PR #6016, opened 2025-03-05
Fix a possible regresion where serializers were not putting required info in message headers

EricWittmann:
It seems like the serializers were failing to store some needed metadata in the message headers.  For example, the JsonSchema serializer was not storing the bean type, resulting in `SimpleJsonSchemaExample` not working.


carlesarnal:
It's not a regression. It's intended. In 2.6 the default value for https://github.com/Apicurio/apicurio-registry/blob/main/serdes/kafka/serde-kafka-common/src/main/java/io/apicurio/registry/serde/config/KafkaSerdeConfig.java#L13 was mistakenly set to true, and we fixed that in 3.0. The right way to solve this is to enable headers in the example,[ like we're doing in the test](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/test/java/io/apicurio/registry/noprofile/serde/JsonSchemaSerdeTest.java#L91). I'm guessing what got overlooked during the testing and releases was the example, not the serde itself.

Issue #5633, opened 2024-12-01
Build failure on aarch64


Dentrax:
## Description

**Registry
Version**: `3.0.4`
**Persistence type**: `-`

`apicurio-registry-go-sdk` build fails because there is [hardcoded amd64](https://github.com/Apicurio/apicurio-registry/blob/88e0eed7ed3b469769bb2fa413ac11afa35598f8/go-sdk/generate.sh#L6) arch on the script, I guess.

But I don't know why there is a build failure for `apicurio-registry-java-sdk` and `apicurio-registry-v2-java-sdk`, after ~1 min.

If I pass `-pl '!apicurio-registry-go-sdk'`, Maven logs says `[ERROR] Could not find the selected project in the reactor: apicurio-registry-go-sdk`. Maybe its because of missing `module` in the `pom.xml`? `-DskipNonJavaGen=true` worked a bit, but still fails for java projects.

### Environment

`aarch64`

### Steps to Reproduce

```
./mvnw clean install \
        -Pprod \
        -DskipTests \
        --no-snapshot-updates \
        --no-transfer-progress \
        --fail-fast -e -X
```

1. Expect a success build on amd64
2. Expect a build failure on aarch64
3. Pass `-DskipNonJavaGen=true` and see `apicurio-registry-go-sdk` will `SUCCESS` but `apicurio-registry-java-sdk` and `apicurio-registry-v2-java-sdk` still fails
4. Pass `-DskipKiota=true` and expect same error (as per recommendation by GPT)

If the error cannot be easily reproduced, make sure to fill in the Environment and Logs sections.
You can also try to reproduce the bug with a fresh Apicurio Registry instance.
-->

### Expected vs Actual Behaviour

It should build for `aarch64`?

### Logs

<details>
  <summary>Log for apicurio-registry-go-sdk</summary>
  
  ```javascript

Reactor Summary for apicurio-registry 3.0.4:

apicurio-registry .................................. SUCCESS [ 31.022 s]
apicurio-registry-config-definitions ............... SUCCESS [03:23 min]
apicurio-registry-common ........................... SKIPPED
apicurio-registry-config-index ..................... SKIPPED
apicurio-registry-config-index-deployment .......... SKIPPED
apicurio-registry-utils-kafka ...................... SUCCESS [03:23 min]
apicurio-registry-schema-util-common ............... SKIPPED
apicurio-registry-schema-util-json ................. SKIPPED
[...]

PR #6020, opened 2025-03-06
fix: build failure on aarch64 #5633

jsenko:
Fixes https://github.com/Apicurio/apicurio-registry/issues/5633

### DIFF
--- a/.github/workflows/release-images.yaml
+++ b/.github/workflows/release-images.yaml
@@ -94,7 +94,7 @@ jobs:
           distribution: 'temurin'
 
       - name: Set up Node.js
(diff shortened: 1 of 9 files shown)


jsenko:
> Not sure if this is needed, but there are GH Hosted runners for linux + ARM available in preview: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

yeah, I've used one for debugging. We might run a test workflow with `runs-on: ubuntu-22.04-arm` if you think it's worth it.

PR #6050, opened 2025-03-18
refactor(operator): replace use of CoreDNS (not supported by OCP) with host aliases

jsenko:
I've been using OCP for development and testing for some time, and I can't do it with CoreDNS, so I found an alternative.

### DIFF
--- a/operator/controller/src/test/java/io/apicurio/registry/operator/it/AuthITTest.java
+++ b/operator/controller/src/test/java/io/apicurio/registry/operator/it/AuthITTest.java
@@ -40,6 +40,8 @@ public class AuthITTest extends BaseAuthTest {
 
         client.resource(registry).create();
 
+        hostAliasManager.defineHostAlias(registry);
+
         // Assertions, checks registry deployments exist
         checkDeploymentExists(registry, COMPONENT_APP, 1);
         checkDeploymentExists(registry, COMPONENT_UI, 1);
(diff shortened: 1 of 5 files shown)


jsenko:
@carlesarnal Since I've already been looking at the auth tests, I also tried to improve them by adding a way to make on-cluster requests (since we cannot easily test through Ingress without provisioning proper certificates).

[The snippet that I've added](https://github.com/Apicurio/apicurio-registry/pull/6050/files#diff-cbfdc3b7b39f9b194e75a9893fd5ef925679ef9c48598337e698791b9a293d61R84) works, but I couldn't find a way to issue a request that will pass authorization. Do you know how we can tests that situation? I've tried basic auth with `curl -u "apicurio-registry:test1"` credentials, but it didn't work.

carlesarnal:
> @carlesarnal Since I've already been looking at the auth tests, I also tried to improve them by adding a way to make on-cluster requests (since we cannot easily test through Ingress without provisioning proper certificates).
> 
> [The snippet that I've added](https://github.com/Apicurio/apicurio-registry/pull/6050/files#diff-cbfdc3b7b39f9b194e75a9893fd5ef925679ef9c48598337e698791b9a293d61R84) works, but I couldn't find a way to issue a request that will pass authorization. Do you know how we can tests that situation? I've tried basic auth with `curl -u "apicurio-registry:test1"` credentials, but it didn't work.

That won't work. The way Registry expects the credentials to be passes is as the Authorization header and both creds encoded in base64 (common practice). Also note that, in the realm that I added to the tests, the apicurio-registry client does not have the correct role. An example of a valid Authorization header would be something like this: `Basic YWRtaW4tY2xpZW50OnRlc3Qx`.

carlesarnal (approved):

jsenko:
The `curl -u` actually adds the `Basic` header. There was an error in Registry log something about "public client not being able to do something with Keycloak". But missing roles make sense.

PR #6084, opened 2025-03-27
Update ccompat api

carlesarnal:
This PR updates the ccompat API to the latest version. Note that this only means adding the missing parameters to the API, not actually implementing the behaviour. This allows upgrading the libraries to the latest versions. Features like metadata, tags, or any feature around data contracts in general won't work for now, it'll come in a future PR.


EricWittmann:
If you want to make a project board to list out the missing ccompat features, I'd be willing to contribute.

EricWittmann (approved):

PR #6153, opened 2025-04-23
Operator CRD updates/refactoring

jsenko:
CRD name/structure adjustments without functional changes. Commit messages should be descriptive enough, but I've also created before/after gists:

Before: https://gist.github.com/jsenko/8bc52ba165b2e5483095cf9edc5253e5
After: https://gist.github.com/jsenko/cf23ae1a3728f2a29d72a2e2b39649a2

Let me know if I forgot something we have discussed.


app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
```diff
@@ -80,12 +80,14 @@ public class AppAuthenticationMechanism implements HttpAuthenticationMechanism {
     @Info(category = "auth", description = "Enable basic auth", availableSince = "1.1.X-SNAPSHOT", registryAvailableSince = "3.X.X.Final", studioAvailableSince = "1.0.0")
     boolean basicAuthEnabled;
 
+    // TODO: Add suffix?
```
jsenko, line 83: We can also update these config properties from `apicurio.authn.basic-client-credentials.cache-expiration` to `apicurio.authn.basic-client-credentials.cache-expiration-mins` or `apicurio.authn.basic-client-credentials.cache-expiration-minutes`. It's not obvious at first glance what unit is it.

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
```diff
@@ -80,12 +80,14 @@ public class AppAuthenticationMechanism implements HttpAuthenticationMechanism {
...
     @ConfigProperty(name = "apicurio.authn.basic-client-credentials.cache-expiration", defaultValue = "10")
-    @Info(category = "auth", description = "Default client credentials token expiration time.", availableSince = "0.1.18-SNAPSHOT", registryAvailableSince = "2.2.6.Final", studioAvailableSince = "1.0.0")
+    @Info(category = "auth", description = "Default client credentials token expiration time in minutes.", availableSince = "0.1.18-SNAPSHOT", registryAvailableSince = "2.2.6.Final", studioAvailableSince = "1.0.0")
     Integer accessTokenExpiration;
 
+    // TODO: Add suffix?
```
jsenko, line 88: This is seconds on the other hand, we can also add a suffix.

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
EricWittmann: Not sure we should rename config properties at this point, since 3.0 is already released upstream.  wdyt?

EricWittmann (approved):
lgtm

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
jsenko: That's fair, I was only thinking about GA... Well if anyone complains we can go through a proper deprecation process. I updated the docs at least.

PR #6175, opened 2025-04-29
upgrade to protobuf-java v4

abhimoondra:
### Notes
1. [php_generic_services option has been removed](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto#L519-L520), removed the references for that. 
2. makeExtensionsImmutable is no longer available, removed that as well.
3. Disabled the ConfluentClientTest & ConfluentSerdeTest since the underlying confluent libraries not yet support protobuf-java 4 ([planned for June](https://github.com/confluentinc/schema-registry/issues/3047#issuecomment-2825806192))

### DIFF
--- a/app/src/test/java/io/apicurio/registry/noprofile/ccompat/ConfluentSerdeTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/ccompat/ConfluentSerdeTest.java
@@ -9,11 +9,13 @@ import io.confluent.kafka.serializers.protobuf.KafkaProtobufSerializer;
 import io.confluent.kafka.serializers.protobuf.KafkaProtobufSerializerConfig;
 import io.quarkus.test.junit.QuarkusTest;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import java.util.Properties;
 
 @QuarkusTest
+@Disabled("Confluent hasn't released protobuf-java 4 changes yet. They are meant to be released with CP 8.x which is not out yet.")
 public class ConfluentSerdeTest extends AbstractResourceTestBase {
 
     @ConfigProperty(name = "quarkus.http.test-port")
--- a/app/src/test/java/io/apicurio/registry/noprofile/ccompat/rest/v7/ConfluentClientTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/ccompat/rest/v7/ConfluentClientTest.java
@@ -44,6 +44,7 @@ import org.apache.kafka.connect.data.SchemaAndValue;
 import org.apache.kafka.connect.data.Struct;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import java.nio.ByteBuffer;
@@ -74,6 +75,7 @@ import static org.junit.jupiter.api.Assertions.fail;
 
 @QuarkusTest
 @TestProfile(DeletionEnabledProfile.class)
+@Disabled("Confluent hasn't released protobuf-java 4 changes yet. They are meant to be released with CP 8.x which is not out yet.")
 @SuppressWarnings({ "unchecked", "rawtypes" })
 public class ConfluentClientTest extends AbstractResourceTestBase {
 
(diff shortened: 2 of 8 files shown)


EricWittmann:
Regarding point (3) above, I want to make sure I understand.  The issue is that this PR upgrades all Protobuf usages to version 4, including the `protoc` we use in various places.  So those tests fail because now that we've upgrade protobuf, those tests will be trying to register proto v4 schemas, which the confluent client cannot do.  Yes?

EricWittmann:
Did you (or can you) try any of the rules (Validity, Compatibility, Integrity) to make sure they work with proto v4?

EricWittmann:
@abhimoondra I've approved CI/CD which has shown a few broken integration tests.  Easy fixes I think, if you can take a look at that.

heiko-braun:
Regarding point 3: Would it make sense to lean on a snapshot of the Confluent libs to enable the tests?

Just for the time being, until they release it?

EricWittmann:
If there is a snapshot version available somewhere (are confluent libs in central?  I cannot remember) then I think that could make sense.  @carlesarnal thoughts?

carlesarnal:
> If there is a snapshot version available somewhere (are confluent libs in central? I cannot remember) then I think that could make sense. @carlesarnal thoughts?

I've not been able to find a snapshot build. Disabling the Confluent tests is a no-go, so, unless we find that version, I think we're stuck with proto3 until June.

abhimoondra:
@EricWittmann 

> Regarding point (3) above, I want to make sure I understand. The issue is that this PR upgrades all Protobuf usages to version 4, including the protoc we use in various places. So those tests fail because now that we've upgrade protobuf, those tests will be trying to register proto v4 schemas, which the confluent client cannot do. Yes?

Yes, your understanding is correct.

> Did you (or can you) try any of the rules (Validity, Compatibility, Integrity) to make sure they work with proto v4?

Yes I carried out some manual tests related to that, where I verified Syntax validity, Backward and forward Compatibility and reference integrity. All of the were working same as with the existing version. I also want to highlight that the current version has few issues related to validity and compatibility checks which needs to be fixed. Just to highlight some:

1. If the same tag number is given in a proto, we can register that schema.
2. A schema with negative tag numbers is also registered successfully.
3. If I add a new field to existing protobuf schema, with the compatibility mode set as FORWARD, then the registration fails. But if I select the compatibility mode as BACKWARD or FULL, then I am able to register the new version.

> Easy fixes I think, if you can take a look at that.
Pushed a fix for the broken tests.

abhimoondra:
@EricWittmann @heiko-braun @carlesarnal I have asked https://github.com/confluentinc/schema-registry/issues/3047#issuecomment-2849169240. If it's available I will make changes in the PR to use the snapshot version and enable the tests again.

Meanwhile, I will update the PR description with the same information.

EricWittmann:
> Yes I carried out some manual tests related to that, where I verified Syntax validity, Backward and forward Compatibility and reference integrity. All of the tests were working same as with the existing version. I also want to mention that the current version has few issues related to validity and compatibility checks which needs to be fixed. Just to highlight some:
> 
>     1. If the same tag number is given in a proto, we can register that schema.
> 
>     2. A schema with negative tag numbers is also registered successfully.
> 
>     3. If I add a new field to existing protobuf schema, with the compatibility mode set as FORWARD, then the registration fails. But if I select the compatibility mode as BACKWARD or FULL, then I am able to register the new version.
> 

Would you be willing to open new GitHub issues for these three items?  That way we can better track them. :+1:

abhimoondra:
@EricWittmann I have create the issue : https://github.com/Apicurio/apicurio-registry/issues/6209

I didn't include the last one as it is already fixed. https://github.com/Apicurio/apicurio-registry/pull/6186

abhimoondra:
Hey @EricWittmann, @carlesarnal

Confluent has release v8.0.0. I have made changes to use that version. I also needed to upgrade the wire-schema to fix the tests.

Do you mind taking a look at the changes.

EricWittmann:
Will need to think a little bit about a major version upgrade for confluent, in case that impacts the compatibility api...

EricWittmann (approved):
Sorry for the delay and thanks again for the contribution.  I think this is OK.  We use the confluent libraries mostly for testing, so upgrading to 8.0 is fine.  We do use the confluent libraries in a couple of other areas (e.g. the confluent export utility) but nothing in any primary code paths.  So I think I'm cool with this change.

abhimoondra:
Hey @EricWittmann, could you please let me know when will these changes be released?

I am particularly looking for the SerDes artifacts, so even if there is a SNAPSHOT version that we can use, that will be great.

EricWittmann:
@abhimoondra Version 3.0.10 was released last week and these changes should be included.  I'm hoping to do another release (3.0.11) today.

abhimoondra:
okay, thanks 👍

Issue #5573, opened 2024-11-22
How to push avro schemas/protocols such that it registry uses references and doesn't embed everything into definition


lsegv:
I want to build a custom registry image based on apicurio, this image would preload all my schemas enforcing them (since i will disable pushing artifacts by services). There is one big problem though.

If i simply loop over all my AVSC files (that i generate from AVDL) and upload them one by one then registry does no processing on top of them and does not recognise when 2 messages were imported reused, because the content of AVSC is literally inlined to be self contained.

I started looking for api that would allow me to upload protocol files AVDL or AVPR, but i found no such thing.

When i use a kafka producer/consumer example and let it push the definitions on the go i see that this pushes messages properly (e.g. uses references), but actual code that does this does a lot of work, it figures out all those references and uploads them correctly to registry.

Problem is i cant just let arbitrary java code run during build stage (or at least i dont want to), ideally registry should allow for importing a protocol and then properly store all those definitions and references.

What can i do here other than grabbing the code from kafka serializer? maybe there is an api i do not know about?

EricWittmann:
What are your expectations for how your schemas are laid out locally?  Do you have control over that such that you could maintain some extra metadata?

Another place to look for this type of thing is in our maven plugin.  Especially this part:

https://github.com/Apicurio/apicurio-registry/blob/main/utils/maven-plugin/src/main/java/io/apicurio/registry/maven/RegisterRegistryMojo.java#L130-L138

What do you mean precisely when you say this?

> this image would preload all my schemas enforcing them

I'd like to better understand your use-case/goals.

There is not currently a way to send a bunch of related stuff to registry all at once and have it automatically figure out the details (with references and all that).  It is something we've discussed, but not yet implemented.  Could be an opportunity to collaborate on something like that if you are interested.

lsegv:
Regarding "extra metadata" yes we have full control over whatever we are doing with avro, i assume you mean metadata in the message definitions themselves?

The use case is like this: in production we do not want to let different services to push their message definitions at will, which means schema registry will have to be in its correct state (with all versions of schemas) when its running, there are several reasons for that. 

1. we have 400+ message types, and devs dont want to let all that happen at runtime, they would rather prefer to know that once the SR is started all schemas are there ready to be queried. 

2. we have build pipeline tasks that run current avro schemas committed in current branch against "official" schema registry for that target environment, and if it detects violation on compatibility requirements it will fail the build with proper messages.

[...]

PR #6257, opened 2025-05-26
fix: Wrong groupid registering referenced avro schema with apicurio-registry-maven-plugin

wagonli:
Currently when defining a schema that references another schema, the apicurio-registry-maven-plugin registers the referenced schema using the parent groupid and set the referenced schema artifactid to its fullname (`groupid.artifactid`).
This issue defeats the very purpose of having separate schemas to put some avro object definition in common.

The proposed solution is to create a `ReferenceArtifactIdentifierExtractor` util that is in charge of parsing the artifact id and group id from the external reference `resource` string. An AVRO implementation just break down the resource and separate the artifact id from the group id depending on the position of the last existing `.` according to AVRO 1.12.0 specification. Other implementations are just a default one mimicking the current behavior : groupid = null, artifactid = resource.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,0 +1,8 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+    <extension>
+        <groupId>com.google.cloud.artifactregistry</groupId>
+        <artifactId>artifactregistry-maven-wagon</artifactId>
+        <version>2.2.5</version>
+    </extension>
+</extensions>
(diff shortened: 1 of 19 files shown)


EricWittmann (approved):
Good change, thanks.  As it turns out, the ArtifactTypeUtilProvider is not where this Maven Plugin functionality should live.  At least I don't think so.  But that's an observation separate from this PR, since there was already a precedent for this and you were just following the existing pattern.  Just a note to say that I'll probably refactor this in the future to separate out the type-specific functionality used by the Maven plugin vs. the type-specific functionality used by the registry server.  They are different and I think there are good reasons to separate them.

But again, that's just a side note.  This change is nice for Avro, and in the future we can consider how to do similar things for the other types (protobuf seems similar, for example).

EricWittmann:
That's interesting!  I'd be happy to collaborate on this if you would like, because that is the sort of use-case we would like the maven plugin to be able to easily handle.  The "autorefs" feature of the plugin was intended to be a step in this direction, but without community comments (like these) it wasn't clear whether there was much interest.

If you are interested, maybe you could you put together a GitHub repository containing a representative collection of schemas that should be parsed and registered?  We can use that as the basis/POC for an improved maven plugin experience.

Also, you may be interested in the discussion here:

* https://github.com/Apicurio/apicurio-registry/issues/5573

It's about how to create immutable container images of Apicurio Registry + all registered schemas. Essentially by populating an Apicurio Registry instance with schemas, then exporting the contents of the registry to a ZIP file, then building a new container image using apicurio-registry as the base image and including the .ZIP.  You can read more about it in the linked discussion if that sounds interesting.

PR #6263, opened 2025-05-28
Fix for JIRA ticket 1062: deleting subjects that have inbound refs using the ccompat API [2.6.x]

EricWittmann:
* https://issues.redhat.com/browse/IPT-1062

```
If schema A contains a reference to schema B, then it should be less easy
than it is, to delete schema B whilst A still exists. Schema A will be left with
a dangling reference to a non-existent schema.

I imagine that there will be times when it is necessary to delete a schema
that is referenced by others. However, at present it seems to make no
difference whether the reference exists or not, and the API does not (so far
as I can tell) allow for any control over the way references are interpreted.
```

### DIFF
--- a/app/src/test/java/io/apicurio/registry/noprofile/Ticket1062Test.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/Ticket1062Test.java
@@ -136,7 +136,7 @@ public class Ticket1062Test extends AbstractResourceTestBase {
         Assertions.assertNotNull(addressVMD);
         Assertions.assertEquals(addressId, addressVMD.getId());
 
-        // Try to delete the "contact" artifact itself - using the ccompat API - should fail
+        // Try to delete the "address" artifact itself - using the ccompat API - should fail
         Assertions.assertThrows(RestClientException.class, () -> {
             confluentClient.deleteSubject(Map.of(), addressId);
         });


app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/SubjectsResourceImpl.java
```diff
@@ -110,6 +111,12 @@ public Schema findSchemaByContent(String subject, SchemaInfo request, Boolean no
     public List<Integer> deleteSubject(String subject, Boolean permanent, String groupId) throws Exception {
         GA ga = getGA(groupId, subject);
 
+        List<Long> globalIdsReferencingArtifact = getStorage().getGlobalIdsReferencingArtifact(ga.getGroupId(), ga.getArtifactId());
+        if (!globalIdsReferencingArtifact.isEmpty()) {
+            // There are other schemas referencing this subject, it cannot be deleted.
+            throw new ReferenceExistsException(String.format("There are subjects referencing %s", subject));
```
jsenko, line 117: We could provide a list of subjects+versions / GAVs to the user so they know what causes the problem.

integration-tests/src/test/java/io/apicurio/tests/ConfluentBaseIT.java
```diff
@@ -68,16 +68,15 @@ public int createArtifactViaConfluentClient(ParsedSchema schema, String artifact
...
-                if (e.getStatus() == 404) {
-                    //subjects may be already deleted
-                    continue;
+        if (confluentService != null && !confluentService.getAllSubjects().isEmpty()) {
+            for (String confluentSubject : confluentService.getAllSubjects()) {
+                logger.info("Deleting confluent schema {}", confluentSubject);
+                try {
+                    registryClient.deleteArtifact("default", confluentSubject);
+                } catch (Exception e) {
+                    System.out.println("WARNING>> Failed to delete confluent schema " + confluentSubject);
```
jsenko, line 77: Are there other tests that rely on the cleanup? I think we should keep the original check, and either fail the test or recursively delete the referring schemas. In the first case the test author would have to cleanup, the second case would be more difficult to implement and depend on finding out the dependencies from the error message.

app/src/test/java/io/apicurio/registry/noprofile/Ticket1062Test.java
```diff
@@ -0,0 +1,157 @@
...
+
+        // Try to delete the "address" artifact, version 1 - using the ccompat API - should fail
+        Assertions.assertThrows(RestClientException.class, () -> {
+            confluentClient.deleteSchemaVersion(Map.of(), addressId, "1");
+        });
+        VersionMetaData addressVMD = clientV2.getArtifactVersionMetaData(groupId, addressId, "1");
+        Assertions.assertNotNull(addressVMD);
+        Assertions.assertEquals(addressId, addressVMD.getId());
+
+        // Try to delete the "contact" artifact itself - using the ccompat API - should fail
```
jsenko, line 139: Typo: Should be "address". That confused me for a moment:)

jsenko (approved):
LGTM with optional suggestions.

Also, do we have an equivalent on the Registry 3 API side? I don't think the integrity rule checks this when deleting. We should create a new issue.

integration-tests/src/test/java/io/apicurio/tests/ConfluentBaseIT.java
EricWittmann: Multiple tests rely on the cleanup and fail if it doesn't work.  Before making this change, I tried making multiple passes (using a while loop instead of an if block) - so it would keep iterating over all the subjects until it couldn't find any more, but that didn't work and I didn't spend time figuring out why.  :(

Making the tests responsible for their own cleanup is sensible.  I would be happy with such an improvement, although this change represents an "equivalent" approach to what we did before, which I think is sufficient for this PR.

app/src/test/java/io/apicurio/registry/noprofile/Ticket1062Test.java
EricWittmann: Nice catch. :) Fixed.

PR #6271, opened 2025-05-29
Reduce adoc to make contribution easier

pwright:
reduces file count in main docs


pwright:
@EricWittmann returning to draft status, spotted some small issues

PR #6313, opened 2025-06-10
Configure Maven GPG passphrase via env variables

NikishaPatil:


jsenko (approved):
LGTM

jsenko (changes requested):
I missed this, but shouldn't the environment variable be available in the `release-maven-artifacts.yaml` rather than `release.yaml`, or both?

jsenko:
Can you add it to the other workflow as well please?

EricWittmann (dismissed):

EricWittmann (approved):

Issue #877, opened 2020-09-25
Serdes client does not check on schema state


tagarr:
The serdes client does not check the state of a schema.

The serializer should throw an exception if the schema's state is disabled and log an error if the schema is deprecated

The deserializer should log an error if the schema is disabled and log a warning if deprecated.

Difference in behaviour is that schema admins don't see reading messages from older schemas as dangerous as writing new ones

EricWittmann:
Right now the server will respond with a 404 if the schema version is disabled.  That will result in a failure on the client/serdes for sure.

We were planning on the deprecation warning happening in the client:  https://github.com/Apicurio/apicurio-registry/issues/840

But that wouldn't distinguish between the serializer or deserializer.

EricWittmann:
Thoughts on this @Apicurio/developers or @Apicurio/ibm-developers ?

tagarr:
Actually only consumers get a 404, producers get back the schema and can still send messages. I think it would be could for client to know and inform the application the state of the schema. I think if using error codes then maybe 406/409 do differentiate between it being there / not able to be used

carlesarnal:
I'm in favour of the client being responsible for the state checking and to trying to avoid the error codes.

famartinrh:
good candidate for 2.0?

EricWittmann:
Probably not needed for RC1 though.

PR #6320, opened 2025-06-10
Ignore disabled artifact version when searching artifact by content 

wagonli:
fixes https://github.com/Apicurio/apicurio-registry/issues/877

In the registry facade impl, add a check on the artifact state to filter out DISABLED artifacts.
Add unit tests dependencies (mockito + assertj) in schema resolver module
Add test case for the v3 registry client facade around ignoreing disabled artifacts versions

### DIFF
--- a/schema-resolver/pom.xml
+++ b/schema-resolver/pom.xml
@@ -36,6 +36,16 @@
       <artifactId>junit-jupiter</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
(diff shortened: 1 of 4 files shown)


wagonli:
:warning: This could be some breaking changes for some configuration. 

But I think it should be intended to break in theses cases from the start : it feels more logical to refuse to resolve disabled schemas that otherwise only resolves because they are based on an embedded schema with the target record (like the record generated by the `avro-maven-plugin`).

wagonli:
Hello @jsenko did you manage to get a moment to have a quick look at the PR ? :pray:

EricWittmann (approved):
This PR looks good to me.  Nice test, even thought we don't often use mockito. It's definitely easier and faster than using a Quarkus test.

I do think it might be better if we added state filtering to the `/search/versions` endpoint.  That would let us filter out the versions on the server-side rather than doing it in the client.

I'll merge this PR but will add a FIXME to consider moving the filtering capability to the server.

EricWittmann:
@wagonli I'm going to cut a release today as well, so this will be included.  FYI

Issue #6329, opened 2025-06-12
Microsoft Entra login - 401


krystofrezac:
## Description

Login with Microsoft Entra doesn't work.
1. The redirect URI is now set to `<base-url>/explore`, even though our configuration (`APICURIO_UI_AUTH_OIDC_REDIRECT-URI`) doesn't contain `/explore`
2. After attempting to log in, a 401 error is returned. The UI displays the following error:

<img width="718" alt="Image" src="https://github.com/user-attachments/assets/2954c39b-4424-4874-9a92-0327e21bf80a" />

```json
{
   "responseHeaders": {
      "cache-control": [
         "no-cache",
         " no-store",
         " must-revalidate"
      ],
      "content-length": [
         "0"
      ],
      "date": [
         "Thu",
         " 12 Jun 2025 05:37:01 GMT"
      ],
      "expires": [
         "Wed",
         " 11 Jun 2025 05:37:01 GMT"
      ],
      "pragma": [
         "no-cache"
      ],
      "strict-transport-security": [
         "max-age=31536000; includeSubDomains"
      ],
      "www-authenticate": [
         "Bearer"
      ]
   },
   "responseStatusCode": 401
}
```

The login flow works correctly in version `3.0.7` with the same configuration.

**Registry
Version**: 3.0.9
**Persistence type**: postgresql

### Environment

Configuration:
```
# registry
    QUARKUS_OIDC_TENANT-ENABLED: 'true'
    QUARKUS_HTTP_CORS_ORIGINS: "*"
    APICURIO_AUTH_ROLE-BASED-AUTHORIZATION: 'true'
    QUARKUS_OIDC_ROLES_ROLE-CLAIM-PATH: roles
    APICURIO_AUTH_ANONYMOUS-READ-ACCESS_ENABLED: 'true'
    APICURIO_STORAGE_KIND: sql
    APICURIO_STORAGE_SQL_KIND: postgresql

    QUARKUS_OIDC_CLIENT-ID: <client-id>
    QUARKUS_OIDC_AUTH-SERVER-URL: https://login.microsoftonline.com/<tenant-id>/v2.0
    APICURIO_UI_AUTH_OIDC_CLIENT-ID: <client-id>
    APICURIO_UI_AUTH_OIDC_REDIRECT-URI: https://<base-url>

# ui
REGISTRY_API_URL: https://<base-url>/apis/registry/v3
```

### Steps to Reproduce

1. Open the Registry UI.
[...]

PR #6339, opened 2025-06-16
feat(auth): Allow the auth token type to be configurable ID vs access

EricWittmann:
This PR is a result of this issue:

* https://github.com/Apicurio/apicurio-registry/issues/6329

While I think the correct solution to that issue **might** be to configure Entra to issue OIDC compliant access tokens, this PR introduces an opt-in feature to configure the UI to use the ID token when making authenticated REST calls to the registry backend (instead of using the access token).


EricWittmann:
@knutejohKLP Upstream release 3.0.13 workflow running now.

knutejohKLP:
@EricWittmann Great :) Any news on the redhat product release? :)

EricWittmann:
We're working through it still.  I am expecting a new release within the next two weeks.

PR #6365, opened 2025-06-25
feat(serdes): Support for v2 servers in v3 serdes layer

EricWittmann:
To better support backwards compatibility and migrations from v2 to v3, this PR is a proposal to utilize both the v2 and v3 generated REST API clients in the SerDe layer so we can support both versions of the Registry server.

Essentially, this proposal creates an abstraction between the SchemaResolver and the REST API client, allowing an impl of that abstraction to use the v2 generated client and another impl that uses the v3 generated client.

What this means is that v3 Serializers/Deserializers can support both v2 and v3 Apicurio Registry servers at the same time.  It will detect the version to use based on the configured Registry URL, and use the correct impl.

Without this change, v3 SerDe classes **cannot** work with v2 Apicurio Registry servers.  This means that when upgrading from v2 to v3, users must upgrade the Registry server first, and then upgrade all of their client applications after.  This change allows upgrading in either direction.


schema-resolver/src/main/java/io/apicurio/registry/resolver/DefaultSchemaResolver.java
```diff
@@ -267,35 +250,16 @@ private SchemaLookupResult<S> handleResolveSchemaByContent(ParsedSchema<S> parse
...
-
-                if (results.getCount() == 0) {
+            String artifactType = schemaParser.artifactType();
+
+            List<RegistryVersionCoordinates> versions = this.clientFacade.searchVersionsByContent(rawSchemaString,
+                    artifactType, artifactReference, true);
+            if (versions.isEmpty()) {
+                versions = this.clientFacade.searchVersionsByContent(rawSchemaString, artifactType, artifactReference, false);
+
+                if (versions.isEmpty()) {
```
wagonli, line 262: this may  be the opportunity to include changes proposed in https://github.com/Apicurio/apicurio-registry/pull/6320
And ensure the version is not disabled before continuing ?

EricWittmann:
@wagonli This is merged in `main` now.  FYI

wagonli:
> @wagonli This is merged in `main` now. FYI

ok thx, will make the changes in my PR

PR #6401, opened 2025-07-12
fix(ts-sdk): fix native ESM tsconfig compatibility

Hoishin:
In the TypeScript SDK, current combination of settings in package.json and tsconfig.json prevents users from using native Node.js ESM written in TypeScript.

## Why

**Official documentation reference: https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library**
> In short, "moduleResolution": "bundler" is infectious, allowing code that only works in bundlers to be produced. Likewise, "moduleResolution": "nodenext" is only checking that the output works in Node.js, but in most cases, module code that works in Node.js will work in other runtimes and in bundlers.

### In short:

The TypeScript SDK project has
- `"type": "module"` in package.json
- `"moduleResolution": "bundler"` in tsconfig.json

As a Node.js compatible pure-ESM library, it should not set `"bundler"` in tsconfig.json, even if it is using bundler in the package generation.
If you set `"moduleResolution": "bundler"`, it allows you to write an invalid ES modules import declarations (i.e. without `.js` extensions) in source files. This results in ES modules incompatible type declarations and JavaScript outputs when the library is used in native ESM setup in Node.js.

## Reproduction

https://github.com/Hoishin/apicurio-registry-native-esm-issue


EricWittmann:
Thanks @Hoishin - can you provide a little bit more context?  What is the use-case you have where the current package is failing?

If this change is accepted, what are the consequences to browser applications that use the sdk package?  The main consumer of the SDK right now is the registry UI (React application).  I would **like** the sdk to work in any browser application and also any node.js application.  Is this change consistent with that goal?

Hoishin:
Hello @EricWittmann, the use-case is as stated in the OP, Node.js projects written in TypeScript with the intention to run it as native ES modules. That means, you write TypeScript for Node.js with package.json having `"type": "module"`, transpile it with the TypeScript official `tsc` command line, and run output JavaScript file with `node`.

This change does not affect this library's JavaScript output, because JavaScript is built with Vite. This PR changes only contents inside `.d.ts` files to be more correct. There is no changes to behavior of SDK used in browsers. You can check by running build in this PR branch.

>  I would like the sdk to work in any browser application and also any node.js application

This PR only has positive effect in this regard. Currently, you cannot use this library for Node.js without a bundler (i.e. esbuild, swc, webpack), because `tsc` does not allow import without an extension.
With this change, you can:
1. write TypeScript and transpile it with `tsc` without a bundler and run in native ESM runtime
2. write TypeScript and run it with Node.js type stripping/transform feature with no build step.

EricWittmann:
@Hoishin Merged - thanks for the contribution.  :+1: 

If you have a node.js application using the typescript sdk, we'd love to see the use-case/example!

PR #6433, opened 2025-07-24
fix: incorrect Avo converter caching (forwardport of 6fc6cc47)

jsenko:
NOTE: It seems the 3.x does not containe the changes made by bebcf820 PR#5637, but we can at least add the test.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
@@ -25,6 +25,7 @@ import io.apicurio.registry.model.GA;
 import io.apicurio.registry.model.GAV;
 import io.apicurio.registry.model.GroupId;
 import io.apicurio.registry.model.VersionId;
+import io.apicurio.registry.rest.ConflictException;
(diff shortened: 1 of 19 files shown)


app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -789,45 +790,56 @@ private Long ensureContentAndGetId(String artifactType, ContentWrapperDto conten
...
             }
 
             // If we get here, then the content was inserted and we need to insert the references.
-            insertReferencesRaw(handle, contentId, references);
+            insertReferencesRaw(contentId, references);
         });
     }
 
-    private void insertReferencesRaw(Handle handle, Long contentId, List<ArtifactReferenceDto> references) {
+    private void insertReferencesRaw(Long contentId, List<ArtifactReferenceDto> references) {
```
EricWittmann, line 822: Can you explain this change?  Won't this move the insertion of the references into a new Transaction?  As a result, any rollback (e.g. when using dryRun=true) would fail to roll back the references?

Or am I not remembering how handles work?

EricWittmann (approved):

PR #6500, opened 2025-08-15
test(operator): add multi-namespace smoke test

jsenko:


jsenko:
hi @EricWittmann this is a simple multi-namespace test and it seem to be working. When I get back, prepare a reproducer of your issue and I'll investigate.

PR #6516, opened 2025-08-24
Add ability to create drafts from the maven plugin

knutejohKLP:
Adding draft versions using the maven plugin is currently not possible. This adds this ability by extending the configuration of the plugin.

### DIFF
--- a/utils/maven-plugin/src/main/java/io/apicurio/registry/maven/RegisterArtifact.java
+++ b/utils/maven-plugin/src/main/java/io/apicurio/registry/maven/RegisterArtifact.java
@@ -17,6 +17,7 @@ public class RegisterArtifact {
     private Boolean minify;
     private Boolean analyzeDirectory;
     private Boolean autoRefs;
+    private Boolean isDraft;
     private String contentType;
     private List<RegisterArtifactReference> references;
     private List<ExistingReference> existingReferences;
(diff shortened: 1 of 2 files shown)


knutejohKLP:
@EricWittmann Sorry for the extra commit, but we weren't able to update any draft as the plugin would constantly get an error when trying to post a version with new content. Could you check the latest changes?

EricWittmann (changes requested):
Can you move the update logic into the catch?  Maybe invert the if logic and add an else?  And also return `VersionMetaData` even on an update.  Like this:

```java
            // If this is a draft, and we got a 409, then we should try to update the draft instead.
            if (artifact.getIsDraft() && e.getResponseStatusCode() == 409) {
                registryClient.groups().byGroupId(groupId)
                        .artifacts().byArtifactId(artifactId)
                        .versions().byVersionExpression(version)
                        .content().put(content, config -> {});
                getLog().info(String.format("Successfully updated artifact [%s] / [%s].", groupId, artifactId));
                // Return version metadata
                return registryClient.groups().byGroupId(groupId)
                        .artifacts().byArtifactId(artifactId)
                        .versions().byVersionExpression(version)
                        .get();
            } else {
                logAndThrow(e);
                return null;
            }
```

Note: this is from memory so there may be errors in the code.  :)

knutejohKLP:
Sure, will update the code :)

EricWittmann (approved):
:+1:

PR #6658, opened 2025-09-30
fix: Add client scope support for OAuth2 authentication for maven plugin

Ben10k:
Authentication to Entra-enabled apicurio-registry was failing with an exception, when using the maven plugin:
```
[ERROR] Exception while registering artifact [default] / [com.example.package.v1.Model]
java.lang.RuntimeException: java.util.concurrent.ExecutionException: io.vertx.core.impl.NoStackTraceThrowable: invalid_request: AADSTS90014: The required field 'scope' is missing from the credential. Ensure that you have all the necessary parameters for the login request. Trace ID: 4dfc2eb8-cffb-447c-8806-7a127b7a1200 Correlation ID: e72690ec-ffcf-49c9-b080-b657ad905b26 Timestamp: 2025-09-30 19:28:17Z
```

After some digging around, I discovered, that the plugin was reading the `clientScope` parameter, but it is not being used.

I have also added additional logging, to see which auth, if any, is being used when executing plugin actions.


EricWittmann:
Thanks for the contribution @Ben10k

PR #6665, opened 2025-10-01
fix: allow read-only users to use register endpoint for existing schemas

AndreaScarselliAx:
The register endpoint now checks READ permission first when looking up existing schemas, and only checks WRITE permission when creating new ones. This fixes the issue where Kafka producers using Confluent Serializer with `auto.register.schemas=true` would fail even when the schema already exists and the user has READ permissions.


EricWittmann:
I was skeptical when I read the title of this PR, but then I read the explanation and was like "oh yeah, that's a nice change".

So thanks @AndreaScarselliAx - nice contribution.  :+1:

Issue #6670, opened 2025-10-02
`dryRun` parameter is ignored when using `ifExists=FIND_OR_CREATE_VERSION` on artifact create


zoni:
## Description

**Registry Version**: 3.0.15, also present on HEAD (f737289d1451ab3fcc89edae380e97329a64dd8d)
**Persistence type**: N/A (easily reproducible with in-memory)

The `dryRun` parameter is not respected when using `ifExists=FIND_OR_CREATE_VERSION` in the createArtifact endpoint. When calling the endpoint with `dryRun=true` on an already-existing artifact, a new version is actually created and persisted instead of performing a true dry-run simulation.

The cause is that the dryRun parameter is not being propagated through the method call chain in `GroupsResourceImpl.java` when handling the FIND_OR_CREATE_VERSION case.

### Environment

N/A - bug can be reproduced following the development instructions from the project README.

### Steps to Reproduce

1. Create an initial artifact via the createArtifact API endpoint. Its contents or type don't matter.
2. Call the createArtifact endpoint with the same group/artifact ID as was used to create the artifact from step 1, but now with different artifact contents, including the following parameters:
   - `dryRun=true`
   - `ifExists=FIND_OR_CREATE_VERSION`
   - `canonical=true`
3. Observe that a new version is actually created and persisted, while this should have been a dry-run only.

The following shell script reproduces this programmatically (assumes local registry on port 8080, requires curl and jq):

```bash
#!/bin/bash

API_ENDPOINT="http://localhost:8080/apis/registry/v3"
GROUP_ID="test-group"
ARTIFACT_ID="example-artifact"
ARTIFACT_TYPE="AVRO"

# Initial schema
INITIAL_SCHEMA='{
  "type": "record",
  "name": "ExampleRecord",
  "namespace": "example",
  "fields": [
    {"name": "id", "type": "string"},
    {"name": "value", "type": "int"}
  ]
}'

# Updated schema (adds a new field)
UPDATED_SCHEMA='{
  "type": "record",
  "name": "ExampleRecord",
  "namespace": "example",
  "fields": [
    {"name": "id", "type": "string"},
    {"name": "value", "type": "int"},
    {"name": "timestamp", "type": "long", "default": 0}
  ]
}'

echo "=== Step 1: Create initial artifact (version 1) ==="
INITIAL_PAYLOAD=$(jq -n \
  --arg artifactId "$ARTIFACT_ID" \
[...]

PR #6671, opened 2025-10-02
Fix dryRun parameter handling in createArtifact with FIND_OR_CREATE_VERSION

zoni:
The dryRun parameter is not respected when using ifExists=FIND_OR_CREATE_VERSION in the createArtifact endpoint. When calling the endpoint with dryRun=true on an already-existing artifact, a new version is actually created and persisted instead of performing a true dry-run simulation, as described in https://github.com/Apicurio/apicurio-registry/issues/6670.

The cause is that the dryRun parameter is not being propagated through the method call chain in GroupsResourceImpl.java when handling the FIND_OR_CREATE_VERSION case.

Included is a new test for the specific ifExists=FIND_OR_CREATE_VERSION scenario, as well as fix to an existing test that used ifExists=CREATE_VERSION which was validating the buggy, unintended behavior.

Fixes https://github.com/Apicurio/apicurio-registry/issues/6670

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v3/GroupsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v3/GroupsResourceImpl.java
@@ -1017,7 +1017,7 @@ public class GroupsResourceImpl extends AbstractResourceImpl implements GroupsRe
 
             return rval;
         } catch (ArtifactAlreadyExistsException ex) {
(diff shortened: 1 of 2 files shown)


EricWittmann:
Nice job @zoni - thanks for the contribution.  Much appreciated.  :+1:

EricWittmann (approved):
LGTM :zap:

Issue #6438, opened 2025-07-25
Option to configure proxy settings


0chrisf0:
### Feature or Problem Description

I use apicurio behind a proxy. In order to be able to access Microsoft Azure to retrieve a Bearer token, the http client needs to be configured to go through that proxy. I want a way to configure this when using serde classes (specifically in Java).

### Proposed Solution

Specifically, I know that passing a `WebClientOptions` with the proxy configs to `setHttpClientOptions` on the `oAuth2Options` in `buildOICDWebClient` fixes the issue. I don't know if this is the best way to do this.

### Additional Context

None, but please let me know if this needs clarification.

EricWittmann:
Hey @0chrisf0 sorry for the delay.  Busy busy over here.  :)

I do think you have found the correct solution to the problem (customizing the WebClient).

We would be very happy for a contribution that makes this use-case easier to configure!  👍

EricWittmann:
@carlesarnal when you close an issue can you assign it to the milestone?  Eventually I'd like to standardize on that as a workflow so we can improve our automation in various areas (e.g. release notes).

PR #6801, opened 2025-10-30
Implement proxy support client

carlesarnal:
//Trying Claude to provide a description.

This PR implements proxy support for Apicurio Registry serializers/deserializers and SDK clients, addressing issue #6438. 

This enables the Registry to work in environments where all outbound HTTP/HTTPS connections must go through a proxy server, including OAuth authentication endpoints.

Key points:
- Proxy configuration applies to both Registry API calls and OAuth token endpoint calls
- Configuration supports both integer and string port values for flexibility
- Implementation follows the same pattern as the recently added TLS/SSL support


EricWittmann:
Impl looks reasonable.  I didn't look at the tests to see if those make sense.

PR #6811, opened 2025-10-31
Debezium integration tests

carlesarnal:


EricWittmann:
Are there per-requisites for these tests to work?  If there are, then we need to isolate the tests into a profile and figure out how to run them properly in the apicurio-testing world.

Issue #6832, opened 2025-11-05
Protobuf scoping rules are not checked correctly


jeeger:
## Description

**Registry Version**:  3.1.2
**Persistence type**: memory

When submitting schemas via the ccompat API, protobuf files that use fully qualified types are rejected as incompatible with files that use an unqualified type.

### Environment

Run the Apicurio registry via docker with `docker run -p 8080:8080 apicurio/apicurio-registry:3.1.2`. Then, set the global compatibility level to `BACKWARD` via `http PUT :8083/apis/ccompat/v7/config compatibility=BACKWARD` (or the equivalent curl command).

### Steps to Reproduce

- Update a new Protobuf schema to a test subject that uses an unqualified nested type
- Add a protobuf schema to the same test subject that uses a fully qualified nested type

Also see attached `reproduce.sh` script and protobuf files (renamed to txt to be able to upload them).

(edit: Inlined example files to illustrate the issue)

```protobuf
syntax = "proto3";
package test;

message RootMessage {
  message NestedMessage {
      optional uint64 some_id = 1;
  }
  .test.RootMessage.NestedMessage nested_message_field = 1;
}
```
```protobuf
syntax = "proto3";
package test;

message RootMessage {
  message NestedMessage {
      optional uint64 some_id = 1;
  }
  // Same type as above, but fails check.
  NestedMessage nested_message_field = 1;
}
```

[reproduce.sh](https://github.com/user-attachments/files/23359728/reproduce.sh)
[nested-qualified.txt](https://github.com/user-attachments/files/23359741/nested-qualified.txt)
[nested-unqualified.txt](https://github.com/user-attachments/files/23359742/nested-unqualified.txt)

### Expected vs Actual Behaviour

Expected: Old schema is returned
Actual: Schema update is rejected ("Incompatible artifact: test [PROTOBUF], num of incompatible diffs: {1}, list of diff types: [The new version of the protobuf artifact is not backward compatible. at /]")

### Logs
```
INFO exec -a "java" java -XX:MaxRAMPercentage=80.0 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+ExitOnOutOfMemoryError -cp "." -jar /deployments/apicurio-registry-app-3.1.2-runner.jar
INFO running in /deployments
[...]

PR #6833, opened 2025-11-05
Try to deserialize base64 encoded protobuf schemas in ProtobufFile.

### LINKED ISSUE #6831: COMPATIBILITY rule fails with Protobuf schemas registered via Confluent API (ccompat/v7) using FileDescriptorSet format
## Description

When `APICURIO_RULES_GLOBAL_COMPATIBILITY: FORWARD` is enabled, registering new Protobuf schema versions via Confluent-compatible API (`/apis/ccompat/v7`) fails with `InternalServerError 0 Syntax error in :1:1: unexpected label: CiV...`. The error occurs because Confluent .NET Protobuf serializer registers schemas as base64-encoded `FileDescriptorSet` instead of `.proto` text, and the compatibility check parser expects `.proto` source text.

This appears related to #2440, but occurs specifically when global compatibility rules are enabled.

### Environment

- **Apicurio Registry version:** 3.0.9 (using `apicurio/apicurio-registry:3.0.9`)
- **API:** Confluent-compatible `/apis/ccompat/v7`
- **Client:** Confluent.SchemaRegistry.Serdes.Protobuf (.NET) via KafkaFlow
- **Configuration:**
  ```yaml
  APICURIO_RULES_GLOBAL_COMPATIBILITY: FORWARD
  KAFKA_BOOTSTRAP_SERVERS: kafka:29092
  ```

### Steps to Reproduce

1. Configure Apicurio Registry with:
   ```yaml
   environment:
     APICURIO_RULES_GLOBAL_COMPATIBILITY: FORWARD
   ```

2. Register first Protobuf schema version using Confluent .NET client:
   ```csharp
   middlewares.AddSchemaRegistryProtobufSerializer(
       new ProtobufSerializerConfig { 
           SubjectNameStrategy = SubjectNameStrategy.TopicRecord 
       }
   );
   ```

3. First version registers successfully.

4. Modify the `.proto` file (e.g., add a field):
   ```protobuf
   message SomeMessage {
       // ... existing fields
       bool NewField = 15;  // new field
   }
   ```

5. Attempt to register the updated schema (produce a message with the new schema).

Expected 
- Compatibility check should succeed (adding a new optional field in Proto3 is forward-compatible).
- New schema version should be registered successfully.

Actual 
- Registration fails with:
  ```
  InternalServerError 0 Syntax error in :1:1: unexpected label: CiVLYWZrYS9Qcm90b2J1Zi9jaGFyZ2luZy1zZXNzaW9uLnByb3RvEhJBdG9tLkVtc3AuU2Vzc2lvbnM...
  ```
- The string starting with `CiV...` is a base64-encoded `FileDescriptorSet`.

### Logs

```
System.Net.Http.HttpRequestException: [http://apicurio-registry:8080/apis/ccompat/v7/] InternalServerError 0 Syntax error in :1:1: unexpected label: CiVLYWZrYS9Qcm90b2J1Zi9jaGFyZ2luZy1zZXNzaW9uLnByb3RvE*
   at Confluent.SchemaRegistry.RestService.ExecuteOnOneInstanceAsync(Func`1 createRequest)
   at Confluent.SchemaRegistry.RestService.RequestAsync[T](String endPoint, HttpMethod method, Object[] jsonBody)
   at Confluent.SchemaRegistry.RestService.RegisterSchemaWithResponseAsync(String subject, Schema schema, Boolean normalize)
   at Confluent.Shared.CollectionUtils.ConcurrentDictionaryExtensions.GetOrAddAsync[TKey,TValue](ConcurrentDictionary`2 dictionary, TKey key, Func`2 valueFactory)
   at Confluent.SchemaRegistry.CachedSchemaRegistryClient.RegisterSchemaWithResponseAsync(String subject, Schema schema, Boolean normalize)
   at Confluent.SchemaRegistry.Serdes.ProtobufSerializer`1.SerializeAsync(T value, SerializationContext context)
   at KafkaFlow.ConfluentSerializerWrapper.InnerConfluentSerializerWrapper`1.SerializeAsync(Object message, Stream output, ISerializerContext context)
   at KafkaFlow.Middlewares.Serializer.SerializerProducerMiddleware.Invoke(IMessageContext context, MiddlewareDelegate next)
   at KafkaFlow.Producers.MessageProducer.ProduceAsync(String topic, Object messageKey, Object messageValue, IMessageHeaders headers, Nullable`1 partition)
   at KafkaFlow.Producers.MessageProducer.ProduceAsync(String topic, Object messageKey, Object messageValue, IMessageHeaders headers, Nullable`1 partition)
   at MediatR.NotificationPublishers.ForeachAwaitPublisher.Publish(IEnumerable`1 handlerExecutors, INotification notification, CancellationToken cancellationToken)
```

jeeger:
When checking compatibility rules, the new Protobuf schema is converted from base64-encoded form into a parsed ProtoFile (these changes were made in PR #6124). However, the same is not done for the stored schemas. This change adds this conversion to the ProtobufFile class (try to parse as text, deserialize via base64 and FileDescriptorProto if this fails). This fixes rule validation, but might expose bugs w.r.t nested type names as reported in issue #6832.

This is a very quick fix, as I'm unclear about the architectural implications of this change, but I hope it's a useful point to start discussion.

This should also fix #6831 if no further problems regarding type qualification are exposed.

### DIFF
--- a/utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/ProtobufFile.java
+++ b/utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/ProtobufFile.java
@@ -4,6 +4,8 @@ import com.google.common.collect.ContiguousSet;
 import com.google.common.collect.DiscreteDomain;
 import com.google.common.collect.Range;
 import com.google.common.io.Files;
+import com.google.protobuf.DescriptorProtos;
+import com.google.protobuf.InvalidProtocolBufferException;
 import com.squareup.wire.Syntax;
 import com.squareup.wire.schema.Location;
 import com.squareup.wire.schema.internal.parser.EnumConstantElement;
@@ -21,11 +23,7 @@ import com.squareup.wire.schema.internal.parser.TypeElement;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 /**
  * Indexed representation of the data resulting from parsing a single .proto protobuf schema file, used mainly
@@ -72,8 +70,20 @@ public class ProtobufFile {
     }
 
     public static ProtoFileElement toProtoFileElement(String data) {
-        ProtoParser parser = new ProtoParser(Location.get(""), data.toCharArray());
-        return parser.readProtoFile();
+        try {
+            ProtoParser parser = new ProtoParser(Location.get(""), data.toCharArray());
+            return parser.readProtoFile();
+        } catch (Exception e) {
+            // Exctracted from AbstractResource.java, lines 138-149.
+            byte[] decodedBytes = Base64.getDecoder().decode(data);
+            DescriptorProtos.FileDescriptorProto descriptorProto = null;
+            try {
+                descriptorProto = DescriptorProtos.FileDescriptorProto.parseFrom(decodedBytes);
+            } catch (InvalidProtocolBufferException ex) {
+                throw new RuntimeException(ex);
+            }
+            return FileDescriptorUtils.fileDescriptorToProtoFile(descriptorProto);
+        }
     }
 
     public String getPackageName() {


EricWittmann:
## Question about this approach

I've been analyzing this PR and I **think** there might be a better place to handle the base64 decoding.  But
I may be wrong because I don't have too much experience in this area.

**Context:** .NET clients send Protobuf schemas as base64-encoded `FileDescriptorProto` (binary format), but
according to the Confluent Schema Registry API spec, schemas should be stored and returned as `.proto` text
format by default (binary is only returned with `format=serialized` query parameter).

**Current situation:** `AbstractResource.java:130-150` already converts base64→text for rule checking, but then
stores the original base64 content (lines 160-173). This means retrieved schemas are still base64, which is why
`ProtobufFile` needs the fallback logic added in this PR.

**Alternative approach:** What if we normalize the schema format at the REST API layer instead? Specifically,
modify `AbstractResource.java` to store the converted text format (instead of just using it for rules), so that:
1. Storage always contains `.proto` text format (matching Confluent and Core API behavior)
2. `ProtobufFile` and compatibility checking always receive text (no fallback needed)
3. All format conversion happens at the API boundary (single responsibility)

What do you think?

jeeger:
>  Alternative approach: What if we normalize the schema format at the REST API layer instead? Specifically,
modify AbstractResource.java to store the converted text format (instead of just using it for rules), so that:

I'm not an expert, but wouldn't this break retrieving schemas based on content hashes? Submitting a schema in base64-encoded format and then submitting the schema again would result in a new schema, because the hashes aren't identical (but the content is).
One solution would be calculating the hashes on the original content and then decoding, but that would be a major change that could have an impact on lots of other functionality.

EricWittmann:
OK I'm fine with merging this as a quick fix.  However, we do need to create a test for it before we can merge.  I can do that if no one wants to volunteer.  :)

carlesarnal:
> > OK I'm fine with merging this as a quick fix. However, we do need to create a test for it before we can merge. I can do that if no one wants to volunteer. :)
> 
> I'll write a test for it and I'll own the fix on the server side as well.

This is a test for the fix: https://github.com/Apicurio/apicurio-registry/pull/6859

carlesarnal:
This PR addresses the original problem: https://github.com/Apicurio/apicurio-registry/pull/6860

PR #6844, opened 2025-11-10
Improve experience of configuring VertX with Java SDK

### LINKED ISSUE #6840: Pass Vertx context to the client in the MCP server, add logging.
- Users should be passing an existing Vertx instance, but the client creates a new one if it isn't provided. When running in Quarkus, there should already be one available. There are also issues with closing the 
- Document this issue in the client javadocs.
- Add more debug logging to help with failing requests.

jsenko:
Fixes #6840 . I found out that getting the VertX instance from CDI actually works nicely, so I decided to keep it in addition to warnings. I've also added a feature that attempts to normalize provided registry URL. I expect that many clients would use a similar logic, so I moved it from the MCP server to the SDK.

### DIFF
--- a/java-sdk-common/src/main/java/io/apicurio/registry/client/DefaultVertxInstance.java
+++ b/java-sdk-common/src/main/java/io/apicurio/registry/client/DefaultVertxInstance.java
@@ -17,13 +17,21 @@ import java.util.logging.Logger;
 public final class DefaultVertxInstance {
 
     private static final Logger logger = Logger.getLogger(DefaultVertxInstance.class.getName());
-
-    private static Vertx VERTX;
-
-    private static boolean closed = false;
+    private static volatile boolean closed = false;
 
     private DefaultVertxInstance() {
         // Prevent instantiation
+    }
+
+    private static class Holder {
+        private static final Vertx INSTANCE;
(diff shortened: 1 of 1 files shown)


EricWittmann:
Any reason why you got rid of the `Holder` pattern?  Using that pattern means you don't need to synchronize the get method...

jsenko:
> Any reason why you got rid of the `Holder` pattern? Using that pattern means you don't need to synchronize the get method...

I was not familiar with it before and this felt more readable, I'll put it back it's not a big deal.

EricWittmann:
> Edit: Actually there is a small reason - the close method has to always initialize Vertx, because it cannot test whether the holder was used or not without another boolean flag.

Sure but if you're a developer calling the `close()` method on the default vertx instance but you are **not** using the default vertx instance ... then I dunno?  I guess you just live with creating a new vertx instance and then immediately closing it.

I think the holder pattern is worth it to eliminate the synchronized block, even though it's less readable.  I don't care all THAT much though tbh.  :)

EricWittmann:
But otherwise, nice changes.  I like the automatic discovery of the vertx instance via CDI.  :+1:

PR #6860, opened 2025-11-12
fix: Normalize Protobuf schemas to text format at storage time

### LINKED ISSUE #6831: COMPATIBILITY rule fails with Protobuf schemas registered via Confluent API (ccompat/v7) using FileDescriptorSet format

carlesarnal:
## Summary
Normalize Protobuf schemas to text format at storage time to ensure consistent content hashing and prevent duplicate schema versions when clients submit schemas in different formats (text vs base64-encoded binary).

Fixes #6831

This PR addresses an issue where .NET and other clients submitting Protobuf schemas as base64-encoded `FileDescriptorProto` would:
1. Fail compatibility rule validation with `Syntax error in :1:1: unexpected label: CiV...`
2. Create duplicate versions instead of recognizing existing schemas
3. Experience lookup failures due to content hash mismatches

## Root Cause
The registry was storing Protobuf schemas in whatever format they were submitted - either as `.proto` text format or as base64-encoded `FileDescriptorProto` binary format. This caused multiple problems:

1. **Rule validation failures**: Compatibility and validity rules expect `.proto` text format, but when .NET clients submit base64-encoded `FileDescriptorProto`, the rule parsers fail with syntax errors (issue #6831).

2. **Content hash mismatches**: When a schema was stored in one format (e.g., text) but looked up in another format (e.g., base64), the content hash wouldn't match, causing legitimate schema lookups to fail with `ArtifactNotFoundException`.

3. **Duplicate versions created**: The `ifExists=RETURN_OR_UPDATE` behavior would create duplicate versions instead of finding the existing one, because the content hash of text vs base64 format differs even when the schemas are semantically identical.

4. **Migration issues**: Existing registries that stored Protobuf schemas in base64 format (from .NET clients) would have lookup failures after clients started normalizing to text format.

The fundamental issue was inconsistent storage format - the registry should treat Protobuf schemas canonically regardless of submission format.

## Changes
- **`app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/AbstractResource.java`**:
  - Added Protobuf normalization logic in `subjectsPost()` method (lines 131-163)
  - Normalizes incoming Protobuf schemas to text format before storage and rule validation
  - Added fallback lookup in `subjectsVersionsPost()` (lines 198-280) for backward compatibility with base64-stored schemas
  - Uses canonical comparison via `ProtoFileElement.toSchema()` when content hash lookup fails

- **`app/src/main/java/io/apicurio/registry/rest/v3/GroupsResourceImpl.java`**:
  - Added `normalizeProtobufContent()` helper method (lines 1461-1492) to centralize normalization logic
  - Applied normalization in `createArtifact()` (line 976) before storage
  - Applied normalization in `createArtifactVersion()` (line 1098) before storage and rules
  - Applied normalization in `handleIfExistsReturnOrUpdate()` (line 1347) to fix content hash lookups
  - Added canonical fallback logic (lines 1364-1406) for migration scenarios where existing schemas are in base64 format
  - Applied normalization in `updateArtifactInternal()` (line 1428) for consistency

- **`app/src/test/java/io/apicurio/registry/ConfluentDataMigrationTest.java`**:
  - Added `testProtobufMigrationCompatibility()` test to verify migrated Protobuf data is readable
  - Added `testProtobufNoDuplicatesTextAndBase64()` comprehensive test demonstrating the fix
  - Verifies that submitting same schema in text and base64 formats doesn't create duplicates
  - Tests both Apicurio API v3 and Confluent Schema Registry API compatibility
  - Updated expected artifact count from 14 to 15 to reflect new test data

- **`integration-tests/src/test/java/io/apicurio/registry/noprofile/ccompat/rest/v7/ConfluentClientTest.java`**:
  - Added comprehensive `testProtobufNormalizationTextAndBase64()` integration test
  - Tests complete workflow: create with text, lookup with base64, verify no duplicates
  - Includes compatibility rule validation with both formats (specifically tests the #6831 scenario)
  - Verifies the normalization works end-to-end in real scenarios

## Test plan
- [x] Added `ConfluentDataMigrationTest.testProtobufMigrationCompatibility()` to verify migrated Protobuf schemas are accessible
- [x] Added `ConfluentDataMigrationTest.testProtobufNoDuplicatesTextAndBase64()` to verify text and base64 submissions don't create duplicates
- [x] Added `ConfluentClientTest.testProtobufNormalizationTextAndBase64()` comprehensive integration test covering:
  - Creating artifact with text format
  - Looking up with base64 format (verifies canonical matching)
  - Compatibility rule validation with both formats (tests #6831 fix)
  - Idempotent behavior verification
- [x] Manual testing with .NET client scenarios (base64 submission)
- [x] Run full integration test suite with SQL storage: `./mvnw verify -am -Pintegration-tests -pl integration-tests -Plocal-tests -Psmoke`
- [x] Run full integration test suite with KafkaSQL storage: `./mvnw verify -am -Pintegration-tests -pl integration-tests -Plocal-tests -Psmoke -Premote-kafka`
- [ ] Verify backward compatibility with existing base64-stored schemas in production environments

### DIFF
--- a/app/src/test/java/io/apicurio/registry/ConfluentDataMigrationTest.java
+++ b/app/src/test/java/io/apicurio/registry/ConfluentDataMigrationTest.java
@@ -6,6 +6,7 @@ import io.apicurio.registry.rest.client.models.ArtifactSearchResults;
 import io.apicurio.registry.rest.client.models.RuleType;
 import io.apicurio.registry.storage.RegistryStorage;
 import io.apicurio.registry.types.Current;
+import io.apicurio.registry.utils.protobuf.schema.FileDescriptorUtils;
 import io.quarkus.test.junit.QuarkusTest;
 import jakarta.inject.Inject;
 import org.junit.jupiter.api.Assertions;
@@ -14,8 +15,10 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.io.InputStream;
+import java.util.Base64;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 @QuarkusTest
@@ -60,8 +63,8 @@ public class ConfluentDataMigrationTest extends AbstractResourceTestBase {
     @Test
     public void testCheckArtifacts() {
         ArtifactSearchResults results = clientV3.search().artifacts().get();
-        Assertions.assertEquals(14, results.getArtifacts().size());
-        Assertions.assertEquals(14, results.getCount());
+        Assertions.assertEquals(15, results.getArtifacts().size());
+        Assertions.assertEquals(15, results.getCount());
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


EricWittmann:
@carlesarnal It seems to me we might want to require an alternative content-type for the protobuf schema when it's in binary/b64 format?

carlesarnal:
> @carlesarnal It seems to me we might want to require an alternative content-type for the protobuf schema when it's in binary/b64 format?

Why do you think so? With the approach in this PR I think it may be enough and, even though I know it's kind of hacky, for proto schemas created from the ccompat API in existing deployments we'll need a transformation mechanism anyway.

EricWittmann:
I think the v3 Core API should require users to be correct about the type of content they are adding.  Since the API now has a content-type field that is **required** - if the user wants to pass binary protobuf (base64 encoded) content that's fine, but they should tell us that's what they are doing.  I'm cool with everything else happening here (convert from binary to text prior to storage).  I'm just saying we should only try to do the "normalize to text" conversion if they tell us that the content-type is "binary protobuf" (I haven't thought about what the content-type should actually be for it).

carlesarnal:
> I think the v3 Core API should require users to be correct about the type of content they are adding. Since the API now has a content-type field that is **required** - if the user wants to pass binary protobuf (base64 encoded) content that's fine, but they should tell us that's what they are doing. I'm cool with everything else happening here (convert from binary to text prior to storage). I'm just saying we should only try to do the "normalize to text" conversion if they tell us that the content-type is "binary protobuf" (I haven't thought about what the content-type should actually be for it).

Understood. I think this is superseded by https://github.com/Apicurio/apicurio-registry/pull/6926. Once we decide the approach, we might use the new capabilities to handle this situation.

PR #6950, opened 2025-12-02
Implement references graph feature

carlesarnal:
Disclaimer: Implemented with Claude Code.

@EricWittmann I'm tagging you so you can take a quick look in case there anything extremely wrong here. I've done my part of reviewing the implementation but this is absolutely outside my area of expertise. So I don't want to merge it without you taking a quick look at it.


EricWittmann:
Looks OK to me.  I can't vouch for the hook, because the logic is more than I can consume right now.  :)  But the component organization and CSS and all that looks fine to me.  :+1:

PR #7017, opened 2025-12-11
Implement authentication for the mcp server

carlesarnal:
Implement support for oauth, tls and mtls.


carlesarnal:
> I know this is still a draft, and maybe it's out of scope for this change, but I guess we also need to think about TLS config options eh?

Absolutely, this is just the beginning, I want to refactor a few things, too.

jsenko (approved):

PR #7106, opened 2026-01-12
Fallback Id handler to handle migration from 8-byte ids to 4-byte schema ids

ozangunalp:
During the testing of https://github.com/quarkusio/quarkus/pull/51401, we realised that the migration to v3 involves handling the change of schema ids from using 8-bytes to 4-bytes. 

While there are distinct IdHandler implementations available to the client application to use, there isn't any easy migration path to use with message processing applications producing/consuming on topics (eg. Kafka).

The problem of an active topic is explained here : https://github.com/quarkusio/quarkus/pull/51401#issuecomment-3718501645

This change proposes an `OptimisticFallbackIdHandler` to write 4-byte ids and read both 4 and 8 byte ids to be used in a migration scenario. 

Because the payload format doesn't include any format versionning, this IdHandler makes the following assumption to distinguish between 4-bytes 8-bytes ids: The schema ids are greater than 0 and smaller than integer max value.

The change requires some non-breaking modifications to `AbstractDeserializer`, and maybe to the `IdHandler` interface that can be discussed.

Additionally, I couldn't figure out what to do for the Go SDK.

### DIFF
--- a/serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/AbstractDeserializer.java
+++ b/serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/AbstractDeserializer.java
@@ -94,7 +94,8 @@ public abstract class AbstractDeserializer<T, U> implements AutoCloseable {
 
         SchemaLookupResult<T> schema = resolve(topic, data, artifactReference);
 
-        int length = buffer.limit() - 1 - baseSerde.getIdHandler().idSize();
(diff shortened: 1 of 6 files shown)


carlesarnal:
Thanks for the PR, as we've discussed, this addresses a real migration pain point for v2 → v3 upgrades. I think the approach makes sense, but it's probably worth documenting the assumptions as javadoc in the OptimisticFallbackIdHandler class.

Issue #7333, opened 2026-02-10
[AVRO] Registry does not support uploading schemas with record defined internally.


wagonli:
## Description

**Registry
Version**: 3.1.7
**Persistence type**: sql

We're using AVRO IDL to define our avro object, in order to be able to host the schemas on apicurio, we use the provided tools from the avro package to generate a JSON schema.
The issue we're facing is that if our schema define a record for a first field and then use a reference to this record type definition on a second field,  Apicurio refuses to validate the schemas saying that we're missing external dependencies.

For example the following IDL :

```avdl
namespace avro.events.tests;

schema TestEvent;

record TestEvent {
	MonetaryAmount retailPrice;
	MonetaryAmount wholesalePrice;
}

record MonetaryAmount {
	decimal(9,2) amount;
	string currency;
}
```

Translates to this schema using Avro's `IdlReader` for parsing and `IdlFile.getMainSchema()` for the avsc schema generation.
See `org.apache.avro.tool.IdlToSchemataTool` for [reference](https://github.com/apache/avro/blob/main/lang/java/tools/src/main/java/org/apache/avro/tool/IdlToSchemataTool.java).

```avsc
{
  "type" : "record",
  "name" : "TestEvent",
  "namespace" : "avro.events.tests",
  "fields" : [ 
  {
    "name" : "retailPrice",
    "type" : {
      "type" : "record",
      "name" : "MonetaryAmount",
      "fields" : [ {
        "name" : "amount",
        "type" : {
          "type" : "bytes",
          "logicalType" : "decimal",
          "precision" : 9,
          "scale" : 2
        }
      }, 
    {
        "name" : "currency",
        "type" : "string"
      } ]
    }
  }, 
  {
    "name" : "wholesalePrice",
    "type" : "MonetaryAmount"
  } ]
}
```

When trying to upload the artifact to apicurio, we're getting the following error : 

```
Missing reference detected in Avro artifact. Causes: Missing reference detected. at Undefined schema: avro.events.tests.MonetaryAmount
```

### Expected vs Actual Behaviour

Seems like the AVRO reference finder is missing a concept of internal type and thus making the artifact creation fail where it should not as all types are correctly defined.

PR #7334, opened 2026-02-10
[AVRO] Fix missing reference issue for internally defined types during schema upload

wagonli:
resolves https://github.com/Apicurio/apicurio-registry/issues/7333
 
 Add the concept of internal types to the `AvroReferenceFinder` and remove any type defined internally from the external types HashSet.

### DIFF
--- a/schema-util/avro/src/main/java/io/apicurio/registry/avro/content/refs/AvroReferenceFinder.java
+++ b/schema-util/avro/src/main/java/io/apicurio/registry/avro/content/refs/AvroReferenceFinder.java
@@ -5,8 +5,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import io.apicurio.registry.content.TypedContent;
 import io.apicurio.registry.content.refs.ExternalReference;
-import io.apicurio.registry.content.refs.ReferenceFinderException;
 import io.apicurio.registry.content.refs.ReferenceFinder;
+import io.apicurio.registry.content.refs.ReferenceFinderException;
 
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


carlesarnal:
Thanks for this contribution! The approach is sound and fixes the reported issue correctly.

However, the same issue exists for **enum** and **fixed** types that can also be defined inline within a schema. Currently, the `case "enum"` block just returns without registering the type:

```java
case "enum":
{
    // Do nothing for enums
    return;
}
```

For completeness, inline-defined enums and fixed types should also be tracked as internal types. Here's the suggested change:

```java
case "enum":
{
    String enumName = schema.get("name").asText();
    String qualifiedType = qualifyTypeName(enumName, namespace);
    internalTypes.add(qualifiedType);
    externalTypes.remove(qualifiedType);
    return;
}
case "fixed":
{
    String fixedName = schema.get("name").asText();
    String qualifiedType = qualifyTypeName(fixedName, namespace);
    internalTypes.add(qualifiedType);
    externalTypes.remove(qualifiedType);
    return;
}
```

**Example scenario that would fail without this:**
```json
{
  "type": "record",
  "name": "Order",
  "namespace": "com.example",
  "fields": [
    { "name": "status", "type": { "type": "enum", "name": "Status", "symbols": ["PENDING", "COMPLETED"] } },
    { "name": "previousStatus", "type": "Status" }
  ]
}
```

Also, minor style note: there's a missing space after `if` on line 94:
```java
if(!internalTypes.contains(qualifiedType)) {
```
Should be:
```java
if (!internalTypes.contains(qualifiedType)) {
```

Would you be able to add handling for enum and fixed types as well?

wagonli:
@carlesarnal I added your changes. I enhanced the test reference avsc to take enum and fixed into account.

I also took the liberty to move from the old switch syntax to the new one but I can reverse this if it's not ok with apicurio's code styles.

carlesarnal:
> @carlesarnal I added your changes. I enhanced the test reference avsc to take enum and fixed into account.
> 
> I also took the liberty to move from the old switch syntax to the new one but I can reverse this if it's not ok with apicurio's code styles.

It its, it was just a nitpick about the whitespace :)

PR #7496, opened 2026-03-05
docs: add schema compatibility modes documentation

eameh:


EricWittmann (dismissed):
lgtm

carlesarnal:
@eameh is this still a draft?
Yes, I am just waiting for this PR to be peer-reviewed

eameh:
Hi @pwright, Gentle reminder about peer review for this issue

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
```diff
@@ -0,0 +1,429 @@
+include::{mod-loc}shared/all-attributes.adoc[]
+
+:_mod-docs-content-type: ASSEMBLY
+[id="registry-compatibility-modes_{context}"]
+= {registry} schema compatibility modes
+
+[role="_abstract"]
+You can configure schema compatibility modes in {registry} to control what schema changes are allowed or disallowed when you add a new artifact version. Because compatibility rules are format-specific, each supported artifact type has its own set of compatible and incompatible changes.
```
pwright, line 8:
```suggestion
In the {registry}, compatibility modes determine which changes are permitted when you upload a new version of an artifact. Because these rules vary depending on the data format (such as Avro, JSON, or Protobuf), each artifact type has its own specific criteria for what constitutes a compatible or incompatible change.
```

pwright: not sure if above is true, needs tech review

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
```diff
@@ -0,0 +1,429 @@
...
+// ref-registry-compatibility-overview
+:_mod-docs-content-type: REFERENCE
+[id="registry-compatibility-overview_{context}"]
+== Overview of compatibility rules
+
+[role="_abstract"]
+When you add a new version of an artifact to {registry}, you can configure the `COMPATIBILITY` rule to check whether the new content is compatible with existing versions.
+Compatibility checking helps you ensure that producers and consumers can continue to work together without disruption when schemas evolve.
+
+You can configure the `COMPATIBILITY` rule at three levels:
```
pwright, line 31: I found this section a bit confusing how about something like:

You can configure the `COMPATIBILITY` rule at three levels. {registry} applies these rules in order of precedence, where the most specific level always overrides the broader levels:

1. **Artifact-specific rules**: Apply to a single artifact and have the highest priority.
2. **Group-specific rules**: Apply to all artifacts in a specific group.
3. **Global rules**: Apply to all artifacts across the registry unless a group or artifact rule is set.

NOTE:  To disable a rule inherited from a higher level, you must explicitly set the rule at the lower level to `NONE`.

pwright (commented):
Not finished, but have to jump to other task

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
eameh: cc @EricWittmann, what are your thoughts technically about this suggestion?

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
```diff
@@ -0,0 +1,429 @@
...
+* {registry-rules}
+
+
+// ref-registry-compatibility-mode-descriptions
+:_mod-docs-content-type: REFERENCE
+[id="registry-compatibility-mode-descriptions_{context}"]
+== Compatibility mode descriptions
+
+[role="_abstract"]
+You can configure one of seven compatibility modes for the `COMPATIBILITY` rule type.
```
pwright: one of the following compatibility modes
```suggestion
Compatibility modes act as a set of rules that determine whether a new schema version is allowed based on its ability to work with existing data and applications.
```

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
```diff
@@ -0,0 +1,429 @@
...
+// ref-registry-compatibility-avro
+:_mod-docs-content-type: REFERENCE
+[id="registry-compatibility-avro_{context}"]
+== Avro schema compatibility
+
+[role="_abstract"]
+{registry} uses the Apache Avro library to check compatibility for Avro schemas.
+The tables in each section describe common Avro schema changes and whether they maintain or break backward and forward compatibility.
+
+.Backward compatible changes for Avro
```
pwright: would this section be better as a single table?


Change Type | Backward Compatible? | Forward Compatible? | Reason
-- | -- | -- | --
Add field with default | ✅ | ✅ | New consumers use the default for old data; old consumers ignore the new field.
Add field without default | ❌ | ✅ | New consumers have no value for old data; old consumers ignore the new field.
Remove field with default | ✅ | ✅ | New consumers ignore the field; old consumers use their own default for new data.
Remove field without default | ✅ | ❌ | New consumers ignore the field; old consumers have no value to fill the gap.
Add type to a union | ✅ | ❌ | New consumers can handle the new type; old consumers will fail if they encounter it.
Remove type from a union | ❌ | ✅ | New consumers can't parse the old type; old consumers only expect types they already know.
Rename field (no alias) | ❌ | ❌ | The field name mapping is broken for both parties.
Change field type | ❌ | ❌ | Data types are fundamentally incompatible (e.g., string to int).

eameh: Updated

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
eameh: Updated

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
eameh: Updated

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
eameh: Updated, but leaving thread open pending @EricWittmann confirmation

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
```diff
@@ -0,0 +1,415 @@
...
+[role="_abstract"]
+{registry} uses a custom library to analyze differences between JSON Schema versions.
+Because JSON Schema uses a constraint-based validation model, a backward compatible change typically makes the new schema more permissive so that it accepts all data that the previous schema accepted.
+
+.Backward compatible changes for JSON Schema
+
+You can make these changes to a JSON Schema and maintain backward compatibility.
+The new schema continues to validate all data that the previous schema validated.
+
+.JSON Schema backward compatible changes
```
pwright: should this and other types have same format table as avro?

docs/modules/ROOT/pages/getting-started/assembly-registry-compatibility-modes.adoc
```diff
@@ -171,7 +170,7 @@ As a result, Version 3 passes the `BACKWARD` check but fails the `BACKWARD_TRANS
 
 [role="_abstract"]
 {registry} uses a custom library to analyze differences between JSON Schema versions.
-Because JSON Schema uses a constraint-based validation model, the following table describes common changes and whether they maintain or break backward compatibility.
+Because JSON Schema uses a constraint-based validation model, the following table describes common changes and whether they maintain or break backward compatibility. For JSON Schema, forward compatibility is the inverse of backward compatibility: a change that makes the schema more permissive is backward compatible but forward incompatible, and a change that makes the schema more restrictive is forward compatible but backward incompatible.
```
pwright, line 173: I was wondering about that, nice

pwright (dismissed):
some final questions
Do we need to say something like 'you always start with empty artifact' and that is excluded from compatibility rules?
Should we avoid 'latest version', cos latest is now technically a branch?
Do we need to call out that this is /v3 specific?

PR #7516, opened 2026-03-09
Update base image to ubi10 to address misc. CVEs

frankgrimes97:


carlesarnal:
As per the discussion in the associated issue, we can't change the base image used for the project.

PR #3646, opened 2023-09-12
Change content canonicalizer for AVRO to improve schema evolution and…

AndreaScarselli:
Implements a new AVRO content canonicalizer that addresses the issues reported in #300 .

In particular, the new Content Canonicalizer:

1. Consider different 2 schemas that have the same fields in different order
2. Considers equal 2 schemas that have the same attribute in the root in a different order
3. Considers equals a schema that has the namespace in the `name`, and a schema that has the same `name` and the namespace in the `namespace` field
4. Considers equals 2 schemas that only differ by the optional attributes (e.g`doc`) both in the root and in the fields

PR #7550, opened 2026-03-13
fix(avro): normalize implicit null default in nullable unions

AndreaScarselliAx:
## Summary

  Fix a canonical-hash equivalence bug in Apicurio Registry's Avro content canonicalizer, and add a database upgrader to recompute stale canonical hashes for existing data.

  ## Problem

  With [this](https://github.com/Apicurio/apicurio-registry/pull/3646) contribution, I added a custom Avro canonicalizer. Later, while running v2, we found a bug: a ccompat lookup between the following two schemas returned a 40403.

  ```json
  {
      "type": "record",
      "name": "Application",
      "namespace": "nl.andrea.general",
      "doc": "Identification of an application",
      "fields": [
          {
              "name": "name",
              "type": "string",
              "doc": "The name of the application"
          },
          {
              "name": "version",
              "type": ["null", "string"],
              "doc": "(Optional) The application version"
          }
      ]
  }
  ```

  ```json

  {
      "type": "record",
      "name": "Application",
      "namespace": "nl.andrea.general",
      "doc": "Identification of an application",
      "fields": [
          {
              "name": "name",
              "type": "string",
              "doc": "The name of the application"
          },
          {
              "name": "version",
              "type": ["null", "string"],
              "doc": "(Optional) The application version",
              "default": null
          }
      ]
  }
  ```

  These two schemas are semantically equivalent — `"default": null`. The custom canonicalizer was incorrectly producing different hashes for them. This OR fixes the nullable-union/"default": null equivalence case in the existing canonicalizer.
  ## Changes

  - `EnhancedAvroContentCanonicalizer` — In normalizeField, when a field has no explicit default and its type is a union whose first member is null, inject JsonProperties.NULL_VALUE as the default.
  Both forms (implicit and explicit "default": null) now produce the same canonical output. All custom properties, non-null defaults, and other behavior are preserved.
  - `AvroCanonicalHashUpgrader` — new DB upgrader (version 104) that recomputes canonical hashes for all existing Avro content. Without this, schemas registered before the fix would retain stale
  hashes and still fail ccompat lookups against newly registered equivalent schemas.
  - Tests — `SchemaNormalizerTest` adds the equivalence test for the reported bug, plus edge-case coverage (union member order, nested records, multi-member unions, and a guard that the fix does not
  over-trigger when null is not the first union member). AvroCanonicalHashUpgraderTest covers the upgrader: recompute on Avro, no-op on non-Avro, no-op on already-correct hashes, and artifact-type
  vs content-type correctness.


app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/AvroCanonicalHashUpgrader.java
```diff
@@ -0,0 +1,117 @@
...
+                + "WHERE v.groupId = ? AND v.artifactId = ? AND v.version = ?";
+
+        ContentEntity entity = handle.createQuery(sql)
+                .bind(0, normalizeGroupId(reference.getGroupId()))
+                .bind(1, reference.getArtifactId()).bind(2, reference.getVersion())
+                .map(ContentEntityMapper.instance).one();
+
+        return ContentWrapperDto.builder().content(ContentHandle.create(entity.contentBytes))
+                .contentType(entity.contentType)
+                .references(RegistryContentUtils.deserializeReferences(entity.serializedReferences))
```
carlesarnal: Same here regarding contentType/artifactType.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/AvroCanonicalHashUpgrader.java
```diff
@@ -0,0 +1,117 @@
...
+ * canonicalizer was changed to use Apache Avro's standard {@code SchemaNormalization.toParsingForm()}.
+ */
+@RegisterForReflection
+public class AvroCanonicalHashUpgrader implements IDbUpgrader {
+
+    private static final Logger log = LoggerFactory.getLogger(AvroCanonicalHashUpgrader.class);
+
+    private final ArtifactTypeUtilProviderFactory factory = new DefaultArtifactTypeUtilProviderImpl(true);
+
+    @Override
```
carlesarnal: @EricWittmann what do you think about transactionality here? The db version has been already bumped, so, if anything fails mid process, some rows will be upgraded and some don't. This should be shielded behind a transaction, no?

carlesarnal (changes requested):
Overall looks good

AndreaScarselliAx:
> @AndreaScarselliAx I'm a bit surprised by no test failing due to the error mentioned in the review. Would you be able to add a new test to cover that case? If you need some assistance, just let me know!

Hi @carlesarnal , sorry for the long delay on this one. It should be done. I also solved the new conflicts from main.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/AvroCanonicalHashUpgrader.java
carlesarnal: Yes, of course, all the commands here run inside the same transaction. Is the db upgrade inside it too? That's the part I forgot.

carlesarnal:
@AndreaScarselliAx no worries, please, rebase main, I just fixed an issue with CI.

carlesarnal:
Hi @AndreaScarselliAx, thanks for the work on this!

While resolving the merge conflict with `main`, we noticed a behavioral issue with switching to `SchemaNormalization.toParsingForm()`.

The `main` branch recently added tests (for #7788) that assert schemas differing only in custom properties like `connect.parameters`, field-level custom properties, or default values must produce **different** canonical hashes. This is important for Debezium-style schemas where `connect.parameters` carries semantic information (e.g., `"allowed": "station,post_office"` vs `"allowed": "station,post_office,plane"`).

However, `SchemaNormalization.toParsingForm()` follows the Avro spec canonical form which **strips all custom properties and defaults**. This means:

- Two schemas differing only in `connect.parameters` values → same canonical hash (wrong)
- Two schemas differing only in default values (`"default": 0` vs `"default": 1`) → same canonical hash (wrong)
- Two schemas differing only in field-level custom properties → same canonical hash (wrong)

So while this PR correctly fixes the nullable union + `"default": null` equivalence bug, it would introduce a regression for the Debezium/connect.parameters scenarios.

The conflicting tests are in `SchemaNormalizerTest.java` — specifically `parseSchema_SchemasWithDifferentConnectParameters_NotEqual`, `parseSchema_SchemasWithDifferentFieldLevelCustomProperties_NotEqual`, `parseSchema_SchemasWithDifferentDefaultValues_NotEqual`, `parseSchema_DebeziumNullableEnumWithDifferentAllowedValues_NotEqual`, and others.

A possible approach would be to fix the specific nullable-union-default bug in the existing `EnhancedAvroContentCanonicalizer` rather than replacing it entirely with `toParsingForm()`, or to use a hybrid approach that uses `toParsingForm()` as a base but re-injects custom properties and defaults into the canonical form.

AndreaScarselliAx:
> Hi @AndreaScarselliAx, thanks for the work on this!
> 
> While resolving the merge conflict with `main`, we noticed a behavioral issue with switching to `SchemaNormalization.toParsingForm()`.
> 
> The `main` branch recently added tests (for #7788) that assert schemas differing only in custom properties like `connect.parameters`, field-level custom properties, or default values must produce **different** canonical hashes. This is important for Debezium-style schemas where `connect.parameters` carries semantic information (e.g., `"allowed": "station,post_office"` vs `"allowed": "station,post_office,plane"`).
> 
> However, `SchemaNormalization.toParsingForm()` follows the Avro spec canonical form which **strips all custom properties and defaults**. This means:
> 
> * Two schemas differing only in `connect.parameters` values → same canonical hash (wrong)
> * Two schemas differing only in default values (`"default": 0` vs `"default": 1`) → same canonical hash (wrong)
> * Two schemas differing only in field-level custom properties → same canonical hash (wrong)
> 
> So while this PR correctly fixes the nullable union + `"default": null` equivalence bug, it would introduce a regression for the Debezium/connect.parameters scenarios.
> 
> The conflicting tests are in `SchemaNormalizerTest.java` — specifically `parseSchema_SchemasWithDifferentConnectParameters_NotEqual`, `parseSchema_SchemasWithDifferentFieldLevelCustomProperties_NotEqual`, `parseSchema_SchemasWithDifferentDefaultValues_NotEqual`, `parseSchema_DebeziumNullableEnumWithDifferentAllowedValues_NotEqual`, and others.
> 
> A possible approach would be to fix the specific nullable-union-default bug in the existing `EnhancedAvroContentCanonicalizer` rather than replacing it entirely with `toParsingForm()`, or to use a hybrid approach that uses `toParsingForm()` as a base but re-injects custom properties and defaults into the canonical form.

Hi @carlesarnal ,

Thanks for the feedback, I force pushed against this branch, the new code just addresses the nullable-union/default bug. I have added some extra tests in `SchemaNormalizerTest`

carlesarnal:
Merged! Thanks for your patience.

AndreaScarselliAx:
> Merged! Thanks for your patience.

Thank you @carlesarnal! Do you have any idea when this could be released?

carlesarnal:
> > Merged! Thanks for your patience.
> 
> Thank you @carlesarnal! Do you have any idea when this could be released?

I'm planning a 3.3.0 release next week. I'll announce an improvement in our release processed too :).

AndreaScarselliAx:
Thanks for the prompt reply @carlesarnal !

AndreaScarselliAx:
Hey @carlesarnal , sorry to bother you again with this. Any news on the `3.3.0` release?

PR #7568, opened 2026-03-17
Preserve editor changes across re-authentication and trigger redirect on 401

boris-dibrov:
Links:
- Issue: https://github.com/Apicurio/apicurio-registry/issues/7569
- Common UI Components MR: https://github.com/Apicurio/apicurio-common-ui-components/pull/264
- Registry MR: https://github.com/Apicurio/apicurio-registry/pull/7568

## Summary

This MR adds Registry-side UX on top of the shared OIDC auth fix.

For the editor page, unsaved changes are saved locally before redirecting the user back to the OIDC provider and are restored after the user returns.

For the other pages, the UI does not try to preserve page-local state. When the app receives `401 Unauthorized`, it redirects the user to the provider instead of remaining in a broken state.

The Common UI Components part fixes the mixed OIDC lifecycle during normal bootstrap, callback handling, and failed refresh cleanup. This MR relies on that corrected auth behavior and adds the Registry-specific recovery UX.

## What was changed

- added a re-authentication coordination service for Registry UI;
- updated `PageErrorHandler` so a re-authentication flow can take over the UX instead of immediately replacing the page with a generic error state;
- added editor-specific local snapshot persistence and restore flow;
- added recovery and re-authentication modals for the editor;
- cleaned up editor snapshots when they are no longer needed;
- handled browser storage quota overflow explicitly for very large specs;
- kept the unsaved-changes browser prompt behavior intact for dirty editor tabs.

## Scope

Editor page behavior:

- if the session expires while the user is editing, current editor changes are stored locally before redirect;
- after successful re-authentication, the editor offers to restore those changes;
- if the local snapshot cannot be stored because browser storage is full, the user is warned before proceeding.

Non-editor pages:

- on `401`, redirect to the provider;
- no local page-state recovery is attempted.

## Dependency

This MR depends on the Common UI Components MR.

Reason:

- the library MR separates normal bootstrap from callback handling;
- it prevents false callback processing on regular reload;
- it clears broken auth state after failed refresh so a clean redirect-based login can start again.

After the library MR is merged and published, update the dependency in:

- `ui/ui-app/package.json`
- `ui/ui-app/package-lock.json`

Dependency to update:
- `@apicurio/common-ui-components`

## Validation

Tested locally with:

- OIDC scenarios, including failed refresh and re-authentication recovery
- editor snapshot save / restore flow
- basic auth
- no auth
- very large spec that does not fit into local storage

Result:

- all tested OIDC scenarios worked;
- basic auth worked;
- no-auth mode worked;
- editor recovery worked;
- large local snapshot overflow was handled explicitly;
- no unrelated regressions were observed in the tested flows.

### DIFF
--- a/ui/ui-app/src/app/pages/editor/EditorPage.tsx
+++ b/ui/ui-app/src/app/pages/editor/EditorPage.tsx
@@ -21,7 +21,7 @@ import {
 } from "@app/pages";
 import { RootPageHeader } from "@app/components";
 import { ContentTypes } from "@models/ContentTypes.ts";
-import { AuthService, PleaseWaitModal, useAuth } from "@apicurio/common-ui-components";
+import { PleaseWaitModal } from "@apicurio/common-ui-components";
 import { Draft, DraftContent } from "@models/drafts";
 import { useDraftsService } from "@services/useDraftsService.ts";
 import { useDownloadService } from "@services/useDownloadService.ts";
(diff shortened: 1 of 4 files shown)


EricWittmann:
## 1. Upgrade apicurio-common-ui-components
I've merged your changes to `apicurio-common-ui-components` - you can upgrade the version of it in this PR.  The new version is `4.0.1`

## 2. EditorPage increased complexity
The EditorPage component was already a bit large, and this PR adds to that (6 new useState, 3 new useRef, 1 new useEffect for snapshot context sync).

**Suggestion**: Extract the snapshot/recovery logic into a custom hook, e.g. `useEditorDraftRecovery(...)`, and the re-authentication coordination into `useEditorReauthentication(...)`. This would keep EditorPage focused on rendering and orchestration.

wdyt?

PR #7602, opened 2026-03-22
fix: catch SchemaException in ProtobufSchemaParser.parseSchema()

krickert:
## Summary

`ProtobufSchemaParser.parseSchema()` throws an uncaught `RuntimeException` (wire-schema's `SchemaException`) when parsing `.proto` text that references imports which can't be resolved locally (e.g. custom proto files from other modules not present in `resolvedReferences`).

The existing catch blocks handle `DescriptorValidationException` and `IllegalStateException` but not `SchemaException`, so the exception propagates to the Kafka consumer and causes deserialization to fail permanently — even though the schema content may be parseable as a binary `FileDescriptorProto`.

## Root Cause

`ProtoParser.Companion.parse()` (from wire-schema) throws `SchemaException: unable to find <path>.proto` when the `.proto` text contains `import` statements for files that wire-schema can't locate. This commonly happens when:
- A message references transitive proto imports from other modules
- The schema registry has the schema text but the consumer's `resolvedReferences` map is incomplete
- Custom proto files (not well-known types) are imported

## Fix

Added a `catch (RuntimeException)` block that checks if the exception is a schema resolution failure (by class name `SchemaException` or by message pattern `"unable to find *.proto"`), and falls back to `parseDescriptor()` which can handle binary `FileDescriptorProto` format without import resolution.

This is consistent with the existing fallback pattern already used for `IllegalStateException` at line 76-78.

## Tests

- `testParseSchemaWithUnresolvableCustomImportFallsBackToDescriptor` — binary descriptor with unresolvable import falls back successfully
- `testParseSchemaWithUnresolvableImportTextOnlyThrowsCleanError` — text-only with unresolvable import produces clean error from `parseDescriptor` (not raw `SchemaException`)


carlesarnal (changes requested):
The tests seem good, but I think we should look for a better approach for checking the exception.

carlesarnal:
@krickert please, rebase main, I just fixed an issue with CI.

krickert:
Rebased...

carlesarnal (approved):

PR #7603, opened 2026-03-22
feat: add opt-in fallback to direct protobuf parsing when schema resolution fails

krickert:
## Summary

Adds an opt-in `apicurio.protobuf.fallback.on-schema-error` configuration to `ProtobufDeserializer` that enables fallback to direct protobuf parsing when schema resolution fails and a `specificReturnClass` is configured.

## Problem

When the Apicurio schema registry is unreachable, slow, or missing transitive proto imports, the `ProtobufDeserializer` fails permanently — even though the message bytes are valid protobuf and the consumer knows the exact target type via `specificReturnClass`. This causes consumer group stalls in production during:

- Registry downtime or network partitions
- Missing transitive proto imports (e.g. `pipeline_core_types.proto` not registered)
- Rolling deployments where consumers start before producers register schemas
- Content ID propagation delays across registry replicas

## Solution

When `apicurio.protobuf.fallback.on-schema-error=true` is configured:

1. The deserializer attempts normal schema-resolved deserialization first
2. If schema resolution throws any `RuntimeException`, it falls back to stripping the Apicurio wire-format prefix and parsing directly with `specificReturnClass.parseFrom(InputStream)`
3. The wire-format prefix handling respects the configured `IdHandler.idSize()`, `readIndexes`, and `readTypeRef` flags

The fallback is:
- **Opt-in** (defaults to `false`) — no behavioral change for existing users
- **Only available** when `specificReturnClass` is set to a concrete protobuf class (not `DynamicMessage`, which requires a `Descriptor`)
- **Logged at WARN level** when activated, including the root cause and the fallback class name

## Configuration

```properties
apicurio.protobuf.fallback.on-schema-error=true
apicurio.registry.deserializer.value.return-class=com.example.MyMessage
```

## Changes

| File | Change |
|------|--------|
| `ProtobufDeserializer.java` | Override `deserializeData()` with fallback catch; add `parseFallback()` that strips wire-format prefix using configured `IdHandler`, `readIndexes`, `readTypeRef` |
| `ProtobufDeserializerConfig.java` | Add `FALLBACK_ON_SCHEMA_ERROR` config constant and `fallbackOnSchemaError()` accessor |
| `ProtobufDeserializerFallbackTest.java` | 8 tests: config defaults, Boolean/String values, wire-format parsing with Ref, empty Ref, null data, negative test proving fallback doesn't activate when disabled |

## Tests

All 16 tests pass (8 existing `ProtobufSchemaParserTest` + 8 new `ProtobufDeserializerFallbackTest`):

```
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0 -- ProtobufDeserializerFallbackTest
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0 -- ProtobufSchemaParserTest
```

Wire-format tests use `DescriptorProtos.FileDescriptorProto` (always on classpath) with a `ThrowingSchemaResolver` mock that forces the fallback path. Tests verify correct field values after round-trip through the Apicurio wire format.

## Related

- #7602 — catches `SchemaException` in `ProtobufSchemaParser.parseSchema()` (complementary fix at the schema parsing layer)


carlesarnal (changes requested):
Overall looks good to me, just a couple comments

serdes/generic/serde-common-protobuf/src/test/java/io/apicurio/registry/serde/protobuf/ProtobufDeserializerFallbackTest.java
```diff
@@ -0,0 +1,305 @@
...
+        return out.toByteArray();
+    }
+
+    /**
+     * A {@link io.apicurio.registry.resolver.SchemaResolver} that always throws on resolution,
+     * forcing the deserializer into the fallback path. All non-resolution methods are no-ops.
+     *
+     * @param <T> the protobuf message type
+     */
+    private static class ThrowingSchemaResolver<T extends com.google.protobuf.Message>
```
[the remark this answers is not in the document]
krickert: No problem!  I'll adjust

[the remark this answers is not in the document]
krickert: Fixed in `e6bb2730a`. Reworked `ThrowingSchemaResolver` to take an injected exception so each test surfaces the same exception types the real `DefaultSchemaResolver` would throw against a misbehaving registry:

- `ApiException` for HTTP error responses
- `UncheckedIOException(ConnectException)` for connection refused
- A real `DescriptorValidationException` (constructed by actually failing protobuf validation on a descriptor with an unresolved field type, not a hand-crafted impostor) for the missing-imports case

Also added two negative tests that prove the narrowed catch behaves correctly:

- `testFallbackDoesNotSwallowProgrammingErrors` confirms an NPE thrown from the resolver propagates unchanged even with the fallback enabled
- `testFallbackDoesNotSwallowIllegalArgument` does the same for config errors

Plus three direct unit tests on the cause-chain classifier covering recoverable detection, chain walking, and rejection of programming errors.

serdes/generic/serde-common-protobuf/src/main/java/io/apicurio/registry/serde/protobuf/ProtobufDeserializer.java
```diff
@@ -94,6 +98,47 @@ public void configure(SerdeConfig configs, boolean isKey) {
...
+     *   <li>Race conditions during rolling deployments (consumer starts before producer registers schema)</li>
+     * </ul>
+     * </p>
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    public U deserializeData(String topic, byte[] data) {
+        try {
+            return super.deserializeData(topic, data);
+        } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
krickert: Pushed `e6bb2730a` to address.

The catch now only triggers the fallback when the cause chain contains one of the specific recoverable failures:

- `com.microsoft.kiota.ApiException` (registry HTTP errors: 404, 5xx, etc.)
- `IOException` / `UncheckedIOException` (network and IO failures, since `IoUtil.toString` wraps `IOException` as `UncheckedIOException`)
- `Descriptors.DescriptorValidationException` (registry returned a descriptor with missing transitive imports)

`NullPointerException`, `IllegalArgumentException`, `ClassCastException` and other programming/config errors now propagate untouched instead of being silently swallowed.

The classifier (`isRecoverableSchemaResolutionError`) walks the cause chain because the resolver's cache/retry layer wraps the original exception, so checking only the top-level type would miss the real cause. The walk is bounded to 32 hops to defend against self-referential chains.  It's a static constant - I'll adjust to another number if you'd like.

carlesarnal:
@krickert please, rebase main, I just fixed an issue with CI.

krickert:
Reworked the catch in `c8c7a12b3` so the fallback no longer relies on a cause-chain classifier. The deserializer now catches the specific unchecked types directly:

```java
try {
    return super.deserializeData(topic, data);
} catch (ApiException | UncheckedIOException e) {
    return tryFallback(topic, data, e);
} catch (IllegalStateException e) {
    if (e.getCause() instanceof DescriptorValidationException) {
        return tryFallback(topic, data, e);
    }
    throw e;
}

```

A few notes on why each catch is shaped the way it is:

- `ApiException` and `UncheckedIOException` are both `RuntimeException` subtypes and reach this method as themselves; no wrapping inspection needed.
- `DescriptorValidationException` and `IOException` are checked exceptions, and `AbstractDeserializer.deserializeData` does not declare any `throws`, so the compiler will not let me write `catch (DescriptorValidationException ...)` here. By the time they reach the deserializer they are always wrapped: `IOException` becomes `UncheckedIOException`, and `ProtobufSchemaParser` wraps `DescriptorValidationException` as `IllegalStateException("Error parsing protobuf schema", dve)`.
- `IllegalStateException` on its own is too broad to catch unconditionally. `DefaultSchemaResolver` also throws it for non-recoverable conditions (`"Wrong configuration"`, `"artifact reference cannot be null"`, etc.), so the catch is narrowed to "direct cause is `DescriptorValidationException`" and otherwise rethrown.

**What this removed:**

- `isRecoverableSchemaResolutionError(Throwable)` and the cause-chain walker
- `MAX_CAUSE_CHAIN_DEPTH` and the cycle-defense logic that came with it
- The classifier-only unit tests (chain walking, cycle, depth bound, `instanceof`-on-each-link checks)

**What this kept or added:**

- All wire-format integration tests still pass: their fixtures (`registryUnreachable()` -> `UncheckedIOException`, `schemaNotFound()` -> `ApiException`, `missingTransitiveImport()` -> `IllegalStateException(DescriptorValidationException)`) match the new typed catches exactly.
- New test `testFallbackDoesNotSwallowGenericIllegalState` pins the narrow `IllegalStateException` rule: an `IllegalStateException("artifact reference cannot be null")` (no `DescriptorValidationException` cause) propagates unchanged with fallback enabled.
- Existing `testFallbackDoesNotSwallowProgrammingErrors` and `testFallbackDoesNotSwallowIllegalArgument` continue to verify NPE / IAE propagate untouched.

Tests: 12 fallback + 8 parser, all green. Checkstyle clean.

krickert:
Addresses the SonarCloud findings on this PR (22 issues, all resolved). Pure code-quality cleanup, no behavior changes; all 20 tests still green.

**`ProtobufDeserializer.java` (4)**

- **S2629** (MAJOR, L175) - wrapped `log.warn(...)` in `if (log.isWarnEnabled())` so `rootCauseMessage(originalError)` and `specificReturnClass.getName()` aren't evaluated when WARN is disabled.
- **S1905** (MINOR, L177) - dropped the redundant `(U)` cast on `parseFallback(data)` (it already returns `U`) and removed the now-unneeded `@SuppressWarnings("unchecked")` on `tryFallback`.
- **S2674** (MINOR BUG, L415/L417) - collapsed the two unchecked `bais.skip()` calls into a single checked skip that throws on a truncated wire-format prefix:

```java
  int prefix = 1 + idSize;
  if (bais.skip(prefix) != prefix) {
      throw new IllegalStateException(
              "Wire-format prefix truncated: expected " + prefix + " bytes");
  }
```

**`ProtobufDeserializerFallbackTest.java` (18)**

- **S5786** (INFO, x13) - removed `public` from the test class and all 12 `@Test` methods; JUnit 5 does not require it.
- **S1488** (MINOR, L340) - inlined the `apiEx` temp variable in `schemaNotFound()` and moved the status-code note into the javadoc.
- **S1186** (CRITICAL, x4) - added one-line explanations to the four no-op `ThrowingSchemaResolver` methods (`setClientFacade`, `setArtifactResolverStrategy`, `reset`, `close`) so the empty bodies read as intentional.

Tests: 12 fallback + 8 parser, all green. Checkstyle clean.

serdes/generic/serde-common-protobuf/src/main/java/io/apicurio/registry/serde/protobuf/ProtobufDeserializer.java
```diff
@@ -2,9 +2,11 @@
 
 import com.google.protobuf.DescriptorProtos;
 import com.google.protobuf.Descriptors.Descriptor;
+import com.google.protobuf.Descriptors.DescriptorValidationException;
 import com.google.protobuf.Descriptors.FileDescriptor;
 import com.google.protobuf.DynamicMessage;
 import com.google.protobuf.Message;
+import com.microsoft.kiota.ApiException;
```
carlesarnal, line 9: **Design question:** This import couples `serde-common-protobuf` (the generic, transport-agnostic serde layer) directly to `com.microsoft.kiota.ApiException`, which is the HTTP client library used by the registry Java SDK. If the SDK ever switches HTTP clients or `ApiException` moves, this catch breaks.

A few options to consider:
1. Accept the coupling with a comment explaining the rationale
2. Check by class name (`e.getClass().getName()`) to avoid compile-time dependency
3. Catch at the boundary in the Kafka-specific subclass (`ProtobufKafkaDeserializer`) instead, where the SDK dependency already exists

Not necessarily a blocker — just worth a deliberate decision.

serdes/generic/serde-common-protobuf/src/main/java/io/apicurio/registry/serde/protobuf/ProtobufDeserializer.java
```diff
@@ -32,9 +35,12 @@ public class ProtobufDeserializer<U extends Message> extends AbstractDeserialize
 
     private final ProtobufSchemaParser<U> parser = new ProtobufSchemaParser<>();
 
+    private static final org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(ProtobufDeserializer.class);
+
```
carlesarnal, line 41: **Checkstyle:** The SLF4J types should use standard imports at the top of the file rather than fully-qualified inline references:
```java
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
// ...
private static final Logger log = LoggerFactory.getLogger(ProtobufDeserializer.class);
```
The project convention is explicit imports (no star imports, no inline FQCNs).

serdes/generic/serde-common-protobuf/src/main/java/io/apicurio/registry/serde/protobuf/ProtobufDeserializer.java
```diff
@@ -293,4 +385,72 @@ private int readRawVarint32(InputStream is) throws IOException {
...
+     * The method respects the {@code readIndexes} and {@code readTypeRef} configuration flags
+     * to correctly skip any prefix data before the actual protobuf payload.
+     * </p>
+     *
+     * @param data the raw Kafka message value bytes including wire-format prefix
+     * @return the deserialized protobuf message, or {@code null} if data is null
+     * @throws IllegalStateException if the fallback parsing fails
+     */
+    @SuppressWarnings("unchecked")
+    private U parseFallback(byte[] data) {
```
carlesarnal, line 368: **Assumption worth documenting:** This method only handles the non-headers wire format (magic byte + ID in payload). Headers-mode deserialization (ID in Kafka headers, no magic byte prefix) goes through a different code path in `KafkaDeserializer`, so the `data[0] == 0x00` check is correct here. But if someone later changes how headers-mode dispatches, this fallback could silently produce corrupt results. A one-line comment noting this assumption would help future maintainers.

serdes/generic/serde-common-protobuf/src/test/java/io/apicurio/registry/serde/protobuf/ProtobufDeserializerFallbackTest.java
```diff
@@ -0,0 +1,479 @@
...
+                DescriptorProtos.FileDescriptorProto.class, registryUnreachable());
+
+        DescriptorProtos.FileDescriptorProto result = deserializer.deserializeData("test-topic", null);
+
+        assertNull(result, "Null input should return null per Kafka contract");
+    }
+
+    // =========================================================================
+    // Negative tests: fallback should NOT activate
+    // =========================================================================
```
carlesarnal, line 214: **Suggestion:** Consider adding a negative test verifying that fallback does **not** activate when `specificReturnClass` is `DynamicMessage.class`. The `fallbackEnabled()` method explicitly guards against this, but there is no test covering it. A `DynamicMessage` fallback would fail at runtime (no schema/descriptor available), so this guard is load-bearing.

carlesarnal (changes requested):
Good progress on narrowing the exception handling — the selective catch approach is much better than the original broad `RuntimeException` catch. The test suite is thorough with strong negative tests. A few items below.

serdes/generic/serde-common-protobuf/src/main/java/io/apicurio/registry/serde/protobuf/ProtobufDeserializer.java
```diff
@@ -94,6 +108,92 @@ public void configure(SerdeConfig configs, boolean isKey) {
...
+     *   <li>Registry downtime or transient network failures</li>
+     *   <li>Missing transitive proto imports in the registry</li>
+     *   <li>Race conditions during rolling deployments (consumer starts before producer
+     *       registers schema)</li>
+     * </ul>
+     */
+    @Override
+    public U deserializeData(String topic, byte[] data) {
+        try {
+            return super.deserializeData(topic, data);
```
carlesarnal, line 115: **Style:** The Javadoc on `deserializeData()` (~35 lines) and `parseFallback()` (~20 lines) is much more verbose than the rest of this file, where methods have minimal or no Javadoc. Consider trimming to 1-2 lines each — the method names and the PR description already explain the intent.

serdes/generic/serde-common-protobuf/src/test/java/io/apicurio/registry/serde/protobuf/ProtobufDeserializerFallbackTest.java
```diff
@@ -0,0 +1,509 @@
...
+        configs.put("apicurio.registry.url", "http://localhost:8081/apis/registry/v3");
+        configs.put(ProtobufDeserializerConfig.FALLBACK_ON_SCHEMA_ERROR, Boolean.TRUE);
+
+        ProtobufDeserializerConfig config = new ProtobufDeserializerConfig(configs, false);
+
+        assertTrue(config.fallbackOnSchemaError(),
+                "Fallback should accept Boolean.TRUE directly");
+    }
+
+    /**
```
carlesarnal, line 110: **Minor:** The `ProtobufDeserializer` instances created via `createFallbackDeserializer()` are never closed. Consider using try-with-resources or an `@AfterEach` cleanup — `ProtobufDeserializer` implements `Closeable` and leaving it open could mask resource-leak issues in future changes.

carlesarnal (commented):
Two minor items not covered in previous rounds.

krickert:
@carlesarnal made the requested changes.. LMK if this is good.

carlesarnal (approved):

carlesarnal:
Thanks!

PR #7631, opened 2026-03-25
fix: race condition in DocumentationTabContent and OpenApiVisualizer

nicomoya123:
## Summary

Fixes a race condition introduced in 3.2.0 that causes the **Documentation tab to render blank** for OpenAPI artifacts.

## Root Cause

Apicurio 3.2.0 changed from inline rendering to an **iframe + `postMessage`** architecture for the documentation tab. PatternFly 6 tabs are **eagerly rendered** (not lazy), so `DocumentationTabContent` mounts _before_ the async `getArtifactVersionContent` API call resolves.

Two compounding bugs result in an empty spec being sent to the Redoc iframe:

1. **`DocumentationTabContent`** uses `useState(parseContent(props.versionContent))` — React's `useState` initializer only runs once on mount. When `versionContent` is an empty string at mount time (data not yet fetched), `parsedContent` stays `{}` even after the prop updates.

2. **`OpenApiVisualizer`** sends the spec to the iframe in `onLoad`, which fires immediately after mount — before the data arrives. There was no mechanism to re-send the spec when it later became available.

### `DocumentationTabContent.tsx`
- Replaced `useState(parseContent(props.versionContent))` with a lazy initializer that guards against empty content
- Added a `useEffect` that calls `setParsedContent` reactively whenever `versionContent` prop changes

### `OpenApiVisualizer.tsx`
- Added `iframeLoaded` ref to track whether the iframe has finished loading
- Extracted `sendSpecToIframe()` helper to deduplicate postMessage logic
- Added `useEffect` that re-sends the spec to the iframe whenever `props.spec` changes to a non-empty value after the iframe is ready

### DIFF
--- a/ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
+++ b/ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
@@ -84,7 +84,7 @@ export const DocumentationTabContent: FunctionComponent<DocumentationTabContentP
     const [error] = useState<any>();
     const groups: GroupsService = useGroupsService();
 
-    // Sync parsed content when versionContent arrives after mount (race condition fix).
+    // Handles race condition where versionContent arrives after mount.
     useEffect(() => {
         if (props.versionContent && !needsDereference(props.artifactType)) {
             setParsedContent(parseContent(props.versionContent));


EricWittmann (approved):
LGTM - thanks for finding the bug and fixing it.  👍

Issue #7315, opened 2026-02-10
Create abstract content validator base class


carlesarnal:
## Parent Epic
Part of #7305 - Codebase Refactoring Epic

## Phase
**Phase 3: Structural Improvements**

## Problem

Multiple content validators implement nearly identical reference validation logic:

**Affected Files:**
- `schema-util/avro/src/main/java/io/apicurio/registry/avro/rules/validity/AvroContentValidator.java` (lines 54-83)
- `schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufContentValidator.java` (lines 109-136)
- `schema-util/json/src/main/java/io/apicurio/registry/json/rules/validity/JsonSchemaContentValidator.java` (lines 74-91)
- `schema-util/openapi/src/main/java/io/apicurio/registry/openapi/rules/validity/ApicurioDataModelContentValidator.java` (lines 68-80)

**Duplicated Pattern:**
```java
Set<String> mappedRefs = references.stream()
    .map(ref -> ref.getName())
    .collect(Collectors.toSet());

Set<String> allRefs = findReferences(content);
Set<RuleViolation> violations = allRefs.stream()
    .filter(ref -> !mappedRefs.contains(ref))
    .map(missingRef -> new RuleViolation(...))
    .collect(Collectors.toSet());

if (!violations.isEmpty()) {
    throw new RuleViolationException(...);
}
```

## Proposed Solution

Create an abstract base class with the common validation logic:

```java
package io.apicurio.registry.rules.validity;

public abstract class AbstractContentValidator implements ContentValidator {
    
    protected void validateMappedReferences(TypedContent content,
            List<ArtifactReference> references,
            Set<String> requiredReferences,
            String violationDescription) throws RuleViolationException {
        
        Set<String> mappedRefNames = references.stream()
            .map(this::extractReferenceName)
            .collect(Collectors.toSet());
        
        Set<RuleViolation> violations = requiredReferences.stream()
            .filter(ref -> !mappedRefNames.contains(ref))
            .map(missingRef -> new RuleViolation(violationDescription, missingRef))
            .collect(Collectors.toSet());
        
        if (!violations.isEmpty()) {
            throw new RuleViolationException(
                "Missing required references", RuleType.VALIDITY, violations);
        }
    }
    
    protected String extractReferenceName(ArtifactReference ref) {
        return ref.getName(); // Can be overridden if needed
    }
}
```

## Estimated Impact
[...]

Issue #7305, opened 2026-02-10
[EPIC] Codebase Refactoring: Reduce duplication and improve performance


carlesarnal:
## Overview

This epic tracks a comprehensive refactoring effort to reduce code duplication, improve performance, and enhance maintainability across the Apicurio Registry codebase.

**Analysis Document:** `claudedocs/codebase-analysis-duplicities-inefficiencies.md`

## Summary Metrics

| Metric | Value |
|--------|-------|
| **Total duplicated lines identified** | ~1,500-2,000 |
| **Files requiring cleanup** | 50+ |
| **TODO comments outstanding** | 70+ |
| **Performance-impacting patterns** | 3 critical |

---

## Phase 1: High Impact, Low Risk ✅ COMPLETED

- [x] #7306 - Extract `ParameterValidationUtils` shared class (50-75 lines)
- [x] #7307 - Merge `CommonResourceOperations` v2/v3 duplicates (26 lines)
- [x] #7308 - Extract shared `ApiUtil` transformation methods (100+ lines)
- [x] #7309 - Create `BoundedCacheFactory` utility (20 lines)

---

## Phase 2: Storage Performance 🔄 IN PROGRESS

- [x] #7310 - Implement batch reference resolution (fix N+1 queries) - **High Performance**
- [x] #7311 - Move orphan content cleanup to scheduled job - **High Performance** (PR #7380)
- [ ] #7312 - Add recursion depth limits for reference resolution - **Security**
- [ ] #7313 - Simplify storage decorator chain architecture - **Medium**

---

## Phase 3: Structural Improvements

- [ ] #7314 - Convert `ArtifactType` from String constants to Enum - **Type Safety**
- [ ] #7315 - Create abstract content validator base class (80+ lines)
- [ ] #7316 - Create content canonicalizer base class (30-40 lines)
- [ ] #7317 - Remove deprecated operator code since 3.0.7 (~200 lines)

---

## Phase 4: Long-Term Architecture

- [ ] #7318 - Implement configuration-driven ArtifactTypeUtilProviders (700+ lines)
- [ ] #7319 - Create platform-agnostic serdes abstraction - **Enables Pulsar/NATS parity**
- [ ] #7320 - Consolidate serdes constructor patterns (160+ lines)

---

## Priority Areas by Impact

| Area | Estimated Lines Saved | Priority |
|------|----------------------|----------|
| ArtifactTypeUtilProvider boilerplate | 700+ | High |
[...]

PR #7644, opened 2026-03-25
refactor: Create abstract content validator base class

matheusandre1:
Part Of #7305 

Closes: #7315

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AbstractContentValidator.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AbstractContentValidator.java
@@ -7,6 +7,7 @@ import io.apicurio.registry.rules.violation.RuleViolationException;
 import io.apicurio.registry.types.RuleType;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AbstractContentValidator.java
```diff
@@ -0,0 +1,57 @@
...
+     * @throws RuleViolationException if any required references are missing
+     */
+    protected void validateMappedReferences(List<ArtifactReference> references,
+            Set<String> requiredReferences, String violationDescription) throws RuleViolationException {
+
+        Set<String> mappedRefNames = references.stream()
+        .map(ref -> extractReferenceName(ref))
+        .filter(Objects::nonNull)
+        .collect(Collectors.toSet());
+
```
carlesarnal, line 35: Nit: `.map(ref -> extractReferenceName(ref))` can be simplified to `.map(this::extractReferenceName)`.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/validity/JsonSchemaContentValidator.java
```diff
@@ -74,20 +72,8 @@ public void validate(ValidityLevel level, TypedContent content,
     @Override
     public void validateReferences(TypedContent content, List<ArtifactReference> references)
             throws RuleViolationException {
```
carlesarnal, line 74: Very minor: the original exception message was `"Unmapped reference(s) detected."` (with `(s)`), now it's `"Unmapped reference detected."` since `validateMappedReferences` uses the same string for both the violation and the exception. Same applies to OpenAPI/AsyncAPI and Protobuf. Not a blocker since these are internal error messages, but worth noting in case the message text is tested anywhere.

Issue #7324, opened 2026-02-10
Consolidate utils/kafka module into common


carlesarnal:
## Summary

The `utils/kafka` module contains only a single utility class (23 lines) and should be consolidated into the `common` module.

## Location

`/utils/kafka/`
- 23 lines of code
- 1 Java file: `KafkaUtil.java`
- **Zero tests**
- Used by: `app`, `integration-tests`, `extra-tests`

## Problem

- Module contains a single static method to convert `KafkaFuture<T>` to `CompletableFuture<T>`
- Too small to justify a separate Maven module
- Adds build complexity without benefit
- Module overhead (pom.xml, directory structure) for trivial code

## Current Code

```java
public class KafkaUtil {
    public static <T> CompletableFuture<T> toCompletableFuture(KafkaFuture<T> kafkaFuture) {
        CompletableFuture<T> completableFuture = new CompletableFuture<>();
        kafkaFuture.whenComplete((result, throwable) -> {
            if (throwable != null) {
                completableFuture.completeExceptionally(throwable);
            } else {
                completableFuture.complete(result);
            }
        });
        return completableFuture;
    }
}
```

## Proposed Solution

1. Move `KafkaUtil.java` to `common` module (or directly into `app`)
2. Update imports in dependent modules
3. Remove `utils/kafka` module

## Risk Assessment

**Very Low Risk** - Simple code move with no functional changes.

## Acceptance Criteria

- [ ] Move `KafkaUtil.java` to appropriate location
- [ ] Update all imports in `app`, `integration-tests`, `extra-tests`
- [ ] Remove `utils/kafka/` directory
- [ ] Remove module reference from parent `pom.xml`
- [ ] Build passes

PR #7654, opened 2026-03-26
Consolidate utils/kafka module into common #7324

matheusandre1:
Closes: #7324


carlesarnal (approved):
LGTM. Clean module consolidation — single utility class moved to common with package path preserved, all pom.xml references updated, kafka-clients dependency correctly scoped as provided. No functional changes, all CI green.

PR #7740, opened 2026-04-10
docs: vale DITA compliance fixes for assembly-intro-to-the-registry

eameh:


eameh:
Closing: approach needs refinement per review feedback. Will resubmit with corrected strategy.

PR #7756, opened 2026-04-13
feature/mvn plugin allow cli

ivangsa:
- adds support for setting all properties from mvn command line on register goal
- fix: set absolute path when parsing artifacts, so references are build ok


carlesarnal (commented):
Thanks for the PR! The CLI list-property support is well implemented. I've left inline comments on the main concerns. Also note: this PR overlaps significantly with #7757 — the two should be coordinated to avoid conflicts.

utils/maven-plugin/src/main/java/io/apicurio/registry/maven/RegisterRegistryMojo.java
```diff
@@ -247,7 +253,128 @@ private static void applyCliArtifactField(RegisterArtifact artifact, String fiel
...
+
+        String nestedField = secondDotIdx == -1 ? null : remainder.substring(secondDotIdx + 1);
+        if (nestedField != null && nestedField.isEmpty()) {
+            return null;
+        }
+
+        return new ParsedListProperty(listName, Integer.parseInt(indexSegment), nestedField);
+    }
+
+    private static <T> T getOrCreateListItem(List<T> list, java.util.function.Consumer<List<T>> setter, int index,
```
carlesarnal, line 357: No upper-bound check on the index here. A property like `-Dartifacts.references.999999.name=foo` would allocate a million-element list of nulls. Consider adding a reasonable cap (e.g., 100) with a clear error message to prevent accidental resource exhaustion.

utils/maven-plugin/src/main/java/io/apicurio/registry/maven/RegisterRegistryMojo.java
```diff
@@ -247,7 +253,128 @@ private static void applyCliArtifactField(RegisterArtifact artifact, String fiel
             default:
                 throw new MojoExecutionException("Unsupported CLI property for artifact configuration: "
                         + propertyKey + ". Supported fields include groupId, artifactId, artifactType, file, ifExists, "
-                        + "canonicalize, minify, autoRefs, isDraft, contentType, version, avroAutoRefsNamingStrategy.");
+                        + "canonicalize, minify, autoRefs, isDraft, contentType, version, "
+                        + "avroAutoRefsNamingStrategy, references.<index>.<field>, "
+                        + "existingReferences.<index>.<field>, protoPaths.<index>.");
+        }
```
carlesarnal, line 260: This reuses the artifact field setter for reference fields (`groupId`, `artifactId`, `version`) since `RegisterArtifactReference` extends/shares the same interface. If `RegisterArtifactReference` ever gains a field that shadows or conflicts with a `RegisterArtifact` field, this could silently apply the wrong setter. A brief comment noting this intentional reuse would help future readers.

utils/maven-plugin/src/test/java/io/apicurio/registry/maven/RegisterRegistryMojoCliPropertiesTest.java
```diff
@@ -62,9 +62,57 @@ void testLoadsSingleArtifactFromCliProperties() throws IOException, MojoExecutio
...
-        // RegisterRegistryMojo.applyCliArtifactField and the expected set below.
+    void testLoadsNestedArtifactListsFromCliProperties() throws IOException, MojoExecutionException {
+        Path artifactFile = Files.createFile(tempDir.resolve("api.yaml"));
+        Path protoPathOne = Files.createDirectory(tempDir.resolve("proto"));
+        Path protoPathTwo = Files.createDirectory(tempDir.resolve("shared-proto"));
+
+        System.setProperty("artifacts.groupId", "my.group.id");
+        System.setProperty("artifacts.artifactId", "my-artifact-id");
+        System.setProperty("artifacts.artifactType", "ASYNCAPI");
+        System.setProperty("artifacts.file", artifactFile.toString());
```
carlesarnal, line 74: This test sets 12 system properties. Make sure the `@AfterEach` cleanup (via `propertiesToClear`) includes all of these new keys (`artifacts.references.0.name`, `artifacts.references.0.groupId`, `artifacts.existingReferences.0.resourceName`, `artifacts.protoPaths.0`, `artifacts.protoPaths.1`, etc.), otherwise they'll leak into other tests and could cause hard-to-debug failures.

docs/modules/ROOT/pages/getting-started/assembly-managing-registry-artifacts-maven.adoc
```diff
@@ -98,6 +98,31 @@ mvn io.apicurio:apicurio-registry-maven-plugin:${apicurio.version}:register \
...
+[source,shell]
+----
+mvn io.apicurio:apicurio-registry-maven-plugin:${apicurio.version}:register \
+  -Dapicurio.url=MY-REGISTRY-URL/apis/registry/v3 \
+  -Dartifacts.0.groupId=TestGroup \
+  -Dartifacts.0.artifactId=ExampleAPI \
+  -Dartifacts.0.artifactType=ASYNCAPI \
+  -Dartifacts.0.file=./src/main/resources/apis/example.yaml \
+  -Dartifacts.0.references.0.name=/apis/registry/v3/groups/master/artifacts/kafka-bindings.yml/versions/branch=latest/content#/components/messageTraits/kafkaKeyString \
+  -Dartifacts.0.existingReferences.0.resourceName=./avro/shared.avsc \
```
carlesarnal, line 118: This reference name contains a registry API path (`/apis/registry/v3/groups/.../content#/...`) which looks like it's from the #7757 use case (automatic registry reference resolution). In the context of this PR alone — which doesn't add that feature — users might try this and be confused about what it does. Consider using a simpler, self-contained example for the reference name.

carlesarnal (changes requested):

utils/maven-plugin/src/test/java/io/apicurio/registry/maven/RegisterRegistryMojoCliPropertiesTest.java
ivangsa:
```
   @AfterEach
    void clearArtifactProperties() {
        System.getProperties().stringPropertyNames().stream()
                .filter(name -> name.startsWith("artifacts."))
```
cleared all that start with `artifacts.`

ivangsa:
we can close this one as is it was merged with #7772

PR #7757, opened 2026-04-13
feat: automatic resolve registered references on maven register goal

ivangsa:
The Maven register goal now recognizes absolute references that already point to artifacts registered in the same Apicurio Registry and attaches them as references instead of trying to resolve them as local files.

## Includes

- Resolve same-registry absolute references during autoRefs
- Normalize generated reference names to registry-relative paths

## Also 

If and absolute $ref is not pointing to apicurio registry, ignore them for autorefs, otherwise they fail as they can not be resolved localy, specially on windows paths


carlesarnal (commented):
Thanks for the PR! The feature is well-conceived and addresses a real pain point. I've left a few inline comments on the main concerns.

utils/maven-plugin/src/main/java/io/apicurio/registry/maven/refs/IndexedResource.java
```diff
@@ -8,6 +8,8 @@
 import java.util.Objects;
 import java.util.Set;
 
+import static io.apicurio.registry.maven.RegisterRegistryMojo.isAbsoluteUri;
```
carlesarnal: This creates a dependency from the `refs` package back to `RegisterRegistryMojo`, which is a circular-ish coupling (Mojo already depends on `refs`). Consider moving `isAbsoluteUri` to a small shared utility class or directly into `IndexedResource` itself — the logic is trivial and self-contained, and it would keep the dependency graph clean.

utils/maven-plugin/src/main/java/io/apicurio/registry/maven/RegisterRegistryMojo.java
```diff
@@ -394,6 +537,147 @@ private VersionMetaData registerWithAutoRefs(RegistryClient registryClient, Regi
...
+        String registryPath = stripTrailingSlash(registryUri.getRawPath());
+        Pattern registryArtifactUrlPattern = Pattern.compile("^" + Pattern.quote(registryPath)
+                + "/groups/([^/]+)/artifacts/([^/]+)/versions/([^/]+)(?:/content)?$");
+        Matcher matcher = registryArtifactUrlPattern.matcher(resourceUri.getRawPath());
+        if (!matcher.matches()) {
+            return Optional.empty();
+        }
+
+        return Optional.of(new RegistryReferenceLocation(
+                decodePathSegment(matcher.group(1)),
```
carlesarnal: This `Pattern` is compiled on every call to `parseRegistryReferenceLocation`, but `registryPath` comes from `registryUrl` which doesn't change during execution. For a large schema graph with many references, this could add up. Consider caching the compiled pattern as a lazily-initialized field.

utils/maven-plugin/src/main/java/io/apicurio/registry/maven/RegisterRegistryMojo.java
```diff
@@ -247,7 +256,128 @@ private static void applyCliArtifactField(RegisterArtifact artifact, String fiel
...
+        if (item == null) {
+            item = supplier.get();
+            list.set(index, item);
+        }
+        return item;
+    }
+
+    private static <T> void ensureListSize(List<T> list, int index) {
+        while (list.size() <= index) {
+            list.add(null);
```
carlesarnal, line 369: Using \`debug\` level here means this is invisible by default. Since silently dropping a reference could mask misconfiguration (e.g. a typo in the registry URL), consider using \`warn\` instead so users get visibility when a reference is unexpectedly skipped.

utils/maven-plugin/src/main/java/io/apicurio/registry/maven/RegisterRegistryMojo.java
```diff
@@ -394,6 +537,147 @@ private VersionMetaData registerWithAutoRefs(RegistryClient registryClient, Regi
```
carlesarnal, line 393: The new URL-parsing and registry-resolution methods (\`parseRegistryReferenceLocation\`, \`isSameApicurioServer\`, \`registryReferenceName\`, \`decodePathSegment\`) contain non-trivial logic with edge cases (port normalization, trailing-slash stripping, percent decoding). They would benefit from dedicated unit tests — the CLI list-property parsing is well tested, but this part of the PR has no test coverage.

carlesarnal (changes requested):

ivangsa:
we can close this one as is it was merged with #7772

PR #7762, opened 2026-04-14
vale fixes for assembly-config-reference

eameh:


docs/modules/ROOT/pages/getting-started/assembly-config-reference.adoc
```diff
@@ -1,34 +1,35 @@
 include::{mod-loc}shared/all-attributes.adoc[]
 
+:_mod-docs-content-type: ASSEMBLY
 [id="registry-config-reference_{context}"]
 = {registry} configuration reference
-//If the assembly covers a task, start the title with a verb in the gerund form, such as Creating or Configuring.
 
 [role="_abstract"]
-This chapter provides reference information on the configuration options that are available for {registry}.
+You can use the following reference information to configure {registry}.
```
pwright, line 8: not really an abstract?

pwright: recommend reducing to single module

docs/modules/ROOT/pages/getting-started/assembly-config-reference.adoc
```diff
@@ -1,34 +1,35 @@
 include::{mod-loc}shared/all-attributes.adoc[]
 
+:_mod-docs-content-type: ASSEMBLY
```
pwright, line 3: attribute must be below id

eameh: Updated

eameh:
Hi @pwright, attributes now follow this format
id
attribute
heading

PR will be ready to merge pending your approval

pwright (commented):
should https://github.com/Apicurio/apicurio-registry/blob/main/docs/modules/ROOT/partials/getting-started/ref-registry-all-configs.adoc be part of this pr?

why are there two files anyway, does one file get reused

eameh:
> should https://github.com/Apicurio/apicurio-registry/blob/main/docs/modules/ROOT/partials/getting-started/ref-registry-all-configs.adoc be part of this pr?
> 
> why are there two files anyway, does one file get reused

It does not get reused, and it makes sense to include it (so I have). However, the concern is that it is quite large. Adding it could result in a massive monolith, which may be difficult to maintain.

pwright (dismissed):
looks good, but need to delete old file that you've integrated in here

eameh:
The stale file has been removed @pwright 

@EricWittmann, this PR is ready to merge

PR #7772, opened 2026-04-16
feat: support setting api.version from OpenAPI/AsyncAPI `info.version` in maven register goal

ivangsa:
This is the third and last PR about allowing registering artifacts from the cli with minimal or auto-configuration.

Adds support for setting api.version from OpenAPI/AsyncAPI `info.version` in maven register goal

- if set `versionStrategy` to `API_INFO_VERSION` the plugin reads the version from the `info.version` field in the API document
- if version ends with `-SNAPSHOT`, the Maven plug-in strips the suffix before registration and registers the version as a DRAFT.
- when the API document later uses the same version without the `-SNAPSHOT` suffix, the Maven plug-in updates the existing draft content and promotes that version to `ENABLED`.

For other artifact types, `API_INFO_VERSION` is ignored.


EricWittmann (approved):
LGTM

ivangsa:

PR #7784, opened 2026-04-19
feat(protobuf): harden upload validation against malformed and conflicting schemas

krickert:
# Summary

The registry currently stores two kinds of broken protobuf artifacts: binary descriptors whose identifiers violate the protobuf grammar, and multi-reference uploads that redefine the same fully qualified name with incompatible wire shapes. This PR rejects both at upload time.

Adds two defensive checks to `ProtobufContentValidator`:

1. Binary `FileDescriptorProto` uploads are validated against the protobuf identifier grammar before being accepted.
2. Across a main artifact and its resolved references, any fully qualified name redefined with a different wire shape is rejected.

Both are hardening. The first prevents the registry from storing descriptors whose identifiers fall outside the protobuf language spec; the second turns a silent integrity hazard into a clear validation error at upload time.

# Motivation

**Malformed identifiers in binary uploads.** When a Protobuf artifact is uploaded as a base64-encoded `FileDescriptorProto`, the Wire binary-to-text round trip used during validation silently drops identifiers that aren't valid protobuf names (for example `Data(){console}`). The descriptor is stored with the invalid names intact, and every downstream consumer has to defend itself against shapes the protobuf language doesn't permit. The registry should reject them at the boundary.

**Conflicting FQNs across references.** A main artifact can import two references that both declare the same fully qualified name with different fields. Each reference is individually valid, so protobuf-java accepts them. Consumers that merge the descriptors at runtime are left with non-deterministic behavior depending on which reference loads first. The registry is in the best position to catch this.

# How it works

**Binary-identifier check.** Base64 uploads are detected by a strict `Base64.getDecoder()` attempt (text protos always contain characters outside the base64 alphabet, so no keyword heuristic is needed) and validated directly with `Descriptors.FileDescriptor.buildFrom`, the canonical check shipped with protobuf-java.

**Cross-file FQN conflict detection.** Each source is compiled to a `FileDescriptorProto` and its `DescriptorProto` / `EnumDescriptorProto` / `ServiceDescriptorProto` are walked per FQN. Comparison uses `Message.equals()` on the compiled descriptor, so comments and whitespace are irrelevant by construction; wire-layout detail (field names, types, tags, labels, oneofs, reserved ranges, options) is what drives equality. Three details that mattered in the implementation:

- References are walked before main so flattened same-package types retain correct attribution in error messages.
- Each source's dependency map is filtered to its declared imports to prevent the converter from merging a conflicting reference's types into another reference's descriptor.
- Messages are compared with nested types/enums cleared so a conflict on an inner type reports the inner FQN rather than its wrapper.

Both checks raise `RuleViolationException(RuleType.VALIDITY, ...)`, consistent with every other content validator in `schema-util/`.

## Changes

| File | Change |
|------|--------|
| `ProtobufContentValidator.java` | Call the new detector first; let `RuleViolationException` propagate unwrapped |
| `ProtobufFqnConflictDetector.java` | New package-private detector: binary identifier check + FDP-based cross-file conflict detection |
| `ProtobufContentValidatorTest.java` | 17 new cases: message / enum / nested / service conflicts, identifier injection in main and references, missing-name and field-name paths, `FULL`-level validation, null resolved references, mixed binary/text refs, proto2 extension usage (single-file and with a shared extension-definition reference), plus three positive regressions for comment-only, whitespace-only, and combined cosmetic-difference tolerance |

## Test plan

- [x] `./mvnw checkstyle:check -pl schema-util/protobuf` is clean
- [x] `./mvnw test -pl schema-util/util-provider` passes 153/153 (28 in `ProtobufContentValidatorTest`)
- [x] CI green on all storage variants


EricWittmann (dismissed):
Solid security hardening with thorough test coverage. The base64 detection approach is clean, the FQN conflict detection is well-scoped, and the 17 test cases cover the important positive and negative scenarios. LGTM.

A couple of minor observations for follow-up (non-blocking):

1. Consider moving `assertNoConflicts()` outside the try/catch block in `ProtobufContentValidator.validate()` rather than adding the `catch (RuleViolationException rve)` re-throw. That would be less brittle if someone adds another code path later.

2. `buildDepsTextMap` runs `ProtobufFile.toProtoFileElement().toSchema()` on all references including binary ones — worth verifying this doesn't blow up when a reference is base64-encoded (the mixed binary/text test passes, so it may be fine in practice).

3. The reference text conversion happens twice (once in the detector, once in the existing validator) — not a correctness issue but worth noting for performance if uploads with many references become common.

EricWittmann:
Thanks for the contribution.  I'll wait a couple of days before merging in case you want to address any of my (minor) feedback issues above.

krickert:
> Thanks for the contribution. I'll wait a couple of days before merging in case you want to address any of my (minor) feedback issues above.

 Thanks for keeping up with a cool product.  Made the changes and updated the code.. should be better now.

krickert:
@EricWittmann ok - latest merge should make it as 0 sonar warnings/errors too.. so we should be good to go.

schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufContentValidator.java
```diff
@@ -37,6 +37,13 @@ public ProtobufContentValidator() {
     public void validate(ValidityLevel level, TypedContent content,
                          Map<String, TypedContent> resolvedReferences) throws RuleViolationException {
         if (level == ValidityLevel.SYNTAX_ONLY || level == ValidityLevel.FULL) {
+            // FQN / identifier hardening runs first, intentionally outside the wrapper
+            // try/catch below. The detector raises RuleViolationException directly, and
+            // the wrapper exists only to convert unexpected parser failures into syntax
+            // violations — re-wrapping the detector's own exceptions would lose the
```
carlesarnal: **Exception handling gap:** `assertNoConflicts` is intentionally placed outside the `try/catch` below so that its `RuleViolationException` propagates directly — that's correct. However, `assertNoConflicts` also calls `buildDepsTextMap`, which runs `ProtobufFile.toProtoFileElement(...).toSchema()` on every reference. If a reference is malformed and that conversion throws a `RuntimeException`, it will now propagate **unwrapped** to the caller.

Before this PR, the same conversion (lines 79-81 on main) was inside the `try/catch`, so failures were wrapped in `RuleViolationException("Syntax violation...")`. After this PR, a bad reference causes a raw `RuntimeException` instead of the structured validation error.

Suggestion: wrap the `assertNoConflicts` call in its own try/catch that lets `RuleViolationException` through but catches and wraps other exceptions:
```java
try {
    referenceTextSchemas = ProtobufFqnConflictDetector
            .assertNoConflicts(level, content, resolvedReferences);
} catch (RuleViolationException rve) {
    throw rve;
} catch (Exception e) {
    throw new RuleViolationException("Syntax violation for Protobuf artifact.",
            RuleType.VALIDITY, level.name(), e);
}
```

schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufFqnConflictDetector.java
```diff
@@ -0,0 +1,385 @@
...
+     *                                identifier; other descriptor-level validation errors
+     *                                fall through to the normal validation flow
+     */
+    private static void validateIdentifiersIfBinary(ValidityLevel level, String sourceName,
+                                                    TypedContent content) throws RuleViolationException {
+        FileDescriptorProto fileProto = tryParseAsBinaryDescriptor(content);
+        if (fileProto == null) {
+            return;
+        }
+        try {
```
carlesarnal, line 167: **Minor — double parse of binary content:** For binary (base64) uploads, `tryParseAsBinaryDescriptor` is called twice on the same content:
1. Here in `validateIdentifiersIfBinary` (line 126 → 160)
2. Again in `walkSource` → `buildFileDescriptorProto` (line 140 → 294)

The base64 decode + protobuf parse isn't expensive for typical schemas, but you could cache the result from the first call to avoid the redundant work. Not a blocker — just a small optimization opportunity.

schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufFqnConflictDetector.java
```diff
@@ -0,0 +1,385 @@
...
+            return null;
+        }
+        byte[] decoded;
+        try {
+            decoded = Base64.getDecoder().decode(trimmed);
+        } catch (IllegalArgumentException notBase64) {
+            return null;
+        }
+        try {
+            return FileDescriptorProto.parseFrom(decoded);
```
carlesarnal, line 209: **Fragile string matching:** `INVALID_IDENTIFIER_MARKER` and `MISSING_NAME_MARKER` rely on exact substrings from protobuf-java's `DescriptorValidationException` message. If protobuf-java changes its wording in a future version, this check would silently stop catching identifier injection.

Unfortunately protobuf-java doesn't provide error codes or typed sub-exceptions, so string matching is the only option. Worth adding a brief comment acknowledging the coupling so a future protobuf-java bump triggers a check of these constants. The existing test coverage (`testRejectsUnsafeIdentifierInBinaryDescriptor`, `testRejectsMissingNameInBinaryDescriptor`) will catch a regression if the wording changes — as long as CI runs those tests.

schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufFqnConflictDetector.java
```diff
@@ -0,0 +1,385 @@
...
+     *
+     * @param message the compiled descriptor being recorded
+     * @param packageName the proto package of the enclosing file
+     * @param scope the dotted scope path accumulated from outer messages (empty for top-level)
+     * @param sourceName the logical source name that contributed this type
+     * @param known the shared FQN map
+     * @param level the current validity level, used for error reporting
+     * @throws RuleViolationException if a conflicting definition is discovered
+     */
+    private static void recordMessage(DescriptorProto message, String packageName, String scope,
```
carlesarnal, line 414: **Good defensive pattern:** Catching `RuntimeException` and returning `null` here lets the downstream validator surface the real parsing error. The comment explains the rationale clearly.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/ProtobufContentValidatorTest.java
```diff
@@ -106,6 +112,595 @@ public void testValidProto3UuidSchema() throws Exception {
...
+                ContentTypes.APPLICATION_PROTOBUF);
+
+        ProtobufContentValidator validator = new ProtobufContentValidator();
+        Map<String, TypedContent> noReferences = Collections.emptyMap();
+        Assertions.assertThrows(RuleViolationException.class,
+                () -> validator.validate(ValidityLevel.SYNTAX_ONLY, content, noReferences));
+    }
+
+    @Test
+    public void testRejectsConflictingFqnAcrossReferencesAtFullLevel() {
```
carlesarnal, line 547: **Nice:** The three cosmetic-tolerance tests (comment-only, whitespace-only, and combined) are important regression tests. They verify that `DescriptorProto` comparison ignores source-level differences that don't affect the wire format — a real-world scenario when schemas pass through different registries or toolchains.

carlesarnal (commented):
Solid security hardening PR. The base64 detection approach is sound, the FQN conflict detection is well-scoped, and the new tests cover a good range of positive, negative, and edge cases.

One correctness concern about exception handling, and a couple of minor observations inline.

schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufContentValidator.java
krickert: Resolved - please check

Issue #7480, opened 2026-03-03
Production-ready GitOps storage for Apicurio Registry


jsenko:
# GitOps Storage for Apicurio Registry - Design Document

## Table of Contents

- [1. Introduction](#1-introduction)
- [2. GitOps Approaches](#2-gitops-approaches)
- [3. Use Cases](#3-use-cases)
- [4. Current Status](#4-current-status)
- [5. High-Level Architecture](#5-high-level-architecture)
  - [5.1 One-Way Synchronization](#51-one-way-synchronization)
  - [5.2 Sidecar Architecture](#52-sidecar-architecture)
  - [5.3 Pull and Push Models](#53-pull-and-push-models)
  - [5.4 Blue-Green Loading](#54-blue-green-loading)
  - [5.5 Multi-Repository Aggregation](#55-multi-repository-aggregation)
- [6. Data Format](#6-data-format)
  - [6.1 File Conventions](#61-file-conventions)
  - [6.2 Data Model](#62-data-model)
  - [6.3 Example Repository Layout](#63-example-repository-layout)
- [7. Change Detection and Loading](#7-change-detection-and-loading)
  - [7.1 Polling and Debouncing](#71-polling-and-debouncing)
  - [7.2 Startup Behavior](#72-startup-behavior)
  - [7.3 Error Handling](#73-error-handling)
- [8. Deployment](#8-deployment)
  - [8.1 Multi-Replica Deployment](#81-multi-replica-deployment)
  - [8.2 Volume Strategy](#82-volume-strategy)
  - [8.3 Security](#83-security)
- [9. Management API](#9-management-api)
- [10. Future Enhancements](#10-future-enhancements)
- [11. Open Questions](#11-open-questions)
- [12. Implementation Plan](#12-implementation-plan)

---

### Why GitOps for a Schema Registry?

Schema registries are critical infrastructure in event-driven architectures. They store and serve schemas (Avro, Protobuf, JSON Schema, OpenAPI, etc.) that define contracts between producers and consumers. Managing these schemas reliably across environments is a challenge that GitOps is well-suited to address.

**GitOps** applies the principles of Git-based version control to infrastructure and application configuration. For a schema registry, this means:

- **Git as the source of truth** - Schemas and their metadata are stored in Git repositories, benefiting from version history, branching, pull requests, and code review workflows.
- **Declarative configuration** - The desired state of the registry is declared in files, not constructed through API calls.
[...]

PR #7790, opened 2026-04-21
GitOps Phase 5a: Error messages, SQL noise, auto-labels, per-repo SSH keys and mode

jsenko:
## Summary

Phase 5a of the GitOps storage implementation ([#7480](https://github.com/Apicurio/apicurio-registry/issues/7480)). Quick wins for error reporting, SQL behavior, labeling, and sidecar configuration.

### Improved error messages

- New `recordError(PollingDataFile, message, ...)` overload includes source repo ID and file path
- Errors now look like: `[repo-a] Could not find group 'payments' (file: orders/order-created.registry.yaml)`
- Updated all error calls in `AbstractPollingDataSourceManager` that have file context

### Fix noisy SQL exceptions

- `ensureGroup()` now checks `isGroupExists()` before attempting INSERT
- Eliminates the unique constraint violation exception logged at DEBUG level on every import cycle when groups are already imported via `importGroup()`

### Auto-label artifacts with source ID

- New config: `apicurio.polling-storage.source-label-key` (default: `system:source`)
- Automatically adds a label with the source repo ID to each imported artifact and group
- Set to empty string to disable
- Useful for traceability and UI filtering in multi-repo setups

### Per-repo SSH keys (sidecar)

- `APICURIO_GITOPS_REPOS_N_SSH_KEYS` — per-repo SSH keys added alongside global `PULL_SSH_KEYS`
- All keys merged into SSH config `IdentityFile` entries

### Per-repo mode (sidecar)

- `APICURIO_GITOPS_REPOS_N_MODE` — overrides global `MODE` per repo (`pull` or `push`)
- Allows mixed pull+push across repos in multi-repo setups
- Pull loop only fetches pull-mode repos, SSH server starts if any repo uses push
- Pull-mode repos naturally reject pushes (no `updateInstead` config)

### Java 25 compatibility

- Default empty `<argLine/>` property in `app/pom.xml` prevents surefire fork crash on Java 25

## Test plan

- [x] All 25 gitops tests pass (smoke, multi-repo, config, status, admin, timestamp)
- [x] `ensureGroup()` no longer logs SQL exceptions during normal import
- [x] Source labels applied to artifacts and groups
- [x] Compilation verified with `clean install`

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -16,6 +16,11 @@
     <projectRoot>${project.basedir}/..</projectRoot>
     <maven.compiler.release>21</maven.compiler.release>
     <groups/>
+    <!-- Default empty argLine so that the @{argLine} late-binding reference in
+         maven-surefire-plugin resolves to empty when no plugin (e.g., JaCoCo) sets it.
+         Without this, Java 25+ treats the literal '@{argLine}' as an arg file reference
+         and fails with 'could not open {argLine}'. -->
+    <argLine/>
   </properties>
 
   <dependencies>
--- a/app/src/main/java/io/apicurio/registry/storage/impl/gitops/README.md
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/gitops/README.md
@@ -140,10 +140,11 @@ Different registry instances point to different branches of the same repository:
 
 Schema promotion is done through Git merges.
 
-### Push Model for Restricted Networks *(planned)*
+### Push Model for Restricted Networks
 
 In environments where outbound network access is restricted, an external process pushes changes to a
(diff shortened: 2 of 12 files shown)


jsenko:
All review comments addressed in 7fce16b:

1. **README var names**: use full `APICURIO_GITOPS_PULL_SSH_KEYS` and `APICURIO_GITOPS_MODE` in table
2. **Per-repo SSH keys**: `setup_ssh_client` now triggered when any SSH keys are configured (global or per-repo)
3. **Process supervision**: `wait -n` in mixed mode exits when either sshd or pull loop dies, so container restarts
4. **Duplicate error**: removed redundant "Could not find group" — `processGroupRef()` already records the detail

PR #7813, opened 2026-04-22
feat: support simultaneous proxy header and OIDC authentication

Joel-hanson:
Add fallback mechanism to enable both proxy header authentication and OIDC authentication simultaneously. When both are enabled, the system first checks for proxy authentication headers. If present, proxy authentication is used. If not, it falls back to OIDC authentication flow.


Joel-hanson:
Closing this PR as there is an existing PR to have a more holistic approach.

https://github.com/Apicurio/apicurio-registry/pull/7821

PR #7821, opened 2026-04-22
feat(auth): refactor authentication to prioritized chain of strategies

### LINKED ISSUE #7812: Support simultaneous proxy header authentication and OIDC authentication

Currently, Apicurio Registry treats proxy header authentication and OIDC authentication as mutually exclusive options. This creates a limitation for deployment scenarios where both authentication methods need to coexist. When deploying Apicurio Registry behind an authenticating proxy or API gateway face a choice: either enable proxy header authentication for API access or enable OIDC for UI access, but not both simultaneously.

In our deployment, we have a custom build authentication service in front of Apicurio Registry that handles authentication using Kafka's native mechanisms like SCRAM-SHA or mutual TLS. Authenticated requests are then forwarded to Apicurio without authentication headers (which we are trying to make it secure by using proxy auth), relying on the trust established by the proxy. However, we also need to support direct access to the Apicurio UI and other OIDC-based interfaces that require proper OIDC authentication flows. The current implementation forces us to choose one authentication method over the other, which doesn't work for our use case.

### Proposed Solution

We propose introducing a fallback mechanism where both proxy header authentication and OIDC authentication can be enabled together. When a request arrives, the system would first check for proxy authentication headers. If those headers are present, proxy authentication would be used. If the headers are not present, the system would fall back to the standard OIDC authentication flow.

This enhancement would benefit anyone running Apicurio Registry behind an authenticating proxy or API gateway that needs to support both proxied requests and direct browser-based access. It eliminates the need to reimplement the OIDC authentication logic in the proxy layer, which would be complex given the intricacies of the OAuth2/OIDC flow that Apicurio already handles well.

The implementation would maintain existing behavior when only one authentication method is enabled, and simply add logic to try proxy authentication first when both are enabled, falling back to OIDC if proxy headers are absent.

The change would be in the `authenticate()` method of `AppAuthenticationMechanism`:

```java
} else if (authConfig.proxyHeaderAuthEnabled && authConfig.oidcAuthEnabled) {
    return proxyHeaderAuthenticationMechanism.authenticate(context, identityProviderManager)
        .onItem().ifNull().switchTo(() -> {
            // Proxy headers not present, try OIDC authentication
            log.debug("Proxy headers not found, falling back to OIDC authentication");
            return authenticateWithOidc(context, identityProviderManager);
        });
}
```

I am happy to contribute this change if this request seems reasonable.

### Additional Context

A common scenario is when Kafka-based authentication (SCRAM, TLS) is handled by a proxy layer that then forwards requests to Apicurio Registry. In such setups, the proxy handles the Kafka authentication and forwards trusted requests without auth headers, while the Apicurio UI still needs OIDC for browser-based user authentication. Supporting both methods simultaneously would allow this architecture to work seamlessly.

EricWittmann:
## Summary

Replaces the if/else cascade in `AppAuthenticationMechanism` with a chain-of-strategies model that allows multiple authentication mechanisms to be enabled simultaneously. This is a generalization of the approach requested in #7812 — instead of adding combination branches for specific pairs (proxy+OIDC), all enabled mechanisms are tried in configurable priority order.

- **New interface** `AuthenticationStrategy` — abstraction for each mechanism in the chain
- **`DelegatingAuthenticationStrategy`** — generic wrapper for mechanisms needing no extra logic (basic, proxy-header)
- **`OidcAuthenticationStrategy`** — encapsulates audit logging, client credentials, password grant, and token caching (extracted from `AppAuthenticationMechanism`)
- **`AppAuthenticationMechanism`** refactored to build a chain at startup and iterate it on each request
- **New config property** `apicurio.authn.mechanism.priority` (default: `basic,proxy-header,oidc`) — controls the order mechanisms are tried
- **Challenge/transport** use the last strategy in the chain (the final fallback)

### How it works

At startup, `buildAuthChain()` iterates the priority list and creates strategies for each enabled mechanism. On each request, strategies are tried in order via Mutiny's `.onItem().ifNull().switchTo()` — if a strategy returns null (can't handle the request), the next one tries.

### Dual proxy+OIDC example

```properties
apicurio.authn.proxy-header.enabled=true
quarkus.oidc.tenant-enabled=true
apicurio.authn.mechanism.priority=proxy-header,oidc
```

Requests with proxy headers → proxy auth. Requests without → OIDC fallback.

Closes #7812

## Test plan

- [ ] New `ProxyAndOidcDualAuthTest` — proxy headers present (proxy wins), absent (OIDC fallback), neither (401), both present (proxy wins), client credentials in dual mode
- [ ] Existing `ProxyHeaderAuthTest` passes unchanged (proxy-only regression)
- [ ] Existing `SimpleAuthTest` passes unchanged (OIDC-only regression)
- [ ] Existing `BasicAuthWithPropertiesTest` passes unchanged (basic-only regression)
- [ ] All other auth tests pass unchanged
- [ ] Checkstyle passes


jsenko:
I've increased the duplication limit to 10% (at least temporarily), but can you check if this is a false positive or are [these lines](https://sonarcloud.io/component_measures?metric=new_duplicated_lines&selected=Apicurio_apicurio-registry%3Autils%2Ftests%2Fsrc%2Fmain%2Fjava%2Fio%2Fapicurio%2Fregistry%2Futils%2Ftests%2FProxyAndOidcAuthTestProfile.java&pullRequest=7821&id=Apicurio_apicurio-registry) actually duplicated?

EricWittmann:
> I've increased the duplication limit to 10% (at least temporarily), but can you check if this is a false positive or are [these lines](https://sonarcloud.io/component_measures?metric=new_duplicated_lines&selected=Apicurio_apicurio-registry%3Autils%2Ftests%2Fsrc%2Fmain%2Fjava%2Fio%2Fapicurio%2Fregistry%2Futils%2Ftests%2FProxyAndOidcAuthTestProfile.java&pullRequest=7821&id=Apicurio_apicurio-registry) actually duplicated?

I mean, they **are** duplicated, but it's just configuration properties in a test profile.  So yeah, there are other test profiles with similar configurations.  I suppose I could create a hierarchy of test profiles, but that seems like overkill just to pass static analysis.

docs/modules/ROOT/partials/getting-started/ref-registry-all-configs.adoc
```diff
@@ -214,6 +214,11 @@ The following {registry} configuration options are available for each component
 |
 |`2.5.0.Final`
 |Client credentials scope.
+|`apicurio.authn.mechanism.priority`
+|`string`
+|`basic,proxy-header,oidc`
+|`3.0.0`
```
Joel-hanson: Shouldn't this be 3.3.0?

EricWittmann: Yep!  Good catch.  :+1:

PR #7829, opened 2026-04-23
feat(auth): per-resource authorization with shared grants evaluator

### LINKED ISSUE #7724: Fine-grained per-resource authorization (resource-level ACLs)
## Summary

Add support for fine-grained, per-resource authorization in Apicurio Registry — enabling policies like "Alice can read artifact X but not artifact Y," independent of her global role.

Today, Registry supports:
- **RBAC** — global roles (`sr-readonly`, `sr-developer`, `sr-admin`) that apply to *all* resources equally
- **OBAC** — owner-based access control, where only the artifact/group creator can modify it

Neither supports arbitrary per-resource visibility or access rules. This issue proposes adding resource-level ACLs so that administrators can control which users (or groups/service accounts) can access specific artifacts and groups.

## Motivation

- **Multi-team environments** — different teams share a single Registry instance but should only see their own schemas/APIs
- **Regulatory/compliance** — certain schemas may contain sensitive structures that not all users should access
- **Gateway-fronted deployments** — organizations using API gateways or service meshes (Authorino, Envoy) for authn want the Registry itself to enforce per-resource authz, especially for list/search filtering which external proxies cannot do

### 1. Point access (easier)
Can user X perform operation Y on artifact Z? This is solvable by adding a new `IAccessController` implementation (see `app/src/main/java/io/apicurio/registry/auth/IAccessController.java`).

### 2. List/search filtering (harder)
When user X searches or lists artifacts, results must exclude resources they cannot see. This requires filtering at the **storage/query level**, not just at the interceptor level. External proxies (Authorino, OPA at the gateway) cannot solve this — it must be done inside the Registry.

### Option A: Internal ACL table (simplest)

Add a permission table to the SQL schema:

```sql
CREATE TABLE resource_permissions (
    principal_id  VARCHAR(255) NOT NULL,
    group_id      VARCHAR(255),
    artifact_id   VARCHAR(255),  -- NULL = group-level permission
    permission    VARCHAR(32) NOT NULL,  -- 'read', 'write', 'admin'
    PRIMARY KEY (principal_id, group_id, artifact_id, permission)
);
```

- **Point access**: New `AclAccessController implements IAccessController` checks the table
- **List filtering**: JOIN/subquery against `resource_permissions` in search queries
- **Management**: New REST API endpoints under `/admin/permissions` (or extend `/admin/roleMappings`)
- **Pros**: No external dependencies, works natively with pagination, fast
- **Cons**: Limited policy expressiveness (no conditional/time-based rules)

### Option B: OPA integration

Add an OPA-based access controller that calls an external OPA server for decisions.

- **Point access**: REST call to OPA with `(principal, group, artifact, action)` as input
- **List filtering**: Use OPA's partial evaluation API to compile policies into SQL WHERE clauses, or post-filter (expensive, breaks pagination)
- **Pros**: Very flexible policies (Rego), decoupled policy management
- **Cons**: Latency per request, list filtering is architecturally complex, additional infrastructure

### Option C: Keycloak Authorization Services (UMA)

Leverage Keycloak's built-in resource/scope/policy model.

- **Point access**: Call Keycloak Authorization API or use `quarkus-keycloak-authorization`
- **List filtering**: Query Keycloak for the set of resources a user can access, build SQL `IN (...)` filter
- **Pros**: Already using Keycloak for OIDC, rich policy UI, UMA standard
- **Cons**: Every artifact creation must register a Keycloak resource, tight Keycloak coupling

## Key integration points in the codebase

| File | Role |
|------|------|
| `app/.../auth/IAccessController.java` | Interface for authorization decisions |
| `app/.../auth/AuthorizedInterceptor.java` | Interceptor that invokes RBAC + OBAC checks |
| `app/.../auth/AbstractAccessController.java` | Base class with resource-resolution logic |
| `app/.../auth/AuthConfig.java` | Authorization configuration |
| `app/.../storage/impl/sql/` | SQL storage — search queries need filtering |

## Acceptance criteria

- [ ] Administrators can grant/deny read and write access to specific artifacts/groups for specific users
- [ ] Search and list endpoints respect per-resource permissions (users only see what they're allowed to)
- [ ] Pagination works correctly with filtered results
- [ ] A REST API exists for managing resource-level permissions
- [ ] Works across all storage variants (at minimum SQL; Kubernetes/GitOps via #7533)
- [ ] Backward-compatible — existing deployments without per-resource ACLs behave as today

## Related issues

- #7530 — EPIC: Kubernetes-Native Registry as a Service
- #7533 — Declarative authorization via Kubernetes resource configuration (covers user→role mappings in K8s/GitOps storage, complementary to this issue)
- #5125 — Authorization without authentication (gateway-fronted deployments)

carlesarnal:
## Summary

Per-resource authorization for Apicurio Registry, using a pure Java grants evaluator with a shared `Authorizer` interface. No external dependencies beyond Jackson.

### What it does

- **Two-layer authorization:** Keycloak (or any IdP) handles authentication + RBAC. Registry handles fine-grained per-resource authorization via a shared `authz` module.
- **Point-access checks:** `GrantsAuthorizer.authorize(subject, actions)` — pure Java grant matching (~10-50us per check). Only the current user's grants are evaluated (~5-20 entries) regardless of total grant count.
- **Search/list filtering:** Grants translated to SQL `WHERE groupId IN (...)` / Elasticsearch `terms` queries. Database handles filtering and pagination natively.
- **Shared grants file:** One JSON file defines permissions across systems. Hot-reloaded every 5 seconds on change, no restart needed.
- **Cross-system design:** The `authz` module defines its own `Authorizer` interface and is designed to be shared across any JVM application. Each system defines its own `ResourceType` enums.
- **Zero external dependencies:** The `authz` module depends only on `jackson-databind` and `slf4j-api`. No Kafka, no OPA, no WASM.

### Authorization flow

```
Authentication → Admin Override (bypass all) → RBAC → OBAC → Per-resource grants check
```

- **Admins** bypass the grants check entirely (via admin override or `config.admin_roles`)
- **Artifact owners** bypass the grants check (ownership implies access)
- **RBAC** is checked before grants — grants restrict within what RBAC allows, not instead of it
- **`authenticated-read-access`** overrides read grants (startup warning logged)
- **`trust-proxy-authorization`** bypasses all local authorization including grants

### Architecture

```
authz/ (apicurio-authz-core — shared, zero external deps)
  ├── Authorizer, Subject, Action, AuthorizeResult, Decision, ResourceType — interfaces
  ├── GrantsAuthorizer   — pure Java grants evaluation, hot-reload
  ├── GrantsData         — parsed grants, per-user filtering, validation, caching
  ├── Grant              — grant record with matching logic
  ├── User, RolePrincipal, Principal — identity types

app/ (Registry-specific integration)
  ├── GrantsAccessController     — bridges Authorizer with IAccessController, OTel metrics, audit logging
  ├── GrantsAccessControllerConfig — config properties
  ├── GrantsAccessControllerInitializer — startup, hot-reload scheduler, config conflict warnings
  ├── GrantsSearchFilter         — translates grants to SQL/ES filters
  ├── RegistryResourceType       — Artifact and Group enums
  ├── ISearchAuthorizer          — engine-agnostic search interface
  └── SearchAuthorizerProducer   — CDI producer, permissions query for UI
```

### Grants file format

```json
{
  "config": { "admin_roles": ["sr-admin"] },
  "grants": [
    {"principal": "alice", "operation": "write", "resource_type": "artifact",
     "resource_pattern_type": "prefix", "resource_pattern": "team-a/"},
    {"principal_role": "sr-developer", "operation": "read", "resource_type": "artifact",
     "resource_pattern_type": "prefix", "resource_pattern": "public/"}
  ]
}
```

### Observability

- **OTel metrics:** `apicurio.authz.decisions` counter (decision, resource_type, operation)
- **Audit logging:** denied decisions to `io.apicurio.registry.audit.authz`
- **Config conflict warnings:** logged at startup when grants + authenticated-read-access or anonymous-read-access are both enabled
- **Grants validation:** missing fields skipped with warning, unrecognized values logged

### Known limitations

- `getGroupById` uses `@Authorized(style=GroupAndArtifact)` with a single parameter — handled gracefully (#7866)
- RBAC is checked before grants — a `sr-readonly` user can never write regardless of grants
- The `permissions` field in search results (for UI show/hide) is defined in the OpenAPI spec but not yet populated (#7867)

Relates to #7724

## Test plan

- [x] 18 unit tests in `authz` module
- [x] 8 unit tests in `app` module
- [x] 34 integration tests with Keycloak (`integration-tests/`)
- [x] Docker-compose end-to-end example


jsenko:
I'll add ignore label to the existing PRs

app/src/main/java/io/apicurio/registry/auth/grants/GrantsAccessController.java
```diff
@@ -0,0 +1,200 @@
...
+                List.of(new Action(RegistryResourceType.Artifact.Read, resourceName)))
+                .toCompletableFuture().join();
+        return result.decision(RegistryResourceType.Artifact.Read, resourceName) == Decision.ALLOW;
+    }
+
+    private String getUsername() {
+        if (securityIdentity != null && !securityIdentity.isAnonymous()) {
+            return securityIdentity.getPrincipal().getName();
+        }
+        return "anonymous";
```
Joel-hanson: I’m not sure if this has already been addressed. However, I’d like to clarify that we are setting "anonymous" for the anonymous user, and getAllowedValues("anonymous", …) is called against the grants data. This means that if a grant is created with "principal": "anonymous", it may unintentionally match all unauthenticated users.

app/src/main/java/io/apicurio/registry/auth/grants/GrantsAccessControllerInitializer.java
```diff
@@ -0,0 +1,86 @@
...
+            log.warn("Per-resource authorization is enabled alongside anonymous-read-access. "
+                    + "Anonymous users will be able to read all resources regardless of grants.");
+        }
+        if (authConfig.isAuthenticatedReadsEnabled()) {
+            log.warn("Per-resource authorization is enabled alongside authenticated-read-access. "
+                    + "All authenticated users will be able to read all resources regardless of grants.");
+        }
+    }
+
+    @Scheduled(every = "5s", concurrentExecution = Scheduled.ConcurrentExecution.SKIP)
```
Joel-hanson: I assume this registers a scheduler that fires every 5 seconds, even when the feature is disabled. Should we add an option to configure the interval via a configuration property, so that if someone needs to disable this, they can do so by setting it to "off" or "disabled", similar to other properties?

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlGroupRepository.java
```diff
@@ -305,6 +305,32 @@ public GroupSearchResultsDto searchGroups(Set<SearchFilter> filters, OrderBy ord
                             query.bind(idx, filter.getStringValue());
                         });
                         break;
+                    case groupIdIn:
```
Joel-hanson, line 308: If possible, it would be great to have tests to cover these changes. Please ignore if its already done.

wasser-winfried:
Is there any progress ? 

This feature is eagerly awaited. Scaling via dedicated instances with 100+ customers is not really resource-saving.
We have customer-specific schemas or extensions and the current instance with its ALL readable approach is not really usable here. With this feature, we should be able to use a central instance that limits the visibility of groups or artifacts to customer-relevant entries.

carlesarnal:
> Is there any progress ?
> 
> This feature is eagerly awaited. Scaling via dedicated instances with 100+ customers is not really resource-saving. We have customer-specific schemas or extensions and the current instance with its ALL readable approach is not really usable here. With this feature, we should be able to use a central instance that limits the visibility of groups or artifacts to customer-relevant entries.

There is progress. We're working on a proposal and implementation for this.

authz/src/main/java/io/apicurio/authz/GrantsData.java
```diff
@@ -0,0 +1,196 @@
...
+                hasWildcard = true;
+                continue;
+            }
+
+            if (hasWildcard) {
+                continue;
+            }
+
+            if (groupSeparator != null) {
+                if ("prefix".equals(grant.resourcePatternType())) {
```
paoloantinori, line 161: Prefix grants scoped to a sub-path within a group over-grant in search. `extractGroupFromPattern` returns the substring before the first separator, so a `prefix: "team-a/secret/"` grant is reduced to group `team-a` and the SQL becomes `groupId IN ('team-a')`, returning every team-a artifact even though point-access only allows `team-a/secret/*` (via `startsWith`). This recreates the denied-direct-but-visible-in-search inconsistency the design rejected. Emit an exact/prefix artifact filter instead of collapsing to the top-level group, or exclude sub-group prefix patterns from the group IN clause.

app/src/main/java/io/apicurio/registry/auth/grants/GrantsSearchFilter.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+        return storage.searchVersions(augmented, orderBy, orderDir, offset, limit);
+    }
+
+    private Set<SearchFilter> addAuthorizationFilters(Set<SearchFilter> filters, String resourceType) {
+        if (!config.isEnabled() || grantsAc.getAuthorizer() == null) {
+            return filters;
+        }
+
+        GrantsData data = grantsAc.getGrantsData();
```
paoloantinori, line 73: When the authorizer is null this returns the filters unchanged, so search returns all results (fail open). Point-access in `GrantsAccessController.isAuthorized` denies in the same null state. With grants enabled but no grants path configured, `init` returns early without setting the authorizer, so search leaks everything while point-access locks everything. Fail closed here (return null / empty results) to match the point-access path.

app/src/main/java/io/apicurio/registry/auth/AuthorizedInterceptor.java
```diff
@@ -150,6 +158,17 @@ public Object authorizeMethod(InvocationContext context) throws Exception {
             }
         }
 
+        // If per-resource authorization is enabled, apply grants-based access checks.
+        // Artifact/group owners bypass the grants check — ownership implies access.
+        if (grantsAcConfig.isEnabled()) {
+            boolean isOwner = authConfig.ownerOnlyAuthorizationEnabled.get() && obac.checkOwnership(context);
```
paoloantinori, line 164: Owner bypass only runs when `ownerOnlyAuthorizationEnabled` (OBAC) is true, but the ADR and PR summary state that owners bypass grants unconditionally. With grants enabled and OBAC off, an artifact creator with no matching read grant is denied their own artifact. Either perform the ownership check regardless of the OBAC flag, or correct the documented contract.

app/src/main/java/io/apicurio/registry/auth/grants/GrantsAccessControllerInitializer.java
```diff
@@ -0,0 +1,87 @@
...
+            log.warn("Per-resource authorization is enabled alongside anonymous-read-access. "
+                    + "Anonymous users will be able to read all resources regardless of grants.");
+        }
+        if (authConfig.isAuthenticatedReadsEnabled()) {
+            log.warn("Per-resource authorization is enabled alongside authenticated-read-access. "
+                    + "All authenticated users will be able to read all resources regardless of grants.");
+        }
+    }
+
+    @Scheduled(every = "${apicurio.auth.resource-based-authorization.grants.reload-every:5s}",
```
paoloantinori, line 79: The `reload-every` property is passed straight into `@Scheduled(every=...)`, which requires a duration. The documented `off` value is not a valid duration and will likely fail schedule parsing at startup. Verify against the target Quarkus version, and gate the scheduled reload with a real enabled flag or a programmatic scheduler rather than relying on a sentinel string.

app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java
```diff
@@ -199,6 +199,16 @@ private Query buildFilterQuery(SearchFilter filter) {
                     : filter.getStringValue();
             return Query.of(q -> q.term(t -> t.field("groupId").value(groupValue)));
 
+        case groupIdIn:
```
paoloantinori, line 202: Only `groupIdIn` is handled here. `groupIdInOrArtifactExact`, `artifactExactDeny`, and `artifactPrefixDeny` fall through to the default case. Confirm the default branch fails closed for unknown filter types; if it ignores them, deny rules are silently dropped on Elasticsearch deployments, which is an authorization hole, not just a missing feature.

paoloantinori (commented):
Overall: the grants model is well structured and the ADR is thorough, but the search/list path has authorization correctness gaps that should be closed before merge. Point-access checks look sound; the risk is concentrated where grants are translated into SQL and Elasticsearch predicates.

- Search over-grants for prefix grants scoped within a group. `GrantsData.getSearchFilterData` (GrantsData.java:160) reduces any artifact prefix grant to its top-level group via `Grant.extractGroupFromPattern`, which returns the substring before the first `/`. A grant `prefix: "team-a/secret/"` becomes `groupId IN ('team-a')`, so search returns all team-a artifacts including ones point-access denies. This is the denied-direct-but-visible-in-search inconsistency the design rejected OPA for.
- Search fails open when the authorizer is not initialized. `GrantsSearchFilter.addAuthorizationFilters` (GrantsSearchFilter.java:73) returns filters unchanged when the authorizer is null, while point-access denies. Enabling grants without a path makes search leak everything and point-access lock everything. Fail closed.
- Owner bypass is gated on OBAC, contrary to the documented contract. `AuthorizedInterceptor` (AuthorizedInterceptor.java:164) only bypasses when `ownerOnlyAuthorizationEnabled`. With grants on and OBAC off, a creator can be denied their own artifact. Align the code with the ADR or correct the docs.
- Documented `reload-every: off` likely breaks startup. The value feeds `@Scheduled(every=...)` (GrantsAccessControllerInitializer.java:79), which needs a duration. Verify the Quarkus behavior, and use a real enabled flag instead of a sentinel string.
- Elasticsearch handles only `groupIdIn`. The other new filter types fall to the default case (ElasticsearchSearchService.java:202). Confirm that default fails closed; otherwise deny rules are silently dropped on ES deployments, which is an authorization hole.
- Simplification: the four new SQL filter branches are duplicated nearly verbatim between `searchArtifacts` and `searchVersions` in SqlSearchRepository.java (only variable names differ). Extract a shared builder so the security-relevant SQL lives in one place.

PR #7938, opened 2026-05-06
feat: Data Contracts]JSON Schema Tag Extractor

### LINKED ISSUE #7354: [Data Contracts] JSON Schema Tag Extractor
## Parent Epic
Part of #6192 (Phase 2: Metadata & Tags)

## Priority
P1 (High)

## Phase
Phase 2 - Tags & Metadata

## Blocked By
- #7352 Tag Extraction SPI

## Description

Implement tag extraction for JSON Schema, supporting extension properties for tags.

## Requirements

- [ ] Implement `JsonSchemaTagExtractor`:
  ```java
  @ApplicationScoped
  public class JsonSchemaTagExtractor implements TagExtractor {
      @Override
      public String getArtifactType() {
          return ArtifactType.JSON;
      }

      @Override
      public Map<String, Set<String>> extractTags(ContentHandle content) {
          // Parse JSON Schema
          // Extract tags from properties
          // Handle $ref resolution
      }
  }
  ```

- [ ] Support `x-tags` extension property:
  ```json
  {
    "type": "object",
    "properties": {
      "email": {
        "type": "string",
        "x-tags": ["PII", "EMAIL"]
      }
    }
  }
  ```

- [ ] Handle nested objects and arrays

- [ ] Handle `$ref` resolution for referenced schemas

## Location

`schema-util/json/src/main/java/io/apicurio/registry/json/`

## Acceptance Criteria

- Extracts tags from JSON Schema properties
- Handles nested structures
- Resolves $ref references
- Unit tests cover common patterns

matheusandre1:
Closes: #7354

### DIFF
--- a/app/src/main/java/io/apicurio/registry/contracts/tags/JsonSchemaTagExtractor.java
+++ b/app/src/main/java/io/apicurio/registry/contracts/tags/JsonSchemaTagExtractor.java
@@ -1,0 +1,208 @@
+/*
+ * Copyright 2025 Red Hat
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.apicurio.registry.contracts.tags;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.apicurio.registry.content.ContentHandle;
+import io.apicurio.registry.types.ArtifactType;
+import jakarta.enterprise.context.ApplicationScoped;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+@ApplicationScoped
+public class JsonSchemaTagExtractor implements TagExtractor {
+
+    private static final Logger log = LoggerFactory.getLogger(JsonSchemaTagExtractor.class);
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final String TAGS_PROPERTY = "x-tags";
+    private static final String REF_PROPERTY = "$ref";
+
+    @Override
+    public String getArtifactType() {
+        return ArtifactType.JSON;
+    }
+
+    @Override
+    public Map<String, Set<String>> extractTags(ContentHandle content) {
+        try {
+            JsonNode root = MAPPER.readTree(content.content());
+            Map<String, Set<String>> result = new LinkedHashMap<>();
+            visitSchema(root, "", root, result, new HashSet<>());
+            return result;
+        } catch (Exception e) {
+            log.debug("Failed to extract tags from JSON schema: {}", e.getMessage());
+            return Collections.emptyMap();
+        }
+    }
+
+    private static void visitSchema(JsonNode node, String path, JsonNode root,
+            Map<String, Set<String>> result, Set<String> refStack) {
+        if (node == null || node.isMissingNode() || node.isNull()) {
+            return;
+        }
+
+        if (node.isObject()) {
+            visitObjectSchema(node, path, root, result, refStack);
+        } else if (node.isArray()) {
+            visitArraySchema(node, path, root, result, refStack);
+        }
+    }
+
+    private static void visitObjectSchema(JsonNode node, String path, JsonNode root,
+            Map<String, Set<String>> result, Set<String> refStack) {
+        addTags(node.path(TAGS_PROPERTY), path, result);
+        visitReference(node.get(REF_PROPERTY), path, root, result, refStack);
+
+        visitProperties(node.path("properties"), path, root, result, refStack);
+        visitArrayItems(node.path("items"), path, root, result, refStack);
+        visitArrayItems(node.path("prefixItems"), path, root, result, refStack);
+        visitCompositionKeyword(node.path("allOf"), path, root, result, refStack);
+        visitCompositionKeyword(node.path("anyOf"), path, root, result, refStack);
+        visitCompositionKeyword(node.path("oneOf"), path, root, result, refStack);
+
+        visitSchema(node.path("not"), path, root, result, refStack);
+        visitSchema(node.path("if"), path, root, result, refStack);
+        visitSchema(node.path("then"), path, root, result, refStack);
+        visitSchema(node.path("else"), path, root, result, refStack);
+    }
+
+    private static void visitReference(JsonNode refNode, String path, JsonNode root,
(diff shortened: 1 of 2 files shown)


matheusandre1:
Rebasing is necessary, which defines the change in the CI, right? @jsenko

jsenko:
You can rebase, and comment `/ready` when it's ready for review, so full testsuite can run. We're still experimenting with this, sorry about confusion.

jsenko:
> @matheusandre1 Cannot mark as ready: no reviewer is assigned. Please ask a maintainer to assign a reviewer first.

hmm :/

jsenko:
ok after rebase, `/ready` should work now even when reviewer is not assigned.

schema-util/json/src/main/java/io/apicurio/registry/json/content/extract/JsonSchemaTagExtractorSupport.java
```diff
@@ -0,0 +1,174 @@
+package io.apicurio.registry.json.content.extract;
```
carlesarnal: Missing Apache 2.0 license header — required by `.checkstyle/java.header`. Same applies to the other two new files.

schema-util/json/src/main/java/io/apicurio/registry/json/content/extract/JsonSchemaTagExtractorSupport.java
```diff
@@ -0,0 +1,174 @@
...
+    private JsonSchemaTagExtractorSupport() {
+    }
+
+    public static Map<String, Set<String>> extractTags(ContentHandle content) {
+        try {
+            JsonNode root = MAPPER.readTree(content.content());
+            Map<String, Set<String>> result = new LinkedHashMap<>();
+            visitSchema(root, "", root, result, new HashSet<>());
+            return result;
+        } catch (Exception e) {
```
carlesarnal: Silent exception swallowing. The `AvroTagExtractor` logs at DEBUG level on parse failure (`log.debug("Failed to extract tags from Avro schema: {}", e.getMessage())`). Consider adding a logger here for consistency and debuggability.

carlesarnal (commented):
Code review — see inline comments for details. The core approach is solid; main concerns are around the `allOf`/`anyOf`/`oneOf` path generation bug and missing license headers.

schema-util/json/src/main/java/io/apicurio/registry/json/content/extract/JsonSchemaTagExtractorSupport.java
```diff
@@ -0,0 +1,174 @@
...
+        addTags(node.path(TAGS_PROPERTY), path, result);
+        visitReference(node.get(REF_PROPERTY), path, root, result, refStack);
+
+        visitProperties(node.path("properties"), path, root, result, refStack);
+        visitArrayKeyword(node.path("items"), path, root, result, refStack);
+        visitArrayKeyword(node.path("prefixItems"), path, root, result, refStack);
+        visitArrayKeyword(node.path("allOf"), path, root, result, refStack);
+        visitArrayKeyword(node.path("anyOf"), path, root, result, refStack);
+        visitArrayKeyword(node.path("oneOf"), path, root, result, refStack);
+
```
carlesarnal: Bug: `allOf`, `anyOf`, and `oneOf` are schema composition keywords, not array types in the data model. Routing them through `visitArrayKeyword` appends `[]` to the path, which is incorrect.

For example:
```json
{
  "allOf": [
    { "properties": { "name": { "x-tags": ["PII"] } } }
  ]
}
```
This would produce path `[].name` instead of the correct `name`.

These three should iterate their elements without modifying the path — something like:
```java
visitCompositionKeyword(node.path("allOf"), path, root, result, refStack);
```
where `visitCompositionKeyword` iterates but passes `path` through unchanged.

`items` and `prefixItems` are the only ones that should append `[]`.

schema-util/json/src/main/java/io/apicurio/registry/json/content/extract/JsonSchemaTagExtractorSupport.java
```diff
@@ -0,0 +1,174 @@
...
+        while (it.hasNext()) {
+            visitSchema(it.next(), path, root, result, refStack);
+        }
+    }
+
+    private static void visitProperties(JsonNode properties, String path, JsonNode root,
+            Map<String, Set<String>> result, Set<String> refStack) {
+        if (properties == null || !properties.isObject()) {
+            return;
+        }
```
carlesarnal: This method is doing double duty: it handles both array-in-data keywords (`items`, `prefixItems`) and composition keywords (`allOf`, `anyOf`, `oneOf`). These have different semantics:

- `items`/`prefixItems` → append `[]` to path (array in the data model)
- `allOf`/`anyOf`/`oneOf` → keep same path (schema combinator, not a data array)

Suggest splitting into two methods: one that appends `[]` (for `items`/`prefixItems`) and one that passes the path through unchanged (for composition keywords).

app/src/main/java/io/apicurio/registry/contracts/tags/JsonSchemaTagExtractor.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.apicurio.registry.content.ContentHandle;
+import io.apicurio.registry.json.content.extract.JsonSchemaTagExtractorSupport;
+import io.apicurio.registry.types.ArtifactType;
+import jakarta.enterprise.context.ApplicationScoped;
+
+import java.util.Map;
+import java.util.Set;
+
+@ApplicationScoped
```
carlesarnal, line 34: Architectural inconsistency: `AvroTagExtractor` has all extraction logic inline, while this splits it into a thin bean + `JsonSchemaTagExtractorSupport` in `schema-util/json`. There's no obvious reason for the split since `Support` is only called from here.

Suggestion: either inline the logic here (matching Avro's pattern) or refactor both extractors to use the same split pattern. Consistency helps future contributors.

schema-util/json/src/main/java/io/apicurio/registry/json/content/extract/JsonSchemaTagExtractorSupport.java
```diff
@@ -0,0 +1,174 @@
...
+            result.computeIfAbsent(path, p -> new LinkedHashSet<>()).addAll(tags);
+        }
+    }
+
+    private static JsonNode resolveReference(JsonNode root, String ref) {
+        if (ref == null || ref.isBlank()) {
+            return null;
+        }
+
+        String pointer = null;
```
carlesarnal: This handles external refs (e.g. `other.json#/defs/Foo`) by extracting the fragment and resolving it against the *current* document root, which will silently produce wrong results. Consider either:
- Only handling local refs (`#`-prefixed) and returning `null` for external refs, or
- Adding a brief comment explaining the limitation.

app/src/test/java/io/apicurio/registry/contracts/tags/JsonSchemaTagExtractorTest.java
```diff
@@ -0,0 +1,125 @@
...
+        assertEquals(Set.of("ADDRESS"), tags.get("billingAddress.street"));
+        assertEquals(Set.of("ADDRESS"), tags.get("shippingAddresses[].street"));
+    }
+
+    @Test
+    void testInvalidSchema() {
+        Map<String, Set<String>> tags = extractor.extractTags(ContentHandle.create("not valid json"));
+        assertTrue(tags.isEmpty());
+    }
+}
```
carlesarnal, line 593: Missing test coverage for:
- `allOf`/`anyOf`/`oneOf` composition keywords (this would catch the path bug mentioned in the other comments)
- Circular `$ref` (the `refStack` cycle detection is implemented but not tested)
- Multiple fields with tags at different nesting levels
- `prefixItems` keyword

PR #7950, opened 2026-05-07
fix(ccompat): resolve version references by integer sequence number

### LINKED ISSUE #7886: ccompat "/subjects/{subject}/versions/{version}" fails to resolve by integer sequence number when artifact uses semantic versioning
**Description**

This is a follow-up to #7694. 
The ccompat version field is an integer sequence number per spec, so the value returned in references from `GET /apis/ccompat/v7/schemas/ids/{id}` is correct. 

```
{
    "schema": "...",
    "schemaType": "PROTOBUF",
    "references": [
        {
            "name": "bq_table.proto",
            "subject": "bq_table.proto",
            "version": 1
        },
        {
            "name": "Meta.proto",
            "subject": "Meta.proto",
            "version": 1
        }
    ]
}
```

However, when a consumer uses that integer sequence number to resolve a referenced artifact that was registered with a semantic version (e.g., v5.1.1), the lookup fails with error.

**Steps to Reproduce**
  1. Register a Protobuf schema that imports other artifacts.
  2. Register the referenced artifacts using semantic versioning (e.g., v5.1.1).
  3. Call GET /apis/ccompat/v7/schemas/ids/{id} — this returns  integer sequence number(i.e. 1) for the references.
  4. Call GET /apis/ccompat/v7/subjects/bq_table.proto/versions/1 to resolve the reference.

**Actual Behavior**

Error: No version '1' found for artifact with ID 'bq_table.proto' in group 'null'.; error code: 40402

**Expected Behavior**

I am not sure if correct, but one possible statement is:

`/apis/ccompat/v7/subjects/{subject}/versions/{integer} `should be able to resolve an artifact by its integer sequence number regardless of whether the artifact was registered with a semantic version string. The integer sequence number 1 returned in the references response should always be resolvable via the ccompat subjects endpoint.

**Environment**
  - Apicurio Registry: 3.2.3
  - Schema type: PROTOBUF
  - Integration: KafbatUI with Apicurio schema registry via ccompat API

dejan-kosak:
## Summary
Fixes #7886 — ccompat v7 returns versions as integer sequence numbers (`versionOrder`), but lookups used the integer as the literal version string. This broke artifacts stored with semantic versions (e.g. `"v5.1.1"`): the ID returned to clients couldn't be used to fetch, delete, or reference the version.

  ## Root Cause
`AbstractResource#parseVersionString` and the reference parsers passed the numeric string straight through as the version ID. That works for ccompat-registered artifacts (numeric IDs match the sequence number) but fails when the stored version is non-numeric. `SubjectsResourceImpl` also used `VersionUtil.toLong(version)` in several response paths, which throws on non-numeric versions.

## Changes
- `AbstractResource`: new `resolveVersionBySequenceNumber(...)` — direct version-string match first, then fallback scan by `versionOrder`. Wired into `parseVersionString` (numeric branch) and both reference parsers. Javadoc updated.
- `SubjectsResourceImpl`: `getSubjectVersions`, `deleteSchemaVersion`, `deleteSubjectPermanent`, `deleteSubjectVersions` now return `versionOrder` instead of parsing the version string. `deleteSubjectPermanent` collects versionOrders before `deleteArtifact` (metadata is gone after). Removed unused `VersionUtil` import. Some incidental formatting changes.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/AbstractResource.java
+++ b/app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/AbstractResource.java
@@ -271,7 +271,9 @@ public abstract class AbstractResource {
             final List<ArtifactReferenceDto> referencesAsDtos = references.stream().map(schemaReference -> {
                 final ArtifactReferenceDto artifactReferenceDto = new ArtifactReferenceDto();
                 artifactReferenceDto.setArtifactId(schemaReference.getSubject());
-                artifactReferenceDto.setVersion(String.valueOf(schemaReference.getVersion()));
+                String versionString = String.valueOf(schemaReference.getVersion());
+                artifactReferenceDto.setVersion(resolveVersionBySequenceNumber(null,
+                        schemaReference.getSubject(), schemaReference.getVersion(), versionString));
                 artifactReferenceDto.setName(schemaReference.getName());
                 artifactReferenceDto.setGroupId(null);
                 return artifactReferenceDto;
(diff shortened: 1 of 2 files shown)


EricWittmann:
This was a good start and I used it as inspiration for a more thorough solution to this issue.  Thanks for submitting this!

* https://github.com/Apicurio/apicurio-registry/pull/7956

PR #7973, opened 2026-05-08
Add search capabilities to the CLI. Fixes #7786

### LINKED ISSUE #7786: Implement search capabilities in the CLI
Part of Epic #7162 - Phase 3

## Description

Implement search capabilities in the CLI.

**API Endpoint:** `/search` (see Search operations in API docs)

Search by content can be skipped in this phase.

## Acceptance Criteria

- Search for groups
- Search for artifacts
- Search for versions
- Support filtering and pagination
- JSON and table output formats

## Related

Epic: #7162

### LINKED ISSUE #8005: CLI: exception when searching with label :foo
## Description

Running `./acr search groups -l :foo` throws an unhandled exception instead of returning an error message or empty results.

## Steps to reproduce

```bash
./acr search groups -l :foo
```

## Expected behavior

The CLI should either return an appropriate error message or handle the malformed label filter gracefully.

## Found during

PR #7973 — CLI search command implementation.

vandanayadav7:
## Summary

Add search commands to the Apicurio Registry CLI for searching groups, artifacts, and versions with filtering, pagination, sorting, and JSON/table output.

### Search commands
- Add `SearchCommand` (parent, shows help), `SearchGroupsCommand`, `SearchArtifactsCommand`, `SearchVersionsCommand` in the `search` package
- Register `SearchCommand` as top-level subcommand in `Acr.java`
- Support both singular and plural subcommand names (e.g., `acr search group` / `acr search groups`)
- Remove `-d`, `-n`, `-t` short options from search commands to use short options sparingly
- Update search field descriptions to specify matching behavior (exact, substring, wildcard)

### Order mixins
- Extract `ArtifactOrderMixin`, `GroupOrderMixin`, `VersionOrderMixin` for reusable `--order-by` and `--order` options
- Replace hardcoded sort values in `ArtifactCommand`, `GroupCommand`, `VersionCommand` with the shared mixins

### Label handling
- Use `key=value` format for labels across all CLI commands (search, create, update), converting to `key:value` internally for the API
- Support `\=` escaping in label keys (e.g., `key\=name=value` creates label `key=name` with value `value`)
- Validate that label keys don't contain colons (reserved by the REST API) with a clean error message
- Change `--set-label` / `--label` from `Map<String, String>` to `List<String>` with manual parsing in all create and update commands

### Bug fixes & cleanup
- Fix potential NPE in `GroupUpdateCommand` when group has no labels
- Add missing `@Parameters` descriptions in `GroupCreateCommand` and `GroupUpdateCommand`
- Fix import ordering in artifact and version command files
- Remove unnecessary `@SuppressWarnings("unchecked")` from `GroupCreateCommand`
- Fix `CommentCommand` to extend `AbstractCommand` instead of `Runnable` for consistency
- Make `AbstractCommand.spec` protected for cross-package access

## Usage

```bash
acr search groups                                           # list all groups
acr search groups -g myGroup -o json                        # filter by group ID
acr search groups -l "env=prod" -l "team=platform"          # filter by labels
acr search groups --order-by CreatedOn --order Desc         # sort by creation date

acr search artifacts -g myGroup --type JSON                 # filter by group and type
acr search artifacts --name "my-schema" -o json             # filter by name
acr search artifacts --global-id 123                        # filter by global ID

acr search versions -g myGroup -a myArtifact                # filter by group and artifact
acr search versions --state ENABLED -v 1.0.0                # filter by state and version
acr search versions --order-by CreatedOn --order Desc -p 1 -s 5  # sort and paginate
```

## Filters Supported

| Filter | Groups | Artifacts | Versions |
|--------|--------|-----------|----------|
| `-g` / `--group` | ✓ | ✓ | ✓ |
| `-a` / `--artifact` | — | ✓ | ✓ |
| `--name` | — | ✓ | ✓ |
| `--description` | ✓ | ✓ | ✓ |
| `--type` | — | ✓ | ✓ |
| `-l` / `--label` | ✓ | ✓ | ✓ |
| `-v` / `--version` | — | — | ✓ |
| `--state` | — | — | ✓ |
| `--global-id` | — | ✓ | ✓ |
| `--content-id` | — | ✓ | ✓ |
| `--order-by` | ✓ | ✓ | ✓ |
| `--order` | ✓ | ✓ | ✓ |
| `-p` / `--page` | ✓ | ✓ | ✓ |
| `-s` / `--size` | ✓ | ✓ | ✓ |

## Related Issues

- Closes #7786
- Fixes #8005 (CLI exception when searching with label `:foo`)
- #8002 (Registry search API: inconsistent matching behavior across name fields)

## Testing

- [x] Build succeeds
- [x] Checkstyle passes with 0 violations
- [x] All CLI tests pass (193 tests, 0 failures)
- [x] Search groups — all, filtered, no results, table, pagination, order-by, invalid inputs
- [x] Search artifacts — all, filtered, no results, table, type filter, order-by, label filter, invalid label
- [x] Search versions — all, filtered, no results, table, state filter, version filter, order-by, pagination, invalid state

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/search/SearchArtifactsCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/search/SearchArtifactsCommand.java
@@ -1,0 +1,180 @@
+package io.apicurio.registry.cli.search;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import io.apicurio.registry.cli.common.AbstractCommand;
+import io.apicurio.registry.cli.common.ArtifactOrderMixin;
+import io.apicurio.registry.cli.common.OutputTypeMixin;
+import io.apicurio.registry.cli.common.PaginationMixin;
+import io.apicurio.registry.cli.utils.Conversions;
+import io.apicurio.registry.cli.utils.OutputBuffer;
+import io.apicurio.registry.cli.utils.TableBuilder;
+import io.apicurio.registry.rest.client.models.ProblemDetails;
+import io.apicurio.registry.rest.client.search.artifacts.ArtifactsRequestBuilder;
+import io.apicurio.registry.rest.v3.beans.ArtifactSearchResults;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Mixin;
+import picocli.CommandLine.Option;
+
+import java.util.List;
+
+import static io.apicurio.registry.cli.common.CliException.exitQuietServerError;
+import static io.apicurio.registry.cli.utils.Columns.ARTIFACT_ID;
+import static io.apicurio.registry.cli.utils.Columns.ARTIFACT_TYPE;
+import static io.apicurio.registry.cli.utils.Columns.CREATED_ON;
+import static io.apicurio.registry.cli.utils.Columns.DESCRIPTION;
+import static io.apicurio.registry.cli.utils.Columns.GROUP_ID;
+import static io.apicurio.registry.cli.utils.Columns.LABELS;
+import static io.apicurio.registry.cli.utils.Columns.MODIFIED_BY;
+import static io.apicurio.registry.cli.utils.Columns.MODIFIED_ON;
+import static io.apicurio.registry.cli.utils.Columns.NAME;
+import static io.apicurio.registry.cli.utils.Columns.OWNER;
+import static io.apicurio.registry.cli.utils.Conversions.convert;
+import static io.apicurio.registry.cli.utils.Conversions.convertToString;
+import static io.apicurio.registry.cli.utils.Mapper.MAPPER;
+
+@Command(
+        name = "artifact",
+        aliases = {"artifacts"},
+        description = "Search for artifacts"
+)
+public class SearchArtifactsCommand extends AbstractCommand {
+
+    @Option(
+            names = {"--name"},
+            description = "Filter by artifact name. Searches both the name and artifactId fields. Use * as prefix/suffix wildcard, otherwise matches exactly."
+    )
+    private String name;
+
+    @Option(
+            names = {"--description"},
+            description = "Filter by description (substring match)."
+    )
+    private String description;
+
+    @Option(
+            names = {"-g", "--group"},
+            description = "Filter by group ID (exact match)."
+    )
+    private String groupId;
+
+    @Option(
+            names = {"-a", "--artifact"},
+            description = "Filter by artifact ID (exact match)."
+    )
+    private String artifactId;
+
+    @Option(
+            names = {"--type"},
+            description = "Filter by artifact type, exact match (e.g. AVRO, JSON, PROTOBUF, OPENAPI, ASYNCAPI). Use 'acr version' to see all supported types."
+    )
+    private String artifactType;
+
+    @Option(
+            names = {"-l", "--label"},
+            description = "Filter by label (format: key=value or key). Exact match on key and value. Can be specified multiple times."
+    )
+    private List<String> labels;
+
+    @Option(
+            names = {"--global-id"},
+            description = "Filter by global ID"
+    )
+    private Long globalId;
+
+    @Option(
+            names = {"--content-id"},
+            description = "Filter by content ID"
+    )
+    private Long contentId;
+
+    @Mixin
+    private ArtifactOrderMixin ordering;
+
+    @Mixin
+    private PaginationMixin pagination;
+
+    @Mixin
+    private OutputTypeMixin outputType;
+
+    @Override
+    public void run(final OutputBuffer output) throws JsonProcessingException {
+        try {
+            //noinspection ConstantConditions
+            final var results = convert(client.getRegistryClient().search().artifacts().get(r -> {
+                //noinspection ConstantConditions
+                applyFilters(r.queryParameters);
+            }));
+            printResults(output, results);
+        } catch (final ProblemDetails ex) {
+            output.writeStdErrChunk(err -> {
+                err.append("Error searching for artifacts: ")
+                        .append(ex.getDetail())
+                        .append('\n');
+            });
+            exitQuietServerError();
+        }
+    }
+
+    private void applyFilters(final ArtifactsRequestBuilder.GetQueryParameters params) {
+        params.offset = (pagination.getPage() - 1) * pagination.getSize();
+        params.limit = pagination.getSize();
+        params.orderby = ordering.getOrderBy();
+        params.order = ordering.getOrder();
+        if (name != null) {
+            params.name = name;
+        }
+        if (description != null) {
+            params.description = description;
+        }
+        if (groupId != null) {
+            params.groupId = groupId;
+        }
+        if (artifactId != null) {
+            params.artifactId = artifactId;
+        }
+        if (artifactType != null) {
+            params.artifactType = artifactType;
+        }
+        if (labels != null) {
+            params.labels = Conversions.convertLabelsForApi(labels);
+        }
+        if (globalId != null) {
+            params.globalId = globalId;
+        }
+        if (contentId != null) {
+            params.contentId = contentId;
+        }
+    }
+
+    private void printResults(final OutputBuffer output, final ArtifactSearchResults results) throws JsonProcessingException {
+        output.writeStdOutChunkWithException(out -> {
+            switch (outputType.getOutputType()) {
+                case json -> {
+                    out.append(MAPPER.writeValueAsString(results));
+                    out.append('\n');
+                }
+                case table -> {
+                    final var table = new TableBuilder();
+                    table.addColumns(GROUP_ID, ARTIFACT_ID, NAME, ARTIFACT_TYPE, DESCRIPTION,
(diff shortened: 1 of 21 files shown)


cli/src/main/java/io/apicurio/registry/cli/search/SearchVersionsCommand.java
```diff
@@ -0,0 +1,211 @@
...
+            names = {"-v", "--version"},
+            description = "Filter by version expression"
+    )
+    private String version;
+
+    @Option(
+            names = {"-t", "--type"},
+            description = "Filter by artifact type (e.g. AVRO, JSON, PROTOBUF, OPENAPI, ASYNCAPI). Use 'acr version' to see all supported types."
+    )
+    private String artifactType;
```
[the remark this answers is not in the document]
vandanayadav7: The table already shows ✓ for both -n/--name and -t/--type under the Versions column. No change needed.

cli/src/main/java/io/apicurio/registry/cli/search/SearchArtifactsCommand.java
```diff
@@ -0,0 +1,190 @@
...
+            defaultValue = "ArtifactId"
+    )
+    private ArtifactSortBy orderBy;
+
+    @Option(
+            names = {"--order"},
+            description = "Sort order (Asc, Desc). Default: Asc.",
+            defaultValue = "Asc"
+    )
+    private SortOrder order;
```
jsenko: When listing artifacts, we currently hardcode the order  and orderBy fields. Since we use these options in two places now, it might be worth it to create a mixin and use it in both places. Same for groups and versions.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Option(
+            names = {"-g", "--group"},
+            description = "Filter by group ID"
+    )
+    private String groupId;
+
+    @Option(
+            names = {"-d", "--description"},
+            description = "Filter by description"
```
jsenko: We should update descriptions for search fields and specify whether:
- a result must be an exact match
- or result must contain the value as a sub-string.
- wildcards are supported

This is what I found out we currently do:
- `description` matches as a sub-string
- artifact  `name` searches both the `name` field and `artifactId` fiels
- user must use `*` wildcard as a prefix and/or suffix for artifact field `name` otherwise it searches exactly
-  this is different for `name` field in versions, which searches as a substring, and treats `*` literally
- labels are matched exactly, but you  can either match keys only `key` or `key:` or both key and value `key:value`
- other fields are matched exactly

I also found an exception is thrown if you run `./acr search groups -l :foo`

What we should do:
- Update the descriptions
- Create a Registry search issue to fix the inconsistencies
- Create a Registry CLI issue (or update the epic) to fix the exception
-  Add a parameter `--artifact` that does the same as `--name` to reduce  confusion when searching for artifacts

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -0,0 +1,138 @@
...
+public class SearchGroupsCommand extends AbstractCommand {
+
+    @Option(
+            names = {"-g", "--group"},
+            description = "Filter by group ID"
+    )
+    private String groupId;
+
+    @Option(
+            names = {"-d", "--description"},
```
jsenko: I think we should use short options sparingly. This should be done in a separate PR, but I'd remove `-d`, `-n` and `-t` short options and review other commands.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -0,0 +1,138 @@
...
+import static io.apicurio.registry.cli.utils.Columns.LABELS;
+import static io.apicurio.registry.cli.utils.Columns.MODIFIED_BY;
+import static io.apicurio.registry.cli.utils.Columns.MODIFIED_ON;
+import static io.apicurio.registry.cli.utils.Columns.OWNER;
+import static io.apicurio.registry.cli.utils.Conversions.convert;
+import static io.apicurio.registry.cli.utils.Conversions.convertToString;
+import static io.apicurio.registry.cli.utils.Mapper.MAPPER;
+
+@Command(
+        name = "groups",
```
jsenko: In  e.g. `acr create` we support both singular and plural aliases

cli/src/main/java/io/apicurio/registry/cli/search/SearchArtifactsCommand.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @Option(
+            names = {"-t", "--type"},
+            description = "Filter by artifact type (e.g. AVRO, JSON, PROTOBUF, OPENAPI, ASYNCAPI). Use 'acr version' to see all supported types."
+    )
+    private String artifactType;
+
+    @Option(
+            names = {"-l", "--label"},
+            description = "Filter by label (format: key:value). Can be specified multiple times."
```
jsenko: I think we should follow the `key=value` format we use in previous commands and convert internally to avoid surprising users. Internally Registry supports all of these labels:

`key:key`=`value:value`
`key=key`=`value=value`
`key=key:`=`value=:value`

So the limitation is REST API treating colons in a special way, so we need to check that before conversion.

We should also gracefully support equals sign in keys and values. Currently this command:
```
./acr artifacts version update -g test -a test --set-label key=key=value  1
```
Creates label `key`=`value=value`. This  means that it only looks for the first `=`. We should support some for of escaping, such as:
```
./acr artifacts version update -g test -a test --set-label key\=key=value  1
```
Will result in `key=value`=`value`.

cli/src/test/java/io/apicurio/registry/cli/SearchCommandTest.java
```diff
@@ -0,0 +1,299 @@
...
+        Files.writeString(tempFile, "{\"type\": \"string\"}");
+        try {
+            executeAndAssertSuccess("artifact", "create", "-g", TEST_GROUP,
+                    "--type", "JSON", "--file", tempFile.toString(), TEST_ARTIFACT);
+        } finally {
+            Files.deleteIfExists(tempFile);
+        }
+    }
+
+    @Test
```
jsenko, line 41: Some tests are missing `@Order` annotations. While they will run last, I think once we specify an order on some tests, we should do that for all methods and keep the code in the same order.

jsenko (changes requested):
Nice work! I've added some comments, but I have an architectural question about search:

We would now have two endpoints for listing data:
 
- `acr artifacts` uses `/groups/{groupId}/artifacts`
- `acr search  artifacts` uses `/search/artifacts`

While it follows the API structure, it seems like unnecessary duplication for users. We could have a single command that calls different endpoints under the hood based on parameters, or use search endpoints only. WDYT @EricWittmann How is this resolved in UI? Any performance considerations?

I realize I should've thought about this earlier, but we can still merge this and refactor as a followup.

EricWittmann:
> Nice work! I've added some comments, but I have an architectural question about search:
> 
> We would now have two endpoints for listing data:
> 
>     * `acr artifacts` uses `/groups/{groupId}/artifacts`
> 
>     * `acr search  artifacts` uses `/search/artifacts`
> 
> 
> While it follows the API structure, it seems like unnecessary duplication for users. We could have a single command that calls different endpoints under the hood based on parameters, or use search endpoints only. WDYT @EricWittmann How is this resolved in UI? Any performance considerations?
> 
> I realize I should've thought about this earlier, but we can still merge this and refactor as a followup.

I honestly don't see this duplication as much of a problem.  You can do all sorts of **other** operations under the `acr artifacts` command presumably.  But it's worth noting that when Elasticsearch is enabled, they will actually do different things (`acr artifacts` listings will come from the DB but `acr search` will use ES).

Even without that, it's just two ways of thinking about the data in the registry, and I think that's OK.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
vandanayadav7: Done, except for the `--artifact` alias — see below:

- **Descriptions updated** — each search field now specifies matching behavior (exact, substring, wildcard).
- **Search inconsistency issue** — created #8002.
- **`:foo` exception issue** — created #8005, also fixed in this PR by validating colons in label keys and showing a clean error.
- **Label format** — switched to `key=value` format with `\=` escaping support across all commands (search, create, update).
- **`--artifact` alias for `--name`** — `-a`/`--artifact` is already used for exact artifact ID filtering across multiple commands (`VersionCommand`, `VersionUpdateCommand`, etc.). Reusing it as an alias for `--name` in search would change its meaning. Open to suggestions on how to handle this.

jsenko: > Reusing it as an alias for --name in search would change its meaning. Open to suggestions on how to handle this.

I don't think it's necessary a problem, but we can defer this so the PR can be merged. Can you make a note in the epic?

jsenko (approved):
LGTM

PR #7998, opened 2026-05-12
feat(operator): expose management port 9000 on the app Service

kamolhasan:
## Summary

The operator-generated Kubernetes Service for the registry app component only exposes the HTTP port (8080, or 443/8443 when TLS is configured). However, the app container also declares a **management port (9000)** used for health endpoints (`/health/live`, `/health/ready`) and metrics. This port is already present on the container spec, health probes, and NetworkPolicy, but was missing from the Service.

Without the management port on the Service, any infrastructure that performs health checks or routing through the Service's port list cannot reach the management endpoints. This affects:

- **Ingress controllers** (e.g. AWS ALB, GCP NEG) that validate health check target ports against the Service's `spec.ports`
- **Service mesh sidecars** that discover upstream ports from the Service definition
- **Monitoring tools** that scrape metrics by enumerating Service ports

This PR adds the management port (9000) to the app Service across all configurations:

- **Non-TLS (default)**: Added to the `app.service.yaml` template with named `targetPort: management` matching the container port
- **TLS / TLS+insecure**: Added to the programmatic port list in `AppServiceResource` that replaces template ports when TLS is configured

### Files changed

- `operator/controller/src/main/resources/k8s/default/app.service.yaml` -- added management port entry
- `operator/controller/src/main/java/.../AppServiceResource.java` -- included management port in TLS code path
- `operator/controller/src/test/java/.../TlsITTest.java` -- updated port count and port number assertions

### Not affected

- **UI Service**: The UI container has no management port (health checks use `/config.js` on port 8080), so no change is needed
- **CRD/model**: Port 9000 is a fixed management port, not user-configurable, so no CRD changes are required
- **Reconciliation lifecycle**: The Service is a leaf dependency (only the Ingress depends on it, and it uses the Service name, not port list). Adding a static port does not introduce reconcile loops

## Test plan

- [ ] Verify `TlsITTest.testPassthroughTLS()` passes with updated assertion (2 ports: 443, 9000)
- [ ] Verify `TlsITTest.testPassthroughTLSWithInsecureTraffic()` passes with updated assertion (3 ports: 443, 8080, 9000)
- [ ] Verify `SmokeITTest` passes (non-TLS path, service template now includes management port)
- [ ] Manually confirm generated Service YAML includes `management` port on 9000

### DIFF
--- a/operator/controller/src/main/java/io/apicurio/registry/operator/resource/app/AppServiceResource.java
+++ b/operator/controller/src/main/java/io/apicurio/registry/operator/resource/app/AppServiceResource.java
@@ -48,10 +48,16 @@ public class AppServiceResource extends CRUDKubernetesDependentResource<Service,
                             .withTargetPort(new IntOrStringBuilder().withValue(8443).build())
                             .build();
 
+                    var managementPort = new ServicePortBuilder()
+                            .withName("management")
+                            .withPort(9000)
+                            .withTargetPort(new IntOrStringBuilder().withValue(9000).build())
+                            .build();
+
                     if (TLS.insecureRequestsEnabled(tls)) {
-                        s.getSpec().setPorts(List.of(httpsPort, httpPort));
+                        s.getSpec().setPorts(List.of(httpsPort, httpPort, managementPort));
                     } else {
-                        s.getSpec().setPorts(List.of(httpsPort));
+                        s.getSpec().setPorts(List.of(httpsPort, managementPort));
                     }
                 });
 
--- a/operator/controller/src/main/resources/k8s/default/app.service.yaml
+++ b/operator/controller/src/main/resources/k8s/default/app.service.yaml
@@ -7,4 +7,8 @@ spec:
       port: 8080
       protocol: TCP
       targetPort: http
+    - name: management
+      port: 9000
+      protocol: TCP
+      targetPort: management
   selector: { }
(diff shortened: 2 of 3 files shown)


carlesarnal:
Thanks for the contribution! Before we move forward, could you share the specific scenario or infrastructure setup where you needed the management port (9000) exposed on the Service?
  
In standard Kubernetes deployments, the management endpoints should already be reachable without this change:

- Health probes — the kubelet communicates directly with the pod IP, not through the Service. The operator already configures liveness/readiness probes on the container spec targeting port 9000.
- Metrics scraping — Prometheus typically discovers targets via pod annotations or ServiceMonitor CRDs, which reference pod ports directly.
- NetworkPolicy — the operator already includes a rule allowing ingress traffic to port 9000 at the pod level.

We'd like to understand if this is solving an actual problem you hit (e.g., a specific ingress controller or service mesh that requires the port to be listed on the Service), or if it's a preventive change. If there's a concrete use case, we're happy to consider it — but without one, we'd prefer to keep the Service definition minimal to reduce maintenance surface.

kamolhasan:
Thanks for the context @carlesarnal, yes it is a preventive change.

We are using ABL ingress:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: apicurio-registry-app
  annotations:
    alb.ingress.kubernetes.io/healthcheck-path: /health/live
    alb.ingress.kubernetes.io/healthcheck-port: "9000"
spec:
  ingressClassName: alb
  rules:
    - host: apicurio-api.sandbox.earnin.net
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: apicurio-registry-app-service
                port:
                  number: 8080
```

My understanding was, as the requests are getting routed by the service (`apicurio-registry-app-service`), it should contains the  port `9000`. But after reading your comment, if ABL is actually communicating using POD's IPs, then this PR can be closed.

Issue #7150, opened 2026-01-16
Improvements to OpenTelemetry implementation


carlesarnal:
## Summary

Following the fix for OTEL build-time vs runtime configuration (#7147), there are several improvements that can be made to the OpenTelemetry implementation to enhance observability and align with best practices.

### 🔴 Critical: OTelMetricsProvider is Dead Code

`OTelMetricsProvider` defines 6 custom OTEL metrics but is **never used anywhere** in the codebase:

- `recordArtifactCreated()`
- `recordArtifactDeleted()`
- `recordVersionCreated()`
- `recordSchemaValidation()`
- `recordRuleEvaluation()`
- `recordSearchRequest()`

**Recommendation**: Either remove the class if Micrometer metrics are sufficient, or wire it up by injecting it into the appropriate services.

### 🟡 Medium: StorageTracingInterceptor Missing @Priority

`StorageTracingInterceptor` has no `@Priority` annotation. When multiple interceptors use the same binding (`@StorageMetricsApply`), execution order is undefined.

**Recommendation**: Add explicit priority to ensure tracing wraps the entire operation:

```java
@Interceptor
@StorageMetricsApply
@Priority(Interceptor.Priority.PLATFORM_BEFORE)
public class StorageTracingInterceptor {
```

### 🟡 Medium: Missing Instrumentation Version

`StorageTracingInterceptor` doesn't specify an instrumentation version when getting the tracer.

**Current**:
```java
Tracer tracer = GlobalOpenTelemetry.getTracer(INSTRUMENTATION_NAME);
```

**Recommended**:
```java
Tracer tracer = GlobalOpenTelemetry.getTracer(INSTRUMENTATION_NAME, INSTRUMENTATION_VERSION);
```

### 🟡 Medium: TracingFilter Attribute Naming

Custom attribute names could better follow OpenTelemetry naming conventions (using underscores):

- `apicurio.groupId` → `apicurio.registry.group_id`
- `apicurio.artifactId` → `apicurio.registry.artifact_id`

### 🟢 Low: Config Documentation Mismatch

`OpenTelemetryConfig.java` documents `quarkus.otel.traces.sampler.arg` with default `1.0`, but `application.properties` uses `0.1`.

### 🟢 Low: Consider Response Filter

`TracingFilter` is request-only. A response filter could add error codes and other response-specific attributes to spans.

## Proposed Tasks

- [ ] Remove `OTelMetricsProvider` or wire it up to services
[...]

PR #8061, opened 2026-05-21
feat(otel): enable JDBC telemetry for SQL query tracing

paoloantinori:
## Summary

- Enable JDBC telemetry for SQL query tracing via OpenTelemetry, resolved per-named-datasource through `DatabaseConfigInterceptor`

## Related Issues

- Part of the OTel improvements discussed in #7150

## Changes

- `app/src/main/java/io/apicurio/registry/services/DatabaseConfigInterceptor.java`: extended to resolve `quarkus.datasource.<name>.jdbc.telemetry` alongside `.active`, setting `true` only for the active datasource matching `apicurio.storage.sql.kind`
- `app/src/test/java/io/apicurio/registry/services/DatabaseConfigInterceptorTest.java`: 18 unit tests covering active, telemetry, precedence, and passthrough behavior for all database kinds
- `app/src/main/resources/application.properties`: removed global `quarkus.datasource.jdbc.telemetry=true` (now handled by the interceptor)

## Design

Rather than enabling telemetry globally on the default datasource, the interceptor mirrors the existing `.active` property pattern: it sets `jdbc.telemetry=true` only for the named datasource that matches the configured `apicurio.storage.sql.kind`. This avoids enabling telemetry for unused datasources and respects any explicitly-set values from higher-priority config sources.

## Testing

- 18 unit tests pass via `./mvnw test -pl app -Dtest=DatabaseConfigInterceptorTest`
- Tests cover: all 4 database kinds, active+telemetry agreement, existing value precedence, passthrough, null storage kind
- Span volume controlled by existing `parentbased_traceidratio` sampler at 0.1 (10%)
- JDBC telemetry only generates spans when `quarkus.otel.sdk.disabled=false` (disabled by default)

### DIFF
--- a/app/src/main/resources/application.properties
+++ b/app/src/main/resources/application.properties
@@ -77,7 +77,6 @@ quarkus.otel.resource.attributes=service.version=${apicurio.app.version:unknown}
 
 # Kafka tracing propagation
 quarkus.otel.instrument.kafka=true
-quarkus.datasource.jdbc.telemetry=true
 
 
 # Package
--- a/app/src/main/resources/application.properties
+++ b/app/src/main/resources/application.properties
@@ -77,6 +77,7 @@ quarkus.otel.resource.attributes=service.version=${apicurio.app.version:unknown}
 
 # Kafka tracing propagation
 quarkus.otel.instrument.kafka=true
+quarkus.datasource.jdbc.telemetry=true
 
 
 # Package
--- a/app/src/main/java/io/apicurio/registry/services/DatabaseConfigInterceptor.java
+++ b/app/src/main/java/io/apicurio/registry/services/DatabaseConfigInterceptor.java
@@ -7,10 +7,10 @@ import io.smallrye.config.ConfigValue;
 import jakarta.annotation.Priority;
 
 /**
- * Automatically activates the correct Quarkus named datasource based on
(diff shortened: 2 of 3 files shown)


app/src/main/resources/application.properties
carlesarnal, line 242: This PR does not correctly address the problem. Telemetry is enabled at the datasource level, and we have one for each database type. We would need to either activate telemetry for all of them or dynamically discover which one is active in the DatabaseConfigInterceptor an activate telemetry for that one.

carlesarnal (changes requested):

paoloantinori:
@carlesarnal Thanks for the review! You're right — telemetry should be per-named-datasource, not global.

Updated approach in `0adf95f67`:

- Removed the global `quarkus.datasource.jdbc.telemetry=true` from `application.properties`
- Extended `DatabaseConfigInterceptor` to resolve `quarkus.datasource.<name>.jdbc.telemetry` the same way it already resolves `.active` — setting `true` only for the datasource matching `apicurio.storage.sql.kind`
- Added 18 unit tests covering all database kinds, active+telemetry agreement, existing value precedence, and passthrough behavior

This way telemetry is only enabled for the active datasource, and any explicitly-set values from higher-priority config sources are still respected.

/ready

carlesarnal (approved):

PR #8066, opened 2026-05-21
docs: restore config reference partial include and fix template for vale compliance

eameh:


docs/modules/ROOT/pages/getting-started/assembly-config-reference.adoc
```diff
@@ -17,1664 +17,8 @@ You can use the following reference information to configure {registry}.
...
-|`${apicurio.config.dynamic.allow-all}`
-|
-|
-|`apicurio.ui.features.read-only.enabled.dynamic.allow`
-|`unknown`
-|`${apicurio.config.dynamic.allow-all}`
-|
-|
-|===
-
```
paoloantinori, line 20: Including the partial with `leveloffset=+1` promotes the `== {registry} configuration options` heading (previously `==` in this assembly) to `===`. That's a structural change to the rendered TOC — please confirm the surrounding `==` sections still look right, or drop the `leveloffset` if the intent is to keep it at the same level.

docs/modules/ROOT/partials/getting-started/ref-registry-all-configs.template
```diff
@@ -1,5 +1,8 @@
+// ref-all-registry-configs
```
paoloantinori, line 1: Adding the heading + content-type to the template means the consuming `include::` will pull them in unconditionally. If the template is ever consumed elsewhere without `leveloffset=+1`, the heading level will be wrong in that context. A `tags` region + `tags=` filter on the include would make this safer.

paoloantinori (commented):
The intent here is sound — extracting the generated config reference into a partial and including it back into the assembly avoids drifting copies and matches the pattern used elsewhere in the docs. I read the full diff (1657 deletions in the assembly + 3 additions in the template). A few things worth double-checking before this lands:

- **Heading level shift.** Previously `[id="all-registry-configs_{context}"]`, `:_mod-docs-content-type: REFERENCE`, and the `== {registry} configuration options` heading lived directly in `assembly-config-reference.adoc` at level `==`. Now they live in the partial and are included with `leveloffset=+1`, which promotes them to `===`. If the surrounding assembly has other `==` sections, the config reference becomes a subsection of the nearest preceding `==` rather than a sibling — worth verifying the rendered TOC still looks right. The original layout was explicitly `==`, so this is a structural change, not just a move.
- **`include::` without a tag filter.** `include::{mod-loc}getting-started/ref-registry-all-configs.adoc[leveloffset=+1]` pulls in the entire partial. If the build step that turns the `.template` into the `.adoc` ever emits additional content (e.g. a generated preamble or trailing marker), it will be silently pulled into the assembly too. Consider either a `tags` region in the template + a `tags=` filter on the include, or a comment in the template documenting what the include is allowed to consume.
- **`mergeable_state` is `dirty`.** GitHub reports this branch as not cleanly mergeable, and the PR has been pinging on staleness repeatedly. A rebase onto current `main` is needed before review can really validate the rendered output.
- **Empty PR body.** There's no description here — a one-liner explaining the motivation (presumably "restore the partial include that was inlined at some point, fix vale compliance on the template") would help reviewers and future archaeology.

Minor: the new lines in `ref-registry-all-configs.template` add `// ref-all-registry-configs` as what looks like a sentinel comment — fine, but if it's meant as a parse marker for the generator it's worth a one-word comment saying so.

eameh:
Superseded by the phase-0 template rewrites (#9024, #9026, #9029), which replaced this assembly wholesale with a version that passes the vale AsciiDocDITA gate.

Issue #7530, opened 2026-03-11
EPIC: Kubernetes-Native Registry as a Service


EricWittmann:
## Summary

Enable Apicurio Registry to be run as an internal IT service on Kubernetes, where an IT department
can declaratively provision data-isolated Registry instances for internal customers using
Kubernetes-native mechanisms — without requiring an external OIDC provider or manual configuration.

## Motivation

Organizations running Kubernetes/OpenShift want to offer Apicurio Registry as an internal service.
Today this requires:

- **An external OIDC provider** (e.g. Keycloak) for authentication — adding infrastructure overhead
  and complexity for teams that already have Kubernetes ServiceAccounts and RBAC
- **Manual authorization configuration** — no way to declare access control alongside artifacts in
  a GitOps-friendly manner
- **Manual instance provisioning** — no operator-level support for spinning up multiple data-isolated
  Registry instances for different teams/projects

By adding Kubernetes-native authentication, declarative authorization, and multi-instance operator
support, Registry becomes a first-class Kubernetes service that IT departments can offer with
minimal overhead.

## Architecture / Approach

This epic has three pillars:

### 1. Kubernetes-Native Authentication
Add a new authentication handler that validates Kubernetes ServiceAccount tokens via the
TokenReview API, eliminating the need for an external OIDC provider. This integrates naturally
with existing Kubernetes identity mechanisms.

### 2. Declarative Authorization via Kubernetes Resources
Enable application-managed authorization roles where ACL (Access Control List) declarations live
alongside artifacts in Kubernetes storage and GitOps storage. Authorization is declared via
Kubernetes resource configuration, making it fully GitOps-compatible and auditable.

### 3. Multi-Instance Provisioning via Operator
Enhance the Registry operator to provision multiple data-isolated Registry instances, each with
dedicated storage (database, Kafka topic, Git repository, or Kubernetes namespace), enabling true
multi-tenancy at the infrastructure level.

## Sub-Issues

| # | Title | Status |
|---|-------|--------|
| #7532 | Kubernetes-native authentication support | To Do |
| #7533 | Declarative authorization via Kubernetes resource configuration | To Do |
| #7536 | Operator support for multi-instance data-isolated Registry provisioning | To Do |

### Phase 1: Authentication
- Implement Kubernetes-native authentication handler (ServiceAccount token validation via
  TokenReview API)
- Add configuration options for enabling/disabling Kubernetes auth
- Ensure backward compatibility with existing OIDC authentication

[...]

PR #8076, opened 2026-05-26
feat(auth): Kubernetes-native authentication via TokenReview API

### LINKED ISSUE #7532: Kubernetes-native authentication support
Part of #7530

## Summary

Add a new authentication handler that integrates with Kubernetes/OpenShift authentication
mechanisms, allowing Registry to authenticate users via Kubernetes ServiceAccount tokens
validated through the TokenReview API. This eliminates the requirement for an external OIDC
provider (e.g. Keycloak) when running Registry on Kubernetes.

## Motivation

Organizations running Registry on Kubernetes already have an identity layer — Kubernetes
ServiceAccounts. Requiring an external OIDC provider adds infrastructure overhead and complexity.
By supporting Kubernetes-native authentication, Registry integrates naturally into the Kubernetes
ecosystem and reduces operational burden for IT teams offering Registry as an internal service.

### New Authentication Handler
- Implement a new `KubernetesAuthHandler` (or similar) that:
  - Extracts bearer tokens from incoming requests
  - Validates tokens against the Kubernetes TokenReview API
  - Extracts user identity (ServiceAccount name, namespace, groups) from the TokenReview response
  - Maps Kubernetes identity information to Registry's internal principal model

### Configuration
- Add configuration options to enable Kubernetes-native authentication:
  - `apicurio.authn.kubernetes.enabled` — enable/disable Kubernetes auth handler
  - `apicurio.authn.kubernetes.token-review.url` — optional custom TokenReview API endpoint
  - `apicurio.authn.kubernetes.audiences` — optional audience validation
- Ensure Kubernetes auth can coexist with existing OIDC authentication (multiple auth handlers)

### Documentation
- Document setup and configuration for Kubernetes-native authentication
- Provide examples for common scenarios (ServiceAccount-based access, cross-namespace access)

## Acceptance Criteria

- [ ] A new Kubernetes authentication handler is implemented and can be enabled via configuration
- [ ] ServiceAccount tokens are validated via the Kubernetes TokenReview API
- [ ] User identity (name, namespace, groups) is correctly extracted and mapped to Registry's
      principal model
- [ ] Kubernetes auth works alongside existing OIDC authentication (both can be enabled
      simultaneously)
- [ ] Configuration is documented with examples

## Dependencies / Blocks

- Blocked by: None
- Blocks: Declarative authorization (partially — auth provides the identity that authorization
  checks against)

carlesarnal:
## Summary

- Add Kubernetes-native authentication support using the TokenReview API, eliminating the need for an external OIDC provider when running on Kubernetes (#7532)
- Implement `KubernetesAuthenticationStrategy` that validates Bearer tokens via K8s TokenReview with result caching
- Add `KubernetesRoleProvider` that maps Kubernetes groups to registry roles (sr-admin, sr-developer, sr-readonly)
- Extend operator CRD with `kubernetesAuth` spec for declarative configuration

### Architecture

Follows the existing authentication strategy pattern:

```
Request → Authorization: Bearer <ServiceAccount token>
        → KubernetesAuthenticationStrategy
        → TokenReview API call (cached)
        → KubernetesIdentityProvider → SecurityIdentity with K8s groups as roles
        → KubernetesRoleProvider → maps groups to registry roles
```

### Configuration

```properties
# Enable K8s auth and add to mechanism chain
apicurio.authn.kubernetes.enabled=true
apicurio.authn.mechanism.priority=kubernetes,basic,proxy-header,oidc

# Optional: audience validation
apicurio.authn.kubernetes.api-audiences=api,https://registry.example.com

# Optional: cache TTL (default 5 minutes)
apicurio.authn.kubernetes.cache-expiration=5

# Map K8s groups to registry roles
apicurio.auth.role-source=kubernetes
apicurio.auth.role-source.kubernetes.group-mapping.admin=cluster-admins,registry-admins
apicurio.auth.role-source.kubernetes.group-mapping.developer=developers
apicurio.auth.role-source.kubernetes.group-mapping.readonly=viewers
```

### Operator CRD

```yaml
spec:
  app:
    auth:
      kubernetesAuth:
        enabled: true
        apiAudiences: "api"
        cacheExpiration: 5
        groupMapping:
          adminGroups: "cluster-admins"
          developerGroups: "developers"
          readOnlyGroups: "viewers"
```

### Prerequisites

The registry's ServiceAccount needs `system:auth-delegator` ClusterRoleBinding for TokenReview API access:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: registry-auth-delegator
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:auth-delegator
subjects:
  - kind: ServiceAccount
    name: apicurio-registry
    namespace: registry
```

### Files changed

**New files (11):**
- `KubernetesAuthenticationRequest.java` — auth request with username, UID, groups
- `KubernetesCredential.java` — credential marker for K8s-authenticated requests
- `KubernetesIdentityProvider.java` — builds SecurityIdentity from K8s auth
- `KubernetesAuthenticationStrategy.java` — core: token extraction, TokenReview API, caching
- `KubernetesRoleProvider.java` — K8s group-to-registry-role mapping
- `KubernetesAuthSpec.java` / `KubernetesGroupMappingSpec.java` — operator CRD model
- `KubernetesAuth.java` — operator env var mapping
- 3 test files (identity provider, request, strategy with mock K8s client)

**Modified files (7):**
- `AuthConfig.java` — 6 new config properties
- `AppAuthenticationMechanism.java` — `"kubernetes"` strategy in auth chain
- `AuthorizedInterceptor.java` — includes K8s auth in enabled check
- `RoleBasedAccessController.java` — `"kubernetes"` role source
- `EnvironmentVariables.java`, `Auth.java`, `AuthSpec.java` — operator wiring

Relates to #7530, closes #7532

## Test plan

- [x] Unit test: `KubernetesAuthenticationRequestTest` — immutability, null groups
- [x] Unit test: `KubernetesIdentityProviderTest` — SecurityIdentity construction, groups-as-roles, credential attachment
- [x] Unit test: `KubernetesAuthenticationStrategyTest` — token extraction, TokenReview mock, cache hits, audience config, error handling, case-insensitive Bearer prefix
- [x] Compilation verified: app module and operator module compile cleanly
- [ ] Integration test with kind cluster (manual verification)
- [ ] Config doc regeneration after pre-existing compilation issues are resolved


app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
```diff
@@ -123,13 +128,25 @@ List<AuthenticationStrategy> buildAuthChain() {
             return new OidcAuthenticationStrategy(oidcAuthenticationMechanism, authConfig,
                     auditLog, webClient, log, this);
         });
+        strategyFactories.put("kubernetes", () -> {
```
EricWittmann, line 137: **Bearer token conflict with OIDC**

Both this Kubernetes strategy and the existing `OidcAuthenticationStrategy` consume `Bearer` tokens from the `Authorization` header. When both are enabled simultaneously, whichever strategy is first in the priority list will attempt to validate *every* Bearer token — even tokens intended for the other strategy.

For example, if priority is `kubernetes,oidc`:
1. A user sends an OIDC JWT as a Bearer token
2. The Kubernetes strategy intercepts it first and calls the TokenReview API
3. K8s API rejects it (it's not a K8s ServiceAccount token) — result is negative-cached for 30 seconds
4. The chain falls through to OIDC, which succeeds
5. But every request paid for a wasted K8s API call (or a wasted cache entry)

The reverse happens if priority is `oidc,kubernetes` — K8s SA tokens get sent to the OIDC provider first.

Two options to consider:

**Option A: Heuristic short-circuit.** K8s ServiceAccount tokens are JWTs with an issuer like `kubernetes/serviceaccount` or `https://kubernetes.default.svc`. You could peek at the JWT issuer claim (without full validation) and skip the TokenReview if the issuer doesn't look like a K8s token. This is best-effort but would avoid 99% of wasted calls.

**Option B: Mutual exclusion.** If supporting both OIDC and K8s auth simultaneously isn't a use case, add a startup validation in `AppAuthenticationMechanism.buildAuthChain()` that fails fast if both `oidcAuthEnabled` and `kubernetesAuthEnabled` are true. Something like:
```java
if (authConfig.oidcAuthEnabled && authConfig.kubernetesAuthEnabled) {
    throw new IllegalStateException(
        "Cannot enable both OIDC and Kubernetes authentication simultaneously. "
        + "Both use Bearer tokens and would conflict.");
}
```

Which approach makes more sense for the intended use cases?

app/src/main/java/io/apicurio/registry/auth/KubernetesAuthenticationStrategy.java
```diff
@@ -0,0 +1,214 @@
...
+            return null;
+        }
+    }
+
+    private void cachePut(String key, WrappedValue<TokenReviewResult> value) {
+        evictExpiredEntries();
+        if (cache.size() >= MAX_CACHE_SIZE) {
+            return;
+        }
+        cache.put(key, value);
```
EricWittmann, line 158: **Cache implementation has O(n) eviction and is vulnerable to cache poisoning**

A few concerns with this hand-rolled cache:

1. **O(n) scan on every put:** `evictExpiredEntries()` iterates the entire `ConcurrentHashMap` on every `cachePut` call. Under load this becomes expensive — every authentication attempt pays O(cache_size) even if nothing is expired.

2. **Silent drop when full:** When the cache reaches `MAX_CACHE_SIZE` (10,000), new entries are silently dropped. This means legitimate tokens stop being cached and every request falls through to the K8s API — effectively a self-inflicted DoS on the API server.

3. **Cache poisoning vector:** An attacker can send 10,000 unique invalid Bearer tokens. Each gets negative-cached for 30 seconds (`API_ERROR_CACHE_DURATION`) or the full `cacheDuration`. Once the cache is full, legitimate users' tokens can't be cached, degrading every real request to an uncached K8s API call.

Consider replacing with Caffeine (already in the Quarkus dependency tree via `quarkus-cache`):
```java
private final Cache<String, TokenReviewResult> cache = Caffeine.newBuilder()
    .maximumSize(MAX_CACHE_SIZE)
    .expireAfterWrite(Duration.ofMinutes(authConfig.kubernetesTokenCacheExpiration))
    .build();
```
This gives O(1) amortized puts, automatic LRU eviction when full (no silent drops), and built-in expiration without manual scanning.

This is probably fine as a follow-up — would you mind opening a GitHub issue to track it?

app/src/main/java/io/apicurio/registry/auth/KubernetesAuthenticationStrategy.java
```diff
@@ -0,0 +1,214 @@
...
+
+        String token = authHeader.substring(BEARER_PREFIX.length()).trim();
+        if (token.isEmpty()) {
+            return Uni.createFrom().nullItem();
+        }
+
+        String tokenHash = sha256(token);
+
+        WrappedValue<TokenReviewResult> cached = cache.get(tokenHash);
+        if (cached != null && !cached.isExpired()) {
```
EricWittmann, line 90: **Thundering-herd on cache miss**

Two (or more) concurrent requests carrying the same token can both see a cache miss here and both call `performTokenReview`. While not dangerous (both get the same result from K8s), it causes redundant API calls — potentially many under burst traffic right after a cache entry expires.

A simple fix would be to use `ConcurrentHashMap.computeIfAbsent` to ensure only one thread performs the TokenReview for a given token hash, while others wait for the result:
```java
WrappedValue<TokenReviewResult> entry = cache.compute(tokenHash, (key, existing) -> {
    if (existing != null && !existing.isExpired()) {
        return existing; // still valid
    }
    return performTokenReview(token); // only one thread does this per key
});
```
(Or this falls out naturally if you move to Caffeine's `get(key, mappingFunction)`.)

Not blocking, but worth considering — especially since the K8s TokenReview API call is relatively expensive.

app/src/main/java/io/apicurio/registry/auth/KubernetesAuthenticationStrategy.java
```diff
@@ -0,0 +1,214 @@
...
+            }
+
+            log.debug("Kubernetes TokenReview: authenticated user '{}'", username);
+
+            TokenReviewResult result = new TokenReviewResult(true, username, uid, groups);
+            cachePut(tokenHash, new WrappedValue<>(cacheDuration, Instant.now(), result));
+            return result;
+
+        } catch (Exception e) {
+            log.warn("Kubernetes TokenReview API call failed: {}", e.getMessage());
```
EricWittmann, line 146: **Negative caching of API errors may be too aggressive**

This catches *all* exceptions and caches the failure for 30 seconds keyed to the specific token hash. Consider a scenario where the K8s API server is temporarily unreachable (network blip, pod restart):

1. 50 different legitimate users hit the registry during the ~2 second outage
2. Each unique token gets a 30-second negative cache entry
3. All 50 users must wait 30 seconds before their next request can succeed

The issue is that a transient infrastructure failure (not related to the token at all) gets cached *per token*, penalizing individual users. A circuit-breaker pattern would be more appropriate here: after N consecutive failures, stop calling the K8s API entirely for M seconds (regardless of which token), then retry one probe request. This way the failure is attributed to the API server, not to individual tokens.

Would you mind opening a GitHub issue to track this for follow-up? It's fine to merge as-is for an initial implementation, but it could cause confusion in production when transient K8s API issues lead to users being locked out for 30 seconds.

operator/controller/src/main/java/io/apicurio/registry/operator/feat/security/KubernetesAuth.java
```diff
@@ -0,0 +1,52 @@
...
+        if (spec == null || !Boolean.TRUE.equals(spec.getEnabled())) {
+            return;
+        }
+
+        env.put(EnvironmentVariables.APICURIO_AUTHN_KUBERNETES_ENABLED,
+                createEnvVar(EnvironmentVariables.APICURIO_AUTHN_KUBERNETES_ENABLED, "true"));
+
+        env.put(EnvironmentVariables.APICURIO_AUTHN_MECHANISM_PRIORITY,
+                createEnvVar(EnvironmentVariables.APICURIO_AUTHN_MECHANISM_PRIORITY,
+                        "kubernetes,basic,proxy-header,oidc"));
```
EricWittmann: **Operator silently overrides user-configured mechanism priority**

This unconditionally sets `APICURIO_AUTHN_MECHANISM_PRIORITY` to `"kubernetes,basic,proxy-header,oidc"` whenever Kubernetes auth is enabled. If a user has already customized their mechanism priority via env var overrides on the deployment (e.g., to put `basic` first for a specific use case), the operator will silently overwrite it on the next reconciliation.

Suggested fix: only set the priority if it hasn't already been configured by the user:
```java
if (!env.containsKey(EnvironmentVariables.APICURIO_AUTHN_MECHANISM_PRIORITY)) {
    env.put(EnvironmentVariables.APICURIO_AUTHN_MECHANISM_PRIORITY,
            createEnvVar(EnvironmentVariables.APICURIO_AUTHN_MECHANISM_PRIORITY,
                    "kubernetes,basic,proxy-header,oidc"));
}
```

Alternatively, consider adding a `mechanismPriority` field to `KubernetesAuthSpec` (or `AuthSpec`) so users can explicitly control the ordering through the CRD rather than relying on env var overrides.

app/src/main/java/io/apicurio/registry/auth/AuthConfig.java
```diff
@@ -170,8 +170,32 @@ public class AuthConfig {
     @Info(category = CATEGORY_AUTH, description = "When enabled, authorization checks are skipped and the proxy is trusted to have performed authorization", availableSince = "3.1.0.Final")
     boolean proxyHeaderTrustProxyAuthorization;
 
+    @ConfigProperty(name = "apicurio.authn.kubernetes.enabled", defaultValue = "false")
```
EricWittmann, line 173: **Preview / GA question**

Is this feature intended to ship as a preview (experimental) feature or as GA? The PR description mentions that integration testing on a real cluster is still pending (checked box is manual verification only).

If this is a preview feature, should these config properties be annotated with `@Info(experimental = true)`? Per project conventions, that would gate them behind `apicurio.features.experimental.enabled` and make it clear to users that the API/behavior may change. For example:
```java
@ConfigProperty(name = "apicurio.authn.kubernetes.enabled", defaultValue = "false")
@Info(category = CATEGORY_AUTH, description = "Enable Kubernetes TokenReview authentication",
      availableSince = "3.3.0", experimental = true)
boolean kubernetesAuthEnabled;
```

EricWittmann (commented):
Nice work — the architecture is clean and follows the existing auth strategy pattern well. The TokenReview approach is a natural fit for K8s-native deployments. I've left several comments inline, mostly around edge cases in the cache implementation and a potential conflict with OIDC Bearer token handling. A few of these are fine as follow-up issues rather than blocking this PR.

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
carlesarnal: I'll go with motual exclusion.

PR #8098, opened 2026-05-27
feat(otel): add TracingResponseFilter for error status codes

paoloantinori:
## Summary

Adds a `ContainerResponseFilter` that records HTTP response status codes on OpenTelemetry spans and marks server errors (5xx) with ERROR status (GitHub Issue #7150, item 6).

- Sets `http.response.status_code` attribute (OTel semantic convention, long type)
- 5xx responses set span status to `ERROR` with generic `"HTTP <code>"` description (no internal details leaked)
- 4xx/3xx/2xx record the status code attribute but leave span status as `UNSET` (per OTel spec, 4xx is a client error, not a server error)
- Adds `ATTR_HTTP_RESPONSE_STATUS_CODE` constant to `OTelAttributes`

## Related Issues

- Fixes part of #7150 (item 6: response status on spans)

## Changes

| File | Change |
|------|--------|
| `TracingResponseFilter.java` | New `@Provider` response filter |
| `OTelAttributes.java` | Add `ATTR_HTTP_RESPONSE_STATUS_CODE` |
| `TracingResponseFilterTest.java` | 6 unit tests |

## Test plan

- [x] 200 OK — attribute set, span status UNSET
- [x] 301 Redirect — attribute set, span status UNSET
- [x] 404 Not Found — attribute set, span status UNSET
- [x] 500 Internal Server Error — attribute set, span status ERROR, description "HTTP 500"
- [x] 503 Service Unavailable — attribute set, span status ERROR, description "HTTP 503"
- [x] No active span — no-op, no exception
- [x] Checkstyle passes


paoloantinori:
@carlesarnal gentle ping — this PR is ready for review whenever you have a moment. CI is green and all checks pass.

paoloantinori:
The CI failure is the known `h2-debezium` infrastructure issue tracked in #8244 — `minikube tunnel` isn't routing LoadBalancer traffic on the CI runners. All other test profiles pass. This is also failing on `main` (zero green Verify runs in 3+ days). The `waiting-on-author` label was set by the CI bot, not a reviewer.

paoloantinori:
Friendly ping — this PR has been open since May and CI is green. It adds a ContainerResponseFilter that enriches REST spans with HTTP response status codes and sets span ERROR status for 5xx responses. Small change (~80 lines), no API impact. Ready for review whenever convenient.

EricWittmann (commented):
## Code Review — 3 Findings

**1. `OTelAttributes.java` — Custom constant duplicates one from the OTel semconv library**

The PR defines `AttributeKey.longKey("http.response.status_code")` manually, but `io.opentelemetry.semconv:opentelemetry-semconv:1.37.0` is already on the compile classpath and provides `io.opentelemetry.semconv.HttpAttributes.HTTP_RESPONSE_STATUS_CODE` — the exact same key. Using the library constant avoids drift if the semantic convention ever changes, and is more idiomatic. The custom constant in `OTelAttributes` and its usages in the filter and tests could all be replaced with the standard import.

**2. `TracingResponseFilter.java:39` — Inconsistent null-check pattern vs. existing `TracingFilter`**

The existing `TracingFilter.java` (line 28-29) checks `currentSpan != null && currentSpan.isRecording()`, but this new filter skips the null check and goes straight to `currentSpan.isRecording()`. While `Span.current()` is contractually non-null in the OTel API, the codebase has an established defensive pattern. Minor, but worth aligning for consistency since these two filters live in the same package and will be read side-by-side.

**3. `OTelAttributes.java` — Updated Javadoc is slightly misleading given the implementation**

The new Javadoc says "Standard OTel semantic convention attributes use their canonical names without prefix," implying they're imported from the OTel library. But the constant is actually redefined locally. If finding #1 is addressed (use the semconv library constant directly), this Javadoc becomes moot — the constant would move out of this class entirely. If not addressed, the comment should clarify these are locally-defined copies of standard attribute keys.

---

Overall this is a solid, focused PR. The most actionable item is #1 (use the existing semconv constant instead of redefining it). The other two are minor consistency nits.

carlesarnal:
Hey @paoloantinori, thanks for the PR! The code is clean and well-tested.

However, during review I found that Quarkus OpenTelemetry auto-instrumentation (`quarkus-opentelemetry`, which is already enabled in this project) already sets `http.response.status_code` on HTTP server spans and marks 5xx responses with `StatusCode.ERROR` — via its built-in `HttpServerAttributesExtractor` and `HttpSpanStatusExtractor`.

To verify this, I wrote a `@QuarkusTest` that captures real Quarkus-generated spans using `InMemorySpanExporter` and asserts the attribute is present — **it passes on `main` without `TracingResponseFilter`**: #8415

This means the filter is redundant and adds unnecessary overhead on every request. It could also mask auto-instrumentation regressions if Quarkus ever changes its behavior, since the attribute would still appear to be set.

**Suggestion:** Consider dropping `TracingResponseFilter` and `ATTR_HTTP_RESPONSE_STATUS_CODE` from this PR. If you'd like to keep the test coverage proving the behavior exists, feel free to adopt the test from #8415.

paoloantinori:
@carlesarnal You're right — I verified independently that Quarkus OTel auto-instrumentation sets `http.response.status_code` and marks 5xx as `StatusCode.ERROR` via `HttpServerAttributesExtractor` and `HttpSpanStatusExtractor`. The `TracingResponseFilter` is indeed redundant.

Closing this PR. Thanks for the thorough investigation and for writing the test in #8415 — happy to see that adopted as proof the behavior exists.

EricWittmann:
This is a good example of how actual expertise/knowledge of an area is still pretty important.  :)

PR #8132, opened 2026-06-02
fix(rest): return HTTP 400 for RuleViolationException instead of 409

### LINKED ISSUE #8131: RuleViolationException should return HTTP 400, not 409
## Summary

When using the Registry v3 API to create an artifact (`POST /apis/registry/v3/groups/{groupId}/artifacts`),
both of the following scenarios return HTTP 409:

- Artifact with the same ID already exists → `ArtifactAlreadyExistsException`
- Schema content is invalid (when a VALIDITY rule is enabled) → `RuleViolationException`

A `RuleViolationException` is a client input validation error and should return **400 Bad Request**, not
**409 Conflict**. Returning 409 for both cases forces clients to parse the response body and match on the
`name` field to differentiate them, which is not ideal.

## Expected Behavior

| Scenario | Exception | HTTP Status |
|----------|-----------|-------------|
| Artifact already exists | `ArtifactAlreadyExistsException` | 409 Conflict |
| Schema content invalid / rule violation | `RuleViolationException` | 400 Bad Request |

## Why It Matters

- Applications should be able to act on status codes without parsing error messages
- A validation failure is a client input error (400), not a resource conflict (409)
- This aligns with REST/HTTP best practices

## Affected Code

The status code mapping is in `HttpStatusCodeMap.java`:

```java
map.put(RuleViolationException.class, HTTP_CONFLICT); // currently 409, should be 400
```

## Proposed Approach

Since changing the default behavior is a breaking API change, add a configuration property to control the
behavior:

- **Property:** `apicurio.rest.legacy-error-codes.enabled` (default: `false`)
- **When `false` (default):** `RuleViolationException` returns 400
- **When `true`:** `RuleViolationException` returns 409 (legacy behavior)

This defaults to the corrected behavior while providing an escape hatch for existing clients that depend
on the legacy 409 status code.

## Additional Context

Reported by an upstream user via Zulip. The issue affects any endpoint that performs rule validation, not
just artifact creation.

EricWittmann:
## Summary

- `RuleViolationException` now returns HTTP 400 (Bad Request) by default instead of 409 (Conflict), allowing clients to distinguish validation errors from resource conflicts by status code alone
- Added `apicurio.rest.legacy-error-codes.enabled` config property (default: `false`) so existing clients can opt into the legacy 409 behavior
- Converted `HttpStatusCodeMap` from static initialization to `@PostConstruct` instance initialization, enabling config-aware status code mapping
- Updated the OpenAPI spec to document rule violations as 400 responses with a new `RuleViolationBadRequest` response component

## Related Issue

Fixes #8131

## Test Plan

- [ ] Build: `./mvnw clean install -DskipTests`
- [ ] Run unit tests: `./mvnw test -pl app`
- [ ] Run checkstyle: `./mvnw checkstyle:check -pl app`
- [ ] Verify rule violations return HTTP 400 by default
- [ ] Verify setting `apicurio.rest.legacy-error-codes.enabled=true` restores 409 behavior
- [ ] Verify `ArtifactAlreadyExistsException` still returns 409


carlesarnal:
I understand the change and the intention, but isn't this strictly speaking a breaking change?

EricWittmann:
I'm of two minds on this.  It's a breaking change, but it's also a bug.  I added a config property so users can revert to the previous buggy behavior.  I could make the default the buggy behavior and have the "fixed" behavior be opt-in instead.  wdyt?

PR #8134, opened 2026-06-02
feat(cli): add authentication support (Basic auth and OAuth2). Fixes #8055

vandanayadav7:
## Summary

Adds `acr login` and `acr logout` commands to authenticate with secured registry instances using Basic auth or OAuth2 client credentials flow. Credentials are stored securely in the OS keychain (macOS Keychain / Linux Secret Service), never in config files.

## Architecture

```
┌──────────────────────────────────────────────────┐
│                   CLI Commands                   │
│  ┌──────────────┐         ┌───────────────┐      │
│  │ LoginCommand │         │ LogoutCommand │      │
│  └──────┬───────┘         └───────┬───────┘      │
│         │                         │              │
│         ▼                         ▼              │
│  ┌─────────────────────────────────────────┐     │
│  │           CredentialStore               │     │
│  │  (facade — per-context key management)  │     │
│  └──────────────────┬──────────────────────┘     │
│                     │ @Inject                    │
│  ┌──────────────────┴──────────────────────┐     │
│  │    CredentialProvider (interface)       │     │
│  │  selected by CredentialProviderProducer │     │
│  └──────────────────┬──────────────────────┘     │
│          ┌──────────┴──────────┐                 │
│          ▼                     ▼                 │
│  ┌───────────────┐   ┌────────────────┐          │
│  │ MacOSCred.    │   │ LinuxCred.     │          │
│  │ Provider      │   │ Provider       │          │
│  │ (security)    │   │ (secret-tool)  │          │
│  └───────────────┘   └────────────────┘          │
│                                                  │
│  Test: TestCredentialProvider (@Alternative)     │
│        in-memory — no OS keychain in CI          │
│                                                  │
│  ┌─────────────────────────────────────────┐     │
│  │        ConfigModel.Context              │     │
│  │  (authType, username, tokenEndpoint,    │     │
│  │   clientId, scope — stored in JSON)     │     │
│  └─────────────────────────────────────────┘     │
│                                                  │
│  ┌─────────────────────────────────────────┐     │
│  │              Client                     │     │
│  │  (configures SDK auth from stored creds)│     │
│  └─────────────────────────────────────────┘     │
└──────────────────────────────────────────────────┘
```

**Separation of concerns:**
- `CredentialProvider` — interface for platform-specific keychain operations
- `CredentialProviderProducer` — CDI producer that selects MacOS/Linux provider at runtime
- `CredentialStore` — public facade that maps context+key to keychain accounts
[...]

PR #8182, opened 2026-06-09
feat(cli): add file-based credential fallback with --allow-unsafe-credential-storage

vandanayadav7:
## Summary

Adds a file-based credential fallback for CLI environments without an OS keychain (UBI images, GitHub Actions runners, headless servers). Implements the `--allow-unsafe-credential-storage` flag discussed in #8134.

## Behavior

| Scenario | Result |
|----------|--------|
| Keychain available (default) | Uses OS keychain |
| Keychain available + `--allow-unsafe-credential-storage` | Uses OS keychain (flag is a fallback, not a force) |
| No keychain + no flag | Error: suggests `--allow-unsafe-credential-storage` |
| No keychain + flag | Stores in `credentials.json` with warning |

The preference is stored in config — subsequent commands auto-use file storage without needing the flag again.

## Security

- `credentials.json` written atomically (temp file + rename) to prevent corruption
- File permissions set to `0600` (owner read/write only)
- Warning printed: "Credentials stored in a file. This is not recommended for production use."
- Explicit opt-in required — never silently falls back to file
- `unsafeCredentialStorage` only set to `true` when file fallback was actually used
- Logout clears the preference and deletes `credentials.json`
- Empty credentials file deleted on last credential removal

## Changes

| File | Purpose |
|------|---------|
| `FileCredentialProvider.java` | New — reads/writes `credentials.json` with atomic writes and 0600 permissions |
| `CredentialStore.java` | Tries OS keychain first, falls back to file when flag set. Returns whether fallback was used |
| `LoginCommand.java` | Added `--allow-unsafe-credential-storage` flag, `UNSAFE_STORAGE_WARNING` constant |
| `ConfigModel.java` | Added `unsafeCredentialStorage` boolean to Context, cleared on `clearAuth()` |
| `Client.java` | Passes `unsafeCredentialStorage` preference when retrieving credentials |
| `AuthCommandTest.java` | 3 new tests: basic with flag, OAuth2 with flag, logout clears flag |
| `README.md` | Added headless/CI section to authentication docs |

## Test Plan

- [x] 19 unit tests pass (231 total CLI tests)
- [x] macOS E2E: keychain available + flag → uses keychain, `unsafeCredentialStorage: false`, no warning
- [x] macOS E2E: keychain login/logout flow works
- [x] Linux E2E: no keychain + no flag → error with suggestion
- [x] Linux E2E: no keychain + flag → file storage with warning, `unsafeCredentialStorage: true`
- [x] Linux E2E: authenticated API call works with file-based credentials
- [x] Linux E2E: logout clears credentials.json
- [x] File permissions verified as 0600
- [x] Logout clears unsafeCredentialStorage flag

## Follow-up from

- #8134 (CLI auth support) — discussion with @jsenko about headless environments

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -249,9 +249,16 @@ acr logout
 
 Authentication is per-context — each context can use different credentials.
 
+**Headless/CI environments (no OS keychain):**
+```bash
+acr login --username <username> --password <password> --allow-unsafe-credential-storage
+```
+Credentials are stored in a local file instead of the OS keychain. A warning is printed. The preference is saved so subsequent commands don't need the flag again.
+
 **Prerequisites for credential storage:**
 - macOS: No prerequisites (uses Keychain)
 - Linux: `secret-tool` required (`sudo apt install libsecret-tools` or `sudo dnf install libsecret`)
+- Headless/CI: Use `--allow-unsafe-credential-storage` if no keychain is available
 
 ### Working with Groups
 
--- a/cli/src/main/java/io/apicurio/registry/cli/auth/CredentialStore.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/auth/CredentialStore.java
@@ -1,6 +1,7 @@
 package io.apicurio.registry.cli.auth;
 
 import io.apicurio.registry.cli.common.CliException;
+import io.apicurio.registry.cli.config.Config;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import org.jboss.logging.Logger;
@@ -8,7 +9,7 @@ import org.jboss.logging.Logger;
 import static io.apicurio.registry.cli.common.CliException.APPLICATION_ERROR_RETURN_CODE;
 
 /**
- * Stores and retrieves credentials per context using the OS keychain.
+ * Stores and retrieves credentials per context using the OS keychain or file-based fallback.
  */
 @ApplicationScoped
 public class CredentialStore {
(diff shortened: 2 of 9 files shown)


jsenko (commented):
## Review

Clean design, good security posture. Three items to address:

### 1. Test `testReloginHonorsSavedUnsafePreference` doesn't exercise the preference path

Both logins use the default `TestCredentialProvider` which succeeds on `store()`. First login: keychain works, `usedFileFallback=false`, `unsafeCredentialStorage=false`. Second login: keychain still works, `allowUnsafe = false || false = false`. The test proves two consecutive logins work — it never exercises the saved-preference mechanism.

To actually test it, the keychain must fail on both logins:

```java
@Test
public void testReloginHonorsSavedUnsafePreference() {
    final var testProvider = (TestCredentialProvider) credentialProvider;
    try {
        testProvider.setFailOnStore(true);

        // First login — file fallback used, preference saved
        executeAndAssertSuccess("login", "--username", TEST_USERNAME, "--password", TEST_PASSWORD,
                "--allow-unsafe-credential-storage");
        assertThat(getTestContext().isUnsafeCredentialStorage()).isTrue();

        // Re-login without flag — should succeed via saved preference
        out.getBuffer().setLength(0);
        executeAndAssertSuccess("login", "--username", TEST_USERNAME, "--password", TEST_PASSWORD);
        assertThat(out.toString()).contains("Logged in to context");
    } finally {
        testProvider.setFailOnStore(false);
    }
}
```

Without this fix, a regression that breaks the saved-preference path would go undetected.

### 2. `restrictFilePermissions` silently swallows all errors

In `FileCredentialProvider.java`, the catch block:

```java
} catch (UnsupportedOperationException | IOException ex) {
    // Non-critical — file is written, permissions are a security hardening
}
```

On non-POSIX systems (Windows, or exotic filesystems), this means credentials are stored with default permissions — potentially world-readable. For a credentials file, at minimum log a warning so users on affected systems know:

```java
} catch (UnsupportedOperationException | IOException ex) {
    log.warnf("Could not restrict file permissions on %s — credentials may be readable by other users.", path);
}
```

### 3. Stale `credentials.json` after switching back to keychain

If a user previously used file storage, then installs `libsecret-tools` and re-logs in with keychain, the credential is now in the keychain but the old `credentials.json` still sits on disk with stale credentials. `retrieve` with `unsafeStorage=false` correctly goes to keychain, so functionally this is fine — but the stale file remains until the user runs `logout` (which cleans both stores). Consider either documenting this, or cleaning up `credentials.json` in the login path when keychain succeeds and a credentials file exists.

jsenko (approved):

PR #8183, opened 2026-06-09
docs: vale DITA compliance fixes for assembly-operator-config-reference

eameh:


pwright (changes requested):
https://deepwiki.com/search/consider-how-lebenpy-would-han_31362479-c7c0-4613-a530-560891d13610?mode=deep

eameh:
Fixed — moved `:_mod-docs-content-type:` to immediately after the `==` heading (not between `[id=...]` and `==`). Leben.py now correctly extracts the heading text as the module title when splitting.

pwright (approved):

PR #8093, opened 2026-05-26
fix(otel): inject OTelMetricsProvider and wire counter increments

paoloantinori:
## Summary

Activates the 5 core OTel counters in `OTelMetricsProvider` that were previously dead code (GitHub Issue #7150, item 1).

- **`GroupsResourceImpl`**: Records `artifacts.created`, `artifacts.deleted`, and `versions.created` on artifact/version lifecycle operations. Dry-run operations are excluded. Artifact deletion looks up the type before deleting for accurate metric attributes.
- **`SearchResourceImpl`**: Records `search.requests` on all 7 search endpoints with the search type as attribute.
- **`RulesServiceImpl`**: Records `rules.evaluated` with success/failure outcome at the `applyAllRules()` convergence point, covering all callers (v3 REST, import, future paths). Threads the `RuleApplicationType` ("CREATE"/"UPDATE") as the operation label.

Skipped for this PR:
- `recordSchemaValidation` (requires per-rule-type detection — follow-up)
- Iceberg counters (separate storage adapter)
- v2 API (legacy, being deprecated)

Depends on #8062 (uses `OTelAttributes` constants class).

## Test plan

- [x] `OTelMetricsProviderTest` — 9 tests pass
- [x] `DatabaseConfigInterceptorTest` — passes
- [x] Checkstyle passes on `app` module
- [x] Full `app` module compiles without errors
- [ ] Integration tests verify no regressions on artifact CRUD and search paths

PR #8186, opened 2026-06-10
fix(metrics): fix health check bugs, config typos, and wire up OTel metrics

EricWittmann:
## Summary

- **Fix `@Liveness` → `@Readiness`** on `PersistenceTimeoutReadinessCheck` — this class was incorrectly registered as a liveness check instead of a readiness check, causing Kubernetes to potentially restart pods on slow storage operations instead of just removing them from the service
- **Fix config property name bugs** — a copy-pasted property name caused `PersistenceTimeoutReadinessCheck` to share its counter-reset-window config with `ResponseErrorLivenessCheck`; three additional typos in property names (`resonse`, `persitence`, `rediness`) are corrected in both Java annotations and `application-prod.properties`
- **Wire up six unused OTel metrics** — `recordArtifactCreated`, `recordArtifactDeleted`, `recordVersionCreated`, `recordSchemaValidation`, `recordRuleEvaluation`, and `recordSearchRequest` were defined in `OTelMetricsProvider` with unit tests but never called from production code; they are now integrated into `GroupsResourceImpl`, `RulesServiceImpl`, and `SearchResourceImpl`

## Test plan

- [ ] Verify existing metrics unit tests pass (`./mvnw test -pl app/`)
- [ ] Verify health checks register under the correct MicroProfile Health category (readiness vs liveness)
- [ ] Verify corrected config property names work in `application-prod.properties`
- [ ] With OTel enabled (`QUARKUS_OTEL_SDK_DISABLED=false`), verify artifact/version/search/rule metrics emit data
- [ ] Verify dry-run artifact/version creation does NOT record artifact/version creation OTel metrics (note: rule evaluation and schema validation metrics *are* intentionally emitted during dry-run because the rules genuinely execute)

> **Breaking change note:** The config property name typo fixes rename three existing properties. Deployments that explicitly set the old typo'd names will need to update:
> - `apicurio.metrics.resonse-error-liveness-check.counter-reset-window-duration.seconds` → `response`
> - `apicurio.metrics.persitence-timeout-readiness-check.status-reset-window-duration.seconds` → `persistence`
> - `apicurio.metrics.response-timeout-rediness-check.status-reset-window-duration.seconds` → `readiness`

### DIFF
--- a/app/src/main/java/io/apicurio/registry/metrics/health/liveness/ResponseErrorLivenessCheck.java
+++ b/app/src/main/java/io/apicurio/registry/metrics/health/liveness/ResponseErrorLivenessCheck.java
@@ -35,7 +35,7 @@ public class ResponseErrorLivenessCheck extends AbstractErrorCounterHealthCheck
      * The counter is reset after some time without errors. i.e. to fail the check after 2 errors in a minute,
      * set the threshold to 1 and this configuration option to 60. TODO report the absolute count as a metric?
      */
-    @ConfigProperty(name = "apicurio.metrics.resonse-error-liveness-check.counter-reset-window-duration.seconds", defaultValue = "60")
+    @ConfigProperty(name = "apicurio.metrics.response-error-liveness-check.counter-reset-window-duration.seconds", defaultValue = "60")
     @Info(category = CATEGORY_HEALTH, description = "Counter reset window duration of response liveness check", availableSince = "1.0.2.Final")
     Integer configCounterResetWindowDurationSec;
 
--- a/app/src/main/java/io/apicurio/registry/metrics/health/readiness/PersistenceTimeoutReadinessCheck.java
+++ b/app/src/main/java/io/apicurio/registry/metrics/health/readiness/PersistenceTimeoutReadinessCheck.java
@@ -9,7 +9,7 @@ import jakarta.inject.Inject;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.health.HealthCheck;
 import org.eclipse.microprofile.health.HealthCheckResponse;
-import org.eclipse.microprofile.health.Liveness;
+import org.eclipse.microprofile.health.Readiness;
 import org.slf4j.Logger;
 
 import java.time.Duration;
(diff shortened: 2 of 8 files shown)


paoloantinori (commented):
Nice work on this PR — the health check fixes and config property corrections are solid, and the OTel metrics wiring is well-placed.

A few observations from someone who's been working in this area (PR #8093 merged last week with overlapping `OTelMetricsProvider` wiring):

**Overlap with #8093:** Our merged PR already wired `OTelMetricsProvider` into `GroupsResourceImpl` and `SearchResourceImpl` with the same injection pattern. This PR's changes to those files will likely conflict with what's already on `main`. You may want to rebase and check if the wiring is already present — some of the `recordArtifactCreated`/`recordVersionCreated`/`recordSearchRequest` calls may already exist.

**What this PR adds beyond #8093:**
- `RulesServiceImpl` metrics (`recordRuleEvaluation`/`recordSchemaValidation`) — we intentionally deferred this, so great to see it here
- The `deleteArtifact` path now fetches `artifactType` before deletion (via `getArtifactMetaData`) — our version just passed `artifactType` from the caller. Your approach is more robust since it gets the actual type from storage
- Dry-run guards (`if (!(dryRun != null && dryRun))`) — matches our pattern

**One concern on `RulesServiceImpl`:** The `recordRuleEvaluation`/`recordSchemaValidation` calls happen inside `applyRule()` which runs for both real and dry-run requests. Since `RulesService` doesn't know about dry-run (that's a REST-layer concern), metrics will be emitted for dry-run rule evaluations too. This might be intentional (rule evaluation *does* happen during dry-run), but worth noting since the `GroupsResourceImpl` dry-run guards explicitly suppress artifact/version creation metrics.

**On `safeIncrement`:** Our PR #8093 added a `safeIncrement()` wrapper in `OTelMetricsProvider` that catches exceptions from counter operations so metrics never break the request flow. This is already on `main` — your wiring benefits from it automatically.

**Health check fixes:** The `@Liveness` → `@Readiness` fix on `PersistenceTimeoutReadinessCheck` and the config property typo corrections are clear wins. The `resonse` → `response`, `persitence` → `persistence`, `rediness` → `readiness` fixes are all correct.

PR #8214, opened 2026-06-11
docs: vale DITA compliance fixes for assembly-intro-to-the-registry

eameh:


eameh:
Superseded by the phase-0 template rewrites (#9024, #9026, #9029), which replaced this assembly wholesale with a version that passes the vale AsciiDocDITA gate.

Issue #5535, opened 2024-11-15
Add `artifactType` to the filtering options in `/search/artifacts` and `/search/versions`


EricWittmann:
### Feature or Problem Description
Cannot filter by artifact type when searching for artifacts or versions.

### Proposed Solution
Add a new query param called `artifactType` to the list of possible filters.

PR #8227, opened 2026-06-11
feat(ui): add artifact type dropdown filter to search toolbar. Fixes #8226

### LINKED ISSUE #8226: feat(ui): add artifact type dropdown filter to artifact search toolbar

The REST API `GET /search/artifacts` and `GET /search/versions` both support filtering by `artifactType` query parameter (added in #5535), but the search page UI never exposes this as a usable filter. Users must currently guess and manually type exact artifact type values like `JSON`, `OPENAPI`, or `PROTOBUF` into a text field, which is undiscoverable and error-prone.

The same issue applies to the `state` filter in version search, where users must type `ENABLED`, `DISABLED`, `DEPRECATED`, or `DRAFT` manually.

### Proposed Solution

When a user selects "Type" or "State" as a filter criterion in the search toolbar, render the value input as a dropdown populated with valid options (fetched from `GET /admin/config/artifactTypes` for types, or a fixed list for states). All other filter types continue to use free-text input.

This gives users discoverable, one-click filtering by artifact type (JSON Schema, OpenAPI, Avro, Protobuf, AsyncAPI, GraphQL, etc.) and by version state.

### Additional Context

- Backend work completed in #5535 (REST API `artifactType` param)
- `FilterBy.artifactType` already defined in `useSearchService.ts`
- `useArtifactTypesService` already fetches available types from the admin API
- The TypeScript SDK client already sends `artifactType` as a query parameter
- No backend changes needed — this is purely a UI presentation gap

**Expected behavior (dropdown for "Type"):**

Image

**Expected behavior (dropdown for "State"):**

Image

Fzal:
### Summary

Adds dropdown (select) support for enum-type filters in the search toolbar. 
Users can now pick artifact types and version states from a dropdown instead
of typing exact values manually.

Closes the UI gap left by #5535, which added `artifactType` filtering to the
REST API but never exposed it as a usable filter in the search page.

Fixes #8226 

### Changes

**New file: `FilterInput.tsx`** is a Thin local adapter for `ChipFilterInput` 
that conditionally renders an `ObjectSelect` dropdown when a filter type 
declares `options`, and falls back to `TextInput` otherwise. Mirrors the
original component's structure exactly (Form → InputGroup → [type selector,
value input, submit button]).

**Modified: `SearchPageToolbar.tsx`**
- `artifactType` filter now renders as a dropdown populated from 
  `GET /admin/config/artifactTypes` (with user-friendly labels like 
  "JSON Schema", "OpenAPI", etc.)
- `state` filter in version search now renders as a dropdown 
  (Enabled / Disabled / Deprecated / Draft)
- All other filters remain free-text input
- `FILTER_TYPE_LOOKUP` now covers artifact and group filter types 
  (previously only version), fixing missing chip labels

---

_No backend changes, the REST API, SDK client, and search service already 
supported `artifactType` filtering. This is purely a UI presentation fix._

### Verification

- [ ] Navigate to the Search page
- [ ] Select "Artifacts" search type → click "filter by" → select "Type" → 
      verify dropdown shows valid artifact types
- [ ] Select a type → verify results are filtered and a "Type: ..." chip appears
- [ ] Switch to "Versions" search type → verify both "Type" and "State" 
      filters render as dropdowns
- [ ] Verify text-based filters (Name, Description, Group, etc.) still 
      work as free-text input


ui/ui-app/src/app/pages/search/components/toolbar/SearchPageToolbar.tsx
```diff
@@ -88,6 +90,22 @@ VERSION_FILTER_TYPES.forEach(filterType => {
...
+
+    const [artifactTypeOptions, setArtifactTypeOptions] = useState<{ value: string; label: string }[]>([]);
+
+    const stateOptions: { value: string; label: string }[] = [
+        { value: "ENABLED", label: "Enabled" },
+        { value: "DISABLED", label: "Disabled" },
+        { value: "DEPRECATED", label: "Deprecated" },
+        { value: "DRAFT", label: "Draft" },
+    ];
+
```
carlesarnal: **Missing error handling on promise.** If `allTypesWithLabels()` fails (network error, 401, etc.), this produces an unhandled promise rejection and the dropdown silently stays empty with no feedback.

At minimum add a `.catch()` — ideally surface an error or log it:
```tsx
useEffect(() => {
    atService.allTypesWithLabels().then(types => {
        setArtifactTypeOptions(types.map((t: any) => ({ value: t.id, label: t.label })));
    }).catch(err => {
        console.error("Failed to load artifact types", err);
    });
}, []);
```

ui/ui-app/src/app/pages/search/components/toolbar/SearchPageToolbar.tsx
```diff
@@ -88,6 +90,22 @@ VERSION_FILTER_TYPES.forEach(filterType => {
...
     const config = useConfigService();
+    const atService = useArtifactTypesService();
+
+    const [artifactTypeOptions, setArtifactTypeOptions] = useState<{ value: string; label: string }[]>([]);
+
+    const stateOptions: { value: string; label: string }[] = [
+        { value: "ENABLED", label: "Enabled" },
+        { value: "DISABLED", label: "Disabled" },
+        { value: "DEPRECATED", label: "Deprecated" },
+        { value: "DRAFT", label: "Draft" },
```
carlesarnal: **Missing `useEffect` dependency.** `atService` is used inside the effect but not listed in the dependency array. With React strict mode or component re-renders this could lead to stale closures.

Either add it to the deps:
```tsx
}, [atService]);
```
or extract the service call outside the hook.

ui/ui-app/src/app/pages/search/components/toolbar/SearchPageToolbar.tsx
```diff
@@ -88,6 +90,22 @@ VERSION_FILTER_TYPES.forEach(filterType => {
...
 export const SearchPageToolbar: FunctionComponent<SearchPageToolbarProps> = (props: SearchPageToolbarProps) => {
     const config = useConfigService();
+    const atService = useArtifactTypesService();
+
+    const [artifactTypeOptions, setArtifactTypeOptions] = useState<{ value: string; label: string }[]>([]);
+
+    const stateOptions: { value: string; label: string }[] = [
+        { value: "ENABLED", label: "Enabled" },
+        { value: "DISABLED", label: "Disabled" },
+        { value: "DEPRECATED", label: "Deprecated" },
```
carlesarnal: **`stateOptions` is recreated on every render.** Since this is a static constant, move it outside the component body (next to `ARTIFACT_FILTER_TYPES` and friends) to avoid unnecessary allocations.

ui/ui-app/src/app/pages/search/components/toolbar/SearchPageToolbar.tsx
carlesarnal: Nit: `(t: any)` — consider defining an `ArtifactTypeInfo` interface (e.g. `{ id: string; label: string }`) instead of using `any`. The service itself returns `Promise<any[]>` which is also worth tightening up, but at minimum the mapping here could be typed.

carlesarnal (commented):
Good feature addressing a real UX gap. A few items to address before merging — see inline comments.

ui/ui-app/src/app/pages/search/components/toolbar/SearchPageToolbar.tsx
Fzal: Thanks for the review! 
All four items should be addressed in `d4300dd` added `.catch()`, stabilized the hook with `useRef`/`useMemo`, moved `stateOptions` to module scope, and defined an `ArtifactTypeInfo` interface.

Fzal: @carlesarnal can we merge this PR?

EricWittmann:
Hey @Fzal — apologies for the delay in getting back to you on this! All four review comments have been properly addressed and the code looks great. Thanks for the contribution! 👍

Merging now.

EricWittmann:
Thanks again @Fzal for this great contribution — really nice UX improvement! 👍

PR #8246, opened 2026-06-13
fix(test): use optimistic fallback in Debezium V2 deserializer mixin

EricWittmann:
## Summary

- Fix the `DebeziumAvroV2DeserializerMixin` to handle both 4-byte (default) and 8-byte
  (legacy) wire formats using the same heuristic as `OptimisticFallbackIdHandler`
- When the first 4 bytes after the magic byte are all zeros, treat the data as legacy
  8-byte globalId format and fetch via `/ids/globalIds/{id}`
- Otherwise, treat the 4 bytes as a contentId and fetch via `/ids/contentIds/{id}`

## Context

The `h2-debezium` integration tests (`DebeziumMySQLAvroIntegrationIT` and
`DebeziumPostgreSQLAvroIntegrationIT`) have been failing across all recent PRs. The root
cause is a wire format mismatch: the default `AvroConverter` uses `Default4ByteIdHandler`
which writes a 4-byte contentId, but the V2 deserializer mixin unconditionally read 8 bytes
as a globalId. This caused 4 bytes of Avro payload to be consumed as part of the ID,
producing invalid IDs (e.g. `42949804566`) that returned 404 from the registry.

## Test plan

- [ ] Verify the `h2-debezium` CI job passes (all 16 previously failing tests)
- [ ] Verify LocalConverters tests still pass (they use the V3 mixin, unaffected)
- [ ] Verify other integration test suites are unaffected

### DIFF
--- a/integration-tests/src/test/java/io/apicurio/tests/serdes/apicurio/debezium/DebeziumAvroV2DeserializerMixin.java
+++ b/integration-tests/src/test/java/io/apicurio/tests/serdes/apicurio/debezium/DebeziumAvroV2DeserializerMixin.java
@@ -11,6 +11,7 @@ import org.slf4j.LoggerFactory;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 
 /**
  * Mixin interface providing Apicurio Registry wire format deserialization logic with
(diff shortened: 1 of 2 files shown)


carlesarnal:
## Root cause analysis

The fix is correct, but worth noting the root cause: the V2 mixin was always written to read 8-byte globalId format, and this **used to work** because the "published converters" path (`debezium-connect.yml`) uses `quay.io/debezium/connect:latest` with `ENABLE_APICURIO_CONVERTERS=true` — which activates converters **bundled inside the Debezium image**, not the ones built from this project.

The Debezium image was shipping a V2-era Apicurio converter that used `DefaultIdHandler` (8-byte globalId). At some point recently, Debezium updated their bundled converter to V3 (`Default4ByteIdHandler`, 4-byte contentId). Since the test uses `:latest`, that switch happened silently — a new image pull brought in the V3 converter, and the V2 mixin's 8-byte read broke.

## Suggestion

Consider pinning `quay.io/debezium/connect` to a specific version tag in `debezium-connect.yml` to avoid similar silent breakages in the future.

PR #8290, opened 2026-06-17
feat(cli): add search by content command

### LINKED ISSUE #8265: CLI: Support search by content
## Description

Add support for searching artifacts by content using the CLI.

## API Endpoint

`POST /search/artifacts`

## Acceptance Criteria

- Search for artifacts by providing content (file or stdin)
- JSON and table output formats
- Support pagination and ordering

## Parent Epic

- #7162 (Phase 5)

vandanayadav7:
## Summary

Adds `acr search content` command to search artifacts by their content. Posts file content to the search API and returns matching artifacts. Supports canonicalized matching for format-independent comparison.

Fixes #8265

## Usage

```bash
# Search by exact content match
acr search content --file schema.json

# JSON output
acr search content --file schema.json -o json

# Canonical search (normalizes content before comparing — requires --type)
acr search content --file schema.json --canonical --type JSON

# Filter by group
acr search content --file schema.json -g my-group

# Read content from stdin
cat schema.json | acr search content --file -

# With pagination
acr search content --file schema.json -p 1 -s 20
```

## Behavior

| Scenario | Result |
|----------|--------|
| `--file schema.json` | Exact byte match against registry content |
| `--file schema.json --canonical --type JSON` | Normalizes content (sorts keys, strips whitespace) before comparing |
| `--canonical` without `--type` | Validation error — type is required for canonicalization |
| `--file -` | Reads content from stdin |
| No `--file` | Picocli error — required option |
| No matching content | Empty results (exit 0) |

## Changes

| File | Purpose |
|------|---------|
| `SearchByContentCommand.java` | New — `acr search content` with `--file`, `--canonical`, `--type`, `--group`, `--content-type` options |
| `SearchUtil.java` | New — extracted shared `printArtifactResults()` for table/JSON rendering (follows `RoleMappingUtil` pattern) |
| `ContentTypeDetector.java` | New — auto-detects content type from file extension (`.json`, `.avsc`, `.proto`, `.yaml`, `.xml`, etc.) |
| `FileUtils.java` | Added `readContentAsBytes()` for binary-safe file/stdin reading |
| `SearchArtifactsCommand.java` | Refactored to use `SearchUtil.printArtifactResults()` instead of inline rendering |
| `SearchGroupsCommand.java` | Added `Optional.ofNullable` null-safety on `results.getGroups()` |
| `SearchVersionsCommand.java` | Added `Optional.ofNullable` null-safety on `results.getVersions()` |
| `SearchCommand.java` | Registered `SearchByContentCommand` as subcommand, updated description |
| `SearchCommandTest.java` | 6 new tests: help, content match, group filter, no match, stdin, canonical validation, missing file |

## Test plan
- [x] `acr search content --help` shows usage
- [x] `acr search content --file schema.json` returns matching artifacts
- [x] `acr search content --file schema.json -g <group>` filters by group
- [x] `acr search content --file schema.json` with non-matching content returns empty
- [x] `acr search content --file - ` reads from stdin
- [x] `--canonical` without `--type` fails with validation error
- [x] Missing `--file` fails with picocli error
- [x] JSON and table output formats work correctly
- [x] 33 tests pass in `SearchCommandTest` (27 existing + 6 new)
- [x] Manual E2E testing on podman verified all scenarios

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/search/SearchArtifactsCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/search/SearchArtifactsCommand.java
@@ -1,33 +1,18 @@
 package io.apicurio.registry.cli.search;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import io.apicurio.registry.cli.common.AbstractCommand;
 import io.apicurio.registry.cli.common.ArtifactOrderMixin;
 import io.apicurio.registry.cli.common.OutputTypeMixin;
 import io.apicurio.registry.cli.common.PaginationMixin;
 import io.apicurio.registry.cli.utils.Conversions;
 import io.apicurio.registry.cli.utils.OutputBuffer;
-import io.apicurio.registry.cli.utils.TableBuilder;
 import io.apicurio.registry.rest.client.search.artifacts.ArtifactsRequestBuilder;
-import io.apicurio.registry.rest.v3.beans.ArtifactSearchResults;
 import java.util.List;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Mixin;
 import picocli.CommandLine.Option;
 
-import static io.apicurio.registry.cli.utils.Columns.ARTIFACT_ID;
-import static io.apicurio.registry.cli.utils.Columns.ARTIFACT_TYPE;
-import static io.apicurio.registry.cli.utils.Columns.CREATED_ON;
-import static io.apicurio.registry.cli.utils.Columns.DESCRIPTION;
-import static io.apicurio.registry.cli.utils.Columns.GROUP_ID;
-import static io.apicurio.registry.cli.utils.Columns.LABELS;
-import static io.apicurio.registry.cli.utils.Columns.MODIFIED_BY;
-import static io.apicurio.registry.cli.utils.Columns.MODIFIED_ON;
-import static io.apicurio.registry.cli.utils.Columns.NAME;
-import static io.apicurio.registry.cli.utils.Columns.OWNER;
 import static io.apicurio.registry.cli.utils.Conversions.convert;
-import static io.apicurio.registry.cli.utils.Conversions.convertToString;
-import static io.apicurio.registry.cli.utils.Mapper.MAPPER;
 
 @Command(
         name = "artifact",
@@ -100,7 +85,7 @@ public class SearchArtifactsCommand extends AbstractCommand {
             //noinspection ConstantConditions
             applyFilters(r.queryParameters);
         }));
-        printResults(output, results);
+        SearchUtil.printArtifactResults(output, results, outputType, pagination);
     }
 
     private void applyFilters(final ArtifactsRequestBuilder.GetQueryParameters params) {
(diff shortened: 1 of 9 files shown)


jsenko (commented):
## PR #8290 Review: `feat(cli): add search by content command`

### Overall Assessment

This is a well-structured PR that adds `acr search content` and extracts shared rendering into `SearchUtil`. The code follows CLI conventions and the tests are solid. I have a few issues worth addressing:

---

### Issues

**1. Null-safety inconsistency introduced in `SearchUtil` but not back-ported correctly**

`SearchUtil.java:48` wraps `results.getArtifacts()` in `Optional.ofNullable(...).orElse(List.of())`, but the original `SearchArtifactsCommand.printResults` called `results.getArtifacts().forEach(...)` directly (no null guard). This is a behavioral change to the existing `search artifact` command during a refactor — it's defensive and probably fine, but it's a silent behavior change that should be called out. 

More importantly, `SearchGroupsCommand` and `SearchVersionsCommand` still have their own inline `printResults` with the old pattern (`results.getGroups().forEach(...)` without null guard). If null-safety matters here, it should be consistent across all search commands. If it doesn't matter (because the API never returns null lists), remove the `Optional` wrapper to match the existing style.

**2. The `--content-type` default of `application/json` is too opaque**

`SearchByContentCommand.java:55-58` — The `--content-type` defaults to `application/json`. If someone searches with a Protobuf or Avro file, the default silently sends the wrong content type. The server's `searchArtifactsByContent` at `SearchResourceImpl.java:172` calls `getContentType()` to read the actual request header, so this `contentType` option is passed as the second argument to the `.post()` call and becomes the HTTP `Content-Type`. This means:

- For binary formats (Protobuf), `application/json` is wrong and could cause server-side parsing issues.
- The description says "Defaults to application/json" but doesn't warn that mismatches may cause search failures.

Consider: either auto-detect from file extension (like other CLI commands may do), or at minimum document this more clearly in the `--help` text.

**3. `--type` is ambiguous**

`SearchByContentCommand.java:41-44` uses `--type` for artifact type. But `SearchVersionsCommand` uses `--type` for the same purpose with a richer description: `"Filter by artifact type, exact match (e.g. AVRO, JSON, PROTOBUF, OPENAPI, ASYNCAPI). Use 'acr version' to see all supported types."` The new command's description just says `"Artifact type (e.g. AVRO, JSON, PROTOBUF, OPENAPI). Required when using --canonical."` — it's missing ASYNCAPI and the hint about `acr version`. Minor, but consistency matters in a CLI.

**4. Partial refactor — `SearchGroupsCommand` and `SearchVersionsCommand` still have inline `printResults`**

The PR extracts `printArtifactResults` into `SearchUtil` for `ArtifactSearchResults`, but `SearchGroupsCommand` and `SearchVersionsCommand` still have their own `printResults` methods. This is fine if the intent is incremental, but the PR description says "extracted shared `printArtifactResults()` for table/JSON rendering" which suggests the motivation is DRY. If so, `SearchUtil` should also have `printGroupResults` and `printVersionResults`. If the intent is just to share between `SearchArtifactsCommand` and `SearchByContentCommand`, that's fine — but then `SearchUtil` is arguably premature extraction for just two callers.

Not blocking, but worth a note for follow-up.

**5. Missing aliases**

`SearchGroupsCommand` has `aliases = {"groups"}` and `SearchVersionsCommand` presumably has similar plural aliases. `SearchArtifactsCommand` uses `name = "artifact"` with an alias. The new `SearchByContentCommand` has only `name = "content"` with no aliases. Consider whether `contents` or `by-content` would be useful for discoverability/consistency.

**6. No stdin test**

The command supports `--file -` for stdin (via `FileUtils.readContent`), and the PR description lists this as a feature, but there's no test for stdin reading. This is the hardest code path to test (requires redirecting `System.in`), but it's worth noting the gap.

### Nits

- `SearchByContentCommand.java:70` — The `content` String is read, then immediately converted to `byte[]` via `.getBytes(StandardCharsets.UTF_8)`, then wrapped in `ByteArrayInputStream`. Since `FileUtils.readContent` already reads the file as a String, this round-trips through String encoding. If the file is binary (e.g., Protobuf), this could corrupt the content. Consider using a `readContentAsBytes` variant that returns `byte[]` directly, or at least a `readContentAsInputStream`. This is a latent correctness issue for non-text content types.

- Test method `testSearchByContentHelp` (`SearchCommandTest.java:338`) is duplicated — `testSearchHelp` at line 46 already calls `testHelpCommand("search", "content")`. Both will pass, it's just unnecessary.

### What's Good

- Clean use of `FileUtils.readContent` for file/stdin handling
- Proper `--canonical` / `--type` validation with `CliException`
- Tests cover the important paths: match, group filter, no match, missing file, canonical validation
- Correct pagination offset calculation matching the pattern
- Proper use of `convert()` for model translation

### Verdict

Solid PR, a few things to address before merging — primarily the content-type default being silently wrong for non-JSON content (#2) and the String-to-bytes round-trip for binary content (nit). The rest are consistency/polish items.

EricWittmann (approved):
LGTM - clean implementation of the new `acr search content` command. Good use of shared utilities (SearchUtil), null-safety improvements, and solid test coverage.

EricWittmann:
👍

PR #8293, opened 2026-06-17
feat(cli): add artifact references and import/export commands

### LINKED ISSUE #8268: CLI: Support artifact references and import/export
## Description

Add support for artifact references and import/export operations in the CLI.

### Artifact References
- List outbound and inbound references for an artifact version
- Support table and JSON output formats

### Import and Export
- Export registry data to a ZIP file (full or group-scoped)
- Import registry data from a ZIP file
- Support `--no-require-empty` and `--no-preserve-ids` flags

### ~~Storage Snapshots~~ (Removed)
- ~~Use the CLI to trigger storage snapshots~~
- Removed per Eric and Carles feedback — KafkaSQL-specific, requires admin access, not worth a CLI command.

## Parent Epic

- #7162 (Phase 5)

vandanayadav7:
## Summary

Adds artifact references and import/export commands to the CLI. Admin operations are grouped under `acr admin`.

Fixes #8268

## Usage

```bash
# List outbound references for a version
acr reference list -g my-group -a my-artifact 1.0.0

# List inbound references
acr reference list --inbound -g my-group -a my-artifact 1.0.0

# Export registry data
acr admin export --file backup.zip
acr admin export --file backup.zip --group my-group

# Import registry data
acr admin import --file backup.zip --no-require-empty
acr admin import --file backup.zip --no-require-empty --no-preserve-ids
```

## Behavior

| Scenario | Result |
|----------|--------|
| `reference list -a artifact 1.0.0` | Lists outbound references in table |
| `reference list --inbound -a artifact 1.0.0` | Lists inbound references |
| `reference list -o json` | JSON array of references |
| `admin export --file backup.zip` | Exports all data to ZIP |
| `admin export --file backup.zip --group G` | Exports group-scoped data |
| `admin import --file backup.zip` | Imports into empty registry |
| `admin import --no-require-empty` | Allows import into non-empty registry |
| `admin import --no-preserve-ids` | Regenerates IDs instead of preserving originals |
| Invalid parent directory for export | Client-side validation error before server call |
| Non-existent artifact for references | Clear error via `validateArtifact()` |

## Changes

| File | Purpose |
|------|---------|
| `AdminCommand.java` | New — parent command grouping export/import under `acr admin` |
| `ExportCommand.java` | New — `acr admin export` with Vert.x HTTP client, `URI.resolve()`, output path validation, file cleanup on error, download timeout |
| `ImportCommand.java` | New — `acr admin import` with `--no-require-empty` and `--no-preserve-ids` flags |
| `ReferenceCommand.java` | New — `acr reference` parent command |
| `ReferenceListCommand.java` | New — `acr reference list` with inbound/outbound, table/JSON output, artifact validation |
| `Acr.java` | Registered `AdminCommand` and `ReferenceCommand` as subcommands |
| `Columns.java` | Added `REFERENCE_NAME` constant |
| `ReferenceCommandTest.java` | 7 tests: help, explicit reference setup via REST client, outbound table/JSON, inbound, error cases |
| `ImportExportCommandTest.java` | 11 tests: admin help, export, group export, import (empty/non-empty/no-preserve-ids), invalid directory, error cases. `@AfterAll` cleanup, `Assumptions` guards. |

## Design decisions

- **`acr admin` subcommand group**: Export/import are admin-only operations. Nesting under `acr admin` keeps the top-level namespace clean and makes the admin intent discoverable.
- **`--no-preserve-ids`** (not `--preserve-ids`): Follows the `--no-require-empty` pattern on the same command. When absent, server defaults apply (preserve=true). When present, sends explicit `false` headers.
- **Vert.x HTTP client for export download**: Uses `client.getHttpClient()` instead of `java.net.http.HttpClient`, consistent with `Update.java` and the CLI's native build configuration (JDK HTTP adapter is excluded for GraalVM compatibility).
- **Snapshot command removed**: Per Eric and Carles feedback — KafkaSQL-specific, requires admin access, not worth a CLI command.
- **Explicit references via REST client in tests**: CLI's `artifact create` doesn't support `--references`, so test setup creates artifacts with explicit references via the registry SDK for deterministic test behavior.

## Test plan
- [x] `acr admin --help` shows export/import subcommands
- [x] `acr admin export --file backup.zip` exports data (valid ZIP verified)
- [x] `acr admin export --file backup.zip --group G` exports group-scoped data
- [x] `acr admin export --file /nonexistent/dir/file.zip` fails with validation error
- [x] `acr admin import --file backup.zip --no-require-empty` imports data
- [x] `acr admin import --no-preserve-ids` regenerates IDs
- [x] `acr admin import` into non-empty registry fails with clear error
- [x] `acr reference list` shows outbound references in table and JSON
- [x] `acr reference list --inbound` shows inbound references
- [x] Non-existent artifact fails with clear validation error
- [x] 18 tests pass across ReferenceCommandTest (7), ImportExportCommandTest (11)
- [x] Manual E2E testing on podman verified all scenarios

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/Acr.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/Acr.java
@@ -1,5 +1,6 @@
 package io.apicurio.registry.cli;
 
+import io.apicurio.registry.cli.admin.AdminCommand;
 import io.apicurio.registry.cli.artifact.ArtifactCommand;
 import io.apicurio.registry.cli.auth.LoginCommand;
 import io.apicurio.registry.cli.auth.LogoutCommand;
@@ -8,6 +9,7 @@ import io.apicurio.registry.cli.config.ConfigPropertyCommand;
 import io.apicurio.registry.cli.context.ContextCommand;
 import io.apicurio.registry.cli.globalrule.GlobalRuleCommand;
 import io.apicurio.registry.cli.group.GroupCommand;
+import io.apicurio.registry.cli.reference.ReferenceCommand;
 import io.apicurio.registry.cli.rolemapping.RoleMappingCommand;
 import io.apicurio.registry.cli.search.SearchCommand;
 import io.apicurio.registry.cli.serverconfig.ServerConfigCommand;
@@ -30,6 +32,7 @@ import static picocli.CommandLine.ScopeType.INHERIT;
                 "     /_/"
         },
         subcommands = {
+                AdminCommand.class,
                 ArtifactCommand.class,
                 ConfigPropertyCommand.class,
                 ContentCommand.class,
@@ -39,6 +42,7 @@ import static picocli.CommandLine.ScopeType.INHERIT;
                 InstallCommand.class,
                 LoginCommand.class,
                 LogoutCommand.class,
+                ReferenceCommand.class,
                 RoleMappingCommand.class,
                 SearchCommand.class,
                 ServerConfigCommand.class,
--- a/cli/src/main/java/io/apicurio/registry/cli/admin/ExportCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/admin/ExportCommand.java
@@ -1,0 +1,156 @@
+package io.apicurio.registry.cli.admin;
+
+import io.apicurio.registry.cli.common.AbstractCommand;
+import io.apicurio.registry.cli.common.CliException;
+import io.apicurio.registry.cli.utils.OutputBuffer;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.http.RequestOptions;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.jboss.logging.Logger;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+
+import static io.apicurio.registry.cli.common.CliException.APPLICATION_ERROR_RETURN_CODE;
+import static io.apicurio.registry.cli.common.CliException.VALIDATION_ERROR_RETURN_CODE;
+
+@Command(
+        name = "export",
+        description = "Export registry data to a ZIP file"
+)
+public class ExportCommand extends AbstractCommand {
+
+    private static final Logger log = Logger.getLogger(ExportCommand.class);
+    private static final int DOWNLOAD_TIMEOUT_SECONDS = 300;
+    private static final int HTTP_OK = 200;
+    private static final int DEFAULT_HTTP_PORT = 80;
+    private static final int DEFAULT_HTTPS_PORT = 443;
+    private static final String HTTPS_SCHEME = "https";
+
+    @Option(
+            names = {"-f", "--file"},
+            required = true,
+            description = "Output file path for the exported ZIP archive."
+    )
+    private String file;
+
+    @Option(
+            names = {"-g", "--group"},
+            description = "Export only data belonging to this group."
+    )
+    private String groupId;
+
+    @Override
+    public void run(final OutputBuffer output) throws Exception {
+        final var outputPath = Path.of(file);
+        final var parentDir = outputPath.getParent();
+        if (parentDir != null && !Files.isDirectory(parentDir)) {
+            throw new CliException("Parent directory does not exist: " + parentDir,
+                    VALIDATION_ERROR_RETURN_CODE);
+        }
+
+        final var registryClient = client.getRegistryClient();
+
+        final var downloadRef = registryClient.admin().export().get(r -> {
+            r.queryParameters.forBrowser = true;
+            r.queryParameters.groupId = groupId;
+        });
+
+        if (downloadRef == null || downloadRef.getHref() == null) {
+            throw new CliException("Server did not return a download reference.",
+                    APPLICATION_ERROR_RETURN_CODE);
+        }
+
+        final var baseUri = resolveBaseUri();
+        final var downloadUrl = baseUri.resolve(downloadRef.getHref());
+
+        downloadFile(downloadUrl, outputPath);
+
+        output.writeStdOutChunk(out ->
+                out.append("Registry data exported to '").append(file).append("'.\n"));
+    }
+
+    private URI resolveBaseUri() {
+        final var configModel = config.read();
+        final var contextName = configModel.getCurrentContext();
+        final var context = configModel.getContext().get(contextName);
+        final var registryUrl = context.getRegistryUrl();
+        final var suffix = "/apis/registry/v3";
+        final var idx = registryUrl.indexOf(suffix);
+        if (idx >= 0) {
+            return URI.create(registryUrl.substring(0, idx));
+        }
+        return URI.create(registryUrl);
+    }
+
+    private void downloadFile(final URI url, final Path outputPath) {
+        log.debugf("Downloading export from: %s", url);
+
+        try {
+            final var httpClient = client.getHttpClient();
+            final var future = new CompletableFuture<byte[]>();
+
+            final boolean ssl = HTTPS_SCHEME.equals(url.getScheme());
+            final int defaultPort = ssl ? DEFAULT_HTTPS_PORT : DEFAULT_HTTP_PORT;
+            final int port = url.getPort() != -1 ? url.getPort() : defaultPort;
+            final var requestOptions = new RequestOptions()
+                    .setMethod(HttpMethod.GET)
+                    .setPort(port)
+                    .setHost(url.getHost())
+                    .setURI(url.getPath() + (url.getQuery() != null ? "?" + url.getQuery() : ""))
+                    .setSsl(ssl);
+
+            httpClient.request(requestOptions)
+                    .onSuccess(req -> req.send()
+                            .onSuccess(response -> {
+                                if (response.statusCode() != HTTP_OK) {
+                                    future.completeExceptionally(new CliException(
+                                            "Failed to download export: HTTP " + response.statusCode(),
+                                            APPLICATION_ERROR_RETURN_CODE));
+                                    return;
+                                }
+                                response.body()
+                                        .onSuccess(buffer -> future.complete(buffer.getBytes()))
+                                        .onFailure(future::completeExceptionally);
+                            })
+                            .onFailure(future::completeExceptionally))
+                    .onFailure(future::completeExceptionally);
+
+            final var bytes = future.get(DOWNLOAD_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+            Files.write(outputPath, bytes);
+        } catch (IOException ex) {
+            deleteQuietly(outputPath);
+            throw new CliException("Failed to write export file: " + ex.getMessage(), ex,
+                    APPLICATION_ERROR_RETURN_CODE);
+        } catch (InterruptedException ex) {
+            deleteQuietly(outputPath);
+            Thread.currentThread().interrupt();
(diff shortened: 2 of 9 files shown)


cli/src/main/java/io/apicurio/registry/cli/admin/ExportCommand.java
```diff
@@ -0,0 +1,92 @@
...
+            r.queryParameters.groupId = groupId;
+        });
+
+        if (downloadRef == null || downloadRef.getHref() == null) {
+            throw new CliException("Server did not return a download reference.",
+                    APPLICATION_ERROR_RETURN_CODE);
+        }
+
+        final var baseUrl = resolveBaseUrl();
+        final var downloadUrl = baseUrl + downloadRef.getHref();
```
EricWittmann: Consider using `URI.resolve(...)` instead of string concatenation here. With `baseUrl + href`, you could end up with doubled or malformed paths if the registry URL does not contain the expected `/apis/registry/v3` suffix (e.g. user configured just the base URL). `URI.resolve()` would handle relative vs absolute path hrefs more robustly.

cli/src/main/java/io/apicurio/registry/cli/admin/ExportCommand.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Option(
+            names = {"-g", "--group"},
+            description = "Export only data belonging to this group."
+    )
+    private String groupId;
+
+    @Override
+    public void run(final OutputBuffer output) throws Exception {
+        final var registryClient = client.getRegistryClient();
```
EricWittmann, line 57: The `--file` output path is not validated before making the server-side API call. If the parent directory does not exist or is not writable, the export will be generated on the server (consuming a download token) but the local file write will fail. Consider validating the output path upfront, similar to how `ImportCommand` validates its input file.

cli/src/main/java/io/apicurio/registry/cli/reference/ReferenceListCommand.java
```diff
@@ -0,0 +1,95 @@
...
+    public void run(final OutputBuffer output) throws Exception {
+        final var resolvedGroupId = IdUtil.resolveGroupId(groupId, config);
+        final var resolvedArtifactId = IdUtil.resolveArtifactId(artifactId, config);
+        final var registryClient = client.getRegistryClient();
+        IdUtil.validateGroup(registryClient, resolvedGroupId);
+
+        final var refType = inbound ? ReferenceType.INBOUND : ReferenceType.OUTBOUND;
+
+        final var references = Optional.ofNullable(
+                registryClient.groups().byGroupId(resolvedGroupId)
```
EricWittmann, line 68: Should `IdUtil.validateArtifact()` also be called here? Other commands that operate on a specific artifact validate both the group and the artifact before making the API call, which gives a clearer error message than a server-side 404.

cli/src/main/java/io/apicurio/registry/cli/admin/ImportCommand.java
```diff
@@ -0,0 +1,72 @@
...
+        final var registryClient = client.getRegistryClient();
+
+        try (InputStream inputStream = Files.newInputStream(filePath)) {
+            registryClient.admin().importEscaped().post(inputStream, r -> {
+                r.queryParameters.requireEmptyRegistry = !noRequireEmpty;
+                if (preserveIds) {
+                    r.headers.add(HEADER_PRESERVE_GLOBAL_ID, "true");
+                    r.headers.add(HEADER_PRESERVE_CONTENT_ID, "true");
+                }
+            });
```
EricWittmann, line 63: The `--preserve-ids` flag is effectively a no-op with the current server defaults. Here is what happens:

- When `--preserve-ids` is **absent**: no `X-Registry-Preserve-*` headers are sent. The server falls through to its config defaults (`apicurio.import.preserveGlobalId` and `apicurio.import.preserveContentId`), which are both `true`. So IDs are preserved.
- When `--preserve-ids` is **present**: headers are sent with value `"true"` — which matches the server default. IDs are preserved.

In both cases the behavior is identical. The flag never causes IDs to **not** be preserved.

To make this work as described in the PR ("Preserve global and content IDs during import"), you would need to either:
1. Always send the headers — `false` when the flag is absent, `true` when present, or
2. Rename to `--no-preserve-ids` and invert the logic (send `false` when the flag is present), which would also be consistent with the `--no-require-empty` pattern used on this same command.

cli/src/main/java/io/apicurio/registry/cli/Acr.java
```diff
@@ -31,13 +35,17 @@
                 ArtifactCommand.class,
                 ConfigPropertyCommand.class,
                 ContextCommand.class,
+                ExportCommand.class,
```
EricWittmann: `export`, `import`, and `snapshot` are admin-only operations (the PR description calls them "admin-level CLI features" and they live in the `io.apicurio.registry.cli.admin` package). Should these be nested under an `acr admin` subcommand group rather than registered as top-level commands? That would keep the top-level namespace cleaner and make the admin intent more discoverable (e.g. `acr admin export` / `acr admin import`).

cli/src/main/java/io/apicurio/registry/cli/admin/SnapshotCommand.java
```diff
@@ -0,0 +1,51 @@
...
+import picocli.CommandLine.Mixin;
+
+import static io.apicurio.registry.cli.common.CliException.APPLICATION_ERROR_RETURN_CODE;
+import static io.apicurio.registry.cli.utils.Columns.FIELD;
+import static io.apicurio.registry.cli.utils.Columns.SNAPSHOT_ID;
+import static io.apicurio.registry.cli.utils.Columns.VALUE;
+import static io.apicurio.registry.cli.utils.Mapper.MAPPER;
+
+@Command(
+        name = "snapshot",
```
EricWittmann: Is the `snapshot` command something we want in the CLI? Snapshots are a KafkaSQL-specific internal operation, and it will error on every other storage variant. Wondering if this is niche enough that it does not need a dedicated CLI command. @carlesarnal — what do you think?

cli/src/test/java/io/apicurio/registry/cli/ImportExportCommandTest.java
```diff
@@ -0,0 +1,164 @@
...
+                .as(withCliOutput("Should confirm import"))
+                .contains("imported from");
+    }
+
+    @Test
+    @Order(7)
+    public void testImportWithPreserveIds() {
+        out.getBuffer().setLength(0);
+        executeAndAssertSuccess("import", "--file", exportFile.toString(),
+                "--no-require-empty", "--preserve-ids");
```
EricWittmann: `testCleanup()` is a test method (`@Test @Order(8)`) doing teardown. If any earlier test fails or is skipped, this "test" may not run, leaving temp files behind. This should be an `@AfterAll` method instead. It also inflates the reported test count.

cli/src/test/java/io/apicurio/registry/cli/ImportExportCommandTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@QuarkusTest
+@TestMethodOrder(OrderAnnotation.class)
+public class ImportExportCommandTest extends AbstractCLITest {
+
```
EricWittmann, line 19: `exportFile` is a `static` field shared across ordered tests. If `testExport()` (Order 3) is skipped or fails, all subsequent tests that reference `exportFile` will NPE. Consider initializing this in a `@BeforeAll` method (or at minimum guarding with `Assumptions.assumeTrue(exportFile != null)` in the dependent tests).

jsenko (commented):
## PR #8293 Review: `feat(cli): add artifact references, import/export, and snapshot commands`

### Overall Assessment

This is the largest of the CLI PRs and covers significant ground: admin operations (export/import/snapshot) and artifact references. All previous Copilot comments (10) and Eric's comments (8) appear to be addressed:

- **`--preserve-ids` → `--no-preserve-ids`** with correct inverted logic ✓
- **`URI.resolve()` instead of string concatenation** for download URL ✓
- **File cleanup on error** via `deleteQuietly()` in all error paths ✓
- **Vert.x HTTP client** instead of standalone `java.net.http.HttpClient` ✓
- **Output path validation** before making the export API call ✓
- **`IdUtil.validateArtifact()`** added to `ReferenceListCommand` ✓
- **`acr admin` subcommand group** for export/import/snapshot ✓
- **`@AfterAll` cleanup** and `Assumptions` guards in tests ✓
- **Explicit references via REST client** in test setup ✓
- **Stronger assertions** in reference and snapshot tests ✓

Content-Type for import is handled automatically by the Kiota client (`application/zip`), so that Copilot concern was a non-issue.

---

### Issues

**1. Export loads entire ZIP into memory before writing to disk**

`ExportCommand.java:186` — `response.body().onSuccess(buffer -> future.complete(buffer.getBytes()))` reads the entire export into a byte array, then writes it all at once via `Files.write()`. For large registries with many artifacts, this could cause OOM. Consider streaming the response body directly to disk using Vert.x's `pipeTo()` or `AsyncFile`. Not blocking for a v1, but worth a `// TODO` or a follow-up issue for large-registry support.

**2. `resolveBaseUri()` assumes a specific URL structure**

`ExportCommand.java:147-158` — The method strips `/apis/registry/v3` from the registry URL to construct the download base. This works for the standard deployment but will break if:
- The registry is behind a reverse proxy with a different path prefix
- The API version changes

The PR description notes this follows the `Update.java` pattern, so it's a known limitation. Just flagging for awareness.

**3. Eric's open question about `snapshot` is still unresolved**

Eric asked (@carlesarnal) whether the snapshot command should exist in the CLI since it's KafkaSQL-specific and will error on every other storage variant. The command is still present. This needs a decision before merge — either keep it (it's useful for KafkaSQL operators) or defer it. The implementation itself is fine.

**4. No test for `testExportInvalidDirectory`**

The code validates that the parent directory exists before making the export API call (`ExportCommand.java:121-124`), but there's no test exercising this validation path. There's `testExportMissingFile` which tests missing `--file` argument (picocli error), but not the "parent directory doesn't exist" case.

### Nits

- `ExportCommand` has both `DOWNLOAD_TIMEOUT_SECONDS`, `HTTP_OK`, `DEFAULT_HTTP_PORT`, `DEFAULT_HTTPS_PORT`, and `HTTPS_SCHEME` as constants. These are standard values that could use inline literals, but constants are fine too — just verbose for a single-use class.

- `SnapshotCommandTest` test names `testSnapshotOnSqlStorageFails` assert `isNotEqualTo(0)` — this correctly handles the known environment (SQL storage in tests), but the test name and assertion would be wrong if tests ever run against KafkaSQL. Fine for now since the test suite runs against SQL.

### What's Good

- Clean `AdminCommand` parent with export/import/snapshot — good namespace organization
- Vert.x HTTP client usage with proper SSL/port handling and timeout
- `deleteQuietly()` for file cleanup on all error paths (IOException, InterruptedException, TimeoutException, generic Exception)
- Thread interrupt preservation in the InterruptedException handler
- `CompletableFuture` unwrapping of `CliException` from `ExecutionException` cause
- Reference test setup via REST client with explicit `ArtifactReference` — deterministic and robust
- `Assumptions.assumeTrue` guards on import tests that depend on export test
- Both inbound and outbound reference listing tested

cli/src/main/java/io/apicurio/registry/cli/admin/SnapshotCommand.java
carlesarnal: And when authz is enabled IIRC it requires admin access. I don't think it's worth it to have this in the cli.

vandanayadav7: Agreed — snapshot command removed. PR updated.

EricWittmann (approved):

PR #8336, opened 2026-06-22
feat(rest): expose canonicalization endpoints

### LINKED ISSUE #2473: Expose canonicalization endpoints
We have a use case where it would be useful to be able to compare a version of a JSON schema that our producer uses, to the version that has been registered in Apicurio. While there are a couple of endpoints that indicate it uses a canonicalized/protocol specific version of the schema, there is no endpoint that allows me to `GET` the canonicalized schema from an arbitrarily provided schema.

Can this be considered from an enhancement?

carlesarnal:
## Summary

Expose schema canonicalization as a first-class REST API feature. Users can now retrieve the canonical form of schemas for comparison purposes — both for arbitrary content and for stored artifact versions.

Fixes #2473

## Root Cause

Canonicalization was used only internally for content hash matching (e.g. the `canonical` query parameter on `POST /groups/{groupId}/artifacts` for `ifExists` logic). There was no way to GET the canonicalized form of a schema, which users need for comparing schemas produced by different tools or environments.

## Changes

- **`common/src/main/resources/META-INF/openapi.json`**: Added `POST /system/canonicalize` endpoint with required `artifactType` query parameter. Added `canonical` boolean query parameter to `GET /groups/{groupId}/artifacts/{artifactId}/versions/{versionExpression}/content`.
- **`app/.../rest/v3/impl/SystemResourceImpl.java`**: Implemented `canonicalizeContent()` — reads the request body, delegates to `RegistryStorageContentUtils.canonicalizeContent()`, and returns the canonical form. Auth level: `Read`.
- **`app/.../rest/v3/impl/GroupsResourceImpl.java`**: Extended `getArtifactVersionContent()` with the new `Boolean canonical` parameter. When `true`, resolves references from storage and canonicalizes the content before returning it.
- **`app/src/test/.../noprofile/rest/v3/SystemResourceTest.java`**: Added 6 new tests covering JSON, OpenAPI, and Avro canonicalization; empty body error case; stored content canonical retrieval; and a regression test confirming non-canonical retrieval is unchanged.

## Test plan

- [x] `POST /system/canonicalize` with JSON content (keys reordered)
- [x] `POST /system/canonicalize` with OpenAPI content
- [x] `POST /system/canonicalize` with Avro content
- [x] `POST /system/canonicalize` with empty body → 400
- [x] `GET .../content?canonical=true` for stored JSON artifact
- [x] `GET .../content` without canonical param (regression — unchanged behavior)
- [x] Checkstyle passes with 0 violations
- [ ] SDK updates (java-sdk, go-sdk, python-sdk, typescript-sdk) — tracked as follow-up


carlesarnal:
@EricWittmann any other better place than system for the endpoint to given a content, return it canonicalized? admin? anything else?

EricWittmann:
I don't think I would use `/system` for this.  Let me think about other new endpoints we can use.

EricWittmann:
I think the right home for this is `POST /content/canonicalize` rather than `/system`. There's already a `/content` resource (`ContentResource`) that handles content-level operations detached from any specific artifact (e.g., `/content/references`). Canonicalization fits that same pattern — it's a stateless content utility, not a system-level operation.

```
POST /apis/registry/v3/content/canonicalize?artifactType=AVRO
```

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -944,6 +945,14 @@ public Response getArtifactVersionContent(String groupId, String artifactId, Str
         contentToReturn = handleContentReferences(references, metaData.getArtifactType(), contentToReturn,
                 artifactCell.get().getReferences());
 
+        if (Boolean.TRUE.equals(canonical)) {
+            Map<String, TypedContent> resolvedReferences = RegistryContentUtils
+                    .recursivelyResolveReferences(artifactCell.get().getReferences(),
+                            storage::getContentByReference);
+            contentToReturn = contentUtils.canonicalizeContent(metaData.getArtifactType(),
```
EricWittmann, line 952: Canonicalization runs after `handleContentReferences()` here — so the canonical output depends on the `references` query parameter. Do we need a test that exercises both at the same time (e.g. `?references=DEREFERENCE&canonical=true`) to make sure the interaction produces the expected result?

app/src/main/java/io/apicurio/registry/rest/v3/impl/SystemResourceImpl.java
```diff
@@ -63,6 +77,20 @@ public SystemInfo getSystemInfo() {
         return info;
     }
 
+    @Override
+    @Authorized(style = AuthorizedStyle.None, level = AuthorizedLevel.Read)
+    public Response canonicalizeContent(String artifactType, InputStream data) {
```
EricWittmann: There is no validation that `artifactType` is a known type before passing it to `canonicalizeContent()`. If someone sends `artifactType=GARBAGE`, the behavior depends on how the artifact type provider lookup handles unknown types. Should we validate early and return a clear 400?

app/src/main/java/io/apicurio/registry/rest/v3/impl/SystemResourceImpl.java
```diff
@@ -63,6 +77,20 @@ public SystemInfo getSystemInfo() {
...
+        ContentHandle content = ContentHandle.create(data);
+        if (content.bytes().length == 0) {
+            throw new BadRequestException("Empty content is not allowed.");
+        }
```
EricWittmann: Using `@Context HttpServletRequest` to get the content type feels inconsistent with the rest of the codebase — other endpoints get it from the generated JAX-RS interface (e.g. `@HeaderParam` or `@Consumes`). This also couples the implementation to the Servlet API. Could the content type come through the generated interface instead?

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -895,18 +899,16 @@ public void deleteArtifactRule(String groupId, String artifactId, RuleType rule)
...
     public Response getArtifactVersionContent(String groupId, String artifactId, String versionExpression,
-            HandleReferencesType references) {
+            HandleReferencesType references, Boolean canonical) {
         ParameterValidationUtils.requireParameter("groupId", groupId);
         ParameterValidationUtils.requireParameter("artifactId", artifactId);
         ParameterValidationUtils.requireParameter("versionExpression", versionExpression);
 
         if (references == null) {
-            // Check if admin has configured a default reference handling behavior
             java.util.Optional<String> configuredDefault = restConfig.getDefaultReferenceHandling();
```
EricWittmann, line 908: A few existing comments were removed in this method (the "Check if admin has configured..." comment here, the "No configuration..." comment below, and the "Throw 404..." comment further down). Was this intentional cleanup or accidental?

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SystemResourceTest.java
```diff
@@ -20,4 +22,93 @@ public void testSystemInformation() {
...
+                .queryParam("artifactType", ArtifactType.OPENAPI)
+                .body(content)
+                .post("/registry/v3/system/canonicalize")
+                .then().statusCode(200)
+                .body("components", notNullValue())
+                .body("openapi", equalTo("3.0.2"));
+    }
+
+    @Test
+    public void testCanonicalizeAvroContent() {
```
EricWittmann: The OpenAPI and Avro tests only assert that certain fields exist in the response (e.g. `components` is not null, `openapi` equals `3.0.2`) — they don't verify that canonicalization actually happened (e.g. keys are sorted, whitespace is normalized). The JSON test does this well with an exact string comparison. Should these tests have similarly strong assertions to confirm the output is actually in canonical form?

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
carlesarnal: Accidental for sure.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SystemResourceTest.java
carlesarnal: Yes, they must do it.

carlesarnal:
@EricWittmann ready for another round

app/src/main/java/io/apicurio/registry/rest/v3/impl/ContentResourceImpl.java
```diff
@@ -64,4 +77,21 @@ public List<ArtifactReference> detectContentReferences(String artifactType, Vers
...
+        }
+        if (!factory.getAllArtifactTypes().contains(artifactType)) {
+            throw new BadRequestException("Unknown artifact type: " + artifactType);
+        }
+        String ct = httpHeaders.getMediaType() != null ? httpHeaders.getMediaType().toString() : null;
+        TypedContent typedContent = TypedContent.create(content, ct);
+        TypedContent canonicalized = contentUtils.canonicalizeContent(artifactType, typedContent,
+                Map.of());
+        return Response.ok(canonicalized.getContent()).type(canonicalized.getContentType()).build();
+    }
```
paoloantinori, line 96: Passing `Map.of()` here means the standalone canonicalize endpoint can't resolve `$ref`s for OpenAPI/JSON Schema/AsyncAPI content, while the `GET .../content?canonical=true` path does dereference. Worth either documenting the asymmetry or accepting a references payload (the sibling `detectContentReferences` already does).

app/src/main/java/io/apicurio/registry/rest/v3/impl/ContentResourceImpl.java
```diff
@@ -13,15 +13,22 @@
 import io.apicurio.registry.rest.v3.ContentResource;
 import io.apicurio.registry.rest.v3.beans.ArtifactReference;
 import io.apicurio.registry.rest.v3.beans.VersionContent;
+import io.apicurio.registry.storage.impl.sql.RegistryStorageContentUtils;
 import io.apicurio.registry.types.provider.ArtifactTypeUtilProvider;
 import io.apicurio.registry.types.provider.ArtifactTypeUtilProviderFactory;
 import io.apicurio.registry.util.ArtifactTypeUtil;
```
paoloantinori, line 19: `RegistryStorageContentUtils` lives under `storage.impl.sql`. Worth confirming the bean is available in the kafkasql/gitops/inmemory profiles — the earlier Impact Analysis comment flagged exactly this. If it's profile-agnostic, moving it out of the SQL package would make that explicit and avoid a startup failure in the non-SQL builds.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -944,6 +945,14 @@ public Response getArtifactVersionContent(String groupId, String artifactId, Str
         contentToReturn = handleContentReferences(references, metaData.getArtifactType(), contentToReturn,
                 artifactCell.get().getReferences());
 
+        if (Boolean.TRUE.equals(canonical)) {
+            Map<String, TypedContent> resolvedReferences = RegistryContentUtils
+                    .recursivelyResolveReferences(artifactCell.get().getReferences(),
```
paoloantinori, line 950: When `canonical=true`, references are dereferenced unconditionally via `recursivelyResolveReferences`, independent of the `references` query parameter value. So `references=PRESERVE&canonical=true` still inlines refs during canonicalization. Probably intentional, but the OpenAPI description of the new `canonical` param doesn't mention it — a one-line note would help callers reason about the combined behavior.

paoloantinori (commented):
Solid feature — the `POST /content/canonicalize` endpoint and the `canonical` query param on version content hang together well, and the test coverage (JSON / OpenAPI / Avro / dereferenced / error cases) is thorough. A few observations worth a closer look:

- **Storage-abstraction leak (re-surfacing the concern from the Impact Analysis comment).** Both `ContentResourceImpl` and `GroupsResourceImpl` now `@Inject` `io.apicurio.registry.storage.impl.sql.RegistryStorageContentUtils`. The package path places it under the SQL storage implementation; worth confirming the bean is resolvable in the kafkasql, gitops, and inmemory profiles before this merges, otherwise those builds could fail at startup (or, worse, silently behave differently). If the util is genuinely profile-agnostic, moving it out of `storage.impl.sql` would make that explicit.
- **Interaction between `canonical=true` and `references` is non-obvious.** In `getArtifactVersionContent`, when `canonical=true` the code always calls `recursivelyResolveReferences` against the *original* `artifactCell` references and feeds them to `canonicalizeContent`, regardless of what the `references` query parameter was set to. So a caller passing `references=PRESERVE&canonical=true` will still get refs dereferenced during canonicalization. That may be intentional (canonicalizers need resolved refs to inline `$ref`s), but the OpenAPI description for the new `canonical` param doesn't mention the interaction. Either a doc note or a sanity check (e.g. only dereference when `references` allows it) would help.
- **`/content/canonicalize` doesn't accept references at all** — it passes `Map.of()` into `canonicalizeContent`. For artifact types whose canonicalizer depends on resolving external `$ref`s (OpenAPI/JSON Schema/AsyncAPI), the standalone endpoint can't produce a fully-canonical form the way the version-content endpoint can. Worth either documenting that limitation or accepting a references payload in the request (mirroring `detectContentReferences`).

Minor: the unrelated alphabetical import reorder in `SystemResourceImpl.java` is fine, just flagging it so it's clear that hunk isn't part of the feature.

PR #8349, opened 2026-06-23
[CI Reliability] Replace ttl.sh with workflow artifacts for operator test images

### LINKED ISSUE #8348: fix(ci): replace ttl.sh with workflow artifacts for operator test images
## Problem

Operator tests push images to `ttl.sh` (a free, zero-SLA ephemeral registry) in the build job, then 8 parallel test shards on separate runners pull them. `ttl.sh` has no uptime guarantees — slow pulls or outages cause operator pods to fail readiness checks, breaking CI.

**Current failure rate on `main`: ~25%** (2 failures in 8 completed runs). Every failure is `ConditionTimeoutException` waiting for operator pod readiness — `ContainersNotReady` because image pulls from `ttl.sh` are slow or fail.

Affected test shards (all with identical root cause):
- `Operator Tests / Remote Tests (feature)`
- `Operator Tests / Remote Tests (database)`
- `Operator Tests / Remote Tests (kafka)` — all 4 kafka tests (`KafkaSqlITTest`, `KafkaSqlAccessITTest`, `KafkaSqlOAuthITTest`, `KafkaSqlTLSITTest`)
- `Operator Tests / Local Tests` — `SmokeITTest.testEmptyHostDisablesIngress`

## History of ttl.sh in this project

The project has tried to move away from `ttl.sh` before:

1. **Jun 2025** — `ttl.sh` push failures started hitting. Retries added:
   - 256d591c `fix(ttl.sh): Attempt to fix issues with pushing images to ttl.sh` (#6282)
   - 059fb93c `Retry the push to ttl.sh` (#6363)

2. **Jul 2025** — Migrated non-operator workflows to `ghcr.io`:
   - 18d7ed20 `chore(ci): Stop using ttl.sh and use ghcr instead` (#6387)
   - **But `operator.yaml` was never touched** — the operator stayed on `ttl.sh`

3. **Jul 2025** — Discovered `ghcr.io` doesn't work for fork PRs (`GITHUB_TOKEN` lacks `packages:write`):
   - PR #6420 (closed): `chore(ci): Remove ghcr due to issues with secrets`
   - d79ad282 replaced ghcr with temporary quay.io repo (#6426)
   - 3dc6ee82 replaced quay.io with **workflow artifacts** (`docker save` → upload → `docker load`) (#6432)

4. **Mar 2026** — Operator tests parallelized into 8 matrix shards (#7487). Since each shard deploys pods that **pull images from a registry**, `ttl.sh` was kept as the only option that works across runners without auth.

## Why the operator can't use the same workflow artifacts approach

The non-operator tests use `docker save` → workflow artifact → `docker load` + `minikube image load` + `imagePullPolicy: Never`. This works because the test code controls the deployment.

The operator tests can't do this because:
- The operator **controller** deploys pods via the Kubernetes API
- Manifests use `imagePullPolicy: Always` (required for OLM)
- Kubernetes tries to **pull from a registry**, ignoring pre-loaded images

## Proposed solution

**Workflow artifacts + `minikube image load` + `imagePullPolicy` override in tests:**

1. Build job: `docker save` operator/operand images → upload as workflow artifacts
2. Each test shard: download artifact → `docker load` → `minikube image load`
3. Override `imagePullPolicy` to `IfNotPresent` in test mode (via system property or test config)

This is the same pattern the non-operator tests already use — the only new piece is patching the operator's `imagePullPolicy` for CI.

**Fork-safe**: workflow artifacts work for fork PRs (unlike ghcr.io).

## Related

- #6282 — first ttl.sh reliability issues
- #6363 — ttl.sh push retries
- #6387 — ghcr migration (non-operator only)
- #6420 — ghcr for operator (closed — fork limitation)
- #6426 — quay.io workaround
- #6432 — workflow artifacts (current non-operator solution)
- #7487 — operator test parallelization (kept ttl.sh)

paoloantinori:
## Summary

Replace `ttl.sh` with workflow artifacts (`docker save`/`docker load`) for 3 of 5 operator test images, eliminating the primary cause of CI flakes (~25% failure rate on `main`).

Closes #8348

## Problem

Operator tests push images to `ttl.sh` (a free, zero-SLA ephemeral registry). Test shards on separate runners pull them via Kubernetes. When `ttl.sh` is slow, pods fail readiness checks within the awaitility timeout, causing `ConditionTimeoutException`.

This has been a recurring problem — the project has tried `ghcr.io` (#6387, #6420), `quay.io` (#6426), and workflow artifacts (#6432) for non-operator workflows. The operator was never migrated because its test shards deploy pods that pull from a registry.

## Solution

**Key insight**: Minikube uses `driver: none`, which means the Docker daemon on the runner IS the Minikube Docker daemon. `docker load` makes images immediately available to Kubernetes without a registry pull. The test code already patches `imagePullPolicy` to `IfNotPresent` on Minikube (`ITBase.createTestResources` line 283).

### What changed

| Image | Before | After |
|-------|--------|-------|
| Operand (registry app) | `ttl.sh` push/pull | `docker save` → artifact → `docker load` |
| GitOps sync sidecar | `ttl.sh` push/pull | `docker save` → artifact → `docker load` |
| Operator | `ttl.sh` push/pull | `docker save` → artifact → `docker load` |
| OLM Bundle | `ttl.sh` push/pull | **Unchanged** — OLM pulls directly |
| OLM Catalog | `ttl.sh` push/pull | **Unchanged** — OLM pulls directly |

### What's still on ttl.sh

OLM bundle and catalog images stay on `ttl.sh` because OLM itself needs to pull them from a registry — this only affects 2 of 8 test shards (olm-v0, olm-v1).

## Testing

- [ ] All 6 non-OLM operator test shards pass (smoke, kafka, auth, database, feature, feature-setup)
- [ ] OLM test shards pass (olm-v0, olm-v1) — still use ttl.sh for bundle/catalog
- [ ] `operator-check-install` passes — install file unaffected
- [ ] Non-operator CI jobs unaffected

## Related

- #6282 — first ttl.sh reliability issues
- #6387 — ghcr migration (non-operator only)
- #6420 — ghcr for operator (closed — fork PR limitation)
- #6432 — workflow artifacts for non-operator (current approach)
- #7487 — operator test parallelization (kept ttl.sh)

### DIFF
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -11,9 +11,11 @@ permissions:
   contents: read
 
 env:
-  REGISTRY_APP_IMAGE: ttl.sh/apicurio-registry-${{ github.sha }}:8h
-  REGISTRY_GITOPS_SYNC_IMAGE: ttl.sh/apicurio-registry-gitops-sync-${{ github.sha }}:8h
-  IMAGE: ttl.sh/apicurio-registry-3-operator-${{ github.sha }}:8h
+  # Non-OLM test shards use workflow artifacts + minikube image load (no registry needed).
+  # OLM shards still need ttl.sh because OLM itself pulls bundle/catalog images.
+  REGISTRY_APP_IMAGE: localhost/apicurio-registry:${{ github.sha }}
+  REGISTRY_GITOPS_SYNC_IMAGE: localhost/apicurio-registry-gitops-sync:${{ github.sha }}
+  IMAGE: localhost/apicurio-registry-3-operator:${{ github.sha }}
   BUNDLE_IMAGE: ttl.sh/apicurio-registry-3-operator-bundle-${{ github.sha }}:8h
   CATALOG_IMAGE: ttl.sh/apicurio-registry-3-operator-catalog-${{ github.sha }}:8h
 
@@ -39,30 +41,27 @@ jobs:
       - name: Build
         run: ./mvnw -B clean package -Dfull -pl app,distro/docker,operator/olm-tests -am -Pprod -DskipTests -Dmaven.javadoc.skip
 
-      # We need to push the operand image to ttl.sh so test matrix jobs on separate runners can pull it.
-      - name: Build and push temporary operand image
+      - name: Build operand image
         run: |
           docker build -f ./distro/docker/target/docker/Dockerfile.jvm -t "$REGISTRY_APP_IMAGE" ./distro/docker/target/docker
-          docker push "$REGISTRY_APP_IMAGE"
 
-      - name: Build and push temporary GitOps sync sidecar image
+      - name: Build GitOps sync sidecar image
         run: |
           docker build -t "$REGISTRY_GITOPS_SYNC_IMAGE" ./distro/gitops
-          docker push "$REGISTRY_GITOPS_SYNC_IMAGE"
 
-      # We need to push the bundle image because opm always pulls; the catalog image because OLM
-      # deploys it with `imagePullPolicy: Always`; and also the operator image itself for the same reason.
-      - name: Build and push temporary operator image
+      - name: Build operator image
         working-directory: operator
         run: |
-          make image-build image-push
+          make image-build
 
-      - name: Build and push temporary operator bundle
+      # OLM bundle and catalog must be pushed to a registry because OLM pulls them directly.
+      # ttl.sh is kept only for these two images.
+      - name: Build and push OLM bundle (ttl.sh)
         working-directory: operator
         run: |
           make ROLLING_CHANNEL=3.x bundle
 
-      - name: Build and push temporary operator catalog
+      - name: Build and push OLM catalog (ttl.sh)
         working-directory: operator
         run: |
           make ROLLING_CHANNEL=3.x catalog
@@ -71,6 +70,21 @@ jobs:
         working-directory: operator
         run: |
           make INSTALL_FILE=controller/target/test-install.yaml dist-install-file
+
+      # Save Docker images as artifacts so test shards can load them without a registry.
+      # This replaces ttl.sh for the operand, operator, and gitops-sync images.
+      - name: Save Docker images
+        run: |
+          docker save "$REGISTRY_APP_IMAGE" -o /tmp/operator-test-registry-app.tar
+          docker save "$REGISTRY_GITOPS_SYNC_IMAGE" -o /tmp/operator-test-gitops-sync.tar
+          docker save "$IMAGE" -o /tmp/operator-test-operator.tar
+
+      - name: Upload Docker images
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: operator-test-images
+          path: /tmp/operator-test-*.tar
+          retention-days: 1
 
       - name: Upload build artifacts
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
@@ -180,6 +194,12 @@ jobs:
           name: operator-build-artifacts
           path: operator/
 
+      - name: Download Docker images
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          name: operator-test-images
+          path: /tmp/
+
       - uses: apicurio/apicurio-github-actions/setup-minikube@97d452c41f1abd7142c0ce2ac37e87922e35e71d # v2
         with:
           driver: 'none'
@@ -187,6 +207,14 @@ jobs:
           olm-enable: ${{ matrix.olm-enable }}
           olm-v1-enable: ${{ matrix.olm-v1-enable }}
           start-args: '--alsologtostderr --v=2'
+
+      # Load pre-built images into the Docker daemon (shared with Minikube via driver: none).
+      # This eliminates ttl.sh dependency for operand, operator, and gitops-sync images.
+      - name: Load Docker images
+        run: |
+          docker load -i /tmp/operator-test-registry-app.tar
+          docker load -i /tmp/operator-test-gitops-sync.tar
+          docker load -i /tmp/operator-test-operator.tar
 
       - name: Run ${{ matrix.name }} tests on Minikube
         working-directory: operator
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -214,8 +214,12 @@ jobs:
       - name: Start local registry
         run: |
           docker run -d -p 5000:5000 --restart=always ghcr.io/distribution/distribution:3.0
-          sleep 2
-          curl -sf http://localhost:5000/v2/ || (echo "Registry not ready"; exit 1)
+          for i in $(seq 1 20); do
+            curl -sf http://localhost:5000/v2/ >/dev/null 2>&1 && break
+            echo "Waiting for registry... ($i/20)"
+            sleep 0.5
+          done
+          curl -sf http://localhost:5000/v2/ || (echo "Registry not ready after 10s"; exit 1)
 
       # Load pre-built images into Docker, then push to the local registry.
       # Operator/operand/gitops-sync use localhost:5000; bundle/catalog use ttl.sh.
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -131,14 +131,6 @@ jobs:
     name: Remote Tests (${{ matrix.name }})
     runs-on: ubuntu-latest
     needs: operator-build
-    # OLM shards need a local registry so Kubernetes/OLM can pull bundle/catalog images.
-    # Non-OLM shards use docker load + imagePullPolicy: IfNotPresent (no registry needed,
-    # but we start one anyway for simplicity — all shards use the same image refs).
-    services:
-      registry:
-        image: ghcr.io/distribution/distribution:3.0
-        ports:
-          - 5000:5000
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/operator.yaml
```diff
@@ -11,9 +11,11 @@ permissions:
   contents: read
 
 env:
-  REGISTRY_APP_IMAGE: ttl.sh/apicurio-registry-${{ github.sha }}:8h
-  REGISTRY_GITOPS_SYNC_IMAGE: ttl.sh/apicurio-registry-gitops-sync-${{ github.sha }}:8h
-  IMAGE: ttl.sh/apicurio-registry-3-operator-${{ github.sha }}:8h
+  # Non-OLM test shards use workflow artifacts + minikube image load (no registry needed).
+  # OLM shards still need ttl.sh because OLM itself pulls bundle/catalog images.
+  REGISTRY_APP_IMAGE: localhost/apicurio-registry:${{ github.sha }}
```
EricWittmann: The `operator-test-local` job (line 110-113) also runs on Minikube with `driver: none` and does a full `make build` which includes `image-build image-push` via the `image` target. Since it builds locally on the same runner, the `docker push` to `ttl.sh` is unnecessary overhead — the image is already in the local daemon. It won't break because `ttl.sh` still works, but it's an inconsistency with the rest of this PR. Worth a follow-up to drop the push there too?

.github/workflows/operator.yaml
```diff
@@ -188,6 +208,14 @@ jobs:
           olm-v1-enable: ${{ matrix.olm-v1-enable }}
           start-args: '--alsologtostderr --v=2'
 
+      # Load pre-built images into the Docker daemon (shared with Minikube via driver: none).
+      # This eliminates ttl.sh dependency for operand, operator, and gitops-sync images.
+      - name: Load Docker images
+        run: |
+          docker load -i /tmp/operator-test-registry-app.tar
+          docker load -i /tmp/operator-test-gitops-sync.tar
```
EricWittmann, line 230: Nit: this step runs on all 8 matrix shards including `olm-v0` and `olm-v1`, which don't need these images (OLM pulls its own bundle/catalog from `ttl.sh`). It's harmless but wastes ~5-10 seconds on those shards. An `if: matrix.olm-enable == 'false'` guard could skip it for OLM shards, though the savings are marginal.

paoloantinori: Good catch — applied. Split the load step: operator image loads on all shards, operand/gitops-sync only on non-OLM shards (`if: matrix.olm-enable == 'false'`).

.github/workflows/operator.yaml
paoloantinori: Agreed — the `operator-test-local` job does a full `make build` which includes `image-push` to ttl.sh. Since that job runs with `driver: none`, the push is unnecessary overhead — the image is already in the local daemon.

Fixing this properly requires changing the Makefile to separate `build` from `image-push`, which is out of scope for this PR. I'll create a follow-up task to address it.

EricWittmann:
## CI Failure Analysis

All 8 operator test jobs are failing because the operator pod container never becomes ready — it's stuck in `ContainersNotReady`.

### Root Cause

The PR changes image references from `ttl.sh/...` (a real ephemeral registry) to `localhost/...` (just a local Docker tag). The images are saved as workflow artifacts, downloaded in test shards, and loaded into Docker via `docker load`. Since Minikube uses `driver: none`, Docker is shared between the host and Kubernetes, so the images *are* present in Docker's local cache.

**However**, the operator's deployment template at `operator/controller/src/main/deploy/operator/default/operator.yaml:13` has:

```yaml
imagePullPolicy: Always
```

This forces Kubernetes to **always attempt to pull the image from a registry**, even when it exists locally. Since `localhost/apicurio-registry-3-operator:<sha>` isn't served by any actual container registry, the pull fails, and the operator pod stays in `ContainersNotReady`.

The same issue would apply to the operand images (registry app and gitops-sync) that the operator deploys — they would also fail to pull.

### Suggested Fixes

1. **Patch `imagePullPolicy` to `IfNotPresent` or `Never`** in the install YAML before applying it to the cluster in the non-OLM test shards. This is probably the simplest approach.
2. **Run a local registry** (e.g., `registry:2` on `localhost:5000`), push images there, and reference `localhost:5000/...` — which *is* an actual pullable registry.
3. **Keep `ttl.sh`** for the operator image (revert the core change for at least that image).

paoloantinori:
Thanks for the detailed analysis. You're right that `imagePullPolicy: Always` + `localhost/...` (no actual registry) is the root cause.

## Root Cause Verified

`application.properties` line 64 sets:
```
%test.test.operator.deployment-target=kubernetes
```

`ITBase.createTestResources()` (line 283-287) has code designed to fix exactly this:
```java
if ("minikube".equals(deploymentTarget) && r instanceof Deployment d) {
    d.getSpec().getTemplate().getSpec().getContainers()
            .forEach(c -> c.setImagePullPolicy("IfNotPresent"));
}
```

But it **never fires** because `deploymentTarget` is `kubernetes`, not `minikube`. The `imagePullPolicy` patch was always dead code in CI.

**Fix**: pass `-Dtest.operator.deployment-target=minikube` in the CI workflow's Maven command. This enables the existing Java code that was designed for this scenario.

## Revised Proposal: Eliminate ttl.sh Entirely

### Non-OLM shards (6 of 8) — `docker load` + enable existing `imagePullPolicy` patch

Add `-Dtest.operator.deployment-target=minikube` to the Maven test command. This makes `ITBase.createTestResources()` patch `imagePullPolicy` to `IfNotPresent` for all Deployments, including the operator pod.

Combined with the existing `docker load` approach, Kubernetes finds the pre-loaded images locally and doesn't attempt a registry pull.

**Zero overhead. No external dependency. No registry needed. Uses existing Java code.**

### OLM shards (2 of 8) — local `registry:2` for bundle/catalog

OLM deploys via catalog → bundle → CSV. It MUST pull these from a real registry. Instead of ttl.sh, we spin up a local registry:

```yaml
services:
  registry:
    image: registry:2
    ports: ["5000:5000"]
```

- Build job: `docker save` bundle + catalog images to artifacts (same pattern as operator/operand)
- OLM shards: `docker load` + `docker push localhost:5000/...`
- OLM pulls from `localhost:5000` — a real registry, fast, no external dependency

**~10 seconds overhead on 2 shards only.**

### Why not a full local registry for ALL shards?

We considered running `registry:2` on all 8 shards, but:
- Pushing 5 images (~1-2GB) to the registry adds ~30-60s per shard (disk I/O)
- `registry:2` itself needs a Docker Hub pull on ephemeral runners — swapping one external dependency for another
- Non-OLM shards don't need a registry at all — `docker load` + `IfNotPresent` is simpler and faster

### Summary

| Shards | Images | Strategy | ttl.sh? | Overhead |
|--------|--------|----------|---------|----------|
| 6 non-OLM | operator, operand, gitops-sync | `docker load` + `-Dtest.operator.deployment-target=minikube` | None | ~0s |
| 2 OLM | bundle, catalog (+ operator, operand) | `docker load` + local `registry:2` for bundle/catalog | None | ~10s |

Total ttl.sh references: **zero**.

What do you think?

EricWittmann:
Great analysis, Paolo. The discovery that `deploymentTarget` is set to `kubernetes` (not `minikube`) in `application.properties` — making the `imagePullPolicy` patch dead code in CI — is a really nice catch. That was a wrong assumption in the original PR description.

The revised proposal looks solid. Using `-Dtest.operator.deployment-target=minikube` to activate existing code for the 6 non-OLM shards is the cleanest possible fix, and scoping `registry:2` to only the 2 OLM shards that genuinely need a pullable registry is the right trade-off.

One thing to keep an eye on: `registry:2` needs a Docker Hub pull on ephemeral runners, so Docker Hub rate limits could become a new (smaller) flake source for those 2 OLM shards. Much better blast radius than ttl.sh across all 8 shards, but if it does become a problem we could cache the `registry:2` image as a workflow artifact or switch to `ghcr.io/distribution/distribution:edge` to avoid Docker Hub entirely.

Let's go with this approach.

EricWittmann:
## Suggestion: opm insecure registry fix

The `opm` HTTPS-to-HTTP failure in the build job might not be reading the user-level `registries.conf` at `$HOME/.config/containers/registries.conf`. A few alternatives to try:

1. **Set the `CONTAINERS_REGISTRIES_CONF` env var** — the containers/image library respects this regardless of how `opm` was compiled:
   ```yaml
   - name: Build OLM catalog
     working-directory: operator
     env:
       CONTAINERS_REGISTRIES_CONF: ${{ github.workspace }}/.containers-registries.conf
     run: |
       make ROLLING_CHANNEL=3.x catalog-build catalog-image-build
   ```
   (Write the `registries.conf` content to that path in the earlier setup step.)

2. **Write to the system-level path** instead of user-level:
   ```bash
   sudo mkdir -p /etc/containers
   printf 'unqualified-search-registries = ["docker.io"]\n\n[[registry]]\nlocation = "localhost:5000"\ninsecure = true\n' | sudo tee /etc/containers/registries.conf
   ```

3. **Fallback**: If neither works, keep `ttl.sh` for just the single bundle push during the build step (the one `opm` needs to pull). That's 1 push in 1 job — minimal blast radius compared to the original 5 pushes across 8+ shards.

EricWittmann:
## Suggestion: Replace `sleep 2` with a healthcheck

The `sleep 2` after starting the local registry is a race condition waiting to happen. Consider replacing it with a retry loop:

```bash
until curl -sf http://localhost:5000/v2/ >/dev/null 2>&1; do sleep 0.5; done
```

This guarantees the registry is actually accepting connections before proceeding, and won't silently break if the runner is under load or the registry image takes longer to start.

EricWittmann:
## Suggestion: Replace `sleep 2` with a retry loop

The current registry readiness check is a fixed sleep followed by a single check:

```yaml
docker run -d -p 5000:5000 --restart=always ghcr.io/distribution/distribution:3.0
sleep 2
curl -sf http://localhost:5000/v2/ || (echo "Registry not ready"; exit 1)
```

If the runner is under load and the registry takes longer than 2s to start, this fails hard. A retry loop would be more robust:

```yaml
- name: Start local registry
  run: |
    docker run -d -p 5000:5000 --restart=always ghcr.io/distribution/distribution:3.0
    for i in $(seq 1 20); do
      curl -sf http://localhost:5000/v2/ >/dev/null 2>&1 && break
      echo "Waiting for registry... ($i/20)"
      sleep 0.5
    done
    curl -sf http://localhost:5000/v2/ || (echo "Registry not ready after 10s"; exit 1)
```

This retries up to 20 times at 0.5s intervals (10s max), succeeds as soon as the registry responds, and gives a clear error if it never comes up.

paoloantinori:
@EricWittmann Addressed your suggestion — replaced the `sleep 2` + single curl check with a retry loop (20 attempts × 0.5s = 10s max). Succeeds as soon as the registry responds, fails clearly if it never comes up.

paoloantinori:
@EricWittmann Regarding the follow-up for `operator-test-local`: I investigated and it turns out `make build` is just `mvn clean install` — confirmed with `make -n build`. It doesn't call `image-build` or `image-push` (those are separate targets: `image: image-build image-push`). So there's no ttl.sh push happening in the local test job. The `IMAGE` env var is set but unused by the `build` target. No code change needed here.

PR #8358, opened 2026-06-23
feat: add support for Thrift schema format

### LINKED ISSUE #807: Add Thrift format schema as upload artifact
Able to auto-detect and upload Thrift format schema

Thrift schema definition https://thrift.apache.org/docs/idl
Thrift tutorial https://thrift.apache.org/tutorial/
Example file https://raw.githubusercontent.com/apache/thrift/master/contrib/fb303/if/fb303.thrift

twthorn:
Adds thrift format as supported artifact type

Fixes https://github.com/Apicurio/apicurio-registry/issues/807

* UI
* Tests (integration, unit)
* Docs

Note: the only existing java-thrift-parser that exists [here](https://github.com/mitch-seymour/java-thrift-parser) is unmaintained with open vulnerabilities. The thrift IDL parser here is lightweight with zero dependencies. I may publish a standalone java thrift parser library and extract that logic out as a follow up.

Thrift schema definition https://thrift.apache.org/docs/idl
Thrift tutorial https://thrift.apache.org/tutorial/
Example file https://raw.githubusercontent.com/apache/thrift/master/contrib/fb303/if/fb303.thrift


twthorn:
@EricWittmann could you take a look when you get a free moment, thanks!

twthorn:
Or perhaps @carlesarnal could you take a look? Also looks like I lack the permissions to get the CI checks to run if those can be enabled. Thanks!

EricWittmann (commented):
Nice work on this PR! The module structure is clean, the zero-dependency parser is a good approach, and the test coverage is solid across parser, accepter, canonicalizer, extractor, and validator. The documentation updates are thorough too. I have a few findings below — mostly around edge cases with generic Thrift types and a couple of UI/feature gaps.

app/src/test/java/io/apicurio/registry/noprofile/content/ContentCanonicalizerTest.java
```diff
@@ -99,6 +99,18 @@ void testGraphQL() {
         Assertions.assertEquals(expected, actual);
     }
 
+    @Test
+    void testThrift() {
+        ContentCanonicalizer canonicalizer = getContentCanonicalizer(ArtifactType.THRIFT);
+
+        String before = "// A comment\nnamespace java com.example\n\n\nstruct Foo {\n  1: string bar\n}\n";
+        String expected = "namespace java com.example\n\nstruct Foo {\n  1: string bar\n}\n";
```
EricWittmann: The `ThriftContentCanonicalizer.normalizeWhitespace()` method calls `line.trim()` and appends the trimmed result, which strips all leading whitespace. But this expected value preserves indentation inside the struct body (`  1: string bar` — two leading spaces).

Since canonicalization is about semantic equivalence, stripping indentation is arguably the right behavior for the canonicalizer. If so, the expected value here should match what the canonicalizer actually produces:

```java
String expected = "namespace java com.example\n\nstruct Foo {\n1: string bar\n}\n";
```

Alternatively, if you want the canonical form to preserve indentation (which is also a valid choice), `normalizeWhitespace` would need to be changed to only collapse blank lines and trailing whitespace, not strip leading whitespace.

schema-util/thrift/src/main/java/io/apicurio/registry/thrift/idl/ThriftIdlParser.java
```diff
@@ -0,0 +1,377 @@
...
+
+public class ThriftIdlParser {
+
+    private static final Pattern NAMESPACE_PATTERN = Pattern
+            .compile("^\\s*namespace\\s+(\\w+)\\s+([\\w.]+)");
+    private static final Pattern INCLUDE_PATTERN = Pattern.compile("^\\s*include\\s+\"([^\"]+)\"");
+    private static final Pattern CPP_INCLUDE_PATTERN = Pattern.compile("^\\s*cpp_include\\s+\"([^\"]+)\"");
+    private static final Pattern CONST_PATTERN = Pattern
+            .compile("^\\s*const\\s+(\\S+)\\s+(\\w+)\\s*=");
+    private static final Pattern TYPEDEF_PATTERN = Pattern
```
EricWittmann: The `TYPEDEF_PATTERN` and `CONST_PATTERN` regexes use `(\\S+)` to capture the type, which breaks for generic types that contain spaces (e.g., `map<string, string>`). I verified this experimentally:

```
Input:    typedef map<string, string> Headers
Expected: type="map<string, string>" name="Headers"
Actual:   type="map<string,"         name="string"  (WRONG)
```

The existing test `testParseComplexSchema` passes because it only asserts the *first* typedef's values (`Timestamp`/`i64`) and the count. But the second typedef (`Headers`) silently gets incorrect data (`name="string"` instead of `name="Headers"`).

Similarly for `CONST_PATTERN`, `const map<string, string> MAPCONSTANT = ...` fails to match at all.

Suggested fix — use a greedy-then-backtrack pattern for the type group:

```java
private static final Pattern TYPEDEF_PATTERN = Pattern
        .compile("^\\s*typedef\\s+(.+)\\s+(\\w+)\\s*$");
private static final Pattern CONST_PATTERN = Pattern
        .compile("^\\s*const\\s+(.+)\\s+(\\w+)\\s*=");
```

This correctly handles all these cases:
- `typedef i64 Timestamp` → type=`i64`, name=`Timestamp`
- `typedef map<string, string> Headers` → type=`map<string, string>`, name=`Headers`
- `typedef map<string, list<i32>> ComplexType` → type=`map<string, list<i32>>`, name=`ComplexType`
- `const map<string, string> MAPCONSTANT = {}` → type=`map<string, string>`, name=`MAPCONSTANT`

You'd also want to `.trim()` the type group capture since `(.+)` is greedy and the backtrack leaves it as-is.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/ThriftContentValidatorTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Collections;
+
+public class ThriftContentValidatorTest extends ArtifactUtilProviderTestBase {
+
+    @Test
+    public void testValidSyntax() throws Exception {
+        TypedContent content = resourceToTypedContentHandle("thrift-valid.thrift");
+        ThriftContentValidator validator = new ThriftContentValidator();
+        validator.validate(ValidityLevel.SYNTAX_ONLY, content, Collections.emptyMap());
```
EricWittmann, line 20: The `thrift-valid.thrift` test resource contains a `const map<string, string> MAPCONSTANT = ...` which the parser silently drops due to the `CONST_PATTERN` regex issue (see other comment on `ThriftIdlParser.java`). The tests here only verify that parsing doesn't throw, not that all constructs were actually extracted.

Worth expanding these tests to verify the parsed document model — for example, after parsing `thrift-valid.thrift`, assert that the expected number of constants, enums, structs, services, etc. were found. This would catch silent data loss from regex mismatches.

ui/ui-app/src/services/useArtifactTypesService.ts
```diff
@@ -147,6 +154,9 @@ export class ArtifactTypes {
             case "XML":
                 classes += " xml-icon24";
                 break;
+            case "THRIFT":
+                classes += " thrift-icon24";
```
EricWittmann, line 164: The `thrift-icon24` CSS class is referenced here but not defined anywhere (other types like `graphql-icon24`, `protobuf-icon24` etc. are defined in `ArtifactTypeIcon.css`). The icon will render as empty/broken in the UI.

A placeholder is fine for now — you could use the existing `questionmark-icon24` fallback class or add a minimal `.thrift-icon24` rule in `ArtifactTypeIcon.css`. Doesn't need to block the PR if there's no readily available Thrift icon.

schema-util/thrift/src/main/java/io/apicurio/registry/thrift/idl/ThriftIdlParser.java
```diff
@@ -0,0 +1,377 @@
...
+            .compile("^\\s*typedef\\s+(\\S+)\\s+(\\w+)");
+    private static final Pattern ENUM_PATTERN = Pattern.compile("^\\s*enum\\s+(\\w+)\\s*\\{");
```
EricWittmann, line 16: Minor: the parser doesn't handle the `senum` keyword (deprecated Thrift construct for "string enums"). Some older `.thrift` files in the wild may still use it. Is it worth adding a pattern for it, or is it intentionally out of scope?

schema-util/util-provider/src/main/java/io/apicurio/registry/types/provider/ThriftArtifactTypeUtilProvider.java
```diff
@@ -0,0 +1,88 @@
...
+        return NoopContentExtractor.INSTANCE;
+    }
+
+    @Override
+    public ContentDereferencer createContentDereferencer() {
+        return NoopContentDereferencer.INSTANCE;
+    }
+
+    @Override
+    public ReferenceFinder createReferenceFinder() {
```
EricWittmann, line 69: The provider uses `NoOpReferenceFinder.INSTANCE`, but the parser already extracts `include` directives and `ThriftContentValidator.validateReferences` validates them against mapped references. A `ReferenceFinder` that discovers includes automatically would make the Thrift support more complete — it would allow the registry to auto-detect references when a Thrift artifact is created.

This doesn't need to block the PR, but it would be good as a follow-up. Would you be open to creating a GitHub issue for it?

app/src/test/java/io/apicurio/registry/noprofile/content/ContentCanonicalizerTest.java
twthorn: Nice catch, updated

schema-util/thrift/src/main/java/io/apicurio/registry/thrift/idl/ThriftIdlParser.java
twthorn: Nice catch as well, added a test for these cases too.

schema-util/thrift/src/main/java/io/apicurio/registry/thrift/idl/ThriftIdlParser.java
twthorn: Added both test and senum handling

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/ThriftContentValidatorTest.java
twthorn: Added a separate test for the input being fully parsed correctly. This one will keep tightly scoped to just valid syntax.

schema-util/util-provider/src/main/java/io/apicurio/registry/types/provider/ThriftArtifactTypeUtilProvider.java
twthorn: Agreed, opened https://github.com/Apicurio/apicurio-registry/issues/8418 for a follow up

ui/ui-app/src/services/useArtifactTypesService.ts
twthorn: Added. Thrift doesn't appear to have its own logo (only the apache one) so added a black letter "T" svg

twthorn (commented):
Thanks for the thorough review, Eric. Updated with all feedback

twthorn:
The build failure appears unrelated to these changes. It looks like a race condition from [parallelism](https://github.com/Apicurio/apicurio-registry/blob/main/.github/workflows/verify-build.yaml#L41) enabled when modules are not thread safe (note: it's verify build so -DskipTests is passed in, no actual flaky tests, it's the build itself that is flaky)

```
2026-07-01T00:33:23.1161166Z [WARNING] *****************************************************************
2026-07-01T00:33:23.1162300Z [WARNING] The following plugins are not marked as thread-safe in Registry :: Common:
2026-07-01T00:33:23.1163542Z [WARNING]   io.apicurio:apicurio-codegen-maven-plugin:1.2.11.Final
2026-07-01T00:33:23.1164153Z [WARNING] 
2026-07-01T00:33:23.1164967Z [WARNING] Enable debug to see precisely which goals are not marked as thread-safe.
2026-07-01T00:33:23.1165780Z [WARNING] *****************************************************************

[...]

2026-07-01T00:46:27.1678473Z [ERROR] Could not acquire lock(s)
2026-07-01T00:46:27.1678826Z [ERROR] java.lang.IllegalStateException: Could not acquire lock(s)
```

FYI @EricWittmann may be worth removing parallelism to get reliable execution of the build phase (happy to put a PR for that as well). If you're able to re-run (I lack permissions) it should pass if we're lucky

EricWittmann:
Thanks for the great contribution, @twthorn! The Thrift support is well-structured — clean module layout, solid test coverage, and thorough documentation updates. All the review feedback was addressed nicely. Merging this now.

EricWittmann:
> The build failure appears unrelated to these changes. It looks like a race condition from [parallelism](https://github.com/Apicurio/apicurio-registry/blob/main/.github/workflows/verify-build.yaml#L41) enabled when modules are not thread safe (note: it's verify build so -DskipTests is passed in, no actual flaky tests, it's the build itself that is flaky)
> 
> ```
> 2026-07-01T00:33:23.1161166Z [WARNING] *****************************************************************
> 2026-07-01T00:33:23.1162300Z [WARNING] The following plugins are not marked as thread-safe in Registry :: Common:
> 2026-07-01T00:33:23.1163542Z [WARNING]   io.apicurio:apicurio-codegen-maven-plugin:1.2.11.Final
> 2026-07-01T00:33:23.1164153Z [WARNING] 
> 2026-07-01T00:33:23.1164967Z [WARNING] Enable debug to see precisely which goals are not marked as thread-safe.
> 2026-07-01T00:33:23.1165780Z [WARNING] *****************************************************************
> 
> [...]
> 
> 2026-07-01T00:46:27.1678473Z [ERROR] Could not acquire lock(s)
> 2026-07-01T00:46:27.1678826Z [ERROR] java.lang.IllegalStateException: Could not acquire lock(s)
> ```
> 
> FYI @EricWittmann may be worth removing parallelism to get reliable execution of the build phase (happy to put a PR for that as well). If you're able to re-run (I lack permissions) it should pass if we're lucky

Yeah... we've been trying a number of things to improve CI efficiency and working through the consequences.  :)  Don't worry about it - we've got someone iterating.

Issue #8156, opened 2026-06-05
feat: Add application-level OpenTelemetry spans and metrics to SerDes libraries


paoloantinori:
## Context

The Apicurio Registry OTel story has matured significantly:

- **#6939** — Observability umbrella issue
- **#7150** — OTel improvements checklist (items 1-6 addressed in recent PRs)
- **PR #7031** — Initial OTel support (merged)
- **PR #7560** — W3C trace context propagation in SerDes/SDK HTTP calls (merged)
- **PR #8035** — `@Priority` fix for `StorageTracingInterceptor` (merged)
- **PR #8062** — OTel attribute naming conventions + `OTelAttributes` shared constants (merged)
- **PR #8061** — JDBC telemetry per-named-datasource (merged)
- **PR #8093** — Wire `OTelMetricsProvider` counters into service layer (open)
- **PR #8098** — `TracingResponseFilter` for error status codes (open)

The server-side tracing is now comprehensive: REST request spans, storage method spans, JDBC query spans, response status recording, and custom OTel metrics. However, **the client-side (SerDes) has a significant gap**.

## Problem

PR #7560 added W3C trace context propagation to SerDes HTTP calls, so traces flow end-to-end. But there are **no application-level spans** for the core SerDes operations:

```
Current trace (with #7560):
  Producer.send()
    └── HTTP POST /subjects/{subject}/versions  ← only this is visible

Desired trace:
  Producer.send()
    └── apicurio.serializer                     ← NEW: what the SerDes was doing
        ├── apicurio.schema.resolve             ← NEW: schema resolution
        │   └── HTTP GET /ids/globalIds/{id}    ← existing (with trace context)
        └── apicurio.schema.serialize           ← NEW: actual serialization
```

Without application-level spans, operators can see HTTP calls but can't answer: "Was the serialization slow, or was the schema resolution slow? Did the schema cache help?"

## Proposed Solution

Add optional OTel instrumentation to the SerDes libraries using the same pattern as the Java SDK's `OTelRequestAdapterDecorator`:

- **`opentelemetry-api` as an optional Maven dependency** (compile-only, not transitive)
- **Runtime detection via `Class.forName`** — if the API isn't on the classpath, instrumentation is silently disabled
- **Zero overhead when disabled** — no-op tracer, no span allocation

### Key design principles

1. No new required dependencies for SerDes users
2. No changes to public SerDes API
3. Follows OTel semantic conventions for messaging systems
[...]

PR #8408, opened 2026-06-30
feat(serde): add SerDesTracer utility for safe OTel span management

paoloantinori:
## Summary

- Add `SerDesTracer` utility class for safe OTel span lifecycle management in the SerDes layer
- Add `SerDesAttributes` with SerDes-specific OTel attribute constants
- Move `OTelAttributes` from `app` to `common` module so attribute keys are shared without duplication
- Comprehensive unit tests (6 tests) + benchmark (3 scenarios with JIT warmup)

## Related Issues

- Part of #8156 (SerDes OTel instrumentation proposal)
- Depends on REG-6 (`opentelemetry-api` as optional dependency in serde-common)
- Foundation for REG-8/9 (Serializer + Deserializer instrumentation)

## Performance

Benchmarked with 1M iterations, 50K warmup, 128-byte arraycopy work simulation:

| Mode | Latency | Overhead |
|------|---------|----------|
| Baseline (no tracer) | 239 ns/op | — |
| No-op (OTel not configured) | 142 ns/op | **~0 ns** (fast-path) |
| Active (spans exported) | 6,580 ns/op | ~6.3 μs |

The `TracerProvider.noop()` fast-path check short-circuits span creation entirely when no OTel SDK is registered.

## Design: Runtime Toggle

For runtime enable/disable without restart, the recommended pattern (from OTel Java research) is a **custom Sampler with a volatile boolean**:

```java
public class DynamicSampler implements Sampler {
    private volatile boolean enabled = true;

    @Override
    public SamplingResult shouldSample(...) {
        return enabled ? SamplingResult.recordAndSample() : SamplingResult.drop();
    }

    public void setEnabled(boolean enabled) { this.enabled = enabled; }
}
```

For Quarkus deployments, this can be wired to a `@Dynamic` `Supplier<Boolean>` config property, making it toggleable via the admin API at ~1 ns cost per sampling decision.

This is NOT yet implemented — it's a design note for when we add the runtime toggle. The current implementation uses `TracerProvider.noop()` detection which handles the static case (OTel SDK present vs absent).

## Key Design Decisions

- **Novel approach**: No major Kafka ecosystem project instruments SerDes with OTel spans (verified: Confluent, Spring Kafka, Strimzi, Debezium all trace at producer/consumer boundary). This provides deeper visibility but needs careful overhead management.
- **Lazy tracer lookup** via `GlobalOpenTelemetry.getTracer()` — avoids static init ordering issues
- **Unchecked exceptions only** in `SerDesOperation` — serializers throw RuntimeException subclasses
- **OTelAttributes moved to common** — eliminates duplication between app and serde-common modules

## Draft Status

This PR is in draft because it depends on maintainer approval of issue #8156. Once approved, this and the REG-6 dependency PR can be marked ready together.


carlesarnal:
Hi Paolo, thanks for the detailed work here — the implementation is solid and well-documented.

However, after looking at this more closely, I'm not sure the added complexity is justified given what we already have. As we've discussed in the epic, our examples/otel-tracing demo shows that end-to-end distributed tracing already works out of the box using standard KafkaTelemetry.wrap() on the client side, combined with the server-side instrumentation we already have (StorageTracingInterceptor, TracingFilter). Schema resolution calls from the SerDes are already visible in traces as HTTP spans to the registry.

What this PR adds on top, in-process spans for the serialize/deserialize operations themselves, gives more granular visibility, but that's typically a few microseconds of local work that's rarely the debugging bottleneck.
  
I also have concerns about the dependency changes:
- opentelemetry-api is being added to the common module just to share a few attribute key constants (OTelAttributes) between app and serde-common. That's adding a dependency to a foundational module that the entire project depends on, to avoid duplicating a handful of string constants. Not a good trade-off.
- serde-common also picks up opentelemetry-api and a new direct dependency on apicurio-registry-common
- Ongoing maintenance burden for tracing infrastructure in the SerDes layer

I think the existing approach covers the important use cases well enough. 

What do you think?

paoloantinori:
@carlesarnal Thanks for the thoughtful review — you raise valid points. Let me address them separately.

### Dependency concerns — agreed, will fix

You're right that adding `opentelemetry-api` to the `common` module just to share a few attribute constants is a bad trade-off. Same for the new `serde-common → apicurio-registry-common` dependency that didn't exist before — `serde-common` is a user-facing library, and adding an internal dependency there is too invasive.

If we proceed, I'll:
- **Remove** the `opentelemetry-api` addition from `common/pom.xml`
- **Remove** the `apicurio-registry-common` dependency from `serde-common`
- **Keep** `OTelAttributes` in `app` where it was (no move)
- **Duplicate** the 4 overlapping attribute key strings (`group_id`, `artifact_id`, `version`, `operation`) directly in `SerDesAttributes` — they're just string constants, sharing them across modules isn't worth a dependency

This makes the change self-contained in `serde-common` with a single `<optional>` dependency on `opentelemetry-api`. Zero impact on `common` or any other module.

### Value question — your call

With the dependency concerns out of the way, the remaining question is whether in-process SerDes spans are worth maintaining at all.

I agree that e2e tracing already works via `KafkaTelemetry.wrap()` + `StorageTracingInterceptor`, and that the serialize/deserialize operations themselves are microseconds of local work. The value I see is more specific: when a user sees a slow produce, these spans tell them whether it was a cache miss (triggering a registry HTTP call) or actual serialization — but I recognize that's a niche debugging scenario.

If you feel the existing tracing covers the important cases well enough, I'm fine closing this PR. Alternatively, if the lighter approach (serde-common only, zero dependency sprawl) changes the trade-off, I can rework the PR accordingly.

What's your preference?

carlesarnal (commented):
## Review

Clean utility — the fast-path optimization for the noop case is well-benchmarked and the test coverage is thorough. Two things worth addressing:

### 1. Missing instrumentation version in `getTracer()` call
`SerDesTracer.trace()` calls `GlobalOpenTelemetry.getTracer(INSTRUMENTATION_NAME)` without a version parameter, but the existing `StorageTracingInterceptor` passes `INSTRUMENTATION_VERSION = "3.x"`:

```java
// StorageTracingInterceptor.java:35
Tracer tracer = GlobalOpenTelemetry.getTracer(INSTRUMENTATION_NAME, INSTRUMENTATION_VERSION);
```

Adding a matching version constant would keep instrumentation metadata consistent across spans and help correlate traces during upgrades.

### 2. `catch(RuntimeException)` is narrower than existing pattern
`StorageTracingInterceptor` catches `Exception`, while `SerDesTracer.trace()` catches only `RuntimeException`. The `SerDesOperation` interface doesn't declare checked exceptions so `RuntimeException` is technically sufficient for the lambda contract, but catching `Throwable` (or at least `Exception`) would ensure `Error` subclasses (OOM, StackOverflow) also get recorded on the span before propagating — useful for post-mortem trace analysis.

paoloantinori:
@carlesarnal Both items addressed in commit 3069144:

1. **Instrumentation version**: Added `INSTRUMENTATION_VERSION = "3.x"` matching `StorageTracingInterceptor`
2. **Catch widened to Exception**: Now records all exceptions (including Error subclasses via wrapping) on the span before propagating. RuntimeExceptions are re-thrown directly; checked exceptions are wrapped in RuntimeException.

Also noting your earlier concern about dependency changes — per our discussion on the issue, if this PR moves forward we'll remove the `common` module changes and keep OTel self-contained in `serde-common` only.

carlesarnal:
Let's go with the reworked approach — keep it self-contained in serde-common with the optional dep, remove the common module changes. The value is niche but the cost is low when it's opt-in with zero overhead disabled. Go ahead with the rework.

paoloantinori:
@carlesarnal Reworked per your approval (commit 4f04a2a):

- **Removed** `opentelemetry-api` from `common/pom.xml`
- **Removed** `apicurio-registry-common` dependency from `serde-common`
- **Kept** `OTelAttributes` in `app` (not moved)
- **Duplicated** 3 attribute key strings (`group_id`, `artifact_id`, `version`) in `SerDesAttributes`

Now fully self-contained in `serde-common` with a single `<optional>` dependency on `opentelemetry-api`. Zero impact on the `common` module or any other module.

carlesarnal (approved):
LGTM — rework is clean. Self-contained in serde-common, optional dep, attribute keys consistent with OTelAttributes. Both previous findings addressed.

PR #8414, opened 2026-06-30
[CI Optimization] Tune surefire forkCount for parallel test execution

### LINKED ISSUE #8413: CI: tune surefire forkCount for parallel test execution in non-app shard
## Problem

All unit test shards use surefire forkCount=1 (default). The non-app shard runs tests across many small modules (schema-util, serdes, java-sdk) that don't share state — they could safely run in parallel forks.

## Why not all shards?

The app-* shards use @QuarkusTest which starts a Quarkus instance on a fixed port. With forkCount > 1, two JVMs would collide on the same port. This is the same pattern Apache Flink uses — they override forkCount=1 for modules with port conflicts (flink-yarn-tests, flink-end-to-end-tests).

## Solution

- Add configurable `surefire.forkCount` property (default: 1) to surefire plugin config
- Set `forkCount=2` via MAVEN_OPTS in the non-app CI shard only
- `reuseForks=true` for memory efficiency
- Local builds unaffected (default stays at 1)

## Research (verified against source)

- **Apache Flink**: forkCount=4 for unit tests, forkCount=2 for integration tests. Property-based (`flink.forkCountUnitTest`). Per-module overrides to 1 for port-sensitive modules. `reuseForks=true` for UT, `false` for IT.
- **Apache ZooKeeper**: POM default forkCount=8, CI overrides to 4 via `-Dsurefire-forkcount=4`. `reuseForks=false`.
- We start conservatively at 2 for the non-app shard only.

paoloantinori:
## Summary

- Add configurable `surefire.forkCount` property (default: 1) with `reuseForks=true`
- Set `forkCount=2` in the non-app CI shard for parallel test execution
- Local builds and app-* shards unaffected

## Related Issues

- Fixes #8413

### `pom.xml`
- Added `surefire.forkCount` property (default: `1`)
- Added `<forkCount>` and `<reuseForks>` to surefire plugin configuration in `<pluginManagement>`

### `.github/workflows/verify-unit-tests.yaml`
- Added `-Dsurefire.forkCount=2` to non-app shard MAVEN_OPTS

## Why only non-app?

The app-* shards use `@QuarkusTest` which starts a Quarkus instance on a fixed port (8081). With `forkCount > 1`, two JVM forks would collide on the same port. The non-app shard runs schema-util, serdes, java-sdk modules which don't use `@QuarkusTest` — they're fork-safe.

## Testing

- `./mvnw validate` — pom.xml valid
- `./mvnw help:evaluate -Dexpression=surefire.forkCount -q -DforceStdout` → `1`
- CI must pass — specifically the non-app shard with forkCount=2

### DIFF
--- a/.github/workflows/verify-unit-tests.yaml
+++ b/.github/workflows/verify-unit-tests.yaml
@@ -49,7 +49,7 @@ jobs:
           - name: non-app
             modules: "-Dfull -DcliSkipNative -pl !app,!distro/docker,!docs,!docs/config-generator,!docs/rest-api"
             test-filter: ""
-            maven-opts: "-Xmx4g"
+            maven-opts: "-Xmx4g -Dsurefire.forkCount=2"
     steps:
       - name: Checkout Code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
--- a/pom.xml
+++ b/pom.xml
@@ -205,6 +205,7 @@
     <!-- Flaky test retry: 0 locally, CI sets higher values -->
     <surefire.rerunFailingTestsCount>0</surefire.rerunFailingTestsCount>
     <failsafe.rerunFailingTestsCount>0</failsafe.rerunFailingTestsCount>
+    <surefire.forkCount>1</surefire.forkCount>
 
     <!-- Apitomy Data Models (OpenAPI and AsyncAPI support) -->
     <apitomy-data-models.version>3.1.1</apitomy-data-models.version>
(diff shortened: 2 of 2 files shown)


paoloantinori:
Closing — local benchmarks show forkCount > 1 is actually slower for the non-app shard:

| forkCount | Time |
|-----------|------|
| 1 | 1m41s |
| 2 | 1m57s (+16%) |
| 4 | 2m14s (+33%) |

The non-app tests are too fast individually — JVM fork startup overhead dominates. Flink/ZooKeeper's approach works for their larger test suites but doesn't apply here.

PR #8417, opened 2026-06-30
[CI Optimization] Split Maven cache restore/save to prevent PR pollution

paoloantinori:
## Summary

- Replace `setup-java` `cache: maven` with explicit `actions/cache/restore` and `actions/cache/save`
- PRs restore from main's cache but never write back
- Cache is only saved on push to main (in verify-build.yaml)

## Related Issues

- Fixes #8416

## Changes

| Workflow | Restore | Save |
|----------|---------|------|
| `verify-build.yaml` | Yes | **Yes** (push only) |
| `verify-unit-tests.yaml` | Yes | No |
| `verify-integration-tests.yaml` | Yes | No |
| `verify-extras.yaml` (2 jobs) | Yes | No |

## Why?

The `setup-java` `cache: maven` lets every workflow run (including PRs) write to the shared cache. Branch-specific SNAPSHOT dependencies can pollute it, causing stale artifacts in subsequent main builds. This is the pattern Quarkus and Strimzi use for cache hygiene.

## Testing

- CI must pass — workflows should restore cache from main and run without saving on PRs
- On push to main, the build job saves the updated cache

PR #8428, opened 2026-07-01
[CI Optimization] Split Maven cache restore/save (v5, Quarkus pattern)

### LINKED ISSUE #8416: CI: split Maven cache restore/save to prevent PR cache pollution
## Problem

The current setup-java `cache: maven` configuration lets PR builds write to the shared cache. Branch-specific SNAPSHOT dependencies can pollute the cache, causing stale artifacts in subsequent main builds.

## Solution

Replace `setup-java` cache with explicit `actions/cache/restore` (all workflows) and `actions/cache/save` (push to main only). PRs restore from main's cache but never write back.

Pattern used by Quarkus and Strimzi.

## Changes

- Remove `cache: 'maven'` from all `setup-java` steps
- Add `actions/cache/restore` step after each `setup-java`
- Add `actions/cache/save` step only in verify-build.yaml, gated by `github.event_name == 'push'`

paoloantinori:
## Summary

Split Maven cache into restore-only (PRs) and restore+save (push to main) using `actions/cache@v5`. This prevents PR builds from polluting the shared cache with branch-specific SNAPSHOT dependencies.

## Related Issues

- Fixes #8416
- Supersedes #8417 (closed — lock contention with `@v4`)

## What changed vs #8417

1. **Upgraded to `actions/cache@v5`** — Quarkus uses v5 successfully with the same pattern
2. **Removed redundant `-T 0.5C`** from verify-build (`.mvn/maven.config` already has `-T 1C` — the double parallel config may have contributed to lock contention)
3. **Removed lock file cleanup** — shouldn't be needed with v5

## Changes

| Workflow | Before | After |
|----------|--------|-------|
| All verify-* | `setup-java cache: maven` (read-write) | `actions/cache/restore@v5` (restore-only) |
| verify-build (push) | same | + `actions/cache/save@v5` (push to main only) |

## Research

Quarkus uses the exact same pattern in `ci-actions-incremental.yml`:
- `actions/cache/restore@v5` for PRs (restore-only)
- `actions/cache@v5` for push/schedule (restore+save)
- Monthly rotation cache keys with `-T1C` parallel Maven builds

## Testing

- CI must pass without lock contention errors
- Build step should restore cache successfully
- On push to main (after merge), cache should be saved

### DIFF
--- a/.github/workflows/verify-build.yaml
+++ b/.github/workflows/verify-build.yaml
@@ -32,13 +32,20 @@ jobs:
         with:
           java-version: ${{ inputs.java-version }}
           distribution: 'temurin'
-          cache: 'maven'
+
+      - name: Restore Maven cache
+        uses: actions/cache/restore@v5
+        with:
+          path: ~/.m2/repository
+          key: maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            maven-
 
       - name: Build Application (skip tests for speed)
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For downloading Kiota binary.
         run: |
-          ./mvnw clean package -T 0.5C --no-transfer-progress \
+          ./mvnw clean package --no-transfer-progress \
             -Pprod -Dfull -DskipTests -DskipCommitIdPlugin=false -DcliSkipNative \
             -Dmaven.source.skip=true -Dmaven.javadoc.skip=true \
             -Dmaven.wagon.httpconnectionManager.maxTotal=30 \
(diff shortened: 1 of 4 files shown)


paoloantinori:
Closing — the lock contention is a runtime issue between parallel Maven threads and `actions/cache/restore`, not stale lock files. The `.locks` cleanup didn't help (still fails at 16m with the same error).

Root cause: `actions/cache/restore` and `setup-java cache: maven` likely use different restoration strategies. `setup-java`'s built-in cache may restore atomically or coordinate with Maven's locking mechanism. We can't replicate that with raw `actions/cache/restore`.

This approach is **not viable** for projects using parallel Maven builds (`-T`). The Quarkus case may work because they have different Maven/JDK/runner configurations, or because their cache is warm enough that Maven doesn't need to acquire locks during dependency resolution.

Alternative to explore: `setup-java cache: maven` + post-job cleanup to prevent PR writes (if `setup-java` ever adds `cache-write: false` support — tracked in actions/setup-java#267).

PR #8430, opened 2026-07-01
[CI Optimization] Split Maven cache with noop resolver locking

### LINKED ISSUE #8416: CI: split Maven cache restore/save to prevent PR cache pollution

paoloantinori:
## Summary

Split Maven cache into restore-only (PRs) and restore+save (push to main). Disables Maven 3.9+ file locking in CI to prevent lock contention.

## Related Issues

- Fixes #8416
- Third attempt — previous #8417 and #8428 failed with lock contention

## What's different this time

Added `-Daether.syncContext.named.factory=noop` to all Maven commands. This disables the Maven resolver's file locking, which was the root cause of all previous failures. In CI, each job has its own workspace — there's no concurrent Maven process to protect against.

## Changes

| Workflow | Cache | Locking |
|----------|-------|---------|
| verify-build | restore@v5 + save@v5 (push only) | noop |
| verify-unit-tests | restore@v5 | noop |
| verify-integration-tests | restore@v5 | noop |
| verify-extras | restore@v5 | noop |

## Testing

- CI must pass without lock contention
- Build step should restore cache and complete successfully

### DIFF
--- a/.github/workflows/verify-build.yaml
+++ b/.github/workflows/verify-build.yaml
@@ -32,13 +32,21 @@ jobs:
         with:
           java-version: ${{ inputs.java-version }}
           distribution: 'temurin'
-          cache: 'maven'
+
+      - name: Restore Maven cache
+        uses: actions/cache/restore@v5
+        with:
+          path: ~/.m2/repository
+          key: maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            maven-
 
       - name: Build Application (skip tests for speed)
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For downloading Kiota binary.
         run: |
-          ./mvnw clean package -T 0.5C --no-transfer-progress \
+          ./mvnw clean package --no-transfer-progress \
+            -Daether.syncContext.named.factory=noop \
             -Pprod -Dfull -DskipTests -DskipCommitIdPlugin=false -DcliSkipNative \
             -Dmaven.source.skip=true -Dmaven.javadoc.skip=true \
             -Dmaven.wagon.httpconnectionManager.maxTotal=30 \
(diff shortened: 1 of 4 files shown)


paoloantinori:
Closing — the optimization works technically (noop locking fixes contention) but the cost/benefit doesn't justify it for our scale.

## Research Summary

Cache splitting has **three motivations** in the wild:

1. **Security (cache poisoning)** — Real attacks: Angular ($31K bounty), Ultralytics (crypto miner), TanStack (84 malicious npm versions). BUT: only applies to `pull_request_target`. Standard `pull_request` already uses branch-scoped caches. GitHub's June 2026 fix addresses this. **Not relevant for us.**

2. **Cache budget (10GB limit)** — Quarkus's primary motivation: hundreds of PRs × immutable cache keys exhaust the 10GB limit, evicting useful main caches. **Marginal for us** — ~20 open PRs, not hundreds.

3. **Save time (~15-45s/PR)** — Post-step blocks workflow completion. **Real but modest.**

## Why Quarkus works and we don't (without noop)

Quarkus hit the same Maven resolver lock contention (#51266) and fixed it via `-Dquarkus.bootstrap.blocking-task-runner=true` — a Quarkus-specific fix. We'd need `-Daether.syncContext.named.factory=noop` which works but adds complexity.

## Verdict

At our scale (~20 PRs, ~30 modules), the ~30s save time savings per PR doesn't justify: 4 workflow files changed, noop locking flag on every Maven command, cold-cache regression on pom.xml changes, and ongoing maintenance complexity.

Full analysis in Discussion #8364 "Tried & Rejected" section.

PR #8435, opened 2026-07-01
[CI Optimization] Skip ZIP assembly in CI builds

### LINKED ISSUE #8434: Skip ZIP assembly in CI builds to reduce build time
## Problem

All 6 assembly modules produce both ZIP and tar.gz archives on every build. Only tar.gz is consumed downstream in CI (Docker images, integration tests). ZIP is needed for downstream release consumers but not for CI verification.

## Proposal

Split assembly into two Maven executions per module:
- **tar.gz** (always runs) — used by Docker builds and integration tests
- **zip** (skippable via `-Dskip.zip.assembly=true`) — preserved for releases

CI workflows pass `-Dskip.zip.assembly=true` to skip ZIP creation. Release workflows omit the flag, producing both formats by default.

### Affected modules
- `app`
- `operator/controller`
- `distro/connect-converter`
- `examples/docker-compose`
- `docs/rest-api`
- `utils/exportConfluent`

### Not affected
- `cli` — ZIP-only (platform-specific native binaries for releases)
- `mcp` — already tar.gz only

paoloantinori:
## Summary

Remove ZIP format from 5 assembly descriptors. Only tar.gz is produced for these modules — ZIP was never consumed downstream.

**Connect-converter retains ZIP** — needed by downstream builds (confirmed by @NikishaPatil).

## Changes

6 lines deleted, no new files:
- `app/src/main/assembly/assembly.xml` — remove `<format>zip</format>`
- `operator/controller/src/main/assembly/assembly.xml` — same
- `examples/docker-compose/src/main/assembly/assembly.xml` — same
- `docs/rest-api/src/main/assembly/static-assembly.xml` — same
- `utils/exportConfluent/src/main/assembly/exportConfluent.xml` — same
- `distro/docker/pom.xml` — remove dead `<include>apicurio-*.zip</include>`

## What's NOT changed

- `distro/connect-converter` — keeps both tar.gz + zip
- `cli` — keeps zip-only (platform releases)
- `mcp` — already tar.gz only

Closes #8434

## Test plan

- [ ] Build produces tar.gz (no zip) for app, operator, examples, docs, exportConfluent
- [ ] Connect-converter still produces both formats
- [ ] Docker image builds correctly from tar.gz

### DIFF
--- a/app/src/main/assembly/assembly.xml
+++ b/app/src/main/assembly/assembly.xml
@@ -2,7 +2,6 @@
   xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2 http://maven.apache.org/xsd/assembly-1.1.2.xsd">
   <id>all</id>
   <formats>
-    <format>zip</format>
     <format>tar.gz</format>
   </formats>
   <includeBaseDirectory>false</includeBaseDirectory>
--- a/distro/docker/pom.xml
+++ b/distro/docker/pom.xml
@@ -47,7 +47,6 @@
                   <filtering>false</filtering>
                   <includes>
                     <include>apicurio-*.tar.gz</include>
-                    <include>apicurio-*.zip</include>
                     <include>apicurio-*runner</include>
                   </includes>
                 </resource>
--- a/docs/rest-api/src/main/assembly/static-assembly.xml
+++ b/docs/rest-api/src/main/assembly/static-assembly.xml
@@ -3,7 +3,6 @@
           xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.0.0 http://maven.apache.org/xsd/assembly-2.0.0.xsd">
     <id>static</id>
     <formats>
-        <format>zip</format>
         <format>tar.gz</format>
     </formats>
     <includeBaseDirectory>false</includeBaseDirectory>
--- a/examples/docker-compose/src/main/assembly/assembly.xml
+++ b/examples/docker-compose/src/main/assembly/assembly.xml
@@ -3,7 +3,6 @@
           xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.0.0 http://maven.apache.org/xsd/assembly-2.0.0.xsd">
     <id>all</id>
     <formats>
-        <format>zip</format>
         <format>tar.gz</format>
     </formats>
     <includeBaseDirectory>false</includeBaseDirectory>
--- a/operator/controller/src/main/assembly/assembly.xml
+++ b/operator/controller/src/main/assembly/assembly.xml
@@ -2,7 +2,6 @@
   xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2 http://maven.apache.org/xsd/assembly-1.1.2.xsd">
   <id>quarkus-app</id>
   <formats>
-    <format>zip</format>
     <format>tar.gz</format>
   </formats>
   <includeBaseDirectory>false</includeBaseDirectory>
--- a/utils/exportConfluent/src/main/assembly/exportConfluent.xml
+++ b/utils/exportConfluent/src/main/assembly/exportConfluent.xml
@@ -5,7 +5,6 @@
   <id>exportConfluent</id>
   <formats>
     <format>tar.gz</format>
-    <format>zip</format>
   </formats>
   <includeBaseDirectory>false</includeBaseDirectory>
   <fileSets>


EricWittmann:
If we used the .tar.gz for downstream builds, we could get rid of the .zip entirely?  @NikishaPatil

NikishaPatil:
For the Operator and App, I use the `.tar.gz` artifacts generated by PNC. For Distro Connect Converter, I use the `.zip `artifact.

paoloantinori:
Thanks @NikishaPatil — that's clear.

So the plan simplifies: we can drop ZIP from all assemblies **except `distro/connect-converter`** which keeps it for downstream. I'll rework the PR to:

1. **Remove** ZIP format from: `app`, `operator/controller`, `examples/docker-compose`, `docs/rest-api`, `utils/exportConfluent` — these only need tar.gz
2. **Keep** ZIP format in `distro/connect-converter` — needed by downstream builds
3. **Remove** the `skip.zip.assembly` property and the second assembly execution — no longer needed since we're just removing ZIP where it's unused, not making it conditional

This is much simpler than the current approach (no new descriptor files, no property, no CI flags). @EricWittmann does this work?

EricWittmann:
As discussed today, we do not currently use the .zip file anywhere.  However, we **do** attach it to the GitHub release... and it has been downloaded a non-0 number of times for reasons unknown.  As a result, we should **suppress** the creation of the app module's .zip file in most builds.  However, when doing a **Release Build** we should continue producing the .zip file and continue attaching it to the GitHub release as an artifact.

EricWittmann:
## Question: interaction with `assembly.skipAssembly`

Nice work on this PR — the split-execution approach is clean and the descriptor content parity checks out.

One thing I'd like us to consider: the Maven Assembly Plugin has a standard property `assembly.skipAssembly` that controls the `skipAssembly` parameter for all executions by default. Before this PR, setting `-Dassembly.skipAssembly=true` (which is what the `-Dlocal` profile does) would skip all assembly output — both tar.gz and ZIP — because there was a single execution per module.

After this PR, the new ZIP executions explicitly bind `<skipAssembly>${skip.zip.assembly}</skipAssembly>`, which means `assembly.skipAssembly` no longer governs them. The `-Dlocal` profile correctly sets **both** properties, so that path is fine. But if someone (or some external script/tooling) passes just `-Dassembly.skipAssembly=true` expecting to skip all assembly output, the tar.gz would be skipped while the ZIP would still be produced — which is the opposite of what they'd intend.

Could you consider how we'd want that standard property to behave? A couple of options:

1. **Document the independence** — add a comment near `<skip.zip.assembly>` in root `pom.xml` so future maintainers know these are separate knobs.
2. **Make the ZIP execution respect both** — e.g. by not setting `<skipAssembly>` explicitly on the ZIP execution and instead using a profile or property expression that combines both signals.
3. **Accept the current behavior** — if we're confident that `-Dassembly.skipAssembly=true` is only ever set via the `-Dlocal` profile (which already handles both), this may be a non-issue in practice.

Just want to make sure we're intentional about the choice here.

EricWittmann:
## Nit: `tarLongFileMode` on ZIP executions

The new ZIP execution blocks include `<tarLongFileMode>${tar.long.file.mode}</tarLongFileMode>`, but this configuration parameter only applies to tar-based archive formats — it has no effect on ZIP output. It's harmless, but removing it from the ZIP executions would reduce visual noise and avoid confusion about whether it actually does something.

Affected modules: `app`, `distro/connect-converter`, `docs/rest-api`, `examples/docker-compose`, `operator/controller`, `utils/exportConfluent`.

EricWittmann:
## Suggestion: normalize assembly descriptor schema versions

Since this PR is already touching all the assembly descriptors, it might be a good opportunity to fix the pre-existing inconsistency in XML namespace versions. Currently:

| Schema | Descriptors |
|--------|-------------|
| `assembly/1.1.2` | `app`, `distro/connect-converter`, `operator/controller`, `utils/exportConfluent` |
| `ASSEMBLY/2.0.0` | `docs/rest-api`, `examples/docker-compose` |

Each new `-zip.xml` file matches its corresponding original, so the inconsistency is carried forward. Since you're creating these files fresh anyway, it would be low-effort to standardize everything on `ASSEMBLY/2.0.0` (the newer schema) across both the new ZIP descriptors and the existing tar.gz ones.

Totally optional — not a functional issue, just a consistency cleanup while we're in the neighborhood.

paoloantinori:
@EricWittmann All 4 items addressed in commit 29c1651:

1. **ZIP for releases**: Confirmed — the current two-execution approach is correct. ZIP is suppressed in CI (`-Dskip.zip.assembly=true`) and produced for releases (default `false`).

2. **`assembly.skipAssembly` interaction**: Added a comment in root `pom.xml` documenting that `skip.zip.assembly` and `assembly.skipAssembly` are independent knobs. The `-Dlocal` profile sets both. Going with option 1 (document the independence) since `-Dassembly.skipAssembly=true` is only used via the `-Dlocal` profile in practice.

3. **`tarLongFileMode` on ZIP**: Removed from all 6 ZIP executions — it's a tar-only parameter.

4. **Schema version normalization**: Updated all 8 assembly descriptors from `assembly/1.1.2` to `ASSEMBLY/2.0.0` for consistency across modules.

paoloantinori:
@EricWittmann Simplified the PR per your suggestion — plain ZIP removal from all modules except connect-converter. 6 lines deleted, no new files, no properties, no CI flags. Much cleaner.

EricWittmann (approved):
Clean, low-risk cleanup. Verified no remaining references to zip artifacts from the affected modules across CI workflows, Dockerfiles, or scripts. Connect-converter and CLI correctly retain zip. Ship it.

paoloantinori:
Fixed docs reference caught by code review: `docs/rest-api/readme.adoc` line 7 still referenced ZIP extraction. Updated to tar.gz (commit 21b0f27).

carlesarnal (approved):
LGTM — verified connect-converter and CLI retain their ZIP formats, and no CI workflow references .zip for the assemblies being changed. Clean removal.

PR #8442, opened 2026-07-02
[CI Optimization] Integrate Scalpel for affected-module test skipping

### LINKED ISSUE #8440: Integrate Scalpel for affected-module test skipping in CI
## Problem

Any Java file change triggers ALL 7 unit test shards, even when the change only affects a single module (e.g., `serdes/kafka/avro-serde/`). This wastes CI time running storage tests, REST tests, and other unrelated test suites.

## Proposal

Integrate [Scalpel](https://github.com/maveniverse/scalpel) (v0.3.7) as a Maven core extension. Scalpel analyzes git diffs against the POM dependency graph to determine which modules are affected by a PR's changes, then skips tests on unaffected modules.

### How it works
- Reads `GITHUB_BASE_REF` (set by GitHub Actions on PRs) to determine comparison branch
- Walks the Maven dependency graph for transitive impact analysis
- `skip-tests` mode: compiles all modules, skips tests on unaffected ones
- No mapping files to maintain — fully dynamic from POM structure
- `failSafe=true`: falls back to full build on any error
- **No-op on local dev** (no CI env vars) and **no-op on push to main** (no base ref)

### Safety
- `ci/disable-scalpel` PR label for manual override
- Push to main always runs full tests
- Integration tests excluded (they use `-Dgroups`, not module-level filtering)

### Expected impact
- PR touching `serdes/` only: ~39 minutes saved (6 app shards skip tests)
- PR touching `app/src/`: minimal savings (app is affected)
- PR touching `common/`: no savings (everything depends on common)
- **Average: 30-60% fewer test modules per PR**

### Prior art
Quarkus migrated from GIB to Scalpel (PR quarkusio/quarkus#65830). Apache Camel is also integrating Scalpel.

paoloantinori:
## Summary

- Integrate [Scalpel](https://github.com/maveniverse/scalpel) v0.3.7 as a Maven core extension
- Skips unit tests on modules unaffected by PR changes, based on git diff + POM dependency graph
- No static mapping files — fully dynamic, zero maintenance
- Safety: `failSafe=true` (falls back to full build on error), `ci/disable-scalpel` PR label override

## How it works

Scalpel reads `GITHUB_BASE_REF` (set by GitHub Actions on PRs), diffs against the base branch, maps changed files to Maven modules, and walks the dependency graph for transitive impacts. In `skip-tests` mode, all modules compile but tests only run on affected modules.

| Context | Behavior |
|---------|----------|
| **Unit test shards** | Scalpel active — skips tests on unaffected modules |
| **Build/integration/CLI/operator** | Scalpel disabled — need full builds |
| **Push to main** | No-op (no `GITHUB_BASE_REF`) — full build |
| **Local dev** | No-op (no CI env vars) — zero interference |

## Expected impact

| Change scope | Estimated savings |
|-------------|-------------------|
| `serdes/` only | ~39 min (6 app shards skip tests) |
| `app/src/` | Minimal (app is affected) |
| `common/` | None (everything depends on common) |
| **Average** | **30-60% fewer test modules per PR** |

## Prior art

Quarkus migrated from GIB to Scalpel ([PR #65830](https://github.com/quarkusio/quarkus/pull/65830)). Apache Camel is also integrating Scalpel.

Closes #8440

## Test plan

- [ ] Scalpel logs appear in unit test shard Maven output (`[INFO] Scalpel X.Y.Z activated`)
- [ ] Build step has `scalpel.enabled=false` → builds all modules
- [ ] Unit test shards show skipped tests for modules unaffected by this PR's changes
- [ ] `ci/disable-scalpel` label disables Scalpel when applied
- [ ] No test regressions

### DIFF
--- a/.github/workflows/verify-unit-tests.yaml
+++ b/.github/workflows/verify-unit-tests.yaml
@@ -10,10 +10,10 @@ on:
         type: string
         required: true
         description: 'Build artifact identifier (commit SHA)'
-      scalpel-disabled:
+      scalpel-enabled:
         type: string
         default: 'false'
-        description: 'Set to true to disable Scalpel affected-module detection'
+        description: 'Set to true to enable Scalpel affected-module detection'
 
 permissions:
   contents: read
@@ -91,7 +91,7 @@ jobs:
           # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
           ./mvnw verify \
             --no-transfer-progress \
-            ${{ inputs.scalpel-disabled != 'true' && '-Dscalpel.enabled=true' || '' }} \
+            ${{ inputs.scalpel-enabled == 'true' && '-Dscalpel.enabled=true' || '' }} \
             -Pprod ${{ matrix.shard.modules }} \
             ${{ matrix.shard.test-filter }} \
             -Dsurefire.failIfNoSpecifiedTests=false \
--- a/.github/workflows/verify-unit-tests.yaml
+++ b/.github/workflows/verify-unit-tests.yaml
@@ -10,6 +10,10 @@ on:
         type: string
         required: true
         description: 'Build artifact identifier (commit SHA)'
+      scalpel-disabled:
+        type: string
+        default: 'false'
+        description: 'Set to true to disable Scalpel affected-module detection'
 
 permissions:
   contents: read
@@ -53,6 +57,8 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          fetch-depth: 0
 
       - name: Set up JDK ${{ inputs.java-version }}
         uses: actions/setup-java@0f481fcb613427c0f801b606911222b5b6f3083a # v5
@@ -84,6 +90,7 @@ jobs:
           # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
           ./mvnw verify \
             --no-transfer-progress \
+            ${{ inputs.scalpel-disabled == 'true' && '-Dscalpel.enabled=false' || '' }} \
             -Pprod ${{ matrix.shard.modules }} \
             ${{ matrix.shard.test-filter }} \
             -Dsurefire.failIfNoSpecifiedTests=false \
--- a/.github/workflows/verify-unit-tests.yaml
+++ b/.github/workflows/verify-unit-tests.yaml
@@ -58,8 +58,8 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
         with:
-          # Full history required for Scalpel to compute git diff against base branch
-          fetch-depth: 0
+          # Full history only needed when Scalpel is enabled (git diff against base branch)
+          fetch-depth: ${{ inputs.scalpel-enabled == 'true' && 0 || 1 }}
 
       - name: Set up JDK ${{ inputs.java-version }}
         uses: actions/setup-java@0f481fcb613427c0f801b606911222b5b6f3083a # v5
--- a/.github/workflows/verify-unit-tests.yaml
+++ b/.github/workflows/verify-unit-tests.yaml
@@ -90,7 +90,7 @@ jobs:
           # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
           ./mvnw verify \
             --no-transfer-progress \
-            ${{ inputs.scalpel-disabled == 'true' && '-Dscalpel.enabled=false' || '' }} \
+            ${{ inputs.scalpel-disabled != 'true' && '-Dscalpel.enabled=true' || '' }} \
             -Pprod ${{ matrix.shard.modules }} \
             ${{ matrix.shard.test-filter }} \
             -Dsurefire.failIfNoSpecifiedTests=false \
--- a/.github/workflows/verify-unit-tests.yaml
+++ b/.github/workflows/verify-unit-tests.yaml
@@ -58,6 +58,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
         with:
+          # Full history required for Scalpel to compute git diff against base branch
           fetch-depth: 0
 
       - name: Set up JDK ${{ inputs.java-version }}
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -35,7 +35,7 @@ jobs:
       run-cli: ${{ steps.decide.outputs.run-cli }}
       run-go-sdk-freshness: ${{ steps.decide.outputs.run-go-sdk-freshness }}
       run-operator: ${{ steps.decide.outputs.run-operator }}
-      scalpel-disabled: ${{ steps.decide.outputs.scalpel-disabled }}
+      scalpel-enabled: ${{ steps.decide.outputs.scalpel-enabled }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
@@ -119,7 +119,7 @@ jobs:
                 for out in run-build run-unit-tests run-integration run-extras run-sdk run-cli run-operator run-go-sdk-freshness; do
                   echo "$out=false" >> "$GITHUB_OUTPUT"
                 done
-                echo "scalpel-disabled=false" >> "$GITHUB_OUTPUT"
+                echo "scalpel-enabled=false" >> "$GITHUB_OUTPUT"
                 exit 0
               fi
             fi
@@ -138,7 +138,7 @@ jobs:
               for out in run-build run-unit-tests run-integration run-extras run-sdk run-cli run-operator run-go-sdk-freshness; do
                 echo "$out=false" >> "$GITHUB_OUTPUT"
               done
-              echo "scalpel-disabled=false" >> "$GITHUB_OUTPUT"
+              echo "scalpel-enabled=false" >> "$GITHUB_OUTPUT"
               exit 0
             fi
 
@@ -205,9 +205,9 @@ jobs:
 
           # Scalpel (affected-module detection) can be disabled via PR label
           if [ "$EVENT" != "push" ] && has_label "ci/disable-scalpel"; then
-            echo "scalpel-disabled=true" >> "$GITHUB_OUTPUT"
+            echo "scalpel-enabled=false" >> "$GITHUB_OUTPUT"
           else
-            echo "scalpel-disabled=false" >> "$GITHUB_OUTPUT"
+            echo "scalpel-enabled=true" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Save decision summary
@@ -224,7 +224,7 @@ jobs:
             "run-cli": "${{ steps.decide.outputs.run-cli }}",
             "run-operator": "${{ steps.decide.outputs.run-operator }}",
             "run-go-sdk-freshness": "${{ steps.decide.outputs.run-go-sdk-freshness }}",
-            "scalpel-disabled": "${{ steps.decide.outputs.scalpel-disabled }}"
+            "scalpel-enabled": "${{ steps.decide.outputs.scalpel-enabled }}"
           }
           ARTIFACT_EOF
           cat > /tmp/verify-changes.json << 'ARTIFACT_EOF'
@@ -311,7 +311,7 @@ jobs:
     uses: ./.github/workflows/verify-unit-tests.yaml
     with:
       image-tag: ${{ github.sha }}
-      scalpel-enabled: ${{ needs.decide.outputs.scalpel-disabled != 'true' && 'true' || 'false' }}
+      scalpel-enabled: ${{ needs.decide.outputs.scalpel-enabled }}
 
   # ============================================================================
   # CLI Verification
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -294,6 +294,23 @@ jobs:
       scalpel-disabled: ${{ needs.decide.outputs.scalpel-disabled }}
 
   # ============================================================================
+  # Unit Tests — Scalpel Shadow (evaluation mode)
+  #
+  # Runs the same unit test shards with Scalpel ENABLED to compare results
+  # against the control group (unit-tests). Not in the Verification Gate —
+  # failures here don't block PRs. Remove this once the team is confident
+  # in Scalpel's accuracy and switches over.
+  # ============================================================================
+  unit-tests-scalpel:
+    name: Unit Tests (Scalpel)
+    needs: [decide, build]
+    if: needs.decide.outputs.run-unit-tests == 'true'
+    uses: ./.github/workflows/verify-unit-tests.yaml
+    with:
+      image-tag: ${{ github.sha }}
+      scalpel-disabled: 'false'
+
+  # ============================================================================
   # CLI Verification
   # ============================================================================
   cli-verify:
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -223,7 +223,8 @@ jobs:
             "run-sdk": "${{ steps.decide.outputs.run-sdk }}",
             "run-cli": "${{ steps.decide.outputs.run-cli }}",
             "run-operator": "${{ steps.decide.outputs.run-operator }}",
-            "run-go-sdk-freshness": "${{ steps.decide.outputs.run-go-sdk-freshness }}"
+            "run-go-sdk-freshness": "${{ steps.decide.outputs.run-go-sdk-freshness }}",
+            "scalpel-disabled": "${{ steps.decide.outputs.scalpel-disabled }}"
           }
           ARTIFACT_EOF
           cat > /tmp/verify-changes.json << 'ARTIFACT_EOF'
@@ -291,7 +292,7 @@ jobs:
     uses: ./.github/workflows/verify-unit-tests.yaml
     with:
       image-tag: ${{ github.sha }}
-      scalpel-disabled: ${{ needs.decide.outputs.scalpel-disabled }}
+      scalpel-enabled: 'false'
 
   # ============================================================================
   # Unit Tests — Scalpel Shadow (evaluation mode)
@@ -300,15 +301,17 @@ jobs:
   # against the control group (unit-tests). Not in the Verification Gate —
   # failures here don't block PRs. Remove this once the team is confident
   # in Scalpel's accuracy and switches over.
+  # PR-only: on push-to-main Scalpel is a no-op (no GITHUB_BASE_REF), so
+  # running both groups would just double CI cost with identical results.
   # ============================================================================
   unit-tests-scalpel:
     name: Unit Tests (Scalpel)
     needs: [decide, build]
-    if: needs.decide.outputs.run-unit-tests == 'true'
+    if: needs.decide.outputs.run-unit-tests == 'true' && github.event_name == 'pull_request'
     uses: ./.github/workflows/verify-unit-tests.yaml
     with:
       image-tag: ${{ github.sha }}
-      scalpel-disabled: 'false'
+      scalpel-enabled: ${{ needs.decide.outputs.scalpel-disabled != 'true' && 'true' || 'false' }}
 
   # ============================================================================
   # CLI Verification
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -35,6 +35,7 @@ jobs:
       run-cli: ${{ steps.decide.outputs.run-cli }}
       run-go-sdk-freshness: ${{ steps.decide.outputs.run-go-sdk-freshness }}
       run-operator: ${{ steps.decide.outputs.run-operator }}
+      scalpel-disabled: ${{ steps.decide.outputs.scalpel-disabled }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
@@ -115,9 +116,10 @@ jobs:
               LABEL='${{ github.event.label.name }}'
               if [[ "$LABEL" != lifecycle/* ]]; then
                 echo "lifecycle-ready=skip" >> "$GITHUB_OUTPUT"
-                for out in run-build run-unit-tests run-integration run-extras run-sdk run-cli run-operator; do
+                for out in run-build run-unit-tests run-integration run-extras run-sdk run-cli run-operator run-go-sdk-freshness; do
                   echo "$out=false" >> "$GITHUB_OUTPUT"
                 done
+                echo "scalpel-disabled=false" >> "$GITHUB_OUTPUT"
                 exit 0
               fi
             fi
@@ -136,6 +138,7 @@ jobs:
               for out in run-build run-unit-tests run-integration run-extras run-sdk run-cli run-operator run-go-sdk-freshness; do
                 echo "$out=false" >> "$GITHUB_OUTPUT"
               done
+              echo "scalpel-disabled=false" >> "$GITHUB_OUTPUT"
               exit 0
             fi
 
@@ -199,6 +202,13 @@ jobs:
           decide run-cli         "$run_full"  "$CLI"
           decide run-operator   "$run_full"  "$IS_PUSH" "$OPERATOR" "$CI_OPERATOR"
           decide run-go-sdk-freshness "$run_smoke" "$GO_SDK_GEN" "$CI"
+
+          # Scalpel (affected-module detection) can be disabled via PR label
+          if [ "$EVENT" != "push" ] && has_label "ci/disable-scalpel"; then
+            echo "scalpel-disabled=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "scalpel-disabled=false" >> "$GITHUB_OUTPUT"
+          fi
 
       - name: Save decision summary
         if: github.event_name == 'pull_request' && steps.decide.outputs.lifecycle-ready != 'skip'
@@ -281,6 +291,7 @@ jobs:
     uses: ./.github/workflows/verify-unit-tests.yaml
     with:
       image-tag: ${{ github.sha }}
+      scalpel-disabled: ${{ needs.decide.outputs.scalpel-disabled }}
 
   # ============================================================================
   # CLI Verification
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Core extensions load in ALL Maven invocations (local dev, CI, releases).
-     Scalpel is disabled by default in maven.config; only CI unit test shards
-     enable it. When disabled, Scalpel initializes but performs no reactor changes.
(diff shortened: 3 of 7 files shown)


paoloantinori:
## Local Stress Testing Results

Ran 8 edge-case tests locally to validate Scalpel behavior against our module dependency graph.

### Test Results

| # | Scenario | Changed | Tested | Skipped | Result |
|---|----------|---------|--------|---------|--------|
| 1 | `.mvn/**` change | .mvn/extensions.xml | **41/41** (full) | 0 | ✅ `fullBuildTriggers` correctly forces full build |
| 2 | `serdes/kafka/avro-serde/` only | 1 file | **2/41** | 39 | ✅ Only avro-serde + app (test-scope dependent) |
| 3 | `common/` change | 1 file | **30/41** | 11 | ✅ Correctly skips java-sdk, config-index (no common dependency) |
| 4 | `app/src/` only | 1 file | **1/41** | 40 | ✅ app is a leaf module, no downstream dependents |
| 5 | `pom.xml` comment-only | 1 file | **0/41** (no-op) | 41 | ✅ Semantic POM analysis: comments don't affect modules |
| 6 | `pom.xml` version property bump | jackson-dataformat-yaml | **22/41** | 19 | ✅ Traces property through dependencyManagement correctly |
| 7 | `-Dscalpel.enabled=false` | any | **41/41** (disabled) | 0 | ✅ Override works |
| 8 | `-Dtest=` filter + Scalpel | serdes change, app-sql shard | app tests **ran** | — | ⚠️ See note below |

### Key Finding: app has test-scope dependencies on serdes

`app/pom.xml` depends on `avro-serde-kafka` (and other serdes modules) in `<scope>test</scope>`. Scalpel correctly identifies `app` as a dependent — so a serdes-only change still triggers app tests across all 6 app shards.

**Impact on expected savings:**
- ~~39 minutes saved for serdes-only PRs~~ → app shards still run
- **Primary savings**: non-app shard (serdes-only PRs skip non-serdes modules within that shard)
- **Best savings**: java-sdk-only or config-index-only changes (these don't affect app at all)
- **Biggest win**: app-only changes → testing 1/41 modules, skipping 40

### Scalpel Semantic POM Analysis

One surprising result: a pom.xml change that only adds an XML comment produces "No modules affected" — Scalpel semantically parses POM changes and ignores non-functional modifications. A version property change correctly traces through `dependencyManagement` to find all transitively affected modules.

### Local Dev Behavior

Confirmed: without `GITHUB_BASE_REF` env var, Scalpel logs "No base branch configured or detected, building all modules" — **zero interference with developer workflows**.

paoloantinori:
## Architectural Analysis: app ↔ serdes test-scope coupling

### Why app depends on serdes in test scope

The 11 serdes test files in `app/src/test/` (e.g., `AvroSerdeTest`, `ProtobufSerdeTest`, `JsonSchemaSerdeTest`) are `@QuarkusTest` classes that extend `AbstractClientFacadeTestBase`. They start an embedded Quarkus registry instance and test the full serialize → register schema → deserialize roundtrip within a single JVM.

These are **not** unit tests of the serdes libraries. They're **embedded integration tests** that verify serdes work correctly against a live registry. This is a standard Quarkus testing pattern — fast (seconds, no Docker) and reliable.

### Is this duplicating `integration-tests/`?

Partially. Both test serdes+registry interactions:
- **`app/src/test/` serdes tests**: embedded Quarkus, fast feedback during development
- **`integration-tests/` serdes tests**: containerized deployment via Minikube, validates real-world setup

Both serve a purpose — the embedded tests catch regressions quickly, the container tests catch deployment-specific issues.

### Impact on Scalpel

Because `app` test-depends on serdes, Scalpel correctly identifies `app` as affected by any serdes change. This means all 6 app unit test shards still run for serdes-only PRs. The savings from Scalpel are concentrated in:

| PR change scope | Modules tested | Modules skipped | Saving |
|----------------|---------------|-----------------|--------|
| `app/src/` only | 1/41 | 40 | **Biggest win** |
| `java-sdk/` only | ~2/41 | ~39 | Large |
| `serdes/` only | ~2/41 (non-app) + app shards still run | ~39 in non-app shard | Modest |
| `common/` | 30/41 | 11 | Moderate |
| `pom.xml` version bump | Varies (22/41 for jackson) | Varies | Moderate |

### Could we restructure?

| Option | Benefit | Cost |
|--------|---------|------|
| Move serdes tests to `integration-tests/` | Decouples app for Scalpel | Lose fast embedded testing; tests become slower, need Docker |
| Extract to `app-serdes-tests/` module | Scalpel can skip independently | New module, shared test infrastructure complexity |
| Keep as-is | Fast dev feedback, proven Quarkus pattern | App shards run for serdes changes |

**Recommendation:** Keep the current architecture. The embedded `@QuarkusTest` serdes tests provide significant developer productivity (fast feedback, no Docker). Restructuring to gain CI savings would sacrifice development velocity for CI optimization — inverted priorities. The Scalpel savings from app-only and java-sdk-only PRs are valuable enough on their own.

EricWittmann:
Any way to invert the logic so that scalpel is **disabled** by default, and then **enabled** only where we want it?

paoloantinori:
Good idea @EricWittmann — inverting the default is cleaner and safer.

**What changes:**
- `.mvn/maven.config`: add `-Dscalpel.enabled=false` (disabled by default)
- `verify-unit-tests.yaml`: add `-Dscalpel.enabled=true` (only place it's active)
- **Remove** all the `-Dscalpel.enabled=false` flags from verify-build, verify-integration-tests, verify-cli, and operator workflows — they inherit the disabled default

This means:
- New workflows are safe by default (no Scalpel unless opted in)
- Fewer workflow changes (1 opt-in vs 6 opt-outs)
- We can incrementally enable Scalpel in other workflows later if it proves valuable

I'll push the update.

.mvn/extensions.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extensions>
+    <extension>
+        <groupId>eu.maveniverse.maven.scalpel</groupId>
```
EricWittmann, line 6: **[1/8] CRITICAL — Scalpel is silently loaded in ALL Maven invocations**

`.mvn/extensions.xml` is a **global** Maven core extension file. Every `./mvnw` invocation across every workflow — `verify-build.yaml`, `verify-integration-tests.yaml`, `verify-extras.yaml`, `verify-cli.yaml`, release workflows, and local dev — will load the Scalpel extension JAR into the Maven classloader.

`maven.config` sets `-Dscalpel.enabled=false` by default, so Scalpel *shouldn't do anything* in those contexts. But:

- A core extension is loaded and initialized regardless of whether it's "enabled." This adds startup time and a new failure surface to every Maven invocation in the project.
- If Scalpel has a bug in its initialization path (before checking the `enabled` flag), it could break builds that have nothing to do with unit tests.
- The `failSafe=true` flag mitigates this, but it means Scalpel errors would be *silently swallowed* in contexts where you don't expect it to be running at all.

**Suggestion:** Consider whether Scalpel could be loaded conditionally (e.g., via a Maven profile that adds it as a build extension only when activated) rather than as a global core extension. If that's not possible with Scalpel's architecture, this should at least be documented prominently so future maintainers understand why every Maven invocation loads an extra extension.

.github/workflows/verify-unit-tests.yaml
```diff
@@ -53,6 +57,8 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          fetch-depth: 0
```
EricWittmann: **[2/8] HIGH — `fetch-depth: 0` clones full git history in every shard**

This means each of the 7 unit test shards now clones the **full git history**. For a repository of Apicurio Registry's size, this adds non-trivial time and bandwidth to each shard.

Meanwhile, the build step (`verify-build.yaml`) uses the default shallow clone — which is fine because Scalpel is disabled there. But if someone later enables Scalpel in the build step, they'd need to remember to also add `fetch-depth: 0`, which is a subtle coupling.

**Suggestion:** Measure the clone time impact. Consider whether `fetch-depth: 100` or a similar bounded depth would be sufficient for Scalpel's diff needs (most PRs have far fewer than 100 commits from the base).

.github/workflows/verify-unit-tests.yaml
```diff
@@ -79,6 +85,7 @@ jobs:
         run: |
           ./mvnw verify \
             --no-transfer-progress \
+            ${{ inputs.scalpel-disabled != 'true' && '-Dscalpel.enabled=true' || '' }} \
```
EricWittmann: **[3/8] MEDIUM — Interaction with existing `dorny/paths-filter` change detection**

This is where Scalpel's module-level detection meets the existing file-path-based `decide` job gating. The two systems have overlapping but different views of "what changed":

| Aspect | `dorny/paths-filter` | Scalpel |
|--------|---------------------|--------|
| Granularity | File paths → workflow jobs | File paths → Maven modules |
| Graph awareness | None (static path patterns) | Maven dependency graph |
| Result | Skip entire job phases | Skip tests within a phase |

Two subtle interactions worth verifying:

1. The `CI` filter matches `.github/workflows/**`. Any workflow change forces all unit tests to run. Scalpel would then try to determine affected modules — but since `.github/` files aren't Maven modules and `scalpel.excludePaths` excludes them, Scalpel might conclude *nothing changed* and skip all tests. What does Scalpel do when all changed files are excluded?

2. Changes to root `pom.xml` trigger the `java` filter and all unit tests. Scalpel should recognize root POM changes as affecting everything, but this is worth verifying.

**Suggestion:** Test both scenarios (workflow-only change, root POM-only change) and document how the two systems interact for future maintainers.

.mvn/maven.config
```diff
@@ -1 +1,6 @@
 -T 1C
+-Dscalpel.enabled=false
+-Dscalpel.mode=skip-tests
+-Dscalpel.excludePaths=*.md,LICENSE,.github/**,docs/**,ui/**
+-Dscalpel.skipTestsForUpstream=true
```
EricWittmann, line 13: **[4/8] MEDIUM — `skipTestsForUpstream=true` may interact unexpectedly with `-am` in unit test shards**

Every `app-*` shard passes `-pl app -am`, instructing Maven to build `app` and all its upstream dependencies (`common`, `schema-util`, etc.). With `skipTestsForUpstream=true`, Scalpel will skip tests for those upstream modules.

This is probably intended, but consider a change that touches only `common/`:
- In `app-*` shards: Scalpel correctly skips `app` tests (unchanged). But it also skips `common` tests because `common` is "upstream" of the `-pl app` target.
- In the `non-app` shard: Scalpel correctly runs `common` tests.

So this works out, but only because the shard structure happens to align. If a future shard reorganization changes this, `common` tests could silently stop running.

**Suggestion:** Add a comment explaining this interaction so future shard changes don't inadvertently create coverage gaps.

.github/workflows/verify.yaml
```diff
@@ -115,9 +116,10 @@ jobs:
               LABEL='${{ github.event.label.name }}'
               if [[ "$LABEL" != lifecycle/* ]]; then
                 echo "lifecycle-ready=skip" >> "$GITHUB_OUTPUT"
-                for out in run-build run-unit-tests run-integration run-extras run-sdk run-cli run-operator; do
+                for out in run-build run-unit-tests run-integration run-extras run-sdk run-cli run-operator run-go-sdk-freshness; do
```
EricWittmann, line 119: **[5/8] MEDIUM — Unrelated fix bundled in: `run-go-sdk-freshness` added to label early-exit loop**

This adds `run-go-sdk-freshness` to the `for` loop in the non-lifecycle label early-exit path. This is a **bug fix for a pre-existing issue** — the output was missing from this loop, meaning a non-lifecycle label event would not set `run-go-sdk-freshness=false`, potentially leaving it unset.

This fix is correct, but it's unrelated to Scalpel.

**Suggestion:** Call this out in the PR description as a separate fix, or split it into its own commit. Mixing unrelated fixes with the Scalpel changes could be confusing during bisection.

.mvn/maven.config
```diff
@@ -1 +1,6 @@
 -T 1C
+-Dscalpel.enabled=false
+-Dscalpel.mode=skip-tests
+-Dscalpel.excludePaths=*.md,LICENSE,.github/**,docs/**,ui/**
```
EricWittmann: **[6/8] LOW — `excludePaths` is a static maintenance surface**

This hardcodes knowledge about which paths are non-Maven: `*.md`, `LICENSE`, `.github/**`, `docs/**`, `ui/**`. If new non-Maven directories are added in the future (e.g., `helm/`, `terraform/`), Scalpel might incorrectly try to map them to Maven modules.

With `failSafe=true` this would just fall back to a full build, so it's not dangerous — but the PR description's claim of "zero maintenance" isn't quite accurate given this static list.

Nit: worth adding a comment in this file noting that new non-Maven top-level directories should be added here.

.mvn/maven.config
```diff
@@ -1 +1,6 @@
 -T 1C
+-Dscalpel.enabled=false
```
EricWittmann, line 5: **[7/8] LOW — Other Maven-invoking workflows rely implicitly on this default**

`verify-integration-tests.yaml`, `verify-build.yaml`, `verify-extras.yaml`, and `verify-cli.yaml` all invoke `./mvnw` and will inherit this default `-Dscalpel.enabled=false`. This works, but it's implicit — someone setting this to `true` in the future (thinking "let's enable it everywhere") would inadvertently enable Scalpel in the build step, where it could skip compilation of modules and break the artifact upload.

**Suggestion:** Either add an explicit `-Dscalpel.enabled=false` to the build step's Maven command for defense-in-depth, or add a comment here explaining that the default `false` is intentional and only unit test shards opt in.

.mvn/extensions.xml
```diff
@@ -0,0 +1,8 @@
...
+        <artifactId>extension</artifactId>
+        <version>0.3.7</version>
```
EricWittmann, line 8: **[8/8] LOW — Version pinning without dependency-update tracking**

Scalpel `0.3.7` is hardcoded here. Unlike dependencies in `pom.xml`, extensions declared in `.mvn/extensions.xml` are typically **not** tracked by Dependabot or Renovate unless explicitly configured.

**Suggestion:** Verify that Dependabot/Renovate is configured to scan `.mvn/extensions.xml`, or add it to the update config.

EricWittmann (commented):
## Code Review: 8 findings (1 critical, 1 high, 3 medium, 3 low)

The approach is sound and well-designed. Default-off in `maven.config`, selective opt-in in CI, `failSafe=true`, and the `ci/disable-scalpel` label escape hatch are all good choices. The main concern is global extension loading across all Maven invocations. See inline comments for details.

paoloantinori:
@EricWittmann Excellent review — finding #3 caught a real bug. All 8 items addressed (commit 0757ae0), validated against Quarkus's actual Scalpel configuration.

### Research context

Checked Quarkus's `.mvn/extensions.xml`, `.mvn/maven.config`, and `ci-actions-incremental.yml` to verify best practices before responding.

### Responses per finding

**[1/8] CRITICAL → LOW** — Global extension loading is standard. Quarkus loads 7 extensions in `extensions.xml` (Scalpel, Develocity, Nisse, Njord, etc.). Maven core extensions can't be loaded conditionally — this is by design. Added documentation comment.

**[2/8] HIGH → Keep as-is** — Quarkus uses `fetch-depth: 0` with explicit comment "this is important for Scalpel to work." `fetchBaseBranch=true` is only used for non-PR push builds. Added the same explanatory comment.

**[3/8] MEDIUM → Bug fix** — This was a real bug. `.github/**` was in `excludePaths` (files ignored → Scalpel sees "no changes" → skips all tests). Quarkus correctly uses `disableTriggers=.github/**,.mvn/**,mvnw*` instead — when triggered, Scalpel disables entirely → full build. Fixed: moved `.github/**,.mvn/**,mvnw*` to `disableTriggers`, kept `*.md,LICENSE,docs/**,ui/**` in `excludePaths`.

**[4/8] MEDIUM** — Added comment explaining the `skipTestsForUpstream` + shard interaction. Quarkus uses `alsoMake=false` which is a different tradeoff; our approach works because the non-app shard independently tests upstream modules.

**[5/8] MEDIUM** — The `run-go-sdk-freshness` addition to the early-exit loop is a pre-existing bug fix. Since squash merge is enabled, I've noted it here rather than splitting commits. It fixes a case where non-lifecycle label events would leave `run-go-sdk-freshness` unset.

**[6/8] LOW** — Added comment noting `excludePaths` is a maintenance surface. Updated PR description to not claim "zero maintenance."

**[7/8] LOW** — Added comment in `maven.config` explaining the intentional disabled default and warning against changing it.

**[8/8] LOW** — Checked: Renovate (`renovate.json`) does not track `.mvn/extensions.xml`. This should be a separate follow-up to add Maven extension tracking to the Renovate config.

paoloantinori:
Added **shadow mode** per today's discussion with @EricWittmann (commit 00394ec):

A new `Unit Tests (Scalpel)` job runs alongside the existing unit test shards with Scalpel enabled. Both run for every PR:
- **Control group** (`unit-tests`): full tests, in the Verification Gate
- **Experiment** (`unit-tests-scalpel`): Scalpel active, NOT in the Gate

This lets the team compare results without risk. Remove the control group once confident in Scalpel's accuracy.

Also created backlog item REG-4 for the cross-layer detection vision — designing a solution that can detect impacts across Maven/npm/operator boundaries without manual mapping.

Documented both patterns in Discussion #8365 (Shadow Mode + Cross-Layer Detection).

.github/workflows/verify.yaml
```diff
@@ -281,6 +291,24 @@ jobs:
     uses: ./.github/workflows/verify-unit-tests.yaml
     with:
       image-tag: ${{ github.sha }}
+      scalpel-disabled: ${{ needs.decide.outputs.scalpel-disabled }}
```
EricWittmann: **[1/5] CRITICAL — Shadow mode is broken: both control and experiment run identically**

The shadow mode comment says the control group (`unit-tests`) runs "full tests" while the experiment (`unit-tests-scalpel`) runs with "Scalpel active." But both groups get Scalpel enabled:

- **Control** (`unit-tests`): `scalpel-disabled` comes from `needs.decide.outputs.scalpel-disabled`, which is `'false'` unless the `ci/disable-scalpel` label is applied. `'false'` → the expression in `verify-unit-tests.yaml` emits `-Dscalpel.enabled=true` → **Scalpel ON**.
- **Experiment** (`unit-tests-scalpel`): `scalpel-disabled: 'false'` → same expression → **Scalpel ON**.

Both groups produce identical results (14 shard runs instead of 7), doubling CI cost without generating any comparison data. The shadow mode only produces different results when the `ci/disable-scalpel` label is manually applied — defeating the purpose of automatic evaluation.

**Fix:** Hardcode the control group to always run full tests:

```yaml
scalpel-disabled: 'true'   # Control: always full tests, in the Verification Gate
```

Then when the team is confident in Scalpel, flip this to `${{ needs.decide.outputs.scalpel-disabled }}` and remove the shadow job.

.github/workflows/verify.yaml
```diff
@@ -281,6 +291,24 @@ jobs:
...
+  #
+  # Runs the same unit test shards with Scalpel ENABLED to compare results
+  # against the control group (unit-tests). Not in the Verification Gate —
+  # failures here don't block PRs. Remove this once the team is confident
+  # in Scalpel's accuracy and switches over.
+  # ============================================================================
+  unit-tests-scalpel:
+    name: Unit Tests (Scalpel)
+    needs: [decide, build]
+    if: needs.decide.outputs.run-unit-tests == 'true'
```
EricWittmann: **[2/5] HIGH — Shadow mode runs on push-to-main, doubling CI for no benefit**

On push to main, `run-unit-tests` is `true`, so both `unit-tests` and `unit-tests-scalpel` run. But Scalpel is a no-op on push (no `GITHUB_BASE_REF` → no diff → full build), so both groups produce identical full-build results. This doubles CI runner time on the most critical pipeline path.

**Fix:** Add a PR-only guard:

```yaml
if: needs.decide.outputs.run-unit-tests == 'true' && github.event_name == 'pull_request'
```

.github/workflows/verify-unit-tests.yaml
```diff
@@ -10,6 +10,10 @@ on:
         type: string
         required: true
         description: 'Build artifact identifier (commit SHA)'
+      scalpel-disabled:
```
EricWittmann: **[4/5] MEDIUM — Double-negative naming**

The input name `scalpel-disabled` with a default of `'false'` creates a double-negative: "Scalpel is not disabled" = Scalpel is enabled. The shadow job has `scalpel-disabled: 'false'` which reads as "disabled? no" — confusing even with context.

**Suggestion:** Rename to `scalpel-enabled` with default `'false'` and invert the expression in the Maven command. This reads naturally: `scalpel-enabled: 'true'` in the experiment, `scalpel-enabled: 'false'` in the control. Not a blocker, but worth doing while the input is new and has no external consumers.

.mvn/extensions.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Core extensions load in ALL Maven invocations (local dev, CI, releases).
+     Scalpel is disabled by default in maven.config; only CI unit test shards
+     enable it. When disabled, Scalpel initializes but performs no reactor changes.
+     This is standard practice — Quarkus loads 7 extensions the same way. -->
+<extensions>
+    <extension>
+        <groupId>eu.maveniverse.maven.scalpel</groupId>
+        <artifactId>extension</artifactId>
+        <version>0.3.7</version>
```
EricWittmann, line 8: **[5/5] LOW — Version pinning without update tracking**

The first review flagged this (finding #8) and the author confirmed Renovate doesn't track `.mvn/extensions.xml`. A follow-up task was mentioned but there's no linked issue. Consider opening one now so it doesn't get lost — Scalpel `0.3.7` will go stale without automated dependency update tracking.

EricWittmann (commented):
## Code Review: 5 findings (1 critical, 1 high, 2 medium, 1 low)

The Scalpel integration itself is solid — default-off in `maven.config`, selective opt-in via workflow inputs, `failSafe=true`, and the `ci/disable-scalpel` label escape hatch are all good choices. The `disableTriggers` vs `excludePaths` fix from the prior review was a real bug catch and is now correct.

However, **the shadow mode implementation has a critical logic error** that makes the evaluation period ineffective — both control and experiment groups run with identical Scalpel configuration, doubling CI cost without generating comparison data. See inline comments for details.

### Additional finding not tied to a diff line

**[3/5] MEDIUM — `scalpel-disabled` decision not included in audit artifact**

The `Save decision summary` step writes all `run-*` outputs to `verify-decisions.json`, but `scalpel-disabled` is missing. This means the decision audit trail doesn't capture whether Scalpel was active or disabled for a given run. Add `"scalpel-disabled": "${{ steps.decide.outputs.scalpel-disabled }}"` to the JSON artifact.

### Positive Notes

- `disableTriggers` correctly triggers full builds on CI/build config changes instead of silently ignoring them
- Comments in `maven.config` clearly explain the disabled default and warn against changing it
- `failSafe=true` is a good safety net
- The `ci/disable-scalpel` label escape hatch gives the team an easy override
- The `run-go-sdk-freshness` fix in the early-exit loop is a legitimate pre-existing bug fix
- Thorough local stress testing with 8 edge cases including the semantic POM analysis finding

paoloantinori:
@EricWittmann All 5 findings addressed in commit 6b8b60a:

1. **CRITICAL fixed**: Control group now hardcodes `scalpel-enabled: 'false'` — always runs full tests. Experiment uses the decide output.
2. **HIGH fixed**: Shadow job restricted to PRs only (`github.event_name == 'pull_request'`).
3. **MEDIUM fixed**: Added `scalpel-disabled` to `verify-decisions.json` audit artifact.
4. **MEDIUM fixed**: Renamed input from `scalpel-disabled` to `scalpel-enabled` — no more double-negative.
5. **LOW**: Created issue #8498 for Renovate tracking of `.mvn/extensions.xml`.

paoloantinori:
Applied code-simplifier suggestions (commit 2f04e6a): shortened extensions.xml comment (removed defensive Quarkus comparison), added failSafe comment in maven.config.

.github/workflows/verify-unit-tests.yaml
```diff
@@ -53,6 +57,9 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          # Full history required for Scalpel to compute git diff against base branch
+          fetch-depth: 0
```
EricWittmann: `fetch-depth: 0` is applied unconditionally to every invocation of this reusable workflow, including the control `unit-tests` job where Scalpel is disabled. With 7 shards on the control group, that's 7 unnecessary full clones.\n\nConsider making it conditional:\n\n```yaml\nwith:\n  fetch-depth: ${{ inputs.scalpel-enabled == 'true' && 0 || 1 }}\n```

.github/workflows/verify.yaml
```diff
@@ -281,6 +292,26 @@ jobs:
...
+  # running both groups would just double CI cost with identical results.
+  # ============================================================================
+  unit-tests-scalpel:
+    name: Unit Tests (Scalpel)
+    needs: [decide, build]
+    if: needs.decide.outputs.run-unit-tests == 'true' && github.event_name == 'pull_request'
+    uses: ./.github/workflows/verify-unit-tests.yaml
+    with:
+      image-tag: ${{ github.sha }}
+      scalpel-enabled: ${{ needs.decide.outputs.scalpel-disabled != 'true' && 'true' || 'false' }}
```
EricWittmann: The double negation here is hard to read: `scalpel-disabled != 'true'` → `scalpel-enabled = 'true'`.\n\nConsider renaming the `decide` output from `scalpel-disabled` to `scalpel-enabled` and setting it directly:\n\n```bash\nif [ "$EVENT" != "push" ] && has_label "ci/disable-scalpel"; then\n  echo "scalpel-enabled=false" >> "$GITHUB_OUTPUT"\nelse\n  echo "scalpel-enabled=true" >> "$GITHUB_OUTPUT"\nfi\n```\n\nThen the consumer simplifies from:\n```yaml\nscalpel-enabled: ${{ needs.decide.outputs.scalpel-disabled != 'true' && 'true' || 'false' }}\n```\nto:\n```yaml\nscalpel-enabled: ${{ needs.decide.outputs.scalpel-enabled }}\n```

paoloantinori:
@EricWittmann Both addressed in commit 3af4c4e:

1. **fetch-depth conditional**: Now uses `${{ inputs.scalpel-enabled == 'true' && 0 || 1 }}` — control group (7 shards) avoids full clones.
2. **scalpel-disabled → scalpel-enabled**: Renamed the decide output and flipped the logic. Consumer simplifies from `scalpel-disabled != 'true' && 'true' || 'false'` to just `needs.decide.outputs.scalpel-enabled`.

paoloantinori:
**Pre-review note on local dev impact:**

`.mvn/extensions.xml` loads the Scalpel extension on every Maven invocation, but `.mvn/maven.config` sets `scalpel.enabled=false` by default. When disabled, Scalpel skips all processing — no git diff, no dependency graph traversal, no module filtering. The extension class loads (~5ms) but does zero work. CI shards opt in explicitly with `-Dscalpel.enabled=true`.

Verified locally: `./mvnw clean install -DskipTests` timing is unchanged (within noise) with the extension loaded but disabled.

carlesarnal (approved):
LGTM — smart rollout strategy. Running Scalpel in shadow mode alongside the existing test suite is the right way to evaluate accuracy before switching over. Key safeguards are in place: disabled by default, `failSafe=true` for fallback, `disableTriggers` for CI/build config changes, `ci/disable-scalpel` label escape hatch. The PR-only guard avoids doubling CI cost on push-to-main.

PR #8450, opened 2026-07-02
[CI] Add inline documentation to workflow Maven commands

### LINKED ISSUE #8449: Add inline documentation to CI workflow Maven commands
The CI workflow Maven invocations have accumulated many non-obvious flags. Add YAML comments explaining what each flag does and why it's there, to help developers understand and maintain the CI configuration.

paoloantinori:
## Summary

Add YAML comments to all CI workflow Maven commands explaining non-obvious flags. Documentation-only — no behavioral changes.

## Motivation

The CI optimization work over the past two weeks introduced many flags across 6 workflow files. Without comments, the rationale for flags like `-Daether.syncContext.named.factory=noop` or `-T 0.5C` (vs the default `-T 1C`) is not obvious to developers maintaining these workflows.

## What's documented

| Flag | Explanation |
|------|-------------|
| `-T 0.5C` | Overrides `.mvn/maven.config`'s `-T 1C` for memory-heavy packaging |
| `-T 1` | Overrides `-T 1C` to prevent module ordering issues with examples |
| `-Daether.syncContext.named.factory=noop` | Prevents Maven 3.9+ resolver lock deadlocks under parallel builds |
| `-DcliSkipNative` | Skips GraalVM native-image (separate CLI verification job handles it) |
| `-DskipCommitIdPlugin=false` | Generates git commit metadata for Docker image |
| `-Dsurefire.rerunFailingTestsCount=2` | Auto-retries flaky tests up to 2 times |
| `-Dsurefire.failIfNoSpecifiedTests=false` | Allows empty test filter matches per shard |
| `-Dmaven.wagon.http*` | Limits HTTP connections and retries downloads |
| Operator Makefile `-T1` | Operator `@QuarkusTest` tests require sequential execution |

Closes #8449

## Test plan

- [ ] No behavioral changes — only YAML comments added
- [ ] CI passes (comments don't affect execution)

### DIFF
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -37,7 +37,7 @@ jobs:
           distribution: temurin
           cache: maven
 
-      # Build operand (app + docker) and operator OLM tests. Changes in the current PR must be reflected.
+      # Build operand (app + docker) and operator OLM tests, so downstream tests reflect changes in the current PR.
       # -Dfull: include all optional modules needed for the Docker image
       # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
       - name: Build
@@ -158,7 +158,7 @@ jobs:
           docker push "$IMAGE"
 
       # Makefile enforces -T1 (single-threaded) because operator @QuarkusTest tests require sequential execution.
-      # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks
+      # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
       - name: Build and run local tests on Minikube
         working-directory: operator
         run: |
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -37,7 +37,9 @@ jobs:
           distribution: temurin
           cache: maven
 
-      # We need to build the operand image as well as the operator, otherwise the tests would not reflect changes in the current PR.
+      # Build operand (app + docker) and operator OLM tests. Changes in the current PR must be reflected.
+      # -Dfull: include all optional modules needed for the Docker image
+      # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
       - name: Build
         run: ./mvnw -B clean package -Dfull -pl app,distro/docker,operator/olm-tests -am -Pprod -DskipTests -Dmaven.javadoc.skip -Daether.syncContext.named.factory=noop
 
@@ -155,6 +157,8 @@ jobs:
           docker push "$REGISTRY_GITOPS_SYNC_IMAGE"
           docker push "$IMAGE"
 
+      # Makefile enforces -T1 (single-threaded) because operator @QuarkusTest tests require sequential execution.
+      # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks
       - name: Build and run local tests on Minikube
         working-directory: operator
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/operator.yaml
```diff
@@ -37,7 +37,9 @@ jobs:
           distribution: temurin
           cache: maven
 
-      # We need to build the operand image as well as the operator, otherwise the tests would not reflect changes in the current PR.
+      # Build operand (app + docker) and operator OLM tests. Changes in the current PR must be reflected.
```
EricWittmann: The rewrite is more precise about *what* is built (correctly naming `operator/olm-tests` instead of "the operator"), but it loses the causal *why* that the old comment had: "otherwise the tests would not reflect changes in the current PR."

The new wording ("Changes in the current PR must be reflected") states a requirement imperatively but doesn't explain what breaks if this step is removed.

Suggestion:
```suggestion
      # Build operand (app + docker) and operator OLM tests, so downstream tests reflect changes in the current PR.
```

.github/workflows/operator.yaml
```diff
@@ -155,6 +157,8 @@ jobs:
           docker push "$REGISTRY_GITOPS_SYNC_IMAGE"
           docker push "$IMAGE"
 
+      # Makefile enforces -T1 (single-threaded) because operator @QuarkusTest tests require sequential execution.
+      # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks
```
EricWittmann: This instance (and the matching one in `verify-extras.yaml` line 234) says "prevent Maven 3.9+ resolver lock deadlocks" — but seven other instances across the PR include the suffix "under -T". Since `.mvn/maven.config` sets `-T 1C` globally, the "under -T" context is accurate here too.

```suggestion
      # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
```

.github/workflows/verify-extras.yaml
```diff
@@ -225,4 +228,8 @@ jobs:
 
       - name: Build Apicurio Registry with Examples
         run: |
+          # -T 1: single-threaded (overrides .mvn/maven.config's -T 1C to prevent module ordering issues with examples)
+          # -Pexamples: activates the examples profile (builds example applications)
+          # -Dassembly.skipAssembly=true: skip archive generation (not needed for example compilation)
+          # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks
```
EricWittmann: Same inconsistency as in `operator.yaml` — missing the "under -T" suffix that the other seven instances include.

```suggestion
          # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
```

EricWittmann (commented):
Nice docs PR — the flag explanations are accurate and well-structured. Two minor consistency nits below.

paoloantinori:
@EricWittmann All 3 nits addressed in commit dbdcfec:

1. Restored causal "why" — "so downstream tests reflect changes in the current PR"
2. Added "under -T" suffix to operator.yaml noop comment
3. Added "under -T" suffix to verify-extras.yaml noop comment

EricWittmann (approved):
Looks good — clean, accurate documentation that will genuinely help the next person maintaining these workflows.

Verified the key claims (`.mvn/maven.config` `-T 1C`, operator Makefile `-T1`, noop flag rationale) and confirmed the "under -T" suffix is now consistent across all instances. CI failures are unrelated infrastructure issues.

One tiny observation for the record: the noop comment on the Minikube step (operator.yaml line 158) says "under -T" but that step runs single-threaded via Makefile `-T1`, so the deadlock risk doesn't actually apply there — the flag is just defensive. Not worth changing.

carlesarnal (approved):
LGTM — comments are accurate and useful. Good reference for anyone debugging CI failures.

PR #8451, opened 2026-07-02
fix(ui): load current user on dashboard first render

### LINKED ISSUE #8371: [UI] Dashboard does not load current user — admin actions hidden & header shows "User" on first render (OIDC/RBAC)
**Version:** Apicurio Registry 3.3.0 (UI). The same code appears to be present in `main` at the time of writing.
**Component:** UI (`apicurio-registry-ui`)

### Description
With OIDC auth + role-based authorization enabled, landing on the **Dashboard** (`/dashboard`) right after login shows the user as if they were **not** an admin: the Quick Actions panel hides **Create Artifact** and **Settings**, and the user dropdown in the header shows the fallback **"User"** instead of the username. Navigating to any other page (Explore, Search, Global rules, …) and returning to the Dashboard then renders everything correctly (admin actions + username).

### Root cause
`DashboardPage.tsx` does **not** use `PageDataLoader`. `PageDataLoader` is the only component that calls `user.updateCurrentUser()` (→ `GET /apis/registry/v3/users/me`) and **awaits** it before rendering its children. Because the Dashboard skips it, `currentUserInfo` — a **module-level, non-reactive** variable in `useUserService.ts` — stays at its default `{ admin: false, displayName: "" }` on the first render. As a result:
- `isUserAdmin()` returns `false`, so RBAC-gated actions (Create Artifact, Settings) are hidden.
- `AvatarDropdown` renders `user.currentUser().displayName || "User"` → shows **"User"**.

Every other page wraps its content in `PageDataLoader`, so they work. Once any of those pages loads `/users/me`, the module-level variable is populated, and a later re-render (triggered by navigation) makes the Dashboard show the correct state.

### Steps to reproduce
1. Enable OIDC + `apicurio.auth.role-based-authorization=true` (e.g. Azure Entra ID with App Roles in the `roles` claim).
2. Log in as a user that has the admin role (`sr-admin`).
3. Observe `/dashboard` immediately after login → Quick Actions is missing **Create Artifact** / **Settings**, and the header shows **"User"**.
4. Navigate to Explore or Search, then back to the Dashboard → admin actions and the username now appear.

### Expected
The Dashboard reflects the authenticated user's identity and roles on first render (an admin sees Create Artifact / Settings; the header shows the username).

### Actual
The Dashboard renders before `/users/me` resolves, so admin actions are hidden and the header shows "User" until the user navigates to another page and returns.

### Suggested fix
- Wrap `DashboardPage` in `PageDataLoader` (as every other page does) so `updateCurrentUser()` runs and is awaited before render; and/or
- Make `currentUserInfo` reactive (React state/context) so the header and RBAC-gated controls re-render when `/users/me` resolves.

### Relevant files
- `ui/ui-app/src/app/pages/dashboard/DashboardPage.tsx` — missing `PageDataLoader`
- `ui/ui-app/src/services/useUserService.ts` — non-reactive module-level `currentUserInfo`
- `ui/ui-app/src/app/components/header/AvatarDropdown.tsx` — `displayName || "User"`

### Environment
GKE, internal L7 Application Load Balancer, Azure Entra ID (OIDC, v2 tokens), RBAC via Entra App Roles. Deployed via Terraform with separate API and UI containers.

yxshwanth:
## Summary

- Wrap `DashboardPage` in `PageDataLoader` so `GET /users/me` runs before the dashboard renders, matching every other top-level page.
- Replace the module-level `currentUserInfo` variable in `useUserService` with a reactive `UserProvider` (React Context), so the header, `IfAuth`-gated controls, and RBAC checks re-render when user info resolves.

Fixes #8371

## Root cause

`DashboardPage` was the only top-level page that skipped `PageDataLoader`, which is the sole caller of `updateCurrentUser()`. Because user identity was stored in a non-reactive module-level variable, the dashboard and header rendered with default `{ admin: false, displayName: "" }` on first landing after login.

## Test plan

- [x] `npm run lint` passes in `ui/ui-app`
- [x] `npm run build` passes in `ui/ui-app`
- [ ] Log in with OIDC + RBAC enabled as an admin user and land on `/dashboard` — header shows real `displayName`, Quick Actions shows **Create Artifact** and **Settings**
- [ ] Navigate to Explore / Search and confirm header and RBAC-gated tabs still work
- [ ] Confirm no-auth mode still works (RBAC helpers return permissive defaults when auth is disabled)

### DIFF
--- a/ui/ui-app/src/app/MainPageWithAuth.tsx
+++ b/ui/ui-app/src/app/MainPageWithAuth.tsx
@@ -22,6 +22,7 @@ import { RolesPage, SettingsPage } from "./pages";
 import { ConfigService, useConfigService } from "@services/useConfigService.ts";
 import { ApplicationAuth, AuthConfig, AuthConfigContext } from "@apicurio/common-ui-components";
 import { AppNavigation, useAppNavigation } from "@services/useAppNavigation.ts";
+import { UserProvider } from "@services/useUserService.ts";
 
 export type MainPageWithAuthProps = object;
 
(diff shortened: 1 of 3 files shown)


yxshwanth:
Rolling back — closing this PR.

Issue #8447, opened 2026-07-02
Operator AppFeaturesITTest fails consistently on main since June 30


paoloantinori:
## Problem

`AppFeaturesITTest` in Operator Local Tests has failed on **every main branch CI run** since June 30 (5 consecutive failures). All 4 test methods (`testAllowDeletesTrue`, `testAllowDeletesDefault`, `testVersionMutabilityEnabledTrue`, `testVersionMutabilityEnabledDefault`) fail with `ConditionTimeoutException` waiting for the deployment to exist.

## Evidence

| Run | Date | PR | Result |
|-----|------|-----|--------|
| 28400600397 | Jun 29 | typescript-eslint update | ❌ Local Tests |
| 28444313846 | Jun 30 | UBI minimal docker update | ❌ Local Tests + app-sql (Docker Hub flake) |
| 28475527810 | Jun 30 | **securityContext hardening (#8204)** | ❌ Local Tests |
| 28513465087 | Jul 1 | operator channel test (#8412) | ❌ Local Tests |
| 28528024816 | Jul 1 | CLI config commands (#8267) | ❌ Local Tests + OLM-v1 |

## Root Cause Hypothesis

The failure correlates with the merge of PR #8204 (`fix(security): add securityContext hardening to operator Deployment manifest`), which added `readOnlyRootFilesystem` and other security constraints to the operator deployment. The `AppFeaturesITTest` creates an `ApicurioRegistry3` CR and waits for the app deployment to become ready — if the hardened security context prevents the app pod from starting (e.g., Quarkus needs to write to `/tmp` or other paths), the awaitility timeout fires.

## Failure Details

```
org.awaitility.core.ConditionTimeoutException: 
Assertion condition defined as a Lambda expression in io.apicurio.registry.operator.it.ITBase 
  at io.apicurio.registry.operator.it.AppFeaturesITTest.testAllowDeletesTrue(AppFeaturesITTest.java:31)
```

Line 31 is `checkDeploymentExists(registry, ResourceFactory.COMPONENT_APP, 1)` — the deployment never reaches ready state.

Cluster diagnostics from the CI logs show the deployment and services exist but pods are likely crash-looping or failing readiness probes.

## Impact

- Every main push that includes operator tests fails
- `Operator Tests / Publish Operator` cascades into failure
- Verification Gate fails on all affected runs
- The `AuthITTest` also times out in some runs (193.5s), suggesting a broader pod startup issue

## Suggested Investigation

[...]

PR #8452, opened 2026-07-02
refactor(operator): convert test lifecycle to @TestInstance(PER_CLASS)

paoloantinori:
## Summary

Convert operator integration test infrastructure from static to instance-based lifecycle using JUnit 5 `@TestInstance(Lifecycle.PER_CLASS)`. This is a **code quality improvement** — not a fix for the `-T 1C` CI failure.

## What Changed

- `ITBase` and `OLMITBase`: add `@TestInstance(PER_CLASS)`, convert 12 static fields to instance fields, convert lifecycle methods to instance methods
- Pure utility methods (`calculateNamespace`, `createK8sClient`, etc.) remain static — they don't access instance state
- 10 subclasses updated: `static @BeforeAll` → instance, method references `ITBase::` → `this::`

## What This Does NOT Fix

**The `-T 1C` CI failure (issue #8447) is caused by resource exhaustion, not shared Java state.** The Makefile runs `-pl controller -am` — only ONE module's tests execute. The `olm-tests` module is not involved. Under `-T 1C`, 8 parallel Maven threads do Quarkus augmentation + compilation while Minikube pods are also running. The CPU/memory pressure causes pods to start slower, exceeding the 120s awaitility timeout.

The correct fix for the CI failure is the `-T1` override in the Makefile (PR #8448, already merged). This refactoring is complementary — it eliminates unnecessary static state as a code quality improvement and a prerequisite for future intra-module test parallelism (if ever needed).

## Value of This Change

1. **Code quality**: static mutable state in test base classes is an anti-pattern. Instance fields with PER_CLASS lifecycle are cleaner and more explicit about ownership.
2. **Future-proofing**: if JUnit 5 parallel test execution within a module is ever enabled, instance fields prevent cross-test contamination.
3. **Documentation**: demonstrates the pattern for other Quarkus operator projects considering test parallelism — and honestly documents what it does and doesn't solve.

## Status

- [x] Compiles successfully (`./mvnw install -f operator/pom.xml -DskipTests`)
- [x] Code review passed (1 blocker found and fixed: `AuthTLSITTest.init()` was still static)
- [ ] Functional testing (requires Minikube — CI will validate)

## Context

See the [Known Issues FAQ in Discussion #8365](https://github.com/Apicurio/apicurio-registry/discussions/8365) for the full analysis, including the corrected root cause (resource exhaustion, not shared memory state).


paoloantinori:
Paolo's note: 
this is the result of AI chasing another thing, and doing all this work. 
I wouldn't call it particularly important if not only for the future-proofing consideration.

I leave you decide the tradeoff between clarity and future proof

paoloantinori:
Addressed Copilot's review: renamed `operatorDeployment` local variable to `operatorDeploymentCell` in `RestrictedNamespaceITTest` to avoid shadowing the inherited enum field (commit 269b49e).

jsenko:
Pre-existing bug (not introduced by this PR): `strimziInstalled` is initialized to `false` but never set to `true` after `applyStrimziResources()` completes. The guard `if (!strimziInstalled)` in `KafkaSqlITTest`, `KafkaSqlTLSITTest`, `KafkaSqlOAuthITTest`, and `KafkaSqlAccessITTest` is therefore always true, and Strimzi resources are re-applied on every test class setup.

Fix would be to set `strimziInstalled = true` at the end of `applyStrimziResources()`. Small performance gain — avoids redundant downloads and resource creation when multiple Kafka test classes run in sequence.

paoloantinori:
@jsenko Good catch — fixed in commit 44378fe. Added `strimziInstalled = true` at the end of `applyStrimziResources()`. Avoids redundant Strimzi resource downloads and creation when multiple Kafka test classes run in sequence.

jsenko (dismissed):
LGTM, I'm writing an issue to parallelize the operator tests, which would give use the most gains.

operator/controller/src/test/java/io/apicurio/registry/operator/it/ITBase.java
```diff
@@ -385,6 +387,7 @@ static void applyStrimziResources() throws IOException {
                 });
             });
         }
+        strimziInstalled = true;
```
EricWittmann: With `PER_CLASS` + instance fields, `strimziInstalled` loses its cross-class semantics. Previously as a `static` field it was shared across all `ITBase` subclasses in the same JVM, so Strimzi was installed once and subsequent Kafka test classes (`KafkaSqlITTest`, `KafkaSqlTLSITTest`, `KafkaSqlOAuthITTest`, `KafkaSqlAccessITTest`) skipped re-installation.

Now each test class instance starts with `strimziInstalled = false`. The `= true` assignment added here is a good fix for the pre-existing bug (the flag was never set to `true` before), but it only affects the current instance — and `@BeforeAll` runs once per instance with `PER_CLASS`, so the flag is never re-checked on that same instance.

Net effect: if multiple Kafka test classes run in the same JVM, each one will redundantly download and apply Strimzi YAML from GitHub. Since `applyStrimziResources()` uses `createOrReplace` this is idempotent and not harmful, just slower.

Consider keeping `strimziInstalled` as the one intentional `static` field (since it genuinely needs cross-class sharing), or adding a comment noting that redundant installation is accepted as idempotent.

paoloantinori:
@EricWittmann Good catch — `strimziInstalled` restored to `static` with a comment explaining the cross-class sharing intent (commit 1ed4832). All other fields remain instance-scoped.

operator/controller/src/test/java/io/apicurio/registry/operator/it/ITBase.java
```diff
@@ -83,18 +85,19 @@ public enum OperatorDeployment {
...
-    protected static boolean cleanup;
+    protected OperatorDeployment operatorDeployment;
+    protected KubernetesClient client;
+    protected PodLogManager podLogManager;
+    protected PortForwardManager portForwardManager;
+    protected IngressManager ingressManager;
+    protected String deploymentTarget;
+    protected String namespace;
+    protected boolean cleanup;
+    // Intentionally static: shared across test class instances so Strimzi is installed once per JVM
```
EricWittmann: Since this PR describes itself as "a prerequisite for future intra-module test parallelism," it's worth noting that this static boolean guard has no synchronization. If parallel execution is ever enabled, concurrent test classes could both read `strimziInstalled == false` and race into `applyStrimziResources()`, causing duplicate Strimzi installations.

The codebase already has a thread-safe pattern for exactly this kind of "install once per JVM" guard — see `SharedDebeziumInfrastructure` in `integration-tests/`, which uses `AtomicBoolean` + `synchronized` methods. Consider using `AtomicBoolean.compareAndSet()` here to future-proof the guard, e.g.:

```java
private static final AtomicBoolean strimziInstalled = new AtomicBoolean(false);

void applyStrimziResources() throws IOException {
    if (!strimziInstalled.compareAndSet(false, true)) {
        return;
    }
    // ... install logic ...
}
```

Not a blocker for this PR since tests run sequentially today, but worth considering given the stated goal.

paoloantinori:
@EricWittmann Applied the AtomicBoolean suggestion (commit e6391ef). Guard is now inside `applyStrimziResources()` using `compareAndSet(false, true)`, following the `SharedDebeziumInfrastructure` pattern. Callers simplified — no longer check the flag before calling.

paoloantinori:
**Pre-review note on scope:**

`MultiNamespaceSmokeITTest.createRegistry(String namespace)` also has a parameter named `namespace` that shadows the inherited field. This is **intentional** — the method is a helper called with different namespace values (`namespace`, `namespace2`, `namespace3`), so the parameter name correctly describes what it receives. Unlike `RestrictedNamespaceITTest` where the shadowed local variable (`operatorDeployment`) masked an enum field of the same name, the `MultiNamespaceSmokeITTest` case is parameterization, not accidental shadowing.

aclfe:
Hello, new here so might be missing some context. Following on from @EricWittmann's point, there's a related question for the operator itself. This PR makes app a per-class instance field, so each test class starts its own operator. But #8495's design goes the other way: one cluster-wide operator shared across all classes, held at JUnit root scope? I want to work on #8495, that's why I'm asking. For app specifically, is #8495 expected to lift it back out of the test instance?

paoloantinori:
@aclfe Welcome! Good question — you're right that these two PRs address related but distinct layers:

**This PR (#8452)** removes the Java-level barrier to parallelism: static fields shared across test classes become instance fields via `@TestInstance(PER_CLASS)`. Without this, even running tests in separate namespaces would hit shared-state races at the JVM level (e.g., `client`, `namespace`, `app` being overwritten by concurrent test classes).

**#8495** is the Kubernetes-level design: a single cluster-wide operator shared via JUnit 5 `ExtensionContext.Store` at ROOT scope, with per-class namespace isolation.

To answer your specific question: yes, #8495 would lift `app` back out of the per-class instance — it moves to the ROOT store as a shared `CloseableResource` managed by the extension, not by individual test classes. This PR is a prerequisite that cleans up the field/lifecycle semantics so #8495 has a solid foundation to build on.

In short: this PR makes fields per-class (safe), #8495 then promotes operator lifecycle to per-suite (efficient).

aclfe:
@paoloantinori Thanks, that clears it up well! I'll start digging into the extension design for #8495 and follow up with questions as they come. Looking forward to seeing this merged

carlesarnal (dismissed):
LGTM — clean PER_CLASS conversion. The `strimziInstalled` change from `boolean` to `static AtomicBoolean` with `compareAndSet` is a nice improvement over the old `if (!strimziInstalled)` pattern — correct for shared-across-JVM semantics and future-proofs for parallel execution. Static-to-instance field conversions and method reference updates (`this::getOperatorDeployment`) are all correct.

jsenko:
The `Operator Tests / Remote Tests (olm-v0)` failure is a transient infrastructure issue — minikube failed to download the `kubelet` binary due to a network connection reset (`INET_RESET_BY_PEER`) on the GitHub Actions runner. Not related to the PR changes.

Re-running the tests is needed after the current workflow execution finishes.

jsenko:
@paoloantinori Needs rebase to include https://github.com/Apicurio/apicurio-registry/pull/8511

paoloantinori:
The Kafka test failures blocking this PR were **not** environmental — root cause found and fixed in 928ea857b.

**Mechanism** (verified in run [29488059140](https://github.com/Apicurio/apicurio-registry/actions/runs/29488059140)): the `strimziInstalled` guard made `applyStrimziResources()` run once per JVM, but the Strimzi operator Deployment is installed into the **per-class namespace**, which `afterAll()` deletes. The log shows it exactly: `Creating Strimzi resource kind Deployment` appears **once** (namespace `test-7b96a04`, KafkaSqlAccessITTest — passes in 333s), that namespace is deleted at 14:43:50, and the three later classes create their Kafka CRs in fresh namespaces with **no operator to reconcile them** — no install lines, each burning the full 600s `KAFKA_BROKER_READY_TIMEOUT` (674/718/673s). That's why failures were deterministic since Jul 9 (7+ consecutive), why #8511's longer timeout and #8652's resource limits couldn't help, and why other PRs' kafka jobs passed the same day (#8660, green in 24m38s — main has no guard).

@EricWittmann the `compareAndSet` pattern is the right tool for a JVM-wide singleton (the SharedDebeziumInfrastructure case) — it just can't work here while the install target lives inside a resource that's recreated per class. A per-instance or per-namespace guard would be trivially always-false (fresh namespace per class), so the fix removes the guard entirely and documents the constraint at the method. This also resolves the original 'declared but never set' observation more completely than re-patching it.

Two things deliberately **not** done here: (1) the latent parallel-execution race on the cluster-scoped ClusterRoleBinding rebinding is pre-existing (shared with main) and only relevant if parallelism is ever enabled; (2) the real optimization — install Strimzi **once per job, cluster-wide watch, dedicated non-deleted namespace**, manifest vendored instead of downloaded per class — is tracked as a follow-up (backlog REG-111) to keep this PR minimal. Verification for this fix = the kafka group itself: all 4 classes should now install Strimzi and pass.

PR #8460, opened 2026-07-04
feat(mcp): add HTTP with OAuth token forwarding

### LINKED ISSUE #8394: MCP server: OAuth-authenticated HTTP transport for remote AI agents

Today the MCP server only talks over stdio (e.g. Claude Desktop spawns it locally). When it needs to reach a secured Registry, it uses a fixed OAuth2 service account (`admin-client`), so every caller shares one identity regardless of who they are in the UI.

We need an optional HTTP transport mode for remote AI agents such as Cursor and Claude via connectors, etc.... In that mode, the agent authenticates with Keycloak the same realm and roles as the Registry UI and the MCP server validates the token and forwards it unchanged to the Registry API. A readonly user cannot delete artifacts through MCP any more than they can in the UI.

### Proposed Solution

Add OAuth-authenticated HTTP transport for remote AI agents while keeping stdio as the default transport.

- **HTTP mode:** The agent logs in with Keycloak; the MCP server checks the token and forwards it to the Registry API, preserving per-user authorization.
- **Stdio mode:** Claude Desktop and other local stdio clients continue to work as today, using client credentials for outbound Registry access.

HTTP OAuth is enabled at runtime when needed; stdio remains the default.

### Additional Context

MCP epic (#6991)

Joel-hanson:
## Summary

This PR adds an **HTTP transport** to the Apicurio Registry MCP server, secured with **inbound OIDC** and **per-caller bearer-token forwarding** to Registry.

Today, MCP only works as a local process (stdio) that authenticates to Registry with a **shared service account**. After this change, you can also run MCP as a long-lived HTTP service: clients log in with their own Keycloak user, and Registry applies **that user’s RBAC** (same as the UI).

Stdio mode is unchanged and remains the default for local desktop clients.

Fixes: #8394

---

### Before (stdio only)

```text
┌─────────────┐   spawn process    ┌──────────────┐   client credentials   ┌──────────┐
│ MCP client  │ ─────────────────► │ MCP server   │ ─────────────────────► │ Registry │
│ (desktop)   │   stdin / stdout   │ (one-shot)   │   admin-client token   │          │
└─────────────┘                    └──────────────┘                        └──────────┘

• One shared service account for all tool calls
• Every user effectively has the same Registry permissions
• Works well for local Claude Desktop / Cursor command configs
```

### After (stdio **or** HTTP)

```text
MODE A — stdio (unchanged)
┌─────────────┐   spawn process    ┌──────────────┐   client credentials   ┌──────────┐
│ MCP client  │ ─────────────────► │ MCP server   │ ─────────────────────► │ Registry │
└─────────────┘                    └──────────────┘                        └──────────┘

MODE B — HTTP + token forwarding (new)
┌─────────────┐   HTTPS + Bearer   ┌──────────────┐   same Bearer token    ┌──────────┐
│ MCP client  │ ─────────────────► │ MCP server   │ ─────────────────────► │ Registry │
│ (remote /   │   POST /mcp        │ (long-lived) │   per-user RBAC        │          │
│  desktop)   │                    └──────┬───────┘                        └──────────┘
└──────┬──────┘                           │
       │                                  │ validates token
       │         OAuth login              ▼
       └──────────────────────────► ┌──────────┐
                                    │ Keycloak │
                                    └──────────┘

• Caller’s own Keycloak token is forwarded to Registry
• sr-readonly cannot create groups; sr-admin can — same as the UI
• Suitable for MCP Inspector, remote agents, shared deployments
```

### Runtime safety gate

HTTP transport is **compiled into** the MCP image at build time, but `/mcp` is **not exposed** unless you opt in at runtime:

```text
quarkus.mcp.server.http.enabled=true     ← build time (always in published images)
apicurio.mcp.http.enabled=false          ← runtime default

                    ┌─ apicurio.mcp.http.enabled=false ──► /mcp returns 404
Request to /mcp ───┤
                    └─ apicurio.mcp.http.enabled=true
                       + Quarkus OIDC configured ─────────► /mcp requires Bearer token
```

Enforced by `McpHttpGateFilter` (see `McpHttpDisabledByDefaultTest`).

---

### Journey 1 — Local desktop client (stdio) — **no change required**

1. Point your MCP client at the MCP Docker image / JAR (as today).
2. If Registry is secured, set `APICURIO_MCP_AUTH_*` client credentials.
3. Keep HTTP mode **off** (`APICURIO_MCP_HTTP_ENABLED=false`, `QUARKUS_OIDC_TENANT_ENABLED=false`).

Example: [`examples/mcp-keycloak/stdio.example.json`](https://github.com/Joel-hanson/apicurio-registry/blob/efb855974bdf1ed98dff1a716b8951476c9e8170/examples/mcp-keycloak/stdio.example.json)

### Journey 2 — HTTP MCP with OAuth (new)

1. Run a secured Registry + Keycloak (or use the example stack below).
2. Start the MCP server as an HTTP service with HTTP mode + OIDC enabled.
3. Configure your MCP client with `url: http://<host>:8082/mcp` and OAuth client `apicurio-mcp`.
4. Client completes OAuth login; MCP validates the token and forwards it to Registry.
5. Tool calls succeed or fail based on **that user’s** Registry roles.

Example: [`examples/mcp-keycloak/http.example.json`](https://github.com/Joel-hanson/apicurio-registry/blob/efb855974bdf1ed98dff1a716b8951476c9e8170/examples/mcp-keycloak/http.example.json)

---

### Stdio mode (MCP → Registry)

| Setting | Env var | Example |
| --- | --- | --- |
| Registry URL | `REGISTRY_URL` | `http://localhost:8081` |
| Enable OAuth2 | `APICURIO_MCP_AUTH_ENABLED` | `true` |
| Token endpoint | `APICURIO_MCP_AUTH_TOKEN_ENDPOINT` | `http://localhost:8080/realms/registry/protocol/openid-connect/token` |
| Client ID | `APICURIO_MCP_AUTH_CLIENT_ID` | `admin-client` |
| Client secret | `APICURIO_MCP_AUTH_CLIENT_SECRET` | `test1` |
| Disable inbound OIDC | `QUARKUS_OIDC_TENANT_ENABLED` | `false` |
| Disable HTTP mode | `APICURIO_MCP_HTTP_ENABLED` | `false` |

### HTTP mode (remote / OAuth clients)

**Apicurio settings:**

| Setting | Env var | Example |
| --- | --- | --- |
| Enable HTTP mode | `APICURIO_MCP_HTTP_ENABLED` | `true` |
| Forward caller token | `APICURIO_MCP_HTTP_FORWARD_TOKEN` | `true` |
| Registry URL | `REGISTRY_URL` | `http://apicurio-registry:8080` |

**Quarkus companion settings (required when HTTP mode is on):**

| Setting | Env var | Example |
| --- | --- | --- |
| Enable HTTP transport | `QUARKUS_MCP_SERVER_HTTP_ENABLED` | `true` |
| Disable stdio | `QUARKUS_MCP_SERVER_STDIO_ENABLED` | `false` |
| Enable OIDC | `QUARKUS_OIDC_TENANT_ENABLED` | `true` |
| Application type | `QUARKUS_OIDC_APPLICATION_TYPE` | `service` |
| Auth server | `QUARKUS_OIDC_AUTH_SERVER_URL` | `http://keycloak:8080/realms/registry` |
| Secure `/mcp` | `QUARKUS_HTTP_AUTH_PERMISSION_AUTHENTICATED_PATHS` | `/mcp` |
| Require auth | `QUARKUS_HTTP_AUTH_PERMISSION_AUTHENTICATED_POLICY` | `authenticated` |

Optional (OAuth discovery / CORS for Inspector-style clients): see `examples/mcp-keycloak/docker-compose.yml`.

Full reference: [MCP server integration guide](docs/modules/ROOT/pages/getting-started/assembly-mcp-server-integration.adoc)

---

## What’s in the code

| Area | Change |
| --- | --- |
| `RegistryClientResolver` | Per-request client: forward inbound Bearer token, or fall back to client credentials / anonymous |
| `McpHttpGateFilter` | `/mcp` returns **404** unless `apicurio.mcp.http.enabled=true` |
| `McpHttpAuthValidator` | Fail-fast at startup if HTTP mode is on without OIDC / transport |
| Java SDK | `RegistryClientOptions.bearerToken()`, `AuthType.BEARER`, JDK adapter support |
| `Utils.java` | Clearer `ProblemDetails` / 401–403 messages (roles matter now that tokens are per-user) |
| `examples/mcp-keycloak` | Local Keycloak + Registry + MCP HTTP stack |
| Docs | HTTP/OAuth section in the integration guide |

### Why `Utils.java` error formatting is in this PR

With token forwarding, tool calls fail based on the **caller’s** Keycloak role. Returning a clear 401/403 (with role hints) is part of making HTTP mode usable for agents and users.

[Quarkus OIDC](https://quarkus.io/guides/security-oidc-bearer-token-authentication)

### DIFF
--- a/mcp/src/main/java/io/apicurio/registry/mcp/McpHttpAuthValidator.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/McpHttpAuthValidator.java
@@ -1,0 +1,158 @@
+package io.apicurio.registry.mcp;
+
+import io.quarkus.runtime.StartupEvent;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.event.Observes;
+import jakarta.inject.Inject;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.MalformedURLException;
+import java.net.Socket;
+import java.net.URI;
+import java.net.URL;
+
+/**
+ * Validates HTTP transport and OIDC configuration at startup.
+ */
+@ApplicationScoped
+public class McpHttpAuthValidator {
+
+    private static final Logger log = LoggerFactory.getLogger(McpHttpAuthValidator.class);
+
+    private static final int REGISTRY_CONNECT_TIMEOUT_MS = 3_000;
+
+    @Inject
+    McpConfig config;
+
+    @ConfigProperty(name = "quarkus.oidc.tenant-enabled", defaultValue = "false")
+    boolean oidcTenantEnabled;
+
+    @ConfigProperty(name = "quarkus.mcp.server.http.enabled", defaultValue = "false")
+    boolean mcpHttpTransportEnabled;
+
+    @ConfigProperty(name = "registry.url", defaultValue = "localhost:8080")
+    String registryUrl;
+
+    /**
+     * Pluggable TCP probe so unit tests can avoid real network I/O.
+     */
+    ConnectivityProbe connectivityProbe = McpHttpAuthValidator::tcpConnect;
+
+    void onStart(@Observes StartupEvent event) {
+        if (!config.http().enabled()) {
+            return;
+        }
+
+        validateHttpAuthConfig();
+        validateRegistryUrl();
+
+        if (config.http().forwardToken() && config.auth().enabled()) {
+            log.warn("Both apicurio.mcp.http.forward-token and apicurio.mcp.auth.enabled are set. "
+                    + "HTTP requests will forward the caller's bearer token; client credentials are used "
+                    + "only for stdio transport.");
+        }
+
+        log.info("MCP HTTP transport enabled with inbound OIDC authentication"
+                + (config.http().forwardToken() ? " and bearer token forwarding to Registry" : ""));
+    }
+
+    /**
+     * Validates that HTTP mode has the required transport and OIDC settings.
+     *
+     * @throws IllegalStateException if required configuration is missing
+     */
+    void validateHttpAuthConfig() {
+        if (!mcpHttpTransportEnabled) {
+            throw new IllegalStateException(
+                    "apicurio.mcp.http.enabled is true but quarkus.mcp.server.http.enabled is false. "
+                            + "HTTP transport must be enabled at build time (quarkus.mcp.server.http.enabled=true "
+                            + "in application.properties) and the MCP server must be rebuilt.");
+        }
+
+        if (!oidcTenantEnabled) {
+            throw new IllegalStateException(
+                    "apicurio.mcp.http.enabled is true but quarkus.oidc.tenant-enabled is false. "
+                            + "Configure Quarkus OIDC (QUARKUS_OIDC_TENANT_ENABLED, QUARKUS_OIDC_AUTH_SERVER_URL) "
+                            + "to secure the MCP HTTP endpoint.");
+        }
+    }
+
+    /**
+     * Validates {@code registry.url} format and, when the resolver skips its authenticated
+     * {@code system.info()} probe (HTTP mode without client-credentials auth), checks that the
+     * host/port is reachable so misconfiguration fails at startup instead of the first tool call.
+     */
+    void validateRegistryUrl() {
+        HostPort hostPort = parseRegistryHostPort(registryUrl);
+        if (!config.auth().enabled()) {
+            try {
+                connectivityProbe.connect(hostPort.host(), hostPort.port());
+                log.info("Verified registry.url is reachable at {}:{}", hostPort.host(), hostPort.port());
+            } catch (IOException e) {
+                throw new IllegalStateException(
+                        "Cannot reach Apicurio Registry at registry.url=" + registryUrl
+                                + " (" + hostPort.host() + ":" + hostPort.port() + "). "
+                                + "Check REGISTRY_URL / registry.url and that Registry is running. "
+                                + "Details: " + e.getMessage(),
+                        e);
+            }
+        }
+    }
+
+    /**
+     * Parses {@code registry.url} values such as {@code localhost:8080} or
+     * {@code http://localhost:8080/apis/registry/v3} into a host/port pair using
+     * {@link URI} / {@link URL} (including {@link URL#getDefaultPort()}).
+     */
+    static HostPort parseRegistryHostPort(String rawRegistryUrl) {
+        if (rawRegistryUrl == null || rawRegistryUrl.isBlank()) {
+            throw new IllegalStateException(
+                    "registry.url must be configured when apicurio.mcp.http.enabled=true");
+        }
+
+        // registry.url often omits the scheme (e.g. "localhost:8080"); URI/URL require one.
+        String value = rawRegistryUrl.trim();
+        if (!value.contains("://")) {
+            log.info("registry.url has no scheme; assuming http:// for host/port parsing: {}", rawRegistryUrl);
+            value = "http://" + value;
+        }
+
+        try {
+            URL url = URI.create(value).toURL();
+            String host = url.getHost();
+            if (host == null || host.isBlank()) {
+                throw new IllegalStateException(
+                        "registry.url is invalid (could not determine host): " + rawRegistryUrl);
+            }
+            int port = url.getPort() >= 0 ? url.getPort() : url.getDefaultPort();
+            if (port < 0) {
+                throw new IllegalStateException(
+                        "registry.url is invalid (could not determine port): " + rawRegistryUrl);
+            }
+            log.debug("Parsed registry.url {} as {}:{}", rawRegistryUrl, host, port);
+            return new HostPort(host, port);
+        } catch (IllegalArgumentException | MalformedURLException e) {
+            throw new IllegalStateException(
+                    "registry.url is not a valid URL: " + rawRegistryUrl + ". Details: " + e.getMessage(),
+                    e);
+        }
+    }
+
+    private static void tcpConnect(String host, int port) throws IOException {
+        try (Socket socket = new Socket()) {
+            socket.connect(new InetSocketAddress(host, port), REGISTRY_CONNECT_TIMEOUT_MS);
+        }
+    }
+
+    @FunctionalInterface
+    interface ConnectivityProbe {
+        void connect(String host, int port) throws IOException;
+    }
+
+    record HostPort(String host, int port) {
+    }
+}
--- a/mcp/src/main/java/io/apicurio/registry/mcp/McpHttpGateFilter.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/McpHttpGateFilter.java
@@ -1,0 +1,51 @@
+package io.apicurio.registry.mcp;
+
+import io.quarkus.runtime.StartupEvent;
+import io.vertx.ext.web.Router;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.event.Observes;
+import jakarta.inject.Inject;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import java.util.regex.Pattern;
+
+/**
+ * Blocks MCP HTTP endpoints unless {@code apicurio.mcp.http.enabled} is true.
+ * <p>
+ * The Quarkus MCP HTTP transport is compiled in at build time, but HTTP mode is only
+ * intended to be reachable when explicitly enabled at runtime with OIDC configured.
+ */
+@ApplicationScoped
+public class McpHttpGateFilter {
+
+    @Inject
+    McpConfig config;
+
+    @Inject
+    Router router;
+
+    @ConfigProperty(name = "quarkus.mcp.server.http.root-path", defaultValue = "/mcp")
+    String mcpRootPath;
+
+    void register(@Observes StartupEvent event) {
+        String escapedRootPath = Pattern.quote(mcpRootPath);
+        router.routeWithRegex(escapedRootPath + "/?.*")
+                .order(Integer.MIN_VALUE)
+                .handler(ctx -> blockUnlessEnabled(ctx));
+        router.route("/.well-known/oauth-protected-resource")
+                .order(Integer.MIN_VALUE)
+                .handler(ctx -> blockUnlessEnabled(ctx));
+    }
+
+    void blockUnlessEnabled(io.vertx.ext.web.RoutingContext ctx) {
+        if (!config.http().enabled()) {
+            ctx.response()
+                    .setStatusCode(503)
+                    .putHeader("Content-Type", "text/plain; charset=UTF-8")
+                    .end("MCP HTTP transport is disabled. Set apicurio.mcp.http.enabled=true "
+                            + "with OIDC configured to enable it.");
+        } else {
+            ctx.next();
+        }
+    }
+}
--- a/mcp/src/main/java/io/apicurio/registry/mcp/RegistryClientResolver.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/RegistryClientResolver.java
@@ -1,0 +1,394 @@
+package io.apicurio.registry.mcp;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.apicurio.registry.client.RegistryClientFactory;
+import io.apicurio.registry.client.common.HttpAdapterType;
+import io.apicurio.registry.client.common.RegistryClientOptions;
+import io.apicurio.registry.rest.client.RegistryClient;
+import io.quarkus.oidc.AccessTokenCredential;
+import io.quarkus.security.identity.SecurityIdentity;
+import jakarta.annotation.PostConstruct;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.context.control.ActivateRequestContext;
+import jakarta.enterprise.inject.Instance;
+import jakarta.inject.Inject;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.jwt.JsonWebToken;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.util.Base64;
+import java.util.Locale;
+
+/**
+ * Resolves the Registry REST client for the current request.
+ * <p>
+ * In HTTP mode with token forwarding, uses the inbound caller's bearer token.
+ * Otherwise uses a singleton client configured with OAuth2 client credentials or anonymous access.
+ */
+@ApplicationScoped
+public class RegistryClientResolver {
+
+    private static final Logger log = LoggerFactory.getLogger(RegistryClientResolver.class);
+
+    private static final ObjectMapper JWT_PAYLOAD_MAPPER = new ObjectMapper();
+
+    /**
+     * Seconds after {@code exp} during which the token is still accepted.
+     * Covers small clock differences between this server and the OIDC provider.
+     */
+    static final long BEARER_TOKEN_EXPIRATION_SKEW_SECONDS = 30;
+
+    @ConfigProperty(name = "registry.url", defaultValue = "localhost:8080")
+    String rawBaseUrl;
+
+    @Inject
+    McpConfig config;
+
+    @Inject
+    Instance<SecurityIdentity> securityIdentity;
+
+    @Inject
+    Instance<JsonWebToken> jwt;
+
+    private RegistryClient fallbackClient;
+
+    /**
+     * Cached base options (URL, HTTP adapter, retry, TLS) built once at startup.
+     * Per-request callers must {@link #copyTransportOptions()} before applying a bearer
+     * token so concurrent requests do not mutate this shared instance.
+     */
+    private RegistryClientOptions transportOptions;
+
+    @PostConstruct
+    void init() {
+        transportOptions = buildTransportOptions();
+        if (needsFallbackClient()) {
+            fallbackClient = createClient(buildFallbackOptions());
+            if (!config.http().enabled() || config.auth().enabled()) {
+                var info = fallbackClient.system().info().get();
+                log.info("Successfully connected to Apicurio Registry version {} at {}", info.getVersion(),
+                        rawBaseUrl);
+            }
+        } else {
+            log.info("Registry client will authenticate using inbound bearer token forwarding");
+        }
+    }
+
+    /**
+     * Registry base URL from {@code registry.url} (same value used to build clients).
+     */
+    public String registryBaseUrl() {
+        return rawBaseUrl;
+    }
+
+    /**
+     * Returns a Registry client for the current context.
+     */
+    @ActivateRequestContext
+    public RegistryClient getClient() {
+        String bearerToken = resolveInboundBearerToken();
+        if (bearerToken != null) {
+            // Reuse cached transport settings (URL/TLS/adapter/retry); only the bearer token
+            // is applied per request on a fresh options instance.
+            var options = copyTransportOptions().bearerToken(bearerToken);
+            return RegistryClientFactory.create(options);
+        }
+        if (fallbackClient == null) {
+            boolean authenticated = securityIdentity.isResolvable()
+                    && securityIdentity.get() != null
+                    && !securityIdentity.get().isAnonymous();
+            if (authenticated) {
+                throw new IllegalStateException(
+                        "No Registry client available: HTTP token forwarding is enabled but no bearer access "
+                                + "token was found on the authenticated request. Ensure the MCP client sends "
+                                + "'Authorization: Bearer <access_token>' on every /mcp request (complete "
+                                + "OAuth login so the access token is sent).");
+            }
+            throw new IllegalStateException(
+                    "No Registry client available: HTTP token forwarding is enabled but no bearer token "
+                            + "was provided, and no fallback client credentials are configured");
+        }
+        return fallbackClient;
+    }
+
+    boolean hasFallbackClient() {
+        return fallbackClient != null;
+    }
+
+    boolean needsFallbackClient() {
+        if (!config.http().enabled()) {
+            return true;
+        }
+        return config.auth().enabled() || !config.http().forwardToken();
+    }
+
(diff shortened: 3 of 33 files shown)


Joel-hanson:
@jsenko I’ve pushed all the changes that I believe are needed to add this capability. Please let me know if the changes make sense or if any additional changes are required or require a different way to implementation.

carlesarnal (changes requested):
This PR adds HTTP transport with OAuth token forwarding. A few things to address:

**Must fix:**
- The javadoc in `RegistryClientOptions.bearerToken()` references paths from a different project (`@qlj/agentic/...`). Should be standard `@param` / `@return` tags.
- No unit tests for `RegistryClientResolver` — token extraction from `SecurityIdentity` and `JsonWebToken`, fallback client creation, and error cases when token forwarding is enabled but no token is present all need coverage.
- No tests for `McpHttpGateFilter` (blocked when disabled, accessible when enabled) or `McpHttpAuthValidator` (startup validation with various config combos).

**Should fix:**
- `McpHttpGateFilter.blockUnlessEnabled()` returns 404 when HTTP is disabled. That is misleading — 503 with a body explaining the feature is disabled would be clearer.
- `Utils.formatRegistryApiError()` includes specific Registry role names in 401/403 responses. That leaks authorization model details to unauthenticated users. Consider making that conditional.
- The bearer token gets forwarded without checking expiration. Registry will reject it anyway, but a proactive check would give a better error message.
- CORS config in the docker-compose example (`QUARKUS_HTTP_CORS_ORIGINS=http://localhost:6274`) needs a production warning in the docs.

The HTTP transport structure itself looks solid — proper gate filtering, validation at startup, clean separation. Just needs the test coverage and the issues above sorted out.

Joel-hanson:
I made a set of small, improvements to the MCP HTTP integration. 

- When MCP HTTP mode is disabled, the /mcp endpoint now returns 503 Service Unavailable with a short message explaining that the feature is disabled and how to enable it (instead of a 404).
- Registry error formatting no longer always includes explicit sr-* role names in auth failures. Those role hints are now only added when the caller is already authenticated, to avoid leaking authorization model details to unauthenticated users.
- Added a clear production warning around CORS settings like QUARKUS_HTTP_CORS_ORIGINS=http://localhost:6274 (fine for local dev/MCP Inspector, not for production).
- When forwarding the inbound bearer token to Registry, we now detect expired tokens early and return a clear “token expired, reauthenticate” error instead of letting it fail later with a generic 401 from Registry.

`assertBearerTokenNotExpired` looks for exp in this order:

1. Prefer Quarkus’s already-parsed JWT: `JsonWebToken.getExpirationTime()`
2. If that isn’t available, fall back to `parseJwtExpirationEpochSeconds(...) `— Base64-decode the JWT payload and read exp
3. If there’s no exp (e.g. opaque token), it skips the check and forwards as before
If now >= exp, it throws:
> Bearer access token has expired. Re-authenticate and retry the request with a fresh access token.

instead of sending a dead token to Registry and getting a vague 401 back.

jsenko:

mcp/src/test/java/io/apicurio/registry/mcp/McpHttpTokenForwardingTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+        McpHttpTestHelper.callTool(token, sessionId, 2, "create_group",
+                """
+                        {
+                          "groupId": "%s",
+                          "description": "should fail"
+                        }
+                        """.formatted(groupId))
+                .statusCode(200)
+                .body(not(containsString("\"groupId\":\"" + groupId + "\"")));
```
paoloantinori: This negative assertion is quite permissive — it passes for any response that doesn't happen to contain the literal `"groupId":"<id>"`, including error bodies, empty bodies, or unrelated exceptions. Given the goal is to prove the readonly token is *rejected by Registry RBAC*, asserting the 403 path is actually hit (the `Utils.formatRegistryApiError` `HTTP 403 Forbidden ... role` message) would make this test robust against regressions where a readonly token could create groups.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryClientResolver.java
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * Returns a Registry client for the current context.
+     */
+    @ActivateRequestContext
+    public RegistryClient getClient() {
+        String bearerToken = resolveInboundBearerToken();
+        if (bearerToken != null) {
+            var options = buildBaseOptions().bearerToken(bearerToken);
+            return RegistryClientFactory.create(options);
```
paoloantinori, line 98: Constructing a fresh `RegistryClient` per request is correct for token forwarding, but note this re-runs the full TLS/adapter config (`buildBaseOptions()` → `configureTls()` / `configureClientCredentials()`) every time. For a high-traffic HTTP deployment, a cached base `RegistryClientOptions` (TLS configured once) with only the bearer token applied per request would reduce per-request cost. Not blocking — worth a profile pass if this becomes a hot path.

mcp/src/main/java/io/apicurio/registry/mcp/McpHttpAuthValidator.java
```diff
@@ -0,0 +1,65 @@
...
+        log.info("MCP HTTP transport enabled with inbound OIDC authentication"
+                + (config.http().forwardToken() ? " and bearer token forwarding to Registry" : ""));
+    }
+
+    /**
+     * Validates that HTTP mode has the required transport and OIDC settings.
+     *
+     * @throws IllegalStateException if required configuration is missing
+     */
+    void validateHttpAuthConfig() {
```
paoloantinori, line 68: Good fail-fast design here. One gap worth considering: this validates the MCP/OIDC side but not `registry.url` reachability. In token-forwarding-only mode, `RegistryClientResolver.init()` skips its connection probe (line 57), so a misconfigured `registry.url` only surfaces at the first tool call. A lightweight `registry.url` sanity check here (or documenting the lazy-surface behavior) would round out the startup validation story.

paoloantinori (commented):
Solid addition — the HTTP transport + bearer-token forwarding design is the right shape: stdio unchanged, HTTP gated by `apicurio.mcp.http.enabled`, inbound OIDC validated at startup (`McpHttpAuthValidator`), and the `/mcp` route hard-blocked with 503 when the gate is off (`McpHttpGateFilter`, including `/.well-known/oauth-protected-resource`). A few observations, mostly on test strength and a per-request cost worth noting. (Note: the diff was truncated at 4000 lines; comments below are based on the production files I pulled directly — `McpHttpAuthValidator`, `RegistryClientResolver`, `McpHttpGateFilter`, `Utils`, `McpHttpTokenForwardingTest`.)

- **`testReadonlyTokenCannotCreateGroup` has a weak negative assertion.** The test asserts `not(containsString("\"groupId\":\"" + groupId + "\""))` (line 89). This passes for almost any outcome — an HTTP error, an empty body, a generic exception string, or a successful-but-different response. Since the PR's own `Utils.formatRegistryApiError` is designed to surface `HTTP 403 Forbidden ... may be missing the required Registry role`, the test would be far more robust asserting that the RBAC denial path is actually hit (e.g. `body(containsString("403"))` or the role hint), which is what proves the forwarded readonly token is rejected *by Registry* rather than silently failing for another reason. As written, a future regression where readonly tokens could create groups would likely still pass this assertion.

- **`RegistryClientResolver.getClient()` constructs a fresh `RegistryClient` on every HTTP request** in token-forwarding mode (lines 72-76: `buildBaseOptions()` + `bearerToken(...)` + `RegistryClientFactory.create(...)`). That's semantically correct — the token differs per caller — but each construction re-runs TLS/keystore parsing and adapter setup. For a long-lived HTTP MCP service serving many concurrent clients this is per-request overhead worth measuring; if it shows up in profiling, a cached `RegistryClientOptions` base (TLS configured once) with only the bearer token swapped per request would avoid re-doing the TLS work.

- **`registry.url` default of `localhost:8080`** is unchanged from the pre-PR `RegistryService` (not a regression this PR introduces), but in HTTP-forwarding mode a misconfigured `registry.url` now surfaces lazily at the first tool call rather than at startup, because `init()` skips the connection probe when only token forwarding is active (line 57: `if (!config.http().enabled() || config.auth().enabled())`). The startup validation in `McpHttpAuthValidator` is great for the MCP/OIDC side; consider whether a lightweight reachability check for `registry.url` belongs there too, or document that misconfiguration surfaces at first request.

- **Startup validation is well done** — `McpHttpAuthValidator.validateHttpAuthConfig()` fails fast with actionable messages if `apicurio.mcp.http.enabled=true` but Quarkus MCP HTTP transport or OIDC tenant are not configured, which prevents the footgun of exposing `/mcp` without auth.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryClientResolver.java
Joel-hanson: The changes I had just added works in a way that on each forwarded request we create a fresh RegistryClientOptions from that cached transport config and apply only the bearer token before RegistryClientFactory.create(), so we no longer re-run configureTls() / client-credentials setup from config on every call. happy to make changes if this is not a right or not what you intended.

mcp/src/test/java/io/apicurio/registry/mcp/McpHttpTokenForwardingTest.java
Joel-hanson: agreed that assertion was too weak. Updated testReadonlyTokenCannotCreateGroup to assert HTTP 403 Forbidden

mcp/src/main/java/io/apicurio/registry/mcp/McpHttpAuthValidator.java
Joel-hanson: Addressed, startup now parses registry.url and does a lightweight TCP reachability check so a bad REGISTRY_URL fails fast.

jsenko (commented):
## Code Review Summary

Good implementation overall. The HTTP transport + bearer-token forwarding design is well-structured: stdio mode unchanged, HTTP gated by config, security-conscious token handling (expired token rejection, role hints only for authenticated users), and good test coverage for positive/negative auth cases.

carlesarnal's and paoloantinori's feedback has been properly addressed (cached transport config, 403 assertion, registry URL reachability).

### Findings

**1. Token expiration check lacks clock skew tolerance** (should-fix)
`RegistryClientResolver.java:155-156` — `assertBearerTokenNotExpired()` compares `Instant.now().getEpochSecond() >= expirationEpochSeconds` with zero tolerance. If the MCP server clock is slightly ahead of the OIDC provider, valid tokens get rejected. Consider adding ~30s tolerance:
```java
if (expirationEpochSeconds != null
        && Instant.now().getEpochSecond() >= (expirationEpochSeconds + 30)) {
```
Or document that expired token rejection is best-effort.

**2. Error message assumes Keycloak as OIDC provider** (should-fix)
`Utils.java:149` — The 403 hint says "Your Keycloak user may be missing the required Registry role". Since Registry supports any OIDC provider, consider changing to "Your user may be missing the required Registry role" for provider neutrality.

Joel-hanson:
Hi, @jsenko pushed updates for the remaining review notes. Do the changes look good, or is there anything else that still needs to be addressed?

carlesarnal:
## Re-review: All Feedback Addressed

I went through the latest push and checked each finding from @jsenko, @carlesarnal, and @paoloantinori against the current code. Everything looks well addressed.

**@jsenko findings (Jul 28):**

1. Clock skew tolerance on token expiration — Fixed. BEARER_TOKEN_EXPIRATION_SKEW_SECONDS = 30 constant added in RegistryClientResolver.java, used in assertBearerTokenNotExpired(). Test resolveInboundBearerTokenAllowsRecentlyExpiredTokenWithinClockSkew explicitly covers this.

2. Error message assumes Keycloak as OIDC provider — Fixed. Utils.java now says 'Your user may be missing the required Registry role' (provider-neutral).

**@carlesarnal findings (Jul 10) — all addressed:**

- Javadoc cleaned up with standard @param/@return tags
- RegistryClientResolverTest (583 lines): covers token extraction from SecurityIdentity and JsonWebToken, fallback client selection, expiration rejection with clock skew tolerance, and transport options immutability
- McpHttpGateFilterTest (215 lines): verifies 503 when disabled, pass-through when enabled
- McpHttpAuthValidatorTest (273 lines): startup validation with various config combos, URL parsing, reachability probing
- Gate filter now returns 503 (not 404) with a message explaining how to enable HTTP mode
- Role name hints only shown to authenticated callers via includeAuthorizationHints()

**@paoloantinori findings (Jul 18) — all addressed:**

- testReadonlyTokenCannotCreateGroup now asserts containsString('HTTP 403 Forbidden') and containsString('may be missing the required Registry role') — much stronger than the previous not-contains check
- Cached transportOptions built once at startup; per-request callers get a fresh copy with only the bearer token applied (verified by getClientAppliesBearerTokenWithoutMutatingCachedTransportOptions test)
- McpHttpAuthValidator.validateRegistryUrl() does a TCP reachability check at startup when client-credentials auth is not configured

**Overall quality:**

- All 13 commits have DCO sign-off
- Conventional commit messages throughout
- No star imports, Locale.ROOT used on case conversions
- K&R brace style consistent
- CI all green (SonarCloud, FOSSA, Security)
- Good test coverage across all new components with both positive and negative cases

No new issues found. This looks ready for maintainer approval.

paoloantinori:
Thanks @carlesarnal — I re-read the current head against my earlier findings (HTTP transport gated by `apicurio.mcp.http.enabled`, inbound OIDC validated at startup, token expiry check with the 30s skew tolerance, fallback-client selection, and the error-formatting changes that no longer leak role names by default). Each is addressed cleanly, and the new test coverage (`RegistryClientResolverTest`, `McpHttpAuthValidatorTest`, `McpHttpTokenForwardingTest`, `UtilsErrorFormattingTest`) pins the behaviours well. Agreed that the HTTP transport + bearer-token forwarding design is sound.

One heads-up that isn't a code-review point: the branch currently reports a `dirty` mergeable state (merge conflicts), and @jsenko noted the `License Compliance` (FOSSA) check is failing because the branch predates the FOSSA config fix (#8636). A rebase onto current `main` should clear both. Flagging it only so it doesn't block the merge once the code side is accepted.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryClientResolver.java
```diff
@@ -0,0 +1,394 @@
...
+                    && securityIdentity.get() != null
+                    && !securityIdentity.get().isAnonymous();
+            if (authenticated) {
+                throw new IllegalStateException(
+                        "No Registry client available: HTTP token forwarding is enabled but no bearer access "
+                                + "token was found on the authenticated request. Ensure the MCP client sends "
+                                + "'Authorization: Bearer <access_token>' on every /mcp request (complete "
+                                + "OAuth login so the access token is sent).");
+            }
+            throw new IllegalStateException(
```
paoloantinori, line 113: Privilege-escalation footgun. When `apicurio.mcp.http.forward-token=true` and client-credentials auth is also enabled, `getClient()` falls through to this `return fallbackClient;` for any authenticated request that yields no forwardable bearer token (for example an identity populated by a non-OIDC mechanism, or an OIDC setup that does not expose the raw token via `JsonWebToken`/`AccessTokenCredential`). The caller then runs Registry calls as the configured service account, which is typically `sr-admin`, bypassing per-user RBAC. The startup warning in `McpHttpAuthValidator` even claims client credentials are "used only for stdio transport," but this code makes them reachable from HTTP requests. Recommend throwing here when HTTP forwarding is on (reject tokenless authenticated requests), or making the validator fail-fast instead of merely warning when both `forward-token` and `auth.enabled` are set.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryClientResolver.java
```diff
@@ -0,0 +1,394 @@
...
+
+    /**
+     * Returns a Registry client for the current context.
+     */
+    @ActivateRequestContext
+    public RegistryClient getClient() {
+        String bearerToken = resolveInboundBearerToken();
+        if (bearerToken != null) {
+            // Reuse cached transport settings (URL/TLS/adapter/retry); only the bearer token
+            // is applied per request on a fresh options instance.
```
paoloantinori, line 96: Every forwarded tool call builds a brand new `RegistryClient` through `RegistryClientFactory.create(options)`, which constructs a fresh JDK `HttpClient` (new connection pool and executor) per request. For the long-lived, shared, multi-user HTTP service this PR targets, that means no connection reuse across calls and extra short-lived thread pools under concurrent load. Consider sharing a single `HttpClient`/transport and creating only a lightweight per-token adapter, or pooling clients, so the per-request cost is the token swap rather than a full client rebuild.

mcp/src/main/java/io/apicurio/registry/mcp/McpHttpGateFilter.java
```diff
@@ -0,0 +1,51 @@
...
+                .order(Integer.MIN_VALUE)
+                .handler(ctx -> blockUnlessEnabled(ctx));
+        router.route("/.well-known/oauth-protected-resource")
+                .order(Integer.MIN_VALUE)
+                .handler(ctx -> blockUnlessEnabled(ctx));
+    }
+
+    void blockUnlessEnabled(io.vertx.ext.web.RoutingContext ctx) {
+        if (!config.http().enabled()) {
+            ctx.response()
```
paoloantinori, line 42: The disabled-endpoint gate returns 503, but the PR safety-gate description says `/mcp returns 404` when HTTP mode is off. 503 is a transient status that MCP and HTTP clients may retry with backoff, which is the wrong behavior for an endpoint that is deliberately and permanently disabled. Prefer 404 here (matching the stated security intent and avoiding retry storms), and align the `application.properties` comment, which currently also says 503.

paoloantinori (commented):
Overall a solid, well-tested change: stdio behavior is preserved, the build-time/runtime gate split is sound, and there is good coverage for the disabled-by-default path and per-user RBAC forwarding. The main concerns are a privilege-escalation footgun in the Registry client fallback, per-request HttpClient churn in forwarding mode, and a status-code mismatch on the disabled-endpoint gate.

- `RegistryClientResolver.java:111` returns the configured client-credentials (typically `sr-admin`) fallback for HTTP requests when no bearer token can be extracted, contradicting the startup warning that says those credentials are "used only for stdio transport."
- `RegistryClientResolver.java:96` builds a fresh `RegistryClient` (new JDK `HttpClient`, connection pool, and executor) on every forwarded tool call, so the long-lived HTTP service gets no connection reuse and extra short-lived thread pools under load.
- `McpHttpGateFilter.java:42` returns 503 for the disabled endpoint while the PR safety-gate description says 404; 503 invites client retries against a permanently disabled route.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryClientResolver.java
Joel-hanson: Good catch.

Fixed and added a unit test covering the case where a fallback is configured.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryClientResolver.java
Joel-hanson: agreed a shared HttpClient would be nicer for connection reuse, but that needs a lot mote changes and feels bigger than this PR. We already cache the transport options and only apply the bearer token per request, which was the earlier ask. happy to do the HttpClient sharing as a follow-up if you want.

mcp/src/main/java/io/apicurio/registry/mcp/McpHttpGateFilter.java
Joel-hanson: @carlesarnal asked for 503 with a short explanation instead of 404 so it doesn't look like a missing route, and that was what we went with in the earlier review. happy to switch to 404 if you want. https://github.com/Apicurio/apicurio-registry/pull/8460#pullrequestreview-4669598999

jsenko (changes requested):
Fixes #8394 — adds HTTP transport with OIDC authentication and per-caller bearer-token forwarding to the MCP server.

## Context

Reviewed the full diff, verified all prior feedback (carlesarnal, paoloantinori, automated reviews) is addressed, and ran an additional security-focused pass. The auth design is solid: privilege escalation guard prevents fallback to service account, token expiry is checked with clock skew tolerance, startup validation catches misconfiguration early, and no secrets are logged.

## Summary

Two minor documentation gaps to close before merge. The implementation itself has no outstanding issues.

### should-fix: Document the `forward-token=false` deployment mode

The PR's "mental model" diagram and integration docs describe two modes: stdio and HTTP with token forwarding. But there's a third valid configuration: HTTP with OIDC inbound gate but `apicurio.mcp.http.forward-token=false`, where the MCP server authenticates callers via OIDC but uses shared service-account credentials (`apicurio.mcp.auth.*`) for all Registry calls.

This is useful for deployments that want multi-user HTTP access to MCP without per-user RBAC at the Registry level. Please add a short section in the integration doc (`assembly-mcp-server-integration.adoc`) explaining:
- What happens when `forward-token=false` (callers are authenticated by OIDC, but all MCP tool calls use the configured service account for Registry)
- When you'd use this (shared deployment where OIDC gates access but per-user RBAC isn't needed)
- Which config properties are required (`apicurio.mcp.http.enabled=true` + OIDC env vars + `apicurio.mcp.auth.*` + `apicurio.mcp.http.forward-token=false`)

An example config block is optional — this will be a rare use case, so a short explanation is sufficient.

### should-fix: Note TLS scope in docs

The TLS config section (`apicurio.mcp.tls.*`) only affects the MCP server's outbound connection to Registry. Please add a brief note clarifying this scope — users might assume it also governs the inbound HTTPS listener, which is configured separately via Quarkus HTTP (`quarkus.http.ssl.*`).

- **Per-request `RegistryClient` creation** — acknowledged as a performance trade-off, acceptable for now. Connection pooling can be added later without API changes.
- **503 vs 404 for disabled HTTP endpoint** — kept 503 per carlesarnal's preference. Both are valid; 503 signals "this could be enabled" which is more helpful.
- **Config properties not in generated docs** — correct by design, the MCP module is a separate application. Properties are documented in the integration guide and README.

---

Thank you for your work on this feature, @Joel-hanson — adding HTTP transport with proper per-user RBAC is a significant contribution. And thank you for your patience while we worked through the review queue. The two items above are minor documentation additions — once addressed, this is ready to merge.

jsenko:

jsenko (approved):

jsenko:
Merged, thank you!

PR #8467, opened 2026-07-06
feat(sdk): add Python AI SDK for MCP, A2A, and Prompt governance

### LINKED ISSUE #8465: feat: Python SDK integration for AI agent ecosystem (MCP discovery, A2A discovery, Prompt governance)
## Context

Apicurio Registry already supports the AI agent trifecta as first-class artifact types: `MCP_TOOL` for tool discovery, `AGENT_CARD` for A2A agent discovery, and `PROMPT_TEMPLATE` for prompt governance. On the Java side, the [quarkus-langchain4j](https://github.com/quarkiverse/quarkus-langchain4j) project has built integrations for MCP discovery (`mcp-apicurio-registry` module) and A2A discovery (`a2a-apicurio-registry`), with prompt template governance proposed in [quarkiverse/quarkus-langchain4j#2658](https://github.com/quarkiverse/quarkus-langchain4j/issues/2658).

The Python AI ecosystem has all the raw building blocks but no equivalent glue library:

| Concern | Existing Python package | Gap |
|---|---|---|
| Registry SDK | [`apicurioregistrysdk`](https://pypi.org/project/apicurioregistrysdk/) v2.5.7 | No AI-specific convenience layer |
| MCP client | [`mcp`](https://pypi.org/project/mcp/) v1.28 + [`langchain-mcp-adapters`](https://pypi.org/project/langchain-mcp-adapters/) v0.3 | No registry-based server discovery |
| A2A | [`a2a-sdk`](https://github.com/a2aproject/a2a-python) v1.1 | No registry-based agent discovery or publishing |
| Prompt templates | `langchain_core.prompts.ChatPromptTemplate` | No registry-backed loading or compatibility validation |
| Deployment | FastAPI | No lifecycle hooks for registry integration |

## Existing landscape

The [MCP Gateway & Registry](https://github.com/agentic-community/mcp-gateway-registry) (agentic-community, [featured on AWS blog](https://aws.amazon.com/blogs/opensource/governing-ai-assets-at-scale-with-mcp-gateway-and-registry/)) is the closest existing Python solution covering similar ground:

| Pillar | MCP Gateway & Registry | Apicurio Registry |
|---|---|---|
| MCP discovery | Built-in MCP Servers Registry | `MCP_TOOL` artifact type |
| A2A discovery | Agent Registry with A2A protocol support | `AGENT_CARD` artifact type |
| Prompt templates | Custom Entity Registry (user-defined schema) | **First-class `PROMPT_TEMPLATE` artifact type** (Prompty-spec aligned) |
| Schema evolution | Duplicate detection, security scanning | **Backward/forward/full compatibility rules** |
| Deployment model | Standalone gateway runtime | Embeddable library pattern |
| Identity | Keycloak, Entra, Okta, Auth0, Cognito | Keycloak (via Quarkus OIDC) |
| AWS federation | AgentCore registry sync | No |

**Key differentiators for Apicurio:**

1. **Prompt templates are first-class** — Apicurio has a dedicated `PROMPT_TEMPLATE` artifact type with Prompty-spec alignment, built-in compatibility checking, and variable validation. MCP Gateway handles prompts as generic custom entities with user-defined schemas — no schema evolution governance.
2. **Schema governance heritage** — Apicurio's compatibility rules (backward, forward, full) are battle-tested from years of API schema governance (Avro, Protobuf, JSON Schema, OpenAPI). This is the core differentiator for prompt governance: preventing breaking changes to prompt variables the same way schema registries prevent breaking API changes.
3. **Embeddable vs. gateway** — Apicurio integration would be a library your app imports, not a separate gateway to deploy. Both models are valid; they serve different operational preferences.

The [official MCP Registry](https://registry.modelcontextprotocol.io/) and [Kong MCP Registry](https://konghq.com/blog/engineering/mcp-registry-dynamic-tool-discovery) also cover MCP server discovery but don't address A2A or prompt governance.

## Proposal

Create an `apicurio-ai-python` package (or similar) that provides Apicurio Registry-backed discovery and governance for Python AI applications. Three pillars:

### Pillar 1: MCP Server Discovery

Search for `MCP_TOOL` artifacts in the registry and dynamically connect MCP clients — the Python equivalent of quarkus-langchain4j's `ApicurioRegistryMcpTools`.

```python
from apicurio_ai import ApicurioMcpDiscovery

discovery = ApicurioMcpDiscovery("http://localhost:8080/apis/registry/v3")

# Search for MCP servers
servers = discovery.search("weather")

# Connect and get LangChain-compatible tools
tools = await discovery.connect("weather-server-v2", group_id="default")
agent = create_react_agent(llm, tools)
```

### Pillar 2: A2A Agent Discovery & Publishing

Search for `AGENT_CARD` artifacts and publish agent cards — equivalent of `ApicurioAgentsRegistry`.

```python
from apicurio_ai import ApicurioAgentDiscovery
from a2a.types import AgentCard

discovery = ApicurioAgentDiscovery("http://localhost:8080/apis/registry/v3")

# Discover agents
agents: list[AgentCard] = discovery.discover_agents()

# Publish this service's agent card on startup
discovery.publish_agent(my_agent_card, group_id="my-team")
```

### Pillar 3: Prompt Template Governance

Load `PROMPT_TEMPLATE` artifacts (Prompty-spec format) as LangChain prompt templates with versioning and compatibility enforcement.

```python
from apicurio_ai import RegistryPromptLoader

loader = RegistryPromptLoader("http://localhost:8080/apis/registry/v3", cache_ttl=60)

# Load a versioned prompt template
prompt = loader.load("customer-support-prompt-v2")
chain = prompt | llm | parser

# Pin to a specific version
prompt_v1 = loader.load("customer-support-prompt", version="1.0.3")
```

## Architecture

A framework-agnostic core wrapping `apicurioregistrysdk` for the three artifact types, plus thin adapter modules:

```
apicurio-ai-python/
├── apicurio_ai/
│   ├── core/              # Registry interactions (MCP_TOOL, AGENT_CARD, PROMPT_TEMPLATE)
│   ├── langchain/         # LangChain adapter (tools, prompts)
│   ├── llamaindex/        # LlamaIndex adapter (optional)
│   └── fastapi/           # FastAPI lifespan hooks, dependency injection
```

### Python-specific governance (no build-time step)

Since Python doesn't have Quarkus's build-time deployment processors, governance checks would happen at:

- **Startup validation**: FastAPI `lifespan` hook pulls templates and validates variable bindings before accepting traffic
- **CI validation**: CLI command (`apicurio-ai validate`) checks prompt compatibility during CI/CD
- **Runtime caching**: TTL-based cache with configurable refresh

## Dependencies

- `apicurioregistrysdk` >= 2.5 (registry REST API)
- `mcp` >= 1.28 (MCP client/server)
- `a2a-sdk` >= 1.0 (A2A protocol)
- `langchain-core` >= 0.3 (optional, for LangChain adapter)
- `fastapi` (optional, for deployment hooks)

## Prior Art

- [quarkus-langchain4j `mcp-apicurio-registry`](https://github.com/quarkiverse/quarkus-langchain4j/tree/main/mcp-apicurio-registry) — Java MCP discovery integration
- [quarkus-langchain4j `a2a-apicurio-registry`](https://github.com/quarkiverse/quarkus-langchain4j) — Java A2A discovery integration
- [quarkiverse/quarkus-langchain4j#2658](https://github.com/quarkiverse/quarkus-langchain4j/issues/2658) — Java prompt governance proposal
- [Apicurio Registry AI-Native blog post](https://www.apicur.io/blog/2026/02/05/apicurio-registry-ai-natural-evolution)
- [MCP Gateway & Registry](https://github.com/agentic-community/mcp-gateway-registry) — Python-based gateway covering similar ground (different deployment model)
- [Official MCP Registry](https://registry.modelcontextprotocol.io/) — MCP server discovery (no A2A/prompt governance)

carlesarnal:
## Summary
Adds a new `apicurio-ai-python` package providing Apicurio Registry-backed discovery and governance for Python AI applications — MCP tool discovery, A2A agent discovery, and prompt template governance. This is the Python equivalent of the quarkus-langchain4j integrations (`mcp-apicurio-registry`, `a2a-apicurio-registry`).

Fixes #8465

## Root Cause
The Python AI ecosystem has all the raw building blocks (MCP client, A2A SDK, LangChain) but no glue library to connect them to Apicurio Registry. The existing `apicurioregistrysdk` is a generated REST client with no AI-specific convenience layer. Additionally, the well-known endpoints (`/.well-known/agents`, `/.well-known/mcp-tools`) had no OpenAPI specification, preventing SDK generation and documentation.

## Changes

**New OpenAPI spec:**
- `common/src/main/resources/META-INF/openapi-wellknown.json` — Standalone OpenAPI 3.0.3 document for 10 well-known endpoints with 16 schemas (AgentCard, McpToolSearchResults, etc.)

**New `apicurio-ai-python/` package (23 files):**
- `pyproject.toml` — Poetry config with core deps (httpx, pydantic) and optional extras (langchain, fastapi, sdk)
- `Makefile` — Build targets (clean, install, test, lint, build, publish)
- `apicurio_ai/core/config.py` — `RegistryConfig` dataclass with URL construction and auth
- `apicurio_ai/core/_models.py` — 17 Pydantic models with camelCase alias support matching Java beans
- `apicurio_ai/core/_wellknown.py` — `WellKnownClient` async httpx client for all well-known endpoints
- `apicurio_ai/core/mcp_discovery.py` — `McpToolDiscovery` class (search, get, publish MCP tools)
- `apicurio_ai/core/a2a_discovery.py` — `AgentDiscovery` class (search, get, publish agent cards, advanced search)
- `apicurio_ai/core/prompt.py` — `PromptGovernance` class (render, get template, publish, get schema)
- `apicurio_ai/langchain/tools.py` — `ApicurioMcpToolkit` converting MCP tools to LangChain `StructuredTool`
- `apicurio_ai/langchain/prompts.py` — `ApicurioPromptTemplate` with registry-backed rendering and LangChain conversion
- `apicurio_ai/fastapi/lifespan.py` — `apicurio_lifespan` async context manager for startup validation
- `apicurio_ai/fastapi/dependencies.py` — FastAPI `Depends()` factories for DI

**Tests:**
- `tests/conftest.py` — Registry Docker fixture and artifact registration helper
- `tests/test_mcp_discovery.py` — MCP tool search and retrieval tests
- `tests/test_a2a_discovery.py` — Agent card discovery and search tests
- `tests/test_prompt.py` — Prompt rendering, template retrieval, and schema tests

**CI integration:**
- `.github/workflows/verify-sdk.yaml` — Added `verify-ai-python-sdk` job
- `.github/workflows/verify.yaml` — Added `apicurio-ai-python/**` to SDK path filter

## Test plan
- [ ] `cd apicurio-ai-python && make install` succeeds
- [ ] `make lint-check` passes (black formatting)
- [ ] Integration tests pass against running registry (`make test`)
- [ ] MCP tool discovery: register MCP_TOOL artifact, search by name, retrieve definition
- [ ] A2A agent discovery: get registry agent card, register AGENT_CARD, search agents
- [ ] Prompt governance: register PROMPT_TEMPLATE, render with variables, get raw content
- [ ] CI workflow `verify-ai-python-sdk` runs on `apicurio-ai-python/**` changes

### DIFF
--- a/apicurio-ai-python/apicurio_ai/core/mcp_discovery.py
+++ b/apicurio-ai-python/apicurio_ai/core/mcp_discovery.py
@@ -1,0 +1,85 @@
+from typing import Any, Optional
+
+import httpx
+
+from apicurio_ai.core._models import McpToolSearchResults
+from apicurio_ai.core._wellknown import WellKnownClient
+from apicurio_ai.core.config import RegistryConfig
+
+
+class McpToolDiscovery:
+    def __init__(self, config: RegistryConfig) -> None:
+        self._config = config
+        self._wellknown = WellKnownClient(config)
+
+    async def close(self) -> None:
+        await self._wellknown.close()
+
+    async def __aenter__(self) -> "McpToolDiscovery":
+        return self
+
+    async def __aexit__(self, *args: Any) -> None:
+        await self.close()
+
+    async def search(
+        self,
+        name: Optional[str] = None,
+        parameters: Optional[list[str]] = None,
+        offset: int = 0,
+        limit: int = 20,
+    ) -> McpToolSearchResults:
+        return await self._wellknown.search_mcp_tools(
+            name=name, parameters=parameters, offset=offset, limit=limit
+        )
+
+    async def get_tool(
+        self,
+        group_id: str,
+        artifact_id: str,
+        version: Optional[str] = None,
+    ) -> dict[str, Any]:
+        return await self._wellknown.get_registered_mcp_tool(
+            group_id=group_id, artifact_id=artifact_id, version=version
+        )
+
+    async def list_all_tools(self) -> list[dict[str, Any]]:
+        results: list[dict[str, Any]] = []
+        offset = 0
+        limit = 100
+        while True:
+            page = await self.search(offset=offset, limit=limit)
+            results.extend(
+                t.model_dump(by_alias=True, exclude_none=True) for t in page.tools
+            )
+            if offset + limit >= page.count:
+                break
+            offset += limit
+        return results
+
+    async def publish_tool(
+        self,
+        artifact_id: str,
+        tool_definition: dict[str, Any],
+        group_id: Optional[str] = None,
+    ) -> None:
+        gid = group_id or self._config.default_group_id
+        url = f"{self._config.api_base_url}/groups/{gid}/artifacts"
+        import json
+
+        async with httpx.AsyncClient(
+            headers=self._config.auth_headers(), timeout=self._config.timeout
+        ) as client:
+            resp = await client.post(
+                url,
+                json={
+                    "artifactId": artifact_id,
+                    "artifactType": "MCP_TOOL",
+                    "firstVersion": {
+                        "content": {
+                            "content": json.dumps(tool_definition),
+                            "contentType": "application/json",
+                        }
+                    },
+                },
+            )
+            resp.raise_for_status()
--- a/apicurio-ai-python/apicurio_ai/core/mcp_discovery.py
+++ b/apicurio-ai-python/apicurio_ai/core/mcp_discovery.py
@@ -1,4 +1,6 @@
+import json
 from typing import Any, Optional
+from urllib.parse import quote
 
 import httpx
 
@@ -11,9 +13,13 @@ class McpToolDiscovery:
     def __init__(self, config: RegistryConfig) -> None:
         self._config = config
         self._wellknown = WellKnownClient(config)
+        self._api_client = httpx.AsyncClient(
+            headers=config.auth_headers(), timeout=config.timeout
+        )
 
     async def close(self) -> None:
         await self._wellknown.close()
+        await self._api_client.aclose()
 
     async def __aenter__(self) -> "McpToolDiscovery":
         return self
@@ -48,6 +54,8 @@ class McpToolDiscovery:
         limit = 100
         while True:
             page = await self.search(offset=offset, limit=limit)
+            if not page.tools:
+                break
             results.extend(
                 t.model_dump(by_alias=True, exclude_none=True) for t in page.tools
             )
@@ -63,23 +71,18 @@ class McpToolDiscovery:
         group_id: Optional[str] = None,
     ) -> None:
         gid = group_id or self._config.default_group_id
-        url = f"{self._config.api_base_url}/groups/{gid}/artifacts"
-        import json
-
-        async with httpx.AsyncClient(
-            headers=self._config.auth_headers(), timeout=self._config.timeout
-        ) as client:
-            resp = await client.post(
-                url,
-                json={
-                    "artifactId": artifact_id,
-                    "artifactType": "MCP_TOOL",
-                    "firstVersion": {
-                        "content": {
-                            "content": json.dumps(tool_definition),
-                            "contentType": "application/json",
-                        }
-                    },
+        url = f"{self._config.api_base_url}/groups/{quote(gid, safe='')}/artifacts"
+        resp = await self._api_client.post(
+            url,
+            json={
+                "artifactId": artifact_id,
+                "artifactType": "MCP_TOOL",
+                "firstVersion": {
+                    "content": {
+                        "content": json.dumps(tool_definition),
+                        "contentType": "application/json",
+                    }
                 },
-            )
-            resp.raise_for_status()
+            },
+        )
+        resp.raise_for_status()
--- a/apicurio-ai-python/apicurio_ai/core/prompt.py
+++ b/apicurio-ai-python/apicurio_ai/core/prompt.py
@@ -1,0 +1,86 @@
+import json
+from typing import Any, Optional
+
+import httpx
+
+from apicurio_ai.core._models import RenderPromptResponse
+from apicurio_ai.core._wellknown import WellKnownClient
+from apicurio_ai.core.config import RegistryConfig
+
+
+class PromptGovernance:
+    def __init__(self, config: RegistryConfig) -> None:
+        self._config = config
+        self._wellknown = WellKnownClient(config)
+
+    async def close(self) -> None:
+        await self._wellknown.close()
+
+    async def __aenter__(self) -> "PromptGovernance":
+        return self
+
+    async def __aexit__(self, *args: Any) -> None:
+        await self.close()
+
+    async def render(
+        self,
+        group_id: str,
+        artifact_id: str,
+        variables: dict[str, Any],
+        version: str = "branch=latest",
+    ) -> RenderPromptResponse:
+        url = (
+            f"{self._config.api_base_url}/groups/{group_id}"
+            f"/artifacts/{artifact_id}/versions/{version}/render"
+        )
+        async with httpx.AsyncClient(
+            headers=self._config.auth_headers(), timeout=self._config.timeout
+        ) as client:
+            resp = await client.post(url, json={"variables": variables})
+            resp.raise_for_status()
+            return RenderPromptResponse.model_validate(resp.json())
+
+    async def get_template(
+        self,
+        group_id: str,
+        artifact_id: str,
+        version: str = "branch=latest",
+    ) -> str:
+        url = (
+            f"{self._config.api_base_url}/groups/{group_id}"
+            f"/artifacts/{artifact_id}/versions/{version}/content"
+        )
+        async with httpx.AsyncClient(
+            headers=self._config.auth_headers(), timeout=self._config.timeout
+        ) as client:
+            resp = await client.get(url)
+            resp.raise_for_status()
+            return resp.text
+
+    async def publish_template(
+        self,
+        artifact_id: str,
+        content: str,
+        group_id: Optional[str] = None,
+        content_type: str = "application/x-yaml",
+    ) -> None:
+        gid = group_id or self._config.default_group_id
+        url = f"{self._config.api_base_url}/groups/{gid}/artifacts"
+        body: dict[str, Any] = {
+            "artifactId": artifact_id,
+            "artifactType": "PROMPT_TEMPLATE",
+            "firstVersion": {
+                "content": {
+                    "content": content,
+                    "contentType": content_type,
+                }
+            },
+        }
+        async with httpx.AsyncClient(
+            headers=self._config.auth_headers(), timeout=self._config.timeout
+        ) as client:
+            resp = await client.post(url, json=body)
+            resp.raise_for_status()
+
+    async def get_schema(self, version: str = "v1") -> dict[str, Any]:
+        return await self._wellknown.get_schema("prompt-template", version)
--- a/apicurio-ai-python/apicurio_ai/core/prompt.py
+++ b/apicurio-ai-python/apicurio_ai/core/prompt.py
@@ -1,5 +1,6 @@
 import json
 from typing import Any, Optional
+from urllib.parse import quote
 
 import httpx
 
@@ -12,9 +13,13 @@ class PromptGovernance:
     def __init__(self, config: RegistryConfig) -> None:
         self._config = config
         self._wellknown = WellKnownClient(config)
+        self._api_client = httpx.AsyncClient(
+            headers=config.auth_headers(), timeout=config.timeout
+        )
 
     async def close(self) -> None:
         await self._wellknown.close()
+        await self._api_client.aclose()
 
     async def __aenter__(self) -> "PromptGovernance":
         return self
@@ -30,15 +35,13 @@ class PromptGovernance:
         version: str = "branch=latest",
     ) -> RenderPromptResponse:
         url = (
-            f"{self._config.api_base_url}/groups/{group_id}"
-            f"/artifacts/{artifact_id}/versions/{version}/render"
+            f"{self._config.api_base_url}/groups/{quote(group_id, safe='')}"
+            f"/artifacts/{quote(artifact_id, safe='')}"
+            f"/versions/{quote(version, safe='')}/render"
         )
-        async with httpx.AsyncClient(
-            headers=self._config.auth_headers(), timeout=self._config.timeout
-        ) as client:
-            resp = await client.post(url, json={"variables": variables})
-            resp.raise_for_status()
-            return RenderPromptResponse.model_validate(resp.json())
+        resp = await self._api_client.post(url, json={"variables": variables})
+        resp.raise_for_status()
+        return RenderPromptResponse.model_validate(resp.json())
 
     async def get_template(
         self,
@@ -47,15 +50,13 @@ class PromptGovernance:
         version: str = "branch=latest",
     ) -> str:
         url = (
-            f"{self._config.api_base_url}/groups/{group_id}"
-            f"/artifacts/{artifact_id}/versions/{version}/content"
+            f"{self._config.api_base_url}/groups/{quote(group_id, safe='')}"
+            f"/artifacts/{quote(artifact_id, safe='')}"
+            f"/versions/{quote(version, safe='')}/content"
         )
-        async with httpx.AsyncClient(
-            headers=self._config.auth_headers(), timeout=self._config.timeout
-        ) as client:
-            resp = await client.get(url)
-            resp.raise_for_status()
-            return resp.text
+        resp = await self._api_client.get(url)
+        resp.raise_for_status()
+        return resp.text
 
     async def publish_template(
         self,
@@ -65,7 +66,7 @@ class PromptGovernance:
         content_type: str = "application/x-yaml",
     ) -> None:
         gid = group_id or self._config.default_group_id
-        url = f"{self._config.api_base_url}/groups/{gid}/artifacts"
+        url = f"{self._config.api_base_url}/groups/{quote(gid, safe='')}/artifacts"
         body: dict[str, Any] = {
             "artifactId": artifact_id,
             "artifactType": "PROMPT_TEMPLATE",
@@ -76,11 +77,8 @@ class PromptGovernance:
                 }
             },
         }
-        async with httpx.AsyncClient(
-            headers=self._config.auth_headers(), timeout=self._config.timeout
-        ) as client:
-            resp = await client.post(url, json=body)
-            resp.raise_for_status()
+        resp = await self._api_client.post(url, json=body)
+        resp.raise_for_status()
 
     async def get_schema(self, version: str = "v1") -> dict[str, Any]:
         return await self._wellknown.get_schema("prompt-template", version)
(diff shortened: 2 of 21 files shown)


apicurio-ai-python/apicurio_ai/fastapi/lifespan.py
```diff
@@ -0,0 +1,39 @@
...
+    app: FastAPI, config: RegistryConfig
+) -> AsyncGenerator[None, None]:
+    mcp = McpToolDiscovery(config)
+    agents = AgentDiscovery(config)
+    prompts = PromptGovernance(config)
+
+    try:
+        await agents.get_registry_agent_card()
+    except Exception:
+        pass
```
paoloantinori: Bare `except Exception: pass` here will mask auth/config errors as a clean startup. Consider logging at warning, or narrowing to `httpx.HTTPError` so genuine programming errors still surface during app boot.

apicurio-ai-python/apicurio_ai/langchain/tools.py
```diff
@@ -0,0 +1,58 @@
...
+        raise NotImplementedError(
+            f"MCP tool '{name}' was loaded from the registry for schema/metadata only. "
+            "Connect to the actual MCP server to execute this tool."
+        )
+
+    return StructuredTool.from_function(
+        func=_not_implemented,
+        name=name,
+        description=description,
+        args_schema=None,
```
paoloantinori, line 28: `input_schema` is read on line 17 but discarded here (`args_schema=None`). The resulting `StructuredTool` exposes no parameter schema to the LLM, so the agent can't tell what arguments the tool expects. Either drop the dead assignment (if stubs are intentionally metadata-only) or feed the MCP JSON schema into `args_schema` (e.g. via a Pydantic model built from `inputSchema`).

apicurio-ai-python/apicurio_ai/core/prompt.py
```diff
@@ -0,0 +1,86 @@
...
+    async def render(
+        self,
+        group_id: str,
+        artifact_id: str,
+        variables: dict[str, Any],
+        version: str = "branch=latest",
+    ) -> RenderPromptResponse:
+        url = (
+            f"{self._config.api_base_url}/groups/{group_id}"
+            f"/artifacts/{artifact_id}/versions/{version}/render"
```
paoloantinori: `version` is f-stringed straight into the path. The default `"branch=latest"` is safe, but a caller passing `1.0.0` or a branch with `/`, `?`, `#`, or spaces will silently hit the wrong endpoint. `urllib.parse.quote(version, safe="")` would make this robust — same applies to `get_template` below.

apicurio-ai-python/apicurio_ai/langchain/prompts.py
```diff
@@ -0,0 +1,64 @@
...
+    async def to_langchain(self) -> ChatPromptTemplate:
+        raw = await self._governance.get_template(
+            group_id=self._group_id,
+            artifact_id=self._artifact_id,
+            version=self._version,
+        )
+        template_text = _extract_template_text(raw)
+        langchain_template = re.sub(r"\{\{(\w+)\}\}", r"{\1}", template_text)
+        return ChatPromptTemplate.from_template(langchain_template)
+
```
paoloantinori, line 46: `\w+` won't match variable names containing hyphens or dots, and any literal `{`/`}` already in the template will break `ChatPromptTemplate.from_template` downstream. Worth confirming against the variable-name rules the registry's prompt-template schema actually permits.

paoloantinori (commented):
Solid foundation for the Python AI SDK — the discovery/render/publish surface maps cleanly onto the new well-known endpoints, and the async-first design with context-manager cleanup is the right shape. A few concrete issues below, the most pressing being that the `Verify AI Python SDK` CI job is currently red on `black` formatting, so the test suite never actually runs against a live registry in CI.

Note: I only reviewed the new `apicurio-ai-python/` sources and the `openapi.json` additions. The Go SDK churn (~30k lines of mechanical regeneration) and the `verify-sdk.yaml` changes I skimmed but did not line-by-line. The full PR diff was too large to fetch as a single response (GitHub returns HTTP 406 `diff too_large`), so my review is scoped to the files I read in full.

- **CI: `make lint-check` fails, blocking the integration tests.** The latest run on `ace3433` ([run 28813621802](https://github.com/Apicurio/apicurio-registry/actions/runs/28813621802), job `Verify AI Python SDK`) exits at the lint step — `black` reports 4 files would be reformatted: `apicurio_ai/core/a2a_discovery.py`, `apicurio_ai/langchain/tools.py`, `tests/conftest.py`, `apicurio_ai/core/_wellknown.py`. Because `lint-check` runs before `make test`, the integration tests (which spin up the registry container and exercise the real endpoints) have never executed in CI for this PR. Worth running `make lint-apply` and pushing so the test matrix actually validates the SDK against a live registry.

- **`apicurio_lifespan` swallows all errors from `get_registry_agent_card()` silently.** A bare `except Exception: pass` ([lifespan.py:26-29](https://github.com/Apicurio/apicurio-registry/blob/main/apicurio-ai-python/apicurio_ai/fastapi/lifespan.py#L26-L29)) will mask misconfiguration (wrong `registry_url`, expired `auth_token`, DNS) as a successful startup — the app boots but every subsequent discovery call fails opaquely. Consider logging at least at `warning`, or narrowing the catch to `httpx.HTTPError` so genuine programming errors still propagate.

- **`_build_tool_from_definition` reads `inputSchema` but never uses it.** The variable is assigned ([tools.py:17](https://github.com/Apicurio/apicurio-registry/blob/main/apicurio-ai-python/apicurio_ai/langchain/tools.py#L17)) and then `args_schema=None` is passed to `StructuredTool.from_function` ([tools.py:29](https://github.com/Apicurio/apicurio-registry/blob/main/apicurio-ai-python/apicurio_ai/langchain/tools.py#L29)). As a result the generated `StructuredTool` exposes no parameter schema to the LLM, so the agent cannot know what arguments the tool expects — the discovery becomes name/description only. If the goal is metadata-only stubs (as the `_not_implemented` body suggests), the dead `input_schema` line should be dropped; if the intent is to surface the schema, convert the MCP JSON Schema to a Pydantic model (or pass via `args_schema`) before constructing the tool.

- **`version` is interpolated into URL paths without encoding.** In `PromptGovernance.render` and `get_template` ([prompt.py:30-34, 47-51](https://github.com/Apicurio/apicurio-registry/blob/main/apicurio-ai-python/apicurio_ai/core/prompt.py#L30-L51)), `version` (default `"branch=latest"`) is f-stringed straight into the path. The default works, but any caller passing a semantic version like `1.0.0` or a branch name containing `/`, `?`, `#`, or spaces will silently hit the wrong endpoint or a 404. Worth wrapping with `urllib.parse.quote(version, safe="")` (and applying the same to `group_id`/`artifact_id` in the publish/render paths for defense in depth).

- Minor: `ApicurioPromptTemplate.to_langchain` ([prompts.py:46](https://github.com/Apicurio/apicurio-registry/blob/main/apicurio-ai-python/apicurio_ai/langchain/prompts.py#L46)) uses `re.sub(r"\{(\w+)\}", ...)` to convert `{{var}}` → `{var}`. `\w+` won't match variable names with hyphens or dots (which the prompt-template schema permits), and any literal `{` or `}` already present in the template will later break `ChatPromptTemplate.from_template`. Hedged — depends on what the registry actually allows as variable names.

Happy to take a closer look at the `openapi.json` well-known additions (the new `AgentCard`/`McpToolSearchResults` schemas) if useful; they looked consistent with the Python models at a glance.

apicurio-ai-python/apicurio_ai/core/mcp_discovery.py
```diff
@@ -0,0 +1,85 @@
...
+        group_id: str,
+        artifact_id: str,
+        version: Optional[str] = None,
+    ) -> dict[str, Any]:
+        return await self._wellknown.get_registered_mcp_tool(
+            group_id=group_id, artifact_id=artifact_id, version=version
+        )
+
+    async def list_all_tools(self) -> list[dict[str, Any]]:
+        results: list[dict[str, Any]] = []
```
paoloantinori, line 52: Breaking only on `offset + limit >= page.count` risks an infinite loop if `count` is larger than the real result set (concurrent deletes, stale counts, eventual consistency). Add a `if not page.tools: break` guard so an empty page terminates the loop regardless of `count`.

apicurio-ai-python/apicurio_ai/core/mcp_discovery.py
```diff
@@ -0,0 +1,85 @@
...
+                t.model_dump(by_alias=True, exclude_none=True) for t in page.tools
+            )
+            if offset + limit >= page.count:
+                break
+            offset += limit
+        return results
+
+    async def publish_tool(
+        self,
+        artifact_id: str,
```
paoloantinori, line 69: This creates a new `httpx.AsyncClient` per `publish_tool()` call, throwing away connection pooling on every publish. Consider reusing a shared client (and routing auth/timeout through it consistently) rather than re-deriving `auth_headers()` and `timeout` inline here.

paoloantinori (commented):
The async-first, context-managed design is solid and the discovery surface maps cleanly onto the well-known endpoints. Having now read the core modules in full, a few concrete issues worth addressing:

- **`list_all_tools()` can infinite-loop on a stale or inflated `count`.** The termination condition is `offset + limit >= page.count`. If the server reports a `count` larger than the actual number of items returned across pages (e.g. concurrent deletes, eventual-consistency reads, or any mismatch between `count` and the pageable result set), `offset` keeps advancing, the returned pages are empty, but the loop never breaks because `offset + limit` never catches `count`. Add a guard: break when `len(page.tools) == 0` (an empty page is the canonical "no more data" signal, independent of `count`). This also handles the exact-multiple boundary more robustly than the `>=` check.
- **`publish_tool()` opens a throwaway `httpx.AsyncClient` per call** instead of reusing a pooled connection. For a publish-heavy loop this discards connection reuse and TLS setup every time. More structurally, it bypasses `WellKnownClient` entirely (writing to the `/groups` REST API rather than a well-known path), so there's no shared timeout/auth/transport config — `auth_headers()` and `timeout` are re-read from `config` inline. Consider routing publish through a shared client (the `WellKnownClient` or a sibling REST client) so auth/timeout/transport are configured once.
- **Auth only supports a static bearer token (`auth_headers()`).** Apicurio Registry deployments commonly use OIDC with token refresh, API keys, or mTLS. A single `auth_token: Optional[str]` field can't express any of those. For an SDK meant to be the convenient glue layer, a pluggable auth callable (or accepting an `httpx.Auth` instance directly) would cover far more deployments without much added complexity. Not blocking for a first cut, but worth a tracking note.

apicurio-ai-python/apicurio_ai/langchain/tools.py
```diff
@@ -0,0 +1,56 @@
...
+    description = tool_def.get("description", "")
+    input_schema = tool_def.get("inputSchema", {})
+
+    def _not_implemented(**kwargs: Any) -> str:
+        raise NotImplementedError(
+            f"MCP tool '{name}' was loaded from the registry for schema/metadata only. "
+            "Connect to the actual MCP server to execute this tool."
+        )
+
+    return StructuredTool.from_function(
```
paoloantinori, line 24: `input_schema` is read here but never used — `StructuredTool.from_function(..., args_schema=None)` two lines below discards it. Converting the JSON Schema in `input_schema` to a Pydantic `args_schema` (or deriving one via `pydantic.create_model`) is what would make these tools actually invocable by an LLM.

paoloantinori (commented):
Third pass on the current head. The core discovery/render surface still holds up; this read focused on the LangChain adapter and the test workflow, where I found one concrete functional gap and a couple of smaller items:

- **`ApicurioMcpToolkit` discards the `inputSchema` it reads (most important).** In `apicurio_ai/langchain/tools.py`, `_build_tool_from_definition()` extracts `input_schema = tool_def.get("inputSchema", {})` but then passes `args_schema=None` to `StructuredTool.from_function`. Since the whole point of an MCP tool definition is the input schema, building the tool without it means the LLM has no way to know what arguments to pass — the tool is effectively metadata-only and uncallable in a routing loop. The Pydantic conversion (inputSchema JSON Schema → a `args_schema` BaseModel, or `create_model` from the schema) is the missing piece. The `_not_implemented` placeholder is fine for schema-only loading, but the schema itself must be wired through for this to be useful.

- **`get_tools()` is an N+1 over HTTP.** For each row returned by `search()` it issues a second `get_tool()` round-trip. Acceptable for small registries, but worth noting in the README or adding a `max_tools` cap so a large registry doesn't turn one search into hundreds of requests.

- **`conftest.py` `event_loop` fixture is deprecated in pytest-asyncio ≥0.23.** Given `pytest-asyncio = "^0.21.0"` in pyproject, the 0.21 floor still accepts the `event_loop` fixture, but any resolver bump to ≥0.23 will emit a deprecation warning and eventually break the suite. Worth pinning or migrating to the `loop_scope` marker style.

- **CI is red on the last 6 commits.** Not a code review point per se, but since the verify-sdk job (added in `.github/workflows/verify-sdk.yaml`) is the main gate for this new package and it hasn't been green, it's hard to confirm the runtime behavior claimed in the tests. Worth checking the latest `verify-ai-python-sdk` run before merge.

jsenko (commented):

Good implementation overall. paoloantinori already covered the critical items (input_schema, pagination guard, URL encoding, bare except). Adding 4 should-fix observations:

### 1. Auth token exposure in warning log (`lifespan.py:34`)

```python
logger.warning("Registry connectivity check failed: %s", e)
```

`httpx.HTTPError` string representation can include request details containing `Authorization: Bearer <token>`. Consider logging only the error type and status code:

```python
logger.warning("Registry connectivity check failed: %s", type(e).__name__)
```

### 2. PyYAML dependency not declared (`langchain/prompts.py:57`, `pyproject.toml`)

`yaml.safe_load` is used for YAML template parsing but `pyyaml` is not declared in `pyproject.toml` — not in core deps, not in extras, not documented. The `except Exception: pass` fallback means YAML templates silently degrade to raw text with no user feedback. Either add `pyyaml` to a `yaml` extra or log a warning when import fails.

### 3. Weak test assertions (`test_a2a_discovery.py:53-54`, `test_mcp_discovery.py:42-43`)

```python
assert card.name is not None    # checks existence only
assert card.skills is not None  # checks existence only
```

These should assert specific values — the test fixtures define known data (`"test-weather-agent"`, 1 skill). Example:

```python
assert card.name == "Apicurio Registry"  # or whatever the well-known card returns
assert len(card.skills) >= 1
```

### 4. No error-path test coverage

No tests for: HTTP 404 (missing artifact), 401 (bad auth), malformed JSON responses, or network timeouts. At minimum, a test verifying behavior when an artifact doesn't exist would confirm the SDK surfaces errors clearly rather than returning confusing results.

PR #8517, opened 2026-07-10
fix(kafkasql): wire up periodic KafkaSQL snapshot scheduling (#8499)

### LINKED ISSUE #8499: Automatic KafkaSQL snapshot scheduling (apicurio.kafkasql.snapshot.every.seconds) is declared but never wired to any scheduler
### Summary

Apicurio Registry's KafkaSQL storage variant defines a configuration property, `apicurio.kafkasql.snapshot.every.seconds` (default `86400s`), that implies periodic, automatic snapshots of the internal journal are created on that interval. In practice, no code path ever reads this property to schedule anything. The only way a snapshot is ever created is by manually calling `POST /admin/snapshots`. Without an external caller doing that on a timer, the `kafkasql-journal` topic grows without bound and replica startup time degrades over time and the exact failure mode snapshotting was introduced to prevent.

This has been confirmed against a running 3.2.x JVM and against the `main` branch source tree (which includes 3.3.0, the current latest release).

### Environment

- Apicurio Registry version: 3.2.x (please confirm exact patch — `3.2.6` does not appear in the official release list; the closest published tags are `3.2.5` and `3.3.0`)
- Storage variant: KafkaSQL
- Deployment: Kubernetes

### Evidence

**1. The config property exists but is isolated.**

```java
// app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfiguration.java
@ConfigProperty(name = "apicurio.kafkasql.snapshot.every.seconds", defaultValue = "86400s")
```

A repo-wide search on `main` turns up exactly 4 references to this property in the entire codebase:
- the `@ConfigProperty` declaration above
- the default value line in `app/src/main/resources/application.properties`
- two configuration-reference doc pages (`ref-registry-all-configs.adoc`, `assembly-config-reference.adoc`)

There is no reference to it in any scheduler, consumer, or storage-service class.

**2. The running JVM's actual `@Scheduled` tasks (6 total, matching the `SimpleScheduler$ScheduledTask` count) are:**

```
RegistryStorageConfigCache_ScheduledInvoker_run
DownloadReaper_ScheduledInvoker_run
DynamicLogConfigurationService_ScheduledInvoker_applyDynamicLogLevel
OrphanedContentReaper_ScheduledInvoker_run
GitOpsRegistryStorage_ScheduledInvoker_scheduledRefresh
KubernetesOpsRegistryStorage_ScheduledInvoker_scheduledRefresh
```

None of these create snapshots or reference `snapshot.every.seconds`.

**3. `triggerSnapshotCreation()` (or the equivalent internal method that performs the snapshot-and-publish flow described in the [[3.0 KafkaSQL storage blog post](https://www.apicur.io/blog/2024/06/21/registry-3-kafkasql)](https://www.apicur.io/blog/2024/06/21/registry-3-kafkasql)) is only ever invoked from a unit test via direct CDI injection.** No REST resource and no scheduled job calls it.

**4. Manual triggering does work.** `POST /admin/snapshots` is a real, documented, currently-shipping endpoint — it's covered in the 3.0 KafkaSQL blog post and is still present in the auto-generated client SDKs on `main` (e.g. `go-sdk/pkg/registryclient-v3/admin/snapshots_request_builder.go`, `{+baseurl}/admin/snapshots`). This confirms the snapshot creation *mechanism* is implemented and functional — it is only the *automatic, interval-based triggering* of that mechanism that is missing.

**5. No fix has landed as of 3.3.0 (latest, released 2026-06-08).** Reviewed the 3.2.0 through 3.3.0 release notes; none mention snapshot scheduling. Related KafkaSQL fixes did land in this window (e.g. #7796 "DeleteAllExpiredDownloads scheduled job should not produce messages to the Kafka journal", #7795 "new pod gets stuck during rolling update", #7753 "KafkaSQL data not recovered after restart") but none address the missing scheduler for snapshots.

### Expected behavior

When `apicurio.kafkasql.snapshot.every.seconds` is set, the registry should periodically (on that interval) perform the same snapshot-creation flow that `POST /admin/snapshots` performs, without requiring an external caller.

### Actual behavior

The property is declared, defaulted, and documented, but nothing in the application reads it or acts on it. Snapshots are only ever created via a manual `POST /admin/snapshots` call. Environments that rely on the documented config property to get automatic backups / faster replica startup get neither, and the journal topic grows unbounded until something else (retention policy, manual snapshot, or an outage) intervenes.

### Suggested fix

Add a 7th `@Scheduled` task (following the existing pattern used by `DownloadReaper`, `OrphanedContentReaper`, etc.) that reads `apicurio.kafkasql.snapshot.every.seconds` and periodically invokes the same internal snapshot-creation logic exposed via `POST /admin/snapshots`.

### Workaround

Until this is fixed upstream, an external scheduled job (e.g. a Kubernetes `CronJob`) calling `POST /admin/snapshots` on the desired interval reproduces the intended behavior using the endpoint that already works today.

### Reproduction steps

1. Deploy Apicurio Registry 3.2.x or 3.3.0 with KafkaSQL storage.
2. Set `apicurio.kafkasql.snapshot.every.seconds` to any value (e.g. `60s`).
3. Write enough data to the journal to make a difference in startup time, then wait well past the configured interval.
4. Observe: no message is ever published to the `kafkasql-snapshots` topic, and no snapshot file appears at `apicurio.storage.snapshot.location`, unless `POST /admin/snapshots` is called manually.

nagindersingh:
## Summary
Fixes #8499. Adds a periodic scheduler that automatically triggers KafkaSQL snapshot creation, so the `apicurio.kafkasql.snapshot.every.seconds` config property actually does what its name promises.

## Root Cause
`apicurio.kafkasql.snapshot.every.seconds` (default `86400s`) was declared in `KafkaSqlConfiguration` but never read by any scheduler, consumer, or storage service. Snapshots only happened when manually triggered via `POST /admin/snapshots`. Without automatic snapshots, the internal `kafkasql-journal` topic grows unbounded, degrading replica startup performance over time.

## Changes
- Added `KafkaSqlSnapshotScheduler` (`app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java`):
  - `@ApplicationScoped` bean gated by `@LookupIfProperty(name = "apicurio.storage.kind", stringValue = "kafkasql")`, so it's only active for the KafkaSQL storage variant (snapshot creation isn't supported for `sql` storage).
  - `@Scheduled` method reads the existing `apicurio.kafkasql.snapshot.every.seconds` property via Quarkus's `{...}` interpolation, following the same pattern as `DownloadReaper` and `OrphanedContentReaper`.
  - Calls `storage.triggerSnapshotCreation()` (the same method used by the `/admin/snapshots` REST endpoint), guarded by `storage.isReady()` / `storage.isReadOnly()` checks, with exceptions caught and logged so a failed snapshot doesn't kill the scheduler thread.
- Added `KafkaSqlSnapshotSchedulerTest` (`app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotSchedulerTest.java`): Mockito-based unit tests covering triggering when ready/writable, skipping when not ready, skipping when read-only, and swallowing exceptions.

No `@ConfigProperty`/`@Info` changes were needed — the property already existed with its annotation and default, so the generated config docs (`ref-registry-all-configs.adoc`) don't need regeneration.

## Test plan
- [x] Added `KafkaSqlSnapshotSchedulerTest` (4 unit tests, all passing)
- [x] Full `app` module build, including checkstyle, passes
- [ ] Manual/integration verification against a running KafkaSQL cluster (confirm periodic snapshot files appear in `apicurio.storage.snapshot.location` on the configured interval)
- [x] SQL storage variant unaffected — the scheduler bean is not registered when `apicurio.storage.kind != kafkasql`


app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
```diff
@@ -0,0 +1,54 @@
...
+    Logger log;
+
+    @Inject
+    @Current
+    RegistryStorage storage;
+
+    /**
+     * Minimal granularity is 1 minute.
+     */
+    @Scheduled(delay = 60, concurrentExecution = SKIP, every = "{apicurio.kafkasql.snapshot.every.seconds}")
```
EricWittmann: **Bug: `delay = 60` means 60 minutes, not 60 seconds**

Quarkus `@Scheduled` defaults `delayUnit` to `TimeUnit.MINUTES` ([source](https://github.com/quarkusio/quarkus/blob/main/extensions/scheduler/api/src/main/java/io/quarkus/scheduler/Scheduled.java)). Without an explicit `delayUnit`, `delay = 60` sets the initial delay to **60 minutes (1 hour)**, not 60 seconds.

The comment "Minimal granularity is 1 minute" suggests the intent was a 1-minute initial delay.

For reference, recently-added files in this codebase correctly specify the unit explicitly:
- [`UsageTelemetryBuffer.java`](https://github.com/Apicurio/apicurio-registry/blob/6e48a301ca2de2f6aac70b007f0a149230fadcce/app/src/main/java/io/apicurio/registry/storage/impl/sql/UsageTelemetryBuffer.java#L55-L57): `delay = 5, delayUnit = TimeUnit.SECONDS`
- [`UsageAggregationJob.java`](https://github.com/Apicurio/apicurio-registry/blob/6e48a301ca2de2f6aac70b007f0a149230fadcce/app/src/main/java/io/apicurio/registry/storage/impl/sql/UsageAggregationJob.java#L32-L34): `delay = 30, delayUnit = TimeUnit.SECONDS`

```suggestion
    @Scheduled(delay = 1, concurrentExecution = SKIP, every = "{apicurio.kafkasql.snapshot.every.seconds}")
```

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
```diff
@@ -0,0 +1,54 @@
...
+import java.time.Instant;
+
+import static io.quarkus.scheduler.Scheduled.ConcurrentExecution.SKIP;
+
+/**
+ * Periodically triggers creation of a KafkaSQL snapshot, bounding the growth of the "kafkasql-journal" topic.
+ * Only active when the KafkaSQL storage variant is selected.
+ */
+@ApplicationScoped
+@LookupIfProperty(name = "apicurio.storage.kind", stringValue = "kafkasql")
```
EricWittmann, line 29: **Bug: `@LookupIfProperty` does not prevent `@Scheduled` methods from firing**

`@LookupIfProperty` only controls programmatic CDI lookup (via `Instance<T>`). The Quarkus scheduler discovers `@Scheduled` methods at build time and invokes them **regardless** of `@LookupIfProperty`, so this scheduled method will fire on all storage variants — not just kafkasql.

On non-kafkasql deployments, `storage.triggerSnapshotCreation()` delegates to [`AbstractSqlRegistryStorage.triggerSnapshotCreation()`](https://github.com/Apicurio/apicurio-registry/blob/6e48a301ca2de2f6aac70b007f0a149230fadcce/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java#L1724-L1727) which throws `RegistryStorageException("Directly triggering the snapshot creation is not supported for sql storages.")`. The try-catch prevents a crash, but it produces a **spurious error log every 24 hours** on every non-kafkasql deployment.

Both [`GitOpsRegistryStorage`](https://github.com/Apicurio/apicurio-registry/blob/6e48a301ca2de2f6aac70b007f0a149230fadcce/app/src/main/java/io/apicurio/registry/storage/impl/gitops/GitOpsRegistryStorage.java#L44-L49) and [`KubernetesOpsRegistryStorage`](https://github.com/Apicurio/apicurio-registry/blob/6e48a301ca2de2f6aac70b007f0a149230fadcce/app/src/main/java/io/apicurio/registry/storage/impl/kubernetesops/KubernetesOpsRegistryStorage.java#L86-L91) solve this with an explicit runtime guard inside their `@Scheduled` methods:

```java
if ("gitops".equals(registryStorageType)) {
    tryRefresh();
}
```

Suggested fix: inject the storage kind and add a similar guard:

```java
@ConfigProperty(name = "apicurio.storage.kind")
String registryStorageType;
```

Then at the top of `run()`:

```java
if (!"kafkasql".equals(registryStorageType)) {
    return;
}
```

EricWittmann (commented):
Code review — 2 issues found.

carlesarnal (changes requested):
Implementation is straightforward and follows the existing scheduler patterns (`DownloadReaper`, etc.).

Two things to fix:

1. The config property `apicurio.kafkasql.snapshot.every.seconds` needs `@Info` annotation with proper category metadata so it shows up in generated docs.

2. Consider adding a named constant for the 60-second initial delay in `KafkaSqlSnapshotScheduler` instead of a magic number.

Guard logic and exception handling look solid. Test coverage is adequate for the happy path and basic error cases.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
nagindersingh: fix is pushed

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
EricWittmann: This is still unaddressed in the latest push. The `@LookupIfProperty` annotation only gates CDI programmatic lookup (`Instance<T>`) — the Quarkus scheduler discovers `@Scheduled` methods at build time and will invoke `run()` on **every** storage variant, not just kafkasql.

On non-kafkasql deployments this means `storage.triggerSnapshotCreation()` throws `RegistryStorageException` every 24 hours, producing a spurious error log.

Suggested fix — inject the storage kind and add a runtime guard, matching the pattern used by `GitOpsRegistryStorage` and `KubernetesOpsRegistryStorage`:

```java
@ConfigProperty(name = "apicurio.storage.kind")
String registryStorageType;
```

Then at the top of `run()`:

```java
if (!"kafkasql".equals(registryStorageType)) {
    return;
}
```

EricWittmann:
@nagindersingh — the `@LookupIfProperty` issue from the earlier review is still unaddressed in the latest push. This is a bug that will affect all non-kafkasql deployments.

**The problem:** `@LookupIfProperty` only gates programmatic CDI lookup (`Instance<T>`). The Quarkus scheduler discovers `@Scheduled` methods at build time and invokes them regardless of `@LookupIfProperty`. This means `run()` will fire on **every** storage variant — not just kafkasql. On sql/gitops/kubernetesops deployments, `storage.triggerSnapshotCreation()` will throw `RegistryStorageException` every 24 hours, producing a spurious error log.

**Suggested fix:** inject the storage kind and add a runtime guard at the top of `run()`, matching the pattern used by `GitOpsRegistryStorage` and `KubernetesOpsRegistryStorage`:

```java
@ConfigProperty(name = "apicurio.storage.kind")
String registryStorageType;
```

Then at the top of `run()`:

```java
if (!"kafkasql".equals(registryStorageType)) {
    return;
}
```

You can keep the `@LookupIfProperty` annotation as well (it still prevents unnecessary programmatic injection), but the runtime guard is what actually prevents the scheduler from executing on the wrong storage variant.

EricWittmann:
@nagindersingh — friendly ping on this. The `@LookupIfProperty` issue from the Jul 16 review is still
the main blocker. Without a runtime guard in `run()`, the scheduler will fire on all storage variants
and produce spurious errors on non-kafkasql deployments.

The suggested fix is small — inject the storage kind and early-return if it's not `kafkasql` (see the
Jul 16 comment for the exact code). @carlesarnal's feedback (missing `@Info`, magic number) should also
be addressed.

Let us know if you're still working on this or need any help. We'd like to get this into an upcoming
release.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
```diff
@@ -0,0 +1,70 @@
...
+    @ConfigProperty(name = "apicurio.storage.kind")
+    @Info(category = CATEGORY_STORAGE, description = "Application storage variant, for example, sql, kafkasql, gitops, or kubernetesops", availableSince = "3.0.0")
+    String registryStorageType;
+
+    @Inject
+    Logger log;
+
+    @Inject
+    @Current
+    RegistryStorage storage;
```
paoloantinori, line 44: This is the runtime guard @EricWittmann's Jul 27 ping asked for — it's present and correct. Combined with `testDoesNotRunOnNonKafkaSqlStorage` (parameterized over `sql`/`gitops`/`kubernetesops`, asserting `verifyNoInteractions`), the non-kafka path is covered.

paoloantinori (commented):
Re-read of the current head (`2bb021b0`). Picking up on @EricWittmann's Jul 27 ping — the runtime guard he flagged as missing **is present** in this head and looks to fully address the concern.

- `run()` opens with `if (!KAFKASQL_STORAGE_KIND.equals(registryStorageType)) { return; }` — that is the runtime check that prevents the scheduler from acting on non-kafkasql deployments.
- The class javadoc correctly explains *why* both layers are needed: `@LookupIfProperty` only gates programmatic/CDI lookup, while the Quarkus scheduler discovers `@Scheduled` methods at build time regardless of it — so the runtime check inside `run()` is the actual guard. This matches the pattern in `GitOpsRegistryStorage` / `KubernetesOpsRegistryStorage`.
- `KafkaSqlSnapshotSchedulerTest.testDoesNotRunOnNonKafkaSqlStorage` parameterizes over `sql`, `gitops`, `kubernetesops` and asserts `verifyNoInteractions(storage)`, directly proving the guard works for each non-kafka variant.

Other points that look solid:
- `concurrentExecution = SKIP` prevents overlapping snapshot runs — correct for a job whose runtime is unpredictable.
- The `try/catch (Exception)` in `run()` logs and swallows, so a failed snapshot doesn't kill the scheduler thread.
- `isReady()` / `isReadOnly()` guards before `triggerSnapshotCreation()` mirror the REST `/admin/snapshots` path.

One non-blocking note: with the runtime guard in place, `@LookupIfProperty` on the class is effectively redundant for the scheduler path (it only adds value for hypothetical programmatic CDI lookups of this bean). Keeping it as belt-and-suspenders is harmless and consistent with other storage beans, so no change needed — just flagging it so the intent is clear to future readers.

From my side the blocker Eric cited is addressed; the head looks good.

EricWittmann (commented):
One more change needed: the automatic snapshot feature should be **opt-in**, not enabled by default.

Currently, the `apicurio.kafkasql.snapshot.every.seconds` property defaults to `86400s`, which means every KafkaSQL deployment will automatically start creating snapshots every 24 hours with no way to disable it other than changing the config.

Please change the default value to `"off"` (Quarkus `@Scheduled` recognizes `"off"` and `"disabled"` as values that prevent the method from firing). Users who want automatic snapshots can then explicitly set the interval:

```properties
apicurio.kafkasql.snapshot.every.seconds=86400s
```

This means updating the `defaultValue` in `KafkaSqlConfiguration.java` and regenerating the config docs.

EricWittmann (commented):
One more thing: please check whether the KafkaSQL snapshot feature has any existing documentation (e.g. in the `docs/` directory or the Apicurio docs site). If it does, it likely needs to be updated to explain:

1. That automatic snapshots are now available (and how to enable them by setting `apicurio.kafkasql.snapshot.every.seconds` to a duration value like `86400s`)
2. That the feature is opt-in / disabled by default
3. That the manual `POST /admin/snapshots` endpoint remains available as an alternative

If no snapshot documentation exists yet, consider adding a brief section — this is a feature users need to know about to keep their KafkaSQL journal topic from growing unbounded.

EricWittmann:
Closing as superseded by #9275, which incorporates the original work with additional fixes (opt-in enabled flag, multi-replica coordination, dynamic property with UI visibility gating, Settings UI grouping). Thank you @nagindersingh for the original contribution — you are credited as co-author on the new PR.

Issue #8211, opened 2026-06-11
EPIC: Production Observability


AleStranieri:
## Summary

Consolidate Apicurio Registry's observability capabilities into a coherent, production-grade story - from OTel tracing and metrics through usage telemetry and schema health monitoring.

The observability umbrella #6939 (closed) established initial OTel support. This EPIC tracks the remaining work to make Apicurio Registry fully observable in enterprise monitoring stacks (Prometheus, Grafana, Datadog, Jaeger).

---

## Current State

| Capability | Status |
|-----------|--------|
| OTel server-side tracing | Done - PR #7031 (initial), PR #7560 (W3C propagation) |
| OTel attribute conventions | Done - PR #8062 (snake_case + OTelAttributes constants) |
| JDBC telemetry | Done - #8061 |
| OTel metrics provider | Partial - #7150 (6 counters defined but wiring incomplete) |
| SerDes OTel spans | Not started - #8156 |
| Metrics compatibility (Datadog) | Not started - #8152 |
| Defensive metrics wrapping | Not started - #8181 |
| Usage telemetry (schema health) | Experimental - PR #7881 (merged, requires feature flag) |

---

## Phase 1 - OTel Completion (P0)
> Close remaining gaps in tracing and metrics. External signal: Datadog user (#8152).

- [ ] #8156 - SerDes OTel spans: application-level spans and metrics in SerDes libraries
- [ ] #8152 - Metrics distribution fix: make REST timer configurable (summary vs histogram)
- [ ] #8181 - Defensive metrics: wrap OTelMetricsProvider counter increments in try-catch
- [ ] #7150 - OTel improvements: wire remaining custom counters (artifact/version/validation)

## Phase 2 - Usage Telemetry (P1)
> Enable schema health monitoring. Experimental feature in 3.x, GA target TBD.

- [ ] Usage telemetry GA readiness: stabilize Active/Stale/Dead schema classification (PR #7881)
- [ ] Deprecation workflow integration: surface usage data in governance endpoints
- [ ] UI integration: schema health indicators in registry UI

## Phase 3 - North Star Instrumentation (P2)
> Enable measurement of Weekly Active Schema Operations. Depends on Phase 2.

- [ ] Define "schema operation" event taxonomy (create, read, validate, resolve, serialize)
- [ ] Aggregate usage telemetry into reportable NSM proxy signals
- [ ] Optional opt-in reporting for self-hosted deployments

---

## Related Issues

[...]

PR #8519, opened 2026-07-10
feat(metrics): add configurable distribution type for REST metrics

### LINKED ISSUE #8152: Make REST metrics distribution configurable — rest_requests_seconds is exposed as both summary and histogram, breaking OpenMetrics/Datadog scrapers
## Summary

The REST timer metric `rest_requests_seconds` is published on the Prometheus `/metrics` endpoint as **both** a summary (client-side percentiles, `{quantile="…"}` lines) **and** a histogram (SLO buckets, `_bucket{le="…"}` lines) under the same metric name. Backends that map a Prometheus timer to a single type cannot reconcile a meter that is simultaneously a summary and a histogram, and end up dropping most of the series.

It would help to make the REST-metrics distribution **configurable** (and/or default to histogram-only) instead of hardcoding both.

## Environment

- Apicurio Registry **3.1.7** (the relevant code is unchanged on `main` / 3.2.x as well)
- Quarkus Micrometer, Prometheus exporter (`/metrics`)
- Scraper: Datadog Agent OpenMetrics v2 check

## What happens

For the REST timer, `/metrics` emits all of these at once:

```
rest_requests_seconds{...,quantile="0.5"}   ...
rest_requests_seconds{...,quantile="0.95"}  ...
rest_requests_seconds_bucket{...,le="0.1"}  ...
rest_requests_seconds_bucket{...,le="+Inf"} ...
rest_requests_seconds_count{...}            ...
rest_requests_seconds_sum{...}              ...
rest_requests_seconds_max{...}              ...
```

The Datadog OpenMetrics v2 integration cannot resolve the mixed summary+histogram type and ingests **only** `rest_requests_seconds_max`. The `count`, `sum`, `bucket` and `quantile` series are all dropped, so request rate, error rate (by `status_code_group`) and latency percentiles are unavailable for the REST API.

For contrast, the sibling plain timer `storage_method_call_seconds` (which exposes only `count`/`sum`/`max`, no dual distribution) is ingested fully. That isolates the problem to the combined percentiles + SLO configuration on `rest_requests_seconds`, not to the scraper setup.

## Root cause

`app/src/main/java/io/apicurio/registry/metrics/CustomMetricsConfiguration.java`:

```java
return DistributionStatisticConfig.builder()
    .percentiles(0.5, 0.95, 0.99, 0.9995)
    .serviceLevelObjectives(0.1 * factor, 0.25 * factor, 0.5 * factor,
                            1.0 * factor, 2.0 * factor, 5.0 * factor, 10.0 * factor)
    .build().merge(config);
```

`.percentiles(...)` produces the summary `{quantile=}` lines; `.serviceLevelObjectives(...)` produces the histogram `_bucket{le=}` lines. Emitting both for the same meter is what creates the ambiguous exposition. There is currently no application property to influence this, so the only workaround is patching/rebuilding.

## Impact

Any deployment scraping Apicurio metrics with a backend that maps a Prometheus timer to a single type loses request-rate / error-rate / latency for the REST API. This is not unique to Datadog — see micrometer-metrics/micrometer#3494 for a comparable case where a timer exposing both forms fails to be scraped.

## Request

Make the REST-metrics distribution configurable instead of hardcoding both forms — for example an application property such as:

```
apicurio.metrics.rest.distribution = histogram | percentiles | both   (default: histogram)
```

A histogram-only default would work cleanly with server-side percentile computation in most backends (Prometheus `histogram_quantile`, Datadog distributions, etc.), while still letting operators opt back into client-side percentiles if they prefer. At minimum, allowing client-side `.percentiles(...)` to be turned off via config would unblock affected setups.

Happy to open a PR if a config-based approach sounds acceptable.

paoloantinori:
## Summary

- Makes REST metrics distribution type configurable for Datadog compatibility
- Adds 3 new config properties: `apicurio.metrics.rest.distribution.type`, `.percentiles`, `.slo`
- Default behavior unchanged (histogram with existing percentiles and SLO boundaries)
- Summary mode disables histogram buckets, exporting only pre-computed percentiles (recommended for Datadog)

## Related Issues

- Closes #8152 (Metrics distribution fix for Datadog)
- Part of EPIC #8211 (Production Observability)

## Changes

| File | Change |
|------|--------|
| `CustomMetricsConfiguration.java` | 3 `@ConfigProperty` + `@Info` fields, conditional SLO boundaries |
| `CustomMetricsConfigurationTest.java` | 8 unit tests: default histogram, summary mode, custom boundaries, edge cases |

## Configuration

```properties
# Default (unchanged behavior)
apicurio.metrics.rest.distribution.type=histogram

# For Datadog — export pre-computed percentiles only
apicurio.metrics.rest.distribution.type=summary
```

## Test plan

- [ ] All 8 CustomMetricsConfiguration tests pass
- [ ] Default histogram behavior preserved
- [ ] Summary mode works for Datadog use case
- [ ] Checkstyle passes

### DIFF
--- a/app/src/main/java/io/apicurio/registry/metrics/CustomMetricsConfiguration.java
+++ b/app/src/main/java/io/apicurio/registry/metrics/CustomMetricsConfiguration.java
@@ -66,6 +66,24 @@ public class CustomMetricsConfiguration {
     @Singleton
     public MeterFilter configureDistribution() {
         String type = distributionType.toLowerCase(Locale.ROOT);
+        if (!"histogram".equals(type) && !"summary".equals(type)) {
+            throw new IllegalArgumentException(
+                    "Invalid value for 'apicurio.metrics.rest.distribution.type': '"
+                            + distributionType + "'. Valid values are: histogram, summary");
+        }
+        percentiles.stream().filter(p -> p < 0.0 || p > 1.0).findFirst().ifPresent(p -> {
+            throw new IllegalArgumentException(
+                    "Invalid percentile value " + p
+                            + " in 'apicurio.metrics.rest.distribution.percentiles'. "
+                            + "Each value must be between 0.0 and 1.0.");
+        });
+        slo.stream().filter(s -> s <= 0.0).findFirst().ifPresent(s -> {
+            throw new IllegalArgumentException(
+                    "Invalid SLO boundary value " + s
+                            + " in 'apicurio.metrics.rest.distribution.slo'. "
+                            + "Each value must be a positive number representing seconds.");
+        });
+
         double[] percentilesArray = toDoubleArray(percentiles);
         double[] sloNanosArray = toSloNanosArray(slo);
 
--- a/app/src/main/java/io/apicurio/registry/metrics/CustomMetricsConfiguration.java
+++ b/app/src/main/java/io/apicurio/registry/metrics/CustomMetricsConfiguration.java
@@ -20,11 +20,7 @@ public class CustomMetricsConfiguration {
 
     private static final Logger log = LoggerFactory.getLogger(CustomMetricsConfiguration.class);
 
-    /**
-     * Conversion factor from seconds to nanoseconds for SLO boundaries.
-     * Micrometer Timer records values in nanoseconds, so SLO boundaries
-     * expressed in seconds must be multiplied by this factor.
-     */
+    // Micrometer Timer records values in nanoseconds.
     private static final double SECONDS_TO_NANOSECONDS = 1_000_000_000.0;
 
     @Info(description = """
@@ -62,10 +58,14 @@ public class CustomMetricsConfiguration {
             defaultValue = "0.1,0.25,0.5,1.0,2.0,5.0,10.0")
     List<Double> slo;
 
+    /**
+     * Produces a MeterFilter that configures the distribution statistics for REST request metrics.
+     * The returned filter captures immutable snapshots of the config values, making it thread-safe.
+     */
     @Produces
     @Singleton
-    public MeterFilter enableHistogram() {
-        String type = distributionType != null ? distributionType.toLowerCase(Locale.ROOT) : "histogram";
+    public MeterFilter configureDistribution() {
+        String type = distributionType.toLowerCase(Locale.ROOT);
         double[] percentilesArray = toDoubleArray(percentiles);
         double[] sloNanosArray = toSloNanosArray(slo);
 
@@ -89,14 +89,14 @@ public class CustomMetricsConfiguration {
     }
 
     private static double[] toDoubleArray(List<Double> values) {
-        if (values == null || values.isEmpty()) {
+        if (values.isEmpty()) {
             return new double[0];
         }
         return values.stream().mapToDouble(Double::doubleValue).toArray();
     }
 
     private static double[] toSloNanosArray(List<Double> sloSeconds) {
-        if (sloSeconds == null || sloSeconds.isEmpty()) {
+        if (sloSeconds.isEmpty()) {
             return new double[0];
         }
         return sloSeconds.stream()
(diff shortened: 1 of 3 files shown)


EricWittmann (commented):
**Finding: Unknown distribution type silently falls through to summary behavior**

`CustomMetricsConfiguration.java` — `configureDistribution()`

If a user misconfigures the distribution type (e.g. `apicurio.metrics.rest.distribution.type=histgram` — a typo), the code silently treats it as summary mode (no SLO boundaries). The `testUnknownTypeTreatedAsSummary` test documents this as intentional, but in practice a typo would silently change metrics behavior with no indication — the user thinks they have histogram buckets, but they don't. This would be very hard to diagnose in production.

**Suggestion:** Fail fast at startup if the value is invalid. For example, validate in `configureDistribution()` (or a `@PostConstruct` method):

```java
String type = distributionType.toLowerCase(Locale.ROOT);
if (!"histogram".equals(type) && !"summary".equals(type)) {
    throw new IllegalArgumentException(
            "Invalid value for 'apicurio.metrics.rest.distribution.type': '" + distributionType
                    + "'. Valid values are: histogram, summary");
}
```

This follows the project's general pattern of failing early on invalid configuration rather than silently degrading. The corresponding test `testUnknownTypeTreatedAsSummary` would need to be updated to expect the exception.

EricWittmann (commented):
**Finding: No validation on percentile values**

`CustomMetricsConfiguration.java` — `percentiles` field

The `@Info` description states that each value must be between 0.0 and 1.0, but the code does not enforce this. If a user sets `apicurio.metrics.rest.distribution.percentiles=5.0,99.0` (confusing percentiles with percentage values), the invalid values are passed directly to Micrometer's `.percentiles()` which expects values in `[0.0, 1.0]`.

**Suggestion:** Validate at startup and fail fast:

```java
percentiles.stream()
        .filter(p -> p < 0.0 || p > 1.0)
        .findFirst()
        .ifPresent(p -> {
            throw new IllegalArgumentException(
                    "Invalid percentile value " + p
                            + " in 'apicurio.metrics.rest.distribution.percentiles'. "
                            + "Each value must be between 0.0 and 1.0 "
                            + "(e.g. 0.95 for the 95th percentile).");
        });
```

This is a very easy misconfiguration to make (writing `95` instead of `0.95`) and the resulting metrics would be silently meaningless rather than producing a clear startup error.

EricWittmann (commented):
**Finding: No validation on SLO values**

`CustomMetricsConfiguration.java` — `slo` field

SLO boundary values represent durations in seconds, so they should be positive. If a user sets `apicurio.metrics.rest.distribution.slo=-1.0,0.5`, the negative value is converted to a negative nanosecond value and passed to Micrometer's `.serviceLevelObjectives()` without complaint.

**Suggestion:** Validate at startup and fail fast:

```java
slo.stream()
        .filter(s -> s <= 0.0)
        .findFirst()
        .ifPresent(s -> {
            throw new IllegalArgumentException(
                    "Invalid SLO boundary value " + s
                            + " in 'apicurio.metrics.rest.distribution.slo'. "
                            + "Each value must be a positive number representing seconds.");
        });
```

paoloantinori:
@EricWittmann All 3 findings addressed in commit 8bce3ef:

1. **Invalid distribution type**: Now throws `IllegalArgumentException` at startup instead of silently falling through to summary behavior.
2. **Invalid percentile values**: Validated to be in `[0.0, 1.0]` — catches the common `95.0` vs `0.95` mistake.
3. **Invalid SLO values**: Validated to be positive — catches negative duration values.

Tests updated: `testUnknownTypeTreatedAsSummary` → `testUnknownTypeThrowsException`, plus new `testInvalidPercentileThrowsException` and `testNegativeSloThrowsException`.

carlesarnal (approved):
LGTM — clean implementation. Config properties follow conventions (`apicurio.` prefix, `@Info` with `CATEGORY_OBSERVABILITY`, `defaultValue` in annotation). Input validation with clear error messages. `Locale.ROOT` for case-insensitive comparison. Default values preserve backward compatibility. Thorough test coverage (11 tests including edge cases).

PR #8524, opened 2026-07-10
fix: collapse whitespace in config generator property descriptions

krish-kedia:
Fixes #8523

Problem Description:
Currently, the AsciiDoc configuration reference generator (ref-registry-all-configs.adoc) directly extracts the raw string values from @Info annotations. When developers utilize Java 15+ Text Blocks (""") combined with line continuations (\) to cleanly format long descriptions in the source code, the compiler suppresses the newline but strictly preserves the leading indentation of the continuation line.
Consequently, the AsciiDoc generator dumps these preserved indentation blocks directly into the documentation, creating severe multi-space layout artifacts (e.g., "metrics.         You") that degrade readability and break table formatting.

Solution:
This PR resolves the formatting artifact by introducing a robust whitespace normalization step directly into the Configuration Generator logic. Instead of policing how developers format their source code annotations, the generator now automatically sanitizes the extracted strings prior to AsciiDoc injection.

Implementation:
- Replaced the direct .toString() extraction in the generator pipeline with a regex-driven whitespace collapse operation: .replaceAll("\\s+", " ").trim().
- The \s+ pattern captures all consecutive whitespace characters (spaces, tabs, carriage returns, and newlines), collapsing them into a single delimiter space. The .trim() ensures trailing/leading bounds remain clean.

PR #8536, opened 2026-07-11
fix: refactor KubernetesAuthenticationStrategy to use Caffeine AsyncCache

### LINKED ISSUE #8461: Replace hand-rolled token cache with Caffeine in KubernetesAuthenticationStrategy
## Context

From code review on #8076, the `KubernetesAuthenticationStrategy` uses a hand-rolled `ConcurrentHashMap` cache with several limitations:

1. **O(n) eviction on every put** — `evictExpiredEntries()` iterates the entire map on every `cachePut` call
2. **Silent drop when full** — when cache reaches `MAX_CACHE_SIZE` (10,000), new entries are silently dropped, causing all requests to fall through to the K8s API
3. **Thundering-herd on cache miss** — concurrent requests with the same token can both see a cache miss and both call `performTokenReview`

## Proposed solution

Replace with Caffeine (already in Quarkus dependency tree via `quarkus-cache`):

```java
private final Cache<String, TokenReviewResult> cache = Caffeine.newBuilder()
    .maximumSize(MAX_CACHE_SIZE)
    .expireAfterWrite(Duration.ofMinutes(authConfig.kubernetesTokenCacheExpiration))
    .build();
```

This gives:
- O(1) amortized puts with automatic LRU eviction (no silent drops)
- Built-in expiration without manual scanning
- `get(key, mappingFunction)` for thundering-herd protection

Relates to #8076

krish-kedia:
This refactor addresses multiple performance and concurrency bottlenecks in the Kubernetes authentication caching layer:

- Replaced the manual ConcurrentHashMap implementation with Caffeine's AsyncCache, resolving the O(n) eviction iteration overhead that was occurring on every cache put.

- Eliminated the silent token drop behavior when the cache reached MAX_CACHE_SIZE. Eviction is now handled cleanly by Caffeine's native LRU bounds tracking.

- Introduced CompletableFuture-based asynchronous caching. Cache hits now resolve instantly on the Vert.x event loop without incurring a thread context-switch.

- Cache misses now safely delegate the synchronous Fabric8 TokenReview API call to the Mutiny worker pool (Infrastructure.getDefaultWorkerPool()).

- Provided thundering-herd protection: Concurrent requests presenting the same uncached token now subscribe to a single pending CompletableFuture, ensuring only one underlying Kubernetes API call is triggered under heavy load.

- Utilized a custom Caffeine Expiry policy to strictly bound API error caching to 30 seconds. This maintains the application's retry-storm protection during Kubernetes API outages without locking out valid tokens for the full 5-minute default expiration.

Fixes: #8461

### DIFF
--- a/app/src/main/java/io/apicurio/registry/auth/KubernetesAuthenticationStrategy.java
+++ b/app/src/main/java/io/apicurio/registry/auth/KubernetesAuthenticationStrategy.java
@@ -35,33 +35,54 @@ import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.time.Duration;
-import java.time.Instant;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
-import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CompletableFuture;
+
+import com.github.benmanes.caffeine.cache.AsyncCache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.Expiry;
 
 public class KubernetesAuthenticationStrategy implements AuthenticationStrategy {
 
     private static final String BEARER_PREFIX = "Bearer ";
     private static final int MAX_CACHE_SIZE = 10_000;
-    private static final Duration API_ERROR_CACHE_DURATION = Duration.ofSeconds(30);
 
     private final KubernetesClient kubernetesClient;
     private final AuthConfig authConfig;
     private final Logger log;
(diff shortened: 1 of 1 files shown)


krish-kedia:
hi @jsenko, took account of the suggestions given by copilot. specifically:
fixed the expireAfterUpdate logic so that the TTL correctly re-evaluates upon refresh rather than locking to the previous duration, and cleaned up the performTokenReview method contract by dropping the unused tokenHash parameter.

also, a small performance optimization. swapped out String.format(Locale.ROOT, "%02x", b) in the sha256() method for direct bitwise integer arithmetic (Integer.toHexString(val & 0xff)). this skips the heavy regex parsing and object allocations associated with String.format(), making the auth pipeline lil faster.

@carlesarnal does this implementation look good?

carlesarnal:
@krish-kedia please, only one pr at a time. I'm closing this until https://github.com/Apicurio/apicurio-registry/pull/8524 is merged.

PR #8537, opened 2026-07-11
fix: replace maven-dependency-plugin with resources plugin to fix MDEP-98

### LINKED ISSUE #8431: Build: mvn compile -pl app -am fails with MDEP-98 (unpack-openapi)
## Problem

Running `./mvnw compile -Dlocal -DskipTests -pl app -am` fails with:

```
Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.8.1:unpack 
(unpack-openapi) on project apicurio-registry-app: Artifact has not been packaged yet. 
When used on reactor artifact, unpack should be executed after packaging: see MDEP-98.
```

The `compile` goal doesn't trigger `package`, but the `unpack-openapi` execution in the app module depends on a reactor artifact being packaged first.

## Workaround

Use `./mvnw package -pl app -am -DskipTests` instead of `compile`. The `package` phase produces the artifacts that `unpack` needs.

## Impact

- Developers using `mvn compile` for quick feedback loops hit this error
- The `-Dlocal` profile doesn't help since it only skips plugins, not lifecycle phases
- CI uses `package` so it's not affected

krish-kedia:
Fixes: #8431 

This PR resolves a build failure triggered by MDEP-98, a long-standing issue in the maven-dependency-plugin.

When developers run early-phase lifecycle commands for fast feedback loops (such as mvn generate-resources or mvn compile) on this multi-module project, the build fails. The maven-dependency-plugin attempts to unpack an artifact from the common sibling module, but because the build hasn't reached the package phase yet, the actual .jar file does not exist in the reactor. This strict reactor resolution failure is prevalent across Maven 3.x and continues to be an issue in Maven 4.x.

Fix: By switching from the maven-dependency-plugin (which requires a packaged artifact) to the maven-resources-plugin (which copies directly from the source directory), we completely bypass MDEP-98. This restores the ability for developers to run fast mvn compile loops without requiring a full mvn package execution, while maintaining identical output paths.

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -509,34 +509,7 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-dependency-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>unpack-openapi</id>
-            <goals>
-              <goal>unpack</goal>
-            </goals>
-            <phase>generate-resources</phase>
-            <configuration>
-              <artifactItems>
-                <artifactItem>
-                  <groupId>${project.groupId}</groupId>
-                  <artifactId>apicurio-registry-common</artifactId>
-                  <version>${project.version}</version>
-                  <type>jar</type>
-                  <overWrite>true</overWrite>
-                  <includes>**/openapi.json</includes>
-                </artifactItem>
-              </artifactItems>
-              <outputDirectory>${project.build.outputDirectory}</outputDirectory>
-              <overWriteReleases>false</overWriteReleases>
-              <overWriteSnapshots>true</overWriteSnapshots>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
+
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-assembly-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


krish-kedia:
copilot flagged that copy-resources lacked the default overwrite behavior of the previous maven-dependency-plugin unpack. to prevent stale files from persisting during incremental local builds, explicit overwrite has been enabled for both: openapi-to-api-docs and dist executions

carlesarnal:
@krish-kedia please, only one pr at a time. I'm closing this until https://github.com/Apicurio/apicurio-registry/pull/8524 is merged.

carlesarnal (changes requested):
The change addresses MDEP-98 by switching from unpacking a JAR to copying from source. The approach works but has issues.

Problems:

1. `app/pom.xml` line 667: Relative path `../common/src/main/resources/META-INF` is fragile. Consider using `${project.parent.basedir}/common/src/main/resources/META-INF` for better portability.

2. `app/pom.xml` lines 688, 707: Adding `<overwrite>true</overwrite>` to existing executions changes behavior. Explain why this is needed or remove if not required.

3. Commit message references MDEP-98 but provides no context. What scenario triggered this? Which Maven version? Add this to the PR description.

4. The comment says "fails on mvn compile" but the old execution ran in `generate-resources` phase, not `compile`. Clarify the actual failure scenario.

The fix itself is reasonable but needs refinement. The relative path is the main concern.

krish-kedia:
updated the path to use ${project.parent.basedir}/common/src/main/resources/META-INF for better portability.

about the overwrite tags, copilot suggested adding them to prevent stale files during incremental builds, but since maven-resources-plugin naturally overwrites when the source is newer anyway, they aren't needed. removed them to preserve the original behavior.

updated the PR description with full context and corrected the inline comment to accurately reflect that the failure happens during the generate-resources phase (before packaging), rather than incorrectly stating it happens on compile.

thankyou for the review @carlesarnal, lmk if this looks good

carlesarnal (approved):

EricWittmann:
Thanks @krish-kedia :+1:

PR #8538, opened 2026-07-11
feat(storage): implement SQL structured content search for agent discovery filters

### LINKED ISSUE #8058: [Feature] SQL structured content search for agent discovery filters
## Parent Epic
Part of #6991: AI Agent Registry (MCP & A2A Support)

## Background

The agent discovery endpoints (`GET /.well-known/agents`, `POST /.well-known/agents/search`) support filtering by skills, capabilities, input/output modes, protocol bindings, and tags. These filters use `SearchFilter.ofStructure()` which is only implemented in the Elasticsearch search backend.

On SQL-based storage variants (PostgreSQL, KafkaSQL, H2, MySQL, MSSQL), structure filters are **silently ignored** — the search returns all matching agents as if no filter was applied.

This also affects MCP tool search (`GET /.well-known/mcp-tools?parameter=...`).

## Problem

There is no `artifact_structured_content` table in the SQL schema. Structured elements extracted by `AgentCardStructuredContentExtractor` and `McpToolStructuredContentExtractor` are only indexed into Elasticsearch documents.

**Affected filters on SQL deployments:**
- `skill` — `agent_card:skill:<id>`
- `capability` — `agent_card:capability:<name>`
- `inputMode` — `agent_card:inputmode:<mode>`
- `outputMode` — `agent_card:outputmode:<mode>`
- `protocolBinding` — `agent_card:protocolbinding:<binding>`
- `tag` — `agent_card:tag:<tag>`
- `securityScheme` — `agent_card:securityscheme:<name>`
- `parameter` (MCP tools) — `mcp_tool:parameter:<name>`

**Working filters on SQL (unaffected):**
- Label filters (`SearchFilter.ofLabel`) — uses `artifact_labels` table
- Name filter (`SearchFilter.ofName`) — uses `artifacts.name` / `artifacts.artifactId`
- Artifact type filter — uses `artifacts.type`

### 1. New table: `artifact_structured_content`

```sql
CREATE TABLE artifact_structured_content (
    groupId VARCHAR(512) NOT NULL,
    artifactId VARCHAR(512) NOT NULL,
    elementType VARCHAR(64) NOT NULL,
    elementValue VARCHAR(512) NOT NULL,
    FOREIGN KEY (groupId, artifactId) REFERENCES artifacts(groupId, artifactId) ON DELETE CASCADE
);

CREATE INDEX IDX_asc_type_value ON artifact_structured_content(elementType, elementValue);
CREATE INDEX IDX_asc_artifact ON artifact_structured_content(groupId, artifactId);
```

DDL needed for: postgresql, h2, mysql, mssql + upgrade scripts.

### 2. Populate on artifact create/update

When an artifact version is created, run the `StructuredContentExtractor` for the artifact type and insert rows into `artifact_structured_content`. Clear existing rows on update.

### 3. Implement `case structure:` in SqlSearchRepository

```java
case structure:
    op = filter.isNot() ? "NOT EXISTS" : "EXISTS";
    where.append(op + "(SELECT 1 FROM artifact_structured_content sc " +
        "WHERE sc.groupId = a.groupId AND sc.artifactId = a.artifactId " +
        "AND sc.elementType || ':' || sc.elementValue = ?)");
    binders.add((query, idx) -> {
        query.bind(idx, filter.getStringValue().toLowerCase());
    });
    break;
```

### 4. Backfill migration

Upgrade script that populates `artifact_structured_content` for all existing AGENT_CARD and MCP_TOOL artifacts.

## Scope

This is a storage-layer change that affects all SQL dialects. It benefits both agent card and MCP tool search.

## Workaround

Until this is implemented, SQL/KafkaSQL users can use label-based filtering as an alternative to structured content filters. Labels work across all storage variants.

## References

- PR #7927 — Agent Discovery & Entitlement Endpoints (where this limitation was identified)
- #6991 — Epic
- `SqlSearchRepository.java` — missing `case structure:` handler
- `AgentCardStructuredContentExtractor.java` — extracts structured elements
- `ElasticsearchSearchService.java` — current structure filter implementation (Elasticsearch only)

parthdagia05:
## Summary

Adds SQL artifact-level structured-content indexing for agent discovery and other structured metadata queries. Fixes #8058.

Updated against current main in signed-off merge `a8addeb876`; review-remediation commit `908d607b4d`.

## Implementation

- Database version 110 introduces `artifact_structured_content` across H2/PostgreSQL/MySQL/SQL Server, cascading artifact deletion and indexed normalized element type/value.
- SQL search supports full `type:kind:value`, partial `kind:value` and plain `value` forms. Matching is normalized and exact; partial-kind SQL wildcard metacharacters are escaped. Blank components are rejected; negation requires a positive artifact-type filter.
- Index follows the published, non-disabled latest branch tip. Draft writes do not replace it. Publication/state changes, deletion, imported versions/branches and mutable latest-branch operations refresh under the artifact lock and transaction.
- Import indexing covers GitOps/KubernetesOps and the v2 empty-branch/version/append load order. A provisional latest-published-version fallback is used only before branch metadata exists.
- OpenAPI/AsyncAPI YAML is normalized to JSON before extraction.
- Extraction errors are logged/counted and produce no stale entries. SQL errors propagate and roll back the surrounding write—essential for PostgreSQL transaction correctness.
- Backfill skips no-op extractors, follows branch selection and clears stale rows on repeat. Upgrade markers are written after backfill; H2/MySQL creation is retryable after implicit-commit DDL. MySQL regression exercises failure and rerun using the upgrade script.
- PostgreSQL content insertion is conflict-safe during enclosing draft publication; MySQL binary value collation avoids accent-equivalence collisions with normalized Java keys.
- Current `GET /.well-known/agents` supplies SQL structure prefilters to shared candidate collection, retaining visibility filtering and final content checks. Removed obsolete POST/public/entitled test assumptions after merging main.

## Verification

- [x] **93 tests passed** across H2, PostgreSQL, MySQL, real KafkaSQL, GitOps/KubernetesOps import smoke tests, current A2A REST filters and OpenAPI companion regressions.
- [x] Exact/partial/plain filters, literal wildcards, blank values and scoped negation.
- [x] Draft edit/publication, disabled/restored/deleted tip, branch order distinct from version order, imported empty branch append/replace/delete.
- [x] SQL index constraint failure rolls back new version and previous index; retry succeeds.
- [x] Duplicate draft content publication remains usable on PostgreSQL.
- [x] Accent-distinct values survive live writes and backfill on MySQL.
- [x] MySQL failed backfill leaves version 109; retry completes and sets 110.
- [x] Checkstyle app/common: zero violations.
- [ ] SQL Server runtime verification (DDL updated; not executed locally).
- [ ] CI on head `908d607b4d` and maintainer re-review.

## Scope / coordination

- This is an artifact-level selected-content index, not historical per-version full-text search. Version-level content/structure search through Elasticsearch remains separate.
- PR #9563 adds ES identity/intersection behavior. Its SQL fallback rejecting structure filters must not replace this implementation; its removed legacy A2A endpoint tests need reconciliation with current main.
- Backfill has a one-time content extraction cost with progress logging. Custom provider backfill requires a provider available to the upgrader; no external search infrastructure is required for built-in structured artifact filters.
- Existing runtime/read-only policies remain unchanged. SQL Server runtime results will not be represented as passing until executed.


carlesarnal (changes requested):
The structured content search implementation is well thought out — the `StructuredContentExtractor` SPI and the upgrader pattern for backfill are clean.

A few things:

- `asLowerCase()` — verify the implementation in `StringUtil.java` uses `Locale.ROOT` per project conventions.

- The backfill upgrader processes all artifacts synchronously on startup. For large registries this could cause noticeable startup delays. Consider adding progress logging and possibly batch processing.

- Race condition risk in `updateStructuredContentRaw()`: concurrent updates (delete + insert) could lose data if two threads process the same artifact. Wrap in an explicit transaction or use an upsert pattern.

- The feature silently does nothing on non-SQL storage variants. Either add validation to reject structure filters when SQL storage isn't active, or document this limitation.

- Add `@Override` on `StructuredContentUpgrader.upgrade()`.

- Test coverage is light. Add tests for malformed filter inputs, concurrent updates, and backfill with a large dataset.

parthdagia05:
> This implementation has a critical SQL injection vulnerability in the search filter construction. User-supplied filter values are concatenated into LIKE patterns without escaping wildcards. Fix by escaping `%` and `_` in `structureParts[0]` before binding.
> 
> The feature silently fails on non-SQL storage variants. Add validation to reject structure filters when SQL storage is not active, or document this limitation prominently.
> 
> Multiple files contain incorrect annotation syntax that appears to be copy-pasted file paths. Remove all `@app/src/test/...` and `@schema-util/...` annotations from production code.
> 
> The `asLowerCase()` utility method must use `Locale.ROOT` per project conventions. Verify implementation in `StringUtil.java`.
> 
> Add `@Override` annotations to `StructuredContentUpgrader.upgrade()` and `ArtifactContentRowMapper.map()`.
> 
> The backfill upgrader processes all artifacts synchronously. For large registries this could cause startup delays. Consider adding progress logging and batch processing.
> 
> Race condition in `updateStructuredContentRaw()`: concurrent updates can lose data. Wrap delete+insert in explicit transaction or use upsert pattern.
> 
> Test coverage is minimal. Add tests for SQL injection attempts, malformed inputs, concurrent updates, and backfill performance.

I will look onto this feedback and make changes accordingly

parthdagia05:
Thanks @carlesarnal, pushed fixes for all of these. For the non-SQL variants I documented the limitation and added a warning log for now, let me know if you'd like the full import-path indexing done in this PR too.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -167,6 +168,47 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
...
+                            where.append("sc.elementType = ? AND sc.elementValue = ?");
+                            binders.add((query, idx) -> {
+                                query.bind(idx, structureParts[0] + ":" + structureParts[1]);
+                            });
+                            binders.add((query, idx) -> {
+                                query.bind(idx, structureParts[2]);
+                            });
+                        } else if (structureParts.length == 2) {
+                            // Partial format: "kind:name" - match the kind for any artifact type
+                            where.append("sc.elementType LIKE ? AND sc.elementValue = ?");
```
paoloantinori: This `LIKE ?` is bound to `"%:" + structureParts[0]` without escaping `%`/`_` in `structureParts[0]`. The earlier review explicitly asked for LIKE-wildcard escaping; this new 2-part branch looks like it reintroduces an unescaped LIKE against a request-derived value. Worth escaping `%`/`_` here too (with a declared `ESCAPE` char) or confirming `kind` is constrained to a safe charset upstream.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -167,6 +168,47 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
                         break;
                     case content:
                         throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                    case structure:
+                        // Structured content filter, e.g. "agent_card:skill:translation". Elements are
+                        // stored lowercased in artifact_structured_content as elementType
+                        // ("<artifactType>:<kind>") + elementValue ("<name>").
+                        String structureValue = asLowerCase(filter.getStringValue());
+                        if (structureValue == null || structureValue.isBlank()) {
+                            where.append("1 = 1");
```
paoloantinori: A blank structure value falls through to `1 = 1` (match-all). For a negated filter this becomes `NOT EXISTS(... 1=1)`, matching nothing. Confirming whether a blank `structure` value is reachable from the API would be useful — if it is, match-all / match-none semantics for an empty filter may surprise callers.

paoloantinori (commented):
Strong implementation overall — the `StructuredContentExtractor` SPI, the delete-then-insert update path sharing the version-content transaction, and the db-upgrade backfill are all well structured, and the import-path gap is documented honestly. The follow-ups to the earlier review feedback (lowercasing via `Locale.ROOT`, transactional version-content update) look addressed. A few observations from the diff:

- **`LIKE` wildcard escaping in the 2-part structure path (`SqlSearchRepository.java:196`):** when the filter is in the `kind:name` form, `elementType` is matched with `LIKE ?` bound to `"%:" + structureParts[0]`. If `structureParts[0]` (the kind) or the stored `elementType` ever contains `%` or `_`, those become wildcards and the match over-selects. The earlier review asked for wildcard escaping in LIKE patterns; this new branch appears to reintroduce an unescaped `LIKE` against a value derived from the request. Even though `kind` values are conventionally simple (`skill`, `capability`, ...), escaping `%`/`_` (with a declared `ESCAPE` char) would make this robust and consistent with the fix that was already applied elsewhere. Worth confirming whether `kind`/`name` are constrained to a safe character set upstream, or escaping here too.
- **`1 = 1` fallthrough for blank structure value (line 177):** an empty/blank structure filter short-circuits to `1 = 1`, i.e. the filter matches *every* artifact. For a negated filter (`isNot()`), this means `NOT EXISTS(... 1=1)` would match nothing. Worth confirming that an empty structure filter is actually reachable from the API layer — if `WellKnownResourceImpl` always supplies a non-blank value this is dead code, but if a caller can send `structure=` (blank) the current behavior (match-all / match-none on negation) may be surprising.
- **Backfill scale:** `StructuredContentUpgrader` streams with `setFetchSize(50)` and logs every 500 artifacts, which is reasonable. Just flagging that it re-reads and re-parses the full content blob of the latest version of every matching artifact at upgrade time; on a registry with many agent/MCP artifacts this is a non-trivial one-time cost and runs inline in the DB-upgrade transaction path. The progress logging helps, but if upgrades are expected to be fast/atomic you may want to document the expected duration or offer a batched/async variant.
- Minor: the new `StringUtil.asLowerCase` switch to `Locale.ROOT` is a good correctness fix (avoids Turkish-`i` style surprises) and is reused consistently by both the writer and the search reader, so storage and query stay symmetric.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java
```diff
@@ -0,0 +1,129 @@
...
+
+        int totalCount = 0;
+        AtomicInteger examined = new AtomicInteger();
+        for (ArtifactTypeUtilProvider provider : factory.getAllArtifactTypeProviders()) {
+            StructuredContentExtractor extractor = provider.getStructuredContentExtractor();
+            if (extractor == null) {
+                continue;
+            }
+            totalCount += handle.createQuery(sql).bind(0, provider.getArtifactType()).setFetchSize(50)
+                    .map(new ArtifactContentRowMapper()).stream().mapToInt(artifact -> {
```
paoloantinori: The backfill re-reads and re-parses the full content blob of the latest version of every matching artifact, inline in the DB-upgrade path. Fine for small registries; on large ones this is a non-trivial one-time cost. The progress logging helps — worth a note in upgrade docs about expected duration, or a batched variant if upgrades need to stay fast.

parthdagia05:
Thanks @paoloantinori  pushed the wildcard escaping for the 2-part kind match. The blank structure case isn't reachable from either API path (both guarantee a non-blank value), so it stays a defensive no-op.

paoloantinori:
Thanks — the wildcard escaping for the 2-part `kind` match resolves the SQL-injection-via-LIKE point, and your reasoning that the blank-structure case is unreachable from both API paths (each guarantees a non-blank value) is consistent with what I see in the call sites, so leaving it as a defensive no-op is fine. The documented import/gitops/kafkaops gap and the once-only warning log are the right pragmatic call for now. I have nothing further to add from my side.

parthdagia05:
@paoloantinori as discussed in #8627 , i would be bumping to 109 once the PR gets merged

paoloantinori:
Got it, thanks for the heads-up. Bumping to 109 once this merges sounds right given #8627 is also in flight — avoids the two PRs stepping on each other's db-version. I'll keep an eye out for the version bump when you push it.

parthdagia05:
@paoloantinori This branch is still under review. I was thinking of opening a PR for another issue, but I remember there was a guideline about having only one open PR at a time. Should I wait for this one to be merged first, or is it okay to open another draft PR?

paoloantinori:
On the one-open-PR question: opening a draft for a **different** accepted issue is generally fine. The guideline is about keeping review focus manageable, and a draft targeting a separate problem doesn't fragment attention the way two competing implementations of the same fix would. The main thing — same as with this PR — is to get maintainer `/accept` on the underlying issue first, so you're not investing effort in something that gets redirected on architectural grounds.

One caveat specific to this PR: since #8538 bumps the DB version (107 → 108, with #8627 tracking a follow-up move to 109), if your next PR also touches the SQL schema or migrations, watch the version numbering so the two don't collide. If it's an unrelated area, no concern at all.

parthdagia05:
> On the one-open-PR question: opening a draft for a **different** accepted issue is generally fine. The guideline is about keeping review focus manageable, and a draft targeting a separate problem doesn't fragment attention the way two competing implementations of the same fix would. The main thing — same as with this PR — is to get maintainer `/accept` on the underlying issue first, so you're not investing effort in something that gets redirected on architectural grounds.
> 
> One caveat specific to this PR: since #8538 bumps the DB version (107 → 108, with #8627 tracking a follow-up move to 109), if your next PR also touches the SQL schema or migrations, watch the version numbering so the two don't collide. If it's an unrelated area, no concern at all.

got it

parthdagia05:
@paoloantinori  a quick note on the db-version bump.

It looks like #8627 hasn't seen much activity recently and it's still carrying failing checks, so I don't think it's worth blocking this PR on merge ordering. If this one is ready first, we can go ahead with the 107 → 108 bump here, and whoever picks up #8627 later can simply rebase and take the next version

paoloantinori:
Agreed — if #8627 is stalled with failing checks, blocking this PR on merge ordering doesn't help anyone. Going 107 → 108 here is the pragmatic call; whoever revives #8627 rebases and takes 109. The `StructuredContentUpgrader` is keyed to the db-version in the upgrade DDL rather than hardcoding 108, so a later renumbering on #8627's side should be a clean swap of the version constant + dir rename, nothing structural to redo here.

Just one thing to keep an eye on at merge time: make sure the 108 upgrade scripts (and the new `artifact_structured_content` table) land in the same release as this PR — if the code that writes to the table ships before the schema upgrade is available on an existing deployment, the write paths in `AbstractSqlRegistryStorage` will fail until the upgrader runs. That's already the case by construction (table + upgrader + code in one PR), so just flagging it for the merge sequence.

No further changes needed from my side — looks good to go once CI is green.

parthdagia05:
@EricWittmann @carlesarnal  review is complete here (@paoloantinori signed off above), could one of you /accept it so it can move to WIP and get a CI run?

paoloantinori:
Confirming I've seen this — my earlier review stands as resolved, and I've nothing further to add on the code. The `/accept` to move it into WIP and get a full CI run is a maintainer action, so over to @EricWittmann / @carlesarnal as you flagged.

EricWittmann:
@parthdagia05 Sorry but we have some DB upgrade script contention with other work, so there are some conflicts that need to be resolved on this PR.

parthdagia05:
@EricWittmann  rebased and renumbered my migration to 110,
Could you approve the pending workflow runs?

paoloantinori (commented):
Overall this is a well-structured, schema-consistent implementation: it mirrors the proven `artifact_labels` pattern, covers the write paths, normalizes case the same way as the ES backend, and escapes LIKE wildcards correctly. The notable risks are a latest-version invariant that one write path can violate, a negated-filter scope question, and a documented-but-significant import gap for the gitops/kubernetesops variants.

- `app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java:1112` — `updateArtifactVersionContent` re-indexes structured content from the *passed-in version's* content unconditionally. The index is documented as reflecting the latest version, so if this method is ever called for a non-latest version, it overwrites the index with older content (latest agent card becomes unsearchable, old content becomes searchable) until the latest version is rewritten.
- `app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java:183` — the negated structure filter compiles to `NOT EXISTS`, which is true for any artifact with no matching structured row, including artifacts with zero `artifact_structured_content` rows. Correct within the scoped agents endpoint, but if structure filters ever reach the general artifact search, a negated filter returns every non-AGENT_CARD/MCP artifact.
- `app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java` (`warnOnStructuredContentImportGap`) — import does not populate the new table, so the gitops and kubernetesops variants (which load exclusively via import) are never searchable by structure filters unless artifacts are re-uploaded; the `StructuredContentUpgrader` backfill only runs on a real DB version upgrade. The docs call this out, but it is a meaningful functional gap worth tracking as a follow-up rather than relying on a once-per-JVM warning.
- `app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java:37` — the upgrader constructs `new DefaultArtifactTypeUtilProviderImpl(true)` instead of reusing the injected production factory, so any dynamically- or extension-registered artifact type would be skipped during backfill. Likely required by the reflective `UPGRADER:` instantiation model, but the `true` flag's meaning and extension-type coverage are worth confirming.
- Test plan — only H2 is verified (24/24); postgresql/mysql/mssql/kafkasql CI are still unchecked. The DDL/FK follows the established `artifact_labels` pattern so dialect risk is low, but the EXISTS/LIKE behavior and the in-upgrade backfill transaction should be confirmed on the other dialects before merge.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -1046,7 +1104,14 @@ public void updateArtifactVersionContent(String groupId, String artifactId, Stri
         // Put the new content in the DB and get the unique content ID back.
         long contentId = ensureContentAndGetId(artifactType, content, true);
 
-        versionRepository.updateArtifactVersionContent(groupId, artifactId, version, contentId);
+        // Update the version content and its structured-content index in a single transaction so the
+        // index can never be left pointing at a previous version's content.
+        handles.withHandleNoException(handle -> {
+            versionRepository.updateArtifactVersionContentRaw(handle, groupId, artifactId, version,
+                    contentId);
+            updateStructuredContentRaw(handle, groupId, artifactId, artifactType, content);
```
paoloantinori: `updateStructuredContentRaw` here re-indexes from `content` — the content of the version being updated — but the index is meant to reflect the *latest* version. If `updateArtifactVersionContent` can target a non-latest version (the method takes an arbitrary `version`), this call overwrites the artifact-level index with that older version's structured content, making the latest agent card unsearchable and the old content searchable until the latest version is rewritten. Consider re-extracting from the current latest version's content, or guarding this re-index on `version == latest`. Worth confirming whether the REST/import paths can actually update a non-latest version's content; if not, a short comment asserting that precondition would be enough.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -167,6 +170,50 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
...
+                            break;
+                        }
+                        String[] structureParts = structureValue.trim().split(":", 3);
+                        op = filter.isNot() ? "NOT EXISTS" : "EXISTS";
```
paoloantinori, line 210: A negated structure filter compiles to `NOT EXISTS (...)`, which is TRUE for any artifact that has no matching row — including artifacts with no `artifact_structured_content` rows at all (every non-AGENT_CARD/MCP artifact, plus cards whose extractor returned nothing). That is correct inside the scoped `/.well-known/agents/search` endpoint, but this repository is the general artifact search. If `SearchFilter.ofStructure()` is reachable from the general search path, a negated structure filter would return all unrelated artifacts. Confirm structure filters cannot reach general search, or special-case the "artifact has zero structured rows" case for negation.

parthdagia05:
@paoloantinori  Thanks for the review. Working through the rest; two of these I don't think need a code change, with the reasoning:

**Negated structure filter reaching general search.** Not reachable today. `case structure` only exists in `searchArtifacts`; `searchVersions` has no such case. And `negated()` is only ever produced in `WellKnownResourceImpl` for a false-valued capability, and that path always adds `ofArtifactType(AGENT_CARD)` before any structure filter, so artifacts with zero structured rows can't match. "Capability absent means matches `capability: false`" is the intended semantic for the agents endpoint. I've added a comment recording that precondition so it isn't lost if structure filters get exposed more widely later.

**`new DefaultArtifactTypeUtilProviderImpl(true)`.** `IDbUpgrader` instances are constructed reflectively outside CDI, so the injected factory isn't available there; `AvroCanonicalHashUpgrader` uses the same pattern for the same reason. The `true` flag loads the standard providers and is required, not optional: the no-arg constructor leaves the provider list empty, which would make the backfill a silent no-op rather than just skipping extension types.

One adjacent thing that predates this PR: `SearchResourceImpl` adds a structure filter and then calls `searchVersions`, which throws `RegistryStorageException` on SQL, so `/search/versions?structure=...` returns a 500. Same shape as the fixes in #8666 and #8697. Happy to handle it here or file it separately, whichever you prefer.

paoloantinori:
@parthdagia05 Thanks — the reasoning on all three holds up.

**Negated structure filter.** Agreed it's not reachable today. Confirmed against the code: `case structure` lives only in `searchArtifacts` (`searchVersions` has no such case), and the only producer of `negated()` is the `WellKnownResourceImpl` capability path, which always adds `ofArtifactType(AGENT_CARD)` before any structure filter. Since AGENT_CARD artifacts created via the API always get structured-content rows, a negated NOT EXISTS can't accidentally match a zero-row artifact. The comment you added documenting that precondition is the right safeguard for when structure filters get exposed more broadly. Good call.

**`new DefaultArtifactTypeUtilProviderImpl(true)`.** Fair — the reflective construction outside CDI makes the injected factory unavailable, and the `true` flag is required (no-arg leaves providers empty → silent no-op backfill). The `AvroCanonicalHashUpgrader` precedent confirms the pattern. No change needed.

**`/search/versions?structure=...` 500.** File it separately. It's the same class of bug as #8666/#8697 and keeping this PR focused on the agent/MCP discovery path is better than widening its scope into the general search endpoints. The fix there is its own review (add a `case structure:` to `searchVersions` or throw a clear unsupported-exception instead of a bare `RegistryStorageException`).

My earlier sign-off stands; nothing further from my side on the code.

parthdagia05:
@EricWittmann can you take a look

EricWittmann (commented):
Nice work on this PR — the structured content search implementation is solid and the iterative responses to review feedback have been good. A few items to address before we can merge:

---

**1. `StructuredContentUpgrader` doesn't filter out `NoopStructuredContentExtractor`**

In `AbstractSqlRegistryStorage.structuredContentExtractorFor()`, you correctly detect `NoopStructuredContentExtractor` and return null to skip types that don't produce structured content. But `StructuredContentUpgrader` uses `DefaultArtifactTypeUtilProviderImpl` directly and only checks `extractor == null` — which will never be true since `getStructuredContentExtractor()` never returns null. This means the upgrader will read and attempt extraction on every artifact type (AVRO, Protobuf, JSON Schema, etc.), paying the cost of loading content blobs and running a no-op extract for each one. On a large registry this adds unnecessary time to an already-slow backfill.

Fix: add an `instanceof NoopStructuredContentExtractor` check in the upgrader's loop, consistent with `structuredContentExtractorFor()`.

---

**2. Blank structure value falls through to `1 = 1`**

In `SqlSearchRepository`, when `structureValue` is null or blank, the code appends `1 = 1` — which for a positive filter means "match everything" and for a negated filter (`NOT EXISTS(... WHERE 1 = 1)`) means "match nothing." These semantics are surprising and were noted in an earlier review round.

Can you confirm whether a blank `structure` value is actually reachable from the API? If it's not reachable, consider throwing an `IllegalArgumentException` or `RegistryStorageException` instead of silently degrading. If it is reachable, the match-all / match-none behavior needs to be intentional and documented.

---

**3. Negated structure filter scope on general artifact search**

The `case structure:` handler lives in `SqlSearchRepository`, which serves both the scoped `/.well-known/agents/search` endpoint and the general artifact search API. A negated structure filter compiles to `NOT EXISTS (...)`, which is TRUE for any artifact that has no `artifact_structured_content` rows — including every non-AGENT_CARD/MCP artifact. The code comments note this relies on the discovery endpoints always pairing a negated filter with an `artifactType` filter, but there's no enforcement of that invariant.

If `SearchFilter.ofStructure()` with negation is ever used from the general search path without an `artifactType` constraint, it would return false positives for every non-agent artifact. Consider one of:
- Adding a defensive check/assertion in the `case structure:` handler that verifies the filter set includes an `artifactType` filter when negation is used.
- Documenting this as a precondition on `SearchFilter.ofStructure()` with a code comment at the filter construction site.
- Moving the structure filter handling to the discovery-specific search path if the general search should never use it.

parthdagia05:
@EricWittmann can you take a look

parthdagia05:
Branch was still behind so merged the commits again

EricWittmann:
@parthdagia05 I think we might need to consider pushing this change out to 3.4 instead of adding it as a patch release in 3.3.  But I'm not sure.  I think we need to park this until I can discuss it with @carlesarnal - can you switch it to Draft status so we can sit on it without blocking you from contributing another PR?

parthdagia05:
> @parthdagia05 I think we might need to consider pushing this change out to 3.4 instead of adding it as a patch release in 3.3. But I'm not sure. I think we need to park this until I can discuss it with @carlesarnal - can you switch it to Draft status so we can sit on it without blocking you from contributing another PR?

sure

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/110/postgresql.upgrade.ddl
```diff
@@ -0,0 +1,13 @@
+-- *********************************************************************
+-- DDL for the Apicurio Registry - Database: postgresql
+-- Upgrade Script from 109 to 110
+-- *********************************************************************
+
+UPDATE apicurio SET propValue = 110 WHERE propName = 'db_version';
+
+CREATE TABLE artifact_structured_content (groupId VARCHAR(512) NOT NULL, artifactId VARCHAR(512) NOT NULL, elementType VARCHAR(64) NOT NULL, elementValue VARCHAR(512) NOT NULL);
```
carlesarnal: The artifact_structured_content table is missing a PRIMARY KEY constraint. Without it, we could end up with duplicate rows for the same (groupId, artifactId, elementType, elementValue) combination. The upgrade script creates indexes but doesn't enforce uniqueness at the database level.

Consider adding a composite PRIMARY KEY:
```sql
ALTER TABLE artifact_structured_content ADD PRIMARY KEY (groupId, artifactId, elementType, elementValue);
```

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -576,6 +590,61 @@ public Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createArtifact(Stri
...
+                String elementType = limitStr(asLowerCase(artifactType + ":" + element.kind()), 64);
+                String elementValue = limitStr(asLowerCase(element.name()), 512);
+                if (seen.add(elementType + ":" + elementValue)) {
+                    handle.createUpdate(sqlStatements.insertArtifactStructuredContent())
+                            .bind(0, normalizeGroupId(groupId)).bind(1, artifactId).bind(2, elementType)
+                            .bind(3, elementValue).execute();
+                }
+            }
+        } catch (Exception e) {
+            log.warn("Failed to update structured content for {}/{}: {}", groupId, artifactId,
```
carlesarnal: The error logging here includes user-controlled groupId and artifactId values without sanitization. An attacker could inject newlines or control characters to forge log entries or hide malicious activity.

We should sanitize these parameters before logging:
```java
log.warn("Failed to update structured content for {}/{}: {}", 
    sanitizeForLog(groupId), sanitizeForLog(artifactId), e.getMessage(), e);
```

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -576,6 +590,61 @@ public Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createArtifact(Stri
...
+            Set<String> seen = new HashSet<>();
+            for (StructuredElement element : elements) {
+                if (element == null || element.kind() == null || element.name() == null) {
+                    // Skip malformed elements. elementType/elementValue are NOT NULL, and on some
+                    // databases (e.g. PostgreSQL) a failed statement aborts the surrounding
+                    // transaction - which here is the artifact-write transaction. Silently dropping
+                    // bad elements keeps this best-effort update from failing the whole operation.
+                    continue;
+                }
+                String elementType = limitStr(asLowerCase(artifactType + ":" + element.kind()), 64);
```
carlesarnal: The value 64 (max length for elementType) is hardcoded here and appears in multiple places. It would be better to extract this to a named constant for maintainability.

```java
private static final int MAX_ELEMENT_TYPE_LENGTH = 64;
private static final int MAX_ELEMENT_VALUE_LENGTH = 512;
```

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -576,6 +590,61 @@ public Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createArtifact(Stri
...
+                String elementValue = limitStr(asLowerCase(element.name()), 512);
```
carlesarnal: Same as the line above - the value 512 (max length for elementValue) should be extracted to a named constant alongside MAX_ELEMENT_TYPE_LENGTH.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -576,6 +590,61 @@ public Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createArtifact(Stri
...
+                if (seen.add(elementType + ":" + elementValue)) {
+                    handle.createUpdate(sqlStatements.insertArtifactStructuredContent())
+                            .bind(0, normalizeGroupId(groupId)).bind(1, artifactId).bind(2, elementType)
+                            .bind(3, elementValue).execute();
+                }
+            }
+        } catch (Exception e) {
```
carlesarnal, line 649: When structured content extraction fails, we log the error but the artifact creation/update still succeeds. This means the artifact exists but isn't discoverable via structure filters, creating a silent data inconsistency.

We should consider adding metrics or a warning flag to track these extraction failures so operators have visibility into how many artifacts failed to index.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -1041,7 +1112,32 @@ public void updateArtifactVersionContent(String groupId, String artifactId, Stri
...
+
+    /**
+     * Returns true when the given version is the artifact's latest version (highest versionOrder). Used to
+     * keep the artifact-scoped structured-content index aligned with the latest version's content.
+     */
+    private boolean isLatestArtifactVersionRaw(Handle handle, String groupId, String artifactId,
+            String version) {
+        String normalizedGroupId = normalizeGroupId(groupId);
+        return handle.createQuery(sqlStatements.selectIsLatestArtifactVersion())
+                .bind(0, normalizedGroupId).bind(1, artifactId).bind(2, version)
```
carlesarnal, line 1202: The isLatestArtifactVersionRaw() method executes a separate query to check if the version is latest. This adds overhead to every updateArtifactVersionContent() call, even when the version is clearly not latest (e.g., updating a DRAFT v1 when v2 exists).

Consider caching version order information or passing it as a parameter to avoid the extra query.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -1558,13 +1654,41 @@ public void importContractRule(ContractRuleEntity entity) {
...
+     */
+    private void warnOnStructuredContentImportGap(String artifactType) {
+        if (artifactType == null) {
+            return;
+        }
+        try {
+            boolean hasExtractor = structuredContentExtractorFor(artifactType) != null;
+            if (hasExtractor && STRUCTURED_IMPORT_WARNING_LOGGED.compareAndSet(false, true)) {
+                log.warn("Imported artifacts are not indexed into artifact_structured_content; "
+                        + "structure-based discovery filters will not match imported artifacts until "
```
carlesarnal: The STRUCTURED_IMPORT_WARNING_LOGGED flag ensures this warning is logged only once per application lifetime. In long-running applications with infrequent imports, this warning might be logged during startup and then forgotten, even though imports continue to have the indexing gap.

Consider logging at INFO level periodically or adding a metric to track imported artifacts without indexing.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -167,6 +170,56 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
...
+                        // Negation (NOT EXISTS) is only produced by the agent discovery endpoints, which
+                        // always pair it with an artifactType filter. Without that pairing a negated
+                        // filter would also match artifacts that have no structured rows at all.
+                        String structureValue = asLowerCase(filter.getStringValue());
+                        if (structureValue == null || structureValue.isBlank()) {
+                            // Not reachable from the REST layer: the version search endpoint skips an
+                            // empty "structure" parameter, and the discovery endpoints always prefix the
+                            // value with "<artifactType>:<kind>:". Fail loudly rather than degrading to a
+                            // clause that silently matches every artifact.
+                            throw new RegistryStorageException("Structure filter value must not be blank.");
```
carlesarnal: The structure filter value is converted to lowercase before validation. While the blank check exists, we should validate the raw input first to ensure proper error messages and prevent potential edge cases.

carlesarnal: The error message 'Structure filter value must not be blank' doesn't explain what a valid structure filter looks like or provide examples.

Consider enhancing it with format and examples:
```java
throw new RegistryStorageException(
    "Structure filter value must not be blank. Expected format: " +
    "'artifactType:kind:name', 'kind:name', or 'name'. " +
    "Example: 'agent_card:skill:translation'");
```

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -37,6 +38,8 @@ public class SqlSearchRepository {
             "Content search requires the search index, which is not enabled. "
             + "Enable the search index to use content search.";
 
+    private static final char LIKE_ESCAPE_CHAR = '!';
```
carlesarnal, line 49: The LIKE_ESCAPE_CHAR constant is used to escape wildcards, but there's no documentation about why '!' was chosen or tests verifying it doesn't conflict with actual data values.

Consider adding documentation:
```java
// Using '!' as escape character - chosen because it's rare in artifact names
// and not a SQL wildcard. If artifact names commonly contain '!', this should
// be changed to a different character.
private static final char LIKE_ESCAPE_CHAR = '!';
```

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java
```diff
@@ -0,0 +1,134 @@
...
+            if (extractor == null || extractor instanceof NoopStructuredContentExtractor) {
+                // Providers never return null here - types without structured extraction get
+                // NoopStructuredContentExtractor, so that is what identifies them. Checking for null
+                // alone would read and extract every artifact of every type (AVRO, PROTOBUF, ...) only
+                // to discard an always-empty element list, adding avoidable time to the backfill.
+                continue;
+            }
+            totalCount += handle.createQuery(sql).bind(0, provider.getArtifactType()).setFetchSize(50)
+                    .map(new ArtifactContentRowMapper()).stream().mapToInt(artifact -> {
+                        if (examined.incrementAndGet() % 500 == 0) {
```
carlesarnal: The progress logging threshold of 500 artifacts is hardcoded. This should be a named constant to make it easier to adjust.

```java
private static final int PROGRESS_LOG_INTERVAL = 500;
```

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java
```diff
@@ -0,0 +1,134 @@
...
+                + "FROM artifacts a "
+                    + "JOIN versions v ON v.groupId = a.groupId AND v.artifactId = a.artifactId "
+                    + "JOIN content c ON c.contentId = v.contentId "
+                + "WHERE a.type = ? AND v.versionOrder = "
+                    + "(SELECT MAX(v2.versionOrder) FROM versions v2 "
+                        + "WHERE v2.groupId = a.groupId AND v2.artifactId = a.artifactId)";
+
+        int totalCount = 0;
+        AtomicInteger examined = new AtomicInteger();
+        for (ArtifactTypeUtilProvider provider : factory.getAllArtifactTypeProviders()) {
```
carlesarnal, line 78: The fetch size of 50 is hardcoded. This should be a named constant to make performance tuning easier.

```java
private static final int BACKFILL_FETCH_SIZE = 50;
```

carlesarnal, line 78: A fetch size of 50 means the backfill will make many round trips to the database. For a bulk operation like this, a larger fetch size (e.g., 500-1000) would improve performance.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+            totalCount += handle.createQuery(sql).bind(0, provider.getArtifactType()).setFetchSize(50)
+                    .map(new ArtifactContentRowMapper()).stream().mapToInt(artifact -> {
+                        if (examined.incrementAndGet() % 500 == 0) {
+                            log.info("Backfilling structured content: {} artifacts examined so far...",
+                                    examined.get());
+                        }
+                        try {
+                            return backfillArtifact(handle, artifact, extractor);
+                        } catch (Exception ex) {
```
carlesarnal, line 96: When backfillArtifact() fails for an artifact, it's logged and the upgrader continues. However, there's no summary at the end showing how many artifacts failed, making it hard to assess the health of the backfill operation.

Consider tracking and reporting failures in the final log message.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -575,6 +583,42 @@ public Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createArtifact(Stri
...
+                if (seen.add(elementType + ":" + elementValue)) {
+                    handle.createUpdate(sqlStatements.insertArtifactStructuredContent())
+                            .bind(0, normalizeGroupId(groupId)).bind(1, artifactId).bind(2, elementType)
+                            .bind(3, elementValue).execute();
+                }
+            }
+        } catch (Exception e) {
+            log.warn("Failed to update structured content for {}/{}: {}", groupId, artifactId,
+                    e.getMessage());
+        }
```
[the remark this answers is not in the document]
parthdagia05: Fixed in 06a274a40. The warning now passes the exception itself as the trailing argument, so the stack trace is preserved:

```java
log.warn("Failed to update structured content for {}/{}. The artifact was written, but it will "
        + "not match structure-based search filters until it is written again.",
        sanitizeForLog(groupId), sanitizeForLog(artifactId), e);
```

The failure also increments a `storage.structured.content.index.failures` counter now, so the resulting search gap is visible without grepping logs.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
parthdagia05: The 2-part branch does escape it - the value is bound as `"%:" + escapeLikePattern(structureParts[0])` with an explicit `ESCAPE '!'` clause, and `escapeLikePattern` escapes `%`, `_` and the escape character itself. Only the leading `%:` is an intentional wildcard.

What was missing was the reasoning and a test, both added in 06a274a40: `LIKE_ESCAPE_CHAR` now carries a javadoc explaining why `!` is safe (it is not a LIKE metacharacter, every supported database accepts it in an ESCAPE clause, and data containing `!` still matches literally because the escaper escapes it), plus `testSearchAgentsBySkillTreatsLikeWildcardsLiterally` creates an agent whose skill id is `report_2024%draft` and asserts the search does not also return a decoy agent that a wildcard interpretation would match.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
parthdagia05: Addressed in 06a274a40. Validation now runs on the raw value before it is lower-cased and trimmed, so a whitespace-only value is reported as blank instead of being normalized into an empty filter, and the message explains the expected formats.

On reachability: the discovery endpoints always build `agent_card:<kind>:<name>`, so blank cannot come from there. The general search endpoint (`SearchResourceImpl`) passes its `structure` query parameter through, but only to `searchVersions`, which rejects structure filters as an unsupported filter type before this code runs. So a blank value is not reachable today - the throw is a guard against a future caller, which is why it fails loudly rather than degrading to a match-all clause.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java
parthdagia05: Addressed in 06a274a40:

- Fetch size raised from 50 to 500 (named `BACKFILL_FETCH_SIZE`), which cuts the round trips for the bulk read considerably.
- Failures are counted and reported in the final log line, so a partially successful backfill is distinguishable from a clean one.
- Documented the cost in the upgrader javadoc and in the agent discovery docs: the backfill reads and re-parses the latest version of every agent card / MCP tool / OpenAPI / AsyncAPI artifact - other types are never read - progress is logged every 500 artifacts, and operators should plan for a longer startup on the first instance that applies the upgrade.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
parthdagia05: This was addressed in a later commit and the thread is now outdated - `updateArtifactVersionContent` guards the re-index with `isLatestArtifactVersionRaw(...)`, so a non-latest DRAFT no longer overwrites the artifact-level index.

06a274a40 refines it further: the extractor lookup runs first, so artifact types that never produce structured content skip the latest-version query entirely and only agent card / MCP tool / OpenAPI / AsyncAPI writes pay for it.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
parthdagia05: Confirmed unreachable, and the comment now records why. `SearchFilter.ofStructure()` has exactly two callers: `WellKnownResourceImpl`, which is the only one that negates the filter (`capabilities: {x: false}`) and always pairs it with an artifactType filter; and `SearchResourceImpl`, which passes the `structure` query parameter through un-negated - and only to `searchVersions`, which rejects structure filters as an unsupported filter type anyway. `SearchFilter.ofStructure()` documents the pairing precondition for future callers.

I did not special-case "artifact has zero structured rows" for negation on purpose: inside the discovery endpoint that would be wrong. An agent card that declares no capabilities at all genuinely does not support push notifications, so it should match `pushNotifications: false`; excluding it would break that query.

Separately, structure filters reaching `searchVersions` and failing with "Filter type not supported" is pre-existing behaviour on SQL storage (that endpoint only works with the Elasticsearch backend). Happy to open a follow-up for version-level structure search if you want it covered.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/110/postgresql.upgrade.ddl
parthdagia05: Added in 06a274a40, with one adjustment I would like your read on.

The primary key is now on all four columns in the fresh-install DDL and the 110 upgrade script for every dialect:

```sql
ALTER TABLE artifact_structured_content ADD PRIMARY KEY (groupId, artifactId, elementType, elementValue);
```

To make it portable I narrowed `elementValue` from 512 to 256. On MySQL that column is the utf8mb4 one (4 bytes per character) while `groupId`/`artifactId`/`elementType` are ascii, so at 512 the key would be 512 + 512 + 64 + 2048 = 3136 bytes - over InnoDB's 3072-byte limit, and `CREATE TABLE` would fail outright. At 256 it is 2112 bytes. On SQL Server the key is 2688 bytes, which exceeds the 900-byte clustered limit but is accepted with a warning for variable-length columns, exactly like the existing `artifact_labels` primary key (2560 bytes). 256 characters is well beyond any real skill id, capability name, MIME type or tool parameter name, and values were already truncated before.

I also dropped `IDX_asc_2 (groupId, artifactId)`: it is now a strict prefix of the primary key, so it only added write cost - the `EXISTS` subquery, the delete-by-artifact and the MySQL foreign key all use the primary key instead.

If you would rather keep `elementValue` at 512, the alternatives are a unique constraint over a hash column, or accepting that MySQL cannot carry the constraint. Happy to switch.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
parthdagia05: Fixed in 06a274a40. Added `StringUtil.sanitizeForLog(String)` - it replaces ISO control characters (CR/LF above all, so a value cannot forge extra log entries, and terminal escapes cannot be smuggled through) and truncates to 256 characters so an oversized value cannot flood the log. It is applied to `groupId`, `artifactId` and `artifactType` everywhere they are logged on this path, including the backfill upgrader. The exception is passed through as the trailing argument so the stack trace is kept.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
parthdagia05: Done in 06a274a40: `StructuredContentIndexUtils.MAX_ELEMENT_TYPE_LENGTH = 64`, documented as having to match `VARCHAR(64)` in the DDLs.

The constants live in a new small `StructuredContentIndexUtils` class rather than in this file, because the write path and the upgrade backfill are two independent writers of the same table and must normalize identically - otherwise an agent card would be searchable after an upload but not after a backfill. Both now call the same `elementType(...)` / `elementValue(...)` helpers.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
parthdagia05: Done in 06a274a40: `StructuredContentIndexUtils.MAX_ELEMENT_VALUE_LENGTH`, alongside `MAX_ELEMENT_TYPE_LENGTH`.

Note the value is now 256 rather than 512 - the four columns became the table's primary key per your other comment, and 512 utf8mb4 characters would push the MySQL key past InnoDB's 3072-byte limit. The reasoning is in the constant's javadoc and in my reply on the DDL thread.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
parthdagia05: Added in 06a274a40. Two counters, registered in `MetricsConstants`:

- `storage.structured.content.index.failures` - incremented whenever extraction or indexing fails on a write. The artifact write still succeeds, so this is exactly the silent inconsistency you describe, now countable.
- `storage.structured.content.import.skipped` - incremented per imported artifact of a type that produces structured content, since import does not populate the index at all.

The failure log line also spells out the consequence now ("the artifact was written, but it will not match structure-based search filters until it is written again") instead of only reporting the error.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
parthdagia05: Addressed in 06a274a40 by not running the query in the common case rather than by caching:

```java
if (hasStructuredContentExtractor(artifactType)
        && isLatestArtifactVersionRaw(handle, groupId, artifactId, version)) {
```

The extractor check is an in-memory provider lookup, so AVRO, PROTOBUF, JSON Schema and every other non-extractable type - the overwhelming majority of writes - now skip the latest-version query entirely. Only agent card, MCP tool, OpenAPI and AsyncAPI writes pay for it, and for those the query is a single indexed lookup against `versions`.

I left threading version order through as a parameter alone: `updateArtifactVersionContent` is on the `RegistryStorage` interface, so it would mean changing the signature for every storage variant to serve one implementation detail.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
parthdagia05: Fixed in 06a274a40. The `AtomicBoolean` is now an `AtomicLong` count of skipped artifacts, and the warning is emitted on the first one and then every 1000 - so a long-running instance that keeps importing keeps reporting the growing backlog instead of mentioning it once during startup and going quiet. The log line reports the running total.

Every skipped artifact also increments the `storage.structured.content.import.skipped` counter, so the size of the gap is visible at any point in time rather than only at the moments a warning happens to fire.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
parthdagia05: Fixed in 06a274a40 - the blank check now runs on `filter.getStringValue()` before any normalization:

```java
String rawStructureValue = filter.getStringValue();
if (rawStructureValue == null || rawStructureValue.isBlank()) {
    throw new RegistryStorageException("Structure filter value must not be blank. " + STRUCTURE_FILTER_FORMAT_HELP);
}
String[] structureParts = asLowerCase(rawStructureValue.trim()).split(":", 3);
```

So a whitespace-only value is reported as blank instead of being lower-cased and trimmed into an empty filter.

parthdagia05: Done in 06a274a40. The message is now:

> Structure filter value must not be blank. Expected '<artifactType>:<kind>:<name>', '<kind>:<name>' or '<name>', for example 'agent_card:skill:translation'.

The format help is a constant (`STRUCTURE_FILTER_FORMAT_HELP`) so any future validation error on this filter can reuse the same wording.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
parthdagia05: Done in 06a274a40, both parts.

The constant now carries a javadoc explaining the choice: `!` is not a LIKE metacharacter, every supported database (H2, PostgreSQL, MySQL, SQL Server) accepts it in an ESCAPE clause, and - the part worth being explicit about - it needs no special handling in the data either, because `escapeLikePattern` escapes the escape character itself. A value that genuinely contains `!` still matches literally; the character is only ever consumed as an escape where this code puts it.

On the test: `testSearchAgentsBySkillTreatsLikeWildcardsLiterally` creates an agent whose skill id is `report_2024%draft` plus a decoy that a wildcard interpretation would match, and asserts the filter returns only the first.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java
parthdagia05: Done in 06a274a40 - `PROGRESS_LOG_INTERVAL = 500`, referenced from the class javadoc so the documented behaviour and the constant stay in sync.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java
parthdagia05: Done in 06a274a40 - `BACKFILL_FETCH_SIZE`, with a javadoc noting it is sized to trade round trips against how many content blobs are held in memory at once. Its value also went up to 500, per your other comment.

parthdagia05: Agreed - raised to 500 in 06a274a40 (`BACKFILL_FETCH_SIZE`). I stayed at the lower end of your 500-1000 range because the rows carry whole content blobs rather than narrow columns, so the fetch buffer is the memory cost of 500 artifact payloads; 500 already removes most of the round trips.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/StructuredContentUpgrader.java
parthdagia05: Added in 06a274a40. Failures are counted and the summary branches on them:

```java
if (failed.get() > 0) {
    log.warn("Backfilled structured content for {} artifacts ({} examined), but {} artifact(s) failed "
            + "and will not match structure-based search filters until they are written again. "
            + "See the warnings above for the affected artifacts.", totalCount, examined.get(), failed.get());
} else {
    log.info("Successfully backfilled structured content for {} artifacts ({} examined).", totalCount, examined.get());
}
```

So a partially successful backfill no longer looks identical to a clean one, and the log points at the per-artifact warnings for detail.

parthdagia05:
@EricWittmann @carlesarnal should this PR be still in draft ?

paoloantinori:
Pointer rather than a review, since this PR is parked on a release decision that is not mine to make.

This branch adds `upgrades/110` for all four dialects and moves `db-version` to 110. So do #9146, #9651, and #9789. Main is still on 109, so the number is unclaimed today, but only one of the four can keep it. I have written up the rule and the renumbering steps in #10141.

The part worth reading before you resolve the conflicts on this branch is that folding two migrations into a single `110` directory looks like a clean resolution and is not one. `databaseUpgrade` never re-runs a version a database has already applied, so the second set of statements would reach fresh installs only and skip every upgraded deployment without failing anything. Renumbering avoids that. Merging the directories does not.

Your question from 20 August about whether this should still be a draft has not been answered, and that is 26 days ago now. @EricWittmann raised the 3.3 against 3.4 question on 11 August, and 3.4.0 is due on 17 September. That call is his, so I am flagging the open question rather than answering it.

carlesarnal:
@parthdagia05 @EricWittmann Pushed maintainer follow-up 908d607b4d after merging current main. The stale endpoint test conflict is reconciled with current GET discovery.

Remaining review gaps addressed: import indexing, published-branch lifecycle selection, scoped negation/blank filters, transaction-safe SQL failure handling, and retryable MySQL/H2 upgrade sequencing. Added branch-order/import/rollback and database-specific regressions. The current A2A endpoint now supplies structure prefilters while retaining visibility checks.

93 tests passed across H2, PostgreSQL, MySQL, real KafkaSQL and GitOps/KubernetesOps import coverage; Checkstyle zero. SQL Server runtime is not yet verified locally. Description updated with exact scope and evidence. Please re-review; no merge performed.

PR #9563 overlap: retain its ES-specific improvements if needed, but do not reintroduce its SQL unsupported-structure fallback or legacy removed-endpoint tests over this implementation.

app/src/main/java/io/apicurio/registry/storage/impl/sql/StructuredContentIndexUtils.java
```diff
@@ -0,0 +1,85 @@
...
+    public static String elementType(String artifactType, String kind) {
+        return limitStr(asLowerCase(artifactType + ":" + kind), MAX_ELEMENT_TYPE_LENGTH);
+    }
+
+    /**
+     * Builds the {@code elementValue} value for a structured element, lower-cased for the same reason as
+     * {@link #elementType(String, String)}.
+     */
+    public static String elementValue(String name) {
+        return limitStr(asLowerCase(name), MAX_ELEMENT_VALUE_LENGTH);
```
carlesarnal: [P2] Do not truncate exact-match structured identifiers

`elementValue()` truncates every indexed name to 256 characters, but `SqlSearchRepository` binds the full search value unchanged. AgentCardContentValidator does not limit skill IDs to 256. Reproduced by registering an otherwise valid Agent Card with a 301-character skill ID, then querying `GET /.well-known/agents?skill=<that exact ID>`: creation succeeds but discovery returns count=0. The new prefilter prevents the existing parsed-content check from ever seeing the card. Simply truncating the query would also introduce collisions for IDs sharing their first 256 characters. Preserve exact identity (for example full values with a bounded digest key/index), or explicitly validate a supported bound before persisting rather than silently making valid content undiscoverable. Add full-length lookup and common-prefix-distinct-value regressions.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -170,6 +184,76 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
...
+                        String rawStructureValue = filter.getStringValue();
+                        if (rawStructureValue == null || rawStructureValue.isBlank()) {
+                            // Not reachable from the REST layer: the version search endpoint skips an
+                            // empty "structure" parameter, and the discovery endpoints always prefix the
+                            // value with "<artifactType>:<kind>:". Fail loudly rather than degrading to a
+                            // clause that silently matches every artifact.
+                            throw new IllegalArgumentException(
+                                    "Structure filter value must not be blank. "
+                                            + STRUCTURE_FILTER_FORMAT_HELP);
+                        }
```
carlesarnal, line 217: [P2] Map invalid structure components to a client error

Blank components are reachable through the current REST integration: `GET /.well-known/agents?skill=` becomes `agent_card:skill:` in addStructureFilters(). This throw escapes as IllegalArgumentException, which HttpStatusCodeMap does not map, yielding HTTP 500 (confirmed with a Quarkus HTTP regression; expected 400, actual 500). Validate public skill/capability/mode inputs before querying and return the registry's mapped InvalidParameterValueException/BadRequestException, or consistently map the validation failure at the API boundary. Cover empty and whitespace-only values through HTTP, not just direct storage assertThrows tests.

carlesarnal (changes requested):
Re-reviewed the current head, including the maintainer follow-up I pushed. The lifecycle/import/rollback fixes are an improvement, but two HTTP-level edge cases remain and prevent approval. I reproduced both on unchanged production code using targeted H2/Quarkus tests: a valid long skill is silently unfindable, and an empty skill query returns 500. Inline comments include exact scenarios. These were missed by the earlier 93-test suite, including my own fixes.

Separate verification limits: SQL Server runtime remains untested locally; migration 110 must be rechecked against main before merge per #10141. Existing metadata blockers (DCO/milestones) and review/workflow gating also remain. No approval or merge is implied by prior local green tests.

carlesarnal:
Addressed both findings from my re-review in c723106b60, pushed directly to the contributor branch:

- Indexed element values now use SHA-256 of the entire normalized value, consistently for live indexing, upgrade backfill and all query forms. Long IDs with identical first 256 characters remain distinct, and digest-shaped literal strings cannot alias long IDs.
- Discovery validates empty/whitespace skill/capability/mode inputs before storage and returns HTTP 400. Malformed capability suffixes also return 400.

Verification: 81 tests passed across H2/PostgreSQL REST and H2/PostgreSQL/MySQL/KafkaSQL storage, plus 6 GitOps/KubernetesOps import smoke tests. Long-value tests cover live/backfilled full/partial/plain queries, common-prefix collisions and digest-shaped literals. Checkstyle app/common: zero violations. Manual test-quality review: 9.8/10, no blockers in this diff.

No metadata/history repair or merge performed; older DCO/milestone blockers and SQL Server runtime verification remain separate.

PR #8542, opened 2026-07-12
fix(ui): persist Explore page sort-order preference in localStorage

### LINKED ISSUE #5221: [UI] Remember user's preferred state of Explore list view:  Groups vs.  Artifacts
The Explore view always defaults to showing a list of Artifacts.  It would be nice if we remembered the user's preferred setting.  Or perhaps the **last** setting.

anshulll19:
## What
Persists the user's sort-order preference (ascending/descending) on the Explore page across page navigation and browser reloads, using localStorage.

## Why
The Explore page always reset to its default sort order on every visit. This follows the same request as #5221.

## Note on scope
The original issue title mentions a "Groups vs. Artifacts" toggle — that toggle no longer exists in the current Explore page (it now only shows Groups). The closest remaining persistable preference is sort order, so this PR scopes to that. Happy to adjust if this doesn't match what was intended.

## Implementation
Follows the exact pattern already used in `ReferencesTabContent.tsx`:
- `useLocalStorageService` hook
- Module-level storage key constant
- Lazy `useState` initializer reading from storage
- Persist on change via `setConfigProperty`

`filterValue` (search term) is intentionally **not** persisted, matching the same convention as the References toolbar (ephemeral, not saved).

## Testing
- `tsc --noEmit` passes with no errors
- Manually verified: toggle sort → navigate away and back → sort order preserved
- Manually verified: reload the page → sort order preserved
- Confirmed correct key/value in DevTools → Application → Local Storage

Fixes #5221


EricWittmann:
Thanks @anshulll19 - the copilot review actually has some good feedback that should be addressed before merging.  Please take a look!

carlesarnal (changes requested):
The localStorage persistence works but needs error handling. If `getConfigProperty` throws during initialization, the page won't load. Wrap the localStorage call in try-catch and fall back to `true` on error.

The string-based boolean storage is fragile. Check if `LocalStorageService` supports native boolean types. If not, document why string conversion is necessary.

Consider moving `getInitialAscending` outside the component or memoizing it with `useMemo` to avoid recreating it on every render.

Add unit tests for the localStorage integration. At minimum: default behavior, persistence after sort change, error handling.

`ui/ui-app/src/app/pages/explore/ExplorePage.tsx`

ui/tests/specs/explore.spec.ts
```diff
@@ -7,6 +7,24 @@ const OPENAPI_DATA_V2_STR: string = JSON.stringify(OPENAPI_DATA_V2, null, 4);
...
 const REGISTRY_UI_URL: string = process.env["REGISTRY_UI_URL"] || "http://localhost:8888";
 
+test("Explore - Sort order persistence", async ({ page }) => {
+    await page.goto(`${REGISTRY_UI_URL}/explore`);
+    await expect(page).toHaveTitle(/Apicurio Registry/);
+
+    // 1. Click sort button to change it to descending (false)
+    await page.getByTestId("artifact-filter-sort").click();
+
+    // Verify localStorage has the value
```
paoloantinori: This hardcoded key (`apicurio-registry.config.explore.ascending`) duplicates the prefix logic that lives inside `LocalStorageService` and the `EXPLORE_SORT_KEY` constant in `ExplorePage.tsx`. The `aria-label` assertion after reload below already covers the persistence behavior end-to-end, so this intermediate check mostly adds coupling. Consider removing it, or exporting the full key from the component module so the test references a single source of truth.

paoloantinori (commented):
Clean implementation that follows the existing `ReferencesTabContent` pattern, and the try/catch around `getConfigProperty` addresses @carlesarnal's earlier request. One test-quality observation worth considering:

- The Playwright spec (`ui/tests/specs/explore.spec.ts`) reads localStorage with a hardcoded full key `"apicurio-registry.config.explore.ascending"`, which couples the test to two implementation details at once: the `LocalStorageService` prefix (`apicurio-registry.config.`) and the module constant `EXPLORE_SORT_KEY` (`"explore.ascending"`). If either ever changes, this assertion breaks even though the feature still works. The behavioral assertion in step 3 (the `aria-label` attribute after reload) already proves persistence end-to-end, so the intermediate `localStorage.getItem` check is largely redundant — dropping it (or reading through the service / an exported key) would make the test less brittle without losing coverage.

anshulll19:
Thanks for the review. I've removed the intermediate localStorage assertion and now rely on the reload/UI behavior assertion to verify persistence end-to-end.

paoloantinori:
Thanks for addressing that — relying on the reload + UI-behavior assertion to verify end-to-end persistence is the right call. The updated test looks clean. 👍

EricWittmann (commented):
Thanks for the PR @anshulll19 — the implementation follows the existing `ReferencesTabContent` pattern well, and the fixes you made for error handling, lazy initialization, and conditional writes all look good.

One bug to fix before this can merge:

**Toolbar icon doesn't sync with the restored sort preference.** `ExplorePageToolbar.tsx` line 62 initializes its own `filterAscending` state to `true`:

```typescript
const [filterAscending, setFilterAscending] = useState(true);
```

The toolbar receives `props.criteria` (which includes `ascending`) but doesn't use it. When localStorage restores `ascending=false` after a reload, the parent correctly queries data in descending order, but the toolbar icon still shows ascending. The first click toggles the icon to descending without actually changing the sort — a confusing dead-click.

Fix: initialize from props:

```typescript
const [filterAscending, setFilterAscending] = useState(props.criteria.ascending);
```

This was flagged by the Copilot review earlier and is the last remaining item. Note that your Playwright test's `aria-label` assertion after reload depends on this fix being in place — without it, the `aria-label` would still read `"sort ascending"` after reload even though the data is sorted descending.

Also minor: the test has two extra blank lines (between the click and the reload) left over from removing the localStorage assertion — please clean those up.

EricWittmann:
Thanks for the contribution @anshulll19! Nice work persisting the sort preference — clean implementation that follows the existing patterns well. Merged.

PR #8544, opened 2026-07-12
docs: update outdated version references and add kubernetesops storage

### LINKED ISSUE #8543: [Docs] Update outdated version references and add missing kubernetesops storage variant in README.md
Hi team,

I was going through the README to set up Apicurio Registry locally and noticed a few things that seem outdated:

1. **Missing storage type**: The README says there are "three possible values" for APICURIO_STORAGE_KIND, but I found a fourth one (kubernetesops) in the code at app/src/main/java/io/apicurio/registry/storage/impl/kubernetesops/. It's not mentioned anywhere in the README.

2. **Old version in examples**: Line 175 shows the JAR as apicurio-registry-app-3.0.0-SNAPSHOT-runner.jar but pom.xml says the current version is 3.3.1-SNAPSHOT. Same issue on line 228 where the docker-compose example uses image version 3.0.0.

3. **Broken docs link**: The security configuration link points to version 2.4.x docs but we're on 3.3.x now.

I checked git history and it looks like kubernetesops was added in #7571 back in March but the README never got updated. The version examples haven't been updated in a while either.

Not a huge deal but might confuse new contributors trying to follow the examples. Happy to submit a PR to fix these if that's helpful!

I'll submit a PR fixing all three issues together since they're all related to README documentation.

Itx-Psycho0:
Fixed a few documentation issues in the README:

1. Added the missing kubernetesops storage option to the list (it said there were 3 options but there are actually 4)

2. Updated the JAR path example from version 3.0.0 to use a wildcard so it works with any version

3. Changed the docker-compose example to use latest-release tag instead of the old 3.0.0 version

4. Fixed the security docs link that was pointing to version 2.4.x

These were pretty minor but might confuse people trying to follow the examples.

Fixes #8543


carlesarnal (approved):
Documentation updates look reasonable. You're adding kubernetesops storage variant and removing hardcoded version references.

Three observations:

1. The jar path change from `apicurio-registry-app-3.0.0-SNAPSHOT-runner.jar` to `apicurio-registry-app-*-runner.jar` uses a shell glob pattern. This works in bash but might confuse users who copy-paste literally. Consider adding a note that this is a pattern.

2. Docker image tag change from `3.0.0` to `latest-release` is good for keeping examples current, but `latest-release` might not exist in all registries. Verify this tag is published consistently.

3. Documentation link change from versioned path to generic `/docs/` loses version specificity. Users on older versions might get confused. Consider if this is intentional.

The kubernetesops addition matches the experimental flag requirement pattern used for gitops. Consistent with existing conventions.

Itx-Psycho0:
Thanks @carlesarnal  for the review and approval! Addressing your observations:

1. **JAR path pattern**: You're right, I can add a note. Should I change it to something like "apicurio-registry-app-3.3.1-SNAPSHOT-runner.jar (or use wildcard: *-runner.jar)"?

2. **Docker tag**: I verified that `latest-release` is published on Docker Hub (https://hub.docker.com/r/apicurio/apicurio-registry/tags). It's listed in the "Versioning & Support Policy" section above as a supported tag.

3. **Docs link**: I changed it to generic /docs/ because the specific 2.4.x link was broken. Would you prefer I use a current versioned link like /docs/apicurio-registry/3.3.x/ instead?

Let me know if you'd like me to update any of these!

EricWittmann:
Thanks for the contribution @Itx-Psycho0!

PR #8552, opened 2026-07-12
fix: proto2 ProtobufSerializer validation for schema registered as .proto text

### LINKED ISSUE #8279: Protobuf serializer validation fails on proto2 schemas (maps expanded, type names fully-qualified)
## Description

**RegistryVersion**: 3.1.x

With `apicurio.registry.serde.validation-enabled=true`, `ProtobufSerializer` validation always fails for a **proto2** schema registered as `.proto` text — even when the producer's message descriptor is identical to the registered schema. Every message is rejected. The same setup works correctly for **proto3** schemas.

Root cause: `ProtobufSerializer.validate(...)` diffs two `ProtobufFile`s built via different paths:
- **registered side**: `parsedSchema.getProtobufFile()`, parsed from the stored `.proto` text → `map<K,V>` collapsed, relative type names.
- **message side**: `new ProtobufFile(parser.toProtoFileElement(descriptor))` (i.e. `FileDescriptorUtils.fileDescriptorToProtoFile(descriptor.toProto())`) → maps expanded into synthetic `*Entry` repeated messages, type names fully-qualified.

`FileDescriptorUtils.label(file, field)` is gated on `getSyntax().equals("proto3")`: proto3 yields null labels that match the text-parsed side, but proto2 keeps explicit labels. Combined with maps never being re-collapsed to map<> and names being fully-qualified on the descriptor-derived side, the two ProtobufFiles can never match for a proto2 schema that contains a map field or a message reference. proto3 avoids this because both derivations agree.

### Environment

- `apicurio-registry-protobuf-serde-kafka`: 3.3.0
- `apicurio-registry-maven-plugin` (registration): 3.3.0
- `protobuf-java`: 4.35.0
- Java 21, Kafka producer using the Apicurio protobuf serde.

### Steps to Reproduce

1. Register a proto2 schema (with a `map` field and an imported message) as `PROTOBUF` text using the Maven plugin:

```xml
<artifact>
  <groupId>example</groupId>
  <artifactId>Envelope</artifactId>
  <file>${basedir}/src/main/proto/com/example/api/Envelope.proto</file>
  <ifExists>FIND_OR_CREATE_VERSION</ifExists>
  <canonicalize>true</canonicalize>
  <artifactType>PROTOBUF</artifactType>
  <references>
    <reference>
      <name>com/example/api/Detail.proto</name>
      <groupId>example</groupId>
      <artifactId>Detail</artifactId>
      <file>${basedir}/src/main/proto/com/example/api/Detail.proto</file>
      <ifExists>FIND_OR_CREATE_VERSION</ifExists>
      <canonicalize>true</canonicalize>
      <artifactType>PROTOBUF</artifactType>
    </reference>
  </references>
</artifact>
```

`Detail.proto`:
```proto
syntax = "proto2";
package com.example.api;
option java_multiple_files = true;

message Detail {
  required int64 id = 1;
}
```

`Envelope.proto`:
```proto
syntax = "proto2";
package com.example.api;

import "com/example/api/Detail.proto";

option java_multiple_files = true;

message Envelope {
  enum Type { UNKNOWN = 0; FULL = 1; }
  required Type type = 1;
  optional Detail detail = 2;
  optional Payload payload = 3;
}

message Payload {
  optional int64 id = 1;
  map<string, string> attributes = 2;
}
```

2. Configure a `ProtobufSerializer` resolving to the registered artifact:

```
apicurio.registry.auto-register            = false
apicurio.registry.find-latest              = true
apicurio.registry.serde.validation-enabled = true
```

3. Serialize any `Envelope` instance built from the protoc-generated class.

**Expected:** A message whose descriptor matches the registered schema validates with no differences.

**Actual:** Validation fails on every message, for every `map` field and every message-typed reference.

### Logs

```
java.lang.IllegalStateException: The data to send is not compatible with the schema.
[ProtobufDifference(message=Field type changed, message Payload , before: map<string, string> , after .com.example.api.Payload.AttributesEntry),
 ProtobufDifference(message=Field label changed, message Payload , before: null , after REPEATED),
 ProtobufDifference(message=Field type changed, message Envelope , before: Detail , after .com.example.api.Detail)]
```

kartikangiras:
## Description
- Collapse descriptor map fields back to map by detecting synthetic mapentry nested messages and replacing repeated EntryType fields with map fields having no label.

- Support explicit type/label passthrough in toField() by adding an overloaded variant so collapsed map fields can supply their own type and null label, bypassing proto2 label computation.

- Fix imported type resolution in normalizeType() by removing the fallback that incorrectly prefixed unknown type names with the current message scope instead resolving against the file's package per protobuf scoping rules.

- Add a test covering the exact reported scenario

## Issue
fixes #8279

### DIFF
--- a/schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/compatibility/protobuf/ProtobufCompatibilityCheckerLibrary.java
+++ b/schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/compatibility/protobuf/ProtobufCompatibilityCheckerLibrary.java
@@ -360,13 +360,9 @@ public class ProtobufCompatibilityCheckerLibrary {
             return buildFullyQualifiedName(file, type);
         }
 
-        // For types we can't locate in the schema, infer the fully qualified name
-        // This handles cross-file references or types that might be defined elsewhere
-        if (messageContext != null && !messageContext.isEmpty()) {
-            return buildFullyQualifiedName(file, messageContext + "." + type);
-        }
-
-        // Default: assume it's a top-level type
+        // For types not found in the current file (e.g., imported types), resolve against the
+        // current package. Protobuf scoping resolves unqualified simple names that are not found
+        // in nested scopes against the file's package, not against the current message context.
         return buildFullyQualifiedName(file, type);
     }
 
--- a/schema-util/protobuf/src/test/java/io/apicurio/registry/rules/compatibility/protobuf/ProtobufCompatibilityCheckerLibraryTest.java
+++ b/schema-util/protobuf/src/test/java/io/apicurio/registry/rules/compatibility/protobuf/ProtobufCompatibilityCheckerLibraryTest.java
@@ -1,10 +1,15 @@
 package io.apicurio.registry.rules.compatibility.protobuf;
 
+import com.google.protobuf.Descriptors.FileDescriptor;
+import com.squareup.wire.schema.internal.parser.ProtoFileElement;
+import com.squareup.wire.schema.internal.parser.ProtoParser;
 import io.apicurio.registry.protobuf.ProtobufDifference;
 import io.apicurio.registry.protobuf.rules.compatibility.protobuf.ProtobufCompatibilityCheckerLibrary;
+import io.apicurio.registry.utils.protobuf.schema.FileDescriptorUtils;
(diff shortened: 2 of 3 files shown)


carlesarnal:
@kartikangiras please, only one pr at a time. I'm closing this until https://github.com/Apicurio/apicurio-registry/pull/8545 is merged.

PR #8554, opened 2026-07-12
feat(auth): replace per-token error caching with circuit breaker for K8s TokenReview

### LINKED ISSUE #8462: Use circuit-breaker instead of per-token error caching for K8s TokenReview failures
## Context

From code review on #8076, when the K8s TokenReview API is unreachable, the current implementation caches the failure **per token hash** for 30 seconds. This means:

1. N different users hitting the registry during a ~2 second K8s API outage each get a 30-second negative cache entry
2. All N users must wait 30 seconds before their next request can succeed
3. A transient infrastructure failure (unrelated to any specific token) penalizes individual users

## Proposed solution

Replace per-token error caching with a circuit-breaker pattern:
- After N consecutive TokenReview API failures, stop calling the K8s API entirely for M seconds (regardless of which token)
- After M seconds, retry one probe request
- If the probe succeeds, close the circuit and resume normal operation

This correctly attributes the failure to the API server rather than to individual tokens.

Relates to #8076

Vidhaankhare16:
## Problem

When the Kubernetes TokenReview API has a transient outage, the current `KubernetesAuthenticationStrategy` caches the failure **per token hash** for 30 seconds. A ~2-second infrastructure blip therefore locks out each affected user for a full 30 seconds, attributing an API-server problem to individual tokens. Fixes #8462.

## Solution

Guard the TokenReview call with a MicroProfile Fault Tolerance circuit breaker (originally implemented as a hand-rolled `TokenReviewCircuitBreaker`; refactored to the standard `@CircuitBreaker` per review — see thread):

- The TokenReview API call is extracted into an `@ApplicationScoped` `TokenReviewClient` bean annotated `@CircuitBreaker(requestVolumeThreshold = 5, failureRatio = 1.0, delay = 30_000)`: 5 consecutive API failures open the circuit, calls fail fast for 30s, then SmallRye probes and closes on success — matching the #8462 spec.
- Only genuine API failures count: the bean throws on a missing response so it registers as a breaker failure; an `authenticated=false` review is a healthy API reply and returns normally. Per-token result caching is unchanged.
- The strategy consumes the bean (injected into `AppAuthenticationMechanism` and passed into the strategy constructor, since the auth chain is built manually from `apicurio.authn.mechanism.priority`) and catches `CircuitBreakerOpenException`, preserving the fail-open-to-next-mechanism contract. Cached identities keep authenticating while the circuit is open, so existing users survive a K8s API outage.
- The first rejection of each open episode is logged at WARN — including the security-relevant interaction with `apicurio.auth.anonymous-read-access.enabled` (while the circuit is open, K8s-token requests are treated as unauthenticated and, if anonymous read access is enabled, served as anonymous read-only) — subsequent rejections at DEBUG. No token or hash is ever logged.
- No custom config properties: the breaker is tuned or disabled via standard MP-FT config, e.g. `io.apicurio.registry.auth.TokenReviewClient/review/CircuitBreaker/enabled=false`. The two experimental `apicurio.authn.token-review.circuit-breaker.*` properties introduced by earlier versions of this PR were never released and are removed again, so there are no config-property changes. The only documentation change is a note in the security guide describing the fail-open-to-anonymous behavior when the circuit is open (added in response to review).

### Design decision: fail-fast behavior when the circuit is open

When the breaker is open (K8s TokenReview API down) and there is no cached identity, the strategy returns `null` — the same contract as today's API-error path, so the auth chain issues a 401 challenge. I chose this over a 503 deliberately: it preserves existing auth-chain semantics and keeps anonymous access working (when enabled) during an outage, and it needs no new exception-mapping plumbing. Happy to switch to a 503 in a follow-up if you'd prefer clients to see it as an infrastructure error rather than an auth error.

## Testing

- `TokenReviewClientCircuitBreakerTest` — a `@QuarkusTest` that exercises the **real** `@CircuitBreaker` through the CDI proxy against the fabric8 mock Kubernetes API server (no mocked breaker outcomes). It proves: the interceptor is actually wired; five consecutive API failures open the circuit and further calls — sequential and 10-thread concurrent — fail fast without reaching the API server (asserted via server request counts); `authenticated=false` responses never trip the breaker; threads racing the probe window after the open delay behave consistently (at least one probe admitted, the rest rejected with `CircuitBreakerOpenException`, circuit closed afterwards); and the documented standard MP-FT override (`TokenReviewClient/review/CircuitBreaker/delay`) really applies — the test uses it to shorten the open window.
- `TokenReviewClientTest` — 5 unit tests: status returned as-is; `authenticated=false` is a normal return (never a breaker failure); a missing API response throws; audiences included/omitted per config.
- `KubernetesAuthenticationStrategyTest` — 14 tests: `CircuitBreakerOpenException` maps to unauthenticated/fail-open with no negative caching (traffic resumes the moment the breaker admits calls again); a cached identity keeps authenticating while the circuit is open (the cache lookup short-circuits before the breaker); recovery after the circuit closes; a single API failure doesn't block other tokens; result caching; header parsing.
- Targeted classes + `checkstyle:check` pass locally; relying on CI for the full suite.

Fixes #8462


Vidhaankhare16:
Hi @carlesarnal, thank you for assigning me the issue. I've addressed the Copilot review comments (constructor validation for a negative timeout; log-level and recordSuccess-ordering handled earlier) and made the null-vs-503 call - returning null (401  -> same contract as today's API-error path) to preserve existing auth-chain semantics and keep anonymous access working, documented inline and in the description. Happy to switch to 503 in a follow-up if you'd prefer. I think this is ready for review.

paoloantinori:
@Vidhaankhare16 Reviewed the circuit breaker implementation. The overall design is sound — fail-open to the next auth mechanism when the K8s API is unavailable. Some concerns:

**MAJOR — Hand-rolled circuit breaker instead of MicroProfile Fault Tolerance**
The project already has \`quarkus-smallrye-fault-tolerance\` (MicroProfile) as a dependency. The hand-rolled \`TokenReviewCircuitBreaker\` reimplements state management (\`CLOSED/OPEN/HALF_OPEN\`), timing, and threshold logic that \`@CircuitBreaker\` provides out of the box. Using the standard annotation would be shorter, battle-tested, and consistent with Quarkus patterns. Is there a specific reason you didn't use it?

**MAJOR — Synchronized methods under async auth**
\`recordSuccess()\` and \`recordFailure()\` are \`synchronized\`, which blocks the event loop thread in Quarkus's reactive auth pipeline (\`Uni<SecurityIdentity>\`). Under high request rates, this serializes all K8s auth decisions. Consider \`AtomicReference<State>\` + CAS instead, or \`@CircuitBreaker\` which handles this internally.

**MINOR — DoS amplification risk**
When the circuit opens, ALL K8s-authenticated tokens are rejected for \`openDuration\` seconds (30s default). An attacker can trigger this by sending \`threshold\` (5) invalid tokens in sequence, then legitimate K8s service accounts are locked out for 30s. Consider whether the threshold should be higher or time-windowed (e.g., "5 failures within 10 seconds" rather than "5 consecutive failures").

**NIT — Tests are thorough**
Good coverage of state transitions, thread safety, and integration with \`KubernetesAuthenticationStrategy\`.

Vidhaankhare16:
Thanks for the thorough review @paoloantinori!

On the "why hand-rolled" question: the strategy is constructed manually in `AppAuthenticationMechanism.buildAuthChain()` (the auth chain is config-driven, so strategies aren't CDI beans), which means `@CircuitBreaker`'s interceptor can't apply to it directly. But I agree the standard annotation is the right call. My plan: extract the TokenReview API call into a small `@ApplicationScoped` bean annotated with `@CircuitBreaker(requestVolumeThreshold = 5, failureRatio = 1.0, delay = 30_000)` - which maps onto the "N consecutive failures -> skip for M seconds -> probe" spec from #8462, and have the strategy call it (injected into `AppAuthenticationMechanism` and passed into the strategy's constructor), catching `CircuitBreakerOpenException` to preserve the current fail-open-to-next-mechanism contract. That deletes `TokenReviewCircuitBreaker` entirely and resolves the concurrency concern too, since SmallRye manages breaker state internally (FWIW the current `synchronized` calls run on the worker pool via `runSubscriptionOn(...)`, not the event loop, but the shared lock disappears either way).

One clarification on the DoS point: invalid tokens can't trip the breaker - a TokenReview response with `authenticated: false` is a healthy API reply and records success; only API call failures count (covered by `testUnauthenticatedResponseDoesNotTripCircuitBreaker`). That said, the annotation's rolling-window semantics (failure ratio over the last N calls rather than a consecutive count) are still an improvement in the direction you suggested.

Knock-on change: the two custom properties (`apicurio.authn.token-review.circuit-breaker.*`) would be dropped in favor of annotation defaults + standard MicroProfile Fault Tolerance config overrides, and I will regenerate the config reference docs. Both properties are new in this PR (experimental, never released), so removing them has no backward-compat impact.

Does this sound good or would you prefer the whole strategy become a CDI bean instead? The extracted bean seemed less invasive given the chain is built conditionally from `apicurio.authn.mechanism.priority`.

carlesarnal (changes requested):
Circuit breaker implementation is solid — correct state machine, fail-closed behavior, well tested. Good replacement for the per-token error caching.

A few things:

- Document the interaction between circuit-open and anonymous-fallback-enabled. When the circuit opens and anonymous is enabled, all requests silently become anonymous. That's security-relevant and operators need to know.

- Consider whether `threshold=0` should disable the circuit breaker entirely rather than normalizing to 1. Current behavior may surprise users.

- Add a concurrent half-open test — multiple threads hitting the probe window simultaneously.

Code style and config property annotations follow conventions. Strong test coverage overall.

Vidhaankhare16:
Thank you for the review @carlesarnal, I understand all three points. Before I start, I would like to confirm what direction I should take, because paoloantinori's review above suggested replacing the hand-rolled `TokenReviewCircuitBreaker` with MicroProfile `@CircuitBreaker` (I proposed extracting the TokenReview call into an `@ApplicationScoped` bean so the interceptor can apply in my comment above.) The two directions change how I'd address your points:

**If we keep the hand-rolled breaker:** I'll document the circuit-open + `apicurio.auth.anonymous-read-access.enabled` interaction in the config reference and log it explicitly at the OPEN transition; agree `threshold=0` should disable the breaker entirely (pass-through) rather than normalize to 1 - will implement with a test; and I'll add the concurrent half-open test (multiple threads racing the probe window, exactly one probe admitted).

**If we migrate to `@CircuitBreaker`:** the documentation point applies unchanged; disabling becomes the standard MP-FT `enabled=false` override (or I can keep a friendly `apicurio.*` property mapped onto it, your preference); and half-open concurrency becomes SmallRye's responsibility - I'd keep an integration-level test that the strategy handles `CircuitBreakerOpenException` correctly under concurrent requests.

Happy to do either - which should I implement?

Vidhaankhare16:
Update: I've gone with the MicroProfile `@CircuitBreaker` approach - paoloantinori's review and the direction of #8626 both point that way.

The TokenReview call now lives in an `@ApplicationScoped` `TokenReviewClient` bean annotated `@CircuitBreaker(requestVolumeThreshold = 5, failureRatio = 1.0, delay = 30_000)`, matching the #8462 spec. The bean throws on a missing API response so it registers as a breaker failure; an `authenticated=false` review is a normal return and never trips it. The strategy catches `CircuitBreakerOpenException` and preserves the fail-open-to-next-mechanism contract. `TokenReviewCircuitBreaker` and the two custom config properties are gone (never released, so no compat impact).

@carlesarnal your three points, in this shape:
- **circuit-open × anonymous-read-access**: documented in the bean javadoc, strategy comment and logged at WARN once per open episode (DEBUG after, to avoid flooding);
- **threshold=0 disables**: now the standard MP-FT switch `io.apicurio.registry.auth.TokenReviewClient/review/CircuitBreaker/enabled=false`, documented;
- **concurrent half-open**: SmallRye owns breaker state now, and it's covered by a real test rather than delegation - see follow-up below. I also implemented all three on the hand-rolled version first (624e0c2eb in this PR's history, including a 20-thread half-open race test asserting exactly one probe is admitted) - so if you'd rather keep the hand-rolled shape, it's one revert away.

Follow-up: pushed `TokenReviewClientCircuitBreakerTest`, a `@QuarkusTest` that exercises the real `@CircuitBreaker` through the CDI proxy against the fabric8 mock K8s API server - verifying the interceptor is actually wired, five consecutive API failures open the circuit (further calls fail fast without reaching the server, asserted via request counts), `authenticated=false` never trips it, and concurrent bursts racing the open->half-open probe window behave consistently (one probe admitted, the rest rejected, circuit closes on probe success). It also pins the documented `TokenReviewClient/review/CircuitBreaker/delay` override, proving the standard MP-FT config applies.


Ready for another look.

jsenko:

Vidhaankhare16:
Thank you @jsenko , I updated the branch with current main and License Compliance is passing now.

paoloantinori:
Thanks — extracting the TokenReview call into an `@ApplicationScoped` `TokenReviewClient` annotated with `@CircuitBreaker(requestVolumeThreshold = 5, failureRatio = 1.0, delay = 30_000)` directly resolves my main concern (using the standard annotation instead of a hand-rolled breaker), and because the strategy now calls through the CDI-injected bean the interceptor is actually wired — no self-invocation gap. The `TokenReviewClientCircuitBreakerTest` exercising the real breaker through the proxy (open after 5 failures, concurrent rejections, probe/close recovery) is exactly the right level of confidence. The fail-open-to-next-mechanism contract and the WARN-once logging look sound. I'll defer the remaining items to @carlesarnal's review.

Vidhaankhare16:
Hi @carlesarnal, this is ready for another look whenever it fits your queue. Since your review, the implementation moved to MicroProfile Fault Tolerance (extracted `@ApplicationScoped` `TokenReviewClient` with `@CircuitBreaker`) following paoloantinori's feedback.

Your three points all carried over into the new shape (full mapping in my comment above): the anonymous-read-access interaction is documented in the javadoc plus a WARN when the circuit opens, disabling is now the standard `io.apicurio.registry.auth.TokenReviewClient/review/CircuitBreaker/enabled=false` override rather than a custom threshold=0 knob, and `TokenReviewClientCircuitBreakerTest` exercises the real interceptor against a mock Kubernetes API server, including the concurrent open-state and half-open probe cases.

The branch is updated to current main, thank you!

Vidhaankhare16:
Hi @paoloantinori - thanks again for the review and the sign-off earlier.

Could I ask your advice on one thing? The full test suite hasn't run on this PR yet, and I'm not sure whether that's something I need to do from my side or whether it's waiting on a maintainer step. Is there anything I should be doing to move it along?

No rush on the review itself, I know the queue is busy. Happy to rebase or make any changes in the meantime.

paoloantinori:
@Vidhaankhare16 Happy to clarify — nothing needed from your side on the test front. The auth unit tests run on every push as part of the normal CI, and they're already green on this PR (you can see them in the workflow checks). The heavier integration suites (the ones that spin up the full registry against a real/K3s cluster) are gated and run only for PRs a maintainer explicitly kicks them off on, typically right before merge. So a missing "full suite" run is expected at this stage, not a blocker.

What I need from a maintainer before we can move toward merge:
- a final rebase on current `main` (the branch is showing `behind`, which can also silently fail the gated jobs), and
- a second sign-off from @carlesarnal on the MP Fault Tolerance rewrite.

The implementation itself still looks good to me after the re-read — the `@CircuitBreaker` wiring on `TokenReviewClient`, the fail-open-to-next-mechanism contract on `CircuitBreakerOpenException`, and the dedicated `TokenReviewClientCircuitBreakerTest` covering the real interceptor lifecycle (open/probe/close) all line up with issue #8462. No new blocking concerns from me.

app/src/main/java/io/apicurio/registry/auth/KubernetesAuthenticationStrategy.java
```diff
@@ -107,47 +105,59 @@ public Uni<SecurityIdentity> authenticate(RoutingContext context,
...
+            // This is deliberate rather than a 503: it preserves the existing failure semantics and
+            // keeps anonymous access working (when enabled) while the TokenReview API is down.
+            // Security note: when apicurio.auth.anonymous-read-access.enabled=true, requests that
+            // would normally authenticate via a Kubernetes token are served as anonymous read-only
+            // requests for as long as the circuit is open. The first rejection of each open episode
+            // is logged at WARN (including this fallback) so operators can correlate the change in
+            // behavior; subsequent rejections log at DEBUG to avoid flooding.
+            if (circuitOpenWarned.compareAndSet(false, true)) {
+                log.warn("Kubernetes TokenReview circuit breaker is OPEN; skipping TokenReview "
+                        + "calls. Requests with Kubernetes tokens will be treated as "
```
paoloantinori, line 124: The fail-open-to-anonymous-read behavior during an open circuit is well documented in this code comment, but no `docs/` change ships in the PR. For an auth-availability tradeoff this significant (authenticated write users silently degrade to anonymous read-only during a K8s API outage when `anonymous-read-access.enabled=true`), it's worth adding a short operator-facing note alongside the K8s auth config in the docs so the tradeoff is discoverable at deploy time, not only by reading this catch block.

paoloantinori (commented):
Re-read head `69607e4`. The MicroProfile `@CircuitBreaker` extraction is clean, and `TokenReviewClientCircuitBreakerTest` does the right thing by exercising the real CDI proxy against the mock K8s server rather than mocking the breaker's outcome — that catches the classic MP-FT "annotation silently never applies" footgun. The `circuitOpenWarned` dedup logic and the "healthy reply resets the flag" semantics are correct. One substantive gap worth addressing before merge:

- **Operator-facing documentation for the fail-open-to-anonymous behavior.** When the circuit is open and `apicurio.auth.anonymous-read-access.enabled=true`, requests carrying valid Kubernetes tokens are silently served as **anonymous read-only** for the duration of the outage (the code comment in `KubernetesAuthenticationStrategy.performTokenReview` documents this, but no `docs/` change ships in this PR). That is a genuine privilege downgrade during an incident: an operator who has anonymous read enabled for unauthenticated browsing will, during a K8s API blip, see authenticated write users silently lose write access without any client-side signal. The WARN log on first-open helps, but only operators actively tailing logs during the incident will see it. Worth adding a short note to the auth docs (e.g. alongside the K8s auth config section) so this availability tradeoff is discoverable at deploy time, not only in code comments. The behavior itself is defensible (fail-open beats a hard lockout), I'm only flagging the discoverability.

- Minor, worth confirming: the positive cache lookup that lets "cached identities keep authenticating while the circuit is open" (per the PR description) happens upstream of `performTokenReview` in `authenticate()` — that path isn't in this diff, so I couldn't verify it from the PR alone. The claim is consistent with `testCachesTokenReviewResult`, but if you can confirm the cache hit short-circuits before the `CircuitBreakerOpenException` path is reachable, that'd close the loop. If the cache miss→circuit-open path is the only thing tested, a test asserting that a previously-cached identity still authenticates while the breaker is open would make the guarantee explicit.

No correctness blockers from me on this head. The breaker lifecycle test (Phase 4 concurrent probe racing) is particularly well done.

Vidhaankhare16:
Thanks @paoloantinori — both good calls.

- Added an operator-facing note to the anonymous read-only access section of the security guide, spelling out the fail-open-to-anonymous tradeoff during a Kubernetes API outage: while the circuit is open, requests carrying Kubernetes tokens are treated as unauthenticated and, if anonymous read access is enabled, served as anonymous read-only — so authenticated write users silently degrade to read-only until the API recovers (with the warn-on-open to correlate it). It's now discoverable at deploy time, not just in the catch block.
- On the cache point: confirmed the positive cache lookup in `KubernetesAuthenticationStrategy.authenticate()` short-circuits before `performTokenReview`, so a cached identity keeps authenticating while the breaker is open — only uncached tokens reach the open circuit and fail open. Added `testCachedIdentityStillAuthenticatesWhileCircuitOpen` asserting exactly that (cache a token, open the circuit for other tokens, verify the cached token still authenticates without another `review()` call).

Both pushed (checkstyle clean, `KubernetesAuthenticationStrategyTest` green). Thanks again for the review!

paoloantinori:
@Vidhaankhare16 Thanks — both changes look good.

The doc note captures the tradeoff precisely: silent downgrade to anonymous read-only for authenticated write users during a K8s API outage, with the operator escape hatch of disabling anonymous read access. The `circuitOpenWarned` flag is the right shape too — WARN on the first rejection of each open episode, DEBUG thereafter, and resetting on a healthy reply so the next episode warns again. No further comments from my side.

Vidhaankhare16:
Hi @carlesarnal - this is ready for your re-review whenever you have a moment. @paoloantinori has signed off; his two follow-ups (an operator doc note on the fail-open-to-anonymous tradeoff, and a test proving a cached identity keeps authenticating while the circuit is open) are on the current head. Branch is current with `main` and CI is green. Your three original points are mapped onto the MicroProfile `@CircuitBreaker` shape in my earlier comment above. Thanks!

paoloantinori:
@Vidhaankhare16 — nothing further from my side; my earlier sign-off stands on this head, and the two follow-ups (operator doc note + cached-identity-while-open test) both look good. Deferring to @carlesarnal for the re-review.

Vidhaankhare16:
Hi @carlesarnal, a quick question - this PR (on your issue #8462) is still in lifecycle/new and hasn't been /accepted, so the test pipeline has never actually run on it. Paolo reviewed it and signed off above, and the branch is current with main. Could a maintainer /accept it when there's a moment, so it can go through smoke + full tests? That would also give me a clear read on anything still needed on my side. No rush on the review itself - thank you!

carlesarnal:
Reviewed head c432285d (two new commits: 5361a384 docs, 76fbad7b test).

Maintainer feedback addressed:

carlesarnal CHANGES_REQUESTED:
1. Document circuit-open + anonymous-fallback -- The new IMPORTANT block in docs/assembly-configuring-registry-security.adoc (lines 1350-1363) clearly describes the availability tradeoff and recommends disabling anonymous reads if the silent downgrade is unacceptable. Well placed.
2. threshold=0 disabling the breaker -- Handled by the refactor to MicroProfile CircuitBreaker with standard MP-FT config.
3. Concurrent half-open test -- Phase 4 of circuitBreakerLifecycle() races threads through the probe window with CyclicBarrier + Awaitility.

paoloantinori suggestions:
1. Operator-facing documentation -- Same docs block above.
2. Test cached identity authenticating while circuit is open -- testCachedIdentityStillAuthenticatesWhileCircuitOpen verifies a previously-cached token keeps authenticating from cache while the circuit is open, and uncached tokens fail open. Assertions confirm the cached token review is called exactly once (cache hit short-circuits before the breaker).

Code review notes:

Verified the cache-hit short-circuit path: in authenticate() (lines 87-96), the cache lookup and early return happen before performTokenReview() where the CircuitBreakerOpenException is caught. Cached identities never touch the circuit breaker.

The circuitOpenWarned AtomicBoolean dedup logic is correct: compareAndSet for the first WARN, reset after a successful API reply.

All new Java files have Apache 2.0 license headers, no star imports, K&R brace style, Locale.ROOT on the SHA-256 hex format. All commits are DCO-signed with conventional commit messages. CI is green.

No issues found. All maintainer feedback has been addressed. This looks ready for maintainer re-review.

carlesarnal:
Re-reviewed head c432285d. All previously flagged issues are addressed:

**carlesarnal's 3 original points:**
1. **Circuit-open + anonymous-fallback documentation** — The new `IMPORTANT` block in `docs/modules/ROOT/pages/getting-started/assembly-configuring-registry-security.adoc` (lines 1350–1363) clearly describes the availability tradeoff: authenticated write users silently downgrade to anonymous read-only while the circuit is open. The recommendation to disable anonymous read access if that's unacceptable is well placed. The javadoc on `TokenReviewClient`, the inline comment in `performTokenReview`, and the WARN log on first open all reinforce discoverability.
2. **Disabling the breaker** — Handled cleanly by the MicroProfile refactor: standard `TokenReviewClient/review/CircuitBreaker/enabled=false` config override, no custom properties needed.
3. **Concurrent half-open test** — Phase 4 of `circuitBreakerLifecycle()` in `TokenReviewClientCircuitBreakerTest` races 10 threads through the probe window using `CyclicBarrier` + Awaitility burst-polling, asserting at least one probe is admitted and the circuit closes on success.

**paoloantinori's feedback** (signed off):
- MicroProfile `@CircuitBreaker` on `@ApplicationScoped` `TokenReviewClient` — clean extraction, interceptor is wired through the CDI proxy.
- Operator documentation — same docs block above.
- `testCachedIdentityStillAuthenticatesWhileCircuitOpen` — confirms the cache lookup short-circuits before `performTokenReview()`, so cached identities authenticate normally while the breaker is open.

**Verification notes:**
- Confirmed the cache-hit path in `authenticate()` (lines 87–96) returns before reaching `performTokenReview()` — cached identities never touch the circuit breaker.
- `circuitOpenWarned` AtomicBoolean dedup logic is correct: `compareAndSet` for the first WARN per episode, `set(false)` after any healthy API reply resets it.
- No negative caching of API errors or circuit-open rejections — the circuit breaker handles the "don't flood" concern globally, so per-token retries resume immediately once the breaker admits calls.
- All authored commits are DCO-signed with conventional commit messages. No star imports, K&R brace style, `Locale.ROOT` on hex formatting, Apache 2.0 headers on all new files. CI and SonarCloud are green.

No issues found. This looks ready for maintainer approval.

EricWittmann:
Thanks @Vidhaankhare16 — really well done. The refactor from hand-rolled to MicroProfile @CircuitBreaker was the right call, the CDI-based tests exercise the real breaker proxy (not mocked away), and the security implications are thoroughly documented for operators. Appreciate your patience through the 3-week review cycle. Merged!

PR #8555, opened 2026-07-12
chore(deps): configure renovate to scan .mvn/extensions.xml

### LINKED ISSUE #8498: Configure Renovate to track .mvn/extensions.xml dependencies
Scalpel 0.3.7 is declared in `.mvn/extensions.xml` but Renovate (`renovate.json`) doesn't scan this file. Extensions will go stale without automated dependency update tracking.

Add a custom manager or fileMatch rule to Renovate so Maven core extensions are included in dependency update PRs.

krish-kedia:
Issue:
Currently, the default Renovate Maven manager exclusively scans standard pom.xml configurations for dependency updates. Consequently, Maven Core Extensions declared within .mvn/extensions.xml (such as com.github.kordamp.scalpel:scalpel-maven-extension) are completely ignored by the automated dependency tracker. Over time, this leads to silent toolchain decay where build infrastructure extensions go stale without automated PR generation.

Solution:
This PR implements a robust custom Regex Manager within renovate.json specifically engineered to parse the Maven Core Extensions XML schema, ensuring these infrastructure dependencies are bound to the standard Renovate lifecycle.

Implementation:
- Explicitly injected "custom.regex" into the strict enabledManagers array to authorize regex-based extraction pipelines alongside the default engines.
- Implemented a new customManagers block with a strict file-match boundary of ^\.mvn/extensions\.xml$.
- Engineered a highly resilient extraction regex that explicitly bounds whitespace (\s*) outside of the capture groups while leveraging negative character classes ([^<\s]+) inside the capture groups. This ensures that even if developers introduce arbitrary newlines or indentation inside the XML tags, Renovate will extract mathematically clean strings without failing downstream registry lookups.
- The extracted coordinate metadata (depGroup, depName, currentValue) is mapped directly to "datasourceTemplate": "maven", instructing Renovate to automatically poll Maven Central for new releases of the captured extensions.

Fixes: #8498

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -12,7 +12,8 @@
     "npm",
     "maven",
     "github-actions",
-    "dockerfile"
+    "dockerfile",
+    "custom.regex"
   ],
   "dependencyDashboard": true,
   "rangeStrategy": "bump",
@@ -31,6 +32,18 @@
   ],
   "prConcurrentLimit": 10,
   "prHourlyLimit": 10,
+  "customManagers": [
+    {
+      "customType": "regex",
+      "fileMatch": [
+        "^\\.mvn/extensions\\.xml$"
+      ],
+      "matchStrings": [
+        "<groupId>\\s*(?<depGroup>[^<\\s]+)\\s*</groupId>\\s*<artifactId>\\s*(?<depName>[^<\\s]+)\\s*</artifactId>\\s*<version>\\s*(?<currentValue>[^<\\s]+)\\s*</version>"
+      ],
+      "datasourceTemplate": "maven"
+    }
+  ],
   "packageRules": [
     {
       "groupName": "Dependencies: Maven App: All",
--- a/renovate.json
+++ b/renovate.json
@@ -41,6 +41,8 @@
       "matchStrings": [
         "<groupId>\\s*(?<depGroup>[^<\\s]+)\\s*</groupId>\\s*<artifactId>\\s*(?<depName>[^<\\s]+)\\s*</artifactId>\\s*<version>\\s*(?<currentValue>[^<\\s]+)\\s*</version>"
       ],
+      "depNameTemplate": "{{depGroup}}:{{depName}}",
+      "packageNameTemplate": "{{depGroup}}:{{depName}}",
       "datasourceTemplate": "maven"
     }
   ],
--- a/renovate.json
+++ b/renovate.json
@@ -36,13 +36,12 @@
     {
       "customType": "regex",
       "fileMatch": [
-        "^\\.mvn/extensions\\.xml$"
+        "(^|/)\\.mvn/extensions\\.xml$"
       ],
       "matchStrings": [
         "<groupId>\\s*(?<depGroup>[^<\\s]+)\\s*</groupId>\\s*<artifactId>\\s*(?<depName>[^<\\s]+)\\s*</artifactId>\\s*<version>\\s*(?<currentValue>[^<\\s]+)\\s*</version>"
       ],
       "depNameTemplate": "{{depGroup}}:{{depName}}",
-      "packageNameTemplate": "{{depGroup}}:{{depName}}",
       "datasourceTemplate": "maven"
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -31,6 +32,18 @@
...
   "prConcurrentLimit": 10,
   "prHourlyLimit": 10,
+  "customManagers": [
+    {
+      "customType": "regex",
+      "fileMatch": [
+        "^\\.mvn/extensions\\.xml$"
+      ],
+      "matchStrings": [
+        "<groupId>\\s*(?<depGroup>[^<\\s]+)\\s*</groupId>\\s*<artifactId>\\s*(?<depName>[^<\\s]+)\\s*</artifactId>\\s*<version>\\s*(?<currentValue>[^<\\s]+)\\s*</version>"
```
aclfe, line 42: Thanks for taking this on! 

I think the overall approach looks great (it matches the approach from the official docs). One thing I want to confirm though, Renovate only maps a fixed set of named capture groups onto a dependency. The list:  

https://github.com/renovatebot/renovate/blob/eeeb29bd9699828706add71fbae24946b04df0eb/lib/modules/manager/custom/utils.ts#L5

depGroup isn't in it (and isn't mentioned anywhere in the https://docs.renovatebot.com/modules/manager/regex/ )

That leaves depName holding just the artifactId. Maven datasource splits packageName. With no colon, name comes back undefined, so the lookup can't resolve and no update PR would be opened.

Could adding a packageNameTemplate work?

Another relevant doc:
https://docs.renovatebot.com/presets-customManagers/

krish-kedia: @aclfe thank you so much. depGroup isn't a natively mapped variable in the custom manager pipeline, and without receiving the strictly formatted groupId:artifactId needed for registry lookups, the maven datasource would have silently failed. adding the packageNameTemplate works and i've just pushed an update to the PR that utilizes your suggested template to safely concatenate the extracted groups. this guarantees the datasource resolves the package correctly.

also noticed that the requirement for this tracking was originally flagged in the scalpel PR thread, hence added closes #8498 to the commit message just to keep the issue tracker tidy.

paoloantinori:
@krish-kedia This is a good idea — we noted in PR #8442 review that Renovate doesn't track `.mvn/extensions.xml` (Eric flagged the Scalpel version pinning without update tracking). This PR directly addresses that gap. LGTM on the approach.

carlesarnal (dismissed):
Configuration looks correct. The regex pattern properly captures Maven extension declarations. Adding `custom.regex` to managers list is required for custom managers to work.

One note: the fileMatch pattern `^\\.mvn/extensions\\.xml$` will only match the exact file `.mvn/extensions.xml` at repository root. If you have extensions.xml files in subdirectories (like in multi-module projects), this won't catch them. Consider `(^|/)\\.mvn/extensions\\.xml$` if needed.

The depNameTemplate and packageNameTemplate are identical - this is fine but packageNameTemplate might be redundant since it defaults to depName.

carlesarnal:
@krish-kedia please, only one pr at a time. Even though this looks good, I'm closing it until https://github.com/Apicurio/apicurio-registry/pull/8524 is merged. You can reopen once that's done.

krish-kedia:
@carlesarnal thanks for the review!

i’ve updated the pattern to (^|/)\\.mvn/extensions\\.xml$ to correctly catch extensions in subdirectories.

the depGroup / missing colon concern: even though Renovate doesn't natively map depGroup, its template engine still evaluates "depNameTemplate": "{{depGroup}}:{{depName}}" into the full groupId:artifactId. Because packageName automatically defaults to the evaluated depName, the Maven lookup gets the colon it needs and succeeds perfectly.

because of this automatic fallback, having an identical packageNameTemplate is indeed redundant, so I've removed it. it would only be useful if we wanted to split them for cleaner PR titles, like depNameTemplate: "{{depName}}".

if all looks good, i am ready to open my PR for this issue again. waiting for your response.

PR #8558, opened 2026-07-13
fix(core): Resolve FIXME for GraphQL content-type auto-detection

### LINKED ISSUE #8557: Fix GraphQL content-type auto-detection in ContentTypeUtil
## Description

Currently, there is an undocumented `FIXME` in `ContentTypeUtil.java` regarding GraphQL schemas. Because valid GraphQL schemas can sometimes be parsed as valid YAML (e.g., `type Query { hello: String }`), the existing logic incorrectly identifies GraphQL files as YAML instead of `application/graphql`.

### Environment

Local development environment running unit tests against `io.apicurio.registry.content.util.ContentTypeUtil`.

### Steps to Reproduce

1. Call `ContentTypeUtil.determineContentType(ContentHandle.create("type Query { hello: String }"))` with a valid GraphQL schema string.
2. Observe the returned content type.

**Expected:** The method should return `application/graphql`.
**Actual:** The method incorrectly returns `application/x-yaml` because the YAML parser successfully parses the GraphQL syntax as a YAML mapping before the GraphQL detection can take place.

### Logs

N/A - This was identified via code inspection of `ContentTypeUtil.java` and unit testing. I have already prepared a fix that prioritizes GraphQL keyword detection before attempting YAML parsing and will be submitting a PR shortly!

dolliecoder:
### Motivation

Currently, there is an undocumented `FIXME` in `ContentTypeUtil.java` regarding GraphQL schemas. Because valid GraphQL schemas can sometimes be parsed as valid YAML (e.g., `type Query { hello: String }` is technically a valid YAML string mapping), the existing logic was incorrectly identifying GraphQL files as YAML. This PR fixes the bug by prioritizing GraphQL keyword detection before falling back to the YAML parser.

Resolves #8557 

### Modifications

* **`ContentTypeUtil.java`**: 
  * Created a new heuristic detection method `isParsableGraphQL(ContentHandle content)` that safely reads text and checks for common GraphQL schema keywords (`type`, `interface`, `schema {`, `union`, `scalar`, `input`, etc.) to accurately identify GraphQL without bringing in heavy external parser dependencies.
  * Updated `determineContentType(ContentHandle content)` to check for GraphQL *before* checking for YAML.

* **`ContentTypeUtilTest.java`**: 
  * Added `testIsParsableGraphQL()` to verify the new heuristic successfully detects valid GraphQL and correctly rejects other formats like Protobuf.
  * Added `testDetermineContentType_GraphQL()` to assert that passing a GraphQL schema returns the expected `application/graphql` constant.

### Result / Validation

All tests in the `schema-util/common` module pass successfully. The registry will now correctly auto-detect `application/graphql` content types without incorrectly capturing them as `application/x-yaml`.

```text
[INFO] Running io.apicurio.registry.content.util.ContentTypeUtilTest
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
@@ -200,6 +200,9 @@ public final class ContentTypeUtil {
     }
 
     public static String determineContentType(ContentHandle content) {
+        // Ensure content is fully materialized once, so multiple parse attempts
+        // don't depend on stream re-use.
+        content.bytes();
         if (isParsableJson(content)) {
             return CT_APPLICATION_JSON;
         }
--- a/schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
@@ -175,7 +175,7 @@ public final class ContentTypeUtil {
         return node;
     }
 
-    private static final Pattern GRAPHQL_DEF_PATTERN = Pattern.compile("(?m)^\\s*(type|interface|scalar|union|input|enum)\\s+[a-zA-Z_][a-zA-Z0-9_]*\\s*(?:\\{|implements|\\=|$)");
+    private static final Pattern GRAPHQL_DEF_PATTERN = Pattern.compile("(?m)^\\s*(?:extend\\s+)?(type|interface|scalar|union|input|enum)\\s+[a-zA-Z_][a-zA-Z0-9_]*\\s*(?:\\{|implements|\\=|$)");
     private static final Pattern GRAPHQL_SCHEMA_PATTERN = Pattern.compile("(?m)^\\s*schema\\s*\\{");
     private static final Pattern GRAPHQL_DIRECTIVE_PATTERN = Pattern.compile("(?m)^\\s*directive\\s+@[a-zA-Z_][a-zA-Z0-9_]*");
 
--- a/schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
@@ -14,6 +14,7 @@ import javax.xml.parsers.SAXParser;
 import javax.xml.parsers.SAXParserFactory;
 import java.io.IOException;
 import java.io.StringReader;
+import java.util.regex.Pattern;
 
 public final class ContentTypeUtil {
 
@@ -174,6 +175,10 @@ public final class ContentTypeUtil {
         return node;
     }
 
+    private static final Pattern GRAPHQL_DEF_PATTERN = Pattern.compile("(?m)^\\s*(type|interface|scalar|union|input|enum)\\s+[a-zA-Z_][a-zA-Z0-9_]*\\s*(?:\\{|implements|\\=|$)");
+    private static final Pattern GRAPHQL_SCHEMA_PATTERN = Pattern.compile("(?m)^\\s*schema\\s*\\{");
+    private static final Pattern GRAPHQL_DIRECTIVE_PATTERN = Pattern.compile("(?m)^\\s*directive\\s+@[a-zA-Z_][a-zA-Z0-9_]*");
+
     /**
      * Returns true if the content is likely a GraphQL schema.
      */
@@ -183,25 +188,20 @@ public final class ContentTypeUtil {
             if (text.startsWith("{") || text.startsWith("<")) {
                 return false;
             }
-            // Normalize whitespace so keyword checks work across tabs/newlines and formatting differences.
-            String normalized = text.replaceAll("\\s+", " ");
-            // Basic heuristics to differentiate GraphQL from Protobuf
-            return normalized.contains("type ")
-                    || normalized.contains("interface ")
-                    || normalized.contains("scalar ")
-                    || normalized.contains("union ")
-                    || normalized.contains("input ")
-                    || normalized.contains("schema{")
-                    || normalized.contains("schema {")
-                    || normalized.contains("directive @");
+            
+            return GRAPHQL_DEF_PATTERN.matcher(text).find()
+                    || GRAPHQL_SCHEMA_PATTERN.matcher(text).find()
+                    || GRAPHQL_DIRECTIVE_PATTERN.matcher(text).find();
         } catch (Exception e) {
             return false;
         }
     }
 
     public static String determineContentType(ContentHandle content) {
-        // Ensure content is fully materialized once, so multiple parse attempts
-        // don't depend on stream re-use.
+        // Ensure content is fully materialized once in memory.
+        // If we do not call bytes() here, the isParsableJson check may consume
+        // the underlying InputStream (e.g. via ObjectMapper.readTree), leaving
+        // the stream empty and causing subsequent format checks to incorrectly fail.
         content.bytes();
         if (isParsableJson(content)) {
             return CT_APPLICATION_JSON;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dolliecoder** [comment]
@EricWittmann @carlesarnal PTAL

**Comment by dolliecoder** [comment]
@jsenko fixed the copilot's suggestions PTAL

**Comment by dolliecoder** [comment]
@carlesarnal PR updated with the extra tests you requested on #8557: GraphQL schema → application/graphql YAML still works and edge cases covered. PTAL when you get a moment

**Review by carlesarnal — CHANGES_REQUESTED** [review]
Tests won't compile. Every test method has a file path as annotation instead of `@Test`. Fix this first.

The GraphQL detection heuristic is too simplistic. Checking for `contains("type ")` will match YAML documents like `type: Query` or any text with those words. After normalizing whitespace to single spaces, you don't need both `"schema{"` and `"schema {"` checks - pick one.

The `content.bytes()` call needs justification or removal. Comment says it's for materialization but doesn't explain why or what problem it solves.

Add test for YAML containing GraphQL keywords to prove it doesn't cause false positives. Current tests only check obvious non-GraphQL formats. Real risk is YAML like:
```yaml
type: Query
fields:
  - name: hello
```

Consider using a proper GraphQL parser for validation instead of keyword matching, or at least check that keywords appear in valid syntax positions (e.g., `type` followed by identifier and `{`).

Detection order matters. GraphQL before YAML means YAML-like GraphQL keywords could be misclassified. Document this decision or add safeguards.

dolliecoder:
@carlesarnal Fixed!

1. Tests: Cleaned up the file annotations 
2. GraphQL: Replaced basic contains() with strict regex (Pattern.compile) to enforce proper syntax positions.
3. YAML: Added your type: Query example to the tests; the regex correctly rejects it and falls back to YAML.
4. Bytes: Added a comment explaining that content.bytes() is needed because Jackson consumes the stream during the JSON check

jsenko:

dolliecoder:
@jsenko Thanks I've just rebased the branch onto main to pick up the FOSSA fix

schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
```diff
@@ -174,11 +175,40 @@ public static JsonNode parseJsonOrYaml(TypedContent content) throws IOException
         return node;
     }
 
-    // FIXME this doesn't work for GraphQL
+    private static final Pattern GRAPHQL_DEF_PATTERN = Pattern.compile("(?m)^\\s*(type|interface|scalar|union|input|enum)\\s+[a-zA-Z_][a-zA-Z0-9_]*\\s*(?:\\{|implements|\\=|$)");
```
paoloantinori: This pattern won't match GraphQL SDL extension syntax (`extend type Foo { ... }`, `extend enum Bar`, etc.), since those lines start with `extend` rather than the keyword itself. For a registry that may store extension-only schema fragments, that would still misclassify as YAML — the same bug this PR targets, just for the extension form. An optional `(?:extend\s+)?` prefix before the keyword group would close the gap. See https://spec.graphql.org/draft/#sec-Type-System.Extension

paoloantinori (commented):
Solid fix — the regex-based `isParsableGraphQL` is a clear improvement over the old `contains("type ")` heuristic, and the added tests (notably `testIsParsableGraphQL_rejectsYamlWithKeywords`) directly cover the original YAML-misclassification regression. A couple of observations:

- **GraphQL `extend` keyword is not detected.** `GRAPHQL_DEF_PATTERN` matches `type|interface|scalar|union|input|enum` only when one of those keywords starts a line, but standard GraphQL SDL extension syntax (`extend type Foo { ... }`, `extend enum Bar`, `extend input Baz`) starts with `extend` (see the [SDL extension grammar](https://spec.graphql.org/draft/#sec-Type-System.Extension)). A schema artifact registered as pure extensions would not match any of the three patterns and would still fall through to YAML detection — i.e. the exact bug this PR fixes, just for the extension form. An optional `(?:extend\s+)?` prefix before the keyword group would close the gap.
- **Test coverage for the new patterns.** `testIsParsableGraphQL` exercises the `schema {` and `type` paths, but neither the `directive @...` pattern nor an `extend`-prefixed definition is asserted. Adding one positive case each would lock in the intended behavior.

Neither point is a blocker; the core direction looks right.

schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
dolliecoder: Fixed added the (?:extend\s+)? prefix to perfectly catch those extension fragments

dolliecoder:
@paoloantinori thanks for catching that I've updated the regex and added the missing positive test cases for both edge cases PTAL

schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
paoloantinori: The `enum` keyword here is shared with Protobuf's `enum` definition. Since `determineContentType` checks GraphQL before protobuf, a `.proto` file with `enum Status { ... }` could be misclassified as `application/graphql`. Consider adding a negative test with Protobuf enum content, or narrowing the keyword set if this is a practical concern.

paoloantinori (commented):
Solid iteration on the GraphQL detection heuristic — the regex-based patterns and the `content.bytes()` stream-materialization guard are real improvements over the original `contains("type ")` approach, and the test suite (including the YAML-rejection case) is well targeted.

A few observations worth considering:

- **Protobuf `enum` collision**: The `GRAPHQL_DEF_PATTERN` includes `enum` as a type-system keyword, and `determineContentType` now checks GraphQL *before* the protobuf fallback. A Protobuf file containing an `enum` definition (e.g. `enum Status { ACTIVE = 0; }`) would match the regex and be misclassified as `application/graphql`. The existing Protobuf test only covers `message Hello { ... }`, which correctly doesn't match — but a standalone `enum` would. Worth either narrowing the keyword set (e.g. dropping `enum` and relying on the other patterns, since GraphQL enums are relatively uncommon as top-level definitions) or adding a negative test with a Protobuf enum to confirm whether this is a real regression in practice.

- **`content.bytes()` materialization**: The defensive call to force full stream materialization before the first `isParsableJson` check is sound and the comment is helpful. This forces every content-type detection call to fully buffer content into memory; for typical schema sizes this is negligible, but worth being aware of if `determineContentType` is ever called on large payloads.

- **`schema {` pattern requires block on same line**: `GRAPHQL_SCHEMA_PATTERN` expects `schema {` on one line. SDL also permits `schema\n{` on separate lines, which this wouldn't match. The single-line form is far more common, so this is likely fine in practice.

schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
dolliecoder: I removed enum from the regex and added a negative test to guarantee Protobuf enums fall through correctly

dolliecoder:
@paoloantinori I've removed enum from the regex to fix the Protobuf collision, and added test cases for both the Protobuf enum and the multi-line schema\n{ scenario PTAL

paoloantinori:
Thanks — dropping `enum` from `GRAPHQL_DEF_PATTERN` resolves the Protobuf `enum Status { ... }` collision, and the multi-line `schema\n{\n...}` case is now caught by `GRAPHQL_SCHEMA_PATTERN`. The negative tests (protobuf message, protobuf enum, JSON-like, XML-like, empty) give good confidence the YAML fallback still holds. No further concerns from me.

EricWittmann:
LGTM :+1:

PR #8559, opened 2026-07-13
fix(export): URL encode path elements in import/export ZIP entries

dolliecoder:
### Description
This PR resolves a known `TODO` in both `v2/EntityWriter` and `v3/EntityWriter` regarding the URL-encoding of path elements during ZIP export. 

**The Bug:**
Previously, if a user had a `groupId` or `artifactId` containing a slash (`/`) or other invalid filesystem characters, the `String.format` path construction would blindly append them. This created unintended nested directories, corrupting the internal directory structure of the export ZIP file and causing import/export failures. 

**The Fix:**
I have implemented `URLEncoder.encode(..., StandardCharsets.UTF_8)` for the `groupId`, `artifactId`, and `fileName` variables just before they are used to construct the `ZipEntry` paths. 

Because `EntityReader` relies entirely on parsing the embedded JSON contents (which retain the unencoded IDs) rather than extracting metadata from the file paths, this change is fully backward-compatible and safe.

### Motivation
Fixing structural data corruption in ZIP serialization and cleaning up acknowledged technical debt. 

### Checklist
- [x] Fixed `v2/EntityWriter.java`
- [x] Fixed `v3/EntityWriter.java`
- [x] Verified `utils/importexport` module compiles and tests pass successfully
- [x] Added `Developer's Certificate of Origin` sign-off to commits

### DIFF
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/ImportExportTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/ImportExportTest.java
@@ -67,6 +67,8 @@ public class ImportExportTest extends AbstractResourceTestBase {
          */
 
         String groupId = "PrimaryTestGroup";
+        String complexGroupId = "Group/With\"And Spaces";
+        String complexArtifactId = "Artifact/With\"And Spaces";
 
         // Create the group
         CreateGroup createGroup = new CreateGroup();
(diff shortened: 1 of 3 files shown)


dolliecoder:
@jsenko fixed the copilot's suggestions PTAL

carlesarnal:
@dolliecoder please, only one pr at a time. I'm closing this until https://github.com/Apicurio/apicurio-registry/pull/8558 is merged.

carlesarnal (changes requested):
Request changes.

`utils/importexport/src/main/java/io/apicurio/registry/utils/impexp/v2/EntityWriter.java` and `utils/importexport/src/main/java/io/apicurio/registry/utils/impexp/v3/EntityWriter.java` change the ZIP entry naming format. The fix makes sense, but this is a compatibility-sensitive area. I only see an app-level v3 round-trip test in `app/src/test/java/io/apicurio/registry/noprofile/rest/v3/ImportExportTest.java`.

Add coverage for v2 as well. This PR changes both v2 and v3 writers, but only v3 is exercised here.

Add edge-case tests for `%`, `+`, Unicode, and default group handling. `URLEncoder` semantics are close to what you want after normalization, but this is exactly the kind of code that breaks on reserved characters.

If the importer depends on parsing ZIP paths back into IDs, make sure decode behavior is explicitly covered. This patch is safe only if read and write stay symmetric.

PR #8546, opened 2026-07-12
fix(auth): use GroupOnly for getGroupById authorization

niy-ati:
Fixes #7866

getGroupById only takes groupId, so AuthorizedStyle.GroupAndArtifact can throw ArrayIndexOutOfBoundsException when the access controller reads a second parameter. Switched to GroupOnly (same as deleteGroupById) in v2 and v3.

Tests added (per review):
AbstractAccessControllerOwnerCheckTest : verifies getGroupById uses GroupOnly in v2/v3 and documents the single-parameter GroupAndArtifact failure
SimpleAuthTest#testGetGroupByIdGroupReadSufficientWithoutArtifactPermissions : readonly user can fetch group metadata without artifact owner permissions; artifact write remains forbidden

PR #8560, opened 2026-07-13
fix(protobuf): ignore google/protobuf imports in ALL_REFS_MAPPED integrity checks

### LINKED ISSUE #4877: Unmapped reference error for google import in protobuf schema
## Description
When adding a protobuf schema artifact with integrity check `ALL_REFS_MAPPED` enabled I get an "unmapped reference" violation exception because this reference is not specified. However the de-/serializers actually don't need that reference because they are importing the google referenced by themselves.

### Environment
**Registry Version**: 2.5.5

Registry is running with the Apicurio Operator in Openshift 4.14 / K8s v1.27.13

### Steps to Reproduce

1. Create protobuf schema with a google import
```
syntax = "proto3";

import "google/protobuf/descriptor.proto";

extend google.protobuf.FileOptions {
  string version = 99999;
}
```
3. enable integrity check ALL_REFS_MAPPED
4. Upload schema

google/protobuf/descriptor.proto is reported with the error "Unmapped reference detected".

Since the de-/serializers do not need that reference to be able to work with the schema I would expect that the google imports are excluded from the reference mapping check.

niy-ati:
## description
fixes #4877

`ALL_REFS_MAPPED` was treating well-known `google/protobuf/*` imports (descriptor, timestamp, etc.) as unmapped references. those come from the protobuf runtime, so they shouldn't need registry mappings.

the integrity path was parsing imports itself, while `ProtobufReferenceFinder` already filters `google/protobuf/`. aligned them by reusing the finder in `ProtobufContentValidator.validateReferences` (same pattern as avro).

## test plan
- [x] `./mvnw -pl schema-util/util-provider -am test -Dtest=ProtobufContentValidatorTest -Dsurefire.failIfNoSpecifiedTests=false`
- [x] schema with only google/protobuf imports + empty mappings does not fail
- [x] custom imports still require mappings when mixed with google imports

### DIFF
--- a/schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufContentValidator.java
+++ b/schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufContentValidator.java
@@ -4,6 +4,8 @@ import com.squareup.wire.schema.SchemaException;
 import com.squareup.wire.schema.internal.parser.MessageElement;
 import com.squareup.wire.schema.internal.parser.ProtoFileElement;
 import io.apicurio.registry.content.TypedContent;
+import io.apicurio.registry.content.refs.ExternalReference;
+import io.apicurio.registry.protobuf.content.refs.ProtobufReferenceFinder;
 import io.apicurio.registry.rest.v3.beans.ArtifactReference;
 import io.apicurio.registry.rules.validity.AbstractContentValidator;
 import io.apicurio.registry.rules.validity.ValidityLevel;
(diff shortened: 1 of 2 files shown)


carlesarnal:

carlesarnal (changes requested):
The fix correctly addresses issue #4877 by reusing `ProtobufReferenceFinder` to filter well-known google/protobuf imports. This creates consistency with auto-reference discovery behavior.

Critical problem: Both test methods have corrupted annotations. Line 782 and 804 in `schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/ProtobufContentValidatorTest.java` show file paths instead of `@Test`. This will not compile.

Fix the annotations:
```java
@Test
public void testValidateReferencesIgnoresGoogleProtobufImports() throws Exception {
```

and

```java
@Test
public void testValidateReferencesStillRequiresCustomImportsWithGoogleOnes() throws Exception {
```

After fixing annotations, the change is good. The logic is correct and test coverage is adequate.

niy-ati:
reopening now that #8546 is merged. addressed the @Test annotation review on 1a3b220 ,both regression tests pass locally.

PR #8581, opened 2026-07-13
fix(storage): include artifact references in gitops/kubernetesops content hashes

### LINKED ISSUE #8540: GitOps/KubernetesOps storage hash diverges SQL storage
## Description

  **Persistence type**: gitops / kubernetesops (compared against sql)

  The polling-based storage variants (`gitops`, `kubernetesops`) compute both `contentHash` and
  `canonicalHash` while ignoring artifact references, even when those references are present and
  stored. The `sql` storage includes references in both hashes. As a result, identical
  content + references produce **different** hashes depending on which storage backend loaded them.

  ### Environment

  Storage backends: `gitops` and `kubernetesops` (both extend the polling storage). Reproduces on any
  platform; not client- or serdes-specific. Compared against the `sql` backend.

  ### Steps to Reproduce

  1. In a `sql`-backed registry, register an Avro (or Protobuf) schema that has at least one artifact
     reference. Note the stored `contentHash` and `canonicalHash` for the version's content.
  2. Export that artifact (including its references) and load the same content + references through a
     `gitops` (or `kubernetesops`) instance.
  3. Query the stored `contentHash` / `canonicalHash` on the polling-backed instance.
  4. Observe that both hashes differ from the values produced by the `sql` backend for the same
     content and references.

  **Expected:** All storage variants compute identical `contentHash` and `canonicalHash` for the same
  content and the same set of references.

  **Actual:** `gitops`/`kubernetesops` compute both hashes as if the content had no references, so the
  values diverge from `sql` whenever the content has references.

niy-ati:
## Summary
- fixes #8540
- polling storages (`gitops` / `kubernetesops`) were computing `contentHash` and `canonicalHash` with `null` references, then attaching refs only when building the `ContentEntity`
- SQL includes serialized references in both hashes, so identical content+refs produced different hashes across backends
- now extract refs from content-metadata first, hash with them (same as SQL), and resolve refs via `storage::getContentByReference` for canonical hashing

## Test plan
- [x] `RegistryStorageContentUtilsTest` — contentHash changes when refs are present
- [x] `AbstractPollingDataSourceManagerReferencesTest` — extractReferences mapping + sql-style hash match

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/polling/AbstractPollingDataSourceManager.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/polling/AbstractPollingDataSourceManager.java
@@ -431,8 +431,12 @@ public abstract class AbstractPollingDataSourceManager<MARKER extends SourceMark
             // Determine artifact type from content if not specified
             String resolvedArtifactType = utils.determineArtifactType(typedContent, artifactType);
 
-            // Calculate content hash for deduplication
-            String contentHash = utils.getContentHash(typedContent, null);
+            // Build references before hashing so gitops/kubernetesops match SQL storage
+            // (both contentHash and canonicalHash include serialized references when present).
+            List<ArtifactReferenceDto> references = extractReferences(contentMetadata);
+
+            // Calculate content hash for deduplication (must include references when present)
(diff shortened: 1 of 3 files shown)


carlesarnal:

Issue #6298, opened 2025-06-06
Artifact search returns incorrect artifacts


jonsnowseven:
SDK client returns an artifact incorrectly if the artifact name **is contained in an existing one**.

## Description

**Registry
Version**: 3.0.9 
**Persistence type**: `sql` 

### Steps to Reproduce

1. Register an AVRO artifact with group id `group` and name `name-test`
2. Run
    ```java
     client.search().artifacts().get(
                getRequestConfiguration -> {
                  checkNotNull(getRequestConfiguration.queryParameters);
                  getRequestConfiguration.queryParameters.name = "name";
                  getRequestConfiguration.queryParameters.groupId = "group";
                });
      ```
3. It returns the artifact with group id `group` and name `name-test`

### Expected vs Actual Behaviour

The SDK should not return any result but returns an artifact.

EricWittmann:
I think this is actually working as intended, but I agree could be improved.  Some of the filters (query params) on the search endpoints are intended to be "LIKE" queries essentially.  That said, it would be nice if the filtering supported both `=` and `LIKE` queries, based on the value of the query param.  So perhaps a value of `*name*` would result in a `LIKE` query, but a value of `name` would result in an exact match.

Thoughts?

jonsnowseven:
Thank you for the detailed answer!
I should have been more specific: I meant what would change from the client's perspective (if anything)?

EricWittmann:
Sorry I never answered this.  From the client perspective, you could still do this:

```java
 client.search().artifacts().get(
            getRequestConfiguration -> {
              checkNotNull(getRequestConfiguration.queryParameters);
              getRequestConfiguration.queryParameters.name = "name";
              getRequestConfiguration.queryParameters.groupId = "group";
            });
```

But it would only return artifacts with a name of exactly "name" AND a group of exactly "group".  But you could then also do this:

```java
 client.search().artifacts().get(
            getRequestConfiguration -> {
              checkNotNull(getRequestConfiguration.queryParameters);
              getRequestConfiguration.queryParameters.name = "name*";
              getRequestConfiguration.queryParameters.groupId = "group";
            });
```

Now with the `*` wildcard in the name query param, this would return all artifacts with a group of exactly "group" but with a name that **contains** the word "name".

PR #8587, opened 2026-07-13
fix(search): make version name search match artifact name behavior

### LINKED ISSUE #8002: Search API: inconsistent matching behavior across name fields
## Description

The search API has inconsistent matching behavior across different entity types:

- **Artifact `name`**: searches both the `name` field and `artifactId` field. User must use `*` wildcard as prefix/suffix for partial matching, otherwise matches exactly.
- **Version `name`**: searches as a substring match, and treats `*` literally (not as a wildcard).
- **`description`**: matches as a substring for all entity types.
- **Labels**: matched exactly, supporting `key`, `key:`, or `key:value` formats.
- **Other fields**: matched exactly.

## Expected behavior

The matching behavior for `name` fields should be consistent across artifacts and versions (either both use wildcard syntax or both use substring matching).

## Found during

PR #7973 — CLI search command implementation.

Me-Priyank:
Fixes #8002

## What

Version search with a `name` filter matches as a substring (`LIKE '%value%'`) and treats `*` literally, whereas artifact search matches `name` exactly, with optional `*` prefix/suffix wildcards, against both the `name` and `artifactId` fields. The same conceptual filter behaves differently across the two endpoints.

This aligns version `name` search with artifact `name` search.

## Why

The artifact-side behavior was established in #6298 / #6868, which deliberately moved artifact name search away from substring matching (that issue describes substring matching as returning incorrect results). The correction was applied to artifact search only; version search was never updated, which is what #8002 tracks.

The Elasticsearch backend (`ElasticsearchSearchService.buildNameQuery`) already searches name + artifactId for versions, so this change is limited to the SQL/default backend. Note that ES matches `name` with full-text search, a separate, pre-existing difference from SQL that applies to artifacts too, and is intentionally left as-is here.

## Changes

- **`SqlSearchRepository`**: extract the artifact name-matching logic into a shared `buildNameClause(...)` helper and use it for **both** artifact and version `name` filters (matched over `name` OR `artifactId`). Split the version `name`/`description` case so `description` keeps its substring behavior, it is already consistent across artifacts and versions and is out of scope. The helper also guards a wildcard-only value (e.g. `name=*`) so it matches everything rather than throwing (the previous inline artifact code would `StringIndexOutOfBoundsException` on `*`).
- **`VersionSearchTest`**: add `testVersionNameSearchWildcards`, mirroring `ArtifactSearchTest#testArtifactNameSearchWildcards` (exact, `name*`, `*name`, `*name*`, no-match, and match-by-artifactId).
- **`SearchVersionsCommand`** (CLI): update the `--name` help text to describe the new matching behavior (it previously said "substring match").

No changes to the REST API surface, the OpenAPI schema, or the Elasticsearch path (which already searches name + artifactId for versions; its full-text matching is unchanged).

## Behavior change

`GET /search/versions?name=…` now matches exactly (or by `*` wildcard) instead of as a substring, consistent with `GET /search/artifacts?name=…`. This is the intended correction per #6298; substring name matching was treated as a bug.

## Testing

- Added `VersionSearchTest#testVersionNameSearchWildcards` (exact, `name*`, `*name`, `*name*`, no-match, match-by-artifactId, and wildcard-only `*`).
- `VersionSearchTest`, `ArtifactSearchTest`, `SearchVersionsTest`, `SearchArtifactsTest` pass locally (in-memory / SQL storage). CI covers the other storage variants.


EricWittmann (dismissed):
Thanks for the contribution, @Me-Priyank! Clean refactoring and solid test coverage. LGTM!

Me-Priyank:
Thanks for reviewing @EricWittmann ... Pushed a fix for the failing `app-other` CI shard. `ImportReindexSearchTest` searched version `name = "Pet Store"` and relied on the old substring matching to hit the version named `"Pet Store API"`

EricWittmann:
:+1:

PR #8598, opened 2026-07-13
fix(storage): implement semantic version sorting for artifact versions

### LINKED ISSUE #7871: [Artifact Versions] sorting by version is not respecting semantic version
## Description

**Registry
Version**: 3.2.x
**Persistence type**: gitops

### Environment

does not matter

### Steps to Reproduce

Navigate on UI to Artifact Versions and click on version header sorting

Actual Behaviour:
When I sort by version in table, i get 

- 1.0.0
- 1.0.1
- 1.0.10
- 1.0.11
- 1.0.2
- ...

Expected Behaviour:
When I sort by version in table, i want get 

- 1.0.0
- 1.0.1
- 1.0.2
- ...
- 1.0.10
- 1.0.11

### Logs

not relevant here

raghavgupta000001:
Fixes #7871

**What changed:**
- Added a `compareVersions` helper method in `AbstractSqlRegistryStorage.java` to parse and numerically sort semantic version strings.
- Applied the custom SemVer sort when `OrderBy.version` is requested.
- Verified local sorting behavior for ascending and descending requests.

**Signed-off-by:** Raghav Gupta raghav.gupta000001@gmail.com

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
@@ -1014,13 +1014,23 @@ public abstract class AbstractSqlRegistryStorage implements RegistryStorage {
         return versionRepository.getArtifactVersions(groupId, artifactId, filterBy);
     }
 
-    @Override
-    public VersionSearchResultsDto searchVersions(Set<SearchFilter> filters, OrderBy orderBy,
-            OrderDirection orderDirection, int offset, int limit, boolean skipCount)
-            throws RegistryStorageException {
+@Override
+public VersionSearchResultsDto searchVersions(Set<SearchFilter> filters, OrderBy orderBy,
+        OrderDirection orderDirection, int offset, int limit, boolean skipCount)
+        throws RegistryStorageException {
 
-        return searchRepository.searchVersions(filters, orderBy, orderDirection, offset, limit, skipCount);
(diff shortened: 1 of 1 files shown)


EricWittmann:
I like the idea, but this implementation is sorting **after** pagination.  That is not going to work. What will happen is that the database will order the results and return the first page of results.  And THEN this code will run and sort by semver within page 1.

We'll need a different architectural approach to this that pushes semver based sorting to the DB layer.  Any ideas?

raghavgupta000001:
> I like the idea, but this implementation is sorting **after** pagination. That is not going to work. What will happen is that the database will order the results and return the first page of results. And THEN this code will run and sort by semver within page 1.
> 
> We'll need a different architectural approach to this that pushes semver based sorting to the DB layer. Any ideas?

@EricWittmann  are completely right about the pagination issue with the in-memory sorting. I agreed we needed to push this to the DB layer, so I took a completely new architectural approach to solve it.

I have introduced a versionSortKey column directly into the versions database table. By generating a sortable string on insertion, this allows the database itself to handle the semantic sorting natively in the query, long before pagination happens.

I have this working locally with the H2 database and opened a Draft PR with the core logic here: https://github.com/Apicurio/apicurio-registry/pull/8627

Before I finalize it, my plan is to:
1. Update the dialect-specific SQL files (PostgreSQL, MySQL, etc.) to handle the new column.
2. Write the database migration scripts (.ddl / ALTER TABLE) to ensure existing deployments upgrade smoothly.

Does this new database-layer approach look good to you?

PR #8601, opened 2026-07-13
fix(version): improve file extension detection based on content type

### LINKED ISSUE #8172: Ensure Content Download honors Content MIME-TYPE and it's file extension
## Description

**Registry Version**: >= 3.2
**Persistence type**: *

Artifact Content Download of an OpenAPI when YAML is selected, downloads a OpenAPI with valid YAML content but filename is always .json

I'm expecting that dependent on selected content type, the downloaded file has the right file extension.

JSON: ".json"
YAML: ".yaml"
XML: ".xml"
...

kartikangiras:
## Description

The artifact content download on the version page always used .json as the file extension, even when the content was YAML or a different format. This happened because the extension was derived solely from the artifact type, which defaults to json for types like OpenAPI, AsyncAPI, JSON Schema, and Avro, all of which can be stored in either JSON or YAML format.

This fix uses the existing detectContentType utility to inspect the actual content string and determine the correct file extension at download time.
## Changes:
- Added fileExtensionForContentType() helper in content.utils.ts so maps content type strings to file extensions (e.g., application/x-yaml → yaml, application/json → json, application/xml → xml, etc.)
- Refactored fileExtensionForDraft() to reuse the new helper
- Updated doDownloadVersion() in VersionPage.tsx to detect content type from the content string and apply the correct extension, while preserving special-case extensions for WSDL (.wsdl) and XSD (.xsd)

### Issue
fixes #8172

### DIFF
--- a/ui/ui-app/src/app/pages/version/VersionPage.tsx
+++ b/ui/ui-app/src/app/pages/version/VersionPage.tsx
@@ -25,7 +25,7 @@ import {
     MetaData,
     RootPageHeader
 } from "@app/components";
-import { ContentTypes } from "@models/ContentTypes.ts";
+import { detectContentType, fileExtensionForContentType } from "@utils/content.utils.ts";
 import { PleaseWaitModal } from "@apicurio/common-ui-components";
 import { AppNavigation, useAppNavigation } from "@services/useAppNavigation.ts";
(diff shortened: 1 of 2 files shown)


carlesarnal:
@kartikangiras please, only one pr at a time. I'm closing this until https://github.com/Apicurio/apicurio-registry/pull/8545 is merged.

Issue #3662, opened 2023-09-15
Apicurio registry 2.4.3 and 2.4.4 causes endless rebalance loop on Kafka


obabec:
Apicurio version: 2.4.3 and 2.4.4
Openshift version: 4.12
Kafka version: 3.4.0, 3.5.0
Persistent type: kafkasql
Convertor: SerdeAvro

The issue is reproducible when running the Debezium test suite against openshift 4.12. The test suite uses installation via OLM so you have to make the necessary changes that would allow you to install the operator via operator bundle to hit the issue.

Once tests connect the consumer to the plain listener on Kafka (the same one that Apicurio uses) the consumer falls into an endless rebalance loop. When we run tests in different ways via external TLS listeners everything works. We have tried version 2.3.0, 2.4.0, 2.4.1, 2.4.2, 2.4.3, 2.4.4. The issue occurs in version >= 2.4.3.

The issue is 100% reproducible in this environment.

This issue is just a follow-up to discussions that started on Monday.

novotnyJiri:
Additional info, due to my mistake in testing runs we wrongly assumed TLS on version 2.4.3 works. That is not true. It only works on version 2.4.2 and probably all earlier versions.

carlesarnal:
@jsenko I think this has been addressed already, no?

jsenko:
Not yet

jcechace:
Any progress on this? Downstream wise this is actually a pretty significant issue.

carlesarnal:
Due to a bug in our converter implementation where the default for enabling the headers was mistakenly set to true, in `2.4.3.Final` the converter started using headers instead of sending the magic byte and the schemaId in the message payload. To revert that behaviour, `apicurio.registry.headers.enabled` has to be set to false in the connector configuration. That is a workaround, not a fix. In 3.0, this has been addressed and the default value has been correctly set. We cannot change it back in 2.6.x since there are existing consumers relying on this, so our best option for older versions is very likely to document this.

jcechace:
@obabec  ^^

malladinagarjuna2:
hey @carlesarnal  @jsenko @EricWittmann 
i ghave worked on this issue and have find the roor cause and a fix 
[...]

PR #8604, opened 2026-07-13
fix(security): add missing HSTS headers to error pages (#2411)

### LINKED ISSUE #2411: Add missing HSTS headers to error pages

malladinagarjuna2:
Fixes #2411

## Problem
`HSTSFilter` sets `Strict-Transport-Security` via the servlet chain, but
`RegistryApplicationServletFilter` runs before it: on a disabled-API hit it calls
`response.reset()` (wiping all headers), sets 404, and returns early — so `HSTSFilter`
never runs and that 404 ships with no headers.

## Fix
- Re-apply HSTS after `reset()` using the existing (previously unused)
  `HSTSFilter.addHstsHeaders()` helper.
- Add `<dispatcher>ERROR</dispatcher>` to the `HSTSFilter` mapping so container-generated
  error pages get it too.

## Testing
New `HstsHeaderTest` checks the header on 5 responses (200, mapper 404, unknown-path 404,
disabled-API 404, 400). Written first: it failed only on the disabled-API 404 (zero headers
returned); all 5 pass with the fix. `DisableApisFlagsTest` and checkstyle still pass.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/RegistryApplicationServletFilter.java
+++ b/app/src/main/java/io/apicurio/registry/rest/RegistryApplicationServletFilter.java
@@ -1,6 +1,7 @@
 package io.apicurio.registry.rest;
 
 import io.apicurio.registry.services.DisabledApisMatcherService;
+import io.apicurio.registry.ui.servlets.HSTSFilter;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.servlet.Filter;
@@ -43,6 +44,7 @@ public class RegistryApplicationServletFilter implements Filter {
                 HttpServletResponse httpResponse = (HttpServletResponse) response;
                 httpResponse.reset();
                 httpResponse.setStatus(HttpServletResponse.SC_NOT_FOUND);
+                HSTSFilter.addHstsHeaders(httpResponse);
                 // important to return, to stop the filters chain
                 return;
             }
--- a/app/src/test/java/io/apicurio/registry/headers/HstsAuthHeaderTest.java
+++ b/app/src/test/java/io/apicurio/registry/headers/HstsAuthHeaderTest.java
@@ -1,0 +1,55 @@
+package io.apicurio.registry.headers;
+
+import io.apicurio.registry.AbstractResourceTestBase;
+import io.apicurio.registry.types.ArtifactType;
+import io.apicurio.registry.types.ContentTypes;
+import io.apicurio.registry.utils.tests.ApicurioTestTags;
+import io.apicurio.registry.utils.tests.ProxyHeaderAuthTestProfile;
+import io.apicurio.registry.utils.tests.TestUtils;
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import io.restassured.http.ContentType;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.containsString;
+
+// #2411: 401/403 flow through the normal filter chain (unlike the disabled-API 404), so HSTSFilter
+// already covers them - this locks that in.
+@QuarkusTest
+@TestProfile(ProxyHeaderAuthTestProfile.class)
+@Tag(ApicurioTestTags.SLOW)
+public class HstsAuthHeaderTest extends AbstractResourceTestBase {
+
+    private static final String HSTS = "Strict-Transport-Security";
+    private static final String ARTIFACT_CONTENT = "{\"name\":\"redhat\"}";
+
+    private static final String HEADER_USERNAME = "X-Forwarded-User";
+    private static final String HEADER_EMAIL = "X-Forwarded-Email";
+
+    @Override
+    protected void deleteGlobalRules(int expectedDefaultRulesCount) throws Exception {
+        // Skip: the base-class cleanup would fail with a 401 here, since it sends no proxy auth headers.
+    }
+
+    @Test
+    public void testHstsOnUnauthorized401() {
+        given().when().get("/registry/v3/search/artifacts").then().statusCode(401).header(HSTS,
+                containsString("max-age="));
+    }
+
+    @Test
+    public void testHstsOnForbidden403() {
+        String groupId = TestUtils.generateGroupId();
+        String artifactId = TestUtils.generateArtifactId();
+
+        // Authenticated but no roles (no groups header) -> 403.
+        given().header(HEADER_USERNAME, "testuser").header(HEADER_EMAIL, "testuser@example.com")
+                .contentType(ContentType.JSON)
+                .body(TestUtils.clientCreateArtifact(artifactId, ArtifactType.JSON, ARTIFACT_CONTENT,
+                        ContentTypes.APPLICATION_JSON))
+                .when().pathParam("groupId", groupId).post("/registry/v3/groups/{groupId}/artifacts").then()
+                .statusCode(403).header(HSTS, containsString("max-age="));
+    }
+}
--- a/app/src/test/java/io/apicurio/registry/headers/HstsBoomResource.java
+++ b/app/src/test/java/io/apicurio/registry/headers/HstsBoomResource.java
@@ -1,0 +1,20 @@
+package io.apicurio.registry.headers;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Alternative;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+
+// Test-only endpoint that always throws, so HstsError500Test can assert the HSTS header on a real 500.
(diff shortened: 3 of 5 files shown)


malladinagarjuna2:
@carlesarnal  @EricWittmann could you please review this pr

malladinagarjuna2:
@EricWittmann  could you please accept this pr ?

carlesarnal (changes requested):
Implementation is correct — adding `HSTSFilter.addHstsHeaders(httpResponse)` to the 404 path in `RegistryApplicationServletFilter` and adding the ERROR dispatcher to web.xml closes the security gap properly.

Commit message needs conventional commit format: `fix(security): add missing HSTS headers to error pages (#2411)`.

Test coverage looks good for the 200/404 cases. Could also verify 500/403/401, but not blocking.

malladinagarjuna2:
ok sir, sure will do that

malladinagarjuna2:
hey @carlesarnal  thanks for the review,
added the 401/403/500 cases you suggested:
HstsAuthHeaderTest — 401 (unauthenticated) and 403 (authenticated, no role)
HstsError500Test — 500, via a small test-only endpoint that throws, since a 500 has no natural trigger
i wasnt able to check the 500 tests locally.

malladinagarjuna2:
@EricWittmann  @carlesarnal could you please review the pr ?

malladinagarjuna2:
@carlesarnal  all checks are passing except the  licence compilance, i think the license compilance is not passing for every pr so i think this is ready for review, if i am wrong kindly correct me

jsenko:

malladinagarjuna2:
hey @carlesarnal  @jsenko could you please accept the pr

jsenko:
> hey @carlesarnal @jsenko could you please accept the pr

Done, workflows should run now.

EricWittmann:
## Suggestion: Scope `HstsBoomResource` to its test profile

`HstsBoomResource` is in `src/test/java` so it won't ship in production, but since it's an `@ApplicationScoped` JAX-RS resource, it's active during **all** `@QuarkusTest` runs in the `app` module — not just `HstsError500Test`. A stray endpoint producing 500s could confuse future test debugging.

You can gate it behind a test profile using `@Alternative` + `getEnabledAlternatives()`:

```java
@Alternative
@Path("/apis/test/hsts-500-boom")
@ApplicationScoped
public class HstsBoomResource {

    @GET
    public String boom() {
        throw new RuntimeException("Intentional failure to verify HSTS headers on 500 responses.");
    }
}
```

```java
@QuarkusTest
@TestProfile(HstsError500Test.BoomProfile.class)
public class HstsError500Test {

    public static class BoomProfile implements QuarkusTestProfile {
        @Override
        public Set<Class<?>> getEnabledAlternatives() {
            return Set.of(HstsBoomResource.class);
        }
    }

    // ... tests unchanged
}
```

This way the endpoint only exists when `HstsError500Test` runs. Non-blocking — the fix itself looks good.

app/src/main/resources-unfiltered/META-INF/web.xml
```diff
@@ -46,6 +46,7 @@
         <url-pattern>/*</url-pattern>
         <dispatcher>REQUEST</dispatcher>
         <dispatcher>FORWARD</dispatcher>
+        <dispatcher>ERROR</dispatcher>
```
paoloantinori: This ERROR dispatcher is not exercised by any test in the PR — `HstsError500Test` triggers a 500 via a thrown JAX-RS exception, which does not go through the ERROR `RequestDispatcher`. Either add a test that genuinely hits the ERROR path, or drop the line if it turns out to be a no-op given the programmatic fix above.

app/src/main/java/io/apicurio/registry/rest/RegistryApplicationServletFilter.java
```diff
@@ -43,6 +44,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 HttpServletResponse httpResponse = (HttpServletResponse) response;
                 httpResponse.reset();
                 httpResponse.setStatus(HttpServletResponse.SC_NOT_FOUND);
+                HSTSFilter.addHstsHeaders(httpResponse);
```
paoloantinori, line 47: Imports `HSTSFilter` from the `ui.servlets` package — app-layer to UI-layer coupling. Functional, but architecturally a little surprising; see the body note about a follow-up.

paoloantinori (commented):
On top of @carlesarnal's and @EricWittmann's good notes — the core security-gap fix is correct: calling `HSTSFilter.addHstsHeaders(httpResponse)` after `response.reset()` on the disabled-API path is the right way to restore the header the filter chain never reaches there (`RegistryApplicationServletFilter` short-circuits with `return` before `HSTSFilter` runs, and `reset()` already wiped any header set earlier in the chain). Two observations worth considering before merge:

- **The new `<dispatcher>ERROR</dispatcher>` in the `HSTSFilter` mapping (`web.xml`) is not exercised by any test in this PR.** `HstsError500Test` produces a 500 via a JAX-RS endpoint that throws — that flows through the normal `REQUEST` chain (RESTEasy maps the exception to 500), not through a container `ERROR` dispatch. So whether the ERROR dispatcher is actually needed, or is dead config, is not demonstrated. If the intent is to cover container-generated static error pages (e.g. a default 404 with no mapped resource), a test that genuinely triggers one would lock that in; otherwise the programmatic fix alone already covers the cases you have listed.
- `RegistryApplicationServletFilter` now reaches into `io.apicurio.registry.ui.servlets.HSTSFilter` — a cross-layer coupling (REST/app layer depending on a UI-layer servlet class). Pragmatic, but if a second security header ever needs to survive `reset()`, the "re-apply explicitly here" pattern will not scale. Worth a thought for a follow-up, not blocking.

malladinagarjuna2:
@EricWittmann @carlesarnal @jsenko could you oplease accept the pr

jsenko:
## Review

### Prior feedback status

All prior review feedback has been addressed:

| Reviewer | Feedback | Status |
|----------|----------|--------|
| @carlesarnal | Conventional commit format | ✅ PR title updated |
| @carlesarnal | Test 500/403/401 | ✅ `HstsAuthHeaderTest` (401/403) and `HstsError500Test` (500) added |
| @EricWittmann | Scope `HstsBoomResource` with `@Alternative` + `getEnabledAlternatives()` | ✅ Implemented exactly as suggested |
| @paoloantinori | `<dispatcher>ERROR</dispatcher>` not tested | ✅ Correctly removed — the web.xml change was dead config since no container `<error-page>` mappings exist |
| @paoloantinori | Cross-layer coupling (`rest` → `ui.servlets.HSTSFilter`) | ℹ️ Acknowledged non-blocking, no action needed for this PR |

### Fix assessment

The core fix is correct and minimal: after `response.reset()` wipes all headers in `RegistryApplicationServletFilter`, `HSTSFilter.addHstsHeaders(httpResponse)` restores the HSTS header. `addHstsHeaders()` is a simple static method that sets a single header — no side effects.

There are no other `response.reset()` calls in the app module, so this is the only code path that needed the fix.

### Test coverage

Good coverage across all relevant HTTP status codes:

- `HstsHeaderTest`: 200, mapper-404, unknown-path-404, **disabled-API-404** (the actual bug), 400
- `HstsAuthHeaderTest`: 401, 403 (using existing `ProxyHeaderAuthTestProfile`)
- `HstsError500Test`: 500 (via `@Alternative`-scoped `HstsBoomResource`)

### New observation (non-blocking, separate issue)

`RedirectFilter` has the same bug pattern. It runs **first** in the filter chain (see `web.xml` ordering: RedirectFilter → RegistryApplicationServletFilter → ResourceCacheControlFilter → HSTSFilter). When `apicurio.redirects.enabled=true` and a path matches, `RedirectFilter` calls `response.sendRedirect()` and returns early — `HSTSFilter` never runs, so the 302 response has no HSTS header.

This is outside the scope of this PR (disabled-API 404 fix), but worth tracking as a follow-up since HSTS on redirects is important (it's how browsers learn to upgrade HTTP → HTTPS before the redirect even happens).

malladinagarjuna2:
@jsenko should i raise a different issue for redirect filter or should i implement it in the pr ?

malladinagarjuna2:
hey @jsenko  i have given a soltuion for the bug https://github.com/Apicurio/apicurio-registry/issues/3662
could you please review that solution

paoloantinori:
Thanks @jsenko for the detailed summary — confirming all three items from my earlier review are addressed on this head: the `HSTSFilter.addHstsHeaders(httpResponse)` call after `response.reset()` on the disabled-API path, the `<dispatcher>ERROR</dispatcher>` addition for container error pages, and the scoped `@Alternative` + `getEnabledAlternatives()` pattern for `HstsBoomResource`. The test coverage (200, 404×3, 400, plus 401/403/500) is thorough. Looks good from my side.

jsenko (approved):

jsenko:
> @jsenko should i raise a different issue for redirect filter or should i implement it in the pr ?

Can you create a new issue for that? Thanks. I'll merge this when the tests pass again.

malladinagarjuna2:
alright @jsenko

PR #8606, opened 2026-07-14
 add missing X-Content-Type-Options: nosniff header (#2412)

### LINKED ISSUE #2412: Add missing X-Content-Type-Options: nosniff header

malladinagarjuna2:
Fixes #2412

## Problem
Registry responses never set `X-Content-Type-Options`, so browsers are free to MIME-sniff
a response body away from its declared content type.

## Fix
- New `ContentTypeOptionsFilter`, mapped to `/*`, sets `X-Content-Type-Options: nosniff`
  on all responses (`REQUEST`/`FORWARD`/`ERROR` dispatchers).
- `RegistryApplicationServletFilter` re-applies the header on the disabled-API path: it
  calls `response.reset()` (which wipes all headers) and returns early, skipping the rest
  of the filter chain — so the header must be set explicitly there.

## Testing
New `ContentTypeOptionsHeaderTest` asserts the header on 5 responses: 200, mapper 404,
unknown-path 404, disabled-API 404, and 400. Written first — all 5 failed before the fix
(header absent everywhere); all 5 pass now. `DisableApisFlagsTest` still passes and
checkstyle is clean.

Also verified at runtime in dev mode: `GET /apis/registry/v3/system/info` with
`apicurio.disable.apis` set now returns `404` **with** `X-Content-Type-Options: nosniff`
(previously the response carried no headers but `Content-Length: 0`).


carlesarnal:
@malladinagarjuna2 please, only one pr at a time. I'm closing this until https://github.com/Apicurio/apicurio-registry/pull/8604 is merged.

app/src/main/java/io/apicurio/registry/ui/servlets/ContentTypeOptionsFilter.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public static void addContentTypeOptionsHeaders(HttpServletResponse httpResponse) {
+        httpResponse.setHeader("X-Content-Type-Options", CONTENT_TYPE_OPTIONS_HEADER);
+    }
+
+    @Override
+    public void init(FilterConfig config) throws ServletException {
+    }
+
+    @Override
```
paoloantinori, line 34: Setting the header here (before `chain.doFilter()`) means any downstream `response.reset()` wipes it — which is exactly why you had to re-add it manually in `RegistryApplicationServletFilter`. An alternative that's reset-proof: move this into a `try { chain.doFilter(...) } finally { addContentTypeOptionsHeaders(httpResponse); }`, so the header is applied on the way back out regardless of what downstream did. Worth considering if there turn out to be other `reset()` paths.

app/src/main/java/io/apicurio/registry/rest/RegistryApplicationServletFilter.java
```diff
@@ -43,6 +44,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
                 HttpServletResponse httpResponse = (HttpServletResponse) response;
                 httpResponse.reset();
                 httpResponse.setStatus(HttpServletResponse.SC_NOT_FOUND);
+                ContentTypeOptionsFilter.addContentTypeOptionsHeaders(httpResponse);
```
paoloantinori: Good catch that this path needs explicit handling — `response.reset()` immediately above clears all headers, so the filter-chain-set header would be lost. This call restores it correctly. Just confirming the intent is intentional and documented in the PR body, which it is.

paoloantinori (commented):
Clean, well-scoped security fix — a single `/*`-mapped filter plus the explicit re-application on the disabled-API path (where `response.reset()` would otherwise wipe the header) is the right shape, and the 5-case test covers the meaningful response classes (200, mapper 404, unknown-path 404, disabled-API 404, 400). A couple of things worth double-checking:

- **Header-erasure surfaces beyond the disabled-API path.** `addContentTypeOptionsHeaders()` runs *before* `chain.doFilter()`, so any downstream component that calls `response.reset()` (or `setHeader("X-Content-Type-Options", ...)`) would drop or overwrite it. The disabled-API branch is handled explicitly; it's worth a quick grep for other `response.reset()`/`sendError`-with-fresh-response calls in the filter/servlet chain to make sure there isn't a second path with the same shape. If there is, the more robust pattern is to set the header in a `try/finally` after `chain.doFilter()` (i.e. on the way back out), which survives resets downstream.
- **`mergeable_state: dirty`.** This won't merge cleanly onto `main` right now — needs a rebase before it can land. Worth resolving before `ready`.
- Minor: the test's `@TestProfile` sets `apicurio.disable.apis=/apis/registry/v3/system/.*` at the class level, so it applies to all 5 cases. That's correct for the cases here (only `testNoSniffOnDisabledApi404` actually needs it), but if more cases are added later that assume full API availability, they'd silently inherit the disabled profile. Fine as-is; flagging for future-awareness.

EricWittmann:
@malladinagarjuna2 Unfortunately there are some conflicts that need to be fixed now!  :t-rex:

malladinagarjuna2:
@EricWittmann  will definetly fix it now .

app/src/main/java/io/apicurio/registry/ui/servlets/ContentTypeOptionsFilter.java
```diff
@@ -0,0 +1,43 @@
...
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        HttpServletResponse httpResponse = (HttpServletResponse) response;
```
paoloantinori, line 37: Because the header is set **before** `chain.doFilter()`, any downstream filter that calls `response.reset()` (the disabled-API 404 path in `RegistryApplicationServletFilter`) wipes it — which is exactly why `resetKeepingHeaders(...)` has to re-list `X-Content-Type-Options` explicitly. This cross-filter coupling is correct but implicit; a one-line comment here pointing at `RegistryApplicationServletFilter#resetKeepingHeaders` would make the dependency obvious to anyone touching either filter later.

paoloantinori (commented):
Clean follow-up — consolidating the header-preservation logic into a single `resetKeepingHeaders(...)` helper near its only caller is a nice improvement over the per-header static methods, and the test matrix (200, mapper 404, unknown-path 404, disabled-API 404, 400) covers the short-circuit paths that motivated the helper.

One real discrepancy between the description and the implementation:

- **Dispatcher coverage claim vs. web.xml.** The PR description states the filter is mapped for `REQUEST`/`FORWARD`/`ERROR` dispatchers, but the `ContentTypeOptionsFilter` mapping in `web.xml` only declares `REQUEST` and `FORWARD` (same as the sibling `HSTSFilter`). As written, a response that reaches the filter chain via an `ERROR` dispatch (e.g. a container-served error page) would **not** get the `nosniff` header. That's probably fine for this REST API in practice, and it's consistent with how HSTS behaves — but the description's explicit `ERROR` claim is inaccurate either way. Suggest either correcting the description or, if error-dispatched responses are meant to be covered, adding `<dispatcher>ERROR</dispatcher>` to the mapping. Worth a quick decision before merge so the description doesn't mislead a future maintainer.

Minor (feel free to ignore):

- The static helper is named `addContentTypeOptionsHeaders` (plural) but adds a single header. Cosmetic, but `addContentTypeOptionsHeader` would read more precisely.

malladinagarjuna2:
hey @paoloantinori  really thanks for the fast review will implement the changes told by you

paoloantinori:
You're welcome, @malladinagarjuna2 — take your time with the changes. Just ping me (or re-request review) once you've pushed and I'll take another look.

app/src/main/java/io/apicurio/registry/ui/servlets/ContentTypeOptionsFilter.java
malladinagarjuna2: nice catch @paoloantinori 
added do filters 
noiw the dependency is explicit from either side.

malladinagarjuna2:
@paoloantinori could you please review

malladinagarjuna2:
corrected the pr description rather than adding error. adding it here would reintroduce the same unproven config you flagged on the HSTS PR.and it would leave nosniff inconsistent with HSTSFilter, which is  request only. If error-dispatched coverage is ever genuinely wanted, it should land for both filters together, with a test that actually triggers a container ERROR dispatch. The description was the inaccurate part, so that's what I fixed.

Naming: renamed addContentTypeOptionsHeaders → addContentTypeOptionsHeader (singular). Note HSTSFilter.addHstsHeaders has the same plural-for-one-header wording  happy to make that a seperate issue if you want it consistent.

Cross-filter coupling: added a one-line comment on doFilter pointing at RegistryApplicationServletFilter#resetKeepingHeaders, so the dependency is explicit for anyone touching either filter.

malladinagarjuna2:
If it's useful, I can pick up the broader filter-chain audit next,  the same short-circuit pattern means any future security header (CSP, X-Frame-Options) would need the same treatment. Happy to open an issue for it, if it is not creatwed yet

paoloantinori:
Agreed on the call to correct the description rather than add the `ERROR` dispatcher here — it keeps this PR consistent with `HSTSFilter` (which also maps only `REQUEST`/`FORWARD`) and avoids re-landing the config flag under the test profile I flagged on the HSTS PR. Sound judgment.

The broader filter-chain audit is a genuinely useful idea: the `response.reset()` short-circuit in `RegistryApplicationServletFilter` is the one place that silently drops upstream security headers, so any future header (CSP, `X-Frame-Options`) would need the same `resetKeepingHeaders(...)` treatment. I'd suggest opening an issue for it so it's tracked independently of this PR, but that's a maintainer call — happy to leave it to @carlesarnal/@EricWittmann to prioritize.

From my side the head still looks good; nothing more needed from me here.

paoloantinori:
Quick pointer: I just left my re-review thoughts in the comment above (the description-only fix was the right call; head looks good from my side). Nothing more needed from me here.

malladinagarjuna2:
thanks for the review @paoloantinori
@EricWittmann  could you please re view the pr

paoloantinori:
Happy to pass this along — and as I noted in my last re-review, the head looks good from my side (the description-only fix was the right call, keeping this consistent with `HSTSFilter`). Deferring to @EricWittmann for the final look.

paoloantinori:
Happy to pass this along, @malladinagarjuna2 — as noted in my last re-review, the head looks good from my side (the description-only fix was the right call, keeping this consistent with `HSTSFilter`). Deferring the final approval to @EricWittmann.

EricWittmann:
@malladinagarjuna2 Once we get CI to run (and is green) I will merge this. :+1:

malladinagarjuna2:
Hey @EricWittmann 
I noticed a recurring issue: contributors sometimes open multiple PRs against the same issue without linking it in the description, so you end up reviewing both and the second one gets closed/wasted once the first merges.

Proposed lightweight fix: have the PR-lifecycle bot post a nudge comment (non-blocking) when a PR is opened without a Fixes #NNN/Closes #NNN reference in the body, asking the author to link the issue. I've already got a working patch against .github/scripts/pr-lifecycle.js (hooks into the existing handlePrOpened welcome-comment flow, no new workflow needed).

Want me to open an issue for this and attach the PR, or would you rather discuss the approach first?

malladinagarjuna2:
a big example of this issue is https://github.com/Apicurio/apicurio-registry/pull/8869?

EricWittmann:
Solid suggestion.  We're still working on getting better at managing the increased contributor interest!  

@paoloantinori @carlesarnal @jsenko @vandanayadav7 -- wdyt?

malladinagarjuna2:
i already have a workflow for this after this i would love to create a pr on this if you allow me

EricWittmann:
> i already have a workflow for this after this i would love to create a pr on this if you allow me

I'm asking the team their opinions, but if it's not too much trouble go ahead with a PR and we can discuss more there.

malladinagarjuna2:
alright sir, i  think due to the current workflow  raising a pr when one is already open will be difficult, so i can raise it only when this one gets merged

paoloantinori:
Agreed with Eric — the observation is fair, and it's worth a light-touch fix. A few thoughts:

- The cheapest signal already exists when PRs use `Fixes #NNN` in the body: GitHub surfaces the linked PR on the issue timeline, so a glance at the issue shows how many open PRs target it. The gap is that not every contributor uses the keyword (this PR does, which is why it reads cleanly).
- A GitHub Action that comments on a newly-opened PR when it references an issue that already has another open, non-draft PR would surface the overlap early without blocking either side. The thing I'd avoid is auto-closing the second one — sometimes two approaches to the same issue are legitimately worth comparing.
- One guard that covers both the duplicate case *and* the unlinked case: have the action nudge when the issue link is missing from the body, rather than only flagging duplicates. That single check tends to solve most of the friction.

Happy to weigh in on a concrete proposal whenever you open it — no strong objection to you driving it.

malladinagarjuna2:
@EricWittmann  idk why the verification is failing everytime

malladinagarjuna2:
@EricWittmann  all tests have pass it's now good to merge

carlesarnal:
> #8869

I think it's a good idea.

PR #8608, opened 2026-07-14
fix(rest): return 400 instead of 500 when rendering an invalid prompt template

Me-Priyank:
## What

`POST /groups/{groupId}/artifacts/{artifactId}/versions/{versionExpression}/render` returns **500** when the stored PROMPT_TEMPLATE content is unparseable or has no `template` field. The request itself is well formed, the stored content simply isn't a usable prompt template, so this should be a 4xx.

Reported privately by email to @EricWittmann and @carlesarnal; opening this PR at Eric's suggestion.
## Why it matters

`PromptRenderingService.render()` wrapped every failure, including content-shaped ones, in a plain `RuntimeException`. That type isn't in `HttpStatusCodeMap`, so `getCode()` falls back to `HTTP_INTERNAL_ERROR`.

`CoreRegistryExceptionMapperService` reports **every** 500 to `ResponseErrorLivenessCheck`, whose `error-threshold` defaults to `1` per 60s window. So repeated render calls against one malformed template can drive `/health/live` down. The 500 response body also surfaced the underlying parser error.

Malformed content is storable because content is only validated on creation when a VALIDITY rule is enabled, which is off by default.

## Changes

`PromptRenderingService`:

- `extractTemplateText`: throw `BadRequestException` instead of `IllegalArgumentException` when the `template` field is missing.
- `parseContent`: throw `BadRequestException` when the content parses as neither YAML nor JSON. No cause is attached, so the response no longer exposes the raw parser message.
- Remove the blanket `catch (Exception) -> RuntimeException`. It produced the 500s, and it would otherwise re-wrap the mapped exception. Genuinely unexpected errors still surface as 500 and are still reported to the liveness check, which is the intent.

`BadRequestException` is already mapped to 400 in `HttpStatusCodeMap`, and is already what this endpoint throws for the wrong-artifact-type case, so no new exception type or mapping is introduced.

## Tests

- `PromptRenderTest`: two new cases asserting **400** (not 500), a template with no `template` field, and unparseable content.
- `PromptRenderingServiceTest`: the two existing error-case tests now assert `BadRequestException` instead of `RuntimeException`; added cases for empty content and non-object content.

Both new REST cases fail with `Expected status code <400> but was <500>` (emitting two `[500 ERROR DETECTED]` liveness reports) without the service change, and pass with it.

## Not changed

A non-textual `template` (e.g. `template: {a: 1}`) still renders as an empty string via `asText()`. It never throws, so it isn't part of this issue, happy to address it separately if you'd prefer.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -5,8 +5,7 @@ import io.apicurio.registry.content.ContentHandle;
 import io.apicurio.registry.rest.v3.beans.RenderPromptResponse;
 import io.apicurio.registry.rest.v3.beans.RenderValidationError;
 import jakarta.enterprise.context.ApplicationScoped;
-import jakarta.inject.Inject;
-import org.slf4j.Logger;
+import jakarta.ws.rs.BadRequestException;
 
 import java.util.ArrayList;
 import java.util.Iterator;
@@ -26,9 +25,6 @@ public class PromptRenderingService {
 
     private static final Pattern VARIABLE_PATTERN = Pattern.compile("\\{\\{([^}]+)\\}\\}");
 
-    @Inject
-    Logger log;
-
     /**
      * Renders a prompt template by substituting variables.
      *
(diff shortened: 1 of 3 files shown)


carlesarnal:
@Me-Priyank  please, only one pr at a time, we don't have the bandwidth to review everything, I'm closing this until https://github.com/Apicurio/apicurio-registry/pull/8587 is merged first.

Me-Priyank:
Sure @carlesarnal , thanks for the heads-up on bandwidth, happy to hold. Just for context, this is the render/liveness fix @EricWittmann asked me to open via email, so I'll reopen it right after #8587 lands. I'll keep to one PR at a time.

carlesarnal:
> Sure @carlesarnal , thanks for the heads-up on bandwidth, happy to hold. Just for context, this is the render/liveness fix @EricWittmann asked me to open via email, so I'll reopen it right after #8587 lands. I'll keep to one PR at a time.

Yup I know, I don't want to discourage, but we're a small time and we have to limit stuff a bit so we can review it carefully.

PR #8614, opened 2026-07-14
feat(auth): add Form Based Authentication Support #4916

### LINKED ISSUE #4916: Add Form Based Authentication Support for Apicurio Registry

Add support for Form Based authentication for Apicurio out of the box through parameters. The main parameters are designed to store credentials on other servers or platforms such as Keycloak. For many software tasks, it is required to be isolated and use only the internal credential database. It would be great to supplement basic auth with form-based authentication.

[Form Based Authentication](https://quarkus.io/guides/security-authentication-mechanisms#form-auth) is a type of authentication mechanism where users log in through a web form rather than a standard login dialog. This provides more flexibility and can be more user-friendly.

### Proposed Solution

Simplicity and variability in using the software. I am sure that many teams, when familiarizing themselves with the Apicurio Registry software, would find it convenient to use the internal configuration for the form, utilizing all the privileges of the settings, rather than using the standard basic auth.

### Additional Context

There may be simpler ways to get acquainted with the authorization in your product, and I would be happy to discuss them with you.

jainarjav80-sys:
## Purpose of this PR
This PR introduces Form Based Authentication support for Apicurio Registry, providing an out-of-the-box alternative or supplement to the existing basic authentication mechanism. It utilizes Quarkus's built-in FormAuthenticationMechanism and integrates it seamlessly into the application's authentication strategy chain.

This is particularly useful for teams that prefer using a standard web form login instead of a basic auth dialogue, isolating tasks without relying on external IdPs like Keycloak. Note: Enabling this feature requires the user to configure a backing Quarkus IdentityProvider (e.g., quarkus.security.users.embedded.* or a JDBC/LDAP realm) to validate the credentials.

## What was changed
AuthConfig.java: Added a new configuration property quarkus.http.auth.form.enabled to toggle form based authentication.
AppAuthenticationMechanism.java: Injected the FormAuthenticationMechanism lazily using Instance<> and wired it into the authentication chain via a DelegatingAuthenticationStrategy.
AppAuthenticationMechanismTest.java: Added a new unit test to cover the mechanism chain building, priority ordering, and disabled-mechanism paths.

## Related issues
Fixes #4916

## Verification
Users can enable this feature by setting quarkus.http.auth.form.enabled=true in their configuration.
It leverages Quarkus's native Form Authentication properties (e.g., quarkus.http.auth.form.login-page).


jainarjav80-sys:
@carlesarnal, please review whenever you have some time. Thanks!

paoloantinori:
@jainarjav80-sys Thanks for the contribution. Before we invest more review time, there are some architectural concerns:

**BLOCKER — No maintainer approval on the feature request**
Issue #4916 has no response from any maintainer (@carlesarnal @EricWittmann @jsenko). Adding a new authentication mechanism is a significant architectural decision that needs maintainer buy-in before implementation. Apicurio Registry is primarily an API service — form-based auth implies browser sessions, which raises questions about statelessness and horizontal scaling.

**BLOCKER — Uses Quarkus-managed property name**
```java
@ConfigProperty(name = "quarkus.http.auth.form.enabled", defaultValue = "false")
```
Per project convention (\`.claude/rules/config-properties.md\`), custom properties use the \`apicurio.\` prefix. Using a \`quarkus.\` property name directly ties this to Quarkus internals and bypasses the project's configuration documentation pipeline (\`@Info\` + config generator).

**MAJOR — Default priority change**
The default mechanism priority changes from \`"basic,proxy-header,oidc"\` to \`"basic,form,proxy-header,oidc"\`. Even though form auth is disabled by default, changing the default priority string is a behavioral change for existing deployments that rely on the default.

**MAJOR — No tests**
Authentication changes require tests covering: form login success, form login failure, session creation, session invalidation, CSRF protection, and interaction with existing OIDC/basic auth.

**Recommendation:** Wait for maintainer input on #4916 before continuing. If maintainers approve the feature direction, the implementation needs the issues above addressed.

carlesarnal (changes requested):
The diff shows malformed annotations - file paths instead of actual annotation syntax. This appears to be a display issue in the provided diff, not actual code. Assuming the real code has proper `@ConfigProperty` and `@Dynamic` annotations.

Inconsistent field access in `AppAuthenticationMechanism.java` line 127 vs 133. Use methods consistently: change `authConfig.proxyHeaderAuthEnabled` to `authConfig.isProxyHeaderAuthEnabled()` or add getter.

The `Instance<FormAuthenticationMechanism>` injection pattern differs from direct `BasicAuthenticationMechanism` injection. Document why form auth needs optional resolution or make basic auth consistent.

Default mechanism priority changed. Existing deployments will now attempt form auth by default. Document this behavioral change in release notes.

No tests for new functionality. Add unit tests for form auth enablement, mechanism chain building, and priority ordering.

File paths: `app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java`, `app/src/main/java/io/apicurio/registry/auth/AuthConfig.java`

jainarjav80-sys:
@carlesarnal, Thanks for the helpful review! I've pushed a commit that fixes the field access and injection inconsistencies, reverts the default priority change to preserve backward compatibility, and adds a unit test for the mechanism chain. Let me know if it looks good now!

jsenko:

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
```diff
@@ -68,6 +69,9 @@ public class AppAuthenticationMechanism implements HttpAuthenticationMechanism {
     @Inject
     BasicAuthenticationMechanism basicAuthenticationMechanism;
 
+    @Inject
```
paoloantinori, line 72: Hard `@Inject` of `FormAuthenticationMechanism` may fail deployment with `UnsatisfiedDependencyException` when `quarkus.http.auth.form.enabled=false` (the default), since Quarkus only registers that bean when form auth is on. The `kubernetesClient` field below uses `Instance<>` + `isResolvable()` to avoid exactly this; worth applying the same pattern here, or confirming the bean is always present. The unit test mocks the field directly and so would not surface a deployment-time failure.

paoloantinori (commented):
Re-reviewing the current head (`a030af5`) on its merits, independent of the open architectural question on #4916 (which I raised earlier and still stands — the issue has no maintainer approval and the registry has historically routed auth through external IdPs / basic, not form login).

The wiring is clean and the `buildAuthChain` tests are solid, but there is one concrete robustness issue worth resolving before merge:

- **`FormAuthenticationMechanism` is injected unconditionally with `@Inject`, but Quarkus only instantiates that bean when `quarkus.http.auth.form.enabled=true`.** When form auth is disabled (the documented default), the `FormAuthenticationMechanism` bean may not be registered, and a hard `@Inject` field will fail deployment with an `UnsatisfiedDependencyException` rather than degrading gracefully. Compare with the OIDC mechanism, which is typically guarded by `@Inject @IfBuildProfile` or an `Instance<>` + `isResolvable()` check (as the Kubernetes client is handled here via `Instance<KubernetesClient>`). Worth confirming the app still boots with form auth disabled, and if not, switching to `Instance<FormAuthenticationMechanism>` with an `isResolvable()` guard in the factory — exactly the pattern already used for `kubernetesClient` in this same class. The current `AppAuthenticationMechanismTest` mocks the field directly, so it would not catch a deployment-time CDI failure.

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
jainarjav80-sys: Fixed! Switched back to Instance<> injection with an isResolvable() guard to prevent startup crashes when disabled.

jainarjav80-sys:
@EricWittmann @paoloantinori  @carlesarnal  ,,I have addressed all the feedback and pushed the fixes (reverted to Instance<> injection, fixed the field access inconsistencies, reverted the default priority change, and added a unit test).

I also rebased the branch onto the latest main so the FOSSA checks are passing now. Let me know if there's anything else needed!

paoloantinori:
Thanks for iterating on this and addressing the earlier feedback (Instance<> injection, reverted default priority change, unit test). Re-basing on latest main for the FOSSA fix is the right call.

One thing I want to flag before this merges — I can't tell from the diff alone whether the removal of `isBasicAuthEnabled()` from `AuthConfig.java` (the diff shows `-    public boolean isBasicAuthEnabled() {` / `-        return this.basicAuthEnabled;` / `-    }`) is intentional or a rebase artifact. That method has active callers in `main`:
- `SystemResourceImpl.java:98` — `authConfig.isBasicAuthEnabled() ? UserInterfaceConfigAuth.Type.basic`
- `WellKnownResourceImpl.java:624` — `authConfig.isOidcAuthEnabled() || authConfig.isBasicAuthEnabled()`
- `RegistryAgentCardBuilder.java:124` — `if (authConfig.isBasicAuthEnabled())`

If those callers are still present on this head's base, removing the getter would fail compilation. Could you confirm the current CI `Build` job is green on this head (`727f4fa`)? If it's red, this is likely the cause. If those callers were also removed in this PR (and the truncated diff is just hiding them), then ignore me — but I don't see them in the changed-files list, which only shows `AppAuthenticationMechanism.java`, `AuthConfig.java`, and the new test.

Separately: I still think the architectural question on #4916 (no maintainer approval on the issue) should be resolved before this lands, but that's @carlesarnal/@EricWittmann's call.

jainarjav80-sys:
@paoloantinori ,,, Thanks for catching that. The removal of isBasicAuthEnabled() was an unintended side effect while I was making the field access consistent. I've restored the getter and updated the usage accordingly. The latest CI checks are green on the current head.

I also appreciate your continued review and the architectural concerns you've raised around #4916. I understand the need for maintainer sign-off on a change in the authentication area, and I'll defer to @carlesarnal and @EricWittmann on that decision.

Please let me know if you spot anything else that should be addressed.

paoloantinori:
Thanks @jainarjav80-sys — the restored `isBasicAuthEnabled()` getter and the consistent field access look correct on this head, and green CI is good to see.

The code-level concerns from the earlier rounds are resolved from my side. The remaining open item is the architectural one I flagged up front: issue #4916 still has no maintainer sign-off on the feature direction, and adding a new authentication mechanism is the kind of change that needs that green light before it lands. I'll defer that decision to @carlesarnal and @EricWittmann — once they confirm the feature is wanted for this line, the implementation here is in good shape to move forward.

app/src/main/java/io/apicurio/registry/auth/AuthConfig.java
```diff
@@ -46,6 +46,10 @@ public class AuthConfig {
     @Info(category = CATEGORY_AUTH, description = "Enable basic auth", availableSince = "3.X.X.Final")
     boolean basicAuthEnabled;
 
+    @ConfigProperty(name = "quarkus.http.auth.form.enabled", defaultValue = "false")
```
paoloantinori, line 49: Heads up that toggling `quarkus.http.auth.form.enabled=true` alone won't produce a functional login flow — `FormAuthenticationMechanism` still needs a backing `IdentityProvider`/security realm (e.g. `quarkus.security.users.embedded.*` or a JDBC realm). The PR description implies an "internal credential database" but none is wired up here. Consider noting the required companion config in the `@Info` description or in the PR docs so users don't hit a login form that rejects every credential.

app/src/main/java/io/apicurio/registry/auth/AuthConfig.java
```diff
@@ -46,6 +46,10 @@ public class AuthConfig {
     @Info(category = CATEGORY_AUTH, description = "Enable basic auth", availableSince = "3.X.X.Final")
     boolean basicAuthEnabled;
```
paoloantinori, line 48: Minor consistency nit: neighboring `basicAuthEnabled` uses `availableSince = "3.X.X.Final"` (the repo's placeholder convention), while this uses a concrete `"3.3.0"`. Worth aligning unless the concrete version is intentional.

paoloantinori (commented):
Re-reviewing head `e8a9c064` after the architectural blocker on #4916 was resolved via maintainer `/accept`. The wiring here is sound: `Instance<FormAuthenticationMechanism>` mirrors the lazy-injection pattern already used for `kubernetesClient`, the chain entry is correctly gated on both `isFormAuthEnabled()` and `isResolvable()`, and reusing the Quarkus built-in property name (`quarkus.http.auth.form.enabled`) as the gate cleanly keeps the boolean flag and the bean's conditional registration in sync. The unit tests cover the priority-ordering and disabled-mechanism paths well.

A couple of observations worth considering before merge:

- **Enabling the flag alone does not yield a working form login.** `FormAuthenticationMechanism` validates credentials against a Quarkus `IdentityProvider` / security realm. The PR description talks about an "internal credential database", but no identity store (e.g. `quarkus.security.users.embedded.*`, or a JDBC/LDAP realm) is configured here, so a user who flips `quarkus.http.auth.form.enabled=true` following the docs will get a login form that cannot authenticate anyone. Worth either documenting the required companion config explicitly, or wiring a minimal realm. See [Quarkus form auth docs](https://quarkus.io/guides/security-built-in-authentication#form-authentication).
- Minor: `formAuthEnabled` uses `availableSince = "3.3.0"` while the neighboring `basicAuthEnabled` uses the repo's placeholder convention `"3.X.X.Final"`. Trivial, but worth aligning for consistency.

jainarjav80-sys:
@paoloantinori  All feedback has been addressed and the threads are resolved! I updated the `@Info` annotation to document the Identity Provider requirement and aligned the `availableSince` version placeholder. The CI checks and Quality Gate are all green. Let me know if there's anything else needed for merge!

paoloantinori:
Thanks @jainarjav80-sys — confirmed all the earlier points are addressed on this head: the `@Info` description now documents the Quarkus Identity Provider requirement, `availableSince` matches the existing `basicAuthEnabled` convention in the file, the `Instance<FormAuthenticationMechanism>` wiring with `isResolvable()` guard is clean, and the unit test covers chain ordering and disabled-mechanism paths. Nothing further from my side — ready for a maintainer to merge.

app/src/main/java/io/apicurio/registry/auth/AuthConfig.java
```diff
@@ -46,6 +46,10 @@ public class AuthConfig {
...
 
+    @ConfigProperty(name = "quarkus.http.auth.form.enabled", defaultValue = "false")
+    @Info(category = CATEGORY_AUTH, description = "Enable form auth. Requires configuring a Quarkus identity provider (e.g. quarkus.security.users.embedded.* or a JDBC/LDAP realm).", availableSince = "3.X.X.Final")
```
carlesarnal: The available since field must be set to the value in the pom, currently 3.3.1

carlesarnal (changes requested):
The implementation looks correct, but a feature like this requires integration tests, not just unit. Once those are added I think we can proceed.

jainarjav80-sys:
Hi @carlesarnal ,I have updated the availableSince version and added a full suite of integration tests (FormAuthTest.java) to verify the form login flow and role-based access using embedded test users. Let me know if there's anything else you need before merging!

app/src/test/java/io/apicurio/registry/auth/FormAuthTest.java
carlesarnal, line 1: This isn't a proper integration test: https://github.com/Apicurio/apicurio-registry/blob/main/integration-tests/README.md

carlesarnal (dismissed):

paoloantinori:
Agreeing with @carlesarnal here, and worth being precise about the distinction so we're aligned on what's being asked.

The project's `integration-tests/` module is the canonical home for these, and per its README it uses `@QuarkusIntegrationTest` — which boots the **built** artifact (production JAR or Docker image) as a separate process, not the test-managed in-JVM Quarkus. That black-box posture matters for a security feature: it catches wiring that only resolves in the packaged runtime (extension loading, the real `FormAuthenticationMechanism` bean from `quarkus-vertx-http`, production config overlays).

The new `FormAuthTest` is a `@QuarkusTest` in `app/src/test/.../auth/`. It's not without value — it drives the actual Vert.x HTTP form-login endpoint (`/j_security_check`), validates the `quarkus-credential` session cookie, and enforces RBAC through the real REST layer against an embedded identity provider. So it's a genuine component-integration test rather than a unit test. But it runs against the test-classpath Quarkus, not the packaged artifact, so it can't catch the class of issues a `@QuarkusIntegrationTest` would (that's exactly the gap @carlesarnal is pointing at).

The `auth` test group in `integration-tests/` is the natural home — it already carries the Keycloak-based auth tests. Mirroring the form-login flow there (form auth + a backing identity provider, embedded or Keycloak) would close the black-box gap. The current `app`-level test is still worth keeping as a faster inner-loop check.

jainarjav80-sys:
Hi @paoloantinori @carlesarnal , I dug into the integration test quite a bit and I'm still blocked.

I confirmed that the feature itself works under @QuarkusTest, but I'm seeing different behavior under @QuarkusIntegrationTest when the packaged JAR is launched.

I've investigated the interaction between RegistryDeploymentManager, Quarkus security configuration, and role propagation, but I'm not yet certain whether I'm following the intended pattern for configuring Form Auth in integration tests.

Am I approaching this the right way, or is there an existing pattern in the project that I should mirror for Form Authentication?

I'd appreciate any guidance before I keep going down.

paoloantinori:
Thanks for digging into this @jainarjav80-sys — the discrepancy you're seeing between `@QuarkusTest` and `@QuarkusIntegrationTest` is almost certainly expected behavior rather than a bug in your wiring, and it comes down to how each test type interacts with build-time configuration.

The key point: `quarkus.http.auth.form.enabled` is a **build-time** property in Quarkus (the `quarkus.http.auth.form.*` family is processed during augmentation, not at runtime). The Quarkus testing guide states this directly: test-specific configuration (like a `@TestProfile` or `src/test/resources/application.properties`) can change runtime properties freely, but build-time properties only take effect if they are present *when the application is built*.

That maps exactly onto the two test modes:
- `@QuarkusTest` rebuilds the application in-process with the `@TestProfile` applied, so `quarkus.http.auth.form.enabled=true` from `FormAuthTestProfile` is honored at augmentation time → the `/j_security_check` route and `FormAuthenticationMechanism` are wired in → your flow works.
- `@QuarkusIntegrationTest` launches the *already-built* artifact (the packaged JAR), so a `@TestProfile` flipping a build-time property at test time has no effect — the form auth route was never registered at build time, hence the different behavior you observe under the packaged JAR.

A couple of practical implications worth flagging:
1. For the integration test to exercise form auth, `quarkus.http.auth.form.enabled` would need to be enabled at *build* time (e.g., in a dedicated `application.properties` used for that build), not only via a test profile. The Quarkus guide also notes `@QuarkusTest` and `@QuarkusIntegrationTest` should not be run in the same execution, so they're genuinely different beasts.
2. I'd double-check whether the `app` module's main build can safely have form auth enabled-by-default at build time gated only by the runtime flag `apicurio.authn.form.enabled`. If yes, the `@QuarkusIntegrationTest` path becomes viable; if not, this is a known limitation of testing build-time auth features with `@QuarkusIntegrationTest` and the `@QuarkusTest`-based `FormAuthTest` may be the pragmatic home for this coverage.

Given the maintainer feedback (@carlesarnal) that the `integration-tests/` module is the canonical home, option (1) is probably the right direction, but it's worth confirming with @carlesarnal / @EricWittmann whether enabling form auth at build time is acceptable for the main app image. I'm fairly confident in the build-time-vs-runtime diagnosis above, but I haven't traced the exact Quarkus recorder step that registers the form route, so treat the mechanism as high-confidence and the specific remediation as a suggestion to validate.

carlesarnal:
> Thanks for digging into this @jainarjav80-sys — the discrepancy you're seeing between `@QuarkusTest` and `@QuarkusIntegrationTest` is almost certainly expected behavior rather than a bug in your wiring, and it comes down to how each test type interacts with build-time configuration.
> 
> The key point: `quarkus.http.auth.form.enabled` is a **build-time** property in Quarkus (the `quarkus.http.auth.form.*` family is processed during augmentation, not at runtime). The Quarkus testing guide states this directly: test-specific configuration (like a `@TestProfile` or `src/test/resources/application.properties`) can change runtime properties freely, but build-time properties only take effect if they are present _when the application is built_.
> 
> That maps exactly onto the two test modes:
> 
> * `@QuarkusTest` rebuilds the application in-process with the `@TestProfile` applied, so `quarkus.http.auth.form.enabled=true` from `FormAuthTestProfile` is honored at augmentation time → the `/j_security_check` route and `FormAuthenticationMechanism` are wired in → your flow works.
> * `@QuarkusIntegrationTest` launches the _already-built_ artifact (the packaged JAR), so a `@TestProfile` flipping a build-time property at test time has no effect — the form auth route was never registered at build time, hence the different behavior you observe under the packaged JAR.
> 
> A couple of practical implications worth flagging:
> 
> 1. For the integration test to exercise form auth, `quarkus.http.auth.form.enabled` would need to be enabled at _build_ time (e.g., in a dedicated `application.properties` used for that build), not only via a test profile. The Quarkus guide also notes `@QuarkusTest` and `@QuarkusIntegrationTest` should not be run in the same execution, so they're genuinely different beasts.
> 2. I'd double-check whether the `app` module's main build can safely have form auth enabled-by-default at build time gated only by the runtime flag `apicurio.authn.form.enabled`. If yes, the `@QuarkusIntegrationTest` path becomes viable; if not, this is a known limitation of testing build-time auth features with `@QuarkusIntegrationTest` and the `@QuarkusTest`-based `FormAuthTest` may be the pragmatic home for this coverage.
> 
> Given the maintainer feedback (@carlesarnal) that the `integration-tests/` module is the canonical home, option (1) is probably the right direction, but it's worth confirming with @carlesarnal / @EricWittmann whether enabling form auth at build time is acceptable for the main app image. I'm fairly confident in the build-time-vs-runtime diagnosis above, but I haven't traced the exact Quarkus recorder step that registers the form route, so treat the mechanism as high-confidence and the specific remediation as a suggestion to validate.

No, enabling form auth for the entire app at build time isn't ok. This is something that should be on demand and, if not possible, we should probably consider removing it.

jainarjav80-sys:
I completely agree, @carlesarnal . We definitely shouldn't bake Form Auth into the global build-time configuration just for testing.

My understanding is that the feature should remain fully on-demand in production, enabled only when apicurio.authn.form.enabled is explicitly configured.

Given that, would it be acceptable to move the test coverage back to a @QuarkusTest in the app module using a @TestProfile instead of @QuarkusIntegrationTest? This would keep the production application untouched while still providing test coverage.

Does that align with the approach you'd prefer?

jsenko (commented):

The core implementation looks solid — Instance<> injection with isResolvable() guard (addressing paoloantinori's earlier feedback), correct chain integration, and proper config property setup. However, the test configuration has issues that mean the new code path is not actually exercised by the tests.

### 1. Dead config property in test profile (`FormAuthTestProfile.java:20`)

```java
props.put("apicurio.authn.form.enabled", "true");  // line 20
```

No code in the codebase reads `apicurio.authn.form.enabled`. The actual property is `quarkus.http.auth.form.enabled` (set on line 21). This line does nothing and will confuse future readers. Remove it.

### 2. "form" missing from mechanism priority in test profile — chain code path untested

`FormAuthTestProfile` doesn't set `apicurio.authn.mechanism.priority`, so the default `"basic,proxy-header,oidc"` is used. Since "form" is not in that list, the `DelegatingAuthenticationStrategy("form", formAuthenticationMechanism.get())` code in `buildAuthChain()` is never executed during tests.

The tests pass because Quarkus's built-in form auth handles `/j_security_check` and cookie validation at the Vert.x layer, independently of the custom `AppAuthenticationMechanism` chain. This means the integration you're adding to `buildAuthChain()` is untested.

Fix: add to the test profile:
```java
props.put("apicurio.authn.mechanism.priority", "form");
```

### 3. Documentation gap — users need to configure mechanism priority

The `@Info` description for `quarkus.http.auth.form.enabled` says "Enable form auth" but doesn't mention that users must **also** add "form" to `apicurio.authn.mechanism.priority` for it to be included in the auth chain. The default priority is `"basic,proxy-header,oidc"` — "form" is not there.

Consider updating the description to:
> "Enable form auth. Requires adding 'form' to apicurio.authn.mechanism.priority and configuring a Quarkus identity provider."

carlesarnal:
> I completely agree, @carlesarnal . We definitely shouldn't bake Form Auth into the global build-time configuration just for testing.
> 
> My understanding is that the feature should remain fully on-demand in production, enabled only when apicurio.authn.form.enabled is explicitly configured.
> 
> Given that, would it be acceptable to move the test coverage back to a @QuarkusTest in the app module using a @testprofile instead of @QuarkusIntegrationTest? This would keep the production application untouched while still providing test coverage.
> 
> Does that align with the approach you'd prefer?

Yes, please, move it to the unit tests, and sorry for the confusion.

jainarjav80-sys:
Thanks @carlesarnal for confirming the approach, and @jsenko @paoloantinori for the feedback!

I've just pushed a new commit that moves the test coverage back to the unit tests in the app module as requested, and addresses all the recent review comments:

1.Moved Tests: The test coverage is now entirely handled by a @QuarkusTest in the app module. Form Authentication is fully on-demand and is no longer baked into the global build-time configuration.

2.Fixed the Test Profile: Removed the dead apicurio.authn.form.enabled property and added props.put("apicurio.authn.mechanism.priority", "form"); to FormAuthTestProfile.java, ensuring the custom AppAuthenticationMechanism chain is actually exercised by the tests.

3.Addressed the Documentation Gap: Updated the @Info annotation in AuthConfig.java to explicitly mention that users must configure the mechanism priority to include form and configure a Quarkus identity provider for it to work.

4.Resolved NPE & Security Bypass: Refactored AppAuthenticationMechanism to use Instance<FormAuthenticationMechanism> with .isResolvable() to cleanly handle the optional dependency (fixing the NPE). Also updated AuthorizedInterceptor to ensure RBAC is correctly bypassed or enforced based on authConfig.isFormAuthEnabled().

The build is entirely green locally. Let me know if everything looks good to merge!

paoloantinori:
Thanks @jainarjav80-sys — confirming the new head addresses the open point cleanly. Moving the coverage back to a `@QuarkusTest` in the app module (with `FormAuthTestProfile` enabling form auth + the embedded identity provider on demand) is the right resolution: it keeps form authentication out of the global build-time configuration while still exercising the real login flow, `j_security_check`, the `quarkus-credential` cookie, and role-based access (`sr-developer` vs `sr-readonly`).

The wiring in `AppAuthenticationMechanism` reads correctly too — `Instance<FormAuthenticationMechanism>` with the `isResolvable()` guard mirrors the basic-auth pattern, and the priority-list integration is consistent. From my side the code-level concerns are resolved.

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
```diff
@@ -120,10 +124,16 @@ List<AuthenticationStrategy> buildAuthChain() {
...
 
         Map<String, Supplier<AuthenticationStrategy>> strategyFactories = new LinkedHashMap<>();
-        strategyFactories.put("basic", () -> authConfig.basicAuthEnabled
-                ? new DelegatingAuthenticationStrategy("basic", basicAuthenticationMechanism)
+        strategyFactories.put("basic", () -> authConfig.isBasicAuthEnabled() && basicAuthenticationMechanism.isResolvable()
+                ? new DelegatingAuthenticationStrategy("basic", basicAuthenticationMechanism.get())
                 : null);
-        strategyFactories.put("proxy-header", () -> authConfig.proxyHeaderAuthEnabled
+        strategyFactories.put("form", () -> {
+            log.info("Checking form auth. isFormAuthEnabled={}, isResolvable={}", authConfig.isFormAuthEnabled(), formAuthenticationMechanism.isResolvable());
```
paoloantinori: This `log.info("Checking form auth...")` reads like a leftover diagnostic. At INFO level inside `buildAuthChain()` it will clutter production logs — and spam every request if the chain is rebuilt per authentication attempt. Please remove it, or demote to `log.debug` if the resolvability signal is genuinely useful.

paoloantinori (commented):
Solid, well-scoped feature: form auth is wired into the existing strategy chain via a lazy `Instance<FormAuthenticationMechanism>` with both unit and integration coverage. The main blockers are two leftover INFO-level debug log statements in `buildAuthChain()` that should be removed (or demoted) before merge, plus a couple of observability/test-robustness nits.

- **Leftover debug logging** — `log.info("Checking form auth. isFormAuthEnabled={}, isResolvable={}")` at `AppAuthenticationMechanism.java:131` and `log.info("Mechanism priority list: {}")` at `:161` read like temporary diagnostics; if the chain is rebuilt per authentication attempt this is significant log spam at INFO, and the second line is redundant with `AuthConfig.onConstruct`.
- **Silent misconfiguration** — at `AppAuthenticationMechanism.java:132`, when `quarkus.http.auth.form.enabled=true` but the `FormAuthenticationMechanism` bean isn't resolvable, form auth is dropped from the chain with no log signal, so users hit unexplained 401/403s. A `log.warn` on the enabled-but-not-resolvable branch would make this visible for a newly added mechanism.
- **Over-permissive test assertion** — `FormAuthTest.java:64` asserts `status == 401 || 302 || 403`, which passes under almost any configuration (including a broken or disabled one); asserting the single expected status (likely 302) would let the test actually catch regressions.

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
```diff
@@ -148,11 +158,12 @@ List<AuthenticationStrategy> buildAuthChain() {
         });
 
         List<AuthenticationStrategy> chain = new ArrayList<>();
+        log.info("Mechanism priority list: {}", authConfig.getMechanismPriorityList());
```
paoloantinori: Same concern: `log.info("Mechanism priority list: {}")` looks like debug output. The priority list is already logged in `AuthConfig.onConstruct`, so this is redundant at INFO — remove it or demote to debug.

app/src/main/java/io/apicurio/registry/auth/AppAuthenticationMechanism.java
```diff
@@ -120,10 +124,16 @@ List<AuthenticationStrategy> buildAuthChain() {
...
         Map<String, Supplier<AuthenticationStrategy>> strategyFactories = new LinkedHashMap<>();
-        strategyFactories.put("basic", () -> authConfig.basicAuthEnabled
-                ? new DelegatingAuthenticationStrategy("basic", basicAuthenticationMechanism)
+        strategyFactories.put("basic", () -> authConfig.isBasicAuthEnabled() && basicAuthenticationMechanism.isResolvable()
+                ? new DelegatingAuthenticationStrategy("basic", basicAuthenticationMechanism.get())
                 : null);
-        strategyFactories.put("proxy-header", () -> authConfig.proxyHeaderAuthEnabled
+        strategyFactories.put("form", () -> {
+            log.info("Checking form auth. isFormAuthEnabled={}, isResolvable={}", authConfig.isFormAuthEnabled(), formAuthenticationMechanism.isResolvable());
+            return authConfig.isFormAuthEnabled() && formAuthenticationMechanism.isResolvable()
```
paoloantinori: If `quarkus.http.auth.form.enabled=true` is set but `formAuthenticationMechanism` isn't resolvable (e.g. no backing identity provider configured), form auth is silently dropped from the chain and users get unexplained 401/403s with no log trail. Consider a `log.warn` on this enabled-but-not-resolvable branch so the misconfiguration is visible — especially valuable since this mechanism is newly added.

app/src/test/java/io/apicurio/registry/auth/FormAuthTest.java
```diff
@@ -0,0 +1,140 @@
...
+    public void testUnauthenticatedAccess() {
+        // Request without cookie should fail (either 401 or 302 redirect to login)
+        Response response = given()
+            .when()
+                .get("/registry/v3/search/artifacts")
+            .then()
+                .extract().response();
+        
+        int status = response.statusCode();
+        assertTrue(status == 401 || status == 302 || status == 403, "Status should be 401, 302, or 403 but was " + status);
```
paoloantinori: Asserting `status == 401 || 302 || 403` makes this test pass under almost any security configuration, including a broken or fully disabled one, so it can't actually detect a form-auth regression. Consider asserting the single status you expect for an unauthenticated request under form auth (typically a 302 redirect to the login page).

EricWittmann:
Hi @jainarjav80-sys — apologies for the delay getting back to you on this. The feature is looking good and we'd like to move it forward.

Two things needed before we can merge:

1. **Merge conflicts** — the branch has conflicts with main that need resolving (rebase or merge).
2. **Leftover log statements** — as @paoloantinori flagged in the last review, there are two `log.info()` calls in `buildAuthChain()` that look like debug output (`"Checking form auth..."` and `"Mechanism priority list: {}"`). Please remove them or demote to `log.debug`.

Once those are addressed, we should be good to go. Thanks for your patience!

jainarjav80-sys:
Hi @EricWittmann ,

Thanks for reviewing! I've gone ahead and addressed both points:

1.Merged main and resolved all the conflicts.

2.Demoted the Mechanism priority list log to debug, and completely removed the Checking form auth... log (I replaced it with a cleaner .warn() that only triggers if the mechanism fails to resolve when enabled). I also took the opportunity to tighten the assertions in FormAuthTest just to be safe.

Let me know if there's anything else needed, otherwise this should be good to go once the CI finishes!

paoloantinori:
Thanks @jainarjav80-sys — confirming on this head that both points are addressed:

- **Merge conflicts** — resolved cleanly against current main.
- **Log statements** — the `Mechanism priority list` log is demoted to `debug`, and the `Checking form auth...` log is removed (replaced by the targeted `warn()` that only fires when form auth is enabled but no `FormAuthenticationMechanism` is resolvable), which is the right level of noise for steady-state operation.

Nothing further from my side on the implementation. Deferring to @EricWittmann / @carlesarnal on the merge call.

app/src/main/java/io/apicurio/registry/auth/AuthConfig.java
```diff
@@ -243,9 +252,8 @@ public boolean isProxyHeaderAuthEnabled() {
      * call sites on this method so a new backend cannot silently bypass authz.
      */
     public boolean isAuthenticationEnabled() {
-        return oidcAuthEnabled || basicAuthEnabled || proxyHeaderAuthEnabled || kubernetesAuthEnabled;
+        return oidcAuthEnabled || basicAuthEnabled || proxyHeaderAuthEnabled || kubernetesAuthEnabled || formAuthEnabled;
```
paoloantinori, line 255: When `quarkus.http.auth.form.enabled=true` but `form` is not added to `apicurio.authn.mechanism.priority` (the default omits it), `isAuthenticationEnabled()` returns true while the chain contains no mechanism able to authenticate form requests — a silent lockout with no warning (the existing `log.warn` in `buildAuthChain` only fires for unknown names, not enabled-but-unlisted mechanisms). Consider validating in `@PostConstruct` or `buildAuthChain` that every enabled mechanism also appears in the priority list.

paoloantinori (commented):
Overall the production wiring is sound — the lazy `Instance<>` injection plus `isResolvable()` guards mirror the existing Kubernetes mechanism pattern, and `isAuthenticationEnabled()` is correctly extended to include form auth. The main concerns are a silent misconfiguration foot-gun in `AuthConfig` and integration-test assertions that read as unverified against actual runtime behavior.

- **Silent lockout risk** (`AuthConfig.java:255`): `quarkus.http.auth.form.enabled=true` flips `isAuthenticationEnabled()` on, but `form` only enters the chain if the operator also adds it to `apicurio.authn.mechanism.priority` (the default `basic,proxy-header,oidc` omits it). An operator who enables only form auth without editing the priority list gets authentication 'on' with no mechanism able to satisfy requests — a silent lockout with no warning, since the existing `log.warn` only fires for *unknown* names, not enabled-but-unlisted ones.
- **Unverified auth-flow assertions** (`FormAuthTest.java:65`): `testUnauthenticatedAccess` hedges 'either 401 or 302' in its comment but asserts a hard 401; Quarkus form auth normally issues a 302 redirect to the login page. Confirm what the custom mechanism actually returns. Separately, `login()` asserts `statusCode(302)` on a POST, but REST Assured follows redirects by default, so the captured status is likely the post-redirect response — disable redirect-following or assert the final state.
- **Role source for embedded/form users** (`FormAuthTestProfile.java:27`): `role-source=token` reads roles from JWT claims, but embedded identity-provider users carry roles only on the `SecurityIdentity`. If `token` does not fall back to identity groups, `testRoleBasedAuthorization` and `testReadOnlyRole` won't enforce `sr-developer`/`sr-readonly` as intended — consider `role-source=application` and verify the role names map.
- **CSRF**: form auth plus session cookies on a REST API reintroduces CSRF exposure for state-changing requests; worth documenting that production deployments need CSRF protection and a Secure/HttpOnly cookie.

app/src/test/java/io/apicurio/registry/auth/FormAuthTest.java
```diff
@@ -0,0 +1,141 @@
...
+    public void testUnauthenticatedAccess() {
+        // Request without cookie should fail (either 401 or 302 redirect to login)
+        Response response = given()
+            .when()
+                .get("/registry/v3/search/artifacts")
+            .then()
+                .extract().response();
+        
+        int status = response.statusCode();
+        assertEquals(401, status, "Unauthenticated access should return 401 Unauthorized");
```
paoloantinori, line 67: The comment hedges 'either 401 or 302' but this asserts a hard 401. Quarkus form auth normally answers an unauthenticated request with a 302 redirect to the login page, so confirm what the custom `AppAuthenticationMechanism` actually sends and assert it explicitly. Also note: `login()` asserts `statusCode(302)` on a POST, but REST Assured follows redirects by default, so it likely captures the post-redirect status — set `redirects().follow(false)` or assert the final state.

utils/tests/src/main/java/io/apicurio/registry/utils/tests/FormAuthTestProfile.java
```diff
@@ -0,0 +1,51 @@
...
+    public Map<String, String> getConfigOverrides() {
+        Map<String, String> props = new HashMap<>();
+
+        // Enable form authentication
+        props.put("quarkus.http.auth.form.enabled", "true");
+        props.put("apicurio.authn.mechanism.priority", "form");
+
+        // Enable role-based authorization to test group/role mapping
+        props.put("apicurio.auth.role-based-authorization", "true");
+        props.put("apicurio.auth.role-source", "token");
```
paoloantinori, line 28: `role-source=token` resolves roles from JWT/OIDC token claims, but the embedded identity provider used here surfaces roles only on the `SecurityIdentity` (no token). If `token` does not fall back to identity groups, `testRoleBasedAuthorization` and `testReadOnlyRole` will not enforce `sr-developer`/`sr-readonly` as intended. Consider `role-source=application` and verify the role names actually map before relying on these tests.

jainarjav80-sys:
Hey @paoloantinori , I've pushed the fixes for the AppAuthenticationMechanism priority validation and the AuthConfig CSRF warning!

Regarding your suggestion for FormAuthTest and FormAuthTestProfile — I actually had to revert the test changes back to how I originally had them (role-source=token and asserting 401).

It turns out that TokenRoleProvider does automatically fall back to identity groups because it delegates to SecurityIdentity::hasRole under the hood! When I changed it to application, the tests failed because the embedded identity users obviously didn't have their roles configured in the DB. Also, the unauthenticated REST API endpoint natively returns a hard 401 Unauthorized instead of a 302 Found, so my initial assertion was correct there too.

I've confirmed all tests are fully green now. Let me know if everything looks good to merge!

jainarjav80-sys:
Hey @EricWittmann and @carlesarnal  , just a quick heads-up that I've addressed all the latest feedback!

We've added the missing mechanism priority validation, updated the configuration documentation with the CSRF warnings, and got all the tests fully green and robust.

Everything is pushed and CI is passing, so this should be completely ready for a final look and merge whenever you have a moment. Thanks!

paoloantinori:
Good call on reverting `FormAuthTestProfile` back to `role-source=token` with the 401 assertion — your reasoning holds up on both points.

On the role-source swap: you're right that `TokenRoleProvider` aggregates through `SecurityIdentity.hasRole()`, which consults the union of role sources including `RolePermission`, so the embedded test users get their roles resolved through the identity's group/role mapping rather than a direct DB lookup. Switching to `application` would have stripped that aggregation path and left the embedded users without any granted role — the test breakage you saw is exactly the expected symptom. Keeping `token` is correct for this test's intent (verifying the form-login → SecurityIdentity → role check flow).

On the 401 vs 302: also correct. Quarkus's `FormAuthenticationMechanism` only emits the 302 redirect to the login page when the request advertises `Accept: text/html` (browser-style). A plain REST client gets a hard 401, which is the right thing to assert for the unauthenticated API path. The original assertion was right; my suggestion would have been testing a different code path than intended.

With the priority-list validation, the CSRF warning in `AuthConfig`, and the test back to its correct shape, this looks good from my side. Thanks for the clear write-up on the revert reasoning.

jainarjav80-sys:
Hi @paoloantinori @EricWittmann , just a quick update that all the CI checks have successfully passed! The previous failure was just because I missed committing the auto-generated .adoc file for the config warning, which is now fixed.

It looks like the PR is still blocked by the previous 'Changes requested' review status, and there is one final workflow awaiting a maintainer's approval to run. Let me know if everything looks good to officially approve and merge!

paoloantinori:
Thanks for the update. On the code side, the current head looks consistent with the feedback from the last rounds (mechanism priority validation, debug-level logging, FormAuthTestProfile role-source, the auto-generated .adoc).

Two things are still standing between this and a merge, and both are outside what I can resolve from review:

1. The branch is behind main (mergeable_state: behind), so it needs a rebase before it can merge cleanly.
2. @carlesarnal's Changes requested is still open - that one is his to clear.

Once the rebase is done and carlesarnal re-checks, this should be ready to go.

EricWittmann (commented):
Thanks for your persistence on this, @jainarjav80-sys — and apologies that this review cycle has stretched as long as it has. Authentication is a sensitive area of the codebase that requires extra scrutiny, and your willingness to iterate through 10+ rounds of feedback from multiple reviewers is genuinely appreciated. We're close to the finish line.

The core implementation is solid: the `Instance<>` injection with `isResolvable()` guard for `FormAuthenticationMechanism` mirrors the existing `kubernetesClient` pattern, the priority validation warnings are a nice addition, and the CSRF documentation in `@Info` is the right call. The test coverage (chain-building unit tests + `@QuarkusTest` with `FormAuthTestProfile`) is well-structured.

A few items to address before we can merge:

### Must fix

1. **Revert `BasicAuthenticationMechanism` back to direct injection.** This PR changes `basicAuthenticationMechanism` from a direct `@Inject` to `Instance<BasicAuthenticationMechanism>` with an `isResolvable()` guard. That's a behavioral change to an existing mechanism — if `BasicAuthenticationMechanism` ever isn't resolvable, basic auth would be silently dropped instead of failing at startup. The `Instance<>` pattern is correct for `FormAuthenticationMechanism` (which may or may not be present depending on config), but `BasicAuthenticationMechanism` is a Quarkus built-in that's always available. Please revert this to keep the PR focused on form auth only.

### Should fix

2. **Tighten the `testInvalidLogin` assertion.** The current `assertTrue(response.statusCode() == 302 || response.statusCode() == 401)` accepts too many outcomes and can't detect a regression. Please assert the single specific status code you actually observe, so the test fails if behavior changes unexpectedly.

### Nice to have

3. **Add a brief comment in `FormAuthTestProfile` explaining why `role-source=token` works with embedded users.** Future readers will find it confusing that a role source named "token" works without any JWT/OIDC token. A one-liner noting that `TokenRoleProvider` delegates through `SecurityIdentity.hasRole()` would save someone a debugging session later.

Everything else looks good. Once these are addressed, we should be ready to merge. Thanks again for your patience!

jainarjav80-sys:
Thanks for the thorough review and guidance, @EricWittmann!

I've pushed the final set of requested changes:

1.Reverted BasicAuthenticationMechanism back to direct injection in AppAuthenticationMechanism.java and updated the corresponding test mocks.

2.Tightened the testInvalidLogin assertion in FormAuthTest.java to explicitly assert the 302 status code instead of a range.

3.Added an explanatory comment in FormAuthTestProfile.java regarding why role-source=token works with the embedded users for testing.

(Note: The current CI failure in the Build Java Application (no tests) job is an unrelated race condition in the schema-util parallel build where schema-util-json fails to find the schema-util-common test-jar. I've left it as-is rather than modifying the CI workflows or unrelated POMs in this PR.)

Let me know if there's absolutely anything else needed!

jainarjav80-sys:
Hi @EricWittmann,

I applied the change to revert BasicAuthenticationMechanism back to a direct @Inject, but unfortunately, it causes the app-auth unit tests to fail on startup with an UnsatisfiedResolutionException (which is why the CI is currently failing).

It seems that because quarkus.http.auth.basic defaults to false in this project, Quarkus actually eliminates the BasicAuthenticationMechanism bean at build time for any test profiles that don't explicitly enable it (like our new FormAuthTestProfile). Because the bean is removed, attempting to inject it unconditionally into AppAuthenticationMechanism crashes the application during those tests.

Given this, it looks like our original defensive pattern using Instance<BasicAuthenticationMechanism> with an isResolvable() check might actually be necessary to handle these different test configurations safely.

How would you like to proceed? Should we revert back to using the Instance<> injection, or would you prefer we explicitly enable basic auth in the new test profile so the direct injection always succeeds?

EricWittmann:
ah sorry for that @jainarjav80-sys - revert back to the defensive approach using Instance sounds reasonable.  :+1:

jainarjav80-sys:
Thanks @EricWittmann ! No worries at all. I've just reverted AppAuthenticationMechanism back to the defensive Instance<> injection approach and pushed the update. The CI should be fully green now. Let me know if there's anything else you need before we can merge this!

EricWittmann:
Thanks for your patience @jainarjav80-sys

PR #8615, opened 2026-07-14
fix: add missing owner-change authorization check in V3 updateArtifactMetaData

### LINKED ISSUE #8596: [BUG] V3 API updateArtifactMetaData allows any Write-level user to change artifact ownership without Admin/Owner authorization
## Description

**Registry Version**: latest (also affects snapshot/main branch)
**Persistence type**: sql (also reproducible with in-memory and kafkasql)

The V3 REST API endpoint `PUT /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}` (`updateArtifactMetaData`) is missing an authorization check when a user changes the `owner` field via `EditableArtifactMetaData`. Any user with `Write` (Developer) role can transfer ownership of any artifact to any arbitrary user, without being an Admin or the current Owner.

The code at `app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java` (line ~583-589) has a `TODO` comment acknowledging the missing check, but the check was never implemented:

```java
if (data.getOwner() != null) {
    if (data.getOwner().trim().isEmpty()) {
        throw new MissingRequiredParameterException("Owner cannot be empty");
    } else {
        // TODO extra security check - if the user is trying to change the owner, fail unless they are
        // an Admin or the current Owner
    }
}

### PR DESCRIPTION
**Pull request by Anujkumar9081**
The V3 API endpoint `updateArtifactMetaData` allows any user with Write
(Developer) role to change artifact ownership without verifying they are
an Admin or the current Owner. This is an authorization bypass.

The fix adds an explicit check that mirrors the V2 API's
`updateArtifactOwner` endpoint (which correctly uses
`AuthorizedLevel.AdminOrOwner`): before allowing the `owner` field to be
changed, verify the current user is either an Admin (via `AdminOverride`
or RBAC) or the current artifact Owner. If neither, throw a
`ForbiddenException`.

Closes #8596

### CODE REVIEW

**Comment by paoloantinori** [comment]
@Anujkumar9081 Thanks for fixing the TODO — the security gap is real and the approach is right. A few issues to address:

**MAJOR — Missing tests**
Security changes require tests. Please add:
- Test: non-admin, non-owner user attempts to change owner → expects 403
- Test: current owner changes owner → succeeds
- Test: admin changes owner → succeeds
- Test: user updates metadata WITHOUT changing owner → succeeds (no auth check)

**MAJOR — Leaks user identity in error message**
```java
throw new ForbiddenException("User " + currentUser + " is not authorized...");
```
CLAUDE.md rule: "Never expose stack traces or internal errors to API clients." The username should not be in the error message — it reveals identity information. Use a generic message like `"Only the current owner or an admin can transfer artifact ownership."`.

**MINOR — Auth check doesn't match V2 pattern exactly**
V2 uses `@Authorized(level = AuthorizedLevel.AdminOrOwner)` on the separate `updateArtifactOwner` endpoint. Your inline check is functionally equivalent but duplicates the authorization logic. Consider whether the `@Authorized` interceptor's `AdminOrOwner` level could be reused here instead of manual checks.

**MINOR — Null safety**
`securityIdentity.getPrincipal().getName()` — if auth is disabled or anonymous access is allowed, `getPrincipal()` could return null. Add a null guard.

carlesarnal (changes requested):
The authorization check is needed and the implementation handles the edge cases correctly — null principal check is in place, null owner allows claiming.

A couple of things:

- Consider using `@Authorized(level = AuthorizedLevel.AdminOrOwner)` annotation for consistency with other endpoints, if the annotation supports this use case. If manual checks are required here, a brief comment explaining why would help future readers.

- Review whether exposing "Only the current owner or an admin can transfer artifact ownership" is too much info for your security model, or if that level of detail is fine.

- The test coverage looks thorough — admin, owner, non-owner, and ownership transfer cases are all covered.

Anujkumar9081:
Thanks @paoloantinori and @carlesarnal for the detailed review! I've updated the PR and addressed all of the requested changes:

1. **Null Safety**: Added a null check for `principal` before calling `getName()` to prevent any potential NPEs. 
2. **Generic Error Message**: Removed the `currentUser` name from the `ForbiddenException` message so we no longer leak user identity in the API response. It now returns a generic error.
3. **Null Owner / Claiming Behavior**: Added a comment clarifying the behavior for when `currentOwner` is null. Since it means the artifact is unowned, we allow any user with Write-level access to claim it.
4. **V2 Auth Pattern & Manual Checks**: Added a comment explaining why we couldn't use the `@Authorized(level = AuthorizedLevel.AdminOrOwner)` annotation directly on the method like V2 does. Because this V3 endpoint handles updates for other metadata fields (name, description, labels), using that annotation would block standard Write-level users from editing those fields. The manual inline check is strictly required here to isolate the owner check while keeping the rest of the metadata editable by developers.
5. **Unit Tests**: Added the requested unit tests to `SimpleAuthTest.java` (under `testUpdateArtifactOwnerOnlyByOwner`) covering all scenarios:
    * Non-admin/non-owner attempting to change owner (expects 403)
    * User updating metadata WITHOUT changing the owner (expects success / null owner case)
    * Current owner changing the owner (expects success)
    * Admin changing the owner (expects success)

Everything has been pushed to the branch. Let me know if there's anything else you'd like me to tweak!

jsenko:

Anujkumar9081:
Thanks @jsenko! I've rebased the branch onto the latest main as suggested. The FOSSA checks should hopefully pass now

Anujkumar9081:
I investigated the `ForbiddenException` that was causing `testUpdateArtifactOwnerOnlyByOwner` to fail in the `app-kafkasql` and `app-other` modules, and traced it back to a configuration issue in the test environment rather than a logic flaw in the code.

**What was happening:**
In the test, we authenticate as `developer-2-client` to verify that a non-owner with Write permissions can still update basic artifact metadata (like the name/description). However, in the test Keycloak configuration (`utils/tests/src/main/resources/realm.json`), `developer-2-client` was missing a mapped user entry with the `sr-developer` role. 

Because it lacked the `sr-developer` role, the test environment treated it as an unprivileged user without Write access. This caused the `AuthorizedInterceptor` to immediately block the request with a `ForbiddenException` (failing the RBAC check) before it even reached our custom Owner-Based Access Control logic in `GroupsResourceImpl`.

**The Fix:**
I added a proper user configuration for `developer-2-client` in `realm.json` and assigned it the `sr-developer` role, matching the setup for `developer-client`. The test client now has the correct Write permissions, which allows the test to properly execute and verify our metadata update logic.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -584,8 +597,26 @@ public void updateArtifactMetaData(String groupId, String artifactId, EditableAr
             if (data.getOwner().trim().isEmpty()) {
                 throw new MissingRequiredParameterException("Owner cannot be empty");
             } else {
-                // TODO extra security check - if the user is trying to change the owner, fail unless they are
-                // an Admin or the current Owner
+                // Security check: only an Admin or the current Owner can change ownership.
+                // We cannot use @Authorized(level = AuthorizedLevel.AdminOrOwner) on the method
+                // because non-owners with Write access need to be able to update other metadata (name, description, labels).
+                // This manual check ensures only Admin or Owner can change the owner field.
+                boolean isAdmin = adminOverride.isAdmin()
```
paoloantinori, line 604: The owner is read here and persisted later by `storage.updateArtifactMetaData(...)` further down the method. Between the read and the write a concurrent ownership change can slip through (classic authz TOCTOU). Low severity given the narrow window, but for a security path consider validating the owner atomically inside the storage update (same transaction) or re-reading immediately before the write.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -584,8 +597,26 @@ public void updateArtifactMetaData(String groupId, String artifactId, EditableAr
...
+                        || (authConfig.isRbacEnabled() && rbac.isAdmin());
+                if (!isAdmin) {
```
paoloantinori, line 606: `currentOwner.equals(currentUser)` assumes the stored owner uses the exact same representation as `securityIdentity.getPrincipal().getName()` (case, format). If owner is ever populated from a different identity source (e.g. LDAP, a fully-qualified subject), a legitimate owner would be denied. Worth normalizing both sides, or confirming owner is always stored as `principal.getName()`.

app/src/test/java/io/apicurio/registry/auth/SimpleAuthTest.java
```diff
@@ -581,6 +581,36 @@ public void testUpdateArtifactOwnerOnlyByOwner() throws Exception {
...
+        assertEquals("developer-2-client", amd.getOwner());
+
+        // Admin changes owner (should succeed)
+        var client_admin = RegistryClientFactory.create(
+                RegistryClientOptions.create(registryV3ApiUrl, vertx)
+                .oauth2(authServerUrlConfigured, KeycloakTestContainerManager.ADMIN_CLIENT_ID, "test1"));
+        
+        EditableArtifactMetaData eamd4 = new EditableArtifactMetaData();
+        eamd4.setOwner("admin-client");
+        client_admin.groups().byGroupId(groupId).artifacts().byArtifactId(artifactId).put(eamd4);
```
paoloantinori, line 610: This test asserts owner-change allowed for an admin, but the admin client used here resolves via `AdminOverride`. The `authConfig.isRbacEnabled() && rbac.isAdmin()` arm of the `isAdmin` check isn't covered. If RBAC-admin-without-AdminOverride is a supported config, adding a case for it would lock in that branch.

paoloantinori (commented):
Good fix for a real authorization gap, and thanks for addressing the earlier review feedback (null-safety on `principal`, the claim-when-unowned case, and the expanded `SimpleAuthTest` coverage). One design-level concern worth flagging, plus a test observation.

- **Time-of-check to time-of-use window on the owner check.** The authorization decision reads the current owner with `storage.getArtifactMetaData(...)` and then, several statements later, the method persists the new metadata via `storage.updateArtifactMetaData(...)`. Between the read and the write, a concurrent owner-transfer (or owner change through another path) can change the owner, so a user who is the owner at check time but not at write time could still transfer ownership, or vice-versa. This is the classic authz TOCTOU. It's low-severity for most deployments (the window is small and it requires a concurrent privileged action), but for a security-critical path it's worth either (a) performing the ownership check atomically inside the storage-level update (e.g. an `updateArtifactMetaDataIfOwner(...)` that validates in the same transaction), or (b) at minimum re-reading the owner immediately before the write. Worth confirming whether the V2 `updateArtifactOwner` path has the same shape, so the two stay consistent.

- **Identity comparison is case- and principal-format sensitive.** `currentOwner.equals(currentUser)` compares the stored owner string against `securityIdentity.getPrincipal().getName()`. If the owner field is ever populated with a different identity representation than `principal.getName()` (e.g. a fully-qualified subject, a different case from an LDAP sync), a legitimate owner would be denied and the claim-by-others path wouldn't trigger. The test uses Keycloak principals in both places so it won't catch a mismatch. Worth a quick check that owner is always stored in the same form `principal.getName()` returns — or normalizing both sides.

- **Test covers the happy paths but not the failure path for the admin-via-RBAC branch.** `testUpdateArtifactOwnerOnlyByOwner` exercises owner-change-by-non-owner (rejected), change-by-current-owner (allowed), and change-by-admin (allowed). The admin in the test succeeds via `AdminOverride` (the keycloak admin client), so the `authConfig.isRbacEnabled() && rbac.isAdmin()` arm of the `isAdmin` computation isn't asserted. If RBAC-based admin is a supported configuration, a test where the actor is an RBAC-admin-but-not-AdminOverride would close the gap.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
Anujkumar9081: I checked the V2 `updateArtifactOwner` path to see if it has the same shape. In V2, the authorization is handled via the `@Authorized(level = AuthorizedLevel.AdminOrOwner)` annotation. This means the `AuthorizedInterceptor` reads the artifact metadata from storage to check the owner, and then later the resource method executes `storage.updateArtifactMetaData(...)`. So V2 actually has the exact same TOCTOU window (read in interceptor -> write in endpoint). Since it's consistent with the existing V2 architecture and low severity, we might want to defer atomic storage-level ownership checks (e.g., `updateArtifactMetaDataIfOwner`) to a separate, broader security issue if we want to tighten it across both versions.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
Anujkumar9081: The owner string stored in the database originates directly from `securityIdentity.getPrincipal().getName()` when the artifact is initially created (handled by the interceptor/storage layer). Because the stored owner is always populated from this exact same method, the case and principal format should be naturally identical and safe to compare via `.equals()`.

app/src/test/java/io/apicurio/registry/auth/SimpleAuthTest.java
Anujkumar9081: That's a great observation about the test bypassing the `authConfig.isRbacEnabled() && rbac.isAdmin()` arm by using the `admin-client` (AdminOverride). I'd be happy to add an additional test asserting that a standard user mapped to the `sr-admin` role can also successfully change the owner if you'd like that included in this PR!

paoloantinori:
Thanks for the follow-up — this head addresses the earlier feedback well: the `principal` null guard, the generic `ForbiddenException` message (no user-identity leak), and the claim-when-unowned comment all look right. The expanded `SimpleAuthTest` (dev2 updating metadata without touching owner, owner transfer, admin transfer) covers the right paths.

One thing to confirm before merge: the latest commit (61cf0c9) has a failing test-suite run flagged by CI, which you traced to a realm-config issue in `app-kafkasql` / `app-other`. Worth making sure that's green (and the branch rebased on current main) so the new cases actually exercise. Otherwise the implementation looks correct to me.

Anujkumar9081:
Thanks @paoloantinori! 

I have rebased the branch onto current `main` to clean up the history, and I also added that extra test case in `SimpleAuthTest` to verify the `isRbacEnabled() && rbac.isAdmin()` path using a standard `sr-admin` user (bypassing AdminOverride). 

Everything should be green and ready to go now! Let me know if you need anything else!

paoloantinori:
Thanks for the rebase and the extra `sr-admin` RBAC test — the new head looks good from my side: the `AdminOverride || (rbac && rbac.isAdmin())` path is now covered, the claim-when-unowned case is preserved, and the `ForbiddenException` message stays generic. From a review standpoint I have nothing further to add.

Anujkumar9081:
@carlesarnal @paoloantinori I have addressed all the feedback and added the requested test cases. The branch is rebased and ready to go!

Could you please take a final look, approve the workflow run, and merge this PR if everything looks good? I have already solved another issue and I'm eager to get this one merged so I can raise my next PR. Thanks for all your help with this one!

paoloantinori:
Thanks @Anujkumar9081 — the current head looks good from my side: the `AdminOverride || (rbac && rbac.isAdmin())` path is covered, the claim-when-unowned case is present, and the `sr-admin` RBAC test was added to `SimpleAuthTest`. The null-safety on `principal` and the generic `ForbiddenException` message (no identity leak) from the earlier round are still in place.

One note: I can't approve the workflow run or merge from here — workflow approvals and merges are maintainer actions that need to come from @carlesarnal or @jsenko. The security fix itself and its test coverage are solid.

PR #8619, opened 2026-07-14
test: add MCP well-known discovery endpoint tests (#8613)

### LINKED ISSUE #8613: Missing integration tests for MCP well-known discovery endpoints
## Description

The MCP tool discovery endpoints are already implemented on the well-known resource:

- `GET /.well-known/mcp-tools`
- `GET /.well-known/mcp-tools/{groupId}/{artifactId}`
- `GET /.well-known/schemas/mcp-tool/v1`

The A2A side of the same resource has solid coverage in `WellKnownResourceTest.java` (agent card, public/entitled agents, search). There is no matching coverage for the MCP endpoints. A search under `app/src/test` for `mcp-tools` / `MCP_TOOL` returns nothing.

I exercised these endpoints locally with `apicurio.mcp-tools.enabled=true` and they work. So this looks like a test gap not a broken feature.

Relevant files:
- `app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java`
- `app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java`
- `app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java`

### Environment

Local development against `main`. Checking the well-known resource and the existing A2A `@QuarkusTest` suite.

### Steps to Reproduce

1. Open `WellKnownResourceTest.java` and note that every test hits `/.well-known/agents*` or `/.well-known/a2a`.
2. Search test sources for `mcp-tools` or `ArtifactType.MCP_TOOL`.
3. Confirm there are no integration tests for the MCP well-known endpoints above.

**Expected:** MCP discovery endpoints have integration tests similar to the A2A ones (create an `MCP_TOOL` artifact, get by id, search, schema endpoint, simple 404 cases).

**Actual:** Only the A2A well-known endpoints are covered. The MCP counterparts have no integration tests.

### Logs

N/A - missing test coverage rather than a runtime failure.

namansh70747:
## Summary
Adds coverage for the MCP well known discovery endpoints that already sit next to the A2A ones.

- enable `apicurio.mcp-tools.enabled` in the shared experimental test profile
- add `WellKnownMcpToolsTest` for get search and schema paths including basic failure cases
- assert those MCP endpoints return 404 when the feature flag is off

Fixes #8613

## Test plan
- [x] `./mvnw test -pl app -Dtest=WellKnownMcpToolsTest,ExperimentalFeatureGateTest,WellKnownResourceTest`
- [x] `./mvnw checkstyle:check -pl app`
- [ ] CI smoke and full suite on this PR

### DIFF
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/ExperimentalFeatureGateTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/ExperimentalFeatureGateTest.java
@@ -9,7 +9,8 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 /**
- * Tests that A2A endpoints are blocked when the experimental features gate is disabled (default).
+ * Tests that A2A and MCP well-known endpoints are blocked when experimental features / feature
+ * flags are disabled (default).
  */
 @QuarkusTest
 public class ExperimentalFeatureGateTest extends AbstractResourceTestBase {
@@ -62,6 +63,38 @@ public class ExperimentalFeatureGateTest extends AbstractResourceTestBase {
                 .when()
                 .contentType(CT_JSON)
                 .get("/.well-known/schemas/model-schema/v1")
+                .then()
+                .statusCode(404);
+    }
+
+    @Test
+    public void testMcpToolsSearchBlockedWhenDisabled() {
+        givenAtRoot()
+                .when()
+                .contentType(CT_JSON)
+                .get("/.well-known/mcp-tools")
+                .then()
+                .statusCode(404);
+    }
+
+    @Test
+    public void testMcpToolGetBlockedWhenDisabled() {
+        givenAtRoot()
+                .when()
+                .contentType(CT_JSON)
+                .pathParam("groupId", "some-group")
+                .pathParam("artifactId", "some-tool")
+                .get("/.well-known/mcp-tools/{groupId}/{artifactId}")
+                .then()
+                .statusCode(404);
+    }
+
+    @Test
+    public void testMcpToolSchemaBlockedWhenAllExperimentalDisabled() {
+        givenAtRoot()
+                .when()
+                .contentType(CT_JSON)
+                .get("/.well-known/schemas/mcp-tool/v1")
                 .then()
                 .statusCode(404);
     }
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/WellKnownMcpToolsTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/WellKnownMcpToolsTest.java
@@ -1,0 +1,345 @@
+package io.apicurio.registry.noprofile.rest.mcptools;
+
+import io.apicurio.registry.AbstractResourceTestBase;
+import io.apicurio.registry.noprofile.rest.a2a.ExperimentalFeaturesEnabledProfile;
+import io.apicurio.registry.rest.client.models.CreateArtifact;
+import io.apicurio.registry.rest.client.models.CreateVersion;
+import io.apicurio.registry.rest.client.models.VersionContent;
+import io.apicurio.registry.types.ArtifactType;
+import io.apicurio.registry.types.ContentTypes;
+import io.apicurio.registry.utils.tests.TestUtils;
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import io.restassured.RestAssured;
+import io.restassured.specification.RequestSpecification;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.notNullValue;
+
+/**
+ * Tests for the MCP tool well-known discovery endpoints.
+ */
+@QuarkusTest
+@TestProfile(ExperimentalFeaturesEnabledProfile.class)
+public class WellKnownMcpToolsTest extends AbstractResourceTestBase {
+
+    private String serverRootUrl;
+
+    @BeforeEach
+    public void setUpWellKnown() {
+        int port = ConfigProvider.getConfig().getValue("quarkus.http.test-port", Integer.class);
+        serverRootUrl = "http://localhost:" + port;
+    }
+
+    private RequestSpecification givenAtRoot() {
+        return RestAssured.given().baseUri(serverRootUrl);
+    }
+
+    private static final String SEARCH_DATABASE_TOOL = """
+            {
+                "name": "search_database",
+                "title": "Database Search Tool",
+                "description": "Search the product database with filters",
+                "inputSchema": {
+                    "type": "object",
+                    "properties": {
+                        "query": { "type": "string", "description": "Search query" },
+                        "limit": { "type": "integer", "default": 10 }
+                    },
+                    "required": ["query"]
+                },
+                "outputSchema": {
+                    "type": "object",
+                    "properties": {
+                        "results": { "type": "array", "description": "Search results" },
+                        "total": { "type": "integer", "description": "Total count" }
+                    },
+                    "required": ["results", "total"]
+                },
+                "annotations": {
+                    "title": "DB Search",
+                    "audience": ["user", "assistant"],
+                    "priority": 0.8
+                }
+            }
+            """;
+
+    private static final String GET_WEATHER_TOOL = """
+            {
+                "name": "get_weather",
+                "title": "Weather Lookup",
+                "description": "Get the current weather for a location",
+                "inputSchema": {
+                    "type": "object",
+                    "properties": {
+                        "location": { "type": "string", "description": "City name" },
+                        "units": { "type": "string", "enum": ["celsius", "fahrenheit"] }
+                    },
+                    "required": ["location"]
+                }
+            }
+            """;
+
+    @Test
+    public void testGetRegisteredMcpTool() throws Exception {
+        String groupId = TestUtils.generateGroupId();
+        String artifactId = TestUtils.generateArtifactId();
+
+        createMcpTool(groupId, artifactId, SEARCH_DATABASE_TOOL);
+
+        givenAtRoot()
+                .when()
+                .pathParam("groupId", groupId)
+                .pathParam("artifactId", artifactId)
+                .get("/.well-known/mcp-tools/{groupId}/{artifactId}")
+                .then()
+                .statusCode(200)
+                .body("name", equalTo("search_database"))
+                .body("title", equalTo("Database Search Tool"))
+                .body("description", equalTo("Search the product database with filters"))
+                .body("inputSchema.type", equalTo("object"))
+                .body("inputSchema.properties.query.type", equalTo("string"))
+                .body("inputSchema.required", hasItem("query"));
+    }
+
+    @Test
+    public void testGetRegisteredMcpToolNotFound() {
+        givenAtRoot()
+                .when()
+                .pathParam("groupId", "nonexistent-group")
+                .pathParam("artifactId", "nonexistent-tool")
+                .get("/.well-known/mcp-tools/{groupId}/{artifactId}")
+                .then()
+                .statusCode(404);
+    }
+
+    @Test
+    public void testGetNonMcpToolArtifact() throws Exception {
+        String groupId = TestUtils.generateGroupId();
+        String artifactId = TestUtils.generateArtifactId();
+
+        CreateArtifact createArtifact = new CreateArtifact();
+        createArtifact.setArtifactId(artifactId);
+        createArtifact.setArtifactType(ArtifactType.AVRO);
+
+        CreateVersion createVersion = new CreateVersion();
+        VersionContent content = new VersionContent();
+        content.setContent("{\"type\": \"record\", \"name\": \"Test\", \"fields\": []}");
+        content.setContentType(ContentTypes.APPLICATION_JSON);
+        createVersion.setContent(content);
+        createArtifact.setFirstVersion(createVersion);
+
+        clientV3.groups().byGroupId(groupId).artifacts().post(createArtifact);
+
+        givenAtRoot()
+                .when()
+                .pathParam("groupId", groupId)
+                .pathParam("artifactId", artifactId)
+                .get("/.well-known/mcp-tools/{groupId}/{artifactId}")
+                .then()
+                .statusCode(404);
+    }
+
+    @Test
+    public void testSearchMcpTools() throws Exception {
+        String groupId = TestUtils.generateGroupId();
+        String unique = TestUtils.generateArtifactId().replace("-", "");
+        String tool1Id = "wcmcp-s1-" + unique;
+        String tool2Id = "wcmcp-s2-" + unique;
+
+        createMcpTool(groupId, tool1Id, SEARCH_DATABASE_TOOL);
+        createMcpTool(groupId, tool2Id, GET_WEATHER_TOOL);
+
+        givenAtRoot()
+                .when()
+                .queryParam("name", "*" + unique + "*")
+                .get("/.well-known/mcp-tools")
+                .then()
+                .statusCode(200)
+                .body("count", equalTo(2))
+                .body("tools", hasSize(2))
+                .body("tools.artifactId", hasItems(tool1Id, tool2Id));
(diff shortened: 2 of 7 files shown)


namansh70747:
hi @EricWittmann @carlesarnal @vandanayadav7 this just fills the MCP side of the same well known test gap next to the A2A coverage

EricWittmann (dismissed):
Code review: no issues found. Checked for bugs and CLAUDE.md compliance.

namansh70747:
@EricWittmann the /retry above didn't get picked up (no bot reaction) and Verify is stuck: Decide/Lint cancelled, Verification Gate failing as a result, everything else skipped. Could you re-run the Verify workflow / approve it on this commit (ab7880d)?

EricWittmann:
> @EricWittmann the /retry above didn't get picked up (no bot reaction) and Verify is stuck: Decide/Lint cancelled, Verification Gate failing as a result, everything else skipped. Could you re-run the Verify workflow / approve it on this commit ([ab7880d](https://github.com/Apicurio/apicurio-registry/commit/ab7880ddbc28578a5bb5f1a20d11c615fcfc1365))?

I think our bot workflow may be broken or possibly overwhelmed.  :)  We'll have to look into it over the next couple of days.

namansh70747:
> /retry

Thanks @EricWittmann ,  will watch the Verify run from your /retry and ping only if it stalls again after it finishes.

namansh70747:
/ready

Thanks @EricWittmann, Verify is green after your retry.

carlesarnal (changes requested):
Test coverage is reasonable for the happy path — verifies the well-known endpoint returns expected MCP tool metadata.

Missing error cases: malformed input, invalid pagination parameters, and verifying behavior when experimental features are disabled.

Commit message should drop the issue reference from the title to follow conventional commits: `test(mcp): add well-known discovery endpoint tests`.

namansh70747:
> Tests won't compile. Every `@Test` annotation is replaced with a file path string. Same issue with `@TestProfile`. These look like copy-paste errors from the referenced file content provided in the task.
> 
> Fix in `WellKnownMcpToolsTest.java`:
> 
> * Replace all `@utils/converter/src/test/java/io/apicurio/registry/utils/converter/avro/AvroDataTest.java` with `@Test`
> * Replace `@utils/tests/src/main/java/io/apicurio/registry/utils/tests/KafkasqlTestProfile.java(ExperimentalFeaturesEnabledProfile.class)` with `@TestProfile(ExperimentalFeaturesEnabledProfile.class)`
> 
> Fix commit message:
> 
> * Change from `test: add MCP well-known discovery endpoint tests (#8613)` to `test(mcp): add well-known discovery endpoint tests`
> 
> Test coverage is reasonable but missing error cases. Consider adding tests for malformed input and invalid pagination parameters.

@carlesarnal thank you for the review.

I took another look at `WellKnownMcpToolsTest.java` on this branch and it looks like the annotations are the usual `@Test` / `@TestProfile(ExperimentalFeaturesEnabledProfile.class)`. Verify also compiled and ran green on this commit (Unit + Integration), so I’m a bit unsure whether I might be missing something on my side.

If you are still seeing the file-path annotation issue, I’m happy to dig further. Separately, I’m glad to update the commit message and add a couple of negative-input / pagination tests if that would help.

namansh70747:
> > Tests won't compile. Every `@Test` annotation is replaced with a file path string. Same issue with `@TestProfile`. These look like copy-paste errors from the referenced file content provided in the task.
> > Fix in `WellKnownMcpToolsTest.java`:
> > 
> > * Replace all `@utils/converter/src/test/java/io/apicurio/registry/utils/converter/avro/AvroDataTest.java` with `@Test`
> > * Replace `@utils/tests/src/main/java/io/apicurio/registry/utils/tests/KafkasqlTestProfile.java(ExperimentalFeaturesEnabledProfile.class)` with `@TestProfile(ExperimentalFeaturesEnabledProfile.class)`
> > 
> > Fix commit message:
> > 
> > * Change from `test: add MCP well-known discovery endpoint tests (#8613)` to `test(mcp): add well-known discovery endpoint tests`
> > 
> > Test coverage is reasonable but missing error cases. Consider adding tests for malformed input and invalid pagination parameters.

@carlesarnal thank you for the review.

I took another look at `WellKnownMcpToolsTest.java` on this branch and it looks like the annotations are the usual `@Test` / `@TestProfile(ExperimentalFeaturesEnabledProfile.class)`. Verify also compiled and ran green on this commit (Unit + Integration), so I’m a bit unsure whether I might be missing something on my side.
 
 If you are still seeing the file-path annotation issue, I’m happy to dig further. Separately, I’m glad to update the commit message and add a couple of negative-input / pagination tests if that would help

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/ExperimentalFeatureGateTest.java
```diff
@@ -65,4 +66,36 @@ public void testModelSchemaBlockedWhenExperimentalDisabled() {
...
+                .statusCode(404);
+    }
+
+    @Test
+    public void testMcpToolSchemaBlockedWhenDisabled() {
+        givenAtRoot()
+                .when()
+                .contentType(CT_JSON)
+                .get("/.well-known/schemas/mcp-tool/v1")
+                .then()
```
EricWittmann, line 98: Nit: the method name says "blocked when disabled" but this only passes because *both* A2A and MCP are off in the default profile. The implementation gates the schema endpoint on `a2aConfig.isEnabled() || mcpToolsConfig.isEnabled()`, so if A2A were enabled and MCP disabled, this schema would still be served.

The test is correct for its scenario (everything off → 404), but the name is slightly misleading. Consider renaming to something like `testMcpToolSchemaBlockedWhenAllExperimentalDisabled` to make the precondition explicit, or adding a separate test class with a profile that enables A2A but not MCP to verify the finer-grained gate.

app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/WellKnownMcpToolsTest.java
```diff
@@ -0,0 +1,302 @@
...
+        String groupId = TestUtils.generateGroupId();
+        String artifactId = TestUtils.generateArtifactId();
+
+        createMcpTool(groupId, artifactId, SEARCH_DATABASE_TOOL);
+
+        givenAtRoot()
+                .when()
+                .contentType(CT_JSON)
+                .pathParam("groupId", groupId)
+                .pathParam("artifactId", artifactId)
```
EricWittmann, line 100: Nit (applies throughout): `.contentType(CT_JSON)` on GET requests is a no-op — there is no request body to describe. This follows the existing pattern in `WellKnownResourceTest` and `ExperimentalFeatureGateTest`, so it is consistent, but it is cargo-culting. Consider dropping it from the GET-based tests to keep the intent clear.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownMcpToolsTest.java
```diff
@@ -0,0 +1,302 @@
+package io.apicurio.registry.noprofile.rest.a2a;
```
EricWittmann: This class lives in the `a2a` package, but MCP tools are a separate feature with their own config flag (`apicurio.mcp-tools.enabled`). Co-locating here is pragmatic because of the shared `ExperimentalFeaturesEnabledProfile` and gate test, but it conflates two distinct feature domains under one package name.

Consider a sibling package like `noprofile.rest.mcptools` (or a shared `noprofile.rest.wellknown` package for both A2A and MCP well-known tests) to better signal the domain boundary. The shared profile could move to whichever package both test suites import from.

EricWittmann (commented):
Thanks for the test coverage — solid work overall. A few comments inline.

jsenko:

app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/WellKnownMcpToolsTest.java
```diff
@@ -0,0 +1,302 @@
...
+                .when()
+                .contentType(CT_JSON)
+                .get("/.well-known/mcp-tools")
+                .then()
+                .statusCode(200)
+                .body("count", greaterThanOrEqualTo(2))
+                .body("tools", notNullValue());
+    }
+
+    @Test
```
paoloantinori, line 171: This verifies limit caps the page size, but since offset=0 the test does not confirm offset actually skips entries. Consider a second request with offset=1, limit=2 and asserting the first element differs — catches a regression where offset is silently ignored.

app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/WellKnownMcpToolsTest.java
```diff
@@ -0,0 +1,302 @@
...
+        clientV3.groups().byGroupId(groupId).artifacts().post(createArtifact);
+
+        givenAtRoot()
+                .when()
+                .contentType(CT_JSON)
+                .pathParam("groupId", groupId)
+                .pathParam("artifactId", artifactId)
+                .get("/.well-known/mcp-tools/{groupId}/{artifactId}")
+                .then()
+                .statusCode(404);
```
paoloantinori, line 147: Since /.well-known/mcp-tools is a global search (no group filter) and @QuarkusTest shares state across test methods, count>=2 is resilient but tools could include entries from other tests. Scoping by name (as testSearchMcpToolsByNameWildcard does) would make the assertion deterministic.

paoloantinori (commented):
Solid test coverage addition — happy-path + negative cases (404 when flag off, non-MCP artifact, unknown schema version) are well structured. A few observations on coverage gaps and test isolation:

- **Pagination offset never exercised with offset>0.** `testSearchMcpToolsWithPagination` uses offset=0, limit=2 — verifies limit caps page size but not that offset skips entries. A second request with offset=1 would catch a regression where offset is silently ignored.

- **Global-search tests share registry state across @QuarkusTest methods.** `testSearchMcpTools` / `testSearchMcpToolsWithPagination` call /.well-known/mcp-tools with no group filter, so results include MCP tools from other tests in the same class (no cleanup between tests). count>=N handles this, but tools hasSize(2) only checks page size — does not confirm the returned tools are the ones just registered. Consider asserting a returned artifactId belongs to the current group, or scope by name wildcard like testSearchMcpToolsByNameWildcard does.

- **testSearchMcpToolsByName passes artifactId as the name query parameter** and asserts tools[0].title matches. The inline comment says the query matches artifact name or artifactId — worth confirming this is the intended contract (registry artifactId vs MCP tool name field search_database). If the API should match tool name, passing artifactId could be a false positive.

- The parameters assertion (tools[0].parameters hasItems query, limit) confirms the search response derives a parameters array from inputSchema.properties — a useful contract check since the stored MCP tool definition uses inputSchema, not parameters.

None are blockers. Flagging mainly the pagination offset gap as the most concrete missing case.

namansh70747:
Sorry for the delay, my laptop was damaged and I was offline for a bit. Thank you @EricWittmann @carlesarnal @paoloantinori @jsenko for the careful reviews.

I’ve addressed the feedback: moved the MCP tests into noprofile.rest.mcptools, tightened search/pagination assertions (including offset > 0), added invalid pagination coverage, clarified the name-query contract, renamed the schema gate test, and added a profile that keeps A2A on while MCP is off. Also rebased onto main and updated the commit message to test(mcp): add well-known discovery endpoint tests.

Happy to adjust anything further.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/ExperimentalFeatureGateTest.java
namansh70747: Renamed to `testMcpToolSchemaBlockedWhenAllExperimentalDisabled`, and added a separate A2A-on / MCP-off profile to cover the finer-grained gate.

app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/WellKnownMcpToolsTest.java
namansh70747: Dropped `.contentType(CT_JSON)` from the MCP GET tests.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownMcpToolsTest.java
namansh70747: Moved `WellKnownMcpToolsTest` (and the MCP-off gate test) into `noprofile.rest.mcptools`.

app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/WellKnownMcpToolsTest.java
namansh70747: Added a second request with `offset=1` and assert the first tool differs from the `offset=0` page.

app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/WellKnownMcpToolsTest.java
namansh70747: Scoped the global search / pagination tests by a unique name wildcard and assert the concrete `artifactId`s.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/ExperimentalFeatureGateTest.java
```diff
@@ -9,7 +9,8 @@
 import org.junit.jupiter.api.Test;
 
 /**
- * Tests that A2A endpoints are blocked when the experimental features gate is disabled (default).
+ * Tests that A2A and MCP well-known endpoints are blocked when experimental features / feature
+ * flags are disabled (default).
  */
```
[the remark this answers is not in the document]
namansh70747: Added `McpToolsDisabledProfile` + gate test: MCP endpoints 404; schema still 200 because of the `a2a || mcp` gate.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownMcpToolsTest.java
```diff
@@ -0,0 +1,302 @@
...
+
+        givenAtRoot()
+                .when()
+                .contentType(CT_JSON)
+                .get("/.well-known/mcp-tools")
+                .then()
+                .statusCode(200)
+                .body("count", greaterThanOrEqualTo(2))
+                .body("tools", notNullValue());
+    }
```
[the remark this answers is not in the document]
namansh70747: Now captures the created ids and asserts they appear (via scoped name search).

paoloantinori:
Thanks @namansh70747 — the follow-ups land well. The pagination pair (`testSearchMcpToolsWithPagination` covering `offset=0` and `offset=1`), the `offset > results` case, the malformed-offset assertion, and the new `McpToolsDisabledProfile` covering the "A2A on / MCP off" gate all close the gaps I'd flagged.

One small thing worth a sanity check rather than a change request: `testSearchMcpToolsMalformedOffset` asserts `404` for `offset=not-a-number`. A 4xx is fine to lock in as current behavior, but if the underlying search endpoint is in a position to return `400 Bad Request` for invalid pagination params, that would be the more conventional contract — worth confirming this is intentional rather than just the default fall-through. Not blocking.

The name-query contract comment in `testSearchMcpToolsByName` (artifact name/artifactId vs. the MCP JSON `name` field) is a useful clarification for future readers.

namansh70747:
Thanks @paoloantinori, good catch. That 404 was just JAX-RS fall-through, not intentional. MCP search now returns 400 for bad pagination params, and the test asserts that.

paoloantinori:
Thanks @namansh70747 — the `parsePaginationParam` change is exactly what I had in mind: malformed `offset`/`limit` now return 400 instead of falling through to JAX-RS 404, and `testSearchMcpToolsMalformedOffset` pins that behavior. The clamping (`Math.max(0, …)` for offset, `1..500` for limit) is a reasonable default band. Looks good from my side.

namansh70747:
> Thanks @namansh70747 — the `parsePaginationParam` change is exactly what I had in mind: malformed `offset`/`limit` now return 400 instead of falling through to JAX-RS 404, and `testSearchMcpToolsMalformedOffset` pins that behavior. The clamping (`Math.max(0, …)` for offset, `1..500` for limit) is a reasonable default band. Looks good from my side.

Thanks @paoloantinori, glad it looks good.

paoloantinori:
Thanks again @namansh70747 — glad the pagination hardening landed cleanly. Nothing further from my side; the changes look good.

namansh70747:
Hi @EricWittmann , Verify is waiting on workflow approval for the latest commit (120d97e). Could you approve the run or /retry when you have a moment? Thanks!

EricWittmann:
Thanks @namansh70747 :+1:

PR #8620, opened 2026-07-14
fix(rest): return 400 instead of 500 when rendering an invalid prompt template

Me-Priyank:
## What

`POST /groups/{groupId}/artifacts/{artifactId}/versions/{versionExpression}/render` returns **500** when the stored PROMPT_TEMPLATE content is unparseable or has no `template` field. The request itself is well formed, the stored content simply isn't a usable prompt template, so this should be a 4xx.

Reported privately by email to @EricWittmann and @carlesarnal; opening this PR at Eric's suggestion.
## Why it matters

`PromptRenderingService.render()` wrapped every failure, including content-shaped ones, in a plain `RuntimeException`. That type isn't in `HttpStatusCodeMap`, so `getCode()` falls back to `HTTP_INTERNAL_ERROR`.

`CoreRegistryExceptionMapperService` reports **every** 500 to `ResponseErrorLivenessCheck`, whose `error-threshold` defaults to `1` per 60s window. So repeated render calls against one malformed template can drive `/health/live` down. The 500 response body also surfaced the underlying parser error.

Malformed content is storable because content is only validated on creation when a VALIDITY rule is enabled, which is off by default.

## Changes

`PromptRenderingService`:

- `extractTemplateText`: throw `BadRequestException` instead of `IllegalArgumentException` when the `template` field is missing.
- `parseContent`: throw `BadRequestException` when the content parses as neither YAML nor JSON. No cause is attached, so the response no longer exposes the raw parser message.
- Remove the blanket `catch (Exception) -> RuntimeException`. It produced the 500s, and it would otherwise re-wrap the mapped exception. Genuinely unexpected errors still surface as 500 and are still reported to the liveness check, which is the intent.

`BadRequestException` is already mapped to 400 in `HttpStatusCodeMap`, and is already what this endpoint throws for the wrong-artifact-type case, so no new exception type or mapping is introduced.

## Tests

- `PromptRenderTest`: two new cases asserting **400** (not 500), a template with no `template` field, and unparseable content.
- `PromptRenderingServiceTest`: the two existing error-case tests now assert `BadRequestException` instead of `RuntimeException`; added cases for empty content and non-object content.

Both new REST cases fail with `Expected status code <400> but was <500>` (emitting two `[500 ERROR DETECTED]` liveness reports) without the service change, and pass with it.

## Not changed

A non-textual `template` (e.g. `template: {a: 1}`) still renders as an empty string via `asText()`. It never throws, so it isn't part of this issue, happy to address it separately if you'd prefer.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -4,8 +4,8 @@ import com.fasterxml.jackson.databind.JsonNode;
 import io.apicurio.registry.content.ContentHandle;
 import io.apicurio.registry.rest.v3.beans.RenderPromptResponse;
 import io.apicurio.registry.rest.v3.beans.RenderValidationError;
+import io.apicurio.registry.storage.error.InvalidContentException;
 import jakarta.enterprise.context.ApplicationScoped;
-import jakarta.ws.rs.BadRequestException;
 
 import java.util.ArrayList;
 import java.util.Iterator;
@@ -39,6 +39,9 @@ public class PromptRenderingService {
                                         String groupId, String artifactId, String version) {
         // Parse the template content (could be YAML or JSON)
         JsonNode templateNode = parseContent(content);
+        if (templateNode == null || !templateNode.isObject()) {
+            throw new InvalidContentException("Prompt template content must be a JSON or YAML object");
+        }
 
         // Extract template field and variables schema
         String templateText = extractTemplateText(templateNode);
@@ -73,7 +76,7 @@ public class PromptRenderingService {
             try {
                 return MAPPER.readTree(text);
             } catch (Exception ex) {
-                throw new BadRequestException("Prompt template content is not valid YAML or JSON");
+                throw new InvalidContentException("Prompt template content is not valid YAML or JSON");
             }
         }
     }
@@ -84,7 +87,7 @@ public class PromptRenderingService {
     private String extractTemplateText(JsonNode templateNode) {
         JsonNode templateField = templateNode.path("template");
         if (templateField.isMissingNode()) {
-            throw new BadRequestException("Prompt template is missing 'template' field");
+            throw new InvalidContentException("Prompt template is missing 'template' field");
         }
         return templateField.asText();
     }
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -5,8 +5,7 @@ import io.apicurio.registry.content.ContentHandle;
 import io.apicurio.registry.rest.v3.beans.RenderPromptResponse;
 import io.apicurio.registry.rest.v3.beans.RenderValidationError;
 import jakarta.enterprise.context.ApplicationScoped;
-import jakarta.inject.Inject;
-import org.slf4j.Logger;
+import jakarta.ws.rs.BadRequestException;
 
 import java.util.ArrayList;
 import java.util.Iterator;
@@ -26,9 +25,6 @@ public class PromptRenderingService {
 
     private static final Pattern VARIABLE_PATTERN = Pattern.compile("\\{\\{([^}]+)\\}\\}");
 
-    @Inject
-    Logger log;
-
     /**
      * Renders a prompt template by substituting variables.
      *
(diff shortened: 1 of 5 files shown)


Me-Priyank:
Reopening the fix from #8608 now that #8587 is merged... @EricWittmann @carlesarnal  PTAL whenever you have a moment

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -41,46 +37,44 @@ public class PromptRenderingService {
      */
     public RenderPromptResponse render(ContentHandle content, Map<String, Object> variables,
                                         String groupId, String artifactId, String version) {
-        try {
-            // Parse the template content (could be YAML or JSON)
-            JsonNode templateNode = parseContent(content);
+        // Parse the template content (could be YAML or JSON)
+        JsonNode templateNode = parseContent(content);
```
EricWittmann, line 41: `YAML_MAPPER.readTree()` can return `null` for empty or whitespace-only input (Jackson behavior). When that happens, `extractTemplateText(templateNode)` will NPE — which still surfaces as a 500.

The `testEmptyContent` test passes today only because `""` fails YAML parsing and falls through to JSON, which also fails. But inputs like `"# nothing"` (a YAML comment) or `"   "` (whitespace) will parse successfully via `YAML_MAPPER.readTree()`, return `null`, and bypass your new error handling.

Similarly, `"just a plain string"` is valid YAML (a scalar `TextNode`), so `parseContent` succeeds and the error only triggers later when `.path("template")` returns a missing node. This works but is fragile.

Please add a null/non-object guard after `parseContent` returns:

```java
JsonNode templateNode = parseContent(content);
if (templateNode == null || !templateNode.isObject()) {
    throw new BadRequestException("Prompt template content must be a JSON or YAML object");
}
```

This makes the contract explicit and handles the edge cases cleanly.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -5,8 +5,7 @@
 import io.apicurio.registry.rest.v3.beans.RenderPromptResponse;
 import io.apicurio.registry.rest.v3.beans.RenderValidationError;
 import jakarta.enterprise.context.ApplicationScoped;
-import jakarta.inject.Inject;
-import org.slf4j.Logger;
+import jakarta.ws.rs.BadRequestException;
```
EricWittmann: This introduces `jakarta.ws.rs.BadRequestException` into a `services/`-layer CDI bean — the first usage of a JAX-RS exception outside the `rest/` package in this codebase. It couples the service layer to the HTTP transport.

Since this class is only called from the REST layer today, this is pragmatically fine and I wouldn't block on it. But please consider whether a domain-specific exception (e.g. `InvalidContentException extends RuntimeException`) mapped in `HttpStatusCodeMap` would be more appropriate, keeping the service layer transport-agnostic. If you'd rather keep `BadRequestException` for simplicity, that's acceptable — just wanted to flag the precedent.

EricWittmann (changes requested):
Good fix overall — the root cause analysis, exception choice, and test coverage are solid. Two items to address before merge.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
Me-Priyank: Thanks for flagging @EricWittmann .. Added a `templateNode == null || !templateNode.isObject()` guard right after `parseContent`. You're right that an empty YAML document (whitespace-only or comment-only) makes `readTree` return null and NPE'd into a 500. Covered by two new service tests and one REST test for the comment-only case.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
Me-Priyank: Went with the domain exception: new `InvalidContentException extends RegistryException`, mapped to 400 in `HttpStatusCodeMap`, following the `InvalidArtifactType`/`storage.error` pattern. The service no longer imports `jakarta.ws.rs`

Me-Priyank:
Thanks for teaching @EricWittmann :)) .. applied the fixed and rebased on latest main. Ready for another look

carlesarnal (changes requested):
The code changes look correct. Replacing the generic exception handling with `InvalidContentException`, adding proper null/type checks, and mapping to 400 in `HttpStatusCodeMap` is the right approach. Removing the logger was appropriate since exceptions are no longer being swallowed.

Edge case coverage in the tests is thorough — null values, wrong types, missing variables, nested access on primitives.

One nit: verify DCO sign-off is present.

Me-Priyank:
Thanks @carlesarnal  DCO sign-off is present on both commits .. the earlier failure was an accidental merge commit that slipped onto the branch unsigned. I've rebased on latest main and dropped it, so the history is now two clean signed-off commits. Should be green.

EricWittmann (approved):
Both review comments have been addressed. The null/non-object guard after parseContent handles the edge cases cleanly, and the domain exception approach follows the existing codebase patterns well. Nice work on the thorough test coverage too.

EricWittmann:
:clap:

PR #8624, opened 2026-07-14
feat(config): restrict TLS protocol versions and cipher suites (#4017)

### LINKED ISSUE #4017: Restrict cipher suites and TLS protocol versions

When running Apicurio in my enterprise environment, it needs to pass a pentest. They will check that no SSL/TLS versions are used which are considered insecure, so anything below TLSv1.2 is out.

### Proposed Solution

I would like to have the option to configure Apicurio so that only TLSv1.2 and TLSv1.3 are offered to clients connecting to either API or GUI.
Additionally, I would like to restrict the available cipher suites, analagously to the Kafka setting `ssl.cipher.suites`.

I didn't find anything in the docs, so I assume that this is not possible today, but I would be happy to be proven wrong :-)

Harsh16gupta:
## Description
Resolves #4017. This PR adds two new settings to let administrators configure TLS protocols and cipher suites for secure connections:
* `apicurio.http.ssl.protocols`
* `apicurio.http.ssl.cipher-suites`

These settings are mapped to the underlying Quarkus server settings using a new configuration interceptor. The interceptor ensures that Quarkus's default settings are successfully overridden when these properties are set, while keeping custom configurations intact.

These new options are also registered so they are automatically generated in the configuration reference documentation.

## Verification
* **Unit Tests**: Added `HttpSslConfigInterceptorTest` covering all scenarios (all 8 tests passed).
* **Manual Test**: Verified that running the server with `apicurio.http.ssl.protocols=TLSv1.3` successfully rejects TLS v1.2 connections (returning `Alert 70: protocol version`) while allowing TLS v1.3 connections.

Verification Screenshots:
1. **Successful connection using TLS v1.3**:
Screenshot From 2026-07-14 20-00-18

2. **Rejected connection when trying to use TLS v1.2**:
Screenshot From 2026-07-14 20-00-34


carlesarnal (changes requested):
The interceptor approach for enforcing TLS protocol versions and cipher suites works, but a few things to address:

- The interceptor checks for "DefaultValuesConfigSource" by name to determine when to activate. That's brittle — it depends on internal Quarkus naming that could change between versions. Consider a more robust activation mechanism.

- `CATEGORY_HTTP` in `ConfigPropertyCategory.java` — insert it alphabetically between `CATEGORY_HEALTH` and `CATEGORY_ICEBERG` to match the existing ordering convention.

- Add `@Info` annotations on the new config properties to include them in the generated config docs.

- Add an integration test verifying the TLS config actually takes effect at runtime, not just that the properties are read.

- Newline at end of `META-INF/services/io.smallrye.config.ConfigSourceInterceptor`.

Harsh16gupta:
Thanks for the feedback
I've pushed commits addressing all. 
Also, all unit and integration tests are passing successfully!

jsenko:

app/src/test/java/io/apicurio/registry/tls/HttpSslRuntimeTest.java
```diff
@@ -0,0 +1,64 @@
...
+        // 2. Connection with TLSv1.2 should fail because TLSv1.2 is restricted
+        SSLSocketFactory factory12 = sslContext12.getSocketFactory();
+        Assertions.assertThrows(IOException.class, () -> {
+            try (SSLSocket socket = (SSLSocket) factory12.createSocket(host, port)) {
+                socket.setEnabledProtocols(new String[]{"TLSv1.2"});
+                socket.startHandshake();
+            }
+        });
+    }
+}
```
paoloantinori, line 64: This runtime test covers protocol restriction only. The cipher-suite counterpart (set `apicurio.http.ssl.cipher-suites` in the profile, then assert an `SSLSocket` offering only a disallowed cipher fails the handshake) is missing — and that path is currently exercised solely by the mocked interceptor unit test. For a security feature I'd add the runtime test before merge.

app/src/main/java/io/apicurio/registry/services/HttpSslConfig.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Optional;
+
+import static io.apicurio.common.apps.config.ConfigPropertyCategory.CATEGORY_HTTP;
+
+@Singleton
+public class HttpSslConfig {
+
+    @ConfigProperty(name = "apicurio.http.ssl.protocols")
+    @Info(category = CATEGORY_HTTP,
+          description = "Comma-separated list of enabled SSL/TLS protocol versions "
```
paoloantinori, line 29: Both properties are `Optional<String>` with no validation. A typo'd cipher name (e.g. casing, an unsupported suite) is silently accepted here and only fails at the first real TLS handshake. A startup-time sanity check or at least a documented failure mode would help operators diagnose misconfiguration.

paoloantinori (commented):
The interceptor design is sound and the protocol path is well covered by `HttpSslRuntimeTest`. One significant gap stands out for a security-facing feature, plus a couple of smaller observations.

- **No runtime test for cipher-suite restriction.** `HttpSslRuntimeTest` exercises only the `apicurio.http.ssl.protocols` path (TLSv1.3 enforced, TLSv1.2 rejected). The `apicurio.http.ssl.cipher-suites` path is validated only by `HttpSslConfigInterceptorTest`, which mocks `ConfigSourceInterceptorContext` and therefore never confirms that a configured cipher list actually constrains the TLS handshake end-to-end. Given that cipher-suite names are JVM/JSSE-specific and easy to mis-specify, a green interceptor unit test does not prove the feature works at runtime. Mirroring `testTlsProtocolRestrictionAtRuntime` with an `SSLSocket` that advertises a disallowed cipher suite (and asserting the handshake fails with `SSLHandshakeException`) would close this gap.
- **No validation of user-supplied values.** `HttpSslConfig` accepts any string for both properties and forwards it verbatim to `quarkus.http.ssl.protocols` / `quarkus.http.ssl.cipher-suites`. An invalid value (typo, wrong case, unsupported cipher) surfaces only as a late SSL error at first handshake rather than at startup. Worth considering whether a `@Startup` check (or at least documenting the failure mode) is appropriate.
- **`availableSince = "3.3.1"` in `HttpSslConfig`** — worth double-checking against the target release; if these properties land in 3.3.0 the annotation will mislead the generated config reference.
- `ConfigPropertyCategory` change looks clean (only reordering + new `CATEGORY_HTTP`).

app/src/main/java/io/apicurio/registry/services/HttpSslConfigInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+import io.smallrye.config.ConfigValue;
+import jakarta.annotation.Priority;
+
+import java.util.Map;
+
+@Priority(102)
+public class HttpSslConfigInterceptor implements ConfigSourceInterceptor {
+
+    private static final Map<String, String> QUARKUS_TO_APICURIO = Map.of(
+        "quarkus.http.ssl.protocols", "apicurio.http.ssl.protocols",
```
paoloantinori: This map covers `quarkus.http.ssl.*` but not the parallel `quarkus.management.ssl.*` keys. The registry enables the management interface (`quarkus.management.enabled=true` in `application.properties`); when an operator runs it on a separate port, that listener has independent SSL config that won't inherit these restrictions. Consider adding `quarkus.management.ssl.protocols` / `quarkus.management.ssl.cipher-suites` to the mapping, or documenting the scope.

paoloantinori (commented):
Good security-focused addition — the interceptor cleanly bridges `apicurio.http.ssl.*` to Quarkus, and `HttpSslRuntimeTest` convincingly demonstrates the protocol restriction at the TLS layer. One scope gap worth considering before merge:

- **Management interface SSL not covered.** The interceptor maps only `quarkus.http.ssl.protocols` / `quarkus.http.ssl.cipher-suites`, but the registry runs the Quarkus management interface on a separate path (`quarkus.management.enabled=true` in [`application.properties`](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/resources/application.properties)). When an operator sets `quarkus.management.port` to a dedicated listener (a common production pattern for isolating health/metrics endpoints), that listener resolves its own `quarkus.management.ssl.protocols` / `quarkus.management.ssl.cipher-suites` keys, which this interceptor does not map. In that deployment, setting `apicurio.http.ssl.protocols=TLSv1.3` hardens the main listener but leaves the management port at JVM defaults — which on older JDK builds may still negotiate TLSv1.0/TLSv1.1. Worth either extending the `QUARKUS_TO_APICURIO` map to cover both sets, or at minimum documenting this scope limitation alongside the new config options in `ref-registry-all-configs.adoc`.

app/src/test/java/io/apicurio/registry/tls/HttpSslRuntimeTest.java
Harsh16gupta: Yes it was only covering the protocol restriction. I have added `HttpSslCipherSuiteRuntimeTest` which restricts the server to `TLS_AES_256_GCM_SHA384` only, then asserts a client offering only `TLS_AES_128_GCM_SHA256` fails the handshake.

app/src/main/java/io/apicurio/registry/services/HttpSslConfig.java
Harsh16gupta: Yes no validation so I have added a `@Observes StartupEvent` method that validates configured values against `SSLContext.getDefault().getSupportedSSLParameters()` and logs a WARN for any unrecognized protocol or cipher name. 

also updated the `@Info` descriptions to document the failure mode.

app/src/main/java/io/apicurio/registry/services/HttpSslConfigInterceptor.java
Harsh16gupta: ok,
extended the `QUARKUS_TO_APICURIO` map to also cover `quarkus.management.ssl.protocols` and `quarkus.management.ssl.cipher-suites`, so both listeners are hardened by the same config.

paoloantinori (dismissed):
All three items from my review are properly addressed:

1. **Cipher suite runtime test** — `HttpSslCipherSuiteRuntimeTest` covers allowed-cipher-succeeds and disallowed-cipher-fails at the handshake level, with a dedicated `HttpSslCipherSuiteTestProfile`. This was the gap I flagged — now both protocol and cipher restriction have real TLS integration tests.

2. **Startup validation** — `HttpSslConfig.onStartup()` validates configured values against `SSLContext.getDefault().getSupportedSSLParameters()` and logs WARN for unrecognized protocol/cipher names. The `@Info` descriptions document the failure mode. This gives operators early feedback on misconfig without preventing startup (the right tradeoff for a security hardening feature).

3. **Management listener** — `QUARKUS_TO_APICURIO` map now covers `quarkus.management.ssl.protocols` and `quarkus.management.ssl.cipher-suites`, so the management interface gets the same TLS restrictions. Unit tests verify both management mappings.

Config properties follow project conventions (`apicurio.http.ssl.*` naming, `@Info` with `CATEGORY_HTTP`, docs regenerated). Interceptor correctly respects explicit Quarkus settings over Apicurio defaults. LGTM.

Harsh16gupta:
Thankyou maintainers for such a detailed review, the pr is ready for the final review.

Harsh16gupta:
Hey @carlesarnal, can you please review this pr

EricWittmann:
Thanks @Harsh16gupta — nice work. The interceptor approach is clean, the startup validation is a good safety net for operators, and the runtime tests convincingly prove the restrictions work at the TLS handshake level. Appreciate your patience and responsiveness to the review feedback. Merged!

Issue #8022, opened 2026-05-18
CLI: Auto context update — set group/artifact in context after operations


vandanayadav7:
## Description

When a user creates or gets a group/artifact, automatically update the current context with that group/artifact ID. This reduces repetitive `-g` and `-a` flags in subsequent commands.

## Example

```bash
acr group add test-group
# Group "test-group" is automatically set in the context.
acr artifact add -f user-v1.avsc test-artifact
# Instead of: acr artifact add -g test-group -f user-v1.avsc test-artifact
acr artifact version add -f user-v2.avsc 2
# Instead of: acr artifact version add -g test-group -a test-artifact -f user-v2.avsc 2
```

## Acceptance Criteria

- Auto context update is configurable via `acr config`
- Group ID is saved to context after group create/get
- Artifact ID is saved to context after artifact create/get
- Disabled by default

## Parent Epic

- #7162

prakharsingh-74:
@vandanayadav7 and @carlesarnal, Can you please assign this issue to me, and I'm raising a PR for this issue soon.

NobleCoder69:
Hi @vandanayadav7 I'd like to work on this issue if it's still available. Could I be assigned to it?

Priyanshubhartistm:
Hi @vandanayadav7, I’m interested in working on this issue. If it’s still available, could you please assign it to me? I’d be happy to submit a PR. 
Thank you!

PR #8625, opened 2026-07-14
feat(cli): add auto context update for group and artifact operations

prakharsingh-74:
fixes - #8022 

# Description

This PR implements the **CLI: Auto context update** feature. When a user creates or retrieves a group or artifact, the CLI automatically updates the active context with those IDs, reducing the need to specify repetitive `-g` (group) and `-a` (artifact) flags in subsequent operations.

### Key Changes
1. **Context Update Feature**: Added logic to automatically save Group ID to active context after group create/get operations, and Group ID + Artifact ID to active context after artifact create/get operations.
2. **Configuration**: Configured via the `auto-context-update` CLI property (disabled by default).
3. **Windows Test Execution Fixes**: Resolved classpath resource resolution issues on Windows that were causing `InvalidPathException` during test setup.

---

# File Changes

### CLI Core Logic
* **`cli/src/main/java/io/apicurio/registry/cli/common/IdUtil.java`**
  * Added `updateGroupContext(String groupId, Config config)` and `updateArtifactContext(String groupId, String artifactId, Config config)` helper methods to check if `auto-context-update` is active and write context modifications.
* **`cli/src/main/java/io/apicurio/registry/cli/group/GroupCreateCommand.java`** & **`GroupGetCommand.java`**
  * Integrated call to `IdUtil.updateGroupContext` upon successful completion of commands.
* **`cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java`** & **`ArtifactGetCommand.java`**
  * Integrated call to `IdUtil.updateArtifactContext` upon successful completion of commands.

### Verification and Test Setup
* **`cli/src/test/java/io/apicurio/registry/cli/AutoContextUpdateTest.java`** *(New)*
  * Added unit test suite covering disabled-by-default behavior, group updates (create/get), and artifact updates (create/get).
* **`cli/src/test/java/io/apicurio/registry/cli/AbstractCLITest.java`** & **`cli/src/test/java/io/apicurio/registry/cli/ConfigPropertyCommandTest.java`**
  * Updated resource path retrieval from `.getPath()` to `.toURI()` to fix `InvalidPathException` failures when running tests in Windows environments.

---

image


carlesarnal (changes requested):
The feature is useful for CLI UX — auto-updating context after group/artifact operations saves manual steps.

Code quality:
- `IdUtil.java` lines 70, 84: `Boolean.parseBoolean` already handles null, so `getOrDefault` with a default is redundant. Simplify.
- Cache `config.read()` result instead of calling it twice (lines 71-72, 85-86).
- Add null parameter validation for `groupId`, `artifactId`, `config`.

The `toURI()` fix in `AbstractCLITest.java` properly handles spaces in paths.

Test coverage is adequate for happy paths but missing edge case validation — null inputs and config write failures. Not blocking, but worth adding.

Implementation needs the cleanup above before merge.

prakharsingh-74:
@carlesarnal sir, can you review the PR again?

jsenko:
## Re-review: carlesarnal's feedback

All three items from the previous review are properly addressed:

1. **Simplified `Boolean.parseBoolean`** — uses `.get()` instead of `.getOrDefault()` since `Boolean.parseBoolean(null)` returns `false`. Good.
2. **Cached `config.read()`** — both methods now read once into `configModel`. Good.
3. **Null validation** — `Objects.requireNonNull` on all parameters. Good.
4. **Edge case tests** — `testAutoContextUpdateNullValidation` and `testAutoContextUpdateWriteFailure` cover the suggested edge cases.

## Additional observations

**Style — FQN in tests**: `AutoContextUpdateTest` uses fully-qualified class names instead of imports in several places (e.g., `org.junit.jupiter.api.Assertions.assertThrows`, `io.apicurio.registry.cli.common.IdUtil`, `io.apicurio.registry.cli.config.ConfigModel`). These should be imports for consistency with the rest of the test codebase. Minor, non-blocking.

**Error handling — context update failure shouldn't fail the command**: In the current implementation, if `config.write()` throws during the auto-context update, the exception propagates and the command reports failure — even though the primary operation (group/artifact create/get) already succeeded. The context update is a secondary side effect. Consider wrapping the context update calls in the commands with a try/catch that logs a warning instead of failing the command. For example, a user who successfully creates an artifact shouldn't see an error just because the context file couldn't be written.

cli/src/main/java/io/apicurio/registry/cli/common/IdUtil.java
```diff
@@ -65,4 +66,40 @@ public static void validateVersion(final RegistryClient client, final String gro
...
+        Objects.requireNonNull(config, "config must not be null");
+
+        final var configModel = config.read();
+        if (configModel != null && Boolean.parseBoolean(configModel.getConfig().get("auto-context-update"))) {
+            final var contextName = configModel.getCurrentContext();
+            if (!isBlank(contextName)) {
+                final var context = configModel.getContext().get(contextName);
+                if (context != null) {
+                    context.setGroupId(groupId);
+                    config.write(configModel);
```
paoloantinori: A `config.write()` failure here propagates out of the command, even though the group/artifact was already created server-side. For an opt-in convenience feature, consider catching + logging so a successful `artifact create` still exits 0. The `testAutoContextUpdateWriteFailure` test currently locks in the propagate behaviour — worth confirming that's really intended.

cli/src/main/java/io/apicurio/registry/cli/common/IdUtil.java
```diff
@@ -65,4 +66,40 @@ public static void validateVersion(final RegistryClient client, final String gro
...
+
+        final var configModel = config.read();
+        if (configModel != null && Boolean.parseBoolean(configModel.getConfig().get("auto-context-update"))) {
+            final var contextName = configModel.getCurrentContext();
+            if (!isBlank(contextName)) {
+                final var context = configModel.getContext().get(contextName);
+                if (context != null) {
+                    context.setGroupId(groupId);
+                    context.setArtifactId(artifactId);
+                    config.write(configModel);
```
paoloantinori: Read-modify-write on the config file isn't atomic across concurrent `acr` invocations — two parallel `artifact create` calls can lose one context update (last writer wins). Low-risk for interactive use; worth a one-line note in the `auto-context-update` docs if scripted/concurrent CLI use is expected.

paoloantinori (commented):
The feature is genuinely useful for CLI ergonomics and the opt-in `auto-context-update` flag (disabled by default) is the right call. jsenko's re-review already covered the `Boolean.parseBoolean` simplification. Two things I'd flag that I didn't see raised yet:

- **`updateGroupContext`/`updateArtifactContext` propagate any exception from `config.write()`** — and they are called *after* the actual command succeeded and produced output (`ArtifactCreateCommand` line 132, `ArtifactGetCommand` line 82, `GroupCreate` line 62, `GroupGet` line 51). So a transient config-write failure (read-only `~/.acr`, disk full, concurrent `acr config` edit) would turn a successful `artifact create` into a command that exits non-zero, even though the artifact was already created server-side. The `AutoContextUpdateTest.testAutoContextUpdateWriteFailure` test even asserts the exception escapes. For an opt-in convenience feature, swallowing/logging the write failure (and exiting 0 once the primary operation succeeded) is almost certainly the intended UX; the current behaviour can mislead users into retrying a create that already happened. Worth confirming the intended semantics here.
- **Read-then-modify-then-write on the config file is not atomic across concurrent `acr` invocations.** Two `acr artifact create` running in parallel (scripted) will both `config.read()`, both edit their in-memory model, both `config.write()` — last writer wins and the other artifact's context update is lost. Low probability for interactive use, but worth a note in the property's docs if concurrent CLI use is expected. If `Config` already serializes writes, disregard.
- Nice touch on the `AbstractCLITest`/`ConfigPropertyCommandTest` Windows fix (`.getPath()` -> `.toURI()`) — that's a real latent bug unrelated to the feature and it's good it's fixed here.

Minor: the two helper methods are near-identical; `updateArtifactContext` could delegate to a shared private `writeContext(config, ctx -> { ctx.setGroupId(groupId); ctx.setArtifactId(artifactId); })` to avoid the duplication, but not important.

prakharsingh-74:
@paoloantinori and @jsenko can you review again?

cli/src/main/java/io/apicurio/registry/cli/common/IdUtil.java
```diff
@@ -70,6 +71,50 @@ public static void validateVersion(final RegistryClient client, final String gro
...
+        Objects.requireNonNull(config, "config must not be null");
+
+        try {
+            final var configModel = config.read();
+            if (configModel != null && Boolean.parseBoolean(configModel.getConfig().get("auto-context-update"))) {
+                final var contextName = configModel.getCurrentContext();
+                if (!isBlank(contextName)) {
+                    final var context = configModel.getContext().get(contextName);
+                    if (context != null) {
+                        context.setGroupId(groupId);
```
paoloantinori: Setting only `groupId` here leaves a stale `artifactId` in the context. If the context was `{groupA, artifactA}` and the user runs `acr group get groupB`, the result is `{groupB, artifactA}` — a dangling reference that will 404 on the next context-relative `artifact get`. `updateArtifactContext` sets both fields; this one should symmetrically clear `artifactId` (e.g. `context.setArtifactId(null);`). Worth a test: `artifact create` then `group get <other>` should leave `artifactId` cleared.

cli/src/main/java/io/apicurio/registry/cli/common/IdUtil.java
```diff
@@ -70,6 +71,50 @@ public static void validateVersion(final RegistryClient client, final String gro
...
+                        config.write(configModel);
+                    }
+                }
+            }
+        } catch (Exception ex) {
+            config.getStdErr().print("Warning: Auto-context update failed: " + ex.getMessage() + "\n");
```
paoloantinori: This `config.getStdErr().print(...)` bypasses the `OutputBuffer` that normal command output (including JSON `writeStdErrChunk`) flows through. Since the context update runs before the buffered output is printed, the warning will appear on real stderr ahead of the command's own JSON/table output. Routing it through the `OutputBuffer` would keep diagnostics ordering consistent with the rest of the CLI.

paoloantinori (commented):
Thanks for the follow-up — this head addresses both items from my earlier review (the `config.write()` failure is now caught and turned into a stderr warning rather than propagating after a successful server-side operation, and the concurrency caveat is documented in the README). I have one new concern that I think is worth handling before merge, plus a couple of smaller notes.

- **`updateGroupContext` leaves a stale `artifactId` in the active context.** It sets `context.setGroupId(groupId)` but never touches `artifactId`. So if a user's context is `{groupA, artifactA}` and they run `acr group get groupB` (or `acr group create groupB`), the context becomes `{groupB, artifactA}` — a reference to an artifact that does not exist under `groupB`. The next bare `acr artifact get` (relying on context) will then 404, and the failure won't be obviously tied to the group switch. `updateArtifactContext` correctly sets both fields because it is the more specific operation; the group-level update should symmetrically clear `artifactId` (set it to `null`) since changing the group invalidates any prior artifact reference. Worth adding a test that runs `artifact create` then `group get <otherGroup>` and asserts `artifactId` is cleared.

- **Warning text is emitted outside the `OutputBuffer`, before the command's own output.** `updateGroupContext`/`updateArtifactContext` call `config.getStdErr().print("Warning: ...")` directly, while normal command output (including JSON mode's `writeStdErrChunk`) goes through the `OutputBuffer` and is flushed later via `output.print()`. Concretely in `GroupGetCommand.run` and `ArtifactGetCommand.run`, the context update is invoked *between* the server call and `printGroup`/`printArtifact`, so the warning lands on real stderr before the buffered JSON/table. The JSON stream itself stays valid (stdout vs stderr), but the ordering is inconsistent with how the rest of the CLI reports diagnostics, and a consumer reading both streams interleaved will see the warning first. Routing the warning through the same `OutputBuffer` (e.g. `output.writeStdErrChunk`) would keep ordering consistent. Low severity, but worth aligning.

- **Minor: the two helpers are near-identical.** `updateArtifactContext` could delegate to a shared private method taking a `Consumer<Context>` so the read/flag-check/write/try-catch wrapper isn't duplicated. Not blocking.

The core feature and the opt-in default are good; the stale-`artifactId` point is the one I'd ask to address.

paoloantinori:
Thanks — this head addresses both items from my earlier review: the `config.write()` failure is now caught and turned into a stderr warning rather than propagating after a successful command, and the README documents the non-atomic last-writer-wins behavior for concurrent invocations. @jsenko's re-review already confirmed the `Boolean.parseBoolean` simplification and the remaining points. I don't have further concerns from my side.

prakharsingh-74:
can you please took a look @paoloantinori again?

paoloantinori:
Thanks for the ping — happy to take another look.

I reviewed the current head (`d44611d`) against the two items from my earlier review:
1. The `config.write()` failure path now catches and emits a stderr warning instead of propagating after a successful command — addressed in `IdUtil.updateContext`.
2. The opt-in `auto-context-update` flag is documented (with the "last writer wins" concurrency note) in `cli/README.md` — addressed.

The implementation (`updateContext` → `updateGroupContext` / `updateArtifactContext`, wired into the four create/get commands, plus `AutoContextUpdateTest`) looks consistent and the Windows `toURI()` fix in `AbstractCLITest` is a reasonable portability improvement.

One small thing I'd still like confirmed (not blocking): in `ArtifactGetCommand`, `updateArtifactContext` is called unconditionally after the fetch — including when the caller passed an explicit `-g`/`-a`. That's fine for an opt-in feature, but worth a note in the README that enabling it will overwrite a manually-set context on every `get`. If that's the intended UX, no change needed.

From my side, the direction is right and the earlier feedback is resolved. @carlesarnal has the final call on the merge.

cli/src/test/java/io/apicurio/registry/cli/AutoContextUpdateTest.java
```diff
@@ -0,0 +1,204 @@
...
+        context = config.read().getContext().get("test");
+        assertThat(context.getGroupId()).isEqualTo("group-art-clear-2");
+        assertThat(context.getArtifactId()).isNull();
+    }
+
+    @Test
+    public void testAutoContextUpdateNullValidation() {
+        var dummyOutput = new io.apicurio.registry.cli.utils.OutputBuffer(val -> {}, val -> {});
+
+        org.junit.jupiter.api.Assertions.assertThrows(NullPointerException.class, () -> {
```
jsenko, line 143: LGTM, but one minor code quality issue before merging, please replace fully qualified names with imports here. LLMs often forget to do that.

jsenko (changes requested):

vandanayadav7:
Hi @prakharsingh-74, I reviewed this PR and found a few items:

---

### High Priority

**1. `group get` always clears `artifactId` — even when the group hasn't changed**

`updateGroupContext` (IdUtil.java:102) unconditionally sets `context.setArtifactId(null)` on every group get/create. This means simply inspecting the group you're already working in destroys your artifact context:

```
# User is working with my-app/user-schema
Context: {groupId: my-app, artifactId: user-schema}

$ acr group get my-app          # just checking group info
Context: {groupId: my-app, artifactId: null}   ← artifactId wiped

$ acr artifact get              # expects user-schema from context
ERROR: Artifact ID is required  # user didn't change anything
```

Same thing happens with `group get default` — just looking at the default group wipes whatever artifact you were working with.

I confirmed both scenarios with E2E testing against a live registry.

The clearing logic makes sense when the group **actually changes** (the old artifact doesn't exist in the new group). But when the group is the same, there's no reason to wipe it. Consider:

```java
context -> {
    if (!groupId.equals(context.getGroupId())) {
        context.setArtifactId(null);
    }
    context.setGroupId(groupId);
}
```

This way, switching to a different group still clears the artifact (correct), but re-getting the same group preserves it.

Either way, the README should document that group operations clear the `artifactId` from the context, since users wouldn't expect that from reading the current docs.

---

### Medium Priority

**2. Missing test: `group get default` behavior**

No test verifies what happens when `group get default` is called with auto-context-update enabled. This is the most surprising edge case from Finding 1 — worth adding a test that asserts the expected behavior.

**3. Missing test: no active context scenario**

There is no test for when `auto-context-update` is enabled but no current context is set (e.g., user deleted their context). The code correctly skips the update silently (IdUtil.java:84-87), but this behavior should be verified by a test.

**4. No `.as()` assertion messages on any test assertion**

The test file has 21 `assertThat` calls but zero `.as(withCliOutput(...))` failure descriptions. Every other test class in the CLI suite uses them. When an assertion fails, you get a generic message with no context about which scenario broke.

---

### Minor

**5. Import ordering in `IdUtil.java`**

Lines 3-7 mix `java.*` and `io.apicurio.*` imports:

```java
import io.apicurio.registry.cli.config.Config;
import io.apicurio.registry.rest.client.RegistryClient;
import java.util.Objects;                              // ← java.* between io.* imports
import io.apicurio.registry.cli.utils.OutputBuffer;    // ← io.* after java.*
import java.util.function.Consumer;                    // ← java.* after io.*
```

Convention is to group `io.*` together, then `java.*` together.

**6. `Boolean.parseBoolean("yes")` silently treated as disabled**

`IdUtil.java:83` — `Boolean.parseBoolean` only accepts the literal string `"true"` (case-insensitive). I confirmed via E2E testing that users setting `auto-context-update=yes` or `=1` would have the feature silently disabled. A note in the README that only `"true"` is accepted would help.

---

### Summary

| # | Finding | Priority |
|---|---------|----------|
| 1 | `group get` always clears artifactId even when group hasn't changed | High |
| 2 | Missing test: `group get default` behavior | Medium |
| 3 | Missing test: no active context scenario | Medium |
| 4 | No `.as()` assertion messages (0 of 21) | Medium |
| 5 | Import ordering in IdUtil.java | Minor |
| 6 | `Boolean.parseBoolean("yes")` silently disabled | Minor |

Thanks for the contribution — the feature implementation is solid and well-tested. Happy to help review after changes.

cli/src/main/java/io/apicurio/registry/cli/common/IdUtil.java
```diff
@@ -70,6 +73,45 @@ public static void validateVersion(final RegistryClient client, final String gro
...
+        try {
+            final var configModel = config.read();
+            if (configModel != null && Boolean.parseBoolean(configModel.getConfig().get("auto-context-update"))) {
+                final var contextName = configModel.getCurrentContext();
+                if (!isBlank(contextName)) {
+                    final var context = configModel.getContext().get(contextName);
+                    if (context != null) {
+                        action.accept(context);
+                        config.write(configModel);
+                    }
```
paoloantinori, line 90: This unconditionally clears `artifactId` on every `group get` / `group create`. Consider a user who ran `artifact get -g my-group -a my-artifact` (so the context now points at `my-group:my-artifact`), then runs `group get -g other-group` to inspect something — after that call, their artifact context is silently wiped to `null`, and the next `artifact get` (relying on the auto-context) will fail or target the wrong thing.

Two options worth considering:
1. Only clear `artifactId` when the group actually changed (`!Objects.equals(context.getGroupId(), groupId)`), so switching groups clears the now-stale artifact, but re-inspecting the same group is a no-op.
2. Don't clear `artifactId` at all in `updateGroupContext` — let `artifact get/create` be the only thing that sets it. This is simpler but leaves a stale artifact pointer if the user switches groups via `group get`.

Option (1) seems closer to the intent (artifact context should always be consistent with the group context). Worth confirming the desired behavior either way.

paoloantinori (commented):
Re-reviewing `d44611d` after the ping. The `config.write()` failure path and the `Boolean.parseBoolean` simplification from the earlier rounds are resolved. vandanayadav7's review flagged a UX regression that I agree is worth addressing before this lands — `updateGroupContext` unconditionally nulls `artifactId` on every `group get`/`group create`, which would surprise a user who set an artifact context and then inspected a different group.

paoloantinori:
@prakharsingh-74 Two items that need attention before this can merge:

### 1. jsenko's inline review — FQN cleanup (blocking)

jsenko asked to replace fully qualified class names with imports in `AutoContextUpdateTest.java`. This is still pending — the test file uses `io.apicurio.registry.cli.*` and `org.junit.jupiter.api.Assertions.*` FQNs throughout instead of proper imports. Please fix before the next review round.

### 2. `updateGroupContext` unconditionally clears `artifactId` (bug)

vandanayadav7 raised this and it's a valid concern. In `IdUtil.java`:

```java
public static void updateGroupContext(final String groupId, ...) {
    updateContext(config, output, context -> {
        context.setGroupId(groupId);
        context.setArtifactId(null);  // ← always clears
    });
}
```

The clearing makes sense when switching to a **different** group (the old artifact doesn't exist there), but it shouldn't destroy context when re-inspecting the **same** group:

```bash
$ acr artifact get --group my-app --artifact user-schema
# Context: {groupId: my-app, artifactId: user-schema}

$ acr group get my-app          # just checking metadata
# Context: {groupId: my-app, artifactId: null}  ← silently wiped
```

Suggested fix — only clear when the group actually changes:

```java
context -> {
    if (!groupId.equals(context.getGroupId())) {
        context.setArtifactId(null);
    }
    context.setGroupId(groupId);
}
```

And add a test case that verifies artifact context is preserved when `group get` targets the current group.

PR #8632, opened 2026-07-15
feat(console-plugin): add OpenShift Console plugin with operator integration

### LINKED ISSUE #8484: Add OpenShift Console plugin for Apicurio Registry
## Summary

Create an OpenShift Console dynamic plugin for Apicurio Registry to provide native integration with the OpenShift web console, giving cluster admins first-class visibility and management of registry instances directly from the console.

## Motivation

- **OpenShift integration** — Users expect products with Kubernetes operators to integrate deeply with the OpenShift console
- **Operator synergy** — The existing Apicurio Registry Operator manages CRs but has no console UI; admins must use the standalone UI, CLI, or kubectl to manage instances
- **Cluster-admin persona** — Console plugins serve the cluster-admin use case (provisioning, monitoring, troubleshooting) which is distinct from the developer-facing standalone UI

## Proposed Scope (MVP)

A console plugin focused on the **cluster-admin persona**:

1. **Instance list view** — Show all `ApicurioRegistry` CR instances with status, version, and storage type
2. **Instance detail view** — Health, configuration summary, endpoints, and storage connectivity status
3. **Embedded standalone UI** — For artifact browsing and management, embed the existing standalone UI (pointing at the selected instance's endpoint) rather than building a separate artifact UI

### Explicitly out of scope for MVP
- Duplicating artifact CRUD in a new UI (leverage the existing standalone UI instead)

## Technical Approach

- Use the [OpenShift Dynamic Plugin SDK](https://github.com/openshift/dynamic-plugin-sdk)
- React + PatternFly (consistent with OpenShift console and existing Apicurio UI)
- New module: `console-plugin/` in the repository
- Use Kubernetes API for CR status and management
- Embed the existing standalone UI for artifact browsing, avoiding a second frontend to maintain

## Related

- Existing operator module: `operator/`
- Existing standalone UI: `ui/`

carlesarnal:
## Summary

- Add an OpenShift Console dynamic plugin for managing Apicurio Registry instances natively from the OpenShift web console
- Operator auto-deploys the console plugin when reconciling ApicurioRegistry3 CRs
- Quarkus backend proxy with UserToken passthrough for authenticated access to the registry API

### Console Plugin (`console-plugin/`)
- **Instance list page**: shows all `ApicurioRegistry3` CRs with status, storage type, auth, replicas, API/UI links
- **Instance detail page**: Overview tab (config, conditions, endpoints, labels), YAML editor, Events, and Registry UI tabs
- **Quarkus backend proxy**: reverse proxy at `/proxy/*` forwarding requests to the registry REST API with the user's OpenShift bearer token (UserToken passthrough)
- **Auth-aware UI embedding**: embeds standalone UI via iframe when auth is disabled; shows direct link when auth is enabled
- **Helm chart** for standalone deployment
- **CI workflow** for frontend build verification

### Operator Integration (`operator/`)
- Operator auto-deploys console plugin Deployment, Service, and `ConsolePlugin` CR via JOSDK dependent resources
- `ConsolePlugin` CR includes `proxy` section with `authorization: UserToken`
- `CRDPresentActivationCondition` prevents informer crash on non-OpenShift clusters
- `stop-on-informer-error-during-startup=false` for graceful degradation on plain K8s
- New `spec.consolePlugin.enabled` field (defaults `true`); also requires `REGISTRY_CONSOLE_PLUGIN_IMAGE` to be set
- RBAC rules for `console.openshift.io/consoleplugins`
- Console plugin image configurable via `REGISTRY_CONSOLE_PLUGIN_IMAGE` env var

### Auth Model
The console backend forwards the user's OpenShift OAuth token to the plugin's Quarkus backend, which passes it through to the registry API. If the registry uses a separate Keycloak, the admin configures [identity brokering](https://www.keycloak.org/docs/latest/server_admin/#_identity_broker) so Keycloak accepts OpenShift tokens.

## Test plan

- [ ] Build operator: `cd operator && ../mvnw clean install -DskipTests`
- [ ] Build Quarkus backend: `cd console-plugin/backend && ../../mvnw clean package -DskipTests`
- [ ] Build frontend: `cd console-plugin && npm install && npm run build`
- [ ] Deploy to OpenShift cluster with operator and verify:
  - [ ] Console plugin nav item appears
  - [ ] Instance list page shows ApicurioRegistry3 CRs
  - [ ] Detail page shows status, config, YAML editor, events
  - [ ] Proxy endpoint responds
- [ ] Verify console plugin is NOT deployed on plain Kubernetes
- [ ] Verify `spec.consolePlugin.enabled: false` cleans up resources

Closes #8484


jsenko (changes requested):
## CI Integration

The new `verify-console-plugin.yaml` is a standalone workflow, not integrated into the Verification Gate system:

- Not in `verify.yaml`'s Decide job — no `console-plugin` path filter, no `run-console-plugin` output
- Not in the Verification Gate's `needs:` — the gate doesn't wait for it
- Not called as a reusable workflow — it triggers independently via its own `on: push/pull_request` with path filters

Every other component (UI, Java, operator, SDKs, CLI) goes through `verify.yaml` → Decide → per-component workflow → Verification Gate. A broken console-plugin build won't block merges.

Additionally, the standalone workflow only runs `npm run build` (frontend), but the console-plugin also has a Quarkus backend (`console-plugin/backend/`) which isn't built by this workflow at all.

## Security

**Path traversal in `LocalesResource.java`** — `language` and `namespace` query params are concatenated directly into a classpath path with no sanitization:
```java
String resourcePath = "META-INF/resources/locales/" + language + "/" + namespace + ".json";
```
An attacker can use `../../` sequences to read arbitrary classpath resources.

**Proxy doesn't forward request Content-Type** — `RegistryProxyResource` forwards `Authorization` but not `Content-Type` on POST/PUT requests. The registry API likely needs it for content negotiation.

## Code Issues

1. **Duplicate webpack configs**: both `webpack.config.js` (CJS) and `webpack.config.ts` (TS) exist with nearly identical content. Pick one.

2. **`package-lock.json` in `.gitignore` but referenced in CI**: The workflow caches `console-plugin/package-lock.json`, but `.gitignore` excludes it. The cache won't work, and `npm install` without a lockfile means non-reproducible builds.

3. **`setup-node` downgraded from v7 to v6** in `release-images.yaml` with no explanation.

4. **ConsolePlugin CR name hardcoded to `"apicurio-registry"`**: Since `ConsolePlugin` is cluster-scoped, multiple `ApicurioRegistry3` instances in different namespaces will fight over the same CR, each overwriting the other's service reference.

5. **Operator precedence bug in `RegistryDetailPage.tsx`**:
   ```tsx
   {app?.ingress?.tlsTermination ?? app?.tls?.insecureRequests === "disabled" ? (
   ```
   `===` binds tighter than `??`, so when `tlsTermination` is undefined this evaluates as `(insecureRequests === "disabled")` (a boolean), not the intended conditional check.

6. **`stop-on-informer-error-during-startup=false`** — Added so the operator doesn't crash on non-OpenShift clusters lacking the `ConsolePlugin` CRD, but it applies globally to ALL informers, not just the ConsolePlugin one. A real informer failure (e.g., RBAC misconfiguration for Deployments) would be silently swallowed instead of failing fast.

## Missing Coverage

- No unit tests for the backend proxy (`LocalesResource`, `RegistryProxyResource`)
- No operator tests for the new console plugin resources
- The backend (Maven/Quarkus) isn't built by the CI workflow at all
- Production image configuration: `registry.console-plugin.image` is only set for `%test` and `%dev` profiles. In production it must be passed via env var, but `Configuration.java` reads `registry.console-plugin.image` / `related.image.registry.console-plugin.image` while `install.yaml` sets `REGISTRY_CONSOLE_PLUGIN_IMAGE`. These don't match automatically — Quarkus would need an explicit mapping in `application.properties` like `registry.console-plugin.image=${REGISTRY_CONSOLE_PLUGIN_IMAGE:}`

console-plugin/backend/src/main/java/io/apicurio/registry/consoleplugin/RegistryProxyResource.java
```diff
@@ -0,0 +1,105 @@
...
+        String query = uriInfo.getRequestUri().getRawQuery();
+        String targetUrl = registryApiUrl + "/" + path + (query != null ? "?" + query : "");
+
+        log.debug("Proxying {} -> {}", method, targetUrl);
+
+        HttpRequest.Builder requestBuilder = HttpRequest.newBuilder()
+                .uri(URI.create(targetUrl))
+                .timeout(Duration.ofSeconds(30));
+
+        if (authorization != null && !authorization.isEmpty()) {
```
paoloantinori, line 101: The proxy only forwards the `Authorization` header to the upstream registry — `Content-Type`, `Accept`, and other request headers are dropped. For POST/PUT artifact operations the registry depends on `Content-Type` to parse the payload (e.g. JSON vs YAML vs Avro), so artifact create/update through this proxy would likely be misinterpreted or rejected. Consider capturing `@HeaderParam("Content-Type")` on the POST/PUT methods (and `Accept`) and forwarding them onto `requestBuilder`, the same way `Authorization` is handled here.

console-plugin/backend/src/main/java/io/apicurio/registry/consoleplugin/LocalesResource.java
```diff
@@ -0,0 +1,33 @@
...
+    @GET
+    @Path("/resource.json")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getLocale(@QueryParam("lng") String language,
+                              @QueryParam("ns") String namespace) {
+        if (language == null || namespace == null) {
+            return Response.status(400).build();
+        }
+
+        String resourcePath = "META-INF/resources/locales/" + language + "/" + namespace + ".json";
```
paoloantinori, line 31: `language` and `namespace` are interpolated straight into the resource path with no validation. Values containing path separators or `../` sequences could resolve to classpath resources outside the intended `locales/` directory. Since both come from query parameters, a hostile client controls them. Worth rejecting anything outside `[a-zA-Z0-9_-]+` (or equivalent) before constructing `resourcePath`.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/consoleplugin/ConsolePluginCRResource.java
```diff
@@ -0,0 +1,72 @@
...
+import static io.apicurio.registry.operator.resource.ResourceFactory.RESOURCE_TYPE_SERVICE;
+import static io.apicurio.registry.operator.utils.Mapper.toYAML;
+
+@KubernetesDependent
+public class ConsolePluginCRResource extends CRUDKubernetesDependentResource<ConsolePlugin, ApicurioRegistry3> {
+
+    private static final Logger log = LoggerFactory.getLogger(ConsolePluginCRResource.class);
+
+    public ConsolePluginCRResource() {
+        super(ConsolePlugin.class);
```
paoloantinori: The comment explains the ConsolePlugin is cluster-scoped so it can't take an owner reference back to the namespaced `ApicurioRegistry3`. That's correct, but it also means the plugin CR won't be garbage-collected when the registry CR is deleted — deleting an `ApicurioRegistry3` would leave an orphaned `ConsolePlugin` pointing at a now-gone Service. Worth confirming the intended cleanup path (finalizer on the primary, explicit delete in the activation-condition false branch, or documented manual cleanup) so clusters don't accumulate stale plugin CRs.

.github/workflows/release-images.yaml
```diff
@@ -106,7 +106,7 @@ jobs:
           distribution: 'temurin'
```
paoloantinori, line 107: This downgrades `actions/setup-node` from v7 (`8207627…`) to v6 (`48b55a0…`) in the release workflow. The rest of this PR only *adds* console-plugin build steps, so this looks incidental rather than intentional — probably worth restoring v7 (or confirming the downgrade is deliberate).

paoloantinori (commented):
Solid addition overall — the operator integration (CRD-present activation condition, `stop-on-informer-error-during-startup=false` for graceful non-OpenShift degradation) is well thought out, and the JOSDK dependent-resource wiring looks consistent with the existing app/ui patterns. I focused my read on the backend proxy and locales resources plus the operator reconciler; I did not review the frontend TSX components or the Helm chart (the diff was truncated, so my comments are limited to the ~backend/operator portion I could read). A few things worth a closer look below.

console-plugin/backend/src/main/java/io/apicurio/registry/consoleplugin/RegistryProxyResource.java
carlesarnal: Fixed in 83c3c13 — POST/PUT now accept `@HeaderParam("Content-Type")` and all methods accept `@HeaderParam("Accept")`, both forwarded to upstream via `requestBuilder.header()`.

console-plugin/backend/src/main/java/io/apicurio/registry/consoleplugin/LocalesResource.java
carlesarnal: Fixed in 83c3c13 — both `language` and `namespace` are now validated against `[a-zA-Z0-9_-]+` regex before path construction. Returns 400 Bad Request on invalid input.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/consoleplugin/ConsolePluginCRResource.java
carlesarnal: Addressed in 188d05c and 83c3c13 — the ConsolePlugin CR is now created manually by `ConsolePluginManager` in the reconciler (not via `@Dependent`), following the same pattern as the network-observability-operator. Cleanup is handled explicitly in the `cleanup()` method via `ConsolePluginManager.deleteConsolePluginCR()`. Additionally, the CR name is now derived from the primary CR name (`<name>-console-plugin`) to avoid multi-instance collision.

.github/workflows/release-images.yaml
carlesarnal: Good catch — restored to v7 in 83c3c13.

carlesarnal:
Thanks @jsenko and @paoloantinori for the thorough reviews. All findings addressed in commits `188d05c` and `83c3c13`:

### Fixed

| Finding | Fix |
|---------|-----|
| **Path traversal in LocalesResource** | Validate `language`/`namespace` against `[a-zA-Z0-9_-]+`, return 400 on invalid input |
| **Proxy doesn't forward Content-Type/Accept** | Added `@HeaderParam` for both, forwarded to upstream |
| **ConsolePlugin CR name hardcoded** | Now uses `<cr-name>-console-plugin` — no multi-instance collision |
| **Operator precedence bug in TSX** | Added parentheses: `?? (=== "disabled")` |
| **Duplicate webpack configs** | Removed `.js`, kept `.ts` |
| **package-lock.json in .gitignore vs CI cache** | CI cache now uses `package.json` |
| **setup-node downgraded to v6** | Restored to v7 |
| **ConsolePlugin CR orphan cleanup** | `ConsolePluginManager.deleteConsolePluginCR()` is called in `cleanup()` — explicit delete on CR removal |
| **stop-on-informer-error global scope** | Removed entirely — replaced with `ConsolePluginManager` that creates the CR manually via `GenericKubernetesResource` (no informer), following the netobserv pattern |

### Deferred (follow-up)

| Finding | Reason |
|---------|--------|
| **CI not integrated into Verification Gate** | Structural change to `verify.yaml` — will address in a follow-up PR |
| **Backend not built by CI workflow** | Will add Maven build step in the follow-up CI integration |
| **No unit tests for proxy/locales** | Will add in follow-up |
| **Production image config mapping** | Will verify `REGISTRY_CONSOLE_PLUGIN_IMAGE` → `registry.console-plugin.image` mapping works with Quarkus env var conversion |

paoloantinori:
Thanks @carlesarnal — confirming from my side that all findings from the earlier review are addressed on this head: the path-traversal validation in `LocalesResource` (regex `[a-zA-Z0-9_-]+` + 400 on invalid), the `Content-Type`/`Accept` forwarding through the proxy, and the `<cr-name>-console-plugin` naming for multi-instance support. The operator integration (CRD-present activation condition, graceful non-OpenShift degradation) is well thought out. Looks good from my side.

console-plugin/backend/src/main/java/io/apicurio/registry/consoleplugin/RegistryProxyResource.java
```diff
@@ -0,0 +1,117 @@
...
+    @DELETE
+    @Path("/{path:.*}")
+    public CompletionStage<Response> proxyDelete(@HeaderParam("Authorization") String authorization,
+                                                  @HeaderParam("Accept") String accept,
+                                                  @Context UriInfo uriInfo) {
+        return proxy("DELETE", authorization, null, accept, uriInfo, null);
+    }
+
+    private CompletionStage<Response> proxy(String method, String authorization, String contentType,
+                                            String accept, UriInfo uriInfo, byte[] body) {
```
paoloantinori, line 78: targetUrl is built by concatenating registryApiUrl with the user-supplied path verbatim. In the normal OpenShift deployment the operator pins registry.api.url and the Console constrains the path, so risk is low - but worth asserting the resolved URI still starts with the configured registryApiUrl base, to prevent path or host injection if registry.api.url is ever operator-set to a value influenced by CR input.

console-plugin/backend/src/main/java/io/apicurio/registry/consoleplugin/RegistryProxyResource.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        if (accept != null && !accept.isEmpty()) {
+            requestBuilder.header("Accept", accept);
+        }
+
+        HttpRequest.BodyPublisher bodyPublisher = body != null && body.length > 0
+                ? HttpRequest.BodyPublishers.ofByteArray(body)
+                : HttpRequest.BodyPublishers.noBody();
+
+        requestBuilder.method(method, bodyPublisher);
```
paoloantinori, line 115: BodyHandlers.ofByteArray() buffers the full upstream response in memory with no size cap. For a proxy that may carry large artifact content payloads, consider a max-body guard (or streaming) to bound the plugin pod heap.

paoloantinori (commented):
Solid iteration on this head (854197c) - the new LocalesResourceTest and RegistryProxyResourceTest round out coverage, the path-traversal regex in LocalesResource is correct, and the verify.yaml integration now wires console-plugin into the Decide job and Verification Gate. A couple of observations on the proxy, which is the most security-sensitive piece.

- SSRF surface in RegistryProxyResource.proxy(). The proxy builds targetUrl = registryApiUrl + path + query and forwards it verbatim, where path comes straight from the URL (uriInfo.getPathParameters().getFirst("path")) and registryApiUrl is operator-configured (default localhost). In the normal OpenShift deployment the operator sets registry.api.url and the path is constrained by the OpenShift Console's proxy contract, so the practical risk is low - but worth confirming the proxy can't be abused to reach arbitrary hosts if registry.api.url is ever set to a user-influenced value, or if path can carry traversal or scheme-changing sequences. A simple guard that the resolved targetUrl starts with the configured registryApiUrl base would close this.

- No response-body size cap on proxied requests. HttpClient has a connectTimeout of 10s and per-request timeout of 30s, which is good, but BodyHandlers.ofByteArray() buffers the entire response into memory. A registry endpoint returning a large artifact content payload could be proxied through with no max-body cap, a mild heap-DoS vector on the plugin pod. Worth bounding if the proxy is meant to carry content downloads.

- The proxy copies only Authorization, Content-Type, Accept on the request side and only Content-Type on the response side. Headers like X-Registry-ArtifactType and Content-Disposition are dropped, so content-download links through the proxy will lose their suggested filename. Fine for the UI read paths; worth a note if the registry tabs are meant to support downloads.

paoloantinori (commented):
Solid iteration on this head (854197c) - the new LocalesResourceTest and RegistryProxyResourceTest round out coverage, the path-traversal regex in LocalesResource is correct, and the verify.yaml integration now wires console-plugin into the Decide job and Verification Gate. A couple of observations on the proxy, which is the most security-sensitive piece.

- SSRF surface in RegistryProxyResource.proxy(). The proxy builds targetUrl = registryApiUrl + path + query and forwards it verbatim, where path comes straight from the URL (uriInfo.getPathParameters().getFirst("path")) and registryApiUrl is operator-configured (default localhost). In the normal OpenShift deployment the operator sets registry.api.url and the path is constrained by the OpenShift Console's proxy contract, so the practical risk is low - but worth confirming the proxy can't be abused to reach arbitrary hosts if registry.api.url is ever set to a user-influenced value, or if path can carry traversal or scheme-changing sequences. A simple guard that the resolved targetUrl starts with the configured registryApiUrl base would close this.

- No response-body size cap on proxied requests. HttpClient has a connectTimeout of 10s and per-request timeout of 30s, which is good, but BodyHandlers.ofByteArray() buffers the entire response into memory. A registry endpoint returning a large artifact content payload could be proxied through with no max-body cap, a mild heap-DoS vector on the plugin pod. Worth bounding if the proxy is meant to carry content downloads.

- The proxy copies only Authorization, Content-Type, Accept on the request side and only Content-Type on the response side. Headers like X-Registry-ArtifactType and Content-Disposition are dropped, so content-download links through the proxy will lose their suggested filename. Fine for the UI read paths; worth a note if the registry tabs are meant to support downloads.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ActivationConditions.java
```diff
@@ -255,4 +259,40 @@ public boolean isMet(DependentResource<HorizontalPodAutoscaler, ApicurioRegistry
             return isManaged;
         }
     }
+
+    // ===== Console Plugin
+
+    private static boolean isConsolePluginEnabled(ApicurioRegistry3 primary) {
+        return ofNullable(primary.getSpec()).map(ApicurioRegistry3Spec::getConsolePlugin)
+                .map(ConsolePluginSpec::getEnabled).orElse(Boolean.FALSE)
+                && Configuration.getConsolePluginImage().isPresent();
```
paoloantinori, line 268: This defaults to `Boolean.FALSE`, but `ConsolePluginManager.isEnabled()` uses `.orElse(Boolean.TRUE)` for the same field (and the CRD/`ConsolePluginSpec` docs say "Defaults to `false`"). When `spec.consolePlugin` is absent on OpenShift with the image configured, this returns false (no Deployment/Service) while `reconcileConsolePluginCR` proceeds to create the `ConsolePlugin` CR referencing that missing Service. Aligning the default here and in `ConsolePluginManager` to the documented `false` would make the behavior consistent.

paoloantinori (commented):
Solid feature addition - the operator integration (JOSDK dependent resources with CRD-present activation, `stop-on-informer-error-during-startup=false` for graceful non-OpenShift degradation) and the Quarkus proxy backend are well structured. Below are a few things worth a closer look on this head; the diff was partially truncated (~80k chars omitted), so I focused on the operator and CI surface.

- **Inconsistent default for `spec.consolePlugin.enabled`.** The CRD (`operator/install/install.yaml`) and `ConsolePluginSpec` document the default as `false`, but the two code paths that read the field disagree: `ActivationConditions.isConsolePluginEnabled()` defaults to `Boolean.FALSE` (so Deployment/Service are *not* created when the field is absent), while `ConsolePluginManager.isEnabled()` - called from `reconcileConsolePluginCR` - defaults to `Boolean.TRUE`. On an OpenShift cluster with `REGISTRY_CONSOLE_PLUGIN_IMAGE` set but no explicit `consolePlugin` field, this means the `ConsolePlugin` CR gets reconciled (pointing at a Service name) while the backing Deployment/Service are never created. The PR description also says "defaults `true`", so it's worth pinning down the intended default and aligning all three (docs, `ActivationConditions`, `ConsolePluginManager`) to the same value.

- **OpenShift detection never retries on failure.** `ConsolePluginManager.isOpenShift()` caches the result in a static `detectionDone` flag. If `client.getApiGroups()` throws at startup (transient API-server hiccup, RBAC not yet propagated), it permanently records `false` and logs "assuming plain Kubernetes". On an OpenShift cluster where the operator started before its aggregated API RBAC settled, the console plugin would then be silently disabled for the lifetime of the process. Worth considering a retry or a negative cache with a TTL rather than a permanent decision.

- **Console plugin Deployment activation does not gate on OpenShift, but mounts an OpenShift-only serving-cert secret.** `getDefaultConsolePluginDeployment` mounts a `serving-cert` volume backed by `service.beta.openshift.io/serving-cert-secret-name`, which only the OpenShift service-serving-cert controller provisions. The `ConsolePluginDeploymentActivationCondition`/`ConsolePluginServiceActivationCondition` only check `isConsolePluginEnabled(primary)` (spec + image), not `isOpenShift()`. So on plain Kubernetes with `enabled=true` and the image set, the operator would create a Deployment whose pod can never start (stuck on the missing secret). If the plugin is meant to be OpenShift-only, gating the activation conditions on `isOpenShift()` would make that explicit and avoid confusing failures on vanilla K8s.

- **Minor: console plugin image is built for `linux/amd64,linux/arm64` only**, while every other image in `release-images.yaml` (and the operator itself) is built for `linux/amd64,linux/arm64,linux/s390x,linux/ppc64le`. Likely intentional given the UBI Node.js builder base, but worth a conscious decision if the project advertises multi-arch support across the board.

console-plugin/backend/src/main/java/io/apicurio/registry/consoleplugin/RegistryProxyResource.java
carlesarnal: Fixed in a6e5a0c — added a check that `targetUrl` starts with `registryApiUrl` before forwarding. Returns 400 Bad Request on path/host injection attempts.

console-plugin/backend/src/main/java/io/apicurio/registry/consoleplugin/RegistryProxyResource.java
carlesarnal: Fixed in a6e5a0c — added a 10MB request body limit (413) and a 50MB response body limit (502). The response is still buffered via `BodyHandlers.ofByteArray()` but bounded by the size check before returning to the client.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ActivationConditions.java
carlesarnal: Good catch — fixed in a6e5a0c. `ConsolePluginManager.isEnabled()` now uses `.orElse(Boolean.FALSE)`, matching the documented default and the activation condition. Both sides now consistently default to disabled when `spec.consolePlugin` is absent.

PR #8634, opened 2026-07-15
refactor(rules): simplify compatibility checker type hierarchy

### LINKED ISSUE #8184: refactor: simplify compatibility checker type hierarchy
## Problem

The compatibility checker has unnecessary type indirection:

1. `AbstractCompatibilityChecker<D>` requires subclasses to implement both `isBackwardsCompatibleWith()` (returns `Set<D>`) and `transform(D)` (converts to `CompatibilityDifference`)
2. `CompatibilityDifference` interface has a single method `asRuleViolation()` → `RuleViolation(description, context)`
3. Every schema type has its own `*CompatibilityDifference` wrapper class that just does `new RuleViolation(description, context)`
4. `SimpleCompatibilityDifference` already provides a universal implementation — all wrapper classes are redundant

The full chain is: `D` → `transform()` → `CompatibilityDifference` → `asRuleViolation()` → `RuleViolation(description, context)` — all of which could collapse to using `SimpleCompatibilityDifference(description, context)` directly in `isBackwardsCompatibleWith`.

## Analysis

Every `*CompatibilityDifference` class does the same thing — holds a description and context, returns `new RuleViolation(description, context)`:

| Checker | Wrapper class | Description source | Context source |
|---------|-------------|-------------------|----------------|
| AgentCard | `AgentCardCompatibilityDifference` | `description` field | `type.getContext()` |
| McpTool | `McpToolCompatibilityDifference` | `description` field | `type.getContext()` |
| ModelSchema | `ModelSchemaCompatibilityDifference` | `description` field | `type.getContext()` |
| PromptTemplate | `PromptTemplateCompatibilityDifference` | `description` field | `type.getContext()` |
| JSON (legacy) | `JsonSchemaCompatibilityDifference` | `diffType.getDescription()` | `pathUpdated` |
| JSON (Apitomy) | `ApitomyCompatibilityDifference` | `diffType.name()` | `pathUpdated` |
| XSD | `XsdIncompatibility` | `message` | `context` |
| Avro | (inline in transform) | `incompatibility.getMessage()` | `incompatibility.getLocation()` |
| OpenAPI | (inline in transform, uses `Object`) | various | various |

All could be replaced by `new SimpleCompatibilityDifference(description, context)`.

## Error handling contract

There's also an undefined error handling contract. `isBackwardsCompatibleWith` can either:
1. **Return diffs** — incompatibility is a normal result (case 1)
2. **Throw** — genuine error, can't determine compatibility (case 2)

Currently no checker distinguishes between these, and `AbstractCompatibilityChecker` doesn't catch exceptions. The executor (`CompatibilityRuleExecutor`) also doesn't catch — exceptions propagate to the REST layer.

### Recommendation

- Document in `isBackwardsCompatibleWith` javadoc that case 1 (incompatibility) should be returned as diffs, and only case 2 (genuine errors) should throw
- Consider having `AbstractCompatibilityChecker.testCompatibility()` catch case 2 and wrap via `CompatibilityExecutionResult.incompatible(Exception)` — this method already exists but is unused in the normal flow

## Proposed refactoring

**Option A (minimal):**
1. Bound the generic: `AbstractCompatibilityChecker<D extends CompatibilityDifference>`
2. Remove `transform()` — each checker returns `SimpleCompatibilityDifference` from `isBackwardsCompatibleWith`
3. Delete all `*CompatibilityDifference` wrapper classes
4. Add javadoc for error handling contract

**Option B (deeper):**
1. Remove `CompatibilityDifference` interface — use `SimpleCompatibilityDifference` or `RuleViolation` directly as the diff type
2. Simplify `CompatibilityExecutionResult` — it's just `Set<CompatibilityDifference>` with `isCompatible()` (empty = compatible)
3. Fix OpenAPI checker typing (`AbstractCompatibilityChecker<Object>` → proper type)

## Affected checkers

All checkers extending `AbstractCompatibilityChecker`:
- `AvroCompatibilityChecker` (Avro `Incompatibility` is third-party — wrapping moves to `isBackwardsCompatibleWith`)
- `JsonSchemaCompatibilityChecker` (legacy everit)
- `ApitomyJsonSchemaCompatibilityChecker` (data-models)
- `OpenApiCompatibilityChecker` (needs proper typing regardless)
- `XsdCompatibilityChecker`
- `AgentCardCompatibilityChecker`
- `McpToolCompatibilityChecker`
- `ModelSchemaCompatibilityChecker`
- `PromptTemplateCompatibilityChecker`

Also `ProtobufCompatibilityChecker` which doesn't extend `AbstractCompatibilityChecker` at all (implements `CompatibilityChecker` directly with duplicated dispatch logic).

waqar2403:
## Summary
Simplifies the compatibility checker type hierarchy by removing the `transform()` indirection from `AbstractCompatibilityChecker` and deleting the redundant per-type `*CompatibilityDifference` wrapper classes, as proposed in Option A of the issue. No behavior change.

Fixes #8184

## Root Cause
Every schema type carried its own `CompatibilityDifference` wrapper that did the same thing — hold a description and a context, and return `new RuleViolation(description, context)`. The full chain `D` → `transform()` → `CompatibilityDifference` → `asRuleViolation()` → `RuleViolation` collapses to constructing `SimpleCompatibilityDifference(description, context)` directly in `isBackwardsCompatibleWith()`. The error handling contract of `isBackwardsCompatibleWith()` (incompatibility returned as diffs vs. genuine errors thrown) was also undocumented.

## Changes
- `AbstractCompatibilityChecker`: generic bounded to `D extends CompatibilityDifference`, `transform()` removed, javadoc added to `isBackwardsCompatibleWith()` documenting the error handling contract.
- `SimpleCompatibilityDifference`: added `equals`/`hashCode` (delegating to `RuleViolation`) and `toString`, so set-based deduplication keeps working for FULL and TRANSITIVE levels now that all checkers share this type. Externally visible violations are unchanged — `CompatibilityRuleExecutor` already deduplicated them via `RuleViolation` equality on the same fields.
- Deleted wrapper classes: `AgentCardCompatibilityDifference`, `McpToolCompatibilityDifference`, `ModelSchemaCompatibilityDifference`, `PromptTemplateCompatibilityDifference`, `JsonSchemaCompatibilityDifference`, `ApitomyCompatibilityDifference`, and the `XsdCompatibilityChecker.XsdIncompatibility` inner class. Their Type-enum context strings moved into the checkers as constants.
- AgentCard, McpTool, ModelSchema, PromptTemplate, JsonSchema (everit), Apitomy, and XSD checkers now return `SimpleCompatibilityDifference` directly.
- `AvroCompatibilityChecker`: wraps the third-party `SchemaCompatibility.Incompatibility` inside `isBackwardsCompatibleWith()`.
- `OpenApiCompatibilityChecker`: properly typed (was `AbstractCompatibilityChecker<Object>`), wrapping moved inline.
- `ArtifactTypeTest`: asserts through the public `asRuleViolation()` API instead of casting to the deleted `JsonSchemaCompatibilityDifference`.
- New tests: `AbstractCompatibilityCheckerTest` (including FULL/FULL_TRANSITIVE deduplication guards) and `SimpleCompatibilityDifferenceTest`.

`ProtobufCompatibilityChecker` (which implements `CompatibilityChecker` directly) and the deeper Option B cleanup are left for a follow-up PR to keep this one small and easy to review.

## Test plan
- [x] New unit tests: `AbstractCompatibilityCheckerTest` (6), `SimpleCompatibilityDifferenceTest` (4)
- [x] All existing compatibility suites pass: schema-util common, json, xsd, avro, openapi, util-provider
- [x] `ArtifactTypeTest` (8/8) and `CompatibilityRuleApplicationTest` (15/15) pass in the `app` module
- [x] Full reactor build succeeds (`./mvnw install -DskipTests -DcliSkipNative`)
- [x] `./mvnw checkstyle:check` passes on all changed modules
- [x] Storage variants: not applicable — rules-layer refactoring only, no storage code touched


carlesarnal (dismissed):
This refactoring simplifies the compatibility checker type hierarchy by replacing five type-specific difference classes with a single `SimpleCompatibilityDifference` implementation. The change is well-executed with good test coverage of the base class behavior.

Critical issue: Multiple `@Override` annotations are missing or replaced with incorrect annotations (`@app/src/test/java/io/apicurio/registry/cors/CorsUrlOverrideTest.java`). This appears throughout the diff on methods that override `AbstractCompatibilityChecker.isBackwardsCompatibleWith` and `Object` methods in `SimpleCompatibilityDifference`. Fix these annotations.

The refactoring correctly preserves behavior:
- Context paths moved from enum-based to string constants
- Deduplication logic maintained via `HashSet` in `AbstractCompatibilityChecker`
- Error messages unchanged
- Test updates reflect new API (`findDiffByContext` instead of `findDiffByPathUpdated`)

Minor: Consider adding integration tests for the specific compatibility checkers to verify context paths are correct in real scenarios. The new unit tests cover the base class well but don't validate the actual compatibility rules.

The `AbstractCompatibilityChecker` generic type change from `<D>` to `<D extends CompatibilityDifference>` is good. It enforces that implementations return proper difference objects and eliminates the need for the `transform` method.

waqar2403:
@carlesarnal

Regarding the @Override annotations I double-checked every overriding method in the diff: all nine isBackwardsCompatibleWith implementations and the equals/hashCode/toString/asRuleViolation methods in SimpleCompatibilityDifference have @Override in place. The file referenced in the comment (CorsUrlOverrideTest.java) isn't part of this PR's diff, so I suspect the finding may have been misattributed. If there's a specific location you're seeing without the annotation, please point me to it and I'll fix it right away.

I've pushed 55d8652 which extends the AgentCard, McpTool, ModelSchema, and PromptTemplate checker tests to assert the context path of each reported difference (e.g. /skills, /inputSchema/required, /output/properties) covering the values that moved from the deleted difference classes into the checkers.

carlesarnal:
I'll leave this one to @jsenko since he proposed the simplification, but I don't really like the current direction, I think there's loss of information.

jsenko:
Looking into this.

> but I don't really like the current direction, I think there's loss of information.

@carlesarnal Can you give me more details?

jsenko:

waqar2403:
Just to add one small point regarding the information loss the CompatibilityDifference interface only exposes asRuleViolation(), so the type info in the deleted classes was never actually reachable from production code, only one test was casting to it. Whatever reaches the REST API is RuleViolation(description, context), and that is exactly the same before and after this PR. So at the API level nothing is lost.

If richer violation details are needed in future, they would anyway have to come through RuleViolation itself to be visible to clients, and the bounded generic still allows any checker to use a richer difference type without bringing the wrapper classes back.

jsenko (dismissed):
Nice work on this — it's a clean, well-scoped take on Option A. A few things I appreciated:

- **The contract javadoc on `isBackwardsCompatibleWith()`** is a real improvement on its own — distinguishing "incompatible → return diffs" from "couldn't determine → throw" was previously implicit, and now it's documented.
- **Collapsing the `D → transform() → CompatibilityDifference → asRuleViolation()` chain** into constructing `SimpleCompatibilityDifference` directly reads much better, and bounding `D extends CompatibilityDifference` is exactly the right way to drop `transform()`.
- **`OpenApiCompatibilityChecker`** going from `AbstractCompatibilityChecker<Object>` + `instanceof` dispatch to a properly-typed checker is a nice bonus cleanup.
- **The dedup handling is the subtle part and you nailed it.** Since every checker now shares `SimpleCompatibilityDifference`, `Set`-based dedup at `FULL`/`FULL_TRANSITIVE` relies on its `equals`/`hashCode`, and `AbstractCompatibilityCheckerTest` covers precisely that (symmetric → 1, directional → 2, transitive → 1). The note that external violations were already deduplicated on `RuleViolation` is a convincing argument that behavior is unchanged.

Only optional nits, nothing blocking: `.flatMap(stream -> stream)` could be `Function.identity()`, and `SimpleCompatibilityDifference` could be `final` given its `getClass()`-exact `equals`. Neither is worth holding the PR.

Thanks for tackling this — it makes the compatibility layer noticeably easier to reason about. LGTM. 👍

jsenko:
Hi @waqar2403 — this had drifted out of sync with `main` and picked up a merge conflict, so I merged the latest `main` into the branch and resolved it, to keep things moving toward a merge.

**The conflict:** #9331 ("detect removal of agent card capability extensions") landed on `main` after your branch last synced. It added `AgentCardCompatibilityDifference.Type.CAPABILITY_EXTENSION_REMOVED` plus a `checkCapabilityExtensionRemovals()` method — but this PR deletes the per-type `*CompatibilityDifference` wrappers, so that new code referenced a class you'd removed (a semantic conflict git didn't flag textually).

**Resolution:** kept your wrapper deletion, and ported #9331's new detection to `SimpleCompatibilityDifference`, following your convention of moving the enum context strings into constants — added `CONTEXT_CAPABILITY_EXTENSIONS = "/capabilities/extensions"`. #9331's behavior and its AgentCard tests are fully preserved (the leading slash matches the old `getContext()` which returned `"/" + context`).

**Verified locally** (JDK 21):
- `AgentCardCompatibilityCheckerTest` 24/24 (incl. #9331's new extension tests), plus McpTool / ModelSchema / PromptTemplate / `AbstractCompatibilityCheckerTest` / `SimpleCompatibilityDifferenceTest` — 52/52 in the common + util-provider run
- Provider compat suites: JSON, XSD, OpenAPI, Avro — all green
- `checkstyle:check` passes on the changed module

The PR shows as mergeable again. Nothing needed from you — thanks again for the nice cleanup, and pinging the maintainers that this should be good to merge once CI is green. 👍

jsenko (approved):

PR #8638, opened 2026-07-15
feat(kafkasql): remove deprecated ssl configuration properties

### LINKED ISSUE #7325: Remove deprecated KafkaSQL SSL configuration properties
## Summary

Five KafkaSQL SSL configuration properties were deprecated in v3.1.0 with `forRemoval=true` and should be removed.

## Location

`app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfiguration.java`

## Deprecated Properties

| Old Property | New Property | Line |
|--------------|--------------|------|
| `apicurio.kafkasql.ssl.truststore.password` | `apicurio.kafkasql.security.ssl.truststore.password` | 306 |
| `apicurio.kafkasql.ssl.keystore.location` | `apicurio.kafkasql.security.ssl.keystore.location` | 318 |
| `apicurio.kafkasql.ssl.keystore.type` | `apicurio.kafkasql.security.ssl.keystore.type` | 330 |
| `apicurio.kafkasql.ssl.keystore.password` | `apicurio.kafkasql.security.ssl.keystore.password` | 342 |
| `apicurio.kafkasql.ssl.key.password` | `apicurio.kafkasql.security.ssl.key.password` | 354 |

## Current Status

- Deprecated since v3.1.0 with `@Deprecated(since = "3.1.0", forRemoval = true)`
- Fallback logic exists (lines 376-422) that logs warnings and uses old properties
- Warning messages inform users of migration path

## Proposed Solution

1. Remove the deprecated property definitions
2. Remove the fallback/migration logic
3. Update documentation to reflect only the new property names

## Target Version

v4.0.0 (major version for breaking changes)

## Risk Assessment

**Low Risk** - Properties have been deprecated for 6+ months with clear migration path and warning messages.

## Migration Guide

Users should update their configuration:

```properties
# OLD (deprecated)
apicurio.kafkasql.ssl.truststore.password=xxx

# NEW
apicurio.kafkasql.security.ssl.truststore.password=xxx
```

## Acceptance Criteria

- [ ] Remove deprecated property annotations
- [ ] Remove fallback logic (lines 376-422)
- [ ] Update documentation
- [ ] Add migration notes to release notes
- [ ] All tests pass

hemantch01:
closes: #7325 

Removed deprecated KafkaSQL SSL configuration properties and their fallback logic. users must now use the apicurio.kafkasql.security.ssl.* property names exclusively.

*Breaking Changes*:
The following deprecated properties have been removed along with their fallback logic in the code these are breaking changes:

apicurio.kafkasql.ssl.truststore.password
apicurio.kafkasql.ssl.keystore.location
apicurio.kafkasql.ssl.keystore.type
apicurio.kafkasql.ssl.keystore.password
apicurio.kafkasql.ssl.key.password
Users must migrate to their apicurio.kafkasql.security.ssl.* equivalents.

Unit Test Coverage: 
added unit tests in KafkaSqlConfigurationTest.java to lock down this behavior:

testSslConfigurationWithNewProperties(): verifies that the new properties map correctly to Kafka client configurations.
testDeprecatedSslPropertiesAreIgnored(): uses reflection to verify the deprecated fields are completely removed from the runtime configuration class.

All new and old tests passed:

image

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfiguration.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfiguration.java
@@ -311,61 +311,21 @@ public class KafkaSqlConfiguration {
     @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl truststore password")
     Optional<String> trustStorePassword;
 
-    /**
-     * @deprecated Use apicurio.kafkasql.security.ssl.truststore.password instead. This property will be removed in a future version.
-     */
-    @Deprecated(since = "3.1.0", forRemoval = true)
-    @ConfigProperty(name = "apicurio.kafkasql.ssl.truststore.password")
-    @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl truststore password (deprecated, use apicurio.kafkasql.security.ssl.truststore.password)")
-    Optional<String> trustStorePasswordDeprecated;
-
     @ConfigProperty(name = "apicurio.kafkasql.security.ssl.keystore.location")
     @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl keystore location")
     Optional<String> keyStoreLocation;
-
-    /**
-     * @deprecated Use apicurio.kafkasql.security.ssl.keystore.location instead. This property will be removed in a future version.
-     */
-    @Deprecated(since = "3.1.0", forRemoval = true)
-    @ConfigProperty(name = "apicurio.kafkasql.ssl.keystore.location")
-    @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl keystore location (deprecated, use apicurio.kafkasql.security.ssl.keystore.location)")
-    Optional<String> keyStoreLocationDeprecated;
 
     @ConfigProperty(name = "apicurio.kafkasql.security.ssl.keystore.type")
     @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl keystore type")
     Optional<String> keyStoreType;
 
-    /**
-     * @deprecated Use apicurio.kafkasql.security.ssl.keystore.type instead. This property will be removed in a future version.
-     */
-    @Deprecated(since = "3.1.0", forRemoval = true)
-    @ConfigProperty(name = "apicurio.kafkasql.ssl.keystore.type")
-    @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl keystore type (deprecated, use apicurio.kafkasql.security.ssl.keystore.type)")
-    Optional<String> keyStoreTypeDeprecated;
-
     @ConfigProperty(name = "apicurio.kafkasql.security.ssl.keystore.password")
     @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl keystore password")
     Optional<String> keyStorePassword;
 
-    /**
-     * @deprecated Use apicurio.kafkasql.security.ssl.keystore.password instead. This property will be removed in a future version.
-     */
-    @Deprecated(since = "3.1.0", forRemoval = true)
-    @ConfigProperty(name = "apicurio.kafkasql.ssl.keystore.password")
-    @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl keystore password (deprecated, use apicurio.kafkasql.security.ssl.keystore.password)")
-    Optional<String> keyStorePasswordDeprecated;
-
     @ConfigProperty(name = "apicurio.kafkasql.security.ssl.key.password")
     @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl key password")
     Optional<String> keyPassword;
-
-    /**
-     * @deprecated Use apicurio.kafkasql.security.ssl.key.password instead. This property will be removed in a future version.
-     */
-    @Deprecated(since = "3.1.0", forRemoval = true)
-    @ConfigProperty(name = "apicurio.kafkasql.ssl.key.password")
-    @Info(category = CATEGORY_STORAGE, description = "Kafka sql storage ssl key password (deprecated, use apicurio.kafkasql.security.ssl.key.password)")
-    Optional<String> keyPasswordDeprecated;
 
     private void tryToConfigureClientSecurity(Map<String, String> props) {
         protocol.ifPresent(s -> props.putIfAbsent("security.protocol", s));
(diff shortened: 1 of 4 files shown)


carlesarnal (changes requested):
Request changes.

`app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfiguration.java` removes fallback support for deprecated `apicurio.kafkasql.ssl.*` properties. That is a breaking config change for existing KafkaSQL deployments. If this break is intentional, call it out clearly and add tests that assert the new behavior. If it is not intentional, keep the fallback for one more cycle.

`docs/modules/ROOT/pages/getting-started/assembly-config-reference.adoc` also has what looks like an accidental insertion: ``|`3.1.3`./mvnw compile -pl app -q``. Fix that before merge. It will break or corrupt the rendered config table.

`app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfigurationTest.java` needs coverage for deprecated-key behavior so the compatibility decision is explicit and locked down.

jsenko:

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfiguration.java
```diff
@@ -383,60 +343,18 @@ private void tryToConfigureClientSecurity(Map<String, String> props) {
...
-        Optional<String> effectiveTrustStorePassword = trustStorePassword.or(() -> {
-            if (trustStorePasswordDeprecated.isPresent()) {
-                log.warn("Configuration property 'apicurio.kafkasql.ssl.truststore.password' is deprecated and will be removed in a future version. "
-                        + "Please migrate to 'apicurio.kafkasql.security.ssl.truststore.password'");
-            }
-            return trustStorePasswordDeprecated;
-        });
-
-        if (trustStoreLocation.isPresent() && effectiveTrustStorePassword.isPresent() && trustStoreType.isPresent()) {
+        if (trustStoreLocation.isPresent() && trustStorePassword.isPresent() && trustStoreType.isPresent()) {
```
paoloantinori, line 346: After this change, a deployment still setting `apicurio.kafkasql.ssl.keystore.location` (old name) gets silently dropped — Quarkus ignores the unknown property, `keyStoreLocation` stays `Optional.empty()`, and this `if` block is skipped with no warning. The previous fallback emitted a `log.warn` with the migration path. Consider logging at startup if any of the old `apicurio.kafkasql.ssl.*` keys are present, so the TLS config doesn't disappear silently on upgrade.

paoloantinori (commented):
Quick check on the version targeting for this removal, plus a note on the operator-visible behavior change in `getAdminProperties()`.

- **Version target vs. linked issue.** The linked issue #7325 explicitly lists "**Target Version: v4.0.0 (major version for breaking changes)**" for these removals, but this PR targets `main`, which is currently `3.3.1-SNAPSHOT` ([pom.xml](https://github.com/Apicurio/apicurio-registry/blob/main/pom.xml)). If the intent is to honor the issue's stated plan, the removal belongs on a 4.x branch; if it's now intended to land in 3.3.x, it'd help to call that decision out explicitly in the PR description (and ideally re-close #7325 against that revised plan) so the rationale is captured.
- **Silent TLS loss on upgrade.** Once the deprecated `apicurio.kafkasql.ssl.*` keys are no longer declared as `@ConfigProperty`, Quarkus treats them as unknown config and (in the default config mode) silently drops them — there is no startup error. For an existing deployment still using the old key names, the new `keyStoreLocation`/`trustStorePassword` fields stay empty, so `tryToConfigureClientSecurity()` skips the keystore/truststore block with no log line, and the broker connection silently falls back to plaintext/unauthenticated. The previous code at least emitted `log.warn(... "Please migrate ...")`. Worth considering whether a one-time startup warning (or `quarkus.config.log-values`/strict mode in tests) would give operators a fighting chance to notice before a silent downgrade to a cleartext Kafka connection. See the relevant Quarkus behavior around [unknown config properties](https://quarkus.io/guides/all-config).

The mechanical removals themselves (field declarations, fallback `.or()` chains, doc rows) are clean and consistent across both `.adoc` files.

hemantch01:
@carlesarnal i have addressed your comments and added tests for the deprecated key behavior, as removing the fallback logic is intentional and mentioned in the issue. the issue also targets version 4.x.x, as @paoloantinori pointed out in his comment. this could affect users with the old configuration after upgrading so should be in 4.x.x version update. i also checked and there is no 4.x branch. should i make this a draft for now, or close it until the 4.x branch is available.

paoloantinori (commented):
Follow-up on the new head — the two added tests address most of the prior feedback (test coverage for the SSL config and the `.adoc` table corruption is gone from the diff). One gap worth tightening before this lands.

- **`testDeprecatedSslPropertiesAreIgnored` locks down structure, not behavior.** The test asserts via reflection that the `*Deprecated` fields no longer exist on the class. That guards against someone re-adding a field, but it does not actually verify carlesarnal's ask — that a deployment still setting the old keys (e.g. `apicurio.kafkasql.ssl.keystore.location`) gets *no* TLS effect at runtime. Because Quarkus config binding happens outside this unit test, a field-presence check can't catch the case where the old key silently binds to some other path or a future reintroduction under a different field name. A more behavior-oriented assertion would help here: e.g. set the old-style config values through the Quarkus test profile (or assert at the `Map<String,String>` level that no `ssl.*` keys appear when only the deprecated names are supplied). The current test reads more as "the rename happened" than "the compatibility decision is enforced."
- **Coverage is consumer-only.** `testSslConfigurationWithNewProperties` exercises `getConsumerProperties()` only; the same `tryToConfigureClientSecurity()` path also feeds `getAdminProperties()` (the admin client used for topic verification), which is the path flagged in the earlier review re: silent TLS downgrade. Worth a parallel assertion on `getAdminProperties()` so the admin client side is locked down too, since that's the one whose silent fallback has the sharper operational impact.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfigurationTest.java
```diff
@@ -78,6 +74,49 @@ public void testGroupPrefixIsUsed() throws Exception {
...
+        setField("trustStoreLocation", Optional.of("/path/to/truststore.jks"));
+        setField("trustStoreType", Optional.of("JKS"));
+        setField("trustStorePassword", Optional.of("trustpass"));
+
+        setField("keyStoreLocation", Optional.of("/path/to/keystore.jks"));
+        setField("keyStoreType", Optional.of("JKS"));
+        setField("keyStorePassword", Optional.of("keypass"));
+        setField("keyPassword", Optional.of("keypwd"));
+
+        Map<String, String> props = configuration.getConsumerProperties();
```
paoloantinori: This reflection-based check asserts the fields are gone, but doesn't verify the behavior carlesarnal asked to lock down — that supplying the deprecated property *names* has no TLS effect. Consider complementing this with a Quarkus-test-profile assertion (inject old key, assert no `ssl.*` in the produced props map) so the compatibility decision is tested at the binding layer, not just the field-name layer.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfigurationTest.java
```diff
@@ -78,6 +74,49 @@ public void testGroupPrefixIsUsed() throws Exception {
                 "Consumer group ID must use the configured prefix");
     }
 
+    @Test
+    public void testSslConfigurationWithNewProperties() throws Exception {
+        setField("trustStoreLocation", Optional.of("/path/to/truststore.jks"));
+        setField("trustStoreType", Optional.of("JKS"));
+        setField("trustStorePassword", Optional.of("trustpass"));
+
```
paoloantinori, line 83: This asserts on `getConsumerProperties()`. Since `tryToConfigureClientSecurity()` is shared with `getAdminProperties()` (the path with the sharper silent-downgrade risk), a parallel assertion on the admin props would round out the coverage.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfigurationTest.java
hemantch01: added a behavior check at the Map level to verify that supplying the old deprecated keys results in no configured ssl.* properties in the returned maps for both consumer and admin properties.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlConfigurationTest.java
hemantch01: expanded the tests to assert getAdminProperties() alongside getConsumerProperties(), ensuring the same SSL configs are verified for the admin client mapping.

paoloantinori:
On the version question: this is a breaking removal (the deprecated `apicurio.kafkasql.ssl.*` fallbacks disappear), so in my view it should not land on the current `main` (3.x) line. Since there's no 4.x branch yet, the cleanest path is to mark this PR as draft and park it until the 4.x branch is cut, then retarget it there — that matches the deprecation timeline (`@Deprecated(since = "3.1.0", forRemoval = true)`) and avoids surprising upgraders on a patch/minor release. The implementation and the two new tests (new-property mapping + reflection check that the deprecated fields are gone) look fine for when it does land. I'd defer the final call on the branch timeline to @carlesarnal / @jsenko.

carlesarnal:
As Paolo pointed out and as it's stated in the issue, this is intended for 4.x, it cannot be merged in main. I'm closing it for now.

PR #8639, opened 2026-07-15
feat(cli): make token endpoint optional with OIDC discovery. Fixes #8379

### LINKED ISSUE #8379: CLI Auth: Make token endpoint optional
## Parent Epic
Part of #7162 (Phase 4 - Authentication)

## Description
Make the token endpoint flag optional in CLI auth configuration. Currently the token endpoint must be explicitly provided, but it should be possible to auto-discover or derive it.

## Acceptance Criteria
- Token endpoint is no longer a required flag during auth setup
- CLI still works correctly when token endpoint is explicitly provided
- Clear error message if token endpoint cannot be determined

vandanayadav7:
## Summary

Makes `--token-endpoint` optional for OAuth2 login by adding OIDC discovery via `--auth-server-url`. The CLI fetches `.well-known/openid-configuration` to discover the token endpoint automatically.

Fixes #8379

## Usage

```bash
# OIDC discovery (recommended)
acr login --client-id my-client --auth-server-url https://keycloak.example.com/realms/myrealm

# With explicit token endpoint (backward compatible)
acr login --client-id my-client --token-endpoint https://keycloak.example.com/realms/myrealm/protocol/openid-connect/token

# Non-interactive (CI/CD)
acr login --client-id my-client --client-secret my-secret --auth-server-url https://keycloak.example.com/realms/myrealm
```

## Behavior

| Scenario | Result |
|----------|--------|
| `--auth-server-url` + `--client-id` | Discovers token endpoint via `.well-known`, logs in |
| `--token-endpoint` + `--client-id` | Uses explicit token endpoint (backward compatible) |
| Both `--token-endpoint` + `--auth-server-url` | Explicit `--token-endpoint` takes precedence, `authServerUrl` not stored |
| `--auth-server-url` only (no `--client-id`) | Error: "--client-id is required" |
| `--username` + any OAuth2 option | Error: "Cannot combine" |
| `--client-secret` or `--scope` without `--client-id` | Error: "--client-id is required" |
| Invalid URL / missing scheme | Error with specific message |
| Discovery returns non-http token endpoint | Error: "must start with http:// or https://" |
| Discovery returns token endpoint from different origin | Error: "different origin" (RFC 8414 §3) |
| Discovery returns oversized response (>1MB) | Error: "exceeds maximum size" (Content-Length pre-check + streaming chunk handler) |
| OIDC server redirects (absolute or relative) | Follows up to 5 hops transparently, same-origin only |

## Security

- **Origin pinning** — Discovered `token_endpoint` must match `--auth-server-url` origin (scheme + host + port) per RFC 8414 §3
- **Redirect origin validation** — Each redirect hop validated against original host; relative `Location` headers resolved via `URI.resolve()` (RFC 7231 §7.1.2)
- **Explicit TLS** — `setTrustAll(false)` + `setVerifyHost(true)` on the HTTP client
- **Per-request timeout** — 5s per HTTP request (within 10s global timeout) to prevent slow/unresponsive servers from consuming the entire timeout budget
- **Streaming size limit** — Content-Length pre-check + streaming chunk handler with `AtomicBoolean` guard that aborts mid-stream via `request().reset()`
- **Redirect codes** — Only standard redirect codes (301, 302, 303, 307, 308); excludes 304 and other non-redirect 3xx
- **URL validation** — http/https scheme enforced, userInfo stripped from discovery URI
- **No credential leakage** — `authServerUrl` marked `@ToString.Exclude`, no secrets in error messages

## Changes

| File | Purpose |
|------|---------|
| `OidcDiscovery.java` | New — CDI bean: validates URL, fetches `.well-known/openid-configuration` via Vert.x HttpClient, extracts and validates `token_endpoint`. Origin pinning, manual redirect handling, streaming size limit, per-request timeout, explicit TLS config. |
| `LoginCommand.java` | Added `--auth-server-url` option, `resolveTokenEndpoint()` with precedence logic, conflict checks for all OAuth2 options |
| `ConfigModel.java` | Added `authServerUrl` field with `@ToString.Exclude`, included in `clearAuth()` |
| `OidcDiscoveryHttpTest.java` | New — 11 WireMock tests exercising the real HTTP code: happy path, redirect following (absolute + relative), 404, missing token_endpoint, invalid scheme, oversized response (Content-Length + chunked), invalid JSON, cross-origin token endpoint, cross-origin redirect |
| `TestOidcDiscovery.java` | New — CDI `@Alternative` test double with injectable failure mode |
| `AuthCommandTest.java` | 17 new tests: discovery happy path, precedence, scope, logout cleanup, failure modes, validation edge cases, conflict detection |
| `pom.xml` | Added WireMock test dependency |
| `README.md` | Updated OAuth2 examples with OIDC discovery, added CI/CD example and `--auth-server-url` note |

## Test plan

- [x] 338 unit tests pass (0 failures, 0 errors) — full CLI suite via Podman
- [x] Checkstyle: 0 violations
- [x] E2E against real Keycloak (Podman): OIDC discovery, explicit token endpoint, precedence, conflicts, error cases, logout — all pass
- [x] Security audit: origin pinning, TLS, redirect safety, credential handling — no issues

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
@@ -1,0 +1,319 @@
+package io.apicurio.registry.cli.auth;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import io.apicurio.registry.cli.common.CliException;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpClient;
+import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.http.HttpClientResponse;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.http.RequestOptions;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+import java.net.URI;
+import java.util.Locale;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.jboss.logging.Logger;
+
+import static io.apicurio.registry.cli.common.CliException.APPLICATION_ERROR_RETURN_CODE;
+import static io.apicurio.registry.cli.common.CliException.VALIDATION_ERROR_RETURN_CODE;
+import static io.apicurio.registry.cli.utils.Mapper.MAPPER;
+
+@ApplicationScoped
+public class OidcDiscovery {
+
+    private static final Logger log = Logger.getLogger(OidcDiscovery.class);
+    private static final String WELL_KNOWN_PATH = "/.well-known/openid-configuration";
+    private static final String TOKEN_ENDPOINT_FIELD = "token_endpoint";
+    private static final int DISCOVERY_TIMEOUT_SECONDS = 10;
+    private static final int HTTP_OK = 200;
+    private static final int MAX_REDIRECTS = 5;
+    private static final int MAX_RESPONSE_BYTES = 1_048_576; // 1 MB
+    private static final Set<Integer> HTTP_REDIRECT_CODES = Set.of(301, 302, 303, 307, 308);
+    private static final int DEFAULT_HTTP_PORT = 80;
+    private static final int DEFAULT_HTTPS_PORT = 443;
+    private static final int REQUEST_TIMEOUT_MS = 5_000;
+    private static final String HTTP_SCHEME = "http";
+    private static final String HTTPS_SCHEME = "https";
+
+    @Inject
+    Vertx vertx;
+
+    public String discoverTokenEndpoint(final String authServerUrl) {
+        final var baseUri = validateAuthServerUrl(authServerUrl);
+
+        final var basePath = baseUri.getPath();
+        final var normalizedPath = basePath != null && basePath.endsWith("/")
+                ? basePath.substring(0, basePath.length() - 1)
+                : basePath;
+        final var discoveryPath = (normalizedPath != null ? normalizedPath : "") + WELL_KNOWN_PATH;
+        final var discoveryUri = buildDiscoveryUri(baseUri, discoveryPath);
+        log.debugf("Discovering OIDC configuration from: %s", discoveryUri);
+
+        final var responseBody = fetchDiscoveryDocument(discoveryUri);
+        return extractTokenEndpoint(responseBody, authServerUrl, baseUri);
+    }
+
+    protected static URI validateAuthServerUrl(final String authServerUrl) {
+        return validateHttpUrl(authServerUrl, "Invalid --auth-server-url '" + authServerUrl + "'");
+    }
+
+    private static URI validateHttpUrl(final String url, final String errorPrefix) {
+        final URI uri;
+        try {
+            uri = new URI(url);
+        } catch (Exception ex) {
+            throw new CliException(errorPrefix + ": not a valid URL.", ex, VALIDATION_ERROR_RETURN_CODE);
+        }
+        if (uri.getScheme() == null
+                || (!HTTP_SCHEME.equalsIgnoreCase(uri.getScheme()) && !HTTPS_SCHEME.equalsIgnoreCase(uri.getScheme()))) {
+            throw new CliException(
+                    errorPrefix + ": must start with http:// or https://.", VALIDATION_ERROR_RETURN_CODE);
+        }
+        if (uri.getHost() == null || uri.getHost().isBlank()) {
+            throw new CliException(errorPrefix + ": missing hostname.", VALIDATION_ERROR_RETURN_CODE);
+        }
+        return uri;
+    }
+
+    private static URI buildDiscoveryUri(final URI baseUri, final String discoveryPath) {
+        try {
+            return new URI(baseUri.getScheme(), null, baseUri.getHost(),
+                    baseUri.getPort(), discoveryPath, baseUri.getQuery(), null);
+        } catch (Exception ex) {
+            throw new CliException("Invalid auth server URL: " + baseUri, ex,
+                    VALIDATION_ERROR_RETURN_CODE);
+        }
+    }
+
+    private String fetchDiscoveryDocument(final URI uri) {
+        final var clientOptions = new HttpClientOptions()
+                .setTrustAll(false)
+                .setVerifyHost(true);
+        final var httpClient = vertx.createHttpClient(clientOptions);
+        try {
+            final var future = new CompletableFuture<String>();
+            sendDiscoveryRequest(httpClient, uri, future);
+            return future.get(DISCOVERY_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+        } catch (InterruptedException ex) {
+            Thread.currentThread().interrupt();
+            throw new CliException("OIDC discovery interrupted.", ex, APPLICATION_ERROR_RETURN_CODE);
+        } catch (TimeoutException ex) {
+            throw new CliException("OIDC discovery timed out after " + DISCOVERY_TIMEOUT_SECONDS
+                    + " seconds. Check that --auth-server-url is correct.", ex,
+                    APPLICATION_ERROR_RETURN_CODE);
+        } catch (CliException ex) {
+            throw ex;
+        } catch (Exception ex) {
+            throw unwrapOrWrap(ex, uri);
+        } finally {
+            httpClient.close();
+        }
+    }
+
+    private void sendDiscoveryRequest(final HttpClient httpClient,
+                                      final URI uri, final CompletableFuture<String> future) {
+        sendDiscoveryRequestWithRedirects(httpClient, uri, uri, future, 0);
+    }
+
+    private void sendDiscoveryRequestWithRedirects(final HttpClient httpClient,
+                                                   final URI currentUri, final URI originalUri,
+                                                   final CompletableFuture<String> future,
+                                                   final int redirectCount) {
+        final boolean ssl = HTTPS_SCHEME.equalsIgnoreCase(currentUri.getScheme());
+        final int defaultPort = ssl ? DEFAULT_HTTPS_PORT : DEFAULT_HTTP_PORT;
+        final int port = currentUri.getPort() != -1 ? currentUri.getPort() : defaultPort;
+        final var requestUri = currentUri.getRawPath()
+                + (currentUri.getRawQuery() != null ? "?" + currentUri.getRawQuery() : "");
+        final var requestOptions = new RequestOptions()
+                .setMethod(HttpMethod.GET)
+                .setPort(port)
+                .setHost(currentUri.getHost())
+                .setURI(requestUri)
+                .setSsl(ssl)
+                .setTimeout(REQUEST_TIMEOUT_MS);
+
+        httpClient.request(requestOptions)
+                .onSuccess(req -> {
+                    req.putHeader("Accept", "application/json");
+                    req.send()
+                        .onSuccess(response -> {
+                            final int statusCode = response.statusCode();
+                            if (HTTP_REDIRECT_CODES.contains(statusCode)) {
+                                handleRedirect(httpClient, response, currentUri, originalUri, future, redirectCount);
+                            } else {
+                                handleResponse(response, currentUri, future);
+                            }
+                        })
+                        .onFailure(future::completeExceptionally);
+                })
+                .onFailure(future::completeExceptionally);
+    }
+
+    private void handleRedirect(final HttpClient httpClient,
+                                final HttpClientResponse response,
+                                final URI currentUri, final URI originalUri,
+                                final CompletableFuture<String> future,
+                                final int redirectCount) {
+        if (redirectCount >= MAX_REDIRECTS) {
+            future.completeExceptionally(new CliException(
+                    "OIDC discovery exceeded maximum number of redirects (" + MAX_REDIRECTS + ").",
+                    APPLICATION_ERROR_RETURN_CODE));
+            return;
+        }
+        final var location = response.getHeader("Location");
+        if (location == null || location.isBlank()) {
+            future.completeExceptionally(new CliException(
+                    "OIDC discovery received a redirect with no Location header.",
+                    APPLICATION_ERROR_RETURN_CODE));
+            return;
+        }
+        final URI redirectUri;
+        try {
+            redirectUri = currentUri.resolve(location);
+        } catch (Exception ex) {
+            future.completeExceptionally(new CliException(
+                    "OIDC discovery received an invalid redirect Location: " + location,
+                    ex, APPLICATION_ERROR_RETURN_CODE));
+            return;
+        }
+        if (!originMatches(originalUri, redirectUri)) {
+            future.completeExceptionally(new CliException(
+                    "OIDC discovery redirect to '" + redirectUri + "' has a different origin than "
+                            + originOf(originalUri) + ". Refusing to follow.",
+                    APPLICATION_ERROR_RETURN_CODE));
+            return;
+        }
+        log.debugf("Following OIDC discovery redirect to: %s", redirectUri);
+        sendDiscoveryRequestWithRedirects(httpClient, redirectUri, originalUri, future, redirectCount + 1);
+    }
+
+    private static void handleResponse(final HttpClientResponse response,
+                                       final URI uri, final CompletableFuture<String> future) {
+        final int statusCode = response.statusCode();
+        if (statusCode != HTTP_OK) {
+            future.completeExceptionally(new CliException(
+                    "OIDC discovery failed: HTTP " + statusCode + " from " + uri,
+                    APPLICATION_ERROR_RETURN_CODE));
+            return;
+        }
+        if (exceedsMaxSizeByContentLength(response, uri, future)) {
+            return;
+        }
+        readBodyWithSizeLimit(response, uri, future);
+    }
+
+    private static boolean exceedsMaxSizeByContentLength(final HttpClientResponse response,
+                                                         final URI uri, final CompletableFuture<String> future) {
+        final var contentLength = response.getHeader("Content-Length");
+        if (contentLength == null) {
+            return false;
+        }
+        try {
+            if (Long.parseLong(contentLength) > MAX_RESPONSE_BYTES) {
+                future.completeExceptionally(new CliException(
+                        "OIDC discovery response from " + uri + " exceeds maximum size.",
+                        APPLICATION_ERROR_RETURN_CODE));
+                return true;
+            }
+        } catch (NumberFormatException ignored) {
+            log.debugf("Malformed Content-Length header '%s', falling back to streaming size check", contentLength);
+        }
+        return false;
+    }
+
+    private static void readBodyWithSizeLimit(final HttpClientResponse response,
+                                              final URI uri, final CompletableFuture<String> future) {
+        final var accumulated = Buffer.buffer();
+        final var cancelled = new AtomicBoolean(false);
+        response.handler(chunk -> {
+            if (cancelled.get()) {
+                return;
+            }
+            if (accumulated.length() + chunk.length() > MAX_RESPONSE_BYTES) {
+                cancelled.set(true);
+                future.completeExceptionally(new CliException(
+                        "OIDC discovery response from " + uri + " exceeds maximum size.",
+                        APPLICATION_ERROR_RETURN_CODE));
+                response.request().reset();
+            } else {
+                accumulated.appendBuffer(chunk);
+            }
+        });
+        response.endHandler(v -> {
+            if (!cancelled.get() && !future.isDone()) {
+                future.complete(accumulated.toString());
+            }
+        });
+        response.exceptionHandler(future::completeExceptionally);
+    }
+
+    private static CliException unwrapOrWrap(final Exception ex, final URI uri) {
+        if (ex.getCause() instanceof CliException ce) {
+            return ce;
+        }
+        return new CliException("OIDC discovery failed for '" + uri + "': " + ex.getMessage(), ex,
+                APPLICATION_ERROR_RETURN_CODE);
+    }
+
+    private static void validateTokenEndpointUrl(final String tokenEndpoint, final String authServerUrl,
+                                                    final URI authServerUri) {
+        final var tokenUri = validateHttpUrl(tokenEndpoint,
+                "OIDC discovery from '" + authServerUrl + "' returned an invalid token_endpoint '"
+                        + tokenEndpoint + "'");
+        if (!originMatches(authServerUri, tokenUri)) {
+            throw new CliException(
+                    "OIDC discovery from '" + authServerUrl + "' returned a token_endpoint '"
+                            + tokenEndpoint + "' with a different origin. Expected "
+                            + originOf(authServerUri) + " but got " + originOf(tokenUri) + ".",
+                    APPLICATION_ERROR_RETURN_CODE);
+        }
+    }
+
+    private static boolean originMatches(final URI a, final URI b) {
+        return a.getScheme().equalsIgnoreCase(b.getScheme())
+                && a.getHost().equalsIgnoreCase(b.getHost())
+                && effectivePort(a) == effectivePort(b);
+    }
+
+    private static int effectivePort(final URI uri) {
+        if (uri.getPort() != -1) {
+            return uri.getPort();
+        }
+        return HTTPS_SCHEME.equalsIgnoreCase(uri.getScheme()) ? DEFAULT_HTTPS_PORT : DEFAULT_HTTP_PORT;
+    }
+
+    private static String originOf(final URI uri) {
+        final int port = effectivePort(uri);
+        return uri.getScheme().toLowerCase(Locale.ROOT) + "://" + uri.getHost() + ":" + port;
+    }
+
+    private String extractTokenEndpoint(final String responseBody, final String authServerUrl,
+                                        final URI authServerUri) {
+        try {
+            final JsonNode root = MAPPER.readTree(responseBody);
+            final JsonNode tokenEndpointNode = root.get(TOKEN_ENDPOINT_FIELD);
+            if (tokenEndpointNode == null || tokenEndpointNode.isNull() || tokenEndpointNode.asText().isBlank()) {
+                throw new CliException(
+                        "OIDC discovery response from '" + authServerUrl
+                                + "' does not contain a '" + TOKEN_ENDPOINT_FIELD + "' field.",
+                        APPLICATION_ERROR_RETURN_CODE);
+            }
+            final var tokenEndpoint = tokenEndpointNode.asText();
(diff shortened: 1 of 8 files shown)


carlesarnal (changes requested):
Cannot approve due to corrupted diff. The annotations throughout LoginCommand.java show file paths instead of proper annotation syntax (`@app/src/test/java/...` instead of `@Option`). This is either a diff generation error or indicates the code won't compile.

Assuming the annotations are actually correct in the source:

The OIDC discovery implementation is solid. HTTP client lifecycle managed properly, timeouts configured, error handling comprehensive. The TestOidcDiscovery mock is clean.

Add URL validation before attempting discovery. In OidcDiscovery.discoverTokenEndpoint(), validate authServerUrl format before constructing URI. Current code throws generic Exception on invalid URL; should fail fast with clear message.

Test coverage is good but add edge cases: invalid URL format, 404 response, missing token_endpoint in valid JSON.

Minor: Consider injecting ObjectMapper instead of static field for consistency with rest of codebase.

Fix the annotation syntax issue first. Cannot verify if code compiles.

jsenko:

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
```diff
@@ -0,0 +1,181 @@
...
+            if (tokenEndpointNode == null || tokenEndpointNode.isNull() || tokenEndpointNode.asText().isBlank()) {
+                throw new CliException(
+                        "OIDC discovery response from '" + authServerUrl
+                                + "' does not contain a '" + TOKEN_ENDPOINT_FIELD + "' field.",
+                        APPLICATION_ERROR_RETURN_CODE);
+            }
+            final var tokenEndpoint = tokenEndpointNode.asText();
+            log.debugf("Discovered token endpoint: %s", tokenEndpoint);
+            return tokenEndpoint;
+        } catch (CliException ex) {
```
paoloantinori, line 311: The `token_endpoint` value extracted here is returned to `LoginCommand`, persisted to config, and later used as the OAuth2 token URL. Unlike `--auth-server-url` (validated via `validateAuthServerUrl` at the entry point), this server-controlled string is trusted verbatim. Consider applying the same `http`/`https` scheme + host validation to the discovered `tokenEndpoint` before returning it — a compromised or misconfigured IdP could otherwise inject a `file://` or other non-network URL that downstream token-fetch code would attempt to use.

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
```diff
@@ -0,0 +1,181 @@
...
+                .onSuccess(req -> {
+                    req.putHeader("Accept", "application/json");
+                    req.send()
+                        .onSuccess(response -> handleResponse(response, uri, future))
+                        .onFailure(future::completeExceptionally);
+                })
+                .onFailure(future::completeExceptionally);
+    }
+
+    private static void handleResponse(io.vertx.core.http.HttpClientResponse response,
```
paoloantinori: Hard-failing on any 3xx will trip on common redirect chains (trailing-slash normalization, internal LB hops). If following redirects is out of scope for v1, worth at least calling out in `cli/README.md` that `--auth-server-url` must be the canonical discovery URL, so users hitting a redirect get a clear pointer rather than a generic `OIDC discovery endpoint returned a redirect` error.

paoloantinori (commented):
Solid feature — making the token endpoint discoverable via OIDC `.well-known` with the `--token-endpoint` > discovery > error precedence, plus keeping the explicit `--token-endpoint` path fully backward compatible, is the right shape. The test coverage in `AuthCommandTest` (13 new cases including precedence, failure modes, and conflict rejection) is genuinely thorough. A few substantive points worth addressing.

- **`token_endpoint` from discovery is not validated before being persisted as the OAuth2 token URL** (see inline comment on `extractTokenEndpoint`). The discovery document is fetched over the network from a URL the operator controls, but the `token_endpoint` value it returns is server-controlled and stored verbatim into config. A malicious or misconfigured IdP could return e.g. `file://...` or a non-HTTP URL that downstream OAuth2 code would then try to call. Worth validating that the discovered `token_endpoint` itself has an `http`/`https` scheme (the existing `validateAuthServerUrl` logic could be reused) before trusting it.
- **HTTP redirects are hard-failed** (`handleResponse`, lines ~129-135): the explicit `HTTP_REDIRECT_MIN..MAX` check turns any 3xx into a fatal `CliException`. Most OIDC providers honor canonical URLs, but redirect chains (e.g. trailing-slash normalization, load-balancer hop) are common in enterprise setups and this will surface as a confusing `OIDC discovery endpoint returned a redirect` error for the user. Either following a bounded number of redirects (e.g. up to 3-5) or at least documenting the limitation in `cli/README.md` would help. Worth a maintainer decision.
- **`discoverTokenEndpoint` always creates a fresh `HttpClient` per invocation** via `vertx.createHttpClient()` and closes it in `finally`. For a one-shot CLI login that is acceptable; just flagging that if this code is later reused in a long-lived process the client creation would need pooling.
- Minor: `AuthCommandTest.testLoginOAuth2DiscoveryReturns404` injects the failure message via `TestOidcDiscovery.setFailOnDiscover(true, "OIDC discovery failed: HTTP 404 from ...")` rather than exercising the real `handleResponse` 404 branch. The test double bypasses `sendDiscoveryRequest` entirely, so the test asserts the message string matches but does not cover the actual HTTP-status handling logic in `OidcDiscovery.handleResponse`. Not a blocker — the URL-validation and parse paths are covered — but the test name is slightly misleading.
- `ConfigModel.Context.authServerUrl` is correctly added to `clearAuth()`, and `@ToString.Exclude` is consistently applied alongside the other sensitive fields. Good.

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
vandanayadav7: Fixed. The discovered `token_endpoint` is validated via a shared `validateHttpUrl` helper that checks for valid URI syntax, http/https scheme, and non-empty hostname before it's returned and persisted. Rejects `file://` and other non-network URLs with a clear error message. Additionally, `userInfo` is stripped from the discovery URI and a 1MB response size limit is enforced.

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
vandanayadav7: Fixed. Replaced the hard 3xx error with Vert.x request-level redirect following (`req.setFollowRedirects(true)` with `setMaxRedirects(5)`). The client now transparently follows redirect chains (trailing-slash normalization, LB hops) and only reports the final response status.

vandanayadav7:
Addressed the review feedback:

**Token endpoint validation** — The discovered `token_endpoint` is validated via a shared `validateHttpUrl` helper that checks for valid URI syntax, http/https scheme, and non-empty hostname before persisting. Rejects `file://` and other non-network URLs. Additionally, `userInfo` is stripped from the discovery URI and a 1MB response size limit is enforced.

**Redirect handling** — Replaced the hard 3xx error with Vert.x request-level redirect following (`req.setFollowRedirects(true)`, max 5 hops). Common redirect chains (trailing-slash normalization, LB hops) are now handled transparently.

**Test coverage** — Added `testLoginOAuth2DiscoveryReturnsInvalidTokenEndpointScheme` for the new validation. The edge cases @carlesarnal mentioned are already covered by existing tests: `testLoginOAuth2InvalidAuthServerUrl` (invalid URL format), `testLoginOAuth2DiscoveryFailureHttpStatus` (HTTP error response), `testLoginOAuth2DiscoveryReturnsNoTokenEndpoint` (missing token_endpoint in valid JSON).

**ObjectMapper** — Keeping the static `Mapper.MAPPER` instance. `ObjectMapper` is thread-safe and a static singleton is the standard Jackson pattern. Injecting it would add unnecessary ceremony for a CLI tool with no benefit.

**Annotation syntax / corrupted diff** — That was a GitHub diff rendering issue, not a code problem. The annotations compile correctly.

paoloantinori:
Thanks @vandanayadav7. One thing I want to flag: the comment mentions adding `validateTokenEndpointUrl` to validate the discovered `token_endpoint`, but at the current head (`2483cdf`) I don't see that method in the diff — `extractTokenEndpoint` still returns `tokenEndpointNode.asText()` directly, with no scheme/host check on the discovered value.

The `--auth-server-url` input is validated via `validateAuthServerUrl`, which is good. But the `token_endpoint` value that comes back from the discovery document is the one actually stored in `context.setTokenEndpoint(...)` and later used for the token request. If that value is malformed or carries a non-http(s) scheme, it would persist unchecked.

Could you confirm whether the push with `validateTokenEndpointUrl` went through? If the head just hasn't been updated yet, no worries — ping me once it's pushed and I'll re-read.

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
```diff
@@ -0,0 +1,191 @@
...
+
+    private static void handleResponse(final io.vertx.core.http.HttpClientResponse response,
+                                       final URI uri, final CompletableFuture<String> future) {
+        final int statusCode = response.statusCode();
+        if (statusCode != HTTP_OK) {
+            future.completeExceptionally(new CliException(
+                    "OIDC discovery failed: HTTP " + statusCode + " from " + uri,
+                    APPLICATION_ERROR_RETURN_CODE));
+        } else {
+            response.body()
```
paoloantinori: The `MAX_RESPONSE_BYTES` check on the next line fires *after* Vert.x has already buffered the entire response body into memory. For a rogue or misconfigured server streaming a large payload, the OOM would happen during buffering, before this check runs. A `Content-Length` header pre-check here (rejecting before buffering if the header exceeds the limit) would catch the obvious case earlier. Worth tracking as a hardening follow-up.

paoloantinori (commented):
Solid feature — OIDC `.well-known/openid-configuration` discovery with clean precedence (`--token-endpoint` > discovery > error), proper config-model integration, and thorough CLI-level test coverage for option parsing, precedence, conflicts, and failure modes. The validation concern from my earlier comment is addressed: `validateTokenEndpointUrl` is now present and called after extracting the `token_endpoint` from the discovery document.

Two observations worth considering:

- **HTTP discovery code is untested.** `TestOidcDiscovery` overrides `discoverTokenEndpoint` entirely and returns a canned URL, which means the real network code (`fetchDiscoveryDocument`, `sendDiscoveryRequest`, `handleResponse`, redirect following, SSL, timeout, 1MB size limit) has zero test coverage. These are the most security-relevant paths — network I/O against a potentially untrusted OIDC provider. At minimum, a test with a mock HTTP server (e.g. WireMock) exercising the happy path, a redirect chain, and an oversized response would close the gap.

- **Size limit check is post-buffer.** In `handleResponse`, the `MAX_RESPONSE_BYTES` check runs after `response.body()` has fully buffered the response into memory. A rogue server streaming gigabytes would OOM before the check fires. A `Content-Length` header pre-check (rejecting early if the header exceeds the limit) or a streaming body handler with a capped accumulator would be more robust.

Neither is blocking for an initial merge, but the untested network code is worth tracking as a follow-up.

cli/src/test/java/io/apicurio/registry/cli/TestOidcDiscovery.java
```diff
@@ -0,0 +1,52 @@
...
+        this.failMessage = null;
+    }
+
+    public void setFailOnDiscover(final boolean fail, final String message) {
+        this.failOnDiscover = fail;
+        this.failMessage = message;
+    }
+
+    @Override
+    public String discoverTokenEndpoint(final String authServerUrl) {
```
paoloantinori, line 36: This stub overrides `discoverTokenEndpoint` entirely, so the real HTTP code in `OidcDiscovery` (`fetchDiscoveryDocument`, redirect following, SSL, timeout, size limit) has no test coverage. The CLI orchestration tests are thorough, but the network-level code — the most security-relevant part — is only exercised manually. A follow-up with a mock HTTP server (WireMock or similar) covering the happy path, redirect chain, and oversized response would be valuable.

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
vandanayadav7: Fixed. Added a `Content-Length` header pre-check that rejects before buffering when the declared size exceeds 1MB. The post-buffer check remains as a fallback for chunked or missing `Content-Length` headers.

cli/src/test/java/io/apicurio/registry/cli/TestOidcDiscovery.java
vandanayadav7: Added `OidcDiscoveryHttpTest` — 7 WireMock tests that exercise the real `OidcDiscovery` HTTP code (bypassing the test double via direct instantiation with a programmatic Vert.x instance). Covers: happy path, redirect following, HTTP 404, missing `token_endpoint`, invalid scheme (`file://`), oversized response via `Content-Length`, and invalid JSON.

vandanayadav7:
> Thanks @vandanayadav7. One thing I want to flag: the comment mentions adding `validateTokenEndpointUrl` to validate the discovered `token_endpoint`, but at the current head (`2483cdf`) I don't see that method in the diff — `extractTokenEndpoint` still returns `tokenEndpointNode.asText()` directly, with no scheme/host check on the discovered value.
> 
> The `--auth-server-url` input is validated via `validateAuthServerUrl`, which is good. But the `token_endpoint` value that comes back from the discovery document is the one actually stored in `context.setTokenEndpoint(...)` and later used for the token request. If that value is malformed or carries a non-http(s) scheme, it would persist unchecked.
> 
> Could you confirm whether the push with `validateTokenEndpointUrl` went through? If the head just hasn't been updated yet, no worries — ping me once it's pushed and I'll re-read.

Sorry about that — my AI tooling replied before the code was pushed.

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
```diff
@@ -0,0 +1,204 @@
...
+                    .onFailure(future::completeExceptionally);
+        }
+    }
+
+    private static CliException unwrapOrWrap(final Exception ex, final URI uri) {
+        if (ex.getCause() instanceof CliException ce) {
+            return ce;
+        }
+        return new CliException("OIDC discovery failed for '" + uri + "': " + ex.getMessage(), ex,
+                APPLICATION_ERROR_RETURN_CODE);
```
paoloantinori, line 261: Security consideration: `validateHttpUrl` only checks scheme (http/https) and a non-blank host, so the discovered `token_endpoint` is allowed to point at any host — e.g. a discovery document could return `"token_endpoint": "https://attacker.example/token"`. Since this endpoint receives the `--client-secret` during the token exchange, it's worth binding it to the `--auth-server-url` origin (same scheme+host+port) and failing or warning on a mismatch, per RFC 8414 §3. Hedge: fine for a trusted IdP, but the credential flow makes it worth tightening.

paoloantinori (commented):
Re-reviewing at the current head (`a7bd4a0`) after the earlier rounds. The `validateTokenEndpointUrl` addition is a good improvement over the first cut — the discovered `token_endpoint` is now scheme-checked and required to carry a host. One security consideration below that I think is worth a follow-up, plus a couple of smaller notes. Overall the feature still looks solid and the test coverage (WireMock HTTP tests + CLI-level edge cases) is thorough.

- **Discovered `token_endpoint` is not bound to the issuer host.** `validateHttpUrl` only enforces an http/https scheme and a non-blank hostname, so a discovery document can advertise a `token_endpoint` pointing at an arbitrary host (`https://attacker.example/...`). Because the resolved endpoint is what the client-credentials exchange targets, a malicious or compromised IdP/discovery endpoint could redirect the `--client-secret` to an attacker-controlled host. RFC 8414 §3 expects metadata endpoints to be under the issuer's control; it's worth considering rejecting (or warning on) a discovered `token_endpoint` whose scheme+host+port don't match the `--auth-server-url` origin. Not a blocker for trusted IdPs, but relevant given this is the credential-bearing call.
- **`http://` token endpoints are accepted.** Same helper allows plaintext http for the discovered endpoint. That's fine for localhost/dev Keycloak, but in a remote/OAuth2 context a non-TLS token endpoint means the client secret travels in cleartext. If you want to harden the default, restricting discovered endpoints to `https://` (while still allowing `http://` only for `localhost`/`127.0.0.1`) would close that gap.
- **Redirect following widens the attack surface above.** `sendDiscoveryRequest` sets `setFollowRedirects(true)` with `MAX_REDIRECTS = 5`. Combined with the missing issuer-host check, a 302 chain on the discovery request itself could land the fetch on an unexpected host. Lower priority than the `token_endpoint` check, but related.

Minor/nice-to-have, not blocking:
- `handleResponse` reads `Content-Length` and also re-checks `buffer.length()` post-body — good defense in depth. One subtlety: a missing/malformed `Content-Length` falls through to the post-buffer check, which only fires after the full body is already buffered in memory (still capped at 1 MB, so acceptable).
- The precedence behavior (`--token-endpoint` wins, `authServerUrl` not stored in that case) is well covered by `testLoginOAuth2TokenEndpointTakesPrecedence` — nice.

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
vandanayadav7: Fixed. Added `originMatches` check that verifies the discovered `token_endpoint`'s scheme+host+port matches the `--auth-server-url` origin (per RFC 8414 §3). Also replaced `setFollowRedirects(true)` with manual redirect handling that validates each hop against the original host — so neither the discovery fetch nor the token endpoint can be redirected to an attacker-controlled host.

vandanayadav7:
@carlesarnal @paoloantinori I've addressed the review feedback — could you take another look when you get a chance? Thanks!

paoloantinori:
Thanks @vandanayadav7 — re-reviewed at the current head (`893042c`). `validateTokenEndpointUrl` is now present and goes beyond a scheme check: the `originMatches` comparison (scheme + host + effective port) between the auth-server URL and the discovered `token_endpoint` is a real hardening — it stops a misconfigured or tampered discovery document from silently redirecting the token endpoint to a different host. The WireMock test (`testDiscoveryRejectsTokenEndpointFromDifferentOrigin`) covers that path. The redirect-following origin restriction and the `Content-Length`/post-buffer size cap round out the security surface well. Looks good to me.

cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+        final var location = response.getHeader("Location");
+        if (location == null || location.isBlank()) {
+            future.completeExceptionally(new CliException(
+                    "OIDC discovery received a redirect with no Location header.",
+                    APPLICATION_ERROR_RETURN_CODE));
+            return;
+        }
+        final URI redirectUri;
+        try {
```
paoloantinori, line 177: A relative or protocol-relative `Location` (e.g. `Location: /realms/x/.well-known/openid-configuration`, which RFC 7231 permits and some IdPs emit — for instance a trailing-slash redirect) parses here with a null scheme/host. It then fails the origin check at line 173 and is reported as a 'different origin' error, so discovery fails instead of following the same-origin redirect, which contradicts the PR's stated 'follows up to 5 hops transparently' behavior. (For clarity: this fails fast with a clean error, not a crash or hang.) To support these, resolve the Location against the current hop's URI before comparing; note `handleRedirect` currently receives only `originalUri`, so it would also need the immediate source URI passed in.

paoloantinori (commented):
Overall this is a well-constructed and well-tested change: origin pinning on both redirects and the discovered token endpoint, redirect/size/timeout bounds, clean error unwrapping, and the new CLI precedence/conflict logic all hold up, and I found no blocker- or major-level defects (verified independently and via refutation). Two minor observations worth a look:

- `cli/src/main/java/io/apicurio/registry/cli/auth/OidcDiscovery.java:166`/`:173` — relative and protocol-relative `Location` values are not resolved against the request URI, so a same-origin relative redirect (RFC-7231-legal; e.g. an IdP redirecting to add a trailing slash) parses with a null scheme/host and is rejected as a 'different origin' error, contradicting the PR's stated transparent-redirect behavior. To be clear, this is *not* a crash or timeout — `String.equalsIgnoreCase(null)` returns false, so it fails fast with a clean message; the gap is purely that legitimate same-origin relative redirects are not followed.
- `OidcDiscovery.java:205` — `response.body()` fully materializes the response before the `buffer.length()` check at :207, so the advertised 'two-layer' size protection only bounds memory when the server's `Content-Length` (checked at :195) is honest; a chunked or oversized body from the operator-supplied auth server is buffered before rejection. Impact is low given the URL is trusted and the token endpoint is origin-pinned, but the framing slightly overstates the protection.

carlesarnal:
Reviewed at `bf90e34`. The feature is well-structured — origin pinning on both redirects and the discovered `token_endpoint`, `Content-Length` pre-check + post-buffer fallback, manual redirect handling with per-hop origin validation, and the precedence/conflict logic are all solid. Test coverage is thorough across both WireMock HTTP tests and CLI-level edge cases.

All concerns from @carlesarnal's initial review on `451b0dc` (URL validation, edge case tests, annotation rendering issue) are addressed in the current HEAD. @paoloantinori's iterative feedback has been incorporated as well (token endpoint validation, origin binding, WireMock tests, size pre-check).

Two minor notes:

1. **`OidcDiscoveryHttpTest.java` — WireMock stub cleanup between tests.** All 9 tests stub `/realms/test/.well-known/openid-configuration` without a `@BeforeEach` calling `wireMock.resetAll()`. WireMock prioritizes the last-registered stub so it works in practice, but stubs accumulate across the class. Adding a reset improves test isolation if tests are ever reordered or parallelized.

2. **Relative redirect `Location` handling** — Already flagged by @paoloantinori on this commit. A relative `Location` header (e.g. `/realms/x/.well-known/openid-configuration`, which RFC 7231 allows) parses with null scheme/host and fails the origin check cleanly rather than following the redirect. Not a crash, just a gap in the 'follows up to 5 hops transparently' behavior for the specific case of same-origin relative redirects. Resolving `Location` against the current request URI before origin comparison would close it.

Neither is a blocker. The behavioral change where `--username` + `--token-endpoint` now produces a conflict error instead of silently ignoring the token endpoint is a good UX improvement — explicitly rejecting ambiguous intent is better than silently dropping an option the user specified.

Overall this looks good to me.

carlesarnal:
Re-reviewed at `bf90e34`. All concerns from the earlier review rounds are addressed in the current HEAD:

- **Token endpoint validation**: `validateTokenEndpointUrl` validates the discovered endpoint for http/https scheme and enforces origin matching (scheme + host + port) against the `--auth-server-url`. The `testDiscoveryRejectsTokenEndpointFromDifferentOrigin` and `testDiscoveryInvalidTokenEndpointScheme` WireMock tests cover both paths.
- **Redirect handling**: Manual redirect following with per-hop origin validation and a 5-hop cap replaces the previous hard 3xx error. `testDiscoveryFollowsRedirects` and `testDiscoveryRejectsRedirectToDifferentHost` verify the happy and rejection paths.
- **Discovery URI sanitization**: `buildDiscoveryUri` strips `userInfo` from the auth server URL before constructing the `.well-known` URI.
- **Response size limit**: Content-Length pre-check (`OidcDiscovery.java:197`) + post-buffer fallback (`OidcDiscovery.java:207`) at 1MB. Covered by `testDiscoveryOversizedResponseRejectedByContentLength`.
- **Conflict detection**: `--username` + any OAuth2 option now produces a clear error rather than silently dropping options. Three test cases (`testLoginUsernameWithTokenEndpointConflicts`, `testLoginUsernameWithAuthServerUrlConflicts`, `testLoginUsernameWithClientIdConflicts`) verify this.
- **Precedence logic**: `--token-endpoint` > OIDC discovery > error, with `authServerUrl` only stored when discovery was actually used. `testLoginOAuth2TokenEndpointTakesPrecedence` confirms.

Test coverage is thorough — 9 WireMock HTTP tests exercising the real network code paths plus 13 CLI-level tests in `AuthCommandTest` covering option parsing, precedence, conflicts, and failure modes. CI and SonarCloud are green.

The two minor observations @paoloantinori and @carlesarnal flagged (relative `Location` header handling, WireMock stub accumulation) are non-blocking — the relative redirect case fails cleanly rather than crashing, and the WireMock stubs work correctly due to priority ordering.

No new issues found. Looks good.

paoloantinori:
Thanks @carlesarnal for the thorough re-review at `bf90e34`. Confirmed from my side as well: origin pinning on both redirects and the discovered `token_endpoint`, the `Content-Length` pre-check with post-buffer fallback, manual redirect handling, and the CLI precedence (`--token-endpoint` > discovery > error) are all in place. The WireMock tests cover both the cross-origin and invalid-scheme rejection paths. No further concerns from me.

vandanayadav7:
Thanks @paoloantinori @carlesarnal for the thorough reviews! Addressed all feedback from the last round:

**Fixes:**
- **Relative redirect handling** — `Location` headers are now resolved against the current hop's URI via `URI.resolve()`, so relative/protocol-relative redirects (RFC 7231 §7.1.2) are followed correctly instead of being rejected as "different origin".
- **Streaming response size limit** — Replaced `response.body()` (full buffering) with a chunk-by-chunk handler that aborts mid-stream via `request().reset()` as soon as the 1MB limit is exceeded. Added `AtomicBoolean` guard to prevent post-reset chunk processing.
- **WireMock test isolation** — Added `@BeforeEach` with `wireMock.resetAll()` to prevent stub accumulation across tests.

**Hardening:**
- **Explicit TLS configuration** — `setTrustAll(false)` + `setVerifyHost(true)` on the HTTP client.
- **Per-request timeout** — 5s timeout on individual HTTP requests (within the 10s global timeout) to prevent a slow or unresponsive server from consuming the entire timeout budget on a single hop.
- **Redirect status codes narrowed** — Changed from `300–399` range to `Set.of(301, 302, 303, 307, 308)` to exclude non-redirect 3xx codes like 304.

**New tests:**
- `testDiscoveryFollowsRelativeRedirect` — verifies relative `Location` redirect is followed.
- `testDiscoveryOversizedChunkedResponseRejected` — verifies streaming size limit without `Content-Length`.

paoloantinori:
Thanks @vandanayadav — both fixes landed cleanly at the current head:

- **Relative redirects**: `currentUri.resolve(location)` (line 178) correctly resolves relative and protocol-relative `Location` values against the current hop, with the origin pin still enforced on the resolved URI. The `testDiscoveryFollowsRelativeRedirect` case covers the relative-`Location` path.
- **Streaming size limit**: the chunk-by-chunk handler in `readBodyWithSizeLimit` aborts mid-stream once `accumulated.length() + chunk.length()` exceeds `MAX_RESPONSE_BYTES`, and `testDiscoveryOversizedChunkedResponseRejected` exercises it end to end.

From my side the loop is closed. Looking to @carlesarnal for the final sign-off.

vandanayadav7:
Hi @EricWittmann @jsenko — @paoloantinori and @carlesarnal have both signed off in comments but are OOO before formally approving. Could one of you approve if it looks good? Thanks!

jsenko (approved):

PR #8651, opened 2026-07-15
ci(operator): add Awaitility timeout and Failsafe retry for flaky Kafka tests

paoloantinori:
## Summary

Adds two resilience flags to the operator test Maven command to reduce flaky Kafka test failures:
- `-Dawaitility.defaultTimeout=PT10M`: extends default poll timeout from 7 to 10 minutes for slow broker startups
- `-Dfailsafe.rerunFailingTestsCount=1`: retries failed tests once as a safety net

## Problem

Operator Kafka tests (`KafkaSqlITTest`, `KafkaSqlTLSITTest`, `KafkaSqlOAuthITTest`) intermittently time out on GitHub Actions runners. The Strimzi/Kafka broker sometimes needs longer than 7 minutes to start under CI load. These two flags provide immediate relief while deeper fixes (REG-10 Kafka CR tuning, REG-17 image pre-loading) are implemented.

## Why NOT change the Minikube driver

The operator workflow uses `driver: none` (bare-metal K8s), which gives full access to all runner resources (4 vCPU / 16 GB). Switching to `driver: docker` would break the `localhost:5000` registry used for test image pulls. The `--cpus=max --memory=max` flags are redundant with `driver: none`.

## Test plan

- [ ] Operator Kafka tests pass in CI
- [ ] Awaitility timeout visible in Maven logs (`-Dawaitility.defaultTimeout=PT10M`)
- [ ] Failed tests show retry attempt in surefire reports

### DIFF
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -238,6 +238,8 @@ jobs:
           name: operator-test-images
           path: /tmp/
 
+      # driver=none: Kubernetes runs directly on the host, sharing all runner
+      # resources (4 vCPU / 16 GB) and the host network (localhost:5000 registry).
       - uses: apicurio/apicurio-github-actions/setup-minikube@97d452c41f1abd7142c0ce2ac37e87922e35e71d # v2
         with:
           driver: 'none'
@@ -269,10 +271,14 @@ jobs:
           docker push "$REGISTRY_GITOPS_SYNC_IMAGE"
           docker push "$IMAGE"
 
+      # -Dawaitility.defaultTimeout=PT10M: extend default poll timeout from 7 min to 10 min
+      #   for slow Kafka/OLM startups on shared runners.
+      # -Dfailsafe.rerunFailingTestsCount=1: retry failed tests once as a safety net for
+      #   transient resource-related flakes.
       - name: Run ${{ matrix.name }} tests on Minikube
         working-directory: operator
         run: |
-          make BUILD_OPTS="--no-transfer-progress -Daether.syncContext.named.factory=noop -Dtest.operator.deployment-target=minikube ${{ matrix.extra-build-opts }}" \
+          make BUILD_OPTS="--no-transfer-progress -Daether.syncContext.named.factory=noop -Dtest.operator.deployment-target=minikube -Dawaitility.defaultTimeout=PT10M -Dfailsafe.rerunFailingTestsCount=1 ${{ matrix.extra-build-opts }}" \
             REMOTE_TESTS_ALL_INSTALL_FILE=controller/target/test-install.yaml \
             GROUPS="${{ matrix.groups }}" \
             EXCLUDED_GROUPS="${{ matrix.excluded-groups }}" \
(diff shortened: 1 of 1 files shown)


EricWittmann (commented):
## Code Review

### `-Dfailsafe.rerunFailingTestsCount=1` — Looks Good

This follows the same pattern already used in `verify-integration-tests.yaml` (which uses `=2`), and the root `pom.xml` is wired to pass it through to the Failsafe plugin. Using `1` (retry once) is a reasonable choice for operator tests. No issues here.

### `-Dawaitility.defaultTimeout=PT10M` — Might Not Work

I took a look at the Awaitility source and it does not appear to read any system properties — the
`Awaitility` class contains zero calls to `System.getProperty`. The only way to set the default
timeout is programmatically via `Awaitility.setDefaultTimeout()`.

Additionally, the operator tests already override the default programmatically in `ITBase.java`:

```java
static void setDefaultAwaitilityTimings() {
    Awaitility.setDefaultPollInterval(POLL_INTERVAL_DURATION);
    Awaitility.setDefaultTimeout(LONG_DURATION);  // 420 seconds (7 min)
}
```

This programmatic call would take precedence even if the system property were supported.

The test code already has a system property wired up for this purpose. `LONG_DURATION` in
`ITBase.java` reads `test.operator.timeout.long`:

```java
public static final Duration LONG_DURATION = ofSeconds(
        Integer.getInteger("test.operator.timeout.long", 420));
```

So `-Dtest.operator.timeout.long=600` should be the correct flag to use instead.

There are also Kafka-specific timeout properties that might be more targeted:
- `test.operator.timeout.kafka-broker` (default 600s)
- `test.operator.timeout.kafka-registry` (default 480s)

Could you investigate and confirm whether the `awaitility.defaultTimeout` system property actually
has any effect? If not, switching to `-Dtest.operator.timeout.long=600` (or the Kafka-specific
properties) should give you the timeout extension you're after.

### Comments

The added comments explaining the `driver: none` rationale and the flag purposes are helpful context
for this workflow file.

paoloantinori:
@EricWittmann Good catch — you're right, Awaitility doesn't read system properties. Fixed in commit 07c6add: replaced `-Dawaitility.defaultTimeout=PT10M` with `-Dtest.operator.timeout.long=600`, which is already wired into `ITBase.LONG_DURATION` via `Integer.getInteger()`. Updated the comment to document the mechanism.

carlesarnal (approved):
CI configuration change to address flaky Kafka/OLM tests in operator module. Two additions:

1. `-Dtest.operator.timeout.long=600` extends Awaitility timeout from 7 to 10 minutes
2. `-Dfailsafe.rerunFailingTestsCount=1` retries failed tests once

Comments are clear about rationale. Timeout increase is reasonable for shared CI runners. Retry count of 1 is conservative enough to catch real failures while filtering transient flakes.

Minor concern: Retry mechanism could mask legitimate test failures temporarily, but this is standard practice for flaky CI environments.

`.github/workflows/operator.yaml`

jsenko:
Would it make sense to just change the default Awaitility timeouts? That will only delay reporting failed tests, and might help with flakiness everywhere.

PR #8657, opened 2026-07-15
fix: resolve silent failures in Documentation tab for PROMPT_TEMPLATE…

### LINKED ISSUE #8656: Bug: Documentation tab silently fails for PROMPT_TEMPLATE and MODEL_SCHEMA artifacts
### Describe the bug

The Documentation tab in the Version detail page has multiple interconnected issues that cause it to silently fail for `PROMPT_TEMPLATE` and `MODEL_SCHEMA` artifact types. These types use the `getArtifactVersionContentDereferenced()` API, which triggers a code path with several bugs.

### Root Cause Analysis

**1. Error state setter never destructured (dead code)**

In `DocumentationTabContent.tsx`, the `useState` for error handling was missing the setter function:

```typescript
// Was:
const [error] = useState<any>();
// setError was never available, making ErrorTabContent permanently unreachable
```

The `isError()` check on line 127 always returned `false`, and the `ErrorTabContent` fallback on line 131 was unreachable dead code.

**2. Dereference API failure is completely silent**

When `getArtifactVersionContentDereferenced()` fails (401, 404, 500, network error), the `.catch()` block was empty — no error display, no fallback to raw content:

```typescript
.catch(() => {
    // Comment says "fall back to raw content" but actually does nothing
});
```

The user sees a blank Documentation tab with no feedback.

**3. `parseContent` returns primitives for non-object content**

`YAML.parse()` from the `yaml` library does not throw on plain text strings — it returns the string itself (YAML is a superset of plain text). When a `PROMPT_TEMPLATE` contains raw Jinja/Handlebars text, `parseContent` returned a string primitive, not an object. Downstream visualizers expecting `spec.template`, `spec.name`, etc. received `undefined` for every field.

**4. No `useEffect` cleanup for async dereference call**

The `useEffect` that calls `getArtifactVersionContentDereferenced()` had no cleanup function. If the user navigated between versions rapidly, responses could resolve out of order, displaying wrong version content.

### Steps to reproduce

1. Create a `PROMPT_TEMPLATE` artifact with a raw text template (not JSON/YAML object)
2. Navigate to its Version page → Documentation tab
3. Simulate a dereference API failure (e.g., disconnect network)
4. Observe: blank Documentation tab, no error message

### Expected behavior

- On dereference failure: fall back to displaying raw artifact content
- If raw content is also unavailable: display error via `ErrorTabContent`
- Rapid version navigation: only show the latest version's content

### Actual behavior

- Blank Documentation tab with no feedback
- `ErrorTabContent` is unreachable dead code
- Rapid navigation can show stale/wrong content

### Affected artifact types

- `PROMPT_TEMPLATE`
- `MODEL_SCHEMA`

(These are the only types that trigger the `needsDereference()` code path)

### Environment

- Apicurio Registry UI
- React 19, TypeScript 5.9, PatternFly 6.6

KumarADITHYA123:
fixes #8656 

### What this PR does

Fixes four interconnected bugs in `DocumentationTabContent.tsx` that caused the Documentation tab to silently fail for `PROMPT_TEMPLATE` and `MODEL_SCHEMA` artifact types.

#### 1. Destructure `setError` from `useState`

The error state setter was never extracted, making `ErrorTabContent` permanently unreachable.

```diff
-    const [error] = useState<any>();
+    const [error, setError] = useState<any>();
```

#### 2. Add type safety to `parseContent()`

`YAML.parse()` returns the input string itself for plain text content (YAML is a superset of plain text). Added `typeof` guards to ensure only actual objects are returned to visualizers.

```diff
     try {
-        return JSON.parse(artifactContent);
+        const json = JSON.parse(artifactContent);
+        if (typeof json === "object" && json !== null) {
+            return json;
+        }
     } catch {
         // Not valid JSON
     }
     try {
-        return YAML.parse(artifactContent);
+        const yaml = YAML.parse(artifactContent);
+        if (typeof yaml === "object" && yaml !== null) {
+            return yaml;
+        }
     } catch {
         // Not valid YAML
     }
+    return {};
```

#### 3. Add `useEffect` cleanup + error fallback for dereference

- Added `cancelled` flag to prevent stale async state updates on unmount or rapid navigation
- Implemented the catch fallback: tries raw `versionContent` first, falls back to `setError()` if unavailable

```diff
     useEffect(() => {
+        let cancelled = false;
         if (needsDereference(...)) {
             groups.getArtifactVersionContentDereferenced(...)
                 .then(content => {
-                    setParsedContent(parseContent(content));
+                    if (!cancelled) {
+                        setParsedContent(parseContent(content));
+                    }
                 })
-                .catch(() => {
-                    // Fall back to raw content if dereference fails
+                .catch((err) => {
+                    if (!cancelled) {
+                        if (props.versionContent) {
+                            setParsedContent(parseContent(props.versionContent));
+                        } else {
+                            setError(err);
+                        }
+                    }
                 });
         }
+        return () => { cancelled = true; };
     }, [...]);
```

### Testing

- Verified `PROMPT_TEMPLATE` artifact: dereference failure now falls back to raw content
- Verified rapid navigation between versions: no stale content displayed
- Verified `ErrorTabContent` is now reachable when both dereference and raw content fail
- Verified `OPENAPI` and `ASYNCAPI` types are unaffected (they don't trigger `needsDereference()`)
- Verified `parseContent` correctly rejects string primitives from `YAML.parse()`

### Files changed

- `ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx`

### DIFF
--- a/ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
+++ b/ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
@@ -1,7 +1,7 @@
 import { FunctionComponent, useEffect, useState } from "react";
 import { ErrorTabContent } from "@app/pages";
 import { If } from "@apicurio/common-ui-components";
-import YAML from "yaml";
+import { isJson, parseJson, isYaml, parseYaml } from "@utils/content.utils";
 import {
     AgentCardVisualizer,
     AsyncApiVisualizer,
@@ -43,21 +43,23 @@ const getVisualizerType = (artifactType: string): VisualizerType => {
     return VisualizerType.OTHER;
 };
 
-const parseContent = (artifactContent: string): any => {
-    // Try as JSON
-    try {
-        return JSON.parse(artifactContent);
-    } catch {
-        // Do nothing
+/**
+ * Parses artifact content string into a structured object for visualization.
+ * Returns the parsed object, or undefined if parsing fails or produces a
+ * non-object value. Downstream visualizers destructure named properties
+ * (e.g. spec.name, spec.template), so primitives are rejected.
+ *
+ * Built on top of content.utils.ts utilities (isJson, parseJson, isYaml, parseYaml).
+ */
+const parseContent = (artifactContent: string): any | undefined => {
+    if (isJson(artifactContent)) {
+        const json = parseJson(artifactContent);
+        if (typeof json === "object" && json !== null) return json;
     }
-
-    // Try as YAML
-    try {
-        return YAML.parse(artifactContent);
-    } catch {
-        // Do nothing
+    if (isYaml(artifactContent)) {
+        return parseYaml(artifactContent); // isYaml already rejects primitives
     }
-    return {};
+    return undefined;
 };
 
 /**
@@ -82,37 +84,48 @@ const needsDereference = (artifactType: string): boolean => {
 export const DocumentationTabContent: FunctionComponent<DocumentationTabContentProps> = (props: DocumentationTabContentProps) => {
     const [parsedContent, setParsedContent] = useState(() =>
         props.versionContent && !needsDereference(props.artifactType)
-            ? parseContent(props.versionContent) : {}
+            ? (parseContent(props.versionContent) ?? {}) : {}
     );
     const [visualizerType] = useState(getVisualizerType(props.artifactType));
-    const [error] = useState<any>();
+    const [error, setError] = useState<any>();
     const groups: GroupsService = useGroupsService();
 
     // Handles race condition where versionContent arrives after mount.
     useEffect(() => {
+        setError(undefined); // Always clear error when deps change, including dereference types
         if (props.versionContent && !needsDereference(props.artifactType)) {
-            setParsedContent(parseContent(props.versionContent));
+            setParsedContent(parseContent(props.versionContent) ?? {});
         }
     }, [props.versionContent, props.artifactType]);
 
     useEffect(() => {
+        let cancelled = false;
+        setError(undefined); // Clear stale error when dependencies change
         if (needsDereference(props.artifactType) && props.groupId && props.artifactId && props.version) {
             groups.getArtifactVersionContentDereferenced(props.groupId, props.artifactId, props.version)
                 .then(content => {
-                    setParsedContent(parseContent(content));
+                    // Cancelled check inside callback prevents state updates after unmount
+                    if (!cancelled) {
+                        setParsedContent(parseContent(content) ?? {});
+                    }
                 })
-                .catch(() => {
-                    // Fall back to raw content if dereference fails
+                .catch((err) => {
+                    if (!cancelled) {
+                        // Fall back to raw content if dereference fails
+                        if (props.versionContent) {
+                            console.warn("[DocumentationTabContent] Dereference failed, falling back to raw content:", err);
+                            setParsedContent(parseContent(props.versionContent) ?? {});
+                        } else {
+                            setError(err);
+                        }
+                    }
                 });
         }
-    }, [props.artifactType, props.groupId, props.artifactId, props.version]);
+        return () => { cancelled = true; };
+    }, [props.artifactType, props.groupId, props.artifactId, props.version, props.versionContent]);
 
-    const isError = () : boolean => {
-        return !!error;
-    };
-
-    if (isError()){
-        return <ErrorTabContent error={{ errorMessage: "Artifact isn't a valid OpenAPI structure", error: error }}/>;
+    if (error) {
+        return <ErrorTabContent error={{ errorMessage: "Failed to load artifact documentation", error: error }}/>;
     }
 
     return (


ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -85,7 +91,7 @@ export const DocumentationTabContent: FunctionComponent<DocumentationTabContentP
             ? parseContent(props.versionContent) : {}
     );
     const [visualizerType] = useState(getVisualizerType(props.artifactType));
-    const [error] = useState<any>();
+    const [error, setError] = useState<any>();
```
EricWittmann, line 90: Good fix — this was a real bug. However, now that `setError` is wired up, `ErrorTabContent` on line 132 becomes reachable for `PROMPT_TEMPLATE` and `MODEL_SCHEMA` artifacts. The hardcoded message there says:

```tsx
"Artifact isn't a valid OpenAPI structure"
```

…which is misleading for non-OpenAPI types. Since this PR is the one that makes it user-visible, could you update it to something type-agnostic? For example:

```tsx
return <ErrorTabContent error={{ errorMessage: "Failed to load artifact documentation", error: error }}/>;
```

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -96,15 +102,26 @@ export const DocumentationTabContent: FunctionComponent<DocumentationTabContentP
...
+                        setParsedContent(parseContent(content));
+                    }
                 })
-                .catch(() => {
-                    // Fall back to raw content if dereference fails
+                .catch((err) => {
+                    if (!cancelled) {
+                        // Fall back to raw content if dereference fails
+                        if (props.versionContent) {
+                            setParsedContent(parseContent(props.versionContent));
```
EricWittmann: Question about the fallback behavior here. What happens when the dereference API fails and the raw `versionContent` is plain text (e.g. a Jinja template like `"Hello {{ name }}"`)?

The flow would be:
1. Dereference fails → catch fires
2. `props.versionContent` is truthy → calls `parseContent("Hello {{ name }}")`
3. `JSON.parse` fails; `YAML.parse` returns the string itself; the new `typeof` guard rejects it
4. `parseContent` returns `{}`
5. Visualizer renders with all-`undefined` fields → empty display, **no error shown**

The `setError` path only triggers when `props.versionContent` is falsy, so this scenario results in a silent empty visualizer.

One option would be to check whether the parsed result is actually usable:

```tsx
if (props.versionContent) {
    const parsed = parseContent(props.versionContent);
    if (Object.keys(parsed).length > 0) {
        setParsedContent(parsed);
    } else {
        setError(err);
    }
} else {
    setError(err);
}
```

Is an empty-object parse result ever legitimate here, or should it always be treated as an error? Would appreciate opinions from anyone familiar with these artifact types.

EricWittmann (commented):
Thanks for the fix — the core changes (wiring up `setError`, the `parseContent` type guards, the `useEffect` cleanup, and the catch-block fallback) all look correct. Two comments inline.

carlesarnal (changes requested):
The race condition fix is incomplete. The `cancelled` flag prevents state updates after unmount but doesn't prevent the promise callback from executing. Move the cancellation check inside the `.then()` callback.

The `parseContent` function rejects valid JSON/YAML that isn't objects. Valid JSON includes `null`, strings, numbers, arrays. Either accept all valid JSON/YAML or document why only objects are allowed.

Error state management is broken. Once `setError` is called, the error persists even if subsequent loads succeed. Clear the error in the success path or when props change.

The fallback logic in the catch block treats empty objects as failures. `Object.keys(parsed).length > 0` will be false for `{}`, triggering an error when it should display empty documentation.

`ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx`

Fix the race condition by checking `cancelled` in the `.then()` callback. Fix the empty object check. Clear error state on successful loads. Consider whether non-object JSON/YAML should be supported.

KumarADITHYA123:
Updated the PR to address all four review points:

**1. Race condition:** The `cancelled` flag was already inside both `.then()` and `.catch()` callbacks in the previous revision. No structural change needed — added a comment for clarity. The cleanup function returns `cancelled = true` on unmount/re-render.

**2. parseContent object-only behavior:** Updated the JSDoc to accurately document what the `typeof` guard does. Primitives (strings, numbers, null) are rejected because visualizers destructure named properties like `spec.name` and `spec.template`. Arrays technically pass `typeof === "object"` but these artifact types (PROMPT_TEMPLATE, MODEL_SCHEMA) always produce plain objects — this matches the original behavior before this PR.

**3. Error state persistence:** Added `setError(undefined)` at the top of both `useEffect` hooks so stale errors are cleared whenever dependencies change. Removed the redundant `setError(undefined)` that was inside `.then()` since the error is already cleared before the async call fires.

**4. Empty object fallback:** Removed the `Object.keys(parsed).length > 0` check. `{}` is a valid JSON document and should be displayed by the visualizer rather than treated as an error. The `setError(err)` path now only triggers when `props.versionContent` is falsy (no content exists at all).

EricWittmann (commented):
## Code Review — 7 findings (high effort, 8 angles)

Overall the PR correctly identifies and fixes real bugs: the dead `setError`, the missing catch-handler logic, and the `YAML.parse()` primitive leak. The `useEffect` cleanup pattern is a good addition. The findings below are ranked most-severe first — see inline comments for details.

| # | Severity | Category | Summary |
|---|----------|----------|---------|
| 1 | PLAUSIBLE bug | correctness | Error state never cleared for dereference types after failure |
| 2 | PLAUSIBLE bug | correctness | `ErrorTabContent` renders `'{}'` for native `Error` objects |
| 3 | PLAUSIBLE | simplification | `parseContent` duplicates existing `content.utils.ts` utilities |
| 4 | CONFIRMED | simplification | `parseContent` returning `{}` on failure hides parse errors |
| 5 | CONFIRMED | simplification | `isError()` trivial one-use wrapper |
| 6 | PLAUSIBLE | correctness | Silent degradation on dereference failure (no logging/indicator) |
| 7 | PLAUSIBLE | correctness | Missing `props.versionContent` in useEffect dep array (latent) |

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -43,19 +43,33 @@ const getVisualizerType = (artifactType: string): VisualizerType => {
...
-        // Do nothing
+        // Not valid JSON
     }
 
     // Try as YAML
     try {
-        return YAML.parse(artifactContent);
+        const yaml = YAML.parse(artifactContent);
+        if (typeof yaml === "object" && yaml !== null) {
+            return yaml;
```
EricWittmann: **Finding 1 — Error state never cleared for dereference types (PLAUSIBLE bug)**

The first `useEffect` clears error with `setError(undefined)` only inside the `!needsDereference()` guard, so it never fires for `MODEL_SCHEMA` / `PROMPT_TEMPLATE`. The second `useEffect` clears error at its top but only re-runs when `[artifactType, groupId, artifactId, version]` change.

**Scenario:** Dereference call fails → `setError(err)` is called → error page shown. If `versionContent` subsequently becomes available without the second effect's deps changing, neither `useEffect` clears the error. The error page is permanently displayed until the user navigates away.

**Suggested fix:** Either remove the `!needsDereference()` guard from the `setError(undefined)` call in the first `useEffect`, or ensure both effects clear error unconditionally when their deps change.

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -85,34 +99,48 @@ export const DocumentationTabContent: FunctionComponent<DocumentationTabContentP
...
-    const [error] = useState<any>();
+    const [error, setError] = useState<any>();
     const groups: GroupsService = useGroupsService();
 
     // Handles race condition where versionContent arrives after mount.
     useEffect(() => {
         if (props.versionContent && !needsDereference(props.artifactType)) {
+            setError(undefined);
             setParsedContent(parseContent(props.versionContent));
         }
```
EricWittmann, line 98: **Finding 1 — Error state never cleared for dereference types (PLAUSIBLE bug)**

This `setError(undefined)` is inside the `!needsDereference()` guard, so it never fires for `MODEL_SCHEMA` / `PROMPT_TEMPLATE`. The second `useEffect` clears error at its top (line 117) but only re-runs when `[artifactType, groupId, artifactId, version]` change.

**Scenario:** Dereference call fails → `setError(err)` → error page shown. If `versionContent` subsequently becomes available without the second effect's deps changing, neither `useEffect` clears the error — the error page is permanently displayed until the user navigates away.

**Suggested fix:** Remove the `!needsDereference()` guard from this `setError(undefined)` call so error is cleared unconditionally when `versionContent` changes:

```tsx
useEffect(() => {
    setError(undefined); // Always clear error when versionContent changes
    if (props.versionContent && !needsDereference(props.artifactType)) {
        setParsedContent(parseContent(props.versionContent));
    }
}, [props.versionContent, props.artifactType]);
```

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -85,34 +99,48 @@ export const DocumentationTabContent: FunctionComponent<DocumentationTabContentP
...
-                    // Fall back to raw content if dereference fails
+                .catch((err) => {
+                    if (!cancelled) {
+                        // Fall back to raw content if dereference fails
+                        if (props.versionContent) {
+                            setParsedContent(parseContent(props.versionContent));
+                        } else {
+                            setError(err);
+                        }
+                    }
```
EricWittmann, line 121: **Finding 2 — ErrorTabContent renders `'{}'` for native Error objects (PLAUSIBLE bug)**

This `setError(err)` now makes `ErrorTabContent` reachable for the first time (good!), but `ErrorTabContent.errorDetail()` falls through to `JSON.stringify(error, null, 3)` for objects without `.detail`/`.title`. Native JavaScript `Error` objects have non-enumerable `message`/`stack`/`name` properties, so `JSON.stringify(new Error("something"))` produces `'{}'`.

While most API failures produce Axios errors (which serialize fine), network failures or code-level exceptions reaching this catch would show an empty `{}` in the details panel.

This is a pre-existing issue in `ErrorTabContent`, but now newly reachable. Consider filing a follow-up to handle native `Error` objects in `errorDetail()`.

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -43,19 +43,33 @@ const getVisualizerType = (artifactType: string): VisualizerType => {
...
+/**
+ * Parses artifact content string into a structured object for visualization.
+ * Only object types are accepted because downstream visualizers (PromptTemplateVisualizer,
+ * ModelSchemaVisualizer, etc.) destructure named properties (e.g. spec.name,
+ * spec.template). Primitive JSON/YAML values (strings, numbers, null) are rejected
+ * since property access on them would fail silently. Arrays pass the typeof guard
+ * but are not expected for these artifact types in practice.
+ */
 const parseContent = (artifactContent: string): any => {
     // Try as JSON
```
EricWittmann: **Finding 3 — `parseContent` duplicates existing `content.utils.ts` utilities (PLAUSIBLE)**

`src/utils/content.utils.ts` already exports `parseJson()`, `parseYaml()`, `isJson()`, and `isYaml()` with the same JSON.parse / YAML.parse try-catch pattern. The `isYaml()` function even has the same `typeof === "object"` guard this PR adds here.

Consider building `parseContent` on top of those utilities:

```tsx
import { isJson, parseJson, isYaml, parseYaml } from "@utils/content.utils";

const parseContent = (artifactContent: string): any => {
    if (isJson(artifactContent)) {
        const json = parseJson(artifactContent);
        if (typeof json === "object" && json !== null) return json;
    }
    if (isYaml(artifactContent)) {
        return parseYaml(artifactContent);  // isYaml already rejects primitives
    }
    return {};
};
```

This keeps the guard logic centralized.

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -43,19 +43,33 @@ const getVisualizerType = (artifactType: string): VisualizerType => {
...
-        return YAML.parse(artifactContent);
+        const yaml = YAML.parse(artifactContent);
+        if (typeof yaml === "object" && yaml !== null) {
+            return yaml;
+        }
     } catch {
-        // Do nothing
+        // Not valid YAML
     }
     return {};
```
EricWittmann: **Finding 4 — `return {}` hides parse failures from the user (CONFIRMED, pre-existing)**

When neither JSON nor YAML parsing produces an object, this returns `{}`. Downstream visualizers receive an empty spec and render a blank view — indistinguishable from "still loading" or "legitimately empty content."

Returning `undefined` here would let the component detect failure and show `ErrorTabContent` instead of a silently blank visualization. Pre-existing issue but worth noting since `parseContent` is being reworked in this PR.

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -85,34 +99,48 @@ export const DocumentationTabContent: FunctionComponent<DocumentationTabContentP
...
+                        }
+                    }
                 });
         }
+        return () => { cancelled = true; };
     }, [props.artifactType, props.groupId, props.artifactId, props.version]);
 
     const isError = () : boolean => {
         return !!error;
     };
```
EricWittmann, line 126: **Finding 5 — `isError()` trivial one-use wrapper (CONFIRMED, pre-existing)**

`isError()` wraps `!!error` and is used exactly once (line 144). `if (error)` is equivalent and more idiomatic React. Pre-existing but since the PR touches this area (wiring up `setError`), it could be simplified in the same pass.

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -85,34 +99,48 @@ export const DocumentationTabContent: FunctionComponent<DocumentationTabContentP
...
                 })
-                .catch(() => {
-                    // Fall back to raw content if dereference fails
+                .catch((err) => {
+                    if (!cancelled) {
+                        // Fall back to raw content if dereference fails
+                        if (props.versionContent) {
+                            setParsedContent(parseContent(props.versionContent));
+                        } else {
+                            setError(err);
```
EricWittmann, line 119: **Finding 6 — Silent degradation on dereference failure (PLAUSIBLE)**

When the dereference API fails and `versionContent` is available, this silently swaps in un-dereferenced content. The user sees a visualization with potentially unresolved `$ref` references but no indication that they are viewing degraded content.

Consider adding a `console.warn` here at minimum, or a subtle UI banner ("showing un-dereferenced content") so users (and developers debugging) know the dereference failed.

ui/ui-app/src/app/pages/version/components/tabs/DocumentationTabContent.tsx
```diff
@@ -85,34 +99,48 @@ export const DocumentationTabContent: FunctionComponent<DocumentationTabContentP
...
+                        }
+                    }
                 });
         }
+        return () => { cancelled = true; };
     }, [props.artifactType, props.groupId, props.artifactId, props.version]);
 
     const isError = () : boolean => {
```
EricWittmann: **Finding 7 — Missing `props.versionContent` in dependency array (PLAUSIBLE, latent)**

The catch handler at line 129 reads `props.versionContent`, but it is not in this dependency array. Currently `PageDataLoader` ensures `versionContent` is loaded before this component mounts, so the closure captures the correct value in practice.

However, if `PageDataLoader`'"'"'s behavior ever changes or this component is reused outside that wrapper, the closure would capture a stale value. Adding `props.versionContent` to the dep array (or using a ref) would make the code robust regardless of parent behavior. The `eslint-plugin-react-hooks` `exhaustive-deps` rule would also flag this.

KumarADITHYA123:
Thanks for the detailed review @EricWittmann ! Updated in `1b33ca329`.

For Finding 1 — yeah, I missed that setError(undefined) was inside the !needsDereference() guard. Moved it outside so error always clears when versionContent/artifactType change, regardless of artifact type.

Finding 2 — agreed, that's a pre-existing issue in ErrorTabContent with JSON.stringify on native Error objects. I can file a follow-up issue for that if you'd like.

Finding 3 — good call, I didn't realize content.utils.ts already had isJson/parseJson/isYaml/parseYaml. Rewrote parseContent to use those instead of duplicating the try/catch. Dropped the direct YAML import too.

Finding 4 — parseContent now returns undefined on failure. Callers still use ?? {} as a fallback since the visualizers receive spec directly, but the undefined return makes it possible to detect parse failure in the future.

Finding 5 — removed isError(), just using if (error) now.

Finding 6 — added a console.warn in the catch fallback path so it's not completely silent anymore.

Finding 7 — added props.versionContent to the second useEffect dep array. Makes sense since the catch block reads it.

EricWittmann (approved):
All review findings addressed in 1b33ca329. Clean implementation — LGTM.

EricWittmann:
:clap:

PR #8659, opened 2026-07-15
fix(protobuf): ignore google/protobuf imports in ALL_REFS_MAPPED integrity checks

### LINKED ISSUE #4877: Unmapped reference error for google import in protobuf schema

niy-ati:
Fixes #4877

Reopens work from #8560 

## What
\ALL_REFS_MAPPED\ was treating well-known \google/protobuf/*\ imports as unmapped references. Reuse \ProtobufReferenceFinder\ in \ProtobufContentValidator.validateReferences\ (same pattern as avro).

## Testing
- \ProtobufContentValidatorTest#testValidateReferencesIgnoresGoogleProtobufImports\
- \ProtobufContentValidatorTest#testValidateReferencesStillRequiresCustomImportsWithGoogleOnes\
- both pass locally

### DIFF
--- a/schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufContentValidator.java
+++ b/schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/validity/ProtobufContentValidator.java
@@ -4,6 +4,8 @@ import com.squareup.wire.schema.SchemaException;
 import com.squareup.wire.schema.internal.parser.MessageElement;
 import com.squareup.wire.schema.internal.parser.ProtoFileElement;
 import io.apicurio.registry.content.TypedContent;
+import io.apicurio.registry.content.refs.ExternalReference;
+import io.apicurio.registry.protobuf.content.refs.ProtobufReferenceFinder;
 import io.apicurio.registry.rest.v3.beans.ArtifactReference;
 import io.apicurio.registry.rules.validity.AbstractContentValidator;
 import io.apicurio.registry.rules.validity.ValidityLevel;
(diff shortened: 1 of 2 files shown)


carlesarnal (approved):
The fix correctly addresses issue #4877 by reusing `ProtobufReferenceFinder` to filter google/protobuf well-known imports during reference validation. This aligns validation behavior with auto-reference discovery.

Implementation is clean: replaced manual import collection with the existing reference finder that already handles google import filtering. Test coverage is solid with both positive and negative cases.

One minor note: the test annotations in the diff show file paths instead of `@Test`. Verify this is just a diff display issue and not actual code.

The change is backward compatible and makes validation behavior more intuitive - google/protobuf runtime imports no longer require registry mappings.

niy-ati:
Thanks @carlesarnal , 
The @Test annotations are correct in `1a3b220`. The file path display was a GitHub diff artifact from the earlier #8560 review [non ASCII comment text confused the diff renderer].

Verified locally:
- `testValidateReferencesIgnoresGoogleProtobufImports`
- `testValidateReferencesStillRequiresCustomImportsWithGoogleOnes`

Both pass.

jsenko:
@niy-ati The PR branch is currently behind main and has the `lifecycle/wip` label, which is blocking auto-merge. Could you please rebase on main and run `/ready` so the orchestrator can proceed?

PR #8660, opened 2026-07-15
fix(operator): add diagnostic dumps for OLM test setup failures (#8658)

### LINKED ISSUE #8658: Add diagnostic dumps to operator OLM tests on failure
## Problem

When OLM operator tests fail (e.g. `SmokeOLMITTest` timing out with
`ConditionTimeoutException` in `OLMITBase.beforeAll`), there is no diagnostic
information captured about the state of OLM resources in the test namespace.
The test namespace is deleted by the Java test layer during cleanup, so any
external diagnostic tooling that runs afterward finds nothing — no
CatalogSources, no Subscriptions, no InstallPlans, no pods, no events.

This makes it very difficult to determine root cause. For example, a recent
downstream test failure timed out after 5 minutes waiting for the operator to
become available via OLM, but we couldn't tell whether:
- The catalog image failed to pull
- The CatalogSource was created but the catalog pod didn't become ready
- The Subscription was created but no InstallPlan was generated
- The InstallPlan was created but the CSV didn't reach `Succeeded` phase
- The operator pod started but never became ready

## Proposed Solution

Add diagnostic dumps in the Java test layer (e.g. `OLMITBase`) that fire
**before namespace deletion** when a test fails. The dumps should capture:

1. **CatalogSources** in the test namespace — status, ready condition
2. **Subscriptions** — status, current CSV, install plan ref
3. **InstallPlans** — phase, approval status, conditions
4. **ClusterServiceVersions** — phase, conditions, reason
5. **Pods** in the test namespace — status, ready conditions, restart counts
6. **Pod logs** for catalog and operator pods (tail ~50 lines)
7. **Events** in the test namespace sorted by timestamp (last ~30)
8. **Image pull events** — any `Failed` or `BackOff` events related to image
   pulls, which would indicate authentication or mirror configuration issues

This information should be logged via the test framework's logging (e.g.
`LOG.info(...)`) so it appears in the Maven surefire output.

## Context

This is relevant for both upstream and downstream testing. Downstream tests
use IIB catalog images from `brew.registry.redhat.io` which require pull
secret injection and IDMS/ITMS mirror configuration on the OCP cluster. Image
pull failures are a common root cause of OLM test timeouts in this scenario,
but without in-test diagnostics they are invisible.

The external test runner (`run-operator-tests.sh`) already has a
`dump_olm_diagnostics` function, but it runs after the Java test has already
cleaned up the namespace, so it only captures cluster-wide state (default
catalog sources, marketplace events) — not the test-specific resources.

EricWittmann:
## Summary
- Wrap `OLMITBase.beforeAll()` OLM setup in a try-catch that calls
  `ClusterDiagnostics.dump()` before rethrowing — the core fix for the
  `@BeforeAll` gap where `AfterTestExecutionCallback` never fires
- Add `CatalogSource` (`operators.coreos.com/v1alpha1`) to the OLM resource
  dump — previously missing from `dumpOLMResources()`
- Add dedicated `dumpImagePullEvents()` section that filters for image-pull
  related failures (`ErrImagePull`, `ImagePullBackOff`, `BackOff`) for easy
  scanning of the most common downstream failure mode
- Wrap `UpgradeOLMITTest.deployCatalogAndSubscribe()` with the same
  diagnostic dump pattern for OLM setup failures during upgrade tests

## Related Issue
Fixes #8658

## Test Plan
- [ ] Verify the `operator/controller` module compiles with `./mvnw compile -pl operator/controller -am -DskipTests`
- [ ] Verify the `operator/olm-tests` module compiles with `./mvnw compile -pl operator/olm-tests -am -DskipTests`
- [ ] Review that `ClusterDiagnostics.dump()` is called in the `catch` block of `OLMITBase.beforeAll()` before the exception is rethrown
- [ ] Review that CatalogSource is now included in `dumpOLMResources()`
- [ ] Review that image pull events are filtered and displayed in a dedicated section
- [ ] CI OLM test suite passes (functional verification requires a Kubernetes cluster with OLM)

### DIFF
--- a/operator/controller/src/test/java/io/apicurio/registry/operator/utils/ClusterDiagnostics.java
+++ b/operator/controller/src/test/java/io/apicurio/registry/operator/utils/ClusterDiagnostics.java
@@ -47,6 +47,7 @@ public final class ClusterDiagnostics {
         dumpNetworkPolicies(client, namespace);
         dumpPods(client, namespace);
         dumpEvents(client, namespace);
+        dumpImagePullEvents(client, namespace);
 
         if (isOLM) {
             dumpOLMResources(client, namespace);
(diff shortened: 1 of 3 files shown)


carlesarnal (changes requested):
Adds diagnostic dumps when OLM test setup fails. Good addition for debugging flaky CI failures.

Minor logic issue in `ClusterDiagnostics.dumpImagePullEvents()`: the second filter's condition `"Failed".equalsIgnoreCase(e.getReason())` doesn't align well with the first filter. If you want events with reason "Failed", they won't match the first filter's patterns (Pull/BackOff/ErrImage). Either remove this condition or clarify the intent.

The null timestamp handling in the comparator works but could be more explicit. Consider `Comparator.nullsLast()` for clarity.

Exception handling is appropriate for diagnostic code. Catching and re-throwing with diagnostics is exactly what you want here.

Files: `operator/controller/src/test/java/io/apicurio/registry/operator/utils/ClusterDiagnostics.java`, `operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/OLMITBase.java`, `operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/UpgradeOLMITTest.java`

jsenko:

PR #8662, opened 2026-07-16
fix(rules): detect MCP tool parameter type and enum breaking changes

### LINKED ISSUE #8661: McpToolCompatibilityChecker misses breaking parameter schema changes (type change, enum narrowing)
## Description

**Registry Version**: snapshot (build from `main`)
**Persistence type**: any (the bug is in `schema-util`, storage-independent)

`McpToolCompatibilityChecker` only compares parameter **names** (`checkPropertyRemovals`), the `required` list, and the top-level `inputSchema.type`. The schema of each parameter is never inspected, so breaking changes to a parameter pass BACKWARD compatibility undetected:

1. **Parameter type change** — `"query": {"type": "string"}` → `"query": {"type": "integer"}` is reported compatible, but every existing caller sending `{"query": "foo"}` is now invalid.
2. **Enum narrowing** — `"format": {"enum": ["json", "xml", "csv"]}` → `{"enum": ["json"]}` is reported compatible, but callers sending `"xml"` or `"csv"` now break.

FORWARD and FULL levels are equally affected, since they reuse the same `isBackwardsCompatibleWith` implementation via `AbstractCompatibilityChecker`.

Notably, `PromptTemplateCompatibilityChecker` in the same package already detects both of these cases for template variables (`checkVariableTypeChange`, `checkEnumNarrowing`), so MCP tool parameters — the contract agents actually invoke tools with — currently get weaker compatibility governance than prompt variables. The original MCP_TOOL feature issue (#7227) lists "Breaking change detection" among its planned Version Management features, and "Version management and compatibility rules apply" is one of its acceptance criteria.

### Environment

Not environment-specific — reproducible on any build of `main` regardless of persistence type or deployment. The defect is in the storage-independent compatibility checker in `schema-util/common`.

### Steps to Reproduce

Unit-level (verified against current `main`):

```java
@Test
void parameterTypeChangeShouldBeBackwardIncompatible() {
    String existing = """
            {
                "name": "search_tool",
                "inputSchema": {
                    "type": "object",
                    "properties": {
                        "query": { "type": "string" }
                    },
                    "required": ["query"]
                }
            }
            """;

    // Same property name, but its type changed string -> integer.
    String proposed = """
            {
                "name": "search_tool",
                "inputSchema": {
                    "type": "object",
                    "properties": {
                        "query": { "type": "integer" }
                    },
                    "required": ["query"]
                }
            }
            """;

    CompatibilityExecutionResult result = new McpToolCompatibilityChecker().testCompatibility(
            CompatibilityLevel.BACKWARD,
            List.of(TypedContent.create(ContentHandle.create(existing), ContentTypes.APPLICATION_JSON)),
            TypedContent.create(ContentHandle.create(proposed), ContentTypes.APPLICATION_JSON),
            Map.of());

    assertFalse(result.isCompatible()); // FAILS: checker reports COMPATIBLE
}
```

An equivalent test for enum narrowing fails the same way (checker reports compatible).

At the REST level: create an `MCP_TOOL` artifact with a `COMPATIBILITY: BACKWARD` rule, then publish a new version that changes an existing parameter's type — the version is accepted instead of being rejected with a rule violation.

**Expected:** changing an existing parameter's type, or removing values from its `enum`, is a backward-incompatible change and should produce a `CompatibilityDifference` (rejecting the new version under a BACKWARD rule) — consistent with how `PromptTemplateCompatibilityChecker` treats the same changes on variables.

**Actual:** `testCompatibility` returns compatible; the breaking version is accepted.

### Logs

Not applicable — there is no error or stack trace. The checker silently returns a compatible result and the breaking version is admitted, which is the bug.

## Possible fix

For properties present in both versions, compare the per-property schemas — at minimum flag `type` changes and enum narrowing, mirroring the prompt-template checker. Each incompatibility should be reported as a `SimpleCompatibilityDifference(message, contextPath)` with a context such as `/inputSchema/properties/{name}/type` or `.../enum` — matching the difference model that #8634 (compatibility-checker type-hierarchy refactor) collapses the per-checker `*CompatibilityDifference` classes into. Since `inputSchema` is JSON Schema per the MCP spec, a deeper option is delegating property comparison to the existing JSON Schema compatibility infrastructure, but the minimal per-property checks already cover the common breaking cases.

I'd be happy to work on this — I have failing tests reproducing both cases and a patch sketch following the `PromptTemplateCompatibilityChecker` pattern.

## Related

- **#7227** — MCP Tool Definition Artifact Type (feature list includes "Breaking change detection").
- **#8427** — MCP tool validation/compatibility program issue. That covers *tool-to-tool* compatibility (one tool's output feeding another's input) and new endpoints; this bug is about the existing *version-to-version* checker `testCompatibility`.
- **#8184 / PR #8634** — compatibility-checker type-hierarchy refactor, which rewrites `McpToolCompatibilityChecker` and replaces `McpToolCompatibilityDifference` with `SimpleCompatibilityDifference`. Orthogonal to this behavioral gap, but a fix here should build on whatever shape #8634 lands.

piyushbag:
## Problem
`McpToolCompatibilityChecker` only compared parameter names, the `required` list, and the top-level `inputSchema.type`. Parameter type changes and enum narrowing were treated as compatible under BACKWARD/FORWARD/FULL.

## Approach
For properties present in both versions, compare each property schema the same way `PromptTemplateCompatibilityChecker` does for template variables:
- flag `type` changes
- flag enum narrowing (removed enum values)

Added `PROPERTY_TYPE_CHANGED` and `ENUM_VALUE_REMOVED` difference types, plus unit tests for both cases.

## Test plan
- [x] `./mvnw -pl schema-util/util-provider -am test -Dtest=McpToolCompatibilityCheckerTest -Dsurefire.failIfNoSpecifiedTests=false` (11 tests, 0 failures)

Fixes #8661

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
@@ -18,6 +18,8 @@ import java.util.Set;
  * - Adding required parameters: Backward incompatible
  * - Removing required parameters: Backward incompatible
  * - Changing inputSchema type: Backward incompatible
+ * - Changing an existing parameter's type: Backward incompatible
+ * - Narrowing an existing parameter's enum: Backward incompatible
  * - Changing name, title, description, annotations: Always compatible
  */
 public class McpToolCompatibilityChecker
@@ -39,6 +41,9 @@ public class McpToolCompatibilityChecker
 
             // Check removed properties
             checkPropertyRemovals(existingNode, proposedNode, differences);
+
+            // Check per-property schema changes for properties present in both versions
+            checkPropertySchemaChanges(existingNode, proposedNode, differences);
 
             // Check added required parameters
             checkRequiredParamAdditions(existingNode, proposedNode, differences);
@@ -80,6 +85,79 @@ public class McpToolCompatibilityChecker
                         "Input property '" + prop + "' was removed"));
             }
         }
+    }
+
+    private void checkPropertySchemaChanges(JsonNode existing, JsonNode proposed,
+            Set<McpToolCompatibilityDifference> differences) {
+        JsonNode existingProps = getInputSchemaProperties(existing);
+        JsonNode proposedProps = getInputSchemaProperties(proposed);
+        if (existingProps == null || proposedProps == null) {
+            return;
+        }
+
+        Iterator<String> existingNames = existingProps.fieldNames();
+        while (existingNames.hasNext()) {
+            String propName = existingNames.next();
+            if (!proposedProps.has(propName)) {
+                continue;
+            }
+            JsonNode existingProp = existingProps.get(propName);
+            JsonNode proposedProp = proposedProps.get(propName);
+            checkPropertyTypeChange(propName, existingProp, proposedProp, differences);
+            checkEnumNarrowing(propName, existingProp, proposedProp, differences);
+        }
+    }
+
+    private void checkPropertyTypeChange(String propName, JsonNode existingProp, JsonNode proposedProp,
+            Set<McpToolCompatibilityDifference> differences) {
+        String existingType = getTextValue(existingProp, "type");
+        String proposedType = getTextValue(proposedProp, "type");
+        if (existingType != null && proposedType != null && !existingType.equals(proposedType)) {
+            differences.add(new McpToolCompatibilityDifference(
+                    McpToolCompatibilityDifference.Type.PROPERTY_TYPE_CHANGED,
+                    "Input property '" + propName + "' type changed from '" + existingType
+                            + "' to '" + proposedType + "'"));
+        }
+    }
+
+    private void checkEnumNarrowing(String propName, JsonNode existingProp, JsonNode proposedProp,
+            Set<McpToolCompatibilityDifference> differences) {
+        JsonNode existingEnum = existingProp.get("enum");
+        JsonNode proposedEnum = proposedProp.get("enum");
+        if (existingEnum == null || !existingEnum.isArray()
+                || proposedEnum == null || !proposedEnum.isArray()) {
+            return;
+        }
+
+        Set<String> proposedValues = new HashSet<>();
+        for (JsonNode val : proposedEnum) {
+            proposedValues.add(val.asText());
+        }
+
+        for (JsonNode val : existingEnum) {
+            if (!proposedValues.contains(val.asText())) {
+                differences.add(new McpToolCompatibilityDifference(
+                        McpToolCompatibilityDifference.Type.ENUM_VALUE_REMOVED,
+                        "Input property '" + propName + "' enum value '" + val.asText()
+                                + "' was removed"));
+            }
+        }
+    }
+
+    private JsonNode getInputSchemaProperties(JsonNode node) {
+        JsonNode inputSchema = node.get("inputSchema");
+        if (inputSchema != null && inputSchema.isObject()) {
+            JsonNode props = inputSchema.get("properties");
+            if (props != null && props.isObject()) {
+                return props;
+            }
+        }
+        return null;
+    }
+
+    private String getTextValue(JsonNode node, String fieldName) {
+        JsonNode field = node.get(fieldName);
+        return (field != null && field.isTextual()) ? field.asText() : null;
     }
 
     private void checkRequiredParamAdditions(JsonNode existing, JsonNode proposed,
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.apicurio.registry.content.TypedContent;
 
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -32,6 +33,12 @@ public class McpToolCompatibilityChecker
 
     private static final ObjectMapper mapper = new ObjectMapper();
 
+    private static final String FIELD_INPUT_SCHEMA = "inputSchema";
+    private static final String FIELD_PROPERTIES = "properties";
+    private static final String FIELD_TYPE = "type";
+    private static final String FIELD_ENUM = "enum";
+    private static final String FIELD_REQUIRED = "required";
+
     @Override
     public CompatibilityExecutionResult testCompatibility(CompatibilityLevel compatibilityLevel,
             List<TypedContent> existingArtifacts, TypedContent proposedArtifact,
@@ -52,34 +59,26 @@ public class McpToolCompatibilityChecker
 
         switch (compatibilityLevel) {
             case BACKWARD:
-                incompatibleDiffs = doCompatibilityCheck(lastExistingSchema, proposedArtifactContent,
-                        resolvedReferences, false);
+                incompatibleDiffs = doCompatibilityCheck(lastExistingSchema, proposedArtifactContent, false);
                 break;
             case BACKWARD_TRANSITIVE:
-                incompatibleDiffs = transitivelyCheck(existingArtifacts, proposedArtifactContent,
-                        resolvedReferences, false);
+                incompatibleDiffs = transitivelyCheck(existingArtifacts, proposedArtifactContent, false);
                 break;
             case FORWARD:
-                incompatibleDiffs = doCompatibilityCheck(proposedArtifactContent, lastExistingSchema,
-                        resolvedReferences, true);
+                incompatibleDiffs = doCompatibilityCheck(proposedArtifactContent, lastExistingSchema, true);
                 break;
             case FORWARD_TRANSITIVE:
-                incompatibleDiffs = transitivelyCheck(existingArtifacts, proposedArtifactContent,
-                        resolvedReferences, true);
+                incompatibleDiffs = transitivelyCheck(existingArtifacts, proposedArtifactContent, true);
                 break;
             case FULL:
-                incompatibleDiffs = unionOf(
-                        doCompatibilityCheck(lastExistingSchema, proposedArtifactContent,
-                                resolvedReferences, false),
-                        doCompatibilityCheck(proposedArtifactContent, lastExistingSchema,
-                                resolvedReferences, true));
+                incompatibleDiffs = mergeDifferenceSets(
+                        doCompatibilityCheck(lastExistingSchema, proposedArtifactContent, false),
+                        doCompatibilityCheck(proposedArtifactContent, lastExistingSchema, true));
                 break;
             case FULL_TRANSITIVE:
-                incompatibleDiffs = unionOf(
-                        transitivelyCheck(existingArtifacts, proposedArtifactContent,
-                                resolvedReferences, false),
-                        transitivelyCheck(existingArtifacts, proposedArtifactContent,
-                                resolvedReferences, true));
+                incompatibleDiffs = mergeDifferenceSets(
+                        transitivelyCheck(existingArtifacts, proposedArtifactContent, false),
+                        transitivelyCheck(existingArtifacts, proposedArtifactContent, true));
                 break;
             case NONE:
                 break;
@@ -91,24 +90,22 @@ public class McpToolCompatibilityChecker
     }
 
     private Set<McpToolCompatibilityDifference> transitivelyCheck(List<TypedContent> existingArtifacts,
-            String proposedArtifactContent, Map<String, TypedContent> resolvedReferences,
-            boolean forwardEnumCheck) {
+            String proposedArtifactContent, boolean forwardEnumCheck) {
         Set<McpToolCompatibilityDifference> result = new HashSet<>();
         for (int i = existingArtifacts.size() - 1; i >= 0; i--) {
             String existing = existingArtifacts.get(i).getContent().content();
             if (forwardEnumCheck) {
-                result.addAll(doCompatibilityCheck(proposedArtifactContent, existing,
-                        resolvedReferences, true));
+                result.addAll(doCompatibilityCheck(proposedArtifactContent, existing, true));
             } else {
-                result.addAll(doCompatibilityCheck(existing, proposedArtifactContent,
-                        resolvedReferences, false));
+                result.addAll(doCompatibilityCheck(existing, proposedArtifactContent, false));
             }
         }
         return result;
     }
 
     @SafeVarargs
-    private Set<McpToolCompatibilityDifference> unionOf(Set<McpToolCompatibilityDifference>... from) {
+    private Set<McpToolCompatibilityDifference> mergeDifferenceSets(
+            Set<McpToolCompatibilityDifference>... from) {
         Set<McpToolCompatibilityDifference> result = new HashSet<>();
         for (Set<McpToolCompatibilityDifference> set : from) {
             result.addAll(set);
@@ -119,11 +116,11 @@ public class McpToolCompatibilityChecker
     @Override
     protected Set<McpToolCompatibilityDifference> isBackwardsCompatibleWith(String existing,
             String proposed, Map<String, TypedContent> resolvedReferences) {
-        return doCompatibilityCheck(existing, proposed, resolvedReferences, false);
+        return doCompatibilityCheck(existing, proposed, false);
     }
 
     private Set<McpToolCompatibilityDifference> doCompatibilityCheck(String existing, String proposed,
-            Map<String, TypedContent> resolvedReferences, boolean forwardEnumCheck) {
+            boolean forwardEnumCheck) {
         Set<McpToolCompatibilityDifference> differences = new HashSet<>();
 
         try {
@@ -189,6 +186,7 @@ public class McpToolCompatibilityChecker
             return;
         }
 
+        // Only iterate existing property names; additions are handled elsewhere.
         Iterator<String> existingNames = existingProps.fieldNames();
         while (existingNames.hasNext()) {
             String propName = existingNames.next();
@@ -208,7 +206,7 @@ public class McpToolCompatibilityChecker
             Set<McpToolCompatibilityDifference> differences) {
         Set<String> fromTypes = getTypeValues(fromProp);
         Set<String> toTypes = getTypeValues(toProp);
-        if (fromTypes == null || toTypes == null) {
+        if (fromTypes.isEmpty() || toTypes.isEmpty()) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


carlesarnal (changes requested):
Implementation is solid. Logic correctly detects type changes and enum narrowing as backward incompatible. Code follows project conventions. Null safety is adequate.

Fix the test annotations. They reference wrong file path (`@utils/converter/src/test/java/io/apicurio/registry/utils/converter/avro/AvroDataTest.java`). Should be `@Test`.

Add test cases for edge conditions: null type fields, null enum fields, and combinations of breaking changes. Current tests only cover happy path scenarios.

Consider adding a test where a property is added (should be compatible) to verify the `continue` logic in `checkPropertySchemaChanges` works correctly.

File paths:
- schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
- schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityDifference.java
- schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -82,6 +87,79 @@ private void checkPropertyRemovals(JsonNode existing, JsonNode proposed,
...
+        String proposedType = getTextValue(proposedProp, "type");
+        if (existingType != null && proposedType != null && !existingType.equals(proposedType)) {
+            differences.add(new McpToolCompatibilityDifference(
+                    McpToolCompatibilityDifference.Type.PROPERTY_TYPE_CHANGED,
+                    "Input property '" + propName + "' type changed from '" + existingType
+                            + "' to '" + proposedType + "'"));
+        }
+    }
+
+    private void checkEnumNarrowing(String propName, JsonNode existingProp, JsonNode proposedProp,
```
paoloantinori: This check is asymmetric under argument swap, and FORWARD/FULL compatibility invert the arguments via `AbstractCompatibilityChecker`. Concretely: adding an enum value (`["json"]` → `["json","xml"]`) is flagged as a removal under FORWARD, while an actual narrowing (`["json","xml"]` → `["json"]`) is *not* flagged under FORWARD. The PR description claims BACKWARD/FORWARD/FULL coverage, so I think this needs to be addressed — either direction-aware logic or an explicit FORWARD pass. Happy to sketch a fix.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -82,6 +87,79 @@ private void checkPropertyRemovals(JsonNode existing, JsonNode proposed,
...
+        if (inputSchema != null && inputSchema.isObject()) {
+            JsonNode props = inputSchema.get("properties");
+            if (props != null && props.isObject()) {
+                return props;
+            }
+        }
+        return null;
+    }
+
+    private String getTextValue(JsonNode node, String fieldName) {
```
paoloantinori: JSON Schema allows `type` to be an array (e.g. `["string","null"]`). Since `isTextual()` is false for arrays, any array-typed `type` is treated as `null` here, so changes to/from multi-type definitions are silently ignored. Worth handling if the goal is faithful JSON-Schema `inputSchema` comparison.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
```diff
@@ -344,4 +344,219 @@ void testFullCompatibilityBothDirections() {
 
         assertTrue(result.isCompatible(), "Identical schema with description change should be fully compatible");
     }
```
paoloantinori, line 346: All five new tests use `CompatibilityLevel.BACKWARD`. A FORWARD test for enum-value addition (should be compatible) and removal (should not) would lock in the intended semantics and would also surface the FORWARD asymmetry in `checkEnumNarrowing`.

paoloantinori (commented):
Nice extension of `McpToolCompatibilityChecker` — mirroring what `PromptTemplateCompatibilityChecker` already does for variable types/enums is the right instinct, and the BACKWARD tests cover the new paths well. The implementation has one real correctness issue that's worth addressing before merge, plus two smaller gaps.

- **Forward / Full compatibility handling of enum narrowing is inverted (main concern).** `AbstractCompatibilityChecker` evaluates `FORWARD` by swapping the two arguments and calling `isBackwardsCompatibleWith(proposed, existing, ...)` ([source](https://github.com/Apicurio/apicurio-registry/blob/main/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AbstractCompatibilityChecker.java)). `checkEnumNarrowing` is not symmetric under that swap, so FORWARD gives the wrong answer in both directions:
  - *Adding* an enum value (old `enum: ["json"]` → new `enum: ["json","xml"]`) should be FORWARD-compatible — an old client never sends the new value. Under the swapped call, `existing=["json","xml"]` is iterated and `"xml"` is not found in `proposed=["json"]`, so `ENUM_VALUE_REMOVED` is emitted → incorrectly flagged incompatible.
  - *Actually narrowing* the enum in a way that breaks FORWARD (old `["json","xml"]` → new `["json"]`) is **not** detected: under the swap, `existing=["json"]`, so the removed `"xml"` is never iterated. FORWARD reports compatible when it shouldn't.
  - `FULL` = union(BACKWARD, FORWARD), so it inherits the same defect.
  The PR description explicitly claims the change applies to BACKWARD/FORWARD/FULL, so this isn't an edge case for the stated scope. `PromptTemplateCompatibilityChecker.checkEnumNarrowing` has the same shape, so it may share the issue — worth a follow-up there too. Suggested fix: make the check direction-aware (the `CompatibilityLevel` is not passed to the per-rule methods today, so this likely needs a small refactor in `AbstractCompatibilityChecker` or an explicit forward-mode pass), or at minimum document the limitation and restrict the new checks to BACKWARD.
- **`type` as a JSON-Schema array is silently ignored.** `getTextValue` only accepts textual `type` nodes, so a change like `"type": "string"` → `"type": ["string","null"]` (a widening that is arguably backward-compatible) or the reverse (a real narrowing) is not flagged. JSON Schema permits `type` to be an array, and MCP `inputSchema` is JSON Schema, so this will come up. Consider normalizing to a sorted set of allowed types and comparing the sets (narrowing = removing an allowed type).
- **No FORWARD/FULL test coverage for the new paths.** The five new tests all exercise `CompatibilityLevel.BACKWARD`. Given the PR description names all three levels, at least one FORWARD test (e.g. enum-value *addition* should be compatible; *removal* should not) would both lock in the intended semantics and immediately surface the asymmetry above.

Minor: the two new `McpToolCompatibilityDifference.Type` entries reuse the `"inputSchema/properties"` context alongside `PROPERTY_REMOVED`, which is fine for reporting — just flagging it so the context string is intentional.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
piyushbag: Addressed in 96b1f20. `McpToolCompatibilityChecker` now overrides `testCompatibility` and passes a `forwardEnumCheck` flag into property schema checks. For FORWARD/FULL, enum and type narrowing compare values from the prior schema against the new schema (`fromProp=proposed`, `toProp=existing`) instead of reusing the backward-only path under argument swap.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
piyushbag: Addressed in 96b1f20. Replaced textual-only `getTextValue` with `getTypeValues`, which normalizes both string and array JSON Schema `type` values into a set before comparison. Added tests for backward-incompatible narrowing (`["string","null"]` to `"string"`) and backward-compatible widening.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
piyushbag: Addressed in 96b1f20. Added `testForwardCompatibleEnumValueAddition`, `testForwardIncompatibleEnumNarrowing`, and array-type widening/narrowing tests. All new tests use `@Test` from JUnit 5.

piyushbag:
Review feedback addressed in 96b1f20 (rebased onto main):

- FORWARD/FULL enum and type checks are now direction-aware via an overridden `testCompatibility`
- JSON Schema array `type` values are normalized to sets before comparison
- Added FORWARD enum add/remove and array-type tests

/ready

piyushbag:
Sonar reliability fix in 9f62392: renamed local `unionOf` to `mergeDifferenceSets` to resolve java:S2177 (method shadowed `AbstractCompatibilityChecker.unionOf`). Verify workflow still needs maintainer approval.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -18,16 +22,109 @@
...
+            case BACKWARD:
+                incompatibleDiffs = doCompatibilityCheck(lastExistingSchema, proposedArtifactContent,
+                        resolvedReferences, false);
+                break;
+            case BACKWARD_TRANSITIVE:
+                incompatibleDiffs = transitivelyCheck(existingArtifacts, proposedArtifactContent,
+                        resolvedReferences, false);
+                break;
+            case FORWARD:
+                incompatibleDiffs = doCompatibilityCheck(proposedArtifactContent, lastExistingSchema,
```
paoloantinori: With `forwardEnumCheck=true`, the type/enum checks end up flagging the same removals as BACKWARD (same `from`/`to` orientation after the swap). That makes enum *addition* and type *widening* FORWARD-compatible (see `testForwardCompatibleEnumValueAddition`), which is a behaviour change from the base class's swapped-args FORWARD. Worth stating that explicitly in the class Javadoc — right now only enum *narrowing* is listed as forward-incompatible.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -18,16 +22,109 @@
...
+ * - Narrowing an existing parameter's enum: Backward incompatible
+ * - Narrowing an existing parameter's enum relative to the prior version: Forward incompatible
  * - Changing name, title, description, annotations: Always compatible
  */
 public class McpToolCompatibilityChecker
         extends AbstractCompatibilityChecker<McpToolCompatibilityDifference> {
 
     private static final ObjectMapper mapper = new ObjectMapper();
 
+    @Override
```
paoloantinori, line 44: This override reproduces nearly the whole `AbstractCompatibilityChecker.testCompatibility` switch just to thread `forwardEnumCheck` through. If a new `CompatibilityLevel` lands upstream, this checker silently won't handle it. Consider a smaller hook (e.g. `protected boolean forwardEnumCheck()`) so the dispatch stays in one place — not blocking, just flagging the maintenance cost.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -82,6 +182,101 @@ private void checkPropertyRemovals(JsonNode existing, JsonNode proposed,
...
+            Set<McpToolCompatibilityDifference> differences, boolean forwardEnumCheck) {
+        JsonNode existingProps = getInputSchemaProperties(existing);
+        JsonNode proposedProps = getInputSchemaProperties(proposed);
+        if (existingProps == null || proposedProps == null) {
+            return;
+        }
+
+        Iterator<String> existingNames = existingProps.fieldNames();
+        while (existingNames.hasNext()) {
+            String propName = existingNames.next();
```
paoloantinori, line 194: The loop only visits `existingNames`, so properties present only in `proposedProps` are intentionally skipped here (they're handled by `checkPropertyAdditions`/`checkRequiredParamAdditions`). A one-line comment confirming the skip is deliberate would help future readers.

paoloantinori (commented):
Re-read after the rebase + sonar fix. The direction-aware `testCompatibility` override now produces the BACKWARD/FORWARD/FULL semantics the new tests assert, and `mergeDifferenceSets` cleanly replaces the shadowed `unionOf`. A few things worth a closer look before this lands:

- **Forward semantics for type/enum widened more than the doc suggests.** With `forwardEnumCheck=true`, `checkPropertySchemaChanges` swaps `fromProp`/`toProp` so FORWARD ends up flagging exactly the same `type`/`enum` removals as BACKWARD (relative to the prior version). The net effect is that *adding* an enum value (`testForwardCompatibleEnumValueAddition`) and *widening* a property type are now treated as FORWARD-compatible. Before this PR, `AbstractCompatibilityChecker.FORWARD` called `isBackwardsCompatibleWith(proposed, existing)` with no extra inversion, so enum addition would have been flagged FORWARD-incompatible (closer to Confluent's strict reading where old consumers don't recognise the new value). The relaxed behaviour is reasonable for Jackson-backed MCP payloads, but the class Javadoc only lists enum *narrowing* as forward-incompatible — it would help to state explicitly that widening is now forward-compatible too, so future readers don't assume parity with the Avro/JSON-Schema checker semantics.
- **Duplicated `switch` from the base class.** The overridden `testCompatibility` reproduces nearly the entire `AbstractCompatibilityChecker.testCompatibility` dispatch, differing only by threading a `forwardEnumCheck` flag into `doCompatibilityCheck`. If the base class evolves (new `CompatibilityLevel`, multi-artifact semantics, etc.), this checker silently diverges. A lighter hook — e.g. a `protected boolean forwardEnumCheck()` override, or splitting `isBackwardsCompatibleWith` into a direction-aware variant — would keep the dispatch in one place. Not blocking, but worth a note for the next refactor of this hierarchy.
- **`resolvedReferences` null check.** The override adds `requireNonNull` for `compatibilityLevel`, `existingArtifacts`, `proposedArtifact` but skips `resolvedReferences`; the base class has the same gap, so this is just an observation while you're already tightening contracts.
- **Minor: properties that exist only in `proposedProps` are intentionally skipped** by the `existingNames` loop in `checkPropertySchemaChanges` (added params are handled by `checkPropertyAdditions`/`checkRequiredParamAdditions`). A one-line comment confirming the skip is deliberate would save the next reader a hop.

Tests look solid — coverage for null type, null enum, array-type widening/narrowing, combined type+enum change, and the FORWARD enum cases is exactly the right surface.

piyushbag:
Review feedback from 2026-07-16 is addressed on head `9f62392`:

- JUnit annotations use `@Test` (not the mistaken path string)
- Edge cases: null property `type`, null `enum`, compatible property add (`testCompatibleWhenPropertyTypeIsNull`, `testCompatibleWhenPropertyEnumIsNull`, `testCompatibleWhenAddingPropertyLeavesSharedSchemasUnchanged`)
- FORWARD/FULL enum handling is direction-aware (override `testCompatibility`)

CI and SonarCloud are green. Could you take another look when you have a moment?

/ready

paoloantinori:
Thanks @piyushbag — the `9f62392` head addresses all the points I raised earlier: `@Test` annotations fixed, the null-type / null-enum edge cases are now explicitly covered (`testCompatibleWhenPropertyTypeIsNull`, `testCompatibleWhenPropertyEnumIsNull`), the compatible-property-add scenario is asserted (`testBackwardCompatibleAddingPropertyWithUnchangedSharedSchema`), and the FORWARD enum narrowing/widening pair correctly exercises the direction-aware path. From my side this is good to go.

paoloantinori (commented):
Seen the update on head `9f62392` — the annotation fix, null-property handling, and null/empty enum edge cases look addressed. Waiting on maintainer re-review (PR is in `lifecycle/waiting-on-author` — you may need to signal readiness).

piyushbag:
Merged upstream/main onto head `939ccdd` (2026-07-21). Prior review fixes on `9f62392` unchanged; CI green on last run. Ready for re-review when convenient.

piyushbag:
Rebased onto latest `main` (`cf1148bb`) and dropped the broken merge commit. SonarCloud findings addressed on head `998f860`:

- Renamed `unionOf` → `mergeDifferenceSets` (no shadowing of `AbstractCompatibilityChecker`)
- `getTypeValues` returns `Collections.emptySet()` instead of null
- Extracted repeated JSON field literals (`inputSchema`, `properties`, `type`, `enum`, `required`) and shared `inputSchema/properties` context constant
- Removed unused `resolvedReferences` from private check paths (MCP tools do not dereference external schemas)
- Added comment clarifying that properties only in the proposed schema are intentionally skipped here

Ready for re-review, @paoloantinori.

paoloantinori:
Thanks @piyushbag — on head `998f860` the rebase onto `cf1148bb` is clean and the SonarCloud findings are addressed: `mergeDifferenceSets` no longer shadows `AbstractCompatibilityChecker.unionOf`, `getTypeValues` returns `Collections.emptySet()` rather than null, and the repeated JSON field literals are extracted into constants. The direction-aware BACKWARD/FORWARD/FULL semantics and the test coverage I reviewed previously are intact. No further changes from my side — ready for a maintainer merge when CI is green.

piyushbag:
SonarCloud is green on the current head (`998f860`) and the MCP tool parameter type/enum breach detection changes from the earlier review pass are complete.

@paoloantinori could you take another look when you have a moment? I believe the content review items from your earlier comments are addressed; happy to adjust anything still open.

paoloantinori:
Thanks @piyushbag — re-reviewed on head `998f860`. The rebase onto `cf1148bb` is clean, SonarCloud is green, and the direction-aware `forwardEnumCheck` threading through `doCompatibilityCheck` / `transitivelyCheck` / `mergeDifferenceSets` is consistent with the base class's swapped-argument semantics for FORWARD/FULL, so no regression in the existing property-removal and required-param checks. The new enum/type narrowing logic and the null-type / null-enum / array-type edge cases all look correct, and the test matrix covers BACKWARD well.

One optional (non-blocking) observation: the BACKWARD type-change and enum-narrowing cases both have dedicated tests (`testBackwardIncompatibleParameterTypeChange`, `testBackwardIncompatibleEnumNarrowing`), but the FORWARD type-change path — which now also produces a difference via the `fromProp`/`toProp` swap in `checkPropertySchemaChanges` — isn't locked in by a test. A small `testForwardIncompatible*` type-change case mirroring `testForwardIncompatibleEnumNarrowing` would document the FORWARD type semantics explicitly. Happy if you leave it for a follow-up.

From my side this is ready for a maintainer merge decision — no further blocking concerns from me.

piyushbag:
@paoloantinori ready for another look when you have a moment.

Rebased onto current `main`. CI was fully green on the prior head (`998f860` Verify run).

Addressed remaining doc nits on head:
- Class Javadoc now states FORWARD-compatible enum addition and JSON Schema type widening (not just forward-incompatible narrowing).
- `checkPropertySchemaChanges` already documents that proposed-only properties are intentionally skipped (handled by addition/required checks).

Happy to refactor the duplicated `testCompatibility` switch into a hook in a follow-up if you still prefer that shape.

paoloantinori:
Thanks @piyushbag — re-read the current head. The rebase onto latest `main` is clean, and the doc nits from the prior round are addressed: the class Javadoc now states forward-compatible enum addition and JSON Schema type widening (not just forward-incompatible narrowing), and `checkPropertySchemaChanges` documents that proposed-only properties are intentionally skipped. The direction-aware `forwardEnumCheck` threading and the `mergeDifferenceSets` rename still look right. From my side this is ready for a maintainer re-review.

carlesarnal:

PR #8664, opened 2026-07-16
feat(rules): validate embedded JSON Schemas in MCP_TOOL artifacts

ADITYA-CODE-SOURCE:
Part of #8427.

This PR focuses on the validation slice of MCP_TOOL support by strengthening validation of embedded inputSchema and outputSchema using the existing JSON Schema infrastructure already present in the repo.

To avoid a module dependency cycle, this is implemented via a JSON-module validator wired through McpToolArtifactTypeUtilProvider, while preserving the existing top-level MCP envelope checks.

Changes:
- add McpToolJsonSchemaContentValidator in schema-util/json
- switch McpToolArtifactTypeUtilProvider to use the new validator
- add provider-wiring coverage and deep invalid embedded schema test cases
- add invalid nested input/output schema fixtures

This PR intentionally avoids overlap with:
- #8662
- #8619
- #8538
- #8634

Test plan:
- ./mvnw -pl schema-util/json,schema-util/util-provider checkstyle:check
- ./mvnw -pl schema-util/util-provider -am test -Dtest=McpToolContentValidatorTest -Dsurefire.failIfNoSpecifiedTests=false

PR #8665, opened 2026-07-16
Fix : McpToolCompatibilityChecker inverted verdicts for FORWARD/FULL levels

Ady0333:
## Problem

McpToolCompatibilityChecker extends AbstractCompatibilityChecker, which builds FORWARD and FULL by swapping the arguments to isBackwardsCompatibleWith. That works for a directional checker like Avro or JSON Schema, but this checker's rules are structural, checkPropertyRemovals just diffs a fixed old vs new argument order. Swapping the arguments relabels additions as removals and hides real removals.

So in practice, FULL rejected adding an optional input parameter, even though the class doc says that's always compatible. And FORWARD silently accepted removing an input parameter, even though the doc says that's incompatible.

---

## Why it matters

MCP_TOOL is a first class artifact type. Any registry using FORWARD or FULL compatibility on MCP tool artifacts gets wrong governance, FULL blocks safe tool evolution, FORWARD lets breaking changes through silently until a tool caller breaks.

---

## What I changed

Overrode testCompatibility in McpToolCompatibilityChecker so every level evaluates the rules once in a fixed existing to proposed direction, instead of relying on the base class swapping arguments:

```java
switch (compatibilityLevel) {
    case BACKWARD_TRANSITIVE:
    case FORWARD_TRANSITIVE:
    case FULL_TRANSITIVE:
        for (TypedContent existing : existingArtifacts) {
            differences.addAll(isBackwardsCompatibleWith(existing.getContent().content(), proposed, resolvedReferences));
        }
        break;
    default:
        String latest = existingArtifacts.get(existingArtifacts.size() - 1).getContent().content();
        differences.addAll(isBackwardsCompatibleWith(latest, proposed, resolvedReferences));
        break;
}
```

isBackwardsCompatibleWith itself is unchanged.

---

## Tests

Added testFullCompatibilityAllowsAddingOptionalParameter and testForwardCompatibilityRejectsRemovingParameter to the existing test class. Full suite: 11/11 pass.


Ady0333:
Heads up, the License Compliance (FOSSA) check is failing here, but it's failing identically on `main` (as of the latest commit before this PR) and on several other currently open PRs (#8667, #8666, #8664, #8662), all with the same "10 issues found" error. This PR only modifies two existing Java files and adds no new dependencies, so it shouldn't be able to introduce a dependency-license issue. Flagging in case it needs a separate look, but it doesn't look related to this change.

carlesarnal (changes requested):
The fix is incomplete. You've added a method override but the logic doesn't actually fix the inverted verdicts. The implementation calls `isBackwardsCompatibleWith(existing, proposed)` for all compatibility levels, which only checks backward compatibility.

For FORWARD level, you need to swap the arguments: `isBackwardsCompatibleWith(proposed, existing)`. For FULL level, you need to check both directions and require both to pass.

The test cases pass because they're testing the wrong behavior. `testFullCompatibilityAllowsAddingOptionalParameter` should actually test BACKWARD compatibility, and `testForwardCompatibilityRejectsRemovingParameter` should test BACKWARD compatibility rejection.

Also fix the annotation placeholders in `McpToolCompatibilityChecker.java` - lines showing file paths like `@app/src/test/java/io/apicurio/registry/cors/CorsUrlOverrideTest.java` should be `@Override`.

Files to fix:
- `schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java` - Implement correct logic for FORWARD/FULL, fix annotations
- `schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java` - Add correct test cases for all compatibility levels

Ady0333:
Thankyou @carlesarnal for accepting the pr. I am working on the changes, will push them shortly.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -25,6 +28,44 @@ public class McpToolCompatibilityChecker
...
+        requireNonNull(compatibilityLevel, "compatibilityLevel MUST NOT be null");
+        requireNonNull(existingArtifacts, "existingArtifacts MUST NOT be null");
+        requireNonNull(proposedArtifact, "proposedArtifact MUST NOT be null");
+
+        if (compatibilityLevel == CompatibilityLevel.NONE || existingArtifacts.isEmpty()) {
+            return CompatibilityExecutionResult.compatible();
+        }
+
+        String proposed = proposedArtifact.getContent().content();
+        Set<CompatibilityDifference> differences = new HashSet<>();
```
paoloantinori, line 50: This transitive loop is the branch most likely to hide an ordering/dedup bug, but the two new tests only exercise it with a single existing artifact. Consider adding a `FULL_TRANSITIVE` test with two existing versions so the multi-version path is actually covered.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -25,6 +28,44 @@ public class McpToolCompatibilityChecker
...
+            case BACKWARD_TRANSITIVE:
+            case FORWARD_TRANSITIVE:
+            case FULL_TRANSITIVE:
+                for (TypedContent existing : existingArtifacts) {
+                    differences.addAll(
+                            isBackwardsCompatibleWith(existing.getContent().content(), proposed,
+                                    resolvedReferences));
+                }
+                break;
+            default:
```
paoloantinori, line 64: This duplicates the base class `get(size()-1)` latest selection and surrounding plumbing. Fine for now, but a follow-up to expose a directional opt-out hook on `AbstractCompatibilityChecker` would keep the shared logic in one place and avoid silent divergence if the base evolves.

paoloantinori (commented):
Thanks for the clear write-up — the diagnosis is correct. `McpToolCompatibilityChecker.isBackwardsCompatibleWith` is genuinely directional: `checkPropertyRemovals` only flags properties present in `existing` and absent in `proposed`, so when `AbstractCompatibilityChecker` swaps the args for `FORWARD`/`FULL` ([source](https://github.com/Apicurio/apicurio-registry/blob/main/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AbstractCompatibilityChecker.java)), additions get flagged as removals and real removals slip through. Overriding `testCompatibility` to always evaluate existing→proposed is the right call.

A few observations, none blocking:

- **Transitive path has no multi-version coverage.** The newly-exercised `BACKWARD_TRANSITIVE`/`FORWARD_TRANSITIVE`/`FULL_TRANSITIVE` loop in the override is the branch most likely to hide an ordering or dedup issue, but both new tests pass a single-element `existingArtifacts` list. A third test with two existing versions (e.g. an older version that had an extra param, plus a newer one that dropped it) under `FULL_TRANSITIVE` would lock in the intended semantics and guard against a future regression in how the latest is selected.
- **The override re-derives the base class structure** (the `requireNonNull` guards, the `NONE`/empty short-circuit, the `get(size()-1)` latest selection, the transitive/non-transitive split). About 35 lines duplicated from `AbstractCompatibilityChecker.testCompatibility`. It is correct today, but if the base later changes how it picks the latest version or handles the empty case, this override will silently diverge. Worth considering a more surgical hook — e.g. a `protected boolean isDirectional()` flag on the base class that structural checkers can opt out of — so the shared plumbing lives in one place. Not something to block this fix on, but a good follow-up.
- **Semantic note to confirm:** with this change `FORWARD`, `BACKWARD`, and `FULL` all collapse to the same evaluation (a single `isBackwardsCompatibleWith(existing, proposed)` call), since `FULL` was the union of the two and both halves are now identical. That matches the class javadoc claim that the rules are direction-agnostic, but it does mean users who set `FORWARD` expecting the traditional "new code tolerates old data" meaning will get back-compat semantics instead. If that is the intended product behavior, great — just flagging it so it is a conscious decision rather than an emergent one.

Tests look solid for the two regression cases called out in the PR description.

paoloantinori (commented):
Thanks for the follow-up — this head resolves the concern from the earlier review. Overriding `testCompatibility` so every level evaluates the rules in a fixed existing→proposed direction is the right fix: the structural rules in `isBackwardsCompatibleWith` (notably `checkPropertyRemovals`) only mean what the class doc claims when `existing` is the baseline, so the base class's argument swapping for FORWARD/FULL was indeed producing inverted verdicts. The three new tests cover the previously broken cases well — the FULL + optional-parameter case, the FORWARD + parameter-removal case, and the FULL_TRANSITIVE walk over prior versions are exactly the regressions I would have asked for.

A couple of minor observations, not blockers:
- Since the rules are direction-agnostic, `BACKWARD_TRANSITIVE`, `FORWARD_TRANSITIVE` and `FULL_TRANSITIVE` collapse to the same evaluation, and so do the non-transitive levels under `default`. The current `switch` is correct and readable; a short comment noting that the explicit case-by-case is intentional (or alternatively collapsing them with a clarifying note) would help future readers realise the symmetry is by design.
- The non-transitive branch assumes `existingArtifacts.get(size - 1)` is the most recent version. Worth a one-line comment confirming that ordering contract is guaranteed by `AbstractCompatibilityChecker`'s callers, so a future refactor of the base class doesn't silently break this assumption.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
paoloantinori, line 50: Since the structural rules are direction-agnostic, `BACKWARD_TRANSITIVE`, `FORWARD_TRANSITIVE` and `FULL_TRANSITIVE` all evaluate identically here (and the non-transitive levels collapse under `default`). The explicit fall-through is fine for readability; consider a one-line comment stating the levels are listed separately on purpose even though they collapse, so a future reader doesn't read the missing per-level logic as a bug.

Ady0333:
Thanks for confirming the direction and for the detailed trace, that helped
me double check my own reasoning too. i have added both comments you suggested,
the direction-agnostic note above the switch and the ordering-contract note
near the latest-version selection.

paoloantinori (commented):
Thanks @Ady0333 — the direction-agnostic comments and the additional test coverage look good. Waiting on maintainer review (PR is in `lifecycle/wip` — mark `/ready` when you're done).

carlesarnal:

PR #8666, opened 2026-07-16
fix(search): return 400 instead of 500 for content search without the index

### LINKED ISSUE #8582: Version search with content filter returns 500 (malformed SQL: empty AND () predicate) on SQL storage
## Description

**Registry Version**: snapshot (build from `main`)
**Persistence type**: sql (H2 in-memory default; same code path for PostgreSQL/MySQL/MSSQL)

`GET /apis/registry/v3/search/versions?content=<anything>` returns **HTTP 500** on any SQL-backed deployment without the optional Elasticsearch index (`apicurio.search.index.enabled=false`, the default).

Root cause: in `SqlSearchRepository`, the `content` filter type is delegated to the search index and skipped for SQL, but the skip is an empty `case content: break;`. The surrounding loop still wraps every filter in `(` … `)`, so a `content` filter emits an empty predicate `()`. The generated SQL becomes `... WHERE (1 = 1) AND () ORDER BY ...`, which the database rejects. The in-code comment ("Content search is handled by the search index only; skip in SQL") shows the intent is to *ignore* the filter, but the implementation produces invalid SQL instead of a no-op.

Affected: `SqlSearchRepository` version-search `case content:` (and the matching artifact-search case).

Secondary: the 500 body echoes the full failing SQL statement in `detail`, leaking internal query/schema structure.

### Environment

Apicurio Registry snapshot (`main`), in-memory (H2) persistence, default config (no ES index), Java 21 (Temurin), plain JVM run, no auth. `AND ()` is invalid on H2, PostgreSQL, MySQL, and SQL Server.

### Steps to Reproduce

1. Start the registry with defaults: `java -jar app/target/quarkus-app/quarkus-run.jar`
2. `curl -i "http://localhost:8080/apis/registry/v3/search/versions?content=hello"`

**Expected**: content search is index-only, so on SQL the `content` filter should be a clean no-op (return versions matching the other filters), or a well-formed 4xx/501, not a 500.

**Actual**: HTTP 500:
`JdbcSQLDataException: Data conversion error converting "ROW to BOOLEAN"; SQL statement: SELECT v.*, a.type FROM versions v JOIN artifacts a ON ... WHERE (1 = 1) AND () ORDER BY v.globalId asc LIMIT ? OFFSET ?`

### Logs

`[500 ERROR DETECTED] ... WHERE (1 = 1) AND () ... org.h2.jdbc.JdbcSQLDataException: Data conversion error converting "ROW to BOOLEAN"`

Me-Priyank:
Fixes #8582

## Problem

`GET /apis/registry/v3/search/versions?content=<x>` returned 500 on SQL storage with the Elasticsearch search index disabled (the default). The `content` filter is index-only; the SQL path had an empty `case content` branch that emitted a malformed predicate (`AND ()`), so the query failed and the 500 body leaked the SQL statement.

## Fix

Content search is only served by the search index, so the SQL path now throws a mapped `ContentSearchNotSupportedException` (400) with a clear message, matching what `ElasticsearchSearchDecorator` already reports when the index is unavailable. No SQL is leaked, and a 400 (not a 500) means the liveness check is not involved.

## Tests

`SearchVersionsTest`: a content filter without the index returns 400 and the response no longer contains the SQL statement.

## Scope

Scoped to the reported `content` defect. The sibling `structure` filter already returns a clean, non-leaking error via the default branch and is left unchanged.


Me-Priyank:
@carlesarnal @EricWittmann PTAL

carlesarnal (changes requested):
The fix correctly changes content search without index from 500 to 400. Error message is clear. Test verifies no SQL leakage. 

Issues:

1. Test annotation is malformed: `@utils/converter/src/test/java/io/apicurio/registry/utils/converter/avro/AvroDataTest.java` should be `@Test`. This looks like a diff corruption or copy-paste error. Fix before merge.

2. Missing test for `/search/artifacts` endpoint. The same fix applies to `searchArtifacts()` method but only `/search/versions` is tested. Add test for artifacts endpoint.

3. Consider testing content filter combined with other valid filters to ensure exception is thrown even when other filters are present.

File paths:
- `app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java` - both methods fixed
- `app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java` - annotation needs fix, add artifacts test

Me-Priyank:
> The fix correctly changes content search without index from 500 to 400. Error message is clear. Test verifies no SQL leakage.
> 
> Issues:
> 
> 1. Test annotation is malformed: `@utils/converter/src/test/java/io/apicurio/registry/utils/converter/avro/AvroDataTest.java` should be `@Test`. This looks like a diff corruption or copy-paste error. Fix before merge.
> 2. Missing test for `/search/artifacts` endpoint. The same fix applies to `searchArtifacts()` method but only `/search/versions` is tested. Add test for artifacts endpoint.
> 3. Consider testing content filter combined with other valid filters to ensure exception is thrown even when other filters are present.
> 
> File paths:
> 
> * `app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java` - both methods fixed
> * `app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java` - annotation needs fix, add artifacts test

Thanks @carlesarnal. Addressed point 3, added `testSearchVersionsByContentFilterWithOtherFiltersIsBadRequest`, which verifies a `content` filter combined with another filter (`groupId`) is still rejected with a 400, so the exception is thrown regardless of the other filters present.

On the other two:

1. `AvroDataTest.java` isn't part of this PR, it's not in the diff and is unchanged on `main` (its `@Test` annotations are all intact) so there's nothing to change there.

2. The `/search/artifacts` endpoint doesn't expose a `content` query parameter (only `/search/versions` does), so there's no artifacts endpoint that reaches this path. I kept the `searchArtifacts()` guard for consistency since both methods share the same filter switch, but it isn't reachable through the REST API. Happy to add a storage-level test or scope the change strictly to `/search/versions` if you'd prefer.

Pushed and ready for another look.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -32,6 +33,10 @@
  */
 public class SqlSearchRepository {
 
+    private static final String CONTENT_SEARCH_UNSUPPORTED_MESSAGE =
+            "Content search requires the Elasticsearch search index, which is not enabled. "
```
paoloantinori: Minor: the message names "Elasticsearch" as the search backend. This exception is thrown from the SQL storage path whenever the `content` filter is used without an index, so it might be more accurate to phrase it generically ("a search index") unless Elasticsearch is guaranteed to be the only supported backend. Worth confirming against the search-index abstraction so the response stays correct for all deployment types.

paoloantinori (commented):
Clean, well-scoped fix — the empty `case content` branch was emitting a malformed predicate (`AND ()`) that failed the query and leaked the SQL statement in a 500. Throwing a mapped `ContentSearchNotSupportedException` (400) with a non-leaking message, plus two tests, is the right call. One wording consideration below.

- The `not(containsStringIgnoringCase("select"))` assertion is a nice touch — it guards against future SQL-leakage regressions, not just the status code.
- Worth double-checking: the message names "Elasticsearch" explicitly. If v3 supports (or may support) other search-index backends, hardcoding the vendor name in a user-facing 400 could mislead operators running a different one. It may also be worth aligning the exact wording with what `ElasticsearchSearchDecorator` reports when the index is configured but unavailable, so clients see a consistent message whether the request is routed through the SQL path or the decorator.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
Me-Priyank: Thanks for flagging @paoloantinori ... the feature is toggled by the generic `apicurio.search.index.enabled` property (Elasticsearch is just the current backend implementation), so I've dropped the vendor name the message now reads "Content search requires the search index, which is not enabled." That keeps it accurate for any deployment and matches the config terminology.

EricWittmann (commented):
Nice fix — clean and well-scoped. One request:

`ContentSearchNotSupportedException` currently extends `RegistryException`, but every other exception in the `storage/error` package extends `RegistryStorageException`. Since this is thrown from a storage class and lives in that package, it should extend `RegistryStorageException` for consistency.

Me-Priyank:
Thanks @EricWittmann .. I initially switched it to `RegistryStorageException` as you suggested, but that adds a 6th parent to the hierarchy and trips SonarCloud's `java:S110` ("6 parents, max 5 authorized"), which failed the quality gate.

I've kept it on `RegistryException`, which actually matches the other 400-mapped input exceptions in this package, `InvalidContentException` (from the prompt-render PR), `InvalidArtifactTypeException`, `InvalidGroupIdException` all extend `RegistryException` directly. So it stays consistent with that family and passes S110. Happy to suppress S110 and switch to `RegistryStorageException` instead if you'd prefer, but this seemed cleaner than a suppression.

The 400 mapping is unaffected either way since `HttpStatusCodeMap` resolves by exact class. And thanks for splitting the `searchArtifacts` dead-code point into #8680, happy to pick that up separately.

EricWittmann:
> Thanks @EricWittmann .. I initially switched it to `RegistryStorageException` as you suggested, but that adds a 6th parent to the hierarchy and trips SonarCloud's `java:S110` ("6 parents, max 5 authorized"), which failed the quality gate.

OK that's cool.  I don't actually much care what SonarCloud thinks most of the time, but I don't feel strongly enough to fight with it.  :rofl:

PR #8672, opened 2026-07-16
docs(contracts): clarify artifact refs and fix ODCS projection scoping

### LINKED ISSUE #8104: Document and clarify how artifact references interact with data contracts
## Context

Data contracts (ODCS projection, CEL rules, PII tagging) are scoped to individual artifacts. When an artifact references other artifacts (e.g., an Avro schema referencing an `Address` record from a separate artifact), the contract system does not resolve or follow those cross-artifact references.

This is actually the correct architectural choice, but it's not documented and will surprise users.

## Current Behavior

**Rule evaluation works fine with nested data.** When a Kafka record is serialized, it's fully materialized — an Avro `GenericRecord` contains the actual nested `Address` object inline, not a reference. So the `Map<String, Object>` conversion produces:

```json
{
  "orderId": "ORD-123",
  "address": { "street": "123 Main St", "zipCode": "90210" }
}
```

A CEL rule like `address.zipCode != ""` already works — nested fields from referenced schemas are accessible at evaluation time.

**The gap is in contract projection and PII tagging.** The ODCS `schemas[].fields` section only lets you declare metadata/PII on fields from the directly referenced artifact. You can't declare `address.street` as PII from the `OrderEvent` contract because the projection engine doesn't resolve cross-artifact references to discover those fields.

## Why This Is the Right Design

Each schema should have its own contract declaring its own PII fields. The `Address` schema has a contract saying `street` is PII. The `OrderEvent` contract doesn't need to know about `Address` internals — that's `Address`'s concern. This follows the same ownership boundary that contracts are meant to enforce.

Resolving reference chains during projection would create coupling: changing the `Address` schema would require re-projecting every contract that references it. It also raises the question of whose contract "wins" if both declare conflicting metadata on the same field.

## What Needs to Happen

1. **Document this explicitly** — in the data contracts documentation (`assembly-data-contracts.adoc`), explain that contracts are per-artifact and cross-artifact references are not resolved. Clarify that CEL rules CAN access nested fields at evaluation time (records are fully materialized), but PII/metadata tagging is per-artifact.

2. **Add a note to the ODCS contract YAML schema** — explain that each `schemas[]` entry is independent, and if users want to govern referenced schemas, they should list them as separate entries in `schemas[]` or create separate contracts.

3. **Consider a validation warning** — when an ODCS contract declares `fields` metadata on a field path that doesn't exist in the schema (because it's in a referenced artifact), emit a warning rather than silently ignoring it.

## Related

- Epic: #6192
- ODCS projection engine: `app/src/main/java/io/apicurio/registry/contracts/odcs/`
- Tag extractors: `app/src/main/java/io/apicurio/registry/contracts/tags/`

niy-ati:
Fixes #8104

## Summary
- Document that each schemas[] entry is independent and projection does not follow cross-artifact schema references
- Fix OdcsTagProjector to apply field tags only from the matching schemas[] entry (review feedback on #8548)
- Add OdcsSchemaLocations helper for shared location parsing
- Add docs section, Javadoc, sample YAML comments, and OdcsTagProjectorTest

## Review feedback addressed
- Javadoc for location now reflects optional groupId and version/branch
- PII/CEL capitalization fixed in sample YAML comments
- Projector behavior now matches documented per-artifact scoping

## Test plan
- [x] OdcsTagProjectorTest passes locally

### DIFF
--- a/app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java
+++ b/app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java
@@ -102,7 +102,10 @@ public class OdcsTagProjector {
         String versionExpression = null;
         String location = schema.getLocation();
         if (location != null && location.contains(":")) {
-            versionExpression = location.substring(location.indexOf(':') + 1);
+            String afterColon = location.substring(location.indexOf(':') + 1);
+            if (!afterColon.isEmpty() && !afterColon.isBlank()) {
+                versionExpression = afterColon;
+            }
         }
 
         if (versionExpression != null && !versionExpression.isBlank()) {
--- a/app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java
+++ b/app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java
@@ -72,6 +72,9 @@ public class OdcsTagProjector {
     /**
      * Returns the {@code schemas[]} entry whose {@code location} resolves to the target artifact.
      * Group may be omitted in {@code location} (defaults to {@code groupId}).
+     * <p>
+     * If multiple entries resolve to the same {@code (groupId, artifactId)}, the first match in
+     * list order is used intentionally — duplicate locations should be avoided in contract YAML.
      */
     static OdcsSchema findMatchingSchema(List<OdcsSchema> schemas, String groupId, String artifactId) {
         if (schemas == null) {
--- a/app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java
+++ b/app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java
@@ -16,6 +16,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 
 @ApplicationScoped
 public class OdcsTagProjector {
@@ -34,7 +35,12 @@ public class OdcsTagProjector {
             return 0;
         }
         try {
-            String targetVersion = resolveTargetVersion(contract, groupId, artifactId);
+            OdcsSchema matchingSchema = findMatchingSchema(contract.getSchemas(), groupId, artifactId);
+            if (matchingSchema == null) {
+                return 0;
+            }
+
+            String targetVersion = resolveTargetVersion(matchingSchema, groupId, artifactId);
             if (targetVersion == null) {
                 warnings.add("No versions found for " + groupId + "/" + artifactId);
                 return 0;
@@ -44,11 +50,8 @@ public class OdcsTagProjector {
 
             var labels = new LinkedHashMap<String, String>();
             int count = 0;
-            for (OdcsSchema schema : contract.getSchemas()) {
-                if (schema.getFields() == null) {
-                    continue;
-                }
-                for (var entry : schema.getFields().entrySet()) {
+            if (matchingSchema.getFields() != null) {
+                for (var entry : matchingSchema.getFields().entrySet()) {
                     count += addTags(labels, tagPrefix, entry.getKey(),
                             entry.getValue(), warnings);
                 }
@@ -66,14 +69,37 @@ public class OdcsTagProjector {
         }
     }
 
-    private String resolveTargetVersion(OdcsContract contract, String groupId,
-            String artifactId) {
-        String versionExpression = null;
-        if (contract.getSchemas() != null && !contract.getSchemas().isEmpty()) {
-            String location = contract.getSchemas().get(0).getLocation();
-            if (location != null && location.contains(":")) {
-                versionExpression = location.substring(location.indexOf(':') + 1);
+    /**
+     * Returns the {@code schemas[]} entry whose {@code location} resolves to the target artifact.
+     * Group may be omitted in {@code location} (defaults to {@code groupId}).
+     */
+    static OdcsSchema findMatchingSchema(List<OdcsSchema> schemas, String groupId, String artifactId) {
+        if (schemas == null) {
+            return null;
+        }
+        for (OdcsSchema schema : schemas) {
+            if (matchesTarget(schema, groupId, artifactId)) {
+                return schema;
             }
+        }
+        return null;
+    }
+
+    static boolean matchesTarget(OdcsSchema schema, String groupId, String artifactId) {
+        if (schema == null) {
+            return false;
+        }
+        String[] parsed = OdcsSchemaLocations.parse(schema.getLocation(), groupId);
+        return OdcsSchemaLocations.isValid(parsed)
+                && Objects.equals(groupId, parsed[0])
+                && Objects.equals(artifactId, parsed[1]);
+    }
+
+    private String resolveTargetVersion(OdcsSchema schema, String groupId, String artifactId) {
+        String versionExpression = null;
+        String location = schema.getLocation();
+        if (location != null && location.contains(":")) {
+            versionExpression = location.substring(location.indexOf(':') + 1);
         }
 
         if (versionExpression != null && !versionExpression.isBlank()) {
--- a/app/src/test/java/io/apicurio/registry/contracts/odcs/OdcsTagProjectorTest.java
+++ b/app/src/test/java/io/apicurio/registry/contracts/odcs/OdcsTagProjectorTest.java
@@ -1,5 +1,7 @@
 package io.apicurio.registry.contracts.odcs;
 
+import io.apicurio.registry.model.BranchId;
+import io.apicurio.registry.model.GAV;
 import io.apicurio.registry.storage.RegistryStorage;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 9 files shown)


niy-ati:
hello @carlesarnal ,  all review feedback from the earlier PR is addressed :
- OdcsTagProjector now projects field tags only from the matching schemas[] entry
- OdcsSchemaLocations helper + OdcsTagProjectorTest added
- Docs, Javadoc, and sample YAML comments updated

OdcsTagProjectorTest passes locally.

app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java
```diff
@@ -34,7 +35,12 @@ public int project(OdcsContract contract, String contractId, String groupId,
             return 0;
         }
         try {
-            String targetVersion = resolveTargetVersion(contract, groupId, artifactId);
+            OdcsSchema matchingSchema = findMatchingSchema(contract.getSchemas(), groupId, artifactId);
```
paoloantinori, line 38: Nice — this is the core fix. Tagging only `matchingSchema.getFields()` (and resolving the version from `matchingSchema.getLocation()` at line 43 instead of `contract.getSchemas().get(0)`) stops fields from sibling schema entries leaking onto the target artifact. Worth noting in release notes that artifacts not listed in any `schemas[]` entry now get zero tags where they previously inherited all schemas' fields.

paoloantinori (commented):
Solid fix — scoping field-tag projection to the single matching `schemas[]` entry (and pulling the version expression from that same entry instead of `contract.getSchemas().get(0)`) closes a real cross-artifact bleed. A couple of notes below.

- **Behavioral change worth flagging:** `project()` now returns `0` and applies no tags when no `schemas[]` entry resolves to `(groupId, artifactId)`. Previously it iterated *every* schema and applied their fields (using `schemas.get(0)`'s version). The new outcome is correct, but contracts that happened to rely on the old bleed-through will emit fewer labels after this — worth a release-note line.
- **Version-resolution fix** (`OdcsTagProjector.java:43`): reading the location from `matchingSchema.getLocation()` instead of `contract.getSchemas().get(0).getLocation()` fixes version resolution for any artifact that is not the first schema entry. Good catch.
- **First-match semantics:** `findMatchingSchema` returns the first entry whose location resolves to the target. If two entries resolve to the same `(group, artifact)` (e.g. both omit the group), only the first contributes tags. Probably fine, but worth confirming it is intentional.
- **Test coverage:** the new `OdcsTagProjectorTest` exercises the static helpers (`OdcsSchemaLocations.parse`, `matchesTarget`, `findMatchingSchema`) well, but the headline behavior — *projecting artifact A yields only A's tags, not B's* — is not asserted through `project()` itself. A test that stubs the version lookup and asserts the resulting label set for a target artifact excludes another schema's fields would lock in the fix. The parallel result-projection path lives in [GroupsResourceImpl.projectOdcsContract](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java).
- The docs section (`assembly-data-contracts.adoc`) and the per-artifact scoping rationale are clearly written; the CEL-vs-projection distinction is well explained.

app/src/test/java/io/apicurio/registry/contracts/odcs/OdcsTagProjectorTest.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Ensures field-tag projection only uses the {@code schemas[]} entry for the target artifact (#8104 / PR review).
+ */
+class OdcsTagProjectorTest {
```
paoloantinori, line 31: The helper methods are well covered, but the actual projection scoping — the headline change of this PR — isn't exercised through `project()`. Consider adding a test that stubs the version/content lookup and asserts that, given two `schemas[]` entries, projecting artifact A yields only A's field tags and none of B's. That would guard the behavioral fix directly rather than only via `matchesTarget`/`findMatchingSchema`.

app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java
niy-ati: Thanks @paoloantinori , i agree on the behavioral change.

but that’s intentional: `project()` now returns 0 and applies no tags when no schemas[] entry resolves to the target (groupId, artifactId). I added an upgrade NOTE in `assembly-data-contracts.adoc` calling this out ,earlier releases bled tags from every `schemas[]` entry and contracts that relied on that may need a dedicated entry per governed artifact.

Version resolution now reads from `matchingSchema.getLocation()` instead of schemas[0], so non first entries resolve correctly too.

app/src/test/java/io/apicurio/registry/contracts/odcs/OdcsTagProjectorTest.java
niy-ati: Good point , added `project()` level tests with a mocked RegistryStorage:

- projecting orders/OrderEvent applies only orderId|PII, not Address’s street tag
- unmatched targets return 0 and never call mergeVersionLabels
- version "2" is taken from shared/Address:2, not from the first schemas[] entry

Also documented first match semantics on` findMatchingSchema` (duplicate locations -> first entry wins) with a test for that case.
All 7 OdcsTagProjectorTest cases pass locally.

niy-ati:
@carlesarnal , @paoloantinori , can u please see the changes made and mark them as accepted if the direction is aligned now.

paoloantinori:
Thanks — the direction is aligned with the earlier feedback. Pulling the location parser out into `OdcsSchemaLocations` (shared between the projector and `GroupsResourceImpl`), scoping projection to the single matching `schemas[]` entry, and reading the version expression from that same entry are all the right moves. The `OdcsTagProjectorTest` (first-match-on-duplicates, no-match-returns-zero, version-from-matching-entry) and the docs section on artifact references round it out. I can't mark it accepted (reviews only), but I have no further concerns from my side.

carlesarnal (dismissed):
Refactoring correctly fixes projection scoping issue. Each `schemas[]` entry now projects only onto its own artifact. Logic is sound, tests cover main paths, documentation is thorough.

The breaking change is properly documented with upgrade note. Teams with contracts that relied on cross-entry tag bleed-through need to add dedicated entries per artifact.

Minor: `resolveTargetVersion()` could validate extracted version expression isn't empty before the blank check, but current code is safe. Test coverage could add branch resolution and null-fields cases.

File paths checked:
- `app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsSchemaLocations.java` (new utility class)
- `app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java` (projection logic fixed)
- `app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java` (uses new utility)
- `app/src/test/java/io/apicurio/registry/contracts/odcs/OdcsTagProjectorTest.java` (comprehensive tests)
- `docs/modules/ROOT/pages/getting-started/assembly-data-contracts.adoc` (upgrade note and new section)

niy-ati:
Thanks @carlesarnal . addressed the minor points in the latest push:

- resolveTargetVersion() now ignores an empty/blank version suffix after ':' (falls back to latest version instead of treating it as an expression)
- Added project()-level tests for branch= resolution, null fields (zero tags, no merge), and trailing-colon locations

Together with the earlier review fixes (project() scoping tests, first-match semantics, upgrade note), OdcsTagProjectorTest is at 10 cases and all pass locally. Ready from my side whenever you can merge.

niy-ati:
@carlesarnal @paoloantinori ,please see the changes and approve them if they align correctly, and merge so that i can take up to other issues assigned to me after this pr closes

paoloantinori:
Thanks @niy-ati — took another pass over the latest push. The empty-suffix handling in `resolveTargetVersion` (falling back to latest instead of treating `""` as a version) and the shared `OdcsSchemaLocations` helper both look right, and the new tests cover the version/branch and empty-suffix paths cleanly. No further comments from my side. @carlesarnal also has eyes on this — merging is his call.

niy-ati:
Thanks @paoloantinori , I appreciate the second look and confirming the empty-suffix handling and test coverage look good.

paoloantinori:
Thanks @niy-ati — no further concerns from my side. The latest push (5609d95) keeps the empty-suffix handling, per-artifact scoping, and the extra `project()` test cases in good shape. Deferring to @carlesarnal for the final re-review and merge.

niy-ati:
@carlesarnal when you have a moment, could you please re-review the latest commit (5609d95) and approve/merge if aligned? Your earlier approval was dismissed after the follow-up fixes for empty version suffix, branch resolution, and the extra project() test cases.

paoloantinori (approved):
Looks good @niy-ati. The scoped projection via `findMatchingSchema` + `OdcsSchemaLocations` parser is correct — each `schemas[]` entry now projects only onto its own target artifact, which closes the cross-artifact tag bleed.

The empty/blank version-suffix handling (falling back to latest instead of treating `":"` as an explicit empty version) is the right call, and I see it's covered in the tests. The `OdcsSchemaLocations` parser is well-guarded (null, blank, multi-slash, colon-only inputs all return `INVALID`).

Documentation updates in `assembly-data-contracts.adoc` and the example contract are consistent with the behavioral change.

From my side this is ready to merge once CI is green.

niy-ati:
thanks @paoloantinori for the approval. @carlesarnal when you have a moment, could you re-approve and /accept so CI can run in a testable lifecycle state?

paoloantinori:
Already approved this one @niy-ati — nothing new to add. As mentioned before, @carlesarnal needs to re-approve so the CI lifecycle can advance.

paoloantinori:
I already approved this one — nothing new to add from my side. Deferring to @carlesarnal for the final re-approval and merge.

niy-ati:
> A maintainer can merge it with `/merge`, or enable auto-merge with `/auto-merge`

@EricWittmann , @carlesarnal  can u please /merge or /auto-merge if you find it right

PR #8637, opened 2026-07-15
ci(operator): publish multi-arch operator images (arm64, s390x, ppc64le)

aclfe:
Add an `image-buildx-push` target to `operator/Makefile` that builds and pushes the operator image with `docker buildx build --push --platform ...` instead of the single-arch `docker build`

Addresses #7665

```
  docker buildx build --builder apicurio-multiarch \
    --platform linux/amd64,linux/arm64,linux/s390x,linux/ppc64le \
    -f operator/controller/src/main/docker/Dockerfile.jvm \
    -o type=oci,dest=$S/operator-multiarch.tar $S/opctx
```
---
- [x] ./mvnw clean install -pl operator -am -DskipTests works
- [x] Verified `registry.access.redhat.com/ubi10/openjdk-21-runtime` publishes linux/amd64, arm64, s390x, and ppc64le variants
- [x] Ran `docker buildx build --platform linux/amd64,linux/arm64,linux/s390x,linux/ppc64le`

PR #8676, opened 2026-07-16
fix(kafkasql): retry topic config verification on UnknownTopicOrPartitionException

### LINKED ISSUE #8583: fix: KafkaSql tests flaky in CI — topic metadata propagation race in verifyTopicConfiguration()
## Problem

KafkaSql unit tests intermittently fail in CI with `java.lang.RuntimeException: Failed to start quarkus`, caused by `UnknownTopicOrPartitionException` during storage initialization. The tests pass locally and sometimes pass in CI (non-deterministic).

## Affected Tests

- `io.apicurio.registry.event.kafkasql.KafkaSqlEventsTest`
- `io.apicurio.registry.storage.impl.kafkasql.KafkaSqlRegistryStorageTest`
- `io.apicurio.registry.auth.KafkaSqlAuthTest`

All in the `app-kafkasql` unit test shard.

## Root Cause

**Kafka topic metadata propagation race condition** in `KafkaSqlRegistryStorage.initialize()`.

The initialization sequence at lines 169-176 of `KafkaSqlRegistryStorage.java`:

```java
// Lines 169-171: create topics
createTopicIfDoesNotExist(journalTopic, ...);
createTopicIfDoesNotExist(snapshotsTopic, ...);
createTopicIfDoesNotExist(eventsTopic, ...);

// Lines 174-176: immediately verify configuration
verifyTopicConfiguration(journalTopic, ...);
verifyTopicConfiguration(snapshotsTopic, ...);
verifyTopicConfiguration(eventsTopic, ...);
```

`verifyTopicConfiguration()` in `KafkaAdminUtil.java` (line 132) calls `adminClient.describeConfigs()`, which fails with `UnknownTopicOrPartitionException` because Kafka's `createTopics()` API is asynchronous: the controller accepts the creation request and returns success, but the topic metadata may not be propagated to brokers yet.

On GitHub Actions runners (2 vCPU, 7 GB RAM), the Strimzi/Testcontainers Kafka broker is CPU-starved, widening the metadata propagation window from milliseconds to hundreds of milliseconds — enough for `describeConfigs()` to fire before the topic is fully registered.

## CI Evidence

**PR #8442 run 29243593340** (2026-07-13):
```
KafkaSqlEventsTest -- Time elapsed: 24.42 s <<< ERROR!
java.lang.RuntimeException: java.lang.RuntimeException: Failed to start quarkus
Caused by: org.apache.kafka.common.errors.UnknownTopicOrPartitionException
```

**PR #8519 run 29089290766** (2026-07-10):
```
KafkaSqlRegistryStorageTest -- Time elapsed: 7.781 s <<< ERROR!
Run 1: KafkaSqlRegistryStorageTest » Runtime Failed to start quarkus
Run 2: KafkaSqlRegistryStorageTest » Runtime Failed to start quarkus
Run 3: KafkaSqlRegistryStorageTest » Runtime Failed to start quarkus
```

Three consecutive surefire retries all failed — the race window is consistent once the runner is under load.

## Why It's Intermittent

- **Locally**: Fast hardware closes the propagation window before `verifyTopicConfiguration()` runs.
- **CI (sometimes passes)**: If the runner isn't under heavy load, metadata propagates in time.
- **CI (fails)**: When multiple test shards or workflows run concurrently, CPU contention slows the broker, widening the race.
- `surefire.rerunFailingTestsCount=2` masks some occurrences, but not when the broker stays slow for the entire retry window.

## Proposed Fix

Add a retry with exponential backoff to `verifyTopicConfiguration()` in `KafkaAdminUtil.java`. The `describeConfigs()` call should catch `UnknownTopicOrPartitionException` and retry:

```java
// In KafkaAdminUtil.verifyTopicConfiguration()
int maxRetries = 3;
long backoffMs = 100;
for (int attempt = 0; attempt <= maxRetries; attempt++) {
    try {
        // existing describeConfigs() call
        break;
    } catch (ExecutionException e) {
        if (e.getCause() instanceof UnknownTopicOrPartitionException && attempt < maxRetries) {
            Thread.sleep(backoffMs);
            backoffMs *= 2;
            continue;
        }
        throw e;
    }
}
```

Alternatively, `createTopicIfDoesNotExist()` could confirm metadata availability via `adminClient.describeTopics()` before returning.

## Related Issues

- Not the same as the KafkaSQL snapshot-topic bootstrap busy-wait (#8495 context) — that concerns unbounded `while (assigned.isEmpty())` loops which happen *after* topic verification
- Not the same as operator Kafka test timeouts (PR #8511) — different layer (app unit tests vs operator integration tests on Minikube)

kartikangiras:
## Description

- `describeConfigs()` in `verifyTopicConfiguration()` fails intermittently in CI with UnknownTopicOrPartitionException because Kafka's createTopics() is async, topic metadata may not be propagated to brokers yet when the verify call fires. On resource-constrained CI runners this race window widens.

- Added exponential backoff retry (3 retries, 100ms → 400ms) to KafkaAdminUtil.verifyTopicConfiguration() to tolerate transient metadata unavailability.

### Issue
fixes #8583 

### Proof Manifests

- All the affected now pass succesfully.
Screenshot 2026-07-14 at 02 20 10
Screenshot 2026-07-14 at 02 20 50

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
@@ -138,7 +138,6 @@ public class KafkaAdminUtil {
         var key = new ConfigResource(ConfigResource.Type.TOPIC, topic);
         // includeSynonyms should ensure we get effective values (including default configs).
         var options = new DescribeConfigsOptions().includeSynonyms(true);
-
         blockOn(toJavaFuture(adminClient.get().get().describeConfigs(singleton(key), options).all())
                 .thenAccept(d -> {
                     var config = d.get(key);
--- a/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
@@ -8,6 +8,7 @@ import io.apicurio.registry.storage.impl.kafkasql.serde.KafkaSqlKeyDeserializer;
 import io.apicurio.registry.storage.impl.kafkasql.v2compat.BootstrapKey;
 import io.apicurio.registry.storage.impl.kafkasql.v2compat.UpgraderKey;
 import io.quarkus.arc.lookup.LookupIfProperty;
+import io.smallrye.faulttolerance.api.ExponentialBackoff;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Inject;
@@ -17,6 +18,7 @@ import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.config.TopicConfig;
 import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
+import org.eclipse.microprofile.faulttolerance.Retry;
 import org.slf4j.Logger;
 
 import java.util.List;
@@ -130,35 +132,15 @@ public class KafkaAdminUtil {
      *
      * @param topic
      */
+    @Retry(retryOn = UnknownTopicOrPartitionException.class, maxRetries = 3, delay = 100, jitter = 50)
+    @ExponentialBackoff
     public void verifyTopicConfiguration(String topic) {
         var key = new ConfigResource(ConfigResource.Type.TOPIC, topic);
         // includeSynonyms should ensure we get effective values (including default configs).
         var options = new DescribeConfigsOptions().includeSynonyms(true);
 
-        int maxRetries = 3;
-        long backoffMs = 100;
-        for (int attempt = 0; attempt <= maxRetries; attempt++) {
-            try {
-                blockOn(toJavaFuture(adminClient.get().get().describeConfigs(singleton(key), options).all())
-                        .thenAccept(configs -> assertTopicConfiguration(topic, key, configs)));
-                break;
-            } catch (UnknownTopicOrPartitionException e) {
-                if (attempt < maxRetries) {
-                    log.warn(
-                            "Topic '{}' configuration is not yet visible to Kafka metadata ({}). Retrying configuration verification in {} ms (attempt {}/{}).",
-                            topic, e.getMessage(), backoffMs, attempt + 1, maxRetries);
-                    try {
-                        Thread.sleep(backoffMs);
-                    } catch (InterruptedException ie) {
-                        Thread.currentThread().interrupt();
-                        throw new RuntimeException(ie);
-                    }
-                    backoffMs *= 2;
-                } else {
-                    throw e;
-                }
-            }
-        }
+        blockOn(toJavaFuture(adminClient.get().get().describeConfigs(singleton(key), options).all())
+                .thenAccept(configs -> assertTopicConfiguration(topic, key, configs)));
     }
 
     private void assertTopicConfiguration(String topic, ConfigResource key, Map<ConfigResource, Config> configs) {
--- a/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
@@ -140,28 +140,26 @@ public class KafkaAdminUtil {
         var options = new DescribeConfigsOptions().includeSynonyms(true);
 
         blockOn(toJavaFuture(adminClient.get().get().describeConfigs(singleton(key), options).all())
-                .thenAccept(configs -> assertTopicConfiguration(topic, key, configs)));
-    }
+                .thenAccept(d -> {
+                    var config = d.get(key);
+                    assertConfiguration(topic, config, TopicConfig.CLEANUP_POLICY_CONFIG, "delete"::equals, """
+                            Topic must not have '%s=%s'. While Apicurio Registry will work with topic compaction, it will not have any effect. \
+                            Use '%s=delete', '%s=-1', and '%s=-1' to disable automatic deletion of messages, and perform any cleanup manually after a backup has been made\
+                            """.formatted(TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_COMPACT,
+                            TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
 
-    private void assertTopicConfiguration(String topic, ConfigResource key, Map<ConfigResource, Config> configs) {
-        var config = configs.get(key);
-        assertConfiguration(topic, config, TopicConfig.CLEANUP_POLICY_CONFIG, "delete"::equals, """
-                Topic must not have '%s=%s'. While Apicurio Registry will work with topic compaction, it will not have any effect. \
-                Use '%s=delete', '%s=-1', and '%s=-1' to disable automatic deletion of messages, and perform any cleanup manually after a backup has been made\
-                """.formatted(TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_COMPACT,
-                TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
+                    if (!topic.equals(configuration.get().getEventsTopic())) { // Events topic is allowed to use retention. If configured, old events should be eventually deleted.
+                        if (!configuration.get().isTopicConfigurationVerificationOverrideEnabled()) {
+                            assertConfiguration(topic, config, TopicConfig.RETENTION_MS_CONFIG, "-1"::equals,
+                                    "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
+                                            .formatted(TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
+                        }
 
-        if (!topic.equals(configuration.get().getEventsTopic())) {
-            if (!configuration.get().isTopicConfigurationVerificationOverrideEnabled()) {
-                assertConfiguration(topic, config, TopicConfig.RETENTION_MS_CONFIG, "-1"::equals,
-                        "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
-                                .formatted(TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
-            }
-
-            assertConfiguration(topic, config, TopicConfig.RETENTION_BYTES_CONFIG, "-1"::equals,
-                    "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
-                            .formatted(TopicConfig.RETENTION_BYTES_CONFIG, TopicConfig.RETENTION_MS_CONFIG));
-        }
+                        assertConfiguration(topic, config, TopicConfig.RETENTION_BYTES_CONFIG, "-1"::equals,
+                                "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
+                                        .formatted(TopicConfig.RETENTION_BYTES_CONFIG, TopicConfig.RETENTION_MS_CONFIG));
+                    }
+                }));
     }
 
     private static void assertConfiguration(String topic, Config config, String property, Predicate<String> condition, String errorMessage) {
--- a/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
@@ -16,6 +16,7 @@ import org.apache.kafka.clients.admin.DescribeConfigsOptions;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.config.TopicConfig;
+import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
 import org.slf4j.Logger;
 
 import java.util.List;
@@ -133,27 +134,52 @@ public class KafkaAdminUtil {
         var key = new ConfigResource(ConfigResource.Type.TOPIC, topic);
         // includeSynonyms should ensure we get effective values (including default configs).
         var options = new DescribeConfigsOptions().includeSynonyms(true);
-        blockOn(toJavaFuture(adminClient.get().get().describeConfigs(singleton(key), options).all())
-                .thenAccept(d -> {
-                    var config = d.get(key);
-                    assertConfiguration(topic, config, TopicConfig.CLEANUP_POLICY_CONFIG, "delete"::equals, """
-                            Topic must not have '%s=%s'. While Apicurio Registry will work with topic compaction, it will not have any effect. \
-                            Use '%s=delete', '%s=-1', and '%s=-1' to disable automatic deletion of messages, and perform any cleanup manually after a backup has been made\
-                            """.formatted(TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_COMPACT,
-                            TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
 
-                    if (!topic.equals(configuration.get().getEventsTopic())) { // Events topic is allowed to use retention. If configured, old events should be eventually deleted.
-                        if (!configuration.get().isTopicConfigurationVerificationOverrideEnabled()) {
-                            assertConfiguration(topic, config, TopicConfig.RETENTION_MS_CONFIG, "-1"::equals,
-                                    "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
-                                            .formatted(TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
-                        }
-
-                        assertConfiguration(topic, config, TopicConfig.RETENTION_BYTES_CONFIG, "-1"::equals,
-                                "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
-                                        .formatted(TopicConfig.RETENTION_BYTES_CONFIG, TopicConfig.RETENTION_MS_CONFIG));
+        int maxRetries = 3;
+        long backoffMs = 100;
+        for (int attempt = 0; attempt <= maxRetries; attempt++) {
+            try {
+                blockOn(toJavaFuture(adminClient.get().get().describeConfigs(singleton(key), options).all())
+                        .thenAccept(configs -> assertTopicConfiguration(topic, key, configs)));
+                break;
+            } catch (UnknownTopicOrPartitionException e) {
+                if (attempt < maxRetries) {
+                    log.warn(
+                            "Topic '{}' configuration is not yet visible to Kafka metadata ({}). Retrying configuration verification in {} ms (attempt {}/{}).",
+                            topic, e.getMessage(), backoffMs, attempt + 1, maxRetries);
+                    try {
+                        Thread.sleep(backoffMs);
+                    } catch (InterruptedException ie) {
+                        Thread.currentThread().interrupt();
+                        throw new RuntimeException(ie);
                     }
-                }));
+                    backoffMs *= 2;
+                } else {
+                    throw e;
+                }
+            }
+        }
+    }
+
+    private void assertTopicConfiguration(String topic, ConfigResource key, Map<ConfigResource, Config> configs) {
+        var config = configs.get(key);
+        assertConfiguration(topic, config, TopicConfig.CLEANUP_POLICY_CONFIG, "delete"::equals, """
+                Topic must not have '%s=%s'. While Apicurio Registry will work with topic compaction, it will not have any effect. \
+                Use '%s=delete', '%s=-1', and '%s=-1' to disable automatic deletion of messages, and perform any cleanup manually after a backup has been made\
+                """.formatted(TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_COMPACT,
+                TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
+
+        if (!topic.equals(configuration.get().getEventsTopic())) {
+            if (!configuration.get().isTopicConfigurationVerificationOverrideEnabled()) {
+                assertConfiguration(topic, config, TopicConfig.RETENTION_MS_CONFIG, "-1"::equals,
+                        "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
+                                .formatted(TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
+            }
+
+            assertConfiguration(topic, config, TopicConfig.RETENTION_BYTES_CONFIG, "-1"::equals,
+                    "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
+                            .formatted(TopicConfig.RETENTION_BYTES_CONFIG, TopicConfig.RETENTION_MS_CONFIG));
+        }
     }
 
     private static void assertConfiguration(String topic, Config config, String property, Predicate<String> condition, String errorMessage) {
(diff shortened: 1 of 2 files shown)


kartikangiras:
Hi @paoloantinori @carlesarnal 
Could have a look at this PR, I was assigned the issue. The last PR was closed and was unable to open it due to some conflict issues. So opened a new one.

jsenko (commented):
The fix looks good — the retry logic is correct, the extracted `assertTopicConfiguration` method improves readability, and the test covers the retry scenario well. The maximum total retry sleep is ~700ms, so there's no risk of interfering with Kubernetes probes.

One suggestion: the codebase already has a couple of retry-with-backoff patterns that could be reused here instead of hand-rolling the loop:

**Option 1: `@Retry` annotation (simplest)**

`KafkaAdminUtil` is already `@ApplicationScoped`, so MicroProfile Fault Tolerance works out of the box — same pattern used by `AppAuthenticationMechanism` and `HttpClientService`:

```java
@Retry(retryOn = UnknownTopicOrPartitionException.class, maxRetries = 3, delay = 100, jitter = 50)
public void verifyTopicConfiguration(String topic) {
    // original code, no retry loop needed
}
```

**Option 2: Generic retry utility (more reusable)**

`ConnectionRetryUtil` in `storage/impl/sql/` already implements configurable exponential backoff. Its core loop could be extracted into a generic utility (e.g., a `RetryUtil.withRetry(Callable, retryPredicate, maxRetries, initialBackoffMs, multiplier)`) that both the SQL connection retry and this Kafka config verification call into.

The current inline approach works, but consolidating on one of these would make the retry behavior more consistent and maintainable across the codebase. Would you be interested in refactoring to one of these approaches?

kartikangiras:
Hi 

> The fix looks good — the retry logic is correct, the extracted `assertTopicConfiguration` method improves readability, and the test covers the retry scenario well. The maximum total retry sleep is ~700ms, so there's no risk of interfering with Kubernetes probes.
> 
> One suggestion: the codebase already has a couple of retry-with-backoff patterns that could be reused here instead of hand-rolling the loop:
> 
> **Option 1: `@Retry` annotation (simplest)**
> 
> `KafkaAdminUtil` is already `@ApplicationScoped`, so MicroProfile Fault Tolerance works out of the box — same pattern used by `AppAuthenticationMechanism` and `HttpClientService`:
> 
> ```java
> @Retry(retryOn = UnknownTopicOrPartitionException.class, maxRetries = 3, delay = 100, jitter = 50)
> public void verifyTopicConfiguration(String topic) {
>     // original code, no retry loop needed
> }
> ```
> 
> **Option 2: Generic retry utility (more reusable)**
> 
> `ConnectionRetryUtil` in `storage/impl/sql/` already implements configurable exponential backoff. Its core loop could be extracted into a generic utility (e.g., a `RetryUtil.withRetry(Callable, retryPredicate, maxRetries, initialBackoffMs, multiplier)`) that both the SQL connection retry and this Kafka config verification call into.
> 
> The current inline approach works, but consolidating on one of these would make the retry behavior more consistent and maintainable across the codebase. Would you be interested in refactoring to one of these approaches?

Hi @jsenko,
I have refactored to the first approach as it made more sense to me for this exact use case. As it is the exact pattern followed by many other use cases and also more maintainable.

jsenko (commented):
The latest push addresses the feedback well — the hand-rolled retry loop has been replaced with `@Retry(retryOn = UnknownTopicOrPartitionException.class, maxRetries = 3, delay = 100, jitter = 50)` plus `@ExponentialBackoff`, which is cleaner and consistent with how retry is handled elsewhere in the codebase (`AppAuthenticationMechanism`, `HttpClientService`).

The test was also updated appropriately — the success-path test verifies the assertion logic directly, and a separate annotation-presence test confirms `@Retry` is configured. The extracted `assertTopicConfiguration` method is still there, keeping things readable.

Looks good to me.

app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
```diff
@@ -129,31 +132,36 @@ public CompletableFuture<Boolean> createTopicIfDoesNotExistAsync(String topic, M
...
+                .thenAccept(configs -> assertTopicConfiguration(topic, key, configs)));
+    }
 
-                    if (!topic.equals(configuration.get().getEventsTopic())) { // Events topic is allowed to use retention. If configured, old events should be eventually deleted.
-                        if (!configuration.get().isTopicConfigurationVerificationOverrideEnabled()) {
-                            assertConfiguration(topic, config, TopicConfig.RETENTION_MS_CONFIG, "-1"::equals,
-                                    "Topic must have '%s=-1' and '%s=-1' to prevent accidental loss of data"
-                                            .formatted(TopicConfig.RETENTION_MS_CONFIG, TopicConfig.RETENTION_BYTES_CONFIG));
-                        }
+    private void assertTopicConfiguration(String topic, ConfigResource key, Map<ConfigResource, Config> configs) {
```
jsenko: Can you restore the original code structure to before it was moved to a separate method? Also this comment is missing:
```
if (!topic.equals(configuration.get().getEventsTopic())) { // Events topic is allowed to use retention. If configured, old events should be eventually deleted.
```
I assume it was done as part of the previous approach, but it's best to keep the original structure unless it's being deliberately changed. This also helps with reviewers to see what has changed. Otherwise LGTM, thanks!

jsenko (changes requested):

app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java
kartikangiras: I have restored the original code structure along with the comment. I guess it's GTG now.

paoloantinori:
Thanks for the new PR. The `@Retry(retryOn = UnknownTopicOrPartitionException.class, ...)` + `@ExponentialBackoff` on `verifyTopicConfiguration` is a cleaner fit than the hand-rolled loop, and the reflection-based test confirms the annotation is present. One thing worth double-checking: MicroProfile Fault Tolerance interceptors only apply when the method is invoked through the CDI proxy — a self-call (`this.verifyTopicConfiguration(...)`) would bypass `@Retry` entirely. Worth confirming the call site reaches `verifyTopicConfiguration` via the injected bean rather than a direct/internal invocation, so the retry actually fires. @jsenko already has an open change request on this one, so I'll defer the rest to that thread.

kartikangiras:
> Thanks for the new PR. The `@Retry(retryOn = UnknownTopicOrPartitionException.class, ...)` + `@ExponentialBackoff` on `verifyTopicConfiguration` is a cleaner fit than the hand-rolled loop, and the reflection-based test confirms the annotation is present. One thing worth double-checking: MicroProfile Fault Tolerance interceptors only apply when the method is invoked through the CDI proxy — a self-call (`this.verifyTopicConfiguration(...)`) would bypass `@Retry` entirely. Worth confirming the call site reaches `verifyTopicConfiguration` via the injected bean rather than a direct/internal invocation, so the retry actually fires. @jsenko already has an open change request on this one, so I'll defer the rest to that thread.

@paoloantinori, I agree with you on the point but I think it doesn't apply here as verifyTopicConfiguration is called via the injected kafkaAdmin CDI proxy from a different bean, not this self call, so `@Retry` fires.

paoloantinori:
Thanks for iterating. The `@Retry(retryOn = UnknownTopicOrPartitionException.class, ...)` + `@ExponentialBackoff` replacement for the hand-rolled loop looks good, and the reflection-based annotation assertion plus the success-path `describeConfigs` test are reasonable coverage given the method is exercised end-to-end elsewhere.

On the one open question I raised earlier — whether `verifyTopicConfiguration` is ever reached via a `this.` self-invocation that would bypass the MicroProfile Fault Tolerance CDI proxy: I can't confirm it from the diff alone, but the method is `public` and the callers I'd expect (topic bootstrap/verification) are external to the util class, so the risk looks low. If you can point at the call site (or confirm `KafkaAdminUtil` is always invoked as a CDI bean, not via `new`), that would fully close it out — otherwise I'm comfortable leaving it as a noted follow-up rather than a blocker.

carlesarnal (changes requested):
Cannot approve due to corrupted annotations in the diff. The code shows file paths where annotation types should be (`@app/src/main/java/...` instead of `@Retry`). This won't compile.

If the diff is just corrupted but the actual code is correct:

1. Remove the reflection-based annotation test in `KafkaAdminUtilTest.testVerifyTopicConfigurationHasRetryAnnotation()`. It tests implementation details, not behavior.
2. Add a real retry test: mock `describeConfigs` to throw `UnknownTopicOrPartitionException` twice, succeed on third attempt, verify it works.
3. Test exhausted retries: throw the exception 4 times, verify it propagates.
4. Test that other exceptions don't trigger retry.
5. Consider using `ConnectionRetryConfig` values instead of hardcoding retry parameters in `KafkaAdminUtil.verifyTopicConfiguration()`, or explain why you're not using it.

The retry logic itself is sound for the stated problem (topic not yet visible after creation), but the test coverage is inadequate.

kartikangiras:
@carlesarnal 
I have made a few changes to the Test coverage:

Added two new unit tests in KafkaAdminUtilTest
- testVerifyTopicConfigurationCallsDescribeConfigs: Mocks the Admin client and verifies describeConfigs is called exactly once for a healthy topic.
- testVerifyTopicConfigurationHasRetryAnnotation: Uses reflection to assert the `@Retry annotation` is present.

Also for the hardcoded retry parameters, ConnectionRetryConfig is scoped to database connectivity and isn't relevant to Kafka metadata propagation. So, the hardcoded values are the best and only way to propagate the behaviour.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaAdminUtilTest.java
aclfe, line 1: Question, this is a plain JUnit test with new KafkaAdminUtil() but isn't the retryOnUnknownTopic helper doing the retrying, not the annotation, so these would still pass even with `@Retry` removed at KafkaAdminUtil? 

Because if we want to prove the annotation fires, would we need a `@QuarkusTest`? Maybe we could just assert that [verifyTopicConfiguration](https://github.com/Apicurio/apicurio-registry/blob/fb1bd0803f84db111c0e8949f4215bc7d8e20883/app/src/main/java/io/apicurio/registry/storage/impl/util/KafkaAdminUtil.java#L135-L137) throws [UnknownTopicOrPartitionException](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/errors/UnknownTopicOrPartitionException.java) directly

aclfe (commented):
Hello, the fix looks right and should stop the flakiness (currently https://github.com/Apicurio/apicurio-registry/pull/8637 is blocked for a while so I'm keeping track).

Maybe a full `@QuarkusTest` exercising the annotation end to end is what the maintainers would prefer, since it checks the behavior more thoroughly.

paoloantinori:
@aclfe You're raising a valid point. As written, these tests don't actually exercise the `@Retry` interceptor — the retry logic is reproduced in the hand-rolled `retryOnUnknownTopic(Runnable, int)` helper, which mirrors the declared `maxRetries = 3` + `retryOn = UnknownTopicOrPartitionException.class` semantics. So yes, the tests would still pass if `@Retry` were removed from `KafkaAdminUtil.verifyTopicConfiguration`, because the helper is what's driving the retry loop, not the CDI interceptor.

That means the tests are essentially validating the mock wiring (does `describeConfigs` get called the expected number of times, do non-retryable exceptions propagate, does the success path terminate), not the production retry behavior. To actually prove the annotation fires you'd need a managed bean: either a `@QuarkusTest` that injects `KafkaAdminUtil` (so the interceptor is in the chain) with a mocked `Admin`, or a dedicated `@RetryingTest`/fault-tolerance profile. The plain-JUnit approach can't see the interceptor.

Two practical options:
- Keep these tests but rename them to make the scope explicit (e.g., `testVerifyTopicConfiguration_describeConfigsCalledNTimesUnderSimulatedRetry`) so it's clear they verify the contract, not the annotation. That's a fair characterization and still has value.
- Add at least one `@QuarkusTest` that exercises the real interceptor to close the gap aclfe pointed out.

I'd lean toward the second for at least the retry-then-succeed case — that's the path that matters in production — but I'll defer to @kartikangiras and @carlesarnal on whether it's worth the extra Quarkus startup cost here.

paoloantinori:
@aclfe raises a valid point. We filed #8583 (the underlying issue), so adding some context:

## The test doesn't exercise `@Retry`

The test is plain JUnit — `new KafkaAdminUtil()` bypasses CDI entirely. MicroProfile Fault Tolerance annotations (`@Retry`, `@CircuitBreaker`, etc.) are CDI interceptor bindings: they only fire when the bean is container-managed and injected. On a `new`-constructed instance, the `@Retry` annotation is inert.

The test's `retryOnUnknownTopic()` helper manually reimplements the retry loop, so all four test methods pass regardless of whether `@Retry` is present on the production method. You could remove `@Retry` and `@ExponentialBackoff` from `KafkaAdminUtil` and every test would still be green.

The codebase itself documents this CDI requirement — `OidcAuthenticationStrategy.java` has: `@param parent the parent CDI bean (hosts getAccessToken for @Retry support)`.

## Two options

**Option A (pragmatic):** Keep the tests as unit tests of the exception propagation behavior — they prove that `verifyTopicConfiguration()` throws `UnknownTopicOrPartitionException` when metadata hasn't propagated, which is the precondition for `@Retry` to help. Reframe the test names and Javadoc accordingly: they validate the *retry strategy* (that this specific exception type is retryable), not the annotation mechanism. `@Retry` itself is a framework guarantee — SmallRye/Quarkus ensures it works when CDI is active.

**Option B (thorough):** Write a `@QuarkusTest` with a `@TestProfile` that sets `apicurio.storage.kind=kafkasql`, inject `KafkaAdminUtil` as a CDI bean, and mock the Kafka `Admin` client to throw `UnknownTopicOrPartitionException`. This would genuinely exercise the interceptor. However, it requires more test infrastructure and may be disproportionate for a one-line annotation.

For context: the two other `@Retry` usages in the codebase (`AppAuthenticationMechanism.getAccessToken()` and `HttpClientService.post()`) have no dedicated retry-behavior tests either — they're covered implicitly through integration tests where CDI is active.

Either approach is acceptable, but the current test framing ("faithfully reproduces the declared retry semantics") should be adjusted to reflect what's actually being tested.

kartikangiras:
Hi @paoloantinori ,
I iterated over your feedback and made the changes to the test coverage with changing the names and updating the relevant javadoc. Now the `kafkaadminutiltest` replicate the retry and exception behaviour instead of annotation mechanism.

paoloantinori:
Thanks for iterating — the new test layout addresses the concern cleanly. The split into `testVerifyTopicConfigurationSucceedsImmediately` (happy path, 1 call), `...ThrowsUnknownTopicExceptionWhenMetadataUnavailable` (recover on the 3rd attempt, 3 calls), `...PropagatesAfterRetryBudgetExhausted` (1 initial + 3 retries = 4 calls, then propagate), and `...PropagatesNonRetryableExceptionImmediately` (`TopicExistsException`, 1 call) covers the four meaningful branches of the `@Retry` contract.

A couple of small notes worth a sanity check:

- The javadoc on `retryOnUnknownTopic` is honest about why the helper exists (`@Retry` is a CDI interceptor binding and is inert under `new KafkaAdminUtil()`), which is the right framing. If you want belt-and-suspenders, a one-line comment at the top of the test class pointing readers at the helper would help — but not blocking.
- Worth double-checking that `KafkaFuture.thenApply(...).get()` actually surfaces the underlying `UnknownTopicOrPartitionException` unwrapped (vs. wrapped in `ExecutionException`) in the way `verifyTopicConfiguration` consumes it. If the production code already unwraps `ExecutionException`'s cause, these assertions hold; if not, the `assertThrows(...)` cases may need to expect the wrapper. The "tests pass" screenshots suggest this is fine, but it's the one assumption worth eyeballing in the production method.

From my side, no further changes requested — once CI is green and the unwrap check above is confirmed, this is good to go.

jsenko:
@carlesarnal @paoloantinori — Looking for your input on the test coverage question here.

After reviewing the discussion thread, I'm **leaning towards removing the tests entirely** and keeping just the 2-annotation production fix. Here's my reasoning:

1. **It tests the framework, not our code.** `@Retry` is a SmallRye/Quarkus CDI interceptor — its behavior is a framework guarantee. The two other `@Retry` usages in the codebase (`AppAuthenticationMechanism`, `HttpClientService`) have zero dedicated retry tests.
2. **`KafkaAdminUtil` is obviously a CDI bean** (`@ApplicationScoped`, injected everywhere via CDI) — there's no realistic risk of the annotation being bypassed via self-call.
3. **No dangerous failure mode.** If it didn't work, the flaky test would stay flaky — there's no data loss or silent corruption. We'd notice immediately.
4. **CI stability is the real test.** If the flaky tests stabilize after merge, the fix works. That's more meaningful than any unit test we could write here.

As aclfe and paoloantinori correctly pointed out, the current tests don't actually exercise `@Retry` — they reimplement the retry loop in a test helper, so they'd pass even if the annotation were removed. They document the exception contract, but that gives a false sense of coverage.

### If we did want to properly test the retry

We'd need a `@QuarkusTest` with CDI active, plus a way to make `describeConfigs` throw `UnknownTopicOrPartitionException`. A few options:

- **`@Alternative` mock bean:** A `@TestProfile` that provides a mock `Admin` client via `@Alternative`/`getEnabledAlternatives()`, similar to the `HstsBoomResource` pattern in #8604.
- **Test-only code path in `KafkaAdminUtil`:** A configuration flag or injectable hook that allows forcing an exception during tests.
- **CDI interceptor:** A test-scoped interceptor that intercepts `verifyTopicConfiguration` and throws on the first N calls.

All of these add significant test infrastructure for a one-line annotation. I don't think it's worth it here.

What do you think — are you OK with dropping the tests and just keeping the production fix?

---

@kartikangiras — Thanks for your patience iterating on this. We're just working through the test coverage question among the maintainers. The production fix itself (the `@Retry` + `@ExponentialBackoff` annotations) looks good.

kartikangiras:
> @carlesarnal @paoloantinori — Looking for your input on the test coverage question here.
> 
> After reviewing the discussion thread, I'm **leaning towards removing the tests entirely** and keeping just the 2-annotation production fix. Here's my reasoning:
> 
> 1. **It tests the framework, not our code.** `@Retry` is a SmallRye/Quarkus CDI interceptor — its behavior is a framework guarantee. The two other `@Retry` usages in the codebase (`AppAuthenticationMechanism`, `HttpClientService`) have zero dedicated retry tests.
> 2. **`KafkaAdminUtil` is obviously a CDI bean** (`@ApplicationScoped`, injected everywhere via CDI) — there's no realistic risk of the annotation being bypassed via self-call.
> 3. **No dangerous failure mode.** If it didn't work, the flaky test would stay flaky — there's no data loss or silent corruption. We'd notice immediately.
> 4. **CI stability is the real test.** If the flaky tests stabilize after merge, the fix works. That's more meaningful than any unit test we could write here.
> 
> As aclfe and paoloantinori correctly pointed out, the current tests don't actually exercise `@Retry` — they reimplement the retry loop in a test helper, so they'd pass even if the annotation were removed. They document the exception contract, but that gives a false sense of coverage.
> 
> ### If we did want to properly test the retry
> We'd need a `@QuarkusTest` with CDI active, plus a way to make `describeConfigs` throw `UnknownTopicOrPartitionException`. A few options:
> 
> * **`@Alternative` mock bean:** A `@TestProfile` that provides a mock `Admin` client via `@Alternative`/`getEnabledAlternatives()`, similar to the `HstsBoomResource` pattern in [fix(security): add missing HSTS headers to error pages (#2411) #8604](https://github.com/Apicurio/apicurio-registry/pull/8604).
> * **Test-only code path in `KafkaAdminUtil`:** A configuration flag or injectable hook that allows forcing an exception during tests.
> * **CDI interceptor:** A test-scoped interceptor that intercepts `verifyTopicConfiguration` and throws on the first N calls.
> 
> All of these add significant test infrastructure for a one-line annotation. I don't think it's worth it here.
> 
> What do you think — are you OK with dropping the tests and just keeping the production fix?
> 
> @kartikangiras — Thanks for your patience iterating on this. We're just working through the test coverage question among the maintainers. The production fix itself (the `@Retry` + `@ExponentialBackoff` annotations) looks good.

Agreed. The tests are really not providing real test coverage and it would be better to avoid the unnecessary extra load for 1-line notations.

paoloantinori:
Agreeing with the direction here, with one nuance.

I was the one who flagged that the original tests claimed to exercise `@Retry` while actually replaying the retry loop in a hand-rolled helper — that concern has been fixed by the clearer names/javadoc, but jsenko's point takes it one step further and I think it's the right call: the `retryOnUnknownTopic(Runnable, int)` helper literally re-implements the `@Retry` contract in the test, so those tests assert the helper, not the framework. They will drift silently if the annotation changes.

My suggestion is a middle ground rather than "remove all tests": keep the tests that assert a *contract of our code*, drop the ones that only re-assert the framework.

**Keep (they verify something `KafkaAdminUtil` is actually responsible for):**
- `testVerifyTopicConfigurationSucceedsImmediately` — happy path, `describeConfigs` is reached once. This is the only test that actually exercises the production method end-to-end against a mocked Admin.
- A single "propagates `UnknownTopicOrPartitionException` when metadata is unavailable" test — this proves the *precondition* that makes `@Retry(retryOn = ...)` effective: the exception must escape `verifyTopicConfiguration` unwrapped. That is a real bug class (someone catching and logging inside the method would silently break retry), and it's worth one test.
- `testVerifyTopicConfigurationPropagatesNonRetryableExceptionImmediately` — confirms the method does not swallow arbitrary exceptions; useful as a guard against over-broad `catch (Exception)`.

**Drop (they duplicate what `@Retry` guarantees):**
- `testVerifyTopicConfigurationThrowsUnknownTopicExceptionWhenMetadataUnavailable` as currently written (3 calls, then success via the helper) — the "eventually succeeds after N retries" behavior is the framework's job.
- `testVerifyTopicConfigurationPropagatesAfterRetryBudgetExhausted` — asserts the framework's `maxRetries = 3` budget, including the exact "4 calls" count. That count is a SmallRye detail, not our contract; if someone bumps `maxRetries` the test breaks for no real reason.
- The `retryOnUnknownTopic` helper itself — once the two tests above are gone, nothing uses it.

Net: ~3 tests instead of 4, no hand-rolled retry loop in the test, and the production `@Retry` + `@ExponentialBackoff` two-liner stays exactly as jsenko approved it.

One caveat worth flagging on the "remove everything" option: the two existing `@Retry` usages jsenko references (`AppAuthenticationMechanism`, `HttpClientService`) live behind CDI in production, so their retry *does* fire — the asymmetry here is that `KafkaAdminUtil.verifyTopicConfiguration` is also called from a CDI bean, so the interceptor will fire in production too. The argument for keeping *some* test isn't "we don't trust `@Retry`", it's "we want a regression guard that the method keeps throwing the right exception type" — which is what the framework keys on.

kartikangiras:
@paoloantinori @jsenko 
I have reduced the tests from 4 to just 3 and removed the explicit test which was only used for annotation validation. Also dropped the retry helper.

Along with this I have resolved the wrapping method and now the test directly calls them.

paoloantinori:
@kartikangiras Thanks for the iteration — this layout looks good to me.

Dropping the hand-rolled retry helper and the dedicated annotation-validation test, and keeping the three remaining tests as **contract guards** (happy path, exception propagation so the CDI interceptor can see `UnknownTopicOrPartitionException`, and non-retryable exception passing through), is the right call. The javadocs are now honest about what these verify versus what `@Retry` does in production.

From my side: no further blockers. @jsenko @carlesarnal feel free to take it over the line.

paoloantinori:
@jsenko On reflection, I'm fine with dropping the tests entirely.

The one argument for keeping the exception-propagation test was as a guard against someone swallowing `UnknownTopicOrPartitionException` inside the method — that would silently break `@Retry` because the interceptor never sees the exception. But realistically, the method is 15 lines and the intent is clear. And if someone did break that precondition, the flaky test would come right back and point at the change — it's its own canary.

So: +1 to just the two annotations, no tests. Let's land it.

kartikangiras:
> @jsenko On reflection, I'm fine with dropping the tests entirely.
> 
> The one argument for keeping the exception-propagation test was as a guard against someone swallowing `UnknownTopicOrPartitionException` inside the method — that would silently break `@Retry` because the interceptor never sees the exception. But realistically, the method is 15 lines and the intent is clear. And if someone did break that precondition, the flaky test would come right back and point at the change — it's its own canary.
> 
> So: +1 to just the two annotations, no tests. Let's land it.

@paoloantinori Applied, dropped all the tests.

paoloantinori:
@kartikangiras Thanks for the ping — as I noted above, I'm aligned with dropping the tests entirely. The retry annotations on `verifyTopicConfiguration` are self-documenting and the happy-path contract is clear at 15 lines. No further concerns from me.

jsenko (approved):

jsenko:
Waiting for the tests and it's good to go, thanks!

aclfe:
Just saw the changes, everything looks great!

jsenko:
Thanks @kartikangiras !

PR #8681, opened 2026-07-16
fix: number input fields reject zero value in Prompt Template test panel

### LINKED ISSUE #8678: Bug: Number input fields reject zero value in Prompt Template test panel
#### Describe the bug

In `PromptTemplateTestPanel.tsx`, the number input field for template variables has two interconnected bugs that prevent users from entering zero:

**1. Display bug (line 123):** The `value` prop uses the `||` operator:
` ` `typescript
value={values[name] || ""}
` ` `
When `values[name]` is `0`, `0 || ""` evaluates to `""` because `0` is falsy in JavaScript. The input field displays empty even when the state correctly holds `0`.

**2. Setter bug (line 124):** The `onChange` handler also uses `||`:
` ` `typescript
onChange={(_event, val) => setValue(name, type === "integer" ? parseInt(val) || "" : parseFloat(val) || "")}
` ` `
`parseInt("0")` returns `0`, and `0 || ""` evaluates to `""`. The state is set to empty string instead of `0`.

Both bugs stem from the same root cause: using `||` (logical OR) instead of `??` (nullish coalescing) or explicit `isNaN()` checks with falsy numeric values.

#### Steps to reproduce

1. Create a PROMPT_TEMPLATE artifact with a numeric variable (e.g., `temperature: { type: "number", default: 0.7 }`)
2. Navigate to the Documentation tab → click "Test Prompt"
3. Clear the number field and type `0`
4. Observe: the field clears itself, `0` cannot be entered

#### Expected behavior

- `0` should be accepted and displayed as a valid numeric value
- Only `NaN` results (from non-numeric input like `"abc"`) should clear the field

#### Environment

- File: `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx`, lines 123-124

KumarADITHYA123:
fixes #8678 

### What this PR does

Fixes a bug where users cannot enter `0` as a value for numeric template variables in the Prompt Template test panel.

#### Root cause

Both the `value` prop and the `onChange` handler use the `||` (logical OR) operator for fallback. Since `0` is falsy in JavaScript, `0 || ""` evaluates to `""`, causing zero to be rejected as both a display value and a state value.

#### Fix

Replace `||` with `??` (nullish coalescing) for the display value, and use explicit `isNaN()` for the setter:

```diff
 case "integer":
 case "number":
     return (
         <TextInput
             type="number"
-            value={values[name] || ""}
-            onChange={(_event, val) => setValue(name, type === "integer" ? parseInt(val) || "" : parseFloat(val) || "")}
+            value={values[name] ?? ""}
+            onChange={(_event, val) => {
+                const n = type === "integer" ? parseInt(val) : parseFloat(val);
+                setValue(name, isNaN(n) ? "" : n);
+            }}
             aria-label={name}
         />
     );
```

#### Input trace verification

| Input | parseInt | isNaN? | setValue | value display (`?? ""`) |
|-------|---------|--------|---------|------------------------|
| `"0"` | `0` | `false` | `0` | `0` ✅ |
| `"42"` | `42` | `false` | `42` | `42` ✅ |
| `""` | `NaN` | `true` | `""` | `""` ✅ |
| `"abc"` | `NaN` | `true` | `""` | `""` ✅ |
| `"-5"` | `-5` | `false` | `-5` | `-5` ✅ |
| `"3.5"` (integer) | `3` | `false` | `3` | `3` ✅ |
| `"3.5"` (number) | `3.5` | `false` | `3.5` | `3.5` ✅ |

#### Files changed

- `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx`

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -120,8 +120,11 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                 return (
                     <TextInput
                         type="number"
-                        value={values[name] || ""}
-                        onChange={(_event, val) => setValue(name, type === "integer" ? parseInt(val) || "" : parseFloat(val) || "")}
+                        value={values[name] ?? ""}
+                        onChange={(_event, val) => {
+                            const n = type === "integer" ? parseInt(val) : parseFloat(val);
+                            setValue(name, isNaN(n) ? "" : n);
+                        }}
                         aria-label={name}
                     />
                 );


KumarADITHYA123:
@EricWittmann @carlesarnal @paoloantinori ptal!
When you are free?

paoloantinori:
Thanks for the ping. The fix is correct: `0 || ""` evaluates to `""` (the bug), and switching the display value to `??` plus the explicit `isNaN()` guard for the setter means `0` is preserved while empty/non-numeric input still clears to `""`. The input-trace table in the PR description checks out. Clean, minimal change — no concerns from me. Deferring to @carlesarnal / @EricWittmann for the merge.

KumarADITHYA123:
Thanks @paoloantinori for the review! 🙏

paoloantinori:
Happy to help — the `??` + `isNaN()` fix is correct and the truth table in the PR description makes the behavior easy to verify. Looks good to me from my side.

carlesarnal (changes requested):
The fix correctly addresses the zero value bug. Original code used `||` which treats `0` as falsy. New code uses `??` to only default on null/undefined, and properly handles NaN with explicit check. The logic is sound.

Missing test coverage for this bug. Add a test that verifies zero values are accepted for both integer and float inputs in `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx`.

KumarADITHYA123:
Thanks for the review! I looked into adding tests but the UI app currently has no test framework set up — no Jest/Vitest, no React Testing Library, no test scripts in package.json, and zero existing test files in ui-app/src.

Setting up the test infrastructure (installing vitest + @testing-library/react, adding config, adding npm scripts) would be a significantly larger change than the 2-line fix itself. Would you prefer:

1. I set up vitest + testing-library in this same PR and add the test
2. I open a separate issue/PR for test infrastructure setup, and add this specific test after that lands

Happy to go either way — just want to keep the PR scope aligned with your preference.

KumarADITHYA123:
@carlesarnal Just checking in — any preference on the test approach? Happy to go either way.

EricWittmann:
Thanks @KumarADITHYA123 — the fix is correct and the input trace table in the PR description was a nice touch. Merged!

FYI, since you opened this PR, Vitest has been added to `ui-app` (there's now a `test` script in `package.json` and existing `.test.ts` files). If you'd like to open a follow-up PR adding a test for this fix, that would be welcome — but it's not required. Apologies for the delay in getting back to you on the test question.

PR #8668, opened 2026-07-16
fix(cli): enhance default group handling and null-safe date conversions

DSingh0304:
**Fixes #8025**

### Description

This PR addresses the handling of the implicit `"default"` group in the Apicurio Registry CLI. Since the registry server treats the `"default"` group as implicit, it does not store a record for it in the database. When fetching artifacts in the default group, the server returns a `null` or blank `groupId`. 

This previously caused a few inconsistencies in the CLI:
1. CLI output tables displayed empty spaces for the group ID instead of the string `"default"`.
2. CRUD operations (create, get, update, delete) on the `"default"` group attempted to call the server API directly, resulting in unexpected errors.

This PR implements the following fixes to handle the default group consistently:
- **Centralized Validation**: Added `IdUtil.isDefaultGroup(String groupId)` to centralize the check for the default group, handling both `null`/blank values and the literal `"default"` string.
- **Display Fixes**: Added `IdUtil.displayGroupId(String groupId)` to ensure `"default"` is properly rendered in all output tables (applied to `ArtifactCommand`, `ArtifactGetCommand`, `VersionCommand`, `VersionGetCommand`, `SearchUtil`, etc.).
- **Graceful CRUD Operations**: Refactored the group commands (`GroupCreateCommand`, `GroupDeleteCommand`, `GroupGetCommand`, `GroupUpdateCommand`) to intercept operations on the implicit default group and return meaningful validation errors or stubbed responses without attempting failing API calls.
- **Testing**: Added comprehensive tests in `GroupCommandTest.java` to ensure all edge cases for the default group are correctly handled.

### Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

### Checklist:
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] My changes generate no new warnings (Checkstyle and compilation pass)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have signed off all my commits (DCO)

PR #8683, opened 2026-07-16
fix(search): normalize negative offset/limit instead of returning 500

### LINKED ISSUE #8611: Search API: negative offset or limit returns 500 and leaks SQL error instead of 400
## Description

**Registry Version**: snapshot (main, commit 63ed0f5d6)
**Persistence type**: sql (in-memory H2, the default)

The v3 search endpoints return HTTP 500 when the `offset` or `limit` query parameter is negative. The error response body also contains the raw database exception, including the generated SQL statement. These endpoints do not require authentication, so any anonymous caller can trigger this.

The affected endpoints are:

- `GET /apis/registry/v3/search/artifacts`
- `GET /apis/registry/v3/search/versions`
- `GET /apis/registry/v3/search/groups`

The value is passed straight to the storage layer, which builds a `LIMIT ? OFFSET ?` query and binds the negative number. The database rejects it, and the resulting exception is not mapped to a client error, so the caller gets a 500 with internal details instead of a 400.

For comparison, `WellKnownResourceImpl.searchAgents` already sanitizes these values with `Math.max(0, offset)` and `Math.max(1, Math.min(limit, 500))` before calling storage, so the intended behaviour is clearly to guard against out of range paging values. The core v3 search methods in `SearchResourceImpl` guard against `null` but not against negative values.

### Environment

Fresh Apicurio Registry instance built from `main`, running with default in-memory (H2) storage, Java 21. No authentication configured. Reproduced with a plain HTTP GET, no client library involved.

### Steps to Reproduce

1. Start a fresh Registry with default settings.
2. Send a search request with a negative offset: `GET /apis/registry/v3/search/artifacts?offset=-1`
3. Observe an HTTP 500 response.
4. Repeat with `?limit=-1` on `/search/artifacts`, `/search/versions`, and `/search/groups`. All return 500.

**Expected**: An invalid pagination value from the client is a client error. The endpoint should return 400 Bad Request (or normalize the value the way `searchAgents` does), and it should not expose any internal SQL or database error in the response.

**Actual**: The endpoint returns 500 Internal Server Error, and the response body contains the database exception and the generated SQL statement.

### Logs

Response body for `GET /apis/registry/v3/search/artifacts?offset=-1`:

```json
{
  "detail": "JdbcSQLDataException: Invalid value \"-1\" for parameter \"result OFFSET\"; SQL statement:\nSELECT a.* FROM artifacts a   ORDER BY coalesce(a.name, a.artifactId) asc LIMIT ? OFFSET ? [90008-240]",
  "title": "org.h2.jdbc.JdbcSQLDataException",
  "status": 500
}
```

Response body for `GET /apis/registry/v3/search/artifacts?limit=-1`:

```json
{
  "detail": "JdbcSQLDataException: Invalid value \"-1\" for parameter \"result FETCH\"; SQL statement:\nSELECT a.* FROM artifacts a   ORDER BY coalesce(a.name, a.artifactId) asc LIMIT ? OFFSET ? [90008-240]",
  "title": "org.h2.jdbc.JdbcSQLDataException",
  "status": 500
}
```

Because these are unmapped 500 responses, they are also reported to the error based liveness check, so malformed pagination values sent by an anonymous client can affect the instance health reporting.

### Suggested fix

Validate `offset` and `limit` in the three `SearchResourceImpl` search methods after the existing null checks, and reject negative values with a `BadRequestException` (the class is already imported and used in this file for other invalid parameters). Optionally add `minimum: 0` to the `offset` and `limit` parameters in the OpenAPI definition so the contract states the constraint. I am happy to open a PR for this once my current PR (#8587) is merged.

Me-Priyank:
Fixes #8611

## Problem

The v3 search endpoints only null-checked the `offset` and `limit` query parameters before passing them to the storage layer. A negative value produced an invalid SQL query (`OFFSET -1` / `LIMIT -1`), which surfaced as a `500` that leaked the generated SQL statement to the caller and, being unmapped, tripped the `ResponseErrorLivenessCheck`.

`WellKnownResourceImpl#searchAgents` already guards against this with `Math.max(0, offset)` and a lower-bounded limit. That inconsistency is the root cause.

## Fix

Clamp `offset` to a minimum of `0` and `limit` to a minimum of `1` before querying storage, via two small private helpers that mirror the `searchAgents` behavior:

- `offset = -1` is normalized to `0` (returns the full result set)
- `limit = -1` is normalized to `1`

Negative values now return `200` with a sensible page instead of a `500`, and nothing internal is leaked.

The helper is applied to every search method in `SearchResourceImpl`, the three GET endpoints from the issue plus the by-content and contracts endpoints, so the whole class is consistent and no search path can hit the invalid-query 500.

I did not port `searchAgents`' upper bound (`Math.min(limit, 500)`): the general search endpoints have never capped the page size, and adding a cap would change existing pagination behavior beyond the scope of this fix.

## Tests

Added negative offset/limit coverage to `SearchArtifactsTest`, `SearchGroupsTest`, and `SearchVersionsTest`, asserting each endpoint returns `200` with the normalized page size (offset `-1` returns the full set, limit `-1` returns a single item).

## Note on adding `minimum: 0` to the OpenAPI spec

@carlesarnal you suggested documenting the constraint with `minimum: 0` in the spec. I left it out deliberately: in this repository the code generator compiles OpenAPI `minimum`/`maximum` into `@DecimalMin`/`@DecimalMax` bean-validation constraints (see the `depth` parameter, which generates `@DecimalMax(value = "10") @DecimalMin(value = "0")` in `GroupsResource`). Adding `minimum` to `offset`/`limit` would therefore reject negatives with a validation `400` *before* the normalization runs, which contradicts the "offset=-1 becomes 0" normalization you preferred.

If you'd rather enforce a hard `400` at the contract layer instead of normalizing, I'm happy to drop the Java normalization and add the `minimum` constraints to the spec, that path is spec-only, no Java needed. Just let me know which you'd prefer.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
@@ -58,6 +58,8 @@ public class SearchResourceImpl implements SearchResource {
 
     private static final String EMPTY_CONTENT_ERROR_MESSAGE = "Empty content is not allowed.";
     private static final String CANONICAL_QUERY_PARAM_ERROR_MESSAGE = "When setting 'canonical' to 'true', the 'artifactType' query parameter is also required.";
+    private static final BigInteger MAX_INT_VALUE = BigInteger.valueOf(Integer.MAX_VALUE);
+    private static final BigInteger MAX_LIMIT = BigInteger.valueOf(1000);
 
     @Inject
     @Current
@@ -137,8 +139,8 @@ public class SearchResourceImpl implements SearchResource {
             filters.add(SearchFilter.ofContentId(contentId));
         }
 
-        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir, offset.intValue(),
-                limit.intValue(), skipCount != null && skipCount);
+        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir, normalizeOffset(offset),
+                normalizeLimit(limit), skipCount != null && skipCount);
         otelMetrics.recordSearchRequest("artifacts");
         return V3ApiUtil.dtoToSearchResults(results);
     }
@@ -187,8 +189,8 @@ public class SearchResourceImpl implements SearchResource {
             filters.add(SearchFilter.ofGroupId(new GroupId(groupId).getRawGroupIdWithNull()));
         }
 
-        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir, offset.intValue(),
-                limit.intValue(), skipCount != null && skipCount);
+        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir, normalizeOffset(offset),
+                normalizeLimit(limit), skipCount != null && skipCount);
         otelMetrics.recordSearchRequest("artifactsByContent");
         return V3ApiUtil.dtoToSearchResults(results);
     }
@@ -244,8 +246,8 @@ public class SearchResourceImpl implements SearchResource {
             }).forEach(filters::add);
         }
 
-        GroupSearchResultsDto results = storage.searchGroups(filters, oBy, oDir, offset.intValue(),
-                limit.intValue());
+        GroupSearchResultsDto results = storage.searchGroups(filters, oBy, oDir, normalizeOffset(offset),
+                normalizeLimit(limit));
         otelMetrics.recordSearchRequest("groups");
         return V3ApiUtil.dtoToSearchResults(results);
     }
@@ -329,8 +331,8 @@ public class SearchResourceImpl implements SearchResource {
             filters.add(SearchFilter.ofStructure(structure));
         }
 
-        VersionSearchResultsDto results = storage.searchVersions(filters, oBy, oDir, offset.intValue(),
-                limit.intValue(), skipCount != null && skipCount);
+        VersionSearchResultsDto results = storage.searchVersions(filters, oBy, oDir, normalizeOffset(offset),
+                normalizeLimit(limit), skipCount != null && skipCount);
         otelMetrics.recordSearchRequest("versions");
         return V3ApiUtil.dtoToSearchResults(results);
     }
@@ -384,8 +386,8 @@ public class SearchResourceImpl implements SearchResource {
             throw new BadRequestException(CANONICAL_QUERY_PARAM_ERROR_MESSAGE);
         }
 
-        VersionSearchResultsDto results = storage.searchVersions(filters, oBy, oDir, offset.intValue(),
-                limit.intValue(), skipCount != null && skipCount);
+        VersionSearchResultsDto results = storage.searchVersions(filters, oBy, oDir, normalizeOffset(offset),
+                normalizeLimit(limit), skipCount != null && skipCount);
         otelMetrics.recordSearchRequest("versionsByContent");
         return V3ApiUtil.dtoToSearchResults(results);
     }
@@ -476,9 +478,19 @@ public class SearchResourceImpl implements SearchResource {
         }
 
         ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir,
-                offset.intValue(), limit.intValue(), false);
+                normalizeOffset(offset), normalizeLimit(limit), false);
         otelMetrics.recordSearchRequest("contracts");
         return V3ApiUtil.dtoToSearchResults(results);
+    }
+
+    // Clamp the offset to [0, Integer.MAX_VALUE] so it never reaches storage as an invalid SQL query (500). See #8611.
+    private static int normalizeOffset(BigInteger offset) {
+        return offset.max(BigInteger.ZERO).min(MAX_INT_VALUE).intValue();
+    }
+
+    // Negative limit -> 1 (limit=0 keeps its empty-page semantics); cap at MAX_LIMIT to bound result size. See #8611.
+    private static int normalizeLimit(BigInteger limit) {
+        return limit.signum() < 0 ? 1 : limit.min(MAX_LIMIT).intValue();
     }
 
     /**
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchArtifactsTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchArtifactsTest.java
@@ -288,4 +288,57 @@ public class SearchArtifactsTest extends AbstractResourceTestBase {
                 .body("count", equalTo(2));
     }
 
+    @Test
+    public void testSearchArtifactsLimitAndOffsetEdgeCases() throws Exception {
+        String group = UUID.randomUUID().toString();
+        String artifactContent = resourceToString("openapi-empty.json");
+
+        for (int idx = 0; idx < 3; idx++) {
+            String artifactId = "Empty-" + idx;
+            String name = "empty-" + idx;
+            this.createArtifact(group, artifactId, ArtifactType.OPENAPI,
+                    artifactContent.replaceAll("Empty API", name), ContentTypes.APPLICATION_JSON, (ca) -> {
+                        ca.setName(name);
+                        ca.getFirstVersion().setName(name);
+                    });
+        }
+
+        // A negative limit is normalized to 1, not passed to storage as an invalid query (#8611).
+        given().when().queryParam("groupId", group).queryParam("limit", -1)
+                .get("/registry/v3/search/artifacts").then().statusCode(200)
+                .body("count", equalTo(3)).body("artifacts.size()", equalTo(1));
+
+        // A negative offset is normalized to 0, returning the full result set.
+        given().when().queryParam("groupId", group).queryParam("offset", -1)
+                .get("/registry/v3/search/artifacts").then().statusCode(200)
+                .body("count", equalTo(3)).body("artifacts.size()", equalTo(3));
+
+        // limit=0 keeps its current semantics (empty page); only negative values are normalized.
+        given().when().queryParam("groupId", group).queryParam("limit", 0)
+                .get("/registry/v3/search/artifacts").then().statusCode(200)
+                .body("count", equalTo(3)).body("artifacts.size()", equalTo(0));
+
+        // Boundary values: offset=0 and limit=1 are valid and behave normally.
+        given().when().queryParam("groupId", group).queryParam("offset", 0)
+                .get("/registry/v3/search/artifacts").then().statusCode(200)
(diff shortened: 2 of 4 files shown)


Me-Priyank:
@EricWittmann @carlesarnal @paoloantinori ptal

app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
```diff
@@ -476,11 +476,21 @@ public ArtifactSearchResults searchContracts(String status, String ownerTeam,
...
     }
 
+    // Clamp a negative offset to 0 so it never reaches storage as an invalid SQL query (500). See #8611.
+    private static int normalizeOffset(BigInteger offset) {
+        return Math.max(0, offset.intValue());
+    }
+
+    // Clamp a limit below 1 to 1, matching searchAgents and avoiding an invalid SQL query (500). See #8611.
+    private static int normalizeLimit(BigInteger limit) {
+        return Math.max(1, limit.intValue());
```
paoloantinori: Worth double-checking the intended behavior for `limit=0`. `Math.max(1, ...)` turns a previously-valid `limit=0` (empty page on backends that accept it) into a 1-element page. If `limit=0` is meant to stay a no-rows call (some clients use it for count-only requests with `count=true`), consider `limit < 0 ? 1 : limit.intValue()` so only negatives are normalized.

paoloantinori (commented):
Good, targeted fix — clamping at the REST boundary is the right layer, and the tests cover all five search methods plus the `searchAgents` precedent you cite. A couple of observations:

- **Behavior change for `limit=0` (not just negatives):** `normalizeLimit` maps any value below 1 — including the previously-valid `limit=0` — to `1`. Depending on the storage backend, `limit=0` today is either an empty page or a 500; after this PR it silently returns a 1-element page. If a client relied on `limit=0` as a cheap "count-only" call (together with `count=true`), it will now receive an artifact it did not ask for. Worth confirming this is intended, or narrowing the clamp to only negatives (`limit < 0 ? 1 : limit`) so the `0` case keeps its current semantics.
- **Pre-existing, not introduced here:** `offset.intValue()` / `limit.intValue()` silently lose magnitude for oversized `BigInteger` values (e.g. `offset` larger than `Integer.MAX_VALUE` wraps around). This PR inherits that behavior via `Math.max(0, offset.intValue())`; flagging only because the issue (#8611) is about input validation, and if the team wants defensive clamping at this boundary this would be the natural place.
- Tests are solid. One small note: the three new tests each create 3 artifacts/groups just to assert `count=3` and page sizes of 1 or 3; the count assertions would hold with a single seeded entity, but the multi-entity setup is reasonable for realism.

app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
Me-Priyank: Thanks @paoloantinori, both good catches.

- **`limit=0`**: you're right, that was an unintended behavior change. Only negative limits are normalized now (`limit.signum() < 0 ? 1 : ...`), so `limit=0` keeps its current empty-page semantics. Added a test asserting that.
- **`Overflow`**: took your suggestion, since this boundary is where the input validation lives, both helpers now clamp the `BigInteger` to `[0, Integer.MAX_VALUE]` before converting to `int`, so oversized values are capped instead of wrapping around. Added tests with `offset`/`limit` = 2^31 (the offset case previously wrapped to a negative int and hit the same 500). The storage layer binds offset and limit as separate SQL parameters, so `Integer.MAX_VALUE` is safe downstream.

paoloantinori:
I left a review on this head earlier (the clamping helpers, the `BigInteger` overflow guard, and the negative offset/limit coverage across the three search test classes). Nothing has changed on the head since then, so those comments still stand — no new concerns from me. Deferring to @EricWittmann / @carlesarnal for the merge call.

carlesarnal (changes requested):
The normalization logic is correct and prevents 500 errors from reaching storage. However, there are several issues:

1. The `@Inject@app/src/main/java/io/apicurio/registry/cdi/Current.java` line looks malformed. Verify this is a diff artifact and not actual code.

2. Test methods use file path annotations instead of `@Test`. This will cause tests to not run. Fix: Replace `@utils/converter/src/test/java/io/apicurio/registry/utils/converter/avro/AvroDataTest.java` with `@Test` on all three new test methods in SearchArtifactsTest.java, SearchGroupsTest.java, and SearchVersionsTest.java.

3. Missing null checks on offset/limit parameters. If these can be null, add guards.

4. This changes API behavior from 500 to 200 for invalid inputs. Document in release notes and consider if this is the desired behavior vs returning 400 Bad Request.

5. Test coverage missing boundary cases (offset=0, limit=1, both params invalid).

6. Consider adding a reasonable upper limit for `limit` parameter (e.g., 1000) instead of allowing Integer.MAX_VALUE which could cause performance issues.

Fix the test annotations and verify the @Inject line. Consider whether silent normalization is better than explicit 400 errors for invalid input.

Me-Priyank:
> The normalization logic is correct and prevents 500 errors from reaching storage. However, there are several issues:
> 
> 1. The `@Inject@app/src/main/java/io/apicurio/registry/cdi/Current.java` line looks malformed. Verify this is a diff artifact and not actual code.
> 2. Test methods use file path annotations instead of `@Test`. This will cause tests to not run. Fix: Replace `@utils/converter/src/test/java/io/apicurio/registry/utils/converter/avro/AvroDataTest.java` with `@Test` on all three new test methods in SearchArtifactsTest.java, SearchGroupsTest.java, and SearchVersionsTest.java.
> 3. Missing null checks on offset/limit parameters. If these can be null, add guards.
> 4. This changes API behavior from 500 to 200 for invalid inputs. Document in release notes and consider if this is the desired behavior vs returning 400 Bad Request.
> 5. Test coverage missing boundary cases (offset=0, limit=1, both params invalid).
> 6. Consider adding a reasonable upper limit for `limit` parameter (e.g., 1000) instead of allowing Integer.MAX_VALUE which could cause performance issues.
> 
> Fix the test annotations and verify the @Inject line. Consider whether silent normalization is better than explicit 400 errors for invalid input.

Thanks @carlesarnal. I found the root of the confusion: a `git pull` had accidentally merged two unrelated PRs (#8668, #8452) plus a merge commit onto this branch, so the diff you reviewed included `cdi/Current.java` and the `@TestInstance` test-annotation refactor. I've rebased the branch clean, it's now a single linear commit touching only `SearchResourceImpl.java` and the three search test classes, which resolves points 1 and 2 (my test methods do use `@Test`).

On the rest:

- **Null checks (3):** already covered, each method null-defaults `offset`→0 and `limit`→20 before the clamping helpers run, so they never receive null.
- **500→200 (4):** this follows your guidance on the issue ("I'd prefer normalizing the values over throwing 400… more forgiving to clients"). I kept normalization; happy to switch to a hard 400 if you've reconsidered. I didn't find a release-notes file in the repo, let me know if there's a process I should add to.
- **Boundary tests (5):** added `offset=0`, `limit=1`, and both-params-invalid cases.
- **Upper limit (6):** `limit` is now capped at `1000` (`MAX_LIMIT`); `searchAgents` uses 500 for its in-memory path, so I can align to 500 instead if you'd prefer a single shared bound.

Me-Priyank:
@EricWittmann @carlesarnal @paoloantinori do i need to change anything else here ?

paoloantinori:
From my side the current head (`cdaf33a`) addresses the points I raised earlier: the `BigInteger` overflow guard via `MAX_INT_VALUE`, the `MAX_LIMIT` cap, and the negative offset/limit coverage in all three search test classes are all in place, and SonarCloud is green. Nothing else outstanding from me.

The remaining change requests are @carlesarnal's — I'll defer to him on those.

paoloantinori (dismissed):
Looks good @Me-Priyank. The `limit=0` behavior is preserved correctly (empty page, not normalized to 1), and the `MAX_LIMIT` cap at 1000 plus the `BigInteger` overflow guard are both sensible additions. Test coverage across all three search endpoints with edge cases (negative, zero, oversized) is thorough.

One small thing I noticed: the `searchContracts` method also got the normalization — good, that was easy to miss.

From my side this is ready. CI needs to be green before merge — the latest run shows a failure, so worth checking if it's related to your changes or a flaky test.

paoloantinori:
Thanks for the ping @Me-Priyank. I already approved this PR on an earlier head — the current head (`733ff80`) looks good from my side as well. The normalization helpers, the `MAX_LIMIT` cap, and the `BigInteger` overflow guard are all in place and the test coverage is thorough. CI just needs to go green. 👍

Me-Priyank:
> Thanks for the ping @Me-Priyank. I already approved this PR on an earlier head — the current head (`733ff80`) looks good from my side as well. The normalization helpers, the `MAX_LIMIT` cap, and the `BigInteger` overflow guard are all in place and the test coverage is thorough. CI just needs to go green. 👍

Thanks @paoloantinori ...  I dug into the CI mechanism a bit, from what I can tell, the PR is still `lifecycle/new`, so the orchestrator is skipping all the test jobs and the gate can't go green yet, If I'm reading it right, an approval alone doesn't move it out of `new` either. So I think it may need a `/accept` from a maintainer first, then I'll `/ready` it right away and the full suite will run against 733ff80.

That said, please correct me if I've misunderstood how the flow works. happy to do whatever's the right next step on my end.

Screenshot 2026-07-21 112706

paoloantinori:
No worries @Me-Priyank — as I noted earlier, my approval is already on the current head. The lifecycle gate is a repo-admin / CI orchestrator concern, not something I can unblock from the review side. @carlesarnal or @EricWittmann would need to move it to a reviewable lifecycle state.

Me-Priyank:
@carlesarnal @EricWittmann @paoloantinori  Converting this to a draft for now, the red **Build** check isn't from this change. It fails at "Check for uncommitted changes" on the generated `docs/.../ref-registry-all-configs.adoc`, which is a `main`-wide breakage: **#8524** changed the config-doc generator to collapse whitespace but didn't regenerate the committed `.adoc`. It's not PR-specific, `main`'s own build is currently red for the same reason ([run](https://github.com/Apicurio/apicurio-registry/actions/runs/29851761921/job/88714312711)), so every full Java build against current `main` trips it.

I'll open a small separate PR that just regenerates `ref-registry-all-configs.adoc` to unblock `main`, then rebase this one and re-run once that's in. This PR doesn't touch that file.

paoloantinori:
Thanks for the heads-up @Me-Priyank. Acknowledged — I'll leave the lifecycle/build gating to the repo admins.

Me-Priyank:
Closing this as of now because of the orchestrator's `one-open-PR-per-contributor` rule... will make this pr again

PR #8684, opened 2026-07-16
fix(cli): apply per-package log levels from config (#8645)

### LINKED ISSUE #8645: CLI: Investigate why logging configuration only works for root logger
## Parent Epic

## Description
The `--verbose` flag and log level configuration only affect the root logger. Per-package log levels (e.g. enabling debug only for HTTP calls or a specific CLI command) do not work as expected.

## Steps to Reproduce
1. Set a per-package log level in the CLI configuration
2. Observe that only the root logger level takes effect

## Expected Behavior
- Per-package log levels should work (e.g. `io.apicurio.registry.cli.services=DEBUG`)
- `--verbose` should continue to work as a global override

## Acceptance Criteria
- Root cause identified
- Per-package log level configuration works
- `--verbose` behavior unchanged

dobbydobap:
## What

`--verbose` enables debug logging, and per-package levels in the CLI config let you quiet
individual packages while it is on. For example, with
`quarkus.log.category."io.netty".level=WARNING` in the config, `acr --verbose ...` shows debug
output for everything except `io.netty`.

Per-package levels only take effect together with `--verbose`. Without it the CLI stays quiet,
whatever the config contains.

## Why

`AbstractCommand.configureVerboseLogging()` only raised the root logger when `--verbose` was
set. There was no code path that read or applied per-package configuration. The packaged CLI
ships `quarkus.log.level=OFF`, and since Quarkus resolves `quarkus.log.*` at build time,
per-package levels supplied at runtime were never picked up by Quarkus's own logging setup.
They have to be applied programmatically, the same way `--verbose` already was.

## How

Renamed the hook to `configureLogging()`. It returns early unless `--verbose` is set. When it
is, it raises the root logger to `FINE`, then reads the `quarkus.log.category."<pkg>".level`
keys from the CLI config and applies each one to its named logger, so a coarser entry
suppresses that package. `toJulLevel(...)` maps Quarkus style names (`DEBUG`, `TRACE`, `WARN`,
...) to plain JUL levels, since the packaged CLI runs on plain JUL where `Level.parse` does
not recognise them. Console handler levels are only ever lowered, never raised, via
`loweredHandlerLevel(...)`, so channels the user did not configure are never suppressed.
Reading the config is guarded so commands that run before install don't regress.

## Testing

`LoggingConfigTest` (`@QuarkusTest`) covers per-package levels applied under `--verbose`,
ignored without it, a noisy package suppressed while the root logger stays at `FINE`, an
unconfigured package left untouched, and `--verbose` raising the root logger.

`LogLevelMappingTest` runs on plain JUL with no JBoss LogManager and covers the level name
mapping plus the only-ever-lowering handler rule.

`cli/README.md` documents the config key format, the `--verbose` requirement and the accepted
level names.

`checkstyle:check` passes.

Closes #8645

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -42,7 +42,7 @@ public abstract class AbstractCommand implements Callable<Integer> {
     public Integer call() {
         var output = new OutputBuffer(config.getStdOut(), config.getStdErr());
         try {
-            configureVerboseLogging();
+            configureLogging();
             updateNotifier.checkAndNotify(getTopLevelCommandName());
             run(output);
             return OK_RETURN_CODE;
@@ -96,18 +96,56 @@ public abstract class AbstractCommand implements Callable<Integer> {
         return current.name();
     }
 
-    private void configureVerboseLogging() {
+    private static final String LOG_CATEGORY_PREFIX = "quarkus.log.category.\"";
+    private static final String LOG_CATEGORY_SUFFIX = "\".level";
+
+    // Applied at runtime because the packaged CLI ships quarkus.log.level=OFF and Quarkus
+    // resolves quarkus.log.* at build time, so per-package levels in the CLI config are
+    // never picked up by Quarkus's own logging setup.
+    private void configureLogging() {
+        var rootLogger = java.util.logging.Logger.getLogger("");
+        java.util.logging.Level finest = null;
+
         var root = spec.root().userObject();
         if (root instanceof Acr acr && acr.isVerbose()) {
-            // Set the root logger level to FINE (DEBUG equivalent)
-            var rootLogger = java.util.logging.Logger.getLogger("");
             rootLogger.setLevel(java.util.logging.Level.FINE);
-            // Also lower handler levels — Quarkus sets quarkus.log.level=WARN on the
-            // console handler, which filters debug messages even when the logger allows them.
-            for (var handler : rootLogger.getHandlers()) {
-                handler.setLevel(java.util.logging.Level.FINE);
-            }
-            log.debug("Verbose logging enabled.");
+            finest = java.util.logging.Level.FINE;
         }
+
+        for (var entry : readLogCategoryLevels().entrySet()) {
+            java.util.logging.Level level;
+            try {
+                level = java.util.logging.Level.parse(entry.getValue().trim());
+            } catch (IllegalArgumentException ex) {
+                log.warnf("Ignoring invalid log level '%s' for category '%s'.", entry.getValue(), entry.getKey());
+                continue;
+            }
+            java.util.logging.Logger.getLogger(entry.getKey()).setLevel(level);
+            if (finest == null || level.intValue() < finest.intValue()) {
+                finest = level;
+            }
+        }
+
+        // Handlers filter by their own level, so open them to the finest level we enabled.
+        if (finest != null) {
+            for (var handler : rootLogger.getHandlers()) {
+                handler.setLevel(finest);
+            }
+        }
+    }
+
+    private java.util.Map<String, String> readLogCategoryLevels() {
+        java.util.Map<String, String> levels = new java.util.HashMap<>();
+        try {
+            config.read().getConfig().forEach((key, value) -> {
+                if (key.startsWith(LOG_CATEGORY_PREFIX) && key.endsWith(LOG_CATEGORY_SUFFIX)
+                        && key.length() > LOG_CATEGORY_PREFIX.length() + LOG_CATEGORY_SUFFIX.length()) {
+                    levels.put(key.substring(LOG_CATEGORY_PREFIX.length(), key.length() - LOG_CATEGORY_SUFFIX.length()), value);
+                }
+            });
+        } catch (CliException ex) {
+            // Config not available yet (e.g. before install) — logging config is best-effort.
+        }
+        return levels;
     }
 }
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -126,10 +126,11 @@ public abstract class AbstractCommand implements Callable<Integer> {
             }
         }
 
-        // Handlers filter by their own level, so open them to the finest level we enabled.
+        // Handlers filter by their own level, so lower them to reveal the finest level we enabled.
+        // Only ever lower, never raise, so we don't suppress unrelated channels the user didn't touch.
         if (finest != null) {
             for (var handler : rootLogger.getHandlers()) {
-                handler.setLevel(finest);
+                handler.setLevel(loweredHandlerLevel(handler.getLevel(), finest));
             }
         }
     }
@@ -164,5 +165,14 @@ public abstract class AbstractCommand implements Callable<Integer> {
             case "ALL" -> java.util.logging.Level.ALL;
             default -> throw new IllegalArgumentException("Unknown log level: " + value);
         };
+    }
+
+    // Returns floor only when the handler is currently coarser (or unset), so a handler is never
+    // raised (which would hide records from channels the user did not configure).
+    static java.util.logging.Level loweredHandlerLevel(java.util.logging.Level current, java.util.logging.Level floor) {
+        if (current == null || current.intValue() > floor.intValue()) {
+            return floor;
+        }
+        return current;
     }
 }
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -115,7 +115,7 @@ public abstract class AbstractCommand implements Callable<Integer> {
         for (var entry : readLogCategoryLevels().entrySet()) {
             java.util.logging.Level level;
             try {
-                level = java.util.logging.Level.parse(entry.getValue().trim());
+                level = toJulLevel(entry.getValue());
             } catch (IllegalArgumentException ex) {
                 log.warnf("Ignoring invalid log level '%s' for category '%s'.", entry.getValue(), entry.getKey());
                 continue;
@@ -147,5 +147,22 @@ public abstract class AbstractCommand implements Callable<Integer> {
             // Config not available yet (e.g. before install) — logging config is best-effort.
         }
         return levels;
+    }
+
+    // Maps Quarkus/JBoss level names (DEBUG, TRACE, WARN, ...) to plain JUL levels. The packaged
+    // CLI runs on plain JUL, where Level.parse does not recognise DEBUG/TRACE.
+    static java.util.logging.Level toJulLevel(String value) {
+        return switch (value.trim().toUpperCase(java.util.Locale.ROOT)) {
+            case "OFF" -> java.util.logging.Level.OFF;
+            case "FATAL", "ERROR", "SEVERE" -> java.util.logging.Level.SEVERE;
+            case "WARN", "WARNING" -> java.util.logging.Level.WARNING;
+            case "INFO" -> java.util.logging.Level.INFO;
+            case "CONFIG" -> java.util.logging.Level.CONFIG;
+            case "DEBUG", "FINE" -> java.util.logging.Level.FINE;
+            case "FINER" -> java.util.logging.Level.FINER;
+            case "TRACE", "FINEST" -> java.util.logging.Level.FINEST;
+            case "ALL" -> java.util.logging.Level.ALL;
+            default -> throw new IllegalArgumentException("Unknown log level: " + value);
+        };
     }
 }
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -8,8 +8,12 @@ import io.apicurio.registry.cli.utils.OutputBuffer;
 import io.apicurio.registry.rest.client.models.ProblemDetails;
 import io.apicurio.registry.rest.client.models.RuleViolationProblemDetails;
 import jakarta.inject.Inject;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.Callable;
+import java.util.logging.Level;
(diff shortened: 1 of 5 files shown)


paoloantinori (commented):
Clean fix for the missing per-package log-level path, and the test coverage (applied / untouched / `--verbose` still raises root) is the right shape. A couple of observations after reading the diff:

- **`Level.parse()` and non-standard level names (`DEBUG`, `TRACE`):** the test resource configures `quarkus.log.category."io.apicurio.registry.cli.services".level=DEBUG`, and `LoggingConfigTest#perPackageLevelFromConfigIsApplied` asserts the resolved level name is `"DEBUG"`. In plain `java.util.logging`, `Level.parse("DEBUG")` throws `IllegalArgumentException` because `DEBUG`/`TRACE` are not among the built-in names (`FINE`, `FINER`, `FINEST`, `INFO`, ...). The test passes because the `@QuarkusTest` runtime has a `Level` named `DEBUG` already registered (Quarkus/JBoss logging registers one). The packaged CLI, however, runs on plain JUL with `quarkus.log.level=OFF` resolved at build time — worth double-checking whether `DEBUG`/`TRACE` in a real `config.json` resolves the same way there as it does under the test. If not, users following the PR description's own example (`level=DEBUG`) would hit the `Ignoring invalid log level` warning and get no per-package logging. The existing `IllegalArgumentException` catch means it fails safe rather than crashing, but it may be worth either documenting the accepted level names or mapping `DEBUG`/`TRACE` to their JUL equivalents (`FINE`/`FINEST`).
- **Handler-level side effect:** opening every root handler to the finest configured level (`AbstractCommand.java` ~line 130) is correct for the targeted package, but it also widens the filter for *every* logger that inherits the root handlers — so if a third library logs at `FINE` for an unrelated category, those messages will now appear once any per-package level is set. This mirrors the existing `--verbose` behavior, so it is consistent; just flagging it since the blast radius is now config-driven rather than only on an explicit `--verbose` flag.
- Minor: `readLogCategoryLevels()` walks the whole config map on every command invocation. Fine for a CLI; flagging only in case the config grows large or is read from a remote source later.

cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
```diff
@@ -96,18 +96,56 @@ private String getTopLevelCommandName() {
...
-            // console handler, which filters debug messages even when the logger allows them.
+            finest = java.util.logging.Level.FINE;
+        }
+
+        for (var entry : readLogCategoryLevels().entrySet()) {
+            java.util.logging.Level level;
+            try {
+                level = java.util.logging.Level.parse(entry.getValue().trim());
+            } catch (IllegalArgumentException ex) {
+                log.warnf("Ignoring invalid log level '%s' for category '%s'.", entry.getValue(), entry.getKey());
```
paoloantinori, line 133: `Level.parse(value)` here will throw `IllegalArgumentException` for `DEBUG`/`TRACE`, which are not built-in JUL level names. The `@QuarkusTest` runtime registers a `DEBUG` level so the test resolves it, but the packaged CLI on plain JUL likely won't — worth verifying that the `DEBUG` example in the PR body actually resolves in the shipped CLI rather than hitting the `Ignoring invalid log level` branch.

cli/src/test/java/io/apicurio/registry/cli/LoggingConfigTest.java
```diff
@@ -0,0 +1,93 @@
...
+        int exitCode = cmd.execute("config", "get", "update.check-enabled");
+        assertThat(exitCode).isEqualTo(0);
+
+        assertThat(Logger.getLogger(CONFIGURED_CATEGORY).getLevel())
+                .isNotNull()
+                .extracting(Level::getName)
+                .isEqualTo("DEBUG");
+    }
+
+    @Test
```
paoloantinori, line 79: This assertion relies on a `Level` named `DEBUG` being resolvable, which only exists because the Quarkus runtime registers it. Consider adding a case (or a comment) clarifying that `FINE`/`FINEST` are the portable JUL equivalents, so users copying the example into a real `config.json` don't get the `Ignoring invalid log level` warning.

carlesarnal (changes requested):
`cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java` has a functional issue. `java.util.logging.Level.parse()` will not parse the configured value shown in `cli/src/test/resources/acr-home-logging/config.json` (`DEBUG`). That means the category level is ignored and the feature does not work for the config format this PR is testing. Add an explicit mapping from Quarkus-style levels to JUL levels before calling `Logger.setLevel()`.

`cli/src/test/java/io/apicurio/registry/cli/LoggingConfigTest.java` also looks wrong in the diff. The three test methods do not have visible `@Test` annotations, and the assertion expects `"DEBUG"` as the JUL level name. If this is the actual code, the tests either will not run or will assert the wrong value. Fix the test annotations and assert the mapped JUL level, likely `Level.FINE`.

dobbydobap:
Thanks both, you were right, and good catch on the `Level.parse` false positive. `java.util.logging.Level.parse` does not recognise `DEBUG/TRACE` on plain `JUL,` so on the packaged CLI the level was silently dropped. My test only passed because the `@QuarkusTest` runtime registers a `DEBUG `level via JBoss LogManager.

Fixed in the new commit:

- Added an explicit `toJulLevel(...)` mapping (`DEBUG` to `FINE`, `TRACE `to `FINEST`, `WARN` to `WARNING`, `ERROR/FATAL` to `SEVERE,` plus the native JUL names) and use it instead of `Level.parse`. Unknown values still fall through to the "ignoring invalid level" warning.

- `LoggingConfigTest` now asserts the mapped `Level.FINE` rather than the name "`DEBUG".`

- Added a plain (non-`@QuarkusTest`) `LogLevelMappingTest` that runs on plain JUL and asserts `toJulLevel("DEBUG") == Level.FINE` and friends, so the mapping is proven with no JBoss LogManager in the picture.

On the `@Test` note: the three `LoggingConfigTest` methods do have `@Test` annotations and do run (the local run reports "Tests run: 3"), the diff view likely hid them. The new plain test makes the coverage unambiguous regardless.
On the two non-blocking notes: the handler widening is intentional and mirrors the existing `--verbose `behavior, so I kept it consistent; and the per-invocation config walk is negligible for a local CLI, happy to revisit if the config ever grows or moves to a remote source.

cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
```diff
@@ -96,18 +96,73 @@ private String getTopLevelCommandName() {
...
+                continue;
+            }
+            java.util.logging.Logger.getLogger(entry.getKey()).setLevel(level);
+            if (finest == null || level.intValue() < finest.intValue()) {
+                finest = level;
+            }
+        }
+
+        // Handlers filter by their own level, so open them to the finest level we enabled.
+        if (finest != null) {
```
paoloantinori: Opening every root handler to `finest` means *all* JVM loggers (Netty, Jackson, Vert.x, ...) emit at that level, not just the configured package(s). Worth scoping this to the configured loggers, or confirming the global widening is intentional.

paoloantinori (commented):
Follow-up review on the current head (`051609a`). The `toJulLevel` mapping and the `LogLevelMappingTest` (plain-JUL, proving Quarkus-style names resolve without JBoss LogManager) directly resolve the `Level.parse` false-positive from the previous round — good.

Two observations worth a look before this lands:

- **Root console handlers are opened to the finest configured package level, which affects every logger channel, not just the configured ones.** `configureLogging()` tracks `finest` across the per-package levels and then sets `handler.setLevel(finest)` on every handler attached to the *root* logger. Root handlers are shared, so any other logger in the JVM (e.g. Netty, Jackson, the Vert.x internals) that logs at that level or finer will now *also* be emitted, even though only one package was explicitly configured. On a packaged CLI this is usually a minor noise issue, but if a user sets `quarkus.log.category."io.apicurio.registry.cli.services".level=FINE` they probably do not expect Netty `FINE` traffic too. Worth confirming this is the intended behaviour or scoping the handler-level change to only the configured loggers (e.g. by attaching a dedicated `ConsoleHandler` to the configured loggers rather than widening the root handlers).
- **No guard against widening handlers when only `--verbose` is set with no per-package config.** The `--verbose` path correctly sets the root logger to `FINE` and `finest=FINE`, so the handler-widening block runs — which is the existing behaviour and fine. But if a future caller ever passes per-package `OFF`/`SEVERE` only, `finest` could end up coarser than a default root handler, which would *raise* the effective level of unrelated channels. A comment stating that handler-widening only ever lowers (never raises) the level, or a `min(current, finest)` clamp, would make the invariant explicit.

dobbydobap:
Thanks, both addressed in the new commit. The handler change is now only-ever-lowering, via a small `loweredHandlerLevel(current, floor)` helper that lowers a handler only when it is coarser than the level we enabled, and never raises it.

On point 1 (leaking other channels): the CLI leaves the console handler at Quarkus's default ALL (nothing sets `quarkus.log.console.level)`, so` loweredHandlerLevel(ALL, ...) `is a no-op and the shared handler is not touched at all in the normal case. When the handler is more restrictive it is lowered just enough to reveal the configured category, and every other channel stays at the root level (`OFF` in the packaged CLI), so it still emits nothing. I kept the shared root handler rather than attaching a dedicated `ConsoleHandler` per configured logger, because a new handler would lose Quarkus's stderr routing and console formatter, so its output would not match the rest of the CLI.

On point 2 (raising the handler): that is exactly what the only-lower rule now prevents. New plain-JUL test `LogLevelMappingTest#handlerLevelOnlyLowersNeverRaises `asserts it: `(INFO, FINE) -> FINE, (INFO, SEVERE) -> INFO` (not raised), `(ALL, FINE) -> ALL` (permissive default untouched), and `(null, FINE) -> FINE.`

paoloantinori (commented):
Thanks @dobbydobap — the only-ever-lowering approach with `loweredHandlerLevel` is cleaner. Looks good from my side. Waiting on maintainer `/accept`.

dobbydobap:
Hi @carlesarnal, I've addressed your review: added the Quarkus→JUL level mapping so
DEBUG/TRACE resolve correctly on the packaged CLI, fixed the test annotations/assertions
to assert the mapped JUL level, and made the handler change only-ever-lowering per
@paoloantinori's follow-up. Paolo confirmed it looks good on his side, and CI + SonarQube
are green. Whenever you have a moment, could you take another look? Thanks for the thorough
review!

paoloantinori:
Nothing further from my side — I already confirmed the only-ever-lowering handler approach and the Quarkus→JUL mapping look correct, and the test coverage (applied / untouched / `--verbose` still raises root) is solid. Deferring to @carlesarnal for the final sign-off.

jsenko:
Thanks for the work here, @dobbydobap — the implementation is solid and the Quarkus→JUL mapping + handler-level invariant are well done.

Before merging, I'd like to discuss the design of how per-package config interacts with `--verbose`, and how users would actually use this feature.

**1. Per-package config should only take effect when `--verbose` is set.**

Currently, per-package levels are applied unconditionally — so setting `io.apicurio.registry.cli.services=DEBUG` in config would produce debug output from that one package even without `--verbose`, while everything else stays silent. That's a confusing UX.

The intended use case is noise reduction: `--verbose` turns on debug logging globally, and per-package config lets you suppress noisy packages (e.g. `io.netty=WARNING`). Without `--verbose`, all logging should stay off regardless of config. This means moving the per-package config reading inside the `if (verbose)` block.

**2. Code style: use imports instead of fully-qualified names.**

`java.util.logging.Level`, `java.util.Map`, `java.util.HashMap`, and `java.util.Locale` are used with full paths throughout. I understand `java.util.logging.Logger` conflicts with the existing `org.jboss.logging.Logger` import, but the others have no conflict and should be imported normally. For the JUL Logger, you could use a qualified name just for that one, or alias it.

**3. Code style: use Javadoc for method-level documentation.**

The `//` comments above `configureLogging()`, `toJulLevel()`, and `loweredHandlerLevel()` should be Javadoc (`/** ... */`) since they document method purpose and behavior.

**4. Documentation.**

The feature needs at least a brief note explaining the config key format and intended usage (suppress noisy packages when using `--verbose`). The CLI has a `README.md` (`cli/README.md`) that could be a good place for this.

@vandanayadav7 — since you filed the original issue (#8645), does the "per-package config as noise suppression for `--verbose`" model match your expectation? Or did you have a different use case in mind?

dobbydobap:
Thanks @jsenko, good catches. All four addressed in the new commit.

1. Per-package config now requires `--verbose`. `configureLogging()` returns early when
`--verbose` is not set, so the config is not even read and nothing is touched. With
`--verbose` the root logger goes to `FINE` and per-package levels are applied on top, so a
coarser entry like `io.netty=WARNING` quiets that package while everything else stays at
debug. That matches the noise reduction model you described.

For tests, `perPackageLevelIsIgnoredWithoutVerbose` pins the new behaviour, and
`noisyPackageIsSuppressedUnderVerbose` covers the suppression case using a `WARNING` category
added to the test config. I checked that the first one genuinely fails against the old code
before the change, so it is not a false positive. I also added `--verbose` to
`unconfiguredPackageIsUntouched`, since without it that test would now pass trivially.

One side effect worth noting: this also resolves @paoloantinori's earlier point about
`readLogCategoryLevels()` walking the config on every invocation, since it now only runs
under `--verbose`.

2. Imported `Level`, `Map`, `HashMap` and `Locale`. I kept `java.util.logging.Logger`
qualified at its two call sites because of the existing `org.jboss.logging.Logger` import
for the `log` field.

3. Converted the comments on `configureLogging`, `readLogCategoryLevels`, `toJulLevel` and
`loweredHandlerLevel` to Javadoc, and documented the `--verbose` requirement on
`configureLogging` itself.

4. Added a Logging section to `cli/README.md` under Configuration, covering the config key
format, the `--verbose` requirement, the suppression use case and the accepted level names.

I have also updated the PR description, which still described the old unconditional behaviour.

@vandanayadav7, if the use case you had in mind for #8645 was different, happy to adjust.

paoloantinori:
Thanks @dobbydobap — re-read the latest head (`e56002c`). Nothing further from my side; the changes addressing @jsenko's design points (early-return unless `--verbose`, only-ever-lowering handlers, the README level list) are all consistent with what I reviewed earlier. Deferring to @jsenko on the design discussion.

jsenko (approved):

jsenko:
Thank you for contributing!

Issue #8495, opened 2026-07-08
Parallelize operator integration tests using JUnit 5 extensions


jsenko:
## Problem

The operator integration test suite runs **all ~28 test classes sequentially** in a single JVM, taking **90+ minutes** in CI (see #7485 for the detailed breakdown). Each test class performs a full operator deploy/undeploy cycle even though most tests don't modify the operator itself. Only 3 out of 28 test classes actually mutate the operator deployment.

The current `ITBase` design couples operator lifecycle to test class lifecycle (`@BeforeAll` starts operator, `@AfterAll` stops it), which prevents running multiple test classes in parallel. PR #8452 converts static fields to `@TestInstance(PER_CLASS)` instance fields, which removes Java-level state sharing across subclasses, but does not address the Kubernetes-level isolation needed for actual parallelism.

## Goal

Run the majority of operator integration tests **in parallel**, each in its own namespace, sharing a single cluster-wide operator instance. Tests that mutate the operator itself run sequentially in a separate phase.

### Single operator, per-test namespaces

Deploy the operator **once in cluster-wide mode** (watching all namespaces) at the start of the test suite. Each test class gets its own unique namespace (already the case via `calculateNamespace()`), creates its CR there, and the shared operator reconciles it. The operator is shut down after all tests complete.

In **local mode**, the in-process operator starts with cluster-wide namespace watching. In **remote mode**, the operator is deployed once to the cluster with cluster-wide config.

### JUnit 5 extension as the orchestration layer

Replace the current `ITBase` base class inheritance with a **JUnit 5 extension** (`OperatorTestExtension` or a redesigned version) that handles:

1. **Operator lifecycle (root store scope):** Lazy-initialized via `ExtensionContext.Store` at the `ROOT` level using `CloseableResource`. The first test class to run triggers operator startup; `CloseableResource.close()` stops it when JUnit shuts down.

2. **Namespace lifecycle (per-class scope):** The extension creates a unique namespace in `BeforeAllCallback` and deletes it in `AfterAllCallback`. Tests don't manage their own namespaces.

3. **Shared infrastructure (root store, lazy):** Resources like Strimzi are installed on-demand using the same `getOrComputeIfAbsent` pattern. The first Kafka test triggers Strimzi installation; cleanup happens at JUnit shutdown via `CloseableResource.close()`. An annotation like `@NeedsStrimzi` on the test class can signal the requirement.

[...]

PR #8686, opened 2026-07-16
test(operator): install Strimzi once per cluster with cluster-wide watch

paoloantinori:
## Summary

Follow-up to #8452. The Kafka operator ITs currently download the Strimzi release manifest from GitHub and install the operator into every test class's throwaway namespace — 4 downloads + 4 installs (~80-90s each) per kafka CI job. This PR installs Strimzi **once per cluster**, into a dedicated long-lived namespace, configured to watch all namespaces (`STRIMZI_NAMESPACE=*`), so test classes only create their `Kafka` CRs.

Expected saving: ~3-5 min per kafka job (3 installs + 3 downloads eliminated); measured numbers will be posted to discussion #8364 after 3 green runs. It also removes the per-class rebinding of cluster-scoped RBAC subjects, which was safe only under strictly sequential execution — a prerequisite for any future test parallelism.

## Changes

- **New `StrimziClusterWideInstaller`**: idempotent install (JVM guard + cluster readiness probe, correct here — unlike the #8452 per-class guard — because the target namespace is never deleted); transforms the manifest per the Strimzi watch-all-namespaces procedure (RoleBinding→ClusterRoleBinding for the 3 documented bindings, leader-election stays namespaced); fail-fast allow-list so a version bump cannot silently escalate a new namespaced binding to cluster scope; 409-tolerant namespace creation; namespace overridable via `test.operator.strimzi-namespace`.
- **Vendored manifest** (`src/test/resources/strimzi/`, byte-identical to upstream 0.47.0, README with bump procedure) — no more GitHub download at test runtime; excluded from the published tests-classifier jar.
- **`ITBase.applyStrimziResources()`** now delegates; the 4 Kafka test classes are unchanged.
- **Failure diagnostics**: `OperatorTestContext.extraDiagnosticNamespaces()` (default empty) lets `OperatorTestExtension` dump the Strimzi namespace on failure — operator reconcile logs are the primary Kafka diagnostic and no longer live in the test namespace.
- **Unit test** `StrimziClusterWideInstallerTest` (tagged `kafka`, no cluster needed) asserts the full transform against the real vendored manifest: env `*`, exact ClusterRoleBinding name set, preserved roleRefs, subject rebinding, leader-election kept namespaced.
- `operator/README.md`: documents the new property.

## Known limitation

On persistent dev clusters, a manifest version bump does not trigger reinstall while an old operator Deployment is still ready — cleanup is manual (documented in the vendored README). CI clusters are ephemeral, so this does not affect CI.

## Testing

- `StrimziClusterWideInstallerTest` passes locally (transform validated without a cluster).
- Full operator reactor compiles; checkstyle clean.
- Real verification is this PR's kafka CI group: expect a single Strimzi install in the job log and all 4 classes green.

Backlog: REG-111. Related: #8452, discussion #8364.


carlesarnal (changes requested):
The refactoring from per-class to cluster-wide Strimzi install is solid. Reduces test overhead and follows Strimzi's documented cluster-wide watch procedure. Three issues need fixing:

1. **Critical**: Malformed annotations in `StrimziClusterWideInstallerTest.java`. Lines show `@app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsTagProjector.java(Tags.KAFKA)` and `@utils/converter/src/test/java/io/apicurio/registry/utils/converter/avro/AvroDataTest.java` which are file paths, not annotations. Should be `@Tag(Tags.KAFKA)` and `@Test` respectively.

2. **Bug**: Race condition in `ensureInstalled()`. If operator becomes ready between the `installedInThisJvm` check and `isOperatorReady()` call, the flag never gets set. Move the flag assignment before the ready check or restructure the logic.

3. **Minor**: Namespace creation catches 409 but doesn't verify it's actually a conflict. Add `ex.getMessage().contains("AlreadyExists")` or similar check before swallowing the exception.

The manifest transformation logic is well-tested and fail-fast on unexpected RBAC layout. The vendored manifest approach is good for supply chain security. Documentation is clear about cleanup and version bumps.

File paths: `operator/controller/src/test/java/io/apicurio/registry/operator/it/StrimziClusterWideInstaller.java` (race condition), `StrimziClusterWideInstallerTest.java` (annotations).

paoloantinori:
@carlesarnal Addressed in commit 498ba34:

**Finding #1** (malformed annotations): Already fixed in a previous commit — the test file has correct `@Tag(Tags.KAFKA)` and `@Test`.

**Finding #2** (race condition in `ensureInstalled()`): The method is `synchronized` and the flag is set in both paths (line 88 for already-ready, line 99 after install). If `waitForOperatorReady` throws, the flag stays false and the next call correctly retries. No race exists.

**Finding #3** (409 check): Fixed. Now uses the idiomatic Fabric8 pattern:
```java
if (ex.getStatus() == null || !"AlreadyExists".equals(ex.getStatus().getReason())) {
    throw ex;
}
```
This checks the K8s API `reason` field on the `Status` object instead of string-matching `getMessage()`, and is null-safe.

paoloantinori:
Looked into the failing run. The only red job is **Operator Tests / Remote Tests (olm-v1)** — specifically `UpgradeOLMITTest.testManualApprovalUpgrade`, which times out (622s) waiting for `Deployment apicurio-registry-operator-v3.2.4 should exist` in the `olm-tests` module. That test has nothing to do with the Strimzi cluster-wide installer this PR changes — it's an OLM upgrade-path test that pulls a specific published operator version.

All the tests this PR actually touches are green:
- Remote Tests (kafka): success ✅
- Remote Tests (feature-setup, feature, auth, database, olm-v0): success ✅
- Build Operator, Check Install File: success ✅

This looks like a flaky/unrelated OLM infra failure (the v3.2.4 catalog entry or its mirror), not a regression from `StrimziClusterWideInstaller`. A `/retry` on the workflow should confirm — if `olm-v1` goes red again on the same test with the same version pin, it's an upstream OLM catalog issue, not this PR.

paoloantinori:
@carlesarnal All three findings addressed (see above comment) and CI is fully green now — including olm-v1 which passed on the latest run. Ready for re-review whenever you get a chance.

jsenko:
@paoloantinori There is already an issue assigned to an upstream contributor that includes similar work https://github.com/Apicurio/apicurio-registry/issues/8495 . Do not merge this please until that's discussed.

paoloantinori:
Acknowledged, @jsenko — I'll hold this PR until the overlap with #8495 is discussed. Thanks for flagging it; happy to coordinate with the upstream contributor to avoid duplicating effort or to fold this work into their direction.

aclfe:
Hello, there's no conflict with #8495 as I'm working on the issue. Frankly Shared Strimzi is just one bullet of Phase 1 in that issue. #8495 will stack on this cleanly, same way this stacked on #8452. Merging it doesn't duplicate effort, plus I'm watching this PR in order to be in sync. Thanks for keeping me in the loop everyone, I appreciate it 

Separately, I've left a longer comment on #8495 regarding some incompatibility / issues, but I checked there's some workarounds so that can be discussed when the PR itself comes out.

paoloantinori:
Thanks @aclfe, that's helpful context — appreciate you confirming the layering. If #8495 stacks cleanly on top of this (same way this one stacked on #8452) and you're actively tracking it, the overlap concern is resolved from my side. The cluster-wide Strimzi install here is a clean prerequisite either way.

paoloantinori:
@carlesarnal thanks for the thorough review. All three items are addressed in the commits already on the branch:

1. **Annotations** — fixed in c976c73a. `StrimziClusterWideInstallerTest.java` uses proper `@Tag(Tags.KAFKA)` and `@Test` annotations.

2. **Race condition** — fixed in 9de23133. `ensureInstalled()` is `synchronized`, the `installedInThisJvm` flag is set after both the early-exit path (`isOperatorReady()` returns true) and the install path. The redundant `volatile` was also dropped since all access is behind the synchronized block.

3. **409 check** — fixed in 1ea0d037. `ensureNamespace()` now checks `!"AlreadyExists".equals(ex.getStatus().getReason())` and re-throws non-conflict errors.

Ready for re-review when you get a chance.

jsenko (commented):
## Review: Approach is sound

I reviewed the same `StrimziClusterWideInstaller` code as part of PR #8818 (which stacks on top of this). The cluster-wide install approach is correct:

- Idempotent install with JVM + cluster-state guards
- `CONVERTIBLE_ROLE_BINDINGS` allowlist prevents silent scope escalation on version bumps
- Vendored manifest avoids GitHub download failures at test time
- `OperatorTestContext.extraDiagnosticNamespaces()` is a good extension point for failure diagnostics
- carlesarnal's requested changes appear addressed per paoloantinori's follow-up comments

**On the DO NOT MERGE label**: jsenko flagged overlap with #8495. aclfe (the #8495 assignee) confirmed the work is complementary — #8818 stacks cleanly on this PR. The overlap concern appears resolved. jsenko should confirm whether the DO NOT MERGE can be lifted.

operator/controller/src/test/resources/strimzi/strimzi-cluster-operator-0.47.0.yaml
carlesarnal, line 1: I don't think checking this in is the right solution. We must download it from the resource, as we were previously doing.

carlesarnal (changes requested):

paoloantinori:
@carlesarnal fair pushback — let me lay out why I went with vendoring rather than a once-per-cluster download, and I'm happy to reconsider if the tradeoff reads differently to you.

The previous per-class path downloaded `strimzi-cluster-operator-0.47.0.yaml` from `github.com/strimzi/...` four times per kafka job (~80-90s each, plus the install). Moving to one download per cluster removes 3 of those 4, but keeps the remaining one — so the job is still dependent on GitHub being reachable, still subject to rate-limiting on fork PRs, and still pays the cold-JVM latency every time. Vendoring the manifest (byte-identical to the 0.47.0 release, and excluded from the published test-jar via the `pom.xml` `<excludes>strimzi/**</excludes>` so olm-tests don't carry the ~1MB) makes the install deterministic and local-only — which is the property I was actually after, not just the time saving.

The transform logic (RoleBinding→ClusterRoleBinding for the three watch-all-namespaces bindings, the fail-fast allow-list, leader-election staying namespaced) lives in code regardless of where the bytes come from, so that part is unchanged either way. And the bump procedure is documented in `src/test/resources/strimzi/README.md`, with `StrimziClusterWideInstallerTest` pinning the transform so a version bump that changes the binding structure fails fast instead of silently escalating scope.

If the concern is the maintenance cost of carrying a 19k-line YAML in the tree, that's a fair point and I'd switch to a single per-cluster download if you'd rather not vendor — but I'd want to keep the determinism argument on the table, since GitHub-availability flakiness is what originally motivated this.

aclfe (commented):
Maybe we can pluck things out from both ideas and make everyone happy. The best part might be the "Moving to one download per cluster"  (single cluster-wide operator) change as paolo said. 

Though the vendoring that rode in on the same PR does seems scary... Checking in the `19k-line` / `~1MB` release manifest isn't great. It's third-party artifact that rot in the tree, and it shows up in every diff, blame and grep for no benefit.

It seems vendoring won't be a huge concern for flakiness, because we're down to one install per job as paolo changed. But I also don't think "just download it at runtime like before" is the best endpoint, because we still take a live GitHub dependency inside the test JVM. 

**So here's a third option:**
Tier 1: Land the cluster-wide install (4→1) as-is. It's the REG-111 follow-up, kills 3 of 4 downloads/installs, removes the RBAC churn, and doesn't depend on manifest sourcing.
Tier 2: Agree with carlesarnal not to vendor 19k lines of third-party YAML. But instead of runtime download, fetch it once at build time. (**precedent: integration-tests/pom.xml already does this**), download the manifest into target/test-classes/strimzi/. The existing getResourceAsStream("/strimzi/...") code works unchanged. Nothing in tree and version pinned in pom.xml. This is just applying the module's existing convention 

Also one thing worth a look before merge, unrelated to the manifest debate: if two JVMs ever share a cluster, can't both read the operator as not-ready and both run the full createOrReplace over the CRDs and ClusterRoles at once. ensureNamespace AlreadyExists, but the resource loop doesn't have the same guard. Probably fine for now, but maybe in the future cluster-wide-into-a-shared-namespace is where it could bite, so maybe a conflict-tolerant create or at least a note>?

jsenko (changes requested):
Thanks @paoloantinori — the cluster-wide Strimzi install approach is sound and the code is clean. The transformation logic, allowlist fail-fast, and idempotent install are all well done. Three changes before merging:

**1. Replace vendored manifest with runtime download (should-fix)**

The 19k-line / ~1MB `strimzi-cluster-operator-0.47.0.yaml` checked into git is too heavy. Instead, download it once per cluster at runtime from the Strimzi GitHub release, driven by a `test.operator.strimzi-version` property. This gives the same savings (1 install vs 4) without the git bloat, and version bumps become a property change instead of re-vendoring 19k lines. Aligns with @carlesarnal's feedback.

**2. Add version check for reused clusters (should-fix)**

Currently, `ensureInstalled()` skips if any Strimzi Deployment is ready, regardless of version. On persistent dev clusters, a version bump silently uses the old operator. When `isOperatorReady()` finds a running Strimzi, compare its container image tag against the expected version. If mismatched, tear down and reinstall. This removes the "manual cleanup on dev clusters" limitation documented in the PR.

**3. Remove `StrimziClusterWideInstallerTest` (nice-to-have)**

This test validates the manifest transformation against a specific Strimzi version's resource layout — but the `CONVERTIBLE_ROLE_BINDINGS` allowlist already provides the same fail-fast safety with clear, actionable error messages (names the unexpected binding, points to the README). The integration tests would surface the failure anyway. The unit test is testing test infrastructure and is redundant with the allowlist mechanism.

aclfe:
hi @paoloantinori, any update on jsenko's 3 review items?

paoloantinori:
Hi @aclfe — thanks for the nudge. jsenko's three items (runtime download instead of vendored manifest, version check for reused clusters, and dropping `StrimziClusterWideInstallerTest`) are noted and on my list. I haven't pushed the changes yet; I'll address them in the next batch of commits on this branch and ping here when it's ready for re-review.

aclfe:
@paoloantinori re-ping

paoloantinori:
Thanks for the re-ping, and for confirming the #8495 layering directly — that removes the hold I'd put on this. With the overlap concern resolved (your #8495 stacks on this cleanly, shared-Strimzi being one phase of the larger work), I'll address jsenko's three items on the next push:

1. Runtime download vs. vendored manifest.
2. Version check for reused clusters.
3. Dropping `StrimziClusterWideInstallerTest`.

I'll have those up shortly. Appreciate you tracking the coordination.

carlesarnal:
This is already done.

PR #8690, opened 2026-07-16
fix(app): reject null and non-string prompt template field

### LINKED ISSUE #8669: bug: Null template field bypasses isMissingNode check, renders as literal "null"
### Description

extractTemplateText guards against a missing 'template' field using templateField.isMissingNode(), but isMissingNode() only returns true for MissingNode. When the stored PROMPT_TEMPLATE artifact has content like {"template": null}, Jackson returns a NullNode. NullNode.isMissingNode() is false, so the guard is skipped. NullNode.asText() returns the 4-character string "null", which is then used as the template text, producing rendered: "null" with no error.

### Steps to reproduce

1. Open `app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java` around line 89
2. A PROMPT_TEMPLATE artifact with {"template": null} content causes the render endpoint to silently succeed and return rendered: "null" instead of a 400 error. No validation error is reported.

### Expected behavior

isMissingNode() returns true only for MissingNode (absent key). NullNode (present key with JSON null value) returns isMissingNode() == false and asText() == "null". The fix used isMissingNode() but did not also check isNull(), leaving the null-value case unguarded. should be handled correctly.

### Actual behavior

A PROMPT_TEMPLATE artifact with {"template": null} content causes the render endpoint to silently succeed and return rendered: "null" instead of a 400 error. No validation error is reported.

deepak0x:
## What

`PromptRenderingService.extractTemplateText` guarded a missing `template` field with `isMissingNode()`, which only returns true for a `MissingNode` (an absent key). A stored `PROMPT_TEMPLATE` artifact with content like `{"template": null}` parses to a `NullNode`, for which `isMissingNode()` is `false`, so the guard was skipped and `NullNode.asText()` returned the literal 4-character string `"null"`. The render endpoint then silently succeeded and returned `rendered: "null"` instead of reporting an error.

This adds a check that the `template` field is textual. A missing key keeps its existing "missing 'template' field" error, and an explicit null or any non-string value (number, boolean, object, array) is now rejected with `InvalidContentException`, which maps to a 400 the same way as the other invalid-content cases in this service.

## Why

An invalid prompt template should fail loudly with a validation error, not silently render the string `"null"` (or a numeric/boolean value coerced via `asText()`). This closes the `NullNode`/non-string gap left by the `isMissingNode()`-only guard.

## Tests

Added three regression tests to `PromptRenderingServiceTest`:
- `testNullTemplateFieldJson` — `{"template": null}` throws `InvalidContentException`
- `testNullTemplateFieldYaml` — YAML `template: null` throws `InvalidContentException`
- `testNonStringTemplateField` — `{"template": 123}` throws `InvalidContentException`

All three fail against the previous code (which rendered `"null"`/`"123"`) and pass with this change. `./mvnw checkstyle:check -pl app` passes.

Closes #8669

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -89,6 +89,9 @@ public class PromptRenderingService {
         if (templateField.isMissingNode()) {
             throw new InvalidContentException("Prompt template is missing 'template' field");
         }
+        if (!templateField.isTextual()) {
+            throw new InvalidContentException("Prompt template 'template' field must be a string");
+        }
         return templateField.asText();
     }
 
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -89,10 +89,17 @@ public class PromptRenderingService {
         if (templateField.isMissingNode()) {
             throw new InvalidContentException("Prompt template is missing 'template' field");
         }
+        if (templateField.isNull()) {
+            throw new InvalidContentException("Prompt template 'template' field must not be null");
+        }
         if (!templateField.isTextual()) {
             throw new InvalidContentException("Prompt template 'template' field must be a string");
         }
-        return templateField.asText();
+        String templateText = templateField.asText();
+        if (templateText.isBlank()) {
+            throw new InvalidContentException("Prompt template 'template' field must not be blank");
+        }
+        return templateText;
     }
 
     /**
--- a/app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
+++ b/app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
@@ -545,6 +545,55 @@ public class PromptRenderingServiceTest {
     }
 
     @Test
+    public void testNullTemplateFieldJson() {
+        String jsonContent = """
+            {
+              "templateId": "null-template",
+              "template": null
+            }
+            """;
+
+        ContentHandle content = ContentHandle.create(jsonContent);
+        Map<String, Object> variables = Map.of();
+
+        Assertions.assertThrows(InvalidContentException.class, () -> {
+            renderingService.render(content, variables, "default", "null-template", "1.0");
+        });
+    }
+
+    @Test
+    public void testNullTemplateFieldYaml() {
+        String yamlContent = """
+            templateId: null-template
+            template: null
+            """;
+
+        ContentHandle content = ContentHandle.create(yamlContent);
+        Map<String, Object> variables = Map.of();
+
+        Assertions.assertThrows(InvalidContentException.class, () -> {
+            renderingService.render(content, variables, "default", "null-template", "1.0");
+        });
+    }
+
+    @Test
+    public void testNonStringTemplateField() {
+        String jsonContent = """
+            {
+              "templateId": "numeric-template",
+              "template": 123
+            }
+            """;
+
+        ContentHandle content = ContentHandle.create(jsonContent);
+        Map<String, Object> variables = Map.of();
+
+        Assertions.assertThrows(InvalidContentException.class, () -> {
+            renderingService.render(content, variables, "default", "numeric-template", "1.0");
+        });
+    }
+
+    @Test
     public void testInvalidYamlContent() {
         String invalidContent = "{{{{not valid yaml or json}}}}";
 
--- a/app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
+++ b/app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
@@ -594,6 +594,40 @@ public class PromptRenderingServiceTest {
     }
 
     @Test
+    public void testEmptyTemplateField() {
+        String jsonContent = """
+            {
+              "templateId": "empty-template",
+              "template": ""
+            }
+            """;
+
+        ContentHandle content = ContentHandle.create(jsonContent);
+        Map<String, Object> variables = Map.of();
+
+        Assertions.assertThrows(InvalidContentException.class, () -> {
+            renderingService.render(content, variables, "default", "empty-template", "1.0");
+        });
+    }
+
+    @Test
+    public void testWhitespaceTemplateField() {
+        String jsonContent = """
+            {
+              "templateId": "whitespace-template",
+              "template": "   "
+            }
+            """;
+
+        ContentHandle content = ContentHandle.create(jsonContent);
+        Map<String, Object> variables = Map.of();
+
+        Assertions.assertThrows(InvalidContentException.class, () -> {
+            renderingService.render(content, variables, "default", "whitespace-template", "1.0");
+        });
+    }
+
+    @Test
     public void testInvalidYamlContent() {
         String invalidContent = "{{{{not valid yaml or json}}}}";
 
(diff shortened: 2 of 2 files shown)


app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -89,6 +89,9 @@ private String extractTemplateText(JsonNode templateNode) {
         if (templateField.isMissingNode()) {
             throw new InvalidContentException("Prompt template is missing 'template' field");
         }
+        if (!templateField.isTextual()) {
+            throw new InvalidContentException("Prompt template 'template' field must be a string");
```
paoloantinori, line 96: Worth noting: `isTextual()` is `true` for an empty/whitespace string, so `{"template": ""}` still renders as `""` without an error. If that's intentional (WIP templates), ignore — otherwise a follow-up `isBlank()` check would complete the "invalid templates fail loudly" goal.

paoloantinori, line 96: Minor UX nit: an explicit `templateField.isNull()` branch with a more specific message (`"'template' is null"`) would help users distinguish a stored null from a genuinely wrong type (number/array/object), since the root cause is different.

paoloantinori (commented):
Solid fix — adding the `isTextual()` guard closes the `NullNode`/`asText()` trap (where `asText()` on a `NullNode` returns the literal `"null"`), and the regression tests for JSON null, YAML null, and a non-string value cover the regression well. Two small follow-ups worth considering, neither of which should block this:

- **Empty / whitespace-only `template` is still rendered silently.** A `template` of `""` or `"   "` has `isTextual() == true`, so it passes the new guard and `substituteVariables("", variables)` returns `""`, yielding `rendered: ""` without any error. That's at odds with the PR's stated goal ("an invalid prompt template should fail loudly"), so it might be worth a follow-up — either a separate `isBlank()` check or a quality rule, depending on whether an intentionally empty template is a legitimate state.
- **Exception message could distinguish explicit `null` from other non-string values.** For `{"template": null}` the message `"...must be a string"` is technically correct but less diagnostic than a dedicated `isNull()` branch that says e.g. `"'template' is null"`. Minor, but it would help users debugging why their stored artifact trips the guard.

One thing I'd confirm: the new case relies on the existing `InvalidContentException` → HTTP 400 mapping (untested at the integration level here). That mapping is shared with the other invalid-content cases in this service, so I don't think there's real risk — just flagging that the 400 claim is by inheritance rather than by a direct test.

deepak0x:
Thanks @paoloantinori, both good calls — addressed in 83ca0d2:

1. **Empty / whitespace template.** Added an `isBlank()` check after `asText()`, so `{"template": ""}` and `{"template": "   "}` now fail with `InvalidContentException` instead of rendering `""`. I treated an empty template as invalid to match the "fail loudly" goal — if you consider an intentionally-empty template a legitimate (e.g. WIP) state, I'm happy to drop this guard or gate it behind a quality rule instead.

2. **Distinguish explicit null.** Added a dedicated `templateField.isNull()` branch with `"'template' field must not be null"`, so an explicit null now gets a specific message and non-string types keep `"must be a string"`.

On the 400 mapping: I checked the wiring rather than leave it implicit — `HttpStatusCodeMap` maps `InvalidContentException` to `HTTP_BAD_REQUEST` (line 104), the same entry every other invalid-content case here relies on. You're right that there's still no endpoint-level integration test asserting the 400 for this path; the new tests assert the exception at the service layer only.

Verification (ran `PromptRenderingServiceTest` via `mvnw test -pl app`):
- All 32 tests pass with the guards in place.
- Removing the `isTextual` guard fails the 3 null/non-string tests; removing `isBlank` fails the 2 empty/whitespace tests — each reporting "Expected InvalidContentException to be thrown, but nothing was thrown". So each guard is what makes its tests pass.

checkstyle:check on `app` is clean.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
deepak0x: Good catch. Added an isBlank() check after asText(), so empty and whitespace-only templates throw InvalidContentException now instead of quietly rendering "". I went with treating empty as invalid to stay in line with the fail-loudly goal, but if an empty template is a valid WIP state I'm fine dropping it or moving it to a quality rule instead. Two new tests cover it, in 83ca0d2.

deepak0x: Done in 83ca0d2. Added an isNull() branch with its own message ("'template' field must not be null"), so an explicit null reads differently from a wrong type, which still gets "must be a string".

paoloantinori:
Thanks — this head addresses both points cleanly. The `isBlank()` guard after `asText()` closes the empty/whitespace gap, and splitting the null / non-string / blank cases into distinct `InvalidContentException` messages makes the failures much easier to diagnose. The five regression tests (JSON null, YAML null, numeric, empty, whitespace) give good coverage. No further concerns from me.

carlesarnal (changes requested):
Solid validation improvement. The code correctly rejects null, non-string, and blank template fields with clear error messages. Test coverage is thorough for common cases.

Two notes:

1. Verify test annotations are correct. The diff shows file paths as annotations which looks wrong.

2. The DynamicConfigPropertyIndexImpl.java/@OdcsInfo annotation convention appears violated, but that file isn't touched by this PR. Clarify if this is a project-wide issue or if the convention statement is outdated.

The blank check subsumes the empty check, but having both tests doesn't hurt. Consider adding a test for non-primitive JSON types (array/object) if you want complete coverage.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java - validation logic is correct and follows existing patterns.

app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java - comprehensive test cases, verify annotation syntax.

deepak0x:
Thanks @carlesarnal, addressed below.

1. Annotations: the five new tests use plain JUnit 5 `@Test`, same as the rest of the class. I think the "file paths as annotations" is a rendering artifact of the review rather than something in the source. The file compiles and the tests run: `mvnw test -pl app` gives 34 passing, and `checkstyle:check` on `app` is clean.

2. `DynamicConfigPropertyIndexImpl.java` / `@OdcsInfo`: this PR doesn't touch that file, so it's out of scope here. I can't say whether the convention note is outdated or a wider issue. Happy to look into it separately if it's worth its own ticket.

3. Added `testArrayTemplateField` (`{"template": ["a","b"]}`) and `testObjectTemplateField` (`{"template": {"nested":"value"}}`) in 49807af. Both get rejected by the existing `isTextual()` guard with the "must be a string" message, so array and object values are covered now too.

deepak0x:
Thanks for the reviews @paoloantinori @carlesarnal — I've addressed the feedback in the latest commits. This PR is still on `lifecycle/new`, so the Verification Gate stays red and the test pipeline never runs. Could a maintainer `/accept` it so the smoke and full tests can kick off? Happy to rebase on `main` as well if that helps.

paoloantinori:
Thanks @deepak0x — the latest head addresses both earlier points cleanly (the `isBlank()` guard for empty/whitespace templates, and the split null / non-string / blank `InvalidContentException` messages). The test coverage on `PromptRenderingServiceTest` is thorough.

On the CI question: you're right that this is still on `lifecycle/new`, so the Verification Gate stays red and the smoke pipeline won't run until it's accepted — and the `/accept` is a maintainer-side gate action that I don't trigger from review. Pinging @carlesarnal as the other reviewer here to take the accept call. Happy to re-review once it's off `lifecycle/new`, or after a rebase onto `main` if that's preferred.

paoloantinori (commented):
Seen the fixes in 83ca0d2 — the `isBlank()` check and the separate `isNull()` branch with distinct error messages are both correct. Waiting on maintainer `/accept` to trigger CI (current run shows failure, likely due to needing workflow approval).

EricWittmann:
Thanks @deepak0x — clean fix with distinct error messages for each invalid case and thorough test coverage. Apologies for the delay. Merged!

PR #8692, opened 2026-07-17
fix: accept all 2xx status codes and restore interrupt flag in HttpCl…

### LINKED ISSUE #8691: Bug: HttpClientService rejects valid HTTP 2xx responses and swallows interrupt flag
### Describe the bug

`HttpClientService.java` has two interconnected bugs in the `post()` method:

**1. Only HTTP 200 treated as success (line 36):**

```java
if (httpResponse.statusCode() == 200) {
    return (O) httpResponse.bodyAsJson(outputClass);
} else {
    throw new HttpClientException(...)
}
```

HTTP 201 (Created), 202 (Accepted), and 204 (No Content) all throw `HttpClientException` and trigger SmallRye `@Retry` with 8 retries and exponential backoff. For webhook endpoints, 202 is the standard acknowledgment response per the CloudEvents specification.

**2. InterruptedException swallows thread interrupt flag (lines 43-44):**

```java
} catch (InterruptedException e) {
    throw new HttpClientException(e);
}
```

Per the Java concurrency contract, catching `InterruptedException` clears the thread's interrupt flag. Without calling `Thread.currentThread().interrupt()` before rethrowing, callers (including SmallRye Fault Tolerance `@Retry`) that check `Thread.interrupted()` will not see the interruption, potentially causing retry loops to continue when they should abort or thread pool shutdown to hang.

### Steps to reproduce

1. Configure a webhook endpoint that returns HTTP 202 Accepted
2. Trigger an artifact event that sends a webhook
3. Observe: `HttpClientException` thrown, retries triggered despite successful delivery

### Expected behavior

- HTTP 200-299 treated as success
- HTTP 204 and empty body responses return null
- `InterruptedException` restores the interrupt flag before rethrowing

### Environment

- File: `app/src/main/java/io/apicurio/registry/http/HttpClientService.java`

Kakumanu-Harshitha:
Fixes #8691

## What this PR does

Fixes two interconnected bugs in `HttpClientService.post()`.

### 1. Accept all 2xx HTTP status codes

**Before:** Only `200` was treated as success. `201`, `202`, `204` all threw `HttpClientException`, triggering SmallRye `@Retry` with 8 retries and exponential backoff despite the request succeeding.

**After:** The full `200-299` range is treated as success. `204` and null body responses return `null` instead of attempting JSON deserialization on empty content.

This is critical for webhook endpoints that return `202 Accepted` — the standard CloudEvents acknowledgment. Without this fix, successful webhook deliveries trigger 8 unnecessary retries.

```diff
-            if (httpResponse.statusCode() == 200) {
-                return (O) httpResponse.bodyAsJson(outputClass);
+            int status = httpResponse.statusCode();
+            if (status >= 200 && status < 300) {
+                if (status == 204 || httpResponse.body() == null) {
+                    return null;
+                }
+                return (O) httpResponse.bodyAsJson(outputClass);
             } else {
-                throw new HttpClientException("Webhook request failed (" + httpResponse.statusCode() + "): " + httpResponse.statusMessage());
+                throw new HttpClientException("Webhook request failed (" + status + "): " + httpResponse.statusMessage());
             }
```

### 2. Restore interrupt flag on InterruptedException

**Before:** `InterruptedException` caught and wrapped without restoring `Thread.currentThread().interrupt()`.

**After:** Interrupt flag restored before rethrowing, per the Java concurrency contract. This follows the same pattern used in `KubernetesManager.java`.

```diff
         } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
             throw new HttpClientException(e);
         }
```

Without this fix, SmallRye `@Retry` may not detect interruption, and thread pool shutdown can hang.

### Caller safety analysis

All current callers of `post()` go through `AbstractConfiguredArtifactTypeUtil.invokeHook()` for artifact type validation webhooks, which always expect a JSON response body (never 204). All callers wrap the call in `catch (Throwable e)` blocks, so a null return would be safely caught.

### Input trace verification

| Status | Body | Result |
|--------|------|--------|
| 200 | `{"id": 1}` | Parsed and returned ✅ |
| 201 | `{"id": 1}` | Parsed and returned ✅ |
| 202 | `null` | Returns `null` ✅ |
| 204 | (no body) | Returns `null` ✅ |
| 200 | `null` | Returns `null` ✅ |
| 400 | error | Throws `HttpClientException` ✅ |
| 500 | error | Throws `HttpClientException` ✅ |

### Files changed

- `app/src/main/java/io/apicurio/registry/http/HttpClientService.java`

### DIFF
--- a/app/src/main/java/io/apicurio/registry/http/HttpClientService.java
+++ b/app/src/main/java/io/apicurio/registry/http/HttpClientService.java
@@ -21,7 +21,18 @@ public class HttpClientService {
     @Inject
     WebClient webClient;
 
-    @Retry(maxRetries = 8, delay = 100, jitter = 50)
+    /**
+     * Posts a JSON request to the given URL and returns the parsed response.
+     *
+     * <p>All HTTP 2xx status codes are treated as success. A {@code 204 No Content} response
+     * or any 2xx response with a {@code null} body will return {@code null} rather than
+     * attempting JSON deserialization. Callers that require a non-null response should
+     * handle this case accordingly.
+     *
+     * @throws HttpClientException on non-2xx responses or communication errors
+     * @throws HttpClientInterruptException if the calling thread is interrupted (aborts retry)
+     */
+    @Retry(maxRetries = 8, delay = 100, jitter = 50, abortOn = HttpClientInterruptException.class)
     @ExponentialBackoff
     @Timeout(value = 10, unit = ChronoUnit.SECONDS)
     public <I, O> O post(String url, I body, Class<O> outputClass) throws HttpClientException {
@@ -33,15 +44,20 @@ public class HttpClientService {
 
             // Wait for the response (vert.x is async).
             HttpResponse<Buffer> httpResponse = future.toCompletionStage().toCompletableFuture().get();
-            if (httpResponse.statusCode() == 200) {
+            int status = httpResponse.statusCode();
+            if (status >= 200 && status < 300) {
+                if (status == 204 || httpResponse.body() == null || httpResponse.body().length() == 0) {
+                    return null;
+                }
                 return (O) httpResponse.bodyAsJson(outputClass);
             } else {
-                throw new HttpClientException("Webhook request failed (" + httpResponse.statusCode() + "): " + httpResponse.statusMessage());
+                throw new HttpClientException("Webhook request failed (" + status + "): " + httpResponse.statusMessage());
             }
         } catch (ExecutionException e) {
             throw new HttpClientException(e);
         } catch (InterruptedException e) {
-            throw new HttpClientException(e);
+            Thread.currentThread().interrupt();
+            throw new HttpClientInterruptException(e);
         }
     }
 }
--- a/app/src/test/java/io/apicurio/registry/http/HttpClientServiceTest.java
+++ b/app/src/test/java/io/apicurio/registry/http/HttpClientServiceTest.java
@@ -1,0 +1,198 @@
+package io.apicurio.registry.http;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpServer;
+import jakarta.inject.Inject;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests for {@link HttpClientService} verifying correct HTTP status code handling
+ * and interrupt flag restoration.
+ *
+ * <p>Uses a Vert.x HTTP stub server (no external dependencies required) following
+ * the same pattern as {@code WebhookArtifactTypesTest}.
+ */
+@QuarkusTest
+public class HttpClientServiceTest {
+
+    @Inject
+    HttpClientService httpClientService;
+
+    private static Vertx vertx;
+    private static HttpServer server;
+    private static int serverPort;
+    private static final AtomicInteger requestCount = new AtomicInteger(0);
+    private static CountDownLatch slowRequestReceived;
+
+    /**
+     * Simple response class for JSON deserialization in tests.
+     */
+    public static class TestResponse {
+        public String message;
+    }
+
+    @BeforeAll
+    static void startStubServer() throws Exception {
+        vertx = Vertx.vertx();
+        CountDownLatch latch = new CountDownLatch(1);
+
+        server = vertx.createHttpServer();
+        server.requestHandler(req -> {
+            requestCount.incrementAndGet();
+            String path = req.path();
+
+            req.bodyHandler(body -> {
+                switch (path) {
+                    case "/status-200":
+                        req.response()
+                                .setStatusCode(200)
+                                .putHeader("Content-Type", "application/json")
+                                .end("{\"message\":\"ok\"}");
+                        break;
+                    case "/status-202":
+                        req.response()
+                                .setStatusCode(202)
+                                .putHeader("Content-Type", "application/json")
+                                .end("{\"message\":\"accepted\"}");
+                        break;
+                    case "/status-204":
+                        req.response()
+                                .setStatusCode(204)
+                                .end();
+                        break;
+                    case "/slow":
+                        // Signal that the request was received, then do not respond —
+                        // connection stays open for interrupt testing
+                        slowRequestReceived.countDown();
+                        break;
+                    default:
+                        req.response().setStatusCode(404).end();
+                        break;
+                }
+            });
+        });
+
+        server.listen(0).onComplete(ar -> {
+            if (ar.succeeded()) {
+                serverPort = ar.result().actualPort();
+                latch.countDown();
+            }
+        });
+
+        latch.await();
+    }
+
+    @AfterAll
+    static void stopStubServer() {
+        if (server != null) {
+            server.close();
+        }
+        if (vertx != null) {
+            vertx.close();
+        }
+    }
+
+    @BeforeEach
+    void resetState() {
+        requestCount.set(0);
+        slowRequestReceived = new CountDownLatch(1);
+    }
+
+    /**
+     * Verifies that an HTTP 202 Accepted response is treated as success,
+     * returns the parsed body, and does not trigger SmallRye @Retry.
+     */
+    @Test
+    void testHttp202DoesNotThrowAndDoesNotRetry() throws HttpClientException {
+        TestResponse response = httpClientService.post(
+                stubUrl("/status-202"), "{}", TestResponse.class);
+
+        assertNotNull(response, "202 with body should return parsed response");
+        assertEquals("accepted", response.message);
+        assertEquals(1, requestCount.get(),
+                "Server should receive exactly 1 request (no retries on 202)");
+    }
+
+    /**
+     * Verifies that an HTTP 204 No Content response returns null
+     * without attempting JSON deserialization.
+     */
+    @Test
+    void testHttp204ReturnsNullWithoutDeserialization() throws HttpClientException {
+        TestResponse response = httpClientService.post(
+                stubUrl("/status-204"), "{}", TestResponse.class);
+
+        assertNull(response, "204 No Content should return null");
+        assertEquals(1, requestCount.get(),
+                "Server should receive exactly 1 request (no retries on 204)");
+    }
+
+    /**
+     * Verifies that an HTTP 200 response with a JSON body is parsed correctly
+     * (baseline behavior preserved).
+     */
+    @Test
+    void testHttp200ReturnsBody() throws HttpClientException {
+        TestResponse response = httpClientService.post(
+                stubUrl("/status-200"), "{}", TestResponse.class);
+
+        assertNotNull(response);
+        assertEquals("ok", response.message);
+        assertEquals(1, requestCount.get(),
+                "Server should receive exactly 1 request (no retries on 200)");
+    }
+
+    /**
+     * Verifies that interrupting a thread blocked on {@code Future.get()} results in
+     * an {@link HttpClientException} and that the thread's interrupt flag is restored
+     * per the Java concurrency contract.
+     */
+    @Test
+    void testInterruptRestoresFlag() throws Exception {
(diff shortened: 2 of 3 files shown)


paoloantinori (commented):
Both fixes are correct and the issue (#8691) is real — accepting only `200` would make a `202 Accepted` webhook acknowledgment trigger the full SmallRye `@Retry` (8 retries, exponential backoff) for an operation that already succeeded, and swallowing the interrupt flag on `InterruptedException` is the classic anti-pattern. The implementation looks sound. Two observations, the first one more important.

- **No test coverage for either fix, and the class is currently untested.** There is no existing test referencing `HttpClientService` anywhere in the repo, and this PR adds none. These are behavior changes to the webhook delivery path: (a) the 2xx-range acceptance and the `204`/null-body → `return null` short-circuit, and (b) the interrupt-flag restoration. Both are easy to regress — a future refactor that narrows the range back to `== 200`, or drops the `Thread.currentThread().interrupt()` call, would reintroduce #8691 silently. A focused test using `WireMock`/`MockWebServer` (or a stubbed `WebClient`) asserting: (1) a `202` response does not throw and does not retry beyond one call, (2) a `204` returns `null` without attempting JSON deserialization, and (3) interrupting the blocked `.get()` surfaces a `HttpClientException` with `Thread.currentThread().isInterrupted() == true` afterward. Given this method carries `@Retry(maxRetries = 8)`, the absence of tests for the retry-trigger boundary is the gap I'd most like to see closed.

- **`return null` changes the method contract for callers expecting a non-null `O`.** Any caller of `post(...)` that does not null-check the result could now NPE on a `204` where it previously got an exception (from `bodyAsJson` on empty content). From a quick look the current callers consume webhook acknowledgments and don't use the typed return, so this is probably fine, but it's worth a comment on the method (or in the issue) that a 2xx-with-empty-body now yields `null` rather than an exception, so future callers handle it.

Nice, tight fix — just add the tests so this doesn't silently revert.

Kakumanu-Harshitha:
Thanks for the review @paoloantinori — updated the commit with both changes:

1. **Tests added** — `HttpClientServiceTest` (Vert.x stub server, same pattern as `WebhookArtifactTypesTest`): asserts 202 doesn't throw/retry, 204 returns null, and interrupt restores the flag with `isInterrupted() == true`.

2. **Javadoc added** — documents that `post()` now returns `null` on 204/empty-body 2xx responses, so future callers know to handle it.

app/src/main/java/io/apicurio/registry/http/HttpClientService.java
```diff
@@ -33,14 +43,19 @@ public <I, O> O post(String url, I body, Class<O> outputClass) throws HttpClient
...
+            int status = httpResponse.statusCode();
+            if (status >= 200 && status < 300) {
+                if (status == 204 || httpResponse.body() == null) {
+                    return null;
+                }
                 return (O) httpResponse.bodyAsJson(outputClass);
             } else {
-                throw new HttpClientException("Webhook request failed (" + httpResponse.statusCode() + "): " + httpResponse.statusMessage());
+                throw new HttpClientException("Webhook request failed (" + status + "): " + httpResponse.statusMessage());
             }
```
paoloantinori, line 55: Restoring the flag here is correct in isolation (the JVM clears it when throwing `InterruptedException`), but note the interaction with `@Retry(maxRetries = 8)` above: the wrapped `HttpClientException` is retried (default `retryOn = Exception.class`), and since the flag is now re-set to `true`, the next `future...get()` throws `InterruptedException` *immediately* — so a single interrupt cycles through all 8 retries in a tight loop instead of cancelling. If interrupts should cancel promptly, consider excluding this case from retry (e.g. `abortOn`, or a distinct exception subtype).

app/src/test/java/io/apicurio/registry/http/HttpClientServiceTest.java
```diff
@@ -0,0 +1,193 @@
...
+     * per the Java concurrency contract.
+     */
+    @Test
+    void testInterruptRestoresFlag() throws Exception {
+        AtomicReference<Throwable> caught = new AtomicReference<>();
+        AtomicBoolean interruptedAfter = new AtomicBoolean(false);
+
+        Thread testThread = new Thread(() -> {
+            try {
+                // This will block on .get() because /slow never responds
```
paoloantinori, line 173: The `Thread.sleep(500)` assumes the worker has entered the blocking `.get()` before the interrupt fires — timing-dependent under CI load. With `@Retry` active plus the re-set flag, the 8 retries also each re-hit the interrupt; worth confirming the `join(10_000)` bound holds against that path (not a single interrupt) so this doesn't become a slow/flaky test.

paoloantinori (commented):
Quick look at #8692. The two core fixes are correct — treating the full `200-299` range as success is the right call for webhook/CloudEvents acks (a `202` triggering 8 retries was a real bug), and restoring the interrupt flag is the right Java concurrency hygiene. One interaction worth thinking through before this lands, below.

- **`@Retry` will replay an interrupted call up to 8 times.** `post(...)` is annotated `@Retry(maxRetries = 8, ...)` with the default `retryOn = Exception.class`, so the `HttpClientException` wrapping the `InterruptedException` is retried. The added `Thread.currentThread().interrupt()` correctly re-sets the flag (the JVM clears it when throwing `InterruptedException`), but that means *each retry immediately re-encounters the interrupt* at `future...get()` (which checks the flag and throws `InterruptedException` again), so a single interrupt spins through all 8 retries in a tight loop rather than cancelling promptly. An interrupt is usually meant as "stop this operation now", so 8 immediate re-attempts work against that intent. Worth confirming whether `HttpClientException` wrapping an interrupt should instead be excluded from retry (e.g. `@Retry(abortOn = ...)` for the interrupt-wrapped case, or a distinct exception type), or whether the current spin-loop behaviour is acceptable here.
- **`testInterruptRestoresFlag` is timing-dependent and may be slow/flaky under CI load.** It `Thread.sleep(500)` assuming the worker has already entered the blocking `.get()` on `/slow`; under load that window can shrink. More importantly, with `@Retry` active and the flag re-set by the new code, the 8 retries each re-hit the interrupt and the `@Timeout(value = 10, ...)` — worth double-checking the test bounds (the `join(10_000)` cap) hold against the 8-retry × interrupt path rather than a single interrupt.
- Minor: `httpResponse.body() == null` guards the deserialization, but Vert.x `HttpResponse.body()` for an *empty* (non-204) 2xx body typically returns an empty `Buffer`, not `null`, so such a response would still flow into `bodyAsJson(outputClass)`. Probably benign (JSON deserialization of empty input returns null), but the guard's intent ("don't attempt JSON deserialization on empty content") may not be fully achieved for empty-but-non-null bodies.

carlesarnal (changes requested):
Missing Apache 2.0 license headers on both new files. Add them.

`serialVersionUID = 0` in HttpClientInterruptException should be `1L`.

The 2xx status code handling is correct. The interrupt flag restoration follows Java concurrency best practices. The `abortOn` configuration prevents wasted retry attempts on interrupts.

Test coverage is solid for the happy path but missing error cases. Add tests for non-2xx responses and 200 with empty body.

The null return for 204/empty bodies is a behavior change. Verify existing call sites handle null. Document this in the PR description if not already done.

The body length check is safe due to short-circuit evaluation but could be clearer. Consider reordering for readability.

File paths: `app/src/main/java/io/apicurio/registry/http/HttpClientInterruptException.java`, `app/src/main/java/io/apicurio/registry/http/HttpClientService.java`, `app/src/test/java/io/apicurio/registry/http/HttpClientServiceTest.java`

paoloantinori:
Thanks — this head addresses both items. The `HttpClientServiceTest` with the Vert.x stub server (202 doesn't throw/retry, 204 returns null, interrupt restores the flag) is solid coverage, and the Javadoc on `post()` documenting the null-on-204 behavior will save future callers from an NPE surprise. From my side the two core fixes are done; the remaining `CHANGES_REQUESTED` from @carlesarnal (license headers, `serialVersionUID`) is in his court.

Kakumanu-Harshitha:
Thanks @carlesarnal — all addressed in the latest push:

Added Apache 2.0 license headers to both new files
Changed serialVersionUID to 1L
Added testNon2xxThrowsHttpClientException — verifies 500 throws HttpClientException with status code in message
Added testHttp200EmptyBodyReturnsNull — verifies 200 with empty body returns null
Extracted responseBody variable for readability in the null/empty check
Verified call site: invokeHook() passes through the return value directly, callers already handle null

Kakumanu-Harshitha:
hey @carlesarnal PTAL! when you are free?

paoloantinori (commented):
Seen the fixes (license headers, `serialVersionUID`, test additions). Looks good from my side — waiting on Carles's re-review.

EricWittmann:
@Kakumanu-Harshitha — the CI failure in `testInterruptRestoresFlag` is consistent across all 3 retry
attempts. The issue is that Quarkus CDI (Arc) wraps your `HttpClientInterruptException` in an
`ArcUndeclaredThrowableException` before it reaches the caller:

```
expected: <io.apicurio.registry.http.HttpClientInterruptException>
but was:  <io.quarkus.arc.ArcUndeclaredThrowableException>
```

The test asserts on the top-level exception type, but since `HttpClientService` is a CDI bean, Arc's
proxy intercepts the undeclared exception and wraps it.

To fix the test, you'll need to assert on the cause chain rather than the top-level type — e.g.,
check that `getCause()` is an instance of `HttpClientInterruptException`.

There's also a deeper design question worth investigating: if Arc wraps the exception before
SmallRye Fault Tolerance sees it, does `@Retry(abortOn = HttpClientInterruptException.class)`
actually match and abort? If SmallRye only sees `ArcUndeclaredThrowableException`, the `abortOn`
won't fire and the interrupt will still spin through all 8 retries. You may need to either:

- Add `ArcUndeclaredThrowableException` to `abortOn` as well, or
- Make `HttpClientInterruptException` extend `RuntimeException` so Arc doesn't wrap it, or
- Verify empirically that SmallRye unwraps the cause chain when matching `abortOn`

Please push a fix and let us know what you find on the `abortOn` behavior.

paoloantinori:
@EricWittmann's diagnosis is spot-on, and it surfaces a question that goes beyond fixing the test assertion.

The `ArcUndeclaredThrowableException` wrapping happens at the Arc client-proxy boundary. Whether FT's `@Retry` sees the raw `HttpClientInterruptException` (and thus matches `abortOn`) or the already-wrapped `ArcUndeclaredThrowableException` (and thus does **not** match `abortOn`) depends entirely on interceptor ordering — and that's exactly what needs empirical confirmation rather than assumption. SmallRye FT's `abortOn` checks `isInstance()` on the thrown exception directly; it does not walk the cause chain. So if FT only sees the wrapper, `abortOn = HttpClientInterruptException.class` silently fails to fire and the interrupt still cycles through all 8 retries — which would mean the core fix of this PR (point #2) doesn't actually work at runtime.

The good news: the test already has the instrument to settle this definitively. `requestCount` on the stub server increments on every incoming request. If `abortOn` fires, the server sees exactly **1** request. If Arc wraps before FT sees it, the server sees up to **8**. Adding

```java
assertEquals(1, requestCount.get(),
    "interrupt should abort retry, not cycle through all attempts");
```

to `testInterruptRestoresFlag` verifies the actual end-to-end behavior — far more valuable than asserting on exception types alone, since it proves the retry contract holds regardless of which layer wraps first.

For the assertion itself: unwrapping via the cause chain (`assertInstanceOf(HttpClientInterruptException.class, caught.get().getCause())`) is the right fix, as Eric suggests.

EricWittmann (commented):
Thanks for the work on this @Kakumanu-Harshitha — the 2xx fix and interrupt flag restoration are both correct and address a real production issue. A few things to resolve to get this ready for merge:

**1. Fix the failing `testInterruptRestoresFlag` test (blocking).** As @EricWittmann diagnosed, Quarkus CDI (Arc) wraps your `HttpClientInterruptException` in an `ArcUndeclaredThrowableException` at the proxy boundary. But before simply fixing the test assertion, you need to answer the deeper question @paoloantinori raised: **does `@Retry`'s `abortOn = HttpClientInterruptException.class` actually match through the Arc wrapper?**

If Arc wraps the exception *before* the SmallRye FT interceptor evaluates `abortOn`, then the `abortOn` won't match, and interrupts will still trigger all 8 retries — defeating the purpose of `HttpClientInterruptException`. Please verify this empirically (e.g., add a test that confirms the interrupt does NOT trigger retries when `abortOn` is configured). If `abortOn` doesn't work through the Arc wrapping, you'll need a different approach (e.g., aborting on `ArcUndeclaredThrowableException`, or restructuring the exception handling).

**2. Replace `Thread.sleep(500)` with proper synchronization.** The interrupt test assumes the worker thread has entered the blocking `.get()` call within 500ms. Under CI load this can race. Use a `CountDownLatch` or `CompletableFuture` to signal when the worker is ready for the interrupt, rather than relying on timing.

**3. @carlesarnal's review items.** Make sure the license headers and `serialVersionUID = 1L` changes are in the current commit. The `CHANGES_REQUESTED` review is still active — once the above items are addressed, we can get a re-review.

The 2xx status code fix and the interrupt flag restoration are solid — it's the `abortOn` + Arc interaction and the test reliability that need your attention. Looking forward to the update!

EricWittmann:
@Kakumanu-Harshitha — closing this PR for now due to inactivity. The reviews from Jul 27 (Arc proxy wrapping diagnosis, `abortOn` verification, test fix) haven't received a response in ~2 weeks.

The underlying bug is real and the core fix is solid — feel free to re-open this PR (or reach out on the Apicurio Slack/Zulip) when you're ready to pick it back up. The main items to address are outlined in the Jul 27 reviews above.

Thanks for the contribution!

PR #8694, opened 2026-07-17
fix(gitops): trust configured repositories with arbitrary UIDs

BevisDev:
## Summary

- Configure Git `safe.directory` entries for the GitOps workspace and configured repositories.
- Use environment-based Git configuration so the fix works when `/home/git` is not writable.
- Support repositories configured at the workspace root with `dir: .`.
- Avoid the unrestricted `safe.directory=*` setting.

Fixes #<issue-number>

## Problem

When the GitOps repository is configured with `dir: .`, its root is the `/repos` volume mount point. Kubernetes or OpenShift may assign ownership of this directory to root or another UID.

Git 2.35.2+ then rejects fetch operations with:

```text
fatal: detected dubious ownership in repository at '/repos'

### DIFF
--- a/distro/gitops/entrypoint.sh
+++ b/distro/gitops/entrypoint.sh
@@ -247,6 +247,34 @@ validate_repo_list() {
     done
 }
 
+# Trust only the workspace paths managed by this sidecar. Environment-based
+# configuration works with OpenShift arbitrary UIDs even when $HOME is not
+# writable, unlike `git config --global`.
+configure_git_safe_directories() {
+    local config_count="${GIT_CONFIG_COUNT:-0}"
+    local repo_dir repo_path
+    local workspace_path="${WORKSPACE%/}"
+    [ -n "${workspace_path}" ] || workspace_path="/"
+
+    export "GIT_CONFIG_KEY_${config_count}=safe.directory"
+    export "GIT_CONFIG_VALUE_${config_count}=${workspace_path}"
+    config_count=$((config_count + 1))
+
+    for repo_dir in "${REPO_DIRS[@]}"; do
+        # The workspace entry above already covers a repo configured with dir=.
+        if [ "${repo_dir}" = "." ]; then
+            continue
+        fi
+
+        repo_path="${workspace_path%/}/${repo_dir}"
+        export "GIT_CONFIG_KEY_${config_count}=safe.directory"
+        export "GIT_CONFIG_VALUE_${config_count}=${repo_path}"
+        config_count=$((config_count + 1))
+    done
+
+    export GIT_CONFIG_COUNT="${config_count}"
+}
+
 # ---------------------------------------------------------------------------
 # Security validation
 # ---------------------------------------------------------------------------
@@ -713,6 +741,7 @@ main() {
     log "Apicurio Registry GitOps sidecar starting"
 
     build_repo_list
+    configure_git_safe_directories
 
     log "  Workspace:     ${WORKSPACE}"
     log "  Security:      ${SECURITY}"
--- a/distro/gitops/entrypoint.sh
+++ b/distro/gitops/entrypoint.sh
@@ -253,8 +253,13 @@ validate_repo_list() {
 configure_git_safe_directories() {
     local config_count="${GIT_CONFIG_COUNT:-0}"
     local repo_dir repo_path
-    local workspace_path="${WORKSPACE%/}"
+    local workspace_path="${WORKSPACE:-}"
     [ -n "${workspace_path}" ] || workspace_path="/"
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by carlesarnal — CHANGES_REQUESTED** [review]
Syntax error in array expansion will break the script. Line with `"${REPO_DIRS[ @]}"` needs the space removed: `"${REPO_DIRS[@]}"`. 

The approach is sound - using GIT_CONFIG_KEY/VALUE environment variables instead of git config --global correctly handles OpenShift arbitrary UIDs. The workspace path normalization logic is correct. The skip for repo_dir="." avoids redundant configuration.

Fix the array syntax and this is good to merge. Consider adding shell script tests for the gitops sidecar if none exist.

Path: distro/gitops/entrypoint.sh

**Comment by BevisDev** [comment]
@carlesarnal  Thanks for the review,
The array expansion is already `${REPO_DIRS[@]}` without a space. I also added local shell tests for the `safe.directory` configuration as suggested. All tests pass.
Could you please take another look?

File: distro/gitops/entrypoint.sh [hunk 1]
```diff
@@ -247,6 +247,34 @@ validate_repo_list() {
...
+
+    for repo_dir in "${REPO_DIRS[@]}"; do
+        # The workspace entry above already covers a repo configured with dir=.
+        if [ "${repo_dir}" = "." ]; then
+            continue
+        fi
+
+        repo_path="${workspace_path%/}/${repo_dir}"
+        export "GIT_CONFIG_KEY_${config_count}=safe.directory"
+        export "GIT_CONFIG_VALUE_${config_count}=${repo_path}"
```
paoloantinori, line 280: Worth double-checking: `${workspace_path%/}` only strips a single trailing slash. If `WORKSPACE` ever arrives with multiple trailing slashes (unlikely from K8s volume mounts, but possible from user config), `repo_path` could end up with a double slash like `/repos//default`. A `readlink -f`-style normalization or a `while` strip would be more defensive, though this is low-risk as written.

paoloantinori (commented):
Solid fix for the Git 2.35.2+ "dubious ownership" failure under OpenShift arbitrary UIDs. The approach is the right one for this environment: using `GIT_CONFIG_KEY_*` / `GIT_CONFIG_VALUE_*` / `GIT_CONFIG_COUNT` environment variables instead of `git config --global` works when `$HOME` is not writable (common under OpenShift's restricted SCC), and scoping `safe.directory` to only the workspace + configured repo paths (rather than the unrestricted `safe.directory=*`) is the security-conscious choice. The `dir: .` deduplication (skipping the redundant entry since the workspace path already covers it) is a nice touch.

A couple of observations worth considering:

- **Path construction when WORKSPACE has no trailing slash.** In `configure_git_safe_directories()`, `repo_path="${workspace_path%/}/${repo_dir}"` strips one trailing slash from `workspace_path`, then appends `/<repo_dir>`. That's correct for `WORKSPACE=/repos` → `/repos/default`. But if `WORKSPACE` somehow ended up with multiple trailing slashes (e.g. `/repos//`), `${workspace_path%/}` only strips one, leaving `/repos//default`. Minor and unlikely from Kubernetes volume mounts, but a `while`-style strip or a `readlink -f` normalization would be more defensive.

- **The test extracts the function via `sed` regex.** `tests/test-safe-directory.sh` loads `configure_git_safe_directories()` by `sed -n '/^configure_git_safe_directories()/,/^}/p'`. This is fragile to refactoring: if the function body ever gains a nested `}` on column 1, or the function moves, the test silently breaks or loads an empty body. The `declare -F` guard afterward catches the empty-body case, which is good. Worth a comment in the test noting this coupling so future editors know to keep the function definition matching the regex.

The test coverage itself (named repo, `dir=.`, pre-existing `GIT_CONFIG_COUNT` preservation, empty workspace) is thorough and directly exercises the env-var assembly logic.

distro/gitops/entrypoint.sh
BevisDev: Thanks for the thorough review, @paoloantinori 

- Trailing slashes: switched to a `while`-strip that removes all trailing slashes while keeping `/` intact for the root case, and guarded the repo path join so `WORKSPACE=/` no longer produces `//<dir>`. I went with `while` over `readlink -f` since the function runs before the repos are cloned, so the paths may not exist on disk yet.
- Test coupling: added a comment above the `sed` extraction noting it depends on the function staying flat (no nested `}` at column 0), so future edits keep the regex working.

Also added a test case for repeated trailing slashes (`/repos//`). All safe.directory tests pass.

jsenko (approved):
LGTM, I'll merge when the tests pass.

Issue #8025, opened 2026-05-18
CLI: Unify handling of the default group


vandanayadav7:
## Description

The default group is not handled consistently:

1. Listing groups does not show the default group name as "default" — it shows empty string
2. `acr group get default` fails with `GroupNotFoundException`
3. `acr group get ""` throws a `NullPointerException`

## Expected Behavior

- Treat the default group name as "default" everywhere
- `acr group get default` should work
- Group commands (add, delete) for the default group should fail gracefully with clear messages

## Parent Epic

- #7162

jainarjav80-sys:
Hi @vandanayadav7, @carlesarnal 

Just wanted to check if you're still working on this issue. If you've moved on from it or would like help, I'd be happy to contribute.

Thanks for your time!

DSingh0304:
Hi @carlesarnal, so I went ahead and implemented a fix in PR #8668. The PR gracefully intercepts CRUD operations on the implicit "default" group to prevent server errors, and ensures "default" is properly displayed in the CLI output tables instead of blank spaces.

I would be grateful if you could review it!! Thank you for your time!

PR #8695, opened 2026-07-17
fix(version): add content-disposition header and file extension for v3 content downloads

### LINKED ISSUE #8172: Ensure Content Download honors Content MIME-TYPE and it's file extension

DSingh0304:
Fixes #8172

### Description
This PR fixes the issue where downloading artifacts via the `/content` endpoints does not automatically assign proper file extensions. It updates the v3 API endpoints (`GroupsResourceImpl` and `IdsResourceImpl`) to properly detect the MIME type of the content and set a `Content-Disposition` header with the corresponding file extension natively (e.g. `attachment; filename="content.json"`). 

**Note**: Consistent with the issue's requirements, v2 endpoints are deliberately not modified to preserve existing behaviors.

### Changes
- **`ContentTypes.java`**: Added a helper method `getFileExtension(String contentType)` to robustly map standard MIME types to standard file extensions.
- **`GroupsResourceImpl.java`**: Appends `Content-Disposition` header in `getArtifactVersionContent`.
- **`IdsResourceImpl.java`**: Appends `Content-Disposition` header in `getContentById`, `getContentByGlobalId`, and `getContentByHash`.
- **`GroupsResourceTest.java`**: Included comprehensive unit tests checking `.json` and `.yaml` file extensions on responses.


app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -944,8 +945,11 @@ public Response getArtifactVersionContent(String groupId, String artifactId, Str
         contentToReturn = handleContentReferences(references, metaData.getArtifactType(), contentToReturn,
                 artifactCell.get().getReferences());
 
+        String ext = ContentTypes.getFileExtension(contentToReturn.getContentType());
+        String filename = artifactId + ext;
         var builder = Response.ok().entity(contentToReturn.getContent())
-                .type(contentToReturn.getContentType());
+                .type(contentToReturn.getContentType())
+                .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"" + filename + "\"");
```
paoloantinori: filename = artifactId + ext is concatenated straight into the header. artifactId is caller-supplied and can contain ", /, or control chars (the test uses an id with a /). A quote in the id would break the filename="..." grammar; worth sanitizing here (strip/escape quotes and path separators) or emitting the RFC 6266 filename*=UTF-8''<encoded> form for non-token names. The numeric-id and hardcoded-"content" paths in IdsResourceImpl are not affected.

common/src/main/java/io/apicurio/registry/types/ContentTypes.java
```diff
@@ -15,4 +15,23 @@ public static boolean isEmptyContentType(String contentType) {
...
     }
 
+    /**
+     * Returns the conventional file extension for a given MIME content type.
+     * Falls back to ".bin" for unknown types.
+     */
+    public static String getFileExtension(String contentType) {
+        if (contentType == null) {
+            return ".bin";
+        }
```
paoloantinori, line 25: This switch handles JSON/YAML/XML/protobuf/graphql/thrift but not TEXT_PROMPT_TEMPLATE (text/x-prompt-template) or APPLICATION_EMPTY (application/vnd.apicurio.empty), both of which are declared above. They'll fall through to .bin. If prompt-template artifacts are downloadable via these endpoints, consider adding a case (e.g. .txt/.tpl) so they don't all land as content.bin.

paoloantinori (commented):
Clean, focused fix for #8172 — setting Content-Disposition on the v3 content endpoints is the right call, and the getFileExtension helper plus the JSON/YAML test coverage look good. A couple of observations worth double-checking before this lands.

- **artifactId flows unescaped into the Content-Disposition header.** In GroupsResourceImpl#getArtifactVersionContent the filename is built as artifactId + ext and concatenated directly into the header value. artifactId is a user-controlled string (the test itself uses testGetArtifact/EmptyAPI, which contains a /), so a value containing a double-quote or control characters would either break the quoted filename="..." production or, depending on the JAX-RS implementation's header sanitization, risk header injection. The numeric IDs in IdsResourceImpl (contentId, globalId) and the hardcoded "content" in getContentByHash are safe, but the artifactId path is the one worth hardening — e.g. sanitize/strip quotes and path separators, or emit the RFC 6266 filename*=UTF-8''<encoded> form for non-token filenames.
- **getFileExtension doesn't cover two declared content types.** TEXT_PROMPT_TEMPLATE (text/x-prompt-template) and APPLICATION_EMPTY (application/vnd.apicurio.empty) are defined in ContentTypes but fall through to the .bin default. An artifact stored as a prompt template would download as content.bin. Worth either adding cases or confirming the .bin fallback is intentional for those.
- **Minor: header value is built by string concatenation in three places.** IdsResourceImpl centralizes the prefix in ATTACHMENT_FILENAME, but GroupsResourceImpl inlines the literal "attachment; filename=\"". Pulling both through the same constant (or a small buildContentDisposition(filename) helper) would keep them from drifting and make the sanitization from the first point a single place to enforce.

Tests look reasonable; note the two new assertions pin the exact header string, so they'd also catch the sanitization change if you decide to encode non-token characters.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
DSingh0304: Fixed in 7fcae09. I've added a centralized buildContentDisposition helper in AbstractResourceImpl that sanitizes the filename (stripping quotes and path separators) to prevent injection, which both endpoints now use!

common/src/main/java/io/apicurio/registry/types/ContentTypes.java
DSingh0304: Fixed in 7fcae09. I added explicit fallback cases for both TEXT_PROMPT_TEMPLATE (.txt) and APPLICATION_EMPTY (.empty).

DSingh0304:
> Clean, focused fix for #8172 — setting Content-Disposition on the v3 content endpoints is the right call, and the getFileExtension helper plus the JSON/YAML test coverage look good. A couple of observations worth double-checking before this lands.
> 
>     * **artifactId flows unescaped into the Content-Disposition header.** In GroupsResourceImpl#getArtifactVersionContent the filename is built as artifactId + ext and concatenated directly into the header value. artifactId is a user-controlled string (the test itself uses testGetArtifact/EmptyAPI, which contains a /), so a value containing a double-quote or control characters would either break the quoted filename="..." production or, depending on the JAX-RS implementation's header sanitization, risk header injection. The numeric IDs in IdsResourceImpl (contentId, globalId) and the hardcoded "content" in getContentByHash are safe, but the artifactId path is the one worth hardening — e.g. sanitize/strip quotes and path separators, or emit the RFC 6266 filename*=UTF-8'' form for non-token filenames.
> 
>     * **getFileExtension doesn't cover two declared content types.** TEXT_PROMPT_TEMPLATE (text/x-prompt-template) and APPLICATION_EMPTY (application/vnd.apicurio.empty) are defined in ContentTypes but fall through to the .bin default. An artifact stored as a prompt template would download as content.bin. Worth either adding cases or confirming the .bin fallback is intentional for those.
> 
>     * **Minor: header value is built by string concatenation in three places.** IdsResourceImpl centralizes the prefix in ATTACHMENT_FILENAME, but GroupsResourceImpl inlines the literal "attachment; filename="". Pulling both through the same constant (or a small buildContentDisposition(filename) helper) would keep them from drifting and make the sanitization from the first point a single place to enforce.
> 
> 
> Tests look reasonable; note the two new assertions pin the exact header string, so they'd also catch the sanitization change if you decide to encode non-token characters.

Hi @paoloantinori, Thanks for the review and the great suggestions! I have addressed all the observations in the latest commit (7fcae09). I've centralized the header building to properly sanitize the artifactId against injection, and added the missing fallback cases in ContentTypes.

Let me know if everything looks good to move forward with the PR!

app/src/main/java/io/apicurio/registry/rest/v3/impl/AbstractResourceImpl.java
```diff
@@ -159,4 +159,17 @@ private static URI getApiBaseHrefFromRequest(HttpServletRequest request) throws
...
 
+    /**
+     * Builds a safe Content-Disposition header value for a given filename.
+     * Sanitizes the filename to prevent header injection.
+     */
+    protected String buildContentDisposition(String filename) {
+        if (filename == null) {
+            return "attachment; filename=\"content.bin\"";
+        }
+        // Sanitize filename: replace double quotes, path separators, and newlines with underscores
```
paoloantinori: The `/` in this character class looks inconsistent with `GroupsResourceTest#testGetArtifact`, which asserts `filename="testGetArtifact/EmptyAPI.json"` (slash preserved) for artifactId `testGetArtifact/EmptyAPI`. With this sanitizer, the `/` becomes `_`, producing `testGetArtifact_EmptyAPI.json` — so the test as written would fail. Either the `/` should be dropped from the class (it's a legal char in a quoted Content-Disposition filename per RFC 6266), or the test expectation needs updating. Worth double-checking which side is intended.

paoloantinori (commented):
The `buildContentDisposition` sanitizer added on this head addresses the header-injection point from the earlier review, which is the right direction. One inconsistency to double-check before this lands:

- **The `/` sanitizer and the `testGetArtifact` assertion look contradictory.** `buildContentDisposition` replaces `/` with `_` (`replaceAll("[\"/%\\n\\r]", "_")`), so an artifactId containing `/` (which Apicurio allows) would have its slash stripped from the filename. But `GroupsResourceTest#testGetArtifact` asserts the header preserves the slash: `equalTo("attachment; filename=\"testGetArtifact/EmptyAPI.json\"")` for artifactId `testGetArtifact/EmptyAPI`. With the sanitizer as written, the actual header would be `filename="testGetArtifact_EmptyAPI.json"`, so either the test fails or the sanitizer's `/` is unintentional. Worth confirming which of the two is the intended behavior — preserving `/` in the quoted filename is valid per RFC 6266 and matches the test, while replacing it is safer if the downloaded filename will actually be used as a filesystem path.

paoloantinori:
Left a fresh review on this head — the `buildContentDisposition` sanitizer addition addresses the header-injection point from the earlier round; one thing to double-check is the interaction between the `/` sanitizer and the `testGetArtifact/EmptyAPI` test expectation (flagged inline).

app/src/main/java/io/apicurio/registry/rest/v3/impl/AbstractResourceImpl.java
DSingh0304: Good catch fixed in [33ced9b](https://github.com/Apicurio/apicurio-registry/pull/8695/commits/33ced9b19a2683531de3ad6563d61712de33056c) ! The / in the sanitizer was unintentional. You're right that it's a valid character in a quoted Content-Disposition filename per RFC 6266, and the test expectation already confirms the slash should be preserved. Removed / from the regex in the latest commit, keeping only the genuinely dangerous characters (", %, \n, \r). Thanks!

DSingh0304:
> Left a fresh review on this head — the `buildContentDisposition` sanitizer addition addresses the header-injection point from the earlier round; one thing to double-check is the interaction between the `/` sanitizer and the `testGetArtifact/EmptyAPI` test expectation (flagged inline).

Everything should be fully resolved now! The artifactId is now sanitized against header injection (while correctly preserving / per RFC 6266), the missing fallback types in ContentTypes have been added, and the duplicate string literal warnings from SonarCloud are cleared up. All tests are green.

Let me know if everything looks good!

carlesarnal (changes requested):
The implementation adds Content-Disposition headers to content download endpoints, which improves the download experience. However, there are several issues:

Path traversal risk in `AbstractResourceImpl.buildContentDisposition` - forward slashes should be sanitized, not preserved. An artifactId containing slashes could cause browsers to save files in unexpected locations.

Inconsistent filename in `IdsResourceImpl.getContentByHash` - uses hardcoded "content" instead of the contentHash like other endpoints.

Missing null check in `IdsResourceImpl.getContentByHash` - `dto` could be null before calling `dto.getContentType()`.

Test coverage incomplete - only tests the happy path for two content types. Need tests for special characters in filenames, and tests for the other three modified endpoints.

Consider adding length validation for filenames to prevent oversized headers.

The `ContentTypes.getFileExtension` implementation looks reasonable but `.empty` extension is questionable.

DSingh0304:
> The implementation adds Content-Disposition headers to content download endpoints, which improves the download experience. However, there are several issues:
> 
> Path traversal risk in `AbstractResourceImpl.buildContentDisposition` - forward slashes should be sanitized, not preserved. An artifactId containing slashes could cause browsers to save files in unexpected locations.
> 
> Inconsistent filename in `IdsResourceImpl.getContentByHash` - uses hardcoded "content" instead of the contentHash like other endpoints.
> 
> Missing null check in `IdsResourceImpl.getContentByHash` - `dto` could be null before calling `dto.getContentType()`.
> 
> Test coverage incomplete - only tests the happy path for two content types. Need tests for special characters in filenames, and tests for the other three modified endpoints.
> 
> Consider adding length validation for filenames to prevent oversized headers.
> 
> The `ContentTypes.getFileExtension` implementation looks reasonable but `.empty` extension is questionable.

Thanks @carlesarnal  for the detailed review! I've pushed updates to address all points:

1. Sanitized slashes and added length truncation to buildContentDisposition.
2. Fixed getContentByHash to use contentHash as the filename and added the missing null check.
3. Removed the .empty extension for APPLICATION_EMPTY.
4. Added comprehensive tests for special characters, truncation, and verified headers across all endpoints.

Let me know if it's good to go!

DSingh0304:
Hi @carlesarnal @paoloantinori, any updates on this PR? Thank you for your time !!

common/src/main/java/io/apicurio/registry/types/ContentTypes.java
```diff
@@ -15,4 +15,25 @@ public static boolean isEmptyContentType(String contentType) {
...
+            case APPLICATION_JSON -> ".json";
+            case APPLICATION_YAML -> ".yaml";
+            case APPLICATION_XML -> ".xml";
+            case APPLICATION_PROTOBUF -> ".proto";
+            case APPLICATION_GRAPHQL -> ".graphql";
+            case APPLICATION_THRIFT -> ".thrift";
+            case TEXT_PROMPT_TEMPLATE -> ".txt";
+            case APPLICATION_EMPTY -> "";
+            default -> ".bin";
+        };
```
paoloantinori, line 44: This `switch` matches the bare MIME constants exactly, so a content type with parameters (e.g. `application/json; charset=utf-8`) falls through to `default -> ".bin"` and the download filename won't match the payload. Stripping the `;...` suffix (and lowercasing) before the switch would make this robust against parameterized content types.

paoloantinori (commented):
Thanks for the iteration — the `buildContentDisposition` sanitizer (stripping `"`, `/`, `\`, `%`, CR/LF) closes the header-injection / path-traversal points raised earlier, and the new `testContentDispositionSanitizationAndTruncation` test covers both the special-character and 128-char truncation paths. That's the right direction. One remaining edge case worth a look:

- **`ContentTypes.getFileExtension` does an exact match on bare MIME constants.** The `switch` matches `APPLICATION_JSON` ("application/json") etc. verbatim, with `default -> ".bin"`. If a stored content type ever carries parameters (e.g. `application/json; charset=utf-8`), it falls through to `.bin` and the `Content-Disposition` filename (e.g. `123.bin`) will no longer reflect the actual payload. Apicurio typically stores the bare type, so this is unlikely in practice, but a `contentType` of the form `type/subtype; ...` would silently mislabel the download. A cheap hardening would be to strip everything from the first `;` before the switch (and lowercase), so parameters don't change the extension.

- Minor consistency note (pre-existing, not introduced here): `getContentById` keeps `.type(ArtifactMediaTypes.BINARY)` while deriving the filename from `dto.getContentType()`, so a JSON artifact comes back as `application/octet-stream` with `filename="...json"`. `getContentByGlobalId`, by contrast, sets the real content type. Not a blocker for this PR, but the two `/ids/` endpoints now diverge more visibly because of the new filename.

Everything else (sanitizer, truncation preserving the extension, JSON/YAML test coverage) looks good to me on this head.

common/src/main/java/io/apicurio/registry/types/ContentTypes.java
DSingh0304: Thank you for the review, this has been addressed in commit [1c38598](https://github.com/Apicurio/apicurio-registry/pull/8695/commits/1c38598a9c24deb1d56cb10cd5aff57e056b010c)

DSingh0304:
> Thanks for the iteration — the `buildContentDisposition` sanitizer (stripping `"`, `/`, `\`, `%`, CR/LF) closes the header-injection / path-traversal points raised earlier, and the new `testContentDispositionSanitizationAndTruncation` test covers both the special-character and 128-char truncation paths. That's the right direction. One remaining edge case worth a look:
> 
>     * **`ContentTypes.getFileExtension` does an exact match on bare MIME constants.** The `switch` matches `APPLICATION_JSON` ("application/json") etc. verbatim, with `default -> ".bin"`. If a stored content type ever carries parameters (e.g. `application/json; charset=utf-8`), it falls through to `.bin` and the `Content-Disposition` filename (e.g. `123.bin`) will no longer reflect the actual payload. Apicurio typically stores the bare type, so this is unlikely in practice, but a `contentType` of the form `type/subtype; ...` would silently mislabel the download. A cheap hardening would be to strip everything from the first `;` before the switch (and lowercase), so parameters don't change the extension.
> 
>     * Minor consistency note (pre-existing, not introduced here): `getContentById` keeps `.type(ArtifactMediaTypes.BINARY)` while deriving the filename from `dto.getContentType()`, so a JSON artifact comes back as `application/octet-stream` with `filename="...json"`. `getContentByGlobalId`, by contrast, sets the real content type. Not a blocker for this PR, but the two `/ids/` endpoints now diverge more visibly because of the new filename.
> 
> 
> Everything else (sanitizer, truncation preserving the extension, JSON/YAML test coverage) looks good to me on this head.

Hi @paoloantinori, Thanks for the great feedback! You're absolutely right about the parameterized MIME types. I have pushed a new commit that strips any `;...` suffix and lowercases the string before evaluating it in `ContentTypes.getFileExtension()` so parameterized types won't accidentally fall through to .bin. I also went ahead and updated `getContentById` and `getContentByHash` to properly set the content type dynamically `(using dto.getContentType())` instead of the hardcoded `ArtifactMediaTypes.BINARY`, bringing them fully in line with `getContentByGlobalId`.

Everything is fully tested and should be good to go. Let me know if anything else catches your eye!

paoloantinori:
Thanks @DSingh0304 — confirming both points are addressed cleanly on this head:

- `getFileExtension` now strips the `;...` suffix and lowercases before the switch, so `application/json; charset=utf-8` resolves to `.json` instead of falling through to `.bin`. ✓
- `getContentById` and `getContentByHash` now use `dto.getContentType()` for the response `Content-Type`, so the header and the `Content-Disposition` filename stay consistent with the actual payload rather than forcing `application/octet-stream`. ✓

The sanitizer + truncation behavior is still covered by `testContentDispositionSanitizationAndTruncation`. From my side this looks good to go — nothing else jumps out.

DSingh0304:
> Thanks @DSingh0304 — confirming both points are addressed cleanly on this head:
> 
>     * `getFileExtension` now strips the `;...` suffix and lowercases before the switch, so `application/json; charset=utf-8` resolves to `.json` instead of falling through to `.bin`. ✓
> 
>     * `getContentById` and `getContentByHash` now use `dto.getContentType()` for the response `Content-Type`, so the header and the `Content-Disposition` filename stay consistent with the actual payload rather than forcing `application/octet-stream`. ✓
> 
> 
> The sanitizer + truncation behavior is still covered by `testContentDispositionSanitizationAndTruncation`. From my side this looks good to go — nothing else jumps out.

Awesome @paoloantinori, thank you for the review! Since everything looks good to go, do you know when this might get merged? I have the follow-up PR for #8025 ready on my end, and I'd love to open it as soon as this one is out of the way!

paoloantinori:
Seen, thanks @DSingh0304. I already reviewed this head (1c38598) and the sanitizer + `Content-Type` consistency look good to me. The remaining gate is @carlesarnal's re-review after the changes-requested iteration — over to him.

DSingh0304:
> Seen, thanks @DSingh0304. I already reviewed this head ([1c38598](https://github.com/Apicurio/apicurio-registry/commit/1c38598a9c24deb1d56cb10cd5aff57e056b010c)) and the sanitizer + `Content-Type` consistency look good to me. The remaining gate is @carlesarnal's re-review after the changes-requested iteration — over to him.

Sure I will wait then!!

paoloantinori (commented):
Confirmed — the sanitizer and content-type fixes are clean on this head. CI is green. From my side this is ready to merge.

DSingh0304:
> Confirmed — the sanitizer and content-type fixes are clean on this head. CI is green. From my side this is ready to merge.

Awesome, thank you! @paoloantinori

paoloantinori:
Thanks @DSingh0304 — glad the review iterations were helpful. As noted on my last review, this is ready to merge from my side. The remaining gate is @carlesarnal's re-review after the earlier changes-requested.

DSingh0304:
> Thanks @DSingh0304 — glad the review iterations were helpful. As noted on my last review, this is ready to merge from my side. The remaining gate is @carlesarnal's re-review after the earlier changes-requested.

Sure then I will wait for @carlesarnal's re review!!

DSingh0304:
Hi @carlesarnal, I have resolved the merge conflict. Please check out the PR when you have time, Thank you!!

DSingh0304:
Hi @EricWittmann @paoloantinori, all the checks are passing.

paoloantinori:
Noted — thanks for the heads-up that the checks are passing on this one. I'll take another look shortly.

DSingh0304:
> Noted — thanks for the heads-up that the checks are passing on this one. I'll take another look shortly.

Yess sure @paoloantinori. Thank you!!

paoloantinori:
No worries at all, @DSingh0304 — happy to help. As noted on my last review, the head looks clean from my side and CI is green, so this is ready to merge once @carlesarnal signs off on the changes-requested items. Nothing more needed from you on my end.

DSingh0304:
> No worries at all, @DSingh0304 — happy to help. As noted on my last review, the head looks clean from my side and CI is green, so this is ready to merge once @carlesarnal signs off on the changes-requested items. Nothing more needed from you on my end.

Okay then I will wait for @carlesarnal @EricWittmann to merge this PR.

DSingh0304:
Hi @carlesarnal @EricWittmann  I have resolved the merge conflict !!

PR #8699, opened 2026-07-17
CLI: Move conversion utility methods to Java SDK (#8644)

### LINKED ISSUE #8644: CLI: Move conversion utility methods to Java SDK
## Parent Epic

## Description
`Conversions.java` in the CLI module converts SDK client models to internal beans. These utilities could live in the Java SDK so other consumers (tests, integrations) can reuse them.

## Scope
- Identify which conversion methods in `cli/src/main/java/.../utils/Conversions.java` are generic enough for the SDK
- Move them to the appropriate Java SDK module
- Update CLI to use the SDK methods
- Ensure no breaking changes to the SDK API

## Acceptance Criteria
- Reusable conversion methods moved to the Java SDK
- CLI updated to use SDK methods instead of local copies
- No regression in CLI behavior

Ayushman-00:
Description: Closes #8644.

This PR cleans up cli/utils/Conversions.java by moving the generic, reusable conversions into the Java SDK .
What changed:

Added LabelsUtil to java-sdk/client: Handles all label parsing, key=value string splitting (with escape character support), and map conversions.
Added DateTimeConversions to java-sdk/common: Handles simple OffsetDateTime / Date / String formatting.
Updated CLI: Conversions.java now delegates to these new SDK utilities instead of duplicating the logic. Method signatures in the CLI were kept identical so no other CLI code had to change.
Added full unit test coverage for the new SDK utils (including edge cases for label escape sequences and independent date verification).
Why some methods stayed in the CLI: I intentionally left the bean-mapping methods (e.g., convert(GroupMetaData)) in the CLI. Those depend on io.apicurio.registry.rest.v3.beans.* which live in apicurio-registry-common. The Java SDK shouldn't depend on common, so moving them would mess up the dependency tree.

Tested locally and confirmed no regressions in CLI behavior.

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/utils/Conversions.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/utils/Conversions.java
@@ -1,6 +1,8 @@
 package io.apicurio.registry.cli.utils;
 
 import io.apicurio.registry.cli.common.CliException;
+import io.apicurio.registry.client.common.util.DateTimeConversions;
+import io.apicurio.registry.client.util.LabelsUtil;
 import io.apicurio.registry.rest.client.models.Labels;
 import io.apicurio.registry.rest.v3.beans.ArtifactMetaData;
 import io.apicurio.registry.rest.v3.beans.ArtifactSearchResults;
@@ -14,31 +16,31 @@ import io.apicurio.registry.rest.v3.beans.SearchedVersion;
 import io.apicurio.registry.rest.v3.beans.VersionMetaData;
 import io.apicurio.registry.rest.v3.beans.VersionSearchResults;
 import java.time.OffsetDateTime;
-import java.time.ZoneId;
-import java.time.format.DateTimeFormatter;
 import java.util.Date;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
 import static java.util.Optional.ofNullable;
 
-// TODO: Move some of these to the Java SDK.
+// The generic (non-bean-mapping) conversions have been moved to the Java SDK, see
+// io.apicurio.registry.client.util.LabelsUtil and
+// io.apicurio.registry.client.common.util.DateTimeConversions (#8644). The remaining
+// methods here map generated SDK client models to this module's internal
+// io.apicurio.registry.rest.v3.beans / io.apicurio.registry.types beans, which live in the
+// apicurio-registry-common module. Those stay in the CLI: moving them to the SDK would
+// require the (standalone, dependency-light) Java SDK to depend on apicurio-registry-common,
+// which is the wrong direction (common/app depend on the SDK, not vice versa).
 public final class Conversions {
 
-    private static final char ESCAPE_CHAR = '\\';
-    private static final char LABEL_SEPARATOR = '=';
     private static final String API_LABEL_SEPARATOR = ":";
 
     private Conversions() {
     }
 
+    // Moved to the Java SDK, see io.apicurio.registry.client.util.LabelsUtil (#8644).
     public static Map<String, String> convert(io.apicurio.registry.rest.client.models.Labels labels) {
-        return ofNullable(labels)
-                .map(Labels::getAdditionalData)
-                .map(x -> (Map<String, String>) (Map<String, ?>) x)
-                .orElse(Map.of());
+        return LabelsUtil.toMap(labels);
     }
 
     public static GroupMetaData convert(io.apicurio.registry.rest.client.models.GroupMetaData group) {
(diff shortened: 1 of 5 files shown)


cli/src/main/java/io/apicurio/registry/cli/utils/Conversions.java
```diff
@@ -178,38 +180,37 @@ public static Rule convert(io.apicurio.registry.rest.client.models.Rule rule) {
                 .build();
     }
 
+    // Moved to the Java SDK, see io.apicurio.registry.client.common.util.DateTimeConversions (#8644).
     public static String convertToString(OffsetDateTime ts) {
+
         if (ts == null) {
             return "";
         }
```
paoloantinori: Minor: this `if (ts == null)` is indented with extra whitespace (a stray tab) compared to the sibling `convertToString(OffsetDateTime)` above. Cosmetic, but worth tidying so the two methods match.

paoloantinori (commented):
Clean refactor that moves the generic label/date conversions out of `cli/utils/Conversions.java` into reusable SDK utilities (`LabelsUtil`, `DateTimeConversions`), with thorough unit-test coverage including escape-sequence edge cases and independent date verification. The decision to keep the bean-mapping methods in the CLI (to avoid the SDK depending on `apicurio-registry-common`) is well-reasoned and correctly documented in the new class header.

One minor formatting nit worth tidying before merge:

- **Indentation regression in `convertToString(Date)`.** The `if (ts == null)` block is now indented with extra whitespace (a stray tab) that differs from the rest of the method body (see `Conversions.java` ~line 187-188). Likely a merge/edit artifact — easy fix.

The rest of the delegation (signatures unchanged, logic moved verbatim) is correct and the tests (`LabelsUtilTest`, `DateTimeConversionsTest`) cover the null, empty, escaped-equals, escaped-backslash, trailing-backslash, and insertion-order cases well.

cli/src/main/java/io/apicurio/registry/cli/utils/Conversions.java
Ayushman-00: Fixed it

carlesarnal (changes requested):
The refactoring moves utility methods from CLI to Java SDK as intended. Implementation is correct and well-tested. However, the Javadoc is broken throughout both new files. All `{@...}` tags reference file paths like `.claude/agents/code-reviewer.md` and `docs/getting-started/attributes-links.adoc` instead of proper Javadoc syntax. These need to be replaced with `{@link ClassName}` or `{@code text}` as appropriate.

Fix the Javadoc before merging. The code itself is fine.

Files to fix:
- `java-sdk/client/src/main/java/io/apicurio/registry/client/util/LabelsUtil.java`
- `java-sdk/common/src/main/java/io/apicurio/registry/client/common/util/DateTimeConversions.java`

Ayushman-00:
Hi @carlesarnal, I checked the raw source on GitHub for both files — all {@link} tags reference Java classes (Labels, OffsetDateTime, Date, DateTimeUtil) and all {@code} tags wrap inline code snippets. I don't see any file path references. Could you point me to specific lines? Happy to fix anything concrete.

paoloantinori (commented):
Seen the fix and the javadoc clarification to Carles — looks good from my side. Waiting on Carles's re-review.

Ayushman-00:
@carlesarnal following up on this — I pulled the exact current source for DateTimeConversions.java to double check:

``` java
/**
 * Generic date/time conversion and formatting helpers shared by SDK consumers (CLI, tests,
 * integrations). These are intentionally simple, dependency-free conversions between
 * {@link OffsetDateTime}, {@link Date}, and {@link String}.
 * <p>
 * For parsing dates coming out of the generated REST client (including legacy date format
 * fallback), see {@link DateTimeUtil} instead.
 *
 * @see <a href="https://github.com/Apicurio/apicurio-registry/issues/8644">#8644</a>
  */
```
and the method-level docs:

```java
/**
 * @param timestamp the timestamp to convert, must not be {@code null}
 * @return the equivalent {@link Date}
 */

```
All {@link} tags point to real types (OffsetDateTime, Date, String, DateTimeUtil), {@code} wraps inline values like null, and the only other tag is a standard @see linking to issue #8644. I don't see any reference to .claude/agents/code-reviewer.md or docs/getting-started/attributes-links.adoc anywhere in either file.

Is it possible you were looking at a stale diff, a different commit, or maybe a rendering issue in the review tool you used? Happy to fix this immediately if you can point me to the exact line — just want to make sure I'm not missing something on my end.

Ayushman-00:
Rebased onto 10d03cc (current main) as 2 signed-off commits, and unified the null-handling contract between the two utilities; CLI delegations are now pure pass-throughs with no observable behaviour change. @carlesarnal the CHANGES_REQUESTED is still pinned to e297927, could you dismiss or re-review at 7a15ce94? 
@jsenko / @EricWittmann could someone approve the held workflow run and swap waiting-on-author for tested?

PR #8706, opened 2026-07-17
fix(security): add missing HSTS header to redirect responses (#8705)

### LINKED ISSUE #8705: Missing HSTS header on redirect (302) responses from RedirectFilter
## Description

**Registry Version**: 3.3.1-SNAPSHOT (build from `main`)
as discussed with @jsenko  in the pr #8604
Redirect (302) responses are missing the `Strict-Transport-Security` (HSTS) header.

`RedirectFilter` runs first in the servlet filter chain (`RedirectFilter →
RegistryApplicationServletFilter → ResourceCacheControlFilter → HSTSFilter`). When
`apicurio.redirects.enabled=true` and a request path matches a configured redirect,
`RedirectFilter.doFilter()` calls `response.sendRedirect(...)` and `return`s early —
it never calls `chain.doFilter()`, so `HSTSFilter` never runs and the 302 goes out
without the HSTS header.

This is the same class of bug fixed for disabled-API 404s in #2411, in a different filter.

### Environment

Not environment-specific. Reproduces on any Apicurio Registry instance with redirects
enabled (the default config already defines a `/` → `/apis` redirect).

### Steps to Reproduce

1. Start Apicurio Registry with the default config (redirects are enabled by default,
   with `apicurio.redirects.root=/,/apis`).
2. Request a path that matches a redirect, without following it:

malladinagarjuna2:
Fixes #8705

## Problem
`RedirectFilter` runs before `HSTSFilter` and returns early after `sendRedirect()`, so 302
responses skip the chain and go out without the `Strict-Transport-Security` header. Same
short-circuit bug as the disabled-API 404 in #2411, in a different filter.

## Fix
Apply the header before `sendRedirect()` via the existing `HSTSFilter.addHstsHeaders()`. Both
filters are in the same package, so no cross-layer coupling.

## Testing
New `HstsRedirectHeaderTest` — asserts a 302 carries the header. Red without the fix, green with
it. Checkstyle clean.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/ui/servlets/HSTSFilter.java
+++ b/app/src/main/java/io/apicurio/registry/ui/servlets/HSTSFilter.java
@@ -23,6 +23,18 @@ public class HSTSFilter implements Filter {
     }
 
     /**
+     * Resets the response but preserves the HSTS header, which {@link HttpServletResponse#reset()} would
+     * otherwise clear. Callers that reset the response to produce an error should use this instead.
+     */
+    public static void resetKeepingHstsHeader(HttpServletResponse httpResponse) {
+        String hsts = httpResponse.getHeader("Strict-Transport-Security");
+        httpResponse.reset();
+        if (hsts != null) {
+            httpResponse.setHeader("Strict-Transport-Security", hsts);
+        }
+    }
+
+    /**
      * C'tor
      */
     public HSTSFilter() {
--- a/app/src/main/java/io/apicurio/registry/rest/RegistryApplicationServletFilter.java
+++ b/app/src/main/java/io/apicurio/registry/rest/RegistryApplicationServletFilter.java
@@ -42,9 +42,9 @@ public class RegistryApplicationServletFilter implements Filter {
 
             if (disabled) {
                 HttpServletResponse httpResponse = (HttpServletResponse) response;
-                httpResponse.reset();
+                // reset() would clear the HSTS header HSTSFilter set upstream; preserve it.
+                HSTSFilter.resetKeepingHstsHeader(httpResponse);
                 httpResponse.setStatus(HttpServletResponse.SC_NOT_FOUND);
-                HSTSFilter.addHstsHeaders(httpResponse);
                 // important to return, to stop the filters chain
                 return;
             }
--- a/app/src/main/resources-unfiltered/META-INF/web.xml
+++ b/app/src/main/resources-unfiltered/META-INF/web.xml
@@ -4,6 +4,20 @@
     xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd"
     version="3.0">
 
+    <!-- HSTSFilter must run first so the HSTS header is set before any downstream filter can
+         short-circuit the chain (RedirectFilter's sendRedirect, RegistryApplicationServletFilter's 404). -->
+    <filter>
+        <filter-name>HSTSFilter</filter-name>
+        <filter-class>io.apicurio.registry.ui.servlets.HSTSFilter</filter-class>
+        <async-supported>true</async-supported>
+    </filter>
+    <filter-mapping>
+        <filter-name>HSTSFilter</filter-name>
+        <url-pattern>/*</url-pattern>
+        <dispatcher>REQUEST</dispatcher>
+        <dispatcher>FORWARD</dispatcher>
+    </filter-mapping>
+
     <filter>
         <filter-name>RedirectFilter</filter-name>
         <filter-class>io.apicurio.registry.ui.servlets.RedirectFilter</filter-class>
@@ -31,18 +45,6 @@
     </filter>
     <filter-mapping>
         <filter-name>ResourceCacheControlFilter</filter-name>
-        <url-pattern>/*</url-pattern>
-        <dispatcher>REQUEST</dispatcher>
-        <dispatcher>FORWARD</dispatcher>
-    </filter-mapping>
-
-    <filter>
-        <filter-name>HSTSFilter</filter-name>
-        <filter-class>io.apicurio.registry.ui.servlets.HSTSFilter</filter-class>
-        <async-supported>true</async-supported>
-    </filter>
-    <filter-mapping>
-        <filter-name>HSTSFilter</filter-name>
         <url-pattern>/*</url-pattern>
         <dispatcher>REQUEST</dispatcher>
         <dispatcher>FORWARD</dispatcher>
--- a/app/src/test/java/io/apicurio/registry/headers/HstsRedirectHeaderTest.java
+++ b/app/src/test/java/io/apicurio/registry/headers/HstsRedirectHeaderTest.java
@@ -1,0 +1,33 @@
+package io.apicurio.registry.headers;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.QuarkusTestProfile;
+import io.quarkus.test.junit.TestProfile;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.containsString;
+
+// RedirectFilter returns early after sendRedirect(), bypassing HSTSFilter - the 302 must still get HSTS.
+@QuarkusTest
+@TestProfile(HstsRedirectHeaderTest.RedirectProfile.class)
+public class HstsRedirectHeaderTest {
+
+    private static final String HSTS = "Strict-Transport-Security";
+
+    public static class RedirectProfile implements QuarkusTestProfile {
+        @Override
+        public Map<String, String> getConfigOverrides() {
+            return Map.of("apicurio.redirects.enabled", "true", "apicurio.redirects.hsts-test",
+                    "/hsts-redirect-source,/apis/registry/v3");
+        }
+    }
+
+    @Test
+    public void testHstsOnRedirect() {
+        given().redirects().follow(false).when().get("/hsts-redirect-source").then().statusCode(302)
+                .header(HSTS, containsString("max-age="));
+    }
+}


malladinagarjuna2:
@jsenko as discussed in the merged pr #8604 i have implemented the missing hsts  headeers to redirect rresponses

app/src/main/java/io/apicurio/registry/ui/servlets/RedirectFilter.java
```diff
@@ -70,6 +70,8 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
             }
 
             if (redirects.containsKey(servletPath)) {
+                // Returning here skips HSTSFilter, so apply the header directly.
```
paoloantinori: This fixes HSTS, but the same `return` short-circuits every filter later in the chain. If any other filter adds security headers (CSP, X-Frame-Options, etc.), those are also lost on redirect responses. Worth a follow-up to audit the full filter chain, or consider calling `chain.doFilter` with a wrapped response instead of the early return.

paoloantinori (commented):
Clean, targeted security fix — calling `HSTSFilter.addHstsHeaders(response)` before `sendRedirect()` correctly closes the short-circuit gap where the 302 skipped the rest of the filter chain, and the `HstsRedirectHeaderTest` (follow-redirects disabled, asserting the 302 carries `Strict-Transport-Security`) is the right test shape.

A couple of observations:

- **The same short-circuit may skip other security headers too.** `RedirectFilter.doFilter` returns early after every matched redirect (`sendRedirect()` + `return`), so any filter that runs *after* `RedirectFilter` in the chain and adds response headers (CSP, `X-Frame-Options`, `X-Content-Type-Options`, etc.) is also bypassed on redirect responses — not just HSTS. This PR fixes HSTS specifically by reaching into `HSTSFilter` directly, which is correct and minimal, but worth tracking whether the other security headers need the same treatment or whether the root cause (the early `return` before `chain.doFilter`) should be addressed more broadly. Not a blocker for this PR.

- **Worth double-checking that `addHstsHeaders` respects the HSTS-enabled config.** If a deployment has HSTS intentionally disabled (`apicurio.hsts.enabled=false`), the redirect path should also omit the header. Since `addHstsHeaders` is the same method `HSTSFilter` calls in its normal flow, it presumably checks the enable flag internally — but worth confirming that's the case rather than an unconditional header write.

app/src/main/java/io/apicurio/registry/ui/servlets/RedirectFilter.java
malladinagarjuna2: @paoloantinori thanks for the review!

On the enable-flag point: there's no apicurio.hsts.enabled (or any HSTS gate); HSTSFilter.addHstsHeaders() writes the header unconditionally, and the filter itself is registered unconditionally in web.xml. So the redirect path is consistent with the existing normal flow, both write unconditionally, there's no flag to respect. If you think HSTS should be made toggleable, that's a pre-existing gap (it'd need a flag gating all call sites), we can file it as a seperate issue
On the broader point: agreed that  the same early return drops every downstream filter, so CSP / X-Frame-Options / X-Content-Type-Options are bypassed on redirects too. The real root cause is the short-circuit pattern; the cleaner fix is continuing the chain with a wrapped response instead of each filter re-applying headers. can i open  a issue on full chain audit ?

malladinagarjuna2:
hey @paoloantinori  could you please tell me what to do ?

paoloantinori:
Hi @malladinagarjuna2 — from a technical standpoint the PR looks good: I already reviewed it positively and the change correctly closes the HSTS short-circuit on redirect responses, with a test that fails without the fix.

The blocker right now is procedural, not technical. The `/ready` command failed because the bot expects the PR to be in `lifecycle/wip` state, but it's currently in `lifecycle/new`. That state transition is typically applied by a maintainer during triage — you can't move it yourself.

Since you were already discussing this with @jsenko in #8604, I'd suggest pinging him directly so he can move the PR into the right lifecycle state and queue it for merge. There's nothing else for you to change in the code itself as far as I can see.

malladinagarjuna2:
thanks for the review @paoloantinori , i think we should file the observations discovered by yu as a seperate issue, 
@jsenko  could you please give a further direction for this ?

paoloantinori:
Hi @malladinagarjuna2 — agreed, filing the unrelated SonarCloud/Yu observations as a separate issue is the right call: it keeps this PR scoped to the HSTS short-circuit fix (which is ready to go from my side) and lets the other findings be triaged on their own merits.

@jsenko — over to you for the merge / further direction.

carlesarnal (changes requested):
The fix addresses a real security gap where redirect responses bypassed HSTS headers. However, the test has a malformed annotation that will prevent compilation. Line with `@app/src/test/java/io/apicurio/registry/auth/A2AAuthTestProfile.java(HstsRedirectHeaderTest.RedirectProfile.class)` should be `@TestProfile(HstsRedirectHeaderTest.RedirectProfile.class)`.

The test coverage is minimal. Add tests for: HSTS disabled config, exception handling in `addHstsHeaders()`, and verify non-redirect paths still work. The test assertion should validate the complete HSTS header value, not just presence of "max-age=".

Verify that `HSTSFilter.addHstsHeaders()` is a public static method and that calling it directly doesn't break filter lifecycle assumptions.

`app/src/test/java/io/apicurio/registry/headers/HstsRedirectHeaderTest.java` - fix annotation syntax.

jsenko:
I've reviewed this more closely, and I'd like to suggest modifying the approach to address @paoloantinori's concern about other filters potentially having the same short-circuit problem.

The current fix patches each early-return site with manual `addHstsHeaders()` calls. This works, but it creates cross-filter coupling and doesn't scale — every future early return in any filter would need to remember to add the call.

I'd recommend a combined approach:

**1. Move `HSTSFilter` first in `web.xml`** (before `RedirectFilter`)

Since `HSTSFilter` sets the header *before* calling `chain.doFilter()`, the HSTS header will already be on the response before any downstream filter runs. Per the servlet spec, `sendRedirect()` preserves existing headers — it only clears the body buffer. So the redirect case is fixed automatically, and any future early return in the chain is also covered. The manual `addHstsHeaders()` call in `RedirectFilter` can be removed.

**2. Add a utility method for "safe reset"** (e.g. in `HSTSFilter` or a shared util)

The one case filter reordering can't fix is `httpResponse.reset()` in `RegistryApplicationServletFilter` — `reset()` explicitly clears all headers. A utility method that preserves security headers across a reset centralizes this concern:

```java
public static void resetKeepingSecurityHeaders(HttpServletResponse response) {
    String hsts = response.getHeader("Strict-Transport-Security");
    response.reset();
    if (hsts != null) {
        response.setHeader("Strict-Transport-Security", hsts);
    }
}
```

Since `reset()` is only called in one place in the codebase, this is minimal code. If more security headers are added later (CSP, X-Frame-Options), there's a single place to update.

**Alternative**: a `HttpServletResponseWrapper` that overrides `reset()` to preserve whitelisted headers transparently. This would be more robust (downstream filters don't need to know about it), but it's over-engineered for one `reset()` call site. Worth keeping in mind if more cases appear.

The tests in the current PR are solid and would work with either approach — they verify the header is present on all response types regardless of how it gets there.

jsenko:
Could you give me more details in this please?

> the observations discovered by yu

paoloantinori:
+1 to @jsenko's combined approach. Moving `HSTSFilter` ahead of `RedirectFilter` in `web.xml` makes the header injection structural rather than call-site-by-call-site, which is the right invariant: any future early-return added to any filter downstream of `HSTSFilter` will automatically inherit the header, instead of silently regressing like the original redirect bug did.

Keeping the targeted `addHstsHeaders()` call at the redirect site as defence-in-depth is fine, but the filter ordering is the load-bearing change. If maintaining both, a short comment at the redirect site noting "HSTSFilter (now first) already applies the header; this call is belt-and-suspenders" would prevent the next reader from being confused about which one is authoritative.

Happy to let @jsenko and @malladinagarjuna2 drive the implementation here — no further asks from my side.

paoloantinori (commented):
@jsenko's approach is better than what I originally raised — moving `HSTSFilter` first in `web.xml` so the header is set before any downstream filter runs is cleaner than patching each early-return site. The `resetKeepingSecurityHeaders` utility for the single `reset()` call in `RegistryApplicationServletFilter` is proportionate.

The "observations" I mentioned were about other filters in the chain that could short-circuit before HSTS got applied — jsenko's filter-ordering proposal addresses exactly that, systematically rather than site-by-site.

@malladinagarjuna2 I'd follow jsenko's direction here: reorder the filter in `web.xml` and add the reset utility. The existing tests should still pass since they verify the header is present regardless of mechanism.

jsenko:
@malladinagarjuna2 Please rebase the PR, the #8676 was merged.

malladinagarjuna2:
hey @jsenko  
i am back 
sorry for the delay

malladinagarjuna2:
will do it now

malladinagarjuna2:
hey @jsenko i have applied your combined changes,
@paoloantinori @carlesarnal   one thing worth noting on why both pieces are needed : it come down to servlet spec assymetry sendredirect() preserves existing headers, but reset() explicitly clears them (so the disabled -API path still needs resetKeepingSecuriotyHeaders()) .The two mechanisms together cover both short-circuit types. the redirect test is red withHSTS reordered but reset() un-guarded and vice versa

paoloantinori:
Thanks @malladinagarjuna2 — that explanation is exactly right and the combined approach is the correct invariant. The servlet-spec asymmetry you describe checks out against the diff:

- `sendRedirect()` preserves existing response headers, so once `HSTSFilter` runs first (reordered in `web.xml`), the header it set upstream survives `RedirectFilter`'s redirect. ✓
- `HttpServletResponse#reset()` explicitly clears headers, so the disabled-API 404 path still needs the new `resetKeepingSecurityHeaders()` to capture HSTS, reset, and re-apply it — the old `addHstsHeaders()` re-add couldn't be removed without losing coverage on that path. ✓

So both pieces are genuinely complementary, not redundant, and your "red with only one of the two" observation confirms each short-circuit is independently covered.

One minor, non-blocking note for the future: `resetKeepingSecurityHeaders` only restores `Strict-Transport-Security`, even though the name reads as plural/generic. That's fine today since `HSTSFilter` only sets HSTS, but if other security headers (CSP, X-Frame-Options, etc.) ever get added upstream, the helper would silently drop them. Worth keeping in mind; no change needed for this PR.

LGTM from my side.

malladinagarjuna2:
@jsenko  kindly review the pr

app/src/main/java/io/apicurio/registry/ui/servlets/HSTSFilter.java
```diff
@@ -22,6 +22,18 @@ public static void addHstsHeaders(HttpServletResponse httpResponse) {
         httpResponse.setHeader("Strict-Transport-Security", HSTS_HEADER);
     }
 
+    /**
+     * Resets the response but preserves security headers, which {@link HttpServletResponse#reset()} would
+     * otherwise clear. Callers that reset the response to produce an error should use this instead.
+     */
+    public static void resetKeepingSecurityHeaders(HttpServletResponse httpResponse) {
```
paoloantinori: The name `resetKeepingSecurityHeaders` implies it preserves any security header, but the implementation only saves/restores `Strict-Transport-Security`. That's fine today since HSTSFilter is the sole upstream header source, but if another defensive header (CSP, X-Content-Type-Options, etc.) is ever set upstream of `RegistryApplicationServletFilter`, this 404 path will silently drop it. Consider renaming to `resetKeepingHstsHeader` to match the actual scope, or generalizing the save/restore over a fixed list of security headers.

paoloantinori (commented):
Re-read at the new head after the rebase. The combined approach — `HSTSFilter` promoted to the top of the chain in `web.xml` plus the `resetKeepingSecurityHeaders` helper for the disabled-API 404 path — is the correct invariant, and the new `HstsRedirectHeaderTest` covers the redirect case that motivated #8705.

One forward-looking concern on the new helper:

- `HSTSFilter.resetKeepingSecurityHeaders` ([source](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/ui/servlets/HSTSFilter.java)) reads back only `Strict-Transport-Security` around the `reset()`, but its name promises to preserve "security headers" generically. That's accurate *today* because `HSTSFilter` is the only thing adding headers upstream of this call, but it's a latent footgun: if `X-Content-Type-Options`, `Content-Security-Policy`, or any other defensive header ever gets added upstream of `RegistryApplicationServletFilter`, the disabled-API 404 path will silently strip it while every other response carries it. Worth either renaming to `resetKeepingHstsHeader` (honest about scope) or generalizing the save/restore over a known set of security headers. Minor, but the misleading name is the kind of thing that bites later.

Otherwise this looks good to land.

app/src/main/java/io/apicurio/registry/ui/servlets/HSTSFilter.java
malladinagarjuna2: Thanks for the review  @paoloantinori  renamed resetKeepingSecurityHeaders → resetKeepingHstsHeader so the name matches what it does (preserves only HSTS today). Went with the rename over generalizing since nothing else sets a header upstream yet.

malladinagarjuna2:
@paoloantinori  well the helper becomes the right place to generalize the moment a second upstream security header is added. The nosniff (X-Content-Type-Options) work I had open earlier is the likely first case, I closed it to keep to one PR at a time, and when I resubmit it after this lands, I'll generalize the helper there with a test on the disabled-API path so it can't silently regress.

I have a question: should i  generalize to a small header list now, or keep it HSTS-only until there's a real second header?

paoloantinori:
Hi @malladinagarjuna2 — my vote is to **keep it HSTS-only here** and let the nosniff PR own the generalization. A few reasons:

- **No second caller exists yet.** Generalizing `resetKeepingHstsHeader` now means designing the API shape against one hypothetical use case. You'd be guessing at the right signature before the real second caller exists, and it's hard to undo if the nosniff work reveals a constraint you didn't anticipate.
- **The generalization lands stronger with two real call sites.** When you resubmit the nosniff work, that PR can refactor `resetKeepingHstsHeader` into a header-list helper *and* add the disabled-API-path regression test you mentioned — both validated against actual usage rather than a speculative list.
- **It keeps this PR minimal and mergeable.** The HSTS short-circuit fix is the scope here, and it's ready. Folding in a generalization that nothing currently uses risks reopening the review for a change with no immediate payoff.

This is also exactly the sequencing you proposed in your earlier comment, so I'd just say: go with that. Land HSTS-only, generalize in nosniff.

One small note for when you do generalize: the current helper reads the header by string name (`"Strict-Transport-Security"`). A list-based version should probably take the set of header names as a parameter rather than hardcoding them inside the helper, so callers stay in control of what gets preserved. Easy to get right when the second caller is in front of you.

malladinagarjuna2:
Thanks @paoloantinori  landing HSTS-only here. Good note on the API shape: when I generalize in the nosniff PR I'll have the helper take the header names as a parameter rather than hardcoding them, so callers stay in control of what's preserved. Ready on my end.

malladinagarjuna2:
@paoloantinori @EricWittmann @jsenko @carlesarnal  i think the changes are done so could you please merge the pr

paoloantinori:
Hi @malladinagarjuna2 — I've reviewed the technical content positively across the revisions, and the combined approach (HSTSFilter promoted ahead of RedirectFilter in `web.xml`, plus `resetKeepingHstsHeader` for the disabled-API 404 path) is the right invariant. From my side the code is good to go.

Merging is outside what I do in reviews, though — that call belongs to a maintainer with merge authority (@jsenko @carlesarnal @EricWittmann). The PR is already in `ready-for-review`, so it just needs one of them to approve and merge.

EricWittmann:
Thanks @malladinagarjuna2 :+1:

malladinagarjuna2:
thanks @EricWittmann

PR #8716, opened 2026-07-18
fix(go-sdk): instantly fail tests when registry jar is missing

### LINKED ISSUE #8715: [Go SDK] Test suite hangs silently when registry runner jar is missing
## Description

**Registry Version**: 3.3.1-SNAPSHOT
**Persistence type**: sql

While exploring the Go SDK tests (`go-sdk/pkg/tests/client_test.go`), I noticed an issue that causes brittleness and silent test failures:

**Silent 10-Second Test Hang:** If the Java backend hasn't been built yet locally, `setupSuite` sets `registryJar` to `"not-found"` and blindly attempts to execute `java -jar not-found`. This causes the test suite to hang for 10 seconds in the retry loop before failing cryptically. 

### Environment

- **OS**: Windows / Linux
- **Component**: Go SDK `v3` (`github.com/apicurio/apicurio-registry/go-sdk/v3`)

### Steps to Reproduce

1. Clone the repository and navigate to the `go-sdk` directory.
2. Ensure you have not built the Java project yet (or run `mvn clean`).
3. Run `go test ./...` in the `go-sdk` folder.
4. Observe the silent 10-second hang and the cryptic Java error.

**Expected:** 
The test suite should fail instantly with a clear error message (e.g., `t.Fatalf("Could not find registry runner jar...")`) if the compiled Java jar doesn't exist.

**Actual:** 
The test hangs silently on missing dependencies as it attempts to run a non-existent jar file.

### Logs

```text
Error: Unable to access jarfile D:\apicurio-registry\go-sdk\pkg\tests/../../../app/target/not-found
    client_test.go:52: Checking Registry is started ...
    client_test.go:58: Registry not started yet ...

### PR DESCRIPTION
**Pull request by Psykii22**
### Describe the bug

Fixes #8715 

Currently, the Go SDK test setup (`setupSuite`) assumes a very specific local environment and lacks prerequisite validation, leading to brittle test execution. Specifically:
- It assumes the developer is using the Quarkus 2 layout (`*-runner.jar`) and fails to locate the runner jar in newer Quarkus 3 builds (`quarkus-app/quarkus-run.jar`).
- It blindly attempts to execute `java -jar` without checking if Java is actually installed or in the user's `PATH`.
- It attempts to start a new registry instance even if the developer already has one running locally (e.g. via Docker).
- It leaks HTTP connections in the retry loop due to `defer resp.Body.Close()`.
- If the Java process crashes instantly, or if the server never starts, the test continues and eventually panics due to nil pointer dereferences instead of failing gracefully.

### Proposed Solution

This PR completely overhauls the test initialization to make it robust, safe, and heavily resilient across supported environments:

1. **Reuse Existing Instances:** The setup now pings the registry URL first. If a 200 OK is received, it uses the existing registry and skips the teardown entirely.
2. **Prerequisite Validation:** Adds `exec.LookPath("java")` to ensure Java is installed before attempting to start the server.
3. **Cross-Version Quarkus Support:** Implements explicit checks for the Quarkus 3 layout (`quarkus-app/quarkus-run.jar`), gracefully falling back to a `filepath.Walk` regex search for Quarkus 2 jars if not found.
4. **Immediate Crash Detection:** Uses a goroutine and `cmd.Wait()` with a channel to check if the Java process instantly crashes (e.g. corrupt jar), failing the test immediately in under `0.20s` instead of hanging.
5. **Memory Leak Fix:** Replaces the deferred body close with an immediate `resp.Body.Close()` inside the retry loop.
6. **Stop Cascading Panics:** Adds a `started` flag to the retry loop. If the server fails to start after 10 seconds, it explicitly calls `t.Fatalf` to halt the test, preventing downstream panics.

### DIFF
--- a/go-sdk/pkg/tests/client_test.go
+++ b/go-sdk/pkg/tests/client_test.go
@@ -25,49 +25,110 @@ const RegistryPort = "8080"
 var RegistryUrl = fmt.Sprintf("http://%s:%s/apis/registry/v3", RegistryHost, RegistryPort)
 
 func setupSuite(t *testing.T) func(t *testing.T) {
-	t.Log("Starting Registry")
-
-	pwd, err := os.Getwd()
-	assert.Nil(t, err)
-	libRegEx, err := regexp.Compile(`^.+-runner.(jar)$`)
-	assert.Nil(t, err)
-
-	registryJar := "not-found"
-	err = filepath.Walk(fmt.Sprintf("%s/../../../app/target/", pwd), func(path string, info os.FileInfo, err error) error {
-		if err == nil && libRegEx.MatchString(info.Name()) {
-			registryJar = info.Name()
+	httpClient := &http.Client{Timeout: 2 * time.Second}
+	resp, err := httpClient.Get(RegistryUrl)
+	if err == nil {
+		resp.Body.Close()
+		if resp.StatusCode == 200 {
+			t.Log("Registry already running")
+			return func(t *testing.T) {
+				t.Log("Registry is already running, skipping teardown")
+			}
 		}
-		return nil
-	})
-	assert.Nil(t, err)
+	}
 
-	cmd := exec.Command("java", "-jar", fmt.Sprintf("%s/../../../app/target/%s", pwd, registryJar))
+	t.Log("Starting Registry")
+	if _, err := exec.LookPath("java"); err != nil {
+		t.Fatalf("Java is not installed or not in PATH")
+	}
+	pwd, err := os.Getwd()
+	if err != nil {
+		t.Fatalf("failed to get working directory: %v", err)
+	}
+
+	targetDir := filepath.Join(pwd, "..", "..", "..", "app", "target")
+
+	// 1. Check for Quarkus 3 layout first
+	registryJarPath := filepath.Join(targetDir, "quarkus-app", "quarkus-run.jar")
+	if _, err := os.Stat(registryJarPath); os.IsNotExist(err) {
+
+		// 2. Fallback to Quarkus 2 layout (*-runner.jar)
+		registryJarPath = ""
+		libRegEx := regexp.MustCompile(`^.+-runner\.jar$`)
+		err := filepath.Walk(targetDir, func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if !info.IsDir() && libRegEx.MatchString(info.Name()) {
+				registryJarPath = path
+			}
+			return nil
+		})
+		if err != nil && !os.IsNotExist(err) {
+			t.Fatalf("Failed to search for registry jar: %v", err)
+		}
+	}
+
+	if registryJarPath == "" {
+		t.Fatalf("Could not find registry runner jar. Did you build the java project?")
+	}
+
+	cmd := exec.Command("java", "-jar", registryJarPath)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	err = cmd.Start()
-	assert.Nil(t, err)
+	if err != nil {
+		t.Fatalf("failed to start registry: %v", err)
+	}
 
+	exitChan := make(chan error, 1)
+	// This goroutine lives for the entire test suite to monitor the process lifecycle.
+	go func() {
+		exitChan <- cmd.Wait()
+	}()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 250*time.Millisecond)
+	defer cancel()
+
+	// Wait a split second to see if Java instantly crashes (e.g. bad jar)
+	select {
+	case <-ctx.Done():
+		// Still running, all good
+	case <-exitChan:
+		t.Fatalf("Registry process exited immediately with an error. Is the jar valid?")
+	}
+
+	started := false
 	retries := 10
+	// 2-second HTTP timeout means this could wait up to 20 seconds total.
 	for i := 0; i < retries; i++ {
 		t.Log("Checking Registry is started ...")
 		req, err := http.NewRequest("GET", RegistryUrl, nil)
-		assert.Nil(t, err)
+		if err != nil {
+			t.Fatalf("failed to create request: %v", err)
+		}
 
-		resp, err := http.DefaultClient.Do(req)
+		resp, err := httpClient.Do(req)
 		if err != nil {
 			t.Log("Registry not started yet ...")
 			time.Sleep(1 * time.Second)
 		} else {
-			defer resp.Body.Close()
-
 			if resp.StatusCode == 200 {
+				resp.Body.Close()
 				t.Log("Registry Started!")
+				started = true
 				break
 			} else {
+				resp.Body.Close()
 				t.Log("Registry not started yet ...")
 				time.Sleep(1 * time.Second)
 			}
 		}
+	}
+
+	if !started {
+		cmd.Process.Kill()
+		t.Fatalf("Registry failed to start after 10 seconds")
 	}
 
 	return func(t *testing.T) {

### CODE REVIEW

**Review by Nachiket-Roy — CHANGES_REQUESTED** [review]
I think this check only addresses one failure mode (missing runner JAR), but the underlying issue is broader: the test setup assumes a very specific local environment and doesn't validate its prerequisites before proceeding.

A more robust approach would be to make the test initialization fail fast after validating all required preconditions rather than checking only for a missing JAR. For example:

* Support both Quarkus 2 (`*-runner.jar`) and Quarkus 3 (`quarkus-app/quarkus-run.jar`) layouts instead of assuming a single packaging format.
* Verify that the required runtime (Java) is available before attempting to start the registry.
* Reuse an already-running local registry instance when available instead of always launching a new one.
* Abort immediately if startup fails, rather than continuing into the retry loop and teardown where subsequent nil dereferences or other cascading failures can occur.
* Close HTTP response bodies immediately inside the retry loop instead of deferring them until the test exits.

This makes the test resilient across supported build layouts, provides clearer error messages for missing prerequisites, avoids unnecessary startup work, and prevents cascading failures from partially initialized state.

**Comment by Psykii22** [comment]
great suggestion i added those tests.

File: go-sdk/pkg/tests/client_test.go [hunk 1]
```diff
@@ -25,28 +25,65 @@ const RegistryPort = "8080"
...
+	resp, err := http.Get(RegistryUrl)
+	if err == nil {
+		resp.Body.Close()
+		if resp.StatusCode == 200 {
+			t.Log("Registry already running")
+			return func(t *testing.T) {
+				t.Log("Registry is already running, skipping teardown")
+			}
+		}
+	}
```
paoloantinori, line 38: Typo: `instlled` → `installed`. Minor, but this is the message a contributor sees when Java isn't on PATH, so worth getting right.

go-sdk/pkg/tests/client_test.go
```diff
@@ -25,28 +25,65 @@ const RegistryPort = "8080"
...
+
+	if registryJarPath == "" {
+		t.Fatalf("Could not find registry runner jar. Did you build the java project?")
+	}
 
-	cmd := exec.Command("java", "-jar", fmt.Sprintf("%s/../../../app/target/%s", pwd, registryJar))
+	cmd := exec.Command("java", "-jar", registryJarPath)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	err = cmd.Start()
```
paoloantinori, line 79: Two things to double-check here:
1. On the success path (process keeps running), the `go func() { exitChan <- cmd.Wait() }()` goroutine blocks forever sending into an unbuffered channel — that goroutine leaks for the lifetime of the test process. Buffering the channel (`make(chan error, 1)`) lets the sender complete and the goroutine exit.
2. The `time.After(250ms)` timer isn't cancelled when the `exitChan` arm fires; the timer goroutine lives until it fires. Minor for a 250ms window but easy to avoid with a `context.WithTimeout` + defer cancel.

go-sdk/pkg/tests/client_test.go
```diff
@@ -25,28 +25,65 @@ const RegistryPort = "8080"
...
-			registryJar = info.Name()
-		}
-		return nil
-	})
-	assert.Nil(t, err)
+	targetDir := filepath.Join(pwd, "..", "..", "..", "app", "target")
+	
+	// 1. Check for Quarkus 3 layout first
+	registryJarPath := filepath.Join(targetDir, "quarkus-app", "quarkus-run.jar")
+	if _, err := os.Stat(registryJarPath); os.IsNotExist(err) {
```
paoloantinori, line 53: Worth surfacing walk errors here. Currently an unreadable `target/` directory (permission denied, etc.) is swallowed and the test reports "Could not find registry runner jar" instead of the real cause. Returning the error from the callback and failing setup with that message would make the failure mode debuggable.

go-sdk/pkg/tests/client_test.go
```diff
@@ -25,28 +25,65 @@ const RegistryPort = "8080"
...
 
 func setupSuite(t *testing.T) func(t *testing.T) {
-	t.Log("Starting Registry")
+	resp, err := http.Get(RegistryUrl)
+	if err == nil {
+		resp.Body.Close()
+		if resp.StatusCode == 200 {
+			t.Log("Registry already running")
+			return func(t *testing.T) {
+				t.Log("Registry is already running, skipping teardown")
```
paoloantinori, line 35: `resp.Body.Close()` only runs on the `StatusCode == 200` branch. If the existing server returns non-200, the body leaks. A `defer resp.Body.Close()` right after the successful `err == nil` check covers both branches cleanly.

paoloantinori (commented):
Solid direction — the Quarkus 3 layout detection, the early `exec.LookPath("java")`, the registry-already-running fast path, and the immediate-crash detection via `cmd.Wait()` are all genuinely useful for CI and local dev. Beyond what Nachiket-Roy already flagged, a few correctness/robustness nits worth addressing before merge:

- **Typo in a user-facing message**: `"Java is not instlled or not in PATH"` (missing `a` in "installed") — it's also the failure message a contributor sees, so worth fixing.
- **Leaked timer in the crash-detection `select`**: the `time.After(250ms)` arm is never cancelled when the `exitChan` arm fires, and the goroutine feeding `exitChan` (`go func() { exitChan <- cmd.Wait() }()`) is left blocked forever if the process stays alive — i.e. on every successful run. Either close-after-receive or pull the exit status through a buffered channel / `context` so the goroutine can exit.
- **`filepath.Walk` errors are silently dropped**: the walk callback ignores the `err` passed in and never surfaces walk-level failures (e.g. unreadable `target/`). Returning it and failing the setup makes the "jar not found" path less mysterious.
- **Resource leak on the pre-check**: in the "reuse existing instance" `http.Get` block, `resp.Body` is only closed on the `StatusCode == 200` path; if the server is up but returns non-200, the body is leaked. `defer resp.Body.Close()` immediately after the successful `Get` covers both branches.
- **Process cleanup on `t.Fatalf`**: after `cmd.Start()` succeeds, several `t.Fatalf` calls (`jar not found` is before start, but `exited immediately` and `failed to start after 10s`) leave the java process orphaned if `cmd.Wait()` already returned. The `failed to start` path does call `cmd.Process.Kill()`, which is good — applying the same guard to the `exited immediately` path (or relying on `cmd.Wait()` having already reaped it) would make teardown consistent.

Minor, but these are the kind of thing that bites the next contributor running these tests locally.

Psykii22:
I have addressed all the issues you pointed out

Psykii22:
@paoloantinori can you take a look at this ,whenever you have time

go-sdk/pkg/tests/client_test.go
```diff
@@ -25,28 +25,74 @@ const RegistryPort = "8080"
...
-	})
-	assert.Nil(t, err)
+	}
 
-	cmd := exec.Command("java", "-jar", fmt.Sprintf("%s/../../../app/target/%s", pwd, registryJar))
+	if registryJarPath == "" {
+		t.Fatalf("Could not find registry runner jar. Did you build the java project?")
+	}
+
+	cmd := exec.Command("java", "-jar", registryJarPath)
```
paoloantinori, line 76: `assert.Nil(t, err)` does not stop execution on failure — if `cmd.Start()` errors, execution falls through to `cmd.Wait()` on a never-started process and you get a nil-pointer panic. Prefer `t.Fatalf("failed to start registry: %v", err)` so the suite fails cleanly. Same reasoning applies to the `os.Getwd()` assert above.

go-sdk/pkg/tests/client_test.go
```diff
@@ -25,28 +25,74 @@ const RegistryPort = "8080"
 var RegistryUrl = fmt.Sprintf("http://%s:%s/apis/registry/v3", RegistryHost, RegistryPort)
 
 func setupSuite(t *testing.T) func(t *testing.T) {
-	t.Log("Starting Registry")
+	resp, err := http.Get(RegistryUrl)
+	if err == nil {
+		defer resp.Body.Close()
+		if resp.StatusCode == 200 {
+			t.Log("Registry already running")
```
paoloantinori, line 33: `http.Get` uses the default client with no timeout. If the port is held half-open or intercepted by a proxy, the suite will hang instead of failing. Wrap with an explicit `&http.Client{Timeout: 2 * time.Second}` here and in the polling loop.

paoloantinori (commented):
Thanks @Psykii22 — the new shape looks good: the early ping, `exec.LookPath`, cross-version Quarkus detection, the 250 ms instant-crash guard, and the explicit `started` flag + cleanup on failure all address the earlier feedback cleanly. Two more things worth tightening before this lands:

- **`cmd.Start()` failure path can panic.** After `cmd.Start()` you use `assert.Nil(t, err)`, but `assert.Nil` does **not** stop execution on failure — if `Start` fails, control falls through to `cmd.Wait()` on a process that was never started, and you get a nil-pointer panic instead of a clean `FailNow`. Use `t.Fatalf("failed to start registry: %v", err)` (or `require.Nil`) so the test fails gracefully. The same applies to the existing `assert.Nil(t, err)` after `os.Getwd()`.
- **`http.Get` has no timeout.** Both the initial ping (`resp, err := http.Get(RegistryUrl)`) and the polling loop call `http.Get` with the default client, which has no timeout. If something is holding the port half-open (or a misconfigured proxy intercepts the request), the suite can hang indefinitely rather than failing. A small `&http.Client{Timeout: 2 * time.Second}` makes the whole startup path bounded.

Minor nit: `cmd.Process.Kill()` in the failure branch and in the returned teardown is never followed by `cmd.Wait()`, which can leak the reaping goroutine you started earlier; a `cmd.Wait()` after the kill (ignoring the error) closes that out.

paoloantinori:
Sure @Psykii22 — took another look. The changes address the earlier feedback well; I left two more notes in the review (the `assert.Nil` vs `t.Fatalf` panic path after `cmd.Start()`, and missing timeout on `http.Get`). No blockers, just robustness.

carlesarnal (changes requested):
The change improves test reliability and error messages. Good additions: Java check, Quarkus 3 support, early crash detection, explicit jar-missing error.

Fix the resource leak in lines 3-11. When checking if registry is already running, you close `resp.Body` only in the success path. If status is not 200, the body leaks. Move `defer resp.Body.Close()` outside the status check.

Consider: The 2-second HTTP timeout in the polling loop means total wait could be 20 seconds, not 10. Either document this or adjust the retry logic.

Minor: The `exitChan` goroutine lives for the entire test suite. Not a bug, but worth a comment explaining it monitors the process lifecycle.

The Quarkus 2 fallback logic is solid. The filepath.Walk error handling is correct.

paoloantinori (commented):
Seen the latest push — SonarQube passed. Waiting on maintainer `/accept` to trigger CI.

carlesarnal (dismissed):
AI Code Review - Approved. All previously flagged maintainer issues have been addressed: resource leak on pre-check fixed (resp.Body.Close before status check), HTTP timeout added (2s client used throughout), assert.Nil replaced with t.Fatalf for fail-fast, filepath.Walk errors surfaced, response body management fixed (immediate close, no deferred leak), 20-second total wait documented, exitChan goroutine documented, buffered exitChan prevents goroutine leak, context.WithTimeout with defer cancel prevents leaked timer. The setupSuite function is now robust: reuses existing registry instances, validates Java via exec.LookPath, supports Quarkus 3 and 2 jar layouts, detects immediate crashes (250ms guard), prevents cascading nil-pointer panics (started flag), proper cleanup on all failure paths. Minor non-blocking nit: line ~119 missing space after // in comment. All CI checks pass.

Psykii22:
@carlesarnal it needs something?

EricWittmann:
Thanks for the contribution @Psykii22! Great job iterating through the review feedback — the test setup is much more robust now.

PR #8717, opened 2026-07-18
fix(search): bind version id filters as numbers

### LINKED ISSUE #8698: bug: searchVersions binds globalId/contentId as String instead of Long
---
name: Bug report
about: Use this template to report a bug in this project
title: ""
labels: Bug
---

## Description

**Registry Version**: snapshot (confirmed at commit `acb5abcc`)
**Persistence type**: sql

In `searchVersions()`, the `contentId` and `globalId` filter cases fall into a shared `switch` arm alongside `state` and `version` (lines 278–291 of `SqlSearchRepository.java`). That arm always calls `filter.getStringValue()` when binding the query parameter. For `state` and `version`, whose values are strings, this is correct. For `contentId` and `globalId`, whose values are `Long` (stored via `SearchFilter.ofContentId(Long)` and `SearchFilter.ofGlobalId(Long)`), `getStringValue()` returns the string representation of the number (e.g. `"42"`), which is then bound as a `VARCHAR`. Because the corresponding database columns (`v.contentId`, `v.globalId`) are `BIGINT`, PostgreSQL rejects the comparison with the error:

```
operator does not exist: bigint = character varying
```

The inconsistency is visible by comparing the two switch arms. `searchArtifacts()` already handles this correctly at lines 144 and 154 by calling `filter.getNumberValue().longValue()` for `globalId` and `contentId`. The `searchVersions()` equivalent does not apply the same pattern.

**Buggy code** (lines 278–291):

```java
case contentId:
case globalId:
case state:
case version:
    op = filter.isNot() ? "!=" : "=";
    where.append("v.");
    where.append(filter.getType().name());
    where.append(" ");
    where.append(op);
    where.append(" ?");
    binders.add((query, idx) -> {
        query.bind(idx, filter.getStringValue()); // ← binds Long as VARCHAR for contentId/globalId
    });
    break;
```

Permalink: https://github.com/Apicurio/apicurio-registry/blob/acb5abcc81a367bbcc7f9b9078fe46bb9923baf0/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java#L289

**Suggested fix:** Split `contentId` and `globalId` into their own arm and bind using `filter.getNumberValue().longValue()`, matching the pattern already used in `searchArtifacts()`:

```java
case contentId:
case globalId:
    op = filter.isNot() ? "!=" : "=";
    where.append("v.");
    where.append(filter.getType().name());
    where.append(" ");
    where.append(op);
    where.append(" ?");
    binders.add((query, idx) -> {
        query.bind(idx, filter.getNumberValue().longValue()); // binds as BIGINT
    });
    break;
case state:
case version:
    op = filter.isNot() ? "!=" : "=";
    where.append("v.");
    where.append(filter.getType().name());
    where.append(" ");
    where.append(op);
    where.append(" ?");
    binders.add((query, idx) -> {
        query.bind(idx, filter.getStringValue());
    });
    break;
```

### Environment

SQL persistence backed by PostgreSQL. The type error is PostgreSQL-specific (`operator does not exist: bigint = character varying`); other databases with implicit numeric/string coercion may silently return wrong results rather than failing.

### Steps to Reproduce

1. Run Apicurio Registry with SQL persistence against a PostgreSQL database.
2. Issue a search-versions request filtered by `globalId` or `contentId` — for example via the REST API:
   ```
   GET /search/versions?globalId=42
   ```
   or via the Java client using `SearchFilter.ofGlobalId(42L)` / `SearchFilter.ofContentId(42L)`.
3. Observe a `500 Internal Server Error`.

The inconsistency can also be confirmed by static inspection:
- `SearchFilter.ofGlobalId(Long)` and `SearchFilter.ofContentId(Long)` store a `Long` value.
- `getStringValue()` on a `Long` value returns `value.toString()` — a `String`.
- `searchVersions` line 289 binds that `String` against `v.globalId` / `v.contentId` (`BIGINT` columns).
- `searchArtifacts` lines 144 and 154 call `filter.getNumberValue().longValue()` for the same filters.

**Expected:** Searching versions by `globalId` or `contentId` binds the filter value as a `BIGINT`, matching the column type, and returns matching version records.

**Actual:** The filter value is bound as `VARCHAR`, causing PostgreSQL to throw `operator does not exist: bigint = character varying`, which surfaces to the caller as a `500 Internal Server Error`.

### Logs

```
ERROR: operator does not exist: bigint = character varying
Hint: No operator matches the given name and argument types.
You might need to add explicit type casts.
```

(Exact stack trace will vary by deployment; enable `LOG_LEVEL=DEBUG` to capture the full SQL and JDBC binding trace.)

hcthakur2004:
## Summary

- bind `searchVersions` `globalId` and `contentId` filters with numeric values instead of strings
- keep `state` and `version` filters on string binding
- add regression coverage for version search by `globalId` and `contentId`

Fixes #8698

## Testing

- `git diff --check`
- `./mvnw.cmd -version`

I attempted `./mvnw.cmd -pl app -Dtest=VersionSearchTest test` and `./mvnw.cmd -pl app -DskipTests compile` locally with JDK 21, but both exceeded the local 5 minute command timeout before producing test reports.


paoloantinori (commented):
The fix correctly mirrors the pattern already used in `searchArtifacts()` (lines 144 and 154 of `SqlSearchRepository.java` on main) and resolves the PostgreSQL `operator does not exist: bigint = character varying` error described in #8698. A few observations:

- **Correctness of the split.** The new arm binds via `filter.getNumberValue().longValue()`, matching `SearchFilter.ofContentId(Long)` / `SearchFilter.ofGlobalId(Long)`. The `op = filter.isNot() ? "!=" : "="` line is preserved, so negated filters keep working. This looks right.
- **Column name reuse.** `where.append("v.").append(filter.getType().name())` produces `v.contentId` / `v.globalId`, which align with the actual `t_versions` columns. Worth confirming the casing matches the DDL on all supported backends (Postgres is case-sensitive on quoted identifiers), but since the `state`/`version` arm in the same method already relies on the same `filter.getType().name()` interpolation without issue, this should be safe.
- **Test coverage gap (minor).** `testFilterByGlobalId` and `testFilterByContentId` cover the positive path only. Neither exercises the negated form (`globalId!=...`, `contentId!=...`), which is the branch where the new arm diverges most from the old shared one. A third assertion in each test using the `Not` filter variant would pin that behavior and guard against a future regression collapsing the arms again.
- **Missing `isNot` assertion.** Related to the above: consider asserting that a negated `globalId` filter returns the *other* artifact(s) in the group rather than the matched one, which would also verify the `!=` operator path end-to-end.

Overall the change is small, well-scoped, and directly addresses the root cause. The added tests are a good baseline; expanding them to cover the negated path would make the regression coverage airtight.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -277,6 +277,16 @@ public VersionSearchResultsDto searchVersions(Set<SearchFilter> filters, OrderBy
...
                     case contentId:
                     case globalId:
+                        op = filter.isNot() ? "!=" : "=";
+                        where.append("v.");
+                        where.append(filter.getType().name());
+                        where.append(" ");
+                        where.append(op);
+                        where.append(" ?");
+                        binders.add((query, idx) -> {
+                            query.bind(idx, filter.getNumberValue().longValue());
```
paoloantinori: This is the right fix — binding via `getNumberValue().longValue()` matches the `searchArtifacts()` pattern at lines 144/154. One small note: `op = filter.isNot() ? "!=" : "="` is shared with the `state`/`version` arm below, so the negation path is preserved, but there's currently no test covering a negated `contentId`/`globalId` filter. Worth adding to guard this branch specifically.

app/src/test/java/io/apicurio/registry/noprofile/VersionSearchTest.java
```diff
@@ -90,6 +90,44 @@ void testFilterByLabels() throws Exception {
...
+    void testFilterByContentId() throws Exception {
+        String groupId = TestUtils.generateGroupId();
+
+        CreateArtifactResponse car1 = createArtifact(groupId, "content-id-artifact-1", ArtifactType.JSON,
+                "{}", ContentTypes.APPLICATION_JSON);
+        createArtifact(groupId, "content-id-artifact-2", ArtifactType.JSON, "{\"type\":\"string\"}",
+                ContentTypes.APPLICATION_JSON);
+
+        VersionSearchResults results = clientV3.search().versions().get(config -> {
+            config.queryParameters.groupId = groupId;
```
paoloantinori, line 145: Consider adding a companion assertion using the `Not` variant of the filter (e.g. `globalId != car1's id`) that returns `content-id-artifact-2` instead. That would exercise the `!=` operator path in the new arm, which is the part that diverges most from the old shared string-binding behavior.

hcthakur2004:
Thanks for the review. I added storage-level coverage for negated globalId/contentId version filters, since the REST query parameters currently expose the positive id filters only. This now exercises both numeric binding paths and the != operator path.

paoloantinori:
Thanks @hcthakur2004 — the new cases look right to me. Splitting `contentId`/`globalId` out of the `state`/`version` fallthrough and binding `filter.getNumberValue().longValue()` matches the `searchArtifacts()` pattern, and the negated-path coverage in `AbstractRegistryStorageTest#testSearchVersionsByNegatedGlobalIdAndContentId` is a nice addition. The REST-level tests exercise the positive path end-to-end. From my side this looks good.

carlesarnal (changes requested):
The fix correctly binds globalId and contentId as numbers instead of strings. SQL query construction looks good with proper parameterization.

Missing null check on `filter.getNumberValue()` in app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java line 8. Add validation before calling `.longValue()` or document that SearchFilter guarantees non-null for these filter types.

Test coverage is adequate for happy path but missing edge cases. Add tests for null/invalid number values and combined filters.

The negated filter test in app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java is good but should also exist in VersionSearchTest.java for consistency with the positive case tests.

Minor: Consider adding a comment explaining why globalId/contentId use numeric binding while other filters use string binding. Makes the code more maintainable.

deepak0x:
Hey @hcthakur2004 u raise pr of mine issue without asking me🙂
@maintainers pls take appropriate action

hcthakur2004:
Thanks for the review @carlesarnal. I pushed an update addressing the requested changes: added explicit validation for numeric search filters, documented why globalId/contentId use numeric binding, and expanded coverage for null/invalid numeric filters, combined filters, and negated id filters in VersionSearchTest.

The targeted local Quarkus test run still exceeded my local timeout, but git diff --check passes and CI has started on the updated branch.

paoloantinori (commented):
On the ownership question raised by @deepak0x: this is for the maintainers (@jsenko, @carlesarnal) to sort out. The project's CONTRIBUTING.md now documents the claiming protocol — issues should be claimed with a comment before starting work. If there's an overlap here, the maintainers will decide how to proceed.

On the code itself: @hcthakur2004's latest push addressing Carles's feedback looks reasonable. Waiting on maintainer decision on the ownership question before further review.

hcthakur2004:
Thanks for clarifying. I picked this up because the issue was open and I did not see an assignment or claim before starting. I did not intend to step on anyone’s work. I’m happy to follow whatever the maintainers decide here.

deepak0x:
@EricWittmann hi sir 
he raise pr for mine issue
without asking me 
it's not a good open source practice 
pls take appropriate action
caue it takes time to find issue and someone else raise pr without even asking

EricWittmann:
@deepak0x Do you mean that you opened the GitHub issue with the intention of fixing it, but because we do not yet have any enforced "Issue must be assigned to begin work" policy, @hcthakur2004 opened a PR before you had a chance?

We're still working on our OSS policies and enforcement automation, so apologies if that is what happened.

I think going forward we will probably try to use the issue assignment feature in GitHub to prevent contention like this in the future.

deepak0x:
Yes correct
Also i raise issue and wait for maintainer reply on it too
So that i can have some conversation too of how can i approach it or if it in progress internally
Have contributed in other organisations too and they follow same rule

EricWittmann:
That's fair, and sorry again that this happened.  We don't have any issue assignment policies in place yet, so I think this was just an unfortunate consequence of that.

As I say - we are still working through processes (for example:  you'll notice that all PRs are now being triaged and assigned to specific maintainers).  Things might stay a bit messy until we get things ironed out.

For now, I would say that there are still plenty of ways to contribute if you still want to!

PR #8718, opened 2026-07-18
feat(rest): Enable HTTP request/response compression (#5523)

### LINKED ISSUE #5523: Enable compression in the REST API
The registry REST API does not currently support compression of the request/response payloads.

### Proposed Solution
Enable the Quarkus feature for http compression:

* https://quarkus.io/guides/http-reference#http-compression

parthdude07:
---

**PR Title:** `feat(rest): Enable HTTP request/response compression (#5523)`

**PR Description:**

## What
Enables HTTP compression and decompression on the registry server REST API, adds a custom `WriterInterceptor` for RFC-compliant response compression, and re-enables client-side compression middleware support in the TypeScript and Go SDKs.

## Why
To optimize network transfer efficiency and resolve existing client-side workarounds where Kiota clients had to explicitly disable compression to prevent garbled requests due to the server previously failing to support decompression.

### Backend (`app`)
- Enabled Quarkus native HTTP compression (`quarkus.http.enable-compression=true`) and decompression (`quarkus.http.enable-decompression=true`)
- Added `HttpCompressionWriterInterceptor` — a custom JAX-RS `WriterInterceptor` that gzip-compresses responses for media types in `quarkus.http.compress-media-types` when the client sends `Accept-Encoding: gzip`, since RESTEasy Classic bypasses Vert.x's compression route chain
- RFC 9110 §12.5.3 compliant `Accept-Encoding` parsing with quality-value and wildcard support
- Emits `Vary: Accept-Encoding` on compressed responses for CDN/reverse-proxy cache correctness
- Added `apicurio.rest.compression.enabled` config property (default `true`) as a runtime kill switch — operators can disable the interceptor's compression via environment variable without a rebuild
- Regenerated config docs (`ref-registry-all-configs.adoc`)

### TypeScript SDK
- Replaced the hardcoded custom middleware list with Kiota's default middleware chain using `getPerformanceMiddlewares()`, resolving the manual override

### Go SDK
- Removed the `CompressionHandler` bypass workaround from example implementations and integration tests, reverting to default `NewNetHttpRequestAdapter` initialization

## ⚠️ Breaking Change (TypeScript SDK)
Request-body compression is now **on by default** in the TypeScript SDK. Callers pointed at an older registry instance (pre-decompression support) can opt out via `useDefaultMiddlewares=false` or by constructing a custom middleware chain. See the SDK README for details.

## Verification
- Integration tests (`HttpCompressionTest`) disable RestAssured's auto-decompress to inspect raw wire bytes and headers:
  - Response compression with `Accept-Encoding: gzip` (validates `Content-Encoding: gzip`, `Vary: Accept-Encoding`, byte-level round-trip)
  - Request decompression of gzip-encoded request bodies
  - Non-compression for excluded media types (`application/zip`)
  - Non-compression when client refuses gzip (`gzip;q=0`)
  - Non-compression without `Accept-Encoding` header
- Unit tests for `clientAcceptsGzip()` covering edge cases (wildcards, quality values, malformed tokens)
- `quarkus.http.limits.max-body-size` (50MB) applies to the **decompressed** payload (Vert.x behavior), mitigating decompression bomb risk
- TypeScript SDK builds and passes linting
- Checkstyle passes with 0 violations

Fixes #5523


app/src/main/resources/application.properties
```diff
@@ -25,6 +25,10 @@ quarkus.http.access-log.enabled=true
 quarkus.http.access-log.pattern="apicurio-registry.access method="%{METHOD}" path="%{REQUEST_URL}" response_code="%{RESPONSE_CODE}" response_time="%{RESPONSE_TIME}" remote_ip="%{REMOTE_IP}" remote_user="%{REMOTE_USER}" user_agent="%{i,User-Agent}""
 quarkus.http.proxy.proxy-address-forwarding=false
 
+## HTTP Compression
```
paoloantinori, line 28: These are Quarkus build-time properties, so they can't be overridden per-environment at runtime. For deployments that already compress at a reverse proxy or CDN, this will cause double-compression. Consider scoping with `quarkus.http.compress-media-types` (e.g. to `application/json`) to avoid compressing already-compressed payloads or binary content, and/or documenting this as a deployment-tier behavior change.

typescript-sdk/lib/sdk/factory.ts
```diff
@@ -40,15 +36,14 @@ export class RegistryClientFactory {
...
-        ];
+        let finalMiddlewares: Middleware[] | undefined = middlewares;
+        if (middlewares.length === 0) {
+            finalMiddlewares = undefined;
+        } else {
+            finalMiddlewares = [...MiddlewareFactory.getDefaultMiddlewares(), ...middlewares];
+        }
 
-        const http = KiotaClientFactory.create(undefined, [...defaultMiddlewares, ...middlewares]);
+        const http = KiotaClientFactory.create(undefined, finalMiddlewares);
```
paoloantinori, line 46: Behavior change worth flagging: when `middlewares` is non-empty, the caller's middlewares are now appended to `getDefaultMiddlewares()` rather than replacing the default chain. Existing callers that passed a custom list expecting full replacement will now also get kiota's defaults (compression, retry, redirect, etc.). If that's intended, a one-line note in the SDK changelog would help downstream consumers.

paoloantinori (commented):
Clean, well-scoped change. Enabling Quarkus HTTP compression/decompression in `application.properties` and dropping the kiota client-side workarounds (kiota-http-go#130 was resolved upstream in Jan 2024, so the bypass in `client_test.go` and `main.go` is correctly removed) is the right direction. A few things worth double-checking before merge:

- **No runtime opt-out.** `quarkus.http.enable-compression=true` / `enable-decompression=true` are now hardcoded for every deployment. Environments that terminate compression at a reverse proxy (nginx/ingress) or that run behind a CDN with its own gzip layer will now double-compress unless they override at the proxy. Quarkus exposes these as build-time properties, so a profile override won't work either — consider whether this should be documented as a breaking behavior change for operator-managed deployments, or whether the intent is to handle compression only at the registry tier.
- **BREACH/CRIME surface.** Enabling response compression on endpoints that reflect attacker-controllable input (artifact content, search queries, error messages) expands the information-leakage surface for compression-based oracle attacks. For a schema registry the practical risk is low (responses are mostly large, low-entropy schema payloads rather than secrets), but it's worth a conscious decision rather than a blanket `enable-compression=true`. Quarkus supports `quarkus.http.compress-media-types` and per-path `quarkus.http.filter.*` to scope compression — scoping to JSON content types would be a reasonable hardening if not already the default.
- **Test impact.** No integration tests in this PR exercise the compressed path. If any existing test asserts on raw `Content-Length` or response body encoding, enabling compression globally could subtly break them (Quarkus only compresses when the client sends `Accept-Encoding: gzip`, so most REST-Assured tests should be unaffected — but worth confirming CI is green on this PR beyond the smoke tier).
- **TypeScript `factory.ts` semantics.** When `middlewares.length === 0`, `finalMiddlewares` is set to `undefined`, so `KiotaClientFactory.create(undefined, undefined)` uses kiota's full default chain (including compression). When middlewares are provided, they're *prepended* with `getDefaultMiddlewares()`. This is a behavior change from the previous code, which *replaced* the default chain with the caller's list — callers that previously passed a custom chain expecting it to fully replace defaults will now get the defaults too. Worth a release-note line for SDK consumers.

The Go SDK and `examples/sdk/go-sdk/main.go` changes are straightforward cleanups; the workaround comments referencing the upstream bug are correctly removed.

go-sdk/pkg/tests/client_test.go
```diff
@@ -99,18 +99,7 @@ func TestCreateAnArtifact(t *testing.T) {
...
-		kiotaHttp.NewRedirectHandler(),
-		kiotaHttp.NewParametersNameDecodingHandler(),
-		// NewCompressionHandler(),
-		kiotaHttp.NewUserAgentHandler(),
-		kiotaHttp.NewHeadersInspectionHandler(),
-	)
-
-	adapter, err := kiotaHttp.NewNetHttpRequestAdapterWithParseNodeFactoryAndSerializationWriterFactoryAndHttpClient(&authProvider, nil, nil, httpClient)
-	// adapter, err := kiotaHttp.NewNetHttpRequestAdapter(&authProvider)
+	adapter, err := kiotaHttp.NewNetHttpRequestAdapter(&authProvider)
```
Psykii22, line 163: Great cleanup, just a quick observation should we not check the err before calling .SetBaseUrl()?
I noticed the original code did it in this order too, but if NewNetHttpRequestAdapter ever returns an error, calling .SetBaseUrl() will panic instantly since assert doesn't stop the test execution, moving the check up ensures that we at least get to see the actual error message printed in the logs before the panic happens.

paoloantinori:
Good catch, @Psykii22 — agreeing with the observation on `go-sdk/pkg/tests/client_test.go`. The new code is:

```go
adapter, err := kiotaHttp.NewNetHttpRequestAdapter(&authProvider)
adapter.SetBaseUrl(RegistryUrl)
assert.Nil(t, err)
```

If `NewNetHttpRequestAdapter` ever returns `(nil, err)` in a future kiota release, `adapter.SetBaseUrl(...)` panics with a nil-pointer dereference before `assert.Nil(t, err)` even runs, so the test would fail with a confusing panic rather than the clean assertion we want. The minimal fix is to swap the order:

```go
adapter, err := kiotaHttp.NewNetHttpRequestAdapter(&authProvider)
assert.Nil(t, err)
adapter.SetBaseUrl(RegistryUrl)
```

(or `assert.NoError`). Worth folding into this PR since the file is already being touched. The `main.go` example doesn't have the issue because it already checks `err` via `panic(err.Error())` right after the call.

parthdude07:
Thanks for the great catches, @paoloantinori and @Psykii22 ! I've addressed all the feedback in the latest commit:

- **Compression Scoping:** Scoped HTTP compression to `application/json` via `quarkus.http.compress-media-types` in `application.properties`.
- **Operator Docs:** Added a note to `operator/README.md` warning about the build-time compression behavior and potential double-compression when using reverse proxies/CDNs.
- **TypeScript SDK:** Added a release note to `typescript-sdk/README.md` clarifying the new behavior where custom middlewares append to the Kiota defaults rather than replacing them.
- **Go SDK Tests:** Swapped the assertion order in `client_test.go` to check `err` before calling `.SetBaseUrl()` to prevent nil-pointer panics. 

Let me know if there's anything else you'd like me to tweak

paoloantinori:
Thanks @parthdude07 — all the feedback is addressed cleanly:

- `quarkus.http.compress-media-types=application/json` scopes compression the way we discussed.
- The operator README note about build-time-only behavior and reverse-proxy/CDN double-compression is exactly the kind of caveat operators need.
- The TS SDK release note on the `middlewares` semantics change (append-to-defaults vs. replace) is the right call given this is observable behavior.
- Bonus: the `assert.Nil(t, err)` move before `adapter.SetBaseUrl(...)` in `client_test.go` is the correct ordering — good catch on that too.

LGTM from my side.

carlesarnal (changes requested):
The compression feature is straightforward but has implementation issues.

`app/src/main/resources/application.properties`: Only compressing `application/json` may miss other content types. Document this limitation or expand the list.

`typescript-sdk/lib/sdk/factory.ts`: The middleware merging logic breaks backward compatibility. Previously, passing custom middleware replaced defaults. Now it appends to defaults. This is a breaking change. Either preserve the old behavior with a new parameter, or bump the major version and clearly document the breaking change in the release notes.

`operator/README.md`: The claim that compression "cannot be disabled via runtime environment variables" is wrong. Quarkus properties can be overridden at runtime. Correct this.

`go-sdk/pkg/tests/client_test.go`: Moving `SetBaseUrl` after nil check is good. But the comment in `examples/sdk/go-sdk/main.go` is misleading - the default adapter doesn't "automatically configure compression middleware" on the server, it just uses Kiota's client-side defaults.

Missing: Tests that verify compression actually works. Add integration tests that check response headers and payload sizes.

paoloantinori (commented):
Thanks @parthdude07 — scoping compression to `application/json` and the other fixes look right. Waiting on maintainer `/accept` to trigger CI.

parthdude07:
Hi @carlesarnal 
I reviewed the current changes and can confirm that almost all of the requested feedback has been properly addressed (the `application.properties` configuration, the middleware merging backward compatibility option in the TS SDK, the go-sdk client test fixes, and the new integration tests for HTTP compression). 

However, regarding the feedback to correct `operator/README.md` ("The claim that compression 'cannot be disabled via runtime environment variables' is wrong. Quarkus properties can be overridden at runtime.") — **the original text in the README was actually correct**. 

In Quarkus, `quarkus.http.enable-compression` is a `BUILD_AND_RUN_TIME_FIXED` property. It is baked in at build time and explicitly cannot be overridden at runtime via an environment variable in the pod.

To verify this, I built the application and attempted to override it at runtime. Here is the command sequence and the resulting output:

```bash
# 1. Build the app module
./mvnw clean package -DskipTests -pl app -am

# 2. Run the compiled application, attempting to disable compression via env var
QUARKUS_HTTP_ENABLE_COMPRESSION=false java -jar app/target/quarkus-app/quarkus-run.jar
```

**Output:**
```text
2026-07-23 10:37:55 WARN [io.quarkus.runtime.configuration.ConfigRecorder] (main) Build time property cannot be changed at runtime:
 - quarkus.http.enable-compression is set to 'false' but it is build time fixed to 'true'. Did you change the property quarkus.http.enable-compression after building the application?
```

Because Quarkus ignores the runtime override for this specific property, setting the `env:` in the Operator CR genuinely has no effect. I recommend we keep the original text in `operator/README.md` since it correctly warns users about this exact limitation. 

Let me know if you'd like me to approve the PR as-is with the original README text preserved!

app/src/main/resources/application.properties
```diff
@@ -25,6 +25,16 @@ quarkus.http.access-log.enabled=true
...
+quarkus.http.enable-compression=true
+quarkus.http.enable-decompression=true
+quarkus.resteasy.gzip.enabled=true
+# Setting compress-media-types replaces Quarkus's built-in default list rather than extending
```
paoloantinori, line 36: Worth double-checking the interaction between `quarkus.resteasy.gzip.enabled=true` and the custom `HttpCompressionWriterInterceptor`. For classic RESTEasy this flag enables RESTEasy's built-in `GZIPEncodingInterceptor` (priority 4000, per the comment in `Priorities.java`) alongside the request-side `GZIPDecodingInterceptor`. The built-in encoding interceptor compresses on `Accept-Encoding` only and ignores `compress-media-types`, so for non-listed types like `application/octet-stream` / `application/zip` it could still compress — defeating the scoping the custom interceptor is there to enforce. If that's the case, either the built-in needs to be suppressed for responses or a test should assert binary content is left uncompressed.

app/src/main/java/io/apicurio/registry/rest/HttpCompressionWriterInterceptor.java
```diff
@@ -0,0 +1,70 @@
...
+        try {
+            context.proceed();
+        } finally {
+            gzipStream.finish();
+        }
+    }
+
+    private boolean isCompressible(WriterInterceptorContext context) {
+        String acceptEncoding = httpHeaders.getHeaderString(HttpHeaders.ACCEPT_ENCODING);
+        if (acceptEncoding == null || !acceptEncoding.contains("gzip")) {
```
paoloantinori: `acceptEncoding.contains("gzip")` matches `gzip;q=0`, which per RFC 7231 is an explicit client refusal of gzip — the response would then be compressed when the client asked it not to be. Low impact in the wild, but if you want strictness, splitting on `,` and parsing the `q` value (or at least checking the token isn't suffixed with `;q=0`) would handle it.

paoloantinori (commented):
The new head layers a custom `HttpCompressionWriterInterceptor` on top of the property/SDK changes from earlier revisions, and `HttpCompressionTest` (disabling RestAssured's auto-decompression to inspect raw bytes) is exactly the right way to verify compression on the wire. Two points worth confirming before this lands:

- **`quarkus.resteasy.gzip.enabled` vs. the custom interceptor.** For classic RESTEasy this flag wires up RESTEasy's built-in `GZIPEncodingInterceptor` (response compression) in addition to `GZIPDecodingInterceptor` (request decompression). If the encoding interceptor is active, it compresses based on `Accept-Encoding` alone and does not consult `quarkus.http.compress-media-types`, which would defeat the media-type scoping the custom interceptor enforces — binary artifact downloads (`application/octet-stream`, `application/zip`) the list deliberately excludes could still get compressed. Worth confirming which mechanism owns response compression, and ideally adding a test asserting a non-listed content type is *not* compressed.

- **`Accept-Encoding` parsing.** `acceptEncoding.contains("gzip")` treats `gzip;q=0` (an explicit client refusal of gzip) as acceptance. Low impact in practice, but a tokenized parse respecting `q=0` would be more correct.

app/src/main/resources/application.properties
parthdude07: Great catch. After looking into it, you're absolutely right that `GZIPEncodingInterceptor` creates a fragile state where any future `@GZIP` annotations would bypass our media-type scoping and double-compress. 

To fix this, I have completely removed `quarkus.resteasy.gzip.enabled=true`. I verified that Vert.x's `quarkus.http.enable-decompression=true` is fully handling the request decompression on its own before requests even reach RESTEasy, so we don't actually need the RESTEasy flag at all. By dropping it, we eliminate the `GZIPEncodingInterceptor` entirely while keeping our custom `HttpCompressionWriterInterceptor` strictly in charge of response compression. 
 
I also added a new integration test (`testResponseNotCompressedForNonListedContentType`) using the `/admin/export` endpoint (`application/zip`) to definitively assert that binary content is left uncompressed even when the client advertises gzip support.

app/src/main/java/io/apicurio/registry/rest/HttpCompressionWriterInterceptor.java
parthdude07: I've updated the logic to replace the naive `.contains("gzip")` check with a robust, RFC 9110 compliant tokenized parser (`clientAcceptsGzip`). 

The new parser splits on commas, handles varying whitespace and case insensitivity, and properly extracts and checks the `q` value to explicitly refuse compression for `q=0` (while gracefully degrading to acceptance for malformed `q` values). 

 I've backed this parser with a new suite of 23 unit tests (`HttpCompressionWriterInterceptorTest`) to cover all the edge cases, and added a new integration test (`testResponseNotCompressedWhenClientRefusesGzip`) to verify that an explicit client refusal successfully prevents compression on the wire.

app/src/main/java/io/apicurio/registry/rest/HttpCompressionWriterInterceptor.java
```diff
@@ -0,0 +1,70 @@
...
+            return false;
+        }
+        if (context.getHeaders().getFirst(HttpHeaders.CONTENT_ENCODING) != null) {
+            return false;
```
paoloantinori, line 72: Heads-up: `contains("gzip")` is a substring check, so `Accept-Encoding: gzip;q=0` (a client that explicitly *refuses* gzip) matches and the response gets compressed anyway. Worth parsing the q-value, or at least special-casing `q=0`, so a client that says "not gzip" actually doesn't get gzip.

app/src/main/resources/application.properties
```diff
@@ -25,6 +25,16 @@ quarkus.http.access-log.enabled=true
 quarkus.http.access-log.pattern="apicurio-registry.access method="%{METHOD}" path="%{REQUEST_URL}" response_code="%{RESPONSE_CODE}" response_time="%{RESPONSE_TIME}" remote_ip="%{REMOTE_IP}" remote_user="%{REMOTE_USER}" user_agent="%{i,User-Agent}""
 quarkus.http.proxy.proxy-address-forwarding=false
 
+## HTTP Compression
+quarkus.http.enable-compression=true
+quarkus.http.enable-decompression=true
+quarkus.resteasy.gzip.enabled=true
```
paoloantinori: With this enabled, RESTEasy registers its own `GZIPEncodingInterceptor` alongside the custom `HttpCompressionWriterInterceptor`. The ordering (RESTEasy at 4000 runs outside the custom one at 5000) should make RESTEasy skip in the normal path, but it'd be good to confirm there's no double-gzip on any response path. Also worth a one-line comment clarifying *why* this flag is still needed now that the custom interceptor handles response compression — is it solely for request decompression?

paoloantinori (commented):
Solid iteration on this head — the custom `HttpCompressionWriterInterceptor` is a reasonable workaround for the Vert.x/RESTEasy dispatch split, and `HttpCompressionTest` correctly disables RestAssured's auto-decompression so it actually inspects bytes on the wire. Two things worth checking before this lands:

- **`isCompressible` treats `gzip;q=0` as accepting gzip.** The `acceptEncoding.contains("gzip")` check is a substring match, so a client that sends `Accept-Encoding: gzip;q=0` (i.e. *explicitly refusing* gzip, per RFC 7231 §5.3.4) still gets a gzipped body it may not be able to decode. Same for a token like `x-gzip`-only client. A minimal fix is to split on `,`, strip the `;q=` parts, and treat `q=0` as a refusal; or reuse an existing content-negotiation helper if one is available.
- **Three overlapping compression mechanisms.** This head stacks `quarkus.http.enable-compression`, `quarkus.http.enable-decompression`, `quarkus.resteasy.gzip.enabled=true`, *and* the custom interceptor. With `resteasy.gzip.enabled=true` RESTEasy registers its own `GZIPEncodingInterceptor` (priority `ENTITY_CODER` = 4000, i.e. it runs *outside* this interceptor at USER = 5000), so in the normal path it should see no `Content-Encoding` header yet and skip — but the interaction is subtle and the test only exercises one media type on a happy path. Worth confirming there's no double-gzip on any path, and whether `resteasy.gzip.enabled=true` is needed at all once the custom interceptor owns response compression (it may be there for request decompression — if so, a one-line comment stating that would help future readers).
- Minor: the `compress-media-types` list now includes `application/x-protobuf` and `application/x-thrift`. These are already-dense wire formats where gzip typically buys little and costs CPU per request; the comment above the list explains the octet-stream/zip exclusion, so I assume protobuf/thrift are deliberate — flagging in case they slipped in when expanding beyond `application/json`.

parthdude07:
> Solid iteration on this head — the custom `HttpCompressionWriterInterceptor` is a reasonable workaround for the Vert.x/RESTEasy dispatch split, and `HttpCompressionTest` correctly disables RestAssured's auto-decompression so it actually inspects bytes on the wire. Two things worth checking before this lands:
> 
> * **`isCompressible` treats `gzip;q=0` as accepting gzip.** The `acceptEncoding.contains("gzip")` check is a substring match, so a client that sends `Accept-Encoding: gzip;q=0` (i.e. _explicitly refusing_ gzip, per RFC 7231 §5.3.4) still gets a gzipped body it may not be able to decode. Same for a token like `x-gzip`-only client. A minimal fix is to split on `,`, strip the `;q=` parts, and treat `q=0` as a refusal; or reuse an existing content-negotiation helper if one is available.
> * **Three overlapping compression mechanisms.** This head stacks `quarkus.http.enable-compression`, `quarkus.http.enable-decompression`, `quarkus.resteasy.gzip.enabled=true`, _and_ the custom interceptor. With `resteasy.gzip.enabled=true` RESTEasy registers its own `GZIPEncodingInterceptor` (priority `ENTITY_CODER` = 4000, i.e. it runs _outside_ this interceptor at USER = 5000), so in the normal path it should see no `Content-Encoding` header yet and skip — but the interaction is subtle and the test only exercises one media type on a happy path. Worth confirming there's no double-gzip on any path, and whether `resteasy.gzip.enabled=true` is needed at all once the custom interceptor owns response compression (it may be there for request decompression — if so, a one-line comment stating that would help future readers).
> * Minor: the `compress-media-types` list now includes `application/x-protobuf` and `application/x-thrift`. These are already-dense wire formats where gzip typically buys little and costs CPU per request; the comment above the list explains the octet-stream/zip exclusion, so I assume protobuf/thrift are deliberate — flagging in case they slipped in when expanding beyond `application/json`.

i didnt pushed the latest changes yet

parthdude07:
@EricWittmann how i can run other smoke tests `/ready` is not working now

EricWittmann (commented):
Thanks for the work on this @parthdude07 — a couple of items to address before we can move forward.

**1. Merge conflicts.** The branch is currently in a conflicting state against `main` (likely from recent changes to `go-sdk/pkg/tests/client_test.go`). Please rebase and resolve.

**2. Unrelated `pom.xml` change.** The `maven-antrun-plugin` upgrade (3.1.0 → 3.2.0) is unrelated to HTTP compression. Please remove it from this PR — it should be submitted separately if needed.

**3. `Accept-Encoding` parsing — RFC 9110 §12.5.3 compliance.**

Good news: the `gzip;q=0` concern @paoloantinori raised is properly handled. I traced through `clientAcceptsGzip` against the spec and all common cases work correctly:

| Header | Result | Correct? |
|--------|--------|----------|
| `gzip` | `true` | Yes |
| `gzip, deflate, br` | `true` | Yes |
| `gzip;q=0.8` | `true` | Yes |
| `gzip;q=0` | `false` | Yes |
| `br;q=1.0, gzip;q=0.8` | `true` | Yes |
| `GZIP` (uppercase) | `true` | Yes |
| `gzip;q=0, *;q=0.5` | `false` | Yes (specific coding overrides wildcard per spec) |

However, there's one spec gap: **the wildcard `*` is not handled.** Per RFC 9110 §12.5.3, `*` matches "any available content coding not explicitly listed in the field." So:

| Header | Current Result | Spec-Correct Result |
|--------|---------------|----------------|
| `*` | `false` | `true` — wildcard includes gzip |
| `*;q=0.5` | `false` | `true` — gzip at q=0.5 via wildcard |
| `br, *` | `false` | `true` — wildcard covers gzip |

In practice, most HTTP clients send explicit `Accept-Encoding: gzip, deflate, br` rather than `*`, so this is unlikely to cause real issues. But it is technically a spec violation.

What do you think — is this edge case worth fixing in this PR, or is it acceptable to defer given the low likelihood of clients using the wildcard form? Either way, please add a brief comment in the code noting that `*` is not currently handled so future maintainers are aware.

parthdude07:
> Thanks for the work on this @parthdude07 — a couple of items to address before we can move forward.
> 
> **1. Merge conflicts.** The branch is currently in a conflicting state against `main` (likely from recent changes to `go-sdk/pkg/tests/client_test.go`). Please rebase and resolve.
> 
> **2. Unrelated `pom.xml` change.** The `maven-antrun-plugin` upgrade (3.1.0 → 3.2.0) is unrelated to HTTP compression. Please remove it from this PR — it should be submitted separately if needed.
> 
> **3. `Accept-Encoding` parsing — RFC 9110 §12.5.3 compliance.**
> 
> Good news: the `gzip;q=0` concern @paoloantinori raised is properly handled. I traced through `clientAcceptsGzip` against the spec and all common cases work correctly:
> 
> Header	Result	Correct?
> `gzip`	`true`	Yes
> `gzip, deflate, br`	`true`	Yes
> `gzip;q=0.8`	`true`	Yes
> `gzip;q=0`	`false`	Yes
> `br;q=1.0, gzip;q=0.8`	`true`	Yes
> `GZIP` (uppercase)	`true`	Yes
> `gzip;q=0, *;q=0.5`	`false`	Yes (specific coding overrides wildcard per spec)
> However, there's one spec gap: **the wildcard `*` is not handled.** Per RFC 9110 §12.5.3, `*` matches "any available content coding not explicitly listed in the field." So:
> 
> Header	Current Result	Spec-Correct Result
> `*`	`false`	`true` — wildcard includes gzip
> `*;q=0.5`	`false`	`true` — gzip at q=0.5 via wildcard
> `br, *`	`false`	`true` — wildcard covers gzip
> In practice, most HTTP clients send explicit `Accept-Encoding: gzip, deflate, br` rather than `*`, so this is unlikely to cause real issues. But it is technically a spec violation.
> 
> What do you think — is this edge case worth fixing in this PR, or is it acceptable to defer given the low likelihood of clients using the wildcard form? Either way, please add a brief comment in the code noting that `*` is not currently handled so future maintainers are aware.

@EricWittmann Implementing wildcard ` * ` support in `clientAcceptsGzip`. The change is small and spec-correct — worth doing now rather than leaving a known gap.

parthdude07:
@EricWittmann please allow the  CI/CD tests to run

paoloantinori:
Quick note from my side on the current head (`dc74935`): the unrelated `pom.xml` change Eric flagged earlier is no longer in the diff, so that concern looks resolved. The compression implementation (`HttpCompressionWriterInterceptor` + the `application.properties` scoping) still looks good to me from the earlier rounds.

Deferring to @EricWittmann on the workflow approval / merge-conflict state — those are maintainer-side actions outside my scope here.

EricWittmann (commented):
Thanks for your patience on this @parthdude07 — the implementation is looking good overall, and apologies for not catching this earlier in the review cycle.

One more required change before we can merge:

**Add a runtime kill switch for response compression.**

The `HttpCompressionWriterInterceptor` is registered unconditionally as a `@Provider` with no way to disable it at runtime. The Quarkus properties (`quarkus.http.enable-compression`, `quarkus.http.enable-decompression`) are `BUILD_AND_RUN_TIME_FIXED`, so operators can't override them via environment variables on a running deployment either. We need an escape valve in case compression causes problems in production (e.g. double-compression behind a reverse proxy, unexpected client behavior, performance issues).

Please add a config property like:

```java
@ConfigProperty(name = "apicurio.rest.compression.enabled", defaultValue = "true")
@Info(category = "rest", description = "Enable gzip compression of REST API responses", availableSince = "3.0.0")
boolean compressionEnabled;
```

Then short-circuit at the top of `isCompressible()`:

```java
if (!compressionEnabled) {
    return false;
}
```

This won't cover the Quarkus-level build-time properties, but it gives operators a runtime toggle for the REST API response compression that the custom interceptor handles — which is the main surface area.

After adding the property, remember to regenerate the config docs:

```bash
./mvnw clean install -pl :apicurio-registry-config-generator -am -DskipTests
```

And commit the updated `ref-registry-all-configs.adoc`.

app/src/main/java/io/apicurio/registry/rest/HttpCompressionWriterInterceptor.java
```diff
@@ -0,0 +1,129 @@
...
+    List<String> compressMediaTypes;
+
+    @Override
+    public void aroundWriteTo(WriterInterceptorContext context) throws IOException, WebApplicationException {
+        if (!isCompressible(context)) {
+            context.proceed();
+            return;
+        }
+
+        context.getHeaders().putSingle(HttpHeaders.CONTENT_ENCODING, "gzip");
```
paoloantinori, line 49: Compression is applied conditionally based on the client's `Accept-Encoding`, but no `Vary: Accept-Encoding` is emitted on the response. Behind a shared cache (CDN/reverse proxy — which the operator README explicitly anticipates), a cached gzipped response could be served to a client that didn't send `Accept-Encoding: gzip`, producing a body that client can't decode. RESTEasy's own `GZIPEncodingInterceptor` sets this header. Suggest `context.getHeaders().add(HttpHeaders.VARY, "Accept-Encoding")` here — use `add` rather than `putSingle` so an existing `Vary: Origin` (CORS) is preserved — unless a global response filter already adds it.

app/src/main/resources/application.properties
```diff
@@ -25,6 +25,20 @@ quarkus.http.access-log.enabled=true
 quarkus.http.access-log.pattern="apicurio-registry.access method="%{METHOD}" path="%{REQUEST_URL}" response_code="%{RESPONSE_CODE}" response_time="%{RESPONSE_TIME}" remote_ip="%{REMOTE_IP}" remote_user="%{REMOTE_USER}" user_agent="%{i,User-Agent}""
 quarkus.http.proxy.proxy-address-forwarding=false
 
+## HTTP Compression
+quarkus.http.enable-compression=true
+quarkus.http.enable-decompression=true
+# Do NOT set quarkus.resteasy.gzip.enabled=true here. That flag registers RESTEasy's
```
paoloantinori, line 31: This line is the load-bearing piece of the request-side fix — the registry historically couldn't decompress request bodies, which is exactly why the Kiota `CompressionHandler` bypasses existed. Two things worth confirming before merge: (1) `quarkus.http.enable-decompression` is actually a recognized key in the target Quarkus version — Quarkus hasn't always exposed Vert.x request decompression, and unknown config keys are silently ignored, so `testGzippedRequestBodyIsDecompressedByServer` is the real proof it works and must run green in CI (not be skipped). (2) Verify request-size limits (`quarkus.http.limits.max-body-size`) apply to the decompressed payload; if the limit is checked against compressed bytes, a small gzip request can decompress into a very large one (decompression bomb).

paoloantinori (commented):
Solid, well-instrumented change: the `Accept-Encoding` parser is RFC-aware and heavily unit-tested, and the integration tests disable RestAssured's auto-decompress to inspect raw bytes — the right way to verify wire behavior rather than just configuration. Two gaps stand out: a missing `Vary: Accept-Encoding` (caching correctness behind a CDN/proxy) and confirming that request decompression is genuinely enabled and size-bounded.

- `app/src/main/java/io/apicurio/registry/rest/HttpCompressionWriterInterceptor.java:44-45` — gzip is applied conditionally on the client's `Accept-Encoding`, but the response never advertises `Vary: Accept-Encoding`. A shared cache/CDN could then serve a gzipped body to a client that didn't request gzip (corruption on decode); RESTEasy's own `GZIPEncodingInterceptor` sets this header. Recommend `context.getHeaders().add(HttpHeaders.VARY, "Accept-Encoding")` (use `add`, not `putSingle`, so an existing `Vary: Origin` from CORS survives) unless a global response filter already sets it.
- `app/src/main/resources/application.properties:31` — `quarkus.http.enable-decompression=true` is the crux of the request-side fix (the reason the Kiota compression bypasses are being removed). Worth confirming before merge: (1) this key is actually recognized in the target Quarkus version — Quarkus hasn't always exposed Vert.x request decompression, and unknown keys are silently ignored, so the integration test `testGzippedRequestBodyIsDecompressedByServer` is the real proof and must be green in CI; (2) that request-size limits apply to the decompressed payload, else a small gzip request can expand into a very large one (decompression bomb).
- `typescript-sdk/lib/sdk/factory.ts` — request-body compression is now on by default, which breaks callers pointed at an older (pre-decompression) registry; the `useDefaultMiddlewares=false` opt-out plus the README note handle this, but it warrants a prominent release-note flag. (Minor: the PR description references `getDefaultMiddlewares()` while the code and README use `getPerformanceMiddlewares()` — worth reconciling.)

parthdude07:
> Solid, well-instrumented change: the `Accept-Encoding` parser is RFC-aware and heavily unit-tested, and the integration tests disable RestAssured's auto-decompress to inspect raw bytes — the right way to verify wire behavior rather than just configuration. Two gaps stand out: a missing `Vary: Accept-Encoding` (caching correctness behind a CDN/proxy) and confirming that request decompression is genuinely enabled and size-bounded.
> 
> * `app/src/main/java/io/apicurio/registry/rest/HttpCompressionWriterInterceptor.java:44-45` — gzip is applied conditionally on the client's `Accept-Encoding`, but the response never advertises `Vary: Accept-Encoding`. A shared cache/CDN could then serve a gzipped body to a client that didn't request gzip (corruption on decode); RESTEasy's own `GZIPEncodingInterceptor` sets this header. Recommend `context.getHeaders().add(HttpHeaders.VARY, "Accept-Encoding")` (use `add`, not `putSingle`, so an existing `Vary: Origin` from CORS survives) unless a global response filter already sets it.
> * `app/src/main/resources/application.properties:31` — `quarkus.http.enable-decompression=true` is the crux of the request-side fix (the reason the Kiota compression bypasses are being removed). Worth confirming before merge: (1) this key is actually recognized in the target Quarkus version — Quarkus hasn't always exposed Vert.x request decompression, and unknown keys are silently ignored, so the integration test `testGzippedRequestBodyIsDecompressedByServer` is the real proof and must be green in CI; (2) that request-size limits apply to the decompressed payload, else a small gzip request can expand into a very large one (decompression bomb).
> * `typescript-sdk/lib/sdk/factory.ts` — request-body compression is now on by default, which breaks callers pointed at an older (pre-decompression) registry; the `useDefaultMiddlewares=false` opt-out plus the README note handle this, but it warrants a prominent release-note flag. (Minor: the PR description references `getDefaultMiddlewares()` while the code and README use `getPerformanceMiddlewares()` — worth reconciling.)



---

**On `Vary: Accept-Encoding` (HttpCompressionWriterInterceptor.java:44-45):**

> Good catch — fixed. Added `context.getHeaders().add(HttpHeaders.VARY, "Accept-Encoding")` using `add()` (not `putSingle()`) so existing `Vary` values like `Origin` from CORS are preserved. The integration test `testResponseIsCompressedWhenClientAcceptsGzip` now also asserts `.header("Vary", containsString("Accept-Encoding"))`.

---

**On `quarkus.http.enable-decompression=true` (application.properties:31):**

> Both concerns checked:
>
> 1. **Recognized in target Quarkus version** — we're on Quarkus 3.33.2.1; `quarkus.http.enable-decompression` has been supported since Quarkus 2.x (it maps directly to Vert.x's `setDecompressionSupported`). The integration test `testGzippedRequestBodyIsDecompressedByServer` is the real proof — it sends a gzip-encoded request body and verifies the server correctly round-trips the content. This test must stay green in CI and is not skipped.
>
> 2. **Decompression bomb** — `quarkus.http.limits.max-body-size` (set to 50MB at line 22) is enforced by Vert.x on the **decompressed** payload, not the compressed bytes. A small gzip payload that expands beyond 50MB will be rejected before the application layer sees it, so we're covered.

---

**On TypeScript SDK naming inconsistency (`getDefaultMiddlewares()` vs `getPerformanceMiddlewares()`):**

> Good eye — will reconcile the PR description to match the actual code which uses `getPerformanceMiddlewares()`. The release-note flag for the breaking change in request-body compression defaults is noted and will be included.

---

parthdude07:
@EricWittmann please have a review .

carlesarnal:
Review of latest changes - a few minor items remain.

**1. Missing DCO sign-off on commit 015a8d7d**
This commit has no Signed-off-by line. All other contributor commits have it. Can be fixed with git rebase -i --signoff, or resolved if the project squash-merges.

**2. availableSince version is wrong for the new config property**
In RestConfig.java:53 and ref-registry-all-configs.adoc, the new apicurio.rest.compression.enabled property uses availableSince 3.0.0 but the project is at 3.3.2-SNAPSHOT. Should be 3.3.2 or whatever version this ships in. Regenerate docs after fixing.

**3. (Nit) Trim values in getCompressMediaTypes()**
In HttpCompressionWriterInterceptor.java:88-89, the split values are not trimmed. If anyone configures the property with spaces after commas, the media type comparison would silently fail. Adding .map(String::trim) after split would make this robust.

All substantive feedback from prior reviews (Vary header, RFC wildcard handling, runtime kill switch, integration tests, TS SDK backward compat) has been properly addressed. The RFC 9110 parser is well-tested, integration tests disable RestAssured auto-decompression to inspect raw wire behavior, and the double-compression guard is correct. Nice work @parthdude07.

parthdude07:
> Review of latest changes - a few minor items remain.
> 
> **1. Missing DCO sign-off on commit [015a8d7](https://github.com/Apicurio/apicurio-registry/commit/015a8d7d720286e941eaa3b7ffadcaabf3c243a7)** This commit has no Signed-off-by line. All other contributor commits have it. Can be fixed with git rebase -i --signoff, or resolved if the project squash-merges.
> 
> **2. availableSince version is wrong for the new config property** In RestConfig.java:53 and ref-registry-all-configs.adoc, the new apicurio.rest.compression.enabled property uses availableSince 3.0.0 but the project is at 3.3.2-SNAPSHOT. Should be 3.3.2 or whatever version this ships in. Regenerate docs after fixing.
> 
> **3. (Nit) Trim values in getCompressMediaTypes()** In HttpCompressionWriterInterceptor.java:88-89, the split values are not trimmed. If anyone configures the property with spaces after commas, the media type comparison would silently fail. Adding .map(String::trim) after split would make this robust.
> 
> All substantive feedback from prior reviews (Vary header, RFC wildcard handling, runtime kill switch, integration tests, TS SDK backward compat) has been properly addressed. The RFC 9110 parser is well-tested, integration tests disable RestAssured auto-decompression to inspect raw wire behavior, and the double-compression guard is correct. Nice work @parthdude07.

addressed all the suggestion. Thnks a lot for review

carlesarnal:
## Re-review

Checked the latest head (`7f32084` + merge `80782a1`) against the three items from the previous round:

| # | Issue | Status |
|---|-------|--------|
| 1 | Missing DCO sign-off on commit `015a8d7` | ✅ Fixed — commit was rebased out; all non-merge commits now carry `Signed-off-by` |
| 2 | Wrong `availableSince` version (`3.0.0` → `3.3.2`) in `RestConfig.java` and docs | ✅ Fixed — both `RestConfig.java:53` and `ref-registry-all-configs.adoc` now show `3.3.2` |
| 3 | Missing `.trim()` in `getCompressMediaTypes()` | ✅ Fixed — `.map(String::trim)` applied at `HttpCompressionWriterInterceptor.java:91` |

All three items are resolved. I don't see new issues in the diff.

**One minor suggestion (non-blocking):** The `operator/README.md` HTTP Compression section tells users the Quarkus compression properties are `BUILD_AND_RUN_TIME_FIXED` and baked into the image, but doesn't mention the runtime kill switch `apicurio.rest.compression.enabled` that was added specifically for this use case. Operator users might not realize they can set `APICURIO_REST_COMPRESSION_ENABLED=false` in their CR's `env:` block to disable compression at runtime. Could be worth a one-liner addition, but fine to do in a follow-up.

Everything else looks good — the interceptor logic, RFC 9110 parsing, integration and unit tests, SDK changes, and docs are solid.

parthdude07:
@carlesarnal Done with the all requested changes .thanks for pointing me out

app/src/main/resources/application.properties
```diff
@@ -25,6 +25,20 @@ quarkus.http.access-log.enabled=true
 quarkus.http.access-log.pattern="apicurio-registry.access method="%{METHOD}" path="%{REQUEST_URL}" response_code="%{RESPONSE_CODE}" response_time="%{RESPONSE_TIME}" remote_ip="%{REMOTE_IP}" remote_user="%{REMOTE_USER}" user_agent="%{i,User-Agent}""
 quarkus.http.proxy.proxy-address-forwarding=false
 
+## HTTP Compression
```
paoloantinori, line 28: Enabling request decompression opens a decompression-bomb attack surface. The PR's mitigation relies on `quarkus.http.limits.max-body-size` (50MB) bounding the *decompressed* payload, but that Vert.x behavior is version-dependent — the HTTP body limit has historically applied to wire/compressed bytes in several versions — and there is no test confirming it holds for this stack. If the limit applies to compressed bytes instead, a small request can decompress far past 50MB. Recommend adding a test that POSTs a high-ratio zip bomb (a few KB compressing to >50MB) and asserts the server rejects it (413), which verifies the mitigation and guards against regression.

app/src/main/java/io/apicurio/registry/rest/HttpCompressionWriterInterceptor.java
```diff
@@ -0,0 +1,147 @@
...
+            return false;
+        }
+        if (context.getHeaders().getFirst(HttpHeaders.CONTENT_ENCODING) != null) {
+            return false;
+        }
+        MediaType mediaType = context.getMediaType();
+        if (mediaType == null) {
+            return false;
+        }
+        return getCompressMediaTypes().contains(mediaType.getType() + "/" + mediaType.getSubtype());
```
paoloantinori: Media types are case-insensitive, but this comparison uses the raw `type/subtype` against an all-lowercase list, so a response carrying a non-lowercase media type (e.g. `Application/JSON`) would silently skip compression (and would not get a `Vary: Accept-Encoding` header). Lowercasing both sides — e.g. `.toLowerCase(Locale.ROOT)` on the built string — is a trivial, RFC-correct hardening. Low real-world impact today since responses are lowercase, but easy to make robust.

paoloantinori (commented):
Overall this is a well-built change: the custom WriterInterceptor is clearly reasoned, the Accept-Encoding parser is RFC-conscientious, and the integration tests correctly disable RestAssured's auto-decompress to inspect raw wire bytes. The main items worth addressing before merge are confirming the decompression-bomb mitigation and a couple of smaller robustness points.

- Request decompression (`application.properties:28`) is a real attack surface, and the PR's mitigation rests entirely on `quarkus.http.limits.max-body-size` applying to the *decompressed* payload — a Vert.x behavior that is version-dependent and currently unverified, with no test covering it. Recommend a test that POSTs a high-ratio zip bomb (a few KB → well over 50MB) and asserts a 413, which both verifies the property and locks it against regression.
- Media-type matching at `HttpCompressionWriterInterceptor.java:78` is case-sensitive: it compares the raw `type/subtype` against an all-lowercase list, so a non-lowercase media type would silently skip compression. Lowercasing both sides is a trivial, RFC-correct hardening.
- Worth confirming the active RESTEasy flavor: the interceptor's whole design assumes Classic RESTEasy bypasses Vert.x compression (`HttpCompressionWriterInterceptor.java:24`). If the build is on `quarkus-resteasy-reactive`, Vert.x already honors `enable-compression` for these media types and this interceptor could double-compress. The single-gunzip assertion in `HttpCompressionTest` would catch that if run on the shipped stack, so verify the test profile matches production rather than a classic-only setup.
- The runtime kill switch (`apicurio.rest.compression.enabled`) only disables the custom interceptor's *response* compression; request decompression (`enable-decompression`, build-time-fixed) and any Vert.x response compression stay on. The operator README's claim that it works 'regardless of the baked-in Quarkus settings' slightly overstates this. This config also has no test coverage — a test toggling it to false and asserting uncompressed responses would be a cheap addition.
- (Minor) Consider interaction with `Range` requests: the interceptor does not skip 206 responses, so if any artifact endpoint advertises Range, a client sending both `Range` and `Accept-Encoding: gzip` would receive a compressed full body rather than the requested slice.

parthdude07:
@paoloantinori 
Good call. Added `testDecompressionBombRejectedByBodySizeLimit` — it constructs a valid JSON payload (~60MB uncompressed, a few KB on the wire) and POSTs it with `Content-Encoding: gzip`, asserting the server rejects it.

 One finding: Vert.x does decompress the payload before the application sees it, so the mitigation works — but the rejection comes as a 400 (the app layer rejects the oversized content) rather than a clean 413 from the HTTP body-size limit. The test asserts that the request does **not** succeed (200/201), accepting either 413 or 400 as valid rejection. This locks the behavior against regression either way.

Also added a kill-switch test (`HttpCompressionDisabledTest`) for `apicurio.rest.compression.enabled=false`, since you flagged the gap there.

app/src/main/java/io/apicurio/registry/rest/HttpCompressionWriterInterceptor.java
parthdude07: Fixed — both sides are now lowercased with Locale.ROOT: the composed type/subtype at the comparison site and the config values when building the list in getCompressMediaTypes(). Agreed it's low real-world risk today but trivial to make correct.

paoloantinori:
Thanks @parthdude07 — this is a genuinely useful finding, and the test is the right kind to add.

Two things worth pulling out of it:

**1. The mitigation holds.** The decompression-bomb payload (~60 MB uncompressed, a few KB on the wire) is rejected, so the body-size budget is enforced on the *decompressed* size, not the wire size. That's the property that actually matters for resource-exhaustion protection — a gzip stream that decodes to 60 MB cannot sneak past the limit as a 4 KB request. Good.

**2. The 400-vs-413 mismatch is real but low-severity.** Vert.x's decompression runs before the HTTP body-size limiter sees the bytes, so the rejection surfaces as a 400 from the application layer rather than a clean 413 from the transport. Functionally the client gets the same outcome (request refused, no resource leak), but the status code lies a little about *why*. Two options, neither blocking:

- **Document it** — a one-line note in the compression README/section that oversized compressed bodies are rejected with 400, not 413, because decompression precedes the body-size check. Cheapest, and it stops a future reader from filing a "wrong status code" bug.
- **Align it** — if Vert.x exposes the decompressed size early enough to short-circuit with 413, that's the semantically honest path. But only worth it if it's a one-liner; I wouldn't reshape the interceptor around it.

My read: ship with the test as-is (it correctly pins current behaviour) and add the doc note. The 413 alignment can be a follow-up if anyone hits a client that strictly keys on 413.

Nothing else from me on this head — the WriterInterceptor and the SDK middleware changes still look solid from the earlier rounds.

parthdude07:
@paoloantinori Pushed  the recent changes . please review them

paoloantinori:
Thanks @parthdude07 — the latest head looks consistent with the feedback from the previous rounds. The `application/json` scoping, the kill-switch test, the DCO sign-off, and the decompression-bomb test all read cleanly, and the merge-conflict / `pom.xml` items Eric flagged earlier are resolved.

Nothing blocking from my side on this head. The interceptor design and the RFC-aware `Accept-Encoding` parser are solid, and the test suite disables RestAssured's auto-decompress the right way to inspect raw bytes. This is ready for maintainer merge once CI is green.

parthdude07:
@paoloantinori @carlesarnal @EricWittmann Thanks for your multiple reviews to get me through this PR.Looking forward to contribute to this amazing community.
I am also applying for LFX mentorship term 3 projects. Looking forward for more contribution apportunities.

parthdude07:
It would be Great if the maintainers allow the CI to run

paoloantinori:
Thanks @parthdude07 — glad the review rounds were useful. The compression work looks good and it's a solid contribution. Good luck with the LFX term-3 applications.

parthdude07:
> Thanks @parthdude07 — glad the review rounds were useful. The compression work looks good and it's a solid contribution. Good luck with the LFX term-3 applications.
> 
> 

Thanks a lot 😊

EricWittmann:
Thanks @parthdude07 — solid work on this. The RFC 9110-compliant Accept-Encoding parsing, the runtime kill switch, the decompression bomb test, and the SDK cleanup all came together well. Appreciate your persistence through the review iterations. Merged!

parthdude07:
> Thanks @parthdude07 — solid work on this. The RFC 9110-compliant Accept-Encoding parsing, the runtime kill switch, the decompression bomb test, and the SDK cleanup all came together well. Appreciate your persistence through the review iterations. Merged!

Thanks alot.

PR #8724, opened 2026-07-18
feat(cli): detect terminal width when rendering tables

### LINKED ISSUE #8647: CLI: Detect terminal width when rendering tables
## Parent Epic

## Description
Tables currently use a fixed width regardless of the terminal size. Use [Jansi](https://fusesource.github.io/jansi) or a similar library to detect the actual terminal width and wrap or truncate columns accordingly.

## Scope
- Add Jansi (or alternative) dependency to the CLI module
- Detect terminal width at runtime
- Adapt `TableBuilder` to truncate or wrap columns based on available width
- Graceful fallback to a default width if detection fails (e.g. piped output)

## Acceptance Criteria
- Tables adapt to the terminal width
- Wide terminals show more data, narrow terminals truncate gracefully
- Piped output (non-TTY) uses a sensible default width
- GraalVM native image compatibility verified

AbdulWasih05:
## What

Tables rendered by the CLI now adapt to the terminal width instead of hard-wrapping every column at 25 characters.

- New `TerminalUtils` detects the terminal width once at startup and caches it:
  1. Non-TTY output (piped/redirected) falls back to a default of 120, using `System.console() == null` as the trigger. On JDK 22+ `System.console()` is non-null even for redirected streams, so `Console::isTerminal` is additionally consulted (reflectively, since the project compiles at an older source level). This matters for the native binary: the Mandrel builder image is JDK 23.
  2. Interactive terminals use Jansi's `AnsiConsole.getTerminalWidth()`.
  3. If Jansi reports 0 or fails to load, the `COLUMNS` environment variable is consulted (helps in Docker/CI where terminal emulation is inconsistent).
  4. Otherwise the default of 120 applies.
- `TableBuilder` gains a fluent `setMaxWidth(int)` (defaults to the detected width; `<= 0` disables the limit). Column widths are allocated fair-share: columns narrower than their share keep their natural width, the freed space is redistributed among wider columns, and no column shrinks below the existing 3-char minimum. Cell lines that exceed their column's allocation are truncated with `...`.

## Why

Fixed-width tables waste wide terminals and overflow narrow ones. See #8647 and the implementation notes from @carlesarnal there, this follows that approach.

## Notes for reviewers

- **Behavior change:** long cell values were previously chunk-wrapped at 25 chars into multiple lines; they are now truncated with an ellipsis at the column's allocated width (explicit `\n` line breaks are still honored). Happy to switch to wrapping-at-allocation instead if you prefer to keep all content visible.
- The trailing column separator after the last column is kept as-is to minimize the diff, so rendered lines carry 3 trailing spaces beyond the budget. Can trim if wanted.
- Jansi comes from the Quarkus BOM (2.4.0), no explicit version.
- Native image: added `quarkus.native.resources.includes` for Jansi's bundled native libraries, and a reflect-config entry for `Console::isTerminal`. Needs validation in the native CI build, I'd appreciate a maintainer triggering that.

## Testing

- `TableBuilderTest`: exact-output test for the fits case; truncation; fair-share allocation (narrow column keeps natural width); multi-line cells; `maxWidth 0` = unlimited; minimum width clamp; pagination footer; empty table.
- `TerminalUtilsTest`: width is positive and cached; `COLUMNS` parsing edge cases.
- Manual (JVM mode, `acr context` with a deliberately long registry URL):
  - Piped output on the host: non-TTY detected, table rendered at the 120 default with ellipsis truncation.
  - `docker run` (no `-t`), `eclipse-temurin:21-jre`: non-TTY, 120 default.
  - `docker run -t` from a non-terminal parent: pty allocated with no usable size, Jansi reports 0, no `COLUMNS` set, graceful 120 default (the "inconsistent terminal emulation" case from the issue discussion).
  - `docker run -t -e COLUMNS=60`: Jansi reports 0, `COLUMNS` fallback honored- table rendered at exactly 60 with truncation.
  - Interactive terminal (Windows, PowerShell in VS Code): Jansi detected the real window width. At full width the table expanded past the 120 default and only mildly truncated the longest URL; after narrowing the window the same command rendered a compact table with the wide column shrunk to its fair share and narrow columns untouched.

Fixes #8647


cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -133,17 +212,7 @@ private static class Cell {
...
-                    // Split the line into chunks of MAX_COLUMN_WIDTH
-                    for (int i = 0; i < line.length(); i += MAX_COLUMN_WIDTH) {
-                        int end = min(i + MAX_COLUMN_WIDTH, line.length());
-                        lines.add(line.substring(i, end));
-                    }
-                } else {
-                    lines.add(line);
-                }
-            });
+            lines.addAll(List.of(value.split("\n")));
```
paoloantinori: This removes the old line-splitting that preserved full cell content. Combined with `fit()` truncating to the allocated width, long values are now silently clipped on narrow terminals instead of wrapping. For a registry CLI where users inspect artifact/description text, consider wrapping within the shrunk column width (multi-line cells already work via `Cell.getLines()`) so no data is lost, or document this as an intentional trade-off.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -100,11 +115,75 @@ public void print(StringBuilder out) {
...
+            }
+        }
+        return widths;
+    }
+
+    /**
+     * Pads the value to the given width, truncating with an ellipsis if it does not fit.
+     */
+    private String fit(String value, int width) {
+        if (value.length() <= width) {
```
paoloantinori, line 325: `value.length()` counts UTF-16 code units, not display columns. CJK characters (width 2) and surrogate pairs (emoji) will misalign columns. Fine for an ASCII-heavy first version, but worth a comment noting the limitation so it is not a surprise when someone files a bug about misaligned tables with international content.

cli/src/main/java/io/apicurio/registry/cli/utils/TerminalUtils.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        return width;
+    }
+
+    private static int detectWidth() {
+        if (!isTerminal()) {
+            return DEFAULT_WIDTH;
+        }
+        try {
+            var width = AnsiConsole.getTerminalWidth();
```
paoloantinori, line 47: Worth confirming in the native binary that `AnsiConsole.getTerminalWidth()` returns a positive value without a prior `AnsiConsole.systemInstall()` call — some Jansi versions return `-1` from the static method until the console is installed, which would silently skip this branch on every native invocation. A quick manual check of `apicurio-registry-cli` in a real terminal (not just CI) would close the gap the unit tests cannot cover.

paoloantinori (commented):
Nice work overall — the iterative fair-share allocator in `computeColumnWidths` is clean and the test coverage for the layout logic is solid. A few observations worth considering before this lands:

- **Truncation vs. wrapping is a behavior regression on narrow terminals.** The old `Cell` constructor split lines at `MAX_COLUMN_WIDTH` (25) so *all* cell content was preserved across multiple visual lines. The new `fit()` truncates with an ellipsis, which means long artifact names, descriptions, or group IDs are silently clipped when the terminal cannot accommodate them. For a CLI often used to inspect registry content, losing data in the display is more surprising than a taller table. Worth considering: wrap within the allocated column width (preserving the old multi-line behavior at a smaller width) rather than truncate, or expose a `--no-truncate` / `setMaxWidth(0)` escape hatch for scripted use.
- **`String.length()` is not display width for non-ASCII content.** `fit()` and `computeColumnWidths()` both rely on `String.length()`, which counts UTF-16 code units. CJK characters occupy 2 terminal columns and emoji/surrogate pairs throw the count off further, so tables containing such content will not align. Probably acceptable for a first cut given that registry identifiers skew ASCII, but worth noting as a known limitation.
- **Native-image path for `AnsiConsole.getTerminalWidth()` is not exercised.** The unit tests cover `parseColumns` and the cache, but not the TTY-detection to Jansi to `COLUMNS` fallback chain. In particular, some Jansi versions return `-1` from the static `getTerminalWidth()` unless `AnsiConsole.systemInstall()` has been called first — worth verifying in the actual native binary (Mandrel/JDK 23) that this returns a positive value and does not silently fall through to `COLUMNS`/default in every real terminal.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
AbdulWasih05: Switched to wrapping cell content at the allocated width (ellipsis only for headers now), and also fixed the trailing separator overflowing the terminal width by 3 spaces, which showed up as a blank line after every row.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
AbdulWasih05: Added the note to the class javadoc (UTF-16 code units vs display columns, CJK and emoji may misalign), with an actual wcwidth style measure as a possible follow-up.

cli/src/main/java/io/apicurio/registry/cli/utils/TerminalUtils.java
AbdulWasih05: Checked jansi 2.4.2 (what the BOM resolves): static getTerminalWidth() lazily calls initStreams() via out(), so no systemInstall() is needed, and I verified in a real Windows terminal in JVM mode that tables re-wrap on resize; the native binary I could not build locally, but the COLUMNS/default fallback keeps the worst case at 120 columns.

carlesarnal (dismissed):
The terminal width detection works but has a logic flaw in the JDK version check. `TerminalUtils.isTerminal()` returns true on pre-JDK 22 when it should return false for redirected streams. Fix the reflection fallback.

The wrapping algorithm is solid. Tests are thorough. The null handling in `Cell` constructor needs attention - either add a null check or document that nulls are unsupported.

Minor: `fit()` truncation behavior is inconsistent when width is very small. Either always show ellipsis or document why it's omitted.

The jansi dependency is fine but increases native binary size. Make sure this is acceptable.

Files to fix:
- `cli/src/main/java/io/apicurio/registry/cli/utils/TerminalUtils.java` - isTerminal() logic
- `cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java` - Cell constructor null handling

AbdulWasih05:
Thanks for the review! All addressed:

- **isTerminal()**: on pre-JDK 22, System.console() itself returns null when streams are redirected, so a non-null console already implies a real terminal there. That said, the catch was too broad, so I narrowed it: NoSuchMethodException (pre-JDK 22) still returns true, any other reflective failure on JDK 22+ now falls back to the default width.
- **Cell null handling**: nulls are now normalized to empty cells in the Cell constructor, with a test.
- **fit() ellipsis**: documented why it is omitted at widths up to the ellipsis length; at the minimum column width of 3 it would consume the whole column, while a plain cut keeps identifying characters.
- **Jansi size**: it is a new direct dependency, ~250 KB as a jar. In the native image the classes are trimmed to the reachable subset, but the resources include currently embeds the bundled native libraries for all platforms (~600 KB uncompressed); since the native binary is built per OS anyway, I can narrow that include to the build platform's library if the size is a concern.

paoloantinori:
Thanks @AbdulWasih05 — the narrowed catch in `isTerminal()` reads correctly now: `NoSuchMethodException` (pre-JDK 22) keeps the "non-null console ⇒ real terminal" semantics, while the broader `ReflectiveOperationException` fails toward the safe default width. Combined with the `reflect-config.json` entry for `Console::isTerminal`, the native path is covered too. Good to go from my side.

paoloantinori (commented):
Thanks @AbdulWasih05 — all three points addressed well. The wrapping behavior (ellipsis only for headers), the UTF-16 vs display-column caveat in the javadoc, and the jansi `getTerminalWidth()` verification on Windows all look good. Waiting on maintainer `/accept`.

AbdulWasih05:
The test suite failure is fixed. `SearchCommandTest.testSearchVersionsTableOutput` failed because tables applied the default width of 120 to piped output, truncating the `Artifact ID` header. Output that is not a terminal now renders at natural width, matching the behavior before this PR, and the default width only applies inside a terminal whose width cannot be detected.

The Verify run for the latest commit is waiting for workflow approval, @jsenko could you approve it so the full suite can confirm?

paoloantinori:
Confirmed — returning `0` (no width limit) from `detectWidth()` when `isTerminal()` is false is the right call. That keeps piped/redirected output at natural width, which is what `SearchCommandTest.testSearchVersionsTableOutput` expects, and `TerminalUtilsTest.testGetTerminalWidthIsUnlimitedWhenNotATerminal` pins the behavior. Good to see the regression test is back to green.

AbdulWasih05:
hey @carlesarnal @EricWittmann , the Verify run for the latest commit is still waiting for workflow approval, could you approve it so the full suite can run? 

I tried /retry but it replied with nothing to re-trigger, since it only covers runs that concluded failed or cancelled. There does not seem to be a lifecycle command that can start a run that is waiting for approval, so a manual approve looks like the only way forward here.

AbdulWasih05:
@EricWittmann the last Verify failure was not from this PR. That run used a merge snapshot frozen on July 21, which had the config generator change from #8524 but not the regenerated docs from #8797, so the docs freshness check failed before any tests ran. The branch is now updated with the latest main.

Could you approve the Verify run for the new commit? `/retry` cannot start it since it only re-triggers failed or cancelled runs.

AbdulWasih05:
hey @carlesarnal, @EricWittmann, tests are green and main is merged in.

PTAL and merge whenever you're ready!

jsenko (commented):
## Code Review — 0 findings (observations only)

Solid implementation. The width allocation algorithm is correct, the fallback chain (Jansi → COLUMNS env → default 120) is robust, and the native image configuration is well-scoped. paoloantinori's feedback was addressed (wrapping instead of truncation, UTF-16 documented, Jansi systemInstall not needed).

Two minor test gap observations (not blocking):
- No test for single-column tables (boundary case for the allocation algorithm, though code handles it correctly)
- No test where all columns equally exceed fair share and must shrink uniformly

The CLI module doesn't use license headers on any file, so the missing header on TerminalUtils.java is consistent with the module convention.

Overall well-done. The PR description is thorough with good manual testing coverage across TTY/non-TTY/Docker scenarios.

AbdulWasih05:
Hey @jsenko, added both the tests, PTAL!
 Could you approve the waiting Verify run, and merge if everything looks good?

AbdulWasih05:
Hey @carlesarnal @EricWittmann, all tests are green and every review point is addressed, but the PR still shows lifecycle/waiting-on-author because the Jul 20 changes requested review is still the latest one, so it probably does not surface in review queues. PTAL!

aclfe:
noting for visibility: there's a separate open PR (#9256) touching the same TODO in TableBuilder.java, opened independently without reference to issue #8647. that one only resolves single-cell wrapping via COLUMNS/stty detection, it doesn't do fair-share column allocation across the table and doesn't reuse Jansi.

this PR looks like the fuller fix for #8647 and already has multiple rounds of review addressed with CI green, might make sense to close #9256 in favor of this one once it merges.

AbdulWasih05:
Hey @jsenko @EricWittmann, this one has been green since Jul 30 with both of the tests you asked for, but it is still pinned by the Jul 20 changes requested review, which keeps the reconciler re-adding lifecycle/waiting-on-author.

The same TODO now also has PR #9256 and issue #9479 open against it, and the fair share formula proposed in #9479 is what this PR already implements, so it would be good to settle which one to take.

Could you dismiss the stale review or merge if it looks good to you?

vandanayadav7 (dismissed):
Thorough review — code quality, security, edge cases, algorithm correctness, and naming conventions all checked. E2e tested in a container: piped output renders at unlimited width (correct for scripts), JSON unaffected, tables display correctly. 30 unit tests pass covering every branch of the fair-share algorithm.

The Jansi detection + fair-share column allocation is the right solution for #8647. Clean separation between TerminalUtils (detection) and TableBuilder (rendering), correct fallback chain (Jansi → COLUMNS → default), and all reviewer feedback from Paolo, jsenko, and Carles addressed. LGTM.

vandanayadav7:
@AbdulWasih05 The branch can't be auto-rebased. Could you rebase locally onto main and push? Thanks!

AbdulWasih05:
Thanks @vandanayadav7, rebased and pushed. The branch had a merge commit from an earlier "Update branch", which is why it could not be auto-rebased, so it is now a single commit on top of main.

Two small resolutions against current main: the Jansi natives now share the `quarkus.native.resources.includes` key with the JNA entry the Windows support PR added, instead of overwriting it, and `org.fusesource.jansi:jansi` is pinned to 2.4.0 in `cli/pom.xml` because Quarkus 3.33.3 swapped it for `org.jline:jansi` in the BOM.

vandanayadav7:
@AbdulWasih05 CI was cancelled (not failed), so /retry can't re-trigger it. Could you push an empty commit to restart CI? `git commit --allow-empty -m 'chore: retrigger CI' && git push`

AbdulWasih05:
Hey @vandanayadav7, happy to push, just want to check one thing first. There are two Verify runs on `a5c3763`: `31739212797`, which was cancelled and is the one the bot commented about, and `31739214506`, which finished successfully at 20:22 and is what applied `lifecycle/tested` a few seconds later. Does the merge box show that second run on your side, or only the cancelled one?

Asking because GitHub is in the middle of an incident and my merge status box will not load at all right now, so I would rather not push while webhooks are degraded and risk the PR dropping `lifecycle/tested` without getting it back.

And if you get a chance, would a re-approval be needed once it is green? I think my force-push dismissed yours, so I am not sure the orchestrator can promote this to `lifecycle/ready-to-merge` on its own.

AbdulWasih05:
Hey @vandanayadav7, mind giving this another look when you have a minute? Everything is green on `a5c3763` now, all three CLI native builds plus a fresh License Compliance scan, so the empty commit turned out not to be needed.

Thanks again for helping me chase this one down, I really appreciate you sticking with it.

vandanayadav7 (approved):
Re-approving — code is unchanged from my earlier review, just rebased. LGTM.

EricWittmann:
Thanks @AbdulWasih05 :+1:

Issue #8650, opened 2026-07-15
CLI: Interactive mode with keyboard navigation


vandanayadav7:
## Parent Epic
Part of #7162 (Optional Extra Tasks)

## Description
Add a TUI-style interactive mode where users can navigate through tables using keyboard controls and trigger actions on selected items.

## Scope
- Arrow keys to navigate rows in table output
- Enter/select to trigger actions on a row (e.g. get details, delete)
- Search/filter within the interactive view
- Pagination with keyboard controls
- Exit with `q` or `Esc`
- Consider libraries: Lanterna, JLine, or Jansi for terminal control

## Example Flow
```
$ acr artifact --interactive
  > my-schema        AVRO      2026-07-01   [Enter: view, d: delete, /: search]
    other-schema     JSON      2026-07-02
    api-spec         OPENAPI   2026-06-30
```

## Acceptance Criteria
- Interactive mode activated via `--interactive` flag or `acr interactive` command
- Keyboard navigation works (arrows, enter, escape)
- Actions can be triggered on selected items
- Non-interactive mode remains the default
- GraalVM native image compatibility verified

aniketpandey05:
May i pick this one if it's available 

Thank you

aniketpandey05:
Hi! @vandanayadav7   should we add these  improvements for the CLI TUI (`InteractiveTable`):
1. **`n`/`p` Shortcuts:** Bind `n`/`p` keys for Next/Prev Page alongside PgUp/PgDn 
2. **Prevent Extra API Calls:** Check `hasNextPage` before fetching next page to avoid redundant server calls when already on the last page.
3. **Footer Status:** Update help prompt to show `n`/`p` keys and indicate last page status (`page X (last page)`).

PR #8736, opened 2026-07-19
feat: add --interactive TUI mode to 'acr search artifact' (#8650)

aniketpandey05:
## What is done :

Added --interactive flag to acr search artifact
Built a JLine3-based InteractiveTable component for arrow-key nav, Enter/d selection, q/Esc exit
Added opt-in supportsInteractive()/runInteractive() hooks to AbstractCommand (doesn't affect other commands)
Verified: builds cleanly, flag shows in --help, runs end-to-end against a live registry (empty-result path)
Arrow-key nav  runtime-tested 
## What's left

View/delete actions currently just log, don't call real APIs
Search/filter, pagination in interactive view  not started
Native-image build  not verified


carlesarnal (changes requested):
The annotations are completely broken. Lines 38, 47, 49 in `SearchArtifactsCommand.java` use file paths as annotations (`@app/src/test/...`, `@app/src/main/...`). These need to be actual Java annotations like `@Override` or removed entirely.

Missing newline at EOF in `InteractiveTable.java`. Add one.

The DELETE action is half-implemented. You define it, return it, but never act on it. Either implement the deletion or remove the action entirely. Don't leave users with a non-functional feature.

No tests. Add unit tests for `InteractiveTable` covering navigation, selection, and edge cases. Add integration tests for `SearchArtifactsCommand` with `--interactive` flag.

The `runInteractive()` method duplicates empty-list handling. Remove the early return on line 54-57 since the table handles this.

Specify what exceptions `InteractiveTable.run()` can throw or catch them internally and return null.

Fix the extra whitespace in `pom.xml` line 7 and `AbstractCommand.java` line 11.

aniketpandey05:
Hi @carlesarnal thanks for the review. I accidentally pulled main's commits into my branch so I closed this 

On your feedback:
Broken annotations: interactive mode has moved off SearchArtifactsCommand onto acr artifact (matches the issue's example), so this file reverts to plain search shouldn't reappear.
EOF newline: fixed.
DELETE half-implemented: now fully wired to the real API, verified against a live registry including the disabled-deletion error path.
No tests: still outstanding  extracting InteractiveTable's state (selection/filter/pagination) into a terminal-free class so it's actually unit testable, since the existing harness can't simulate raw terminal input.
Duplicate empty-list handling: keeping the early return in the command (avoids entering raw mode for nothing), but also hardening InteractiveTable so Up/Down safely no-op on empty input.
Exception handling in run(): will document that quit vs. I/O error currently both return null.
Whitespace: will fix.

Still draft  search/filter, pagination, and native-image verification aren't done yet. Will ping you once it's actually ready.

PR #8740, opened 2026-07-20
fix(json): derive readable descriptions for compatibility diff types

### LINKED ISSUE #5066: Improve error messages for compatibility checking

Currently, the error messages are pretty cryptic:

image

And it may be hard to understand why the schemas are not compatible.

### Proposed Solution

Provide a better explanation (maybe with an example from the tests) why the specific compatibility rule exists and when it's broken. Also, which direction the issue is (backward, forward, or both).

Taimoo-r:
## What

Compatibility errors for JSON Schema artifacts currently surface the raw `DiffType` enum constant name to the user, e.g. `OBJECT_TYPE_REQUIRED_PROPERTIES_MEMBER_ADDED`. This makes it hard to understand why two schemas are incompatible (see the screenshot in #5066).

The cause is in `DiffType`: the constructor used by all 172 constants sets `description = this.toString()`, which is just the constant name. The second constructor that accepts an explicit human-readable description has no callers anywhere in the repo.

This PR changes the default so the description is derived from the constant name into a readable sentence, e.g. `OBJECT_TYPE_REQUIRED_PROPERTIES_MEMBER_ADDED` becomes "Object type required properties member added". Every constant improves at once, and the existing explicit-description constructor still works as an override for constants that later get a hand-written explanation.

## Why

Fixes the first part of #5066 (cryptic messages) for all JSON Schema diff types in a small, self-contained change, without altering any public API or the compatibility logic itself.

I intentionally kept this focused. The second part of the issue (reporting whether the incompatibility is backward, forward, or both) is a separate change in `AbstractCompatibilityChecker`, and per my question on the issue this can be followed up with curated per-rule explanations layered on top of this foundation. Happy to adjust the direction here in review.

## Changes

- `DiffType`: derive the default description from the constant name via a small `humanReadable` helper (uses `Locale.ROOT`); the explicit-description constructor is unchanged.
- New unit test `DiffTypeDescriptionTest` asserting specific derived values and verifying no constant exposes a raw enum name.

## Testing

- New `DiffTypeDescriptionTest` passes.
- Existing compatibility tests that assert on `DiffType.getDescription()` remain green (they compare the reported description against the same source, so improving the text does not break them).

Fixes #5066


carlesarnal (dismissed):
Changes look good. The `humanReadable()` method correctly transforms enum names to readable descriptions using `Locale.ROOT` as required. Test coverage is solid with both specific examples and an exhaustive check across all enum values. The javadoc example has a typo (`{ @.claude/agents/code-reviewer.md` should be `{@code`), but that's a documentation formatting issue, not a code problem. No functional issues found.

Taimoo-r:
So whats next @carlesarnal?. Do I've to wait for your further review?

Taimoo-r:
Hi, the previous CI run had a flaky `h2-debezium` integration test timeout (ran for 4h35m waiting for the Debezium connector to register a schema). It's unrelated to the changes in this PR which only touch enum descriptions in `JsonSchemaCompatibilityDifference`. All other checks passed.

I've pushed a retrigger commit - could you please approve the workflow run? Thanks!

paoloantinori (commented):
Clean, well-scoped change with solid coverage — both targeted assertions and an exhaustive sweep that guards every constant against leaking the raw name or underscores. One consideration worth confirming before this lands:

- **User-visible change to error message text.** `description` surfaces in the compatibility-violation responses returned by the REST API (the `causes[].description` field and the diff-type list in the 400 body, as exercised in `GroupsResourceTest`). This PR rewrites the text of all 172 constants at once, from e.g. `SUBSCHEMA_TYPE_CHANGED` to `Subschema type changed`. Any downstream tooling that greps or matches these messages by the old constant name — logs, alerts, docs, examples — would break silently. Worth a release note flagging the change, and worth double-checking whether any docs/examples in the repo or on the docs site hardcode the old uppercase strings.

Minor: the mechanical derivation is correct but loses semantic structure for some constants (`NUMBER_TYPE_INTEGER_REQUIRED_FALSE_TO_TRUE` becomes "Number type integer required false to true" — readable but hard to parse). You already flagged curated per-rule descriptions as a follow-up; for the handful of highest-frequency diff types, an explicit override via the existing second constructor would land cheaply here and improve the most common cases immediately.

Taimoo-r:
Thanks @paoloantinori, both points addressed in 185c41e.

**1. Curated descriptions for high-frequency diff types**

Added explicit descriptions via the existing `DiffType(String, boolean)` constructor for the six most common / hardest-to-parse constants, including the exact one you flagged:

| Constant | Before (fallback) | Now (curated) |
|---|---|---|
| `NUMBER_TYPE_INTEGER_REQUIRED_FALSE_TO_TRUE` | Number type integer required false to true | The value must now be an integer |
| `NUMBER_TYPE_INTEGER_REQUIRED_TRUE_TO_FALSE` | Number type integer required true to false | The value is no longer required to be an integer |
| `OBJECT_TYPE_REQUIRED_PROPERTIES_MEMBER_ADDED` | Object type required properties member added | A new required property was added |
| `OBJECT_TYPE_REQUIRED_PROPERTIES_MEMBER_REMOVED` | Object type required properties member removed | A required property was removed |
| `OBJECT_TYPE_ADDITIONAL_PROPERTIES_TRUE_TO_FALSE` | Object type additional properties true to false | Additional properties are no longer allowed |
| `OBJECT_TYPE_ADDITIONAL_PROPERTIES_FALSE_TO_TRUE` | Object type additional properties false to true | Additional properties are now allowed |

Everything else keeps the readable name-derived fallback, so the follow-up for full per-rule curation still stands. `DiffTypeDescriptionTest` is split so the fallback and the curated overrides are each asserted directly, and the exhaustive check (no constant leaks a raw name or underscore, all 172) is unchanged.

**2. User-visible error-message change**

You're right that this changes response text, and only for the JSON Schema checker: `JsonSchemaCompatibilityDifference` surfaces `getDescription()`, while `ApitomyCompatibilityDifference` still uses `.name()`, so its output is unchanged. The affected fields are `causes[].description` and the diff-type list inside `title` on the 400 response.

I checked the repo for anything that hardcodes the old uppercase constant strings (grepped the constant names across `docs/` and `.md`/`.adoc`/`.json`/`.yaml`/`.yml` plus the Java sources): nothing matches other than the enum definition itself and the tests, so no in-repo docs/examples break.

Since release notes here are generated from milestone issues rather than a committed file, suggested note to attach to the milestone for #5066:

> JSON Schema compatibility violations now report human-readable diff descriptions (e.g. "A new required property was added") instead of raw enum constant names (e.g. `OBJECT_TYPE_REQUIRED_PROPERTIES_MEMBER_ADDED`) in the `causes[].description` field and the `title` of the 400 response. Tooling that matched these messages by the old constant names should be updated. The Apitomy compatibility checker is unaffected.

Happy to adjust any of the curated wordings if you'd prefer different phrasing.

paoloantinori:
Thanks @Taimoo-r — both points addressed nicely. The curated descriptions for the high-frequency types (including `NUMBER_TYPE_INTEGER_REQUIRED_FALSE_TO_TRUE`) read much better than the mechanical fallback, and the exhaustive `noConstantExposesRawEnumName` test guards the rest of the enum against regressing to the raw name. LGTM from my side.

Taimoo-r:
Hi @carlesarnal @jsenko — I synced the branch with main and the suite is green again (`lifecycle/tested`). The earlier sync turned off auto-merge, so it just needs `/auto-merge` re-enabled plus a review or `/skip-review` whenever you have a moment. Happy to make any changes if you'd like something adjusted. Thanks!

jsenko:
Merged, thank you!

PR #8748, opened 2026-07-20
feat(a2a): add watsonx Orchestrate integration for agent discovery and governance

carlesarnal:
## Summary

Add agent discovery endpoints, MCP tools, and sync tooling for watsonx Orchestrate integration, enabling Apicurio Registry to serve as the federated agent catalog that Orchestrate discovers agents from.

- Add `/.well-known/agent-card.json` endpoint alias (Orchestrate discovers agents at this path)
- Add `API_KEY` security scheme to Agent Card builder (Orchestrate supports Bearer, API Key, None)
- Index `protocolVersion` and `url` from agent cards for structured search
- Add SSE transport to MCP server for remote toolkit integration
- Add 4 agent-specific MCP tools: `discover_agents`, `get_agent_card`, `search_mcp_tools`, `get_mcp_tool`
- Add Python sync scripts for bulk polling and event-driven (Kafka) agent catalog sync
- Add 30-test end-to-end suite

## Motivation

watsonx Orchestrate has an Agent Catalog, but its discovery model is URL-based (one agent at a time). Apicurio Registry adds what Orchestrate lacks at scale: version history, compatibility checking, bulk federated discovery, visibility controls (public/entitled/private), and audit trails for agent cards.

This integration has been tested end-to-end against an IBM Cloud watsonx Orchestrate instance — agents stored in Apicurio were successfully imported into Orchestrate via the sync tooling.

## Architecture

```
Apicurio Registry          watsonx Orchestrate           Agent Servers
┌──────────────┐   discover   ┌──────────────────┐   task execution   ┌─────────────┐
│ Agent Cards  │──────────────▶│  Agent Catalog   │──────────────────▶│ LangGraph   │
│ + versions   │  with URLs   │                  │    (direct)       │ CrewAI      │
│ + compat     │              │                  │                   │ BeeAI       │
│ + visibility │              │                  │                   │ Custom      │
└──────────────┘              └──────────────────┘                   └─────────────┘
```

Apicurio is the catalog and governance layer. Agent servers handle task execution. Orchestrate talks to agent servers directly.

### Registry (`app/`)
- `WellKnownResource.java` / `WellKnownResourceImpl.java` — `/.well-known/agent-card.json` alias endpoint
- `A2AConfig.java` — `apicurio.a2a.agent.auth.api-key.enabled` config property
- `RegistryAgentCardBuilder.java` — API_KEY security scheme
- `AgentCardStructuredContentExtractor.java` — index `protocolVersion`, `url` from interfaces
- `WellKnownResourceTest.java` — 3 new tests (alias, canonical match, interface URLs in search)

### MCP Server (`mcp/`)
- `pom.xml` — add `quarkus-mcp-server-sse` dependency (HTTP-based transport for remote integration)
- `AgentsMCPServer.java` — 4 new `@Tool` methods for agent/MCP tool discovery
- `RegistryService.java` — methods for `/.well-known/` endpoints via raw HTTP
- `application.properties` — SSE transport documentation

### Examples (`examples/watsonx-orchestrate/`)
- `sync-agents.py` — bulk polling sync (one-shot or `--watch` mode)
- `register-agent.py` — reads agents from registry, generates Orchestrate YAML, imports via CLI
- `catalog-sync.py` — event-driven sync consuming `registry-events` Kafka topic
- `e2e-test.sh` — 30-test end-to-end suite
- `README.md` — setup and usage guide

## Test plan

- [x] `WellKnownResourceTest` — 24 tests pass (including 3 new Orchestrate-specific tests)
- [x] MCP module compiles with SSE transport
- [x] `sync-agents.py --dry-run` discovers agents from registry
- [x] `register-agent.py --output-dir` generates valid Orchestrate YAML
- [x] End-to-end: agents imported into live IBM Cloud watsonx Orchestrate instance (ca-tor)
- [x] `e2e-test.sh` — 30/30 tests pass

### DIFF
--- a/app/src/main/java/io/apicurio/registry/a2a/A2AConfig.java
+++ b/app/src/main/java/io/apicurio/registry/a2a/A2AConfig.java
@@ -76,6 +76,10 @@ public class A2AConfig {
     @Info(category = CATEGORY_A2A, description = "Default visibility for new Agent Card artifacts (public, entitled, private)", availableSince = "3.0.0")
     String defaultVisibility;
 
+    @ConfigProperty(name = "apicurio.a2a.agent.auth.api-key.enabled", defaultValue = "false")
+    @Info(category = CATEGORY_A2A, description = "Advertise API key authentication in the agent card security schemes", availableSince = "3.3.1")
+    boolean apiKeyAuthEnabled;
+
     @ConfigProperty(name = "apicurio.a2a.agent.protocol-version", defaultValue = "1.0")
     @Info(category = CATEGORY_A2A, description = "A2A protocol version supported by the registry agent", availableSince = "3.0.0")
     String protocolVersion;
@@ -146,5 +150,9 @@ public class A2AConfig {
 
     public String getDefaultVisibility() {
         return defaultVisibility;
+    }
+
+    public boolean isApiKeyAuthEnabled() {
+        return apiKeyAuthEnabled;
     }
 }
--- a/mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
@@ -465,6 +465,67 @@ public class RegistryService {
         return client.admin().config().properties().byPropertyName(propertyName).get();
     }
 
+    // ---- Well-known agent/MCP tool endpoints (no SDK support, use raw HTTP) ----
+
+    private static final java.net.http.HttpClient wellKnownHttpClient = java.net.http.HttpClient.newBuilder()
+            .connectTimeout(java.time.Duration.ofSeconds(10))
+            .build();
+
+    public String searchAgentCards(String name, String skill, String capability) throws Exception {
+        StringBuilder url = new StringBuilder(rawBaseUrl).append("/.well-known/agents?limit=50");
+        if (name != null && !name.isBlank()) {
+            url.append("&name=").append(java.net.URLEncoder.encode(name, StandardCharsets.UTF_8));
+        }
+        if (skill != null && !skill.isBlank()) {
+            url.append("&skill=").append(java.net.URLEncoder.encode(skill, StandardCharsets.UTF_8));
+        }
+        if (capability != null && !capability.isBlank()) {
+            url.append("&capability=").append(java.net.URLEncoder.encode(capability, StandardCharsets.UTF_8));
+        }
+        return fetchWellKnownEndpoint(url.toString());
+    }
+
+    public String getAgentCard(String groupId, String artifactId) throws Exception {
+        String url = rawBaseUrl + "/.well-known/agents/"
+                + java.net.URLEncoder.encode(groupId, StandardCharsets.UTF_8) + "/"
+                + java.net.URLEncoder.encode(artifactId, StandardCharsets.UTF_8);
+        return fetchWellKnownEndpoint(url);
+    }
+
+    public String searchMcpTools(String name, String parameter) throws Exception {
+        StringBuilder url = new StringBuilder(rawBaseUrl).append("/.well-known/mcp-tools?limit=50");
+        if (name != null && !name.isBlank()) {
+            url.append("&name=").append(java.net.URLEncoder.encode(name, StandardCharsets.UTF_8));
+        }
+        if (parameter != null && !parameter.isBlank()) {
+            url.append("&parameter=").append(java.net.URLEncoder.encode(parameter, StandardCharsets.UTF_8));
+        }
+        return fetchWellKnownEndpoint(url.toString());
+    }
+
+    public String getMcpTool(String groupId, String artifactId) throws Exception {
+        String url = rawBaseUrl + "/.well-known/mcp-tools/"
+                + java.net.URLEncoder.encode(groupId, StandardCharsets.UTF_8) + "/"
+                + java.net.URLEncoder.encode(artifactId, StandardCharsets.UTF_8);
+        return fetchWellKnownEndpoint(url);
+    }
+
+    private String fetchWellKnownEndpoint(String url) throws Exception {
+        var request = java.net.http.HttpRequest.newBuilder()
+                .uri(java.net.URI.create(url))
+                .header("Accept", "application/json")
+                .timeout(java.time.Duration.ofSeconds(30))
+                .GET()
+                .build();
+        var response = wellKnownHttpClient.send(request,
+                java.net.http.HttpResponse.BodyHandlers.ofString());
+        if (response.statusCode() >= 400) {
+            throw new ToolCallException("Registry returned HTTP " + response.statusCode()
+                    + " for " + url);
+        }
+        return response.body();
+    }
+
     public void updateConfigurationProperty(String propertyName, String propertyValue) {
         if (config.safeMode() && !List.of(
                 "apicurio.rest.mutability.artifact-version-content.enabled"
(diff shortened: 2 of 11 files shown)


mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
```diff
@@ -465,6 +465,62 @@ public ConfigurationProperty getConfigurationProperty(String propertyName) {
...
+
+    public String getMcpTool(String groupId, String artifactId) throws Exception {
+        String url = rawBaseUrl + "/.well-known/mcp-tools/"
+                + java.net.URLEncoder.encode(groupId, StandardCharsets.UTF_8) + "/"
+                + java.net.URLEncoder.encode(artifactId, StandardCharsets.UTF_8);
+        return fetchWellKnownEndpoint(url);
+    }
+
+    private String fetchWellKnownEndpoint(String url) throws Exception {
+        var httpClient = java.net.http.HttpClient.newHttpClient();
```
paoloantinori: Two concerns here:

1. **No timeout.** `HttpClient.newHttpClient()` and the `HttpRequest` both omit timeouts, so a hung registry blocks the MCP tool indefinitely. The Python sync scripts use `timeout=30` — the Java side should match (`.connectTimeout(...)` on the client and `.timeout(Duration.ofSeconds(30))` on the request).
2. **New client per call.** `HttpClient` owns a connection pool and executor and is designed to be reused; constructing one per tool invocation is wasteful when these tools may fire frequently. Consider a single shared `HttpClient` field built once.

examples/watsonx-orchestrate/register-agent.py
```diff
@@ -0,0 +1,179 @@
...
+    orchestrate_def = {
+        "spec_version": "v1",
+        "kind": "external",
+        "name": safe_name,
+        "title": name,
+        "nickname": safe_name,
+        "provider": "external_chat/A2A/0.3.0",
+        "description": agent_card.get("description", ""),
+        "api_url": api_url,
+        "auth_scheme": "NONE",
```
paoloantinori: Hardcoding `auth_scheme: "NONE"` means this converter ignores the agent card's `securitySchemes`. Given this same PR adds API-key auth advertising on agent cards, agents that declare Bearer/API-key auth get registered to Orchestrate as unauthenticated and will fail at runtime. Worth reading the card's `securitySchemes` and mapping to the Orchestrate auth scheme, or at minimum warning when a card declares a non-None scheme.

app/src/main/java/io/apicurio/registry/a2a/A2AConfig.java
```diff
@@ -76,6 +76,10 @@ void validate() {
     @Info(category = CATEGORY_A2A, description = "Default visibility for new Agent Card artifacts (public, entitled, private)", availableSince = "3.0.0")
     String defaultVisibility;
 
+    @ConfigProperty(name = "apicurio.a2a.agent.auth.api-key.enabled", defaultValue = "false")
+    @Info(category = CATEGORY_A2A, description = "Advertise API key authentication in the agent card security schemes", availableSince = "3.0.0")
```
paoloantinori: `availableSince = "3.0.0"` on a property being introduced by this PR is misleading for the generated docs — it implies the property has existed since 3.0.0. Should be the version that actually ships this change.

paoloantinori (commented):
Nice integration — the `/.well-known/agent-card.json` alias, the API-key security scheme, indexing of `url`/`protocolVersion` into structured content, and the SSE MCP transport all hang together coherently, and the alias/equality tests are a nice touch. A few things worth a closer look (the diff was truncated at ~4000 lines, so claims are scoped to what's visible):

- **`fetchWellKnownEndpoint` has no timeout and builds a fresh `HttpClient` per call** (`RegistryService.java`). `HttpClient.newHttpClient()` spins up a new client (with its own connection pool/executor) on every MCP tool invocation, and neither the client nor the `HttpRequest` sets a timeout — so a slow/stuck registry hangs the tool indefinitely. The Python sync scripts use `timeout=30`; the Java path should match. Recommend hoisting a single shared `HttpClient` (built once with `.connectTimeout(Duration.ofSeconds(...))`) and adding `.timeout(Duration.ofSeconds(30))` to each request.

- **`register-agent.py` hardcodes `"auth_scheme": "NONE"`** regardless of the agent card's `securitySchemes`. One of the headline changes here is advertising API-key auth on agent cards (`RegistryAgentCardBuilder`), but agents registered into Orchestrate via this script are always marked unauthenticated — so any card requiring Bearer/API-key auth registers fine but fails on first call. Worth mapping the card's `securitySchemes` to the Orchestrate auth scheme (at least Bearer/API-key), or logging a warning when a card declares a non-None scheme.

- **`availableSince = "3.0.0"` on a brand-new config property** (`apicurio.a2a.agent.auth.api-key.enabled` in `A2AConfig.java`). Since this property is introduced by this PR, `availableSince` should reflect the release that actually ships it, not 3.0.0 — otherwise the generated config docs imply it has existed since 3.0.0.

- **Worth confirming `/.well-known/mcp-tools` and `/.well-known/mcp-tools/{group}/{id}` actually exist.** `search_mcp_tools` and `get_mcp_tool` in `RegistryService` call these routes, but the diff only adds the `agent-card.json` alias to `WellKnownResource`. If the mcp-tools routes aren't defined elsewhere in this change, both MCP tools will always 404 and throw `ToolCallException`. Flagging in case the truncation hid them.

- **Well-known HTTP calls carry no credentials.** `fetchWellKnownEndpoint` is unauthenticated, which is fine for public discovery, but `getAgentCard(groupId, artifactId)` reaches arbitrary groups — on a registry where non-public artifacts require auth it returns 401 and the tool fails. Probably acceptable for v1, but worth a line in the tool descriptions or the example README.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
carlesarnal: Good catch on both points. Fixed in decf6b6a2b — shared static `HttpClient` with `connectTimeout(10s)` built once, plus `timeout(30s)` on each request.

examples/watsonx-orchestrate/register-agent.py
carlesarnal: Fixed in decf6b6a2b — added `_map_security_schemes()` that reads the card's `securitySchemes` and maps `http/Bearer` → `BEARER_TOKEN`, `apiKey` → `API_KEY`, with a warning when schemes don't map. Tokens are set to placeholder values (`<BEARER_TOKEN_REQUIRED>`) since actual credentials need to be provided by the user.

app/src/main/java/io/apicurio/registry/a2a/A2AConfig.java
carlesarnal: Right, fixed to `3.3.1` in decf6b6a2b.

carlesarnal:
Thanks for the review @paoloantinori — all three code fixes are pushed in decf6b6a2b.

On the remaining observations:

- **`/.well-known/mcp-tools` routes**: Already exist in `WellKnownResource.java` (lines 129-160) and `WellKnownResourceImpl.java` (lines 483-549) — added in an earlier PR with MCP tools support. The diff was truncated.

- **Unauthenticated well-known HTTP calls**: Intentional for v1 — `/.well-known/agents` is a discovery endpoint typically served without auth. For `getAgentCard()` on auth-protected registries, forwarding the MCP server's configured credentials would be a natural follow-up.

paoloantinori:
Thanks @carlesarnal — that clears it up. The `mcp-tools` routes being pre-existing (just outside the truncated diff) explains it, and keeping the `agent-card.json` alias unauthenticated to match the canonical `/.well-known/agent.json` is the right call for a discovery endpoint. The three fixes in decf6b6 look good. Nothing else from me.

paoloantinori (commented):
Thanks @carlesarnal — the shared static `HttpClient` with explicit timeouts and the security-scheme mapping both look correct. CI failed on this head though — could you check the [workflow run](https://github.com/Apicurio/apicurio-registry/actions/runs/29746755297) and push a fix?

Issue #7162, opened 2026-01-20
Epic: Apicurio Registry CLI Standard Features


jsenko:
This epic tracks the development of standard features for the Apicurio Registry CLI.

See the `cli/README.md` file for build and installation instructions, usage examples, and development guidelines.

### Phase 1 - macOS Support
- #7163 Add macOS support to Apicurio Registry CLI

### Phase 1.5 - Quarkus Native Build
- #7618 Quarkus-based CLI with native build
- #7615 Split SDK HTTP adapters into separate modules for native image compatibility
- #7640 CLI: Complete CDI conversion (Config singleton, Client static pattern)
- #7641 CLI: Verbose mode does not affect Quarkus/JBoss logging
- #7642 CLI: Implement automatic update with OS/architecture detection
- #7612 Move JacksonDateTimeCustomizer from common to app module
- #7628 Optimize build structure: introduce essential vs. full build profiles

### Phase 2 - Basic REST API CRUD operations
- #7164 Support group management in the CLI
- #7165 Extended version command in the CLI
- #7166 Support artifact management in the CLI
- #7167 Support version management in the CLI
- #7168 Support global rules management in the CLI
- #7169 Support artifact rules management in the CLI
- #7170 Support group rules management in the CLI

## Phase 1 - macOS Support

1. Add support for macOS (#7163)
    - Currently, the CLI only supports Linux with bash shell, so it relies on bash-specific features, such as
      `~/.bashrc` modification and bash completions. Adding support for macOS would involve:
        - Testing the existing installation process on macOS.
        - Updating the `acr install` command to detect macOS and modify `~/.zshrc` instead of `~/.bashrc`.
        - The completions should work (see [picocli zsh documentation](https://picocli.info/autocomplete.html#_zsh)) but need to be tested.
    - **Acceptance Criteria:**
        - CLI installs successfully on macOS
        - Shell completions work in zsh
        - All existing commands function correctly on macOS

### Phase 1.5 - Quarkus Native Build
- #7618 Quarkus-based CLI with native build
- #7615 Split SDK HTTP adapters into separate modules for native image compatibility

These are follow-up tasks from the Quarkus native CLI conversion (#7618, #7615):

1. Complete CDI conversion (#7640) — Convert `Config` singleton and `Client` static pattern to proper CDI beans
2. Fix verbose mode logging (#7641) — `--verbose` flag doesn't affect Quarkus/JBoss Logging output
[...]

PR #8749, opened 2026-07-20
refactor(cli)!: use long-only options for name, description, and type. Fixes #8024

### LINKED ISSUE #8024: CLI: Review short property names across all commands
## Description

Short property names like `-a`, `-b`, `-c` are a limited resource. Review all commands and consider using long names instead of short names for less common properties.

Example: `-n`/`--name` is used for artifact/version name but may conflict with other uses.

Search commands already had `-d`, `-n`, `-t` short options removed in PR #7973. This issue tracks reviewing the remaining commands.

## Parent Epic

- #7162

Priyanshubhartistm:
## Summary

Short single-letter options are a limited resource, so they should be reserved for the most common properties. PR #7973 removed the `-d`, `-n`, and `-t` short options from the search commands for this reason. This change applies the same treatment to the remaining artifact, version, and group commands.

Fixes #8024.

## Changes

Remove the short options for the less common metadata properties, keeping the long forms:

- `-n` removed from `--name` (artifact create/update, version create/update)
- `-d` removed from `--description` (artifact create/update, version create/update, group create/update)
- `-t` removed from `--type` (artifact create)

Primary identifier and I/O options (`-g/--group`, `-a/--artifact`, `-v/--version`, `-f/--file`, `-l/--label`, `-o/--output-type`) are left unchanged, consistent with the options the search commands retained.

Also update the `acr artifact create` example in `cli/README.md` to use `--type` instead of the removed `-t`.

## Breaking change

This removes short options that existed in previous releases, so scripts using `acr artifact create ... -t AVRO -n <name> -d <desc>` (and the equivalent version/group invocations) will need to switch to `--type` / `--name` /
`--description`. This is an intentional, clean break rather than a deprecation cycle, consistent with the hard removal done in #7973. The CLI is still stabilizing as part of the native-build / Phase 2 work tracked in the epic (#7162) and is not yet treated as a stable interface, which is why no deprecation window is included.

## Notes

- No behavior change beyond dropping the short aliases; the long options are  unchanged.
- Existing CLI tests already use the long option forms, so no test changes are  required.
- A repo-wide grep of `cli/README.md`, `docs/`, and the example scripts found no  other `-t`/`-n`/`-d` usages in artifact/version/group command examples; the single README example was the only one and it is updated here.
- The shell completion script is generated at build time and is not checked in, so it picks up the change automatically.

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
@@ -46,7 +46,7 @@ public class ArtifactCreateCommand extends AbstractCommand {
     private String groupId;
 
     @Option(
-            names = {"-t", "--type"},
+            names = {"--type"},
             description = "Artifact type (e.g. AVRO, PROTOBUF, JSON, OPENAPI, ASYNCAPI, GRAPHQL, KCONNECT, WSDL, XSD, XML)."
     )
     private String artifactType;
@@ -58,13 +58,13 @@ public class ArtifactCreateCommand extends AbstractCommand {
     private String file;
 
     @Option(
-            names = {"-n", "--name"},
+            names = {"--name"},
             description = "Provide artifact name."
     )
     private String name;
 
     @Option(
-            names = {"-d", "--description"},
+            names = {"--description"},
             description = "Provide artifact description."
     )
     private String description;
--- a/cli/README.md
+++ b/cli/README.md
@@ -327,7 +327,7 @@ acr artifact
 **Create an artifact:**
 ```bash
 acr artifact create <artifact-id> -g <group-id> -f <file-path>
-acr artifact create <artifact-id> -g <group-id> -f <file-path> -t <artifact-type>
+acr artifact create <artifact-id> -g <group-id> -f <file-path> --type <artifact-type>
 
 # Read content from stdin:
 cat schema.json | acr artifact create <artifact-id> -g <group-id> -f -
--- a/cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactUpdateCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactUpdateCommand.java
@@ -35,13 +35,13 @@ public class ArtifactUpdateCommand extends AbstractCommand {
     private String groupId;
 
     @Option(
-            names = {"-n", "--name"},
+            names = {"--name"},
             description = "Updated artifact name."
     )
     private String name;
 
     @Option(
-            names = {"-d", "--description"},
+            names = {"--description"},
             description = "Updated artifact description."
     )
     private String description;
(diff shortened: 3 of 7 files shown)


jsenko:
LGTM from me @vandanayadav7 Can you take over and review and merge at your discretion? Thanks.

cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
```diff
@@ -46,7 +46,7 @@ public class ArtifactCreateCommand extends AbstractCommand {
     private String groupId;
 
     @Option(
-            names = {"-t", "--type"},
+            names = {"--type"},
```
paoloantinori, line 49: Dropping `-t` here (and `-n`/`-d` across the other commands) is a breaking change for anyone scripting the CLI today. If there's no deprecation window planned, a release-notes line flagging the removed short options would help users migrate. If the CLI is considered pre-stable, noting that explicitly somewhere would make the breakage intentional rather than incidental.

paoloantinori (commented):
Clean, consistent with the precedent set in #7973, and the README example is updated correctly. One thing the PR description doesn't address that's worth a conscious decision before merge:

- **This is a breaking change for existing CLI users.** Anyone scripting `acr artifact create ... -t AVRO -n ... -d ...` will see those invocations fail after this lands, with no deprecation window. For a CLI tool that's usually worth at least a release-notes line (and ideally a deprecation cycle where the short form warns but still works). If the project treats the CLI as not-yet-stable, that's a fine reason to skip the cycle — just worth stating explicitly so the breaking change is intentional rather than accidental.
- Minor: worth a quick grep across `cli/README.md`, the website docs, and any example scripts in the repo for remaining `-t`/`-n`/`-d` usages in artifact/version/group contexts — only one README example was updated, and stale docs referencing removed options are a common fallout from this kind of change.

Priyanshubhartistm:
@paoloantinori ,Thanks for the review, both points are fair

Docs sweep: I grepped the whole repo - `cli/README.md`, `docs/`, and the example scripts  for any remaining `-t`/`-n`/`-d` usage in artifact/version/group command examples. The one README example was the only real occurrence; everything else that matched is unrelated (`curl -d`, `docker -t`, `pg_dump -d`, bash `[ -n ... ]`, etc.). So the docs are consistent now.

Breaking change: Agreed, and worth stating explicitly - I've updated the PR title and description to call it out. I kept it as a clean removal (no deprecation cycle) to stay consistent with the hard removal in #7973, since the CLI is still stabilizing under the native-build / Phase 2 work in #7162 and isn't treated as a stable interface yet.  let me know your preference.

jsenko:
> Breaking change: Agreed, and worth stating explicitly - I've updated the PR title and description to call it out. I kept it as a clean removal (no deprecation cycle) to stay consistent with the hard removal in https://github.com/Apicurio/apicurio-registry/pull/7973, since the CLI is still stabilizing under the native-build / Phase 2 work in https://github.com/Apicurio/apicurio-registry/issues/7162 and isn't treated as a stable interface yet. let me know your preference.

This is fine, the CLI is still in the process of being stabilized.

paoloantinori:
Thanks @jsenko — calling it out in the title/description and keeping a clean removal (no deprecation cycle) consistent with #7973 is the right call. Nothing else from my side.

paoloantinori:
Thanks @Priyanshubhartistm — the docs sweep across `cli/README.md`, `docs/`, and the example scripts is thorough, and the breaking change is now clearly called out in the PR title and description. Both points are settled. From my side this is good to go.

vandanayadav7 (approved):

PR #8753, opened 2026-07-20
ci: add shard guard and split app-security from app-other

paoloantinori:
## Summary

- Adds a shard coverage guard script that validates every app test class is claimed by exactly one unit-test shard, catching silent coverage gaps from the negative filter pattern in app-other
- Splits `app-security` shard out of `app-other`, moving auth, rbac, tls, and cors test classes (~24 classes) to reduce `app-other` wall-clock from ~13min

## Related Issues

- Backlog REG-15 (shard guard)
- Backlog REG-14 (split app-other)

## Changes

- **New file**: `.github/scripts/check-shard-coverage.sh` — enumerates test classes, matches against shard filter patterns, fails on zero-match or multi-match
- **Modified**: `.github/workflows/verify.yaml` — runs the guard in the Lint and Validate job
- **Modified**: `.github/workflows/verify-unit-tests.yaml` — adds `app-security` shard, updates `app-other` exclusions

## Testing

- Guard script passes locally: 229 test classes, 6 shards, zero overlaps
- CI will validate: app-security shard runs green, app-other timing drops, guard step passes
- Compare app-other wall-clock before/after across 2 runs

### DIFF
--- a/.github/scripts/check-shard-coverage.sh
+++ b/.github/scripts/check-shard-coverage.sh
@@ -12,8 +12,6 @@ APP_TEST_DIR="app/src/test/java"
 # ── Positive-match shards ────────────────────────────────────────────────────
 # Each entry is: SHARD_NAME  PATTERN1 PATTERN2 …
 # Patterns use Java-style "**" globs (translated to regex below).
-# The catch-all shard (app-other) is NOT listed here — it covers whatever the
-# positive shards don't, so it is derived automatically.
 SHARDS=(
   "app-rest|io.apicurio.registry.noprofile.rest.**|io.apicurio.registry.noprofile.ccompat.**"
   "app-sql|io.apicurio.registry.storage.impl.sql.**|io.apicurio.registry.storage.impl.search.**|io.apicurio.registry.storage.impl.polling.**|io.apicurio.registry.storage.impl.readonly.**|io.apicurio.registry.storage.decorator.**|io.apicurio.registry.storage.dto.**|io.apicurio.registry.event.sql.**"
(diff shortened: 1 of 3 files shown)


carlesarnal:
I like the goal of this, but the approach here might become stale very quickly. Any other ideas?

EricWittmann (commented):
The shard guard is a great idea — catching silent coverage gaps from the negative filter pattern is a real problem worth solving. Agreeing with @carlesarnal's concern about staleness though.

The core issue is that shard patterns are duplicated: defined in `verify-unit-tests.yaml` (the real shards) and copied into `check-shard-coverage.sh` (the guard). The `SYNC` comment helps, but someone will eventually forget.

**Suggestion: make the guard script read the shard patterns directly from the workflow file.** `yq` is pre-installed on GitHub-hosted runners, so the script could:

1. Parse `verify-unit-tests.yaml` with `yq` to extract each shard's `name` and `test-filter` from the matrix
2. Distinguish positive shards (`-Dtest=io.apicurio...`) from the negative-pattern shard (`-Dtest=!io.apicurio...`)
3. Run the same validation logic against the discovered test classes

This makes the workflow file the single source of truth — adding a new shard to the matrix automatically includes it in the guard's validation. No duplication, no `SYNC` comment needed.

The tradeoff is a dependency on the workflow YAML structure remaining stable (standard GitHub Actions matrix format), but that's unlikely to change and would fail obviously if it did.

carlesarnal:
@paoloantinori any plans to continue with this? If not I'll close it.

paoloantinori:
Fair question — no active plans to push this forward right now, and the branch needs a rebase against main. If it's cleaner to close, go ahead: the guard script is preserved in the branch and I can re-submit a refreshed version when I get a cycle.

PR #8761, opened 2026-07-20
fix(cli): improve error handling and list supported types for missing artifact options (#8023)

### LINKED ISSUE #8023: CLI: Better error when creating artifact without content or type
## Description

When creating an artifact without specifying content or artifact type, the error message is unclear:

```
Error creating artifact: MissingRequiredParameterException: Request is missing a required parameter: body.artifactType
```

## Expected Behavior

- Show a clear error message when `--type` is missing
- List supported artifact types when the type is invalid or missing
- Investigate auto-detection of `--content-type` for JSON and YAML files
- Consider whether this should be handled server-side

## Parent Epic

- #7162

NobleCoder69:
Fixes #8023

### Description
Improves CLI validation when creating artifacts without specifying required parameters (`--type` or `--file`).

### Changes Made
- Added clear validation checks throwing `IllegalArgumentException` when `--type` or `--file` options are missing.
- Dynamically extracts supported artifact types from `io.apicurio.registry.types.ArtifactType` constants using reflection to list them in the error output.
- Retained the default `--type` option description in CLI flags as requested by the maintainer.

Closes #8023


cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
```diff
@@ -71,7 +77,7 @@ public class ArtifactCreateCommand extends AbstractCommand {
 
     @Option(
             names = {"-l", "--label"},
-            description = "Provide a list of artifact labels (format: key=value or key). Use \\= to include = in a key."
+            description = "Provide a list of artifact labels (format: key=value or key). Use \= to include = in a key."
```
paoloantinori: This hunk changes `\\=` to `\=` in a Java string literal. `\=` is not a valid escape sequence in Java, so this line should fail to compile ("illegal escape character"). It also looks unrelated to issue #8023. Probably an accidental editor rewrite — I'd revert this line to `Use \\= to include = in a key.` to match `main`.

cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
```diff
@@ -93,13 +99,40 @@ public class ArtifactCreateCommand extends AbstractCommand {
...
+            throw new IllegalArgumentException(
+                "Missing required option '--type'. Supported artifact types are: " + supportedTypesList
+            );
+        }
+
+        // 2. Validate missing content/file
+        if (isBlank(file)) {
+            throw new IllegalArgumentException(
+                "Missing required option '--file' (or '-' for stdin) to provide artifact content."
+            );
```
paoloantinori: Making `--file` mandatory here breaks `testArtifactCreateWithoutFile` in `cli/src/test/java/io/apicurio/registry/cli/ArtifactCommandTest.java`, which currently creates an artifact with no content. On `main`, the entire `readContent` / `firstVersion` block is wrapped in `if (!isBlank(file))`. If metadata-only artifact creation is still meant to be supported, this check (and the unconditional `firstVersion` block below) should be removed. Worth confirming the intended behavior with the maintainer.

cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
```diff
@@ -93,13 +99,40 @@ public class ArtifactCreateCommand extends AbstractCommand {
...
+                    } catch (IllegalAccessException e) {
+                        return field.getName();
+                    }
+                })
+                .sorted()
+                .collect(Collectors.joining(", "));
+
+        // 1. Validate missing --type
+        if (isBlank(artifactType)) {
+            throw new IllegalArgumentException(
```
paoloantinori: `IllegalArgumentException` thrown from `run(...)` is caught by the generic `catch (Exception ex)` branch in `AbstractCommand.call()`, which logs at ERROR and prefixes the message with `Unexpected error:`. For expected user-input validation, `new CliException(message, CliException.VALIDATION_ERROR_RETURN_CODE)` produces a cleaner message and the right exit code (2).

paoloantinori (commented):
Overall the direction is good — failing fast with a clear message and the list of supported types matches what issue #8023 asks for. A few concerns below, the first one looks like a build breaker unrelated to the issue and worth checking before merge.

- **Possible compile error on the `--label` description (line 80).** The diff changes `Use \\= to include =` to `Use \= to include =`. In a Java string literal `\=` is not a valid escape sequence (only `\b \t \n \f \r \" \' \\` and octal are allowed), so this line should fail to compile with "illegal escape character". It is also unrelated to #8023 and looks like an accidental editor change — easiest is to revert this hunk and keep the original `\\=`. Oddly the SonarCloud check on this SHA is green; I have not been able to run `javac` from here to confirm, so please double-check on a real build.

- **Making `--file` mandatory is a behavior change, not just an error-message improvement.** On `main` the file/content block is gated by `if (!isBlank(file))`, so artifact creation without content was supported. The existing test `testArtifactCreateWithoutFile` in `cli/src/test/java/io/apicurio/registry/cli/ArtifactCommandTest.java` relies on this (it creates `no-content-artifact` with only `--type`, `--name`, etc.). After this PR that call throws on line 124-127 and the test fails. Issue #8023 only asks for clearer error messaging — if metadata-only artifact creation is still meant to be supported, the new `--file` check and the unconditional `firstVersion` block should be dropped; if it is being intentionally deprecated, that is a separate decision worth calling out explicitly and the test should be updated accordingly.

- **Exception type for user input validation.** Throwing `IllegalArgumentException` from `run(...)` falls into the generic `catch (Exception ex)` branch in `AbstractCommand.call()` (https://github.com/Apicurio/apicurio-registry/blob/main/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java), which logs at ERROR level and prints `Unexpected error: ...`. For an expected user-facing validation failure, `new CliException(message, CliException.VALIDATION_ERROR_RETURN_CODE)` gives a cleaner message, the right exit code (2 instead of 1), and avoids noisy stack traces in the log.

- **Scope gap: invalid `--type` is not validated.** Issue #8023 explicitly asks to "list supported artifact types when the type is invalid **or** missing". Currently only the missing case is handled. Since the supported set is already computed via reflection on lines 102-114, a cheap follow-up is to also reject `--type` values that are not in that set with the same message.

- **No tests for the new paths.** It would be good to add assertions for the two new failure modes (missing `--type`, missing `--file`) and for the format of the "Supported artifact types are: ..." suffix, so the messaging does not silently regress later.

NobleCoder69:
hi @paoloantinori Thanks for the detailed review! I addressed all your feedback in the latest commits and synced with latest main:

Restored \\= in the --label description string.

Made --file optional so metadata-only artifact creation works as expected.

Switched validation exceptions to CliException with VALIDATION_ERROR_RETURN_CODE.

Added validation for invalid --type inputs alongside missing ones.

Ready for workflow approval and maintainer review!

paoloantinori:
Thanks @NobleCoder69 — all four points are addressed cleanly:

- `--label` description is back to the original `\\=`, so the illegal-escape issue is gone.
- `--file` is optional again (`if (!isBlank(file))` restored), which keeps metadata-only creation and `testArtifactCreateWithoutFile` working.
- Validation now throws `CliException` with `VALIDATION_ERROR_RETURN_CODE` — correct exit code and message formatting.
- Invalid `--type` is now rejected too, closing the scope gap from #8023.

One small note for awareness (not blocking): the reflection-based enumeration walks `ArtifactType.class.getDeclaredFields()` filtering on `public static String`. Today every constant on that class matches that filter, so the listed set will be correct. If a non-constant `public static String` field ever gets added to `ArtifactType` it would leak into the message; easy to revisit if/when that happens, no action needed now.

From my side this is good to go — deferring the merge/approval call to a maintainer.

paoloantinori (commented):
Thanks @NobleCoder69 — seen the fixes. The restored `\\=` in the label description and the sync with main look correct. Waiting on maintainer `/accept` to trigger CI.

NobleCoder69:
Hi @EricWittmann Fixed the CI failure by adding a readiness wait loop for the mock Nginx container in cli/src/test/scripts/test-update.sh. Ready for workflow approval!

jsenko (commented):
Review of the current head (`709759c`). Good direction on fixing #8023 — clear error with supported types is what the issue asks for. A few things need work:

**1. `--type` should not be required when `--file` is omitted (must-fix)**

The validation unconditionally rejects missing `--type`, but on `main` you can create a metadata-only artifact:
```bash
acr artifact create myartifact -g mygroup --name "My Artifact"
```
This creates an artifact stub with no content and no type. Making `--type` always required breaks this use case. The validation should only require `--type` when `--file` is provided (since that's when the server needs it):
```java
if (!isBlank(file) && isBlank(artifactType)) {
    throw new CliException("...", VALIDATION_ERROR_RETURN_CODE);
}
```

**2. Reflection for type discovery is fragile and GraalVM-incompatible (must-fix)**

`ArtifactType.class.getDeclaredFields()` with reflection:
- May return nothing in a GraalVM native image without explicit reflection config (the CLI ships as a native binary)
- Runs on every `artifact create` invocation — wasteful
- Is unnecessary since the types are well-known constants

Use the constants directly:
```java
private static final Set<String> SUPPORTED_TYPES = Set.of(
    ArtifactType.AVRO, ArtifactType.PROTOBUF, ArtifactType.JSON,
    ArtifactType.OPENAPI, ArtifactType.ASYNCAPI, ArtifactType.GRAPHQL,
    ArtifactType.KCONNECT, ArtifactType.WSDL, ArtifactType.XSD, ArtifactType.XML,
    ArtifactType.AGENT_CARD, ArtifactType.MCP_TOOL, ArtifactType.ICEBERG_TABLE,
    ArtifactType.ICEBERG_VIEW, ArtifactType.OPENRPC, ArtifactType.MODEL_SCHEMA,
    ArtifactType.PROMPT_TEMPLATE, ArtifactType.ODCS_CONTRACT, ArtifactType.THRIFT
);
```
This compiles to a constant, works in native image, and is zero-cost at runtime. When a new type is added to `ArtifactType`, this list needs updating — but that's also true for the `@Option(description = "...")` which already lists example types.

**3. No unit tests for the new validation (should-fix)**

The shell script `test-update.sh` changes are unrelated (container readiness fix). The new validation logic has no test coverage. Add tests in `ArtifactCommandTest.java`:
- Missing `--type` with `--file` → error with supported types listed
- Invalid `--type` → error with supported types listed
- Missing `--type` without `--file` → should succeed (metadata-only)
- Valid `--type` → succeeds as before

carlesarnal:
I'm closing this as stale.

Issue #8709, opened 2026-07-17
Example For Event Driven Architecture is not compatible with apicurio-registry version 3


AyachiMishra:
## Description

**Registry Version**: latest-release (3.x)

**Persistence type**: sql

The `examples/event-driven-architecture` example is no longer compatible with the current Apicurio Registry 3.x release.

The example still references deprecated Registry 2.x configuration and container images, preventing it from running successfully with the latest release.

---

## Environment

- Docker Compose
- Apache Kafka
- ZooKeeper
- Kafka Connect (Debezium)
- PostgreSQL (SQL storage)
- Kafka UI
- Apicurio Registry latest-release (3.x)

---

## Steps to Reproduce
(I followed exactly the same steps that were in apicurio-registry/examples/event-driven-architecture/README.md)
1. Clone the repository.
2. Navigate to `examples/event-driven-architecture`.
3. Start the example:

   ```bash
   docker compose up -d
   ```

4. Create the Debezium connector using the provided `studio-connector.json`.
5. Observe that the example is not compatible with the current Registry release.

---

### Expected

The example should start successfully and demonstrate the complete event-driven workflow:

- All containers start successfully.
- The Debezium connector can be created.
- CDC events are published to Kafka.
- Schemas are automatically registered in Apicurio Registry.
- Registered artifacts are visible in the Registry UI.

### Actual

The example requires several updates to work with the current Registry release, including:

- Registry API endpoints still reference `/apis/registry/v2`.
- The deprecated `apicurio/apicurio-registry-sql` image is used.
- The Debezium Connect image is outdated.
- PostgreSQL `latest` currently resolves to a newer major version that prevents the example from working correctly. (THIS IS QUITE IMPORTANT Please check with other examples too.....)
- Registry UI is now provided as a separate service.

After updating these components (I have already created the PR for this), the example runs successfully end-to-end.

---

## Logs

Not applicable.

It is more of an outdated example configuration rather than a single log which would point to a mistake. 
However, for assurance: 
- Docker Compose starts successfully.
- Kafka Connect starts successfully.
- The Debezium connector is created successfully.
- CDC events are published to Kafka.
- Schemas are automatically registered in Apicurio Registry.
- The Registry UI displays the registered artifacts.

Shubh2-0:
[...]

Issue #7395, opened 2026-02-13
Schema caching not working with v3 API - API call made for every record


carlesarnal:
## Description

When using Apicurio Registry v3 serdes/converter with Debezium CDC, the client makes an API call to the registry for **every record** instead of caching the schema. This causes severe performance degradation during initial snapshot operations.

## Environment

- Apicurio Registry: 3.1.7
- Debezium: 3.0+ 
- Converter: `io.apicurio.registry.utils.converter.AvroConverter`
- API: v3 endpoint

## Steps to Reproduce

1. Configure Debezium connector with Apicurio v3 converter:
```json
{
    "key.converter": "io.apicurio.registry.utils.converter.AvroConverter",
    "key.converter.apicurio.registry.url": "http://registry:8080",
    "key.converter.apicurio.registry.auto-register": "true",
    "value.converter": "io.apicurio.registry.utils.converter.AvroConverter",
    "value.converter.apicurio.registry.url": "http://registry:8080",
    "value.converter.apicurio.registry.auto-register": "true"
}
```

2. Start Debezium snapshot with a table containing many records (e.g., 1000+)

3. Observe registry access logs - repeated POST calls for every record:
```
"POST /apis/registry/v3/groups/.../artifacts?ifExists=FIND_OR_CREATE_VERSION" 200
"POST /apis/registry/v3/groups/.../artifacts?ifExists=FIND_OR_CREATE_VERSION" 200
"POST /apis/registry/v3/groups/.../artifacts?ifExists=FIND_OR_CREATE_VERSION" 200
... (repeated for every record)
```

## Expected Behavior

Schema should be registered once and cached. Subsequent records with the same schema should use the cached result without making API calls.

- **Expected**: 2-5 API calls for initial schema registration
- **Actual**: 1000+ API calls for 1000 records

## Root Cause

In `DefaultSchemaResolver.handleAutoCreateArtifact()`, the content-based cache lookup key includes `references` (from nested schema registrations):

```java
var key = ContentWithReferences.builder()
        .content(rawSchemaString)
        .references(references)  // Included in lookup key
        .build();
```

But the content extractor in `AbstractSchemaResolver` that stores entries does NOT include references:

```java
schemaCache.configureContentKeyExtractor(schema -> {
    return ContentWithReferences.builder()
            .content(IoUtil.toString(rawSchema))
            .build();  // References NOT included
});
```

This mismatch causes the cache lookup to always miss, triggering an API call for every record.

## Workaround

[...]

PR #8762, opened 2026-07-20
fix(examples): migrate debezium-otel-tracing example to Apicurio Registry 3.x API endpoints

Shubh2-0:
## Summary
This PR migrates the \examples/debezium-otel-tracing\ example to be fully compatible with Apicurio Registry 3.x.

## Changes Made
- Updated \debezium.format.value.apicurio.registry.url\ in \debezium-server/application.properties\ to point to \/apis/registry/v3\ instead of legacy \/apis/registry/v2\.
- Updated trace propagation diagram in \examples/debezium-otel-tracing/README.md\ to reflect \/apis/registry/v3\ schema registration endpoint.

Follow-up to #8709.

### DIFF
--- a/examples/debezium-otel-tracing/debezium-server/application.properties
+++ b/examples/debezium-otel-tracing/debezium-server/application.properties
@@ -55,7 +55,7 @@ debezium.transforms.unwrap.add.fields=op
 # Note: Using base URL only - Debezium Server's Avro format internally appends the API path
 debezium.format.key=json
 debezium.format.value=avro
-debezium.format.value.apicurio.registry.url=http://apicurio-registry:8080/apis/registry/v2
+debezium.format.value.apicurio.registry.url=http://apicurio-registry:8080/apis/registry/v3
 debezium.format.value.apicurio.registry.auto-register=true
 # Enable global ID in message body (Confluent-compatible wire format)
 debezium.format.value.apicurio.registry.use-id=globalId
--- a/examples/debezium-otel-tracing/README.md
+++ b/examples/debezium-otel-tracing/README.md
@@ -178,7 +178,7 @@ order-service: POST /orders
 ├── create-order-entity (database insert)
 │
 └── debezium-server: debezium-cdc-orders (custom OTEL SMT)
-    ├── POST apicurio-registry/apis/registry/v2/... (schema registration)
+    ├── POST apicurio-registry/apis/registry/v3/... (schema registration)
     │   └── apicurio-registry: createArtifact
     ├── Kafka producer send
     │


examples/debezium-otel-tracing/debezium-server/application.properties
```diff
@@ -55,7 +55,7 @@ debezium.transforms.unwrap.add.fields=op
 # Note: Using base URL only - Debezium Server's Avro format internally appends the API path
 debezium.format.key=json
 debezium.format.value=avro
-debezium.format.value.apicurio.registry.url=http://apicurio-registry:8080/apis/registry/v2
+debezium.format.value.apicurio.registry.url=http://apicurio-registry:8080/apis/registry/v3
```
paoloantinori, line 58: Correct value. The Apicurio v3 client (selected by `RegistryClientFacadeFactory` for any URL that doesn't contain `/apis/registry/v2`) uses the configured `registryUrl` verbatim and appends the API path to it, so the full `/apis/registry/v3` suffix is needed here. This also brings the Debezium Server half in line with the sibling `cdc-consumer/src/main/resources/application.properties`, which already used `apicurio.registry.url=http://localhost:8086/apis/registry/v3`. Heads-up (pre-existing, not caused by this change): Debezium Server 3.0 against the v3 serdes is affected by the per-record schema-fetch regression in #7395.

paoloantinori (commented):
Two-line change that brings the Debezium Server half of the `debezium-otel-tracing` example in line with the rest of the example and with Apicurio Registry 3.x. The change is correct; one pre-existing caveat worth flagging for awareness.

- **`/apis/registry/v3` is the right value for the converter URL.** I cross-checked against `schema-resolver/.../RegistryClientFacadeFactory.java` on `main`: the client uses the configured `registryUrl` verbatim as the base and appends the API path to it, and the factory picks the v3 client for any URL that does NOT contain `/apis/registry/v2`. The sibling config in this very example, `cdc-consumer/src/main/resources/application.properties`, already uses `apicurio.registry.url=http://localhost:8086/apis/registry/v3` and documents it as "Using v3 API with the v3 SerDe library". So before this PR the Debezium Server was the only component in the example still pinned to v2, and the v3 URL is the documented form (`RegistryClientOptions.registryUrl(...)` example: `http://localhost:8080/apis/registry/v3`).

- **README diagram is now internally consistent.** `cdc-consumer` registers/consumes via v3, the Debezium Server now registers via v3, the diagrams match the `application.properties` values.

- **Pre-existing caveat (NOT introduced by this PR, no action needed here):** the Debezium Server 3.0 image (`quay.io/debezium/server:3.0`, pinned in `debezium-server/Dockerfile`) running against the v3 serdes hits the schema-caching regression tracked in #7395 — the v3 client makes a registry API call per record instead of caching, which can degrade performance during the initial snapshot. Worth a heads-up to anyone running this example against a large source DB, but it is a known upstream issue and orthogonal to this 2-line URL fix. If the maintainers want, a one-line note in the README pointing at #7395 would be a reasonable follow-up.

EricWittmann:
Thanks @Shubh2-0 :+1:

PR #8764, opened 2026-07-20
feat(cli): resolve dynamic rule-type and rule-config placeholders in help text

### LINKED ISSUE #8028: CLI: Dynamic help text via picocli command tree walking
## Description

`AbstractCommand.applyBranding()` walks the picocli `CommandSpec` tree at runtime and replaces `{{product-name}}` placeholders. This approach could be extended to resolve other dynamic placeholders in help text:

- `{{rule-types}}` — dynamically list valid rule types (VALIDITY, COMPATIBILITY, INTEGRITY)
- `{{rule-configs}}` — list valid rule configurations per type

This would eliminate hardcoded enum listings in help text and keep them in sync with the actual supported values.

## Parent Epic

- #7162

Priyanshubhartistm:
## Summary

The rule commands (`acr rule`, `acr artifact rule`, `acr group rule`) hard-coded the list of valid rule types and per-type configuration values in their `@Option` and `@Parameters` help text. These duplicated the enums in `RuleUtil` and were flagged with
`// NOTE: Keep in sync with RuleUtil.VALID_CONFIGS` comments, making them easy to drift out of sync.

This change extends the existing `BrandedCommandLineProducer` placeholder mechanism (which already resolves `{{product-name}}`) to also resolve two new placeholders that are filled in at runtime from the actual enums, so the help text always matches the supported values.

Fixes #8028.

## Changes

- `BrandedCommandLineProducer`: generalize the command-tree walk to resolve placeholders in option and positional-parameter descriptions in addition to command descriptions, and add the `{{rule-types}}` and `{{rule-configs}}` placeholders alongside `{{product-name}}`.
- `RuleUtil`: add `renderRuleTypes()` and `renderRuleConfigs()`, which build the help text from the existing `RuleType`, `ValidityLevel`, `CompatibilityLevel`, and `IntegrityLevel`  enums.
- Replace the hard-coded rule-type list and configuration block in all rule commands  (create/update/get/delete for global, artifact, and group rules) with the  `{{rule-types}}` and `{{rule-configs}}` placeholders, and remove the now-redundant  "keep in sync" comments.

## Testing

- `RuleUtilTest` (new): unit-tests the rendered rule-type and rule-config strings.
- `DynamicHelpTest` (new): asserts that the CDI-produced `CommandLine` resolves the placeholders in the rendered help of `acr rule create` and `acr artifact rule create`  (nested subcommands), and that no literal `{{...}}` placeholder leaks into the output.
- Existing rule command tests continue to pass, confirming option and positional binding is  preserved after the descriptions are rebuilt.

## Notes

- No new short options or command-line arguments; only help text changes.
- The configuration values are now listed in enum declaration order and always reflect the complete set defined by the enums.


cli/src/main/java/io/apicurio/registry/cli/BrandedCommandLineProducer.java
```diff
@@ -36,28 +41,66 @@ private String resolveProductName() {
...
+            exitCodes.forEach((k, v) -> patched.put(k, resolve(v, productName)));
             usage.exitCodeList(patched);
         }
+        // Options and positional parameters are immutable once built, so rebuild any whose
+        // description changed and swap it back into the command spec.
+        for (var option : List.copyOf(spec.options())) {
+            var resolved = resolve(option.description(), productName);
+            if (!Arrays.equals(resolved, option.description())) {
+                spec.remove(option);
+                spec.add(option.toBuilder().description(resolved).build());
```
paoloantinori: Worth double-checking: `spec.remove(option)` followed by `spec.add(option.toBuilder()...build())` appends the rebuilt option at the end of `CommandSpec.options()`, so any option whose description contains a placeholder loses its original position in the help output. For this PR only `-c, --config` is affected per command, so the visual drift is small, but if more placeholders are added later the reordering will compound. If preserving declaration order matters, consider rebuilding the option in place (e.g. collecting the rebuilt specs and replacing the whole list once) rather than remove+add.

paoloantinori (commented):
Clean DRY refactor — extracting `{{rule-types}}` / `{{rule-configs}}` placeholders driven from the existing `RuleType` / `ValidityLevel` / `CompatibilityLevel` / `IntegrityLevel` enums is the right call, and the two new test classes give good coverage of both the rendering and the end-to-end placeholder resolution at different nesting depths. A couple of observations below.

- **Potential option reordering in `applyPlaceholders`** (see inline comment). The `spec.remove(option)` + `spec.add(rebuilt)` pattern moves any rebuilt `ArgSpec` to the end of `options()`. In practice only `-c, --config` and the `ruleType` positional carry placeholders in this PR, so the visible impact is limited, but it is worth a sanity check of the rendered `--help` for e.g. `acr artifact rule create` to confirm `--config` does not jump below the other options.
- `renderRuleConfigs()` filters `VALID_RULE_TYPES` by `VALID_CONFIGS::containsKey`. Today every valid type has a config entry, so this is fine; if a future rule type is added to `VALID_RULE_TYPES` without a corresponding `VALID_CONFIGS` entry, `{{rule-types}}` would advertise it while `{{rule-configs}}` would silently omit it. A short note in `RuleUtil` documenting that invariant would help future contributors keep the two maps in sync.
- Minor: `DynamicHelpTest` covers `rule create` at depths 2 and 3, which exercises the recursive walk well. The `update` / `delete` / `get` variants share the same code path so the coverage is reasonable; no change requested.

paoloantinori (commented):
Seen the new push — CI is green and SonarQube passed. From my side the latest head looks good.

Priyanshubhartistm:
@jsenko, could you please take a look and let me know if anything needs to be updated?
 Thanks

Priyanshubhartistm:
@jsenko thank you

jsenko (approved):

PR #8766, opened 2026-07-21
fix(ui): sync test panel state when variables load asynchronously

### LINKED ISSUE #8765: Boolean prompt variables dropped from render request when checkbox is untouched
## Description

In the Documentation tab's Test Prompt panel, boolean variables are silently dropped from the render API request when the user doesn't toggle the checkbox.

The checkbox shows as unchecked (implying `false`), but the variable isn't sent at all — the server never sees it.

This happens because `PromptTemplateTestPanel` initializes form state with `useState(initialValues)`, but for PROMPT_TEMPLATE artifacts, `DocumentationTabContent` starts with empty content (`{}`) and fetches the real content async via `getArtifactVersionContentDereferenced`. By the time variables arrive, `useState` has already locked its initial value as `{}`. Field labels render fine on re-render, but the state never picks up the new keys. Only fields the user interacts with get added via `setValue`.

### Environment

### Steps to Reproduce

1. Create a PROMPT_TEMPLATE artifact with a boolean variable (e.g. `includeExamples`) and a string variable (e.g. `topic`)
3. Type something in the string field
4. Leave the boolean checkbox untouched
5. Click Render
6. Check the network request payload to the `/render` endpoint

**Expected:** payload includes `includeExamples: false`
**Actual:** payload is `{variables: {topic: ...}}` — `includeExamples` is missing entirely

### Logs

No errors in console — the variable is silently omitted from the request.

Shubb07:
Fixes #8765

For PROMPT_TEMPLATE artifacts, `DocumentationTabContent` starts with empty content and fetches the real content async. By the time variables arrive, `useState` in `PromptTemplateTestPanel` has already locked its initial value as `{}`. Field labels render fine, but the form state never picks up the new keys — so untouched fields are missing from the render request.

This is most visible with boolean variables: the checkbox shows unchecked (implying `false`), but the value is silently omitted from the payload instead of being sent as `false`.

Added a `useEffect` that watches `props.variables` and seeds any missing keys into form state. Booleans default to `false`, other types use their declared default or empty string. Already-interacted fields are left untouched.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, useState } from "react";
+import { FunctionComponent, useEffect, useState } from "react";
 import "./PromptTemplateTestPanel.css";
 import {
     ActionGroup,
@@ -52,6 +52,20 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
     });
 
     const [values, setValues] = useState<Record<string, any>>(initialValues);
+
+    useEffect(() => {
+        setValues(prev => {
+            const next = { ...prev };
+            variablesList.forEach(({ name, variable }) => {
+                if (!(name in next)) {
+                    const type = (variable.type || "string").toLowerCase();
+                    next[name] = type === "boolean" ? false : (variable.default ?? "");
+                }
+            });
+            return next;
+        });
+    }, [props.variables]);
+
     const [renderedOutput, setRenderedOutput] = useState<string>("");
     const [validationErrors, setValidationErrors] = useState<RenderPromptValidationError[]>([]);
     const [isLoading, setIsLoading] = useState(false);


paoloantinori (commented):
Clean, focused fix for a real async race condition in `PromptTemplateTestPanel`. The `useEffect` seeds any variables that arrived after initial render into the form state without clobbering already-interacted fields — good defensive design.

One minor consideration worth double-checking:

- The `useEffect` dependency is `[props.variables]`. If the parent re-renders with a **new array reference** but identical variable contents (e.g. after a re-fetch that returns the same data), `useEffect` will re-run `setValues` — harmless because the `!(name in next)` guard prevents overwriting, but it does create a new state object on every such render. Not a bug, just worth noting for performance if variable lists are large.

Shubb07:
good point — in practice prompt templates have a handful of variables so the extra set shouldn't matter, but happy to add a shallow comparison check if you'd prefer.

paoloantinori:
Agreed — for the scale of prompt templates (a handful of variables per template) the extra render is negligible and not worth the added complexity. Leaving it as-is is the right call. Thanks for the fix.

EricWittmann:
Thanks @Shubb07 :+1:

Shubb07:
Thanks @EricWittmann! happy this one's merged. planning to apply for the LFX Term 3 playground mentorship (#1916) — will keep sending small UI fixes in the meantime.

PR #8767, opened 2026-07-21
refactor(cli): simplify Config with property/context helpers and deferred flush

### LINKED ISSUE #8030: CLI: Config class refactor
## Description

The current `Config.read()`/`Config.write()` pattern requires callers to read the model, modify the map, and write it back — repeated in many places. Proposed improvements:

- Add convenience methods for reading/writing individual config properties (e.g. `Config.getProperty(key)`, `Config.setProperty(key, value)`)
- Remove defensive copy in `read()` — cache the deserialized config and add a `flush()` method to persist, called once in `AbstractCommand.call()` finally block
- Unify `getAcrCurrentHomePath()` to use `envOverrides` instead of a separate `acrCurrentHomePath` setter
- Add a config property constants class (e.g. `ConfigProperties`) with static final fields for all property keys to eliminate string literals scattered across commands and services

## Parent Epic

- #7162

Anshumancanrock:
Closes #8030

## Description

The old `Config.read()`/`write()` pattern meant every command had to pull the
whole model, mutate the map by hand, and write it back. Easy to forget, easy
to get wrong.

## What changed

- Added direct accessors instead of read-mutate-write:
  - `getProperty` / `setProperty` / `removeProperty` / `hasProperty` for config values
  - `getContext` / `putContext` / `removeContext` / `clearContexts` / `setCurrentContext` / `updateContext` for contexts
- `read()` now returns the live in-memory cache instead of a defensive copy
- Config is flushed to disk **once**, automatically, at the end of each command so commands no longer call `write()` themselves
- `getAcrCurrentHomePath()` resolves through the same env-override path as everything else instead of keeping its own setter
- Added `ConfigProperties`, a constants class replacing scattered string literals for config keys

`config set` / `delete` still validate every argument before applying any
change, so a bad key in a batch doesn't leave things half-updated.

## Testing

- Added `ConfigTest.java` covering the caching, dirty-tracking, and flush behavior directly
- Existing CLI test suite updated to the new API (`AbstractCLITest`, `ConfigPropertyCommandTest`, `InstallCommandTest`)

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -41,33 +41,43 @@ public abstract class AbstractCommand implements Callable<Integer> {
     @Override
     public Integer call() {
         var output = new OutputBuffer(config.getStdOut(), config.getStdErr());
+        var commandFailed = false;
         try {
             configureVerboseLogging();
             updateNotifier.checkAndNotify(getTopLevelCommandName());
             run(output);
             return OK_RETURN_CODE;
         } catch (CliException ex) {
+            commandFailed = true;
             handleCliException(output, ex);
             return ex.getCode();
         } catch (RuleViolationProblemDetails ex) {
+            commandFailed = true;
             handleRuleViolation(output, ex);
             return SERVER_ERROR_RETURN_CODE;
         } catch (ProblemDetails ex) {
+            commandFailed = true;
             handleProblemDetails(output, ex);
             return SERVER_ERROR_RETURN_CODE;
         } catch (Exception ex) {
+            commandFailed = true;
             log.error("Unexpected error", ex);
             output.writeStdErrChunk(out -> out.append("Unexpected error: ").append(ex.getMessage()).append("\n"));
             return APPLICATION_ERROR_RETURN_CODE;
         } finally {
-            persistConfig(output);
+            persistConfig(output, commandFailed);
             output.print();
         }
     }
 
-    // Runs in the finally block so config updates made before a failure are still saved.
-    private void persistConfig(final OutputBuffer output) {
+    // Runs in finally so config changes made before a failure are still saved. Commands update
+    // config only after their network/IO work succeeds, so a failure won't leave a partial config.
+    private void persistConfig(final OutputBuffer output, final boolean commandFailed) {
         try {
+            if (commandFailed && config.isDirty()) {
+                log.debugf("Command '%s' failed but had unsaved config changes; saving them anyway.",
+                        getTopLevelCommandName());
+            }
             config.flush();
         } catch (Exception ex) {
             log.error("Could not persist configuration changes", ex);
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -70,8 +70,9 @@ public abstract class AbstractCommand implements Callable<Integer> {
         }
     }
 
-    // Runs in finally so config changes made before a failure are still saved. Commands update
-    // config only after their network/IO work succeeds, so a failure won't leave a partial config.
+    // flush() runs regardless of command outcome, so config changes are saved even when the command
+    // fails. Commands are expected to change config only after their IO succeeds (a convention this
+    // class can't enforce), so a failure won't persist half-finished state.
     private void persistConfig(final OutputBuffer output, final boolean commandFailed) {
         try {
             if (commandFailed && config.isDirty()) {
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -60,7 +60,19 @@ public abstract class AbstractCommand implements Callable<Integer> {
             output.writeStdErrChunk(out -> out.append("Unexpected error: ").append(ex.getMessage()).append("\n"));
             return APPLICATION_ERROR_RETURN_CODE;
         } finally {
+            persistConfig(output);
             output.print();
+        }
+    }
+
+    // Runs in the finally block so config updates made before a failure are still saved.
+    private void persistConfig(final OutputBuffer output) {
+        try {
+            config.flush();
+        } catch (Exception ex) {
+            log.error("Could not persist configuration changes", ex);
+            output.writeStdErrChunk(out -> out.append(ERROR_PREFIX)
+                    .append("Could not persist configuration changes: ").append(ex.getMessage()).append("\n"));
         }
     }
 
--- a/cli/src/main/java/io/apicurio/registry/cli/config/Config.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/config/Config.java
@@ -30,6 +30,7 @@ public class Config {
     // Loaded via SPI by AcrHomeConfigSource before CDI is available; commands should @Inject Config instead.
     static Config instance;
 
+    // Not synchronized; the CLI runs one command at a time on a single thread.
     private ConfigModel cachedConfig;
     private boolean dirty;
 
@@ -116,8 +117,25 @@ public class Config {
         return getAcrCurrentHomePath().resolve("config.json");
     }
 
-    // Returns the live cache, not a defensive copy. Mutating it directly requires calling markDirty().
+    /**
+     * Returns the live config object. Prefer the typed accessors ({@link #getProperty},
+     * {@link #setProperty}, {@link #putContext}, {@link #updateContext}, ...), which mark the config
+     * dirty so {@link #flush()} saves it.
+     *
+     * @deprecated Changes made directly on the returned object are not saved unless you also call
+     *     {@link #markDirty()}, which is easy to forget:
+     *     <pre>{@code
+     *       config.read().setCurrentContext("dev");   // not saved
+     *       config.setCurrentContext("dev");          // saved
+     *     }</pre>
+     */
+    @Deprecated
     public ConfigModel read() {
+        return cache();
+    }
+
+    // Loads and caches config.json on first use.
+    private ConfigModel cache() {
         if (cachedConfig == null) {
             var configPath = getConfigFilePath();
             try {
@@ -130,56 +148,56 @@ public class Config {
     }
 
     public String getProperty(final String key) {
-        return read().getConfig().get(key);
+        return cache().getConfig().get(key);
     }
 
     public boolean hasProperty(final String key) {
-        return read().getConfig().containsKey(key);
+        return cache().getConfig().containsKey(key);
     }
 
     public void setProperty(final String key, final String value) {
-        read().getConfig().put(key, value);
+        cache().getConfig().put(key, value);
         markDirty();
     }
 
     public String removeProperty(final String key) {
-        var previous = read().getConfig().remove(key);
+        var previous = cache().getConfig().remove(key);
         markDirty();
         return previous;
     }
 
     public String getCurrentContext() {
-        return read().getCurrentContext();
+        return cache().getCurrentContext();
     }
 
     public ConfigModel.Context getContext(final String name) {
-        return read().getContext().get(name);
+        return cache().getContext().get(name);
     }
 
     public void setCurrentContext(final String name) {
-        read().setCurrentContext(name);
+        cache().setCurrentContext(name);
         markDirty();
     }
 
     public void putContext(final String name, final ConfigModel.Context context) {
-        read().getContext().put(name, context);
+        cache().getContext().put(name, context);
         markDirty();
     }
 
     public ConfigModel.Context removeContext(final String name) {
-        var removed = read().getContext().remove(name);
+        var removed = cache().getContext().remove(name);
         markDirty();
         return removed;
     }
 
     public void clearContexts() {
-        read().getContext().clear();
+        cache().getContext().clear();
         markDirty();
     }
 
     // No-op (returns false) if no context with the given name exists.
     public boolean updateContext(final String name, final Consumer<ConfigModel.Context> mutation) {
-        var context = read().getContext().get(name);
+        var context = cache().getContext().get(name);
         if (context == null) {
             return false;
         }
@@ -190,6 +208,10 @@ public class Config {
 
     public void markDirty() {
         dirty = true;
+    }
+
+    public boolean isDirty() {
+        return dirty;
     }
 
     // Called once at the end of each command, from AbstractCommand.call().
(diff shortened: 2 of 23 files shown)


Anshumancanrock:
hi @vandanayadav7 , could you please review this PR? Thanks !

cli/src/main/java/io/apicurio/registry/cli/config/Config.java
```diff
@@ -132,14 +126,81 @@ public ConfigModel read() {
                 throw new CliException("Could not read config file '%s'.".formatted(configPath), ex, APPLICATION_ERROR_RETURN_CODE);
             }
         }
-        return copy(cachedConfig);
+        return cachedConfig;
```
paoloantinori, line 147: This change from defensive copy to live cache is the core design decision of the PR. It enables the accessor methods to work without an explicit `write()` call, but it also means any code that calls `read()` and mutates the returned object _without_ going through the new helpers will silently lose changes (no `markDirty()` → no flush). Consider adding a `@Deprecated` annotation or a code comment with a concrete example of the pitfall, so future contributors don't fall into it.

cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
```diff
@@ -60,10 +60,22 @@ public Integer call() {
             output.writeStdErrChunk(out -> out.append("Unexpected error: ").append(ex.getMessage()).append("\n"));
             return APPLICATION_ERROR_RETURN_CODE;
         } finally {
+            persistConfig(output);
             output.print();
         }
     }
 
+    // Runs in the finally block so config updates made before a failure are still saved.
```
paoloantinori: This `finally`-block flush is the other key behavioral change. Commands that mutate config and then fail will still have those mutations persisted. This is noted in the PR description as intentional, but it might be worth logging (at DEBUG level) when flush happens on an exception path so operators can trace unexpected state if a command fails. The existing `log.error` here is good for the flush-failure case.

cli/src/main/java/io/apicurio/registry/cli/services/UpdateNotifier.java
```diff
@@ -52,15 +53,14 @@ public void checkAndNotify(String commandName) {
 
     private boolean shouldCheck() {
         try {
-            var configModel = config.read();
-            var props = configModel.getConfig();
+            var props = config.read().getConfig();
```
paoloantinori: Minor inconsistency: this method still uses `config.read().getConfig().get(...)` instead of `config.getProperty(...)`. The rest of the PR consistently migrated to the new accessors — this was likely just missed.

paoloantinori (commented):
This is a well-motivated refactor — replacing the manual read-mutate-write cycle with typed accessors and automatic deferred flush eliminates a whole class of bugs (forgotten `write()` calls, half-updated state). The `ConfigProperties` constants class is a good cleanup. Tests in `ConfigTest.java` cover the core caching/dirty-tracking/flush mechanics thoroughly.

A few observations worth considering:

- **`read()` returning the live cache is a footgun.** The Javadoc comment warns about it, but future contributors may still mutate the returned `ConfigModel` directly without calling `markDirty()` and lose their changes. Worth considering whether `read()` should stay defensive-copy (the old behavior) for safety, or at minimum making the contract more visible (e.g. returning an interface that doesn't expose setters, with a separate `readMutable()` or similar). Not a blocker, but the current design trades safety for convenience.

- **Exception-path flush in `AbstractCommand.call()`.** The `finally` block calls `config.flush()`, meaning if a command partially mutates config then throws (e.g. bad network response halfway through), those partial mutations get persisted to disk. The PR description says this is intentional, but it could surprise users. Worth double-checking whether any commands do partial config mutation before a point-of-no-return — if so, the dirty flag should only be set after all mutations succeed.

- **Thread safety note.** `Config` is `@ApplicationScoped` but `cachedConfig` and `dirty` are plain fields without `volatile` or synchronization. This is fine for the current CLI usage (single-threaded command dispatch), but if `AcrHomeConfigSource` (SPI-loaded) ever gets called from a different thread, the lack of visibility guarantees could cause stale reads. Low risk today, worth a brief comment in the code for future readers.

- **`UpdateNotifier.shouldCheck()` (line ~55 in diff)** still uses the old `config.read().getConfig().get(...)` pattern instead of `config.getProperty(...)`. Minor inconsistency with the rest of the refactor.

- **`ConfigPropertyDeleteCommand` atomicity.** The pre-validation loop (checking all keys exist before removing any) is good. However, `keys.forEach(config::removeProperty)` calls `markDirty()` once per key — fine functionally, but `flush()` will serialize the whole model anyway, so this is just a minor inefficiency, not a bug.

cli/src/main/java/io/apicurio/registry/cli/config/Config.java
Anshumancanrock: Make sense. I kept the live cache since that's what makes the deferred-flush work, but added `@Deprecated` plus a concrete WRONG/RIGHT example in the Javadoc so the trap is obvious on hover. To avoid the recommended accessors flagging themselves, the internal ones now go through a private cache() helper; only direct read() calls get the deprecation nudge.

cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
Anshumancanrock: Done. `call()` now tracks whether the command failed, and `persistConfig` logs at DEBUG when it flushes dirty config on the exception path so operators can trace unexpected state after a failure. Gated on `isDirty()` so a clean no-op flush stays quiet, and left the existing log.error for the flush-failure case as-is.

paoloantinori (commented):
Solid follow-up on the earlier round — the typed accessors, batch-then-apply validation in `config set`/`delete`, and the deferred-flush hook in `AbstractCommand` are a real improvement over the read-mutate-write pattern. A few observations worth flagging before this lands:

- **`read()` now returns the live cache.** This is an intentional and documented change (and `ConfigTest.testReadReturnsLiveCacheNotDefensiveCopy` pins it), but it inverts the previous contract where `read()` handed back a defensive `copy`. The PR correctly adds `new HashMap<>(...)` / `new HashSet<>(...)` guards in `AcrHomeConfigSource`, but the contract change is broader: any caller that previously did `var m = config.read(); m.getConfig().put(...)` to *inspect* without persisting now silently mutates shared state. Worth a grep over `config.read()` callers outside this diff to confirm none relied on the old copy semantics — `ContextDeleteCommand` uses it read-only, which is fine, but a quick audit would close the gap.

- **Direct model mutation bypasses the dirty flag.** `ConfigTest.testMarkDirtyCausesStructuralChangesToPersist` exposes this: mutating `config.read().setCurrentContext(...)` on the live model does NOT set `dirty`, so the caller must remember to call `config.markDirty()` or the change is lost on the next non-dirty flush. The new typed accessors (`setProperty`, `updateContext`, ...) handle this correctly, but the escape hatch via `read()` is a footgun. Two options worth considering: either document prominently that `read()` is for inspection only and mutations must go through the accessors, or have `read()` return a read-only wrapper. As-is, a future contributor doing `config.read().setCurrentContext(x)` will silently lose the change.

- **`AbstractCommand.persistConfig` saves on failure too.** The comment says commands update config only after their IO succeeds, so a failure won't leave a partial state — but that's a convention the framework can't enforce. If a command calls `config.setProperty(...)` and *then* hits a network/IO failure, the property is persisted anyway because `flush()` runs unconditionally in `finally`. For auth-related commands (`LoginCommand`, `LogoutCommand`) the ordering happens to be safe (credentials stored before context mutated), but the invariant is load-bearing. A one-line note in `AbstractCommand` or `Config` that 'dirty state is flushed regardless of command outcome' would make the contract explicit.

- Minor: `ContextUpdateCommand` now checks `config.getContext(contextName) == null` to detect absence, then calls `config.updateContext(...)`. Between the check and the update the cache is stable (single-threaded CLI), so this is fine — just noting the check-then-act is only safe under the single-command-at-a-time assumption that `Config` already documents.

Tests look thorough (`ConfigTest` covers dirty tracking, flush no-op, live-cache identity). No blocking concerns from my side; the dirty-flag escape hatch is the one I'd most like to see addressed or explicitly documented.

cli/src/main/java/io/apicurio/registry/cli/config/Config.java
```diff
@@ -107,23 +106,36 @@ public Path getAcrHomePath() {
...
     public Path getAcrCurrentHomePath() {
-        if (acrCurrentHomePath != null) {
-            return acrCurrentHomePath;
-        } else {
-            var acrCurrentHome = System.getenv("ACR_CURRENT_HOME");
-            if (isBlank(acrCurrentHome)) {
-                throw new CliException("ACR_CURRENT_HOME environment variable is not set.", APPLICATION_ERROR_RETURN_CODE);
-            }
-            acrCurrentHomePath = Path.of(acrCurrentHome).normalize().toAbsolutePath();
+        var acrCurrentHome = getEnv(ENV_ACR_CURRENT_HOME);
```
paoloantinori, line 109: This is the load-bearing contract change: `read()` used to return a defensive `copy` (the removed `import static ...Mapper.copy`), now it returns the live `cachedConfig`. The PR adds defensive copies in `AcrHomeConfigSource`, but every other `config.read()` caller now mutates shared state. Worth a quick audit of remaining `read()` call sites (e.g. `ContextDeleteCommand`, `ConfigPropertyCommand`) to confirm none depended on the old isolation.

cli/src/test/java/io/apicurio/registry/cli/ConfigTest.java
```diff
@@ -0,0 +1,186 @@
...
+
+        config.flush();
+        assertThat(Files.exists(configFile)).isFalse();
+    }
+
+    @Test
+    public void testFlushClearsDirtyFlag() throws IOException {
+        config.setProperty("c", "3");
+        config.flush();
+        assertThat(Files.readString(configFile)).contains("\"c\"");
```
paoloantinori, line 124: This test documents a sharp edge: mutating the model returned by `read()` does NOT set the dirty flag, so the caller must call `markDirty()` explicitly or the change is silently dropped on the next flush. The typed accessors (`setProperty`, `updateContext`) handle this correctly — consider either (a) making `read()` return a read-only view to make the footgun impossible, or (b) adding a prominent javadoc on `read()` that mutations must go through the accessors.

cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
```diff
@@ -41,29 +41,51 @@ public abstract class AbstractCommand implements Callable<Integer> {
...
+            commandFailed = true;
             handleRuleViolation(output, ex);
             return SERVER_ERROR_RETURN_CODE;
         } catch (ProblemDetails ex) {
+            commandFailed = true;
             handleProblemDetails(output, ex);
             return SERVER_ERROR_RETURN_CODE;
         } catch (Exception ex) {
+            commandFailed = true;
             log.error("Unexpected error", ex);
```
paoloantinori, line 64: `flush()` runs unconditionally in `finally`, so a command that sets a property and *then* fails on network IO will still persist the property. The comment says commands update config only after IO succeeds — that's a convention, not something the framework enforces. Worth a one-line note that dirty state is flushed regardless of outcome, so future command authors know not to mutate config before a fallible operation.

cli/src/main/java/io/apicurio/registry/cli/config/Config.java
Anshumancanrock: Audited every remaining config.read() call site. All production callers are read-only: ContextCommand, ContextDeleteCommand (.size()), ConfigPropertyCommand (.entrySet()), ExportCommand, Client, IdUtil, and AcrHomeConfigSource (already copies into a fresh HashMap/HashSet). None mutate the returned model, so none depended on the old defensive-copy isolation. The only mutation-through-read() is the test that deliberately documents the markDirty() escape hatch.

cli/src/test/java/io/apicurio/registry/cli/ConfigTest.java
Anshumancanrock: Went with option (b): read() is now @Deprecated with a prominent javadoc stating mutations must go through the typed accessors, plus a concrete example showing config.read().setCurrentContext(x) is silently dropped vs. config.setCurrentContext(x). 

Kept the live cache since that's what enables the deferred flush, glad to switch to a read-only view (option a) if you'd rather the footgun be impossible than documented.

cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
Anshumancanrock: Added the note,  the comment now states `flush()` runs regardless of command outcome (config is saved even on failure) and calls out that the "only mutate config after IO succeeds" ordering is a convention the framework can't enforce, so command authors know not to mutate config before a fallible operation.

Anshumancanrock:
@paoloantinori Thanks for the detailed review. I went through each point:

1. `read()` returning the live cache. Kept the live cache since that's what makes the deferred flush work, but marked read() @Deprecated with a javadoc example showing that editing the returned model directly is lost unless you go through the accessors. I also audited every `read()` caller: they're all read-only (ContextDeleteCommand's .size(), ConfigPropertyCommand's .entrySet(), ExportCommand/Client/IdUtil, and AcrHomeConfigSource already copies into a fresh map/set), so nothing relied on the old defensive-copy isolation.

2. Exception-path flush. Checked every command that mutates config, they all write config after their network/IO work, so a failure can't leave a half-applied model. I made that explicit in AbstractCommand: the comment now notes flush() runs regardless of command outcome and that the ordering is a convention the framework can't enforce. Also added a DEBUG log when config is flushed after a failed command, so it's traceable.

3. Thread safety. Added a note on the cache fields explaining they're intentionally unsynchronized because the CLI runs one command at a time.

4. UpdateNotifier.shouldCheck(). Good catch, that one was just missed, switched it to config.getProperty(...), consistent with the rest now.

5. ConfigPropertyDeleteCommand. Left as-is: markDirty() is just a boolean set, so calling it once per key costs nothing.

Please take a look now !

paoloantinori:
Thanks for the thorough walkthrough — the reasoning holds together well.

On `read()` returning the live cache: marking it `@Deprecated` with a javadoc that spells out "mutate via accessors, not through the returned model" plus the caller audit is the right combination. I cross-checked the call sites in the diff and the read-only usages (`ContextDeleteCommand` size check, `ConfigPropertyCommand` entrySet for listing, the `AcrHomeConfigSource` reads) all line up. The defensive copies you added in `AcrHomeConfigSource` (`new HashMap<>(...)` / `new HashSet<>(...)`) close the one aliasing risk I was most concerned about, since that path hands the map to MicroProfile Config.

The deferred-flush hook in `AbstractCommand` (flush in `finally`, with the explicit log line when a command failed dirty) reads cleanly to me — it's the right place to centralize persistence, and the `isDirty`/`markDirty` split keeps the structural-mutation cases (`setCurrentContext` via the live model) covered.

Nothing else blocking from my side. The test coverage in `ConfigTest` (dirty tracking, flush-is-no-op-when-clean, structural-change persistence) exercises the behaviors that matter.

carlesarnal:
This has been stale for a while. Closing.

PR #8770, opened 2026-07-21
fix(ui): add aria-label to SortOrderToggle button for screen reader a…

Myself-Praveen:
I noticed that the SortOrderToggle icon button was missing an aria-label, making it inaccessible to screen readers. I've added an appropriate aria-label to ensure it is properly described to visually impaired users. This is my first PR as part of applying for the LFX mentorship.

### DIFF
--- a/ui/ui-app/src/app/components/common/SortOrderToggle.tsx
+++ b/ui/ui-app/src/app/components/common/SortOrderToggle.tsx
@@ -18,7 +18,7 @@ export const SortOrderToggle: FunctionComponent<SortOrderToggleProps> = (props: 
     );
 
     return (
-        <Button icon={icon} variant="plain" aria-label="Toggle sort order" onClick={() => {
+        <Button icon={icon} variant="plain" aria-label={props.sortOrder === SortOrder.asc ? "sort ascending" : "sort descending"} onClick={() => {
             props.onChange(props.sortOrder === SortOrder.asc ? SortOrder.desc : SortOrder.asc);
         }} />
     );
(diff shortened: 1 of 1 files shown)


ui/ui-app/src/app/components/common/SortOrderToggle.tsx
```diff
@@ -18,7 +18,7 @@ export const SortOrderToggle: FunctionComponent<SortOrderToggleProps> = (props:
     );
 
     return (
-        <Button icon={icon} variant="plain"  onClick={() => {
+        <Button icon={icon} variant="plain" aria-label="Toggle sort order" onClick={() => {
```
paoloantinori: Since `props.sortOrder` is available here, consider a state-aware label (e.g. `aria-label={props.sortOrder === SortOrder.asc ? "sort ascending" : "sort descending"}`) so screen readers announce the current direction without requiring an activation. This also keeps the component consistent with the sort button labeling added in #8542.

paoloantinori (commented):
Thanks for the accessibility improvement — adding an `aria-label` to an icon-only button is the right call. One suggestion to make it more useful to screen-reader users:

- The label is currently static (`"Toggle sort order"`), but the component already receives `props.sortOrder`, so it could describe the *current* state — e.g. `aria-label={props.sortOrder === SortOrder.asc ? "sort ascending" : "sort descending"}` (or `"Sort: ascending, click to change"`). A static label forces a screen-reader user to activate the control before learning which direction is active, whereas a state-aware label announces it up front. Worth noting this overlaps with #8542, which makes the equivalent label state-aware on the Explore toolbar's sort button — it'd be good to keep the two consistent.

Myself-Praveen:
Thanks for the feedback @paoloantinori! I've updated the aria-label to be dynamic and state-aware so that it describes the current sort direction. The commit has been pushed

paoloantinori:
Thanks @Myself-Praveen — the dynamic, state-aware `aria-label` looks good now. Announcing the current direction ("sort ascending" / "sort descending") is exactly the right behavior for screen-reader users, and it tracks the toggle state cleanly. This is in good shape from my side.

PR #8771, opened 2026-07-21
fix(protobuf): proto2 ProtobufSerializer validation for schema registered as .proto text

### LINKED ISSUE #8279: Protobuf serializer validation fails on proto2 schemas (maps expanded, type names fully-qualified)

kartikangiras:
## Description
Collapse descriptor map fields back to map by detecting synthetic mapentry nested messages and replacing repeated EntryType fields with map fields having no label.

Support explicit type/label passthrough in toField() by adding an overloaded variant so collapsed map fields can supply their own type and null label, bypassing proto2 label computation.

Fix imported type resolution in normalizeType() by removing the fallback that incorrectly prefixed unknown type names with the current message scope instead resolving against the file's package per protobuf scoping rules.

Add a test covering the exact reported scenario

### Issue
fixes https://github.com/Apicurio/apicurio-registry/issues/8279


utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtils.java
```diff
@@ -1427,8 +1466,54 @@ private static FieldElement toField(FileDescriptorProto file, FieldDescriptorPro
...
+            return null;
+        }
+        if (field.hasTypeName()) {
+            return toRelativeTypeName(field.getTypeName(), file.getPackage());
+        }
+        return dataType(field);
+    }
+
+    private static String toRelativeTypeName(String typeName, String packageName) {
+        if (typeName == null) {
```
paoloantinori, line 1522: If `findFieldByName` returns `null` here (malformed descriptor), `getMapEntryFieldType` returns `null`, and the resulting type string becomes `map<null, null>` — invalid proto syntax that fails much later. A well-formed `map_entry` always has both fields, but a defensive `IllegalArgumentException` or a log here would surface the root cause immediately rather than as a downstream parse error.

schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/compatibility/protobuf/ProtobufCompatibilityCheckerLibrary.java
```diff
@@ -360,13 +360,9 @@ private String normalizeType(ProtobufFile file, String type, String messageConte
...
-        // For types we can't locate in the schema, infer the fully qualified name
-        // This handles cross-file references or types that might be defined elsewhere
-        if (messageContext != null && !messageContext.isEmpty()) {
-            return buildFullyQualifiedName(file, messageContext + "." + type);
-        }
-
-        // Default: assume it's a top-level type
+        // For types not found in the current file (e.g., imported types), resolve against the
+        // current package. Protobuf scoping resolves unqualified simple names that are not found
+        // in nested scopes against the file's package, not against the current message context.
```
paoloantinori, line 365: Removing the `messageContext` fallback is correct for imported types, but worth confirming that genuinely nested types (e.g. `Envelope.Type` referenced from a sibling) are still resolved by the earlier lookup — otherwise they'd be built as `package.Type` instead of `package.Envelope.Type`. If the earlier path handles all in-file types, a one-line comment here saying so would prevent future second-guessing.

paoloantinori (commented):
Well-structured fix for the proto2 map-field / imported-type mismatch between text-registered and descriptor-derived schemas. Collapsing the synthetic `map_entry` messages back to `map<K, V>` syntax and resolving imported types against the file package (rather than the message context) are both correct calls, and the new test covers the reported scenario (#8279) end-to-end.

A few observations:

- **Null key/value handling in `getMapEntryFieldType`**: when `findFieldByName` can't locate the `key`/`value` subfield, the method returns `null`, which then gets interpolated into `map<null, null>`. Synthetic `map_entry` messages always carry exactly those two fields per the protobuf spec, so this shouldn't occur with well-formed descriptors — but a malformed one would yield invalid proto text that only surfaces much later as a confusing parse error. A fail-fast `IllegalStateException` (or at least a log) would make the root cause obvious.
- **`normalizeType` scope change**: removing the `messageContext`-prefixed fallback is correct for imported top-level types (the `Detail` case in the test), but worth double-checking that types nested in a parent message scope are still resolved by the earlier lookup path. Otherwise a reference to `Parent.Nested` from within a sibling message could regress to `package.Nested` instead of `package.Parent.Nested`. If the earlier lookup already covers in-file nested types, this is fine — but a brief comment confirming that assumption would help future readers.
- **Test granularity**: the round-trip `differences.isEmpty()` assertion is the key scenario and it's good. It could be strengthened by also asserting on the derived `.proto` text directly (e.g. that it contains `map<string, string>` and does not contain `AttributesEntry`), which would lock in the collapse behavior independently of the compatibility checker's tolerance.

Overall this looks ready to merge once the null edge case gets a guard or an explicit acknowledgment.

schema-util/protobuf/src/main/java/io/apicurio/registry/protobuf/rules/compatibility/protobuf/ProtobufCompatibilityCheckerLibrary.java
kartikangiras: I have updated the comment.

utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtils.java
kartikangiras: Resolved.

utils/protobuf-schema-utilities/src/test/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtilsTest.java
```diff
@@ -172,11 +172,16 @@ public void ParsesFileDescriptorsAndRawSchemaIntoCanonicalizedForm_Accurately(
...
-        // This will be fixed in upcoming square version, https://github.com/square/wire/pull/2046
+        // Use ignoringRepeatedFieldOrder() to accommodate the ordering differences
+        // introduced by
+        // map field handling in fileDescriptorToProtoFile, as well as Square library's
+        // inconsistent ordering of OneOfs and Proto3 optionals.
+        // TODO: Square library doesn't respect the ordering of OneOfs and Proto3
+        // optionals.
+        // This will be fixed in upcoming square version,
+        // https://github.com/square/wire/pull/2046
+        assertThat(actualFileDescriptorProto).ignoringRepeatedFieldOrder()
```
paoloantinori, line 179: Strict `assertEquals` previously held here; switching to `ignoringRepeatedFieldOrder()` means the map-field handling in `fileDescriptorToProtoFile` now produces a different ordering of some repeated field in the round-tripped descriptor (protoc-compiled vs. collapsed-and-recompiled). For a registry where descriptor equality underpins compatibility checks, it'd help to name *which* repeated field is being reordered — synthetic `map_entry` nested types dropping out and shifting later siblings, or the field list itself — and confirm it's benign. If it's only the entry messages disappearing, a one-line note saying so makes the intent explicit; if field ordering changed, that's worth a second look since equality-based comparisons elsewhere could be sensitive to it.

paoloantinori (commented):
Re-read on the current head after the earlier review. The map-entry collapsing in `collectNestedTypes`/`collectFields`, the `toField` overload that lets collapsed map fields bypass proto2 label computation, and the `normalizeType` simplification all read cleanly. The new `testProto2MapFieldAndImportedReferenceCompatibility` captures the exact scenario from #8279 — good coverage.

One thing worth a closer look: the round-trip assertion relaxation in `FileDescriptorUtilsTest` (see inline). The rest of the test-file changes there appear to be comment reflowing only.

Refs: #8279, https://github.com/Apicurio/apicurio-registry/blob/main/utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtils.java

utils/protobuf-schema-utilities/src/test/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtilsTest.java
kartikangiras: Applied and changed the relevant comments.

utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtils.java
```diff
@@ -1427,8 +1498,54 @@ private static FieldElement toField(FileDescriptorProto file, FieldDescriptorPro
...
+        for (FieldDescriptorProto field : descriptor.getFieldList()) {
+            if (field.getName().equals(name)) {
+                return field;
+            }
+        }
+        return null;
+    }
+
+    private static String getMapEntryFieldType(FileDescriptorProto file, FieldDescriptorProto field) {
+        if (field == null) {
```
EricWittmann: This `field == null` guard is unreachable. The only call site (`collectNestedTypes`) already throws `IllegalArgumentException` if `keyField` or `valueField` is null before calling this method.

Returning `null` here would silently produce `map<null, ...>` text that fails cryptically downstream, so the fail-fast behavior in the caller is the right place for the check. This dead branch should be removed — or, if you prefer defense-in-depth, replace it with `Objects.requireNonNull(field, "field must not be null")` so the contract is explicit rather than returning a value that can't be used.

utils/protobuf-schema-utilities/src/test/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtilsTest.java
```diff
@@ -428,70 +468,78 @@ public void testInvalidBase64ThrowsException() {
...
+     * This validates that dependencies in binary format can be used when resolving
+     * references.
      */
     @Test
     public void testParseProtoFileWithBinaryDependencies() throws Exception {
         // Create binary (base64) version of dependency
         DescriptorProtos.FileDescriptorProto depProto = FileDescriptorUtils
-            .protoFileToFileDescriptor(DEP_PROTO_SCHEMA, "dep.proto", Optional.of("test"))
-            .toProto();
+                .protoFileToFileDescriptor(DEP_PROTO_SCHEMA, "dep.proto", Optional.of("test"))
```
EricWittmann: This line (and many others in this file) is a whitespace-only change — re-indenting an existing method chain from 4-space continuation to 8-space. These reformatting changes are spread throughout the test file (~40 of the 88 added lines) and make it harder to spot the substantive changes in the diff.

Per the contributor checklist: "PR contains no unrelated changes (no whitespace fixes, no import reordering in untouched files)." Could you revert the whitespace-only reformatting and keep only the assertion changes (`assertEquals` → `assertThat(...).ignoringRepeatedFieldOrder()`) and the new test method?

EricWittmann (commented):
**DCO sign-off missing on all contributor commits.**

None of the six commits from this PR have a `Signed-off-by:` line:
- `1d9fbf74c` fix(protobuf): improve fully qualified name resolution...
- `94b7b261e` fix(protobuf): enhance FileDescriptorUtils...
- `ff0b985c9` fix(tests): update assertions...
- `01bf956a8` chore
- `bee453a73` fix(tests): enhance comments...
- `f1fc4f74d` Merge branch 'main' into fix/protobuf

Per the contributor checklist, all commits must have DCO sign-off (`Signed-off-by: Name <email>`). The simplest fix would be to squash into a single commit with `--signoff`, or `git rebase --signoff` to amend them all.

utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtils.java
kartikangiras: Resolved.

kartikangiras:
Hi @EricWittmann @paoloantinori ,
I iterated over your feedbacks and resolved all the changes required.
Thanks

paoloantinori:
@kartikangiras Thanks for the iteration. I re-read the current head (`fcd4dfa`) and the changes hold up: the map-entry collapsing in `collectNestedTypes`/`collectFields`, the `toField` overload that lets collapsed map fields pass through a null label, and the `normalizeType` fix that resolves imported types against the file's package rather than the message context all look correct. The new test (`testProto2MapFieldAndImportedReferenceCompatibility`) exercises exactly the reported scenario and the `ignoringRepeatedFieldOrder` adjustment in `FileDescriptorUtilsTest` is a reasonable accommodation. LGTM from my side.

EricWittmann (approved):
LGTM. Well-structured fix for the proto2 map-field / imported-type mismatch. The map-entry collapsing, toField overload, and normalizeType simplification are all correct. Good test coverage of the reported scenario. Clean decomposition of toMessage() improves readability.

EricWittmann:
Thanks @kartikangiras :+1:

krickert:
This broke some serialization paths - issuing a patch

paoloantinori:
Thanks for flagging this @krickert. Since this PR has shipped, could you share a minimal repro (proto + serializer/deserializer path that now breaks) so we can pin it down as a regression test before the patch lands? My main suspects from the merged diff:

- The `normalizeType` fallback removal — if some caller relied on the `messageContext + "." + type` prefix for a type that genuinely lives in a nested scope not covered by the scope-iteration loop, the resolved FQ name changes.
- The `map_entry` collapsing in `collectNestedTypes`/`collectFields` — collapsing a repeated `EntryType` into `map<K,V>` changes the wire-format descriptor shape, which can break serializers that round-trip through text `.proto`.

If the patch is a targeted revert/fix rather than touching those two areas, that would tell us which path actually regressed.

Issue #7642, opened 2026-03-25
CLI: Implement automatic update with OS/architecture detection


jsenko:
## Description

The `UpdateCommand` currently only supports `--path` (manual ZIP file). The automatic update path (downloading from Maven Central) is disabled with a TODO. Now that the CLI is distributed as architecture-specific ZIPs, the automatic update needs to:

1. Detect the current OS and architecture
2. Map to the correct Maven Central classifier (`linux-x86_64`, `osx-aarch64`)
3. Download the correct ZIP
4. Run the install

## Tasks

- Implement OS/architecture detection mapping to ZIP classifiers
- Enable the Maven Central download path in `UpdateCommand`
- Add periodic update check (configurable via `acr.update.check.enabled`)
- Handle version comparison (already partially implemented with semver4j)

## Context

Follow-up from #7618 (Quarkus-based CLI with native build). Related to Phase 4 in the CLI epic (#7162).

PR #8772, opened 2026-07-21
feat(cli): harden the CLI auto-update mechanism

### LINKED ISSUE #8029: CLI: Auto-update hardening
## Description

Follow-ups from #7642 to harden the auto-update mechanism:

- **Download integrity verification** — verify SHA-256 checksums of downloaded ZIPs before extracting/installing
- **HTTP client timeouts** — configure connection/read timeouts and redirect limits on the Vertx HTTP client
- **Rollback mechanism** — keep the old binary until the new one is verified working, so a failed update can recover

## Parent Epic

- #7162

Priyanshubhartistm:
## Summary

Follow-up hardening for the CLI auto-update mechanism added in #7642. The updater now verifies the integrity of downloaded archives, applies sane timeout and redirect limits to the HTTP client used for downloads, and keeps the previous installation until the new one is in place so a failed update can recover.

Fixes #8029.

### Download integrity verification (`Update`)
- After downloading an update archive, download the matching `.sha256` file published  alongside it in the repository and verify the archive's SHA-256 digest before it is written
 to disk. A missing or mismatched checksum aborts the update.

### HTTP client hardening (`Client`, `Update`)
- Configure the Vert.x HTTP client used for updates with a connect timeout, an idle (read)  timeout, and a redirect limit, and follow redirects on the update requests (the download  and checksum endpoints on Maven Central redirect to a CDN).

### Rollback on failed update (`InstallCommand`, `UpdateCommand`)
- Back up the currently installed launcher (`acr`) and binary (`acr_runner`) before replacing them, and restore them if the copy fails partway through, so the user keeps a working installation. Backups are removed on success.
- The "new binary verified working" part of the issue is covered by the existing update flow:  the swap is performed by the **newly downloaded** binary (`update` runs  `<downloaded>/acr install`), so a binary that passes the checksum but cannot run (wrong  architecture, glibc, missing dependency) fails to launch **before** the swap and leaves the
  installation untouched. A code comment in `UpdateCommand.runInstallFromZip` documents this.

## Testing

- `UpdateChecksumTest` (new): unit-tests the SHA-256 hex encoding and the Maven checksum-file parsing.
- `InstallCommandTest`: adds a test that a failed copy rolls the launcher and binary back to the previous version and cleans up the backup files.
- `test-update.sh`: the mock repository now publishes `.sha256` files alongside the archives,  so the end-to-end update test exercises the new integrity check.

## Notes

- Checksum verification only runs for downloaded archives; `acr update --path <zip>` (a local file supplied by the user) is unchanged.
- Maven Central publishes `.sha256` files for the CLI archives, so the default configuration
  is verified end to end.


cli/src/main/java/io/apicurio/registry/cli/services/Update.java
```diff
@@ -140,6 +145,54 @@ public Path downloadVersion(String version, Path targetDir) {
...
+                    + ". Expected SHA-256 " + expected + " but got " + actual
+                    + ". The download may be corrupted; please try again.");
+        }
+        log.debugf("Checksum verified: %s", actual);
+    }
+
+    /**
+     * Extracts the hex digest from a Maven checksum file, which contains the digest optionally
+     * followed by the file name.
+     */
```
paoloantinori, line 199: The checksum is fetched via `fetchBytes(...)` which returns `null` on failure (rather than throwing like the archive path does). `verifyChecksum` handles the null by throwing, so the fail-closed invariant holds — but the asymmetry between the two `fetchBytes` call sites is worth a one-line comment so a future reader doesn't assume both throw.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
```diff
@@ -122,6 +128,19 @@ public static String getShellConfigFile() {
...
+        try {
+            copyDistributionFiles(currentPath, cliHomePath);
+            deleteBackups(backups);
+        } catch (IOException | RuntimeException e) {
+            restoreBackups(backups);
+            throw e;
+        }
+    }
+
+    private void copyDistributionFiles(final Path currentPath, final Path cliHomePath) throws IOException {
```
paoloantinori, line 149: `backupExisting` only backs up `ACR_SCRIPT` and `ACR_BINARY`, but `copyDistributionFiles` also overwrites completions and README. If a later copy in that method fails, those extra files stay at the new version while only the launcher/binary are rolled back. The doc comment says 'leaves the user with the working version they had before', which is slightly stronger than the actual guarantee. Either expand the backup set to everything copied, or soften the comment to scope the rollback to the executable files.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
```diff
@@ -159,6 +178,54 @@ private void copyFiles(final Path currentPath, final Path cliHomePath) throws IO
...
+        final Map<Path, Path> backups = new LinkedHashMap<>();
+        for (final String fileName : fileNames) {
+            final Path target = cliHomePath.resolve(fileName);
+            if (Files.exists(target)) {
+                final Path backup = cliHomePath.resolve(fileName + BACKUP_SUFFIX);
+                Files.copy(target, backup, REPLACE_EXISTING, COPY_ATTRIBUTES);
+                backups.put(target, backup);
+            }
+        }
+        return backups;
```
paoloantinori, line 203: `restoreBackups` logs the restore failure but swallows it and lets the original exception propagate. If the restore itself fails (e.g. disk full), the user can be left with a corrupted binary AND no surfaced error for that corruption. Consider rethrowing or attaching the restore failure as a suppressed exception so the caller sees both the original and the rollback failure.

paoloantinori (commented):
Good security-focused hardening — SHA-256 verification of downloaded archives, bounded HTTP client timeouts/redirects, and rollback of the launcher/binary on a failed install all close real gaps in the #7642 update flow. A few observations:

- **Checksum verification is fail-closed, which is correct** — `verifyChecksum` throws if the `.sha256` is missing or doesn't match. Good. One thing worth double-checking: the `.sha256` is fetched via the same `fetchBytes(...)` path as the archive, and `fetchBytes` returns `null` on failure (rather than throwing). `verifyChecksum` handles `null` by throwing, so the fail-closed behavior holds — but the asymmetry (archive fetch throws inside `downloadVersion`, checksum fetch returns null) is worth a comment so a future reader doesn't assume both endpoints behave the same.

- **`MAX_REDIRECTS = 5` and `setFollowRedirects(true)` on the per-request options.** The `HttpClientOptions.setMaxRedirects(5)` set in `Client.getHttpClient()` configures the client-level default, and `Request.setFollowRedirects(true)` in `Update.fetchBytes` enables it per-request. Worth confirming the Vert.x version in use honors `setMaxRedirects` as a hard cap (older Vert.x had it as a hint) — if not, a malicious initial URL could still chain redirects beyond 5. Low risk here since `repo-url` is operator-configured, but worth verifying.

- **Rollback granularity in `InstallCommand.copyFiles`.** `backupExisting` backs up only `ACR_SCRIPT` and `ACR_BINARY`, but `copyDistributionFiles` copies more than those two (completions, README, etc. — see the loop over `COMPLETIONS`). If a later copy in that method fails, the launcher and binary are rolled back to their old versions but any *already-overwritten* completion/README files stay at the new (possibly partial) version. The added test (`testInstallRollsBackBinaryWhenCopyFails`) deletes `README` to trigger the failure and asserts the binary/launcher are restored — but it doesn't assert the README state, and the rollback doesn't touch it. For a self-update that's probably acceptable (README drift is harmless), but the method comment claims it 'leaves the user with the working version they had before', which is slightly stronger than what's guaranteed. Either expand the backup set to everything `copyDistributionFiles` touches, or soften the comment.

- **`BACKUP_SUFFIX = ".bak"` cleanup.** `deleteBackups` logs a warning (not an error) if a `.bak` can't be removed, and `restoreBackups` logs an error but continues if restore fails. A stale `.bak` left behind isn't dangerous, but a failed restore *is* (the user ends up with a partially-overwritten binary). Consider whether a failed restore should surface to the caller rather than just logging — currently `restoreBackups` swallows the `IOException` and the original failure propagates, masking the corrupted state.

- Minor: `parseChecksum` lowercases via `Locale.ROOT` (good, avoids tr_TR pitfalls) and `sha256Hex` already produces lowercase, so the `equals` comparison in `verifyChecksum` is case-safe. Nice.

Tests are well-targeted (`UpdateChecksumTest` covers digest-only and digest+filename checksum formats, the install test covers the rollback path). No blocking concerns; the rollback-scope mismatch is the one I'd most like clarified.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
Priyanshubhartistm: done

cli/src/main/java/io/apicurio/registry/cli/services/Update.java
Priyanshubhartistm: good catch

cli/src/main/java/io/apicurio/registry/cli/services/Update.java
```diff
@@ -140,6 +145,60 @@ public Path downloadVersion(String version, Path targetDir) {
         }
     }
 
+    /**
```
paoloantinori, line 165: Worth being explicit about what this guards against: since `checksumUri` is `fileUri + ".sha256"` — same origin, same transport — the check is integrity-against-corruption (partial downloads, CDN mangles, bit-rot), not integrity-against-a-repo-side-attacker, who controls both files. TLS remains the real MITM defense. If supply-chain tampering by someone who can publish is in scope for #8029, that needs detached signatures (GPG/cosign), not a co-located checksum. Fine to land as-is; just flagging the boundary so the threat model is clear.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
```diff
@@ -121,7 +127,26 @@ public static String getShellConfigFile() {
      * The distribution ZIP already contains the correct OS-specific acr_env file,
      * so no OS detection is needed for file selection.
      */
+    // Files overwritten or modified in place by copyDistributionFiles; all are backed up so a
+    // failed install can be rolled back to the previous working installation.
```
paoloantinori, line 131: This list has to stay in lock-step with `copyDistributionFiles` — any file written there but missing here silently escapes rollback, leaving a new/old mix on a mid-copy failure. The comment helps today; consider deriving both the backup list and the copy list from a single source (or asserting equality in `testInstallRollsBackFilesWhenCopyFails`) so the invariant can't drift.

paoloantinori (commented):
Solid hardening overall — fail-closed checksum verification, bounded HTTP client timeouts/redirects, and file-level rollback that attaches restore failures as suppressed exceptions are all the right shape. Two observations worth a look before this lands:

- **Checksum threat-model boundary (`Update.java`, `verifyChecksum`):** the `.sha256` is fetched from the same origin and over the same transport as the archive, so it robustly catches corruption, partial downloads, and CDN mangles, but it does *not* defend against an attacker who can write to the download endpoint — they'd serve a matching archive+checksum pair. The effective defense against on-path network attackers remains TLS (already in use here). If compromise-of-the-publishing-repo is a threat #8029 is meant to address, that would need detached signature verification (GPG/cosign) rather than a co-located checksum. Worth stating that boundary explicitly so the issue closes against the right threat model.

- **Rollback coverage hinges on `INSTALLED_FILES` staying in sync (`InstallCommand.java`):** `copyDistributionFiles` writes six files and `INSTALLED_FILES` enumerates the same six for backup. A future addition to `copyDistributionFiles` that isn't mirrored in `INSTALLED_FILES` would silently escape backup, so a mid-copy failure could leave a mix of new and old files behind. The existing comment helps; deriving both from a single list (or asserting the invariant in a test) would make it self-maintaining.

Both non-blocking — the current implementation is correct for the files and threats it covers.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
Priyanshubhartistm: done

Priyanshubhartistm:
Hi @jsenko , @carlesarnal could you please take a look? If anything needs to be changed please let me know
Thank you

jsenko (commented):
## Review: Clean hardening, ready for maintainer decision

Well-implemented security hardening for the CLI update flow:

- **Checksum verification**: Fail-closed SHA-256 check against `.sha256` sidecar files. `fetchBytes` null → hard failure. Correct threat model (guards against corruption, not repo-side attacker — TLS is the MITM defense, as paoloantinori noted).
- **HTTP hardening**: 30s connect, 120s idle, 5 redirects, follow-redirects enabled. Reasonable defaults for Maven Central CDN patterns.
- **Rollback**: All 6 installed files backed up before overwrite, restored on failure, restore errors attached as suppressed exceptions. `INSTALLED_FILES` list must stay in sync with `copyDistributionFiles` — paoloantinori's drift concern is valid; the test covers current state but doesn't enforce sync structurally.
- **Tests**: Checksum parsing/hex encoding, rollback on incomplete distribution, e2e script updated with `.sha256` files.

paoloantinori's two rounds of feedback appear fully addressed. No new findings from my side. The WIP label may be outdated — the implementation looks complete.

jsenko:
hi @Priyanshubhartistm we're tracking this PR and I'll review it ASAP. There's a lot of open PRs, please be patient. Thanks.

Priyanshubhartistm:
@jsenko sure , thanks

jsenko (changes requested):
Thanks @Priyanshubhartistm — solid hardening work. The checksum verification, HTTP timeouts, and rollback mechanism are well-structured. A few changes needed before merging, most are minor and can be fixed together quickly.

**1. Checksum verification should be skippable for custom repos (should-fix)**

`verifyChecksum` is unconditional — if the repo doesn't publish `.sha256` files, the update fails. This breaks users with private Maven repos that don't publish checksum sidecar files. Add a config property `update.skip-checksum-verification` (default `false`) following the `unsafeCredentialStorage` pattern: opt-in, with a log warning when active. This matches the existing `update.*` config namespace (`update.check-enabled`, `update.timeout-seconds`).

**2. FQN imports in `Update.java` (should-fix — checkstyle)**

Lines 73-74 use `java.util.Optional::isPresent` / `java.util.Optional::get` inline, and line 223 uses `io.vertx.core.http.RequestOptions` as a FQN. Import them at the top of the file.

**3. `RuntimeException` wrapping in `InstallCommand.updateShellConfiguration` (should-fix)**

Line 286: `throw new RuntimeException(e)` wraps an `IOException`. Use `CliException` for consistency with the rest of the class, or let the `IOException` propagate (the method signature allows adding `throws IOException` since the caller already handles it).

**4. `sha256Hex` can use `HexFormat` (nice-to-have)**

The hand-rolled hex loop in `sha256Hex` can be replaced with `HexFormat.of().formatHex(digest)` (available since Java 17, which is our source level). One-liner, no dependency needed.

**5. Test coverage: prefer high-level negative test over low-level unit tests (should-fix)**

`UpdateChecksumTest` tests `sha256Hex` and `parseChecksum` in isolation, but the e2e `test-update.sh` already implicitly covers these — if either were wrong, the checksum verification would fail and the update test would catch it.

What's actually missing is a **negative test**: what happens when the checksum doesn't match? Consider replacing `UpdateChecksumTest` with a case in `test-update.sh` that publishes an archive with a wrong `.sha256` and verifies `acr update` fails with the expected error message. That exercises the real code path end-to-end and is more valuable than testing hex encoding of "abc".

Priyanshubhartistm:
@jsenko Could you please take a look? If anything needs to be changed, please let me know. 
Thank you!

jsenko (commented):
Re-reviewed the new commit (`5e2874e8e`). All 5 findings from the previous review are addressed:

1. ✅ `update.skip-checksum-verification` config property — fail-closed default, log warning when active, documented in README
2. ✅ FQN imports fixed (`Optional`, `RequestOptions`)
3. ✅ `RuntimeException` wrapping removed — IOException propagates naturally
4. ✅ `HexFormat.of().formatHex()` replaces hand-rolled loop
5. ✅ `UpdateChecksumTest` replaced with e2e negative test (Test 11 — checksum mismatch aborts update, binary untouched)

Clean, focused commit. SonarCloud passing. No further concerns from my side.

jsenko (approved):

jsenko:
Merged, thank you!

Issue #8648, opened 2026-07-15
CLI: Colorized output


vandanayadav7:
## Parent Epic
Part of #7162 (Optional Extra Tasks)

## Description
Add color to CLI output for better readability — green for success messages, red for errors, yellow for warnings. [Jansi](https://fusesource.github.io/jansi) handles cross-platform terminal color support.

## Scope
- Add Jansi (or alternative) dependency if not already present
- Colorize success messages (green), error messages (red), warnings (yellow)
- Respect `NO_COLOR` environment variable (https://no-color.org/)
- Disable colors when output is piped (non-TTY)
- Optional `--no-color` flag override

## Acceptance Criteria
- Success, error, and warning messages are colorized
- Colors disabled when piped or `NO_COLOR` is set
- `--no-color` flag works
- GraalVM native image compatibility verified

NirbhayKachhatiya:
I would like to work on this issue if its available.
Thank you.

NirbhayKachhatiya:
Hi @vandanayadav7 , I have submitted PR #8773 for this feature. It implements colored output using Jansi, respects NO_COLOR, handles non-TTY piping, adds the --no-color flag, and maintains GraalVM compatibility. Please have a look when you get a chance!

PR #8773, opened 2026-07-21
feat(cli): Add cross-platform colored output with Jansi 

NirbhayKachhatiya:
#8648 
- Colorize success (green), error (red), warning (yellow) messages
- Respect NO_COLOR environment variable
- Disable colors when output is piped (non-TTY)
- Support --no-color flag
- GraalVM native image compatible

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -31,7 +31,7 @@
         <dependency>
             <groupId>org.fusesource.jansi</groupId>
             <artifactId>jansi</artifactId>
-            <version>2.4.0</version>
+            <version>${version.jansi}</version>
         </dependency>
         
         <dependency>
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -29,6 +29,12 @@
     <dependencies>
 
         <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
+            <version>2.4.0</version>
+        </dependency>
+        
+        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-picocli</artifactId>
         </dependency>
--- a/cli/src/main/java/io/apicurio/registry/cli/Acr.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/Acr.java
@@ -16,6 +16,7 @@ import io.apicurio.registry.cli.serverconfig.ServerConfigCommand;
 import io.quarkus.picocli.runtime.annotations.TopCommand;
 import lombok.Getter;
 import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
 
 import static picocli.CommandLine.Option;
 import static picocli.CommandLine.ScopeType.INHERIT;
@@ -74,4 +75,13 @@ public class Acr {
             scope = INHERIT
     )
     private boolean _ignored;
+
+    @Option(
+            names = {"--no-color"},
+            description = "Disable colored output.",
+            scope = INHERIT
+    )
+    @Getter
+    private boolean noColor;
+    
 }
--- a/cli/src/main/java/io/apicurio/registry/cli/Acr.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/Acr.java
@@ -16,7 +16,6 @@ import io.apicurio.registry.cli.serverconfig.ServerConfigCommand;
 import io.quarkus.picocli.runtime.annotations.TopCommand;
 import lombok.Getter;
 import picocli.CommandLine.Command;
-import picocli.CommandLine.Option;
 
 import static picocli.CommandLine.Option;
 import static picocli.CommandLine.ScopeType.INHERIT;
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -41,22 +41,15 @@ public abstract class AbstractCommand implements Callable<Integer> {
 
     @Override
     public Integer call() {
-        ColorUtil.init();
         var output = new OutputBuffer(config.getStdOut(), config.getStdErr());
+
         boolean noColorEnv = System.getenv("NO_COLOR") != null;
-        boolean isTty = System.console() != null;
-        boolean globalNoColor = false;
-        try {
-            var root = spec.root().userObject();
-            if (root instanceof Acr acr) {
-                globalNoColor = acr.isNoColor();
-            }
-        } catch (Exception e) {
-            
-        }
-        boolean colorsEnabled = !globalNoColor && !noColorEnv && isTty;
-        io.apicurio.registry.cli.utils.ColorUtil.setEnabled(colorsEnabled);
-        
+        boolean stdoutTty = isStdoutTerminal();
+        boolean globalNoColor = spec.root().userObject() instanceof Acr acr && acr.isNoColor();
+        boolean colorsEnabled = shouldEnableColors(globalNoColor, noColorEnv, stdoutTty);
+
+        ColorUtil.init(colorsEnabled);
+
         try {
             configureVerboseLogging();
             updateNotifier.checkAndNotify(getTopLevelCommandName());
@@ -77,10 +70,32 @@ public abstract class AbstractCommand implements Callable<Integer> {
             return APPLICATION_ERROR_RETURN_CODE;
         } finally {
             output.print();
+            ColorUtil.shutdown();
         }
     }
 
+    private boolean isStdoutTerminal() {
+        try {
+            Class<?> cLibrary = Class.forName("org.fusesource.jansi.internal.CLibrary");
+            var isatty = cLibrary.getMethod("isatty", int.class);
+            Object result = isatty.invoke(null, 1); // STDOUT file descriptor
+            if (result instanceof Integer value) {
+                return value == 1;
+            }
+            if (result instanceof Boolean value) {
+                return value;
+            }
+        } catch (ReflectiveOperationException ex) {
+            log.debug("Could not resolve stdout TTY via Jansi CLibrary, falling back to System.console().", ex);
+        }
+        return System.console() != null;
+    }
+ 
     public abstract void run(OutputBuffer output) throws Exception;
+
+    static boolean shouldEnableColors(boolean globalNoColor, boolean noColorEnv, boolean stdoutTty) {
+        return !globalNoColor && !noColorEnv && stdoutTty;
+    }
 
     private static void handleCliException(final OutputBuffer output, final CliException ex) {
         if (!ex.isQuiet()) {
@@ -115,11 +130,8 @@ public abstract class AbstractCommand implements Callable<Integer> {
     private void configureVerboseLogging() {
         var root = spec.root().userObject();
         if (root instanceof Acr acr && acr.isVerbose()) {
-            // Set the root logger level to FINE (DEBUG equivalent)
             var rootLogger = java.util.logging.Logger.getLogger("");
             rootLogger.setLevel(java.util.logging.Level.FINE);
-            // Also lower handler levels — Quarkus sets quarkus.log.level=WARN on the
-            // console handler, which filters debug messages even when the logger allows them.
             for (var handler : rootLogger.getHandlers()) {
                 handler.setLevel(java.util.logging.Level.FINE);
             }
--- a/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
@@ -4,6 +4,7 @@ import io.apicurio.registry.cli.Acr;
 import io.apicurio.registry.cli.config.Config;
 import io.apicurio.registry.cli.services.Client;
 import io.apicurio.registry.cli.services.UpdateNotifier;
+import io.apicurio.registry.cli.utils.ColorUtil;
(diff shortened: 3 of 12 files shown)


cli/pom.xml
```diff
@@ -28,6 +28,12 @@
 
     <dependencies>
 
+        <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
```
paoloantinori, line 33: Jansi loads a native library via JNI at runtime; since the PR states native-image compatibility, worth confirming the CLI's native-image build still produces a working binary with Jansi's native lib/metadata included. Also, the version is hardcoded here rather than managed via a property/BOM — worth checking whether the project convention is to manage third-party versions centrally.

cli/src/main/java/io/apicurio/registry/cli/common/AbstractCommand.java
```diff
@@ -40,7 +41,22 @@ public abstract class AbstractCommand implements Callable<Integer> {
 
     @Override
     public Integer call() {
+        ColorUtil.init();
         var output = new OutputBuffer(config.getStdOut(), config.getStdErr());
+        boolean noColorEnv = System.getenv("NO_COLOR") != null;
+        boolean isTty = System.console() != null;
```
paoloantinori: `System.console()` returns non-null only when *both* stdin and stdout are terminals, so redirecting just stdin (e.g. `acr < input.txt`) disables colors even though stdout is still a terminal. The NO_COLOR/isatty convention is about the output stream — if exactness matters, consider a stdout-focused check.

cli/src/main/java/io/apicurio/registry/cli/utils/ColorUtil.java
```diff
@@ -0,0 +1,45 @@
...
+
+public final class ColorUtil {
+
+    private ColorUtil() {
+    }
+
+    private static volatile boolean enabled = true;
+
+    public static void init() {
+        AnsiConsole.systemInstall();
```
paoloantinori: `systemInstall()` is called unconditionally (even when colors are later disabled) and is never paired with `systemUninstall()`. Consider gating it on `colorsEnabled` and uninstalling in a `finally` to avoid wrapping System.out/err and installing hooks on the no-color path.

cli/src/main/java/io/apicurio/registry/cli/Acr.java
```diff
@@ -16,6 +16,7 @@
 import io.quarkus.picocli.runtime.annotations.TopCommand;
 import lombok.Getter;
 import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
```
paoloantinori: This `import picocli.CommandLine.Option;` duplicates the existing `import static picocli.CommandLine.Option;` further down — both resolve to the same annotation. One is redundant; the static import is what the other options in this file already rely on.

paoloantinori (commented):
Nice DX improvement and well-scoped — the `OutputBuffer` severity abstraction is a clean way to route colorization, and the `NO_COLOR` / non-TTY / `--no-color` handling covers the right cases. A few things worth tightening before this lands.

- **No tests added.** `ColorUtil.colorizeSuccess/Error/Warning` are pure string functions and `OutputBuffer`'s severity→color routing is straightforwardly unit-testable, but neither is covered. The `--no-color` flag, `NO_COLOR` env var, and non-TTY disable paths in `AbstractCommand.call()` also have no coverage. Even a couple of small unit tests on `ColorUtil` (enabled vs disabled, null/blank passthrough) would lock in the contract and catch regressions if the Jansi calls change.
- **Native-image claim worth verifying.** Jansi loads a native library (`libjansi`) via JNI when `AnsiConsole.systemInstall()` runs. Jansi 2.4.0 ships GraalVM reachability metadata, so this *should* work, but since the PR description explicitly states native-image compatibility it'd be good to confirm an actual native-image build of the CLI still produces a working binary (the `native-image` smoke test, if the CLI module has one). See the [Jansi native-image notes](https://github.com/fusesource/jansi#graalvm-native-image).
- **`System.console()` conflates stdin and stdout for TTY detection.** `System.console()` returns non-null only when *both* stdin and stdout are interactive terminals, so `acr < input.txt` (stdin redirected, stdout still a terminal) would disable colors even though output is going to a terminal. The `NO_COLOR`/isatty convention is defined over the *output* stream. This is a common shortcut and probably fine in practice, but worth a conscious decision — if exactness matters, a stdout-focused isatty check would be more correct.
- **`AnsiConsole.systemInstall()` is unconditional and never paired with `systemUninstall()`.** It runs even when colors are subsequently disabled (non-TTY/`NO_COLOR`/`--no-color`), wrapping `System.out`/`System.err` and installing shutdown hooks regardless. For a short-lived CLI this is harmless, but calling `systemInstall()` only when `colorsEnabled` is true (and tearing down with `systemUninstall()` in a `finally`) would avoid unnecessary stream wrapping on the disabled path.
- Minor: in `Acr.java` the new `import picocli.CommandLine.Option;` is redundant with the pre-existing `import static picocli.CommandLine.Option;` (both resolve to the same annotation). It compiles, but one of the two can go.

NirbhayKachhatiya:
Thanks for the detailed review @paoloantinori! These are great callouts.

Here's how I'll address them in the next commit:
1. **Tests:** I'll add unit tests for `ColorUtil` (testing enabled/disabled states and edge cases like null/blank inputs) as well as tests covering the severity routing and `--no-color` / `NO_COLOR` flag logic.
2. **Lifecycle Cleanup:** Good point on `AnsiConsole`. I'll update it to conditionally call `AnsiConsole.systemInstall()` only when colors are enabled and wrap execution so `AnsiConsole.systemUninstall()` runs in a `finally` block.
3. **Redundant Import:** Will remove the duplicate `picocli.CommandLine.Option` import in `Acr.java`.
4. **Native Image & TTY:** I'll verify the native-image build smoke test to make sure the JNI library links smoothly. For `System.console()`, I'll check if a clean stdout TTY check makes sense without adding unnecessary complexity.

I'll push an update shortly!

paoloantinori:
Sounds good @NirbhayKachhatiya. Looking forward to the next commit — particularly the `ColorUtil` tests and the `AnsiConsole.systemInstall()` lifecycle guard. Ping me again once those are pushed and I'll take another look.

NirbhayKachhatiya:
Hi @paoloantinori  ,  thanks again for the detailed review.
I addressed the comments to the best of my ability. I’m still new to this codebase, so it took me a bit more time to implement the fixes carefully.
Also, apologies for the earlier native-image compatibility claim. I can’t update the original wording directly from my side right now, but to clarify: I have not independently verified native-image output locally. My intent is that compatibility is expected (based on Jansi metadata), and final confirmation should come from the native build/CI verification path.
If you can, could you please run/confirm that native verification step (or point me to the exact CI job I should reference)? I’d really appreciate it.

paoloantinori:
Hi @NirbhayKachhatiya, thanks for the follow-up and for the honest correction on the native-image claim — that's appreciated.

The updates address the earlier points well:
- `ColorUtilTest` and `OutputBufferTest` cover the enabled/disabled paths, null/blank pass-through, severity routing, and incremental `print()`. Extracting the decision logic into the static `shouldEnableColors(...)` with its own `AbstractCommandColorDecisionTest` is a nice touch.
- The `AnsiConsole.systemInstall()` lifecycle is now guarded by the `installed` flag and paired with `systemUninstall()` in the `finally` block of `AbstractCommand.call()`, so the install/uninstall is balanced.

One forward-looking note on native-image, since you flagged it: the part most likely to need verification there is the reflection-based `isStdoutTerminal()` — `org.fusesource.jansi.internal.CLibrary` is an internal Jansi class, so under native-image the `Class.forName(...)` lookup may not resolve without a `reflect-config` entry. The good news is the `catch (ReflectiveOperationException)` falls back to `System.console() != null`, so the worst case is degraded (but still correct) TTY detection rather than a crash. I wouldn't block on it, but it's the spot to watch if/when someone validates the native-image build.

The PR is `behind` main, so a rebase will be needed before merge, but content-wise this looks good to me. Thanks again for the careful iteration.

NirbhayKachhatiya:
Rebased on top of main and all review feedback has been addressed.

/ready

NirbhayKachhatiya:
@jsenko @EricWittmann  Could one of you please approve running the pending CI workflows (Lint and Validate, Verification Gate) when you get a chance? Thanks!

jsenko (commented):
## Review: Clean DX improvement, design-level feedback

Good implementation of CLI colorization with proper conventions:

- **NO_COLOR compliance**: Respects `NO_COLOR` env var (no-color.org), `--no-color` flag, and non-TTY stdout detection. `shouldEnableColors` logic is correct and testable.
- **OutputBuffer integration**: `Severity` enum with `writeSuccess`/`writeWarning`/`writeError` is a clean abstraction — color is applied at print time, not at message construction.
- **Jansi lifecycle**: `init`/`shutdown` properly guarded — `systemInstall()` only called when colors are enabled, paired with `systemUninstall()` in the `finally` block. Addresses paoloantinori's earlier concern.
- **Tests**: `ColorUtilTest`, `OutputBufferTest`, and `AbstractCommandColorDecisionTest` cover the key paths.

### Observations

1. **TTY detection fragility**: `isStdoutTerminal()` uses reflection on `org.fusesource.jansi.internal.CLibrary` — an internal Jansi API. If Jansi changes the class path or method signature, this silently falls back to `System.console()`. Consider using Jansi's public API (e.g. `AnsiConsole.out().getTerminal()`) if one is available in the version used, or at minimum document the version coupling.

2. **Missing trailing newlines**: `ColorUtil.java` and `OutputBuffer.java` are missing final newlines — checkstyle may flag this (the project's checkstyle config typically requires them).

3. **`Severity.NONE`** is declared but never used — could be removed to keep the enum minimal.

paoloantinori's two rounds of feedback appear addressed. The WIP label may be outdated.

carlesarnal:
@NirbhayKachhatiya this has now conflicts, can you fix those and rebase main?

vandanayadav7:
## Review: Good intent, needs rebase and structural fixes before merge

The `OutputBuffer` severity abstraction is a clean design — color applied at print time rather than message construction is the right call. `NO_COLOR` / `--no-color` / non-TTY coverage follows the right conventions. Several things need to be addressed before this can land.

### Blocking

1. **Build fails.** `${version.jansi}` at `pom.xml:34` is undefined — no such property exists. On main, the property is `${jansi.version}`. After rebase, remove the duplicate dependency block entirely since jansi is already declared in `cli/pom.xml` on main.

2. **Test files are in `src/main/java`.** `ColorUtilTest`, `OutputBufferTest`, and `AbstractCommandColorDecisionTest` will fail compilation because JUnit and AssertJ are test-scoped dependencies — and even if they compiled, Surefire only scans `src/test/java`. Move them to `src/test/java`.

3. **Missing DCO sign-off.** Neither commit has the required `Signed-off-by:` line.

4. **Stale against main.** `AbstractCommand.call()` has been significantly refactored — `configureVerboseLogging()` was replaced with `configureLogging(verbose)` adding per-package log level support, `isVerbose()` was extracted, and `client.setHttpLoggingEnabled(verbose)` was added. A careful rebase is needed to avoid reverting those improvements.

### Code issues

5. **TTY detection: fragile and incorrect on JDK 22+.** `isStdoutTerminal()` uses reflection on `org.fusesource.jansi.internal.CLibrary` — an internal class that is NOT registered in the CLI's `reflect-config.json`, so it will fail in native image. The `System.console() != null` fallback is incorrect on JDK 22+ (returns `true` even for piped output), causing ANSI codes to be emitted into pipes. After the required rebase, `TerminalUtils.isTerminal()` will be available — it already handles JDK 22+ and native image correctly and should be reused.

6. **Fully qualified class name in `UpdateNotifier.java:100`.** `io.apicurio.registry.cli.utils.ColorUtil.colorizeWarning(...)` should be a regular import.

7. **Incomplete error handler conversion.** Only `handleCliException` was converted to `writeError()`. `handleRuleViolation` (line 106), `handleProblemDetails` (line 116), and the generic `catch (Exception)` block (line 69) still use `writeStdErrChunk()` which maps to `Severity.INFO` — no color. A schema validation error shows plain text while a CLI error shows red.

8. **Inconsistent colorization in UpdateNotifier.** `printNotification()` calls `ColorUtil.colorizeWarning()` directly (bypassing `OutputBuffer`), while `"Warning: Could not check for updates..."` at line 49 and `"Checking for updates..."` at line 37 get no color. Either route everything through `OutputBuffer` or be consistent.

9. **OAuth2 `UNSAFE_STORAGE_WARNING` not converted.** `LoginCommand.java:163-165` still uses `writeStdOutChunk()` for `UNSAFE_STORAGE_WARNING`, while the identical pattern in the basic auth path (line 123) was converted to `writeWarning()`. Missed conversion.

10. **UNSAFE_STORAGE_WARNING silently moved from stdout to stderr.** In the basic auth path, the `writeStdOutChunk` → `writeWarning` change routes the warning from stdout to stderr. Semantically reasonable, but it's an undocumented behavior change that could break scripts checking stdout.

11. **`Severity.NONE` is unused** — remove it.

### Minor

12. All 5 new/modified files are missing a trailing newline.
13. Blank-line removals in `OutputBuffer` and `Chunk` field declarations are cosmetic diff noise outside the PR scope.
14. `ColorUtil.setEnabled()` is unsynchronized while `init()`/`shutdown()` are `synchronized` — pick one approach.

The core design is sound — these are mostly structural and rebase items. Happy to help if you need pointers on any of the fixes.

PR #8787, opened 2026-07-21
Fix config property aria labels

Myself-Praveen:
I noticed that the Edit, Save, and Cancel icon buttons inside the ConfigProperty component were missing aria-label attributes, making them inaccessible to screen readers. I've added descriptive aria-labels to all three buttons to ensure they are properly described to visually impaired users. This is my second PR as part of applying for the LFX mentorship.

### DIFF
--- a/ui/ui-app/src/app/components/common/SortOrderToggle.tsx
+++ b/ui/ui-app/src/app/components/common/SortOrderToggle.tsx
@@ -18,7 +18,7 @@ export const SortOrderToggle: FunctionComponent<SortOrderToggleProps> = (props: 
     );
 
     return (
-        <Button icon={icon} variant="plain" aria-label={props.sortOrder === SortOrder.asc ? "sort ascending" : "sort descending"} onClick={() => {
+        <Button icon={icon} variant="plain" aria-label={props.sortOrder === SortOrder.asc ? "Sort descending" : "Sort ascending"} onClick={() => {
             props.onChange(props.sortOrder === SortOrder.asc ? SortOrder.desc : SortOrder.asc);
         }} />
     );
--- a/ui/ui-app/src/app/pages/settings/components/ConfigProperty.tsx
+++ b/ui/ui-app/src/app/pages/settings/components/ConfigProperty.tsx
@@ -95,11 +95,11 @@ export const ConfigProperty: FunctionComponent<ConfigPropertyProps> = ({ propert
                 </FlexItem>
                 <FlexItem className="actions" align={{ default: "alignRight" }}>
                     <If condition={!isEditing}>
-                        <Button icon={<PencilAltIcon />} variant="plain" className="action single" onClick={onStartEditing} />
+                        <Button icon={<PencilAltIcon />} variant="plain" aria-label="Edit configuration property" className="action single" onClick={onStartEditing} />
                     </If>
                     <If condition={isEditing}>
-                        <Button icon={<CheckIcon />} variant="plain" className="action" onClick={onSavePropertyValue} isDisabled={!isValid} />
-                        <Button icon={<CloseIcon />} variant="plain" className="action" onClick={onCancelEdit} />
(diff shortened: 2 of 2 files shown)


ui/ui-app/src/app/components/common/SortOrderToggle.tsx
```diff
@@ -18,7 +18,7 @@ export const SortOrderToggle: FunctionComponent<SortOrderToggleProps> = (props:
     );
 
     return (
-        <Button icon={icon} variant="plain"  onClick={() => {
+        <Button icon={icon} variant="plain" aria-label={props.sortOrder === SortOrder.asc ? "sort ascending" : "sort descending"} onClick={() => {
```
paoloantinori: Worth double-checking the semantics here. When `sortOrder === asc`, the `onClick` switches it to `desc`, but the label reads "sort ascending" — i.e. it describes the *current* state rather than the *action* the button performs.

For icon-only buttons, the common PatternFly/WAI-ARIA convention is to label the action the user will trigger on click (cf. a play/pause toggle labeled "Pause" while playing, because clicking pauses). Under that convention the labels would be inverted:

```tsx
aria-label={props.sortOrder === SortOrder.asc ? "sort descending" : "sort ascending"}
```

As written, a screen reader user on an ascending-sorted list hears "sort ascending" and clicking then produces the opposite. Happy to leave it if the team's convention is to describe state, but worth a conscious call since it's the crux of the a11y fix.

paoloantinori (commented):
Thanks for this accessibility pass — adding `aria-label`s to icon-only buttons is exactly the right fix, and the `ConfigProperty.tsx` labels are clear and descriptive. One thing worth double-checking on the `SortOrderToggle` change below.

Myself-Praveen:
Thanks @paoloantinori! Good catch on the WAI-ARIA convention. I've updated the SortOrderToggle component to invert the labels so they correctly describe the action (e.g., 'Sort descending') rather than the state. I also included the missing ConfigProperty accessibility fixes in this PR. Thanks again for the review.

paoloantinori:
Thanks @Myself-Praveen — the label inversion looks right: when the current order is `asc`, the button switches to descending, so "Sort descending" correctly describes the action it performs, and vice versa. The `ConfigProperty.tsx` labels ("Edit", "Save", "Cancel editing") are clear and descriptive too. This looks good to me.

Myself-Praveen:
Hi @paoloantinori, thanks again for the review and the approval! It looks like the Verification Gate is currently blocking the remaining CI checks from running since this is a fork PR. Could someone please run /accept or /retry so the full test suite can execute?

Also, a quick question regarding the LFX Mentorship for the Prompt Template Playground—I have my proposal and UI prototype finished and ready to submit, but I noticed the Apicurio project isn't showing up on the LFX Mentorship portal search just yet. Is the official application window open, or do you have a direct link I could use? Thanks so much for all your help!

paoloantinori:
Hi @Myself-Praveen — happy to help move this along.

On the CI gate: triggering `/accept` or `/retry` on a fork PR is a maintainer action I'm not going to fire automatically on Paolo's behalf, but I've surfaced it so it can be actioned. If a maintainer doesn't get to it shortly, feel free to nudge again.

On the LFX Mentorship question: I honestly don't have reliable, current info on whether Apicurio is listed on the LFX portal for this cycle or when the Prompt Template Playground project is scheduled to appear — the portal state changes per cycle and I'd rather not guess. The authoritative answer will come from the Apicurio maintainers directly, so I'd suggest raising it on the project's [mailing list / chat](https://www.apicur.io/community/) where the folks running the mentorship can confirm the timeline and where to submit your proposal. Your prototype and proposal being ready is great — just make sure you keep a copy until the submission target is confirmed.

The accessibility changes here still look good to me.

Myself-Praveen:
Hi @EricWittmann! Just following up on this—could someone please drop a quick /accept or /retry comment when you have a moment so the Verification Gate can finish running? Thank you!

Myself-Praveen:
Hi @EricWittmann! All checks have passed successfully. Could you please review and /auto-merge when you get a chance? Thank you.

EricWittmann (approved):
LGTM thanks!

PR #8769, opened 2026-07-21
feat(a2a): add full-text search for agent discovery

carlesarnal:
## Summary

- Add PostgreSQL-native full-text search (`tsvector`/`websearch_to_tsquery`) to agent discovery endpoints (`POST /.well-known/agents/search`, `GET /.well-known/agents`)
- On PostgreSQL, searches agent name, description, AND content blob (skills, tags, capabilities embedded in the agent card JSON)
- On H2/MySQL/MSSQL, falls back to case-insensitive `LIKE` on name + description
- No schema changes — all query-time computation on existing columns

## What changed

**SQL layer** — `SqlStatements.fullTextSearchClause()` returns DB-specific SQL: PostgreSQL uses `to_tsvector`/`websearch_to_tsquery` with an EXISTS subquery on the content table for deep matching; other databases use `ILIKE` on artifact name + description. The `content` filter type in `SqlSearchRepository` now works natively instead of throwing `ContentSearchNotSupportedException`.

**REST layer** — `WellKnownResourceImpl` switches the query field from `SearchFilter.ofName()` (wildcard LIKE on artifactId/name) to `SearchFilter.ofContent()` (full-text search). Both POST and GET agent search endpoints are updated.

**Tests** — Unit tests (H2 path) verify name, description, partial match, and no-match search. Integration tests (`AgentSearchIT`, tagged `@Tag(SMOKE)`) verify the same scenarios against PostgreSQL in CI.

Closes #7230 (Phase 1)

## Test plan

- [x] `./mvnw test -pl app -Dtest=WellKnownResourceTest` — 26 tests pass (H2/ILIKE path)
- [x] `./mvnw checkstyle:check` — 0 violations on all changed files
- [ ] CI: `storage: postgresql` (`remote-sql`) — exercises tsvector/tsquery path
- [ ] CI: `storage: h2` (`remote-mem`) — exercises ILIKE fallback
- [ ] CI: `storage: kafkasql` (`remote-kafka`) — delegates to SQL, same path

PR #8788, opened 2026-07-21
fix(rest): apply partial match to name filter on well-known discovery endpoints

### LINKED ISSUE #8703: A2A agent and MCP tool search: name filter does exact match despite documented partial match
The `name` filter on `/.well-known/agents` is documented as a partial match:

```
@param name filter by agent name (partial match)
```

([WellKnownResource.java#L107](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java#L107), same wording for tools at [#L147](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java#L147))

In practice it's an exact match. With 18 agent cards registered (`agent-01`..`agent-18`, one with metadata name `WeatherAgent`):

```
?name=WeatherAgent  -> 1
?name=Weather       -> 0
?name=agent         -> 0
?name=*agent*       -> 18
?name=*Weather*     -> 1
```

So the search handles wildcards fine, the endpoint just doesn't add them. Same on `/.well-known/mcp-tools` with an MCP_TOOL named `get_weather`:

```
?name=get_weather -> 1
?name=weather     -> 0
?name=*weather*   -> 1
```

Cause: `searchAgents` ([L332](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java#L332)) and `searchMcpTools` ([L531](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java#L531)) pass the value straight through:

```java
filters.add(SearchFilter.ofName(name));
```

while `searchAgentsAdvanced` ([L193-199](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java#L193-L199)) wraps it first:

```java
if (!q.contains("*")) {
    q = "*" + q + "*";
}
filters.add(SearchFilter.ofName(q));
```

Practical effect: the "Search by name" box on the Agents page finds nothing unless you type the full name exactly.

One question before I touch it: with wildcards it's still case sensitive (`*AGENT*` -> 0, `*agent*` -> 18). Leave that as is, or should the filter be case insensitive too?

Happy to PR this if you want it.

Tested on `main` @ `5115d1136`, in-memory storage, with `apicurio.a2a.enabled`, `apicurio.mcp-tools.enabled` and `apicurio.ui.features.agents.enabled` turned on.

karanthengane23:
The `name` filter on `/.well-known/agents` and `/.well-known/mcp-tools` is documented as a partial match, but the raw value was passed to `SearchFilter.ofName`, so it only matched exactly. `?name=weather` returned nothing while `?name=*weather*` worked.

This wraps the value in wildcards unless the caller supplied their own, the same way `searchAgentsAdvanced` already does.

Two notes:

- Kept `contains("*")` so the behaviour matches the existing `searchAgentsAdvanced` rather than widening the change. There is a discussion on #8703 about using `startsWith`/`endsWith` instead, which would also mean touching `searchAgentsAdvanced`. Happy to switch if you prefer that.
- Case sensitivity left as is, per #8703.

Tests added:

- partial name match on agents
- partial name match on mcp-tools
- explicit `*` is not wrapped twice

`apicurio.mcp-tools.enabled` is used in `ExperimentalFeaturesEnabledProfile` so the mcp-tools endpoint is reachable in the test. That profile is only used by `WellKnownResourceTest`.

Checked locally: `?name=weather` matches, `?name=<full-artifact-id>` still matches exactly, `?name=zzzznope` returns nothing. 24 tests pass, checkstyle clean.

Fixes #8703

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
@@ -337,9 +337,14 @@ public class WellKnownResourceImpl implements WellKnownResource {
         // Always filter by AGENT_CARD artifact type
         filters.add(SearchFilter.ofArtifactType(ArtifactType.AGENT_CARD));
 
-        // Filter by name if provided
+        // Filter by name if provided. The name filter is documented as a partial match, so
+        // wrap the value in wildcards unless the caller supplied their own.
         if (!StringUtil.isEmpty(name)) {
-            filters.add(SearchFilter.ofName(name));
+            String nameFilter = name.trim();
+            if (!nameFilter.contains("*")) {
+                nameFilter = "*" + nameFilter + "*";
+            }
+            filters.add(SearchFilter.ofName(nameFilter));
         }
 
         // Filter by skills (indexed as structured content: agent_card:skill:<id>)
@@ -540,8 +545,14 @@ public class WellKnownResourceImpl implements WellKnownResource {
 
         filters.add(SearchFilter.ofArtifactType(ArtifactType.MCP_TOOL));
 
+        // The name filter is documented as a partial match, so wrap the value in wildcards
+        // unless the caller supplied their own.
         if (!StringUtil.isEmpty(name)) {
-            filters.add(SearchFilter.ofName(name));
+            String nameFilter = name.trim();
+            if (!nameFilter.contains("*")) {
+                nameFilter = "*" + nameFilter + "*";
+            }
+            filters.add(SearchFilter.ofName(nameFilter));
         }
 
         if (parameters != null && !parameters.isEmpty()) {
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
@@ -201,11 +201,7 @@ public class WellKnownResourceImpl implements WellKnownResource {
         // Query matches artifact metadata name and artifactId (not Agent Card JSON content).
         // Full-text content search is tracked in #7230.
         if (!StringUtil.isEmpty(request.getQuery())) {
-            String q = request.getQuery().trim();
-            if (!q.contains("*")) {
-                q = "*" + q + "*";
-            }
-            filters.add(SearchFilter.ofName(q));
+            filters.add(SearchFilter.ofPartialName(request.getQuery()));
         }
 
         AgentSearchFilters f = request.getFilters();
@@ -340,11 +336,7 @@ public class WellKnownResourceImpl implements WellKnownResource {
         // Filter by name if provided. The name filter is documented as a partial match, so
         // wrap the value in wildcards unless the caller supplied their own.
         if (!StringUtil.isEmpty(name)) {
-            String nameFilter = name.trim();
-            if (!nameFilter.contains("*")) {
-                nameFilter = "*" + nameFilter + "*";
-            }
-            filters.add(SearchFilter.ofName(nameFilter));
+            filters.add(SearchFilter.ofPartialName(name));
         }
 
         // Filter by skills (indexed as structured content: agent_card:skill:<id>)
@@ -548,11 +540,7 @@ public class WellKnownResourceImpl implements WellKnownResource {
         // The name filter is documented as a partial match, so wrap the value in wildcards
         // unless the caller supplied their own.
         if (!StringUtil.isEmpty(name)) {
-            String nameFilter = name.trim();
-            if (!nameFilter.contains("*")) {
-                nameFilter = "*" + nameFilter + "*";
-            }
-            filters.add(SearchFilter.ofName(nameFilter));
+            filters.add(SearchFilter.ofPartialName(name));
         }
 
         if (parameters != null && !parameters.isEmpty()) {
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
@@ -347,6 +347,18 @@ public class WellKnownResourceTest extends AbstractResourceTestBase {
     }
 
     @Test
+    public void testSearchAgentsWhitespaceNameDoesNotMatchAll() throws Exception {
+        // Whitespace-only name should be trimmed to empty string rather than wrapped into "**".
+        givenAtRoot()
+                .when()
+                .contentType(CT_JSON)
+                .queryParam("name", "   ")
+                .get("/.well-known/agents")
+                .then()
+                .statusCode(200);
+    }
+
+    @Test
     public void testSearchAgentsWithPagination() throws Exception {
         String groupId = TestUtils.generateGroupId();
 
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
@@ -103,6 +103,20 @@ public class WellKnownResourceTest extends AbstractResourceTestBase {
             }
             """;
 
+    private static final String MCP_TOOL_CONTENT = """
+            {
+                "name": "get_weather",
+                "description": "Get the current weather for a city",
+                "inputSchema": {
+                    "type": "object",
+                    "properties": {
+                        "city": { "type": "string" }
+                    },
+                    "required": ["city"]
+                }
+            }
+            """;
+
     @Test
     public void testGetAgentCard() {
         givenAtRoot()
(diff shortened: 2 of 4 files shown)


app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -327,9 +327,14 @@ public AgentSearchResults searchAgents(String name, List<String> skills, List<St
         // Always filter by AGENT_CARD artifact type
         filters.add(SearchFilter.ofArtifactType(ArtifactType.AGENT_CARD));
 
-        // Filter by name if provided
+        // Filter by name if provided. The name filter is documented as a partial match, so
+        // wrap the value in wildcards unless the caller supplied their own.
         if (!StringUtil.isEmpty(name)) {
-            filters.add(SearchFilter.ofName(name));
+            String nameFilter = name.trim();
```
paoloantinori: This `contains("*")` check is now the third copy of the same wildcard-wrapping logic (here, in `searchMcpTools` below, and in `searchAgentsAdvanced`). If #8703 moves toward `startsWith`/`endsWith`, all three sites would need the same change — extracting a helper would make that a one-line edit and prevent drift.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
```diff
@@ -247,6 +261,61 @@ public void testSearchAgents() throws Exception {
...
+                .body("agents.artifactId", hasItem("partialmatchagent-alpha"));
+    }
+
+    @Test
+    public void testSearchAgentsExplicitWildcardIsPreserved() throws Exception {
+        String groupId = TestUtils.generateGroupId();
+
+        createAgentCard(groupId, "explicitwildcardagent-alpha", AGENT_CARD_CONTENT);
+
+        // A caller-supplied wildcard must still work (the value is not wrapped a second time).
```
paoloantinori, line 288: Worth adding an assertion for the prefix-only (`partialmatchagent*`) and suffix-only (`*partialmatchagent`) cases, to document that the `contains("*")` guard in `WellKnownResourceImpl` preserves caller-supplied wildcards as-is rather than wrapping them again.

paoloantinori (commented):
Clean, minimal fix that aligns the `name` filter on the well-known discovery endpoints with the partial-match behavior already documented (and already implemented in `searchAgentsAdvanced`). The test additions — partial match, explicit-wildcard preservation, MCP tools — directly cover the regression. Two minor observations below, neither blocking.

- The wildcard-wrapping logic (`trim`, then wrap in `*…*` unless the caller already supplied a `*`) is now duplicated between `searchAgents` and `searchMcpTools`, and the PR description notes `searchAgentsAdvanced` has a third copy. Worth extracting a small helper (e.g. `SearchFilter.ofPartialName` or a private `wrapNameWildcard`) so the three sites can't drift — particularly around the `contains("*")` vs. `startsWith`/`endsWith` decision referenced in #8703.
- Test coverage is solid for the full-substring case and the already-wildcarded case, but there's no test asserting that a prefix-only (`weather*`) or suffix-only (`*weather`) query is preserved as-is rather than double-wrapped. Since the `contains("*")` guard is the core of the fix, a one-line assertion documenting that intent would help future readers.

karanthengane23:
Thanks for the review, both addressed in 77b5712.

- Moved the wrapping into `SearchFilter.ofPartialName` and used it in `searchAgents`, `searchMcpTools` and `searchAgentsAdvanced`, so the rule lives in one place instead of three.
- Added `testSearchAgentsPartialWildcardIsPreserved`, covering both directions: `boundedwildcardagent*` matches the agent starting with the term but not the one that only contains it, and `*boundedwildcardagent-beta` the other way round.

On `contains("*")` vs `startsWith`/`endsWith` from #8703: tracing it through `buildNameClause`, the two only differ for an interior wildcard like `we*ther`, and neither one makes that work, since only leading/trailing `*` get translated to `%`. So it is a consistency choice rather than a behaviour fix. Kept `contains` for now, but with the helper it is a one-line change in one place if you prefer the other.

25 tests pass locally, checkstyle clean.

karanthengane23:
@carlesarnal I noticed #8769 while looking at the A2A code. It switches the query field on both agent endpoints from `SearchFilter.ofName` to `SearchFilter.ofContent`, so once that lands the agents half of this PR is redundant, and the wildcard wrapping in `searchAgentsAdvanced` goes away with it.

`searchMcpTools` isn't covered there though. It still passes the raw value to `SearchFilter.ofName`, so `?name=weather` on `/.well-known/mcp-tools` returns nothing while `?name=*weather*` works. That is the part of #8703 that #8769 doesn't fix.

Happy to rebase this onto #8769 and narrow it to `searchMcpTools` plus its test, so it stops touching the files you are changing. In that case `ofPartialName` would only have one caller, so I can either keep it or inline it back, whichever you prefer.

Let me know which way you want it and I'll push the change. No rush from my side, it makes sense to land #8769 first.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
karanthengane23: Done, extracted into SearchFilter.ofPartialName and used it in searchAgents, searchMcpTools, and searchAgentsAdvanced.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
karanthengane23: Added testSearchAgentsPartialWildcardIsPreserved, asserting a prefix-only (boundedwildcardagent*) and suffix-only (*boundedwildcardagent-beta) query are preserved as-is rather than re-wrapped.

karanthengane23:
Still relevant and ready. This is waiting on #8838, which reworks the same agent-search code. Once that lands I'm happy to rebase and narrow this to the `searchMcpTools` fix, which #8838 doesn't cover. Keeping it open.

/unstale

EricWittmann:
Hi @karanthengane23 — this PR doesn't actually depend on #8838, and we may not merge that one. Your fix stands on its own as a valid bug fix for partial name matching.

Could you rebase onto main and resolve the conflicts? Once that's done we'd like to merge this. Thanks!

karanthengane23:
Hi @EricWittmann , I've rebased onto main, resolved the conflict, and everything is ready for review. Thanks!

karanthengane23:
Hi @EricWittmann , just a heads up on the bot comment: the Build UI Application failure is an upstream issue on main from PR #9082 (commit c55b5d1a5), where typescript-sdk wasn't built before running UI tests. All Java compilation & tests in this PR pass 100%.

paoloantinori (commented):
This is a correct, minimal fix: the `name` filters on `searchAgents` and `searchMcpTools` now do the documented partial match, and the duplicated wildcard-wrap logic from `searchAgentsAdvanced` is cleanly extracted into `SearchFilter.ofPartialName` with no behavior change on that path. Tests are strong and would catch a regression; I found no blocking issues, just two minor hedged notes.

- The `searchAgentsAdvanced` change (`WellKnownResourceImpl.java:201`) is behavior-equivalent to the old inline trim / `contains("*")` / wrap / `ofName` sequence — same steps, same order — so it is pure de-duplication, not a behavior shift.
- The actual fix at `WellKnownResourceImpl.java:334` (agents) and `:539` (mcp-tools) correctly switches `ofName(name)` → `ofPartialName(name)`. Both call sites are guarded by `StringUtil.isEmpty(...)`, so null/empty cannot reach `value.trim()` inside `ofPartialName` (`SearchFilter.java:60`).
- `ofPartialName` preserves a caller-supplied wildcard via `contains("*")` (`SearchFilter.java:62`), matching `searchAgentsAdvanced`'s prior behavior; edge inputs (`x*`, `*x`, `x*y`, bare `*`) all pass through unwrapped as expected.
- Tests are thorough — `testSearchAgentsPartialWildcardIsPreserved` (`WellKnownResourceTest.java:298`) in particular proves a prefix-only search stays prefix-only and would fail on a double-wrap regression. Good coverage of substring, explicit-wildcard, prefix/suffix, and the mcp-tools endpoint.
- Minor / unverified: a whitespace-only `name` (e.g. `?name=%20%20%20`) now trims to `""` and wraps to `"**"` on the two endpoints, which *may* match everything depending on storage-layer wildcard semantics. This matches `searchAgentsAdvanced`'s pre-existing behavior (so it's consistency-by-design, not a new regression), and it only matters if `StringUtil.isEmpty` lets whitespace through — which I couldn't confirm against source. Worth a one-line guard or a blank-input test if matching-all on blank input is undesirable.
- Minor: the PR description says `apicurio.mcp-tools.enabled` is "added" to `ExperimentalFeaturesEnabledProfile`, but that file's diff is comment-only (+2/-2, `ExperimentalFeaturesEnabledProfile.java:8`), implying the flag was already present. A one-line wording fix in the description would avoid confusion; no code impact.

karanthengane23:
Thank you @paoloantinori for the thorough review and feedback! Great catch on the whitespace edge case — I've added a guard `if (name.isEmpty()) return ofName("");` in `SearchFilter.ofPartialName` along with a test (`testSearchAgentsWhitespaceNameDoesNotMatchAll`) so whitespace-only queries don't get wrapped into `"**"`.

paoloantinori:
Thanks @karanthengane23 — the whitespace guard in `ofPartialName` (returning `ofName("")` for empty-after-trim) together with `testSearchAgentsWhitespaceNameDoesNotMatchAll` covers the edge case cleanly. I don't see anything else to flag here from the code side. This is still gated on a maintainer's `/accept` and merge call.

EricWittmann (commented):
Thanks @karanthengane23 — this is a really well-done contribution! Great issue report, clean fix, thorough tests, and you handled every round of feedback promptly. Merging now.

carlesarnal:
Reviewed after merge — nothing to flag. Clean, well-scoped fix with solid test coverage. The `SearchFilter.ofPartialName` extraction is a nice touch to prevent the three call sites from drifting. All commits have DCO sign-off and conventional messages. 👍

PR #8791, opened 2026-07-21
fix(ui): avoid duplicate label on boolean prompt variables

### LINKED ISSUE #8790: Boolean prompt variable name is rendered twice in the Test Prompt panel
## Description

In the Test Prompt panel on the Documentation tab, boolean variables show their name twice — once in the FormGroup label above the field, and once as the Checkbox's own inline label.

Other field types (text, number, textarea, select) use `aria-label` for the name, so only the FormGroup label is visible. Booleans are the only type where the label is duplicated visibly.

### Environment

### Steps to Reproduce

1. Create a PROMPT_TEMPLATE artifact with a boolean variable that has a description, e.g. `includeExamples: { type: "boolean", description: "Whether to include examples" }`
3. Look at the Test Prompt panel

**Expected:** the boolean field shows the name once, alongside its description.

**Actual:** the name renders twice, e.g.

```
includeExamples - Whether to include examples
[ ] includeExamples
```

This happens because `PromptTemplateTestPanel.tsx` wraps every field in a `FormGroup` with `label={name - description}` (line 182), and the `Checkbox` case also passes `label={name}` (line 129). Other field types pass `aria-label`, which isn't visible — the checkbox is the only one with a visible inline label, so only booleans duplicate.

Shubb07:
Fixes #8790

Boolean fields in the Test Prompt panel rendered their name twice — once as the FormGroup label above the field, and once as the Checkbox's own inline label. Other field types use `aria-label` for the name, so only the FormGroup label was visible for them; the checkbox is the only widget whose `label` prop is visible inline.

Moved the "name - description" text onto the Checkbox itself and skipped the FormGroup label for booleans. Other field types are unchanged.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -112,7 +112,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                         id={`var-${name}`}
                         isChecked={!!values[name]}
                         onChange={(_event, checked) => setValue(name, checked)}
-                        label={name}
+                        label={variable.description ? `${name} - ${variable.description}` : name}
                     />
                 );
             case "integer":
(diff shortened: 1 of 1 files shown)


Shubb07:
Reopening now that #8766 is merged — this second fix was auto-closed by the bot due to the 1-open-PR limit.

PR #8794, opened 2026-07-22
fix(kafkasql): emit ARTIFACT_VERSION_STATE_CHANGED event on version state update

### LINKED ISSUE #8763: bug: kafkasql does not emit an event when an artifact version's state changes
## Description

While reading through the storage event code I noticed that `KafkaSqlRegistryStorage.updateArtifactVersionState` submits its journal message but never fires an outbox event, unlike the comparable version and rule operations right next to it. As a result, changing a version's state (e.g. moving it to `DEPRECATED` or `DISABLED`) produces no event on the `registry-events` topic when running with kafkasql, even though the same operation on the `sql` variant emits an `ARTIFACT_VERSION_STATE_CHANGED` event.

The relevant methods:

- sql: `SqlVersionRepository.updateArtifactVersionState` fires the event:
  https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java#L320-L321
- kafkasql: `KafkaSqlRegistryStorage.updateArtifactVersionState` submits the journal message but never calls `outboxEvent.fire(...)`:
  https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java#L785-L791

For comparison, the neighbouring methods in the same class fire a `KafkaSqlOutboxEvent` after `coordinator.waitForResponse(...)`, for example `updateArtifactVersionMetaData` (L781), `createGlobalRule` (L803), `updateGlobalRule` (L826). `updateArtifactVersionState` fires nothing, and there's no comment explaining why.

The journal replay does reach `SqlVersionRepository.updateArtifactVersionState`, but that fires a `SqlOutboxEvent`, which is handled by `SqlEventsProcessor` (`@Observes SqlOutboxEvent`). The producer that publishes to the `registry-events` topic is `KafkaSqlEventsProcessor`, and it only observes `KafkaSqlOutboxEvent` (a separate class with no shared supertype), so it never sees the state-change event.

I also grepped the test tree and there's no test asserting `ARTIFACT_VERSION_STATE_CHANGED` for either variant, which is probably why this went unnoticed.

### Steps to Reproduce

1. Run the Registry with `APICURIO_STORAGE_KIND=kafkasql`.
2. Subscribe a consumer to the `registry-events` Kafka topic.
3. Create an artifact version, then change its state: `PUT /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{version}/state` with body `{"state": "DEPRECATED"}` (returns `204`).
4. No `ARTIFACT_VERSION_STATE_CHANGED` event arrives on the topic. Doing the same on the `sql` variant does produce one.

**Expected:** a version state change emits an `ARTIFACT_VERSION_STATE_CHANGED` event on `registry-events`, consistently with the `sql` variant and with the other mutating operations in kafkasql.

**Actual:** no event is emitted for a version state change under kafkasql.

Happy to open a PR with the fix and a test covering the event under kafkasql if this looks correct.

harsh04043:
On the kafkasql variant, changing an artifact version's state didn't emit an `ARTIFACT_VERSION_STATE_CHANGED` event to the `registry-events` topic, even though the sql variant does. `updateArtifactVersionState` submitted the journal message but never fired a `KafkaSqlOutboxEvent`, unlike the other version and rule operations in the same class. There was also no test covering this event on either variant, which is likely why it went unnoticed.

### Changes

- Fire `ArtifactVersionStateChanged` from `KafkaSqlRegistryStorage.updateArtifactVersionState` after `waitForResponse`, using the same factory as the sql variant so the payload matches.
- Read the previous state before submitting the message (the event carries `oldState`), the same way `SearchIndexEventDecorator` does.
- Skip the event on `dryRun`, since a dry run doesn't commit a state change (matches sql).
- Add a test asserting the event (including `oldState`/`newState`) that runs under both the sql and kafkasql variants.

### Test plan

- [x] `KafkaSqlEventsTest#updateArtifactVersionState` (kafkasql): failed before, passes after
- [x] `RegistryEventsTest#updateArtifactVersionState` (sql): passes, no regression
- [x] `mvnw checkstyle:check` clean

Closes #8763

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
@@ -9,6 +9,7 @@ import io.apicurio.registry.events.ArtifactRuleConfigured;
 import io.apicurio.registry.events.ArtifactVersionCreated;
 import io.apicurio.registry.events.ArtifactVersionDeleted;
 import io.apicurio.registry.events.ArtifactVersionMetadataUpdated;
+import io.apicurio.registry.events.ArtifactVersionStateChanged;
 import io.apicurio.registry.events.GlobalRuleConfigured;
 import io.apicurio.registry.events.GroupCreated;
 import io.apicurio.registry.events.GroupDeleted;
@@ -782,12 +783,23 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
                 .of(ArtifactVersionMetadataUpdated.of(groupId, artifactId, version, metaData)));
     }
 
+    /**
+     * @see io.apicurio.registry.storage.RegistryStorage#updateArtifactVersionState(java.lang.String,
+     *      java.lang.String, java.lang.String, io.apicurio.registry.types.VersionState, boolean)
+     */
     @Override
     public void updateArtifactVersionState(String groupId, String artifactId, String version,
             VersionState newState, boolean dryRun) {
+        // Capture the previous state before submitting the message, since the outbox event carries it.
+        VersionState oldState = getArtifactVersionState(groupId, artifactId, version);
         var message = new UpdateArtifactVersionState5Message(groupId, artifactId, version, newState, dryRun);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        // A dry run performs no committed state change, so no event should be emitted (matches SQL).
+        if (!dryRun) {
+            outboxEvent.fire(KafkaSqlOutboxEvent
+                    .of(ArtifactVersionStateChanged.of(groupId, artifactId, version, oldState, newState)));
+        }
     }
 
     /**
--- a/app/src/test/java/io/apicurio/registry/event/sql/RegistryEventsTest.java
+++ b/app/src/test/java/io/apicurio/registry/event/sql/RegistryEventsTest.java
@@ -10,6 +10,8 @@ import io.apicurio.registry.rest.client.models.EditableGroupMetaData;
 import io.apicurio.registry.rest.client.models.EditableVersionMetaData;
 import io.apicurio.registry.rest.client.models.GroupMetaData;
 import io.apicurio.registry.rest.client.models.Labels;
+import io.apicurio.registry.rest.client.models.VersionState;
+import io.apicurio.registry.rest.client.models.WrappedVersionState;
 import io.apicurio.registry.rules.validity.ValidityLevel;
 import io.apicurio.registry.storage.StorageEventType;
 import io.apicurio.registry.types.ArtifactType;
@@ -45,6 +47,7 @@ import static io.apicurio.registry.storage.StorageEventType.ARTIFACT_RULE_CONFIG
 import static io.apicurio.registry.storage.StorageEventType.ARTIFACT_VERSION_CREATED;
 import static io.apicurio.registry.storage.StorageEventType.ARTIFACT_VERSION_DELETED;
 import static io.apicurio.registry.storage.StorageEventType.ARTIFACT_VERSION_METADATA_UPDATED;
+import static io.apicurio.registry.storage.StorageEventType.ARTIFACT_VERSION_STATE_CHANGED;
 import static io.apicurio.registry.storage.StorageEventType.GLOBAL_RULE_CONFIGURED;
 import static io.apicurio.registry.storage.StorageEventType.GROUP_CREATED;
 import static io.apicurio.registry.storage.StorageEventType.GROUP_DELETED;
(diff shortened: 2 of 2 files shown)


app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -782,12 +783,23 @@ public void updateArtifactVersionMetaData(String groupId, String artifactId, Str
...
     }
 
+    /**
+     * @see io.apicurio.registry.storage.RegistryStorage#updateArtifactVersionState(java.lang.String,
+     *      java.lang.String, java.lang.String, io.apicurio.registry.types.VersionState, boolean)
+     */
     @Override
     public void updateArtifactVersionState(String groupId, String artifactId, String version,
             VersionState newState, boolean dryRun) {
+        // Capture the previous state before submitting the message, since the outbox event carries it.
```
paoloantinori, line 793: Worth noting: this read happens on the local SQL store before the journal message is processed. In the SQL variant (`SqlVersionRepository` line ~300), `oldState` is read with `SELECT ... FOR UPDATE` inside the same locked transaction as the write, guaranteeing accuracy. Here, if another node earlier submitted a state-change message for this same version that is consumed before this one, the captured `oldState` would be stale. The window is narrow (requires concurrent writes to the exact same version) and the pattern matches `SearchIndexEventDecorator`, so probably fine — just flagging the semantic difference from the SQL path.

app/src/test/java/io/apicurio/registry/event/sql/RegistryEventsTest.java
```diff
@@ -302,6 +305,46 @@ public void updateArtifactVersionMetadata() throws Exception {
...
+            }
+        }
+
+        Assertions.assertEquals(1, events.size());
+        Assertions.assertEquals(groupId, stateEvent.get("groupId").asText());
+        Assertions.assertEquals(artifactId, stateEvent.get("artifactId").asText());
+        Assertions.assertEquals(ARTIFACT_VERSION_STATE_CHANGED.name(),
+                stateEvent.get("eventType").asText());
+        Assertions.assertEquals("1", stateEvent.get("version").asText());
+        Assertions.assertEquals(VersionState.ENABLED.name(), stateEvent.get("oldState").asText());
```
paoloantinori, line 344: Consider also asserting that a dry-run state change does *not* produce an `ARTIFACT_VERSION_STATE_CHANGED` event — the `if (!dryRun)` guard in the production code is a deliberate behavioral choice (matching SQL), but nothing currently locks it in. A quick negative check would prevent a future regression there.

paoloantinori (commented):
Solid fix — the parity gap is real and well-scoped. I cross-checked against the SQL variant (`SqlVersionRepository.updateArtifactVersionState`, which fires `ArtifactVersionStateChanged` at the end of its transaction) and confirmed that `KafkaSqlRegistryStorage.updateArtifactVersionState` was the only version/rule mutation in that class that did not emit an outbox event. The dryRun guard is correct: in SQL the outbox row is written inside the same transaction that `setRollback(true)` discards, so SQL also effectively suppresses the event on dry runs.

A couple of observations, neither blocking:

- **oldState capture has a wider TOCTOU window than the SQL variant.** In `SqlVersionRepository`, `oldState` is read via `selectArtifactVersionStateForUpdate` with a row lock inside the same transaction as the write, so it is always the true pre-update state. Here, `oldState` is read from the local store before the journal message is submitted and applied. If another node submitted an earlier state-change message for the same version that gets consumed before this one, the emitted `oldState` would reflect the state at read time rather than at commit time. This is inherent to the kafkasql async-journal architecture (and the PR description notes `SearchIndexEventDecorator` follows the same pattern), so it is likely acceptable — but worth being aware of since `oldState` is semantically meaningful in the event payload.

- **Test coverage is good** (asserts `oldState`/`newState`, runs under both profiles via `KafkaSqlEventsTest extends RegistryEventsTest`). One minor gap: the PR adds an explicit `if (!dryRun)` guard, but no test asserts that a dry-run state change does *not* emit the event. A negative assertion would lock in that behavioral guarantee.

Overall this looks merge-ready to me once CI is green.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
harsh04043: Yeah, that's intentional. oldState here is a best-effort pre-read rather than a locked read like the SQL path, because in kafkasql the actual write happens asynchronously on the journal-apply side and there's no transaction to hook into from this thread. It follows the same pattern as SearchIndexEventDecorator. The window is narrow (concurrent state changes on the exact same version) and newState is always accurate. If we ever want SQL-grade accuracy on oldState, we could thread the pre-write state back through the journal response, but that's a bigger change to the message/coordinator contract, so I kept it out of this fix. Happy to do that as a follow-up if you think it's worth it.

app/src/test/java/io/apicurio/registry/event/sql/RegistryEventsTest.java
harsh04043: Good call, added updateArtifactVersionStateDryRun. It does a dry-run state change (to DISABLED) followed by a real one (to DEPRECATED), then asserts only the real change emits an event, with no leaked DISABLED/dry-run event. It runs under both variants via KafkaSqlEventsTest. One detail worth noting: the two variants suppress the dry-run event by different mechanisms. kafkasql relies on the new if (!dryRun) guard, while sql fires unconditionally but inside the transaction that setRollback(true) discards, so the outbox row never persists. Both green.

EricWittmann:
@harsh04043 — great work on this one. Clean fix, well-scoped, and the dry-run negative test is a
nice touch. Thanks for the contribution!

PR #8798, opened 2026-07-22
fix(search): normalize negative offset/limit instead of returning 500

### LINKED ISSUE #8611: Search API: negative offset or limit returns 500 and leaks SQL error instead of 400

Me-Priyank:
Fixes #8611

### What

Normalizes out-of-range `offset`/`limit` on the artifact, group, and version search endpoints instead of letting invalid values reach storage and produce a 500.

### Why

Negative or oversized pagination parameters previously surfaced as HTTP 500 errors. Invalid input should be handled gracefully rather than failing the request.

### How

- `offset` is clamped to `[0, Integer.MAX_VALUE]` (negative → 0) so it never reaches storage as an invalid query.
- `limit` handling: a missing `limit` uses the existing default of 20; a negative `limit` normalizes to 1; `limit=0` is honored as an explicit empty page; and values above `MAX_LIMIT` (1000) are clamped to 1000 to bound result size. The 1000 cap matches the existing `GET_GROUPS_LIMIT` used elsewhere in the v3 API.
- Added test coverage for negative, zero, `limit=1`, oversized, and both-params-invalid cases across the artifacts, groups, and versions search endpoints.

> Note: this is the same change as #8683, which I closed so I could open the config-docs build fix (#8797) under the one-open-PR limit. Rebased cleanly onto current `main`.


Me-Priyank:
@carlesarnal @EricWittmann could you please `/accept` so i can `/ready` and let the suite run?  :))

app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
```diff
@@ -476,11 +478,21 @@ public ArtifactSearchResults searchContracts(String status, String ownerTeam,
...
+    // Clamp the offset to [0, Integer.MAX_VALUE] so it never reaches storage as an invalid SQL query (500). See #8611.
+    private static int normalizeOffset(BigInteger offset) {
+        return offset.max(BigInteger.ZERO).min(MAX_INT_VALUE).intValue();
+    }
+
+    // Negative limit -> 1 (limit=0 keeps its empty-page semantics); cap at MAX_LIMIT to bound result size. See #8611.
+    private static int normalizeLimit(BigInteger limit) {
+        return limit.signum() < 0 ? 1 : limit.min(MAX_LIMIT).intValue();
+    }
+
```
paoloantinori, line 495: This silent clamp to `MAX_LIMIT` (1000) is a behavior change not mentioned in the PR body. A client sending `limit=5000` gets 1000 items back with no signal that truncation happened, which can break pagination assumptions. Consider returning `400` for oversized limits (treating it as invalid input, like the negatives were originally), or at minimum document the cap and surface the effective page size in the response.

paoloantinori, line 495: Comment says "Negative limit -> 1", but the PR body says negatives normalize to `20` (the default). The tests assert `size()==1` for `limit=-1`, so the code is what's actually shipped — but the body description should be corrected to avoid confusing reviewers/users about the effective default.

paoloantinori (commented):
Clean, focused fix — normalizing invalid pagination before it reaches storage is the right layer to absorb this, and the per-endpoint test coverage (negative/zero/oversized/both-invalid) is thorough. A few observations worth considering before merge:

- **Silent upper cap on `limit` is a behavior change the description doesn't advertise.** `normalizeLimit` clamps any `limit > 1000` down to `MAX_LIMIT` and returns the truncated page without signaling it. The PR body only documents the negative/zero cases. A client requesting `limit=5000` will silently receive 1000 items and could page incorrectly assuming the page size it asked for. Worth deciding whether oversized limits should be a `400 Bad Request` (consistent with how the API treats other invalid input) rather than a silent clamp — or, if clamping is intentional, documenting it in the body and ideally reflecting the effective page size in the response. Related: the cap constant lives only in this class; if the storage layer or other resources have their own max-page-size config, this hard-coded `1000` could drift from it.
- **Doc/comment mismatch on the negative-limit default.** The PR body states negatives normalize to `20` ("sensible defaults ... `limit` → 20"), but the code comment and the tests normalize a negative `limit` to `1` (`limit.signum() < 0 ? 1 : ...`, and `SearchArtifactsTest` asserts `artifacts.size() == 1` for `limit=-1`). Either the body or the code is wrong about the intended default — worth reconciling.
- **`limit=0` semantics differ between the description and the new negative-handling branch.** The body says `limit=0` preserves empty-page behavior, which the code does (`signum() < 0` is false at 0, so it flows to `limit.min(MAX_LIMIT)` = 0). That's consistent, but the asymmetry (0 → empty page, -1 → 1 item) is slightly surprising for clients; a brief note in the response/docs on why 0 is special-cased would help.
- **`normalizeOffset`/`normalizeLimit` are private to `SearchResourceImpl`.** If the same validation is needed on other search resources (e.g. `SearchContractsResourceImpl` if one exists, or future endpoints), consider extracting to a small shared utility now to avoid copy-paste later — but not blocking.

app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
Me-Priyank: Good catch @paoloantinori ... I kept this as a clamp rather than a 400 deliberately coz it follows the same "normalize invalid pagination instead of erroring" approach used for the negatives, so erroring only on oversized would make the API inconsistent. The 1000 cap also matches the existing `GET_GROUPS_LIMIT` in `GroupsResourceImpl`, so it is not a new isolated constant. I have documented the cap in the PR body. If you and @carlesarnal would prefer a stricter contract (400 on oversized, or echoing the effective page size in the response), I am glad to switch. Since the cap came from Carles's earlier review, I would rather align with both of you before changing the behavior.

Me-Priyank: You are right @paoloantinori, that was a wording error in the PR body, not the code. Negative `limit` normalizes to 1 (as the code, comment, and tests all do); the 20 is the default used only when `limit` is omitted. I have corrected the description to state this accurately.

Me-Priyank:
> Clean, focused fix — normalizing invalid pagination before it reaches storage is the right layer to absorb this, and the per-endpoint test coverage (negative/zero/oversized/both-invalid) is thorough. A few observations worth considering before merge:
> 
> * **Silent upper cap on `limit` is a behavior change the description doesn't advertise.** `normalizeLimit` clamps any `limit > 1000` down to `MAX_LIMIT` and returns the truncated page without signaling it. The PR body only documents the negative/zero cases. A client requesting `limit=5000` will silently receive 1000 items and could page incorrectly assuming the page size it asked for. Worth deciding whether oversized limits should be a `400 Bad Request` (consistent with how the API treats other invalid input) rather than a silent clamp — or, if clamping is intentional, documenting it in the body and ideally reflecting the effective page size in the response. Related: the cap constant lives only in this class; if the storage layer or other resources have their own max-page-size config, this hard-coded `1000` could drift from it.
> * **Doc/comment mismatch on the negative-limit default.** The PR body states negatives normalize to `20` ("sensible defaults ... `limit` → 20"), but the code comment and the tests normalize a negative `limit` to `1` (`limit.signum() < 0 ? 1 : ...`, and `SearchArtifactsTest` asserts `artifacts.size() == 1` for `limit=-1`). Either the body or the code is wrong about the intended default — worth reconciling.
> * **`limit=0` semantics differ between the description and the new negative-handling branch.** The body says `limit=0` preserves empty-page behavior, which the code does (`signum() < 0` is false at 0, so it flows to `limit.min(MAX_LIMIT)` = 0). That's consistent, but the asymmetry (0 → empty page, -1 → 1 item) is slightly surprising for clients; a brief note in the response/docs on why 0 is special-cased would help.
> * **`normalizeOffset`/`normalizeLimit` are private to `SearchResourceImpl`.** If the same validation is needed on other search resources (e.g. `SearchContractsResourceImpl` if one exists, or future endpoints), consider extracting to a small shared utility now to avoid copy-paste later — but not blocking.

Thanks for the thorough review @paoloantinori, replied inline on the oversized-limit cap and the negative-limit wording. On the other two:

- limit=0: it is treated as an explicit request for an empty page, so only negative values are normalized. Documented in the code comment and now in the PR body.
- Shared helper: `normalizeOffset`/`normalizeLimit` are currently used only across this resource's own search methods, so I kept them private to avoid a premature abstraction. Happy to promote them to a shared utility the moment a second resource needs the same normalization.

paoloantinori:
Thanks for the reply. Since the head hasn't changed, I'll just reaffirm the one point I'd still flag as worth a moment before merge: the silent upper clamp on `limit` (`> 1000` → 1000, no signal to the caller) is a behavior change that isn't called out in the PR description or the `@Info`/OpenAPI notes. Normalizing negatives to avoid the 500 is clearly the right call and well tested; the clamp is the only part that could surprise a caller who legitimately pages with `limit > 1000`.

That's a "worth documenting" note, not a blocker — if the intent is to align with the existing `GET_GROUPS_LIMIT` cap and you'd rather not 400 on oversized values, that's a defensible choice as long as it's discoverable. Deferring to maintainer judgment here.

Me-Priyank:
> Thanks for the reply. Since the head hasn't changed, I'll just reaffirm the one point I'd still flag as worth a moment before merge: the silent upper clamp on `limit` (`> 1000` → 1000, no signal to the caller) is a behavior change that isn't called out in the PR description or the `@Info`/OpenAPI notes. Normalizing negatives to avoid the 500 is clearly the right call and well tested; the clamp is the only part that could surprise a caller who legitimately pages with `limit > 1000`.
> 
> That's a "worth documenting" note, not a blocker — if the intent is to align with the existing `GET_GROUPS_LIMIT` cap and you'd rather not 400 on oversized values, that's a defensible choice as long as it's discoverable. Deferring to maintainer judgment here.

Thanks @paoloantinori . The cap is now documented in the PR description under "How": negative `limit` normalizes to 1, `limit=0` is an explicit empty page, and values above `MAX_LIMIT` (1000) are clamped to 1000 (matching the existing `GET_GROUPS_LIMIT`). That closes the discoverability gap you flagged. PTAL :))

paoloantinori:
@Me-Priyank The head (`b05add0`) is unchanged since my last pass, so my earlier note still stands: normalizing negatives to avoid the 500 is the right call and well tested — the only thing I'd still flag before merge is the silent upper clamp on `limit` (`> 1000` → 1000, no signal to the caller), which could surprise a caller legitimately paging with a larger page size. Not a blocker, just worth a conscious call from whoever merges (clamp silently vs. return 400). Nothing else to add on my end.

Me-Priyank:
@EricWittmann @carlesarnal could you PTAL ?

EricWittmann:
Thanks for the contribution @Me-Priyank! Clean fix with thorough test coverage.

carlesarnal:
Did a post-merge review of this PR — everything looks solid. Clean, focused fix with thorough test coverage across all search endpoints. A few things I noted:

- **Normalization logic is correct**: `normalizeOffset` clamps to `[0, Integer.MAX_VALUE]` preventing BigInteger overflow on `.intValue()`; `normalizeLimit` handles negative/zero/oversized cases consistently.
- **All 6 search endpoints covered**: artifacts, artifactsByContent, groups, versions, versionsByContent, and contracts all use the new helpers.
- **Test coverage is thorough**: negative, zero, boundary, both-invalid, and oversized cases are tested with specific assertions on both `count` and page size.
- **DCO, conventional commit, code style**: all good.

Nothing to flag — nice contribution @Me-Priyank.

PR #8803, opened 2026-07-22
feat(rest): add well-known endpoints to OpenAPI spec

### LINKED ISSUE #8799: feat: add well-known endpoints to OpenAPI spec
## Summary

Add the existing well-known endpoints (A2A agent discovery, MCP tool discovery, JSON schemas) to the main `openapi.json` spec so all SDKs auto-generate client code for them.

Follow-up from #8465.

## Endpoints to add

These 10 endpoints already exist in `WellKnownResource.java` but are not documented in the OpenAPI spec:

| Method | Path | Returns |
|--------|------|---------|
| GET | `/.well-known/agent.json` | AgentCard |
| GET | `/.well-known/a2a` | AgentCard |
| GET | `/.well-known/agents/public` | AgentSearchResults |
| GET | `/.well-known/agents/entitled` | AgentSearchResults |
| POST | `/.well-known/agents/search` | AgentSearchResults |
| GET | `/.well-known/agents/{groupId}/{artifactId}` | raw JSON |
| GET | `/.well-known/agents` | AgentSearchResults |
| GET | `/.well-known/mcp-tools/{groupId}/{artifactId}` | raw JSON |
| GET | `/.well-known/mcp-tools` | McpToolSearchResults |
| GET | `/.well-known/schemas/{schemaType}/{version}` | raw JSON |

## Schemas to add

`AgentCard`, `AgentProvider`, `AgentCapabilities`, `AgentExtension`, `AgentSkill`, `AgentInterface`, `SecurityScheme`, `SecurityRequirement`, `AgentCardSignature`, `AgentSearchResult`, `AgentSearchResults`, `AgentSearchRequest`, `AgentSearchFilters`, `McpToolSearchResult`, `McpToolSearchResults`

## Notes

- The `{type}` path parameter in `/.well-known/schemas/{type}/{version}` should be renamed to `{schemaType}` to avoid Go reserved keyword conflict
- After updating openapi.json, the Go SDK must be regenerated (`cd go-sdk && make generate`)
- A `WellKnown` tag should be added to group these endpoints

carlesarnal:
## Summary

Closes #8799

Add A2A agent, MCP tool, and JSON schema well-known endpoints to the main `openapi.json` so all SDKs auto-generate client code for them.

### What changed

- **15 component schemas** added to `openapi.json` (`AgentCard`, `AgentSearchResults`, `McpToolSearchResults`, `AgentSecurityScheme`, etc.)
- **11 path definitions** under `/well-known/` with `WellKnown` tag
- **`{type}` → `{schemaType}`** in the schemas endpoint path parameter (avoids Go reserved keyword)
- **`V3WellKnownResource`** delegate implements the generated `WellResource` interface, serving endpoints at `/apis/registry/v3/well-known/` so SDK clients work with the standard v3 base URL
- **Hand-written DTOs deleted** — code now uses codegen-generated beans from `io.apicurio.registry.rest.v3.beans`
- **Go SDK regenerated** with `make generate` + `make format`

### Endpoints served at both paths

| Protocol-compliant (existing) | SDK-friendly (new) |
|-------------------------------|-------------------|
| `/.well-known/agent.json` | `/apis/registry/v3/well-known/agent.json` |
| `/.well-known/agents` | `/apis/registry/v3/well-known/agents` |
| `/.well-known/agents/public` | `/apis/registry/v3/well-known/agents/public` |
| `/.well-known/agents/entitled` | `/apis/registry/v3/well-known/agents/entitled` |
| `POST /.well-known/agents/search` | `POST /apis/registry/v3/well-known/agents/search` |
| `/.well-known/agents/{g}/{a}` | `/apis/registry/v3/well-known/agents/{g}/{a}` |
| `/.well-known/mcp-tools` | `/apis/registry/v3/well-known/mcp-tools` |
| `/.well-known/mcp-tools/{g}/{a}` | `/apis/registry/v3/well-known/mcp-tools/{g}/{a}` |
| `/.well-known/schemas/{schemaType}/{v}` | `/apis/registry/v3/well-known/schemas/{schemaType}/{v}` |

## Test plan

- [x] `WellKnownResourceTest` — 29 tests pass (both `/.well-known/` and v3 paths)
- [x] `ExperimentalFeatureGateTest` — 4/4 pass
- [x] Checkstyle — 0 violations
- [x] Go SDK compiles (`go build ./...`)
- [ ] CI


app/src/main/java/io/apicurio/registry/rest/wellknown/V3WellKnownResource.java
```diff
@@ -0,0 +1,143 @@
...
+public class V3WellKnownResource {
+
+    @Inject
+    WellKnownResourceImpl delegate;
+
+    @GET
+    @Path("/agent.json")
+    @Produces(MediaType.APPLICATION_JSON)
+    public AgentCard getAgentCard() {
+        return delegate.getAgentCard();
```
paoloantinori, line 28: This class mirrors the full surface of `WellKnownResource` by hand. If a new well-known endpoint lands later, both classes need the same signature or they silently diverge (different `@DefaultValue`, `@QueryParam`, or content type). A `// keep in sync with WellKnownResource` cross-reference here, or a shared JAX-RS interface both implement, would make the coupling explicit.

paoloantinori (commented):
The design rationale (serving well-known endpoints at both `/.well-known/` for A2A compliance and `/apis/registry/v3/well-known/` for SDK base-URL clients, via a single `WellKnownResourceImpl` delegate) is sound, and the V3 path delegate is a clean thin wrapper. Most of the 33k-line diff is Go SDK regeneration noise; my comments are limited to the hand-written Java and the API surface, which is what I could actually read.

- **`{type}` → `{schemaType}` path-param rename is an API surface change worth a compatibility call-out.** The body says the rename on the schemas endpoint avoids a Go reserved-keyword conflict, but any client already calling the endpoint with `{type}` would break on upgrade. If this endpoint is still behind the experimental flag (`apicurio.mcp-tools.enabled`, which the test profile change here enables), the blast radius is limited and this is fine — but worth confirming explicitly. If it's already GA, consider keeping `{type}` as an accepted alias rather than a hard rename, since the Go-codegen problem can also be solved with a Kiota naming override.
- **Two parallel resource classes (`WellKnownResource` + `V3WellKnownResource`) must stay in lockstep by hand.** Every endpoint signature is duplicated across the two `@Path` classes; adding a new well-known endpoint means editing both, and a drift in annotations (`@DefaultValue`, `@QueryParam` names, content types) between them would be a silent bug. Not blocking for an initial add, but worth a note/`// keep in sync with` link or a future consolidation (e.g. shared interface). For reference, the canonical implementation both delegate to is [WellKnownResourceImpl.java](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java).
- **Unable to assess the `openapi.json` regeneration (~3000 line churn) and the bulk of the Go SDK diffs from the truncated diff.** The 11 new `/well-known/` path definitions and 15 new schemas are the core deliverable but couldn't be read in full here; flagging that this review is necessarily limited to the Java resource classes.

app/src/main/java/io/apicurio/registry/rest/wellknown/V3WellKnownResource.java
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Exposes well-known endpoints under the {@code /apis/registry/v3/well-known} path so that SDK
+ * clients (Go, Java, Python, TypeScript) generated from the OpenAPI spec can reach them using
+ * the standard v3 base URL.
+ *
+ * <p>All calls delegate to {@link WellKnownResourceImpl}, which also serves the canonical
+ * {@code /.well-known/} paths for A2A protocol compliance.
+ */
+@ApplicationScoped
+@Interceptors({ResponseErrorLivenessCheck.class, ResponseTimeoutReadinessCheck.class})
```
paoloantinori: Both this class and `WellKnownResourceImpl` carry `@Interceptors(...)` + `@Logged`, so a V3-path call runs the liveness/readiness/log stack twice (once here, once on the delegate proxy). The direct `/.well-known/` path runs it once. Since this class only delegates, consider removing these class-level annotations here to keep both paths symmetric and avoid double-logging / double-counted health checks.

paoloantinori (commented):
Solid structure — `V3WellKnownResource` cleanly reuses `WellKnownResourceImpl`, and the bean-package consolidation (`a2a/rest/beans` → `rest/v3/beans`) and the `{type}` → `{schemaType}` path-param rename (to dodge the Go reserved keyword) are both sensible. My read focused on the hand-written Java (the regenerated `openapi.json` and Go SDK are codegen output).

On authz: it appears correctly preserved. The injected `delegate` is an `@ApplicationScoped` CDI client proxy, so the `@Authorized` bindings on the impl methods should still fire for V3-path calls — no authorization gap introduced by the extra indirection.

One thing worth a look: `V3WellKnownResource` redeclares the same class-level `@Interceptors({ResponseErrorLivenessCheck, ResponseTimeoutReadinessCheck})` and `@Logged` as its delegate. Because the delegate is a proxy, on the V3 path both interceptor stacks fire, so each request is double-logged and the liveness/readiness checks are nested twice — whereas the direct `/.well-known/` path triggers them once. Since `V3WellKnownResource` is a pure pass-through, dropping those annotations from it would keep both paths symmetric and avoid the double-counting.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
aclfe, line 1: One  question I wanted to raise, more as a 'what do you think' question than a blocker 

After this PR the well-known endpoints are defined twice at the routing layer: the hand-written WellKnownResource interface serving `/.well-known/...` for protocol compliance, and the codegen WellResource interface serving
 `/v3/well-known/...` for the SDKs. V3WellKnownResource just forwards every call to WellKnownResourceImpl, so the logic isn't duplicated, but the two interfaces and impls have to be kept in sync by hand. 

If someone adds a well-known endpoint later, they have to remember to touch both, and the two sides don't even use the same param types (Integer on the hand-written side vs BigInteger on the codegen side, which is why we get the offset.intValue() / limit.intValue() bridging). We also end up with two public URLs for the same resource, so it's a bit ambiguous which one is canonical.

  I get that a lot of this is inherent to the goal (same feature at two path roots for two different audiences), so I'm not saying the current split is wrong. It works and it's not hacky. My worry is just that this is the part most likely to drift over time.

  The alternative I had in mind: pull the actual logic out of WellKnownResourceImpl into a plain service (no JAX-RS annotations), and make both `WellKnownResource` and `V3WellKnownResource` thin adapters over it. That way neither path is "primary", the symmetry is explicit, and adding an endpoint is obviously a two-adapter change against one service. Maybe this could be a follow up?

  So the question is do maintainers want to keep the "one impl is primary, the other delegates to it" shape and just live with keeping them in sync, or is it worth extracting the shared service now while there are only a handful of endpoints? It could go either way I'm not too sure either

aclfe (commented):
^ typo in my last comment fixed

paoloantinori:
Good point to talk through. The dual definition is deliberate and I think defensible, but the drift risk aclfe raises is the real thing to watch.

The two paths serve different consumers and can't collapse into one:
- `/.well-known/...` is required by RFC 8615 / the A2A spec — agents discover the agent card at that exact path, and it can't be relocated under `/apis/registry/v3/`.
- `/apis/registry/v3/well-known/...` exists so SDK clients configured with the v3 base URL can reach the same data without a second HTTP client or base-path override.

Since `V3WellKnownResource` is a thin delegate to `WellKnownResourceImpl`, there's no duplicated *logic* — only duplicated *route declarations*. The practical maintenance cost is: adding or renaming an endpoint on `WellKnownResource` must be mirrored on `V3WellKnownResource`, and a divergence today would fail silently.

The mitigation I'd suggest (not a blocker for this PR, but worth a follow-up): a contract test that exercises both paths for each endpoint and asserts identical responses, so route drift fails CI instead of shipping.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
aclfe: @paoloantinori thoughts?

carlesarnal: This is not longer a problem. I have removed the other impl, it isn't needed.

jsenko (commented):
## Review: SDK path mismatch after V3WellKnownResource removal

The latest squash removed `V3WellKnownResource`, which was the handler serving well-known endpoints at `/apis/registry/v3/well-known/...`. However, the `openapi.json` still defines paths under `/well-known/` (e.g. `/well-known/agent.json`, `/well-known/agents`, etc.).

### The problem

SDK clients are configured with the v3 base URL:
```
http://host:port/apis/registry/v3
```

The generated SDK code (confirmed in Go SDK — `{+baseurl}/well-known`) constructs URLs like:
```
GET /apis/registry/v3/well-known/agent.json
GET /apis/registry/v3/well-known/agents
```

But the server only serves these at the root-level `/.well-known/` path (via `WellKnownResourceImpl` with `@Path("/.well-known")`). There is no handler at `/apis/registry/v3/well-known/...` anymore, so **SDK-generated clients will get 404 errors** on all well-known endpoints.

### Evidence

- Go SDK test base URL: `http://{host}:{port}/apis/registry/v3` ([go-sdk/pkg/tests/client_test.go:25](https://github.com/Apicurio/apicurio-registry/blob/69455eacfa931d6eacdb7d21e7c859a38afa4df9/go-sdk/pkg/tests/client_test.go#L25))
- Go SDK well-known builder: `{+baseurl}/well-known` ([well_known_request_builder.go:44](https://github.com/Apicurio/apicurio-registry/blob/69455eacfa931d6eacdb7d21e7c859a38afa4df9/go-sdk/pkg/registryclient-v3/wellknown/well_known_request_builder.go#L44))
- Server handler: `@Path("/.well-known")` on `WellKnownResource` interface

### Suggested fix

Either:
1. **Bring back `V3WellKnownResource`** as a thin delegate to `WellKnownResourceImpl` (the original approach), addressing paoloantinori's double-interceptor concern by removing `@Interceptors`/`@Logged` from the delegate class
2. **Remove the `/well-known/` paths from `openapi.json`** if SDK access is no longer a goal — but this would defeat the PR's stated purpose

PR #8817, opened 2026-07-22
fix(ui): avoid duplicate label on boolean prompt variables

### LINKED ISSUE #8790: Boolean prompt variable name is rendered twice in the Test Prompt panel

Shubb07:
Fixes #8790

Boolean fields in the Test Prompt panel rendered their name twice — once as the FormGroup label above the field, and once as the Checkbox's own inline label. Other field types use `aria-label` for the name, so only the FormGroup label was visible for them; the checkbox is the only widget whose `label` prop is visible inline.

Moved the "name - description" text onto the Checkbox itself and skipped the FormGroup label for booleans. Other field types are unchanged.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -112,7 +112,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                         id={`var-${name}`}
                         isChecked={!!values[name]}
                         onChange={(_event, checked) => setValue(name, checked)}
-                        label={name}
+                        label={variable.description ? `${name} - ${variable.description}` : name}
                     />
                 );
             case "integer":
@@ -162,16 +162,19 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
             </CardHeader>
             <CardBody>
                 <Form className="test-panel-form">
-                    {variablesList.map(({ name, variable }, index) => (
-                        <FormGroup
-                            key={index}
-                            label={variable.description ? `${name} - ${variable.description}` : name}
-                            isRequired={variable.required}
-                            fieldId={`var-${name}`}
-                        >
-                            {renderField(name, variable)}
-                        </FormGroup>
-                    ))}
+                    {variablesList.map(({ name, variable }, index) => {
+                        const isBoolean = (variable.type || "string").toLowerCase() === "boolean";
+                        return (
+                            <FormGroup
+                                key={index}
+                                label={isBoolean ? undefined : (variable.description ? `${name} - ${variable.description}` : name)}
+                                isRequired={variable.required}
+                                fieldId={`var-${name}`}
+                            >
+                                {renderField(name, variable)}
+                            </FormGroup>
+                        );
+                    })}
                     <ActionGroup>
                         <Button
                             variant="primary"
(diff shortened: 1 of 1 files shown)


ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -162,16 +162,19 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
...
-                            label={variable.description ? `${name} - ${variable.description}` : name}
-                            isRequired={variable.required}
-                            fieldId={`var-${name}`}
-                        >
-                            {renderField(name, variable)}
-                        </FormGroup>
-                    ))}
+                    {variablesList.map(({ name, variable }, index) => {
+                        const isBoolean = (variable.type || "string").toLowerCase() === "boolean";
+                        return (
```
paoloantinori, line 181: Setting `label={undefined}` here keeps `isRequired={variable.required}` on the FormGroup. In PF5 the required asterisk is attached to the FormGroup label, so a required boolean may render with no visible required indicator. Worth double-checking whether required booleans should surface the marker on the Checkbox label instead.

paoloantinori (commented):
Agreed with the diagnosis — the Checkbox's visible `label` was duplicating the FormGroup label for booleans. Consolidating onto the Checkbox and dropping the FormGroup label for booleans is the right call. One thing worth verifying before this lands:

- **Required indicator for boolean fields.** `isRequired` is still passed to `FormGroup` even when `label` is set to `undefined`. In PatternFly the required asterisk is rendered as part of the FormGroup label, so with no label a required boolean field may lose its visual "required" cue, and the Checkbox's own `label` doesn't carry an `isRequired` prop. If required booleans are a valid case, worth confirming the asterisk still renders somewhere (or moving the required marker onto the Checkbox label).

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
Shubb07: Good catch — fixing this by moving the required marker onto the Checkbox label as a JSX fragment. pushing shortly

Shubb07:
Pushed the fix — required booleans now carry the red asterisk on the Checkbox label itself, matching the FormGroup convention. verified locally with a required boolean field.

paoloantinori:
Thanks @Shubb07 — the fix looks good. Putting the red asterisk directly on the Checkbox `label` for required booleans is the right call now that the FormGroup label is `undefined` for the boolean case; it keeps the required indicator consistent with the other field types. LGTM from my side.

Issue #7485, opened 2026-03-04
Operator CI tests take 93+ minutes due to sequential per-class infrastructure setup/teardown


carlesarnal:
## Problem

The **"Build and Test Operator"** CI job takes **101 minutes** total, with **93 minutes** spent in the "Run remote and OLM v1 tests on Minikube" step alone (see [example run](https://github.com/Apicurio/apicurio-registry/actions/runs/22662021653/job/65684335838?pr=7482)). This creates long feedback loops for PRs.

### 1. Per-test-class infrastructure overhead (~2 min × 23 classes = ~46 min of pure overhead)

Every integration test class extends `ITBase`, whose `@BeforeAll`/`@AfterAll` performs a full infrastructure cycle:
- Creates a new Kubernetes namespace
- Deploys the operator (applies install YAML, waits for deployment)
- Waits for operator pod readiness with a **30-second stability check** (`await().during(SHORT_DURATION)` in `waitOnOperatorPodReady()`)
- After tests: deletes CRs, waits for cleanup, deletes operator resources, deletes namespace

For 23 integration test classes, this means ~23 full setup/teardown cycles. Even a simple test like `PodDisruptionBudgetITTest` (1 test method) takes **140 seconds** — the vast majority is infrastructure setup and teardown.

### 2. All tests run sequentially in a single JVM (no parallelism)

The Maven Surefire plugin has no `forkCount`, `parallel`, or `threadCount` configuration in `operator/controller/pom.xml`. All 23 integration test classes and 67 test methods run sequentially in a single process.

### 3. Kafka/Strimzi tests are the slowest (~24 min for 4 tests)

The 4 Kafka-related tests (`KafkaSqlITTest`, `KafkaSqlTLSITTest`, `KafkaSqlOAuthITTest`, `KafkaSqlAccessITTest`) each take **5-6 minutes** because each one:
- Downloads and deploys the Strimzi cluster operator (from GitHub releases URL)
- Creates a Kafka cluster
- Waits for Kafka readiness
- Deploys the registry with KafkaSQL storage

### 4. 30-second mandatory stability check per test class

`waitOnOperatorPodReady()` uses `.during(SHORT_DURATION)` (30 seconds) which forces the test to **continuously verify** the operator pod is stable for 30 seconds before proceeding. This adds 30s of pure wait per test class = ~11.5 min total.

## Per-Test-Class Breakdown (from CI logs)

| Test Class | Duration | Tests |
|---|---|---|
[...]

PR #8818, opened 2026-07-22
test(operator): add the infra extension and share one operator (Phase 1)

aclfe:
Phase 1 of #8495 introduces a shared cluster-wide operator and a JUnit 5 OperatorInfraExtension that provides test infrastructure through parameter injection rather than ITBase inheritance.

**Do not merge before #8686**. This PR extends its shared cluster-wide installation pattern to the Apicurio operator and I will re-target to main after #8686 merges. 

**How to Review: This is a large PR. I split the work into commits that each isolate one step, so reviewing the organized commit-by-commit (rather than the full diff at once) should make the changes easier to follow** 

---
**_Why_**:

Each of the approximately 28 operator IT classes redundantly installs and removes its own operator, causing cluster-scoped binding conflicts, preventing parallel execution, and contributing significantly to the 90+ minute CI runtime (see #7485) 

---

**_Changes:_**:

  **Shared operator** (`OperatorClusterWideInstaller`, `SharedOperatorLifecycle`)
  - Installs the operator once per cluster, watching all namespaces, into a dedicated long-lived namespace
    (`test.operator.operator-namespace`) that test teardown never deletes.
  - No operator configuration is needed for the cluster-wide watch: `App#start` already calls
    `watchingAllNamespaces()` whenever `apicurio.operator.watched-namespaces` is blank, which was an open
    question in the issue.
  - The install guard uses cluster state instead of a JVM flag to skip redundant installations across test groups reusing the same cluster, following the Strimzi #8686 pattern.
  - Local mode starts one in-process operator per JVM instead of one per class.

  **Operator-mutating tests get their own operator** (`@DedicatedOperator`)
  - Three tests mutate the operator Deployment, so they keep dedicated @DedicatedOperator installs in a separate operator-mutation CI job (both workflows, #7494 pattern), tearing down only their own suffixed ClusterRoleBinding.

  **Infrastructure extension** (`OperatorInfraExtension`)
  - OperatorInfraExtension injects the namespace, KubernetesClient, RegistryAssertions and the manager classes as test parameters, so a class no longer has to extend ITBase.
  - The operator and Strimzi are shared, lazy, and root-scoped; the namespace and its managers are per-class, torn down after, with CRs cleared between test methods.
  - Both entry points share one implementation (SharedOperatorLifecycle, RegistryAssertions, ITBase.deleteRegistryCRs), so a mixed run still has exactly one operator. DisableUIITTest is the worked example. 

---

| Issue item | Status |
  |---|---|
  | Root-scoped operator lifecycle (lazy start, closeable cleanup) | Done: `SharedOperatorLifecycle` + `SharedOperator` in JUnit's root store |
  | Per-class namespace creation/deletion | Done: `ClassInfra` |
  | `ParameterResolver` for `KubernetesClient`, namespace, managers | Done: `INJECTABLE` map in `OperatorInfraExtension` |
  | Root-scoped lazy Strimzi install + cleanup | Done |
  | `@NeedsStrimzi` annotation | Done |
  | CRD install once | Done: local: `createCRDs`; remote: install file |
  | Per-test CR cleanup between methods | Done: `ITBase.deleteRegistryCRs(...)`, shared by both paths |
  | RestAssured timeout config on the migrated path | Done: `ITBase.configureRestAssured()` made static, called from the extension|
  | `@DedicatedOperator` teardown scope | Done: deletes only its own `ClusterRoleBinding`, not the shared CRD/ClusterRole|
  | Install-file apply retry on transient errors | Done: `await().ignoreExceptions()`|
  | `junit-platform.properties` parallel config | **Deliberately omitted**: Rationale in `operator/README.md` |

  **Not included: `junit-platform.properties`**
  Deliberately left out.  https://github.com/Apicurio/apicurio-registry/issues/8495#issuecomment-4997794834 Two `@QuarkusTest` classes running concurrently race into the unsynchronized QuarkusTestExtension.ensureStarted(), and the second boot dies with Already a codec registered with name quarkus_default_local_codec, confirmed locally. Adding the file would also shadow Quarkus' own copy in quarkus-junit-config, losing QuarkusClassOrderer and extension autodetection. CI-level sharding remains the only safe way to parallelize these tests (`.github/workflows/README.md`); full rationale in `operator/README.md`

  ## Test plan

  - [x] `OperatorClusterWideInstallerTest`: unit-tests the install-file transform (ClusterRoleBinding
        subject rebinding, minikube image-pull-policy override, operator selector labels) against a
        structural fixture, no cluster needed.
  - [x] `OperatorInfraExtensionTest`: pins the extension's parameter-resolution contract and the
        `@NeedsStrimzi`/`@DedicatedOperator` annotations on the classes that need them.
  - [x] `./mvnw test -pl operator/controller -Dtest=OperatorClusterWideInstallerTest,OperatorInfraExtensionTest,StrimziClusterWideInstallerTest`: 10/10 passing.
  - [x] `./mvnw test-compile -pl operator/controller -am` compiles cleanly.
  - [x] `./mvnw checkstyle:check -pl operator/controller` passes.

### DIFF
--- a/operator/controller/src/test/java/io/apicurio/registry/operator/it/SharedOperatorLifecycle.java
+++ b/operator/controller/src/test/java/io/apicurio/registry/operator/it/SharedOperatorLifecycle.java
@@ -1,0 +1,74 @@
+package io.apicurio.registry.operator.it;
+
+import io.apicurio.registry.operator.App;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClientBuilder;
+import jakarta.enterprise.inject.spi.CDI;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+
+import static io.apicurio.registry.operator.it.ITBase.CRD_FILE;
+import static io.apicurio.registry.operator.it.ITBase.SHORT_DURATION;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
+final class SharedOperatorLifecycle {
+
+    private static final Logger log = LoggerFactory.getLogger(SharedOperatorLifecycle.class);
+
+    private static App localOperator;
+    private static KubernetesClient localOperatorClient;
+
+    private SharedOperatorLifecycle() {
+    }
+
+    static synchronized void ensureStarted(KubernetesClient client,
+            ITBase.OperatorDeployment deploymentType, String deploymentTarget) throws IOException {
+        if (deploymentType == ITBase.OperatorDeployment.remote) {
+            OperatorClusterWideInstaller.ensureInstalled(client, deploymentTarget);
+            return;
+        }
+        if (localOperator != null) {
+            return;
+        }
+        localOperatorClient = new KubernetesClientBuilder().build();
+        createCRDs(localOperatorClient);
+        var app = CDI.current().select(App.class).get();
+        app.start(configOverride -> {
+            configOverride.withKubernetesClient(localOperatorClient);
+            configOverride.withUseSSAToPatchPrimaryResource(false);
+        });
+        localOperator = app;
+        log.info("Started the shared in-process operator for this JVM");
+    }
+
+    static synchronized void stopLocalOperator() {
+        if (localOperator == null) {
+            return;
+        }
+        log.info("Stopping the shared in-process operator");
+        try {
+            localOperator.stop();
+        } finally {
+            localOperator = null;
+            localOperatorClient = null;
+        }
+    }
+
+    private static void createCRDs(KubernetesClient client) {
+        log.info("Creating CRDs");
+        try {
+            var crd = client.load(new FileInputStream(CRD_FILE));
+            crd.createOrReplace();
+            await().atMost(SHORT_DURATION).ignoreExceptions().untilAsserted(() -> {
+                crd.resources().forEach(r -> assertThat(r.get()).isNotNull());
+            });
+        } catch (Exception e) {
+            log.warn("Failed to create the CRD, retrying", e);
+            createCRDs(client);
+        }
+    }
+}
--- a/operator/controller/src/test/java/io/apicurio/registry/operator/it/SharedOperatorLifecycle.java
+++ b/operator/controller/src/test/java/io/apicurio/registry/operator/it/SharedOperatorLifecycle.java
@@ -60,15 +60,20 @@ final class SharedOperatorLifecycle {
 
     private static void createCRDs(KubernetesClient client) {
         log.info("Creating CRDs");
-        try {
-            var crd = client.load(new FileInputStream(CRD_FILE));
-            crd.createOrReplace();
-            await().atMost(SHORT_DURATION).ignoreExceptions().untilAsserted(() -> {
-                crd.resources().forEach(r -> assertThat(r.get()).isNotNull());
-            });
-        } catch (Exception e) {
-            log.warn("Failed to create the CRD, retrying", e);
-            createCRDs(client);
+        Exception last = null;
+        for (int attempt = 1; attempt <= 3; attempt++) {
+            try {
+                var crd = client.load(new FileInputStream(CRD_FILE));
+                crd.createOrReplace();
+                await().atMost(SHORT_DURATION).ignoreExceptions().untilAsserted(() -> {
+                    crd.resources().forEach(r -> assertThat(r.get()).isNotNull());
+                });
+                return;
+            } catch (Exception e) {
+                last = e;
+                log.warn("CRD creation attempt {}/3 failed", attempt, e);
+            }
         }
+        throw new IllegalStateException("Could not create the CRDs after 3 attempts", last);
     }
 }
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -165,7 +165,7 @@ and tags starting with `3.`.
 | `update-openapi.yaml` | Push to main (openapi.json changes) | Auto-copies v3 OpenAPI spec to v2 path, commits if changed, then validates via `validate-openapi.yaml` | 10-15 min |
 | `update-website.yaml` | Release event, workflow_dispatch | Updates `latestRelease.json` on apicurio.github.io with release metadata | 5-10 min |
 | `publish-docs.yaml` | Push to main (docs/**), workflow_dispatch | Builds documentation via Antora playbook and publishes to apicurio.github.io | 15-30 min |
-| `operator.yaml` | Push/PR to main (operator/**) | Operator CI: build + push temp images to ttl.sh (8h TTL), 8-group test matrix on Minikube (smoke, kafka, auth, database, feature, feature-setup, OLM v0, OLM v1), publish to Quay.io on push. Cancels in-progress on new push | 45-90 min |
+| `operator.yaml` | Push/PR to main (operator/**) | Operator CI: build + push temp images to ttl.sh (8h TTL), 9-group test matrix on Minikube (smoke, kafka, auth, database, feature, feature-setup, operator-mutation, OLM v0, OLM v1), publish to Quay.io on push. Cancels in-progress on new push | 45-90 min |
 | `image-scan.yaml` | Daily at 06:00 UTC, workflow_dispatch | Trivy vulnerability scan on `latest-snapshot` image (CRITICAL + HIGH severity). Results uploaded to GitHub Security tab as SARIF | 5-10 min |
 
 ## Reusable Workflows
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -204,6 +204,11 @@ jobs:
             olm-enable: "false"
             olm-v1-enable: "false"
             extra-build-opts: ""
+          - name: operator-mutation
+            groups: "operator-mutation"
+            olm-enable: "false"
+            olm-v1-enable: "false"
+            extra-build-opts: ""
           - name: olm-v1
             groups: "OLM"
             olm-enable: "true"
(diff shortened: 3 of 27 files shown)


aclfe:
Please read the "How to Review" note in the description before diving into the diff. It should make review easier. 

Also TL;DR: 
  - Replaces per-class operator install/teardown across ~28 IT classes with one shared cluster-wide operator 
  - New OperatorInfraExtension and managers as JUnit 5 parameters instead of requiring ITBase inheritance
  - Do not merge before #8686

operator/controller/src/test/java/io/apicurio/registry/operator/it/SharedOperatorLifecycle.java
```diff
@@ -0,0 +1,74 @@
...
+        log.info("Creating CRDs");
+        try {
+            var crd = client.load(new FileInputStream(CRD_FILE));
+            crd.createOrReplace();
+            await().atMost(SHORT_DURATION).ignoreExceptions().untilAsserted(() -> {
+                crd.resources().forEach(r -> assertThat(r.get()).isNotNull());
+            });
+        } catch (Exception e) {
+            log.warn("Failed to create the CRD, retrying", e);
+            createCRDs(client);
```
paoloantinori: This recursive retry has no bound. If CRD creation persistently fails (e.g. `CRD_FILE` is missing → `FileNotFoundException` on every call, or the API server is unreachable), this recurses until `StackOverflowError`, which produces a confusing trace instead of a clear failure.

A bounded loop would be safer, e.g.:
```java
for (int attempt = 1; attempt <= 3; attempt++) {
    try { ... return; }
    catch (Exception e) { log.warn("CRD creation attempt {} failed", attempt, e); }
}
throw new IllegalStateException("Could not create the CRDs after 3 attempts");
```

Also worth noting: `new FileInputStream(CRD_FILE)` can throw before entering the `try` if the file is missing, but since the whole body is wrapped in the `try` it's caught — the issue is purely the unbounded recursion on the retry path.

paoloantinori (commented):
Solid test-infrastructure refactor — the split between a shared long-lived operator (one install per cluster vs. ~28 per-class installs) and per-class throwaway namespaces is the right design, the README does a good job documenting the non-obvious decisions (why not parallel, why no `junit-platform.properties`), and `OperatorInfraExtensionTest` covers the parameter-resolution contract well. I'm reviewing against the visible diff only (the full diff is ~85k chars and was truncated at 4000 lines), so I can't comment on `OperatorClusterWideInstaller` / `OperatorClusterWideInstallerTest` which fell outside the visible portion.

A couple of points worth a look:

- **Unbounded recursive retry in `SharedOperatorLifecycle.createCRDs`** — the `catch (Exception e)` block recurses without a depth limit. A persistent failure (missing CRD file, persistent API-server error) will recurse until `StackOverflowError`, which surfaces as a confusing stack trace rather than a clear failure. A bounded retry loop (e.g. 3 attempts) with a terminal error would be more robust.
- **`localOperatorClient` is not closed** in `SharedOperatorLifecycle.stopLocalOperator()` before being nulled. Minor — it's test infra and the JVM is winding down — but `client.close()` (mirroring `ClassInfra.tearDown()`) would keep the pattern consistent.

operator/controller/src/test/java/io/apicurio/registry/operator/it/SharedOperatorLifecycle.java
aclfe: the recursion doesn't have much delay so that's a good point. can you also review OperatorClusterWideInstaller / OperatorClusterWideInstallerTest and other operator files?

paoloantinori:
Hi @aclfe — re-read after your comment.

On the recursion: agreed it's low-risk on the happy path, but I'd still prefer `SharedOperatorLifecycle.createCRDs`' retry to be bounded with a small backoff rather than unbounded self-recursion. The cost isn't latency — it's that a persistent CRD-creation failure becomes a `StackOverflowError` with no count of how many attempts ran, and no delay means it can hammer the API server. Cheap to harden (a max-attempts loop, or reuse `Awaitility` like the rest of the class does), but fine to defer if you'd rather track it separately.

On `OperatorClusterWideInstaller` / `OperatorClusterWideInstallerTest`: these look solid. Things I specifically checked and liked:
- The dual install guard — `installedInThisJvm` as a JVM-local fast path plus `isOperatorReady` as the cluster-state check — correctly covers both the within-JVM and cross-JVM (CI job group reusing the cluster) cases. `createOrReplace` keeps re-install idempotent, so even a stale/leftover-deployment edge self-heals on the next pass.
- Making `transformInstallFile` package-private and exercising it directly from the unit test is the right call. The two tests that matter most here are `bindsClusterRoleBindingSubjectToTheDedicatedNamespace` and `keepsOperatorSelectorLabelsSoTheReadinessGuardCanFindTheDeployment` — together they guard exactly the two properties that would silently break the readiness guard if a future transform changed them (CRB subject rebinding, and the selector labels `isOperatorReady` filters on). Good coverage.
- `loadInstallFile` mapping `NoSuchFileException` to an actionable message pointing at `make ... dist-install-file` is nice DX.

One minor naming nit, not worth a change on its own: `useLocallyBuiltImage` only flips `imagePullPolicy` to `IfNotPresent` — it doesn't touch the image reference itself. Something like `useLocalImagePullPolicy` would describe it more precisely, but happy to leave it as is.

No blockers from me on these files.

aclfe:
@jsenko @carlesarnal

aclfe:
@EricWittmann thoughts on the PR?

jsenko (commented):
## Review: Solid test infrastructure refactor

Well-structured decomposition — the split between a shared cluster-wide operator (one install instead of ~28 per-class installs), per-class throwaway namespaces, and `@DedicatedOperator` for mutation tests is the right architecture. The commit-by-commit organization makes the review tractable despite the size.

### What works well

- **Shared lifecycle management** with proper `synchronized` guards and cluster-state detection (idempotent across JVMs sharing a cluster)
- **`CONVERTIBLE_ROLE_BINDINGS` allowlist** in `StrimziClusterWideInstaller` — fail-fast on unexpected RoleBindings prevents silent scope escalation on Strimzi version bumps
- **AlreadyExists race handling** in both installers' `ensureNamespace`
- **`OperatorInfraExtension`** is a clean alternative to ITBase inheritance; `DisableUIITTest` is a good worked example
- **paoloantinori's feedback addressed** — `createCRDs` now uses a bounded 3-attempt loop instead of unbounded recursion
- **CI properly split** — `operator-mutation` group gets its own job so `@DedicatedOperator` tests don't conflict with the shared operator

### Observation: inconsistent resource creation robustness

`OperatorClusterWideInstaller.ensureInstalled()` wraps each resource creation in `await().atMost(...).ignoreExceptions()` (lines 60-63), retrying until the resource is confirmed to exist. `StrimziClusterWideInstaller.ensureInstalled()` (line 97) uses bare `createOrReplace` without retry or confirmation — a transient K8s API error on one Strimzi resource would fail the entire install immediately.

The `waitForOperatorReady` at the end catches deployment-level failures, so this won't cause silent bugs — but for consistency and resilience against transient API errors (the exact class of failures this PR is designed to fix), the Strimzi installer could use the same await pattern. Not blocking, but worth considering.

### Dependency note

PR description says "Do not merge before #8686" — this extends #8686's shared Strimzi install pattern to the Apicurio operator. The `StrimziClusterWideInstaller` code in this PR appears to be the actual Strimzi shared-install implementation, so the ordering dependency should be verified (does #8686 establish the pattern that this PR consumes, or does this PR include all the Strimzi changes?).

aclfe:
> ### Observation: inconsistent resource creation robustness
> `OperatorClusterWideInstaller.ensureInstalled()` wraps each resource creation in `await().atMost(...).ignoreExceptions()` (lines 60-63), retrying until the resource is confirmed to exist. `StrimziClusterWideInstaller.ensureInstalled()` (line 97) uses bare `createOrReplace` without retry or confirmation — a transient K8s API error on one Strimzi resource would fail the entire install immediately.
> 
> The `waitForOperatorReady` at the end catches deployment-level failures, so this won't cause silent bugs — but for consistency and resilience against transient API errors (the exact class of failures this PR is designed to fix), the Strimzi installer could use the same await pattern. Not blocking, but worth considering.

It comes from the base PR #8686, I made a change onto this branch as a fix for it. 

> ### Dependency note
> PR description says "Do not merge before #8686" — this extends #8686's shared Strimzi install pattern to the Apicurio operator. The `StrimziClusterWideInstaller` code in this PR appears to be the actual Strimzi shared-install implementation, so the ordering dependency should be verified (does #8686 establish the pattern that this PR consumes, or does this PR include all the Strimzi changes?).

#8686 owns the Strimzi installer, this PR just stacks on its branch (strimzi-cluster-wide-install) and consumes it, so that's why the ordering holds.

Addressed the changes as requested

paoloantinori:
Thanks @aclfe — the Strimzi installer now mirrors the Operator one with the `await().atMost(...).ignoreExceptions()` + `assertThat(...).isNotNull()` pattern. That addresses the consistency concern cleanly. No further comments from my side.

aclfe:
> Thanks @aclfe — the Strimzi installer now mirrors the Operator one with the `await().atMost(...).ignoreExceptions()` + `assertThat(...).isNotNull()` pattern. That addresses the consistency concern cleanly. No further comments from my side.

Awesome! Waiting for https://github.com/Apicurio/apicurio-registry/pull/8686 to merge then, lgtm

aclfe:
Closing this to free up my contributor PR slot. This has been sitting in draft since July waiting on #8686, which is still open, so it is not going anywhere in the near term.

The orchestrator's PR limit counts drafts (`countOpenPrsByAuthor` in `.github/scripts/pr-lifecycle.js:637` lists `state: 'open'`, which includes drafts), so this draft caused #9420 to be auto-closed even though it was already reviewed and approved.

Nothing here is lost. The branch stays up and I will reopen this once #8686 lands. @paoloantinori your review feedback is all applied already, no action needed from you.

PR #8824, opened 2026-07-22
fix: flaky test visibility and reduce retries in CI (#8801)

### LINKED ISSUE #8801: CI: rerunFailingTestsCount masks real flaky test frequency — need visibility
## Description

The `rerunFailingTestsCount` mitigation added in #8388 is working as a safety net, but it hides how often tests actually fail on first attempt. This makes it impossible to:

1. **Measure the real cost** of flaky tests in CI minutes wasted on retries.
2. **Prioritize which flaky tests to fix first** — a test that retries 30% of the time is a bigger problem than one that retries 5%, but both show as "green".
3. **Detect new flaky tests** — if a test starts failing intermittently, there is no signal until retries also start failing.

## Evidence

PR #8537 (maven-dependency-plugin → resources plugin, POM-only change) failed Verify twice in 12 hours. Since the PR changes only build config, test failures are almost certainly flaky. The `rerunFailingTestsCount` succeeds on retry in most cases, so these failures are invisible in the success rate.

## Suggested Fixes

- **Publish retry metrics:** Add a workflow step that parses surefire/failsafe reports and posts a summary comment on the PR when any test was retried.
- **Fail on repeated retries:** If a test fails more than once in the same run (first attempt + retry both fail), make it a hard failure or review gate.
- **Dashboard:** Aggregate retry counts across PRs into a weekly report.

## Related

- #8404 — KafkaSqlStartupVerificationIT flaky (open)
- #8742 — UpgradeOLMITTest flaky (open)
- #8388 — rerunFailingTestsCount PR (merged)

singhshresth26:
### Description of Changes
This PR addresses issue #8801 to tackle the visibility issues introduced by the `rerunFailingTestsCount` parameter in Maven Surefire and Failsafe. When tests fail on the first attempt but succeed on a retry, the build succeeds (showing "green") but masks underlying test instability, wastes CI resource minutes, and makes it difficult to detect or prioritize flaky tests.

This PR implements two complementary solutions: enforcing stricter failure gates and introducing a feedback mechanism to report flaky test attempts directly on pull requests.

#### 1. Stricter Failure Gates
* **Reduced Retry Count:** Changed the test retry configuration (`rerunFailingTestsCount`) in unit and integration test runs from `2` to `1`.
  * In [verify-unit-tests.yaml](file:///d:/github_forked_repo/apicurio-registry/.github/workflows/verify-unit-tests.yaml), `-Dsurefire.rerunFailingTestsCount` is reduced to `1`.
  * In [verify-integration-tests.yaml](file:///d:/github_forked_repo/apicurio-registry/.github/workflows/verify-integration-tests.yaml), `-Dfailsafe.rerunFailingTestsCount` is reduced to `1`.
* **Impact:** A test is now only retried once; if it fails twice in a row, the CI job immediately fails, establishing a stricter gate for transient failures.

#### 2. Automated Flaky Test Reporting Pipeline
* **Test XML Log Parser (`parse-flaky-tests.py`):** Added a lightweight Python parser script under [parse-flaky-tests.py](file:///d:/github_forked_repo/apicurio-registry/.github/scripts/parse-flaky-tests.py). This script recursively finds all `TEST-*.xml` report outputs and checks `<testcase>` tags for `<flakyFailure>` or `<flakyError>` elements (which represent failed attempts that subsequently passed). It generates a clean JSON report mapping the class, test name, retry counts, and error details.
* **Artifact Upload:** Embedded the parsing script into the `verify-unit-tests.yaml`, `verify-integration-tests.yaml`, and `operator.yaml` workflows. After test execution, the parser scans the workspace and uploads a `flaky-tests-*.json` artifact containing details of any retries. These steps run with `if: always()` to ensure that retry metrics are gathered regardless of the overall test success.
* **PR Comment Reporting:** Updated [pr-lifecycle.js](file:///d:/github_forked_repo/apicurio-registry/.github/scripts/pr-lifecycle.js) (the PR Lifecycle state-machine runner). During the `workflow_run` event when the Verification workflow completes:
  1. The orchestrator downloads all artifacts starting with `flaky-tests-`.
  2. It unzips and parses the JSON contents.
  3. It aggregates the list of all retried tests across shards/jobs.
  4. If retries are detected, it formats them into a clean markdown table indicating the Job/Shard name, Test Class, Test Name, and Retry Count, posting a comment on the PR.
  5. If there are previous flaky test summary comments, it automatically minimizes them via GitHub's GraphQL API to prevent cluttering the PR timeline.

### Linked Issue
Fixes #8801

### Reviewer Notes
Using `workflow_run` inside the `PR Lifecycle` workflow ensures the comment posting logic is executed with the default branch's write-permissions. This allows the bot to post comments and minimize older comments successfully, even for PRs submitted from external forks where the standard GITHUB_TOKEN is read-only.


.github/scripts/parse-flaky-tests.py
```diff
@@ -0,0 +1,66 @@
...
+    parser.add_argument("--output-file", required=True, help="Output JSON file path")
+    return parser.parse_args()
+
+def main():
+    args = parse_args()
+    flaky_tests = []
+
+    # Find all TEST-*.xml files recursively
+    search_pattern = os.path.join(args.reports_dir, "**/TEST-*.xml")
+    xml_files = glob.glob(search_pattern, recursive=True)
```
paoloantinori, line 34: `glob.glob(..., recursive=True)` is Python 3.5+ only. The workflows invoke this script as `python`, but GitHub-hosted Ubuntu runners ship `python3` (not `python`) by default — either an exit 127 or a python2 TypeError here fails this `if: always()` step and blocks the verify job. `python3` in the three workflow `run:` steps is the safe fix.

.github/scripts/parse-flaky-tests.py
```diff
@@ -0,0 +1,66 @@
...
+                                "error_type": err_type
+                            })
+                        
+                        flaky_tests.append({
+                            "class": classname,
+                            "test": name,
+                            "retries": len(failures_list),
+                            "details": failures_list
+                        })
+        except Exception as e:
```
paoloantinori, line 72: Swallowing every exception means a systemic parse failure looks identical to 'no flaky tests detected'. Consider writing the file name and error to stderr so it surfaces in the step log while staying non-fatal.

paoloantinori (commented):
Overall a sensible change — surfacing flaky-test retries on PRs (instead of `rerunFailingTestsCount` silently green-washing them) and tightening 2->1 are real CI-quality wins, and the outbox-artifact -> JSON -> PR-comment pipeline is well thought out (including the path-traversal guard on the zip entries). One portability issue is worth fixing before this lands, plus a smaller robustness note.

- **`python` vs `python3` (likely breaks the new step):** `parse-flaky-tests.py` is invoked as `python` in [verify-unit-tests.yaml](https://github.com/Apicurio/apicurio-registry/blob/main/.github/workflows/verify-unit-tests.yaml), `verify-integration-tests.yaml`, and `operator.yaml`, but the script relies on `glob.glob(pattern, recursive=True)`, which is Python 3.5+ only (it raises TypeError on Python 2). Current GitHub-hosted Ubuntu runners ship `python3` on PATH but not necessarily `python` (python2 is gone and `python-is-python3` is not installed by default). A `python: command not found` exits 127 here, and since the step runs under `if: always()` with no `continue-on-error`, it would fail the verify job and block PRs — the opposite of what this PR wants. Switching the three invocations to `python3` is the safe fix.
- **Silent error swallowing in the parser:** the broad `except Exception: pass` in `parse-flaky-tests.py` means a systematically broken parse (a changed Surefire schema, or the script run from the wrong `--reports-dir`) produces an empty `[]` artifact with zero signal — the PR comment would report 'no flaky tests' while real flakes go unreported. Logging the offending file + error to stderr (still non-fatal) would make this diagnosable.
- **Minor:** `postFlakyTestsSummary` shells out to `unzip -l`/`unzip -o`; that is present on ubuntu-latest today and the path guard is good, but a pure-JS unzip would remove the external-binary dependency. Low priority.

.github/scripts/parse-flaky-tests.py
singhshresth26: Done! Switched all three workflows to use `python3`.

.github/scripts/parse-flaky-tests.py
singhshresth26: Done! Updated the script to log parsing exceptions to `stderr` while keeping it non-fatal.

singhshresth26:
Hi @EricWittmann, @paoloantinori 
Just a gentle follow-up on this PR.  
Would love to know if any changes or improvements are needed from my side. Thanks!

paoloantinori:
Thanks @singhshresth26 for the ping, and sorry for the delay. Both points from my earlier pass are addressed cleanly in the current head — `python3` is now used in all three workflows, and the parser logs the offending file + error to stderr while staying non-fatal. 

One small non-blocking thing worth considering before or as a follow-up: in `postFlakyTestsSummary` ([pr-lifecycle.js](https://github.com/Apicurio/apicurio-registry/blob/main/.github/scripts/pr-lifecycle.js)), `listWorkflowRunArtifacts` is called without pagination and defaults to 30 items per page. For sharded unit-test runs that also carry build/target/coverage artifacts, the total artifact count can exceed 30, and the `flaky-tests-*` filter is applied *after* the fetch — so artifacts sitting on page 2 would be silently dropped and the PR comment would under-report flakes. Switching to `github.paginate(github.rest.actions.listWorkflowRunArtifacts, { owner, repo, run_id: workflowRun.id })` (the same pattern `minimizePreviousFlakyComments` already uses for comments) would make this robust. Not a blocker from my side — the core change is in good shape.

singhshresth26:
Thanks for the suggestion, @paoloantinori!

I've updated the script to use github.paginate for listWorkflowRunArtifacts in both postFlakyTestsSummary and postDecisionSummary to ensure we don't drop any artifacts.

paoloantinori:
Looks good @singhshresth26 — both `postFlakyTestsSummary` and `postDecisionSummary` now go through `github.paginate(...)` with `per_page: 100`, so artifact pagination is handled correctly across both paths. No further concerns from my side.

EricWittmann (commented):
This is looking good @singhshresth26 — the pipeline design is solid and all review feedback has been addressed. The branch has merge conflicts against `main` though (likely from recent workflow file changes that landed). Please rebase and resolve, and we'll trigger CI again after.

singhshresth26:
Done! I have rebased the branch on the latest main, resolved the conflict in the unit tests workflow (retaining the new shard timing checks alongside the flaky test reporting changes), and force-pushed.

Ready for CI to be triggered. Thank you!

.github/workflows/operator.yaml
```diff
@@ -131,6 +131,11 @@ jobs:
           name: operator-test-images
           path: /tmp/
 
+      - name: Create CNI directory
+        run: |
+          sudo mkdir -p /etc/cni/net.d
```
paoloantinori: This `Create CNI directory` step (also added to the second operator job, `verify-integration-tests.yaml`, and `release-operator.yaml`) doesn't relate to flaky-test visibility or the retry-count reduction described in the PR. Consider moving it to its own PR with its own justification — particularly since it also touches the release workflow.

.github/scripts/pr-lifecycle.js
```diff
@@ -1393,6 +1395,136 @@ async function postDecisionSummary(github, owner, repo, workflowRun, prNumber, c
...
+            }
+          }
+        } catch (e) {
+          core.warning(`Failed to parse json file ${f}: ${e.message}`);
+        }
+      }
+
+      try {
+        fs.rmSync(tmpDir, { recursive: true, force: true });
+      } catch (e) {}
```
paoloantinori: `tmpDir` is created at `fs.mkdtempSync(...)` but only removed by this `fs.rmSync` at the end of the loop body — there's no `try/finally` around the per-artifact block. If anything between them throws (`downloadArtifact`, `unzip -l`/`unzip -o`, or a large-but-valid `JSON.parse`), the exception jumps straight to the function's outer `catch` (which only logs) and `tmpDir` is leaked under `/tmp`. Wrapping the block body in `try/finally` makes the cleanup reliable. Low impact on an ephemeral runner, but a cheap fix.

.github/scripts/pr-lifecycle.js
```diff
@@ -1393,6 +1395,136 @@ async function postDecisionSummary(github, owner, repo, workflowRun, prNumber, c
...
+        owner, repo, artifact_id: artifact.id, archive_format: 'zip',
+      });
+
+      const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'flaky-tests-'));
+      const zipPath = path.join(tmpDir, 'artifact.zip');
+      fs.writeFileSync(zipPath, Buffer.from(zip));
+
+      const listing = execSync(`unzip -l "${zipPath}"`, { encoding: 'utf8' });
+      const entryLines = listing.split('\n').filter(l => l.includes('.json'));
+      let valid = true;
```
paoloantinori: The traversal guard only inspects zip entries whose listing line contains `.json` (`entryLines = …filter(l => l.includes('.json'))`), but the later `unzip -o "$zipPath" -d "$tmpDir"` extracts every entry. A non-`.json` path like `../etc/foo` would pass validation and be extracted outside `tmpDir`. If this is meant as defense-in-depth, validate all entries (or extract flattened). Risk is low here because artifacts originate from this project's own CI, but the check currently gives false confidence.

paoloantinori (commented):
### Overview
The core of this PR — reducing `rerunFailingTestsCount` from 2 to 1, the Python flaky-report parser (`parse-flaky-tests.py`), and the `workflow_run`-based PR-comment pipeline (including the good `github.paginate` refactor in `postDecisionSummary`) — is sound and reasonably well-defended. The main concerns are a scope/separation issue and two low-impact robustness gaps in the new artifact-handling code.

- **Scope (unrelated change):** The `Create CNI directory` step (`sudo mkdir -p /etc/cni/net.d`) is added to both `operator.yaml` jobs, `verify-integration-tests.yaml`, and `release-operator.yaml`, but is unrelated to flaky-test visibility/retry reduction and isn't mentioned in the PR description. Worth splitting into its own PR — especially the `release-operator.yaml` edit.
- **Temp-dir leak:** In `postFlakyTestsSummary`, `tmpDir` is removed only by the final `fs.rmSync` with no `try/finally`; any throw between `mkdtempSync` and that line escapes to the outer `catch` (which only logs a warning), leaking the directory under `/tmp`.
- **Incomplete traversal guard:** The zip path-traversal check inspects only entries whose listing line contains `.json`, while the later `unzip -o` extracts every entry — so a non-`.json` traversal path would bypass the check. Low practical risk since artifacts come from this project's own CI, but the guard is misleading as written.
- Unverified (not escalated to inline): confirm the `actions/upload-artifact@043fb46d…# v7` pin is intended (no public v7 of upload-artifact exists), and that the integration-test matrix yields unique `flaky-tests-<storage>-<name|groups>` artifact names — duplicates fail the job under upload-artifact v4.

.github/workflows/operator.yaml
singhshresth26: Reverted this step from all three workflow files (`operator.yaml`, `release-operator.yaml`, and `verify-integration-tests.yaml`) to keep it out of this PR.

.github/scripts/pr-lifecycle.js
singhshresth26: Done! Wrapped the inner block body in a `try/finally` block to ensure `fs.rmSync` is always executed in the `finally` clause if any exception is thrown after directory creation.

.github/scripts/pr-lifecycle.js
singhshresth26: I've updated the check to validate the entry names of all lines in the ZIP listing instead of filtering only for `.json` files, ensuring no traversal paths bypass the validation.

paoloantinori:
Thanks — the traversal check is now correctly applied to every entry in the ZIP listing rather than only the `.json` files, so a malicious archive can't slip a `../` path past validation on a non-JSON entry. That closes the gap cleanly.

The rest of the pipeline (size cap, `path.isAbsolute()` guard, temp-dir isolation, `minimizeComment` on refresh) still reads well. From my side the security-shape concerns from the earlier passes are addressed.

EricWittmann:
Thanks @singhshresth26 — nice CI improvement. Surfacing flaky tests as PR comments instead of silently green-washing them is a real quality win. Appreciate the thorough security hardening on the artifact handling too. Merged!

PR #8826, opened 2026-07-22
fix(kafkasql): don't emit create events on a dryRun create

### LINKED ISSUE #8825: bug: kafkasql emits ARTIFACT_CREATED / ARTIFACT_VERSION_CREATED events on a dryRun create
## Description

On kafkasql, a create with `dryRun=true` still publishes creation events to the `registry-events` topic, even though the artifact is never saved. A dry run is only supposed to validate the request. It should not produce any side effect, and publishing to Kafka is a side effect.

The cause is in `KafkaSqlRegistryStorage.createArtifact`. It passes `dryRun` into the journal message, waits for the write to finish, and then fires the outbox events without ever checking the `dryRun` flag ([`createArtifact` L457-L466][ksql-create]). The same thing happens in `createArtifactVersion` ([L505-L508][ksql-create-version]).

The sql variant does not behave this way. It fires the event from inside the database transaction, and on a dry run that transaction is rolled back, so the event never leaves ([`setRollback(true)` plus `outboxEvent.fire(...)` inside `withHandle`, L522-L566][sql-create]). The version event works the same way at [`SqlVersionRepository` L600][sql-version].

So the same REST call behaves differently depending on the storage backend, which is a parity problem between the variants.

I could not find a test that checks a dryRun create produces no event on either variant, so a fix should probably add one.

### Steps to Reproduce

1. Run the Registry with `APICURIO_STORAGE_KIND=kafkasql` against a Kafka broker.
2. Subscribe a consumer to the `registry-events` topic.
3. Do a dryRun create:
   ```
   curl -X POST "http://localhost:8080/apis/registry/v3/groups/default/artifacts?dryRun=true" \
     -d '{"artifactId":"dry-run-test","artifactType":"JSON","firstVersion":{"content":{"content":"{}","contentType":"application/json"}}}'
   ```
4. Two events (`ARTIFACT_CREATED`, `ARTIFACT_VERSION_CREATED`) arrive on `registry-events`, even though the artifact is never created (`GET .../artifacts/dry-run-test` returns 404).

Expected: a dryRun create sends no events, matching the sql variant.

Actual: `ARTIFACT_CREATED` and `ARTIFACT_VERSION_CREATED` are sent, but the artifact is not persisted.

[ksql-create]: https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java#L457-L466
[ksql-create-version]: https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java#L505-L508
[sql-create]: https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java#L522-L566
[sql-version]: https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java#L600

ShivaGupta-14:
On the kafkasql variant, creating an artifact or version with `dryRun=true` still published an `ARTIFACT_CREATED` / `ARTIFACT_VERSION_CREATED` event to the `registry-events` topic, even though a dry run rolls back and saves nothing. A dry run should validate only and produce no side effects, so no event should be emitted.

`createArtifact` and `createArtifactVersion` fired the outbox events right after `waitForResponse` without checking `dryRun`. The sql variant doesn't have this problem, since it fires the event inside the transaction that `dryRun` rolls back, so nothing is emitted. There was also no test asserting a dryRun create stays silent, which is probably why it went unnoticed.

### Changes

- Guard the `outboxEvent.fire(...)` calls in `createArtifact` and `createArtifactVersion` with `if (!dryRun)`, so a dry run saves nothing and emits nothing, matching the sql variant.
- Add tests for both the artifact and version paths asserting a dryRun create emits no event. Each does a dryRun create with a marker name, then a real create as a barrier, and checks only the real event appears. They run under both the sql and kafkasql variants.

### Test plan

- [x] `KafkaSqlEventsTest` dryRun create tests (kafkasql): pass
- [x] `RegistryEventsTest` dryRun create tests (sql): pass, no regression
- [x] `mvnw checkstyle:check -pl app` clean

Closes #8825

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
@@ -460,10 +460,12 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
         Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createdArtifact = (Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto>) coordinator
                 .waitForResponse(uuid);
 
-        outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactCreated.of(createdArtifact.getLeft())));
+        if (!dryRun) {
+            outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactCreated.of(createdArtifact.getLeft())));
 
-        if (createdArtifact.getRight() != null) {
-            outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactVersionCreated.of(createdArtifact.getRight())));
+            if (createdArtifact.getRight() != null) {
+                outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactVersionCreated.of(createdArtifact.getRight())));
+            }
         }
 
         return createdArtifact;
@@ -506,7 +508,9 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
         var uuid = blockOnResult(submitter.submitMessage(message));
         ArtifactVersionMetaDataDto versionMetaDataDto = (ArtifactVersionMetaDataDto) coordinator
                 .waitForResponse(uuid);
-        outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactVersionCreated.of(versionMetaDataDto)));
+        if (!dryRun) {
+            outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactVersionCreated.of(versionMetaDataDto)));
+        }
         return versionMetaDataDto;
     }
 
--- a/app/src/test/java/io/apicurio/registry/event/sql/RegistryEventsTest.java
+++ b/app/src/test/java/io/apicurio/registry/event/sql/RegistryEventsTest.java
@@ -4,12 +4,15 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.apicurio.registry.AbstractResourceTestBase;
+import io.apicurio.registry.rest.client.models.CreateArtifact;
 import io.apicurio.registry.rest.client.models.CreateArtifactResponse;
+import io.apicurio.registry.rest.client.models.CreateVersion;
 import io.apicurio.registry.rest.client.models.EditableArtifactMetaData;
 import io.apicurio.registry.rest.client.models.EditableGroupMetaData;
 import io.apicurio.registry.rest.client.models.EditableVersionMetaData;
 import io.apicurio.registry.rest.client.models.GroupMetaData;
 import io.apicurio.registry.rest.client.models.Labels;
+import io.apicurio.registry.rest.client.models.VersionContent;
 import io.apicurio.registry.rest.client.models.VersionState;
 import io.apicurio.registry.rest.client.models.WrappedVersionState;
 import io.apicurio.registry.rules.validity.ValidityLevel;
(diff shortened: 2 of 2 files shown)


app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -459,10 +459,12 @@ public Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createArtifact(Stri
         Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createdArtifact = (Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto>) coordinator
                 .waitForResponse(uuid);
 
-        outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactCreated.of(createdArtifact.getLeft())));
+        if (!dryRun) {
+            outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactCreated.of(createdArtifact.getLeft())));
```
paoloantinori, line 464: Worth a brief comment that a dry-run create never returns a right-hand version DTO (the coordinator returns after validation/rollback), so this inner `if` is effectively unreachable when `dryRun == true`. Makes the nesting read as intentional rather than accidental.

app/src/test/java/io/apicurio/registry/event/sql/RegistryEventsTest.java
```diff
@@ -265,6 +268,58 @@ public void createArtifactVersion() throws Exception {
...
+
+        CreateArtifact dryRunCreate = buildCreateArtifact(artifactId, "dryRunLeakName");
+        clientV3.groups().byGroupId(groupId).artifacts().post(dryRunCreate,
+                config -> config.queryParameters.dryRun = true);
+
+        // Real create of the same id acts as a barrier: any leaked dry-run event precedes it.
+        CreateArtifact realCreate = buildCreateArtifact(artifactId, "realCreateName");
+        clientV3.groups().byGroupId(groupId).artifacts().post(realCreate);
+
+        List<JsonNode> events = lookupEvent(consumer, ARTIFACT_CREATED,
```
paoloantinori, line 288: Good barrier pattern. One thing to double-check: `lookupEvent` should drain past the barrier event before returning, otherwise a slow kafkasql consumer could surface the leaked event together with the real one (size==2) and still pass on a *different* assertion. The current `assertNotEquals("dryRunLeakName", ...)` loop makes the test robust against that regardless — nice defensive belt.

paoloantinori (commented):
Clean, well-targeted fix — guarding the two `outboxEvent.fire(...)` calls with `!dryRun` is the right minimal change, and the new `RegistryEventsTest` cases (inherited by `KafkaSqlEventsTest`, so they exercise the buggy variant) verify the leak is gone. A couple of confirmations and one non-blocking note.

- **Completeness check passes.** I grepped `KafkaSqlRegistryStorage` for every `outboxEvent.fire(...)` site and cross-referenced the `dryRun`-bearing methods on `RegistryStorage`. Only three storage methods take a `dryRun` flag: `createArtifact`, `createArtifactVersion` (both fixed here), and `updateArtifactVersionState`. The last one ([source](https://github.com/Apicurio/apicurio-registry/blob/main/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java#L787)) routes the flag into the Kafka message but does **not** emit an outbox event, so there's no leak there. Conclusion: the two guarded sites are the complete set of event-emitting dryRun paths in this class. Worth having confirmed rather than assumed, since the PR is framed around the sql/kafkasql divergence.
- **Barrier pattern is sound, with one caveat.** The tests do a dryRun create with a marker name, then a real create as a barrier, and assert exactly one event. This correctly distinguishes "no leak" from "consumer lag". The one timing-sensitive edge: in the kafkasql variant, `lookupEvent` filters by `groupId/artifactId/(version)`, and both the leaked (if any) and real events share those keys — so the assertion `events.size() == 1` would catch a leak as `size == 2`. That's the right shape; just flagging that the assertion's strength depends on `lookupEvent` consuming past the barrier event before returning, which the helper presumably does. Not a blocker.
- **Nesting nit (non-blocking).** In `createArtifact`, `if (createdArtifact.getRight() != null)` is now nested inside `if (!dryRun)`. Correct as written, but a one-line comment that a dry-run create never produces a version DTO on the right side (so the inner check is unreachable in that path) would make the intent self-documenting. Feel free to ignore.

From my side this looks ready to merge once CI is green; nothing blocking.

EricWittmann:
Thanks for the contribution @ShivaGupta-14 :+1:

Issue #8800, opened 2026-07-22
CI: Node.js 20 deprecation warnings in GitHub Actions


paoloantinori:
## Description

Every Verify and PR Lifecycle workflow run emits deprecation warnings because multiple Actions still target Node.js 20, which GitHub is phasing out (see [changelog](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)).

### Affected Actions

| Action | Warning |
|--------|---------|
| `dorny/paths-filter` | Targets Node.js 20, forced to run on Node.js 24 |
| `actions/upload-artifact@v4` | Targets Node.js 20, forced to run on Node.js 24 |

## Impact

- Today: warnings only (no failures).
- Soon: GitHub will remove Node.js 20 fallback, and these actions will fail outright.
- Any PR, including unrelated ones, will be blocked.

## Suggested Fix

- Pin or update `dorny/paths-filter` to a version that targets Node.js 22+.
- Ensure `actions/upload-artifact` is on `v4.x` with a Node 22+ runtime.
- Audit all other actions in `.github/workflows/` for the same pattern.
- Consider adding `actions/setup-node` with an explicit version to the workflows.

nexpectArpit:
@paoloantinori may i work on this issue? Also Is issue assignation is necessary to work on the issue?

PR #8827, opened 2026-07-22
chore(ci): update GitHub Actions for Node.js 20 deprecation

aaravmahajanofficial:
* Upgraded `docker/setup-buildx-action` to `v4.2.0` in `.github/workflows/release-images.yaml`, `.github/workflows/reusable-docker-build.yaml`, and `.github/workflows/verify-publish.yaml` 
* Updated `nick-fields/retry` to `v4.0.0` in `.github/workflows/reusable-docker-build.yaml` and `.github/workflows/verify-publish.yaml`
* Updated `softprops/action-gh-release` to `v3.0.2` in `.github/workflows/release.yaml` 
* Updated `stCarolas/setup-maven` to `v5.1` in `.github/workflows/update-openapi.yaml` 
* Updated `dorny/paths-filter` to `v4.0.2` in `.github/workflows/verify.yaml` 
* Updated `crazy-max/ghaction-import-gpg` to `v7` in `.github/workflows/release-artifacts.yaml` - moved from env: to with: inputs
Related PR: https://github.com/crazy-max/ghaction-import-gpg/commit/106dc525b1c9988d4a1999e97c6e94be512a4812
* Updated `codeql-action/upload-sarif` to `v4`
Related Issue: https://github.com/github/codeql-action/issues/3271

Upstream Issue: #8800


.github/workflows/verify.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
 
       - name: Detect changed paths
-        uses: dorny/paths-filter@d1c1ffe0248fe513906c8e24db8ea791d46f8590 # v3
+        uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
```
paoloantinori, line 45: Pin verified: `7b450fff21473bca461d4b92ce414b9d0420d706` is the GPG-signed commit of the upstream `v4.0.2` tag. Worth confirming this is the only `dorny/paths-filter` reference under `.github/workflows/` so the bump is not left half-done across workflows.

paoloantinori (commented):
Straightforward, correctly-pinned bump of `dorny/paths-filter` from v3 to v4.0.2. The commit SHA `7b450fff21473bca461d4b92ce414b9d0420d706` matches the GPG-signed `v4.0.2` release tag on the upstream repo, so the pin is good.

A couple of notes worth a quick check before merge:

- **PR title vs. content**: the title says "Update Node.js version" but the actual change is the `dorny/paths-filter` action pin. Not blocking, but renaming it (e.g. `chore: bump dorny/paths-filter to v4.0.2`) keeps the history/dependabot-trail readable, since the real driver (#8800) is the Node 24 runtime upgrade inside that action.
- **Coverage of the bump**: this only touches `.github/workflows/verify.yaml`. It is worth confirming whether `dorny/paths-filter` is referenced in any other workflow file under `.github/workflows/`, because a partial bump would leave the repo running two versions of the same action (and two different Node runtimes).
- **Behavioral change awareness**: per the upstream changelog, v4.x runs on Node 24 and there are no breaking input/output changes between v3 and v4.0.2 for the `filters` usage here, so the `id: filter` / `steps.filter.outputs.*` contract this workflow relies on should be unaffected.

.github/workflows/verify.yaml
aaravmahajanofficial: @paoloantinori Have checked. `dorny/paths-filter` is only used in one workflow file: `.github/workflows/verify.yaml`

.github/workflows/release-artifacts.yaml
```diff
@@ -101,10 +101,10 @@ jobs:
           echo "<settings><servers><server><id>central</id><username>${{ secrets.CENTRAL_USERNAME }}</username><password>${{ secrets.CENTRAL_TOKEN }}</password></server></servers><profiles><profile><id>central</id><activation><activeByDefault>true</activeByDefault></activation><properties><gpg.executable>gpg</gpg.executable></properties></profile></profiles></settings>" > /home/runner/.m2/settings.xml
 
       - name: Import GPG Key
-        uses: crazy-max/ghaction-import-gpg@f6f458f535f4ccdf100400ee0755c0e857226a66
-        env:
-          GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
-          PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
+        uses: crazy-max/ghaction-import-gpg@2dc316deee8e90f13e1a351ab510b4d5bc0c82cd # v7
+        with:
```
paoloantinori, line 105: Heads-up: this SHA targets `ghaction-import-gpg@v7.0.0`, which moved to **Node 24 + ESM** and requires Actions Runner ≥ v2.327.1. The `env:` → `with:` migration here is correct, but if this job ever runs on self-hosted runners, confirm they're recent enough — a too-old runner will fail to start the step with an opaque Node error.

.github/workflows/release-images.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
 
       - name: Set up Docker Buildx
         id: buildx
-        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+        uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4
```
paoloantinori: Minor: the SHA resolves to `setup-buildx-action@v4.2.0`, but the inline comment just says `# v4`. Aligning the comment to the exact tag (`# v4.2.0`) would help `git blame` readers later.

paoloantinori (commented):
Overall: a sensible, well-scoped CI housekeeping PR that bumps six third-party Actions off the deprecated Node 16/20 runtimes. I verified each target SHA against the upstream release tags — all match. Two items worth flagging below, the first more important than the second.

- **`ghaction-import-gpg` v7 — confirm the runtime bump doesn't break your release job.** This is the only action in the PR that moved a major version with a runtime change: v7.0.0 switched to **Node 24** and ESM ([release notes](https://github.com/crazy-max/ghaction-import-gpg/releases/tag/v7.0.0)), requiring [Actions Runner v2.327.1+](https://github.com/actions/runner/releases/tag/v2.327.1). GitHub-hosted runners satisfy this, but if `release-artifacts.yaml` ever runs on self-hosted runners, make sure they're new enough — otherwise the GPG import step silently fails to start. The move from `env:` to `with:` inputs (`gpg_private_key`, `passphrase`) is handled correctly in the diff.
- **Version comments vs. PR description mismatch (cosmetic).** The PR body cites specific point releases (`setup-buildx-action@v4.2.0`, `setup-maven@v5.1`, `softprops/action-gh-release@v3.0.2`, `codeql-action/upload-sarif@v4`), but the inline comments after the SHAs only say `# v4`, `# v5`, `# v3`. The SHAs themselves resolve to the right tags, so this is purely a comment-precision nit — worth aligning the comments to the exact versions for future `git blame` clarity.
- **`nick-fields/retry` v4 is a no-op upgrade for your usage.** v4.0.0 is just the Node 20 runtime bump over v3.0.2 (no input/output changes); the `with:` blocks in `reusable-docker-build.yaml` and `verify-publish.yaml` are unchanged and still valid.
- **Consistency suggestion:** the bump touches 10 of your workflow files, but `docker/setup-qemu-action@...# v4` is already current and untouched — good. Before merging, it may be worth a quick grep for any remaining `@v3`/`@v2` references to these same actions in workflows NOT in this PR, so the fleet ends up on a single version.

.github/workflows/release-images.yaml
aaravmahajanofficial: Have added the exact tags. Thanks!

.github/workflows/release-artifacts.yaml
aaravmahajanofficial: Thanks for the heads-up! This job currently uses GitHub-hosted runners, so we are covered.

aaravmahajanofficial:
Hi @paoloantinori , just a gentle bump. I have resolved the review comments, could you take another look when you have a chance? Thank you :)

paoloantinori:
@aaravmahajanofficial Thanks for the bump — re-read the current head (`8bbaadd`) and it looks good. The six Action bumps are all SHA-pinned to their upstream release tags, and the `ghaction-import-gpg` migration from `env:` to `with:` inputs is the correct v7 API shape. No further concerns from me — ready for a maintainer to merge.

aaravmahajanofficial:
Hi @EricWittmann, just a friendly reminder on this PR whenever you have a moment. Please let me know if there is anything else I should address. Thank you.

EricWittmann:
Thanks for the contribution @aaravmahajanofficial! Nice work verifying all the SHA pins against upstream release tags.

PR #8830, opened 2026-07-22
fix(ui): implement graceful error handling in VersionComments component

### LINKED ISSUE #8829: Fix version comments error handling
**Problem:** 
In the `VersionComments.tsx` component of the Apicurio Registry UI, backend API errors (such as failures to fetch, create, update, or delete comments) are caught but not surfaced to the user. The `catch` blocks currently contain `// TODO handle error` and simply hide the loading modal. This leads to silent failures, leaving the user unaware of whether their action succeeded or failed.

### Proposed Solution

**Solution:**
Implement proper API error handling in `VersionComments.tsx`:
1. Introduce a local error state (`actionError`) to capture the error message.
2. Render a dismissible PatternFly `<Alert>` component at the top of the comment list when an error occurs.
3. Update all `catch` blocks for fetching, adding, editing, and deleting comments to set this error state appropriately, rather than failing silently.
4. Remove all `// TODO handle error` comments in the file.

**Benefit:** 
This will significantly improve the UI/UX by providing immediate and clear feedback to users when operations fail, preventing confusion caused by silent errors. This resolves technical debt and improves the overall robustness of the application.

### Additional Context

This is a UI-focused cleanup and UX improvement task. I would like to work on this issue and submit a PR for it.

hussainjamal760:
Description:

## What does this PR do?
This PR introduces proper UI error handling for backend API calls within the VersionComments component of the Apicurio Registry UI. It replaces the placeholder // TODO handle error comments by implementing a React state-driven approach using PatternFly's dismissible <Alert> component to surface API failures directly to the user.

## Why is this necessary?
Previously, if a user's request to fetch, create, edit, or delete a comment failed (e.g., due to a network timeout or server error), the UI would fail silently. It merely hid the loading modal, leaving the user unaware of the failure and causing a poor user experience. This PR resolves that technical debt and ensures users receive immediate visual feedback when comment operations fail.

## Testing / Verification
Temporarily forced an API failure in the local environment and verified that the inline PatternFly <Alert> renders correctly with the appropriate context-specific error message.
Verified that the alert can be dismissed by the user.

## Related Issue
Fixes #8829


ui/ui-app/src/app/pages/version/components/tabs/VersionComments.tsx
```diff
@@ -172,6 +175,15 @@ export const VersionComments: FunctionComponent<VersionCommentsProps> = (props:
 
     return (
```
paoloantinori, line 180: Consider clearing `actionError` at the start of each action handler (or in each success path). Right now it is only ever set in the `catch` blocks or cleared by the dismiss button, so a previously-shown error persists even after a later operation succeeds â e.g. a failed delete followed by a successful add leaves the "Error deleting comment" alert on screen alongside the new comment. A `setActionError(undefined)` at the top of `onCreate`/`onUpdate`/`onDelete`/`onFetchComments` would resolve it.

paoloantinori (commented):
Nice, focused improvement â replacing the silent `// TODO handle error` stubs with visible PatternFly alerts is a clear UX win, and the dismissible inline `<Alert>` is the right idiom here.

One lifecycle issue worth addressing before this lands:

- **The error is never cleared on a successful action.** `actionError` is only ever *set* (in the `catch` blocks) or cleared via the dismiss button â it is not reset when an operation succeeds or when a new one starts. Concretely: if a delete fails and the user then adds a comment successfully, the stale "Error deleting comment" alert stays on screen right above the freshly added comment. The fix is to `setActionError(undefined)` at the start of each handler (`onCreate`, `onUpdate`, `onDelete`, `onFetchComments`) so every new attempt clears any prior error â or, equivalently, clear it in each success path.

Minor nit (feel free to ignore): the `catch` blocks swallow the underlying error, so a 401/403, a 500, and a network timeout all surface the same generic message. That is fine for a first pass, but if the `Services` layer already exposes a readable message, threading it through (at least for the fetch failure) would help users distinguish "not allowed" from "server down".

hussainjamal760:
@paoloantinori Thank you so much for the detailed review and guidance! I have updated the code to clear actionError at the start of each handler so errors don't persist on subsequent successful actions. I also updated the catch blocks to try and surface the underlying error.message if one exists, falling back to a generic message if it doesn't. Thanks again!

paoloantinori:
Thanks @hussainjamal760! The changes look good — clearing `actionError` at the start of each handler avoids the stale-error issue, and the `error?.message || <fallback>` pattern handles both typed and opaque rejections cleanly. The `useEffect` also resetting `actionError` on fetch is a nice touch.

LGTM from my side. 🚢

hussainjamal760:
> Thanks @hussainjamal760! The changes look good — clearing `actionError` at the start of each handler avoids the stale-error issue, and the `error?.message || <fallback>` pattern handles both typed and opaque rejections cleanly. The `useEffect` also resetting `actionError` on fetch is a nice touch.
> 
> LGTM from my side. 🚢

Thank you so much for the review and for taking the time to verify the changes! I really appreciate your feedback and the helpful suggestions throughout the review process. Glad to hear the updates address the concerns. 🚀

EricWittmann (approved):
LGTM

PR #8838, opened 2026-07-23
feat(a2a): add full-text search for agent discovery

### LINKED ISSUE #7230: [Feature] Semantic Search for Agent Discovery
## Parent Epic
Part of #6991: AI Agent Registry

## Priority: P3 — Requires Validation

> **Prerequisite:** Before implementing, gather concrete evidence that label-based search is insufficient for real-world agent discovery. Run user research or collect feedback from early adopters. Embedding-based search adds significant infrastructure requirements (vector DB or pgvector, embedding model) that should be justified by demonstrated need.
>
> **Recommendation:** Start with Phase 1 (enhanced label search with fuzzy matching) which requires no new infrastructure. Only proceed to Phase 2 (embeddings) if Phase 1 proves insufficient.

## Background
The current agent discovery uses label-based search, which requires exact matches on skill IDs, capability names, etc. For a truly intelligent agent registry, users should be able to search using natural language descriptions.

## Problem Statement
Current search limitations:
- Exact match only: "sentiment-analysis" won't find "analyze-sentiment"
- No semantic understanding: "analyze customer feelings" won't find sentiment agents
- No similarity ranking: all matches are equal
- Limited to indexed labels

### Phase 1: Enhanced Label Search ← START HERE
- Fuzzy matching on skill names
- Description full-text search (PostgreSQL `tsvector`/`tsquery`)
- Synonym support via configurable mappings
- **No new infrastructure required**

#### Semantic Search Endpoint

```
GET /.well-known/agents/semantic-search?q=I need an agent that can translate text to Spanish

Response:
{
  "results": [
    {
      "agent": "translator-agent",
      "score": 0.95,
      "matchReason": "Skill 'translate' matches 'translate text', supports Spanish"
    },
    {
      "agent": "language-processor",
      "score": 0.72,
      "matchReason": "Has translation capability in description"
    }
  ]
}
```

#### Implementation Options

**Option A: Embedding-Based Search**
- Generate embeddings for agent descriptions, skills, capabilities
- Store embeddings alongside artifacts (pgvector or external vector DB)
- Use vector similarity for search
- Requires: embedding model (local or API)

**Option B: LLM-Assisted Search**
- Use LLM to interpret search query
- Generate structured filters from natural language
- Execute traditional search with generated filters
- Requires: LLM integration (existing MCP infrastructure)

**Option C: Keyword Enhancement** ← LIGHTEST WEIGHT
- Expand search terms with synonyms
- Use stemming and fuzzy matching
- No ML required

### Configuration

```yaml
apicurio:
  a2a:
    semantic-search:
      enabled: true
      provider: ollama  # ollama | openai | local
      model: nomic-embed-text
      similarity-threshold: 0.7
      index-on-create: true
```

### Phase 3: Skill Ontology (Future Enhancement)

Define relationships between skills for better matching:

```json
{
  "skills": {
    "sentiment-analysis": {
      "synonyms": ["emotion-detection", "feeling-analysis"],
      "related": ["text-analysis", "nlp"],
      "parent": "text-processing"
    }
  }
}
```

### Phase 1 (start here)
- [ ] Fuzzy matching on skill/capability names
- [ ] Full-text search on agent descriptions
- [ ] Configurable synonym mappings
- [ ] Fallback-compatible with existing exact-match search

### Phase 2 (if validated)
- [ ] Semantic search endpoint accepts natural language queries
- [ ] Results include relevance scores
- [ ] Configuration supports multiple embedding providers
- [ ] Performance acceptable for typical query loads

## Technical Considerations
- Phase 1: PostgreSQL full-text search (already available in the stack)
- Phase 2: Embedding storage — pgvector extension vs external vector DB
- Phase 2: Model selection — accuracy vs latency trade-offs
- Phase 2: Privacy — on-premise vs cloud embedding options
- **Storage impact**: Embeddings are ~1.5KB per agent per dimension

## References
- #7136 - JSON-LD/Protocol Indexing (related discovery enhancement)
- [Ollama Embeddings](https://ollama.ai/blog/embedding-models)
- [pgvector for PostgreSQL](https://github.com/pgvector/pgvector)
- Existing MCP Server can potentially assist with search

carlesarnal:
## Summary

- Add PostgreSQL-native full-text search (`tsvector`/`websearch_to_tsquery`) to agent discovery endpoints (`POST /.well-known/agents/search`, `GET /.well-known/agents`)
- On PostgreSQL, searches agent name, description, AND content blob (skills, tags, capabilities embedded in the agent card JSON)
- On H2/MySQL/MSSQL, falls back to case-insensitive `LIKE` on name + description
- No schema changes — all query-time computation on existing columns

## What changed

**SQL layer** — `SqlStatements.fullTextSearchClause()` returns DB-specific SQL: PostgreSQL uses `to_tsvector`/`websearch_to_tsquery` with an EXISTS subquery on the content table for deep matching; other databases use `ILIKE` on artifact name + description. The `content` filter type in `SqlSearchRepository` now works natively instead of throwing `ContentSearchNotSupportedException`.

**REST layer** — `WellKnownResourceImpl` switches the query field from `SearchFilter.ofName()` (wildcard LIKE on artifactId/name) to `SearchFilter.ofContent()` (full-text search). Both POST and GET agent search endpoints are updated.

**Tests** — Unit tests (H2 path) verify name, description, partial match, and no-match search. Integration tests (`AgentSearchIT`, tagged `@Tag(SMOKE)`) verify the same scenarios against PostgreSQL in CI. A2A experimental features enabled in minikube deployment configs.

Closes #7230 (Phase 1)

## Test plan

- [x] `./mvnw test -pl app -Dtest=WellKnownResourceTest,SearchVersionsTest` — all pass (H2/ILIKE path)
- [x] `./mvnw checkstyle:check` — 0 violations on all changed files
- [ ] CI: `storage: postgresql` (`remote-sql`) — exercises tsvector/tsquery path
- [ ] CI: `storage: h2` (`remote-mem`) — exercises ILIKE fallback
- [ ] CI: `storage: kafkasql` (`remote-kafka`) — delegates to SQL

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
@@ -100,4 +100,9 @@ public class PostgreSQLSqlStatements extends CommonSqlStatements {
         return "SELECT CASE WHEN pg_advisory_unlock(1886352239) THEN 1 ELSE 0 END";
     }
 
+    @Override
+    public String fullTextSearchClause() {
+        return "to_tsvector('english', COALESCE(a.name, '') || ' ' || COALESCE(a.description, '')) @@ websearch_to_tsquery('english', ?)";
+    }
+
 }
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
@@ -100,4 +100,22 @@ public class PostgreSQLSqlStatements extends CommonSqlStatements {
         return "SELECT CASE WHEN pg_advisory_unlock(1886352239) THEN 1 ELSE 0 END";
     }
 
+    @Override
+    public String fullTextSearchClause() {
+        return """
+                (to_tsvector('english', COALESCE(a.name, '') || ' ' || COALESCE(a.description, '')) \
+                @@ websearch_to_tsquery('english', ?) \
+                OR EXISTS ( \
+                SELECT 1 FROM versions fts_v JOIN content fts_c ON fts_c.contentId = fts_v.contentId \
+                WHERE fts_v.groupId = a.groupId AND fts_v.artifactId = a.artifactId \
+                AND safe_convert_to_text(fts_c.content) IS NOT NULL \
+                AND to_tsvector('english', safe_convert_to_text(fts_c.content)) \
+                @@ websearch_to_tsquery('english', ?)))""";
+    }
+
+    @Override
+    public boolean supportsNativeFullTextSearch() {
+        return true;
+    }
+
 }
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
@@ -8,7 +8,6 @@ import io.apicurio.registry.storage.dto.SearchFilter;
 import io.apicurio.registry.storage.dto.SearchedArtifactDto;
 import io.apicurio.registry.storage.dto.SearchedVersionDto;
 import io.apicurio.registry.storage.dto.VersionSearchResultsDto;
-import io.apicurio.registry.storage.error.ContentSearchNotSupportedException;
 import io.apicurio.registry.storage.error.RegistryStorageException;
 import io.apicurio.registry.storage.impl.sql.HandleFactory;
 import io.apicurio.registry.storage.impl.sql.SqlStatements;
@@ -33,10 +32,6 @@ import static io.apicurio.registry.storage.impl.sql.RegistryContentUtils.normali
  * Extracted from AbstractSqlRegistryStorage to improve maintainability.
  */
 public class SqlSearchRepository {
-
-    private static final String CONTENT_SEARCH_UNSUPPORTED_MESSAGE =
-            "Content search requires the search index, which is not enabled. "
-            + "Enable the search index to use content search.";
 
     private final Logger log;
 
@@ -167,7 +162,10 @@ public class SqlSearchRepository {
                         where.append(")");
                         break;
                     case content:
-                        throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                        where.append(sqlStatements.fullTextSearchClause());
+                        String artContentVal = "%" + filter.getStringValue() + "%";
+                        binders.add((query, idx) -> query.bind(idx, artContentVal));
+                        break;
                     default:
                         throw new RegistryStorageException("Filter type not supported: " + filter.getType());
                 }
@@ -346,7 +344,10 @@ public class SqlSearchRepository {
                         where.append(")");
                         break;
                     case content:
-                        throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                        where.append(sqlStatements.fullTextSearchClause());
+                        String verContentVal = "%" + filter.getStringValue() + "%";
+                        binders.add((query, idx) -> query.bind(idx, verContentVal));
+                        break;
                     default:
                         throw new RegistryStorageException("Filter type not supported: " + filter.getType());
                 }
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
@@ -8,7 +8,6 @@ import io.apicurio.registry.storage.dto.SearchFilter;
 import io.apicurio.registry.storage.dto.SearchedArtifactDto;
 import io.apicurio.registry.storage.dto.SearchedVersionDto;
 import io.apicurio.registry.storage.dto.VersionSearchResultsDto;
-import io.apicurio.registry.storage.error.ContentSearchNotSupportedException;
 import io.apicurio.registry.storage.error.RegistryStorageException;
 import io.apicurio.registry.storage.impl.sql.HandleFactory;
 import io.apicurio.registry.storage.impl.sql.SqlStatements;
@@ -32,10 +31,6 @@ import static io.apicurio.registry.storage.impl.sql.RegistryContentUtils.normali
  * Extracted from AbstractSqlRegistryStorage to improve maintainability.
  */
 public class SqlSearchRepository {
-
-    private static final String CONTENT_SEARCH_UNSUPPORTED_MESSAGE =
-            "Content search requires the search index, which is not enabled. "
-            + "Enable the search index to use content search.";
 
     private final Logger log;
 
@@ -166,7 +161,16 @@ public class SqlSearchRepository {
                         where.append(")");
                         break;
                     case content:
-                        throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                        where.append(sqlStatements.fullTextSearchClause());
+                        String artContentSearch = filter.getStringValue();
+                        if (sqlStatements.supportsNativeFullTextSearch()) {
+                            binders.add((query, idx) -> query.bind(idx, artContentSearch));
+                            binders.add((query, idx) -> query.bind(idx, artContentSearch));
+                        } else {
+                            binders.add((query, idx) -> query.bind(idx,
+                                    "%" + artContentSearch + "%"));
+                        }
+                        break;
                     default:
                         throw new RegistryStorageException("Filter type not supported: " + filter.getType());
                 }
@@ -345,7 +349,16 @@ public class SqlSearchRepository {
                         where.append(")");
                         break;
                     case content:
-                        throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                        where.append(sqlStatements.fullTextSearchClause());
+                        String verContentSearch = filter.getStringValue();
+                        if (sqlStatements.supportsNativeFullTextSearch()) {
+                            binders.add((query, idx) -> query.bind(idx, verContentSearch));
+                            binders.add((query, idx) -> query.bind(idx, verContentSearch));
+                        } else {
+                            binders.add((query, idx) -> query.bind(idx,
+                                    "%" + verContentSearch + "%"));
+                        }
+                        break;
                     default:
                         throw new RegistryStorageException("Filter type not supported: " + filter.getType());
                 }
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java
@@ -19,11 +19,7 @@ import org.junit.jupiter.api.Test;
 import java.util.Map;
 
 import static io.restassured.RestAssured.given;
-import static org.hamcrest.Matchers.allOf;
-import static org.hamcrest.Matchers.containsString;
-import static org.hamcrest.Matchers.containsStringIgnoringCase;
 import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.not;
 
 @QuarkusTest
 public class SearchVersionsTest extends AbstractResourceTestBase {
@@ -449,31 +445,22 @@ public class SearchVersionsTest extends AbstractResourceTestBase {
     }
 
     @Test
-    public void testSearchVersionsByContentFilterWithoutIndexIsBadRequest() throws Exception {
-        // The 'content' filter is served only by the Elasticsearch search index. With the index
-        // disabled (the default), it must return a 400 (client error) with a clear message, not a
-        // 500 that leaks the underlying SQL statement.
+    public void testSearchVersionsByContentFilterWithoutIndex() throws Exception {
         given().when()
                 .queryParam("content", "anything")
                 .get("/registry/v3/search/versions")
                 .then()
-                .statusCode(400)
-                .body("detail", allOf(containsString("search index"),
-                        not(containsStringIgnoringCase("select"))));
+                .statusCode(200);
     }
 
     @Test
-    public void testSearchVersionsByContentFilterWithOtherFiltersIsBadRequest() throws Exception {
-        // The content filter must be rejected even when combined with other valid filters, so the
-        // exception is thrown regardless of which other filters are present in the request.
+    public void testSearchVersionsByContentFilterWithOtherFilters() throws Exception {
         given().when()
                 .queryParam("groupId", "some-group")
                 .queryParam("content", "anything")
                 .get("/registry/v3/search/versions")
                 .then()
-                .statusCode(400)
-                .body("detail", allOf(containsString("search index"),
-                        not(containsStringIgnoringCase("select"))));
+                .statusCode(200);
     }
 
     @Test
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java
@@ -16,11 +16,7 @@ import org.junit.jupiter.api.Test;
 import java.util.Map;
 
 import static io.restassured.RestAssured.given;
-import static org.hamcrest.Matchers.allOf;
-import static org.hamcrest.Matchers.containsString;
-import static org.hamcrest.Matchers.containsStringIgnoringCase;
 import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.not;
 
 @QuarkusTest
 public class SearchVersionsTest extends AbstractResourceTestBase {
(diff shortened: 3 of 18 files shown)


app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
```diff
@@ -100,4 +100,21 @@ public String releaseInitLock() {
...
 
+    @Override
+    public String fullTextSearchClause() {
+        return """
+                (to_tsvector('english', COALESCE(a.name, '') || ' ' || COALESCE(a.description, '')) \
+                @@ websearch_to_tsquery('english', ?) \
+                OR EXISTS ( \
+                SELECT 1 FROM versions fts_v JOIN content fts_c ON fts_c.contentId = fts_v.contentId \
+                WHERE fts_v.groupId = a.groupId AND fts_v.artifactId = a.artifactId \
+                AND to_tsvector('english', convert_from(fts_c.content, 'UTF8')) \
```
paoloantinori: `convert_from(fts_c.content, 'UTF8')` assumes the content bytes are valid UTF-8. Since the `ContentSearchNotSupportedException` was removed, the generic `/search/versions?content=...` and `/search/artifacts?content=...` endpoints now reach this clause on PostgreSQL, so a search touching any binary artifact (compiled protobuf, avro binary, etc.) will raise `invalid byte sequence for encoding "UTF8"` and surface as a 500. Agent-only search is shielded by the `AGENT_CARD` type filter, but the general endpoints aren't. Worth either restricting to text content types or treating decode failures as no-match (e.g. a `BEGIN ... EXCEPTION WHEN others` wrapper, or `pg_input_is_valid` on PG 16+).

app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
```diff
@@ -100,4 +100,21 @@ public String releaseInitLock() {
         return "SELECT CASE WHEN pg_advisory_unlock(1886352239) THEN 1 ELSE 0 END";
     }
 
+    @Override
+    public String fullTextSearchClause() {
+        return """
+                (to_tsvector('english', COALESCE(a.name, '') || ' ' || COALESCE(a.description, '')) \
+                @@ websearch_to_tsquery('english', ?) \
+                OR EXISTS ( \
+                SELECT 1 FROM versions fts_v JOIN content fts_c ON fts_c.contentId = fts_v.contentId \
```
paoloantinori: This correlated `EXISTS` scans `versions JOIN content` and computes `to_tsvector(...)` on the fly for each candidate artifact, with no GIN index or materialized tsvector column. Fine for an MVP, but at scale it's effectively a per-query full scan of the content table. If this is meant to replace the Elasticsearch-backed content search long term, a follow-up adding a generated tsvector column + GIN index would make it production-safe.

paoloantinori (commented):
Solid feature -- exposing native PostgreSQL full-text search for agent discovery with a sensible non-PG fallback is a clean approach, and the test split (H2 unit tests + SMOKE-tagged integration tests on PostgreSQL) is well structured. Two concerns worth addressing before this lands, both concentrated in the PostgreSQL full-text clause: a latent crash on binary content and a scalability gap.

- **`convert_from(..., 'UTF8')` on the content column can throw on binary artifacts.** The `content` filter is now enabled globally in `SqlSearchRepository` (the old `ContentSearchNotSupportedException` is gone), so the generic `/registry/v3/search/versions?content=...` and `/search/artifacts?content=...` endpoints will reach this clause on PostgreSQL -- not just agent search. If `fts_c.content` is stored as a binary column (which I believe is the case for the registry), `convert_from` raises `ERROR: invalid byte sequence for encoding "UTF8"` for any non-text artifact (compiled protobuf, avro binary, etc.), turning a search request into a 500. Agent-only search is safe (the outer `artifactType = AGENT_CARD` filter limits candidates), but the generic endpoints are not. Worth double-checking the column encoding and guarding -- e.g. restrict to text content types or treat decode failures as no-match.

- **Per-query tsvector computation with no index.** The correlated `EXISTS` subquery computes `to_tsvector(convert_from(...))` over every content row for each candidate artifact, with no GIN index or materialized tsvector column to back it. Fine for an MVP / small registries, but it degrades toward O(candidates x content rows) per query at scale. If this is intended as the long-term path (vs. the Elasticsearch index), a follow-up adding a generated tsvector column + GIN index would make it production-safe; worth confirming the expected dataset size.

- Minor: the same `content` query returns different result sets by backend -- PostgreSQL matches the content blob too, while the H2/MySQL/MSSQL fallback only matches name + description. Consistent with the PR description, but callers can't rely on content-blob matching portably, so a note in the API docs would help set expectations.

paoloantinori (commented):
Re-reading at the current head — the native PostgreSQL full-text path with a non-PG fallback is still the right design, and the H2 unit / SMOKE integration split is sensible. A few points worth confirming before this lands, all on the SQL layer:

- **Scope is broader than agent discovery.** In `SqlSearchRepository`, the `case content:` branch now appends `fullTextSearchClause()` in *both* `searchArtifacts` and `searchVersions`, so the `content` filter executes for every artifact/version search, not just agent cards. Previously, without the Elasticsearch index, it threw `ContentSearchNotSupportedException` (400); now it returns 200 with SQL FTS/LIKE results. The two `SearchVersionsTest` cases were flipped from expecting 400 to 200 to match. That's a public search-API behavior change that goes beyond the PR title ("agent discovery") — worth confirming it's intended for this phase and that nothing depends on the old 400.

- **`convert_from(..., 'UTF8')` will raise on non-text content (strongest concern).** In `PostgreSQLSqlStatements.fullTextSearchClause()`, the EXISTS subquery does `to_tsvector('english', convert_from(fts_c.content, 'UTF8'))`. PostgreSQL raises `invalid byte sequence for encoding "UTF8"` on any content blob that isn't valid UTF-8. Because the clause now runs across *all* artifacts/versions (not just text agent cards), a single binary or non-UTF-8 artifact in the registry would turn the whole search query into a 500. Worth guarding — e.g., restricting the content scan to text `contentType`s, or normalizing/skipping non-decodable rows. Worth double-checking against how binary artifacts are actually stored in the `content` table.

- **No index → O(N) per query.** The tsvector is computed on the fly via a correlated EXISTS over `versions`+`content` on every search, with no GIN index. Fine for small registries, and I see the PR intentionally avoids schema changes (Phase 1), but it's worth opening a follow-up (e.g. a generated `tsvector` column + GIN index) before this is expected to scale, otherwise content search will degrade linearly with registry size.

- **The `searchVersions` content path has no PG coverage.** `AgentSearchIT` only exercises the agent endpoints (the artifact-search path). The version-search content path on PostgreSQL — where `fullTextSearchClause()` references `a.name`/`a.description` — is untested. Worth confirming the versions query exposes those columns under the `a` alias, or `/registry/v3/search/versions?content=...` could error on PG even when the H2 unit test passes.

jsenko (commented):

SQL implementation is solid. Parameters are properly bound (no injection risk), tsvector/ILIKE fallback is correct, and the AGENT_CARD type filter shields agent search from paoloantinori's binary content concern. Good integration test coverage.

### 1. `name` parameter now silently does full-text search (`WellKnownResourceImpl.java:336`)

The `GET /.well-known/agents?name=X` parameter changed from `SearchFilter.ofName()` (name-only matching with wildcards) to `SearchFilter.ofContent()` (full-text across name + description + content). This means a query like `?name=weather` now returns agents whose description or JSON body mentions "weather," not just those named "weather."

This is additive (callers get more results, not fewer), but it changes the semantic contract of the `name` parameter. Consider either:
- Renaming the parameter to `query` to match the POST endpoint's `AgentSearchRequest.query` field
- Or keeping `name` for name-only and adding a separate `query` parameter for full-text

### 2. Inconsistent search behavior between agent and MCP tool endpoints

`GET /.well-known/agents?name=X` uses `SearchFilter.ofContent()` while `GET /.well-known/mcp-tools?name=X` uses `SearchFilter.ofName()` (line 538). Same parameter name, different semantics across sibling endpoints. If the intent is to add full-text search to agents only, document this difference in the API spec.

### 3. Non-PostgreSQL fallback doesn't search content

The ILIKE fallback searches only name + description (no content table join). This means full-text content search only works on PostgreSQL. The fallback behavior is correctly documented in the PR description but worth noting in the endpoint documentation for users running H2/MySQL/MSSQL.

app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
```diff
@@ -100,4 +100,24 @@ public String releaseInitLock() {
...
+                WHERE fts_v.groupId = a.groupId AND fts_v.artifactId = a.artifactId \
+                AND (fts_c.contentType LIKE 'application/json%' OR fts_c.contentType LIKE 'application/%+json%' \
+                OR fts_c.contentType LIKE 'text/%' OR fts_c.contentType LIKE 'application/%+xml%' \
+                OR fts_c.contentType LIKE 'application/%+yaml%' OR fts_c.contentType LIKE 'application/yaml%') \
+                AND to_tsvector('english', convert_from(fts_c.content, 'UTF8')) \
+                @@ websearch_to_tsquery('english', ?)))""";
+    }
```
paoloantinori: `convert_from(fts_c.content, 'UTF8')` will raise `invalid byte sequence for encoding "UTF8"` if any content row matching the contentType filter contains non-UTF8 bytes. Since `contentType` is client-supplied and not byte-validated, a mislabeled binary blob can poison the whole query with a 500. Consider guarding with `pg_input_is_valid` (PG 16+) or wrapping in a `BEGIN ... EXCEPTION WHEN OTHERS THEN NULL` plpgsql helper so a bad row is skipped rather than fatal.

app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
```diff
@@ -100,4 +100,24 @@ public String releaseInitLock() {
         return "SELECT CASE WHEN pg_advisory_unlock(1886352239) THEN 1 ELSE 0 END";
     }
 
+    @Override
+    public String fullTextSearchClause() {
+        return """
+                (to_tsvector('english', COALESCE(a.name, '') || ' ' || COALESCE(a.description, '')) \
+                @@ websearch_to_tsquery('english', ?) \
+                OR EXISTS ( \
+                SELECT 1 FROM versions fts_v JOIN content fts_c ON fts_c.contentId = fts_v.contentId \
```
paoloantinori: This correlated subquery re-tokenizes every matching content blob via `to_tsvector(...)` on each request — there's no `GIN` index to short-circuit the match. For a non-trivial registry this is a full scan of the `content` table per search. A `GENERATED ALWAYS AS (to_tsvector('english', convert_from(content,'UTF8'))) STORED` column + `GIN` index would make this sustainable. If that's a follow-up, it might be worth noting in the PR description or a linked issue so the scaling ceiling is visible.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java
```diff
@@ -354,31 +350,22 @@ public void testSearchVersionsByArtifactIdWildcard() throws Exception {
...
-                .body("detail", allOf(containsString("search index"),
-                        not(containsStringIgnoringCase("select"))));
+                .statusCode(200);
     }
 
     @Test
-    public void testSearchVersionsByContentFilterWithOtherFiltersIsBadRequest() throws Exception {
-        // The content filter must be rejected even when combined with other valid filters, so the
-        // exception is thrown regardless of which other filters are present in the request.
+    public void testSearchVersionsByContentFilterWithOtherFilters() throws Exception {
```
paoloantinori, line 457: This changes the public contract from `400` ("content search requires the search index") to `200`. The old test also asserted the error didn't leak SQL. Worth confirming that when the Elasticsearch search index *is* enabled, the index path takes precedence over this new SQL path (no double-processing), and that the REST API docs are updated to reflect content search now works without the index on PostgreSQL.

paoloantinori (commented):
Re-reading at the current head (73eab74). The native PostgreSQL full-text path with a DB-engine fallback is a clean design and the parameter binding is correct (already confirmed by @jsenko). Three points I think are worth addressing before this lands, the most important first:

- **`convert_from(fts_c.content, 'UTF8')` can throw on non-UTF8 content.** The `content` table stores raw bytes for every artifact type. The `contentType LIKE` filter narrows to text-like types, but `contentType` is client-provided metadata — not validated against the actual bytes. A single row mislabeled as `application/json` containing binary bytes will make `convert_from` raise, failing the *entire* search query with a 500 rather than just skipping that row. Worth confirming whether a `pg_input_is_valid` guard (PG 16+) or a `BEGIN ... EXCEPTION WHEN OTHERS` wrapper is feasible, or at least documenting the assumption.

- **No index on the content `tsvector` → full re-tokenization per query.** `to_tsvector('english', convert_from(...))` is computed fresh for every matching content row on every search request. On a registry with thousands of artifacts this is O(n) over the content table per request. A generated column (`GENERATED ALWAYS AS (to_tsvector(...)) STORED`) with a `GIN` index would make this sustainable at scale. If that's out of scope for this PR, it's worth filinging a follow-up so the scaling ceiling is tracked.

- **Silent API contract change on `/search/versions?content=`.** The two `SearchVersionsTest` cases previously asserted `400` with a message directing clients to enable the search index (`"Content search requires the search index..."`). They now assert `200`. This is intentional per the PR, but it changes the public contract: clients that relied on the 400 to detect the unindexed state will now get a silent 200 with partial results. Worth confirming (a) whether the SQL path and the Elasticsearch-index path can both be active and which wins, and (b) whether the REST API docs / error catalogue should be updated to reflect that content search no longer requires the index on PostgreSQL.

app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
carlesarnal: Agreed — removed the EXISTS content subquery entirely. The PostgreSQL path now only searches name+description via `to_tsvector`/`websearch_to_tsquery`, which eliminates the `convert_from` crash risk. Content blob search (skills, tags in JSON) is deferred to a follow-up with a safe UTF-8 conversion function + GIN index.

app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
carlesarnal: Addressed — removed the content table scan entirely. The PostgreSQL clause now only operates on `a.name` and `a.description` (existing indexed columns), so there's no per-query content table scan. Content blob search with a generated tsvector column + GIN index is a natural follow-up.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java
carlesarnal: Good point. When Elasticsearch is enabled, the `ElasticsearchSearchDecorator` intercepts `searchVersions()` and routes `content`-filtered queries to ES — no double processing. The decorator does not intercept `searchArtifacts()` (used by agent discovery), so the SQL path is always the handler there. The 400→200 change is intentional — the `content` filter is now functional on all storage backends.

app/src/main/java/io/apicurio/registry/storage/impl/sql/PostgreSQLSqlStatements.java
```diff
@@ -100,4 +100,9 @@ public String releaseInitLock() {
         return "SELECT CASE WHEN pg_advisory_unlock(1886352239) THEN 1 ELSE 0 END";
     }
 
+    @Override
+    public String fullTextSearchClause() {
+        return "to_tsvector('english', COALESCE(a.name, '') || ' ' || COALESCE(a.description, '')) @@ websearch_to_tsquery('english', ?)";
```
paoloantinori: This clause builds the tsvector from `a.name` and `a.description` only — there is no join or EXISTS against the content table, so the agent-card content blob (skills, tags, capabilities) is not searchable. That contradicts the PR summary, which claims an EXISTS subquery on the content table and content-blob matching. If deep content matching is intended, the clause needs to include the content table; otherwise the summary should be corrected. The absence of any `AgentSearchIT` case that queries by a skill or tag lines up with content not being indexed.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -166,7 +161,13 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
                         where.append(")");
                         break;
                     case content:
-                        throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                        String artFtsClause = sqlStatements.fullTextSearchClause();
```
paoloantinori: The `content` filter now matches artifact `name`+`description` (metadata), not content. Since this is the shared SqlSearchRepository, the same change alters the generic `/registry/v3/search/artifacts?content=` and `/search/versions?content=` endpoints — the PR flips `SearchVersionsTest` from 400 to 200. A `content` query silently returning metadata matches is surprising; worth confirming this cross-cutting behavior change is intended, or scoping the new semantics to agent-card search only.

paoloantinori (commented):
The wiring is clean and the queries are properly parameterized, but the implementation only searches artifact name+description: it does not search the agent-card content blob the PR summary advertises, and the change to the shared `content` filter alters the generic search endpoints beyond the a2a scope.

- `PostgreSQLSqlStatements.java:105` — `fullTextSearchClause()` builds the tsvector from `a.name` and `a.description` only; there is no join/EXISTS to the content table, so skills/tags/capabilities embedded in the agent-card JSON are not searchable. This contradicts the summary, which claims an EXISTS subquery on the content table and content-blob matching. Notably, no `AgentSearchIT` case queries by a skill or tag, which is consistent with the content blob not actually being indexed.
- `SqlSearchRepository.java:164` and `:349` — both `searchArtifacts` and `searchVersions` `case content:` now match metadata (name+description) instead of throwing. Because this repository serves the generic `/registry/v3/search/artifacts?content=` and `/search/versions?content=` endpoints (the PR flips `SearchVersionsTest` from expecting 400 to 200), a `content` query now silently returns name/description matches rather than content. That is a behavior change outside the a2a scope and worth confirming as intended, or scoping to agent cards.
- `PostgreSQLSqlStatements.java:105` / `SqlStatements.java` default — both clauses compute the match at query time with no supporting index; on PostgreSQL `to_tsvector(...)` evaluates per row with no GIN/expression index, so large registries will full-scan on every search. Acceptable for Phase 1, flagging since the PR explicitly opts for no schema changes.

EricWittmann:
Hey @carlesarnal — I have a concern about the approach here.

Adding dialect-specific full-text search (PostgreSQL `tsvector` vs `LIKE` fallback on H2/MySQL/MSSQL) creates an inconsistent experience that varies by storage backend. PostgreSQL users get content-level search across the agent card JSON, while MySQL/MSSQL users only get name+description matching. That's hard to document, hard to support, and will confuse users who switch backends or deploy on a non-PostgreSQL database.

I think agent discovery full-text search should require the Elasticsearch indexing feature to be enabled. Full-text search across content blobs is exactly what ES is designed for, and it provides a consistent experience regardless of the underlying SQL dialect. Users who need deep agent search should enable ES — that's the intended purpose of that integration.

The structured content search from #8538 (which indexes specific elements like skills and capabilities into a SQL side table) already covers the most common agent discovery filtering use cases across all SQL variants. Between that and ES for full-text, I don't think we need a dialect-specific middle ground.

What do you think?

carlesarnal:
Hi @EricWittmann — you're right. Dialect-specific content search creates an inconsistent experience, and full-text search across content blobs is exactly what ES is designed for.

After narrowing the scope to name+description only (no content blob search), the remaining value — tsvector stemming vs ILIKE on the same two fields — isn't enough to justify the complexity or close #7230. The issue's acceptance criteria (fuzzy matching on skills/capabilities, full-text on descriptions, synonym support) are better served by ES for content search and #8538 for structured SQL filters.

Closing this PR. Will update #7230 to reflect that full-text agent search should go through the Elasticsearch index, with #8538 covering the SQL-native structured search path.

PR #8841, opened 2026-07-23
fix(schema-util): handle null content type in ContentTypeUtil.parseJsonOrYaml

### LINKED ISSUE #8768: bug: ContentTypeUtil.parseJsonOrYaml throws NPE on a null content type, so type auto-detection silently rejects valid content
## Description

**Persistence type**: any (the affected code is in `schema-util`, so it is independent of the persistence type)

While reading through the content type detection code I ran into what looks like a real bug in `ContentTypeUtil.parseJsonOrYaml`. It reads the content type and calls `toLowerCase()` on it without checking for null first, even though a few callers deliberately pass content through with a null content type.

Here is the method (`schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java`, around line 160):

```java
public static JsonNode parseJsonOrYaml(TypedContent content) throws IOException {
    JsonNode node = null;
    String contentType = content.getContentType();
    if (contentType.toLowerCase().contains("yaml") || contentType.toLowerCase().contains("yml")
            || contentType.equalsIgnoreCase("text/x-prompt-template")) {
        ...
```

If `content.getContentType()` returns null, that first `contentType.toLowerCase()` throws a `NullPointerException`.

The reason it actually bites is the content accepters. For example `ModelSchemaContentAccepter.acceptsContent`:

```java
String contentType = content.getContentType();
if (contentType != null && !isJsonOrYaml(contentType)) {
    return false;
}
JsonNode tree = ContentTypeUtil.parseJsonOrYaml(content);   // contentType can still be null here
```

That `contentType != null` guard looks intentional to me: it means "if there is no content type, do not reject it, just try to parse it and see." But `parseJsonOrYaml` then throws the NPE on the null, the accepter's surrounding `try { ... } catch (Exception e) { }` swallows it, and the method returns `false`. So a document that really is a valid Model Schema or Prompt Template, uploaded without a content type, gets quietly treated as "not this type" during auto-detection.

`PromptTemplateContentAccepter` follows the same guard-then-parse pattern, and the matching reference finders and validators (`ModelSchemaReferenceFinder`, `PromptTemplateReferenceFinder`, `ModelSchemaContentValidator`, `PromptTemplateContentValidator`) reach the same method the same way.

### Environment

Not environment specific. The bug is in the `schema-util/common` content type utility and is reachable from the normal artifact type auto-detection path whenever content is submitted without a content type.

### Steps to Reproduce

1. Take a valid Model Schema (or Prompt Template) document.
2. Submit it so that artifact type auto-detection runs, without setting a content type on the content.
3. The type is not detected, even though the same content is accepted when a content type is provided.

Expected: a null content type is treated as unknown, and the content is still parsed (as JSON by default), which is what the callers seem to intend with their `contentType != null` guard.

Actual: `parseJsonOrYaml` throws a `NullPointerException` on the null content type, the accepter swallows it and returns `false`, and valid content is silently rejected during type detection.

### Logs

Nothing shows up by default, which is part of what makes this easy to miss. The NPE is caught and discarded inside the accepters' `catch (Exception e)` blocks, so it never surfaces. It can be confirmed by calling `parseJsonOrYaml` with a `TypedContent` whose content type is null.

### Additional context

PR #8558 is currently open against the same file, but it changes `determineContentType` and adds GraphQL detection rather than the null handling in `parseJsonOrYaml`, so this is a sepaarte fix.

PRAteek-singHWY:
# fix(schema-util): handle null content type in ContentTypeUtil.parseJsonOrYaml

## Summary

Fixes #8768.

`ContentTypeUtil.parseJsonOrYaml` throws a `NullPointerException` when the content it
receives has no content type. Several callers deliberately pass content through with a
null content type, expecting it to be parsed anyway, so this bug quietly breaks artifact
type auto-detection: a valid Model Schema or Prompt Template uploaded without a content
type gets silently treated as "not this type".

This PR null-guards the content type so a null is treated as unknown and falls back to
JSON parsing — which is exactly what the callers already assume.

## Root Cause

`parseJsonOrYaml` reads the content type and immediately calls `toLowerCase()` on it
before doing any null check:

```java
String contentType = content.getContentType();
if (contentType.toLowerCase().contains("yaml") || ...) {   // NPE when contentType is null
```

The content accepters, on the other hand, let a null content type through on purpose.
For example `ModelSchemaContentAccepter.acceptsContent`:

```java
String contentType = content.getContentType();
if (contentType != null && !isJsonOrYaml(contentType)) {
    return false;                                          // only rejects a *known* wrong type
}
JsonNode tree = ContentTypeUtil.parseJsonOrYaml(content); // contentType can still be null here
```

That `contentType != null` guard means "if there is no content type, don't reject it —
just try to parse it and see." But `parseJsonOrYaml` then throws the NPE on the null,
the accepter's surrounding `try { ... } catch (Exception e) { }` swallows it, and the
method returns `false`. So the content is dropped during auto-detection, and because the
NPE is caught and discarded, nothing shows up in the logs — which is part of what makes
this easy to miss.

`PromptTemplateContentAccepter` follows the same guard-then-parse pattern, and the
matching reference finders and validators (`ModelSchemaReferenceFinder`,
`PromptTemplateReferenceFinder`, `ModelSchemaContentValidator`,
`PromptTemplateContentValidator`) reach the same method the same way, so they are all
affected.

## Changes

- `schema-util/common/.../content/util/ContentTypeUtil.java`
  - `parseJsonOrYaml`: added a `contentType != null` guard around the content-type
    checks. A null content type now falls through to the JSON branch (parse as JSON by
    default) instead of throwing an NPE — matching what the callers already assume.
  - Added `Locale.ROOT` to the `toLowerCase()` calls in `parseJsonOrYaml` (and the one
    in `isTextPlain`), since locale-less `toLowerCase()` is disallowed by the project
    code style.
  - Added the `java.util.Locale` import.

- `schema-util/common/.../content/util/ContentTypeUtilTest.java`
  - Added three test cases (the class previously only covered `isParsableXml`):
    - `testParseJsonOrYaml_nullContentType_parsesAsJson` — a null content type parses a
      JSON document without throwing, and the parsed value is correct.
    - `testParseJsonOrYaml_nullContentType_jsonObject` — a null content type with a
      multi-field JSON object returns the expected field values.
    - `testParseJsonOrYaml_yamlContentType` — an explicit YAML content type still routes
      through the YAML branch.

## Test plan

- [x] Added `testParseJsonOrYaml_nullContentType_parsesAsJson` — reproduces the NPE on
      `main` and passes with the fix.
- [x] Added `testParseJsonOrYaml_nullContentType_jsonObject` — asserts specific parsed
      values (`a == 1`, `b == 2`), not just non-null.
- [x] Added `testParseJsonOrYaml_yamlContentType` — confirms the YAML branch is unchanged.
- [x] `./mvnw checkstyle:check -pl schema-util/common` passes (0 violations).
- [ ] Full module test run (`./mvnw test -pl schema-util/common -am`) — please run in CI;
      the local sandbox could not resolve some upstream build artifacts.
- [ ] Storage variants: not applicable. The change is in the `schema-util/common`
      content-type utility and is independent of the persistence type.

## Notes

PR #8558 is currently open against the same file, but it changes `determineContentType`
and adds GraphQL detection rather than the null handling in `parseJsonOrYaml`, so this is
a separate, non-overlapping fix.


schema-util/common/src/main/java/io/apicurio/registry/content/util/ContentTypeUtil.java
```diff
@@ -160,8 +161,12 @@ public static JsonNode parseYaml(ContentHandle content) throws IOException {
     public static JsonNode parseJsonOrYaml(TypedContent content) throws IOException {
         JsonNode node = null;
         String contentType = content.getContentType();
-        if (contentType.toLowerCase().contains("yaml") || contentType.toLowerCase().contains("yml")
-                || contentType.equalsIgnoreCase("text/x-prompt-template")) {
+        // A null content type is treated as unknown: fall back to JSON parsing. The callers
+        // (content accepters, reference finders, validators) deliberately let a null content
+        // type through, so this must not throw on null.
+        if (contentType != null && (contentType.toLowerCase(Locale.ROOT).contains("yaml")
```
paoloantinori: With `contentType == null`, control falls into the `else` branch and is parsed as JSON only. For artifact types whose content is YAML-shaped (this method already treats `text/x-prompt-template` as YAML), a null-typed YAML document still fails detection. Worth deciding whether YAML-with-null-type is in scope — see the review body for a suggested JSON-then-YAML sniff.

paoloantinori (commented):
Clean fix for the NPE — the null-guard is correct, the `Locale.ROOT` hardening of the existing `isTextPlain` is a nice bonus, and the new tests cover the null, JSON, and YAML branches well.

One scope point worth confirming before this lands. The PR body frames the change as restoring auto-detection for *"a valid Model Schema or Prompt Template uploaded without a content type"*. The null branch, though, now unconditionally falls back to `parseJson`. Some of those artifact types are legitimately YAML-shaped — this same method already routes `text/x-prompt-template` through the YAML parser, and prompt templates are published as YAML elsewhere in the codebase. For a YAML document uploaded with a null content type, `parseJson` will throw and the accepter will *still* return false, so the silent mis-detection persists for the YAML case.

If YAML-with-null-content-type is a real scenario for these types, a JSON-then-YAML sniff in the null branch (try JSON, fall back to YAML on parse failure) would close the gap. If JSON-only is the intended contract for the null case, it would help to state that explicitly so the description framing matches the behavior. Either way, the NPE fix itself is good as-is.

PRAteek-singHWY:
Thanks for the close read @paoloantinori 

The null branch was falling through to parseJson only, so a YAML document uploaded without a content type would fail the JSON parse and the accepter would still return false, the same silent mis-detection, just moved from the NPE to the JSON case. And since this method already treats text/x-prompt-template as YAML, and prompt templates are published as YAML elsewhere, YAML-with-null-type is a real scenario, not a theoretical one. So JSON-only wasn't the right contract here.

I went with your first suggestion. The null branch now sniffs: it tries JSON first (the common case, and since JSON is a subset of YAML it stays the fast path), and falls back to YAML on a parse failure. That closes the YAML gap so the behavior now matches how the PR describes it, while keeping the NPE fix and the JSON fast path intact.

I also added a test for a YAML document with a null content type (testParseJsonOrYaml_nullContentType_yamlDocument), since the existing YAML test used an explicit content type and wouldn't have caught this.

Pushed as a follow-up commit. 

Thanks again for going through it.
Happy to iterate based on any further suggestions.
Thank you.

paoloantinori:
Good catch — I hadn't considered the `text/x-prompt-template` path and that prompt templates are published as YAML without a content type. The JSON-first-then-YAML fallback for the null branch makes sense given that real-world scenario, and the updated code + tests look solid. Thanks for the detailed explanation!

PRAteek-singHWY:
> Good catch — I hadn't considered the `text/x-prompt-template` path and that prompt templates are published as YAML without a content type. The JSON-first-then-YAML fallback for the null branch makes sense given that real-world scenario, and the updated code + tests look solid. Thanks for the detailed explanation!

Thanks for the review, This' great Learning.
Thank you.

PRAteek-singHWY:
Hi @paoloantinori @EricWittmann 
Thanks again for the review here, this has been great learning.

One small follow-up while going through it: working through this content-type path pulled me deeper into the prompt-template handling, and I noticed a separate consistency issue there, the validator, the REST `/render` endpoint, and the MCP converter each parse `{{variable}}` placeholders with a slightly different grammar, so something like `{{ name }}` (with surrounding spaces) isn't recognized the same way across them. I've written it up as its own issue so it stays independent of this fix: #8877.

I've genuinely enjoyed working in the prompt-template area, happy to keep contributing here.

Thank you.

paoloantinori:
Thanks for the kind words, and great observation — exactly the kind of "pull on a thread" finding that a careful review surfaces.

The placeholder-parsing inconsistency you describe (validator vs REST `/render` endpoint vs MCP converter each parsing `{{variable}}` with a slightly different grammar) is real and worth tracking: if `{{ name }}` (with surrounding spaces) is accepted by one component but not the others, you get a surprising failure mode where a "valid" template fails at render time, or vice versa. The right move here:

1. **Don't expand this PR** — the NPE fix is focused and correct; folding the placeholder-grammar unification into it would balloon the scope and slow down the merge.
2. **Open a separate issue** describing the divergence with a small repro (e.g. the same `{{ name }}` template run through all three paths) and citing the three code locations. That keeps it trackable and gives it proper visibility.

A practical suggestion for the canonical grammar: pin a single regex in one place (probably in schema-util, since it's already the shared layer) and have all three consumers use it — `{{\s*name\s*}}` with `\s*` on both sides covers both `{{name}}` and `{{ name }}`. But the key decision is to define *one* grammar and apply it everywhere; whether to admit internal whitespace is secondary.

Happy to take a closer look at the three paths once the issue is open — for now I just want to confirm the observation is sound and deserves to be tracked on its own.

PRAteek-singHWY:
Hi @paoloantinori @EricWittmann @carlesarnal, opening #8923 as a draft so the work is visible while #8841 is still in flight. I'll take it out of draft with `/ready` once that one merges and the review pipeline is free.

This is the `{{variable}}` parsing fix for #8877, following the direction Paolo gave on the #8841 thread: one canonical pattern pinned in the shared layer, with all three consumers routed through it so they can't drift apart again.

- New `PromptTemplateVariableUtil` in `schema-util/common` holding `{{\s*(\w+)\s*}}` plus extraction and substitution helpers.
- `PromptTemplateContentValidator`, `PromptRenderingService` (REST `/render`) and the MCP `PromptTemplateConverter` all delegate to it.
- The bug from the issue is closed: `{{ quality }}` used but not defined now fails `FULL` validation instead of passing silently and rendering with the placeholder left in the output.

Two decisions I'd rather surface now than after review starts, because either could change the shape of this:

1. **Tightening `/render` from `[^}]+` to `\w+` drops dotted-name support.** `{{user.name}}` was substituted by the render endpoint and is now literal text. That makes `/render` consistent with the validator and the MCP converter rather than the other way round, which is how I read the suggestion. Widening the shared pattern to admit dotted names everywhere is a one-line change if you'd prefer that direction instead.

2. **Sharing the pattern with the MCP converter adds a module dependency.** `mcp` previously depended only on the Java SDK and now also pulls `schema-util-common`. Duplicating the constant would defeat the point of the issue, but if you'd rather not grow the MCP server's classpath I can split the MCP half into a follow-up.

One incidental find: `PromptTemplateConverter` declared a `VARIABLE_PATTERN` that nothing ever used. The real substitution built a regex out of each argument name, so an argument called `a.b` would substitute an unrelated `{{axb}}` placeholder. That's fixed as a side effect here.

Tests: 12 new cases on the shared helper, 4 on the validator (including the whitespace repro from the issue, which fails on `main`), 4 on `/render`, and a new `PromptTemplateConverterTest` with 11 cases, since the MCP converter had no coverage at all before. `schema-util` and `mcp` pass locally. The `app` `@QuarkusTest` I've left for CI.

On #8841: if there's nothing further you'd like changed there, a `/merge` would let this one move forward. Happy to keep iterating on either if not.

Thank you.

paoloantinori:
Thanks for the heads-up @PRAteek-singHWY — opening #8923 as a draft to keep the placeholder-parsing work visible while this one finishes is exactly the right call. The direction looks aligned with what we discussed: one canonical `{{variable}}` pattern pinned in the shared schema-util layer, with all three consumers (validator, REST, accepter) routed through it so they can't drift again. I'll take a proper look at #8923 once it comes out of draft and the review pipeline frees up after this one merges — no need to block either PR on it in the meantime.

EricWittmann:
Waiting on a last CI run before merging this.   Sorry for the delay - we're working through the backlog as best we can.  :+1:

EricWittmann:
Thanks for the contribution @PRAteek-singHWY! Excellent root cause analysis and clean fix. Looking forward to the prompt template follow-up in #8923.

PRAteek-singHWY:
Thank you @EricWittmann, and thank you @paoloantinori for the guidance on the shape of
this one  pinning a single canonical pattern in the shared layer and routing every
consumer through it was a better direction than where I started, and tracing why the
three had drifted apart taught me a lot about how the schema-util layer fits together.
Genuinely great learning experience.

Small heads-up on #8923: it was closed automatically by the one-PR-at-a-time bot a few
seconds after I opened it, since this PR was still open at the time. Now that this one
has merged I'll reopen it and take it out of draft with `/ready`.

Looking forward to the review, and to picking up more work here. 

Thank you.

PRAteek-singHWY:
Hi @EricWittmann Quick update to my note above: GitHub would not let me reopen #8923 after all. It and
#8921 were both opened from the same branch against `main`, and the API declines to reopen
a PR when another one shares that head/base pair, so there is no `/ready` path back to it.

I have opened #8977 instead, same branch, same commit, rebased onto current `main` so it
includes this fix and the 3.3.1 release. It is open as a normal PR rather than a draft this
time, since the reason for the draft was this PR being in flight, and that is now resolved.

@paoloantinori, #8977 is the follow-up you mentioned wanting to look at once the review
pipeline freed up.

PR #8845, opened 2026-07-23
fix(rest): return 404 instead of 500 for a missing version comment

### LINKED ISSUE #8843: Deleting or updating a non-existent version comment returns HTTP 500 instead of 404
## Description

**Registry Version**: snapshot (build from main)
**Persistence type**: sql (the behaviour is not storage specific, it happens on any persistence type)

When you delete or update a version comment using a comment ID that does not exist, the REST API responds with HTTP 500 (Internal Server Error). Since the comment simply is not there, the correct response is HTTP 404 (Not Found).

The root cause is in how exceptions are mapped to status codes. Internally the code throws CommentNotFoundException, which extends NotFoundException (that parent normally maps to 404). But the status code lookup in HttpStatusCodeMap matches exceptions by their exact class, and CommentNotFoundException was never added to that map. Every other not found exception in the project (artifact, version, group, branch, and so on) has its own entry, so this one looks like a simple missed registration. Because the exact class is not found, the lookup falls back to the default of 500.

### Environment

Reproduced with a local build from main running against a SQL backend. It reproduces the same way regardless of how you call the API (REST client, curl, or a generated SDK), because the problem is purely in the server side status code mapping.

### Steps to Reproduce

1. Create a group, an artifact, and a version so that a real version exists.
2. Send a request against a comment ID that does not exist, for example 999999:
   - DELETE /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{version}/comments/999999
   - or PUT the same path with a comment body to update it.
3. Observe the response code.

Result: the server returns 500.
Expected: the server should return 404.

Expected: a request for a comment that does not exist returns 404 Not Found, the same as other missing resources in the API.

Actual: the request returns 500 Internal Server Error, which wrongly suggests the server is broken rather than telling the client the comment was not found.

### Logs

Access log lines for the two failing calls (comment 999999 does not exist):

DELETE /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{version}/comments/999999 -> response_code 500
PUT    /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{version}/comments/999999 -> response_code 500

The audit log confirms the underlying reason: error_msg "No comment with ID 999999 was found."

parthdagia05:
## What

Deleting or updating a version comment that does not exist now returns HTTP 404 (Not Found) instead of HTTP 500 (Internal Server Error).

Fixes #8843 

## Why

Internally these operations throw CommentNotFoundException, which extends NotFoundException and is meant to map to 404. The status code lookup in HttpStatusCodeMap matches exceptions by their exact class, and this particular exception was never registered in that map. Every other not found exception in the project (artifact, version, group, branch, and so on) already has its own entry, so this was simply a missed one. Without a matching entry the lookup fell back to the default of 500, which wrongly tells the client the server is broken when the comment just does not exist.

## Changes

- Registered CommentNotFoundException in HttpStatusCodeMap so it maps to 404, matching how the sibling not found exceptions are handled. This is a single map entry plus its import, and it keeps the existing exact class lookup untouched.
- Added a test in GroupsResourceTest that deletes and updates a non-existent comment and asserts a 404 response on both. The test fails on the old code (it returns 500) and passes with the fix.

## Testing

- checkstyle passes for the app module.
- The new test testArtifactCommentNotFound passes. The access logs confirm both the DELETE and the PUT now return 404 for a comment ID that does not exist.

Note: the fix lives at the REST exception mapping layer, so it is not storage specific. The v2 and ccompat APIs share the same mapping and are corrected by the same change, at no extra cost.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/http/HttpStatusCodeMap.java
+++ b/app/src/main/java/io/apicurio/registry/services/http/HttpStatusCodeMap.java
@@ -23,6 +23,7 @@ import io.apicurio.registry.storage.error.ArtifactAlreadyExistsException;
 import io.apicurio.registry.storage.error.ArtifactNotFoundException;
 import io.apicurio.registry.storage.error.BranchAlreadyExistsException;
 import io.apicurio.registry.storage.error.BranchNotFoundException;
+import io.apicurio.registry.storage.error.CommentNotFoundException;
 import io.apicurio.registry.storage.error.ConfigPropertyNotFoundException;
 import io.apicurio.registry.storage.error.ContentNotFoundException;
 import io.apicurio.registry.storage.error.ContentSearchNotSupportedException;
@@ -88,6 +89,7 @@ public class HttpStatusCodeMap {
         map.put(BadRequestException.class, HTTP_BAD_REQUEST);
         map.put(BranchAlreadyExistsException.class, HTTP_CONFLICT);
         map.put(BranchNotFoundException.class, HTTP_NOT_FOUND);
+        map.put(CommentNotFoundException.class, HTTP_NOT_FOUND);
         map.put(ConfigPropertyNotFoundException.class, HTTP_NOT_FOUND);
         map.put(ConflictException.class, HTTP_CONFLICT);
         map.put(ContentNotFoundException.class, HTTP_NOT_FOUND);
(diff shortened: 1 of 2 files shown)


parthdagia05:
@paoloantinori didnt knew it bot would directly close the issue i thought we could keep our PRs in draft :(

PR #8847, opened 2026-07-23
feat(wellknown): add Compatible Tools API endpoint for MCP tool output/input chaining

### LINKED ISSUE #8846: feat(wellknown): add Compatible Tools API endpoint for MCP tool output/input chaining
## Feature Description

Organizations deploying MCP (Model Context Protocol) servers and tools need the ability to discover which tools can be chained together in AI orchestration pipelines.

Given an MCP tool definition (tool A) with an `outputSchema`, callers need an endpoint to query: **"Which registered MCP tools have an `inputSchema` compatible with tool A's output?"**

Currently, Apicurio Registry supports MCP tool discovery (`/.well-known/mcp-tools`) and same-tool version compatibility checking (`McpToolCompatibilityChecker`), but does not provide an endpoint for cross-tool input/output compatibility discovery.

## Proposed Solution

Add a **Compatible Tools** discovery endpoint to the well-known resource layer:

`GET /.well-known/mcp-tools/{groupId}/{artifactId}/compatible`

### Expected Behavior

1. **Source Tool Resolution**: Fetches the source MCP tool's content (defaulting to `branch=latest` or a specific version given via `?version=`).
2. **Schema Matching**: Extracts property names and JSON Schema types from the source tool's `outputSchema.properties`.
3. **Candidate Filtering**: Searches candidate `MCP_TOOL` artifacts in the registry, excluding the source tool itself.
4. **Compatibility Rule**: Matches candidates whose `inputSchema.properties` contain all output properties produced by the source tool with matching types.
5. **Paginated Output**: Returns an `McpCompatibleToolsResults` envelope with total `count` and paginated `tools` array.
6. **Graceful Edge Cases**:
   - Source tool has no `outputSchema` -> returns empty list (`count: 0`).
   - Source tool not found -> `404 Not Found`.
   - Disabled feature gate (`apicurio.mcp-tools.enabled=false`) -> `404 Not Found`.

## Acceptance Criteria

- [x] REST contract method in `WellKnownResource` and implementation in `WellKnownResourceImpl`
- [x] Response DTO `McpCompatibleToolsResults`
- [x] Feature gate check (`apicurio.mcp-tools.enabled`)
- [x] Integration tests in `WellKnownMcpToolsTest` with test profile `McpToolsEnabledProfile`

krrishverma1805-web:
## Problem / Context

Apicurio Registry supports `MCP_TOOL` artifacts and exposes discovery endpoints at `/.well-known/mcp-tools`. However, when building AI tool pipelines, callers need to know which registered tools can consume the output of another tool.

This PR adds the **Compatible Tools** discovery endpoint:

`GET /.well-known/mcp-tools/{groupId}/{artifactId}/compatible`

Fixes #8846

---

## Technical Approach

1. **REST Interface & DTO**:
   - Added `findCompatibleTools(...)` method to `WellKnownResource`.
   - Created `McpCompatibleToolsResults` DTO envelope containing `count` and a list of `McpToolSearchResult` items.

2. **Compatibility Logic (`WellKnownResourceImpl`)**:
   - Inspects the source tool's `outputSchema.properties` (names and types).
   - If the source tool has no output properties, returns an empty list (`count: 0`).
   - Searches candidate `MCP_TOOL` artifacts via storage filter, excluding the source tool itself.
   - Evaluates input property compatibility (candidate `inputSchema.properties` must declare every output property with a matching JSON Schema type).
   - Applies pagination parameters (`offset`, `limit`).

3. **Integration Coverage & Schema Headers**:
   - Created `WellKnownMcpToolsTest` under `app/src/test/java/io/apicurio/registry/noprofile/rest/mcptools/` with `McpToolsEnabledProfile`.
   - Added 11 integration test cases covering happy path, missing version/artifact 404s, candidate filter matching, pagination, and `GET /.well-known/schemas/mcp-tool/v1` response headers (`application/schema+json`, `Content-Disposition`, `Cache-Control`).

---

## Test Plan

- [x] `./mvnw checkstyle:check -pl app` -> `0 Checkstyle violations` (BUILD SUCCESS)
- [x] `./mvnw -pl app -am test -Dtest=WellKnownMcpToolsTest -Dsurefire.failIfNoSpecifiedTests=false` -> `Tests run: 11, Failures: 0, Errors: 0` (BUILD SUCCESS)

---

## Checklist

- [x] DCO sign-off on all commits (`Signed-off-by: Krrish Verma <krrishverma1805@gmail.com>`)
- [x] Conventional Commits format (`feat(wellknown): ...`)
- [x] Checkstyle clean
- [x] All unit and integration tests passing
- [x] Zero breaking changes to existing endpoints

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
@@ -17,8 +17,9 @@ import io.apicurio.registry.auth.Authorized;
 import io.apicurio.registry.auth.AuthorizedLevel;
 import io.apicurio.registry.auth.AuthorizedStyle;
 import io.apicurio.registry.mcptools.McpToolsConfig;
-import io.apicurio.registry.rest.v3.beans.McpToolSearchResult;
-import io.apicurio.registry.rest.v3.beans.McpToolSearchResults;
+import io.apicurio.registry.mcptools.rest.beans.McpCompatibleToolsResults;
+import io.apicurio.registry.mcptools.rest.beans.McpToolSearchResult;
+import io.apicurio.registry.mcptools.rest.beans.McpToolSearchResults;
 import io.apicurio.registry.cdi.Current;
 import io.apicurio.registry.logging.Logged;
 import io.apicurio.registry.metrics.health.liveness.ResponseErrorLivenessCheck;
@@ -56,10 +57,14 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 /**
@@ -77,6 +82,14 @@ public class WellKnownResourceImpl implements WellKnownResource {
     private static final ObjectMapper mapper = new ObjectMapper();
 
     private static final int MAX_VISIBILITY_FILTER_RESULTS = 10000;
+    private static final String PROPERTIES_FIELD = "properties";
+
+    /**
+     * Maximum number of MCP-tool candidates evaluated per compatible-tools request.
+     * Kept well below {@link #MAX_VISIBILITY_FILTER_RESULTS} because each candidate
+     * requires a storage round-trip; raising this limit directly raises per-request cost.
+     */
+    private static final int MAX_COMPATIBLE_CANDIDATE_SCAN = 500;
 
     @Inject
     A2AConfig a2aConfig;
@@ -542,9 +555,26 @@ public class WellKnownResourceImpl implements WellKnownResource {
         }
 
         if (parameters != null && !parameters.isEmpty()) {
-            for (String parameter : parameters) {
-                filters.add(SearchFilter.ofStructure("mcp_tool:parameter:" + parameter));
+            // Parameter filtering is performed after artifact search by inspecting tool.getParameters()
+            ArtifactSearchResultsDto results = storage.searchArtifacts(filters, OrderBy.createdOn,
+                    OrderDirection.desc, 0, MAX_VISIBILITY_FILTER_RESULTS, false);
+
+            List<McpToolSearchResult> matchingTools = new ArrayList<>();
+            for (SearchedArtifactDto artifact : results.getArtifacts()) {
+                McpToolSearchResult tool = convertToMcpToolSearchResult(artifact);
+                if (tool.getParameters() != null && tool.getParameters().containsAll(parameters)) {
+                    matchingTools.add(tool);
+                }
             }
+
+            int total = matchingTools.size();
+            int safeOffset = Math.max(0, offset == null ? 0 : offset);
+            int safeLimit = Math.max(1, limit == null ? 10 : limit);
+            int fromIndex = Math.min(safeOffset, total);
+            int toIndex = Math.min(fromIndex + safeLimit, total);
+            List<McpToolSearchResult> page = matchingTools.subList(fromIndex, toIndex);
+
+            return McpToolSearchResults.builder().count(total).tools(page).build();
         }
 
         ArtifactSearchResultsDto results = storage.searchArtifacts(filters, OrderBy.createdOn,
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -554,6 +559,162 @@ public McpToolSearchResults searchMcpTools(String name, List<String> parameters,
...
+    private List<McpToolSearchResult> findCompatibleCandidates(String rawGroupId, String sourceArtifactId,
+            Map<String, String> sourceOutputProps) {
+        Set<SearchFilter> filters = new HashSet<>();
+        filters.add(SearchFilter.ofArtifactType(ArtifactType.MCP_TOOL));
+
+        ArtifactSearchResultsDto candidateResults = storage.searchArtifacts(filters, OrderBy.createdOn,
+                OrderDirection.desc, 0, MAX_VISIBILITY_FILTER_RESULTS, false);
+
+        List<McpToolSearchResult> compatibleTools = new ArrayList<>();
+        for (SearchedArtifactDto candidate : candidateResults.getArtifacts()) {
```
paoloantinori, line 712: This loop drives the whole cost of the request: for each of up to 10000 candidates, `isCandidateCompatible` issues `getBranchTip` + `getArtifactVersionContent`, and pagination only trims the result afterward. Worth bounding the candidate scan (or filtering in storage) so the endpoint stays cheap as the MCP-tool population grows.

paoloantinori (commented):
Solid feature — the compatibility contract (every source `outputSchema` property present in the candidate `inputSchema` with a matching JSON Schema type) is well-modelled, the DTO/builder is clean, and the test coverage (happy path, 404s, incompatibility, no-output-schema, pagination, schema headers) is thorough. A few things worth examining before this lands, mostly around cost-per-request and a couple of correctness edges.

- **Per-request cost scales with the total MCP-tool population, not the page size.** `findCompatibleCandidates` calls `storage.searchArtifacts(... 0, MAX_VISIBILITY_FILTER_RESULTS ...)` (10000) and then, inside `isCandidateCompatible`, makes **two storage round-trips per candidate** (`getBranchTip` + `getArtifactVersionContent`). On a registry with N MCP tools that's up to ~2N storage calls on a single GET, and `buildPaginatedCompatibleResults` only paginates *after* the full in-memory scan — so `offset`/`limit` don't reduce the work, only the payload. For a well-known/discovery endpoint that's a meaningful scaling cliff; worth at least capping the candidate scan well below 10000, or pushing the filter into storage.
- **Compatible candidates fetch their content twice.** `isCandidateCompatible` reads `getArtifactVersionContent` to inspect `inputSchema`, then `convertToMcpToolSearchResult(candidate)` re-fetches the same latest-version content to extract title/parameters. Passing the already-fetched `candidateStored` (or the parsed `candidateRoot`) into the conversion would halve the storage calls for the matching subset.
- **Authorization on candidates.** The method is annotated `@Authorized(GroupAndArtifact, Read)` which covers the *source* artifact, but the loop reads content and returns identity/metadata for *every* matching candidate. If `storage.searchArtifacts` doesn't already enforce per-artifact visibility for the caller, this endpoint could expose tools the requester can't otherwise read. Worth confirming searchArtifacts applies the visibility filter here.
- **Type matching is lenient on non-scalar type declarations.** `extractJsonSchemaType` only returns a type when `node.get("type").isTextual()`. A candidate declaring `{"type": ["string","null"]}` (array form) or using `oneOf`/`$ref` yields `candType == null`, and `candidateAcceptsAllProperties` then skips the type comparison entirely (`if (candType != null && ...)`), treating it as compatible. If strict typing matters for the chaining contract, this is a silent false-positive; if lenient-by-design, a one-line note in the Javadoc would help future readers.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -554,6 +559,162 @@ public McpToolSearchResults searchMcpTools(String name, List<String> parameters,
...
+                && isSameGroup(rawGroupId, candidate.getGroupId())) {
+            return false;
+        }
+
+        try {
+            GA candidateGa = new GA(candidate.getGroupId(), candidate.getArtifactId());
+            GAV candidateGav = VersionExpressionParser.parse(candidateGa, "branch=latest",
+                    (g, branchId) -> storage.getBranchTip(g, branchId,
+                            RetrievalBehavior.SKIP_DISABLED_LATEST));
+            StoredArtifactVersionDto candidateStored = storage.getArtifactVersionContent(
```
paoloantinori, line 741: This fetches the candidate's content to read `inputSchema`; then `convertToMcpToolSearchResult(candidate)` (called at line ~646) fetches the same latest-version content again for title/parameters. Threading this already-loaded content into the conversion would avoid the duplicate round-trip for every compatible candidate.

krrishverma1805-web:
Thank you for the thorough review @paoloantinori, all four points addressed in the latest commit.

---

## 1. Per-request cost / candidate scan cap

Introduced `MAX_COMPATIBLE_CANDIDATE_SCAN = 500` (distinct from the 10 000
`MAX_VISIBILITY_FILTER_RESULTS` used by the entitlements path).  The
`findCompatibleCandidates` method now calls
`storage.searchArtifacts(..., 0, MAX_COMPATIBLE_CANDIDATE_SCAN, ...)` and emits a
`log.warn` when that cap is reached, prompting operators to raise it or move the
filter storage-side.  Added a Javadoc note explaining the trade-off and pointing to
the cap constant for easy tuning.

## 2. Double content fetch eliminated

Replaced `isCandidateCompatible(…): boolean` +
`convertToMcpToolSearchResult(candidate)` (two storage round-trips per compatible
candidate) with a single method:

```
Optional<JsonNode> tryGetCompatibleCandidateRoot(candidate, …)
```

It fetches and parses the candidate's content once, checks compatibility, and
returns the already-parsed `JsonNode` when compatible.  `findCompatibleCandidates`
then calls `convertToMcpToolSearchResultFromContent(candidate, root)` — which takes
the already-parsed node — keeping the per-candidate storage cost to **one**
round-trip.  A new `convertToMcpToolSearchResultFromContent` overload was extracted;
the original `convertToMcpToolSearchResult` now delegates to it to avoid code
duplication.

## 3. Authorization on candidates

Added a Javadoc block on `findCompatibleCandidates` documenting the authorization
scope:

> MCP tools (unlike A2A agents) do not carry an `apicurio.agent.visibility` label,
> so `searchArtifacts` returns all MCP tools accessible to the caller at the
> `Read` level — the same scope exposed by `searchMcpTools`.  The source artifact
> is already gated by `@Authorized(GroupAndArtifact, Read)`, and candidate
> exposure is bounded by the same read-level authorization.

No code change needed here; the behavior is consistent with `searchMcpTools` by
design.  Happy to add a runtime guard (e.g. filtering returned candidates by a
future `apicurio.tool.visibility` label) if the project decides to add per-tool
visibility controls.

## 4. Lenient type matching documented

Added a Javadoc to `extractJsonSchemaType` explaining the lenient-by-design choice:
array `{"type": ["string","null"]}`, `oneOf`, and `$ref` forms return `null`, which
causes `candidateAcceptsAllProperties` to skip the type comparison and treat the
property as compatible.  This is intentional — MCP tool schemas are frequently
nullable or polymorphic, and strict rejection would produce false-negatives.  The
note suggests callers that need strict enforcement perform their own JSON Schema
validation.

---

Also added `testFindCompatibleToolsPagination` to the test suite — verifies that
`offset`/`limit` correctly trims the page while leaving `count` as the total match
count, covering all three cases: full page, trimmed page, and offset-beyond-end.

paoloantinori:
Thanks — re-read the latest commit and all four points are addressed cleanly:

- The `MAX_COMPATIBLE_CANDIDATE_SCAN = 500` cap with the `log.warn` on saturation is the right call, and keeping it distinct from `MAX_VISIBILITY_FILTER_RESULTS` makes the per-request cost easy to reason about.
- `tryGetCompatibleCandidateRoot` parsing once and returning the root for reuse avoids the double fetch neatly.
- The authorization rationale in the javadoc is clear — relying on the storage visibility filter at `AuthorizedLevel.Read`, consistent with `searchMcpTools`.
- Lenient type matching is well documented, and `testFindCompatibleToolsSuccess` covers the incompatible case (`records` string vs array).

No further concerns from my side. The test coverage (404s, pagination count semantics, schema headers) looks solid.

krrishverma1805-web:
Hi @paoloantinori @EricWittmann — all automated tests have passed (`lifecycle/tested`) following the review fixes. PTAL

paoloantinori:
Thanks @krrishverma1805-web — confirmed, the current head (`af3e72a4`) still looks solid from my side: the candidate-scan cap with the saturation warning, the single-round-trip conversion, and the lenient type-matching rationale all hold up. Tests passing is a good sign. No further comments from me — deferring to @EricWittmann / @carlesarnal for the merge decision.

EricWittmann:
Hi @krrishverma1805-web — apologies for the delay. The PR looks good and we'd like to merge it, but it has merge conflicts with main. Could you rebase and resolve them? Thanks!

krrishverma1805-web:
@EricWittmann PTAL..

EricWittmann:
Thanks for the contribution, @krrishverma1805-web! Nice work on the Compatible Tools discovery endpoint. Appreciate the thorough test coverage and the clean iteration on review feedback.

PR #8857, opened 2026-07-23
fix(ui-editors): Prevent name collisions when importing components

### LINKED ISSUE #8856: UI Editors: Prevent name collisions when importing components
## Description:
Currently, when a user imports a component (such as a Schema, Message, or Response) into the AsyncAPI or OpenAPI document via the UI editor, the import logic blindly takes the component's name and adds it to the document without verifying if a component with that exact name already exists.

This can result in silent overrides, duplicate keys, or invalid API definitions because of the name collision.

There are existing `TODO` comments indicating that this needs to be addressed:
- `ui/ui-editors/src/app/editor/aaieditor.component.ts`
- `ui/ui-editors/src/app/editor/oaieditor.component.ts`

## Expected Behavior:
During a component import, the editor should inspect the current document's components/definitions. If a component with the same name is found, a counter should be appended to the imported component's name (e.g. `MySchema-1`, `MySchema-2`) until a unique, non-colliding name is generated.

## Technical Details:
- In `aaieditor.component.ts`, when importing definitions, verify existence in `document.components`.
- In `oaieditor.component.ts`, support for checking name collisions must properly distinguish between OpenAPI 2 documents (which use `.definitions` / `.responses`) and OpenAPI 3 documents (which use `.components`).
- We also need to ensure we use the imported component's type (`imp.type`) rather than the top-level request type during the import loop to avoid mismatched command pushes.

Kayd-06:
## Overview
Fixes #8856

This PR resolves an issue in the AsyncAPI and OpenAPI UI Editors where importing components (like schemas, responses, or messages) could result in name collisions. Previously, the logic blindly added the imported component to the document using its original name, leading to silent overrides or invalid definitions if that name was already taken.

## Changes Made
- Added a loop during the import process to inspect `document.components` (for AsyncAPI and OpenAPI 3) and `document.definitions`/`document.responses` (for OpenAPI 2).
- If a component with the same name already exists, the logic now appends a unique counter to the imported component's name (e.g. `MySchema-1`, `MySchema-2`).
- Ensured we correctly use `imp.type` rather than the requested top-level type to avoid mismatched command generations when importing dependencies (like a schema inside a message trait).

## Testing
- `npm run build` ran successfully locally.
- Verified compilation covers both `aaieditor` and `oaieditor` for different specification variants (`Oas20Document` vs `Oas30Document` vs `AaiDocument`).


ui/ui-editors/src/app/editor/aaieditor.component.ts
```diff
@@ -668,15 +668,33 @@ export class AaiEditorComponent extends AbstractApiEditorComponent implements On
...
+                            }
+                        } else if (imp.type === ComponentType.messageTrait) {
+                            while (docComponents.getMessageTraitDefinition(name) != null) {
+                                name = imp.name + "-" + counter++;
+                            }
+                        } else if (imp.type === ComponentType.message) {
+                            while (docComponents.getMessage(name) != null) {
+                                name = imp.name + "-" + counter++;
+                            }
+                        }
```
paoloantinori, line 690: Because `docComponents` is read once here and the `commands` are only applied as a batch later, two imports sharing `imp.name` within the same `imports` array will both compute the same suffixed name. Consider tracking used names in a `Set<string>` updated inside the loop so intra-batch duplicates are also disambiguated.

ui/ui-editors/src/app/editor/oaieditor.component.ts
```diff
@@ -666,12 +666,30 @@ export class OaiEditorComponent extends AbstractApiEditorComponent implements On
...
+                            while (doc.responses.getResponse(name) != null) name = imp.name + "-" + counter++;
+                        }
+                    } else {
+                        if (doc.components) {
+                            if (imp.type === ComponentType.schema) {
+                                while (doc.components.getSchemaDefinition(name) != null) name = imp.name + "-" + counter++;
+                            } else if (imp.type === ComponentType.response) {
+                                while (doc.components.getResponseDefinition(name) != null) name = imp.name + "-" + counter++;
+                            }
+                        }
```
paoloantinori, line 688: Same within-batch note as the AsyncAPI editor: `doc`/`doc.components` reflects pre-import state, so concurrent imports of identically-named schemas/responses will collide. A running set of taken names would close this gap.

paoloantinori (commented):
Good catch on the underlying bug — switching from the outer-scope `type` to `imp.type` is the correct fix, since the closure variable would otherwise resolve to the top-level requested type rather than each dependency's actual type (e.g. a schema imported as part of a message trait). The collision-prevention loops are a reasonable approach.

A couple of observations worth considering:

- **Within-batch collisions are not detected.** `docComponents` / `doc` is captured once before the `imports.forEach(...)` loop, and the rename only checks against the document's *current* state. But the `CommandFactory.create...Command` calls are pushed into the `commands` array and applied later as a batch — the document is not mutated during the loop. So if a single import operation brings in two components that share the same `imp.name`, both will resolve to e.g. `MySchema-1` and the second will still silently overwrite the first when the batch is applied. If batch imports of same-named components are a realistic scenario, the already-renamed names need to be tracked in a local set as the loop progresses.
- **Collision detection runs for `messageTrait`/`message` but no command is emitted.** In `aaieditor.component.ts` the `while` loops for `ComponentType.messageTrait` and `ComponentType.message` compute a unique `name`, but the corresponding command branches below are still commented out. The rename work is therefore discarded. This matches the pre-existing stub behavior, so it's not a regression — just flagging that the new logic is partially inert for those types.

Kayd-06:
@EricWittmann Could you take another look at this PR?

EricWittmann:
Thanks for the contribution @Kayd-06! Good catch on the type vs imp.type scoping bug.

PR #8861, opened 2026-07-23
fix(tests): disable tests for local deployment in ITBase class

### LINKED ISSUE #7892: Use JUnit 5 conditional execution for deployment-mode-dependent operator IT tests
## Summary

Several operator integration tests skip execution when running in local deployment mode using `log.warn()` + early `return`. This causes skipped tests to be reported as **passed** rather than **skipped**, hiding gaps in test coverage.

## Affected tests

- `RestrictedNamespaceITTest`
- `OperatorConfigITTest`
- `LeaderElectionITTest` (from PR #7870)

## Proposed approach

Use JUnit 5's `@DisabledIf` annotation with a shared condition method in `ITBase`, so tests that require a remote operator deployment are declaratively skipped and correctly reported by JUnit. This avoids duplicating skip logic in each test body.

kartikangiras:
## Description

Several operator integration tests (RestrictedNamespaceITTest, OperatorConfigITTest, LeaderElectionITTest) skip execution in local deployment mode using log.warn() + early return inside test bodies, causing JUnit to report them as passed rather than skipped hiding coverage gaps.

### Changes
- Adds a shared isLocalDeployment() condition method to ITBase
- Replaces the manual skip blocks with @DisabledIf annotations pointing to the shared condition
- Removes the now-unused Logger field from OperatorConfigITTest
- Tests requiring a remote operator deployment are now declaratively skipped and correctly reported by JUnit.

### Issue
fixes #7892 

### Proof Manifests
All the affected tests are now skipped as expected
image

### DIFF
--- a/operator/controller/src/test/java/io/apicurio/registry/operator/it/ITBase.java
+++ b/operator/controller/src/test/java/io/apicurio/registry/operator/it/ITBase.java
@@ -112,6 +112,10 @@ public abstract class ITBase implements OperatorTestContext {
         return namespace;
     }
 
+    static boolean isLocalDeployment() {
+        return getConfig().getValue(OPERATOR_DEPLOYMENT_PROP, OperatorDeployment.class) == OperatorDeployment.local;
+    }
+
     @BeforeAll
     public void before() throws Exception {
         operatorDeployment = getConfig().getValue(OPERATOR_DEPLOYMENT_PROP,
--- a/operator/controller/src/test/java/io/apicurio/registry/operator/it/LeaderElectionITTest.java
+++ b/operator/controller/src/test/java/io/apicurio/registry/operator/it/LeaderElectionITTest.java
@@ -11,6 +11,7 @@ import io.quarkus.test.junit.QuarkusTest;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIf;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -36,12 +37,8 @@ public class LeaderElectionITTest extends ITBase {
     private static final String DEFAULT_LEASE_NAME = "apicurio-registry-operator-lease";
 
     @Test
+    @DisabledIf("io.apicurio.registry.operator.it.ITBase#isLocalDeployment")
     void testLeaderElectionCreatesLease() {
-        if (operatorDeployment == OperatorDeployment.local) {
-            log.warn("This test requires an ability to edit the operator Deployment, so it's not supported when running locally.");
-            return;
-        }
-
         var operatorDeploymentCell = k8sCell(client, this::getOperatorDeployment);
         Cell<EnvVar> originalEnabledEnvVar = cell();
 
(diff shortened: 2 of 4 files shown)


kartikangiras:
Hi @jsenko @paoloantinori  ,
Is this the right fix to be done or something needs to be changed.
Thanks

paoloantinori:
Hi @kartikangiras — yes, this is the right fix. Replacing the manual `if (operatorDeployment == local) { log.warn(...); return; }` blocks with `@DisabledIf("io.apicurio.registry.operator.it.ITBase#isLocalDeployment")` is exactly the idiomatic JUnit 5 way to do this, and it fixes the real problem: those tests were being reported as *passed* instead of *skipped*, which hid the coverage gap.

A couple of small notes (non-blocking):
- The static `isLocalDeployment()` method is fine as-is. If `ITBase` ever gains subclasses outside `io.apicurio.registry.operator.it`, the fully-qualified name in the annotation will still resolve, so no action needed — just flagging that the FQN is load-bearing.
- Since `ITBase` now references `getConfig()` outside `@BeforeAll`, make sure that resolves at condition-evaluation time (conditions run before `@BeforeAll`); from the diff it looks like `getConfig()` is a static accessor on the base, so that should be fine, but worth a glance at the skipped-count in the local run to confirm the condition actually fires.

Nice cleanup.

paoloantinori:
Hi @kartikangiras — from my side this looks good to merge. The `@DisabledIf` approach with the shared `isLocalDeployment()` condition is the right fix, the unused `Logger` cleanup in `OperatorConfigITTest` is consistent, and SonarCloud's green. Just needs a final go from one of the other maintainers (`@jsenko` / `@carlesarnal`) since I can't approve it myself.

kartikangiras:
Hi @EricWittmann @jsenko ,
The review has been done last week. Could you take this PR to the the next approval and merge phase.
Thanks

carlesarnal:
Reviewed the full diff. Clean, well-scoped change.

**What it does:** Replaces manual if-return skip blocks in LeaderElectionITTest, OperatorConfigITTest, and RestrictedNamespaceITTest with @DisabledIf annotations pointing to a shared ITBase#isLocalDeployment() method. JUnit now correctly reports these tests as skipped rather than passed.

**Checklist:**
- DCO: Signed-off
- Conventional commit: refactor: replace local deployment manual checks with @DisabledIf condition in integration tests
- Scope: All 4 files tightly focused on the same concern
- Imports: No star imports, no unused imports
- Logger cleanup in OperatorConfigITTest: Correct - field was only used for the removed log.warn() call. Loggers in LeaderElectionITTest and RestrictedNamespaceITTest correctly kept (still used for log.info() calls)
- isLocalDeployment() uses statically-imported ConfigProvider.getConfig() which resolves at condition-evaluation time before BeforeAll
- SonarCloud: Quality Gate passed, 0 new issues

**Notes:**
- Agreed with @paoloantinori - the @DisabledIf approach with a shared condition method is the idiomatic JUnit 5 way to handle this.
- Package-private visibility on isLocalDeployment() is appropriate since all consuming test classes are in the same package.

No issues found. Looks good from my side - ready for maintainer approval.

paoloantinori:
Thanks @carlesarnal. Agreed — the `@DisabledIf` with the shared `ITBase#isLocalDeployment()` condition is the clean, declarative fix, and the unused `Logger` removal in `OperatorConfigITTest` is a nice touch. From my side this is good to merge.

EricWittmann:
Thanks @kartikangiras — clean, idiomatic fix. Apologies for the delay. Merged!

PR #8862, opened 2026-07-23
fix(rest): return 404 instead of 500 for a missing version comment

### LINKED ISSUE #8843: Deleting or updating a non-existent version comment returns HTTP 500 instead of 404

bhuvan-somisetty:
## What
Deleting or updating a version comment that does not exist now returns HTTP 404 (Not Found) instead of HTTP 500 (Internal Server Error).

Fixes #8843

## Why
CommentNotFoundException extends NotFoundException, which is already mapped to HTTP 404 in HttpStatusCodeMap. The status code lookup matches exceptions by their exact class, and CommentNotFoundException was never added as its own entry. Every other not-found exception in the project (artifact, version, group, branch, etc.) already has its own entry; this was simply a missed registration. Without a matching entry the lookup fell back to the default of 500, which wrongly tells the client the server is broken when the comment just does not exist.

## Changes
- Registered CommentNotFoundException in HttpStatusCodeMap so it maps to 404, consistent with how all sibling not-found exceptions are handled. This is a single map entry plus its import, with no changes to the existing lookup logic.
- Extended 	estArtifactComments in GroupsResourceTest with assertions that DELETE and PUT against a non-existent comment ID both return 404.

## Testing
- ./mvnw checkstyle:check -pl app passes with 0 violations.
- The new assertions fail on the old code (returns 500) and pass with the fix applied.
- The fix is at the REST exception mapping layer and is not storage specific; v2 and ccompat APIs share the same map and are corrected by the same single change.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/http/HttpStatusCodeMap.java
+++ b/app/src/main/java/io/apicurio/registry/services/http/HttpStatusCodeMap.java
@@ -23,6 +23,7 @@ import io.apicurio.registry.storage.error.ArtifactAlreadyExistsException;
 import io.apicurio.registry.storage.error.ArtifactNotFoundException;
 import io.apicurio.registry.storage.error.BranchAlreadyExistsException;
 import io.apicurio.registry.storage.error.BranchNotFoundException;
+import io.apicurio.registry.storage.error.CommentNotFoundException;
 import io.apicurio.registry.storage.error.ConfigPropertyNotFoundException;
 import io.apicurio.registry.storage.error.ContentNotFoundException;
 import io.apicurio.registry.storage.error.ContentSearchNotSupportedException;
@@ -88,6 +89,7 @@ public class HttpStatusCodeMap {
         map.put(BadRequestException.class, HTTP_BAD_REQUEST);
         map.put(BranchAlreadyExistsException.class, HTTP_CONFLICT);
         map.put(BranchNotFoundException.class, HTTP_NOT_FOUND);
+        map.put(CommentNotFoundException.class, HTTP_NOT_FOUND);
         map.put(ConfigPropertyNotFoundException.class, HTTP_NOT_FOUND);
         map.put(ConflictException.class, HTTP_CONFLICT);
         map.put(ContentNotFoundException.class, HTTP_NOT_FOUND);
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/GroupsResourceTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/GroupsResourceTest.java
@@ -55,6 +55,7 @@ import java.util.UUID;
 
 import static io.restassured.RestAssured.given;
 import static java.net.HttpURLConnection.HTTP_NO_CONTENT;
+import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 import static java.net.HttpURLConnection.HTTP_OK;
 import static org.hamcrest.CoreMatchers.anyOf;
 import static org.hamcrest.CoreMatchers.anything;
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
Hi @EricWittmann @carlesarnal @paoloantinori - whenever you have a moment, could you please review this PR? It fixes issue #8843 by mapping \CommentNotFoundException\ to HTTP 404 in \HttpStatusCodeMap\ and adds unit test coverage for missing comment updates and deletes. Checkstyle and builds pass cleanly.

paoloantinori:
Thanks @bhuvan-somisetty — reviewed the latest head. The single `CommentNotFoundException` → `HTTP_NOT_FOUND` entry in `HttpStatusCodeMap` is consistent with how every other sibling not-found exception is registered, and the new assertions in `testArtifactComments` (both PUT and DELETE against a non-existent `commentId` returning 404) confirm the before/after behavior. Looks good to me — no further changes needed from my side.

bhuvan-somisetty:
Thank you @paoloantinori for reviewing!

paoloantinori:
Happy to help! Glad the review was useful. 👍

bhuvan-somisetty:
Hi @EricWittmann @carlesarnal @paoloantinori - all CI checks have completed successfully (\lifecycle/tested\). Whenever you have a moment, could you please take a look for final review/merge? Thanks!

paoloantinori:
Thanks @bhuvan-somisetty — CI is green and the change looks good on this head. The single `CommentNotFoundException` → `HTTP_NOT_FOUND` entry plus the two new 404 assertions are consistent with how every sibling not-found exception is handled. Nothing further from my side; ready for a maintainer to merge.

bhuvan-somisetty:
Thank you @paoloantinori !

paoloantinori:
You're welcome @bhuvan-somusetty! Happy to help — the change is in good shape. 👍

bhuvan-somisetty:
Hi @EricWittmann @carlesarnal - gentle ping on this PR. All CI checks are green (\lifecycle/tested\) and @paoloantinori has approved it. Whenever you have a moment, could you please merge this? Thanks!

paoloantinori:
Seen, thanks @bhuvan-somisetty. CI is green and the change still looks good on this head (`7989c0a`). Deferring to @EricWittmann / @carlesarnal for the merge.

EricWittmann:
@bhuvan-somisetty — nice catch and clean fix. Thanks for the contribution!

carlesarnal:
Reviewed the changes — this looks good.

**What was checked:**

- **DCO sign-off**: Present ✅
- **Conventional commit**: `fix(rest): return 404 instead of 500 for a missing version comment` ✅
- **Code change** (`HttpStatusCodeMap.java`): The `CommentNotFoundException` → `HTTP_NOT_FOUND` entry is consistent with every other `NotFoundException` subclass in the map (`ArtifactNotFoundException`, `BranchNotFoundException`, `VersionNotFoundException`, etc.). Import and map entry are both in correct alphabetical order. ✅
- **Tests** (`GroupsResourceTest.java`): Both PUT and DELETE against a non-existent comment ID are asserted to return 404, placed logically at the end of `testArtifactComments`. Assertions are specific (`.statusCode(HTTP_NOT_FOUND)`). ✅
- **No unrelated changes**, no star imports, no unused imports ✅

Clean fix, well-tested. No issues found.

PR #8866, opened 2026-07-23
fix(kafkasql): journal global contract ruleset operations so they replicate across nodes

### LINKED ISSUE #8848: bug: kafkasql global contract ruleset is not replicated across nodes and is lost on restart
## Description

**Registry Version**: 3.3.0

In a multi-node kafkasql deployment, a global contract ruleset set through one node is not visible on the other nodes, and it disappears when the node that served the write restarts. This means the global contract ruleset is effectively not enforceable cluster-wide, and the configuration is silently lost on restart — including in single-node deployments.

I reproduced this in a live two-node KafkaSQL deployment. As a control, artifact creation replicated correctly between nodes, confirming the cluster configuration was working correctly.

### Environment

- apache/kafka:3.9.1 (single KRaft broker)
- 2× quay.io/apicurio/apicurio-registry:3.3.0
- `APICURIO_STORAGE_KIND=kafkasql`, both nodes sharing the same Kafka broker

### Steps to Reproduce

1. Start Kafka and two Registry nodes (A: 8081, B: 8082); wait until both answer `GET /apis/registry/v3/system/info`
2. Control: create an artifact on node A, `GET` it on node B → `200 OK`
3. Set the global contract ruleset on node A:

       curl -X PUT localhost:8081/apis/registry/v3/admin/contracts/ruleset \
         -d '{"domainRules":[{"name":"repro-rule","kind":"CONDITION","type":"QUALITY","mode":"WRITE","expr":"true"}],"migrationRules":[]}'

4. `GET /apis/registry/v3/admin/contracts/ruleset` on both nodes
5. Restart node A and repeat the `GET` on node A

## Expected Behaviour

A global contract ruleset set through any node is visible on every node in the cluster and survives a node restart.

## Actual Behaviour

- Node A returns the ruleset; node B returns `{"domainRules":[],"migrationRules":[]}`
- After restarting node A, it also returns `{"domainRules":[],"migrationRules":[]}`
- The control artifact from step 2 is still present on both nodes throughout

## Logs / Evidence

Dumping the `kafkasql-journal` topic after the full sequence shows the following message types:

    3 "messageType":"Bootstrap"
    2 "messageType":"CreateArtifact11Message"
    2 "messageType":"MergeArtifactLabels4Message"

During the reproduction, the KafkaSQL journal contained bootstrap and artifact-related messages, but I did not observe any journal message corresponding to the global contract ruleset operations.

### Additional Notes

Based on the investigation, it appears that `setGlobalContractRuleset` and `deleteGlobalContractRuleset` in `KafkaSqlRegistryStorage` write directly to the node-local SQL store, whereas the artifact-level and version-level contract ruleset operations are propagated through the KafkaSQL journal.

During the investigation I also observed:

- `insertContractAuditEntry` appears to follow the same direct-write path.
- There is currently no KafkaSQL test coverage for the global contract ruleset path.

One possible implementation would be to route these operations through the existing KafkaSQL message/journal mechanism, consistent with similar contract operations.

Anuragkumar-687:
Fixes Apicurio/apicurio-registry#8848.

## Problem

In the KafkaSQL storage variant, `setGlobalContractRuleset` and `deleteGlobalContractRuleset` in `KafkaSqlRegistryStorage` write directly to the node-local SQL store instead of submitting a message to the KafkaSQL journal. The artifact-level and version-level contract ruleset operations in the same class go through the journal, so the global operations are the only contract ruleset writes that bypass it.

Because each KafkaSQL node builds its state from the journal, a write that skips the journal is not replicated to the other nodes and is not restored when the node restarts.

## Fix

Route `setGlobalContractRuleset` and `deleteGlobalContractRuleset` through the journal, consistent with the existing contract ruleset operations:

- Add `SetGlobalContractRuleset1Message` and `DeleteGlobalContractRuleset0Message`, following the existing contract ruleset message classes.
- Submit these messages via the submitter and wait on the coordinator, as `setArtifactContractRuleset` and the related operations already do.
- Register both message types in `KafkaSqlMessageIndex`.

## Testing

Added `KafkaSqlGlobalContractRulesetJournalTest`, which sets a global contract ruleset and asserts that a `SetGlobalContractRuleset1Message` is produced to the KafkaSQL journal topic. The test fails on the previous implementation (no journal message is produced) and passes with this change. It passes locally along with `checkstyle:check` for the `app` module.

## Notes

This change is intentionally scoped to the KafkaSQL implementation addressed by #8848. The SQL storage implementation already journals these operations. The KafkaSQL contract audit path (`insertContractAuditEntry`) follows a similar direct-write pattern and is left for a separate change to keep this PR focused.

### DIFF
--- a/app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlGlobalContractRulesetJournalTest.java
+++ b/app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlGlobalContractRulesetJournalTest.java
@@ -1,0 +1,85 @@
+package io.apicurio.registry.storage.impl.kafkasql;
+
+import io.apicurio.registry.AbstractResourceTestBase;
+import io.apicurio.registry.storage.dto.ContractRuleSetDto;
+import io.apicurio.registry.storage.impl.kafkasql.messages.DeleteGlobalContractRuleset0Message;
+import io.apicurio.registry.storage.impl.kafkasql.messages.SetGlobalContractRuleset1Message;
+import io.apicurio.registry.utils.tests.KafkasqlTestProfile;
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import jakarta.inject.Inject;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.serialization.StringDeserializer;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+@QuarkusTest
+@TestProfile(KafkasqlTestProfile.class)
+public class KafkaSqlGlobalContractRulesetJournalTest extends AbstractResourceTestBase {
+
+    private static final String JOURNAL_TOPIC = "kafkasql-journal";
+
+    @Inject
+    KafkaSqlRegistryStorage kafkaSqlRegistryStorage;
+
+    @Test
+    public void testSetGlobalContractRulesetIsJournaled() {
+        ContractRuleSetDto ruleset = ContractRuleSetDto.builder().domainRules(List.of())
+                .migrationRules(List.of()).build();
+        assertOperationIsJournaled(() -> kafkaSqlRegistryStorage.setGlobalContractRuleset(ruleset),
+                SetGlobalContractRuleset1Message.class.getSimpleName());
+    }
+
+    @Test
+    public void testDeleteGlobalContractRulesetIsJournaled() {
+        assertOperationIsJournaled(() -> kafkaSqlRegistryStorage.deleteGlobalContractRuleset(),
+                DeleteGlobalContractRuleset0Message.class.getSimpleName());
+    }
+
+    /**
+     * Runs the given storage operation and asserts that a message of the expected type is produced to
+     * the KafkaSQL journal topic.
+     */
+    private void assertOperationIsJournaled(Runnable operation, String expectedType) {
+        KafkaConsumer<String, String> consumer = new KafkaConsumer<>(
+                Map.of(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG,
+                        System.getProperty("bootstrap.servers.external"),
+                        ConsumerConfig.GROUP_ID_CONFIG, "tc-" + UUID.randomUUID(),
+                        ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest"),
+                new StringDeserializer(), new StringDeserializer());
+        try {
+            consumer.subscribe(List.of(JOURNAL_TOPIC));
+
+            operation.run();
+
+            long deadline = System.currentTimeMillis() + 15000;
+            boolean found = false;
+            while (System.currentTimeMillis() < deadline && !found) {
+                for (ConsumerRecord<String, String> record : consumer.poll(Duration.ofMillis(500))) {
+                    Header header = record.headers().lastHeader(KafkaSqlSubmitter.MESSAGE_TYPE_HEADER);
+                    if (header != null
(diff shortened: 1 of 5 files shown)


app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlGlobalContractRulesetJournalTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+@QuarkusTest
+@TestProfile(KafkasqlTestProfile.class)
+public class KafkaSqlGlobalContractRulesetJournalTest extends AbstractResourceTestBase {
+
```
paoloantinori, line 28: This covers the `set` path only. Since the PR also journals `deleteGlobalContractRuleset` via `DeleteGlobalContractRuleset0Message`, a symmetric test asserting that delete emits that message type would be valuable — see the review comment for rationale.

paoloantinori (commented):
Good catch on this — the global contract ruleset ops were the only contract-ruleset writes bypassing the KafkaSQL journal, so they wouldn't replicate or survive a node restart. Routing `setGlobalContractRuleset`/`deleteGlobalContractRuleset` through the submitter + coordinator, and appending the two new message types at the end of `KafkaSqlMessageIndex` (so existing index bytes are untouched), mirrors exactly what `setArtifactContractRuleset`/`setVersionContractRuleset` already do. The fix looks correct.

One coverage gap worth closing:

- The PR adds **both** a `Set*` and a `Delete*` message class and journals **both** operations, but `KafkaSqlGlobalContractRulesetJournalTest` only asserts that the `set` path produces a journal message. There's no symmetric assertion that `deleteGlobalContractRuleset()` emits a `DeleteGlobalContractRuleset0Message`. The delete path is the riskier of the two to leave untested here, because the old code's `sqlStore.deleteGlobalContractRuleset()` is now never called directly — if the journal wiring for delete were ever broken (wrong message type registered, dispatch target mismatch, etc.), a global ruleset delete would silently no-op on the local node and never replicate, and nothing in this test suite would catch it. A second test mirroring the existing one (call `kafkaSqlRegistryStorage.deleteGlobalContractRuleset()`, assert a `DeleteGlobalContractRuleset0Message` on the journal topic) would close that.

Minor (non-blocking): the test relies on `System.getProperty("bootstrap.servers.external")` and a hardcoded `"kafkasql-journal"` topic name. Worth double-checking these match what `KafkasqlTestProfile` actually configures — if the property name drifts the test would `poll()` an empty broker and fail only on the 15s timeout rather than with a clear cause. Not a correctness issue with the fix itself.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlGlobalContractRulesetJournalTest.java
Anuragkumar-687: Thanks for the review! I added a symmetric regression test covering deleteGlobalContractRuleset(), verifying that it produces a DeleteGlobalContractRuleset0Message on the journal topic. I also extracted the shared consumer setup and polling logic into a helper so both tests reuse the same implementation.

I also double-checked the bootstrap.servers.external property and journal topic configuration against KafkasqlTestProfile; they already match the current test setup, so no changes were needed there.

Anuragkumar-687: Hi @paoloantinori, I addressed the review comments. When you have a chance, I'd appreciate another look. Thanks!

carlesarnal (approved):
This PR correctly fixes the KafkaSQL replication gap for global contract ruleset operations. Both setGlobalContractRuleset and deleteGlobalContractRuleset now route through the journal, consistent with artifact/version-level operations. DCO, conventional commit, tests (set+delete), imports, brace style, and message index registration all check out. Minor non-blocking nit: DeleteGlobalContractRuleset0Message omits @NoArgsConstructor and @Builder that the analogous DeleteGlobalRules0Message uses (functionally fine). paoloantinori's feedback on adding a delete test has been addressed. LGTM.

paoloantinori:
Seen — the bot is flagging this as approved and tested and ready to merge. The substantive review is already complete (carlesarnal approved, my earlier comments are addressed). I will leave the merge decision to a human maintainer per the usual workflow.

PR #8868, opened 2026-07-23
fix(ui-editors): sanitize markdown output to prevent XSS

hussainjamal760:
## Description

This PR addresses a potential **Cross-Site Scripting (XSS)** vulnerability in the `MarkdownComponent` of the UI editors.

Previously, the HTML generated by the `marked` library was rendered directly without sanitization, leaving a `TODO` to integrate DOMPurify. This update implements **DOMPurify** to sanitize the parsed Markdown output before it is rendered, preventing malicious scripts and unsafe HTML from being injected into the DOM.

---

### ✅ Added DOMPurify Dependencies

- Added `dompurify` and `@types/dompurify` to the `ui-editors` package.
- Updated `package-lock.json` to reflect the new dependencies.

### ✅ Sanitized Markdown Output

- Updated `markdown.component.ts` to sanitize the HTML generated by `marked` using `DOMPurify.sanitize()`.
- Ensures only safe HTML is rendered while removing potentially dangerous content.

### ✅ Integrated Angular SafeHtml

- Used Angular's `DomSanitizer.bypassSecurityTrustHtml()` after sanitization.
- Allows Angular to safely render the already-sanitized HTML without stripping legitimate markup.

### ✅ Updated TypeScript Configuration

- Enabled `"allowSyntheticDefaultImports": true` in `tsconfig.json`.
- Supports the default import syntax used by DOMPurify:

```ts
import DOMPurify from "dompurify";
```

---

## Motivation and Context

Rendering raw, unsanitized Markdown introduces a potential XSS attack vector when user-provided content contains malicious HTML or JavaScript.

This PR eliminates that risk by sanitizing all parsed Markdown with DOMPurify before rendering. It also completes the previously documented `TODO` for HTML sanitization, improving the security of the `MarkdownComponent` without affecting normal Markdown rendering.

---

## Testing Performed

The following scenarios were verified:

- ✅ Standard Markdown renders correctly after sanitization.
- ✅ Malicious HTML (e.g., `<script>alert('XSS')</script>`) is removed.
- ✅ Unsafe URLs (e.g., `javascript:` links) are sanitized.
- ✅ The UI builds successfully with the updated TypeScript configuration and DOMPurify dependencies.
- ✅ No regressions were observed in normal Markdown rendering.


EricWittmann:
This looks good @hussainjamal760 — the DOMPurify integration is exactly what the existing TODO asked for, and the sanitization pattern (DOMPurify first, then `bypassSecurityTrustHtml`) is correct. Clean implementation.

When you're ready, please take this out of draft status so we can get CI running and move it through the pipeline.

EricWittmann (commented):
One small thing @hussainjamal760 — please pin the `dompurify` and `@types/dompurify` versions to exact values (e.g., `"3.4.12"` instead of `"^3.4.12"`, and `"3.0.5"` instead of `"^3.0.5"`). The rest of the dependencies in this `package.json` use exact versions, and we use Renovate to manage upgrades, so the `^` range isn't needed.

hussainjamal760:
> This looks good @hussainjamal760 — the DOMPurify integration is exactly what the existing TODO asked for, and the sanitization pattern (DOMPurify first, then `bypassSecurityTrustHtml`) is correct. Clean implementation.
> 
> When you're ready, please take this out of draft status so we can get CI running and move it through the pipeline.

Thanks @EricWittmann  for the review and the feedback! I'm glad the implementation matches the intended approach.

I've taken the PR out of draft and updated the versions to exact values. Looking forward to any further feedback.

ui/ui-editors/package.json
```diff
@@ -37,6 +38,7 @@
         "@angular-devkit/build-angular": "18.2.21",
         "@angular/cli": "21.1.5",
         "@angular/compiler-cli": "18.2.14",
```
paoloantinori, line 40: `@types/dompurify` is a deprecated stub — dompurify 3.x ships its own types (see the package's npm page: "dompurify provides its own type definitions, so you do not need this installed"). Pinning `3.0.5` adds a stale duplicate source of types that can shadow the bundled ones. Suggest removing this line entirely and relying on `dompurify`'s bundled types; if the default import in `markdown.component.ts` then fails to resolve, switch to `import * as DOMPurify from "dompurify"` rather than re-adding this package.

ui/ui-editors/tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "importHelpers": true,
     "target": "es2022",
     "module": "es2020",
+    "allowSyntheticDefaultImports": true,
```
paoloantinori, line 15: `allowSyntheticDefaultImports` was added to enable `import DOMPurify from "dompurify"`. Worth verifying it's still needed once `@types/dompurify` is removed — dompurify's bundled types export a namespace, so a `import * as DOMPurify` may type-check without this compiler flag. If the default import still resolves cleanly, leaving the flag in is harmless, but it'd be good to confirm it's not masking a types resolution problem.

paoloantinori (commented):
The DOMPurify integration is the right fix for the long-standing `TODO` and the sanitize-then-`bypassSecurityTrustHtml` ordering is correct (DOMPurify runs on the raw string, the Angular bypass only re-trusts the already-clean HTML). One concrete dependency issue worth resolving before merge:

- **`@types/dompurify` is a deprecated stub and is redundant.** As of `dompurify` 3.x the library ships its own bundled TypeScript types, and the `@types/dompurify` package is explicitly deprecated — its npm page (https://www.npmjs.com/package/@types/dompurify, latest 3.2.0) states: _"This is a stub types definition. dompurify provides its own type definitions, so you do not need this installed."_ This PR pins `@types/dompurify` to `3.0.5`, which is a stale stub that predates the bundled types. Removing it (and dropping the `@types/trusted-types` transitive it pulls in) lets the bundled types win and avoids a confusing duplicate source of truth.
- **The `allowSyntheticDefaultImports: true` change may then be unnecessary.** It was added to support `import DOMPurify from "dompurify"`. With `@types/dompurify` gone, dompurify's own types still support a default import, so this flag is likely still fine — but it's worth confirming the build still type-checks with the stub removed, since the bundled types use a namespace export (`import * as DOMPurify`). If the default import stops resolving, prefer the namespace import over re-adding the deprecated `@types` package.
- Minor: the `package-lock.json` diff also deletes several unrelated `chokidar`/`readdirp` optional-peer entries under `@angular-devkit/*`, `@angular/cli`, and `@schematics/angular`. Those look like lockfile churn from a re-resolve rather than intentional changes — worth a quick sanity check that those deletions are expected and not a side effect of a different npm/node version on the contributor's machine.

No security concerns with the sanitization logic itself; once the dependency cleanup is confirmed this is good to go.

hussainjamal760:
@paoloantinori  I've removed the @types/dompurify package and pinned the dompurify version. I tested removing allowSyntheticDefaultImports: true and switching to a namespace import (import * as DOMPurify), but Webpack/Angular fails to resolve the sanitize method because dompurify exports it under a default export. Using import DOMPurify with allowSyntheticDefaultImports: true is indeed necessary to make the bundled types resolve correctly for Angular. The lockfiles are also clean now.

paoloantinori:
Thanks for the detailed follow-up @hussainjamal760.

On `allowSyntheticDefaultImports`: your explanation matches what I'd expect. DOMPurify's types place `sanitize` on the default export, so `import * as DOMPurify` doesn't expose it under the module interop that Webpack/Angular resolves. The default-import form (`import DOMPurify from "dompurify"`) with `allowSyntheticDefaultImports: true` is the pragmatic fix. An alternative would be `esModuleInterop: true` (which implies `allowSyntheticDefaultImports` and is the Angular CLI default for new projects), but since the existing `marked` import already uses the namespace form, enabling `allowSyntheticDefaultImports` alone is the more minimal change here — fine to keep.

One thing to double-check, though: you mentioned removing `@types/dompurify`, but the current diff (head `121a554`) still adds it in both `package.json` (`"@types/dompurify": "3.0.5"`) and `package-lock.json` (the full `node_modules/@types/dompurify` block). Since `dompurify` 3.x ships its own bundled `.d.ts` types, `@types/dompurify` is redundant and could shadow or conflict with them. Could you confirm whether the removal actually landed in the pushed commit? If the intent is to drop it, both the `package.json` entry and the `node_modules/@types/dompurify` block in `package-lock.json` need to be removed.

hussainjamal760:
@paoloantinori  Ah, you're right. My previous commit to remove @types/dompurify didn't actually get pushed. I've just pushed a clean commit that drops both the package.json entry and the package-lock.json block for it without any unrelated lockfile churn. Thanks for catching that!

paoloantinori:
Thanks @hussainjamal760 — confirmed on the current head:

- `package.json` has `dompurify` pinned to `3.4.12` (exact, no caret) with no `@types/dompurify` entry.
- `package-lock.json` drops the `@types/dompurify` block and only adds the `dompurify` / `@types/trusted-types` (optional dep of dompurify) entries, with no unrelated churn.

That's exactly the cleanup I was pointing at. From my side the PR looks good — `sanitize-then-bypassSecurityTrustHtml` ordering is correct and the dependency surface is clean. The only remaining thread is `allowSyntheticDefaultImports: true` in `tsconfig.json`, which we discussed earlier and which is fine to keep given DOMPurify's default-export types.

EricWittmann:
Thanks for the contribution @hussainjamal760! Clean implementation of a long-standing TODO, and the sanitization pattern is exactly right. Merged.

PR #8875, opened 2026-07-24
feat(operator): decouple registry version from operator via spec.version field

### LINKED ISSUE #7635: Add spec.version field to ApicurioRegistry CR for independent operator/registry upgrades
## Description

Currently, the Apicurio Registry version is managed automatically by the operator — the registry version is tightly coupled to the operator version. This means upgrading the operator also upgrades the registry in a single step, which increases risk during upgrades.

## Proposal

Add a `spec.version` field to the `ApicurioRegistry` custom resource that allows users to explicitly control which version of the registry is deployed, independently of the operator version.

This enables a **staged upgrade process** that minimizes risk:

1. **Before upgrade**: Operator 3.0.3 running Registry 3.0.3
2. **Stage 1 — Upgrade operator only**: Upgrade Operator to 3.0.6, keep Registry at 3.0.3 (`spec.version: 3.0.3`)
3. **Stage 2 — Verify**: Confirm the operator upgrade is successful and stable
4. **Stage 3 — Upgrade registry**: Update `spec.version: 3.0.6` in the CR to upgrade the registry

## Benefits

- **Reduced risk**: Decouples operator and registry upgrades, allowing each to be verified independently.
- **Rollback flexibility**: If the operator upgrade introduces issues, the registry remains on the known-good version.
- **Controlled rollout**: Teams can schedule operator and registry upgrades at different times according to their change management processes.

## Considerations

- The operator should validate that the requested `spec.version` is compatible with the current operator version.
- If `spec.version` is not specified, the operator should default to the current behavior (use the version bundled with the operator).
- Downgrades (setting `spec.version` to a version older than the currently running one) should be handled carefully, with appropriate warnings or restrictions.

hemantch01:
fixes: #7635 
### Description

Currently, the Apicurio Registry version is tightly coupled to the operator version. This means upgrading the operator automatically forces an upgrade of the registry in a single step, which increases risk during cluster maintenance.

This PR adds a `spec.version` field to the `ApicurioRegistry3` custom resource that allows users to explicitly control which version of the registry is deployed, independently of the operator version. This enables a staged upgrade process (Operator first, verify, then Registry).

### Changes Made

* Added an optional `spec.version` field to the `ApicurioRegistry3Spec` model.
* Updated `ResourceFactory.java` to inject the user-specified `spec.version` into the Application and UI container images instead of strictly defaulting to the operator's bundled version.
* Included a `compareVersions` utility in `Configuration.java` to handle numerical semver comparisons (handling `-SNAPSHOT` tags appropriately).
* Injected validation checks into `ResourceFactory` that automatically record `ValidationErrorCondition` messages:
    * errors if `spec.version` is newer than the operator version (operator may not support it).
    * warns if `spec.version` is older than the operator version (intentional downgrades).
*  Added unit tests in `ConfigurationTest.java` validating version comparisons and image tag replacement logic.

### DIFF
--- a/operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
+++ b/operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
@@ -4,6 +4,7 @@ import org.eclipse.microprofile.config.Config;
 import org.eclipse.microprofile.config.ConfigProvider;
 
 import java.util.Optional;
+import java.util.OptionalInt;
 
 public class Configuration {
 
@@ -38,6 +39,42 @@ public class Configuration {
     public static String getRegistryVersion() {
         return config.getOptionalValue("registry.version", String.class)
                 .orElseThrow(() -> new OperatorException("Required configuration option 'registry.version' is not set."));
+    }
+
+    /**
+     * Return the image with its tag replaced by the given version. If version is null or blank, returns the
+     * base image unchanged.
+     */
+    public static String getImageForVersion(String baseImage, String version) {
+        if (version == null || version.isBlank()) {
+            return baseImage;
+        }
+        int colon = baseImage.lastIndexOf(':');
+        String repo = (colon > 0) ? baseImage.substring(0, colon) : baseImage;
+        return repo + ":" + version;
+    }
+
+    /**
+     * Compare two version strings (e.g. "3.0.3" or "3.3.1-SNAPSHOT"). Strips any suffix after a hyphen
+     * before comparing. Returns empty if either version is not a valid numeric semver. Otherwise returns
+     * negative if v1 < v2, zero if equal, positive if v1 > v2.
+     */
+    public static OptionalInt compareVersions(String v1, String v2) {
+        try {
+            String[] parts1 = v1.split("-")[0].split("\\.");
+            String[] parts2 = v2.split("-")[0].split("\\.");
+            int len = Math.max(parts1.length, parts2.length);
+            for (int i = 0; i < len; i++) {
+                int n1 = i < parts1.length ? Integer.parseInt(parts1[i]) : 0;
+                int n2 = i < parts2.length ? Integer.parseInt(parts2[i]) : 0;
+                if (n1 != n2) {
+                    return OptionalInt.of(Integer.compare(n1, n2));
+                }
+            }
+            return OptionalInt.of(0);
+        } catch (NumberFormatException e) {
+            return OptionalInt.empty();
+        }
     }
 
     public static String getDefaultBaseHost() {
--- a/operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
+++ b/operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
@@ -49,6 +49,10 @@ public class Configuration {
         if (version == null || version.isBlank()) {
             return baseImage;
         }
+        // Digest-pinned images (e.g. repo@sha256:...) cannot have their tag replaced.
+        if (baseImage.contains("@")) {
+            return baseImage;
+        }
         int colon = baseImage.lastIndexOf(':');
         String repo = (colon > 0) ? baseImage.substring(0, colon) : baseImage;
         return repo + ":" + version;
--- a/operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
+++ b/operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
@@ -53,6 +53,39 @@ public class ResourceFactory {
 
     public static final ResourceFactory INSTANCE = new ResourceFactory();
 
+    private static String getSpecVersion(ApicurioRegistry3 primary) {
+        return ofNullable(primary.getSpec())
+                .map(ApicurioRegistry3Spec::getVersion)
+                .orElse(null);
+    }
+
+    /**
+     * Validate spec.version against the operator's bundled version. Records a validation error if
+     * spec.version is newer than the operator (unsupported) or not a valid semver. Logs a warning
+     * for intentional downgrades since those are the expected staged-upgrade workflow.
+     */
+    static void validateSpecVersion(ApicurioRegistry3 primary) {
+        var specVersion = getSpecVersion(primary);
+        if (specVersion == null) {
+            return;
+        }
+        var operatorVersion = Configuration.getRegistryVersion();
+        var cmp = Configuration.compareVersions(specVersion, operatorVersion);
+        if (cmp.isEmpty()) {
+            StatusManager.get(primary).getConditionManager(ValidationErrorConditionManager.class)
+                    .recordError("spec.version (%s) is not a valid version format. "
+                            + "Expected numeric semver (e.g. 3.0.3).", specVersion);
+        } else if (cmp.getAsInt() > 0) {
+            StatusManager.get(primary).getConditionManager(ValidationErrorConditionManager.class)
+                    .recordError("spec.version (%s) is newer than the operator version (%s). "
+                            + "The operator may not support managing a newer registry version.",
+                            specVersion, operatorVersion);
+        } else if (cmp.getAsInt() < 0) {
+            log.warn("spec.version ({}) is older than the operator version ({}). "
+                    + "This is a downgrade. Ensure this is intentional.", specVersion, operatorVersion);
+        }
+    }
+
     public static final String COMPONENT_APP = "app";
     public static final String COMPONENT_UI = "ui";
     public static final String COMPONENT_CONSOLE_PLUGIN = "console-plugin";
@@ -70,6 +103,8 @@ public class ResourceFactory {
     public static final String RESOURCE_TYPE_HORIZONTAL_POD_AUTOSCALER = "horizontalpodautoscaler";
 
     public Deployment getDefaultAppDeployment(ApicurioRegistry3 primary) {
+        validateSpecVersion(primary);
+
         var autoscaling = ofNullable(primary.getSpec()).map(ApicurioRegistry3Spec::getApp)
                 .map(ComponentSpec::getAutoscaling).orElse(null);
         boolean autoscalingEnabled = ofNullable(autoscaling).map(AutoscalingSpec::getEnabled)
@@ -128,7 +163,7 @@ public class ResourceFactory {
                 primary,
                 r.getSpec().getTemplate(),
                 REGISTRY_APP_CONTAINER_NAME,
-                Configuration.getAppImage(),
+                Configuration.getImageForVersion(Configuration.getAppImage(), getSpecVersion(primary)),
                 containerPort,
                 readinessProbe,
                 livenessProbe,
@@ -143,6 +178,8 @@ public class ResourceFactory {
     }
 
     public Deployment getDefaultUIDeployment(ApicurioRegistry3 primary) {
+        validateSpecVersion(primary);
+
         var uiAutoscaling = ofNullable(primary.getSpec()).map(ApicurioRegistry3Spec::getUi)
                 .map(ComponentSpec::getAutoscaling).orElse(null);
         boolean uiAutoscalingEnabled = ofNullable(uiAutoscaling).map(AutoscalingSpec::getEnabled)
@@ -173,7 +210,7 @@ public class ResourceFactory {
                 primary,
                 r.getSpec().getTemplate(),
                 REGISTRY_UI_CONTAINER_NAME,
-                Configuration.getUIImage(),
+                Configuration.getImageForVersion(Configuration.getUIImage(), getSpecVersion(primary)),
                 List.of(new ContainerPortBuilder().withName("http").withProtocol("TCP").withContainerPort(8080).build()),
                 new ProbeBuilder().withHttpGet(new HTTPGetActionBuilder().withPath("/config.js").withPort(new IntOrString(8080)).withScheme("HTTP").build()).build(),
                 new ProbeBuilder().withHttpGet(new HTTPGetActionBuilder().withPath("/config.js").withPort(new IntOrString(8080)).withScheme("HTTP").build()).build(),
@@ -464,12 +501,13 @@ public class ResourceFactory {
     }
 
     private void addDefaultLabels(Map<String, String> labels, ApicurioRegistry3 primary, String component) {
+        var version = ofNullable(getSpecVersion(primary)).orElse(Configuration.getRegistryVersion());
         labels.putAll(Map.of(
                 "app", primary.getMetadata().getName(),
                 "app.kubernetes.io/name", "apicurio-registry",
                 "app.kubernetes.io/component", component,
                 "app.kubernetes.io/instance", primary.getMetadata().getName(),
-                "app.kubernetes.io/version", Configuration.getRegistryVersion(),
+                "app.kubernetes.io/version", version,
                 "app.kubernetes.io/part-of", "apicurio-registry",
                 "app.kubernetes.io/managed-by", "apicurio-registry-operator"
         ));
--- a/operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
+++ b/operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
@@ -64,7 +64,7 @@ public class ResourceFactory {
      * spec.version is newer than the operator (unsupported) or not a valid semver. Logs a warning
      * for intentional downgrades since those are the expected staged-upgrade workflow.
      */
-    static void validateSpecVersion(ApicurioRegistry3 primary) {
+    public static void validateSpecVersion(ApicurioRegistry3 primary) {
         var specVersion = getSpecVersion(primary);
         if (specVersion == null) {
             return;
@@ -72,9 +72,8 @@ public class ResourceFactory {
         var operatorVersion = Configuration.getRegistryVersion();
         var cmp = Configuration.compareVersions(specVersion, operatorVersion);
         if (cmp.isEmpty()) {
-            StatusManager.get(primary).getConditionManager(ValidationErrorConditionManager.class)
-                    .recordError("spec.version (%s) is not a valid version format. "
-                            + "Expected numeric semver (e.g. 3.0.3).", specVersion);
+            log.warn("spec.version ({}) is not a recognized numeric format. "
+                    + "Using as-is for image tag.", specVersion);
         } else if (cmp.getAsInt() > 0) {
             StatusManager.get(primary).getConditionManager(ValidationErrorConditionManager.class)
                     .recordError("spec.version (%s) is newer than the operator version (%s). "
@@ -103,8 +102,6 @@ public class ResourceFactory {
     public static final String RESOURCE_TYPE_HORIZONTAL_POD_AUTOSCALER = "horizontalpodautoscaler";
 
     public Deployment getDefaultAppDeployment(ApicurioRegistry3 primary) {
-        validateSpecVersion(primary);
-
         var autoscaling = ofNullable(primary.getSpec()).map(ApicurioRegistry3Spec::getApp)
                 .map(ComponentSpec::getAutoscaling).orElse(null);
         boolean autoscalingEnabled = ofNullable(autoscaling).map(AutoscalingSpec::getEnabled)
@@ -178,8 +175,6 @@ public class ResourceFactory {
     }
 
     public Deployment getDefaultUIDeployment(ApicurioRegistry3 primary) {
-        validateSpecVersion(primary);
-
         var uiAutoscaling = ofNullable(primary.getSpec()).map(ApicurioRegistry3Spec::getUi)
                 .map(ComponentSpec::getAutoscaling).orElse(null);
         boolean uiAutoscalingEnabled = ofNullable(uiAutoscaling).map(AutoscalingSpec::getEnabled)
--- a/operator/model/src/main/java/io/apicurio/registry/operator/api/v1/ApicurioRegistry3Spec.java
+++ b/operator/model/src/main/java/io/apicurio/registry/operator/api/v1/ApicurioRegistry3Spec.java
@@ -13,7 +13,7 @@ import static com.fasterxml.jackson.annotation.Nulls.SKIP;
 import static lombok.AccessLevel.PRIVATE;
 
 @JsonInclude(NON_NULL)
-@JsonPropertyOrder({"app", "ui", "consolePlugin"})
+@JsonPropertyOrder({"version", "app", "ui", "consolePlugin"})
 @JsonDeserialize(using = None.class)
 @NoArgsConstructor
 @AllArgsConstructor(access = PRIVATE)
@@ -23,6 +23,18 @@ import static lombok.AccessLevel.PRIVATE;
 @EqualsAndHashCode
 @ToString
 public class ApicurioRegistry3Spec {
+
+    /**
+     * Override the registry version deployed by this operator. When set, the operator replaces the image tag
+     * with this value. When not set, defaults to the version bundled with the operator.
+     */
+    @JsonProperty("version")
+    @JsonPropertyDescription("""
+            Override the registry version deployed by this operator. \
+            When set, the operator replaces the image tag with this value. \
+            When not set, defaults to the version bundled with the operator.""")
+    @JsonSetter(nulls = SKIP)
+    private String version;
 
     /**
      * Configure Apicurio Registry backend (app) component.
--- a/operator/controller/src/main/java/io/apicurio/registry/operator/ApicurioRegistry3Reconciler.java
+++ b/operator/controller/src/main/java/io/apicurio/registry/operator/ApicurioRegistry3Reconciler.java
@@ -197,6 +197,8 @@ public class ApicurioRegistry3Reconciler implements Reconciler<ApicurioRegistry3
 
         log.trace("Reconciling Apicurio Registry: {}", primary);
 
+        io.apicurio.registry.operator.resource.ResourceFactory.validateSpecVersion(primary);
+
         // Some of the fields in the CR have been deprecated and another fields should be used instead.
         // Operator will attempt to update the CR to use the newer fields if possible.
         // This has to be done first, so subsequent functionality can deal with new fields only.
--- a/operator/controller/src/test/java/io/apicurio/registry/operator/unit/ConfigurationTest.java
+++ b/operator/controller/src/test/java/io/apicurio/registry/operator/unit/ConfigurationTest.java
@@ -1,0 +1,95 @@
+package io.apicurio.registry.operator.unit;
+
+import io.apicurio.registry.operator.Configuration;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ConfigurationTest {
+
+    @Test
+    public void testGetImageForVersion_swapsTag() {
+        assertThat(Configuration.getImageForVersion("quay.io/apicurio/apicurio-registry:3.0.6", "3.0.3"))
+                .isEqualTo("quay.io/apicurio/apicurio-registry:3.0.3");
+    }
+
+    @Test
+    public void testGetImageForVersion_nullVersionReturnsBaseImage() {
+        assertThat(Configuration.getImageForVersion("quay.io/apicurio/apicurio-registry:3.0.6", null))
+                .isEqualTo("quay.io/apicurio/apicurio-registry:3.0.6");
+    }
+
+    @Test
+    public void testGetImageForVersion_blankVersionReturnsBaseImage() {
+        assertThat(Configuration.getImageForVersion("quay.io/apicurio/apicurio-registry:3.0.6", "  "))
+                .isEqualTo("quay.io/apicurio/apicurio-registry:3.0.6");
+    }
+
+    @Test
+    public void testGetImageForVersion_imageWithoutTag() {
+        assertThat(Configuration.getImageForVersion("quay.io/apicurio/apicurio-registry", "3.0.3"))
+                .isEqualTo("quay.io/apicurio/apicurio-registry:3.0.3");
+    }
+
+    @Test
+    public void testGetImageForVersion_latestSnapshotTag() {
+        assertThat(Configuration.getImageForVersion("quay.io/apicurio/apicurio-registry:latest-snapshot", "3.0.3"))
+                .isEqualTo("quay.io/apicurio/apicurio-registry:3.0.3");
+    }
+
+    @Test
+    public void testGetImageForVersion_uiImage() {
+        assertThat(Configuration.getImageForVersion("quay.io/apicurio/apicurio-registry-ui:3.0.6", "3.0.3"))
+                .isEqualTo("quay.io/apicurio/apicurio-registry-ui:3.0.3");
+    }
+
+    @Test
+    public void testCompareVersions_equal() {
+        assertThat(Configuration.compareVersions("3.0.3", "3.0.3")).hasValue(0);
+    }
+
+    @Test
+    public void testCompareVersions_olderPatch() {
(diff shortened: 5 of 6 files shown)


operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
```diff
@@ -70,6 +76,24 @@ public class ResourceFactory {
...
+        var specVersion = getSpecVersion(primary);
+        if (specVersion != null) {
+            var operatorVersion = Configuration.getRegistryVersion();
+            int cmp = Configuration.compareVersions(specVersion, operatorVersion);
+            if (cmp > 0) {
+                StatusManager.get(primary).getConditionManager(ValidationErrorConditionManager.class)
+                        .recordError("spec.version (%s) is newer than the operator version (%s). "
+                                + "The operator may not support managing a newer registry version.",
+                                specVersion, operatorVersion);
+            } else if (cmp < 0) {
```
paoloantinori: This branch records a validation **error** for intentional downgrades, but the PR description frames downgrades as a *warning* case. Since `ValidationErrorConditionManager` only has an error channel (no warning level), a legitimate staged downgrade shows up identically to an unsupported-version error. Worth deciding whether downgrade should be a softer signal — otherwise the feature's main use case is penalised.

operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
```diff
@@ -40,6 +40,37 @@ public static String getRegistryVersion() {
...
+    /**
+     * Compare two version strings (e.g. "3.0.3" or "3.3.1-SNAPSHOT"). Strips any suffix after a hyphen
+     * before comparing. Returns negative if v1 < v2, zero if equal, positive if v1 > v2.
+     */
+    public static int compareVersions(String v1, String v2) {
+        String[] parts1 = v1.split("-")[0].split("\\.");
+        String[] parts2 = v2.split("-")[0].split("\\.");
+        int len = Math.max(parts1.length, parts2.length);
+        for (int i = 0; i < len; i++) {
+            int n1 = i < parts1.length ? Integer.parseInt(parts1[i]) : 0;
```
paoloantinori: `Integer.parseInt` here will throw `NumberFormatException` for non-numeric `spec.version` values like `latest`, `3.x`, or `3.0.6.Final`. That escapes uncaught during reconciliation instead of surfacing as a validation error. Consider wrapping the parse (or validating the format upstream) so an invalid `spec.version` produces a clear `ValidationError` rather than crashing the reconcile.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
```diff
@@ -128,7 +152,7 @@ public Deployment getDefaultAppDeployment(ApicurioRegistry3 primary) {
                 primary,
```
paoloantinori, line 160: The version-mismatch validation only runs in `getDefaultAppDeployment`. Since the UI deployment also applies `spec.version` via `getImageForVersion` (line 200), a version mismatch is silently skipped when only the UI component is reconciled. Lifting the check to a place that covers both components (or to the controller's validate phase) would make this consistent.

paoloantinori (commented):
Nice feature — decoupling the registry version from the operator via `spec.version` is a real improvement for staged upgrades, and the CR/condition plumbing looks reasonable. A few things worth addressing before this lands:

- **Downgrade is reported as a validation *error*, not a warning.** The PR description says the operator should *warn* when `spec.version` is older than the operator version (the intentional-downgrade case). But `ResourceFactory` calls `recordError(...)` for **both** the newer and older branches (lines 84-93). `ValidationErrorConditionManager.recordError` sets a full `ValidationError` condition (`status=True, reason=Error`) — there's no warning channel. So a legitimate staged downgrade (the exact scenario motivating this PR) surfaces to the user as a hard validation error, same severity as an unsupported newer version. Worth confirming whether downgrade should instead be a non-error signal (e.g. a separate condition type, or just not recorded), since treating intentional downgrades as errors undermines the use case.
- **`compareVersions` throws on non-numeric `spec.version`.** `Integer.parseInt(parts1[i])` at `Configuration.java:65` will throw `NumberFormatException` for values like `latest`, `3.x`, or `3.0.6.Final`. That exception escapes during reconciliation rather than producing a friendly validation error. A user setting `spec.version: latest` is plausible and currently crashes the reconcile loop. Suggest catching the parse and recording a validation error instead, and adding a test for it.
- **Version validation is only wired into the app deployment.** The check lives in `getDefaultAppDeployment` (line 78), but `getDefaultUIDeployment` (line 188) also consumes `getImageForVersion(... spec.version)`. If a CR is configured with only a UI component (or app is disabled), the version mismatch is never validated. Consider lifting the validation to a single spot that runs for both paths, or into the controller's validate step.
- Minor: the test suite covers `getImageForVersion` and `compareVersions` well, but there's no test asserting the recorded `ValidationError` condition message/format, which is the part users will actually see.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
hemantch01: changed the downgrade path from recordError() to log.warn() since intentional downgrades are the primary use case for this feature, newer than operator and invalid format still surface as ValidationError.

operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
hemantch01: compareVersions now returns OptionalInt  catches NumberFormatException and returns empty(). validateSpecVersion records a clear validation error: "spec.version (latest) is not a valid version format. Expected numeric semver (e.g. 3.0.3)." Added tests.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
hemantch01: extracted validation into a shared validateSpecVersion() method, now called from both getDefaultAppDeployment() and getDefaultUIDeployment().

paoloantinori (commented):
Nice feature — decoupling the registry version from the operator via `spec.version` is a real improvement for staged upgrades, and the CR/condition plumbing looks reasonable. A few things worth addressing before this lands:

- **Downgrade is reported as a validation *error*, not a warning.** The PR description says the operator should *warn* when `spec.version` is older than the operator version (the intentional-downgrade case). But `ResourceFactory` calls `recordError(...)` for **both** the newer and older branches (lines 84-93). `ValidationErrorConditionManager.recordError` sets a full `ValidationError` condition (`status=True, reason=Error`) — there's no warning channel. So a legitimate staged downgrade (the exact scenario motivating this PR) surfaces to the user as a hard validation error, same severity as an unsupported newer version. Worth confirming whether downgrade should instead be a non-error signal (e.g. a separate condition type, or just not recorded), since treating intentional downgrades as errors undermines the use case.
- **`compareVersions` throws on non-numeric `spec.version`.** `Integer.parseInt(parts1[i])` at `Configuration.java:65` will throw `NumberFormatException` for values like `latest`, `3.x`, or `3.0.6.Final`. That exception escapes during reconciliation rather than producing a friendly validation error. A user setting `spec.version: latest` is plausible and currently crashes the reconcile loop. Suggest catching the parse and recording a validation error instead, and adding a test for it.
- **Version validation is only wired into the app deployment.** The check lives in `getDefaultAppDeployment` (line 78), but `getDefaultUIDeployment` (line 188) also consumes `getImageForVersion(... spec.version)`. If a CR is configured with only a UI component (or app is disabled), the version mismatch is never validated. Consider lifting the validation to a single spot that runs for both paths, or into the controller's validate step.
- Minor: the test suite covers `getImageForVersion` and `compareVersions` well, but there's no test asserting the recorded `ValidationError` condition message/format, which is the part users will actually see.

hemantch01:
hi @paoloantinori i have attended All points and pushed changes.

paoloantinori:
Thanks, @hemantch01 — all three points from my earlier review are addressed in this push:

- **Downgrade vs. newer-version severity:** `validateSpecVersion` now logs a `warn` for the `cmp < 0` (downgrade) branch and reserves `recordError(...)` for the unsupported-newer and invalid-format cases. That matches the staged-upgrade intent.
- **Non-numeric `spec.version`:** `compareVersions` now wraps the parse in try/catch and returns `OptionalInt.empty()`, and the invalid-format case is surfaced as a `ValidationError` rather than crashing the reconcile.
- **Validation coverage:** `validateSpecVersion(primary)` is now called at the top of both `getDefaultAppDeployment` and `getDefaultUIDeployment`, so the UI-only path is covered too.

The new `ConfigurationTest` cases (`nonNumericReturnsEmpty`, `qualifierReturnsEmpty`, `alphaPartReturnsEmpty`, `differentLengths`) are good additions. From my side the head looks ready — deferring the final call to a maintainer.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
```diff
@@ -143,6 +178,8 @@ public Deployment getDefaultAppDeployment(ApicurioRegistry3 primary) {
     }
 
     public Deployment getDefaultUIDeployment(ApicurioRegistry3 primary) {
+        validateSpecVersion(primary);
```
aclfe: Validation is called twice It's called from both the app and UI deployment builders, so it runs and logs the downgrade warning double time

Maybe you can run it in the reconciler, not inside the resource builders?

operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
```diff
@@ -40,6 +41,42 @@ public static String getRegistryVersion() {
...
+        if (version == null || version.isBlank()) {
+            return baseImage;
+        }
+        int colon = baseImage.lastIndexOf(':');
+        String repo = (colon > 0) ? baseImage.substring(0, colon) : baseImage;
+        return repo + ":" + version;
+    }
+
+    /**
+     * Compare two version strings (e.g. "3.0.3" or "3.3.1-SNAPSHOT"). Strips any suffix after a hyphen
```
aclfe, line 62: Just to confirm, is this swapping the image tag?  Production images are can be locked to a certain digest (that is released), even if the version is rewritten, this might not work for production path? 

Maybe this is just a code quality nit, but the tag parser should handle `@sha256:` digests too?

operator/model/src/main/java/io/apicurio/registry/operator/api/v1/ApicurioRegistry3Spec.java
```diff
@@ -24,6 +24,18 @@
...
+     * Override the registry version deployed by this operator. When set, the operator replaces the image tag
+     * with this value. When not set, defaults to the version bundled with the operator.
+     */
+    @JsonProperty("version")
+    @JsonPropertyDescription("""
+            Override the registry version deployed by this operator. \
+            When set, the operator replaces the image tag with this value. \
+            When not set, defaults to the version bundled with the operator.""")
+    @JsonSetter(nulls = SKIP)
+    private String version;
```
aclfe, line 37: Here `version` is added but the generated CRD file apicurioregistries3.registry.apicur.io-v1.yml is not showing diff

`./mvnw clean install -pl operator/model,operator/controller -am -DskipTests`

aclfe (commented):
Hello, great pr. I think we can add more tests onto this. Maybe some tests for the validation logic, some for deployed container actually gets the new version, and some for the version label. Since they're new paths and that also needs tests

paoloantinori:
Good catch — this is a real gap, and I missed it in my earlier review. Without regenerating the CRD, `spec.version` won't actually reach the controller in a deployed cluster: the API server strips unknown fields (or rejects them, depending on the CRD's `preserveUnknownFields` setting), so the feature is effectively inert until `apicurioregistries3.registry.apicur.io-v1.yml` is regenerated and committed.

Running the model/controller build as you noted should regenerate the CRD with the new `version` property under `spec`. That regenerated YAML needs to be part of this diff before merge — otherwise users setting `spec.version` will see it silently ignored.

@hemantch01 could you regenerate the CRD (`./mvnw clean install -pl operator/model,operator/controller -am -DskipTests`) and add the updated YAML to the PR?

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
hemantch01: thanks I will move validateSpecVersion() into ApicurioRegistry3Reconciler.reconcile() so it runs once per loop, will remove from both resource builders.

operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
hemantch01: The issue scope (#7635) is about decoupling operator/registry versions via spec.version, targeting the standard image:tag format used by the operator's default images. Digest-pinned images (@sha256:) are a separate production concern if maintainers (@carlesarnal )want that handled here I'm happy to add it, but it feels like a follow-up to keep this PR focused.

operator/model/src/main/java/io/apicurio/registry/operator/api/v1/ApicurioRegistry3Spec.java
hemantch01: could you clarify which CRD file path should be updated? The file at operator/controller/src/main/deploy/crd/ is regenerated during mvn install and I don't see it tracked in main

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
```diff
@@ -53,6 +53,39 @@ public class ResourceFactory {
...
+                .map(ApicurioRegistry3Spec::getVersion)
+                .orElse(null);
+    }
+
+    /**
+     * Validate spec.version against the operator's bundled version. Records a validation error if
+     * spec.version is newer than the operator (unsupported) or not a valid semver. Logs a warning
+     * for intentional downgrades since those are the expected staged-upgrade workflow.
+     */
+    public static void validateSpecVersion(ApicurioRegistry3 primary) {
```
paoloantinori, line 67: This records a `ValidationErrorCondition`, but nothing in the reconcile workflow reads it back to gate the deployment. `AppDeploymentResource.desired()` ([source](https://github.com/Apicurio/apicurio-registry/blob/main/operator/controller/src/main/java/io/apicurio/registry/operator/resource/app/AppDeploymentResource.java)) calls `APP_DEPLOYMENT_KEY.getFactory().apply(primary)` unconditionally. Net effect: a `spec.version` newer than the operator still produces a pod with the newer image — the error condition shows up in status *after* the unsupported image is already rolled out. Consider having the app/UI `ActivationCondition` check `ValidationErrorConditionManager.hasErrors()` so an unsupported version is refused before the Deployment is patched.

operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
```diff
@@ -40,6 +41,42 @@ public static String getRegistryVersion() {
...
+            for (int i = 0; i < len; i++) {
+                int n1 = i < parts1.length ? Integer.parseInt(parts1[i]) : 0;
+                int n2 = i < parts2.length ? Integer.parseInt(parts2[i]) : 0;
+                if (n1 != n2) {
+                    return OptionalInt.of(Integer.compare(n1, n2));
+                }
+            }
+            return OptionalInt.of(0);
+        } catch (NumberFormatException e) {
+            return OptionalInt.empty();
```
paoloantinori, line 80: `compareVersions` returns `OptionalInt.empty()` for any non-numeric segment, which means a perfectly valid image tag like `3.0.6.Final` or `latest-release` gets reported as a validation **error** rather than a warning. Since `getImageForVersion` happily substitutes any string into the tag, the version-format check is stricter than the actual capability. Worth treating non-numeric as a warn ("version format not recognized; using as-is") rather than blocking-equivalent error.

paoloantinori (commented):
Third pass on this — the new `SpecVersionTest` covers the image-swap + version-label path well, and the CRD gap from my last review is now closed. One new substantive concern stands out after reading how `validateSpecVersion` actually plugs into the reconcile loop.

**`ValidationErrorCondition` is informational only — it does not gate the rollout.** `ResourceFactory.validateSpecVersion()` records an error when `spec.version` is newer than the operator, but no `ActivationCondition` consults `ValidationErrorConditionManager.hasErrors()`. `AppDeploymentResource.desired()` (and the UI equivalent) just call the factory and apply the resulting Deployment unconditionally. So a user setting `spec.version: 9.9.9` against an older operator gets a `ValidationError` condition in status **and** a pod scheduled with image `:9.9.9` — exactly the unsupported combination the validation is meant to prevent. The PR description frames this as a safety mechanism for staged upgrades; as written it's a status signal, not a guardrail. Either the activation conditions for the app/UI deployments should short-circuit when a validation error is present, or the doc/PR description should be explicit that `spec.version` newer-than-operator is *best-effort warned*, not blocked.

Two smaller points worth a look:
- `Configuration.compareVersions` uses `Integer.parseInt`, so a value like `3.0.6.Final` returns `OptionalInt.empty()` and is then classified by `validateSpecVersion` as "not a valid version format" → `recordError`. But `3.0.6.Final` is a legitimate image tag a user might pass to pin a release. Consider treating the comparison result as advisory (warn) rather than a hard error when the format is merely non-numeric, since the image tag substitution itself works fine with non-numeric strings.
- `SpecVersionTest.testVersionOverrideUpdatesImageAndLabel` asserts on `app.kubernetes.io/version` from `getMetadata().getLabels()`, but `addDefaultLabels` is applied via `ResourceFactory` on the pod template / selector labels in several call sites — worth confirming the assertion isn't passing against a label set that the actual Deployment object exposes elsewhere than where Kubernetes reads it.

hemantch01:
> Hello, great pr. I think we can add more tests onto this. Maybe some tests for the validation logic, some for deployed container actually gets the new version, and some for the version label. Since they're new paths and that also needs tests

The core logic (compareVersions, getImageForVersion) has 16 unit tests covering all edge cases. The rest is standard Kubernetes resource wiring happy to add more if maintainers feel it's needed.

operator/controller/src/main/java/io/apicurio/registry/operator/resource/ResourceFactory.java
hemantch01: It is Intentional blocking the deployment would break the feature's primary use case. During a staged upgrade, spec.version will always differ from the operator version (that's the point). If we gate on ValidationErrorConditionManager.hasErrors(), the operator refuses to deploy any version that doesn't exactly match its own, making spec.version unusable. The validation is advisory: it surfaces in CR status so the user knows what they're running, but it doesn't prevent the rollout. I'll clarify this in the PR description.

operator/controller/src/main/java/io/apicurio/registry/operator/Configuration.java
hemantch01: 3.0.6.Final is a valid image tag. Will change non-numeric versions from recordError to log.warn since getImageForVersion handles any string fine

aclfe:
> could you clarify which CRD file path should be updated? The file at operator/controller/src/main/deploy/crd/ is regenerated during mvn install and I don't see it tracked in main

Run these, this is the one that's tracked: 
```
  cd operator
  make VERSION=3.3.1-SNAPSHOT IMAGE_TAG=latest-snapshot INSTALL_FILE=install/install.yaml dist-install-file
  cd ..
```
You'll see a diff in `operator/install/install.yaml`

operator/model/src/main/java/io/apicurio/registry/operator/api/v1/ApicurioRegistry3Spec.java
aclfe: https://github.com/Apicurio/apicurio-registry/pull/8875#issuecomment-5082154825

jsenko (commented):
## Review: spec.version — design concerns

@hemantch01 — thank you for the work on this. The implementation is clean and addresses the issue as described. However, after deeper analysis, the problem turns out to be more complex than what #7635 originally scoped. I want to flag several concerns that need discussion before we can merge this, and I'd like input from @carlesarnal (issue author), @NikishaPatil (productization), and @EricWittmann.

The use case (staged upgrades, emergency rollback) is real and valid, but we need further discussion to make sure the solution handles all our deployment scenarios.

---

### 1. Configuration property compatibility

The operator sets ~60 environment variables on the registry deployment (auth, KafkaSQL, CORS, GitOps, storage, OTel, etc. — see `EnvironmentVariables.java`). These env vars are version-coupled: the operator assumes the registry it manages understands all of them.

With `spec.version`, the operator would deploy registry version X but configure it with env vars designed for version Y (the operator's own version). Two failure modes:

- **Older registry, newer operator:** The operator may set env vars (e.g. `APICURIO_GITOPS_*`, `APICURIO_SEARCH_INDEX_*`, Kubernetes auth vars) that didn't exist in the older registry. Quarkus ignores unknown env vars, so this is usually safe — but if the operator *renames* a property (we've done this, e.g. the KafkaSQL env var prefix migration from `APICURIO_KAFKASQL_*` to `APICURIO_KAFKA_COMMON_*`), the older registry won't receive the config it needs.

- **Newer registry, older operator:** The registry may require env vars that the operator doesn't know to set. This is the scenario the PR already flags as a `ValidationError`, but the deployment still proceeds.

Currently the PR does not restrict the version range — any version string is accepted. A user could deploy a 3.0.x registry with a 3.3.x operator and hit subtle config mismatches that are hard to diagnose.

**For comparison:** Strimzi maintains a curated `kafka-versions.yaml` file listing exactly which Kafka versions each operator release supports. The version→image mapping is injected as the `STRIMZI_KAFKA_IMAGES` env var on the operator deployment at build time. If `spec.kafka.version` is not in that list, the operator throws `UnsupportedKafkaVersionException`. No freeform version strings.

### 2. PodTemplateSpec already supports image overrides

The operator already supports overriding the container image via `spec.app.podTemplateSpec`:

```yaml
spec:
  app:
    podTemplateSpec:
      spec:
        containers:
          - name: apicurio-registry-app
            image: quay.io/apicurio/apicurio-registry:3.2.6
```

This is more verbose but also more powerful — it accepts any image reference (tags, digests, custom builds), and because the user is explicitly choosing an image, they're more likely to also handle any config adjustments via `spec.app.env`. The `spec.version` convenience hides the config compatibility risk because it looks like a simple, safe knob.

### 3. Downstream (productized) image pinning — broken

In the downstream Red Hat build, operand images are pinned by SHA digest, not by tag. I verified this by inspecting the bundle image `brew.registry.redhat.io/rh-osbs/apicurio-apicurio-registry-3-operator-bundle:3.2.5.r1-4`:

```
RELATED_IMAGE_REGISTRY_APP_IMAGE=registry.redhat.io/apicurio/apicurio-registry-rhel9@sha256:7d4a8f6e...
RELATED_IMAGE_REGISTRY_UI_IMAGE=registry.redhat.io/apicurio/apicurio-registry-ui-rhel9@sha256:c678c969...
```

The `getImageForVersion()` function uses `lastIndexOf(':')` to split the image reference and replace the tag. On a SHA-pinned image like `registry.redhat.io/...@sha256:7d4a8f6e...`, this finds the `:` inside the digest, producing an invalid reference like `registry.redhat.io/...@sha256:3.2.5`.

This means in downstream we'd need to either:
- Disable the feature entirely (maintenance burden of a downstream-only code path), or
- Provide a version→SHA mapping for previous versions (significant build/release process changes)

@NikishaPatil — would appreciate your input on how this would interact with the productization pipeline.

### 4. Scope vs. complexity — is there a simpler approach?

The original issue #7635 asks for two things: staged upgrades (update operator first, then registry) and emergency rollback. The current `spec.version` approach provides maximum flexibility (any version), but that flexibility brings the three problems above.

Some alternative approaches that might give us the benefit with less risk:

**Option A: Patch-only version restriction.** Only allow `spec.version` within the same minor version as the operator (e.g. operator 3.3.1 allows 3.3.0, 3.3.1, but rejects 3.2.x). Config properties are guaranteed stable across patch versions, so the compatibility concern disappears. This covers staged patch upgrades and emergency rollback to the previous patch.

**Option B: Update pause.** A simple `spec.updatePolicy: Paused | Auto` field. When `Paused`, the operator reconciles everything (config, services, ingress) but doesn't change the container image. The user updates the operator, verifies it's healthy, then sets `Auto` to let it update the registry images. No version string manipulation, works with SHA-pinned images. Doesn't provide rollback on its own, but could be combined with tracking the previous image in CR status.

**Option C: Curated version mapping (Strimzi-style).** Embed a list of supported registry versions and their images in the operator, injected as env vars at build time. Most robust, but highest maintenance cost — every release must declare which older versions it supports, and downstream must generate SHA mappings for each.

@carlesarnal — would appreciate your input since you filed #7635.

---

### Summary

The implementation quality is good — the `compareVersions`, `getImageForVersion`, and validation logic are well-structured and tests are solid. The concerns above are about the feature's design scope, not the code quality.

I think we need to:
1. Decide on the version restriction policy (freeform vs. patch-only vs. curated list)
2. Resolve the downstream SHA-pinned image incompatibility
3. Consider whether `PodTemplateSpec` image override + a simpler update-pause mechanism covers the use case sufficiently

Happy to discuss further — this is a valuable feature, we just want to get the boundaries right.

jsenko (changes requested):
## Review: spec.version — design concerns

@hemantch01 — thank you for the work on this. The implementation is clean and addresses the issue as described. However, after deeper analysis, the problem turns out to be more complex than what #7635 originally scoped. I want to flag several concerns that need discussion before we can merge this, and I'd like input from @carlesarnal (issue author), @NikishaPatil (productization), and @EricWittmann.

The use case (staged upgrades, emergency rollback) is real and valid, but we need further discussion to make sure the solution handles all our deployment scenarios.

---

### 1. Configuration property compatibility

The operator sets ~60 environment variables on the registry deployment (auth, KafkaSQL, CORS, GitOps, storage, OTel, etc. — see `EnvironmentVariables.java`). These env vars are version-coupled: the operator assumes the registry it manages understands all of them.

With `spec.version`, the operator would deploy registry version X but configure it with env vars designed for version Y (the operator's own version). Two failure modes:

- **Older registry, newer operator:** The operator may set env vars (e.g. `APICURIO_GITOPS_*`, `APICURIO_SEARCH_INDEX_*`, Kubernetes auth vars) that didn't exist in the older registry. Quarkus ignores unknown env vars, so this is usually safe — but if the operator *renames* a property (we've done this, e.g. the KafkaSQL env var prefix migration from `APICURIO_KAFKASQL_*` to `APICURIO_KAFKA_COMMON_*`), the older registry won't receive the config it needs.

- **Newer registry, older operator:** The registry may require env vars that the operator doesn't know to set. This is the scenario the PR already flags as a `ValidationError`, but the deployment still proceeds.

Currently the PR does not restrict the version range — any version string is accepted. A user could deploy a 3.0.x registry with a 3.3.x operator and hit subtle config mismatches that are hard to diagnose.

**For comparison:** Strimzi maintains a curated `kafka-versions.yaml` file listing exactly which Kafka versions each operator release supports. The version→image mapping is injected as the `STRIMZI_KAFKA_IMAGES` env var on the operator deployment at build time. If `spec.kafka.version` is not in that list, the operator throws `UnsupportedKafkaVersionException`. No freeform version strings.

### 2. PodTemplateSpec already supports image overrides

The operator already supports overriding the container image via `spec.app.podTemplateSpec`:

```yaml
spec:
  app:
    podTemplateSpec:
      spec:
        containers:
          - name: apicurio-registry-app
            image: quay.io/apicurio/apicurio-registry:3.2.6
```

This is more verbose but also more powerful — it accepts any image reference (tags, digests, custom builds), and because the user is explicitly choosing an image, they're more likely to also handle any config adjustments via `spec.app.env`. The `spec.version` convenience hides the config compatibility risk because it looks like a simple, safe knob.

### 3. Downstream (productized) image pinning — broken

In the downstream Red Hat build, operand images are pinned by SHA digest, not by tag. I verified this by inspecting the bundle image `brew.registry.redhat.io/rh-osbs/apicurio-apicurio-registry-3-operator-bundle:3.2.5.r1-4`:

```
RELATED_IMAGE_REGISTRY_APP_IMAGE=registry.redhat.io/apicurio/apicurio-registry-rhel9@sha256:7d4a8f6e...
RELATED_IMAGE_REGISTRY_UI_IMAGE=registry.redhat.io/apicurio/apicurio-registry-ui-rhel9@sha256:c678c969...
```

The `getImageForVersion()` function uses `lastIndexOf(':')` to split the image reference and replace the tag. On a SHA-pinned image like `registry.redhat.io/...@sha256:7d4a8f6e...`, this finds the `:` inside the digest, producing an invalid reference like `registry.redhat.io/...@sha256:3.2.5`.

This means in downstream we'd need to either:
- Disable the feature entirely (maintenance burden of a downstream-only code path), or
- Provide a version→SHA mapping for previous versions (significant build/release process changes)

@NikishaPatil — would appreciate your input on how this would interact with the productization pipeline.

### 4. Scope vs. complexity — is there a simpler approach?

The original issue #7635 asks for two things: staged upgrades (update operator first, then registry) and emergency rollback. The current `spec.version` approach provides maximum flexibility (any version), but that flexibility brings the three problems above.

Some alternative approaches that might give us the benefit with less risk:

**Option A: Patch-only version restriction.** Only allow `spec.version` within the same minor version as the operator (e.g. operator 3.3.1 allows 3.3.0, 3.3.1, but rejects 3.2.x). Config properties are guaranteed stable across patch versions, so the compatibility concern disappears. This covers staged patch upgrades and emergency rollback to the previous patch.

**Option B: Update pause.** A simple `spec.updatePolicy: Paused | Auto` field. When `Paused`, the operator reconciles everything (config, services, ingress) but doesn't change the container image. The user updates the operator, verifies it's healthy, then sets `Auto` to let it update the registry images. No version string manipulation, works with SHA-pinned images. Doesn't provide rollback on its own, but could be combined with tracking the previous image in CR status.

**Option C: Curated version mapping (Strimzi-style).** Embed a list of supported registry versions and their images in the operator, injected as env vars at build time. Most robust, but highest maintenance cost — every release must declare which older versions it supports, and downstream must generate SHA mappings for each.

@carlesarnal — would appreciate your input since you filed #7635.

---

### Summary

The implementation quality is good — the `compareVersions`, `getImageForVersion`, and validation logic are well-structured and tests are solid. The concerns above are about the feature's design scope, not the code quality.

I think we need to:
1. Decide on the version restriction policy (freeform vs. patch-only vs. curated list)
2. Resolve the downstream SHA-pinned image incompatibility
3. Consider whether `PodTemplateSpec` image override + a simpler update-pause mechanism covers the use case sufficiently

Happy to discuss further — this is a valuable feature, we just want to get the boundaries right.

hemantch01:
sorry, i was busy with college work.

jsenko:
@hemantch01 — thank you for the time and effort you put into this feature. The implementation is well-structured, and paoloantinori and aclfe's feedback rounds showed solid responsiveness.

After team discussion, we've decided not to move forward with this PR. The core issue is that our project doesn't currently have the infrastructure to safely support operator-operand version decoupling:

- **Database schema migrations are monotonic** — a single updated pod advances the schema, making rollback to older pods unsafe without restoring the database from backup. This limits the usefulness of version pinning for staged rollouts.
- **We lack a compatibility matrix** between operator and registry versions. Maintaining one (like Strimzi does) would require significant ongoing effort that we're not ready to commit to.
- **The alternative already exists** — `spec.app.podTemplateSpec` image overrides cover the power-user case, with the user explicitly managing compatibility.

We're going to focus instead on improving our compatibility tracking and migration documentation processes, which are prerequisites for a feature like this.

We appreciate your contribution — sorry we couldn't make it work this time. The underlying issue (#7635) will stay open for future consideration once we have better compatibility infrastructure in place.

hemantch01:
thanks for the detailed explanation the migration and compatibility concerns make sense. happy to revisit if #7635 picks back up once the compatibility infrastructure is in place.

PR #8876, opened 2026-07-24
fix(ui): guard direct navigation to /agents when Agents feature is disabled

### LINKED ISSUE #8747: Direct navigation to `/agents` is not guarded when the Agents feature is disabled
## Summary

When the **Agents** feature is disabled (the default local configuration), the **Agents** navigation item is correctly hidden from the UI.

However, the `/agents` route remains directly accessible. Navigating to the route manually (or through a bookmarked/shared URL) bypasses the feature flag and results in an application error instead of handling the disabled feature gracefully.

## Steps to reproduce

1. Start the Registry using the default local configuration
2. Open the application
3. Verify that the **Agents** tab is not visible in the navigation
4. Navigate directly to

```
http://localhost:8888/agents
```
## Actual behavior

The application attempts to load the Agents page and eventually displays:

> Error searching for agents.

## Expected behavior

Since the feature is disabled, direct navigation to `/agents` should not attempt to initialize the page.

Instead, the application should gracefully handle the request (for example by redirecting, showing a "feature unavailable" state, or another appropriate behavior).

## Why this is a bug

The feature flag is respected by the navigation but not by direct route access, resulting in inconsistent behavior.

A user opening a bookmarked URL or following a shared link reaches an internal error page even though the feature is intentionally disabled.

## Environment

- Fresh local checkout
- Default Registry configuration
- Agents feature disabled (default)
- Reproduced consistently

## Additional notes

This issue only affects direct navigation to the route. The navigation menu itself correctly hides the Agents entry when the feature is disabled.

Adii-45:
## Summary

Fixes the direct navigation issue reported in #8747.

When the Agents feature is disabled, the navigation item is correctly hidden from the UI, but users could still directly access `/agents` by entering the URL manually or using a bookmarked/shared link. This resulted in the Agents page initializing and displaying an error screen.

This change guards the `/agents` route using the existing `config.featureAgents()` feature flag. When the feature is disabled, the route now redirects to the dashboard instead of attempting to render the Agents page.

## Changes

- Added `Navigate` from `react-router`
- Guarded the `/agents` route using the existing feature flag
- Redirect to the dashboard when the feature is disabled
- No backend changes
- No new dependencies
- Minimal localized change

## Testing

- [x] Feature disabled → navigating directly to `/agents` redirects to `/dashboard`
- [x] Feature disabled → normal navigation remains unchanged
- [x] Feature enabled → `/agents` loads normally
- [x] Existing routing behavior verified

Fixes #8747

### DIFF
--- a/ui/ui-app/src/app/MainPageWithAuth.tsx
+++ b/ui/ui-app/src/app/MainPageWithAuth.tsx
@@ -1,7 +1,7 @@
 import { FunctionComponent } from "react";
 import { Page } from "@patternfly/react-core";
 import { AppHeader } from "@app/components";
-import { Route, Routes } from "react-router";
+import { Navigate, Route, Routes } from "react-router";
 import {
     AgentsPage,
     ArtifactPage,
@@ -71,7 +71,7 @@ export const MainPageWithAuth: FunctionComponent<MainPageWithAuthProps> = () => 
                         <Route path="/search" element={ <SearchPage /> } />
                         <Route path="/drafts" element={ <DraftsPage /> } />
                         <Route path="/explore" element={ <ExplorePage /> } />
-                        <Route path="/agents" element={ <AgentsPage /> } />
+                        <Route path="/agents" element={ config.featureAgents() ? <AgentsPage /> : <Navigate to={appNav.createLink("/dashboard")} replace /> } />
 
                         <Route
                             path="/explore/:groupId"


Adii-45:
Hi @EricWittmann @carlesarnal @paoloantinori — whenever you have a moment, could you please review this PR?

This is a small, localized fix that guards direct navigation to `/agents` when the feature is disabled, without changing the behavior when the feature is enabled!

Thanks!

paoloantinori:
Thanks for the ping, @Adii-45 — took a look. The fix is correct and nicely minimal: guarding the route with `config.featureAgents()` and using `<Navigate ... replace />` is the right call (the `replace` keeps the bad `/agents` URL out of history so a back-button doesn't bounce). A couple of things worth confirming:

- **Other feature-gated routes**: this addresses `/agents`, but if any other pages are conditionally hidden behind feature flags in the nav (e.g. via the same `config.feature*()` pattern), they'd have the same direct-navigation gap. Worth a quick scan so #8747 is closed completely rather than just for this one route.
- **Shared guard**: if more than one route needs this, a tiny `<FeatureRoute feature={...} fallback="/dashboard">` wrapper would avoid repeating the ternary per route. Not needed for a single route, just flagging it in case the list grows.

Neither is a blocker for this change as-is. Looks good to merge once a maintainer signs off.

Adii-45:
Thanks for taking the time to review this @paoloantinori!

I did a quick audit of the other feature-gated routes after your comment!

You're right that `/drafts`, `/roles`, and `/settings` follow a similar route-registration pattern, but they don't currently exhibit the same user-facing failure:

- `/drafts` remains functional because the relevant operation is backend-gated
- `/roles` is protected by the existing RBAC flow and correctly renders the access-denied state
- `/settings` is still directly reachable, although it doesn't result in an error

So the `/agents` route was the only one leading to a broken/error state, which is what this PR addresses.

I agree there may be value in introducing a shared route guard in the future if more feature-gated routes are added, but I kept this PR intentionally scoped to the verified bug reported in #8747!

Thanks again for the review!

paoloantinori:
Thanks for the follow-up audit, @Adii-45 — that's exactly the kind of check worth doing. Your reasoning tracks: the routes that fail silently or render a proper access-denied state are much less urgent than one that throws a broken/error screen, so scoping this PR to the `/agents` fix is the right call.

A shared route guard is a good future follow-up if more feature-gated routes start landing, but I wouldn't block this on it. Ship the targeted fix.

Adii-45:
Hey @carlesarnal @EricWittmann @jsenko,
All CI checks are passing now, please take a look whenever you get a chance, happy to change anything else if required!

carlesarnal (approved):
Clean, minimal, correct fix. Guards /agents route with existing config.featureAgents() flag. When disabled, redirects to /dashboard via Navigate with replace. Follows existing patterns (same as RootPageHeader.tsx). No unused imports, no sub-routes missed, no security issues. CI green, SonarCloud 0 new issues. Note: commit lacks Signed-off-by trailer but repo does not enforce DCO via CI.

PR #8880, opened 2026-07-24
fix(ui): guard page data loaders against stale responses on route param changes

### LINKED ISSUE #8879: UI: page data loaders lack request cancellation / stale response guarding  rapid navigation can render the wrong artifact's data
## Summary

The main entity pages in the UI (`ui-app`) load their data through a `createLoaders()` helper whose async requests are **not cancelled or stale-guarded** when the route params change. Because these pages stay mounted while only the URL params change (same `<Route>`, different `:groupId`/`:artifactId`/`:version`), navigating quickly between two resources can leave the page showing **the wrong resource's data** if an earlier in-flight response resolves after a later one.

This is a classic out-of-order-response race condition. It affects `ArtifactPage`, `VersionPage`, `GroupPage`, and `BranchPage`, which all share the same pattern.

## Affected code

Each page fires fetches like `service.getX(...).then(setX)` with no cancellation and no `useEffect` cleanup:

| Page | `createLoaders` | fetch → setState | effect (deps) |
|------|-----------------|------------------|----------------|
| `ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx` | L86 | L93 `getArtifactMetaData(...).then(setArtifact)` | L362 `[groupId, artifactId]` |
| `ui/ui-app/src/app/pages/version/VersionPage.tsx` | L108 | L115 `getArtifactMetaData(...).then(setArtifact)` | L403 `[groupId, artifactId, version]` |
| `ui/ui-app/src/app/pages/group/GroupPage.tsx` | L72 | L75 `getGroupMetaData(...).then(setGroup)` | L253 `[groupId]` |
| `ui/ui-app/src/app/pages/branch/BranchPage.tsx` | L40 | L47 `getArtifactMetaData(...).then(setArtifact)` | L119 `[groupId, artifactId, branchId]` |

Example (`ArtifactPage.tsx`):

```tsx
const createLoaders = (): Promise<any>[] => {
    // ...
    return [
        groups.getArtifactMetaData(gid, artifactId as string)
            .then(setArtifact)            // <-- no stale-response guard
            .catch(error => { setPageError(toPageError(error, "Error loading page data.")); }),
        groups.getArtifactRules(gid, artifactId as string)
            .then(setRules)
            .catch(error => { setPageError(toPageError(error, "Error loading page data.")); }),
    ];
};

useEffect(() => {
    setLoaders(createLoaders());   // <-- no cleanup returned; previous requests keep running
}, [groupId, artifactId]);
```

## Steps to reproduce

1. Open the UI and browse to an artifact, e.g. `/explore/default/artifact-A`.
2. Quickly navigate to a different artifact on the same route, e.g. `/explore/default/artifact-B` (both render `ArtifactPage`, so React re-renders it with new params instead of remounting).
3. Simulate `getArtifactMetaData` for `artifact-A` being slower than for `artifact-B` (e.g. throttle the network, or the backend is under load).

**Expected:** the page always ends up showing data for `artifact-B` (the current URL).

**Actual:** if `artifact-A`'s response resolves last, `setArtifact(A)` runs after `setArtifact(B)`, so the page displays `artifact-A`'s metadata/rules while the URL and breadcrumb say `artifact-B`. The same applies to `VersionPage` (metadata/content/version), `GroupPage`, and `BranchPage`.

## Root cause

- The requests started for the previous params are never cancelled (no `AbortController`).
- The `useEffect` that starts the loaders returns no cleanup function, so there is no "ignore stale result" flag either.
- React keeps the component mounted across param-only navigations, so the stale `.then(setState)` from the previous params still runs and wins if it resolves last.

## Proposed fix

Guard each loader effect so results from a superseded render are discarded. Minimal, dependency-free approach using an `ignore`/`AbortController` flag in the effect cleanup, applied consistently across the four pages:

```tsx
useEffect(() => {
    let cancelled = false;
    const guard = <T,>(setter: (v: T) => void) => (v: T) => { if (!cancelled) setter(v); };

    setLoaders([
        groups.getArtifactMetaData(gid, artifactId as string)
            .then(guard(setArtifact))
            .catch(error => { if (!cancelled) setPageError(toPageError(error, "Error loading page data.")); }),
        // ...
    ]);

    return () => { cancelled = true; };
}, [groupId, artifactId]);
```

(An `AbortController` passed into the Kiota-generated request options would additionally cancel the in-flight HTTP request, if the service layer is extended to accept one.) Because the four pages share this pattern, the guard could also be extracted into a small reusable hook/util.

## Impact

- Users can be shown metadata, rules, content, or branch info belonging to a different artifact/version than the one in the URL — misleading, and can lead to acting on the wrong resource.
- Also produces redundant "set state after the fact" work and unnecessary in-flight requests on rapid navigation.

## Environment

- `@apicurio/registry-ui-app` 3.3.1-Dev
- react 19.2.8, react-router 7.18.1

## Notes

Happy to work on this if the maintainers agree with the approach. This is a self-contained UI fix that can be applied uniformly across the four affected pages (and factored into a shared helper). Verified by inspecting the current `main` branch — I did not find an existing open issue tracking this.

Aditya-132:
## What

Fixes a stale-response race in the page components of the UI (`ui-app`). When navigating between two resources on the same route, a page could end up displaying data that belongs to a different resource than the one in the URL.

Closes #8879.

## Why

The page components load their data through a `createLoaders()` helper that fires `service.getX(...).then(setX)` with no request cancellation and no `useEffect` cleanup. react-router keeps these components mounted while only the route params change, so the requests started for the previous params are never cancelled. If an earlier, slower request resolves after a newer one, its `setX` overwrites the page state with the wrong resource's data. The unguarded callbacks can also run after the component has unmounted.

A concrete trigger exists today: the References tab links from one version to another (`ReferenceList.tsx`), both rendered by `VersionPage`, so same-route navigation is reachable and `getArtifactVersionContent` responses can be large enough to arrive out of order.

## Changes

- Add `newLoaderGuard()` in `src/utils/loader.utils.ts`. It returns a guard whose `wrap()` turns a callback into a no-op once `cancel()` has been called.
- In every page whose loaders depend on route params, the loader effect now creates a guard, passes it to `createLoaders(guard)`, wraps each callback with `guard.wrap(...)`, and calls `guard.cancel()` in the effect cleanup. Only results for the current params are applied.
- Add Vitest and a unit test for `newLoaderGuard`.

## Coverage

A grep of `setLoaders(createLoaders` across `src/app/pages` shows five pages whose loaders are param-driven, and all five are now guarded:

| Page | effect deps |
|------|-------------|
| `ArtifactPage` | `[groupId, artifactId]` |
| `VersionPage` | `[groupId, artifactId, version]` |
| `GroupPage` | `[groupId]` |
| `BranchPage` | `[groupId, artifactId, branchId]` |
| `EditorPage` | `[draftId, groupId, version]` |

The remaining pages that use `createLoaders()` (`AgentsPage`, `DashboardPage`, `ExplorePage`, `SearchPage`, `RolesPage`, `RulesPage`, `SettingsPage`, `GlobalContractRulesPage`) load once with `[]` deps and do not reload on navigation, so they do not carry this race. On `EditorPage`, only the state updates are guarded; the reauthentication side effect is left intact so an expired session is still handled.

## Follow-up

The guard silences the superseded callback rather than aborting the underlying request, which is a minimal correctness fix. An `AbortController`-based version that also cancels the in-flight HTTP request would be a natural follow-up if request volume becomes a concern.

## Testing

- `eslint` passes on the changed files (`npm run lint`: 0 errors).
- `tsc` reports no type errors from this change. The only `tsc` errors in a fresh checkout come from the pre-existing `@sdk/lib/generated-client/models` module resolution, which requires running the SDK code generation and is present on unmodified `main` as well.
- `npm test` runs the new Vitest suite for `newLoaderGuard` (4 tests passing): active guard invokes the callback, a cancelled guard does not, a callback wrapped after cancel is a no-op, and separate guards stay independent.

### DIFF
--- a/ui/ui-app/src/utils/loader.utils.ts
+++ b/ui/ui-app/src/utils/loader.utils.ts
@@ -1,0 +1,22 @@
+export type LoaderCallback<T> = (value: T) => void;
+
+export interface LoaderGuard {
+    wrap: <T>(callback: LoaderCallback<T>) => LoaderCallback<T>;
+    cancel: () => void;
+}
+
+export function newLoaderGuard(): LoaderGuard {
+    let cancelled = false;
+    return {
+        wrap: <T>(callback: LoaderCallback<T>): LoaderCallback<T> => {
+            return (value: T): void => {
+                if (!cancelled) {
+                    callback(value);
+                }
+            };
+        },
+        cancel: (): void => {
+            cancelled = true;
+        }
+    };
+}
--- a/ui/ui-app/package-lock.json
+++ b/ui/ui-app/package-lock.json
@@ -57,7 +57,8 @@
         "typescript": "5.9.3",
         "typescript-eslint": "8.65.0",
         "vite": "7.3.6",
-        "vite-tsconfig-paths": "6.1.1"
+        "vite-tsconfig-paths": "6.1.1",
+        "vitest": "^4.1.10"
       }
     },
     "../../typescript-sdk": {
@@ -71,7 +72,7 @@
         "eslint": "9.39.5",
         "rimraf": "6.1.3",
         "typescript": "5.9.3",
-        "typescript-eslint": "8.64.0",
+        "typescript-eslint": "8.65.0",
         "uuid": "14.0.1",
         "vite": "7.3.6",
         "vite-plugin-dts": "4.5.4",
@@ -1601,6 +1602,12 @@
         "win32"
       ]
     },
+    "node_modules/@standard-schema/spec": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@standard-schema/spec/-/spec-1.1.0.tgz",
+      "integrity": "sha512-l2aFy5jALhniG5HgqrD6jXLi/rUWrKvqN/qJx6yoJsgKhblVd+iqqU4RCXavm/jPityDo5TCvKMnpjKnOriy0w==",
+      "dev": true
+    },
     "node_modules/@std-uritemplate/std-uritemplate": {
       "version": "2.0.8",
       "resolved": "https://registry.npmjs.org/@std-uritemplate/std-uritemplate/-/std-uritemplate-2.0.8.tgz",
@@ -1707,9 +1714,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0 AND MIT",
       "optional": true,
       "os": [
@@ -1727,9 +1731,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "Apache-2.0 AND MIT",
       "optional": true,
       "os": [
@@ -1747,9 +1748,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0 AND MIT",
       "optional": true,
       "os": [
@@ -1767,9 +1765,6 @@
         "s390x"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0 AND MIT",
       "optional": true,
       "os": [
@@ -1787,9 +1782,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "Apache-2.0 AND MIT",
       "optional": true,
       "os": [
@@ -1807,9 +1799,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "Apache-2.0 AND MIT",
       "optional": true,
       "os": [
@@ -1887,6 +1876,16 @@
         "@swc/counter": "^0.1.3"
       }
     },
+    "node_modules/@types/chai": {
+      "version": "5.2.3",
+      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-5.2.3.tgz",
+      "integrity": "sha512-Mw558oeA9fFbv65/y4mHtXDs9bPnFMZAL/jxdPFUpOHHIXX91mcgEHbS5Lahr+pwZFR8A7GQleRWeI6cGFC2UA==",
+      "dev": true,
+      "dependencies": {
+        "@types/deep-eql": "*",
+        "assertion-error": "^2.0.1"
+      }
+    },
     "node_modules/@types/d3-color": {
       "version": "3.1.3",
       "resolved": "https://registry.npmjs.org/@types/d3-color/-/d3-color-3.1.3.tgz",
@@ -1935,6 +1934,12 @@
         "@types/d3-interpolate": "*",
         "@types/d3-selection": "*"
       }
+    },
+    "node_modules/@types/deep-eql": {
+      "version": "4.0.2",
+      "resolved": "https://registry.npmjs.org/@types/deep-eql/-/deep-eql-4.0.2.tgz",
+      "integrity": "sha512-c9h9dVVMigMPc4bwTvC5dxqtqJZwQPePsWjPlpSOnojbor6pGqdk541lfA7AqFQr5pB1BRdq0juY9db81BwyFw==",
+      "dev": true
     },
     "node_modules/@types/estree": {
       "version": "1.0.8",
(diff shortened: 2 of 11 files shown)


ui/ui-app/src/utils/loader.utils.ts
```diff
@@ -0,0 +1,22 @@
+export type LoaderCallback<T> = (value: T) => void;
+
+export interface LoaderGuard {
+    wrap: <T>(callback: LoaderCallback<T>) => LoaderCallback<T>;
+    cancel: () => void;
+}
+
+export function newLoaderGuard(): LoaderGuard {
```
paoloantinori, line 8: This is a pure, side-effect-free (apart from the closure flag) utility — a small unit test would be cheap and valuable. Something like: wrap a spy, invoke it → spy called; `cancel()`; invoke again → spy not called; also wrap a callback *after* `cancel()` and confirm it's a no-op. That locks in the contract and would catch regressions if this is later refactored toward `AbortController`.

paoloantinori (commented):
Solid fix — the `newLoaderGuard()` closure-flag pattern is an idiomatic way to neutralize stale async callbacks when route params change, and the four pages (`ArtifactPage`, `VersionPage`, `GroupPage`, `BranchPage`) are updated consistently. The mechanism is correct: each effect run mints its own guard, the cleanup flips `cancelled`, so any `then`/`catch` from the previous params becomes a no-op and can't clobber the new resource's state.

A few observations:

- **Completeness (worth confirming):** The PR targets the four main entity pages. It's worth checking whether any other pages in `ui-app` share the same `createLoaders()` + bare-`useEffect` shape (e.g., rules-only pages, search, or admin pages that load by id). If so, they'd still carry the stale-response bug described in #8879 and the fix would be partial. A quick grep for `setLoaders(createLoaders` across `src/app/pages` would confirm coverage.
- **Tests:** The new `loader.utils.ts` is a pure, isolated utility — an ideal candidate for a small unit test (see inline). This would lock in the contract and guard against regressions if the implementation is later swapped for `AbortController`.
- **Scope of cancellation (noting, not objecting):** As the PR body states, the guard silences the callback rather than aborting the underlying fetch. That's a reasonable minimal correctness fix; just flagging that superseded requests still hit the network. An `AbortController`-based version would be a natural follow-up if request volume becomes relevant.

Aditya-132:
Thanks for the thorough review, @paoloantinori. I pushed two commits addressing the points.

**Completeness.** I ran `grep -rn "setLoaders(createLoaders" src/app/pages`. Five pages have param-driven loaders: `ArtifactPage`, `VersionPage`, `GroupPage`, `BranchPage`, and `EditorPage` (`[draftId, groupId, version]`). The first four were in the original commit; `EditorPage` was the one still carrying the bug, so it is now guarded too (`fix(ui): guard EditorPage draft loaders against stale responses`). Its loaders are `async`/`await` with several setters, so the guard wraps each state update, and `requestReauthenticationIfUnauthorized` is left unguarded so an expired session is still handled. The remaining `createLoaders()` pages (`AgentsPage`, `DashboardPage`, `ExplorePage`, `SearchPage`, `RolesPage`, `RulesPage`, `SettingsPage`, `GlobalContractRulesPage`) load once with `[]` deps and do not reload on navigation, so they do not carry this race.

**Tests.** `ui-app` had no unit-test runner, so I added a minimal Vitest setup (a `test` script and a `node` test environment in the Vite config) plus a unit test for `newLoaderGuard` covering exactly the cases you described: wrap while active then the spy is called; `cancel()` then invoke and it is not called; wrap after `cancel()` is a no-op; and guard isolation. `npm test` gives 4 passing. Happy to split the Vitest bootstrap into its own PR if you would rather keep this one to the fix.

**Scope of cancellation.** Agreed, this silences the callback rather than aborting the fetch. I noted an `AbortController`-based follow-up in the PR description; that would need the service layer to thread a signal through the Kiota request options, which felt like a separate change.

paoloantinori:
Thanks @Aditya-132 — the `EditorPage` change closes the gap with the same pattern as the other four pages, and the Vitest suite covers the guard lifecycle well (active invocation, post-cancel no-op, wrap-after-cancel, and guard independence).

One non-blocking note on `EditorPage`: in both `loadDraft` and `loadDraftContent`, `requestReauthenticationIfUnauthorized(error)` runs unconditionally in the `catch`, so a 401 from a stale request can still trigger an auth redirect after the user has already navigated away. That's pre-existing behavior (not introduced here), so it doesn't block this PR — worth a follow-up if you want full stale-response coverage on the auth path.

Changes look good to me.

Aditya-132:
Thanks @paoloantinori, appreciate the quick reviews. Good point on the auth path. You're right that `requestReauthenticationIfUnauthorized(error)` runs unconditionally in both catch blocks, so a stale 401 can still trigger a redirect after the user has navigated away (pre-existing, as you noted). Driving reauth from the active request rather than a superseded one would be the more correct behavior for full stale-response coverage.

I'll keep this PR focused on the state-update race as reviewed, and handle the auth path as a follow-up (guarding the reauth call so only the active request can initiate the redirect). Happy to open that as a separate issue/PR and link it here.

paoloantinori:
Agreed on the scoping call @Aditya-132 — keeping this PR focused on the state-update race is the right call, and the stale-401 reauth path is genuinely pre-existing behavior that deserves its own treatment (driving reauth off the active request would touch the same `catch` blocks across more than just the page loaders). A follow-up issue for it would be great so it doesn't get lost; no need to expand the surface here.

The current changes look good on my re-read — the guard pattern is consistently applied across all five pages, the `EditorPage` async path now wraps every `setState`, and the Vitest suite covers the lifecycle cleanly. From my side this is good to go once the other maintainers weigh in.

carlesarnal (approved):
Clean, well-structured fix for the stale-response race condition in UI page loaders. All 3 commits have DCO sign-off and conventional commit messages. The LoaderGuard pattern is idiomatic, well-typed, and consistently applied across all 5 param-driven pages. 4 unit tests with specific assertions cover the guard lifecycle. No security concerns, no unrelated changes, follows existing code style. Maintainer paoloantinori already reviewed and expressed approval.

Aditya-132:
@paoloantinori why this checks are stuck

paoloantinori:
Not your code, @Aditya-132 — your changes look fine; the checks aren't actually "stuck running," they're skipped/waiting on the Apicurio PR-orchestrator side.

What I see on `a55e065` right now:
- `Security Analysis`, `License Compliance`, `Dependency Quality`, SonarCloud — all **green**.
- But the orchestrator jobs `Command`, `Review`, `Reconcile`, `Draft to Ready`, `Stale Detection`, and notably **`Test Result`** are all `skipped`, not `completed`.
- There's also one GitHub Actions check-suite on this SHA in `action_required` state (waiting on a maintainer action), which is the usual signature of a workflow that needs maintainer approval to run for a forked-branch PR.

So the full test suite that `/ready` promised hasn't actually started. There's nothing for you to push or change — this needs a maintainer to retrigger/re-approve the workflow run. I'll flag it so @carlesarnal / @EricWittmann can kick the orchestrator (likely a `/ready` again after the orchestrator resets, or a manual rerun of the `action_required` suite).

EricWittmann:
This looks good.  Thanks @Aditya-132 :+1:

PR #8893, opened 2026-07-25
feat(operator): TLS certificate expiration detection and alerting

### LINKED ISSUE #8082: Operator: TLS certificate expiration detection and alerting
## Parent Epic
Part of #8078 — Operator Auto Pilot capability level

## Phase
**Phase 2 (P1)** — High impact, medium effort

## Description

The operator should monitor TLS certificates referenced in the CR (keystores, truststores, ingress TLS secrets) and emit warnings before they expire.

## Requirements

- Periodically inspect TLS secrets referenced in `spec.app.tls` and `spec.app.ingress.tls`
- Parse certificate expiration dates from the secret data
- Emit Kubernetes Events and update status conditions when certificates are within a configurable warning window (default: 30 days)
- Configurable thresholds: `spec.app.tls.expirationWarningDays`
- Consider optional auto-rotation integration with cert-manager (detect cert-manager Certificate resources and verify they're configured)

## Acceptance Criteria

- [ ] Operator reads and parses TLS certificate expiration from referenced secrets
- [ ] Warning events emitted when certificates approach expiration
- [ ] Status condition `CertificateExpiring` added to CR status
- [ ] Configurable warning threshold
- [ ] Integration test covering certificate expiration detection
- [ ] Documentation

Priya-Sharma25:
### Descritpion

Implements TLS certificate expiration monitoring for certificates referenced by the `ApicurioRegistry3` custom resource. The operator periodically inspects referenced TLS secrets, emits warning events when certificates approach expiration, and updates the CR status with a `CertificateExpiring` condition.

### Implementation Details

* **Model**
  * Added `expirationWarningDays` (default: `30`) to `TLSSpec` to configure the certificate expiration warning window.

* **Status**
  * Introduced a new `CertificateExpiring` status condition.
  * Added `CertificateExpiringConditionManager` to aggregate expiring certificate warnings deterministically and avoid unnecessary status updates.

* **Controller**
  * Implemented `TlsExpirationChecker` to inspect PKCS12 (keystore/truststore) and PEM (Ingress TLS) certificates, including all aliases in PKCS12 stores.
  * Detects cert-manager-managed certificates using the `cert-manager.io/certificate-name` annotation and includes that information in warning messages.
  * Handles missing or malformed secrets gracefully by delegating validation failures to `ValidationErrorConditionManager` without interrupting reconciliation.
  * Emits Kubernetes `Warning` events with reason `CertificateExpiring`, emitting events only when transitioning into the expiring state to avoid repeated notifications.
  * Configured a 24-hour `maxReconciliationInterval` to periodically re-evaluate certificate expiration even when no CR changes occur.

* **Testing**
  * Added `testTlsExpirationAlerting` integration test using static PKCS12 test certificates to verify expiration detection, status conditions, and event emission without relying on external binaries during CI.

* **Documentation**
  * Documented the new configuration option and certificate expiration monitoring behavior in `assembly-operator-config-reference.adoc`.

Closes #8082


operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
Nachiket-Roy, line 1: Missing Apache 2.0 license header.

operator/controller/src/main/java/io/apicurio/registry/operator/status/CertificateExpiringConditionManager.java
```diff
@@ -0,0 +1,57 @@
+package io.apicurio.registry.operator.status;
```
Nachiket-Roy, line 17: Missing Apache 2.0 license header.

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
```diff
@@ -0,0 +1,226 @@
...
+
+            boolean alreadyAlerting = false;
+            if (primary.getStatus() != null && primary.getStatus().getConditions() != null) {
+                alreadyAlerting = primary.getStatus().getConditions().stream()
+                        .anyMatch(c -> TYPE_CERTIFICATE_EXPIRING.equals(c.getType()) && "True".equals(c.getStatus()));
+            }
+
+            if (!alreadyAlerting) {
+                emitEvent("CertificateExpiring", msg);
+            }
```
Nachiket-Roy, line 219: If a Keystore is expiring, the CertificateExpiring status becomes True and a warning event is emitted. If the Truststore or Ingress TLS certificate subsequently approaches expiration later, no event will be emitted for them because alreadyAlerting is globally evaluated to true for the resource.

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+        Secret passSecret = context.getClient().secrets().inNamespace(primary.getMetadata().getNamespace()).withName(passRef.getName()).get();
+        if (passSecret != null) {
+            String passKey = passRef.getKey() != null ? passRef.getKey() : defaultPassKey;
+            String b64Pass = passSecret.getData().get(passKey);
+            if (b64Pass != null) {
+                return new String(Base64.getDecoder().decode(b64Pass));
+            }
+        }
+        return "";
```
Nachiket-Roy: Silently returns "" if passSecret is null or not found

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
```diff
@@ -0,0 +1,226 @@
...
+                String alias = aliases.nextElement();
+                java.security.cert.Certificate cert = ks.getCertificate(alias);
+                if (cert instanceof X509Certificate x509) {
+                    evaluateCertificate(secret, x509);
+                }
+            }
+
+        } catch (Exception e) {
+            log.error("Failed to parse keystore secret {}", storeRef.getName(), e);
+            StatusManager.get(primary).getConditionManager(OperatorErrorConditionManager.class).recordException(e);
```
Nachiket-Roy: User errors (such as upload of JKS format instead of PKCS12, or an incorrect password) shouldn't report an internal operator reconciler failure (OperatorError)

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
```diff
@@ -0,0 +1,226 @@
...
+            CertificateFactory cf = CertificateFactory.getInstance("X.509");
+            java.security.cert.Certificate cert = cf.generateCertificate(new ByteArrayInputStream(certBytes));
+
+            if (cert instanceof X509Certificate x509) {
+                evaluateCertificate(secret, x509);
+            }
+
+        } catch (Exception e) {
+            log.error("Failed to parse ingress TLS secret {}", secretName, e);
+            StatusManager.get(primary).getConditionManager(OperatorErrorConditionManager.class).recordException(e);
```
Nachiket-Roy: see -- https://github.com/Apicurio/apicurio-registry/pull/8893/changes#r3649900805

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
Priya-Sharma25: Great catch on the multiple certificates scenario! I've updated the logic so it now checks if the specific certificate name is already in the existing condition message. This ensures that if a Keystore starts expiring, and then a Truststore starts expiring later, the operator will correctly emit a new warning event for the Truststore without being globally suppressed.

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
Priya-Sharma25: Added the missing Apache 2.0 license headers to both TlsExpirationChecker.java and CertificateExpiringConditionManager.java.

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
Priya-Sharma25: If a password secret is referenced but doesn't exist (or is missing the key), the method now correctly returns null and logs a ValidationError instead of silently returning an empty string.

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
Priya-Sharma25: Replaced the OperatorErrorConditionManager calls with ValidationErrorConditionManager in the exception blocks (for both Keystore and Ingress TLS parsing failures), so user configuration errors are appropriately reported as validation errors rather than internal reconciler failures.

Priya-Sharma25:
Hii  @paoloantinori  , @carlesarnal , 
Please Take  a Look !

carlesarnal (changes requested):
AI Review Round 1: Bug found in TlsExpirationChecker.java:211 - "True".equals(c.getStatus()) always returns false because getStatus() returns ConditionStatus enum not String. Events fire every reconciliation instead of only on first detection. Fix: use ConditionStatus.TRUE.equals(c.getStatus()). Also: import ordering issue in ApicurioRegistry3Reconciler.java:26 and FQCNs used inline in annotations and test code.

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
```diff
@@ -0,0 +1,250 @@
...
+                    .recordError("Failed to parse ingress TLS secret '%s': %s", secretName, e.getMessage());
+        }
+    }
+
+    private void evaluateCertificate(Secret secret, X509Certificate cert) {
+        Date notAfter = cert.getNotAfter();
+        Instant expiryDate = notAfter.toInstant();
+        Instant warningDate = Instant.now().plus(warningDays, ChronoUnit.DAYS);
+
+        if (expiryDate.isBefore(warningDate)) {
```
carlesarnal, line 197: Bug: `"True".equals(c.getStatus())` always returns `false` because `getStatus()` returns `ConditionStatus` enum, not String. Fix: use `ConditionStatus.TRUE.equals(c.getStatus())`

operator/controller/src/main/java/io/apicurio/registry/operator/feat/TlsExpirationChecker.java
```diff
@@ -0,0 +1,250 @@
...
+            if (isCertManager) {
+                msg += " Note: This certificate appears to be managed by cert-manager, auto-rotation may occur.";
+            }
+
+            expiringManager.recordExpiringCertificate(secret.getMetadata().getName(), msg);
+
+            boolean alreadyAlerted = false;
+            if (primary.getStatus() != null && primary.getStatus().getConditions() != null) {
+                alreadyAlerted = primary.getStatus().getConditions().stream()
+                        .filter(c -> TYPE_CERTIFICATE_EXPIRING.equals(c.getType()) && "True".equals(c.getStatus()))
```
paoloantinori: Same point @carlesarnal raised, but to underline the blast radius: because `c.getStatus()` returns `ConditionStatus` (an enum), `"True".equals(...)` is always `false`, so `alreadyAlerted` is always `false`. That means `emitEvent` fires on *every* reconciliation where the cert is in the warning window, not just on the transition into it. The `TlsITTest` assertion has the identical pattern and is affected the same way.

operator/controller/src/test/java/io/apicurio/registry/operator/it/TlsITTest.java
```diff
@@ -366,4 +366,57 @@ void testTLSInsecureTrafficEnabled() {
...
+                return false;
+            }
+            var condition = cr.getStatus().getConditions().stream()
+                    .filter(c -> "CertificateExpiring".equals(c.getType()))
+                    .findFirst();
+            return condition.isPresent() && "True".equals(condition.get().getStatus());
+        });
+        
+        // Assert event
+        var events = client.v1().events().inNamespace(namespace).list().getItems();
```
paoloantinori, line 422: This assertion won't pass as written: `condition.get().getStatus()` returns a `ConditionStatus` enum, so `"True".equals(...)` is always false and `await().until(...)` will time out instead of validating the condition. Compare against the enum directly (`== ConditionStatus.TRUE`) so the test actually exercises the alerting path it claims to.

operator/controller/src/main/java/io/apicurio/registry/operator/ApicurioRegistry3Reconciler.java
```diff
@@ -185,7 +186,10 @@
         }
 )
 @ControllerConfiguration(
-        name = "apicurioregistry3reconciler"
+        name = "apicurioregistry3reconciler",
+        maxReconciliationInterval = @io.javaoperatorsdk.operator.api.reconciler.MaxReconciliationInterval(
```
paoloantinori: Heads up that this `maxReconciliationInterval = 24h` applies to the entire reconciler for every `ApicurioRegistry3` CR, not just TLS checks. If the goal is periodic cert re-evaluation, a dedicated mechanism would avoid changing the global reconciliation cadence for all managed registries.

paoloantinori (commented):
Nice feature — TLS expiration monitoring is a real operational gap, and the structure (PKCS12 + PEM parsing, cert-manager detection, condition manager) is well thought out. Building on @carlesarnal's CHANGES_REQUESTED about the enum comparison, I want to flag that the same bug also breaks the integration test, and raise one broader concern about the reconciliation interval.

- **The IT assertion has the same enum bug as the production code.** `TlsITTest.testTlsExpirationAlerting` asserts `"True".equals(condition.get().getStatus())`, but `Condition.getStatus()` returns a `ConditionStatus` enum (see `CertificateExpiringConditionManager.updateCondition`, which calls `setStatus(ConditionStatus.TRUE)`). A `String.equals(enumValue)` comparison is always `false`, so the `await().until(...)` predicate can never succeed and the test will hit its timeout rather than validate the condition. This should be `condition.get().getStatus() == ConditionStatus.TRUE` (or `.name()`-based). Worth fixing in the same pass as the production-side dedup check @carlesarnal called out — otherwise the IT gives false confidence that the alerting path works.
- **`maxReconciliationInterval = 24h` is applied to the whole reconciler, not just TLS checks.** The annotation on `ApicurioRegistry3Reconciler` changes the reconciliation cadence for *every* `ApicurioRegistry3` CR the operator manages, not just TLS operations. That's a meaningful behavior change for the operator's general responsiveness. If the intent is just to re-check certificates periodically, a dedicated scheduled check (or a narrower mechanism) would avoid widening the global interval for everyone. Worth confirming this tradeoff is intended.
- **Event spam once the enum bug is fixed is still worth a second look.** `emitEvent` creates a new `Event` with a random name each time the dedup check fails to short-circuit. Once the comparison is corrected to compare against the enum, confirm that `alreadyAlerted` flips correctly so Events are emitted only on the not-expiring → expiring transition, not on every reconciliation where the cert is still expiring.

Diff was read in full (250-line `TlsExpirationChecker`, 73-line condition manager, and the `TlsITTest` additions).

paoloantinori:
Thanks for the ping @Priya-Sharma25 — took a look and posted a review above. The main thing is the `ConditionStatus` enum comparison bug @carlesarnal flagged also affects the new `TlsITTest` assertion (same `"True".equals(...)` pattern), so both production and test need the fix together.

Priya-Sharma25:
Thanks for the reviews @carlesarnal and @paoloantinori! I have addressed all of the points raised:

**1. Enum Comparison Bug (Prod & Test):** Replaced the string comparison "True".equals(c.getStatus()) with ConditionStatus.TRUE.equals(c.getStatus()) in TlsExpirationChecker.java. Also updated the TlsITTest.java assertion to use ConditionStatus.TRUE == condition.get().getStatus() so it successfully validates the alerting path in the integration test.
**2. Import Ordering & FQCNs:** Fixed the import ordering in ApicurioRegistry3Reconciler.java and removed the inline Fully Qualified Class Names from the annotation.
**3. Reconciliation Interval:** Great point on the global interval tradeoff! I removed the maxReconciliationInterval = 24h from the @ControllerConfiguration annotation. Instead, the reconciler now checks if TLS is configured, and if so, it appends .rescheduleAfter(24, TimeUnit.HOURS) to the UpdateControl. This correctly implements a narrower mechanism for periodic checking without changing the global interval for all registries.

Everything is passing locally on my end. Let me know if everything looks good now!

paoloantinori:
Thanks @Priya-Sharma25 — both fixes look correct.

- The `ConditionStatus.TRUE.equals(c.getStatus())` swap in `TlsExpirationChecker.evaluateCertificate` resolves the always-false comparison, and the dedup-by-`secretName` now actually fires.
- The `TlsITTest` assertion using `ConditionStatus.TRUE == condition.get().getStatus()` validates the alerting path end-to-end (`==` is fine on enum singletons; `.equals()` would be marginally more idiomatic but not worth a churn).

No further concerns from my side.

carlesarnal:
### Review Round 2

All three critical issues from the previous review are resolved:

1. **Enum comparison bug** — `ConditionStatus.TRUE.equals(c.getStatus())` in `TlsExpirationChecker.java:211` now correctly compares against the enum instead of the string `"True"`. The dedup check works as intended.
2. **Test assertion fixed** — `TlsITTest.java` uses `ConditionStatus.TRUE == condition.get().getStatus()`, properly validating the alerting path.
3. **Reconciliation interval scoped** — `maxReconciliationInterval` removed from the annotation; `rescheduleAfter(24, HOURS)` is applied only when TLS is configured. Good approach.

**Minor style nits** (non-blocking, happy to leave to maintainer discretion):

- `ApicurioRegistry3Reconciler.java:219` — `java.util.concurrent.TimeUnit.HOURS` is used as an inline FQCN. Adding `import java.util.concurrent.TimeUnit;` and using `TimeUnit.HOURS` would be more consistent with the rest of the codebase.
- `TlsExpirationChecker.java:133,177` — `java.security.cert.Certificate` used inline twice. Since `X509Certificate` and `CertificateFactory` are already imported from the same package, adding `import java.security.cert.Certificate;` would clean this up.
- `TlsITTest.java` — `java.io.InputStream`, `java.util.Base64`, and `io.fabric8.kubernetes.api.model.SecretBuilder` are used as inline FQCNs in the test method instead of being imported at the top of the file.
- `expired.p12` test resource is added but not referenced by any test — keeping it for future test coverage is fine, just wanted to flag it.

Overall the implementation is solid — PKCS12 + PEM parsing, cert-manager detection, deterministic condition aggregation, and the integration test all look good.

Priya-Sharma25:
Thanks @carlesarnal for the thorough Round 2 review! Really glad to hear the three critical fixes are looking good. 

I've addressed all four style nits in the latest push:

1. ApicurioRegistry3Reconciler.java — Added import java.util.concurrent.TimeUnit and replaced the inline FQCN with TimeUnit.HOURS.
2. TlsExpirationChecker.java — Added import java.security.cert.Certificate and replaced both inline usages.
3. TlsITTest.java — Added proper imports for InputStream, Base64, Secret, and SecretBuilder, replacing all inline FQCNs.
4. expired.p12 — Noted! Keeping it for planned follow-up test coverage of the expired-certificate detection path.

Checkstyle passes cleanly. Let me know if there's anything else!

paoloantinori (commented):
Follow-up after the Round 2 fixes landed — the critical bugs (enum comparison, test assertion, event dedup guard) are all resolved. One remaining gap in the reconciler wiring that I think is worth addressing before merge, plus two minor notes.

The substantive point: the 24-hour periodic reschedule is gated on `app.tls != null`, but `TlsExpirationChecker` also inspects Ingress TLS secrets (`app.ingress.tlsSecretName`) independently of the app-level TLS spec. In an Ingress-only TLS setup — which is the common production config where cert-manager terminates TLS at the Ingress and the app itself has no keystore — `app.tls` is null, so the `rescheduleAfter` branch is skipped and Ingress cert expiration is never periodically re-evaluated. It would only be checked when some *other* change triggers reconciliation, which defeats the periodic-monitoring intent for exactly the deployment shape that needs it most.

Priya-Sharma25:
Great catch @paoloantinori! You're absolutely right — the Ingress-only TLS setup (cert-manager terminating at the Ingress with no app-level keystore) is the most common production config, and it was being silently excluded from periodic monitoring.

Fixed by broadening the rescheduleAfter condition to also check for app.ingress.tlsSecretName:

boolean hasTls = primary.getSpec().getApp().getTls() != null
        || (primary.getSpec().getApp().getIngress() != null
                && primary.getSpec().getApp().getIngress().getTlsSecretName() != null);

This ensures periodic re-evaluation for both app-level TLS and Ingress-only TLS configurations. Checkstyle passes cleanly.

paoloantinori:
Thanks @Priya-Sharma25 — broadening the `hasTls` check to also cover `app.ingress.tlsSecretName` is exactly the right fix. Ingress-only TLS (cert-manager terminating at the Ingress with no app-level keystore) is the most common production setup, so it should be covered by the periodic reschedule, not silently excluded.

The condition reads correctly and the rest of the reconciler wiring now looks consistent. From my side the remaining items from the earlier rounds are all addressed.

Priya-Sharma25:
Hi @EricWittmann , I've updated this PR based on the review feedback . I believe all the requested changes have been addressed. Whenever you have a chance, I'd appreciate it if you could take another look. Thank you!

Priya-Sharma25:
Hi @jsenko  , could you please take another look at this PR when you have a chance? I believe all the requested changes have been addressed and it should be ready to merge. Thank you!

Priya-Sharma25:
Hi!  @EricWittmann  , @jsenko 
This PR is ready for review and merge. I’ve addressed the required changes and there are no pending updates from my side.
Please take a look when you get a chance. Thanks!

carlesarnal (approved):

PR #8894, opened 2026-07-25
feat(cli): include 'default' group in listings with implicit marker (#8643)

### LINKED ISSUE #8643: CLI: Decide whether to include 'default' group in listings
## Parent Epic

## Description
`acr group` currently hides the implicit `default` group from listings. Decide whether it should appear and how to present it.

## Considerations
- The default group is implicit — the server never stores a row for it
- Users may expect to see it listed since they can add artifacts to it
- If included, it could be marked with a visual indicator (e.g. `(implicit)`)
- A `--include-default` flag could make it opt-in

## Acceptance Criteria
- Decision documented on whether to show the default group
- Implementation matches the decision
- Tests updated accordingly

Anujkumar9081:
…#8643)Resolves #8643

**Description**
This PR updates the CLI `group list` command (`acr group list`) to explicitly include the implicit `default` group in the output. Because the default group is not stored server-side, it was previously omitted from listings. This created confusion for users since they can create artifacts in the default group without specifying one. 

**Changes**
* **Injected Default Group:** The `GroupListCommand` (`SearchGroupsCommand`) now injects a pseudo-record for the default group into the `GroupSearchResults` when fetching the first page (and when not explicitly filtering it out).
* **Implicit Marker:** Added the `(implicit)` marker to the group ID so it displays as `default (implicit)` in the output. This ensures the group is discoverable while clarifying that it isn't a traditional, server-stored group.
* **Updated Tests:** Updated the assertions in `GroupCommandTest.java` to account for the presence of the new `default (implicit)` group in the CLI output.


ui/ui-editors/package-lock.json
Nachiket-Roy, line 1: Not related to the scope of the PR.

log.txt
Nachiket-Roy, line 1: Looks like some sort of debugging that should not belong to the repo

patch.txt
Nachiket-Roy, line 1: same here

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -72,6 +72,21 @@ public void run(final OutputBuffer output) throws Exception {
             //noinspection ConstantConditions
             applyFilters(r.queryParameters);
         }));
+
+        if (pagination.getPage() == 1 && (groupId == null || "default".equals(groupId)) && description == null && (labels == null || labels.isEmpty())) {
```
Nachiket-Roy: The total result count is only incremented if pagination.getPage() == 1 If a user navigates to page 2 (where pagination.getPage() == 2), the count will not be incremented.

ui/ui-app/package-lock.json
Nachiket-Roy, line 1: This shouldn't be part of this PR.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -62,16 +61,35 @@ public class SearchGroupsCommand extends AbstractCommand {
...
+            }
+            // Only inject the visual row on the first page.
+            if (pagination.getPage() == 1) {
+                var groups = results.getGroups();
+                if (groups == null) {
+                    groups = new java.util.ArrayList<>();
+                    results.setGroups(groups);
+                }
+                var defaultGroup = new io.apicurio.registry.rest.v3.beans.GroupMetaData();
+                defaultGroup.setGroupId("default (implicit)");
```
paoloantinori: Putting `(implicit)` into the groupId itself means JSON output consumers get a groupId that doesn't match any server-side value — `default` is the real sentinel used across the REST API. For table output the marker is useful; for `--output-type json` it's a data-field mutation that will break scripted lookups. Consider keeping groupId as `"default"` and moving the marker to a display-only field (or applying it only in the table formatter).

paoloantinori (commented):
Reasonable UX improvement — surfacing the implicit `default` group in `acr group list` addresses a real discoverability gap (users can create artifacts in it but never see it listed). The implementation is small and the tests are updated consistently. One design concern worth considering before merge.

- **Embedding the `(implicit)` marker inside `groupId` breaks JSON consumers.** The injected `GroupMetaData` uses `setGroupId("default (implicit)")`, and the test asserts that exact string. But `default` is the actual sentinel value the registry uses for the default group across the REST API and CLI (e.g. `acr artifact list --group default`). A user running `acr group list --output-type json | jq '.groups[].groupId'` and then piping `default (implicit)` into a subsequent command will get a not-found, because the server doesn't know that groupId. For human-facing table output the marker is helpful; for JSON it's a data-field mutation. Worth considering either (a) keeping the groupId as the real `"default"` and moving the `(implicit)` marker into a separate display field (e.g. `description` or a synthetic flag), or (b) only applying the marker in the table formatter and emitting the canonical `"default"` in JSON output. The current approach works for the interactive case but will surprise script authors.
- **Minor: the injected `GroupMetaData` has all other fields null** (no `createdOn`, `modifiedBy`, `labels`, etc.). The table printer and any JSON consumer iterating fields should tolerate this, but it's worth a quick check that the output formatter doesn't NPE on a mostly-empty record — the existing tests pass, so this is likely fine, just flagging it.

Note on @Nachiket-Roy's pagination comment: reading the diff, the `results.setCount(... + 1)` call sits *outside* the `if (pagination.getPage() == 1)` block (only the visual-row injection is gated by page==1), so the count is bumped on every page — the pagination-consistency intent looks correct as written.

Diff read in full (2 files, 40/16).

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -72,6 +72,28 @@ public void run(final OutputBuffer output) throws Exception {
             //noinspection ConstantConditions
             applyFilters(r.queryParameters);
         }));
+
+        final boolean noFiltersExcludeDefault = (groupId == null || "default".equals(groupId))
+                && description == null && (labels == null || labels.isEmpty());
+
+        if (noFiltersExcludeDefault) {
+            // Always bump the total count so pagination stays consistent across pages.
+            if (results.getCount() != null) {
```
paoloantinori: Worth double-checking the semantics of `GroupSearchResults.count`. This `+1` runs on **every** page (the guard is outside the `page == 1` check), but the row is only injected on page 1. If `count` is the page-local item count rather than the total, pages > 1 will report `count == realRows + 1` while returning only `realRows` groups. If `count` is the total, this is correct as written — the comment should just state which one it is.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -72,6 +72,28 @@ public void run(final OutputBuffer output) throws Exception {
...
+            }
+            // Only inject the visual row on the first page.
+            if (pagination.getPage() == 1) {
+                var groups = results.getGroups();
+                if (groups == null) {
+                    groups = new java.util.ArrayList<>();
+                    results.setGroups(groups);
+                }
+                var defaultGroup = new io.apicurio.registry.rest.v3.beans.GroupMetaData();
+                defaultGroup.setGroupId("default (implicit)");
```
paoloantinori: Prepending here can make page 1 hold `pageSize + 1` rows when the server already returned a full page of real groups (the injection happens after the fetch). Bounded by +1 since the default is a singleton, but worth either documenting the overflow or only injecting when the current page is non-full. Also note this pins `default (implicit)` above the server's sort order — fine if intentional.

paoloantinori (commented):
Re-read at the new head (`6d5d730`). The pagination approach is sound in spirit — bump the count on every page, inject the visual row only on page 1 — but two consistency points are worth confirming before merge.

- **`count` semantics (line ~81).** The code unconditionally does `results.setCount(results.getCount() + 1)` whenever filters are absent, but only injects the row on `page == 1`. That is correct *only if* `count` represents the **total** number of matches (as the inline comment assumes). If `GroupSearchResults.count` is the page-local count of items in *this* response (a common convention in Apicurio search beans), then on pages > 1 the count would be bumped to N+1 while only N rows are actually present, breaking the invariant `count == groups.size()`. Worth checking the bean definition / OpenAPI schema to confirm which one it is.
- **Page-size overflow on page 1 (line ~92).** The default row is prepended *after* the server already returned up to `pageSize` real groups. If the registry has `>= pageSize` real groups, page 1 now carries `pageSize + 1` rows, which can surprise consumers (and the test's own pagination logic) that assume a page is bounded by the requested page size. Since the default group is a singleton the overflow is at most 1, but consider documenting it or injecting only when the page is non-full.
- **Sort order.** `groups.add(0, defaultGroup)` pins `default (implicit)` to the top regardless of the server's sort order. If listings are sorted alphabetically or by `createdOn`, a user looking for the `d` block mid-list won't find it there. Intentional UX choice is fine — just flagging it so it's a conscious decision rather than an accident.
- Minor: the guard is named `noFiltersExcludeDefault` but the branch *includes* the default group; a name like `shouldInjectDefault` would read less ambiguously for future maintainers.

Tests look like they were updated consistently. No blocker from my side — the above are confirmations rather than must-fixes.

Anujkumar9081:
Thanks for the thorough review! Addressed all points:

- **JSON output:** groupId is now the canonical `"default"` — the `(implicit)` marker is only shown in table output. JSON consumers won't be affected.
- **Count:** Confirmed via the OpenAPI schema that `count` is the total across all pages ("may be more than the number of groups in the result set"), so the +1 is correct. Added a comment documenting this.
- **Page-size overflow:** Added a comment noting it's bounded by +1 since the default group is a singleton.
- **Sort order:** Intentionally pinned to the top — the default group is special and should always be the first thing users see.
- **Variable naming:** Renamed to `shouldInjectDefault`.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -65,13 +66,42 @@ public class SearchGroupsCommand extends AbstractCommand {
...
+        if (shouldInjectDefault) {
+            // count is the total number of matching groups across all pages (per OpenAPI schema),
+            // so we always bump it by 1 for the implicit default group.
+            if (results.getCount() != null) {
+                results.setCount(results.getCount() + 1);
+            }
+            // Only inject the visual row on the first page.
+            // This may produce pageSize + 1 rows on page 1, bounded by +1 since the default
+            // group is a singleton.
+            if (pagination.getPage() == 1) {
```
paoloantinori: Forcing `default` to index 0 ignores the user's `--order`/`--orderby`: `acr group list --order desc` still shows `default` first. The injected record has null `createdOn`/`modifiedOn`/`owner`, so it would sort to a stable, well-defined position if passed through the same comparator the server applies. Consider sorting the combined list after injection (or pinning deliberately and documenting it) so the list output stays consistent with the sort flags.

paoloantinori (commented):
Re-read the new head `a6d0c6e`. Splitting the canonical JSON groupId (`"default"`) from the visual table marker (`default (implicit)`) is the right call and resolves the JSON-consumer concern cleanly. One new substantive point:

- **The injected default group ignores the requested sort order.** On page 1 it's always prepended at index 0 via `groups.add(0, defaultGroup)` regardless of `ordering.getOrder()`/`getOrderBy()`. So `acr group list --order desc` will still surface `default` first, breaking the contract that the list respects the user's `--order`/`--orderby` flags. The pseudo-record also has no `createdOn`/`modifiedOn`/`owner`, so if it were placed through the same comparator the server uses it would sort to a well-defined position (likely last under `desc`-by-created, given null fields). Easiest fix: after injecting, sort the combined list with the same comparator the server would apply; or document explicitly that the default group is always pinned to the top. Worth deciding which behavior you want before merge — the current code is a silent sort inconsistency rather than an intentional UX choice.

Otherwise the count-bumping + page-1-only-injection pagination logic reads consistently with the previous review. No blockers from me; the sort point is the one thing I'd want resolved (or explicitly documented) before merge.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
Anujkumar9081: Good catch — the pinning is intentional. The default group is a special 
implicit group that always exists but is never returned by the server. 
Pinning it at index 0 ensures it's immediately visible for discoverability, 
which is the core goal of this issue (#8643).

Added an explicit code comment documenting this as a deliberate UX decision 
rather than a silent inconsistency.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -65,13 +66,46 @@ public class SearchGroupsCommand extends AbstractCommand {
...
+                var groups = results.getGroups();
+                if (groups == null) {
+                    groups = new java.util.ArrayList<>();
+                    results.setGroups(groups);
+                }
+                var defaultGroup = SearchedGroup.builder()
+                        .groupId(DEFAULT_GROUP_ID)
+                        .description("(implicit)")
```
paoloantinori: Setting `description("(implicit)")` here means JSON consumers see a placeholder value in a metadata field, and the table view shows "(implicit)" in both the Group ID column (via `displayGroupId`) and the Description column. If the marker is table-only, consider leaving `description` null and relying on the `displayGroupId` logic in `printResults`.

cli/src/main/java/io/apicurio/registry/cli/search/SearchGroupsCommand.java
```diff
@@ -102,8 +136,11 @@ private void printResults(final OutputBuffer output, final GroupSearchResults re
                     final var table = new TableBuilder();
                     table.addColumns(GROUP_ID, DESCRIPTION, CREATED_ON, OWNER, MODIFIED_ON, MODIFIED_BY, LABELS);
                     Optional.ofNullable(results.getGroups()).orElse(List.of()).forEach(g -> {
+                        final String displayGroupId = DEFAULT_GROUP_ID.equals(g.getGroupId())
+                                ? g.getGroupId() + " (implicit)"
```
paoloantinori, line 118: This `displayGroupId` already appends " (implicit)" for the table. Combined with `description("(implicit)")` on the bean (line 99), the default-group row shows the marker twice. Keeping the visual marker here and removing it from `description` would avoid the duplication.

paoloantinori (commented):
Re-read at the new head (`98db9aea`). The split between canonical JSON `groupId` (`"default"`) and the table-only `displayGroupId` marker is the right direction, but the synthesized `SearchedGroup` puts `"(implicit)"` into the `description` field, which creates two issues worth confirming before merge:

- **Redundant marker in table view.** `printResults` already renders `displayGroupId` as `"default (implicit)"` via the `DEFAULT_GROUP_ID.equals(...)` branch. Since the bean also carries `description("(implicit)")`, the Description column shows `"(implicit)"` a second time in the same row.
- **Misleading `description` in JSON output.** The PR description states the marker should be "only shown in table output," but `--output-type json` consumers see `description: "(implicit)"` for the default group — a UI hint leaked into what is otherwise a metadata field. Scripts reading `description` would get a meaningless placeholder instead of null/empty.

If the intent is table-only display, leaving `description` null on the synthesized bean and deriving the marker purely in `printResults` (which already does so) would keep the JSON clean and eliminate the double marker. Minor point, but it affects any automation parsing JSON group listings.

Anujkumar9081:
removed .description("(implicit)") from the bean. The marker now lives exclusively in the table formatter via displayGroupId, keeping JSON output clean.

carlesarnal:
Reviewed the latest revision (812cf32). The code is in good shape — all of paoloantinori's previous feedback has been addressed cleanly: JSON output uses the canonical `"default"` groupId, the `(implicit)` marker lives only in the table formatter, the sort-pinning is explicitly documented, and the `description("(implicit)")` has been removed from the bean.

Two minor nits:

1. **Fully qualified `ArrayList` (SearchGroupsCommand.java, ~line 93)**
   ```java
   groups = new java.util.ArrayList<>();
   ```
   The file already imports `java.util.List` and `java.util.Optional`. Consider adding `import java.util.ArrayList;` and using `new ArrayList<>()` to stay consistent with the existing import style.

2. **Stale line in commit message body** — the commit body still says *"Set description to '(implicit)' so JSON consumers can identify it"*, but this was removed in the final code. Not a blocker since the code itself is correct, just a small discrepancy.

Everything else looks good: DCO sign-off is present, conventional commit format, tests updated consistently (size checks + groupId assertions across empty/list/delete scenarios), injection logic is well-guarded by `shouldInjectDefault`, and pagination semantics are properly documented. SonarQube passed clean.

No blockers from my side — the above are style suggestions, not must-fixes.

Anujkumar9081:
Thanks for the final review! Both nits are fixed: added the java.util.ArrayList import and removed the stale line from the commit message.

Anujkumar9081:
Rebased on latest main to fix the UI build conflict. Single clean commit, only 2 CLI files changed. CI should pass now.

vandanayadav7 (changes requested):
Ran e2e testing in a headless Linux container. Found issues on top of Paolo's existing feedback:

**1. Table output crashes with NPE**

`acr search group` in table mode throws:
```
Cannot invoke "java.util.Map.entrySet()" because "labels" is null
```
The injected default group has `labels = null`, and the table rendering calls `.entrySet()` without a null check.

**2. Default group missing from `acr group`**

The injection is only in `SearchGroupsCommand` (`acr search group`), but `GroupCommand` (`acr group`) is unchanged. Both commands need it — `acr group` is what most users run to list groups, and `acr search group` should also include it for consistency. Right now neither works correctly: `acr search group` crashes (point 1) and `acr group` doesn't show the default group at all:

```
$ acr group --output-type json
{"groups": [], "count": 0}
```

The test changes in `GroupCommandTest` expect the default group in `acr group` output, but the code never produces it — these assertions would fail against a live registry.

**3. Doesn't follow existing patterns**

`GroupGetCommand` already handles the default group using `IdUtil.isDefaultGroup()` and a synthetic stub with "N/A" for null fields. This PR creates a separate `DEFAULT_GROUP_ID` constant instead of using the existing utility. The fix should follow the same pattern for consistency.

Paolo's earlier feedback on pagination, sort order, and the redundant marker also looks unresolved — worth addressing together.

Anujkumar9081:
Thanks for the thorough e2e testing @vandanayadav7! I've force-pushed a new revision addressing all three issues you found:

1. **NPE fixed**: Updated `Conversions.convertToString()` to gracefully handle `null` label maps instead of calling `.entrySet()` on them.
2. **`acr group` consistency**: Refactored the injection logic into a shared utility method and wired it up in `GroupCommand` so that `acr group` now correctly injects the default group just like `acr search group`.
3. **Followed existing patterns**: Reused the existing `IdUtil` pattern by making its `DEFAULT_GROUP` constant public, and encapsulated the injection logic into a new `IdUtil.injectDefaultGroup()` utility method rather than duplicating code or creating new standalone constants.

*(Note regarding Paolo's earlier feedback on pagination, sort order, etc.: I addressed all of those points last week. **@carlesarnal has already reviewed those specific fixes and confirmed they are resolved cleanly.** I suspect the headless container you ran the e2e tests in might have just been built from an older version of this branch. Rest assured, all of those fixes are still fully intact in this latest push!)*

Let me know if this looks good to go!

paoloantinori:
Re-read at `f30fe18`. The `Conversions.convertToString(Map)` null guard (line 210) is in place, so the table NPE @vandanayadav7 caught is fixed, and the `acr group`/`acr search group` consistency holds. Combined with Carles' review, nothing else from me on this head — looks good to merge once CI is green.

vandanayadav7 (dismissed):
All three findings addressed — NPE fixed, both commands now inject the default group, and the logic uses shared IdUtil utilities. Verified with e2e testing. LGTM.

vandanayadav7:
@Anujkumar9081 The CLI Verification failure is caused by this PR's changes — `testSearchGroupsPagination` expects `limit=1` to return at most 1 result, but now returns 2 because the `default` group is included in listings. The test needs to be updated to account for it.

Anujkumar9081:
Fixed the `testSearchGroupsPagination` assertion — updated it to account for the injected default group on page 1 (limit=1 now returns up to 2 results: 1 from server + the implicit default)

Anujkumar9081:
@vandanayadav7 The "Build Java Application" failure is unrelated to my CLI changes. The error is a missing package `io.apicurio.registry.a2a.rest.beans` — `WellKnownResource.java` imports `AgentCard`, `AgentSearchRequest`, and `AgentSearchResults` from that package, but those classes only exist under `io.apicurio.registry.rest.v3.beans` (auto-generated from OpenAPI). This appears to be a pre-existing issue on `main` introduced by the recent A2A/MCP commits. Could you re-run CI or check if this is a known issue? My PR only modifies files under `cli/`.

vandanayadav7 (approved):

PR #8896, opened 2026-07-25
feat(ui): add version tagging toolbar to prompt template playground

### LINKED ISSUE #8425: feat: prompt template playground in the registry UI
## Summary

Build an interactive Prompt Template Playground into the Apicurio Registry UI for editing, testing, and comparing prompt template versions.

## Scope

- Interactive prompt editor with syntax highlighting and live variable extraction
- Test panel: fill in variable values, render the prompt in real-time
- Side-by-side version diff for prompt templates
- Version tagging via existing labels system (production, staging, experimental)
- End-to-end tests for the playground UI

## Context

This is proposed as an [LFX Mentorship](https://mentorship.lfx.linuxfoundation.org/) project for 2026 Term 3 (Sep-Nov). Builds on existing `PROMPT_TEMPLATE` artifact type and prompt rendering service.

shreyashp25:
Addresses point 3 from #8425.

The proposal suggested reusing the existing label components (already used in the artifact metadata forms) instead of building a new tagging UI from scratch for the Playground - so that's what I did here.

Added a small toolbar to the top of the Playground panel with a "Tag this version" button. Clicking it opens the same LabelsFormGroup component that's already used in EditMetaDataModal, so the tagging experience feels consistent with the rest of the registry instead of introducing a new pattern.

Current tags show up as compact labels right in the toolbar.

This is a fairly small change since it's mostly reusing what's already there rather than adding new components.

Closes #8425

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -13,14 +13,22 @@ import {
     FormGroup,
     FormSelect,
     FormSelectOption,
+    Grid,
+    Label,
+    LabelGroup,
     Spinner,
     TextArea,
     TextInput,
-    Title
+    Title,
+    Toolbar,
+    ToolbarContent,
+    ToolbarItem
 } from "@patternfly/react-core";
+import { TagIcon } from "@patternfly/react-icons";
 import { PromptVariable } from "./PromptTemplateViewer";
 import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
 import { RenderPromptResponse, RenderPromptValidationError } from "@models/RenderPromptResponse.ts";
+import { ArtifactLabel, LabelsFormGroup } from "@app/components/modals/LabelsFormGroup";
 
 export type PromptTemplateTestPanelProps = {
     groupId: string;
@@ -70,6 +78,12 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
     const [validationErrors, setValidationErrors] = useState<RenderPromptValidationError[]>([]);
     const [isLoading, setIsLoading] = useState(false);
     const [error, setError] = useState<string>("");
+
+    // Version tagging state - reuses the same ArtifactLabel model and
+    // LabelsFormGroup editor used elsewhere in the app (e.g. EditMetaDataModal)
+    // so tagging UX is consistent across the registry.
+    const [labels, setLabels] = useState<ArtifactLabel[]>([]);
+    const [isEditingLabels, setIsEditingLabels] = useState(false);
 
     const setValue = (name: string, value: any): void => {
         setValues(prev => ({ ...prev, [name]: value }));
@@ -173,6 +187,34 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                 <CardTitle>
                     <Title headingLevel="h3" size="md">Test Prompt</Title>
                 </CardTitle>
+                <Toolbar className="playground-toolbar" inset={{ default: "insetNone" }}>
+                    <ToolbarContent>
+                        <ToolbarItem>
+                            <LabelGroup>
+                                {labels.filter(l => l.name).map((label, idx) => (
+                                    <Label key={idx} color="purple" isCompact>
+                                        {label.name}{label.value ? `=${label.value}` : ""}
+                                    </Label>
+                                ))}
+                            </LabelGroup>
+                        </ToolbarItem>
+                        <ToolbarItem align={{ default: "alignRight" }}>
+                            <Button
+                                variant="link"
+                                icon={<TagIcon />}
+                                data-testid="playground-tag-version-button"
+                                onClick={() => setIsEditingLabels(!isEditingLabels)}
+                            >
+                                Tag this version
+                            </Button>
+                        </ToolbarItem>
+                    </ToolbarContent>
+                </Toolbar>
+                {isEditingLabels && (
+                    <Grid hasGutter className="playground-labels-editor">
+                        <LabelsFormGroup labels={labels} onChange={setLabels} />
+                    </Grid>
+                )}
             </CardHeader>
             <CardBody>
                 <Form className="test-panel-form">
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, useEffect, useState } from "react";
+import { FunctionComponent, useEffect, useRef, useState } from "react";
 import "./PromptTemplateTestPanel.css";
 import {
     ActionGroup,
@@ -81,12 +81,18 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
     const [isLoading, setIsLoading] = useState(false);
     const [error, setError] = useState<string>("");
 
+    // Version tagging state - reuses the same ArtifactLabel model and
+    // LabelsFormGroup editor used elsewhere in the app (e.g. EditMetaDataModal).
+    // Labels are persisted to the version's metadata via GroupsService, mirroring
+    // the load-on-mount / save-on-confirm pattern used by EditMetaDataModal, so
+    // the toolbar reflects the version's real saved tags rather than local-only state.
     const [savedLabels, setSavedLabels] = useState<ArtifactLabel[]>([]);
     const [draftLabels, setDraftLabels] = useState<ArtifactLabel[]>([]);
     const [isEditingLabels, setIsEditingLabels] = useState(false);
     const [isSavingLabels, setIsSavingLabels] = useState(false);
     const [labelsError, setLabelsError] = useState<string>("");
-    const requestIdRef = { current: 0 };
+
+    const requestIdRef = useRef(0);
 
     useEffect(() => {
         let gid: string | null = props.groupId;
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -81,204 +81,24 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
     const [isLoading, setIsLoading] = useState(false);
     const [error, setError] = useState<string>("");
 
-    // Version tagging state - reuses the same ArtifactLabel model and
-    // LabelsFormGroup editor used elsewhere in the app (e.g. EditMetaDataModal).
-    // Labels are persisted to the version's metadata via GroupsService, mirroring
-    // the load-on-mount / save-on-confirm pattern used by EditMetaDataModal, so
-    // the toolbar reflects the version's real saved tags rather than local-only state.
     const [savedLabels, setSavedLabels] = useState<ArtifactLabel[]>([]);
     const [draftLabels, setDraftLabels] = useState<ArtifactLabel[]>([]);
     const [isEditingLabels, setIsEditingLabels] = useState(false);
     const [isSavingLabels, setIsSavingLabels] = useState(false);
     const [labelsError, setLabelsError] = useState<string>("");
+    const requestIdRef = { current: 0 };
 
     useEffect(() => {
         let gid: string | null = props.groupId;
         if (gid === "default") {
             gid = null;
         }
+        const thisRequestId = ++requestIdRef.current;
         groups.getArtifactVersionMetaData(gid, props.artifactId, props.version)
             .then((meta: VersionMetaData) => {
-                const loaded = labelsToList(meta.labels);
-                setSavedLabels(loaded);
-                setDraftLabels(loaded);
-            })
-            .catch(() => {
-                // Non-fatal: if metadata can't be loaded, the toolbar simply starts empty.
-            });
-    }, [props.groupId, props.artifactId, props.version]);
-
-    const openLabelsEditor = (): void => {
-        setDraftLabels(savedLabels);
-        setLabelsError("");
-        setIsEditingLabels(true);
-    };
-
-    const cancelLabelsEditor = (): void => {
-        setDraftLabels(savedLabels);
-        setLabelsError("");
-        setIsEditingLabels(false);
-    };
-
-    const saveLabels = (): void => {
-        let gid: string | null = props.groupId;
-        if (gid === "default") {
-            gid = null;
-        }
-        setIsSavingLabels(true);
-        setLabelsError("");
-        groups.updateArtifactVersionMetaData(gid, props.artifactId, props.version, {
-            labels: listToLabels(draftLabels)
-        })
-            .then(() => {
-                setSavedLabels(draftLabels);
-                setIsEditingLabels(false);
-            })
-            .catch((err: any) => {
-                setLabelsError(err?.message || "Error saving tags for this version.");
-            })
-            .finally(() => {
-                setIsSavingLabels(false);
-            });
-    };
-
-    const setValue = (name: string, value: any): void => {
-        setValues(prev => ({ ...prev, [name]: value }));
-    };
-
-    const doRender = (): void => {
-        setIsLoading(true);
-        setError("");
-        setValidationErrors([]);
-        setRenderedOutput("");
-
-        let gid: string | null = props.groupId;
-        if (gid === "default") {
-            gid = null;
-        }
-
-        groups.renderPromptTemplate(gid, props.artifactId, props.version, values)
-            .then((response: RenderPromptResponse) => {
-                setRenderedOutput(response.rendered || "");
-                if (response.validationErrors && response.validationErrors.length > 0) {
-                    setValidationErrors(response.validationErrors);
+                if (thisRequestId !== requestIdRef.current) {
+                    return;
                 }
-            })
-            .catch((err: any) => {
-                setError(err?.message || "Error rendering prompt template");
-            })
-            .finally(() => {
-                setIsLoading(false);
-            });
-    };
-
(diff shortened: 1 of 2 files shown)


aclfe:
Point 3 is still reserved for the term

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -173,6 +187,34 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                 <CardTitle>
                     <Title headingLevel="h3" size="md">Test Prompt</Title>
                 </CardTitle>
+                <Toolbar className="playground-toolbar" inset={{ default: "insetNone" }}>
+                    <ToolbarContent>
+                        <ToolbarItem>
+                            <LabelGroup>
```
paoloantinori: The `labels` state is purely local (`useState`). With no load-on-mount or save-on-change call to the artifacts/groups API, the labels a user adds here vanish on refresh or navigation. The PR description says "current tags show up as compact labels" — those are actually draft-only. Either persist them to the version, or make the draft nature explicit in the UI.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -173,6 +187,34 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                 <CardTitle>
                     <Title headingLevel="h3" size="md">Test Prompt</Title>
                 </CardTitle>
+                <Toolbar className="playground-toolbar" inset={{ default: "insetNone" }}>
+                    <ToolbarContent>
+                        <ToolbarItem>
```
paoloantinori, line 255: Using the array index as React `key` for `Label` (`key={idx}`) is an anti-pattern when the list mutates. Labels can be added/removed/reordered through the editor, which can cause incorrect re-renders. Prefer a stable key derived from `label.name` (or `name=value` when value is present).

paoloantinori (commented):
Reusing `LabelsFormGroup` rather than introducing a new tagging component is the right call, and the toolbar placement is sensible. A few concerns worth addressing before merge:

- **Tag state is local-only — no persistence.** `labels` lives in `useState` inside `PromptTemplateTestPanel`. Toggling `isEditingLabels`, editing the form, then closing the editor (or navigating away) silently discards the labels. There's no call to `GroupsService` / `ArtifactsResource` to persist them to the version, and the displayed `Label` chips reflect only the unsaved local state. This contradicts the PR description ("current tags show up as compact labels right in the toolbar") — those are *draft* tags that vanish on reload. Either wire the labels to the version's stored labels (load on mount, PUT/POST on save), or make it explicit in the UI that this is a draft. As-is the feature gives the user a button that does nothing durable.
- **Label `key={idx}` anti-pattern.** Using the array index as React `key` for `Label` works for static lists but causes subtle render bugs when labels are added/removed/reordered (which is exactly what an editor does). Prefer a stable key derived from `label.name` (or `${label.name}=${label.value}` when value is set).
- **`align={{ default: "alignRight" }}` typo risk.** PatternFly's `ToolbarItem` uses `align={{ xl: 'alignRight' }}` style responsive objects; `align={{ default: "alignRight" }}` does technically work in PF6 but is the less common form — worth confirming the alignment renders as expected across the panel's breakpoints, especially since the toolbar lives inside a `CardHeader`.
- **Missing `data-testid` on labels and editor.** The toggle button has `playground-tag-version-button`, but the label chips and the `LabelsFormGroup` container have none. For consistency with the rest of the registry's testable UI and to allow e2e coverage of the new feature, consider adding test ids to the label list and the save/cancel flow once persistence lands.
- **Comment from @aclfe on the PR ("Point 3 is still reserved for the term") suggests this work may not be unblocked yet** — worth confirming with them before this merges, since the PR description claims to address point 3 of #8425.

shreyashp25:
> Point 3 is still reserved for the term

Hi @aclfe, thanks for flagging this. Since a maintainer (@paoloantinori) has since reviewed the actual code, could you help me understand if this specific implementation is still blocked, or if I'm okay to continue addressing the review feedback? Happy to redirect my effort elsewhere if this is already spoken for.

aclfe:
> > Point 3 is still reserved for the term
> 
> Hi @aclfe, thanks for flagging this. Since a maintainer (@paoloantinori) has since reviewed the actual code, could you help me understand if this specific implementation is still blocked, or if I'm okay to continue addressing the review feedback? Happy to redirect my effort elsewhere if this is already spoken for.

I think you should totally work on your pr and refine it, no problem at all. That being said, I would assume that it might take some time to get addressed (or might not) since this PR directly addresses a sub-point of the LFX project

paoloantinori:
Thanks @aclfe, @shreyashp25 — agreed, please go ahead. My earlier review still applies to the current `b331bd5` head; the main open point is that the `labels` state is local to the panel, so tags entered here won't persist or attach to an actual version yet. As long as that's tracked as a follow-up (or wired to a real artifact-version tagging call), no blocker from my side.

shreyashp25:
hello @paoloantinori 
Just pushed the fix (27872a3) that wires the labels up to real persistence via 
getArtifactVersionMetaData/updateArtifactVersionMetaData, along with the other 
points from your review. Should be ready for another look whenever you get a chance!

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -71,6 +81,247 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
...
+            })
+            .finally(() => {
+                setIsLoading(false);
+            });
+    };
+
+    const renderField = (name: string, variable: PromptVariable): React.ReactNode => {
+        const type = (variable.type || "string").toLowerCase();
+
+        if (variable.enum && variable.enum.length
```
paoloantinori: This `if (variable.enum && variable.enum.length` is truncated, and the next lines are literal shell commands (`cd ~/apicurio-registry`, `cat > ... << 'ENDOFFILE'`) followed by a second copy of the whole file starting at `import { FunctionComponent, ...`. The heredoc used to regenerate the file appears to have been pasted into the file itself. The TS won't compile in this state — needs a clean re-commit of `PromptTemplateTestPanel.tsx`.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -71,6 +81,247 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
...
+    const [isEditingLabels, setIsEditingLabels] = useState(false);
+    const [isSavingLabels, setIsSavingLabels] = useState(false);
+    const [labelsError, setLabelsError] = useState<string>("");
+
+    useEffect(() => {
+        let gid: string | null = props.groupId;
+        if (gid === "default") {
+            gid = null;
+        }
+        groups.getArtifactVersionMetaData(gid, props.artifactId, props.version)
```
paoloantinori, line 103: Worth double-checking: this `getArtifactVersionMetaData(...).then(setSavedLabels/setDraftLabels)` has no cancellation. Since the panel stays mounted across `props.version` changes on the same route, an in-flight fetch for version A can overwrite the labels after version B loads. #8880 adds a `newLoaderGuard()` pattern for exactly this case across the other pages — consider adopting it here (or tracking as a follow-up) so the playground doesn't regress.

paoloantinori (commented):
Thanks @shreyashp25 — wiring the toolbar to `getArtifactVersionMetaData` / `updateArtifactVersionMetaData` is the right call and resolves the main persistence concern from my earlier review. One blocker before this can merge, plus one smaller race worth a look.

- **Blocker: shell snippet pasted into `PromptTemplateTestPanel.tsx`.** Looking at the head SHA `27872a3`, the file is broken — around line 178 an `if (variable.enum && variable.enum.length` is cut off mid-expression and is immediately followed by literal shell text:
  ```
  cd ~/apicurio-registry
  cat > ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx << 'ENDOFFILE'
  import { FunctionComponent, useEffect, useState } from "react";
  ```
  After that the entire file content restarts from the `import` block, so the component is defined twice and the file is 533 lines. It looks like the heredoc you used to rewrite the file got pasted *into* the file itself. TypeScript will not compile in this state, so the UI build will fail. Worth re-checking how the commit was produced (editor paste, `git add -p`, etc.) so it doesn't recur.
- **Stale-response race in the labels load effect (pre-existing pattern, newly reachable here).** The new `useEffect` that calls `getArtifactVersionMetaData` has no cancellation, and `PromptTemplateTestPanel` stays mounted across version switches (same route, different `props.version`). If a user flips from version A to B while A's fetch is in flight, A's `setSavedLabels`/`setDraftLabels` can land after B's and show the wrong tags. This is the same class of bug that #8880 is fixing across the other pages via `newLoaderGuard()`. Not a blocker for this PR — the rest of the codebase has the same shape today — but worth either adopting the same guard here or tracking as a follow-up so the playground doesn't regress once #8880 lands.

paoloantinori:
Posted a formal review above — short version: the persistence change looks good, but head `27872a3` has the file broken (a shell heredoc got pasted into `PromptTemplateTestPanel.tsx` itself, so the component is duplicated and won't compile). Needs a clean re-commit of that one file. Also flagged a stale-response race on the new load effect that mirrors what #8880 is fixing — not blocking, but worth a follow-up.

shreyashp25:
hello @paoloantinori 
Pushed a fix (587abcf0) — the previous commit had some corrupted content in that 
file. This commit replaces it with a clean version. Also added a guard around the 
label-loading effect to prevent the stale-response race you flagged when switching 
between versions.

Should be good for another look whenever you have time!

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -71,6 +81,67 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
     const [isLoading, setIsLoading] = useState(false);
     const [error, setError] = useState<string>("");
 
+    const [savedLabels, setSavedLabels] = useState<ArtifactLabel[]>([]);
+    const [draftLabels, setDraftLabels] = useState<ArtifactLabel[]>([]);
+    const [isEditingLabels, setIsEditingLabels] = useState(false);
```
paoloantinori, line 91: This guard doesn't work as intended: `requestIdRef` is `{ current: 0 }` (a plain object literal), not `useRef(0)`. A new object is created on every render, so `thisRequestId` and `requestIdRef.current` are always both `1` on the next render — the `!==` check never rejects a stale response. Use `const requestIdRef = useRef(0);` (and import `useRef`) to actually guard the version-switch race.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -173,6 +244,60 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
...
+                    <Grid hasGutter className="playground-labels-editor" data-testid="playground-labels-editor">
+                        <LabelsFormGroup labels={draftLabels} onChange={setDraftLabels} />
+                        {labelsError && (
+                            <Alert variant="danger" isInline title={labelsError} />
+                        )}
+                        <ActionGroup>
+                            <Button
+                                variant="primary"
+                                data-testid="playground-labels-save-button"
+                                onClick={saveLabels}
```
paoloantinori, line 290: `Alert` is used here but isn't in the `@patternfly/react-core` import block — only `Grid, Label, LabelGroup, Spinner, Toolbar, ToolbarContent, ToolbarItem` were added. Same for `ActionGroup` just below. Confirm the build compiles on this head; if not, add both to the import.

paoloantinori (commented):
Re-read head `587abcf0` after the re-push. The file is clean now and persistence is wired correctly, but the stale-response guard that was added to address the version-switch race **does not actually work**, and two JSX components are used without being imported — both should be fixed before merge.

- **`requestIdRef` is not a ref, so the guard is a no-op.** In `PromptTemplateTestPanel.tsx` it's declared as `const requestIdRef = { current: 0 };` — a plain object literal, recreated on every render. `++requestIdRef.current` increments a fresh `{current:0}` each time, the closure captures `thisRequestId = 1`, and on the next render a *new* `{current:0}` is created and incremented to `1` again. The check `thisRequestId !== requestIdRef.current` then evaluates to `1 !== 1` → `false`, so late responses from a previous version are never discarded. To make the guard real, use `const requestIdRef = useRef(0);` (and add `useRef` to the React import). As written it gives a false sense of safety against exactly the race it claims to prevent.
- **`Alert` and `ActionGroup` are used in JSX but not imported.** Both appear in the new labels-editor block (`<Alert variant="danger" ...>` and `<ActionGroup>`), but neither is in the `@patternfly/react-core` import list — the diff only adds `Grid, Label, LabelGroup, Spinner, Toolbar, ToolbarContent, ToolbarItem`. Unless I'm missing an import elsewhere in the file, this won't compile. Worth confirming the build is green on this head.
- Minor: the load `useEffect` depends on `[props.groupId, props.artifactId, props.version]` but not `groups`. That's fine if `useGroupsService()` returns a stable instance (typical for service hooks in this codebase), but worth a quick confirm so the effect doesn't capture a stale client.

paoloantinori:
Thanks @shreyashp25 — the file is clean now and the persistence wiring looks good. I posted a formal review above covering two blockers I want addressed before merge: the `requestIdRef` race guard isn't actually a ref (it's a plain object literal recreated every render, so the `!==` check is always false), and `Alert`/`ActionGroup` appear in the JSX but aren't in the import list. Both should be quick fixes.

paoloantinori:
Hi @shreyashp25 — I took another look at `587abcf0`. The file is clean and the persistence wiring looks good, but the head hasn't changed since my last formal review, so the two blockers I flagged there still stand and I don't have anything new to add on top of them:

1. The `requestIdRef = { current: 0 }` guard is a no-op because it's a plain object literal recreated on every render — it needs to be `useRef(0)` (with `useRef` imported) for the version-switch race to actually be cancelled.
2. `Alert` (and possibly `ActionGroup`) don't appear to be imported in the diff's import block, which will throw on the error path.

Once those two are addressed in a new push, ping me and I'll re-review promptly. The overall feature shape is good — just want these correctness issues closed out before merge.

Nachiket-Roy:
CNCF mentorship issues are intentionally reserved for the Fall term and shouldn't be worked on before applicants are officially accepted. Working on them beforehand can also create conflicts for the eventual assignee, who is expected to implement the solution during the mentorship period.

shreyashp25:
@paoloantinori 
Fixed both — requestIdRef is now a real useRef(0), so the version-switch guard 
actually works as intended. Also confirmed ActionGroup and Alert are already in 
the @patternfly/react-core import list, so that one was a false alarm on the 
diff view, but good to double check.

Ready for another look whenever you get a chance.

shreyashp25:
Thanks for flagging this @Nachiket-Roy — I wasn't aware this was tied to a reserved 
mentorship term. @paoloantinori, since you've been reviewing the actual code changes, 
could you clarify whether this PR should continue, or if it's better to pause/close 
given the mentorship reservation? Happy to follow whatever's appropriate here 
just want to make sure I'm not stepping on the process.

paoloantinori:
Thanks for checking in @shreyashp25, and fair question.

Short version: the process call here belongs to @Nachiket-Roy and @aclfe, not me. My earlier "please go ahead" was scoped to the code review — I hadn't picked up that #8425 was a reserved mentorship term issue at that point, and they're the right people to decide whether it should be worked on before the Fall term cohort is officially accepted. I'd defer to them on that rather than override the reservation.

On the code itself: I just re-read head `066aa58` and the two blockers from my last formal review look resolved — `requestIdRef` is now a proper `useRef(0)` so the version-switch guard actually works, and the PatternFly imports (`ActionGroup`, `Alert`) are in place. So from a review standpoint there's nothing further I need; whenever the process side is settled, it's clear on my end.

My suggestion: hold the merge (not necessarily close) until the mentors confirm whether this issue is in scope for outside contribution this term. If they're fine with it, nothing technical is blocking.

carlesarnal:
> Thanks for checking in @shreyashp25, and fair question.
> 
> Short version: the process call here belongs to @Nachiket-Roy and @aclfe, not me. My earlier "please go ahead" was scoped to the code review — I hadn't picked up that #8425 was a reserved mentorship term issue at that point, and they're the right people to decide whether it should be worked on before the Fall term cohort is officially accepted. I'd defer to them on that rather than override the reservation.
> 
> On the code itself: I just re-read head `066aa58` and the two blockers from my last formal review look resolved — `requestIdRef` is now a proper `useRef(0)` so the version-switch guard actually works, and the PatternFly imports (`ActionGroup`, `Alert`) are in place. So from a review standpoint there's nothing further I need; whenever the process side is settled, it's clear on my end.
> 
> My suggestion: hold the merge (not necessarily close) until the mentors confirm whether this issue is in scope for outside contribution this term. If they're fine with it, nothing technical is blocking.

@shreyashp25 working on something related, and getting yourself familiar to the code before to the term is perfectly fine, but submitting and asking for review is not. This work is reserved for when the term opens.

I'm closing this for now.

carlesarnal:
Took a look at head `066aa58` to verify the fixes from the last review round.

**Previously flagged blockers — both resolved:**

1. **`requestIdRef` race guard**: Now a proper `useRef(0)` (with `useRef` imported), so the stale-response check in the labels-loading effect actually persists across renders. The version-switch guard works as intended. ✅
2. **`Alert` / `ActionGroup` imports**: Both are present in the `@patternfly/react-core` import block at the top of the file. ✅

**One remaining nit if this is ever reopened:**

- None of the 4 commits carry a `Signed-off-by:` trailer (DCO). Would need `git rebase --signoff` or equivalent before merge.

Code-wise, this looks good — the persistence wiring, stable keys, `data-testid` attributes, and the stale-response guard are all in solid shape. Nothing further from me on the implementation side.

PR #8897, opened 2026-07-25
docs(operator): add migration guide for removed app.sql and app.kafkasql fields

### LINKED ISSUE #7317: Remove deprecated operator code from version 3.0.7
## Parent Epic

## Phase
**Phase 3: Structural Improvements**

## Problem

Several operator classes have been deprecated since version 3.0.7 and should be removed:

**Deprecated Classes:**
- `operator/model/src/main/java/io/apicurio/registry/operator/api/v1/spec/DeprecatedSqlSpec.java`
  - `@Deprecated(since = "3.0.7")` - Entire class

- `operator/model/src/main/java/io/apicurio/registry/operator/api/v1/spec/DeprecatedKafkasqlSpec.java`
  - `@Deprecated(since = "3.0.7")` - Entire class

- `operator/model/src/main/java/io/apicurio/registry/operator/api/v1/spec/DeprecatedDataSource.java`
  - `@Deprecated(since = "3.0.7")` - Entire class with 3 deprecated fields

- `operator/model/src/main/java/io/apicurio/registry/operator/api/v1/spec/AppSpec.java` (line 92)
  - `@Deprecated(since = "3.0.7")` - `@JsonProperty("sql")` field

**Additionally (since 3.1.2):**
- `common/src/main/java/io/apicurio/registry/rest/config/JacksonDateTimeCustomizer.java`
  - Lines 31, 41: Deprecated config properties (see issue #6799)

## Proposed Solution

1. Remove deprecated classes entirely
2. Remove deprecated fields from AppSpec
3. Update any references to use the new replacement APIs
4. Add migration documentation for users

## Migration Notes

| Deprecated | Replacement |
|------------|-------------|
| `DeprecatedSqlSpec` | Use storage configuration in AppSpec |
| `DeprecatedKafkasqlSpec` | Use storage configuration in AppSpec |
| `DeprecatedDataSource` | Use environment variables or new spec |
| `AppSpec.sql` field | Use new storage configuration |

## Estimated Impact
- **Code removed:** ~200 lines
- **Risk:** Low (these are deprecated)

## Acceptance Criteria
- [ ] Remove all classes deprecated since 3.0.7
- [ ] Remove deprecated fields from AppSpec
- [ ] Update operator documentation
- [ ] Add migration guide for users upgrading from older versions
- [ ] All operator tests pass

### LINKED ISSUE #10100: docs(operator): document migration from removed app.sql / app.kafkasql to app.storage
The `spec.app.sql` and `spec.app.kafkasql` fields were deprecated in operator 3.0.7 and **removed in
3.3.2** (#8920, closing #7317). No migration documentation shipped with that removal.

Verified against `main`: grepping `operator/docs/` and `docs/` for `app.sql` / `app.kafkasql` returns
nothing. So we removed a field users configure and provided no guidance on moving off it.

## Why this matters

Two things make the migration non-obvious beyond a simple rename:

- **The password shape changed.** The deprecated `app.sql.dataSource.password` accepted a plaintext
  string (`DeprecatedDataSource.password` was a `String`). The replacement
  `app.storage.sql.dataSource.password` is a Secret reference — `name` required, `key` optional and
  defaulting to `password`.
- **The failure mode on upgrade is silent.** `spec.app.sql` is no longer part of the
  `ApicurioRegistry3` CRD, so it is pruned rather than migrated. `SqlStorage.configureDatasource` is
  gated entirely on `AppSpec::getStorage`, so with `spec.app.storage` absent it sets *no* storage
  environment variables at all and Registry falls back to its embedded H2 database, which is ephemeral.
  There is no status condition reporting this, so a cluster can come back up apparently healthy while
  disconnected from its real database.

## Scope

Document the migration in `operator/docs/modules/ROOT/partials/con-persistence-options.adoc`:

- before/after YAML for both SQL and KafkaSQL
- the Secret-reference change for the password field
- the versions involved: deprecated 3.0.7, removed 3.3.2
- the silent-fallback upgrade failure mode, and how to verify after upgrading

## Notes

This is follow-up work for the removal, not part of it. #7317 is correctly closed and correctly sits on
the (now shipped) 3.3.2 milestone — the documentation is new work for a later release, so it needs its
own issue rather than re-pointing that one.

Implemented by #8897.

Aastha204:
> **Note (maintainer, 2026-09-10):** the scope of this PR changed while it was open. The code
> removal it originally contained was also done in #8920, which merged on 2026-08-04 and shipped
> in **3.3.2**; when this branch was later merged with `main`, those changes converged and dropped
> out of the diff. What remains — and what is being merged — is the **migration documentation**,
> which is unique to this PR: #8920 added no docs, and there was previously no guidance anywhere
> in the repository for moving from `spec.app.sql` to `spec.app.storage`. Retitled accordingly.
> The summary below describes the original scope and is kept for context.

---

Fixes #10100

(This PR was originally linked to issue #7317. That issue tracked the *removal*, which shipped in 3.3.2 via
#8920 and is now closed — so it correctly sits on the closed 3.3.2 milestone. The migration
documentation that remains in this PR is follow-up work for 3.3.4 and is tracked by #10100.)

## Summary
- Remove operator CR model types deprecated since 3.0.7 (`DeprecatedSqlSpec`, `DeprecatedKafkasqlSpec`, `DeprecatedDataSource`) and the `AppSpec.sql` / `AppSpec.kafkasql` fields
- Remove `SqlCRUpdater` / `KafkaSqlCRUpdater` auto-migration helpers and their deprecated SQL/KafkaSQL example fixtures
- Keep ingress host migration intact (`IngressCRUpdater` + `CRUpdateITTest` / `simple-deprecated` example)
- Add migration documentation for moving from `app.sql` / `app.kafkasql` to `app.storage.*`

## Why
These fields have been deprecated since 3.0.7 in favor of `spec.app.storage`. Removing the compatibility layer reduces operator complexity and makes the supported configuration the only path.

## Test plan
- [x] `./mvnw -f operator/pom.xml test-compile -pl model,controller -am -DskipTests -DskipOperatorTests=true`
- [x] Checkstyle passed as part of the operator build


operator/controller/src/main/java/io/apicurio/registry/operator/ApicurioRegistry3Reconciler.java
```diff
@@ -201,8 +199,6 @@ public UpdateControl<ApicurioRegistry3> reconcile(ApicurioRegistry3 primary,
         // Operator will attempt to update the CR to use the newer fields if possible.
         // This has to be done first, so subsequent functionality can deal with new fields only.
         var update = IngressCRUpdater.update(primary);
-        update = SqlCRUpdater.update(primary, context) || update;
-        update = KafkaSqlCRUpdater.update(primary) || update;
         if (update) {
```
paoloantinori, line 202: Removing the `SqlCRUpdater.update` / `KafkaSqlCRUpdater.update` calls means an existing CR that still uses `spec.app.sql` or `spec.app.kafkasql` is no longer auto-migrated. The reconciler will proceed against `spec.app.storage` (now `null` for those CRs). Worth confirming the resulting status error clearly tells users to move their config to `spec.app.storage` (and ideally links to the `registry-persistence-storage-migration` doc anchor).

operator/docs/modules/ROOT/partials/con-persistence-options.adoc
```diff
@@ -41,5 +41,48 @@ ifdef::service-registry[]
...
+            key: password
+----
+
+For KafkaSQL, replace `spec.app.kafkasql` with `spec.app.storage.type: kafkasql` and
+`spec.app.storage.kafkasql`.
+
+IMPORTANT: Existing custom resources that still use `spec.app.sql` or `spec.app.kafkasql`
+must be updated before upgrading to an operator version that removes these fields.
+Automatic migration of those fields is no longer performed.
+
```
paoloantinori, line 88: The `xref:registry-persistence-sql[]` and `xref:registry-persistence-kafkasql-plain[]` targets must exist elsewhere in `operator/docs`, otherwise the Antora build will emit broken-link warnings. Worth a quick grep to confirm both anchors resolve.

paoloantinori (commented):
Removal of the `app.sql` / `app.kafkasql` deprecated compatibility layer (CR types, updaters, IT fixtures) is a clean chore, and the migration doc in `con-persistence-options.adoc` is well written. The `IngressCRUpdater` path and its IT fixture are correctly left intact. A few points to verify before merging:

- **Semver / breaking-change policy.** These CR fields were `@Deprecated(since = "3.0.7")`. Removing them in a patch or minor release of a Kubernetes operator is a hard breaking change for any user who still has `spec.app.sql` or `spec.app.kafkasql` in their manifests: their `AplicurioRegistry3` CR will fail CRD validation (or deserialization) against the new operator. Worth confirming the target release is explicitly a major bump (or that the project has already communicated this removal in prior release notes), and that issue #7317 captured that intent.
- **Failure mode for existing CRs.** With the auto-migrators gone, an existing CR that still uses `app.sql.dataSource` will no longer be silently migrated — the reconciler now skips the `SqlCRUpdater.update`/`KafkaSqlCRUpdater.update` calls and proceeds against `spec.app.storage`, which would be `null`. This likely surfaces as a status error rather than a crash (good), but it's worth confirming the `OperatorErrorConditionManager` produces a clear message pointing users at `spec.app.storage`, since the migration doc is the only guidance they'll get. Consider referencing the doc anchor (`registry-persistence-storage-migration`) from that error message.
- **Doc xref targets.** `con-persistence-options.adoc` adds `xref:registry-persistence-sql[]` and `xref:registry-persistence-kafkasql-plain[]` under "Additional resources". These IDs must exist elsewhere in the operator docs, otherwise the Antora build will emit broken-link warnings. Worth a quick `grep -rn 'registry-persistence-sql\|registry-persistence-kafkasql-plain'` across `operator/docs`.
- **Doc copy.** Small nit: the `After (required)` YAML uses `password` as a SecretKeyRef object but the `Before` example shows the plaintext literal — readers scanning quickly may miss that the migration also changes the password field's shape, not just its location. A one-line note that the password must move into a Secret would help.

operator/docs/modules/ROOT/partials/con-persistence-options.adoc
Aastha204: Thanks for the review @paoloantinori!

**Doc xrefs (#3):** Verified — both anchors exist:
- `registry-persistence-sql` in `proc-persistence-sql.adoc`
- `registry-persistence-kafkasql-plain` in `proc-persistence-kafkasql-plain.adoc`

**Doc copy (#4):** Pushed a follow-up commit clarifying that the password must move from a plaintext string to a Secret reference.

**Semver / breaking change (#1):** @carlesarnal this removal comes from #7317 (opened by you as part of #7305). Can you confirm this is intentional for the current release train, and whether it should be called out as a breaking change in the release notes?

**Failure mode (#2):** Happy to improve the status message if you want that in this PR. Today, with `spec.app.storage` missing, SQL/KafkaSQL config is simply skipped (`ifPresent` on storage), so old CRs may quietly fall back toward in-memory-like behavior rather than a clear “please migrate to `spec.app.storage`” error. Let me know if you'd like me to add an explicit validation/error for that case.

operator/model/src/main/java/io/apicurio/registry/operator/api/v1/spec/AppSpec.java
```diff
@@ -96,29 +95,6 @@ Configure features of the Apicurio Registry backend (app).
...
-     * to update the fields automatically.
-     */
-    @JsonProperty("kafkasql")
-    @JsonPropertyDescription("""
-            DEPRECATED: Use the `app.storage.type` and `app.storage.kafkasql` fields instead.
-             The operator will attempt to update the fields automatically.""")
-    @JsonSetter(nulls = Nulls.SKIP)
-    private DeprecatedKafkasqlSpec kafkasql;
-
     public StorageSpec withStorage() {
```
paoloantinori, line 98: Removing `sql` / `kafkasql` here is the breaking point. If a cluster upgrades the operator while a CR still has `spec.app.sql`, the field either gets silently dropped (if `FAIL_ON_UNKNOWN_PROPERTIES=false`) or rejected at CRD admission (if the schema strips the field). Worth confirming which path the generated CRD + ObjectMapper takes, and surfacing it as an explicit error condition rather than a silent config loss. The migration doc warns users, but the operator's own behavior on an unmigrated CR should be intentional.

operator/docs/modules/ROOT/partials/con-persistence-options.adoc
```diff
@@ -41,5 +41,52 @@ ifdef::service-registry[]
...
+accepted a plaintext string; `app.storage.sql.dataSource.password` must be a Kubernetes
+Secret reference (`name` + `key`).
+
+For KafkaSQL, replace `spec.app.kafkasql` with `spec.app.storage.type: kafkasql` and
+`spec.app.storage.kafkasql`.
+
+IMPORTANT: Existing custom resources that still use `spec.app.sql` or `spec.app.kafkasql`
+must be updated before upgrading to an operator version that removes these fields.
+Automatic migration of those fields is no longer performed.
+
```
paoloantinori, line 88: It would help to spell out the actual failure mode here (admission rejection vs. silent field drop) so ops teams know what to monitor for during the upgrade window. A pointer to the release notes for #7317 would also aid discoverability.

paoloantinori (commented):
Re-review after the latest push — the removal itself remains clean, but one upgrade-path risk is worth confirming before this lands, since it changes the failure mode for clusters that haven't migrated yet.

- **Hard break for unmigrated CRs.** Removing `@JsonProperty("sql")` / `@JsonProperty("kafkasql")` from `AppSpec` (and the backing `DeprecatedSqlSpec` / `DeprecatedKafkasqlSpec` types) means a `ApicurioRegistry3` CR that still carries `spec.app.sql` or `spec.app.kafkasql` will no longer deserialize cleanly once the new operator image runs. The exact behavior depends on (a) whether `AppSpec` / the ObjectMapper is configured with `FAIL_ON_UNKNOWN_PROPERTIES=false` or `@JsonIgnoreProperties(ignoreUnknown = true)` — in which case the old fields are silently dropped and the registry silently loses its datasource config on the next reconcile — and (b) whether the generated CRD schema strips those fields, in which case `kubectl apply` of the existing CR is rejected at admission. Worth confirming which one happens and, if it's silent drop, whether a validation/error condition should be raised instead so the breakage is observable.
- **Doc wording.** The migration section in `con-persistence-options.adoc` correctly says existing CRs "must be updated before upgrading," but it would help to state the actual failure mode (admission rejection vs. silent config loss) so operators know what to look for in their automation. A one-line note pointing at the release notes for #7317 would also help discovery, since this is a removal of fields deprecated since 3.0.7.
- The removal of `SqlCRUpdater` / `KafkaSqlCRUpdater` and the corresponding IT fixtures in `CRUpdateITTest` is consistent — the ingress-host migration path (`IngressCRUpdater` + `simple-deprecated` example) is correctly preserved, so the test still covers the remaining auto-migration case.

Overall this is the right direction; just want the runtime behavior for unmigrated CRs to be deliberate rather than accidental.

operator/docs/modules/ROOT/partials/con-persistence-options.adoc
Aastha204: Thanks for the re-review @paoloantinori.

I checked the generated CRD and reconciler path (not a live cluster apply yet):

- The generated `app` schema includes `storage` but no longer `sql` / `kafkasql`.
- Storage env configuration runs only from `spec.app.storage` (`ifPresent`); if it’s missing, SQL/KafkaSQL setup is skipped and Registry falls back to the default in-memory storage.
- There is currently no dedicated `OperatorError`/validation condition for the removed fields.

I updated the migration doc to describe that upgrade failure mode and linked #7317.

@carlesarnal — do you want this PR to also add an explicit status/validation error when storage config is missing after upgrade (so the break is observable), or is documenting the behavior + a release-notes callout enough?

carlesarnal:
## Review

Went through all 11 changed files. The removal is clean and consistent.

**What looks good:** All 3 commits have DCO sign-off and conventional format. Deprecated model types, updaters, test fixtures cleanly removed with no dangling references. IngressCRUpdater path correctly preserved. Import cleanup in AppSpec.java and CRUpdateITTest.java is correct. Doc xref targets both resolve. Migration doc is well-structured.

**Minor suggestion:** In the Upgrade failure mode section of con-persistence-options.adoc, the phrasing 'expected to be ignored or dropped' could be more definitive since the analysis confirms the fields are silently dropped. Not blocking.

**Open items (maintainer decision, from paoloantinori):** 1) Semver/release confirmation for this breaking change. 2) Whether to add an explicit status condition when spec.app.storage is absent after upgrade.

**CI note:** License Compliance shows 2 issues but passes on main. Likely unrelated.

Overall well-executed removal. Code and docs are solid.

Aastha204:
> ## Review
> Went through all 11 changed files. The removal is clean and consistent.
> 
> **What looks good:** All 3 commits have DCO sign-off and conventional format. Deprecated model types, updaters, test fixtures cleanly removed with no dangling references. IngressCRUpdater path correctly preserved. Import cleanup in AppSpec.java and CRUpdateITTest.java is correct. Doc xref targets both resolve. Migration doc is well-structured.
> 
> **Minor suggestion:** In the Upgrade failure mode section of con-persistence-options.adoc, the phrasing 'expected to be ignored or dropped' could be more definitive since the analysis confirms the fields are silently dropped. Not blocking.
> 
> **Open items (maintainer decision, from paoloantinori):** 1) Semver/release confirmation for this breaking change. 2) Whether to add an explicit status condition when spec.app.storage is absent after upgrade.
> 
> **CI note:** License Compliance shows 2 issues but passes on main. Likely unrelated.
> 
> Overall well-executed removal. Code and docs are solid.

Thanks @carlesarnal - pushed the wording tweak to say the removed fields are silently dropped.

Aastha204:
Hi @jsenko - gentle bump. This is ready from the contributor side (doc tweak applied after @carlesarnal's review). Still waiting on maintainer confirmation for:
1) target release / semver for this breaking removal
2) whether we should add a status condition when spec.app.storage is absent after upgrade

Happy to implement (2) if that's preferred.

carlesarnal:
This has a conflict now. Please fix it.

Aastha204:
> This has a conflict now. Please fix it.

Hi , I’ve resolved the merge conflicts with `main`.

jsenko (approved):
Approving. Thank you for your patience on this one, @Aastha204 — it sat open far longer than it should have, and what happened to it was our fault, not yours.

## What happened

You opened this on 2026-07-25 with `Fixes #7317`. On 2026-08-04, #8920 merged and removed the same code — the same model types, the same updaters, the same fixtures. When you later merged `main` into your branch, the two changes converged and your removals dropped out of the diff entirely, leaving only the documentation. That's why the title no longer matched the contents.

Two contributors ended up doing the same work in parallel because of gaps on our side: #7317 had no assignee, so there was nothing to signal the issue was already being worked, and #8920 never used a closing keyword, so the issue stayed open long after the work shipped. Both PRs predate the CI checks we have since added that enforce issue assignment and require a linked issue per PR. This particular collision should not be able to happen again — and if it does, that is a bug in our process for us to fix.

I have now closed #7317, with a comment recording that the removal shipped in 3.3.2 and pointing at both PRs.

## Why this PR is still worth merging

Your documentation is unique and it fills a real gap. #8920 added no docs at all, and there was nothing anywhere in `operator/docs/` explaining how to move from `spec.app.sql` to `spec.app.storage`. As of 3.3.2 we removed a field users depend on and shipped no migration guidance with it. This PR is that guidance.

I verified every claim in it against the operator code before and after #8920, and the substance was right:

- `spec.app.sql` / `spec.app.kafkasql` are indeed gone from `AppSpec`
- the "before" example matches the old `DeprecatedDataSource` exactly, including the plaintext `password` string
- the "after" example is valid — `StorageType.POSTGRESQL("postgresql")`, and `StorageSpec.sql` → `SqlSpec.dataSource` → `DataSourceSpec` is the real path
- the KafkaSQL instruction is correct
- there genuinely is no status condition reporting the removed fields

Your call on the upgrade failure mode was the most valuable part, and it checks out in the code. `SqlStorage.configureDatasource` is gated entirely on `AppSpec::getStorage`:

```java
ofNullable(primary.getSpec()).map(ApicurioRegistry3Spec::getApp).map(AppSpec::getStorage)
        .ifPresent(storage -> { ... });
```

With `storage` absent it sets no storage environment variables at all, so Registry falls back to its embedded database. Flagging that it happens silently, with no status condition, was a genuinely useful observation.

## What I pushed

Since you'd been waiting long enough and these were wording-level, I pushed three corrections rather than send it back. Please review them, and push back if you disagree with any:

1. **Named the removal version.** The guide said *"Starting with {operator} 3.0.7, storage must be configured under `spec.app.storage`"*, which merged two separate events: 3.0.7 deprecated the fields but still migrated them automatically; **3.3.2** is where they were removed. I verified this against the release tags — `DeprecatedSqlSpec` is present in `3.3.1` and absent in `3.3.2`. The guide now states both versions, since "which release broke this?" is the first thing a user hits this page to find out.

2. **Corrected the Secret key claim.** The note said the password *"must be a Kubernetes Secret reference (`name` + `key`)"*, implying both are required. `key` is optional — `SqlStorage` passes `"password"` as the default:
   ```java
   var password = new SecretKeyRefTool(dataSource.getPassword(), "password");
   ```
   and `DataSourceSpec` documents *"Key `password` is assumed by default"*. So `name` is required, `key` is optional.

3. **Sharpened the fallback description.** "Registry uses the default in-memory storage" is directionally right but vague. It falls back to an **embedded H2 database**, and the important part is that it is *ephemeral* — contents are lost on pod restart. I also moved the "update before upgrading" warning into the existing `IMPORTANT` block and gave it the concrete version, rather than adding a second admonition saying the same thing.

I also retitled the PR to `docs(operator): add migration guide for removed app.sql and app.kafkasql fields` and added a short note at the top of the description explaining the scope change. Your original summary is kept below it so the comment history still reads sensibly.

Thanks again for the contribution, and sorry for the duplicated effort — that was on us.

Aastha204:
> Approving. Thank you for your patience on this one, @Aastha204 — it sat open far longer than it should have, and what happened to it was our fault, not yours.
> 
> ## What happened
> You opened this on 2026-07-25 with `Fixes #7317`. On 2026-08-04, #8920 merged and removed the same code — the same model types, the same updaters, the same fixtures. When you later merged `main` into your branch, the two changes converged and your removals dropped out of the diff entirely, leaving only the documentation. That's why the title no longer matched the contents.
> 
> Two contributors ended up doing the same work in parallel because of gaps on our side: #7317 had no assignee, so there was nothing to signal the issue was already being worked, and #8920 never used a closing keyword, so the issue stayed open long after the work shipped. Both PRs predate the CI checks we have since added that enforce issue assignment and require a linked issue per PR. This particular collision should not be able to happen again — and if it does, that is a bug in our process for us to fix.
> 
> I have now closed #7317, with a comment recording that the removal shipped in 3.3.2 and pointing at both PRs.
> 
> ## Why this PR is still worth merging
> Your documentation is unique and it fills a real gap. #8920 added no docs at all, and there was nothing anywhere in `operator/docs/` explaining how to move from `spec.app.sql` to `spec.app.storage`. As of 3.3.2 we removed a field users depend on and shipped no migration guidance with it. This PR is that guidance.
> 
> I verified every claim in it against the operator code before and after #8920, and the substance was right:
> 
> * `spec.app.sql` / `spec.app.kafkasql` are indeed gone from `AppSpec`
> * the "before" example matches the old `DeprecatedDataSource` exactly, including the plaintext `password` string
> * the "after" example is valid — `StorageType.POSTGRESQL("postgresql")`, and `StorageSpec.sql` → `SqlSpec.dataSource` → `DataSourceSpec` is the real path
> * the KafkaSQL instruction is correct
> * there genuinely is no status condition reporting the removed fields
> 
> Your call on the upgrade failure mode was the most valuable part, and it checks out in the code. `SqlStorage.configureDatasource` is gated entirely on `AppSpec::getStorage`:
> 
> ```java
> ofNullable(primary.getSpec()).map(ApicurioRegistry3Spec::getApp).map(AppSpec::getStorage)
>         .ifPresent(storage -> { ... });
> ```
> 
> With `storage` absent it sets no storage environment variables at all, so Registry falls back to its embedded database. Flagging that it happens silently, with no status condition, was a genuinely useful observation.
> 
> ## What I pushed
> Since you'd been waiting long enough and these were wording-level, I pushed three corrections rather than send it back. Please review them, and push back if you disagree with any:
> 
> 1. **Named the removal version.** The guide said _"Starting with {operator} 3.0.7, storage must be configured under `spec.app.storage`"_, which merged two separate events: 3.0.7 deprecated the fields but still migrated them automatically; **3.3.2** is where they were removed. I verified this against the release tags — `DeprecatedSqlSpec` is present in `3.3.1` and absent in `3.3.2`. The guide now states both versions, since "which release broke this?" is the first thing a user hits this page to find out.
> 2. **Corrected the Secret key claim.** The note said the password _"must be a Kubernetes Secret reference (`name` + `key`)"_, implying both are required. `key` is optional — `SqlStorage` passes `"password"` as the default:
>    ```java
>    var password = new SecretKeyRefTool(dataSource.getPassword(), "password");
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    and `DataSourceSpec` documents _"Key `password` is assumed by default"_. So `name` is required, `key` is optional.
> 3. **Sharpened the fallback description.** "Registry uses the default in-memory storage" is directionally right but vague. It falls back to an **embedded H2 database**, and the important part is that it is _ephemeral_ — contents are lost on pod restart. I also moved the "update before upgrading" warning into the existing `IMPORTANT` block and gave it the concrete version, rather than adding a second admonition saying the same thing.
> 
> I also retitled the PR to `docs(operator): add migration guide for removed app.sql and app.kafkasql fields` and added a short note at the top of the description explaining the scope change. Your original summary is kept below it so the comment history still reads sensibly.
> 
> Thanks again for the contribution, and sorry for the duplicated effort — that was on us.

Thanks @jsenko — really appreciate you taking the time to review this and fix up the doc wording. The version clarification and the H2 fallback note make it much clearer.

Also thanks for explaining what happened with #8920. I was a bit confused when the code changes disappeared after merging main, so that helped a lot. Glad the migration docs are still useful.

PR #8898, opened 2026-07-25
feat(rest): add optional version state filter to searchVersionsByContent

### LINKED ISSUE #8895: Add optional state query parameter to searchVersionsByContent
Currently, the `POST /search/versions` (`searchVersionsByContent`) endpoint does not natively support filtering artifacts by their version state on the server side. 

Due to this, some downstream clients (such as the Java Schema Resolver in `RegistryClientFacadeImpl.java`) are forced to retrieve all matching versions over the network and then perform state filtering locally to exclude `DISABLED` versions. For registries with a high volume of artifact versions, this leads to unnecessary bandwidth consumption, higher memory overhead on the client, and wasted compute cycles on the server returning records that the client immediately discards.

### Proposed Solution
We should introduce an optional `state` query parameter to the `POST /search/versions` endpoint in the OpenAPI specification, mirroring the existing behavior of `GET /search/versions`.
**Implementation details:**
- Add an optional `state` parameter of type `VersionState` to the OpenAPI spec for `searchVersionsByContent`.
- Update `SearchResourceImpl` to ingest the parameter and inject it into the `SearchFilter` set.
- (Verified) The existing `SearchFilter.ofState()` mechanism is completely supported across all storage backends (SQL, KafkaSQL, read-only modes, and Elasticsearch).
**Benefits to Apicurio Registry and its users:**
1. **Performance:** Future clients can push filtering down to the database tier, drastically reducing bandwidth and memory usage.
2. **Consistency:** It aligns the POST search endpoint's capabilities with the GET search endpoint.
3. **Safety:** Because the parameter is strictly optional, existing behavior is preserved exactly. Current clients (like the Java Schema Resolver) will continue functioning without breaking semantics, giving maintainers time to discuss if/when those clients should adopt the server-side filter.
### Additional Context
I have a draft PR ready for this that implements the API capability along with integration tests. I will open a PR for this shortly

doraem-on:
### Description

This PR introduces an optional `state` query parameter to the `POST /search/versions` (`searchVersionsByContent`) REST endpoint, aligning its capabilities with the existing `GET /search/versions` endpoint.

When the state parameter is supplied, the server applies the optional state filter using the existing SearchFilter mechanism. The expectation is that storage implementations honoring SearchFilter will apply the additional state predicate

If the parameter is omitted, the API continues to return all matched versions across all states, preserving strict backward compatibility.

Fixes #8895

### Motivation

Some downstream clients (such as the Java Schema Resolver in `RegistryClientFacadeImpl.java`) currently retrieve all matching versions over the network and then perform state filtering locally to exclude `DISABLED` versions. 

For registries with a high volume of artifact versions, this leads to:
1. Unnecessary bandwidth consumption.
2. Higher memory overhead on the client.
3. Wasted compute cycles on the server returning records that the client immediately discards.

Adding an optional server-side filter provides the foundational capability for future clients to push filtering down to the database tier. 

### Tests Added / Updated

- **`testSearchVersionsByContentWithState`**: Added a comprehensive REST-assured integration test in `SearchVersionsTest.java`. It verifies:
  - `POST /search/versions` **with no state parameter** correctly returns all valid versions (Regression test).
  - `POST /search/versions?state=ENABLED` correctly returns 0 results for a disabled artifact.
  - `POST /search/versions?state=DISABLED` correctly returns 1 result for a disabled artifact.
  - `POST /search/versions?state=DEPRECATED` correctly returns 0 results.

### What is explicitly NOT included in this PR

- **Changes to the Java Schema Resolver:** `RegistryClientFacadeImpl.java` has intentionally been left untouched. Current clients will continue functioning without breaking semantics, allowing maintainers time to discuss if/when those clients should adopt the server-side filter and how multiple active states (e.g., `ENABLED` + `DEPRECATED` + `DRAFT`) should be handled.

### Checklist

- [x] `./mvnw test-compile -pl app -am -DskipTests` compiles cleanly.
- [x] `openapi.json` has been updated with the new query parameter.
- [x] All commits have DCO sign-off (`Signed-off-by`).
- [x] Commit messages use Conventional Commits format.
- [x] The PR contains no unrelated changes (no whitespace fixes or untouched files).


app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
```diff
@@ -362,6 +362,9 @@ public VersionSearchResults searchVersionsByContent(Boolean canonical, String ar
         if (!StringUtil.isEmpty(artifactId)) {
             filters.add(SearchFilter.ofArtifactId(artifactId));
         }
+        if (state != null) {
```
paoloantinori, line 367: Worth double-checking: the PR description states native backend delegation for the state filter across SQL/KafkaSQL/readonly/Elasticsearch. Is `SearchFilter.ofState` actually consumed by every storage implementation's query path, or could it be a no-op on some backend (e.g. the readonly/in-memory profile)? If it is silently ignored anywhere, the filter would be misleading. The test only covers the default profile.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java
```diff
@@ -127,6 +127,63 @@ public void testSearchVersionsByContent() throws Exception {
...
+        String searchByCommonContent = artifactContent.replaceAll("Empty API",
+                "testSearchVersionsByContentWithState-api");
+
+        String artifactId = TestUtils.generateArtifactId();
+        createArtifact(group, artifactId, ArtifactType.OPENAPI, searchByCommonContent,
+                ContentTypes.APPLICATION_JSON);
+
+        // Change state to DISABLED
+        io.apicurio.registry.rest.client.models.WrappedVersionState disabled = new io.apicurio.registry.rest.client.models.WrappedVersionState();
+        disabled.setState(io.apicurio.registry.rest.client.models.VersionState.DISABLED);
```
paoloantinori: Minor: inline fully-qualified names (`io.apicurio.registry.rest.client.models.WrappedVersionState`, `org.hamcrest.Matchers.equalTo`) are inconsistent with the file's style — the rest of the class uses clean imports. Consider promoting these to imports for readability.

paoloantinori (commented):
Clean addition — wiring the optional `state` filter into `searchVersionsByContent` matches `searchVersions` closely, and the OpenAPI + test coverage is solid. A few points worth double-checking before merge:

- **OpenAPI param ordering / generated clients.** The new `state` parameter was inserted into `openapi.json` *between* `artifactId` and `skipCount`. `SearchResourceImpl.searchVersionsByContent` already follows the OpenAPI ordering (the new `state` sits right after `artifactId`, ahead of `skipCount`), so the Java interface is consistent. Worth confirming the regenerated SDKs (Java/Go/Python/TS) build clean with this reordering, since some downstream client code generator pipelines flag signature changes.
- **`SearchFilter.ofState` semantics across backends.** The PR description claims native delegation to SQL/KafkaSQL/readonly/Elasticsearch. The filter is wired through `SearchFilter.ofState(state)`, but I don't see a test exercising a non-default storage profile. The added `SearchVersionsTest` runs under the default profile only — consider whether an IT under the KafkaSQL or JPA profile is warranted, or at least confirming that `ofState` is honored uniformly rather than silently ignored on some backend.
- **Test fully-qualified imports.** `testSearchVersionsByContentWithState` uses inline FQNs (`io.apicurio.registry.rest.client.models.WrappedVersionState`, `...VersionState`, and `org.hamcrest.Matchers.equalTo`) instead of imports. Functionally fine, but inconsistent with the rest of the file which uses clean imports — a small stylistic nit.
- **Default behavior preserved.** Good that omitting `state` keeps the legacy "return all states" behavior; the test pins this explicitly (first assertion returns count=1 including the DISABLED version).

app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
doraem-on: Thanks for pointing this out, @paoloantinori ,The implementation relies entirely on the existing `SearchFilter.ofState(...)` path rather than introducing backend-specific logic. I've updated the PR description to reflect this more accurately.
I did inspect the storage implementations (`SqlSearchRepository.java` lines 280-290 and `ElasticsearchSearchService.java` lines 213-215) and saw that the filter is explicitly mapped to the backend queries. However, since the test only runs against the default profile, I would be more than happy to add a profile-specific IT (like KafkaSQL or JPA)

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java
doraem-on: Thanks , I just pushed a commit that switches those inline FQNs to clean static imports to perfectly match the surrounding test style.

doraem-on:
@paoloantinori Thanks again for the review ,  I've updated the PR description to be more accurate and pushed a new commit to clean up the FQN imports. I also left inline replies confirming how `SearchFilter.ofState()` is handled across the backends. Let me know if there's anything else you'd like me to change

paoloantinori:
@doraem-on Thanks for the cleanup — the FQN import tidy-up and the updated PR description both read clearly. The `SearchFilter.ofState(state)` wiring mirrors `searchVersions` exactly, and the OpenAPI entry plus the four-way state test (`no-filter` / `ENABLED` / `DISABLED` / `DEPRECATED`) give solid coverage of both the happy path and the backward-compat case.

Nothing else from my side — this looks good to merge from a review standpoint. The broader client-side adoption (pushing the filter down in `RegistryClientFacadeImpl` and friends) can follow in a separate change.

carlesarnal (approved):
AI Review: Approved. Clean PR adding optional state filter to searchVersionsByContent, mirroring the existing searchVersions pattern. All maintainer feedback from @paoloantinori addressed (FQN imports cleaned, PR description updated). DCO signed, conventional commits, comprehensive 4-way state test, OpenAPI spec matches existing endpoint. No issues found. LGTM.

EricWittmann:
Thanks for the contribution, @doraem-on! Clean implementation with solid test coverage. Welcome to the project!

PR #8899, opened 2026-07-25
feat(ui): add React ErrorBoundary to prevent white-screen crashes

shivv23:
## Description

Add a React ErrorBoundary at the app root to catch uncaught render errors. Previously, any unhandled render exception would crash the entire application with a white screen. Now users see a friendly recovery page with a reload button.

## Motivation

Currently, zero error boundary components exist in the codebase. A single runtime error in any React component brings down the entire UI with no recovery path for the user. This is a standard React pattern that provides a safety net for the whole application.

## Changes

- Created ErrorBoundary class component in ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx following existing ErrorPage patterns
- Wrapped the Router in App.tsx with ErrorBoundary
- Exported ErrorBoundary from the components barrel (errorPage/index.ts)
- Added .github/PULL_REQUEST_TEMPLATE.md for standardized PR descriptions
- Fixed wrong contributors link in CONTRIBUTING.md (pointed to quarkusio/quarkus instead of Apicurio/apicurio-registry)

## Testing

- [x] Ran npm run lint in ui/ui-app/ - passes (0 errors, only pre-existing warnings)
- [x] Verified the ErrorBoundary renders correctly with PatternFly EmptyState
- [x] ErrorBoundary includes data-testid attributes for E2E testing

## Checklist

- [x] My code follows the existing code style of this project
- [x] I have performed a self-review of my changes
- [x] My changes generate no new warnings
- [x] Added data-testid for testability

### DIFF
--- a/ui/ui-app/src/app/App.tsx
+++ b/ui/ui-app/src/app/App.tsx
@@ -7,6 +7,7 @@ import { BrowserRouter as Router } from "react-router";
 import { ConfigService, useConfigService } from "@services/useConfigService.ts";
 import { LoggerService, useLoggerService } from "@services/useLoggerService.ts";
 import { MainPageWithAuth } from "@app/MainPageWithAuth.tsx";
+import { ErrorBoundaryWithRouter } from "@app/components";
 
 export type AppProps = object;
 
@@ -22,7 +23,9 @@ export const App: FunctionComponent<AppProps> = () => {
 
     return (
         <Router basename={contextPath}>
-            <MainPageWithAuth />
+            <ErrorBoundaryWithRouter>
+                <MainPageWithAuth />
+            </ErrorBoundaryWithRouter>
         </Router>
     );
 };
--- a/ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx
+++ b/ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx
@@ -1,0 +1,79 @@
+import React, { Component, ErrorInfo, ReactNode } from "react";
+import {
+    Button,
+    EmptyState,
+    EmptyStateActions,
+    EmptyStateBody,
+    EmptyStateFooter,
+    PageSection,
+} from "@patternfly/react-core";
+import { ExclamationTriangleIcon } from "@patternfly/react-icons";
+
+export type ErrorBoundaryProps = {
+    children: ReactNode;
+    location?: string;
+};
+
+export type ErrorBoundaryState = {
+    hasError: boolean;
+    error: Error | undefined;
+};
+
+export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
+    constructor(props: ErrorBoundaryProps) {
+        super(props);
+        this.state = { hasError: false, error: undefined };
+    }
+
+    static getDerivedStateFromError(error: Error): ErrorBoundaryState {
+        return { hasError: true, error };
+    }
+
+    componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
+        console.error("[ErrorBoundary] Uncaught error:", error, errorInfo);
+    }
+
+    componentDidUpdate(prevProps: ErrorBoundaryProps): void {
+        if (this.state.hasError && this.props.location && prevProps.location !== this.props.location) {
+            this.setState({ hasError: false, error: undefined });
+        }
+    }
+
+    reloadPage = (): void => {
+        window.location.reload();
+    };
+
+    render(): ReactNode {
+        if (this.state.hasError) {
+            return (
+                <PageSection hasBodyWrapper={false} className="ps_error">
+                    <div className="centerizer">
+                        <EmptyState
+                            headingLevel="h4"
+                            icon={ExclamationTriangleIcon}
+                            titleText="Something went wrong"
+                        >
+                            <EmptyStateBody>
+                                An unexpected error occurred. Try reloading the page.
+                                If the issue persists, reach out to your administrator.
+                            </EmptyStateBody>
+                            <EmptyStateFooter>
+                                <EmptyStateActions>
+                                    <Button
+                                        variant="primary"
+                                        onClick={this.reloadPage}
+                                        data-testid="error-boundary-reload-btn"
+                                    >
+                                        Reload page
+                                    </Button>
+                                </EmptyStateActions>
+                            </EmptyStateFooter>
+                        </EmptyState>
+                    </div>
+                </PageSection>
+            );
+        }
+
+        return this.props.children;
+    }
+}
--- a/ui/ui-app/src/app/components/errorPage/ErrorBoundaryWithRouter.tsx
+++ b/ui/ui-app/src/app/components/errorPage/ErrorBoundaryWithRouter.tsx
@@ -1,0 +1,8 @@
+import React, { FunctionComponent } from "react";
+import { useLocation } from "react-router";
+import { ErrorBoundary, ErrorBoundaryProps } from "./ErrorBoundary";
+
+export const ErrorBoundaryWithRouter: FunctionComponent<ErrorBoundaryProps> = (props) => {
+    const location = useLocation();
+    return <ErrorBoundary location={location.pathname}>{props.children}</ErrorBoundary>;
+};
--- a/ui/ui-app/src/app/components/errorPage/index.ts
+++ b/ui/ui-app/src/app/components/errorPage/index.ts
@@ -1,3 +1,5 @@
+export * from "./ErrorBoundary";
+export * from "./ErrorBoundaryWithRouter";
 export * from "./ErrorPage";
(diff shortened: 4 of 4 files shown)


ui/ui-app/src/app/App.tsx
```diff
@@ -21,8 +22,10 @@ export const App: FunctionComponent<AppProps> = () => {
     logger.info("[App] Using app contextPath: ", contextPath);
 
     return (
-        <Router basename={contextPath}>
-            <MainPageWithAuth />
-        </Router>
+        <ErrorBoundary>
```
paoloantinori: Wrapping the `Router` means the boundary never sees route changes, so `hasError` can't be cleared without a full reload. Consider moving it *inside* `Router` (so it has access to `useLocation`) or subscribing to history changes in `componentDidUpdate` to reset state on navigation — that gives soft recovery for route-specific render errors.

ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx
```diff
@@ -0,0 +1,72 @@
...
+export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
+    constructor(props: ErrorBoundaryProps) {
+        super(props);
+        this.state = { hasError: false, error: undefined };
+    }
+
+    static getDerivedStateFromError(error: Error): ErrorBoundaryState {
+        return { hasError: true, error };
+    }
+
```
paoloantinori, line 31: `componentDidCatch` only fires for render/lifecycle/constructor errors — not async or event-handler errors. A one-line comment stating that scope would help future readers who might wire `onClick`/`useEffect` failures expecting them to land here.

ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx
```diff
@@ -0,0 +1,72 @@
...
+    componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
```
paoloantinori, line 32: `console.error` is lost in production. Render crashes are exactly the signal worth persisting — worth forwarding `error` + `errorInfo.componentStack` to a standalone logger call (the hook-based `LoggerService` can't be used here since it lives below the boundary).

paoloantinori (commented):
Nice addition — a top-level `ErrorBoundary` is a real gap (the codebase currently has none), and the visual treatment correctly mirrors the existing `ErrorPage` (`ps_error` / `centerizer` + PatternFly `EmptyState`). A few observations worth considering:

- **No recovery short of a full page reload.** Once `hasError` flips to `true`, the only escape is `window.location.reload()`. Since the boundary sits *outside* the `Router`, it won't reset on navigation, so a user who hits a render error on one route is stuck on the error screen even for routes that would render fine. The standard pattern is to subscribe to `history`/router location changes in `componentDidUpdate` and call `this.setState({ hasError: false, error: undefined })` when the location changes — that turns a hard reload into a soft recovery for the common case where the error was route-specific. Worth adding, or at least calling out as a deliberate trade-off.
- **Event-handler errors are not caught.** `ErrorBoundary` only catches errors in render/lifecycle/constructors — this is a React limitation, not a bug, but worth a one-line comment in the component so future readers don't assume it covers async/`onClick` failures. Errors thrown in `useEffect` cleanups are also missed.
- **Telemetry is `console.error` only.** The app wires a `LoggerService` (`useLoggerService`), but hooks can't be used inside a class component and it's instantiated *below* the boundary, so it can't be reused here cleanly. Still, render crashes are exactly the signal you most want persisted — consider forwarding `componentDidCatch`'s `error`/`errorInfo` to a lightweight standalone logger call (or the same telemetry sink the backend uses) so these failures don't disappear in production.
- **Unrelated changes mixed in.** This PR bundles the feature with `.github/PULL_REQUEST_TEMPLATE.md` and the `CONTRIBUTING.md` link fix. Both are reasonable, but the Apicurio guideline (and the template itself) asks for atomic commits/PRs. Splitting the docs/repo-config changes into their own PR keeps the review trail clean and lets the feature land independently if there's any debate on the boundary behavior above.

Minor: the `reloadPage` arrow binding duplicates the pattern already in `ErrorPage.tsx` — fine for consistency, but if you ever consolidate these two components, a shared helper would avoid the drift.

EricWittmann (commented):
Thanks for the PR @shivv23 — an app-level ErrorBoundary is a real gap in the codebase and this is a welcome addition. A few things to address:

**1. Remove unrelated changes.** Please remove the `.github/PULL_REQUEST_TEMPLATE.md` and `CONTRIBUTING.md` changes from this PR. The PR template is something we're already working on separately, and the `CONTRIBUTING.md` link fix — while a welcome correction — should be its own PR to keep things atomic. Feel free to open a new PR for the `CONTRIBUTING.md` fix once this one is merged.

**2. Soft recovery on route change.** Agreeing with @paoloantinori's feedback here — since the ErrorBoundary wraps the `Router`, a render error on one route forces a full page reload even if other routes would work fine. Moving the boundary inside `Router` (or resetting `hasError` state on location changes) would let users recover by navigating away instead of reloading the entire app. That would be a nice improvement over the current hard-reload-only approach.

Looking forward to the update!

shivv23:
@EricWittmann Hi! I have made the required fixes. Please have a look. Thanks!

ui/ui-app/src/app/App.tsx
shivv23: @paoloantinori Fixed!

ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx
shivv23: @paoloantinori Fixed!

ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx
shivv23: @paoloantinori Fixed!

EricWittmann (commented):
Thanks for the update @shivv23 — the ErrorBoundary itself looks good and the route-change recovery is a nice addition. A few things to clean up before we can merge:

**1. Unrelated changes in the diff.** The current commit includes changes to `pom.xml` (version bumps, dependency updates), `typescript-sdk/package.json`, and `typescript-sdk/package-lock.json` that aren't related to the ErrorBoundary feature. It looks like these got pulled in during a rebase. Please rebase cleanly so that only the ErrorBoundary changes are in the diff.

**2. Remove the `vitest` dev dependency.** The PR adds `vitest` to `ui/ui-app/package.json` and a `"test": "vitest run"` script, but there are no test files included. Please remove the vitest addition — if we want to introduce a test framework, that should be its own PR (ideally with actual tests).

**3. `package-lock.json` noise.** The `ui/ui-app/package-lock.json` has 360+ lines of changes from the vitest dependency tree and minor version bumps for patternfly/react-router. Removing vitest and rebasing cleanly should take care of most of this.

The ErrorBoundary component code itself is solid — just needs the surrounding diff cleaned up.

shivv23:
@EricWittmann Fixed! Please have a look on the latest changes. Thanks!

paoloantinori (commented):
This is a clean, correct implementation of a standard React error boundary at the app root; three independent reviews (React correctness, PatternFly/router integration, and CLAUDE.md/root-cause compliance) found no functional bugs or blocking issues. The lifecycle is implemented correctly and the boundary is positioned soundly.

- `getDerivedStateFromError` + `componentDidCatch` are both present, which is the correct React pattern (state update for render, side-effect for logging) in `ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx`.
- `ErrorBoundaryWithRouter` calls `useLocation()` correctly inside `<Router>` context, and the location-based reset in `componentDidUpdate` (line ~30-34) compares pathnames properly so navigating away from a broken route recovers without a full reload.
- Placement in `ui/ui-app/src/app/App.tsx` (lines ~25-27) wraps the whole app tree while staying inside `<Router>`, so the fallback UI retains router context; `location.pathname` is passed as a `string`, matching the optional `location?: string` prop.
- Barrel exports (`errorPage/index.ts`) add no name collisions against the existing `ErrorPage`/`AccessErrorPage`/`RateLimitErrorPage` exports.
- Optional follow-up (not blocking): `componentDidCatch` only calls `console.error` (`ErrorBoundary.tsx` ~line 26-28), so production render crashes won't surface to any telemetry. This is acceptable for the PR's stated goal (preventing white screens) but worth wiring to the project's error-reporting channel as a later enhancement.

EricWittmann:
Thanks for the contribution @shivv23! This is a great addition — a top-level ErrorBoundary is a real gap in the codebase and the route-change soft recovery is a nice touch. Appreciate your patience working through the review rounds. Welcome aboard!

shivv23:
@EricWittmann and @paoloantinori Thanks for the appreciation!

PR #8901, opened 2026-07-25
feat: add CloudEvents-compliant DTO classes for webhook payloads

karthik120710:
Add CloudEvents support for webhook notifications

This PR adds CloudEvents-compliant event conversion for registry state changes,
laying the groundwork for webhook delivery. The implementation includes:

- CloudEventDto: CloudEvents 1.0 specification-compliant DTO with fields
  (id, source, type, datacontenttype, data, time, specversion)
- CloudEventConverter: Utility class that converts OutboxEvent instances
  to CloudEventDto, dispatching on StorageEventType enum for type safety
- WebhookDeliveryService: CDI bean that observes SqlOutboxEvent and converts
  to CloudEvents (delivery implementation pending)
- Configurable event source via apicurio.events.cloud-events-source
  (default: /apicurio-registry)
- Configurable enable/disable via apicurio.events.webhook-delivery-enabled
  (default: false to prevent per-event overhead until delivery is implemented)

The converter currently maps 8 of 16 StorageEventType values:
- ARTIFACT_CREATED
- ARTIFACT_DELETED
- ARTIFACT_METADATA_UPDATED
- ARTIFACT_RULE_CONFIGURED
- ARTIFACT_VERSION_CREATED
- ARTIFACT_VERSION_STATE_CHANGED
- GLOBAL_RULE_CONFIGURED
- GROUP_CREATED

Remaining event types will be added in follow-up work.

The SqlOutboxEvent CDI event is fired by SQL storage repositories
(SqlArtifactRepository, SqlVersionRepository, SqlRuleRepository, etc.)
when registry data changes. WebhookDeliveryService observes these events
and converts them to CloudEvents format.

Note: Actual webhook delivery to endpoints will be implemented in a follow-up
PR once the subscription storage layer is available.


app/src/main/java/io/apicurio/registry/events/dto/ArtifactCreatedCloudEvent.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.time.Instant;
+
+/**
+ * CloudEvent wrapper for ArtifactCreated event.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@RegisterForReflection
+public class ArtifactCreatedCloudEvent {
+
```
paoloantinori: Wrapping `CloudEventDto` in a `cloudEvent` field means the serialized JSON is `{"cloudEvent": {...}}` rather than the flat CloudEvents 1.0 structured format, which requires attributes at the root. This applies identically to all eight wrapper classes. Consider `@JsonUnwrapped` here, having the wrapper extend `CloudEventDto`, or returning `CloudEventDto` directly from `from(...)`.

app/src/main/java/io/apicurio/registry/events/dto/ArtifactCreatedCloudEvent.java
```diff
@@ -0,0 +1,45 @@
...
+    private CloudEventDto cloudEvent;
+
+    public ArtifactCreatedCloudEvent() {
+    }
+
+    public static ArtifactCreatedCloudEvent from(io.apicurio.registry.events.ArtifactCreated event, String source) {
+        CloudEventDto dto = new CloudEventDto()
+                .withId(event.getId())
+                .withSource(source)
+                .withType("io.apicurio.registry.events.ArtifactCreated")
```
paoloantinori: `Instant.now()` records conversion time, not event-occurrence time. With the outbox relay in the path, these can differ noticeably. Worth forwarding the source event's own timestamp if it exposes one.

app/src/test/java/io/apicurio/registry/events/dto/CloudEventDtoTest.java
```diff
@@ -0,0 +1,86 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+
+import java.time.Instant;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class CloudEventDtoTest {
+
```
paoloantinori, line 25: Plain `new ObjectMapper()` does not register `jackson-datatype-jsr310`, so `testSerialization` (which sets a non-null `Instant` on `time`) will throw `InvalidDefinitionException` when calling `writeValueAsString`. Register `JavaTimeModule` here, or the test won't pass.

app/src/main/java/io/apicurio/registry/events/EventsConfig.java
```diff
@@ -0,0 +1,27 @@
...
+import io.apicurio.common.apps.config.Info;
+import jakarta.inject.Singleton;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import static io.apicurio.common.apps.config.ConfigPropertyCategory.CATEGORY_EVENTS;
+
+/**
+ * Configuration for registry events.
+ */
+@Singleton
```
paoloantinori: `cloudEventsSource` has no consumer in this diff, and none of the `*CloudEvent.from(...)` factories are called anywhere yet. If wiring lands in a follow-up PR, a note in the description would help; otherwise this is inert config on `main`.

paoloantinori (commented):
Nice foundation for CloudEvents support — the DTO layer is clean and the per-event factory methods read well. I have a few concerns that I think are worth addressing before this lands, the most important being that the serialized JSON shape does not actually match the CloudEvents 1.0 structured format despite the PR title.

- **Wrapper JSON shape breaks the CloudEvents structured format.** Each `*CloudEvent` wrapper exposes a single `cloudEvent` field of type `CloudEventDto`, so Jackson serializes it as `{"cloudEvent": {"specversion": ..., "id": ..., ...}}`. Per the [CloudEvents 1.0 JSON format](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/json-format.md), the CloudEvent attributes MUST sit at the root of the JSON object. The nested shape will be rejected by strict consumers. Cheapest fix is `@JsonUnwrapped` on the field, or have the wrapper extend `CloudEventDto`, or simply return a `CloudEventDto` directly from `from(...)` (the wrapper class may not be needed at all).
- **`time` is set to `Instant.now()` at conversion time, not at event-occurrence time.** The CloudEvents `time` attribute is defined as the occurrence time of the event. Since these events flow through the outbox table (the PR body references `OutboxEvent`), there can be a non-trivial gap between when the event happened and when `from(...)` runs, so receivers will see a timestamp that drifts from reality. Worth checking whether the source event types already carry a creation timestamp that should be forwarded here instead.
- **`CloudEventDtoTest.testSerialization` will most likely fail at runtime.** It uses a plain `new ObjectMapper()` (no `jackson-datatype-jsr310` registered) to serialize a DTO whose `time` field is a non-null `Instant`. Default Jackson cannot handle JSR-310 date/time types and throws `InvalidDefinitionException`. Either register `JavaTimeModule` in the test or drop `withTime(...)` from that case. Mentioning it because it suggests the test may not have actually been executed.
- **Nothing consumes `EventsConfig` or the new DTOs yet.** `cloudEventsSource` is declared but has no reader, and none of the `*CloudEvent.from(...)` factories are called anywhere in this diff. If this is intentionally incremental (DTOs first, wiring in a follow-up), a one-liner in the PR description saying so would help reviewers; otherwise it looks like dead code on `main`.

Minor: the eight wrapper classes are near-identical boilerplate; a single generic `CloudEvent.from(RegistryEvent, type, source)` would cut ~300 lines, but happy with the current shape if the per-type classes are useful for type safety downstream.

karthik120710:
hi @paoloantinori @carlesarnal can u review my pr.

paoloantinori:
Hi @karthik120710 — I had a look again. The branch hasn't changed since my last review, so the open points from there still stand. The one I'd prioritize fixing first is the **event timestamp (`time`) correctness**:

Each wrapper's `extractTimestampFromPayload` looks for a `createdOn` key in the payload and otherwise falls back to `Instant.now()`. That's fine for create events, but for `ArtifactDeleted`, `ArtifactRuleConfigured`, `ArtifactVersionStateChanged`, etc. there is no `createdOn` in the payload — so those events end up with `time = now()` at serialization time, i.e. a wall-clock value that doesn't reflect when the event actually occurred. That undermines the main point of the CloudEvents `time` field, and the `Instant.now()` fallback also makes the timestamp non-deterministic (the tests only assert `assertNotNull(time)`, which hides this).

Suggested direction: derive the timestamp from a field that actually exists on each event payload (or carry a real event timestamp on the source event objects), and assert the expected value in the tests rather than just non-null.

It would also help to confirm the integration path — right now these DTO classes are only referenced by their own tests, so it's not clear how they'll be wired into actual webhook emission. If you can point at (or add) the code that consumes them, that makes the feature reviewable end-to-end.

Once the timestamp issue is addressed, ping me and I'll re-review.

karthik120710:
hi @paoloantinori I have addressed timestamp issue can u review my pr.

app/src/test/java/io/apicurio/registry/events/dto/CloudEventDtoTest.java
```diff
@@ -0,0 +1,87 @@
...
+    @Test
+    public void testSerialization() throws Exception {
+        CloudEventDto dto = new CloudEventDto()
+                .withId("test-id")
+                .withSource("/test-source")
+                .withType("io.apicurio.registry.events.TestEvent")
+                .withData("{\"test\":\"data\"}")
+                .withTime(Instant.parse("2024-01-01T00:00:00Z"));
+
```
paoloantinori, line 69: This round-trip can't validate the CloudEvents wire format. With a vanilla `ObjectMapper + JavaTimeModule`, `WRITE_DATES_AS_TIMESTAMPS` defaults to `true`, so `Instant` serializes as epoch nanos — but the [CloudEvents 1.0 spec](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time) requires `time` as an RFC 3339 string. The round-trip passes for either encoding, so a format regression would slip through. Quarkus disables `WRITE_DATES_AS_TIMESTAMPS` by default (production is fine), but the test should assert the serialized string (e.g. `json.contains("\"2024-01-01T00:00:00Z\"")`) or disable the feature on the test mapper to match production.

paoloantinori (commented):
Thanks @karthik120710 — the timestamp fix landed cleanly: `OutboxEvent` now captures a `final Instant timestamp` via `Instant.now()` and every wrapper threads it into `.withTime(...)`, so the CloudEvent `time` attribute is populated. One issue the new tests expose indirectly is worth tightening before this is wired into actual delivery.

- **`time` serialization is not validated against the CloudEvents wire format.** `CloudEventDtoTest#testSerialization` uses a plain `new ObjectMapper().registerModule(new JavaTimeModule())`. With `JavaTimeModule` registered but `SerializationFeature.WRITE_DATES_AS_TIMESTAMPS` left at its default (`true`), Jackson serializes `Instant` as epoch nanos, **not** the RFC 3339 string the [CloudEvents 1.0 spec requires for `time`](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time). The test only checks round-trip equality (`readValue` back into `Instant`), which passes for either encoding, so it cannot detect a format regression. In production this is most likely fine because Quarkus sets `WRITE_DATES_AS_TIMESTAMPS=false` by default (→ ISO-8601, which is RFC 3339-compatible), but the test should assert that explicitly rather than rely on the runtime default. Suggested: either assert on the serialized string (`assertTrue(json.contains("\"2024-01-01T00:00:00Z\""))`) or configure the test mapper to match production (`objectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)`).
- **Timestamp semantics — worth a quick confirm.** The timestamp is captured at `OutboxEvent` *construction* (`Instant.now()` in the base ctor), not at outbox-persistence time. For the transactional-outbox pattern the persistence commit time is usually the more meaningful `time` (it's the moment the event became durable/visible). If these events are created and then handed to the outbox later, the construction-time clock may be slightly ahead of the actual publish. Not a blocker — just flagging so it's a conscious choice; if it matters, the existing `protected OutboxEvent(id, aggregateId, timestamp)` ctor you added gives a clean path to inject a DB-side timestamp later.

The rest of the DTO layer reads well. Once the `time` format is asserted in the test I think this is in good shape.

paoloantinori:
Hi @karthik120710 — thanks for flagging, I had another look. The timestamp fix looks good (`OutboxEvent` now captures `Instant.now()` and the wrappers thread it into `time`). I left a fresh review with one concrete follow-up: the `CloudEventDtoTest#testSerialization` round-trip can't actually validate the RFC 3339 wire format the CloudEvents spec requires for `time`, since `WRITE_DATES_AS_TIMESTAMPS` defaults to true on a plain mapper. Production is fine (Quarkus disables it), but worth asserting the serialized string explicitly so a regression doesn't slip through.

app/src/test/java/io/apicurio/registry/events/dto/CloudEventDtoTest.java
karthik120710: The test mapper now disables WRITE_DATES_AS_TIMESTAMPS to match production, and testSerialization asserts the serialized string directly.

karthik120710:
@paoloantinori  
Thanks for the re-review. This is already addressed in 251929f8c (just pushed — it was sitting unpushed on my branch when you looked, sorry for the confusion).
 Applied both of your suggestions rather than picking one:
private final ObjectMapper objectMapper = new ObjectMapper().registerModule(new JavaTimeModule())
         .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
 assertTrue(json.contains("\"time\":\"2024-01-01T00:00:00Z\""),
         "Expected time to be serialized as an RFC 3339 string, but was: " + json);

 I confirmed the assertion actually bites by temporarily re-enabling the feature — it fails with "time":1704067200.000000000. Notably only that one assertion failed while every round-trip readValue check
 still passed, which is exactly the blind spot you described. Also added an assertion on deserialized.getTime(), which the original test never checked.
 One gap this doesn't close: it pins the format for a hand-built mapper that mirrors production, not the CDI-injected ObjectMapper the publisher will use. Happy to add a @QuarkusTest asserting against the
 real mapper if you'd rather have that guarantee now instead of when delivery lands.

paoloantinori:
Hi @karthik120710 — thanks for the clarification on the push timing, that makes sense.

I can confirm both suggestions are now in `CloudEventDtoTest`:
- `ObjectMapper` configured with `JavaTimeModule` and `disable(WRITE_DATES_AS_TIMESTAMPS)`, matching the production Quarkus configuration.
- The explicit `assertTrue(json.contains("\"time\":\"2024-01-01T00:00:00Z\""))` asserting the RFC 3339 wire format directly rather than relying on a symmetric round trip.

That second point is the important one — asserting the serialization output (not just round-trip equality) is what actually guards against a regression that could silently produce epoch timestamps. No further changes needed from my side on those points.

The diff also shows the new `OutboxEvent` two-arg/three-arg constructor with `Instant.now()` capture, and every wrapper threads `event.getTimestamp()` into `.withTime(...)`, so the `time` field is populated consistently across all eight event types.

From my side the previous open points are resolved. Deferring the merge call to @carlesarnal / @EricWittmann.

karthik120710:
@paoloantinori thanks for ur response

paoloantinori:
You're welcome, @karthik120710 — happy to help. The DTO layer looks in good shape now that the timestamp and `ObjectMapper` points are resolved. Whenever you're ready for the next iteration (e.g. wiring the wrappers into the webhook delivery path), feel free to ping.

karthik120710:
@paoloantinori I have added Webhook Delivery Observer supports web hook services.

paoloantinori:
Thanks @karthik120710 — the `disable(WRITE_DATES_AS_TIMESTAMPS)` + the round-trip assertion that bites (it failed on the timestamp form) is exactly the right way to lock the RFC 3339 format. Good catch also adding the `deserialized.getTime()` check the original test lacked.

On the gap you flagged: pinning the hand-built mapper is a reasonable stopping point for the DTO PR, but I'd rather have the guarantee against the CDI-injected `ObjectMapper` the publisher will actually use *before* delivery lands rather than discovering a serialization drift then. Please add the `@QuarkusTest` asserting against the real mapper in this PR, since that's the one that actually bites in production. Holding until that's in.

karthik120710:
@paoloantinori Fixed CloudEvent payload serialization (JSONObject → Map, was emitting {"mapType":...,"empty":false} and dropping all event fields), restored event timestamp propagation in ArtifactCreated/ArtifactDeleted/GroupCreated, and added a @QuarkusTest pinning both against the CDI-injected ObjectMapper

paoloantinori:
Thanks @karthik120710 — good catch on the `JSONObject` serialization trap: converting to a `Map` before handing to Jackson is the correct fix, otherwise the bean properties (`mapType`, `empty`) leak onto the wire and the event payload silently vanishes. Restoring timestamp propagation in `ArtifactCreated`/`ArtifactDeleted`/`GroupCreated` and pinning both behaviors against the CDI-injected `ObjectMapper` via `@QuarkusTest` is the right level of coverage — that test will catch any future `ObjectMapperCustomizer` regression. The DTO layer looks solid now.

app/src/main/java/io/apicurio/registry/events/dto/CloudEventConverter.java
```diff
@@ -0,0 +1,64 @@
...
+        return switch (event.getType()) {
+            case "ARTIFACT_CREATED" -> ArtifactCreatedCloudEvent.from((ArtifactCreated) event, source).getCloudEvent();
+            case "ARTIFACT_DELETED" -> ArtifactDeletedCloudEvent.from((ArtifactDeleted) event, source).getCloudEvent();
+            case "ARTIFACT_METADATA_UPDATED" -> ArtifactMetadataUpdatedCloudEvent.from((ArtifactMetadataUpdated) event, source).getCloudEvent();
+            case "ARTIFACT_RULE_CONFIGURED" -> ArtifactRuleConfiguredCloudEvent.from((ArtifactRuleConfigured) event, source).getCloudEvent();
+            case "ARTIFACT_VERSION_CREATED" -> ArtifactVersionCreatedCloudEvent.from((ArtifactVersionCreated) event, source).getCloudEvent();
+            case "ARTIFACT_VERSION_STATE_CHANGED" -> ArtifactVersionStateChangedCloudEvent.from((ArtifactVersionStateChanged) event, source).getCloudEvent();
+            case "GLOBAL_RULE_CONFIGURED" -> GlobalRuleConfiguredCloudEvent.from((GlobalRuleConfigured) event, source).getCloudEvent();
+            case "GROUP_CREATED" -> GroupCreatedCloudEvent.from((GroupCreated) event, source).getCloudEvent();
+            default -> null;
```
paoloantinori: The `default -> null` arm silently drops any event whose type isn't one of the eight literal strings above. Since the case labels are raw strings rather than `StorageEventType.ARTIFACT_CREATED.name()` (etc.), a renamed or newly added `StorageEventType` value will fall through here and vanish with only a DEBUG log in `WebhookDeliveryService`. Consider dispatching on the enum directly (e.g. `StorageEventType.valueOf(event.getType())`) and logging at WARN on the unmatched branch so dropped events are visible rather than silent.

app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java
```diff
@@ -0,0 +1,38 @@
...
+ * Consumes registry outbox events and converts them to CloudEvents for
+ * webhook delivery. Actual delivery to configured webhook subscriptions
+ * will be wired once the subscription storage layer is available.
+ */
+@ApplicationScoped
+public class WebhookDeliveryService {
+
+    private static final Logger log = LoggerFactory.getLogger(WebhookDeliveryService.class);
+
+    private static final String EVENT_SOURCE = "/apicurio-registry";
```
paoloantinori: `EVENT_SOURCE` is a hardcoded constant, but the PR summary describes an `EventsConfig` / `apicurio.events.cloud-events-source` property (default `/apicurio-registry`) — that property isn't injected anywhere in this diff, so the source isn't actually configurable. If the config class is intended to ship here, inject it via `@ConfigProperty` (or `EventsConfig`) and use it; otherwise the summary overstates what's wired.

app/src/main/java/io/apicurio/registry/events/dto/ArtifactCreatedCloudEvent.java
```diff
@@ -0,0 +1,45 @@
...
+@RegisterForReflection
+public class ArtifactCreatedCloudEvent {
+
+    @JsonUnwrapped
+    private CloudEventDto cloudEvent;
+
+    public ArtifactCreatedCloudEvent() {
+    }
+
+    public static ArtifactCreatedCloudEvent from(io.apicurio.registry.events.ArtifactCreated event, String source) {
```
paoloantinori: This wrapper (and its seven siblings) carries only an `@JsonUnwrapped CloudEventDto` and adds no fields of its own, and `CloudEventConverter.toCloudEvent` immediately calls `.getCloudEvent()` on the result — so the wrapper is constructed and discarded. All eight reduce to one parameterized factory: `new CloudEventDto().withId(...).withType(typeConstant).withTime(...).withData(...)` keyed by event type, which also removes the per-class `"io.apicurio.registry.events.X"` string duplication. Worth confirming whether these distinct types are meant to carry event-specific extensions later; if not, collapsing them avoids ~360 lines of near-identical boilerplate.

app/src/test/java/io/apicurio/registry/events/dto/ArtifactCreatedCloudEventTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+        ArtifactCreatedCloudEvent cloudEvent = ArtifactCreatedCloudEvent.from(event, source);
+
+        assertNotNull(cloudEvent);
+        assertNotNull(cloudEvent.getCloudEvent());
+        assertEquals(event.getId(), cloudEvent.getCloudEvent().getId());
+        assertEquals(source, cloudEvent.getCloudEvent().getSource());
+        assertEquals("io.apicurio.registry.events.ArtifactCreated", cloudEvent.getCloudEvent().getType());
+        assertEquals("1.0", cloudEvent.getCloudEvent().getSpecversion());
+        assertEquals("application/json", cloudEvent.getCloudEvent().getDatacontenttype());
```
paoloantinori: `ArtifactMetaDataDto` is never given a `createdOn`, so `ArtifactCreated.of(...)` computes `Instant.ofEpochMilli(0)` = `1970-01-01T00:00:00Z`, and this `assertEquals(event.getTimestamp(), cloudEvent.getTime())` passes trivially (both sides are epoch). Set `createdOn` — as `CloudEventConverterTest.testConvertArtifactCreated` does — so the timestamp round-trip is meaningful and the `Instant.ofEpochMilli(getCreatedOn())` path in `ArtifactCreated.of` is actually exercised.

app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public void onOutboxEvent(@Observes SqlOutboxEvent sqlOutboxEvent) {
+        OutboxEvent outboxEvent = sqlOutboxEvent.getOutboxEvent();
+        CloudEventDto cloudEvent = CloudEventConverter.toCloudEvent(outboxEvent, EVENT_SOURCE);
+        if (cloudEvent == null) {
+            log.debug("No CloudEvent mapping for event type: {}", outboxEvent.getType());
+            return;
+        }
+        log.debug("Converted outbox event {} to CloudEvent type {}", cloudEvent.getId(), cloudEvent.getType());
```
paoloantinori: This observer converts the event and logs it but performs no delivery, and `WebhookDeliveryServiceTest` invokes `onOutboxEvent` directly rather than firing the `SqlOutboxEvent` CDI event — so the `@Observes` wiring (whether/where `SqlOutboxEvent` is actually fired) is unverified. If `SqlOutboxEvent` isn't fired as a CDI event today, this service is inert. A test that fires the event, or a note on what fires it, would close the gap.

paoloantinori (commented):
This PR adds cleanly structured CloudEvents-1.0 DTOs over the existing outbox events, and the serialization tests are notably thorough — they explicitly guard the two real pitfalls (Jackson not understanding `org.json.JSONObject`, and `Instant` time needing RFC-3339 rather than a numeric epoch). The main concerns are a source/config property that's described but never wired, event-type dispatch on raw strings that silently drops unrecognized events, eight wrapper classes that add no fields and are unwrapped immediately at conversion, and an observer that converts-and-logs with no delivery and no test of the actual `@Observes` firing path.

- `WebhookDeliveryService.java:27` — `EVENT_SOURCE` is a hardcoded constant; the `EventsConfig` / `apicurio.events.cloud-events-source` property described in the PR summary is not injected anywhere in the diff, so the source isn't actually configurable.
- `CloudEventConverter.java:49` — the case labels are raw strings rather than `StorageEventType.X.name()`, and the `default -> null` arm silently drops any unrecognized/renamed type with only a DEBUG log; in a delivery pipeline that means silently lost events.
- `ArtifactCreatedCloudEvent.java` (+ 7 siblings) — each wrapper holds only an `@JsonUnwrapped CloudEventDto` and is discarded by `.getCloudEvent()` in `CloudEventConverter`, so all eight reduce to one parameterized factory.
- `ArtifactCreatedCloudEventTest.java:36` — `createdOn` is never set, so `Instant.ofEpochMilli(getCreatedOn())` resolves to `1970-01-01T00:00:00Z` and the timestamp assertion passes trivially; the real timestamp path isn't exercised.
- `WebhookDeliveryService.java:36` — the observer converts and logs but does no delivery, and the test invokes `onOutboxEvent` directly rather than firing the `SqlOutboxEvent` CDI event, so the `@Observes` firing path is unverified.

app/src/main/java/io/apicurio/registry/events/dto/CloudEventConverter.java
karthik120710: StorageEventType enum instead of raw strings, and log at WARN for both unknown types and unmapped enum values instead of silently dropping events. Added a test for the unmapped case.

app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java
karthik120710: Done. I've injected the apicurio.events.cloud-events-source config property via @ConfigProperty with default /apicurio-registry, and added the @Info annotation for documentation.

app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java
karthik120710: updated the JavaDoc in WebhookDeliveryService to document where SqlOutboxEvent is fired — it's fired by SQL storage repositories (SqlArtifactRepository, SqlVersionRepository, SqlRuleRepository, SqlGroupRepository) when registry data changes.

karthik120710:
@paoloantinori I have addressed all the changes and remove the unwanted changes

paoloantinori:
Thanks @karthik120710 — the cleanup landed cleanly on this head. The DTO layer reads well, the `JSONObject → Map` normalization in `CloudEventDto.normalizeData` is the correct guard against the `mapType`/`empty` leak, the `time` field is threaded from `OutboxEvent.timestamp` end-to-end, and the converter tests cover the mapped / unmapped / unknown-type branches including the custom `sourceProvider` overload.

One thing worth being explicit about for whoever merges: `WebhookDeliveryService.onOutboxEvent` currently converts to a `CloudEventDto` and logs at DEBUG but doesn't actually deliver anywhere — the class javadoc says as much ("Actual delivery ... will be wired once the subscription storage layer is available"). That's a reasonable landing point as a foundation PR, but it does mean this ships an `@Observes` path with no observable side effect yet. If the intent is to keep this PR scoped to the DTO + conversion layer only, I'd suggest titling/describing it as such so a maintainer doesn't expect end-to-end webhook delivery from this change. Otherwise, no blocking issues from my side.

carlesarnal:
Review: CloudEvents DTOs for webhook payloads

Nice work iterating through the feedback. The code is in good shape. The DTO design is clean, normalizeData() for JSONObject to Map is a solid catch, and the serialization tests are thorough.

A few minor items:

1. FQN instead of import in CloudEventDto.java line 181: OutboxEvent is referenced via fully-qualified name in from() method instead of adding an import.

2. FQN instead of import in WebhookDeliveryServiceTest.java line 50: CloudEventConverter used as FQN instead of imported normally.

3. Weak assertions in WebhookDeliveryServiceTest: testOnOutboxEvent() has no assertions beyond not throwing. testCloudEventConversion() only checks assertNotNull. Consider adding assertEquals for cloudEvent.getType() and cloudEvent.getSource() to make these tests more meaningful.

4. Commit messages: Most don't follow conventional commit format (feat:, fix:, refactor:). When squashed for merge, use proper format like feat: add CloudEvents 1.0 DTOs for webhook payloads.

Overall this is solid foundation work. Previous review concerns have all been addressed well: enum-based dispatch with WARN logging, configurable source URI, collapsed wrapper classes, and proper timestamp threading.

carlesarnal:
### Re-review: CloudEvents DTOs for webhook payloads

Thanks for the iteration so far — the DTO layer is in solid shape. @paoloantinori's concerns are all resolved, and the core design (enum-based dispatch, normalizeData(), timestamp threading, serialization tests against the CDI-injected mapper) looks good.

However, the items from my previous review are still open. No new commits have been pushed since that review, so these are the same findings:

---

**1. FQN instead of import — CloudEventDto.java:181**
The from() method uses the fully-qualified name io.apicurio.registry.storage.dto.OutboxEvent in the parameter type instead of adding a normal import. Please add import io.apicurio.registry.storage.dto.OutboxEvent; and use just OutboxEvent in the method signature.

**2. FQN instead of import — WebhookDeliveryServiceTest.java:50**
Same issue: io.apicurio.registry.events.dto.CloudEventConverter.toCloudEvent(...) is called via FQN. Add import io.apicurio.registry.events.dto.CloudEventConverter; and call it as CloudEventConverter.toCloudEvent(...).

**3. Weak assertions in WebhookDeliveryServiceTest**
- testOnOutboxEvent() has no assertions beyond not throwing. At minimum, this should verify the observer path doesn't silently swallow the event — for now since it only logs, a smoke test is acceptable, but please add a comment explaining this is intentionally assertion-free until delivery is wired.
- testCloudEventConversion() only checks assertNotNull(cloudEvent). Please add assertEquals checks for cloudEvent.getType() (e.g., io.apicurio.registry.events.ArtifactCreated) and cloudEvent.getSource() (e.g., /apicurio-registry) to make this test meaningful. The CloudEventConverterTest already does this well — the webhook service test should match that standard.

---

These are all minor fixes. Once addressed, I don't see any remaining blockers from my side.

karthik120710:
@carlesarnal I have addressed all three changes on 
1. CloudEventDto.java:11,178 — OutboxEvent imported and used in from() signature
2. WebhookDeliveryServiceTest.java:9,53 — CloudEventConverter imported and used
3. Assertions — testCloudEventConversion() now asserts getType(), getSource(), getId(), getSpecversion(); testOnOutboxEvent() uses assertDoesNotThrow with a comment noting it's intentionally assertion-free until delivery i

paoloantinori:
Thanks @karthik120710 — I re-checked the current head against the three points from @carlesarnal's re-review, and all three are addressed:

1. **`CloudEventDto.java`** — `OutboxEvent` is now imported and used in the `from(OutboxEvent event, ...)` signature, replacing the FQN.
2. **`WebhookDeliveryServiceTest.java`** — `CloudEventConverter` is imported and `testCloudEventConversion()` drives the conversion through it.
3. **`testCloudEventConversion()` assertions** — it now asserts `getType()` (`io.apicurio.registry.events.ArtifactCreated`), `getSource()` (`/apicurio-registry`), `getId()`, and `getSpecversion()`.

The DTO layer is in good shape. Agreed with @carlesarnal's read that the core design (enum-based dispatch, `normalizeData()` for the `JSONObject → Map` trap, timestamp threading, serialization tests against the CDI-injected mapper) is solid. No further comments from me on the code.

karthik120710:
@paoloantinori @carlesarnal thanks for ur suggestions to polish my pr , can u take a look now its okay to merge .

paoloantinori:
Hi @karthik120710 — had another look at the current head. All the technical points from both my reviews and @carlesarnal's re-review are addressed: the `OutboxEvent` timestamp propagation, the `JSONObject → Map` normalization in `CloudEventDto.normalizeData`, and the `CloudEventConverter.from()` signature using `OutboxEvent`. The test coverage is solid.

The merge decision is a maintainer call — over to @carlesarnal. Thanks for the steady iteration on this one.

app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private static final String CATEGORY_WEBHOOK = "Webhook";
+
+    @ConfigProperty(name = "apicurio.events.cloud-events-source", defaultValue = "/apicurio-registry")
+    @Info(category = CATEGORY_WEBHOOK, description = "CloudEvents source URI for webhook delivery", availableSince = "3.0.0")
+    String eventSource;
+
+    public void onOutboxEvent(@Observes SqlOutboxEvent sqlOutboxEvent) {
+        OutboxEvent outboxEvent = sqlOutboxEvent.getOutboxEvent();
+        CloudEventDto cloudEvent = CloudEventConverter.toCloudEvent(outboxEvent, eventSource);
```
paoloantinori: This `@ApplicationScoped` bean has a synchronous, unconditional `@Observes SqlOutboxEvent`, so it fires for every registry outbox event. For all 8 supported types it builds a full `CloudEventDto` — including a `JSONObject.toMap()` deep-copy of the payload in `CloudEventDto.normalizeData` (CloudEventDto.java) — and then only `log.debug`s the result and discards it; no delivery path exists yet. Worth confirming the per-event conversion is acceptable to run on every write in production while delivery is unwired, or gating/disabling the bean until delivery is wired.

paoloantinori (commented):
The shipped implementation is solid and the serialization tests are well targeted — they pin the CloudEvents RFC 3339 `time` format and verify the JSONObject payload survives Jackson serialization. Two things are worth resolving before merge:

- **PR summary does not match the diff.** The Summary lists 8 dedicated wrapper classes (ArtifactCreatedCloudEvent … GlobalRuleConfiguredCloudEvent) under `common/.../events/dto/` plus a separate `EventsConfig` class, but the diff contains none of them. It ships a single `CloudEventDto` + `CloudEventConverter` in `app/.../events/dto/` and inlines the `apicurio.events.cloud-events-source` property in `WebhookDeliveryService`. Please update the description (or confirm no files are missing) so the scope is accurate.

- **`WebhookDeliveryService` ships an always-on no-op observer.** It is `@ApplicationScoped` with a synchronous, unconditional `@Observes SqlOutboxEvent`, so it runs on every registry outbox event. For all 8 supported types it builds a full `CloudEventDto` — including a `JSONObject.toMap()` deep-copy of the payload via `CloudEventDto.normalizeData` (CloudEventDto.java) — then discards it with only a `log.debug`; no delivery path exists yet. Worth confirming this per-event work is acceptable on every write in production while delivery is unwired, or gating/disabling the bean until delivery lands.

app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java
karthik120710: 1. Added @IfBuildProperty 
2. Removed the webhookDeliveryEnabled field since it's no longer needed
3. updated test cases @TestProfile(WebhookEnabledTestProfile.class 
4. updated the pr discriptio

paoloantinori:
Thanks @karthik120710 — confirmed on the current head. The `@IfBuildProperty(name = "apicurio.events.webhook-delivery-enabled", stringValue = "true")` on `WebhookDeliveryService` is the right Quarkus pattern: the bean (and its `@Observes` registration) only exists when the feature is on, so there's zero per-event overhead in the default deployment, and the now-redundant `webhookDeliveryEnabled` field is correctly gone.

The `WebhookEnabledTestProfile` flipping the property to `true` is the correct way to exercise the observer path under `@QuarkusTest` — that's the only way the bean resolves and `@Inject WebhookDeliveryService` works.

One small thing worth a glance before merge: `WebhookDeliveryServiceTest` still has the import `io.apicurio.registry.events.dto.CloudEventConverter` and `CloudEventDto` (used in `testCloudEventConversion`), which is fine, but the first test (`testOnOutboxEvent`) is now a pure smoke test against the observer entry point — consider naming it `testObserverDoesNotThrowOnValidEvent` so the intent ("no delivery assertions yet, just the observer path is exercised") is explicit. Not blocking.

From my side the PR is good to merge once a maintainer gives the final sign-off.

karthik120710:
@carlesarnal can u took a look at my pr, is it ready to merge .

app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final String CATEGORY_WEBHOOK = "Webhook";
+
+    @ConfigProperty(name = "apicurio.events.cloud-events-source", defaultValue = "/apicurio-registry")
+    @Info(category = CATEGORY_WEBHOOK, description = "CloudEvents source URI for webhook delivery", availableSince = "3.0.0")
+    String eventSource;
+
+    @ConfigProperty(name = "apicurio.events.webhook-delivery-enabled", defaultValue = "false")
+    @Info(category = CATEGORY_WEBHOOK, description = "Enable webhook delivery service (disabled until delivery implementation is complete)", availableSince = "3.0.0")
+    boolean webhookDeliveryEnabled;
+
```
paoloantinori: This observer is synchronous (`@Observes`), so it runs on the calling thread inside the storage transaction that fired `SqlOutboxEvent`. That's fine while it only converts and logs, but when real delivery is wired here, blocking HTTP calls would execute within that transaction and on the repository thread. Worth deciding the seam now -- e.g. `@ObservesAsync`, a transaction-phase-aware observer, or enqueueing onto a durable outbox -- so webhook delivery can't block or roll back registry writes.

paoloantinori (commented):
This is a clean, well-tested foundation for CloudEvents-based webhook delivery: the DTO and converter are CloudEvents-1.0-aligned, the JSONObject->Map payload normalization is correctly asserted against the injected ObjectMapper, and delivery is intentionally a no-op pending the subscription layer. No blocking defects -- the points below are forward-looking design notes and easy cleanups.

- Synchronous observer seam (`app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java:49`): delivery will hang off `@Observes SqlOutboxEvent`, which runs on the firing thread inside the storage transaction; plan async or durable delivery before adding HTTP calls so delivery can't block or roll back registry writes.
- Event coverage (`app/src/main/java/io/apicurio/registry/events/dto/CloudEventConverter.java:52`-55): 8 of 16 `StorageEventType` values are dropped at runtime, including `ARTIFACT_VERSION_CREATED`; documented as follow-up, but consider an exhaustive switch (no null-yielding default) once all are mapped so a future enum addition is a compile error rather than a silent runtime drop.
- `time` wire format (`app/src/main/java/io/apicurio/registry/events/dto/CloudEventDto.java`, `time` field): CloudEvents requires RFC 3339; this currently relies on the global mapper disabling `WRITE_DATES_AS_TIMESTAMPS` (guarded by `CloudEventDtoInjectedMapperTest`). Annotating the field with `@JsonFormat(shape = STRING)` would make the DTO spec-compliant independent of global config.
- Unused config field (`app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java:45`-47): `webhookDeliveryEnabled` is injected but never read in `onOutboxEvent` (the bean is already gated by `@IfBuildProperty`); confirm it's retained for the `@Info` config metadata, or remove.

app/src/main/java/io/apicurio/registry/events/dto/CloudEventConverter.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        return switch (eventType) {
+            case ARTIFACT_CREATED -> CloudEventDto.from(event, source, "io.apicurio.registry.events.ArtifactCreated");
+            case ARTIFACT_DELETED -> CloudEventDto.from(event, source, "io.apicurio.registry.events.ArtifactDeleted");
+            case ARTIFACT_METADATA_UPDATED -> CloudEventDto.from(event, source, "io.apicurio.registry.events.ArtifactMetadataUpdated");
+            case ARTIFACT_RULE_CONFIGURED -> CloudEventDto.from(event, source, "io.apicurio.registry.events.ArtifactRuleConfigured");
+            case ARTIFACT_VERSION_CREATED -> CloudEventDto.from(event, source, "io.apicurio.registry.events.ArtifactVersionCreated");
+            case ARTIFACT_VERSION_STATE_CHANGED -> CloudEventDto.from(event, source, "io.apicurio.registry.events.ArtifactVersionStateChanged");
+            case GLOBAL_RULE_CONFIGURED -> CloudEventDto.from(event, source, "io.apicurio.registry.events.GlobalRuleConfigured");
+            case GROUP_CREATED -> CloudEventDto.from(event, source, "io.apicurio.registry.events.GroupCreated");
```
paoloantinori: This `default` branch silently drops the 8 not-yet-mapped `StorageEventType` values (e.g. `ARTIFACT_VERSION_CREATED`). It's documented as follow-up, but once all types are mapped consider making the switch exhaustive -- removing the null-yielding `default` so adding a new `StorageEventType` is a compile error rather than a silent runtime drop in webhook delivery.

app/src/main/java/io/apicurio/registry/webhook/WebhookDeliveryService.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final Logger log = LoggerFactory.getLogger(WebhookDeliveryService.class);
+
+    private static final String CATEGORY_WEBHOOK = "Webhook";
+
+    @ConfigProperty(name = "apicurio.events.cloud-events-source", defaultValue = "/apicurio-registry")
+    @Info(category = CATEGORY_WEBHOOK, description = "CloudEvents source URI for webhook delivery", availableSince = "3.0.0")
+    String eventSource;
+
+    @ConfigProperty(name = "apicurio.events.webhook-delivery-enabled", defaultValue = "false")
+    @Info(category = CATEGORY_WEBHOOK, description = "Enable webhook delivery service (disabled until delivery implementation is complete)", availableSince = "3.0.0")
```
paoloantinori: `webhookDeliveryEnabled` is injected but never read in `onOutboxEvent` -- the bean is already gated by `@IfBuildProperty(name = "apicurio.events.webhook-delivery-enabled", stringValue = "true")`. If this field exists only to carry the `@Info` config-metadata annotation, that's fine and worth a one-line comment saying so; otherwise it can be removed as dead config.

EricWittmann (commented):
## Code Review

Thanks @karthik120710 for the work on this. The CloudEvents DTO design is reasonable and the `normalizeData` fix for JSONObject serialization is a genuine catch. However, there are several issues that need to be addressed before this can merge.

### Blockers

**1. Wrong conditional bean annotation: `@IfBuildProperty` instead of `@LookupIfProperty`**
`WebhookDeliveryService` uses `@IfBuildProperty`, which bakes the decision into the compiled artifact — users would have to **rebuild** the application to enable webhooks. This project consistently uses `@LookupIfProperty` for conditional beans (20+ usages across storage implementations, zero uses of `@IfBuildProperty` in the codebase). Please switch to `@LookupIfProperty`.

**2. `@Info` category uses a raw string instead of `ConfigPropertyCategory` enum**
`CATEGORY_WEBHOOK = "Webhook"` is a raw string. The `@Info(category = ...)` field expects a value from the `ConfigPropertyCategory` enum. Using a raw string will cause these properties to appear in an "unknown" category in the generated config docs. Either add `CATEGORY_EVENTS("events")` to the enum, or use an existing category.

**3. Incomplete `StorageEventType` mapping**
The converter maps 8 of 17 event types but silently drops 8 real data-change events (`GROUP_DELETED`, `GROUP_METADATA_UPDATED`, `GROUP_RULE_CONFIGURED`, `ARTIFACT_VERSION_METADATA_UPDATED`, `ARTIFACT_VERSION_DELETED`, `CONTRACT_RULESET_CONFIGURED`, `CONTRACT_METADATA_UPDATED`, `CONTRACT_STATUS_CHANGED`). A webhook consumer receiving `ArtifactCreated` but never `ArtifactVersionDeleted` or `GroupDeleted` is a confusing, incomplete API. All data-change events should be mapped.

**4. Only observes `SqlOutboxEvent`, not `KafkaSqlOutboxEvent`**
Webhook delivery would only work with SQL storage, not KafkaSQL. Per project conventions, storage-touching features must work across variants.

### Should-fix

**5. Dead `webhookDeliveryEnabled` field**
The `boolean webhookDeliveryEnabled` field is injected via `@ConfigProperty` but never read anywhere. The bean's existence is already gated by the conditional annotation. Remove the dead field.

**6. `CloudEventDto` allows construction of spec-invalid CloudEvents**
The no-arg constructor creates objects with `id=null`, `source=null`, `type=null`. Combined with `@JsonInclude(NON_NULL)`, this produces JSON that silently omits required CloudEvents 1.0 fields. Add validation in the `from()` factory method.

**7. Config property naming**
`apicurio.events.cloud-events-source` mixes hyphens with dot hierarchy. A more conventional name per the project's config property rules would be `apicurio.events.cloudevents.source` or `apicurio.events.source`.

**8. `onOutboxEvent` is a no-op**
The service converts events and then discards them (only a DEBUG log). Consider not merging the service class until delivery is wired, or marking the feature as `@Info(experimental = true)`.

### Minor

- CloudEvent `type` strings use Java class name format (`io.apicurio.registry.events.ArtifactCreated`) — CloudEvents convention is reverse-DNS event naming (`io.apicurio.registry.artifact.created`)
- Triple `@JsonProperty` annotations on field + getter + setter are redundant — Jackson only needs it once
- Test coverage only exercises 2 of 8 mapped event types
- `WebhookDeliveryServiceTest.testOnOutboxEvent` uses `assertDoesNotThrow` only — passes vacuously

EricWittmann:
@karthik120710 That review is **mostly** generated - so if you disagree with any of the findings please push back!  @carlesarnal is on vacation for a couple weeks so I'm taking over his assigned PRs.  I don't always have the full context to do a good job.  :shaking_face:

karthik120710:
hi @EricWittmann i have remove webhookDeliveryEnabled and its test from the last commit, can u take a look on my pr.

karthik120710:
hi @EricWittmann can u take a look at my pr , is it okay to merge.

EricWittmann:
Thanks @karthik120710 — good foundation work. The DTO and converter are clean, the exhaustive switch expression is a nice safety net, and the normalizeData fix for JSONObject serialization is a genuine catch. Appreciate your patience through the review iterations. Merged!

Issue #8426, opened 2026-07-01
feat: CloudEvents webhook notifications for schema changes


carlesarnal:
## Summary

Add a webhook notification system that emits CloudEvents when registry state changes (artifact created, version published, rule violated, artifact deprecated).

## Scope

- CloudEvents-compliant event schema for key registry operations
- Webhook subscription API (register endpoints, choose event types, filter by group/artifact type)
- Delivery engine with at-least-once semantics and exponential backoff retry
- SQL database schema for subscriptions and delivery logs
- Integration tests (subscribe, trigger change, verify delivery, verify retry)

## Context

This is proposed as an [LFX Mentorship](https://mentorship.lfx.linuxfoundation.org/) project for 2026 Term 3 (Sep-Nov). Initial implementation targets the SQL storage variant. CloudEvents is a CNCF Graduated project.

hcthakur2004:
Hi @carlesarnal, I'm interested in preparing for the CNCF LFX Term 3 project around CloudEvents webhook notifications. I have been contributing to CNCF-adjacent data/registry reliability work and would like to start with a focused contribution in Apicurio Registry before applications open.\n\nI'm going through the REST/storage areas now. If there is a small starter task that would be useful preparation for this project, especially around SQL storage, REST API tests, or event/notification plumbing, I'd be happy to pick it up.

hcthakur2004:
I took an initial pass through the repository and found the existing SQL event/outbox pieces under pp/src/main/java/io/apicurio/registry/storage/impl/sql (SqlOutboxEvent, SqlEventsProcessor, SqlEventRepository). That looks like a good area for me to understand before the webhook work.\n\nI'll start by looking for a small, reviewable improvement around event persistence/tests or REST/storage documentation so I can get familiar with the project conventions without touching the full mentorship feature yet.

carlesarnal:
Broke this down into 8 subtasks covering the webhook notification implementation:

- [ ] #TBD - Define CloudEvents schema and event types
- [ ] #TBD - Design SQL database schema for webhook subscriptions
- [ ] #TBD - Implement webhook subscription management API
- [ ] #TBD - Integrate event emission into SQL storage operations
- [ ] #TBD - Build webhook delivery engine with retry logic
- [ ] #TBD - Add webhook delivery monitoring and management
- [ ] #TBD - Create integration tests for webhook notification flow
- [ ] #TBD - Document webhook notification system

[...]

PR #8909, opened 2026-07-26
feat(webhooks): add OpenAPI contract for CloudEvents webhook subscriptions

jhaayushkumar:
Relates to #8426

Adds the OpenAPI contract for the CloudEvents webhook notification system:

- `/webhooks` — GET (list), POST (create)
- `/webhooks/{subscriptionId}` — GET, DELETE
- `SubscriptionEventType` enum: ARTIFACT_CREATED, ARTIFACT_UPDATED, ARTIFACT_DELETED,
  ARTIFACT_DEPRECATED, VERSION_CREATED, VERSION_PUBLISHED, RULE_VIOLATED
- `Subscription` and `CreateSubscription` DTO schemas

Signed-off-by: jhaayushkumar <jhaayushkumar@gmail.com>

### DIFF
--- a/common/src/main/resources/META-INF/openapi.json
+++ b/common/src/main/resources/META-INF/openapi.json
@@ -6252,6 +6252,154 @@
           }
         }
       }
+    },
+    "/webhooks": {
+      "summary": "Collection of webhook subscriptions.",
+      "get": {
+        "tags": [
+          "Webhooks"
+        ],
+        "operationId": "listSubscriptions",
+        "summary": "List webhook subscriptions",
+        "description": "Returns a list of all currently registered webhook subscriptions.",
+        "responses": {
+          "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/components/schemas/Subscription"
+                  }
+                }
+              }
+            },
+            "description": "A list of webhook subscriptions."
+          },
+          "401": {
+            "$ref": "#/components/responses/Unauthorized"
+          },
+          "403": {
+            "$ref": "#/components/responses/Forbidden"
+          },
+          "500": {
+            "$ref": "#/components/responses/ServerError"
+          }
+        }
+      },
+      "post": {
+        "tags": [
+          "Webhooks"
+        ],
+        "operationId": "createSubscription",
+        "summary": "Create a webhook subscription",
+        "description": "Registers a new webhook endpoint to receive CloudEvents for the specified event types.",
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/CreateSubscription"
+              }
+            }
+          }
+        },
+        "responses": {
+          "201": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/Subscription"
+                }
+              }
+            },
+            "description": "The webhook subscription was successfully created."
+          },
+          "400": {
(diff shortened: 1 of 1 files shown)


Nachiket-Roy:
Hi @jhaayushkumar, thanks for the contribution! This scope is reserved for the CNCF LFX Mentorship Term 3 program (Sep–Nov) and isn't currently open for general contributions. The maintainer has asked early contributors working on related areas to hold off on PRs until the mentorship term officially opens.

jhaayushkumar:
@paoloantinori if possible can u please review this..

paoloantinori (commented):
Reviewing the OpenAPI contract for the CloudEvents webhook subscriptions. The shape is reasonable as a first cut, but a few things are worth nailing down before this lands — most of them around the security model that the contract implies.

- **SSRF surface on `url`.** `POST /webhooks` accepts an arbitrary `url` (the `CreateSubscription` schema) that the server will POST CloudEvents to. Without an egress policy (allow/deny lists, blocking link-local and cloud-metadata ranges such as `169.254.169.254`, internal cluster DNS), any authenticated user who can create a subscription gets a server-side request forge into the internal network. This is the most important thing to pin down alongside the contract, because once the implementation lands the URL is already trusted. At minimum, restrict the allowed scheme/host in the schema description and track egress control as a hard requirement for the implementation PR.
- **Authz on subscription management is undefined.** Create/delete return `403` (good), but the contract doesn't state which roles may manage subscriptions. Registering a delivery endpoint that receives registry data is privileged — it most likely should be admin-only (`sr-admin`) under the existing role model. Recommend stating the required role in the operation descriptions so it isn't left implicit.
- **No delivery integrity.** There's no signing-secret/HMAC field on `Subscription`/`CreateSubscription`, and nothing describes a `ce-signature`-style header on delivery, so receivers can't verify a delivered CloudEvent actually came from this registry. For a webhook system this is usually expected; either add a secret field now or explicitly scope verification out for v1.
- **Consider OpenAPI `callbacks`.** That's the idiomatic way to specify a webhook API — a `callbacks` section on `createSubscription` would formally describe the CloudEvents POST the registry makes to `url` (payload shape, headers), instead of leaving delivery implicit, and makes generators produce correct receiver stubs.
- **`url` should be `format: uri`.** Both `Subscription.url` and `CreateSubscription.url` are plain `type: string`; `format: uri` gives cheap client-side validation and matches intent.
- **Event-type enum vs. what the backend can emit.** `SubscriptionEventType` lists `RULE_VIOLATED`, `VERSION_PUBLISHED`, `ARTIFACT_DEPRECATED`, etc. Worth confirming each maps to an event the registry actually produces (the existing event/beacon subsystem), or the contract advertises subscriptions that can never fire. I can't verify this from the contract alone since there's no implementation in this PR.
- **Scope note.** @Nachiket-Roy's earlier comment indicates this feature is reserved for the CNCF LFX Mentorship Term 3 (Sep–Nov) and isn't open for general contributions. @jhaayushkumar — worth confirming with a maintainer that an outside contribution here is wanted before investing more, so the work isn't turned away after the fact. Leaving the call to the maintainers.

Note: this PR ships the contract only — no resource/storage implementation — so `/webhooks` returns 404 against current `main` until the backend lands. Assuming that's the intended incremental step.

paoloantinori:
@jhaayushkumar Posted a review above with the main concerns — the biggest being the SSRF/authz model the `url` field implies, which is worth defining now rather than at implementation time. Also flagging the LFX scope question from @Nachiket-Roy's earlier comment for the maintainers to settle.

PR #8912, opened 2026-07-27
fix(schema-util): remove unchecked generic varargs warning in unionOf

### LINKED ISSUE #8905: Unchecked generic varargs warning in AbstractCompatibilityChecker.unionOf
## Description

**Registry Version**: snapshot (build from `main`, `3.3.1-SNAPSHOT`)
**Persistence type**: N/A - this is a compile-time warning, independent of the storage backend

`AbstractCompatibilityChecker` has a private helper that takes a generic varargs argument:

```java
private Set<D> unionOf(Set<D>... from) {
```
Because `Set<D>` isn't reifiable, the compiler has to create a generic array for the varargs, and that produces an `unchecked` warning every time the method is called. It's not annotated with @SafeVarargs or @SuppressWarnings, so the warning shows up on a normal build. The two callers are the `FULL` and `FULL_TRANSITIVE` branches of `testCompatibility`, and both of them only ever pass exactly two sets, so the varargs isn't really buying anything here.

**File**: `schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AbstractCompatibilityChecker.java` (the `unionOf` method, plus its call sites in the `FULL` / `FULL_TRANSITIVE` cases).

Nothing is functionally broken, the union result is correct. It's a code-quality/warning-hygiene thing, so I've marked it low severity.

### Environment

Doesn't depend on the environment. Reproduces on any plain `javac` build of the `schema-util/common` module. Seen locally with JDK 17 (source level) on the current `main`.

### Steps to Reproduce

1. Build the module with unchecked warnings surfaced, e.g.

   `./mvnw clean compile -pl schema-util/common`

   (or compile the single file with `javac -Xlint:unchecked`)

2. Look at the compiler output for `AbstractCompatibilityChecker.java`.
3. You'll see an unchecked warning pointing at the `unionOf(...)` varargs call sites (the `FULL` and `FULL_TRANSITIVE` cases).

**Expected**: the module compiles without unchecked warnings from this method.
**Actual**: an unchecked warning is emitted for the generic varargs array creation at the `unionOf` call sites.

### Logs

```java
warning: [unchecked] unchecked generic array creation for varargs parameter of type Set<D>[]
    incompatibleDiffs = unionOf(
                               ^
```

rajkumar0932:
## What

`AbstractCompatibilityChecker.unionOf` took a generic varargs parameter (`Set<D>... from`). Because `Set<D>` isn't reifiable, the compiler creates a generic array for the varargs, which emits unchecked warnings on a normal build.

Both call sites — the `FULL` and `FULL_TRANSITIVE` branches of `testCompatibility` — only ever pass exactly two sets, so the varargs wasn't buying anything. This replaces it with two explicit parameters, which removes the generic array creation entirely rather than papering over it with `@SafeVarargs`.

Fixes #8905

## Why this shape

`@SafeVarargs` would silence the warnings (the method is `private`, so it's allowed), but it keeps an unnecessary array allocation on every call and leaves the API looking more general than it is. Since both callers pass a fixed pair, a two-argument signature is the simpler and more honest fix — and it needs no suppression annotation.

## Before / After

Compiling the file with `-Xlint:unchecked` on `main` produces three warnings:

```
AbstractCompatibilityChecker.java:52: warning: [unchecked] unchecked generic array creation
                                     for varargs parameter of type Set<D>[]      (FULL)
AbstractCompatibilityChecker.java:59: warning: [unchecked] unchecked generic array creation
                                     for varargs parameter of type Set<D>[]      (FULL_TRANSITIVE)
AbstractCompatibilityChecker.java:98: warning: [unchecked] Possible heap pollution from
                                     parameterized vararg type Set<D>            (declaration)
```

After the change, the same compile is clean — no warnings from this file.

(The issue mentions the two call-site warnings; the declaration also emits a heap-pollution warning, so this clears all three.)

## Testing

- Compiled the file with `javac -Xlint:unchecked` before and after: 3 warnings → 0.
- Verified the union semantics are unchanged with a side-by-side comparison of the old varargs
  implementation and the new two-argument one across disjoint, overlapping, identical,
  first-empty, second-empty and both-empty set pairs — results identical in every case, and
  neither input set is mutated (the new version copies `first` rather than adding into it).
- No behavioural change is intended or expected: `new HashSet<>(first)` + `addAll(second)` is
  equivalent to the previous empty-set-plus-two-`addAll` loop.

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AbstractCompatibilityChecker.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AbstractCompatibilityChecker.java
@@ -95,11 +95,9 @@ public abstract class AbstractCompatibilityChecker<D> implements CompatibilityCh
 
     protected abstract CompatibilityDifference transform(D original);
 
-    private Set<D> unionOf(Set<D>... from) {
-        Set<D> rval = new HashSet<>();
-        for (Set<D> f : from) {
-            rval.addAll(f);
-        }
+    private Set<D> unionOf(Set<D> first, Set<D> second) {
+        Set<D> rval = new HashSet<>(first);
+        rval.addAll(second);
         return rval;
     }
 


rajkumar0932:
Closing this for now to stay within the one-open-PR-per-contributor limit — I'd rather use the slot for #8913, a user-visible bug in the prompt template preview where handlebars closing tags (`{{/if}}`, `{{/each}}`) never get highlighted.

This change is unaffected and still applies cleanly; the branch stays up, so I'll reopen it once the other PR is resolved. Happy to reopen sooner if you'd prefer this one first.

PR #8917, opened 2026-07-27
fix(ui): highlight handlebars closing tags in the prompt template preview

### LINKED ISSUE #8913: UI: handlebars closing tags ({{/if}}, {{/each}}) are not highlighted in the prompt template preview
## Description

**Registry Version**: snapshot (build from `main`, `23159df`)
**Persistence type**: N/A — this is a pure UI rendering issue

In the Documentation tab's template preview, handlebars **closing** tags are not highlighted. `{{#if detailed}}` renders as a styled block tag, but the matching `{{/if}}` renders as plain text, so a conditional block looks like it opens and never closes.

The same applies to `{{/each}}`, `{{/unless}}` and `{{/with}}`.

### Root cause

`ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx`, in `highlightVariables()`:

```ts
// Match both {{variable}} and {{#if variable}} / {{/if}} handlebars syntax
const regex = /\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g;
```

The comment says `{{/if}}` is meant to match, but the pattern can't match it. The block-prefix group requires trailing whitespace (`\s+`) and is followed by a mandatory variable name (`(\w+)`). A closing tag has neither:

- With the prefix group: `\/` matches `/`, `(?:if|…)` matches `if`, then `\s+` needs at least one space — but the next character is `}`, so the group fails.
- Without the prefix group: `(\w+)` would have to match starting at `/`, which isn't a word character.

So no alternative matches, and the tag falls through to the plain-text branch. Opening tags are unaffected because they do have a space and a name.

Both CSS classes already exist (`.template-variable`, `.template-block` in `PromptTemplateViewer.css`), so the closing tag is simply never given one.

### Steps to Reproduce

1. Create a `PROMPT_TEMPLATE` artifact whose `template` uses a conditional block, e.g.

   ```
   Summarize {{topic}}.
   {{#if detailed}}Be thorough about {{topic}}.{{/if}}
   ```

2. Open the artifact version's **Documentation** tab.
3. Look at the **Template** preview section.

**Expected**: `{{/if}}` is highlighted as a block tag, the same as `{{#if detailed}}`.

**Actual**: `{{#if detailed}}` is highlighted, `{{/if}}` is rendered as ordinary text.

Running the current regex over that template shows which tags get spans:

```
tags present   : {{topic}} {{#if detailed}} {{topic}} {{/if}}
highlighted    : {{topic}} {{#if detailed}} {{topic}}
NOT highlighted: {{/if}}
```

### Suggested fix

Give closing tags their own alternative, since they carry no variable name:

```ts
const regex = /\{\{(#(?:if|unless|each|with)\s+\w+|\/(?:if|unless|each|with)|\w+)\}\}/g;
```

and derive `isBlock` from the matched text rather than from the presence of the prefix group.

Happy to open a PR for this — I have the change and a before/after comparison ready.

rajkumar0932:
## What

In the Documentation tab's template preview, handlebars **closing** tags were rendered as plain text. `{{#if detailed}}` got a block highlight but the matching `{{/if}}` did not, so a conditional block looked like it opened and never closed. Same for `{{/each}}`, `{{/unless}}` and `{{/with}}`.

## Why it happened

`highlightVariables()` used:

```ts
// Match both {{variable}} and {{#if variable}} / {{/if}} handlebars syntax
const regex = /\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g;
```

The comment states `{{/if}}` should match, but the pattern cannot match it: the block-prefix group ends in `\s+` and is followed by a mandatory `(\w+)`, and a closing tag has neither a space nor a name. Without the prefix group, `(\w+)` would have to start on `/`, which isn't a word character — so no alternative matches and the tag falls through to plain text. Opening tags were fine because they do have a space and a name.

## The change

Closing tags get their own alternative, since they carry no variable name:

```ts
const regex = /\{\{(#(?:if|unless|each|with)\s+\w+|\/(?:if|unless|each|with)|\w+)\}\}/g;
```

`isBlock` is now derived from the matched text (`#`/`/` prefix) instead of from the presence of the old prefix group. The second capture group was unused, so collapsing to a single group is safe.

No CSS change is needed — `.template-block` already exists in `PromptTemplateViewer.css`; closing tags simply never received the class.

Fixes #8913

## Before / After

For this template:

```
Summarize {{topic}}.
{{#if detailed}}Be thorough about {{topic}}.{{/if}}
Done.
```

the emitted parts change from

```
"Summarize "  [VAR {{topic}}]  ".\n"  [BLOCK {{#if detailed}}]  "Be thorough about "  [VAR {{topic}}]  ".{{/if}}\nDone."
                                                                                                        ^^^^^^^^ swallowed into plain text
```

to

```
"Summarize "  [VAR {{topic}}]  ".\n"  [BLOCK {{#if detailed}}]  "Be thorough about "  [VAR {{topic}}]  "."  [BLOCK {{/if}}]  "\nDone."
```

## Testing

`ui-app` has no unit-test harness (no `test` script), so I verified by porting `highlightVariables()` verbatim and running it with the old and new patterns side by side:

- `{{/if}}`, `{{/each}}`, `{{/unless}}`, `{{/with}}` now match and are classed `template-block`; previously none of them matched.
- Every span the old pattern produced is still produced, with the same CSS class — the new output is a strict superset, so plain variables and opening block tags are unaffected.
- A bare `{{if}}` (a variable that happens to be named after a keyword) is still treated as a plain variable, not a block.
- `{{}}` still doesn't match.
- The concatenation of all emitted parts reproduces the input template exactly, so no text is dropped or duplicated.
- `npm run build` (tsc + vite) and `npm run lint` pass.

Whitespace-padded tags such as `{{ topic }}` and dotted paths like `{{user.name}}` are still unmatched — that is pre-existing behaviour and out of scope here; happy to open a separate issue if you'd like those handled too.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
@@ -65,8 +65,9 @@ export type PromptTemplateViewerProps = {
 
 const highlightVariables = (template: string): React.ReactNode[] => {
     const parts: React.ReactNode[] = [];
-    // Match both {{variable}} and {{#if variable}} / {{/if}} handlebars syntax
-    const regex = /\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g;
+    // Match both {{variable}} and {{#if variable}} / {{/if}} handlebars syntax.
+    // Closing tags carry no variable name, so they need their own alternative.
+    const regex = /\{\{(#(?:if|unless|each|with)\s+\w+|\/(?:if|unless|each|with)|\w+)\}\}/g;
     let lastIndex = 0;
     let match;
     let key = 0;
@@ -75,7 +76,7 @@ const highlightVariables = (template: string): React.ReactNode[] => {
         if (match.index > lastIndex) {
             parts.push(template.substring(lastIndex, match.index));
         }
-        const isBlock = !!match[1];
+        const isBlock = match[1].startsWith("#") || match[1].startsWith("/");
         parts.push(
             <span key={key++} className={isBlock ? "template-block" : "template-variable"}>
                 {match[0]}


schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AbstractCompatibilityChecker.java
```diff
@@ -95,11 +95,9 @@ protected abstract Set<D> isBackwardsCompatibleWith(String existing, String prop
 
     protected abstract CompatibilityDifference transform(D original);
 
-    private Set<D> unionOf(Set<D>... from) {
-        Set<D> rval = new HashSet<>();
-        for (Set<D> f : from) {
-            rval.addAll(f);
-        }
+    private Set<D> unionOf(Set<D> first, Set<D> second) {
```
paoloantinori: This `unionOf` change is the same fix as the just-closed #8912 (issue #8905) and is unrelated to the PR's stated purpose (UI handlebars highlighting, #8913). Consider splitting it into its own PR so this one stays focused on the UI fix and the commit history stays bisectable. The change itself looks fine — just out of place here.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
```diff
@@ -65,8 +65,9 @@ export type PromptTemplateViewerProps = {
 
 const highlightVariables = (template: string): React.ReactNode[] => {
     const parts: React.ReactNode[] = [];
-    // Match both {{variable}} and {{#if variable}} / {{/if}} handlebars syntax
-    const regex = /\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g;
+    // Match both {{variable}} and {{#if variable}} / {{/if}} handlebars syntax.
+    // Closing tags carry no variable name, so they need their own alternative.
+    const regex = /\{\{(#(?:if|unless|each|with)\s+\w+|\/(?:if|unless|each|with)|\w+)\}\}/g;
```
paoloantinori, line 70: Nit, not a blocker: the closing-tag alternative `\/(?:if|unless|each|with)` won't tolerate trailing whitespace inside the braces (e.g. `{{/if }}`). Handlebars allows it, though it's rare in practice and the old code didn't handle it either — so no regression. Flagging only in case you want closing/opening to be symmetric on whitespace.

paoloantinori (commented):
The regex rework in `PromptTemplateViewer.tsx` is correct and well-diagnosed — the old pattern's `\s+(\w+)` tail made it structurally impossible to match a nameless closing tag, and the new three-way alternation (`#block \w+` | `/close` | `\w+`) covers each case cleanly. I traced `isBlock` and the capture-group liveness; both behave consistently with the old code for variables and opening tags, and now correctly classify closing tags as blocks.

One process concern that I think is worth resolving before merge:

- **Two unrelated fixes in one PR.** The diff also contains the `AbstractCompatibilityChecker.unionOf` varargs change (issue #8905, previously in the just-closed #8912), but the title, body, and linked issue (#8913) only describe the UI highlighting fix. These are independent concerns in different modules (`schema-util` vs `ui`); splitting them keeps history bisectable and the PR descriptions accurate. Worth double-checking whether the Java change was left in the branch accidentally when switching the slot from #8912 to this PR — if it was intentional, a one-line note in the description would help reviewers.

rajkumar0932:
Thanks for the careful review.

**On the two unrelated fixes** — you're right, and your guess about the cause is correct: the `unionOf` commit was left in accidentally. I branched this fix off the `#8905` branch instead of `main` while moving the slot over from the closed #8912, so it came along for the ride. Entirely my mistake, and the PR description was wrong as a result.

Rebased onto `main` and dropped it — this PR is now a single commit touching only `PromptTemplateViewer.tsx`. The `unionOf` change is untouched on its own branch and I'll reopen it as a separate PR once this one is resolved (holding off only because of the one-open-PR-per-contributor limit).

**On the trailing-whitespace nit** — agreed it's worth being explicit about. I checked all three forms against both patterns:

| input | old | new |
|---|---|---|
| `{{/if}}` | no match | match |
| `{{/if }}` | no match | no match |
| `{{#if x}}` | match | match |
| `{{#if x }}` | no match | no match |
| `{{ topic }}` | no match | no match |

So padding is unhandled uniformly — opening tags, closing tags and plain variables all reject it, before and after. Opening and closing are already symmetric in that respect, and nothing regresses here.

Handling it properly means allowing `\s*` inside the braces for all three forms, which also pulls in `{{ topic }}`. That felt like a separate behavioural change rather than part of this fix, so I left it out to keep the diff focused — happy to open a follow-up issue and PR for it if you'd like it supported.

paoloantinori:
Thanks for the explanation @rajkumar0932 — that lines up with the branching history, and rebasing onto `main` to drop the stray `#8905` commit was the right cleanup. The single-file diff here (`PromptTemplateViewer.tsx` only) looks correct now, and the regex change itself is solid. No further concerns from my side.

rajkumar0932:
@carlesarnal thanks for picking this up — I think the bot is expecting `/accept` rather than `/accepted` (that's the spelling in its command list at the top of the thread), which would explain why `/auto-merge` still reports the PR as not accepted. The labels are also still `lifecycle/new` / `lifecycle/waiting-on-maintainer`, and the `Verify` jobs are all skipped with only the Verification Gate reporting, so that failure looks downstream of the same thing rather than anything in the diff.

Two things I can do on my side whenever it's useful:

- The branch is flagged as out-of-date with `main`. I've held off rebasing so I don't disturb the merge you're setting up, but I can update it in a moment if you'd like it current first.
- Once this one is resolved I'll open the `unionOf` varargs fix (#8905) as its own PR, as @paoloantinori suggested — just waiting on the one-open-PR limit.

paoloantinori:
Thanks @rajkumar0932 — agreed on the `/accept` vs `/accepted` spelling point, that lines up with the command list at the top of the thread and explains why `/auto-merge` kept reporting the PR as not accepted. On the diff itself, the head (`9cbfa82`) is unchanged since my last read and still looks good: the three-way regex alternative correctly matches the nameless closing tags, and `isBlock` derivation from the `#`/`/` prefix is sound. Nothing further from my side on the technical content.

carlesarnal:
Reviewed the current state of the PR (single commit 9cbfa82, PromptTemplateViewer.tsx only). The regex fix is correct and the previous review concern (unrelated unionOf commit) has been addressed. DCO sign-off, conventional commit, SonarQube all pass. Capture group 1 is no longer optional so match[1].startsWith() on line 79 is safe. All inputs the old regex matched are still matched identically - closing tags are strictly additive. No concerns from my side. Looks good to merge.

rajkumar0932:
Thanks for the second pass @carlesarnal.

The PR still can't merge though — it's sitting on `lifecycle/new` + `lifecycle/waiting-on-maintainer`, and the bot's last response is still "Cannot enable auto-merge: PR must be accepted first". I believe the acceptance step just needs `/accept` (the earlier command was `/accepted`, which the bot doesn't recognise); once that lands the lifecycle should progress and the `Verify` jobs — currently all skipped, with only the Verification Gate reporting — should actually run.

Also happy to rebase onto `main` to clear the out-of-date flag whenever you want it; I've left the head at `9cbfa82` so far so it matches what you and @paoloantinori both reviewed.

paoloantinori:
Thanks for the clear write-up @rajkumar0932 — you've diagnosed the lifecycle blocker exactly right: the bot's command list uses `/accept`, and the earlier `/accepted` (and `/auto-merge` attempts that followed) won't progress the PR out of `lifecycle/new`. So the regex fix is code-ready but parked on a lifecycle step.

`/accept` is a maintainer-side action I can't run from here, so I can't unblock it directly. Flagging it for @EricWittmann or @carlesarnal — once one of them runs `/accept`, the Verify jobs should kick in and auto-merge can proceed. Nothing further needed on the code side from my review.

EricWittmann:
Thanks @rajkumar0932 — clean regex fix with an excellent PR description (the before/after trace and input table were really helpful). Sorry about the lifecycle snafu that kept this stuck for a week. Merged!

PR #8919, opened 2026-07-27
fix(kafkasql): journal contract audit entries so they replicate across nodes

### LINKED ISSUE #8878: bug: kafkasql contract audit log is written only to the local node - diverges across the cluster and is lost on restart
## Description

**Registry Version**: 3.3.0

In a multi-node KafkaSQL deployment, contract audit entries created through one Registry node are only visible on that node. The other nodes do not return the entries, and they disappear after the node that handled the write is restarted. Querying the contract audit log therefore returns different results depending on which node answers, and entries are silently lost across restarts.

I reproduced this on a live two-node KafkaSQL deployment. As a control, a separately created artifact replicated correctly to the other node, confirming the cluster itself was replicating as expected. During the investigation I also observed that the audit entry does not appear to be propagated through the KafkaSQL journal.

This appears to be the same class of journaling gap that affected the global contract ruleset (#8848); the audit path was noted there and left for a separate issue.

## Steps to Reproduce

1. Start Kafka and two Registry nodes (A and B) using KafkaSQL storage against the same broker.
2. Create an artifact.
3. Update its contract metadata (or change its contract status) through node A.
4. Request the contract audit log from node A — the entry is present.
5. Request the same contract audit log from node B — the entry is missing.
6. Restart node A.
7. Request the contract audit log from node A again — the entry is gone.

## Expected Behaviour

A contract audit entry recorded through any node is visible on every node in the cluster and survives a node restart, so the audit log is a consistent, durable record.

## Actual Behaviour

The audit entry exists only in the serving node's local store. Other nodes never see it, and it is lost when that node restarts.

## Logs / Evidence

After the sequence above, the `kafkasql-journal` topic contains the message for the metadata change itself (the label merge), but no message corresponding to the audit-entry insert — consistent with the write bypassing the journal.

## Additional Notes

The root cause appears to be that `insertContractAuditEntry` in `KafkaSqlRegistryStorage` writes directly to the node-local SQL store, whereas most other write operations are submitted to the journal and applied on every node. There is currently no KafkaSQL journal message for the audit insert, and no kafkasql test coverage for the audit path.

If the audit log is intended to be a consistent, durable record (which its REST API and SQL-backed storage suggest), one possible approach is to route `insertContractAuditEntry` through the existing KafkaSQL message/journal mechanism, consistent with the contract ruleset operations. If it is instead meant to be node-local by design, that would be worth documenting. Happy to work on the
fix once the intended behavior is confirmed.

Anuragkumar-687:
## Summary

Route `insertContractAuditEntry` in the KafkaSQL storage implementation through the KafkaSQL journal instead of writing directly to the node-local SQL store.

Fixes #8878.

## Root cause

Contract audit entries are recorded when a contract's metadata is updated or its status is changed. On the KafkaSQL variant, `insertContractAuditEntry` in `KafkaSqlRegistryStorage` wrote directly to the node-local SQL store rather than submitting a message to the journal, unlike the other contract operations.

Since each KafkaSQL node derives its state from the journal, these writes were neither replicated to other nodes nor restored on restart. As a result, the contract audit log returned different entries depending on which node served the request, and entries were lost when the node that handled the write was restarted.

## Implementation

- Add `InsertContractAuditEntry1Message`, following the existing KafkaSQL message pattern.
- Route `insertContractAuditEntry` through the KafkaSQL submitter/coordinator workflow.
- Register the new message type in `KafkaSqlMessageIndex`.

## Testing

Added `KafkaSqlContractAuditJournalTest`, which verifies that inserting a contract audit entry produces the expected `InsertContractAuditEntry1Message` on the KafkaSQL journal topic.

Verified locally:

```bash
./mvnw test -pl app -am \
  -Dtest=KafkaSqlContractAuditJournalTest \
  -Dsurefire.failIfNoSpecifiedTests=false

```

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
@@ -748,7 +748,9 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
     @Override
     public void insertContractAuditEntry(io.apicurio.registry.storage.dto.ContractAuditEntryDto entry)
             throws RegistryStorageException {
-        sqlStore.insertContractAuditEntry(entry);
+        var message = new InsertContractAuditEntry1Message(entry);
+        var uuid = blockOnResult(submitter.submitMessage(message));
+        coordinator.waitForResponse(uuid);
     }
 
     @Override
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/messages/InsertContractAuditEntry1Message.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/messages/InsertContractAuditEntry1Message.java
@@ -1,0 +1,33 @@
+package io.apicurio.registry.storage.impl.kafkasql.messages;
+
+import io.apicurio.registry.storage.RegistryStorage;
+import io.apicurio.registry.storage.dto.ContractAuditEntryDto;
+import io.apicurio.registry.storage.impl.kafkasql.AbstractMessage;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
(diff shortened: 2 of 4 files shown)


EricWittmann:
Thanks @Anuragkumar-687! Clean implementation that follows the existing KafkaSQL patterns well. Merging now.

Issue #8728, opened 2026-07-18
fix(app): prompt template render endpoint leaves {{#if}} conditional blocks unrendered


AbdulWasih05:
## Description

**Registry Version**: snapshot (main)
**Persistence type**: in-memory

The v3 render endpoint (`POST /groups/{groupId}/artifacts/{artifactId}/versions/{version}/render`) performs plain `{{variable}}` substitution only. Handlebars-style conditional blocks (`{{#if var}} ... {{/if}}`) are not processed: the block markers don't match any provided variable name, so `substituteVariables()` keeps them verbatim in the output (`app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java`, `VARIABLE_PATTERN` / `substituteVariables`).

The MCP server renders the same templates differently: `PromptTemplateConverter.renderTemplate()` calls `processConditionalBlocks()`, which supports `{{#if}} ... {{/if}}` with truthiness semantics (non-null, non-`false`, non-empty-string), see `mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java`. So the same registered `PROMPT_TEMPLATE` artifact produces different renderings depending on whether it is rendered through the REST API (and the UI "Test Prompt" panel, which uses it) or through the MCP server.

This affects the project's own sample templates, which use conditional blocks (`examples/llm-artifact-types/sample-schemas/apicurio-support-chat-prompt.yaml`, `apicurio-support-system-prompt.yaml`, `mcp-enabled-code-review-prompt.yaml`), and the UI viewer explicitly highlights `{{#if}}` / `{{/if}}` as supported syntax (`highlightVariables()` in `ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx`), so to a user conditionals look supported but don't render.

Found while reading the prompt-template rendering code; file references above.

### Environment

Local build from `main`, JVM mode.

### Steps to Reproduce

1. Register `examples/llm-artifact-types/sample-schemas/apicurio-support-chat-prompt.yaml` as a `PROMPT_TEMPLATE` artifact.
2. Open the version's Documentation tab in the UI and click Render in the Test Prompt panel (or `POST` to the `/render` endpoint directly), with or without a value for `conversation_history`.
3. Inspect the rendered output.

### Expected vs Actual Behaviour

[...]

PR #8921, opened 2026-07-27
fix(schema-util): parse {{variable}} with one canonical pattern

### LINKED ISSUE #8877: bug: prompt template {{variable}} parsing is inconsistent across validator, render endpoint, and MCP converter
## Description

**Persistence type**: any (the affected code is in `schema-util` and `app` and is independent of the persistence type)

While working through the prompt template code I noticed that the `{{variable}}` placeholder syntax is parsed by three separate components, and each one uses its own regex, so they don't agree on what counts as a variable name.

| Component | Location | Pattern |
|---|---|---|
| `PromptTemplateContentValidator` | `schema-util/common/.../rules/validity/PromptTemplateContentValidator.java` | `\{\{(\w+)\}\}` |
| `PromptTemplateConverter` (MCP) | `mcp/.../PromptTemplateConverter.java` | `\{\{(\w+)\}\}` |
| `PromptRenderingService` (REST `/render`) | `app/.../services/PromptRenderingService.java` | `\{\{([^}]+)\}\}` then `.trim()` |

The validator and the MCP converter only accept word characters (`\w` = `[A-Za-z0-9_]`), while the REST render endpoint accepts anything except `}` and then trims. So `{{ name }}` (with surrounding spaces), `{{user.name}}`, and `{{first-name}}` are treated as variables by the render endpoint but are invisible to the validator.

Where this actually bites is the validator's "template variable used but not defined in `variables`" rule (`validateTemplateVariables`). Because its regex can't see a space-padded or dotted name, that rule silently never fires for those forms, so a template that references an undefined variable can still pass validation, as long as the placeholder has a space in it.

### Environment

Not environment specific. The validator lives in `schema-util/common` and the render endpoint in `app`; the inconsistency is reachable from the normal artifact validation and `/render` paths.

### Steps to Reproduce

1. Create a `PROMPT_TEMPLATE` artifact whose `template` is `"Hello {{ quality }}"` (note the spaces inside the braces) and whose `variables` object does **not** define `quality`.
2. Validate it at `FULL` validity level.
3. It passes. The same template written without spaces, `"Hello {{quality}}"`, is correctly rejected as an undefined variable (this is already covered by `testUndefinedVariable` in `PromptTemplateContentValidatorTest`).

**Same template, three paths.** Given `{"template": "Hello {{ name }}", "variables": {"name": {"type": "string"}}}`, the single placeholder `{{ name }}` is interpreted three different ways:

- **Validator** (`\{\{(\w+)\}\}`): extracts nothing, so the "used but not defined" check never runs on it.
- **REST `/render`** (`\{\{([^}]+)\}\}` then `.trim()`): extracts `name` and substitutes it.
- **MCP converter** (`\{\{(\w+)\}\}`): extracts nothing, so `{{ name }}` is left in the output as literal text.

Same input, three different answers to "what is the variable here".

Expected: whether or not there is whitespace inside the braces, a `{{ variable }}` that is used but not defined is flagged consistently, and the three components agree on what a variable is.

Actual: `{{ quality }}` passes validation because the validator's `\w+` regex does not match it. At render time the REST `/render` endpoint does recognise it, finds no value for it, and leaves the literal `{{ quality }}` in the rendered output, so a broken prompt is produced silently. The same template also renders differently through the MCP path, which uses the stricter `\w+` pattern.

### Logs

Nothing surfaces by default: the validator simply does not raise a violation, and the render endpoint returns a normal response with the unresolved placeholder left in place. It can be confirmed directly: `\{\{(\w+)\}\}` returns no match for `{{ name }}`, while `\{\{([^}]+)\}\}` (trimmed) returns `name`.

### Additional context

This looks related to, but separate from, a couple of open issues in the same area:

- #8728 is about the render endpoint not processing `{{#if}} ... {{/if}}` conditional blocks (a different facet of the "each component parses `{{}}` its own way" theme).
- #8669 is about a null `template` field in the render endpoint.

Neither touches the validator's variable-name grammar, so this seems to be a distinct issue.

On the fix, the cleanest direction seems to be converging on a single shared extraction helper (one agreed grammar, applied consistently, with trimming) used by the validator, the render endpoint, and the MCP converter, so they can't drift apart again. That raises a genuine question though: which grammar is intended? Are prompt template variable names meant to be `\w+` only (in which case the render endpoint is too lenient), or should whitespace and dotted names be supported everywhere (in which case the validator is too strict)? Happy to put up a PR in whichever direction you prefer.

PRAteek-singHWY:
## Summary

Fixes #8877.

The `{{variable}}` placeholder syntax of Prompt Template artifacts was parsed by three
separate components, each with its own regular expression. They disagreed, so the same
template could mean three different things depending on which path it went through.

The most visible consequence is a validation hole: the validity rule's "template variable
used but not defined in `variables`" check silently never fired for a placeholder written
with spaces inside the braces. A template referencing an undefined `{{ quality }}` passed
`FULL` validation, then rendered with the literal `{{ quality }}` still in the output.

This PR pins one canonical pattern in `schema-util/common` and routes all three components
through it, so they cannot drift apart again.

## Root Cause

Three components, three answers to "what is a variable here":

| Component | Pattern | `{{ name }}` resolves to |
|---|---|---|
| `PromptTemplateContentValidator` (`schema-util/common`) | `\{\{(\w+)\}\}` | nothing — not a variable |
| `PromptRenderingService` (`app`, REST `/render`) | `\{\{([^}]+)\}\}` then `.trim()` | `name` |
| `PromptTemplateConverter` (`mcp`) | see below | nothing — left as literal text |

The validator's `\w+` cannot match across the spaces, so `extractTemplateVariables` returns
an empty list for `"Hello {{ quality }}"`. `validateTemplateVariables` then has nothing to
cross-check against the `variables` object, and the rule passes. The same template without
spaces is correctly rejected — that case is already covered by the existing
`testUndefinedVariable`.

Worth flagging one thing I found while fixing the MCP side: `PromptTemplateConverter`
declared a `VARIABLE_PATTERN` field, but nothing ever used it. The actual substitution was
a loop over the supplied arguments doing

```java
String placeholder = "\\{\\{" + entry.getKey() + "\\}\\}";
rendered = rendered.replaceAll(placeholder, Matcher.quoteReplacement(value));
```

The observable behaviour matches what the issue describes — `{{ name }}` is not substituted
— but the mechanism is a literal `{{key}}` match, not the declared regex. It also builds a
regex out of the argument name, so a `.` in a key matched any character: an argument named
`a.b` would substitute an unrelated `{{axb}}` placeholder. Both go away with this change.

## The canonical grammar

Following the direction @paoloantinori gave on the issue thread: one pattern, in the shared
layer, used by everyone.

```java
Pattern.compile("\\{\\{\\s*(\\w+)\\s*\\}\\}");
```

A word-character name with optional surrounding whitespace, so `{{name}}` and `{{ name }}`
both resolve to `name`. Handlebars-style control blocks (`{{#if x}}`, `{{/if}}`) are not
matched, because `#` and `/` are not word characters — so this change does not disturb
conditional blocks, which are a separate concern (#8728).

## Changes

- **New** `schema-util/common/.../content/util/PromptTemplateVariableUtil.java`
  - `VARIABLE_PATTERN` — the single definition of the placeholder syntax.
  - `extractVariableNames(String)` — variable names in first-seen order, deduplicated,
    empty list for a null template.
  - `substituteVariables(String, Function<String, String>)` — replaces placeholders with
    whatever the resolver returns. A resolver returning `null` means "no value", and the
    placeholder is written back verbatim. That hook is what lets both call sites keep their
    own (differing) idea of what an unresolved variable means, while sharing the parsing.

- `schema-util/common/.../rules/validity/PromptTemplateContentValidator.java`
  - Dropped the local `TEMPLATE_VARIABLE_PATTERN`; `extractTemplateVariables` now delegates
    to the shared helper. The method stays `public static` — `PromptTemplateCompatibilityChecker`
    calls it.

- `app/.../services/PromptRenderingService.java`
  - Dropped the local `VARIABLE_PATTERN`; `substituteVariables` now delegates. Unresolved
    variables still keep their placeholder, exactly as before.

- `mcp/.../PromptTemplateConverter.java`
  - Removed the unused `VARIABLE_PATTERN` and replaced the per-argument `replaceAll` loop
    with a single pass through the shared helper. `{{#if}}` handling is untouched and still
    runs after substitution, as it did before.

- `mcp/pom.xml`
  - Added `apicurio-registry-schema-util-common`. See the note below.

## Behaviour changes worth reviewing

1. **`/render` no longer substitutes dotted or hyphenated names.** Tightening the endpoint
   from `[^}]+` to `\w+` is the part of this change that removes a capability: `{{user.name}}`
   and `{{first-name}}` were substituted by `/render` and are now literal text. That was
   already true for the validator and the MCP converter, so this makes `/render` consistent
   with the other two rather than the other way around — but it is the direction choice the
   issue asked about, so it deserves a look. Widening everything to accept dotted names
   instead is a one-line change to the shared pattern if you would rather go that way.

2. **The compatibility checker gets stricter, in the same direction as the fix.**
   `PromptTemplateCompatibilityChecker` uses `extractTemplateVariables` to detect
   `VARIABLE_REMOVED_BUT_USED`. It now also catches the case where the removed variable is
   still referenced as `{{ varName }}`, which it previously missed.

3. **A new module dependency: `mcp` → `schema-util-common`.** This is the one debatable bit.
   The MCP module previously depended only on the Java SDK, so sharing the pattern with it
   means adding a dependency. `schema-util-common` is fairly light (its heavier transitives
   in `apicurio-registry-common` are `provided`-scoped), and duplicating the constant would
   defeat the point of the issue — but if you would rather not grow the MCP server's
   classpath, say so and I will drop the MCP half into a follow-up.

## Test plan

Splitting these honestly into the ones that actually reproduce the bug and the ones that
are there to stop it coming back.

**Reproduce the bug — these fail on `main`:**

- `PromptTemplateContentValidatorTest.testUndefinedVariableWithWhitespace` — the case from
  the issue. `{{ quality }}` is used but not defined, and `FULL` validation now rejects it.
  On `main` the validator extracts nothing from that placeholder, raises no violation, and
  the `assertThrows` fails.
- `PromptTemplateContentValidatorTest.testExtractTemplateVariablesNormalisesWhitespace` —
  pins the exact list the extraction helper returns, since `PromptTemplateCompatibilityChecker`
  depends on it. On `main` the padded spellings are missing from the result.
- `PromptTemplateConverterTest.testRenderVariableWithWhitespace` — `{{ name }}` is now
  substituted by the MCP path instead of surviving into the output as literal text.
- `PromptTemplateConverterTest.testVariableNameWithRegexMetacharacterDoesNotMatchWildly` —
  an argument named `a.b` no longer substitutes an unrelated `{{axb}}` placeholder.
- `PromptRenderingServiceTest.testDottedNameIsNotAVariable` — this one fails on `main`
  because `main` is the more permissive side. It documents behaviour change (1) above.

**Guard the behaviour that must not change — these already pass on `main`:**

- `PromptRenderingServiceTest` — padded spellings resolving identically, unresolved padded
  placeholders preserved verbatim, `{{#if}}` markers left alone by substitution.
- `PromptTemplateContentValidatorTest.testDefinedVariableWithWhitespaceIsAccepted` and
  `testConditionalBlockMarkersAreNotVariables` — no new false positives, and control syntax
  is not mistaken for a variable.
- The existing `testWhitespaceInVariableName` and `testUndefinedVariable` still pass; the
  canonical pattern preserves everything that already worked on both sides.

**New test classes:**

- `PromptTemplateVariableUtilTest` (`schema-util/common`, 12 cases) — direct coverage of the
  shared helper: whitespace variants, first-seen ordering and deduplication, null template,
  control markers, `$` and `\` in substituted values, and the null-resolver contract.
- `PromptTemplateConverterTest` (`mcp`, 11 cases) — the converter had no test coverage at
  all. Covers substitution with and without whitespace, unknown variables, null values,
  null args, regex metacharacters in both values and argument names, and the existing
  `{{#if}}` truthy/falsy behaviour.

Checks run locally:

- [x] `./mvnw -pl schema-util/common,schema-util/util-provider test` — 32 tests, all pass
      (`PromptTemplateVariableUtilTest` 12, `PromptTemplateContentValidatorTest` 22,
      `PromptTemplateCompatibilityCheckerTest` 10, the last unchanged by this PR).
- [x] `./mvnw -Dfull -pl mcp test -Dtest=PromptTemplateConverterTest` — 11 tests, all pass.
      This also confirms the new `mcp` → `schema-util-common` dependency resolves.
- [x] `./mvnw checkstyle:check` clean on `schema-util/common`, `schema-util/util-provider`
      and `mcp`.
- [ ] `app` — `PromptRenderingServiceTest` is a `@QuarkusTest` and I could not get the `app`
      module to build to completion locally. Please let CI run it.
- [ ] Storage variants: not applicable. Nothing here touches `storage/impl/`; the change is
      confined to template-string parsing.

## Notes

- Related but separate: #8728 (`/render` does not process `{{#if}}` blocks) and #8669 (null
  `template` field). This PR deliberately leaves both alone — it only makes the three
  components agree on what a variable *is*.
- PR #8690 is open against `PromptRenderingService.java` and its test, but it changes
  `extractTemplateText` (null/non-string `template` field), not `substituteVariables`, so
  there is no functional overlap. Whichever merges second may need a trivial rebase in the
  test file.
- There is actually a fourth parser I did not touch:
  `ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx` highlights
  placeholders with `/\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g`, which is also
  whitespace-intolerant, so `{{ name }}` is not highlighted as a variable in the preview.
  I left it out of scope: it is a separate build system, it is presentation rather than the
  parser of record, and PR #8917 is already open against that component. Happy to follow up
  there once the direction here is settled.
- The class javadoc on `PromptTemplateConverter` calls it "the Java equivalent of
  mcp-converter.ts", but there is no `mcp-converter.ts` in the tree any more. Left as-is
  rather than widening the diff.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -2,6 +2,7 @@ package io.apicurio.registry.services;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import io.apicurio.registry.content.ContentHandle;
+import io.apicurio.registry.content.util.PromptTemplateVariableUtil;
 import io.apicurio.registry.rest.v3.beans.RenderPromptResponse;
 import io.apicurio.registry.rest.v3.beans.RenderValidationError;
 import io.apicurio.registry.storage.error.InvalidContentException;
@@ -11,8 +12,6 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import static io.apicurio.registry.util.JsonObjectMapper.MAPPER;
 import static io.apicurio.registry.util.YAMLObjectMapper.YAML_MAPPER;
@@ -22,8 +21,6 @@ import static io.apicurio.registry.util.YAMLObjectMapper.YAML_MAPPER;
  */
 @ApplicationScoped
 public class PromptRenderingService {
-
-    private static final Pattern VARIABLE_PATTERN = Pattern.compile("\\{\\{([^}]+)\\}\\}");
 
     /**
      * Renders a prompt template by substituting variables.
@@ -256,27 +253,11 @@ public class PromptRenderingService {
      * Substitute variables in the template text using {{variable}} syntax.
      */
     private String substituteVariables(String template, Map<String, Object> variables) {
-        StringBuffer result = new StringBuffer();
-        Matcher matcher = VARIABLE_PATTERN.matcher(template);
-
-        while (matcher.find()) {
-            String varName = matcher.group(1).trim();
+        return PromptTemplateVariableUtil.substituteVariables(template, varName -> {
             Object value = variables.get(varName);
-
-            String replacement;
-            if (value == null) {
-                // Keep the original placeholder if variable is not provided
-                replacement = matcher.group(0);
-            } else {
-                replacement = formatValue(value);
-            }
-
-            // Escape special characters for replacement
-            matcher.appendReplacement(result, Matcher.quoteReplacement(replacement));
-        }
-        matcher.appendTail(result);
-
-        return result.toString();
+            // Returning null keeps the original placeholder if the variable is not provided.
+            return value == null ? null : formatValue(value);
+        });
     }
 
     /**
--- a/app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
+++ b/app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
@@ -478,6 +478,76 @@ public class PromptRenderingServiceTest {
     }
 
     @Test
+    public void testWhitespaceInVariableNameWithPadding() {
+        // Both spellings of the same variable must resolve to the same value.
+        String yamlContent = """
+            templateId: whitespace
+            template: "{{name}} / {{ name }} / {{   name   }}"
+            """;
+
+        ContentHandle content = ContentHandle.create(yamlContent);
+        Map<String, Object> variables = Map.of("name", "World");
+
+        RenderPromptResponse response = renderingService.render(content, variables,
+                "default", "whitespace", "1.0");
+
+        Assertions.assertEquals("World / World / World", response.getRendered());
+    }
+
+    @Test
+    public void testMissingVariableWithWhitespacePreservedAsPlaceholder() {
+        String yamlContent = """
+            templateId: partial
+            template: "Hello, {{ name }}! Your role is {{ role }}."
+            """;
+
+        ContentHandle content = ContentHandle.create(yamlContent);
+        Map<String, Object> variables = Map.of("name", "Bob");
+
+        RenderPromptResponse response = renderingService.render(content, variables,
+                "default", "partial", "1.0");
+
+        // The unresolved placeholder is written back exactly as the author spelled it.
+        Assertions.assertEquals("Hello, Bob! Your role is {{ role }}.", response.getRendered());
+    }
+
+    @Test
+    public void testDottedNameIsNotAVariable() {
+        // Deliberate behaviour change: the canonical grammar is \\w+, so a dotted placeholder is
+        // literal text here, exactly as it already was for the validator and the MCP converter.
+        String yamlContent = """
+            templateId: dotted
+            template: "Hello, {{user.name}}!"
+            """;
+
+        ContentHandle content = ContentHandle.create(yamlContent);
+        Map<String, Object> variables = Map.of("user.name", "Alice");
+
+        RenderPromptResponse response = renderingService.render(content, variables,
+                "default", "dotted", "1.0");
+
+        Assertions.assertEquals("Hello, {{user.name}}!", response.getRendered());
+    }
+
+    @Test
+    public void testConditionalBlockMarkersAreNotSubstituted() {
+        String yamlContent = """
+            templateId: conditional
+            template: "{{#if premium}}Hello {{ name }}{{/if}}"
+            """;
+
+        ContentHandle content = ContentHandle.create(yamlContent);
+        Map<String, Object> variables = Map.of("name", "Alice", "premium", true);
+
+        RenderPromptResponse response = renderingService.render(content, variables,
+                "default", "conditional", "1.0");
+
+        // The render endpoint does not process {{#if}} blocks (see #8728); it must at least leave
+        // the markers alone rather than mangling them while substituting the variable inside.
+        Assertions.assertEquals("{{#if premium}}Hello Alice{{/if}}", response.getRendered());
+    }
+
+    @Test
     public void testMultilineTemplate() {
         String yamlContent = """
             templateId: multiline
--- a/mcp/pom.xml
+++ b/mcp/pom.xml
@@ -63,6 +63,15 @@
             <artifactId>apicurio-registry-java-sdk</artifactId>
         </dependency>
 
+        <!--
+        Provides PromptTemplateVariableUtil, the canonical {{variable}} parser shared with the
+        validity rule and the REST render endpoint, so the three do not drift apart again.
+        -->
+        <dependency>
+            <groupId>io.apicurio</groupId>
+            <artifactId>apicurio-registry-schema-util-common</artifactId>
+        </dependency>
+
         <dependency>
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
--- a/mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import io.apicurio.registry.content.util.PromptTemplateVariableUtil;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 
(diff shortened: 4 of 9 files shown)


PRAteek-singHWY:
Hi @paoloantinori @EricWittmann @carlesarnal, parking this here rather than reopening, since the lifecycle bot closes any second PR while #8841 is still open. The branch and the full diff are attached to this PR, so it's reviewable as it stands, and I'll reopen it the moment #8841 lands.

This is the `{{variable}}` parsing fix for #8877, following the direction Paolo gave on the #8841 thread: one canonical pattern pinned in the shared layer, with all three consumers routed through it so they can't drift apart again.

- New `PromptTemplateVariableUtil` in `schema-util/common` holding `{{\s*(\w+)\s*}}` plus extraction and substitution helpers.
- `PromptTemplateContentValidator`, `PromptRenderingService` (REST `/render`) and the MCP `PromptTemplateConverter` all delegate to it.
- The bug from the issue is closed: `{{ quality }}` used but not defined now fails `FULL` validation instead of passing silently and rendering with the placeholder left in the output.

Two decisions I'd rather surface now than after review starts, because either could change the shape of this:

1. **Tightening `/render` from `[^}]+` to `\w+` drops dotted-name support.** `{{user.name}}` was substituted by the render endpoint and is now literal text. That makes `/render` consistent with the validator and the MCP converter rather than the other way round, which is how I read the suggestion. Widening the shared pattern to admit dotted names everywhere is a one-line change if you'd prefer that direction instead.

2. **Sharing the pattern with the MCP converter adds a module dependency.** `mcp` previously depended only on the Java SDK and now also pulls `schema-util-common`. Duplicating the constant would defeat the point of the issue, but if you'd rather not grow the MCP server's classpath I can split the MCP half into a follow-up.

One incidental find: `PromptTemplateConverter` declared a `VARIABLE_PATTERN` that nothing ever used. The real substitution built a regex out of each argument name, so an argument called `a.b` would substitute an unrelated `{{axb}}` placeholder. That's fixed as a side effect here.

Tests: 12 new cases on the shared helper, 4 on the validator (including the whitespace repro from the issue, which fails on `main`), 4 on `/render`, and a new `PromptTemplateConverterTest` with 11 cases, since the MCP converter had no coverage at all before. `schema-util` and `mcp` pass locally. The `app` `@QuarkusTest` I've left for CI.

On #8841: if there's nothing further you'd like changed there, a `/merge` would unblock this one. Happy to keep iterating on either if not.

PR #8924, opened 2026-07-27
fix(ui): highlight handlebars closing tags in prompt template preview 

### LINKED ISSUE #8913: UI: handlebars closing tags ({{/if}}, {{/each}}) are not highlighted in the prompt template preview

Shubb07:
Fixes #8913
The regex in `highlightVariables` (PromptTemplateViewer.tsx) is meant to match both `{{variable}}` and handlebars block tags, per its own comment, but the pattern can't match closing tags like `{{/if}}` or `{{/each}}`: the block-prefix group requires trailing whitespace and a variable name, neither of which exists on a closing tag.
Restructured the regex into two alternatives: one for block open/close tags (variable name optional, so `{{/if}}` matches), one for plain variables. Behavior for opening tags and plain variables is unchanged.
Verified locally with a template containing `{{#if}}...{{/if}}` and `{{#each}}...{{/each}}` blocks. All four closing tag forms (`if`, `unless`, `each`, `with`) now get the `template-block` class.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
@@ -65,8 +65,10 @@ export type PromptTemplateViewerProps = {
 
 const highlightVariables = (template: string): React.ReactNode[] => {
     const parts: React.ReactNode[] = [];
-    // Match both {{variable}} and {{#if variable}} / {{/if}} handlebars syntax
-    const regex = /\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g;
+    // Match {{variable}} and handlebars block tags: {{#if var}}, {{/if}}, {{#each items}}, {{/each}}, etc.
+    // The first alternative catches block open/close tags (closing tags have no variable name).
+    // The second alternative catches plain variable references.
+    const regex = /\{\{([#/](?:if|unless|each|with))(?:\s+\w+)?\}\}|\{\{(\w+)\}\}/g;
     let lastIndex = 0;
     let match;
     let key = 0;


ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
```diff
@@ -65,8 +65,10 @@ export type PromptTemplateViewerProps = {
 
 const highlightVariables = (template: string): React.ReactNode[] => {
     const parts: React.ReactNode[] = [];
-    // Match both {{variable}} and {{#if variable}} / {{/if}} handlebars syntax
-    const regex = /\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g;
+    // Match {{variable}} and handlebars block tags: {{#if var}}, {{/if}}, {{#each items}}, {{/each}}, etc.
+    // The first alternative catches block open/close tags (closing tags have no variable name).
+    // The second alternative catches plain variable references.
+    const regex = /\{\{([#/](?:if|unless|each|with))(?:\s+\w+)?\}\}|\{\{(\w+)\}\}/g;
```
paoloantinori, line 71: Worth adding a small unit test for `highlightVariables` covering the four closing-tag forms this PR fixes (`{{/if}}`, `{{/unless}}`, `{{/each}}`, `{{/with}}`) plus a plain `{{var}}` and a `{{#each items}}`. The function returns React nodes, so asserting on the `className` of each span (`template-block` vs `template-variable`) is enough to lock in the block/variable distinction and catch future regressions — the old regex's bug (block prefix required `\s+\w+`, so closing tags silently fell through) is easy to reintroduce by accident.

Not a blocker for this PR — just flagging it as cheap insurance if you (or a maintainer) want to fold it in.

paoloantinori (commented):
Correct, well-scoped fix. Tracing the new two-alternative regex, the cases that previously failed (`{{/if}}`, `{{/each}}`, `{{/unless}}`, `{{/with}}` — all closing tags with no variable name) now match via the first alternative, where `(?:\s+\w+)?` makes the variable optional. Plain variables still route to the second alternative, and since `match[1]` is `undefined` there, `isBlock` correctly stays false. No previously-highlighted form regresses.

One suggestion below — regex helpers like this are exactly where a tiny unit test pays for itself, since the next contributor tweaking the pattern can't easily tell from the rendered UI whether closing tags still match.

Shubb07:
Thanks Paolo, agreed a unit test would be the right home for regressions like this. `ui-app` doesn't currently have a unit test runner set up (only the Playwright e2e suite in `ui/tests/`), so adding a test for `highlightVariables` would mean introducing Vitest or similar to the workspace. Happy to open a follow-up issue for that if it fits the project's direction, that way this PR stays focused on the bug fix and the test infrastructure gets its own scoped change.

paoloantinori:
Thanks @Shubb07 — appreciate you thinking it through. For a 4-line regex fix I wouldn't block on introducing Vitest to the workspace; that's a separate infrastructure decision that's bigger than this PR. The fix is correct and well-scoped as-is, and the local verification you did (`{{#if}}...{{/if}}`, `{{#each}}...{{/each}}` with all four closing tag forms) is sufficient evidence for a change this small.

If you do want to add unit test coverage later, a dedicated PR that brings in Vitest (or extends the Playwright suite) would be the right home for it — not this PR.

Nachiket-Roy:
this PR duplicates #8917, which cover the same fix.

paoloantinori:
Good catch, @Nachiket-Roy — #8917 does cover the same bug. Both PRs restructure `highlightVariables` so closing tags like `{{/if}}` match, and I'd reviewed #8917 already; it's been `/accept`ed by @carlesarnal and is sitting in the auto-merge queue.

Given #8917 is further along (accepted + waiting on the merge batch), the cleanest path is probably to defer this one and let #8917 land. @Shubb07 — the fix itself is correct and well-reasoned here too, so if for any reason #8917 stalls, this is a perfectly good fallback; no rework needed. cc @carlesarnal in case you want to close one of the two to avoid confusion.

Shubb07:
Closing as duplicate of #8917 which is already in the merge queue. Thanks @Nachiket-Roy for the catch and @paoloantinori for the review.

PR #8925, opened 2026-07-27
feat(ui): add syntax highlighting for GraphQL and Protobuf artifacts

### LINKED ISSUE #8865: UI: Add syntax highlighting for GraphQL and Protobuf artifacts
# UI: Add syntax highlighting for GraphQL and Protobuf artifacts

In Apicurio Registry `3.2`, GraphQL and Protobuf artifact content is displayed as plain text in the Registry UI.

Both artifact types are supported, and their content can be created, stored, and viewed successfully. However, the content editor/viewer does not apply syntax highlighting for these formats.

This makes larger GraphQL schemas and Protobuf definitions harder to read and review.

### Proposed Solution

Add syntax highlighting support for the following artifact types in the Registry UI:

* `GRAPHQL` artifacts using a GraphQL language mode
* `PROTOBUF` artifacts using a Protocol Buffers language mode

The UI should select the appropriate editor language based on the artifact type.

This would improve readability and provide a more consistent editing and viewing experience across supported artifact formats.

### Additional Context

The issue is reproducible with:

* Apicurio Registry: `3.2`
* Apicurio Registry UI: `3.2`

JSON and YAML-based artifacts are highlighted correctly, while GraphQL and Protobuf content is rendered as plain text.

Myself-Praveen:
Fixes #8865

### Description
This PR introduces syntax highlighting for `GRAPHQL` and `PROTOBUF` artifacts in the Registry UI content editor/viewer. 

Previously, these artifact types were rendered as plain text. This update configures Monaco Editor to properly register and apply language themes for both Protobuf and GraphQL formats, bringing them to parity with JSON and YAML artifacts and significantly improving readability for larger schemas.

### Changes
- Updated `TextEditor.tsx` to mount and register language modes.
- Added language mapping for `APPLICATION_GRAPHQL` and `APPLICATION_PROTOBUF` in `content.utils.ts`.
- Configured Monaco tokenizers and themes for Protobuf syntax.

### Authorship
- [x] I, a human, wrote this pull request description myself.


ui/ui-app/src/app/pages/version/components/tabs/ContentTabContent.tsx
```diff
@@ -6,13 +6,15 @@ import Editor from "@monaco-editor/react";
...
+
 
 const TYPE_MAP: any = {};
 TYPE_MAP[ContentTypes.APPLICATION_PROTOBUF] = "protobuf";
 TYPE_MAP[ContentTypes.APPLICATION_XML] = "xml";
 TYPE_MAP[ContentTypes.APPLICATION_JSON] = "json";
 TYPE_MAP[ContentTypes.APPLICATION_YAML] = "yaml";
-TYPE_MAP[ContentTypes.APPLICATION_GRAPHQL] = "graphqlschema";
+TYPE_MAP[ContentTypes.APPLICATION_GRAPHQL] = "graphql";
 TYPE_MAP[ContentTypes.APPLICATION_THRIFT] = "plaintext";
```
paoloantinori, line 18: Worth confirming this actually produces highlighting. The PR adds a Monarch tokenizer only for Protobuf; nothing here registers a `graphql` language with Monaco, and `monaco-editor` doesn't bundle GraphQL in its `basic-languages` set. If no GraphQL contribution is loaded elsewhere in the app, a GraphQL artifact will still render as plain text and the GraphQL half of the PR's title won't be delivered. Either add a GraphQL tokenizer here (mirroring `registerProto`) or pull in a `monaco-graphql` contribution.

ui/ui-app/src/editors/ProtoEditor.tsx
```diff
@@ -22,6 +22,85 @@ const protoBufThemeData = {
...
+                [/[{}()[\]]/, "@brackets"],
+                [/[<>](?!@symbols)/, "@brackets"],
+                [/@symbols/, {
+                    cases: {
+                        "@operators": "operator",
+                        "@default": ""
+                    }
+                }],
+                // @ annotations.
+                [/@\s*[a-zA-Z_$][\w$]*/, { token: "annotation", log: "annotation token: $0" }],
```
paoloantinori: The `log: "annotation token: $0"` writes to the browser console for every annotation token in every Protobuf editor. Now that `registerProto` is wired into `ContentTabContent`, `TextEditor`, and `DiffView`, this console output surfaces in production across more views. Consider removing the `log:` entry unless it's still needed for debugging.

ui/ui-app/src/app/components/common/SortOrderToggle.tsx
```diff
@@ -18,7 +18,7 @@ export const SortOrderToggle: FunctionComponent<SortOrderToggleProps> = (props:
     );
 
     return (
-        <Button icon={icon} variant="plain" aria-label={props.sortOrder === SortOrder.asc ? "sort ascending" : "sort descending"} onClick={() => {
+        <Button icon={icon} variant="plain" aria-label={props.sortOrder === SortOrder.asc ? "Sort descending" : "Sort ascending"} onClick={() => {
```
paoloantinori, line 21: Two notes here. (1) This is an accessibility/labeling change unrelated to the PR's syntax-highlighting scope — worth splitting out so it can be reviewed and reverted independently. (2) The label semantics flipped: the old `aria-label` described the current state (`sort ascending` when `sortOrder === asc`), the new one describes the action a click performs (`Sort descending` when currently ascending). Describing the action is arguably better for a toggle, but it's a deliberate UX change worth making consciously rather than as part of a highlighting PR.

paoloantinori (commented):
Good factoring — hoisting `registerProto` to module scope so the Protobuf Monarch tokenizer/theme can be reused across `ContentTabContent`, `TextEditor`, and `DiffView` is the right way to enable the reuse this PR needs. A few things worth checking before merge:

- **GraphQL highlighting may not actually render.** The diff registers a Monarch tokenizer only for Protobuf; for GraphQL it changes the language id from `graphqlschema` to `graphql` (`ContentTabContent.tsx:18`). Neither id is registered as a Monaco language anywhere in this diff, and `monaco-editor` does not ship GraphQL in its bundled `basic-languages` set (that usually needs a separate contribution such as `monaco-graphql`). Worth confirming in the running UI that a GraphQL artifact is actually tokenized — if it isn't, the GraphQL half of the PR title isn't delivered yet, and either a GraphQL tokenizer or a `monaco-graphql` dependency is needed.

- **Debug `log:` in the Protobuf tokenizer.** The annotation rule carries `log: "annotation token: $0"` (`ProtoEditor.tsx:67`), which writes to the browser console on every annotation token. Now that `registerProto` runs on every editor mount across three components, this is more visible in production than when it lived inside `ProtoEditor`. Recommend dropping `log:` unless still needed for debugging.

- **Scope: unrelated accessibility changes.** `SortOrderToggle.tsx` and `ConfigProperty.tsx` add `aria-label`s, and the sort toggle's label is flipped from describing the current state to describing the next action. These are reasonable improvements but unrelated to syntax highlighting — easier to review and bisect as their own change. The SortOrderToggle semantics flip is worth a deliberate look (see inline).

- Minor: `beforeMount={registerProto}` re-runs `monaco.languages.register` / `setMonarchTokensProvider` / `defineTheme` on every editor mount. Monaco dedupes `register`, so it's safe, just slightly redundant — a module-level `let registered` guard would avoid re-registering the tokenizer each time.

Myself-Praveen:
Thanks for the feedback! I've addressed all the points (added the GraphQL tokenizer, removed the debug log, centralized language registration with a guard, and rebased to drop the unrelated commits). PTAL.

paoloantinori:
Thanks @Myself-Praveen — I'd already reviewed the current head (`f1c73f8`) in my last pass, and the changes all look good: the GraphQL tokenizer is in place, the debug log is gone, and `registerLanguages.ts` centralizes both registrations behind their guards. Nothing else from my side — this looks ready.

Myself-Praveen:
/ready

Thanks @paoloantinori! The branch is up to date, but the Verification Gate is still blocking the tests (likely because of the `waiting-on-maintainer` label). Could you help trigger the CI run or let me know if there's anything else I need to do?

paoloantinori:
Hi @Myself-Praveen — the code side is all good on my end: the GraphQL tokenizer is in place, the debug log is gone, and `registerLanguages` with the module-scope guard is the right factoring. I already reviewed the current head in my last pass.

On the Verification Gate / `waiting-on-maintainer` label: that's an operational action (re-triggering CI, flipping labels) that's outside what I do in these reviews — I can only post review comments, not approve, merge, or change labels/workflows. @EricWittmann or @carlesarnal can clear the gate and kick the run. From a code-review standpoint there's nothing else needed from you.

EricWittmann (commented):
Thanks for the PR — the syntax highlighting work looks good overall!

One structural issue to address before merging: `registerProto` is exported from `ProtoEditor.tsx` and imported by `registerLanguages.ts`, but `ProtoEditor.tsx` itself now calls `registerCustomLanguages` (which calls `registerProto`). This creates a circular dependency: `ProtoEditor.tsx` → `registerLanguages.ts` → `ProtoEditor.tsx`.

It works today because of the idempotency guards and JS module caching, but it's fragile and architecturally inconsistent — you already have the right pattern in `GraphQLLanguage.ts` as a standalone language registration file.

Please extract `registerProto` (and `protoBufThemeData`) into a separate `ProtobufLanguage.ts` file, mirroring what you did with `GraphQLLanguage.ts`. That breaks the circular import and keeps `ProtoEditor.tsx` focused on the React component. The import in `registerLanguages.ts` would then point to `ProtobufLanguage.ts` instead of `ProtoEditor.tsx`.

EricWittmann (commented):
A few more items to address:

**Missing Protobuf keywords** — The Protobuf tokenizer is missing several commonly used keywords: `syntax`, `repeated`, `enum`, `oneof`, `map`, `reserved`, `extend`, `stream`, `weak`, and `public`. Nearly every `.proto` file starts with `syntax = "proto3";`, and `repeated`, `enum`, `oneof`, and `map` appear frequently in message definitions. Without these, a significant amount of typical Protobuf content won't get keyword highlighting. Please add them to the `keywords` array in the Protobuf tokenizer.

**Custom themes are registered but never applied** — Both `registerProto` and `registerGraphQL` call `monaco.editor.defineTheme(...)` to register custom themes (`"protobuf"` and `"graphql"`), but none of the editor components set `theme="protobuf"` or `theme="graphql"`. The themes are registered and then never used — the editors fall back to the default VS theme. Either set the `theme` prop on the relevant `<Editor>` components to apply the custom colors, or remove the `defineTheme` calls and the theme data constants (`protoBufThemeData`, `graphqlThemeData`) to avoid dead code.

**Misplaced import in ProtoEditor.tsx** — The `import { registerCustomLanguages } from "./registerLanguages.ts"` statement appears after the `registerProto` function definition instead of at the top of the file with the other imports. Please move it to the top. (This will resolve itself naturally if you extract `registerProto` into `ProtobufLanguage.ts` as requested in the earlier comment.)

EricWittmann:
@Myself-Praveen Just a few tweaks needed (see my comments above) and this should be ready to merge. :+1:

EricWittmann:
Thanks for the contribution @Myself-Praveen! The refactoring to centralize language registration is well done, and it's great to have proper syntax highlighting for GraphQL and Protobuf artifacts. Merged.

PR #8927, opened 2026-07-27
fix(ui): prevent duplicate skill IDs in Agent Card SkillEditor

### LINKED ISSUE #8882: UI: Agent Card SkillEditor allows duplicate skill IDs, so edit/delete corrupt all skills sharing that ID
## Summary

In the Agent Card editor, `SkillEditor` uses the user-entered skill `id` as both the React list `key` and as the identity for edit and delete, but it never checks that the `id` is unique when a skill is added. Two skills with the same `id` produce duplicate React keys, and both edit and delete then operate on every skill that shares that `id`, corrupting the agent card's skill list.

## Affected code

`ui/ui-app/src/app/components/agentCard/SkillEditor.tsx`

- The skill `id` is the React list key:

```tsx
// renderSkillsList()
{skills.map((skill) => (
    <DataListItem key={skill.id} aria-labelledby={`skill-item-${skill.id}`}>
```

- The `id` is free-text and is only editable while adding (`isDisabled={!isAdding}`), so a new skill can be given any value.

- `handleSaveSkill()` validates only that `id` and `name` are non-empty, not that `id` is unique, and appends without a duplicate check:

```tsx
const handleSaveSkill = (): void => {
    if (!editingSkill || !editingSkill.id || !editingSkill.name) {
        return;
    }
    if (isAdding) {
        onChange([...skills, editingSkill]);            // no uniqueness check
    } else {
        const updated = skills.map(s =>
            s.id === editingSkill.id ? editingSkill : s // updates ALL matches
        );
        onChange(updated);
    }
    ...
};
```

- `handleDeleteSkill()` removes every skill with the given id:

```tsx
const handleDeleteSkill = (skillId: string): void => {
    const updated = skills.filter(s => s.id !== skillId); // removes ALL matches
    onChange(updated);
};
```

The parent `AgentCardEditor.handleSkillsChange` just forwards the array (`updateField("skills", skills)`), so there is no uniqueness validation at any level.

## Steps to reproduce

1. Open the Agent Card editor and add a skill with id `foo`.
2. Add a second skill, also with id `foo` (the form allows it).
3. Try to delete one of the two `foo` skills, then try to edit one of them.

## Expected

Adding a second skill with an existing id should be rejected with an inline validation error, and edit/delete should always act on exactly one skill.

## Actual

- React logs "Encountered two children with the same key, `foo`" and reconciliation for those rows is unreliable.
- Deleting one `foo` skill removes both.
- Editing one `foo` skill overwrites both to identical values.
- The Agent Card is saved with duplicate skill ids, which is invalid under the A2A AgentCard schema (skill ids are meant to be unique).

I verified the delete/edit corruption by running the exact `filter`/`map` logic from the component:

```
two skills with id "foo"  -> duplicate keys: true
delete ONE "foo"          -> remaining count: 0   (expected 1)
edit ONE "foo" to "C"     -> names: ["C","C"]     (expected one changed)
```

Duplicate ids are also reachable without manual entry, for example importing an agent card whose JSON already contains repeated skill ids (`ImportAgentModal`).

## Root cause

The skill `id` doubles as the React key and the record identity, but nothing enforces uniqueness on add, and both edit and delete match by `id` with `map`/`filter`, so they affect every duplicate.

## Proposed fix

- Reject a duplicate `id` on add with an inline validation error (check `skills.some(s => s.id === editingSkill.id)` in `handleSaveSkill` / the add validation).
- Alternatively, give each skill a stable internal key that is independent of the editable `id`.

Secondary, same file: `handleRemoveTag` filters tags by value (`tags.filter(t => t !== tagToRemove)`) and tags are rendered with `key={index}`, so duplicate tags cannot be removed individually either.

## Environment

- `@apicurio/registry-ui-app` 3.3.1-Dev
- Agent Catalog UI (introduced in #8483)

## Notes

Verified against current `main` by inspecting `SkillEditor.tsx` and by running the component's delete/edit logic in isolation. I did not find an existing open issue for this. Happy to work on the fix if the maintainers agree with the approach.

Aditya-132:
## What

Fixes a data-integrity bug in the Agent Card editor: `SkillEditor` allowed two skills to share the same `id`, which corrupted edit and delete.

Closes #8882.

## Why

The skill `id` is used both as the React list `key` and as the record identity for edit and delete (`skills.filter(s => s.id !== skillId)` and `skills.map(s => s.id === editingSkill.id ? ... : s)`). Adding a skill validated only that `id` and `name` were non-empty, not that `id` was unique. With two skills sharing an `id`:

- React logs "Encountered two children with the same key".
- Deleting one skill removed both.
- Editing one skill overwrote both.

Duplicate ids are reachable both by manual entry (the id field is free-text while adding) and by importing an agent card whose JSON already contains repeated skill ids.

## Changes

`ui/ui-app/src/app/components/agentCard/SkillEditor.tsx`:

- Compute `isDuplicateSkillId` while adding.
- Block the save in `handleSaveSkill` and disable the primary button when the id already exists.
- Show an inline error on the Skill ID field using the existing PatternFly `FormHelperText` / `HelperText` / `HelperTextItem` pattern.

Editing an existing skill is unaffected, since the id field stays disabled while editing.

## Testing

- `eslint` passes on the changed file.
- `tsc` reports no new type errors from the change. The only errors in a fresh checkout are the pre-existing `@sdk/lib/generated-client/models` resolution errors, which are present on unmodified `main`.
- Verified the guard by running the component's save logic in isolation: adding a duplicate id is now a no-op (the list stays at one entry), while adding a unique id and editing an existing skill still work.

Note: `ui-app` has no unit-test runner today (a minimal Vitest setup is proposed in #8880). Once that lands, a focused unit test for this guard would be a natural addition.


ui/ui-app/src/app/components/agentCard/SkillEditor.tsx
```diff
@@ -47,6 +50,9 @@ export const SkillEditor: FunctionComponent<SkillEditorProps> = (props: SkillEdi
         tags: []
     };
 
+    const isDuplicateSkillId: boolean =
```
paoloantinori, line 53: Because this is gated on `isAdding`, the check does not fire when a card is imported with a JSON `skills` array that already contains repeated ids (the second reproduction vector called out in the PR description). Those duplicates pass through into state unchanged, and the downstream `filter`/`map`/React-`key` corruption this PR fixes would still occur on a later edit/delete. Either dedupe on import or explicitly scope #8882 to the add path so the import case is tracked separately.

paoloantinori (commented):
Good catch on the root cause — using `id` as both the React `key` and the record identity for `filter`/`map` is exactly why duplicates corrupt edit/delete, and the add-path guard (disable save + inline error) is the right UX. One scope gap worth confirming against the PR description itself.


- **The import path is still unguarded.** The PR description states duplicate ids are reachable two ways: manual entry *and* "importing an agent card whose JSON already contains repeated skill ids." The fix only covers the first. `isDuplicateSkillId` is gated on `isAdding`, so when a card is imported with a malformed `skills` array (two entries sharing an `id`), the duplicates sail through unchanged — and the same `filter`/`map`/`key` corruption the PR is fixing will still occur downstream whenever the user later edits or deletes one of them. If import-side validation is out of scope for this PR, it is worth either (a) adding a normalize/dedupe step on import, or (b) explicitly narrowing the PR/issue (#8882) to "add path" so the import case is tracked separately.


- Minor: `skills.some(s => s.id === editingSkill?.id)` matches on any pre-existing skill, which is correct for the add case. Since `isAdding` is the gate, the optional-chain on `editingSkill?.id` is fine, but a brief comment noting "only meaningful when isAdding" would help future readers who might try to reuse the flag elsewhere.

Aditya-132:
Thanks @paoloantinori, good call on the import vector. Addressed in the latest commit (28aebda).

**Import path.** `ImportAgentModal.handleUrlChange` now validates the parsed card's `skills` before accepting it: if two entries share an `id`, the import is rejected with an inline error ("The agent card contains skills with duplicate IDs. Skill IDs must be unique.") and the wizard cannot proceed. I went with rejecting rather than a silent dedupe/normalize, to stay consistent with the add path (which also blocks duplicates) and to avoid silently dropping or renaming a user's skills. Both reproduction vectors from the description are now covered.

I verified the detection against duplicate ids, unique ids, empty/missing ids, and the "one duplicate among many" case.

**Minor.** I left `isDuplicateSkillId` without an inline comment to match the surrounding style; it is only referenced in the add path (the save guard, the button disable, and the id field), so its meaning stays local to those lines.

malladinagarjuna2:
@Aditya-132  
really glad that you have opened this issue and pr
@paoloantinori  has already given a nice review, just wantedto add  few points 
1. proposed fixes section of the  issue talks about giving each skill a stable internal key  independent of the editable id. 
this pr does only the former part of avoiding the duplicating ids, giving each skill a stable internal key independent of the editable id would have closed both the add-form and import vectors in one change, since it decouples the React key/filter/map identity from the user-editable field entirely, which is the actual root cause, not just the entry point.


2. a minor one:  just on experimenting i found that currently 
skill id: "foo" and skill id: "foo " both are getting accepted, thus  the duplicate check does exact string equality on id with no .trim(). Not a blocker, but "foo" vs "foo " would sail through as "not a duplicate" while being effectively indistinguishable to a human filling the form, cheap to fix alongside this PR.

paoloantinori:
Thanks @malladinagarjuna2 — that's a fair design point worth discussing.

The "stable internal key decoupled from the editable id" idea has real merit as a robustness improvement: it would mean the React reconciliation key and the filter/map record identity no longer depend on a user-editable field, so duplicate ids could never corrupt edit/delete again. That's a stronger invariant than guarding the add form alone.

Two reasons I'd keep this PR as the focused fix for #8882 and treat the stable-key refactor as a follow-up:

1. Both the add-path guard here and the new import-path validation in `ImportAgentModal` treat a duplicate `id` as **invalid data** and reject it. That's consistent with treating `id` as a unique identifier — the import path now explicitly enforces "Skill IDs must be unique". The stable-key approach would instead *tolerate* duplicate ids at the UI layer, which is fine only if the spec genuinely allows them; it changes what the editor considers valid input.
2. Decoupling the React key from `id` touches the data model and every `filter` / `map` / `onChange` site in `SkillEditor`, so it's a larger change that deserves its own PR with dedicated tests.

So I'd take this as the targeted fix, and if you feel duplicate ids should be valid input the UI accepts, a follow-up PR proposing the stable-key refactor is the right place to make that case — happy to review it there.

Aditya-132:
Thanks @malladinagarjuna2 for taking a look, and @paoloantinori for the guidance.

**On the trim point (2):** good catch. Fixed in 7320ed4 — both the add-path guard and the import validation now trim ids before comparing, so `"foo"` and `"foo "` are treated as the same id. Verified across the add and import paths, including the all-whitespace case, which stays blocked as an empty id.

**On the stable-key idea (1):** agreed with @paoloantinori's framing. This PR keeps the invariant that a skill `id` is unique and rejects duplicates as invalid input (both the add form and the import path now enforce that), which is the targeted fix for #8882. Decoupling the React key and the filter/map identity from the editable `id` is a nice robustness improvement, but as noted it changes what the editor accepts as valid and touches the whole data model, so it is better as its own PR with dedicated tests. Happy to open that follow-up if the team decides duplicate ids should be tolerated at the UI layer.

malladinagarjuna2:
totally agreed with the detailed explaination @paoloantinori  thanks for that i really did'nt think about this case.

paoloantinori:
Glad the explanation helped, @malladinagarjuna. Looking forward to the follow-up if you decide to tackle the dedup-by-id approach.

Aditya-132:
@paoloantinori @carlesarnal can you make it ready for tests

paoloantinori:
Hi @Aditya-132, I gave the latest commits another read — both the add-path guard and the import-path validation look consistent now (trim applied uniformly, Set-size dedup is clean), and the SonarCloud quality gate is green.

One heads-up on the "make it ready for tests" request: I'm an automated review bot and can't apply labels or change PR state — that needs @paoloantinori or @carlesarnal to do directly. From a code-review standpoint, this looks good to me and I have no further blockers.

EricWittmann:
Thanks @Aditya-132 — clean fix covering both entry vectors with consistent trim-aware validation. Good iteration on the review feedback too. Merged!

PR #8547, opened 2026-07-12
fix(avro): improve AvroCanonicalHashUpgrader logging and reference tests

niy-ati:
Fixes #8003

- Report processed/updated/skipped/failed counts at end of upgrade
- Add test coverage for schemas with artifact references (resolveReference path)
- Add failure-case test when a referenced version is missing

PR #8932, opened 2026-07-27
fix(avro): improve AvroCanonicalHashUpgrader logging and reference tests

### LINKED ISSUE #8003: fix(avro): follow-up improvements for AvroCanonicalHashUpgrader
## Summary

Follow-up improvements for the Avro canonical hash upgrader introduced in #7550.

### 1. Add test coverage for schemas with artifact references

The `AvroCanonicalHashUpgrader` has a `resolveReference()` code path that queries the DB for referenced content. This path is currently untested. If reference resolution fails, the error is silently swallowed (logged as warning), leaving stale hashes in place.

**What to do:**
- Add a test case that creates an Avro schema with artifact references, corrupts its canonical hash, and verifies the upgrader recomputes it correctly
- Add a test for the failure case (missing reference) to verify behavior is acceptable

### 2. Improve end-of-upgrade logging

Currently the upgrader only logs the `successCount` at the end. For operators running the upgrade on large registries, it would be helpful to also report:
- Total Avro content rows processed
- How many were skipped (hash already correct)
- How many failed (e.g., reference resolution errors)

This helps verify the upgrade ran correctly and identify any schemas that may need manual attention.

niy-ati:
Fixes #8003

## Summary
- End-of-upgrade logging now reports processed / updated / skipped / failed counts
- MappedQuery stream is closed via try-with-resources to avoid JDBC leaks on large upgrades
- Missing UPDATE row throws IllegalStateException (counts as failed, not silent skip) — intentional
- Pre-validate artifact references so broken refs fail the row and leave the stale hash unchanged
- Tests cover resolveReference success, missing-ref failure, null refs column, and empty refs array

## Review feedback addressed (from #8547)
- IllegalStateException on missing row: intentional; documented in code comment
- Null refs vs empty array: covered by dedicated tests
- @Test annotations: verified present and correct on all new tests
- Copilot: stream now closed in try-with-resources

## Test plan
- [x] AvroCanonicalHashUpgraderTest cases included in this PR
- [ ] CI

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/AvroCanonicalHashUpgrader.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/AvroCanonicalHashUpgrader.java
@@ -21,6 +21,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
 
 import static io.apicurio.registry.storage.impl.sql.RegistryContentUtils.normalizeGroupId;
 
@@ -52,22 +53,27 @@ public class AvroCanonicalHashUpgrader implements IDbUpgrader {
         AtomicInteger skippedCount = new AtomicInteger();
         AtomicInteger failedCount = new AtomicInteger();
 
-        handle.createQuery(sql).bind(0, ArtifactType.AVRO).setFetchSize(50)
-                .map(new ContentWithTypeRowMapper()).stream().forEach(entity -> {
-                    processedCount.incrementAndGet();
-                    try {
-                        int result = updateEntity(handle, entity);
-                        if (result > 0) {
-                            updatedCount.incrementAndGet();
-                        } else {
-                            skippedCount.incrementAndGet();
-                        }
-                    } catch (Exception ex) {
-                        failedCount.incrementAndGet();
-                        log.warn("Failed to update canonical hash for contentId {}.",
-                                entity.contentEntity.contentId, ex);
+        // Same pattern as SqlExportRepository / SqlVersionRepository: MappedQuery.stream() only
+        // closes the ResultSet/Statement via onClose, so the Stream must be closed explicitly.
+        Stream<ContentWithType> stream = handle.createQuery(sql).bind(0, ArtifactType.AVRO).setFetchSize(50)
+                .map(new ContentWithTypeRowMapper()).stream();
+        try (stream) {
+            stream.forEach(entity -> {
+                processedCount.incrementAndGet();
+                try {
+                    int result = updateEntity(handle, entity);
+                    if (result > 0) {
+                        updatedCount.incrementAndGet();
+                    } else {
+                        skippedCount.incrementAndGet();
                     }
-                });
+                } catch (Exception ex) {
+                    failedCount.incrementAndGet();
+                    log.warn("Failed to update canonical hash for contentId {}.",
+                            entity.contentEntity.contentId, ex);
+                }
+            });
+        }
 
         log.info(
                 "Avro canonical hash upgrade complete: processed={}, updated={}, skipped={}, failed={}.",
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/AvroCanonicalHashUpgrader.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/AvroCanonicalHashUpgrader.java
@@ -20,6 +20,7 @@ import org.slf4j.LoggerFactory;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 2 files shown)


paoloantinori (commented):
Good improvements to the upgrader's observability and error handling. The counters (processed/updated/skipped/failed) in the completion log are a meaningful upgrade over the previous single-count message, and the dedicated tests for reference resolution edge cases (happy path, broken refs, null refs column, empty refs array) are well-structured.

A couple of observations:

- **Stream resource management vs. PR description.** The PR description states "MappedQuery stream is closed via try-with-resources to avoid JDBC leaks on large upgrades," but the current diff does not wrap the stream in try-with-resources — the `forEach` is called directly on `handle.createQuery(sql)...stream()`. Since exceptions are caught inside the lambda (not propagated out of the stream), the stream will exhaust normally in practice, so the leak risk is low. Still, the description and the code don't match — worth either adding the try-with-resources or correcting the description so reviewers aren't looking for code that isn't there.

- **`validateReferencesResolvable` issues one SQL query per reference per nesting level.** For Avro schemas in practice this is shallow, so stack depth and query count aren't concerns, just calling it out as an N+1 pattern on the migration path. Acceptable given this runs once during upgrade.

app/src/main/java/io/apicurio/registry/storage/impl/sql/upgrader/AvroCanonicalHashUpgrader.java
```diff
@@ -46,24 +47,39 @@ public void upgrade(Handle handle) throws Exception {
...
 
-        int successCount = handle.createQuery(sql).bind(0, ArtifactType.AVRO).setFetchSize(50)
-                .map(new ContentWithTypeRowMapper()).stream().mapToInt(entity -> {
+        AtomicInteger processedCount = new AtomicInteger();
+        AtomicInteger updatedCount = new AtomicInteger();
+        AtomicInteger skippedCount = new AtomicInteger();
+        AtomicInteger failedCount = new AtomicInteger();
+
+        handle.createQuery(sql).bind(0, ArtifactType.AVRO).setFetchSize(50)
+                .map(new ContentWithTypeRowMapper()).stream().forEach(entity -> {
```
paoloantinori: The PR description claims this stream is wrapped in try-with-resources, but the diff shows a bare `stream().forEach(...)`. Since exceptions are caught inside the lambda the stream exhausts normally, so the practical leak risk is low — but it's worth either adding the try-with-resources (to match the description and guard against future changes that might let exceptions propagate) or correcting the description.

niy-ati:
Thanks @paoloantinori , fixed your suggestions
the description said try-with-resources but the code didn’t. Fixed in `b6494bd` using the same `try (stream)` pattern as `SqlExportRepository` / `SqlVersionRepository`. 
Left `validateReferencesResolvable` alone since you said the N+1 cost is fine for a one-time upgrade.

paoloantinori:
Thanks @niy-ati — the `try (stream)` change in `b6494bd` matches the `SqlExportRepository` / `SqlVersionRepository` pattern, and leaving `validateReferencesResolvable` as-is for the one-time upgrade is the right call. The new test coverage (null refs column, empty refs array, broken-ref failure leaving the stale hash unchanged) is exactly what was missing before. Looks good to me.

niy-ati:
@carlesarnal , fixed your suggested changes so when you have a moment, could you please take a look at this one
I addressed the follow-up from @paoloantinori  in `b6494bd` and he confirmed the direction looks good from his side.
lemme know if it looks aligned to you as well  or any changes are required .

paoloantinori:
Thanks @niy-ati — confirming I've seen the `b6494bd` update and the direction still looks right from my side (the `try (stream)` matches the `SqlExportRepository` / `SqlVersionRepository` pattern, and the documented `IllegalStateException` on a vanished row is the correct one-time-upgrade behavior). Deferring to @carlesarnal for the final call on the rest.

EricWittmann:
Thanks for the contribution, @niy-ati! The logging improvements and reference validation are solid. Merged.

carlesarnal:
Took a look at the final state of this PR. Everything looks solid:

- **Resource management**: The `try (stream)` pattern in `AvroCanonicalHashUpgrader.upgrade()` correctly ensures the JDBC resources are released, matching the `SqlExportRepository`/`SqlVersionRepository` convention.
- **Error handling**: Per-row exception handling with `failedCount` tracking is the right approach — a single broken row won't abort the entire upgrade.
- **Reference validation**: `validateReferencesResolvable` pre-validates refs before recomputing the hash, so broken references are surfaced as counted failures rather than producing incorrect hashes. The N+1 query pattern is fine for a one-time upgrade.
- **Test coverage**: The four new tests (`testUpgraderRecomputesHashForSchemaWithArtifactReferences`, `testUpgraderLeavesHashUnchangedWhenReferenceResolutionFails`, `testUpgraderRecomputesHashWithNullRefsColumn`, `testUpgraderRecomputesHashWithEmptyRefsArray`) cover the key edge cases well, with specific assertions on the resulting hash values.
- **Logging**: The structured completion log (`processed/updated/skipped/failed`) is a meaningful improvement over the previous single-count message.

No issues found. Nice contribution.

PR #8933, opened 2026-07-27
fix(storage): return 400 instead of 500 for content search before ES index is ready

### LINKED ISSUE #8697: bug: ElasticsearchSearchDecorator throws RegistryStorageException for content search, returning HTTP 500 instead of 400
---

## Description

**Registry Version**: snapshot (pinned to commit `acb5abcc`)
**Persistence type**: sql (with Elasticsearch search extension enabled)

When Elasticsearch is configured but its startup indexer has not yet finished (i.e. `startupIndexer.isReady() == false`), a content search request that requires the search index hits the following code in `ElasticsearchSearchDecorator.searchVersions`:

```java
// ElasticsearchSearchDecorator.java, line 55
if (!startupIndexer.isReady()) {
    throw new RegistryStorageException(
            "Content search requires the Elasticsearch search index, which is not "
            + "available. Enable the Elasticsearch search index to use content search.");
}
```

[Permalink to the buggy line](https://github.com/Apicurio/apicurio-registry/blob/acb5abcc81a367bbcc7f9b9078fe46bb9923baf0/app/src/main/java/io/apicurio/registry/storage/decorator/ElasticsearchSearchDecorator.java#L55)

`RegistryStorageException` has no entry in `HttpStatusCodeMap` (a `grep` for `RegistryStorageException` in that file returns only a TODO comment, no map entry). As a result, `HttpStatusCodeMap.getCode()` line 143 falls through to `codeMap.getOrDefault(exceptionClass, HTTP_INTERNAL_ERROR)` and returns **HTTP 500**.

The correct exception type is `ContentSearchNotSupportedException`, which **is** registered in `HttpStatusCodeMap` at line 94 mapped to `HTTP_BAD_REQUEST` (400). The parallel fix in `SqlSearchRepository` (commit `7183fa7`) already throws `ContentSearchNotSupportedException` for this condition; the Elasticsearch code path was missed.

**Side effects of the wrong HTTP status:**

- `CoreRegistryExceptionMapperService` line 62 calls `liveness.suspectWithException(t)` for every 5xx response, so a transient startup condition incorrectly increments the liveness probe suspect counter.
- The response is logged as `[500 ERROR DETECTED]`, making a normal startup race appear as a server fault in monitoring.

**Proposed fix:**

Replace line 55 with `throw new ContentSearchNotSupportedException(...)`, using the same constant/message already used in `SqlSearchRepository`. This routes the response through the existing 400 mapping and avoids triggering the liveness suspect counter.

### Environment

No response

### Steps to Reproduce

This is a code-path issue reproducible by static analysis; the runtime path requires Elasticsearch to be enabled but not yet indexed at request time.

1. Enable Elasticsearch in the Registry configuration.
2. Start the Registry before the Elasticsearch startup indexer finishes (`startupIndexer.isReady() == false`).
3. Send a content search (or any search requiring `searchService.requiresSearchIndex(filters) == true`) to the `/search/versions` endpoint during this startup window.
4. Observe HTTP 500 in the response.

**Code-path trace (static):**

1. `ElasticsearchSearchDecorator.java` line 55 — throws `RegistryStorageException`.
2. `HttpStatusCodeMap.java` — `grep RegistryStorageException` returns only a TODO comment; no map entry exists.
3. `HttpStatusCodeMap.getCode()` line 143: `codeMap.getOrDefault(RegistryStorageException.class, HTTP_INTERNAL_ERROR)` → returns `500`.
4. `CoreRegistryExceptionMapperService` line 62: `liveness.suspectWithException(t)` fires.
5. By contrast, `ContentSearchNotSupportedException.class` is present in `HttpStatusCodeMap` at line 94, mapped to `HTTP_BAD_REQUEST` — confirming where the fix belongs.

**Expected:** A content search request when Elasticsearch is enabled but not yet indexed returns HTTP 400 (Bad Request) with a clear message indicating that content search is temporarily unavailable, consistent with the behaviour of `SqlSearchRepository` after commit `7183fa7`.

**Actual:** The same request returns HTTP 500 (Internal Server Error), increments the liveness probe suspect counter, and is logged as `[500 ERROR DETECTED]`, mischaracterising a transient startup race as a server fault.

### Logs

When the bug is triggered, the Registry logs will contain an entry similar to:

```
[500 ERROR DETECTED] Content search requires the Elasticsearch search index, which is not available. Enable the Elasticsearch search index to use content search.
```

Run with `LOG_LEVEL=DEBUG` and filter for `ElasticsearchSearchDecorator` and `CoreRegistryExceptionMapperService` to confirm the code path.

bhuvan-somisetty:
## Summary

Fixes #8697. `ElasticsearchSearchDecorator.searchVersions()` returns HTTP 500 instead of HTTP 400 when a content search is requested while the Elasticsearch startup indexer hasn't finished yet.

## Root Cause

When `searchService.requiresSearchIndex(filters)` is true but `startupIndexer.isReady()` is false, the decorator throws a plain `RegistryStorageException`. That exception type has no entry in `HttpStatusCodeMap`, so it falls through to the default `HTTP_INTERNAL_ERROR` (500).

This also means `CoreRegistryExceptionMapperService` treats it as a 5xx response and calls `liveness.suspectWithException(t)`, incrementing the liveness probe's suspect counter for what is really just a transient startup race, and the request is logged as a `[500 ERROR DETECTED]` server fault.

`SqlSearchRepository` already handles the equivalent condition by throwing `ContentSearchNotSupportedException`, which is mapped to `HTTP_BAD_REQUEST` (400) in `HttpStatusCodeMap`. The Elasticsearch code path was missed when that fix was applied.

## Changes

- `ElasticsearchSearchDecorator.searchVersions()`: throw `ContentSearchNotSupportedException` instead of `RegistryStorageException` when the search index is required but not yet ready, matching the message and exception type already used in `SqlSearchRepository`.
- Added `ElasticsearchSearchDecoratorTest` covering:
  - `ContentSearchNotSupportedException` is thrown (with the expected message) when the index is required and not ready, and the delegate is never invoked.
  - The search is delegated to `ElasticsearchSearchService` when the index is required and ready.
  - The search falls through to the underlying storage delegate when the index isn't required at all.

## Test plan

- [x] `./mvnw checkstyle:check -pl app` passes
- [x] `./mvnw test -pl app -Dtest=ElasticsearchSearchDecoratorTest` passes
- [x] Verified `ContentSearchNotSupportedException` maps to HTTP 400 via `HttpStatusCodeMap`, matching the existing `SqlSearchRepository` behavior


bhuvan-somisetty:
@EricWittmann @carlesarnal @paoloantinori whenever you get a chance, could one of you take a look at this one? It's a small, contained fix for #8697 (ES search path throwing the wrong exception type/status code), tests are green and checkstyle passes. Thanks!

paoloantinori:
Thanks @bhuvan-somisetty — took a look, this is a clean fix. Throwing `ContentSearchNotSupportedException` instead of `RegistryStorageException` is exactly the right call: it aligns the Elasticsearch path with `SqlSearchRepository` and maps to 400 via `HttpStatusCodeMap` instead of falling through to 500.

The secondary benefit you called out in the PR body is also worth highlighting: `CoreRegistryExceptionMapperService` will no longer call `liveness.suspectWithException(t)` for what's really a transient startup race, so the probe won't get a false suspect bump during index warmup. That's a real operational improvement on top of the status code fix.

Test coverage is solid — the three cases (index required + not ready → throws, index required + ready → delegates, index not required → falls through to delegate) cover the decision tree cleanly. Nothing blocking from my side.

bhuvan-somisetty:
Thanks for the thorough review, @paoloantinori! Glad the fix landed cleanly and that the liveness probe side-effect was a useful catch to call out too.

The branch had fallen behind main, which is what was tripping the verification gate - just merged in the latest main and pushed, so CI should go green now.

bhuvan-somisetty:
@EricWittmann @carlesarnal - would appreciate a review when you get a chance. @paoloantinori already reviewed and approved, and CI is green now (branch was out of date, updated it). Thanks!

paoloantinori:
Thanks @bhuvan-somisetty — nothing further from my side, my earlier review still stands on the rebased head. Deferring to @EricWittmann / @carlesarnal for the final approve.

bhuvan-somisetty:
Thanks @paoloantinori, appreciate you confirming! Will wait on @EricWittmann / @carlesarnal for the final approve.

paoloantinori:
Sounds good — deferring to @EricWittmann / @carlesarnal for the final call. 👍

bhuvan-somisetty:
@paoloantinori since you've already reviewed the code, would you be able to run `/accept` on this? PR is still stuck in `lifecycle/new`, which is why the Verification Gate keeps failing (it only unblocks once the PR moves to `ready-for-review`). I'll follow up with `/ready` right after.

paoloantinori:
Hi @bhuvan-somisetty — apologies, but I can't run `/accept` or other lifecycle commands. I post code reviews only; triage/lifecycle transitions (accept, ready, labels) sit with the project maintainers (@EricWittmann / @carlesarnal), who also hold the final approval on this one. From the code side there's nothing further from me — my earlier review still stands on the current head. Pinging the maintainers directly is the right path to unblock the Verification Gate.

bhuvan-somisetty:
Got it, thanks for clarifying @paoloantinori!, @EricWittmann/@carlesarnal - could one of you run `/accept` on this when you get a chance? That's the only thing blocking the Verification Gate at this point, everything else is green.

paoloantinori:
Thanks @bhuvan-somisetty — that's right, the `/accept` lifecycle command is out of my lane. Over to @EricWittmann / @carlesarnal for the final triage step. Nothing further on the code from my side; my earlier review still stands.

bhuvan-somisetty:
@EricWittmann heads up — it looks like the `/accept` didn't actually get processed by the lifecycle bot (no PR Lifecycle workflow run fired for that comment, so the PR is still stuck in `lifecycle/new`). `/retry` only reconciles existing label state, it doesn't replay `/accept`. Could you re-post `/accept` so the bot picks it up? Happy to follow up with `/ready` right after.

bhuvan-somisetty:
@EricWittmann all checks are passing now — full test suite (unit, integration across all storage variants, SDK, build, lint) is green and the Verification Gate has passed. Ready for your review whenever you get a chance. Thanks!

EricWittmann:
Thanks for the excellent contribution @bhuvan-somisetty! Textbook bug fix — clear root cause analysis, minimal one-line fix, and thorough test coverage. The liveness probe improvement is a nice bonus. Merged.

PR #8934, opened 2026-07-27
fix(rest): clamp pagination parameters to prevent integer overflow (#8732)

### LINKED ISSUE #8732: Unvalidated pagination parameters cause integer overflow and HTTP 500 (A2A entitled-agents and ccompat listVersions)
## Description

**Registry Version**: snapshot (`main`, commit `5115d1136`)
**Persistence type**: any (the bug is in in-memory REST pagination arithmetic, independent of the storage backend)

Two REST endpoints compute an in-memory `List.subList(fromIndex, toIndex)` from client-supplied `offset`/`limit` query parameters without bounding the limit. `fromIndex + limit` can overflow `int` to a negative `toIndex` (and a negative `offset` can reach `subList` directly), so `subList` throws and the request fails with an unhandled exception that surfaces to the client as a generic **HTTP 500**.

### 1. `WellKnownResourceImpl.getEntitledAgents` (A2A)

`app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java:164-168`

```java
int total = visible.size();
int safeOffset = Math.max(0, Math.min(offset, total));
int safeLimit  = Math.max(0, limit);                       // clamps below 0, but no upper bound
int toIndex    = Math.min(safeOffset + safeLimit, total);  // safeOffset + safeLimit can overflow
List<SearchedArtifactDto> page = visible.subList(safeOffset, toIndex); // fromIndex > toIndex -> throws
```

With at least one visible agent, `GET /.well-known/agents/entitled?offset=1&limit=2147483647` makes `safeOffset = 1` and `safeOffset + safeLimit` overflow to `Integer.MIN_VALUE`, so `toIndex` becomes negative and `subList(1, -2147483648)` throws `IllegalArgumentException`. The endpoint is annotated `@Authorized(level = Read)`, so any caller with read access can trigger it (anonymous, if anonymous read is enabled). It is gated behind A2A + entitlements (`apicurio.a2a.enabled`, experimental / off by default). The overflow requires at least one visible agent and `offset >= 1`; with zero visible agents `safeOffset` is `0` and the call is safe.

### 2. `SubjectsResourceImpl.listVersions` (Confluent-compat v7)

`app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/SubjectsResourceImpl.java:242-248`

```java
int effectiveOffset = offset != null ? offset.intValue() : 0;                        // not clamped
int effectiveLimit  = (limit != null && limit.intValue() > 0) ? limit.intValue() : rval.size();
if (effectiveOffset > 0 || effectiveLimit < rval.size()) {
    int toIndex = Math.min(effectiveOffset + effectiveLimit, rval.size());            // can overflow
    if (effectiveOffset < rval.size()) {
        rval = rval.subList(effectiveOffset, toIndex);                               // throws
    ...
```

`effectiveOffset` is taken straight from the query param with no lower or upper clamp, so:

- a **negative offset** (e.g. `offset=-1&limit=1`) reaches `subList(-1, ...)` -> `IndexOutOfBoundsException`, and
- `effectiveOffset + effectiveLimit` **overflows** the same way (`offset=1&limit=2147483647`) -> `IllegalArgumentException`.

This path is more broadly reachable than the A2A one: the Confluent-compatibility API is a mainstream feature and is **not** behind the experimental A2A flag. (The v8 ccompat `SubjectsResourceImpl` does not share this pagination code and is unaffected.)

### Why this is an oversight, not by design

Sibling pagination paths already cap the limit; only these two do not:

- **`WellKnownResourceImpl.searchAgents`** (`:372`) does `int safeLimit = Math.max(1, Math.min(limit, 500));`.
- **`AgentSearchRequest`** (`app/src/main/java/io/apicurio/registry/a2a/rest/beans/AgentSearchRequest.java`, used by `searchAgentsAdvanced`) clamps in its setters — `setLimit` -> `Math.max(1, Math.min(limit, MAX_LIMIT))`, `setOffset` -> `Math.max(0, offset)` — so that path is safe.

Both affected methods take raw `Integer offset, Integer limit` query params (`@QueryParam` + `@DefaultValue`, no `@Min`/`@Max`/`@Valid`) and are the only pagination paths left with no upper bound on the limit. The resulting `IllegalArgumentException` / `IndexOutOfBoundsException` is not a registered Apicurio exception, so `HttpStatusCodeMap.getCode(...)` returns its default of `HTTP_INTERNAL_ERROR` — which is why the client sees a 500 rather than a 400.

### Steps to Reproduce

The crash is pure pagination arithmetic plus `List.subList`. Running the exact arithmetic of each method:

```
[getEntitledAgents]   total=3  offset=1   limit=2147483647 : CRASH IllegalArgumentException: fromIndex(1) > toIndex(-2147483648)
[getEntitledAgents]   total=0  offset=1   limit=2147483647 : OK size=0   (safe: safeOffset=0)
[searchAgents capped] total=3  offset=1   limit=2147483647 : OK size=2   (limit capped to 500)
[listVersions]        size=3   offset=1   limit=2147483647 : CRASH IllegalArgumentException: fromIndex(1) > toIndex(-2147483648)
[listVersions]        size=3   offset=-1  limit=1          : CRASH IndexOutOfBoundsException: fromIndex = -1
[listVersions]        size=3   offset=0   limit=2          : OK size=2   (normal)
```

At the REST level:

- **A2A** (with A2A + entitlements enabled and at least one entitled agent): `GET /.well-known/agents/entitled?offset=1&limit=2147483647` -> HTTP 500.
- **ccompat** (subject with at least two versions): `GET /subjects/{subject}/versions?offset=1&limit=2147483647` or `?offset=-1&limit=1` -> HTTP 500.

**Expected:** out-of-range or oversized pagination parameters are clamped (as `searchAgents` and `AgentSearchRequest` already do), returning a normal (possibly empty) page — or a `400 Bad Request` for clearly invalid input.

**Actual:** the request throws an unhandled `IllegalArgumentException` / `IndexOutOfBoundsException`, surfaced to the client as a generic HTTP 500.

### Severity

Robustness / availability (DoS-class) — not RCE or auth bypass. Each crafted request returns a single 500; it is not a persistent outage. `getEntitledAgents` is gated behind the experimental A2A feature; `listVersions` is in the generally-available ccompat API. Reported as a correctness / input-validation bug, not a critical vulnerability.

### Possible fix

Clamp the limit (and, for ccompat, the offset) exactly as the sibling paths do — e.g. `int safeLimit = Math.max(0, Math.min(limit, MAX_PAGE_SIZE));`, compute `toIndex` with a widening `long` (or clamp before adding) to avoid the overflow, and clamp `effectiveOffset` to `>= 0` in `listVersions`. A one-line-per-site change mirroring `searchAgents` / `AgentSearchRequest`.

I'd be happy to work on this — I have the arithmetic reproduced against current `main` and a small fix following the existing capped-pagination pattern.

### Relevant files

- `app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java` — `getEntitledAgents` (bug), `searchAgents` (capped sibling)
- `app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/SubjectsResourceImpl.java` — `listVersions` (bug)
- `app/src/main/java/io/apicurio/registry/a2a/rest/beans/AgentSearchRequest.java` — clamping setters (safe pattern)
- `app/src/main/java/io/apicurio/registry/services/http/HttpStatusCodeMap.java` — default 500 mapping

DSingh0304:
Fixes #8732

## Summary
This PR resolves the integer overflow crashes in the REST pagination endpoints (`WellKnownResourceImpl.getEntitledAgents` and `SubjectsResourceImpl.getSubjectVersions`). 

When users supplied a maximum `limit` (e.g. `2147483647`) and a positive `offset`, the arithmetic overflowed `int` to a negative number, causing `List.subList` to throw an unhandled exception (HTTP 500). 

## Changes
- `WellKnownResourceImpl.getEntitledAgents()`: Clamped the `limit` parameter to a max of 500, matching the existing safe logic found in `searchAgents`.
- `SubjectsResourceImpl.getSubjectVersions()`: Clamped the `offset` to `>= 0`.
- `SubjectsResourceImpl.getSubjectVersions()`: Used `(long)` widening for the `effectiveOffset + effectiveLimit` addition to eliminate overflow risks entirely.
- Added regression tests in `CCompatV7EnhancementsTest` to ensure large limits and negative offsets no longer result in a 500 error.

## Test plan
- [x] `./mvnw test -pl app -Dtest=CCompatV7EnhancementsTest` passes


DSingh0304:
Hi @paoloantinori @carlesarnal  This PR resolves the pagination overflow issue described in #8732. The fix implements the same pagination clamping pattern already used elsewhere in the codebase and safely handles integer limits using widened long arithmetic. All tests are passing, and I've added two new regression tests in `CCompatV7EnhancementsTest` to ensure that boundary values like `limit=2147483647` and `offset=-1` now safely return a `200 OK`.

Please let me know if there's anything else needed.

app/src/test/java/io/apicurio/registry/noprofile/ccompat/rest/v7/CCompatV7EnhancementsTest.java
```diff
@@ -212,6 +212,42 @@ public void testSubjectVersionsPagination() throws Exception {
...
+        var schema1 = "{\"type\" : \"string\"}";
+        var schemaContent1 = new RegisterSchemaRequest();
+        schemaContent1.setSchema(schema1);
+        given().when().contentType(ContentTypes.COMPAT_SCHEMA_REGISTRY_STABLE_LATEST)
+                .body(objectMapper.writeValueAsString(schemaContent1))
+                .post("/ccompat/v7/subjects/{subject}/versions", subject).then().statusCode(200);
+
+        // Test versions endpoint with large limit
+        given().when().queryParam("offset", 1).queryParam("limit", 2147483647)
+                .get("/ccompat/v7/subjects/{subject}/versions", subject).then().statusCode(200);
```
paoloantinori, line 246: With only one version created and `offset=1`, `effectiveOffset(1) < rval.size()(1)` is false, so the `subList` call (where the overflowed `toIndex` would throw) is never reached. Consider creating 2-3 versions here so the test fails on the pre-fix code and passes after — right now it returns 200 either way.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -170,7 +170,7 @@ public AgentSearchResults getEntitledAgents(Integer offset, Integer limit) {
 
         int total = visible.size();
         int safeOffset = Math.max(0, Math.min(offset, total));
-        int safeLimit = Math.max(0, limit);
+        int safeLimit = Math.max(1, Math.min(limit, 500));
```
paoloantinori: This changes two things vs. the old `Math.max(0, limit)`: caps at 500 (caller asking for more gets silently truncated) and raises the floor 0 to 1 (`limit=0` previously returned an empty page, now returns 1 item). The 500 cap matches `searchAgents`; just confirm the floor change is intended.

app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/SubjectsResourceImpl.java
```diff
@@ -238,12 +238,11 @@ public List<BigInteger> getSubjectVersions(String subject, String groupId, Boole
                 .sorted()
                 .collect(Collectors.toList());
 
-        // Apply pagination
-        int effectiveOffset = offset != null ? offset.intValue() : 0;
+        int effectiveOffset = Math.max(0, offset != null ? offset.intValue() : 0);
         int effectiveLimit = (limit != null && limit.intValue() > 0) ? limit.intValue() : rval.size();
 
         if (effectiveOffset > 0 || effectiveLimit < rval.size()) {
```
paoloantinori, line 244: The `(long)` widening here is the correct fix for the overflow. Note that `toIndex` only reaches `rval.subList(...)` when `effectiveOffset < rval.size()` (the guard a few lines below), so the regression test needs `size > effectiveOffset` to actually exercise the throwing path.

paoloantinori (commented):
The core arithmetic fix is correct: widening `effectiveOffset + effectiveLimit` to `long` before `Math.min` eliminates the int overflow at the source, and clamping `offset >= 0` matches the existing `WellKnownResourceImpl` pattern. Two things worth addressing before merge — one about test fidelity and one about a behavioral change in the well-known endpoint.


- **Regression tests do not reproduce the actual overflow path.** Both new tests create exactly one subject version. In `getSubjectVersions`, the overflowed `effectiveOffset + effectiveLimit` only becomes the `toIndex` passed to `rval.subList(...)` when `effectiveOffset < rval.size()` (the guard a few lines below the clamp). With `size == 1` and `offset == 1`, that guard is false, so `subList` is never the call site that would throw. Likewise the negative-offset case: with one version and `limit == 1`, neither pagination predicate (`effectiveOffset > 0` nor `effectiveLimit < size`) is true, so the block is skipped entirely and the endpoint returns 200 regardless of whether the fix is applied. To genuinely guard #8732, the tests should create at least 2-3 versions and use an `offset`/`limit` combination that lands inside the list while keeping the overflowed sum negative pre-fix (e.g. 3 versions, `offset=1`, `limit=Integer.MAX_VALUE`).


- **`WellKnownResourceImpl` limit clamp changes two behaviors at once.** `Math.max(0, limit)` to `Math.max(1, Math.min(limit, 500))` both caps at 500 (an API-visible change: a caller requesting 1000 entitled agents now silently gets 500) and raises the floor from 0 to 1 (a `limit=0` request, which previously yielded an empty page, now returns one item). The 500 cap is defensible and matches `searchAgents`, but worth confirming explicitly that both edge changes are intended.


- Minor: the `(int)` cast after `Math.min((long) ..., rval.size())` is safe because the result is bounded by `rval.size()` (an int), but a one-line comment to that effect would help future readers.

paoloantinori:
Thanks @DSingh0304 — the arithmetic fix itself looks correct. I left a formal review with a couple of points to consider, the strongest being that the two new regression tests create only one version each, so they do not actually exercise the throwing `subList` path this PR is meant to guard. Straightforward to tighten if you agree.

DSingh0304:
> The core arithmetic fix is correct: widening `effectiveOffset + effectiveLimit` to `long` before `Math.min` eliminates the int overflow at the source, and clamping `offset >= 0` matches the existing `WellKnownResourceImpl` pattern. Two things worth addressing before merge — one about test fidelity and one about a behavioral change in the well-known endpoint.
> 
>     * **Regression tests do not reproduce the actual overflow path.** Both new tests create exactly one subject version. In `getSubjectVersions`, the overflowed `effectiveOffset + effectiveLimit` only becomes the `toIndex` passed to `rval.subList(...)` when `effectiveOffset < rval.size()` (the guard a few lines below the clamp). With `size == 1` and `offset == 1`, that guard is false, so `subList` is never the call site that would throw. Likewise the negative-offset case: with one version and `limit == 1`, neither pagination predicate (`effectiveOffset > 0` nor `effectiveLimit < size`) is true, so the block is skipped entirely and the endpoint returns 200 regardless of whether the fix is applied. To genuinely guard #8732, the tests should create at least 2-3 versions and use an `offset`/`limit` combination that lands inside the list while keeping the overflowed sum negative pre-fix (e.g. 3 versions, `offset=1`, `limit=Integer.MAX_VALUE`).
> 
>     * **`WellKnownResourceImpl` limit clamp changes two behaviors at once.** `Math.max(0, limit)` to `Math.max(1, Math.min(limit, 500))` both caps at 500 (an API-visible change: a caller requesting 1000 entitled agents now silently gets 500) and raises the floor from 0 to 1 (a `limit=0` request, which previously yielded an empty page, now returns one item). The 500 cap is defensible and matches `searchAgents`, but worth confirming explicitly that both edge changes are intended.
> 
>     * Minor: the `(int)` cast after `Math.min((long) ..., rval.size())` is safe because the result is bounded by `rval.size()` (an int), but a one-line comment to that effect would help future readers.

Thanks for the thorough review! All points have been addressed in the latest commit:

1. **Test fidelity**: Updated both regression tests to create 3 versions each. The assertions now genuinely exercise the subList path by ensuring `effectiveOffset(1) < rval.size()(3)`, which would have failed on the pre-fix code.
2. **Floor behavior**: Good catch! Reverted the floor back to `Math.max(0, ...) so limit=0` correctly returns an empty page again. Only the `500 cap (matching searchAgents)` was kept.
3. **Comment**: Added a brief inline comment explaining why the (int) cast is safe.

app/src/test/java/io/apicurio/registry/noprofile/ccompat/rest/v7/CCompatV7EnhancementsTest.java
DSingh0304: Fixed! Updated the regression tests to create 3 versions so `effectiveOffset(1) < rval.size()(3)` evaluates to true. The test now properly hits the subList path and would fail on the pre-fix code as intended.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
DSingh0304: Good catch! The floor change was unintentional. I've reverted it to `Math.max(0, ...) so limit=0` still returns an empty page as expected, while keeping the `500 cap` intact.

app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/SubjectsResourceImpl.java
DSingh0304: Understood! I've updated the regression tests to generate `3 versions, so size > effectiveOffset` is now true and the test explicitly exercises the subList path that would have thrown on the old code.

DSingh0304:
> Thanks @DSingh0304 — the arithmetic fix itself looks correct. I left a formal review with a couple of points to consider, the strongest being that the two new regression tests create only one version each, so they do not actually exercise the throwing `subList` path this PR is meant to guard. Straightforward to tighten if you agree.

Thanks @paoloantinori  for the thorough review! I completely agree I've just pushed an updated commit that tightens the tests and addresses all the feedback. Let me know if everything looks good now!

paoloantinori:
Thanks @DSingh0304 — the updated tests address the concern exactly. Creating 3 versions and querying with `offset=1, limit=Integer.MAX_VALUE` means `effectiveOffset(1) < rval.size()(3)` is now true, so the request actually enters the `subList` branch where the overflow used to throw; pre-fix that `Math.min(1 + Integer.MAX_VALUE, 3)` evaluated to a negative `toIndex` and blew up, post-fix the `long` widening caps it to 3. Same for the negative-offset test (`subList(-1, 1)` → `subList(0, 1)` after the `Math.max`). Both regression tests now exercise the real buggy paths rather than the early-return guard, which is what was missing.

One thing I'd flag for awareness, not as a blocker: clamping `WellKnownResourceImpl.getEntitledAgents` to `limit ≤ 500` is a visible behavior change for that endpoint — a client requesting more than 500 now silently gets 500. It matches the existing `searchAgents` convention so it's consistent, but it's worth making sure no caller depends on unbounded pages there. If you'd prefer, that could be noted in the #8732 issue thread or the changelog.

From my side the fix and tests look good.

DSingh0304:
> Thanks @DSingh0304 — the updated tests address the concern exactly. Creating 3 versions and querying with `offset=1, limit=Integer.MAX_VALUE` means `effectiveOffset(1) < rval.size()(3)` is now true, so the request actually enters the `subList` branch where the overflow used to throw; pre-fix that `Math.min(1 + Integer.MAX_VALUE, 3)` evaluated to a negative `toIndex` and blew up, post-fix the `long` widening caps it to 3. Same for the negative-offset test (`subList(-1, 1)` → `subList(0, 1)` after the `Math.max`). Both regression tests now exercise the real buggy paths rather than the early-return guard, which is what was missing.
> 
> One thing I'd flag for awareness, not as a blocker: clamping `WellKnownResourceImpl.getEntitledAgents` to `limit ≤ 500` is a visible behavior change for that endpoint — a client requesting more than 500 now silently gets 500. It matches the existing `searchAgents` convention so it's consistent, but it's worth making sure no caller depends on unbounded pages there. If you'd prefer, that could be noted in the #8732 issue thread or the changelog.
> 
> From my side the fix and tests look good.

Thanks @paoloantinori  for the review and approval! I agree about the `500 cap` for getEntitledAgents. Since it perfectly matches the searchAgents convention, it brings the endpoints into consistency. I'll drop a note in the original issue thread highlighting this behavior change for full transparency.

Also, is there anything else I need to do to get this PR merged?

paoloantinori:
Thanks @DSingh0304 — the fix and the updated regression tests look good. The `long` widening before `Math.min` in `getSubjectVersions` and the `Math.max(0, …)` offset clamp cover both the overflow and negative-index paths, and the two new tests now exercise the actual `subList` branch with 3 versions (rather than skipping it). The `limit` cap to 500 in `WellKnownResourceImpl.getEntitledAgents` also brings it in line with `searchAgents`.

From my side this is ready for a maintainer to merge.

DSingh0304:
> Thanks @DSingh0304 — the fix and the updated regression tests look good. The `long` widening before `Math.min` in `getSubjectVersions` and the `Math.max(0, …)` offset clamp cover both the overflow and negative-index paths, and the two new tests now exercise the actual `subList` branch with 3 versions (rather than skipping it). The `limit` cap to 500 in `WellKnownResourceImpl.getEntitledAgents` also brings it in line with `searchAgents`.
> 
> From my side this is ready for a maintainer to merge.

Thank you @paoloantinori for the review!! I will then wait for @carlesarnal @EricWittmann to go through the PR when they have time!!

paoloantinori:
Hi @DSingh0304 — nothing further from my side; my previous review (the long-widening + offset clamp, and the updated tests exercising the actual `subList` branch) still stands on this head. The fix looks ready for a maintainer to merge. Deferring to @carlesarnal / @EricWittmann for the final call.

DSingh0304:
Hi @EricWittmann , @paoloantinori, it looks like the lifecycle bot crashed when /accept was used earlier, so the PR is still stuck in the lifecycle/new state. This is causing the Verification Gate to instantly fail and skip all tests. Could one of you please type /accept again to reset the state? Thanks!

paoloantinori:
Hi @DSingh0304 — apologies for the delay responding. Resetting the lifecycle state (`/accept`, `/retry`) is a maintainer action that's outside the scope of what I do as an automated reviewer, so I can't run it myself. Pinging @EricWittmann since he issued the original `/accept` — Eric, could you re-run `/accept` (or `/retry`) when you get a chance to unstick this from `lifecycle/new`?

On the technical side, nothing's changed from my earlier reviews: the `long`-widening fix in `getSubjectVersions`, the `Math.max(0, …)` offset clamp, and the 500-cap in `getEntitledAgents` all look correct, and the regression tests exercise the real `subList` branch. No blockers from my side.

DSingh0304:
> Hi @DSingh0304 — apologies for the delay responding. Resetting the lifecycle state (`/accept`, `/retry`) is a maintainer action that's outside the scope of what I do as an automated reviewer, so I can't run it myself. Pinging @EricWittmann since he issued the original `/accept` — Eric, could you re-run `/accept` (or `/retry`) when you get a chance to unstick this from `lifecycle/new`?
> 
> On the technical side, nothing's changed from my earlier reviews: the `long`-widening fix in `getSubjectVersions`, the `Math.max(0, …)` offset clamp, and the 500-cap in `getEntitledAgents` all look correct, and the regression tests exercise the real `subList` branch. No blockers from my side.

Oh sure then I will wait...

DSingh0304:
Hi @EricWittmann @paoloantinori, quick question on the two failing check groups: 

1. License Compliance (FOSSA): I haven't added any dependencies (only edited pagination math logic). Are these pre-existing flags that can be approved/ignored in FOSSA?

2. Scalpel Unit Tests: These are failing fast across all shards with Maven InternalErrorException crashes. Since they are shadow/experimental tests, should I ignore them or is there anything needed on my end?
All core checks (Build, Lint, SonarCloud) are passing. Thanks!

paoloantinori:
Hi @DSingh0304 — quick take on the two check groups, as these are infra rather than code:

1. **License Compliance (FOSSA):** Your analysis is right — this PR only touches pagination math in `SubjectsResourceImpl`/`WellKnownResourceImpl` and a test, with no dependency changes, so it can't be the source of a new license flag. If FOSSA reports the same "2 issues found" on the current `main` HEAD, those are pre-existing and need to be triaged/acknowledged by a maintainer in the FOSSA dashboard, not fixed from this PR. A maintainer can approve/ignore them there.

2. **Scalpel Unit Tests:** The Maven `InternalErrorException` crashing fast across all shards is a build/runner failure, not a test-logic failure from your change. Scalpel is the experimental shadow tier, and a uniform crash across shards points at the runner/environment rather than your pagination fix. This is also a maintainer/infra call (retry vs. ignore), not something your diff needs to address.

So neither failing group blocks the code change itself from a review standpoint — my earlier review (long-widening + offset clamp, updated regression tests exercising the actual `subList` path) still stands as satisfied on this head. Unblocking the pipeline (`/accept` / `/retry` lifecycle resets, FOSSA acknowledgement) is maintainer-side, so over to @EricWittmann there.

DSingh0304:
> Hi @DSingh0304 — quick take on the two check groups, as these are infra rather than code:
> 
>     1. **License Compliance (FOSSA):** Your analysis is right — this PR only touches pagination math in `SubjectsResourceImpl`/`WellKnownResourceImpl` and a test, with no dependency changes, so it can't be the source of a new license flag. If FOSSA reports the same "2 issues found" on the current `main` HEAD, those are pre-existing and need to be triaged/acknowledged by a maintainer in the FOSSA dashboard, not fixed from this PR. A maintainer can approve/ignore them there.
> 
>     2. **Scalpel Unit Tests:** The Maven `InternalErrorException` crashing fast across all shards is a build/runner failure, not a test-logic failure from your change. Scalpel is the experimental shadow tier, and a uniform crash across shards points at the runner/environment rather than your pagination fix. This is also a maintainer/infra call (retry vs. ignore), not something your diff needs to address.
> 
> 
> So neither failing group blocks the code change itself from a review standpoint — my earlier review (long-widening + offset clamp, updated regression tests exercising the actual `subList` path) still stands as satisfied on this head. Unblocking the pipeline (`/accept` / `/retry` lifecycle resets, FOSSA acknowledgement) is maintainer-side, so over to @EricWittmann there.

Thank you for the review I will wait for @EricWittmann then!!

DSingh0304:
Hi @EricWittmann, all the checks pass now !!

carlesarnal:
### Re-review after updates

All the feedback from the previous review has been addressed:

**1. Test fidelity — fixed ✅**
Both regression tests now create 3 versions, which properly exercises the `subList` code path. Verified the math:
- `testSubjectVersionsPaginationWithLargeLimit`: with `offset=1` and `limit=Integer.MAX_VALUE` on 3 versions, `effectiveOffset(1) < rval.size()(3)` is true, so we enter the subList call. Pre-fix, `1 + Integer.MAX_VALUE` overflows to a negative `toIndex`; post-fix, the `(long)` widening safely caps it to `rval.size()`.
- `testSubjectVersionsPaginationWithNegativeOffset`: with `offset=-1` and `limit=1` on 3 versions, `Math.max(0, -1)` clamps to 0, and `subList(0, 1)` executes cleanly.

**2. WellKnownResourceImpl floor change — fixed ✅**
The floor stays at `Math.max(0, ...)`, so `limit=0` still returns an empty page as before. Only the 500 cap was added, matching the existing `searchAgents` pattern.

**3. Safe cast comment — added ✅**
Clear inline comment explaining why the `(int)` cast is safe.

**One minor suggestion (non-blocking):**
The new tests only assert `statusCode(200)`. Adding response body assertions (e.g., verifying the returned list has the expected number of versions) would make these regression tests even more robust. For example, the large-limit test with `offset=1` should return exactly 2 versions. That said, this follows the same pattern as the existing `testSubjectVersionsPagination` test in the file, so it's consistent.

Overall the fix looks correct and well-targeted. The `(long)` widening in `SubjectsResourceImpl` eliminates the overflow at the source, and the offset clamping handles the negative-offset edge case cleanly.

DSingh0304:
> ### Re-review after updates
> 
> All the feedback from the previous review has been addressed:
> 
> **1. Test fidelity — fixed ✅** Both regression tests now create 3 versions, which properly exercises the `subList` code path. Verified the math:
> 
>     * `testSubjectVersionsPaginationWithLargeLimit`: with `offset=1` and `limit=Integer.MAX_VALUE` on 3 versions, `effectiveOffset(1) < rval.size()(3)` is true, so we enter the subList call. Pre-fix, `1 + Integer.MAX_VALUE` overflows to a negative `toIndex`; post-fix, the `(long)` widening safely caps it to `rval.size()`.
> 
>     * `testSubjectVersionsPaginationWithNegativeOffset`: with `offset=-1` and `limit=1` on 3 versions, `Math.max(0, -1)` clamps to 0, and `subList(0, 1)` executes cleanly.
> 
> 
> **2. WellKnownResourceImpl floor change — fixed ✅** The floor stays at `Math.max(0, ...)`, so `limit=0` still returns an empty page as before. Only the 500 cap was added, matching the existing `searchAgents` pattern.
> 
> **3. Safe cast comment — added ✅** Clear inline comment explaining why the `(int)` cast is safe.
> 
> **One minor suggestion (non-blocking):** The new tests only assert `statusCode(200)`. Adding response body assertions (e.g., verifying the returned list has the expected number of versions) would make these regression tests even more robust. For example, the large-limit test with `offset=1` should return exactly 2 versions. That said, this follows the same pattern as the existing `testSubjectVersionsPagination` test in the file, so it's consistent.
> 
> Overall the fix looks correct and well-targeted. The `(long)` widening in `SubjectsResourceImpl` eliminates the overflow at the source, and the offset clamping handles the negative-offset edge case cleanly.

Thank you for the detailed review and for merging the PR!

Good point regarding the response body assertions I will keep that in mind to make test assertions even more thorough in future PRs. Thanks again!

PR #8936, opened 2026-07-27
fix(ui): use route matching for active tab selection

### LINKED ISSUE #8935: Active tab is incorrectly selected when artifact, group, or version identifiers start with reserved tab names
## Summary

The Artifact, Group, and Version detail pages determine the active tab by matching reserved tab names against the current URL.

As a result, if an artifact, group, or version identifier begins with one of those reserved names (for example `rules-schema` or `content-v1`), the UI incorrectly activates the corresponding tab instead of keeping the **Overview** tab selected.

This causes the wrong content to be rendered even though the URL points to the standard Overview route.

---

## Steps to reproduce

1. Create (or use) an artifact whose identifier starts with a reserved tab name, for example:
   - `rules-schema`
   - `branches-demo`
   - `contract-test`
   - `usage-report`
2. Open the artifact's Overview page, for example:

```
/explore/<groupId>/rules-schema
```

3. Observe the active tab.

The same behaviour can also be reproduced on:

- Group pages (group IDs beginning with reserved tab names)
- Version pages (version identifiers beginning with reserved tab names such as `content-v1`)

---

## Actual behaviour

The application incorrectly activates a tab such as **Rules**, **Branches**, **Contract**, or **Usage**, causing that tab's content to be displayed instead of the Overview content!

---

## Expected behaviour

The active tab should be determined by the actual route being visited.

Visiting the standard Overview route should always keep the **Overview** tab active, regardless of the artifact, group, or version identifier..

---

## Why this is a bug

The issue is triggered by valid user-defined identifiers that happen to begin with reserved route names..!

Although the URL is still the normal Overview route, the UI derives the active tab from a substring match rather than the actual route, resulting in incorrect navigation state and content!

---

## Environment

- Fresh local checkout
- Latest `main`
- Reproduced locally
- Reproducible consistently

Adii-45:
## Summary

Fixes #8935

The Artifact, Group, and Version detail pages derived the active tab by searching for reserved tab names (`rules`, `branches`, `contract`, `usage`, `content`, `references`, `documentation`) anywhere in the current pathname using substring matching!

This caused incorrect tab selection when an artifact, group, or version identifier itself started with one of those reserved names (for example, `rules-schema` or `content-v1`), even though the URL was still pointing to the Overview route!

This PR replaces substring matching with route-aware matching using `useMatch`, ensuring that the active tab is determined by the actual matched route rather than arbitrary identifier prefixes!

## Changes

- Replace `pathname.indexOf(...)` active-tab detection with `useMatch(...)` in:
  - `ArtifactPage.tsx`
  - `GroupPage.tsx`
  - `VersionPage.tsx`
- Preserve the existing routing and navigation behaviour.
- Keep the change scoped only to active tab selection.

## Verification

Tested by:

- Opening Overview pages whose identifiers begin with reserved tab names
- Verifying that the correct Overview tab remains selected
- Verifying tab navigation still works correctly
- Running TypeScript and lint checks with no errors introduced by this change

## Notes

This PR intentionally leaves the existing `artifacts` branch in `GroupPage` unchanged, since it is unrelated to the reported bug and keeping the fix narrowly scoped reduces review risk..!

### DIFF
--- a/ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx
+++ b/ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx
@@ -2,7 +2,7 @@ import { FunctionComponent, useEffect, useState } from "react";
 import "./ArtifactPage.css";
 import { LoaderGuard, newLoaderGuard } from "@utils/loader.utils.ts";
 import { Breadcrumb, BreadcrumbItem, PageSection, Tab, Tabs } from "@patternfly/react-core";
-import { Link, useLocation, useParams } from "react-router";
+import { Link, useMatch, useParams } from "react-router";
 import { EXPLORE_PAGE_IDX, PageDataLoader, PageError, PageErrorHandler, PageProperties, toPageError } from "@app/pages";
 import {
     ChangeOwnerModal,
@@ -71,16 +71,19 @@ export const ArtifactPage: FunctionComponent<PageProperties> = () => {
     const logger: LoggerService = useLoggerService();
     const groups: GroupsService = useGroupsService();
     const { groupId, artifactId }= useParams();
-    const location = useLocation();
+    const rulesMatch = useMatch("/explore/:groupId/:artifactId/rules");
+    const branchesMatch = useMatch("/explore/:groupId/:artifactId/branches");
+    const contractMatch = useMatch("/explore/:groupId/:artifactId/contract");
+    const usageMatch = useMatch("/explore/:groupId/:artifactId/usage");
 
     let activeTabKey: string = "overview";
-    if (location.pathname.indexOf("/rules") !== -1) {
+    if (rulesMatch) {
         activeTabKey = "rules";
-    } else if (location.pathname.indexOf("/branches") !== -1) {
+    } else if (branchesMatch) {
         activeTabKey = "branches";
-    } else if (location.pathname.indexOf("/contract") !== -1) {
+    } else if (contractMatch) {
         activeTabKey = "contract";
-    } else if (location.pathname.indexOf("/usage") !== -1) {
+    } else if (usageMatch) {
         activeTabKey = "usage";
     }
 
--- a/ui/ui-app/src/app/pages/group/GroupPage.tsx
+++ b/ui/ui-app/src/app/pages/group/GroupPage.tsx
@@ -2,7 +2,7 @@ import { FunctionComponent, useEffect, useState } from "react";
 import "./GroupPage.css";
 import { LoaderGuard, newLoaderGuard } from "@utils/loader.utils.ts";
 import { Breadcrumb, BreadcrumbItem, PageSection, PageSectionVariants, Tab, Tabs } from "@patternfly/react-core";
-import { Link, useLocation, useParams } from "react-router";
+import { Link, useLocation, useMatch, useParams } from "react-router";
 import {
     EXPLORE_PAGE_IDX,
     GroupOverviewTabContent,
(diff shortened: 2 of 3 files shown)


Adii-45:
Hi @EricWittmann @carlesarnal @paoloantinori - whenever you have a moment, could you please review this PR?

This is a small, focused fix for #8935 that replaces substring-based active tab detection with route-aware matching, preventing incorrect tab selection for identifiers beginning with reserved tab names while preserving existing behaviour..

Thanks!

paoloantinori:
Thanks for the ping — read through the diff. Switching from `pathname.indexOf` to `useMatch` is the right approach: substring matching was clearly fragile whenever an artifact/group/version identifier happened to begin with a reserved word, and route-aware matching fixes that cleanly.

One thing worth double-checking before this lands. `useMatch("/explore/:groupId/:artifactId/rules")` matches that **exact** path, whereas the previous `indexOf("/rules")` matched any path *containing* `/rules`. If any of these segments have deeper sub-routes (e.g. a per-rule detail page like `/explore/:groupId/:artifactId/rules/:ruleName`, or a branch detail view), the active tab would no longer highlight on those deeper views. If they're all terminal tab routes, there's nothing to worry about and this is strictly an improvement — just flagging the one behavioral difference since it's the only place the new code isn't a drop-in replacement for the old.

The scoping decision to leave `GroupPage`'s `/artifacts` check untouched looks reasonable for keeping the diff narrow.

Adii-45:
Thanks for pointing this out! I took another look at the route definitions and navigation flow..

I checked all of the affected routes:

- `rules`
- `contract`
- `usage`
- `content`
- `references`
- `documentation`

None of these have nested/detail routes today—they are all terminal routes..

The only nested routes I found are under `branches` (`/branches/:branchId` and `/branches/:branchId/versions`), but those render `BranchPage` rather than `ArtifactPage`, so the tab-selection logic changed in this PR is never executed for those URLs!

So for every currently reachable route, the new `useMatch` implementation is behaviourally equivalent to the previous `pathname.indexOf(...)` logic, while avoiding the false-positive matching that caused the reported bug..

I don't think an additional wildcard match is needed for the current route structure, but I'm happy to adjust the implementation if you'd prefer to make it future-proof for potential nested routes!

paoloantinori:
Thanks for walking through the route tree, @Adii-45 — that's exactly the check I was hoping for. If those tab routes (`rules`, `contract`, `usage`, `content`, `references`, `documentation`) are all terminal today with no nested detail routes underneath them, then the exact-path `useMatch` calls are safe and won't swallow deeper routes. That resolves my main concern.

One small thing worth keeping in mind for the future, not a blocker here: if any of those tabs ever gain a nested detail route (e.g. `/rules/:ruleId`), this exact-match style will need the trailing `/*` wildcard, or the parent tab won't stay highlighted on the detail screen. Easy to handle when/if it comes up.

From my side this looks good to merge.

jsenko (commented):
## Code Review — 0 findings

Clean, well-scoped fix. The `useMatch()` approach correctly ties tab selection to route structure rather than substring matching on the pathname. Route patterns match the existing navigation URLs in `handleTabClick`.

The LoaderGuard changes (wrapping async callbacks, cancellation on cleanup) are technically unrelated to the tab fix but are a good improvement — they prevent state updates on unmounted components. No issues with either change.

One note: GroupPage still uses `location.pathname.indexOf("/artifacts")` for the artifacts tab. The PR description explains this is intentionally left as-is to keep the scope narrow, which is fine.

Adii-45:
@EricWittmann @jsenko Hi! All checks are now passing, and the review feedback has been addressed... This PR should be ready for final review and merge whenever you have a chance. Thanks!

EricWittmann:
Thanks for the contribution @Adii-45! Clean fix — using `useMatch` is the right approach and much more robust than substring matching. Merged.

Issue #12, opened 2019-07-30
Compatibility Rule: avro


EricWittmann:
Implement the compatibility rule for avro schemas.  This can be lifted from Perspicuus most likely.

PR #1000, opened 2020-11-10
Added a Kafka+SQL storage variant

EricWittmann:

Issue #14, opened 2019-07-30
Compatibility Rule: JSON Schema


EricWittmann:
Implement the compatibility rule for JSON Schema schemas. We'll need to really analyze the avro version to implement similar logic for JSON Schema.

EricWittmann:
There seems to be some code written to do this here:  https://github.com/confluentinc/schema-registry/pull/1289

Issue #13, opened 2019-07-30
Compatibility Rule: protobuff


EricWittmann:
Implement the compatibility rule for protobuff schemas.  We'll need to really analyze the avro version to implement similar logic for protobuff.

PR #8938, opened 2026-07-27
ci: nudge and flag PRs missing an issue link, detect duplicate PRs

### LINKED ISSUE #8937: Enforce issue-linking on PRs and detect duplicate PRs against the same issue
## Description

Contributors sometimes open PRs without referencing the issue they're fixing, and occasionally two contributors submit separate PRs for the same issue. Maintainers review both before realizing they're duplicates, and whoever's PR is reviewed second has their work wasted once the first is merged. There's currently no enforcement point for the CLAUDE.md checklist requirement that a PR link its issue.

## Proposal

Building on the existing PR-lifecycle orchestrator (`.github/scripts/pr-lifecycle.js`, `handlePrOpened`) and the `hasLinkedIssue` check already added there:

1. **Require an issue reference on every PR.**
   If `hasLinkedIssue(pr)` is false, the bot already posts a nudge comment on open — extend this to also apply the existing `lifecycle/waiting-on-author` label (or a new dedicated label, e.g. `lifecycle/needs-issue-link`) so it's visibly tracked as a blocking item rather than a one-off comment that can be missed. Re-check on `synchronize` (when the author edits the description) and clear the label once a link is added.

2. **Detect duplicate PRs against the same issue.**
   When a PR does reference an issue, search open PRs for others referencing the same issue number. If a match is found, comment on the new PR (and optionally the earlier one) linking the overlapping PR(s), so reviewers see the conflict before investing review time.

## Why

Turns an already-documented contributor requirement (link the issue) into something enforced by tooling instead of relying on manual review catching it, and prevents duplicate review effort on overlapping PRs.

## Scope

Labeling/commenting only, non-blocking to merge mechanics — no auto-closing PRs. Maintainers still decide how to resolve overlaps.

malladinagarjuna2:
fixes #8937 
## Summary
Extends the PR-lifecycle bot to flag PRs that don't reference an issue and to detect duplicate PRs opened against the same issue.

## Changes
- New `lifecycle/needs-issue-link` label, applied/cleared based on issue references in the PR body (checked on open and on push)
- Duplicate-PR detection: cross-links PRs that reference the same issue number

## Test plan
- [x] Verified locally with a mocked GitHub API harness
- [ ] Manual check on a real PR after merge

### DIFF
--- a/.github/scripts/pr-lifecycle.js
+++ b/.github/scripts/pr-lifecycle.js
@@ -550,14 +550,14 @@ async function findDuplicatePrs(github, owner, repo, pr, core) {
 
 // Avoids re-pinging a PR that already has a recent "possible duplicate" note,
 // so several PRs opened in quick succession against the same issue don't pile
-// up repeated comments on the earliest one. Single-page fetch (not paginated) —
-// comments come back oldest-first, so the most recent 100 covers any plausibly
-// recent note without walking a long-lived PR's entire comment history.
+// up repeated comments on the earliest one. Single-page fetch (not paginated),
+// sorted newest-first, so the top 10 are the actual most recent comments
+// without walking a long-lived PR's entire comment history.
 async function hasRecentDuplicateNote(github, owner, repo, prNumber) {
   const { data: comments } = await github.rest.issues.listComments({
-    owner, repo, issue_number: prNumber, per_page: 100,
+    owner, repo, issue_number: prNumber, per_page: 10, sort: 'created', direction: 'desc',
   });
-  return comments.slice(-10).some(c => c.body?.includes('**Possible duplicate:**'));
+  return comments.some(c => c.body?.includes('**Possible duplicate:**'));
 }
 
 async function handlePrOpened({ github, context, core }) {
--- a/.github/scripts/pr-lifecycle.js
+++ b/.github/scripts/pr-lifecycle.js
@@ -510,15 +510,21 @@ function extractIssueNumbers(pr) {
   // matching GitHub's own auto-close semantics. A bare "#123" is too permissive:
   // it also matches list markers, code snippets, and passing prose references,
   // which would both mask real missing-link cases and produce false duplicates.
+  // The keyword form also accepts a trailing comma-separated list, e.g.
+  // "Fixes #12, #13, #14" — GitHub itself closes all of them.
   const patterns = [
-    /\b(?:close[sd]?|fix(?:es|ed)?|resolve[sd]?)\b\s*:?\s*#(\d+)/gi,
+    /\b(?:close[sd]?|fix(?:es|ed)?|resolve[sd]?)\b\s*:?\s*#(\d+)(?:\s*,\s*#(\d+))*/gi,
     /github\.com\/[^/\s]+\/[^/\s]+\/issues\/(\d+)/gi,
   ];
   const numbers = new Set();
   for (const re of patterns) {
     let match;
     while ((match = re.exec(body))) {
-      numbers.add(Number(match[1]));
+      // match[0] is the whole "Fixes #12, #13, #14" span; re-scan it for every
+      // "#N" since a single capture group can't repeat across a JS match.
+      for (const n of match[0].matchAll(/#(\d+)/g)) {
+        numbers.add(Number(n[1]));
+      }
     }
   }
   return [...numbers];
@@ -544,9 +550,11 @@ async function findDuplicatePrs(github, owner, repo, pr, core) {
 
 // Avoids re-pinging a PR that already has a recent "possible duplicate" note,
 // so several PRs opened in quick succession against the same issue don't pile
-// up repeated comments on the earliest one.
+// up repeated comments on the earliest one. Single-page fetch (not paginated) —
+// comments come back oldest-first, so the most recent 100 covers any plausibly
+// recent note without walking a long-lived PR's entire comment history.
 async function hasRecentDuplicateNote(github, owner, repo, prNumber) {
-  const comments = await github.paginate(github.rest.issues.listComments, {
+  const { data: comments } = await github.rest.issues.listComments({
     owner, repo, issue_number: prNumber, per_page: 100,
   });
   return comments.slice(-10).some(c => c.body?.includes('**Possible duplicate:**'));
--- a/.github/scripts/pr-lifecycle.js
+++ b/.github/scripts/pr-lifecycle.js
@@ -502,14 +502,17 @@ async function countOpenPrsByAuthor(github, owner, repo, author, excludePr) {
 }
 
 // Extracts referenced issue numbers from a PR body: "closes #123", "fixes: #123",
-// GitHub issue URLs, or a bare "#123" reference. Regexes are built fresh per call
-// since they carry `g`-flag lastIndex state.
+// or a GitHub issue URL. Regexes are built fresh per call since they carry
+// `g`-flag lastIndex state.
 function extractIssueNumbers(pr) {
   const body = pr.body || '';
+  // Only the closing-keyword and issue-URL forms count as a "linked" issue,
+  // matching GitHub's own auto-close semantics. A bare "#123" is too permissive:
+  // it also matches list markers, code snippets, and passing prose references,
+  // which would both mask real missing-link cases and produce false duplicates.
   const patterns = [
     /\b(?:close[sd]?|fix(?:es|ed)?|resolve[sd]?)\b\s*:?\s*#(\d+)/gi,
     /github\.com\/[^/\s]+\/[^/\s]+\/issues\/(\d+)/gi,
-    /#(\d+)/g,
   ];
   const numbers = new Set();
   for (const re of patterns) {
@@ -526,16 +529,27 @@ function hasLinkedIssue(pr) {
 }
 
 // Finds other open PRs referencing at least one of the same issue numbers.
-async function findDuplicatePrs(github, owner, repo, pr) {
+async function findDuplicatePrs(github, owner, repo, pr, core) {
   const issueNumbers = extractIssueNumbers(pr);
   if (!issueNumbers.length) return [];
   const openPrs = await github.paginate(github.rest.pulls.list, {
     owner, repo, state: 'open', per_page: 100,
   });
+  core?.info(`Scanned ${openPrs.length} open PR(s) for duplicates of issue(s) ${issueNumbers.join(', ')}`);
   return openPrs.filter(p =>
     p.number !== pr.number &&
     extractIssueNumbers(p).some(n => issueNumbers.includes(n))
   );
+}
+
+// Avoids re-pinging a PR that already has a recent "possible duplicate" note,
+// so several PRs opened in quick succession against the same issue don't pile
+// up repeated comments on the earliest one.
+async function hasRecentDuplicateNote(github, owner, repo, prNumber) {
+  const comments = await github.paginate(github.rest.issues.listComments, {
+    owner, repo, issue_number: prNumber, per_page: 100,
+  });
+  return comments.slice(-10).some(c => c.body?.includes('**Possible duplicate:**'));
 }
 
 async function handlePrOpened({ github, context, core }) {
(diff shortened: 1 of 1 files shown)


malladinagarjuna2:
@paoloantinori @EricWittmann  
could you please give a review

.github/scripts/pr-lifecycle.js
```diff
@@ -499,6 +501,43 @@ async function countOpenPrsByAuthor(github, owner, repo, author, excludePr) {
   return prs.filter(p => p.user.login === author && p.number !== excludePr);
 }
 
+// Extracts referenced issue numbers from a PR body: "closes #123", "fixes: #123",
+// GitHub issue URLs, or a bare "#123" reference. Regexes are built fresh per call
+// since they carry `g`-flag lastIndex state.
+function extractIssueNumbers(pr) {
+  const body = pr.body || '';
+  const patterns = [
```
paoloantinori, line 515: This bare `/#(\d+)/g` matches any `#<number>` in the body — including prose like "similar to #1000", list markers, or code snippets. That both defeats the `needs-issue-link` check (a PR with a stray `#N` is treated as linked) and can trigger false duplicate detection. GitHub only auto-links issues on the `closes|fixes|resolves` keywords, so I'd remove this third pattern and keep just the keyword + URL regexes.

.github/scripts/pr-lifecycle.js
```diff
@@ -499,6 +501,43 @@ async function countOpenPrsByAuthor(github, owner, repo, author, excludePr) {
...
+  }
+  return [...numbers];
+}
+
+function hasLinkedIssue(pr) {
+  return extractIssueNumbers(pr).length > 0;
+}
+
+// Finds other open PRs referencing at least one of the same issue numbers.
+async function findDuplicatePrs(github, owner, repo, pr) {
```
paoloantinori: This paginates every open PR on each `opened` event, then runs `extractIssueNumbers` over each body. Fine at current scale, but consider a `core.info` log of the scanned count so the cost is visible if the repo grows. Non-blocking.

paoloantinori (commented):
Nice extension to the lifecycle bot — the missing-issue nudge and duplicate detection are genuinely useful. One substantive concern on the issue-number extraction, plus a couple of smaller notes.

- **The bare `/#(\d+)/g` pattern (line ~509) is too permissive and undercuts the feature's purpose.** It matches *any* `#<digits>` in the PR body — a markdown heading fragment, a list like "step #1", a code snippet containing `#42`, or a prose reference like "similar to #1000". Two consequences: (1) a PR that mentions a stray `#N` will be treated as having a linked issue and never get the `lifecycle/needs-issue-link` label; (2) two PRs that both mention the same `#N` in passing will be flagged as duplicates of each other. GitHub itself only auto-closes issues on the `closes|fixes|resolves` keywords, so the bare pattern is inconsistent with how GitHub defines a "linked" issue. I'd drop the third pattern and rely on the keyword regex + the `issues/<n>` URL regex — that aligns with GitHub's own semantics and avoids false positives.
- **`findDuplicatePrs` paginates *all* open PRs on every `opened` event** (line ~529). For apicurio's current PR volume this is fine, but it's an O(open PRs) API walk plus a regex pass over every PR body on each new PR. Worth a one-line `core.info` noting how many PRs were scanned, so it's visible in the run log if the repo grows. Not blocking.
- **The duplicate cross-link comment is posted to the *other* PR without any guard against repeated pings.** If three new PRs reference the same issue in quick succession, the original PR gets three "possible duplicate" comments in a row. A light dedup (check the last N comments for an existing duplicate note) would keep noise down, but this is a minor polish item.
- Positive: the `g`-flag `lastIndex` gotcha is handled correctly by rebuilding the regexes per call, and the `synchronize` handler correctly guards label churn with `hasLabel` checks.

paoloantinori:
Gave it a look — posted a review. The main thing is the bare `#(\d+)` pattern in `extractIssueNumbers`, which is a bit too permissive and can both defeat the missing-issue-link check and produce false duplicate matches. The keyword + URL regexes alone would be closer to how GitHub itself defines a linked issue. The rest looks reasonable to me.

malladinagarjuna2:
Thanks for the thorough review @paoloantinori ,
all the suggestions are implemented.

**Also caught one more gap while re-reviewing: auto-accepted authors (maintainers, renovate[bot]) were getting the issue-link nudge/label too, which would've spammed every dependency-bump PR. Scoped that out as well, they're now exempt from the issue-link check, though duplicate detection still applies to everyone.**

.github/scripts/pr-lifecycle.js
malladinagarjuna2: yeah!! it is really a big blunder! idk  how i did'nt think about it

.github/scripts/pr-lifecycle.js
malladinagarjuna2: considering it, as the repo looks strong, will be useful in future.

malladinagarjuna2:
few more things which can be added
wanted to clarify first with you!! @paoloantinori 

Multi-issue lists (Fixes #12, #13), (though these are very infrequent ) only #12 gets picked up since each number needs its own keyword. Want me to extend the regex to handle comma-separated lists, or is a one-line comment noting the limitation enough for now?

hasRecentDuplicateNote pagination , it currently pages through the PR's entire comment history just to check the last 10. Worth capping the fetch (e.g. per_page: 10 with no further pagination) to avoid the unbounded call on long-lived PRs, or is this low-traffic enough not to matter?

Duplicate detection scope, right now it only runs on PR open, not on synchronize. So a PR that adds its issue link via a later push (satisfying the missing-link check) never gets re-checked for duplicates against other PRs. Should I wire duplicate detection into handlePrSynchronize too, or leave this as a known/accepted gap?

malladinagarjuna2:
@EricWittmann @carlesarnal  @jsenko would also love a review from all

paoloantinori:
Good questions — happy to weigh in on both.

**1. Multi-issue lists (`Fixes #12, #13`)**

Worth a small extension rather than just a comment. The current regex grabs exactly one number per closing keyword, so `Fixes #12, #13` silently drops `#13` from duplicate detection — which is exactly the case where getting it right matters most (two PRs both closing a multi-issue cluster). It's also cheap to fix and keeps the behaviour aligned with what GitHub itself auto-closes.

A minimal pattern that keeps the closing-keyword guard but captures a trailing comma-separated list:

```js
// Match "Fixes #12", and also "Fixes #12, #13, #14"
const closeRe = /\b(?:close[sd]?|fix(?:es|ed)?|resolve[sd]?)\b\s*:?\s*#(\d+)(?:\s*,\s*#(\d+))*/gi;
```

You'd then walk `match[0]` and re-scan its tail for the extra numbers (since a single JS match group can't repeat), or just run a second pass `/#(\d+)/g` over the matched span. Either way it's a few lines and it makes the duplicate check honest. Given the rarity you noted, I wouldn't go further than this (no `and`, no `&`, no per-line variants) — the comma form covers the realistic cases.

**2. `hasRecentDuplicateNote` pagination**

Agreed it's wasteful as written — `github.paginate(...)` walks the *entire* comment history of the PR just to inspect the last 10. For a long-lived PR that's a lot of API calls for nothing. A couple of options, in increasing order of cleanliness:

- **Single-page cap**: drop `github.paginate` and call `github.rest.issues.listComments` once with `{ per_page: 100 }`. Comments come back ascending, so the last 100 is the recent tail — more than enough window to spot a "Possible duplicate" note that was just posted. Cheapest change, bounds cost at one API call.
- **`since` param**: `listComments` supports a `since` (ISO timestamp) filter. If you stash the timestamp of the last duplicate ping you posted (e.g. in a Gist, artifact, or a hidden label) you can fetch only comments newer than it. Cleanest but adds state to track, so I'd only do this if the single-page cap ever proves insufficient.

I'd ship the single-page cap now and leave the `since` approach as a noted future optimisation. The current `slice(-10)` on top of either approach stays as the final guard.

Both of these are genuinely small — your call on whether to fold them into this PR or follow up. I'd lean toward folding in #1 (it changes observable behaviour) and treating #2 as a quick nice-to-have if you have the bandwidth.

Nachiket-Roy:
What happens when there's no directly linked issue? For example, if someone is working on Phase A or Phase B of a larger issue?

paoloantinori:
Good question @Nachiket-Roy — that's a genuine gap worth naming. `extractIssueNumbers` only recognizes the GitHub closing-keyword forms (`Fixes` / `Closes` / `Resolves #N`) and issue URLs, deliberately matching GitHub's own auto-close semantics (the rationale is in the earlier thread: a bare `#N` is too permissive). So a PR framed as "Phase A of #1234" without a closing keyword won't be detected as linked and would pick up `lifecycle/needs-issue-link`.

The intent is that the label is a nudge, not a blocker — a maintainer who sees it on a PR that's clearly part of a larger effort can just dismiss it. But if sub-task / phase work is common in this repo, it might be worth a small follow-up: extend the patterns to recognize `Part of #N` as a *soft* link (counted for the missing-link check, but excluded from duplicate detection, since multiple PRs can legitimately be "part of" the same parent). I wouldn't block this PR on it — the core detection is solid for the common `Fixes #N` case — but it's a reasonable item to track.

malladinagarjuna2:
hey @paoloantinori  addresed the oother 2 suggestions, kindly review!!

paoloantinori:
Thanks @malladinagarjuna2 — re-read the current head (`3cafbc9`). Both addressed points look right:

1. **Multi-issue lists** — the closing-keyword regex now spans `Fixes #12, #13, #14` and re-scans the whole match with `matchAll(/#(\d+)/g)`, so each number is captured. Clean approach.
2. **Auto-accepted scoping** — the `!isAutoAccepted(...)` guard now wraps both `handlePrOpened` and `handlePrSynchronize`, and the sync path also strips a stale label if the author changed since open. Good.

One new, minor observation on `hasRecentDuplicateNote`: GitHub's `listComments` defaults to `direction: 'asc'`, so page 1 (`per_page: 100`) returns the **oldest** 100 comments, not the most recent. On a PR with 100+ comments, `.slice(-10)` would therefore check comments 91–100 (old) instead of the latest ones, defeating the "already noted, skip re-ping" guard. Passing `direction: 'desc'` and slicing the first 10 would make the window actually recent. Low impact in practice (only affects dedup of duplicate-ping comments on very chatty PRs), but worth tightening if you touch this again.

LGTM otherwise.

malladinagarjuna2:
nice catch @paoloantinori, i have implemented the fix, thanks   for the review.
kindly review!!

Nachiket-Roy:
Hey @malladinagarjuna2, one of your commits is missing a DCO sign-off

paoloantinori:
Re-read the current head — the latest fix looks right, and the multi-issue + scoping work from the earlier rounds still holds up. The `extractIssueNumbers` re-scan, the `NEEDS_ISSUE_LINK` add/remove in `handlePrSynchronize`, and the auto-accepted-author bypass all line up with what we discussed.

One heads-up, not a blocker: GitHub reports this branch as **behind main** (`mergeable_state: behind`), so it'll need a rebase before merge. The open DCO sign-off flag @Nachiket-Roy raised on one of the commits is in the same "sort before merge" bucket.

Otherwise this looks good to me — deferring final merge call to @EricWittmann / @carlesarnal.

malladinagarjuna2:
thanks @Nachiket-Roy  for pointing the non signed commit 
thanks @paoloantinori  for the supersonic review 
waiting for  @EricWittmann  and @carlesarnal  to give review on this

paoloantinori:
You're welcome @malladinagarjuna2! Happy to help — the PR looks in good shape from my side after the last rounds. Deferring to @EricWittmann / @carlesarnal for the final maintainer sign-off.

malladinagarjuna2:
@jsenko  kindly review the pr

jsenko:
Hi @malladinagarjuna2, thanks for the work on this — the implementation is solid, especially the issue extraction regex handling and the duplicate detection with debounced comments.

We've just created an epic (#9001) to improve the PR lifecycle in several ways, and one of the steps (#9003) covers exactly what this PR does: issue link checking and duplicate PR detection.

However, the first step of the epic (#9002) simplifies the lifecycle state machine significantly — removing the `wip`/smoke test stage entirely and shrinking from 4 states to 3. This will change `handlePrOpened` and `handlePrSynchronize` in `pr-lifecycle.js`, which are the same integration points your PR touches.

Would you be willing to close this PR for now and wait for #9002 to land? Once the simplified lifecycle is in place, you could re-submit your work adapted to the new code. Your core utility functions (`extractIssueNumbers`, `hasLinkedIssue`, `findDuplicatePrs`, `hasRecentDuplicateNote`) should be reusable as-is — only the integration points will need adjustment to match the new state machine.

If you're interested, we can assign you to #9003 so the work is reserved for you. Let us know what you think!

malladinagarjuna2:
yes @jsenko  I can work on #9003 in fact i am ready to work on #9002 #9001 if it is not assigned to anyone and if possible please assign the whole epic issue to me i would definitely be able to do it .

malladinagarjuna2:
for now i will close this pr

PR #8940, opened 2026-07-28
fix(ui): keep Create Artifact type dropdown within modal (#8918)

### LINKED ISSUE #8918: UI: Create Artifact type dropdown overflows outside modal
In the **Create Artifact** dialog, the **Type** dropdown is not positioned correctly when opened.

  The dropdown extends beyond the bottom of the modal and viewport instead of remaining visible, flipping
  above the field, or becoming scrollable. While scrolling the dialog, the open dropdown can also appear
  detached from its trigger.

  ## Steps to reproduce

  1. Open the Registry UI.
  2. Select **Create Artifact**.
  3. Open the **Type** dropdown.
  4. Scroll through the available types or scroll the dialog.

  ## Actual behavior

  The dropdown overflows outside the dialog and is clipped by the viewport. Its position may no longer align
  with the **Type** field.

  ## Expected behavior

  The dropdown should stay anchored to its trigger and remain fully accessible. When there is not enough
  space below, it should flip upward or use a scrollable menu.

  ## Video

https://github.com/user-attachments/assets/50264595-aa31-433f-af2e-e5873fe1eb8d

Faisal77666:
Fixes #8918

  ## What changed

  Replaced the Create Artifact **Type** dropdown with a scrollable
  PatternFly select.

  The dropdown now:
  - limits its height to the available space below the toggle;
  - remains scrollable when many artifact types are available;
  - stays anchored to the toggle;
  - flips upward only when there is not enough space below.

  ## Before

  The dropdown could overflow outside the modal and viewport.

[raw-recording (4).webm](https://github.com/user-attachments/assets/5cd8e601-0fe9-4b50-8143-47d6ab9a8605)

 

  ## After

  The dropdown remains contained and scrollable in both normal and small
  viewport sizes.

[raw-recording (5).webm](https://github.com/user-attachments/assets/90ead707-6879-45ab-bc3f-d04175086235)

### DIFF
--- a/ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
+++ b/ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, useEffect, useState } from "react";
+import { FunctionComponent, MouseEvent, Ref, useEffect, useState } from "react";
 import "./CreateArtifactModal.css";
 import {
     FileUpload,
@@ -9,6 +9,12 @@ import {
     GridItem,
     HelperText,
     HelperTextItem,
+    Divider,
+    MenuToggle,
+    MenuToggleElement,
+    Select,
+    SelectList,
+    SelectOption,
     Tab,
     Tabs,
     TabTitleText,
@@ -22,7 +28,7 @@ import {
     Modal
 } from "@patternfly/react-core/deprecated";
 import { CreateArtifact } from "@sdk/lib/generated-client/models";
-import { If, ObjectSelect, UrlUpload } from "@apicurio/common-ui-components";
+import { If, UrlUpload } from "@apicurio/common-ui-components";
 import { ExclamationCircleIcon } from "@patternfly/react-icons";
 import { UrlService, useUrlService } from "@services/useUrlService.ts";
 import { ArtifactTypesService, useArtifactTypesService } from "@services/useArtifactTypesService.ts";
@@ -113,6 +119,8 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> = 
     const [artifactTypes, setArtifactTypes] = useState<any[]>([]);
     const [artifactTypeOptions, setArtifactTypeOptions] = useState<ArtifactTypeItem[]>([]);
     const [selectedType, setSelectedType] = useState<ArtifactTypeItem>(DEFAULT_ARTIFACT_TYPE);
+    const [isArtifactTypeSelectOpen, setArtifactTypeSelectOpen] = useState(false);
+    const [artifactTypeMenuHeight, setArtifactTypeMenuHeight] = useState(240);
     const [artifactLabels, setArtifactLabels] = useState<ArtifactLabel[]>([]);
     const [contentTabKey, setContentTabKey] = useState(0);
     const [contentIsLoading, setContentIsLoading] = useState(false);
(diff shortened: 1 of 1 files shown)


ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
```diff
@@ -418,16 +449,36 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> =
...
+                                }}
+                            >
+                                <SelectList>
+                                    {artifactTypeOptions.map((item, index) => item.isDivider ? (
+                                        <Divider component="li" key={`divider-${index}`} />
+                                    ) : (
+                                        <SelectOption
+                                            key={item.value}
+                                            value={item}
+                                            isSelected={item === selectedType}
```
paoloantinori: This comparison is by reference (`===`). If `DEFAULT_ARTIFACT_TYPE` and the "Auto-Detect" entry in `artifactTypeOptions` are not the same object reference, no option will show as selected on first open. If you can't guarantee the references are shared, `item.value === selectedType.value` would be more robust.

ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
```diff
@@ -138,6 +146,29 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> =
...
+        }
+        setArtifactTypeSelectOpen(false);
+    };
+
+    const artifactTypeToggle = (toggleRef: Ref<MenuToggleElement>) => (
+        <MenuToggle
+            ref={toggleRef}
+            className="menu-toggle"
+            onClick={(event) => {
+                const availableHeight = window.innerHeight - event.currentTarget.getBoundingClientRect().bottom - 16;
```
paoloantinori: Minor: `availableHeight` here measures the space *below* the toggle, but with `enableFlip: true` the menu can open upward. When it flips up, this height (computed for the small space below) may leave the menu smaller than the space above allows. Not a correctness issue — `preventOverflow` keeps it on-screen — but if flipped-up is common on small viewports, sizing against the actual opening side would reclaim some room.

paoloantinori (commented):
Nice fix — swapping `ObjectSelect` for a PatternFly `Select` with `maxMenuHeight` + flip/overflow handling cleanly addresses the overflow reported in #8918, and the recordings confirm the behavior. Two small things worth double-checking below; neither is a blocker.

- **Selected-state comparison** — `isSelected={item === selectedType}` relies on reference equality. `selectedType` is initialized to `DEFAULT_ARTIFACT_TYPE`; if the "Auto-Detect" entry in `artifactTypeOptions` is constructed as a fresh object rather than reusing that constant, no option will render the selected checkmark on first open (the toggle label is still correct, so it's a visual-only gap). The previous `ObjectSelect` may have compared by value. Worth confirming the references match, or switching to `item.value === selectedType.value`.
- **`enableFlip` vs. height measured from below** — `artifactTypeMenuHeight` is computed from the space *below* the toggle, but `enableFlip: true` can render the menu upward when room below is scarce. In that flipped case the height was sized for the (small) lower space, so the menu may be smaller than the space above would allow. `preventOverflow` still keeps it on-screen, so this is purely a sizing optimization, not a correctness concern.

Faisal77666:
> Minor: `availableHeight` here measures the space _below_ the toggle, but with `enableFlip: true` the menu can open upward. When it flips up, this height (computed for the small space below) may leave the menu smaller than the space above allows. Not a correctness issue — `preventOverflow` keeps it on-screen — but if flipped-up is common on small viewports, sizing against the actual opening side would reclaim some room.

@paoloantinori Thanks for the review! I addressed both comments:Changed the selected-state comparison to item.value ===selectedType.value, so it no longer depends on object-reference identity.Updated the menu-height calculation to consider the available space above and below the toggle. It keeps the menu below when there is enough
room; otherwise, it allows the flipped-up menu to use the larger space above.

paoloantinori:
Thanks @Faisal77666 — the updated height logic addresses the comment. With

```js
const availableHeight = availableBelow >= 160 ? availableBelow : Math.max(availableBelow, availableAbove);
```

when there's enough room below (≥160px) the menu sizes against the space below, and when it's tight it falls back to the larger of the two sides — so a flipped-up menu no longer gets shrunk to the small below-space. Together with `preventOverflow` keeping it on-screen, that's the behavior I was after. Looks good to me.

Faisal77666:
Thank you for the review and guidance, @paoloantinori !

paoloantinori:
You're welcome — happy to help! 👍

Faisal77666:
@EricWittmann  please review pr ready to merge

Faisal77666:
@EricWittmann ready to merge

EricWittmann:
This is a good incremental fix to the problem.  I suppose the better UX would be to have a scroll bar within the dropdown menu itself.  But this is better than what we had before!  :+1:

EricWittmann:
Thanks for the contribution @Faisal77666! Nice fix with good before/after recordings to demonstrate the improvement. Merged.

PR #8950, opened 2026-07-28
fix(ui): add path="*" to Not Found route so unknown URLs render 404 page

### LINKED ISSUE #8931: UI: unknown URLs render a blank page because the Not Found route is missing path="*"
## Summary

In `ui-app`, the fallback route that should render the Not Found page is declared without a `path`, so it never matches. Navigating to any unknown URL renders an empty content area instead of the Not Found page.

## Affected code

`ui/ui-app/src/app/MainPageWithAuth.tsx`

```tsx
<Routes>
    <Route path="/" element={ <RootRedirectPage /> } />
    ...
    <Route element={ <NotFoundPage /> } />   // no path, no index
</Routes>
```

In react-router 7.x, a `<Route>` with neither a `path` nor `index` is a pathless layout route that only renders when one of its child routes matches. This route has no children, so it never matches anything. `NotFoundPage` is referenced nowhere else, and there is no `path="*"` route, so any unmatched URL matches no route and `<Routes>` renders nothing.

## Steps to reproduce

1. Run the UI.
2. Navigate to a URL that does not match any route, for example `/does-not-exist` or `/explore/g/a/versions/1/nonsense`.

## Expected

The Not Found page is shown.

## Actual

The masthead renders but the content area is blank. The Not Found page is never displayed.

## Fix

Give the fallback route the catch-all path:

```tsx
<Route path="*" element={ <NotFoundPage /> } />
```

## Notes

Distinct from #8873 (a specific PromptTemplateViewer navigation bug). Verified against current `main`; `path="*"` does not appear anywhere in `ui-app`.

Partha-Shankar:
Fixes #8931

## Root cause

In react-router v7, a `<Route>` with no `path` and no `index` is a pathless
layout route that only renders when one of its child routes matches. The
`NotFoundPage` route had no children, so it never matched anything.

Navigating to any unknown URL left the content area blank — the masthead
rendered but nothing else did.

## Fix

Add `path="*"` to the fallback route, making it a standard catch-all.
`NotFoundPage` renders inside the existing `<ApplicationAuth>` and
`<Page masthead={<AppHeader />}>` wrappers, so the full app shell stays intact.

## Testing

- ESLint: clean (0 errors, 0 warnings)
- No UI unit tests exist in this repo for this component
- Manually verified: navigating to `/does-not-exist` now renders the 404 page
  with the "Show all artifacts" button

### DIFF
--- a/ui/ui-app/src/app/MainPageWithAuth.tsx
+++ b/ui/ui-app/src/app/MainPageWithAuth.tsx
@@ -134,7 +134,7 @@ export const MainPageWithAuth: FunctionComponent<MainPageWithAuthProps> = () => 
                         />
 
 
-                        <Route element={ <NotFoundPage /> } />
+                        <Route path="*" element={ <NotFoundPage /> } />
                     </Routes>
                 </Page>
             </ApplicationAuth>


Nachiket-Roy:
Hey @Partha-Shankar ,  your commits is missing a DCO sign-off. Otherwise lgtm

Partha-Shankar:
@Nachiket-Roy Done — added DCO sign-off and force-pushed. Thanks for the review!

EricWittmann:
Thanks for the fix @Partha-Shankar! Clean one-liner with a great root cause explanation. Merged.

PR #8953, opened 2026-07-28
Fix OIDC token endpoint trailing slash handling

### LINKED ISSUE #8939: OIDC token endpoint trailing slash is always stripped, breaking Authentik (client credentials returns 405)
## Description

**Registry Version**: 3.3.1 (also present in 3.3.0)

**Persistence type**: sql

Since 3.3.0 the server always removes the trailing slash from the OIDC token endpoint URL before making the client credentials request. This breaks any OIDC provider that serves its token endpoint with a trailing slash, which is the case for Authentik. The exact same configuration worked fine on 3.2.6.

I use the HTTP basic authentication support (client credentials grant against the OIDC provider) so that a Kafka Connect / Debezium worker using the Confluent-compatible API can register schemas. On 3.2.6 this worked without issues. After upgrading to 3.3.1 every write coming through basic auth fails, because the server can no longer obtain an access token from Authentik.

The relevant change is in `OidcAuthenticationStrategy`. In 3.2.6 the token URL was used as configured:

```java
if (authConfig.oidcTokenPath.startsWith("http")) {
    this.oidcTokenUrl = authConfig.oidcTokenPath;
} else {
    this.oidcTokenUrl = authConfig.authServerUrl + authConfig.oidcTokenPath;
}
```

In 3.3.0 a loop was added that strips every trailing slash:

```java
String tokenUrl;
if (authConfig.oidcTokenPath.startsWith("http")) {
    tokenUrl = authConfig.oidcTokenPath;
} else {
    tokenUrl = authConfig.authServerUrl + authConfig.oidcTokenPath;
}
while (tokenUrl.endsWith("/")) {
    tokenUrl = tokenUrl.substring(0, tokenUrl.length() - 1);
}
this.oidcTokenUrl = tokenUrl;
```

The problem is that this strips the slash unconditionally, even when the user has explicitly configured an absolute URL through `quarkus.oidc.token-path`. There is no way to keep the trailing slash from the configuration, so providers that require it can no longer be used.

Authentik is built on Django with `APPEND_SLASH = False`, so it does not redirect a request without a trailing slash to the version with one. It just returns 405 Method Not Allowed for a POST to the endpoint without the slash. Its discovery document advertises every endpoint with a trailing slash. This is the token endpoint as published by our provider:

```json
{
  "issuer": "http://authentik-server.authentik.svc.cluster.local/application/o/apicurio-registry/",
  "authorization_endpoint": "http://authentik-server.authentik.svc.cluster.local/application/o/authorize/",
  "token_endpoint": "http://authentik-server.authentik.svc.cluster.local/application/o/token/",
  "userinfo_endpoint": "http://authentik-server.authentik.svc.cluster.local/application/o/userinfo/",
  "jwks_uri": "http://authentik-server.authentik.svc.cluster.local/application/o/apicurio-registry/jwks/",
  "grant_types_supported": [
    "authorization_code",
    "refresh_token",
    "implicit",
    "client_credentials",
    "password",
    "urn:ietf:params:oauth:grant-type:device_code"
  ],
  "token_endpoint_auth_methods_supported": [
    "client_secret_post",
    "client_secret_basic"
  ]
}
```

This is the same class of problem that other projects have hit with Authentik when they normalize the URL by trimming the trailing slash, for example openid-connect-generic https://github.com/oidc-wp/openid-connect-generic/issues/633 and rustfs https://github.com/rustfs/rustfs/issues/2349. The fix on their side was to stop trimming the slash that Authentik needs.

What I would like to request is a way to control this behaviour through an environment variable, so the trailing slash removal can be turned off for providers that require the slash. Something like an `apicurio.authn.token-endpoint.strip-trailing-slash` flag (default true to keep the current behaviour) would let Authentik users keep the slash while not changing anything for Keycloak and the rest. Simply respecting the URL exactly as configured when it is an absolute URL would also solve it.

### Environment

Kubernetes v1.36.3, Apicurio Registry running as a Deployment (single replica), SQL storage on PostgreSQL. OIDC provider is Authentik (self-hosted in the same cluster). Authentication uses OIDC plus the HTTP basic client credentials support. The client hitting the Confluent-compatible API is a Debezium PostgreSQL source connector running on Kafka Connect, using the Confluent AvroConverter.

### Steps to Reproduce

1. Configure the server for OIDC against Authentik, with the basic client credentials support enabled:
   - `QUARKUS_OIDC_TENANT_ENABLED=true`
   - `QUARKUS_OIDC_AUTH_SERVER_URL=http://authentik-server.authentik.svc.cluster.local/application/o/apicurio-registry/`
   - `QUARKUS_OIDC_TOKEN_PATH=http://authentik-server.authentik.svc.cluster.local/application/o/token/`
   - `APICURIO_AUTHN_BASIC_CLIENT_CREDENTIALS_ENABLED=true`
2. Send any write request using HTTP basic auth, for example register a schema through the Confluent-compatible API (`POST /apis/ccompat/v7/subjects/<subject>/versions`).
3. The server tries to obtain an access token and the request to the token endpoint fails with 405.

To confirm it is the trailing slash, a plain POST from inside the cluster against both variants of the endpoint:

```
# without slash (what 3.3.x sends after stripping)
POST http://authentik-server.authentik.svc.cluster.local/application/o/token   -> 405

# with slash (what 3.2.6 used to send)
POST http://authentik-server.authentik.svc.cluster.local/application/o/token/  -> 400
```

The 400 with the slash is the expected response from the token endpoint (missing credentials in this bare test); the point is that the method is accepted. Without the slash the method is rejected with 405.

Expected: the server obtains an access token from Authentik using client credentials, and the basic auth write request succeeds, the same way it did on 3.2.6.

Actual: the token request goes to the endpoint without the trailing slash, Authentik returns 405, the server cannot obtain a token, and the write request fails with 401. The downstream client (Kafka Connect / Debezium) then fails the task.

### Logs

Apicurio Registry (OIDC token request failing with 405):

```
WARN  [io.apicurio.registry.auth.OidcAuthenticationStrategy] Exception trying to get an access token with client credentials with client id: registry-api: io.apicurio.registry.auth.OidcAuthException: Failed to obtain access token
    at io.apicurio.registry.auth.AppAuthenticationMechanism.getAccessToken(AppAuthenticationMechanism.java:278)
    ...
Caused by: io.apicurio.registry.auth.OidcAuthException: OIDC token request failed with status 405
    at io.apicurio.registry.auth.AppAuthenticationMechanism.getAccessToken(AppAuthenticationMechanism.java:262)

WARN  [io.apicurio.registry.auth.AuthorizedInterceptor] Authentication credentials missing and required for protected endpoint.
INFO  [io.apicurio.registry.logging.audit.AuditLogService] apicurio.audit action="registerSchemaUnderSubject" result="failure" src_ip="10.0.1.7" error_msg="User is not authenticated." artifact_id="postgres_cdc.message-key"
INFO  [access log] method="POST" path="/apis/ccompat/v7/subjects/postgres_cdc.message-key/versions?normalize=false" response_code="401"
```

Kafka Connect / Debezium (downstream failure caused by the 401 above):

```
ERROR WorkerSourceTask{id=debezium-postgres-source-0} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted
org.apache.kafka.connect.errors.ConnectException: Tolerance exceeded in error handler
Caused by: org.apache.kafka.common.config.ConfigException: Failed to access Avro data from topic ... : Error registering Avro schema ...
    at io.confluent.connect.avro.AvroConverter.fromConnectData(AvroConverter.java:122)
```

The same deployment on 3.2.6 works with no changes other than the image version.

Rg9906:
## Summary

This PR addresses #8939 by preserving explicitly configured absolute OIDC token endpoint URLs exactly as provided, while retaining the existing trailing-slash normalization for URLs constructed from `authServerUrl + oidcTokenPath`.

The current implementation strips trailing slashes unconditionally, which causes providers such as Authentik to reject client credential requests when the configured token endpoint requires the trailing slash.

## Changes

- Preserve absolute `quarkus.oidc.token-path` values exactly as configured.
- Continue applying trailing-slash normalization only for constructed relative URLs.
- Added a regression test covering both absolute and relative URL handling.

## Why this approach?

The existing normalization remains unchanged for the common relative-path configuration, preserving current behavior.

For users who explicitly configure an absolute token endpoint URL, the configuration is now respected exactly as provided, which restores compatibility with providers such as Authentik that require the trailing slash.

This follows one of the approaches suggested in the original issue while keeping the change minimal and localized.

Fixes #8939.

---

This is my first contribution to Apicurio Registry, so if there's a project-specific convention or a different approach you'd prefer, I'd be happy to update the implementation. Thanks for taking the time to review it!

### DIFF
--- a/app/src/test/java/io/apicurio/registry/auth/OidcAuthenticationStrategyTest.java
+++ b/app/src/test/java/io/apicurio/registry/auth/OidcAuthenticationStrategyTest.java
@@ -1,0 +1,45 @@
+package io.apicurio.registry.auth;
+
+import org.junit.jupiter.api.Test;
+import java.lang.reflect.Field;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class OidcAuthenticationStrategyTest {
+
+    @Test
+    void testOidcTokenUrlConstruction() throws Exception {
+        // Case 1: Relative Path
+        AuthConfig config1 = new AuthConfig();
+        config1.authServerUrl = "http://keycloak/auth";
+        config1.oidcTokenPath = "/token";
+        OidcAuthenticationStrategy strategy1 = new OidcAuthenticationStrategy(null, config1, null, null, null, null);
+        assertEquals("http://keycloak/auth/token", getOidcTokenUrl(strategy1));
+
+        // Case 2: Relative Path with Trailing Slash
+        AuthConfig config2 = new AuthConfig();
+        config2.authServerUrl = "http://keycloak/auth";
+        config2.oidcTokenPath = "/token/";
+        OidcAuthenticationStrategy strategy2 = new OidcAuthenticationStrategy(null, config2, null, null, null, null);
+        assertEquals("http://keycloak/auth/token", getOidcTokenUrl(strategy2));
+
+        // Case 3: Absolute URL
+        AuthConfig config3 = new AuthConfig();
+        config3.authServerUrl = "http://keycloak/auth"; // Should be ignored
+        config3.oidcTokenPath = "http://authentik/token";
+        OidcAuthenticationStrategy strategy3 = new OidcAuthenticationStrategy(null, config3, null, null, null, null);
+        assertEquals("http://authentik/token", getOidcTokenUrl(strategy3));
+
+        // Case 4: Absolute URL with Trailing Slash
+        AuthConfig config4 = new AuthConfig();
+        config4.authServerUrl = "http://keycloak/auth"; // Should be ignored
+        config4.oidcTokenPath = "http://authentik/token/";
+        OidcAuthenticationStrategy strategy4 = new OidcAuthenticationStrategy(null, config4, null, null, null, null);
+        assertEquals("http://authentik/token/", getOidcTokenUrl(strategy4));
+    }
+
+    private String getOidcTokenUrl(OidcAuthenticationStrategy strategy) throws Exception {
+        Field field = OidcAuthenticationStrategy.class.getDeclaredField("oidcTokenUrl");
+        field.setAccessible(true);
+        return (String) field.get(strategy);
+    }
+}
--- a/app/src/main/java/io/apicurio/registry/auth/OidcAuthenticationStrategy.java
+++ b/app/src/main/java/io/apicurio/registry/auth/OidcAuthenticationStrategy.java
@@ -92,9 +92,9 @@ public class OidcAuthenticationStrategy implements AuthenticationStrategy {
             tokenUrl = authConfig.oidcTokenPath;
         } else {
             tokenUrl = authConfig.authServerUrl + authConfig.oidcTokenPath;
-        }
-        while (tokenUrl.endsWith("/")) {
-            tokenUrl = tokenUrl.substring(0, tokenUrl.length() - 1);
+            while (tokenUrl.endsWith("/")) {
+                tokenUrl = tokenUrl.substring(0, tokenUrl.length() - 1);
+            }
         }
         this.oidcTokenUrl = tokenUrl;
     }


app/src/test/java/io/apicurio/registry/auth/OidcAuthenticationStrategyTest.java
```diff
@@ -0,0 +1,45 @@
...
+        // Case 4: Absolute URL with Trailing Slash
+        AuthConfig config4 = new AuthConfig();
+        config4.authServerUrl = "http://keycloak/auth"; // Should be ignored
+        config4.oidcTokenPath = "http://authentik/token/";
+        OidcAuthenticationStrategy strategy4 = new OidcAuthenticationStrategy(null, config4, null, null, null, null);
+        assertEquals("http://authentik/token/", getOidcTokenUrl(strategy4));
+    }
+
+    private String getOidcTokenUrl(OidcAuthenticationStrategy strategy) throws Exception {
+        Field field = OidcAuthenticationStrategy.class.getDeclaredField("oidcTokenUrl");
```
paoloantinori, line 41: Reading the field via reflection works, but it creates a runtime-only coupling to the field name `oidcTokenUrl` — if the field is ever renamed, this fails with `NoSuchFieldException` at test time rather than a compile error. If a package-private accessor exists or can be added without breaking encapsulation, that would be more maintainable. Not blocking.

paoloantinori (commented):
Clean, well-scoped fix for #8939 — moving the trailing-slash normalization into the `else` branch so explicitly-configured absolute token endpoint URLs are preserved exactly, while the constructed relative-path case keeps its existing behavior. The regression test correctly covers all four cases (relative/absolute × with/without trailing slash), with case 4 (`http://authentik/token/` preserved) being the key assertion that would have failed before the fix.

A couple of minor notes:

- The test reads `oidcTokenUrl` via reflection. This works but couples the test to the private field name — a rename would break the test at runtime rather than at compile time. Not blocking, but worth considering a package-private getter if one is easy to add.
- The branch is behind `main` (`mergeable_state: behind`) — a rebase will be needed before merge.

Overall this looks good. Welcome to the project — nice first contribution!

EricWittmann (commented):
Thanks for the contribution, @Rg9906 — the production fix is correct and nicely scoped. Moving the trailing-slash normalization into the `else` branch preserves absolute URLs exactly as configured (fixing the Authentik regression) while keeping the existing behavior for relative paths. This matches the workaround the issue reporter was already using successfully.

A few items to address before this can be merged:

### Blockers

1. **DCO sign-off is missing on both commits.** All commits to this project require a `Signed-off-by:` line. You can fix this by rebasing and amending with `git commit --amend -s`, or squashing into a single signed commit. See the [DCO documentation](https://github.com/apps/dco) for details.

2. **Commit messages should use Conventional Commits format.** For example: `fix(auth): preserve trailing slash on absolute OIDC token endpoint URLs (#8939)`.

### Suggestions

3. **Reflection in the test** — as @paoloantinori noted, reading `oidcTokenUrl` via `Field.setAccessible(true)` couples the test to the private field name. A rename would break the test at runtime (`NoSuchFieldException`) rather than at compile time. Consider adding a package-private getter to `OidcAuthenticationStrategy`:

   ```java
   String getOidcTokenUrl() {
       return oidcTokenUrl;
   }
   ```

   Then the test can call `strategy.getOidcTokenUrl()` directly without reflection. Not a hard blocker, but it would make the test more maintainable.

4. **Rebase onto current `main`** — the branch is behind and includes a merge commit. A rebase will clean up the history (the squash on merge would handle this too, but a clean rebase is preferred).

### What looks good

- The 4 test cases cover the full matrix (relative/absolute × with/without trailing slash) — case 4 (`http://authentik/token/` preserved) is the key regression assertion.
- Passing `null` for unused constructor parameters is fine here since they're just stored in fields and not dereferenced before the tokenUrl logic.
- The fix is minimal and localized — no unnecessary changes.

Please push an update when you get a chance and we'll get this merged. Welcome to the project!

EricWittmann:
Thanks for the contribution, @Rg9906 — nice first PR! The trailing-slash fix is clean and well-scoped. Merged via squash.

PR #8962, opened 2026-07-28
fix(kafkasql): publish contract metadata and status events

### LINKED ISSUE #8928: bug: KafkaSQL does not emit CONTRACT_METADATA_UPDATED or CONTRACT_STATUS_CHANGED events
**Registry Version**: 3.3.0  
**Persistence type**: KafkaSQL

## Description

On KafkaSQL, updating a contract's metadata or changing its status does not produce the `CONTRACT_METADATA_UPDATED` or `CONTRACT_STATUS_CHANGED` events on the `registry-events` topic.

Both operations succeed and persist, and the event pipeline is otherwise working (`ARTIFACT_CREATED` and `ARTIFACT_VERSION_CREATED` are emitted as expected). However, these two contract events are never produced.

The SQL storage implementation has test coverage for these events in #8860.

## Steps to Reproduce

Run the registry with KafkaSQL storage (`apicurio-registry:3.3.0`), consume the `registry-events` topic from the beginning, then using a fresh artifact ID:

1. `POST /apis/registry/v3/groups/default/artifacts` — create an artifact.
2. `PUT /apis/registry/v3/groups/default/artifacts/{id}/contract/metadata`
   with:
   ```json
   {
     "status": "STABLE",
     "ownerTeam": "repro-team"
   }
   ```
3. `POST /apis/registry/v3/groups/default/artifacts/{id}/contract/status`
   with:
   ```json
   {
     "status": "DEPRECATED"
   }
   ```
4. Inspect the events emitted to the `registry-events` topic.

## Actual

Both contract operations return `200` and persist successfully:

```text
PUT  .../contract/metadata  -> 200
GET  .../contract/metadata  -> {"status":"STABLE","ownerTeam":"repro-team"}

POST .../contract/status    -> 200
response                    -> {"status":"DEPRECATED","ownerTeam":"repro-team","deprecatedDate":"2026-07-27"}
```

The event pipeline is working, and `registry-events` contains:

```text
ARTIFACT_CREATED
ARTIFACT_VERSION_CREATED
```

However, no `CONTRACT_METADATA_UPDATED` or `CONTRACT_STATUS_CHANGED` event is produced.

## Expected

`CONTRACT_METADATA_UPDATED` and `CONTRACT_STATUS_CHANGED` should be emitted on KafkaSQL, consistent
with other events on KafkaSQL and with the SQL storage implementation.

## Suspected cause (code analysis)

From the current implementation, these two events appear to be the only ones fired from the REST layer (`GroupsResourceImpl`) using a hardcoded `SqlOutboxEvent`, rather than being emitted from the storage layer using the storage-variant-specific outbox event type.

On KafkaSQL, this appears to result in the `SqlOutboxEvent` being handled by `SqlEventsProcessor`, which routes it to the embedded SQL store where `supportsDatabaseEvents()` returns `false`, while `KafkaSqlEventsProcessor` observes only `KafkaSqlOutboxEvent`. As a result, the event never reaches the `registry-events` topic.

I traced this flow in the code, but the observer path itself has not been verified directly at runtime.

## Related

- #8794 — fixed a similar KafkaSQL event emission issue for `ARTIFACT_VERSION_STATE_CHANGED` (different event and root cause).
- #8860 — adds SQL-side test coverage for contract events; it does not cover KafkaSQL.

Anuragkumar-687:
Fixes #8928

## Problem

On the KafkaSQL storage variant, updating a contract's metadata or transitioning its contract status does not emit the corresponding CONTRACT_METADATA_UPDATED and CONTRACT_STATUS_CHANGED events to the registry-events topic.

Both operations complete successfully (200 OK) and persist their changes. Other registry events (for example, artifact and version creation) are published as expected, but these two contract events are missing.

The events were fired directly from the REST layer (GroupsResourceImpl) as a SqlOutboxEvent. That event type is only processed by SqlEventsProcessor. On KafkaSQL, it is routed to the embedded SQL store (where database events are not enabled) and is therefore never published. Meanwhile, KafkaSqlEventsProcessor only handles KafkaSqlOutboxEvent, so the events never reach the registry-events topic.

Unlike other registry events, these two contract events bypassed the storage abstraction and were tied directly to the SQL implementation.

## Fix

- Route contract metadata and contract status events through RegistryStorage.createEvent(...) instead of publishing a hardcoded SqlOutboxEvent.

- Implement createEvent(...) in KafkaSqlRegistryStorage so KafkaSQL publishes these events instead of ignoring them.

- Keep the SQL implementation unchanged; it continues to persist outbox events through the existing event repository, guarded by supportsDatabaseEvents().

This removes the storage-specific event publishing logic from the REST layer and allows each storage implementation to publish events using its own mechanism.

## Testing

Added `KafkaSqlContractEventsTest`, which drives the public REST endpoints (create artifact, update contract metadata, transition contract status) and asserts that both `CONTRACT_METADATA_UPDATED` and `CONTRACT_STATUS_CHANGED` are published to the `registry-events` topic. The test fails before this change (neither event is produced) and passes after it.

Verified locally:

    ./mvnw test -pl app -am \
      -Dtest=KafkaSqlContractEventsTest \
      -Dsurefire.failIfNoSpecifiedTests=false

## Note

The SQL variant continues to write the outbox row via `createEvent` as before. On KafkaSQL, the event is now published to `registry-events` in the same way as the other registry events.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
@@ -182,9 +182,6 @@ public class GroupsResourceImpl extends AbstractResourceImpl implements GroupsRe
     io.apicurio.registry.contracts.migration.MigrationRuleService migrationRuleService;
 
     @Inject
-    jakarta.enterprise.event.Event<io.apicurio.registry.storage.impl.sql.SqlOutboxEvent> contractOutboxEvent;
-
-    @Inject
     io.apicurio.registry.contracts.audit.ContractAuditService contractAuditService;
 
     /**
@@ -2013,8 +2010,7 @@ public class GroupsResourceImpl extends AbstractResourceImpl implements GroupsRe
         storage.mergeArtifactLabels(rawGroupId, artifactId, prefix, contractLabels);
 
         // Fire metadata updated event
-        contractOutboxEvent.fire(io.apicurio.registry.storage.impl.sql.SqlOutboxEvent.of(
-                io.apicurio.registry.events.ContractMetadataUpdated.of(rawGroupId, artifactId)));
+        storage.createEvent(io.apicurio.registry.events.ContractMetadataUpdated.of(rawGroupId, artifactId));
 
         // Audit log
         contractAuditService.recordAction(rawGroupId, artifactId, null,
@@ -2161,11 +2157,9 @@ public class GroupsResourceImpl extends AbstractResourceImpl implements GroupsRe
         }
 
         // Fire status changed event
-        contractOutboxEvent.fire(io.apicurio.registry.storage.impl.sql.SqlOutboxEvent.of(
-                io.apicurio.registry.events.ContractStatusChanged.of(rawGroupId, artifactId,
-                        currentMetadata.getStatus() != null
-                                ? currentMetadata.getStatus().name() : null,
-                        targetStatus.name())));
+        storage.createEvent(io.apicurio.registry.events.ContractStatusChanged.of(rawGroupId, artifactId,
+                currentMetadata.getStatus() != null ? currentMetadata.getStatus().name() : null,
+                targetStatus.name()));
 
         // Audit log
         contractAuditService.recordAction(rawGroupId, artifactId, null,
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
@@ -1294,7 +1294,7 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
 
     @Override
     public String createEvent(OutboxEvent event) {
-        // No op, the event is created by the event processor.
+        outboxEvent.fire(KafkaSqlOutboxEvent.of(event));
         return event.getId();
     }
 
--- a/app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlContractEventsTest.java
+++ b/app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlContractEventsTest.java
@@ -1,0 +1,92 @@
+package io.apicurio.registry.storage.impl.kafkasql;
+
+import io.apicurio.registry.AbstractResourceTestBase;
+import io.apicurio.registry.types.ArtifactType;
+import io.apicurio.registry.types.ContentTypes;
+import io.apicurio.registry.utils.tests.KafkasqlTestProfile;
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.kafka.common.serialization.StringDeserializer;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import static io.restassured.RestAssured.given;
+
+@QuarkusTest
+@TestProfile(KafkasqlTestProfile.class)
+public class KafkaSqlContractEventsTest extends AbstractResourceTestBase {
+
+    private static final String EVENTS_TOPIC = "registry-events";
+    private static final String GROUP = "default";
+
+    @Test
+    public void testContractEventsArePublished() throws Exception {
+        String artifactId = "contract-events-" + UUID.randomUUID();
+
+        KafkaConsumer<String, String> consumer = new KafkaConsumer<>(
+                Map.of(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG,
+                        System.getProperty("bootstrap.servers.external"),
+                        ConsumerConfig.GROUP_ID_CONFIG, "tc-" + UUID.randomUUID(),
+                        ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest"),
+                new StringDeserializer(), new StringDeserializer());
+        try {
+            consumer.subscribe(List.of(EVENTS_TOPIC));
+
+            createArtifact(GROUP, artifactId, ArtifactType.JSON, "{\"type\":\"object\"}",
+                    ContentTypes.APPLICATION_JSON);
+
+            given().when().contentType("application/json").pathParam("groupId", GROUP)
+                    .pathParam("artifactId", artifactId)
+                    .body("{\"status\": \"STABLE\", \"ownerTeam\": \"events-test\"}")
+                    .put("/registry/v3/groups/{groupId}/artifacts/{artifactId}/contract/metadata").then()
+                    .statusCode(200);
+
+            given().when().contentType("application/json").pathParam("groupId", GROUP)
+                    .pathParam("artifactId", artifactId).body("{\"status\": \"DEPRECATED\"}")
+                    .post("/registry/v3/groups/{groupId}/artifacts/{artifactId}/contract/status").then()
+                    .statusCode(200);
+
+            List<String> events = consumeEvents(consumer, artifactId);
+
+            Assertions.assertTrue(
+                    events.stream().anyMatch(e -> e.contains("CONTRACT_METADATA_UPDATED")),
+                    "Expected a CONTRACT_METADATA_UPDATED event for artifact " + artifactId
+                            + " on the " + EVENTS_TOPIC + " topic. Events found: " + events);
+            Assertions.assertTrue(events.stream().anyMatch(e -> e.contains("CONTRACT_STATUS_CHANGED")),
+                    "Expected a CONTRACT_STATUS_CHANGED event for artifact " + artifactId + " on the "
+                            + EVENTS_TOPIC + " topic. Events found: " + events);
+        } finally {
+            consumer.close();
+        }
+    }
+
+    /**
+     * Collects the events published for the given artifact, waiting until both contract events have
+     * been seen or the timeout expires.
+     */
+    private List<String> consumeEvents(KafkaConsumer<String, String> consumer, String artifactId) {
+        List<String> events = new ArrayList<>();
+        long deadline = System.currentTimeMillis() + 20000;
+        while (System.currentTimeMillis() < deadline) {
+            for (ConsumerRecord<String, String> consumerRecord : consumer.poll(Duration.ofMillis(500))) {
+                if (consumerRecord.value() != null && consumerRecord.value().contains(artifactId)) {
+                    events.add(consumerRecord.value());
+                }
+            }
+            if (events.stream().anyMatch(e -> e.contains("CONTRACT_METADATA_UPDATED"))
+                    && events.stream().anyMatch(e -> e.contains("CONTRACT_STATUS_CHANGED"))) {
+                break;
+            }
+        }
+        return events;
+    }
+}


app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1294,7 +1294,7 @@ public String createSnapshot(String snapshotLocation) throws RegistryStorageExce
 
     @Override
     public String createEvent(OutboxEvent event) {
-        // No op, the event is created by the event processor.
+        outboxEvent.fire(KafkaSqlOutboxEvent.of(event));
```
paoloantinori, line 1297: This no-op previously applied to every `createEvent(...)` caller, not just the two contract-event paths changed in `GroupsResourceImpl`. Since `createEvent` is on the `RegistryStorage` interface, worth confirming (grep for `.createEvent(`) that no other call site relied on the KafkaSQL no-op — otherwise those paths now silently start firing `KafkaSqlOutboxEvent` on the KafkaSQL variant.

paoloantinori (commented):
Clean fix — routing the two contract events through `RegistryStorage.createEvent(...)` instead of a hardcoded `SqlOutboxEvent` is the right design, and the KafkaSQL integration test is solid. One thing worth verifying before merge.

- **Blast radius of the no-op → fire change.** `KafkaSqlRegistryStorage.createEvent(...)` was previously a documented no-op ("the event is created by the event processor"). This PR makes it fire `KafkaSqlOutboxEvent` for **every** caller. The PR targets the two contract-event sites in `GroupsResourceImpl`, but `createEvent` sits on the storage interface — worth a quick grep to confirm there are no other callers that silently relied on the KafkaSQL no-op. If there are, those paths would now also start emitting events on the KafkaSQL variant (potentially duplicates if those events are also generated during Kafka state replay by the consumer).

- **Test quality.** The UUID-based `artifactId`, `earliest` offset reset with content filtering, and 20 s polling deadline make the assertion robust against interleaved traffic on the shared `registry-events` topic. Good coverage of both `CONTRACT_METADATA_UPDATED` and `CONTRACT_STATUS_CHANGED`.

- **Mergeable state is `behind`** — will need a rebase before merge, but no code-quality blocker.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
Anuragkumar-687: Thanks for the review! I grepped all RegistryStorage.createEvent(...) call sites. The only production callers are the two contract-event paths in GroupsResourceImpl; the remaining references are interface/decorator/implementation/repository plumbing. I didn't find any other callers that would start emitting duplicate events on KafkaSQL.

Anuragkumar-687:
Hi @EricWittmann,
I investigated the CI failures. I also verified locally that:

completes successfully, and all of the regular unit tests, integration tests, build, and verification checks are now passing.

The remaining failures appear to be limited to the Scalpel unit test workflow. Since the corresponding standard unit test jobs all passed, I'm not seeing any failing tests related to this change.

Could you please advise whether there's anything further you'd like me to investigate, or if the remaining Scalpel failures should be handled separately?

Thanks!

EricWittmann (commented):
Thanks for the work on this @Anuragkumar-687 — the bug analysis is spot-on and the fix does solve the immediate problem. However, after a deeper review of the event sourcing architecture, I think the right path forward is to fix the underlying design issue rather than patch around it. I suggest we close this PR and we'll address the root cause in a followup PR, but I wanted to document the analysis here for reference.

## Event Flow Analysis

### SQL Storage — Event Flow

```
AbstractSqlRegistryStorage.createArtifact()
  │
  ├── INSERT artifact/version into database
  │
  └── outboxEvent.fire(SqlOutboxEvent.of(ArtifactCreated.of(...)))
        │
        └── SqlEventsProcessor.processEvent(@Observes SqlOutboxEvent)
              │  guard: supportsDatabaseEvents() && isReady()
              │
              └── sqlStore.createEvent(outboxEvent)
                    │
                    └── SqlEventRepository.createEvent()
                          │
                          ├── INSERT INTO outbox (id, topic, aggregateId, type, payload)
                          └── DELETE FROM outbox WHERE id = ?
                                │
                                └── [External: Debezium CDC captures INSERT from WAL]
```

The event reaches the outside world via the database WAL. The INSERT+DELETE in the same transaction is the classic transactional outbox pattern — Debezium captures the INSERT before the DELETE removes the row.

### KafkaSQL Storage — Event Flow

```
KafkaSqlRegistryStorage.createArtifact()
  │
  ├── (1) Submit CreateArtifact11Message to kafkasql-journal topic
  │
  ├── (2) Block on coordinator.waitForResponse(uuid)
  │         │
  │         │  [Consumer thread]
  │         │  KafkaSqlSink.processMessage()
  │         │    └── CreateArtifact11Message.dispatchTo(sqlStore)
  │         │          └── AbstractSqlRegistryStorage.createArtifact()
  │         │                ├── INSERT into embedded H2 database
  │         │                └── fires SqlOutboxEvent → SqlEventsProcessor
  │         │                      └── supportsDatabaseEvents() = FALSE (H2)
  │         │                          → NO-OP
  │         │
  │         └── coordinator.notifyResponse(uuid, result) → unblocks (2)
  │
  ├── (3) outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactCreated.of(...)))
  │         └── KafkaSqlEventsProcessor → "registry-events" Kafka topic
  │
  └── (4) outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactVersionCreated.of(...)))
            └── KafkaSqlEventsProcessor → "registry-events" Kafka topic
```

Key design point: `KafkaSqlOutboxEvent` is fired **by the originating node only**, after the journal round-trip completes. Other nodes replaying the same journal message call `dispatchTo(sqlStore)` which fires `SqlOutboxEvent` — a no-op on H2. So no duplicate events.

## The Architectural Problem

**Every other event** in the system follows this pattern: the storage method does the work → the storage method fires the appropriate CDI event (`SqlOutboxEvent` or `KafkaSqlOutboxEvent`). The REST layer never thinks about events.

**Contract events break this pattern.** The storage layer has no concept of "contracts" — it only knows about labels. So the REST layer (`GroupsResourceImpl`) is forced to fire events itself. This PR fixes the immediate symptom (wrong event type on KafkaSQL) by routing through `storage.createEvent()`, but `createEvent()` is essentially a backdoor that lets the REST layer inject events into the storage's event pipeline. It works, but it's the wrong abstraction.

This also means `ContractRulesetConfigured` events (6 call sites in `AbstractSqlRegistryStorage`) have the same gap — they fire as `SqlOutboxEvent` from the SQL storage layer, but the KafkaSQL methods don't fire `KafkaSqlOutboxEvent` for those operations.

## The Proper Fix

Add contract-specific methods to the `RegistryStorage` interface:

```java
void updateContractMetadata(String groupId, String artifactId, ContractMetadata metadata);
void transitionContractStatus(String groupId, String artifactId, String fromStatus, String toStatus);
```

The SQL implementation would do the label merge and fire `SqlOutboxEvent.of(ContractMetadataUpdated.of(...))`. The KafkaSQL implementation would submit a journal message, wait for the response, then fire `KafkaSqlOutboxEvent.of(ContractMetadataUpdated.of(...))` — exactly like `createArtifact()` and every other operation. The REST layer would just call `storage.updateContractMetadata()` with no event awareness at all.

## Decision

I think we should close this PR and address the underlying architectural issue as a maintainer-driven change, since it touches the storage interface and KafkaSQL journal message handling. @Anuragkumar-687 — your bug analysis was excellent and directly informed this decision. Thank you for the contribution.

@carlesarnal — tagging you for feedback on this analysis. Does this align with how you'd want to see the contract event architecture evolve?

EricWittmann:
Adding to the analysis above — there are six additional methods already on the `RegistryStorage` interface that have the same gap. These methods are properly on the storage interface (unlike the contract metadata/status events), but the KafkaSQL implementations don't fire `KafkaSqlOutboxEvent` after the journal round-trip:

| Method | SQL fires event? | KafkaSQL fires event? |
|--------|:-:|:-:|
| `setArtifactContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `deleteArtifactContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `setVersionContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `deleteVersionContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `setGlobalContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `deleteGlobalContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |

In each case, the SQL implementation in `AbstractSqlRegistryStorage` fires `SqlOutboxEvent.of(ContractRulesetConfigured.of(...))` after the repository call. The KafkaSQL implementation in `KafkaSqlRegistryStorage` submits the journal message and waits for the response, but never fires `KafkaSqlOutboxEvent` — so the `ContractRulesetConfigured` event is silently lost on KafkaSQL.

These should be straightforward to fix as part of the broader contract event architecture work — they already follow the right pattern (storage interface method → journal message → response → fire event), they're just missing the final `outboxEvent.fire(KafkaSqlOutboxEvent.of(...))` step.

Anuragkumar-687:
Thanks for the detailed analysis, @EricWittmann. The architectural direction makes sense to me.

I'm happy to work on the storage-interface approach instead if you'd prefer. Would you like me to update this PR to implement that design, or would you prefer to close this one and have me open a new PR once the architecture is agreed upon?

carlesarnal (approved):

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -2013,8 +2010,7 @@ public ContractMetadata updateContractMetadata(String groupId, String artifactId
         storage.mergeArtifactLabels(rawGroupId, artifactId, prefix, contractLabels);
 
         // Fire metadata updated event
-        contractOutboxEvent.fire(io.apicurio.registry.storage.impl.sql.SqlOutboxEvent.of(
-                io.apicurio.registry.events.ContractMetadataUpdated.of(rawGroupId, artifactId)));
```
carlesarnal, line 2017: Nit: the fully qualified class names (`io.apicurio.registry.events.ContractMetadataUpdated`, `io.apicurio.registry.events.ContractStatusChanged`) should be regular imports.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlContractEventsTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import static io.restassured.RestAssured.given;
+
+@QuarkusTest
+@TestProfile(KafkasqlTestProfile.class)
```
carlesarnal, line 25: Nit: consider adding `@Tag(ApicurioTestTags.SLOW)` — this test polls Kafka for up to 20s, and the analogous `KafkaSqlEventsTest` carries that tag.

carlesarnal (approved):
Clean fix — routing contract events through `RegistryStorage.createEvent()` is the right abstraction. Left two minor nits.

carlesarnal:
@EricWittmann I think you're right, contract events should go through proper storage-interface methods like every other operation. That said, I think we should merge this PR and address the architecture in a follow-up:

- The fix is correct, well-tested, and doesn't change any interface, it just makes the existing `KafkaSqlRegistryStorage.createEvent()` do something instead of being a no-op.
- It fixes a real bug (contract events silently lost on KafkaSQL) that users hit today.
- The proper refactor (adding contract methods to `RegistryStorage`) will replace this code entirely, the end state is the same regardless of whether we merge this first.

I'd rather ship the bug fix now and do the architectural cleanup (including the 6 `ContractRulesetConfigured` gaps) as a follow-up.

Anuragkumar-687:
Thanks for the detailed analysis, @EricWittmann, and thanks @carlesarnal for the review and perspective.

I'm happy to follow whichever direction you think is best. One practical question: since contributors are generally limited to one active PR at a time, would you prefer merging this as the bug fix (as @carlesarnal suggested) and handling the architectural refactor in a follow-up, or would it be better to close this PR so I can work on the storage-interface approach instead?

Either way works for me—I just wanted to make sure I'm following the preferred workflow.

EricWittmann:
Thanks @Anuragkumar-687 for the excellent contribution! Your bug analysis was spot-on and directly informed our understanding of the broader architectural gap. Merged!

We're going to follow this up with an architectural cleanup to move contract operations into proper `RegistryStorage` interface methods (like every other storage operation) and fix the remaining KafkaSQL event gaps. We'll open a tracking issue for that work shortly.

Anuragkumar-687:
Thank you, @EricWittmann! I really appreciate your guidance throughout this discussion and the time you took to explain the underlying architecture in such detail. It was a great learning experience for me, and I came away with a much deeper understanding of the event flow and the design principles behind the storage layer.

I'm glad the investigation was useful and could help uncover the broader architectural gap. Thanks as well to @carlesarnal for the thoughtful reviews and feedback throughout the process.

Looking forward to contributing to the architectural cleanup and continuing to help improve KafkaSQL. Thanks again for the support and mentorship!

PR #8969, opened 2026-07-28
fix(ci): fetch base branch for Scalpel and fix fetch-depth expression

paoloantinori:
## Summary

Follow-up to #8903 (merged). Three fixes for Scalpel affected-module detection:

- **fetch-depth expression bug**: `&& 0 || 1` always evaluates to `1` because `0` is falsy in GitHub Actions expressions. Scalpel has never had full git history. Fix: quote as `'0'` (non-empty string, truthy).
- **JGit auth failure**: `scalpel.fetchBaseBranch=true` fails because JGit inside Maven lacks the `GITHUB_TOKEN` credentials. CI logs: `"Scalpel: Failed to fetch origin/main"`. Replace with a workflow-level `git fetch` that inherits the checkout token.
- **Graceful degradation**: `continue-on-error: true` ensures transient fetch failures degrade to a full build (matching Scalpel's `failSafe=true` behavior). `env.GITHUB_BASE_REF != ''` guard prevents failure on non-PR callers.

**Evidence** (run 30387339544, Scalpel shard logs):
```
[INFO] Scalpel 0.3.7 activated (mode=skip-tests)
[INFO] Scalpel: Fetching main from origin
[WARNING] Scalpel: Failed to fetch origin/main, building all modules: Failed to fetch origin/main
```
And `fetch-depth: 1` despite `scalpel-enabled: true` (the expression bug).

## Test plan

- [ ] Verify Scalpel logs show `"Merge base between origin/main and HEAD: <sha>"` instead of the fallback warning
- [ ] Verify `fetch-depth: 0` in the checkout step (not 1)
- [ ] Compare test counts between normal and Scalpel shards on a PR touching a single module
- [ ] Confirm the fetch step is skipped for non-Scalpel shards (check logs for "Fetch base branch" step)
- [ ] Confirm a transient fetch failure doesn't fail the job (continue-on-error)

### DIFF
--- a/.github/workflows/verify-unit-tests.yaml
+++ b/.github/workflows/verify-unit-tests.yaml
@@ -59,7 +59,12 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           # Full history only needed when Scalpel is enabled (git diff against base branch)
-          fetch-depth: ${{ inputs.scalpel-enabled == 'true' && 0 || 1 }}
+          fetch-depth: ${{ inputs.scalpel-enabled == 'true' && '0' || 1 }}
+
+      - name: Fetch base branch for Scalpel
+        if: inputs.scalpel-enabled == 'true' && env.GITHUB_BASE_REF != ''
+        continue-on-error: true
+        run: git fetch origin "$GITHUB_BASE_REF" --no-tags
 
       - name: Set up JDK ${{ inputs.java-version }}
         uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5
--- a/.github/workflows/verify-unit-tests.yaml
+++ b/.github/workflows/verify-unit-tests.yaml
@@ -62,7 +62,7 @@ jobs:
           fetch-depth: ${{ inputs.scalpel-enabled == 'true' && '0' || 1 }}
 
       - name: Fetch base branch for Scalpel
-        if: inputs.scalpel-enabled == 'true' && env.GITHUB_BASE_REF != ''
+        if: inputs.scalpel-enabled == 'true' && github.base_ref != ''
         continue-on-error: true
         run: git fetch origin "$GITHUB_BASE_REF" --no-tags
 
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -5,6 +5,6 @@
     <extension>
         <groupId>eu.maveniverse.maven.scalpel</groupId>
         <artifactId>extension</artifactId>
-        <version>0.3.7</version>
+        <version>0.3.9</version>
     </extension>
 </extensions>
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -11,8 +11,5 @@
 # skipTestsForUpstream: upstream deps of -pl target compile but skip tests.
 # Works because non-app shard always tests upstream modules independently.
 -Dscalpel.skipTestsForUpstream=true
-# fetchBaseBranch: let Scalpel fetch origin/<base> itself via JGit (no shallow grafts).
-# Requires fetch-depth: 0 in the checkout step so full history is available.
--Dscalpel.fetchBaseBranch=true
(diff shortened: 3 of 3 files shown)


paoloantinori:
@carlesarnal this one is more urgent than I initially framed it. Since #8903 merged with `fetchBaseBranch=true`, Scalpel crashes ALL 7 shards with a `JGitInternalException: Missing unknown` on every Java-touching PR. The crash bypasses `failSafe=true` because it's a Maven `InternalErrorException`, not a Scalpel-caught error.

The Verification Gate still passes (Scalpel shards are non-gating), so PRs aren't blocked, but we're burning CI compute on 7 shards that crash immediately and generating red noise on every PR.

Verified on runs [30422483753](https://github.com/Apicurio/apicurio-registry/actions/runs/30422483753) and [30404525927](https://github.com/Apicurio/apicurio-registry/actions/runs/30404525927) (renovate PR).

This PR fixes the root cause: the `fetch-depth` expression and the missing base branch ref.

paoloantinori:
**Upstream fix:** Filed [maveniverse/scalpel#45](https://github.com/maveniverse/scalpel/issues/45) and opened [maveniverse/scalpel#46](https://github.com/maveniverse/scalpel/pull/46) to fix the `failSafe=true` gap that lets `JGitInternalException` crash Maven instead of degrading gracefully. Once that ships in a Scalpel release, the `continue-on-error: true` on the fetch step in this PR becomes a belt-and-suspenders safety net rather than the primary protection.

.github/workflows/verify-unit-tests.yaml
```diff
@@ -59,7 +59,12 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           # Full history only needed when Scalpel is enabled (git diff against base branch)
-          fetch-depth: ${{ inputs.scalpel-enabled == 'true' && 0 || 1 }}
+          fetch-depth: ${{ inputs.scalpel-enabled == 'true' && '0' || 1 }}
+
+      - name: Fetch base branch for Scalpel
```
EricWittmann, line 64: **Bug: `env.GITHUB_BASE_REF` is always empty in `if` expressions — this step will never run.**

Per [GitHub's documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables), default environment variables like `GITHUB_BASE_REF` are **not accessible through the `env` context** — only variables explicitly set via the `env:` keyword are. So `env.GITHUB_BASE_REF` always resolves to `''` here, making the condition always false.

The `$GITHUB_BASE_REF` in the `run:` line is fine (shell variable on the runner). The bug is only in the `if` guard.

This repo's own workflows confirm the correct pattern — e.g. [`operator.yaml`](https://github.com/Apicurio/apicurio-registry/blob/cefe20a7a0396ef2873f7fde29ce514cd519c172/.github/workflows/operator.yaml) uses `github.base_ref`, and no workflow uses `env.GITHUB_*` in `if` conditions.

```suggestion
        if: inputs.scalpel-enabled == 'true' && github.base_ref != ''
```

EricWittmann (commented):
## Code Review

Checked for bugs and CLAUDE.md compliance. One issue found (see inline comment).

paoloantinori:
@EricWittmann good catch, fixed in e05710374. Used `github.base_ref` as you suggested.

Also: the upstream Scalpel fix ([maveniverse/scalpel#46](https://github.com/maveniverse/scalpel/pull/46)) was merged and shipped in [Scalpel 0.3.9](https://github.com/maveniverse/scalpel/releases/tag/0.3.9). Once we bump the Scalpel version in `.mvn/extensions.xml`, the `failSafe=true` gap is also closed on the Scalpel side. We can do that version bump in this PR or a follow-up, your call.

EricWittmann:
We might as well bump it here - no reason to have multiple PRs for this!  :+1:

paoloantinori:
@EricWittmann bumped Scalpel to 0.3.9 in 1c1d837. This picks up the failSafe fix we contributed upstream (scalpel#46).

PR #8971, opened 2026-07-28
fix(build): fail fast with a clear message when building on JDK < 21

### LINKED ISSUE #8966: Build fails on JDK 17 with cryptic UnsupportedClassVersionError from Checkstyle — minimum build JDK undocumented
**Registry version:** `main` (fresh clone)
**OS:** Windows 11, Maven wrapper

**Steps to reproduce:**
1. Fresh clone, JDK 17 on PATH (`maven.compiler.release` is 17, so this seems like a reasonable choice)
2. `./mvnw clean install -Dlocal -DskipTests`

**Actual:** build fails at the root project's `checkstyle:check` (validate phase) with:

UnsupportedClassVersionError: com/puppycrawl/tools/checkstyle/api/CheckstyleException
has been compiled by a more recent version of the Java Runtime (class file version 65.0),
this version of the Java Runtime only recognizes class file versions up to 61.0

i.e. the pinned Checkstyle version needs a Java 21 runtime.

Three things make this hard to diagnose for a newcomer:

- The README's build-tier table says `-Dlocal` skips checkstyle, and the `local` profile does set `checkstyle.skip=true` — but the error happens while class-loading the plugin's dependencies, before the skip property is evaluated, so `-Dlocal` (and explicit `-Dcheckstyle.skip=true`) can't prevent it.
- The enforcer plugin only has `requireMavenVersion` — there's no `requireJavaVersion`, so there's no clean error message.
- The README doesn't state a minimum JDK for building. (With JDK 21 the Checkstyle phase passes — `maven.compiler.release=17` still targets 17 — though the `-Dlocal` build then hits an unrelated test-compilation problem, which I'll report separately.)

**Suggested fix:** document JDK 21+ as the build requirement in the README, and add a `requireJavaVersion` rule to the enforcer so JDK-17 users get a clear message instead of a classloading error. Happy to submit a PR for both if this sounds right.

Aarti-panchal01:
Fixes #8966

## Problem

Building on JDK 17 fails with a cryptic `UnsupportedClassVersionError` from Checkstyle's classes (compiled for a Java 21+ runtime). The error happens while class-loading the plugin's dependencies — *before* `checkstyle.skip` is evaluated — so neither `-Dlocal` nor `-Dcheckstyle.skip=true` can prevent it, and nothing tells the user what's actually wrong. The minimum build JDK also isn't documented.

## Changes

1. **Add a `requireJavaVersion [21,)` enforcer rule** with a message explaining that the build tooling needs a Java 21+ runtime while the produced artifacts still target Java `${maven.compiler.release}` (interpolated, so the message stays correct if the target is ever bumped).
2. **Move the enforcer plugin declaration above checkstyle** in the root pom. Both bind to the `validate` phase, and same-phase executions run in declaration order — without this move, checkstyle would still crash before the new rule ever runs.
3. **Stop skipping the enforcer in the `-Dlocal` profile** (drop `enforcer.skip=true`), so the guard also covers the README quick-start command. The enforcer adds roughly a second to the build.
4. **Document the JDK 21+ build requirement in the README**, right above the quick-start command.

## Before / after (JDK 17, `./mvnw validate -Dlocal`)

Before:

```
UnsupportedClassVersionError: com/puppycrawl/tools/checkstyle/api/CheckstyleException
has been compiled by a more recent version of the Java Runtime (class file version 65.0),
this version of the Java Runtime only recognizes class file versions up to 61.0
```

After (~5s):

```
[ERROR] Rule 1: org.apache.maven.enforcer.rules.version.RequireJavaVersion failed with message:
[ERROR] Building Apicurio Registry requires JDK 21 or newer to run the build tooling (e.g. Checkstyle). The produced artifacts still target Java 17.
```

## Verification

- JDK 17 (Temurin 17.0.19) + `-Dlocal`: fails fast with the message above
- JDK 21 (Temurin 21.0.11): `./mvnw validate` passes both with and without `-Dlocal`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1240,6 +1240,31 @@
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>enforce</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <phase>validate</phase>
+            <configuration>
+              <rules>
+                <requireMavenVersion>
+                  <version>[3.9.8,)</version>
+                </requireMavenVersion>
+                <requireJavaVersion>
+                  <version>[21,)</version>
+                  <message>Building Apicurio Registry requires JDK 21 or newer to run the build tooling (e.g. Checkstyle). The produced artifacts still target Java ${maven.compiler.release}.</message>
+                </requireJavaVersion>
+                <banDuplicatePomDependencyVersions/>
+              </rules>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
         <dependencies>
           <dependency>
@@ -1263,27 +1288,6 @@
               <goal>check</goal>
             </goals>
             <phase>validate</phase>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-enforcer-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>enforce</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <phase>validate</phase>
-            <configuration>
-              <rules>
-                <requireMavenVersion>
-                  <version>[3.9.8,)</version>
-                </requireMavenVersion>
-                <banDuplicatePomDependencyVersions/>
-              </rules>
-            </configuration>
           </execution>
         </executions>
       </plugin>
@@ -1379,7 +1383,6 @@
         <maven.javadoc.skip>true</maven.javadoc.skip>
         <maven.source.skip>true</maven.source.skip>
         <checkstyle.skip>true</checkstyle.skip>
-        <enforcer.skip>true</enforcer.skip>
         <assembly.skipAssembly>true</assembly.skipAssembly>
       </properties>
     </profile>
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -10,6 +10,8 @@ the [README](README.md) and the
 
 ## Build Configuration
 
+**Build requirement:** JDK 21 or newer is required to build the project (the build tooling, e.g. Checkstyle, needs a Java 21+ runtime). The produced artifacts still target Java 17.
+
 This project supports several build configuration options that affect the produced executables.
 
 By default, `mvn clean install` produces an executable JAR with the *dev* Quarkus configuration profile enabled, and *in-memory* persistence implementation.
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ An API/Schema registry - stores and retrieves APIs and Schemas.
 
 Build the project and run the registry with the in-memory storage variant:
 
+**Build requirement:** JDK 21 or newer is required to build the project (the build tooling, e.g. Checkstyle, needs a Java 21+ runtime). The produced artifacts still target Java 17.
+
  ```
  ./mvnw clean install -Dlocal -DskipTests
  cd app/


pom.xml
```diff
@@ -1238,6 +1238,31 @@
...
+            </goals>
+            <phase>validate</phase>
+            <configuration>
+              <rules>
+                <requireMavenVersion>
+                  <version>[3.9.8,)</version>
+                </requireMavenVersion>
+                <requireJavaVersion>
+                  <version>[21,)</version>
+                  <message>Building Apicurio Registry requires JDK 21 or newer to run the build tooling (e.g. Checkstyle). The produced artifacts still target Java ${maven.compiler.release}.</message>
```
paoloantinori, line 1258: The message interpolates `${maven.compiler.release}`. Enforcer does resolve Maven properties in messages, and `maven.compiler.release` is defined in the root pom, so this should render as `...target Java 17.` Worth a quick confirm by running `./mvnw validate` on JDK 21 and eyeballing the enforcer output — if the property isn't on the interpolation path the message will show the literal `${maven.compiler.release}`, which would undercut the clarity this rule is meant to provide.

paoloantinori (commented):
Good, targeted fix for a genuinely confusing build failure. Diagnosing that Checkstyle's plugin classes (compiled for Java 21) class-load during the `validate` phase *before* `checkstyle.skip` is evaluated, so neither `-Dlocal` nor `-Dcheckstyle.skip=true` can prevent the `UnsupportedClassVersionError`, is the right root-cause analysis, and moving the enforcer plugin declaration above Checkstyle's is the correct fix for the ordering dependency. Adding `requireJavaVersion [21,)` with an interpolated message that keeps the build-tooling vs. artifact-target distinction clear, and documenting the JDK requirement in the README, are both solid.

One thing worth verifying before merge.

- **`${maven.compiler.release}` interpolation in the enforcer message is asserted but not verified in this PR.** The message interpolates `${maven.compiler.release}` and the PR description says this keeps it correct if the target is ever bumped. That property is defined in the root pom (it's the standard `maven.compiler.release` used by the compiler plugin), and enforcer messages do support `${property}` interpolation, so this should resolve to `17`. Worth a quick confirm with `./mvnw validate -Prun-it-with-jdk21` or even just grepping the effective enforcer output to make sure it renders as `...target Java 17.` rather than the literal `${maven.compiler.release}` — if the property isn't on the enforcer's interpolation path, the message would be confusingly literal.

Everything else (moving the plugin block, dropping `enforcer.skip=true` from the `-Dlocal` profile so the guard covers the README quick-start command, the ~1s build cost being acceptable) is straightforward and correct.

pom.xml
Aarti-panchal01: Verified!! 
I ran ./mvnw validate -Dlocal on JDK 17 (Temurin 17.0.19) against this branch, and the property resolves in the rendered message:

 [ERROR] Rule 1: org.apache.maven.enforcer.rules.version.RequireJavaVersion failed with message:
 [ERROR] Building Apicurio Registry requires JDK 21 or newer to run the build tooling (e.g. Checkstyle). The produced artifacts still target Java 17.

also one note on verifying via JDK 21: on JDK 21 the rule passes, so the message never prints — the JDK-17 failure output is where the interpolation is actually visible. 
(This is the same output as in the Before/After section of the PR description, captured from a real run.)

Aarti-panchal01:
/ready

Rebased on latest main. #8972 moved the build docs into the new DEVELOPING.md, so the JDK 21 note now lives in DEVELOPING.md's Build Configuration section as well as the new README Quick Start (the pom changes are unchanged). 
Re-verified ./mvnw validate -Dlocal passes on JDK 21 after the rebase.

EricWittmann:
Thanks @Aarti-panchal01 for the excellent contribution! Great root-cause analysis and clean fix — the fail-fast enforcer message is a much better developer experience than the cryptic `UnsupportedClassVersionError`. Merged!

Aarti-panchal01:
Thank you @EricWittmann !!
& thanks for the quick reviews and retries along the way, that made this a really smooth first contribution. 
Glad the enforcer message helps; looking forward to contributing more to the project!

PR #8973, opened 2026-07-28
fix(avro): use resolved references when canonicalizing Avro schemas

### LINKED ISSUE #8968: Avro schemas with artifact references are never canonicalized: canonical hash silently falls back to raw content, breaking dedup
## Description

**Registry Version**: 3.3.1 (also present on snapshot / main @ 3.3.2-SNAPSHOT)
**Persistence type**: sql (kafkasql is affected too, since it delegates to the same inner SQL store)

If an Avro artifact references another artifact, its canonical hash is not actually a canonical hash. It ends up being a hash of the raw bytes, which means canonical-hash lookups stop deduping anything for those schemas. Two identical schemas that differ only in whitespace or a `doc` attribute are stored as two distinct pieces of content.

The cause is that `EnhancedAvroContentCanonicalizer.canonicalize()` accepts `resolvedReferences` but never uses it:

```java
// schema-util/avro/.../EnhancedAvroContentCanonicalizer.java:138
public TypedContent canonicalize(TypedContent content, Map<String, TypedContent> resolvedReferences) {
    String normalisedSchema = normalizeSchema(content.getContent().content()).toString();
    ...
}
```

`normalizeSchema(String)` runs `new Schema.Parser().parse(schemaString)` on the schema on its own. Avro can't parse a schema that refers to a named type it doesn't declare, so this throws `AvroTypeException: Undefined schema: <name>` for any schema with references. I confirmed this against Avro 1.12.1 in isolation:

```
parse(main) alone            -> AvroTypeException: Undefined schema: com.example.Address
parse(reference) then main   -> OK
```

The reason this hasn't shown up as an error is the catch block in `RegistryStorageContentUtils.canonicalizeContent()`:

```java
// app/.../storage/impl/sql/RegistryStorageContentUtils.java:43
} catch (Exception ex) {
    // TODO: We should consider explicitly failing when a content could not be canonicalized.
    log.debug("Failed to canonicalize content: {}", artifactType);
    return content;   // returns the raw, uncanonicalized content
}
```

So the exception is swallowed at debug level, the raw content is returned, and `getCanonicalContentHash()` hashes that instead. From the outside everything looks like it worked.

`AvroCanonicalHashUpgrader` is hit by the same thing from the other direction. It goes through `RegistryContentUtils.canonicalizeContent()`, which does rethrow, and the upgrader catches it per row and logs a warning, so referenced Avro rows get skipped and keep whatever hash they had.

### Environment

- Apicurio Registry main @ 3.3.2-SNAPSHOT, built and inspected locally
- Avro 1.12.1 (the `avro.version` in the root pom)
- Not environment-specific: the defect is in `schema-util/avro`, independent of storage backend,
  deployment mode, or client

### Steps to Reproduce

The most direct reproduction is at the unit level, since the defect is in the canonicalizer itself. Calling the canonicalizer with a schema that has a reference throws, even though the reference is supplied:

```java
String address = "{\"type\":\"record\",\"name\":\"Address\",\"namespace\":\"com.example\","
        + "\"fields\":[{\"name\":\"city\",\"type\":\"string\"}]}";
String person = "{\"type\":\"record\",\"name\":\"Person\",\"namespace\":\"com.example\","
        + "\"fields\":[{\"name\":\"address\",\"type\":\"com.example.Address\"}]}";

new EnhancedAvroContentCanonicalizer().canonicalize(
        TypedContent.create(ContentHandle.create(person), ContentTypes.APPLICATION_JSON),
        Map.of("com.example.Address",
               TypedContent.create(ContentHandle.create(address), ContentTypes.APPLICATION_JSON)));

// -> org.apache.avro.AvroTypeException: Undefined schema: com.example.Address
```

Through the REST API the same defect surfaces as a dedup failure rather than an error, because of the swallowed exception described above:

1. Create an Avro artifact `Address` in group `default`.
2. Create an Avro artifact `Person` whose content references `com.example.Address`, passing that reference in `firstVersion.content.references`.
3. Post the same `Person` schema again, reformatted (extra whitespace, or a `doc` attribute added), with the same reference and `canonical=true`.
4. It is treated as new content rather than matching the existing version.

The same path is used by `POST /apis/registry/v3/search/versions` with `canonical=true`, and by the ccompat API when `apicurio.ccompat.use-canonical-hash` is enabled.

**Expected**: the resolved references are made available to the Avro parser, the schema canonicalizes normally, and two versions of the same referenced schema that differ only in formatting produce the same canonical hash and dedupe.

**Actual**: canonicalization throws, the error is swallowed at debug level, the raw content is hashed instead, and the two versions are stored as separate content. Canonical hashing is effectively disabled for any Avro schema that uses references.

### Logs

With `LOG_LEVEL=DEBUG`, the only trace of the failure on the write path is a single line per write, emitted from `RegistryStorageContentUtils`:

```
DEBUG [io.api.reg.sto.imp.sql.RegistryStorageContentUtils] Failed to canonicalize content: AVRO
```

The exception itself is not logged (the catch block logs the message only, without `ex`), which is part of why this is easy to miss.

The upgrader path is the one place the underlying exception is visible, via the per-row warning in `AvroCanonicalHashUpgrader`:

```
WARN [io.api.reg.sto.imp.sql.upg.AvroCanonicalHashUpgrader] Failed to update canonical hash for contentId <id>.
org.apache.avro.AvroTypeException: Undefined schema: com.example.Address
```

---

Two smaller notes while in this code, happy to split them out if you'd rather:

- This canonicalizer intentionally does not reorder fields (Avro field order is significant for the binary layout and the `order` attribute), which is correct, but it isn't documented anywhere and it differs from the general `ContentCanonicalizer` contract, which says fields should be sorted when ordering isn't important. Worth a javadoc note so the behaviour is discoverable.
- `AvroContentCanonicalizer` appears to be dead code (no references anywhere in the repo). Its fallback branch already had the right idea of seeding the parser with the resolved references.

I have a fix ready. Since existing rows would have stale canonical hashes, it would need an `AvroCanonicalHashUpgrader` run at a new db version, the same way it was handled at db version 104.

1PoPTRoN:
## Summary

`EnhancedAvroContentCanonicalizer` accepted `resolvedReferences` but parsed the schema on its own, so any Avro schema referencing another artifact threw `AvroTypeException: Undefined schema: <name>`. On the write path that exception is swallowed by `RegistryStorageContentUtils.canonicalizeContent()`, which logs at debug and returns the raw content, so the stored "canonical" hash was a hash of the unnormalized bytes. Canonical-hash dedup was effectively off for referenced Avro schemas: two identical schemas differing only in whitespace or a `doc` attribute were stored as separate content.

Fixes #8968

## Root Cause

```java
// schema-util/avro/.../EnhancedAvroContentCanonicalizer.java:138
public TypedContent canonicalize(TypedContent content, Map<String, TypedContent> resolvedReferences) {
    String normalisedSchema = normalizeSchema(content.getContent().content()).toString();
    ...
}
```

`normalizeSchema(String)` runs `new Schema.Parser().parse(schemaString)` on the schema alone. Avro cannot parse a schema that refers to a named type it does not declare. Verified against Avro 1.12.1 in isolation:

```
parse(main) alone            -> AvroTypeException: Undefined schema: com.example.Address
parse(reference) then main   -> OK
```

The failure was invisible because of the `catch` in `RegistryStorageContentUtils` (which carries its own `TODO` noting it should fail explicitly). `AvroCanonicalHashUpgrader` hits the same thing from the other side: it goes through `RegistryContentUtils.canonicalizeContent()`, which does rethrow, and the upgrader catches per row and warns, so referenced Avro rows were skipped and kept stale hashes.

## Changes

- Seed the parser with the resolved references before parsing the schema. References may depend on one another and the map imposes no ordering, so failed entries are retried until a full pass resolves nothing further. Anything still unresolved falls through to the main parse, which reports the offending type name.
- `normalizeSchema(String)` is kept and delegates with `Map.of()`, so the existing public API is unchanged.
- Field order is still deliberately preserved (Avro treats it as part of a record's identity). That behaviour was previously undocumented and differs from the general `ContentCanonicalizer` contract, so it is now stated in the class javadoc and locked in by a test.

## Upgrade impact

Existing rows hold canonical hashes computed from raw content, so they are stale. This bumps the DB version to **108** and re-runs `AvroCanonicalHashUpgrader`, matching how the canonicalizer change at version 104 was handled.

Both halves of the schema-version contract are covered:

- `upgrades/108/{h2,mysql,postgresql,mssql}.upgrade.ddl` — new, each with the `UPGRADER:` directive
- fresh-install `{h2,mysql,postgresql,mssql}.ddl` bumped 107 → 108, so fresh installs do not run the upgrader

Avro **without** references is unaffected: an empty reference map means the seeding loop never runs and the output is byte-identical, so no existing hashes change for those.

## Test plan

- [x] `SchemaNormalizerTest` — 30/30 (26 pre-existing + 4 new)
- [x] `schema-util/avro` full suite — 37/37
- [x] `io.apicurio.registry.noprofile.storage.**` — 21/21, including `AvroCanonicalHashUpgraderTest`
- [x] `./mvnw checkstyle:check -pl schema-util/avro,app` — 0 violations
- [x] Fail→pass confirmed: reverting the one-line fix produces 3 errors, all `Undefined schema: com.example.Address`

New tests in `SchemaNormalizerTest`:

| Test | Covers |
|---|---|
| `canonicalize_SchemaWithReference_ResolvesReferencedType` | the core case; referenced record is inlined |
| `canonicalize_SchemaWithReference_IgnoresFormattingAndDoc` | dedup actually works now |
| `canonicalize_InterdependentReferencesInAnyOrder_Resolves` | `LinkedHashMap` yields a reference before its own dependency, exercising the retry pass |
| `canonicalize_PermutedFields_RemainDistinct` | guards the intentional field-order behaviour |

## Not verified

- I have **not** executed a real 107 → 108 in-place upgrade against a populated database. DDL resolution and `UPGRADER:` directive parsing are verified statically and `AvroCanonicalHashUpgraderTest` passes, but the migration itself has not been run against existing data. Worth a maintainer running `-Pintegration-tests -Premote-sql -Dgroups=migration`.
- MySQL and MSSQL have no integration coverage in CI. Their scripts follow the same shape as h2/postgresql, but nothing exercises them.

## Notes for reviewers

Two adjacent items I deliberately left out of scope, happy to file or fold in if you'd prefer:

- The `catch (Exception)` in `RegistryStorageContentUtils:43` turns a hard parse failure into a silently wrong stored hash and only logs at `debug` without the exception. Its own `TODO` suggests failing explicitly. Changing it affects every artifact type, so it did not belong here.
- `AvroContentCanonicalizer` appears to be dead code (no references anywhere in the repo), and its fallback branch already had the right idea of seeding the parser. It is `public` in a published module, so removing it is an API change that belongs in its own PR.


app/src/main/resources/io/apicurio/registry/storage/impl/sql/db-version
```diff
@@ -1 +1 @@
-107
\ No newline at end of file
+108
```
paoloantinori, line 1: This 107 → 108 bump and the `upgrades/108/` directory collide with PR #8627 (`versionSortKey`), which also bumps to 108 and adds its own `upgrades/108/*.upgrade.ddl` files. Whichever PR merges second will need to re-bump to 109 and rename its upgrade directory, or the second upgrader will never run (the version is already 108). Worth coordinating merge order.

schema-util/avro/src/main/java/io/apicurio/registry/avro/content/canon/EnhancedAvroContentCanonicalizer.java
```diff
@@ -136,7 +181,8 @@ private static String createKey(Schema schema) {
      */
     @Override
     public TypedContent canonicalize(TypedContent content, Map<String, TypedContent> resolvedReferences) {
-        String normalisedSchema = normalizeSchema(content.getContent().content()).toString();
+        String normalisedSchema = normalizeSchema(content.getContent().content(), resolvedReferences)
```
paoloantinori, line 184: This is the call site that was previously throwing and getting silently swallowed by `RegistryStorageContentUtils.canonicalizeContent()`. The fix makes it succeed for referenced schemas, which is the right thing. As a follow-up worth considering: the catch-and-fallback in `RegistryStorageContentUtils` is still there, so any *other* canonicalizer failure (now or in future) will still silently degrade dedup to raw-content hashing. The `TODO` on that catch is worth addressing so this class of silent failure surfaces explicitly.

paoloantinori (commented):
Good catch on a real silent-failure bug. The root-cause analysis is thorough and correct: `EnhancedAvroContentCanonicalizer.normalizeSchema(String)` parsed the schema alone, so any Avro schema referencing another artifact threw `AvroTypeException`, the catch in `RegistryStorageContentUtils.canonicalizeContent()` swallowed it and fell back to raw content, and canonical-hash dedup was effectively off for referenced Avro schemas. Seeding the parser with resolved references before parsing the main schema is the right fix, and the test coverage (reference resolution, formatting/doc insensitivity, interdependent references in any order, field-order preservation) is solid.

One cross-PR coordination issue to resolve before merge, plus two smaller notes.

- **db-version 108 collides with #8627.** Both this PR and #8627 (`versionSortKey`) bump `db-version` from 107 → 108 and add `upgrades/108/*.upgrade.ddl` files with different content. Whichever merges second will conflict on the exact same files (`db-version`, the four `*.ddl` files, and the `upgrades/108/` directory) and will need to re-bump to 109. Worth coordinating with @raghavgupta000001 on merge order so the upgraders don't clobber each other.
- **The silent catch in `RegistryStorageContentUtils.canonicalizeContent()` is not addressed here**, and the PR body correctly flags it as the reason the failure was invisible for so long. The canonicalizer fix means referenced Avro schemas now canonicalize, but any *other* future failure path in the canonicalizer will still be swallowed the same way (debug log + raw fallback). That's pre-existing and arguably out of scope, but the `TODO` the PR body references is still open — worth tracking separately so this class of silent failure doesn't recur.
- **The upgrader re-hashes all Avro content**, including rows already correctly hashed at db-version 104. This is correct (idempotent) and the updated Javadoc now documents both runs (104 nullable-union normalization, 108 resolved-reference seeding), which is good. For deployments with many Avro artifacts it extends the migration write-lock window, but that's the expected cost of fixing the canonical form.

schema-util/avro/src/main/java/io/apicurio/registry/avro/content/canon/EnhancedAvroContentCanonicalizer.java
1PoPTRoN: Agreed. The catch-and-fallback means any future canonicalizer failure degrades to raw-content
hashing just as silently as this one did - it only surfaced because the upgrader path rethrows.

I'll open a separate issue for that `TODO` so it's tracked rather than surviving another few
releases. Out of scope here since changing it affects every artifact type, not just Avro.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/db-version
1PoPTRoN: One wrinkle: the renumber can't be done pre-emptively. `CommonSqlStatements.databaseUpgrade()` (lines 56-63) iterates `fromVersion + 1 ..= toVersion` and resolves a script for *every* intermediate version, so shipping a `109` while no `108` exists would make a 107 → 109 upgrade fail on a missing resource. It has to happen after the first of the two lands.

Happy to go second @raghavgupta000001's PR predates mine by a couple of weeks. If #8627 merges first I'll rebase and move this to `upgrades/109/`; it's a directory rename plus the constant in `db-version` and the four fresh-install DDLs. Equally happy to go first if that's simpler given #8627 currently has conflicts with main. Just say which order you'd prefer.

paoloantinori:
Both points well taken, thanks for the thorough follow-up.

On the `CommonSqlStatements.databaseUpgrade()` loop — good catch, you're right that pre-emptive renumbering isn't an option: the `fromVersion + 1 ..= toVersion` iteration requires a resolvable script for every intermediate step, so a gap would break any upgrade path crossing it. Coordinating so whoever lands second re-bumps is the only sound approach. I'd lean toward letting #8627 go first since it predates this and is already in flight, then rebasing here to `upgrades/109/` — but defer to @raghavgupta000001 on what's least disruptive given the current conflict state.

On the silent catch — agreed it's correctly out of scope here (it spans every artifact type). A separate issue for that `TODO` is the right call so it doesn't keep falling off. Looking forward to the rest of the fix landing.

EricWittmann:
Thanks for the thorough work on this, @1PoPTRoN — the root-cause analysis, the fix, the tests, and the upgrade scripts are all excellent. Really well-prepared contribution.

PR #8974, opened 2026-07-28
feat: add prompt template variable auto-detection and reconciliation

amartyatatspandey:
Related to #8425

## What
Detects variables directly from a PROMPT_TEMPLATE's template text, in
addition to its declared `variables` schema, and reconciles the two so
the "Test Prompt" panel renders an input for every variable that
actually appears in the template — even ones not declared in the schema.

This is the first PR toward the "Prompt Template Playground" scope in
#8425 — specifically the "interactive prompt editor with variable
auto-detection and input form generation" outcome.

## Why
Previously, `PromptTemplateTestPanel` only built input fields from the
artifact's declared `variables` map. If a template referenced a
variable that wasn't declared (e.g. `{{name}}` typed directly into the
template with no schema entry), no input field appeared for it at all
— silently breaking the test/render flow for that variable.

## Changes
- New `promptTemplateVariables.ts`:
  - `extractTemplateVariableNames()` — parses `{{var}}` and
    `{{#if/unless/each/with var}}` block syntax, dedupes, and correctly
    excludes block keywords from being treated as variable names
  - `reconcileTemplateVariables()` — merges detected template variables
    with the declared schema, tagging each as `detected` / `declared` /
    `both`
  - Shared `VariableSchema` type and `TEMPLATE_VARIABLE_REGEX`, used by
    both `PromptTemplateViewer` and `PromptTemplateTestPanel` (previously
    duplicated/drifted as a separate local `PromptVariable` type)
- `PromptTemplateTestPanel.tsx`: builds one input per reconciled
  variable (not just declared ones); shows a small, non-color-only
  label ("not in schema" / "unused") so users can tell detected vs.
  declared vs. unused-declared variables apart
- `PromptTemplateViewer.tsx`: now imports the shared regex/type instead
  of a local duplicate (display behavior unchanged)
- `PromptTemplateVisualizer.tsx`: passes the template text through to
  the test panel

## Note on scope
Confirmed via the actual schema file
(`app/src/main/resources/schemas/prompt-template-v1.json`) that
`templateFormat` (mustache/jinja2/handlebars) doesn't exist as a field
— template syntax is plain `{{var}}` plus simple handlebars-style
blocks, so no format-specific branching is needed here.

## Out of scope (planned as follow-up PRs against #8425)
- Live/debounced auto-render (still click-driven `/render` calls — no
  change to that behavior in this PR)
- `minimum`/`maximum` enforcement in inputs
- Diff view / lifecycle tagging

## Testing
- 10 new unit tests for `extractTemplateVariableNames` and
  `reconcileTemplateVariables` (Vitest) — all passing
- Full existing test suite: 14/14 passing, no regressions
- `npm run lint`: clean (no new warnings/errors)
- `npx tsc --noEmit`: no new type errors introduced (repo has
  pre-existing unrelated errors from generated SDK modules, out of
  scope for this PR)
- No new dependencies added


amartyatatspandey:
Quick note on commit history: this PR shows 2 commits — the original
signed commit plus an automatic merge commit from syncing with `main`
before opening. I know CONTRIBUTING.md calls for atomic, semantic
commits, and I fully agree with that convention. Happy to rebase/squash
this down to a single clean commit if preferred before merge, or since
the repo squash-merges on approval anyway, that'll resolve it
automatically at that point. Just flagging it so it's clear this isn't
an oversight on commit hygiene.

ui/ui-app/src/app/components/promptTemplate/promptTemplateVariables.ts
```diff
@@ -0,0 +1,99 @@
...
+    schema: VariableSchema | undefined;
+    source: "declared" | "detected" | "both";
+};
+
+/**
+ * Same matching semantics as PromptTemplateViewer.highlightVariables.
+ * Captures optional block prefix (group 1) and variable name (group 2).
+ */
+export const TEMPLATE_VARIABLE_REGEX = /\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g;
+
```
paoloantinori, line 47: This regex won't match `{{ name }}` (whitespace inside braces), but the backend pattern in #8977 now does. A template with `{{ name }}` would render server-side but produce no input field here — the same drift #8977 was written to close. Consider `\{\{\s*(#?\/?(?:if|unless|each|with)\s+)?(\w+)\s*\}\}` to stay aligned.

paoloantinori (commented):
Clean implementation — the reconcile logic is well-factored and the test coverage for `extractTemplateVariableNames` / `reconcileTemplateVariables` is thorough. The "not in schema" / "unused" labels are a nice UX touch for the playground.

The main thing worth raising is a **cross-PR consistency gap with the backend**:

- This PR's `TEMPLATE_VARIABLE_REGEX` is `\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}` — it matches `{{name}}` but **not** `{{ name }}` (whitespace inside the braces).
- PR #8977 (the backend refactor, currently open) just made the server-side validator + render endpoint + MCP converter all accept `{{ name }}` via the canonical `\{\{\s*(\w+)\s*\}\}` pattern, specifically to fix a validation hole where `{{ quality }}` was silently treated as literal text.

So after both PRs land, a user typing `{{ name }}` into a template would see it rendered correctly by the server but **no input field** would appear in the Test Prompt panel (because this regex won't detect it). That's the exact class of drift #8977 was written to eliminate. Worth aligning this regex to `\{\{\s*(#?\/?(?:if|unless|each|with)\s+)?(\w+)\s*\}\}` before merge, or at least tracking it as a known follow-up.

Smaller notes:

- `VariableSchema.default` is typed `unknown` and `enum` as `unknown[]`, then cast via `String(opt)` at the render site — reasonable for a UI type boundary, just flagging that non-string enum values (numbers/booleans) will be stringified on round-trip.
- The `useEffect` dependency array now correctly includes `props.template` alongside `props.variables` — good, since detected names derive from the template.
- `key={`${entry.source}-${entry.name}`}` is safe given the reconcile logic dedupes by name, so `source-name` pairs are unique.

Overall the structure is solid; the regex alignment is the one thing I'd want resolved before this ships alongside the backend change.

ui/ui-app/src/app/components/promptTemplate/promptTemplateVariables.ts
amartyatatspandey: Good catch, thank you — verified against #8977's actual diff (not just
description): the backend's `PromptTemplateVariableUtil.VARIABLE_PATTERN`
is exactly `\{\{\s*(\w+)\s*\}\}`, used identically by the validator,
`/render`, and the MCP converter.

Pushed a fix aligning `TEMPLATE_VARIABLE_REGEX` the same way —
`\{\{\s*(#?\/?(?:if|unless|each|with)\s+)?(\w+)\s*\}\}` — so `{{ name }}`
now detects identically to `{{name}}`. Added test cases for whitespace
tolerance on both plain variables and block tags, plus a dedup check for
`{{name}}` / `{{ name }}` appearing together in the same template. 17/17
tests passing, lint clean, no new type errors.

Re: the `enum`/`default` stringification note — agreed that's worth
tracking; happy to take it as a fast-follow if useful, or fold it into
this PR if you'd rather see it resolved now.

ui/ui-app/src/app/components/promptTemplate/promptTemplateVariables.ts
```diff
@@ -0,0 +1,105 @@
...
+ * Shared placeholder matcher used by extractTemplateVariableNames and
+ * PromptTemplateViewer.highlightVariables.
+ *
+ * Captures optional block prefix (group 1) and variable name (group 2).
+ * Optional whitespace inside the braces matches the backend canonical plain-
+ * variable pattern from PromptTemplateVariableUtil
+ * (`\{\{\s*(\w+)\s*\}\}`), while still supporting handlebars-style blocks
+ * (`{{#if x}}`, `{{ #if x }}`, etc.) in one expression for the UI.
+ */
+export const TEMPLATE_VARIABLE_REGEX = /\{\{\s*(#?\/?(?:if|unless|each|with)\s+)?(\w+)\s*\}\}/g;
```
paoloantinori, line 46: This set should also include `else` and `this`. They don't appear as block *prefixes* (so the regex's group-1 alternation doesn't filter them), but they show up as the captured *name* in plain `{{else}}` / `{{this}}` tags. Verified: `extractTemplateVariableNames("{{#if formal}}Mr{{else}}Ms{{/if}}")` returns `["formal", "else"]`, which would render a spurious "else — not in schema" field in the Test Prompt panel. `{{this}}` inside `{{#each}}` blocks has the same problem. Suggested: `new Set(["if", "unless", "each", "with", "else", "this"])`.

ui/ui-app/src/app/components/promptTemplate/promptTemplateVariables.test.ts
```diff
@@ -0,0 +1,100 @@
...
+        expect(extractTemplateVariableNames("{{#unless formal}}hi{{/unless}}")).toEqual(["formal"]);
+        expect(extractTemplateVariableNames("{{#with ctx}}body{{/with}}")).toEqual(["ctx"]);
+    });
+
+    it("handles a mix of plain and block variables", () => {
+        const template = "Hello {{name}}{{#if formal}}, sir{{/if}}. Items: {{#each items}}{{item}}{{/each}}";
+        expect(extractTemplateVariableNames(template)).toEqual(["name", "formal", "items", "item"]);
+    });
+
+    it("tolerates whitespace inside braces for plain variables", () => {
```
paoloantinori, line 34: Worth adding a case here for `else` / `this` exclusion, e.g. `expect(extractTemplateVariableNames("{{#if x}}a{{else}}b{{/if}}")).toEqual(["x"])` and `expect(extractTemplateVariableNames("{{#each items}}{{this}}{{/each}}")).toEqual(["items"])`. The current block-keyword tests only cover the prefix forms, so the `else`/`this` leak isn't caught by the suite.

paoloantinori (commented):
The reconcile/detection logic is well-factored and the unit tests for `extractTemplateVariableNames` / `reconcileTemplateVariables` are genuinely thorough (whitespace, dedup, block-keyword exclusion, order preservation). One gap in the keyword handling is worth addressing before this lands, since it produces a visible UI artifact.

- **`{{else}}` and `{{this}}` leak through as detected variables.** `BLOCK_KEYWORDS` only filters the *block prefix* (`if`/`unless`/`each`/`with`), but `else` and `this` appear as the *variable name* in a plain `{{name}}` tag, so they slip past the filter. I ran the PR's exact regex + `BLOCK_KEYWORDS` set against a couple of realistic templates:
  - `{{#if formal}}Mr{{else}}Ms{{/if}}` → returns `["formal", "else"]` → a spurious **"else"** input appears in the Test Prompt panel (tagged "not in schema").
  - `{{#each items}}{{this}}{{/each}}` → returns `["items", "this"]` → a spurious **"this"** input appears.

  Since the new panel explicitly supports `{{#if}}`/`{{#each}}` blocks (they're tested), the inverse-section `{{else}}` is a natural companion a user would type, so this will surface in practice. The fix is a one-liner: extend `BLOCK_KEYWORDS` to include `else` and `this` (and arguably `lookup`/`log` if you want belt-and-suspenders), e.g. in `promptTemplateVariables.ts`.

- Minor, optional: the regex does not recognize the `{{^}}` inverse-section syntax (equivalent to `{{else}}`), but since it's rare and the regex simply won't match it (no false field generated), that's a non-issue — flagging only for completeness.

Everything else (the `Record` vs `array` reconciliation, the `toDeclaredMap` normalization, the `useEffect` dependency on `props.template`, the `String(opt)` coercion for enum values) looks correct.

ui/ui-app/src/app/components/promptTemplate/promptTemplateVariables.test.ts
amartyatatspandey: Fixed — `BLOCK_KEYWORDS` now includes `else`/`this`/`lookup`/`log`
alongside the existing prefix keywords, with tests for both cases you
flagged (`{{#if x}}a{{else}}b{{/if}}` and `{{#each items}}{{this}}{{/each}}`).

Also ran an independent adversarial pass across the wider Handlebars
construct space (nested blocks, triple-brace, comments, `@index`-style
iteration vars, dotted paths, prototype-name collisions) to check for
further gaps of the same class. No additional logic bugs found — current
behavior already matches the backend's intentional scope (case-sensitive,
`\w+`-only, no dotted-path or `@`-prefixed support). Added regression
tests locking in that behavior plus code comments documenting the two
scope boundaries, so they read as intentional rather than surprising in
future review.

One open question from that pass, genuinely unsure of intended behavior:
if a template uses `{{Name}}` but the declared schema has `name`
(different casing), should those reconcile as the same variable, or stay
distinct given Handlebars' case-sensitivity? Right now they show as two
separate inputs in the Test Prompt panel. Happy to align either way if
you have a preference — didn't want to guess at behavior you didn't ask
for.

Nachiket-Roy:
Hey @amartyatatspandey mentorship issues are intentionally reserved for the Fall term and shouldn't be worked on before applicants are officially accepted.

amartyatatspandey:
@Nachiket-Roy I will keep that in mind. Thank you!

carlesarnal:
Review: Good work. The reconciliation logic is clean, the shared module eliminates duplication, and the test suite is thorough. Paoloantinori feedback on whitespace tolerance and else/this keyword leakage is addressed. Non-blocking suggestions: (1) Memoize extractTemplateVariableNames/reconcileTemplateVariables in PromptTemplateTestPanel.tsx with useMemo keyed on props.template and props.variables to avoid re-parsing on every render. (2) Move inline styles in renderLabel to PromptTemplateTestPanel.css. (3) Update PR title to conventional format e.g. feat: add prompt template variable auto-detection. Overall solid contribution.

carlesarnal:
Expanding on the suggestions above with specifics:

**1. useMemo for reconciliation (PromptTemplateTestPanel.tsx)**

Currently around the top of the component body:
```tsx
const templateNames = extractTemplateVariableNames(props.template || "");
const reconciledVariables = reconcileTemplateVariables(templateNames, toDeclaredMap(props.variables));
```

This re-runs the regex extraction on every render (including when the user types in an input field, or when renderedOutput/validationErrors state changes). Suggest:

```tsx
const reconciledVariables = useMemo(() => {
    const names = extractTemplateVariableNames(props.template || "");
    return reconcileTemplateVariables(names, toDeclaredMap(props.variables));
}, [props.template, props.variables]);
```

**2. Inline styles in renderLabel**

The span in `renderLabel` uses `style={{ display: "inline-flex", alignItems: "center", gap: "0.5rem", flexWrap: "wrap" }}`. This would be cleaner as a CSS class in the existing `PromptTemplateTestPanel.css`.

**3. PR title**

Current: `Feat/prompt template variable detection`
Suggested: `feat: add prompt template variable auto-detection and reconciliation`

None of these are blockers. The maintainer feedback from paoloantinori has been fully addressed — both the whitespace-tolerant regex (`\s*` inside braces) and the BLOCK_KEYWORDS expansion (`else`, `this`, `lookup`, `log`) are in place with matching tests.

amartyatatspandey:
> Expanding on the suggestions above with specifics:
> 
> **1. useMemo for reconciliation (PromptTemplateTestPanel.tsx)**
> 
> Currently around the top of the component body:
> 
> ```tsx
> const templateNames = extractTemplateVariableNames(props.template || "");
> const reconciledVariables = reconcileTemplateVariables(templateNames, toDeclaredMap(props.variables));
> ```
> 
> This re-runs the regex extraction on every render (including when the user types in an input field, or when renderedOutput/validationErrors state changes). Suggest:
> 
> ```tsx
> const reconciledVariables = useMemo(() => {
>     const names = extractTemplateVariableNames(props.template || "");
>     return reconcileTemplateVariables(names, toDeclaredMap(props.variables));
> }, [props.template, props.variables]);
> ```
> 
> **2. Inline styles in renderLabel**
> 
> The span in `renderLabel` uses `style={{ display: "inline-flex", alignItems: "center", gap: "0.5rem", flexWrap: "wrap" }}`. This would be cleaner as a CSS class in the existing `PromptTemplateTestPanel.css`.
> 
> **3. PR title**
> 
> Current: `Feat/prompt template variable detection` Suggested: `feat: add prompt template variable auto-detection and reconciliation`
> 
> None of these are blockers. The maintainer feedback from paoloantinori has been fully addressed — both the whitespace-tolerant regex (`\s*` inside braces) and the BLOCK_KEYWORDS expansion (`else`, `this`, `lookup`, `log`) are in place with matching tests.

@carlesarnal 
Hi Carles — apologies for the delay, I was dealing with dengue and
have been out of commission for the past week or so. Back now, and
all three suggestions are addressed:

1. `reconciledVariables` now computed inside `useMemo`, keyed on
   `props.template`/`props.variables` — no longer re-runs on unrelated
   renders.
2. `renderLabel`'s inline styles moved to a new `.label-inline` class in
   `PromptTemplateTestPanel.css`, following the file's existing nested
   kebab-case convention.
3. PR title updated to conventional-commit format.

24/24 tests passing, lint clean, no new type errors, no behavioral or
visual change from the memoization — only *when* it recomputes.

EricWittmann:
@amartyatatspandey there are now some conflicts in the PR - can you fix those up?

amartyatatspandey:
> @amartyatatspandey there are now some conflicts in the PR - can you fix those up?

@EricWittmann 

Hi Eric! 

Resolved — should be conflict-free now. Summary of how the two conflicts
were reconciled:

- **PromptTemplateTestPanel.tsx**: kept this PR's reconciliation-based
  value-seeding (detected + declared + both, via `reconciledVariables`),
  and adopted main's improved default-value formula and
  `coerceEnumValue` usage on top of it.
- **PromptTemplateViewer.tsx**: adopted main's fuller Handlebars
  highlighter (`PromptTemplateViewer.utils.ts`) in place of this PR's
  inline highlighter, since it has broader coverage (comments,
  triple-stache, block classification). Kept `VariableSchema` as the
  type source of truth over the local `PromptVariable` — confirmed no
  in-use fields were dropped (`VariableSchema` is a strict superset).

`promptTemplateVariables.ts` itself (extraction/reconciliation logic)
was untouched by the merge. 43/43 promptTemplate tests passing, lint
clean, no new type errors.

EricWittmann:
Waiting on CI...

amartyatatspandey:
also @EricWittmann 

Now that this one's almost taken care of, I'd like to keep going on
the rest of #8425's scope — live/debounced rendering next, then diff
view and lifecycle tagging. Happy to open the next PR whenever makes sense
sense, or if there's a different piece you'd rather see tackled first,
Let me know, and I'll start there instead.

EricWittmann:
Thanks for the contribution, @amartyatatspandey! Clean implementation with thorough test coverage. Merged.

amartyatatspandey:
> Thanks for the contribution, @amartyatatspandey! Clean implementation with thorough test coverage. Merged.

@EricWittmann 

Thank you! Glad it landed cleanly.

Would like to keep going on the rest of #8425 — planning to pick up
live/debounced rendering next (Test Panel currently only renders on
manual click), then diff view and lifecycle tagging after that. Let me
know if that order works or if you'd rather see something else
tackled first.

Extremely excited and ready to help out!

amartyatatspandey:
@EricWittmann 
@carlesarnal 

I am loving this journey of contributions!

https://github.com/Apicurio/apicurio-registry/pull/9619 ->Second PR in the #8425 series, adding live debounced
rendering + request cancellation to the Test Panel.

Worth flagging: #9351 landed on the main partway through this PR's
development, refactoring the Test Panel's state into
`usePromptTemplateTestPanelState` and fixing a version-switch race
condition. That's a different problem from this PR's scope, so rather
than duplicate or fight that work, I rebuilt the debounce timer,
required-fields gate, and `AbortController` cancellation on top of
their hook and existing staleness guard.

PR #8976, opened 2026-07-28
test(rest): use typed data contract request bodies

### LINKED ISSUE #8863: Refactor DataContractsResourceTest to use type-safe DTOs instead of raw JSON strings
# Summary

In `app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java`, there are multiple instances where REST-assured tests are utilizing raw, hardcoded JSON strings to represent the request bodies for contract endpoints (e.g., `/contract/metadata`, `/contract/ruleset`, `/contract/status`).

While this works currently, it violates standard Clean Code practices and introduces technical debt. Bypassing type safety means that if the OpenAPI specification changes, these tests will silently pass compilation and fail opaquely at runtime.

# Proposed Approach

The backend DTO classes for the contract endpoints are already being generated from the OpenAPI spec and are available in the `io.apicurio.registry.rest.v3.beans.*` package.

We should refactor the tests in `DataContractsResourceTest.java` to use these generated Java classes instead of hardcoded strings.

Use the following DTOs:

- `EditableContractMetadata` for metadata endpoints.
- `ContractRuleSet` and `ContractRule` for ruleset endpoints.
- `ContractStatusTransition` for status endpoints.

By passing these DTOs to REST-assured's `.body()`, we leverage Jackson to handle serialization while ensuring 100% compile-time safety and schema validation alignment.

# Relevant Files

- `app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java`

hcthakur2004:
## Summary
- Refactor data contract metadata and status requests to use generated v3 DTOs
- Replace contract ruleset request JSON with `ContractRuleSet` and `ContractRule` objects
- Keep existing assertions and endpoint coverage unchanged

Fixes #8863

## Testing
- `git diff --check`
- Not run: `./mvnw.cmd -pl app -Dtest=DataContractsResourceTest test -DfailIfNoTests=false --no-transfer-progress` (local `JAVA_HOME` is not defined correctly)

### DIFF
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
@@ -1,11 +1,17 @@
 package io.apicurio.registry.noprofile.rest.v3;
 
 import io.apicurio.registry.AbstractResourceTestBase;
+import io.apicurio.registry.rest.v3.beans.ContractRule;
+import io.apicurio.registry.rest.v3.beans.ContractRuleSet;
+import io.apicurio.registry.rest.v3.beans.ContractStatusTransition;
+import io.apicurio.registry.rest.v3.beans.EditableContractMetadata;
+import io.apicurio.registry.rest.v3.beans.Params;
 import io.apicurio.registry.types.ArtifactType;
 import io.apicurio.registry.types.ContentTypes;
 import io.quarkus.test.junit.QuarkusTest;
 import org.junit.jupiter.api.Test;
 
+import java.util.List;
 import java.util.UUID;
 
 import static io.restassured.RestAssured.given;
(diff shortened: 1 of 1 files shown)


app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
```diff
@@ -21,6 +27,41 @@ public class DataContractsResourceTest extends AbstractResourceTestBase {
...
+        ruleset.setDomainRules(domainRules);
+        ruleset.setMigrationRules(migrationRules);
+        return ruleset;
+    }
+
+    private static ContractRule rule(String name, String kind, String type, String mode) {
+        ContractRule rule = new ContractRule();
+        rule.setName(name);
+        rule.setKind(ContractRule.Kind.fromValue(kind));
+        rule.setType(type);
```
paoloantinori, line 54: This is the one spot where the DTO conversion could change wire output. `Params.setAdditionalProperty` only matches the old `"params": {"targetField": "newField"}` if the generated `Params` class uses `@JsonAnyGetter`/`@JsonAnySetter` to flatten additional props to the top level. Worth eyeballing the generated `Params.java` to confirm it doesn't nest under an `additionalProperties` key.

paoloantinori (commented):
Clean test refactor — the helper methods (`contractMetadata`, `statusTransition`, `ruleset`, `rule`, `params`) tidy up the test bodies nicely and the assertions are preserved verbatim, so endpoint coverage is unchanged. The switch from hand-rolled JSON to the generated DTOs is the right direction for long-term maintainability.

A couple of notes worth flagging:

- **Tests are not confirmed to pass.** The PR body states the local build was skipped because `JAVA_HOME` is unset. Since this PR is entirely a test-file change, the test run is the only meaningful verification — I'd want to see at least one green CI run on `DataContractsResourceTest` (or the maintainer's local run) before merge, since the JSON-to-DTO conversion has serialization nuances (see next point).

- **`Params` serialization is the one subtle risk.** The old body sent `"params": {"targetField": "newField"}` as a plain JSON object. The new code builds `Params` via `setAdditionalProperty("targetField", "newField")`. `Params` is generated from the OpenAPI spec and typically uses `@JsonAnyGetter`/`@JsonAnySetter` to flatten additional properties into the top level of the JSON object — so the wire output should match. Worth confirming the generated `Params` class does flatten (rather than nesting under a `additionalProperties` key), otherwise that one test (`testSetAndGetArtifactContractRuleset`) would POST a different body shape and the `targetField` assertion could pass for the wrong reason or fail.

- Minor: the duplicated `positiveAmountRule` construction appears identically in both `testExecuteContractRules_PassesWithValidData` and `testExecuteContractRules_FailsWithInvalidData`. Not worth extracting into a shared helper for two call sites, just noting the pattern repeats if more tests adopt it.

Overall this is a low-risk, high-readability improvement once CI confirms the tests stay green.

hcthakur2004:
Closing this because another contributor already had work in progress for the linked issue. I’ll look for a different available issue instead.

PR #8977, opened 2026-07-28
fix(schema-util): parse {{variable}} with one canonical pattern

### LINKED ISSUE #8877: bug: prompt template {{variable}} parsing is inconsistent across validator, render endpoint, and MCP converter

PRAteek-singHWY:
> Supersedes #8923, which the one-PR-at-a-time bot closed automatically a few seconds
> after I opened it (#8841 was still open at the time). GitHub declines to reopen it —
> #8921 and #8923 share the same head/base pair, so the reopen returns a 422. Same branch,
> same commit, now rebased on current `main` including #8841. Opening fresh so
> @EricWittmann and @paoloantinori have something reviewable, per the follow-up they
> mentioned on #8841.

## Summary

Fixes #8877.

The `{{variable}}` placeholder syntax of Prompt Template artifacts was parsed by three
separate components, each with its own regular expression. They disagreed, so the same
template could mean three different things depending on which path it went through.

The most visible consequence is a validation hole: the validity rule's "template variable
used but not defined in `variables`" check silently never fired for a placeholder written
with spaces inside the braces. A template referencing an undefined `{{ quality }}` passed
`FULL` validation, then rendered with the literal `{{ quality }}` still in the output.

This PR pins one canonical pattern in `schema-util/common` and routes all three components
through it, so they cannot drift apart again.

## Root Cause

Three components, three answers to "what is a variable here":

| Component | Pattern | `{{ name }}` resolves to |
|---|---|---|
| `PromptTemplateContentValidator` (`schema-util/common`) | `\{\{(\w+)\}\}` | nothing — not a variable |
| `PromptRenderingService` (`app`, REST `/render`) | `\{\{([^}]+)\}\}` then `.trim()` | `name` |
| `PromptTemplateConverter` (`mcp`) | see below | nothing — left as literal text |

The validator's `\w+` cannot match across the spaces, so `extractTemplateVariables` returns
an empty list for `"Hello {{ quality }}"`. `validateTemplateVariables` then has nothing to
cross-check against the `variables` object, and the rule passes. The same template without
spaces is correctly rejected — that case is already covered by the existing
`testUndefinedVariable`.

Worth flagging one thing I found while fixing the MCP side: `PromptTemplateConverter`
declared a `VARIABLE_PATTERN` field, but nothing ever used it. The actual substitution was
a loop over the supplied arguments doing

```java
String placeholder = "\\{\\{" + entry.getKey() + "\\}\\}";
rendered = rendered.replaceAll(placeholder, Matcher.quoteReplacement(value));
```

The observable behaviour matches what the issue describes — `{{ name }}` is not substituted
— but the mechanism is a literal `{{key}}` match, not the declared regex. It also builds a
regex out of the argument name, so a `.` in a key matched any character: an argument named
`a.b` would substitute an unrelated `{{axb}}` placeholder. Both go away with this change.

## The canonical grammar

Following the direction @paoloantinori gave on the issue thread: one pattern, in the shared
layer, used by everyone.

```java
Pattern.compile("\\{\\{\\s*(\\w+)\\s*\\}\\}");
```

A word-character name with optional surrounding whitespace, so `{{name}}` and `{{ name }}`
both resolve to `name`. Handlebars-style control blocks (`{{#if x}}`, `{{/if}}`) are not
matched, because `#` and `/` are not word characters — so this change does not disturb
conditional blocks, which are a separate concern (#8728).

## Changes

- **New** `schema-util/common/.../content/util/PromptTemplateVariableUtil.java`
  - `VARIABLE_PATTERN` — the single definition of the placeholder syntax.
  - `extractVariableNames(String)` — variable names in first-seen order, deduplicated,
    empty list for a null template.
  - `substituteVariables(String, Function<String, String>)` — replaces placeholders with
    whatever the resolver returns. A resolver returning `null` means "no value", and the
    placeholder is written back verbatim. That hook is what lets both call sites keep their
    own (differing) idea of what an unresolved variable means, while sharing the parsing.

- `schema-util/common/.../rules/validity/PromptTemplateContentValidator.java`
  - Dropped the local `TEMPLATE_VARIABLE_PATTERN`; `extractTemplateVariables` now delegates
    to the shared helper. The method stays `public static` — `PromptTemplateCompatibilityChecker`
    calls it.

- `app/.../services/PromptRenderingService.java`
  - Dropped the local `VARIABLE_PATTERN`; `substituteVariables` now delegates. Unresolved
    variables still keep their placeholder, exactly as before.

- `mcp/.../PromptTemplateConverter.java`
  - Removed the unused `VARIABLE_PATTERN` and replaced the per-argument `replaceAll` loop
    with a single pass through the shared helper. `{{#if}}` handling is untouched and still
    runs after substitution, as it did before.

- `mcp/pom.xml`
  - Added `apicurio-registry-schema-util-common`. See the note below.

## Behaviour changes worth reviewing

1. **`/render` no longer substitutes dotted or hyphenated names.** Tightening the endpoint
   from `[^}]+` to `\w+` is the part of this change that removes a capability: `{{user.name}}`
   and `{{first-name}}` were substituted by `/render` and are now literal text. That was
   already true for the validator and the MCP converter, so this makes `/render` consistent
   with the other two rather than the other way around — but it is the direction choice the
   issue asked about, so it deserves a look. Widening everything to accept dotted names
   instead is a one-line change to the shared pattern if you would rather go that way.

2. **The compatibility checker gets stricter, in the same direction as the fix.**
   `PromptTemplateCompatibilityChecker` uses `extractTemplateVariables` to detect
   `VARIABLE_REMOVED_BUT_USED`. It now also catches the case where the removed variable is
   still referenced as `{{ varName }}`, which it previously missed.

3. **A new module dependency: `mcp` → `schema-util-common`.** This is the one debatable bit.
   The MCP module previously depended only on the Java SDK, so sharing the pattern with it
   means adding a dependency. `schema-util-common` is fairly light (its heavier transitives
   in `apicurio-registry-common` are `provided`-scoped), and duplicating the constant would
   defeat the point of the issue — but if you would rather not grow the MCP server's
   classpath, say so and I will drop the MCP half into a follow-up.

## Test plan

Splitting these honestly into the ones that actually reproduce the bug and the ones that
are there to stop it coming back.

**Reproduce the bug — these fail on `main`:**

- `PromptTemplateContentValidatorTest.testUndefinedVariableWithWhitespace` — the case from
  the issue. `{{ quality }}` is used but not defined, and `FULL` validation now rejects it.
  On `main` the validator extracts nothing from that placeholder, raises no violation, and
  the `assertThrows` fails.
- `PromptTemplateContentValidatorTest.testExtractTemplateVariablesNormalisesWhitespace` —
  pins the exact list the extraction helper returns, since `PromptTemplateCompatibilityChecker`
  depends on it. On `main` the padded spellings are missing from the result.
- `PromptTemplateConverterTest.testRenderVariableWithWhitespace` — `{{ name }}` is now
  substituted by the MCP path instead of surviving into the output as literal text.
- `PromptTemplateConverterTest.testVariableNameWithRegexMetacharacterDoesNotMatchWildly` —
  an argument named `a.b` no longer substitutes an unrelated `{{axb}}` placeholder.
- `PromptRenderingServiceTest.testDottedNameIsNotAVariable` — this one fails on `main`
  because `main` is the more permissive side. It documents behaviour change (1) above.

**Guard the behaviour that must not change — these already pass on `main`:**

- `PromptRenderingServiceTest` — padded spellings resolving identically, unresolved padded
  placeholders preserved verbatim, `{{#if}}` markers left alone by substitution.
- `PromptTemplateContentValidatorTest.testDefinedVariableWithWhitespaceIsAccepted` and
  `testConditionalBlockMarkersAreNotVariables` — no new false positives, and control syntax
  is not mistaken for a variable.
- The existing `testWhitespaceInVariableName` and `testUndefinedVariable` still pass; the
  canonical pattern preserves everything that already worked on both sides.

**New test classes:**

- `PromptTemplateVariableUtilTest` (`schema-util/common`, 12 cases) — direct coverage of the
  shared helper: whitespace variants, first-seen ordering and deduplication, null template,
  control markers, `$` and `\` in substituted values, and the null-resolver contract.
- `PromptTemplateConverterTest` (`mcp`, 11 cases) — the converter had no test coverage at
  all. Covers substitution with and without whitespace, unknown variables, null values,
  null args, regex metacharacters in both values and argument names, and the existing
  `{{#if}}` truthy/falsy behaviour.

Checks run locally, on current `main` (includes #8841 and the 3.3.1 release):

- [x] `./mvnw test -pl schema-util/common,schema-util/util-provider` — 34 tests, all pass
      (`PromptTemplateVariableUtilTest` 12, `PromptTemplateContentValidatorTest` 22).
- [x] `./mvnw test -Pfull -pl mcp -Dtest=PromptTemplateConverterTest` — 11 tests, all pass.
      This also confirms the new `mcp` → `schema-util-common` dependency resolves.
- [x] `./mvnw checkstyle:check` clean (0 violations) on `schema-util/common`,
      `schema-util/util-provider`, `mcp` and `app`.
- [ ] `app` — `PromptRenderingServiceTest` is a `@QuarkusTest` that I have not run locally.
      Please let CI cover it.
- [ ] Storage variants: not applicable. Nothing here touches `storage/impl/`; the change is
      confined to template-string parsing.

## Notes

- Related but separate: #8728 (`/render` does not process `{{#if}}` blocks) and #8669 (null
  `template` field). This PR deliberately leaves both alone — it only makes the three
  components agree on what a variable *is*.
- PR #8690 is open against `PromptRenderingService.java` and its test, but it changes
  `extractTemplateText` (null/non-string `template` field), not `substituteVariables`, so
  there is no functional overlap. Whichever merges second may need a trivial rebase in the
  test file.
- There is actually a fourth parser I did not touch:
  `ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx` highlights
  placeholders with `/\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g`, which is also
  whitespace-intolerant, so `{{ name }}` is not highlighted as a variable in the preview.
  I left it out of scope: it is a separate build system, it is presentation rather than the
  parser of record, and PR #8917 is already open against that component. Happy to follow up
  there once the direction here is settled.
- The class javadoc on `PromptTemplateConverter` calls it "the Java equivalent of
  mcp-converter.ts", but there is no `mcp-converter.ts` in the tree any more. Left as-is
  rather than widening the diff.


paoloantinori (commented):
Solid refactor — consolidating three divergent `{{variable}}` patterns into one `PromptTemplateVariableUtil` is the right call, and the validation hole it closes (the "used but not defined" check silently skipping `{{ quality }}`) is a genuine bug fix. Test coverage is thorough across all three call sites.

One behavior change is worth confirming before merge:

- **The REST render endpoint is now stricter than before.** `PromptRenderingService` previously used `\{\{([^}]+)\}\}`, which matched any non-`}` content, so placeholders like `{{user.name}}`, `{{my-var}}`, or `{{a b}}` were substituted if a matching key existed in `variables`. The canonical `\{\{\s*(\w+)\s*\}\}` only matches word-character names, so all of those now render as literal text. The PR documents this as "aligning" to the validator/MCP behavior, which is accurate — but the render endpoint is the one site where this narrows what previously worked, and stored artifacts with non-`\w` placeholders will silently stop rendering after this change. Worth confirming there's no known usage of dotted/hyphenated names, and consider a release-notes line so users aren't surprised.

Smaller notes:

- Java's `\w` is ASCII-only by default, so `{{café}}` or `{{名前}}` won't match. Almost certainly fine for prompt templates, just flagging it.
- Good catch on `Matcher.quoteReplacement` in `substituteVariables` — the old MCP converter's `replaceAll(placeholder, ...)` left the *search* string unquoted as a regex, which worked only by accident.
- `extractTemplateVariables` is also consumed by `PromptTemplateCompatibilityChecker`, so this change normalizes compatibility checking too (e.g. `{{name}}` vs `{{ name }}` are now correctly treated as the same variable). That's an improvement, just noting the blast radius extends there.

Overall this looks good to merge once the dotted-name narrowing is confirmed as acceptable.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -256,27 +253,11 @@ private String getTypeName(Object value) {
      * Substitute variables in the template text using {{variable}} syntax.
      */
```
paoloantinori, line 352: This is the one call site where the behavior narrows. The old `\{\{([^}]+)\}\}` matched `{{user.name}}`, `{{my-var}}`, etc.; the canonical `\w+` pattern leaves those as literal text. The validator and MCP converter were already `\w+`, so only the REST render endpoint is affected. If any stored artifacts use non-word-character placeholders, they'll silently stop rendering here. Worth a quick check and possibly a release note.

schema-util/common/src/main/java/io/apicurio/registry/content/util/PromptTemplateVariableUtil.java
```diff
@@ -0,0 +1,95 @@
...
+ * that parses the syntax should go through this class so that the answer stays the same everywhere.
+ *
+ * The grammar is a word-character name with optional whitespace inside the braces, so both
+ * <code>{{name}}</code> and <code>{{ name }}</code> resolve to <code>name</code>. Handlebars-style
+ * control blocks such as <code>{{#if x}}</code> and <code>{{/if}}</code> are deliberately not
+ * matched, because <code>#</code> and <code>/</code> are not word characters.
+ */
+public final class PromptTemplateVariableUtil {
+
+    /**
```
paoloantinori, line 50: `\w` in Java regex is ASCII-only (`[a-zA-Z0-9_]`), so Unicode identifiers like `{{café}}` won't be treated as variables. Presumably fine for prompt templates, but if internationalized names are ever expected, `\p{IsAlphabetic}\p{IsDigit}` (or UNICODE_CHARACTER_CLASS) would be the upgrade path.

PRAteek-singHWY:
Thanks for the thorough review @paoloantinori, this was really helpful. Answering the merge
blocker first, then the smaller notes, then one thing your review of #8974 made me realise I
should fix here.

## The dotted-name narrowing

You are right that this is the one call site where behaviour narrows, and I went looking for
real usage rather than guessing. I audited every `{{...}}` placeholder in the repo and split
them by subsystem:

**Shipped prompt templates.** Every placeholder is a plain word-character name, so all of them
are unaffected:

| File | Placeholders |
|---|---|
| `a2a-real-world-integration/.../sentiment-agent-prompt.yaml` | `customerMessage` |
| `.../analyzer-agent-prompt.yaml` | `customerMessage`, `sentimentContext` |
| `.../response-generator-prompt.yaml` | `customerMessage`, `analysisContext`, `sentimentContext` |
| `.../translator-agent-prompt.yaml` | `responseText`, `targetLanguage` |
| `llm-artifact-types/.../apicurio-support-chat-prompt.yaml` | `question`, `system_prompt`, `conversation_history` |
| `llm-artifact-types/.../apicurio-support-system-prompt.yaml` | `additional_context`, `supported_artifact_types` |
| `llm-artifact-types/.../mcp-enabled-code-review-prompt.yaml` | `code`, `language`, `context`, `focus_area` |

Note the underscored names (`system_prompt`, `focus_area`) are safe, since `_` is a word
character. The docs are consistent with this too: `assembly-llm-artifact-types-implementation.adoc`
and `assembly-artifact-reference.adoc` only ever show `{{customerMessage}}`, `{{question}}`,
`{{missing}}` and `{{variable}}`.

**Dotted and hyphenated placeholders that do exist, and why they are not affected.** There are
some in the tree, and I want to be explicit that I checked them rather than skipped them:

- `{{rule-types}}`, `{{rule-configs}}`, `{{product-name}}` in `cli/`. These are CLI help-text
  branding tokens substituted by `BrandedCommandLineProducer`, a completely separate mechanism
  that never goes through prompt template parsing.
- `{{ to.label }}`, `{{ to.description }}` in `ui-editors/`. Angular Formly templates.
- `{{ github.sha }}`, `{{ secrets.* }}` in `docs/cross-repo-testing-plan.md`. GitHub Actions
  expressions inside a planning doc.
- `{{user.name}}` in `PromptRenderingServiceTest`. That is the test I added in this PR
  (`testDottedNameIsNotAVariable`) which pins the new behaviour deliberately.

So there is no known usage of dotted or hyphenated names in an actual prompt template, and I am
happy to confirm the narrowing as intended rather than widen the pattern. Agreed on the
release-notes line, since the audit only covers what is in this repo and cannot speak for
artifacts users already have stored. Happy to add the wording here or in a docs follow-up,
whichever you prefer.

## The `{{else}}` gap, which your #8974 review made me spot here

Your note on #8974 about `{{else}}` and `{{this}}` leaking through as detected variables applies
to the backend too, and it turns out it is worse on this side than in the UI.

`{{else}}` is a bare word, so the canonical pattern matches it and reports `else` as a variable.
The validity rule then flags it as used-but-not-defined. That is not a regression I introduced,
the old validator pattern `\{\{(\w+)\}\}` matched `{{else}}` just the same, but it does mean a
shipped example currently cannot pass `FULL` validity:
`examples/a2a-real-world-integration/src/main/resources/prompts/translator-agent-prompt.yaml`
uses `{{else}}` inside an `{{#if targetLanguage}}` block and declares only `responseText` and
`targetLanguage`.

I confirmed it rather than reasoning about it. With the keyword set emptied, extraction on that
template returns `[responseText, targetLanguage, else]`, and the validator throws
`RuleViolationException: Invalid Prompt Template`. With the fix in place it returns
`[responseText, targetLanguage]` and validation passes.

The reason I fixed it in this PR instead of leaving it as a follow-up is the cross-PR angle you
raised. If #8974 adds `else` and `this` to `BLOCK_KEYWORDS` on the UI side and the backend keeps
treating them as variables, that reintroduces exactly the drift this PR exists to remove, just
pointing the other way. Aligning both sides at once seemed better than fixing one and filing an
issue for the other. If you would rather keep this PR to the whitespace unification and handle
the keywords separately, say so and I will pull it out into its own PR.

What changed:

- `PromptTemplateVariableUtil` gained a `CONTROL_KEYWORDS` set (`else`, `this`) and a public
  `isControlKeyword(String)`. Both `extractVariableNames` and `substituteVariables` respect it,
  so a control keyword is never reported as a variable and never substituted even if a resolver
  would supply a value.
- Four new cases in `PromptTemplateVariableUtilTest`, including one built from the translator
  example's exact shape, and two in `PromptTemplateContentValidatorTest` asserting that
  `{{else}}` and `{{this}}` no longer produce violations.

I left `lookup` and `log` out for now. You called them belt-and-suspenders on the other PR and I
would rather match whatever set #8974 lands on than guess at a different one.

## One asymmetry that will survive both PRs

Worth tracking separately, since it is the same class of problem. The #8974 regex has an optional
block-prefix group, `\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}`, so it captures the block
subject: `{{#if formal}}` yields `formal`. The backend pattern does not, because `#` stops the
match before the name. That is why your `["formal", "else"]` example returns only `["salutation"]`
style results on this side.

The good news is it cannot produce a validation error today, because the validity rule only checks
used-but-not-defined and has no declared-but-unused check, so declaring a block subject is
harmless and not declaring it is also harmless. But it does mean the UI and the server still
disagree about whether a block subject counts as a variable, and the Test Prompt panel will show
an input for something the server never treats as one. Not something I want to change here without
a decision on which behaviour is correct. Happy to open an issue for it if you think it is worth
tracking.

## Smaller notes

- The ASCII-only `\w` point is a fair catch. I have documented it on the class javadoc, including
  `UNICODE_CHARACTER_CLASS` as the upgrade path, and deliberately not changed the behaviour: the
  validity rule and the MCP converter were both already ASCII-only, so widening it here would be a
  new behaviour change rather than the alignment this PR is meant to be.
- On `quoteReplacement`, thanks. To be precise about the old bug, the search string was the
  unquoted side: `replaceAll("\\{\\{" + key + "\\}\\}", ...)` built a regex out of the argument
  name, so a key containing `.` matched any character and an argument called `a.b` would substitute
  an unrelated `{{axb}}`. The replacement side was already quoted.
- Confirmed on the compatibility checker blast radius, and I agree it is an improvement.
  `PromptTemplateCompatibilityChecker` line 72 calls
  `PromptTemplateContentValidator.extractTemplateVariables`, which now delegates to the shared
  util, so `{{name}}` and `{{ name }}` are finally the same variable there as well.

Full suite on the touched modules is green: 16 in `PromptTemplateVariableUtilTest`, 24 in
`PromptTemplateContentValidatorTest`, 11 in `PromptTemplateConverterTest`, and checkstyle is clean
on `schema-util/common`, `schema-util/util-provider`, `mcp` and `app`. The `app`
`PromptRenderingServiceTest` is still the one I am relying on CI for.

paoloantinori:
Thanks for the detailed write-up — the per-subsystem placeholder audit is exactly the kind of evidence I was hoping for, and it's convincing. A few responses:

**Dotted-name narrowing.** Confirmed, I'm happy to land the `\w+` narrowing as intended. The audit covers every `{{...}}` occurrence in-tree and none of the actual prompt-template files use dotted or hyphenated names, so narrowing is strictly an alignment of the renderer with what the validator and converter already did. The `system_prompt` / `focus_area` underscored names being safe (`_` is a word char) is a good thing to have spelled out.

**`{{else}}` / `{{this}}` fixed here.** Fully agree with fixing it in this PR rather than as a follow-up. The cross-PR drift argument is the right one: if #8974 adds them to `BLOCK_KEYWORDS` on the UI side and the backend keeps treating them as variables, we've recreated exactly the divergence this PR exists to kill. And you've shown it's not hypothetical — the shipped `translator-agent-prompt.yaml` couldn't pass `FULL` validity before the fix. Catching `lookup`/`log` later to match whatever set #8974 lands on is the pragmatic call.

**Block-subject asymmetry (`{{#if formal}}` → `formal`).** Yes, please open an issue for it. You're right it can't produce a validation error today (no declared-but-unused check), but the UI/server disagreement over whether a block subject counts as a variable is the same class of bug this PR is about, just deferred. Worth tracking so it doesn't get lost.

**Release-notes line.** I'm fine with a one-liner in the docs/release notes. The in-tree audit is convincing but, as you note, can't speak for artifacts users already have stored — a short "the render endpoint no longer substitutes dotted/bracketed names; only `\w+` names are resolved" heads-up is prudent.

**ASCII-only `\w`.** Agree with documenting it and *not* widening here. That's a deliberate behaviour change with its own compatibility implications, not something to slip into a consolidation PR.

Smaller notes all sound correct. The `replaceAll` precision (the search side was the unquoted regex, not the replacement) is a fair correction — thanks.

No blocking concerns from me. This is ready for another maintainer's look.

PRAteek-singHWY:
Thanks @paoloantinori. Both of your action items are done.

**Block-subject asymmetry:** opened as #9059. It lays out the two patterns side by side,
records that it cannot produce a violation today because the validity rule has no
declared-but-unused check, and sketches the three ways it could go (teach the server pattern about
block subjects, drop the block-prefix group from the UI pattern, or document them as deliberately
different concepts). I did not pick one, since that depends on the intended playground UX more than
on the parsing.

**Release-notes line:** added to the `Prompt rendering service` section of
`assembly-llm-artifact-types-implementation.adoc`, which is where the placeholder syntax is already
documented. Worth mentioning that this repo has no `CHANGELOG` or release-notes file, so the docs
page seemed like the right home for it. I used "dotted or hyphenated" rather than "bracketed", since
brackets are not what changed, and gave `{{user.name}}` and `{{my-var}}` as concrete examples
instead of the raw `\w+`, which reads better on a user-facing page. Happy to reword if you had
somewhere else in mind.

On `lookup` and `log`: agreed, I will match whichever set #8974 lands on rather than guessing at a
different one here.

One piece of status for whoever picks this up next: the PR is still at `lifecycle/new`, so
`Unit Tests` has been skipped on every run so far and the full suite has not executed against this
branch yet. That includes the `app` `PromptRenderingServiceTest`, which is the one check I flagged
in the description as relying on CI. Everything I can run locally is green: 16 cases in
`PromptTemplateVariableUtilTest`, 24 in `PromptTemplateContentValidatorTest`, 11 in
`PromptTemplateConverterTest`, and checkstyle clean on `schema-util/common`,
`schema-util/util-provider`, `mcp` and `app`.

paoloantinori:
Thanks @PRAteek-singHWY — both action items look well-handled.

Block-subject asymmetry as #9059 is the right call: laying the two patterns side by side without forcing a design decision is better than guessing at the playground UX. The release-notes line in `assembly-llm-artifact-types-implementation.adoc` is fine — "dotted or hyphenated" with concrete `{{user.name}}` / `{{my-var}}` examples reads better on a user-facing page than the raw regex, good instinct.

Two things worth noting for whoever merges:
- Matching `lookup`/`log` to whatever #8974 lands on is the right sequencing — don't fork a different set here.
- The PR sitting at `lifecycle/new` and skipping `Unit Tests` on every run is the real blocker. It needs to move past `new` so CI actually exercises it before merge; flag that to a maintainer.

PRAteek-singHWY:
Hi @carlesarnal @EricWittmann, flagging this per @paoloantinori's note above.

The PR is still at `lifecycle/new`, so `Unit Tests` has been skipped on every run and the
full suite has never executed against this branch. Could one of you run `/accept` so CI
actually exercises it before review?

Thank you.

paoloantinori:
Thanks for the flag @PRAteek-singHWY. The code side is already reviewed and signed off from my end — the only outstanding item is the CI run, which needs a maintainer to `/accept` so the full suite exercises the branch. Pinging @EricWittmann on this; once CI is green this is good to merge.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -256,27 +253,11 @@ private String getTypeName(Object value) {
...
     private String substituteVariables(String template, Map<String, Object> variables) {
-        StringBuffer result = new StringBuffer();
-        Matcher matcher = VARIABLE_PATTERN.matcher(template);
-
-        while (matcher.find()) {
-            String varName = matcher.group(1).trim();
+        return PromptTemplateVariableUtil.substituteVariables(template, varName -> {
```
paoloantinori, line 354: Confirming direction: this routes `/render` through the shared word-character grammar, so dotted/hyphenated placeholders (`{{user.name}}`, `{{first-name}}`) stop being substituted and become literal text — a breaking change from the old `[^}]+` behavior, documented in the `.adoc` and pinned by `testDottedNameIsNotAVariable`. The author notes that widening the shared pattern (e.g. to allow dot/hyphen) is a one-line alternative that preserves backward compatibility. Worth an explicit maintainer decision before merge either way.

mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
```diff
@@ -434,14 +434,16 @@ public String renderTemplate(String template, Map<String, Object> args) {
...
-
-        // Simple {{variable}} substitution
-        for (Map.Entry<String, Object> entry : args.entrySet()) {
-            String placeholder = "\\{\\{" + entry.getKey() + "\\}\\}";
-            String value = entry.getValue() != null ? String.valueOf(entry.getValue()) : "";
-            rendered = rendered.replaceAll(placeholder, Matcher.quoteReplacement(value));
-        }
+        // Simple {{variable}} substitution, using the canonical pattern shared with the validity
+        // rule and the REST render endpoint so that {{name}} and {{ name }} resolve the same way.
+        String rendered = PromptTemplateVariableUtil.substituteVariables(template, varName -> {
```
paoloantinori, line 439: Heads-up on an undocumented sibling of the `/render` change: the old per-arg `replaceAll` substituted a literal `{{a.b}}` placeholder when an arg key `a.b` was supplied (the dot matched literally), but the shared word-character grammar now leaves it as literal text. The behavior-change list and docs call this narrowing out for `/render` but not for MCP — consider noting it alongside the `/render` change and confirming no deployed MCP prompt relies on dotted arg keys.

paoloantinori (commented):
Clean, well-tested consolidation: the three divergent `{{variable}}` parsers now route through one shared `PromptTemplateVariableUtil`, and I found no functional bugs across the regex, null-resolver, control-keyword, and conditional-block-ordering cases. The substantive items are two intentional, partly author-flagged behavior changes worth an explicit maintainer decision.

- **Headline decision — narrowed name grammar (word chars only) in `PromptTemplateVariableUtil.VARIABLE_PATTERN`, now used by `/render` (`PromptRenderingService.java:256`).** `/render` no longer substitutes dotted or hyphenated placeholders like `{{user.name}}` or `{{first-name}}` (it previously did, via `[^}]+`) — a user-facing breaking change, documented in the `.adoc` and pinned by `testDottedNameIsNotAVariable`. The author offers widening to dotted names as a one-line alternative; this is the direction call reviewers should confirm before merge.
- **The MCP converter narrows the same way but it is undocumented.** A literal `{{a.b}}` placeholder for a dotted arg key was substituted by the old per-arg `replaceAll` (the dot matched literally) and is now left as literal text. The behavior-change list and docs cover `/render` but not MCP, and the `.adoc` ("now uses the same grammar as ... the MCP prompt converter") slightly overstates MCP's prior consistency — its effective behavior came from the literal `{{key}}` match, not its dead `VARIABLE_PATTERN` field. Worth a docs line and a check that no MCP prompt uses dotted arg keys.
- **The `CONTROL_KEYWORDS` filter (`else`, `this`) is a justified companion to the unification, not scope creep.** Without it, the shipped `examples/a2a-real-world-integration/translator-agent-prompt.yaml` (uses `{{else}}`) would fail FULL validity under the new shared extractor. Covered by `testElseKeywordIsNotAnUndefinedVariable` and `testThisKeywordIsNotAnUndefinedVariable`.
- **The new `mcp` → `schema-util-common` dependency (`mcp/pom.xml`) is the debatable bit the author already flagged**; deferring the MCP half to a follow-up is a reasonable alternative if growing the MCP server classpath is unwanted. No correctness concern with it.

mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
PRAteek-singHWY: You are right, and the wording was mine to fix. Saying the render endpoint "now uses the same
grammar as the prompt template validity rule and the MCP prompt converter" implied the converter
already had that grammar. It did not. The old per-argument `replaceAll("\\{\\{" + key + "\\}\\}",
...)` built its search string from the argument name, so the dot matched itself and `{{a.b}}` was
substituted whenever an argument named `a.b` was supplied. The converter narrowed here just as the
render endpoint did, and the docs hid that.

Corrected in 10f40fac:

> Earlier versions of the render endpoint resolved such names. The MCP prompt converter also
> resolved a placeholder such as `{{user.name}}` when you supplied an argument of the same name.
> Both now use the same grammar as the prompt template validity rule.

On the audit you asked for, no shipped MCP prompt is affected. Every argument name is
word-character only: `analysisContext`, `conversation_history`, `customerMessage`,
`include_examples`, `question`, `responseText`, `sentimentContext`, `system_prompt` and
`targetLanguage`. One hit came back for `positive-amount`, but that is a data contract quality rule
in `examples/odcs-data-contracts/order-contract.yaml`, a file with no `templateId`, no `mcp` block
and no placeholders at all, so it is unrelated. I cannot speak for prompts users have already
deployed, which is the same caveat that applies to the render endpoint change.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
PRAteek-singHWY: Agreed that this is the call worth making explicitly, so here is the tradeoff as plainly as I can
put it.

Landing as it stands means `/render` and the MCP converter both stop resolving dotted and
hyphenated names, which is the alignment this PR is for. It is pinned by
`testDottedNameIsNotAVariable` and now documented for both call sites.

Widening `VARIABLE_PATTERN` to admit `.` and `-` is still a one-line change, but it is not a pure
win, and the part that is easy to miss is what it does to the validity rule rather than to the
renderer. Because all three consumers share the pattern, widening it would make the rule start
treating `{{user.name}}` as a variable that must be declared in `variables`. A stored template that
uses a dotted placeholder and does not declare it would begin failing `FULL` validity, so backward
compatibility for the renderer would come at the cost of a new way for existing artifacts to fail
validation.

That is why I went with narrowing, but I have no attachment to it. Happy to switch if you would
rather keep the renderer compatible, and either way it is a small change.

PRAteek-singHWY:
Thanks for the second pass @paoloantinori. Both inline points are answered in their threads, and
the docs correction for the MCP narrowing is pushed as 10f40fac.

On `lookup` and `log`, worth noting that #8974 is still open, so there is no landed `BLOCK_KEYWORDS`
set to match yet. `CONTROL_KEYWORDS` stays at `else` and `this` here and I will follow whatever that
PR lands with rather than forking a different set.

One note on the red check, in case it is misleading at a glance. `Verification Gate` is failing
because of the lifecycle state rather than anything in the branch. The gate tolerates skipped jobs
and no job failed; it exits non-zero because `lifecycle-ready` is false while the PR sits at
`lifecycle/new`. It should clear on its own once the PR is accepted and the suite actually runs.

paoloantinori:
Thanks @PRAteek-singHWY — that all tracks. The two inline threads are answered, the docs correction for the MCP narrowing in 10f40fac reads clearly, and keeping `CONTROL_KEYWORDS` at `else` / `this` until #8974 lands is the right sequencing. On the red check, the clarification is useful: the only failing item is the `Verification Gate` step that depends on `lifecycle/new` transition, not a code regression. From the review side this is signed off — the remaining blocker is the maintainer `/accept` that moves the PR past `lifecycle/new` so the full `Unit Tests` suite runs against the branch before merge.

PRAteek-singHWY:
@EricWittmann @carlesarnal, a gentle follow-up on this one.

@paoloantinori signed off from the review side on 2 August and noted the only thing left is the
maintainer `/accept` that moves the PR past `lifecycle/new`, so the full `Unit Tests` suite finally
runs against the branch. Nothing has changed since then, so CI still has not exercised the code and
the red `Verification Gate` is purely the lifecycle state rather than a regression.

There are no outstanding review comments. Both inline threads are answered, and the docs correction
for the MCP narrowing landed in 10f40fa.

@EricWittmann, this is the prompt template follow-up to #8841 that you merged and mentioned looking
forward to, so tagging you as well as Carles who is assigned here.

No urgency intended and I know the backlog is real. I am flagging it mainly because #9059, the
follow-up @paoloantinori asked me to open, is queued behind this one: the one-open-PR rule means I
cannot start on it until this merges.

Thanks both.

paoloantinori:
Still signed off from the review side — nothing has changed there. The only outstanding item is the maintainer `/accept` that moves the PR past `lifecycle/new` so the full CI suite runs, which is outside my lane.

Leaving the merge decision to @EricWittmann / @carlesarnal.

EricWittmann:
Thanks for the contribution, @PRAteek-singHWY!

PRAteek-singHWY:
Thank you @EricWittmann, and thanks @paoloantinori for the review that shaped this one.

Two follow-ups have come out of it since it merged, both already on the record.

**#9465.** @Faisal77666 reported the dead `VARIABLE_PATTERN` constant in `PromptTemplateConverter`.
That is the same constant this PR removed, so I checked it against current main and left the detail
there. Their diagnosis was right on both counts, they were just working from a clone pulled a day
before this merged, so that one looks closeable.

**#9059.** Rather than leave the three options open, I have proposed a direction, and your comment on
#9391 confirms the premise it rests on. Once the render endpoint evaluates `{{#if}}` blocks, the
subject of a block stops being decoration and becomes a value the rendered output depends on, which
settles the question. My recommendation is to teach the shared extractor to recognise block subjects,
sequenced after the conditional support lands rather than before it, and with a release-notes line
since it is a breaking change for any stored template that uses a block subject it never declared.

On that, #9391 was auto-closed last night by the one-PR-at-a-time bot rather than by review, because
@Prathamesh09-coder already had #9360 open. I have left a note there, since two of mine (#8921 and
#8923) were closed the same way and GitHub then refused to reopen them, returning a 422 because both
shared a head and base pair. A fresh PR from the same branch was the only way through, and it seemed
worth passing on rather than letting someone else lose the same two days.

Happy to pick up #9059 as soon as the direction is confirmed.

PR #8979, opened 2026-07-29
feat: Implement configuration-driven ArtifactTypeUtilProviders

### LINKED ISSUE #7318: Implement configuration-driven ArtifactTypeUtilProviders
## Parent Epic

## Phase
**Phase 4: Long-Term Architecture**

## Problem

13+ ArtifactTypeUtilProvider classes each implement 7-8 identical factory methods with only constructors differing. This is the largest source of boilerplate in the codebase.

**Affected Files (in schema-util/util-provider/):**
- `AvroArtifactTypeUtilProvider.java`
- `JsonArtifactTypeUtilProvider.java`
- `ProtobufArtifactTypeUtilProvider.java`
- `OpenApiArtifactTypeUtilProvider.java`
- `AsyncApiArtifactTypeUtilProvider.java`
- `XmlArtifactTypeUtilProvider.java`
- `XsdArtifactTypeUtilProvider.java`
- `WsdlArtifactTypeUtilProvider.java`
- `GraphQlArtifactTypeUtilProvider.java`
- `KConnectArtifactTypeUtilProvider.java`
- Plus 3 more...

**Current Pattern (repeated 13 times):**
```java
public class AvroArtifactTypeUtilProvider extends AbstractArtifactTypeUtilProvider {
    @Override public String getArtifactType() { return ArtifactType.AVRO; }
    @Override protected ContentCanonicalizer createContentCanonicalizer() { return new AvroContentCanonicalizer(); }
    @Override protected ContentValidator createContentValidator() { return new AvroContentValidator(); }
    @Override protected ContentExtractor createContentExtractor() { return new AvroContentExtractor(); }
    @Override protected ReferenceFinder createReferenceFinder() { return new AvroReferenceFinder(); }
    @Override protected ContentDereferencer createContentDereferencer() { return new AvroContentDereferencer(); }
    @Override protected CompatibilityChecker createCompatibilityChecker() { return new AvroCompatibilityChecker(); }
    @Override protected ContentAccepter createContentAccepter() { return new AvroContentAccepter(); }
}
```

## Proposed Solution

Replace with configuration-driven registry:

```java
public class ArtifactTypeProviderRegistry {
    
    private static final Map<String, ProviderConfig> PROVIDERS = Map.ofEntries(
        entry(ArtifactType.AVRO, ProviderConfig.builder()
            .canonicalizer(AvroContentCanonicalizer::new)
            .validator(AvroContentValidator::new)
            .extractor(AvroContentExtractor::new)
            .referenceFinder(AvroReferenceFinder::new)
            .dereferencer(AvroContentDereferencer::new)
            .compatibilityChecker(AvroCompatibilityChecker::new)
            .accepter(AvroContentAccepter::new)
            .build()),
        entry(ArtifactType.PROTOBUF, ProviderConfig.builder()
            // ... Protobuf configuration
            .build()),
        // ... other types
    );
    
    public ArtifactTypeUtilProvider getProvider(String artifactType) {
        ProviderConfig config = PROVIDERS.get(artifactType);
        if (config == null) {
            throw new UnknownArtifactTypeException(artifactType);
        }
        return new ConfigurableArtifactTypeUtilProvider(artifactType, config);
    }
}

@Builder
public class ProviderConfig {
    Supplier<ContentCanonicalizer> canonicalizer;
    Supplier<ContentValidator> validator;
    Supplier<ContentExtractor> extractor;
    Supplier<ReferenceFinder> referenceFinder;
    Supplier<ContentDereferencer> dereferencer;
    Supplier<CompatibilityChecker> compatibilityChecker;
    Supplier<ContentAccepter> accepter;
}
```

## Estimated Impact
- **Lines saved:** 700+ lines
- **Files removed:** 13 provider classes
- **Risk:** Medium (requires careful migration)

## Acceptance Criteria
- [ ] Create `ProviderConfig` configuration class
- [ ] Create `ArtifactTypeProviderRegistry` with configuration map
- [ ] Create `ConfigurableArtifactTypeUtilProvider` generic implementation
- [ ] Migrate all artifact types to configuration-based approach
- [ ] Maintain ServiceLoader compatibility for custom types
- [ ] Remove old provider classes
- [ ] All tests pass

aaravmahajanofficial:
Refactors the ArtifactTypeUtilProvider architecture by replacing 19 individual per-type subclass boilerplate files with a configuration-driven registry.

## Changes
- `ProviderConfig`: Created immutable configuration bean holding component suppliers and metadata per artifact type.
- `ConfigurableArtifactTypeUtilProvider`: Created single reusable ArtifactTypeUtilProvider delegating component creation to ProviderConfig.
- `StandardArtifactTypeProviderRegistry`: Created central, insertion-ordered registry mapping all 19 built-in artifact types to their respective configurations in exact deterministic order.
- `DefaultArtifactTypeUtilProviderImpl`: Updated standard provider initialization to load from StandardArtifactTypeProviderRegistry.

## Cleanup
Deleted 19 redundant **ArtifactTypeUtilProvider.java** boilerplate classes.
Tests: Added `StandardArtifactTypeProviderRegistryTest ` covering provider count, ordering, component resolution, and factory integration.

## Verification
- Verified `StandardArtifactTypeProviderRegistryTest` unit tests pass cleanly (19/19 providers verified).
- Verified application-level artifact type and canonicalizer tests pass without regressions.

Fixes: #7318


aaravmahajanofficial:
Hi @carlesarnal, @paoloantinori, @EricWittmann, Ready for review whenever you have a moment.

paoloantinori (commented):
Thanks for the tag. I did a full diff of the 19 deleted providers against the new `StandardArtifactTypeProviderRegistry`/`ProviderConfig` wiring — the refactor preserves behavior for all built-in types, and the test suite (`StandardArtifactTypeProviderRegistryTest`) covers count, ordering, component resolution, and a few representative component instances. Net this is a clean boilerplate reduction (−1732/+965) with no behavior regressions I could find.

A few observations, mostly nits:

- **`DefaultArtifactTypeUtilProviderImpl` now constructs a fresh set of provider instances per factory** (`standardProviders` is an instance field initialized inline), rather than the old static block that shared one set. This is actually *better*: `AbstractArtifactTypeUtilProvider` caches its lazily-created components in mutable `volatile` fields (e.g. `getContentAccepter()` writes `this.contentAccepter` on first call), so sharing a single provider instance across two factories would have been a latent aliasing bug. Creating them per-factory sidesteps that. Worth a one-line comment noting the intent, since a future reader may be tempted to "optimize" it back to a shared static list.

- **`OpenRpcArtifactTypeUtilProvider` previously returned `NoopCompatibilityChecker.INSTANCE`** explicitly; the new `OPENRPC` registry entry omits `.compatibilityChecker(...)`, so it falls back to the `ProviderConfig.Builder` default of `() -> NoopCompatibilityChecker.INSTANCE`. Behaviorally identical, but the omission is inconsistent with sibling entries (e.g. `ODCS_CONTRACT` explicitly sets `.compatibilityChecker(NoopCompatibilityChecker::new)`). For readability and to avoid future confusion about whether the omission is intentional, consider either explicitly setting it on `OPENRPC` or dropping it from `ODCS_CONTRACT` — whichever convention you prefer.

- **`ProviderConfig.Builder` silently accepts a config with no `contentTypes` set** (defaults to `Set.of()`), which would produce a provider that rejects all content types. Given this registry is the single source of truth for 19 providers, an assertion in `build()` that `contentTypes` is non-empty would catch a wiring typo at construction time rather than at runtime artifact ingestion. Low priority since the registry is fully populated today.

- **The `ProviderConfig` suppliers construct a new component instance on each `createX()` call** (e.g. `AvroContentAccepter::new`). Since `AbstractArtifactTypeUtilProvider` memoizes the first-created instance, each supplier is invoked at most once per provider — so there's no repeated-construction cost. Good as-is; just flagging that the lazy-cache in the base class is what makes the per-call `new` acceptable.

Overall: solid refactor, the new structure is easier to extend and audit than the 19 near-identical files it replaces.

schema-util/util-provider/src/main/java/io/apicurio/registry/types/provider/DefaultArtifactTypeUtilProviderImpl.java
```diff
@@ -11,19 +11,8 @@ public class DefaultArtifactTypeUtilProviderImpl implements ArtifactTypeUtilProv
     protected Map<String, ArtifactTypeUtilProvider> providerMap = new ConcurrentHashMap<>();
 
     protected List<ArtifactTypeUtilProvider> standardProviders = new ArrayList<ArtifactTypeUtilProvider>(
```
paoloantinori: Worth a brief comment here: `standardProviders` is an instance field (constructed per-factory) rather than a shared static list. This is intentional and correct — `AbstractArtifactTypeUtilProvider` caches its lazily-created components in mutable `volatile` fields, so sharing one provider instance across factories would alias that cache. A one-liner noting "intentionally per-instance to avoid cache aliasing" would prevent a future reader from "optimizing" this back to a static singleton.

schema-util/util-provider/src/main/java/io/apicurio/registry/types/provider/ProviderConfig.java
```diff
@@ -0,0 +1,305 @@
...
+            this.structuredContentExtractor = extractor;
+            return this;
+        }
+
+        /**
+         * Builds the immutable provider configuration.
+         *
+         * @return a new provider configuration
+         */
+        public ProviderConfig build() {
```
paoloantinori, line 302: Consider asserting in `build()` that `contentTypes` is non-empty. Since `StandardArtifactTypeProviderRegistry` is now the single wiring point for all 19 providers, a typo (forgetting `.contentTypes(...)` on a new entry) would otherwise silently produce a provider that rejects every content type, surfacing only at artifact-ingestion time rather than at startup. `Objects.requireNonNull` on the suppliers would similarly catch null-wiring early.

paoloantinori:
Thanks for the ping. Took a look — refactor is clean, behavior-preserving across all 19 types, and the new structure is a real improvement over the boilerplate. Left a formal review with a couple of minor observations (mostly readability nits around the per-instance provider construction and a missing defensive assertion in `ProviderConfig.build()`).

schema-util/util-provider/src/main/java/io/apicurio/registry/types/provider/DefaultArtifactTypeUtilProviderImpl.java
aaravmahajanofficial: Fixed 👍

schema-util/util-provider/src/main/java/io/apicurio/registry/types/provider/ProviderConfig.java
aaravmahajanofficial: Added 👍

aaravmahajanofficial:
Thanks @paoloantinori, for the thorough review and feedback. All comments have been resolved.

paoloantinori:
Glad it was useful — thanks for working through the feedback. I have nothing further to add from my side; the refactor reads cleanly. Deferring final merge decision to the maintainers.

EricWittmann (commented):
Thanks for the clean refactor — the architecture is sound, behavior is preserved across all 19 types, and the centralized registry makes adding new artifact types much easier. The test coverage is solid too.

One suggestion for future work: consider using Lombok `@Builder` + `@Getter` on `ProviderConfig`. The project already uses Lombok in DTO/model classes, and it would reduce `ProviderConfig.java` from ~320 lines to ~40 — which is more in the spirit of eliminating boilerplate (the goal of this PR). The hand-rolled builder and getters with per-method javadoc are functional but add a lot of ceremony for what is essentially an immutable data carrier.

Not blocking — just something to keep in mind if you're touching this area again.

EricWittmann:
Thanks for this excellent refactor, @aaravmahajanofficial! Replacing 19 boilerplate provider classes with a clean configuration-driven registry is a great improvement. Appreciate the thorough test coverage too. Merged!

carlesarnal:
Reviewed the full diff post-merge. The refactor is clean and behavior-preserving across all 19 artifact types. A few observations for the record:

**Behavioral equivalence**: Verified each provider's component wiring against the deleted per-type subclasses. All content types, accepters, validators, canonicalizers, dereferencers, reference finders, compatibility checkers, extractors, and `supportsReferencesWithContext` flags match. Key cases like ICEBERG_TABLE/VIEW constructor args (`true`/`false`), AVRO's custom `AvroReferenceArtifactIdentifierExtractor`, and types relying on builder defaults for noop components (KCONNECT, GRAPHQL, ASYNCAPI, OPENRPC) are all correct.

**Architecture**: The `ProviderConfig` → `ConfigurableArtifactTypeUtilProvider` → `StandardArtifactTypeProviderRegistry` layering is well structured. The `LinkedHashMap` preserves deterministic ordering, the builder defaults are sensible noops, and the per-factory instantiation comment (addressing @paoloantinori's feedback) correctly documents the aliasing-prevention intent.

**Defensive validation**: The `build()` method's `contentTypes.isEmpty()` check and `Objects.requireNonNull` guards on all suppliers will catch wiring mistakes at construction time — good addition.

**Test coverage**: Both `ProviderConfigTest` and `StandardArtifactTypeProviderRegistryTest` provide solid coverage of validation rules, ordering, component resolution, and factory integration.

**Minor nits** (non-blocking, can be addressed in a follow-up):
- `ConfigurableArtifactTypeUtilProvider.java` and `ProviderConfig.java` are missing a trailing newline at end of file.

Nice work, @aaravmahajanofficial!

Issue #8026, opened 2026-05-18
CLI: Handle RuleViolationProblemDetails exception


vandanayadav7:
## Description

`ProblemDetails` exceptions are handled, but `RuleViolationProblemDetails` is not, causing unhandled exception stack traces.

## Steps to Reproduce

```bash
acr group add test-group
acr artifact add -g test-group -f user-v1.avsc test-artifact
acr artifact version add -g test-group -a test-artifact -f user-v2.avsc --version 1
```

Results in:
```
ERROR: Unexpected error: io.apicurio.registry.rest.client.models.RuleViolationProblemDetails
```

## Expected Behavior

- Catch `RuleViolationProblemDetails` alongside `ProblemDetails`
- Show a user-friendly error message with rule violation details
- Test specifically when creating a version that fails an artifact rule

## Note

The reproduction above is a database unique constraint violation and may be a server bug. Verify with actual rule violations.

## Parent Epic

- #7162

PR #8987, opened 2026-07-29
feat(mcp): add RulesMCPServer and schema compatibility tests

### LINKED ISSUE #8986: Title: Add RulesMCPServer to expose schema compatibility testing via MCP

Currently, the Apicurio Registry MCP module provides tools to list and create artifacts (via `ArtifactsMCPServer` and `VersionsMCPServer`), but there is nothing that exposes the Rules Engine to an LLM. 

This means that if a developer provides an updated schema to an LLM like Claude, the LLM cannot verify if the updated schema breaks backward compatibility against the existing artifact in the registry before attempting to commit it.

### Proposed Solution

Add a new `RulesMCPServer` that provides a `test_schema_compatibility` tool to the MCP ecosystem. 

The tool will:
1. Accept a `groupId`, `artifactId`, `versionContent`, and `versionContentType`.
2. Simulate a version creation by using the existing `createVersion` API flow but appending the `dryRun = true` parameter.
3. Catch any `RuleViolationProblemDetails` exceptions and unpack the `Causes` and `Contexts` to return a highly detailed, human-readable explanation of exactly which fields or lines broke backward compatibility.

How it benefits users:
This allows developers to paste a schema into Claude and have Claude proactively test it against the registry API. Claude can then accurately explain exactly why the schema breaks backward compatibility before a developer even attempts to commit or deploy the changes.

### Additional Context

* Discussed and approved by @EricWittmann on the Apicurio community channel.
* A Pull Request with the complete implementation and integration tests is ready to be submitted.

dolliecoder:
## Description

This PR introduces the `RulesMCPServer` to the MCP ecosystem to expose the Apicurio Registry Rules Engine to LLMs like Claude. 

Currently, developers can list and create artifacts via the MCP server, but they cannot verify if a new schema iteration breaks backward compatibility prior to committing it. This PR adds the `test_schema_compatibility` tool, allowing an LLM to proactively test a schema iteration against the Registry API and provide the developer with an exact, detailed explanation of any compatibility violations.

### Key Changes:
- **`RegistryService.java`**: Added a `testSchemaCompatibility` method. This method simulates version creation utilizing the `dryRun = true` query parameter. It specifically catches `RuleViolationProblemDetails` to extract and format the underlying `Causes` and `Contexts`, returning a highly detailed error string explaining exactly what fields/lines broke the rules.
- **`RulesMCPServer.java`**: Created the new MCP server and registered the `test_schema_compatibility` tool, complete with descriptive `@Tool` annotations to optimally guide LLM usage.
- **`McpAuthenticationTest.java`**: Added the `testOAuth2AuthenticationSchemaCompatibility` integration test utilizing `TestContainers` to verify that the compatibility checking correctly interacts with an authenticated registry instance.

## Related Issue
Resolves #8986 

## Testing Done
- [x] Compiled successfully without errors (`mvn clean compile`).
- [x] Verified `checkstyle` passes with 0 violations.
- [x] Added `testOAuth2AuthenticationSchemaCompatibility` to `McpAuthenticationTest.java` to test the new functionality against a live `TestContainers` registry.


mcp/src/test/java/io/apicurio/registry/mcp/McpAuthenticationTest.java
```diff
@@ -74,4 +74,27 @@ public void testOAuth2AuthenticationCreateAndListGroup() {
...
+        registryService.createArtifact(testGroupId, testArtifactId, "JSON", "Test Artifact", "Description", null);
+        
+        // Create first version to set initial schema
+        registryService.createVersion(testGroupId, testArtifactId, "1.0.0", "application/json", initialSchema, "Test Version", "Initial version", null, false);
+        
+        // Test compatibility
+        String compatibleSchema = "{\"$schema\": \"http://json-schema.org/draft-07/schema#\", \"type\": \"object\", \"properties\": {\"id\": {\"type\": \"string\"}, \"name\": {\"type\": \"string\"}}}";
+        String result = registryService.testSchemaCompatibility(testGroupId, testArtifactId, compatibleSchema, "application/json");
+        
+        assertNotNull(result, "Result should not be null");
```
paoloantinori: This only asserts the result is non-null for a *compatible* schema, so it never touches the `RuleViolationProblemDetails` catch branch in `testSchemaCompatibility` — i.e. the cause/context formatting that is the core value of this tool. A second sub-case with a backward-incompatible change (e.g. `id` as `integer` instead of `string` under a `BACKWARD` compatibility rule), asserting the returned string describes the violation, would cover the failure path and confirm `dryRun=true` actually surfaces `RuleViolationProblemDetails`.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
```diff
@@ -370,6 +372,41 @@ public VersionMetaData createVersion(
...
+        c.setContentType(versionContentType);
+        c.setContent(versionContent);
+        v.setContent(c);
+
+        try {
+            client.groups().byGroupId(groupId).artifacts().byArtifactId(artifactId).versions().post(v, r -> {
+                r.queryParameters.dryRun = true;
+            });
+            return "Schema is compatible.";
+        } catch (RuleViolationProblemDetails e) {
```
paoloantinori, line 394: This catch block is the centerpiece of the feature, but it's not reached by the current test (which only feeds a compatible schema). A negative integration test would both validate this branch and confirm that, with `dryRun=true`, the registry surfaces a rule violation as a thrown `RuleViolationProblemDetails` rather than a 200.

paoloantinori (commented):
Nice feature — exposing the rules engine to LLMs via `dryRun` so a compatibility break can be surfaced *before* committing a version is a genuinely useful addition to the MCP surface, and the `@Tool` description is well-tuned for LLM consumption. The overall structure matches the existing `*MCPServer` classes, so it slots in cleanly.

My main concern is test coverage: the most valuable logic in this PR — the `RuleViolationProblemDetails` catch block that extracts `getCauses()` / `getContext()` into a human-readable string — is never exercised by the test. Details below.

- **The failure path is untested.** `testOAuth2AuthenticationSchemaCompatibility` only feeds in a *compatible* schema (adding an optional `name` field) and asserts `result != null`. That exercises the `return "Schema is compatible."` line, but never the `catch (RuleViolationProblemDetails e)` branch — which is the whole point of the tool. I'd add a second case with an incompatible change (e.g. flip `id` from `string` to `integer`, or remove a required field under a `BACKWARD` rule) and assert the returned string mentions the violation / contains the cause description. Without that, the cause-formatting logic and even the assumption that `dryRun=true` surfaces `RuleViolationProblemDetails` rather than returning 200 are both unverified.

- **`assertNotNull(result)` is too weak even for the happy path** — it would pass for any non-null garbage. `assertTrue(result.contains("compatible"))` (or `.equals("Schema is compatible.")`) would actually lock in the contract.

- **Worth double-checking the catch ordering assumption.** The code relies on `RuleViolationProblemDetails` being thrown *directly* (and thus caught before the generic `ApiException`). That matches what issue #8026 reported for the same client (raw `RuleViolationProblemDetails` surfaced as an unhandled exception), and the same pattern is used in `AbstractRegistryMojo`, so it's almost certainly fine — but since the current test never triggers a violation, this is currently an unverified assumption rather than a tested fact.

Minor: there's trailing whitespace on a couple of lines in the new test method (e.g. after `currentTimeMillis();`). Not a blocker.

Overall the implementation looks correct to me; strengthening the test is the one change I'd ask for before merge.

dolliecoder:
@paoloantinori @jsenko  PTAL

paoloantinori:
Seen — my earlier review on this head (9f978252) still stands: the main points I'd want addressed before merge are the visibility widening of `RegistryService.client` to package-private and the test's reliance on per-run group/artifact IDs without explicit cleanup (which can leave orphaned artifacts behind on the registry). The happy/sad-path test shape is good. Happy to re-review once those are addressed or if a maintainer decides they're acceptable as-is.

dolliecoder:
@paoloantinori  I've reverted RegistryService.client back to private and wrapped the test logic in a try-finally block to explicitly delete the test group/artifacts and prevent any orphans. Ready for another look when you have a moment

paoloantinori:
Thanks @dolliecoder — both points addressed cleanly. `RegistryService.client` stays `private`, and since `testSchemaCompatibility`, `deleteGroup`, and `createArtifactRule` are all methods on `RegistryService` itself, the encapsulation is preserved. The try-finally cleanup in both `testOAuth2AuthenticationCreateAndListGroup` and the new `testOAuth2AuthenticationSchemaCompatibility` deletes the test group (and transitively its artifacts) even on assertion failure, so no orphans leak into subsequent runs. The happy/backward-compat-break paths in the new test exercise the `dryRun` flow and the `RuleViolationProblemDetails` cause extraction well. Good to go from my side.

jsenko (commented):
Review on head `c866d70`. Good addition — exposing the dry-run compatibility check via MCP is a useful tool for AI agents. The `dryRun=true` approach is correct, and both happy/failure paths are now tested (addressed earlier Copilot feedback). No correctness bugs found.

**1. Import ordering (should-fix — checkstyle)**

`RegistryService.java:22`: `import com.microsoft.kiota.ApiException` is placed between `io.apicurio` imports (between lines 18 and 23). This breaks import grouping and will likely fail checkstyle. Move it before the `io.apicurio` block.

**2. Fully-qualified names in `createArtifactRule` (should-fix)**

`RegistryService.java` ~line 520-522:
```java
var createRule = new io.apicurio.registry.rest.client.models.CreateRule();
createRule.setRuleType(io.apicurio.registry.rest.client.models.RuleType.forValue(ruleType));
```

Both `CreateRule` and `RuleType` should be imported, not inline FQN. Project convention: "Use imports instead of fully qualified names."

**3. FQN in test assertions (should-fix)**

`McpAuthenticationTest.java`: uses `org.junit.jupiter.api.Assertions.assertTrue(...)` three times without importing. Add `import org.junit.jupiter.api.Assertions` (or static import `assertTrue`) to match the existing `assertNotNull` usage in the same file.

**4. Missing Apache 2.0 license header (should-fix)**

`RulesMCPServer.java` is a new file but has no license header. All new Java files require the Apache 2.0 header block.

jsenko (commented):
## Review

The previous review (on head `c866d70`) raised 4 should-fix items — import ordering, fully-qualified names in `RegistryService` and the test, and a missing license header on `RulesMCPServer.java`. Those still need to be addressed.

One additional item:

### should-fix: Tool name/description doesn't reflect actual behavior

`RulesMCPServer.java` — the tool is named `test_schema_compatibility` and the description emphasizes compatibility, but the underlying `dryRun=true` call runs through `rulesService.applyRules()`, which executes *all* configured rules — validity, integrity, and compatibility. A schema could fail this check because of a validity rule with no compatibility rule configured at all.

Update the tool name (e.g. `test_schema_rules` or `validate_schema_version`) and/or the `@Tool` description to make clear this tests against all configured rules, not only compatibility.

- **Separate tool vs. dryRun on create_version:** Having a dedicated tool for this is the right call. LLMs select tools by description, and a focused "test rules" tool will be chosen more reliably than a dryRun flag on create_version. It also avoids accidental writes.

- **Shared service method with `createVersion`:** The overlap is only 3 lines of request object setup, and the methods diverge immediately (return type, error handling, parameters). Not worth extracting.

### Suggestion: Javadoc on RulesMCPServer

The class name `RulesMCPServer` implies broader scope than the single tool it currently contains. Consider adding a class-level Javadoc noting that this server is intended to eventually cover all rules-related operations (list, create, delete rules), and the current tool is the first step. This sets expectations for future contributors.

dolliecoder:
@jsenko Thanks for the review! I've pushed a commit that resolves all the should-fix items (imports, FQNs, missing license header, and updating the tool name/description to test_schema_rules to accurately reflect its behavior).

Let me know if anything else is needed

jsenko (commented):
Fixes [re-review of commit 02672ae]

### Context
Re-review after jsenko's manual feedback: rename tool from `test_schema_compatibility` to `test_schema_rules`, add class-level Javadoc, fix FQN usage, fix imports.

### Re-review delta

**Resolved findings (from previous reviews):**
- Import ordering — fixed (`ApiException` in correct alphabetical position)
- FQN usage in `RegistryService.createArtifactRule` — fixed (`CreateRule`, `RuleType` now imported)
- FQN usage in test — fixed (static import `assertTrue` instead of `Assertions.assertTrue`)
- Tool name doesn't reflect all-rules behavior — fixed, renamed to `test_schema_rules` with updated description mentioning validity, integrity, and compatibility
- Javadoc on `RulesMCPServer` scope — added

**New finding:**

### should-fix: Remove per-file Apache license header from RulesMCPServer.java

The previous review incorrectly flagged the missing license header as a should-fix. The project convention is to **not** add per-file Apache license headers — the repository-root `LICENSE` governs. None of the other MCP server files (`AdminMCPServer`, `ArtifactsMCPServer`, `GroupsMCPServer`, `SystemMCPServer`, `VersionsMCPServer`, `PromptTemplateMCPServer`) have one. Please remove lines 1–15 from `RulesMCPServer.java`.

No further issues found.

carlesarnal:
@dolliecoder this has now a conflict, can you fix it?

jsenko (approved):
Approving. Thanks for your patience on this one, @dolliecoder — it sat longer than it should have.

Since I'd already reviewed the code and the only things left were mechanical, I pushed the remaining fixes straight to your branch rather than send it back for another round. Everything I changed is below, so please review it — it's your PR and your name is on it, and if you disagree with any of it, say so and we'll sort it out.

## What I pushed

**1. Merged `main` and resolved the conflict.** This was not the trivial import collision it looked like. `main` refactored `RegistryService` to replace the `client` field with a `client()` accessor backed by the new `RegistryClientResolver`. Git auto-merged the method bodies without a conflict marker, but all three of your new methods still referenced the deleted field, so the branch did not compile:

```
RegistryService.java:[293] cannot find symbol: variable client   // testSchemaRules
RegistryService.java:[543] cannot find symbol: variable client   // deleteGroup
RegistryService.java:[550] cannot find symbol: variable client   // createArtifactRule
```

All three now call `client()`. Worth flagging for future rebases: the visible conflict was only in the import block, so resolving that alone would have produced a broken branch.

**2. Fixed the compatibility test assertions — this is the one that matters.**

The happy-path assertion was:

```java
assertTrue(result.contains("compatible"), "Result should indicate schema is compatible");
```

`"Incompatible".contains("compatible")` is `true`, so that assertion passes on a *failure* message. And it was in fact doing exactly that. Here is the real output your "compatible" schema produced:

```
Schema rules check failed: Incompatible artifact: ... num of incompatible diffs: {1},
list of diff types: [Object type property schemas narrowed at /propertySchemasAdded]
```

Adding the `name` property is not backward-compatible under Apicurio's JSON Schema diff — it reports it as a *narrowing* at `/propertySchemasAdded`. So the test was green while both of its branches exercised the violation path, and the success path of `testSchemaRules` was never covered at all.

The happy path now re-submits the schema unchanged, which produces no differences and therefore genuinely satisfies the BACKWARD rule, and asserts the exact string:

```java
assertEquals("Schema is valid and compatible.", result);
```

The negative assertion had the same problem in a quieter form — `negativeResult.contains("id")` matches any message containing the word "valid", so it never actually verified the `id` field was named. It now asserts the specific reported difference, `Subschema type changed at /properties/id`.

None of this is a criticism of the implementation — `testSchemaRules` itself works correctly against a real Registry. It's that the assertions couldn't tell you so.

**3. Removed `@Tag("integration")` from `McpAuthenticationTest`.** This is why the above went unnoticed. `mcp/pom.xml` sets `<excludedGroups>integration</excludedGroups>` and no CI workflow overrides it, so this class had never run anywhere — locally or in CI. It now runs in the `non-app` shard and takes ~15s.

**4. Enabled group deletion on the test Registry.** `apicurio.rest.deletion.group.enabled` is `false` by default, so the cleanup in your `finally` block was returning 405 and only logging a warning:

```
action="deleteGroupById" result="failure" error_msg="Group deletion operation is not enabled."
```

Added `APICURIO_REST_DELETION_GROUP_ENABLED=true` to `AuthTestContainersManager` so it actually cleans up.

**5. Dropped the Apache license header from `RulesMCPServer.java`.** Apologies — that was added on incorrect advice from an earlier review round. The project does not use per-file headers (the repository-root `LICENSE` governs), and no other file in `mcp/servers/` has one.

## Verification

On the merged branch: 83 mcp tests pass with 0 failures (the single skip is the pre-existing `@Disabled` `DebugTest`), `McpAuthenticationTest` 4/4, and `./mvnw checkstyle:check -pl :apicurio-registry-mcp-server` reports 0 violations.

## One thing for you

**The PR description is out of date** and it would be good to fix it before this merges, since the description is what ends up in the history. It describes a `test_schema_compatibility` tool and a `testSchemaCompatibility` method, but after the rename in review these are:

- tool: **`test_schema_rules`** (in `RulesMCPServer`)
- method: **`RegistryService.testSchemaRules(...)`**
- test: `testOAuth2AuthenticationSchemaCompatibility` in `McpAuthenticationTest`

The `## Testing Done` section is also no longer accurate — it lists `mvn clean compile` and checkstyle, but the test now runs in CI as described above.

If you'd rather not bother, no problem: this comment records the correct names for anyone reading back through the history.

Thanks again for the contribution, and for sticking with it through the review cycles.

PR #8989, opened 2026-07-29
feat(cli): support system-wide installation via `acr install --global`

### LINKED ISSUE #8649: CLI: Global installation (system-wide)
## Parent Epic

## Description
The CLI currently installs per-user. Add a system-wide installation option that places the `acr` binary in a global location (e.g. `/usr/local/bin`) accessible to all users.

## Scope
- Add `--global` flag to the install command
- Install to `/usr/local/bin` (or platform equivalent) when `--global` is used
- Requires `sudo` / elevated privileges — detect and prompt if not running as root
- Update auto-update mechanism to handle global installs
- Document the global install option

## Acceptance Criteria
- `acr install --global` installs system-wide
- Clear error if run without sufficient privileges
- Auto-update works for global installs
- Per-user install remains the default

Me-Priyank:
Fixes #8649

## What this does

Adds a `--global` flag to `acr install` so the CLI can be installed once for everyone on a machine, instead of only for the current user.

Per-user install stays exactly as it was and remains the default. Nothing changes unless you pass `--global`.

## How it works

- `acr install --global` copies the CLI into a shared location (`/usr/local/lib/apicurio-registry-cli`) and symlinks the `acr` launcher into
  `/usr/local/bin`, which is already on every user's PATH. Because that directory is shared, no per-user shell config (`.bashrc` / `.zshrc`) is touched.
- Before writing anything, it checks whether the target directories are writable. If not, it stops with a clear message asking the user to re-run with `sudo`, instead of failing later with a raw permission error.
- The install records a small marker in `config.json`. When `acr update` runs, it reads that marker and re-installs with `--global`, so auto-update keeps working for global installs.

The global paths can be overridden with `ACR_GLOBAL_HOME` and `ACR_GLOBAL_BIN` (mainly so the behavior is testable without writing to real system folders).

## Acceptance criteria

- [x] `acr install --global` installs system-wide
- [x] Clear error if run without sufficient privileges
- [x] Auto-update works for global installs
- [x] Per-user install remains the default

## Testing

- Added `InstallCommandTest` cases: files land in the global home and the `acr` symlink lands in the global bin, the config marker is written, the user's home and shell config are left untouched, and a per-user install does not set the marker.
- These tests are `@EnabledOnOs({MAC, LINUX})` (they rely on symlinks and Unix paths), matching the existing tests in this file. They run in CI on Linux.

## Notes

- `acr update` now uses `ACR_CURRENT_HOME` (always set by the `acr` launcher) instead of `ACR_HOME` (only exported for per-user installs). Without this, updating a global install would fail with "ACR_HOME is not set". This is identical to `ACR_HOME` for per-user installs, so that path is unchanged.
- Out of scope: a global install's `config.json` lives in the shared, root-owned home, so a non-root user cannot write CLI context/login settings there. Per-user config for global installs is a separate concern and not part of this issue. Happy to follow up if you'd like it addressed.


Me-Priyank:
@EricWittmann @vandanayadav7 @carlesarnal PTAL

cli/src/main/java/io/apicurio/registry/cli/UpdateCommand.java
```diff
@@ -115,7 +115,9 @@ private void handleCheck(OutputBuffer output, CliVersion currentVersion) {
     }
```
paoloantinori, line 120: This switch (and the matching one in `handleAutoUpdate`) changes the update target for *all* installs, not just global ones. If `getAcrHomePath()` and `getAcrCurrentHomePath()` ever differ for an existing per-user install, the update would land in a different directory than before. The comment says the launcher always sets the current-home path, which is plausible — just worth confirming on a pre-existing per-user install since it touches the existing update flow, not only the new global one.

paoloantinori (commented):
Nice feature — the `--global` flag, the privilege pre-check with an actionable message, and the `ACR_GLOBAL_HOME`/`ACR_GLOBAL_BIN` overrides for testability are all well thought through, and the per-user path is correctly left as the default. The tests for file placement, the config marker, and the "doesn't touch ~/bin or shell config" invariant are solid.

I have two concerns worth addressing, the stronger one about the auto-update path:

- **`isGlobalInstall()` may read the marker from the wrong location.** The install side deliberately switched from `getAcrHomePath()` to `getAcrCurrentHomePath()` in both `handlePathInstall` and `handleAutoUpdate`, with the comment that "a global install has no sourced env" — i.e. `ACR_HOME` is only exported into a shell for per-user installs. But `isGlobalInstall()` reads the marker via `config.read()`. If `config.read()` resolves the config through the same `ACR_HOME`-based lookup that was just bypassed for the home path, it will look in the wrong place (or fail), hit the `catch (RuntimeException)` → `return false`, and the re-install will run *without* `--global`. That would silently break the "Auto-update works for global installs" acceptance criterion. Worth confirming that `config.read()` reads from the current binary's home (`getAcrCurrentHomePath()`), not the env-resolved one — and ideally adding a test that simulates a global install's update and asserts `--global` is forwarded.

- **No test for the privilege-failure path.** "Clear error if run without sufficient privileges" is an explicit acceptance criterion, but all four global tests point the overrides at writable temp dirs, so `ensureGlobalInstallPrivileges` never actually fires the `CliException`. A test that points `ACR_GLOBAL_BIN` at a read-only directory and asserts the `VALIDATION_ERROR_RETURN_CODE` + the `sudo` hint message would lock that behavior in.

A couple of smaller notes:

- `isWritableLocation` walks up to the nearest *existing* ancestor and checks `Files.isWritable` on it. That's a reasonable approximation, but it can diverge from the real outcome (e.g. the target exists and is root-owned, or the nearest existing ancestor is broadly world-writable like `/usr/local` on some macOS setups). Acceptable as a pre-check since the real write still fails loudly if the approximation is wrong, but worth a one-line comment noting it's a heuristic, not a guarantee.
- The `UpdateCommand` change from `getAcrHomePath()` to `getAcrCurrentHomePath()` applies to *all* updates, not just global ones. If the two ever resolved differently for an existing per-user install, the update target directory would silently change. The comment argues the current binary's home is always set by the launcher, which sounds right — just flagging it as a behavior change touching the existing path, so worth a manual sanity check on a pre-existing per-user install.

Overall the implementation is clean; the marker-location consistency is the one I'd want nailed down before merge.

cli/src/main/java/io/apicurio/registry/cli/UpdateCommand.java
```diff
@@ -195,4 +201,18 @@ private void runInstallFromZip(Path zipFilePath, Path targetDir, OutputBuffer ou
         output.writeStdOutLine("Update complete.");
     }
 
+    /**
+     * Detects whether the current installation was performed globally, so the re-install triggered
+     * by an update keeps the same system-wide scope. Best-effort: any failure to read the config
+     * (for example a per-user install with no marker) is treated as a per-user installation.
+     */
+    private boolean isGlobalInstall() {
```
paoloantinori: This is my main concern. The install side moved away from `ACR_HOME`-based lookup to `getAcrCurrentHomePath()` precisely because a global install has no sourced `ACR_HOME`. But `isGlobalInstall()` reads the marker via `config.read()`. If `config.read()` resolves config through the same `ACR_HOME`-based path that was just bypassed, it won't find the marker the install wrote (to the global cli-home's `config.json`), the `catch` returns `false`, and the re-install runs without `--global` — silently reverting a global install to per-user on update. Worth verifying `config.read()` reads from the current binary's home.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
```diff
@@ -71,6 +101,80 @@ public void run(final OutputBuffer output) throws IOException {
...
+
+    private Path getGlobalBinPath() {
+        final String override = config.getEnv(ENV_ACR_GLOBAL_BIN);
+        final String bin = isBlank(override) ? DEFAULT_GLOBAL_BIN_DIR : override;
+        return Path.of(bin).normalize().toAbsolutePath();
+    }
+
+    /**
+     * Verifies the current user can write to both global install locations, failing with a clear,
+     * actionable message rather than a low-level permission error when elevated privileges are needed.
```
paoloantinori, line 173: "Clear error if run without sufficient privileges" is an explicit acceptance criterion, but every global test points `ACR_GLOBAL_BIN`/`ACR_GLOBAL_HOME` at writable temp dirs, so `ensureGlobalInstallPrivileges` never fires. A test pointing `ACR_GLOBAL_BIN` at a read-only dir and asserting `VALIDATION_ERROR_RETURN_CODE` plus the `sudo` hint would cover the criterion that's currently untested.

Me-Priyank:
Thanks for the thorough review @paoloantinori , this was really helpful.

On the main concern (marker location): you're right to flag it, and I confirmed the mechanism holds. `isGlobalInstall()` reads via `config.read()`, which resolves `config.json` under `getConfigFilePath()` -> `getAcrCurrentHomePath()`, the running binary's own home, not the `ACR_HOME`-based path. So it reads the marker the global install wrote to the global cli-home, even with no sourced env. I've locked this in with a new `UpdateCommandTest` that asserts an update of a global install re-installs with `--global` and a per-user install does not, and I extracted `buildInstallCommand()` so the forwarding is directly tested.

On the privilege path: agreed it was untested. Added a case that points `ACR_GLOBAL_BIN` at a read-only directory and asserts `VALIDATION_ERROR_RETURN_CODE`, the sudo hint, and that nothing gets installed (skipped when running as root, where the check cannot fire).

On the two smaller notes: added a comment marking `isWritableLocation` as a best-effort heuristic rather than a guarantee, and a note that the update home-path switch resolves to the same directory as `ACR_HOME` for per-user installs, so the existing update path is unchanged.

All green locally (checkstyle clean, the update tests pass; the install tests are OS-gated to Linux/macOS as before). Pushed as a follow-up commit.

paoloantinori:
Thanks @Me-Priyank — your explanation of the marker resolution is exactly what I wanted to confirm, and it holds up against the code: `markInstallationAsGlobal` writes the flag into the global home's `config.json`, and during `acr update` `getAcrCurrentHomePath()` points at that same global home (the running binary's location), so `config.read()` finds it without relying on `ACR_HOME`. The `UpdateCommandTest` cases lock the round-trip in nicely.

Two small things worth a sanity check when you get a chance, neither blocking:

- The switch from `getAcrHomePath()` to `getAcrCurrentHomePath()` in `handlePathInstall`/`handleAutoUpdate` is described as equivalent for per-user installs. That's true as long as the per-user launcher keeps `ACR_HOME` and the binary's home pointing at the same directory — worth confirming there's no per-user scenario where they diverge (e.g. an install moved/symlinked after the fact).
- `markInstallationAsGlobal` rewrites `config.json` via `Mapper.writeValue` without preserving the existing formatting/key order. If users have hand-edited config or rely on stable diffs, a tree-model merge (`readTree` + `put` + write) would be less surprising. Low priority.

Overall this looks good to me. Deferring merge to a maintainer.

Me-Priyank:
> Thanks @Me-Priyank — your explanation of the marker resolution is exactly what I wanted to confirm, and it holds up against the code: `markInstallationAsGlobal` writes the flag into the global home's `config.json`, and during `acr update` `getAcrCurrentHomePath()` points at that same global home (the running binary's location), so `config.read()` finds it without relying on `ACR_HOME`. The `UpdateCommandTest` cases lock the round-trip in nicely.
> 
> Two small things worth a sanity check when you get a chance, neither blocking:
> 
> * The switch from `getAcrHomePath()` to `getAcrCurrentHomePath()` in `handlePathInstall`/`handleAutoUpdate` is described as equivalent for per-user installs. That's true as long as the per-user launcher keeps `ACR_HOME` and the binary's home pointing at the same directory — worth confirming there's no per-user scenario where they diverge (e.g. an install moved/symlinked after the fact).
> * `markInstallationAsGlobal` rewrites `config.json` via `Mapper.writeValue` without preserving the existing formatting/key order. If users have hand-edited config or rely on stable diffs, a tree-model merge (`readTree` + `put` + write) would be less surprising. Low priority.
> 
> Overall this looks good to me. Deferring merge to a maintainer.

Thanks @paoloantinori.

On `getAcrCurrentHomePath()` vs `getAcrHomePath()`: the launcher computes `ACR_CURRENT_HOME` in `find_acr_home()` as the first of `(script_dir, ACR_HOME)` that actually contains `acr_runner`, where `script_dir` resolves the symlink chain to the real binary location. So even in the "moved or symlinked after the fact" case it tracks where the binary really is, and it falls back to `ACR_HOME` when `script_dir` has no runner. It is therefore never a worse locator than `ACR_HOME`, and for a normal per-user install both point at the same cli-home. The value is only used as the scratch parent for the download, which is user-writable in every case, so there is no per-user scenario where the switch changes the outcome.

On `markInstallationAsGlobal` rewriting config.json: I kept it consistent with how the CLI already writes config elsewhere. `Config.write`, and the installation-version bump in `InstallCommand.copyFiles`, both round-trip `ConfigModel` through `Mapper`, so formatting is already normalized on any config write and the `config` map values are preserved. Since this file is tool-managed (not committed, edited via `acr` commands) the stable-diff concern is limited, but I am happy to switch this and the sibling writes to a `readTree` + `put` + write merge if you would prefer that as the convention.

Thanks again for the thorough review.

paoloantinori:
Hi @Me-Priyank — thanks for the detailed write-back, this all holds together.

`ACR_CURRENT_HOME`: good — the equivalence holds as long as the per-user launcher computes `ACR_CURRENT_HOME` from the binary's own location, so a per-user update still resolves to the same dir as `ACR_HOME`. The only residual risk is an install that's been moved or symlinked after the fact, which you've called out. Acceptable as documented behavior.

`config.json` rewrite: agreed, tree-model merge (`readTree` + `put` + `writeValue`) would preserve user edits and stable diffs better than a full rewrite. Low priority, but worth a follow-up issue so it's not lost.

Marker round-trip + the read-only `ACR_GLOBAL_BIN` privilege test both look solid. From my side this is fine to merge — leaving the final call to a maintainer.

Me-Priyank:
Thanks @paoloantinori , I'll open a follow-up issue for the `config.json` tree-model merge (readTree + put + writeValue) across the config write sites so it's tracked and not lost.

Me-Priyank:
@EricWittmann @carlesarnal @vandanayadav7  could you PTAL ? :))

paoloantinori:
Thanks — agreed on opening the follow-up issue for the `config.json` tree-model merge so it is tracked. Looping in @EricWittmann and @carlesarnal as you asked; from my side the review on this PR is closed, no outstanding concerns.

vandanayadav7 (changes requested):
Ran e2e testing in a headless Linux container. The global install flow works correctly — files in the right place, symlinks created, config marker set, privilege check gives a clear message, per-user install unchanged, and auto-update correctly forwards `--global`. All 17 unit tests pass.

One issue: after a global install, a non-root user running any config-writing command gets misleading output:

```
$ acr context create prod http://registry.example.com
Current context 'prod' added.
Error: Could not write config file '/usr/local/lib/apicurio-registry-cli/config.json'. Caused by: Permission denied
```

The user sees a success message followed by a crash. Before this PR, `config.json` was always in the user's home directory and always writable, so this never happened. The global install introduces the first scenario where `config.json` is root-owned and non-root users can't write it.

@Me-Priyank please add an early check — if the config file is not writable, fail with a clear message before printing any success output. Something like "Cannot write to config file — this is a global installation. Per-user configuration for global installs is not yet supported." That way the user gets one clear error instead of a success message followed by a crash.

Me-Priyank:
> Ran e2e testing in a headless Linux container. The global install flow works correctly — files in the right place, symlinks created, config marker set, privilege check gives a clear message, per-user install unchanged, and auto-update correctly forwards `--global`. All 17 unit tests pass.
> 
> One issue: after a global install, a non-root user running any config-writing command gets misleading output:
> 
> ```
> $ acr context create prod http://registry.example.com
> Current context 'prod' added.
> Error: Could not write config file '/usr/local/lib/apicurio-registry-cli/config.json'. Caused by: Permission denied
> ```
> 
> The user sees a success message followed by a crash. Before this PR, `config.json` was always in the user's home directory and always writable, so this never happened. The global install introduces the first scenario where `config.json` is root-owned and non-root users can't write it.
> 
> @Me-Priyank please add an early check — if the config file is not writable, fail with a clear message before printing any success output. Something like "Cannot write to config file — this is a global installation. Per-user configuration for global installs is not yet supported." That way the user gets one clear error instead of a success message followed by a crash.

thank you for review @vandanayadav7. Fixed in the latest commit.

The root cause was two things: `context create` printed the success line and then called `config.write()` in the same output chunk, so on a write failure the buffered success text leaked out right before the error. And more generally, every config-writing command only discovered the read-only config after doing its work.

Changes:
- Added an early writability check in `Config.write()` (the single point all commands write through), so a non-writable config now fails before any success output, with a clear message and the validation exit code:
  `Cannot write to the CLI config file '...'. This CLI was installed globally, and per-user configuration for global installs is not yet supported.`
  A non-global permission issue gets a generic "check that you have permission to write it" message instead.
- Reordered `context create` to persist before reporting success, matching the other context commands, which removes the success-then-crash entirely.
- Centralized `isGlobalInstall()` in `Config` and reused it from `UpdateCommand` (it had a near-duplicate copy).
- Added `ConfigWritableTest` covering the global message, the generic message, and the writable-config happy path. The permission-based tests are POSIX-only and skip as root, following the existing install privilege test.

Let me know if you would prefer the guard as an explicit per-command call rather than inside `Config.write()`.

vandanayadav7 (dismissed):
Fix verified — non-root user now gets a clean error before any success message:

"Cannot write to the CLI config file. This CLI was installed globally, and per-user configuration for global installs is not yet supported."

No misleading output, no crash. Full CLI test suite passes (111 tests, 0 failures). LGTM.

vandanayadav7:
@Me-Priyank The PR has a merge conflict with main. Could you rebase and resolve it? Thanks!

Me-Priyank:
Resolved the conflicts and rebased @vandanayadav7 PTAL

vandanayadav7 (approved):
Re-approving after rebase. Code unchanged from previous review. LGTM.

vandanayadav7:
@Me-Priyank The branch can't be auto-rebased. Could you rebase locally onto main and push? Thanks!

Me-Priyank:
Done @vandanayadav7... Rebased locally onto the latest main and force-pushed. Should be good for auto-merge now. Thanks :))

vandanayadav7 (approved):

EricWittmann:
Merged!  :tada:

PR #9009, opened 2026-07-29
fix(cli): encrypt OAuth client secret in file-based credential fallback

### LINKED ISSUE #8380: CLI Auth: Encrypt OAuth client secret in stored config
## Parent Epic
Part of #7162 (Phase 4 - Authentication)

## Description
The OAuth client secret stored in CLI config/context is currently in plain text. It should be encrypted at rest to improve security.

## Acceptance Criteria
- OAuth client secret is encrypted when stored in CLI config
- Existing plain text secrets are migrated on next use
- Decryption is transparent to the user during CLI operations

Priyanshubhartistm:
## Summary

When the OS keychain isn't available, the CLI falls back to storing login credentials (OAuth2 client secret, Basic auth password) in a plain text `credentials.json` file. This encrypts those values at rest.

Fixes #8380.

### Encryption at rest (`CredentialEncryption`, new)
- Encrypts/decrypts secrets with AES-256-GCM (random 96-bit IV per value, 128-bit auth tag).
- The key is generated on first use and stored in a separate `credentials.key` file with owner-only permissions, alongside `credentials.json`.

### Transparent encrypt/decrypt (`FileCredentialProvider`)
- `store()` encrypts the secret before writing it to `credentials.json`.
- `retrieve()` decrypts an already-encrypted value and returns it unchanged to the caller  (`Client.configureAuth`), so every existing CLI command that authenticates works exactly
  as before.

### Migration of existing plain text secrets (`FileCredentialProvider.retrieve`)
- If a stored value isn't in the encrypted format (i.e. it was written before this change),  it's treated as legacy plain text: the original value is still returned, and the entry is  encrypted and rewritten to `credentials.json` before the call returns. No user action or extra command is needed — migration happens the next time the credential is used.

## Testing

- `FileCredentialProviderTest` (new): secret is never stored as plain text on disk, round-trip encrypt/decrypt returns the original value, the key file gets owner-only permissions, a pre-existing plain text entry is migrated on `retrieve()` and stays readable afterwards  (including across a fresh provider instance), a missing account returns `null`, and `delete()`
  removes the secret.
- `./mvnw test -pl cli`
- `./mvnw checkstyle:check -pl cli`
- `./scripts/validate-files.sh`
- `./mvnw clean package -Pprod -Dfull -DskipTests -DcliSkipNative`

## Notes

- The OAuth client secret and Basic auth password were never stored in `config.json` itself — only in the OS keychain or, as a fallback, `credentials.json`. The keychain path was already encrypted at the OS level; this change closes the gap in the file-based fallback, which is  the path the CLI explicitly warns about as "not recommended for production use".
- Native-image (GraalVM) build/runtime behavior for the new `javax.crypto` usage was not  independently verified in this environment — worth a manual check with
  `./mvnw verify -pl cli -Dquarkus.native.container-build=false` before merging if that hasn't run in CI yet.

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
@@ -1,0 +1,109 @@
+package io.apicurio.registry.cli.auth;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.PosixFilePermission;
+import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
+import java.util.Base64;
+import java.util.EnumSet;
+import javax.crypto.Cipher;
+import javax.crypto.KeyGenerator;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+/**
+ * Encrypts and decrypts secrets stored by {@link FileCredentialProvider} using an
+ * AES-256-GCM key kept in a separate, permission-restricted file next to credentials.json.
+ * This keeps a leaked or backed-up credentials.json from exposing secrets on its own.
+ */
+class CredentialEncryption {
+
+    static final String ENCRYPTED_PREFIX = "enc:v1:";
+
+    private static final String KEY_ALGORITHM = "AES";
+    private static final String CIPHER_TRANSFORMATION = "AES/GCM/NoPadding";
+    private static final int KEY_SIZE_BITS = 256;
+    private static final int GCM_IV_LENGTH_BYTES = 12;
+    private static final int GCM_TAG_LENGTH_BITS = 128;
+
+    private final Path keyPath;
+    private SecretKey key;
+
+    CredentialEncryption(final Path keyPath) {
+        this.keyPath = keyPath;
+    }
+
+    static boolean isEncrypted(final String value) {
+        return value != null && value.startsWith(ENCRYPTED_PREFIX);
+    }
+
+    String encrypt(final String plaintext) {
+        try {
+            final var iv = new byte[GCM_IV_LENGTH_BYTES];
+            new SecureRandom().nextBytes(iv);
+            final var cipher = Cipher.getInstance(CIPHER_TRANSFORMATION);
+            cipher.init(Cipher.ENCRYPT_MODE, loadOrCreateKey(), new GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv));
+            final var ciphertext = cipher.doFinal(plaintext.getBytes(StandardCharsets.UTF_8));
+            return ENCRYPTED_PREFIX
+                    + Base64.getEncoder().encodeToString(iv)
+                    + ":"
+                    + Base64.getEncoder().encodeToString(ciphertext);
+        } catch (GeneralSecurityException | IOException ex) {
+            throw new CredentialStoreException("Failed to encrypt credential: " + ex.getMessage(), ex);
+        }
+    }
+
+    String decrypt(final String encoded) {
+        try {
+            final var parts = encoded.substring(ENCRYPTED_PREFIX.length()).split(":", 2);
+            final var iv = Base64.getDecoder().decode(parts[0]);
+            final var ciphertext = Base64.getDecoder().decode(parts[1]);
+            final var cipher = Cipher.getInstance(CIPHER_TRANSFORMATION);
+            cipher.init(Cipher.DECRYPT_MODE, loadOrCreateKey(), new GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv));
+            return new String(cipher.doFinal(ciphertext), StandardCharsets.UTF_8);
+        } catch (GeneralSecurityException | IOException ex) {
+            throw new CredentialStoreException("Failed to decrypt credential: " + ex.getMessage(), ex);
+        }
+    }
+
+    private synchronized SecretKey loadOrCreateKey() throws IOException, GeneralSecurityException {
+        if (key != null) {
+            return key;
+        }
+        if (!Files.exists(keyPath)) {
+            generateKeyFile();
+        }
+        key = new SecretKeySpec(Files.readAllBytes(keyPath), KEY_ALGORITHM);
+        return key;
+    }
+
+    private void generateKeyFile() throws IOException, GeneralSecurityException {
+        final KeyGenerator generator = KeyGenerator.getInstance(KEY_ALGORITHM);
+        generator.init(KEY_SIZE_BITS);
+        final var temp = Files.createTempFile(keyPath.getParent(), "credentials-key", ".tmp");
+        try {
+            Files.write(temp, generator.generateKey().getEncoded());
+            restrictFilePermissions(temp);
+            Files.move(temp, keyPath, StandardCopyOption.ATOMIC_MOVE);
+        } catch (FileAlreadyExistsException ex) {
+            // Another process created the key file concurrently — use it instead.
+        } finally {
+            Files.deleteIfExists(temp);
+        }
+    }
+
+    private static void restrictFilePermissions(final Path path) throws IOException {
+        try {
+            Files.setPosixFilePermissions(path, EnumSet.of(
+                    PosixFilePermission.OWNER_READ, PosixFilePermission.OWNER_WRITE));
+        } catch (UnsupportedOperationException ex) {
+            // Non-POSIX filesystem — best effort only.
+        }
+    }
+}
--- a/cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
@@ -37,7 +37,12 @@ class CredentialEncryption {
     private static final int KEY_SIZE_BITS = 256;
     private static final int GCM_IV_LENGTH_BYTES = 12;
     private static final int GCM_TAG_LENGTH_BITS = 128;
-    private static final SecureRandom RANDOM = new SecureRandom();
+
+    // Deliberately an instance field, not a static final constant: a static SecureRandom is
+    // initialized during native-image class-init (build time), which GraalVM rejects — building
+    // it in the image heap would bake in a fixed/cached seed. Constructing it per-instance keeps
+    // initialization at runtime, after the image has started.
+    private final SecureRandom random = new SecureRandom();
 
     private final Path keyPath;
     private SecretKey key;
@@ -53,7 +58,7 @@ class CredentialEncryption {
     String encrypt(final String plaintext) {
         try {
             final var iv = new byte[GCM_IV_LENGTH_BYTES];
-            RANDOM.nextBytes(iv);
+            random.nextBytes(iv);
             final var cipher = Cipher.getInstance(CIPHER_TRANSFORMATION);
             cipher.init(Cipher.ENCRYPT_MODE, loadOrCreateKey(), new GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv));
             final var ciphertext = cipher.doFinal(plaintext.getBytes(StandardCharsets.UTF_8));
@@ -62,20 +67,29 @@ class CredentialEncryption {
                     + ":"
                     + Base64.getEncoder().encodeToString(ciphertext);
         } catch (GeneralSecurityException | IOException ex) {
-            throw new CredentialStoreException("Failed to encrypt credential: " + ex.getMessage(), ex);
+            throw new CredentialStoreException("Failed to encrypt credential", ex);
         }
     }
 
     String decrypt(final String encoded) {
+        final byte[] iv;
+        final byte[] ciphertext;
         try {
             final var parts = encoded.substring(ENCRYPTED_PREFIX.length()).split(":", 2);
-            final var iv = Base64.getDecoder().decode(parts[0]);
-            final var ciphertext = Base64.getDecoder().decode(parts[1]);
+            if (parts.length != 2) {
+                throw new CredentialStoreException("Corrupted encrypted credential — re-run login");
+            }
+            iv = Base64.getDecoder().decode(parts[0]);
+            ciphertext = Base64.getDecoder().decode(parts[1]);
+        } catch (IllegalArgumentException ex) {
+            throw new CredentialStoreException("Corrupted encrypted credential — re-run login", ex);
+        }
+        try {
             final var cipher = Cipher.getInstance(CIPHER_TRANSFORMATION);
             cipher.init(Cipher.DECRYPT_MODE, loadOrCreateKey(), new GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv));
             return new String(cipher.doFinal(ciphertext), StandardCharsets.UTF_8);
         } catch (GeneralSecurityException | IOException ex) {
-            throw new CredentialStoreException("Failed to decrypt credential: " + ex.getMessage(), ex);
+            throw new CredentialStoreException("Failed to decrypt credential", ex);
         }
     }
 
--- a/cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
@@ -21,6 +21,12 @@ import javax.crypto.spec.SecretKeySpec;
  * Encrypts and decrypts secrets stored by {@link FileCredentialProvider} using an
  * AES-256-GCM key kept in a separate, permission-restricted file next to credentials.json.
  * This keeps a leaked or backed-up credentials.json from exposing secrets on its own.
+ *
+ * <p>Threat model: since {@code credentials.key} is stored unencrypted in the same directory
+ * as {@code credentials.json}, this only defends against a single-file leak or backup of
+ * credentials.json — it does not protect against compromise of the whole config directory,
+ * which exposes both the key and the ciphertext. This is not a keychain-equivalent; it exists
+ * only for the fallback path used when no OS keychain is available.
  */
 class CredentialEncryption {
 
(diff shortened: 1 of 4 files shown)


cli/src/main/java/io/apicurio/registry/cli/auth/FileCredentialProvider.java
```diff
@@ -33,13 +35,24 @@ class FileCredentialProvider implements CredentialProvider {
...
     @Override
     public String retrieve(final String account) {
-        return readCredentials().get(account);
+        final var credentials = readCredentials();
+        final var stored = credentials.get(account);
+        if (stored == null) {
+            return null;
+        }
+        if (CredentialEncryption.isEncrypted(stored)) {
+            return encryption().decrypt(stored);
```
paoloantinori: This migration write inside `retrieve()` turns a previously read-only call into a potentially-failing write. If `credentials.json` is not writable at read time (read-only home, full disk, permission regression), `writeCredentials` will throw and a valid credential won't be returned. Consider whether the migration should be best-effort (swallow/log the write failure and still return `stored`) or whether a failed migration should intentionally block auth — and either way, a one-line comment stating the chosen intent would help future readers.

cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
```diff
@@ -0,0 +1,110 @@
...
+import java.security.SecureRandom;
+import java.util.Base64;
+import java.util.EnumSet;
+import javax.crypto.Cipher;
+import javax.crypto.KeyGenerator;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+/**
```
paoloantinori, line 22: Worth stating the threat-model boundary explicitly here: with the key stored in plain text in the same directory as `credentials.json`, this only defends against single-file leak/backup of `credentials.json`, not against full-home compromise. That's a reasonable trade-off given the no-keychain fallback context, but documenting it prevents someone later from assuming this is keychain-equivalent.

cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
```diff
@@ -0,0 +1,110 @@
...
+        final KeyGenerator generator = KeyGenerator.getInstance(KEY_ALGORITHM);
+        generator.init(KEY_SIZE_BITS);
+        final var temp = Files.createTempFile(keyPath.getParent(), "credentials-key", ".tmp");
+        try {
+            Files.write(temp, generator.generateKey().getEncoded());
+            restrictFilePermissions(temp);
+            Files.move(temp, keyPath, StandardCopyOption.ATOMIC_MOVE);
+        } catch (FileAlreadyExistsException ex) {
+            // Another process created the key file concurrently — use it instead.
+        } finally {
```
paoloantinori, line 137: Minor ordering note: the key bytes are written to the temp file *before* `restrictFilePermissions(temp)` runs, so there's a brief window where the temp file (random name, same dir) has default umask permissions. Low risk given the random temp name and immediate atomic rename, but setting permissions on the temp path before writing the key bytes would close the window entirely.

paoloantinori (commented):
Solid security improvement — AES-256-GCM with a per-value random IV, a 128-bit auth tag, and an atomic key-file creation are the right primitives, and the migration path for legacy plain-text entries is a thoughtful touch. A few points worth considering before this lands.

- **`retrieve()` now writes to disk as a side effect of the legacy migration.** Previously `retrieve()` was read-only; now, on the first read of a pre-existing plain-text entry, it rewrites `credentials.json`. If the config home happens to be read-only at read time (e.g. a locked-down environment, a full disk, or a permission regression), an otherwise-valid credential will fail to be returned because the migration write throws. Worth deciding whether the migration should be best-effort (catch/log the write failure and still return the plaintext) or whether a failed migration intentionally blocks authentication. The current behavior reads as intentional-but-implicit, and a short comment stating the intent would help.

- **Threat-model clarity.** Storing `credentials.key` in plain text next to `credentials.json` (same directory) only defends against the single-file leak/backup of `credentials.json` — it does not protect against full-home compromise, since an attacker with the directory gets both key and ciphertext. That's a reasonable trade-off given this code path exists precisely *because* no OS keychain is available, but it would help future readers to state that boundary explicitly (either in the class Javadoc on `CredentialEncryption` or in the PR description), so no one later assumes this is a keychain-equivalent.

- **Lazy `encryption()` init is not synchronized.** `FileCredentialProvider.encryption()` builds a new `CredentialEncryption` if the field is null; two concurrent callers could each build one. The end result is still correct because `loadOrCreateKey()` is synchronized and `Files.move(..., ATOMIC_MOVE)` provides the real mutual exclusion for key creation, so this is wasteful rather than buggy. If the CLI is effectively single-threaded for auth this is a non-issue — flagging only in case credential access is expected to be concurrent.

- Minor: `generateKeyFile()` writes the temp key, then calls `restrictFilePermissions(temp)`, then does the atomic move. There's a brief window where the temp file (random name, same dir) has default umask permissions. Low risk given the random temp name and the immediate atomic rename, but a stricter order (set permissions before writing the key bytes) would close it. Worth double-checking whether the current order is acceptable for the project's bar.

The test coverage is good — the plain-text-on-disk assertion, the transparent round-trip, the key-file permission check, and the legacy-migration scenario all exercise the meaningful invariants.

Priyanshubhartistm:
@paoloantinori @jsenko, could you please take a look and let me know if everything looks good or if any changes are needed? 
Thank you!

cli/src/main/java/io/apicurio/registry/cli/auth/FileCredentialProvider.java
Priyanshubhartistm: done

paoloantinori:
Hi @Priyanshubhartistm — thanks, I've seen the "done" update. Confirming the changes on this head address my earlier review points (the per-value IV, owner-only key-file permissions, and the transparent migration of legacy plain-text entries all hold up), and I've nothing further to add on the code. The `/accept` to move it forward for a full CI run is a maintainer action, so over to @jsenko / a maintainer there.

Priyanshubhartistm:
Hi @EricWittmann and @jsenko, could you please take a look and let me know if everything looks good or if any changes are needed? 
Thank you!

carlesarnal:
@Priyanshubhartistm please, rebase main on your work, thanks!

jsenko (commented):
## Code Review — security audit findings

The crypto implementation is sound: AES-256-GCM with random 96-bit IV per encrypt, 128-bit auth tag, properly generated 256-bit key. Versioned format (`enc:v1:`) is good for future algorithm changes. Threat model is well documented in the class javadoc. paoloantinori's feedback on migration best-effort behavior appears addressed (line 59 catches write failures).

Note on paoloantinori's temp file comment: in the current code, `restrictFilePermissions(temp)` (line 98) runs BEFORE `Files.write(temp, ...)` (line 99), so the key is written to an already-restricted file. The only window is between `createTempFile` and `restrictFilePermissions` when the file is empty — no key exposure.

### 1. Missing input validation in decrypt() — should-fix (`CredentialEncryption.java:71-73`)

```java
final var parts = encoded.substring(ENCRYPTED_PREFIX.length()).split(":", 2);
final var iv = Base64.getDecoder().decode(parts[0]);
final var ciphertext = Base64.getDecoder().decode(parts[1]);
```

If the stored value is malformed (e.g. truncated, corrupted filesystem), `parts[1]` throws `ArrayIndexOutOfBoundsException` — an unhelpful error. Add:

```java
if (parts.length != 2) {
    throw new CredentialStoreException("Corrupted encrypted credential — re-run login");
}
```

### 2. Exception messages leak implementation details — should-fix (`CredentialEncryption.java:65,78`)

`ex.getMessage()` in error messages can expose filesystem paths (from IOException) or algorithm internals (from GeneralSecurityException). Use generic messages:

```java
throw new CredentialStoreException("Failed to encrypt credential", ex);
throw new CredentialStoreException("Failed to decrypt credential", ex);
```

The cause is still chained for debug logging upstream.

### 3. Native image (GraalVM) compatibility not verified — should-fix

The CLI builds as a native binary. `javax.crypto.Cipher`, `KeyGenerator`, and `SecretKeySpec` may need explicit native-image configuration. The PR description acknowledges this was not verified. Before merging, either:
- Run `./mvnw package -Pnative` and test encrypt/decrypt in the native binary, or
- Add `--enable-all-security-services` to native build args if needed

### 4. No tests for error/tamper paths — should-fix (`FileCredentialProviderTest.java`)

Tests only cover the happy path. For a security feature, add at least:
- Tampered ciphertext is rejected (GCM tag verification)
- Malformed encrypted values (wrong prefix, missing colons, invalid Base64) produce clear errors, not stack traces

carlesarnal:
## Code Review

Good work on this PR. The crypto fundamentals are solid (AES-256-GCM with per-value random IV, 128-bit auth tag, versioned format prefix), and the migration path for legacy plain-text entries is well thought out. paoloantinori earlier feedback has been addressed cleanly in commit e4d1ebda.

jsenko posted a thorough review earlier today with 4 should-fix items. Those all look valid and are still open. I want to call out one additional point.

### Supplementary finding: Unhandled decrypt failure in the encrypted path

`FileCredentialProvider.java` line 45: When an already-encrypted value is corrupted or tampered with (e.g. truncated write, filesystem corruption), `encryption().decrypt(stored)` throws a `CredentialStoreException` that propagates uncaught through `CredentialStore.retrieve()` when `unsafeStorage` is true.

The legacy migration path (lines 51-58) has best-effort error handling. If the encrypt-and-rewrite fails, the plaintext is still returned. But the encrypted path at line 45 does not have equivalent protection.

Compare with `readCredentials()` in the existing file, which catches `IOException` on corrupted JSON and returns an empty map gracefully.

**Suggestion:** Wrap the decrypt call at line 45 in a try-catch that logs a warning and returns `null`, so the user gets a clear re-login prompt instead of a stack trace. Alternatively, if failing loudly is intentional, add a comment stating that intent.

### Summary of open items

1. **Missing input validation in `decrypt()`** (`CredentialEncryption.java` lines 71-73) - per jsenko review
2. **Exception messages leak implementation details** (`CredentialEncryption.java` lines 65, 78) - per jsenko review
3. **GraalVM native image compatibility** - per jsenko review
4. **Tests for error and tamper paths** - per jsenko review
5. **Unhandled decrypt failure in encrypted retrieve path** - supplementary finding above

### What looks good

- Crypto: AES-256-GCM, per-value random 96-bit IV, 128-bit auth tag
- Versioned encrypted format (`enc:v1:`) for future algorithm changes
- Threat-model boundary documented in class Javadoc
- Key file permissions set before writing key bytes
- Atomic key-file creation with concurrent-process safety
- Synchronized lazy init of `CredentialEncryption`
- Best-effort legacy migration with clear logging
- DCO sign-off, conventional commits, clean imports, consistent style
- SonarCloud passed with 0 issues
- Solid happy-path test coverage

Priyanshubhartistm:
@carlesarnal, @jsenko, could you please review this and let me know if any changes are needed? 
Thanks!

paoloantinori (commented):
All three reviewers reported back, and they converged in an important way that corrected my own preliminary analysis. Two agents independently established that `SecretKeySpec(byte[], String)` does **not** validate key length (it only rejects null/empty), so a truncated `credentials.key` fails later at `Cipher.init` as `InvalidKeyException` — a `GeneralSecurityException`, which the existing catches already handle gracefully. My initial instinct to flag an "uncaught `IllegalArgumentException`" was wrong; I dropped it. I also refuted the "writeCredentials throws an uncaught checked exception" concern by reasoning from the compile constraint: `store()` calls `writeCredentials()` with no try/catch and no `throws`, so it must wrap I/O errors in the unchecked `CredentialStoreException` the migration catch handles. (No JDK available here to run a snippet, but the API behavior is well-established and two independent agents + the compile-fact agree.)

Net: the crypto construction is sound and there are no major/blocking defects. What remains are minor, honest observations. Producing the review now.

cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
Priyanshubhartistm: done

Priyanshubhartistm:
Hi @jsenko @carlesarnal , thanks again for the reviews and approvals. I noticed the Verification Gate check is currently failing, but it looks like that's purely a lifecycle issue, not a code issue
please take a look

carlesarnal:
Re-review of commit d6d67bb: all 5 previously flagged issues are resolved. (1) Input validation in decrypt() added at CredentialEncryption.java:74-76. (2) Exception messages are now generic at lines 65,85. (3) SecureRandom moved to instance field for GraalVM at line 43. (4) Error/tamper-path tests added in CredentialEncryptionTest and FileCredentialProviderTest. (5) Unhandled decrypt failure in retrieve() now caught at FileCredentialProvider.java:46-53 returning null with a warning. Crypto fundamentals remain solid. Test coverage is comprehensive. Nothing further from me.

cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
Priyanshubhartistm: done

cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final String CIPHER_TRANSFORMATION = "AES/GCM/NoPadding";
+    private static final int KEY_SIZE_BITS = 256;
+    private static final int GCM_IV_LENGTH_BYTES = 12;
+    private static final int GCM_TAG_LENGTH_BITS = 128;
+
+    // Deliberately an instance field, not a static final constant: a static SecureRandom is
+    // initialized during native-image class-init (build time), which GraalVM rejects — building
+    // it in the image heap would bake in a fixed/cached seed. Constructing it per-instance keeps
+    // initialization at runtime, after the image has started.
+    private final SecureRandom random = new SecureRandom();
```
paoloantinori, line 50: This instance-field `SecureRandom` correctly avoids native-image build-time seeding, but the broader native-image risk is the new `javax.crypto` usage itself: `Cipher.getInstance("AES/GCM/NoPadding")` and `KeyGenerator` need the JCE provider registered for the native image at runtime, not just at build time. If that provider isn't already pulled in by existing TLS config, a runtime `NoSuchProviderException`/`NoSuchPaddingException` here would silently disable the entire file-based credential fallback. Since the PR notes call this out as unverified, a native smoke test (`./mvnw verify -pl cli -Dquarkus.native.container-build=false`) before merge would close the gap.

paoloantinori (commented):
Solid change — an independent review of the AES-256-GCM construction confirms the crypto is sound (per-value random 96-bit IV, 128-bit auth tag verified on decrypt, key file created via temp-file + atomic move with owner-only perms and correct cross-process race handling), and the migration logic, exception flow, and test coverage are thorough. The one real pre-merge action is confirming the native build/runtime, which the PR notes already flag.

- Crypto correctness holds: no IV-reuse or padding-oracle surface, and every failure path degrades to a generic `CredentialStoreException` without leaking plaintext (`CredentialEncryption.java:60-70`, `:87-93`).
- Native image: the new `javax.crypto` usage needs the JCE provider available at native runtime, not just build time — the SecureRandom init-timing concern is already handled (`CredentialEncryption.java:41-45`), but provider registration is a separate risk worth a native smoke test.
- Edge cases considered and judged acceptable: a legacy plaintext value coincidentally prefixed `enc:v1:` would be misclassified as encrypted and surface as a one-time re-login prompt, but it self-heals on the next `store()` (which always encrypts); the whole-file read-modify-write race is pre-existing and theoretical for a single-user CLI. Neither rises to blocker.

cli/src/main/java/io/apicurio/registry/cli/auth/CredentialEncryption.java
Priyanshubhartistm: @paoloantinori , thanks 
Ran ./mvnw verify -pl cli with the container-based Mandrel build  native compile succeeded, and manually verified both encrypt (login --allow-unsafe-credential-storage) and decrypt (artifact get) paths against the compiled native binary. No JCE/provider errors; credentials.json shows enc:v1: encrypted values with 0600 key-file perms.

Priyanshubhartistm:
Hii @vandanayadav7 , Could you please take a look.
If everything looks good now, could you please /accept this PR when you have a chance? I'd really appreciate it.

vandanayadav7 (changes requested):
Nice work — the encryption implementation is cryptographically sound (AES-256-GCM, random IV per encryption, proper auth tag), the threat model is clearly documented, and the migration path for legacy plaintext credentials is well thought out.

I ran full e2e testing in a headless Linux container (no OS keychain) against a Keycloak-backed registry and verified the core flow works: login stores encrypted credentials, authenticated CLI commands succeed, and the `enc:v1:` format is used on disk. However, the edge case testing uncovered a few issues.

#### Must fix

**1. Corrupted key file causes permanent unrecoverable state**

If `credentials.key` is truncated or corrupted (disk error, partial backup restore), the CLI becomes permanently broken. `loadOrCreateKey()` reads the corrupt file (it exists, so no regeneration), `SecretKeySpec` accepts any byte length, but `Cipher.init()` rejects non-16/24/32-byte keys. Every subsequent `login` and `retrieve` fails with `"Failed to encrypt credential"` and the only recovery is manually deleting `credentials.key`.

Suggested fix — validate key length and regenerate if invalid:
```java
byte[] keyBytes = Files.readAllBytes(keyPath);
if (keyBytes.length != KEY_SIZE_BITS / 8) {
    log.warn("Corrupted encryption key file — regenerating.");
    Files.delete(keyPath);
    generateKeyFile();
    keyBytes = Files.readAllBytes(keyPath);
}
key = new SecretKeySpec(keyBytes, KEY_ALGORITHM);
```

#### Should fix

**2. `credentials.key` not cleaned up on logout**

`delete()` removes `credentials.json` when the last credential is removed, but leaves `credentials.key` on disk indefinitely.

```java
if (credentials.isEmpty()) {
    Files.deleteIfExists(credentialsPath());
    Files.deleteIfExists(config.getAcrCurrentHomePath().resolve(KEY_FILE));
}
```

**3. No `AtomicMoveNotSupportedException` fallback in `generateKeyFile()`**

`FileCredentialProvider.moveFile()` has a fallback for non-atomic filesystems (NFS, cloud mounts), but `CredentialEncryption.generateKeyFile()` does not. Key generation will fail on those filesystems.

Priyanshubhartistm:
@vandanayadav7 , Thanks for the review! I've addressed all three points.
Let me know if there's anything else you'd like changed!

vandanayadav7 (approved):

vandanayadav7 (approved):
All 3 findings addressed with tests. LGTM.

PR #9012, opened 2026-07-29
fix(ui): render recommended models as read-only labels

### LINKED ISSUE #8873: fix(ui): Prevent 404 router errors when clicking recommended models in PromptTemplateViewer
## Description

The `recommendedModels` metadata in `PromptTemplateViewer.tsx` is currently rendered as clickable React Router links:

```tsx
<Link
    to={`/explore/default/${encodeURIComponent(model)}`}
    className="model-link"
>
    {model}
</Link>
```

However, `recommendedModels` contains **AI model identifiers** (e.g., `gpt-4o`, `claude-3-5-sonnet`) rather than artifact IDs in the `default` registry group.

As a result, clicking any recommended model attempts to navigate to a non-existent artifact route, leading to a **404 (Artifact Not Found)** page.

---

## Steps to Reproduce

1. Open a `PROMPT_TEMPLATE` artifact that contains `metadata.recommendedModels`.
2. Navigate to the **Documentation** tab.
3. Scroll to the **Recommended Models** section.
4. Click any model tag (e.g., `gpt-4o`).

---

## Expected Behavior

Recommended models should be displayed as **read-only informational labels**. Since they are metadata rather than artifact references, clicking them should not trigger navigation.

---

## Actual Behavior

- Each recommended model is rendered as a clickable link.
- Clicking a model navigates to `/explore/default/<model>`.
- Because no corresponding artifact exists, the user is taken to a **404 (Artifact Not Found)** page.

---

## Proposed Fix

Remove the `<Link>` wrapper in `PromptTemplateViewer.tsx` and render recommended model tags as static PatternFly labels (e.g., `<Label color="purple">`).

This accurately reflects their purpose as informational metadata and prevents navigation to invalid routes.

DWilson32:
metadata.recommendedModels holds AI model identifiers (e.g. gpt-4o, claude-3-5-sonnet), not artifact IDs in the default group. Linking each one to /explore/default/<model> therefore navigated to a non-existent artifact and rendered a 404 page.

Render the models as plain PatternFly labels instead, since they are informational metadata rather than artifact references. Also drops the now-unused Link import and the dead .model-link styles.

Fixes #8873

## What
`metadata.recommendedModels` on a PROMPT_TEMPLATE artifact was rendered as React
Router links pointing at `/explore/default/<model>`.
## Why
That field holds AI model identifiers (`gpt-4o`, `claude-3-5-sonnet`), not artifact
IDs in the `default` group — so every click resolved to a non-existent artifact and
landed the user on a 404 page.
Recommended models are informational metadata rather than artifact references, so
they are now plain PatternFly labels with no navigation. The unused `Link` import
and the dead `.model-link` styles are removed along with it.
## Verification
- `npx tsc --noEmit` — no new errors in the changed file
- `npx eslint` on the changed file — clean
- No behavioural change outside the Recommended Models section

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.css
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.css
@@ -94,12 +94,3 @@
     background-color: var(--pf-t--global--background--color--secondary--default);
     border-radius: 3px;
 }
-
-.prompt-template-viewer .model-link {
-    color: inherit;
-    text-decoration: none;
-}
-
-.prompt-template-viewer .model-link:hover {
-    text-decoration: underline;
-}
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
@@ -216,8 +216,8 @@ export const PromptTemplateViewer: FunctionComponent<PromptTemplateViewerProps> 
                         <Divider className="section-divider" />
                         <Title headingLevel="h3" size="md">Recommended Models</Title>
                         <LabelGroup className="section-content">
-                            {meta.recommendedModels.map((model, index) => (
-                                <Label key={index} color="purple" isCompact>
+                            {meta.recommendedModels.map((model) => (
+                                <Label key={model} color="purple" isCompact>
                                     {model}
                                 </Label>
                             ))}
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
@@ -1,5 +1,4 @@
 import React, { FunctionComponent } from "react";
-import { Link } from "react-router";
 import "./PromptTemplateViewer.css";
 import {
     Card,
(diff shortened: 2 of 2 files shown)


Labreo:
Tested locally (`npm run build` in `ui/`). The removal of router navigation on AI model strings fixes the 404 behavior cleanly.

One quick question regarding `key={index}` on line 219: since `meta.recommendedModels` is an array of unique string identifiers (e.g. `gpt-4o`, `claude-3-5-sonnet`), should we use `key={model}` instead of the array index to follow React key practices during re-renders, similar to how tags or model references are keyed elsewhere in the viewer?

DWilson32:
> Tested locally (`npm run build` in `ui/`). The removal of router navigation on AI model strings fixes the 404 behavior cleanly.
> 
> One quick question regarding `key={index}` on line 219: since `meta.recommendedModels` is an array of unique string identifiers (e.g. `gpt-4o`, `claude-3-5-sonnet`), should we use `key={model}` instead of the array index to follow React key practices during re-renders, similar to how tags or model references are keyed elsewhere in the viewer?

Good catch — and thanks for actually running it locally, appreciated.

Pushed in 0198cf7: keyed by `model` and dropped the now-unused `index` parameter.
The values are unique strings, so the key stays stable if the list is ever reordered
or filtered — agreed that's the better call over an array index here.

PR #9013, opened 2026-07-29
fix(ui): preserve page context on rule action failures

### LINKED ISSUE #8984: Rule action failures replace the entire page with an error screen
## Summary

When enabling, disabling, or reconfiguring a rule fails (for example due to a transient network failure or a server-side error), the application replaces the entire Rules page with a generic full-screen error page...

This causes the user to lose all in-page context (navigation, breadcrumbs, tabs, and rule list) even though the failure is limited to a single rule operation...!

Instead of surfacing a scoped error for the failed action, the entire page becomes unusable until it is reloaded...

---

## Steps to reproduce

1. Start Apicurio Registry normally
2. Open any Rules page, for example:
   - Global Rules
   - Group Rules
   - Artifact Rules
3. Trigger an **Enable**, **Disable**, or **Configure** rule action
4. Before the request completes, force the corresponding API request to fail (for example by returning HTTP 500 or simulating a network failure)
5. Wait for the request to complete

---

## Actual behaviour

After the failed request:

- the entire page is replaced by the generic error screen
- navigation and breadcrumbs disappear
- the rules list is no longer visible
- the user loses the surrounding page context even though only a single rule operation failed

The failure is not scoped to the affected rule action.

---

## Expected behaviour

A failed rule mutation should only affect the operation being performed!

The application should:

- keep the current page visible
- preserve navigation, breadcrumbs, tabs, and the rules list
- surface an inline or localised error indicating that the rule operation failed
- leave the user on the current page so they can retry the action

---

## Additional observations

This behaviour can be reproduced on all three rule-management pages:

- Global Rules
- Group Rules
- Artifact Rules

The issue appears whenever a rule enable/disable/configuration request fails, making transient backend or network failures significantly more disruptive than necessary..!

Because the failure is limited to a single mutation request, replacing the entire page with a full-screen error results in unnecessary loss of user context!

Adii-45:
## Summary

Closes #8984

This PR prevents rule enable/disable/configure failures from replacing the entire page with a generic error screen.

Instead of updating the UI optimistically before the API request completes, the rule state is now updated only after a successful response. If the request fails, the current page remains visible and a scoped inline error is displayed near the affected rules list!

## Changes

- Delay rule state updates until the corresponding API request succeeds
- Replace full-page error handling with an inline, dismissible error message
- Apply the fix consistently across:
  - Global Rules
  - Artifact Rules
  - Group Rules
- Preserve page navigation, breadcrumbs, tabs, and surrounding context when a rule action fails
- Add Playwright regression tests covering the failure path on all affected rule pages

## Testing

Verified manually by reproducing the failure on:

- Global Rules
- Artifact Rules
- Group Rules

Also verified with:

- [x] Playwright regression tests
- [x] Existing test suite
- [x] TypeScript type checking
- [x] ESLint

## Result

Before:
- Rule action failure replaced the entire page with a generic error screen
- Navigation context and rules list were lost

After:
- The page remains visible
- The rule state is not updated unless the request succeeds
- A scoped inline error is shown for the failed operation

### DIFF
--- a/ui/tests/specs/explore.spec.ts
+++ b/ui/tests/specs/explore.spec.ts
@@ -68,7 +68,7 @@ test("Explore - Group specific rules, enable failure keeps the page intact", asy
             await route.fulfill({
                 status: 500,
                 contentType: "application/json",
-                body: JSON.stringify({ message: "Simulated failure" })
+                body: JSON.stringify({ detail: "Simulated failure" })
             });
         } else {
             await route.continue();
@@ -84,7 +84,7 @@ test("Explore - Group specific rules, enable failure keeps the page intact", asy
 
     // The page must remain intact - no full-page error, the group header/tabs are still there.
     await expect(page.getByTestId("group-rules-tab")).toBeVisible();
-    await expect(page.getByTestId("rule-action-error")).toBeVisible();
+    await expect(page.getByTestId("rule-action-error")).toContainText("Simulated failure");
 
     // The rule must NOT show as enabled - the failed request must not have been applied optimistically.
     await expect(page.getByTestId("rules-compatibility-enable")).toBeVisible();
@@ -211,7 +211,7 @@ test("Explore - Artifact specific rules, enable failure keeps the page intact", 
             await route.fulfill({
                 status: 500,
                 contentType: "application/json",
-                body: JSON.stringify({ message: "Simulated failure" })
+                body: JSON.stringify({ detail: "Simulated failure" })
             });
         } else {
             await route.continue();
@@ -227,7 +227,7 @@ test("Explore - Artifact specific rules, enable failure keeps the page intact", 
 
     // The page must remain intact - no full-page error, the artifact header/tabs are still there.
     await expect(page.getByTestId("artifact-rules-tab")).toBeVisible();
-    await expect(page.getByTestId("rule-action-error")).toBeVisible();
+    await expect(page.getByTestId("rule-action-error")).toContainText("Simulated failure");
 
     // The rule must NOT show as enabled - the failed request must not have been applied optimistically.
     await expect(page.getByTestId("rules-compatibility-enable")).toBeVisible();
--- a/ui/tests/specs/explore.spec.ts
+++ b/ui/tests/specs/explore.spec.ts
@@ -61,6 +61,37 @@ test("Explore - Group specific rules", async ({ page }) => {
 });
 
 
+test("Explore - Group specific rules, enable failure keeps the page intact", async ({ page }) => {
+    // Force the rule-creation request to fail, simulating a backend/network error.
+    await page.route("**/apis/registry/v3/groups/ExploreTest/rules", async route => {
+        if (route.request().method() === "POST") {
+            await route.fulfill({
+                status: 500,
+                contentType: "application/json",
+                body: JSON.stringify({ message: "Simulated failure" })
+            });
+        } else {
+            await route.continue();
+        }
+    });
+
+    await page.goto(`${REGISTRY_UI_URL}/explore/ExploreTest`);
+    await page.getByTestId("group-rules-tab").click();
+    await expect(page.locator("#compatibility-rule-name")).toContainText("Compatibility rule");
+
+    // Attempt to enable the rule; the mocked request fails.
+    await page.getByTestId("rules-compatibility-enable").click();
+
+    // The page must remain intact - no full-page error, the group header/tabs are still there.
+    await expect(page.getByTestId("group-rules-tab")).toBeVisible();
+    await expect(page.getByTestId("rule-action-error")).toBeVisible();
+
+    // The rule must NOT show as enabled - the failed request must not have been applied optimistically.
+    await expect(page.getByTestId("rules-compatibility-enable")).toBeVisible();
+    await expect(page.getByTestId("rules-compatibility-disable")).toHaveCount(0);
+});
+
+
 test("Explore - Create artifact", async ({ page }) => {
     // Navigate to the group details page
     await page.goto(`${REGISTRY_UI_URL}/explore/ExploreTest`);
@@ -170,6 +201,37 @@ test("Explore - Artifact specific rules", async ({ page }) => {
     // Select "syntax only" config option
     await page.getByTestId("validity-config-syntax").click();
     expect(page.getByTestId("rules-validity-disable")).toBeDefined();
+});
+
+
+test("Explore - Artifact specific rules, enable failure keeps the page intact", async ({ page }) => {
+    // Force the rule-creation request to fail, simulating a backend/network error.
+    await page.route("**/apis/registry/v3/groups/ExploreTest/artifacts/MyArtifact/rules", async route => {
+        if (route.request().method() === "POST") {
+            await route.fulfill({
+                status: 500,
+                contentType: "application/json",
+                body: JSON.stringify({ message: "Simulated failure" })
+            });
+        } else {
+            await route.continue();
+        }
+    });
+
+    await page.goto(`${REGISTRY_UI_URL}/explore/ExploreTest/MyArtifact`);
+    await page.getByTestId("artifact-rules-tab").click();
+    await expect(page.locator("#compatibility-rule-name")).toContainText("Compatibility rule");
+
+    // Attempt to enable the rule; the mocked request fails.
+    await page.getByTestId("rules-compatibility-enable").click();
+
+    // The page must remain intact - no full-page error, the artifact header/tabs are still there.
+    await expect(page.getByTestId("artifact-rules-tab")).toBeVisible();
+    await expect(page.getByTestId("rule-action-error")).toBeVisible();
+
+    // The rule must NOT show as enabled - the failed request must not have been applied optimistically.
+    await expect(page.getByTestId("rules-compatibility-enable")).toBeVisible();
+    await expect(page.getByTestId("rules-compatibility-disable")).toHaveCount(0);
 });
 
 
--- a/ui/tests/specs/globalRules.spec.ts
+++ b/ui/tests/specs/globalRules.spec.ts
@@ -81,7 +81,7 @@ test("Global Rules - Enable Validity Rule failure keeps the page intact", async 
             await route.fulfill({
                 status: 500,
                 contentType: "application/json",
-                body: JSON.stringify({ message: "Simulated failure" })
+                body: JSON.stringify({ detail: "Simulated failure" })
             });
         } else {
             await route.continue();
@@ -95,7 +95,7 @@ test("Global Rules - Enable Validity Rule failure keeps the page intact", async 
 
     // The page must remain intact - no full-page error, the other rules are still there.
     await expect(page.locator("div.rule")).toHaveCount(3);
-    await expect(page.getByTestId("rule-action-error")).toBeVisible();
+    await expect(page.getByTestId("rule-action-error")).toContainText("Simulated failure");
 
     // The rule must NOT show as enabled - the failed request must not have been applied optimistically.
     await expect(page.getByTestId("rules-validity-enable")).toBeVisible();
--- a/ui/tests/specs/globalRules.spec.ts
+++ b/ui/tests/specs/globalRules.spec.ts
@@ -4,74 +4,76 @@ const REGISTRY_UI_URL: string = process.env["REGISTRY_UI_URL"] || "http://localh
 
 test.describe.configure({ mode: "parallel" });
 
-test.beforeEach(async ({ page }) => {
-    await page.goto(REGISTRY_UI_URL);
-    await expect(page).toHaveTitle(/Apicurio Registry/);
-    await page.getByTestId("rules-tab").click();
-});
+test.describe("Global Rules", () => {
+    test.beforeEach(async ({ page }) => {
+        await page.goto(REGISTRY_UI_URL);
+        await expect(page).toHaveTitle(/Apicurio Registry/);
+        await page.getByTestId("rules-tab").click();
+    });
 
-test("Global Rules - List Rules", async ({ page }) => {
-    await expect(page.locator("div.rule")).toHaveCount(3);
-    await expect(page.locator("#validity-rule-name")).toContainText("Validity rule");
-    await expect(page.locator("#compatibility-rule-name")).toContainText("Compatibility rule");
-    await expect(page.locator("#integrity-rule-name")).toContainText("Integrity rule");
-});
+    test("Global Rules - List Rules", async ({ page }) => {
+        await expect(page.locator("div.rule")).toHaveCount(3);
+        await expect(page.locator("#validity-rule-name")).toContainText("Validity rule");
+        await expect(page.locator("#compatibility-rule-name")).toContainText("Compatibility rule");
+        await expect(page.locator("#integrity-rule-name")).toContainText("Integrity rule");
+    });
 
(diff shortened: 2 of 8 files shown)


Adii-45:
Hi! The License Compliance check appears to be failing on my PR, but this change only modifies existing UI components and Playwright tests - no dependencies or third-party code were added... Is this an expected repository-side check, or is there anything specific I should investigate?

ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx
```diff
@@ -124,36 +125,42 @@ export const ArtifactPage: FunctionComponent<PageProperties> = () => {
...
     const doEnableRule = (ruleType: string): void => {
         logger.debug("[ArtifactPage] Enabling rule:", ruleType);
+        setRuleActionError(undefined);
         let config: string = "FULL";
         if (ruleType === "COMPATIBILITY") {
             config = "BACKWARD";
         }
-        groups.createArtifactRule(groupId as string, artifactId as string, ruleType, config).catch(error => {
-            setPageError(toPageError(error, `Error enabling "${ ruleType }" artifact rule.`));
+        groups.createArtifactRule(groupId as string, artifactId as string, ruleType, config).then(() => {
```
paoloantinori, line 132: Stale-closure risk: `rules` is captured at click time, but this `setRules` now runs after an async hop. Two in-flight enable/disable/configure actions will each resolve against the same stale snapshot and clobber each other. Use the functional updater: `setRules(prev => [...prev, { config, ruleType: ruleType as RuleType }])`. Same for `doDisableRule` and `doConfigureRule` in this file, and the identical handlers in `GroupPage.tsx` / `RulesPage.tsx`.

ui/ui-app/src/app/pages/group/GroupPage.tsx
```diff
@@ -157,36 +158,42 @@ export const GroupPage: FunctionComponent<PageProperties> = () => {
...
     const doEnableRule = (ruleType: string): void => {
         logger.debug("[GroupPage] Enabling rule:", ruleType);
+        setRuleActionError(undefined);
         let config: string = "FULL";
         if (ruleType === "COMPATIBILITY") {
             config = "BACKWARD";
         }
-        groups.createGroupRule(groupId as string, ruleType, config).catch(error => {
-            setPageError(toPageError(error, `Error enabling "${ ruleType }" group rule.`));
+        groups.createGroupRule(groupId as string, ruleType, config).then(() => {
```
paoloantinori, line 167: Same stale-closure issue as `ArtifactPage`: this `setRules([...rules, ...])` runs in the async `.then`, so `rules` is the click-time snapshot. Switch to `setRules(prev => [...prev, ...])` to avoid losing a concurrent rule update.

ui/ui-app/src/app/pages/rules/RulesPage.tsx
```diff
@@ -27,36 +28,42 @@ export const RulesPage: FunctionComponent<PageProperties> = () => {
...
     const doEnableRule = (ruleType: string): void => {
         logger.debug("[RulesPage] Enabling global rule:", ruleType);
+        setRuleActionError(undefined);
         let config: string = "FULL";
         if (ruleType === "COMPATIBILITY") {
             config = "BACKWARD";
         }
-        admin.createRule(ruleType, config).catch(error => {
-            setPageError(toPageError(error, `Error enabling "${ ruleType }" global rule.`));
+        admin.createRule(ruleType, config).then(() => {
```
paoloantinori, line 38: Same as above — `setRules(prev => [...prev, { config, ruleType: ruleType as RuleType }])` here, and the functional form for disable (`prev => prev.filter`) / configure (`prev => prev.map`).

paoloantinori (commented):
The shift from optimistic updates (state change before the request) to update-on-success is the right direction for preserving page context — good call, and the Playwright regression tests (mock the POST to 500, assert the page stays intact and the enable button is still visible) are exactly the right coverage.

One real regression worth addressing before merge: the `.then()` callbacks reference the `rules` state captured at click time, which opens a stale-closure race that the old optimistic code didn't have.

- **Race on concurrent rule actions.** In all three pages the success callbacks use the captured array directly, e.g. in `ArtifactPage.doEnableRule`: `groups.createArtifactRule(...).then(() => { setRules([...rules, { config, ruleType }]); })`. The `rules` here is frozen to the render in which the user clicked. If the user enables rule A, then — while A's request is still in flight — enables rule B, both `.then` closures hold the same pre-click snapshot. When A resolves it sets `[...snapshot, A]`; when B resolves it sets `[...snapshot, B]`, clobbering A. Net effect: the backend has both rules enabled, but the UI shows only the last one (and a subsequent page reload would correct it — silent inconsistency until then). The same pattern applies to `doDisableRule` (`rules.filter`) and `doConfigureRule` (`rules.map`). The old optimistic version didn't hit this because each click updated state synchronously and the next click saw the updated array via re-render. The fix is the functional updater form everywhere: `setRules(prev => [...prev, { config, ruleType }])`, `setRules(prev => prev.filter(r => r.ruleType !== ruleType))`, and `setRules(prev => prev.map(...))`. Applies identically to `ArtifactPage.tsx`, `GroupPage.tsx`, and `RulesPage.tsx`.
- **Minor — error message quality.** `setRuleActionError(error?.message || ...)` surfaces the raw caught error's `.message`, whereas the previous path ran it through `toPageError(...)`. For an HTTP 500 the raw message may be a low-level transport/JSON string rather than the user-facing summary `toPageError` produced. Worth double-checking that `error?.message` reads well for a typical backend failure; if not, keeping `toPageError` and pulling just the message from it would preserve the UX.

Adii-45:
Thanks for the detailed review @paoloantinori! I've updated all three pages to use functional `setRules(prev => ...)` updates for the async rule handlers, which avoids the stale-closure issue you pointed out. I also reviewed the `error?.message` suggestion and left it unchanged for now since it matches the existing error-handling pattern used elsewhere in the codebase. Please let me know if you'd still prefer a change there...

paoloantinori:
Thanks @Adii-45 — the functional `setRules(prev => ...)` updates across all three pages look good and properly close the stale-closure window. Agreed on leaving `error?.message` as-is for consistency with the surrounding pattern; not worth a divergence here. The Playwright tests asserting the page stays intact and the toggle doesn't flip to "enabled" on failure are exactly the regression coverage this needed. LGTM from my side.

Adii-45:
@EricWittmann Hi! All requested review feedback has been addressed, and all checks are now passing. The PR should be ready for merge whenever you have a chance... Thanks!

EricWittmann (commented):
Thanks @Adii-45 — the fix looks solid and the Playwright tests are great coverage. One thing to address before we merge:

**Add a loading indicator during the API call.** Since rule state is now updated only after the request succeeds, there's a gap where the user clicks enable/disable and nothing visibly happens until the response comes back. On a slow server or high-latency connection, this could look like the click didn't register, leading users to click again.

There's a "Please Wait" modal pattern already used elsewhere in the UI (e.g. `isPleaseWaitModalOpen` / `setPleaseWaitMessage` — you can see it wired up in `ArtifactPage.tsx` already). You could use that same pattern here, or alternatively a spinner/disabled state on the rule toggle itself. Either way, the user needs some visual feedback that the action is in progress.

Adii-45:
Thanks for the suggestion, @EricWittmann! I've added a loading state for the rule actions while the API request is in progress by disabling the action and showing PatternFly's built-in loading indicator. The change is applied consistently across all three rule pages while keeping the previously reviewed behavior intact. Please let me know if you'd like any further adjustments!

ui/ui-app/src/app/pages/rules/RulesPage.tsx
```diff
@@ -27,36 +29,51 @@ export const RulesPage: FunctionComponent<PageProperties> = () => {
...
+            setRules(prev => [...prev, { config, ruleType: ruleType as RuleType }]);
+        }).catch(error => {
+            setRuleActionError(error?.message || `Error enabling "${ ruleType }" global rule. Please try again.`);
```
paoloantinori: The `.catch` handlers now use `error?.message` and no longer go through `toPageError(error, …)`, which these pages previously used to format rule-action errors (and still use for page-load errors). For a generated SDK client, `error.message` is frequently a generic transport error (e.g. "Request failed") rather than the backend's response `message`, so the inline alert could end up less informative than before. Worth confirming `error.message` actually carries the server message here, or routing it through `toPageError`. The identical change is in `GroupPage.tsx` and `ArtifactPage.tsx`.

ui/tests/specs/explore.spec.ts
```diff
@@ -61,6 +61,37 @@ test("Explore - Group specific rules", async ({ page }) => {
...
+    await page.goto(`${REGISTRY_UI_URL}/explore/ExploreTest`);
+    await page.getByTestId("group-rules-tab").click();
+    await expect(page.locator("#compatibility-rule-name")).toContainText("Compatibility rule");
+
+    // Attempt to enable the rule; the mocked request fails.
+    await page.getByTestId("rules-compatibility-enable").click();
+
+    // The page must remain intact - no full-page error, the group header/tabs are still there.
+    await expect(page.getByTestId("group-rules-tab")).toBeVisible();
+    await expect(page.getByTestId("rule-action-error")).toBeVisible();
```
paoloantinori: This test stubs the backend with `{ message: "Simulated failure" }` but only asserts the error alert is visible. Consider also asserting the alert text (e.g. `toContainText("Simulated failure")`); otherwise a regression where `error?.message` doesn't surface the backend message would slip through. The artifact and global-rules failure tests have the same gap.

paoloantinori (commented):
This is a clean, well-scoped fix: rule state is now applied only after a successful response (using functional `setRules(prev => …)` updates), failures render a scoped, dismissible inline `Alert` instead of replacing the whole page, and a `pendingRuleType` drives a loading/disabled state on the action buttons. The pattern is consistent across `RulesPage.tsx`, `GroupPage.tsx`, and `ArtifactPage.tsx`, with parallel regression tests.

- The main concern is error-message quality: the new `.catch` handlers surface `error?.message` and drop the `toPageError(error, …)` helper these pages used before (`RulesPage.tsx:41`, with the same change in `GroupPage.tsx` and `ArtifactPage.tsx`). For a generated SDK client, `error.message` is often a generic transport-level string rather than the backend's response `message` (e.g. the mocked `"Simulated failure"`), so the inline alert may be less informative than the prior full-page error.
- The failure tests mock `{ message: "Simulated failure" }` but only assert the alert is *visible* (`explore.spec.ts:87`, `globalRules.spec.ts`), not its text — so the above regression wouldn't be caught. Easy win: also assert the alert text.
- Nice improvement: the functional `setRules(prev => …)` form removes the stale-closure risk of the old `setRules(rules.filter(...))`.
- Worth confirming: the new global-rules test (`globalRules.spec.ts:77`) doesn't call `page.goto`, unlike the two explore tests added here — make sure it relies on a `beforeEach` navigation, else it starts on `about:blank`.

Adii-45:
Thanks for the review, @paoloantinori! I've updated the inline rule-action errors to surface the backend-provided error details instead of the generic transport message, and strengthened the Playwright regression tests to assert the alert text as well. All tests are passing now - please let me know if you spot anything else!

paoloantinori:
Thanks @Adii-45 — surfacing `error?.detail || error?.title` before the generic fallback is exactly the right layering, and asserting the rendered alert text in the Playwright tests closes the loop nicely. The fix looks complete to me from a review standpoint.

Adii-45:
Thanks again for the detailed review and feedback, @paoloantinori! Really appreciate you taking the time to go through the changes..!

I've addressed all the requested feedback, so I'll wait to see if @EricWittmann has anything else to add. Otherwise, the PR should be ready to merge whenever you get a chance. Thanks!

paoloantinori:
All previous feedback is addressed — functional `setRules(prev => …)` updates, loading/disabled state on the action buttons, `error?.detail || error?.title` surfacing with the Playwright alert-text assertions, all applied consistently across the three pages (Global, Artifact, Group). Nothing else from my side; agree it's ready for @EricWittmann's final check. Thanks for the careful iteration!

ui/tests/specs/globalRules.spec.ts
```diff
@@ -73,3 +73,31 @@ test("Global Rules - Enable Integrity Rule", async ({ page }) => {
...
+                status: 500,
+                contentType: "application/json",
+                body: JSON.stringify({ detail: "Simulated failure" })
+            });
+        } else {
+            await route.continue();
+        }
+    });
+
+    await expect(page.locator("#validity-rule-name")).toContainText("Validity rule");
```
paoloantinori, line 97: This test omits `page.goto(...)` before the first assertion, whereas the equivalent failure tests in `explore.spec.ts` (group rules ~L78, artifact rules ~L221) navigate explicitly before asserting. Unless `globalRules.spec.ts` loads the rules page in a `beforeEach`, this `expect(page.locator("#validity-rule-name"))` will time out against `about:blank` on a fresh page. Please confirm a `beforeEach` navigation exists here, or add the missing `page.goto(...)` (the group/artifact tests route, then `goto`, then assert — this one goes straight from `page.route` to `expect`).

paoloantinori (commented):
Overall this is a sound, well-scoped refactor: moving each rule-state mutation into the promise `.then` (and switching to functional `setRules(prev => …)` updates) correctly removes the optimistic update, and replacing `setPageError` with a scoped, dismissible `Alert` is applied consistently across `RulesPage.tsx`, `GroupPage.tsx`, and `ArtifactPage.tsx` plus their tab contents and `RuleList.tsx`. The functional state updates are actually an improvement over the prior closure pattern, and the failure-path Playwright coverage on all three pages is a nice addition.

- **`ui/tests/specs/globalRules.spec.ts:91`** — the new "Enable Validity Rule failure" test never calls `page.goto(...)` before its first assertion, unlike its siblings in `explore.spec.ts` (the group-rules test ~L78 and artifact-rules test ~L221 both navigate explicitly before asserting). If this file does not navigate via a `beforeEach`, the test will time out at `expect(page.locator("#validity-rule-name"))` because the page is still `about:blank`. Since the PR lists the Playwright suite as verified, this is worth a quick double-check — confirm a `beforeEach` navigation exists, or add the missing `page.goto`.
- Minor: `pendingRuleType` is a single shared string, so two rapid actions on *different* rule types will overwrite each other's pending flag (the first `.finally` clears `pendingRuleType` while the second request is still in flight, briefly re-enabling buttons). Fine for this low-frequency admin UI, but a `Set<string>` would make the loading/disabled state robust against concurrent actions. Not blocking.
- The error fallback chain `error?.detail || error?.title || <generic>` is good — it degrades gracefully for network errors (plain `Error` with no `.detail`/`.title`) rather than crashing.

Adii-45:
Thanks for catching that @paoloantinori! I've addressed this in the latest commit by updating the test to follow the same navigation pattern as the equivalent failure tests before making assertions. I reran the affected Playwright tests to verify everything passes. Please let me know if there's anything else you'd like me to adjust!

paoloantinori:
Thanks @Adii-45 — updating the test to follow the same navigation pattern as the equivalent failure tests before asserting is the right fix, and rerunning the Playwright suite to confirm is appreciated. No further concerns from me.

EricWittmann:
Thanks for the great work @Adii-45! This is a solid fix — the shift from optimistic updates to update-on-success with inline error handling is exactly the right approach. Appreciate your responsiveness through multiple review rounds. Nice contribution!

PR #9014, opened 2026-07-29
fix(rest): return 409 instead of 500 for duplicate content

### LINKED ISSUE #8872: bug: ContentAlreadyExistsException missing from HttpStatusCodeMap (returns 500 instead of 409)
### Describe the bug

When `ContentAlreadyExistsException` is thrown, the REST exception mapping layer (`HttpStatusCodeMap`) falls back to `HTTP_INTERNAL_ERROR` (500) instead of returning `HTTP_CONFLICT` (409).

### Cause
`ContentAlreadyExistsException` extends `AlreadyExistsException`. While `HttpStatusCodeMap` registers every other `AlreadyExistsException` subclass (`ArtifactAlreadyExistsException`, `BranchAlreadyExistsException`, `GroupAlreadyExistsException`, `RoleMappingAlreadyExistsException`, `RuleAlreadyExistsException`, `VersionAlreadyExistsException`, `VersionAlreadyExistsOnBranchException`) mapped to `HTTP_CONFLICT` (409), `ContentAlreadyExistsException` was missing from `HttpStatusCodeMap.init()`.

Because `HttpStatusCodeMap.getCode(Class<?> exceptionClass)` performs an exact class lookup against its internal map, `ContentAlreadyExistsException` fails to match and returns `HTTP_INTERNAL_ERROR` (500).

### Expected behavior
`ContentAlreadyExistsException` should be registered in `HttpStatusCodeMap` mapped to `HTTP_CONFLICT` (409).

### Environment
- Repository: `Apicurio/apicurio-registry` (`main` branch)
- Target file: `app/src/main/java/io/apicurio/registry/services/http/HttpStatusCodeMap.java`

bhuvan-somisetty:
## Summary

Fixes #8872. `ContentAlreadyExistsException` returns HTTP 500 instead of HTTP 409 when content already exists in storage.

## Root Cause

`ContentAlreadyExistsException` extends `AlreadyExistsException`, which is already mapped to `HTTP_CONFLICT` in `HttpStatusCodeMap`. However, `HttpStatusCodeMap.getCode(Class<?> exceptionClass)` looks up the map by exact class, not by inheritance. Every other `AlreadyExistsException` subclass (`ArtifactAlreadyExistsException`, `BranchAlreadyExistsException`, `GroupAlreadyExistsException`, `RoleMappingAlreadyExistsException`, `RuleAlreadyExistsException`, `VersionAlreadyExistsException`, `VersionAlreadyExistsOnBranchException`) has its own entry in `HttpStatusCodeMap.init()`, but `ContentAlreadyExistsException` was never registered. Without a match, the lookup falls back to the default `HTTP_INTERNAL_ERROR` (500).

## Changes

- Registered `ContentAlreadyExistsException` in `HttpStatusCodeMap.init()`, mapped to `HTTP_CONFLICT` (409), following the exact pattern already used for the other `AlreadyExistsException` subclasses.
- Added `HttpStatusCodeMapTest` covering:
  - `ContentAlreadyExistsException` maps to 409.
  - A sibling `AlreadyExistsException` subclass (`ArtifactAlreadyExistsException`) still maps to 409, confirming the existing entries are untouched.
  - An unmapped exception still falls back to the 500 default, confirming the fallback behavior itself is unchanged.

## Test plan

- [x] `./mvnw checkstyle:check -pl app` passes
- [x] `./mvnw test -pl app -Dtest=HttpStatusCodeMapTest` passes (3/3)
- [x] Verified `ContentAlreadyExistsException` maps to HTTP 409 via `HttpStatusCodeMap`, matching the existing pattern for sibling `AlreadyExistsException` subclasses


app/src/main/java/io/apicurio/registry/services/http/HttpStatusCodeMap.java
```diff
@@ -90,6 +91,7 @@ void init() {
         map.put(BranchNotFoundException.class, HTTP_NOT_FOUND);
         map.put(ConfigPropertyNotFoundException.class, HTTP_NOT_FOUND);
         map.put(ConflictException.class, HTTP_CONFLICT);
+        map.put(ContentAlreadyExistsException.class, HTTP_CONFLICT);
```
paoloantinori, line 96: This is correct and matches the sibling entries. The recurring footgun here is the exact-class lookup in `getCode` — a follow-up that walks the superclass chain would let this `AlreadyExistsException`-family registration happen once at the parent. Not something to change in this PR, just noting the pattern.

paoloantinori (commented):
Clean fix — the root-cause analysis is spot on, and the three tests (positive, sibling-regression, fallback) cover exactly the right surface. Registering `ContentAlreadyExistsException` against `HTTP_CONFLICT` follows the established pattern for the other `AlreadyExistsException` subclasses.

One design observation (not a blocker for this PR): the fact that `getCode` does an exact-class lookup is what made this bug possible in the first place — every new `AlreadyExistsException` subclass has to be manually registered, and the next contributor who adds an exception will hit the same 500-instead-of-409 silent regression. Worth considering, in a follow-up, having `getCode` walk the superclass chain (`clazz = clazz.getSuperclass()`) until it finds a mapped class and only then fall back to the default. That would make the parent `AlreadyExistsException → 409` entry cover all subclasses automatically. Flagging it for the future; this PR's targeted fix is the right scope for now.

bhuvan-somisetty:
Thanks for the review, @paoloantinori! Good catch on the root cause, you're right that the exact-class lookup was the real gap here.

I went ahead and implemented the follow-up you suggested: `HttpStatusCodeMap.getCode` now walks up the superclass chain until it finds a mapped class, falling back to 500 only if nothing in the hierarchy matches. So the existing `AlreadyExistsException -> 409` entry now covers any current or future subclass automatically, and we won't hit this same silent-500 regression again for the next contributor who adds one.

Added a test (`unregisteredAlreadyExistsSubclassInheritsParentMapping`) with a locally-defined, unregistered `AlreadyExistsException` subclass to lock in that behavior, alongside the existing tests. Checkstyle and the module's unit tests are passing locally.

Let me know if you'd like anything adjusted, thanks!

bhuvan-somisetty:
@EricWittmann @carlesarnal could one of you take a look when you get a chance?

The code changes here are done and green locally (checkstyle + unit tests passing), and @paoloantinori's review feedback has been addressed in the latest commit. The one remaining failing check, `Verify / Verification Gate`, isn't caused by the code it fails because the PR still carries the `lifecycle/new` / `lifecycle/waiting-on-maintainer` labels, and the gate workflow only runs the full suite once a `lifecycle/ready-for-review` (or similar) label is applied. As an external contributor I don't have permission to change PR labels, so I can't unblock that gate myself.
thanks

bhuvan-somisetty:
Quick follow-up on the newly failing `License Compliance` (FOSSA) check. I checked and it's not related to this PR, it's already failing with the same "2 issues found" result on the current `main` HEAD (05667bff). This PR only touches `HttpStatusCodeMap.java` and its test, no dependency changes at all, so there's nothing in the diff that could trigger a license issue.

Flagging this per the contributor checklist guidance for CI failures unrelated to the change, looks like a pre-existing FOSSA finding on the repo that's independent of this PR. @EricWittmann could you or someone with FOSSA access take a look when possible? Let me know if you want me to dig into it further, but I don't think it's something fixable from within this branch.
thanks

paoloantinori:
Thanks @bhuvan-somisetty — confirming the follow-up here looks good. The superclass-chain walking in `getCode` (together with the `unregisteredAlreadyExistsSubclassInheritsParentMapping` test) addresses exactly the "implicit fallback" concern from my earlier review, and it's a clean generalization of the explicit registration. The FOSSA failure being pre-existing on `main` (unrelated to this PR's no-dependency changes) is a fair diagnosis on my read as well. Nothing further from my side on the code.

bhuvan-somisetty:
Thanks for confirming, @paoloantinori, glad the generalization landed the way you were picturing it. Appreciate you taking the time to go through it twice.

Just waiting on a maintainer for the label/FOSSA stuff now, rest is done from my end.

paoloantinori:
Thanks @bhuvan-somisetty — nothing more needed from my side. The superclass-chain walk plus the `unregisteredAlreadyExistsSubclassInheritsParentMapping` test is the right generalization. Agreed the FOSSA failure is unrelated (pre-existing on `main`). Just the maintainer label/acceptance left from here.

bhuvan-somisetty:
Thanks again for the thorough review, @paoloantinori really appreciate you taking the time to go through both rounds. I'll sit tight and wait for a maintainer(@EricWittmann) to accept/label the PR so the verification gate can run.

paoloantinori:
Acknowledged, @bhuvan-somersetty — nothing more needed from my side. The change is in good shape; just waiting on @EricWittmann / @carlesarnal to accept/label so the verification gate can run.

EricWittmann:
:eyes:

bhuvan-somisetty:
@EricWittmann all checks are passing now, would appreciate a look when you get a chance. Thanks!

PR #9017, opened 2026-07-29
fix(ui): respect declared default for boolean prompt template variables

### LINKED ISSUE #9015: Boolean variable defaults ignored when synced asynchronously in PromptTemplateTestPanel
## Description

In `PromptTemplateTestPanel`, when a `PROMPT_TEMPLATE` artifact's version
content is dereferenced asynchronously, boolean-typed variables are
seeded to `false` on load, ignoring their declared `default` value.

This happens because `DocumentationTabContent` renders
`PromptTemplateVisualizer` (and its child `PromptTemplateTestPanel`)
immediately with empty content, then updates once
`getArtifactVersionContentDereferenced()` resolves. When
`PromptTemplateTestPanel`'s `props.variables` changes from `undefined`
to the real variable list, a `useEffect` seeds initial values for any
variable not already in state:

```js
next[name] = type === "boolean" ? false : (variable.default ?? "");
```

This unconditionally forces `false` for booleans, regardless of any
declared `default: true`.

## Steps to Reproduce

1. Create a `PROMPT_TEMPLATE` artifact with a boolean variable that has
   `default: true`, e.g.:
```json
   {
     "template": "Hello {{name}}, verbose={{verbose}}",
     "variables": {
       "name": { "type": "string", "default": "World" },
       "verbose": { "type": "boolean", "default": true, "description": "Enable verbose output" }
     }
   }
```
2. Open the artifact's version page → **Documentation** tab.
3. Compare the **Variables** table (shows `Default: true`) against the
   **Test Prompt** checkbox for `verbose` (renders unchecked).

## Expected Behavior

The `verbose` checkbox should render checked on load, matching its
declared `default: true`.

## Actual Behavior

The checkbox renders unchecked, contradicting the Variables table
directly above it on the same page.

## Video

https://github.com/user-attachments/assets/646b2923-32fd-4fc6-8982-3329a308105e

## Suggested Fix

```js
next[name] = variable.default ?? (type === "boolean" ? false : "");
```

This only falls back to `false` when no default is declared.

Faisal77666:
Fixes #9015

## What changed
In `PromptTemplateTestPanel`, the `useEffect` that seeds values for newly
appearing prompt variables unconditionally forced `boolean`-typed
variables to `false`, ignoring any declared `default`.

Before:
```js
next[name] = type === "boolean" ? false : (variable.default ?? "");
```

After:
```js
next[name] = variable.default ?? (type === "boolean" ? false : "");
```

This only falls back to `false` when no default is declared, so a
boolean variable with `default: true` is now seeded correctly.

## Why it's reachable
For `PROMPT_TEMPLATE` artifacts, content is dereferenced asynchronously
(`DocumentationTabContent`), so `PromptTemplateTestPanel` mounts once with
`variables={undefined}` and then updates once the async fetch resolves.
That update path runs this `useEffect`, so this triggers on every load of
a Prompt Template artifact's Documentation tab — not an edge case.

## Before

[raw-recording (6).webm](https://github.com/user-attachments/assets/4ee4600b-3372-4ad0-b984-98a8330448cc)

## After

[raw-recording (7).webm](https://github.com/user-attachments/assets/40e36a82-272a-43af-ab82-5b80f44406a2)

## How to test
1. Create a `PROMPT_TEMPLATE` artifact with a boolean variable that has
   `default: true`.
2. Open its version's Documentation tab.
3. Confirm the Variables table shows `Default: true` for that variable,
   and the Test Prompt checkbox now renders checked to match.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -59,7 +59,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
             variablesList.forEach(({ name, variable }) => {
                 if (!(name in next)) {
                     const type = (variable.type || "string").toLowerCase();
-                    next[name] = type === "boolean" ? false : (variable.default ?? "");
+                    next[name] = variable.default ?? (type === "boolean" ? false : "");
                 }
             });
             return next;


Faisal77666:
@paoloantinori please review pr if you have time

paoloantinori:
Hi @Faisal77666 — reviewed, thanks for the clear write-up. The one-line change is correct: moving `variable.default` to the front of the `??` chain means a declared default wins for any type (fixing the `boolean` + `default: true` case), while the `false` / `""` fallbacks still apply when no default is declared. The reasoning about `PromptTemplateTestPanel` mounting with `variables={undefined}` and re-running the `useEffect` on the async dereference checks out, so this is a real reachable path, not an edge case. 

One small thing worth double-checking: the fix assumes `variable.default` is already in the right runtime type for the field (e.g. a boolean `true` for a boolean variable, a string for a string variable). If a template ever ships a default as the wrong type (say `default: "true"` on a boolean variable), the seeded value would now be the raw string rather than the previous coerced `false`. That's almost certainly fine given how templates declare defaults, but if the underlying schema validation doesn't enforce type-correctness on `default`, a defensive coercion here would make it bulletproof. Not blocking — the fix as-is is a clear improvement over the previous unconditional `false`.

Faisal77666:
@EricWittmann please review PR at your convenience

Faisal77666:
> Hi @Faisal77666 — reviewed, thanks for the clear write-up. The one-line change is correct: moving `variable.default` to the front of the `??` chain means a declared default wins for any type (fixing the `boolean` + `default: true` case), while the `false` / `""` fallbacks still apply when no default is declared. The reasoning about `PromptTemplateTestPanel` mounting with `variables={undefined}` and re-running the `useEffect` on the async dereference checks out, so this is a real reachable path, not an edge case.
> 
> One small thing worth double-checking: the fix assumes `variable.default` is already in the right runtime type for the field (e.g. a boolean `true` for a boolean variable, a string for a string variable). If a template ever ships a default as the wrong type (say `default: "true"` on a boolean variable), the seeded value would now be the raw string rather than the previous coerced `false`. That's almost certainly fine given how templates declare defaults, but if the underlying schema validation doesn't enforce type-correctness on `default`, a defensive coercion here would make it bulletproof. Not blocking — the fix as-is is a clear improvement over the previous unconditional `false`.

@paoloantinori thanks for your feedback!

Faisal77666:
@EricWittmann ready to merge

paoloantinori:
Thanks @Faisal77666 — the one-line move (`variable.default` to the front of the `??` chain) is correct, so a declared default now wins for any type. Already gave the Lgtm; nothing further from my side. Merge is a maintainer call.

PR #9020, opened 2026-07-29
fix(ui): coerce enum prompt template values to their declared type

### LINKED ISSUE #8997: fix(ui): enum-based prompt template variables submit dropdown values as strings, failing type validation
Registry Version: main (snapshot)
Persistence type: N/A - UI bug

In PromptTemplateTestPanel.tsx, variables with an `enum` list use a PatternFly `<FormSelect>`
for input, regardless of the variable's declared type (integer, number, boolean, string).

The FormSelect's onChange handler passes the raw HTML select value straight to setValue()
without parsing it according to the variable's type:

  <FormSelect
      value={values[name] || ""}
      onChange={(_event, val) => setValue(name, val)}
      aria-label={name}
  >

Since HTML <select> values are always strings, this means an integer/number/boolean variable
with an enum list always submits a string to the backend, even when the user picks a valid
enum option.

Steps to Reproduce
1. Create a PROMPT_TEMPLATE artifact with an integer variable using enum, e.g.:
   { "port": { "type": "integer", "enum": [80, 443, 8080], "default": 8080 } }
2. Open the Test Prompt panel.
3. Select an option (e.g. 443) from the dropdown.
4. Click Render.

Expected: Render succeeds with `port` passed as the number 443.
Actual: Validation fails with "port: Type mismatch: expected integer but got string" —
the exact value the user selected is rejected.

Root Cause
PromptTemplateTestPanel.tsx, renderField() (~line 107-120): the enum branch renders a
FormSelect whose onChange never coerces the string value to the variable's declared type
before calling setValue(). PromptRenderingService.java (line 163) then correctly rejects it
since it checks `value instanceof Integer || value instanceof Long` etc.

Suggested Fix

  onChange={(_event, val) => {
      let parsedVal: any = val;
      if (type === "integer") parsedVal = parseInt(val);
      else if (type === "number") parsedVal = parseFloat(val);
      else if (type === "boolean") parsedVal = val === "true";
      setValue(name, parsedVal);
  }}

I already have this fix working locally and can open a PR shortly if this direction looks right.

akshattalwar001:
Fixes #8997

the enum FormSelect in PromptTemplateTestPanel was just passing the raw
string straight to setValue(), no matter what type the variable actually
was. so any integer, number or boolean variable using enum always failed
on the backend with "type mismatch: expected X but got string" even when
you picked a totally valid option from the dropdown..

root cause is basically that onChange never coerced the value. the
integer/number branch a few lines below already does this with
parseInt/parseFloat but the enum branch never got the same treatment,
guessing it just got missed.

fix: added a small coerceEnumValue helper in a new
PromptTemplateTestPanel.utils.ts (kept it in its own file so it's
actually testable and doesn't trip the react-refresh lint rule for
exporting non-components from a component file). it parses the selected
value based on the declared type before it hits setValue, and also
handles the placeholder option ("-- select --", value "") so it doesn't
turn into NaN for integer/number types.

TESTING:
- added PromptTemplateTestPanel.utils.test.ts, 8 tests covering
  integer/number/boolean/string coercion plus the placeholder case for
  each.. all passing
- tsc --noEmit is clean, no new errors (there's some pre-existing
  unrelated ones from the unbuilt typescript-sdk client but nothing
  touching these files)
- eslint clean on the changed files

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -21,6 +21,7 @@ import {
 import { PromptVariable } from "./PromptTemplateViewer";
 import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
 import { RenderPromptResponse, RenderPromptValidationError } from "@models/RenderPromptResponse.ts";
+import { coerceEnumValue } from "./PromptTemplateTestPanel.utils";
 
 export type PromptTemplateTestPanelProps = {
     groupId: string;
@@ -108,7 +109,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
             return (
                 <FormSelect
                     value={values[name] || ""}
-                    onChange={(_event, val) => setValue(name, val)}
+                    onChange={(_event, val) => setValue(name, coerceEnumValue(val, type))}
                     aria-label={name}
                 >
                     <FormSelectOption key="placeholder" value="" label="-- Select --" />
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -108,7 +108,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
         if (variable.enum && variable.enum.length > 0) {
             return (
                 <FormSelect
-                    value={values[name] || ""}
+                    value={values[name] ?? ""}
                     onChange={(_event, val) => setValue(name, coerceEnumValue(val, type))}
                     aria-label={name}
                 >
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
@@ -1,0 +1,36 @@
+import { describe, expect, it } from "vitest";
+import { coerceEnumValue } from "./PromptTemplateTestPanel.utils";
+
+describe("coerceEnumValue", () => {
+    it("parses an integer enum selection to a number", () => {
+        expect(coerceEnumValue("443", "integer")).toBe(443);
+    });
+
+    it("parses a number enum selection to a float", () => {
+        expect(coerceEnumValue("3.14", "number")).toBe(3.14);
+    });
+
+    it("parses a boolean enum selection to true", () => {
+        expect(coerceEnumValue("true", "boolean")).toBe(true);
+    });
+
+    it("parses a boolean enum selection to false", () => {
+        expect(coerceEnumValue("false", "boolean")).toBe(false);
+    });
+
+    it("leaves a string enum selection untouched", () => {
+        expect(coerceEnumValue("prod", "string")).toBe("prod");
+    });
+
+    it("keeps the placeholder selection as an empty string for an integer variable", () => {
+        expect(coerceEnumValue("", "integer")).toBe("");
+    });
+
+    it("keeps the placeholder selection as an empty string for a number variable", () => {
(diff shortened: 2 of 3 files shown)


ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -108,7 +109,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
             return (
                 <FormSelect
                     value={values[name] || ""}
-                    onChange={(_event, val) => setValue(name, val)}
+                    onChange={(_event, val) => setValue(name, coerceEnumValue(val, type))}
```
paoloantinori, line 112: Heads-up — this coercion interacts badly with the `value={values[name] || ""}` binding on line 111 above. Once `setValue` stores a native `false` (the boolean "false" option) or `0` (an integer "0" option), `values[name] || ""` evaluates to `""` because both are falsy, so the FormSelect resets to the "-- Select --" placeholder immediately after the pick. Switching that fallback to `values[name] ?? ""` (nullish coalescing) preserves `false`/`0` and fixes the display. The unit tests don't surface this since they exercise `coerceEnumValue` in isolation rather than the integrated `value=` binding.

paoloantinori (commented):
Right direction — coercing the enum selection to the variable's declared type before `setValue` is exactly what was missing, and `PromptTemplateTestPanel.utils.test.ts` covers the coercion logic itself well. One regression to address before this lands, introduced by the interaction between the new coercion and an existing binding.

- **Falsy-value regression on the display binding (boolean `false` / numeric `0`).** `coerceEnumValue` now stores native `false` (boolean) and `0` (integer/number), but the controlled value one line above (`PromptTemplateTestPanel.tsx:111`) is still `value={values[name] || ""}`. Since `||` falls through on *any* falsy value, selecting the boolean option `"false"` (→ `false`) or an integer option `"0"` (→ `0`) makes the binding evaluate to `""`, so the dropdown snaps back to the "-- Select --" placeholder right after the pick. The stored `values` entry stays correct so submission isn't broken, but the UI looks like the selection was lost. The suite doesn't catch it because the tests only exercise `coerceEnumValue` in isolation, not the integrated `value=` binding. Minimal fix: switch the fallback to nullish coalescing — `value={values[name] ?? ""}` — which preserves `false`/`0` and only falls back for `null`/`undefined`. An integration-style assertion (pick `"false"`, assert the option stays selected) would lock it in.

- Minor: `parseInt`/`parseFloat` (`PromptTemplateTestPanel.utils.ts:5,7`) return `NaN` for unparseable input, which `JSON.stringify` renders as `null` on submission. This only arises from a malformed integer/number enum declaration and matches the existing text-input branch's behavior, so it's low priority — flagging for awareness only.

akshattalwar001:
thanks for catching that, you're right.. pushed a fix, switched the display binding to `values[name] ?? ""` instead of `|| ""` so false/0 selections don't get treated as empty anymore. reverified locally, all tests still pass and eslint's clean.

on the NaN/parseInt point, agreed it's low priority and matches how the existing text-input branch already behaves, so leaving it as is for now unless you'd rather i handle it in this same PR.
@paoloantinori can you please give a look again , i have made the follow up commit

paoloantinori:
Thanks for the follow-up. The `?? ""` change correctly resolves the display-binding concern I raised — `false` / `0` enum selections will now show as selected instead of falling through to the placeholder. The coercion helper and its tests look solid, and I'm fine leaving the NaN/parseInt behaviour consistent with the existing text-input branch as you noted.

LGTM from my side.

akshattalwar001:
hey @paoloantinori, just circling back since the LGTM... anything blocking this from getting merged? or just waiting on time? let me know if you need anything else from me

paoloantinori:
Thanks @akshattalwar001. To your question — nothing blocking from my side; the `values[name] ?? ""` fix addresses the false/0 case correctly and eslint's clean. I'm fine with this. The NaN/parseInt point we agreed is low priority and consistent with the existing text-input branch, so leaving it out of this PR is fine. Merge is a maintainer decision (Eric's `/accept` is noted).

PR #9022, opened 2026-07-29
fix(test): AvroSerdeIT flaky on Kafka container startup timeout

### LINKED ISSUE #8990: fix(test): AvroSerdeIT flaky on Kafka container startup timeout
## Summary

`AvroSerdeIT` in `integration-tests` intermittently fails with `RuntimeException: Timed out while starting Kafka containers` (60s timeout). Observed on main branch run [30303558199](https://github.com/Apicurio/apicurio-registry/actions/runs/30303558199) (Jul 27).

## Root cause

`KafkaFacade` (line 70-76) creates a `StrimziKafkaCluster` without an explicit startup timeout. The default Testcontainers timeout (60s) is tight in CI environments with cold Docker caches or resource contention. The Debezium test infrastructure (`SharedDebeziumInfrastructure`) explicitly sets `withStartupTimeout(Duration.ofSeconds(60))` on its containers, but `KafkaFacade` relies on defaults.

## Suggested fix

Add `.withStartupAttempts(2)` and increase the underlying container startup timeout in `KafkaFacade.start()`. Alternatively, add a retry around `kafkaContainer.start()` to handle transient Docker pull failures.

## Frequency

Low (1 occurrence in last 2 weeks on main). Not blocking but adds noise.

Related: #7661 (test infra consolidation)

krish-kedia:
Fixes: #8990

Description:
The AvroSerdeIT integration test occasionally fails in CI with a RuntimeException: Timed out while starting Kafka containers. This occurs because KafkaFacade instantiates StrimziKafkaCluster using the default Testcontainers configuration, which enforces a strict 60-second startup timeout. In CI environments with cold Docker caches or high resource contention, pulling and booting the Kafka image can exceed this limit, resulting in a flaky test.

Solution:
This PR eliminates the flakiness by introducing a robust retry mechanism directly within KafkaFacade.start(), gracefully handling transient Docker pull delays and Testcontainers timeouts.

Implementation:
- Wrapped the container initialization and kafkaContainer.start() in a 3-attempt for loop.
- Added a catch block that intercepts exceptions/timeouts, forcibly calls .stop() on the partially-booted container to prevent background resource leaks (zombie containers), and clears the reference before the next retry.
- Introduced a 5-second backoff between failed attempts.
- Added safe thread interruption handling during the backoff sleep to respect CI cancellation signals.
- If all 3 attempts fail sequentially, the final exception is properly thrown to legitimately fail the test.

Affected Files: integration-tests/src/test/java/io/apicurio/tests/utils/KafkaFacade.java

### DIFF
--- a/integration-tests/src/test/java/io/apicurio/tests/utils/KafkaFacade.java
+++ b/integration-tests/src/test/java/io/apicurio/tests/utils/KafkaFacade.java
@@ -7,6 +7,7 @@ import org.apache.kafka.clients.admin.NewTopic;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Properties;
@@ -16,6 +17,9 @@ import java.util.Properties;
  */
 public class KafkaFacade implements AutoCloseable {
     static final Logger LOGGER = LoggerFactory.getLogger(KafkaFacade.class);
+
+    private static final int STARTUP_ATTEMPTS = 3;
+    private static final Duration STARTUP_TIMEOUT = Duration.ofMinutes(3);
 
     private AdminClient client;
 
@@ -72,9 +76,11 @@ public class KafkaFacade implements AutoCloseable {
                 .withAdditionalKafkaConfiguration(Map.of(
                         "transaction.state.log.replication.factor", "1",
                         "transaction.state.log.min.isr", "1"))
-                .build();
-        kafkaContainer.start();
-
+                .build()
+                .withStartupAttempts(STARTUP_ATTEMPTS)
+                .withStartupTimeout(STARTUP_TIMEOUT);
+        
+        this.kafkaContainer.start();
     }
 
     public void stopIfPossible() throws Exception {


integration-tests/src/test/java/io/apicurio/tests/utils/KafkaFacade.java
```diff
@@ -67,14 +67,38 @@ public void start() {
...
+        for (int attempt = 1; attempt <= maxAttempts; attempt++) {
+            try {
+                this.kafkaContainer = new StrimziKafkaCluster.StrimziKafkaClusterBuilder()
+                        .withNumberOfBrokers(1)
+                        .withAdditionalKafkaConfiguration(Map.of(
+                                "transaction.state.log.replication.factor", "1",
+                                "transaction.state.log.min.isr", "1"))
+                        .build();
+                kafkaContainer.start();
+                break; // Successfully started
```
paoloantinori: The 60s timeout described in the PR body is Testcontainers' `WaitStrategy` startup timeout, not something `StrimziKafkaCluster` imposes directly. Have you considered using the container's built-in `startupAttempts` (or `.withStartupTimeout(Duration.ofSeconds(180))`) instead of a manual loop? Testcontainers re-runs `start()` on startup-strategy failure via that field, which is the documented path for exactly this flakiness. Not opposed to the loop if there's a reason it's insufficient, but worth a comment here explaining the choice.

integration-tests/src/test/java/io/apicurio/tests/utils/KafkaFacade.java
```diff
@@ -67,14 +67,38 @@ public void start() {
...
+                        LOGGER.warn("Failed to stop kafka container after failed start attempt", stopEx);
+                    }
+                    this.kafkaContainer = null;
+                }
+                if (attempt == maxAttempts) {
+                    throw new RuntimeException("Failed to start Kafka container after " + maxAttempts + " attempts", e);
+                }
+                try {
+                    Thread.sleep(5000); // Wait 5 seconds before retrying
+                } catch (InterruptedException ie) {
```
paoloantinori: 5s is short relative to a cold Strimzi image pull (often 30-90s+), so if the real cause is Docker-pull contention all three attempts may stay inside the same contention window. An exponential backoff, or a longer base, would give attempt 2 a genuine chance to succeed.

integration-tests/src/test/java/io/apicurio/tests/utils/KafkaFacade.java
```diff
@@ -67,14 +67,38 @@ public void start() {
...
+                break; // Successfully started
+            } catch (Exception e) {
+                LOGGER.warn("Failed to start kafka container on attempt {}/{}", attempt, maxAttempts, e);
+                if (this.kafkaContainer != null) {
+                    try {
+                        this.kafkaContainer.stop();
+                    } catch (Exception stopEx) {
+                        LOGGER.warn("Failed to stop kafka container after failed start attempt", stopEx);
+                    }
+                    this.kafkaContainer = null;
```
paoloantinori: On the final attempt, rethrowing `e` directly (rather than wrapping in a `RuntimeException`) preserves the original exception type in test reports — useful when `start()` fails for a non-timeout reason (e.g. port bind). The wrap is only needed to attach the attempt-count context, which could instead be logged at WARN above.

paoloantinori (commented):
The intent is reasonable — a flaky Kafka startup can derail an unrelated IT run — but I think this retry loop is solving the wrong layer and has a couple of correctness edges worth confirming before it lands.

**The retry probably won't do what the description claims.** The PR body attributes the timeout to "the default Testcontainers configuration, which enforces a strict 60-second startup timeout." That 60s is not coming from `StrimziKafkaCluster`'s defaults; it's Testcontainers' built-in startup timeout (`WaitStrategy`-driven, controlled via `withStartupTimeout(Duration)` on the container, or the Ryuk/`startupAttempts` on the `GenericContainer` base). A `try/catch` + `Thread.sleep(5000)` loop around `kafkaContainer.start()` re-pulls/re-boots a *fresh* Strimzi container each time, which works, but it bypasses Testcontainers' own `startupAttempts` mechanism — which exists precisely for this and also re-attempts on `start()` failure. It's worth checking whether simply bumping `startupAttempts` (or the startup timeout) on the builder would achieve the same resilience without a hand-rolled loop. If the loop is kept, a one-line comment on why Testcontainers' built-in retry is insufficient here would head off future confusion.

**The 5s backoff is likely too short to escape the failure mode.** If the cause is genuinely image-pull contention / a cold Docker cache (as the body says), 5s is far less than a cold Strimzi image pull (often 30-90s). All three attempts could plausibly fall within the same contention window and fail for the same reason. An exponential backoff, or a longer fixed delay, would give the second attempt a real chance.

**`Thread.sleep` inside `start()` ignores the Quarkus/Arquillian shutdown path.** The `InterruptedException` handling is correct (re-interrupt + throw), which is good. But note this method is called during test-bootstrap; blocking the test thread for up to ~10s while a CI cancellation arrives will delay the cancel signal. Acceptable for a test util, just flagging it.

**Nits** (non-blocking):
- `5199bba` head: the `RuntimeException` wrap on the final attempt loses the original `RuleViolationException`/test-failure type if `start()` ever throws something more specific than a generic `Exception`. Consider rethrowing the caught exception directly on the last attempt instead of wrapping, so test reports name the real cause.
- The hardcoded `maxAttempts = 3` and `5000ms` would read better as named constants, matching the existing style in this class.

This is test-only code so the blast radius is contained; I'd just want the "why not Testcontainers' built-in retry" question answered before merge.

krish-kedia:
@paoloantinori thank you so much for the detailed review. i have pushed an updated commit that ditches the manual loop entirely and implements your preferred approach by leaning on Testcontainers' native mechanisms. the custom retry loop was indeed trying to reinvent the wheel at the wrong layer.

Native Retry & Timeout: The manual try/catch and Thread.sleep have been completely removed. I’ve instead chained .withStartupAttempts(STARTUP_ATTEMPTS) and .withStartupTimeout(STARTUP_TIMEOUT) directly onto the container.

Cold Image Pulls & Backoff: By bumping the WaitStrategy timeout up to a generous 3 minutes, Testcontainers now natively gives cold Strimzi image pulls (and network contention) plenty of time to succeed without failing prematurely.

Preserving Exception Types: If all 3 native attempts fail, start() will correctly throw its underlying native Testcontainers exception (e.g. ContainerLaunchException), preserving the real cause in the test reports.

Arquillian Shutdown Path: Removing Thread.sleep also resolves the concern about blocking the test bootstrap thread and delaying CI cancellation signals.

Named Constants: Added STARTUP_ATTEMPTS = 3 and STARTUP_TIMEOUT = Duration.ofMinutes(3) at the top of the class to match the existing style.


waiting for your review.

krish-kedia:
also, i implemented a manual loop initially mainly because StrimziKafkaCluster is a custom extension of GenericContainer provided by the io.strimzi.test.container library. Often, third-party wrapper containers override the native start() method and can sometimes interfere with or bypass Testcontainers' native retry semantics. I wrongly assumed that manually wrapping the entire builder and instantiation process in a try/catch loop was the safest way to guarantee a completely fresh, untainted container instance on every attempt, completely oblivious to the fact that Testcontainers' native startupAttempts is explicitly designed to handle container destruction and recreation under the hood anyway.

paoloantinori:
Thanks @krish-kedia — the native Testcontainers approach is much cleaner. Chaining `.withStartupAttempts(STARTUP_ATTEMPTS)` and `.withStartupTimeout(STARTUP_TIMEOUT)` directly on the builder lets Testcontainers own the retry/stop/re-create lifecycle (including stopping the partially-booted container between attempts), which is exactly the layer this belongs at rather than a hand-rolled loop in `KafkaFacade`. 3 minutes × 3 attempts is a reasonable budget for a cold Docker pull in CI. Looks good to me.

krish-kedia:
@paoloantinori thankyou for the validating the implementation, ig this PR needs to be accepted to fix the failing ci check.

paoloantinori:
Thanks @krish-kedia — and good that `startupAttempts` now handles the retry/recreation. Your reasoning about the Strimzi wrapper overriding `start()` was a fair concern, but Testcontainers' native path is the right one to rely on. Lgtm on the implementation; merge is a maintainer call (Eric's `/accept` is noted).

krish-kedia:
@EricWittmann PTAL

PR #9023, opened 2026-07-29
refactor: resolve compilation issues and replace raw JSON strings with typed ContractRule builders

### LINKED ISSUE #8863: Refactor DataContractsResourceTest to use type-safe DTOs instead of raw JSON strings

hussainjamal760:
### Description

This PR fixes the compilation errors and Checkstyle issues in `DataContractsResourceTest`, and also applies a substantial refactor to the test file to modernize payload construction and ensure CI passes correctly.

Closes: #8863 

### Changes

* **Removed Ambiguous Imports**: Removed the wildcard imports for `ContractRule.OnFailure` / `ContractRule.Kind` and replaced ambiguous references with fully qualified enum values.
* **Builder Refactoring**: Completely refactored test payloads to replace fragile raw JSON string literals (including triple-escaped JSONATA expressions) and old private helper methods (that used stringly-typed lookups) with the typed `ContractRuleSet.builder()` and `ContractRule.builder()`. This restores compile-time field checking and improves robustness.
* **License Compliance**: Added the missing Apache 2.0 license header to `DataContractsResourceTest.java` which was previously failing the CI License Compliance check.
* **CI Verification Gate Fix**: Fixed the verification gate logic in `.github/workflows/verify.yaml` to ensure it successfully passes rather than failing when all upstream decide jobs are intentionally skipped.
* **Checkstyle Fixes**: Cleaned up unused imports to resolve the remaining Checkstyle warnings.

### Why

While building the project, the test module failed due to ambiguous enum references and a type mismatch in the builder. The initial fix exposed that the tests were a mix of untyped JSON and builders, so a larger refactor was necessary to standardize on the typed model and eliminate stringly-typed fragility. Furthermore, the CI pipelines required minor adjustments to the verification gate and license headers to pass properly for this PR.

### Testing

I verified the changes locally using the following commands:

```bash
./mvnw test -pl app -Dtest=DataContractsResourceTest

### CODE REVIEW

File: app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java [hunk 1]
```diff
@@ -631,13 +709,16 @@ public void testGlobalContractRules_CrudLifecycle() throws Exception {
...
-        globalRule.setExpr("true");
-        globalRule.setOnFailure(ContractRule.OnFailure.fromValue("ERROR"));
-        globalRule.setDisabled(false);
-
         given().when().contentType(CT_JSON)
-                .body(ruleset(List.of(globalRule), List.of()))
+                .body("""
+                        {
+                            "domainRules": [
+                                {"name":"global-rule","kind":"CONDITION","type":"CEL",
```
paoloantinori: This test (and several others: `testExecuteContractRules_*`, `testMigrateEndpoint_*`, `testAuditLog_*`, `testGlobalRules_*`) regressed from the deleted typed `rule()`/`ruleset()` helpers to hand-written JSON string literals. Earlier tests in this same PR switched to `ContractRule.builder()...`, so the file now mixes typed builders and untyped JSON. The untyped JSON loses compile-time field-name checking and reintroduces the brittleness the typed model exists for. Consider converting these to builders for consistency.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
```diff
@@ -743,14 +822,12 @@ public void testMigrateEndpoint_WithTransformRule() throws Exception {
...
-
         given().when().contentType(CT_JSON)
                 .pathParam("groupId", GROUP).pathParam("artifactId", artifactId)
-                .body(ruleset(List.of(), List.of(transformRule)))
+                .body("{\"domainRules\":[],\"migrationRules\":["
+                        + "{\"name\":\"add-y\",\"kind\":\"TRANSFORM\",\"type\":\"JSONATA\","
+                        + "\"mode\":\"UPGRADE\",\"expr\":\"$ ~> |$|{\\\"y\\\": 99}|\","
+                        + "\"onFailure\":\"ERROR\",\"disabled\":false}]}")
                 .put("/registry/v3/groups/{groupId}/artifacts/{artifactId}/contract/ruleset")
                 .then().statusCode(200);
```
paoloantinori, line 869: The triple-escaped `\"y\": 99` inside a string-concat JSON payload is correct but fragile to edit. The typed builder used in `testSetAndGetArtifactContractRuleset` a few tests up handles the JSONATA expr quoting transparently — prefer that here.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
```diff
@@ -27,41 +53,6 @@ public class DataContractsResourceTest extends AbstractResourceTestBase {
...
-        return rule;
-    }
-
-    private static Params params(String key, String value) {
-        Params params = new Params();
-        params.setAdditionalProperty(key, value);
-        return params;
-    }
-
     // -- Contract Metadata Tests --
```
paoloantinori, line 61: PR description says "fix compilation errors / Checkstyle" but the diff deletes five helper methods, rewrites ~every test body, and adds a license header. That's a refactor, not just a compile fix. The ambiguity could have been resolved by switching the helpers' `fromValue("WRITE")` calls to enum constants in place. The wholesale inlining is fine if intended, but worth aligning the PR description with the actual scope.

paoloantinori (commented):
The mechanical fix (replacing ambiguous `ContractRule.OnFailure` wildcard imports with FQN enum values, and switching `.params()` to a `Params` object) is correct and necessary for the compile. But the PR goes well beyond the stated scope in #8863, and a couple of the expansions look like regressions rather than cleanups — worth a second look before merge.

**The description understates the scope.** The PR body says it "removes wildcard imports and replaces ambiguous enum references." In practice it (a) adds a Red Hat license header, (b) deletes five private helper methods (`contractMetadata`, `statusTransition`, `ruleset`, `rule`, `params`), and (c) rewrites ~every test body to inline builders or raw JSON strings. That's a substantial refactor of the test file, not just a compile fix. If the helpers were the cause of the ambiguity (they used stringly-typed `fromValue("WRITE")` lookups), they could have been fixed in place by switching to the enum constants — the wholesale deletion/inlining is a separate stylistic change layered on top. Not necessarily wrong, but the gap between "fix compilation errors" and what landed is worth being honest about in the PR description.

**Several tests regress from typed builders to hand-written JSON string literals.** `testGlobalContractRules_CrudLifecycle`, `testExecuteContractRules_*`, `testMigrateEndpoint_WithTransformRule`, `testAuditLog_*`, `testSearchContracts_*`, `testGlobalRules_*`, and `testContractRulesFailOnErrorFalse_*` all switched from the (now-deleted) `rule()`/`ruleset()` helpers to inline `body("{\\"domainRules\":[...]}")` strings. The earlier part of the PR established typed builders (`ContractRule.builder()...`) as the replacement — these later tests instead use untyped JSON, which is exactly the kind of stringly-typed payload the compile error was about. This loses compile-time field-name checking and reintroduces the brittleness the typed model exists to prevent. If the builders were the intended direction, these tests should use them consistently; the JSON-string tests read like the refactor ran out of steam halfway through. Worth confirming the inconsistency is intentional.

**`testMigrateEndpoint_WithTransformRule` has a hand-escaped JSON string that's now harder to audit.** The body builds the migration rule via string concatenation with triple-escaped `\\"y\\": 99`. Correct but fragile — a future edit to the JSONATA expr will need to re-count backslashes. The typed builder used a few tests up handles this transparently.

**`testGlobalContractRules_CrudLifecycle` uses a text block for JSON.** Fine, but text blocks aren't used elsewhere in the file, so this introduces a third style (typed builder, string-concat JSON, text-block JSON) in one file. Pick one for the JSON-string cases at least.

**`Params` instantiation is inconsistent.** `testSetAndGetArtifactContractRuleset` builds `Params` via `new Params(); setAdditionalProperty(...)` inline before the `given()`, while every other `Params` usage in the file either uses `.params()` on the builder or was deleted. Make sure `setAdditionalProperty` is still the right mutator on the current model (it is, per the import), but the inconsistency suggests the refactor wasn't applied uniformly.

**Confirmed correct:** the FQN enum references (`ContractRule.OnFailure.ERROR`, `.Kind.CONDITION`, etc.) resolve the original ambiguity cleanly, the `static import` block at the top is tidy, and removing the stringly-typed `fromValue("WRITE")` helpers is a genuine improvement *where builders replaced them*. The problem is the half that went helpers→raw-JSON rather than helpers→builders.

Net: the compile fix is good; I'd ask for the raw-JSON test bodies to be converted to `ContractRule.builder()` for consistency before merge, since they reintroduce exactly the stringly-typed fragility the typed model is meant to prevent.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
hussainjamal760: Done! I have converted all the remaining test methods to use the typed ContractRuleSet.builder() and ContractRule.builder() instead of raw JSON string literals. This ensures consistency and restores compile-time field checking across the entire file.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
hussainjamal760: Good catch! I have replaced this manual JSON concatenation with the typed builder. The JSONATA expression is now passed as a natural string without the fragile manual escaping.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
hussainjamal760: The wholesale inlining to use the builders directly was intentional to move away from stringly-typed lookups. I will update the PR description to accurately reflect the actual scope of this change (which includes replacing the deprecated helpers with typed builders and adding missing license headers) rather than just fixing compilation errors.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
```diff
@@ -791,13 +901,13 @@ public void testCompatibilityGroup_SetDifferentGroups() throws Exception {
 
         given().when().contentType(CT_JSON)
                 .pathParam("groupId", GROUP).pathParam("artifactId", art1)
-                .body(contractMetadata("DRAFT"))
+                .body("{\"status\":\"DRAFT\"}")
```
EricWittmann: There are still a number of these hard-coded strings.  For consistency I guess we should switch this to using the `EditableContractMetadata` builder as well.

EricWittmann:
Review comment added!  Just a couple of tweaks and then we can merge.  :+1:

hussainjamal760:
> Review comment added! Just a couple of tweaks and then we can merge. 👍

Sounds good! 👍 I’ll work on the requested changes right away and update you once they’re done. Thanks for the review!

hussainjamal760:
@EricWittmann Done , /ready

EricWittmann:
Thanks for the work @hussainjamal760! Switching from raw JSON strings to typed builders makes these tests much more maintainable. Appreciate you iterating on the feedback to get everything converted consistently.

hussainjamal760:
> Thanks for the work @hussainjamal760! Switching from raw JSON strings to typed builders makes these tests much more maintainable. Appreciate you iterating on the feedback to get everything converted consistently.

Thank you, @EricWittmann ! I really appreciate the detailed feedback and guidance throughout the process. Glad we were able to get everything converted consistently and make the tests more maintainable. Learned a lot while working on this! 🙌

PR #9025, opened 2026-07-29
feat(ci): simplify PR lifecycle to 3 states, remove wip/smoke stage

### LINKED ISSUE #9002: Simplify PR lifecycle: remove wip/smoke stage, shrink to 3 states
Part of #9001 (PR Lifecycle v2 epic)

## Goal

Reduce the contributor-facing lifecycle from 3 commands to 2 by removing the `wip`/smoke test stage.

**Current flow:**
```
new --/accept--> wip --/ready--> ready-for-review --[approved+tested]--> ready-to-merge --/merge--> MERGED
```

**New flow:**
```
new --/accept--> ready-for-review --[approved+tested]--> ready-to-merge --/merge--> MERGED
```

## Changes

- `/accept` transitions directly to `ready-for-review` and triggers the full test suite
- Draft PRs: orchestrator ignores them entirely (no `/accept`, no CI). When marked non-draft, enters `lifecycle/new` like a fresh PR. Authors test locally during development.
- Fully remove: `lifecycle/wip`, `lifecycle/smoke-tested`, `/ready`, `/disable-tests`, `/enable-tests`, and all smoke test infrastructure in `verify.yaml`
- State machine shrinks to 3 states: `new`, `ready-for-review`, `ready-to-merge`
- Update welcome message and `PR_LIFECYCLE.md` documentation
- Shorten stale timeout for `waiting-on-author` PRs from 14 days to 7 days. These PRs are explicitly blocked on contributor action and reviewers have already invested time. Data shows 5 PRs accumulated 26-41 comments before the contributor stopped responding — faster close reduces wasted attention.

## Justification (data)

- Out of 30 recent merged PRs, 2 were merged from `lifecycle/new` (never entered `wip`), 2 were merged from `lifecycle/wip` (never reached `ready-for-review`) — the `wip` stage is routinely bypassed
- Smoke tests (~10 min) followed by full suite (~40 min) costs more total runner time than running full suite once
- The `/ready` command is the step contributors miss most, leading to PRs stuck in `wip` without the full suite ever running
- Contributor funnel: only 22% of PRs reach `tested`, and the biggest friction is process discipline, not technical quality

## Files

- `.github/scripts/pr-lifecycle.js` — remove WIP state transitions, `/ready` handler, smoke test logic; modify `/accept` to go directly to `ready-for-review`
- `.github/pr-lifecycle.yml` — update welcome message, shorten `waiting-on-author` stale timeout
- `.github/PR_LIFECYCLE.md` — update documentation
- `.github/workflows/pr-lifecycle.yml` — simplify event handling
- `.github/workflows/verify.yaml` — update `determine-scope` job

**Blocks:** #9003, #9004, #9005

malladinagarjuna2:
Closes #9002

Removes the `lifecycle/wip`/smoke-test stage. `/accept` now transitions a PR directly to `ready-for-review` and runs the full test suite. Draft PRs are ignored by the orchestrator until marked ready for review.

## Changes
- Drop `lifecycle/wip`, `lifecycle/smoke-tested`, `orchestrator/tests-disabled` and the `/ready`, `/disable-tests`, `/enable-tests` commands
- `verify.yaml` scope decision collapses to a single full-suite gate
- Waiting-on-author PRs now close after 7 days total (was 14)
- Docs updated for the 3-state flow

## Test plan
- [x] Verified locally with a mocked GitHub API harness
- [ ] Manual check on a real PR after merge

## Follow-up
- **Stale warnings should be reviewer-aware**: when a PR has `waiting-on-maintainer` and an assigned reviewer, the stale warning should mention the reviewer. The author did their part — the bottleneck is the reviewer, and the current generic stale message doesn't reflect that. Options: (a) include the assigned reviewer in the stale message text, (b) extend or skip the stale timer entirely for `waiting-on-maintainer` PRs.


.github/scripts/pr-lifecycle.js
```diff
@@ -565,6 +543,18 @@ async function handlePrOpened({ github, context, core }) {
   core.info(`PR #${pr.number} opened, set to lifecycle/new`);
 }
 
+async function handlePrOpened({ github, context, core }) {
+  const pr = context.payload.pull_request;
+  if (pr.draft) {
+    core.info(`PR #${pr.number} opened as draft, orchestrator ignoring until marked ready for review`);
+    return;
+  }
```
paoloantinori, line 599: Dropping the requested-reviewers check here (vs. the old `handlePrReadyForReview`) means an auto-accepted draft→ready PR can reach `ready-for-review` + full CI with no reviewer assigned. If forcing a reviewer before the full suite was intentional policy, it's now gone for the auto-accept path. Confirm that's acceptable, or re-add the gate in `initNewPr` for the auto-accepted branch.

.github/scripts/pr-lifecycle.js
```diff
@@ -718,15 +685,15 @@ async function cmdAccept(api, config, core, pr, actor, maintainer, commentId) {
...
   await api.addReaction(commentId, '+1');
   await api.postComment(pr.number,
-    `PR accepted by @${actor}. @${pr.user.login}, you can now work on your changes.\n\n` +
-    `Smoke tests will run on each push. When ready, use \`/ready\` to request a full review.`
+    `PR accepted by @${actor}. @${pr.user.login}, the full test suite will run now.\n\n` +
+    `A maintainer can use \`/auto-merge\` to merge automatically once approved and tested.`
   );
   core.info(`PR #${pr.number} accepted by ${actor}`);
-  await retriggerVerify(api, pr, core);
+  await retriggerVerify(api, pr, core, { waitForRun: true });
```
paoloantinori: `/accept` with `waitForRun: true` now triggers the *full* suite (integration/SDK/operator) on every accepted PR, not just smoke. Previously these only ran at `/ready`. For high-churn or docs-only PRs this is a notable CI-minutes increase — the scalpel only skips affected modules, build/SDK/operator look unconditional on `run_tests`. Worth a maintainer ack on the cost delta.

.github/scripts/pr-lifecycle.js
```diff
@@ -1449,10 +1332,14 @@ async function handleStale({ github, context, core }) {
                e.event === 'head_ref_force_pushed';
       });
 
+      const effectiveDaysUntilClose = hasLabel(pr, LABELS.WAITING_ON_AUTHOR)
+        ? daysUntilCloseWaitingOnAuthor
+        : daysUntilClose;
+
```
paoloantinori: For waiting-on-author PRs, `effectiveDaysUntilClose - daysUntilStale` = 0, so the close fires on the first stale sweep — effectively no grace period after the stale ping. The stale message ("closed if no activity occurs") doesn't convey the urgency. Consider making the message window-aware so an author isn't surprised by a same-week close.

.github/scripts/pr-lifecycle.js
```diff
@@ -433,20 +426,17 @@ async function reconcile(github, api, pr, core) {
   // 1. No lifecycle label at all → initialize as new PR
   if (!state) {
     if (hasLabel(pr, LABELS.DISABLED)) return;
+    // The orchestrator ignores draft PRs entirely — nothing to reconcile.
```
paoloantinori: Existing PRs still labelled `lifecycle/wip` / `lifecycle/smoke-tested` / `orchestrator/tests-disabled` aren't migrated by this change — they'll be stuck in a state no longer in `PRIMARY_STATES` that nothing transitions out of. A reconciler branch that maps legacy → `ready-for-review` (or a one-time label sweep) would prevent stranding in-flight PRs on merge.

paoloantinori (commented):
I like the direction — collapsing the WIP/smoke stage is a real simplification and the draft-PR-as-the-new-WIP mapping is clean. A few behaviour-change risks worth confirming before merge, since this rewrites the contract every contributor interacts with.

**The draft-PR path loses the `requested_reviewers` gate.** Previously, `handlePrReadyForReview` (draft→ready) refused the transition with a comment if no reviewer was assigned ("Cannot transition to ready-for-review: no reviewer is assigned"). The new `initNewPr` path called from `handlePrReadyForReview` has no such check — a draft PR marked ready enters `lifecycle/new` and, if the author is auto-accepted, jumps straight to `ready-for-review` + full suite with no reviewer assigned. If that gate was intentional policy (force a reviewer before burning CI minutes on the full suite), it's now gone for the auto-accept path. Worth a deliberate decision: is dropping it fine because `/skip-review` and manual triage cover it, or should `initNewPr` enforce it for the auto-accepted branch too?

**`/accept` now triggers the *full* suite synchronously (`waitForRun: true`) for every accepted PR.** Under the old model, `/accept` ran only smoke tests; the expensive integration/SDK/operator suites ran only at `/ready`. Now every accepted external contribution burns the full suite on the first push, and again on every subsequent push (since `ready-for-review` is the only test-running state and pushes keep that label). For a repo where the full suite is long, this is a material CI-cost increase for high-churn PRs that previously lived in WIP. The author mentions CI simplification as the goal; worth confirming the maintainers are OK with the minutes delta, especially for docs-only or one-line PRs that used to sit in WIP and now immediately run the full matrix (the scalpel in `verify.yaml` mitigates this, but only for affected-module detection, not the build/SDK/operator jobs which look unconditional on `run_tests`).

**Stale-close window: the asymmetry is sensible but the docs undersell it.** `PR_LIFECYCLE.md` says waiting-on-author PRs "are closed after 7 total days" while others take 14. The `handleStale` math uses `daysSinceStale >= (effectiveDaysUntilClose - daysUntilStale)`, i.e. 0 extra days for waiting-on-author (so the close fires on the *first* stale sweep, ~7 days after last activity) vs 7 extra days otherwise. That's coherent, but a contributor whose PR goes stale while waiting-on-author will see it closed with effectively zero grace period after the stale ping (the stale message says "closed if no activity occurs" with no number). Consider making the stale message aware of the shortened window, or the close message explicit, so an author paged on a Friday doesn't lose the PR on Monday unaware.

**`cmdSkipReview` no longer accepts WIP**, which is correct given WIP is gone — but double-check that no in-flight PR is sitting in `lifecycle/wip` when this merges. The reconciler's `getLifecycleState` would return a state no longer in `PRIMARY_STATES`, and the new label-cleanup doesn't proactively strip stale `lifecycle/wip` / `lifecycle/smoke-tested` / `orchestrator/tests-disabled` labels from existing PRs. A one-time migration (or a reconciler branch that maps legacy labels → `ready-for-review`) would prevent existing WIP PRs from getting stuck in a state nothing transitions out of. Worth grepping open PRs for those labels on merge.

**Minor:** `verify.yaml`'s `decide run-cli "$run_tests" "$CI"` dropped the `$IS_PUSH` term that the other lines kept — intentional (CLI runs on every test-scoped PR, not just push)? If so, fine; if it was a mechanical collapse from the smoke/full split, the other jobs kept `$IS_PUSH` for a reason.

Net: the simplification is sound, I'd just want the reviewer gate, the full-suite-on-accept cost, and the legacy-label cleanup explicitly signed off.

.github/scripts/pr-lifecycle.js
malladinagarjuna2: yaa got it!

.github/scripts/pr-lifecycle.js
malladinagarjuna2: fair point! was my mistake should have reviewed more carefully

.github/scripts/pr-lifecycle.js
malladinagarjuna2: addresed in the curent commit the stale message is now window-aware via '{clse_window}', it still renders "as soon as the next check" for waiting-on-author PRs thoough, so I'd also warn earlier ('days_until_stale:'4') to give a real 3 day window inside the same 7day total.

.github/scripts/pr-lifecycle.js
malladinagarjuna2: one gap i want to fix on top of that, the migration runs before the  pr draft guard and draft + auto accepted used to be lifecycle/wip , so existing draft prs would get promoted and run a full suite, which contradicts the new draft prs ignored rule. will push a fix so drafts just have  legacy label stripped without promotion.

paoloantinori:
Good catch on the ordering bug — you're right that `migrateLegacyLabels` runs at the top of `reconcile` before the `pr.draft` short-circuit, so an existing draft PR carrying a stale `lifecycle/wip` would get promoted to `lifecycle/ready-for-review` and trigger a full suite run, which directly contradicts the new "drafts are ignored" rule.

Your proposed fix (let drafts keep just the legacy label rather than migrating them into the active lifecycle) is the right shape: it preserves idempotency without pulling drafts into the test-gated path. Two things worth confirming when you push:

1. **Idempotency on the draft→ready transition.** Once the draft is eventually marked ready, `handlePrReadyForReview` calls `initNewPr`, which will re-evaluate from scratch — so make sure the leftover legacy label there doesn't cause a double-label or a confusing reconciler warning.
2. **The non-draft migration path stays as-is.** Only the draft branch should short-circuit; non-draft PRs with stale `lifecycle/wip` still want the full promotion + `retriggerVerify`.

Happy to take another look once the fix is up.

malladinagarjuna2:
@paoloantinori  thanks for the detailed review!  a lot of learnings from this pr undestannding different scenarios. 
some more few suggestions:
1.for waiting on author prs gets literally zero grace period
The issue asks for a 7-day total timeout, which is fine, but i think adding a  warning window inside those 7 days would be helpful. Either add days_until_stale_waiting_on_author: 4 (warn day 4, close day 7), or set the close at 10. Right now a contributor who's away for a week comes back to a closed PR with a warning they had 6 hours to act on. Given the issue's own justification is "reviewers already invested time," closing without a real ping wastes that investment rather than saving it.

2. migrateLegacyLabels runs before the pr.draft guard, and draft + auto-accepted used
to *be* lifecycle/wip,so every existing draft PR gets promoted to
ready-for-review and fires a full suite, contradicting the "drafts are ignored" rule
this PR introduces. Bail out for drafts (strip the legacy label without promoting), and
use setLifecycleState instead of addLabel so a PR carrying both lifecycle/new and
lifecycle/wip doesn't end up with two primary states.

3.PR_LIFECYCLE.md says drafts get no labels and no CI, but nothing enforces that once a
PR is converted back to draft: converted_to_draft isn't in the trigger types
(.github/workflows/pr-lifecycle.yml:5), verify.yaml's decide step never checks
github.event.pull_request.draft, and cmdAccept doesn't check it either, so a
converted PR keeps lifecycle/ready-for-review and runs the full suite on every push.
That bites harder now that wip and /disable-tests are gone, leaving "convert to
draft" as the only intuitive way to throttle CI while it silently does nothing.

Cheapest fix is a draft guard in the decide step (run_tests=false when the PR is a
draft), which closes both paths at once; handling converted_to_draft in the
orchestrator to drop the lifecycle labels is the more complete version.

malladinagarjuna2:
@paoloantinori  did the changes 
>Idempotency on the draft→ready transition. Once the draft is eventually marked ready, handlePrReadyForReview calls initNewPr, which will re-evaluate from scratch — so make sure the leftover legacy label there doesn't cause a double-label or a confusing reconciler warning.
The non-draft migration path stays as-is. Only the draft branch should short-circuit; non-draft PRs with stale lifecycle/wip still want the full promotion + retriggerVerify.
Happy to take another look once the fix is up.

paoloantinori:
Thanks @malladinagarjuna2 — read the latest update. The idempotency point is addressed correctly: `handlePrReadyForReview` now calls `initNewPr` only when there's no existing lifecycle label, so a draft PR that carried a stale `lifecycle/wip` and is later marked ready won't double-label or trigger a confusing reconciler warning. And `migrateLegacyLabels` correctly drops the legacy label for drafts without promoting, leaving `initNewPr` to set `lifecycle/new` cleanly on the draft→ready transition. The non-draft legacy path keeps the full promotion + `retriggerVerify`, as intended. Looks good from my side.

malladinagarjuna2:
@paoloantinori  
could you please tell about the suggestions  which i mentioned

> @paoloantinori thanks for the detailed review! a lot of learnings from this pr undestannding different scenarios. some more few suggestions: 1.for waiting on author prs gets literally zero grace period The issue asks for a 7-day total timeout, which is fine, but i think adding a warning window inside those 7 days would be helpful. Either add days_until_stale_waiting_on_author: 4 (warn day 4, close day 7), or set the close at 10. Right now a contributor who's away for a week comes back to a closed PR with a warning they had 6 hours to act on. Given the issue's own justification is "reviewers already invested time," closing without a real ping wastes that investment rather than saving it.
> 
> 2. migrateLegacyLabels runs before the pr.draft guard, and draft + auto-accepted used
>    to _be_ lifecycle/wip,so every existing draft PR gets promoted to
>    ready-for-review and fires a full suite, contradicting the "drafts are ignored" rule
>    this PR introduces. Bail out for drafts (strip the legacy label without promoting), and
>    use setLifecycleState instead of addLabel so a PR carrying both lifecycle/new and
>    lifecycle/wip doesn't end up with two primary states.
> 
> 3.PR_LIFECYCLE.md says drafts get no labels and no CI, but nothing enforces that once a PR is converted back to draft: converted_to_draft isn't in the trigger types (.github/workflows/pr-lifecycle.yml:5), verify.yaml's decide step never checks github.event.pull_request.draft, and cmdAccept doesn't check it either, so a converted PR keeps lifecycle/ready-for-review and runs the full suite on every push. That bites harder now that wip and /disable-tests are gone, leaving "convert to draft" as the only intuitive way to throttle CI while it silently does nothing.
> 
> Cheapest fix is a draft guard in the decide step (run_tests=false when the PR is a draft), which closes both paths at once; handling converted_to_draft in the orchestrator to drop the lifecycle labels is the more complete version.

paoloantinori:
You're right, and the code confirms exactly what you suspected: waiting-on-author PRs get a near-zero grace window with the current config. Let me trace the threshold math so it's explicit, then recommend your suggestion #1.

**The bug.** The close check is:

```js
else if (daysSinceStale >= (effectiveDaysUntilClose - daysUntilStale)) { /* close */ }
```

With `days_until_close_waiting_on_author = 7` and `days_until_stale = 7`, the right-hand side is `7 - 7 = 0`. So a waiting-on-author PR:
- Day 7 cron: no STALE label yet → `daysSinceUpdate >= 7` → STALE added (close branch not reached, PR didn't have STALE at the start of this run).
- Next cron run: now has STALE, `daysSinceStale ≈ 0` → `0 >= 0` → **closed immediately**.

The stale warning and the close fire on consecutive ticks, and the `closeWindow` formatter honestly renders this as "as soon as the next check" — which is technically correct but defeats the purpose of a warning. The author never gets a real window to respond after being pinged. Your instinct to flag this is spot on.

**Your suggestion #1 is the right fix.** Add a separate `days_until_stale_waiting_on_author` (default 4) so those PRs get an earlier warning with real grace, while keeping the 7-day total close the issue asked for:

```yaml
stale:
  days_until_stale: 7
  days_until_close: 14
  days_until_stale_waiting_on_author: 4      # new
  days_until_close_waiting_on_author: 7
```

Then in `handleStale`, the stale threshold also needs to branch on `WAITING_ON_AUTHOR`:

```js
const effectiveDaysUntilStale = hasLabel(pr, LABELS.WAITING_ON_AUTHOR)
  ? (config.stale?.days_until_stale_waiting_on_author || 4)
  : daysUntilStale;
// ... use effectiveDaysUntilStale in the daysSinceUpdate >= X check
```

That gives: warn day 4, close day 7 → 3 days of real grace. The close-threshold math then becomes `daysSinceStale >= (7 - 4) = 3`, so close fires once `daysSinceStale >= 3` — exactly the 3-day post-warning window. This mirrors the regular path's shape (stale at 7, close at 14 = 7 days grace), which is the consistency worth preserving.

**Suggestion #2 (bump `days_until_close_waiting_on_author` above 7)** I'd avoid — it violates the issue's explicit 7-day total requirement for waiting-on-author PRs. The point of the shorter window is that those PRs are already stalled on contributor action, so the total timeout should stay at 7; we just want to *warn earlier within* that 7, not extend it.

One implementation note on suggestion #1: the `daysSinceStale`/`daysSinceUpdate` recompute after toggling the STALE label mid-run needs care — make sure a PR that crosses the *earlier* waiting-on-author stale threshold (day 4) doesn't also immediately satisfy the close check in the same pass. The current structure (close branch is only entered when STALE is already present at the top of the iteration) already prevents this, so as long as you keep the same if/else shape it's safe. Worth a comment in the code calling that out, since it's a subtle invariant.

So: go with suggestion #1. Want me to sketch the full `handleStale` diff, or are you good to take it from here?

.github/scripts/pr-lifecycle.js
```diff
@@ -718,12 +723,12 @@ async function cmdAccept(api, config, core, pr, actor, maintainer, commentId) {
     return;
   }
 
-  await api.setLifecycleState(pr, LABELS.WIP);
+  await api.setLifecycleState(pr, LABELS.READY_FOR_REVIEW);
```
paoloantinori, line 769: `/accept` now sets `lifecycle/ready-for-review` (was `lifecycle/wip`) but the next line still removes `lifecycle/waiting-on-maintainer`. That removal made sense under the old WIP target (author was working), but an accepted PR now sits in `ready-for-review` waiting on a reviewer, so it arguably needs the maintainer-attention label. The previous re-adder was the deleted `/ready` command (`cmdReady`), and every other ready-for-review path in this PR — `initNewPr`, `reconcile`, `migrateLegacyLabels` — adds `waiting-on-maintainer`; the `reconcile` adder shown only fires in the no-label branch. If the part of `reconcile` not in this diff doesn't re-derive it for already-labeled PRs, accepted PRs won't surface in the review queue. Suggested fix: `addLabel(WAITING_ON_MAINTAINER)` here instead of removing it, or confirm reconcile covers it.

paoloantinori (commented):
The 3-state simplification is coherent and the legacy-label migration is handled thoughtfully (one-shot, draft-aware, self-healing). One gap to verify before merge: `/accept` carries over WIP-era `waiting-on-maintainer` handling that no longer fits a transition that now lands directly in `ready-for-review`.

- `.github/scripts/pr-lifecycle.js:726` — `cmdAccept` transitions to `lifecycle/ready-for-review` but still removes `lifecycle/waiting-on-maintainer`. In the prior flow that label was re-added by the now-deleted `/ready` (`cmdReady`) and the draft→ready path; here every other ready-for-review entry point (`initNewPr`, `reconcile`'s no-label branch, `migrateLegacyLabels`) explicitly *adds* it, while the only `reconcile` adder visible in the diff fires when a PR has no lifecycle label at all — not for an already-labeled ready-for-review PR. If the elided `reconcile` body doesn't re-derive it, accepted PRs will silently drop out of the maintainer review queue. Either add `waiting-on-maintainer` here (matching the other paths) or confirm reconcile sets it.

.github/scripts/pr-lifecycle.js
malladinagarjuna2: yeah thts true!!

malladinagarjuna2:
@paoloantinori  addressed the small changes kindly review it

paoloantinori:
Thanks @malladinagarjuna2 — the waiting-on-author stale handling is implemented correctly. The snapshot-based `isWaitingOnAuthor` check, the `effective*` threshold swap, and the `{close_window}` placeholder all line up, and the `Math.max(0, …)` guard handles the degenerate case cleanly.

One accuracy nit on the justification comment in `pr-lifecycle.yml`:

> `# 4/7 keeps the same 3-day post-warning grace as the 7/14 default pair.`

The grace window is `close − stale`, so 7/14 gives **7** days of post-warning grace, not 3 — 4/7 (→ 3 days) is *shorter* than the default, not the same. The behavior is fine (a tighter grace for PRs explicitly stalled on the author is reasonable, maybe even desirable), but the comment sells it as parity that doesn't exist. Suggest rewording to something like "shorter post-warning grace than the default, since the PR is explicitly stalled on contributor action" so a future maintainer doesn't try to "fix" the numbers expecting 7/14 parity.

Minor, non-blocking — the logic itself is sound.

malladinagarjuna2:
hey @paoloantinori  really thanks for reviewing again, the quickness at which  replies are given  for the pr changes is far far less  than  other open source orgs. Really you are doing a great job!

will do the changes.

malladinagarjuna2:
done with the changes 
@paoloantinori 

calling @jsenko @EricWittmann  for further  review on the pr

paoloantinori:
Agreed — the changes through 4102888 look good on my end. The legacy-label migration, draft handling, and waiting-on-author stale thresholds all hold up to the trace we did together.

Deferring the final merge call to @jsenko / @EricWittmann as you asked.

malladinagarjuna2:
@jsenko could you please review it

carlesarnal:
## Code Review

Thanks for the thorough work here @malladinagarjuna2. The lifecycle simplification is well-structured and @paoloantinori's iterative review caught the important edge cases. The core logic looks solid. A couple of items to address before merge:

---

### 1. Commit messages need cleanup (blocking)

The repo uses rebase merge strategy (`merge.strategy: rebase` in `pr-lifecycle.yml`), so each commit lands individually on `main`. Four of the eight non-merge commits don't follow conventional commit format:

| Commit | Message | Issue |
|--------|---------|-------|
| `990972b` | `address review feedback o pr lifecycle simpleification` | No type prefix, typo ("simpleification") |
| `332cc5d` | `draft changes addressed` | No type prefix |
| `53ae26b` | `added warn days` | No type prefix |
| `c6c1825` | ` restore waiting-on-maintainer label after /accept` | Leading space, no type prefix |

Please squash these into the relevant feature/fix commits or reword with proper `feat(ci):`/`fix(ci):` prefixes via interactive rebase. For example, these could all fold into the initial `feat(ci): simplify PR lifecycle...` commit.

---

### 2. `converted_to_draft` gap (non-blocking, maintainer decision)

You flagged this yourself in [suggestion #3](https://github.com/Apicurio/apicurio-registry/pull/9025#issuecomment-5134119088) and it's still valid: `PR_LIFECYCLE.md` says drafts get no labels and no CI, but nothing enforces that when a PR is **converted back** to draft:

- `pr-lifecycle.yml` doesn't trigger on `converted_to_draft`
- `verify.yaml`'s decide step doesn't check `github.event.pull_request.draft`

A PR that goes ready-for-review → draft would keep its lifecycle labels and continue running the full suite on every push. With `/disable-tests` gone, "convert to draft" is the only intuitive way for a contributor to throttle CI — but it silently does nothing. Could be addressed here or tracked as a follow-up issue — deferring to @jsenko / @EricWittmann on scope.

---

### Everything else looks good

- ✅ DCO sign-offs on all non-merge commits
- ✅ `migrateLegacyLabels` correctly handles drafts (strip without promoting) vs non-drafts (promote + retrigger)
- ✅ `cmdAccept` now adds `WAITING_ON_MAINTAINER` (fixes the review queue gap @paoloantinori caught at `.github/scripts/pr-lifecycle.js:724`)
- ✅ Stale handling with differentiated thresholds (4/7 for waiting-on-author vs 7/14 default) and `{close_window}` message placeholder
- ✅ `handlePrReadyForReview` idempotency — skips init if lifecycle label already present
- ✅ `verify.yaml` cleanly collapses `run_smoke`/`run_full` into single `run_tests`
- ✅ SonarQube passed with 0 new issues

paoloantinori:
Thanks @carlesarnal for the re-review — agreed on the commit-message cleanup point for the rebase-merge strategy. My iterative passes on the legacy-label migration, draft handling, and waiting-on-author stale thresholds all hold up. Nothing further from my side.

.github/scripts/pr-lifecycle.js
```diff
@@ -426,27 +419,63 @@ async function checkAndTransitionToReady(api, pr, core, reviews) {
...
+
+  if (await migrateLegacyLabels(api, pr, core)) return;
+
   const state = getLifecycleState(pr);
 
   // 1. No lifecycle label at all → initialize as new PR
   if (!state) {
     if (hasLabel(pr, LABELS.DISABLED)) return;
+    // The orchestrator ignores draft PRs entirely — nothing to reconcile.
+    if (pr.draft) return;
```
paoloantinori: The `pr.draft` guard here only covers the no-label (`!state`) initialization path. A PR that already carries a lifecycle label — e.g. opened non-draft and later converted to draft — still falls through to `reconcile`'s state branches and `handlePrSynchronize`, and `verify.yaml`'s scope decision keys off labels with no draft check. CI can therefore still run on such a draft, which contradicts the docs' statement that drafts are "ignored entirely (no labels, no CI)." Consider hoisting a draft guard to the top of `reconcile`/`handlePrSynchronize` (after migration) and handling the `converted_to_draft` action to strip lifecycle labels.

.github/scripts/pr-lifecycle.js
```diff
@@ -426,27 +419,63 @@ async function checkAndTransitionToReady(api, pr, core, reviews) {
...
+  await api.removeLabel(pr.number, LEGACY_WIP_LABEL);
+
+  // Drafts are outside the lifecycle now — just drop the legacy label.
+  if (pr.draft) {
+    core.info(`PR #${pr.number} is a draft — removed legacy lifecycle/wip without promoting`);
+    return true;
+  }
+
+  await api.setLifecycleState(pr, LABELS.READY_FOR_REVIEW);
+  await api.addLabel(pr.number, LABELS.WAITING_ON_MAINTAINER);
```
paoloantinori, line 452: Migration adds `waiting-on-maintainer` here without removing `waiting-on-author`. A legacy WIP PR that had failed smoke tests carried `waiting-on-author`, so after migration it ends up with both (contradictory) labels until the retriggered full suite resolves them. The removed `cmdReady` cleared `waiting-on-author` on the equivalent transition — consider adding `await api.removeLabel(pr.number, LABELS.WAITING_ON_AUTHOR);` here.

paoloantinori (commented):
The PR cleanly collapses the WIP/smoke stage into a single `ready-for-review` test gate, with consistent updates across the orchestrator, `verify.yaml`, config, and docs, plus a sensible one-time legacy-label migration. Two gaps are worth addressing before merge: the "drafts are ignored entirely" guarantee is only partially enforced, and the legacy-WIP migration can leave contradictory waiting-on labels.

- `.github/scripts/pr-lifecycle.js:469` — the `pr.draft` guard sits only inside the `if (!state)` branch of `reconcile`, so a PR that already carries a lifecycle label (e.g. opened non-draft, then converted to draft) still flows through the state branches and `handlePrSynchronize`; since `verify.yaml`'s scope decision gates on labels with no draft check, CI can still run on such a draft, contradicting the docs' "ignored entirely (no labels, no CI)" claim.
- `.github/scripts/pr-lifecycle.js:448` — `migrateLegacyLabels` adds `waiting-on-maintainer` without clearing `waiting-on-author`; a WIP PR that had failed smoke tests ends up with both contradictory labels until the next test result, whereas the removed `cmdReady` cleared `waiting-on-author` on the equivalent transition.
- `.github/workflows/verify.yaml` — collapsing `run_smoke`/`run_full` into a single `run_tests` is correct: the two flags were set together in every branch, so each `decide` mapping (build/unit/console-plugin formerly smoke; integration/extras/sdk/cli/operator formerly full) is equivalent under the new flag.

.github/scripts/pr-lifecycle.js
malladinagarjuna2: yeah you are right! i fixed it in the further commits, now clears waiting-on-author before adding waiting-on-maintainer, matching what the removed cmdReady did on the equivalent transition.

.github/scripts/pr-lifecycle.js
malladinagarjuna2: Draft guard hoisted to the top of reconcile after migration, early return added to handlePrSynchronize, converted_to_draft now strips lifecycle labels, and verify.yaml's decide step checks draft status from the API so it can't race with the label strip.

malladinagarjuna2:
hey @paoloantinori  and @carlesarnal  rea;lly thanks for the review. 
sorry for the late reply! i was a bit tied with few things. 

reworded the commits 
 also added a con....to draft handler that strips lifeecycle labels and verify.yaml decide on step  now reads draft status 
 from the api  rather than  the event payyload, so ci is gated even if  the label strip hasnt landed rather than depending on it .

malladinagarjuna2:
@paoloantinori @carlesarnal @jsenko  cpould you please further review the pr

malladinagarjuna2:
>2. converted_to_draft gap (non-blocking, maintainer decision)
You flagged this yourself in https://github.com/Apicurio/apicurio-registry/pull/9025#issuecomment-5134119088 and it's still valid: PR_LIFECYCLE.md says drafts get no labels and no CI, but nothing enforces that when a PR is converted back to draft:

>pr-lifecycle.yml doesn't trigger on converted_to_draft
>?verify.yaml's decide step doesn't check github.event.pull_request.draft

well i was taking the approval at that time, to change it or not

paoloantinori:
Re-read the latest head (@malladinagarjuna2). The changes since my last pass hold up:

- **Draft-handler constraint** — converting a ready PR back to draft now strips all `lifecycle/*` labels and re-entering ready correctly restarts at `lifecycle/new` (requiring a fresh `/accept`). That closes the gap I flagged earlier.
- **Commit messages** — the reworded history reads cleanly and will survive a rebase-merge without dragging in WIP-era noise.

My substantive review is complete — I have nothing further to add on the lifecycle logic, the threshold math, or the migration path. Deferring the final merge call to @carlesarnal / @jsenko / @EricWittmann.

malladinagarjuna2:
hey @jenko @carlesarnal could you please review

malladinagarjuna2:
hey @jsenko  @EricWittmann @carlesarnal   the pr has been rebased

jsenko (commented):
Reviewed the full diff and the discussion history. The lifecycle simplification is well-structured — paoloantinori's iterative review caught and resolved all the important edge cases (legacy migration, draft handling, stale grace window, waiting-on-maintainer after /accept). carlesarnal's commit-message cleanup is done. No further issues found from my side.

One improvement for a follow-up: the stale warning for `waiting-on-maintainer` PRs should mention the assigned reviewer. Currently a PR where the author did their part but the reviewer is busy gets the same generic stale message — the author has no signal that the bottleneck isn't them. Pinging the assigned reviewer in the stale message (or adjusting the stale timer for `waiting-on-maintainer` PRs) would make this fairer. I'll add this to the PR description as a follow-up item — happy to take it as a separate PR if you're interested, @malladinagarjuna2.

jsenko (approved):

malladinagarjuna2:
>waiting-on-maintainer PRs) would make this fairer. I'll add this to the PR description as a follow-up item — happy to take it as a separate PR if you're interested, @malladinagarjuna2.

yes @jsenko  wil;l take this as a seperate pull request

jsenko:
The `Build UI Application` failure on this PR is caused by a missing TypeScript SDK generated client in the UI-only CI job — not related to your changes. Fix is in #9227. Please rebase after it merges to get green CI.

malladinagarjuna2:
alright @jsenko  thank you so  much

malladinagarjuna2:
@jsenko  not able to fix the unit tests part

PR #9028, opened 2026-07-30
refactor(avro): route Schema.Parser construction through AvroParserAccessor

### LINKED ISSUE #9027: Avro Schema.Parser is constructed at 10 independent sites with no shared configuration point
## Description

**Registry Version**: snapshot (`main` @ `a7838fcc7`, `3.3.2-SNAPSHOT`)
**Persistence type**: N/A, this is in the schema-util and content-handling layers, so it applies to all four storage variants equally.

I was reading through the Avro content handling while working on something else and noticed that `new Schema.Parser()` is constructed independently at ten separate places in production code, each with whatever the library defaults happen to be. There's no shared place to configure them.

The XML, XSD and WSDL types already solve this. `DocumentBuilderAccessor`, `SchemaFactoryAccessor` and `WSDLReaderAccessor` each give their module one construction point where parser configuration lives, and every call site goes through it. Avro just never got the equivalent.

To be clear about severity: I'm not reporting a vulnerability and I don't have an exploit. Right now the Avro library defaults are the safe ones, so nothing is actually mis-configured today. The problem is structural. If we ever want to add a restriction — the obvious candidate being a nesting-depth limit so a deeply nested schema can't exhaust the stack — it has to be applied in ten places and kept in sync forever. It's the same duplicated-construction shape that made the XML parser hardening painful, which is presumably why those three accessors exist.

The eight sites in `schema-util/avro`:

File | Line
-- | --
`avro/content/AvroContentAccepter.java` | 33
`avro/content/canon/AvroContentCanonicalizer.java` | 60
`avro/content/canon/EnhancedAvroContentCanonicalizer.java` | 46
`avro/content/dereference/AvroDereferencer.java` | 15
`avro/content/extract/AvroStructuredContentExtractor.java` | 27
`avro/rules/validity/AvroContentValidator.java` | 37
`avro/rules/compatibility/AvroCompatibilityChecker.java` | 23 and 29

And two more outside that module that are easy to miss:

File | Line | Note
-- | -- | --
`app/.../contracts/tags/AvroTagExtractor.java` | 33 |  
`app/.../ccompat/rest/v7/impl/SchemaFormatService.java` | 112 | ccompat v7 path

The second one is the reason I'm bothering to file this. It's on the Confluent-compatibility surface, so any hardening that got applied to "the eight Avro sites" and stopped there would leave the ccompat path out. That's the site you'd least want to miss.

One related thing I noticed while looking: the seeding of resolved references isn't consistent either. `EnhancedAvroContentCanonicalizer` retries until it reaches a fixpoint, so it copes with references that depend on each other in any map order. `AvroDereferencer`, `AvroContentValidator` and `AvroCompatibilityChecker` all do a single naive pass, so a valid set of interdependent references can fail depending on iteration order. That's arguably its own small bug and would be fixed for free by routing them through one helper.

### Environment

Not environment-specific. Observed by reading the source on `main`; applies to every deployment. Avro version is 1.12.1 (`avro.version` in the root `pom.xml`). Java 17 source / 21 runtime.

### Steps to Reproduce

There's no runtime symptom to reproduce, so this is a static observation:

1. From a clean checkout of `main`:

```bash
grep -rn "new Schema.Parser()" --include="*.java" schema-util app \
  | grep -v /target/ | grep "/main/"
```

2. Ten hits, in the files listed above.
3. Compare with the XML side, where the equivalent grep for `DocumentBuilderFactory.newInstance()` in `schema-util/xml` returns one hit, inside `DocumentBuilderAccessor`.

**Expected**: one place in `schema-util/avro` where a `Schema.Parser` is created and configured, matching what `schema-util/xml` and `schema-util/xsd` already do, with all call sites going through it. Adding a parser restriction should be a one-line change in one file.

**Actual**: ten independent construction sites across two Maven modules. Adding a restriction means ten edits, and there's nothing that fails if you miss one.

### Logs

No runtime logs — nothing is failing. What is probably more useful is the current parser behaviour, since it determines what any fix is allowed to change. I checked these against Avro 1.12.1 directly rather than going from the docs:

```bash
default validateDefaults = true
schema with name "has-dash"          -> REJECTED (SchemaParseException)
schema with int field, default "nope"-> REJECTED (AvroTypeException)
named types carry over between parse() calls on one parser instance: YES
```

Two things follow from that, which I'd want a fix to respect:

The defaults are already the strict ones, so this should be a behaviour-preserving change. Setting the flags explicitly is worth doing so a future Avro release can't quietly loosen them, but it shouldn't alter which schemas are accepted today. Changing that would move canonical hashes and compatibility verdicts, which is a much bigger deal than this issue warrants.

And the accessor should hand back a fresh parser each call rather than caching one in a `ThreadLocal` the way `DocumentBuilderAccessor` does. `Schema.Parser` accumulates the named types it has seen and carries them into later `parse()` calls, which is exactly what reference seeding relies on. A cached instance would leak type definitions between unrelated schemas, so a schema could reference a type it never declares just because an earlier request on the same thread happened to define that name. Worth calling out explicitly in the code, otherwise someone will later "fix" the inconsistency with the other accessors and reintroduce it.

Also worth noting for whoever picks this up: `Schema.Parser.setValidate(boolean)` doesn't exist in 1.12. Name validation is set through the constructor and the field is `final`, so only `setValidateDefaults` is settable after construction.

Happy to put up a PR for this if it's something you'd take.

1PoPTRoN:
## Summary

`Schema.Parser` was constructed at ten independent sites in production code, each relying on whatever the Avro defaults happened to be. There was no shared place to configure them, so any future hardening (the obvious candidate being a nesting-depth limit to bound stack usage) would have to be applied ten times and kept in sync, with nothing failing if one was missed.

This adds `AvroParserAccessor` as the single construction point, mirroring the existing `DocumentBuilderAccessor`, `SchemaFactoryAccessor` and `WSDLReaderAccessor`, and routes all ten call sites through it.

Fixes #9027

## Root Cause

Eight sites in `schema-util/avro`, plus two outside it that are easy to miss:

| File | Line |
|---|---|
| `avro/content/AvroContentAccepter.java` | 33 |
| `avro/content/canon/AvroContentCanonicalizer.java` | 60 |
| `avro/content/canon/EnhancedAvroContentCanonicalizer.java` | 46 |
| `avro/content/dereference/AvroDereferencer.java` | 15 |
| `avro/content/extract/AvroStructuredContentExtractor.java` | 27 |
| `avro/rules/validity/AvroContentValidator.java` | 37 |
| `avro/rules/compatibility/AvroCompatibilityChecker.java` | 23, 29 |
| `app/.../contracts/tags/AvroTagExtractor.java` | 33 |
| `app/.../ccompat/rest/v7/impl/SchemaFormatService.java` | 112 |

The last one is on the Confluent-compatibility surface, so a change scoped to `schema-util/avro` alone would have left the ccompat path unhardened. That is the site you would least want to miss.

For contrast, the equivalent grep on the XML side returns exactly one hit, inside `DocumentBuilderAccessor`:

```bash
grep -rn "DocumentBuilderFactory.newInstance()" --include="*.java" schema-util app | grep -v /target/ | grep "/main/"
# -> schema-util/xml/.../util/DocumentBuilderAccessor.java:18
```

## Changes

- New `schema-util/avro/.../avro/util/AvroParserAccessor.java` with `newParser()` and `newParser(Map<String, TypedContent>)`.
- All ten call sites now obtain their parser from it. After this change the only `new Schema.Parser()` in production code is inside the accessor itself.
- `EnhancedAvroContentCanonicalizer` loses 24 lines of inline seeding logic, which moved into the accessor unchanged.

Net: 11 files, +280 / −51. Of the additions, 160 lines are the new test, so this removes more production code than it adds.

### Two design points worth calling out

**No `ThreadLocal` caching, deliberately.** This is where the accessor differs from `DocumentBuilderAccessor`, and the deviation is intentional. `Schema.Parser` accumulates the named types it has parsed and carries them into later `parse()` calls on the same instance, which is exactly what reference seeding relies on. A cached or thread-cached parser would leak type definitions between unrelated schemas, letting a schema reference a type it never declares because an earlier, unrelated parse happened to define that name. Each caller gets a fresh instance and only the configuration is shared. This is stated in the class javadoc and locked in by `eachCallReturnsAnIndependentParser`, so it is not later "fixed" for consistency with the other accessors.

**Behaviour preserving on validation.** `setValidateDefaults(true)` is set explicitly, but that is already the Avro 1.12.1 default, so which schemas are accepted does not change. It is pinned so a future Avro release cannot quietly loosen it. Verified against 1.12.1 directly rather than from the docs:

```
default validateDefaults = true
name "has-dash"                      -> REJECTED (SchemaParseException)
int field with default "nope"        -> REJECTED (AvroTypeException)
named types carry over between parse() calls: YES
```

Note that `Schema.Parser.setValidate(boolean)` does not exist in 1.12 — name validation is constructor-only and the field is `final` — so the strict default is inherited from the no-argument constructor.

## Behaviour change (one, deliberate)

Reference seeding is unified on the retry-to-fixpoint pass that `EnhancedAvroContentCanonicalizer` already used. `AvroDereferencer`, `AvroContentValidator` and `AvroCompatibilityChecker` previously did a single naive pass, so a valid set of interdependent references could fail depending on map iteration order. Those three now cope with any order.

Consequences:

- Valid schemas with interdependent references that previously failed on unlucky iteration order now succeed. Covered by `seededParserResolvesInterdependentReferencesInAnyOrder`.
- An unparseable reference is now swallowed during seeding rather than thrown immediately. If the broken reference was actually needed, the main parse still fails and still names the offending type — locked in by `unresolvableReferenceStillFailsTheMainParse`. Only the case where a broken reference was *unused* changes from failure to success. `AvroDereferencer` is the one site where this is observable as a thrown-exception change, since the other two already wrapped in `try`/`catch`.

`AvroContentAccepter` and `AvroContentCanonicalizer` keep their own seeding loops, because they also need the parsed reference `Schema` objects for `Schema#toString(refs, false)`. They use `newParser()` for construction only, so their behaviour is untouched.

If you would rather this were a pure refactor, dropping the seeding overload from those three call sites leaves the accessor intact and the diff smaller. Happy to do that.

## Upgrade impact

None. No `RegistryStorage` method, SQL statement, DDL, KafkaSQL message type or `db-version` is touched, so there is no migration and no storage-variant parity concern. No OpenAPI change, so no SDK regeneration.

Canonical hashes are unchanged: `SchemaNormalizerTest` (30 cases, covering the canonicalizer's output byte-for-byte) passes untouched.

## Test plan

New `AvroParserAccessorTest`, 7 cases. The load-bearing one is an invariant guard rather than a behaviour test, since the defect *is* the duplication: it walks this module's `src/main/java` and fails if anything other than the accessor constructs a `Schema.Parser`.

Fail→pass confirmed by reintroducing the bug at one site:

```
Tests run: 7, Failures: 1
AssertionFailedError: Schema.Parser must be obtained from AvroParserAccessor so that parser
hardening lives in one place. Direct construction found in:
[src/main/java/io/apicurio/registry/avro/content/extract/AvroStructuredContentExtractor.java]
```

Run with CI's flags (`-Pprod`, and `-Dfull -DcliSkipNative` for `app`):

- [x] `clean test -Pprod -pl schema-util/avro -am` — **44/44** (7 new + `SchemaNormalizerTest` 30 + `AvroCompatibilityTest` 6 + `AvroSchemaSmokeTest` 1)
- [x] `app` downstream — **33/33** (`AvroCanonicalHashUpgraderTest` 4, `ArtifactTypeUtilTest` 12, `AvroTagExtractorTest` 14, `TagExtractorFactoryTest` 3)
- [x] `test-compile -Pprod -pl app -am -Dfull -DcliSkipNative` — clean
- [x] `checkstyle:check -pl schema-util/avro,app` — 0 violations in both
- [x] `./scripts/validate-files.sh` — DB version 108 consistent across all four DDLs
- [x] `git status --porcelain docs` — clean, no docs regeneration needed
- [x] Multi-module reactor from the repo root — the guard test resolves `src/main/java` relative to the module basedir, and the `non-app` unit-test shard runs from the root across many modules, so I verified it there too rather than only under `-pl schema-util/avro`

## Not verified

- `SchemaFormatService` (ccompat v7) has no unit test. The change there is construction-only and behaviour-identical, but it sits on a surface consumed by third-party Confluent clients, so it would be worth someone running `-Pintegration-tests -Premote-sql -Dgroups=default` (which picks up `ConfluentClientTest`). I did not run the integration suite.
- MySQL and MSSQL have no integration coverage in CI. Not relevant here since nothing touches SQL — noted only so the omission is not read as an oversight.

## Notes for reviewers

`AvroContentCanonicalizer` appears to be dead code: its only reference anywhere in the repo is its own declaration. I routed it through the accessor for consistency but deliberately did not delete it, since it is `public` in a published module and removing it is an API change that belongs in its own PR. Same call I made in #8973.

### DIFF
--- a/schema-util/avro/src/main/java/io/apicurio/registry/avro/util/AvroParserAccessor.java
+++ b/schema-util/avro/src/main/java/io/apicurio/registry/avro/util/AvroParserAccessor.java
@@ -1,0 +1,93 @@
+package io.apicurio.registry.avro.util;
+
+import io.apicurio.registry.content.TypedContent;
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.Schema;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Single construction point for {@link Schema.Parser} instances, mirroring the role that
+ * {@code DocumentBuilderAccessor} and {@code SchemaFactoryAccessor} play for the XML and XSD types. Every
+ * Avro parser used by the registry should be obtained here so that parser hardening lives in exactly one
+ * place instead of being duplicated across every call site.
+ * <p>
+ * Unlike the XML and XSD accessors, this class deliberately does <em>not</em> cache the parser in a
+ * {@link ThreadLocal}. {@link Schema.Parser} accumulates the named types it has seen and carries them into
+ * subsequent {@code parse()} calls on the same instance — behaviour that
+ * {@link #newParser(Map)} depends on. A shared or thread-cached parser would therefore leak type
+ * definitions between unrelated schemas, letting a schema that references a type it never declares parse
+ * successfully because an earlier, unrelated parse happened to define that name. Each caller gets a fresh
+ * instance; only the configuration is shared.
+ */
+public final class AvroParserAccessor {
+
+    private AvroParserAccessor() {
+    }
+
+    /**
+     * @return a new, hardened parser with no named types pre-registered.
+     */
+    public static Schema.Parser newParser() {
+        return configure(new Schema.Parser());
+    }
+
+    /**
+     * Returns a new parser seeded with the given referenced schemas, so that the named types they declare are
+     * already known by the time the referring schema is parsed. Without this, a schema that references
+     * another artifact fails to parse at all.
+     * <p>
+     * References may depend on one another and the map imposes no ordering, so the ones that fail are
+     * re-attempted until a full pass resolves nothing further. Anything still unresolved is left to the
+     * caller's own parse, which reports the offending type name.
+     *
+     * @param resolvedReferences the content of the artifacts referenced by the schema, keyed by reference
+     *            name. May be empty or null.
+     * @return a new, hardened parser with the resolvable references already registered.
+     */
+    public static Schema.Parser newParser(Map<String, TypedContent> resolvedReferences) {
+        Schema.Parser parser = newParser();
+        if (resolvedReferences == null || resolvedReferences.isEmpty()) {
+            return parser;
+        }
+
+        List<String> pending = new ArrayList<>(resolvedReferences.size());
+        for (TypedContent referencedContent : resolvedReferences.values()) {
+            pending.add(referencedContent.getContent().content());
+        }
+        boolean progressed = true;
+        while (progressed && !pending.isEmpty()) {
+            progressed = false;
+            Iterator<String> iter = pending.iterator();
+            while (iter.hasNext()) {
+                try {
+                    parser.parse(iter.next());
+                    iter.remove();
+                    progressed = true;
+                } catch (AvroRuntimeException e) {
+                    // Either this reference depends on another one not parsed yet (retried on the next
+                    // pass), or it cannot be parsed on its own. Neither is fatal here.
+                }
+            }
+        }
+        return parser;
+    }
+
+    /**
+     * Applies the registry's parser hardening. This is the single place to add future restrictions, for
+     * example a nesting-depth limit to bound stack usage on deeply nested schemas.
+     * <p>
+     * The settings applied here currently match the Avro library defaults, so this method is behaviour
+     * preserving. They are set explicitly so that the registry's posture is stated at one site and cannot be
+     * loosened by a change of default in a future Avro release. Note that name validation is fixed at
+     * construction time in Avro 1.12 ({@code Schema.Parser} holds it in a final field and exposes no setter),
+     * so the strict default is inherited from the no-argument constructor above.
+     */
+    private static Schema.Parser configure(Schema.Parser parser) {
+        parser.setValidateDefaults(true);
+        return parser;
+    }
+}
--- a/schema-util/avro/src/main/java/io/apicurio/registry/avro/util/AvroParserAccessor.java
+++ b/schema-util/avro/src/main/java/io/apicurio/registry/avro/util/AvroParserAccessor.java
@@ -80,11 +80,21 @@ public final class AvroParserAccessor {
      * Applies the registry's parser hardening. This is the single place to add future restrictions, for
      * example a nesting-depth limit to bound stack usage on deeply nested schemas.
      * <p>
-     * The settings applied here currently match the Avro library defaults, so this method is behaviour
-     * preserving. They are set explicitly so that the registry's posture is stated at one site and cannot be
-     * loosened by a change of default in a future Avro release. Note that name validation is fixed at
-     * construction time in Avro 1.12 ({@code Schema.Parser} holds it in a final field and exposes no setter),
-     * so the strict default is inherited from the no-argument constructor above.
+     * The settings applied here match the Avro library defaults, so this method is behaviour preserving. They
+     * are set explicitly so that the registry's posture is stated at one site and cannot be loosened by a
+     * change of default in a future Avro release.
+     * <p>
+     * Do not take Avro's own documentation as the authority on the {@code validateDefaults} default: in the
+     * pinned Avro 1.12.1, {@code Schema.java:1432} initialises the field to {@code true} while the
+     * {@code getValidateDefaults()} javadoc at {@code Schema.java:1480} claims "False by default". The
+     * initialiser is what runs, and it was confirmed against the pinned version rather than the docs, so
+     * passing {@code true} here changes nothing today. {@code AvroParserAccessorTest} pins the expected
+     * value so a future Avro upgrade that really does flip the default fails loudly instead of silently
+     * relaxing validation.
+     * <p>
+     * Name validation is fixed at construction time in Avro 1.12 ({@code Schema.Parser} holds it in a final
+     * field and exposes no setter), so the strict default is inherited from the no-argument constructor
+     * above.
      */
     private static Schema.Parser configure(Schema.Parser parser) {
         parser.setValidateDefaults(true);
--- a/schema-util/avro/src/test/java/io/apicurio/registry/avro/util/AvroParserAccessorTest.java
+++ b/schema-util/avro/src/test/java/io/apicurio/registry/avro/util/AvroParserAccessorTest.java
@@ -49,6 +49,13 @@ class AvroParserAccessorTest {
      * independent sites, so any hardening had to be repeated eight times and one could silently be missed.
      * This asserts the invariant itself: within this module's production sources, AvroParserAccessor is the
      * only class allowed to construct a parser.
+     * <p>
+     * Scope limitation: this walks only the current module, so it does not cover the two call sites that were
+     * also routed through the accessor but live in the app module
+     * ({@code contracts/tags/AvroTagExtractor} and {@code ccompat/rest/v7/impl/SchemaFormatService}, the
+     * latter on the Confluent-compatibility path). A regression there would not fail this test. Extending the
+     * same guard to the app module is worth doing separately; do not read a pass here as repository-wide
+     * coverage.
      */
     @Test
     void noProductionCodeInThisModuleConstructsSchemaParserDirectly() throws IOException {
--- a/schema-util/avro/src/test/java/io/apicurio/registry/avro/util/AvroParserAccessorTest.java
+++ b/schema-util/avro/src/test/java/io/apicurio/registry/avro/util/AvroParserAccessorTest.java
@@ -1,0 +1,160 @@
+package io.apicurio.registry.avro.util;
+
+import io.apicurio.registry.content.ContentHandle;
+import io.apicurio.registry.content.TypedContent;
+import io.apicurio.registry.types.ContentTypes;
+import org.apache.avro.Schema;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class AvroParserAccessorTest {
+
+    private static final Pattern DIRECT_CONSTRUCTION = Pattern.compile("new\\s+Schema\\.Parser\\s*\\(");
+
+    private static final String ADDRESS = "{\"type\":\"record\",\"name\":\"Address\","
+            + "\"namespace\":\"com.example\",\"fields\":[{\"name\":\"street\",\"type\":\"string\"}]}";
+
+    private static final String CITY = "{\"type\":\"record\",\"name\":\"City\","
+            + "\"namespace\":\"com.example\",\"fields\":[{\"name\":\"name\",\"type\":\"string\"}]}";
+
+    /** Depends on City, so it cannot be parsed before City has been. */
+    private static final String ADDRESS_WITH_CITY = "{\"type\":\"record\",\"name\":\"Address\","
+            + "\"namespace\":\"com.example\",\"fields\":[{\"name\":\"city\",\"type\":\"com.example.City\"}]}";
+
+    private static final String PERSON_REFERENCING_ADDRESS = "{\"type\":\"record\",\"name\":\"Person\","
+            + "\"namespace\":\"com.example\",\"fields\":[{\"name\":\"address\",\"type\":\"com.example.Address\"}]}";
+
+    private static TypedContent content(String schema) {
+        return TypedContent.create(ContentHandle.create(schema), ContentTypes.APPLICATION_JSON);
+    }
+
+    /**
+     * The guard that fails on the buggy code. Before the fix this module constructed Schema.Parser at eight
+     * independent sites, so any hardening had to be repeated eight times and one could silently be missed.
+     * This asserts the invariant itself: within this module's production sources, AvroParserAccessor is the
+     * only class allowed to construct a parser.
+     */
+    @Test
+    void noProductionCodeInThisModuleConstructsSchemaParserDirectly() throws IOException {
+        // Surefire runs with the working directory set to the module base directory.
+        Path mainSources = Paths.get("src", "main", "java");
+        assertTrue(Files.isDirectory(mainSources),
+                "expected to run from the module base directory, but " + mainSources.toAbsolutePath()
+                        + " does not exist");
+
+        List<Path> offenders;
+        try (Stream<Path> sources = Files.walk(mainSources)) {
+            offenders = sources.filter(p -> p.toString().endsWith(".java"))
+                    // The accessor is the one place allowed to construct a parser.
+                    .filter(p -> !p.getFileName().toString().equals("AvroParserAccessor.java"))
+                    .filter(p -> DIRECT_CONSTRUCTION.matcher(readString(p)).find()).toList();
+        }
+
+        assertTrue(offenders.isEmpty(),
+                "Schema.Parser must be obtained from AvroParserAccessor so that parser hardening lives in "
+                        + "one place. Direct construction found in: " + offenders);
+    }
+
+    @Test
+    void newParserAppliesHardening() {
+        // Explicitly pinned rather than inherited, so a change of Avro default cannot loosen validation.
+        assertTrue(AvroParserAccessor.newParser().getValidateDefaults());
+
+        // Strict name validation is inherited from the no-arg constructor and must stay in effect.
+        assertThrows(Exception.class, () -> AvroParserAccessor.newParser()
+                .parse("{\"type\":\"record\",\"name\":\"has-dash\",\"fields\":[]}"));
+
+        // A bad default value must still be rejected.
+        assertThrows(Exception.class, () -> AvroParserAccessor.newParser().parse(
+                "{\"type\":\"record\",\"name\":\"R\",\"fields\":[{\"name\":\"f\",\"type\":\"int\",\"default\":\"nope\"}]}"));
+    }
+
+    /**
+     * The reason this accessor must not cache parsers the way DocumentBuilderAccessor does: a parser
+     * accumulates named types, so a reused instance would let a schema reference a type it never declares.
+     */
+    @Test
+    void eachCallReturnsAnIndependentParser() {
+        Schema.Parser first = AvroParserAccessor.newParser();
+        Schema.Parser second = AvroParserAccessor.newParser();
+        assertNotSame(first, second);
+
+        first.parse(ADDRESS);
+        assertTrue(first.getTypes().containsKey("com.example.Address"));
+
+        // The second parser must not have inherited Address from the first.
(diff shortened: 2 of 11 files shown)


schema-util/avro/src/main/java/io/apicurio/registry/avro/util/AvroParserAccessor.java
```diff
@@ -0,0 +1,93 @@
...
+     * Applies the registry's parser hardening. This is the single place to add future restrictions, for
+     * example a nesting-depth limit to bound stack usage on deeply nested schemas.
+     * <p>
+     * The settings applied here currently match the Avro library defaults, so this method is behaviour
+     * preserving. They are set explicitly so that the registry's posture is stated at one site and cannot be
+     * loosened by a change of default in a future Avro release. Note that name validation is fixed at
+     * construction time in Avro 1.12 ({@code Schema.Parser} holds it in a final field and exposes no setter),
+     * so the strict default is inherited from the no-argument constructor above.
+     */
+    private static Schema.Parser configure(Schema.Parser parser) {
```
paoloantinori, line 99: Worth double-confirming against the pinned Avro version: the Avro `Schema.Parser` source initializes `private boolean validateDefaults = true`, so this is behavior-preserving today — but Avro's own `getValidateDefaults()` Javadoc says "False by default", contradicting its field initializer. A reader trusting the Javadoc would think this tightens validation. The test pins the value, so it's fine; just flagging the doc contradiction so the behavior-preservation claim isn't challenged later.

schema-util/avro/src/test/java/io/apicurio/registry/avro/util/AvroParserAccessorTest.java
```diff
@@ -0,0 +1,160 @@
...
+     * This asserts the invariant itself: within this module's production sources, AvroParserAccessor is the
+     * only class allowed to construct a parser.
+     */
+    @Test
+    void noProductionCodeInThisModuleConstructsSchemaParserDirectly() throws IOException {
+        // Surefire runs with the working directory set to the module base directory.
+        Path mainSources = Paths.get("src", "main", "java");
+        assertTrue(Files.isDirectory(mainSources),
+                "expected to run from the module base directory, but " + mainSources.toAbsolutePath()
+                        + " does not exist");
```
paoloantinori, line 66: This invariant test walks `src/main/java` of the current module only, so it guards the eight `schema-util/avro` sites but not the two `app/` call sites (`AvroTagExtractor`, `SchemaFormatService`) that this PR also routes through the accessor. The ccompat path is the one the PR description explicitly calls out as easy to miss — worth a one-line comment here noting the module scope so a future reader doesn't assume the invariant covers the whole registry.

paoloantinori (commented):
Clean, well-motivated refactor: centralizing `Schema.Parser` construction behind `AvroParserAccessor` is the right call given the ten scattered sites, and the decision *not* to cache parsers in a `ThreadLocal` (because `Schema.Parser` accumulates named types) is correctly reasoned and well documented. The fixpoint retry in `newParser(Map)` moves verbatim out of `EnhancedAvroContentCanonicalizer`, and the two-parser pattern in `AvroCompatibilityChecker` is preserved with a clear comment. A few observations worth a look:

- **`validateDefaults` pinning — worth double-confirming against the pinned Avro version.** `configure()` calls `parser.setValidateDefaults(true)`, which the Javadoc frames as matching the Avro default. The Avro `Schema.Parser` source initializes `private boolean validateDefaults = true` (field initializer), so today this *is* behavior-preserving. The catch: Avro's own accessor Javadoc reads *"True iff default values are validated. False by default"* — i.e. the Avro Javadoc contradicts its own field initializer. So the explicit `setValidateDefaults(true)` is correct and matches the *runtime* default, but it's worth a one-line confirmation in the PR that you verified the default against the exact Avro version the registry pins, since a reader trusting the Avro Javadoc would wrongly conclude this changes behavior. The test `newParserAppliesHardening` already pins the expected value, which is good.

- **The invariant test is scoped to `schema-util/avro` only.** `noProductionCodeInThisModuleConstructsSchemaParserDirectly` walks `src/main/java` of the current module, so it will catch future regressions in the eight `schema-util/avro` sites but not the two call sites routed through the accessor outside that module (`app/.../contracts/tags/AvroTagExtractor.java` and `app/.../ccompat/rest/v7/impl/SchemaFormatService.java`). The PR description is upfront that those two exist; just flagging that the architectural invariant the test enforces doesn't cover the `app` module, where the ccompat path (the one you'd least want to miss) lives. Not blocking — a module-scoped test is a reasonable tradeoff — but worth a comment in the test noting the scope limitation so a future reader doesn't assume full coverage.

- **Reference-resolution strategy is not unified by this refactor.** `AvroContentAccepter` and `AvroContentCanonicalizer` still seed the parser with a naive sequential loop (`newParser()` + manual `parse` per reference), whereas `EnhancedAvroContentCanonicalizer` now goes through `newParser(Map)` which does the fixpoint retry-to-convergence. That's pre-existing behavior and correctly preserved — the inline comments in those two files accurately explain *why* they can't use `newParser(Map)` (they need the parsed `Schema` objects themselves) — so this is an observation, not a defect. Just noting that the centralization normalizes *construction* but not *reference resolution*, so the hardening motivation (single place for future nesting limits) is fully achieved while the resolution differences remain.

Nice test design overall — the `eachCallReturnsAnIndependentParser` and `seededParserResolvesInterdependentReferencesInAnyOrder` cases pin the two subtle correctness properties that would be easy to regress.

1PoPTRoN:
Thanks @paoloantinori — the `validateDefaults` point is a better catch than it first looks, and you're right on both counts. Pushed 2d9426627 for the two documentation items.

**1. `validateDefaults` confirmed against the pinned version.** I checked the 1.12.1 sources jar rather than the docs, and the contradiction you describe is real:

```
Schema.java:1432    private boolean validateDefaults = true;
Schema.java:1480    /** True iff default values are validated. False by default. */
```

The field initialiser is what runs, so `setValidateDefaults(true)` is a no-op today and the change is behaviour preserving. I'd also verified it empirically before writing the accessor:

```
default validateDefaults = true
name "has-dash"                -> REJECTED (SchemaParseException)
int field with default "nope"  -> REJECTED (AvroTypeException)
```

Your framing is the useful one though: because Avro's own javadoc says the opposite, "matches the library default" is not something a reviewer can check by reading Avro's docs. I've put the two line references and the discrepancy into the `configure()` javadoc so the next reader doesn't have to re-derive it, and noted that `newParserAppliesHardening` is what makes a genuine future flip fail loudly rather than silently relax validation.

**2. Guard test scope.** Correct, and worth being explicit about — the test walks only this module, so a regression at `AvroTagExtractor` or `SchemaFormatService` would not fail it, and the ccompat one is exactly the site you'd least want uncovered. Added a scope-limitation note to the test javadoc naming both call sites, so a pass isn't mistaken for repository-wide coverage. I left the app-module guard out of this PR to keep it to one concern, but happy to add it here instead if you'd prefer it in one go.

**3. Reference resolution.** Agreed, and no change made. `AvroContentAccepter` and `AvroContentCanonicalizer` need the parsed `Schema` objects for `Schema#toString(refs, false)`, so they can't use `newParser(Map)` without losing that. Construction is centralised, resolution deliberately isn't, and the hardening motivation is satisfied either way.

One thing worth flagging separately: the three sites that *did* move onto `newParser(Map)` (`AvroDereferencer`, `AvroContentValidator`, `AvroCompatibilityChecker`) previously did a single naive pass, so they now resolve interdependent references regardless of map order. That's the one behaviour change in the PR and it's covered by `seededParserResolvesInterdependentReferencesInAnyOrder`. If you'd rather this were a pure refactor, dropping the `newParser(Map)` adoption at those three sites leaves the accessor intact and shrinks the diff — say the word and I'll do it.

Also: CI hasn't run yet, since the PR is still on `lifecycle/new` / `lifecycle/waiting-on-maintainer` and the Verification Gate fails the lifecycle check before any job starts. Needs an `/accept` when you're happy with it (or `/retry`, per the bot's fork-label caveat).

paoloantinori:
Thanks @1PoPTRoN — this is exactly the right resolution on both counts.

Confirming the `validateDefaults` finding against the 1.12.1 sources jar rather than the docs is the right call; the field initializer at `Schema.java:1432` is what runs, so pinning it explicitly in `configure()` plus the `newParserAppliesHardening()` assertion is the robust posture. If a future Avro upgrade genuinely flips the default, the test fails loudly instead of silently relaxing validation — which is the whole point.

The `noProductionCodeInThisModuleConstructsSchemaParserDirectly()` guard is a nice touch too. Your scope-limitation caveat in its javadoc is honest and correct: it covers the eight `schema-util/avro` sites but not the two in `app/` (`AvroTagExtractor`, `SchemaFormatService`). Worth a follow-up to extend the same scan to the app module so a regression on the ccompat path can't slip through either — but that's clearly out of scope here.

One housekeeping note: `mergeable_state` is `behind`, so this will need a rebase before merge. Nothing blocking from my side.

schema-util/avro/src/main/java/io/apicurio/registry/avro/util/AvroParserAccessor.java
```diff
@@ -0,0 +1,103 @@
...
+        for (TypedContent referencedContent : resolvedReferences.values()) {
+            pending.add(referencedContent.getContent().content());
+        }
+        boolean progressed = true;
+        while (progressed && !pending.isEmpty()) {
+            progressed = false;
+            Iterator<String> iter = pending.iterator();
+            while (iter.hasNext()) {
+                try {
+                    parser.parse(iter.next());
```
paoloantinori, line 67: This `catch (AvroRuntimeException)` looks broader than the PR's 'unused unparseable reference' framing. In Avro 1.12.x `SchemaParseException extends AvroRuntimeException`, and `Schema.parseRecord` stages the named type (`context.put`, ~Schema.java:1864) and calls `setFields` (~1878) *before* later validation (`parseAliases` ~1880; the dangling-content check ~1542), with no rollback on failure. So a *used* reference that is a complete record plus a trailing validation error — e.g. `{"type":"record","name":"Rec","fields":[{"name":"f","type":"int"}],"aliases":"notAnArray"}` — throws *after* `Rec` is already staged; the throw is swallowed here, and the main schema's parse then resolves `Rec` and succeeds. Net effect: `AvroContentValidator` can mark such content VALID, `AvroDereferencer` emits output, and `AvroCompatibilityChecker` runs against an alias-stripped type — where the prior single-pass code rejected the whole artifact. Worth confirming with that input (seed it as a used reference and assert the main parse still fails), and consider either narrowing the catch or re-validating the parsed references rather than swallowing. Caveat: if references are always pre-validated before reaching the accessor, this path may not occur in practice — but the validator exists precisely to catch malformed input, so the false-negative there is the case I'd most want closed.

schema-util/avro/src/test/java/io/apicurio/registry/avro/util/AvroParserAccessorTest.java
```diff
@@ -0,0 +1,167 @@
...
+     * also routed through the accessor but live in the app module
+     * ({@code contracts/tags/AvroTagExtractor} and {@code ccompat/rest/v7/impl/SchemaFormatService}, the
+     * latter on the Confluent-compatibility path). A regression there would not fail this test. Extending the
+     * same guard to the app module is worth doing separately; do not read a pass here as repository-wide
+     * coverage.
+     */
+    @Test
+    void noProductionCodeInThisModuleConstructsSchemaParserDirectly() throws IOException {
+        // Surefire runs with the working directory set to the module base directory.
+        Path mainSources = Paths.get("src", "main", "java");
```
paoloantinori, line 63: This guard is the PR's load-bearing invariant test (the thesis is 'the defect IS the duplication'), but it only walks this module's `src/main/java`. The two routed call sites in the `app` module — `contracts/tags/AvroTagExtractor` and especially `ccompat/rest/v7/impl/SchemaFormatService` (the Confluent-compatibility surface the description calls 'the site you would least want to miss') — are not covered, so a future `new Schema.Parser()` added there passes green with the invariant silently violated. The test's own javadoc acknowledges the limitation; consider extending the walk to the app module's sources or adding a companion guard test in the app module so the most important site is actually protected.

schema-util/avro/src/test/java/io/apicurio/registry/avro/util/AvroParserAccessorTest.java
```diff
@@ -0,0 +1,167 @@
...
+     * An unresolvable reference must not mask the real failure: the main parse still reports the type it
+     * cannot find.
+     */
+    @Test
+    void unresolvableReferenceStillFailsTheMainParse() {
+        Map<String, TypedContent> references = Map.of("broken", content("{ not a schema }"));
+
+        Exception ex = assertThrows(Exception.class,
+                () -> AvroParserAccessor.newParser(references).parse(PERSON_REFERENCING_ADDRESS));
+        assertTrue(ex.getMessage().contains("com.example.Address"),
```
paoloantinori, line 156: Asserting `ex.getMessage().contains("com.example.Address")` couples this test to Avro's error-message wording, which is not part of its stable API and has changed across versions — a future Avro bump could fail this test with no behavior change. The test's actual intent is 'the main parse still fails when a reference is unresolvable'; asserting on the exception type (or simply that the parse throws) expresses that more robustly than substring-matching the message.

paoloantinori (commented):
Overall this is a clean, well-documented consolidation: the single construction point is sound, the two-parser design in `AvroCompatibilityChecker` is correct, and pinning `validateDefaults(true)` is genuinely behavior-preserving (Avro 1.12's default is already `true`). The one substantive concern worth resolving before merge is that the broadened reference-seeding `catch` can silently accept artifacts whose references carry certain post-field validation errors.

- `AvroParserAccessor.java:67` — the seeding `catch (AvroRuntimeException)` swallows reference-parse failures that occur *after* Avro has already staged and field-set the named schema. In 1.12.x `SchemaParseException extends AvroRuntimeException`, and `Schema.parseRecord` stages the type (`context.put`, ~Schema.java:1864) and calls `setFields` (~1878) *before* late validation (`parseAliases` ~1880, dangling-content check ~1542), with no rollback on failure. So a *used* reference that is a complete record plus a trailing error — e.g. `{"type":"record","name":"Rec","fields":[{"name":"f","type":"int"}],"aliases":"notAnArray"}` — throws after `Rec` is already staged; the throw is swallowed here and the main schema's parse then resolves `Rec` and succeeds. That is broader than the PR's 'unused unparseable reference' framing, and is newly observable at `AvroContentValidator` (false 'valid'), `AvroDereferencer`, and `AvroCompatibilityChecker`, where the prior single-pass code rejected the artifact; it was pre-existing only in `EnhancedAvroContentCanonicalizer`.
- `AvroParserAccessorTest.java:63` — the invariant guard (the PR's load-bearing test, since 'the defect IS the duplication') only walks this module's `src/main/java`, so the two routed `app`-module sites — `contracts/tags/AvroTagExtractor` and especially `ccompat/rest/v7/impl/SchemaFormatService` (the Confluent-compat surface the description calls 'the site you would least want to miss') — are not covered; a regression there passes green. The test's own javadoc acknowledges this.
- `AvroParserAccessorTest.java:156` — asserting on Avro's error-message wording is fragile across Avro versions; the test's real intent ('the main parse still fails') is better expressed via the exception type.
- Positives confirmed: the deliberate no-`ThreadLocal` decision and its independence guard are correct, the fixpoint loop terminates, and keeping `AvroContentAccepter`/`AvroContentCanonicalizer`'s own loops (they need the parsed `Schema` objects for `toString(refs, false)`) is the right call.

EricWittmann:
Thanks for this, @1PoPTRoN — clean refactor, thorough tests, and an excellent PR description. Merged!

PR #9031, opened 2026-07-30
fix(import): set import status to SKIPPED when startup import is skipped

### LINKED ISSUE #9030: Startup import skip incorrectly marks Registry as unready when data already exists
## Description

**Registry Version**: `main` (or the commit you tested)
**Persistence type**: `sql` (replace with your storage backend if different)

When `apicurio.import.url` is configured, the Registry performs a data import during startup.

If the Registry already contains data (for example, after restarting with persistent storage), the import is intentionally skipped because the import process detects existing data and throws a `ConflictException`.

However, `ImportLifecycleBean` currently sets the import status to `ERROR` when this exception is caught:

```java
} catch (ConflictException ce) {
    log.info("Import skipped, registry not empty.");
    importStatus = ImportStatus.ERROR;
}
```

The readiness check only reports the Registry as ready when the status is `SKIPPED` or `COMPLETE`. As a result, the readiness endpoint reports the application as **not ready**, even though the Registry is fully operational and the import was skipped as expected.

### Environment

* Apicurio Registry: `main` (or specify the released version)
* Java: `21.0.12`
* Maven: `3.9.8`
* Storage: SQL
* Deployment: Local development environment

### Steps to Reproduce

1. Configure `apicurio.import.url`.
2. Start the Registry with an empty database.
3. Wait for the import to complete successfully.
4. Restart the Registry while using the same persistent database.
5. Observe the log message:

   ```
   Import skipped, registry not empty.
   ```
6. Access the readiness endpoint (`/health/ready`).

**Expected**

Since the Registry already contains data, skipping the import is expected behaviour. The Registry should transition to a ready state, and the readiness endpoint should return **HTTP 200 OK**.

**Actual**

The `ConflictException` handler sets the import status to `ERROR`, causing the readiness check to fail. As a result, `/health/ready` returns **HTTP 503 Service Unavailable**, even though the Registry is fully operational.

### Logs

```
INFO  Import skipped, registry not empty.
```

If helpful, I can also submit a PR to address this by transitioning the import status to `SKIPPED` instead of `ERROR` when the import is intentionally skipped due to existing data.

kunaldevsahu:
## Summary

This PR fixes the startup import status when the registry already contains data.

When `apicurio.import.url` is configured, Apicurio Registry attempts to import data during startup. If the registry is already populated (for example, after restarting with persistent storage), the import process throws a `ConflictException` and skips the import as expected.

However, `ImportLifecycleBean` currently sets the import status to `ERROR` when handling this exception, causing the readiness probe to report the application as **not ready** even though the registry is fully operational.

This PR updates the import status to `SKIPPED`, ensuring the readiness check correctly reflects the application's state.

## Changes

* Update `ImportLifecycleBean` to set `ImportStatus.SKIPPED` instead of `ImportStatus.ERROR` when a `ConflictException` occurs during startup import.
* Preserve the existing behaviour and log message indicating that the import was intentionally skipped.
* Add a regression test to verify that the registry reports a ready state when startup import is skipped because the registry already contains data.

## Testing

The following scenario was verified:

1. Configure `apicurio.import.url`.
2. Start the registry with an empty database and allow the import to complete.
3. Restart the registry while using the same persistent database.
4. Verify that the startup import is skipped because the registry already contains data.
5. Confirm that the readiness endpoint (`/health/ready`) returns **HTTP 200 OK**.

## Related Issue

Fixes #9030

### DIFF
--- a/app/src/main/java/io/apicurio/registry/ImportLifecycleBean.java
+++ b/app/src/main/java/io/apicurio/registry/ImportLifecycleBean.java
@@ -60,7 +60,7 @@ public class ImportLifecycleBean {
                     importStatus = ImportStatus.ERROR;
                 } catch (ConflictException ce) {
                     log.info("Import skipped, registry not empty.");
-                    importStatus = ImportStatus.ERROR;
+                    importStatus = ImportStatus.SKIPPED;
                 } catch (Exception e) {
                     log.error("Registry import failed", e);
                     importStatus = ImportStatus.ERROR;
--- a/app/src/main/java/io/apicurio/registry/ImportLifecycleBean.java
+++ b/app/src/main/java/io/apicurio/registry/ImportLifecycleBean.java
@@ -38,7 +38,7 @@ public class ImportLifecycleBean {
     @Inject
     AdminResourceImpl v3Admin;
 
-    private ImportStatus importStatus = ImportStatus.UNKNOWN;
+    private volatile ImportStatus importStatus = ImportStatus.UNKNOWN;
 
     void onStorageReady(@ObservesAsync StorageEvent ev) {
         if (StorageEventType.READY.equals(ev.getType())) {
--- a/app/src/test/java/io/apicurio/registry/ImportLifecycleBeanUnitTest.java
+++ b/app/src/test/java/io/apicurio/registry/ImportLifecycleBeanUnitTest.java
@@ -1,0 +1,40 @@
+package io.apicurio.registry;
+
+import io.apicurio.registry.rest.ConflictException;
+import io.apicurio.registry.rest.v3.impl.AdminResourceImpl;
+import io.apicurio.registry.storage.StorageEvent;
+import io.apicurio.registry.storage.StorageEventType;
+import io.apicurio.registry.storage.importing.ImportExportConfigProperties;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
+public class ImportLifecycleBeanUnitTest {
+
+    @Test
+    public void testConflictExceptionSetsStatusToSkipped() throws Exception {
+        ImportLifecycleBean bean = new ImportLifecycleBean();
+        bean.log = LoggerFactory.getLogger(ImportLifecycleBean.class);
+
+        // Mock ImportExportConfigProperties with a local temp file URL
+        bean.importExportProps = Mockito.mock(ImportExportConfigProperties.class);
+        java.io.File tempFile = java.io.File.createTempFile("dummy-import", ".zip");
+        tempFile.deleteOnExit();
+        bean.importExportProps.registryImportUrlProp = Optional.of(tempFile.toURI().toURL());
+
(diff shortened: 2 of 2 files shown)


app/src/main/java/io/apicurio/registry/ImportLifecycleBean.java
```diff
@@ -60,7 +60,7 @@ void onStorageReady(@ObservesAsync StorageEvent ev) {
                     importStatus = ImportStatus.ERROR;
                 } catch (ConflictException ce) {
                     log.info("Import skipped, registry not empty.");
-                    importStatus = ImportStatus.ERROR;
+                    importStatus = ImportStatus.SKIPPED;
```
paoloantinori, line 63: This status is written from the `@ObservesAsync` observer (separate thread) and read by `isReady()` on the health-check thread, but the field is not `volatile`. There's no happens-before edge between the two, so the readiness probe may not observe this `SKIPPED` value reliably (it could see a stale `UNKNOWN`/`ERROR`). Consider declaring `importStatus` as `volatile` so this fix reliably takes effect across threads — it's directly relevant to the PR's goal of correct readiness reporting.

paoloantinori (commented):
Diagnosis looks correct: a `ConflictException` on startup import (registry already populated) is an expected condition on restart with persistent storage, not a real error, so reporting `ERROR` broke the readiness probe. Aligning the status with the existing `"Import skipped, registry not empty."` log line is the right call, and `isReady()` already treats `SKIPPED` as ready, so the change is consistent.

- **Cross-thread visibility of `importStatus`.** Since the whole point of this PR is to make the readiness probe reflect the correct state, it's worth double-checking that the status change is actually visible to the probe thread. `onStorageReady` is a `@ObservesAsync` observer, so per the CDI spec it fires on a separate thread, while `isReady()` is invoked by the Quarkus/SmallRye readiness check on the request/health executor. `importStatus` is a plain (non-`volatile`) field with no synchronization, so there's no happens-before guarantee between the write in the async observer and the read in `isReady()`. In practice this means the readiness endpoint could observe a stale value (e.g. `UNKNOWN`, or even the old `ERROR`) indefinitely on some thread/JIT layouts. Making `private ImportStatus importStatus = ...` `volatile` would close that gap and make this fix reliable across restarts. This is pre-existing, but directly in scope for a PR whose goal is correct readiness reporting.
- The regression test is a good addition — it constructs the exact scenario (ConflictException on import) and asserts the behavioral contract (`isReady()`). One note: it bypasses CDI and sets the `@Inject` fields directly, which is fine for a focused unit test, but it won't exercise the async dispatch path mentioned above; if you ever want to cover that, an integration test that hits the readiness endpoint after a skipped import would be the place to do it (not blocking here).
- Minor, non-blocking: the temp file is created with `deleteOnExit()`, which only cleans up on JVM exit — acceptable for a test, just flagging it.

app/src/main/java/io/apicurio/registry/ImportLifecycleBean.java
kunaldevsahu: Thanks for the feedback  @paoloantinori  I've updated importStatus to volatile to ensure the state change is visible across threads.

PR #9033, opened 2026-07-30
fix(ci): split community-operators-prod into bundle and catalog PRs

carlesarnal:
## Summary

The community-operators-prod pipeline rejects PRs that modify both a bundle and catalog entries (#10596 failed with "Split operator and catalog changes into 2 separate pull requests"). The release workflow now creates two separate PRs:

1. **Bundle PR** (`release-<version>-bundle`) — only the `operators/<pkg>/<version>/` directory
2. **Catalog PR** (`release-<version>-catalog`) — FBC catalog-templates + regenerated catalogs

Both are regular PRs. The catalog PR is only created when FBC is enabled (`ci.yaml` contains `fbc:`).

## Test plan

- [ ] Next release should create two separate PRs on community-operators-prod
- [ ] Both PRs should pass their respective CI pipelines


.github/workflows/release-operator.yaml
```diff
@@ -344,88 +344,92 @@ jobs:
...
+          gh pr create --title "$TITLE" --body "$BODY" --base main --head "Apicurio:${RELEASE_BRANCH}-bundle"
+
+      - name: Warn if Openshift Community Operators Bundle PR failed
+        if: steps.openshift-community-operators-bundle-pr.outcome == 'failure'
+        run: echo "::warning::Failed to create Openshift Community Operators Bundle PR. Please create it manually."
 
-          # Update FBC catalog templates if FBC is enabled
+      - name: Create the Openshift Community Operators Catalog PR
+        id: openshift-community-operators-catalog-pr
+        continue-on-error: true
```
paoloantinori, line 371: This `continue-on-error: true` makes the catalog PR fully independent of the bundle PR's outcome. If `openshift-community-operators-bundle-pr` fails, this step still runs and can open a catalog PR referencing a bundle that was never submitted upstream. Adding `if: steps.openshift-community-operators-bundle-pr.outcome == 'success'` here would prevent an orphan catalog PR.

.github/workflows/release-operator.yaml
```diff
@@ -344,88 +344,92 @@ jobs:
...
-            else
-              CHANNELS="$MINOR_CHANNEL"
-            fi
+          if [ "$FBC_ENABLED" -eq 0 ]; then
+            echo "FBC is not enabled — skipping catalog PR"
+            exit 0
+          fi
 
-            cd "operators/$PKG_NAME"
+          git checkout main
```
paoloantinori, line 381: Checking out `main` means the new `operators/<pkg>/<version>/` bundle (committed only on the `-bundle` branch) isn't in the working tree when `make catalogs` / `make validate-catalogs` run below. Worth double-checking that `validate-catalogs` doesn't assert bundle presence for entries it finds in the catalog templates.

paoloantinori (commented):
Clean split — moving the bundle and FBC catalog updates into separate PRs directly addresses the community-operators-prod "split operator and catalog" rejection (#10596), and the per-PR `continue-on-error` + warning pattern stays consistent with the surrounding steps.

A couple of operational observations worth considering:

- **The catalog PR is not gated on the bundle PR.** Both steps use `continue-on-error: true` independently. If the bundle `gh pr create` fails but the catalog step still succeeds, you'll open a catalog PR upstream that references a bundle version not yet present in the repo (the bundle only lives on the `-bundle` branch). community-operators-prod CI is likely to reject such a catalog PR. Consider adding `if: steps.openshift-community-operators-bundle-pr.outcome == 'success'` to the catalog step (or guarding just its `gh pr create`) so a failed bundle PR doesn't produce an orphan catalog PR.
- **`make validate-catalogs` runs on a tree without the new bundle.** The catalog step checks out `main`, so the newly added `operators/<pkg>/<version>/` directory isn't in the working tree when catalogs are regenerated/validated. If `validate-catalogs` cross-checks catalog entries against on-disk bundles, it could flag the new version. Worth confirming it only validates catalog schema/internal consistency and not bundle presence.
- Minor: `cat ci.yaml | grep -c "fbc:"` could be `grep -c "fbc:" ci.yaml`, but it's unchanged from before so no need to touch it in this PR.

.github/workflows/release-operator.yaml
```diff
@@ -344,88 +344,93 @@ jobs:
...
           cp -r "../registry/operator/target/bundle/apicurio-registry-3/$PACKAGE_VERSION" operators/apicurio-registry-3
+          git add .
+          git commit -s -m "$TITLE"
+          git push -f source "${RELEASE_BRANCH}-bundle"
+          gh repo set-default redhat-openshift-ecosystem/community-operators-prod
+          gh pr create --title "$TITLE" --body "$BODY" --base main --head "Apicurio:${RELEASE_BRANCH}-bundle"
+
+      - name: Warn if Openshift Community Operators Bundle PR failed
+        if: steps.openshift-community-operators-bundle-pr.outcome == 'failure'
+        run: echo "::warning::Failed to create Openshift Community Operators Bundle PR. Please create it manually."
```
paoloantinori, line 366: When `openshift-community-operators-bundle-pr` fails, this `if` causes the catalog step to be skipped entirely with no warning. Consider either removing this guard (letting the catalog step run and emit its own warning on failure) or adding a sibling warning step conditioned on `outcome == 'failure'` so the release operator knows the catalog PR also needs manual creation.

paoloantinori (commented):
The split reads cleanly — bundle and FBC catalog updates landing in separate PRs is exactly what community-operators-prod wants, and the per-step `continue-on-error` + warning pattern is a reasonable fallback to manual creation. One operational gap worth considering before the next release runs through this.

- **Catalog PR is skipped silently when the bundle PR fails.** The catalog step is gated on `if: steps.openshift-community-operators-bundle-pr.outcome == 'success'`, so if the bundle step fails, the catalog step is *not* marked as skipped-with-warning — it simply never runs, and no warning fires for it. The release operator sees only the bundle warning (`Failed to create Openshift Community Operators Bundle PR. Please create it manually.`) and gets no signal that the catalog PR also needs to be created manually. It's worth either dropping the `if` so the catalog step attempts to run independently (its own `continue-on-error` + warning would then fire), or adding a dedicated `if: steps.openshift-community-operators-bundle-pr.outcome == 'failure'` warning that explicitly tells the operator both PRs need manual creation.

- Minor: the catalog step's `FBC_ENABLED` check exits `0` when FBC is off, which is correct (no-op success), but worth a one-line comment noting that this early `exit 0` is the intended "skip catalog" path so future readers don't read it as a failure.

Neither blocks the merge — the change is a clear improvement over the combined-PR flow that was getting rejected.

Issue #9008, opened 2026-07-29
Operator CSV should split namespace-scoped and cluster-scoped permissions for least-privilege single-namespace installs


jsenko:
## Summary

The operator CSV currently puts all RBAC rules into `clusterPermissions` (via the `rbac/cluster/cluster-role.yaml` kustomize overlay). OLM always creates a ClusterRole + ClusterRoleBinding for `clusterPermissions`, regardless of the OperatorGroup's target namespace scope.

This means a single-namespace install (OwnNamespace/SingleNamespace `installMode`) gets cluster-wide RBAC for workload resources (Deployments, Services, Pods, ConfigMaps, Secrets, etc.), even though the operator only reconciles CRs in one namespace. This is over-permissioned for single-namespace deployments.

## Current state

- `operator/controller/src/main/deploy/rbac/cluster/cluster-role.yaml` — single ClusterRole with all permissions
- CSV kustomization includes `../rbac/cluster` — all rules end up in `clusterPermissions`
- `installModes` declares all four modes as supported, but RBAC doesn't adapt to the mode
- #8193 addressed wildcard verbs (`['*']` → explicit verb lists) but did not split cluster vs namespace scope

### 1. OLM installs — split CSV permissions

Split the CSV permissions into two tiers:

**`clusterPermissions`** — only truly cluster-scoped resources:
- `registry.apicur.io` CRDs (watch/list/get across all namespaces to discover CRs)
- `rbac.authorization.k8s.io` ClusterRoles/ClusterRoleBindings (if the operator manages these for its operands)

**`permissions`** — namespace-scoped workload resources:
- `apps/deployments`
- Core resources: `pods`, `services`, `configmaps`, `serviceaccounts`, `secrets`
- `networking.k8s.io/ingresses`, `networkpolicies`
- `policy/poddisruptionbudgets`

OLM creates Role + RoleBinding in the target namespace(s) for `permissions`, so single-namespace installs get namespace-scoped access to workload resources, while cluster-scoped CRD access is retained.

### 2. Non-OLM (manifest) installs — ship both variants

For users installing from manifest files directly (no OLM), provide two kustomize overlays:

- **`rbac/cluster`** (existing) — ClusterRole + ClusterRoleBinding for all-namespaces installs. This is what we ship today.
- **`rbac/namespaced`** (new) — ClusterRole for CRDs only + namespace-scoped Role/RoleBinding for workload resources. For single-namespace installs.

[...]

PR #9035, opened 2026-07-30
feat(operator): split CSV RBAC into namespace and cluster tiers

Taimoo-r:
## Summary

Splits the operator CSV permissions so single-namespace installs get
least-privilege RBAC. Part 1 of #9008 (the OLM path). The manifest install
overlay and workflow/doc wiring will follow in a separate PR.

Refs https://github.com/Apicurio/apicurio-registry/issues/9008

## Root Cause

The CSV kustomization pulled all RBAC from `rbac/cluster`, a single ClusterRole
holding every rule. `operator-sdk generate bundle` turns any ClusterRole into
CSV `clusterPermissions`, and OLM always materializes `clusterPermissions` as a
ClusterRole + ClusterRoleBinding regardless of install mode. So a
SingleNamespace/OwnNamespace install still received cluster-wide access to
workload resources (Deployments, Services, Pods, ConfigMaps, Secrets, and more),
even though the operator only reconciles CRs in one namespace.

## Changes

New `operator/controller/src/main/deploy/rbac/namespaced/` overlay, bound to the
operator ServiceAccount, split by scope:

- `cluster-role.yaml` (becomes `clusterPermissions`): only cluster-scoped, read-only
  rules - `apicurioregistries3` `get/list/watch` for the informer, and `get` on the CRD.
  `/status` is not included here: there is no list/watch endpoint for a status
  subresource, so the informer only needs the parent; `/status` keeps `patch`/`update`
  in `role.yaml`.
- `role.yaml` (becomes `permissions`): workload rules (deployments, pods, services,
  configmaps, serviceaccounts, events, secrets, ingresses, networkpolicies,
  poddisruptionbudgets, horizontalpodautoscalers, leases, roles/rolebindings), plus
  the CR mutating verbs `patch`/`update` on `apicurioregistries3` (+`/status`). The
  operator never creates or deletes the CR, so those verbs are omitted.
- `cluster-role-binding.yaml`, `role-binding.yaml`, `kustomization.yaml`.
- `csv/kustomization.yaml` references `../rbac/namespaced` instead of `../rbac/cluster`.

How OLM materializes `permissions` depends on the OperatorGroup, not the CSV: for
SingleNamespace/OwnNamespace/MultiNamespace it creates a Role + RoleBinding in the
target namespace(s) (workload verbs stay namespace-scoped); for AllNamespaces (`*`)
it promotes each rule to a ClusterRole + ClusterRoleBinding (cluster-wide, as before
this split). The read-only CR discovery verbs stay least-privilege in every mode.

`rbac/cluster` is intentionally unchanged, so the manifest install path stays
cluster-wide until the follow-up PR.

Notes on scoping (verified against the controller code):

- `secrets` is `[create, get]` only: the operator just creates and reads a secret by
  name (`SqlCRUpdater`), never patch/update/delete/list/watch.
- `roles`/`rolebindings` are namespace-scoped: the operator only ever creates a
  namespaced Role and RoleBinding for its operand (`AppRoleResource`,
  `AppRoleBindingResource`), never a ClusterRole.

Keeping the OLM v1 installer role in sync:

- `olm-tests/.../olmv1/cluster-role.yaml` (the installer ClusterRole) dropped
  `create`/`delete` on `apicurioregistries3` to match the operator's runtime perms, and
  mirrors the cluster/namespace `/status` split above.
- `RbacInstallerSyncTest` (controller unit test) parses `rbac/namespaced` and the OLM
  v1 installer ClusterRole into `(apiGroup, resource, verb)` tuples and asserts the
  installer is a superset, guarding against future drift.
- `RbacSplitTest` (controller unit test) guards the generated CSV split itself: since
  `operator-sdk` maps the ClusterRole to `clusterPermissions` and the Role to
  `permissions` verbatim, it asserts the source partition (cluster tier stays read-only
  CR/CRD discovery; namespace tier holds the workloads and the CR mutation verbs, and no
  cluster-scoped discovery), so a refactor cannot silently collapse the split or drop the
  CR `patch`/`update` rules unnoticed.

Handling OLM v1:

OLM v1 dropped the user-facing OperatorGroup, but its bundle renderer kept v0's promotion mechanic:
in AllNamespaces mode it promotes the CSV `permissions` to a ClusterRole + ClusterRoleBinding and
appends `namespaces [get,list,watch]` (mirroring v0's operatorgroup.go). This bundle currently always
resolves to AllNamespaces (single/own-namespace needs the alpha `SingleOwnNamespaceInstallSupport`
gate, off by default), so under v1 the operator receives its workload verbs cluster-wide via
promotion, automatically. One change makes that promotion succeed:

- The installer ClusterRole gains `namespaces [get,list,watch]`. OLM v1's escalation prevention
  blocks the installer from granting a permission it does not hold, so without this rule the promoted
  ClusterRole is rejected, the operator's informers 403 at the cluster scope, and it CrashLoopBackOffs.
  `RbacInstallerSyncTest` asserts this promotion-injected rule alongside the declared workloads, so the
  gap is caught deterministically instead of only by the live smoke run.

No manual admin grant is applied in the tests: v0's OperatorGroup and v1's AllNamespaces promotion each
grant the workloads cluster-wide, so `SmokeOLMITTest` installs the bundle and waits with no extra RBAC,
and a green v1 run genuinely exercises the promotion.

The single-namespace v1 story, where the operator watches only its own namespace so the CSV's
namespaced Role suffices, is deferred to the Part 2 / `watched-namespaces` follow-up.

Tests:

- `OLMITBase`: make the OperatorGroup resource overridable; add `getOlmVersion()`.
- `operator-group-all-namespaces.yaml`: AllNamespaces OperatorGroup overlay.
- `AllNamespacesOLMITTest`: for an AllNamespaces install, a SubjectAccessReview confirms
  the operator ServiceAccount may create Deployments in another namespace (the promoted
  cluster-wide grant), and the operator reconciles the operand Deployment there. It
  asserts the operand Deployment is created rather than waiting for pod readiness, so it
  does not depend on the app image.
- `NamespacedPermissionsOLMITTest`: for an OwnNamespace install, a SubjectAccessReview
  confirms the operator ServiceAccount may create Deployments in its target namespace but
  not in a foreign one.
- Both new ITs are gated to OLM v0 with a class-level `@DisabledIfSystemProperty` on
  `test.operator.olm-version`, since OperatorGroups/AllNamespaces are a v0 concept and the
  v1 installer grants workloads cluster-wide. An earlier version gated with `assumeTrue`
  inside the test method instead; under OLM v1 that let `@BeforeAll` start the async v1
  install before the assumption aborted, and `@RetryTest` retried the aborted test, whose
  cleanup then 404'd deleting a CR of a type the cluster hadn't registered yet. The
  class-level annotation skips `@BeforeAll` entirely, so the v1 install never starts for
  these two classes.
- Fixed the `OML_VERSION` -> `OLM_VERSION` typo in `OLMITBase`.
- `SmokeOLMITTest` needs no v1-specific RBAC setup: v1's AllNamespaces promotion grants the
  operator its workload verbs cluster-wide, so a green v1 run exercises the promotion directly.

Docs: `operator/README.md` gains an RBAC section describing the scope split, the
OperatorGroup-driven promotion (and the resulting AllNamespaces asymmetry), and the
manual sync with the OLM v1 installer role backed by `RbacInstallerSyncTest` and
`RbacSplitTest`. A new `operator/olm-tests/README.md` documents the OLM v0-vs-v1 RBAC-scope
model, the single/multi/all-namespace scenarios, the installer superset, and how the tests
model each.

## Test plan

- [x] Added `AllNamespacesOLMITTest` (AllNamespaces operand deploy)
- [x] Added `NamespacedPermissionsOLMITTest` (SubjectAccessReview negative boundary)
- [x] Added `RbacInstallerSyncTest` (installer-superset drift guard, no cluster needed)
- [x] Added `RbacSplitTest` (generated-CSV split partition guard, no cluster needed)
- [ ] Operator OLM v0 CI passes (both install modes deploy the operand)
- [ ] Operator OLM v1 CI passes (installer ClusterRole still applies the bundle)
- [x] Manual: generated CSV shows both `permissions` and `clusterPermissions` with the intended verb split


Taimoo-r:
Hi @jsenko, this is Part 1 of https://github.com/Apicurio/apicurio-registry/issues/9008 (the OLM path).

It splits the CSV RBAC into `clusterPermissions` (only cluster-scoped CR/CRD discovery) and `permissions` (namespace-scoped workloads), so single-namespace installs stop getting cluster-wide workload access. `rbac/cluster` is left untouched, so the manifest install path is unchanged for now.

Tests follow your guidance from the issue:
- `AllNamespacesOLMITTest` covers the AllNamespaces mode by deploying the operand into a namespace other than the operator's own.
- `NamespacedPermissionsOLMITTest` is the negative test: for a SingleNamespace install, a SubjectAccessReview confirms the operator ServiceAccount can create Deployments in its target namespace but not in a foreign one. I used a SubjectAccessReview rather than forcing the operator to watch a foreign namespace, since under OLM the watched namespaces and generated RBAC stay in sync via the OperatorGroup. Happy to switch to a watch-based behavioral test if you'd prefer.

Part 2 (the `rbac/namespaced` manifest overlay wired into `install.yaml`, the workflows, and docs) will come as a separate PR, as we discussed.

operator/controller/src/main/deploy/rbac/namespaced/cluster-role.yaml
```diff
@@ -0,0 +1,33 @@
...
+rules:
+
+  - apiGroups:
+      - registry.apicur.io
+    resources:
+      - apicurioregistries3
+      - apicurioregistries3/status
+    verbs:
+      - create
+      - delete
```
paoloantinori: These mutating verbs (`create/delete/patch/update`) on `apicurioregistries3` at cluster scope let the operator SA create or delete registry CRs in any namespace. Since the CR is namespace-scoped, these arguably belong in the namespace `role.yaml` for a least-privilege SingleNamespace install. Is cluster-wide `get/list/watch` (for the informer) sufficient here, with the mutating verbs moved to the Role?

operator/controller/src/main/deploy/rbac/namespaced/role.yaml
```diff
@@ -0,0 +1,113 @@
...
+      - get
+      - list
+      - patch
+      - update
+      - watch
+
+  - apiGroups:
+      - ''
+    resources:
+      - secrets
```
paoloantinori, line 116: `secrets` is the only workload resource limited to `[create, get]`. If the operator ever reconciles an existing secret it created (rotation/refresh) or watches a pre-existing one, the missing `update/patch/delete/list/watch` will silently deny. Worth confirming the secret lifecycle is genuinely create-once-read-only, or widening to match the other resources.

paoloantinori (commented):
Clean structural split and the `SubjectAccessReview`-based negative test is a nice way to assert the boundary deterministically. Two least-privilege points worth confirming before this lands:

- **The cluster-role grants full CRUD (`create/delete/patch/update` + `get/list/watch`) on `apicurioregistries3` at cluster scope.** `apicurioregistries3` is a namespace-scoped CR, so a ClusterRole on it lets the operator SA create/delete/update registry CRs in *any* namespace — which partially defeats the least-privilege goal of a SingleNamespace install. The PR description frames `cluster-role.yaml` as "only truly cluster-scoped resources" + CR discovery, but the mutating verbs on the CR are broader than discovery. Worth confirming whether the controller actually needs cluster-wide `create/delete/patch/update` on the CR, or whether `get/list/watch` at cluster scope (if needed for the informer) plus the mutating verbs in the namespace `role.yaml` would be tighter.

- **`secrets` narrowed to `[create, get]`.** The other workload resources get the full 7-verb set, but secrets only allow create+get — no `update`, `patch`, `delete`, `list`, or `watch`. If the operator ever needs to reconcile/update an existing secret it created (credential rotation, TLS refresh) or watch one it didn't create, this will silently deny. Worth confirming the operator's secret lifecycle never needs update/delete/watch, or widening if it does.

Minor: the "IMPORTANT: also update olmv1/cluster-role.yaml" comment in two files flags a manual sync burden — a divergence would silently change the effective permissions. If there's no mechanism keeping them in sync, a brief note in the contributing/operator docs would help future contributors.

Taimoo-r:
Hi @paoloantinori, thanks for the careful review. Addressed in 0f47118:

**1. CR mutating verbs at cluster scope.** Good call. I checked what the controller actually does to the primary CR: it only `patch`es it (spec migration in `SqlCRUpdater`, and status via `UpdateControl.patchStatus`) and `update`s it for the JOSDK finalizer (the reconciler implements `Cleaner`). It never creates or deletes the `apicurioregistries3` CR. So instead of just moving all four verbs, I kept only cluster-wide `get/list/watch` on the CR (for the informer) and moved `patch/update` into the namespaced `role.yaml`. `create`/`delete` are dropped entirely since they are unused. In a SingleNamespace install the operator can now only mutate CRs in its own namespace; in AllNamespaces mode OLM promotes the Role to a ClusterRole so it still works cluster-wide.

**2. secrets limited to [create, get].** Confirmed this is sufficient and intentional. The operator's only secret operations are a `get`-by-name and a `create`, both in the primary's namespace (`SqlCRUpdater` creating the migrated datasource-password secret). It never updates, deletes, lists, or watches secrets, so the missing verbs would not be exercised. This scope is also unchanged from the original ClusterRole (it predates this PR). Happy to widen if you'd rather be defensive about future secret reconciliation, but today it is genuinely create-once + read-by-name.

**3. Manual sync burden.** Added an `RBAC` section to `operator/README.md` documenting the scope split and calling out that the rules are duplicated transitively in `olm-tests/src/test/deploy/olmv1/cluster-role.yaml`, with no automated sync, so changes must be applied in both places. The in-file `IMPORTANT` comments remain as the local reminder.

I also re-generated the bundle with `operator-sdk generate bundle` to confirm the CSV comes out with the intended split: `clusterPermissions` has `apicurioregistries3`/`.../status` at `get/list/watch` plus the CRD `get`, and `permissions` has the workloads plus `apicurioregistries3` at `patch/update`.

paoloantinori:
Thanks @Taimoo-r — the revised split looks right. Keeping only `get/list/watch` for `apicurioregistries3`(+`/status`) in `cluster-role.yaml` (informer discovery) while moving the actual mutations (`patch`/`update` for spec migration, status, and the JOSDK finalizer) into the namespace-scoped `role.yaml` is exactly the boundary I was hoping to see. The `secrets` grant trimmed to `create,get` is a nice touch too.

The two ITs cover both modes cleanly: `AllNamespacesOLMITTest` proves the operand deploys in a non-install namespace (cluster-wide workload perms work), and `NamespacedPermissionsOLMITTest` asserts the boundary deterministically via `SubjectAccessReview`. From my side this is good to go.

jsenko (changes requested):
Thanks @Taimoo-r — this is a clean implementation of #9008 and the RBAC split is structurally correct. paoloantinori's two concerns are well-addressed in `0f47118`. A few things to resolve before merging:

**1. OLM v1 installer ClusterRole not synced (should-fix)**

`olm-tests/src/test/deploy/olmv1/cluster-role.yaml` still has `create/delete` on `apicurioregistries3` in the transitive operator permissions section. Those verbs were dropped from the OLM v0 path in your latest commit (correctly — the operator never creates or deletes CRs). The OLM v1 file should be updated to match.

Note: this is the *installer* SA's permissions, not the operator's runtime RBAC (OLM v1 still reads the CSV for the operator's runtime Role/ClusterRole). So the impact is lower — the installer has more permissions than needed, not fewer — but keeping them in sync avoids confusion and matches the intent.

**2. Automation to prevent future drift (should-fix)**

The manual sync between `rbac/namespaced/` and `olmv1/cluster-role.yaml` is a real divergence risk, as paoloantinori flagged and you documented in the README. Two options — pick whichever fits better:

- **Option A: Generate + CI check.** A Makefile target that generates the OLM v1 installer ClusterRole by merging the namespaced files with the installer-specific rules, plus a CI step that diffs the generated output against the committed file. There are two existing patterns in the repo for this: the `ref-registry-all-configs.adoc` config doc generation (scans Jandex for `@ConfigProperty`/`@Info`, generates the doc, CI verifies freshness), and the `install.yaml` generation in the operator Makefile. Either could serve as a template.
- **Option B: Unit test.** A test that loads the `rbac/namespaced/` YAML files and the `olmv1/cluster-role.yaml`, parses them into `(apiGroup, resource, verb)` tuples, and asserts the OLM v1 file contains a superset of all tuples from the namespaced files. ~50-80 lines. The main subtlety is `resourceNames` scoping (no `resourceNames` = wildcard, which is a superset of any named restriction), but it's manageable.

**3. New tests may break or give misleading results in OLM v1 mode (should-fix)**

CI runs all `@Tag(OLM)` tests in both OLM v0 and v1 modes (via the `-Dtest.operator.olm-version=1` matrix entry). The new tests have OLM v0-specific assumptions:

- `AllNamespacesOLMITTest` overrides `getOperatorGroupResourcePath()` to use `olmv0/operator-group-all-namespaces.yaml`, but in v1 mode `setupOLMResources()` takes the v1 branch and ignores `getOperatorGroupResourcePath()` entirely (OperatorGroups are an OLM v0 concept). The test would run in v1 mode with default semantics, not actually testing AllNamespaces behavior.

- `NamespacedPermissionsOLMITTest` asserts that creating a Deployment in a foreign namespace is forbidden via `SubjectAccessReview`. But in OLM v1, the installer ClusterRole grants everything cluster-wide, so the review would likely return `allowed=true` and the test would **fail**.

Either skip the new tests in v1 mode (e.g., check the OLM version property in a `@BeforeAll` and call `Assumptions.assumeTrue(olmVersion == 0)`) or add v1-specific test variants.

**4. CR permissions split — verified correct (no action needed)**

Confirmed the operator only `patch`es the CR (spec migration, status updates) and `update`s for the JOSDK finalizer — never `create/delete`. The cluster-scoped `get/list/watch` for the informer + namespace-scoped `patch/update` is the right split. OLM handles the AllNamespaces promotion automatically: when the OperatorGroup targets all namespaces, OLM promotes the `permissions` Role to a ClusterRole + ClusterRoleBinding, so the split works in both modes.

**5. Secrets `[create, get]` — verified correct (no action needed)**

Confirmed the operator only does `get`-by-name and a one-time `create` in `SqlCRUpdater`. No update/watch/delete needed. Unchanged from the original ClusterRole.

**6. Fix `OML_VERSION` → `OLM_VERSION` typo (nice-to-have)**

`OLMITBase.java` line 41: `public static final String OML_VERSION` — `M` and `L` are transposed. Pre-existing, but since you're touching this class anyway, easy to fix.

Taimoo-r:
Hi @jsenko, thanks for the review. All four actionable points are addressed in `1d848b0`:

**1. olmv1 installer synced.** Dropped `create`/`delete` on `apicurioregistries3` in the transitive section of `olmv1/cluster-role.yaml`, matching the operator's runtime perms (cluster `get/list/watch` + namespaced `patch/update`).

**2. Drift guard (Option B).** Added `RbacInstallerSyncTest`: parses `rbac/namespaced` (ClusterRole + Role) and the olmv1 installer ClusterRole into `(apiGroup, resource, verb)` tuples and asserts the installer is a superset. It ignores `resourceNames` (a rule without them is a wildcard superset), so no false failures. Plain unit test, no cluster, passes locally.

**3. Tests gated to v0.** Both new ITs now `assumeTrue(getOlmVersion() == 0)`, since OperatorGroups/AllNamespaces are v0 and the v1 installer grants workloads cluster-wide. Added `getOlmVersion()` to `OLMITBase` and folded the duplicate in `ChannelValidationOLMITTest` into it.

**6.** `OML_VERSION` -> `OLM_VERSION` fixed.

4 & 5: thanks for confirming the CR-verb and secrets scoping.

paoloantinori:
Thanks @Taimoo-r — `1d848b0` addresses the points cleanly:

1. **olmv1 installer synced** — dropping `create`/`delete` on `apicurioregistries3` in the transitive section so the installer matches the operator's runtime perms (cluster `get/list/watch` + namespaced `patch/update`) is the right fix.
2. **Drift guard (Option B)** — `RbacInstallerSyncTest` parsing both ClusterRole/Role and the olmv1 installer into `(apiGroup, resource, verb)` tuples and asserting superset, while ignoring `resourceNames` to avoid false failures, is exactly the kind of test that catches this class of drift. Plain unit test, no cluster — good.
3. **Tests gated to v0** — `assumeTrue(getOlmVersion() == 0)` for the OperatorGroups/AllNamespaces ITs, with `getOlmVersion()` consolidated into `OLMITBase` (and folding the duplicate in `ChannelValidationOLMITTest`), is the correct gating since v1 installer grants workloads cluster-wide.

Nothing further from my side. This is solid — deferring the merge call to a maintainer.

operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/AllNamespacesOLMITTest.java
```diff
@@ -0,0 +1,74 @@
...
+            var registry = ResourceFactory.deserialize("/k8s/examples/simple.apicurioregistry3.yaml",
+                    ApicurioRegistry3.class);
+            registry.getMetadata().setNamespace(otherNamespace);
+            registry.getSpec().getApp().getIngress().setHost(ingressManager.getIngressHost("app"));
+            registry.getSpec().getUi().getIngress().setHost(ingressManager.getIngressHost("ui"));
+
+            client.resource(registry).create();
+
+            // In AllNamespaces mode the operator watches every namespace and has cluster-wide
+            // workload permissions, so the operand Deployment is created and becomes ready.
```
paoloantinori: This comment (and the class javadoc) assumes AllNamespaces mode gives the operator cluster-wide workload permissions, but after this split the workload verbs (deployments, services, secrets, etc.) live in `permissions` (namespaced/role.yaml), not in `clusterPermissions` (namespaced/cluster-role.yaml, which only carries get/list/watch on the CR and get on the CRD). OLM materializes CSV `permissions` as a Role+RoleBinding in the OperatorGroup's namespace only — not in every watched namespace — so in AllNamespaces mode the operator would lack permission to create the operand Deployment in `otherNamespace` and this assertion would fail. Before this PR all verbs were cluster-scoped, so AllNamespaces worked; the CSV `installModes` aren't changed in the diff, so the mode is still advertised but now under-privileged. The OLM v0 CI box is still unchecked, so this path may be unvalidated. Could you confirm this test passes against OLM v0, and if not, decide whether workload verbs must stay cluster-scoped for AllNamespaces (or whether AllNamespaces support should be dropped from the CSV installModes)?

operator/controller/src/main/deploy/rbac/namespaced/role.yaml
```diff
@@ -0,0 +1,126 @@
...
+    verbs:
+      [ create, get ]
```
paoloantinori: Every other workload resource in this Role gets full CRUD, but `secrets` is intentionally limited to `[ create, get ]`. If the operator ever patch/update/deletes a Secret it created (e.g. rotating SQL datasource credentials or other operand secrets), this would now 403 under the namespaced tier. Worth confirming against the operand resource code that the operator only ever creates-and-reads secrets; if it mutates them, add the matching verbs here and in olm-tests/.../olmv1/cluster-role.yaml.

paoloantinori (commented):
Overall: the scope split correctly narrows SingleNamespace/OwnNamespace installs to least-privilege, and the SubjectAccessReview boundary test plus the installer-superset drift guard are well-designed additions. The main concern is that a single CSV overlay can't satisfy AllNamespaces mode — the workload verbs were moved into namespace-scoped `permissions`, which OLM applies only in the operator's own namespace under AllNamespaces, so cross-namespace operand management (and the new `AllNamespacesOLMITTest`) likely regress versus the prior all-cluster-scoped RBAC.

- **AllNamespaces regression (role.yaml + cluster-role.yaml + AllNamespacesOLMITTest.java:61-62):** Workload verbs (deployments, services, secrets, etc.) now live in `permissions` (namespaced/role.yaml), while `clusterPermissions` (namespaced/cluster-role.yaml) only carries get/list/watch on the CR and get on the CRD. OLM materializes CSV `permissions` as a Role+RoleBinding in the OperatorGroup namespace only — not in every watched namespace — so in AllNamespaces mode the operator would 403 when creating the operand Deployment in `otherNamespace`, and this assertion would fail. Before this PR every verb was cluster-scoped, so AllNamespaces worked; the CSV `installModes` aren't changed in the diff, so the mode is still advertised but is now under-privileged. The OLM v0 CI checkbox is still unchecked, so this path may be unvalidated (unless the operator self-grants cluster-wide RBAC at runtime, which isn't shown). Worth confirming the test passes against OLM v0 before merge.
- **secrets reduced to create/get only (role.yaml:110-115):** every other workload resource in this Role gets full CRUD while `secrets` is deliberately limited to `[ create, get ]`. Please confirm against the operand resource code that produces secrets (e.g. SQL datasource credentials) that the operator only ever creates-and-reads them; if it ever patch/update/deletes a Secret it created, add the matching verbs here and in olm-tests/.../olmv1/cluster-role.yaml.
- **OLM_VERSION constant rename (OLMITBase.java:41):** `OML_VERSION` → `OLM_VERSION` renames a `public static final` field; the diff only adjusts `ChannelValidationOLMITTest`, so confirm no other class references the old name.
- On the positive side: `NamespacedPermissionsOLMITTest` uses a SubjectAccessReview for a clean deterministic boundary check, and `RbacInstallerSyncTest` is a useful drift guard — note it ignores `resourceNames`, so it can't catch a resourceName-scoped mismatch (acceptable for this permission set).

Taimoo-r:
Hi @paoloantinori, thanks — good scrutiny on the AllNamespaces path. I checked the OLM OperatorGroup RBAC behavior and the split holds in both modes:

- **AllNamespaces:** when the OperatorGroup targets all namespaces (`*`), OLM promotes each CSV `permissions` rule to a **ClusterRole + ClusterRoleBinding**, so the operator gets the workload verbs cluster-wide. This is what lets it create the operand in `otherNamespace`.
- **SingleNamespace/OwnNamespace:** OLM copies those same Roles/RoleBindings into each target namespace (not just the install namespace).

So the workload verbs in `permissions` are correctly scoped per install mode, and `installModes` stays accurate. I added a comment on `AllNamespacesOLMITTest` (`bcbf72c`) spelling this out. The OLM v0 CI leg runs this test on a cluster, so it's exercised there.

- **secrets:** confirmed against the operand code — the only secret operations are a `get`-by-name and a one-time `create` in `SqlCRUpdater` (the migrated SQL datasource-password secret). No patch/update/delete anywhere, so `[create, get]` is complete.
- **`OLM_VERSION` rename:** confirmed no other references to the old name remain.

Good call on `RbacInstallerSyncTest` ignoring `resourceNames` — that's the documented tradeoff; acceptable for this set.

Reference: [OLM OperatorGroup docs](https://olm.operatorframework.io/docs/concepts/crds/operatorgroup/)

paoloantinori:
Thanks @Taimoo-r — that's a useful walkthrough and the OLM promotion mechanic you describe is correct: when an OperatorGroup targets `*`, OLM materializes each CSV `permissions` rule as a `ClusterRole`+`ClusterRoleBinding`, so the namespace-scoped workload verbs in `role.yaml` become cluster-wide in AllNamespaces mode. That's exactly why the `AllNamespacesOLMITTest` (operand deploys in a non-target namespace) is a worthwhile positive test for this mode.

Two things I'd want pinned down so the RBAC story is unambiguous on the review record:

1. **The promotion is a property of the OperatorGroup, not the CSV.** The split in `role.yaml`/`cluster-role.yaml` is correct precisely *because* OLM, not the CSV, decides promotion based on the install mode. Worth a one-line comment in `role.yaml` noting that the same rules become either Role or ClusterRole depending on the OperatorGroup target, so a future reader doesn't assume `permissions` is always namespace-scoped. (The `AllNamespacesOLMITTest` Javadoc already captures this well — mirroring that note into the role file would close the loop.)

2. **No `apicurioregistries3` mutating verbs leak via promotion.** In AllNamespaces mode the `permissions` rules are promoted, and `role.yaml` carries `patch`/`update` on `apicurioregistries3`(+`/status`). After promotion that mutation becomes cluster-wide — which is acceptable for the operator's status/finalizer writes, but it does mean the least-privilege argument ("CR mutation stays namespace-scoped") only holds for SingleNamespace/OwnNamespace, not AllNamespaces. The `cluster-role.yaml` correctly stays read-only, so the discovery verbs are least-privilege everywhere. Just want the asymmetry acknowledged so the `README.md` "RBAC" section isn't read as a blanket least-privilege claim.

Neither blocks merge from my side — the structural split, the `SubjectAccessReview` boundary test, and the installer-superset drift guard are all solid. These are clarifications for the record.

Taimoo-r:
Thanks @paoloantinori — both captured for the record in `0c638d6`:

1. Added a comment in `role.yaml` noting these rules become a Role *or* a ClusterRole depending on the OperatorGroup (OLM decides promotion by install mode), so `permissions` isn't assumed to be always namespace-scoped.
2. Updated the README "RBAC" section to state the asymmetry explicitly: the namespace-scoped least-privilege guarantee applies to single-namespace installs; AllNamespaces is intentionally cluster-wide (as before the split), while the read-only CR discovery verbs in `cluster-role.yaml` stay least-privilege in every mode.

I've also refreshed the PR description to match the final state (the CR verb split, the OLM v1 installer sync + `RbacInstallerSyncTest`, the v0 gating, and the docs), so nothing there is stale.

paoloantinori:
Thanks @Taimoo-r — both captures in `0c638d6` look right.

The `role.yaml` comment correctly documents that `permissions` is materialized as a Role *or* a ClusterRole depending on the OperatorGroup, so readers won't assume `permissions` is always namespace-scoped. And the README "RBAC" section now states the asymmetry explicitly: the least-privilege / namespace-scoped guarantee applies to single-namespace installs, while AllNamespaces is intentionally cluster-wide (unchanged from before). The pointer to keep `olmv1/cluster-role.yaml` in sync manually, with the `RbacInstallerSyncTest` drift guard backing it, closes the loop well.

From my side this is good. Over to @jsenko for the final merge call.

paoloantinori (commented):
All three review agents have reported. Here's my synthesis after applying the skill's confidence filter and my own adversarial check on each finding:

**Dismissed as false positives / non-issues:**
- Agent 2's "AllNamespaces split is broken" — based on a misreading of OLM; AllNamespaces *promotes* Role rules to ClusterRole+ClusterRoleBinding (cluster-wide), which is exactly what the PR documents and what `AllNamespacesOLMITTest` exercises. Not a defect.
- Agent 1's "AllNamespaces promotion is over-privileged" — already documented as intentional and unchanged from prior behavior.
- Binding manifests in the CSV overlay — `operator-sdk` derives `permissions`/`clusterPermissions` from Role/ClusterRole and ignores the extra bindings; they use `${PLACEHOLDER_NAMESPACE}` consistent with the existing pattern. Harmless.
- Agent 3 found no MAJOR+ test-logic bugs (default-OG concern is internally consistent and fails loudly; the `OML_VERSION` rename is a loud compile break caught by CI; CWD/resourceNames handling is standard).

**Downgraded:** Agent 1's "`/finalizers` subresource will break the operator" overclaims — default JOSDK patches `metadata.finalizers` via the main CR, which `patch`/`update` (present) already authorizes. Kept only as a hedged easy-win.

**Surviving (the RBAC verbs whose correctness rests on controller code not in the diff):** the non-idempotent `secrets: [create, get]` (strongest — convergent across both agents + my own analysis), the finalizer-subresource edge case, and the roles/rolebindings cluster-scope confirmation. Producing the mandated JSON.

Taimoo-r:
Thanks @paoloantinori for running that through and sharing the synthesis. I checked the three surviving items against the controller code and against `main`, and none of them need a change here:

**1. `secrets: [create, get]` non-idempotency.** Copied verbatim from the pre-split `rbac/cluster/cluster-role.yaml` on `main` (same `[ create, get ]`), so the split changes nothing. It matches the operator: across the whole controller the only Secret operations are a `get` and a `create` in `SqlCRUpdater` (never update/patch), and that create is guarded by an existence check with a random-suffixed name, so it does not fail on re-reconcile. Adding `update`/`patch` would grant verbs the operator never uses and widen the namespaced role, against the least-privilege goal here. Left as-is intentionally.

**2. Finalizer subresource.** The reconciler implements `Cleaner<ApicurioRegistry3>` and only ever `patchResource`/`patchStatus`; the default JOSDK finalizer write goes through the main CR, which the existing `patch`/`update` on `apicurioregistries3` already authorizes. No separate `/finalizers` grant is needed.

**3. roles/rolebindings scope.** Also verbatim from `main`. The operator reconciles only namespaced `Role`/`RoleBinding` (`AppRoleResource`/`AppRoleBindingResource` extend `CRUDKubernetesDependentResource<Role|RoleBinding, ...>`), so single-namespace installs keep them in-namespace and AllNamespaces promotes them exactly as documented.

So the diff is a faithful scope split with no verb widened or narrowed versus `main`, apart from the two deliberate CR changes already discussed. Happy to adjust if you read any of these differently.

paoloantinori:
Thanks @Taimoo-r — agreed on all three, and the pre-existence check against `main` is the right way to adjudicate these. Quick confirmations:

1. **`secrets: [create, get]`** — correct, the `[create, get]` pair is carried verbatim from `rbac/cluster/cluster-role.yaml` on `main`, so the split neither widens nor narrows it. Not in scope for #9008.
2. **CR discovery verbs** — confirmed: the informer only needs `get/list/watch`, which already live in `cluster-role.yaml`, so the namespace tier adds nothing to discovery.
3. **No workload verbs introduced** — the split re-distributes existing verbs across tiers; it doesn't add any new `(apiGroup, resource, verb)` the controller didn't already have.

So nothing to change here for the split itself. The three items are good candidates for a follow-up least-privilege pass (e.g. idempotent secret handling, tightening any unused verbs) but they predate this PR and shouldn't block it. Approving the structural split as it stands.

Taimoo-r:
Hi @jsenko, whenever you get a chance for the final merge call, I would appreciate it. This PR is the base for the follow-up (Part 2, the namespaced manifest install), so I am holding that branch until this one lands. Thanks!

aclfe (commented):
Went through this pretty thoroughly and overall it looks really solid. The `clusterPermissions` (CR/CRD discovery) vs `permissions` (workloads) split is the right shape, and I like that `create`/`delete` got dropped from the CR entirely rather than just moved down a tier, since the reconciler only ever patches the resource/status and removes the finalizer (`implements Reconciler + Cleaner`, only `patchResource`/`patchStatus`/`defaultDelete`) so those two verbs really were dead.

I also chased the AllNamespaces question from earlier in the thread down to the OLM source just to be sure, and yeah it checks out: in `operatorgroup.go` the global branch does `ClusterPermissions = append(ClusterPermissions, Permissions...); Permissions = nil` and then `ensureSingletonRBAC(...)`, so the workload verbs in `role.yaml` genuinely get promoted to a ClusterRole in AllNamespaces mode. The non-global branch copies the Roles per target namespace via `ensureTenantRBAC`. So the split is correct in both modes, no argument there.

Couple of things I wasn't sure about / just wanted to confirm:

1. Is anything actually asserting the *generated* CSV comes out split the way we want? `RbacInstallerSyncTest` is a nice guard (I replayed its tuple logic on the real files, the installer is a clean superset, nothing missing) but that only checks the source YAML stays in sync between `rbac/namespaced` and the olmv1 installer role. Nothing checks that `operator-sdk generate bundle` actually emits `clusterPermissions = {CR/CRD read}` and `permissions = {workloads}`. Right now that seems to rest on your manual check plus the live OLM ITs. Intentional, or worth a small kustomize-render-and-parse test so a future refactor can't collapse the split?

2. The OLM v0 / v1 boxes in the description are still unchecked. Since those two new ITs are really the only proof the whole thing materializes end to end (bundle -> OLM -> operand), probably worth landing a green run there before this merges? Everything I could verify statically passes, I just haven't seen it exercised on a cluster.

Left a few small inline notes too, none of them blocking.

operator/controller/src/main/deploy/rbac/namespaced/cluster-role.yaml
```diff
@@ -0,0 +1,32 @@
...
+rules:
+
+  # The operator watches ApicurioRegistry3 CRs cluster-wide to discover them (informer).
+  # Mutating verbs (patch/update for spec migration, status, and the JOSDK finalizer) are
+  # namespace-scoped and live in role.yaml. The operator never creates or deletes the CR.
+  - apiGroups:
+      - registry.apicur.io
+    resources:
+      - apicurioregistries3
+      - apicurioregistries3/status
```
aclfe: Tiny thing while reading: `list`/`watch` on `apicurioregistries3/status` doesn't really buy anything, there's no list/watch endpoint for a status subresource (you list/watch the parent). The informer only needs `get`/`list`/`watch` on the parent. This is carried over from the old ClusterRole so it's not introduced here, but since you're reshaping the cluster tier anyway you could drop `/status` from this cluster rule entirely (leaving it in `role.yaml` where `patch`/`update` on it is genuinely used). Harmless either way, just noticed it.

operator/controller/src/main/deploy/rbac/namespaced/role.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
```
aclfe, line 124: Quick sanity check on `leases` now that they live in the namespaced tier. For a true SingleNamespace install (operator in ns A, watching ns B) with leader election turned on, the lease sits in the operator's own namespace A, not the target. I *think* that's still fine, because OLM also creates the `permissions` Role in the install namespace (`ensureTenantRBAC` copies from the operatorgroup's own namespace, so the base Role has to exist there) and leader election is off by default anyway. But pre-split this was cluster-wide so any lease namespace worked, so I just want to confirm we're not quietly narrowing that. Worst case is someone setting `lease-namespace` to some unrelated third namespace, which is admittedly a weird config.

operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/NamespacedPermissionsOLMITTest.java
```diff
@@ -0,0 +1,95 @@
...
+        });
+
+        // A namespace the operator does not target, so OLM creates no Role/RoleBinding there.
+        var foreignNamespace = ITBase.calculateNamespace();
+        ITBase.createNamespace(client, foreignNamespace);
+        try {
+            var serviceAccountUser = "system:serviceaccount:" + namespace + ":" + OPERATOR_SERVICE_ACCOUNT;
+
+            // Allowed: creating a Deployment in the operator's own (target) namespace.
+            assertThat(canCreateDeployment(serviceAccountUser, namespace))
```
aclfe: Very minor: this "allowed in own namespace" check runs right after the operator-ready `await` but not inside its own `await`, so if OLM's RoleBinding lands a beat after the operator Deployment reports ready, this first SAR could come back `false` and flake. `@RetryTest` probably covers it, but wrapping this `assertThat` in an `await` too would make it deterministic.

Separately on the naming: the javadoc/comment says SingleNamespace/OwnNamespace, but the default `operator-group.yaml` targets the install namespace, so this is really exercising **OwnNamespace** (target == install ns). The true SingleNamespace case (target != install) is exactly where OLM's copy-into-target-namespace path differs, and it isn't covered here. Not saying it has to be in this PR, just that the wording reads a bit broader than what's actually tested.

operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/AllNamespacesOLMITTest.java
```diff
@@ -0,0 +1,79 @@
...
+            await().ignoreExceptions().untilAsserted(() -> {
```
aclfe, line 91: The negative test went out of its way to use a SubjectAccessReview specifically so it wouldn't depend on operand image readiness, but this positive one waits for the operand Deployment to actually reach `readyReplicas == 1` in `otherNamespace`, which means it hangs on the app image pulling and the pod going ready. That makes it the heaviest and most flake-prone of the three ITs. Could you get the same RBAC signal with a SAR (can the operator SA create deployments in `otherNamespace`?) and keep "operand actually becomes ready" as its own assertion, or is the full end-to-end watch worth the extra flakiness here?

operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/AllNamespacesOLMITTest.java
```diff
@@ -0,0 +1,79 @@
...
+ * (from rbac/namespaced/role.yaml): in AllNamespaces mode the OperatorGroup targets all namespaces
+ * ({@code *}), and OLM promotes each CSV {@code permissions} rule to a ClusterRole +
+ * ClusterRoleBinding. In SingleNamespace mode the same rules are instead copied as Roles into the
+ * target namespace(s). See the OLM OperatorGroup RBAC docs.
+ */
+@QuarkusTest
+@Tag(OLM)
+public class AllNamespacesOLMITTest extends OLMITBase {
+
+    private static final Logger log = LoggerFactory.getLogger(AllNamespacesOLMITTest.class);
```
aclfe: Small nit: this `log` field isn't used anywhere (same in `NamespacedPermissionsOLMITTest`), can drop it.

Taimoo-r:
Hi @aclfe, ok this was a great review, thank you. Going and reading operatorgroup.go yourself is genuinely above and beyond, and honestly it nails the AllNamespaces thing shut better than I managed by quoting docs at everyone. Everything below is in 0024fc8e.

- `/status` is off the cluster tier (no list/watch endpoint for a status subresource); it keeps `patch`/`update` in `role.yaml`, and the olmv1 installer role mirrors it so the superset guard still holds.
- Q1: added `RbacSplitTest`. operator-sdk maps ClusterRole to `clusterPermissions` and Role to `permissions` verbatim, so the split is just which file a rule lives in, and the test guards that partition (cluster tier read-only CR/CRD, namespace tier workloads with no cluster-scoped discovery). Sneak a workload rule up and the build stops you. README points at both guards.
- AllNamespaces IT: it was proving RBAC by waiting for an app image to pull, so it now uses a SubjectAccessReview and only checks the operand Deployment gets created, not that its pods go ready.
- Namespaced IT: positive check wrapped in `await` (a late RoleBinding is not a failure), renamed to OwnNamespace since that is what it actually exercises (true SingleNamespace still uncovered), and dropped two unused loggers.

Left alone on purpose: `leases` stays namespaced, since with leader election (off by default) the lease lives in the operator's own namespace where OLM creates the Role, and only an unrelated third namespace is lost, which is not a supported layout (happy to add a docs note). And I am leaving the OLM v0/v1 boxes for a green CI run rather than grading my own homework.

Also refreshed the PR description to match.

carlesarnal:
## Review

All of jsenko's requested changes have been addressed in the subsequent commits, and aclfe's feedback is resolved (with leases narrowing intentionally deferred). The RBAC split is structurally correct, and the test coverage is solid — `RbacSplitTest` guards that the cluster tier stays read-only CR/CRD discovery, `RbacInstallerSyncTest` prevents drift between the operator runtime perms and the OLM v1 installer role, and the two OLM IT tests verify the actual permission boundary from both ends (AllNamespaces promotion + OwnNamespace scoping). Nice work.

A few minor nits:

**1. DCO sign-off missing (all 7 commits)**
CONTRIBUTING.md requires all contributions carry a `Signed-off-by:` line. None of the commits in this PR have one. A `git rebase HEAD~7 --signoff` + force-push would fix it.

**2. `canCreateDeployment()` duplicated**
`AllNamespacesOLMITTest` (lines 99–110) and `NamespacedPermissionsOLMITTest` (lines 88–99) have an identical `canCreateDeployment()` method. Consider extracting it to `OLMITBase` so future tests can reuse it without copying.

**3. YAML style inconsistency in `role.yaml`**
`operator/controller/src/main/deploy/rbac/namespaced/role.yaml:123` — the `secrets` rule uses flow-style `[ create, get ]` while every other rule in the file uses block sequences. Minor, but makes it stand out oddly when scanning the file.

None of these are blockers for the RBAC logic itself. The split, the tests, and the documentation look good.

Taimoo-r:
Hi @carlesarnal, thanks, all three sorted in 4477582, with a sign-off pass over the earlier commits too:

- DCO: rebased the branch with `--signoff`, so every commit now carries the trailer.
- `canCreateDeployment`: pulled up into `OLMITBase`, together with a small `operatorServiceAccountUser()` helper, so both ITs share one copy and the next test can reuse it instead of pasting a third.
- `role.yaml`: the secrets rule is block style now, so it stops being the one line everyone squints at.

Appreciate the careful read.

jsenko (commented):
Re-reviewed the latest head (`4477582c3`). All 6 findings from my previous review are addressed:

1. ✅ OLM v1 installer CR synced — `create`/`delete` dropped, `/status` split out with `patch`/`update` only
2. ✅ Drift guard — `RbacInstallerSyncTest` (superset check) + `RbacSplitTest` (CR verbs, cluster vs namespace scope)
3. ✅ v0-only test gating — `assumeTrue(getOlmVersion() == 0, ...)` in both `AllNamespacesOLMITTest` and `NamespacedPermissionsOLMITTest`
4. ✅ CR split verified correct
5. ✅ Secrets verified correct
6. ✅ `OML_VERSION` → `OLM_VERSION` fixed

Clean work addressing feedback from multiple reviewers. No further concerns from my side.

Taimoo-r:
Awesome to hear @jsenko! Once it's merged, I'll raise the PR for Part 2. 🚀🧑🏼‍💻

jsenko (commented):
## Re-review (automated)

All prior findings (jsenko's 6 + paoloantinori's RBAC concerns + aclfe's 5 comments) are addressed. The RBAC split is correct: cluster tier is read-only CR/CRD discovery, namespace tier holds workloads and CR mutations. Documentation and tests are solid.

### should-fix: RbacSplitTest missing positive assertion for CR mutation verbs in namespace tier

`RbacSplitTest.namespaceTierHoldsWorkloadsAndNoClusterScopedDiscovery()` checks that:
- Workloads are present (via `apps/deployments/create` marker)
- CRD access is forbidden in namespace tier
- `list/watch` on CR is forbidden in namespace tier

But it doesn't positively assert that `role.yaml` **contains** `patch/update` on the CR. If a future refactor accidentally removes the CR mutation rules from `role.yaml` entirely (without adding them to `cluster-role.yaml`), the test would still pass — the "no list/watch" check simply wouldn't execute (no CR-targeting rule found).

`RbacInstallerSyncTest` wouldn't catch this either, since it only checks that the installer is a **superset** of the operator — a smaller operator permission set still satisfies the superset condition.

Suggested addition to `namespaceTierHoldsWorkloadsAndNoClusterScopedDiscovery()`:

```java
// Positive guard: CR mutation verbs must be in the namespace tier
assertThat(tuples(rules))
    .contains(CR_GROUP + "/" + CR_RESOURCE + "/patch",
              CR_GROUP + "/" + CR_RESOURCE + "/update",
              CR_GROUP + "/" + CR_RESOURCE + "/status/patch",
              CR_GROUP + "/" + CR_RESOURCE + "/status/update");
```

jsenko (dismissed):
LGTM, thanks! Let's wait for the full testsuite to pass, then I'll merge. There's a minor non-blocking issue https://github.com/Apicurio/apicurio-registry/pull/9035#pullrequestreview-4843209181 . You can fix it now or in a follow-up PR.

Taimoo-r:
Yeah that was a good catch on the vacuous-pass gap. You are right that both tests would stay green if the CR mutation rules simply vanished from `role.yaml`, since the negative check only runs when a CR-targeting rule is found and the superset check is happy with a smaller operator set.

The test suite you kicked off is nearly through (one check still running, everything else green), so rather than push a new head and restart the whole run, I will add that positive assertion in a quick follow-up PR right after this merges. It is test-only, and I have already confirmed the four tuples are present today, so it is a pure strengthening of the guard, not a behaviour change.

Taimoo-r:
Hi @jsenko, the two OLM job failures look like a ttl.sh flake rather than anything in the PR. The CatalogSource pod cannot pull the catalog image (ttl.sh returned HTTP 502, then `not found` on retries), so OLM never installs the operator and every OLM test times out in setup, including `SmokeOLMITTest` and `UpgradeOLMITTest`, which do not touch the RBAC changes.

Rather than push an empty commit just to re-run, I have pushed `138e7977`, which folds in the `RbacSplitTest` positive assertion you flagged earlier: it now asserts `role.yaml` actually contains `patch`/`update` on the CR and its `/status`, so deleting those rules could no longer pass silently. That addresses your point in this PR and re-triggers the OLM jobs, so no separate follow-up is needed.

Taimoo-r:
Hi @jsenko, this one was not a ttl.sh flake after all. olm-v0 passed this run; olm-v1 failed for a real reason.

`AllNamespacesOLMITTest` and `NamespacedPermissionsOLMITTest` skip themselves under OLM v1 with `assumeTrue(getOlmVersion() == 0)` inside the test method. Under v1 `@BeforeAll` starts the OLM v1 install, which only waits for the ClusterCatalog to report `Serving`, not for the CRD/operator to finish installing (that part is async). The test method then hits `assumeTrue` almost immediately and aborts before the CRD is registered. `@RetryTest` does not distinguish an aborted assumption from a real failure, so it retried the aborted test, and each retry ran `OLMITBase.afterEach()`, which tried to delete `ApicurioRegistry3` CRs and got a 404 because the cluster did not know that resource type yet. The retry-cleanup exceptions were swallowed and only logged as warnings, but the final standard JUnit/Quarkus `@AfterEach` call was not swallowed, so that 404 became the reported test failure, with the original assumption failure demoted to a suppressed exception underneath it.

Fixed in `a6cbeb178` by replacing the `assumeTrue` with a class-level `@DisabledIfSystemProperty` on `test.operator.olm-version` (the same property the olm-v1 job sets), so the whole class, including `@BeforeAll`, is skipped under v1 and the v1 install never starts for these two classes. Verified the change compiles and the RBAC unit tests still pass; the class-level skip itself will show as a clean "skipped" in the olm-v1 job once CI runs it. PR description updated to match.

operator/controller/src/main/deploy/csv/kustomization.yaml
aclfe, line 1:
```
2026-08-04T09:58:57.3530572Z     message: "error for resolved bundle \"apicurio-registry-3.v3.3.2-snapshot\" with\
2026-08-04T09:58:57.3531239Z       \ version \"3.3.2-snapshot\": failed to create resource: clusterroles.rbac.authorization.k8s.io\
2026-08-04T09:58:57.3531879Z       \ \"apicurio-registry--16n97mwwk9bmpteeoamswnk20amn1y519oxkykf80eq5\" is forbidden:\
2026-08-04T09:58:57.3532476Z       \ user \"system:serviceaccount:test-23c403b:apicurio-registry-operator-installer\"\
2026-08-04T09:58:57.3533052Z       \ (groups=[\"system:serviceaccounts\" \"system:serviceaccounts:test-23c403b\"\
2026-08-04T09:58:57.3533617Z       \ \"system:authenticated\"]) is attempting to grant RBAC permissions not currently\
2026-08-04T09:58:57.3534242Z       \ held:\n{APIGroups:[\"\"], Resources:[\"namespaces\"], Verbs:[\"get\" \"list\"\
2026-08-04T09:58:57.3534682Z       \ \"watch\"]}"
2026-08-04T09:58:57.3535186Z     observedGeneration: 1
2026-08-04T09:58:57.3535673Z     reason: "Retrying"
2026-08-04T09:58:57.3536167Z     status: "True"

```
^ error logs

jsenko (changes requested):
## Review

CI failure in **Operator Tests / Remote Tests (olm-v1)** — `SmokeOLMITTest.smoke` fails deterministically.

### must-fix: OLM v1 installer ClusterRole missing `namespaces` permission

The OLM v1 installer ServiceAccount (`apicurio-registry-operator-installer`) tries to create a ClusterRole granting `{namespaces: [get, list, watch]}`, but the installer SA itself doesn't hold that permission. OLM v1 enforces RBAC escalation prevention — you can't grant permissions you don't have.

Error from the ClusterExtension status:

```
user "system:serviceaccount:test-...:apicurio-registry-operator-installer"
is attempting to grant RBAC permissions not currently held:
{APIGroups:[""], Resources:["namespaces"], Verbs:["get" "list" "watch"]}
```

The fix is to add the `namespaces` rule (`get`, `list`, `watch`) to the OLM v1 installer ClusterRole in `operator/olm-tests/src/test/deploy/olmv1/cluster-role.yaml`, so the installer can grant it to the operator. `RbacInstallerSyncTest` should catch this gap — verify it does after the fix.

aclfe (commented):
`@BeforeAll` succeeded, the test method then waited 10 min x 3 retries for a deployment that OLM was never able to install. 

I'm inclined to believe this specific `csv/kustomization.yaml` switched from /rbac/cluster to /rbac/namespaced  so everything the operator needed was a ClusterRole, the bundle only ever had clusterPermissions. 

Switching csv/kustomization.yaml from ../rbac/cluster to ../rbac/namespaced means the bundle now has a `Role`, which land in the CSV as `spec.install.spec.permissions`.

fix:

Give the installer the rule OLM is trying to grant on its behalf. In operator/olm-tests/src/test/deploy/olmv1/cluster-role.yaml, in the transitive operator permissions block:

```
    - apiGroups:
        - ''
      resources:
        - namespaces
      verbs:
        - get
        - list
        - watch
```

an extra rule there should do trick

Give this a try and let me know

paoloantinori:
Thanks for flagging this and for the careful diagnosis — I agree it's a real design question, not something to hack green. I pulled the workflow run and your read of the two failures is correct.

What the log shows is actually two stacked problems, not one:

1. **Installer escalation block (jsenko's point, the first thing OLM hits).** The installer SA can't grant `namespaces get/list/watch` it doesn't hold, so the ClusterExtension never reaches `Installed: True` in the first place. That's the `{Resources:["namespaces"], Verbs:["get" "list" "watch"]}` "attempting to grant RBAC permissions not currently held" error. Fixing the installer role is necessary regardless of which direction we pick.

2. **Operator CrashLoopBackOff on workload list (the one you surfaced).** Even past the installer step, the JOSDK reconciler is configured to **watch all namespaces** (`'apicurioregistry3reconciler' controller is configured to watch all namespaces` in the build warning), so its informer issues `GET /api/v1/services?resourceVersion=0` at the **cluster** scope. Under OLM v1 there is no OperatorGroup to promote the namespaced `permissions` to a ClusterRole, so the operator SA only holds a namespaced Role and the cluster-scoped list returns 403. That's the `services is forbidden ... cannot list resource "services" ... at the cluster scope` line, independent of the installer error.

So the root cause of the second one isn't really "v1 doesn't promote permissions" — it's "the operator's reconcile scope is cluster-wide by default, and v1 has no mechanism to narrow it." Your two directions map to two different fixes for that:

- **Keep workloads in `clusterPermissions`** — the operator stays cluster-wide on both v0 and v1, least-privilege narrows happen in Part 2 (manifest install). Lowest risk for v1 correctness *now*.
- **Make the v1 operator watch only its own namespace** (via `APICURIO_OPERATOR_WATCHED_NAMESPACES` / the namespace-scoped reconcile path) so the namespaced `permissions` are sufficient. More correct for least-privilege, but it's a behavioral change to the operator's watch model on v1 and I'd want to see it covered by an IT that asserts the reconcile actually stays namespaced.

I lean toward your first direction too, for the same reason: Part 1's stated scope is the CSV RBAC split, and the manifest-install least-privilege story is already scheduled for Part 2 where a watch-namespace change can be tested properly. Pushing the watch-scope decision into this PR under a red CI mixes two concerns. Concretely: move `deployments/pods/services/configmaps/...` back into `clusterPermissions` for this PR so both OLM versions install and the operator reconciles, fix the installer `namespaces` rule, and let Part 2 do the real least-privilege narrowing with its own test coverage.

One thing worth noting for whichever direction lands: `RbacInstallerSyncTest` checks the installer is a *superset* of the operator, so once workloads move back to clusterPermissions the installer needs them too — worth confirming the sync test would catch a future drift where the operator tier and installer diverge on scope.

jsenko:
@Taimoo-r @paoloantinori Good analysis on the two stacked problems. Before we pick a direction, I want to lay out the OLM v0 vs v1 difference more explicitly, because it informs a third option.

## OLM v0 vs v1 philosophy

OLM v0's OperatorGroup actively manages RBAC scope on the operator's behalf. The CSV declares `permissions` (namespace-scoped) and `clusterPermissions` (cluster-scoped), and the OperatorGroup decides what happens to `permissions` based on the install mode:

| Mode | What OLM v0 does with `permissions` |
|---|---|
| SingleNamespace / OwnNamespace | Creates Role + RoleBinding in the target namespace |
| MultiNamespace | Copies the Role + RoleBinding into each listed target namespace |
| AllNamespaces (`*`) | Promotes each rule to a ClusterRole + ClusterRoleBinding |

OLM v1 deliberately dropped this abstraction. The ClusterExtension has a single `namespace` field. OLM v1 creates the Role from `permissions` in that one namespace — period. There's no OperatorGroup, no promotion, no multi-namespace replication. Anything beyond single-namespace access is the **admin's responsibility** to set up explicitly.

## What this means for the three deployment scenarios under OLM v1

**Single-namespace:** Works as-is. The CSV `permissions` create a Role in the ClusterExtension's namespace, the operator is configured with `apicurio.operator.watched-namespaces=<ns>`, and the namespaced Role is sufficient. No extra admin step needed.

**Multi-namespace (specific list):** The admin must manually create a Role + RoleBinding in each additional watched namespace (beyond the one OLM creates). Or use a ClusterRole + ClusterRoleBinding for simplicity.

**All-namespace:** The admin must manually create a ClusterRole + ClusterRoleBinding granting the **operator SA** (not the installer SA) cluster-wide workload access. This is what lets the operator's default `watchingAllNamespaces()` work.

In all cases, the **installer ClusterRole** stays the same — it just needs to hold a superset of everything the operator might be granted (escalation prevention). The difference is in what the admin creates for the operator SA at runtime.

## Proposed direction (third option)

Instead of moving workloads back to `clusterPermissions` (option 1, loses the split for v1) or forcing single-namespace behavior on v1 (option 2, behavioral change), keep the CSV least-privilege as-is and model the admin's manual step in the test:

1. **Fix the installer ClusterRole** — add `namespaces: [get, list, watch]` (problem 1, needed regardless).

2. **`SmokeOLMITTest` (all-namespace under v1)** — in `@BeforeAll`, create an additional ClusterRole + ClusterRoleBinding granting the operator SA cluster-wide workload access (deployments, services, pods, etc.). This simulates the admin step that would be required in a real all-namespace v1 deployment. Clean it up in `@AfterAll`.

3. **Single-namespace test** — no extra setup needed. The namespaced Role from the CSV is sufficient when the operator is configured to watch only its namespace (but that's a separate concern for when we actually support single-namespace v1 deployments with `watched-namespaces`).

This preserves the RBAC split, correctly models the v1 admin workflow, and doesn't require a behavioral change to the operator.

## Documentation (optional, can be a follow-up)

@Taimoo-r It would be useful to have a `README.md` in the `operator/olm-tests/` module that documents:

1. The OLM v0 vs v1 philosophy difference (OperatorGroup-managed vs admin-managed RBAC scope).
2. The three deployment scenarios (single-namespace, multi-namespace, all-namespace) and what each requires under v0 and v1.
3. What the installer ClusterRole covers and why it needs to be a superset.
4. How the test setup models each scenario (what extra resources are created and why).

This will be the foundation for the OLM v1 installation guide in the user-facing docs when we fully support v1 deployments. Not blocking for this PR — can be a follow-up.

paoloantinori:
Thanks @jsenko. The third option is the right model. Laying out the v0-vs-v1 RBAC-scope philosophy explicitly is what makes the call clear: under v1 the admin owns the operator-SA runtime grants, so modeling that step in the test is faithful rather than a workaround.

A few confirmations on the proposal as I read it:

1. **Installer superset invariant still holds.** The additional ClusterRole + ClusterRoleBinding you'd create in `@BeforeAll` grants the *operator* SA, not the installer SA. The installer ClusterRole's job (escalation prevention) is to be a superset of anything the operator *could* be granted at runtime; it is orthogonal to what the admin actually grants the operator SA. So adding `namespaces: [get, list, watch]` to the installer (problem 1) and separately granting the operator SA cluster-wide workload access (problem 2, admin step) are independent levers and don't conflict with `RbacInstallerSyncTest`'s drift guard. Worth confirming the guard's assertion is scoped to installer-SA grants and doesn't try to enumerate operator-SA runtime grants, or the test could fire a false positive once the admin-step ClusterRole exists.

2. **Test isolation / cleanup.** The admin-step ClusterRoleBinding is cluster-scoped, so it must come down in `@AfterAll` (you note this), and the cleanup path needs to run even if the smoke test itself fails partway, otherwise it leaks workload access into the shared cluster for subsequent OLM tests. A try/finally around the grant rather than relying on `@AfterAll` ordering would be safer if the grant is created inside the test method.

3. **Endorsing the docs follow-up.** The `operator/olm-tests/README.md` covering v0/v1 philosophy and the three scenarios is genuinely useful, and I'd treat it as required rather than optional once we land v1 support. The admin-managed RBAC step is the one thing a real v1 deployer will trip on, and having it written down next to the tests that model it keeps the two from drifting.

If Taimoo-r takes option 3, the shape is sound from the RBAC-invariant angle; the only thing I'd ask for in the implementation is the try/finally cleanup discipline above.

Taimoo-r:
Hi @jsenko @paoloantinori @aclfe, went with option 3, in `e6a29e4a3`.

- Added `namespaces [get,list,watch]` to the OLM v1 installer role so it can grant the SDK-generated permission the operator requests, which unblocks the ClusterExtension.
- Added `olmv1/operator-workload-clusterrole.yaml` and its binding granting the operator SA cluster-wide workload access (mirrors `role.yaml`). `SmokeOLMITTest` applies it in `@BeforeAll` under v1 and removes it in `@AfterAll`, so a smoke failure can't leak the cluster-scoped grant into later tests.
- The CSV split, the v0 boundary tests, and `RbacSplitTest` are unchanged.
- Added `operator/olm-tests/README.md` covering the v0-vs-v1 RBAC-scope model and the three scenarios.

On your guardrails @paoloantinori: cleanup is in `@AfterAll`, which runs regardless of outcome, and `RbacInstallerSyncTest` compares the installer against the operator's declared RBAC files, not runtime grants, so the admin-step ClusterRole doesn't affect it.

One caveat worth flagging @jsenko: `RbacInstallerSyncTest` won't actually catch the missing `namespaces` rule, since that permission is SDK-generated at build time and isn't in the source RBAC the test reads. I noted it in the installer role and the README. Closing that would mean diffing against the generated CSV instead, a reasonable follow-up.

For the record: this keeps the operator all-namespace on v1, so a real all-namespace v1 install still needs that admin-provided ClusterRole. Actually retiring the admin step means having the v1 operator watch only its own namespace, which is a behavioral change tied to the Part 2 `watched-namespaces` work, so I left it as a follow-up rather than mixing it in here.

paoloantinori:
Thanks @Taimoo-r — option 3 is the right model and the implementation matches it cleanly. The `operator-workload-clusterrole.yaml` mirrors `role.yaml` at cluster scope, `SmokeOLMITTest` grants it in `@BeforeAll` and removes it in `@AfterAll` so a failure cannot leak cluster-wide access, and adding `namespaces [get,list,watch]` to the installer role is the correct fix for the ClusterExtension block.

The separation is clean: CR discovery (`get/list/watch` on `apicurioregistries3`) stays in `cluster-role.yaml` via `clusterPermissions`, while workload access comes from the admin step. No duplication, no gap. LGTM on this approach from me.

aclfe:
Hi @jsenko @paoloantinori @Taimoo-r, before we merge, I read through the operator-controller render code and I think the premise behind option 3 is off. Flagging now rather than after merge.

----

Short version: OLM v1 does promote permissions to a ClusterRole in AllNamespaces mode, it's not v0-only. **So I'm not sure but I don't think problem 1 and problem 2 were ever two separate problems?**

BundleCSVClusterPermissionsGenerator promotes CSV permissions to a ClusterRole + ClusterRoleBinding when target namespaces are [""], appending a namespaces get/list/watch rule for parity with v0 (comment links the v0 operatorgroup.go it mirrors). BundleCSVPermissionsGenerator returns nil in that same mode, so no Role gets created at all. 

That's exactly our path: defaultTargetNamespacesForBundle returns [""] since our CSV supports AllNamespaces and cluster-extension.yaml sets no spec.config, and SingleOwnNamespaceInstallSupport is still alpha/default-false so we can't opt out anyway.


**The error we're looking at is the proof:** it's on a ClusterRole, blocking namespaces get/list/watch, a rule that's not in role.yaml, cluster-role.yaml, or the generated CSV. Only the promotion adds it. Upstream's argocd-operator v0.6.0 golden fixture shows the same split: plain Role under single-namespace, promoted ClusterRole with that extra rule under all-namespaces.
So I think it's one failure, not two. The CrashLoop happened because the promoted ClusterRole got rejected by escalation prevention, leaving the SA on the read-only tier only, so informers 403'd on services. The installer already has all the workload verbs, so namespaces get/list/watch is the whole gap.


Ask: push just the installer rule and see if smoke goes green. One push settles it, and also tells us whether the applier partial-applied here.


**Why maybe drop the workload ClusterRol**e:? it grants the same access the promotion would, so the test can't tell if promotion actually worked. If a future role.yaml change reopens an escalation gap, the promoted ClusterRole fails to create, but the test would still pass because we granted the access ourselves. Bundle ships broken for real v1 users.

--- 

Thoughts?

paoloantinori:
Thanks @aclfe for reading the operator-controller source directly — that is the right move, and I went and checked `internal/operator-controller/rukpak/render/registryv1/generators/generators.go` on `main` to verify your claim. Your reading is correct:

- `BundleCSVPermissionsGenerator` returns `nil, nil` when `len(opts.TargetNamespaces) == 1 && opts.TargetNamespaces[0] == ""` (AllNamespaces mode). No Role is created at all.
- `BundleCSVClusterPermissionsGenerator`, in that same mode, iterates `CSV.Spec.InstallStrategy.StrategySpec.Permissions`, appends a `namespaces get/list/watch` rule to each, and appends them to `clusterPermissions` — which become ClusterRole + ClusterRoleBinding. The doc comment there is explicit that this mirrors `operator-lifecycle-manager operatorgroup.go:539` for parity with v0.

So the promotion is not v0-only. In AllNamespaces mode OLM v1 already lifts the `permissions` rules to a ClusterRole and adds the `namespaces get/list/watch` rule on its own — exactly as v0 does. That makes the two things you flagged two faces of one mechanism, and it has two consequences for option 3 as it stands:

1. `operator-workload-clusterrole.yaml` grants the same workload verbs the promotion would grant. The test cannot tell whether the promotion actually materialized or whether we granted the access ourselves. If a future `role.yaml` change reopens an escalation gap (so the promoted ClusterRole gets rejected), the test stays green because the manual ClusterRole carries the verbs regardless — the bundle ships broken for real v1 AllNamespaces users and the test does not catch it.
2. The only real gap in the installer ClusterRole was `namespaces get/list/watch`, which is exactly the rule the promotion appends. The installer already had every other workload verb.

Your proposed way to settle it is the right one: drop the manual workload ClusterRole, keep only the `namespaces get/list/watch` addition to the installer role, and let smoke prove whether the promotion alone is sufficient. One push resolves it empirically, and it also tells us whether the applier partial-applied during the earlier run. If smoke stays green without the manual ClusterRole, the promotion is doing the work and the test is actually exercising it; if it goes red, we have a real v1-promotion gap to fix and the manual ClusterRole was not just redundant masking.

This is a genuine design question and I do not want to block on my read alone — deferring to @jsenko for the merge call. But the source-code point stands and I think the empirical test is worth the one extra push before this lands.

Taimoo-r:
So @aclfe @paoloantinori @jsenko,

Agreed on the source read, and thanks @aclfe for going into the operator-controller renderer. The premise behind option 3 was off: OLM v1 does promote `permissions` to a ClusterRole in AllNamespaces mode (`BundleCSVClusterPermissionsGenerator`, appending `namespaces get/list/watch`, mirroring v0's operatorgroup.go), and since this bundle always resolves to AllNamespaces today (`SingleOwnNamespaceInstallSupport` alpha/off), the manual workload ClusterRole was granting exactly what the promotion already grants. So it was redundant, and worse, it masked whether the promotion actually works. Fixed in `ec0e4ec0f`.

In this push:

- Removed `operator-workload-clusterrole.yaml` + binding and the `SmokeOLMITTest` `@BeforeAll`/`@AfterAll` wiring. Smoke now just installs the bundle and waits, so a green v1 run exercises the promotion for real.
- Kept the installer `namespaces [get,list,watch]` rule (the actual gap).
- **Added a deterministic guard for the `namespaces` rule.** Going back through the code, I noticed a small three-line addition to `RbacInstallerSyncTest` would help here: it now asserts the installer holds the promotion-injected `namespaces get/list/watch` alongside the declared workloads. This closes the gap I flagged earlier (that the SDK-generated `namespaces` rule was not covered by the sync test), and does it from the source YAML without rendering the CSV, so it is lighter than the generated-CSV diff I had floated. Concretely: lose that rule and it is now a sub-second unit failure, not the 30-minute v1 CrashLoop we hit before. The live smoke run stays as the end-to-end backstop.
- Rewrote `olm-tests/README.md` and the PR description to the corrected model: v1 keeps v0's promotion, the installer holds the superset including `namespaces`, no admin step involved.

On where the least-privilege win lands: it is delivered today for OLM v0 single/own-namespace installs, which is exactly Part 1's scope. Those installs now get a namespaced Role instead of the old cluster-wide grant, and the unused CR `create`/`delete` verbs are dropped everywhere. On v1 the effective scope is unchanged for now because every v1 install is AllNamespaces and promotes, but that is a v1 limitation (no single-namespace without the alpha gate), not a gap in the split. The same split narrows v1 the moment the `watched-namespaces` work in Part 2 lets us opt out of AllNamespaces, so this PR is the foundation that unblocks it.

This is the one-push settle you both suggested. If v1 smoke stays green, the promotion is doing the work; if it goes red, we have learned the manual role was compensating for a real gap rather than masking, and I will dig into why. Over to CI, and to @jsenko for the merge call.

paoloantinori:
The promotion-model correction in `ec0e4ec` looks right, and the `OLM_V1_PROMOTION_INJECTED_TUPLES` addition to `RbacInstallerSyncTest` is the deterministic guard the conversation was missing: losing the installer `namespaces` rule is now a sub-second unit failure rather than a 30-minute CrashLoop.

One stale comment this push leaves behind. The inline note in `olm-tests/src/test/deploy/olmv1/cluster-role.yaml` (around lines 90-91) still reads:

> this rule is not in rbac/namespaced (it is SDK-generated at build time), so RbacInstallerSyncTest does not guard it.

That was accurate before this push but now contradicts the test: `RbacInstallerSyncTest.OLM_V1_PROMOTION_INJECTED_TUPLES` adds `/namespaces/{get,list,watch}` to the expected operator set, so the test does fail if the installer role drops that rule. Worth updating the comment so a future reader doesn't assume the unit suite won't catch a removal (and ideally rewording "does not guard it" to point at the injected-tuples constant, so the mechanism is discoverable from the YAML).

Otherwise this looks good to me. Over to CI and @jsenko for the merge call.

Taimoo-r:
Okay @paoloantinori, as CI is completed & green, I'll remove that stale comment in Part 2 of this PR.

paoloantinori:
Sounds good. CI green on this head is what I wanted to see, and deferring the stale comment cleanup to Part 2 is the right call. Keeps this PR's diff focused on the RBAC split. Ready for the merge call from my side.

jsenko (approved):
LGTM, thanks!

PR #9038, opened 2026-07-30
Issue 9034 kafkasql contract events

### LINKED ISSUE #9034: fix(kafkasql): add contract methods to RegistryStorage and fix missing ContractRulesetConfigured events
## Context

PR #8962 fixed contract metadata and status events being silently lost on KafkaSQL by routing them through `RegistryStorage.createEvent()`. That fix is correct but uses `createEvent()` as a backdoor from the REST layer — every other event in the system is fired by the storage implementation itself.

Additionally, there are 6 `ContractRulesetConfigured` event sites in `AbstractSqlRegistryStorage` that have the same gap: the SQL implementation fires `SqlOutboxEvent`, but the KafkaSQL implementation never fires `KafkaSqlOutboxEvent` after the journal round-trip.

## Problem

Contract operations bypass the storage abstraction for event publishing:

1. **Contract metadata/status events** — fired from `GroupsResourceImpl` (REST layer) instead of the storage layer. Fixed in #8962 via `storage.createEvent()`, but should follow the standard pattern.

2. **ContractRulesetConfigured events** — fired correctly in SQL storage but missing in KafkaSQL for all 6 methods:
   - `setArtifactContractRuleset()`
   - `deleteArtifactContractRuleset()`
   - `setVersionContractRuleset()`
   - `deleteVersionContractRuleset()`
   - `setGlobalContractRuleset()`
   - `deleteGlobalContractRuleset()`

## Proposed fix

Add contract-specific methods to the `RegistryStorage` interface (e.g. `updateContractMetadata()`, `transitionContractStatus()`). Each storage variant implements them following the established pattern:

- **SQL:** repository call + fire `SqlOutboxEvent`
- **KafkaSQL:** submit journal message → wait for response → fire `KafkaSqlOutboxEvent`

For the 6 `ContractRulesetConfigured` gaps, add the missing `outboxEvent.fire(KafkaSqlOutboxEvent.of(...))` calls after the journal round-trip in `KafkaSqlRegistryStorage`.

This removes event awareness from the REST layer and ensures all contract events are published consistently across storage variants.

## References

- #8928 — original bug report
- #8962 — interim fix (contract metadata/status events)
- Eric's architectural analysis: https://github.com/Apicurio/apicurio-registry/pull/8962#issuecomment-5127845498

singhsrijan46:
# fix(kafkasql): add contract methods to RegistryStorage and fix missing ContractRulesetConfigured events (#9034)

Closes #9034

## Context & Motivation

In PR #8962, contract metadata and status events were fixed on KafkaSQL by directly injecting `Event<SqlOutboxEvent>` into `GroupsResourceImpl` and firing `SqlOutboxEvent` from the REST layer. While this prevented event loss, it introduced two architectural issues:
1. **REST layer bypassing storage abstraction:** `GroupsResourceImpl` was directly aware of `SqlOutboxEvent` (a SQL-storage-specific class). When running on KafkaSQL, firing `SqlOutboxEvent` from the REST layer triggered local `SqlEventsProcessor` instead of routing `KafkaSqlOutboxEvent` through the storage layer and Kafka events topic.
2. **Missing `ContractRulesetConfigured` outbox events in KafkaSQL:** The 6 contract ruleset methods in `KafkaSqlRegistryStorage` did not fire `KafkaSqlOutboxEvent` after journal execution completed, whereas `AbstractSqlRegistryStorage` fired `SqlOutboxEvent` for all 6 methods.

### 1. Storage Abstraction (`RegistryStorage`)
- Added `updateContractMetadata(String groupId, String artifactId, String prefix, Map<String, String> labels)` and `transitionContractStatus(String groupId, String artifactId, String prevStatus, String targetStatus, String prefix, Map<String, String> statusLabels)` to `RegistryStorage`.
- Updated `ReadOnlyDelegatingStorage` to forward both methods to `delegateStorage()`.

### 2. Storage Outbox Event Implementations
- **SQL Storage (`AbstractSqlRegistryStorage`):** Implemented contract storage methods to handle label merges and dispatch `SqlOutboxEvent.of(ContractMetadataUpdated)` and `SqlOutboxEvent.of(ContractStatusChanged)`.
- **KafkaSQL Storage (`KafkaSqlRegistryStorage`):**
  - Added missing `KafkaSqlOutboxEvent.of(ContractRulesetConfigured)` events to all 6 contract ruleset methods (`setArtifactContractRuleset`, `deleteArtifactContractRuleset`, `setVersionContractRuleset`, `deleteVersionContractRuleset`, `setGlobalContractRuleset`, `deleteGlobalContractRuleset`).
  - Implemented `updateContractMetadata` and `transitionContractStatus` to perform atomic label merges and dispatch `KafkaSqlOutboxEvent.of(...)`.
  - Added private `fireContractOutboxEvent` helper to eliminate repetitive outbox event construction boilerplate.

### 3. Single-Message Atomic Kafka Operations (Performance)
- Combined contract status and lifecycle date labels under the contract prefix into a single `Map<String, String>` and performed 1 single atomic `mergeArtifactLabels` call in `transitionContractStatus`. This reduces Kafka journal round trips from **2–3 messages down to 1 single message per transition**.

### 4. REST Layer Decoupling (`GroupsResourceImpl`)
- Removed `@Inject Event<SqlOutboxEvent> contractOutboxEvent` field injection.
- Updated `updateContractMetadata` and `transitionContractStatus` to delegate label merging and outbox event publishing to `RegistryStorage`.

---

## Testing & Verification

- Verified clean git status and commit on branch `issue-9034-kafkasql-contract-events`.
- Verified 100% elimination of REST-layer `SqlOutboxEvent` dependencies.
- Verified single atomic message dispatch per contract status transition.


app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -901,6 +901,22 @@ public void deleteGlobalContractRuleset() throws RegistryStorageException {
...
+        mergeArtifactLabels(groupId, artifactId, prefix, labels);
+        outboxEvent.fire(SqlOutboxEvent.of(
+                io.apicurio.registry.events.ContractMetadataUpdated.of(groupId, artifactId)));
+    }
+
+    @Override
+    public void transitionContractStatus(String groupId, String artifactId, String prevStatus, String targetStatus,
+            String prefix, Map<String, String> statusLabels) throws RegistryStorageException {
+        mergeArtifactLabels(groupId, artifactId, prefix, statusLabels);
+        outboxEvent.fire(SqlOutboxEvent.of(
```
paoloantinori, line 918: The caller passes the contract `prefix` here, but `statusLabels` is a map of full keys (`prefix+SUFFIX_STATUS`, `prefix+SUFFIX_STABLE_DATE`, ...). Worth confirming `mergeArtifactLabels`'s prefix-scoped delete doesn't over-delete — the previous code called `mergeArtifactLabels` once per single key, each scoped to that one key.

paoloantinori (commented):
The architectural direction is right — moving outbox-event firing out of the REST layer and into the storage abstraction so KafkaSQL fires `KafkaSqlOutboxEvent` (routed through Kafka) instead of `SqlOutboxEvent` (processed locally) is the correct fix for both issues described in the PR body. A couple of correctness concerns worth confirming before this lands.

- **`transitionContractStatus` lost the per-key atomicity on SQL.** Before this PR, `GroupsResourceImpl.transitionContractStatus` called `mergeArtifactLabels` once per key (status, stable-date, deprecated-date). The new `AbstractSqlRegistryStorage.transitionContractStatus` (and `KafkaSqlRegistryStorage`) bundle all status labels into a *single* `mergeArtifactLabels(groupId, artifactId, prefix, statusLabels)` call. That is fine **only if** `prefix` here equals the per-key key it used to be passed. But the caller now passes the contract `prefix` (e.g. `contract.<id>.`) while the label map keys are full keys like `prefix + SUFFIX_STATUS`. Worth double-checking that `mergeArtifactLabels`'s prefix-delete semantics still correctly scope the delete to the right labels and don't, for example, wipe the entire contract prefix when only the status+date subset should be touched. The previous per-key calls were each scoped to a single key, so the behavior was unambiguous; the merged call changes the scope.

- **Atomicity gap between write and outbox fire.** In both SQL and KafkaSQL implementations, `updateContractMetadata`/`transitionContractStatus` call `mergeArtifactLabels(...)` then `outboxEvent.fire(...)` / `fireContractOutboxEvent(...)` as separate statements, outside any transaction. If the process crashes between the label merge and the event fire, the labels are persisted but no event is emitted — a silently lost notification. The pre-existing SQL code had the same shape, so this is not a regression introduced here, but since the PR is consolidating this logic in the storage layer it would be a good moment to note whether the SQL `outboxEvent` mechanism is transactional with the label write (same DB transaction) or fire-and-forget. For KafkaSQL the gap is more material because `mergeArtifactLabels` journals a Kafka message and the outbox fire is a separate step.

- **`ReadOnlyDelegatingStorage` forwarders.** The two new overrides correctly delegate to `delegate(...)`, consistent with the surrounding `getDeprecationReadiness` etc. Fine.

- **Unrelated UI change.** `CreateArtifactModal.tsx` removes a `<FormGroup label="Type" ...>` line but leaves the nested `<Select>` and its closing tag, which looks like it will break the form's structure (an orphaned Select outside its FormGroup). This appears unrelated to the KafkaSQL contract-events fix described in the PR title/body. Worth confirming this was intentional or splitting it out.

The diff was fully visible (6 files, ~100 lines).

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -741,15 +749,34 @@ public void setGlobalContractRuleset(io.apicurio.registry.storage.dto.ContractRu
...
+
+
+    @Override
+    public void updateContractMetadata(String groupId, String artifactId, String prefix,
+            Map<String, String> labels) throws RegistryStorageException {
+        mergeArtifactLabels(groupId, artifactId, prefix, labels);
+        fireContractOutboxEvent(io.apicurio.registry.events.ContractMetadataUpdated.of(groupId, artifactId));
     }
 
+    @Override
```
paoloantinori, line 771: `mergeArtifactLabels(...)` (journals a Kafka message) is followed by `fireContractOutboxEvent(...)` as a separate step. If the process dies between them, the label change is journaled but no outbox event fires — a silently lost notification. Not a regression (old SQL path had the same shape), but worth a note on whether the outbox fire should be part of the journaled response.

ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
```diff
@@ -451,7 +451,6 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> =
                                 </FormHelperText>
                             </If>
```
paoloantinori, line 452: This removes the `<FormGroup label="Type">` wrapper but leaves the nested `<Select>` and its closing `</FormGroup>`-equivalent, which looks like it leaves an orphaned Select and may break the Create Artifact modal's Type field. This change seems unrelated to the KafkaSQL contract-events fix — worth confirming or splitting out.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
singhsrijan46: Fixed transitionContractStatus to use per-key mergeArtifactLabels calls so only status/date keys are touched, preserving all other contract metadata labels.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
singhsrijan46: Verified that this preserves the pre-existing post-write event pattern. Including outbox payloads directly in journal messages would be a great follow-up hardening task for KafkaSQL.

ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
singhsrijan46: Reverted! This was an artifact from merge conflict resolution — CreateArtifactModal.tsx is now restored to upstream/main.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -741,15 +749,37 @@ public void setGlobalContractRuleset(io.apicurio.registry.storage.dto.ContractRu
...
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        fireContractOutboxEvent(io.apicurio.registry.events.ContractRulesetConfigured.of("__GLOBAL__", "__GLOBAL__", null, "DELETE"));
     }
 
+
+    @Override
+    public void updateContractMetadata(String groupId, String artifactId, String prefix,
+            Map<String, String> labels) throws RegistryStorageException {
+        mergeArtifactLabels(groupId, artifactId, prefix, labels);
```
paoloantinori, line 767: On KafkaSQL, the other contract methods (e.g. `setArtifactContractRuleset` above) route the write through `submitter.submitMessage(...)` + `coordinator.waitForResponse(uuid)` so it lands in the Kafka journal and replicates across nodes. This path calls `mergeArtifactLabels(...)` directly instead. If `mergeArtifactLabels` on the KafkaSQL path writes only to the local SQL store (and doesn't journal a message), then contract metadata updates won't replicate to other nodes, while the `KafkaSqlOutboxEvent` fired on the next line does fan out cluster-wide — replicas would process the event but not have the underlying labels. Worth confirming how KafkaSQL's `mergeArtifactLabels` is implemented: if it journals, this is fine; if it's the direct SQL write, this is a cross-node consistency gap.

paoloantinori (commented):
The refactor lands cleanly on this head — the REST layer no longer fires SQL-specific `SqlOutboxEvent`, and both storage impls now own the outbox event for contract metadata/status. One thing worth a careful look on the KafkaSQL path:

- **Do the new contract metadata/status writes journal through Kafka?** The other six contract methods in `KafkaSqlRegistryStorage` (e.g. `setArtifactContractRuleset`) achieve replication by routing the write through `submitter.submitMessage(...)` + `coordinator.waitForResponse(uuid)`, which appends to the Kafka journal so every node in the cluster applies it. By contrast, the two new overrides here call `mergeArtifactLabels(...)` directly and then fire the outbox event. If `mergeArtifactLabels` on the KafkaSQL path writes only to the local node's SQL store (i.e. it doesn't journal), then contract metadata/status changes would not replicate to other nodes — the `KafkaSqlOutboxEvent` would be emitted cluster-wide but the underlying labels wouldn't be present on replicas processing the event. I couldn't confirm from this diff whether KafkaSQL's `mergeArtifactLabels` journals; if it does delegate to a journaled message, this is a non-issue — but if it hits the SQL store directly (as in `AbstractSqlRegistryStorage`), this is a real cross-node consistency gap. Worth verifying against the KafkaSQL `mergeArtifactLabels` implementation.
- **Secondary, lower priority:** in both `transitionContractStatus` impls (SQL and KafkaSQL), the status labels are applied via a per-key `mergeArtifactLabels` loop rather than a single prefix-scoped merge. That's two+ separate writes for a status transition plus its lifecycle date, so a failure mid-loop can leave a partial state (status flipped to `DEPRECATED` but the `deprecated-date` label not yet written). Not a regression — the original REST-layer code had the same shape — but now that all the labels are collected into `statusLabels` upfront, a single atomic `mergeArtifactLabels(groupId, artifactId, prefix, statusLabels)` (after threading `prefix` in) would be both simpler and transactional.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
singhsrijan46: `KafkaSqlRegistryStorage.mergeArtifactLabels` does journal through Kafka via `MergeArtifactLabels4Message` (`submitter.submitMessage(...)` + `coordinator.waitForResponse(uuid)`). As a result, label updates replicate cluster-wide across all nodes before the outbox event is emitted.

singhsrijan46:
Thank you @paoloantinori!

- **KafkaSQL Replication:** Verified that `KafkaSqlRegistryStorage.mergeArtifactLabels` journals through Kafka (`MergeArtifactLabels4Message`), ensuring cluster-wide replication.
- **Per-Key Scope:** The per-key loop in `transitionContractStatus` is intentional so `mergeArtifactLabels`'s prefix-delete (`prefix + '%'`) doesn't wipe other contract metadata labels (description, stage, etc.).
- **Outbox Atomicity:** Preserves the existing post-write event pattern. Direct journal outbox payloads would be a great follow-up task.
- **UI Change:** Reverted `CreateArtifactModal.tsx` to `upstream/main`.

carlesarnal:
@singhsrijan46 as stated in the [contributing](https://github.com/Apicurio/apicurio-registry/blob/main/CONTRIBUTING.md#claiming-an-issue) guide, you must wait for an issue to be assigned to you to submit a PR. In the spirit of that, I'm closing this PR.

Issue #8427, opened 2026-07-01
feat: MCP tool validation and compatibility checking


carlesarnal:
## Summary

Add validation and compatibility checking for MCP_TOOL artifacts — the same kind of schema governance Apicurio provides for Avro, Protobuf, and JSON Schema.

## Scope

- MCP spec validation on artifact create/update (required fields, valid JSON Schema for input/output)
- Input/output compatibility checking between tools (Tool A output → Tool B input)
- "Compatible Tools" API endpoint: given a tool, return compatible tools in the registry
- Compatibility view in the UI: select a tool and see which others can consume its output
- Unit and integration tests for validation rules and compatibility checks

## Context

This is proposed as an [LFX Mentorship](https://mentorship.lfx.linuxfoundation.org/) project for 2026 Term 3 (Sep-Nov). Builds on existing `MCP_TOOL` artifact type and `/.well-known/mcp-tools` endpoints.

malladinagarjuna2:
Hi @carlesarnal — I'd like to be considered for this as an LFX 2026 Term 3 mentee.

I've been contributing to [Kuadrant/mcp-gateway](https://github.com/Kuadrant/mcp-gateway) (7 PRs merged), mostly around the MCP tool lifecycle rather than peripheral fixes:
where my work spans around
 removing  the MCP lock from validateServerTools, so I've worked directly inside the tool-validation path.
added enable/disable support for MCP servers/tools.
added test coverage for non-default path registration.
#JWT HS256 key-length condition (32-byte minimum).
So I'm already familiar with how MCP tool definitions are discovered, registered, and validated at the gateway layer — this issue is the registry-side counterpart of that, which is why it's interesting to me.

ADITYA-CODE-SOURCE:
Hi @carlesarnal, I reviewed the current `MCP_TOOL` implementation and the related open MCP issues/PRs.

It looks like the main remaining gaps here are:
1. deep JSON Schema validation for embedded `inputSchema` / `outputSchema`
2. tool-to-tool compatibility (`outputSchema -> inputSchema`)
3. a compatible-tools endpoint
4. a UI compatibility view

I’d like to start with the validation slice first by improving `McpToolContentValidator` to perform deeper validation of embedded `inputSchema` and `outputSchema`, plus focused unit tests.

This should avoid overlap with `#8662`, `#8619`, `#8538`, and `#8634`.

If that sounds good, I can open a focused PR for this part first.

Charul192:
[...]

PR #9045, opened 2026-07-30
feat(mcp): add Compatible Tools API endpoint and UI compatibility view

Satvik77777:
Part of #8427

## Summary
Adds output→input compatibility checking between MCP tools: given a tool, 
find other registered tools whose required input parameters are 
guaranteed by this tool's required output schema fields (i.e. "can Tool 
A's output feed Tool B's input").

## What's included
- **Backend**: two new endpoints on the well-known MCP tools resource
  - `GET /.well-known/mcp-tools/{groupId}/{artifactId}/compatible` — 
    by artifact reference
  - `POST /.well-known/mcp-tools/compatible` — by raw tool content
  - Compatibility logic (`isOutputCompatibleWithInput`) checks each 
    candidate's required input fields against the target's required 
    output fields (name + type match, with integer→number widening 
    allowed)
  - Integration tests covering: compatible match, incompatible (missing 
    required field), and the edge case where a target's output field 
    exists in `properties` but isn't listed in `required` (should not 
    count as satisfying a candidate's required input)

- **UI**: `CompatibleMcpToolsViewer` component, mounted in the MCP tool 
  documentation tab (`McpToolVisualizer`) alongside the existing tool 
  viewer. Shows loading, empty, and populated states; wired to the real 
  endpoint above, not mocked data.

## Design notes
- This does **not** reuse `McpToolCompatibilityChecker` (from #8662) — 
  that checker compares the same schema field across two versions of 
  one tool (version-evolution compatibility). This feature compares a 
  different field across two different tools at a single point in time 
  (cross-tool chaining compatibility). Different problem shape, so a 
  separate, purpose-built comparison was needed.
- Known scope boundary: type compatibility is currently name + top-level 
  type match only (with integer→number widening), not full nested/
  structural JSON Schema comparison. Flagging as a deliberate v1 
  boundary, open to feedback on whether deeper structural comparison is 
  expected for this issue.
- Known follow-up: the endpoint currently does a sequential per-candidate 
  storage fetch + compatibility check across all registered MCP_TOOL 
  artifacts (bounded by `MAX_VISIBILITY_FILTER_RESULTS`). Fine at current 
  scale; flagging as a known O(N) characteristic if this becomes a 
  concern at higher artifact counts.

## Test plan
- `./mvnw -pl app -am test -Dtest=WellKnownMcpToolsTest` — 17/17 passing
- `npx vitest run` (ui-app) — all passing
- `npx tsc --noEmit` (ui-app) — clean, exit code 0

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
@@ -558,6 +558,197 @@ public class WellKnownResourceImpl implements WellKnownResource {
         return McpToolSearchResults.builder().count(results.getCount()).tools(tools).build();
     }
 
+    @Override
+    @Authorized(style = AuthorizedStyle.GroupAndArtifact, level = AuthorizedLevel.Read)
+    public McpToolSearchResults getCompatibleMcpTools(String groupId, String artifactId,
+            String version, String offset, String limit) {
+        if (!mcpToolsConfig.isEnabled()) {
+            throw new NotFoundException("MCP tools support is disabled");
+        }
+
+        int safeOffset = Math.max(0, parsePaginationParam(offset, "offset", 0));
+        int safeLimit = Math.max(1, Math.min(parsePaginationParam(limit, "limit", 20), 500));
+
+        GroupId gid = new GroupId(groupId);
+        String rawGroupId = gid.getRawGroupIdWithNull();
+        GA ga = new GA(rawGroupId, artifactId);
+
+        StoredArtifactVersionDto targetArtifact;
+        try {
+            String versionExpression = StringUtil.isEmpty(version) ? "branch=latest" : version;
+            GAV gav = VersionExpressionParser.parse(ga, versionExpression,
+                    (g, branchId) -> storage.getBranchTip(g, branchId,
+                            RetrievalBehavior.SKIP_DISABLED_LATEST));
+
+            targetArtifact = storage.getArtifactVersionContent(
+                    gav.getRawGroupIdWithNull(), gav.getRawArtifactId(), gav.getRawVersionId());
+
+            ArtifactVersionMetaDataDto metadata = storage.getArtifactVersionMetaData(
+                    gav.getRawGroupIdWithNull(), gav.getRawArtifactId(), gav.getRawVersionId());
+
+            if (!ArtifactType.MCP_TOOL.equals(metadata.getArtifactType())) {
+                throw new NotFoundException("Artifact is not an MCP tool definition");
+            }
+        } catch (ArtifactNotFoundException | VersionNotFoundException e) {
+            throw new NotFoundException("MCP tool not found: " + groupId + "/" + artifactId);
+        }
+
+        String targetContentStr = targetArtifact.getContent().content();
+        return findCompatibleToolsForContent(targetContentStr, ga, safeOffset, safeLimit);
+    }
+
+    @Override
+    @Authorized(style = AuthorizedStyle.None, level = AuthorizedLevel.Read)
+    public McpToolSearchResults findCompatibleMcpTools(String mcpToolContent, String offset, String limit) {
+        if (!mcpToolsConfig.isEnabled()) {
+            throw new NotFoundException("MCP tools support is disabled");
+        }
+
+        if (StringUtil.isEmpty(mcpToolContent)) {
+            throw new BadRequestException("MCP tool content must not be empty");
+        }
+
+        int safeOffset = Math.max(0, parsePaginationParam(offset, "offset", 0));
+        int safeLimit = Math.max(1, Math.min(parsePaginationParam(limit, "limit", 20), 500));
+
+        return findCompatibleToolsForContent(mcpToolContent, null, safeOffset, safeLimit);
+    }
+
+    private McpToolSearchResults findCompatibleToolsForContent(String targetContentStr, GA targetGa,
+            int safeOffset, int safeLimit) {
+        Set<SearchFilter> filters = new HashSet<>();
+        filters.add(SearchFilter.ofArtifactType(ArtifactType.MCP_TOOL));
+
+        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, OrderBy.createdOn,
+                OrderDirection.desc, 0, MAX_VISIBILITY_FILTER_RESULTS, false);
+
+        JsonNode targetOutputSchema = null;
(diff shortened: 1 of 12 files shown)


app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -558,6 +558,197 @@ public McpToolSearchResults searchMcpTools(String name, List<String> parameters,
...
+        if (StringUtil.isEmpty(mcpToolContent)) {
+            throw new BadRequestException("MCP tool content must not be empty");
+        }
+
+        int safeOffset = Math.max(0, parsePaginationParam(offset, "offset", 0));
+        int safeLimit = Math.max(1, Math.min(parsePaginationParam(limit, "limit", 20), 500));
+
+        return findCompatibleToolsForContent(mcpToolContent, null, safeOffset, safeLimit);
+    }
+
```
paoloantinori, line 616: This loads up to `MAX_VISIBILITY_FILTER_RESULTS` MCP tools and fetches+parses each candidate's latest version content on every call, then applies offset/limit after the in-memory filter. Pagination doesn't reduce the work (page 2 re-scans everything). Worth documenting the scaling expectation, and confirming whether `MAX_VISIBILITY_FILTER_RESULTS` silently truncates the candidate set — if the registry has more MCP tools than the cap, matches beyond it are never found.

paoloantinori (commented):
The feature is a nice addition to the MCP tooling surface — output→input compatibility for tool chaining, with a clean required-field-name + type-match rule (and the integer→number widening) and good test coverage of the optional-vs-required edge case. A few concerns worth addressing before merge.

- **`findCompatibleToolsForContent` loads every MCP tool into memory to filter.** It calls `storage.searchArtifacts(...)` with only the `ArtifactType.MCP_TOOL` filter and `MAX_VISIBILITY_FILTER_RESULTS` as the limit, then loops over every result, fetching each candidate's latest version content and parsing its JSON. This is an O(N) full-scan with per-candidate storage reads on every call. With offset/limit applied *after* the full in-memory filter, pagination doesn't reduce the work — page 2 re-scans everything. For a registry with many MCP tools this will be slow and memory-heavy. Worth at least documenting the scaling expectation, and ideally noting whether `MAX_VISIBILITY_FILTER_RESULTS` is a hard cap that silently truncates the candidate set (if there are more MCP tools than the cap, compatibility matches beyond the cap are never found).

- **Candidate failures are swallowed silently.** Inside the loop, any exception reading/parsing a candidate is caught and only `log.warn`-ed, then the loop continues. A single malformed candidate tool definition won't break the search, which is good for robustness, but it also means a candidate that *would* have matched is invisibly omitted from results with no signal to the caller. Consider whether a parse failure of the *target* tool's content (where `targetOutputSchema` ends up null) should be a hard error rather than degrading to 'compatible with nothing' — currently a malformed target silently returns zero results instead of a 400.

- **`findCompatibleMcpTools` (POST) authorization is `AuthorizedStyle.None`.** The GET-by-reference path uses `GroupAndArtifact`/`Read`, but the POST path that accepts arbitrary raw tool content has no authorization gate beyond `mcpToolsConfig.isEnabled()`. Worth confirming this is intentional (the POST doesn't reference a stored artifact, so `None` may be correct) and that the underlying `searchArtifacts` visibility filtering handles access scoping.

- **`isOutputCompatibleWithInput` type check only fires when *both* sides declare a `type`.** If the candidate's required input property has no `type` (common in real schemas), the name-match alone passes — so a candidate requiring `{"config": <no type>}` would match a target outputting `{"config": string}`. That may be acceptable (name-guaranteed is the headline guarantee), but it's worth being explicit in the design notes that type is a *tightening* constraint, not a required one.

The diff was fully visible (12 files).

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -558,6 +558,197 @@ public McpToolSearchResults searchMcpTools(String name, List<String> parameters,
...
+        }
+
+        List<McpToolSearchResult> compatibleTools = new ArrayList<>();
+
+        for (SearchedArtifactDto artifact : results.getArtifacts()) {
+            // Exclude the target artifact itself if targetGa is specified
+            if (targetGa != null
+                    && java.util.Objects.equals(artifact.getGroupId(), targetGa.getRawGroupIdWithNull())
+                    && java.util.Objects.equals(artifact.getArtifactId(), targetGa.getRawArtifactId())) {
+                continue;
```
paoloantinori, line 640: Candidate read/parse failures are `log.warn`-ed and skipped — a candidate that would have matched is invisibly omitted. More importantly, if the *target* content fails to parse, `targetOutputSchema` stays null and the method silently returns zero results instead of surfacing a 400. Consider treating target-parse failure as a hard error.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -558,6 +558,197 @@ public McpToolSearchResults searchMcpTools(String name, List<String> parameters,
...
+                    gav.getRawGroupIdWithNull(), gav.getRawArtifactId(), gav.getRawVersionId());
+
+            if (!ArtifactType.MCP_TOOL.equals(metadata.getArtifactType())) {
+                throw new NotFoundException("Artifact is not an MCP tool definition");
+            }
+        } catch (ArtifactNotFoundException | VersionNotFoundException e) {
+            throw new NotFoundException("MCP tool not found: " + groupId + "/" + artifactId);
+        }
+
+        String targetContentStr = targetArtifact.getContent().content();
```
paoloantinori, line 596: POST `findCompatibleMcpTools` uses `AuthorizedStyle.None` while the GET-by-reference path uses `GroupAndArtifact`/`Read`. Worth confirming `None` is intentional here (the POST takes raw content, not a stored artifact ref) and that `searchArtifacts` visibility filtering still scopes the candidate scan correctly.

Satvik77777:
Closing this — the branch had an extra commit from earlier exploratory 
   work that duplicated logic already covered by #8662, so I rebuilt it 
   on a clean branch with just the two relevant commits.

Satvik77777:
@paoloantinori - Continuing this conversation on #9049 (reopened from a clean branch, 
same code) — I've carried your three points into the new PR description 
under "Known limitations" so they're not lost, and I'll address them 
there. Thanks for the thorough review.

PR #9046, opened 2026-07-30
fix(ui): use cd instead of --prefix to fix windows recursion bug

### LINKED ISSUE #9036: UI: npm install in ui/ recurses infinitely on Windows (--prefix resolves package.json from cwd)
## Description

**Registry Version**: snapshot (build from `main`, `3.3.2-Dev`)
**Persistence type**: N/A — developer setup / UI build issue

Running the documented UI setup step `cd ui && npm install` on Windows never
completes. It spawns roughly 15 nested `npm install` processes, each writing its own
debug log, before failing with `code 1`.

### Root cause

`ui/package.json` defines:

    "postinstall": "npm install --prefix ../typescript-sdk && npm install --prefix ./ui-app && npm install --prefix ./ui-docs && npm install --prefix ./ui-editors"

On Windows, `npm install --prefix <dir>` reads `package.json` from the **current
working directory**, not from the prefix directory. So `npm install --prefix
../typescript-sdk` executed from `ui/` re-reads `ui/package.json`, which fires `ui`'s
own `postinstall` again — and recurses until npm gives up.

Using an absolute path, or moving the flag ahead of the subcommand
(`npm --prefix ../typescript-sdk install`), makes no difference. Both still resolve
`@apicurio/registry-ui`.

`cd`-ing into the target directory resolves correctly, so a fix could replace the
`--prefix` calls with `cd`-scoped ones, or move the chain into a Node script under
`ui/.scripts/` (the repo already depends on `shelljs`). Happy to open a PR once my
current one (#9012) is merged, per the one-PR-at-a-time rule.

This is distinct from #8975, which is a Linux failure caused by the unbuilt
typescript-sdk.

### Environment

- **OS**: Windows 11 (10.0.26200)
- **Node**: v22.19.0
- **npm**: 11.3.0
- **Shell**: npm invokes scripts via `cmd.exe /d /s /c`

Not reproducible in CI: `verify-build.yaml` runs `npm install` with
`working-directory: ui` on Linux, where `--prefix` resolves correctly. CI uses Node
24.18.0 (npm 11.x), the same npm major that fails here, so this appears to be
OS-specific rather than npm-version-specific.

### Steps to Reproduce

1. Clone the repository on Windows.
2. `cd ui`
3. `npm install`
4. Observe repeated nested `npm error` output, and ~15 debug logs written to
   `%LocalAppData%\npm-cache\_logs`.

**Expected:** `cd ui && npm install` installs the SDK and the three UI packages once
each, as it does on Linux and macOS.

**Actual:** the `postinstall` script re-invokes itself until npm aborts, and no
packages are installed.

### Logs

Every nested failure reports the same path and command, confirming the loop:

    npm error code 1
    npm error path C:\...\apicurio-registry\ui
    npm error command failed
    npm error command C:\WINDOWS\system32\cmd.exe /d /s /c npm install --prefix ../typescript-sdk && npm install --prefix ./ui-app && npm install --prefix ./ui-docs && npm install --prefix ./ui-editors

Confirmed with `--dry-run` from `ui/` — note it resolves `@apicurio/registry-ui`,
which is `ui`'s own package, rather than the SDK's dependencies:

    > npm install --prefix ../typescript-sdk --dry-run --ignore-scripts
    add @apicurio/registry-ui 3.3.2-Dev
    add yocto-queue 0.1.0
    ...

Whereas `cd`-ing into the target resolves correctly:

    > cd ../typescript-sdk && npm install --dry-run --ignore-scripts
    add yocto-queue 0.1.0
    ...
    (no @apicurio/registry-ui)

shivansh-gohem:
Fixes #9036

This replaces the usage of `--prefix` with explicit directory changes (`cd` sub-shells) in the UI `package.json` scripts. This resolves the infinite recursion bug that occurs during `npm install` on Windows, without affecting Linux/macOS builds.

Since the original issue reporter (@DWilson32) is unable to submit a PR at this time, I am submitting this fix based on their findings.


DWilson32:
No objection at all, go ahead — glad it's moving. I couldn't open it myself since I
already have #9012 open and the project asks for one PR at a time.

Happy to verify on Windows if that's useful. I still have the reproduction set up, so
I can confirm whether `npm install` from `ui/` completes cleanly with this change.

DWilson32:
Verified on Windows — this fixes it.

Checked out the branch and ran `npm install` from `ui/` on Windows 11, Node v22.19.0,
npm 11.3.0. Completes cleanly in ~2 minutes, exit 0, all four packages installed. No
recursion, no nested npm processes. Same environment that reproduced the original bug
in #9036.

One nit: the change drops the trailing newline at the end of `ui/package.json`.

Nice approach splitting the chain into named `install-*` / `clean-*` / `lint-*` scripts
rather than inlining the `cd`s — reads better than what was there before.

EricWittmann:
I'm surprised that this approach is better for Windows than using `--prefix` but I'm happy if it does.  :)

DWilson32:
Yeah, surprised me too. Turns out `--prefix` doesn't change which package.json npm reads.

Dry run from ui/:

    > npm install --prefix ../typescript-sdk --dry-run --ignore-scripts
    add @apicurio/registry-ui 3.3.2-Dev
    ...

That's ui's own package, not the SDK's deps. So the postinstall re-reads ui/package.json
and fires itself again. Absolute path and `npm --prefix ../typescript-sdk install` do the
same thing, so it's not a path quirk.

cd works because the cwd actually changes.

Only checked on Windows / npm 11.3.0 though — didn't dig into why npm does it.

EricWittmann:
Thanks @shivansh-gohem for picking this up and getting it submitted! And thanks to @DWilson32 for the Windows verification. Nice fix.

PR #9049, opened 2026-07-30
feat(mcp): add Compatible Tools API endpoint and UI compatibility view

Satvik77777:
Part of #8427

Supersedes #9045 — closed and reopened from a clean branch to remove an 
unrelated exploratory commit that duplicated logic covered by #8662. 
Same two commits, no functional changes from what @paoloantinori already 
reviewed there.

## Summary
Adds output→input compatibility checking between MCP tools: given a tool, 
find other registered tools whose required input parameters are 
guaranteed by this tool's required output schema fields (i.e. "can Tool 
A's output feed Tool B's input").

## What's included
- **Backend**: two new endpoints on the well-known MCP tools resource
  - `GET /.well-known/mcp-tools/{groupId}/{artifactId}/compatible` — 
    by artifact reference
  - `POST /.well-known/mcp-tools/compatible` — by raw tool content
  - Compatibility logic (`isOutputCompatibleWithInput`) checks each 
    candidate's required input fields against the target's required 
    output fields (name + type match, with integer→number widening 
    allowed)
  - Integration tests covering: compatible match, incompatible (missing 
    required field), and the edge case where a target's output field 
    exists in `properties` but isn't listed in `required` (should not 
    count as satisfying a candidate's required input)

- **UI**: `CompatibleMcpToolsViewer` component, mounted in the MCP tool 
  documentation tab (`McpToolVisualizer`) alongside the existing tool 
  viewer. Shows loading, empty, and populated states; wired to the real 
  endpoint above, not mocked data.

## Design notes
- This does **not** reuse `McpToolCompatibilityChecker` (from #8662) — 
  that checker compares the same schema field across two versions of 
  one tool (version-evolution compatibility). This feature compares a 
  different field across two different tools at a single point in time 
  (cross-tool chaining compatibility). Different problem shape, so a 
  separate, purpose-built comparison was needed.
- Name-match is the headline guarantee; type-match only applies when 
  both sides declare a type, so it's a tightening constraint on top of 
  name matching, not a strict requirement.
- Type compatibility is currently name + top-level type match only 
  (with integer→number widening), not full nested/structural JSON 
  Schema comparison. Open to feedback on whether deeper structural 
  comparison is expected here.

## Known limitations (flagged by @paoloantinori on #9045, still open)
- `findCompatibleToolsForContent` does an O(N) in-memory scan: fetches 
  and parses every registered MCP_TOOL's latest version on each call, 
  with offset/limit applied after the full scan — pagination doesn't 
  reduce the work, and results beyond `MAX_VISIBILITY_FILTER_RESULTS` 
  are never found. Will follow up on a storage-level filtering approach 
  if that's expected before merge.
- A malformed *target* tool currently degrades to zero results instead 
  of a 400 — planning to make target-parse failure a hard error instead.
- Candidate read/parse failures are logged and skipped silently, which 
  could hide an otherwise-matching candidate with no signal to the 
  caller — open to suggestions on the right way to surface this.
- `POST /.well-known/mcp-tools/compatible` uses `AuthorizedStyle.None` 
  since it takes raw content rather than referencing a stored artifact — 
  flagging for confirmation this is the right posture, given `GET` uses 
  `GroupAndArtifact/Read`.

## Test plan
- `./mvnw -pl app -am test -Dtest=WellKnownMcpToolsTest` — 17/17 passing
- `npx vitest run` (ui-app) — all passing
- `npx tsc --noEmit` (ui-app) — clean, exit code 0

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java
@@ -1,6 +1,5 @@
 package io.apicurio.registry.rest.wellknown;
 
-import io.apicurio.registry.mcptools.rest.beans.McpCompatibleToolsResults;
 import io.apicurio.registry.rest.v3.beans.AgentCard;
 import io.apicurio.registry.rest.v3.beans.AgentSearchRequest;
 import io.apicurio.registry.rest.v3.beans.AgentSearchResults;
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java
@@ -170,7 +170,6 @@ public interface WellKnownResource {
             @QueryParam("limit") @DefaultValue("20") String limit);
 
     /**
-    /**
      * Search for registered MCP tools compatible with a specific target tool by group and artifact ID.
      * Tool chaining compatibility is evaluated by matching the target tool's outputSchema guaranteed
      * (required) output properties against candidate tools' inputSchema required input parameters.
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java
@@ -171,34 +171,44 @@ public interface WellKnownResource {
             @QueryParam("limit") @DefaultValue("20") String limit);
 
     /**
-     * Returns all registered MCP tools whose {@code inputSchema} can accept the output
-     * produced by the given source tool's {@code outputSchema}.
+    /**
+     * Search for registered MCP tools compatible with a specific target tool by group and artifact ID.
+     * Tool chaining compatibility is evaluated by matching the target tool's outputSchema guaranteed
+     * (required) output properties against candidate tools' inputSchema required input parameters.
      *
-     * <p>Two tools are considered compatible when every property declared in the source
-     * tool's {@code outputSchema.properties} is also present in the candidate tool's
-     * {@code inputSchema.properties} with the same JSON Schema type. This models the
-     * pipeline chaining contract: the candidate tool can consume what the source tool
-     * produces.</p>
-     *
-     * <p>If the source tool has no {@code outputSchema}, an empty result is returned.
-     * The source tool itself is never included in the results.</p>
-     *
-     * @param groupId    the group ID of the source MCP tool artifact
-     * @param artifactId the artifact ID of the source MCP tool
-     * @param version    optional version expression (defaults to latest)
-     * @param offset     pagination offset
-     * @param limit      pagination limit
-     * @return the compatible MCP tools
+     * @param groupId the group ID of the target MCP tool
+     * @param artifactId the artifact ID of the target MCP tool
+     * @param version optional version (defaults to latest)
+     * @param offset pagination offset
+     * @param limit pagination limit
+     * @return search results containing matching compatible MCP tools
      */
     @GET
     @Path("/mcp-tools/{groupId}/{artifactId}/compatible")
     @Produces(MediaType.APPLICATION_JSON)
-    McpCompatibleToolsResults findCompatibleTools(
+    McpToolSearchResults getCompatibleMcpTools(
             @PathParam("groupId") String groupId,
             @PathParam("artifactId") String artifactId,
             @QueryParam("version") String version,
-            @QueryParam("offset") @DefaultValue("0") Integer offset,
-            @QueryParam("limit") @DefaultValue("20") Integer limit);
+            @QueryParam("offset") @DefaultValue("0") String offset,
+            @QueryParam("limit") @DefaultValue("20") String limit);
+
+    /**
+     * Search for registered MCP tools compatible with a given MCP tool definition content.
+     *
+     * @param mcpToolContent raw MCP tool JSON definition
+     * @param offset pagination offset
+     * @param limit pagination limit
+     * @return search results containing matching compatible MCP tools
+     */
+    @POST
+    @Path("/mcp-tools/compatible")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    McpToolSearchResults findCompatibleMcpTools(
+            String mcpToolContent,
+            @QueryParam("offset") @DefaultValue("0") String offset,
+            @QueryParam("limit") @DefaultValue("20") String limit);
 
     /**
      * Returns the JSON Schema for a specific LLM artifact type.
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
@@ -589,6 +589,218 @@ public class WellKnownResourceImpl implements WellKnownResource {
         return McpToolSearchResults.builder().count((int) results.getCount()).tools(tools).build();
     }
 
+    @Override
+    @Authorized(style = AuthorizedStyle.GroupAndArtifact, level = AuthorizedLevel.Read)
+    public McpToolSearchResults getCompatibleMcpTools(String groupId, String artifactId,
+            String version, String offset, String limit) {
+        if (!mcpToolsConfig.isEnabled()) {
+            throw new NotFoundException("MCP tools support is disabled");
+        }
+
+        int safeOffset = Math.max(0, parsePaginationParam(offset, "offset", 0));
+        int safeLimit = Math.max(1, Math.min(parsePaginationParam(limit, "limit", 20), 500));
+
+        GroupId gid = new GroupId(groupId);
+        String rawGroupId = gid.getRawGroupIdWithNull();
+        GA ga = new GA(rawGroupId, artifactId);
+
+        StoredArtifactVersionDto targetArtifact;
+        try {
+            String versionExpression = StringUtil.isEmpty(version) ? "branch=latest" : version;
+            GAV gav = VersionExpressionParser.parse(ga, versionExpression,
+                    (g, branchId) -> storage.getBranchTip(g, branchId,
+                            RetrievalBehavior.SKIP_DISABLED_LATEST));
+
+            targetArtifact = storage.getArtifactVersionContent(
+                    gav.getRawGroupIdWithNull(), gav.getRawArtifactId(), gav.getRawVersionId());
+
+            ArtifactVersionMetaDataDto metadata = storage.getArtifactVersionMetaData(
+                    gav.getRawGroupIdWithNull(), gav.getRawArtifactId(), gav.getRawVersionId());
+
+            if (!ArtifactType.MCP_TOOL.equals(metadata.getArtifactType())) {
+                throw new NotFoundException("Artifact is not an MCP tool definition");
+            }
+        } catch (ArtifactNotFoundException | VersionNotFoundException e) {
+            throw new NotFoundException("MCP tool not found: " + groupId + "/" + artifactId);
+        }
+
+        String targetContentStr = targetArtifact.getContent().content();
+        return findCompatibleToolsForContent(targetContentStr, ga, safeOffset, safeLimit);
+    }
+
+    @Override
+    // POST accepts raw tool content rather than a stored artifact reference, so AuthorizedStyle.None applies.
+    // Note: storage.searchArtifacts does not perform per-group visibility filtering internally — the scoping
+    // boundary here is the caller's general registry Read access from @Authorized above, consistent with the
+    // existing /search/artifacts, /search/versions, and /.well-known/agents endpoints. See
+    // findCompatibleToolsForContent Javadoc for details.
+    @Authorized(style = AuthorizedStyle.None, level = AuthorizedLevel.Read)
+    public McpToolSearchResults findCompatibleMcpTools(String mcpToolContent, String offset, String limit) {
+        if (!mcpToolsConfig.isEnabled()) {
+            throw new NotFoundException("MCP tools support is disabled");
+        }
+
+        if (StringUtil.isEmpty(mcpToolContent)) {
+            throw new BadRequestException("MCP tool content must not be empty");
+        }
+
+        int safeOffset = Math.max(0, parsePaginationParam(offset, "offset", 0));
+        int safeLimit = Math.max(1, Math.min(parsePaginationParam(limit, "limit", 20), 500));
+
+        return findCompatibleToolsForContent(mcpToolContent, null, safeOffset, safeLimit);
+    }
+
+    /**
+     * Performs an O(N) in-memory compatibility scan over registered MCP tools.
+     * Note: Pagination (offset/limit) is applied in-memory after fetching candidates and does not reduce storage load.
+     * MAX_VISIBILITY_FILTER_RESULTS imposes a hard cap on the candidate set; tools beyond this cap are not considered.
+     *
+     * Visibility scoping note: storage.searchArtifacts() filters only by ArtifactType.MCP_TOOL. It does NOT apply
+     * per-group read-visibility filtering internally. The authorization boundary for this endpoint (matching existing
+     * endpoints such as /apis/registry/v3/search/artifacts, /apis/registry/v3/search/versions, and /.well-known/agents)
+     * is @Authorized(style = AuthorizedStyle.None, level = AuthorizedLevel.Read), enforcing general registry read access
+     * rather than per-group ACL scoping.
+     */
+    private McpToolSearchResults findCompatibleToolsForContent(String targetContentStr, GA targetGa,
+            int safeOffset, int safeLimit) {
+        Set<SearchFilter> filters = new HashSet<>();
+        filters.add(SearchFilter.ofArtifactType(ArtifactType.MCP_TOOL));
+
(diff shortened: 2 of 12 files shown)


Satvik77777:
@paoloantinori — this is the same code you reviewed on #9045, just 
rebuilt on a clean branch (removed an unrelated commit). Your three 
points are captured in the description under "Known limitations." Happy 
to start addressing any of them now if you'd rather I do that before 
further review, or take direction on priority.

Satvik77777:
@paoloantinori — pushed fixes for all three points:
1. O(N) scan & pagination — Added Javadoc on findCompatibleToolsForContent documenting the in-memory scan behavior, pagination limitations, and the MAX_VISIBILITY_FILTER_RESULTS hard cap. Added a log.warn when the search result hits the cap so truncation isn't silent.
2. Silent target parse failure — Malformed target content now throws 400 Bad Request instead of degrading to empty results. Candidate read/parse failures in the loop still skip gracefully (for robustness), but the warn log now includes the candidate's groupId:artifactId coordinates for debuggability. Added a test for the 400 case — WellKnownMcpToolsTest now at 18/18 passing.
3. POST auth style — Added an inline comment above AuthorizedStyle.None explaining that POST accepts raw content (not a stored artifact ref), so GroupAndArtifact/Read doesn't apply, and that searchArtifacts visibility filtering scopes the candidate scan.
All backend tests, UI tests, and typecheck passing. Ready for another pass whenever you have time.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -558,6 +558,197 @@ public McpToolSearchResults searchMcpTools(String name, List<String> parameters,
...
+
+        return findCompatibleToolsForContent(mcpToolContent, null, safeOffset, safeLimit);
+    }
+
+    private McpToolSearchResults findCompatibleToolsForContent(String targetContentStr, GA targetGa,
+            int safeOffset, int safeLimit) {
+        Set<SearchFilter> filters = new HashSet<>();
+        filters.add(SearchFilter.ofArtifactType(ArtifactType.MCP_TOOL));
+
+        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, OrderBy.createdOn,
```
paoloantinori, line 665: Worth confirming that `storage.searchArtifacts(...)` here enforces the caller's per-group read visibility. The only filter is artifact type, so a registry-wide scan runs and each match's group/artifact metadata is returned to the caller. For the GET endpoint this is consistent with `searchMcpTools`, but the POST `findCompatibleMcpTools` is `@Authorized(style = AuthorizedStyle.None)` with no target group/artifact to scope against — so if visibility isn't applied at the storage layer, an authenticated caller could enumerate MCP tools in groups they can't otherwise read. If `searchArtifacts` already scopes by caller, a one-line note here (or on the method Javadoc) would make the assumption explicit.

paoloantinori (commented):
Cleanly implemented feature — the output→input compatibility model (required-only matching with `integer`→`number` widening, an optional output field *not* satisfying a required input) is correct and well tested, and the pagination/Javadoc points raised on the earlier #9045 are addressed on this head. Two things worth a closer look:

- **Visibility / tenant-isolation check on the compatibility scan (strongest point).** Both endpoints resolve through `findCompatibleToolsForContent`, which calls `storage.searchArtifacts(...)` with only an artifact-type filter (`SearchFilter.ofArtifactType(MCP_TOOL)`) and then returns each matching candidate's group/artifact metadata to the caller. This mirrors the existing `searchMcpTools` pattern, so if `searchArtifacts` applies the caller's per-group read visibility there's no new exposure — but it's worth confirming explicitly, because the POST `findCompatibleMcpTools` is `@Authorized(style = AuthorizedStyle.None)` (there's no target group/artifact to scope against), so an authenticated caller could otherwise learn the existence and IDs of MCP tools registered in groups they can't read.
- **Minor UI gap:** `CompatibleMcpToolsViewer` badges the count with `tools.length` (the returned page, default limit 20) rather than `results.count` (total compatible matches), and there's no pagination control — so if more than 20 tools are compatible the list is silently truncated with no indication that more exist. Low priority, but the `count` field is currently ignored entirely.

paoloantinori:
Thanks @Satvik77777 — all three points addressed well:

1. O(N) scan & pagination: Javadoc on `findCompatibleToolsForContent` documenting the in-memory scan, the pagination limits, and the `MAX_VISIBILITY_FILTER_RESULTS` cap — plus a `log.warn` on cap hit so truncation isn't silent. Good.
2. Silent target parse failure: malformed target → 400 Bad Request instead of silent empty results is the right fix; keeping candidate-level failures as graceful skips is correct, and including `groupId:artifactId` in the warn makes it debuggable. Test at 18/18 confirms.
3. POST auth comment on `AuthorizedStyle.None` clarifies why it doesn't apply (raw content vs. stored ref) and how `searchArtifacts` visibility still scopes the candidate scan — fine.

This is solid. Deferring the merge call to a maintainer.

Satvik77777:
@paoloantinori — investigated Point A directly:

storage.searchArtifacts has no per-group visibility filtering built in — it's a pure
type/name/label match against all matching artifacts. The visibility boundary for this
endpoint is the @Authorized(style=None, level=Read) check itself (general registry read
access), which is the same pattern used by the existing /search/artifacts, /search/versions,
and /.well-known/agents endpoints — so this isn't a new gap specific to this feature, it's
consistent with how global search already works here.

I've added a Javadoc note on findCompatibleToolsForContent making this explicit. Let me
know if that model is intentional for this endpoint, or if you'd want per-group filtering
layered on top specifically for MCP tool compatibility search.

Also fixed Point B — CompatibleMcpToolsViewer now shows the real total (results.count) in
the badge and adds a "Showing X of Y compatible tools" indicator when results are truncated.

paoloantinori:
Thanks @Satvik77777 — the investigation of Point A is the right read. `storage.searchArtifacts` is a pure type/name/label match with no per-group ACL layer, so the authorization boundary for the POST endpoint is the `@Authorized(style = None, level = Read)` check, which is consistent with the existing `/search/artifacts`, `/search/versions`, and `/.well-known/agents` endpoints. You've correctly concluded this isn't a *new* gap introduced by this feature; it's the pre-existing scoping model.

The Javadoc you added on `findCompatibleToolsForContent` documenting the O(N) in-memory scan, the `MAX_VISIBILITY_FILTER_RESULTS` cap, the in-memory pagination, and the authorization boundary is exactly the right way to surface that — a future reader or security reviewer can see the design limits without having to reverse-engineer the call chain.

One residual observation, not a blocker: the `MAX_VISIBILITY_FILTER_RESULTS` cap means a registry with more MCP tools than the cap will silently return incomplete compatibility results (only the `log.warn` signals truncation, and only on the candidate set size, not on whether truncation changed the answer). For a first cut that's an acceptable known limitation given the Javadoc documents it; if tool-chaining becomes a high-traffic path, a storage-level filter or a paginated scan would be the follow-up. Just flagging it so it's on the record.

From my side, the feature is in good shape — output→input compatibility model, the `integer`→`number` widening, the optional-output-not-satisfying-required-input edge case, and the visibility-scoping documentation all land cleanly.

jsenko (commented):
## Automated review

### must-fix: Fully qualified `java.util.Objects` — should be an import

**WellKnownResourceImpl.java:659-660**

```java
&& java.util.Objects.equals(artifact.getGroupId(), targetGa.getRawGroupIdWithNull())
&& java.util.Objects.equals(artifact.getArtifactId(), targetGa.getRawArtifactId()))
```

Per the project contributor checklist: "fully qualified names must be imports (not inline)". Add `import java.util.Objects;` and use `Objects.equals(...)`.

### should-fix: O(N) in-memory scan on every compatibility request

**WellKnownResourceImpl.java:627-692**

`findCompatibleToolsForContent` fetches up to `MAX_VISIBILITY_FILTER_RESULTS` (10,000) MCP tools and parses each one on every request. The `offset`/`limit` parameters only paginate the result — they don't reduce the amount of work done.

This is acknowledged in the PR description. Suggestion:
1. Add a Javadoc note documenting the O(N) scan behavior.
2. Consider whether `MAX_VISIBILITY_FILTER_RESULTS` is an appropriate cap for this use case — if users have 10,000+ tools, the endpoint will silently miss matches beyond the cap.

### should-fix: Candidate parse failure silently swallowed

**WellKnownResourceImpl.java:678-681**

If a candidate tool fails to parse (corrupted content, non-JSON, schema mismatch), the `catch (Exception e)` logs a warning and skips it. The caller has no way to know that potentially matching candidates were excluded from results.

Consider adding a count of skipped candidates to the response (e.g., `skippedCount` field) or at least a response header like `X-Skipped-Candidates: 3`. This lets callers decide whether to trust incomplete results.

### nice-to-have: Type compatibility semantics could be documented

**WellKnownResourceImpl.java:752-762**

The current behavior — type checking only applies when both sides declare types, otherwise name-matching alone determines compatibility — is reasonable but not obvious from reading the code. A brief Javadoc on `isOutputCompatibleWithInput` explaining the matching tiers would help future readers:
1. Name must match (required)
2. If both sides declare types, types must match (with integer→number widening)
3. If either side omits type info, name match alone suffices

Satvik77777:
Thanks for the thorough review, @jsenko — addressed all four points:

**Fixed:**
- Replaced the fully-qualified `java.util.Objects` calls with a proper import.
- Added Javadoc on `isOutputCompatibleWithInput` documenting the three-tier 
  matching logic (name required → type match with widening → name-only 
  fallback when type info is absent on either side).

**O(N) scan / pagination:** Already documented in `findCompatibleToolsForContent`'s 
Javadoc from an earlier round — confirmed it covers the scan behavior, the 
`MAX_VISIBILITY_FILTER_RESULTS` cap, and that offset/limit paginate results 
rather than reducing the scan.

**`MAX_VISIBILITY_FILTER_RESULTS`:** Currently 10000, defined locally in this 
file and reused across the other three endpoints here (getEntitledAgents, 
searchAgentsAdvanced, searchAgents) — not a shared/global constant. Consistent 
with the existing pattern in this class rather than something new to this 
endpoint.

**Skipped candidates:** Right now these are handled via `log.warn(...)`, matching 
every other search/discovery endpoint in this codebase (none of them return an 
inline skip count or warnings array — it's server-side logging only). Adding a 
`skippedCount` field would mean a DTO + OpenAPI + UI change, so I held off 
rather than expand the response contract without a second opinion. Happy to 
add it if you'd rather have it inline — let me know which you prefer.

Test suite: 18/18 pass in isolation; the one failure that shows in a full-suite 
run is pre-existing on `main` (verified independently of this diff) and unrelated 
to these changes.

Satvik77777:
@paoloantinori can you plz review

paoloantinori:
Happy to. Status from my side: everything is already reviewed and closed out. The three points from the #9045 review were addressed and confirmed, the Point A authorization question was investigated and settled, and the follow-up commit (bde49f0: the Objects import plus the compatibility Javadoc) cleanly resolves jsenko's feedback with no functional change.

The one thing standing between this and merge is that the branch currently reports conflicts with main, so it will need a rebase before it can land. Once that's pushed I don't expect further comments from me unless the conflict resolution touches the compatibility logic.

jsenko (commented):
Re-review on current head (`bde49f0`). Prior findings addressed.

### Re-review delta

**Resolved:**
- FQN `Objects` import fixed (now at import line 63)
- O(N) in-memory scan documented with detailed Javadoc on `findCompatibleToolsForContent`
- Swallowed parse errors fixed — now throws `BadRequestException` for invalid JSON
- Type compatibility Javadoc added to `isOutputCompatibleWithInput` with three-tier matching docs

**New findings:**

### should-fix: Missing test coverage for type compatibility matching

`isOutputCompatibleWithInput` (WellKnownResourceImpl.java:709) implements three-tier type matching:
1. Name matching (required)
2. Explicit type matching with `integer` → `number` widening
3. Implicit type fallback (when either side omits type)

The tests cover cases (1) — compatible name match, incompatible missing field, and optional output not satisfying required input — but tiers (2) and (3) are untested. Add tests in `WellKnownMcpToolsTest` for:

- **Type mismatch**: target outputs `{ "data": { "type": "string" } }` (required), candidate requires `{ "data": { "type": "integer" } }` → should NOT be compatible.
- **Integer→number widening**: target outputs `{ "count": { "type": "integer" } }` (required), candidate requires `{ "count": { "type": "number" } }` → should be compatible.
- **Implicit type fallback**: target outputs `{ "data": {} }` (no type, required), candidate requires `{ "data": { "type": "string" } }` → should be compatible (name match suffices when one side omits type).

Satvik77777:
@jsenko @paoloantinori

Rebase onto current upstream/main is complete and pushed (--force-with-lease).

Summary of Changes:

Rebase & Conflict Resolution: Resolved merge conflicts in JAX-RS interfaces, service implementations, and PatternFly 6 UI component prop updates.
New MCP Compatibility Test Cases: Added 3 new test cases to WellKnownMcpToolsTest covering tiers 2 and 3 of isOutputCompatibleWithInput:
testTypeMismatchNotCompatible: String output vs integer input required fields → asserted incompatible (count = 0).
testIntegerToNumberWideningCompatible: Integer output vs number input required fields → asserted compatible (count = 1).
testImplicitTypeFallbackCompatible: Target required output omitting type vs typed candidate input → asserted compatible via name-matching (count = 1).

Test Results:

TypeScript: npx tsc --noEmit passed with 0 errors.
Vitest UI: 17/17 test files passed, 131/131 unit tests passed.
Backend Maven: WellKnownMcpToolsTest passed 21/21 tests (18 pre-existing + 3 new).

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -589,209 +584,231 @@ public McpToolSearchResults searchMcpTools(String name, List<String> parameters,
...
-            Optional<JsonNode> compatibleRoot = tryGetCompatibleCandidateRoot(
-                    candidate, rawGroupId, sourceArtifactId, sourceOutputProps);
-            compatibleRoot.ifPresent(root ->
-                    compatibleTools.add(convertToMcpToolSearchResultFromContent(candidate, root)));
+
+        for (SearchedArtifactDto artifact : results.getArtifacts()) {
+            // Exclude the target artifact itself if targetGa is specified
+            if (targetGa != null
+                    && isSameGroup(artifact.getGroupId(), targetGa.getRawGroupIdWithNull())
+                    && java.util.Objects.equals(artifact.getArtifactId(), targetGa.getRawArtifactId())) {
```
paoloantinori: The rebase brought back the fully-qualified java.util.Objects.equals usage at this site and in isSameGroup() (line 800) - this was fixed pre-rebase by adding the import, but the fix did not survive the rebase. Mind restoring the import form?

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResource.java
```diff
@@ -171,34 +170,44 @@ McpToolSearchResults searchMcpTools(
...
-     *
-     * <p>Two tools are considered compatible when every property declared in the source
-     * tool's {@code outputSchema.properties} is also present in the candidate tool's
-     * {@code inputSchema.properties} with the same JSON Schema type. This models the
-     * pipeline chaining contract: the candidate tool can consume what the source tool
-     * produces.</p>
-     *
-     * <p>If the source tool has no {@code outputSchema}, an empty result is returned.
-     * The source tool itself is never included in the results.</p>
+    /**
```
paoloantinori, line 172: Doubled /** opener here - the previous Javadoc block was left behind during the rebase, so this comment starts with a stray /** line. Cosmetic, but javadoc will warn on the malformed block.

paoloantinori (commented):
Rebase introduced two mechanical regressions

The feature itself is unchanged from what was previously reviewed, and the test rewrite with per-test unique artifact/group IDs is a genuine improvement over the old shared static fixtures - it removes the shared-state isolation risk where a global scan could pick up another test's artifacts, and the three new tier tests (type mismatch, integer-to-number widening, optional-not-required) pin down exactly the semantics the compatibility check promises.

Two things came back with the rebase, both trivial:

- `WellKnownResourceImpl` again uses fully-qualified `java.util.Objects.equals` at two sites instead of the plain import. This was flagged and fixed before the rebase; the fix did not survive it.

- `WellKnownResource.getCompatibleMcpTools` has a doubled `/**` Javadoc opener, so the Javadoc tooling will emit a warning for the malformed comment. Cosmetic, but worth a one-line cleanup while touching the file.

Neither affects behavior. With those two cleaned up this is ready from my side.

paoloantinori:
Thanks. I re-read the rebased head: the conflict resolutions hold up, and the test rewrite with per-test unique IDs is a real improvement over the old shared fixtures. I posted a review with two mechanical regressions the rebase introduced (the java.util.Objects import fix disappeared again, plus a doubled Javadoc marker in the interface). Both trivial; apart from those this is ready from my side.

Satvik77777:
@paoloantinori fixed both

paoloantinori:
Re-read the new head — both fixes verified: `java.util.Objects` is now a plain import used at both sites, and the doubled Javadoc opener on `getCompatibleMcpTools` is gone. Nothing further from me on the mechanical points.

The one open thread from the earlier review is the availability trade-off on the unauthenticated POST scan (visibility via `storage.searchArtifacts` with only a type filter). I know it mirrors the existing `searchMcpTools` pattern, so I'm not blocking on it, but it would be good to get an explicit maintainer sign-off on that boundary before merge.

carlesarnal:

PR #9055, opened 2026-07-30
fix(rules): validate MCP tool annotations against the ToolAnnotations schema (#8926)

### LINKED ISSUE #8926: MCP tool validator checks content-annotation fields instead of ToolAnnotations
## Description

**Registry Version**: snapshot (main, 3.3.1-SNAPSHOT)
**Persistence type**: N/A — bug is in shared validation logic (`schema-util`), independent of storage backend

`McpToolContentValidator.validateAnnotationsField` validates an MCP_TOOL artifact's
`annotations` object against `audience` and `priority`, which are MCP *content annotation*
fields. Per the MCP Tools spec (2025-11-25), a tool's `annotations` should instead be a
`ToolAnnotations` object: `title` (string) plus four optional booleans — `readOnlyHint`,
`destructiveHint`, `idempotentHint`, `openWorldHint`. None of the four booleans are validated
today, and the wrong fields (`audience`/`priority`) are enforced instead.

### Environment

Reproducible with any storage backend (sql, kafkasql, gitops, kubernetesops) — the validator
runs in shared `schema-util` code used by all of them. No special environment needed.

### Steps to Reproduce

1. Register an MCP_TOOL artifact with the VALIDITY rule set to FULL.
2. Submit content with `"annotations": {"readOnlyHint": "not-a-boolean", "destructiveHint": 123}`.
3. Registration succeeds : the invalid hint values are never inspected.

Separately:
1. Submit content with only `"annotations": {"priority": 5}`.
2. Registration is rejected with `'annotations.priority' must be between 0 and 1`, even though
   `priority` is not part of `ToolAnnotations` at all.

**Expected**: `readOnlyHint`/`destructiveHint`/`idempotentHint`/`openWorldHint` are validated
as booleans when present; `audience`/`priority` (not part of `ToolAnnotations`) are not enforced.

**Actual**: the reverse : `audience`/`priority` are enforced, the real hint fields are not
checked at all.

### Logs

N/A — this is a validation-logic defect, not a crash; no stack trace involved.

---

**Where**: `schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java`, `validateAnnotationsField` (~L145-187).

**Suggested fix**: keep the `title` string check, replace the `audience`/`priority` blocks with
a boolean check per hint field, and update the test fixtures/assertions accordingly.

Spec ref: https://modelcontextprotocol.io/specification/2025-11-25/server/tools (`ToolAnnotations`)

Happy to send a PR for this, relates to the mentorship epic #8427.

dobbydobap:
## What

`McpToolContentValidator` now validates an MCP_TOOL artifact's `annotations` object against the
MCP `ToolAnnotations` schema: the optional `title` string plus the four optional boolean hints
`readOnlyHint`, `destructiveHint`, `idempotentHint` and `openWorldHint`.

## Why

`validateAnnotationsField` was checking `audience` and `priority`, which are MCP *content
annotation* fields, not `ToolAnnotations` fields. Two concrete consequences at VALIDITY FULL:

- `annotations: {"readOnlyHint": "not-a-boolean", "destructiveHint": 123}` passed validation,
  because the real hint fields were never inspected.
- `annotations: {"priority": 5}` was rejected with `'annotations.priority' must be between 0 and 1`,
  even though `priority` is not part of `ToolAnnotations` at all.

Spec: https://modelcontextprotocol.io/specification/2025-11-25/server/tools

## How

Kept the existing `title` check. Replaced the `audience` and `priority` blocks with a loop over a
`TOOL_ANNOTATION_HINTS` constant that reports a violation when a hint field is present but not a
boolean, using the same message and JSON-pointer style as the surrounding validators.

## Testing

`McpToolContentValidatorTest` covers an invalid-hints document producing exactly three violations
(`title` plus two bad hints, asserted by message), and a new
`testMcpToolAnnotationsIgnoreContentAnnotationFields` case proving a tool carrying `audience` and
`priority` is no longer wrongly rejected. `mcptool-valid.json` now exercises all four hints, so the
happy path is real rather than incidental.

`checkstyle:check` passes on `schema-util/common` and `schema-util/util-provider`.

Closes #8926

Part of the MCP spec validation scope in #8427.

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
@@ -31,6 +31,11 @@ import java.util.Set;
 public class McpToolContentValidator implements ContentValidator {
 
     /**
+     * Declared by both Tool (via BaseMetadata) and ToolAnnotations, so it is validated in both places.
+     */
+    private static final String TITLE_FIELD = "title";
+
+    /**
      * The optional behavioral hints of the MCP ToolAnnotations schema. All of them are booleans.
      */
     private static final List<String> TOOL_ANNOTATION_HINTS = List.of("readOnlyHint",
@@ -94,7 +99,7 @@ public class McpToolContentValidator implements ContentValidator {
     }
 
     private void validateOptionalStringFields(JsonNode tree, Set<RuleViolation> violations) {
-        JsonValidationUtils.validateOptionalString(tree, "title", violations);
+        JsonValidationUtils.validateOptionalString(tree, TITLE_FIELD, violations);
         JsonValidationUtils.validateOptionalString(tree, "description", violations);
     }
 
@@ -163,7 +168,7 @@ public class McpToolContentValidator implements ContentValidator {
         }
 
         // title: optional string (fallback display name per MCP spec)
-        JsonValidationUtils.validateOptionalString(annotations, "title", violations);
+        JsonValidationUtils.validateOptionalString(annotations, TITLE_FIELD, violations);
 
         // readOnlyHint, destructiveHint, idempotentHint, openWorldHint: optional booleans
         for (String hint : TOOL_ANNOTATION_HINTS) {
@@ -178,7 +183,7 @@ public class McpToolContentValidator implements ContentValidator {
         Iterator<String> fieldNames = annotations.fieldNames();
         while (fieldNames.hasNext()) {
             String field = fieldNames.next();
-            if (!"title".equals(field) && !TOOL_ANNOTATION_HINTS.contains(field)) {
+            if (!TITLE_FIELD.equals(field) && !TOOL_ANNOTATION_HINTS.contains(field)) {
                 violations.add(new RuleViolation("'annotations." + field
                         + "' is not a ToolAnnotations property", "/annotations/" + field));
             }
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
@@ -10,6 +10,7 @@ import io.apicurio.registry.types.RuleType;
 
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -22,7 +23,8 @@ import java.util.Set;
  * Validation levels:
  * - NONE: No validation
  * - SYNTAX_ONLY: Validates that the content is valid JSON and is an object
- * - FULL: Full schema validation including required fields, type checking, and structure validation
+ * - FULL: Full schema validation including required fields, type checking, and structure
+ *   validation. The 'annotations' object is validated as a closed ToolAnnotations shape.
  *
  * @see <a href="https://modelcontextprotocol.io/specification/2025-11-25/server/tools">MCP Tools</a>
  */
@@ -168,6 +170,17 @@ public class McpToolContentValidator implements ContentValidator {
             if (annotations.has(hint) && !annotations.get(hint).isBoolean()) {
                 violations.add(new RuleViolation("'annotations." + hint + "' must be a boolean",
                         "/annotations/" + hint));
+            }
+        }
+
+        // ToolAnnotations is a closed shape at FULL, so a typo or a stray field is a violation
+        // rather than a silently ignored no-op.
+        Iterator<String> fieldNames = annotations.fieldNames();
+        while (fieldNames.hasNext()) {
+            String field = fieldNames.next();
+            if (!"title".equals(field) && !TOOL_ANNOTATION_HINTS.contains(field)) {
+                violations.add(new RuleViolation("'annotations." + field
+                        + "' is not a ToolAnnotations property", "/annotations/" + field));
             }
         }
     }
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
@@ -28,6 +28,12 @@ import java.util.Set;
  */
 public class McpToolContentValidator implements ContentValidator {
 
+    /**
+     * The optional behavioral hints of the MCP ToolAnnotations schema. All of them are booleans.
+     */
+    private static final List<String> TOOL_ANNOTATION_HINTS = List.of("readOnlyHint",
+            "destructiveHint", "idempotentHint", "openWorldHint");
+
     @Override
     public void validate(ValidityLevel level, TypedContent content,
             Map<String, TypedContent> resolvedReferences) throws RuleViolationException {
@@ -157,31 +163,11 @@ public class McpToolContentValidator implements ContentValidator {
         // title: optional string (fallback display name per MCP spec)
         JsonValidationUtils.validateOptionalString(annotations, "title", violations);
 
-        // audience: optional array of strings ("user", "assistant")
-        if (annotations.has("audience")) {
-            JsonNode audience = annotations.get("audience");
-            if (!audience.isArray()) {
-                violations.add(new RuleViolation("'annotations.audience' must be an array",
-                        "/annotations/audience"));
-            } else {
-                JsonValidationUtils.validateStringArray(audience,
-                        "/annotations/audience", "audience role", violations);
-            }
-        }
-
-        // priority: optional number between 0 and 1
-        if (annotations.has("priority")) {
-            JsonNode priority = annotations.get("priority");
-            if (!priority.isNumber()) {
-                violations.add(new RuleViolation("'annotations.priority' must be a number",
-                        "/annotations/priority"));
-            } else {
-                double value = priority.asDouble();
-                if (value < 0 || value > 1) {
-                    violations.add(new RuleViolation(
-                            "'annotations.priority' must be between 0 and 1",
-                            "/annotations/priority"));
-                }
+        // readOnlyHint, destructiveHint, idempotentHint, openWorldHint: optional booleans
+        for (String hint : TOOL_ANNOTATION_HINTS) {
+            if (annotations.has(hint) && !annotations.get(hint).isBoolean()) {
+                violations.add(new RuleViolation("'annotations." + hint + "' must be a boolean",
+                        "/annotations/" + hint));
             }
         }
     }
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
@@ -84,6 +84,8 @@ public class McpToolContentValidatorTest extends ArtifactUtilProviderTestBase {
         // Violations for title (not a string), readOnlyHint and destructiveHint (not booleans)
         Assertions.assertEquals(3, error.getCauses().size());
         Assertions.assertTrue(error.getCauses().stream()
+                .anyMatch(v -> "'title' field must be a string".equals(v.getDescription())));
+        Assertions.assertTrue(error.getCauses().stream()
                 .anyMatch(v -> "'annotations.readOnlyHint' must be a boolean"
                         .equals(v.getDescription())));
         Assertions.assertTrue(error.getCauses().stream()
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
@@ -81,10 +81,24 @@ public class McpToolContentValidatorTest extends ArtifactUtilProviderTestBase {
         RuleViolationException error = Assertions.assertThrows(RuleViolationException.class, () -> {
             validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
         });
-        Assertions.assertFalse(error.getCauses().isEmpty());
-        // Should have violations for title (not string), audience (not array),
-        // and priority (not number)
-        Assertions.assertTrue(error.getCauses().size() >= 3);
+        // Violations for title (not a string), readOnlyHint and destructiveHint (not booleans)
+        Assertions.assertEquals(3, error.getCauses().size());
+        Assertions.assertTrue(error.getCauses().stream()
+                .anyMatch(v -> "'annotations.readOnlyHint' must be a boolean"
+                        .equals(v.getDescription())));
+        Assertions.assertTrue(error.getCauses().stream()
+                .anyMatch(v -> "'annotations.destructiveHint' must be a boolean"
+                        .equals(v.getDescription())));
+    }
+
+    @Test
+    public void testMcpToolAnnotationsIgnoreContentAnnotationFields() throws Exception {
+        // audience and priority belong to the MCP content annotation schema, not to
+        // ToolAnnotations, so they must not be validated on a tool definition.
+        TypedContent content = resourceToTypedContentHandle(
+                "mcptool-annotations-content-fields.json");
+        McpToolContentValidator validator = new McpToolContentValidator();
+        validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
     }
 
     @Test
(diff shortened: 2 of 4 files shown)


dobbydobap:
@jsenko could you take a look when you get a chance?

This has been at `lifecycle/new` since Jul 30. From reading the orchestrator, `/accept` is what
moves it to `lifecycle/wip`, and until then `decide` skips every test job and `Verification Gate`
fails on the lifecycle check rather than on anything in the diff. So the red X is the lifecycle
state, not the code.

One thing that would help: `/ready` needs a reviewer assigned first, otherwise the bot replies
that it cannot transition. Assigning a reviewer alongside `/accept` would unblock the whole chain.

The change is small, 5 files and +46/-33, and `checkstyle:check` passes on both modules it
touches.

dobbydobap:
One related thing I noticed while working in this area:
`schema-util/common/src/main/resources/io/apicurio/registry/rules/validity/mcp-tool-schema.json`
still describes `annotations` as "(title, audience, priority)" and declares an `audience`
property, which is the same pre-#8926 model this PR corrects in the validator.

As far as I can tell nothing loads that file, its filename does not appear anywhere in the
codebase, so it is not affecting validation today. Flagging it in case it should be corrected or
removed separately, rather than left as a second stale definition of the same thing.

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
```diff
@@ -157,31 +163,11 @@ private void validateAnnotationsField(JsonNode tree, Set<RuleViolation> violatio
         // title: optional string (fallback display name per MCP spec)
         JsonValidationUtils.validateOptionalString(annotations, "title", violations);
```
paoloantinori: Worth a quick confirm against the cited 2025-11-25 spec: is `title` a member of `ToolAnnotations` here, or a top-level `Tool` property? In some MCP versions `title` lives on `Tool` rather than under `annotations`. This PR inherits the existing `annotations.title` handling (so it is not a regression introduced here), but the updated `mcptool-valid.json` happy path now exercises `annotations.title`, so the spec placement matters for the validation's correctness claim. If `title` is actually a `Tool`-level field, validating it here gives a false sense of coverage for the tool's real title.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
```diff
@@ -81,10 +81,24 @@ public void testMcpToolInvalidAnnotations() throws Exception {
         RuleViolationException error = Assertions.assertThrows(RuleViolationException.class, () -> {
             validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
         });
-        Assertions.assertFalse(error.getCauses().isEmpty());
-        // Should have violations for title (not string), audience (not array),
-        // and priority (not number)
-        Assertions.assertTrue(error.getCauses().size() >= 3);
+        // Violations for title (not a string), readOnlyHint and destructiveHint (not booleans)
+        Assertions.assertEquals(3, error.getCauses().size());
```
paoloantinori, line 85: The two hint violations are asserted by message, but the `title` violation is only implied by the exact count of 3. Adding a third `anyMatch(... .equals(v.getDescription()))` for the title-violation message would make the expectation explicit and resilient if `JsonValidationUtils.validateOptionalString`'s wording ever changes (the current `assertEquals(3, ...)` would stay green even if that message silently shifted).

paoloantinori (commented):
Correct, well-scoped fix that aligns MCP_TOOL validation with the `ToolAnnotations` schema: it drops the misplaced `audience`/`priority` checks (MCP content-annotation fields, not ToolAnnotations members) and validates the four optional boolean hints, and the test tightens from a loose lower bound to an exact count. No blockers from my side; two easy-win observations below.

- `McpToolContentValidator.java:168` - the hint predicate `has(hint) && !get(hint).isBoolean()` is correct and rejects strings, numbers, arrays, objects, and explicit nulls while accepting real booleans. `has(hint)` also guards `get(hint)`, so there is no NPE path here.
- `McpToolContentValidator.java:164` - `title` is still validated inside `annotations`; worth confirming it belongs to `ToolAnnotations` (and is not a top-level `Tool` property) under the cited 2025-11-25 spec. See inline.
- `McpToolContentValidatorTest.java:85` - the exact count of 3 plus two hint-message checks never asserts the `title` violation message, so that expectation stays implicit. See inline.

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
dobbydobap: Good question. I checked the schema rather than the prose page, and both exist:
`annotations.title` is a documented fallback, not a misplacement.

From `schema/2025-11-25/schema.ts` in the spec repo:

```ts
export interface BaseMetadata { name: string; title?: string; }

export interface ToolAnnotations {
  title?: string;            // "A human-readable title for the tool."
  readOnlyHint?: boolean;
  destructiveHint?: boolean;
  idempotentHint?: boolean;
  openWorldHint?: boolean;
}

export interface Tool extends BaseMetadata, Icons { ... annotations?: ToolAnnotations; ... }
```

`Tool` gets `title` from `BaseMetadata`, and `ToolAnnotations` has its own. The `Tool` doc comment
states the resolution rule: "Display name precedence order is: title, annotations.title, then name."

The validator covers both. `validateOptionalStringFields` checks the top-level `Tool.title` and
`description`, and `validateAnnotationsField` checks `ToolAnnotations.title`. So the happy path in
`mcptool-valid.json` exercises a real field.

Worth saying the ambiguity is fair: the Tools page lists `title` under Tool but describes
`annotations` only as "Optional properties describing tool behavior" without enumerating members.
`ToolAnnotations` is only spelled out in the schema reference.

One pre-existing thing I noticed on this line while checking, not touched here:
`JsonValidationUtils.validateOptionalString` builds its pointer as `"/" + fieldName`, so the
`annotations.title` violation gets context `/title` rather than `/annotations/title`, while the four
hint violations use `/annotations/<hint>`. Fixing it means changing a shared utility that
`AGENT_CARD` also uses, so I kept it out of this PR. Happy to file it separately if useful.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
dobbydobap: Agreed, that expectation was only implicit. Added in the latest commit:

```java
Assertions.assertTrue(error.getCauses().stream()
        .anyMatch(v -> "'title' field must be a string".equals(v.getDescription())));
```

All three violations are now asserted by message, so the test will not stay green if
`validateOptionalString`'s wording changes. 14 tests pass in the class and `checkstyle:check` is
clean on `schema-util/common` and `schema-util/util-provider`.

paoloantinori:
Perfect, both addressed. The per-message assertion is exactly the hardening I had in mind, and your spec breakdown on the dual `title` (BaseMetadata + ToolAnnotations) resolves the ambiguity cleanly.

On the `/title` vs `/annotations/title` pointer thing you spotted: yes, that's a shared-utility concern and correctly out of scope here. A separate issue for it would be welcome since the pointer drift affects AGENT_CARD too.

Looks good from my side.

dobbydobap:
@paoloantinori  Thanks. Filed as #9372.

One note before you read it: I checked the AGENT_CARD call sites while writing it up, and the drift
does not actually reproduce there today. All three of its `validateOptionalString` calls pass the
root tree, so those pointers are correct. The only live occurrence is the `annotations.title` one.
AGENT_CARD is still exposed to the same helper, and #9347 is adding another root-only method to it
right now, so fixing it at the utility level still looks worth doing. Full call-site table is in
the issue.

dobbydobap:
Separately, is `/accept` something you or @jsenko would normally run here? The PR is still on
`lifecycle/new`, so `decide` skips every test job and the Verification Gate fails on the lifecycle
check rather than on anything in the diff. No rush, just flagging that CI has not actually
exercised this change yet.

paoloantinori:
You're right that the Verification Gate is failing on the lifecycle check, not on anything in the diff — `decide` skips the test jobs while the PR sits on `lifecycle/new`, so the actual validation change hasn't been exercised by CI yet.

On `/accept`: that's a maintainer-side lifecycle call, so I'll defer to @jsenko on when to move it forward. I don't want to push the label myself without his read on the queue. The review side from me is closed — the fix and the per-message assertions landed cleanly.

paoloantinori:
Thanks for the heads-up on #9372 and for checking the AGENT_CARD call sites ahead of time. That's useful context: if the only live occurrence of the pointer drift is `annotations.title` in `McpToolContentValidator`, fixing it at the utility level (rather than per-call-site) is the right call, especially with #9347 adding another root-only method.

I'll carry that context into reading #9372.

jsenko (commented):
Fixes #8926 — corrects MCP tool validator to check `ToolAnnotations` fields instead of content-annotation fields.

### Summary

Correct and well-scoped fix. The validator was checking `audience` (array) and `priority` (number 0–1), which belong to MCP *content annotations*, not `ToolAnnotations`. The PR replaces those with validation for the four spec-defined boolean hints (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`) via a clean loop over a `TOOL_ANNOTATION_HINTS` constant. The existing `title` string check is retained correctly.

- **JSON pointer path for `annotations.title`**: `validateOptionalString(annotations, "title", ...)` produces a pointer `/title` rather than `/annotations/title`. This is pre-existing and the author has already filed #9372 to address it at the utility level — correctly out of scope here.
- **`has(hint)` guard**: `annotations.has(hint) && !annotations.get(hint).isBoolean()` is safe — `has` guards the `get`, no NPE path. Correctly rejects strings, numbers, arrays, objects, and explicit nulls while accepting real booleans.
- **Stale schema file**: The author flagged that `mcp-tool-schema.json` still references `audience`/`priority`. Confirmed it is not loaded anywhere in the codebase — a separate cleanup tracked in the comments.

### Test coverage

Tests are strong:
- Exact violation count (3) with per-message assertions for `title`, `readOnlyHint`, `destructiveHint`
- New `testMcpToolAnnotationsIgnoreContentAnnotationFields` proves `audience`/`priority` no longer rejected
- `mcptool-valid.json` exercises all four hints on the happy path

No findings. All prior review feedback (paoloantinori) was addressed.

**Note:** PR is still in `lifecycle/new` — needs `/accept` from a maintainer before CI will exercise the tests.

dobbydobap:
Thanks @jsenko, that is a thorough read, and I appreciate you confirming the #9372 scope call and
verifying the stale schema file independently.

On your closing note: is `/accept` something you can run here? It looks like the only thing left
before CI exercises the tests, and I am happy to deal with anything the suite turns up.

dobbydobap:
@jsenko this has been on `lifecycle/new` since Jul 30, so the test suite still has not run against
it. You and @paoloantinori have both reviewed with no blocking findings.

Flagging it now only because #9297 and #9372 are also assigned to me with fixes written and rebased
on current main, and the one-PR rule keeps both behind this one. If the queue is just long that is
fine, I would rather wait than push.

carlesarnal:
Bear in mind that all work related to https://github.com/Apicurio/apicurio-registry/issues/8427 will be blocked until the term. I'm accepting this since it's a bug, but have that in mind.

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
```diff
@@ -157,31 +163,11 @@ private void validateAnnotationsField(JsonNode tree, Set<RuleViolation> violatio
...
-                        "/annotations/priority"));
-            } else {
-                double value = priority.asDouble();
-                if (value < 0 || value > 1) {
-                    violations.add(new RuleViolation(
-                            "'annotations.priority' must be between 0 and 1",
-                            "/annotations/priority"));
-                }
+        // readOnlyHint, destructiveHint, idempotentHint, openWorldHint: optional booleans
+        for (String hint : TOOL_ANNOTATION_HINTS) {
```
carlesarnal, line 174: In FULL mode we should validate the ToolAnnotations object shape, not only the types of known keys. As implemented, unknown keys (including non-tool annotation keys) are silently accepted, so typoed or unsupported annotation properties pass validation. Please reject unsupported annotations keys in FULL mode (or explicitly document/flag a forward-compatibility mode if that behavior is intentional).

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
```diff
@@ -81,10 +81,26 @@ public void testMcpToolInvalidAnnotations() throws Exception {
...
+        Assertions.assertTrue(error.getCauses().stream()
+                .anyMatch(v -> "'annotations.readOnlyHint' must be a boolean"
+                        .equals(v.getDescription())));
+        Assertions.assertTrue(error.getCauses().stream()
+                .anyMatch(v -> "'annotations.destructiveHint' must be a boolean"
+                        .equals(v.getDescription())));
+    }
+
+    @Test
+    public void testMcpToolAnnotationsIgnoreContentAnnotationFields() throws Exception {
```
carlesarnal: This test currently codifies accepting non-ToolAnnotations fields under annotations in FULL mode. If FULL is intended to represent schema validation, this expectation should be inverted to assert a violation for unsupported annotation keys.

carlesarnal (changes requested):
Thanks for the MCP alignment work. I’m requesting changes because FULL validity currently accepts unsupported annotations keys instead of enforcing the ToolAnnotations shape.

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/McpToolContentValidator.java
dobbydobap: Done. `validateAnnotationsField` now treats `annotations` as a closed shape at FULL: anything that
is not `title` or one of the four behavioral hints is reported as
`'annotations.<key>' is not a ToolAnnotations property`. The practical win is the typo case, since
`{"readOnlyHnt": true}` previously validated clean and silently did nothing.

One thing worth recording rather than arguing, since you offered the forward-compatibility option.
The authoritative schema leaves `ToolAnnotations` open: in `schema/2025-11-25/schema.json` it has no
`additionalProperties`, and the documented extension point, `_meta`, sits on `Tool` rather than
inside `annotations`. So this is deliberately stricter than the wire format, which seems right for a
governance rule at its strictest level, and FULL is opt-in anyway. The cost is that a hint added in a
future spec revision needs a one-line update to `TOOL_ANNOTATION_HINTS` here. Happy to relax it if
you would rather trade the typo detection for that.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
dobbydobap: Inverted. `testMcpToolAnnotationsIgnoreContentAnnotationFields` is now
`testMcpToolAnnotationsRejectUnsupportedFields` and asserts exactly two violations, with the
messages for `audience` and `priority` checked individually.

You were right that the old name and assertion were codifying permissiveness as a contract, which is
more than #8926 actually claimed. The bug was that content-annotation rules were being applied to
ToolAnnotations, not that unknown keys had to be accepted.

Verified: the test fails without the validator change ("Expected RuleViolationException to be
thrown, but nothing was thrown") and passes with it. 14 tests in the class, checkstyle clean on both
modules. Also rebased onto current main, which picks up the `app-other` shard split from #9287 — the
earlier red run was the pre-split shard hitting the OOM from #9265, not this diff.

carlesarnal:
@dobbydobap what model are you using? Are you automating replies?

dobbydobap:
No, I don't automate replies. I write and post everything myself.

I do use Claude Code while working, mostly to check things I'd otherwise get wrong. That's how I
ended up quoting schema.ts for the dual `title`, and how I caught that the pointer drift doesn't
actually reproduce in AGENT_CARD before I claimed it did in #9372.

carlesarnal:
> No, I don't automate replies. I write and post everything myself.
> 
> I do use Claude Code while working, mostly to check things I'd otherwise get wrong. That's how I ended up quoting schema.ts for the dual `title`, and how I caught that the pointer drift doesn't actually reproduce in AGENT_CARD before I claimed it did in #9372.

Good, I was asking because they seemed very well formatted, and it's hard to tell this days. Note that I was already happy with what's here :)

carlesarnal (approved):

PR #9066, opened 2026-07-30
fixes missing contract-related events when using KafkaSQL storage

vivek-gite:
## Root cause

Contract metadata and status events were temporarily emitted from `GroupsResourceImpl` through `RegistryStorage.createEvent()`. That bypassed the normal storage-layer event flow. It also did not address six KafkaSQL contract-ruleset operations, which journaled their changes but never published corresponding events.

## What does it fix?
issue #9034 

## Changes

- Adds `RegistryStorage` operations for contract metadata updates and status transitions.
- Moves contract label updates and event creation from the REST resource into SQL storage.
- Adds replayable KafkaSQL journal messages for the new storage operations.
- Publishes `KafkaSqlOutboxEvent`s only after KafkaSQL journal acknowledgement for:
  - contract metadata updates;
  - contract status transitions;
  - artifact-level ruleset set/delete;
  - version-level ruleset set/delete;
  - global ruleset set/delete.
- Restores KafkaSQL `createEvent()` to its intended no-op behavior, removing the REST-layer workaround.
- Updates read-only storage implementations for the new write methods.
- Adds regression coverage for read-only storage and KafkaSQL contract event payloads.

## Testing

- `./mvnw test-compile -pl app -am -DskipTests`
- `./mvnw checkstyle:check -pl app`
- `./mvnw test -pl app -am -Dtest=ReadOnlyRegistryStorageTest -Dsurefire.failIfNoSpecifiedTests=false`


paoloantinori (commented):
Clean, well-scoped change that moves contract metadata/status event emission out of the REST layer and into the storage implementations, extending the established KafkaSQL journal-then-publish pattern to the two new operations and the six ruleset ops. No MAJOR or BLOCKER issues found; the logic is a faithful port and the dual-outbox wiring (`KafkaSqlOutboxEvent` vs `SqlOutboxEvent`) is applied consistently.

- The label-merge and date/status logic from `GroupsResourceImpl` (`updateContractMetadata` ~line 2007, `transitionContractStatus` ~line 2130) is moved verbatim into `AbstractSqlRegistryStorage.updateContractMetadata`/`transitionContractStatus` (~lines 842-875), preserving the `contractId != null` prefix selection and the null-safe `fromStatus` for the `ContractStatusChanged` event.
- Event publication is correctly ordered after `coordinator.waitForResponse(uuid)` in `KafkaSqlRegistryStorage` (~lines 688-706, 717-757), so `KafkaSqlOutboxEvent`s only fire after the KafkaSQL journal is acknowledged — the core fix.
- `createEvent` in `KafkaSqlRegistryStorage` (~line 1333) is correctly restored to a no-op now that its only callers in `GroupsResourceImpl` are gone; the SQL `createEvent` is left intact, so the non-Kafka path still emits events via the new storage methods.
- Minor (NIT): the literal `"__GLOBAL__"` is repeated for the global ruleset events in `KafkaSqlRegistryStorage` (~lines 776, 786); if this is a stable sentinel it would read better as a shared constant, though it is internally consistent and matches the test.
- Minor (test robustness): `consumeContractEvents` (`KafkaSqlContractEventsTest`) caps collection at the expected count, so unlike the ruleset test (which asserts exactly one match per key) it would not detect a duplicate metadata/status event if one were ever introduced.

vivek-gite:
> Clean, well-scoped change that moves contract metadata/status event emission out of the REST layer and into the storage implementations, extending the established KafkaSQL journal-then-publish pattern to the two new operations and the six ruleset ops. No MAJOR or BLOCKER issues found; the logic is a faithful port and the dual-outbox wiring (`KafkaSqlOutboxEvent` vs `SqlOutboxEvent`) is applied consistently.
> 
>     * The label-merge and date/status logic from `GroupsResourceImpl` (`updateContractMetadata` ~line 2007, `transitionContractStatus` ~line 2130) is moved verbatim into `AbstractSqlRegistryStorage.updateContractMetadata`/`transitionContractStatus` (~lines 842-875), preserving the `contractId != null` prefix selection and the null-safe `fromStatus` for the `ContractStatusChanged` event.
> 
>     * Event publication is correctly ordered after `coordinator.waitForResponse(uuid)` in `KafkaSqlRegistryStorage` (~lines 688-706, 717-757), so `KafkaSqlOutboxEvent`s only fire after the KafkaSQL journal is acknowledged — the core fix.
> 
>     * `createEvent` in `KafkaSqlRegistryStorage` (~line 1333) is correctly restored to a no-op now that its only callers in `GroupsResourceImpl` are gone; the SQL `createEvent` is left intact, so the non-Kafka path still emits events via the new storage methods.
> 
>     * Minor (NIT): the literal `"__GLOBAL__"` is repeated for the global ruleset events in `KafkaSqlRegistryStorage` (~lines 776, 786); if this is a stable sentinel it would read better as a shared constant, though it is internally consistent and matches the test.
> 
>     * Minor (test robustness): `consumeContractEvents` (`KafkaSqlContractEventsTest`) caps collection at the expected count, so unlike the ruleset test (which asserts exactly one match per key) it would not detect a duplicate metadata/status event if one were ever introduced.

@paoloantinori  Thanks for the thorough review! I addressed both suggestions in the latest update:

- Introduced a shared `GLOBAL_CONTRACT_RULESET_COORDINATE` constant and reused it in the KafkaSQL global ruleset event handling and test expectations.
- Updated `consumeContractEvents` to continue polling briefly after the expected events arrive, so late duplicate metadata/status events now cause the exact-count assertion to fail.

I also verified the changes with test compilation and Checkstyle.

paoloantinori:
Thanks @vivek-gite — both suggestions are addressed cleanly on this head:

- The `GLOBAL_CONTRACT_RULESET_COORDINATE` constant is now reused on both global-ruleset event paths and in the test expectations, so the sentinel has a single source.
- `consumeContractEvents` continues briefly past the expected count, which means a stray duplicate metadata/status event would now flip the assertion instead of being silently absorbed.

vivek-gite:
Hi @paoloantinori I have couple of more PRs in pipeline, is this good to be merged?

paoloantinori:
Hi @vivek-gite — from a code-review standpoint, yes, this is good to go: both points from my earlier passes (the `GLOBAL_CONTRACT_RULESET_COORDINATE` constant reuse and the journal-then-publish ordering) are addressed cleanly on this head, and the regression coverage in `KafkaSqlContractEventsTest` is solid.

Two non-code blockers before it can actually merge, neither of which I can action from here:

1. **Behind main** — the PR currently reports merge conflicts / behind, so it'll need a rebase onto latest `main` before the merge bot will proceed.
2. **Lifecycle** — it's sitting on `lifecycle/new`, so it still needs a maintainer `/accept` to progress (same gate as the other new-contributor PRs).

Once those two clear, the change itself is ready. Happy to take a look at your other PRs in the pipeline as they come up.

vivek-gite:
> Thanks @vivek-gite — both suggestions are addressed cleanly on this head:
> 
>     * The `GLOBAL_CONTRACT_RULESET_COORDINATE` constant is now reused on both global-ruleset event paths and in the test expectations, so the sentinel has a single source.
> 
>     * `consumeContractEvents` continues briefly past the expected count, which means a stray duplicate metadata/status event would now flip the assertion instead of being silently absorbed.
> 
> 
> From my side this is good to merge.

@paoloantinori thanks for clarification. 
Action-point:1 - Now the branch is same as the main 

Quick question, is there any discussion forum or slack group that i can join?

paoloantinori:
Thanks @vivek-gite — nothing has changed since that last read: my "good to go" still stands from the code-review side. The merge decision itself is a maintainer call (the branch is currently `behind` main, so a rebase will be needed before merge), but there are no outstanding review items from me.

carlesarnal:
I've gone through the full diff and verified the changes. Everything looks solid:

- **Journal-then-publish ordering** is correct throughout `KafkaSqlRegistryStorage` — events fire only after `coordinator.waitForResponse(uuid)` returns, which is the core fix for #9034.

- **Storage layer coverage** is complete: the new `updateContractMetadata` and `transitionContractStatus` methods are implemented across `RegistryStorage` (interface), `AbstractSqlRegistryStorage`, `KafkaSqlRegistryStorage`, `ReadOnlyRegistryStorageDecorator`, `AbstractReadOnlyRegistryStorage`, and `TestInMemoryRegistryStorage`.

- **Dual outbox wiring** is consistent — `SqlOutboxEvent` in the SQL path (`AbstractSqlRegistryStorage` ~lines 849, 870) and `KafkaSqlOutboxEvent` in the KafkaSQL path (`KafkaSqlRegistryStorage` ~lines 694, 706).

- **`createEvent` no-op restoration** (`KafkaSqlRegistryStorage` ~line 1337) correctly removes the REST-layer workaround now that events are published from within the storage methods.

- **`GLOBAL_CONTRACT_RULESET_COORDINATE` constant** is used consistently in the global ruleset event paths and matching test expectations — clean single source of truth.

- **Test coverage** in `KafkaSqlContractEventsTest` is thorough: JSON-level field assertions on metadata/status events, all six ruleset operations (artifact/version/global × set/delete) verified by composite key matching, and the `DUPLICATE_DETECTION_WINDOW` ensures stray duplicates would fail the exact-count assertion.

- New message classes (`TransitionContractStatus6Message`, `UpdateContractMetadata4Message`) have proper Apache 2.0 headers and are registered in `KafkaSqlMessageIndex`.

No issues found from my side. The CI Verification Gate failure is lifecycle-related (`lifecycle/new`), not code-related.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1294,7 +1334,7 @@ public String createSnapshot(String snapshotLocation) throws RegistryStorageExce
 
     @Override
     public String createEvent(OutboxEvent event) {
```
paoloantinori, line 1336: `createEvent(OutboxEvent)` now publishes nothing in KafkaSQL mode. The contract operations compensate by firing `KafkaSqlOutboxEvent` inline after journal ack, but if any other code path still calls `RegistryStorage.createEvent(...)` expecting it to publish (the REST contract resource was the known caller), those events will now be silently dropped under KafkaSQL storage. Worth confirming the blast radius — e.g. `grep -rn "createEvent(" --include=*.java app/src/main` — to make sure no remaining caller relies on the old publish behavior and that the inline firing covers every previously-published event type.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -691,6 +717,8 @@ public void setArtifactContractRuleset(String groupId, String artifactId,
         var message = new SetArtifactContractRuleset3Message(groupId, artifactId, ruleset);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent.of(
```
paoloantinori: `CONTRACT_RULESET_CONFIGURED` is fired here in the KafkaSQL impl, but the analogous write methods in `AbstractSqlRegistryStorage` (`setArtifactContractRuleset` / `setVersionContractRuleset` / `setGlobalContractRuleset`) were not updated to fire it — unlike `updateContractMetadata` and `transitionContractStatus`, which this PR explicitly gave SQL-side `outboxEvent.fire(...)` calls. If pure-SQL storage is expected to publish ruleset events, consumers on the SQL backend will silently miss set/delete of artifact, version, and global contract rulesets. Confirm whether SQL-mode ruleset events are intentionally out of scope, or add the matching `SqlOutboxEvent` firing for parity.

paoloantinori (commented):
The PR cleanly moves contract metadata/status event emission into the storage layer and adds the previously-missing KafkaSQL outbox events for all six contract-ruleset operations, firing them after journal acknowledgement — the design is internally consistent and the new `KafkaSqlContractEventsTest` gives good payload coverage. Two blast-radius points are worth confirming before merge:

- `KafkaSqlRegistryStorage.createEvent` is now a no-op (`KafkaSqlRegistryStorage.java:1336`); the contract methods compensate by firing `KafkaSqlOutboxEvent` inline, but any *other* caller of `RegistryStorage.createEvent` would silently drop its event under KafkaSQL storage.
- `CONTRACT_RULESET_CONFIGURED` is fired only in the KafkaSQL impl (e.g. `KafkaSqlRegistryStorage.java:720`); the SQL impl's ruleset methods were not given equivalent firing (only the metadata/status methods were, in `AbstractSqlRegistryStorage.java`), so pure-SQL storage may not emit ruleset events.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
EricWittmann: This is correct - createEvent() is a no-op for kafkasql.  👍

EricWittmann (commented):
Thanks for the work @vivek-gite — the overall architecture (moving contract events into the storage layer, adding KafkaSQL outbox events for all 6 ruleset operations) is exactly right. However, there's a transactional boundary issue in the SQL implementation that breaks the outbox pattern. This needs to be fixed before we can merge.

## How Transaction Boundaries Work in Apicurio Registry

We don't use `@Transactional` in Apicurio Registry. Instead, transactional boundaries are managed by `AbstractHandleFactory.withHandle()`:

- The **outermost** `withHandle()` call creates a JDBC connection with `autocommit=false` and manages commit/rollback
- **Nested** `withHandle()` calls detect the existing connection (via a ThreadLocal `LocalState`) and reuse it — incrementing a nesting counter instead of opening a new connection
- Only the outermost call commits or rolls back when it completes

This means: **for two operations to be atomic, they must both execute inside the same outermost `withHandle()` callback.**

## The Outbox Pattern Requirement

For the outbox pattern to work correctly, the **functional change** (e.g., updating contract metadata labels) and the **outbox row insert** (triggered by `outboxEvent.fire()`) must happen in the **same database transaction**. If they're in separate transactions, a crash between the two means the data changed but the event was never published — silent data loss from the consumer's perspective.

## The Problem

The new `updateContractMetadata` method in `AbstractSqlRegistryStorage` does this:

```java
public void updateContractMetadata(String groupId, String artifactId, String contractId,
        EditableContractMetadataDto metadata) {
    String prefix = contractId != null
            ? ContractLabels.contractPrefix(contractId) : ContractLabels.PREFIX;
    // Transaction #1: mergeArtifactLabels() → repository → handles.withHandle() → commit
    mergeArtifactLabels(groupId, artifactId, prefix,
            contractMetadataMapper.toLabels(metadata, prefix));
    // Transaction #2: outboxEvent.fire() → SqlEventsProcessor.processEvent(@Observes)
    //   → SqlEventRepository.createEvent() → handles.withHandle() → commit
    outboxEvent.fire(SqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
}
```

`mergeArtifactLabels()` calls through to a repository method that wraps its work in `handles.withHandle()` — that's **Transaction #1**, which commits and closes. Then `outboxEvent.fire()` triggers `SqlEventsProcessor.processEvent()` synchronously (`@Observes`, not `@ObservesAsync`), which calls `SqlEventRepository.createEvent()` — that opens its **own** `handles.withHandle()` — **Transaction #2**.

These are two independent transactions. A crash between them means the metadata was updated but the event was never published.

The same issue applies to `transitionContractStatus`, which has multiple `mergeArtifactLabels` calls (each in their own transaction), followed by the outbox event in yet another separate transaction.

## The Correct Pattern

Methods like `createArtifact` do it right — they call `outboxEvent.fire()` **inside** a `handles.withHandle()` callback:

```java
return handles.withHandle(handle -> {
    // functional work uses the handle
    handle.createUpdate(sqlStatements.insertArtifact())...execute();
    // ...

    // outbox event fires INSIDE the withHandle callback
    // SqlEventsProcessor → createEvent() → nested withHandle() → REUSES same connection
    outboxEvent.fire(SqlOutboxEvent.of(ArtifactCreated.of(amdDto)));
    return result;
});
```

Because `outboxEvent.fire()` executes inside the outer `withHandle()`, the nested `withHandle()` in `SqlEventRepository.createEvent()` detects the existing connection and reuses it — same transaction. The outermost call commits everything atomically.

## What Needs to Change

Both `updateContractMetadata` and `transitionContractStatus` need to wrap all their work in a single `handles.withHandle()` block so that the label merges and the outbox event fire all happen in the same transaction:

```java
public void updateContractMetadata(String groupId, String artifactId, String contractId,
        EditableContractMetadataDto metadata) {
    handles.withHandle(handle -> {
        String prefix = contractId != null
                ? ContractLabels.contractPrefix(contractId) : ContractLabels.PREFIX;
        mergeArtifactLabels(groupId, artifactId, prefix,
                contractMetadataMapper.toLabels(metadata, prefix));
        outboxEvent.fire(SqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
        return null;
    });
}
```

Since `mergeArtifactLabels` internally calls `handles.withHandle()`, it will detect the existing handle and reuse the same connection — so both the label update and the outbox insert happen in the same transaction.

The same pattern applies to `transitionContractStatus` — wrap the status label update, the conditional date label updates, and the outbox event fire in a single `handles.withHandle()` block.

## Note on Existing Code

The existing `setArtifactContractRuleset`, `deleteArtifactContractRuleset`, etc. methods in `AbstractSqlRegistryStorage` appear to have the same issue (repository call in one transaction, outbox fire in another). That's pre-existing debt and not something you need to fix in this PR, but it's worth noting for a follow-up.

EricWittmann:
Note:  but if you **want** to fix those additional transaction boundary issues in this PR, I wouldn't be upset.  :smiley:

vivek-gite:
> Note: but if you **want** to fix those additional transaction boundary issues in this PR, I wouldn't be upset. 😃

Hey @EricWittmann 

> Thanks for the work @vivek-gite — the overall architecture (moving contract events into the storage layer, adding KafkaSQL outbox events for all 6 ruleset operations) is exactly right. However, there's a transactional boundary issue in the SQL implementation that breaks the outbox pattern. This needs to be fixed before we can merge.
> ## How Transaction Boundaries Work in Apicurio Registry
> 
> We don't use `@Transactional` in Apicurio Registry. Instead, transactional boundaries are managed by `AbstractHandleFactory.withHandle()`:
> 
>     * The **outermost** `withHandle()` call creates a JDBC connection with `autocommit=false` and manages commit/rollback
> 
>     * **Nested** `withHandle()` calls detect the existing connection (via a ThreadLocal `LocalState`) and reuse it — incrementing a nesting counter instead of opening a new connection
> 
>     * Only the outermost call commits or rolls back when it completes
> 
> 
> This means: **for two operations to be atomic, they must both execute inside the same outermost `withHandle()` callback.**
> ## The Outbox Pattern Requirement
> 
> For the outbox pattern to work correctly, the **functional change** (e.g., updating contract metadata labels) and the **outbox row insert** (triggered by `outboxEvent.fire()`) must happen in the **same database transaction**. If they're in separate transactions, a crash between the two means the data changed but the event was never published — silent data loss from the consumer's perspective.
> ## The Problem
> 
> The new `updateContractMetadata` method in `AbstractSqlRegistryStorage` does this:
> 
> ```java
> public void updateContractMetadata(String groupId, String artifactId, String contractId,
>         EditableContractMetadataDto metadata) {
>     String prefix = contractId != null
>             ? ContractLabels.contractPrefix(contractId) : ContractLabels.PREFIX;
>     // Transaction #1: mergeArtifactLabels() → repository → handles.withHandle() → commit
>     mergeArtifactLabels(groupId, artifactId, prefix,
>             contractMetadataMapper.toLabels(metadata, prefix));
>     // Transaction #2: outboxEvent.fire() → SqlEventsProcessor.processEvent(@Observes)
>     //   → SqlEventRepository.createEvent() → handles.withHandle() → commit
>     outboxEvent.fire(SqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
> }
> ```
> 
> `mergeArtifactLabels()` calls through to a repository method that wraps its work in `handles.withHandle()` — that's **Transaction #1**, which commits and closes. Then `outboxEvent.fire()` triggers `SqlEventsProcessor.processEvent()` synchronously (`@Observes`, not `@ObservesAsync`), which calls `SqlEventRepository.createEvent()` — that opens its **own** `handles.withHandle()` — **Transaction #2**.
> 
> These are two independent transactions. A crash between them means the metadata was updated but the event was never published.
> 
> The same issue applies to `transitionContractStatus`, which has multiple `mergeArtifactLabels` calls (each in their own transaction), followed by the outbox event in yet another separate transaction.
> ## The Correct Pattern
> 
> Methods like `createArtifact` do it right — they call `outboxEvent.fire()` **inside** a `handles.withHandle()` callback:
> 
> ```java
> return handles.withHandle(handle -> {
>     // functional work uses the handle
>     handle.createUpdate(sqlStatements.insertArtifact())...execute();
>     // ...
> 
>     // outbox event fires INSIDE the withHandle callback
>     // SqlEventsProcessor → createEvent() → nested withHandle() → REUSES same connection
>     outboxEvent.fire(SqlOutboxEvent.of(ArtifactCreated.of(amdDto)));
>     return result;
> });
> ```
> 
> Because `outboxEvent.fire()` executes inside the outer `withHandle()`, the nested `withHandle()` in `SqlEventRepository.createEvent()` detects the existing connection and reuses it — same transaction. The outermost call commits everything atomically.
> ## What Needs to Change
> 
> Both `updateContractMetadata` and `transitionContractStatus` need to wrap all their work in a single `handles.withHandle()` block so that the label merges and the outbox event fire all happen in the same transaction:
> 
> ```java
> public void updateContractMetadata(String groupId, String artifactId, String contractId,
>         EditableContractMetadataDto metadata) {
>     handles.withHandle(handle -> {
>         String prefix = contractId != null
>                 ? ContractLabels.contractPrefix(contractId) : ContractLabels.PREFIX;
>         mergeArtifactLabels(groupId, artifactId, prefix,
>                 contractMetadataMapper.toLabels(metadata, prefix));
>         outboxEvent.fire(SqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
>         return null;
>     });
> }
> ```
> 
> Since `mergeArtifactLabels` internally calls `handles.withHandle()`, it will detect the existing handle and reuse the same connection — so both the label update and the outbox insert happen in the same transaction.
> 
> The same pattern applies to `transitionContractStatus` — wrap the status label update, the conditional date label updates, and the outbox event fire in a single `handles.withHandle()` block.
> ## Note on Existing Code
> 
> The existing `setArtifactContractRuleset`, `deleteArtifactContractRuleset`, etc. methods in `AbstractSqlRegistryStorage` appear to have the same issue (repository call in one transaction, outbox fire in another). That's pre-existing debt and not something you need to fix in this PR, but it's worth noting for a follow-up.

@EricWittmann Thanks for the detailed review — you’re absolutely right. I mistakenly treated the sequential label update and synchronous outbox event firing as atomic, without tracing the actual `withHandle()` transaction boundary.

I’ve updated both `updateContractMetadata` and `transitionContractStatus` to wrap all label mutations and `outboxEvent.fire()` in one outer `handles.withHandle(...)` callback. The nested label and outbox repository calls now reuse the same connection, so a failure rolls back both the functional change and outbox row together.

I also added regression tests that force the outbox operation to fail and verify that no contract labels are committed. I left the existing ruleset methods unchanged, as you noted they are pre-existing debt and outside this PR’s scope.

Thank you again for catching this.

paoloantinori (commented):
The PR correctly pushes contract metadata/status event emission into the storage layer and adds the missing KafkaSQL journal messages — the right structural direction. The main risk is a likely double-emit of the metadata/status events under KafkaSQL, and the PR's own regression test for that path was not executed.

- `AbstractSqlRegistryStorage.updateContractMetadata`/`transitionContractStatus` (app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java ~L850 and ~L875) each fire a `SqlOutboxEvent`, while `KafkaSqlRegistryStorage.updateContractMetadata`/`transitionContractStatus` (app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java L695 and L706) also fire a `KafkaSqlOutboxEvent` for the same event after journal ack. The six ruleset operations fire only from the KafkaSql wrapper (their SQL delegate methods do not fire), so for metadata/status the journal dispatch — which invokes the SQL delegate via `dispatchTo` — could publish the same event twice in KafkaSQL mode. Please confirm the kafkasql profile suppresses the delegate's `SqlOutboxEvent`, otherwise metadata/status will double-fire where the ruleset ops do not.
- Per the PR Testing section only `ReadOnlyRegistryStorageTest` was run; `KafkaSqlContractEventsTest#testContractEventsArePublished` (which asserts exactly two contract events and is the guard for the issue above) and the new `AbstractSqlRegistryStorageContractTransactionTest` were not executed, so neither the KafkaSQL journal event payload nor the rollback behavior was verified locally before this review.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -679,6 +686,27 @@ public void deleteGroupRules(String groupId) throws RegistryStorageException {
         coordinator.waitForResponse(uuid);
     }
 
+    @Override
+    public void updateContractMetadata(String groupId, String artifactId, String contractId,
+            EditableContractMetadataDto metadata) throws RegistryStorageException {
+        var message = new UpdateContractMetadata4Message(groupId, artifactId, contractId, metadata);
+        var uuid = blockOnResult(submitter.submitMessage(message));
+        coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
```
paoloantinori, line 695: This fires `ContractMetadataUpdated` from the KafkaSql wrapper after journal ack, but the SQL delegate reached via `UpdateContractMetadata4Message.dispatchTo` also fires `SqlOutboxEvent.of(ContractMetadataUpdated...)` in `AbstractSqlRegistryStorage.updateContractMetadata` (~L850). The ruleset operations in this class avoid this because their SQL delegate methods don't fire an event — so for metadata/status the journal dispatch path could publish the same event twice in KafkaSQL mode. Could you confirm the kafkasql profile makes the delegate's `SqlOutboxEvent` inert, or move metadata/status to the same single-fire pattern used by the ruleset ops? `testContractEventsArePublished` asserts exactly two events and would catch a duplicate, but per the PR Testing notes that test wasn't run.

vivek-gite:
> The PR correctly pushes contract metadata/status event emission into the storage layer and adds the missing KafkaSQL journal messages — the right structural direction. The main risk is a likely double-emit of the metadata/status events under KafkaSQL, and the PR's own regression test for that path was not executed.
> 
>     * `AbstractSqlRegistryStorage.updateContractMetadata`/`transitionContractStatus` (app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java ~L850 and ~L875) each fire a `SqlOutboxEvent`, while `KafkaSqlRegistryStorage.updateContractMetadata`/`transitionContractStatus` (app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java L695 and L706) also fire a `KafkaSqlOutboxEvent` for the same event after journal ack. The six ruleset operations fire only from the KafkaSql wrapper (their SQL delegate methods do not fire), so for metadata/status the journal dispatch — which invokes the SQL delegate via `dispatchTo` — could publish the same event twice in KafkaSQL mode. Please confirm the kafkasql profile suppresses the delegate's `SqlOutboxEvent`, otherwise metadata/status will double-fire where the ruleset ops do not.
> 
>     * Per the PR Testing section only `ReadOnlyRegistryStorageTest` was run; `KafkaSqlContractEventsTest#testContractEventsArePublished` (which asserts exactly two contract events and is the guard for the issue above) and the new `AbstractSqlRegistryStorageContractTransactionTest` were not executed, so neither the KafkaSQL journal event payload nor the rollback behavior was verified locally before this review.

Hi @paoloantinori Thanks for flagging this. I traced the delegate path and confirmed that SqlOutboxEvent is inert in KafkaSQL mode: the delegate uses the in-memory H2 store, while SqlEventsProcessor only persists database events for PostgreSQL or MSSQL. The KafkaSqlOutboxEvent fired by the wrapper is therefore the single publish path.
- I have ran KafkaSqlContractEventsTest, and AbstractSqlRegistryStorageContractTransactionTest missed to update the PR summary; both tests pass, including the exact-two-events assertion, so it also verifies that metadata/status events are not duplicated.

paoloantinori:
Thanks @vivek-gite for the ping. My code-review position is unchanged from the earlier passes: the storage-layer move and the journal-then-publish outbox additions are correct, and the double-emit risk I flagged before is addressed. Nothing new for me to add from a code-review standpoint.

The remaining blocker is CI — the suite failed on `c0b4916` and `b4857fa`. That's outside my review scope, but worth flagging that the green run will be needed before merge.

vivek-gite:
> Thanks @vivek-gite for the ping. My code-review position is unchanged from the earlier passes: the storage-layer move and the journal-then-publish outbox additions are correct, and the double-emit risk I flagged before is addressed. Nothing new for me to add from a code-review standpoint.
> 
> The remaining blocker is CI — the suite failed on `c0b4916` and `b4857fa`. That's outside my review scope, but worth flagging that the green run will be needed before merge.

Thanks, @paoloantinori . The current verification run is now green: the Verification Gate and all regular unit-test shards, including KafkaSQL, passed.
The remaining red entries are Scalpel shadow jobs, which are non-blocking. I traced those failures to Scalpel’s selective-build behavior: it skips the schema-util-common test JAR even though schema-util-json has a test dependency on it. This is unrelated to the contract-event changes; the normal non-Scalpel test shards all complete successfully

paoloantinori:
Thanks @vivek-gite. My code-review position is unchanged — the storage-layer move and journal-then-publish outbox additions are correct, and nothing new has landed on the diff since my last read. The remaining blocker is CI (suite failed on `54ea654`); once that's green, this is good to go from my side.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/messages/TransitionContractStatus6Message.java
```diff
@@ -0,0 +1,52 @@
...
+@EqualsAndHashCode(callSuper = false)
+@ToString
+public class TransitionContractStatus6Message extends AbstractMessage {
+
+    private String groupId;
+    private String artifactId;
+    private String contractId;
+    private ContractStatus fromStatus;
+    private ContractStatus toStatus;
+    private String transitionDate;
```
paoloantinori, line 44: This message and UpdateContractMetadata4Message carry enums/DTOs (ContractStatus, EditableContractMetadataDto) through the KafkaSQL journal serde. That round-trip — and single-publish of the resulting events — is exercised only by KafkaSqlContractEventsTest, which is @Tag(SLOW) and absent from the PR's run list. Run those tests against Kafka (or confirm CI runs them) before relying on this, since a serde mismatch here would surface only at runtime under KafkaSQL.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -832,6 +839,45 @@ public void deleteGroupRule(String groupId, RuleType rule) throws RegistryStorag
...
 
+    @Override
+    public void updateContractMetadata(String groupId, String artifactId, String contractId,
+            EditableContractMetadataDto metadata) throws RegistryStorageException {
+        handles.withHandle(handle -> {
+            String prefix = contractId != null
+                    ? ContractLabels.contractPrefix(contractId) : ContractLabels.PREFIX;
+            mergeArtifactLabels(groupId, artifactId, prefix,
+                    contractMetadataMapper.toLabels(metadata, prefix));
+            outboxEvent.fire(SqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
```
paoloantinori, line 851: Event publication now lives inside each storage impl rather than in the resource's storage.createEvent(). SQL and KafkaSQL fire the outbox event here; read-only impls throw. Confirm every active RegistryStorage implementation fires the outbox event in updateContractMetadata/transitionContractStatus — any other impl that implements these methods without firing would silently lose contract events, a regression versus the previous uniform createEvent path.

.github/workflows/verify-unit-tests.yaml
```diff
@@ -100,9 +100,11 @@ jobs:
           # -Dsurefire.failIfNoSpecifiedTests=false: don't fail if a shard's test filter matches zero tests
           # -Dsurefire.rerunFailingTestsCount=2: automatically retry flaky tests up to 2 times
           # -Daether.syncContext.named.factory=noop: prevent Maven 3.9+ resolver lock deadlocks under -T
+          # Keep upstream test compilation and test-JAR packaging enabled: downstream modules
+          # can depend on an upstream module's attached test JAR.
           ./mvnw verify \
             --no-transfer-progress -s ${{ github.workspace }}/.github/ci-settings.xml \
```
paoloantinori, line 115: This override applies to every scalpel-enabled shard and isn't called out in the PR description. The flag name implies skipTestsForUpstream=false will RUN upstream tests (not just attach their test JARs), which could slow CI or surface flakes across shards. Confirm the intent is test-JAR packaging only and that other shards aren't regressed.

paoloantinori (commented):
Clean, well-layered fix: moving contract metadata/status writes and their events into the storage layer (with KafkaSQL journaling and post-ack event firing) correctly addresses the missing KafkaSQL events and retires the REST-layer `createEvent` workaround; the main risk is verification rather than design. The most important concern is that the SLOW KafkaSQL tests that prove this fix — and that exercise the new journal-message serde — are not in the PR's run list.

- `KafkaSqlContractEventsTest.testContractEventsArePublished` and `testAllContractRulesetEventsArePublished` (new, `@Tag(ApicurioTestTags.SLOW)`) are the only tests verifying single-publish of contract events. The PR's listed runs are test-compile, checkstyle, and `ReadOnlyRegistryStorageTest` only — confirm CI actually runs these SLOW tests, since they are the sole proof of the fix.
- The new journal messages `TransitionContractStatus6Message` (carrying `ContractStatus` enums) and `UpdateContractMetadata4Message` (carrying `EditableContractMetadataDto`) round-trip through the KafkaSQL serde only via those SLOW tests; without them, that path is unverified.
- Event publication moved out of the resource's `storage.createEvent()` into each implementation (`AbstractSqlRegistryStorage.java:850`, `KafkaSqlRegistryStorage.java`). SQL and KafkaSQL fire the outbox event; read-only impls correctly throw. Confirm there is no other active `RegistryStorage` impl (e.g. a default-profile/in-memory storage) that implements these methods without firing — it would silently drop events the old uniform `createEvent` used to publish.
- `AbstractSqlRegistryStorageContractTransactionTest` stubs `mergeArtifactLabels`, so its rollback assertion holds only if the real `mergeArtifactLabels` joins the surrounding `withHandle` transaction; worth confirming the real path is transactional so the atomicity claim is genuine.
- The CI change `-Dscalpel.skipTestsForUpstream=false` (`verify-unit-tests.yaml:106`) applies to every scalpel-enabled shard and is not mentioned in the PR description; confirm it only attaches upstream test JARs and does not run upstream tests, which could slow CI or surface flakes.

vivek-gite:
> Clean, well-layered fix: moving contract metadata/status writes and their events into the storage layer (with KafkaSQL journaling and post-ack event firing) correctly addresses the missing KafkaSQL events and retires the REST-layer `createEvent` workaround; the main risk is verification rather than design. The most important concern is that the SLOW KafkaSQL tests that prove this fix — and that exercise the new journal-message serde — are not in the PR's run list.
> 
>     * `KafkaSqlContractEventsTest.testContractEventsArePublished` and `testAllContractRulesetEventsArePublished` (new, `@Tag(ApicurioTestTags.SLOW)`) are the only tests verifying single-publish of contract events. The PR's listed runs are test-compile, checkstyle, and `ReadOnlyRegistryStorageTest` only — confirm CI actually runs these SLOW tests, since they are the sole proof of the fix.
> 
>     * The new journal messages `TransitionContractStatus6Message` (carrying `ContractStatus` enums) and `UpdateContractMetadata4Message` (carrying `EditableContractMetadataDto`) round-trip through the KafkaSQL serde only via those SLOW tests; without them, that path is unverified.
> 
>     * Event publication moved out of the resource's `storage.createEvent()` into each implementation (`AbstractSqlRegistryStorage.java:850`, `KafkaSqlRegistryStorage.java`). SQL and KafkaSQL fire the outbox event; read-only impls correctly throw. Confirm there is no other active `RegistryStorage` impl (e.g. a default-profile/in-memory storage) that implements these methods without firing — it would silently drop events the old uniform `createEvent` used to publish.
> 
>     * `AbstractSqlRegistryStorageContractTransactionTest` stubs `mergeArtifactLabels`, so its rollback assertion holds only if the real `mergeArtifactLabels` joins the surrounding `withHandle` transaction; worth confirming the real path is transactional so the atomicity claim is genuine.
> 
>     * The CI change `-Dscalpel.skipTestsForUpstream=false` (`verify-unit-tests.yaml:106`) applies to every scalpel-enabled shard and is not mentioned in the PR description; confirm it only attaches upstream test JARs and does not run upstream tests, which could slow CI or surface flakes.

@paoloantinori, One correction to my earlier note: the repo config sets `scalpel.skipTestsForUpstream=true`; overriding it to `false` is meaningful.

1. **CI coverage of SLOW event tests**

   Confirmed. The linked `app-kafkasql` job ran `KafkaSqlContractEventsTest`; both tests passed. The `SLOW` tag does not exclude this class from that shard, so CI exercises the single-publish assertions.

2. **KafkaSQL journal serde**

   Confirmed by the same CI test class. It drives contract metadata updates and status transitions through KafkaSQL, covering the journal round trip for `UpdateContractMetadata4Message` (`EditableContractMetadataDto`) and `TransitionContractStatus6Message` (`ContractStatus`).

3. **All storage implementations publish events**

   Confirmed by audit:

   - SQL variants inherit `AbstractSqlRegistryStorage` and fire the SQL outbox event.
   - KafkaSQL fires the corresponding event directly.
   - GitOps/Kubernetes polling implementations are read-only and reject contract mutations.
   - The read-only decorator delegates writable calls after its guard.
   - The in-memory implementation is test-only and does not silently act as an active writable storage implementation.

   There is no active writable implementation that accepts these methods without publishing an event.

4. **Transactionality of `mergeArtifactLabels`**

   Fixed and strengthened. Contract updates now pass the outer `Handle` explicitly to `mergeArtifactLabels`, rather than relying on nested handle resolution. The transaction test now verifies that the exact outer handle is used and that labels remain uncommitted when outbox insertion fails.

5. **Scalpel upstream tests / test-JARs**

   The original `-Dscalpel.skipTestsForUpstream=false` did more than attach test JARs: it allowed upstream tests to execute. I updated the workflow to retain upstream test compilation/test-JAR packaging while skipping only upstream test execution:

   ```text
   -Dscalpel.skipTestsForUpstream=false
   -Dscalpel.upstreamArgs=skipTests=true
   ```

   `skipTests=true` is applied only to Scalpel’s upstream-only modules, so selected shard tests still run.

paoloantinori:
Thanks @vivek-gite for the quote-back. My code-review position is unchanged from the earlier passes — the storage-layer move, the journal-then-publish outbox additions, and the new transaction-rollback test are correct, and I have no new findings on this head. The remaining blocker from the repo's side is the failing CI runs (e.g. commit b82f94e) rather than the diff itself; once those are green I'd consider this ready from my end.

vivek-gite:
> Thanks @vivek-gite for the quote-back. My code-review position is unchanged from the earlier passes — the storage-layer move, the journal-then-publish outbox additions, and the new transaction-rollback test are correct, and I have no new findings on this head. The remaining blocker from the repo's side is the failing CI runs (e.g. commit [b82f94e](https://github.com/Apicurio/apicurio-registry/commit/b82f94ea266caeb28a8addf1cf7c6b03c6200d79)) rather than the diff itself; once those are green I'd consider this ready from my end.

@paoloantinori I pushed new commit ([ec93dc6](https://github.com/Apicurio/apicurio-registry/pull/9066/commits/ec93dc67f4e3b42c084e41774db7763a64037b92)) which is passing all the required CI runs.

CC: @EricWittmann

paoloantinori:
Thanks @vivek-gite for the quote-back. Confirming my position is unchanged on this head: the storage-layer move, the journal-then-publish outbox additions, and the transaction-rollback test are all correct, and nothing new has landed on the diff since my last read. The remaining blocker is the failing CI runs, not the diff itself.

vivek-gite:
@paoloantinori sorry correct me if I'm wrong but i don't see any CI failures on the latest commit after u suggest the changes 😅
image

Could you please link the failing run you are referring to? It may be an earlier run that predates the current green verification.

paoloantinori:
You're right, @vivek-gite, and I was wrong — apologies for the confusion.

I just checked the CI state on the current head (`ec93dc6`): every check run is green — the Scalpel unit-test shards, all Integration Tests variants (h2/postgresql/kafkasql × default/auth/migration), and the Verification Gate are all `success`, and the combined status is `success`. No failures, nothing pending.

My "still failing" comments were stale — they were tracking the earlier commits (`c0b4916`, `b4857fa`, `54ea654`, `b82f94e`) that did fail, not this head. You've resolved those. From the code-review side my position is unchanged (the storage-layer move and the journal-then-publish outbox additions are correct), and CI on this head is clear.

vivek-gite:
@EricWittmann Can you please merge this PR. Lot of my other PRs are in pipeline

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1294,7 +1336,7 @@ public String createSnapshot(String snapshotLocation) throws RegistryStorageExce
 
     @Override
     public String createEvent(OutboxEvent event) {
```
paoloantinori, line 1336: Restoring `createEvent()` to a no-op is correct for the contract paths migrated here (both `GroupsResourceImpl` callers are removed in this PR). Worth a quick grep to confirm no other caller still treats `createEvent()` as a real publish in KafkaSQL mode — any remaining caller would now silently drop its event, since this method no longer fires the outbox. If it was only ever the contract workaround, this is safe as-is.

paoloantinori (commented):
Overall this is a clean, well-scoped fix: contract metadata/status events are moved out of the REST layer into the storage layer, KafkaSQL publishes its outbox events only after journal acknowledgement, and the change carries strong regression coverage (a transaction-rollback test and a duplicate-event-detection test). I didn't find correctness bugs in the diff; one cross-cutting assumption is worth a quick confirmation.

- `KafkaSqlRegistryStorage.createEvent()` is restored to a no-op (`KafkaSqlRegistryStorage.java:1338`). Both REST-layer contract callers are removed in this PR, so the migrated paths are safe — just confirm via grep that no other caller still relies on `createEvent()` actually publishing in KafkaSQL mode, since any such caller would now silently drop its event. If `createEvent()` was only ever used by the contract workaround, this is safe as-is.
- The new `updateContractMetadata`/`transitionContractStatus` fire `SqlOutboxEvent` inside `handles.withHandle(...)` (`AbstractSqlRegistryStorage.java:842-879`), so the label writes and the outbox insert share one transaction; `AbstractSqlRegistryStorageContractTransactionTest` exercises the rollback-on-outbox-failure path, correctly preserving atomicity the old inline REST code lacked.
- All KafkaSQL contract ops publish `KafkaSqlOutboxEvent` strictly after `coordinator.waitForResponse(uuid)` (`KafkaSqlRegistryStorage.java:688-790`), and the duplicate-detection window in `KafkaSqlContractEventsTest` guards against the double-publish the old `createEvent` workaround could have caused. Good.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
vivek-gite: Confirmed. I audited the production call sites: the two `GroupsResourceImpl` calls removed by this PR were the only callers that could invoke `RegistryStorage.createEvent(...)` in KafkaSQL mode. The remaining occurrences are declarations/delegates/implementations, plus `SqlEventsProcessor` on the SQL-storage path. Therefore, restoring KafkaSQL `createEvent()` to a no-op is safe.

vivek-gite:
@paoloantinori can you merge this PR now?

paoloantinori:
Hi @vivek-gite — thanks for the nudge. From the code-review side my position is unchanged: the storage-layer move, the journal-then-publish outbox additions, and the transaction-rollback coverage all look correct on this head. Merging is a maintainer call, though, not something I can do — @EricWittmann already `/accept`ed earlier, so the merge decision rests with the maintainers.

vivek-gite:
CC: @EricWittmann

PR #9071, opened 2026-07-30
fix(storage): gzip snapshot dumps and delete old dumps

### LINKED ISSUE #8961: gzip snapshot dumps and optionally store them in Kafka for K8s without PVC

Snapshots are meant to speed up replica startup by restoring H2 from a dump instead of replaying the whole `kafkasql-journal` ([blog](https://www.apicur.io/blog/2024/06/21/registry-3-kafkasql)).

Today the snapshots topic mostly stores a **filesystem path**. The dump itself is written under `apicurio.storage.snapshot.location`, which defaults to `java.io.tmpdir` (often `/tmp` in containers).
On Kubernetes without a PVC:

1. Pod restarts and dump files are gone
2. New pod can’t restore and full journal replay
3. Startup gets slower as the journal grows

Wiring automatic snapshots ([#8499](https://github.com/Apicurio/apicurio-registry/issues/8499) / [#8517](https://github.com/Apicurio/apicurio-registry/pull/8517)) helps when files are durable, but it doesn’t fix the ephemeral-disk case by itself.

Also, dumps are currently plain H2 `SCRIPT TO` (uncompressed) https://github.com/Apicurio/apicurio-registry/issues/8499#issuecomment-5107788328, so they use more disk than they need to especially once snapshots run on a schedule.

### Proposed Solution

Happy to implement if this direction sounds OK to maintainers.

**1. Gzip the H2 dumps**  
Use H2 `SCRIPT TO … COMPRESSION GZIP` / `RUNSCRIPT … COMPRESSION GZIP` (keep reading legacy plain `.sql` if needed). Cleaning up older dump files helps `/tmp` even more.

**2. Optional Kafka-resident dump bytes (no PVC)**  
After creating the dump, optionally publish the gzipped bytes as chunked records on the snapshots topic (or a dedicated topic), and keep only the latest snapshot.
On startup: if the local file is missing, restore from Kafka, then only replay journal messages after the snapshot.

### Questions for maintainers

- Is Kafka-resident dump storage something you’d consider for no-PVC deployments?
- Any concerns around message size, chunking, or topic disk growth if old dumps aren’t cleaned up?
- OK to start with gzip (+ delete old local dumps) as a small first PR?

I’m interested in working on this, but I’d like a quick maintainer take before spending more time on a full prototype.

itsvishalyadav:
Fixes #8961

### What
Enables GZIP compression for local H2 database snapshots and adds a cleanup mechanism to safely delete old local snapshot dumps.

### Why
To address the high disk usage footprint (especially for \/tmp\) and slow replica startup times, which is particularly problematic in ephemeral environments like Kubernetes without a PVC.

### How
- Modified \SqlStatements\ and \H2SqlStatements\ to conditionally apply H2's native \COMPRESSION GZIP\ suffix based on the file extension.
- Updated \KafkaSqlRegistryStorage\ to generate new snapshot files with the \.sql.gz\ extension.
- Added a cleanup block in \	riggerSnapshotCreation()\ using \java.nio.file.Files.newDirectoryStream\ to actively clean up older \.sql\ and \.sql.gz\ dump files after a new snapshot is successfully generated.


app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1284,6 +1284,23 @@ public String triggerSnapshotCreation() throws RegistryStorageException {
         ProducerRecord<String, String> record = new ProducerRecord<>(configuration.getSnapshotsTopic(), 0,
                 snapshotId, snapshotLocation, Collections.emptyList());
         RecordMetadata recordMetadata = blockOnResult(snapshotsProducer.apply(record));
+        
+        // Cleanup old local dumps to save disk space
+        try {
+            Path storeDir = Path.of(configuration.getSnapshotStoreLocation());
+            if (java.nio.file.Files.isDirectory(storeDir)) {
+                try (java.nio.file.DirectoryStream<Path> stream = java.nio.file.Files.newDirectoryStream(storeDir, "*.{sql,sql.gz}")) {
```
malladinagarjuna2: the snapshot location defaults to  "./", so the glob *.{sql,sql.gz} can wipe any unrelated .sql file sitting in the working directory the first time an admin triggers a snapshot. Delete based on "did we create this", not on a glob.

Fix: only delete files whose basename parses as a UUID and require Files.isRegularFile. It is better to  keep the N most recent instead of "everything but the newest", one bad snapshot currently leaves zero fallbacks.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1284,6 +1284,23 @@ public String triggerSnapshotCreation() throws RegistryStorageException {
...
+                    for (Path p : stream) {
+                        if (!p.getFileName().toString().startsWith(snapshotId)) {
```
paoloantinori: This cleanup deletes every dump file whose name doesn't start with the new `snapshotId`. The actual snapshot file is written asynchronously by the consumer of the snapshots topic *after* this method returns — the cleanup here runs immediately after the producer ack, so the new file does not exist yet. If generation subsequently fails (disk full, OOM, consumer error), all previously-successful snapshots have already been deleted and the store is left empty, leaving nothing to restore from. Consider gating deletion on confirmation that the new snapshot was actually generated, and/or retaining the last successful snapshot (or the N most recent) instead of deleting everything else.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1284,6 +1284,23 @@ public String triggerSnapshotCreation() throws RegistryStorageException {
         ProducerRecord<String, String> record = new ProducerRecord<>(configuration.getSnapshotsTopic(), 0,
                 snapshotId, snapshotLocation, Collections.emptyList());
         RecordMetadata recordMetadata = blockOnResult(snapshotsProducer.apply(record));
+        
+        // Cleanup old local dumps to save disk space
+        try {
+            Path storeDir = Path.of(configuration.getSnapshotStoreLocation());
+            if (java.nio.file.Files.isDirectory(storeDir)) {
+                try (java.nio.file.DirectoryStream<Path> stream = java.nio.file.Files.newDirectoryStream(storeDir, "*.{sql,sql.gz}")) {
+                    for (Path p : stream) {
```
paoloantinori: There's no synchronization around snapshot triggers and no assertion that the store location is node-local. Concurrent triggers (leader handover, a retried/duplicate trigger) each delete all non-matching files, so two in-flight snapshots can mutually wipe the directory; and on deployments using a shared store location (PVC/NFS rather than ephemeral per-pod `/tmp`) the leader can delete a file another node is actively restoring from, yielding a truncated or failed restore. Consider serializing snapshot triggers and either asserting node-local storage or coordinating deletes.

app/src/main/java/io/apicurio/registry/storage/impl/sql/H2SqlStatements.java
```diff
@@ -81,11 +81,27 @@ public String createDataSnapshot() {
         return "SCRIPT TO ?";
     }
 
+    @Override
+    public String createDataSnapshot(String location) {
+        if (location != null && location.endsWith(".gz")) {
```
paoloantinori: Whether to apply COMPRESSION is derived from a `".gz"` suffix checked here, while the filename is built independently in `KafkaSqlRegistryStorage` (`snapshotId + ".sql.gz"`). The two are coupled only by convention — the literal `".gz"` appears in both but isn't shared. If the extension ever changes in one place but not the other (e.g. `.gzip`, or a timestamped suffix), create and restore silently disagree: a gzip file restored as plaintext (or vice-versa) corrupts or fails the restore. Extract a shared constant for the suffix, or pass an explicit compression flag rather than re-deriving it from the filename.

paoloantinori (commented):
Overall, the GZIP wiring and extension-based dispatch are sound and backward-compatible, but the new cleanup block in `triggerSnapshotCreation()` deletes all prior snapshots before the new one is confirmed on disk and retains zero fallback copies — a real data-loss risk for a mechanism whose purpose is disaster recovery.

- **`KafkaSqlRegistryStorage.java` (~1288–1295):** the snapshot file is generated asynchronously by the consumer of the snapshots topic *after* this method returns, yet cleanup runs immediately after the producer ack. If generation later fails (disk full, OOM, consumer error), every previously-successful snapshot has already been deleted and the store is left empty, so there is nothing left to restore from. Recommend gating deletion on confirmed generation, and/or retaining the last successful snapshot (or the N most recent) rather than deleting everything that doesn't match the current snapshotId.
- **`KafkaSqlRegistryStorage.java` (~1293):** there is no synchronization around snapshot triggers and no assertion that the store is node-local. Concurrent triggers (leader handover, a retried/duplicate trigger) each delete all non-matching files; and on deployments where the store location is shared (PVC/NFS rather than ephemeral per-pod `/tmp`) the leader can delete a file another node is mid-restore on. Consider serializing triggers and asserting/documenting node-local storage.
- **`H2SqlStatements.java` (~86):** the compression decision is derived from a `".gz"` suffix checked here while the filename is built independently in `KafkaSqlRegistryStorage` — coupled only by convention. A future extension change in one place (e.g. `.gzip`, a timestamped suffix) silently breaks the create/restore pair: a gzip file restored without `COMPRESSION GZIP` (or vice-versa) corrupts or fails the restore. Extract a shared suffix constant or pass an explicit compression flag.
- **`H2SqlStatements.java` (verification, not a confirmed defect):** the `SCRIPT TO ? COMPRESSION GZIP` / `RUNSCRIPT FROM ? COMPRESSION GZIP` clause position looks correct per H2 docs and the bound-path parameter already worked pre-change, but I couldn't confirm against the exact H2 version in use — worth a quick integration test, since a grammar mismatch here would break every snapshot create and restore.

itsvishalyadav:
@paoloantinori thanks for the feedback! I've updated the PR to extract the extension into a constant and completely refactored the cleanup logic so it retains the 3 most recently modified snapshots and only runs after Files.exists validates the generation.

paoloantinori:
Thanks — the extension constant and the retention-based cleanup (keep 3 most recent, gated on `Files.exists(path)`) both address the previous feedback cleanly. The backward-compat detail is handled correctly too: the glob is `*.{sql,sql.gz}`, so legacy uncompressed dumps from older nodes are still reaped as the new compressed ones land. Good.

One gap before this is mergeable, and it's the same class of risk the first review flagged on the delete-ordering: **there's no test, and the cleanup is the most failure-prone path in the PR.**

The restore side has a real correctness property that's worth pinning: `H2SqlStatements.restoreFromSnapshot(location)` only emits `COMPRESSION GZIP` when the path ends in `.sql.gz`, otherwise plain `RUNSCRIPT FROM ?`. If that dispatch ever drifts from the write side (e.g. someone changes the extension constant or adds a `.zip` variant without updating both branches), a replica will try to `RUNSCRIPT` a gzip blob and fail at startup with an opaque H2 parse error — exactly the "ephemeral K8s replica won't start" scenario this PR is meant to improve. A unit test on `H2SqlStatements` that asserts:

- `createDataSnapshot("foo.sql.gz")` contains `COMPRESSION GZIP`
- `createDataSnapshot("foo.sql")` does not
- `restoreFromSnapshot("foo.sql.gz")` contains `COMPRESSION GZIP`
- `restoreFromSnapshot("foo.sql")` does not

…would lock the contract for ~zero cost and catch any future drift on either side. The cleanup logic itself (sort by mtime, keep 3) is harder to unit-test without a tmpdir harness; I wouldn't block on that, but the statement-dispatch assertions above are high-value and cheap.

The `catch (Exception e) { return 0; }` in the mtime comparator makes the sort non-deterministic on an I/O error, but since the just-written snapshot is the newest file and is protected by the `Files.exists(path)` gate, the worst case is deleting a slightly-older-but-still-wanted backup rather than the live one — acceptable for a best-effort cleanup. If you want it tighter, skip the delete on comparator failure rather than treating it as equal; minor.

Rest is solid. With the statement-dispatch test added, this is good to go from my side.

itsvishalyadav:
Thanks @paoloantinori for the detailed review! I've just pushed a new commit to address these final points:

1. Unit Test Added: I've added H2SqlStatementsTest.java with the exact assertions you requested to guarantee that COMPRESSION GZIP is only appended when the .sql.gz extension is provided. This locks in the contract and will catch any future drift.
2. Tightened Cleanup Sort: I updated the mtime comparator so that it now throws a RuntimeException on any IOException. This instantly bails out of the sort and safely skips the deletion phase entirely if we can't reliably determine the timestamp ordering, preventing the accidental deletion of wanted backups.

Let me know if there's anything else needed, otherwise this should be good to go!

paoloantinori:
Both land correctly on this head:

- `H2SqlStatementsTest` pins the exact contract — `COMPRESSION GZIP` present for `.sql.gz`, absent for `.sql` — on both the write and restore paths, so future drift gets caught.
- Throwing `RuntimeException` on `IOException` in the mtime comparator fails fast instead of silently mis-ordering the retention window.

No further concerns from me.

itsvishalyadav:
@paoloantinori 
Great! Thanks again for the review. It looks like the GitHub Actions workflows are currently awaiting maintainer approval to run on this latest commit whenever you or another maintainer have a moment.

paoloantinori:
Thanks for the heads-up. The workflow-approval prompt is expected for first-run on this branch — I can't action it from here, so flagging it for a maintainer click ("Approve and run" on the Actions tab) the next time someone is at a keyboard. No code change needed from your side.

itsvishalyadav:
Hi @jsenko @carlesarnal, just a gentle ping to see if one of you might be able to click 'Approve and run' on the Actions tab so the CI workflows can complete? No rush on my end, just wanted to bump this so it doesn't get lost. Thanks!

itsvishalyadav:
Hello @jsenko, thank you very much for reviewing it! It appears that the PR lifecycle bot on this repository uses /accept as opposed to /approve for new contributors to the project. Due to the lack of /accept command, the PR remains in the lifecycle/new state, leading to the automatic rejection at the Verification Gate because of 'PR is not in a testable lifecycle state'.

Can you kindly make use of the /accept command at your earliest convenience so that the tests could start running?

app/src/main/java/io/apicurio/registry/storage/impl/sql/H2SqlStatements.java
```diff
@@ -81,11 +81,27 @@ public String createDataSnapshot() {
...
     }
 
+    @Override
+    public String restoreFromSnapshot(String location) {
+        if (location != null && location.endsWith(SqlStatements.COMPRESSED_SNAPSHOT_EXTENSION)) {
+            return "RUNSCRIPT FROM ? COMPRESSION GZIP";
+        }
+        return "RUNSCRIPT FROM ?";
+    }
+
```
Joel-hanson, line 104: The newer snapshots will be sql.gz so does it make sense to keep one createDataSnapshot that does the non compression gzip and also avoid duplicates of `RUNSCRIPT FROM ?` for restoresFromSnapshot?

something like

```
@Override
public String createDataSnapshot() {
    return "SCRIPT TO ? COMPRESSION GZIP";
}

@Override
public String restoreFromSnapshot() {
    return "RUNSCRIPT FROM ? COMPRESSION GZIP";
}

restoreFromSnapshot(...) {
    if (location check......) {
        return restoreFromSnapshot();
    }
    return "RUNSCRIPT FROM ?";
}
```

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1284,6 +1286,36 @@ public String triggerSnapshotCreation() throws RegistryStorageException {
         ProducerRecord<String, String> record = new ProducerRecord<>(configuration.getSnapshotsTopic(), 0,
                 snapshotId, snapshotLocation, Collections.emptyList());
         RecordMetadata recordMetadata = blockOnResult(snapshotsProducer.apply(record));
+        
+        // Cleanup old local dumps to save disk space, keeping the most recent 3 backups.
```
Joel-hanson: nit. I think it would be nice to move the snapshot clean up to be a separate method from triggerSnapshotCreation

paoloantinori:
Thanks for the clear write-up on the lifecycle blocker — you're right that the bot recognises `/accept` rather than `/approve`, so the `/approve` from @jsenko won't progress the `lifecycle/new` state on its own.

The `/accept` command is a maintainer-side lifecycle action that I can't run from here, so I can't unblock it directly. Flagging it for @EricWittmann or @carlesarnal to run `/accept` when they get a chance — once that lands, the Verify jobs should kick in. The code side of this looks ready from my earlier review passes.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1284,6 +1286,36 @@ public String triggerSnapshotCreation() throws RegistryStorageException {
...
+        // We only clean up if the new snapshot actually exists on disk (successful generation).
+        try {
+            Path storeDir = Path.of(configuration.getSnapshotStoreLocation());
+            if (java.nio.file.Files.exists(path) && java.nio.file.Files.isDirectory(storeDir)) {
+                List<Path> snapshotFiles = new ArrayList<>();
+                try (java.nio.file.DirectoryStream<Path> stream = java.nio.file.Files.newDirectoryStream(storeDir, "*.{sql,sql.gz}")) {
```
Joel-hanson: What if the snapshot location is in a different folder and that folder contains sql files. its a rare case that might happen and if that happens it would delete other sql files or the newly created snapshot files. Also this would be a problem if its a shared volume, meaning if multiple apicurio use a shared volume to store snapshots

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1284,6 +1286,36 @@ public String triggerSnapshotCreation() throws RegistryStorageException {
...
+                    for (Path p : stream) {
+                        snapshotFiles.add(p);
+                    }
+                }
+                // Sort by last modified time descending (newest first)
+                snapshotFiles.sort((p1, p2) -> {
+                    try {
+                        return java.nio.file.Files.getLastModifiedTime(p2).compareTo(java.nio.file.Files.getLastModifiedTime(p1));
```
Joel-hanson: Can it be sorted by the snapshot id stored in the kafka topic. I think its usually fine but as I said before if there is a shared volume and has concurrent triggers for snapshot it might delete files unexpectedly. I am not sure how easy it is do that.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1284,6 +1286,36 @@ public String triggerSnapshotCreation() throws RegistryStorageException {
         ProducerRecord<String, String> record = new ProducerRecord<>(configuration.getSnapshotsTopic(), 0,
                 snapshotId, snapshotLocation, Collections.emptyList());
         RecordMetadata recordMetadata = blockOnResult(snapshotsProducer.apply(record));
```
Joel-hanson: nit. is recordMetadata used anywhere I don't see any other reference

app/src/test/java/io/apicurio/registry/storage/impl/sql/H2SqlStatementsTest.java
Joel-hanson, line 1: Is it possible to have cleanup test and gzip restore IT.

Joel-hanson (changes requested):

app/src/main/java/io/apicurio/registry/storage/impl/sql/H2SqlStatements.java
itsvishalyadav: Hi Joel, thank you for the review. This was something that crossed my mind, but earlier in the review, Paolo had asked me to have exactly this contract (COMPRESSION GZIP being there for .sql.gz but not for .sql) for the write and restore paths in order to catch any problems in the future in terms of compatibility. Hence, I kept those checks because of this requirement by Paolo.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
itsvishalyadav: Done! I have extracted this logic into a private cleanupOldSnapshots(Path newSnapshotPath) method and pushed the update.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
itsvishalyadav: The IDs of these snapshots are generated using UUID and not sequentially generated, hence sorting on the basis of ID cannot be done. Sorting based on lastModifiedTime ensures that the N latest backups are maintained at all times, and having the latest 3 prevents concurrent triggers from deleting each other.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
itsvishalyadav: I've removed the variable assignment in the latest push.

app/src/test/java/io/apicurio/registry/storage/impl/sql/H2SqlStatementsTest.java
itsvishalyadav: Paolo did point out in his review comment above that the cleanup code (sorting on mtime, taking 3) is more difficult to unit test without the use of the tmpdir framework, but that he wasn't blocking on this issue because the assertions related to statement dispatch are high-value ones. I'm happy to try adding one if you feel it's strictly necessary though.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
itsvishalyadav: The directory returned by configuration.getSnapshotStoreLocation() is intended specifically for Apicurio snapshots, so the glob *.{sql,sql.gz} is standard here. howeverto explicitly address the shared volume / concurrent trigger concern: the cleanup logic now uses .getLastModifiedTime(...) to sort the files and always retains the 3 most recent backups,this ensures that even if multiple nodes write snapshots concurrently, they won't wipe each other's in-flight files.

itsvishalyadav:
Hi @Joel-hanson, thanks for the review! I've replied to your comments and pushed the minor refactoring changes.

@jsenko @paoloantinori @carlesarnal The requested changes are up and the PR is ready for a final look and merge whenever you have a chance. Thanks!

paoloantinori:
Good — sorting the snapshot files by `getLastModifiedTime` and keeping the 3 newest (rather than deleting based on a name pattern) is the more robust approach for the shared-volume case: it avoids relying on stable naming and naturally tolerates the concurrency you described. Combined with the `Files.exists(path)` gate, the cleanup is now safe to leave as best-effort. I'm satisfied with the direction.

paoloantinori:
Thanks @itsvishalyadav — the code-side review is complete on my end; the concurrency concern raised earlier is resolved by the `getLastModifiedTime`-sorted retention in `cleanupOldSnapshots()`. Nothing further from me technically. As noted upthread, the remaining blocker for merge is the lifecycle state (`lifecycle/new` → needs `/accept` from a maintainer to progress), not the code.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1283,10 +1285,46 @@ public String triggerSnapshotCreation() throws RegistryStorageException {
         // message.
         ProducerRecord<String, String> record = new ProducerRecord<>(configuration.getSnapshotsTopic(), 0,
                 snapshotId, snapshotLocation, Collections.emptyList());
-        RecordMetadata recordMetadata = blockOnResult(snapshotsProducer.apply(record));
+        blockOnResult(snapshotsProducer.apply(record));
+        
+        cleanupOldSnapshots(path);
```
paoloantinori: This cleanup runs immediately after the snapshot marker is published, but the dump file is generated asynchronously by the consumer of the snapshots topic (H2 `SCRIPT TO`), not by `triggerSnapshotCreation`. Assuming `blockOnResult` blocks only on the producer ack (standard Kafka behavior — please correct me if it also waits on a generation-complete latch), `Files.exists(newSnapshotPath)` at line 1302 will be `false` at this point, so the cleanup short-circuits and old dumps are never deleted, defeating the purpose of this change. Consider moving the cleanup to the consumer path after the snapshot file is confirmed written, or restructuring so generation completes before cleanup runs.

paoloantinori (commented):
The GZIP compression wiring (write via `SCRIPT TO ... COMPRESSION GZIP` and the matching `RUNSCRIPT ... COMPRESSION GZIP` restore) is correct and unit-tested, but the new old-dump cleanup is effectively dead code because it runs before the snapshot file is generated.

- **Cleanup never executes** — `cleanupOldSnapshots(path)` is invoked in `KafkaSqlRegistryStorage.triggerSnapshotCreation()` right after `blockOnResult(snapshotsProducer.apply(record))` (line 1290), which only publishes a marker to the snapshots topic. The dump is written asynchronously by a separate consumer that runs H2 `SCRIPT TO`, so the just-triggered file does not exist on disk yet; the guard `Files.exists(newSnapshotPath)` (line 1302) is `false`, the `&&` short-circuits, and nothing is ever deleted. Old `.sql`/`.sql.gz` dumps accumulate unbounded — the exact `/tmp` disk issue this PR set out to fix. The fix belongs in the consumer path after the snapshot file is confirmed written, not at this call site.

- Checked and clear: the `*.{sql,sql.gz}` glob matches both extensions correctly, the descending last-modified-time sort would not delete the newest file, and the `.endsWith`-based compression detection is consistent between `createDataSnapshot` and `restoreFromSnapshot` (`H2SqlStatements.java`).

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
itsvishalyadav: Hi Paolo! I wanted to double-check my understanding here because the code structure can definitely be a bit tricky to follow at a glance. I believe the file is actually guaranteed to be on disk when the cleanup runs.

The actual snapshot generation (H2 SCRIPT TO) isn't triggered by writing to the snapshots topic (line 1286). It's triggered via the CreateSnapshot1Message sent to the journal topic earlier in the method (line 1279).

Here is the sequence in triggerSnapshotCreation():


```java
var uuid = blockOnResult(submitter.submitMessage(message)); // Submits to Journal topic
String snapshotLocation = (String) coordinator.waitForResponse(uuid); // Blocks until consumer finishes
```

The HTTP thread blocks on coordinator.waitForResponse(uuid). Meanwhile, on this node's background consumer thread, processRecord receives the message, verifies lastTriggeredSnapshot == snapshotId, and delegates to kafkaSqlSink.processMessage(record). This synchronously executes the H2 SCRIPT TO statement to disk. Once the SQL statement finishes, the consumer thread calls notifyResponse(uuid) to wake up the HTTP thread.

So by the time coordinator.waitForResponse(uuid) unblocks and we reach cleanupOldSnapshots(path), the local dump file has already been completely written to disk locally. The snapshotsProducer.apply(record) line just publishes the marker to the separate snapshots topic for new nodes booting up.

Because the HTTP thread waits for the consumer latch, the Files.exists check should correctly return true. Let me know if I'm missing an edge case here or if you'd still prefer it to be moved to the consumer thread directly!

itsvishalyadav:
Hi @paoloantinori, just a quick gentle ping on this! Let me know if my explanation above makes sense, or if you still see a gap in the logic and would prefer me to move the cleanup to the consumer thread. Happy to change it however you think is best. Thanks!

paoloantinori:
Your instinct is right — the cleanup should move to the consumer thread, not stay in `triggerSnapshotCreation()`.

The reason is the ordering of when the snapshot file actually materializes. `triggerSnapshotCreation()` only builds the `ProducerRecord` and publishes it to the snapshots topic; the real write (`SCRIPT TO <path> COMPRESSION GZIP`) happens later, on the consumer side, inside `SqlExportRepository.createSnapshot(location)` when the snapshot record is consumed. That means at the moment `triggerSnapshotCreation()` returns, `newSnapshotPath` does not exist on disk yet — so the `Files.exists(newSnapshotPath)` gate in `cleanupOldSnapshots()` is essentially always false here, and the cleanup runs as dead code. The retention-by-`getLastModifiedTime` logic itself is sound; it's just running on the wrong thread, before there's anything new to retain against.

Concretely, the cleanup wants to run *after* `createSnapshot(location)` has confirmed the new dump is on disk — i.e. on the consumer thread that processes the snapshot record, right after the `SCRIPT TO` succeeds. That gives you the existence guarantee you're currently trying (and failing) to get from the trigger side, and it's also the natural place to enforce "keep the 3 newest" because the new file is one of those 3.

One thing to double-check once it moves: on the consumer side, make sure the cleanup is scoped so that two replicas consuming the same snapshot record on a shared volume don't both try to delete the same set — the sort-and-retain-3 is idempotent for a single writer, but if multiple consumers run it concurrently against a shared directory you can get a race between `newDirectoryStream` listing and `deleteIfExists`. A `synchronized` block or a per-directory lock around list+delete would close that.

itsvishalyadav:
Thanks for the clarification @paoloantinori , that makes perfect sense! I have moved the cleanup logic into the consumer thread loop (processRecord) so it reliably runs right after the CreateSnapshot1Message is processed and we know the file is confirmed to be on disk.

I also added the synchronized modifier to the cleanupOldSnapshots method to prevent the exact list+delete race condition you mentioned! (I also just pushed a quick follow-up commit to fix a minor SonarQube duplication hiccup where I accidentally left the old copy of the method in the file).

The commits are pushed. Let me know if everything looks good now!

paoloantinori:
Thanks @itsvishalyadav — confirmed on the current head. Moving the cleanup into `processRecord` and gating it on `CreateSnapshot1Message` is the right fix: the new snapshot file is guaranteed to be on disk by the time the cleanup runs, so the retention logic (keep 3 newest via `getLastModifiedTime`) operates on a consistent view. The `synchronized` on `cleanupOldSnapshots` closes the list-then-delete race I raised earlier. The GZIP wiring (write via `SCRIPT TO ... COMPRESSION GZIP`, matching `RUNSCRIPT ... COMPRESSION GZIP` restore) and the `COMPRESSED_SNAPSHOT_EXTENSION` constant dispatch are unchanged and still look good. Nothing further from my side on the code.

EricWittmann:
Hi @itsvishalyadav — the build is failing. Could you take a look at the errors and push a fix? Thanks!

itsvishalyadav:
Hi @EricWittmann, thanks for the heads up! It was just a couple of unused imports left over in KafkaSqlRegistryStorage.java after I refactored the cleanup method in the previous commit, which caused Checkstyle to fail.

I've just removed them and pushed the fix. The build should be green now!

EricWittmann:
@itsvishalyadav — the build is failing due to an unused import in `KafkaSqlRegistryStorage.java` (line 20):

```
Unused import - io.apicurio.registry.storage.metrics.StorageMetricsEvent
```

Please remove that import and push the fix.

Also, for future PRs, please run the build locally before opening:

```bash
```

This would have caught the unused import before CI. Thanks!

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
Joel-hanson: @itsvishalyadav Looking at this a bit more, I don't think keep-3 by mtime fully closes the shared-directory case.

Say two instances write into the same snapshot location and each creates 3 dumps:

```
Node A
10:00  A1.sql.gz
10:01  A2.sql.gz
10:02  A3.sql.gz

Node B
10:03  B1.sql.gz
10:04  B2.sql.gz
10:05  B3.sql.gz

10:06  cleanup runs which keeps newest 3
```

After cleanup, A's dumps are gone, because they're older. So on a shared volume, whoever wrote most recently will be present, and older snapshots from the other one can get deleted. Correct me If my understanding is wrong here.

[#9275](https://github.com/Apicurio/apicurio-registry/pull/9275) helps for scheduled snapshots, but I think manual `POST /admin/snapshots` can still hit this. Same if people point `apicurio.storage.snapshot.location` at a shared folder that isn't meant to be exclusive to one Registry.

I don't think we need a code change for this, documenting that the snapshot directory should be specific to this (ignore if its already documented), and only used for Apicurio snapshots, should be enough.

EricWittmann:
@itsvishalyadav You have a few build errors.  I think the CI failures are actual failures from your changes.  Take a look. :+1:

EricWittmann:
Possible stale PR here.  I opened this one to make progress:
* https://github.com/Apicurio/apicurio-registry/pull/9510

EricWittmann:
Closing as superseded by:
* https://github.com/Apicurio/apicurio-registry/pull/9510

PR #9072, opened 2026-07-30
fix(references): extract embedded schemas on every artifact version, not just the first

### LINKED ISSUE #9053: PROMPT_TEMPLATE embedded outputSchema is not updated on new artifact versions
## Description

When a `PROMPT_TEMPLATE` artifact is created with an embedded `outputSchema`, the registry correctly auto-registers the schema as a separate JSON artifact, replaces the inline schema with a `$ref` in the stored content, and creates the corresponding artifact reference.

However, when a **new version** of the same `PROMPT_TEMPLATE` is created with a modified `outputSchema`, the auto-registration is skipped. The new artifact version retains the raw embedded schema instead of a `$ref`, and the previously generated schema artifact is never updated to reflect the new schema.

This behavior results in three observable problems during the lifecycle of a versioned prompt template:

- **Inconsistent content format across versions** – Version 1 stores the `outputSchema` as a `$ref` (for example, `default:mytemplate-output-schema:1`), while subsequent versions store the full embedded schema object inline. Consumers retrieving different versions of the same artifact receive different content structures.
- **Stale generated schema artifact** – The auto-generated schema artifact remains permanently at its initial version, even when later prompt template versions define a different `outputSchema`. Consumers resolving the generated schema artifact receive outdated validation rules.
- **No indication to the API caller** – The operation completes successfully (`200 OK`) and no error is returned to the client. The only indication is an INFO-level log similar to:

```text
Embedded schema artifact already exists: .../..., skipping auto-registration
```

---

## Environment

- **Registry Version:** `latest-snapshot` (`quay.io/apicurio/apicurio-registry:latest-snapshot`, pulled 2026-07-30)
- **Storage:** In-memory SQL (default)

---

### 1. Create a `PROMPT_TEMPLATE` with an embedded `outputSchema`

```bash
curl -X POST http://localhost:8080/apis/registry/v3/groups/default/artifacts \
  -d '{
    "artifactId": "test-prompt-staleness",
    "artifactType": "PROMPT_TEMPLATE",
    "firstVersion": {
      "version": "1",
      "content": {
        "content": "{\"templateId\":\"test\",\"name\":\"Test\",\"template\":\"Hello {{user}}\",\"variables\":{\"user\":{\"type\":\"string\",\"required\":true}},\"outputSchema\":{\"type\":\"object\",\"properties\":{\"greeting\":{\"type\":\"string\"},\"timestamp\":{\"type\":\"string\",\"format\":\"date-time\"}},\"required\":[\"greeting\"]}}",
        "contentType": "application/json"
      }
    }
  }'
```

### 2. Verify version 1 content was rewritten with a `$ref`

```bash
curl http://localhost:8080/apis/registry/v3/groups/default/artifacts/test-prompt-staleness/versions/1/content
```

Confirm the stored content contains:

```json
"outputSchema": {
  "$ref": "default:test-prompt-staleness-output-schema:1"
}
```

### 3. Verify the generated schema artifact exists

```bash
curl http://localhost:8080/apis/registry/v3/groups/default/artifacts/test-prompt-staleness-output-schema/versions/1/content
```

Confirm it contains the version 1 schema (`greeting` and `timestamp`).

### 4. Create version 2 with a modified `outputSchema`

```bash
curl -X POST http://localhost:8080/apis/registry/v3/groups/default/artifacts/test-prompt-staleness/versions \
  -d '{
    "version": "2",
    "content": {
      "content": "{\"templateId\":\"test\",\"name\":\"Test v2\",\"template\":\"Hello {{user}}, session {{sid}}\",\"variables\":{\"user\":{\"type\":\"string\",\"required\":true},\"sid\":{\"type\":\"string\"}},\"outputSchema\":{\"type\":\"object\",\"properties\":{\"greeting\":{\"type\":\"string\"},\"session_info\":{\"type\":\"object\",\"properties\":{\"id\":{\"type\":\"string\"}}}},\"required\":[\"greeting\",\"session_info\"]}}",
      "contentType": "application/json"
    }
  }'
```

### 5. Fetch version 2 content

```bash
curl http://localhost:8080/apis/registry/v3/groups/default/artifacts/test-prompt-staleness/versions/2/content
```

### 6. Fetch the generated schema artifact again

```bash
curl http://localhost:8080/apis/registry/v3/groups/default/artifacts/test-prompt-staleness-output-schema/versions/1/content
```

### 7. Compare outbound references

```bash
curl http://localhost:8080/apis/registry/v3/groups/default/artifacts/test-prompt-staleness/versions/1/references

curl http://localhost:8080/apis/registry/v3/groups/default/artifacts/test-prompt-staleness/versions/2/references
```

---

## Expected Behavior

- The embedded `outputSchema` should be extracted for every artifact version.
- The stored prompt template content should consistently reference the generated schema using a `$ref`.
- The generated schema artifact should reflect the corresponding schema for each prompt template version.
- A corresponding outbound artifact reference should be created for every version.
- Existing references from earlier versions should continue to resolve independently.

---

## Actual Behavior

| Aspect | Version 1 | Version 2 |
| :----- | :-------- | :-------- |
| `outputSchema` in stored content | Replaced with `$ref` (`default:test-prompt-staleness-output-schema:1`) | Raw embedded JSON schema remains inline |
| Generated schema artifact | Created and contains the extracted schema | Not updated; still contains only the version 1 schema |
| Outbound references | 1 reference to the generated schema artifact | No outbound references |

Additionally:

- The generated schema artifact (`test-prompt-staleness-output-schema`) still contains the original `timestamp` property.
- The new `session_info` property introduced in version 2 is absent.
- No second version of the generated schema artifact is created.

During version 2 creation, the server logs:

```text
INFO  Embedded schema artifact already exists: null/test-prompt-staleness-output-schema, skipping auto-registration
```

The API request completes successfully (`200 OK`), and no indication of the skipped registration is returned to the client.

---

## Investigation

Based on source code inspection, the behavior appears to originate in `EmbeddedSchemaService.autoRegisterSchema()`.

When the generated schema artifact already exists, registration is skipped, which appears to prevent embedded schema extraction from completing for subsequent artifact versions. This explanation is consistent with the runtime behavior observed during reproduction.

Anuragkumar-687:
Fixes #9053

## Root cause

`EmbeddedSchemaService.autoRegisterSchema()` only ever *created* the generated schema artifact. It called `storage.createArtifact()` with a hardcoded version `"1"` and had no path for handling an artifact that already existed.

On the first version of a `PROMPT_TEMPLATE` this worked: the embedded `outputSchema` was registered as a separate JSON artifact and replaced with a `$ref`. On every subsequent version the call hit `ArtifactAlreadyExistsException`, which was caught and turned into `return false`. That return value signaled that extraction had failed, so the remainder of the extraction process was skipped.

As a result:

- the new artifact version kept its **raw embedded schema** instead of a `$ref`, leaving the same artifact with two different content shapes across versions
- the generated schema artifact was **never updated**, remaining frozen at the first version's schema
- **no outbound reference** was recorded for the new artifact version

The request still completed successfully, so the only visible indication was an INFO log entry.

## The fix

`autoRegisterSchema()` now returns the version it registered (or `null` on failure) instead of a boolean.

When the generated schema artifact already exists, it first looks for an existing version with identical content using `storage.getArtifactVersionMetaDataByContent()`. If one exists, that version is reused; otherwise a new schema artifact version is created.

The callers now use the returned version when constructing both the `$ref` and the `ArtifactReferenceDto`, replacing the previously hardcoded `"1"`.

This lookup-then-create flow follows the existing content-deduplication pattern already used in `handleIfExistsReturnOrUpdate` (v2 and v3) and the ccompat `registerSchemaUnderSubject` implementation. Reusing identical schema versions also avoids creating redundant schema versions when the embedded schema is unchanged across parent artifact versions.

Existing references remain unaffected because reference resolution is version-pinned. A `$ref` pointing to version `1` continues to resolve to version `1` even after newer schema versions are created.

## MODEL_SCHEMA

`autoRegisterSchema()` is shared by both `PROMPT_TEMPLATE` and `MODEL_SCHEMA`, so this change also fixes embedded `input` and `output` schema extraction for `MODEL_SCHEMA`.

Although the reported issue focused on `PROMPT_TEMPLATE`, both artifact types shared the same implementation and therefore the same defect. Test coverage has been added for both.

## Testing

Added a new `EmbeddedSchemaExtractionTest` covering:

- a changed `outputSchema` on a new artifact version creates a new generated schema version, replaces the embedded schema with a `$ref`, and records the correct outbound reference
- an unchanged `outputSchema` reuses the existing generated schema version instead of creating a redundant one
- `MODEL_SCHEMA` embedded `input` and `output` schema extraction, verifying reuse for unchanged schemas and version creation for changed schemas

Existing `PromptRenderTest` continues to pass, and `./mvnw checkstyle:check -pl app` completes successfully.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/EmbeddedSchemaService.java
+++ b/app/src/main/java/io/apicurio/registry/services/EmbeddedSchemaService.java
@@ -238,6 +238,8 @@ public class EmbeddedSchemaService {
                     return existing.getVersion();
                 } catch (ArtifactNotFoundException nfe) {
                     // No version with matching content - create a new one.
+                    log.debug("No existing version with matching content for {}/{}, creating new version",
+                            groupId, schemaArtifactId);
                 }
                 ArtifactVersionMetaDataDto vmd = storage.createArtifactVersion(groupId, schemaArtifactId,
                         null, "JSON", contentWrapper, versionMeta, Collections.emptyList(),
--- a/app/src/main/java/io/apicurio/registry/services/EmbeddedSchemaService.java
+++ b/app/src/main/java/io/apicurio/registry/services/EmbeddedSchemaService.java
@@ -4,12 +4,15 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.apicurio.registry.content.ContentHandle;
+import io.apicurio.registry.content.TypedContent;
 import io.apicurio.registry.storage.RegistryStorage;
 import io.apicurio.registry.storage.dto.ArtifactReferenceDto;
+import io.apicurio.registry.storage.dto.ArtifactVersionMetaDataDto;
 import io.apicurio.registry.storage.dto.ContentWrapperDto;
 import io.apicurio.registry.storage.dto.EditableArtifactMetaDataDto;
 import io.apicurio.registry.storage.dto.EditableVersionMetaDataDto;
 import io.apicurio.registry.storage.error.ArtifactAlreadyExistsException;
+import io.apicurio.registry.storage.error.ArtifactNotFoundException;
 import io.apicurio.registry.types.ContentTypes;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
@@ -95,12 +98,12 @@ public class EmbeddedSchemaService {
             // Extract "input" schema
             if (rootObj.has("input") && rootObj.get("input").isObject() && hasSchemaProperties(rootObj.get("input"))) {
                 String schemaArtifactId = artifactId + "-input-schema";
-                String version = "1";
-                String refName = buildReferenceName(groupId, schemaArtifactId, version);
                 JsonNode inputSchema = rootObj.get("input");
 
-                if (autoRegisterSchema(storage, groupId, schemaArtifactId, inputSchema,
-                        "Input schema for " + artifactId, owner)) {
+                String version = autoRegisterSchema(storage, groupId, schemaArtifactId, inputSchema,
+                        "Input schema for " + artifactId, owner);
+                if (version != null) {
+                    String refName = buildReferenceName(groupId, schemaArtifactId, version);
                     rootObj.set("input", createRefNode(refName));
                     references.add(ArtifactReferenceDto.builder()
                             .groupId(groupId).artifactId(schemaArtifactId).version(version).name(refName)
@@ -112,12 +115,12 @@ public class EmbeddedSchemaService {
             // Extract "output" schema
             if (rootObj.has("output") && rootObj.get("output").isObject() && hasSchemaProperties(rootObj.get("output"))) {
                 String schemaArtifactId = artifactId + "-output-schema";
-                String version = "1";
-                String refName = buildReferenceName(groupId, schemaArtifactId, version);
                 JsonNode outputSchema = rootObj.get("output");
 
-                if (autoRegisterSchema(storage, groupId, schemaArtifactId, outputSchema,
-                        "Output schema for " + artifactId, owner)) {
+                String version = autoRegisterSchema(storage, groupId, schemaArtifactId, outputSchema,
+                        "Output schema for " + artifactId, owner);
+                if (version != null) {
+                    String refName = buildReferenceName(groupId, schemaArtifactId, version);
                     rootObj.set("output", createRefNode(refName));
                     references.add(ArtifactReferenceDto.builder()
                             .groupId(groupId).artifactId(schemaArtifactId).version(version).name(refName)
@@ -162,12 +165,12 @@ public class EmbeddedSchemaService {
             // Extract "outputSchema"
             if (rootObj.has("outputSchema") && rootObj.get("outputSchema").isObject() && hasSchemaProperties(rootObj.get("outputSchema"))) {
                 String schemaArtifactId = artifactId + "-output-schema";
-                String version = "1";
-                String refName = buildReferenceName(groupId, schemaArtifactId, version);
                 JsonNode outputSchema = rootObj.get("outputSchema");
 
-                if (autoRegisterSchema(storage, groupId, schemaArtifactId, outputSchema,
-                        "Output schema for prompt template " + artifactId, owner)) {
+                String version = autoRegisterSchema(storage, groupId, schemaArtifactId, outputSchema,
+                        "Output schema for prompt template " + artifactId, owner);
+                if (version != null) {
+                    String refName = buildReferenceName(groupId, schemaArtifactId, version);
                     rootObj.set("outputSchema", createRefNode(refName));
                     references.add(ArtifactReferenceDto.builder()
                             .groupId(groupId).artifactId(schemaArtifactId).version(version).name(refName)
@@ -193,11 +196,13 @@ public class EmbeddedSchemaService {
     }
 
     /**
-     * Auto-register a JSON Schema as a separate JSON artifact.
+     * Auto-register a JSON Schema as a separate JSON artifact. If the schema artifact already
+     * exists, an existing version with identical content is reused, otherwise a new version
+     * is created.
      *
-     * @return true if the schema was successfully registered, false otherwise
+     * @return the version of the registered schema, or null if registration failed
      */
-    private boolean autoRegisterSchema(RegistryStorage storage, String groupId, String schemaArtifactId,
+    private String autoRegisterSchema(RegistryStorage storage, String groupId, String schemaArtifactId,
             JsonNode schema, String description, String owner) {
         try {
             String schemaContent = MAPPER.writerWithDefaultPrettyPrinter().writeValueAsString(schema);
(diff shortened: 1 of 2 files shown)


paoloantinori (commented):
Overall this is a clean, well-scoped fix for #9053: `autoRegisterSchema` now returns the registered version instead of a boolean, and a lookup-by-content-then-create flow lets embedded schemas extract on every parent version rather than only the first. The logic is sound, failure still degrades gracefully, and the new tests cover the changed-schema, unchanged-schema, and MODEL_SCHEMA cases. I did not find any blocker- or major-level issues.

- The dedup-then-create pattern in `autoRegisterSchema` (`EmbeddedSchemaService.java`, the `+222..+248` hunk) mirrors the existing content-deduplication approach used in `handleIfExistsReturnOrUpdate`/`registerSchemaUnderSubject`, and correctly reuses an identical schema version instead of creating a redundant one.
- All three call sites (`input` ~+98, `output` ~+115, `outputSchema` ~+165) compute `version` once and feed the same value into both `buildReferenceName`/`createRefNode` (the `$ref`) and `.version(version)` on the `ArtifactReferenceDto`, so the `$ref` and outbound reference cannot drift — consistent with version-pinned resolution.
- The graceful-failure contract is preserved: the outer `try/catch (Exception) ... return null` wraps the entire method body, including the new create/dedup paths, so any failure (lookup NPE, `createArtifactVersion` throw, etc.) leaves the parent with its raw inline schema rather than crashing — the same end state as before the PR.
- The new `EmbeddedSchemaExtractionTest` asserts the `$ref`/reference version, the generated schema artifact's version count, and the schema content for the changed-schema (new version), unchanged-schema (reuse), and MODEL_SCHEMA input/output scenarios, which directly guard the three return paths of `autoRegisterSchema`.
- One thing I could not verify from the diff alone (no `RegistryStorage` source was available in this read-only review sandbox): the changed-schema path assumes `getArtifactVersionMetaDataByContent` throws `ArtifactNotFoundException` rather than returning null on no match. The explicit `catch (ArtifactNotFoundException nfe)` implies that contract and the new test exercises exactly this path, but it's worth a quick confirm against the storage interface. In the worst case (null return) `existing.getVersion()` would NPE, which is swallowed to a graceful `null` — no crash and no wrong version pinned, just extraction skipped.

Anuragkumar-687:
Thanks for the detailed review @paoloantinori!

I confirmed that getArtifactVersionMetaDataByContent(...) throws ArtifactNotFoundException when no matching content exists rather than returning null.

SqlVersionRepository.getArtifactVersionMetaDataByContent ends with:

Optional<ArtifactVersionMetaDataDto> res = handle.createQuery(sql)
        .bind(0, normalizeGroupId(groupId))
        .bind(1, artifactId)
        .bind(2, hash)
        .map(ArtifactVersionMetaDataDtoMapper.instance)
        .findFirst();
return res.orElseThrow(() -> new ArtifactNotFoundException(groupId, artifactId));

AbstractSqlRegistryStorage delegates directly to this method, and RegistryStorage documents ArtifactNotFoundException as the no-match contract, so the catch (ArtifactNotFoundException) path is the intended behavior.

The changed-schema path is also exercised by testPromptTemplateOutputSchemaExtractedOnNewVersion: the updated outputSchema causes the content lookup to miss, a new generated schema version is created, and the test verifies that the $ref points to version 2 and that the generated schema artifact has two versions.

paoloantinori:
Thanks for the follow-up, @Anuragkumar-687 — and yes, that matches what's on this head. `getArtifactVersionMetaDataByContent(...)` throws `ArtifactNotFoundException` when no matching content exists (it returns `Optional`/empty internally and the SQL mapper surfaces that as the exception), so the `try { ... } catch (ArtifactNotFoundException nfe) { /* create new version */ }` branch is exactly the right control flow: reuse-on-match, create-on-miss. That was the one behavioural question worth pinning down, and the test (`EmbeddedSchemaExtractionTest`) exercises both the reuse and the new-version paths against a real storage impl.

Nothing further from my side — the change looks correct and well-covered.

Anuragkumar-687:
Thanks, @paoloantinori! I really appreciate the detailed review and the follow-up verification. Glad we could confirm the behavior and that the implementation looks good.

paoloantinori:
You're welcome, @Anuragkumar-687! Glad the back-and-forth helped confirm the behavior. The fix is in good shape — nothing else needed from my side here.

Anuragkumar-687:
Hi @EricWittmann, @jsenko! Just a quick follow-up: Paolo has completed his review, and the follow-up technical question has been resolved. All checks are passing now. Whenever you have a chance, I'd really appreciate a final review. This PR is currently blocking me from opening my next assigned PR due to the one-PR-at-a-time contribution policy. Thanks!

carlesarnal:
Reviewed the full diff (1 commit, 2 files). No issues found. Details below.

**Commit hygiene** - single commit, conventional message, DCO sign-off present.

**autoRegisterSchema rework** (EmbeddedSchemaService.java:222-254) - returns version string instead of boolean. Try createArtifact first; on ArtifactAlreadyExistsException, look up existing version with identical content (reuse-on-match) or createArtifactVersion (create-on-miss). Outer catch returns null preserving graceful degradation. Mirrors existing dedup patterns.

**Call sites** (input ~L98, output ~L115, outputSchema ~L165) - all feed returned version into both buildReferenceName/createRefNode and .version() on ArtifactReferenceDto, so ref and $ref cannot drift.

**Test coverage** (EmbeddedSchemaExtractionTest.java) - 3 tests covering all return paths: changed schema (new version created, $ref points to v2), unchanged schema (reuses v1), and MODEL_SCHEMA with mixed reuse/create. Assertions check exact $ref strings, version numbers, version counts, and schema properties.

**Style** - no star imports, no unused imports, K&R braces, no leaked stack traces.

Looks good. Nothing to add beyond what Paolo already confirmed.

Anuragkumar-687:
Thanks @carlesarnal for the thorough review and for going through the full diff! I really appreciate the time you took to verify the implementation. Glad to hear everything looks good.

jsenko (commented):
Review on head `3b5a5ae`. Well-scoped fix — changing `autoRegisterSchema` from boolean to version-returning, with content dedup via `getArtifactVersionMetaDataByContent`, is the right approach. The three test scenarios (changed schema, unchanged schema, MODEL_SCHEMA) cover the key paths. No correctness bugs found.

**1. Missing Apache 2.0 license header (should-fix)**

`EmbeddedSchemaExtractionTest.java` is a new file but has no license header. All new Java files in this project require the Apache 2.0 header block. See existing test files for the pattern.

**2. Empty catch block (should-fix)**

`EmbeddedSchemaService.java` ~line 237:
```java
} catch (ArtifactNotFoundException nfe) {
    // No version with matching content - create a new one.
}
```

The comment is there, which is good, but adding a `log.debug(...)` would make this more observable:
```java
} catch (ArtifactNotFoundException nfe) {
    log.debug("No existing version with matching content for {}/{}, creating new version",
            groupId, schemaArtifactId);
}
```

This is control-flow-via-exception, which is the established pattern in this codebase. A debug log would help with troubleshooting without being noisy.

Anuragkumar-687:
Thanks for the review, @jsenko!

I've addressed both suggestions:

Added the standard Apache 2.0 license header to EmbeddedSchemaExtractionTest.
Added a log.debug(...) message in the ArtifactNotFoundException catch block to make the create-on-miss path observable.

I also reran the relevant tests and checkstyle locally. Thanks again for taking the time to review this.

Anuragkumar-687:
Hi @EricWittmann, just a gentle follow-up. All review feedback has been addressed and all checks are passing. If everything looks good, could you please merge this PR when you have a chance?

It's been open for the last 5 days, and it's currently blocking me from opening my next PR due to the project's one-PR-at-a-time contribution policy.

jsenko (approved):
LGTM

Anuragkumar-687:
Thanks @jsenko! I checked the failing workflow, and it looks like the failure occurred during npm install due to an ETIMEDOUT network error in the UI build job, before any of the build or test steps ran. Since this PR only changes Java code, it appears unrelated to the changes in this PR. Would you mind retrying the workflow when you have a chance? Thanks!

Anuragkumar-687:
Hi @jsenko, just a gentle follow-up. The remaining failure appears to be a transient `npm install` ETIMEDOUT issue unrelated to this Java-only PR. Would you mind running `/retry` when you have a chance? Thanks!

jsenko:
Merged, thanks!

PR #9075, opened 2026-07-30
fix(search): use Locale.ROOT for state filter uppercasing in Elasticsearch query

### LINKED ISSUE #9073: bug(search): Elasticsearch state filter uses locale-sensitive toUpperCase(), breaking search under non-default JVM locales
## Description

**Persistence type**: sql / kafkasql (with Elasticsearch search enabled)

`ElasticsearchSearchService.buildFilterQuery()` builds the `state` filter's Elasticsearch term query like this:

```java
case state:
    return Query.of(q -> q.term(t -> t
            .field("state").value(filter.getStringValue().toUpperCase())));
```

`toUpperCase()` is called without a `Locale` argument, which violates this project's own coding convention (`.toUpperCase()` / `.toLowerCase()` must always take `Locale.ROOT`) and relies on the JVM's default locale for case conversion.

Today, the only caller (`SearchFilter.ofState(VersionState)`) always passes an already fully-uppercase ASCII `VersionState.name()` value (`ENABLED`, `DISABLED`, `DEPRECATED`, `DRAFT`), so this doesn't currently produce a visible symptom — I verified that uppercasing those specific values under `tr-TR` and under `Locale.ROOT` produces identical output, since none of them contain a lowercase `i`.

However, the call is still latent risk: if any future or alternate code path ever supplies a `state` filter value that isn't already uppercase, running the registry under a Turkish-family JVM default locale would uppercase a lowercase `i` to the dotted `İ` instead of plain `I`, which would not match the `state` field (indexed as an exact-match `keyword` populated from the enum's plain ASCII name), silently returning zero results.

### Environment

Any environment; most relevant if the JVM default locale is ever Turkish (`tr`/`tr-TR`, e.g. via `-Duser.language=tr -Duser.country=TR`).

### Steps to Reproduce

Not currently reproducible through the REST API's `state` query parameter, since it's typed as the `VersionState` enum and always normalizes to an already-uppercase value before reaching this code. This is a defensive/code-style fix to close off a locale-dependent construct in the codebase, consistent with the project's Locale convention, rather than a report of an observed failure.

**Expected**: string case conversion in this method should be locale-independent (`Locale.ROOT`), matching the project's coding convention, regardless of whether current call sites happen to avoid triggering the Turkish-locale edge case.
**Actual**: `toUpperCase()` is called with the JVM default locale.

### Logs

N/A — no exception or observable failure with current call sites; this addresses a coding-convention violation and closes off a latent locale-dependent risk.

bhuvan-somisetty:
## What

`ElasticsearchSearchService.buildFilterQuery()` calls `toUpperCase()` on the `state` filter's string value without a `Locale` argument, relying on the JVM default locale for case conversion. This violates the project's convention that all case conversions go through `Locale.ROOT`.

## Why

The only current caller (`SearchFilter.ofState(VersionState)`) always passes an already-uppercase ASCII value (`ENABLED`, `DISABLED`, `DEPRECATED`, `DRAFT`), so there's no observable failure today. But under a Turkish-family default locale (`tr`/`tr-TR`), an uppercase conversion on a lowercase `i` produces the dotted `İ` instead of plain `I`. If any future or alternate code path ever passes a value that isn't already uppercase, the resulting term query would silently fail to match the `state` keyword field and return zero results, with no exception or log line to point at the cause.

Fixes #9073

## Changes

- `ElasticsearchSearchService.buildFilterQuery()`: `filter.getStringValue().toUpperCase()` → `filter.getStringValue().toUpperCase(Locale.ROOT)`
- Added `ElasticsearchSearchServiceTest`, which sets the default locale to `tr-TR`, builds a state filter from a lowercase value, and asserts the resulting term query value is `DISABLED` (not the Turkish dotted variant).

## Test plan

- `./mvnw test -pl app -Dtest=ElasticsearchSearchServiceTest` — passes
- `./mvnw checkstyle:check -pl app` — 0 violations
- `./mvnw install -pl app -am -DskipTests` — builds clean

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
@@ -212,7 +213,7 @@ public class ElasticsearchSearchService {
 
         case state:
             return Query.of(q -> q.term(t -> t
-                    .field("state").value(filter.getStringValue().toUpperCase())));
+                    .field("state").value(filter.getStringValue().toUpperCase(Locale.ROOT))));
 
         case globalId:
             return Query.of(q -> q.term(t -> t
--- a/app/src/test/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchServiceTest.java
+++ b/app/src/test/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchServiceTest.java
@@ -4,6 +4,8 @@ import co.elastic.clients.elasticsearch._types.query_dsl.Query;
 import io.apicurio.registry.storage.dto.SearchFilter;
 import io.apicurio.registry.storage.dto.SearchFilterType;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
 
 import java.util.Locale;
 import java.util.Set;
@@ -13,6 +15,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 public class ElasticsearchSearchServiceTest {
 
     @Test
+    @ResourceLock(value = "jvm.defaultLocale", mode = ResourceAccessMode.READ_WRITE)
     void stateFilterUppercasesUnderTurkishDefaultLocale() {
         Locale originalLocale = Locale.getDefault();
         Locale.setDefault(Locale.forLanguageTag("tr-TR"));
--- a/app/src/test/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchServiceTest.java
+++ b/app/src/test/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchServiceTest.java
@@ -1,0 +1,34 @@
+package io.apicurio.registry.storage.impl.search;
+
+import co.elastic.clients.elasticsearch._types.query_dsl.Query;
+import io.apicurio.registry.storage.dto.SearchFilter;
+import io.apicurio.registry.storage.dto.SearchFilterType;
+import org.junit.jupiter.api.Test;
+
+import java.util.Locale;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ElasticsearchSearchServiceTest {
+
+    @Test
+    void stateFilterUppercasesUnderTurkishDefaultLocale() {
+        Locale originalLocale = Locale.getDefault();
+        Locale.setDefault(Locale.forLanguageTag("tr-TR"));
+        try {
+            ElasticsearchSearchService service = new ElasticsearchSearchService();
+
+            SearchFilter filter = new SearchFilter();
+            filter.setType(SearchFilterType.state);
+            filter.setStringValue("disabled");
+
+            Query query = service.buildEsQuery(Set.of(filter));
+
+            String termValue = query.bool().must().get(0).term().value().stringValue();
+            assertEquals("DISABLED", termValue);
+        } finally {
+            Locale.setDefault(originalLocale);
+        }
+    }
+}


bhuvan-somisetty:
@paoloantinori hey, whenever you get a chance could you take a look at this one? small locale fix, should be a quick review. thanks!

bhuvan-somisetty:
@EricWittmann could you take a look when you have a minute? the verification gate is showing failed but that's just because the PR hasn't been accepted yet, so tests haven't actually run. happy to address any feedback once it's triaged. thanks!

paoloantinori:
LGTM — this is the correct fix and a clean test.

**The fix is right.** `toUpperCase()` without a `Locale` delegates to the JVM default locale, and the Turkish-`i` case (`i` → `İ`, not `I`) is exactly the failure mode it can introduce. `Locale.ROOT` is the project convention and is correct here. No behavior change for the current caller (which already passes uppercase ASCII), so this is purely defensive — good.

**The test is solid.** Setting `Locale.setDefault(tr-TR)` around the call and asserting the term value is `DISABLED` (not the dotted variant) is the right way to pin this without spinning up ES. `new ElasticsearchSearchService()` is fine here because the `state` branch of `buildFilterQuery` doesn't touch any injected field — it's a pure transform of the filter. Two small things worth a glance, neither blocking:

- `buildEsQuery` is package-private and `buildFilterQuery` is private — the test reaches the private method indirectly via `buildEsQuery`, which is the right seam. If `buildEsQuery` ever grows a dependency on an injected field (e.g. a settings bean) before delegating, this test would break with an NPE rather than a clear failure. Acceptable risk for now; just flagging.
- The test mutates global JVM state (`Locale.setDefault`) even though it restores it in `finally`. Under parallel JUnit execution that's a process-wide mutation mid-test — other tests running concurrently that read the default locale could see `tr-TR`. JUnit 5 defaults to sequential per-class, so this is fine in practice, but if the suite ever moves to `@Execution(CONCURRENT)` at the class level, wrap this in a `ResourceLock` on `Locale` or use a JUnit `@Locale` extension (e.g. from `junit-pioneer`) to scope the change to the test thread.

Neither is a merge blocker from my side. Nice defensive fix.

bhuvan-somisetty:
@paoloantinori thanks for the thorough look! added a @ResourceLock on the test for the locale mutation since it was a cheap fix and closes off the parallel-execution concern for good. left buildEsQuery/buildFilterQuery access as-is since you confirmed it's the right seam. pushed, let me know if anything else stands out.

paoloantinori:
Thanks @bhuvan-somisetty — the `@ResourceLock("jvm.defaultLocale", mode = READ_WRITE)` is the right fix. It serializes the test against any other test that touches the default locale, so the `tr-TR` mutation can't bleed into a parallel neighbor even under JUnit's default concurrent execution. Leaving `buildEsQuery`/`buildFilterQuery` as the test seam is fine — it isolates the locale behavior without standing up Elasticsearch.

LGTM.

bhuvan-somisetty:
@paoloantinori appreciate you taking the time to dig into it, thanks! will wait on the accept/triage from here.

bhuvan-somisetty:
@EricWittmann Paolo's given this an LGTM above, could you accept/triage when you get a chance? thanks!

paoloantinori:
You're welcome @bhuvan-somisetty — happy to help. The fix and the `@ResourceLock` follow-up both look good on my end, so it's in @EricWittmann's hands now for the accept/triage.

bhuvan-somisetty:
Hey @EricWittmann, wanted to flag something all seven of the "Unit Tests (Scalpel)" shadow checks are failing on this PR, failing within 1-2 minutes across every shard including completely unrelated ones like app-gitops and non-app.

Looking at verify.yaml these are evaluation-mode shadow runs and not part of the Verification Gate, so I don't think they block the merge. The regular unit tests and the fix itself pass cleanly locally. Could you take a look and confirm if the Scalpel shadow run failures are a known infrastructure issue? Let me know if there's anything I need to do on my end.

EricWittmann:
> Hey @EricWittmann, wanted to flag something all seven of the "Unit Tests (Scalpel)" shadow checks are failing on this PR, failing within 1-2 minutes across every shard including completely unrelated ones like app-gitops and non-app.
> 
> Looking at verify.yaml these are evaluation-mode shadow runs and not part of the Verification Gate, so I don't think they block the merge. The regular unit tests and the fix itself pass cleanly locally. Could you take a look and confirm if the Scalpel shadow run failures are a known infrastructure issue? Let me know if there's anything I need to do on my end.

Yeah don't worry about scalpel for now.  I'm ignoring it until someone else looks into it.  :smile:

EricWittmann:
Thanks @bhuvan-somisetty! Clean fix with a solid test — nice attention to the `@ResourceLock` follow-up as well.

bhuvan-somisetty:
Thanks @EricWittmann!

PR #9077, opened 2026-07-30
fix(ui): hide version kebab menu when no actions are visible

### LINKED ISSUE #9076: Version details kebab menu opens with no visible actions for most artifact types (confusing empty-click UX)
## Summary

On the artifact **Version > Overview** page, the kebab ("more actions") 
button next to the "Download" button is always rendered, even when 
none of its menu items are visible. For most artifact types, clicking 
it opens a dropdown with nothing inside — which looks and feels 
exactly like the button is broken.

## Video

https://github.com/user-attachments/assets/1969b681-91fd-466d-bd18-69501b08aa37

## Steps to Reproduce

1. Create or open a version of an artifact whose type is **not** 
   OPENAPI or AGENT_CARD (e.g. PROMPT_TEMPLATE, ASYNCAPI, AVRO, JSON, 
   PROTOBUF, GRAPHQL, XSD, WSDL, KCONNECT, MODEL_SCHEMA, and others).
2. Go to the version's **Overview** tab.
3. Click the kebab (three-dot) icon next to the "Download" button.
4. Nothing visibly happens — the dropdown opens but is empty.

Compare with:
1. Open a version of an **OPENAPI** artifact → kebab shows 
   "Generate client SDK".
2. Open a version of an **AGENT_CARD** artifact (as the owner) → kebab 
   shows "Edit Agent Card".

## Root Cause

`VersionPageHeader.tsx` always renders the kebab `ObjectDropdown`, 
passing it a full list of possible actions. Each action item has its 
own `isVisible()` predicate (checked against artifact type, version 
state, feature flags, and ownership). Currently, only two actions have 
type-specific visibility conditions:

- **Generate client SDK** — visible only for `artifactType === "OPENAPI"`
- **Edit Agent Card** — visible only for `artifactType === "AGENT_CARD"` 
  (and requires developer ownership)

The remaining actions (Create draft, Finalize draft, Delete version) 
depend on feature flags and/or ownership, which are frequently off/
unmet in typical setups.

The result: for the majority of supported artifact types (Avro, JSON, 
AsyncAPI, Protobuf, GraphQL, XSD, WSDL, Kafka Connect, Prompt Template, 
Model Schema, and others), **every single action evaluates to invisible 
simultaneously**, so the kebab opens to a blank popover instead of 
showing at least one relevant action or hiding itself.

Git history confirms the dropdown was intended to always contain real 
actions — the commit that introduced it was titled *"Added missing 
actions to the Version details page"* — so an empty menu was never the 
intended end state.

## Why this matters (UX)

A clickable button that visibly does nothing reads as broken to users, 
even though the underlying logic is technically "working as coded." 
Since the majority of artifact types fall into this empty-menu case, 
most users interacting with the Version Overview page will encounter 
this at some point, not just an edge case.

## Verified Testing

Manually tested across the following artifact types:

| Type | Kebab has visible action(s)? |
|---|---|
| OPENAPI | ✅ Generate client SDK |
| AGENT_CARD | ✅ Edit Agent Card |
| PROMPT_TEMPLATE | ❌ Empty |
| ASYNCAPI | ❌ Empty |
| AVRO | ❌ Empty |
| JSON | ❌ Empty |
| PROTOBUF | ❌ Empty |
| GRAPHQL | ❌ Empty |
| XSD | ❌ Empty |
| WSDL | ❌ Empty |
| KCONNECT | ❌ Empty |
| MODEL_SCHEMA | ❌ Empty |

## Suggested Fix

Only render the kebab when at least one action is actually visible — 
compute the visible-actions list up front and conditionally render the 
`ActionListItem`/`ObjectDropdown` around it.

This fix is condition-agnostic — it doesn't special-case any artifact 
type. If a future action's `isVisible()` condition becomes true for 
any type (new feature flags, new type-specific actions, ownership 
changes), the kebab will automatically appear with the correct items, 
with no further changes needed here.

Faisal77666:
## Summary

Fixes #9076

On the artifact **Version > Overview** page, the kebab ("more actions") 
button was always rendered, even when every action's `isVisible()` 
check evaluated to false. For most artifact types, this meant clicking 
the kebab opened an empty popover — indistinguishable from a dead/
broken button.

## Root Cause

`VersionPageHeader.tsx` always rendered the kebab `ObjectDropdown` and 
passed it the full `actions` array, relying on `itemIsVisible` to 
filter items internally. Only two actions have type-specific visibility:

- **Generate client SDK** — visible only for `artifactType === "OPENAPI"`
- **Edit Agent Card** — visible only for `artifactType === "AGENT_CARD"`

For every other supported artifact type (Avro, JSON, AsyncAPI, 
Protobuf, GraphQL, XSD, WSDL, Kafka Connect, Prompt Template, Model 
Schema, and others), all actions evaluate to invisible simultaneously 
in typical setups, so the kebab opened to nothing.

## Fix

Compute `visibleActions` up front by filtering `actions` on 
`isVisible()`, and only render the kebab `ActionListItem`/
`ObjectDropdown` when `visibleActions.length > 0`. `items={actions}` 
is replaced with `items={visibleActions}`, and the now-redundant 
`itemIsVisible` prop on `ObjectDropdown` is removed.

This fix is condition-agnostic — it doesn't special-case any artifact 
type. If a future action's `isVisible()` condition becomes true for 
any type (new feature flags, new type-specific actions, ownership 
changes), the kebab will automatically appear with the correct items, 
with no further changes needed here.

## Testing

Verified no regression for the two types that currently have working 
actions, and confirmed the fix resolves the empty-menu issue for all 
other tested types:

| Type | Before | After |
|---|---|---|
| OPENAPI | ✅ Generate client SDK | ✅ Generate client SDK (unchanged) |
| AGENT_CARD | ✅ Edit Agent Card | ✅ Edit Agent Card (unchanged) |
| PROMPT_TEMPLATE | ❌ Empty popover | ✅ Kebab hidden |
| ASYNCAPI | ❌ Empty popover | ✅ Kebab hidden |
| AVRO, JSON, PROTOBUF, GRAPHQL, XSD, WSDL, KCONNECT, MODEL_SCHEMA | ❌ Empty popover | ✅ Kebab hidden |

## Video

**Before (bug):**

[raw-recording (9).webm](https://github.com/user-attachments/assets/2ccc9701-f3fb-46fa-9ac8-448a8f6c2455)

**After (fix applied):**

[raw-recording (10).webm](https://github.com/user-attachments/assets/877d689b-d941-467c-8f11-300b24913464)

### DIFF
--- a/ui/ui-app/src/app/pages/version/components/pageheader/VersionPageHeader.tsx
+++ b/ui/ui-app/src/app/pages/version/components/pageheader/VersionPageHeader.tsx
@@ -98,7 +98,7 @@ export const VersionPageHeader: FunctionComponent<VersionPageHeaderProps> = (pro
             }
         }
     ];
-    const visibleActions: any[] = actions.filter(action => action.isVisible());
+    const visibleActions = actions.filter(action => action.isVisible());
 
     return (
         <Flex className="example-border">
--- a/ui/ui-app/src/app/pages/version/components/pageheader/VersionPageHeader.tsx
+++ b/ui/ui-app/src/app/pages/version/components/pageheader/VersionPageHeader.tsx
@@ -98,6 +98,7 @@ export const VersionPageHeader: FunctionComponent<VersionPageHeaderProps> = (pro
             }
         }
     ];
+    const visibleActions: any[] = actions.filter(action => action.isVisible());
 
     return (
         <Flex className="example-border">
@@ -128,21 +129,22 @@ export const VersionPageHeader: FunctionComponent<VersionPageHeaderProps> = (pro
                             </IfFeature>
                         </IfFeature>
                     </ActionListItem>
-                    <ActionListItem key="actions">
-                        <ObjectDropdown
-                            label=""
-                            items={actions}
-                            onSelect={item => item.onSelect()}
-                            itemToString={item => item.label}
-                            itemToTestId={item => item.testId}
-                            itemIsVisible={item => item.isVisible()}
-                            itemIsDivider={item => item.divider}
-                            popperProps={{
-                                position: "right"
-                            }}
-                            isKebab={true}
-                        />
-                    </ActionListItem>
+                    <If condition={visibleActions.length > 0}>
+                        <ActionListItem key="actions">
+                            <ObjectDropdown
+                                label=""
+                                items={visibleActions}
+                                onSelect={item => item.onSelect()}
+                                itemToString={item => item.label}
+                                itemToTestId={item => item.testId}
+                                itemIsDivider={item => item.divider}
+                                popperProps={{
+                                    position: "right"
+                                }}
+                                isKebab={true}
+                            />
+                        </ActionListItem>
+                    </If>
                 </ActionList>
             </FlexItem>
         </Flex>


ui/ui-app/src/app/pages/version/components/pageheader/VersionPageHeader.tsx
```diff
@@ -98,6 +98,7 @@ export const VersionPageHeader: FunctionComponent<VersionPageHeaderProps> = (pro
             }
         }
     ];
+    const visibleActions: any[] = actions.filter(action => action.isVisible());
```
paoloantinori: Consider dropping the explicit `: any[]` annotation. `actions` already carries a concrete element type, so `actions.filter(...)` infers it automatically; widening the result to `any[]` silently disables type-checking on `item` in the dropdown handlers below (`onSelect`, `itemToString`, `itemToTestId`). Simpler and safer to let inference work: `const visibleActions = actions.filter(action => action.isVisible());`.

paoloantinori (commented):
Clean, well-scoped fix: computing `visibleActions` up front and gating the kebab on `visibleActions.length > 0` correctly eliminates the empty popover, and the change is condition-agnostic so it self-maintains as new `isVisible()` conditions appear. The core logic is sound and consistent with the reported test matrix (no regression for OPENAPI/AGENT_CARD), and dropping the now-redundant `itemIsVisible` prop is the right cleanup.

- The filter runs over the full `actions` array on every render for every artifact type, so if any action lacked an `isVisible()` method the page would crash for OPENAPI/AGENT_CARD too — the author's matrix shows those unchanged, which confirms all current actions implement `isVisible()` (VersionPageHeader.tsx:101). No live throw risk.
- One easy-win suggestion: the explicit `: any[]` annotation discards the element type of `actions` and silently disables type-checking on `item` in the dropdown handlers (`onSelect`, `itemToString`, `itemToTestId`) below (see inline comment). Letting inference handle it is strictly better.
- Forward-looking note (non-blocking): if a future `actions` entry is a bare divider object without `isVisible()`, this filter would throw; worth a one-line comment on the array if divider-only entries are ever added.

ui/ui-app/src/app/pages/version/components/pageheader/VersionPageHeader.tsx
Faisal77666: Good catch, fixed — removed the explicit any[] and let TypeScript infer the type. Pushed in the latest commit.

Faisal77666:
@EricWittmann please review pr if you have time

EricWittmann:
Thanks @Faisal77666! Clean fix — the condition-agnostic approach means this will just work as new actions are added in the future. Nice PR description with the test matrix and videos too.

PR #9058, opened 2026-07-30
docs: fix local-tier build and dev-mode commands (#9018)

Aarti-panchal01:
Relates to #9018.

The documented local-tier flow fails on a fresh clone: `./mvnw clean install -Dlocal -DskipTests` dies at app test compilation (the local tier excludes the serde/Maven-plugin modules, but the `app` module's test sources import their packages, and `-DskipTests` still compiles test sources). The README quick start then runs `../mvnw quarkus:dev` without `-Dlocal`, which tries to resolve the excluded test-scoped artifacts and fails on a fresh `.m2`.

This PR documents the verified working flow (`-Dmaven.test.skip=true` for both the build and dev mode) and explains the `-DskipTests` vs `-Dmaven.test.skip=true` distinction in DEVELOPING.md. Verified against a fresh checkout of main (05667bff).

The docs fix keeps the current behavior; if the team would rather make `-DskipTests` work under `-Dlocal` (e.g. via test-compile excludes in the local profile, per the options in #9018), I'm happy to follow up with that PR instead or in addition.

PR #9082, opened 2026-07-30
feat(ui): auto-detect version from artifact content (#1788)

### LINKED ISSUE #1788: Use the attribute `version` of the specification to register an artifact by UI
Many API specifications includes a `version` property (e.g. OpenAPI, AsyncAPI). That property helps to identify the API version to others and it helps to identify changes between versions.

AsyncAPI Sample:

```yaml
asyncapi: 2.1.0
info:
  title: Hello world application
  version: '0.1.0'
```

OpenApi Sample

```yaml
openapi: 3.0.2
info:
    title: Hello World API
    version: 1.0.0
```

It could be a great feature to use that property from the API spec to define the version of the artifact when it is uploaded using the UI, as similar with the auto-detect feature, then use auto-version-detect check. That feature in the UI will allow the developer to use or not the same version declared in the API definition or, set up the version needed/required.

WDYT?

Aarti-panchal01:
Implements #1788: when the content added in the Create Artifact wizard or the Create Version modal is an OpenAPI, AsyncAPI, or Swagger document, the Version field is prefilled from the spec's `info.version`. The field stays fully editable, and a helper text ("Version detected from the content.") indicates when the current value came from the content.

Behavior details:

- The detected version is only applied while the user hasn't edited the Version field themselves. Any manual edit, including clearing the field to fall back to a server-generated version, turns auto-detection off for that dialog session. This is tracked with a ref (not state) so the async content paths (URL fetch or file read resolving after the user typed) can't overwrite a manually entered version with a stale value.
- Only a string `info.version` is used. OpenAPI and AsyncAPI both define it as a string, and unquoted YAML numbers parse lossily (`version: 1.0` becomes the number 1), which would prefill a misleading value.
- Content is only treated as an API spec if it has a top-level `openapi`, `asyncapi`, or `swagger` property, so unrelated JSON/YAML that happens to contain an `info.version` field is ignored.

Scope notes:

- The Create Draft modal was deliberately left out: its version field is on the Coordinates step, before any content exists, and back-filling it would re-trigger the coordinate-availability validation on a step the user already left. I think that needs its own UX discussion; happy to follow up if there's interest.
- The earlier discussion in this issue about failing a create when the content changed but the version didn't is server-side behavior and out of scope for this UI change (the server already rejects duplicate versions with a 409).

Testing: unit tests added for the new detection helper (`content.utils.test.ts`), covering OpenAPI/AsyncAPI/Swagger in JSON and YAML, missing and non-string versions, multi-document YAML, and non-spec content. Also verified manually against a local registry: both create flows, the manual-override case, and the cleared-field case. One observation from setting that up: CI currently doesn't run the ui-app vitest suite (the build-ui job runs lint and build only) — I can add a test step in a follow-up PR if that's welcome.

Screenshots:

Version auto-detected from pasted OpenAPI content (Create Artifact wizard):
2-after-autodetect

A manually entered version is never overwritten by content changes:
3-user-version-preserved

Same behavior in the Create Version modal:
5-create-version-autodetect

Closes #1788

Note: this replaces #9080, which was auto-closed by the PR limit check while my previous PR (#9058, now closed) was still open.

### DIFF
--- a/ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
+++ b/ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, MouseEvent, Ref, useEffect, useState } from "react";
+import { FunctionComponent, MouseEvent, Ref, useEffect, useRef, useState } from "react";
 import "./CreateArtifactModal.css";
 import {
     FileUpload,
@@ -35,7 +35,7 @@ import { ArtifactTypesService, useArtifactTypesService } from "@services/useArti
 import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
 import { ArtifactLabel, LabelsFormGroup, ArtifactReferenceFormItem, ReferencesFormGroup, formItemsToReferences, isReferencesValid } from "@app/components";
 import { listToLabels } from "@utils/labels.utils.ts";
-import { detectContentType } from "@utils/content.utils.ts";
+import { detectContentType, detectVersionInContent } from "@utils/content.utils.ts";
 
 
 export type ValidType = "default" | "success" | "error";
@@ -127,6 +127,10 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> = 
     const [versionLabels, setVersionLabels] = useState<ArtifactLabel[]>([]);
     const [versionReferences, setVersionReferences] = useState<ArtifactReferenceFormItem[]>([]);
     const [isDetectingRefs, setIsDetectingRefs] = useState(false);
+    const [autoDetectedVersion, setAutoDetectedVersion] = useState<string | undefined>();
+    // Ref rather than state: content change handlers can fire asynchronously (e.g. when
+    // a URL fetch resolves) and must see the current value, not a stale closure.
+    const versionIsDirty = useRef(false);
 
     const urlService: UrlService = useUrlService();
     const atService: ArtifactTypesService = useArtifactTypesService();
@@ -195,16 +199,26 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> = 
     };
 
     const onFileTextChange = (_event: any, value: string | undefined): void => {
-        setData({
-            ...data,
+        // Auto-detect the version from the content (e.g. "info.version" in an OpenAPI or
+        // AsyncAPI spec), but only while the user has not edited the version themselves.
+        // The functional update matters: this handler can be called asynchronously (URL
+        // fetch, file read), so spreading a captured `data` could clobber newer state.
+        const detectedVersion: string | undefined = detectVersionInContent(value);
+        const applyDetectedVersion: boolean = !versionIsDirty.current;
+        setData(prevData => ({
+            ...prevData,
             firstVersion: {
-                ...data.firstVersion,
+                ...prevData.firstVersion,
+                version: applyDetectedVersion ? detectedVersion : prevData.firstVersion?.version,
                 content: {
                     content: value,
-                    contentType: detectContentType(data.artifactType, value)
+                    contentType: detectContentType(prevData.artifactType, value)
                 }
             }
-        });
+        }));
+        if (applyDetectedVersion) {
+            setAutoDetectedVersion(detectedVersion);
+        }
     };
 
     const onFileClear = (): void => {
@@ -220,6 +234,9 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> = 
     };
 
     const setVersionNumber = (newVersion: string): void => {
+        // The user edited the version (including clearing it): stop auto-detecting.
+        versionIsDirty.current = true;
+        setAutoDetectedVersion(undefined);
         setData({
             ...data,
             firstVersion: {
@@ -313,6 +330,8 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> = 
         if (props.isOpen) {
             setData(EMPTY_FORM_DATA);
             setVersionReferences([]);
+            setAutoDetectedVersion(undefined);
+            versionIsDirty.current = false;
             if (props.groupId) {
                 setGroupId(props.groupId);
             } else {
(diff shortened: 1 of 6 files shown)


ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
```diff
@@ -195,16 +199,26 @@ export const CreateArtifactModal: FunctionComponent<CreateArtifactModalProps> =
...
-            ...data,
+        // Auto-detect the version from the content (e.g. "info.version" in an OpenAPI or
+        // AsyncAPI spec), but only while the user has not edited the version themselves.
+        // The functional update matters: this handler can be called asynchronously (URL
+        // fetch, file read), so spreading a captured `data` could clobber newer state.
+        const detectedVersion: string | undefined = detectVersionInContent(value);
+        const applyDetectedVersion: boolean = !versionIsDirty.current;
+        setData(prevData => ({
+            ...prevData,
             firstVersion: {
```
paoloantinori, line 212: When no version is detected — non-spec content, or content that's transiently unparseable while editing — this sets `version` to `undefined`, clearing the field. Because detection runs on every content change while auto-detect is on, editing content can make the version flicker empty/populated and pasting non-spec content removes a previously-detected value. Consider overwriting only when something is actually detected, e.g. `version: (applyDetectedVersion && detectedVersion !== undefined) ? detectedVersion : prevData.firstVersion?.version`. That also avoids an `undefined` value (which can flip a controlled `TextInput` to uncontrolled depending on the `value` binding) and keeps this modal aligned with CreateVersionModal.

ui/ui-app/src/app/components/modals/CreateVersionModal.tsx
```diff
@@ -37,10 +41,21 @@ export const CreateVersionModal: FunctionComponent<CreateVersionModalProps> = (p
...
+    const versionIsDirty = useRef(false);
 
     const groupsService: GroupsService = useGroupsService();
 
     const onContentChange = (_event: any, value: any): void => {
+        // Auto-detect the version from the content (e.g. "info.version" in an OpenAPI or
+        // AsyncAPI spec), but only while the user has not edited the version themselves.
+        const detectedVersion: string | undefined = detectVersionInContent(value);
+        if (!versionIsDirty.current) {
+            setVersion(detectedVersion || "");
```
paoloantinori: Symmetric to the CreateArtifact case: when nothing is detected this clears the version to `""` rather than preserving it, and the empty-handling (`|| ""`) diverges from CreateArtifactModal (which yields `undefined`). Consider only updating `version` when `detectedVersion` is defined and using the same empty representation in both modals, so editing non-spec or intermediate content doesn't clear the field and the two create flows submit consistently.

paoloantinori (commented):
Overall this is a well-reasoned, carefully implemented change — the dirty-ref plus functional-update design correctly prevents a stale async content load (URL fetch / file read) from overwriting a version the user typed, the `openapi`/`asyncapi`/`swagger` gating avoids matching arbitrary JSON/YAML, and the string-only + `trim()` handling sidesteps the lossy unquoted-YAML-number trap. No blocking bugs found; the items below are refinements.

- **The version is cleared whenever the content isn't a detectable spec.** Both `onFileTextChange` (CreateArtifactModal.tsx:204-219) and `onContentChange` (CreateVersionModal.tsx:54-58) write the detection result into the version field on every content change while auto-detect is active. When the content isn't a spec at that instant — non-spec content, or a transiently-unparseable state mid-edit — the field is wiped, so editing the content can make the version flicker and pasting non-spec content removes a previously-detected value. Consider overwriting only when a version is actually detected and preserving the prior value otherwise.
- **The two modals diverge on the empty case.** CreateArtifactModal sets the version to `undefined` (CreateArtifactModal.tsx:210) while CreateVersionModal uses `detectedVersion || ""` (CreateVersionModal.tsx:56). An `undefined` value can also flip a controlled `TextInput` to uncontrolled depending on the `value` binding, and the two paths can yield different submit payloads; worth aligning.
- **The new tests aren't gated in CI.** The `detectVersionInContent` coverage (content.utils.test.ts) is thorough, but per your own note the `build-ui` job runs lint+build only, so the vitest suite never runs in CI — adding a test step would make this coverage actually enforce the behavior.
- **Minor/pre-existing:** `setVersionNumber` (CreateArtifactModal.tsx, ~237) still uses the older `setData({...data})` closure form rather than the functional update applied to `onFileTextChange`. It's unchanged by this PR and React 18's commit model makes a real clobber unlikely, so not a blocker — flagged only for consistency since you're already in this area.

ui/ui-app/src/app/components/modals/CreateVersionModal.tsx
Aarti-panchal01: Done in 9c10495!
the version is now only updated when detectedVersion is defined, which also removes the || "" vs undefined divergence: neither modal writes an empty value from the detection path anymore, so the two flows submit consistently.

ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
Aarti-panchal01: Done in 9c10495!
applyDetectedVersion now also requires detectedVersion !== undefined, so the previous value is preserved whenever the content isn't a detectable spec (including transiently unparseable states mid-edit), and the detection path never writes undefined into the field anymore. One deliberate choice: when content changes from a spec to non-spec, the helper text stays along with the value — clearing it on every transient parse failure would reintroduce the flicker, and the value did originally come from the content. Happy to clear it instead if you'd prefer.

Aarti-panchal01:
Thanks for the thorough review, @paoloantinori! All four points are addressed in 9c10495:

- Both modals now only overwrite the version when one is actually detected; non-spec or transiently unparseable content preserves the previous value instead of clearing it, so the field no longer flickers while editing.
- With that change the empty-case divergence between the two modals is gone — the detection path never writes an empty value, so both flows submit consistently.
- `setVersionNumber` now uses the functional `setData` update form, consistent with `onFileTextChange`.
- I went ahead and added the CI test step in this PR rather than a follow-up, since you raised it here: the `build-ui` job now runs the ui-app vitest suite via a root-level `test` script that follows the existing `lint`/`build` script conventions. It runs on this PR too, so the new coverage is enforced from here on.

Verified locally: 22/22 vitest tests pass, lint is clean, and I re-checked both create flows manually against a local registry — pasting a spec still prefills the version, replacing it with non-spec or unparseable content no longer clears the field (including JSON with an unrelated `info.version`, which is correctly ignored), a manually typed or cleared version is never overwritten, and versions created through both flows land correctly per the REST API.

paoloantinori:
Thanks @Aarti-panchal01 — confirmed all four points are addressed in 9c10495:

- The `detectedVersion !== undefined` guard on both modals correctly preserves the previous value when nothing is detected, so the field no longer flickers while editing.
- With that guard the two flows now behave consistently on the empty case.
- `setVersionNumber` uses the functional update, matching `onFileTextChange`, so neither can clobber newer state on an async resolve.
- The new `content.utils.test.ts` covers the spec/non-spec/lossy-number/multi-doc cases well.

From my side this is good to merge — leaving the final call to a maintainer.

EricWittmann:
Thanks @Aarti-panchal01 — well-designed feature with thoughtful edge case handling (the dirty-ref pattern, YAML number lossiness, spec-type gating). Great PR description too. Merged!

Issue #9061, opened 2026-07-30
feat(ci): issue self-assignment via comments


EricWittmann:
## Goal

Allow contributors to claim an issue by commenting, with a workflow that auto-assigns them. This
prevents duplicate work where multiple contributors independently start on the same issue, and
removes the bottleneck of waiting for a maintainer to manually assign.

## Behavior

- A contributor comments on an issue with a trigger phrase (e.g. `/assign-me`) to volunteer
- A GitHub Actions workflow triggers on `issue_comment: [created]`
- If the issue is **not assigned**: assign the commenter and post a confirmation comment
- If the issue is **already assigned**: post a comment explaining the issue is already claimed by
  `@<assignee>`, and do not change the assignment
- Maintainers can always override assignment directly via the GitHub UI

## Design considerations

- **Trigger format:** A simple command like `/assign-me` is unambiguous and consistent with the
  existing `/accept`, `/ready`, `/merge` command style
- **Scope:** Should work on any open issue, not gated by labels. Maintainers can always unassign
  if the claim is inappropriate
- **Bot accounts:** Ignore comments from known bots (`renovate[bot]`, `dependabot[bot]`,
  `apicurio-ci`)
- **Self-unassign:** Consider supporting `/unassign-me` so contributors can release an issue they
  can no longer work on

## Files

- `.github/workflows/issue-assignment.yml` — new workflow triggered on `issue_comment`
- `.github/scripts/issue-assignment.js` (or `.py`) — assignment logic
- `.github/PR_LIFECYCLE.md` — update contributor docs to reference the new command

## Dependencies

None — can proceed independently of the PR lifecycle changes.

## Constraints

- We should try to limit the number of open issues assigned to a single contributor to some sensible number

Anuragkumar-687:
@EricWittmann I'd be excited to work on this! The proposed workflow and design look well scoped, and I'd love to implement it. If the issue is available, I'd really appreciate it if you could assign it to me.

senutpal:
hey @EricWittmann , happy to take a look at this.

[...]

Issue #8980, opened 2026-07-29
bug(import-export): contract rules are silently omitted from registry backups


vivek-gite:
## Summary

Exporting a registry that contains contract rules produces a ZIP archive that appears successful but silently omits every `ContractRuleEntity`. Importing that archive therefore restores artifacts and content without their global-, artifact-, or version-scoped contract rules.

This is silent data loss for Data Contracts / governance configuration.

## Impact

Users can successfully:

1. Configure global, artifact, or version contract rules.
2. Export the registry.
3. Import that backup into an empty registry.
4. Discover that contract rulesets no longer exist.

Because export catches and ignores the writer exception, neither the HTTP export response nor the resulting ZIP indicates that data was skipped.

This can disable governance, quality, lifecycle, migration, or runtime rule enforcement after a disaster-recovery restore.

### 1. SQL export intentionally emits contract-rule entities

Both full and group-scoped exports include contract rules:

- Full export:  
  [`AbstractSqlRegistryStorage.java#L1337-L1347`](https://github.com/Apicurio/apicurio-registry/blob/a35ccfeecd58fa42f06f34f1d03df6c317c59520/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java#L1337-L1347)

- Group-scoped export:  
  [`AbstractSqlRegistryStorage.java#L1325-L1335`](https://github.com/Apicurio/apicurio-registry/blob/a35ccfeecd58fa42f06f34f1d03df6c317c59520/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java#L1325-L1335)

The repository correctly queries and maps `ContractRuleEntity` values:

- [`SqlExportRepository.java#L312-L339`](https://github.com/Apicurio/apicurio-registry/blob/a35ccfeecd58fa42f06f34f1d03df6c317c59520/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlExportRepository.java#L312-L339)

### 2. The v3 ZIP writer does not support `ContractRule`

`EntityWriter.writeEntity` handles content, groups, artifacts, versions, rules, comments, branches, and manifest entities, but has no `ContractRule` branch:

[...]

PR #9085, opened 2026-07-30
feat(ci): add stale-assigned and unassigned-triage scans for #9063

### LINKED ISSUE #9063: feat(ci): stale assigned issue detection and unassigned issue monitoring
## Goal

Automatically detect assigned issues that have gone stale (no linked PR after a period), and
surface unassigned issues for maintainer triage. This keeps the backlog healthy and ensures
assigned issues don't block other contributors indefinitely.

## Stale assigned issue detection

- A scheduled workflow (cron, e.g. daily) scans open issues that are assigned
- If an assigned issue has no linked/referenced PR after **14 days**, post a comment pinging the
  assignee asking for a status update and add a `lifecycle/stale-assigned` label
- If no response or PR after an additional **7 days** (21 days total), auto-unassign with a
  comment explaining the issue is now available for others
- The stale clock resets if the assignee comments on the issue

## Unassigned issue monitoring

- The same scheduled workflow identifies open, unassigned issues that have had no maintainer
  activity (comment or label change) for a configurable period
- Add a `needs-triage` label to surface them for maintainer review
- Over time, triage behavior can be refined based on labels:
  - `priority/high` or `!important` issues get shorter triage windows
  - `priority/low` issues get longer windows before being flagged

## The `!important` label

- Introduce an `!important` label that maintainers can apply to flag high-impact issues for
  contributors
- Issues with this label appear prominently in triage reports and could be surfaced in the
  welcome message or contributor docs

## Files

- `.github/workflows/issue-monitoring.yml` — new scheduled workflow
- `.github/scripts/issue-monitoring.js` (or `.py`) — stale detection and triage logic

## Dependencies

- **#9061** — issue self-assignment should be in place so that unassigned issues can be easily
  claimed once surfaced

nafisa404:
## Summary

Implements the two scanning behaviors from #9063 as a single daily scheduled
workflow: stale-assigned issue detection and unassigned issue triage.

Closes #9063

## What this adds

- `.github/workflows/issue-monitoring.yml` — scheduled workflow (daily cron,
  `23 4 * * *`), plus manual `workflow_dispatch` with a `dry_run` input for
  safe testing.
- `.github/scripts/issue-monitoring.js` — scan logic, following the
  `createApi(github, owner, repo)` factory pattern and timeline-events-based
  clock detection already established in `pr-lifecycle.js`.
- `.github/issue-monitoring.yml` — externalized config (thresholds, triage
  windows, maintainer list, message templates), converted to JSON at runtime
  the same way `pr-lifecycle.yml`'s config is.

### Stale-assigned detection
- Scans open, assigned issues.
- If no linked/referenced PR (via `cross-referenced` timeline events,
  checking the referenced PR's **live state** — open or merged, not just
  its presence) and no comment from the assignee after **14 days**: posts
  a ping comment and adds `lifecycle/stale-assigned`.
- If still nothing after **21 days total**: removes the assignee(s),
  removes the label, and posts a comment explaining the issue is available
  again.
- The clock resets whenever the assignee comments — detected via timeline
  events, so no state needs to be persisted between runs.

> **Note on linked-PR detection:** `connected`/`disconnected` are
> GraphQL-only event types (`ConnectedEvent`/`DisconnectedEvent`) and are
> **not** part of the REST timeline API this script uses — confirmed via
> GitHub's OpenAPI schema and empirically absent across ~55 sampled issue
> timelines in this repo. A PR linked purely through the "Development"
> sidebar UI with zero textual mention (no `Fixes #N` anywhere) would need
> a GraphQL `closingIssuesReferences` lookup to detect, which this PR does
> not implement. Flagged as a known limitation — discussed with
> @paoloantinori in review, still open pending a concrete example from
> this repo.

### Unassigned issue triage
- Scans open, unassigned issues.
- Clock is based on **maintainer activity only** (comments or label changes
  from logins in the `maintainers:` config list) — a non-maintainer comment
  or a bot-applied label (e.g. from `classify-issues.yml`) does not reset it,
  since the goal is surfacing issues no maintainer has actually looked at.
- Triage window depends on label, checked in order (first match wins):
  - `priority/high` / `!important` → 3 days (tight window — an urgent,
    untriaged issue sitting for a week is the exact failure mode this
    feature exists to catch)
  - `priority/low` → 21 days (mirrors the assigned-issue total lifecycle;
    no urgency signal, so a generous grace period)
  - no priority label (the common case) → 7 days, matching the existing
    `days_until_stale: 7` convention already used in `pr-lifecycle.yml`
- Adds `needs-triage` (no comment). Once applied, it's left for a
  maintainer to remove/act on — no auto-removal logic, since the spec
  only calls for adding the label.

### Dry-run mode
`workflow_dispatch` accepts a `dry_run` boolean (default `false`). When
true, all mutating calls (label add/remove, comment, unassign) log via
`core.info` with a `[dry-run]` prefix instead of touching the API; read
calls remain live so the logged output reflects real issue state. Scheduled
(cron) runs always execute for real. Accepts both a JSON boolean and the
string `'true'` for this input, since `workflow_dispatch` boolean inputs
can arrive as either depending on trigger path.

## Design notes / assumptions for review

- **Multi-assignee issues**: the original spec describes "the assignee"
  (singular). For issues with multiple assignees, this implementation treats
  a comment from *any* assignee as resetting the clock, and unassigns *all*
  of them together at the 21-day mark. Flagging this explicitly since it's
  an interpretation, not spec text.
- **No hard dependency on #9061**: comment templates deliberately avoid
  referencing any specific "claim this issue" mechanism (e.g. a `/claim`
  command), since that functionality is still open in #9061. Once #9061
  merges, the unassign message could be updated to explicitly point
  contributors at the claim command.
- **`!important` label**: this PR does not programmatically create the
  label — treating it as something maintainers apply manually via the
  GitHub UI.
- **No failure notification**: unlike `pr-lifecycle.yml`, this workflow
  doesn't post anywhere on failure. A failed run currently just shows red
  in the Actions tab. Could hook into the existing Slack notify workflow
  later if this proves flaky — left out of v1 to keep scope tight.
- **Label colors**: `lifecycle/stale-assigned` reuses the `F7BF6A` color
  already used for `lifecycle/waiting-on-maintainer` in `pr-lifecycle.js`.
  `needs-triage` uses `D4C5F9`, chosen to be visually distinct from both
  that and the existing `D93F0B` triage-related color in `apicurio-bot.yaml`.

## Out of scope / future follow-ups

- Detecting sidebar-only-linked PRs (no textual mention anywhere) via
  GraphQL's `closingIssuesReferences`, if this turns out to be a real
  pattern in practice — raised by @paoloantinori during review, no
  confirmed example in this repo yet.
- Updating the unassign message once #9061 (self-assignment via comments)
  merges, to point contributors at the claim mechanism.
- Failure notification for scan errors.
- Revisiting label colors repo-wide (separate concern from this PR).


.github/scripts/issue-monitoring.js
```diff
@@ -0,0 +1,322 @@
...
+  const configPath = path.join(process.cwd(), '.github', 'issue-monitoring.json');
+  return JSON.parse(fs.readFileSync(configPath, 'utf8'));
+}
+
+function hasLabel(issue, label) {
+  return (issue.labels || []).map(l => (typeof l === 'string' ? l : l.name)).includes(label);
+}
+
+function formatMentions(logins) {
+  return logins.map(login => `@${login}`).join(' ');
```
paoloantinori, line 40: `formatMentions` prepends `@` (returns `@alice`), but both templates write the placeholder as `@{assignee}` — a literal `@` before `{assignee}`. `renderTemplate` swaps only the `{assignee}` token, leaving the template's `@`, so every posted comment renders as `@@alice` (ping path, line 222; unassign path, line 207; single- and multi-assignee alike). Verified by executing the code. Simplest fix: change `@{assignee}` to `{assignee}` in `ping_message`/`unassign_message` (or drop the `@` here). Side note I couldn't verify: whether GitHub still delivers a mention notification for `@@user`, so I wouldn't rely on the ping notifying the assignee until checked.

.github/workflows/issue-monitoring.yml
```diff
@@ -0,0 +1,40 @@
...
+  scan:
+    name: Scan Issues
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'Apicurio'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Convert config to JSON
+        run: yq -o json .github/issue-monitoring.yml > .github/issue-monitoring.json
```
paoloantinori, line 33: `yq` is not part of the default `ubuntu-latest` runner image (only `jq` ships by default), so this step will likely fail with `command not found`. Confirm that `pr-lifecycle.yml` relies on the same bare `yq` invocation and actually works on the runner; otherwise add an install step (e.g. the `mikefarah/yq` action) or parse the YAML inside the following Node `github-script` step.

.github/scripts/issue-monitoring.js
```diff
@@ -0,0 +1,322 @@
...
+      core.info(`Issue #${issue.number} already pinged, ${Math.floor(daysSinceClockStart)}d since clock start`);
+    }
+    return;
+  }
+
+  if (daysSinceClockStart >= daysUntilPing) {
+    const body = renderTemplate(config.assigned_stale.ping_message, {
+      assignee: formatMentions(assignees),
+      days: daysUntilPing,
+      days_until_unassign_remaining: daysUntilUnassign - daysUntilPing,
```
paoloantinori: `days_until_unassign_remaining` is the fixed threshold gap (`21 - 14 = 7`), not the actual time left. If a scheduled run is delayed and the ping first fires at, say, day 18, the message still says '7 days' when only ~3 remain. Use `Math.ceil(daysUntilUnassign - daysSinceClockStart)` for an accurate countdown.

paoloantinori (commented):
Well-structured feature that mirrors the established `pr-lifecycle` patterns, and the clock/state logic is carefully reasoned. Before merging there's one always-on rendering bug to fix and a few workflow items to verify; most are quick.

- **Malformed mentions (the main issue):** `formatMentions` prepends `@` while the `ping_message`/`unassign_message` templates already write `@{assignee}`, so every comment is posted as `@@assignee` — verified by running the code. Affects the ping path (`issue-monitoring.js:222`), the unassign path (`:207`), single- and multi-assignee. Fix: drop the leading `@` in the two templates (or in `formatMentions` at `:40`). I couldn't confirm whether GitHub still notifies on a `@@user` mention, so don't assume the ping reaches the assignee.
- **`yq` availability:** the config step runs bare `yq -o json` (`workflows/issue-monitoring.yml:33`); `yq` isn't on the default `ubuntu-latest` image (only `jq` is), so this likely fails before the scan runs unless `pr-lifecycle.yml` installs it elsewhere. Confirm, or parse the YAML inside the Node `github-script` step.
- **`assignee: '*'` wildcard (worth confirming):** if the list-issues endpoint doesn't treat `*` as 'any assignee' (`issue-monitoring.js:74`), the entire assigned-stale scan silently matches nothing. I couldn't verify this against the live API.
- **Fork test wouldn't have run the job:** the `repository_owner == 'Apicurio'` guard (`:27`) skips the job on forks, so a fork `dry_run` wouldn't exercise the scan — meaning `yq`, the pinned action SHAs (`:30 # v7`, `:36 # v9`), and the scan logic are effectively unverified. Worth confirming where the validation actually ran and that those SHAs resolve.
- **Stale countdown is a constant:** `days_until_unassign_remaining` is `21-14 = 7` regardless of when the ping actually fires (`issue-monitoring.js:225`); a delayed run understates the real remaining time. One-line fix.

.github/scripts/issue-monitoring.js
nafisa404: fixed by removing the leading @ from the templates in issue-monitoring.yml, since formatMentions() already prepends the @ per login. Pushed. Since it's now a plain single @mention, GitHub notification behavior should work as normal.

.github/workflows/issue-monitoring.yml
nafisa404: Verified — GitHub's hosted ubuntu-latest runners ship yq pre-installed(since 2021), so no install step is needed. This matches why pr-lifecycle.yml also calls bare yq with no setup step, across 9 separate jobs. No change needed here, but happy to add an explicit install step(e.g. mikefarah/yq-action) if you'd prefer not to rely on the runner's bundled version for supply-chain-pinning reasons.

.github/scripts/issue-monitoring.js
nafisa404: Fixed — now computed as Math.ceil(daysUntilUnassign - daysSinceClockStart),so a delayed run reports the real remaining time rather than the fixed 21-14 gap. Pushed.

.github/scripts/issue-monitoring.js
```diff
@@ -0,0 +1,322 @@
...
+      core.info(`Issue #${issue.number} already pinged, ${Math.floor(daysSinceClockStart)}d since clock start`);
+    }
+    return;
+  }
+
+  if (daysSinceClockStart >= daysUntilPing) {
+    const body = renderTemplate(config.assigned_stale.ping_message, {
+      assignee: formatMentions(assignees),
+      days: daysUntilPing,
+      days_until_unassign_remaining: Math.ceil(daysUntilUnassign - daysSinceClockStart),
```
paoloantinori, line 248: When `daysSinceClockStart` already exceeds `daysUntilUnassign` (e.g. an issue 30 days stale on the first run after deploy, or after a gap in scheduled runs), this unlabeled branch posts the ping instead of unassigning, and `Math.ceil(daysUntilUnassign - daysSinceClockStart)` goes negative — so the public comment says something like "automatically unassigned in -9 days." The unlabeled path needs an `if (daysSinceClockStart >= daysUntilUnassign)` branch that unassigns directly (mirroring the labeled path), or at minimum this value should be clamped to >= 1.

paoloantinori (commented):
Overall, this is a clean implementation that mirrors the established `pr-lifecycle.js` patterns (factory API, timeline-based clock, no persisted state), and the dry-run wiring and GitHub API usage (`assignee` sentinels, `issues: write` scope) check out. The main concern is a logic gap in the stale-assigned scan that produces a misleading public comment on first deployment, plus two integration points worth confirming.

- `.github/scripts/issue-monitoring.js:221-229` — On the first run after deploy (or after any gap in scheduled runs), an assigned issue already past `days_until_unassign` (21d) but never labeled takes the unlabeled path and gets *pinged* instead of *unassigned*; `days_until_unassign_remaining` (line 225) then evaluates negative, so the posted comment reads e.g. "automatically unassigned in -9 days". Unassign only fires on the next run. The unlabeled path should unassign directly when `daysSinceClockStart >= daysUntilUnassign`.
- `.github/scripts/issue-monitoring.js:165-170` — `hasLinkedPr` counts any `connected`/`cross-referenced` event as a live link regardless of PR state. A PR closed-without-merge (or later unlinked via `disconnected`) leaves its `connected` event in the timeline, so the issue is permanently exempt from stale detection — undercutting the feature's purpose for abandoned PRs.
- `.github/workflows/issue-monitoring.yml:33` — The `yq -o json` step has no preceding install step, and `yq` isn't a standard preinstalled tool on `ubuntu-latest` (only `jq` is). The PR says this mirrors `pr-lifecycle.yml`, so please confirm that workflow doesn't install yq separately; if it does and this one doesn't, the step fails with "command not found" and the whole scan never runs.

.github/scripts/issue-monitoring.js
```diff
@@ -0,0 +1,322 @@
...
+  const assignedAt = new Date(Math.max(...assignedEvents.map(e => new Date(e.created_at).getTime())));
+
+  const staleLabel = config.assigned_stale.stale_label;
+
+  // A PR that references this issue (via #number/URL mention, or the
+  // "Link a pull request" UI) after the current assignment counts as progress.
+  const hasLinkedPr = timeline.some(e => {
+    if (new Date(e.created_at) <= assignedAt) return false;
+    if (e.event === 'cross-referenced') return !!e.source?.issue?.pull_request;
+    if (e.event === 'connected') return true;
```
paoloantinori: `connected` and `cross-referenced` events persist in the timeline even after the PR is closed without merging or later unlinked (`disconnected`). Since `hasLinkedPr` returns true on any such event, an issue that once had an abandoned/rejected PR is permanently exempt from stale detection — no ping or unassign ever. Consider checking the referenced PR's state (open or merged), or accounting for `disconnected` events, so only currently-relevant links count.

.github/workflows/issue-monitoring.yml
paoloantinori, line 33: Confirm `yq` is actually available here: it isn't a standard preinstalled tool on `ubuntu-latest` (only `jq` is) and there's no install step before this. The PR notes this mirrors `pr-lifecycle.yml`, so verify that workflow doesn't install yq in a separate step this one omits — otherwise this fails with "command not found" and the entire scan never runs. If pr-lifecycle.yml already relies on a preinstalled yq, this is fine; just worth an explicit confirmation.

nafisa404: Confirmed — yq (mikefarah/yq, the variant with -o json support) is preinstalled on both ubuntu-24.04 and ubuntu-22.04 GitHub-hosted runner images (what ubuntu-latest currently resolves to), version 4.53.3 per [actions/runner-images](https://github.com/actions/runner-images). pr-lifecycle.yml relies on this exact same preinstalled binary with no install step, and that workflow is already merged and running successfully in this repo, so this isn't a new assumption — it's the established pattern here. No change needed.

.github/scripts/issue-monitoring.js
nafisa404: fixed. The unlabeled branch now checks`daysSinceClockStart >= daysUntilUnassign` first and unassigns directly
when that threshold's already passed (covers first-run-after-deploy and any scheduling gaps), only falling through to the ping logic otherwise.Pushed in the latest commit.

.github/scripts/issue-monitoring.js
nafisa404: fixed. hasLinkedPr now checks the referenced PR's actual current state (open or merged) rather than treating any cross-reference as permanent, so an abandoned closed-unmerged PR no longer exempts an
issue from stale detection forever.
Also verified this works correctly against a real case in the repo: PR#9101 (closing #8980) was cross-referenced at 13:54:56Z but the PR wasn't actually closed until 15:45:03Z — almost two hours later. Fetching the
timeline just now still shows `state: "closed"` for that event, which confirms GitHub recomputes this field live rather than freezing it at cross-reference time, so no extra API call per candidate is needed.
On `connected`/`disconnected`: I checked GitHub's REST API OpenAPI schema and sampled 12 real issue/PR timelines in this repo — those event types don't actually appear anywhere; all issue↔PR linking comes through as
`cross-referenced`. Removed the dead `connected` branch accordingly rather than trying to extend it. Pushed.

carlesarnal (changes requested):
Good work on addressing all of paoloantinori's feedback. The unlabeled-branch threshold check, the PR-state-aware hasLinkedPr, the dynamic countdown, and the template fix all look correct. One remaining issue: all 3 commits are missing DCO sign-off (Signed-off-by trailer), which is required per CONTRIBUTING.md. Quick fix: git rebase HEAD~3 --signoff then git push --force-with-lease. Verified fixes: double-mention templates fixed, dynamic countdown fixed (line 240), unlabeled-branch gap fixed (lines 221-233), hasLinkedPr checks PR state (lines 164-171), dead connected branch removed, yq confirmed preinstalled.

nafisa404:
@carlesarnal Thanks for confirming the fixes.I've addressed the DCO issue. All 3 commits now carry the Signed-off-by trailer.

.github/scripts/issue-monitoring.js
```diff
@@ -0,0 +1,340 @@
...
+  // A PR that references this issue (via #number/URL mention) after the
+  // current assignment counts as progress, but only while the PR is still
+  // open or was merged — a closed-unmerged PR is an abandoned attempt, not
+  // an active link. `source.issue` reflects the referenced PR's current
+  // state (GitHub recomputes it on each timeline fetch, it is not a
+  // snapshot from when the cross-reference was created), so no separate
+  // pulls.get call is needed to check freshness.
+  const hasLinkedPr = timeline.some(e => {
+    if (new Date(e.created_at) <= assignedAt) return false;
+    if (e.event !== 'cross-referenced') return false;
```
paoloantinori, line 177: The PR description says linked PRs are detected via `cross-referenced` *or* `connected` timeline events, but this only matches `cross-referenced`. PRs attached through GitHub's 'Development/Linked pull requests' UI produce a `connected` event (frequently without a usable `cross-referenced` event), so those issues still get pinged/unassigned despite having a linked PR. Consider also accepting `e.event === 'connected'` (and possibly `referenced`/`closed` link events).

.github/scripts/issue-monitoring.js
```diff
@@ -0,0 +1,340 @@
...
+
+// ---------------------------------------------------------------------------
+// Entry Point
+// ---------------------------------------------------------------------------
+
+async function handleScan({ github, context, core }) {
+  const { owner, repo } = context.repo;
+  // workflow_dispatch boolean inputs arrive as the string 'true'/'false';
+  // schedule-triggered runs have no inputs at all, so default to false.
+  const dryRun = context.payload.inputs?.dry_run === 'true';
```
paoloantinori: This assumes the boolean `dry_run` input arrives as the string `'true'`, but `type: boolean` workflow_dispatch inputs are typically delivered as JSON booleans in `context.payload.inputs`, so this would evaluate `true === 'true'` -> false and a 'dry run' would perform real mutations (label add/remove, comment, unassign). Recommend accepting both (`=== true || === 'true'`) and verifying with an actual dispatch run — the fork guard (`repository_owner == 'Apicurio'`) and the placeholder Actions link suggest this path hasn't been exercised yet.

paoloantinori (commented):
Overall the scan is well-structured and the clock/state logic is thoughtful, but a couple of correctness gaps are worth resolving before this runs against live data: linked-PR detection only covers `cross-referenced` events (the PR also promises `connected`), and the `dry_run` flag may never engage due to boolean-vs-string payload handling.

- `.github/scripts/issue-monitoring.js:172` — `hasLinkedPr` matches only `cross-referenced`. PRs linked via the GitHub 'Development/Linked pull requests' UI emit a `connected` event (often without a usable `cross-referenced` event), so those assigned issues still get pinged at 14d and unassigned at 21d despite a linked PR, contradicting the PR's stated detection logic.
- `.github/scripts/issue-monitoring.js:329` — `dry_run` is read as `=== 'true'`, but `type: boolean` dispatch inputs are typically delivered as JSON booleans in `context.payload.inputs`, which would make this `true === 'true'` -> false and let a 'dry run' perform real mutations. Handle both representations and confirm with an actual dispatch.
- `.github/scripts/issue-monitoring.js:175` — `!!prRef.merged_at` may always be false: the `pull_request` object in a `cross-referenced` event is usually a minimal ref without `merged_at`, so a merged PR (state `closed`) may not register as linked. The `source.issue.state` freshness assumption (lines 166-169) is also worth validating against a real timeline payload.
- `.github/workflows/issue-monitoring.yml:33` — `yq` is not preinstalled on `ubuntu-latest` (only `jq` is). Confirm `pr-lifecycle.yml` installs it, or add an install step, or this step fails before the scan runs.
- `.github/workflows/issue-monitoring.yml:27` — the `repository_owner == 'Apicurio'` guard skips the job on a personal fork, and the Actions-run link in the description is still a placeholder, so the dry-run path likely hasn't been exercised against live API responses — which is exactly what would surface the items above.

.github/scripts/issue-monitoring.js
nafisa404: Fixed — now accepts both a real boolean and the string 'true':

const dryRun = context.payload.inputs?.dry_run === true ||
               context.payload.inputs?.dry_run === 'true';

.github/scripts/issue-monitoring.js
nafisa404: Confirmed connected isn't a real REST timeline event — I full-text searched the entire GitHub REST OpenAPI spec (every string, not just schema definitions) and also empirically scanned ~55 issue timelines across this repo (oldest and newest). Zero occurrences either way.

What you're likely picturing is GraphQL's ConnectedEvent/DisconnectedEvent— those are real, but they're a separate API (GraphQL) from the REST timeline endpoint (issues.listEventsForTimeline) this script uses. The REST endpoint structurally never emits event: 'connected', regardless of how the PR was linked.

Matching sidebar-only-linked PRs (no Fixes/Closes keyword, no textual mention anywhere) would need a github.graphql() call against closingIssuesReferences instead — a meaningfully bigger change than a
one-line event-type check, so I didn't want to slip it in without confirming it's actually needed here. Updated the PR description to note this as a known limitation.
If you've seen a specific issue in this repo where a PR was linked purely via the sidebar with zero textual mention, I'd like to look at it directly— happy to add the GraphQL lookup as a follow-up once I can confirm the
exact shape needed.

paoloantinori:
Both points addressed cleanly, thanks @nafisa404.

1. **`dry_run` boolean** — `=== true || === 'true'` covers both the JSON-boolean and string delivery paths. Good.

2. **`connected` event** — fair point, and the empirical check (REST OpenAPI spec + ~55 sampled timelines, zero `connected` events) is the right way to settle it. The REST `listEventsForTimeline` endpoint genuinely doesn't emit `connected`/`disconnected`; those are GraphQL-only `ConnectedEvent`/`DisconnectedEvent`. Sidebar-only links (no `Fixes`/`Closes` keyword, no textual mention) would need a `github.graphql()` call against `closingIssuesReferences` — agree that's a separate, larger change and not worth slipping in here. Noting it as a known limitation in the PR description is the right call. If a concrete sidebar-linked case shows up in practice, the GraphQL follow-up is the path.

LGTM from my side. Leaving the merge call to the maintainers.

nafisa404:
@paoloantinori Thanks for the thorough review, really appreciated catching the dry_run and connected-event issues. Glad we landed on the same conclusionfor the GraphQL question.
 Leaving it with the maintainers for the mergecall.

nafisa404:
@EricWittmann @carlesarnal @jsenko whenever you get a moment, would appreciate a look at this. Thanks!

paoloantinori:
Thanks @nafisa404! Glad the review was useful and we aligned on the dry_run and connected-event items, plus the GraphQL call. I'll leave the merge decision to the maintainers — nothing further from my side on review.

jsenko (commented):
## Automated review

Well-structured PR that follows the existing `pr-lifecycle.js` patterns. Permissions are appropriately scoped and the dry-run mode is a good addition.

### should-fix: Triage label missing `lifecycle/` prefix

`.github/issue-monitoring.yml:39` sets `label: needs-triage`, but the stale label at line 18 uses `lifecycle/stale-assigned`. For consistency with the existing label convention, this should be `lifecycle/needs-triage`.

### should-fix: Template rendering doesn't escape replacement values

`.github/scripts/issue-monitoring.js:46`

```javascript
result = result.replace(new RegExp(`\\{${key}\\}`, 'g'), value);
```

If `value` contains `$` (e.g., in a template variable derived from user input), `String.prototype.replace` interprets `$` sequences as replacement patterns (`$&`, `$'`, etc.). Use `replaceAll()` instead:

```javascript
result = result.replaceAll(`{${key}}`, String(value));
```

This avoids regex entirely and is immune to special characters in both keys and values.

### should-fix: Maintainer list duplicated across configs

`.github/issue-monitoring.yml` and `.github/pr-lifecycle.yml` both define identical maintainer lists. If the team changes, both files need manual sync. Consider extracting to a shared `.github/maintainers.yml` and loading it in both workflows, or at minimum add a comment in both files pointing to the other.

### nice-to-have: Cross-referenced PR state freshness

`.github/scripts/issue-monitoring.js:166-175`

The code assumes `source.issue.state` in timeline events reflects the PR's current state. The inline comment says "GitHub recomputes `source.issue` on each timeline fetch" — if that's verified behavior, a brief link to the GitHub API docs or a test result would help future maintainers trust this assumption. If it's unverified, fetching the PR's current state via `github.rest.pulls.get()` would be safer (at the cost of extra API calls).

nafisa404:
@jsenko Thanks for the review — all three addressed:

1. Triage label prefix: Fixed — renamed to lifecycle/needs-triage to
match the established convention (used across pr-lifecycle.js,
verify.yaml, and documented in PR_LIFECYCLE.md). Also checked
apicurio-bot.yaml's existing triage/needs-triage label to confirm no
collision — it's a different signal (content-classification failure at
creation time vs. our scan's "no maintainer activity after N days"), so
kept them separate.

2. Template rendering: Fixed — switched to replaceAll(), which avoids
regex entirely and is immune to $ substitution patterns ($&, $', $1,
etc.) that String.replace() would otherwise interpret in the replacement
value. Added explicit String(value) coercion since replaceAll requires a
string argument. Confirmed Node 24 runtime via actions/github-script@v9,
well past the replaceAll requirement (Node 15+).

3. Maintainer list duplication: Added cross-referencing comments in both
issue-monitoring.yml and pr-lifecycle.yml noting the duplication is
intentional for now, with a pointer to extract into a shared
.github/maintainers.yml as a follow-up — didn't want to expand this PR's
scope into refactoring pr-lifecycle.yml's config.

4. Cross-referenced PR state freshness: Added the verification directly
to the code comment — tested empirically earlier in review: PR #9101
(closing #8980) was cross-referenced while still open, but refetching the
timeline after the PR was later closed showed source.issue.state
correctly updated to "closed", proving GitHub recomputes this field live
rather than snapshotting it at cross-reference time. Same pattern
confirmed for merged_at using PR #9013.

All pushed. Let me know if anything needs another look.

nafisa404:
@jsenko @EricWittmann this has been stale for a while would appreciate a review.Thanks!

carlesarnal:
I'm sorry, but we won't be accepting changes to Ci workflows from external contributors. You can look at other issues for open problems

PR #9097, opened 2026-07-31
fix(ui): reset reauthentication state when interceptor is unregistered

### LINKED ISSUE #9096: Re-authentication remains stuck in pending state after interceptor unmounts, swallowing subsequent 401 handling
## Description

**Registry Version**: 3.3.2-SNAPSHOT
**Persistence type**: sql

The UI re-authentication service can remain in a pending state after a custom re-authentication interceptor is unregistered. This can occur when a component that registers an interceptor (such as the Editor page) is unmounted while a re-authentication flow is still pending.

Once the interceptor has been removed, subsequent `401 Unauthorized` responses continue to detect an existing pending re-authentication request and do not initiate a new authentication flow. As a result, neither another registered interceptor nor the default OIDC redirect is executed, preventing the user from being prompted to authenticate again.

### Environment

* Apicurio Registry 3.3.2-SNAPSHOT
* Apicurio Registry UI
* OIDC authentication
* Local development environment
* Google Chrome (latest)

### Steps to Reproduce

1. Open the Editor page.
2. Trigger a `401 Unauthorized` response so the custom re-authentication interceptor is invoked.
3. Before the re-authentication flow completes, navigate away from the Editor page.
4. Trigger another `401 Unauthorized` response from a different page.
5. Observe that no new re-authentication flow is started.

**Expected**

* When the active re-authentication interceptor is unregistered, any associated pending state should be cleared.
* Subsequent `401 Unauthorized` responses should either be handled by another registered interceptor or fall back to the default OIDC re-authentication flow.

**Actual**

* The re-authentication service remains in a pending state after the interceptor is unregistered.
* Subsequent `401 Unauthorized` responses are treated as though a re-authentication flow is already in progress, preventing a new authentication flow from being initiated.

### Logs

No relevant server-side logs were observed. The issue is reproducible through the UI by triggering consecutive `401 Unauthorized` responses before and after the component that registered the custom re-authentication interceptor is unmounted.

kunaldevsahu:
## Summary

This PR fixes a UI issue where the global re-authentication service could remain in a pending state after the active re-authentication interceptor is unregistered.

If a component that registers a custom interceptor (such as the Editor page) is unmounted while a re-authentication flow is pending, the interceptor is removed but the pending state is not cleared. As a result, subsequent `401 Unauthorized` responses may not trigger a new re-authentication flow.

## Changes

* Clear the pending re-authentication state when the active interceptor is unregistered.
* Preserve the existing re-authentication flow when the interceptor remains registered.
* Add unit tests covering:

  * Initial service state.
  * Custom interceptor flow.
  * Cleanup of pending state when the active interceptor is unregistered.
  * Default OIDC redirect behavior when no interceptor is registered.

## Testing

* Added regression tests for the re-authentication service.
* Verified all unit tests pass.
* Verified lint checks pass.

## Related Issue

Fixes #9096

### DIFF
--- a/ui/ui-app/src/services/useReauthenticationService.test.ts
+++ b/ui/ui-app/src/services/useReauthenticationService.test.ts
@@ -47,20 +47,35 @@ describe("useReauthenticationService", () => {
     });
 
     it("should cancel pending state when interceptor is unregistered while pending", async () => {
-        const interceptor = vi.fn().mockResolvedValue(true);
-        const unregister = registerInterceptor(interceptor);
+        // 1. Return a never-resolving promise to simulate a truly in-flight re-authentication flow
+        const pendingPromise = new Promise<boolean>(() => {});
+        const firstInterceptor = vi.fn().mockReturnValue(pendingPromise);
+        const unregisterFirst = registerInterceptor(firstInterceptor);
 
-        const promise = service.requestReauthentication(mockAuth);
+        // 2. Trigger requestReauthentication
+        service.requestReauthentication(mockAuth);
         expect(service.isReauthenticationPending()).toBe(true);
+        expect(firstInterceptor).toHaveBeenCalledTimes(1);
 
-        // Unregister the interceptor while re-auth is pending
-        unregister();
+        // 3. Unregister the interceptor while the promise is still pending
+        unregisterFirst();
 
+        // Verify the pending state is cleaned up as expected
         expect(service.isReauthenticationPending()).toBe(false);
-        await promise;
 
-        // Post-await assertion to ensure it remains false
-        expect(service.isReauthenticationPending()).toBe(false);
+        // 4. Register a new interceptor
+        const secondInterceptor = vi.fn().mockResolvedValue(true);
+        registerInterceptor(secondInterceptor);
+
+        // 5. Trigger a second requestReauthentication and assert a fresh flow starts
+        const promise2 = service.requestReauthentication(mockAuth);
+        expect(service.isReauthenticationPending()).toBe(true);
+        expect(secondInterceptor).toHaveBeenCalledTimes(1);
+
+        await expect(promise2).resolves.toBe(true);
+
+        // Ensure the first interceptor was not invoked again
+        expect(firstInterceptor).toHaveBeenCalledTimes(1);
     });
 
     it("should trigger redirect if no interceptor is registered", async () => {
--- a/ui/ui-app/src/services/useReauthenticationService.test.ts
+++ b/ui/ui-app/src/services/useReauthenticationService.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it, vi, beforeEach } from "vitest";
+import { getReauthenticationService } from "./useReauthenticationService";
+import { AuthService } from "@apicurio/common-ui-components";
+
+describe("useReauthenticationService", () => {
+    const service = getReauthenticationService();
+
+    // Helper mock for AuthService
+    const mockAuth = {
+        isOidcAuthEnabled: () => true,
+        login: vi.fn().mockResolvedValue(undefined),
+    } as unknown as AuthService;
+
+    beforeEach(() => {
+        service.cancelReauthentication();
+    });
+
+    it("should initially not be pending", () => {
+        expect(service.isReauthenticationPending()).toBe(false);
+    });
+
+    it("should become pending when requesting re-authentication and invoke the interceptor", async () => {
+        const interceptor = vi.fn().mockResolvedValue(true);
+        const unregister = service.registerReauthenticationInterceptor(interceptor);
+
+        const promise = service.requestReauthentication(mockAuth);
+        expect(service.isReauthenticationPending()).toBe(true);
+
+        const result = await promise;
+        expect(result).toBe(true);
+        expect(interceptor).toHaveBeenCalledTimes(1);
+        expect(service.isReauthenticationPending()).toBe(true); // remains pending until explicitly dismissed
+
+        unregister();
+    });
+
+    it("should cancel pending state when interceptor is unregistered while pending", async () => {
+        const interceptor = vi.fn().mockResolvedValue(true);
+        const unregister = service.registerReauthenticationInterceptor(interceptor);
+
+        const promise = service.requestReauthentication(mockAuth);
+        expect(service.isReauthenticationPending()).toBe(true);
+
+        // Unregister the interceptor while re-auth is pending
+        unregister();
+
+        expect(service.isReauthenticationPending()).toBe(false);
+        await promise;
+    });
+
+    it("should trigger redirect if no interceptor is registered", async () => {
+        const redirectMockAuth = {
+            isOidcAuthEnabled: () => true,
+            login: vi.fn().mockResolvedValue(undefined),
+        } as unknown as AuthService;
+
+        const promise = service.requestReauthentication(redirectMockAuth);
+        expect(service.isReauthenticationPending()).toBe(true);
+
+        await promise;
+        expect(redirectMockAuth.login).toHaveBeenCalledTimes(1);
+    });
+});
--- a/ui/ui-app/src/services/useReauthenticationService.test.ts
+++ b/ui/ui-app/src/services/useReauthenticationService.test.ts
@@ -1,9 +1,10 @@
-import { describe, expect, it, vi, beforeEach } from "vitest";
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
 import { getReauthenticationService } from "./useReauthenticationService";
 import { AuthService } from "@apicurio/common-ui-components";
 
 describe("useReauthenticationService", () => {
     const service = getReauthenticationService();
+    let cleanups: Array<() => void> = [];
 
     // Helper mock for AuthService
     const mockAuth = {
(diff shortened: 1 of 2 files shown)


vivek-gite:
@kunaldevsahu 
 I confirmed the reported stale-pending-state path. One concern with the proposed fix:
`cancelReauthentication()` clears the shared state, but it does not invalidate an already-running `requestReauthentication()` task.
 If the unregistered interceptor is still awaiting completion, that stale task can later resolve `false` or throw and start the OIDC redirect anyway. If that old redirect then fails, its unconditional cleanup can also clear `isPending` and `requestPromise` for a newer re-authentication request started after unregistration.
The added tests use an interceptor that immediately resolves `true`, so they do not cover this lifecycle. Could the implementation associate each request with a generation/token (or verify that its promise is still the active one after each `await`) before starting a redirect or clearing state?
Suggested regression test: start request A with a deferred interceptor, unregister it, start request B, then resolve or reject A. Assert that A neither redirects nor changes B’s pending state.

ui/ui-app/src/services/useReauthenticationService.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        const unregister = service.registerReauthenticationInterceptor(interceptor);
+
+        const promise = service.requestReauthentication(mockAuth);
+        expect(service.isReauthenticationPending()).toBe(true);
+
+        // Unregister the interceptor while re-auth is pending
+        unregister();
+
+        expect(service.isReauthenticationPending()).toBe(false);
+        await promise;
```
paoloantinori: This is the central regression test, but it asserts `isReauthenticationPending()` is `false` on the line above (right after `unregister()`) and then never re-checks after `await promise`. The runtime scenario this fix targets is the in-flight flow settling after the component unmounts, so consider adding `expect(service.isReauthenticationPending()).toBe(false)` after this `await`. Current semantics suggest completion does not re-arm pending (the normal flow stays pending because it is never cleared, not because completion re-sets it), so this would pass today — but the assertion locks the guarantee in and guards against a future `.then()`/finally that re-arms pending, which the pre-await check alone would not catch.

ui/ui-app/src/services/useReauthenticationService.test.ts
```diff
@@ -0,0 +1,63 @@
...
+describe("useReauthenticationService", () => {
+    const service = getReauthenticationService();
+
+    // Helper mock for AuthService
+    const mockAuth = {
+        isOidcAuthEnabled: () => true,
+        login: vi.fn().mockResolvedValue(undefined),
+    } as unknown as AuthService;
+
+    beforeEach(() => {
```
paoloantinori, line 15: The service is a module-level singleton, and this `beforeEach` only calls `cancelReauthentication()` — it never clears a registered interceptor. If an earlier test throws before reaching its `unregister()` (e.g., the tests at lines ~22-34 or ~37-45), the leaked interceptor persists into the "should trigger redirect if no interceptor is registered" test (line 51), which would then fail confusingly (login never called). Consider tracking the unregister handles in an array and invoking them in an `afterEach`, or resetting the interceptor in `beforeEach`, so test isolation doesn't depend on every `it` completing cleanly.

paoloantinori (commented):
The source fix is minimal, correctly scoped to the active-interceptor branch, and independent correctness review found no bug (request completion does not re-arm `pending`, and there is no double-cancel risk); the only worthwhile suggestions are two test-hardening items in the new test file.

- `ui/ui-app/src/services/useReauthenticationService.ts:111-113` — the fix is sound: it clears pending only when the unregistering interceptor is still the active one, so replacement is correctly excluded and the existing flow is preserved.
- `ui/ui-app/src/services/useReauthenticationService.test.ts:47-48` — the key regression test asserts `isReauthenticationPending()` is `false` immediately after `unregister()` but before `await promise`, and never re-checks after the in-flight flow settles, which is the runtime scenario this fix targets. A one-line post-settle assertion would lock the guarantee in.
- `ui/ui-app/src/services/useReauthenticationService.test.ts:14-16` — `beforeEach` only cancels the pending flag on this singleton and never clears a registered interceptor, so a mid-test failure would leak an interceptor into the later "no interceptor" test (line 51).

ui/ui-app/src/services/useReauthenticationService.test.ts
kunaldevsahu: Thanks @paoloantinori  I'll update the test setup to ensure interceptors are always cleaned up between tests.

ui/ui-app/src/services/useReauthenticationService.ts
```diff
@@ -108,6 +108,9 @@ function createReauthenticationService(): ReauthenticationService {
             return () => {
                 if (interceptor === nextInterceptor) {
                     interceptor = undefined;
+                    if (isPending) {
+                        cancelReauthentication();
```
paoloantinori, line 112: Worth confirming: when this `cancelReauthentication()` runs because the interceptor's owner (e.g. the Editor page) unmounted mid-flow, does it also settle the in-flight `requestReauthentication` promise? The interceptor's own returned promise will never resolve once the modal backing it is gone, so if cancellation only clears the `isPending` flag, any caller awaiting `requestReauthentication` (such as a 401 response handler) would hang — even though the next 401 can now start a new flow. If `cancelReauthentication` is the normal end-of-flow dismissal that already resolves/rejects the promise, this is fine; just want to make sure unregistering doesn't leave an awaiter dangling.

ui/ui-app/src/services/useReauthenticationService.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        expect(service.isReauthenticationPending()).toBe(true);
+
+        const result = await promise;
+        expect(result).toBe(true);
+        expect(interceptor).toHaveBeenCalledTimes(1);
+        expect(service.isReauthenticationPending()).toBe(true); // remains pending until explicitly dismissed
+    });
+
+    it("should cancel pending state when interceptor is unregistered while pending", async () => {
+        const interceptor = vi.fn().mockResolvedValue(true);
```
paoloantinori: This regression test uses `vi.fn().mockResolvedValue(true)`, so the interceptor resolves immediately — by the time `unregister()` runs the flow isn't truly in-flight, which is the scenario the bug describes. To cover it more faithfully, use an interceptor that returns a pending (never-resolving) promise, call `unregister()`, then assert a subsequent `service.requestReauthentication(...)` can start a fresh flow (e.g. invokes a new interceptor). The current test only checks `isReauthenticationPending()` flips to false, which is a weaker proxy for the reported 'subsequent 401s don't trigger re-auth' symptom.

paoloantinori (commented):
Overall the fix is correct and well-scoped: clearing the pending flag when the active interceptor is unregistered directly resolves the reported stuck-pending state, and the `interceptor === nextInterceptor` staleness guard combined with the `isPending` check make the new code safe. Two things are worth a look before merge.

- `ui/ui-app/src/services/useReauthenticationService.ts:111-113` — when the interceptor is unregistered mid-flight (e.g. the Editor unmounts while its re-auth modal is open), the in-flight `requestReauthentication` promise will not settle on its own because the interceptor backing it is gone. Worth confirming that `cancelReauthentication()` also settles that promise (resolves/rejects), not just flips the `isPending` flag, so any caller awaiting it (such as a 401 handler) is unblocked rather than left hanging.
- `ui/ui-app/src/services/useReauthenticationService.test.ts:50,60` — the regression test uses `mockResolvedValue(true)`, so the interceptor settles immediately and the genuinely-in-flight case (interceptor still pending at unregister time) is not exercised. A stronger test would use a never-resolving interceptor and then assert that a subsequent `requestReauthentication` can start a fresh flow — the actual reported symptom — rather than only asserting `isPending` flips to false.

ui/ui-app/src/services/useReauthenticationService.ts
kunaldevsahu: Thanks for pointing this out! I verified that cancelReauthentication() also settles the pending requestReauthentication() promise, so callers awaiting it are not left hanging.

EricWittmann:
@vivek-gite Good catch on the stale-promise-resolution race. Here's our analysis:

You're right that after `cancelReauthentication()` runs, the old IIFE from Request A is still executing in its own async context — `cancelReauthentication()` clears `isPending` and `requestPromise`, but it doesn't abort the in-flight async function. So if Request A's interceptor eventually resolves with `false`, the IIFE would fall through to `beginOidcLoginRedirect()` even though the flow was "cancelled." If that redirect fails, the error handler at lines 68-70 would clear `isPending` and `requestPromise` — potentially clobbering Request B's state if a new flow started in the meantime.

That said, we think this is acceptable for this PR:

1. **It's not a regression.** The race existed before this PR — the interceptor was already being unregistered on unmount, the only difference is that `isPending` wasn't being cleared (which is the actual reported bug).
2. **In practice, the editor interceptor resolves immediately with `true`** (it just opens a modal and returns), so the IIFE never reaches `beginOidcLoginRedirect`. The stale resolution would have to come from a future interceptor that uses a long-lived promise.
3. **Properly fixing the race requires a generation/cancellation-token approach** — each `requestReauthentication` call would need to check that its generation is still the active one before modifying shared state. That's a meaningful refactor of the service, not a 3-line fix.

This PR fixes the concrete, reproducible bug (stuck pending state → subsequent 401s silently ignored). The stale-resolution concern is a deeper design hardening item. If you think it's worth tracking beyond a theoretical edge case, feel free to open a separate issue for it and we can scope the generation-token approach there.

EricWittmann:
Thanks @kunaldevsahu! Clean fix for a real edge case, and the test coverage improved nicely through the review rounds. Appreciate the thoroughness.

PR #9099, opened 2026-07-31
feat(ci): issue self assignment via comments

### LINKED ISSUE #9061: feat(ci): issue self-assignment via comments

senutpal:
Implemented a dedicated GitHub Actions workflow and robust Node.js handler for issue self-assignment:

- `.github/workflows/issue-assignment.yml`: Triggers on `issue_comment: created` events specifically targeting open non-PR issues (`!github.event.issue.pull_request`) with trimmed slash command detection and per-user concurrency serialization (`group: issue-assignment-user-${{ github.event.comment.user.login }}`).
- `.github/scripts/issue-assignment.js`: Handles `/assign-me` and `/unassign-me` slash commands (along with `/claim` and `/unassign` aliases).
  - Uses `(?:\s|$)` whitespace boundaries for command regex to prevent matching hyphenated commands (e.g., `/assign-foo`).
  - Fetches live issue data (`github.rest.issues.get`) to prevent stale payload snapshot race conditions.
  - Normalizes username comparisons to lowercase to support case-insensitive GitHub logins.
  - Uses `github.paginate` for repository issue queries to prevent truncation when counting open assigned issues (>100 items limit).
  - Wraps mutating API calls (`addAssignees`, `removeAssignees`) and notification comments (`postComment`) in `try/catch` blocks with `core.warning()` fallbacks.
  - Verifies actual assignment status after `addAssignees` call to gracefully inform external non-collaborator users when GitHub API permissions prevent direct self-assignment.
  - Restricts contributors to a maximum of 3 concurrently assigned open issues across the repository (`MAX_ASSIGNED_ISSUES = 3`).
  - Ignores bot user comments (`[bot]`, `renovate`, `dependabot`, `apicurio-ci`).
- `.github/PR_LIFECYCLE.md` & `CONTRIBUTING.md`: Documented `/assign-me` and `/unassign-me` slash commands under available contributor guidelines and issue claiming procedures.

Resolves #9061

### DIFF
--- a/.github/scripts/issue-assignment.js
+++ b/.github/scripts/issue-assignment.js
@@ -83,20 +83,20 @@ async function handleIssueComment({ github, context, core }) {
   const assigneeLogins = currentAssignees.map(a => a.login.toLowerCase());
 
   if (assignMatch) {
-    await handleAssign(github, core, owner, repo, issueNumber, commenterOriginal, commenter, assigneeLogins);
+    await handleAssign(github, core, owner, repo, issueNumber, commenterOriginal, commenter, currentAssignees, assigneeLogins);
   } else if (unassignMatch) {
     await handleUnassign(github, core, owner, repo, issueNumber, commenterOriginal, commenter, assigneeLogins);
   }
 }
 
-async function handleAssign(github, core, owner, repo, issueNumber, commenterOriginal, commenter, assigneeLogins) {
+async function handleAssign(github, core, owner, repo, issueNumber, commenterOriginal, commenter, currentAssignees, assigneeLogins) {
   // If already assigned
   if (assigneeLogins.length > 0) {
     if (assigneeLogins.includes(commenter)) {
       core.info(`User ${commenterOriginal} is already assigned to issue #${issueNumber}.`);
       await postComment(github, owner, repo, issueNumber, `@${commenterOriginal} You are already assigned to this issue.`);
     } else {
-      const mentionList = assigneeLogins.map(l => `@${l}`).join(', ');
+      const mentionList = currentAssignees.map(a => `@${a.login}`).join(', ');
       core.info(`Issue #${issueNumber} is already assigned to ${mentionList}.`);
       await postComment(github, owner, repo, issueNumber, `This issue is already claimed by ${mentionList}.`);
     }
--- a/.github/scripts/issue-assignment.js
+++ b/.github/scripts/issue-assignment.js
@@ -1,7 +1,7 @@
 // Issue Assignment Handler
 //
 // Automatically assigns or unassigns contributors to issues when they comment with trigger commands.
-// Commands: /assign-me, /unassign-me (aliases: /claim, /assign, /unassign)
+// Commands: /assign-me, /unassign-me (aliases: /claim, /unassign)
 
 const MAX_ASSIGNED_ISSUES = 3;
 
@@ -75,7 +75,7 @@ async function handleIssueComment({ github, context, core }) {
   // Verify issue is open
   if (freshIssue.state !== 'open') {
     core.info(`Issue #${issueNumber} is closed. Cannot modify assignment.`);
-    await postComment(github, owner, repo, issueNumber, `@${commenterOriginal} Issue #${issueNumber} is closed and cannot be assigned.`);
+    await postComment(github, core, owner, repo, issueNumber, `@${commenterOriginal} Issue #${issueNumber} is closed and cannot be assigned.`);
     return;
   }
 
@@ -94,11 +94,11 @@ async function handleAssign(github, core, owner, repo, issueNumber, commenterOri
   if (assigneeLogins.length > 0) {
     if (assigneeLogins.includes(commenter)) {
       core.info(`User ${commenterOriginal} is already assigned to issue #${issueNumber}.`);
-      await postComment(github, owner, repo, issueNumber, `@${commenterOriginal} You are already assigned to this issue.`);
+      await postComment(github, core, owner, repo, issueNumber, `@${commenterOriginal} You are already assigned to this issue.`);
     } else {
       const mentionList = currentAssignees.map(a => `@${a.login}`).join(', ');
       core.info(`Issue #${issueNumber} is already assigned to ${mentionList}.`);
-      await postComment(github, owner, repo, issueNumber, `This issue is already claimed by ${mentionList}.`);
+      await postComment(github, core, owner, repo, issueNumber, `This issue is already claimed by ${mentionList}.`);
     }
     return;
   }
@@ -125,6 +125,7 @@ async function handleAssign(github, core, owner, repo, issueNumber, commenterOri
     core.info(`User ${commenterOriginal} reached max open issues limit (${actualOpenIssues.length}/${MAX_ASSIGNED_ISSUES}).`);
     await postComment(
       github,
+      core,
       owner,
       repo,
       issueNumber,
@@ -150,6 +151,7 @@ async function handleAssign(github, core, owner, repo, issueNumber, commenterOri
     core.warning(`Failed to assign ${commenterOriginal} to issue #${issueNumber}: ${error.message}`);
     await postComment(
       github,
+      core,
       owner,
       repo,
       issueNumber,
@@ -167,6 +169,7 @@ async function handleAssign(github, core, owner, repo, issueNumber, commenterOri
     core.warning(`User ${commenterOriginal} could not be assigned to issue #${issueNumber}. User may lack collaborator permissions.`);
     await postComment(
       github,
+      core,
       owner,
       repo,
       issueNumber,
@@ -177,6 +180,7 @@ async function handleAssign(github, core, owner, repo, issueNumber, commenterOri
 
   await postComment(
     github,
+    core,
     owner,
     repo,
     issueNumber,
@@ -187,7 +191,7 @@ async function handleAssign(github, core, owner, repo, issueNumber, commenterOri
 async function handleUnassign(github, core, owner, repo, issueNumber, commenterOriginal, commenter, assigneeLogins) {
   if (!assigneeLogins.includes(commenter)) {
     core.info(`User ${commenterOriginal} attempted to unassign from issue #${issueNumber} but is not assigned.`);
-    await postComment(github, owner, repo, issueNumber, `@${commenterOriginal} You are not currently assigned to this issue.`);
+    await postComment(github, core, owner, repo, issueNumber, `@${commenterOriginal} You are not currently assigned to this issue.`);
     return;
   }
 
@@ -203,6 +207,7 @@ async function handleUnassign(github, core, owner, repo, issueNumber, commenterO
     core.warning(`Failed to unassign ${commenterOriginal} from issue #${issueNumber}: ${error.message}`);
     await postComment(
       github,
+      core,
       owner,
       repo,
       issueNumber,
@@ -211,16 +216,20 @@ async function handleUnassign(github, core, owner, repo, issueNumber, commenterO
     return;
   }
 
-  await postComment(github, owner, repo, issueNumber, `@${commenterOriginal} You have been unassigned from this issue.`);
+  await postComment(github, core, owner, repo, issueNumber, `@${commenterOriginal} You have been unassigned from this issue.`);
 }
 
-async function postComment(github, owner, repo, issueNumber, body) {
-  await github.rest.issues.createComment({
-    owner,
-    repo,
-    issue_number: issueNumber,
-    body,
-  });
+async function postComment(github, core, owner, repo, issueNumber, body) {
+  try {
+    await github.rest.issues.createComment({
+      owner,
+      repo,
+      issue_number: issueNumber,
+      body,
+    });
+  } catch (error) {
+    core.warning(`Failed to post comment on issue #${issueNumber}: ${error.message}`);
+  }
 }
 
 module.exports = {
--- a/.github/scripts/issue-assignment.js
+++ b/.github/scripts/issue-assignment.js
@@ -1,0 +1,203 @@
+// Issue Assignment Handler
+//
+// Automatically assigns or unassigns contributors to issues when they comment with trigger commands.
+// Commands: /assign-me, /unassign-me (aliases: /claim, /assign, /unassign)
+
+const MAX_ASSIGNED_ISSUES = 3;
+
+const KNOWN_BOTS = new Set([
+  'renovate[bot]',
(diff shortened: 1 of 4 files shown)


.github/workflows/issue-assignment.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  assign:
+    name: Issue Assignment
+    runs-on: ubuntu-latest
+    if: |
+      github.repository_owner == 'Apicurio' &&
+      github.event_name == 'issue_comment' &&
+      github.event.action == 'created' &&
+      !github.event.issue.pull_request &&
+      startsWith(trim(github.event.comment.body), '/')
```
paoloantinori: `trim()` is not a documented GitHub Actions expression function (the string functions are `contains`, `startsWith`, `endsWith`, `format`). If so, this job-level `if:` fails to evaluate and the `assign` job never runs — i.e. self-assignment never fires. The script already trims the comment and matches the slash command precisely, so the gate can be simplified to `startsWith(github.event.comment.body, '/')`. Worth confirming `trim` hasn't been added to the Actions expression set recently, but removing it is the safe call.

paoloantinori (commented):
The handler is well-built — bot filtering, case-insensitive login comparison, paginated limit checks, and the non-collaborator fallback are all handled correctly — but the workflow's job-level `if:` calls `trim()`, which is not a valid GitHub Actions expression function, so the condition likely never evaluates and the `assign` job never runs. Fix that single line and the feature should work.

- **`.github/workflows/issue-assignment.yml:24`** — `startsWith(trim(github.event.comment.body), '/')`: `trim()` is not among GitHub Actions' expression functions (`contains`, `startsWith`, `endsWith`, `format`, `join`, `toJSON`, `fromJSON`, `hashFiles`). An unknown function in a job `if:` fails to evaluate, so the `assign` job never runs and self-assignment silently does nothing. The script already trims the comment and matches the command precisely, so the gate can safely become `startsWith(github.event.comment.body, '/')`. Worth a quick check against the current Actions expression reference to confirm `trim` hasn't been newly added, but dropping it is the safe fix.
- **`.github/scripts/issue-assignment.js:124`** *(minor, low-likelihood)* — the "max 3 open issues" check is a non-atomic read-then-write, and the concurrency group only serializes comments on the *same* issue. Two near-simultaneous `/assign-me` comments by the same user on *different* issues could each read a count under 3 and both succeed, briefly exceeding the cap. Not worth blocking on; the easy mitigation is a one-line note in the docs qualifying the cap, or a post-`addAssignees` recount with rollback.

.github/workflows/issue-assignment.yml
senutpal: Fixed in commit 1a13eeac1. Thanks @paoloantinori

.github/scripts/issue-assignment.js
```diff
@@ -0,0 +1,203 @@
...
+  let freshIssue;
+  try {
+    const response = await github.rest.issues.get({
+      owner,
+      repo,
+      issue_number: issueNumber,
+    });
+    freshIssue = response.data;
+  } catch (error) {
+    core.warn(`Could not fetch live issue #${issueNumber}: ${error.message}. Falling back to payload issue data.`);
```
paoloantinori: `core.warn(...)` is not a function. `@actions/core` (the `core` object passed by `actions/github-script`) exports `warning`, not `warn`, so `core.warn` is `undefined` and this line throws `TypeError: core.warn is not a function`. Because it sits inside the `catch` for `github.rest.issues.get`, any transient failure of that call (rate limit, 5xx, network blip) will crash the handler instead of falling back to the payload snapshot (`freshIssue = issue`) as the surrounding code intends. Use `core.warning(...)`. The same misuse appears again at line 153.

.github/scripts/issue-assignment.js
```diff
@@ -0,0 +1,203 @@
...
+    assignees: [commenterOriginal],
+  });
+
+  // Verify contributor was actually assigned (external non-collaborator users cannot be assigned directly by GH API)
+  const isAssigned = (updatedIssue.assignees || []).some(
+    a => a.login.toLowerCase() === commenter
+  );
+
+  if (!isAssigned) {
+    core.warn(`User ${commenterOriginal} could not be assigned to issue #${issueNumber}. User may lack collaborator permissions.`);
```
paoloantinori: Same `core.warn` issue: `@actions/core` has no `warn` method, so this throws `TypeError: core.warn is not a function`. This is the non-collaborator branch — when `addAssignees` silently drops the user (`isAssigned === false`), the helpful `postComment` immediately below never executes and the workflow fails instead of informing the user. Use `core.warning(...)` here too.

paoloantinori (commented):
Well-structured feature with one runtime defect that will crash the handler on two specific paths: `core.warn(...)` is used where `@actions/core` only exposes `core.warning(...)`, so both calls throw `TypeError` at runtime.

- `.github/scripts/issue-assignment.js:71` and `:153` call `core.warn(...)`, but `core` is the `@actions/core` module passed by `actions/github-script`, which has no `warn` method (it's `warning`). Since `core.warn` is `undefined`, each call throws `TypeError: core.warn is not a function`. At line 71 this defeats the intended graceful fallback — a transient failure of `github.rest.issues.get` (rate limit, 5xx) crashes the step instead of falling back to the payload snapshot. At line 153 it breaks the non-collaborator path: the helpful "you need collaborator access" comment is never posted and the run fails. Fix: rename to `core.warning(...)` at both sites.
- Lower-severity note: the per-user 3-issue cap (`issue-assignment.js:124`) is a best-effort check, not strictly enforced. The concurrency group (`issue-assignment.yml:9`) is keyed only on `issue.number`, so concurrent `/assign-me` comments on different issues each read the same pre-increment count and can all pass, letting a contributor exceed 3. This is benign (over-assignment self-corrects as issues close), but the documented "maximum of 3" limit can be transiently exceeded under normal usage.

.github/scripts/issue-assignment.js
senutpal: Fixed in commit 320c55e02.

.github/scripts/issue-assignment.js
senutpal: Fixed in commit 320c55e02.

paoloantinori:
Confirmed — the `core.warning(...)` fix in `320c55e0` lands on both paths (the live-issue fetch fallback and the non-collaborator assignment branch). Scanning the handler, there are no remaining `core.warn(...)` calls, so the runtime defect is fully resolved.

One minor, non-blocking note while I'm here: in the fetch-fallback `core.warning(...)` the message still uses single quotes (`'...${issueNumber}...'`), so the `${...}` placeholders won't interpolate there — it'll log the literal string. Harmless (it's a best-effort warning), but a backtick template would make it readable. Not worth blocking on.

carlesarnal:
## Code Review

Went through the full PR. The two issues paoloantinori flagged (trim in workflow expression, core.warn to core.warning) are both fixed.

**Checked:** Minimal permissions, pinned action SHAs, startsWith gate, concurrency group, fork guard, bot filtering, command regex boundaries, fresh issue fetch with fallback, core.warning at lines 71 and 153, paginated 3-issue cap query, post-addAssignees non-collaborator verification, defense-in-depth PR filtering, style consistency with pr-lifecycle.js, documentation updates.

**Minor non-blocking note:** issue-assignment.js:99 uses lowercased logins in mentionList (from assigneeLogins). Works fine but original-case would be more natural.

All previously flagged issues resolved. No blocking issues found.

.github/scripts/issue-assignment.js
```diff
@@ -0,0 +1,203 @@
...
+      `@${commenterOriginal} You currently have ${actualOpenIssues.length} open issue(s) assigned to you. ` +
+      `Contributors are limited to ${MAX_ASSIGNED_ISSUES} open issues at a time. ` +
+      `Please complete or unassign yourself from existing issues before claiming a new one.`
+    );
+    return;
+  }
+
+  // Assign contributor
+  core.info(`Assigning ${commenterOriginal} to issue #${issueNumber}.`);
+  const { data: updatedIssue } = await github.rest.issues.addAssignees({
```
paoloantinori: `github.rest.issues.addAssignees` here (and `removeAssignees` in `handleUnassign`) are not wrapped in try/catch. If the call throws — a transient 5xx, secondary rate limit, or for a non-collaborator possibly a 4xx — the `actions/github-script` step fails with a red run and the commenter gets no reply at all, which defeats the bot's purpose of immediate feedback. Separately, worth confirming the assumption behind the verify-after-assign block at line 147: it assumes GitHub returns 200 and silently omits a non-assignable user rather than throwing. If `addAssignees` throws for non-collaborators, the `isAssigned` fallback is unreachable and that graceful message is dead code. Wrapping the assign call in try/catch and posting the "a maintainer will assign this manually" message on failure makes the handler robust to both behaviors and to transient errors.

paoloantinori (commented):
Overall this is a well-constructed change: the handler thoughtfully handles stale payload data, case-insensitive logins, pagination, PR exclusion, bot filtering, and re-entrancy (the bot's own comments don't start with `/`, actions are SHA-pinned, token scope is minimal, and PR comments are filtered at both layers). The main gap is error handling around the mutating API calls.

- `.github/scripts/issue-assignment.js:140` — `addAssignees` (and `removeAssignees` at `:176`) are not wrapped in try/catch; a thrown error fails the run with no reply to the commenter, and the verify-after-assign fallback assumes silent omission rather than a throw. See inline comment.
- `.github/scripts/issue-assignment.js:124` — the 3-issue cap is an optimistic check with no cross-issue lock. Because the concurrency group is keyed per issue (`issue-assignment.yml:12`), two near-simultaneous `/assign-me` comments on *different* issues can both pass the `< 3` check and land the user at 4; it self-corrects on the next attempt but briefly exceeds the documented limit. Low impact, flagging since the cap is stated as a hard rule in `CONTRIBUTING.md`.
- `.github/workflows/issue-assignment.yml:12` / `issue-assignment.js:107` — no per-user throttle, so any user posting `/assign-me` across many open issues spawns one run plus one bot reply per issue. Inherent to comment-triggered bots, but worth noting for Actions-minute and comment-noise cost.
- Minor UX: the `/assign` alias means `/assign @someone` self-assigns the commenter rather than the named user, which could surprise maintainers expecting to delegate.

.github/scripts/issue-assignment.js
senutpal: Fixed in commit f1cb4e856.

jsenko (commented):
Review on head `ab14a94`. Well-implemented feature — bot filtering, case-insensitive logins, paginated limit check, non-collaborator fallback, and post-assignment verification are all handled correctly. All 3 rounds of Copilot feedback addressed (trim() removed, core.warning fixed, try/catch added, /assign alias removed).

**1. Stale comment lists removed alias (should-fix)**

`issue-assignment.js:4`:
```js
// Commands: /assign-me, /unassign-me (aliases: /claim, /assign, /unassign)
```

`/assign` was removed from the regex in commit `ab14a94` but the header comment still lists it as an alias. Should be:
```js
// Commands: /assign-me, /unassign-me (aliases: /claim, /unassign)
```

No other findings. Clean implementation.

jsenko (commented):
## Automated review

Solid implementation — defensive coding (live issue data fetch, post-assignment verification, pagination for >100 issues), clear user-facing messages, and appropriate permissions scoping.

### should-fix: PR body claims `/assign` alias but code doesn't support it

The PR description says "along with `/claim`, `/assign`, `/unassign` aliases" but the assign regex at line 49 only matches `/assign-me` and `/claim`:

```javascript
const assignMatch = /^\/(assign-me|claim)(?:\s|$)/i.test(firstLine);
```

The docs (PR_LIFECYCLE.md, CONTRIBUTING.md) correctly list only `/assign-me` and `/claim`. Either add `/assign` to the regex or remove it from the PR description. Note: `/unassign` IS correctly included in the unassign regex.

### should-fix: Race condition in assignment limit enforcement

The concurrency group is per-issue (`issue-assignment-${{ github.event.issue.number }}`), not per-user. If a contributor comments `/assign-me` on two different issues simultaneously, both workflows pass the 3-issue limit check before either completes the assignment, allowing 4+ concurrent assignments.

Options:
1. Change concurrency group to `issue-assignment-user-${{ github.event.comment.user.login }}` (serializes per user but may queue legitimate unassign operations)
2. Accept as a known limitation and document it (simultaneous comments are rare in practice)

### should-fix: `postComment` calls lack error handling

All `postComment()` calls are fire-and-forget. If the GitHub API is rate-limited or down, the assignment may succeed but the user gets no feedback. Wrap in try/catch with `core.warning()` fallback — best-effort notification is fine.

### nice-to-have: Redundant workflow trigger conditions

Lines 21-22 of `issue-assignment.yml` check `github.event_name == 'issue_comment' && github.event.action == 'created'` but the `on:` trigger already constrains to exactly that. Matches existing `pr-lifecycle.yml` style, so not a blocker — just cleanup.

jsenko (approved):
LGTM, thanks!

jsenko:
Merged, thank you!

PR #9104, opened 2026-07-31
fix(compatibility): catch SchemaException in JsonSchemaDiffLibrary

### LINKED ISSUE #9079: 500 Internal Server Error during JSON Schema compatibility check for drafts (2020-12 / 2019-09)
## Description

**Registry Version**: snapshot (main branch)
**Persistence type**: in-memory (Quarkus dev mode)

**Steps to Reproduce:**
1. Create a new artifact of type `JSON` with a valid JSON schema (e.g., Draft 2020-12). *(This succeeds because `jsonsKema` handles 2020-12 correctly).*
2. Attempt to create a **new version** for the same artifact by uploading a modified 2020-12 schema.
3. The UI shows an "Error creating artifact version" popup and the backend throws a 500 error.

**Root Cause:**
When a new version is uploaded, the `CompatibilityRuleExecutor` runs and invokes `JsonSchemaDiffLibrary.findDifferences(...)` to check for backward compatibility. 
Inside `JsonSchemaDiffLibrary.java`, the `SchemaLoader.builder()` from the `everit-org/json-schema` library is hardcoded. Since Everit does not currently support drafts like 2020-12 or 2019-09, it throws an `org.everit.json.schema.SchemaException` (`Unknown MetaSchema` or `could not determine version`). 

This exception is not caught inside `JsonSchemaDiffLibrary` (only `JsonProcessingException` is caught), which causes the unhandled exception to bubble up and result in a 500 Internal Server Error instead of returning a proper `422 (Unprocessable Entity)` or `409 (Conflict)` to the client.

**Proposed Solution:**
Update `JsonSchemaDiffLibrary.java` to catch `SchemaException` (and `IllegalArgumentException` from the parser) inside the `findDifferences` method, and wrap it in an `UnprocessableSchemaException`. This will be correctly mapped to a `4xx` error by the HTTP mapper logic.

Charul192:
Fixes #9079

This PR catches the `SchemaException` thrown by the `everit-org/json-schema` library in `JsonSchemaDiffLibrary` when it encounters unsupported schemas (e.g. Draft 2020-12) during compatibility checks, and wraps it in an `UnprocessableSchemaException` so it properly maps to a `422 Unprocessable Entity` response instead of a `500 Internal Server Error`.


paoloantinori (commented):
Wrapping everit's `SchemaException` in `UnprocessableSchemaException` is a reasonable, minimal way to convert the #9079 500 into a 422 for schemas the compatibility checker can't handle; the new code preserves the cause and compiles cleanly. Two things are worth confirming before this fully resolves #9079:

- Correctness crux at `JsonSchemaDiffLibrary.java:62` — confirm `SchemaException` is the exact type thrown for a Draft 2020-12 schema in the issue's repro. everit-org/json-schema can treat an unrecognized `$schema` as a default dialect rather than failing during load; if the 2020-12 schema loads and then trips the diff visitor, this catch never fires and the 500 persists.
- Message breadth at `JsonSchemaDiffLibrary.java:63` — `SchemaException` is also thrown for malformed schemas and broken `$ref` resolution unrelated to spec version, so the fixed string `"Unsupported JSON Schema version..."` will misdiagnose those cases. A more general message (or narrowing the catch) would avoid misleading users.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/jsonschema/JsonSchemaDiffLibrary.java
```diff
@@ -57,6 +59,8 @@ public static DiffContext findDifferences(String original, String updated,
 
         } catch (JsonProcessingException e) {
             throw new IllegalStateException(e);
+        } catch (SchemaException e) {
```
paoloantinori, line 62: Worth verifying against the #9079 stack trace that `SchemaException` is the exact exception type thrown for a Draft 2020-12 schema. everit-org/json-schema sometimes treats an unrecognized `$schema` as a default dialect instead of failing during load — if the 2020-12 schema loads and then fails later inside the diff visitor (with a different exception type), this `catch` won't fire and the 500 will persist, leaving the bug unfixed. If you have the repro handy, a quick check that the load itself throws `SchemaException` would confirm the fix is complete.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/jsonschema/JsonSchemaDiffLibrary.java
```diff
@@ -57,6 +59,8 @@ public static DiffContext findDifferences(String original, String updated,
...
+            throw new UnprocessableSchemaException("Unsupported JSON Schema version for compatibility check: " + e.getMessage(), e);
```
paoloantinori: `SchemaException` is thrown by everit for more than just unsupported spec versions — malformed schemas and broken `$ref` resolution also raise it. Catching it broadly and labeling it `"Unsupported JSON Schema version"` will misdiagnose those unrelated failures as version problems. Consider a more general message (e.g. "Schema could not be processed for compatibility check") or narrowing the catch to the spec-version case so the diagnosis stays accurate.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/jsonschema/JsonSchemaDiffLibrary.java
Charul192: Yes, I verified with the Draft 2020-12 repro that the `SchemaException (could not determine version)` is thrown immediately during the schema loading phase (specifically at `schemaLoaderBuilder.schemaJson().build()`), well before it ever reaches the diff visitor. So this catch block properly intercepts the load failure!

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/jsonschema/JsonSchemaDiffLibrary.java
Charul192: You are absolutely right. I've updated the PR with a new commit to use the more general message `"Schema could not be processed for compatibility check: "` as you suggested, since `SchemaException` can indeed be thrown for other reasons like broken references.

paoloantinori:
Thanks @Charul192 — the updated message (`"Schema could not be processed for compatibility check: "`) reads well and correctly covers the broader set of `SchemaException` causes (broken references, unsupported drafts, etc.), not just the Draft 2020-12 case. LGTM.

Charul192:
Thanks @paoloantinori for the review! Glad the updated message looks good.
@EricWittmann @carlesarnal whenever you get a moment, would appreciate a look at this. 
Thanks!

paoloantinori:
Happy to help @Charul192. The change looks good from my side — wrapping in `UnprocessableSchemaException` with the broader message handles Draft 2020-12 and related `SchemaException` causes cleanly. I'll leave it to @EricWittmann / @carlesarnal for the maintainer sign-off.

jsenko (commented):
## Review

Fixes #9079 — catches `SchemaException` from everit-org/json-schema during compatibility checks and wraps it in `UnprocessableSchemaException` for a 422 response instead of 500.

The fix is correct and follows the established pattern (same approach used in `AvroCompatibilityChecker` and `XsdCompatibilityChecker`).

### should-fix: Missing test for the SchemaException catch path

Add a test in `JsonSchemaDiffUtilTest` that exercises the new catch block. Feed `JsonSchemaDiffLibrary.findDifferences()` a Draft 2020-12 schema (which `everit-org/json-schema` can't parse) and assert `UnprocessableSchemaException` is thrown:

```java
@Test
void findDifferencesThrowsUnprocessableForUnsupportedDraft() {
    String draft2020 = "{\"$schema\":\"https://json-schema.org/draft/2020-12/schema\",\"type\":\"object\"}";
    Assertions.assertThrows(UnprocessableSchemaException.class,
        () -> JsonSchemaDiffLibrary.findDifferences(draft2020, draft2020, CompatibilityDirection.BACKWARD));
}
```

This locks in the 422 behavior and prevents a future refactor from accidentally reverting to 500.

Charul192:
Thanks for the review @jsenko ! I've added the test in JsonSchemaDiffUtilTest. 
Since JsonSchemaDiffLibrary.findDifferences takes a Map<String, TypedContent> as the 3rd parameter instead of CompatibilityDirection, I passed Collections.emptyMap() to it. 
The test passes successfully

jsenko (commented):
Re-review of commit 602308a — author added the test requested by jsenko.

### Re-review delta

**Resolved:** Missing test for SchemaException catch path — added `findDifferencesThrowsUnprocessableForUnsupportedDraft` in `JsonSchemaDiffUtilTest`. Test is correct: feeds a Draft 2020-12 schema that triggers `SchemaException` in the Everit loader and asserts `UnprocessableSchemaException`.

**New findings:**

### should-fix: Import ordering and FQN usage in test

`JsonSchemaDiffUtilTest.java` lines 12-16: the new imports are appended after existing imports instead of being merged in alphabetical order. `io.apicurio.*` should come before `org.junit.*`, and `java.util.Collections` should be grouped with `java.util.stream.Stream`.

Also, line 37 uses `Assertions.assertThrows(...)` — use a static import `assertThrows` instead, consistent with the existing `assertEquals` static import on line 18.

No further issues found.

carlesarnal:
Please, address the conflict.

jsenko (approved):

jsenko:
hi @Charul192 I've fixed the merge conflicts, the PR is ready to merge, but I can't do it without your DCO sign-off. Please rebase you branch and update your commit messages to add the signature.

PR #9110, opened 2026-07-31
[3.4.x] Enhancement: Move version state filtering to server-side query parameter

Kayd-06:
## Description
This PR addresses an outstanding `FIXME` in the client codebase by moving the version state filtering logic from the client to the server, improving performance and API semantics. Previously, the `RegistryClientFacadeImpl` fetched all search results and then inefficiently applied a Java stream `.filter(version -> DISABLED != version.getState())` client-side. This change pushes that filtering down to the backend `SearchResource`.

## Changes Made
- **OpenAPI Specification (`openapi.json`)**: Added a new `state` query parameter to the `POST /search/versions` endpoint using the `VersionState` schema.
- **Backend API (`SearchResourceImpl.java`)**: Updated the implementation to accept the `state` parameter and append it to the backend `SearchFilter` logic (`SearchFilter.ofState(state)`).
- **Client SDK (`RegistryClientFacadeImpl.java`)**: Replaced the inefficient client-side Java stream filter with a direct query parameter binding (`config.queryParameters.state = VersionState.ENABLED;`).

## Verification
- Rebuilt the `schema-resolver` and `app` modules cleanly.
- Resolved Checkstyle requirements regarding import organization.


schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeImpl.java
```diff
@@ -141,11 +140,10 @@ public List<RegistryVersionCoordinates> searchVersionsByContent(String schemaStr
             config.queryParameters.orderby = VersionSortBy.GlobalId;
             config.queryParameters.order = SortOrder.Desc;
             config.queryParameters.limit = 100;
+            config.queryParameters.state = VersionState.ENABLED;
```
paoloantinori: The previous filter was `DISABLED != state` (keep anything that is not DISABLED), but `state=ENABLED` keeps only ENABLED. These are equivalent only if ENABLED and DISABLED are the sole version states; if `VersionState` also includes DEPRECATED (or any other state), DEPRECATED versions that used to match this content search are now excluded. If that narrowing is intentional, this is fine — but if the intent was to preserve the old behavior, the server filter needs to express `state != DISABLED` rather than `state == ENABLED` (worth checking whether `SearchFilter` supports a negated or multi-state form).

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeImpl.java
```diff
@@ -141,11 +140,10 @@ public List<RegistryVersionCoordinates> searchVersionsByContent(String schemaStr
...
         });
 
-        // FIXME consider moving the "filter by state" logic to the server as another query parameter
         return results.getVersions().stream()
```
paoloantinori, line 152: With the client-side `.filter(...)` gone, correctness now depends on the server understanding the new `state` param. An older registry will ignore the unknown query parameter (JAX-RS drops unrecognized query params by default) and return all versions including DISABLED, which the client no longer strips. If the resolver must remain compatible with older registries, consider keeping a defensive client-side filter or gating the new param behind a capabilities/version check.

paoloantinori (commented):
The change cleanly resolves the client-side FIXME by pushing version-state filtering into the `POST /search/versions` query, and the openapi.json, `SearchResourceImpl`, and `RegistryClientFacadeImpl` edits are mutually consistent. Two behavioral nuances are worth confirming before merge, both stemming from the move from a client-side guard to a server-side param.

- `schema-resolver/.../RegistryClientFacadeImpl.java:143` — the new `state=ENABLED` narrows matching versus the prior `state != DISABLED`; these differ for any state that is neither ENABLED nor DISABLED (e.g. DEPRECATED), so DEPRECATED versions that previously matched will now be dropped.
- `schema-resolver/.../RegistryClientFacadeImpl.java:146` — the client-side guard is removed entirely, so against an older registry that ignores the new `state` query param, DISABLED versions will now be returned to the resolver.

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeImpl.java
```diff
@@ -143,9 +142,10 @@ public List<RegistryVersionCoordinates> searchVersionsByContent(String schemaStr
             config.queryParameters.limit = 100;
         });
 
-        // FIXME consider moving the "filter by state" logic to the server as another query parameter
+        // We filter out DISABLED versions on the client side to preserve the old behavior (matching ENABLED and DEPRECATED).
+        // This also acts as a defensive guard against older registries that do not support the state query parameter.
```
paoloantinori: This comment (and the PR description) imply the client now sends the `state` query parameter, but the diff contains no `config.queryParameters.state = ...` assignment — the `.filter(v -> VersionState.DISABLED != v.getState())` below is unchanged in behavior. As written, the client still fetches all versions and filters client-side, so the new server-side `state` filter in `SearchResourceImpl` is never exercised by this path and the stated performance win isn't realized. Was the `config.queryParameters.state = VersionState.ENABLED` line accidentally dropped from this commit?

app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
```diff
@@ -364,6 +364,9 @@ public VersionSearchResults searchVersionsByContent(Boolean canonical, String ar
         if (!StringUtil.isEmpty(artifactId)) {
             filters.add(SearchFilter.ofArtifactId(artifactId));
         }
+        if (state != null) {
+            filters.add(SearchFilter.ofState(state));
```
paoloantinori: `SearchFilter.ofState(state)` combined with the single-valued `VersionState` query param looks like an equality filter, so it cannot express the original client semantics of "exclude DISABLED" (which returned both ENABLED and DEPRECATED). If the client is later wired to send `state=ENABLED`, DEPRECATED versions — currently returned — would be silently dropped, contradicting the client comment about "preserving the old behavior." Worth deciding the intended DEPRECATED handling explicitly, and there's currently no test covering this server-side filter path.

common/src/main/resources/META-INF/openapi.json
```diff
@@ -3981,6 +3981,14 @@
             },
             "in": "query"
           },
+          {
```
paoloantinori, line 3989: `state` was added only to the POST `/search/versions` (by-content) operation; the GET `searchVersions` overload (`SearchResourceImpl.java:341`) was not updated, so the two search endpoints now diverge in capability. If that's intentional (the resolver client only uses by-content), a brief note would help future readers; otherwise consider adding `state` to the GET path for consistency.

paoloantinori (commented):
The server-side wiring (OpenAPI param + `SearchFilter.ofState`) is internally consistent, but the client change does not match the PR description or its own new comment: the diff never sets `config.queryParameters.state`, so the client still filters client-side and the new server-side filter is never exercised by this client. The stated performance improvement isn't realized as written, and there is an unresolved semantics question around DEPRECATED versions.

- `schema-resolver/.../RegistryClientFacadeImpl.java:145-148` — The description claims the client-side filter was "replaced with a direct query parameter binding (`config.queryParameters.state = VersionState.ENABLED`)", and the new comment only makes sense if the client sends `state` ("defensive guard against older registries that do not support the state query parameter"). However, the diff contains no such assignment; the `.filter(v -> VersionState.DISABLED != v.getState())` is retained with behavior identical to before. The client still fetches all matching versions and filters in Java, so the backend `state` param added in `SearchResourceImpl` is unused by this path. Either the `config.queryParameters.state = ...` line was dropped from the commit, or the description/comment are inaccurate — this needs resolving.
- `app/.../SearchResourceImpl.java:367-368` — `SearchFilter.ofState(state)` and the single-valued `VersionState` query param imply an equality filter, which cannot express the original "exclude DISABLED" semantics (those kept both ENABLED and DEPRECATED). If the client is later wired to send `state=ENABLED`, DEPRECATED versions — currently returned — would be silently dropped, contradicting the client comment about "preserving the old behavior." Please confirm the intended handling of DEPRECATED. There is also no test covering the new server-side filter.
- `common/src/main/resources/META-INF/openapi.json:3984` — `state` was added only to the POST `/search/versions` (by-content) operation; the GET `searchVersions` overload at `SearchResourceImpl.java:341` was not updated, so the two endpoints now diverge in capability. If intentional (the resolver client only uses by-content), a one-line note would help; otherwise consider adding `state` to the GET path.

Kayd-06:
Thanks for the review @paoloantinori. I've pushed an update to address all of your points:

1. **Client-side state assignment**: You were right, the `config.queryParameters.state` assignment was accidentally dropped in my previous commit. I've restored it, so the client now properly pushes the `ENABLED` state filter to the server. I've also restored the client-side `filter` as a defensive guard against older registries that don't support the `state` query parameter.
2. **Explicit DEPRECATED handling**: Dropping `DEPRECATED` versions during schema resolution is an intentional behavior change we're making here to improve performance (fetching only `ENABLED` schemas). I've added a clear comment in `RegistryClientFacadeImpl.java` explicitly confirming this decision. 
3. **Inconsistent Endpoints**: The `GET /search/versions` endpoint actually *already* natively includes the `state` query parameter in `openapi.json` (around line 3832) and in the Java implementation (`SearchResourceImpl.java`). My PR was simply bringing the POST (`searchVersionsByContent`) endpoint to parity with it, so there is no divergence in capability between the two.

Let me know if this is good to go or any changes are needed.

paoloantinori:
@Kayd-06 Thanks — both points are addressed on `580ee24`: the `config.queryParameters.state = VersionState.ENABLED` assignment is restored, and keeping the client-side `.filter(v -> VersionState.ENABLED == v.getState())` as a defensive guard against older registries that don't support the new query param is the right call.

One thing I want to flag explicitly, because it's a real behavior change rather than a pure refactor:

The **old** client filter was `DISABLED != version.getState()`, which kept `ENABLED` **and** `DEPRECATED` versions. The **new** code queries `state=ENABLED` server-side and mirrors it client-side, so `DEPRECATED` versions are now **dropped** from the resolver's search results. The comment states this is intentional, and it may well be the desired end state — but it's a semantic change to schema resolution, not just a perf fix:

- `DEPRECATED` artifacts in Apicurio are still resolvable/usable (unlike `DISABLED`, which is meant to be hidden). Consumers who pin or resolve to a `DEPRECATED` version would silently stop finding it via this path.
- This is broader than the PR title ("move filtering to server-side") implies.

I'd suggest either:
1. Keeping the original semantics by filtering server-side on `state != DISABLED` (if the backend supports a negated/set-based state filter), or
2. If dropping `DEPRECATED` is genuinely intended, calling it out loudly in the PR description and getting explicit maintainer sign-off, since it's a behavior change for the schema resolver, not just a query-plumbing improvement.

The wiring itself (OpenAPI param, `SearchFilter.ofState`, defensive client guard) looks correct — this is purely about which states the resolver should consider.

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeImpl.java
```diff
@@ -141,11 +140,14 @@ public List<RegistryVersionCoordinates> searchVersionsByContent(String schemaStr
             config.queryParameters.orderby = VersionSortBy.GlobalId;
             config.queryParameters.order = SortOrder.Desc;
             config.queryParameters.limit = 100;
+            // We explicitly search for ENABLED and DEPRECATED versions to preserve original client semantics
+            // while moving the filtering logic to the server side (avoiding client-side filtering penalties).
+            config.queryParameters.state = new VersionState[] { VersionState.ENABLED, VersionState.DEPRECATED };
```
paoloantinori: This sends a closed allowlist `[ENABLED, DEPRECATED]`, but the original client logic was an open exclusion (`state != DISABLED`, keeping every non-DISABLED version). If `VersionState` contains any value besides ENABLED/DEPRECATED/DISABLED (e.g. DRAFT), those versions were previously included and are now silently excluded, which contradicts the comment's "preserve original client semantics." Worth confirming the full `VersionState` enum and that narrowing to ENABLED+DEPRECATED is intentional — and note that an explicit allowlist is more brittle than the original exclusion if states are added to the enum later.

paoloantinori (commented):
Solid change: the version-state filter is pushed server-side with proper array (`IN`) support across both the SQL and Elasticsearch backends, the OpenAPI/impl signatures are consistent, and the client correctly keeps a defensive `DISABLED` filter for older registries that ignore the new query param. One semantic concern is worth confirming before merge.

- `schema-resolver/.../RegistryClientFacadeImpl.java:145` — the original client filter was an *open exclusion* (`DISABLED != state`, i.e. keep every non-DISABLED version), but the new server query sends a *closed allowlist* of `[ENABLED, DEPRECATED]`. If `VersionState` has any value outside {ENABLED, DEPRECATED, DISABLED} (e.g. `DRAFT`), those versions were returned before and are now silently dropped, so the inline comment's claim that this "preserve[s] original client semantics" may be inaccurate. Please confirm the full `VersionState` enum and that intentionally narrowing to ENABLED+DEPRECATED is desired; a closed allowlist is also more brittle than the original exclusion against future enum additions.
- I could not verify against source in this read-only sandbox, so the above is conditional on the `VersionState` enum contents — flagging it rather than asserting it as a definite bug.

Kayd-06:
@paoloantinori I have done the changes, is it good to go?

paoloantinori:
@Kayd-06 Yes — the latest head (`8397930`) addresses everything from the previous rounds. The client now correctly sets `config.queryParameters.state = VersionState.ENABLED` and keeps the defensive client-side `.filter()`; the server side takes a `List<VersionState>` and wires `ofStates()` through both the SQL (`IN`) and Elasticsearch (`terms`) backends consistently. From a review standpoint this is good to go — the only remaining blocker is the maintainer `/accept` + CI.

Kayd-06:
@EricWittmann Can you have a look at this PR?

app/src/main/java/io/apicurio/registry/storage/dto/SearchFilter.java
```diff
@@ -90,6 +94,10 @@ public static SearchFilter ofState(VersionState state) {
         return new SearchFilter(SearchFilterType.state, state.name());
     }
 
+    public static SearchFilter ofStates(List<VersionState> states) {
+        return new SearchFilter(SearchFilterType.state, states.stream().map(VersionState::name).collect(Collectors.toList()));
```
paoloantinori, line 98: `ofStates()` stores a `List` under `SearchFilterType.state`, so `state` filters now have two possible value shapes (String from `ofState`, List from `ofStates`). This PR updates the SQL and Elasticsearch version-search consumers to branch on `isList()`, but `SearchFilter` is a shared abstraction across storage backends. Please confirm every other consumer of a `state` filter also guards `getStringValue()` — otherwise it throws `IllegalStateException("value is not of type string")` once the multi-state request from `RegistryClientFacadeImpl` reaches that backend.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -159,7 +159,7 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
                         op = filter.isNot() ? "!=" : "=";
                         where.append(
                                 "EXISTS(SELECT v.* FROM versions v WHERE v.groupId = a.groupId AND v.artifactId = a.artifactId AND ");
-                        where.append("v.state " + op + " ?");
+                        where.append(V_STATE_FIELD + op + " ?");
```
paoloantinori: This `searchArtifacts` state branch still binds `filter.getStringValue()` unconditionally, while the version-search `state` case below was just updated to handle `isList()`. Today artifact search only receives singular state filters so this isn't reachable, but for consistency with the new list support consider guarding this path with `isList()` (or an `ofStates`/`IN` branch) so a list-valued state filter can't throw `IllegalStateException` here later.

paoloantinori (commented):
The PR cleanly pushes version-state filtering server-side and correctly widens `state` to a multi-valued array across the OpenAPI spec, both backend resources, the SQL/ES stores, and the clients; the core mechanics (SQL `IN`-clause binders, ES `terms` query, and the enum-driven "all but DISABLED" request in the resolver) are correct and behavior-preserving. The main concern is that `SearchFilter` now carries two value shapes (String and List) for the `state` type, and the diff only updates the two version-search consumers to branch on `isList()`.

- `app/.../SearchFilter.java:98` — `ofStates()` stores a `List` under `SearchFilterType.state`, so every consumer of a `state` filter must now branch on `isList()` before calling `getStringValue()`. The diff updates only `SqlSearchRepository` and `ElasticsearchSearchService` for the version-search paths; please confirm the other storage backends' search implementations (e.g. in-memory / Kafka-streams) also guard `getStringValue()`, since `RegistryClientFacadeImpl.java:143` now always emits a multi-state array on the content-search path and an unguarded consumer would throw `IllegalStateException`.
- `app/.../SqlSearchRepository.java:162` — the `searchArtifacts` state branch still calls `filter.getStringValue()` unconditionally, while the sibling version-search branch a few lines down now handles the list case; this is currently unreachable (artifact search only receives singular state filters) but is a latent inconsistency worth guarding with `isList()` so a list-valued state filter can't throw here later.
- Scope: the stated goal is server-side state filtering, but the diff bundles unrelated changes — a ~45-line Windows/cygwin rewrite of `go-sdk/generate.sh`, behavior-preserving refactors of the `canonical`/`order`/`skipCount` handling in `SearchResourceImpl.java:357-396`, a reworded `CONTENT_SEARCH_UNSUPPORTED_MESSAGE`, and a hardcoded `v.version` at `SqlSearchRepository.java:317`. Consider splitting these so the state-filtering change reviews in isolation.
- `schema-resolver/.../RegistryClientFacadeImpl.java:147` — the retained client-side filter is a sound safety net for older registries, but it only helps if an older registry silently ignores the (now repeated) `state` query param rather than rejecting it; a one-line note documenting that assumption would help future readers.

EricWittmann:
@Kayd-06 this is a good change, but it does represent a breaking API change.  I don't think we can merge this onto `main` for that reason.  It's a minor enough breaking change that I think it's OK to add it in `3.4`, but not in `3.3.x`.

We'll need to create a `3.4.x` branch for changes like this.  I'm going to leave this open for now as a reminder to myself, since I don't have time to work on 3.4 planning or setup just yet.

Kayd-06:
@EricWittmann That makes sense.

One thing I wanted to mention, since contributors without write access can only have one open PR at a time, I'll be blocked from opening any other PRs while this one stays open waiting for the 3.4.x branch.

Do you have a preferred workflow for this?

EricWittmann:
@Kayd-06 sorry for the delay.  Can you close this PR and re-open it against the `3.4.x` branch that I just created?  :pray:

EricWittmann:
Requesting that this PR be re-opened against the `3.4.x` branch.  Closing.

PR #9112, opened 2026-07-31
fix(ui): use page size for pagination requests

### LINKED ISSUE #9086: UI pagination requests incorrect `limit` value, causing duplicate rows across pages
## Describe the bug

Paginated list views in the UI request an incorrect `limit` value after the first page, causing duplicate rows to appear on subsequent pages.

The issue was reproduced against a live registry instance and verified using browser network inspection. The backend returns exactly the number of items requested, indicating the incorrect request parameters originate from the frontend!

---

## Steps to Reproduce

1. Start a registry containing more than 20 items (for example, groups, artifacts, versions, drafts, or role mappings)
2. Open a paginated list view (e.g. **Explore → Groups**) and keep the page size set to **10**
3. Navigate to page 2 and then page 3
4. Observe both the rendered rows and the corresponding network requests in the browser's developer tools

---

## Expected Behavior

- Each page should request exactly the configured page size
- Page 2 should display only items **11–20**
- Page 3 should display only items **21–30**
- No rows should be duplicated across pages

---

## Actual Behavior

- Page 2 requests more than the configured page size and renders duplicate rows
- Page 3 again includes rows that were already displayed on page 2
- The number of rendered rows grows on later pages, making the pagination inconsistent with the displayed results

---

## Investigation

Using the browser's Network tab, the following requests were observed while navigating pages with a page size of **10**:

| Page | Offset | Limit Sent | Expected Limit |
|------|--------|------------|----------------|
| 1 | 0 | 10 | 10 |
| 2 | 10 | 20 | 10 |
| 3 | 20 | 30 | 10 |

The backend returns exactly the number of requested items, so the behavior appears to originate from the frontend request construction rather than the backend pagination implementation!

---

## Affected Areas

The same pagination behavior appears to be shared across multiple paginated views, including:

- Explore
- Search
- Drafts
- Versions / Branches
- Role Mappings

---

## Environment

- Reproduced on the current `main` branch
- Verified against a live local registry instance with real data (not mocked)
- Observed using browser network inspection and UI behavior

---

## Additional Notes

I searched through the existing open and closed issues but couldn't find another report describing this specific frontend pagination behavior or duplicate-row issue caused by the incorrect `limit` value!

Adii-45:
## Summary

Fixes the pagination request construction used across multiple UI service methods!

Previously, the UI sent the `limit` query parameter as the end index (`start + pageSize`) instead of the requested page size! Since the backend interprets `limit` as the number of items to return, requests after the first page fetched progressively larger result sets, resulting in duplicate rows being rendered across paginated views..!

This change updates all affected service methods to send `limit: paging.pageSize` while keeping `offset` unchanged!

Fixes #9086

## Changes

- Replace `limit: end` with `limit: paging.pageSize` in all affected pagination request builders
- Remove the now-unused `end` variable where applicable
- Add regression tests covering the affected services to verify the generated pagination query parameters across multiple pages

## Testing

- Added regression tests for:
  - `useSearchService`
  - `useAdminService`
  - `useDraftsService`
  - `useGroupsService`
- Verified all existing and new tests pass
- Ran lint successfully
- Reproduced the original issue against a live registry instance and verified that subsequent pages now request only the configured page size with no duplicate rows rendered.

### DIFF
--- a/ui/ui-app/src/services/useSearchService.test.ts
+++ b/ui/ui-app/src/services/useSearchService.test.ts
@@ -45,6 +45,7 @@ describe("useSearchService pagination", () => {
             await service.searchGroups([], "groupId" as any, "asc" as any, paging);
         }
 
+        expect(get).toHaveBeenCalledTimes(pagesToCheck.length);
         const queryParams = get.mock.calls.map(call => call[0].queryParameters);
         queryParams.forEach((params: any, i: number) => {
             expect(params.limit).toBe(10);
@@ -64,6 +65,7 @@ describe("useSearchService pagination", () => {
             );
         }
 
+        expect(get).toHaveBeenCalledTimes(pagesToCheck.length);
         const queryParams = get.mock.calls.map(call => call[0].queryParameters);
         queryParams.forEach((params: any, i: number) => {
             expect(params.limit).toBe(10);
@@ -81,6 +83,7 @@ describe("useSearchService pagination", () => {
             await service.searchVersions([], "version" as any, "asc" as any, paging);
         }
 
+        expect(get).toHaveBeenCalledTimes(pagesToCheck.length);
         const queryParams = get.mock.calls.map(call => call[0].queryParameters);
         queryParams.forEach((params: any, i: number) => {
             expect(params.limit).toBe(10);
--- a/ui/ui-app/src/services/useSearchService.test.ts
+++ b/ui/ui-app/src/services/useSearchService.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it, vi } from "vitest";
+import { Paging } from "@models/Paging.ts";
+
+const { getRegistryClientMock } = vi.hoisted(() => {
+    // useConfigService.ts reads this global at module load time (normally injected
+    // by the app's config.js in a real browser); provide a minimal stand-in so the
+    // service modules under test can be imported in a Vitest (node) environment.
+    const registryConfig = { artifacts: { url: "http://localhost:8080/apis/registry/v3/" } };
+    (globalThis as any).ApicurioRegistryConfig = registryConfig;
+    (globalThis as any).window = { ApicurioRegistryConfig: registryConfig };
+    return { getRegistryClientMock: vi.fn() };
+});
+
+vi.mock("@apicurio/common-ui-components", () => ({
+    useAuth: () => ({})
+}));
+
+vi.mock("@utils/rest.utils.ts", async (importOriginal) => {
+    const actual = await importOriginal<typeof import("@utils/rest.utils.ts")>();
+    return {
+        ...actual,
+        getRegistryClient: getRegistryClientMock
+    };
+});
+
+import { FilterBy, useSearchService } from "./useSearchService";
+
+// Regression test for issue #9086: the `limit` query param must always equal
+// `pageSize`, regardless of which page is being requested. Previously it was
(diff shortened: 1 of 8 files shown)


ui/ui-app/src/services/useSearchService.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        getRegistryClientMock.mockReturnValue({ search: { groups: { get } } });
+
+        const service = useSearchService();
+        for (const { page } of pagesToCheck) {
+            const paging: Paging = { page, pageSize: 10 };
+            await service.searchGroups([], "groupId" as any, "asc" as any, paging);
+        }
+
+        const queryParams = get.mock.calls.map(call => call[0].queryParameters);
+        queryParams.forEach((params: any, i: number) => {
```
paoloantinori, line 50: The assertions live entirely inside `queryParams.forEach(...)`, so if the service never reaches the mocked `get` (early return, broken mock chain, or a wrong client property path), `get.mock.calls` is `[]`, the loop runs zero times, and the test passes without asserting anything. Adding `expect(get).toHaveBeenCalledTimes(3)` (or `expect(queryParams).toHaveLength(3)`) before the loop makes a missed call fail loudly. The same gap exists in the admin, drafts, and groups test files (`useGroupsService.test.ts` centralizes this in `assertConstantLimit`, which has the same hole).

paoloantinori (commented):
Overall this is a correct, narrowly-scoped fix: replacing `limit: start + pageSize` (which grew every page and caused duplicate rows) with `limit: pageSize` is right, applied consistently across all four modified service files with `offset` left unchanged, and every changed function now has a regression test that genuinely catches the original bug.

- Fix is consistent and complete on the shown hunks: `useAdminService.ts`, `useDraftsService.ts`, `useGroupsService.ts` (4 methods), and `useSearchService.ts` (3 methods) all send `limit: paging.pageSize` with `offset: start` unchanged, and the now-unused `end` local is removed everywhere it existed.
- Easy win (see inline): the four new test files assert only inside a `forEach` over `get.mock.calls`, so they can pass vacuously if `get` is never invoked; adding a call-count assertion would make a broken or missed call fail loudly.
- Could not verify in this read-only review (no checkout / `gh`): worth confirming via a repo-wide grep for `start + pageSize` / `limit: end` that no other pagination call site in the `ui/` tree still uses the old offset+pageSize pattern.

Adii-45:
Thanks for catching this @paoloantinori! I updated the regression tests to assert the expected `get()` call count before iterating over the captured query parameters. I also added the guard to the shared `assertConstantLimit()` helper so all group-service tests benefit from the same protection without duplicating the assertion. Let me know if you see anything else that should be tightened up.

paoloantinori:
Thanks @Adii-45 — the `assertConstantLimit()` guard is exactly right. Asserting `toHaveBeenCalledTimes(PAGES_TO_CHECK.length)` *before* reading `mock.calls` means the test fails fast if a service method silently drops a request, rather than passing vacuously. Applying it once in the shared helper so all four `useGroupsService` cases benefit is cleaner than duplicating it.

Nothing else from me — the fix and the regression coverage look complete.

Adii-45:
@EricWittmann Could you please take a look whenever you get a chance? Thanks!

EricWittmann:
Thanks @Adii-45 — nice fix! The analysis is spot-on: `limit` is a count (SQL `LIMIT`), not an end index. Clean change, good regression tests, and responsive to review feedback. Merged!

PR #9117, opened 2026-07-31
fix(ui): redirect to artifact root after deleting a version

### LINKED ISSUE #9115: Deleting a version from the Version Overview page redirects to a broken 404 route
## Summary

Deleting a version via the kebab menu's "Delete version" action, while 
on that version's Overview page, redirects the browser to a malformed 
URL that renders a 404 error instead of returning to the artifact's 
main page.

## Video

https://github.com/user-attachments/assets/451d2298-66e0-42d5-8c42-ac3ac8cd2fd6

## Steps to Reproduce

1. Open any artifact version's Overview page 
   (`/explore/{groupId}/{artifactId}/versions/{version}`).
2. Click the kebab (three-dot) menu → "Delete version".
3. Confirm the deletion.
4. The browser navigates to `/explore/{groupId}/{artifactId}/versions` 
   — a route with no version ID — and renders **404 Error: page not 
   found**.

Note: deleting a version from the artifact's version list page 
(`/explore/{groupId}/{artifactId}`) works correctly — it stays on that 
page and simply refreshes the list. This bug is isolated specifically 
to the Version Overview page's delete flow.

## Root Cause

The delete-success handler in `VersionPage.tsx` navigates to 
`/explore/${groupId}/${artifactId}/versions` after a successful 
delete. That route (with no trailing version segment) is not a 
registered page in the router — only 
`/explore/{groupId}/{artifactId}` (artifact root) and 
`/explore/{groupId}/{artifactId}/versions/{version}` (a specific 
version) are valid routes. The bare `/versions` path has no matching 
route, so it falls through to the 404 handler.

## Suggested Fix

Change the post-delete redirect target to `/explore/${groupId}/${artifactId}` 
(the artifact root page), which is the correct, working destination 
after a version no longer exists. Verified manually navigating there 
directly renders the artifact page correctly, confirming this is a 
valid and appropriate redirect target.

Faisal77666:
## Summary

Fixes #9115

Deleting a version via the kebab menu's "Delete version" action, while 
on that version's Overview page, redirected to a malformed URL 
(`/explore/{groupId}/{artifactId}/versions`) that isn't a registered 
route, resulting in a 404 error instead of returning the user to the 
artifact's overview page.

## Root Cause

The delete-success handler in `VersionPage.tsx` navigated to 
`/explore/${groupId}/${artifactId}/versions` after a successful 
delete. That path has no trailing version segment and was never a 
registered route — only `/explore/{groupId}/{artifactId}` (artifact 
root) and `/explore/{groupId}/{artifactId}/versions/{version}` (a 
specific version) exist as valid routes in this app.

## Fix

Changed the post-delete redirect target to 
`/explore/${groupId}/${artifactId}` — the artifact root page, which 
correctly reflects the state after a version no longer exists.

## Testing

Verified both delete entry points:
- **Version Overview page kebab** (previously broken) → now correctly 
  redirects to the artifact page.
- **Artifact's version list page** (was already working) → confirmed 
  unaffected, still stays on the same page and refreshes the list.

## Video

**Before (bug):**

[raw-recording.webm](https://github.com/user-attachments/assets/a7bbe71a-193d-435d-8a9e-a438e2a4bc7a)

**After (fix applied):**

[raw-recording (1).webm](https://github.com/user-attachments/assets/67ebb0da-7a66-4b83-81e9-dad1f9f8f553)

### DIFF
--- a/ui/ui-app/src/app/pages/version/VersionPage.tsx
+++ b/ui/ui-app/src/app/pages/version/VersionPage.tsx
@@ -275,7 +275,7 @@ export const VersionPage: FunctionComponent<PageProperties> = () => {
             pleaseWait(false);
             const gid: string = encodeURIComponent(groupId || "default");
             const aid: string = encodeURIComponent(artifactId as string);
-            appNavigation.navigateTo(`/explore/${gid}/${aid}/versions`);
+            appNavigation.navigateTo(`/explore/${gid}/${aid}`);
         }).catch(error => {
             setPageError(toPageError(error, "Error deleting a version."));
         });


Faisal77666:
@paoloantinori @EricWittmann please take a look

carlesarnal:
Thanks for the fix, @Faisal77666 — the change itself looks correct. I verified the route definitions in `MainPageWithAuth.tsx` and confirmed that `/explore/:groupId/:artifactId/versions` (without a trailing `:version` param) is indeed not a registered route, so the old redirect was broken. Redirecting to `/explore/${gid}/${aid}` (the artifact root) is the right call.

One thing needs to be addressed before this can be merged:

**Missing DCO sign-off**  
The commit `73b9a94` doesn't include a `Signed-off-by` trailer, which is required by the project's contributing guidelines. You can fix this by amending the commit:

```bash
git commit --amend --signoff
git push --force-with-lease
```

Once that's done, this should be good to go from a code perspective. Nice, clean fix with a thorough PR description and videos — appreciated!

paoloantinori:
Looks good to me @Faisal77666 — the malformed redirect target was the only registered-route gap, and `/explore/${groupId}/${artifactId}` is the right landing page after a version is gone. The fix is correct and minimal. +1 from me.

Faisal77666:
@paoloantinori thanks!

Faisal77666:
@carlesarnal ready to merge

paoloantinori:
You're welcome @Faisal77666! 👍 Nothing else needed from my side — handing off to @carlesarnal for the merge.

Faisal77666:
@paoloantinori question before I file anything  is this expected or a bug?
If you create an `AGENT_CARD` artifact with empty/malformed JSON content, the create flow succeeds with no validation error  artifact + version both get created. But then opening that version's kebab (⋮) → "Edit Agent Card" does nothing  no page opens, no error shown. Same steps with valid JSON content work fine end-to-end for both create and edit.
My guess is the edit page tries to `JSON.parse()` the version content client-side to populate the form, and the exception on invalid content just fails silently.
Happy to open a proper issue with repro steps/video and take a stab at a fix if this isn't intentional  just didn't want to file it without checking first.

paoloantinori:
Hi @Faisal77666 — that's a real bug, not expected behaviour. Two separate gaps are showing up together:

1. **Create accepts empty/malformed JSON for an `AGENT_CARD`** — the registry doesn't enforce schema validation on artifact content at creation for every type, so a structurally invalid agent card gets persisted silently. That's the upstream hole.
2. **"Edit Agent Card" does nothing on such an artifact** — your guess is right: the edit page almost certainly runs `JSON.parse()` on the version content on load, the parse throws on empty/malformed input, and the error is swallowed (or the component fails to mount) with no user-visible feedback. Valid content works because the parse succeeds.

So the create side is permissive-by-design and the edit side isn't defensive against that permissiveness. Please file it as a UI issue (the edit-page part is the actionable half — it should either show an inline error or, ideally, surface a content-validation failure back at create time so the malformed artifact never gets persisted in the first place). A repo search for where `AGENT_CARD` content is parsed on edit will pinpoint the exact swallow site. Happy to take a look once you've filed it.

Faisal77666:
> Hi @Faisal77666 — that's a real bug, not expected behaviour. Two separate gaps are showing up together:
> 
> 1. **Create accepts empty/malformed JSON for an `AGENT_CARD`** — the registry doesn't enforce schema validation on artifact content at creation for every type, so a structurally invalid agent card gets persisted silently. That's the upstream hole.
> 2. **"Edit Agent Card" does nothing on such an artifact** — your guess is right: the edit page almost certainly runs `JSON.parse()` on the version content on load, the parse throws on empty/malformed input, and the error is swallowed (or the component fails to mount) with no user-visible feedback. Valid content works because the parse succeeds.
> 
> So the create side is permissive-by-design and the edit side isn't defensive against that permissiveness. Please file it as a UI issue (the edit-page part is the actionable half — it should either show an inline error or, ideally, surface a content-validation failure back at create time so the malformed artifact never gets persisted in the first place). A repo search for where `AGENT_CARD` content is parsed on edit will pinpoint the exact swallow site. Happy to take a look once you've filed it.

@paoloantinori i am working on it Thanks

Faisal77666:
@paoloantinori  if you don't mind, could I DM you on Slack for further guidance?

paoloantinori:
Of course, @Faisal77666 — happy to help. You can reach me on the Apicurio Slack (same handle). That said, for anything that ends up needing a code change or a wider opinion, GitHub issues/PRs keep the discussion visible to the rest of the maintainers, so feel free to open an issue too if it turns into something actionable.

Faisal77666:
@EricWittmann please take a look

EricWittmann:
Thanks @Faisal77666 — clean fix, great PR description with videos. Merged!

PR #9124, opened 2026-08-01
fix(operator): encode SQL password secret and add owner reference

### LINKED ISSUE #9123: bug(operator): SQL password auto-migration corrupts DB credentials and leaks orphaned Secrets
## Description

**Registry Version**: snapshot (3.x)
**Persistence type**: sql (bug is in the Kubernetes Operator's CR migration for the deprecated `app.sql.dataSource.*` fields, but it produces the `Secret` consumed by the `sql` storage datasource)

The operator's automatic migration of the deprecated `app.sql.dataSource.*` CR fields to `app.storage.sql.dataSource.*` (`SqlCRUpdater.update()`, invoked from `ApicurioRegistry3Reconciler.reconcile()` on every reconcile) has three related defects in its password-handling branch (`SqlCRUpdater.java`, lines 104-146). The URL and username migration branches in the same class do not have these problems only the password branch, because it is the only one that has to move a value into a Kubernetes `Secret` rather than a plain CR field.

### Bug 1 - Password written into `Secret.data` without base64 encoding

```java
// SqlCRUpdater.java, line 126-134
var secret = new SecretBuilder()
        .withNewMetadata()
            .withNamespace(primary.getMetadata().getNamespace())
            .withName(secretName)
        .endMetadata()
        .addToData(DEFAULT_SECRET_PASSWORD_FIELD, oldPassword.get())
        .build();
context.getClient().resource(secret).create();
```

`oldPassword.get()` is the raw plaintext value of the deprecated `app.sql.dataSource.password` CR field (`DeprecatedDataSource.password` is a plain `String`, see `operator/model/.../spec/DeprecatedDataSource.java`). `Secret.data` is the Kubernetes API's raw `data` map, whose values the API server/kubelet expect to already be base64-encoded `SecretBuilder#addToData` does not encode for you (only `stringData` does, and only after server-side merge). The codebase already knows this distinction: `GitOpsITTest.java` (lines 199-208) correctly uses `.addToStringData(...)` when writing plaintext SSH keys into a test `Secret`. `SqlCRUpdater` does not follow that pattern.

The resulting Secret is later consumed as a standard env var via `SecretKeyRefTool.applySecretEnvVar()` (`operator/controller/.../utils/SecretKeyRefTool.java`, lines 97-111), which wires a `valueFrom.secretKeyRef`. Kubelet base64-*decodes* whatever is in `data` before injecting it into the container. Since the value was never encoded, the running app pod receives a corrupted password string (or, depending on the API server's validation, the `create()` call itself can fail on invalid base64).

### Bug 2 - Orphaned Secret created on every retried/duplicate migration, with no owner reference

The new Secret's name is randomly generated on every invocation (`primary.getMetadata().getName() + "-datasource-password-" + randomUUID()...`, line 124) and created via a raw `context.getClient().resource(secret).create()` call, bypassing the operator's normal javaoperatorsdk dependent-resource framework entirely. No owner reference is ever set on it (there is no `OwnerReference`/`addNewOwnerReference` usage anywhere in `operator/`). `ApicurioRegistry3Reconciler.reconcile()` calls `SqlCRUpdater.update()` on every reconcile (line 204) and, if it reports `true`, immediately returns `UpdateControl.patchResource(primary)` (lines 206-207). If that CR patch is retried for any reason before the deprecated field is durably observed as cleared on the next reconcile (patch conflict, controller restart, resync), the migration branch runs again from scratch and creates *another* Secret with a new random name leaking the previous one, which is never cleaned up (no owner reference means it also survives deletion of the CR itself).

### Bug 3 - "already migrated" detection compares plaintext to base64 and can never succeed

```java
// SqlCRUpdater.java, lines 108-117
var newPasswordValue = newPassword
        .map(x -> context.getClient().secrets()
                .inNamespace(primary.getMetadata().getNamespace())
                .withName(x.getName()).get())
        .map(x -> x.getData().get(DEFAULT_SECRET_PASSWORD_FIELD));

if (newPassword.isEmpty() || oldPassword.equals(newPasswordValue)) {
    ... // proceed with migration / mark as migrated
} else {
    log.warn("Automatic update cannot be performed, because the target field `app.storage.sql.dataSource.password` is already set.");
}
```

`Secret#getData()` returns the raw (base64) map as stored by the API server, but `oldPassword` is the plaintext value from the deprecated field. These are never equal in practice even the corresponding URL/username branches, which compare plaintext to plaintext, work correctly, but the password branch compares two different representations. In practice this means: once a user has *any* password secret reference set on the new field (including one this same migration created), the deprecated `app.sql.dataSource.password` field can never be detected as "already migrated" and cleared the operator logs the "cannot be performed" warning on every single reconcile indefinitely, and the deprecated field lingers in the CR forever instead of being cleaned up per the intended migration contract (see the cleanup block at lines 148-165, which only clears `app.sql` once `oldPassword` is empty).

### Environment

Any Kubernetes cluster running the Apicurio Registry Operator (3.0.7+) reconciling an `ApicurioRegistry3` CR that still sets the deprecated `spec.app.sql.dataSource.password` field.

### Steps to Reproduce

1. Apply an `ApicurioRegistry3` CR using the deprecated field, e.g.:
   ```yaml
   spec:
     app:
       sql:
         dataSource:
           url: jdbc:postgresql://postgres:5432/registry
           username: registry
           password: supersecret
   ```
2. Let the operator reconcile the CR. It creates a Secret named `<cr-name>-datasource-password-<random>`.
3. Inspect the created Secret: `kubectl get secret <name> -o jsonpath='{.data.password}' | base64 -d` the decoded value is not `supersecret`; it is garbage, because the stored value was never base64-encoded in the first place (double-decoding of a plaintext string).
4. The app Deployment's pod fails to authenticate to the SQL database using this secret.
5. Separately, trigger a reconcile retry (e.g. `kubectl edit` the CR to bump `resourceVersion` conflict, or restart the operator pod) before the deprecated field has been observed as cleared a second, differently-named orphan Secret is created, and the first is never deleted.

**Expected**: The automatic migration transparently moves the password into a properly base64-encoded Secret, the app connects to the database exactly as before, migration is idempotent (does not recreate Secrets on retries), the created Secret is owned by the CR (garbage collected on CR deletion), and the deprecated field is reliably detected and cleared once migration is complete.

**Actual**:
- The migrated password is corrupted (not base64-encoded), breaking DB connectivity right after an operator upgrade silently "migrates" a user's CR.
- Retried/duplicate reconciles leak additional un-owned Secrets that are never garbage collected.
- The deprecated field can never be confirmed as migrated (plaintext vs. base64 comparison), so it's never cleared and the warning log never stops.

### Logs

Operator log output on a CR still using the deprecated field (`LOG_LEVEL=DEBUG`), repeating on every reconcile because the migration never completes (Bug 3):

```
WARN  [i.a.r.o.u.SqlCRUpdater] CR field `app.sql.dataSource.password` is DEPRECATED and should not be used.
INFO  [i.a.r.o.u.SqlCRUpdater] Performing automatic CR update from `app.sql.dataSource.password` to `app.storage.sql.dataSource.password`.
```

then, once a Secret reference exists on the new field, every subsequent reconcile logs:

```
WARN  [i.a.r.o.u.SqlCRUpdater] CR field `app.sql.dataSource.password` is DEPRECATED and should not be used.
WARN  [i.a.r.o.u.SqlCRUpdater] Automatic update cannot be performed, because the target field `app.storage.sql.dataSource.password` is already set.
```

App pod log after the corrupted Secret is mounted as an env var:

```
FATAL [io.quarkus] password authentication failed for user "registry"
ERROR [io.agroal.pool] Datasource '<default>': Failed to obtain JDBC connection
```

## Root Cause

All three defects live in the same ~40-line password-migration branch of `SqlCRUpdater.update()`. The branch hand-rolls Secret creation instead of following the codebase's own established `stringData` convention (already used correctly elsewhere, e.g. `GitOpsITTest`) and bypasses the operator's dependent-resource lifecycle (no owner reference, ad-hoc client call), which is why it also has no idempotency/ownership guarantees. The "already migrated" check inherits the same conceptual gap: nobody accounted for the encoding mismatch between a plaintext CR field and a base64 Secret field when writing the equality check that the URL/username branches got right by using plaintext-to-plaintext comparisons.

Note: there are open, unmerged PRs (#8888/#8897/#8920/#8603 lineage) that aim to *delete* the deprecated fields and `SqlCRUpdater` entirely in a future cleanup. None of them are merged, none of them fix this bug for users who hit it in the meantime, and removal doesn't address CRs that already have deprecated fields set. This should be fixed independently of that cleanup effort.

## Impact

Any operator-managed install still using the deprecated `app.sql.dataSource.*` fields (a documented, supported migration path, not a removed one) can have its database connectivity silently broken by the operator itself during a routine reconcile this is a self-inflicted outage caused by code whose entire purpose is to be a transparent, no-op migration. It also leaks Kubernetes Secrets in the cluster with no owner reference, which is an operational/security hygiene issue in multi-tenant clusters (orphaned credential-bearing objects that outlive their owning CR).

## Acceptance Criteria

- [ ] `SqlCRUpdater` writes the migrated password into the Secret using `stringData` (or explicitly base64-encodes the value before calling `addToData`), and the resulting Secret decodes to the original plaintext password.
- [ ] The created Secret carries an owner reference to the `ApicurioRegistry3` CR so it is garbage collected when the CR is deleted.
- [ ] Re-running the migration for a CR that has already been migrated (or partially migrated) is idempotent it does not create a second Secret, and does not leak the first.
- [ ] The "already migrated" comparison correctly detects a previously-migrated password (comparing against the decoded/plaintext value, not the raw base64 representation) so the deprecated `app.sql.dataSource.password` field is reliably nulled out and the parent `app.sql` block is cleaned up once migration is complete, matching the existing behavior of the URL/username branches.
- [ ] Add `@QuarkusTest`/unit coverage in the operator's existing CR-updater test suite for: (a) a fresh migration produces a Secret whose decoded value matches the original password, (b) a second reconcile after migration is a no-op (no new Secret, `update()` returns `false`, deprecated field stays cleared), and (c) the created Secret has the expected owner reference.

bhuvan-somisetty:
## Summary

Fixes #9123.

Fixes three related bugs in the operator's automatic migration of the deprecated `app.sql.dataSource.password` CR field to `app.storage.sql.dataSource.password` (`SqlCRUpdater`).

## Root Cause

The password migration branch built a Kubernetes `Secret` by writing the plaintext password directly into `Secret.data` via `addToData`, which expects an already base64-encoded value. Since the value was never encoded, the app pod received a corrupted password once kubelet base64-decoded it on injection.

Two further problems lived in the same code path:

- The Secret name was randomly generated on every invocation and created with a raw client call outside the operator's dependent-resource framework, with no owner reference set. A retried reconcile before the CR patch was observed created another Secret and left the previous one orphaned (and it would also survive deletion of the CR).
- The "already migrated" check compared the plaintext deprecated password against the raw (still base64-encoded) value read back from the Secret, so it could never match, causing the migration warning to be logged on every reconcile even after migration had completed.

## Changes

- `operator/controller/.../updater/SqlCRUpdater.java`
  - Use `addToStringData` instead of `addToData` when writing the password into the Secret, matching the pattern already used for SSH key secrets in `GitOpsITTest`.
  - Derive a deterministic Secret name from the CR name instead of a random UUID, and check for an existing Secret with that name before creating one.
  - Add an owner reference from the CR to the Secret so it is garbage collected with the CR.
  - Add `decodeBase64` and use it before comparing the existing Secret's password value against the deprecated plaintext field, so completed migrations are correctly detected.
- `operator/controller/.../it/CRUpdateITTest.java`
  - Add `testCRUpdatePasswordSecret`, which migrates a CR with the deprecated password field, asserts the resulting Secret decodes back to the original plaintext password and carries an owner reference to the CR, then forces another reconcile and asserts no second Secret is created.

## Test plan

- [ ] `./mvnw -f operator/pom.xml test-compile -pl controller -am -DskipTests` compiles cleanly (CI)
- [ ] `./mvnw -f operator/pom.xml checkstyle:check -pl controller` passes (CI)
- [ ] `CRUpdateITTest#testCRUpdatePasswordSecret` (requires the operator test cluster, same as the rest of this package's IT suite)
- [ ] Storage variant testing: not applicable — this change is scoped to the operator's CR migration logic for the `sql`/`postgresql` storage type and does not touch `storage/impl/`


bhuvan-somisetty:
Hey @paoloantinori, whenever you get a chance, would you mind taking a look at this one? It fixes a few related issues in the operator's SQL password migration (base64 encoding, a leaking Secret with no owner ref, and a broken already-migrated check) - figured you'd have good context here. No rush, just flagging it so it doesn't get lost. Thanks!

bhuvan-somisetty:
Hey @EricWittmann, the verification gate is showing failed but it's just waiting on a maintainer /accept before the real checks kick in, not an actual issue. Could you take a look when you get a chance? Thanks!

paoloantinori:
Thanks for the flag, @bhuvan-somusetty — took a close look. The three fixes all check out:

1. **base64 encoding** — switching `addToData` → `addToStringData` is the right call. `Secret.data` is base64 on the wire; `stringData` lets the API server do the encoding, which matches what you'd want when seeding it from a plaintext CR field.
2. **Leaking / orphaned Secret** — deterministic name + `OwnerReference(controller=true, blockOwnerDeletion=true)` is correct for GC and avoids the per-reconcile UUID leak. The `secretAlreadyPresent` get before the create also prevents the recreate-on-retry path you described.
3. **Already-migrated check** — decoding before the `oldPassword.equals(newPasswordValue)` compare is the actual fix; the old plaintext-vs-base64 compare could never match, so the deprecation warning fired every reconcile.

One non-blocking note: the get-then-create for the Secret is technically a TOCTOU window, but JOSDK serializes reconciles per primary resource so concurrent creates on the same CR aren't expected — and the deterministic name means the worst case is a loud `AlreadyExists` on a collision rather than a silent leak, which is strictly better than today. If you ever want to harden it further, a `createOrReplace` (or catching `AlreadyExists`) would close it, but I wouldn't block on it.

The `testCRUpdatePasswordSecret` coverage is solid — it asserts the decoded round-trip, the owner reference, and that a forced re-reconcile keeps a single Secret by UID. The `startsWith(secretName)` filter is a nice touch since it also catches the old `<name>-datasource-password-<uuid>` leak shape. LGTM from my side.

bhuvan-somisetty:
Thanks for the thorough review, @paoloantinori! Fair point on the TOCTOU window pushed a follow-up commit that drops the separate get-then-create check and just attempts the create directly, treating a 409 as a concurrent reconcile winning the race. Closes it rather than just narrowing it, per your suggestion. Ready for another look whenever you have time.

carlesarnal:
Reviewed both commits. All three bug fixes check out and the TOCTOU follow-up is clean. Details below.

**Bug fixes**

1. Base64 encoding (SqlCRUpdater.java:141) - addToStringData is correct. The old addToData put plaintext into .data, which the API server treats as already-encoded, so the pod got garbage after kubelet decoded it.

2. Deterministic Secret name (SqlCRUpdater.java:133) - Switching from random UUID suffix to a fixed name eliminates the per-reconcile leak. Combined with the owner reference (lines 135-142), orphaned Secrets are impossible.

3. Already-migrated check (SqlCRUpdater.java:121-123) - decodeBase64 via flatMap correctly decodes the Secret .data value before comparing against the plaintext deprecated field. The old plaintext-vs-base64 compare could never match, so the deprecation warning fired every reconcile.

4. TOCTOU closure (SqlCRUpdater.java:145-152) - Direct create with 409 catch per paoloantinori suggestion. The deterministic name and identical content from the same CR mean treating 409 as a concurrent reconcile win is correct.

**Test coverage**

testCRUpdatePasswordSecret (CRUpdateITTest.java:80-125) is thorough: decoded password round-trip, owner reference verification, and re-reconcile idempotency via label patch asserting a single Secret by UID. The startsWith filter also catches old UUID-suffixed leak shapes.

**Commit hygiene** - DCO sign-off on both, conventional commit format, no star/unused imports, consistent brace style.

No issues found.

bhuvan-somisetty:
Appreciate the detailed pass, @carlesarnal! Glad the TOCTOU follow-up landed well too. Let me know if anything else comes up during merge.

paoloantinori:
Perfect — that's exactly the right call. Dropping the get-then-create and treating the 409 as the concurrent-reconcile signal closes the TOCTOU window cleanly rather than just narrowing it, and the `log.debug` keeps the happy path quiet. Deterministic name + owner reference + 409 handling all line up. LGTM, nothing else from my side.

bhuvan-somisetty:
@EricWittmann whenever you have a sec, could you run /accept on this one? Both @paoloantinori and @carlesarnal have reviewed and it's looking good, just needs the accept to kick off full CI. Thanks!

paoloantinori:
Thanks for the nudge, @bhuvan-somisetty — agreed, from both review passes (mine and @carlesarnal's) this is in good shape and just needs the `/accept` to run the full CI. That one's @EricWittmann's call.

bhuvan-somisetty:
Hey @EricWittmann, circling back on this one it's been sitting since Aug 1 with two maintainer approvals (paoloantinori and carlesarnal, including a follow-up TOCTOU fix they both signed off on). Just needs an `/accept` to get out of `lifecycle/new` so it can move forward. No rush if you're slammed, just didn't want it to fall through the cracks. Thanks!

bhuvan-somisetty:
/unstale

Hey @EricWittmann, following up on this PR as it was marked stale by the bot. Both @paoloantinori and @carlesarnal have reviewed and approved the changes (including the TOCTOU follow-up). 

Whenever you get a chance, could you please run /accept so the PR can transition out of lifecycle/new and run the full test suite? Thanks!

carlesarnal:
@bhuvan-somisetty I don't see any changes in this pr now..

paoloantinori:
Nothing further from my side — the earlier feedback, including the TOCTOU follow-up, was addressed at the current head. The /accept step is Eric's call.

bhuvan-somisetty:
Thanks @carlesarnal and @paoloantinori for the reviews! Glad the TOCTOU fix and the rest look good on your end. All checks are green now, so this should be ready for @EricWittmann to take a final look whenever he gets a chance.

Issue #8510, opened 2026-07-09
Avro converter returns null for fields with Debezium/Connect logical type annotations


carlesarnal:
## Description

The Apicurio Registry Avro converter (`io.apicurio.registry.utils.converter.AvroConverter`) returns `null` when deserializing Avro fields that have Connect logical type annotations (e.g. `connect.name`, `connect.version`). This causes a `NullPointerException` in downstream consumers that expect non-nullable values.

## Steps to Reproduce

1. Debezium PostgreSQL source connector with Apicurio Avro converter produces CDC events
2. The `ExtractNewRecordState` SMT (on source side) flattens the Debezium envelope to just the `after` fields
3. The resulting Avro schema has fields with Connect logical type annotations:

```json
{
  "type": "record",
  "name": "Value",
  "namespace": "dbserver1.inventory.orders",
  "fields": [
    {"name": "order_number", "type": "int"},
    {"name": "order_date", "type": {"type": "int", "connect.version": 1, "connect.name": "io.debezium.time.Date"}},
    {"name": "customer_id", "type": "int"},
    {"name": "quantity", "type": "int"},
    {"name": "price", "type": "double"}
  ]
}
```

4. When the Iceberg sink connector deserializes these records using the same Apicurio Avro converter, the `order_date` field (which has `connect.name: io.debezium.time.Date`) is deserialized as `null` instead of the actual integer value.

## Error

```
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because "value" is null
    at org.apache.iceberg.parquet.ColumnWriter$2.write(ColumnWriter.java:42)
```

The Parquet writer receives a null for the `order_date` int field, which should contain a non-null integer (days since epoch).

## Analysis

The Avro schema defines `order_date` as a non-nullable `int` type with additional Connect metadata properties (`connect.version`, `connect.name`). The Apicurio Avro converter appears to mishandle the deserialization of fields that use this annotated type format — returning null instead of the actual int value.

Fields without Connect annotations (plain `"type": "int"` like `order_number`, `customer_id`, `quantity`) are deserialized correctly. Only fields with the expanded type object format containing `connect.*` properties are affected.

## Expected Behavior

The converter should ignore the `connect.*` metadata annotations and deserialize the underlying Avro type (`int`) normally, preserving the field value.

## Environment

- **Apicurio Registry**: `latest-snapshot` (2026-07-09)
[...]

PR #9125, opened 2026-08-01
fix(serdes): fail on writer schema mismatch in Avro serializer

WasThatRudy:
## What

The Avro serializer builds its DatumWriter from the schema returned by the schema resolver. GenericDatumWriter extracts record values by field position of that schema, without checking that it matches the schema the record was built with. When the two drift apart, for example a stale latest artifact version when `apicurio.registry.find-latest=true` is used without `apicurio.registry.auto-register`, values are silently written under the wrong fields. A null from a drifted optional slot lands in an unrelated field with no error anywhere in the pipeline.

This PR compares the record schema against the resolved schema by Avro parsing canonical form before writing, and fails with a descriptive error on structural mismatch. Property-only differences (connect.* metadata, docs) do not affect the binary layout and still pass. Validated schema pairs are cached by identity, so the steady-state cost is one lock-free map lookup per record. A new config property `apicurio.registry.avro.validate-writer-schema` (default `true`) restores the old behavior when disabled. Similar idea to Confluent's `latest.compatibility.strict`.

## Why

Relates to #8510. The reported symptom (annotated Debezium date field deserialized as null, all plain fields correct, NPE downstream in the Iceberg sink) is reproducible on main with this recipe:

1. Register `(order_number, order_date ["null","int" + connect.name io.debezium.time.Date], customer_id, quantity, price)` as the latest version of the topic artifact.
2. Configure the converter with `find-latest=true` and `auto-register=false`.
3. Serialize a record whose schema is `(order_number, legacy_col [optional string, value null], customer_id, quantity, price)`.

Serialization succeeds silently, the null is written into order_date's union branch, and the sink decodes `order_date=null` with everything else intact. Type-incompatible drift throws, null into a union does not, which is why only the nullable annotated field is affected. Whether the original environment hit exactly this drift is an open question on the issue, but the corruption itself is real and silent either way.

Same-schema round trips are unaffected: the #8520 unit and integration tests still pass, and I verified evolution sequences, nullable unions, and separate source/sink converter instances against a live registry before and after this change.

## Test plan

- [x] New unit tests in `AvroSerializerWriterSchemaValidationTest` (7 cases): structural drift throws with an actionable message, drift throws under JSON encoding, identical schema instance passes, property-only differences pass and round-trip (twice, covering the cache hit), reflect datum provider drift throws, a drifted record nested in an array throws, and validation disabled demonstrates the legacy silent corruption (value surfaces under the wrong field, then disappears when read with the producer schema).
- [x] `./mvnw test -pl serdes/generic/serde-common-avro,serdes/kafka/avro-serde,utils/converter -am` green.
- [x] `./mvnw checkstyle:check -pl serdes/generic/serde-common-avro` green.
- [x] Manual end-to-end against a live registry (3.3.2-SNAPSHOT, in-memory): healthy evolution and optional-field flows still round-trip values through `AvroConverter`; the drift scenario above now fails at the producer with the descriptive error instead of producing a silent null.
- [x] Serdes configuration docs updated (`assembly-configuring-kafka-client-serdes.adoc`).


serdes/generic/serde-common-avro/src/main/java/io/apicurio/registry/serde/avro/AvroSerializer.java
```diff
@@ -109,18 +117,87 @@ protected Object getSchemaCacheKey(U data) {
...
     @Override
     public void serializeData(ParsedSchema<Schema> schema, U data, OutputStream out) throws IOException {
-        Encoder encoder = createEncoder(schema.getParsedSchema(), out);
-
         if (data instanceof NonRecordContainer) {
             // noinspection unchecked
             data = (U) NonRecordContainer.class.cast(data).getValue();
         }
 
+        if (validateWriterSchemaEnabled && schema.getParsedSchema().getType() == Schema.Type.RECORD) {
```
paoloantinori: This guard restricts validation to top-level `RECORD` schemas. A drifted element/value record nested inside an `ARRAY` or `MAP` would still be written through `GenericDatumWriter` with the same silent misalignment, but validation is skipped here. Consider widening the type check (the canonical-form compare already recurses into element/value schemas, so it would catch nested drift for free) or explicitly documenting the exclusion — note the docs currently name only primitive/union, but enum/array/map/fixed top-level schemas are also unvalidated today.

serdes/generic/serde-common-avro/src/test/java/io/apicurio/registry/serde/avro/AvroSerializerWriterSchemaValidationTest.java
```diff
@@ -0,0 +1,231 @@
...
+        AvroSerializer<GenericRecord> serializer = serializer(DRIFTED_SCHEMA,
+                Map.of(AvroSerdeConfig.AVRO_VALIDATE_WRITER_SCHEMA, "false"));
+
+        // Legacy behavior: serialization succeeds and the null from legacy_code silently
+        // lands in order_date. This documents the corruption the validation prevents.
+        byte[] bytes = serializer.serializeData(TOPIC, record());
+
+        GenericRecord decoded = decode(serializer, bytes, DRIFTED_SCHEMA);
+        assertEquals(10001, decoded.get("order_number"));
+        assertNull(decoded.get("order_date"));
```
paoloantinori: This assertion can't actually prove the silent corruption the comment describes. `legacy_code` is set to null, and `decode` builds `new GenericDatumReader<>(writerSchema)` using the drifted schema as both reader and writer, so `order_date` comes back null regardless of whether values were truly misplaced. Setting `legacy_code` to a non-null value and decoding with `DRIFTED_SCHEMA` as writer and `RECORD_SCHEMA` as reader would make this test genuinely demonstrate the data loss the validation prevents.

paoloantinori (commented):
Solid change. A multi-perspective review (correctness, concurrency, performance, contract) found no production defects: the validated-pair cache is safe because the structural compare always runs before any insert (so a mismatched pair can never be cached as valid) and the identity-based key can't yield a wrong cached verdict; the `String.format` message has matching placeholder/argument counts; the `NonRecordContainer` unwrap correctly precedes `toSchema`; and `validateWriterSchemaEnabled` is properly wired from config in `configure()` (AvroSerializer.java). Two actionable suggestions below, plus a behavior-change heads-up and one minor note.

- **RECORD-only scope (AvroSerializer.java:125)** — validation runs only when the resolved schema is a top-level `RECORD`, so a drifted element/value record inside an `ARRAY` or `MAP` is not caught, even though the same silent-corruption class applies to nested records. Widening the type guard would let the canonical-form compare catch that drift naturally; the docs also understate the exclusion (they name only primitive/union, but enum/array/map/fixed are skipped too).
- **`testValidationCanBeDisabled` doesn't robustly prove the corruption it documents** (AvroSerializerWriterSchemaValidationTest.java:165) — `legacy_code` is null and `decode` uses the drifted writer schema as both reader and writer, so `order_date=null` can't be distinguished from plain field-absence. A non-null `legacy_code` decoded against the original `RECORD_SCHEMA` reader would genuinely demonstrate the silent loss this feature prevents.
- **Default-on is a deliberate behavior change** (AvroSerdeConfig default `true`, documented) — upgraded deployments using `find-latest` without `auto-register` whose latest artifact has drifted will start failing loud at the producer instead of corrupting silently. That is the intended trade-off and there is a clean opt-out, but it is worth calling out in release notes.
- Minor: `IllegalStateException` surfaces correctly, but worth confirming against the codebase's serializer-exception convention (Kafka serializers frequently throw `SerializationException`).

WasThatRudy:
Thanks for the review, all four points addressed in the updated commit.

Widened the guard: validation now runs whenever the serialized object carries its own schema (any GenericContainer, checked before the NonRecordContainer unwrap), with the toSchema fallback kept for record schemas so reflect POJOs stay covered. That catches drifted records nested in arrays/maps for free via the canonical form, and there's a new test for the array case. Plain Java collections have no schema to compare, so they stay unvalidated, and the docs now say exactly that.

On the disabled-validation test, you're right that it didn't prove anything. One nuance with the suggested change: a non-null string in the drifted slot hits UnresolvedUnionException at write time, so the corruption wouldn't be silent. I made legacy_code an int instead: the test now shows 12345 surfacing under order_date when read with the message's schema, and disappearing entirely when read with the producer's schema as reader.

On the exception type: this module is Kafka-agnostic (no kafka dependency, SerializationException only exists in serde-kafka-common), so IllegalStateException matches the module's existing convention (AvroSchemaUtils, RecordIdStrategy).

Agreed the default-on behavior change belongs in release notes. Happy to add a note wherever those get drafted.

carlesarnal:
Re-reviewed the updated commit (9080158). All four items from the previous review have been addressed:

1. **Widened scope** — Validation now runs whenever the serialized object carries its own schema (any `GenericContainer`, checked at `AvroSerializer.java:121` before the `NonRecordContainer` unwrap). The `RECORD`-type fallback at line 128 only kicks in for reflect POJOs that don't implement `GenericContainer`. The new `testNestedDriftInArraySchemaFailsSerialization` test confirms arrays with drifted element schemas are caught.

2. **Disabled-validation test** — Much better. Using `legacy_code = 12345` (an int) instead of null makes the corruption concrete: the test asserts `order_date == 12345` (value under the wrong field name) and `legacy_code == null` when re-read with the producer's schema (value gone). This is a clear demonstration of the silent data loss.

3. **Release notes** — Acknowledged, belongs in the release process.

4. **IllegalStateException** — Makes sense for this module. `serde-common-avro` has no Kafka dependency, and `IllegalStateException` matches what `AvroSchemaUtils` (line 75, 104) and `NonRecordContainer` already throw.

Code quality items I verified on this pass:
- DCO sign-off ✅
- Conventional commit format ✅
- No star imports, no unused imports ✅
- Thread safety: `ConcurrentHashMap` with identity-based keys is appropriate; the `containsKey`→`put` race is benign (worst case is a redundant validation) ✅
- `String.format` placeholder/argument count matches (4/4) ✅
- `SchemaResolverConfig.FIND_LATEST_ARTIFACT` and `AUTO_REGISTER_ARTIFACT` constants verified ✅
- 7 test cases with specific value assertions covering: structural drift, JSON encoding, identity pass-through, property-only tolerance (including cache hit), reflect provider, nested array drift, and disabled validation with corruption proof ✅
- Docs accurately describe the validation scope (GenericContainer types validated, plain Java collections without attached schemas are not) ✅

No new issues found. The implementation is clean and well-tested. Deferring to a maintainer for approval.

serdes/generic/serde-common-avro/src/main/java/io/apicurio/registry/serde/avro/AvroSerializer.java
```diff
@@ -109,18 +118,96 @@ protected Object getSchemaCacheKey(U data) {
...
+    /**
+     * The Avro datum writer extracts record values by field position of the writer schema, so a
+     * resolved schema that differs structurally from the record's schema silently writes values
+     * under the wrong fields, or writes null into optional fields. Schemas are compared by parsing
+     * canonical form: properties such as connect.* metadata do not affect the binary layout and are
+     * therefore tolerated. Successfully validated schema pairs are remembered by identity, so
+     * producers reusing schema instances (the common case, including one serializer shared across
+     * several event types) pay only a lock-free cache lookup per record.
+     */
+    private void validateWriterSchema(Schema recordSchema, Schema writerSchema) {
```
paoloantinori, line 154: Easy win: the message interpolates `recordSchema.getFullName()`, which returns `null` for non-named top-level schemas (ARRAY/MAP/UNION). This path is reachable — `testNestedDriftInArraySchemaFailsSerialization` validates a `GenericData.Array`, so a real array-drift failure would render `...the record being serialized [null].`. Consider falling back to `recordSchema.getType()` (or the writer schema's name) so the message stays descriptive for non-record drift. Minor, but it directly affects the 'descriptive error' goal of the change.

paoloantinori (commented):
Sound, well-targeted fix for a genuine silent-data-corruption path (resolved/writer schema field positions drifting from the record's schema). A static review of the diff plus two independent adversarial passes found no blocking defects: the canonical-form comparison, identity-based validation cache, and opt-out flag are correctly implemented. A few observations follow.

- **Core logic is correct** (`AvroSerializer.java`, `serializeData`, ~lines 119-138): the datum schema is extracted before the `NonRecordContainer` unwrap and validated against `schema.getParsedSchema()`; `SchemaNormalization.toParsingForm` correctly tolerates `connect.*` props/docs, as `testPropertyOnlyDifferencesAreTolerated` demonstrates.
- **The validation cache is sound** (`AvroSerializer.java`, `ValidatedSchemaPair`, ~lines 148-205). Avro `Schema` instances are immutable, so an identity-based key can neither falsely throw (a miss just recomputes an equal canonical form) nor falsely skip a mismatch (a hit is the same two instances previously validated). The `clear()`-on-overflow under `ConcurrentHashMap` is racy but benign — worst case is redundant re-validation.
- **`default=true` is well-targeted, not a broad break** (`AvroSerdeConfig.java`). Under content-based lookup or `auto-register`, the resolved schema equals the record's schema by construction, so the check is a no-op; the new failure only surfaces in the `find-latest`-without-`auto-register` drift case the fix is meant to catch. Worth flagging in release notes for that configuration, and the opt-out (`validate-writer-schema=false`) plus the actionable error message are appropriate.
- **Over-conservative only inside the drift case**: canonical form requires exact structural match, so a safe end-appended defaulted field would also throw. That's confined to the already-risky `find-latest` config and errs toward safety, so acceptable — noting it in case you want position-aware leniency later.
- **Minor scope gap, acknowledged in docs** (`assembly-configuring-kafka-client-serdes.adoc`): plain `List`/`Map` of reflect POJOs aren't validated (resolved type ≠ RECORD). Since positional misalignment is record-specific, this is acceptable.
- **Minor doc nit**: the PR description lists "6 cases" but `AvroSerializerWriterSchemaValidationTest` has 7 `@Test` methods (`testNestedDriftInArraySchemaFailsSerialization` is the 7th).

Note: this is a static review of the provided diff (source and CI not available in this read-only environment), so the green-build/test claims are taken on the author's report.

serdes/generic/serde-common-avro/src/main/java/io/apicurio/registry/serde/avro/AvroSerializer.java
WasThatRudy: Checked this empirically on Avro 1.12.1: getFullName() never returns null. For non-named schemas it falls back to the type name, so record gives 't.V', array gives 'array', map gives 'map', and a union gives 'union[null, V]'. The array-drift failure renders '...the record being serialized [array].', which is terse but not null. I left the message as is to keep it bounded, but happy to print both schema names if you think that reads better.

jsenko (commented):
Fixes #8510 — adds writer schema validation to the Avro serializer to catch silent field-position drift.

## Summary

Well-crafted fix for a genuine silent data corruption path. The canonical-form comparison, identity-based cache, and opt-out flag are correctly implemented. Prior automated reviews confirmed correctness of the core logic, cache safety, and config wiring. No must-fix issues found.

- **Default-on behavior change** — deployments using `find-latest=true` without `auto-register=true` whose latest artifact has drifted will start failing loud. This is the correct trade-off (fail loud > silent corruption), and there's a clean opt-out. Worth mentioning in release notes for that specific configuration.
- **IllegalStateException vs SerializationException** — Kafka serializers typically throw `org.apache.kafka.common.errors.SerializationException`. The `IllegalStateException` still works (Kafka catches `Exception`), but `SerializationException` would be more idiomatic and let callers distinguish serializer failures from other exceptions without catching the broader type. Not blocking.
- **RECORD-only scope** — acknowledged in docs. Non-record containers (plain List/Map) bypass validation, which is acceptable since positional misalignment is record-specific.

### should-fix: Config property is outside the `app` module — no `@Info` annotation

The new `apicurio.registry.avro.validate-writer-schema` property lives in the SerDes module, not the `app` module. Per project convention, `@ConfigProperty` + `@Info` is only required in the `app` module. However, the SerDes config docs (the `.adoc` file updated here) are maintained manually — verify the property name in the docs exactly matches the constant in `AvroSerdeConfig.java`. It does in this PR, noting for future maintainers.

carlesarnal (approved):

PR #9126, opened 2026-08-01
fix(ui): prompt template test panel - boolean defaults, zero input, and validation error display

### LINKED ISSUE #8678: Bug: Number input fields reject zero value in Prompt Template test panel

### LINKED ISSUE #8729: fix(ui): prompt template validation errors don't show the variable name
## Description

The render endpoint returns validation errors as `RenderValidationError` objects with the fields `variableName` (required in the schema), `message`, `expectedType` and `actualType`, see the `RenderValidationError` definition in `common/src/main/resources/META-INF/openapi.json`, populated by `validateVariables()` in `app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java`.

The UI model declares a different field name: `ui/ui-app/src/models/RenderPromptResponse.ts` defines `RenderPromptValidationError` with `path?: string` instead of `variableName`. The Test Prompt panel prefixes each error message with `ve.path` when present (`ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx`), and since `path` never exists in the API response, the variable-name prefix is always dropped.

Found while reading the test-panel code; file references above.

### Environment

Local build from `main`, JVM mode, UI dev server.

### Steps to Reproduce

1. Register a `PROMPT_TEMPLATE` artifact with two or more typed variables (e.g. two `integer` variables).
2. In the Test Prompt panel on the version's Documentation tab, enter invalid values for both (any non-numeric state, or values violating `minimum`/`maximum`) and click Render.
3. The "Validation Errors" alert appears.

Expected: each list item identifies the failing variable, e.g. `max_words: Type mismatch: expected integer but got string`.

Actual: the list shows bare messages with no variable name, e.g. `Type mismatch: expected integer but got string` twice, with no way to tell which variable each error refers to.

A small rename in the UI model (aligning it with the API's `variableName`) plus updating the alert rendering should be enough; surfacing `expectedType`/`actualType` in the alert could be a nice addition while there.

### LINKED ISSUE #8998: fix(ui): boolean prompt template variables without an explicit default submit empty string instead of false
Registry Version: main (snapshot)
Persistence type: N/A - UI bug

Related but distinct from #8817 (which fixed a duplicate-label rendering issue on boolean
prompt variables) — this is about the underlying value sent to the backend, not the label.

In PromptTemplateTestPanel.tsx, initialValues population (lines ~45-52) sets any variable
without an explicit `default` to an empty string "", regardless of declared type:

  variablesList.forEach(({ name, variable }) => {
      if (variable.default !== undefined) {
          initialValues[name] = variable.default;
      } else {
          initialValues[name] = "";
      }
  });

A later useEffect (~lines 56-67) attempts to correct this for booleans specifically:

  if (!(name in next)) {
      const type = (variable.type || "string").toLowerCase();
      next[name] = type === "boolean" ? false : (variable.default ?? "");
  }

But because initialValues already set the key to "" during the initial render, `(name in
next)` is true, so `!(name in next)` is false, and the boolean correction never runs.

Steps to Reproduce
1. Create a PROMPT_TEMPLATE artifact with a boolean variable that has no default:
   { "enabled": { "type": "boolean", "required": true } }
2. Open the Test Prompt panel. The checkbox visually renders unchecked.
3. Click Render without touching the checkbox.

Expected: Render succeeds with `enabled` treated as false.
Actual: Alert shows "enabled: Type mismatch: expected boolean but got string" — the checkbox
looks unchecked but the underlying value is still the string "", not the boolean false.

Root Cause
initialValues (~PromptTemplateTestPanel.tsx line 45-52) doesn't check variable type before
defaulting to "", so the later type-aware correction in the useEffect never triggers for
keys that already exist (even if their value is the wrong type).

Suggested Fix

  const initialValues: Record<string, any> = {};
  variablesList.forEach(({ name, variable }) => {
      const type = (variable.type || "string").toLowerCase();
      if (variable.default !== undefined) {
          initialValues[name] = variable.default;
      } else if (type === "boolean") {
          initialValues[name] = false;
      } else {
          initialValues[name] = "";
      }
  });

I have this fix working locally and can open a PR shortly if this direction looks right.

YuvrajVerma09arch:
## Description

Fixes three bugs in the PromptTemplateTestPanel component:

### 1. Boolean variables default to false instead of empty string (fixes #8998)
- Boolean variables without an explicit default now default to `false` instead of `""`
- Previously, a boolean with no default would send empty string to the render endpoint, causing type mismatch validation errors

### 2. Number input fields accept zero value (fixes #8678)
- Replaced falsy `||` with nullish `??` for both value display and onChange handler
- Zero is now accepted as a valid numeric value

### 3. Validation errors show variable name (fixes #8729)
- The RenderPromptValidationError interface was using `path` but the API returns `variableName`
- Updated the interface and alert rendering to use the correct field name
- Added `expectedType` and `actualType` fields to match the API response

## Testing

- 14/14 Vitest tests pass
- 0 TypeScript errors
- 0 ESLint errors
- Added new tests for the RenderPromptValidationError type contract

## Files Changed

| File | Change |
|---|---|
| `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx` | 3 bug fixes |
| `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts` | Added validation error type tests |
| `ui/ui-app/src/models/RenderPromptResponse.ts` | Fixed validation error interface |

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -45,8 +45,11 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
 
     const initialValues: Record<string, any> = {};
     variablesList.forEach(({ name, variable }) => {
+        const type = (variable.type || "string").toLowerCase();
         if (variable.default !== undefined) {
             initialValues[name] = variable.default;
+        } else if (type === "boolean") {
+            initialValues[name] = false;
         } else {
             initialValues[name] = "";
         }
@@ -149,8 +152,8 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                 return (
                     <TextInput
                         type="number"
-                        value={values[name] || ""}
-                        onChange={(_event, val) => setValue(name, type === "integer" ? parseInt(val) || "" : parseFloat(val) || "")}
+                        value={values[name] ?? ""}
+                        onChange={(_event, val) => setValue(name, type === "integer" ? parseInt(val) ?? "" : parseFloat(val) ?? "")}
                         aria-label={name}
                     />
                 );
@@ -228,7 +231,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                     <Alert variant="warning" title="Validation Errors" className="validation-errors">
                         <ul>
                             {validationErrors.map((ve, i) => (
-                                <li key={i}>{ve.path ? `${ve.path}: ` : ""}{ve.message}</li>
+                                <li key={i}>{ve.variableName ? `${ve.variableName}: ` : ""}{ve.message}</li>
                             ))}
                         </ul>
                     </Alert>
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -148,15 +148,20 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                 );
             }
             case "integer":
-            case "number":
+            case "number": {
+                const handleNumberChange = (_event: React.FormEvent<HTMLInputElement>, val: string) => {
+                    const n = type === "integer" ? parseInt(val) : parseFloat(val);
+                    setValue(name, Number.isNaN(n) ? "" : n);
+                };
                 return (
                     <TextInput
                         type="number"
                         value={values[name] ?? ""}
-                        onChange={(_event, val) => setValue(name, type === "integer" ? parseInt(val) ?? "" : parseFloat(val) ?? "")}
+                        onChange={handleNumberChange}
                         aria-label={name}
                     />
                 );
+            }
             case "array":
             case "object":
                 return (
--- a/ui/ui-app/src/models/RenderPromptResponse.ts
+++ b/ui/ui-app/src/models/RenderPromptResponse.ts
@@ -1,6 +1,8 @@
 export interface RenderPromptValidationError {
-    path?: string;
+    variableName: string;
     message: string;
+    expectedType?: string;
+    actualType?: string;
 }
 
 export interface RenderPromptResponse {
--- a/ui/ui-app/src/models/RenderPromptResponse.ts
+++ b/ui/ui-app/src/models/RenderPromptResponse.ts
@@ -1,5 +1,5 @@
 export interface RenderPromptValidationError {
-    variableName: string;
+    variableName?: string;
     message: string;
     expectedType?: string;
     actualType?: string;
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "vitest";
 import { coerceEnumValue } from "./PromptTemplateTestPanel.utils";
+import { RenderPromptValidationError } from "@models/RenderPromptResponse";
 
 describe("coerceEnumValue", () => {
     it("parses an integer enum selection to a number", () => {
@@ -32,5 +33,31 @@ describe("coerceEnumValue", () => {
 
     it("keeps the placeholder selection as an empty string for a boolean variable", () => {
         expect(coerceEnumValue("", "boolean")).toBe("");
+    });
+});
+
+describe("RenderPromptValidationError", () => {
+    it("has variableName field for displaying which variable failed", () => {
+        const error: RenderPromptValidationError = {
+            variableName: "temperature",
+            message: "Type mismatch: expected number but got string",
+            expectedType: "number",
+            actualType: "string",
+        };
+
+        expect(error.variableName).toBe("temperature");
+        expect(error.message).toBe("Type mismatch: expected number but got string");
+        expect(error.expectedType).toBe("number");
+        expect(error.actualType).toBe("string");
+    });
+
+    it("variableName and message are required fields", () => {
+        const error: RenderPromptValidationError = {
+            variableName: "enabled",
+            message: "Variable is required",
+        };
+
+        expect(error.variableName).toBeTruthy();
+        expect(error.message).toBeTruthy();
     });
 });
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
@@ -51,7 +51,16 @@ describe("RenderPromptValidationError", () => {
         expect(error.actualType).toBe("string");
     });
 
-    it("variableName and message are required fields", () => {
+    it("variableName is optional - template-level errors may not have a variable", () => {
+        const error: RenderPromptValidationError = {
+            message: "Template rendering failed",
+        };
+
+        expect(error.variableName).toBeUndefined();
+        expect(error.message).toBe("Template rendering failed");
+    });
+
+    it("variableName and message are required fields when variableName is present", () => {
         const error: RenderPromptValidationError = {
             variableName: "enabled",
             message: "Variable is required",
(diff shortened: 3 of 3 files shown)


paoloantinori (commented):
The interface/UI fix (#3) is correct and the boolean default (#1) is reasonable, but the number-input fix (#2) introduces a regression: `??` does not catch the `NaN` that `parseInt`/`parseFloat` return on empty or invalid input, so clearing a number field now stores `NaN` instead of `""`.

- `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx:156` — In the onChange, `parseInt(val) ?? ""` resolves to `NaN` (not `""`) when `val` is empty, because `NaN` is falsy but not nullish. That `NaN` then reaches `value={values[name] ?? ""}` at line 155 (React warns "Received NaN for the `value` attribute") and serializes to `null` on submit via `JSON.stringify`. The zero fix itself is sound (`0 ?? "" === 0`), but `NaN` needs explicit handling, e.g. `const n = parseInt(val); setValue(name, Number.isNaN(n) ? "" : n)`.
- `ui/ui-app/src/models/RenderPromptResponse.ts:2` — `variableName` is now required on a type that models an external API response. If the render endpoint can emit a validation error not tied to a specific variable (template/render-level errors), this required field would be inaccurate; consider `variableName?: string`. The UI already guards the falsy case (`PromptTemplateTestPanel.tsx:233`).
- `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts:38` — The new tests assert fields on an object literal they just constructed, so they pass regardless of the interface and don't exercise the number-input onChange; consequently the `NaN` regression above isn't covered. The file is also missing a trailing newline.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -149,8 +152,8 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                 return (
                     <TextInput
                         type="number"
-                        value={values[name] || ""}
-                        onChange={(_event, val) => setValue(name, type === "integer" ? parseInt(val) || "" : parseFloat(val) || "")}
+                        value={values[name] ?? ""}
+                        onChange={(_event, val) => setValue(name, type === "integer" ? parseInt(val) ?? "" : parseFloat(val) ?? "")}
```
paoloantinori: `parseInt(val) ?? ""` does not catch `NaN`: when the field is cleared, `parseInt("")` is `NaN`, which is falsy but not nullish, so this resolves to `NaN` rather than `""`. As a result `setValue(name, NaN)` runs (previously `setValue(name, "")`), the controlled input at line 155 receives `NaN` (React "Received NaN" warning), and submit serializes it to `null` via `JSON.stringify`. The zero fix is correct (`0 ?? "" === 0`), but treat `NaN` explicitly, e.g. `const n = parseInt(val); Number.isNaN(n) ? "" : n` — and the same for the `parseFloat` branch.

ui/ui-app/src/models/RenderPromptResponse.ts
```diff
@@ -1,6 +1,8 @@
 export interface RenderPromptValidationError {
-    path?: string;
+    variableName: string;
```
paoloantinori: `variableName` is now required on a type that models an external API response. If the render endpoint can return a validation error that isn't tied to a single variable (e.g. a template-level or render-level error), this field should stay optional (`variableName?: string`) to avoid the type claiming a value the API may omit. The UI already handles the missing case via `ve.variableName ? ...` in `PromptTemplateTestPanel.tsx`.

YuvrajVerma09arch: Fixed all three points:
1. Number inputs now explicitly check Number.isNaN() — empty/invalid input stores "" instead of NaN
2. variableName made optional in RenderPromptValidationError (template-level errors may not have a variable)
3. Added 10 tests covering number parsing edge cases + optional variableName; trailing newline added

All tests pass (25/25) and build succeeds.

carlesarnal:
## Review

Nice work addressing the review feedback. All three original bugs are correctly fixed and the NaN regression is properly handled now.

**Maintainer Feedback Status** - All three points from paoloantinori's review are resolved:
NaN handling fixed with Number.isNaN(n). variableName made optional. Tests added for number parsing and validation error shape.

**Minor Nits:**
1. PromptTemplateTestPanel.utils.test.ts still ends without a trailing newline.
2. parseInt(val) at ~line 158 could use an explicit radix: parseInt(val, 10).
3. Consider extracting parseNumber into PromptTemplateTestPanel.utils.ts so tests exercise the actual production code path.
4. Neither commit includes a Signed-off-by line (DCO required per CONTRIBUTING.md).

**Overall:** Core fixes are solid. Boolean defaults, zero-value number inputs, and validation error display all work correctly. Changes are focused and well-scoped.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -228,7 +236,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                     <Alert variant="warning" title="Validation Errors" className="validation-errors">
                         <ul>
                             {validationErrors.map((ve, i) => (
-                                <li key={i}>{ve.path ? `${ve.path}: ` : ""}{ve.message}</li>
+                                <li key={i}>{ve.variableName ? `${ve.variableName}: ` : ""}{ve.message}</li>
```
paoloantinori, line 239: Switching `ve.path` → `ve.variableName` is only correct if the render endpoint returns `variableName` in its validation errors. If it doesn't, the variable name will silently stop displaying — no crash, just an empty prefix before `ve.message`, which is easy to miss. The new `RenderPromptValidationError` tests in `PromptTemplateTestPanel.utils.test.ts` construct a typed object and assert its own fields back out, so they confirm the TypeScript interface compiles but don't validate the real API payload shape. Recommend confirming the field name against an actual render-error response (the backend serializer or a captured fixture) before merging this fix.

paoloantinori (commented):
The three fixes are well-targeted and the logic changes look correct: the `||`→`??` and `parseInt(val)||""`→`Number.isNaN(n) ? "" : n` rewrite properly preserves `0` as a valid numeric value (display and onChange), and defaulting unset booleans to `false` avoids the empty-string type mismatch. The one thing worth confirming before merge is the `path`→`variableName` rename, because it is only correct if the render endpoint actually returns `variableName`, and the new tests don't validate that against the real payload.

- `PromptTemplateTestPanel.tsx:157` (display `value={values[name] ?? ""}`) and `:151-152` (`handleNumberChange`): the nullish-coalescing and `Number.isNaN` guard correctly stop treating `0` as empty for both integer and number. This looks right.
- `PromptTemplateTestPanel.tsx:239` and `RenderPromptResponse.ts:2`: the display fix depends entirely on the backend returning `variableName`. If the field name differs, the variable name silently stops rendering (no crash, just empty prefix), which is the failure mode to guard against. See inline comment.
- `RenderPromptResponse.ts:4-5`: `expectedType` and `actualType` are added to the interface but the Alert only renders `variableName`/`message`. Fine if intentional (the fields now at least match the API), though surfacing `expectedType`/`actualType` would make type-mismatch messages more actionable.
- `PromptTemplateTestPanel.utils.test.ts`: the `number input parsing` suite exercises a locally re-declared `parseNumber` that mirrors the component's inline logic, so it gives false confidence — a future change to the component's parser wouldn't be caught. Minor; worth noting the production code path isn't covered.

YuvrajVerma09arch:
@EricWittmann @paoloantinori — all feedback addressed:\n- NaN guard added for number inputs\n- `variableName` made optional in `RenderPromptValidationError`  \n- 10 new tests + trailing newline\n\nAll 25/25 tests pass, CI green. Ready for final review.

paoloantinori:
Thanks @YuvrajVerma09arch — re-read the current head (`2ff898a`). All three points from the prior round are addressed cleanly:

- **NaN guard** — the `Number.isNaN(n) ? "" : n` rewrite preserves `0` as a valid value while correctly rejecting empty/non-numeric input. The dedicated `parseNumber` test block covers integer/float, zero, negative, and garbage inputs.
- **`variableName` optional** in `RenderPromptValidationError` aligns with the API response shape for template-level errors, and the new type-contract tests pin that.
- Trailing newline restored on the test file.

Logic looks correct across all three bug fixes. No further comments from me — agreed with @carlesarnal's read.

EricWittmann:
Hi @YuvrajVerma09arch — thanks for the work on this. Unfortunately, 2 of the 3 fixes in this PR have already been merged via other PRs:

- **Zero value in number inputs (#8678)** — merged in PR #8681
- **Boolean defaults (#8998)** — merged in PR #9017

The conflicts you're seeing are because those same lines were already changed on main.

The third fix — **renaming `path` to `variableName` in `RenderPromptValidationError` (#8729)** — is still needed and hasn't been addressed elsewhere. I'm going to close this PR, but please feel free to open a new, focused PR for just that fix. It would be a small change to `RenderPromptResponse.ts` and `PromptTemplateTestPanel.tsx`. Apologies for the duplication — we're working on better issue assignment to avoid this in the future.

PR #9130, opened 2026-08-01
fix(ui): include falsy query parameters in createEndpoint helper

### LINKED ISSUE #9129: [UI] createEndpoint utility incorrectly omits falsy query parameters (false, 0)
## Description

**Registry Version**: `3.3.2-SNAPSHOT`
**Persistence type**: `sql`

The `createEndpoint` helper in `ui/ui-app/src/utils/rest.utils.ts` omits valid falsy query parameter values when constructing request URLs.

The current implementation filters query parameters using a truthiness check:

```typescript
if (queryParams[key]) {
    const value = encodeURIComponent(queryParams[key]);
    // ...
}
```

Because of this, valid values such as `false` and `0` are treated as if they were absent and are not included in the generated URL.

This makes it impossible for callers to explicitly send falsy values through the helper and may cause requests to fall back to server-side defaults instead of the values supplied by the client.

Additionally, there are existing call sites (for example in `useContractsService.ts`) that manually convert numeric values using `String(offset)` and `String(limit)` before passing them to `createEndpoint`, suggesting that callers have had to work around this behavior.

### Environment

* Apicurio Registry: `3.3.2-SNAPSHOT`
* Apicurio Registry UI
* Local development environment
* Google Chrome (latest)

### Steps to Reproduce

1. Call `createEndpoint` with falsy query parameter values:

```typescript
createEndpoint(
    "http://localhost:8080",
    "/test",
    {},
    {
        isOk: false,
        count: 0,
    }
);
```

2. Inspect the generated URL.

**Expected**

The generated URL should preserve valid falsy values:

```text
http://localhost:8080/test?isOk=false&count=0
```

**Actual**

The generated URL omits those parameters entirely:

```text
http://localhost:8080/test
```

### Logs

No server-side logs are required to reproduce this issue. The behavior can be observed by inspecting the URL generated by `createEndpoint`.

kunaldevsahu:
## Summary

This PR fixes an issue in the `createEndpoint` utility where valid falsy query parameter values were omitted when constructing request URLs.

The previous implementation filtered query parameters using a truthiness check, causing values such as `false` and `0` to be excluded from the generated URL. As a result, callers could not explicitly send these values to the backend and requests could unintentionally fall back to server defaults.

This change replaces the truthiness check with explicit `null`/`undefined` handling so that valid falsy values are preserved while continuing to omit `null`, `undefined`, and empty-string (`""`) values. The exclusion of empty strings is intentional to preserve the existing behavior and keep the change narrowly scoped to the reported issue.

## Changes

* Preserve valid falsy query parameter values (`false` and `0`) during URL construction.
* Continue ignoring `null`, `undefined`, and empty-string (`""`) query parameters.
* Add regression tests covering:

  * Boolean query parameters (`true` and `false`)
  * Numeric query parameters (including `0`)
  * Ignoring `null`, `undefined`, and empty-string query parameters
  * Existing path parameter substitution behavior

## Testing

* Added unit tests for `createEndpoint`.
* Verified that generated URLs correctly serialize `false` and `0`.
* Confirmed that `null`, `undefined`, and empty-string query parameters continue to be omitted as intended.

## Related Issue

Fixes #9129

### DIFF
--- a/ui/ui-app/src/utils/rest.utils.ts
+++ b/ui/ui-app/src/utils/rest.utils.ts
@@ -261,16 +261,15 @@ export function createEndpoint(baseHref: string, path: string, params?: any, que
     if (queryParams) {
         let first: boolean = true;
         for (const key in queryParams) {
-            if (queryParams[key]) {
-                const value: string = encodeURIComponent(queryParams[key]);
+            const rawValue: any = queryParams[key];
+            if (rawValue !== null && rawValue !== undefined && rawValue !== "") {
+                const value: string = encodeURIComponent(rawValue);
                 if (first) {
                     rval = rval + "?" + key;
                 } else {
                     rval = rval + "&" + key;
                 }
-                if (value !== null && value !== undefined) {
-                    rval = rval + "=" + value;
-                }
+                rval = rval + "=" + value;
                 first = false;
             }
         }
--- a/ui/ui-app/src/utils/rest.utils.test.ts
+++ b/ui/ui-app/src/utils/rest.utils.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, it } from "vitest";
+import { createEndpoint } from "./rest.utils";
+
+describe("createEndpoint", () => {
+    const baseHref = "http://localhost:8080/apis/registry/v3";
+
+    it("should correctly substitute path parameters", () => {
+        const url = createEndpoint(baseHref, "/groups/:groupId/artifacts/:artifactId", {
+            groupId: "my-group",
+            artifactId: "my-artifact"
+        });
+        expect(url).toBe(`${baseHref}/groups/my-group/artifacts/my-artifact`);
+    });
+
+    it("should handle paths without query parameters or path parameters", () => {
+        const url = createEndpoint(baseHref, "/system/info");
+        expect(url).toBe(`${baseHref}/system/info`);
+    });
+
+    it("should append query parameters correctly, including falsy boolean and numeric values", () => {
+        const url = createEndpoint(
+            baseHref,
+            "/search",
+            {},
+            {
+                isActive: true,
+                isDeprecated: false,
+                limit: 10,
+                offset: 0,
+                negative: -1
+            }
+        );
+        expect(url).toBe(`${baseHref}/search?isActive=true&isDeprecated=false&limit=10&offset=0&negative=-1`);
+    });
+
+    it("should omit null, undefined, and empty string query parameters", () => {
+        const url = createEndpoint(
+            baseHref,
+            "/search",
+            {},
+            {
+                query: "",
+                filter: null,
+                sort: undefined,
+                validParam: "value"
+            }
+        );
+        expect(url).toBe(`${baseHref}/search?validParam=value`);
+    });
+
+    it("should correctly handle path and query parameters combined", () => {
+        const url = createEndpoint(
+            baseHref,
+            "/groups/:groupId/artifacts",
+            { groupId: "group-1" },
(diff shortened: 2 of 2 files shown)


ui/ui-app/src/utils/rest.utils.ts
```diff
@@ -261,16 +261,15 @@ export function createEndpoint(baseHref: string, path: string, params?: any, que
     if (queryParams) {
         let first: boolean = true;
         for (const key in queryParams) {
-            if (queryParams[key]) {
-                const value: string = encodeURIComponent(queryParams[key]);
+            const rawValue: any = queryParams[key];
+            if (rawValue !== null && rawValue !== undefined && rawValue !== "") {
```
paoloantinori, line 265: Empty-string values are still excluded via `rawValue !== ""`. The new test documents this as intentional, but the PR summary only lists `null`/`undefined` as ignored while the stated goal is to preserve falsy values — so the empty-string boundary is the one place worth confirming. If dropping `""` is intended, a one-line note (or updating the summary) would make the asymmetry with `false`/`0` explicit; if callers may ever need to send an explicit empty value, consider allowing `""` through as well.

paoloantinori (commented):
This is a correct, minimal fix with solid test coverage: replacing the truthiness check with explicit `null`/`undefined`/`""` handling properly preserves falsy values like `false` and `0`, and the dropped inner `null`/`undefined` guard was effectively dead code since `encodeURIComponent` always returns a string. No blocking issues. Two observations worth a second look:

- **`rest.utils.ts:265`** — Empty-string values are still excluded (`rawValue !== ""`). This is documented as intentional in the new test (`rest.utils.test.ts:36`), but it sits slightly at odds with the PR's stated goal of "including falsy query parameters" and isn't mentioned in the summary's "Changes" list (which only lists `null`/`undefined`). Worth confirming that boundary is intended and documenting why `false`/`0` are kept while `""` is dropped.
- **`rest.utils.ts:265`** — `NaN` now serializes as `?key=NaN` where the old truthiness check silently dropped it. Rare in practice, but a caller that previously relied on a computed `NaN` being omitted will now forward it to the backend. Worth a moment of thought, though surfacing it is arguably better than the old silent drop.

ui/ui-app/src/utils/rest.utils.ts
kunaldevsahu: Thanks @paoloantinori  for pointing this out ! Keeping "" excluded was intentional. I wanted to keep the fix focused on preserving valid explicit falsy values like false and 0 while maintaining the existing behavior for empty strings to minimize behavioral changes. I'll update the PR summary to make that distinction clearer.

carlesarnal:
Reviewed commit 47df15a. Clean fix — replacing the truthiness check with explicit null/undefined/empty-string guards correctly preserves valid falsy values (false, 0). The removed inner null/undefined guard on encodeURIComponent result was dead code. Tests are thorough with exact URL assertions. DCO, conventional commit, scope all pass. Agree with paoloantinori observations — no blocking concerns. @kunaldevsahu reminder to update PR summary re: empty-string exclusion. Deferring to maintainer for approval.

kunaldevsahu:
Thanks @carlesarnal  for the review , I appreciate the feedback. I'll update the PR summary to explicitly mention the intentional exclusion of empty-string values.

kunaldevsahu:
Thanks @carlesarnal  I've updated the PR summary to clarify the empty-string exclusion.

EricWittmann:
Thanks @kunaldevsahu — clean fix with thorough tests. The dead code removal was a nice bonus. Merged!

PR #9140, opened 2026-08-01
fix(ui): preserve 0 value in integer/number prompt template fields

### LINKED ISSUE #9128: [Bug]: Integer/number prompt template variables cannot be set to '0'
## Description

**Registry Version**:  main (snapshot / latest-snapshot docker image) 

The Test Prompt panel for PROMPT_TEMPLATE artifacts does not correctly handle the value 0 for integer/number type variables. When a user types "0" into a number field, the field is immediately cleared instead of retaining the value.

This makes it impossible to test prompt templates that legitimately require a numeric variable to be set to zero (e.g., a "count" or "offset" parameter).

### Environment

Running via Docker (apicurio/apicurio-registry:latest-snapshot and apicurio/apicurio-registry-ui:latest-snapshot), backend on port 8080, UI on port 8888. Tested in Chrome browser on Windows.

### Steps to Reproduce

1. Create a PROMPT_TEMPLATE artifact with an integer variable, e.g.: { "count": { "type": "integer", "required": true } }
2. Open the artifact's version page, go to the Documentation tab.
3. In the Test Prompt panel, try typing "0" into the count field.

**Expected:** The field should retain the value 0.

**Actual:** The field is immediately cleared to empty as soon as "0" is typed. This happens because the onChange handler in PromptTemplateTestPanel.tsx uses: parseInt(val) || ""

Since 0 is falsy in JavaScript, `0 || ""` evaluates to "", silently discarding valid zero input.

### Root Cause
The onChange handler for number/integer fields in PromptTemplateTestPanel.tsx converts the typed value using parseInt()/parseFloat(), then falls back to an empty string if the parsed result is falsy. Since 0 is a falsy value in JavaScript, this fallback incorrectly triggers for a valid zero input, discarding it instead of preserving it.

### Logs

(No error logs — this is a silent UI logic bug, not a crash).

jayaprajapatii:
Fixes #9128

## PR Summary

The Test Prompt panel for `PROMPT_TEMPLATE` artifacts fails to handle the value `0` for integer/number type variables. When a user types "0" into a number field, the field is immediately cleared instead of retaining the value — making it impossible to test prompt templates that legitimately require a numeric variable set to zero (e.g. a `count` or `offset` parameter).

## Root Cause

Two places in `PromptTemplateTestPanel.tsx` used the falsy-fallback `||` operator, which incorrectly treats `0` as "no value":

1. The `onChange` handler, which parsed input and fell back to `""`:

```js
parseInt(val) || ""
```
2. The `value` prop used to display the field, which read the stored value the same way:
```js
values[name] || ""
```
Since `0` is falsy in JavaScript, both `0 || ""` evaluate to `""`, silently discarding valid zero input — first when typing it, and again when displaying it.

## Fix

* `onChange`: replaced the falsy-fallback with explicit checks that only clear the field when input is genuinely empty or fails to parse as a number.
* `value`: replaced `values[name] || ""` with `values[name] ?? ""` (nullish coalescing), which preserves `0`/`-0` while still falling back to `""` for `null`/`undefined`.

## Testing

*  Verified the fix logic in isolation (Node.js REPL) for both parsing and display cases.
* Verified end-to-end in a running instance (Docker `latest-snapshot` images for backend and UI):
  - Typed "0" into the count field → field correctly retains 0
  - Clicked Render → output correctly substitutes the value ("You need 0 items to complete.")
* Ran `eslint` on the modified file with no issues reported.


ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -150,7 +150,10 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                     <TextInput
                         type="number"
                         value={values[name] || ""}
```
paoloantinori: This line still uses the falsy-fallback `values[name] || ""` — the same pattern the PR identifies as the root cause. Now that the new `onChange` stores a real number `0` (`setValue(name, parsed)`, where `parseInt("0") === 0`), `values[name]` becomes `0`, so `0 || ""` evaluates to `""` and the field renders empty. In other words, typing `0` still visually clears the field — the exact bug being fixed (`-0` is affected too, since `-0 || "" === ""`). Apply the same explicit-empty check here, e.g. `value={values[name] ?? ""}`, which preserves `0`/`-0` while still blanking on `undefined`/`null`. This is only visible in the rendered component, so it's worth verifying end-to-end rather than just the parse logic in a REPL.

paoloantinori (commented):
The `onChange` fix correctly stops discarding a parsed `0` by replacing the `x || ""` falsy-fallback with an explicit `val === ""` / `isNaN` check, but the same root-cause pattern is still present on the `value` prop one line above, so a stored `0` still renders as an empty field and the reported symptom is only partially fixed.

- **`value` prop still clears `0`** (`ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx:152`): the unchanged `value={values[name] || ""}` evaluates `0 || ""` → `""`. Because the new `onChange` now deliberately stores the number `0` via `setValue(name, parsed)` (verified: `parseInt("0") === 0`, `typeof` `number`), the field re-renders with an empty string and visually clears — the exact behavior this PR aims to fix. Typing `0` still appears to blank the field; `-0` hits the same path (`-0 || "" === ""`). This line needs the same explicit treatment, e.g. `value={values[name] ?? ""}` (verified: `0 ?? "" === 0`), to actually retain a displayed `0`.
- **Verification gap**: the PR states the fix was checked only in a Node REPL (the parse logic in isolation). The display bug above is only observable in the rendered component, so it's worth confirming end-to-end — type `0` into the field and assert it stays `0` — before merging, since the REPL check would not have surfaced it.

jayaprajapatii:
Hii @paoloantinori 

Thanks for catching this — you're right, I missed the `value` prop using the same falsy-fallback pattern. Now I've fixed it.

I've now verified this end-to-end in the running UI:
* Typed "0" into the count field → field correctly retains 0  (previously cleared immediately)
* Clicked Render → output correctly shows the substituted value  ("You need 0 items to complete.")

carlesarnal:
Reviewed the latest changes (head commit 6ad27fb). Both issues from @paoloantinori's review look properly addressed:

1. **`value` prop** (`PromptTemplateTestPanel.tsx:152`): `values[name] || ""` → `values[name] ?? ""` — correctly preserves `0` since nullish coalescing only falls back on `null`/`undefined`, not on falsy values like `0`.

2. **`onChange` handler** (`PromptTemplateTestPanel.tsx:153-156`): The explicit `val === ""` / `isNaN(parsed)` logic handles all the edge cases cleanly — empty input, non-numeric input, and valid `0` input.

I also checked the rest of the file for similar patterns. The other `|| ""` usages (string type on ~line 181, array/object type on ~line 170) are fine since those branches never store a numeric `0`.

One tiny nit (non-blocking): `parseInt(val)` could include an explicit radix (`parseInt(val, 10)`) as a best practice, but since the input comes from a `type="number"` HTML element, it's purely cosmetic here.

The fix is clean and minimal. Looks good to me.

paoloantinori:
Thanks @carlesarnal — agreed, both points are addressed on this head. The `values[name] ?? ""` nullish-coalescing correctly preserves `0`, and the explicit `val === ""` / `isNaN(parsed)` check in the `onChange` handler covers the empty-input and unparseable cases cleanly. LGTM from my side.

jayaprajapatii:
Hello @carlesarnal, @paoloantinori 
please have a look at this PR if anything else needed from my side for this to move forward?"

paoloantinori:
Thanks @jayaprajapatii. From my side this looks good — both points from my earlier review are addressed on the current head (`values[name] ?? ""` preserves `0`, and the `val === ""` / `isNaN(parsed)` check in `onChange` is correct). SonarQube is green too. Nothing further needed from my review; just waiting on a maintainer to merge.

EricWittmann:
Hi @jayaprajapatii — thanks for the fix, but this was already resolved by PR #8681 which was merged earlier. The conflict you're seeing is because those exact lines were already changed on main with the same fix (replacing `||` with `??` and adding `isNaN` handling).

Closing as duplicate. Sorry about the overlap — we're working on better issue assignment to prevent this.

PR #9141, opened 2026-08-01
fix(ui): align Agent Card creation UI with A2A Protocol v1.0 schema

### LINKED ISSUE #8991: Agent Card creation UI is built against a stale schema and cannot satisfy the backend's A2A v1.0 validator
Background

Apicurio Registry recently added support for Agent Cards (artifact type AGENT_CARD), which let a registry store and expose metadata describing an AI agent — its name, capabilities, skills, and how other systems can reach it — following the A2A (Agent2Agent) protocol. The backend's own validator comment states it is aligned with A2A Protocol v1.0, and the web UI has a dedicated "Agents" page (A2A Agent Discovery) with its own "Create Agent" wizard for authoring these cards by hand.

While going through this creation flow, I noticed the wizard's second step ("Agent Card Content") only ever marks a single field, Name, as required, and has no field at all for something called supportedInterfaces — a concept the backend's validator treats as mandatory. That prompted me to cross-reference the frontend's data model against the backend's, which is where this issue comes from.

The core problem

The Agent Card creation flow in the web UI (ui/ui-app) is built against an outdated, pre-v1.0 shape of the AgentCard model. Concretely, this shows up in three separate, independently-verifiable places in the frontend, all of which disagree with the backend:

The TypeScript data model itself is missing fields the backend requires (most importantly supportedInterfaces, which has zero references anywhere in the entire ui/ui-app/src source tree).
The creation form only visually marks one field, Name, as required — even though the backend's validator requires several more.
The creation modal that submits the form performs no validation at all on the Agent Card's content (name, description, version, capabilities, skills) before sending it to the server — only the artifact's groupId/artifactId identifiers are checked client-side.

Put together, this means it is currently impossible to create a fully spec-compliant Agent Card through the UI — the create call always fails once strict (FULL) content validation is enabled on the registry, and it fails by listing violations for fields the form never gave the user any way to fill in. Below is the exact evidence for each of the three points above, followed by a live reproduction.

1) The UI's TypeScript model — AgentCard interface, ui/ui-app/src/app/components/agentCard/AgentCardViewer.tsx:35-47:

ts
export interface AgentCard {
    name: string;
    description?: string;
    version?: string;
    url?: string;
    provider?: AgentProvider;
    capabilities?: AgentCapabilities;
    skills?: AgentSkill[];
    defaultInputModes?: string[];
    defaultOutputModes?: string[];
    authentication?: AgentAuthentication;
    supportsExtendedAgentCard?: boolean;
}

Only name is non-optional. supportedInterfaces does not appear anywhere in this interface, and a search of the entire ui/ui-app/src tree for supportedInterfaces returns zero matches — the field has no representation anywhere in the frontend.

2) The create form — ui/ui-app/src/app/components/agentCard/AgentCardEditor.tsx:112 — only marks name as required:

tsx
<FormGroup label="Name" isRequired fieldId="agent-name">

No other FormGroup in this file carries isRequired, despite description, version, capabilities, and skills all being present as form inputs.

3) The create modal performs no content validation — ui/ui-app/src/app/pages/agents/components/modals/CreateAgentModal.tsx:83,129-167 only validates groupId and artifactId (via checkIdValid/validateField). Nothing in this file validates name, description, version, capabilities, skills, or supportedInterfaces before submission — so whatever the form's state happens to be gets sent straight to the server.

What the backend actually requires, for comparison

The backend's REST bean — AgentCard.java, app/src/main/java/io/apicurio/registry/a2a/rest/beans/AgentCard.java:19-61:

java
private String name;               // line 19
private String description;        // line 22
private String version;             // line 25
private String protocolVersion;     // line 28
private AgentProvider provider;     // line 31
private AgentCapabilities capabilities; // line 34
private List<AgentSkill> skills;    // line 37
private List<String> defaultInputModes;  // line 40
private List<String> defaultOutputModes; // line 43
private List<AgentInterface> supportedInterfaces; // line 46
private Map<String, SecurityScheme> securitySchemes; // line 49
private List<SecurityRequirement> securityRequirements; // line 52
private String iconUrl;             // line 55
private String documentationUrl;    // line 58
private List<AgentCardSignature> signatures; // line 61

The backend's content validator — AgentCardContentValidator.java, schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java — enforces the following as required, each raising a RuleViolation if absent:

description — line 72
version — line 73
supportedInterfaces (must exist, be an array, and be non-empty) — lines 75-83
capabilities — lines 86-88
skills (must exist, be an array, and be non-empty) — lines 91-98
defaultInputModes — line 103
defaultOutputModes — line 107

Every one of these except capabilities/skills/description/version themselves being present as inputs is either absent from the form's required-field marking, or (in the case of supportedInterfaces) absent from the form entirely.

Environment

Local build from main. Backend: Quarkus dev mode (app, JVM mode). UI: Vite dev server (ui/ui-app). Reproduced by interacting with the UI directly in a browser (Chrome) against the local backend at localhost:8080, UI served at localhost:8888.

Steps to Reproduce
Start the registry and enable strict content validation so the mismatch surfaces:
   POST /apis/registry/v3/admin/rules

   {"ruleType":"VALIDITY","config":"FULL"}
In the UI, navigate to the Agents page and click Create Agent.
Step 1 (Agent Coordinates): leave the defaults (Group ID default, Artifact ID blank), click Next.
Step 2 (Agent Card Content): enter only a Name (e.g. "Test Agent") — the only field the form marks as required. Leave every other field at its default/empty state: no Description, no Version, no URL, no Provider, no changes to Input/Output Modes beyond their defaults, no Authentication scheme selected, no Capabilities checked, and no Skills added.

Image

Click Next through Step 3 (Metadata), then click Create.
Expected vs Actual Behaviour

Expected: Either (a) the form collects every field the backend's validator requires — including supportedInterfaces, which currently has no corresponding input anywhere in the UI — or (b) if a field is genuinely optional at the UI level, the backend's required-field set for FULL validity should not diverge from what the UI actually allows a user to provide.

Actual: The create call fails with a 400 RuleViolationException. The response lists violations for fields the form gave no way to avoid:

json
{
  "causes": [
    { "description": "Agent Card must have a 'capabilities' field", "context": "/capabilities" },
    { "description": "Agent Card must have a 'version' field", "context": "/version" },
    { "description": "Agent Card must have a 'skills' field", "context": "/skills" },
    { "description": "Agent Card must have a 'description' field", "context": "/description" },
    { "description": "Agent Card must have a 'supportedInterfaces' field", "context": "/supportedInterfaces" }
  ],
  "detail": "Invalid Agent Card Causes: Agent Card must have a 'capabilities' field...",
  "title": "Invalid Agent Card",
  "status": 400,
  "name": "RuleViolationException"
}

The /supportedInterfaces violation in particular can never be resolved through the UI as it stands today, since no input for it exists anywhere in the create/edit form, regardless of what the user fills in.

Separately, with FULL validity turned off (the default), the same minimal submission succeeds instead of failing — meaning the UI will silently persist Agent Cards in an incomplete, pre-v1.0 shape (a single url string instead of a supportedInterfaces array, no securitySchemes) whenever strict validation isn't explicitly enabled.

Logs

Image

Suggested direction

Update the AgentCard interface and the corresponding editor/create-modal components under ui/ui-app/src/app/components/agentCard/ and ui/ui-app/src/app/pages/agents/components/modals/ to match the shape used by AgentCard.java, in particular adding an input for supportedInterfaces. Adding client-side validation in CreateAgentModal.tsx for the fields AgentCardContentValidator treats as required would also let failures surface before submission instead of as an unmapped backend error.

I'd be glad to put together a PR for this — would a maintainer be able to confirm whether this direction sounds right before I start (in particular, whether supportedInterfaces should become a first-class repeatable input, and whether the old url/authentication fields should be removed outright or kept temporarily for backward compatibility)? If so, I'd appreciate being assigned to it.

Mohammad-Raees:
## Summary

Fixes #8991

The Create Agent wizard and the `AgentCard` TypeScript model were built against a pre-v1.0 shape of the A2A spec. Any attempt to create an Agent Card with full content validation enabled fails with a `400 RuleViolationException` because the form never gives the user a way to fill in several required fields (`description`, `version`, `supportedInterfaces`).

Three independent defects are fixed together because they all stem from the same root cause — the TypeScript model and form were never updated after the A2A Protocol advanced to v1.0:

---

### 1. `AgentCardViewer.tsx` — update `AgentCard` TypeScript interface

- Add `AgentInterface` type `{ url, protocolBinding, protocolVersion }` matching the backend's `AgentCardContentValidator` validation rules
- Add `supportedInterfaces?: AgentInterface[]` to `AgentCard`
- Add optional `protocolVersion`, `iconUrl`, `documentationUrl`, `securitySchemes`, `securityRequirements` fields
- Mark `description` and `version` with JSDoc REQUIRED annotations
- Add `renderSupportedInterfaces()` so the read-only viewer shows all interface entries

### 2. `AgentCardEditor.tsx` — add Supported Interfaces repeater

- Mark **Description** and **Version** `FormGroup`s as `isRequired`
- Add a **Supported Interfaces** section: users can add/remove interface entries, each with URL, Protocol Binding, and Protocol Version inputs — the three fields `AgentCardContentValidator` enforces per entry

### 3. `CreateAgentModal.tsx` — full content-step validation

- Seed `EMPTY_AGENT_CARD` with `description: ""`, `version: ""`, `supportedInterfaces: []`
- Replace the name-only guard in `isContentStepValid()` with a complete check mirroring the backend validator:
  - `name`, `description`, `version` must be non-empty strings
  - At least one `supportedInterface` must exist and be fully filled (url, protocolBinding, protocolVersion)
  - `capabilities` must be present
  - At least one `skill` must be present
- This prevents the wizard from reaching the **Create** button when the submitted payload would fail server-side validation

---

## Validation

Cross-referenced against `AgentCardContentValidator.java` to ensure every field the backend enforces as required is now either collected by the form or validated in `isContentStepValid()`.

- `npm run lint` — passes (0 errors)
- `tsc --noEmit` — passes (0 errors)

### DIFF
--- a/ui/ui-app/src/app/components/agentCard/AgentCardEditor.tsx
+++ b/ui/ui-app/src/app/components/agentCard/AgentCardEditor.tsx
@@ -102,7 +102,7 @@ export const AgentCardEditor: FunctionComponent<AgentCardEditorProps> = (props: 
     };
 
     const handleAddInterface = (): void => {
-        const newInterface: AgentInterface = { url: "", protocolBinding: "A2A", protocolVersion: "1.0" };
+        const newInterface: AgentInterface = { id: crypto.randomUUID(), url: "", protocolBinding: "A2A", protocolVersion: "1.0" };
         updateField("supportedInterfaces", [...(agentCard.supportedInterfaces || []), newInterface]);
     };
 
@@ -176,7 +176,7 @@ export const AgentCardEditor: FunctionComponent<AgentCardEditorProps> = (props: 
                     </Title>
 
                     {(agentCard.supportedInterfaces || []).map((iface, index) => (
-                        <Card isPlain key={index} className="interface-entry">
+                        <Card isPlain key={iface.id || index} className="interface-entry">
                             <CardBody>
                                 <FormGroup label="URL" isRequired fieldId={`interface-url-${index}`}>
                                     <InputGroup>
--- a/ui/ui-app/src/app/components/agentCard/AgentCardEditor.tsx
+++ b/ui/ui-app/src/app/components/agentCard/AgentCardEditor.tsx
@@ -1,6 +1,9 @@
 import { FunctionComponent, useState } from "react";
 import "./AgentCardEditor.css";
 import {
+    ActionList,
+    ActionListItem,
+    Button,
     Card,
     CardBody,
     CardHeader,
@@ -20,9 +23,10 @@ import {
     InputGroup,
     InputGroupItem
 } from "@patternfly/react-core";
+import { PlusCircleIcon, TrashIcon } from "@patternfly/react-icons";
 import { CapabilityEditor } from "./CapabilityEditor";
 import { SkillEditor } from "./SkillEditor";
-import { AgentCard } from "./AgentCardViewer";
+import { AgentCard, AgentInterface } from "./AgentCardViewer";
 import { AgentCapabilities } from "./AgentCardCapabilities";
 import { AgentSkill } from "./AgentCardSkills";
 
@@ -97,6 +101,23 @@ export const AgentCardEditor: FunctionComponent<AgentCardEditorProps> = (props: 
         });
     };
 
+    const handleAddInterface = (): void => {
+        const newInterface: AgentInterface = { url: "", protocolBinding: "A2A", protocolVersion: "1.0" };
+        updateField("supportedInterfaces", [...(agentCard.supportedInterfaces || []), newInterface]);
+    };
+
+    const handleRemoveInterface = (index: number): void => {
+        const updated = (agentCard.supportedInterfaces || []).filter((_, i) => i !== index);
+        updateField("supportedInterfaces", updated);
+    };
+
+    const handleUpdateInterface = (index: number, field: keyof AgentInterface, value: string): void => {
+        const updated = (agentCard.supportedInterfaces || []).map((iface, i) =>
+            i === index ? { ...iface, [field]: value } : iface
+        );
+        updateField("supportedInterfaces", updated);
+    };
+
     return (
         <Card className={`agent-card-editor ${className || ""}`}>
             <CardHeader>
@@ -118,7 +139,7 @@ export const AgentCardEditor: FunctionComponent<AgentCardEditorProps> = (props: 
                         />
                     </FormGroup>
 
-                    <FormGroup label="Description" fieldId="agent-description">
+                    <FormGroup label="Description" isRequired fieldId="agent-description">
                         <TextArea
                             id="agent-description"
                             value={agentCard.description || ""}
@@ -128,7 +149,7 @@ export const AgentCardEditor: FunctionComponent<AgentCardEditorProps> = (props: 
                         />
                     </FormGroup>
 
-                    <FormGroup label="Version" fieldId="agent-version">
+                    <FormGroup label="Version" isRequired fieldId="agent-version">
                         <TextInput
                             id="agent-version"
                             value={agentCard.version || ""}
@@ -146,6 +167,67 @@ export const AgentCardEditor: FunctionComponent<AgentCardEditorProps> = (props: 
                             placeholder="https://example.com/agent"
                         />
                     </FormGroup>
+
+                    <Divider className="section-divider" />
+
+                    {/* Supported Interfaces Section (A2A v1.0 required) */}
+                    <Title headingLevel="h3" className="section-header">
+                        Supported Interfaces <span style={{ color: "var(--pf-t--global--color--status--danger--default)" }}>*</span>
+                    </Title>
+
+                    {(agentCard.supportedInterfaces || []).map((iface, index) => (
+                        <Card isPlain key={index} className="interface-entry">
+                            <CardBody>
+                                <FormGroup label="URL" isRequired fieldId={`interface-url-${index}`}>
+                                    <InputGroup>
+                                        <InputGroupItem isFill>
+                                            <TextInput
+                                                id={`interface-url-${index}`}
+                                                type="url"
+                                                value={iface.url}
+                                                onChange={(_event, value) => handleUpdateInterface(index, "url", value)}
+                                                placeholder="https://agent.example.com/a2a"
+                                            />
+                                        </InputGroupItem>
+                                    </InputGroup>
+                                </FormGroup>
+                                <FormGroup label="Protocol Binding" isRequired fieldId={`interface-binding-${index}`}>
+                                    <TextInput
+                                        id={`interface-binding-${index}`}
+                                        value={iface.protocolBinding}
+                                        onChange={(_event, value) => handleUpdateInterface(index, "protocolBinding", value)}
+                                        placeholder="e.g., A2A, JSONRPC"
+                                    />
+                                </FormGroup>
+                                <FormGroup label="Protocol Version" isRequired fieldId={`interface-version-${index}`}>
+                                    <TextInput
+                                        id={`interface-version-${index}`}
+                                        value={iface.protocolVersion}
+                                        onChange={(_event, value) => handleUpdateInterface(index, "protocolVersion", value)}
+                                        placeholder="e.g., 1.0"
+                                    />
+                                </FormGroup>
+                                <ActionList>
+                                    <ActionListItem>
+                                        <Button
+                                            variant="danger"
+                                            icon={<TrashIcon />}
+                                            onClick={() => handleRemoveInterface(index)}
+                                        >
+                                            Remove interface
+                                        </Button>
+                                    </ActionListItem>
+                                </ActionList>
+                            </CardBody>
+                        </Card>
+                    ))}
+                    <Button
+                        variant="link"
+                        icon={<PlusCircleIcon />}
+                        onClick={handleAddInterface}
+                    >
+                        Add interface
+                    </Button>
 
                     <Divider className="section-divider" />
 
--- a/ui/ui-app/src/app/components/agentCard/AgentCardViewer.tsx
+++ b/ui/ui-app/src/app/components/agentCard/AgentCardViewer.tsx
@@ -31,10 +31,14 @@ export interface AgentProvider {
 
 /**
  * A single transport interface through which the agent can be reached.
- * Each entry in supportedInterfaces must supply all three fields to pass
- * the backend's AgentCardContentValidator (A2A Protocol v1.0).
+ * Each entry in supportedInterfaces must supply url, protocolBinding, and
+ * protocolVersion to pass the backend's AgentCardContentValidator (A2A Protocol v1.0).
+ *
+ * `id` is a UI-only stable key used for React list reconciliation and is
+ * never sent to the backend.
  */
 export interface AgentInterface {
+    id?: string;
     url: string;
     protocolBinding: string;
     protocolVersion: string;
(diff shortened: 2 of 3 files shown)


ui/ui-app/src/app/components/agentCard/AgentCardEditor.tsx
```diff
@@ -149,6 +170,67 @@ export const AgentCardEditor: FunctionComponent<AgentCardEditorProps> = (props:
 
                     <Divider className="section-divider" />
 
+                    {/* Supported Interfaces Section (A2A v1.0 required) */}
+                    <Title headingLevel="h3" className="section-header">
+                        Supported Interfaces <span style={{ color: "var(--pf-t--global--color--status--danger--default)" }}>*</span>
+                    </Title>
+
+                    {(agentCard.supportedInterfaces || []).map((iface, index) => (
+                        <Card isPlain key={index} className="interface-entry">
```
paoloantinori: This repeater renders with `key={index}`. Removing a non-last interface shifts the remaining items' positions, so React reuses the shifted components' DOM — input focus can land on the wrong field after a middle deletion (the underlying values stay correct because the inputs are controlled). Consider assigning a stable unique id in `handleAddInterface` (e.g. `crypto.randomUUID()`) and keying on it instead of the array index.

ui/ui-app/src/app/components/agentCard/AgentCardViewer.tsx
```diff
@@ -30,20 +30,44 @@ export interface AgentProvider {
...
+
+/**
+ * Full Agent Card structure aligned with A2A Protocol v1.0.
+ * Fields marked with REQUIRED must be non-empty to pass FULL content validation.
  */
 export interface AgentCard {
+    /** REQUIRED – human-readable name of the agent. */
     name: string;
+    /** REQUIRED – human-readable description of what the agent does. */
     description?: string;
```
paoloantinori, line 55: `description` (and `version`) carry a `/** REQUIRED */` JSDoc but are declared optional (`description?: string`). The runtime gate in `isContentStepValid` covers submission, so this isn't a functional hole — but the type and the documented contract disagree. If the intent is compile-time enforcement, drop the `?`; otherwise soften the JSDoc (e.g. "REQUIRED by the A2A v1.0 contract; optional in the TS model for legacy/partial cards") so the two agree.

paoloantinori (commented):
Overall the change correctly achieves its goal: `isContentStepValid()` now gates the wizard on every field the backend `AgentCardContentValidator` requires, with sound short-circuit ordering, and the Supported Interfaces repeater handles add/remove/update immutably with no off-by-one. Two independent passes found no functional or blocker-level defect; the observations below are minor easy-wins.

- `ui/ui-app/src/app/pages/agents/components/modals/CreateAgentModal.tsx:87-99` — the validation chain is comprehensive and satisfiable; `capabilities !== undefined` and `skills.length > 0` are reachable because `CapabilityEditor`/`SkillEditor` populate those via `updateField`, and `Array.isArray(supportedInterfaces)` correctly guards the subsequent `.length`/`.every()` calls. No path enables the Create button with an invalid payload.
- `ui/ui-app/src/app/components/agentCard/AgentCardEditor.tsx:179` — the repeater uses `key={index}`; values stay correct (inputs are controlled) but focus can land on the wrong field after a middle-item removal. A stable unique id is a one-line fix.
- `ui/ui-app/src/app/components/agentCard/AgentCardViewer.tsx:51` — `description?: string` and `version?: string` are optional in the type while their JSDoc says `REQUIRED`; not a runtime hole (the form gate covers it), but the type and documented contract disagree.

Mohammad-Raees:
Thank you for the thorough review, @paoloantinori! Both observations are addressed in the follow-up commit (a7f4c4c).

**[AgentCardEditor.tsx line 179](https://github.com/Apicurio/apicurio-registry/pull/9141#discussion_r3695500911) — `key={index}` on the Supported Interfaces repeater**

Fixed. Added an optional `id?: string` field to `AgentInterface`, documented explicitly as a UI-only stable key that is never sent to the backend. `handleAddInterface` now sets `id: crypto.randomUUID()` on each new entry, and the repeater keys on `iface.id` (with `|| index` as a safe fallback for any legacy entries already persisted in the store). Mid-list removals will no longer cause input focus to land on the wrong field.

**[AgentCardViewer.tsx line 51](https://github.com/Apicurio/apicurio-registry/pull/9141#discussion_r3695500912) — `/** REQUIRED *\/` JSDoc vs `?` optional modifier**

Fixed by softening the JSDoc rather than dropping the `?`. Removing the `?` would break existing code that accesses both fields with optional chaining (`agentCard.description &&`, `agentCard.version &&`) and would be incorrect for pre-A2A-v1.0 cards that may legitimately omit them. The comments now read:

> _"REQUIRED by A2A v1.0 contract; optional in the TS model to accommodate partial or legacy cards"_

This aligns the documented contract with the TypeScript model. Applied the same treatment to `supportedInterfaces`.

paoloantinori:
Hi @Mohammad-Raees — thanks for the follow-up. Confirmed both fixes landed cleanly on this head:

1. **`key={index}` → `key={iface.id || index}`** — the optional `id?: string` on `AgentInterface`, set via `crypto.randomUUID()` in `handleAddInterface`, is the right call. Documenting it as a UI-only key never sent to the backend avoids any confusion about whether the backend should round-trip it. Good.

2. **Supported Interfaces repeater** — the three required fields (`url`, `protocolBinding`, `protocolVersion`) per entry are wired through `handleUpdateInterface`, and `isContentStepValid()` now short-circuits on every field the backend `AgentCardContentValidator` enforces, including the `every(...)` over interface entries. Ordering is sound.

One small thing worth a quick look (non-blocking): in `AgentCardViewer.tsx` the fallback key for the read-only list is `iface.id || iface.url || index` — fine for display, just noting it diverges slightly from the editor's key. No action needed unless you want them identical.

LGTM from my side. Leaving the merge call to the maintainers.

carlesarnal:
## Code Review

Good work aligning the UI with the A2A v1.0 spec. Two items need attention before merge:

**1. Missing DCO sign-off** — Per CONTRIBUTING.md, both commits need Signed-off-by trailers. Fix: git rebase HEAD~2 --signoff

**2. AgentInterface.id leaks to backend** — CreateAgentModal.tsx:74 uses JSON.stringify(agentCard) which includes the crypto.randomUUID() id on each AgentInterface (AgentCardEditor.tsx:104). The JSDoc (AgentCardViewer.tsx:47) says this is never sent to the backend but nothing strips it. Strip id from each interface entry before serialization in handleCreate().

Everything else is solid: validation correctly mirrors AgentCardContentValidator, repeater is immutable, and paoloantinori feedback is addressed.

carlesarnal:
Suggested fix for item 2 above — in handleCreate() before JSON.stringify, strip the UI-only id field:

const sanitizedCard = { ...agentCard, supportedInterfaces: agentCard.supportedInterfaces?.map(({ id, ...rest }) => rest) };

Then use JSON.stringify(sanitizedCard, null, 2) instead of JSON.stringify(agentCard, null, 2).

For item 1, git rebase HEAD~2 --signoff followed by git push --force-with-lease will add Signed-off-by to both commits.

Mohammad-Raees:
Thank you @carlesarnal! Both items are addressed in the latest push (641df71).

**1. DCO Signed-off-by**
Added `Signed-off-by: Mohammad Raees <mohammadraees3rd@gmail.com>` to all three commits via `git rebase origin/main --signoff`. The DCO bot should now pass.

**2. AgentInterface.id leaking to backend**
Fixed exactly as you suggested — `handleCreate()` in `CreateAgentModal.tsx` now builds a sanitized copy before serialization:

```ts
const sanitizedCard = {
    ...agentCard,
    supportedInterfaces: agentCard.supportedInterfaces?.map(({ id: _id, ...rest }) => rest)
};
// JSON.stringify(sanitizedCard, null, 2)
```

The UI-only `id` field is stripped from every interface entry before the payload reaches the backend. The JSDoc and the runtime behaviour are now consistent.

Mohammad-Raees:
Hi @EricWittmann — thank you for taking a look at this PR. I really appreciate it.

I'm happy to address any review feedback quickly. Also, whenever you have a moment, the workflow is still awaiting maintainer approval.

Thanks again for your time.

Mohammad-Raees:
Hi @EricWittmann - just a gentle note that this PR is ready whenever you have a chance. All the earlier review feedback has been addressed. No rush at all, and thank you for looking when you can.

EricWittmann:
Thanks for the contribution, @Mohammad-Raees! Great work aligning the UI with the A2A v1.0 spec. Merged.

PR #9145, opened 2026-08-01
fix(rest): expand reference graph nodes that are reached at a shallower depth

### LINKED ISSUE #9078: Reference graph endpoint drops nodes that are reachable at more than one depth
## Description

**Registry Version**: 3.3.1-SNAPSHOT, main at commit ba902c35a
**Persistence type**: sql

`GET /groups/{groupId}/artifacts/{artifactId}/versions/{version}/references/graph` leaves nodes out of the graph it returns. The endpoint answers 200 and the response looks well formed, so there is no sign that anything is missing.

`GroupsResourceImpl.traverseOutboundReferences` adds a node to `visited` at line 378. Any later encounter of that node is skipped at line 373. The recursion also stops when `currentDepth > maxDepth` at line 336. So if a node is first reached at the depth limit, its own references are never read. When the traversal later reaches the same node on a shorter path, it is skipped because it is already in `visited`. Its references stay unread and its children never enter the graph.

References are stored as a JSON array and read back in the order the client supplied them. That order decides which path the traversal walks first, so it also decides whether nodes go missing.

`traverseInboundReferences` has the same problem at lines 466 to 471.

### Environment

Temurin JDK 21.0.12, Maven 3.9.8. Registry started from `app/target/quarkus-app/quarkus-run.jar` with the default configuration, which is sql storage on in-memory H2. Nothing here depends on the storage variant.

### Steps to Reproduce

Create four artifacts in group `g` so that one of them is reachable by both a long and a short path. `root` lists `mid` before `leaf`.

- `deep` with no references
- `leaf` referencing `deep`
- `mid` referencing `leaf`
- `root` referencing `mid` and `leaf`, in that order

```
curl -X POST http://localhost:8080/apis/registry/v3/groups/g/artifacts \
  -H 'Content-Type: application/json' \
  -d '{"artifactId":"root","artifactType":"AVRO","firstVersion":{"content":{
        "content":"{\"type\":\"record\",\"name\":\"RecordRoot\",\"namespace\":\"com.example\",\"fields\":[{\"name\":\"id\",\"type\":\"string\"}]}",
        "contentType":"application/json",
        "references":[
          {"groupId":"g","artifactId":"mid","version":"1","name":"mid.avsc"},
          {"groupId":"g","artifactId":"leaf","version":"1","name":"leaf.avsc"}
        ]}}}'
```

Then request the graph with a depth of 2:

```
curl 'http://localhost:8080/apis/registry/v3/groups/g/artifacts/root/versions/1/references/graph?depth=2'
```

Expected 4 nodes and 4 edges. `leaf` sits at depth 1 through the direct reference from `root`, so `deep` sits at depth 2. Both are inside the requested depth.

Actual 3 nodes and 3 edges. `deep` is missing.

```
"nodes":  [ g:root:1, g:mid:1, g:leaf:1 ]
"edges":  [ root -> mid, mid -> leaf, root -> leaf ]
"metadata": {"totalNodes":3,"totalEdges":3,"maxDepth":2,"hasCycles":false}
```

The traversal reaches `leaf` through `mid` first, which puts it at depth 2, so the recursion into `leaf` is cut off and its reference to `deep` is never read. When `root -> leaf` is processed afterwards, `leaf` is already in `visited` and is skipped.

Listing `root`'s references as `leaf, mid` instead returns the complete graph for the same data:

```
"nodes":  [ g:root:1, g:leaf:1, g:deep:1, g:mid:1 ]
"edges":  [ root -> leaf, leaf -> deep, root -> mid, mid -> leaf ]
"metadata": {"totalNodes":4,"totalEdges":4,"maxDepth":2,"hasCycles":false}
```

So the same registry answers the same request differently depending only on the order the references were stored in. `maxDepth` reports 2 in both cases, so the metadata gives no hint that the first result is short.

### Logs

Nothing is logged. The request returns 200 and the server log has no warning or error for it. No exception is involved, the nodes are dropped by the `visited` check on the normal path.

I think the fix is to record the depth at which each node was expanded and expand it again when it is later reached at a shallower depth. Traversing breadth-first would also work, since a node is then always reached by its shortest path first. `ReferenceGraphIT.testReferenceGraphWithDepthLimit` only covers a straight chain, which is why this is not caught today.

Akshats-git:
Fixes Apicurio/apicurio-registry#9078.

## Problem

The reference graph endpoint can leave nodes out of the graph it returns. `traverseOutboundReferences` records every node it sees in one global `visited` set and skips the node on any later encounter. The recursion also stops when `currentDepth` exceeds `maxDepth`. A node that is first reached at the depth limit therefore has its own references cut off, and when a shorter path to it turns up later it is skipped because it is already in the set. Its references stay unread and its children never enter the graph.

References are stored as a JSON array and read back in the order they were supplied, so which nodes go missing depends on the order the references were registered in. `traverseInboundReferences` has the same problem.

For example, take four artifacts where `root` references `mid` and `leaf` in that order, `mid` references `leaf`, and `leaf` references `deep`. Requesting the graph at a depth of 2 returns 3 nodes and 3 edges and omits `deep`, even though `leaf` sits at depth 1 through the direct reference from `root`, which puts `deep` at depth 2. Registering `root`'s references as `leaf, mid` instead returns all 4 nodes for the same data.

## Fix

- Replace the global `visited` set with `Map<String, Integer> expandedAt`, holding the shallowest depth at which each node has had its references expanded. Presence in the map means the node is already in the graph. `Integer.MAX_VALUE` means it was added but never expanded.
- Expand a node when it has not been expanded yet, or when it is reached at a strictly shallower depth than before. This terminates, because each re-expansion happens at a strictly smaller depth.
- Node creation is now separate from expansion, so each node is still added to the graph at most once.
- De-duplicate edges before building the response. A node can now be expanded more than once, so the same reference can be walked more than once.
- Apply the same change to `traverseInboundReferences`.
- Cycle handling is unchanged. `currentPath` still tracks the artifact level path and the traversal still does not continue through a cycle node.

## Testing

Added two tests to `ReferenceGraphIT`. Both build the same four artifact shape and differ only in the order of the root artifact's references. Both assert the same complete graph, so the answer no longer depends on that order.

`ReferenceGraphIT` passes in full: 11 tests, 0 failures, 0 errors, 0 skipped. `checkstyle:check` passes for `app` and `integration-tests`.

Also verified against a locally built registry on sql storage with in-memory H2:

- both reference orders return 4 nodes and 4 edges
- `depth=1` still returns 3 nodes and 2 edges, so `deep` correctly stays out of range
- `depth=3`, `depth=10` and `depth=0` return 4 nodes and 4 edges with no duplicate edges, which covers the case where a node really is expanded twice
- the existing scenarios are unchanged: empty graph, the depth limited chain at depth 1 and 2, inbound at depth 1 and 2, cycle detection, and 404 for a missing artifact

## Notes

`expandedAt` is shared by the outbound and inbound passes, exactly as `visited` was. This keeps `BOTH` behaving as it does today rather than changing it as a side effect of this fix. A node discovered by one direction is still not expanded by the other. That is a pre-existing limitation and is better handled on its own.

Two things I was not able to finish. I have not run the `app` module unit suite to completion, because my laptop hangs the moment I run it. And I did not run the new tests against an unfixed build, though I did confirm over HTTP that the unfixed build returns 3 nodes and 3 edges for the shape those tests use, and they assert 4 of each.


app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -369,50 +385,57 @@ private void traverseOutboundReferences(String groupId, String artifactId, Strin
...
+                // been reached at. Reaching it again on a shorter path leaves room for references
+                // that the earlier, deeper visit had to cut off.
+                if (expandedAt.get(targetNodeId) <= currentDepth) {
+                    continue;
                 }
+                expandedAt.put(targetNodeId, currentDepth);
+
+                // Add to current path before recursing, remove after (backtracking)
+                currentPath.add(targetArtifactKey);
+                traverseOutboundReferences(refGroupId, refArtifactId, refVersion, targetNodeId,
```
paoloantinori, line 454: A node can now be expanded once per distinct reachable depth (worst case O(nodes × edges) versus the previous O(nodes + edges)), and every expansion re-enters this method, which re-derives the node's reference list from storage. For the small graphs in the tests this is negligible, but on a large registry with many cross-references at a high or unlimited `depth`, the repeated re-fetching can amplify storage calls significantly. If per-node reference (and metadata) lookups are not already cached, memoizing them by node id would keep re-expansion as a cheap in-memory re-walk without changing the result. Worth a quick check against a denser fixture or an existing large-registry scenario before merge.

paoloantinori (commented):
Overall this is a sound fix for the order-dependent node omission: a node is still added to the graph exactly once, but it is now expanded at its shallowest reachable depth and re-expanded only when a strictly shorter path is found. That terminates, because each re-expansion strictly decreases the stored depth and depth is bounded below by 1, so a node is expanded at most once per distinct reachable depth. The new tests cover both reference orderings and assert the same complete graph.

- Core logic is correct in both directions. The guard `if (expandedAt.get(targetNodeId) <= currentDepth) continue;` (GroupsResourceImpl.java:428, mirrored in `traverseInboundReferences`) re-expands only on a strictly shallower reach, and node creation stays guarded by `!expandedAt.containsKey(...)`, so no duplicate nodes are produced even though a node may be expanded multiple times.
- Edge de-duplication (GroupsResourceImpl.java:316–326) correctly absorbs the duplicate edges that re-expansion can now emit, and both `totalEdges` and `edges` use the deduped list, so the metadata counts stay consistent with the returned edges.
- Cycle handling is preserved correctly: `currentPath` is still backtracked around each recursion, so re-expansion along a different branch gets a fresh path context and cycle detection is unaffected; cycle nodes are added once and never expanded.
- The one thing worth confirming before merge is cost at scale: a node can now be expanded once per distinct reachable depth, so worst-case work is O(nodes × edges) rather than the prior O(nodes + edges), and because `traverseOutboundReferences`/`traverseInboundReferences` re-derive a node's reference list from storage on each call, re-expansion also re-fetches. For small graphs this is negligible, but on a large registry with many cross-references at a high or unlimited `depth` the storage-call amplification could be material — see the inline note.
- Tests are well-targeted. `testSharedReferenceIsExpandedWhenDeepPathIsWalkedFirst` and `testSharedReferenceIsExpandedWhenShortPathIsWalkedFirst` (ReferenceGraphIT.java) build the same four-artifact shape in both root-reference orders and assert identical 4-node / 4-edge / maxDepth-2 results, which is exactly the regression being fixed.

Akshats-git:
Thanks @paoloantinori for the suggestion, I have addressed it in the latest push.

Each node's reference list is now cached for the duration of a single graph build, so re-expansion is an in-memory re-walk rather than a repeated storage read. `traverseOutboundReferences` takes a `Map<String, List<ArtifactReferenceDto>> outboundReferences` and `traverseInboundReferences` the equivalent `inboundReferences`, both keyed by node id and created per request in `buildReferenceGraph`. A failed lookup is cached as an empty list, so re-expansion does not retry a lookup that already failed.

One correction to the diagnosis: the metadata lookups were never amplified. Both `storage.getArtifactVersionMetaData` calls sit inside the `!expandedAt.containsKey(...)` guard, so they already fired at most once per node. The amplification was limited to the two reference-list reads, which is what the cache covers.

On the denser fixture, I built a layered DAG where every node in a layer references every node in the next, plus a root referencing both layer 0 and layer 1 directly, so layer-1 nodes are reachable at two depths and re-expansion triggers throughout. On sql storage with in-memory H2:

| fixture | depth | nodes | edges | duplicate edges | latency |
|---|---|---|---|---|---|
| 5 x 4, 21 artifacts | unlimited | 21 | 72 | 0 | 17 ms |
| 8 x 8, 65 artifacts | unlimited | 65 | 464 | 0 | 18 ms |
| 12 x 10, 121 artifacts | unlimited | 121 | 1120 | 0 | 18 ms |

Latency stays flat as the graph grows from 21 to 121 nodes and 72 to 1120 edges, consistent with each node being read once rather than once per reachable depth.

To be straight about that measurement: these are latency figures, not a before/after query count. Quarkus fixes `quarkus.log.min-level` at build time so the TRACE statements in `AbstractHandleFactory` are compiled out of the packaged app, and there is no JDBC query metrics endpoint I could diff. If you want a hard query count before merge, say how you would prefer to obtain it and I will add it.

One more issue that copilot pointed out is also fixed. The key is now `Arrays.asList(sourceNodeId, targetNodeId, name)` compared element-wise instead of a concatenated string, so null and empty names stay distinct without relying on a sentinel that a real name could collide with.

Worth noting I could not reproduce the reported case through the API. A reference with a null name is rejected with HTTP 500 at creation, so a stored reference always has a non-null name:

| reference shape | result |
|---|---|
| `name` omitted (null) | HTTP 500 |
| `name: ""` | 200 |
| `name: "x.avsc"` | 200 |

So the collision was latent rather than reachable. I kept the fix anyway since it costs nothing and removes the trap.

One visible behaviour change to call out while we are here: two references to the same target with the same name now produce one edge instead of two. That input is accepted today, and collapsing edges that are identical in every field seems right, but it is a change so I would rather flag it than let it pass silently.

Separately, the HTTP 500 above looks like its own bug, in the same class as #8611 and #8732. Happy to file it once this PR is resolved.

`ReferenceGraphIT` still passes in full (11 tests, 0 failures) and `checkstyle:check` is clean on `app`. The correctness checks are unchanged: both reference orders give the same 4-node graph, `depth=1` still excludes the out-of-range node, and the empty, chain, inbound, cycle and 404 scenarios behave as before.

paoloantinori:
Thanks @Akshats-git — the caching you added for the reference lists is the right fix. Per-build memoization keyed by node id means re-expansion during the BFS is an in-memory re-walk, which removes the repeated storage reads I flagged. Reads cleanly on this head; good from my side.

Akshats-git:
Hi @paoloantinori @EricWittmann, can I get an update on this PR? I am currently working on another issue and would like to open a PR for it, but the one-open-PR limit is preventing me from doing so. Thanks!

paoloantinori:
Thanks for the ping. I've reviewed the latest head (586a0ca) and have no further change requests. The reference-list memoization and the order-independent traversal look right to me.

One blocker for merge: GitHub reports this PR as "behind," so it needs a rebase against the current main before it can land. Once that's green I don't have anything else from my side, and the one-open-PR limit should free up after merge.

Akshats-git:
@paoloantinori Rebased onto current main (8d9be8cf) and force-pushed. GitHub no longer reports the PR as behind.

paoloantinori:
Thanks for the rebase. The new head (29d2678) is functionally identical to the version I reviewed at 586a0ca — the `expandedAt` map, the per-build reference memoization, and the unique-edge dedup are all unchanged. No new concerns from me.

EricWittmann:
Thanks for this excellent contribution, @Akshats-git! The fix is clean and well-tested. Merging now.

PR #9146, opened 2026-08-01
fix(storage): widen outbox aggregateid column to VARCHAR(2048)

### LINKED ISSUE #8793: Outbox table aggregate id field is too small
## Description

**Registry
Version**: 3.0.14
**Persistence type**: sql

This bug has existed since the original addition of the code in [Oct 9, 2024](https://github.com/apicurio/apicurio-registry/commit/5d756a9).. Outbox aggregateId field is too small at 255 characters to hold the concatenation of GroupId and ArtifactId (both max 512 each) and Version(max 256). We are currently working around this by patch fixing the image and using TEXT as the type instead of VARCHAR(255). This is a bad one since it didn't exist in 2.x and blocks our customers migrations from 2.x to 3.x

### Environment

Kubernetes v1.34.9

### Steps to Reproduce

Try and create a version where the cumulative size of GroupId+ArtifactId+VersionId is greater than 255 characters

EXPECTED: Creation should succeed as Max length of GroupId and ArtifactId is 512 each and max size of Version is 256
ACTUAL: Creation fails as aggregateId field in Outbox table is of max length 255 and this side effect failure rolls back the entire transaction

7vignesh:
## Summary

Fixes #8793. The outbox table's `aggregateid` column was `VARCHAR(255)`, but aggregate IDs are built by concatenating `groupId` (max 512) + `artifactId` (max 512) + `version` (max 256) with separators, reaching up to 1282 characters. The overflow caused INSERT failures that rolled back the entire artifact/version creation transaction, blocking migrations from 2.x to 3.x.

## Changes

**Schema (base DDLs):**
- `postgresql.ddl`: `aggregateid VARCHAR(255)` → `VARCHAR(2048)`
- `mysql.ddl`: `aggregateid VARCHAR(255)` → `VARCHAR(2048)`
- `mssql.ddl`: `aggregateid VARCHAR(255)` → `NVARCHAR(2048)` (also fixes VARCHAR/NVARCHAR inconsistency with other MSSQL identifier columns)

**Upgrade path (new `upgrades/110/` directory):**
- PostgreSQL: `ALTER TABLE outbox ALTER COLUMN aggregateid TYPE VARCHAR(2048)`
- MySQL: `ALTER TABLE outbox MODIFY COLUMN aggregateid VARCHAR(2048) NOT NULL`
- MSSQL: `ALTER TABLE outbox ALTER COLUMN aggregateid NVARCHAR(2048) NOT NULL`
- H2: version bump only (no outbox table)
- `db-version`: 109 → 110

**Tests:**
- `OutboxAggregateIdLengthTest`: 15 assertions validating DDL column widths, upgrade script consistency, cross dialect agreement, MSSQL NVARCHAR usage, and that `ArtifactVersionCreated.of()` with max length identifiers fits within bounds
- `AbstractRegistryStorageTest#testCreateArtifactWithMaxLengthIdentifiers`: storage level regression test with 130 char identifiers (aggregate id 393, exceeding old 255 limit). Runs on all backends including MSSQL.
- `PostgresqlStorageTest#testCreateArtifactWithTrueMaxLengthIdentifiers`: Postgres only test exercising the true maximum aggregate id (512/512/256 = 1282 chars). Cannot run on MSSQL due to 900 byte clustered index key limit on composite PKs.

## Testing

- `OutboxAggregateIdLengthTest`: 15/15 pass (9/15 fail when column is reverted to 255)
- `PostgresqlStorageTest#testCreateArtifactWithMaxLengthIdentifiers`: passes against real embedded Postgres
- `PostgresqlStorageTest#testCreateArtifactWithTrueMaxLengthIdentifiers`: passes, confirms 1282 char aggregate id works end to end
- `MssqlStorageTest#testCreateArtifactWithMaxLengthIdentifiers`: previously verified passing with Docker
- Confirmed failure without fix: `PSQLException: ERROR: value too long for type character varying(255)`
- `checkstyle:check`: 0 violations

## Notes for reviewers

- The base DDL sets `db_version=109` (from the preceding VersionSortKeyUpgrader change), so fresh installs run upgrade 110 automatically, which widens the column.
- Confirmed no other outbox event type builds an aggregate id longer than 1282 chars. The widest form is `groupId-artifactId-version` used by ArtifactVersionCreated, ArtifactVersionDeleted, ArtifactVersionMetadataUpdated, and ArtifactVersionStateChanged.
- The version bump before ALTER ordering follows the existing project convention (upgrades 101, 106, 107, 108, 109 all bump first).
- The MSSQL VARCHAR to NVARCHAR correction is intentional: every other MSSQL identifier column uses NVARCHAR, and VARCHAR would corrupt non ASCII group/artifact IDs.


app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/109/mssql.upgrade.ddl
```diff
@@ -0,0 +1,8 @@
+-- *********************************************************************
+-- DDL for the Apicurio Registry - Database: mssql
+-- Upgrade Script from 108 to 109
+-- *********************************************************************
+
+UPDATE apicurio SET propValue = 109 WHERE propName = 'db_version';
+
+ALTER TABLE outbox ALTER COLUMN aggregateid NVARCHAR(2048) NOT NULL;
```
paoloantinori: This MSSQL upgrade is a type change (VARCHAR(255) → NVARCHAR(2048)), not just a width widening, so it is a higher-risk ALTER than the MySQL/Postgres width-only ones. The syntax is valid T-SQL and varchar→nvarchar is a lossless conversion, so the code looks correct — but as the PR notes, neither the MySQL nor MSSQL path was verified against a real 108 database (Docker unavailable). Confirming both upgrades succeed against an actual pre-existing 108 DB before merge would close the one open gap; Postgres is already covered by `PostgresqlStorageTest#testCreateArtifactWithMaxLengthIdentifiers`.

paoloantinori (commented):
This is a correct, well-scoped fix for the outbox `aggregateid` overflow (#8793): the base DDLs and per-dialect upgrade scripts use each database's proper ALTER syntax, the MSSQL VARCHAR→NVARCHAR correction is well-justified, and the tests are thorough and self-validating. I found no defects in the diff; the only open item is confirming the MySQL and MSSQL upgrade paths against a real 108 database, which the PR already flags as unverified.

- **Base DDLs** widen the column to 2048 (`postgresql.ddl:120`, `mysql.ddl:243`, `mssql.ddl:116`), comfortably above the maximum generated id of 1282 (512+1+512+1+256) and leaving headroom.
- **Upgrade scripts** (`upgrades/109/`) are dialect-correct: Postgres `ALTER COLUMN ... TYPE VARCHAR(2048)` preserves the existing NOT NULL; MySQL `MODIFY COLUMN ... NOT NULL` re-specifies it as required; MSSQL `ALTER COLUMN ... NVARCHAR(2048) NOT NULL` is valid T-SQL and re-specifies nullability.
- **MSSQL VARCHAR→NVARCHAR** (`mssql.ddl:116`, `upgrades/109/mssql.upgrade.ddl:8`) is the right call — the source identifier columns (groups/artifacts/versions) are NVARCHAR, so the concatenated `aggregateid` must be too, otherwise non-ASCII group/artifact IDs would be corrupted on the way into the outbox.
- **Tests are strong**: `OutboxAggregateIdLengthTest` cross-checks all dialects, asserts base-DDL and upgrade widths agree, and ties the schema width to the actual output of `ArtifactVersionCreated.of()`; `PostgresqlStorageTest#testCreateArtifactWithMaxLengthIdentifiers` proves the artifact/version transaction now commits end-to-end on real Postgres.
- **Caveat**: the MySQL and MSSQL upgrades (`upgrades/109/mysql.upgrade.ddl:8`, `upgrades/109/mssql.upgrade.ddl:8`) are unverified against a real 108 DB (Docker unavailable) — worth confirming pre-merge, especially the MSSQL type conversion.

7vignesh:
Update: verified the MSSQL upgrade path locally with Docker. The test now passes against real SQL Server (testcontainer) confirming the NVARCHAR(2048) ALTER works correctly. All 17 tests green across Postgres, MSSQL, and the DDL consistency suite.

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
```diff
@@ -205,6 +205,84 @@ public void testCreateArtifactWithLargeMetaData() throws Exception {
...
     }
 
+    /**
+     * Regression test for the outbox aggregate id length. Creating an artifact fires an outbox event whose
+     * aggregate id is "groupId-artifactId" (and "groupId-artifactId-version" for the version event). That
+     * INSERT happens in the same transaction as the artifact write, so if the aggregateid column is too
+     * narrow the whole creation is rolled back and the artifact never appears.
+     *
+     * <p>
+     * Uses the maximum widths the schema permits: groupId 512, artifactId 512, version 256. On Postgres and
```
paoloantinori: This docstring (and the PR summary) say the test exercises the maximum schema widths — groupId 512, artifactId 512, version 256 — but the code below uses `generateString(130)` for all three identifiers (lines 226-228), producing a 393-char aggregate id rather than 1282. The test still reproduces the original bug (393 > 255), but it does not cover the true maximum the column was widened to hold. The inline comment at line 221 gives the actual reason — MSSQL's `branch_versions` clustered-index 900-byte limit prevents 512/512/256 — so consider correcting the docstring/summary to match the 130-char reality, and optionally adding a Postgres-only variant that genuinely uses 512/512/256 (Postgres has no such index limit) to cover the 1282-char case end-to-end.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/109/mysql.upgrade.ddl
```diff
@@ -0,0 +1,8 @@
+-- *********************************************************************
+-- DDL for the Apicurio Registry - Database: MySQL
+-- Upgrade Script from 108 to 109
+-- *********************************************************************
+
+UPDATE apicurio SET propValue = 109 WHERE propName = 'db_version';
```
paoloantinori, line 6: The `db_version` bump precedes the column `ALTER` (`mssql.upgrade.ddl` uses the same order). MySQL and MSSQL DDL is non-transactional and implicitly commits, so if the ALTER fails the version is already 109 and the upgrade won't be retried on the next startup — leaving `aggregateid` at 255 after an apparently successful upgrade. Worth confirming this ordering matches the project's existing upgrade-script convention and that the upgrade runner would still recover; if not, moving the ALTER before the version bump keeps db_version at 108 on failure so the upgrade retries cleanly.

paoloantinori (commented):
The core fix is sound and well-justified: widening `outbox.aggregateid` to VARCHAR(2048) (NVARCHAR on MSSQL) in the base DDLs with matching upgrade scripts resolves the transaction-rollback failure, and the MSSQL VARCHAR→NVARCHAR correction is a good catch for non-ASCII identifiers. The static `OutboxAggregateIdLengthTest` plus the real-Postgres regression test cover the change well, and base-DDL↔upgrade widths are consistent across dialects. Two observations worth a look before merge:

- `AbstractRegistryStorageTest.java:215` — the test's docstring and the PR summary claim max-length identifiers (groupId 512 / artifactId 512 / version 256), but the code uses `generateString(130)` for all three (lines 226-228), so the exercised aggregate id is 393, not 1282. It still reproduces the original bug (393 > 255) but does not cover the true maximum; the actual reason (MSSQL's `branch_versions` 900-byte clustered-index limit, noted in the comment at line 221) should replace the 512/512/256 claim.
- `upgrades/109/mysql.upgrade.ddl:6` (same order in `mssql.upgrade.ddl:6`) — the `db_version` bump runs before the column `ALTER`. On MySQL/MSSQL DDL is non-transactional and auto-commits, so a failed ALTER would leave db_version at 109 with the column still at 255, and the upgrade would not retry on next startup. Worth confirming this matches the project's existing upgrade convention and that the runner recovers; if not, moving the ALTER before the version bump keeps db_version at 108 on failure so it retries.

7vignesh:
Thanks for the detailed review. Addressed both points:

1. Fixed the docstring to accurately describe why 130-char identifiers are used instead of the schema max. MSSQL's branch_versions table has a composite clustered PK on (groupId, artifactId, branchId, version), all NVARCHAR, and the 900-byte index limit makes the full 512/512/256 impossible on that backend. The aggregate id of 393 still exceeds the old 255-char limit that caused the bug.

2. Regarding the version bump ordering: the existing convention in this project (upgrades 101, 106, 107) is to bump db_version first, then apply DDL changes. My scripts follow that same pattern for consistency. If the project wants to change this convention for better retry semantics on non-transactional DDL engines, that would be a broader change across all upgrade scripts.

7vignesh:
@jsenko pls review

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/110/mysql.upgrade.ddl
```diff
@@ -0,0 +1,8 @@
+-- *********************************************************************
+-- DDL for the Apicurio Registry - Database: MySQL
+-- Upgrade Script from 109 to 110
+-- *********************************************************************
+
+UPDATE apicurio SET propValue = 110 WHERE propName = 'db_version';
+
+ALTER TABLE outbox MODIFY COLUMN aggregateid VARCHAR(2048) NOT NULL;
```
paoloantinori, line 8: MySQL DDL implicitly commits and is not transactional, so the version bump on line 6 and this ALTER commit independently. If this ALTER fails after the bump commits, the database is stamped at db_version=110 while aggregateid stays at 255, and the next startup sees 110 == target and skips this upgrade entirely, so the column is never widened and the original INSERT failure resurfaces later. Reversing the order (ALTER before the version bump) leaves db_version at the prior value on failure so the upgrade retries on the next startup. Confirm this matches how the upgrade framework executes scripts and recovers from a mid-script failure.

app/src/test/java/io/apicurio/registry/storage/impl/sql/OutboxAggregateIdLengthTest.java
```diff
@@ -0,0 +1,228 @@
...
+        int groupIdWidth = extractWidth(GROUPS_GROUPID, ddl, "groups.groupId", dialect);
+        int artifactIdWidth = extractWidth(ARTIFACTS_ARTIFACTID, ddl, "artifacts.artifactId", dialect);
+        int versionWidth = extractWidth(VERSIONS_VERSION, ddl, "versions.version", dialect);
+
+        // Sanity-check the identifier widths this assertion depends on, so that widening one of them later
+        // without widening aggregateid is caught here rather than in production.
+        assertEquals(512, groupIdWidth, "Unexpected groups.groupId width in " + dialect + ".ddl");
+        assertEquals(512, artifactIdWidth, "Unexpected artifacts.artifactId width in " + dialect + ".ddl");
+        assertEquals(256, versionWidth, "Unexpected versions.version width in " + dialect + ".ddl");
+
```
paoloantinori, line 88: This asserts 1282 as the widest aggregate id, but only ArtifactVersionCreated is exercised (in testGeneratedAggregateIdFitsColumn). The widening is sufficient only if no other outbox event type builds a longer id. Worth enumerating the other outbox events (ArtifactCreated, ContentCreated, comment and role events, and so on) and confirming none concatenates more or longer fields, otherwise VARCHAR(2048) can still overflow on that path and the bug persists there.

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
```diff
@@ -205,6 +205,86 @@ public void testCreateArtifactWithLargeMetaData() throws Exception {
...
+     * it just confirms that long identifiers round-trip.
+     */
+    @Test
+    public void testCreateArtifactWithMaxLengthIdentifiers() throws Exception {
+        // Use 130 chars for groupId/artifactId/version because MSSQL clustered indexes have a 900-byte
+        // key limit. The branch_versions table has a composite PK on (groupId, artifactId, branchId,
+        // version), all NVARCHAR (2 bytes/char). With branchId ~6 chars ("latest"), the sum must stay
+        // under 450 chars. An aggregate id of 130 + 1 + 130 + 1 + 130 = 393 still well exceeds the old
+        // 255-char column limit that caused the original bug.
+        String longGroupId = generateString(130);
```
paoloantinori, line 230: The 130-char choice is forced by MSSQL's branch_versions clustered-key limit, but PostgreSQL has no such limit, so this shared test never drives the real upper bound (groupId and artifactId at 512 plus version at 256, a 1282-char aggregate id) through an actual outbox INSERT, it tops out at 393. A PostgreSQL-only override using the full 512/512/256 would prove the true maximum round-trips end to end and complement the width checks in OutboxAggregateIdLengthTest.

paoloantinori (commented):
The core fix is correct and the test suite is unusually thorough for a schema change: OutboxAggregateIdLengthTest guards the column width, cross-dialect agreement, MSSQL NVARCHAR usage, and upgrade-versus-base-DDL parity. I could not verify two assumptions against source because the registry tree is not checked out here, so the points below are confirmations to close out rather than confirmed defects.

- app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/110/mysql.upgrade.ddl:6,8 - the db_version bump commits before the ALTER, and MySQL DDL is non-transactional. A failed ALTER after the bump leaves the DB stamped at 110 with the column still 255, and startup then skips upgrade 110. Consider ALTER-then-bump ordering.
- app/src/test/java/io/apicurio/registry/storage/impl/sql/OutboxAggregateIdLengthTest.java:87 - the 1282-char maximum is asserted only for ArtifactVersionCreated. Confirm no other outbox event type builds a longer aggregate id.
- app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java:229 - the end-to-end test caps identifiers at 130 chars (393-char aggregate id) for the MSSQL clustered-key limit, so the true maximum (512/512/256, 1282 chars) is never driven through a real outbox INSERT. A Postgres-only variant would close the gap.
- The summary says db-version 108 to 109 with an upgrades/109/ directory, but the diff is 109 to 110 with upgrades/110/. The code is self-consistent at 110; the description should be updated to match.
- Endorsing the author's own note: the MySQL and MSSQL upgrade paths still need a manual run against a real 108/109 database, since they could not be exercised locally.

7vignesh:
Thanks for the follow up review. All points addressed:

1. PR description updated to correctly reflect 109 to 110 with upgrades/110/. Apologies for the stale text.

2. ALTER vs bump ordering: confirmed this follows the existing convention. Upgrades 101, 106, 107, 108, and 109 all bump db_version before the DDL change. Changing this would be a broader project decision.

3. Confirmed no other outbox event type builds an aggregate id longer than 1282 chars. All event classes were audited: the widest form is groupId + artifactId + version (used by ArtifactVersionCreated, ArtifactVersionDeleted, ArtifactVersionMetadataUpdated, ArtifactVersionStateChanged). All other events use shorter forms (groupId + artifactId, bare groupId, or a rule type enum value).

4. Added PostgresqlStorageTest#testCreateArtifactWithTrueMaxLengthIdentifiers which drives the actual maximum (512/512/256 = 1282 char aggregate id) through a real Postgres outbox INSERT. This closes the gap between the MSSQL constrained test (130 chars) and the true maximum. Verified passing.

paoloantinori:
Thanks for confirming both points. The description now matches the upgrade path, and the bump-before-ALTER ordering is consistent with the existing migrations. No further concerns from me.

paoloantinori (commented):
The root cause analysis is right and the fix direction is safe: the widest aggregate id is 512 + 1 + 512 + 1 + 256 = 1282 chars, VARCHAR(2048) leaves headroom, the outbox PK stays on `id` so MSSQL's 900-byte clustered key limit is untouched, and the VARCHAR to NVARCHAR correction on MSSQL matches every sibling identifier column. No blockers from my side; what remains is one upgrade-safety concern, a missing license header, and two test-strength points. This was a static review of the provided diff, I could not run the test suite or read the surrounding repo, so the reported 15/15 and embedded-Postgres results are taken from the PR description.
- Upgrade scripts set db_version=110 before the ALTER (`upgrades/110/mysql.upgrade.ddl` line 6, same order in the mssql and postgresql scripts). A failed ALTER then leaves the database stamped at 110 with the column still at 255, and the widening is never retried.
- `OutboxAggregateIdLengthTest.java` line 1 starts at `package` with no Apache 2.0 license header; the new DDL files follow the existing banner convention, but the Java file does not follow the Java convention.
- `AbstractRegistryStorageTest.java` line 224 never asserts the outbox row itself, so on H2 (no outbox table) and MySQL (outbox declared but not written to yet) the regression test passes without exercising the outbox INSERT that caused the bug.
- `PostgresqlStorageTest.java` line 33 is a near line-for-line copy of the abstract 130-char test with bigger numbers.
- The regex-based DDL tests in `OutboxAggregateIdLengthTest.java` are coupled to formatting details (mysql.ddl multi-line layout, backticks around `groups`), so a pure reformat of a DDL file can break them; the cross-dialect consistency check they provide is worth the brittleness, just be aware.
- Per the notes, fresh installs still replay upgrade 110 because the base DDL stamps db_version=109. That is harmless here (empty outbox table, idempotent ALTER), but if the convention from the 109 change was to move the DDL stamp along with `db-version`, worth confirming which is intended.
- Keeping `id`, `aggregatetype`, and `type` as VARCHAR on MSSQL is fine since those values are ASCII (UUID and enum-like strings); only aggregateid carries user-supplied identifiers.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/110/mysql.upgrade.ddl
```diff
@@ -0,0 +1,8 @@
+-- *********************************************************************
+-- DDL for the Apicurio Registry - Database: MySQL
+-- Upgrade Script from 109 to 110
+-- *********************************************************************
+
+UPDATE apicurio SET propValue = 110 WHERE propName = 'db_version';
```
paoloantinori, line 6: This script bumps db_version to 110 before the ALTER, and the mssql and postgresql scripts use the same order. If the ALTER fails mid-upgrade (for example a lock timeout on a busy outbox table), the database ends up reporting version 110 while `aggregateid` is still VARCHAR(255), and on the next startup upgrade 110 is skipped, so the widening is never retried and the original bug silently returns. On MySQL this is concrete: the ALTER implicit-commits this UPDATE, so wrapping the script in a transaction does not help. Consider running the ALTER first and the version bump last so a partial failure is retried on the next boot. I see upgrades 101 and 106 through 109 bump first as well, but their follow-up statements are additive; this is the first upgrade where partial application permanently skips the fix.

app/src/test/java/io/apicurio/registry/storage/impl/sql/OutboxAggregateIdLengthTest.java
```diff
@@ -0,0 +1,228 @@
+package io.apicurio.registry.storage.impl.sql;
```
paoloantinori, line 1: The file starts at `package` with no Apache 2.0 license header. Every other Java file in this module carries one, and the new upgrade DDLs follow their own banner convention, so this stands out. Please add the standard header (if app/ tests are exempt from the header check, feel free to ignore).

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
```diff
@@ -206,6 +206,86 @@ public void testCreateArtifactWithLargeMetaData() throws Exception {
...
+     * <p>
+     * Uses 130-char identifiers rather than the schema maximum (512/512/256) because MSSQL's
+     * branch_versions table has a composite clustered PK on (groupId, artifactId, branchId, version),
+     * all NVARCHAR (2 bytes/char), and MSSQL limits clustered index keys to 900 bytes. The resulting
+     * aggregate id of 393 chars still well exceeds the old 255-char column limit that caused the bug.
+     * On Postgres and MSSQL this exercises the real outbox INSERT; on H2 there is no outbox table, so
+     * it just confirms that long identifiers round-trip.
+     */
+    @Test
+    public void testCreateArtifactWithMaxLengthIdentifiers() throws Exception {
```
paoloantinori, line 224: The assertions verify metadata round-trip but never the outbox row itself. On H2 (no outbox table) and MySQL (outbox declared but not yet written to, per the comment in OutboxAggregateIdLengthTest) this test passes without exercising the outbox INSERT that caused the bug, and even on Postgres and MSSQL the insert is only implied by the lack of a rollback. Consider querying the outbox table directly after createArtifact and createArtifactVersion (a count, or a select on the computed `groupId-artifactId-version` aggregate id) on the profiles that write it, so the test fails for the actual regression rather than passing vacuously on two of the four dialects.

app/src/test/java/io/apicurio/registry/storage/impl/sql/PostgresqlStorageTest.java
```diff
@@ -1,15 +1,72 @@
...
 @Typed(PostgresqlStorageTest.class)
 public class PostgresqlStorageTest extends DefaultRegistryStorageTest {
+
+    /**
+     * Postgres-only test that exercises the outbox INSERT with the true maximum aggregate id length
+     * (512 + 1 + 512 + 1 + 256 = 1282 chars). This cannot run on MSSQL due to its 900-byte clustered
+     * index key limit on composite PKs involving these identifier columns.
+     */
+    @Test
+    public void testCreateArtifactWithTrueMaxLengthIdentifiers() throws Exception {
```
paoloantinori, line 33: This is a near line-for-line copy of testCreateArtifactWithMaxLengthIdentifiers in AbstractRegistryStorageTest, differing only in identifier lengths (512/512/256 vs 130/130/130). Consider parameterizing the abstract test instead, for example an overridable `protected int identifierLength()` or a @MethodSource, with the MSSQL profile capping at 130 for the 900-byte clustered key limit. That keeps one copy of the assertions, so future additions (such as the direct outbox row check suggested on the abstract test) land in one place instead of two.

7vignesh:
@paoloantinori  Thanks for the thorough review. Addressing each point:

1. Upgrade ordering follows the existing convention. Upgrades 101, 106, 107, 108, and 109 all bump db_version before the DDL change. Changing this would be a broader project decision across all upgrade scripts.

2. License header is intentionally omitted. The project does not use per file Apache headers (only 143 of 2168 Java files carry one) and the repo root LICENSE governs.

3. The H2/MySQL limitation is documented in the test docstring. The real outbox INSERT path is covered by PostgresqlStorageTest#testCreateArtifactWithTrueMaxLengthIdentifiers which drives the full 1282 char aggregate id through a real Postgres instance.

4. The Postgres only test intentionally overlaps with the abstract test because it exercises the true maximum (512/512/256) that MSSQL cannot handle due to its 900 byte clustered index key limit. The abstract test uses 130 char identifiers to stay within MSSQL constraints while still exceeding the old 255 char column.

5. Acknowledged on the regex brittleness. The cross dialect consistency guarantee is worth it.

6. Base DDL stamping 109 while db_version is 110 is the same pattern used by the preceding upgrades (108 stamped 107, 109 stamped 108). Fresh installs replay the upgrade which is harmless since the table is empty and the ALTER is idempotent.

carlesarnal:
Most comments from Paolo don't have a clear answer as a reply. Concerns raised as comments must be addressed directly as a reply to said comments.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/110/mysql.upgrade.ddl
7vignesh: Follows the existing project convention. Upgrades 101, 106, 107, 108, and 109 all bump db_version before the DDL. Changing this ordering would require a broader decision across all upgrade scripts.

app/src/test/java/io/apicurio/registry/storage/impl/sql/OutboxAggregateIdLengthTest.java
7vignesh: The project does not use per file Apache license headers. Only 143 of 2168 Java files carry one and AGENTS.md explicitly states not to add them. The repo root LICENSE governs.

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
7vignesh: Correct. The outbox INSERT is exercised by the Postgres path (SqlEventRepository.supportsDatabaseEvents returns true for postgresql and mssql). H2 has no outbox table and MySQL does not write to it. The real INSERT coverage is in PostgresqlStorageTest#testCreateArtifactWithTrueMaxLengthIdentifiers which drives 1282 chars through the actual outbox on Postgres.

app/src/test/java/io/apicurio/registry/storage/impl/sql/PostgresqlStorageTest.java
7vignesh: Kept separate intentionally. The abstract test (130 chars) proves cross-backend compatibility including MSSQL. The Postgres test (512/512/256) proves the true maximum works on the one backend where it can actually be stored. Parameterizing would obscure which backend constraint each test size serves.

app/src/test/java/io/apicurio/registry/storage/impl/sql/OutboxAggregateIdLengthTest.java
7vignesh: Confirmed. Audited all 16 outbox event classes. The widest aggregate id form is groupId + "-" + artifactId + "-" + version (1282 chars), used by ArtifactVersionCreated, ArtifactVersionDeleted, ArtifactVersionMetadataUpdated, and ArtifactVersionStateChanged. All other events use shorter forms: groupId + artifactId (1025), bare groupId (512), or a rule type enum value.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/109/mssql.upgrade.ddl
7vignesh: This is now at upgrades/110/. The MSSQL type change (VARCHAR to NVARCHAR) is valid T-SQL and matches every other identifier column in the MSSQL DDL. Tested against a real SQL Server testcontainer.

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
7vignesh: Fixed in the current version. The docstring now accurately describes the 130 char choice and MSSQL constraint.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/109/mysql.upgrade.ddl
7vignesh: Same convention answer as the 110 version. Follows existing project pattern.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/110/mysql.upgrade.ddl
7vignesh: Same as above. Follows existing convention across all upgrade scripts.

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
7vignesh: Addressed. PostgresqlStorageTest#testCreateArtifactWithTrueMaxLengthIdentifiers now exercises the full 512/512/256 (1282 char) aggregate id through a real Postgres outbox INSERT.

paoloantinori:
I set the milestone. This PR and #8793 are both on `3.x`, which clears half of the Validate PR failure without you pushing anything. Setting it needs triage permission, so that part was never yours to fix.

I chose `3.x` rather than `3.4.0` on purpose. 3.4.0 is due on 17 September and this has no approving review yet, so putting it there would list the change in release notes it will not be in. `3.x` is the unscheduled milestone and can be re-pointed once there is a merge date.

The other half of that check is the DCO trailer on `784892b6`, and it is a merge commit rather than one of yours. Your actual change, `4144a76d1`, is signed correctly.

A rebase onto main fixes it:

```
git fetch upstream
git rebase upstream/main
```

The merge commit disappears in a rebase, so the DCO check goes green. GitHub also reports this branch as behind main, and the same command clears that.

There is a third reason to rebase, and it is the one I would raise even if the rest were green. This PR adds `upgrades/110` for all four dialects and moves `db-version` to 110. Three other open PRs do exactly the same: #8538, #9651, and #9789. Main is still on 109, so 110 is unclaimed today, but only one of the four can keep it. Rebasing is how you find out which number is actually free when your turn comes.

I have written the rule and the renumbering steps up in #10141 so it lives in one place rather than four PR threads. The short version is that whichever PR merges first keeps 110 and the rest renumber, and that nobody should resolve the conflict by folding two migrations into a single `110` directory. That variant fails silently on every deployment that already upgraded, because `databaseUpgrade` never re-runs a version it has already applied. #10141 has the detail.

Nothing here needs a decision from you beyond the rebase. I will post the next free number in #10141 as each of these lands.

On the inactivity label: the bot warned on 14 September, so a comment or a push keeps it open. The rebase counts.

PR #9153, opened 2026-08-01
fix(auth): require Admin or current owner for ownership transfers

### LINKED ISSUE #8596: [BUG] V3 API updateArtifactMetaData allows any Write-level user to change artifact ownership without Admin/Owner authorization

batmnnn:
## Summary
- Restrict changing the `owner` field on artifact and group metadata updates to the current owner or an admin
- Keep other metadata fields (name, description, labels) writable for users with Write access
- Add `OwnershipTransferAuthorizer` plus unit and basic-auth integration tests for positive and negative (403) cases

Fixes #8596

## Test plan
- [x] `OwnershipTransferAuthorizerTest` (decision matrix)
- [x] `BasicAuthWithPropertiesTest#testUpdateGroupOwner*` (owner / non-owner 403 / admin)
- [x] Regression: existing artifact-owner auth tests and `GroupMetaDataTest` (auth disabled)
- [x] `./mvnw checkstyle:check -pl app` on touched files
- [ ] CI auth/security suite on this PR

### DIFF
--- a/app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
+++ b/app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2026 Red Hat
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.apicurio.registry.auth;
+
+import io.quarkus.security.ForbiddenException;
+import io.quarkus.security.identity.SecurityIdentity;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+
+import java.security.Principal;
+import java.util.Objects;
+
+/**
+ * Authorizes ownership transfers for artifacts and groups.
+ *
+ * <p>Metadata update endpoints use {@link AuthorizedLevel#Write} so non-owners can still
+ * change name/description/labels. Ownership changes require Admin or the current owner
+ * (mirrors V2 {@code updateArtifactOwner} / {@link AuthorizedLevel#AdminOrOwner}).
+ *
+ * <p>Cannot be expressed solely via {@code @Authorized(level = AdminOrOwner)} because that
+ * would block non-owners from updating non-owner metadata fields on the same endpoint.
+ */
+@ApplicationScoped
+public class OwnershipTransferAuthorizer {
+
+    private static final String TRANSFER_FORBIDDEN =
+            "Only the current owner or an admin can transfer ownership.";
+
+    @Inject
+    AuthConfig authConfig;
+
+    @Inject
+    AdminOverride adminOverride;
+
+    @Inject
+    RoleBasedAccessController rbac;
+
+    @Inject
+    SecurityIdentity securityIdentity;
+
+    /**
+     * Ensures the caller may set ownership to {@code requestedOwner}.
+     *
+     * <ul>
+     *   <li>No-op when authentication is disabled</li>
+     *   <li>No-op when {@code requestedOwner} equals the current owner</li>
+     *   <li>Allowed for admins (admin-override or RBAC admin role)</li>
+     *   <li>Allowed when the resource is unowned ({@code currentOwner == null})</li>
+     *   <li>Allowed when the caller is the current owner</li>
+     *   <li>Otherwise forbidden</li>
+     * </ul>
+     *
+     * @param currentOwner the owner currently stored for the resource (may be null)
+     * @param requestedOwner the non-empty owner value from the request
+     */
+    public void authorizeOwnerChange(String currentOwner, String requestedOwner) {
+        if (!isAuthenticationEnabled()) {
+            return;
+        }
+
+        if (Objects.equals(currentOwner, requestedOwner)) {
+            return;
+        }
+
+        if (isAdmin()) {
+            return;
+        }
+
+        Principal principal = securityIdentity.getPrincipal();
+        String currentUser = principal != null ? principal.getName() : null;
+
+        // Unowned resources may be claimed by any authenticated Write user.
+        if (currentOwner == null || currentOwner.equals(currentUser)) {
+            return;
+        }
+
+        throw new ForbiddenException(TRANSFER_FORBIDDEN);
+    }
+
+    private boolean isAdmin() {
+        return adminOverride.isAdmin() || (authConfig.isRbacEnabled() && rbac.isAdmin());
+    }
+
+    private boolean isAuthenticationEnabled() {
+        return authConfig.isOidcAuthEnabled() || authConfig.isBasicAuthEnabled()
+                || authConfig.isProxyHeaderAuthEnabled() || authConfig.isKubernetesAuthEnabled();
+    }
+}
--- a/app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
+++ b/app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
@@ -27,6 +27,12 @@ import java.util.Objects;
 /**
  * Authorizes ownership transfers for artifacts and groups.
  *
+ * <p>Wired into the only V3 client-writable owner paths:
+ * {@code GroupsResourceImpl#updateArtifactMetaData} and
+ * {@code GroupsResourceImpl#updateGroupById}. Version metadata has no owner field;
+ * import/bulk paths are Admin-only. V2 uses a dedicated
+ * {@code updateArtifactOwner} with {@link AuthorizedLevel#AdminOrOwner}.
+ *
  * <p>Metadata update endpoints use {@link AuthorizedLevel#Write} so non-owners can still
  * change name/description/labels. Ownership changes require Admin or the current owner
  * (mirrors V2 {@code updateArtifactOwner} / {@link AuthorizedLevel#AdminOrOwner}).
@@ -68,7 +74,7 @@ public class OwnershipTransferAuthorizer {
      * @param requestedOwner the non-empty owner value from the request
      */
     public void authorizeOwnerChange(String currentOwner, String requestedOwner) {
-        if (!isAuthenticationEnabled()) {
+        if (!authConfig.isAuthenticationEnabled()) {
             return;
         }
 
@@ -93,10 +99,5 @@ public class OwnershipTransferAuthorizer {
 
     private boolean isAdmin() {
         return adminOverride.isAdmin() || (authConfig.isRbacEnabled() && rbac.isAdmin());
-    }
-
-    private boolean isAuthenticationEnabled() {
-        return authConfig.isOidcAuthEnabled() || authConfig.isBasicAuthEnabled()
-                || authConfig.isProxyHeaderAuthEnabled() || authConfig.isKubernetesAuthEnabled();
     }
 }
--- a/app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
+++ b/app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
@@ -29,9 +29,14 @@ import java.util.Objects;
  *
  * <p>Wired into the only V3 client-writable owner paths:
  * {@code GroupsResourceImpl#updateArtifactMetaData} and
- * {@code GroupsResourceImpl#updateGroupById}. Version metadata has no owner field;
- * import/bulk paths are Admin-only. V2 uses a dedicated
- * {@code updateArtifactOwner} with {@link AuthorizedLevel#AdminOrOwner}.
+ * {@code GroupsResourceImpl#updateGroupById}. Confirmed by grepping
+ * {@code rest/v3/impl} for {@code data.getOwner()} / {@code setOwner} and checking
+ * OpenAPI {@code Editable*} schemas in {@code common/.../openapi.json}: only
+ * {@code EditableArtifactMetaData} and {@code EditableGroupMetaData} expose
+ * {@code owner}; {@code EditableVersionMetaData} / {@code CreateArtifact} /
+ * {@code CreateGroup} do not (create stamps owner from the principal); import is
+ * Admin-only. V2 uses dedicated {@code updateArtifactOwner} with
+ * {@link AuthorizedLevel#AdminOrOwner}.
  *
  * <p>Metadata update endpoints use {@link AuthorizedLevel#Write} so non-owners can still
  * change name/description/labels. Ownership changes require Admin or the current owner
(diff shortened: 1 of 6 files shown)


app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
```diff
@@ -0,0 +1,102 @@
...
+import io.quarkus.security.identity.SecurityIdentity;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+
+import java.security.Principal;
+import java.util.Objects;
+
+/**
+ * Authorizes ownership transfers for artifacts and groups.
+ *
```
paoloantinori, line 29: The class doc says this authorizes ownership transfers for artifacts and groups, but the call is wired into only `updateArtifactMetaData` and `updateGroupById`. Please grep the v3 resource impls for `setOwner`/`getOwner` and confirm no other endpoint — notably artifact *version*-metadata updates and any import/bulk paths — writes the `owner` field. Each unguarded site fully bypasses this control, which would defeat the purpose of the PR.

app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
```diff
@@ -0,0 +1,102 @@
...
+        throw new ForbiddenException(TRANSFER_FORBIDDEN);
+    }
+
+    private boolean isAdmin() {
+        return adminOverride.isAdmin() || (authConfig.isRbacEnabled() && rbac.isAdmin());
+    }
+
+    private boolean isAuthenticationEnabled() {
+        return authConfig.isOidcAuthEnabled() || authConfig.isBasicAuthEnabled()
+                || authConfig.isProxyHeaderAuthEnabled() || authConfig.isKubernetesAuthEnabled();
```
paoloantinori: This hand-rolled auth-enabled check is load-bearing: if a new auth backend is later added to `AuthConfig` without updating this list, `authorizeOwnerChange` silently no-ops and every ownership transfer is allowed — the worst failure mode for an authz primitive. If there is a canonical `isAuthEnabled`/equivalent helper already used by the `@Authorized` interceptor, reuse it here; otherwise add a test asserting this list tracks `AuthConfig`'s auth backends.

paoloantinori (commented):
The core authorization logic is sound: the owner/current-owner/admin/unowned decision matrix in `OwnershipTransferAuthorizer` is correct, null-principal falls through to deny, and the same-owner short-circuit avoids false 403s when clients round-trip the existing owner value. The main diligence item for a hardening PR is confirming *every* owner-writing endpoint is covered; a few smaller observations follow.

- **Coverage of owner-writing endpoints:** the check is wired into `updateArtifactMetaData` (`GroupsResourceImpl.java`, ~line 598) and `updateGroupById` (~line 649) only. Please grep the v3 resource impls for `setOwner`/`getOwner` and confirm version-metadata updates and any import/bulk paths cannot set `owner` — each unguarded site fully bypasses this control.
- **`isAuthenticationEnabled()` is reimplemented** (`OwnershipTransferAuthorizer.java:99-101`): if it drifts from the canonical auth-enabled check (e.g., a new auth backend lands in `AuthConfig`), this authorizer silently no-ops — the worst failure mode for an authz primitive. Reuse the existing helper if one exists, or add a test that fails when the two diverge.
- **Non-atomic vs storage:** the current owner is read (`getArtifactMetaData`/`getGroupMetaData`) then written in a separate call, so concurrent transfers can interleave. This matches the existing resource-layer model and improves on the prior no-check baseline, so it isn't a regression — but storage-side enforcement would close the window.
- **Unowned resources are claimable by any authenticated Write user** (`:88`): documented and intentional; flagging only so it stays a deliberate policy choice.
- **Minor test gap:** `OwnershipTransferAuthorizerTest` never exercises the RBAC-admin branch (`rbac.isAdmin()` is always `false`), so the `(isRbacEnabled && rbac.isAdmin())` path in `isAdmin()` is uncovered.

batmnnn:
Thanks for the review — addressed the actionable items:

**Owner-writing endpoint coverage**
Grep of the V3 resource impls confirms the only client-writable `owner` paths are `updateArtifactMetaData` and `updateGroupById` (both now guarded). `EditableVersionMetaData` has no owner field. Import/bulk is Admin-only. V2 already uses dedicated `updateArtifactOwner` with `AdminOrOwner`. Documented this on `OwnershipTransferAuthorizer`.

**Auth-enabled check drift**
Extracted the canonical check used by `AuthorizedInterceptor` into `AuthConfig.isAuthenticationEnabled()` and call that from both the interceptor and the ownership authorizer, so a new auth backend cannot silently bypass this control.

**RBAC admin test gap**
Added `rbacAdminCanTransfer` covering `(isRbacEnabled && rbac.isAdmin())` with admin-override disabled.

Agreed on the non-atomic read/write note and the intentional unowned-claim policy — no code change for those.

app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
batmnnn: Confirmed: only `updateArtifactMetaData` and `updateGroupById` accept a client-supplied owner in V3. Version metadata has no owner field; import is Admin-only. Documented on the class.

app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
batmnnn: Fixed by moving the check to `AuthConfig.isAuthenticationEnabled()` and reusing it from both `AuthorizedInterceptor` and `OwnershipTransferAuthorizer`.

app/src/test/java/io/apicurio/registry/auth/BasicAuthWithPropertiesTest.java
```diff
@@ -537,4 +540,79 @@ public void testDeleteArtifactIfAdmin() throws Exception {
...
+        createGroup.setGroupId(groupId);
+        createGroup.setDescription("owned by bob1");
+        GroupMetaData created = client_dev1.groups().post(createGroup);
+        assertEquals(DEVELOPER_USERNAME, created.getOwner());
+
+        // Non-owner may still update non-owner metadata (OBAC group limit is off by default).
+        EditableGroupMetaData descriptionUpdate = new EditableGroupMetaData();
+        descriptionUpdate.setDescription("updated by bob2");
+        client_dev2.groups().byGroupId(groupId).put(descriptionUpdate);
+        assertEquals("updated by bob2", client_dev1.groups().byGroupId(groupId).get().getDescription());
```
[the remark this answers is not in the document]
batmnnn: Noted. The authorizer decision matrix (including non-owner deny) is covered in `OwnershipTransferAuthorizerTest`. The BasicAuth IT profile has OBAC on, so it exercises the group-owner transfer path with owner-only auth enabled. An RBAC-on/OBAC-off end-to-end profile would be a useful follow-up but is not required to close the ownership-transfer hole this PR targets.

app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
```diff
@@ -0,0 +1,103 @@
...
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+
+import java.security.Principal;
+import java.util.Objects;
+
+/**
+ * Authorizes ownership transfers for artifacts and groups.
+ *
+ * <p>Wired into the only V3 client-writable owner paths:
```
paoloantinori, line 30: The entire guarantee rests on this claim — that `updateArtifactMetaData` and `updateGroupById` are the only V3 client-writable paths that accept an `owner`. It's worth confirming with a repo-wide grep across the V3 resources (artifact/group create, import, bulk, and version-metadata endpoints) that no other reachable endpoint exposes an owner setter; any such path would bypass this check entirely. If already verified, a one-line note pointing to how it was confirmed would help reviewers.

paoloantinori (commented):
The change correctly restricts ownership transfers on artifact and group metadata to the current owner or an admin while leaving other metadata fields writable at the Write level, and the new decision matrix is correct and well-tested. The `isAuthenticationEnabled()` consolidation is a clean, behavior-preserving refactor that also guards against a future auth backend silently bypassing authorization.

- `OwnershipTransferAuthorizer.java:73-95` — the decision logic is correct across the full matrix (auth-disabled no-op, same-owner no-op, admin via override or RBAC, current-owner allow, unowned-resource allow, else `ForbiddenException`); `OwnershipTransferAuthorizerTest` mirrors each branch.
- `GroupsResourceImpl.java` (`updateArtifactMetaData` ~589-606, `updateGroupById` ~641-656) — both endpoints now fetch current metadata, invoke the authorizer with the trimmed requested owner, and reject empty/whitespace owners; `updateGroupById` additionally gains an empty-owner guard it previously lacked.
- `BasicAuthWithPropertiesTest.java:543-617` — owner / non-owner-403 / admin coverage for the group path is solid; note there is no equivalent artifact-path integration test in this PR (the artifact wiring is exercised only via the authorizer unit test and pre-existing artifact-owner tests).
- `AuthConfig.java` / `AuthorizedInterceptor.java:77` — the `isAuthenticationEnabled()` helper is an exact equivalent of the previous four-field check; good centralization, and the `updateGroupById` description-only update in the test confirms storage treats a null owner as leave-unchanged (no partial-update bypass).
- Minor: there is an inherent check-then-act window between `storage.getArtifactMetaData`/`getGroupMetaData` and the subsequent write (`GroupsResourceImpl.java` ~593/650); low impact for this use case.

app/src/main/java/io/apicurio/registry/auth/OwnershipTransferAuthorizer.java
batmnnn: Re-verified and documented on the class.

**Confirmation method:**
1. Grep `app/.../rest/v3/impl` for `data.getOwner()` / `setOwner` — the only request-body owner writers are `updateArtifactMetaData` and `updateGroupById` (both guarded). Create paths stamp owner from `securityIdentity.getPrincipal()`, not the request.
2. OpenAPI `Editable*` schemas in `common/src/main/resources/META-INF/openapi.json`: only `EditableArtifactMetaData` and `EditableGroupMetaData` have an `owner` property; `EditableVersionMetaData`, `CreateArtifact`, and `CreateGroup` do not.
3. `AdminResourceImpl#importData` is `@Authorized(level = Admin)`.

Added a one-line note on `OwnershipTransferAuthorizer` pointing to that grep + OpenAPI check.

batmnnn:
Done — all review feedback addressed:

- Ownership transfer restricted to current owner / admin on both V3 metadata update paths
- `AuthConfig.isAuthenticationEnabled()` centralized and shared with `AuthorizedInterceptor`
- RBAC-admin unit coverage added
- V3 owner-write coverage re-verified (grep + OpenAPI) and documented on `OwnershipTransferAuthorizer`

Ready for another look when you have a moment.

batmnnn:
@paoloantinori can you take a look at this again, thanks.

paoloantinori:
Thanks @batmnnn — the latest head reads well. The centralized `isAuthenticationEnabled()`, the `OwnershipTransferAuthorizer` decision matrix, and the V3 wiring (both `updateArtifactMetaData` and `updateGroupById`) all look consistent with the earlier feedback. The unit test matrix plus the basic-auth integration tests (owner / non-owner 403 / admin) cover the right cases.

One thing worth confirming before this lands, since the PR's whole purpose is hardening ownership semantics. When the request omits the `owner` field entirely (`data.getOwner() == null`), the guard block is skipped and `dto.setOwner(... : null)` is passed to storage. If `updateArtifactMetaData` / `updateGroupMetaData` treat a `null` owner as "clear ownership" rather than "leave unchanged", then a non-owner with Write access could strip an artifact's owner to null via a routine description/labels update — after which the resource is unowned and claimable by any Write user. That would be a path around the transfer guard.

This depends on the storage semantics for a null owner value, which isn't visible in this diff: if null is a no-op there's nothing to do; if null clears, the guard should also cover the null path (only current-owner/admin may clear). Worth a quick check against the storage impl. If it's already a no-op, this PR is good to go from my side.

batmnnn:
@paoloantinori Confirmed null owner is leave-unchanged, not clear. No bypass.

**Resource layer:** guard runs only when `data.getOwner() != null`. Empty/whitespace is rejected before storage. Omitting `owner` passes `null` on the DTO and skips the authorizer (expected for partial updates).

**Storage (canonical SQL):** both `SqlArtifactRepository#updateArtifactMetaData` and `SqlGroupRepository#updateGroupMetaData` only execute the owner UPDATE when `getOwner() != null && !getOwner().trim().isEmpty()`. There is no `SET owner = NULL` path.

**Other variants:** KafkaSQL applies the same SQL repositories via the journal. GitOps/KubernetesOps reject metadata updates as read-only.

**Existing tests:** `GroupMetaDataTest` (description-only / labels-only leave owner intact) and `BasicAuthWithPropertiesTest#testUpdateGroupOwnerDeniedForNonOwner` (non-owner description update keeps original owner, then owner change → 403).

So a Write user cannot strip ownership via a routine metadata update. PR is good from this angle.

paoloantinori:
Thanks @batmnnn — the breakdown matches my read of the head. The guard placement is right: running only when `data.getOwner() != null`, with empty/whitespace rejected before storage, means null-on-DTO correctly leaves the owner unchanged rather than clearing it. The dual coverage at resource (`GroupsResourceImpl`) and storage (`SqlArtifactRepository` / `SqlGroupRepository` only updating on non-null non-blank owner) closes the bypass window.

Two minor notes, neither blocking:
- The decision matrix in `OwnershipTransferAuthorizer` handles the unowned-resource claim case (any authenticated Write user), which is consistent with V2 behavior. Worth keeping in mind if you later want unowned-claim to be admin-only — that's a policy call, not a bug here.
- Since the PR is behind `main`, a rebase before merge will keep the authz path intact; the `AuthorizedInterceptor` change is centralized through `isAuthenticationEnabled()` so conflicts there, if any, should be trivial to resolve.

LGTM. Tests cover owner/admin/non-owner/403 across both artifact and group paths.

batmnnn:
Thanks @paoloantinori, done with rebase

EricWittmann:
Thanks @batmnnn — excellent work. The dedicated OwnershipTransferAuthorizer, the centralized isAuthenticationEnabled() cleanup, and the thorough test coverage (unit + integration, positive + negative) all made this an easy merge. Appreciate the responsiveness to review feedback too. Merged!

batmnnn:
Thanks @paoloantinori @EricWittmann for the thoughtful reviews and for accepting this PR. I really enjoyed contributing.

Also i’m interested in applying for LFX Mentorship with this project, what’s the best approach to get started or express interest?

PR #9157, opened 2026-08-01
fix(mcp): validate required fields during detection

### LINKED ISSUE #9156: fix(mcp): reject malformed required fields during artifact type detection
## Description

`McpToolContentAccepter` identifies content as an MCP tool whenever the JSON object contains `name` and `inputSchema` keys. It does not verify the required field types, so malformed content such as the following is accepted during artifact type detection:

```json
{name: null, inputSchema: null}
```

The MCP tool validator correctly requires `name` to be a non-empty string and `inputSchema` to be an object, but artifact type detection can run independently of a validity rule. As a result, unrelated or malformed JSON can be classified as `MCP_TOOL`.

## Expected behavior

MCP tool content detection should require:

- `name` to be a non-empty string
- `inputSchema` to be a JSON object

## Proposed fix

Tighten `McpToolContentAccepter` to check the required field shapes and add regression tests for null, incorrectly typed, and empty values.

hcthakur2004:
## Summary

- require MCP tool names to be non-empty strings during artifact type detection
- require MCP tool input schemas to be JSON objects during detection
- add regression coverage for null, incorrectly typed, and empty required fields

## Why

Artifact type detection previously checked only whether the `name` and `inputSchema` keys existed. Malformed JSON such as `{name:null,inputSchema:null}` could therefore be classified as an MCP tool even when validity rules were not enabled.

## Testing

- `git diff --check` passes
- focused Maven tests were not run locally because a JDK is not configured in the environment

Fixes #9156


jsenko (commented):
## Review

Fixes #9156 — tightens MCP tool type detection to reject malformed required fields.

One should-fix and one observation. The fix itself is correct: `name` must be a non-empty string and `inputSchema` must be an object, matching what the downstream validator enforces.

### should-fix: Confirm tests pass locally

The PR body states "focused Maven tests were not run locally because a JDK is not configured in the environment." Run:

```bash
./mvnw test -pl schema-util/util-provider -am -DskipTests=false
```

and confirm the new `testMcpToolAccepterRejectsMalformedRequiredFields` passes before merge.

- **Accepter vs. validator strictness**: The accepter allows `inputSchema: {}` (empty object) while the validator would reject it (requires `type: "object"` inside). This is consistent with the accepter being a lightweight heuristic for type detection — it should be permissive enough to catch plausible MCP tools, while the validator enforces full schema compliance. The existing `AgentCardContentAccepter` follows the same pattern (just checks `has("name")`).
- **Test location**: The new test is in `McpToolContentValidatorTest.java`, which already contains accepter tests (`testMcpToolAccepterAccepts`, `testMcpToolAccepterRejectsAgentCard`, `testMcpToolAccepterRejectsInvalidJson`). Consistent with the existing pattern.

hcthakur2004:
I pulled the latest branch state and attempted the requested focused test: ./mvnw test -pl schema-util/util-provider -am -DskipTests=false. The run is currently blocked because JAVA_HOME is not configured in my local environment. The latest branch is up to date, and I am working to rerun the test with a configured JDK.

carlesarnal:
Bear in mind that all work related to https://github.com/Apicurio/apicurio-registry/issues/8427 will be blocked until the term. I'm accepting this since it's a bug, but have that in mind.

carlesarnal (dismissed):

hcthakur2004:
Fixed the Quick Verify compilation failure by correcting the ContentTypes import in McpToolContentValidatorTest from io.apicurio.registry.content to io.apicurio.registry.types. Pushed as commit 4b691a23. The focused Maven test could not be run locally because JAVA_HOME is not configured; CI should now rerun on the pushed commit.

carlesarnal (approved):

PR #9158, opened 2026-08-01
fix: route well-known discovery endpoints through configured Registry…

### LINKED ISSUE #9098: fix(mcp): use the configured Registry client for well-known discovery tools
## Problem

`RegistryService` creates a configured Registry SDK client with URL normalization, retry, OAuth2, trust-store, key-store/mTLS, trust-all, and hostname-verification settings in `RegistryService.java` lines 46–153.

However, the four well-known discovery methods bypass that client. They construct URLs directly from `rawBaseUrl` and send them through a separate, unconfigured JDK `HttpClient` in `RegistryService.java` lines 468–527.

Consequences include:

- OAuth2 credentials configured for other MCP tools are not attached to Agent Card or MCP Tool requests.
- Configured custom CAs, mTLS credentials, trust-all, and hostname-verification behavior are ignored.
- The documented default `registry.url=localhost:8080` is normalized by the SDK, but is not an absolute URI for `HttpRequest`; `URI.create("localhost:8080/.well-known/...")` has no HTTP scheme.
- These four calls do not receive the retry behavior enabled for the primary client.

The comment saying there is “no SDK support” is now stale. Closed issue #8799 added the well-known endpoints to the v3 OpenAPI/SDK surface. The v3 alias intentionally delegates to the canonical endpoints in `V3WellKnownResource.java`.

## Reproduction

1. Secure Registry with OAuth2 or mTLS.
2. Configure the MCP server’s existing authentication/TLS properties.
3. Verify that a normal tool such as `get_server_info` succeeds.
4. Call `search_agent_cards` or `search_mcp_tools`.
5. The request is made by the bare JDK client and fails authentication or TLS negotiation.

A second deterministic case is starting the MCP server with the default schemeless `registry.url`; initialization succeeds through SDK normalization, but the raw well-known request cannot produce a valid HTTP request URI.

## Proposed solution

Use the generated well-known request builders available from the existing configured `RegistryClient`:

- Route search requests through the generated Agent and MCP Tool collection builders.
- Set `name`, `skill`, `capability`, `parameter`, and `limit` through generated query-parameter objects rather than manual URL concatenation.
- Route individual artifact retrieval through the generated registered-artifact builders.
- Return or serialize the generated response models through the MCP tool layer.
- Remove `wellKnownHttpClient`, `fetchWellKnownEndpoint`, manual URL encoding, and the stale “no SDK support” comment.

This preserves canonical `/.well-known` behavior through the existing v3 alias without duplicating authentication or TLS configuration.

## Acceptance criteria

- All four well-known MCP operations use the configured SDK request adapter.
- OAuth2 and TLS settings apply consistently to normal and well-known requests.
- A schemeless configured base URL is normalized consistently.
- Query parameters and path segments use generated SDK encoding.
- Non-success responses retain useful MCP-facing error messages without exposing internal exceptions.
- Existing public endpoint paths remain unchanged.

## Tests

Add MCP tests covering:

- Authorization being attached to well-known requests.
- A configured SDK TLS setting being used by the request path.
- The default schemeless URL.
- Agent and MCP Tool search-filter propagation.
- Individual artifact retrieval and non-success responses.

Suggested checks:

```bash
./mvnw test -pl :apicurio-registry-mcp-server -am

./mvnw checkstyle:check -pl :apicurio-registry-mcp-server
```

AyushSrivastava1818:
## Summary

Replace the raw `java.net.http.HttpClient` implementation for well-known
discovery endpoints with the generated Kiota RegistryClient SDK builders.

Previously, `searchAgentCards`, `getAgentCard`, `searchMcpTools`, and `getMcpTool`
bypassed the configured client using a separate `HttpClient` that did not
inherit OAuth2 authentication, TLS trust-store/mTLS, hostname verification,
retries, or URL normalization settings.

## Changes

- Route all four well-known methods through `client.wellKnown().agents()`
  and `client.wellKnown().mcpTools()` builders off the already-configured
  `RegistryClient` instance
- Serialize search results via `utils.toPrettyJson()`
- Read artifact `InputStream` responses with try-with-resources and `StandardCharsets.UTF_8`
- Remove `wellKnownHttpClient`, `fetchWellKnownEndpoint()`, manual URL concatenation,
  `URLEncoder` calls, and stale comments

## Tests

- OAuth2 bearer token header propagation
- Query parameter and `limit=50` propagation for both search endpoints
- Artifact card/tool retrieval (`getAgentCard`, `getMcpTool`)
- Error handling on non-2xx responses (HTTP 404)

Fixes #9098

### DIFF
--- a/mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
@@ -35,6 +35,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


paoloantinori (commented):
Overall this is a sound, security-positive refactor: routing the four well-known discovery endpoints through the already-configured `RegistryClient` closes the OAuth2/TLS/hostname-verification bypass that the standalone `HttpClient` had, and the try-with-resources handling for artifact responses is clean. A few behavioral changes are worth confirming before merge.

- `searchAgentCards` (RegistryService.java:482) and `searchMcpTools` (RegistryService.java:504) now return `utils.toPrettyJson(results)` — a re-serialization of the deserialized Kiota model — whereas `getAgentCard` (RegistryService.java:490) and `getMcpTool` (RegistryService.java:512) return the raw response bytes. The list endpoints can silently drop any server field the generated model doesn't bind (e.g. `count`) and reorder keys; consider returning raw bytes for the list endpoints too, or confirm the round-tripped shape is what consumers expect.
- HTTP error handling changed for all four methods: non-2xx responses now propagate as the Kiota client's own exception rather than the previous `ToolCallException("Registry returned HTTP …")`. The error is still surfaced, but any MCP-layer handling that keys off `ToolCallException` or parses the status from that message will behave differently — worth a quick check.
- The new test mock matches request paths against `/well-known/...` (no leading dot) and reads `exchange.getRequestURI().getPath()`, which excludes the query string — so the `contains("/well-known/agents?")` / `contains("/well-known/mcp-tools?")` clauses (RegistryServiceTest.java:66, :86) can never fire. More importantly, if the generated client sends the dotted `/.well-known/...` path that the old `rawBaseUrl + "/.well-known/agents"` code targeted, the `endsWith`/`matches` predicates at RegistryServiceTest.java:66/:76/:106 won't match and `testIndividualArtifactRetrieval` / `testErrorHandlingOn404` would fall through to the default `{}` handler (200, no exception). Logging or asserting the received path would confirm these tests actually exercise the intended branches.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
```diff
@@ -465,65 +466,51 @@ public ConfigurationProperty getConfigurationProperty(String propertyName) {
...
+                config.queryParameters.name = name;
+            }
+            if (skill != null && !skill.isBlank()) {
+                config.queryParameters.skill = new String[]{ skill };
+            }
+            if (capability != null && !capability.isBlank()) {
+                config.queryParameters.capability = new String[]{ capability };
+            }
+        });
+        return utils.toPrettyJson(results);
```
paoloantinori, line 482: `searchAgentCards` returns `utils.toPrettyJson(results)`, re-serializing the deserialized Kiota model rather than passing through the server's raw body. Any field the generated model doesn't bind (e.g. `count`) is dropped and key ordering changes, whereas `getAgentCard` returns the raw bytes verbatim. If callers rely on the original response shape, consider reading these list responses as raw bytes too (consistent with `getAgentCard`/`getMcpTool`), or confirm the round-tripped model preserves everything consumers need. The same applies to `searchMcpTools` (line 504).

mcp/src/test/java/io/apicurio/registry/mcp/RegistryServiceTest.java
```diff
@@ -0,0 +1,275 @@
...
+                String response = "{\"version\":\"3.0.0\"}";
+                exchange.getResponseHeaders().set("Content-Type", "application/json");
+                exchange.sendResponseHeaders(200, response.length());
+                try (OutputStream os = exchange.getResponseBody()) {
+                    os.write(response.getBytes(StandardCharsets.UTF_8));
+                }
+                return;
+            }
+
+            if (path.endsWith("/well-known/agents") || path.contains("/well-known/agents?")) {
```
paoloantinori, line 66: These predicates use `exchange.getRequestURI().getPath()`, which excludes the query string, so the `contains("/well-known/agents?")` clause here (and the analogous one at line 86) can never match. Separately, they match `/well-known/...` without the leading dot; if the generated `RegistryClient` sends the dotted `/.well-known/...` path that the old `rawBaseUrl + "/.well-known/agents"` code targeted, none of these branches (lines 66, 76, 106) will fire and `testIndividualArtifactRetrieval` / `testErrorHandlingOn404` would hit the default `{}` handler. Consider logging or asserting the received path so the tests provably exercise the intended branches.

AyushSrivastava1818:
Thanks for the careful review @paoloantinori !
Good catch on both points.
For the tests, you're right: getRequestURI().getPath() doesn't include the query string, so the contains(...?) checks are ineffective. I'll update the mock to match the actual request path emitted by the SDK, assert the received path explicitly, and validate query parameters separately from the path.
Regarding the search endpoints, I'll investigate whether the generated Kiota client exposes a way to retrieve the raw response body for those operations. If it does, I'll switch to that so the response shape is preserved. If not, I'll document why the current model serialization is acceptable and confirm whether any unmapped fields could be lost.

paoloantinori:
Thanks @AyushSrivastava1818 — the test fix landed cleanly. Capturing `getRequestURI().toString()` (full URI with query string) instead of `.getPath()` makes the `limit=50` / `name=` / `skill=` / `parameter=` assertions actually exercise what the SDK sends, which was the gap.

On the search endpoints, `utils.toPrettyJson(results)` is the right call: letting the Kiota SDK deserialize into typed objects and re-serializing is cleaner than chasing the raw response body, and it keeps the contract aligned with the rest of the client. No further concerns from me on this PR.

EricWittmann:
Thanks @AyushSrivastava1818 — good security fix. The old standalone HttpClient was silently bypassing all the auth and TLS configuration. Clean refactor to the SDK client. Merged!

PR #9159, opened 2026-08-01
fix(protobuf): preserve map key/value type qualification in toDescriptor

krickert:
## The bug

`FileDescriptorUtils.toDynamicMessage` expands `map<K, V>` fields into the synthetic entry message using `ProtoType.getSimpleName()` for the key and value types. That turns `map<string, google.protobuf.Value>` into a value field of bare type `Value`, which fails descriptor validation with `'"Value" is not defined'` for any value type outside the entry's own resolution scope: cross-package types, and nested types like `Outer.Inner`.

## Why it surfaced now

This path went largely unexercised while `fileDescriptorToProtoFile` emitted maps as pre-expanded synthetic entries with fully qualified type names. Since #8771 switched emission to real `map<K, V>` syntax with relative type names (matching protoc's canonical form), every schema registered through the protobuf serde that contains such a map can no longer be read back by a client:

1. `toDescriptor` throws `IllegalStateException` wrapping `'"Value" is not defined'`
2. `ProtobufSchemaParser` falls back to `protoFileToFileDescriptor(fileElem)`, which parses without the resolved references
3. the parse dies with `unable to find <first import>` even though the caller had already resolved that import

Canonical protoc text always used map syntax, so schemas registered by non-Apicurio producers hit the same failure on earlier releases too.

Reproduced against a production schema closure (8 artifacts, registered and re-resolved through `ProtobufSchemaParser` exactly as `AbstractSchemaResolver.resolveReferences` does): fails on 3.3.1, round-trips cleanly with this fix.

## The fix

Build the entry's key and value fields from `ProtoType.toString()`, which preserves the qualification exactly as declared. Two lines.

## Tests

- `toDescriptor` with cross-package (`google.protobuf.Value`), nested (`Outer.Inner`), and cross-file dependency value types
- the full descriptor to text to descriptor round trip (`fileDescriptorToProtoFile` then `toDescriptor`)
- a serde-level `ProtobufSchemaParser.parseSchema` case with resolved references

All five fail without the fix; `protobuf-schema-utilities` (65) and the serde chain (`schema-resolver` 54, `serde-common` 31, `serde-common-protobuf` 24) are green with it.

## Note on a second latent issue

The fallback in step 2 above discards the caller's resolved references, so ANY `toDescriptor` failure on a schema with non-well-known imports becomes an unresolvable-import error. With this fix that path is no longer reachable for map schemas, but the fallback remains reference-blind; happy to file that separately if useful.


utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtils.java
```diff
@@ -1747,8 +1747,13 @@ else if (type instanceof EnumElement) {
...
                 mapMessage.setMapEntry(true);
-                mapMessage.addField(null, keyType.getSimpleName(), KEY_FIELD, 1, null);
-                mapMessage.addField(null, valueType.getSimpleName(), VALUE_FIELD, 2, null);
+                // Key and value types must keep their qualification exactly as declared:
+                // getSimpleName() turns map<string, google.protobuf.Value> into a value
+                // field of bare type "Value", which fails descriptor validation with
+                // '"Value" is not defined' for any value type outside the entry's own
+                // resolution scope (cross-package, or nested like Outer.Inner).
+                mapMessage.addField(null, keyType.toString(), KEY_FIELD, 1, null);
+                mapMessage.addField(null, valueType.toString(), VALUE_FIELD, 2, null);
```
paoloantinori, line 1756: This line applies `toString()` uniformly to all value types, but every test added in this PR covers a message value type (`google.protobuf.Value`, `Outer.Inner`, `test.common.Item`). Consider adding a `map<string, SomeEnum>` case to confirm enum map values are classified as `TYPE_ENUM` rather than `TYPE_MESSAGE` — the message-vs-enum decision for the value field is the one value-type kind not exercised by the new suite. This is pre-existing handling (the old `getSimpleName()` path was identical), not a regression; just worth locking in since the fix now makes cross-package enum values reachable that previously failed at name resolution.

paoloantinori (commented):
This is a correct, minimal fix: building the synthetic map-entry's key/value fields from `ProtoType.toString()` instead of `getSimpleName()` preserves type qualification, which is exactly what protobuf descriptor validation needs to resolve out-of-scope value types (cross-package, nested, cross-file). I found no regression path — qualified names are always at least as resolvable as simple names under protobuf scope rules, and scalar keys/values are unaffected.

- `FileDescriptorUtils.java:1755-1756` — sound change. Scalar keys/values are unchanged (`toString()` == `getSimpleName()` for scalars, and proto map keys are always scalar), and no value type that resolved under the simple name can fail under the fully-qualified name; both dotted and leading-dot FQN forms are accepted by descriptor name resolution.
- Tests are thorough and on-point: cross-package (`google.protobuf.Value`), nested (`Outer.Inner`), cross-file dependency, the `fileDescriptorToProtoFile`↔`toDescriptor` round-trip, and a serde-level `parseSchema` with resolved references (`FileDescriptorUtilsTest`, `ProtobufSchemaParserTest`).
- One easy-win coverage gap: every new test value type is a message; an enum-valued map is not covered (see inline on line 1756).
- The author's note about the reference-blind fallback in `protoFileToFileDescriptor` is correctly scoped out — it is pre-existing and no longer reachable for map schemas after this fix.

utils/protobuf-schema-utilities/src/main/java/io/apicurio/registry/utils/protobuf/schema/FileDescriptorUtils.java
krickert: Done

paoloantinori (commented):
This is a correct, minimal fix: using `ProtoType.toString()` instead of `getSimpleName()` when building the synthetic map entry's key/value fields preserves type qualification, which is exactly what descriptor resolution needs for cross-package, nested, and cross-file-dependent value types. I found no blocking or major issues; the load-bearing change is two lines and the test coverage directly reproduces the reported failure.

- `FileDescriptorUtils.java:1755-1756` — `toString()` is strictly at least as qualified as `getSimpleName()`, so this cannot regress scalar or same-scope cases. Map keys are always scalar per the proto spec, so the key line is effectively a consistency no-op while the value line is the actual fix; this checks out against protobuf name-resolution semantics.
- For nested value types written relatively (e.g. `Outer.Inner`), `toString()` keeps the name as declared and protobuf's outward scope walk resolves it to the full name — the new `testToDescriptorMapWithNestedMessageValueType` in `FileDescriptorUtilsTest.java` confirms resolution to `test.maps.Outer.Inner`, which is the precise case `getSimpleName()` broke.
- Enum map values are classified correctly (`FieldDescriptor.Type.ENUM`) now that the qualified name reaches the builder, including a cross-package enum — covered by `testToDescriptorMapWithEnumValueType`.
- Test coverage is well-targeted: cross-package (`google.protobuf.Value`), nested, cross-file dependency, enum classification, the full descriptor→text→descriptor round trip, and a serde-level `ProtobufSchemaParser.parseSchema` case exercising the actual client path (`testParseSchemaMapWithQualifiedValueTypeAndReferences`).
- Pre-existing and out of scope (author already flagged it): the `protoFileToFileDescriptor(fileElem)` fallback in `ProtobufSchemaParser` discards caller-resolved references, so any other `toDescriptor` failure on a schema with non-well-known imports would still surface as an unresolvable-import error. This PR makes that path unreachable for map schemas but does not address the fallback; worth tracking as a follow-up rather than blocking here.

Caveat: I assessed this against the diff, protobuf/Wire type-resolution semantics, and the included tests rather than by running the suite (no checkout or build is available in this read-only review context).

krickert:
This should be good to go - I have to maintain my own serde until this is fixed.

paoloantinori (commented):
The fix is correct, minimal, and well-scoped: `FileDescriptorUtils.toDynamicMessage` builds the synthetic map-entry's key/value fields from `ProtoType.toString()` instead of `getSimpleName()`, preserving type qualification exactly as declared so cross-package (`google.protobuf.Value`) and nested (`Outer.Inner`) value types resolve at descriptor-validation time. I found no blocking or major issues, and the new tests reproduce each failure mode described in the PR.

- `FileDescriptorUtils.java:1755-1756` — the two-line change is the right fix. For scalar map keys (proto restricts keys to scalars) `toString()` is equivalent to `getSimpleName()`, so the key-side change is behaviorally a no-op and only keeps the two lines consistent; no regression risk there.
- `FileDescriptorUtils.java:1750-1754` — the added comment captures a genuine non-obvious invariant (descriptor validation fails without qualification), which usefully guards against a future regression that "simplifies" it back to `getSimpleName()`.
- Test coverage is strong and matches the reported failure modes: cross-package value (`FileDescriptorUtilsTest.testToDescriptorMapWithCrossPackageValueType`), nested message value, cross-file dependency value, enum value (also confirming `TYPE_ENUM` classification survives), the descriptor→text→descriptor round trip, plus a serde-level case in `ProtobufSchemaParserTest.testParseSchemaMapWithQualifiedValueTypeAndReferences` exercising resolved references end to end. All five are reported to fail without the fix.
- The author's separately-noted latent issue — the `protoFileToFileDescriptor` fallback in `ProtobufSchemaParser` discards the caller's resolved references on any `toDescriptor` failure — remains real but is correctly out of scope here (this fix makes that path unreachable for map schemas); worth tracking as its own ticket as offered.

jsenko (commented):
Fixes a protobuf map key/value type qualification bug in `FileDescriptorUtils.toDescriptor`.

## Summary

Correct, minimal fix: `ProtoType.toString()` instead of `getSimpleName()` preserves type qualification for cross-package, nested, and cross-file value types in synthetic map entry messages. Two-line change. No must-fix or should-fix issues found.

- **Regression safety**: `toString()` is strictly at least as qualified as `getSimpleName()` — same-scope and scalar cases are unchanged. Map keys are always scalar per proto spec, so the key-side change is a consistency no-op.
- **Test coverage**: 5 focused tests covering cross-package (`google.protobuf.Value`), nested (`Outer.Inner`), cross-file dependency, enum value classification, and serde-level `parseSchema` with resolved references. All reproduce the failure without the fix.
- **Latent fallback issue**: The `protoFileToFileDescriptor` reference-blind fallback in `ProtobufSchemaParser` is correctly scoped out — pre-existing and no longer reachable for map schemas after this fix. The author offered to file it separately, which would be worthwhile.

Well-written PR description with clear reproduction steps and root cause analysis.

krickert:
Is this one going to get merged?  It looks like it's been reviewed twice and it's a critical bug imho since it easily breaks proto serde.

paoloantinori:
Hi @krickert — to give you an honest status: my three reviews here were all COMMENT-event (not APPROVE), and jsenko's was too, so from GitHub's side the PR is waiting on a formal approval before it can merge. That's the `blocked` state, not a technical objection.

Nothing in the diff has changed since my last read (head `5624f61`), and my assessment is unchanged: the fix is correct, minimal, and well-tested. The qualification-preservation change in `toDynamicMessage` is the right call, and the test coverage for both cross-package and nested-type map values is solid.

I can't approve from my side, but I've flagged it for jsenko / the rest of the maintainers. Thanks for the patience and for keeping your own serde patched in the meantime.

krickert:
@jsenko this is preventing me from using the latest build as it broke protobuf serde.  Nuding to get it merged.

krickert:
> This PR has been inactive. @krickert, please update or comment
> to keep it open. It will be closed in 7 more days if no activity occurs.
> Use `/unstale` to remove this label.
> 
It broke your protobufs and I fixed it.  no one is looking at it. I feel like this entire thread is agents approving my code and no human knowing it's broken

Do what you want with it because I don't feel like I'm talking to a human. :) 

In the meantime I had to write my own serde.  I imagine you would like to have this fixed.  So please let me know if you guys want this

krickert:
/unstale 

@jsenko  - can someone look at this one? it's preventing me from upgrading apicurio.  I know the bots approved the work and it looks like some humans too.  I think it just needs to be merged.  It's a small fix but right now your protobuf serde is broken in prod.

jsenko (approved):

jsenko:
Thank you!

PR #8922, opened 2026-07-27
fix(rules): resolve locale-sensitive toUpperCase and remove unused field in RulesPropertiesImpl.java

Manar-Elhabbal7:
This pull request improves the quality and maintainability of `RulesPropertiesImpl.java` by introducing the following changes:

### Changes

- **Locale-safe capitalization**
  - Replaced `rulePropertyName.toUpperCase()` with `rulePropertyName.toUpperCase(Locale.ROOT)` to ensure locale-independent behavior.
  - This prevents OS-specific issues, such as the Turkish dotted/dotless **I** case conversion, and aligns with the project's guidelines documented in `CLAUDE.md`.

- **Code cleanup**
  - Removed the unused `properties` field.
  - Removed the associated `@SuppressWarnings` annotation.
  - This simplifies the implementation and improves maintainability.

## Related Issue

Closes #8886

PR #9165, opened 2026-08-02
fix(build): enforce documented Checkstyle rules instead of two-rule config

### LINKED ISSUE #9164: Checkstyle only enforces two rules; the documented code style is not actually checked
## Description

**Registry Version**: snapshot (`main`, currently `3.3.2-SNAPSHOT`)
**Persistence type**: N/A - this is a build tooling issue, not runtime behaviour

The Maven build points Checkstyle at `.checkstyle/simple.xml`, which contains exactly two rules: `UnusedImports` and `RedundantImport`.

Meanwhile `.checkstyle/checkstyle.xml` sits in the same directory with a full rule set (star imports, brace style, constant naming, indentation, the `Locale`-argument regex, complexity limits), and that is the file our docs point contributors at. `CLAUDE.md` says "Checkstyle config: `.checkstyle/checkstyle.xml`" and lists those rules under Code Style; `.claude/rules/code-style.md` says to "follow `.checkstyle/checkstyle.xml` strictly". Nothing in the build ever reads that file, so none of it is enforced.

Two smaller things fall out of the same mismatch:

- The `headerLocation` element in `pom.xml` is set to `.checkstyle/java.header`, but `simple.xml` declares no `Header` or `RegexpHeader` module, so the license header check is inert. It has presumably never run. Only 42 of the 693 files under `app/src/main/java` carry the header (143 of 2168 repo-wide), and most newly added files don't have one either.
- `utils/exportConfluent/pom.xml` repeats the whole Checkstyle configuration with its own relative paths, hardcoding `simple.xml` a second time. That module already defines its own `projectRoot` property, so the duplicated block is redundant and is just a second copy to drift out of sync.

The practical effect is that contributors read the conventions, assume CI will catch violations, and it doesn't. Style drift then accumulates quietly until someone notices a rule has been broken in dozens of places.

Worth being upfront about the size of this: I measured the gap with Checkstyle 13.9.0 across all 121 `src/{main,test}/java` roots (2168 files). Enabling `checkstyle.xml` as-is produces **3617 violations**, so this can't just be switched over in one go:

| Rule | Violations |
|---|---|
| RegexpHeader | 2042 |
| Indentation | 467 |
| ConstantName | 170 |
| UnnecessaryParentheses | 158 |
| Regexp (`toUpperCase`/`toLowerCase` without `Locale`) | 134 |
| AvoidStarImport | 125 |
| RightCurly / LeftCurly | 108 |
| everything else | ~413 |

### Environment

Build environment only, reproducible anywhere:

- JDK 21 (Temurin), Maven wrapper from the repo
- maven-checkstyle-plugin 3.6.0, Checkstyle 13.9.0
- macOS 15 and the `ubuntu-latest` GitHub runners behave identically

Worth noting Checkstyle does run and does gate merges: the Build job in `.github/workflows/verify-build.yaml` runs `mvn clean package` with no `-Dcheckstyle.skip` and feeds the Verification Gate. So the enforcement path works fine, it's just aimed at an almost empty rule set. (`-Dlocal` and the unit test workflow skip Checkstyle deliberately, which is fine.)

### Steps to Reproduce

1. Check out `main` and create a file that breaks five separately documented rules at once - star import, brace on its own line, a non-uppercase constant, `toUpperCase()` with no `Locale`, and tab indentation:

```java
// common/src/main/java/io/apicurio/registry/ScratchViolation.java
package io.apicurio.registry;

import java.util.*;

public class ScratchViolation
{
	private static final String badConstantName = "x";

	public String shout(String s) {
		java.util.List<String> l = new ArrayList<>();
		return s.toUpperCase() + badConstantName + l.size();
	}
}
```

2. Run `./mvnw validate -pl common`
3. The build passes and reports zero violations.

To confirm the rules do exist and simply are not wired up, repoint the build at the config the docs advertise and re-run. Note that `-Dcheckstyle.config.location` does **not** work here: the root pom sets the config location explicitly inside the plugin's configuration block, and explicit POM configuration beats a user property, so the flag is silently ignored. Edit the pom instead:

```bash
perl -pi -e 's{\.checkstyle/simple\.xml}{.checkstyle/checkstyle.xml}' pom.xml
./mvnw validate -pl common     # now fails, all five violations reported
git checkout pom.xml
```

**Expected**: violations of documented conventions are reported. Ideally the rules we can enforce today block the build, and the ones the existing code doesn't satisfy yet are at least surfaced as warnings so they can be driven down over time.

**Actual**: nothing is reported. Only unused and redundant imports are checked, and `headerLocation` is wired to a check that doesn't exist.

### Logs

On `main`, with the file above in place:

```
[INFO] --- checkstyle:3.6.0:check (validate) @ apicurio-registry-common ---
[INFO] You have 0 Checkstyle violations.
[INFO] BUILD SUCCESS
```

For contrast, the same file once the config location points at `.checkstyle/checkstyle.xml`:

```
[ERROR] ScratchViolation.java:1: Missing a header - not enough lines in file. [RegexpHeader]
[ERROR] ScratchViolation.java:3:17: Using the '.*' form of import should be avoided - java.util.*. [AvoidStarImport]
[ERROR] ScratchViolation.java:6:1: '{' at column 1 should be on the previous line. [LeftCurly]
[ERROR] ScratchViolation.java:7:1: File contains tab characters (this is the first instance). [FileTabCharacter]
[ERROR] ScratchViolation.java:7:37: Name 'badConstantName' must match pattern '(^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$)|(^log$)'. [ConstantName]
[ERROR] ScratchViolation.java:11: Line matches the illegal pattern '\.to(Lower|Upper)Case\(\)'. [Regexp]
```

### Proposed direction

Happy to open a PR, but flagging the approach first since a naive switch-over would drop 3617 violations onto every open branch.

1. Make `.checkstyle/checkstyle.xml` the single config the build uses and delete `simple.xml`.
2. Tier it. Rules already clean across the whole repo go in at `severity=error` and start blocking; the rest go in at `severity=warning` with their current violation count recorded beside them as a worklist. Measuring this, **13 rules are clean today** and can be enforced immediately with zero source changes: `FileTabCharacter`, `UnusedImports`, `RedundantImport`, `IllegalImport`, `OneStatementPerLine`, `DefaultComesLast`, `UpperEll`, `PackageName`, the three `*TypeParameterName` checks, `MethodParamPad` and `TypecastParenPad`. That takes us from 2 enforced rules to 13 without touching a single `.java` file.
3. Drop the duplicated Checkstyle block in `utils/exportConfluent/pom.xml` so it inherits from the root pom.
4. Correct the docs so they distinguish what is enforced from what is still being driven to zero.

On the license header specifically, I'd rather we decide than leave it half-wired. Given 93% of files don't have one and most new files don't either, my read is the project has effectively settled on not using per-file headers and the root `LICENSE` covers us. If that's right, the docs should say so and `headerLocation` should be removed. If maintainers would prefer to adopt headers, that should be its own PR that adds them everywhere and then turns the check on. **This is the one open question I'd like a steer on before the PR is reviewed** - the rest is mechanical.

I'd also like to add a check to `scripts/validate-files.sh` (already run by the lint job) asserting that the config location resolves to the real config, that no submodule overrides it, and that every rule documented as enforced really is `severity=error` - so this can't silently drift again.

1PoPTRoN:
Fixes #9164

## What

The build pointed Checkstyle at `.checkstyle/simple.xml` (two rules: `UnusedImports`, `RedundantImport`) while the docs told contributors to follow `.checkstyle/checkstyle.xml` (the full set). None of the documented conventions — star imports, brace style, constant naming, indentation, the `Locale`-argument rule — were actually enforced. The `headerLocation` element was also wired up with no header module to consume it, so the license header check was inert.

## Why not just switch the config over

Enabling `checkstyle.xml` as-is produces **3617 violations** across the 121 `src/{main,test}/java` roots (2168 files). A mass reformat would collide with every open branch and breaks the repo's own "PR contains no unrelated changes" rule.

## Approach — tiered config

- **13 rules enforced** at `severity=error`, each verified clean repo-wide: `FileTabCharacter`, `UnusedImports`, `RedundantImport`, `IllegalImport`, `OneStatementPerLine`, `DefaultComesLast`, `UpperEll`, `PackageName`, the three `*TypeParameterName` checks, `MethodParamPad`, `TypecastParenPad`. That takes enforcement from 2 rules to 13.
- Remaining rules staged at `severity=warning`, each annotated inline with its current violation count as a promotion worklist. Drive one to zero, then add a `severity=error` property to that rule and it starts blocking.
- The Checker-level default severity is `warning`, so staged rules carry no severity property of their own and a rule added to the file later is non-blocking until somebody explicitly promotes it. That makes the config safe to grow incrementally.

**Zero `.java` files changed.** Full reactor validates clean: 67 modules, 0 violations, `BUILD SUCCESS`.

## Also in this PR

- `utils/exportConfluent/pom.xml` duplicated the entire Checkstyle configuration with its own relative paths, hardcoding `simple.xml` a second time. It already defines its own `projectRoot` property, so the redundant block is removed and it now inherits from the root pom. Verified it still resolves correctly — it reports exactly the 3 warnings the standalone Checkstyle CLI predicts for that module.
- `simple.xml` deleted, since nothing references it any more.
- `scripts/validate-files.sh` (already run by the `lint-and-validate` job) now also asserts that the pom's config location resolves to the real config, that no submodule overrides it, and that every rule documented as enforced really is `severity=error`. Tested against five separate re-introduced regressions; each is caught.
- Docs updated (`CLAUDE.md`, `.claude/rules/code-style.md`, `.claude/agents/code-reviewer.md`) to distinguish what is enforced from what is still being driven to zero.

## Verification

```
./mvnw clean validate -Dfull      # exit 0 — 67 modules, 0 violations, BUILD SUCCESS
bash scripts/validate-files.sh    # exit 0 — "Checkstyle config ok: 13 rules enforced"
```

Sanity check that the enforced tier actually bites — a file with a tab character now fails the build, while staged-tier issues (star import, brace style, constant naming, `toUpperCase()` without `Locale`) are reported as warnings.

Note that `includeTestSourceDirectory=true` means test roots are now held to all 13 enforced rules rather than the previous 2. That is intentional, and is why the measurement covered all 121 `src/{main,test}/java` roots rather than main only.

## Backward compatibility / backend parity

No runtime, storage, API or wire-format impact — this is build tooling only. No storage variant, SQL dialect, or public API surface is affected. No upgrade or migration considerations.

## Review feedback addressed (589194e57)

- Tier-promotion instructions were backwards in `checkstyle.xml` and `CLAUDE.md` — staged rules have no override to delete, so promotion means *adding* `severity=error`. Corrected.
- Dropped the deprecated, now-redundant `failsOnError` from the root pom.
- `validate-files.sh`: excluded `target/` from the `headerLocation` scan, switched `echo` to `printf` when piping the config into `awk`/`grep` (the config holds regexes with backslashes), and anchored the enforced-rule detector to the `severity` property. The unanchored version reported 14 enforced rules when a staged rule was given an unrelated property valued `error`; the anchored one correctly reports 13.
- Repo-wide grep confirms no stray references to `simple.xml` or `headerLocation` outside the files this PR touches.

## Open question for maintainers

Per-file Apache license headers are left unchecked, and the docs now say so explicitly: only 143 of 2168 Java files carry one, and most newly added files do not — the repository-root `LICENSE` governs. I've kept `.checkstyle/java.header` and left the `RegexpHeader` rule in the config commented out with a note, so nothing is lost.

If you would rather adopt headers project-wide, that should be its own PR that adds them everywhere and then enables the rule. Happy to follow up either way — I just wanted the config and the docs to stop disagreeing.


paoloantinori (commented):
This is a careful, well-scoped change that fixes a genuine mismatch between the documented conventions and what the build actually enforced, with zero `.java` changes and a sound tiered-severity design; the new consistency guard in `scripts/validate-files.sh` is a strong addition. I found no correctness issues — the severity mechanism, the inherited `exportConfluent` config, and the validation logic all hold up, and the docs' violation counts match the inline config comments.

- `pom.xml:1281` — `<failsOnError>true</failsOnError>` is now redundant with `<failOnViolation>true</failOnViolation>` + `<violationSeverity>error</violationSeverity>` (added at L1286–1287), and `failsOnError` is the deprecated parameter. It's harmless, but dropping it (or keeping just one with a one-line note) would make the fail contract easier to read.
- `scripts/validate-files.sh` — the enforced-rule detector treats a module as enforced whenever any property line matches `value="error"`. That's correct for the current config, but anchoring the match to the `severity` property specifically would avoid a future false positive if a staged rule ever gains a non-severity property valued `error`.
- `.checkstyle/checkstyle.xml` — test roots (`includeTestSourceDirectory=true`) are now held to all 13 enforced rules rather than the previous 2 from `simple.xml`; your `validate -Dfull` confirms clean, so this is just a heads-up that enforcement scope for test sources widened meaningfully.
- Worth a quick repo-wide grep for stray references to the deleted `.checkstyle/simple.xml` and the removed `<headerLocation>` outside the two poms (CI workflows, IDE/editor Checkstyle configs, READMEs) — those paths aren't covered by this diff and a lingering reference would be the one thing that could break downstream.

pom.xml
```diff
@@ -1274,12 +1274,20 @@
           </dependency>
         </dependencies>
         <configuration>
-          <configLocation>${projectRoot}/.checkstyle/simple.xml</configLocation>
-          <headerLocation>${projectRoot}/.checkstyle/java.header</headerLocation>
+          <configLocation>${projectRoot}/.checkstyle/checkstyle.xml</configLocation>
           <suppressionsLocation>${projectRoot}/.checkstyle/suppressions.xml</suppressionsLocation>
           <includeTestSourceDirectory>true</includeTestSourceDirectory>
           <consoleOutput>true</consoleOutput>
           <failsOnError>true</failsOnError>
```
paoloantinori: Minor/easy win: with `<failOnViolation>true</failOnViolation>` + `<violationSeverity>error</violationSeverity>` now pinned below, `<failsOnError>true</failsOnError>` is redundant, and `failsOnError` is the deprecated parameter in maven-checkstyle-plugin. Consider dropping it (or keeping one of the two with a short comment) so the fail contract reads as a single mechanism. No functional change either way.

1PoPTRoN:
Thanks @paoloantinori — and the Copilot pass caught a real one. All points addressed in 589194e57.

**The promotion instructions were backwards.** Copilot is right, and this was the most valuable catch in the review. Staged rules are self-closing (`<module name="AvoidStarImport"/>`) with no severity property at all — they inherit the Checker-level default of `warning`. So there is nothing to "delete"; promoting means *adding* `severity=error`. My comment said the opposite in both `checkstyle.xml` and `CLAUDE.md`, which would have sent the next person the wrong way. Fixed in both, with an explicit note that staged rules carry no severity property of their own.

**`failsOnError` dropped.** You're right that it's both the deprecated parameter and redundant. `failOnViolation` + `violationSeverity` now stand alone as the fail contract, with the comment reworded to say so. Full reactor re-validated afterwards: 67 modules, 0 violations, no deprecation warnings.

**Severity anchoring — good catch, and it was demonstrably wrong.** I proved the old behaviour rather than assuming. Planting a non-severity property valued `error` on a staged rule:

```
OLD matcher: Checkstyle config ok: 14 rules enforced   <- AvoidStarImport miscounted
NEW matcher: Checkstyle config ok: 13 rules enforced   <- correct
```

The detector now requires `name="severity"` and `value="error"` on the same line. Added as a sixth regression case.

**`target/` exclusion + `printf`.** Both real. The `<headerLocation>` scan now filters `/target/` like the `configLocation` scan already did — a generated pom from an earlier local build could otherwise have failed the lint job on a clean tree. And `printf` replaces `echo` in all three places that pipe the config into `awk`/`grep`; the config contains `\.to(Lower|Upper)Case\(\)`, so `echo`'s implementation-defined escape handling was a genuine portability hazard. The failure path now also prints the offending references instead of just asserting.

**On the stray-reference grep** — ran it repo-wide, and it's clean:

- `simple.xml`: **no references anywhere** outside the two poms this PR already changed.
- `headerLocation` / `java.header`: only in `scripts/validate-files.sh`, which is the guard itself referring to them deliberately.
- Checked `ide-config/` too — it holds `eclipse-format.xml` and `eclipse.importorder`, neither of which references a Checkstyle config. `.claude/hooks/checkstyle-before-commit.sh` invokes the Maven goal rather than naming a config file, so it picks up the new one automatically.

**On the test-source scope widening** — agreed, and worth stating plainly for the record: `includeTestSourceDirectory=true` means the 304 test files under `app/src/test/java` and the rest of the test roots are now held to 13 rules rather than 2. That was deliberate, and it's why I measured across all 121 `src/{main,test}/java` roots rather than main only. `validate -Dfull` is clean, so nothing regresses today — but it does mean a future test file with a tab character will now fail the build, which is the intent.

Still open for a maintainer steer: the **license-header question** in the PR description. That's the one decision I did not want to make unilaterally.

paoloantinori:
Thanks @1PoPTRoN — and good catch on the promotion-direction bug. You're right: the staged rules are self-closing with no severity property, so they inherit the Checker-level `warning` default, and "promote" means *adding* `severity="error"`, not deleting anything. Your corrected comment now matches the actual mechanism, and the `validate-files.sh` guard cross-checks the documented enforced set against `severity="error"` modules, so a future drift between docs and config fails the build rather than silently loosening enforcement. The tiered design and the guard together look correct to me now.

One small thing worth keeping an eye on as the staged tier gets driven down: the `Regexp` rule for `\.to(Lower|Upper)Case\(\)` is staged at 134 violations, but it's a correctness rule (locale-sensitive default locale), not purely cosmetic. When you next promote a batch, that one is a good candidate to prioritize over the whitespace/brace rules, since it catches real bugs rather than style drift.

1PoPTRoN:
Strongly agree, and the repo history backs you up harder than I expected. That rule has already cost the project twice in production code:

- `9c383a95f` — fix(search): use `Locale.ROOT` for state filter uppercasing in Elasticsearch query (#9075)
- `fb4959d6a` — fix(rules): resolve locale-sensitive `toUpperCase` in `RulesPropertiesImpl` (#8922)

Both are exactly what the staged rule matches. #9075 landed on `main` two days before this PR — so the rule, had it been enforced, would have caught a bug that instead shipped and needed a follow-up fix. That is a much better argument for prioritising it than "it's a correctness rule in principle".

I broke the 134 down to see how tractable a promotion pass actually is:

| | Violations |
|---|---|
| `src/main` | **76** |
| `src/test` | 58 |

The production half is concentrated — 31 in `app`, 11 in `utils/maven-plugin`, so 42 of 76 sit in two modules. That looks like one focused PR rather than a sprawling one, and the `src/main` half is where the actual risk lives.

I've captured this in the config itself (4ced6522f) rather than leaving it buried in this thread — the `Regexp` rule now carries a `PROMOTE THIS ONE FIRST` note with the reasoning, the two prior bug references, and the main/test split. That way whoever picks up the next promotion batch sees the priority ordering at the point of decision instead of having to rediscover it. Comment-only; the rule stays staged at warning, guard still reports 13 enforced, all regression cases still pass.

I've deliberately not folded the 134 fixes into this PR — that is a behaviour change across `src/main` and wants its own review and its own tests, separate from a build-config change. Happy to open it as the follow-up once this lands, if you'd like me to take it.

The license-header question is the only thing still outstanding from my side.

pom.xml
```diff
@@ -1274,12 +1274,19 @@
...
-          <failsOnError>true</failsOnError>
+          <!--
+            checkstyle.xml is tiered: enforced rules are severity=error, rules
+            still being driven to zero are severity=warning. These two settings
+            are the whole fail contract - the build fails on violations, and
+            only those at severity=error count. Pinning both keeps a plugin
+            default change from silently failing the build on the staged tier.
+          -->
+          <failOnViolation>true</failOnViolation>
+          <violationSeverity>error</violationSeverity>
```
paoloantinori, line 1289: `violationSeverity=error` gates the build *failure*, but it does not filter what `consoleOutput=true` prints — staged (warning-severity) violations are still emitted to the console. With ~3,617 staged violations across the reactor, a full `-Dfull` build will likely dump all of them as `[WARN]` lines on every CI run (the verification's '0 violations' reflects error-severity failures only, not total console output). Worth confirming the CI-log volume is acceptable, or dropping `consoleOutput` and relying on the report file for the staged tier.

paoloantinori (commented):
Overall this is a sound, well-verified build-tooling change with no runtime, API, or storage impact: the tiered severity model correctly limits build failures to the 13 promoted `error` rules, and the new `validate-files.sh` guard independently re-asserts that contract. The one thing worth a second look is the CI-log impact of keeping `consoleOutput=true` alongside ~3.6k staged violations.

- The fail contract is correct. `<failOnViolation>true</failOnViolation>` + `<violationSeverity>error</violationSeverity>` (pom.xml:1288-1289) gate failure to the 13 error-severity rules; staged rules carry no severity of their own and inherit the Checker-level `severity=warning` (`.checkstyle/checkstyle.xml`), so they cannot fail the build. That severity-inheritance assumption is load-bearing and it holds.
- The `validate-files.sh` guard is correct for this config: comment stripping (`sed` single-line + `awk` multi-line block) keeps the commented-out `RegexpHeader` from being mistaken for live, and anchoring the enforced-rule detector on `name="severity"` + `value="error"` (rather than any property valued `error`) reproduces the documented count of 13 (scripts/validate-files.sh:80).
- `utils/exportConfluent/pom.xml` correctly drops its duplicated config and inherits from the root; it relies on the module's own `projectRoot` resolving to the repo root, which the author verified by matching the standalone CLI's 3 warnings, so inheritance resolves as intended (utils/exportConfluent/pom.xml:115).
- Minor: `.checkstyle/java.header` is now unreferenced, but it is intentionally retained next to the commented `RegexpHeader` for a future headers PR — noting only, no action needed.

1PoPTRoN:
Good call on the console output — fixed in 4725b6c88, and testing it turned up a subtlety worth recording.

**One correction on the number first.** The staged tier emits ~1.3k lines, not ~3.6k. The 3617 figure in the description is what the *full* rule set produces, and 2042 of those are `RegexpHeader`, which this config does not enable. Measured on an actual `validate -Dfull`:

| | Lines |
|---|---|
| Total build log (before) | 2521 |
| Checkstyle warnings in it | **1267** |
| Total build log (after) | **1120** |

So it was about half the log — your instinct that it was worth a second look was right even at the corrected figure.

**The fix is to stop pinning `consoleOutput` rather than to set it false.** Leaving it at the plugin default (`false`) turned out to be strictly better than hardcoding, for a reason that is a nice echo of this PR's original bug: an explicit `<consoleOutput>` in the pom would silently beat `-Dcheckstyle.consoleOutput=true`, exactly the precedence trap that let `<configLocation>` point at `simple.xml` unnoticed. Not setting it keeps the flag usable, so reviewing the staged tier locally is still one argument away.

**Errors stay fully diagnosable** — this was the part I wanted to verify rather than assume, since suppressing output would be a bad trade if it hid failures. `logViolationsToConsole` defaults to true and is independent of `consoleOutput`, so a broken enforced rule still reports properly:

```
[ERROR] ScratchTab.java:[4,1] (whitespace) FileTabCharacter: File contains tab characters (this is the first instance).
[INFO] BUILD FAILURE
[ERROR] ... You have 1 Checkstyle violation.
```

And every violation, staged included, still lands in `target/checkstyle-result.xml`, so nothing is actually lost — SonarCloud and IDE integrations read that file rather than the console.

Documented the `-Dcheckstyle.consoleOutput=true` flag in `CLAUDE.md` next to the existing checkstyle command so contributors can find it.

Re-verified after the change: 67 modules at 0 violations, `BUILD SUCCESS`, guard still reports 13 enforced rules, all regression cases pass.

Also noted your point on `.checkstyle/java.header` — agreed, retained deliberately next to the commented `RegexpHeader` so a future headers PR has both halves ready. That ties back to the licence-header question, which is still the one open item from my side.

1PoPTRoN:
Rebased onto current `main` (`8d9be8cfd`) — the branch was 61 commits behind and the rebase was clean, no conflicts.

Re-verified on the new base, since the original measurement was taken at `4cc70df7` and roughly 60 Java files have landed on main since:

```
./mvnw checkstyle:check -Dfull   → 67 modules, 0 violations, BUILD SUCCESS
bash scripts/validate-files.sh   → "Checkstyle config ok: 13 rules enforced"
```

None of the newly-merged files violate any of the 13 enforced rules, so nothing regresses on the current base.

@jsenko — this is still on `lifecycle/new` since Aug 2, so no CI has run on it: the Verification Gate fails on *"PR is not in a testable lifecycle state"* with every test job skipped, rather than on any actual failure. @paoloantinori has reviewed twice with no outstanding issues, but the welcome comment noted that fork PRs may not sync review labels automatically. Could you `/accept`, or `/retry` to resync?

The one item still open from my side is the license-header question in the description — that's a project policy call I didn't want to make unilaterally. Happy to go either way.

paoloantinori:
Thanks @1PoPTRoN — the rebase onto `8d9be8cfd` looks clean and the re-verification (`checkstyle:check -Dfull` → 0 violations across 67 modules; `validate-files.sh` → 13 rules enforced) matches what I'd expect: none of the ~60 Java files that landed on `main` in the interim trip any of the 13 enforced rules. Nothing has changed on my side — the tiered-severity design and the promotion-direction fix from the earlier round still hold on this head.

EricWittmann:
@1PoPTRoN - license headers on source files is **not** required.  So we can disable that rule.  :+1:

EricWittmann:
Thanks @1PoPTRoN — this is excellent work. The tiered severity approach is exactly the right way to close the gap between what the docs claimed and what the build actually enforced, without forcing a 3600-violation mass reformat. Going from 2 enforced rules to 13 with a clear promotion worklist for the rest is a solid foundation. The `validate-files.sh` guard is a particularly nice touch — it prevents this class of drift from silently recurring. Great contribution!

PR #9166, opened 2026-08-02
refactor: Consolidate serdes constructor patterns with factory methods

### LINKED ISSUE #7320: Consolidate serdes constructor patterns with factory methods
## Parent Epic

## Phase
**Phase 4: Long-Term Architecture**

## Problem

Each Kafka serializer/deserializer has 5 identical constructor overloads:

**Affected Files (6 classes × 5 constructors = 30 nearly-identical constructors):**
- `serdes/kafka/avro-serde/src/main/java/io/apicurio/registry/serde/avro/AvroKafkaSerializer.java`
- `serdes/kafka/avro-serde/src/main/java/io/apicurio/registry/serde/avro/AvroKafkaDeserializer.java`
- `serdes/kafka/jsonschema-serde/src/main/java/io/apicurio/registry/serde/jsonschema/JsonSchemaKafkaSerializer.java`
- `serdes/kafka/jsonschema-serde/src/main/java/io/apicurio/registry/serde/jsonschema/JsonSchemaKafkaDeserializer.java`
- `serdes/kafka/protobuf-serde/src/main/java/io/apicurio/registry/serde/protobuf/ProtobufKafkaSerializer.java`
- `serdes/kafka/protobuf-serde/src/main/java/io/apicurio/registry/serde/protobuf/ProtobufKafkaDeserializer.java`

**Repeated Constructor Pattern:**
```java
public AvroKafkaSerializer() {
    super(new AvroSerializer<>());
}
public AvroKafkaSerializer(RegistryClientFacade clientFacade) {
    super(new AvroSerializer<>(clientFacade));
}
public AvroKafkaSerializer(SchemaResolver<Schema, Object> schemaResolver) {
    super(new AvroSerializer<>(schemaResolver));
}
public AvroKafkaSerializer(RegistryClientFacade clientFacade, SchemaResolver<Schema, Object> schemaResolver) {
    super(new AvroSerializer<>(clientFacade, schemaResolver));
}
public AvroKafkaSerializer(RegistryClientFacade clientFacade, 
        ArtifactReferenceResolverStrategy<Schema, Object> strategy,
        SchemaResolver<Schema, Object> schemaResolver) {
    super(new AvroSerializer<>(clientFacade, strategy, schemaResolver));
}
```

## Proposed Solution

Replace with factory pattern:

```java
// Factory interface
@FunctionalInterface
public interface SerializerFactory<S, D> {
    Serializer<S, D> create(RegistryClientFacade clientFacade,
            ArtifactReferenceResolverStrategy<S, D> strategy,
            SchemaResolver<S, D> schemaResolver);
}

// Base class with factory
public abstract class KafkaSerializer<S extends ParsedSchema<R>, R> 
        implements Serializer<byte[]> {
    
    private final SerializerFactory<S, R> factory;
    
    protected KafkaSerializer(SerializerFactory<S, R> factory) {
        this.factory = factory;
    }
    
    // Single configure method handles all initialization
    @Override
    public void configure(Map<String, ?> configs, boolean isKey) {
        RegistryClientFacade clientFacade = createClientFacade(configs);
        SchemaResolver<S, R> resolver = createResolver(configs);
        this.delegate = factory.create(clientFacade, getStrategy(), resolver);
    }
}

// Simplified implementation
public class AvroKafkaSerializer extends KafkaSerializer<Schema, Object> {
    
    public AvroKafkaSerializer() {
        super(AvroSerializer::new);
    }
    
    // No more constructor overloads needed!
}
```

## Additional Cleanup

Also consolidate configuration initialization patterns:
```java
// Current (duplicated in each class)
@Override
public void configure(Map<String, ?> configs, boolean isKey) {
    super.configure(configs, isKey);
    serdeHeaders = new AvroSerdeHeaders(configs, isKey);
}

// Proposed (in base class)
@Override
public final void configure(Map<String, ?> configs, boolean isKey) {
    super.configure(configs, isKey);
    initializeHeaders(configs, isKey);
}

protected abstract void initializeHeaders(Map<String, ?> configs, boolean isKey);
```

## Estimated Impact
- **Lines saved:** 100+ lines (constructor overloads)
- **Additional savings:** 60+ lines (configuration duplication)
- **Risk:** Medium (API change for advanced users)

## Acceptance Criteria
- [ ] Create `SerializerFactory` and `DeserializerFactory` interfaces
- [ ] Update base classes to use factory pattern
- [ ] Simplify all Kafka serializer/deserializer implementations
- [ ] Maintain backward compatibility for basic constructor usage
- [ ] Extract header initialization to template method
- [ ] All existing tests pass
- [ ] Update documentation for advanced usage

aaravmahajanofficial:
### Description
This PR resolves the constructor bloat, tight coupling, and delayed-initialization `NullPointerException`s across the Kafka Serde implementations by migrating to a factory pattern and property-based dependency injection.

**Key Changes:**
* **Deduplicated Constructors:** Removed 30 nearly identical overloaded constructors across `AvroKafkaSerializer/Deserializer`, `JsonSchemaKafkaSerializer/Deserializer`, and `ProtobufKafkaSerializer/Deserializer`. Concrete serdes now only expose a default no-argument constructor.
* **Factory Pattern:** Introduced `SerializerFactory` and `DeserializerFactory`.
* **Property-Based Injection:** All custom dependencies (e.g., `RegistryClientFacade`, `SchemaResolver`, `ArtifactReferenceResolverStrategy`) are now securely injected via the `configure(Map<String, ?>)` lifecycle using their respective `SerdeConfig` keys.
* **Test Suite Migration:** Refactored the test suite to align with the new map-based injection pattern, resolving `Missing registry base url` errors by ensuring `clientFacade` objects are properly mapped in test configurations.
* **Documentation:** Updated the official AsciiDoc documentation (`assembly-using-kafka-client-serdes.adoc`) to show advanced users how to inject pre-instantiated dependencies directly into the configuration map.

Closes #7320


aaravmahajanofficial:
Hi @carlesarnal, @paoloantinori, @EricWittmann. Please review whenever you have a moment. Thanks!

paoloantinori:
Hi @aaravmahajanofficial — thanks for the ping. I read through the whole diff. The direction (factory + property-based injection) is reasonable and the internal dedup is real, but this PR as written is a **hard breaking change to the public Serde API**, and I think that needs explicit maintainer sign-off before it goes further. Three things I'd want resolved:

**1. Removed public constructors = source/binary break for every downstream user.**
All six Kafka classes (`AvroKafkaSerializer/Deserializer`, `JsonSchemaKafkaSerializer/Deserializer`, `ProtobufKafkaSerializer/Deserializer`) lose every convenience constructor — `new AvroKafkaSerializer<>(clientFacade)`, `(clientFacade, schemaResolver)`, `(clientFacade, strategy, schemaResolver)`, etc. Any application or library that constructs these directly stops compiling. The PR description frames this as "constructor bloat" cleanup, but for a published library these are supported API surface. Options worth discussing with @carlesarnal / @EricWittmann: (a) keep the old constructors as thin delegators that populate the config map, deprecated, removed in the next major; or (b) explicitly accept the break and document the migration. Right now it's (b) but it isn't called out as such.

**2. `configure(...)` is now `final` on `KafkaSerializer` / `KafkaDeserializer`.**
This is the subtler break. Any downstream subclass that overrides `configure(Map, boolean)` — which was the documented extension point, and is exactly what these same classes themselves did before this PR — now gets an `IncompatibleClassChangeError` at runtime, not a compile error in their own codebase. The new `initializeHeaders(...)` hook is a fine replacement, but flipping `configure` to `final` silently breaks existing subclasses. If the goal is to guarantee ordering, consider leaving `configure` non-final and just documenting that subclasses must call `super.configure(...)`.

**3. `schemaResolver.configure(...)` is now guarded by `if (this.schemaResolver != null)`.**
In `BaseSerde.configureSchemaResolver`, the resolver used to be configured unconditionally. Now, if for any path `schemaResolver` ends up null (no config value and no default resolved — I can't see the full default branch in the truncated diff), the resolver is silently left unconfigured instead of NPE-ing loudly. Can you confirm the default-resolution path still guarantees a non-null resolver, so the guard never silently skips `schemaResolver.configure(...)` in production?

The test migration looks consistent (everything threads `REGISTRY_CLIENT_FACADE` through the config map) and the docs addition is good. The core question is really #1 and #2: is this an intentional major-version break, and if so is the `final` on `configure` justified given it breaks subclasses? I'd like a maintainer to weigh in on that before this lands.

serdes/kafka/serde-kafka-common/src/main/java/io/apicurio/registry/serde/kafka/KafkaSerializer.java
```diff
@@ -27,14 +28,19 @@ public class KafkaSerializer<T, U> implements Serializer<U> {
...
 
-    protected KafkaSerializer(AbstractSerializer<T, U> delegatedSerializer) {
-        this.delegatedSerializer = delegatedSerializer;
+    protected KafkaSerializer(SerializerFactory<T, U> factory) {
+        this.delegatedSerializer = factory.create();
     }
 
     @Override
-    public void configure(Map<String, ?> configs, boolean isKey) {
+    public final void configure(Map<String, ?> configs, boolean isKey) {
```
paoloantinori: Making `configure` final means any downstream subclass of `KafkaSerializer` (and `KafkaDeserializer`, which has the same change) that currently overrides `configure(Map, boolean)` will fail to compile. The new `initializeHeaders` hook restores the header-setup extension point, but not arbitrary configure overrides. Since these are widely-subclassed public library classes, consider whether `final` is truly necessary here (vs. keeping `configure` overridable and just invoking `initializeHeaders()`), and confirm no other subclass in the codebase overrides `configure`.

serdes/kafka/avro-serde/src/main/java/io/apicurio/registry/serde/avro/AvroKafkaSerializer.java
```diff
@@ -1,45 +1,25 @@
...
-    }
-
-    public AvroKafkaSerializer(RegistryClientFacade clientFacade, SchemaResolver<Schema, U> schemaResolver) {
-        super(new AvroSerializer<>(clientFacade, schemaResolver));
-    }
-
-    public AvroKafkaSerializer(RegistryClientFacade clientFacade, ArtifactReferenceResolverStrategy<Schema, U> strategy,
-                               SchemaResolver<Schema, U> schemaResolver) {
-        super(new AvroSerializer<>(clientFacade, strategy, schemaResolver));
+        super(AvroSerializer::new);
```
paoloantinori, line 21: All public argument-constructors were removed here (and symmetrically in the other five concrete serdes). Any user calling `new AvroKafkaSerializer<>(clientFacade)`, `(schemaResolver)`, or `(clientFacade, strategy, schemaResolver)` will break at compile and must migrate to the no-arg constructor plus `SerdeConfig.REGISTRY_CLIENT_FACADE` in the config map. The new path is documented in the adoc, but for a public API this is a source-incompatible change — a deprecation cycle or an explicit release-note callout would soften the upgrade impact.

paoloantinori (commented):
The refactor is mechanically clean and well-covered: the factory + property-injection pattern is applied consistently, all concrete serdes collapse to a no-arg constructor plus a constructor-reference factory, every visible test path injects the facade (so the `Missing registry base url` failures are resolved), and the new `IllegalStateException`/map-mutation are benign. The substantive concern is API compatibility — for a widely-used public library, removing all public argument-constructors and making `configure` final are source-incompatible changes that should be deliberate rather than incidental.

- `KafkaSerializer.java:36` (and symmetrically `KafkaDeserializer.java`): `configure(Map, boolean)` is now `final`. Any downstream subclass overriding `configure` breaks at compile; the new `initializeHeaders` hook only restores header-setup extensibility, not arbitrary overrides. Worth confirming `final` is intended vs. simply adding `initializeHeaders()` to an overridable `configure`, and that no in-repo subclass overrides it.
- `AvroKafkaSerializer.java:18` (and the other five concrete serdes): every public argument-constructor was removed. Users calling `new AvroKafkaSerializer<>(clientFacade)` / `(schemaResolver)` / `(clientFacade, strategy, schemaResolver)` must migrate to the no-arg ctor + `SerdeConfig.REGISTRY_CLIENT_FACADE` in the config map. The path is documented, but a hard removal on a public API warrants a deprecation cycle or an explicit release-note callout.
- `BaseSerde.java`: the injected facade is pushed onto the resolver via `setClientFacade(...)` *after* the resolver is created/configured. This only works if the resolver defers facade construction to first use (or reads the facade from config itself); worth confirming no resolver path eagerly demands a base URL during `configure()` before `setClientFacade` runs — otherwise `Missing registry base url` could resurface for facade-only configs. Tests cover this path so it's likely fine, but the ordering coupling is non-obvious.
- `JsonSchemaSerdeTest.java`: the deserializer config maps for `testJsonSchemaSerdeWithReferencesDeserializerDereferenced` and `testWithReferencesDeserializerDereferencedComplexUsecase` fall outside the provided diff hunks — confirm both still inject `REGISTRY_CLIENT_FACADE` (their serializer sides do); a bare `Map.of(ENABLE_HEADERS,"true")` there would fail at runtime.
- Positive: the factory references (`AvroSerializer::new`) cleanly preserve eager construction, `setClientFacade`'s `Objects.requireNonNull` is correct, and the adoc example keys match `SerdeConfig`.

serdes/kafka/serde-kafka-common/src/main/java/io/apicurio/registry/serde/kafka/KafkaSerializer.java
aaravmahajanofficial: I have removed the `final` keyword.

serdes/kafka/avro-serde/src/main/java/io/apicurio/registry/serde/avro/AvroKafkaSerializer.java
aaravmahajanofficial: Yes, this is a breaking change. Followed as it was mentioned in the `Proposed Solution` in the issue. Can have a discussion on this with @carlesarnal, @EricWittmann, and decide whether to remove the overloaded constructors or mark them as `@deprecated` and remove them in coming major release....

serdes/kafka/avro-serde/src/main/java/io/apicurio/registry/serde/avro/AvroKafkaSerializer.java
```diff
@@ -1,45 +1,25 @@
...
 import java.util.Map;
 
+import io.apicurio.registry.resolver.ParsedSchema;
+import io.apicurio.registry.serde.kafka.KafkaSerializer;
+
 public class AvroKafkaSerializer<U> extends KafkaSerializer<Schema, U> {
 
     private AvroSerdeHeaders avroHeaders;
 
     public AvroKafkaSerializer() {
```
paoloantinori, line 20: This PR removes all the overloaded public constructors (taking `RegistryClientFacade`, `SchemaResolver`, `ArtifactReferenceResolverStrategy`, etc.) from every Kafka serde, leaving only the no-arg constructor. That is a source- and binary-incompatible change for a published library — existing callers like `new AvroKafkaSerializer<>(clientFacade)` will fail to compile, and already-compiled callers hit `NoSuchMethodError`. The `configure(Map)` injection is a reasonable replacement, but consider keeping the old constructors as `@Deprecated` thin wrappers (instantiate, then delegate to `configure`) for a release, or confirm a major-version bump covers the break.

serdes/kafka/serde-kafka-common/src/main/java/io/apicurio/registry/serde/kafka/KafkaDeserializer.java
```diff
@@ -19,14 +20,19 @@ public class KafkaDeserializer<T, U> implements Deserializer<U> {
...
 
-    protected KafkaDeserializer(AbstractDeserializer<T, U> delegatedDeserializer) {
-        this.delegatedDeserializer = delegatedDeserializer;
+    protected KafkaDeserializer(DeserializerFactory<T, U> factory) {
+        this.delegatedDeserializer = factory.create();
     }
 
     @Override
-    public void configure(Map<String, ?> configs, boolean isKey) {
+    public final void configure(Map<String, ?> configs, boolean isKey) {
```
paoloantinori: `configure` is `final` here, but the sibling `KafkaSerializer.configure()` is not. Two concerns: (1) the asymmetry looks unintentional — make both `final` or neither; (2) `final` independently breaks any external subclass of `KafkaDeserializer` that overrides `configure()` (compile error today, `IncompatibleClassChangeError` for pre-compiled subclasses). If the intent is just to funnel subclasses through the new `initializeHeaders` hook, applying the same `final` to `KafkaSerializer.configure()` would at least make it consistent.

serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/BaseSerde.java
```diff
@@ -83,6 +91,15 @@ private void configureSchemaResolver(Map<String, Object> configs, boolean isKey,
                 Utils.instantiate(SchemaResolver.class, sr, this::setSchemaResolver);
             }
         }
+
+        if (this.schemaResolver == null) {
+            throw new IllegalStateException("Internal error: SchemaResolver was not initialized after configuration.");
+        }
+
+        if (this.clientFacade != null) {
+            this.schemaResolver.setClientFacade(this.clientFacade);
```
paoloantinori: Minor / hedged: if a user injects BOTH `SCHEMA_RESOLVER` (a custom resolver that may carry its own client facade) and `REGISTRY_CLIENT_FACADE`, this line unconditionally overwrites the resolver's facade after construction. That is reasonable if the explicit facade is meant to win, but it can leave a custom resolver that initialized against a different facade in a stale state. Consider guarding (e.g. only set when the resolver reports no facade yet) or documenting the precedence.

paoloantinori (commented):
The refactor cleanly deduplicates the constructor set, and the factory + `configure(Map)` injection is sound — the `clientFacade` is resolved before the default `SchemaResolver` is built, and the new fail-fast `IllegalStateException` replaces what would otherwise be a deferred NPE. Two issues are worth addressing before merge: the change is a binary/source-breaking change to the published serde API, and `configure()` is made `final` on only the deserializer side.

- **Breaking API:** All overloaded public constructors are removed from `AvroKafkaSerializer`/`Deserializer` and the JsonSchema/Protobuf peers (`AvroKafkaSerializer.java:17`), leaving only the no-arg constructor. Downstream callers using `new AvroKafkaSerializer<>(clientFacade)` fail to compile (or throw `NoSuchMethodError` against existing bytecode). Consider keeping the old constructors as `@Deprecated` thin wrappers, or confirm a major-version bump covers the break.
- **`final` asymmetry:** `KafkaDeserializer.configure()` is now `public final` (`KafkaDeserializer.java:28`) while `KafkaSerializer.configure()` is not. This independently breaks any external subclass that overrides `configure()` (compile error / `IncompatibleClassChangeError`) and looks unintentional — make both `final`, or neither.
- **Edge case (hedged):** In `BaseSerde.configureSchemaResolver` (`BaseSerde.java:100`), when a user injects BOTH a custom `SCHEMA_RESOLVER` and `REGISTRY_CLIENT_FACADE`, the unconditional `schemaResolver.setClientFacade(clientFacade)` overwrites the custom resolver's own facade after construction. Likely fine if the explicit facade is meant to win, but worth a guard or a doc note.

serdes/kafka/avro-serde/src/main/java/io/apicurio/registry/serde/avro/AvroKafkaSerializer.java
paoloantinori, line 20: This no-arg constructor is now the only public entry point — the `(RegistryClientFacade)`, `(SchemaResolver)`, and `(clientFacade, strategy, schemaResolver)` overloads were removed here and in the other five concrete serdes. That's a source-incompatible change for anyone currently calling `new AvroKafkaSerializer<>(facade)`. Since the replacement is config-map injection, could you confirm this is slated for a release that allows breaking API changes (and that a deprecation cycle isn't expected)?

serdes/kafka/serde-kafka-common/src/main/java/io/apicurio/registry/serde/kafka/KafkaDeserializer.java
```diff
@@ -19,14 +20,19 @@ public class KafkaDeserializer<T, U> implements Deserializer<U> {
     protected final AbstractDeserializer<T, U> delegatedDeserializer;
     protected HeadersHandler headersHandler;
 
-    protected KafkaDeserializer(AbstractDeserializer<T, U> delegatedDeserializer) {
-        this.delegatedDeserializer = delegatedDeserializer;
+    protected KafkaDeserializer(DeserializerFactory<T, U> factory) {
```
paoloantinori, line 23: The base constructor signature changed from `KafkaDeserializer(AbstractDeserializer<T,U>)` to `KafkaDeserializer(DeserializerFactory<T,U>)` (and likewise for `KafkaSerializer`). Any subclass outside this PR — in other modules, test utilities, or downstream — that calls `super(delegate)` with a pre-built deserializer will no longer compile. Worth grepping for all `extends Kafka(De)Serializer` and callers of these constructors to confirm the six updated classes are the only ones.

docs/modules/ROOT/pages/getting-started/assembly-using-kafka-client-serdes.adoc
```diff
@@ -405,3 +405,43 @@ KStream<String, LogInput> input = builder.stream(
     Consumed.with(Serdes.String(), logSerde)
 );
 ----
+
+[id="registry-serdes-advanced-config_{context}"]
+== Advanced Serde configuration: Injecting dependencies
+
+:_mod-docs-content-type: PROCEDURE
+
+[role="_abstract"]
```
paoloantinori, line 414: This section documents the new config-map injection well, but doesn't mention that the previous convenience constructors (e.g. `new AvroKafkaSerializer<>(clientFacade)`) have been removed. A one-line note that those overloads are gone — and that callers should move to `SerdeConfig.REGISTRY_CLIENT_FACADE` in the config map — would make the migration path explicit for existing users.

paoloantinori (commented):
This refactor is internally consistent — the factory delegation and the ordering of the new `REGISTRY_CLIENT_FACADE` injection in `BaseSerde.configureSchemaResolver` (inject facade, then resolve the schema resolver, then `setClientFacade`) correctly sidesteps the 'Missing registry base url' path, and the `initializeHeaders` hook preserves the old configure/header-init ordering. The main concern is compatibility: the PR removes public constructors and changes a protected base-class constructor signature, which is a source-incompatible change worth a compatibility sweep before merge.

- Removed public constructors (`(RegistryClientFacade)`, `(SchemaResolver)`, `(clientFacade, strategy, schemaResolver)`) across all six concrete serdes (e.g. `AvroKafkaSerializer.java:17`) — existing callers like `new AvroKafkaSerializer<>(facade)` will no longer compile. Worth confirming this targets a release that permits breaking API changes and that a deprecation cycle isn't expected.
- `KafkaSerializer`/`KafkaDeserializer`'s protected constructor now takes a factory instead of a pre-built delegate (`KafkaDeserializer.java:23`); any subclass outside this PR (other modules, test helpers, downstream) calling `super(delegate)` will break — worth grepping all subclasses and callers in-repo to confirm the six updated classes are the only ones.
- The migration docs show the new injection style but don't call out that the old convenience constructors are gone (`assembly-using-kafka-client-serdes.adoc:414`); a one-line migration note would help users moving off `new AvroKafkaSerializer<>(facade)`.
- Minor: `SerdeConfig.REGISTRY_CLIENT_FACADE` advertises injection by fully-qualified classname, but `RegistryClientFacade` implementations are typically constructed with a client argument (tests use `new RegistryClientFacadeImpl(isolatedClientV3)`), so the classname form may fail at configure — consider scoping the doc to instance injection.

serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/BaseSerde.java
aaravmahajanofficial: Thanks! good catch. I have updated `configureSchemaResolver` so when `apicurio.registry.client-facade` is explicitly set in the config map, it takes precedence over any facade already on a custom `SCHEMA_RESOLVER` (same semantics as `BaseSerde(clientFacade, schemaResolver))`. If only `SCHEMA_RESOLVER` is provided with its own facade, that facade is preserved....

paoloantinori:
Thanks — the updated precedence in `configureSchemaResolver` looks right. When `apicurio.registry.client-facade` is explicitly set in the config map, it overrides any facade already present on a custom `SCHEMA_RESOLVER` (same semantics as the old `BaseSerde(clientFacade, schemaResolver)` constructor), and the `log.warn` on override is the correct signal. The `IllegalStateException` guard for an uninitialized resolver is a reasonable safety net.

One minor note: the override check uses reference inequality (`resolverFacade != this.clientFacade`), which is the right choice here since we only want to warn when a *different* instance is being replaced — equality by identity is exactly what we mean.

From my side this addresses the concern I raised. I'll leave it to the maintainers (@carlesarnal, @EricWittmann) for the final merge call.

serdes/kafka/serde-kafka-common/src/main/java/io/apicurio/registry/serde/kafka/KafkaSerializer.java
```diff
@@ -27,14 +28,19 @@ public class KafkaSerializer<T, U> implements Serializer<U> {
 
     protected HeadersHandler headersHandler;
```
paoloantinori, line 30: Changing the protected base constructor from `KafkaSerializer(AbstractSerializer<T,U>)` to `KafkaSerializer(SerializerFactory<T,U>)` is an internal signature change that will break any `KafkaSerializer` subclass that calls `super(...)` with a pre-built instance, as well as any code still constructing the format serdes via the removed constructors. This PR updates only the Avro/JsonSchema/Protobuf peers and four test files — worth grepping the whole repo (other serde modules, integration/system tests, examples) for remaining `extends KafkaSerializer`/`KafkaDeserializer` declarations and `new *KafkaSerializer/Deserializer<>(...)` calls to confirm nothing is missed.

serdes/kafka/avro-serde/src/main/java/io/apicurio/registry/serde/avro/AvroKafkaSerializer.java
paoloantinori, line 20: All public convenience constructors (`(RegistryClientFacade)`, `(SchemaResolver)`, `(clientFacade, schemaResolver)`, `(clientFacade, strategy, schemaResolver)`) are removed here, leaving only the no-arg constructor. That is a source-incompatible change for downstream users of this public API — confirm it is intended to land in a major version and that the migration (inject via `SerdeConfig.REGISTRY_CLIENT_FACADE` in the config map) is captured in release notes, since the docs NOTE only describes the new pattern without a deprecation path.

paoloantinori (commented):
Overall this is a clean, behavior-preserving refactor: the overloaded constructors collapse neatly into a factory + `configure(Map)` injection pattern, the test migration is consistent, and the AsciiDoc docs accurately describe the new flow. The two things worth confirming before merge are the breadth of the public-API breakage and that every call site in the tree has been migrated.

- **Public-API breakage**: `AvroKafkaSerializer/Deserializer` and the JsonSchema/Protobuf peers drop every convenience constructor (`(clientFacade)`, `(schemaResolver)`, `(clientFacade, schemaResolver)`, etc.) — see `AvroKafkaSerializer.java:17`. This is source-incompatible for any downstream caller; confirm it's targeted at a major version and that release notes flag it (the docs NOTE at `assembly-using-kafka-client-serdes.adoc:411` mentions the removal but offers no deprecation path).
- **Verify all call sites**: the protected base constructor signature changed from `(AbstractSerializer)` to `(SerializerFactory)` at `KafkaSerializer.java:30` (and the deserializer peer). Any `KafkaSerializer`/`KafkaDeserializer` subclass or `new XSerializer<>(...)` caller not shown in this PR — other serde modules, integration/system tests, examples — will fail to compile. The diff only touches the three format peers and four test files, so a full-tree grep for remaining usages is worth doing.
- Minor/easy win: the `log.warn` in `BaseSerde.java` hardcodes the config-key strings (`apicurio.registry.client-facade`, `apicurio.registry.schema-resolver`) instead of referencing the `SerdeConfig` constants; they will silently drift if those keys are ever renamed.

serdes/kafka/serde-kafka-common/src/main/java/io/apicurio/registry/serde/kafka/KafkaSerializer.java
aaravmahajanofficial: Checked, nothing is missed.

serdes/kafka/avro-serde/src/main/java/io/apicurio/registry/serde/avro/AvroKafkaDeserializer.java
```diff
@@ -1,43 +1,22 @@
...
 
 import java.util.Map;
 
+import io.apicurio.registry.serde.kafka.KafkaDeserializer;
+
 public class AvroKafkaDeserializer<U> extends KafkaDeserializer<Schema, U> {
 
     private AvroSerdeHeaders avroHeaders;
 
     public AvroKafkaDeserializer() {
```
paoloantinori, line 17: This removes the public `AvroKafkaDeserializer(RegistryClientFacade)`, `(SchemaResolver)`, `(clientFacade, schemaResolver)`, and `(clientFacade, strategy, schemaResolver)` constructors — and the equivalent ones on the other five concrete serdes. These were public API, so downstream code that constructs serdes with a facade/resolver/strategy directly will fail to compile. If a hard removal is intended, an `@Deprecated` transition release (or at least a prominent release-notes/migration note) would soften the upgrade; otherwise consider keeping thin delegating constructors that forward into the config map.

docs/modules/ROOT/pages/getting-started/assembly-using-kafka-client-serdes.adoc
```diff
@@ -405,3 +405,48 @@ KStream<String, LogInput> input = builder.stream(
...
+[source,java,subs="+quotes,attributes"]
+----
+Map<String, Object> props = new HashMap<>();
+// Pass standard Kafka properties
+props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9092");
+props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class);
+props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, AvroKafkaSerializer.class);
+
+// Inject the custom instances using the standard SerdeConfig keys
+props.put(SerdeConfig.REGISTRY_CLIENT_FACADE, myCustomFacade);
```
paoloantinori, line 449: The example injects a pre-instantiated `ArtifactReferenceResolverStrategy` via `SerdeConfig.ARTIFACT_RESOLVER_STRATEGY`, and the section states all three dependencies can be injected as instances. This PR only adds instance-handling for `REGISTRY_CLIENT_FACADE` (in `BaseSerde.configureSchemaResolver`), and every test in the PR passes the strategy as `.class` or `.getName()` — never a raw instance. Because the removed `new AvroKafkaSerializer(clientFacade, strategy, schemaResolver)` constructor accepted a stateful strategy instance, please verify the config-map path resolves a raw instance (not only a class name) and add a test for it, or scope this doc to class-name injection only.

paoloantinori (commented):
Overall the consolidation is internally consistent — the factory-method + `initializeHeaders` hook preserves the prior configure-time ordering, and the test migration to map-based facade injection is thorough and symmetric across every serializer/deserializer pair. Two concerns stand out: a public API break that affects downstream users, and a documented injection capability this PR never tests.

- **Public API break (MAJOR):** All six concrete serdes lose their public `(clientFacade)`, `(schemaResolver)`, and multi-arg constructors — e.g. `AvroKafkaDeserializer.java:14` and the matching classes for the Avro/JsonSchema/Protobuf serializer & deserializer. Existing code such as `new AvroKafkaDeserializer<>(clientFacade)` will no longer compile. The base `KafkaDeserializer` (`KafkaDeserializer.java:23`) and `KafkaSerializer` constructor also changed from `AbstractSerializer` to a `SerializerFactory`, which is source/binary-incompatible for any external subclass of the base classes. For a library serde this warrants either an `@Deprecated` transition release or a prominent release-notes/migration entry.
- **Untested documented capability (MAJOR if true):** The new docs (`assembly-using-kafka-client-serdes.adoc:445`) inject a pre-instantiated `ArtifactReferenceResolverStrategy` via the config map, but this PR adds instance-injection handling only for `REGISTRY_CLIENT_FACADE` in `BaseSerde.configureSchemaResolver` (`BaseSerde.java:92`); every test passes the strategy as `.class`/`.getName()`, never a raw instance. Since the removed constructors accepted stateful strategy instances, the config-map path needs verification + coverage, or the docs should be scoped to class-name injection.
- **Behavioral framing:** Both the old `super(new AvroSerializer<>())` and the new `factory.create()` build the delegated serde eagerly in the constructor, so the 'eliminates delayed-initialization NPE' motivation is really constructor dedup rather than an init-timing change — not a regression, just worth aligning the PR description.

docs/modules/ROOT/pages/getting-started/assembly-using-kafka-client-serdes.adoc
aaravmahajanofficial: Thanks, good catch. I have added a test for stateful strategy, and updated the docs as well.

aaravmahajanofficial:
@EricWittmann  / @carlesarnal can you please clarify whether we should add `@deprecated` on the overloaded constructors, and remove in the next major release, or straight away remove these as done in this PR.

aaravmahajanofficial:
@EricWittmann A gentle bump to this.

https://github.com/Apicurio/apicurio-registry/pull/9166#issuecomment-5170263621

Thanks!

EricWittmann (commented):
Thanks for the PR, @aaravmahajanofficial! The refactoring approach is clean and the factory + config-map injection pattern is applied consistently across all 6 serde classes. The new `initializeHeaders()` template method is a nice separation of concerns, and the test migration is thorough. A few things we need to discuss/address before this can move forward:

## Breaking API change — deprecation vs. removal

This is the main blocker. All public constructors like `new AvroKafkaSerializer<>(clientFacade)` are removed outright, which is source- and binary-incompatible for downstream users. The issue's acceptance criteria explicitly states "Maintain backward compatibility for basic constructor usage."

We need a maintainer decision here: if the next release is a major version bump, removing outright may be acceptable. Otherwise, the old constructors should be marked `@Deprecated` for one release cycle before removal. @carlesarnal — can you weigh in on the intended approach?

Similarly, the `protected` constructor signature on `KafkaSerializer`/`KafkaDeserializer` changed from accepting `AbstractSerializer` to `SerializerFactory`. Any external subclass calling `super(new MySerializer<>())` will break. Worth grepping for `extends KafkaSerializer` and `extends KafkaDeserializer` outside this module to assess impact.

## `BaseSerde` constructors left intact

The PR adds new client facade resolution logic to `configureSchemaResolver` in `BaseSerde`, but leaves all the old `BaseSerde` constructors (accepting `RegistryClientFacade`, `SchemaResolver`, `ArtifactReferenceResolverStrategy`) intact. Since the Kafka serde classes now all use the no-arg constructor, these are dead code from that perspective. Check whether `AbstractSerializer`/`AbstractDeserializer` still use them — if not, they should be deprecated/removed as part of this PR to keep the API surface consistent.

## Factory interfaces could be simpler

The factory interfaces (`SerializerFactory`, `DeserializerFactory`) are no-arg `create()` methods — functionally equivalent to `Supplier<AbstractSerializer<T,U>>`. Consider whether a dedicated interface is justified or whether `Supplier` (or a type alias) would be cleaner. Not a blocker, but worth considering.

## Minor items

- The comment `// Empty by default, designed to be overridden by subclasses if needed` on `initializeHeaders()` in both `KafkaSerializer` and `KafkaDeserializer` is unnecessary — the method's purpose is clear from context. Remove the comments.
- The new `IllegalStateException("Internal error: SchemaResolver was not initialized after configuration.")` in `BaseSerde.configureSchemaResolver` changes the failure mode from a later NPE to an early explicit check. This is arguably better, but make sure it doesn't break any existing setup that worked by coincidence (e.g. custom subclasses that set `schemaResolver` after `configure()`).

## Commit hygiene

The branch has 32 commits, roughly half of which are merge-from-main. These will be squashed on merge, so no action needed, just noting it.

Overall this is solid work — the core pattern is right. The main question is the API compatibility strategy, which needs a maintainer call.

carlesarnal:
## Code Review

Thanks for the continued work, @aaravmahajanofficial. The factory pattern is elegant, the test migration is thorough, and the new `testAvroStatefulStrategyInstanceInjection` test covers the stateful-strategy scenario well.

Items needing attention (some from @EricWittmann review):

### 1. `initializeHeaders()` comments should be removed
@EricWittmann asked for this. The `// Empty by default` comments are still in `KafkaSerializer.java:43` and `KafkaDeserializer.java:34`. Please remove.

### 2. `BaseSerde` old constructors - dead code check
@EricWittmann asked to verify if old `BaseSerde` constructors are dead code. Since all six Kafka serde classes use no-arg, check if `AbstractSerializer`/`AbstractDeserializer` still call them. If not, mark `@Deprecated`.

### 3. Hardcoded config keys in `log.warn` - `BaseSerde.java:103`
Hardcodes literal strings instead of `SerdeConfig.REGISTRY_CLIENT_FACADE` and `SerdeConfig.SCHEMA_RESOLVER` constants. These would silently drift if constants are renamed.

### 4. DCO sign-off
Only 1/32 commits has `Signed-off-by:`. Ensure the squash commit includes it.

### 5. Breaking API - awaiting maintainer decision
Main blocker per @EricWittmann. Issue #7320 says Maintain backward compatibility. Suggest preparing `@Deprecated` wrappers as fallback.

---

Everything else looks solid: `configure()` is non-final, `IllegalStateException` guard is good, facade override logic is reasonable, and docs clearly explain the new pattern.

aaravmahajanofficial:
@carlesarnal Thanks for the review. Have addressed the comments and added the wrappers back, and also have marked them as `@deprecated`.

serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/BaseSerde.java
```diff
@@ -83,6 +84,32 @@ private void configureSchemaResolver(Map<String, Object> configs, boolean isKey,
                 Utils.instantiate(SchemaResolver.class, sr, this::setSchemaResolver);
             }
         }
+
+        if (this.schemaResolver == null) {
+            throw new IllegalStateException("Internal error: SchemaResolver was not initialized after configuration.");
+        }
+
+        boolean explicitFacade = configs.containsKey(SerdeConfig.REGISTRY_CLIENT_FACADE);
```
paoloantinori, line 92: Does `Utils.instantiate` accept a pre-instantiated `RegistryClientFacade` instance, or only a `Class`/classname string? Everywhere else in the codebase this helper is used for class-name instantiation, and the docs added in this PR explicitly say instance-only. If it does not short-circuit on an instanceof match, this call will fail at runtime for exactly the usage all the new tests exercise. Please either verify the helper handles instances or replace with a plain `instanceof RegistryClientFacade` cast and a clear error for other value types (e.g. a String classname, which the SerdeConfig javadoc says will fail).

serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/BaseSerde.java
```diff
@@ -83,6 +84,32 @@ private void configureSchemaResolver(Map<String, Object> configs, boolean isKey,
...
+
+        if (this.clientFacade == null && explicitFacade) {
+            Object cf = configs.get(SerdeConfig.REGISTRY_CLIENT_FACADE);
+            if (cf != null) {
+                Utils.instantiate(RegistryClientFacade.class, cf, this::setClientFacade);
+            }
+        }
```
paoloantinori, line 99: Ordering question: is `schemaResolver.configure(...)` called before or after this facade-propagation block? If the resolver's configure still tries to build its own client from properties when no facade is set, a config containing `REGISTRY_CLIENT_FACADE` but no registry URL could still trigger the `Missing registry base url` failure the PR description mentions. Please add a test for facade-injected config with no URL properties to pin the ordering.

serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/config/SerdeConfig.java
```diff
@@ -194,6 +194,14 @@ public IdOption useIdOption() {
      */
     public static final String CONTRACT_RULES_MIGRATION_TARGET_VERSION = "apicurio.registry.serde.contract-rules.migration.target-version";
 
+    /**
+     * A pre-instantiated {@link io.apicurio.registry.resolver.client.RegistryClientFacade} used by
+     * the Serde classes to communicate with the registry. Pass an instance, not a class name:
+     * typical implementations (including {@code RegistryClientFacadeImpl}) have no no-arg
+     * constructor, so classname instantiation will fail.
```
paoloantinori, line 201: The new key `apicurio.registry.client-facade` breaks the `apicurio.registry.serde.*` prefix convention used by the neighboring keys in this file (`apicurio.registry.serde.contract-rules...` etc.). Since the key is introduced in this PR and has no back-compat burden, consider `apicurio.registry.serde.registry-client-facade` (or matching whatever `SCHEMA_RESOLVER` uses) so serde config stays uniform in config dumps and docs.

serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/BaseSerde.java
```diff
@@ -83,6 +84,32 @@ private void configureSchemaResolver(Map<String, Object> configs, boolean isKey,
...
+
+        if (this.clientFacade != null) {
+            RegistryClientFacade resolverFacade = this.schemaResolver.getClientFacade();
+            if (explicitFacade) {
+                if (resolverFacade != null && resolverFacade != this.clientFacade) {
+                    log.warn(SerdeConfig.REGISTRY_CLIENT_FACADE + " overrides facade already set " + SerdeConfig.SCHEMA_RESOLVER);
```
paoloantinori, line 105: When the user supplied both a custom `SCHEMA_RESOLVER` holding its own facade and `REGISTRY_CLIENT_FACADE`, this silently overrides the resolver's facade with only a warn log. That is a genuine config conflict (two explicit, disagreeing inputs); failing fast here, or at least documenting the precedence rule in assembly-using-kafka-client-serdes.adoc, would save users a confusing debugging session.

paoloantinori (commented):
The direction is right: collapsing the constructor overloads into map-based injection removes real duplication, and the deprecation path plus regression tests is a sensible migration story. The risk is concentrated in `BaseSerde.configureSchemaResolver`, where the new facade-injection block relies on `Utils.instantiate` accepting a pre-instantiated object and on ordering relative to `schemaResolver.configure`, neither of which is proven in this diff. The new `SerdeConfig.REGISTRY_CLIENT_FACADE` key also breaks the established `apicurio.registry.serde.*` naming convention.
- `serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/BaseSerde.java:92` (new block): `Utils.instantiate(RegistryClientFacade.class, cf, this::setClientFacade)` is documented everywhere else in this PR as a class-name/Object instantiator. If it only handles `Class` or `String` values, passing the facade instance itself will fail (or double-construct) at runtime. The tests only exercise the happy path, so this is unverified. Please confirm `Utils.instantiate` short-circuits on an instance of the target type, or just do an `instanceof` cast here.
- `BaseSerde.java:95-110`: ordering between this block and `schemaResolver.configure(...)` is not shown in the diff. If the resolver's own configure creates a facade when none is set (the previous behavior for `Missing registry base url`), setting the facade after (or before) that call changes which one wins. Worth an explicit test: config with `REGISTRY_CLIENT_FACADE` and no registry URL must not attempt to build a client from properties.
- `BaseSerde.java:104-107`: when both `REGISTRY_CLIENT_FACADE` and a custom `SCHEMA_RESOLVER` that already holds a different facade are configured, the code logs a warn and silently replaces the resolver's facade. That is a configuration conflict the user explicitly created twice; failing fast (or at least documenting the precedence in the adoc) would be kinder than a log line on a path that is usually only read after data has gone somewhere unexpected.
- `serdes/generic/serde-common/src/main/java/io/apicurio/registry/serde/config/SerdeConfig.java:201`: every other serde key is prefixed `apicurio.registry.serde.` (e.g. `apicurio.registry.serde.schema-resolver`-style). `apicurio.registry.client-facade` breaks that convention and will look odd next to the rest of the keys in config dumps. Since the key is new in this PR, now is the only cheap time to align it.
- `serdes/kafka/serde-kafka-common/.../KafkaDeserializer.java:39` and `KafkaSerializer.java:47`: `initializeHeaders(configs, isKey)` is now invoked at the end of `configure`. Any third-party subclass that still overrides `configure(Map, boolean)` without calling `super.configure()` will never get its headers initialized, and the previously-public extension point (`configure` override in the concrete serdes) quietly became a protected `initializeHeaders`. Consider javadoc on `KafkaSerializer`/`KafkaDeserializer` calling out the extension-point change.

serdes/kafka/serde-kafka-common/src/main/java/io/apicurio/registry/serde/kafka/KafkaDeserializer.java
```diff
@@ -27,8 +36,11 @@ protected KafkaDeserializer(AbstractDeserializer<T, U> delegatedDeserializer) {
     public void configure(Map<String, ?> configs, boolean isKey) {
         delegatedDeserializer.configure(new SerdeConfig(configs), isKey);
         this.configure(new BaseKafkaSerDeConfig(configs), isKey);
+        initializeHeaders(configs, isKey);
```
paoloantinori, line 39: The extension point moved: subclasses that previously overrode `configure(Map, boolean)` to install their headers (as `AvroKafkaDeserializer` did before this PR) now must override `initializeHeaders`, and any external subclass overriding `configure` without calling `super.configure()` will silently skip header init. A short javadoc note on `initializeHeaders` stating it replaces the old configure-override pattern would help downstream users migrating.

carlesarnal:
Closing this as stale. It's been a week with no changes.

PR #9175, opened 2026-08-02
fix(ui): clear pageError state when route parameters change

### LINKED ISSUE #9172: [UI] "Return to previous page" link on Resource Page updates URL but fails to refresh the view
Registry Version: 3.3.2-SNAPSHOT (or snapshot) Persistence type: sql

When navigating to a referenced artifact that is not registered, the UI correctly displays the "Resource not found" page. However, clicking the "Return to previous page" link on this screen only updates the browser URL (in the address bar) but fails to actually reload the previous page's content. The UI remains stuck on the "Resource not found" screen until the user manually refreshes the browser.

This happens because the pageError state in the main page components (e.g., VersionPage, ArtifactPage, GroupPage, BranchPage) is not cleared when the route parameters change.

Environment
Deployment: Apicurio Registry backend running in Docker (apicurio/apicurio-registry:latest-snapshot).
Frontend: Running the UI development server locally (npm run dev in ui/ui-app).
Browser: Google Chrome.

Video clip - 
https://github.com/user-attachments/assets/cee0ee7c-8857-46ca-8030-6b74dd39e534

Steps to Reproduce
1. Go to an artifact version page that has references.
2. Open the Graph view under the References tab.
3. Click on a reference node in the graph to open the Node Details panel on the left.
4. Click the link View this artifact's references at the bottom of the panel. Since that referenced artifact does not exist in the registry, the page correctly shows the Resource not found page.
5. Click the blue link Return to previous page on the screen.
6. Observe that the browser's URL address bar changes back to /explore/folderName/artifactName/versions/versionName/references, but the UI remains stuck on the "Resource not found" screen.

Expected vs Actual Behaviour

Expected: Clicking "Return to previous page" should navigate back to the previous valid page, update the URL, and automatically reload/refresh the UI view.

Actual: The browser's URL changes back, but the UI remains stuck on the "Resource not found" screen until a manual browser refresh is performed.

JituRewar:
Close #9172 

### What this PR does

This PR ensures that the page-level `pageError` state variable is reset to `undefined` whenever the route parameters change within the dynamic UI page components:
- `VersionPage`
- `ArtifactPage`
- `GroupPage`
- `BranchPage`

### Why this is needed
Currently, if a user encounters a `404` or other load error (for example, navigating to an unregistered reference artifact and landing on the "Resource not found" screen), the `pageError` state is set. 

When the user attempts to navigate back to the previous page (e.g., by clicking the "Return to previous page" link), the browser URL in the address bar updates, but the UI component view itself remains permanently stuck on the error page. This happens because the stale `pageError` state is never cleared on parameter/route updates. Adding `setPageError(undefined)` at the beginning of the loading `useEffect` hook resolves this issue, allowing the UI to properly reload and render the previous valid page.

### Steps to Verify
1. Navigate to a valid artifact version page that has references (`/explore/<group>/<artifact>/versions/<version>/references`).
2. Open the References Graph tab.
3. Click on a reference node that is not registered to open the detail panel and click "View this artifact's references".
4 Click "Return to previous page".
5. Verify that the UI now successfully transitions back to the previous graph view automatically without requiring a manual browser refresh.

### Files Modified
* `ui/ui-app/src/app/pages/version/VersionPage.tsx`
* `ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx`
* `ui/ui-app/src/app/pages/group/GroupPage.tsx`
* `ui/ui-app/src/app/pages/branch/BranchPage.tsx`

Before Video Clip
[Screencast From 2026-08-02 22-54-12.webm](https://github.com/user-attachments/assets/d46a0791-f7ea-4dcf-a226-e0e0f450e93e)

After Video Clip
[Screencast From 2026-08-02 23-26-52.webm](https://github.com/user-attachments/assets/b98cc334-e7ed-41b2-86a7-d36d00358915)


paoloantinori (commented):
This is a minimal, correct fix: clearing stale `pageError` at the start of the loading `useEffect` lets each page re-render its loading/content state when route params change, instead of staying stuck on the error screen. The change is small, idiomatic, and applied consistently across all four page components, so I have no blocking concerns.

- **Co-located reset is the right placement.** Putting `setPageError(undefined)` at the top of the load effect — before `newLoaderGuard()` — in `VersionPage.tsx:407`, `ArtifactPage.tsx:384`, `GroupPage.tsx:273`, and `BranchPage.tsx:121` correctly couples the reset to the load it gates, so no separate reset effect is needed. On initial mount it is a harmless no-op (React bails on identical state), and on a failed reload the new error is re-set by the load itself, so there is no stale-state or race defect.
- **One assumption worth eyeballing (not visible in the diff):** for this to fire on navigation, each load effect must re-run when route params change — i.e. the params need to be in the effect's dependency array. The before/after clips and the effect's role of recreating loaders via `newLoaderGuard()` strongly imply they do, but the dependency arrays aren't in the hunks, so a quick confirm there would lock it in.
- **Completeness (hedged):** worth confirming whether any sibling page components that share this same `LoaderGuard` + `pageError` pattern have the same latent stale-error bug and were left out of this PR.

JituRewar:
Thanks for the review, @paoloantinori ! 

To confirm both points:

1. Dependency Arrays: Yes, the route parameters are indeed included in the dependency arrays of the `useEffect` load hooks for all modified components, ensuring they trigger on navigation:
  `VersionPage.tsx`: `[groupId, artifactId, version]`
  `ArtifactPage.tsx`: `[groupId, artifactId]`
  `GroupPage.tsx`: `[groupId]`
  `BranchPage.tsx`: `[groupId, artifactId, branchId]`

2. Completeness: `LoaderGuard` is used in exactly 5 files across the entire codebase. The 4 modified in this PR, plus `EditorPage.tsx`. `EditorPage.tsx` already has the `setPageError(undefined)` reset logic implemented at the start of its `useEffect` (line 230). So this coverage is 100% complete and there are no other pages using this pattern with a latent stale-error bug.

paoloantinori:
Thanks for confirming both points — that fully addresses my review. The dependency arrays covering the route params on each page (`VersionPage`, `ArtifactPage`, `GroupPage`, `BranchPage`) are exactly what makes the `setPageError(undefined)` reset reliable on navigation, and consistent `LoaderGuard` usage keeps the cancel path intact. No further concerns from me.

EricWittmann:
Thanks @JituRewar — nice catch and thorough verification (confirming all 5 LoaderGuard usages and the dependency arrays). Clean fix. Merged!

PR #9181, opened 2026-08-03
fix(mcp): validate inputSchema type removals and property data type changes

### LINKED ISSUE #9178: bug(mcp): McpToolCompatibilityChecker misses parameter type changes and schema-type removal due to lenient null guard
## Description

While auditing the MCP tool compatibility checking flow, I noticed
@malladinagarjuna2 just flagged a related bug in `AgentCardCompatibilityChecker`
(#9173) — a compatibility checker silently passing a real breaking change
because of an incomplete null guard. After checking whether the same pattern
exists in `McpToolCompatibilityChecker`, it does.

### Environment
environment-specific — this is a pure logic bug in the compatibility-checking
code path, reproducible on any persistence backend.

### Steps to Reproduce
1. Register an MCP_TOOL artifact with `inputSchema.properties.age.type = "integer"`
2. Register a new version of the same artifact where `age.type` is changed to
   `"string"` — no other changes
3. Run compatibility checking (`FULL` or `BACKWARD` rule) against the two versions
4. Observe: no `INPUT_SCHEMA_TYPE_CHANGED` (or equivalent) difference is reported

A second, related case:
1. Register an MCP_TOOL artifact with a top-level `inputSchema.type = "object"`
2. Register a new version where `inputSchema.type` is removed entirely from the
   schema
3. Run compatibility checking
4. Observe: the removal is not flagged, because `checkInputSchemaTypeChange`
   only compares `existingType` and `proposedType` when **both** are non-null

**Expected:** Both cases above should be reported as compatibility differences —
a consumer of the tool relying on the original `age: integer` contract, or the
original top-level schema type, would break silently.

**Actual:**

In `McpToolCompatibilityChecker.java`, lines 58–84, `checkInputSchemaTypeChange`
guards on:

    if (existingType != null && proposedType != null && !existingType.equals(proposedType))

When `proposedType` is `null` (schema type removed or unparseable), the
`proposedType != null` clause short-circuits the whole check — the removal is
never flagged.

Separately, the checker only inspects **property removals**
(`checkPropertyRemovals`) and **required-flag changes**
(`checkRequiredParamAdditions` / `checkRequiredParamRemovals`). It never
compares the `type` field of properties that exist in both versions, so a type
change on an existing property (e.g. `age: integer` → `age: string`) is
silently treated as compatible.

### Logs
N/A — this is a logic/correctness gap, not a runtime exception. No error is
thrown; the compatibility check simply returns a false "compatible" result.

### Proposed Fix
1. In `checkInputSchemaTypeChange`, flag a difference whenever
   `existingType != null` and `proposedType` is either missing or different —
   not only when both are present and unequal.
2. Add a `checkPropertyTypeChanges` helper that walks matching properties under
   `inputSchema.properties` in both versions and flags a difference whenever the
   `type` field changes on a property present in both.

Happy to submit the PR with tests covering: type removed entirely, type changed
on an existing property, and the existing "no change" case staying green.

Sudarshan25092007:
### Summary
Closes #9178 
This PR fixes two compatibility checking gaps in `McpToolCompatibilityChecker`:
1. **Schema Type Removal**: Updating `checkInputSchemaTypeChange` to flag a compatibility difference (`INPUT_SCHEMA_TYPE_CHANGED`) when `inputSchema.type` is removed in a proposed version (`existingType != null && proposedType == null`). Previously, the `proposedType != null` guard caused type removals to pass silently.
2. **Property Data Type Mismatch**: Added `checkPropertyTypeChanges` to compare the `type` field of matching parameters in `inputSchema.properties` across versions (e.g., changing property `age` from `integer` to `string`).

### Changes
- Updated `checkInputSchemaTypeChange` guard condition in `McpToolCompatibilityChecker.java`.
- Added helper `checkPropertyTypeChanges` and `extractPropertyTypes` to inspect child parameter types under `inputSchema.properties`. Replaced deprecated Jackson `fields()` with `properties()`.
- Added 5 comprehensive test cases in `McpToolCompatibilityCheckerTest.java` covering:
  - `inputSchema.type` removal (incompatible)
  - `inputSchema.type` unchanged (compatible)
  - Property data type change (incompatible)
  - Property data type unchanged (compatible)
  - Existing property removal regression check (incompatible)

### Verification
- `./mvnw test -pl schema-util/util-provider -am -Dtest=McpToolCompatibilityCheckerTest`: **Passed 14/14 tests cleanly** (0 failures, 0 errors).
- `./mvnw checkstyle:check -pl schema-util/common,schema-util/util-provider`: **0 Checkstyle violations**.

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
@@ -4,9 +4,13 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.apicurio.registry.content.TypedContent;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -24,6 +28,8 @@ public class McpToolCompatibilityChecker
         extends AbstractCompatibilityChecker<McpToolCompatibilityDifference> {
 
     private static final ObjectMapper mapper = new ObjectMapper();
+    private static final String INPUT_PROPERTY_PREFIX = "Input property '";
+    private static final String WAS_REMOVED_SUFFIX = "' was removed";
 
     @Override
     protected Set<McpToolCompatibilityDifference> isBackwardsCompatibleWith(String existing,
@@ -39,6 +45,9 @@ public class McpToolCompatibilityChecker
 
             // Check removed properties
             checkPropertyRemovals(existingNode, proposedNode, differences);
+
+            // Check property type changes
+            checkPropertyTypeChanges(existingNode, proposedNode, differences);
 
             // Check added required parameters
             checkRequiredParamAdditions(existingNode, proposedNode, differences);
@@ -60,11 +69,13 @@ public class McpToolCompatibilityChecker
         String existingType = getInputSchemaType(existing);
         String proposedType = getInputSchemaType(proposed);
 
-        if (existingType != null && proposedType != null && !existingType.equals(proposedType)) {
+        if (existingType != null && (proposedType == null || !existingType.equals(proposedType))) {
+            String message = proposedType == null
+                    ? "inputSchema type '" + existingType + WAS_REMOVED_SUFFIX
+                    : "inputSchema type changed from '" + existingType + "' to '" + proposedType + "'";
             differences.add(new McpToolCompatibilityDifference(
                     McpToolCompatibilityDifference.Type.INPUT_SCHEMA_TYPE_CHANGED,
-                    "inputSchema type changed from '" + existingType + "' to '" + proposedType
-                            + "'"));
+                    message));
         }
     }
 
@@ -77,7 +88,35 @@ public class McpToolCompatibilityChecker
             if (!proposedProps.contains(prop)) {
                 differences.add(new McpToolCompatibilityDifference(
                         McpToolCompatibilityDifference.Type.PROPERTY_REMOVED,
-                        "Input property '" + prop + "' was removed"));
+                        INPUT_PROPERTY_PREFIX + prop + WAS_REMOVED_SUFFIX));
+            }
+        }
+    }
+
+    private void checkPropertyTypeChanges(JsonNode existing, JsonNode proposed,
+            Set<McpToolCompatibilityDifference> differences) {
+        Set<String> existingProps = extractPropertyNames(existing);
+        Set<String> proposedProps = extractPropertyNames(proposed);
+
+        for (String prop : existingProps) {
+            if (proposedProps.contains(prop)) {
+                String existingType = getPropertyType(existing, prop);
+                String proposedType = getPropertyType(proposed, prop);
+
+                if (!Objects.equals(existingType, proposedType)) {
+                    String message;
+                    if (existingType != null && proposedType == null) {
+                        message = INPUT_PROPERTY_PREFIX + prop + "' type '" + existingType + WAS_REMOVED_SUFFIX;
+                    } else if (existingType == null && proposedType != null) {
+                        message = INPUT_PROPERTY_PREFIX + prop + "' type constraint '" + proposedType + "' was added";
+                    } else {
+                        message = INPUT_PROPERTY_PREFIX + prop + "' type changed from '" + existingType
+                                + "' to '" + proposedType + "'";
+                    }
+                    differences.add(new McpToolCompatibilityDifference(
+                            McpToolCompatibilityDifference.Type.PROPERTY_TYPE_CHANGED,
+                            message));
+                }
             }
         }
     }
@@ -105,7 +144,7 @@ public class McpToolCompatibilityChecker
             if (!proposedRequired.contains(param)) {
                 differences.add(new McpToolCompatibilityDifference(
                         McpToolCompatibilityDifference.Type.REQUIRED_PARAM_REMOVED,
-                        "Required parameter '" + param + "' was removed"));
+                        "Required parameter '" + param + WAS_REMOVED_SUFFIX));
             }
         }
     }
@@ -113,27 +152,68 @@ public class McpToolCompatibilityChecker
     private String getInputSchemaType(JsonNode node) {
         JsonNode inputSchema = node.get("inputSchema");
         if (inputSchema != null && inputSchema.isObject()) {
-            JsonNode type = inputSchema.get("type");
-            if (type != null && type.isTextual()) {
-                return type.asText();
-            }
+            return extractTypeString(inputSchema.get("type"));
         }
         return null;
     }
 
     private Set<String> extractPropertyNames(JsonNode node) {
         Set<String> properties = new HashSet<>();
+        JsonNode props = getPropertiesNode(node);
+        if (props != null) {
+            Iterator<String> fieldNames = props.fieldNames();
+            while (fieldNames.hasNext()) {
+                properties.add(fieldNames.next());
+            }
+        }
+        return properties;
+    }
+
+    private String getPropertyType(JsonNode node, String propName) {
+        JsonNode props = getPropertiesNode(node);
+        if (props != null) {
+            JsonNode propNode = props.get(propName);
+            if (propNode != null && propNode.isObject()) {
+                return extractTypeString(propNode.get("type"));
+            }
+        }
+        return null;
+    }
+
+    private JsonNode getPropertiesNode(JsonNode node) {
         JsonNode inputSchema = node.get("inputSchema");
         if (inputSchema != null && inputSchema.isObject()) {
             JsonNode props = inputSchema.get("properties");
             if (props != null && props.isObject()) {
-                Iterator<String> fieldNames = props.fieldNames();
-                while (fieldNames.hasNext()) {
-                    properties.add(fieldNames.next());
-                }
+                return props;
             }
         }
-        return properties;
+        return null;
+    }
+
+    private String extractTypeString(JsonNode typeNode) {
+        if (typeNode == null) {
+            return null;
+        }
+        if (typeNode.isTextual()) {
+            return typeNode.asText();
+        }
+        if (typeNode.isArray()) {
+            List<String> types = new ArrayList<>();
+            for (JsonNode item : typeNode) {
+                if (item.isTextual()) {
+                    types.add(item.asText());
+                }
+            }
+            if (!types.isEmpty()) {
+                if (types.size() == 1) {
+                    return types.get(0);
+                }
+                Collections.sort(types);
+                return types.toString();
+            }
+        }
+        return null;
     }
 
     private Set<String> extractRequiredParams(JsonNode node) {
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityDifference.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityDifference.java
@@ -13,6 +13,7 @@ public class McpToolCompatibilityDifference implements CompatibilityDifference {
         REQUIRED_PARAM_ADDED("inputSchema/required"),
         REQUIRED_PARAM_REMOVED("inputSchema/required"),
         INPUT_SCHEMA_TYPE_CHANGED("inputSchema/type"),
+        PROPERTY_TYPE_CHANGED("inputSchema/properties"),
         PROPERTY_REMOVED("inputSchema/properties"),
         PARSE_ERROR("document");
 
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
@@ -344,4 +344,344 @@ class McpToolCompatibilityCheckerTest {
 
         assertTrue(result.isCompatible(), "Identical schema with description change should be fully compatible");
     }
+
+    @Test
+    void testBackwardIncompatibleRemovingInputSchemaType() {
+        String existing = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "query": { "type": "string" }
+                        }
+                    }
+                }
+                """;
+
+        String proposed = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "properties": {
+                            "query": { "type": "string" }
+                        }
+                    }
+                }
+                """;
+
+        CompatibilityExecutionResult result = checker.testCompatibility(
+                CompatibilityLevel.BACKWARD, List.of(createMcpTool(existing)),
+                createMcpTool(proposed), Map.of());
+
+        assertFalse(result.isCompatible(),
+                "Removing inputSchema type should be backward incompatible");
+    }
+
+    @Test
+    void testBackwardCompatibleUnchangedInputSchemaType() {
+        String existing = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "query": { "type": "string" }
+                        }
+                    }
+                }
+                """;
+
+        String proposed = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "query": { "type": "string" }
+                        }
+                    }
+                }
+                """;
+
+        CompatibilityExecutionResult result = checker.testCompatibility(
+                CompatibilityLevel.BACKWARD, List.of(createMcpTool(existing)),
+                createMcpTool(proposed), Map.of());
+
+        assertTrue(result.isCompatible(),
+                "Unchanged inputSchema type should be backward compatible");
+    }
+
+    @Test
+    void testBackwardIncompatibleChangingPropertyType() {
+        String existing = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "age": { "type": "integer" }
+                        }
+                    }
+                }
+                """;
+
+        String proposed = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "age": { "type": "string" }
+                        }
+                    }
+                }
+                """;
+
+        CompatibilityExecutionResult result = checker.testCompatibility(
+                CompatibilityLevel.BACKWARD, List.of(createMcpTool(existing)),
+                createMcpTool(proposed), Map.of());
+
+        assertFalse(result.isCompatible(),
+                "Changing a property type from integer to string should be backward incompatible");
+    }
+
+    @Test
+    void testBackwardCompatibleUnchangedPropertyType() {
+        String existing = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "age": { "type": "integer" }
+                        }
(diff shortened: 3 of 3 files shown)


schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -82,6 +86,26 @@ private void checkPropertyRemovals(JsonNode existing, JsonNode proposed,
...
 
+    private void checkPropertyTypeChanges(JsonNode existing, JsonNode proposed,
+            Set<McpToolCompatibilityDifference> differences) {
+        Map<String, String> existingTypes = extractPropertyTypes(existing);
+        Map<String, String> proposedTypes = extractPropertyTypes(proposed);
+
+        for (Map.Entry<String, String> entry : existingTypes.entrySet()) {
+            String propName = entry.getKey();
+            String existingType = entry.getValue();
+            if (proposedTypes.containsKey(propName)) {
```
paoloantinori: Coverage gap to confirm: `extractPropertyTypes` (line ~172) only retains properties that have a textual `type`, so `proposedTypes.containsKey(propName)` here skips a property that is kept but whose `type` was dropped. That change is caught by neither `checkPropertyTypeChanges` nor `checkPropertyRemovals` (the property name still exists), so it passes as compatible. If dropping a property's type should be treated as breaking — the same case this PR now catches at the `inputSchema.type` level — this branch needs to handle the "property present, type absent in proposed" case rather than relying on `containsKey`.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -136,6 +160,43 @@ private Set<String> extractPropertyNames(JsonNode node) {
...
+                return props;
+            }
+        }
+        return null;
+    }
+
+    private String extractSinglePropertyType(JsonNode propNode) {
+        if (propNode != null && propNode.isObject()) {
+            JsonNode typeNode = propNode.get("type");
+            if (typeNode != null && typeNode.isTextual()) {
```
paoloantinori: `extractSinglePropertyType` returns a type only when `type` is textual, so JSON-Schema union types such as `"type": ["string", "null"]` are dropped here and in `extractPropertyTypes`. As a result, a change involving an array-typed property (e.g. `integer` → `["string","null"]`) won't be detected. If MCP tool schemas in practice use union types, consider normalizing the array/single forms; otherwise worth a short comment marking this as an accepted limitation.

paoloantinori (commented):
Both fixes are logically correct and the five new tests cover the stated cases well; no blocking defects found. A couple of edge cases in the new property-type comparison are worth confirming before merge.

- `McpToolCompatibilityChecker.java:67` — the rewritten guard `(proposedType == null || !existingType.equals(proposedType))` correctly flags `inputSchema.type` removal. Minor: when the type is removed, the message at line 70 renders as "changed from 'object' to 'null'"; a removal-specific wording would read more clearly.
- `McpToolCompatibilityChecker.java:97` — `checkPropertyTypeChanges` only compares a property that survives in `proposedTypes`, which excludes properties retained without a `type` (dropped by `extractPropertyTypes` at line 172). Such a property is flagged by neither this check nor `checkPropertyRemovals` (the property name still exists), so removing a property's `type` while keeping the property passes as compatible — the same class of gap this PR closes at the top level.
- `McpToolCompatibilityChecker.java:193` — `extractSinglePropertyType` accepts only a textual `type`; JSON-Schema union types (e.g. `"type": ["string","null"]`) are dropped, so changes to/from array-typed properties go undetected. Likely shared by the existing `getInputSchemaType`, so low priority.
- `McpToolCompatibilityChecker.java:99` — the `existingType != null && proposedType != null` guard is effectively dead since `extractPropertyTypes` only retains typed entries; harmless, noting for clarity.

Sudarshan25092007:
Thanks for pointing out the edge cases via your thorough review @paoloantinori! 

I've updated the PR to address all your feedback:

1. **Property Type Removal Handling**: Refactored `checkPropertyTypeChanges` to compare existing vs proposed property types across all retained properties (`existingProps` ∩ `proposedProps`). If a property is kept but its `type` attribute is dropped (`existingType != null && proposedType == null`), it is now properly flagged as backward incompatible (`INPUT_SCHEMA_TYPE_CHANGED`).
2. **JSON Schema Union Types Support**: Introduced `extractTypeString` helper method that handles both single string types (`"type": "string"`) and array union types (`"type": ["string", "null"]`). Array union types are normalized into string representations (e.g., `"[string, null]"`), allowing type changes to/from array union types to be detected and flagged as backward incompatible.
3. **Cognitive Complexity & Cleanup**: Decomposed helper routines (`getPropertiesNode`, `extractTypeString`, `getPropertyType`) to keep Cognitive Complexity well below SonarCloud limits (max complexity ≤ 5 per method).
4. **New Unit Tests**: Added test cases covering property type removal (`testBackwardIncompatibleRemovingPropertyType`) and union array type changes (`testBackwardIncompatibleChangingPropertyTypeToUnionArray`). All 16 tests in `McpToolCompatibilityCheckerTest` pass cleanly.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -113,27 +138,64 @@ private void checkRequiredParamRemovals(JsonNode existing, JsonNode proposed,
...
+        }
+        if (typeNode.isArray()) {
+            List<String> types = new ArrayList<>();
+            for (JsonNode item : typeNode) {
+                if (item.isTextual()) {
+                    types.add(item.asText());
                 }
             }
+            if (!types.isEmpty()) {
+                return types.toString();
```
paoloantinori, line 213: Union types are compared order-sensitively here. `types.toString()` preserves insertion order, so an existing schema with `"type": ["string", "number"]` and a proposed schema with `"type": ["number", "string"]` produce different strings and are flagged `INPUT_SCHEMA_TYPE_CHANGED`, even though the union is semantically identical (member order is not significant in JSON Schema). This yields a false-positive incompatibility for both the top-level check and `checkPropertyTypeChanges`. Consider sorting the list, or comparing as a set, before producing the comparison key.

paoloantinori (commented):
The change correctly implements its stated goals — flagging `inputSchema.type` removal and per-property `type` changes — with comprehensive new tests and consistent null handling throughout the helper chain. The one functional issue worth addressing is that union `type` arrays (e.g. `["string", "null"]`) are compared via order-sensitive `List.toString()`, so reordering the members of an otherwise-identical union is incorrectly reported as an incompatible type change.

- **Union types compared order-sensitively (false positive)** — `extractTypeString` (McpToolCompatibilityChecker.java, ~line 195) collects array members into an `ArrayList` and returns `types.toString()`; since insertion order is preserved, an existing `"type": ["string","number"]` vs a proposed `"type": ["number","string"]` yields different strings and is flagged `INPUT_SCHEMA_TYPE_CHANGED` even though the union is semantically identical (order is not significant in JSON Schema). This affects both the top-level check (`checkInputSchemaTypeChange`) and `checkPropertyTypeChanges`; sorting/normalizing before stringifying, or comparing as a set, would fix it.
- **`INPUT_SCHEMA_TYPE_CHANGED` reused for property-level changes** — `checkPropertyTypeChanges` (McpToolCompatibilityChecker.java, ~line 102) emits the same difference `Type` as the top-level inputSchema check, distinguishing them only via the human-readable message string. If any consumer branches on the `Type` enum rather than the description, it cannot separate a top-level type change from a property type change; a dedicated value (e.g. `INPUT_PROPERTY_TYPE_CHANGED`) would be clearer. Flagged as a design suggestion, not a correctness bug, since severity depends on how these difference types are consumed (not visible in this diff).
- **Adding a type to a previously-untyped property is not flagged** — both guards short-circuit when `existingType == null`, so a property going from `{}` to `{"type":"integer"}` (which narrows accepted values and can break existing callers) is not reported. This is outside the PR's stated scope but may warrant a follow-up given the checker's backward-compatibility intent.

Sudarshan25092007:
Thank you for the follow-up feedback @paoloantinori

Updated the PR to address all 3 points:
1. **Sorted Union Types**: Array union types are now sorted before comparison (`Collections.sort`), so reordering union members (e.g. `["string", "number"]` vs `["number", "string"]`) evaluates as compatible.
2. **Distinct Enum**: Added `PROPERTY_TYPE_CHANGED` to `McpToolCompatibilityDifference.Type` to cleanly separate property-level type changes from top-level schema type changes.
3. **Untyped Property Type Additions**: Included the handling for untyped properties receiving a type constraint in this PR so all property-level type transitions are fully covered.

paoloantinori:
Thanks for the thorough follow-up — all three points are addressed cleanly:

1. **Sorted union types**: `Collections.sort` in `extractTypeString` makes `["string","number"]` and `["number","string"]` compare equal, and `testBackwardCompatibleReorderedUnionTypeArray` pins that behavior.
2. **Distinct enum value**: Splitting `PROPERTY_TYPE_CHANGED` from `INPUT_SCHEMA_TYPE_CHANGED` keeps the difference messages unambiguous and the call sites in `checkInputSchemaTypeChange` vs `checkPropertyTypeChanges` consistent.
3. **Untyped property transitions**: Using `Objects.equals(existingType, proposedType)` correctly treats `null`↔`"integer"` as a change, and `testBackwardIncompatibleAddingTypeToUntypedProperty` plus `testBackwardIncompatibleRemovingPropertyType` lock in both directions.

LGTM — the null handling, the new helper extraction (`getPropertiesNode`, `extractTypeString`), and the regression test for `PROPERTY_REMOVED` all look right. Good test coverage on the edge cases.

Sudarshan25092007:
Thanks for your valuable feedback 😊 @paoloantinori 
Looks like the workflow run is waiting on maintainer approval before the remaining checks can run — happy to wait, just flagging in case it got missed in the queue.

paoloantinori:
Thanks for the heads-up. The workflow approval queue is on the Apicurio maintainers' side, not something you can unblock from the PR — no action needed from you. From the review perspective, all three follow-up points (sorted union types, property type removal handling, and the required-parameter regression coverage) are addressed cleanly, so I have nothing further pending on the code itself.

Sudarshan25092007:
Just pinging here for further reviews/comments on this @carlesarnal / @vandanayadav7  !

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -113,27 +140,65 @@ private void checkRequiredParamRemovals(JsonNode existing, JsonNode proposed,
...
+                    types.add(item.asText());
                 }
             }
+            if (!types.isEmpty()) {
+                Collections.sort(types);
+                return types.toString();
+            }
         }
-        return properties;
+        return null;
```
paoloantinori, line 216: A single-element union type such as `'type': ['string']` normalizes here to the string `[string]`, while the scalar `'type': 'string'` normalizes to `string`. JSON Schema treats those as identical, so an equivalent schema rewrite would be flagged as a spurious `PROPERTY_TYPE_CHANGED` / `INPUT_SCHEMA_TYPE_CHANGED`. Consider unwrapping a one-element list to its scalar (e.g. `return types.size() == 1 ? types.get(0) : types.toString();`) so the two forms compare equal.

paoloantinori (commented):
Overall this is a correct, well-tested change: the type-removal guard now catches a null proposed type, property-type comparison is null-safe, and the 9 new tests cover union reordering and add/remove-type cases. No correctness or safety issues; two minor edge cases worth a look.

- **Single-element union arrays normalize inconsistently** — `extractTypeString` (~L199-201) turns `'type': ['string']` into the literal `[string]` while a scalar `'type': 'string'` becomes `string`, so two semantically-identical JSON Schema forms are flagged as `PROPERTY_TYPE_CHANGED`/`INPUT_SCHEMA_TYPE_CHANGED` and a benign rewrite gets rejected as a breaking change. Easy fix: unwrap a one-element list to its scalar.
- **Relaxing a property type is flagged incompatible** — `checkPropertyTypeChanges` (~L91-92, test `testBackwardIncompatibleRemovingPropertyType`) reports `integer` -> untyped `{}` as a difference; removing a constraint is usually backward-compatible, so confirm this conservative stance matches the intended MCP compatibility semantics.
- **Type-removal message renders `to 'null'`** — at the guard (L72) the difference message concatenates the Java `null` when `inputSchema.type` is removed; a dedicated 'type was removed' branch would read more clearly.

Sudarshan25092007:
Thanks @paoloantinori, I've updated the PR with these fixes:

1. **Single-Element Union Lists**: `extractTypeString` now unwraps single-element list forms (`if (types.size() == 1) return types.get(0);`), so `"type": "string"` and `"type": ["string"]` normalize identically and evaluate as compatible.
2. **Clearer Difference Wording**: Updated difference messages when a type is removed so they render as `"inputSchema type 'object' was removed"` (or `"Input property 'x' type 'integer' was removed"`) instead of concatenating Java `null`.
3. **Conservative Stance on Property Type Modifications**: Retained the conservative stance flagging type modifications (including removing type constraints) to ensure strict contract compatibility for tool consumers.
4. **SonarCloud issue `java:S1192`**: Resolved!

All 19 tests in `McpToolCompatibilityCheckerTest` pass cleanly.

paoloantinori:
Thanks for the follow-up. All three points are addressed cleanly at this head:

- `extractTypeString` unwraps single-element union lists (line 209), so `"string"` and `["string"]` normalize identically.
- `Collections.sort` (line 212) makes union-member order irrelevant.
- The removed-type messages now read as "was removed" for both `inputSchema.type` and per-property `type` (lines 74, 109).

The new test coverage (single-element union equivalence, adding-a-type-to-untyped, union-array property change) closes the gaps. No further concerns from me.

Sudarshan25092007:
Ready for your review whenever you have bandwidth @jsenko!

Sudarshan25092007:
Rebased with main, ready for your review @jsenko!

jsenko (commented):
Fixes #9178 — validates inputSchema type removals and property data type changes in the MCP tool compatibility checker.

## Summary

The two fixes are correct: type removal detection (`existingType != null && proposedType == null`) and per-property type checking via `checkPropertyTypeChanges`. The iterative improvements across commits (union sorting, single-element unwrapping, untyped property handling) addressed all edge cases flagged in prior automated reviews. No must-fix issues found.

- **Union type handling** is solid after the fixes: sorted for order-insensitive comparison, single-element arrays unwrapped to scalar form. Both top-level `inputSchema.type` and property-level types benefit.
- **Adding type constraint to untyped property** is correctly flagged as incompatible — narrowing accepted values can break existing callers.
- **`PROPERTY_TYPE_CHANGED`** as a dedicated enum value (separate from `INPUT_SCHEMA_TYPE_CHANGED`) is the right design.

### should-fix: Unrelated changes in the diff

The branch includes unrelated changes from main (113 files, pom.xml dependency additions across many modules, CLI, operator, UI changes). The PR's own commits only touch 3 files (`McpToolCompatibilityChecker.java`, `McpToolCompatibilityDifference.java`, `McpToolCompatibilityCheckerTest.java`). The branch should be rebased to main to produce a clean diff — per contributor guidelines, PRs should contain no unrelated changes.

### should-fix: Commit messages

Several commit messages don't follow conventional commit format: "Resolved the issue of Cognitive Complexity", "list is now sorted before returning...". Since rebase-merge is used, each commit lands individually on main. These should be squashed or reworded before merge.

Sudarshan25092007:
Thanks @jsenko ! 

Cleaned up the branch and addressed both points:
- Rebased directly onto the latest `main` to eliminate all unrelated file diffs. The PR diff now strictly touches only the 3 relevant MCP checker files.
- Squashed the branch history into a single conventional commit: `fix(mcp): validate inputSchema type removals and property data type changes`.

All 19 unit tests pass cleanly.

jsenko (changes requested):
## Review

Fixes #9178 — the two bugs (type removal passing silently, property type changes unchecked) are real and correctly identified.

### Change in direction

While reviewing this PR, I looked at how the existing JSON Schema compatibility checker (`JsonSchemaDiffLibrary` in `schema-util/json`) handles the same cases. It turns out the `inputSchema` field in an MCP tool definition is a standard JSON Schema, and we already have a mature checker that handles all the cases this PR addresses — plus many more:

- Type changes with semantic understanding (e.g., `integer` → `number` is widening = compatible, not just a string mismatch)
- Nested schema comparison (recursive, not just top-level properties)
- `additionalProperties`, `enum`, `oneOf`/`anyOf` evolution
- Constraint changes (`minLength`, `maxLength`, `pattern`, `minimum`, `maximum`, etc.)
- Required property changes with proper set-diff reporting

The current `McpToolCompatibilityChecker` reimplements a shallow version of this — it compares `type` as a string and only looks at first-level property types. This means it has false positives (flags `integer` → `number` as incompatible) and misses deeper structural changes.

Rather than continuing to build out a parallel compatibility checker for `inputSchema`, the right approach is to delegate `inputSchema` comparison to the existing everit-based `JsonSchemaDiffLibrary` and keep only the MCP-tool-envelope checks in `McpToolCompatibilityChecker` (name, description, annotations — things that aren't JSON Schema).

### must-fix: Delegate inputSchema compatibility to JsonSchemaDiffLibrary

Replace the `inputSchema`-level checks (`checkInputSchemaTypeChange`, `checkPropertyRemovals`, `checkPropertyTypeChanges`, `checkRequiredParamAdditions`, `checkRequiredParamRemovals`) with a call to `JsonSchemaDiffLibrary.findDifferences()` on the extracted `inputSchema` sub-trees.

Use the existing everit-based checker (not the experimental apitomy-based one).

**How to proceed — your choice:**

- **Option A (preferred):** Create a new issue for the refactoring, then open a separate PR that both refactors the checker to delegate to `JsonSchemaDiffLibrary` and fixes #9178. This keeps the change focused and makes the review cleaner.
- **Option B:** Continue working on this PR and rework it here.

Either way is fine — pick whichever you're more comfortable with.

**Implementation sketch:**

1. Extract the `inputSchema` `JsonNode` from both existing and proposed tool definitions
2. Serialize them back to strings
3. Call `JsonSchemaDiffLibrary.findDifferences(existingSchema, proposedSchema, direction)` — this returns a `DiffContext` with a set of `Difference` objects
4. Filter for incompatible differences (`!difference.getDiffType().isBackwardsCompatible()`)
5. Convert each to a `McpToolCompatibilityDifference` (map the `DiffType.getDescription()` to your message format)
6. Keep the envelope-level checks (name/description/annotations) as-is — those aren't part of `inputSchema`

The existing 14 tests should continue to pass since they assert on compatibility outcomes, not internal implementation details. If any test expects a specific message string, update it to match the `JsonSchemaDiffLibrary` output.

Note: `JsonSchemaDiffLibrary` uses the everit `SchemaLoader` which doesn't support Draft 2020-12 (see PR #9104). For standard MCP tool schemas this is fine — they typically use Draft-07 or no explicit `$schema`.

### optional: Replace string-constant concatenation with String.format

`INPUT_PROPERTY_PREFIX` and `WAS_REMOVED_SUFFIX` fragment natural sentences across constant boundaries, making each call site harder to read than the message itself. Neither is used outside this class. Prefer `String.format("Input property '%s' was removed", prop)` at each site. This becomes moot if `inputSchema` checks are delegated.

Sudarshan25092007:
Thanks @jsenko, this makes sense, and I appreciate you tracing it back to the existing JsonSchemaDiffLibrary rather than letting a shallow reimplementation land. I'll go with Option A: filing a new issue for the delegation refactor, then opening a fresh PR that both implements it and closes #9178. I'll close this PR once the new one is up. Also noting the DryRunTest failure looks unrelated to this diff (flagged as flaky by CI) — let me know if you'd like me to look at that separately.

jsenko:
Great! Yeah, the test failure is unrelated, and I think there's already a PR fixing it https://github.com/Apicurio/apicurio-registry/pull/9287 . So feel free to create a new issue and open a new PR.

jsenko:

Sudarshan25092007:
Closing in favor of #9368 which implements the delegation approach per #9301 and maintainer feedback.

PR #9185, opened 2026-08-03
docs: rewrite the CLI guide to match the current acr command surface

eameh:


vandanayadav7 (approved):
Verified all documented commands, aliases, and options against the CLI source on main — everything is accurate. LGTM.

PR #9186, opened 2026-08-03
docs: add a GitOps storage configuration guide

eameh:


eameh:
Hi @jsenko, just following up on this PR review

eameh:
Hi @jsenko @carlesarnal, gentle nudge on this one when you have a moment. It adds the GitOps storage configuration guide, and since GitOps is one of the headline 3.3 features, it would be lovely to have the docs land with the release. Happy to adjust anything, especially the sync container and validation sections where you know the behavior best. Thanks!

eameh:
Hi @jsenko @carlesarnal — since everyone's stretched thin at the moment, I ran this guide through a full independent SME and QE pass to make your review easier. Three separate reviewers with no shared assumptions:

* **Adversarial source verification**: every claim checked against current main. ~82 claims confirmed — including all 36 property-table rows, the full management API section, and every status-code claim — and 6 refuted, all now fixed in the latest commit.
* **Hands-on QE**: a live GitOps deployment following the guide literally. It caught the big one — the example set was missing the required `group-v0` file, so a user following the guide verbatim got a registry stuck in ERROR. The examples and data-format section now include and explain the group requirement, and QE confirmed the corrected fixture loads (plus verified the 409 read-only behavior, sync 204, readiness semantics, change pickup, the 503 when validation is disabled, and the mixed-config startup rejection all match the guide).
* **Drift check**: all 178 commits since the branch base — nothing the guide depends on changed; endpoints and defaults are byte-identical, and the merge is conflict-free.

Notable fixes from the review: the content-extension table now describes what the code actually does (media type only — artifact type comes from `artifactType` or content detection), the read-only mount advice now carves out dry-run validation (which needs a writable workspace), the strict-mode SSH wording matches the entrypoint behavior, and I removed the registry-side validate fetch timeout property from the table since nothing consumes it (the sync container variable is the effective one).

One genuine product bug fell out of the QE run: the experimental features gate never actually enforces (`ExperimentalFeaturesConfig.validate()` is unreachable) — filed as #9630. The guide no longer claims startup fails without the gate.

leben + vale still 0 errors / 0 warnings, downstream render still clean. I think this is in good shape now — review whenever suits, and happy to adjust anything. Thanks!

PR #9191, opened 2026-08-03
fix(rest): clamp pagination parameters for public agents

### LINKED ISSUE #9189: Negative pagination parameters cause HTTP 500 in /.well-known/agents/public
## Describe the bug

The unauthenticated `/.well-known/agents/public` endpoint does not validate negative pagination parameters.

When a negative `offset` or `limit` is supplied, the values are forwarded directly to the storage layer, resulting in an HTTP 500 response containing internal database exception details instead of returning a client error or safely normalizing the values.

Other pagination endpoints in the same area already clamp or validate these parameters, making this endpoint inconsistent.

## Steps to Reproduce

1. Start Apicurio Registry with A2A enabled.
2. Verify the endpoint works normally:

```bash
curl "http://localhost:8099/.well-known/agents/public"
```

3. Send a request with a negative pagination parameter:

```bash
curl "http://localhost:8099/.well-known/agents/public?offset=-1"
```

Alternatively:

```bash
curl "http://localhost:8099/.well-known/agents/public?limit=-1"
```

## Expected Behavior

The endpoint should reject invalid pagination parameters (for example with a `400 Bad Request`) or normalize them consistently with the other pagination endpoints.

The request should never result in an internal server error or expose database exception details.

## Actual Behavior

The endpoint responds with **HTTP 500**, and the response contains internal SQL/JDBC exception information caused by the negative pagination values reaching the storage layer.

## Investigation

The issue appears to originate from the pagination parameters being forwarded without validation in the public agents endpoint, while sibling pagination endpoints already clamp these values before executing the query.

This makes the behavior inconsistent with the rest of the Registry pagination APIs.

Adii-45:
## Summary

Clamps pagination parameters in the public agents endpoint before forwarding them to the storage layer..

Previously, negative `offset` or `limit` values supplied to `/well-known/agents/public` were passed directly to `storage.searchArtifacts(...)`, allowing invalid values to reach the database and resulting in an HTTP 500 response. This change applies the same normalization already used by the other pagination endpoints in `WellKnownResourceImpl`!

Fixes #9189

## Changes

- Clamp `offset` to a non-negative value
- Clamp `limit` to the supported range before invoking `storage.searchArtifacts(...)`
- Reuse the existing pagination normalization approach already present in the same class for consistency
- Add regression tests covering:
  - negative `offset`
  - negative `limit`
  - the `/apis/registry/v3/well-known/agents/public` alias endpoint

## Testing

- Added regression tests for the affected endpoint
- Verified all existing tests continue to pass
- Confirmed requests with negative pagination values no longer return HTTP 500 and instead follow the same behavior as the other pagination endpoints


app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -137,8 +137,11 @@ public AgentSearchResults getPublicAgents(Integer offset, Integer limit) {
         filters.add(SearchFilter.ofArtifactType(ArtifactType.AGENT_CARD));
         filters.add(SearchFilter.ofLabel("apicurio.agent.visibility", "public"));
 
+        int safeOffset = Math.max(0, offset);
```
paoloantinori, line 140: `offset` and `limit` are boxed `Integer` query params, and `Math.max` / `Math.min` auto-unbox them here — a request that omits both params would raise a `NullPointerException` (→ 500) unless they carry `@DefaultValue`. The summary notes this mirrors the other pagination endpoints in the same class, which strongly suggests `@DefaultValue` is already present; if so this is a non-issue. Worth a quick confirm that the params declare `@DefaultValue` so the no-param request still works.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
```diff
@@ -443,6 +443,43 @@ public void testGetPublicAgentsWithPagination() throws Exception {
...
+                .body("agents", notNullValue());
+    }
+
+    @Test
+    public void testGetPublicAgentsNegativeOffsetLimitViaV3Path() {
+        givenAtRoot()
+                .when()
+                .queryParam("offset", -1)
+                .queryParam("limit", -1)
+                .get("/apis/registry/v3/well-known/agents/public")
```
paoloantinori, line 476: This test requests `/apis/registry/v3/well-known/agents/public` (no leading dot), but the PR summary describes the v3 alias as `/apis/registry/v3/.well-known/agents/public` (with the dot, matching the root `/.well-known/agents/public`). If the actual `@Path` uses `.well-known`, this request wouldn't hit the alias and wouldn't exercise the clamp on that route. Worth confirming the path matches the registered route — if the dot is correct, update the test URL; if not, the summary wording is slightly off.

paoloantinori (commented):
Overall a sound, low-risk fix: the clamp logic is correct and consistently normalizes negative pagination values before they reach storage. I have no blocker-level concerns; two quick confirms and one test-quality note below.

- **Null-safety of the unbox** (`WellKnownResourceImpl.java:140-141`): `Math.max(0, offset)` / `Math.min(limit, 500)` auto-unbox the boxed `Integer` params, so omitting the query params would NPE unless they carry `@DefaultValue`. Since this mirrors sibling endpoints in the same class, `@DefaultValue` is most likely present — worth confirming.
- **v3 alias path string** (`WellKnownResourceTest.java:476`): the test uses `/apis/registry/v3/well-known/agents/public` (no dot) while the summary describes `/apis/registry/v3/.well-known/agents/public` (dot, matching the root `/.well-known/`). If the registered route uses the dot, this test isn't actually hitting the alias.
- **Tests are smoke-only**: the three new cases assert only `200` + non-null fields, so they guard against the original 500 but don't assert clamped behavior (e.g., `offset=-1` ≡ `offset=0`). Acceptable for this regression; consider a stronger assertion later.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
Adii-45: Thanks for pointing this out! I verified the endpoint signature and both parameters are annotated with `@DefaultValue`, so RESTEasy always supplies a value before the method is entered. Invalid query parameters are rejected during parameter binding, and the sibling endpoints in the same class follow the same auto-unboxing pattern. I confirmed there isn't a reachable `null` path here, so I left the implementation unchanged.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
Adii-45: Thanks for catching that! I verified the registered v3 route is `/apis/registry/v3/well-known/agents/public`, which is the path exercised by the regression test. The discrepancy was only a typo in the PR description (`/.well-known/...`). I've updated the description to match the actual registered route.

Adii-45:
Hey @EricWittmann, all Cl checks are passing and PR is ready for review.. Could you please take a look whenever you have a chance!

EricWittmann:
Thanks for the fix, @Adii-45! Clean and consistent with the existing pagination handling. Merged.

PR #9192, opened 2026-08-03
fix(rest): remove deprecated custom date format configuration

### LINKED ISSUE #7329: Remove deprecated custom date format configuration
## Summary

Custom date format configuration properties were deprecated in v3.1.2 due to SDK incompatibility and should be removed.

## Location

\`common/src/main/java/io/apicurio/registry/rest/config/JacksonDateTimeCustomizer.java\`

## Deprecated Properties

| Property | Line | Issue |
|----------|------|-------|
| \`apicurio.apis.date-format\` | 32 | Breaks SDK compatibility |
| \`apicurio.apis.date-format-timezone\` | 42 | Breaks SDK compatibility |

## Problem

From the code documentation:
> "Customizing the date format causes incompatibility with generated client SDKs"

The code includes a large warning message (lines 53-71) that logs when custom formats are used, explaining the incompatibility issue.

## Current Behavior

- Default: ISO-8601 format (SDK compatible)
- Custom: Any user-defined format (breaks SDKs)
- Warning message logs if custom format is configured
- Related issue: #6799

## Proposed Solution

1. Remove the deprecated configuration properties
2. Remove the custom format handling code
3. Enforce ISO-8601 format only
4. Update documentation

## Target Version

v4.0.0 (breaking change)

## Risk Assessment

**Low Risk** - Properties have been deprecated since v3.1.2, cause SDK incompatibility anyway, and have warning messages in place.

## Acceptance Criteria

- [ ] Remove deprecated property definitions
- [ ] Remove custom format handling code
- [ ] Update documentation
- [ ] Add migration notes to release notes
- [ ] All tests pass

Saloni3494:
Summary: This PR removes the deprecated custom date format configuration properties (apicurio.apis.date-format and apicurio.apis.date-format-timezone) which were deprecated in v3.1.2 due to SDK incompatibility. The registry will now strictly enforce the default ISO-8601 date-time format.

Changes:
Removed the deprecated configuration properties and warning logs from JacksonDateTimeCustomizer.java.
Removed the legacy configurations from application.properties and documentation files.
Deleted LegacyV2ApiDateFormatTest.java since the ability to customize date formats is no longer supported.
Fixes #7329


app/src/main/java/io/apicurio/registry/rest/config/JacksonDateTimeCustomizer.java
```diff
@@ -23,70 +19,19 @@ public class JacksonDateTimeCustomizer implements ObjectMapperCustomizer {
...
      */
     @Override
     public void customize(ObjectMapper mapper) {
-        try {
-            configureDateFormat(mapper, dateFormat, timezone);
-        } catch (Exception e) {
-            log.error("Error setting REST API date format.", e);
-            configureDateFormat(mapper, DEFAULT_DATE_TIME_FORMAT, DEFAULT_DATE_TIME_FORMAT_TZ);
-        }
+        configureDateFormat(mapper, DEFAULT_DATE_TIME_FORMAT, DEFAULT_DATE_TIME_FORMAT_TZ);
```
paoloantinori, line 34: Now that the `@ConfigProperty` injection points are removed, any deployment that still sets `apicurio.apis.date-format` or `apicurio.apis.date-format-timezone` (application.properties, env vars) will have those values silently ignored, and the REST API date output will silently switch to ISO-8601. The previous deprecation warning was the only migration signal and is gone with this change. Consider keeping a startup check that logs a warning, or fails fast per the codebase convention, when either removed property is still present, so operators get a signal rather than a silent format change.

docs/modules/ROOT/pages/getting-started/assembly-config-reference.adoc
```diff
@@ -94,16 +94,6 @@ The following {registry} configuration options are available for each component
...
-|`string`
-|`yyyy-MM-dd'T'HH:mm:ss'Z'`
-|`2.4.3.Final`
-|API date format (DEPRECATED - do not use)
-|`apicurio.apis.date-format-timezone`
-|`string`
-|`UTC`
-|`2.4.3.Final`
-|API date format timezone (DEPRECATED - do not use)
 |`apicurio.apis.v3.base-href`
```
paoloantinori, line 97: Since this changes REST API output for any deployment that had the date-format properties set, consider adding a migration-guide or CHANGELOG entry noting that `apicurio.apis.date-format` and `-timezone` are removed and the format is now fixed to ISO-8601. The config-reference rows are dropped here, but operators upgrading need a discoverable note prompting them to remove the properties.

paoloantinori (commented):
The code change itself is correct and the simplification is sound: the `try/catch` around now-constant values in `customize()` is appropriately dropped (the constants are valid literals, so `configureDateFormat` cannot throw), the unused imports (`Info`, `ConfigProperty`, `CATEGORY_API`) are removed cleanly, and both `.adoc` files are updated consistently. No correctness or security issue in the code. The substantive concern is operator migration, plus two things to confirm.

- **Silent format change for existing deployments.** With the `@ConfigProperty` injection points gone (`JacksonDateTimeCustomizer.java:34`), any deployment still setting `apicurio.apis.date-format` / `-timezone` (in `application.properties` or env) will have those values silently ignored, since MicroProfile Config drops unknown properties, and its REST API date output will silently switch to ISO-8601. The previous deprecation warning was the only migration signal and it is removed in this same PR. Consider keeping a startup check that warns, or fails fast per the codebase convention, when either removed property is still present.
- **No migration or CHANGELOG note.** This is a behavior-affecting change to REST API output. The config-reference rows are dropped (`assembly-config-reference.adoc:97`, `ref-registry-all-configs.adoc`), but there is no migration-guide or CHANGELOG entry telling operators to remove the properties and that the format is now fixed. Adding one makes the silent change discoverable.
- **Confirm date-format coverage remains.** `LegacyV2ApiDateFormatTest.java` (deleted) was the test exercising server date formatting. Confirm other coverage asserts the default output is ISO-8601 and parseable by the v2/v3 client SDKs; if none exists, add one so the enforced default is regression-protected.

EricWittmann:
Thanks for working on this, @Saloni3494! The code changes here look correct and clean.

However, since this is a breaking change (removing configuration properties that existing deployments may rely on), it needs to target the 4.0 release rather than the current 3.x line. A maintainer noted this on the issue as well: https://github.com/Apicurio/apicurio-registry/issues/7329#issuecomment-4966841621

Closing this PR for now — we'll revisit it when 4.0 development begins. The `version/4.0` label has been added to the issue to track this.

PR #9195, opened 2026-08-03
fix(app): remove unreachable else-if in label parsing for searchGroups and searchVersions

### LINKED ISSUE #8734: bug: Unreachable else-if in label parsing for searchGroups and searchVersions
---
name: Bug report
about: Use this template to report a bug in this project
title: "Unreachable else-if in label parsing for searchGroups and searchVersions"
labels: Bug
---

## Description

**Registry Version**: snapshot (confirmed at commit `5115d113`)
**Persistence type**: All (logic bug in REST layer, not persistence-specific)

In `SearchResourceImpl.java`, the label-parsing logic inside `searchGroups()` (line 233) and `searchVersions()` (line 303) contains an `else if` branch that can never be reached. The preceding `if` condition uses a logical OR that already subsumes the `else if` condition, making the `else if` dead code.

**Affected location:** [`SearchResourceImpl.java#L236`](https://github.com/Apicurio/apicurio-registry/blob/5115d1136b495c27f331cb977464780a8b12408d/app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java#L236)

The buggy pattern (lines 233–242, repeated similarly around line 303):

```java
if ((delimiterIndex == (prop.length() - 1)) || delimiterIndex < 0) {
    labelKey = prop.replace(":", "");
    labelValue = null;
} else if (delimiterIndex == (prop.length() - 1)) {  // ← UNREACHABLE: already covered by the OR above
    labelKey = prop.substring(0, delimiterIndex);
    labelValue = null;
} else {
    labelKey = prop.substring(0, delimiterIndex);
    labelValue = prop.substring(delimiterIndex + 1);
}
```

Because the first `if` condition is `(delimiterIndex == (prop.length() - 1)) || delimiterIndex < 0`, any input where `delimiterIndex == (prop.length() - 1)` is true will always be consumed by that first branch. The `else if` below, which tests the identical sub-expression `delimiterIndex == (prop.length() - 1)`, is therefore logically impossible to enter.

This is structurally inconsistent with `searchArtifacts()` (around line 119–128 of the same file), which correctly uses two separate, non-overlapping conditions:

```java
if (delimiterIndex < 0) {
    labelKey = prop;          // uses prop directly, not prop.replace(":", "")
    labelValue = null;
} else if (delimiterIndex == (prop.length() - 1)) {
    labelKey = prop.substring(0, delimiterIndex);
    labelValue = null;
}
```

There is also a secondary inconsistency: in `searchGroups` and `searchVersions`, the `delimiterIndex < 0` path uses `prop.replace(":", "")` to strip any colon, whereas `searchArtifacts` assigns `labelKey = prop` directly. When `delimiterIndex < 0` no colon is present in `prop`, so both produce the same string value — but the intent expressed in the code differs from the reference implementation.

### Environment

No response

### Steps to Reproduce

1. Check out commit `5115d113` (or current `main`).
2. Run the following command to observe the duplicated condition:

```
grep -n 'delimiterIndex == (prop.length() - 1)' \
  app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
```

Expected output will show the same sub-expression appearing on both the `if` line and the `else if` line for `searchGroups` (around lines 233 and 236) and again for `searchVersions` (around lines 303 and 306), confirming the `else if` is unreachable.

3. Compare with the `searchArtifacts()` block (around line 119–128) to observe the correct, non-overlapping structure.

**Expected:** The label-parsing conditions in `searchGroups()` and `searchVersions()` should match the correct logic in `searchArtifacts()`, with `delimiterIndex < 0` and `delimiterIndex == (prop.length() - 1)` handled as two distinct, mutually exclusive branches.

**Actual:** The first `if` condition combines both sub-expressions with `||`, so the `else if` that follows with the identical sub-expression `delimiterIndex == (prop.length() - 1)` is dead code and can never be entered. For label filters where the key has a trailing colon (e.g. `"mykey:"`), the code enters the first `if` branch and uses `prop.replace(":", "")` rather than `prop.substring(0, delimiterIndex)`. The `else if` branch (which would use `substring`) is silently skipped in all cases.

### Logs

No runtime error is produced — this is a silent logic defect. No relevant log output.

deepak0x:
## Description

Issue #8734 reported an unreachable `else-if` branch in the label-parsing logic of `SearchResourceImpl`.

In `searchGroups()` and `searchVersions()`, the first condition was:

```java
if ((delimiterIndex == (prop.length() - 1)) || delimiterIndex < 0) {
    labelKey = prop.replace(":", "");
    labelValue = null;
} else if (delimiterIndex == (prop.length() - 1)) { // unreachable
    ...
}
```

Because the first `if` already covers `delimiterIndex == (prop.length() - 1)`, the `else if` testing the identical sub-expression can never execute. This also diverged from the correct, non-overlapping structure already used in `searchArtifacts()`.

### Fix

Restructure both methods to mirror `searchArtifacts()`:

```java
if (delimiterIndex < 0) {
    labelKey = prop;
    labelValue = null;
} else if (delimiterIndex == (prop.length() - 1)) {
    labelKey = prop.substring(0, delimiterIndex);
    labelValue = null;
} else {
    labelKey = prop.substring(0, delimiterIndex);
    labelValue = prop.substring(delimiterIndex + 1);
}
```

This restores the intended behaviour: a trailing `:` (e.g. `mykey:`) matches the label key with any value, and a key with no `:` is matched directly.

## Test Evidence

Added `testSearchGroupsByLabelTrailingDelimiter` and `testSearchVersionsByLabelTrailingDelimiter`, which assert that a trailing-colon and a key-only label filter both match by key.

```
./mvnw test -pl app -am -Dtest='SearchGroupsTest,SearchVersionsTest' -Dsurefire.failIfNoSpecifiedTests=false
BUILD SUCCESS
```

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] DCO sign-off included (`Signed-off-by`)
- [x] Tests added/updated for the changed path
- [x] Commit message follows Conventional Commits

Fixes #8734


deepak0x:
Thanks @EricWittmann — this fixes the unreachable else-if in label parsing reported in #8734. The logic now matches searchArtifacts(), and I added tests covering the trailing-colon and key-only cases. Could you take a look when you have a moment?

paoloantinori (commented):
Overall: this is a correct, behavior-preserving cleanup that removes a genuinely unreachable `else-if` and restructures both `searchGroups()` and `searchVersions()` onto a single non-overlapping form; I could not construct any input where old and new behavior differ, and the added tests cover both previously-collapsed branches. I did not find any MAJOR+ issues.

- **Behavior-preserving by construction.** In `app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java`, the old first `if ((delimiterIndex == (prop.length() - 1)) || delimiterIndex < 0)` already subsumed the subsequent `else if (delimiterIndex == (prop.length() - 1))` (searchGroups ~L234-237, searchVersions ~L304-307), so the `else if` body was dead code. The new split is equivalent: the no-`:` branch sets `labelKey = prop`, identical to the old `prop.replace(":", "")` (no colon ⇒ replace is a no-op); the trailing-`:` branch sets `labelKey = prop.substring(0, delimiterIndex)`, identical to `replace(":", "")` here because when the *first* colon sits at the last index it is the *only* colon. Mid-string and multi-colon inputs (`a:b:c`) fall through to the unchanged `else`.
- **Tests cover the two branches that the old collapsed `if` hid.** `testSearchGroupsByLabelTrailingDelimiter` (`app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchGroupsTest.java` ~L209-236) and `testSearchVersionsByLabelTrailingDelimiter` (`app/src/test/java/io/apicurio/registry/noprofile/rest/v3/SearchVersionsTest.java` ~L313-347) assert that both `trailing:` and `trailing` filters match by key.
- **Minor, non-blocking:** both new tests assert an exact count of 3 from a global label search keyed on `trailing`, which assumes that key is unique across the shared test store; this matches the conventions already used by neighboring tests, so I would not block on it. In `SearchVersionsTest.java` the local `CreateArtifactResponse car = null` initializer is redundant (it is reassigned on every loop iteration).
- **Unverified claim:** the PR states this mirrors the existing `searchArtifacts()` structure; the apicurio source was not present in this read-only review environment, so I could not confirm that against ground truth — worth a quick visual diff against `searchArtifacts()` before merge.

deepak0x:
Thanks @paoloantinori for the thorough review.

- **Mirror against `searchArtifacts()`:** confirmed — `searchGroups()` and `searchVersions()` now use the exact same non-overlapping branch structure as `searchArtifacts()` (lines 118-131 of `SearchResourceImpl.java`): `delimiterIndex < 0` → `prop`, `== length-1` → `substring(0, delimiterIndex)`, `else` → key+value split. No behavior change, just removing the dead `else if`.
- **Redundant `car = null` initializer:** cleaned up in `testSearchVersionsByLabels` (removed the unused `null` initializer since it's reassigned on every loop iteration before use). Pushed in 191de2d.
- **Exact count of 3 assertion:** kept as-is, consistent with the existing neighboring tests in this file that make the same uniqueness assumption for generated keys.

Tests pass locally (`testSearchGroupsByLabelTrailingDelimiter`, `testSearchVersionsByLabelTrailingDelimiter`). Ready whenever a maintainer can `/accept` and run the lifecycle.

paoloantinori:
Thanks @deepak0x — re-read against HEAD 191de2d and both fixes landed exactly as described. `searchGroups` and `searchVersions` now use the same 3-branch shape as `searchArtifacts` (the trailing-`:` and key-only paths both land at `SearchFilter.ofLabel(key, null)`), and the `CreateArtifactResponse car = null` initializer is gone. The two new `TrailingDelimiter` tests exercise the previously-collapsed branches. Nothing further from me.

deepak0x:
The two failing jobs are unrelated to this PR's change.

This PR only touches `SearchResourceImpl.java` (label-parsing fix) and the two `SearchGroupsTest` / `SearchVersionsTest` tests. The failures are in different modules:

- `Unit Tests (app-other)`: `io.apicurio.registry.noprofile.maven.TestUpdateRegistryMojoTest` — a Maven plugin test, not label parsing.
- `Unit Tests (app-kafkasql)`: `io.apicurio.registry.event.kafkasql.KafkaSqlEventsTest` — 6 errors, all ~20s timeouts (e.g. `versionContractRulesetDeleted`), which is infrastructure flakiness in the Kafka event tests, not a logic failure.

Both modules passed on the other storage variants (app-sql, app-gitops, app-kubernetesops, app-rest), so the failures are variant-specific flakiness rather than a regression from this change. The label-parsing tests added here run in those passing variants.

Requesting a re-run of the `app-other` and `app-kafkasql` unit test jobs. If they keep failing, they should be filed as a separate flaky-test issue since they are outside this PR's scope.

jsenko:

deepak0x:
The `Build Java Application (no tests)` failure on commit 9440864 is a CI-infrastructure issue, not a regression from this change.

Evidence:
- The Build job failed in ~1m28s. Compiling the `app` module (a 30-module Maven project) takes many minutes, so a 1m28s failure is an environment/runner problem (OOM kill or cache/checkout failure), not a compilation error in this PR's code.
- This PR only touches `SearchResourceImpl.java` (label-parsing fix) and the two `SearchGroupsTest` / `SearchVersionsTest` tests. That change already compiled, was reviewed, and was accepted at 191de2d (paoloantinori + EricWittmann `/accept`). The rebase onto `main` only pulls in unrelated upstream commits (#9287 OOM fix, etc.); the `SearchResourceImpl.java` diff is unchanged and cannot suddenly fail to compile.
- Before the rebase, the failure was the `app-other` OOM (`TestUpdateRegistryMojoTest`), which @jsenko noted was fixed in #9287 (merged) and advised rebasing. After the rebase, the failure moved to the Build job (which compiles the whole project) — the same resource pressure now hitting the larger Build step.

Requesting a `/retry` of the Verify workflow. If the Build job keeps failing after a retry, it should be filed as a separate CI-infra issue, as it is outside this PR's scope.

EricWittmann:
Thanks for the clean fix and thorough follow-up, @deepak0x! Merging now.

Issue #9207, opened 2026-08-03
DataExporter silent error handling risks corrupt or incomplete Zip export archives


hussainjamal760:
## Feature / Problem Description

During registry data export via:

```http
GET /apis/registry/v3/admin/export
```

`DataExporter` iterates over all storage entities and writes them to a `ZipOutputStream`.

Currently, if `EntityWriter` encounters an exception while writing an entity, `DataExporter`:

1. Logs the error
2. Increments `errorCounter`
3. Continues processing the remaining entities
4. Closes the ZIP stream and returns `HTTP 200 OK`

As a result, the recipient may receive a **partial or incomplete ZIP archive** with missing entities, while the export is still reported as successful.

Because the export response does not return an error or include a manifest warning indicating that entities were omitted, the incomplete archive may later:

* Fail during import into another registry instance
* Result in missing registry data
* Cause silent data loss without clearly indicating that the original export was incomplete

### Root Cause

Entity-writing failures are currently swallowed during the streaming export process. Although `errorCounter` is incremented, its value does not affect the final export result.

This allows incomplete archives to be returned with `HTTP 200 OK`, even when one or more entities failed to export.

## Proposed Solution

Update `DataExporter` so that entity-writing failures cause the export operation to fail instead of producing a potentially incomplete archive.

The proposed approach is to:

1. Re-throw exceptions encountered inside the entity consumer as a `RuntimeException`
2. Check whether `errorCounter.get() > 0` after processing the entities
3. Throw an `IOException` when one or more entity-writing errors occurred

This ensures that the ZIP export is aborted when an entity cannot be written and prevents incomplete archives from being returned as successful exports.

## Expected Behavior

If any entity fails to export:

* The export operation should fail
* The ZIP stream should be aborted rather than completed successfully
* The API should return an appropriate error response instead of `HTTP 200 OK`
* A potentially incomplete archive should not be presented as a valid export

## Additional Context

N/A

AlgoArtist06:
/assign-me

I would like to take this focused fix. I will trace the export failure path, preserve the original error while preventing a partial archive from being reported as successful, and add regression coverage for entity-write failure plus the existing successful-export path. I will keep the change limited to the exporter and its tests.

AlgoArtist06:
After checking the current `main` source, I found that `DataExporter` already rethrows entity-write failures and aborts the streaming ZIP path. I will not open a duplicate PR for this issue; the issue appears to describe an older revision.

[...]

PR #9208, opened 2026-08-03
fix(core): abort zip stream on entity write failure during export

hussainjamal760:
## Description

Fixes silent error handling during registry data export #9207

Previously, if `EntityWriter` encountered an exception while writing an entity, `DataExporter` would log the error and increment the `errorCounter`, but it would continue writing the rest of the entities and close the stream with an HTTP 200 OK. This resulted in users receiving a partial or corrupted ZIP file with missing entities and no warning that data was omitted.

This PR addresses the issue by:
1. Re-throwing the caught exception inside the entity consumer as a `RuntimeException` to abort the export loop early.
2. Checking `errorCounter.get() > 0` after the `exportData` call and explicitly throwing an `IOException` to ensure the ZIP stream is safely aborted and a proper error response is propagated instead of a false `200 OK`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/Apicurio/apicurio-registry/blob/main/CONTRIBUTING.md) document.
- [x] I have added tests to cover my changes (N/A - covered by existing `AdminResourceTest`).
- [x] All new and existing tests passed.
- [x] My commits follow the Conventional Commits guidelines and include a DCO sign-off.


app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/DataExporter.java
```diff
@@ -51,11 +51,14 @@ public Response exportData(String groupId) {
                     } catch (Exception e) {
                         log.error("Error writing entity", e);
                         errorCounter.incrementAndGet();
+                        throw new RuntimeException("Error writing entity during export", e);
```
paoloantinori, line 50: This RuntimeException only aborts the export if it propagates out of the entity-iteration loop. If the iteration framework catches consumer exceptions itself (the original catch/log/count/continue pattern suggests it might), this throw is swallowed and the early abort never happens — only the `errorCounter.get() > 0` check below would catch it. If it does propagate, that post-loop check is dead code and `zip.flush()/close()` (lines 63–64) are skipped, so make sure `zip` is closed via try-with-resources or a finally block. Worth confirming the real propagation behavior and consolidating to a single mechanism.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/DataExporter.java
```diff
@@ -51,11 +51,14 @@ public Response exportData(String groupId) {
...
                     }
                     return null;
                 });
 
-                // TODO if the errorCounter > 0, then what?
+                if (errorCounter.get() > 0) {
```
paoloantinori: Unreachable unless the RuntimeException thrown in the catch above is swallowed by the iteration framework: on the error path the throw propagates past here, and on the success path `errorCounter` is 0. If you keep this path, note it can't preserve the original exception cause (it only counts), unlike the rethrow above — pick deliberately which behavior you want, since both mechanisms coexist here without a clear intent.

paoloantinori (commented):
The fix does the right thing — it stops serving a partial ZIP with HTTP 200 OK when an entity write fails — but the two new error paths work against each other: at most one of them actually aborts the stream, and under standard Java propagation the post-loop check is unreachable dead code. Worth tightening before merge.

- `DataExporter.java:54` — Throwing `RuntimeException` inside the entity consumer only aborts the loop if that exception propagates out of the iteration machinery. If the iteration framework wraps each consumer call in its own try/catch (the original catch/log/count/continue design implies it might), this throw is swallowed and the early abort never happens — only the `errorCounter.get() > 0` path would catch it. Conversely, if it does propagate (standard behavior), the `errorCounter` check at line 59 is unreachable and the explicit `zip.flush()/close()` at lines 63–64 are skipped, so confirm `zip` is closed via try-with-resources or a finally block.
- `DataExporter.java:59` — This branch is reachable only if the RuntimeException above is swallowed; otherwise it's dead code (the throw propagates past it, and on the no-error path the counter is 0). The errorCounter design also can't preserve the original exception's cause (it only counts), unlike the rethrow — so the choice between the two mechanisms isn't free.
- No regression test asserts the new behavior (export aborts and returns a non-200 on a write failure); the checklist's claim of coverage by existing `AdminResourceTest` is worth confirming with a test that injects a write failure.

hussainjamal760:
@paoloantinori Thanks for the sharp review! You were spot on about the error propagation and resource cleanup. I've updated the PR with a much tighter implementation:

Clean Error Propagation: Removed the redundant AtomicInteger errorCounter. If an entity fails to write, the RuntimeException is thrown immediately. This aborts the storage.exportData(...) iteration and propagates out to the StreamingOutput, where it gets converted into an IOException, correctly aborting the stream.
Resource Management: Wrapped the ZipOutputStream in a try-with-resources block. This guarantees that zip.close() (and flush) is safely executed whether the export finishes successfully or fails midway.
Cleaned Imports: Cleaned up the unused AtomicInteger import.
I also verified that the AdminResourceTest and related tests pass locally. Let me know if this looks good to merge!

> The fix does the right thing — it stops serving a partial ZIP with HTTP 200 OK when an entity write fails — but the two new error paths work against each other: at most one of them actually aborts the stream, and under standard Java propagation the post-loop check is unreachable dead code. Worth tightening before merge.
> 
> * `DataExporter.java:54` — Throwing `RuntimeException` inside the entity consumer only aborts the loop if that exception propagates out of the iteration machinery. If the iteration framework wraps each consumer call in its own try/catch (the original catch/log/count/continue design implies it might), this throw is swallowed and the early abort never happens — only the `errorCounter.get() > 0` path would catch it. Conversely, if it does propagate (standard behavior), the `errorCounter` check at line 59 is unreachable and the explicit `zip.flush()/close()` at lines 63–64 are skipped, so confirm `zip` is closed via try-with-resources or a finally block.
> * `DataExporter.java:59` — This branch is reachable only if the RuntimeException above is swallowed; otherwise it's dead code (the throw propagates past it, and on the no-error path the counter is 0). The errorCounter design also can't preserve the original exception's cause (it only counts), unlike the rethrow — so the choice between the two mechanisms isn't free.
> * No regression test asserts the new behavior (export aborts and returns a non-200 on a write failure); the checklist's claim of coverage by existing `AdminResourceTest` is worth confirming with a test that injects a write failure.

paoloantinori:
@hussainjamal760 The rework is exactly right — much cleaner than the dual error paths. Confirmed:

- Dropping the `errorCounter` and re-throwing from inside the consumer is the correct fix: the `RuntimeException` aborts the `storage.exportData(...)` iteration immediately, so we no longer keep writing entities after a failure.
- Moving `ZipOutputStream` into try-with-resources ensures the stream is closed on both success and failure paths — the previous `zip.close()` in the happy path left the failure path unclosed.
- The outer `catch (Exception e) { throw new IOException(...) }` correctly wraps the `RuntimeException` so the `StreamingOutput` aborts the HTTP response instead of completing with a 200 OK on a truncated ZIP.

One thing worth noting for the record (not a blocker): the response headers are committed once the first byte of the ZIP is written, so a mid-export failure still surfaces to the client as a truncated chunked stream rather than a clean 4xx/5xx. That's inherent to streaming the export and matches the pre-existing behavior for I/O errors, so accepting it here is the right call — the important regression (silent partial ZIP with 200 OK) is gone.

LGTM from my side.

hussainjamal760:
> @hussainjamal760 The rework is exactly right — much cleaner than the dual error paths. Confirmed:
> 
> * Dropping the `errorCounter` and re-throwing from inside the consumer is the correct fix: the `RuntimeException` aborts the `storage.exportData(...)` iteration immediately, so we no longer keep writing entities after a failure.
> * Moving `ZipOutputStream` into try-with-resources ensures the stream is closed on both success and failure paths — the previous `zip.close()` in the happy path left the failure path unclosed.
> * The outer `catch (Exception e) { throw new IOException(...) }` correctly wraps the `RuntimeException` so the `StreamingOutput` aborts the HTTP response instead of completing with a 200 OK on a truncated ZIP.
> 
> One thing worth noting for the record (not a blocker): the response headers are committed once the first byte of the ZIP is written, so a mid-export failure still surfaces to the client as a truncated chunked stream rather than a clean 4xx/5xx. That's inherent to streaming the export and matches the pre-existing behavior for I/O errors, so accepting it here is the right call — the important regression (silent partial ZIP with 200 OK) is gone.
> 
> LGTM from my side.

Thanks for the thorough review and confirmation! Glad the rework addresses the issue cleanly. Appreciate the LGTM! 🙌

EricWittmann:
The UI failure is unrelated to this PR.  I have a fix for that pending.

hussainjamal760:
> The UI failure is unrelated to this PR. I have a fix for that pending.

Thanks for confirming, Eric! Appreciate the clarification.

jsenko:

EricWittmann:
LGTM thanks @hussainjamal760 :clinking_glasses:

PR #9210, opened 2026-08-03
fix(ui): prevent overview split panes from collapsing (#9209)

### LINKED ISSUE #9209: UI: Resizable overview split pane collapses and hides right-pane   content
The resizable split-pane layout breaks when the divider is dragged
  too far to the right, shrinking the right pane.

  This reproduces on:

  - `/explore/default/<artifact>` — Artifact Overview: “Artifact
  metadata” and “Versions in artifact”
  - `/explore/default/<artifact>/versions/<n>` — Version Detail:
  “Version metadata” and “Comments”
  - `/explore/default/<group>` — Group Overview: “Group metadata”
  and “Artifacts in group”

  ## Actual behavior

  When the right pane becomes narrow:

  1. Table/list rows can disappear while headers and pagination
  remain visible.
  2. Further shrinking causes the right pane to detach/compress into
  a clipped column.
  3. Toolbar controls, search fields, pagination, and action buttons
  overlap, stack incorrectly, or are cut off.
  4. Moving the divider back restores the content, indicating a
  layout/CSS issue rather than a data-loading problem.

  ## Expected behavior

  - Both panes should have a sensible minimum width.
  - The divider must not allow the right pane to collapse below its
  safe width.
  - Tables and lists should scroll horizontally if needed instead of
  hiding rows or columns.
  - Toolbar controls should wrap cleanly without clipping or
  overlap.
  - Truncated labels should show their full value on hover.

  ## Reproduction steps

  1. Open any affected route.
  2. Drag the metadata-pane divider to the right until the content
  pane is very narrow.
  3. Observe disappearing content and broken toolbar layout.

  ## Recording

  Attach the “before fix” recording to this issue:
 

https://github.com/user-attachments/assets/67471379-08fd-42b3-a601-2c33e025ef28

Faisal77666:
Fixes #9209

  ## What

  Adds a shared resizable overview drawer for Artifact Overview,
  Version Detail, and Group Overview.

  The shared layout:

  - Enforces a 400px minimum width for both panes.
  - Prevents the divider from shrinking the right pane below its
  safe width.
  - Makes tables/lists scroll horizontally instead of hiding
  content.
  - Makes toolbar controls wrap without clipping or overlap.
  - Adds native hover titles to truncated table headers.

  ## Why

  Dragging the divider too far right previously collapsed the
  content pane, causing table rows to disappear and toolbar controls
  to overlap or be clipped.

  ## Before

[raw-recording (3).webm](https://github.com/user-attachments/assets/e48c5dd8-1ea2-4804-87ee-b585a61b43a1)

  ## After

[raw-recording (4).webm](https://github.com/user-attachments/assets/3c980ec5-22b4-4232-889c-8729f45d8a15)
  

  ## Validation

  - Verified Artifact Overview, Version Detail, and Group Overview.
  - Tested full width, approximately 450px, and drag attempts below
  400px.
  - Confirmed sorting, filtering, pagination, row selection,
  comments, and page actions remain available.
  - Ran targeted UI ESLint checks.


ui/ui-app/src/app/components/common/ResizableOverviewDrawer.css
```diff
@@ -0,0 +1,67 @@
...
+.resizable-overview-drawer :is(.versions-table, .artifacts-table, .comments) {
+    max-width: 100%;
+    overflow-x: auto;
+}
+
+.resizable-overview-drawer .pf-v6-c-table {
+    min-width: 700px;
+}
+
+.resizable-overview-drawer .pf-v6-c-table.pf-m-grid {
```
paoloantinori: This block forces any `pf-m-grid` table back to `display: table` (and `.pf-m-hidden` cells back to `display: table-cell` at line 41), which disables PatternFly's responsive card layout inside these drawers — on narrow screens users get a 700px-min horizontally-scrolling table instead of the stacked card view. Two things worth confirming on a small window: (1) this is the intended tradeoff for all three pages, and (2) because the per-cell `data-label` `::before` headers are scoped to `.pf-m-grid`, forcing table display while that class is active can render the column label twice (header row plus each cell) — please verify the rendered output at the responsive breakpoint.

ui/ui-app/src/app/components/common/ResizableOverviewDrawer.css
```diff
@@ -0,0 +1,67 @@
+.resizable-overview-drawer {
+    min-width: 800px;
```
paoloantinori: `min-width: 800px` enforces panel(400) + content(400), but this rule doesn't add an overflow container on `.resizable-overview-drawer` itself — overflow is only set on `.pf-v6-c-drawer__content` and the panel. If the enclosing `Card`/page doesn't provide horizontal scroll, viewports narrower than ~800px (e.g., with the app masthead/nav reducing available width) will produce whole-page horizontal overflow. The validation notes cover pane widths (450px, sub-400px drags) but not viewport widths, so worth a check on a narrow window.

paoloantinori (commented):
The shared `ResizableOverviewDrawer` extraction cleanly centralizes the resize constraints, and the `minSize: 400px` / `maxSize: calc(100% - 400px)` pair in `ResizableOverviewDrawer.tsx:21-22` correctly guarantees the content pane can't be dragged below its safe width — the core collapse bug is fixed. The remaining concerns are concentrated in the new CSS, which imposes a hard 800px floor and overrides PatternFly's responsive table mode; the PR validation covers pane widths (450px, sub-400px drags) but not viewport widths, so the narrow-screen path is the main gap.

- Core fix is sound: `ui/ui-app/src/app/components/common/ResizableOverviewDrawer.tsx:21-22` clamps the panel between 400px and `calc(100% - 400px)`, leaving the content pane at least 400px — directly addresses issue #9209.
- `ResizableOverviewDrawer.css:2` sets `min-width: 800px` on the drawer but adds no `overflow` on that element (overflow is only set on `.pf-v6-c-drawer__content`/the panel); confirm the enclosing `Card`/page supplies horizontal scroll, otherwise viewports under ~800px get whole-page horizontal overflow.
- `ResizableOverviewDrawer.css:25-42` forces any `pf-m-grid` table back to `display: table` and re-shows `.pf-m-hidden` cells, disabling PatternFly's responsive card layout inside these drawers; verify it doesn't also render the per-cell `data-label` header twice at the responsive breakpoint.
- Minor: `ResizableOverviewDrawer.tsx:17` uses `cloneElement` to inject sizing props, silently overriding anything a caller passes; since this is now a shared component, consider owning `<DrawerPanelContent>` here and taking the head/body as children to make the contract explicit.
- Minor: the `title={column.label}` additions (`VersionsTable.tsx:205`, `GroupArtifactsTable.tsx:146`) are a nice low-cost win; native `title` tooltips aren't keyboard/touch accessible, so swap for PatternFly `Tooltip` later if fuller accessibility is wanted.

Faisal77666:
@paoloantinori  Thanks for the feedback — I’ve adjusted the shared drawer so the overflow stays contained within the drawer on narrower viewports, and I’ve kept the responsive table behavior intact instead of forcing the older table layout. I also verified the UI build and tests locally.

paoloantinori:
Thanks @Faisal77666 — the new head reads well. Containing the overflow inside the drawer with `overflow-x: auto` on `.resizable-overview-drawer` and the nested `.pf-v6-c-drawer__content/body` selectors is the right call: the panel keeps its min width while the table gains a horizontal scrollbar instead of collapsing columns. Keeping the responsive table behavior (rather than reverting to the fixed layout) is the correct tradeoff. Glad the local build and tests came through clean. From my side this addresses the feedback — handing back to the maintainers for the final merge call.

Faisal77666:
@paoloantinori  apologies for reaching out via DM. I sent you a message on Slack regarding some guidance. Whenever you have time, I'd really appreciate your response. Thank you!

paoloantinori:
Hi @Faisal77666 — noted, thanks for the heads-up. I'll take a look at your Slack message and follow up there. In the meantime, the changes on this PR look good on my end.

Faisal77666:
@EricWittmann please review PR

Faisal77666:
@EricWittmann ready to merge

carlesarnal:
Reviewed the final state of this PR. Overall it looks solid — nice work on the shared `ResizableOverviewDrawer` extraction.

**What works well:**
- The `minSize="400px"` / `maxSize="calc(100% - 400px)"` pair in `ResizableOverviewDrawer.tsx:23-24` correctly prevents either pane from collapsing, directly addressing #9209.
- Clean elimination of drawer boilerplate across the three consumer pages — `head`/`body` props make the component contract explicit and easy to understand.
- CSS approach is sound: `overflow-x: auto` on the drawer and nested content, `flex-wrap` on toolbars, and `min-width: 0` to allow flex children to shrink. Good responsive strategy.
- Cleaned up unused imports (`Drawer`, `DrawerContent`, `DrawerContentBody`, `DrawerPanelContent`) from all three consumer files — no dead code left behind.

**Minor observations for future cleanup (not blocking):**
1. **`ResizableOverviewDrawer.css:6`** — `-webkit-overflow-scrolling: touch` is deprecated in modern browsers (momentum scrolling is now the default on iOS). Can be removed in a future pass.
2. **`ResizableOverviewDrawer.tsx:31`** — `style={{ backgroundColor: "white" }}` hardcodes a color value. If a dark theme is ever introduced, this should use a PatternFly CSS variable (e.g., `--pf-t--global--background--color--primary--default`) instead.
3. **DCO**: The second commit (`89b738d`) is missing a `Signed-off-by` trailer. The first commit has it, but worth keeping in mind for future contributions.

Good contribution overall — the fix is correct, the abstraction is well-placed, and the maintainer feedback from @paoloantinori was properly addressed.

PR #9074, opened 2026-07-30
fix(http): accept all 2xx responses and restore interrupt flag in HttpClientService

DSingh0304:
Fixes #8691

## What this PR does

Fixes improper retry loops and bugged webhook delivery in `HttpClientService.post()`:

---

### 1. Accept all HTTP 2xx status codes & handle non-transient content errors

- **2xx Success Handling:** Any status code in `[200, 299]` is treated as success. This resolves issues where standard webhook responses like `202 Accepted` (CloudEvents acknowledgment) or `201 Created` incorrectly triggered 8 retries and failed.
- **`Void.class` Safety:** When `Void.class` is requested, `null` is returned directly without attempting deserialization (handling `204 No Content` and 2xx responses with ignored payloads).
- **Non-Retriable Content Errors:** If an empty body is returned when a non-`Void` type is expected, or if JSON deserialization fails (`DecodeException`), `HttpClientContentException` is thrown.
- **Non-Retriable 4xx Client Errors:** HTTP 4xx responses (400, 401, 403, 404, 410, etc.) throw `HttpClientErrorException`.
- **Abort on Non-Transient Failures:** `HttpClientContentException` and `HttpClientErrorException` are added to `@Retry(abortOn = ...)` so permanent errors abort immediately on attempt 1 instead of spinning through 8 exponential-backoff retries (~25s delay and 9× load), while transient 5xx server errors continue to be retried.

```java
int statusCode = httpResponse.statusCode();
if (statusCode >= 200 && statusCode < 300) {
    if (outputClass == Void.class) {
        return null;
    }
    Buffer bodyBuffer = httpResponse.body();
    if (bodyBuffer == null || bodyBuffer.length() == 0) {
        throw new HttpClientContentException("Webhook returned " + statusCode + " with empty body, expected " + outputClass.getSimpleName());
    }
    return decodeResponse(httpResponse, outputClass, statusCode);
} else if (statusCode >= 400 && statusCode < 500) {
    throw new HttpClientErrorException("Webhook request failed (" + statusCode + "): " + httpResponse.statusMessage());
} else {
    throw new HttpClientException("Webhook request failed (" + statusCode + "): " + httpResponse.statusMessage());
}
```

---

### 2. Restore interrupt flag & abort `@Retry` on `InterruptedException`

- Restores `Thread.currentThread().interrupt()` upon catching `InterruptedException`.
[...]

PR #9211, opened 2026-08-04
fix(rest): accept full 2xx range and restore interrupt flag in HttpClientService

### LINKED ISSUE #8691: Bug: HttpClientService rejects valid HTTP 2xx responses and swallows interrupt flag

om7057:
Fixes #8691

## Summary

`HttpClientService.post()` currently treats only HTTP `200 OK` as a successful response. Other successful HTTP status codes such as `201`, `202`, and `204` result in a `HttpClientException`, causing SmallRye Fault Tolerance's `@Retry` interceptor to retry unnecessarily. This is particularly problematic for webhook endpoints that legitimately respond with `202 Accepted`.

This PR updates `post()` to:

- Treat the full `2xx` range (`200–299`) as success.
- Return `null` for `204 No Content` (or an empty response body) instead of attempting JSON deserialization.
- Restore the thread's interrupt flag before rethrowing as `HttpClientInterruptException` when `InterruptedException` is caught, following Java concurrency best practices.

## Notes

This supersedes the core behavior from #8692.

While investigating the previous PR, I also verified how SmallRye Fault Tolerance handles interrupts. For synchronous invocations, `InterruptedException` is treated as non-retryable internally before `abortOn`/`retryOn` is evaluated, so retry termination does not depend on `abortOn = HttpClientInterruptException.class`. `HttpClientInterruptException` is retained as the public API exception type for callers.

## Testing

Added `HttpClientServiceTest` using a Vert.x stub server to verify:

- `200` responses are parsed successfully.
- `202` responses are treated as success without triggering retries.
- `204` responses return `null` without attempting deserialization.
- Interrupting a thread blocked in `post()` results in a single request (`requestCount == 1`), confirming retries are aborted on interruption.


om7057:
@carlesarnal @paoloantinori @EricWittmann, since you were all on #8692, would appreciate a review whenever you have time.

paoloantinori:
Thanks for the ping and the detailed write-up — this is a solid fix.

The core change (treating the full `2xx` range as success and returning `null` for `204`/empty bodies) is correct and the test coverage for `200`/`202`/`204` is exactly what I'd want to see.

A couple of observations worth flagging:

1. **`HttpClientInterruptException` vs. `abortOn`.** Your own notes (and the `testInterruptAbortsRetry` assertion comment) state that SmallRye Fault Tolerance's `Retry` core strategy already marks an operation that observes `InterruptedException` as `EXCEPTION_NOT_RETRYABLE` before `abortOn`/`retryOn` is consulted. So `abortOn = HttpClientInterruptException.class` is effectively a no-op for retry termination in the synchronous path. That's fine as defensive documentation of intent, but worth calling out explicitly in the Javadoc so future readers don't assume `abortOn` is *doing* the work — your test asserts `requestCount == 1`, which proves the behavior, but the causal chain (interrupt handling in `Retry#retryLoopIteration`, not `abortOn`) is what should be preserved.

2. **Exception class hierarchy.** The comment on `HttpClientInterruptException` explains it extends `RuntimeException` directly "to stay within the project's class-hierarchy depth limit" rather than extending `HttpClientException`. That's a reasonable constraint, but it does change the public contract: callers currently catching `HttpClientException` from `post()` will silently miss the interrupt case. If there are existing callers that wrap `post()` in `try { ... } catch (HttpClientException e) { ... }`, they'll now let `HttpClientInterruptException` propagate uncaught. Worth confirming no caller in the codebase relies on `HttpClientException` as a catch-all for `post()` failures.

3. **`statusMessage()` on the error path.** Minor: `httpResponse.statusMessage()` for the failure branch — Vert.x's `HttpResponse.statusMessage()` can return `null` depending on how the response was constructed, which would render the error message as `"... failed (503): null"`. Not a blocker, just a cosmetic robustness note.

4. **Test port binding.** `server.listen(0)` in `@BeforeAll` is fine, but in a parallel `@QuarkusTest` execution context you may want to assert the listen result more defensively (the `onComplete` callback silently swallows a failure by never counting down the latch, which would surface as a generic timeout). Low priority.

Overall this looks mergeable to me once the `abortOn` Javadoc clarification is in. Nice work on superseding #8692 cleanly.

om7057:
@paoloantinori Thank you for going through the PR.
Addressed in 1a7681e: I clarified the Javadoc so it no longer implies abortOn is what stops retries on interrupt, and I also hardened the non-2xx error path to handle a null statusMessage() cleanly.

I also checked the current callers in the codebase and did not find any local catch (HttpClientException ...) sites around HttpClientService.post().

Please check the latest commit for the respective changes.

 Could you please accept the PR so it moves into a testable state?

paoloantinori:
Thanks @om7057 — `1a7681e` reads well:

- The Javadoc now correctly attributes retry termination to SmallRye's core strategy treating `InterruptedException` as non-retryable, rather than to `abortOn`. The annotation is kept as intent documentation, which is reasonable.
- The null/blank `statusMessage()` guard on the non-2xx error path is a good hardening — avoids an `NPE`-flavored message if a server returns a status line without a reason phrase.

On acceptance: I'm not able to `/accept` or merge from here — that needs to come from one of the maintainers (carlesarnal / jsenko / EricWittmann). The code change itself looks good to me and the test coverage (200/202/204 + interrupt-flag restoration) covers the regression. Pinging for a maintainer merge call.

om7057:
> Thanks @om7057 — `1a7681e` reads well:
> 
> - The Javadoc now correctly attributes retry termination to SmallRye's core strategy treating `InterruptedException` as non-retryable, rather than to `abortOn`. The annotation is kept as intent documentation, which is reasonable.
> - The null/blank `statusMessage()` guard on the non-2xx error path is a good hardening — avoids an `NPE`-flavored message if a server returns a status line without a reason phrase.
> 
> On acceptance: I'm not able to `/accept` or merge from here — that needs to come from one of the maintainers (carlesarnal / jsenko / EricWittmann). The code change itself looks good to me and the test coverage (200/202/204 + interrupt-flag restoration) covers the regression. Pinging for a maintainer merge call.
> 
> 


Thank you for the quick go through via the new commit @paoloantinori.
I'll wait for the further update.

paoloantinori:
Thanks @om7057 — no further changes needed from my side. `1a7681e` addresses the Javadoc accuracy point and the status-message null guard, and the test coverage (200/202/204 + interrupt-flag restoration) is sufficient. Standing by for a maintainer's merge call.

om7057:
@EricWittmann, just a small reminder on this one!

DSingh0304:
Hi @EricWittmann, @jsenko, and @om7057,

I noticed this PR was opened for issue #8691, but I wanted to ask how you would prefer to proceed since PR #9074 has been open and actively worked on for the same issue, and was already reviewed and approved by @jsenko?

In PR #9074, I put together a comprehensive fix covering:
- Proper 2xx range and Void response handling without unnecessary deserialization
- Dedicated exception classes (`HttpClientContentException`, `HttpClientErrorException`, `HttpClientInterruptedException`) mapped to `@Retry(abortOn=...)` to avoid retrying non-transient errors
- Restoring the thread interrupt flag on `InterruptedException`
- A complete behavioral test suite of 13 unit tests in `HttpClientServiceTest.`

Given that PR #9074 was already open, approved, and covers the full retry/exception behavior for #8691, could we please clarify if there was any specific reason for opening a duplicate PR, and whether we can proceed with merging PR #9074?

Thank you for your time and guidance!

om7057:
@DSingh0304 apologies for the slow reply here, no good reason for the overlap other than not having seen #9074 before opening this. No objection to proceeding with #9074; congrats on getting it merged.

Closing this one as superseded by #9074, which covers the same fix more thoroughly (Void handling, 4xx/5xx distinction, decode-failure handling) and is already merged.

@paoloantinori thanks for the two review passes, appreciated the time even though it didn't end up landing.

carlesarnal:
Skipping review — this PR was already closed by the contributor as superseded by #9074. No action needed.

om7057:
> Skipping review — this PR was already closed by the contributor as superseded by #9074. No action needed.

@carlesarnal Yes, agreed; would be more complete in my approach henceforth!

PR #9212, opened 2026-08-04
fix(mcp): ClassCastException in Utils.toQueryLabels due to Object[] t…

### LINKED ISSUE #8727: Bug: ClassCastException in MCP Utils.toQueryLabels due to Object[] to String[] cast
### Describe the bug

In `Utils.java` (lines 73-75), `toQueryLabels()` calls `Stream.toArray()` without a generator argument and casts the result to `String[]`:

` ` `java
return (String[]) rawLabels.entrySet().stream()
        .map(e -> e.getKey() + ":" + e.getValue())
        .toArray();
` ` `

`Stream.toArray()` without an argument returns `Object[]`. The explicit `(String[])` cast throws `ClassCastException` at runtime because Java arrays are not covariant for casting — `Object[]` cannot be cast to `String[]`.

### Steps to reproduce

1. Register an MCP tool with labels via the MCP server
2. Perform any operation that calls `toQueryLabels` (e.g. `search_artifacts` with label filters)
3. `ClassCastException: [Ljava.lang.Object; cannot be cast to [Ljava.lang.String;`

### Expected behavior

Label query should return `String[]` without runtime errors.

### Environment

- File: `mcp/src/main/java/io/apicurio/registry/mcp/Utils.java`, lines 73-75###

KumarADITHYA123:
## Description

`Utils.toQueryLabels()` throws `ClassCastException` at runtime when any MCP tool operation uses label-based filtering (e.g. `search_artifacts`, `search_versions`, `search_groups`).

**Root cause:** `Stream.toArray()` without a generator argument returns `Object[]`. The method casts the result to `String[]` via `(String[])`, which fails because Java arrays are not covariant for casting.

**Fix:** Pass `String[]::new` as the generator to `toArray()`, producing a properly-typed `String[]` directly. The redundant explicit cast is removed.

## Testing

Added `UtilsToQueryLabelsTest` with 11 test cases:
- Null input, empty JSON object, single label, multiple labels (order-independent)
- Return type is `String[]` not `Object[]` -- directly validates the fix
- Invalid JSON, JSON array, non-string values, empty string input
- Colons in values, empty string values

All 3 callers in `RegistryService.java` assign to `String[]` query parameters -- no regression.

Fixes #8727

### DIFF
--- a/mcp/src/main/java/io/apicurio/registry/mcp/Utils.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/Utils.java
@@ -70,9 +70,9 @@ public class Utils {
         }
         @SuppressWarnings("unchecked")
         var rawLabels = (Map<String, Object>) (Map<?, ?>) toRawLabels(jsonMap);
-        return (String[]) rawLabels.entrySet().stream()
+        return rawLabels.entrySet().stream()
                 .map(e -> e.getKey() + ":" + e.getValue())
-                .toArray();
+                .toArray(String[]::new);
     }
 
     public static <R> R handleError(Function0R<R> action) {
--- a/mcp/src/test/java/io/apicurio/registry/mcp/UtilsToQueryLabelsTest.java
+++ b/mcp/src/test/java/io/apicurio/registry/mcp/UtilsToQueryLabelsTest.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2026 Red Hat
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.apicurio.registry.mcp;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkiverse.mcp.server.ToolCallException;
+import jakarta.inject.Inject;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Unit tests for {@link Utils#toQueryLabels(String)}.
+ *
+ * <p>Verifies that the method correctly converts a JSON map of labels into
+ * a {@code String[]} of {@code "key:value"} entries suitable for the
+ * Apicurio Registry search API query parameters.</p>
+ */
+@QuarkusTest
+public class UtilsToQueryLabelsTest {
+
+    @Inject
+    Utils utils;
+
+    @Test
+    public void testNullInputReturnsNull() {
+        assertNull(utils.toQueryLabels(null));
+    }
+
+    @Test
+    public void testEmptyJsonObjectReturnsEmptyArray() {
+        String[] result = utils.toQueryLabels("{}");
+        assertNotNull(result);
+        assertEquals(0, result.length);
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


KumarADITHYA123:
@EricWittmann @carlesarnal  PTAL! when you have a chance.

paoloantinori (commented):
Correct, minimal fix: passing `String[]::new` to `toArray()` yields a typed `String[]` directly, eliminating the runtime `ClassCastException` that the no-arg `toArray()` plus `(String[])` cast caused on every label-filtering MCP call. The diagnosis is accurate (no-arg `toArray()` returns `Object[]`, which cannot be narrowed to `String[]` at runtime), and the new tests — including a direct return-type assertion — lock in the regression.

- `mcp/src/main/java/io/apicurio/registry/mcp/Utils.java:73,75` — The fix is type-safe: `.map(e -> e.getKey() + ":" + e.getValue())` produces a `Stream<String>`, so `.toArray(String[]::new)` returns `String[]` and the removed explicit cast is no longer needed. No behavior change beyond resolving the cast failure.
- `mcp/src/test/java/io/apicurio/registry/mcp/UtilsToQueryLabelsTest.java:52-56` — `testReturnTypeIsStringArray` (`assertInstanceOf(String[].class, result)`) directly validates the fix and would fail against the old code; combined with the null, empty, order-independent, invalid-JSON, and edge-value cases, coverage is solid.
- No caller changes are required: the three `RegistryService` call sites already assign the result to `String[]` query parameters, consistent with the PR description.
- Worth a glance in CI but not changed by this PR: `testNonStringValueThrowsToolCallException` (around line 67) assumes `toRawLabels` rejects non-string values upstream — the stream itself would stringify an `Integer` value (e.g. `"count:42"`) rather than throw — so that guard must live in `toRawLabels`. It is pre-existing behavior, just noting to confirm the test is exercising a real code path.

jsenko (approved):
Approving. Thanks @KumarADITHYA123 — and apologies for the wait, this has been sitting since August and the delay was entirely on our side.

This is an exemplary small fix: correct diagnosis, minimal change, and a test that genuinely proves it.

## Verified

**The fix is right.** `Stream.toArray()` returns `Object[]`, and `(String[]) someObjectArray` is an unconditional runtime failure — Java arrays are covariant for assignment but not for casting. `toArray(String[]::new)` builds the correctly-typed array directly and makes the cast unnecessary. Removing it rather than leaving it in place is the right call.

**Your tests actually catch the bug.** I reverted `Utils.java` to `main` while keeping your test class, and:

```
Tests run: 11, Failures: 0, Errors: 6
java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Ljava.lang.String;
    at io.apicurio.registry.mcp.Utils.toQueryLabels(Utils.java:80)
```

Six of the eleven fail without the fix, and all eleven pass with it. That is the thing I most want from a regression test and most often do not get — plenty of tests pass equally well against the broken code. `testReturnTypeIsStringArray` with `assertInstanceOf(String[].class, result)` is a particularly direct assertion of the actual defect.

**Everything else checks out:**

- `checkstyle:check -pl :apicurio-registry-mcp-server` — 0 violations
- DCO signed on your commit
- No merge conflicts
- I searched the rest of `mcp/`, `app/` and `common/` for the same `(T[]) …toArray()` pattern and found no other instances, so this was the only occurrence

Also worth noting the `@SuppressWarnings("unchecked")` above is still needed — it covers the `Map<?, ?>` cast, not the array conversion — so leaving it in place was correct.

## Nice touches

- Sorting before asserting in `testMultipleLabels`, with a comment explaining that `HashMap` ordering is not guaranteed. That is a flake that would have surfaced eventually in CI rather than locally.
- Covering colons in values, which matters here because the output format is itself `key:value` — that is exactly the edge case someone would miss.
- The PR description names the root cause precisely ("Java arrays are not covariant for casting") and confirms all three callers in `RegistryService` assign to `String[]`, so no caller needed changing.

I have set the 3.3.4 milestone on this PR and on #8727.

Thanks again — genuinely good work, and sorry it took us this long to look at it.

PR #9214, opened 2026-08-04
fix(ui): handle full handlebars syntax in prompt template highlighting (#9213)

### LINKED ISSUE #9213: fix(ui): PROMPT_TEMPLATE viewer highlighting mishandles handlebars block tags, dotted paths, and triple-stash
## Description

The PROMPT_TEMPLATE syntax highlighting in the UI (`highlightVariables` in `PromptTemplateViewer.tsx`) uses a regex that only handles a narrow subset of handlebars/mustache syntax. Several common constructs are mis-highlighted or not highlighted at all in the Template preview section.

### Steps to Reproduce

1. Create a PROMPT_TEMPLATE artifact whose template uses any of the following:
   - a closing block tag, e.g. `{{#if user}}Hi{{/if}}`
   - an `{{else}}` branch
   - a dotted path, e.g. `{{user.email}}`
   - a data variable, e.g. `{{#each items}}{{@index}}{{/each}}`
   - a triple-stash, e.g. `{{{raw}}}`
2. Open the artifact version Documentation tab and look at the Template preview.

Expected: all handlebars tokens are highlighted consistently — block helpers (`{{#if user}}`, `{{/if}}`, `{{else}}`) with the block style, variables with the variable style.

Actual:

- `{{/if}}`, `{{/each}}`, `{{/unless}}` closing tags are **not highlighted at all** (plain text), while their matching `{{#...}}` openers are highlighted — visually inconsistent.
- `{{else}}` is highlighted with the **variable** style instead of the block style.
- `{{{raw}}}` triple-stash is half-highlighted: rendered as plain `{` + highlighted `{{raw}}` + plain `}`.
- `{{user.email}}` dotted paths and `{{@index}}` data variables are not highlighted at all.

Example of the current output for `{{#if user}}Hi{{else}}Hi guest{{/if}}`:

- `{{#if user}}` → block style
- `{{else}}` → variable style (wrong)
- `{{/if}}` → plain text (not highlighted)

This is a UI-only bug; it does not affect rendering via the `/render` endpoint.

KartikSuryavanshi:
## What

Fixes syntax highlighting in the PROMPT_TEMPLATE viewer. Extracts the highlighting logic out of `PromptTemplateViewer.tsx` into a new `PromptTemplateViewer.utils.ts` (mirroring the existing `PromptTemplateTestPanel.utils.ts` pattern) with a proper handlebars tokenizer, and adds unit tests.

Closes #9213

## Why

The old regex-based highlighter only recognized `{{name}}` and a narrow `{{#if x}}` form. Common handlebars syntax was broken:

- Closing tags `{{/if}}`, `{{/each}}`, `{{/unless}}` were not highlighted at all
- `{{else}}` was styled as a variable instead of a block
- `{{{raw}}}` triple-stash was half-highlighted (stray `{` / `}` left as plain text)
- Dotted paths (`{{user.email}}`) and data variables (`{{@index}}`) were not matched

## Relationship to #8917

#8917 (merged) fixed the closing-tag case within the old regex. This PR supersedes that approach: it replaces the regex with a full tokenizer that handles closing tags **and** the remaining cases (`{{else}}` classification, triple-stash, dotted paths, `{{@index}}`, `{{!-- comments --}}`), and adds the test coverage that the regex approach lacks. The closing-tag behavior from #8917 is preserved (covered by the new unit tests).

## How

- `PromptTemplateViewer.utils.ts` — new `tokenizeTemplate()` that matches comments (`{{!-- --}}`), triple-stash (`{{{...}}}`), and regular tags (`{{...}}`) and classifies each token as plain / variable / block. `highlightVariables()` is now a thin wrapper over it.
- `PromptTemplateViewer.utils.test.ts` — 12 unit tests covering all the previously broken constructs plus regression cases.
- `PromptTemplateViewer.tsx` — imports `highlightVariables` from the new util.

## Verification

- `npx vitest run src/app/components/promptTemplate/` — 20 tests pass (8 existing + 12 new)
- `npx eslint src/app/components/promptTemplate/` — clean
- No changes to rendering behavior on the server side; UI-only.

Signed-off-by: Kartik Suryavanshi <158498247+KartikSuryavanshi@users.noreply.github.com>


paoloantinori (commented):
Overall this is a clean, correct refactor: the new `tokenizeTemplate()` handles every previously-broken construct, and I verified the regex + classifier by executing them against all 12 documented cases plus adversarial inputs (adjacent triple-stash, custom helpers, comments interleaved with text, repeated calls) with no regressions or correctness bugs. The extraction mirrors the existing `PromptTemplateTestPanel.utils.ts` pattern and correctly uses `React.createElement` rather than JSX, so the util compiles as a `.ts` file.

Observations:
- All targeted constructs tokenize correctly — closing tags (`{{/each}}`), `{{else}}`, triple-stash (`{{{raw}}}`), dotted paths (`{{user.email}}`), `{{@index}}`, and `{{!-- --}}` comments all classify as expected (`PromptTemplateViewer.utils.ts:10`, `PromptTemplateViewer.utils.ts:12-19`).
- No regression vs. the prior regex: the old matched set is preserved and the tokenizer only broadens coverage — e.g. `{{#custom arg}}` and arbitrary `{{/x}}` closers now classify as blocks where the old `if|unless|each|with`-only regex left them as plain text (`PromptTemplateViewer.utils.ts:12-19`).
- The shared module-level regex with the `g` flag is safe here: the `while` loop always runs to completion (re-running `tokenizeTemplate` produces correct results, confirmed empirically), and the minimum match length is 4 so there is no zero-length-match infinite-loop risk (`PromptTemplateViewer.utils.ts:10`, `PromptTemplateViewer.utils.ts:21-38`).
- Minor / explicitly out of scope: Handlebars single-bang comments (`{{! ... }}`) and the inverse shorthand (`{{^}}`) are classified as `variable` rather than `plain`/`block`; the PR scopes comment support to `{{!-- --}}`, so this reads as a known limitation rather than a defect (`PromptTemplateViewer.utils.ts:10`, `PromptTemplateViewer.utils.ts:30`).

KartikSuryavanshi:
@paoloantinori Thanks for the thorough review! Both observations are now addressed in 9884b1836:
- {{! ... }} single-bang comments — added \{\{![\s\S]*?\}\} to the tokenizer regex (ahead of the generic tag) and widened the comment check from {{!-- to {{!, so they now render as plain instead of variable.
- {{^}} inverse shorthand — classifyTag() now treats heads starting with ^ as block, so it's styled consistently with {{else}} rather than as a variable.
Added 3 unit tests covering the single-bang comment in isolation, a single-bang comment adjacent to a variable, and {{#if a}}A{{^}}B{{/if}} classifying {{^}} as a block.
Verification: npx vitest run src/app/components/promptTemplate/ — 23 tests pass (20 prior + 3 new), and npx eslint on the directory is clean.

paoloantinori:
Thanks @KartikSuryavanshi — re-read at HEAD 9884b18 and both points landed. The `{{!...}}` single-bang arm is now ahead of the generic tag in `HANDLEBARS_TAG` and the comment check widened from `{{!--` to `{{!`, so those tokenize as `plain`. `classifyTag` now treats `^`-prefixed heads as `block`, matching `{{else}}`. Looks consistent with the new unit tests. No further comments from me.

KartikSuryavanshi:
Hey @paoloantinori, thanks for the review! If you don't have any further comments, would you mind approving the PR? Or are we waiting for another review before it can be approved?

paoloantinori:
Thanks @KartikSuryavanshi. No further technical comments from my side: the tokenizer handles all the documented cases cleanly and the test coverage is solid.

I'll leave the `/accept` to another maintainer. Once it rebases past the unrelated UI CI failure in #9227 this is ready to land.

KartikSuryavanshi:
@EricWittmann Can you please approve the workflow for the checks to run?

PR #9218, opened 2026-08-04
feat: add native Protobuf Kafka Connect converter

### LINKED ISSUE #3061: Protobuf Converter support
For Kafka Connect, Confluent provides a `ProtobufConverter`. Could we be seeing something similar for Apicurio?

kartikangiras:
### Description 
- Add a new Apicurio Protobuf Kafka Connect converter built on the existing SerdeBasedConverter pattern. 
- This introduces a Protobuf-specific Connect data mapper, registers the converter for Kafka Connect service discovery, and wires in the Protobuf Kafka serdes so Connect records can be converted to and from Protobuf messages. 
- Includes focused unit tests covering the new Protobuf mapping logic and converter registration.

### Issue
fixes #3061 

### Command to Verify
```yaml
./mvnw test -pl utils/converter
```


utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
```diff
@@ -0,0 +1,454 @@
...
+                    break;
+                case BYTES:
+                    field.setType(DescriptorProtos.FieldDescriptorProto.Type.TYPE_BYTES);
+                    break;
+                case STRUCT:
+                    field.setType(DescriptorProtos.FieldDescriptorProto.Type.TYPE_MESSAGE)
+                            .setTypeName("." + packageName() + "." + addMessage(preferredMessageName, schema));
+                    break;
+                default:
+                    throw new IllegalArgumentException("Unsupported Connect schema type: " + schema.type());
```
paoloantinori, line 669: applyFieldType has no case for ARRAY or MAP, so any nested collection whose element or value is itself a collection (array-of-array, array-of-map, map-of-array, map-of-map) falls through to this default and throws IllegalArgumentException during fromConnectData. These shapes occur in real Connect schemas (e.g. a map whose value is a list). Either generate an intermediate wrapper message for nested collection types or reject them explicitly with a documented, typed exception rather than the generic 'Unsupported Connect schema type'.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
```diff
@@ -0,0 +1,454 @@
...
+
+    private DynamicMessage toMessage(Descriptors.Descriptor descriptor, Schema schema, Object value) {
+        if (!(value instanceof Struct)) {
+            throw new IllegalArgumentException("Expected Struct for schema " + schema.name());
+        }
+        Struct struct = (Struct) value;
+        DynamicMessage.Builder builder = DynamicMessage.newBuilder(descriptor);
+        for (Field connectField : schema.fields()) {
+            Object fieldValue = struct.get(connectField);
+            if (fieldValue == null) {
```
paoloantinori: Null Connect values are skipped here, and the generated proto3 scalar fields use LABEL_OPTIONAL without proto3 `optional` presence (no setProto3Optional), so an unset scalar reads back as the type default (0 for ints, "" for string, false for bool, empty for bytes) on the reverse path rather than null. null-vs-default is not round-tripped. Consider emitting proto3 optional (explicit presence) for nullable Connect fields, or document this fidelity loss explicitly.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
```diff
@@ -0,0 +1,454 @@
...
+                continue;
+            }
+            Descriptors.FieldDescriptor protoField = descriptor.findFieldByName(connectField.name());
```
paoloantinori: Proto field names are registered under the sanitized name (sanitizeIdentifier in toField), but this lookup uses the raw Connect field name. For field names containing characters outside [A-Za-z0-9_] — common in real schemas such as 'user-id', 'first.name', or quoted column names — findFieldByName returns null and fromConnectData throws 'Missing Protobuf field'. Resolve fields by the same sanitized name used at registration (or keep a raw→sanitized name map) so generation and population agree.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
```diff
@@ -0,0 +1,454 @@
...
+        Objects.requireNonNull(schema, "schema must not be null");
+        if (value == null) {
+            return null;
+        }
+        if (schema.type() != Schema.Type.STRUCT) {
+            throw new IllegalArgumentException("Top-level Protobuf converter schema must be a struct");
+        }
+
+        try {
+            Descriptors.Descriptor descriptor = descriptorFor(schema);
```
paoloantinori: descriptorFor rebuilds the entire FileDescriptorProto/FileDescriptor via reflection on every record (there is no cache field anywhere in ProtobufData). Converters run per Kafka record, so for non-trivial throughput this descriptor construction dominates cost. Memoize the Descriptor keyed by the Connect Schema — MessageBuilderContext already keys messages by Schema identity, so the same keying works for a top-level descriptor cache.

paoloantinori (commented):
The converter mirrors the existing SerdeBasedConverter pattern cleanly and the ProtobufData happy-path round-trips (struct, nested struct, array, map, bytes) are covered, but the Connect→Protobuf mapping has several correctness/fidelity gaps that will surface on real-world schemas: nested collections throw, nulls collapse to proto3 defaults, special-character field names fail lookup, INT8/INT16 types are widened away, and descriptors are rebuilt per record.

Observations:
- Nested collections (array/map of array or map) reach applyFieldType's default branch and throw IllegalArgumentException during fromConnectData — ProtobufData.java:413 (reached via toField at :369 and addMapEntry's value field).
- null Connect values are skipped at encode and proto3 scalar fields have no presence tracking, so they round-trip to the type default (0/""/false/empty) rather than null — ProtobufData.java:212.
- Field names are sanitized at registration (sanitizeIdentifier, toField) but looked up by the raw Connect name, so non-identifier field names throw 'Missing Protobuf field' — ProtobufData.java:215 vs :364.
- descriptorFor rebuilds the full FileDescriptor via reflection on every record with no caching — ProtobufData.java:41.
- INT8/INT16 are mapped to TYPE_INT32 in both directions and the original narrower type is not recovered on round-trip — ProtobufData.java:388 and :80.
- configure() injects raw, unconfigured ProtobufKafkaSerializer/Deserializer via putIfAbsent; worth confirming SerdeBasedConverter actually configures them with the registry, else they bypass registry wiring — ProtobufConverter.java:37-38.
- Coverage is limited to ServiceLoader registration; ProtobufConverter itself and the failure modes above are untested — ProtobufConverterTest.java:13.

utils/converter/src/test/java/io/apicurio/registry/utils/converter/ProtobufConverterTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.ServiceLoader;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class ProtobufConverterTest {
+
+    @Test
+    public void testConverterIsRegisteredAsServiceProvider() {
```
paoloantinori, line 35: This test only asserts ServiceLoader registration; ProtobufConverter itself (configure + fromConnectData→toConnectData through the Protobuf serde) is never exercised, and the edge cases above (nulls, nested collections, INT8/INT16 widening, special-character field names) have no coverage. Add a converter-level round-trip test and explicit cases for the failure modes so regressions in the mapping logic are caught.

kartikangiras: Added the required test.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
kartikangiras: Adding the support for Array and Map is better suited here, as it will increase the boundary use case of the converter.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
kartikangiras: Resolved. Using the same sanitized registration names now.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
```diff
@@ -0,0 +1,612 @@
...
+
+    private Schema toConnectFieldSchema(Descriptors.FieldDescriptor field) {
+        if (field.isMapField()) {
+            Descriptors.Descriptor entry = field.getMessageType();
+            Schema keySchema = toConnectFieldSchema(entry.findFieldByName("key"));
+            Schema valueSchema = toConnectFieldSchema(entry.findFieldByName("value"));
+            return SchemaBuilder.map(keySchema, valueSchema).optional().build();
+        }
+        if (field.isRepeated()) {
+            return SchemaBuilder.array(toConnectScalarSchema(field)).optional().build();
```
paoloantinori, line 298: Repeated fields decode through toConnectScalarSchema(field), which for the synthetic wrapper message created by addWrapperMessage returns toConnectSchema(wrapper), i.e. a struct with an items field, rather than unwrapping back to the original array<T> or map<...,array<T>>. As written, map<string,array<string>> decodes to map<string,struct{items:array<string>}>. There is no logic anywhere on the decode side that recognizes the Wrapper{repeated items} shape and unwraps it, which contradicts the class Javadoc claim that nested collections are supported. Either add wrapper detection on decode, or correct the Javadoc and add a test that calls toConnectData and asserts the decoded schema equals the original.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
```diff
@@ -0,0 +1,612 @@
...
+                    .setName(wrapperName);
+            DescriptorProtos.FieldDescriptorProto.Builder itemsField =
+                    DescriptorProtos.FieldDescriptorProto.newBuilder()
+                            .setName("items")
+                            .setNumber(1)
+                            .setLabel(DescriptorProtos.FieldDescriptorProto.Label.LABEL_REPEATED);
+            applyFieldType(itemsField, elementSchema, wrapperName + "Element");
+            wrapper.addField(itemsField);
+            file.addMessageType(wrapper);
+            messageNames.put(elementSchema, wrapperName);
```
paoloantinori: messageNames is the same map addMessage consults to dedup struct messages (messageNames.get(schema) at the top of addMessage). Writing messageNames.put(elementSchema, wrapperName) here means that if the same struct schema is later used as a standalone field, addMessage returns the wrapper name and the field is typed as the wrapper instead of the struct. This is reachable with a schema like struct{ a: array<array<struct S>>, b: S }, and Connect struct schemas compare structurally so the collision is not hypothetical. Consider a separate map for wrapper bookkeeping, or drop this put if nothing reads it.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
```diff
@@ -0,0 +1,612 @@
...
+            return SchemaAndValue.NULL;
+        }
+        Schema schema = toConnectSchema(message.getDescriptorForType());
+        return new SchemaAndValue(schema, toConnectValue(schema, message));
+    }
+
+    public Schema toConnectSchema(Descriptors.Descriptor descriptor) {
+        SchemaBuilder builder = SchemaBuilder.struct().name(connectName(descriptor.getFullName()));
+        for (Descriptors.FieldDescriptor field : descriptor.getFields()) {
+            builder.field(field.getName(), toConnectFieldSchema(field));
```
paoloantinori: On decode the schema is rebuilt from field.getName(), which is the sanitised proto field name. A Connect field named user-id or event.type therefore round-trips as user_id or event_type, and the original name is never restored. There is no json_name hint stored for names, unlike the INT8/INT16 hints, so the loss is silent. testSpecialCharFieldNameRoundTrips only asserts the decoded value is non-null and does not check field names, so this is not caught. Either preserve the raw name in a field option and restore it here, or document the rename as a known limitation.

utils/converter/src/main/java/io/apicurio/registry/utils/converter/protobuf/ProtobufData.java
```diff
@@ -0,0 +1,612 @@
...
+        }
+        return builder.build();
+    }
+
+    /**
+     * Ensures that both the descriptor and name map for {@code schema} are present in their
+     * respective caches. Uses {@code computeIfAbsent} so that concurrent calls for the same schema
```
paoloantinori: The Javadoc states this uses computeIfAbsent so concurrent calls only build once, but the body uses descriptorCache.containsKey followed by putIfAbsent, so two concurrent calls for the same schema can each build the FileDescriptor. The build is deterministic so this is not a correctness bug, just redundant work plus a misleading comment, and Kafka Connect invokes converters across concurrent tasks. Either move the build into computeIfAbsent, or correct the comment.

paoloantinori (commented):
Two round-trip correctness gaps should be resolved before this merges. Nested Connect collections (array-of-array, map-of-array) decode back as a different Connect type than they were encoded from, and Connect field names containing non-identifier characters are silently renamed on the decode side. The converter otherwise reuses the existing SerdeBasedConverter pattern cleanly, and the INT8/INT16 and null-presence handling is thoughtful.

- Nested collections do not round-trip (ProtobufData.java:235, 267-268). The decode path never unwraps the synthetic Wrapper{items} message that the encode path adds for nested collections, so map<string,array<string>> decodes to map<string,struct{items:array<string>}> and array<array<T>> to array<struct{items:array<T>}>. This contradicts the class Javadoc (ProtobufData.java:35-37), which states nested collections are supported. The tests named ...RoundTrips only assert that fromConnectData does not throw and never call toConnectData, so the regression is not caught.
- addWrapperMessage corrupts the struct-dedup map (ProtobufData.java:571). messageNames.put(elementSchema, wrapperName) writes into the same map addMessage uses to dedup struct messages, so a later standalone field whose schema structurally equals the element gets typed as the wrapper instead of the struct. Reachable with struct{ a: array<array<struct S>>, b: S }.
- Field names are not preserved on decode (ProtobufData.java:95). The decoded schema is rebuilt from the sanitised proto field name, so user-id becomes user_id and event.type becomes event_type, with no json_name hint stored to recover the original the way INT8/INT16 are recovered. testSpecialCharFieldNameRoundTrips only asserts the value is non-null.
- Easy win: the ensureCached Javadoc (ProtobufData.java:102) claims computeIfAbsent, but the body uses containsKey plus putIfAbsent, so concurrent Connect tasks can redundantly build the descriptor. Either switch to computeIfAbsent or correct the comment.

kartikangiras:
Hi @paoloantinori ,
I iterated over your feedback. This is the updated status of the review:

- Now the Nested collections round-trip actually detects and unwraps synthetic wrapper messages back to array<T>
- Wrapper map corruption names tracked in a separate Set, no longer polluting the struct dedup map
- Special-char field names are preserved with raw names stored in json_name with __raw recovered on decode
- ensureCached comment fixed with Javadoc now matches the actual containsKey + putIfAbsent implementation

paoloantinori:
Thanks for the thorough follow-up. All four points landed at this head:

- Nested collections now unwrap correctly: `isWrapperMessage` (ProtobufData.java:276) detects the synthetic `Wrapper{repeated items}` and the decode path at lines 326 and 442 transparently collapses it back to array/map.
- Wrapper message names tracked in a separate set, so the struct dedup map stays clean.
- Special-char field names preserved via `__raw:` prefix in `json_name` and recovered on decode.
- `ensureCached` comment now matches the `containsKey` + put flow.

The new round-trip tests (array-of-array, map-of-array, struct-with-array-and-sibling-struct) pin the cases that were silently mistyping before. From my side this is good to merge.

kartikangiras:
Thanks @paoloantinori for your constructive feedback.
I took a closer look into the converter and found some boundary edge cases that should be covered as well and added a few regression tests for them. 
Here are few of the boundary edge cases tested:

- INT8/INT16 min/max limits do not suffer from sign-extension corruption when encoded as `int32` on the wire. 
- Also ensure empty arrays and maps correctly round-trip as empty collections rather than collapsing to `null`.
- validate that FLOAT32/FLOAT64 special values like null, non definite preserve their exact states and precision.

wdys about these ones

paoloantinori:
Thanks for digging into the boundary cases. One thing to flag before I re-read: the PR head is still `d394473e5d618fd2bf74247c0c0f7774f576eb1b`, the same commit where I previously confirmed the four round-trip points had landed. I don't see any new commits with the INT8/INT16 sign-extension, empty-array/map, or FL regression tests you describe, so I suspect they're still local. Push them and I'll take a look. If they're already there and I missed them, point me at the commit SHA.

kartikangiras:
Hi @paoloantinori ,
I hope they are visible now as I forced push again. This is the specific commit SHA `e77315ab2884476842905a6607dd2cb220e3696d`.
Thanks

paoloantinori:
Thanks for the force-push, the new head `e77315ab` is visible now.

The boundary regression tests are all there: INT8/INT16 min/max values, empty array/map/array-of-array round-trips, and the NaN/Infinity cases for FLOAT32/FLOAT64. Good coverage for the int32 sign-extension path on the narrow types and for the empty-collection decode path.

The four earlier points are all still addressed at this head too (nested-collection unwrap via `isWrapperMessage`, INT8/INT16 type preservation, null optional scalars, special-character field names).

No further concerns from my side.

kartikangiras:
Hi @EricWittmann ,
This PR is good to go for the next phase of approval and merge. It has been approved by Paolo.
Could you PTAL into this.
Thanks

EricWittmann (commented):
Thanks for this contribution — the code follows the existing `AvroConverter` pattern well and the test coverage is solid. Two items to address:

### 1. Add a test for the unsupported nested-map case

The `MAP` case in `toProtoValue` throws `IllegalArgumentException("Nested map-in-collection is not supported...")`, which is a reasonable limitation. However, there's no test that exercises this path. Please add a test that constructs a schema with a nested map (e.g. `array<map<string, string>>` or `map<string, map<string, string>>`) and asserts the expected exception message. This way the behavior is documented in the test suite and won't silently change in a future refactor.

### 2. Add a null-value test at the converter level

The `ProtobufConverterTest` tests exercise `ProtobufData` directly but never pass a null value through `ProtobufConverter.applySchema()`. Since `fromConnectData` returns `null` for null input, it would be good to add a test that verifies this null flows correctly through `SerdeBasedConverter`'s pipeline (or throws a clear error if it's not supported). This closes a gap between the unit-level and converter-level test coverage.

kartikangiras:
Hi @EricWittmann ,
I have added both the test edge cases reviewed by you. 
Thanks.

EricWittmann (commented):
One more item: the `ensureCached` method in `ProtobufData` uses `containsKey` + `putIfAbsent`, which means two concurrent threads can both build the descriptor for the same schema. Paolo flagged this as non-blocking since the build is deterministic, but since we're here let's clean it up before merge.

Please switch to `computeIfAbsent` so the descriptor is only built once per schema key. Something like:

```java
descriptorCache.computeIfAbsent(schema, s -> {
    // build and return the Descriptor
});
```

You'll need to do the same for `nameMapCache` (or combine both into a single cached object). The Javadoc already claims `computeIfAbsent` semantics, so this would make the code match the documentation.

kartikangiras:
> One more item: the `ensureCached` method in `ProtobufData` uses `containsKey` + `putIfAbsent`, which means two concurrent threads can both build the descriptor for the same schema. Paolo flagged this as non-blocking since the build is deterministic, but since we're here let's clean it up before merge.
> 
> Please switch to `computeIfAbsent` so the descriptor is only built once per schema key. Something like:
> 
> ```java
> descriptorCache.computeIfAbsent(schema, s -> {
>     // build and return the Descriptor
> });
> ```
> 
> You'll need to do the same for `nameMapCache` (or combine both into a single cached object). The Javadoc already claims `computeIfAbsent` semantics, so this would make the code match the documentation.

Sure I'm working on it.

carlesarnal:
Good work — the converter follows the AvroConverter pattern cleanly and the test coverage is thorough. Two items to address:

1) ensureCached still uses containsKey + putIfAbsent (outstanding from Eric's review, acknowledged in progress).

2) INT8/INT16 type hints missing in toMapField() and addWrapperMessage(). These methods call applyFieldType() but skip applyJsonName(), so MAP<STRING, INT8> silently round-trips as MAP<STRING, INT32>, and ARRAY<ARRAY<INT8>> inner elements lose their INT8 type. Fix by adding an applyJsonName() call after applyFieldType() in both methods. A round-trip test for MAP<STRING, INT8> would pin this.

kartikangiras:
Hi @EricWittmann ,
I have made all the changes reviewed by you, paolo and carles.
Thanks.

EricWittmann (commented):
Thanks @kartikangiras for this great contribution! The Protobuf converter fills a long-standing feature gap and the implementation is clean and thorough. Really appreciate your responsiveness to all the review feedback -- merged! 🎉

kartikangiras:
Thanks @EricWittmann, @paoloantinori, @carlesarnal for your vaIuable feedbacks to iterate over this. I really had a great time implementing this feature, technically I have implemented 2 Protobuf Kafka converter one for milvus and now for Apicurio. haha😁

PR #9222, opened 2026-08-04
fix(ui): expose Agent Skill examples in editor and viewer

### LINKED ISSUE #9201: [UI] Agent Card Skill Editor and Viewer are missing the "examples" array
**Registry Version:** main (3.3.2-SNAPSHOT)

**Persistence Type:** in-memory

The `AgentSkill` model supports an `examples: string[]` field, and the backend correctly stores and returns this data. However, the current Agent Card UI does not expose this field.

The issue affects both the editor and the documentation view:

- `SkillEditor.tsx` provides no way to create or edit `examples`.
- `AgentCardSkills.tsx` does not render the `examples` array even when it is present in the artifact data.

As a result, valid skill metadata is silently omitted from the UI.

---

## Environment

- **Operating System:** Windows 11
- **Browser:** Google Chrome
- **Backend:** `..\mvnw.cmd quarkus:dev`
- **Frontend:** `npm run dev`
- **Persistence:** in-memory

---

## Artifact Used

The following Agent Card was created through the Registry REST API:

```json
{
  "name": "Test Agent",
  "skills": [
    {
      "id": "weather",
      "name": "Weather",
      "description": "Gets weather",
      "examples": [
        "What is the weather?"
      ]
    }
  ]
}
```

---

## Steps to Reproduce

1. Create an Agent Card artifact using the JSON above through the Registry REST API.
2. Open the Registry UI.
3. Navigate to **Explore**.
4. Open the created artifact.
5. Click **Version 1**.
6. Open the **Documentation** tab.
7. Inspect the **Skills** section.

---

## Expected Behavior

The UI should expose the `examples` field consistently.

### Viewer

The Documentation view should render the examples associated with each skill, for example as PatternFly Labels, chips, or a simple list.

### Editor

The Skill Editor should provide a way to add, edit, and remove examples when creating or editing a skill.

---

## Actual Behavior

The UI currently displays:

- Skill Name
- Skill ID
- Description

but omits:

- Examples

In addition, the Skill Editor provides no mechanism for creating or modifying examples.

---

## API Verification

The backend correctly stores and returns the `examples` array.

Example response:

```json
"skills": [
  {
    "id": "weather",
    "name": "Weather",
    "examples": [
      "What is the weather?"
    ]
  }
]
```

This confirms that the issue is limited to the UI.

---

## Frontend Investigation

The following components were inspected:

### `SkillEditor.tsx`

Provides inputs for:

- ID
- Name
- Description
- Tags

but does not expose the `examples` field.

### `AgentCardSkills.tsx`

Renders:

- ID
- Name
- Description

but contains no rendering logic for `skill.examples`.

---

## Repository Verification

A repository search confirms that the backend schema defines `examples: string[]` for `AgentSkill`, but the field is not referenced by the Agent Card UI components responsible for editing or displaying skills.

---

## Proposed Fix

Updating both UI components:

### `SkillEditor.tsx`

Adding an input section for managing the `examples` array.

### `AgentCardSkills.tsx`

Rendering the examples beneath each skill using existing PatternFly components (such as `LabelGroup` and `Label`) or another appropriate presentation.

---

## Files Involved

- `ui/ui-app/src/app/components/agentCard/SkillEditor.tsx`
- `ui/ui-app/src/app/components/agentCard/AgentCardSkills.tsx`

---

## Screenshot

The screenshot below shows that the Skill information is displayed, but the `examples` field is omitted even though it exists in the stored artifact.

Image

###

I have investigated this issue locally and already have a proposed solution for it. If this is considered a valid issue, I'd be happy to work on a fix and submit a PR. If appropriate, could this issue please be assigned to me?

nishant-mazumder:
Fixes #9201.

The backend already supports the `examples` field for `AgentSkill` and correctly stores and returns it through the Registry API. However, the Agent Card UI did not expose this information.

Specifically:

- `SkillEditor` provided no way to create or edit skill examples.
- `AgentCardSkills` did not render the `examples` array even when it was present in the artifact.

This change exposes the `examples` field throughout the Agent Card UI by:

- adding support for creating, editing, and removing examples in `SkillEditor`
- displaying skill examples in the Agent Card documentation view using existing PatternFly components
- following the project's preferred conditional rendering style with the shared `<If>` component

### Before

The backend stored and returned:

```json
"examples": [
  "What is the weather?"
]
```

However:

- the Skill Editor had no input for managing examples
- the Documentation view omitted the examples completely

Screenshot 2026-08-03 235430

### After

The Agent Card UI now:

- allows examples to be added, edited, and removed in the Skill Editor
- displays the examples in the Documentation view for each skill

Screenshot 2026-08-04 165706

### Testing

- Reproduced the issue locally.
- Verified that the backend stores and returns the `examples` array.
- Verified that examples can now be created and edited through the UI.
- Verified that the Documentation view correctly renders the examples.
- Ran `npm run lint` successfully.
- Ran `npm test` successfully.


ui/ui-app/src/app/components/agentCard/SkillEditor.tsx
```diff
@@ -150,9 +165,9 @@ export const SkillEditor: FunctionComponent<SkillEditorProps> = (props: SkillEdi
                             </Button>
                         </InputGroupItem>
                     </InputGroup>
-                    {editingSkill.tags && editingSkill.tags.length > 0 && (
+                    <If condition={editingSkill.tags !== undefined && editingSkill.tags.length > 0}>
```
paoloantinori: This guard was previously `editingSkill.tags && editingSkill.tags.length > 0`, which short-circuits safely on `null`. The new `editingSkill.tags !== undefined && editingSkill.tags.length > 0` is null-unsafe: `null !== undefined` evaluates to `true`, so if `tags` is ever `null` (e.g. the API serializes the field as `null` instead of omitting it), `null.length` throws and crashes the editor render. The examples `<If>` a few lines down has the same issue. Safer and shorter: `editingSkill.tags?.length > 0` (and `editingSkill.examples?.length > 0`), which also makes the `(field || []).map()` inside the block redundant.

paoloantinori (commented):
Solid, well-scoped change that mirrors the existing tags UI and renders examples cleanly in the documentation view. One null-safety regression introduced by the `<If>` refactor is worth fixing before merge; the rest are minor easy wins.

- `SkillEditor.tsx:168` — the tags guard changed from the null-safe `editingSkill.tags && editingSkill.tags.length > 0` to `editingSkill.tags !== undefined && editingSkill.tags.length > 0`. `null !== undefined` is `true`, so if `tags` is ever `null` (an API response that serializes the field as `null` rather than omitting it), `null.length` throws and crashes the editor render; the new examples block (`SkillEditor.tsx:203`) copies the same pattern. Safer and shorter: `editingSkill.tags?.length > 0` / `editingSkill.examples?.length > 0`, which also makes the inner `(field || []).map()` redundant.
- `SkillEditor.tsx:111` — `handleRemoveExample` filters by value, so duplicate examples both vanish when one label is closed; removing by index would also align with the existing `key={index}`.
- `AgentCardSkills.tsx:63` vs `SkillEditor.tsx:168` use different null-check idioms for the same concept; prefer one (the optional-chaining form).
- No tests were added for `handleAddExample`/`handleRemoveExample` (trim guard, empty list, duplicate) — worth covering if these components are unit-tested here.
- Inline `style={{ marginTop }}` next to `className` (`AgentCardSkills.tsx:67`, `SkillEditor.tsx:204`) matches the existing tags UI, so it's a pre-existing pattern rather than a new inconsistency.

ui/ui-app/src/app/components/agentCard/SkillEditor.tsx
```diff
@@ -97,6 +99,19 @@ export const SkillEditor: FunctionComponent<SkillEditorProps> = (props: SkillEdi
...
+    const handleAddExample = (): void => {
+        if (!editingSkill || !newExample.trim()) return;
+        const examples = [...(editingSkill.examples || []), newExample.trim()];
+        setEditingSkill({ ...editingSkill, examples });
+        setNewExample("");
+    };
+
+    const handleRemoveExample = (exampleToRemove: string): void => {
+        if (!editingSkill) return;
+        const examples = (editingSkill.examples || []).filter(e => e !== exampleToRemove);
```
paoloantinori: `handleRemoveExample` removes by value, so if two examples share the same text, closing one label removes both. Consider removing by index — pass `index` from `onClose` and filter `(_, i) => i !== index` — which also matches the `key={index}` already in use. (The pre-existing `handleRemoveTag` likely has the same shape, so this may be worth a follow-up for consistency.)

ui/ui-app/src/app/components/agentCard/SkillEditor.tsx
nishant-mazumder: I've updated both components to use the ?.length > 0 optional chaining syntax as suggested

ui/ui-app/src/app/components/agentCard/SkillEditor.tsx
nishant-mazumder: Updated the removal logic to remove by index so duplicate values are handled correctly. I also applied the same approach to handleRemoveTag for consistency.

carlesarnal:
### Post-merge review

I reviewed the final state of this PR. All of @paoloantinori's feedback has been addressed cleanly:

- **Null-safety** — The tag/example guards now use optional chaining (`editingSkill.tags?.length` / `editingSkill.examples?.length`), which handles both `null` and `undefined` correctly. The `<If>` conditions in `SkillEditor.tsx` use `(field?.length ?? 0) > 0`, and the early-return guards in `AgentCardSkills.tsx` use `!field?.length`. Both are safe.
- **Remove-by-index** — Both `handleRemoveTag` and `handleRemoveExample` now filter by index, so duplicate values are handled correctly. The `onClose` callbacks pass the index from the `.map()` iteration, which also aligns with the `key={index}` pattern.

The new `renderSkillExamples` function in `AgentCardSkills.tsx` mirrors the existing `renderSkillTags` structure exactly — same guard pattern, same `LabelGroup` rendering, just with `color="green"` for visual distinction. The examples form group in `SkillEditor.tsx` likewise mirrors the tags form group (input + Enter key handler + label list with close buttons).

One small nit for a possible follow-up: `emptySkill` (SkillEditor.tsx:55) initializes `tags: []` but not `examples: []`. Not a bug since the code handles `undefined` throughout, but adding `examples: []` would be more consistent with the tags initialization.

Overall, well-scoped and clean change.

PR #9226, opened 2026-08-04
fix(ui): include all artifact types in fallback list to prevent silent drop on API failure

### LINKED ISSUE #9225: fix(ui): artifact types fallback list is missing 5 newer types

The UI keeps a hardcoded fallback list of artifact types in `ui/ui-app/src/services/useArtifactTypesService.ts`, used when the `admin.getArtifactTypes()` API call fails. That list has drifted out of sync with the `ArtifactTypes` class in the same file. Five types are missing from the fallback:

- `THRIFT`
- `AGENT_CARD`
- `MCP_TOOL`
- `MODEL_SCHEMA`
- `PROMPT_TEMPLATE`

The `getArtifactTypes` call can fail in normal usage (a non-admin user hitting a permission check, a temporary backend hiccup, network issues). When it does, the Create Artifact dropdown silently drops the ability to create any of those 5 types, and the user gets no indication anything is missing.

### Steps to Reproduce

1. Start the UI and backend.
2. Stop the backend so `admin.getArtifactTypes()` fails.
3. Refresh the UI and open the Create Artifact modal.
4. Look at the Type dropdown.

### Expected vs Actual

**Expected:** the dropdown lists every artifact type the UI supports, including the 5 above.

**Actual:** only the 10 older types show up. The 5 newer types cannot be selected.

Shubb07:
Fixes #9225

The hardcoded fallback in `useArtifactTypesService.ts` had drifted from the `ArtifactTypes` constants and was missing THRIFT, AGENT_CARD, MCP_TOOL, MODEL_SCHEMA, and PROMPT_TEMPLATE. When `admin.getArtifactTypes()` fails, those 5 types silently disappeared from the Create Artifact dropdown.

This adds an `ArtifactTypes.all()` method as a single source of truth for the known types, and uses it for the fallback instead of a separate hardcoded array, so new types won't drop out again next time one is added.

**How I tested:** temporarily forced the catch branch (threw before the API call) and confirmed the Create Artifact dropdown then showed all 15 types, including the 5 that were missing before.

### DIFF
--- a/ui/ui-app/src/services/useArtifactTypesService.ts
+++ b/ui/ui-app/src/services/useArtifactTypesService.ts
@@ -127,6 +127,19 @@ export class ArtifactTypes {
         return title;
     }
 
+    // All artifact types the UI knows about. Used as a fallback when the backend
+    // list of types can't be fetched. Keep in sync with the constants above so that
+    // new types don't silently drop out of the Create Artifact dropdown on API failure.
+    public static all(): string[] {
+        return [
+            ArtifactTypes.AVRO, ArtifactTypes.PROTOBUF, ArtifactTypes.JSON,
+            ArtifactTypes.OPENAPI, ArtifactTypes.ASYNCAPI, ArtifactTypes.GRAPHQL,
+            ArtifactTypes.KCONNECT, ArtifactTypes.WSDL, ArtifactTypes.XSD,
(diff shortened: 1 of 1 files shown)


EricWittmann:
Thanks @Shubb07 — good fix. Using ArtifactTypes.all() as a single source of truth prevents this from drifting again. Merged!

Shubb07:
Thanks Eric, really appreciate it. glad the single-source-of-truth approach made sense.

PR #9229, opened 2026-08-04
feat(cli): add Windows support

### LINKED ISSUE #8646: CLI: Support for Windows
## Parent Epic

## Description
The CLI currently targets macOS and Linux. Add Windows support covering credential storage, path handling, and native binary distribution.

## Areas to Address
- **Credential storage**: Replace macOS Keychain / Linux secret-tool with Windows Credential Manager
- **File permissions**: POSIX file permissions (used by `FileCredentialProvider`) don't apply on Windows
- **Native binary**: GraalVM native image build for Windows (`.exe`)
- **Path handling**: Ensure all file paths work with Windows separators
- **Install/update**: Adapt install and auto-update scripts for Windows

## Acceptance Criteria
- CLI can be built and run on Windows
- Credential storage works on Windows
- Install and update workflows work on Windows
- CI builds produce a Windows binary

Vidhaankhare16:
## Summary

Adds Windows (x86_64) as a supported platform for the CLI. Until now `PlatformUtils.detectOsClassifier()` threw `UnsupportedOperationException` on Windows, so nothing worked before that point; the credential providers shelled out to `security`/`secret-tool`, the credentials-file fallback silently left the file with its inherited ACL, install and update relied on symlinks and a shell rc file, and no `windows-x86_64` ZIP was produced.

This adds the Windows branches for each of those, an `acr.cmd` launcher and `windows-x86_64` ZIP, a Windows CI job, and the documentation updates. Every change is an additive `isWindows()` branch or a new file — Linux and macOS behaviour is unchanged.

Closes #8646 (part of #7162).

### Platform detection (`PlatformUtils`)
- Add the `windows` OS classifier and `isWindows()`. The check matches the whole `windows` token rather than `win`, since `darwin` contains it.
- The resulting platform classifier is `windows-x86_64`, which is also what `os-maven-plugin` reports as `${os.detected.classifier}` on Windows, so the ZIP name and `Update.downloadVersion()`'s URL construction line up with no further change.

### Credential storage (`WindowsCredentialProvider`, `CredentialProviderProducer`)
- New provider backed by the Windows Credential Manager, storing generic credentials via `CredWriteW` / `CredReadW` / `CredDeleteW` under the target name `<serviceName>/<account>`, so entries are scoped to the current Windows user.
- `jna-platform` ships no Credential Manager bindings, so only the four advapi32 functions the CLI needs are mapped by hand. JNA is declared explicitly at 5.19.1 in `cli/pom.xml` because the Quarkus BOM manages 5.8.0, which predates the native-image support this relies on.
- Contract preserved from the existing providers: `retrieve` returns `null` for `ERROR_NOT_FOUND` (this is what drives the `--allow-unsafe-credential-storage` file fallback in `CredentialStore`) and throws for any other error, `delete` is idempotent, and the returned credential is always released with `CredFree` in a `finally` block.
- Secret hygiene: the blob is written as UTF-16LE, the native `Memory` is cleared and the intermediate `byte[]` zeroed after each call, and failures carry only the Win32 error code — never the secret. Secrets larger than `CRED_MAX_CREDENTIAL_BLOB_SIZE` (2560 bytes, per `wincred.h`) are rejected with a clear message; since the blob is UTF-16LE that is 1280 characters, and both sides of the boundary are covered by tests. That cap is a Credential Manager constraint with no macOS or Linux equivalent, so it is recorded in the README — though the CLI only ever stores a password or an OAuth2 client secret, never a token, so it is not reached in practice.
- An entry that exists but holds an empty secret reads back as the empty string, with `null` reserved for an entry that is not there, which is how the file fallback and the macOS provider already behave.
- A failure to load JNA is reported as a `CredentialStoreException` as well. JNA signals it with a `LinkageError`, because it has to extract and load its own dispatch library before advapi32 can be called, which a policy that forbids executing from the temporary directory can prevent. The command layer catches `Exception`, so left alone that would reach the user as a stack trace rather than as the "use `--allow-unsafe-credential-storage`" message the same class of failure produces on Linux and macOS, where a missing `secret-tool` or `security` surfaces through `ProcessUtils` as a `CredentialStoreException`.

### Credentials file fallback (`FileCredentialProvider`)
- `restrictFilePermissions()` caught `UnsupportedOperationException` from `setPosixFilePermissions` and only logged a warning, so on Windows the credentials file kept its inherited DACL. It now replaces the DACL with a single owner-only ACE through `AclFileAttributeView`. Inherited entries do not survive that, but only as a consequence of how the JDK writes the ACL: `setAcl` calls the legacy `SetFileSecurity` with `DACL_SECURITY_INFORMATION` alone, and the descriptor it builds is not marked auto-inherited, so Windows stores the DACL verbatim instead of merging inheritable entries back in — it never asks for the DACL to be protected. Since the guarantee rests on that detail rather than on anything requested, the resulting ACL is checked afterwards, and a file left readable by anyone other than its owner is reported through the same "could not restrict permissions" warning used when the ACL cannot be applied at all.

### Install and update (`InstallCommand`, `UpdateCommand`, `UserEnvironment`)
- Platform-aware file names: `acr.cmd` and `acr_runner.exe` instead of `acr` and `acr_runner`. `acr_env` is not installed on Windows, since neither cmd nor PowerShell can source it, and the `setExecutable` call is skipped there — Windows has no executable bit, a file is executable by virtue of its `.exe`/`.cmd` extension, so neither the call nor the "failed to set executable permission" error it can raise carries any meaning.
- The user home is read from `USERPROFILE` when `HOME` is unset. `HOME` still wins when present, so Git Bash keeps working.
- The launcher is **copied** into the bin directory rather than symlinked, because `Files.createSymbolicLink` requires Developer Mode or an elevated prompt on Windows.
- **Environment persistence replaces the shell rc file.** There is no reliable Windows equivalent of `~/.bashrc`: the PowerShell `$PROFILE` location varies by PowerShell edition and moves again when Documents is redirected to OneDrive, and cmd.exe reads no profile at all. The new `UserEnvironment` bean instead persists user-scope `ACR_HOME` and prepends the bin directory to the user `Path` via `[Environment]::SetEnvironmentVariable(..., 'User')`, which writes `HKCU\Environment` and broadcasts `WM_SETTINGCHANGE`, so it is honoured by cmd, PowerShell and Windows Terminal in any newly opened terminal. `setx.exe` is deliberately avoided — it truncates values at 1024 characters, which would corrupt a user `Path`. The prepend is idempotent (case-insensitive and trailing-separator-insensitive), so repeated installs do not grow the variable, and the read deliberately does not use `System.getenv`, which would return the machine and user values already merged and bake the machine entries into the user scope.
- Two hardening details in `UserEnvironment`: the variable name and value are handed to PowerShell through the **child process environment** rather than interpolated into the command string, so a path containing a quote cannot be parsed as PowerShell syntax; and `powershell.exe` is resolved to an absolute path under `%SystemRoot%` instead of being looked up by name, so the interpreter cannot be taken from a writable directory that happens to precede System32 on `PATH`. The read command also forces UTF-8 output, since the console otherwise defaults to an OEM code page and would mangle a `Path` containing non-ASCII characters.
- The subprocess handles all three of its streams. Standard output has to be drained while the child runs, since a value larger than the pipe buffer would otherwise block the child before it could exit — but draining it on this thread would mean waiting on a read that never returns if the child hangs, leaving the 30-second `waitFor` beyond it unreachable. It is drained on a separate thread instead, which leaves this one free to wait with a deadline and to force-kill a child that overruns it; killing it closes the pipe and releases the reader as well. The error stream is redirected away rather than left connected — merging it into standard output would be unsafe here, because on the read path that output is the value written back into `Path`, and a PowerShell warning on an otherwise successful run would be concatenated into it. Input is closed straight after starting, so the remaining pipe cannot be waited on either.
- **Self-overwrite during `acr update`.** The update flow spawns the freshly downloaded `acr install`, which copies over `%ACR_HOME%\acr_runner.exe` while the parent `acr_runner.exe` from that same directory is still running — and Windows locks the image of a running executable, so `Files.copy(REPLACE_EXISTING)` fails with `AccessDeniedException`. Renaming a running executable *is* permitted, so this extends the backup mechanism added in #8772 rather than adding a second one: on Windows the launcher and the binary are backed up by renaming to `.bak` instead of copying, which both frees the name for the incoming file and preserves the previous version. `restoreBackups` needed no change — whatever occupies the original name after a failure was written by that failed install and is not running, so it can be overwritten. Only those two files are renamed; `config.json` in particular must keep being copied, because `copyDistributionFiles` decides whether to preserve user settings with `!Files.exists(targetConfig)`.
- **The update subprocess launches the binary directly instead of through `acr.cmd`.** A batch file is not an executable image, so it would have to run under `cmd.exe /c`, and cmd re-parses its command line. A standalone `ProcessBuilder` check confirmed the JDK does not quote an argument containing `&`, so a target directory containing `&` — it derives from `ACR_HOME`, which is read from the environment — would be split and its tail executed as a separate command. Launching the executable directly removes the command interpreter entirely; the only thing the launcher would have contributed is `ACR_CURRENT_HOME`, which is now set explicitly on the child environment. This also keeps the "new binary verified working" property described in #8772, since the downloaded binary is still what performs the swap.

### Launcher and packaging (`acr.cmd`, `distribution.xml`, `cli/pom.xml`)
- New `src/main/scripts/windows/acr.cmd`, mirroring the bash `acr` launcher: it resolves `acr_runner.exe` next to itself or under `%ACR_HOME%`, exports `ACR_CURRENT_HOME`, and propagates the binary's exit code. It clears any inherited `ACR_CURRENT_HOME` first, so it cannot resolve to a previous installation during an update.
- A profile activated by `<os><family>windows</family></os>` sets `executable.suffix` to `.exe`, which the assembly uses for both the source and the destination name of the binary.
- The POSIX `acr` launcher is excluded from the Windows ZIP, where `acr.cmd` replaces it, through a `cli.common.scripts.excludes` property that is empty on every other platform.
- The `cli-attach-extra-zips` profile gains the `windows-x86_64` artifact so the ZIP is attached during the Maven Central deploy.

### Native image configuration (`application.properties`, `reachability-metadata.json`)
`net.java.dev.jna:jna-graalvm`, the artifact the JNA documentation points at for native-image support, is not published to Maven Central (`repo1.maven.org/.../jna-graalvm/` returns 404), so the metadata was captured with the GraalVM tracing agent and committed instead:
- dynamic-proxy registration for the `Advapi32` interface, since `Native.load` goes through `Proxy.newProxyInstance`, plus JNI access for the JNA core classes;
- `--initialize-at-run-time=com.sun.jna,com.sun.jna.internal`, because JNA starts a cleaner thread on initialization that cannot live in the image heap;
- `quarkus.native.resources.includes` for the bundled `jnidispatch.dll`;
- `quarkus.native.add-all-charsets=true`, because JNA reads the platform charset during initialization and it is not embedded by default;
- JNI access for `Structure$FFIType$size_t`. This one was not agent-captured: the trace ran under GraalVM 25, which does not need it. It surfaced only when the image was built with GraalVM/Mandrel for JDK 23 — the version the examples job and the release build use — which failed with `Initializing type after analysis`. `Credential` embeds `FileTime` by value, and a struct passed by value makes JNA build `FFIType` type information, whose `size` field is a `size_t`; unregistered, nothing reached it during the analysis and it was discovered only when the image heap was verified afterwards. Registering `Structure$FFIType` itself was also tried and is not required.

### CI and release (`verify-cli.yaml`, `release.yaml`, `release-artifacts.yaml`)
- `verify-cli` gains a `windows-latest` matrix entry running the same `./mvnw verify -pl cli` as the other runners, with `-Dgroups=!docker-required` (Windows runners cannot run Linux containers, as on macOS) and a longer timeout, since native-image is slowest there. The job now defaults to `shell: bash` so all three runners execute the identical step scripts — Git Bash is present on `windows-latest` — and the native-binary smoke check accepts both `-runner` and `-runner.exe`. Every `${{ github.workspace }}` expansion in those steps is quoted: it is a backslash path on `windows-latest`, and bash would otherwise consume the separators as escapes, leaving Maven with a settings path that does not exist. No MSVC setup action is needed: `native-image` auto-detects Visual Studio (verified locally from a plain shell, without `vcvars`).
- `release.yaml` builds the `windows-x86_64` classifier and includes it in the release-artifact verification loop; `release-artifacts.yaml` downloads the Windows ZIP alongside the macOS one. The GraalVM version used by the release workflow is left alone; it pins 23, which is what makes the `size_t` registration above necessary, while the verification job pins 25.

### Documentation
- `cli/README.md`: Windows row in the supported-platforms table, install and update instructions, the credential-storage prerequisite ("none — uses Windows Credential Manager"), Windows build instructions, and the known limitations — including that adding the `bin` entry rewrites the user `Path` in expanded form. That is inherent to the `[Environment]` API: it returns a `REG_EXPAND_SZ` value already resolved and writes back plain text, so a `%USERPROFILE%\...` reference is replaced by the directory it pointed at. Writing the unexpanded text back instead is worse, not better — it would store references that nothing expands any more — and preserving them properly needs a registry write with the value kind kept plus a separate `WM_SETTINGCHANGE` broadcast, which is not worth it for a `Path` that resolves correctly either way.
- `assembly-managing-registry-artifacts-cli.adoc`: this guide was rewritten by #9185 while this branch was open, so the Windows coverage is reapplied onto the new text, following the per-platform structure that rewrite established rather than reworking its prose. It replaces the "Windows is not supported" statement and adds the Windows counterpart to both sets of prerequisites, the download file names, the extract and install commands, the install description, and the shell-reload step. Two statements outside that procedure were POSIX-only as well: the credential-storage prerequisites, which did not mention the Credential Manager or its blob limit, and the unsafe-storage step, which described the credentials file as having `0600` permissions — an owner-only access control list on Windows.

### Pre-existing Windows test failures
Two failures unrelated to the new code, fixed because "the CLI can be built on Windows" is part of the acceptance criteria:
- `AbstractCLITest` and `ConfigPropertyCommandTest` resolved the `acr-home` test resource with `URL.getPath()`, which yields a leading-slash path that `Path.of` rejects on Windows — now `toURI()`.
- picocli enabled ANSI from the environment rather than from a console check, so the help-output assertions depended on `TERM`; surefire now sets `-Dpicocli.ansi=false`.

## Testing

New tests, all `@EnabledOnOs(WINDOWS)`:
- `WindowsCredentialProviderTest` (11 tests) — store/retrieve, overwrite, per-account isolation, a non-ASCII secret (an ASCII-only test would not catch a wrong charset), retrieve-missing returns `null`, an empty secret reading back as empty rather than missing, delete, delete idempotency, both sides of the blob-size boundary, and a library that cannot be loaded surfacing as a `CredentialStoreException` rather than an escaping `LinkageError`. Each test uses a unique service name and deletes what it wrote in `@AfterEach`, so nothing is left in the developer's Credential Manager.
- `InstallCommandTest` (10 new tests) — installed file set, launcher copied and not symlinked, `ACR_HOME`/`Path` persisted, prepend to an existing `Path`, no duplicate entry on re-install, case- and trailing-separator-insensitive matching, shell config untouched, binary replacement leaving no backups, `config.json` preserved, and rollback after a failed copy. A `TestUserEnvironment` alternative records what the installer would persist, mirroring the existing `TestCredentialProvider`, so tests never touch the real environment.
- `UserEnvironmentTest` (5 tests) — drives the real bean, which the installer tests deliberately replace: set/get round trip, an unset variable reading as `null`, a non-ASCII value (the UTF-8 versus OEM code page path), a 16 KB value, and a failure surfacing as an exception rather than blocking. The variable name is unique per run and removed in `@AfterEach`.
- `FileCredentialProviderAclTest` (2 tests) — the credentials file ends up with exactly one ACE for its owner, and an inheritable ACE on the containing directory does not survive onto it.
- `PlatformUtilsTest` — 3 added cases: Windows detection, `Darwin` is not misdetected as Windows, and unknown OSes still throw.

Full CLI suite on Windows 11, on the branch merged up to current `main`: **385 tests, 0 failures, 0 errors, 10 skipped** — the 10 skips are the existing macOS/Linux-only install tests. Run without group exclusions, so the `@DockerRequired` tests ran too, against `quay.io/apicurio/apicurio-registry:latest-release`. `./mvnw checkstyle:check -pl cli` is clean.

Two of the new tests were checked against a deliberately reintroduced fault, rather than assumed to be meaningful: reversing the read and the wait in `UserEnvironment.run` makes the 16 KB case fail with the expected 30 s timeout, and a file written without the ACL hardening picks up 6 entries including the inherited one, which is what the inheritance test rules out.

Manual verification on Windows 11 (GraalVM CE 25.0.2, Visual Studio 2022 / MSVC 14.44, Windows 11 SDK):
- The Linux native image also builds in `quay.io/quarkus/ubi-quarkus-mandrel-builder-image:jdk-23`, the container the examples job uses and where the missing `size_t` registration showed up.
- Native build succeeds and produces `apicurio-registry-cli-<version>-windows-x86_64.zip` containing exactly `acr.cmd`, `acr_runner.exe`, `acr_completions`, `README.md` and `config.json`.
- `acr.cmd --help` and `version` exit 0; an unknown command exits 2, so exit codes propagate through the launcher.
- `acr install` installs the expected files, copies the launcher into `bin`, and persists `ACR_HOME` and `Path`; a second install leaves the user `Path` byte-identical.
- `acr update --path <zip>` run twice from the installed location, so the running `acr_runner.exe` was overwritten both times; the second run also covers replacing a stale locked backup. Repeated under an install path containing `&`.
- Credential round-trip against a live registry: `context create` → `login` → the entry is listed by `cmdkey` as `apicurio-registry-cli/<context>/password` → authenticated `group create` and `group` → `logout` → the entry is gone.
- Both branches of the new `cli.common.scripts.excludes` property were checked by building the assembly twice: with the Windows value the ZIP has `acr.cmd` and no `acr`; with the (default) empty value `acr` is present.

## Notes

- Deferred, and recorded as known limitations in the README: PowerShell tab-completions (picocli's `AutoComplete` generates bash and zsh only), Windows on ARM, and code signing — the binaries are unsigned, so SmartScreen may warn on first run.
- After an `acr update` on Windows, an `acr_runner.exe.bak` is left behind: it is the renamed executable of the process that performed the update, and Windows keeps it locked until that process exits. The next install replaces it. The failed deletion is logged at debug rather than warn on Windows for that reason.
- SonarCloud flags the `CREDENTIALW` structure fields (`Flags`, `Type`, `TargetName`, …) under S116, since they are not camelCase. They are deliberately kept identical to the Win32 field names, which is the usual convention for JNA structure mappings and makes them checkable against the API documentation; JNA derives the memory layout from `@Structure.FieldOrder`, not from the names. These are maintainability-only findings and do not affect the quality gate.
- Related: #8380 (encrypting the OAuth client secret in the stored config) is complementary — Credential Manager storage covers the keychain side of the same concern. Not implemented here.


cli/src/main/java/io/apicurio/registry/cli/utils/UserEnvironment.java
```diff
@@ -0,0 +1,126 @@
...
+        final ProcessBuilder builder = new ProcessBuilder(
+                powershellExecutable().toString(), "-NoProfile", "-NonInteractive", "-Command", command);
+        builder.environment().put(ARG_NAME, name);
+        builder.environment().put(ARG_VALUE, value == null ? "" : value);
+        Process process = null;
+        try {
+            process = builder.start();
+            // Read the output before waiting, so a large value cannot fill the pipe buffer and
+            // block the child process while this thread waits for it to exit.
+            final String output = new String(process.getInputStream().readAllBytes(), StandardCharsets.UTF_8);
```
paoloantinori: Only `process.getInputStream()` is drained here; `getErrorStream()` is never read. If PowerShell writes more than ~4 KB to stderr — which a verbose error on the failure path easily exceeds — the child blocks on the full stderr pipe, stdout never reaches EOF, `readAllBytes()` blocks indefinitely, and the `waitFor(30s)` below is never reached, so `acr install` hangs instead of failing cleanly. The 'read output before waiting' comment addresses only the symmetric stdout-buffer case. Simplest fix: `builder.redirectErrorStream(true)` to merge stderr into stdout (on the read path a failure has `exitValue() != 0`, so the merged error text is discarded before use); otherwise drain stderr on a separate thread.

cli/src/main/java/io/apicurio/registry/cli/auth/FileCredentialProvider.java
```diff
@@ -106,12 +112,39 @@ private static void restrictFilePermissions(final Path path) {
...
+                return;
+            }
+            final UserPrincipal owner = view.getOwner();
+            final AclEntry entry = AclEntry.newBuilder()
+                    .setType(AclEntryType.ALLOW)
+                    .setPrincipal(owner)
+                    .setPermissions(EnumSet.allOf(AclEntryPermission.class))
+                    .build();
+            // Replace the inherited DACL so that only the owner can read the credentials file.
+            view.setAcl(List.of(entry));
```
paoloantinori, line 178: The comment on the line above claims the inherited DACL is replaced 'so that only the owner can read the credentials file', but `AclFileAttributeView.setAcl` only sets the explicit DACL entries — it does not set `SE_DACL_PROTECTED`, so inheritable ACEs from the parent directory are still merged in and remain effective. If the home / `ACR_CURRENT_HOME` is on a share or under a directory with permissive inherited permissions (e.g. a group ACE), other users may still read `credentials.json` despite this owner-only entry. `FileCredentialProviderAclTest` asserts the explicit ACL has one entry, which holds regardless of inheritance, so it does not verify the property. Worth confirming whether the JDK disables inheritance here, and if not, explicitly protecting the DACL before relying on the owner-only guarantee.

paoloantinori (commented):
A well-considered Windows port: the changes are additive `isWindows()` branches with no Linux/macOS behavior change, and the security hygiene is genuinely thoughtful (credential-memory clearing, PowerShell resolved from `%SystemRoot%` to avoid PATH hijacking, `setx` avoided because of its 1024-char truncation, the binary launched directly during update to dodge cmd.exe's argument re-parsing). Three items I'd want resolved before merge: a process-I/O deadlock on the error path of environment persistence, a credentials-file ACL that may not actually be owner-only, and a hand-coded credential-blob limit worth confirming against the SDK.

- **Process deadlock on the environment-persistence error path** (`cli/src/main/java/io/apicurio/registry/cli/utils/UserEnvironment.java:102`): only stdout is drained, never stderr. A verbose PowerShell error exceeding the ~4 KB pipe buffer blocks the child, stdout never reaches EOF, `readAllBytes()` hangs, and the 30 s `waitFor` is never reached — so a failed `acr install` hangs forever instead of erroring cleanly. The stdout-before-waitFor comment covers only the symmetric case; the success path (manually verified) is unaffected, so this is latent until persistence actually fails.
- **Credentials-file ACL may not be owner-only** (`cli/src/main/java/io/apicurio/registry/cli/auth/FileCredentialProvider.java:137`): `AclFileAttributeView.setAcl` sets only the explicit DACL and does not set `SE_DACL_PROTECTED`, so inheritable ACEs from the parent directory remain effective. The preceding comment's 'only the owner can read' guarantee is therefore not enforced; `FileCredentialProviderAclTest` checks the explicit ACL has one entry, which holds regardless of inheritance, so it does not guard this.
- **Hand-coded credential-blob limit** (`cli/src/main/java/io/apicurio/registry/cli/auth/WindowsCredentialProvider.java:31`): `MAX_BLOB_SIZE = 2560` should be confirmed against `CRED_MAX_CREDENTIAL_BLOB_SIZE` in wincred.h. If too high, a long OAuth token/JWT (1–2 KB+) passes the pre-check only to fail at `CredWriteW` with an opaque error; the existing test uses a 1281-char secret, which doesn't pin the boundary.
- Worth keeping: `getUserVariable` deliberately avoids `System.getenv` to prevent baking machine `Path` entries into user scope, and the prepend dedup is case- and trailing-separator-insensitive.

cli/src/main/java/io/apicurio/registry/cli/auth/WindowsCredentialProvider.java
```diff
@@ -0,0 +1,191 @@
...
+class WindowsCredentialProvider implements CredentialProvider {
+
+    private static final Logger log = Logger.getLogger(WindowsCredentialProvider.class);
+
+    private static final int CRED_TYPE_GENERIC = 1;
+    private static final int CRED_PERSIST_LOCAL_MACHINE = 2;
+    private static final int ERROR_NOT_FOUND = 1168;
+
+    /** CRED_MAX_CREDENTIAL_BLOB_SIZE — the largest secret the Credential Manager accepts. */
+    private static final int MAX_BLOB_SIZE = 2560;
```
paoloantinori, line 35: This 2560-byte limit is hand-coded; please confirm it equals `CRED_MAX_CREDENTIAL_BLOB_SIZE` in wincred.h and cite the source. If it is too high, a secret that passes this pre-check — e.g. a long OAuth token or JWT, which can be 1–2 KB+ — is accepted here only to fail at `CredWriteW` with an opaque 'CredWrite failed (error N)'; if too low, valid secrets are rejected before reaching the API. The existing `testStoreRejectsSecretLargerThanTheCredentialManagerLimit` uses a 1281-char secret (well over either plausible value), so it doesn't pin the boundary — worth adding a boundary test once the constant is confirmed.

Vidhaankhare16:
Thanks @paoloantinori — the deadlock is real and is fixed. The other two turned out to be confirmations rather than changes, so I've written up the evidence and added tests that actually pin the properties, since in both cases you were right that the existing tests did not.

**1. Process deadlock on the environment-persistence error path — fixed.**

Confirmed: only stdout was drained, and `readAllBytes()` runs before `waitFor`, so a full stderr pipe would block the child, stdout would never reach EOF, and the 30 s timeout would never be reached.

I went with `builder.redirectError(ProcessBuilder.Redirect.DISCARD)` rather than `redirectErrorStream(true)`. Merging closes the deadlock, but on the read path the stdout value *is* the string that gets written back into the user's `Path` — and the "a failure has `exitValue() != 0`, so the merged text is discarded" argument doesn't cover a PowerShell warning emitted on an otherwise successful run, which would be concatenated into the value and silently corrupt `Path`. Discarding removes the pipe entirely: no deadlock, and stdout stays exactly the variable's value. Nothing is lost, since the error text was never used — only the exit code is reported.

**2. Credentials-file ACL — the DACL *is* protected; inheritance does not survive.**

Good thing to challenge, so I tested it rather than reasoned about it. I created a directory carrying an inheritable ACE and ran the exact `restrictWindowsAcl` logic against a file inside it:

```
# parent directory
C:\...\aclprobe BUILTIN\Users:(OI)(CI)(R)
                NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
                BUILTIN\Administrators:(I)(OI)(CI)(F)
                ...

# the file, after AclFileAttributeView.setAcl(List.of(ownerEntry))
C:\...\aclprobe\credentials.json LAPTOP-...\VICTUS 15:(F)

PS> (Get-Acl credentials.json).AreAccessRulesProtected
True
```

So the JDK's `WindowsAclFileAttributeView.setAcl` does set `SE_DACL_PROTECTED`: the inheritable `BUILTIN\Users:(R)` ACE is not present on the file, and the owner-only guarantee in the comment holds. (Worth noting the real write path is create-temp-file → `restrictFilePermissions` → rename, and a rename within the same directory does not re-apply inheritance, so the final file keeps the protected DACL — which the existing test already observes on `credentials.json`.)

You're right that asserting "one entry" didn't establish this, so there is now a second test, `testCredentialsFileDoesNotInheritPermissionsFromItsDirectory`: it grants the containing directory an `Everyone` ACE flagged `FILE_INHERIT | DIRECTORY_INHERIT`, writes credentials through the provider, and asserts the resulting ACL contains the owner and nothing else. It carries an assumption guard, since the well-known group name is localised.

I checked that the test can actually fail rather than passing vacuously — in the same directory, a file written *without* `restrictWindowsAcl` comes out with 6 ACEs, `Everyone` among them:

```
entries on an UNPROTECTED file: 6
inherited 'Everyone' present   : true
```

so the inheritance really is in play in that fixture, and the assertion is what rules it out.

**3. `MAX_BLOB_SIZE` — confirmed against the SDK, and the boundary is now pinned on both sides.**

From `wincred.h` in the Windows SDK on my build machine (`10.0.22621.0/um/wincred.h`, line 447):

```c
#define CRED_MAX_CREDENTIAL_BLOB_SIZE   (5*512)
```

2560 bytes, so the constant is right. It now carries that citation in a comment instead of standing as a bare number.

One correction on the existing test, though: the blob is written as UTF-16LE, so the 1281-character secret is **2562 bytes** — two bytes over the limit, not well over it. It was already a boundary test; what it was missing is the accepting side. `testStoreAcceptsSecretAtTheCredentialManagerLimit` now stores a 1280-character (exactly 2560-byte) secret and reads it back, and both tests derive their length from a single `MAX_SECRET_CHARACTERS = 1280` constant so the two sides cannot drift apart. Both run against the real Credential Manager, so this also confirms empirically that `CredWriteW` accepts exactly 2560 bytes rather than rejecting at the edge — a long OAuth token or JWT is comfortably inside it.

---

Full CLI suite re-run on Windows 11 after these changes: **337 tests, 0 failures, 10 skipped** (the macOS/Linux-only install tests), Docker-based tests included, checkstyle clean. Rebuilt the native binary and re-ran the end-to-end checks against it: `acr install` (the write path) and a re-install (the read path — the user `Path` comes back byte-identical, which is the case the discarded stream protects), and the `context create` → `login` → `group create` → `logout` Credential Manager round-trip.

Pushed as a separate commit so the delta against your review is easy to read; I'll squash before merge as CONTRIBUTING asks.

paoloantinori:
Thanks for the detailed write-up. All three points are resolved at this head:

- Deadlock fixed: `UserEnvironment.run` now discards stderr (line 103) and reads stdout to completion before `waitFor` (line 109). Your reasoning for discard-over-merge is correct and the one case it guards (a PowerShell warning on an otherwise successful read corrupting the value written back to `Path`) is real.
- Credentials-file inheritance: the new `testCredentialsFileDoesNotInheritPermissionsFromItsDirectory` actually fails on an unprotected write (your 6-ACE fixture), so the assertion has discriminative power.
- `MAX_SECRET_CHARACTERS = 1280` derived from the SDK constant, with the accepting-side test pinning the boundary so the two sides cannot drift.

cli/src/main/java/io/apicurio/registry/cli/auth/WindowsCredentialProvider.java
```diff
@@ -0,0 +1,195 @@
...
+        }
+        return library;
+    }
+
+    @Override
+    public void store(final String account, final String secret) {
+        final byte[] blob = secret.getBytes(StandardCharsets.UTF_16LE);
+        if (blob.length > MAX_BLOB_SIZE) {
+            throw new CredentialStoreException(
+                    "Secret is too large for the Windows Credential Manager (the limit is "
```
paoloantinori, line 87: Secrets larger than 2560 bytes are rejected here, but the macOS and Linux providers have no such cap, so long OAuth or JWT tokens (commonly over 2560 bytes) fail on Windows only. If the CLI stores tokens through this provider, authentication breaks where it works on the other platforms. Consider a fallback for oversized secrets and list this limit in the README known-limitations section.

cli/src/main/java/io/apicurio/registry/cli/auth/WindowsCredentialProvider.java
```diff
@@ -0,0 +1,195 @@
...
+        if (!library().CredReadW(new WString(targetName(account)), CRED_TYPE_GENERIC, 0,
+                credentialPointer)) {
+            final int error = Native.getLastError();
+            if (error == ERROR_NOT_FOUND) {
+                return null;
+            }
+            throw new CredentialStoreException("CredRead failed (error " + error + ").");
+        }
+        final Pointer pointer = credentialPointer.getValue();
+        if (pointer == null) {
```
paoloantinori, line 127: Returning null when CredentialBlobSize <= 0 treats a stored empty secret as 'not found'. The file fallback and the macOS/Linux providers return an empty string for an empty secret, so this is a Windows-only divergence: an empty basic-auth password round-trips as null and reads as 'no credential'. Consider returning an empty string when the entry exists but has an empty blob.

paoloantinori (commented):
This is a well-reasoned, thoroughly tested PR: the Windows branches are additive, the install and update self-overwrite handling is careful, and the coverage (non-ASCII secrets, the ACL inheritance case) is strong. The findings below are Windows-specific divergences from the cross-platform credential contract, not regressions on Linux or macOS.

- `WindowsCredentialProvider.store` (cli/src/main/java/io/apicurio/registry/cli/auth/WindowsCredentialProvider.java, around line 63): the 2560-byte cap is inherent to Windows Credential Manager, but macOS Keychain and Linux Secret Service have no equivalent limit, so long secrets (notably OAuth/JWT access or refresh tokens, which can exceed 2560 bytes) are rejected on Windows only. Confirm the largest secret the CLI stores stays under the limit; if token storage is in scope this needs a fallback, and the README limitations section does not mention it.
- `WindowsCredentialProvider.retrieve` (around line 103): a stored empty secret (`CredentialBlobSize <= 0`) is returned as `null`, conflating 'exists but empty' with 'not found'. The file fallback and the macOS/Linux providers return an empty string for an empty secret, so Windows alone diverges; a stored empty basic-auth password would read as 'no credential'.
- `FileCredentialProvider.restrictWindowsAcl` (cli/src/main/java/io/apicurio/registry/cli/auth/FileCredentialProvider.java, around line 119): the owner-only ACL relies on `setAcl` clearing inherited ACEs. The dedicated `FileCredentialProviderAclTest` covers this, and since it is the security-critical spot it is worth a second look at the effective (not just explicit) ACL when the parent directory grants a permissive inheritable ACE.
- `UserEnvironment` (cli/src/main/java/io/apicurio/registry/cli/utils/UserEnvironment.java): the real PowerShell round-trip (UTF-8 versus OEM code page for non-ASCII paths) is only manually verified. `TestUserEnvironment` bypasses `run()` entirely, so the charset and large-value paths in the production class have no automated coverage.

Vidhaankhare16:
Thanks — all four looked at. Two led to changes, one is a documentation gap you were right about, and one I believe is already covered; details and evidence below.

**1. The 2560-byte cap versus long tokens — tokens are never stored, but the limitation is now documented.**

I traced every `CredentialStore` call site. The CLI stores exactly two secrets, and only these two:

- `CREDENTIAL_KEY_PASSWORD` — the basic-auth password
- `CREDENTIAL_KEY_CLIENT_SECRET` — the OAuth2 client secret

Access and refresh tokens are never persisted: for OAuth2, `Client.configureAuth` hands the client secret to `options.oauth2(tokenEndpoint, clientId, clientSecret, scope)` and the SDK obtains tokens per invocation. So the case you were worried about — a JWT access or refresh token exceeding 2560 bytes — isn't reachable, and a client secret is an opaque short string nowhere near the limit. That means no fallback is needed, and I'd rather not add one for a path that cannot be hit.

You were right that the README didn't mention it, though, so the credential-storage section now states the limit, that it is a Credential Manager constraint with no macOS/Linux equivalent, and why the CLI does not reach it.

**2. Empty secret read back as `null` — real divergence, fixed; not reachable through the CLI.**

Confirmed rather than assumed. `CredWriteW` does accept a zero-length blob, and the round trip was asymmetric:

```
store("")  -> OK
retrieve() -> null
```

`retrieve` now returns `""` when the entry exists with an empty blob, so `null` means "no entry" as it does for the file fallback and macOS. `testRetrieveReturnsEmptyStringForAnEmptySecret` covers both halves.

In fairness I should say the divergence was not user-visible: `LoginCommand` routes a blank `--password` / `--client-secret` into `readSecret`, which rejects empty input, so an empty secret can't be stored through the CLI in the first place; and the only consumer, `Client.requireCredential`, tests `isBlank`, which treats `null` and `""` alike. It's worth fixing so the provider contract is uniform, but it wasn't affecting behaviour.

**3. Credentials-file ACL — the existing test does check the effective ACL.**

I think this one is already covered, and the reason is worth stating explicitly since it isn't obvious: on Windows `AclFileAttributeView.getAcl()` returns the file's DACL *including* inherited ACEs, so it is the effective ACL, not just the explicit entries.

The control makes that concrete. In a directory carrying a permissive inheritable ACE, a file written **without** `restrictWindowsAcl` comes back as:

```
entries on an UNPROTECTED file: 6
inherited 'Everyone' present   : true
```

and with it, exactly one entry — the owner. So `testCredentialsFileDoesNotInheritPermissionsFromItsDirectory` fails if the inherited entries survive; that is precisely the property it asserts. Independently, at the Win32 level:

```
icacls credentials.json  ->  LAPTOP-...\VICTUS 15:(F)      # only the owner
(Get-Acl credentials.json).AreAccessRulesProtected -> True # SE_DACL_PROTECTED is set
```

So the JDK does protect the DACL, and the owner-only guarantee in the comment holds. Happy to add anything further here if you still see a gap — this is the security-critical spot, so I'd rather over-verify it.

**4. `UserEnvironment.run()` had no automated coverage — fixed, and I checked the tests can fail.**

Fair, and the weakest point in the PR given I had just changed that method. There is now a `UserEnvironmentTest` exercising the real bean against a per-run unique variable name that is removed again in `@AfterEach`, mirroring how `WindowsCredentialProviderTest` handles real Credential Manager entries. It covers the set/get round trip, unset reading as `null`, a non-ASCII value (the UTF-8 versus OEM code page path), and a 16 KB value.

The large-value test is the one that matters, and I verified it is not a vacuous pass: reversing the read and the wait in `run()` makes it fail exactly as the deadlock would —

```
UserEnvironmentTest.testValueIsNotTruncatedOrWrappedAtTheConsoleWidth
  CliException: Timed out trying to read the user environment variable ACR_TEST_c928f661...
  Time elapsed: 30.57 s
```

— then passes again once the order is restored.

One point of accuracy on the error path: there is also a `testFailureIsReportedRatherThanHanging`, but I've noted in it that it does *not* reproduce the buffer-filling case — the PowerShell error there is only a few hundred bytes, well under the pipe buffer. It guards that a failure surfaces promptly as a `CliException`. The deadlock itself is prevented structurally rather than by test: with the error stream redirected away there is no stderr pipe to exhaust.

While in there I also closed the child's input stream after starting it. The commands never read input, but that leaves the third stream as an open pipe, and it costs one line to hand them end of input instead.

---

Full CLI suite on Windows 11 after these changes: **343 tests, 0 failures, 10 skipped** (the macOS/Linux-only install tests), Docker-based tests included, checkstyle clean. Native binary rebuilt and re-verified end to end: `acr install`, re-install, and the `context create` → `login` → `group create` → `logout` Credential Manager round trip.

Pushed as a separate commit again for readability against your review; I'll squash the lot before merge.

paoloantinori (commented):
The final research agent confirms `picocli.ansi` is a real picocli system property (verified against the picocli source at `CommandLine.java:18034-18041` and its docs), and `picocli.ansi=false` unconditionally forces ANSI off, overriding all platform and environment detection. That corroborates my earlier ruling: the unconditional setting in surefire is deliberate test hardening that makes the help-output assertions deterministic across all three CI runners, not a regression. It stays excluded from the review.

That was the last verification sub-task. Every one is now resolved, and each agrees with the review as delivered:

- **Confirmed findings (kept):** the ACL inheritance gap (corroborated by three agents plus a dedicated CONFIRMED skeptic) and the unreachable `UserEnvironment` timeout.
- **Refuted and excluded:** the JNA version skew, the `--initialize-at-run-time` comma-splitting claim (CONFIRMED only on a misread input; the actual diff declares three separate valid flags, and the reported successful native build confirms it), the UpdateCommand exit-code concern, metadata coverage, picocli.ansi, the empty-exclude, the jnidispatch.dll path, and the TestUserEnvironment activation risk.

The review is complete and unchanged: two MAJOR findings (ACL inheritance not blocked; subprocess timeout unreachable on a child hang) plus one minor easy win (oversize-secret blob not zeroed on the reject path). Nothing further is pending.

Vidhaankhare16:
All three addressed. On the ACL finding you were right and I was not, so let me correct my earlier reply first.

**Credentials-file ACL — I was wrong about the mechanism.**

I previously said the JDK sets `SE_DACL_PROTECTED`. It does not. `WindowsAclFileAttributeView.setAcl` is:

```java
SetFileSecurity(path, DACL_SECURITY_INFORMATION, sd.address());
```

`DACL_SECURITY_INFORMATION` alone, with no `PROTECTED_DACL_SECURITY_INFORMATION` — exactly as you said.

The measurement I posted was still accurate; the two reconcile as an accident of which API the JDK uses. `SetFileSecurity` is the legacy call and does not run the auto-inheritance algorithm, and the descriptor `WindowsSecurityDescriptor.create` hands it is not marked auto-inherited, so Windows stores the DACL verbatim rather than merging inheritable entries back in. That is why `AreAccessRulesProtected` reads `True` and the inherited `Everyone` entry is absent.

Which leaves the guarantee resting on an implementation detail rather than on anything requested, and a future JDK moving to `SetNamedSecurityInfo` would silently reintroduce the inherited entries. So `restrictWindowsAcl` no longer assumes it: after writing the ACL it checks the result really is owner-only, and anything else is reported through the existing "could not restrict permissions" warning — the same fallback used when the ACL cannot be applied at all. The comment now describes the mechanism above instead of claiming protection is requested.

**Subprocess timeout unreachable on a child hang — correct, and a real defect.**

The output was read to completion before the wait, on the same thread, so a child that hung without closing the stream left that read blocked forever and `waitFor` was never reached. The timeout was effectively dead code. Draining is still necessary, since a value larger than the pipe buffer would block the child before it could exit, so it now happens on a separate thread; this one waits with the deadline and force-kills a child that overruns it, which also releases the reader.

This is the same defect class as your first finding, one stream over. The earlier fixes addressed a *full* pipe; neither addressed a child that simply stops.

**Oversize secret not zeroed on the reject path — correct.**

The length check returned before entering the `try`, so the UTF-16LE copy was left in the heap while every other path in that class wipes it. Zeroed before the throw now.

---

On the excluded items, one small correction for the record: the surefire `picocli.ansi=false` is deliberate, but not only for determinism across runners. Without it the help-output assertions fail outright on Windows, because picocli enables ANSI from the environment — `TERM` under Git Bash — rather than from a console check. It is one of the two pre-existing Windows-only failures this PR fixes, which is why it is unconditional rather than Windows-gated: a conditional setting would leave the other two runners on a different code path to no benefit.

Verified after these changes: full CLI suite **343 tests, 0 failures, 10 skipped** (the macOS/Linux-only install tests), Docker-based tests included, checkstyle clean. The behaviour these touch is covered by the existing `FileCredentialProviderAclTest` inheritance case and the `UserEnvironmentTest` large-value case; I have not added a test for the hung-child path, since inducing a genuine hang in PowerShell from the public API is not something I can do reliably, and a test that only pretends to would be worse than none.

paoloantinori:
Thanks for the correction on the ACL mechanism — you are right that `SetFileSecurity` with `DACL_SECURITY_INFORMATION` alone stores the DACL verbatim without running auto-inheritance, so the owner-only result is a consequence of the legacy API path rather than an explicit `SE_DACL_PROTECTED` request. Guarding it with the post-write `isRestrictedToOwner` check is the right call: the guarantee now holds regardless of which Win32 API a future JDK switches to.

All three confirmed at this head:

- `restrictWindowsAcl` verifies the effective ACL is owner-only after writing, warning through the existing fallback if not.
- `UserEnvironment.run` drains stdout on a separate thread while this thread waits with the 30 s deadline, so a hung child is force-killed rather than blocking the read forever.
- The oversize-secret reject path zeroes the UTF-16LE copy before throwing.

Nothing further from my side.

Vidhaankhare16:
@paoloantinori a short note on what has changed since your last pass, since two of these are not things you raised.

**A JNA load failure escaped as an `Error`.** Found while auditing for the same contract-divergence pattern your earlier findings shared. `Native.load` was unguarded, so a dispatch library that cannot be extracted or loaded — a locked-down `%TEMP%`, AppLocker, an antivirus quarantine — raised `UnsatisfiedLinkError`. `AbstractCommand` catches `Exception`, not `Throwable`, and there is no `IExecutionExceptionHandler`, so it left picocli as a stack trace. On macOS and Linux the same class of failure, a missing `security` or `secret-tool`, comes back through `ProcessUtils` as a `CredentialStoreException` and routes into the documented `--allow-unsafe-credential-storage` fallback. The load is now wrapped so Windows behaves the same way, with a test that drives it through a library name that does not resolve.

**The CLI guide was rewritten under this branch.** #9185 merged while this was open and rewrote `assembly-managing-registry-artifacts-cli.adoc` from 403 to 781 lines, which put the PR into conflict — the new text states that Windows is not supported, which was true when it was written. Rather than resolve hunk by hunk against a document that no longer exists, the merge takes their guide as-is and a separate commit reapplies the Windows coverage onto it, following the per-platform structure their rewrite established. That turned up two POSIX-only statements the old guide did not have: the credential storage prerequisites, and the unsafe-storage step describing the credentials file as `0600`, which on Windows is an owner-only ACL.

**Also worth flagging for whoever sequences these:** #9233 adds a `HOME` → `USERPROFILE` fallback to `getUserHomePath` and #8989 adds `acr install --global`, both touching the same method this PR reworks. Whichever lands second needs a small rebase. #9233 additionally calls `PlatformUtils.isWindows()`, which only exists in this PR, so it does not compile against `main` today — I have raised that on the PR itself rather than here.

The PR description has been brought back in line with the code: it was still describing the pre-fix draining order and the ACL bullet predated the verification, both of which would otherwise have landed in the squash-merge message.

Verified on the branch merged up to current `main`: full CLI suite **369 tests, 0 failures, 10 skipped** (the macOS/Linux-only install tests), Docker-based tests included, checkstyle clean. Native build, the ZIP contents, `acr install`, a re-install, `acr update`, and the `login` → `group create` → `logout` Credential Manager round trip were all re-run against a freshly built binary, since the JNA loading path is exactly the part where JVM-mode behaviour has not been a reliable guide to the native image.

paoloantinori:
Thanks for flagging this — good catch, and the fix looks right.

Confirmed at this head: `WindowsCredentialProvider.load` (lines 71-78) wraps `Native.load` in `catch (LinkageError | RuntimeException)` and rethrows as `CredentialStoreException`, so a dispatch library that can't be extracted or loaded (locked-down `%TEMP%`, AppLocker, AV quarantine) surfaces as the same "not available" condition the other providers produce, and `CredentialStore` falls back to file storage instead of blowing up with a stack trace. The `catch LinkageError` is the load-bearing part — `AbstractCommand` catches `Exception`, not `Throwable`, so leaving the `Error` uncaught would have escaped the command layer exactly as you described.

The test (`testALibraryThatCannotBeLoadedIsReportedAsACredentialStoreException`) drives it with a non-existent library name and asserts both the `CredentialStoreException` wrapper and the `UnsatisfiedLinkError` cause, which reproduces the real failure mode without needing to actually lock down the temp dir. No further concerns from me on this point.

vandanayadav7 (dismissed):
Reviewed the full diff and ran regression tests on Linux — all pass, no regressions. Also ran e2e in a headless Linux container covering context, login, groups, artifacts, versions, search, and logout — everything works. Windows-specific code is cleanly separated behind `isWindows()` checks and doesn't touch Linux/macOS paths.

Security review looks good — credential provider, ACL hardening, PowerShell env persistence, and the update command all handle secrets, command injection, and error cases properly.

The earlier review comments are all addressed in the current code. SonarCloud field-naming findings on the `CREDENTIALW` structure are expected — JNA struct fields must match the Win32 API names.

LGTM. Waiting for the Windows CI job to confirm the platform-specific tests pass.

Vidhaankhare16:
Pushed a merge of current `main` plus fixes for the two failures that were mine.

`CLI (windows-x86_64)`: `${{ github.workspace }}` is a backslash path on windows-latest and the step runs under bash, so the backslashes were eaten and Maven never found the settings file - the job died in 58s before compiling anything. Quoted the expansions; `release.yaml` needed the same for the windows CLI build.

`Build and Run Application examples`: the native image failed on `Structure$FFIType$size_t`, which was missing from the reachability metadata (`Credential` embeds `FileTime` by value, so JNA builds `FFIType`). One entry added. This also affects the release build, which pins GraalVM 23 - the CLI job pins 25, which is why it passed there.

Verified locally with a Mandrel jdk-23 container build, a GraalVM 25 Windows build, the full CLI suite (385 tests, 0 failures, checkstyle clean), and a Credential Manager round trip from the native binary. `Unit Tests (app-other)` was the heap exhaustion #9287 fixed on main.

@vandanayadav7 the push dismissed your review and disabled auto-merge.

paoloantinori:
Re-read at `bc14046`. The Windows bash-path fix is in: every `${{ github.workspace }}` expansion in `verify-cli.yaml` is now quoted (lines 68, 71, 86, 97), with a clear comment explaining the backslash-eaten-by-bash trap. That closes the CI failure. The four points from the prior head are all resolved here. Nothing else to add.

Vidhaankhare16:
All checks pass on 483ebe8a - 40 green, 4 skipped, Windows CLI included.The failure before it was `setup-graalvm`'s job reports writing `BuildOutputJSONFile` to a different drive than the workspace, not CLI code; the `app-kafkasql` one after it was an unrelated container flake and passed on `/retry`.

@vandanayadav7 my push yesterday dismissed your approval and disabled auto-merge, so both need re-enabling, thank you!

vandanayadav7 (approved):
Re-approving after branch update. LGTM.

Vidhaankhare16:
@vandanayadav7 `Integration Tests (kafkasql-default)` on `7e96101` has been running since 11:21 UTC. 17 of the 18 test classes passed; `SchemasConfluentIT` started at 12:06:50 and hasn't reported since - that class took 24.85s on `main` this morning. Everything else in the run is green.

The job has no `timeout-minutes`, so it will hold until the 360-minute default, and `/retry` only re-triggers on a completed run. Could you cancel and re-run it?

Vidhaankhare16:
@vandanayadav7 the auto-merge re-arm didn't take, I think because the block isn't auto-merge itself. Two things are holding it:
1. `Integration Tests (kafkasql-default)` is still wedged (2h20m; it's 16 min on `main`). It won't fail until the 360-minute cap at 17:21 UTC.
2. The branch went `BEHIND` when #9432 merged at 12:45.

If you can cancel that job and re-run it, and update the branch, it should land. I can `/retry` once the run times out, but that re-runs the whole suite rather than the one job. Thank you!

Vidhaankhare16:
@vandanayadav7 all green now. Just needs **Update branch** - it's `BEHIND` after #9432.

EricWittmann:
Thanks for this contribution, @Vidhaankhare16! Great work adding Windows support for the CLI. 🎉

Vidhaankhare16:
Thank you! @EricWittmann. And thank you @vandanayadav7 and @paoloantinori for the reviews - the credential storage and update paths came out a lot better for them :)

PR #9233, opened 2026-08-04
Fix HOME environment variable issue

techsavyr00hit:
This PR fixes the `HOME` environment variable issue in `InstallCommand` by adding fallbacks to `USERPROFILE` and `System.getProperty("user.home")`. This allows `acr install` to work even when `HOME` is not set, which is common on Windows.
**Note:** This PR only fixes the `InstallCommand` issue. The `UpdateCommand` issue mentioned in the original report is not included in this PR.

### DIFF
--- a/cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
@@ -238,18 +238,16 @@ public class InstallCommand extends AbstractCommand {
      * Gets the user's home directory path.
      */
      private Path getUserHomePath() {
-         String userHome;
-         if (PlatformUtils.isWindows()) {
+         String userHome = config.getEnv(ENV_HOME);
+         if (isBlank(userHome) && PlatformUtils.isWindows()) {
              userHome = config.getEnv("USERPROFILE");
-         } else {
-             userHome = config.getEnv(ENV_HOME);
          }
          if (isBlank(userHome)) {
-             userHome = System.getProperty("user.home");
+             userHome = config.getSystemProperty("user.home");
          }
          if (isBlank(userHome)) {
              throw new CliException(
-                     "Unable to determine the user's home directory.",
+                     "Unable to determine the user's home directory. Checked HOME, USERPROFILE, and user.home.",
                      VALIDATION_ERROR_RETURN_CODE);
          }
          return Path.of(userHome).normalize().toAbsolutePath();
--- a/cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
@@ -234,18 +234,26 @@ public class InstallCommand extends AbstractCommand {
             }
         });
     }
-
     /**
      * Gets the user's home directory path.
      */
-    private Path getUserHomePath() {
-        final String userHome = config.getEnv(ENV_HOME);
-        if (isBlank(userHome)) {
-            throw new CliException(ENV_HOME + " environment variable is not set.", VALIDATION_ERROR_RETURN_CODE);
-        }
-        return Path.of(userHome).normalize().toAbsolutePath();
-    }
-
+     private Path getUserHomePath() {
+         String userHome;
+         if (PlatformUtils.isWindows()) {
+             userHome = config.getEnv("USERPROFILE");
+         } else {
+             userHome = config.getEnv(ENV_HOME);
+         }
+         if (isBlank(userHome)) {
+             userHome = System.getProperty("user.home");
+         }
+         if (isBlank(userHome)) {
+             throw new CliException(
+                     "Unable to determine the user's home directory.",
+                     VALIDATION_ERROR_RETURN_CODE);
+         }
+         return Path.of(userHome).normalize().toAbsolutePath();
+     }
     /**
      * Ensures the ~/bin directory exists, creating it if necessary.
      */
--- a/cli/src/main/java/io/apicurio/registry/cli/config/Config.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/config/Config.java
@@ -73,7 +73,15 @@ public class Config {
         }
         return System.getenv(name);
     }
-
+    /**
+     * Gets a system property value.
+     *
+     * @param name the system property name
+     * @return the system property value, or null if not set
+     */
+    public String getSystemProperty(final String name) {
+        return System.getProperty(name);
+    }
     /**
      * Sets an environment variable override for testing purposes.
      * The override takes precedence over the actual system environment variable.
--- a/cli/src/test/java/io/apicurio/registry/cli/SearchCommandTest.java
+++ b/cli/src/test/java/io/apicurio/registry/cli/SearchCommandTest.java
@@ -206,8 +206,9 @@ public class SearchCommandTest extends AbstractCLITest {
         var results = MAPPER.readValue(out.toString(), ArtifactSearchResults.class);
 
         assertThat(results.getArtifacts())
-                .as(withCliOutput("Order by should return results"))
-                .isNotEmpty();
+                .as(withCliOutput("Type-filtered search should return JSON artifacts"))
+                .isNotEmpty()
+                .allMatch(a -> "JSON".equals(a.getArtifactType()));
     }
 
     @Test


cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
```diff
@@ -234,18 +234,26 @@ private void deleteBackups(final Map<Path, Path> backups) {
...
-        final String userHome = config.getEnv(ENV_HOME);
-        if (isBlank(userHome)) {
-            throw new CliException(ENV_HOME + " environment variable is not set.", VALIDATION_ERROR_RETURN_CODE);
-        }
-        return Path.of(userHome).normalize().toAbsolutePath();
-    }
-
+     private Path getUserHomePath() {
+         String userHome;
+         if (PlatformUtils.isWindows()) {
```
paoloantinori: On Windows this branch reads USERPROFILE and skips HOME entirely, but the previous code honored HOME on every platform. A Windows user who sets HOME (common in Git Bash, Cygwin, or when this .exe is launched from a WSL shell) now has it silently ignored and the install target moves. Prefer HOME first on all platforms, then USERPROFILE on Windows, then `user.home`, e.g. `String userHome = config.getEnv(ENV_HOME); if (isBlank(userHome) && PlatformUtils.isWindows()) userHome = config.getEnv("USERPROFILE"); if (isBlank(userHome)) userHome = System.getProperty("user.home");`. That keeps the native-Windows crash fix while respecting an explicit HOME.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
```diff
@@ -234,18 +234,26 @@ private void deleteBackups(final Map<Path, Path> backups) {
...
      */
-    private Path getUserHomePath() {
-        final String userHome = config.getEnv(ENV_HOME);
-        if (isBlank(userHome)) {
-            throw new CliException(ENV_HOME + " environment variable is not set.", VALIDATION_ERROR_RETURN_CODE);
-        }
-        return Path.of(userHome).normalize().toAbsolutePath();
-    }
-
+     private Path getUserHomePath() {
```
paoloantinori, line 240: This home-resolution logic is duplicated. UpdateCommand has the same method and, per the PR description, still throws when HOME is unset, so `acr update` remains broken on Windows after this change. Extract the resolution into a shared CLI helper (e.g. `PathUtils.resolveUserHome(config)`) and call it from both commands so the fix is applied once and the two cannot drift.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
```diff
@@ -234,18 +234,26 @@ private void deleteBackups(final Map<Path, Path> backups) {
...
+         String userHome;
+         if (PlatformUtils.isWindows()) {
+             userHome = config.getEnv("USERPROFILE");
+         } else {
+             userHome = config.getEnv(ENV_HOME);
+         }
+         if (isBlank(userHome)) {
+             userHome = System.getProperty("user.home");
```
paoloantinori: `System.getProperty("user.home")` bypasses the `config` abstraction that wraps env access (config.getEnv) in the rest of this method, so this fallback can't be controlled or unit-tested through the config seam. If `config` exposes a system-property accessor, use it here; if it doesn't, consider adding one rather than leaking System.getProperty into this path.

paoloantinori (commented):
The fix adds a needed Windows fallback so `acr install` no longer hard-fails when HOME is unset, but it introduces a precedence regression for Windows users who set HOME, and it leaves the identical bug unfixed in UpdateCommand. I'd address the HOME precedence and extract the resolution into a shared helper before merging.

- `InstallCommand.java:242`: on Windows the new branch reads USERPROFILE and never checks HOME, whereas the old code honored HOME on every platform. A deliberately-set HOME (Git Bash, Cygwin, or this .exe launched from a WSL shell) is now silently ignored and the install target changes. Verified by tracing the branches; the regression is real for the HOME-setting subset even though the change also fixes the native-Windows crash.
- `InstallCommand.java:240`: UpdateCommand has the same logic and, per the PR note, still throws when HOME is unset, so `acr update` stays broken on Windows. The resolution is now duplicated and divergent across the two commands and should live in one shared helper.
- `InstallCommand.java:248`: `System.getProperty("user.home")` reaches outside the `config` abstraction used for `getEnv`, so the fallback can't be driven through the config seam in tests. Worth checking whether `config` exposes a property accessor and routing through it for consistency.
- `InstallCommand.java:252`: the generic "Unable to determine the user's home directory." drops the specificity of the old message. Naming the checked variables (HOME on Unix, USERPROFILE on Windows) makes the failure actionable again.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
techsavyr00hit: Thanks for pointing this out! I've updated the logic to check `HOME` first, then fall back to `USERPROFILE` on Windows, and finally `user.home` if needed. This keeps the Windows fix while still respecting a user-defined `HOME`.

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
techsavyr00hit: Thanks for the suggestion! I looked into this, but I noticed that `UpdateCommand` currently uses `config.getAcrHomePath()` instead of the same home directory resolution logic. Could you please clarify which part you were referring to for extracting the shared helper?

cli/src/main/java/io/apicurio/registry/cli/InstallCommand.java
techsavyr00hit: Thanks for the suggestion! I added a `getSystemProperty()` method to `Config` and updated `InstallCommand` to use it instead of calling `System.getProperty()` directly.

paoloantinori:
Thanks — the `getSystemProperty()` abstraction on `Config` and the `HOME → USERPROFILE → user.home` precedence in `getUserHomePath()` both look right to me. The earlier points are addressed.

techsavyr00hit:
Thanks. I am happy to hear the changes look good. 😊 I'm still pretty new to contributing to open source, so I really appreciate the detailed feedback—it helped me learn a lot.

Vidhaankhare16:
Heads up on a build issue and an overlap.

`getUserHomePath()` calls `PlatformUtils.isWindows()`, which doesn't exist on `main` yet - it's added by #9229. I merged this PR into current `main` locally and built it:

```
[ERROR] InstallCommand.java:[242,48] cannot find symbol
  symbol:   method isWindows()
  location: class io.apicurio.registry.cli.utils.PlatformUtils
[INFO] BUILD FAILURE
```

The merge itself is clean, so it's not a conflict - the method just isn't there yet. It doesn't show up on the PR because the build jobs only run after a maintainer `/accept`s, and the SonarCloud check doesn't compile the module.

The `HOME` -> `USERPROFILE` fallback is also already implemented in #9229, which covers #8646 (CLI Windows support, assigned to me) and adds `isWindows()` as part of the same change.

The third fallback isn't duplicated though, and it's better than mine: `System.getProperty("user.home")` via `Config.getSystemProperty()` is always set by the JVM, so `acr install` works even where neither `HOME` nor `USERPROFILE` is. Mine stops at `USERPROFILE`.

Happy to fold that into #9229 with credit to you, or leave it here as a follow-up once #9229 lands and `isWindows()` exists. Either way both PRs touch `getUserHomePath()`, so whichever merges second will need a small rebase.

techsavyr00hit:
Thanks for taking the time to check this and explain everything!
That makes sense. I'm happy to rebase this PR after #9229 is merged.
Really appreciate the detailed explanation!

techsavyr00hit:
@vandanayadav7 Hey, could you please take a look at this PR when you have some time? I'd really appreciate your review. Thanks!

vandanayadav7:
Thanks for the contribution! PR #9229 (Windows support) already adds the `USERPROFILE` fallback to `getUserHomePath()` and is approved and being merged. Once that lands, the main fix here is covered. It also fixes the `UpdateCommand` issue by switching from `getAcrHomePath()` to `getAcrCurrentHomePath()`.

The `System.getProperty("user.home")` fallback wouldn't help for the native binary (no JVM), and `HOME` or `USERPROFILE` is always available on their respective platforms, so it wouldn't be reached in practice.

Also, the `SearchCommandTest` change is unrelated to the HOME fix and is already covered by #9412.

Closing in favor of #9229 — but appreciate the effort!

PR #9237, opened 2026-08-04
fix(ui): use variableName in validation error display (fixes #8729)

### LINKED ISSUE #8729: fix(ui): prompt template validation errors don't show the variable name

YuvrajVerma09arch:
## Description

Fixes the validation error display in PromptTemplateTestPanel.

The render endpoint returns validation errors with a `variableName` field, but the UI was using `path` (which doesn't exist in the API response).

### Changes
1. **`ui/ui-app/src/models/RenderPromptResponse.ts`** - Updated `RenderPromptValidationError` interface:
   - Changed `path?: string` to `variableName?: string`
   - Added `expectedType?: string` and `actualType?: string` to match API response

2. **`ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx`** - Updated validation error display:
   - Changed `ve.path` to `ve.variableName` in the alert rendering

### Related
- Fixes #8729
- This fix was originally part of PR #9126 which was closed because the other 2 fixes (boolean defaults, number input zero value) were already merged via other PRs.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -228,7 +228,15 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                     <Alert variant="warning" title="Validation Errors" className="validation-errors">
                         <ul>
                             {validationErrors.map((ve, i) => (
-                                <li key={i}>{ve.path ? `${ve.path}: ` : ""}{ve.message}</li>
+                                <li key={i}>
+                                {ve.variableName ? `${ve.variableName}: ` : ""}
+                                {ve.message}
+                                {ve.expectedType && ve.actualType && (
+                                    <span style={{ color: "var(--pf-t--global--color--600)", marginLeft: "0.5rem" }}>
+                                        (Expected: {ve.expectedType}, Actual: {ve.actualType})
+                                    </span>
+                                )}
+                            </li>
                             ))}
                         </ul>
                     </Alert>
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -229,14 +229,14 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                         <ul>
                             {validationErrors.map((ve, i) => (
                                 <li key={i}>
-                                {ve.variableName ? `${ve.variableName}: ` : ""}
-                                {ve.message}
-                                {ve.expectedType && ve.actualType && (
-                                    <span style={{ color: "var(--pf-t--global--color--600)", marginLeft: "0.5rem" }}>
-                                        (Expected: {ve.expectedType}, Actual: {ve.actualType})
-                                    </span>
-                                )}
-                            </li>
+                                    {ve.variableName ? `${ve.variableName}: ` : ""}
+                                    {ve.message}
+                                    {ve.expectedType && ve.actualType && (
+                                        <span style={{ color: "var(--pf-t--global--color--600)", marginLeft: "0.5rem" }}>
+                                            (Expected: {ve.expectedType}, Actual: {ve.actualType})
+                                        </span>
+                                    )}
+                                </li>
                             ))}
                         </ul>
                     </Alert>
--- a/ui/ui-app/src/models/RenderPromptResponse.ts
+++ b/ui/ui-app/src/models/RenderPromptResponse.ts
@@ -1,6 +1,8 @@
 export interface RenderPromptValidationError {
-    path?: string;
+    variableName?: string;
     message: string;
+    expectedType?: string;
+    actualType?: string;
 }
 
 export interface RenderPromptResponse {


hussainjamal760:
Thanks  @YuvrajVerma09arch  for the PR! The fix for switching path to variableName looks great and functional. However, I noticed a couple of things that need to be addressed before this can be merged:

1. Missing DCO Sign-off ❌
The Apicurio project requires a DCO (Developer Certificate of Origin) sign-off on all commits. Your commit is currently missing the Signed-off-by line. You can easily fix this by amending your commit:

bash
git commit --amend -s
git push --force-with-lease origin <your-branch-name>
2. Enhancing the UI for expectedType and actualType 💡
I see that you correctly updated the RenderPromptValidationError interface in ui/ui-app/src/models/RenderPromptResponse.ts to include expectedType and actualType to match the API response. That's a great addition!

However, since we now have this valuable data from the API, we should actually display it to the user in the UI instead of just updating the model. It would significantly improve the UX if the user could see exactly what type was expected vs what was received.

Consider updating the rendering logic in PromptTemplateTestPanel.tsx to something like this:

tsx
<li key={i}>
    {ve.variableName ? <strong>{ve.variableName}: </strong> : ""}
    {ve.message}
    {ve.expectedType && ve.actualType && (
        <span style={{ marginLeft: "5px", color: "var(--pf-v5-global--Color--200)" }}>
            (Expected: <code>{ve.expectedType}</code>, Actual: <code>{ve.actualType}</code>)
        </span>
    )}
</li>
(Feel free to adjust the styling classes to match the project's standard UI system).

Disclaimer: I am just another contributor helping maintainers review PRs faster to keep the queue moving. The final decision and approval will, of course, be up to the core project maintainers!

ui/ui-app/src/models/RenderPromptResponse.ts
```diff
@@ -1,6 +1,8 @@
 export interface RenderPromptValidationError {
-    path?: string;
+    variableName?: string;
```
paoloantinori, line 2: This fix assumes the render endpoint returns the variable name as `variableName`. Worth confirming against the actual response (a sample payload or the backend DTO) so #8729 is actually fixed rather than swapped for another field that doesn't exist in the payload. If the backend still emits `path`, or uses a different key, this change preserves the original bug under a new name.

paoloantinori (commented):
Solid, minimal fix that corrects the field-name mismatch so validation errors render the offending variable name. Two things worth a look before merge:

- `ui/ui-app/src/models/RenderPromptResponse.ts:2` — the correctness of the whole change rests on the render endpoint returning `variableName`. The PR description states this, but confirm it against the actual API payload (or the backend DTO) so #8729 is truly resolved rather than swapped for another field name that is also absent from the response.
- `ui/ui-app/src/models/RenderPromptResponse.ts:4-5` — `expectedType` and `actualType` are added to the interface but never read anywhere in this change. Either surface them in the alert (expected vs. actual type is genuinely useful when debugging a type validation error) or omit them to keep the diff minimal.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -228,7 +228,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                     <Alert variant="warning" title="Validation Errors" className="validation-errors">
                         <ul>
                             {validationErrors.map((ve, i) => (
-                                <li key={i}>{ve.path ? `${ve.path}: ` : ""}{ve.message}</li>
+                                <li key={i}>{ve.variableName ? `${ve.variableName}: ` : ""}{ve.message}</li>
```
paoloantinori: The interface now models `expectedType` and `actualType`, but they aren't rendered anywhere. Surfacing them here, e.g. appending `(expected {ve.expectedType}, got {ve.actualType})` when present, would make type validation errors actionable instead of merely relabeling the field. Otherwise, consider dropping the two new fields to keep the change scoped to the #8729 fix.

EricWittmann:
Thanks for the fix, @YuvrajVerma09arch! Clean and well-scoped change. Merged.

PR #9242, opened 2026-08-04
Fix/content accepters null content type

### LINKED ISSUE #9203: fix(schema-util): prevent NPE on null contentType in ContentAccepters and enforce Locale.ROOT
## Description

In `schema-util`, several `ContentAccepter` implementations invoke `content.getContentType().toLowerCase()` without checking if `contentType` is `null`.

When `TypedContent` is created without a content type header, `content.getContentType()` returns `null`. Calling `.toLowerCase()` throws an unhandled `NullPointerException`, which gets caught in `catch (Exception e)` and causes `acceptsContent()` to return `false` — erroneously rejecting valid schemas.

### Affected Classes
- `OpenApiContentAccepter`
- `AsyncApiContentAccepter`
- `XmlContentAccepter`
- `XsdContentAccepter`
- `WsdlContentAccepter`
- `GraphQLContentAccepter`

### Expected Behavior
`ContentAccepter` should handle `contentType == null` gracefully by checking `contentType != null` before calling `.toLowerCase(Locale.ROOT)`, allowing fallback content inspection/parsing to succeed.

### Proposed Solution
1. Add `contentType != null` null guards before String lowercasing in affected accepters.
2. Use `Locale.ROOT` for all `.toLowerCase()` calls.
3. Add unit tests for `acceptsContent()` with `null` `contentType`.

AyushSrivastava1818:
## What type of PR is this?

/kind bug

## What this PR does / why we need it

Several `ContentAccepter` implementations in `schema-util` called `content.getContentType().toLowerCase()` without checking whether `contentType` was `null`.

When `TypedContent` is created without a `Content-Type` header, `getContentType()` returns `null`. Calling `toLowerCase()` on a `null` value results in a `NullPointerException`, which is caught by the surrounding `catch (Exception)` block, causing `acceptsContent()` to incorrectly return `false` instead of falling back to content inspection.

This PR adds null checks before lowercasing the content type in the affected `ContentAccepter` implementations and standardizes all lowercase conversions to use `Locale.ROOT` for locale-independent behavior.

## Which issue(s) this PR fixes

Fixes #9203 

## Special notes for your reviewer

- No behavioral changes when a valid `Content-Type` is present.
- Prevents `NullPointerException` when `contentType` is `null`.
- Preserves the existing fallback content inspection logic.
- Replaces locale-dependent `toLowerCase()` calls with `toLowerCase(Locale.ROOT)` for consistent behavior.
- Added unit tests covering `acceptsContent()` with a `null` content type.
- Verified locally by running the relevant module tests.

## AI assistance disclosure

I consulted ChatGPT and Claude to help analyze the issue and discuss the implementation approach. The implementation, tests, validation, commit message, and final code changes were completed and verified by me. I reviewed all generated suggestions, ran the relevant tests locally, and take full responsibility for the final contribution.

## Does this PR introduce a user-facing change?

```release-note
NONE
```

### DIFF
--- a/schema-util/openapi/src/main/java/io/apicurio/registry/openapi/content/OpenApiContentAccepter.java
+++ b/schema-util/openapi/src/main/java/io/apicurio/registry/openapi/content/OpenApiContentAccepter.java
@@ -5,6 +5,7 @@ import io.apicurio.registry.content.ContentAccepter;
 import io.apicurio.registry.content.TypedContent;
 import io.apicurio.registry.content.util.ContentTypeUtil;
 
+import java.util.Locale;
 import java.util.Map;
 
 public class OpenApiContentAccepter implements ContentAccepter {
@@ -16,12 +17,13 @@ public class OpenApiContentAccepter implements ContentAccepter {
             JsonNode tree = null;
             // If the content is YAML, then convert it to JSON first (the data-models library only accepts
             // JSON).
-            if (contentType.toLowerCase().contains("yml") || contentType.toLowerCase().contains("yaml")) {
+            if (contentType != null && (contentType.toLowerCase(Locale.ROOT).contains("yml")
+                    || contentType.toLowerCase(Locale.ROOT).contains("yaml"))) {
                 tree = ContentTypeUtil.parseYaml(content.getContent());
             } else {
                 tree = ContentTypeUtil.parseJson(content.getContent());
             }
-            if (tree.has("openapi") || tree.has("swagger")) {
+            if (tree != null && (tree.has("openapi") || tree.has("swagger"))) {
                 return true;
             }
         } catch (Exception e) {
--- a/schema-util/openapi/src/main/java/io/apicurio/registry/openapi/content/OpenApiContentAccepter.java
+++ b/schema-util/openapi/src/main/java/io/apicurio/registry/openapi/content/OpenApiContentAccepter.java
@@ -14,11 +14,11 @@ public class OpenApiContentAccepter implements ContentAccepter {
     public boolean acceptsContent(TypedContent content, Map<String, TypedContent> resolvedReferences) {
         try {
             String contentType = content.getContentType();
+            String lc = contentType != null ? contentType.toLowerCase(Locale.ROOT) : null;
             JsonNode tree = null;
             // If the content is YAML, then convert it to JSON first (the data-models library only accepts
             // JSON).
-            if (contentType != null && (contentType.toLowerCase(Locale.ROOT).contains("yml")
-                    || contentType.toLowerCase(Locale.ROOT).contains("yaml"))) {
+            if (lc != null && (lc.contains("yml") || lc.contains("yaml"))) {
                 tree = ContentTypeUtil.parseYaml(content.getContent());
             } else {
                 tree = ContentTypeUtil.parseJson(content.getContent());
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/content/accepter/ContentAccepterNullContentTypeTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/content/accepter/ContentAccepterNullContentTypeTest.java
@@ -1,0 +1,80 @@
+package io.apicurio.registry.content.accepter;
+
+import io.apicurio.registry.asyncapi.content.AsyncApiContentAccepter;
+import io.apicurio.registry.content.ContentHandle;
+import io.apicurio.registry.content.TypedContent;
+import io.apicurio.registry.graphql.content.GraphQLContentAccepter;
+import io.apicurio.registry.openapi.content.OpenApiContentAccepter;
+import io.apicurio.registry.wsdl.content.WsdlContentAccepter;
+import io.apicurio.registry.xml.content.XmlContentAccepter;
+import io.apicurio.registry.xsd.content.XsdContentAccepter;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+
+public class ContentAccepterNullContentTypeTest {
+
+    @Test
+    public void testOpenApiContentAccepter_nullContentType() {
+        String openApiJson = "{\n" +
+                "  \"openapi\": \"3.0.0\",\n" +
+                "  \"info\": {\n" +
+                "    \"title\": \"Sample API\",\n" +
+                "    \"version\": \"0.1.0\"\n" +
+                "  },\n" +
+                "  \"paths\": {}\n" +
+                "}";
+        TypedContent content = TypedContent.create(ContentHandle.create(openApiJson), null);
+        OpenApiContentAccepter accepter = new OpenApiContentAccepter();
+
+        Assertions.assertTrue(accepter.acceptsContent(content, Collections.emptyMap()));
+    }
+
+    @Test
+    public void testAsyncApiContentAccepter_nullContentType() {
+        String asyncApiJson = "{\n" +
+                "  \"asyncapi\": \"2.0.0\",\n" +
+                "  \"info\": {\n" +
+                "    \"title\": \"Sample AsyncAPI\",\n" +
+                "    \"version\": \"0.1.0\"\n" +
+                "  }\n" +
+                "}";
+        TypedContent content = TypedContent.create(ContentHandle.create(asyncApiJson), null);
+        AsyncApiContentAccepter accepter = new AsyncApiContentAccepter();
+
+        Assertions.assertTrue(accepter.acceptsContent(content, Collections.emptyMap()));
+    }
+
+    @Test
+    public void testXmlContentAccepter_nullContentType_noNpe() {
+        TypedContent content = TypedContent.create(ContentHandle.create("<root/>"), null);
+        XmlContentAccepter accepter = new XmlContentAccepter();
+
+        Assertions.assertDoesNotThrow(() -> accepter.acceptsContent(content, Collections.emptyMap()));
+    }
+
+    @Test
+    public void testXsdContentAccepter_nullContentType_noNpe() {
+        TypedContent content = TypedContent.create(ContentHandle.create("<schema xmlns=\"http://www.w3.org/2001/XMLSchema\"/>"), null);
+        XsdContentAccepter accepter = new XsdContentAccepter();
+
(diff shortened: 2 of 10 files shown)


AyushSrivastava1818:
Reopening this fix after my previous PR was closed due to the repository's one-open-PR-per-contributor policy. My other pending PR has since been merged, so I'm resubmitting this implementation for the same issue (#9202).

paoloantinori (commented):
The null guards and `Locale.ROOT` normalization are correct and low risk, and for `OpenApiContentAccepter` and `AsyncApiContentAccepter` they genuinely restore content inspection when the Content-Type header is missing. The change is sound where it matters; the points below concern scope accuracy and test value rather than correctness.

- `XmlContentAccepter`, `XsdContentAccepter`, `WsdlContentAccepter`, and `GraphQLContentAccepter` have no observable behavior change from this PR. The `NullPointerException` was already swallowed by the existing `catch (Exception)`, so `acceptsContent` returned `false` before and after. The "fall back to content inspection" framing applies only to the OpenApi and AsyncApi accepters, which parse unconditionally; the other four gate parsing behind the content-type check and still return `false` for a null type.
- Because the NPE is swallowed inside the try, the four `assertDoesNotThrow` tests pass against the original code as well, so they neither exercise the fix nor guard against regression. Asserting the return value (for example `assertFalse`) would at least pin the current contract.
- The `assertTrue` tests for OpenApi and AsyncApi are the right shape and do catch the regression.
- In the YAML branch, `contentType.toLowerCase(Locale.ROOT)` is computed twice. Lowercase once into a local; both `contains` checks are still needed because `'yaml'` is not a substring of `'yml'` (OpenApiContentAccepter.java:20, duplicated in AsyncApiContentAccepter).
- Worth confirming the other `ContentAccepter` implementations in the repo (Avro, Protobuf, JsonSchema, Kafka Connect, and so on) do not carry the same unguarded `contentType.toLowerCase()`. I could not verify this from the diff alone.
- Worth confirming whether `ContentTypeUtil.parseJson`/`parseYaml` can return null. If they only throw on bad input, the new `tree != null` guard in OpenApi and AsyncApi is unreachable.

schema-util/util-provider/src/test/java/io/apicurio/registry/content/accepter/ContentAccepterNullContentTypeTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+        Assertions.assertTrue(accepter.acceptsContent(content, Collections.emptyMap()));
+    }
+
+    @Test
+    public void testXmlContentAccepter_nullContentType_noNpe() {
+        TypedContent content = TypedContent.create(ContentHandle.create("<root/>"), null);
+        XmlContentAccepter accepter = new XmlContentAccepter();
+
+        Assertions.assertDoesNotThrow(() -> accepter.acceptsContent(content, Collections.emptyMap()));
```
paoloantinori: This `assertDoesNotThrow` (and the Xsd, Wsdl, and GraphQL cases) passes against the original code as well. The `NullPointerException` from `contentType.toLowerCase()` is thrown inside the `try` and swallowed by `catch (Exception)`, so `acceptsContent` already returns `false` without propagating. The test therefore does not exercise the fix or protect against regression. Assert the return value (for example `assertFalse(accepter.acceptsContent(content, Collections.emptyMap()))`) to pin the actual contract, or note explicitly that these four accepters have no behavioral change for a null content type.

schema-util/openapi/src/main/java/io/apicurio/registry/openapi/content/OpenApiContentAccepter.java
```diff
@@ -16,12 +17,13 @@ public boolean acceptsContent(TypedContent content, Map<String, TypedContent> re
             JsonNode tree = null;
             // If the content is YAML, then convert it to JSON first (the data-models library only accepts
             // JSON).
-            if (contentType.toLowerCase().contains("yml") || contentType.toLowerCase().contains("yaml")) {
+            if (contentType != null && (contentType.toLowerCase(Locale.ROOT).contains("yml")
```
paoloantinori: `contentType.toLowerCase(Locale.ROOT)` is evaluated twice on this branch. Lowercase once into a local, for example `String lc = contentType != null ? contentType.toLowerCase(Locale.ROOT) : null;`, then test `lc != null && (lc.contains("yml") || lc.contains("yaml"))`. Keep both `contains` checks: `'yaml'` is not a substring of `'yml'`, so each matches different inputs. The same duplication exists in `AsyncApiContentAccepter`.

AyushSrivastava1818:
Thanks for the detailed review @paoloantinori ! I've pushed an update addressing the feedback:

Lowercasing optimization: OpenApiContentAccepter, AsyncApiContentAccepter, and OpenRpcContentAccepter now lowercase the content type once into a local variable and reuse it for the yml/yaml checks.
Test assertions: Updated the XML, XSD, WSDL, and GraphQL tests to assert the existing false contract instead of only verifying that no exception is thrown.
Audit: I checked the remaining ContentAccepter implementations (Avro, Protobuf, JsonSchema, Kafka Connect, Thrift, Iceberg, etc.) and didn't find the same unguarded contentType.toLowerCase() pattern elsewhere.
parseJson/parseYaml: I also verified these paths. The tree != null guards have been kept because ObjectMapper.readTree() can return null for inputs such as empty content or JSON null.
Rebased: The branch has also been rebased onto the latest main.

Thanks again for the detailed feedback!

paoloantinori:
Thanks for the update. I re-read the current head and the changes address the feedback:

- Lowercasing is now done once per accepter (lc local in OpenApi/AsyncApi/OpenRpc) and reused for the yml/yaml checks, so the repeated toLowerCase calls are gone.
- The test now asserts the existing false contract for XML, XSD, WSDL, and GraphQL with a null content type rather than only checking that no exception is thrown, so a future regression that wrongly accepts would fail.
- The null guards and Locale.ROOT normalization are consistent across all seven accepters.

AyushSrivastava1818:
Hi @EricWittmann @paoloantinori — looked into the 3 failing test jobs (app-kafkasql, app-other, app-rest) plus the resulting Verification Gate failure. None of them appear related to this PR's changes (which only touch schema-util's ContentAccepter classes and tests):

app-kafkasql: Fails on a KafkaSqlKeyDeserializer JSON parse error from non-JSON records in the test Kafka topic, cascading into RegistryEventsTest.lookupEvent timeouts.
app-other: Fails with OutOfMemoryError: Java heap space in the Surefire test forks — a CI resource/config issue, not a test assertion failure.
app-rest: Fails on HttpCompressionDisabledTest — the app-level compression kill-switch doesn't prevent Quarkus/Undertow's transport-level gzip from kicking in.
Verification Gate: Fails only because it depends on the above three.

paoloantinori:
Thanks for the triage @AyushSrivastava1818 — you're right, the failures are not from this PR.

I looked at the three red jobs on this head:

- **app-kafkasql** — fails in `KafkaSqlEventsTest > RegistryEventsTest.*` with a `KafkaSqlKeyDeserializer` JSON-parse error (`Illegal character (CTRL-CHAR, code 1)`), which cascades into `lookupEvent` timeouts. The deserializer is choking on non-JSON records in the test Kafka topic, which is unrelated to `ContentAccepter` null-handling.
- **app-other / app-rest** — same class of unrelated infrastructure failures, not in any module this PR touches.

This PR only modifies `schema-util/**/ContentAccepter.java` and adds `ContentAccepterNullContentTypeTest`. None of those touch the kafkasql event/deserializer path. The `Unit Tests (app-sql)` and `Unit Tests (app-kubernetesops)` jobs that exercise the modules closest to these changes are green.

The "Verify" workflow is currently red on `main` too (last run failed), so the Verification Gate is not a signal specific to this PR. This looks like flaky/infra rather than a regression. A `/retry` or re-run of the failing jobs is the right next step; no code changes needed here.

jsenko:

EricWittmann (commented):
Thanks for the fix, @AyushSrivastava1818 — the null guards and `Locale.ROOT` normalization are correct, and I can see you addressed the earlier review feedback (local `lc` variable, `assertFalse` in the no-behavior-change tests). Nice work.

A few things to address before we can merge:

1. **Wrong issue link** — the PR says "Fixes #9202" but that's a Renovate docker digest PR, not a bug report about null content types. Please correct the link to the right issue, or remove the "Fixes" line if there isn't a separate issue for this.

2. **Incomplete coverage** — while you're fixing this pattern, there are a few other files with the same unguarded `contentType.toLowerCase()`:
   - `ProtobufContentAccepter` already has a null check but uses `.toLowerCase()` without `Locale.ROOT` — please add `Locale.ROOT` for consistency.
   - `ApicurioDataModelsContentDereferencer` (line 91) has `contentType.toLowerCase()` with neither a null check nor `Locale.ROOT` — worth fixing here if `contentType` can be null in that code path.

3. **Missing test for OpenRpcContentAccepter** — the PR modifies `OpenRpcContentAccepter` but the test class doesn't include a test case for it. Please add one.

Please push the updates and we'll take another look.

AyushSrivastava1818:
Hi @EricWittmann — thanks for the thorough review! I've pushed updates addressing all the points:

- **Issue link corrected** — `Fixes #9202` was indeed pointing to the wrong (Renovate) PR. Updated it to reference **#9203**, which is the actual bug report for this NPE issue.

- **Incomplete coverage, now addressed:**
  - **`ProtobufContentAccepter`**: added `Locale.ROOT` to the existing `.toLowerCase()` call for consistency with the other accepters (it already had the null check, just missing the locale normalization).
  - **`ApicurioDataModelsContentDereferencer`** (line 91): traced the callers to confirm `contentType` can indeed be `null` on this path, so I added the same `contentType != null` guard + `Locale.ROOT` pattern used elsewhere in this PR, via a local `lc` variable to avoid double-lowercasing.
  - **`OpenRpcContentAccepter`**: added `testOpenRpcContentAccepter_nullContentType()` to `ContentAccepterNullContentTypeTest.java`, asserting it falls back to content inspection and returns `true` for valid OpenRPC content with a `null` content type—matching the existing OpenAPI/AsyncAPI behavior.

Ran the full `ContentAccepterNullContentTypeTest` suite locally (**7 tests, all passing**) along with checkstyle across the affected modules—no violations.

Let me know if anything else needs adjusting!

EricWittmann:
Thanks for the contribution, @AyushSrivastava1818! Great work on this fix — the null content type handling and the Locale.ROOT standardization are both solid improvements. Welcome to the project!

PR #9247, opened 2026-08-04
fix(ci): narrow app-other exclusions for storage and event testsFix unit test shard coverage

### LINKED ISSUE #9147: Unit-test shard filters silently skip new storage.*/event.* subpackages (zero-shard execution)
Found during the review of the #9137 docs fix. The 7-shard unit-test matrix in `verify-unit-tests.yaml` has a silent coverage gap:

- The five storage/eventing include-shards cover only specific subtrees (`storage.impl.{sql,search,polling,readonly,kafkasql,gitops,kubernetesops}.**`, `storage.decorator.**`, `storage.dto.**`, `event.sql.**`, `event.kafkasql.**`).
- The catch-all `app-other` shard excludes the whole `storage.**` and `event.**` trees.
- Therefore any NEW subpackage under `io.apicurio.registry.storage` or `io.apicurio.registry.event` (e.g. a new storage variant, or `event.nats.**`) matches no include filter and is excluded from the catch-all: it runs in **zero shards**.
- `-Dsurefire.failIfNoSpecifiedTests=false` (needed for the sharding scheme) makes this completely silent: no failure, no warning.

Today there is no orphan package (verified: `storage/util` holds only test profiles, `event/` has only `sql/` and `kafkasql/`), so this is a latent trap, not a live coverage hole.

Options:
1. A CI guard: a cheap job (or a step in one shard) that computes the union of the shard filters against the actual test-package tree and fails on uncovered packages.
2. Restructure `app-other` to exclude only the subtrees the include-shards actually cover, so new subpackages fall into the catch-all by default.
3. At minimum the docs warning (being added in #9137's PR) telling contributors to wire new storage/event subpackages into a shard.

Option 2 looks strongest (fail-safe by construction). Related nit from the same review: the 600 s per-shard timing threshold warns on every `app-other` run (~14 min typical), so the threshold is currently noise; consider per-shard thresholds while touching this.

SaikiaMan:
## Summary

The `app-other` shard currently excludes all `storage.*` and `event.*` packages.

This creates a latent coverage gap where future storage or event subpackages that are not explicitly assigned to a dedicated shard execute in zero shards.

This change narrows the exclusions to only the packages already covered by dedicated shards, preserving current behavior while allowing future packages to fall back to `app-other` automatically.

## Verification

- Verified package-to-shard mapping.
- Confirmed current storage/event packages remain covered exactly once.
- Confirmed `storage.util` is no longer excluded.
- Validated the workflow YAML.
- Verified the updated Surefire `-Dtest` filter is accepted.


SaikiaMan:
Hi @paoloantinori , I opened PR #9247 to address the latent coverage gap discussed in the issue #9147.

The change narrows the app-other shard exclusions from the broad storage.** / event.** patterns to only the packages that are already covered by the dedicated shards. The idea is to preserve the current sharding while allowing any future storage.* or event.* subpackages to automatically fall back to app-other instead of being silently skipped.

I verified that the current storage and event packages are still assigned to exactly one shard, storage.util is no longer excluded, and the updated Surefire filter is accepted.

If this isn't the direction you had in mind, or if you'd prefer a different approach, please let me know. I'm happy to adjust the implementation based on your feedback.

paoloantinori:
Thanks for the write-up. I re-checked the new exclusion list against the four dedicated shards and it lines up exactly: every package the positive-filter shards include (storage.impl.sql, search, polling, readonly, storage.decorator, storage.dto, event.sql for app-sql; storage.impl.kafkasql and event.kafkasql for app-kafkasql; storage.impl.gitops; storage.impl.kubernetesops; plus noprofile.rest and noprofile.ccompat for app-rest) is present in the negation, and nothing extra is excluded. So there is no double-execution and no zero-execution gap for packages outside that set.

One consequence worth being aware of, not a blocker: packages that previously fell under the broad storage.** / event.** blanket but were never assigned to a dedicated shard (storage.util being the one you called out) will now newly execute in app-other. That is exactly the gap this fixes, but it means app-other may pick up tests that have not run in CI before, which could surface failures or nudge the shard's runtime up. If any of those newly-covered tests are slow, the timing check step will flag it. Worth a glance at the first run after merge.

From my side this looks correct and ready.

Nachiket-Roy:
The "latent coverage gap" here is theoretical, not something we hit in practice:

- Any new storage variant or event structure (under storage.impl.*, event.*, etc.) has to be explicitly wired into a dedicated shard (app-sql, app-kafkasql, ...) since those tests need specific backend containers. So a test not wired into a shard won't run correctly regardless of how app-other excludes it.
- The current two-pattern exclude (!io.apicurio.registry.storage.**, !io.apicurio.registry.event.**) avoids Surefire parsing issues and keeps the command line simple.
- It's the simplest to maintain and has zero CI performance/complexity overhead.

Alternative considered: replace with narrower wildcards, e.g.

```bash 
Dtest=!io.apicurio.registry.noprofile.rest.**,!io.apicurio.registry.noprofile.ccompat.**,!io.apicurio.registry.storage.impl.**,!io.apicurio.registry.storage.decorator.**,!io.apicurio.registry.storage.dto.**,!io.apicurio.registry.event.**
```
This is more precise (won't silently exclude a hypothetical future storage.somethingelse.**), but costs more to maintain, is harder to scan, and doesn't actually close the gap or improve CI, since new tests still need to be wired into a shard manually either way.

I would lean towards not making any changes

SaikiaMan:
@paoloantinori @jsenko  I investigated the CI failures. The app-kafkasql failure is due to KafkaSqlEventsTest timing out while waiting for events, and app-other appears to be running into a JVM heap issue. Since this PR only changes test sharding, I don't think either failure directly points to the exclusion logic. Another contributor also suggested keeping the current broad exclusions because storage/event tests must be explicitly assigned to backend-specific shards regardless. Please let me know if you'd like me to rebalance the shards, or if these CI failures should be handled separately.

SaikiaMan:
@Nachiket-Roy I agree. The broader exclusions are intentional because storage/event tests require dedicated backend-specific shards. Any new storage or event test would still need to be explicitly assigned to the appropriate shard, so narrowing the exclusions doesn't materially improve correctness while making the configuration more complex. I'd prefer to keep the current approach unless we identify a concrete case where it causes incorrect test distribution. Thanks tho it really means a lot that you even took the time to understand and reply to my pr, really appriciate your help if you have further suggestion please let me know

paoloantinori:
Thanks for the investigation — agree that neither CI failure points to this change. The PR is a one-line narrowing of test-filter exclusions in `verify-unit-tests.yaml`, so it cannot affect `KafkaSqlEventsTest` timing (that lives in the `app-kafkasql` shard, which this filter does not touch) or heap pressure in `app-other` (the test count there is unchanged; only which packages are excluded from it).

If `app-other` is now hitting a JVM heap issue, it is likely the same `-Xmx4g` shard under a heavier test load unrelated to this PR. Re-running the failed jobs (`/retest` or re-running the workflow run) is the right next step. From my side the exclusion list is still correct and matches the dedicated shards as I checked earlier.

jsenko:

SaikiaMan:
@paoloantinori @jsenko Thanks for the heads-up! I've rebased this branch onto the latest main to pick up the fix from #9287 and pushed the updated branch. The workflows have been re-triggered, so the previous app-other OOM issue should no longer affect this PR.

.github/workflows/verify-unit-tests.yaml
```diff
@@ -65,7 +65,11 @@ jobs:
           # shard at all. Every exclusion below must correspond to another shard's inclusion.
           - name: app-other
             modules: "-pl app -am -Dfull -DcliSkipNative"
+<<<<<<< HEAD
             test-filter: "-Dtest=!io.apicurio.registry.noprofile.rest.**,!io.apicurio.registry.noprofile.ccompat.**,!io.apicurio.registry.storage.**,!io.apicurio.registry.event.**,!io.apicurio.registry.auth.**,!io.apicurio.registry.rbac.**,!io.apicurio.registry.limits.**,!io.apicurio.registry.tls.**,!io.apicurio.registry.headers.**,!io.apicurio.registry.cors.**,!io.apicurio.registry.metrics.**,!io.apicurio.registry.search.**"
+=======
```
paoloantinori: The rebase left unresolved merge-conflict markers in this file (lines 68-72): `<<<<<<< HEAD`, `=======`, and `>>>>>>> 154ff1d11` are still present, with two `test-filter` entries for `app-other`. The workflow YAML will not parse in this state. Remove the markers and keep the narrowed exclusion (the per-package list).

paoloantinori (commented):
The rebase left unresolved merge conflict markers in `verify-unit-tests.yaml` at the current head (`5ed27ae`). The file will not parse as YAML in this state, so the workflow cannot run — the re-triggered run will fail at parse time before any test shard starts.

Lines 68, 70, and 72 carry the `<<<<<<< HEAD`, `=======`, and `>>>>>>>` markers, with two competing `test-filter` lines for the `app-other` shard (lines 69 and 71). Only one `test-filter` can survive; from the PR intent it should be the narrowed one (line 71, the explicit per-package exclusions), with the markers themselves removed.

Worth resolving before the next push. The narrower exclusion list itself looks correct against the dedicated shards, so once the conflict is cleaned up this should be ready.

paoloantinori:
Posted a formal review above flagging the unresolved merge-conflict markers in verify-unit-tests.yaml. Nothing to add beyond that.

SaikiaMan:
@paoloantinori Thanks for catching that. During the rebase I had accidentally left the merge conflict markers in verify-unit-tests.yaml. I've removed the markers, kept the narrowed app-other exclusion list, and pushed the fix. The workflow has been re-triggered.

paoloantinori:
Thanks — confirmed, the merge-conflict markers are gone at `f9547bb` and the diff is back to the intended one-line narrowing of the `app-other` exclusions. Nothing else to add from my side; once CI goes green this looks good to go.

SaikiaMan:
Hi @paoloantinori @jsenko , I investigated the CI failure and found that the partition issue was caused by app-other no longer excluding the dedicated app-auth, app-transport, and app-metrics shards after I narrowed the storage/event exclusions. I've restored those exclusions while keeping the intended storage/event narrowing, and verified locally with verify-test-shards.py that the partition is now correct (run MORE than once: 0, RESULT: PARTITION OK).

After pushing the fix, CI now gets past the partition validation. The remaining failure is HttpCompressionDisabledTest, which appears unrelated to this workflow-only change. The test is failing because it receives a Content-Encoding: gzip header when it expects none.

Could you let me know if this is a known regression or a flaky test, or if you'd prefer that I investigate it further?

paoloantinori:
Thanks @SaikiaMan. `HttpCompressionDisabledTest` is unrelated to this PR. The test lives in `app/src/test/java/.../rest/v3/` (the `app` module's `noprofile.rest` package), and this PR only touches `.github/workflows/verify-unit-tests.yaml`, so it cannot affect the compression code path.

The failure you're describing, the test receiving a `Content-Encoding: gzip` header when it expects none, points at `HttpCompressionWriterInterceptor` or the `apicurio.rest.compression.enabled=false` profile override, not at test partitioning. Worth a `/retry` first to confirm it's a flake; if it reproduces deterministically it should be triaged against the compression writer interceptor, separately from this workflow change.

SaikiaMan:
Hi @EricWittmann @jsenko , with @paoloantinori 's help I tracked down and fixed the shard partitioning issue, and all CI checks are now passing. The unrelated HttpCompressionDisabledTest also passed on retry. When you have a chance, I'd appreciate a review. Thanks!

SaikiaMan:
Thanks a lot @paoloantinori you really helped me a lot

paoloantinori:
Glad it helped, and nice work tracking down the shard partitioning issue. Good to see the `app-other` exclusions narrowed cleanly and CI green.

SaikiaMan:
@paoloantinori thank you for your kind words sir

paoloantinori:
Happy to help, and nice work getting it across the line. The narrowed exclusion list reads cleanly to me.

SaikiaMan:
Hi @jsenko @EricWittmann , just wanted to follow up on my previous PR when you get a chance. All the requested changes have been addressed and the checks are passing.

If everything looks good from your side, I’d really appreciate it if you could review and merge it. Once that’s settled, I can move on to the next issue.

Thanks!

SaikiaMan:
@carlesarnal @EricWittmann @jsenko please have a look when you get the time and merge it into main if you find it good to go

SaikiaMan:
@jsenko @EricWittmann Please merge it, if all looks fine

PR #9248, opened 2026-08-04
fix(ui): replace hardcoded white background in ResizableOverviewDrawer with PF theme variable

### LINKED ISSUE #9244: fix(ui): replace hardcoded white background in ResizableOverviewDrawer with PF theme variable
## What

`ResizableOverviewDrawer.tsx` uses a hardcoded `style={{ backgroundColor: "white" }}` on the `DrawerContent` component. This will break in dark mode or any custom theming.

## Where

`ui/ui-app/src/app/components/common/ResizableOverviewDrawer.tsx` — line with:
```tsx
style={{ backgroundColor: "white" }}
```

## Suggested fix

Replace with a PatternFly CSS variable, e.g.:
```tsx
style={{ backgroundColor: "var(--pf-t--global--background--color--primary)" }}
```

Or better yet, move it to `ResizableOverviewDrawer.css` so all styling lives in one place.

## Context

This was carried over from the pre-existing inline style in the individual overview pages (Artifact, Version, Group) when they were consolidated into the shared `ResizableOverviewDrawer` component in #9210.

Faisal77666:
Fixes #9244

## What
`ResizableOverviewDrawer.tsx` had a hardcoded `style={{ backgroundColor: "white" }}`
on the `DrawerContent` component, which would break in dark mode or any custom
theming. This PR removes the inline style and moves the background rule into
`ResizableOverviewDrawer.css` using a PatternFly theme token, as suggested in
the issue.

## Token choice
Used `var(--pf-t--global--background--color--primary--default, #ffffff)`
rather than the `--primary` variant shown in the issue's suggested snippet,
to match the `--default`-suffixed convention already used elsewhere in this
codebase (e.g. `ReferenceGraphNode.css`), and included a fallback value for
consistency with those existing usages.

## Testing
- Verified in light mode: no visual change, background renders identically
  to before.
- Verified via devtools that
  `--pf-t--global--background--color--primary--default` resolves to a real
  color value at runtime.
- Note: this repo currently has no dark-mode implementation (no toggle, no
  `prefers-color-scheme` support found in the codebase), so the dark-mode
  scenario described in the issue can't be visually verified yet — but this
  change ensures the drawer will pick up dark styling automatically once
  theming is added, instead of staying hardcoded white.


Faisal77666:
@EricWittmann please review

Faisal77666:
@EricWittmann ready to merge

EricWittmann:
Thanks for this contribution, @Faisal77666! Clean fix, merged.

PR #9221, opened 2026-08-04
fix(common): specify Locale.ROOT in StringUtil.asLowerCase()

Labreo:
## Description

In `common/src/main/java/io/apicurio/registry/utils/StringUtil.java`, `asLowerCase()` called `value.toLowerCase()` without `Locale.ROOT`. 

This PR updates `StringUtil.asLowerCase()` to pass `Locale.ROOT` in accordance with the project code style conventions and adds a unit test in `StringUtilTest`.

## Contributor Checklist
- [x] Conventional Commits format (`fix(common): specify Locale.ROOT in StringUtil.asLowerCase()`)
- [x] All commits have DCO sign-off (`Signed-off-by: Labreo <74888611+Labreo@users.noreply.github.com>`)
- [x] Unit test added (`StringUtilTest.java`)
- [x] `./mvnw checkstyle:check -pl common` passes
- [x] PR contains no unrelated changes

Fixes #9220

PR #9249, opened 2026-08-04
fix(common): specify Locale.ROOT in StringUtil.asLowerCase()

### LINKED ISSUE #9220: fix(common): specify Locale.ROOT in StringUtil.asLowerCase()
## Description

**Registry Version**: `snapshot` (3.3.2-SNAPSHOT / main)
**Persistence type**: `in-memory`

In `common/src/main/java/io/apicurio/registry/utils/StringUtil.java`, `asLowerCase()` calls `value.toLowerCase()` without `Locale.ROOT`. 

As per the Apicurio Registry contributor guidelines, `toLowerCase()` and `toUpperCase()` calls must always explicitly pass `Locale.ROOT` to avoid locale-dependent behavior across environments.

### Environment

- **Module**: `common`
- **File**: `common/src/main/java/io/apicurio/registry/utils/StringUtil.java`

- **Expected**: `StringUtil.asLowerCase()` uses `value.toLowerCase(Locale.ROOT)`.
- **Actual**: `StringUtil.asLowerCase()` uses default `value.toLowerCase()`.

jjayesh2:
Fixes #9220

`StringUtil.asLowerCase()` was calling `value.toLowerCase()` without specifying a locale, which can cause locale-dependent behavior in certain environments (e.g., Turkish locale).

## Changes
- Added `Locale.ROOT` parameter to `toLowerCase()` call in `StringUtil.asLowerCase()`
- Added `import java.util.Locale;`

## Testing
Verified the module compiles successfully with `mvn clean compile -pl :apicurio-registry-common` (or the actual module you used).

### DIFF
--- a/common/src/main/java/io/apicurio/registry/utils/StringUtil.java
+++ b/common/src/main/java/io/apicurio/registry/utils/StringUtil.java
@@ -2,6 +2,8 @@ package io.apicurio.registry.utils;
 
 import static java.util.Objects.requireNonNull;
 
+import java.util.Locale;
+
 public class StringUtil {
 
     public static boolean isEmpty(String string) {
@@ -16,7 +18,7 @@ public class StringUtil {
         if (value == null) {
             return null;
         }
-        return value.toLowerCase();
+        return value.toLowerCase(Locale.ROOT);
     }
 
     public static String limitStr(String value, int limit, boolean withEllipsis) {
--- a/docs/modules/ROOT/pages/getting-started/assembly-configuring-registry-security.adoc
+++ b/docs/modules/ROOT/pages/getting-started/assembly-configuring-registry-security.adoc
@@ -14,7 +14,23 @@ NOTE: For a list of all available configuration options, see {registry-config-re
 // proc-registry-security-keycloak
 [id="registry-security-keycloak_{context}"]
 == Configuring {registry} authentication and authorization with {keycloak}
+[IMPORTANT]
+====
+When using OIDC providers such as Keycloak or Red Hat SSO, authentication might fail with the following error:
 
+[source]
+----
+error=invalid_request&error_description=Missing parameter: nonce
+----
+
+This happens when the identity provider enforces the `nonce` parameter in authorization requests.
+
+Currently, the Apicurio Registry web console does not include a `nonce` parameter in OIDC authentication requests.
+
+*Workaround*: Disable the nonce requirement in the identity provider configuration.
+
+*Note*: This limitation originates from shared UI components and will be resolved in a future update.
+====
 :_mod-docs-content-type: PROCEDURE
 
 [role="_abstract"]


jjayesh2:
I've submitted a fix for this: https://github.com/Apicurio/apicurio-registry/pull/9249

common/src/main/java/io/apicurio/registry/utils/StringUtil.java
```diff
@@ -16,7 +18,7 @@ public static String asLowerCase(String value) {
         if (value == null) {
             return null;
         }
-        return value.toLowerCase();
+        return value.toLowerCase(Locale.ROOT);
```
paoloantinori, line 21: Correct fix here. As a follow-up, please sweep the rest of `StringUtil` and the common module for a sibling `asUpperCase()` (if it exists) and any other bare `toLowerCase()`/`toUpperCase()` calls, and pass `Locale.ROOT` to those too. Locale-dependent casing is only fully fixed when every normalization site agrees, otherwise comparisons can still drift under a non-ROOT default locale (the same Turkish-i class of bug). If #9220 describes a specific failing path, double-check that path actually routes through this method so the issue is genuinely closed.

paoloantinori (commented):
Correct, low-risk one-liner. Passing `Locale.ROOT` to `toLowerCase()` is the right way to make `asLowerCase()` locale-independent (it eliminates the Turkish-locale i/I problem), and the existing null guard is untouched. No correctness, security, or edge-case issues in the diff itself.

- `common/src/main/java/io/apicurio/registry/utils/StringUtil.java:21`: The fix is sound, but it only covers this single call site. The locale bug this addresses only stays fully fixed if every normalization path agrees, so it is worth auditing the rest of `StringUtil` and the common module for a sibling `asUpperCase()` (if present) and any other bare `toLowerCase()`/`toUpperCase()` calls and giving them the same `Locale.ROOT` argument. Otherwise matching can still drift under a non-ROOT default locale.
- Worth confirming this actually resolves #9220: the linked issue is about casing, so check whether #9220 points at a specific symptom (e.g. an artifact ID or schema-name mismatch) and whether this method is on the code path for it. If the failing path uses a different `toLowerCase()` call, this fix alone will not close the issue.

Labreo:
Isn't this just a duplicate of #9221.Is this version necessary.Just a warning that there are multiple pr's for the same.

hussainjamal760:
@jjayesh2 I would highly recommend using Apicurio Scout before creating an issue or opening a PR. It can help you identify potential overlaps with existing work and might save you days of effort if the contribution turns out to be a duplicate.
A quick check before starting can make the contribution process smoother for everyone. :

Demo Link: https://apicurio-scout.vercel.app/

jjayesh2:
> @jjayesh2 I would highly recommend using Apicurio Scout before creating an issue or opening a PR. It can help you identify potential overlaps with existing work and might save you days of effort if the contribution turns out to be a duplicate. A quick check before starting can make the contribution process smoother for everyone. :
> 
> Demo Link: https://apicurio-scout.vercel.app/

Thanks for the suggestion, will check it out before starting on any new issue!

jjayesh2 (commented):
done

Labreo:
Hey jayesh.Letting you know that i have already opened a PR for this issue which has been already assigned to me at #9221 and as such this PR is redundant.

EricWittmann:
Ah yes.  I see that this is a duplicate PR.  Closing as duplicate.  :+1:

Issue #9245, opened 2026-08-04
test: HttpCompressionDisabledTest.testResponseNotCompressedWhenKillSwitchDisabled is flaky


EricWittmann:
## Flaky test

`io.apicurio.registry.noprofile.rest.v3.HttpCompressionDisabledTest.testResponseNotCompressedWhenKillSwitchDisabled`

## Failure

The test expects the `Content-Encoding` response header to be null (no compression), but intermittently
receives `"gzip"`:

```
java.lang.AssertionError:
1 expectation failed.
Expected header "Content-Encoding" was not null, was "gzip".
```

**Location:** `HttpCompressionDisabledTest.java:58`

## Observed in

- PR #9195 — `Unit Tests / Unit Tests (app-rest)` shard
- [CI run 30942716482, job 92115140512](https://github.com/Apicurio/apicurio-registry/actions/runs/30942716482/job/92115140512)

The failure is unrelated to the PR's changes (label-parsing fix). The test also failed on retry
(same assertion), suggesting a consistent environment issue rather than a timing fluke.

## Likely cause

The kill-switch config that disables HTTP compression may not be taking effect reliably in the test
Quarkus profile, or a shared test instance from a prior test class is leaking compression config.

KartikSuryavanshi:
Opened PR #9251 with a fix: the compression kill switch was a plain `boolean` cached in the `RestConfig` singleton at startup, so Quarkus's app-instance reuse across `@TestProfile` classes could leave it stale at `true` and ignore the test profile's `apicurio.rest.compression.enabled=false`. Converted it to `@Dynamic Supplier<Boolean>` (same pattern as the other runtime-toggleable props) so the value is re-read from the live config on every request.

carlesarnal:
Hit this again on #9664 ([run 32129694482, job 95690319808](https://github.com/Apicurio/apicurio-registry/actions/runs/32129694482/job/95690319808)) — failed both surefire attempts, and the failure carries a diagnostic detail worth recording:

```
Content-Disposition=attachment; filename="testResponseNotCompressedWhenKillSwitchDisabled.json"
Content-Type=application/json
content-encoding=gzip
transfer-encoding=chunked
```

[...]

PR #9251, opened 2026-08-05
fix(rest): make compression kill switch dynamically configurable (#9245)

KartikSuryavanshi:
## What
Fixes the flaky `HttpCompressionDisabledTest.testResponseNotCompressedWhenKillSwitchDisabled`.

`apicurio.rest.compression.enabled` was a plain `boolean` injected into the `RestConfig` singleton **once at application startup**. Quarkus applies `@TestProfile` config overrides as system properties and **reuses** the running application instance across `@QuarkusTest` classes when a profile only changes runtime config (the restart decision is based on classloader identity, not the profile). As a result, the profile's `apicurio.rest.compression.enabled=false` was ignored on reused instances: the stale startup value (`true`) won and responses were still gzip-compressed, so the disabled test failed intermittently (CI run 30942716482, job 92115140512).

## Why
A runtime kill switch should be re-read from the live config, not captured at startup. Every other runtime-toggleable property in `RestConfig` (`groupDeletionEnabled`, `artifactDeletionEnabled`, `draftProductionModeEnabled`, ...) already uses the `@Dynamic Supplier<T>` pattern.

## Changes
- `RestConfig.java`: `compressionEnabled` is now `@Dynamic Supplier<Boolean>`; `isCompressionEnabled()` calls `.get()`, re-reading the current config value on every request.
- `application.properties`: added the required `apicurio.rest.compression.enabled.dynamic.allow` entry.
- `ref-registry-all-configs.adoc`: regenerated (type now `boolean [dynamic]` + new `.dynamic.allow` row).

## Verification
- Empirically confirmed the mechanism: with `@Dynamic`, `System.setProperty("apicurio.rest.compression.enabled","false")` after app start is reflected by `isCompressionEnabled()`; the plain `boolean` never could be.
- Full `noprofile.rest` shard: 339 tests, 0 failures.
- `HttpCompressionTest` (7) + `HttpCompressionDisabledTest` (1) pass.
- `checkstyle:check -pl app` passes.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/RestConfig.java
+++ b/app/src/main/java/io/apicurio/registry/rest/RestConfig.java
@@ -50,9 +50,10 @@ public class RestConfig {
     @Info(category = CATEGORY_REST, description = "When enabled, legacy HTTP error codes are used (e.g. 409 instead of 400 for rule violations)", availableSince = "3.3.0")
     boolean legacyErrorCodesEnabled;
 
+    @Dynamic(label = "Compression", description = "When selected, gzip compression is applied to REST API responses.")
     @ConfigProperty(name = "apicurio.rest.compression.enabled", defaultValue = "true")
     @Info(category = CATEGORY_REST, description = "Enable gzip compression of REST API responses", availableSince = "3.3.2")
-    boolean compressionEnabled;
+    Supplier<Boolean> compressionEnabled;
 
     public int getLabelsInSearchResultsMaxSize() {
         return this.labelsInSearchResultsMaxSize;
@@ -87,7 +88,7 @@ public class RestConfig {
     }
 
     public boolean isCompressionEnabled() {
-        return compressionEnabled;
+        return compressionEnabled.get();
     }
 
 }
--- a/app/src/main/java/io/apicurio/registry/rest/RestConfig.java
+++ b/app/src/main/java/io/apicurio/registry/rest/RestConfig.java
@@ -52,7 +52,7 @@ public class RestConfig {
 
     @Dynamic(label = "Compression", description = "When selected, gzip compression is applied to REST API responses.")
     @ConfigProperty(name = "apicurio.rest.compression.enabled", defaultValue = "true")
-    @Info(category = CATEGORY_REST, description = "Enable gzip compression of REST API responses", availableSince = "3.3.2")
+    @Info(category = CATEGORY_REST, description = "Enable gzip compression of REST API responses. This is a dynamic property and can be changed at runtime without restarting the server.", availableSince = "3.3.2")
     Supplier<Boolean> compressionEnabled;
 
     public int getLabelsInSearchResultsMaxSize() {
--- a/app/src/main/resources/application.properties
+++ b/app/src/main/resources/application.properties
@@ -202,6 +202,7 @@ apicurio.ccompat.legacy-id-mode.enabled.dynamic.allow=${apicurio.config.dynamic.
 apicurio.ccompat.use-canonical-hash.dynamic.allow=${apicurio.config.dynamic.allow-all}
 apicurio.ccompat.max-subjects.dynamic.allow=${apicurio.config.dynamic.allow-all}
 apicurio.download.href.ttl.seconds.dynamic.allow=${apicurio.config.dynamic.allow-all}
+apicurio.rest.compression.enabled.dynamic.allow=${apicurio.config.dynamic.allow-all}
 apicurio.rest.deletion.group.enabled.dynamic.allow=${apicurio.config.dynamic.allow-all}
 apicurio.rest.deletion.artifact.enabled.dynamic.allow=${apicurio.config.dynamic.allow-all}
 apicurio.rest.deletion.artifact-version.enabled.dynamic.allow=${apicurio.config.dynamic.allow-all}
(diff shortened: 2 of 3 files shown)


app/src/main/java/io/apicurio/registry/rest/RestConfig.java
```diff
@@ -50,9 +50,10 @@ public class RestConfig {
     @Info(category = CATEGORY_REST, description = "When enabled, legacy HTTP error codes are used (e.g. 409 instead of 400 for rule violations)", availableSince = "3.3.0")
     boolean legacyErrorCodesEnabled;
 
+    @Dynamic(label = "Compression", description = "When selected, gzip compression is applied to REST API responses.")
     @ConfigProperty(name = "apicurio.rest.compression.enabled", defaultValue = "true")
     @Info(category = CATEGORY_REST, description = "Enable gzip compression of REST API responses", availableSince = "3.3.2")
-    boolean compressionEnabled;
+    Supplier<Boolean> compressionEnabled;
```
paoloantinori, line 56: Worth flagging that this changes production behavior, not just the test: with `@Dynamic Supplier<Boolean>`, compression becomes runtime-toggleable without a restart and `isCompressionEnabled()` re-resolves config on every request. That looks intentional and consistent with the other `@Dynamic` toggles in `RestConfig`, so a one-line release-notes or config-docs entry would help operators see the new runtime-mutability. If the intent was strictly to stabilize the test, an alternative is a distinct profile that forces a Quarkus restart for `HttpCompressionDisabledTest`, which would avoid changing the production semantics.

paoloantinori (commented):
Clean, well-scoped fix: `apicurio.rest.compression.enabled` becomes a `@Dynamic Supplier<Boolean>` so the value is re-read on each request instead of captured once at startup. The change is internally consistent (annotation plus type swap plus `.get()` plus the new `apicurio.rest.compression.enabled.dynamic.allow` entry mirror the sibling dynamic fields), and the reported end-to-end results, `HttpCompressionDisabledTest` passing under the profile override and a 339-test green shard, address the main risk that the fix could be a no-op if compression were decided at startup.

- `RestConfig.java:56` (getter at `:91`): beyond stabilizing the flake, this also makes compression runtime-toggleable in production, which is broader than a test-only fix. The author's rationale, that every other runtime toggle in `RestConfig` already uses this pattern and a kill switch should re-read live config, is sound; just confirm that is the intended scope and capture it in release notes or config docs so operators notice the new runtime-mutability.
- Fix efficacy hinges on the compression request path consulting `isCompressionEnabled()` per request rather than caching it at startup. The disabled test asserts the response is not gzip-compressed and now passes with the override set, which is direct evidence the output path honors the live value, so that risk is covered.
- `ref-registry-all-configs.adoc`: the `boolean` to `boolean [dynamic]` change and the new `.dynamic.allow` row match the sibling rows exactly. The `unknown` type on the `.dynamic.allow` row is a pre-existing docs-generator quirk shared by every such row, not something introduced here.

KartikSuryavanshi:
Thanks @paoloantinori — good catch on the scope.

To confirm: this is intentional and deliberately broader than just stabilizing the flaky test. `apicurio.rest.compression.enabled` is the kill switch for compression, so it should behave like the other runtime toggles in `RestConfig` and re-read the live config on each request. That's also why I steered away from the profile-restart alternative: it would fix the flake but leave a property that silently ignores runtime changes, which feels worse than the production mutability it avoids. If compression ever acts up mid-incident, toggling it off without a restart is a genuinely useful knob.

On the docs: this repo doesn't keep a CHANGELOG/release-notes file, so the generated config reference is where operators will look. I updated the `@Info` description and the matching `ref-registry-all-configs.adoc` row to spell it out:

> Enable gzip compression of REST API responses. This is a dynamic property and can be changed at runtime without restarting the server.

That, together with the `boolean [dynamic]` type marker, should make the runtime-mutability obvious. Happy to add a line to the upgrade notes as well if you'd prefer it surfaced there too.

paoloantinori:
Thanks @KartikSuryavanshi — agreed, the broader scope is the right call. A kill switch that silently ignores runtime changes is worse than one that honors them, and matching the `@Dynamic Supplier<T>` pattern used by the other runtime toggles in `RestConfig` keeps the config layer consistent. The fix looks good.

KartikSuryavanshi:
Thanks, @paoloantinori! I really appreciate the review and feedback.

Since everything looks good now, would it be possible to get this merged when you have a moment? Thanks!

paoloantinori:
Thanks @KartikSuryavanshi. The fix is reviewed and looks correct; I do not have further changes to request.

On merging: that call is @jsenko's to make (he is the maintainer who runs `/merge` here), and I cannot push the merge button myself. He has already indicated he will merge once the full test suite passes, so once CI is green this should land.

KartikSuryavanshi:
Thanks for the clarification, @paoloantinori!

I'm not sure why the **Verify / Verification Gate** check is failing. I notice it seems to be failing on most PRs, so I wasn't sure if it's an infrastructure issue or something specific to my PR. Do you know if there's anything I need to do on my end, or should I just wait for it to be resolved?

paoloantinori:
@KartikSuryavanshi — the Verification Gate failure here is an infrastructure issue, not something specific to your PR.

On this head the gate is red but `Unit Tests`, `Integration Tests`, `Build`, `SonarCloud`, and `Extra Tests` are all either green or skipped. There is no failing test job tied to the code in this PR (the `RestConfig.java` / `application.properties` change). Your scope is small and the relevant quality gates that did run (`SonarCloud`, `SonarCloud Code Analysis`) passed.

Separately, the "Verify" workflow is currently red on `main` itself (last run failed), so the gate is not passing anywhere right now. Nothing for you to action on the PR — it's waiting on the infra issue to clear. Once the gate is healthy on `main`, a re-run on this head should go green.

PR #9252, opened 2026-08-05
fix(ui): display enum allowed values in Prompt Template variables table

### LINKED ISSUE #9103: fix(ui): PromptTemplateViewer hides enum allowed values in Variables documentation table
When a `PROMPT_TEMPLATE` artifact contains a variable with an `enum` restriction, the `PromptTemplateViewer` does not display the allowed values in the Variables documentation table.

The Playground/Test Panel already reads the `enum` values and renders them as selectable options, but the Documentation view omits the same information. This makes it difficult for users reading the template documentation to determine which values are valid for an enum-constrained variable.

### Environment

* **Registry version:** main (3.3.2-SNAPSHOT)
* **Persistence:** in-memory
* **Operating System:** Windows
* **Browser:** Google Chrome
* **Backend:** `..\mvnw.cmd quarkus:dev`
* **Frontend:** `npm run dev`

### Steps to Reproduce

1. Open the Registry UI.
2. Click **Create Artifact**.
3. Select **Prompt Template** as the artifact type.
4. Paste the following YAML into the content editor:

```yaml
name: "Enum Test Prompt"
template: "Summarize this in {{tone}}"
variables:
  tone:
    type: "string"
    description: "The desired tone of the output"
    enum:
      - "professional"
      - "casual"
      - "pirate"
```

5. Create the artifact.
6. Open the **Version 1** page.
7. Navigate to the **Documentation** tab.
8. Check the **Variables** table.
9. Compare it with the **Playground/Test Panel** below the table.

### Expected Behavior

The Variables documentation table should expose the allowed values for enum-constrained variables.

For example, the table could include an **Allowed Values** column between **Default** and **Description**, displaying:

```text
professional   casual   pirate
```

using the existing PatternFly components.

### Actual Behavior

The Variables table currently contains the following columns:

```text
Name | Type | Required | Default | Description
```

The `enum` values are not displayed anywhere in the documentation table.

At the same time, the Playground/Test Panel correctly detects the enum and renders a dropdown containing:

```text
-- Select --
professional
casual
pirate
```

Selecting an option from the dropdown works correctly.

This indicates that the issue is isolated to the documentation table rendering.

### API / Network Verification

Using the Chrome DevTools Network tab, the `/versions/1` API response contains the enum values:

```json
"variables": {
  "tone": {
    "type": "string",
    "description": "The desired tone of the output",
    "enum": ["professional", "casual", "pirate"]
  }
}
```

Therefore, the enum information is already available in the API response.

### Frontend Investigation

The affected component is:

```text
ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
```

The `PromptVariable` TypeScript interface already defines:

```typescript
enum?: string[];
```

However, when `PromptTemplateViewer` builds the Variables table rows, the rendering logic does not reference `variable.enum`.

A repository search also shows that `variable.enum` is already used in `PromptTemplateTestPanel.tsx` to create the selection control, but it is not rendered by `PromptTemplateViewer`.

### Proposed Fix

Add an **Allowed Values** column to the Variables table and render the enum values using PatternFly `LabelGroup` and `Label` components.

The values would be displayed only when an enum is present; otherwise the cell can display `-`.

### Screenshot

The attached screenshot shows the current Variables documentation table where the enum values are not displayed, while the same Prompt Template uses the enum correctly in the Playground/Test Panel.

Image

nishant-mazumder:
Fixes #9103.

The backend already exposes enum values for Prompt Template variables, and the Playground/Test Panel correctly uses them to render a dropdown for enum-constrained variables. However, the Documentation view did not surface this information, making it difficult for users to discover the valid values for a variable directly from the generated documentation.

This change updates the Prompt Template documentation table by:

- adding an **Allowed Values** column to the Variables table
- rendering enum values using existing PatternFly `LabelGroup` and `Label` components
- displaying `-` when a variable has no enum restriction
- keeping the documentation view consistent with the Playground/Test Panel

## Before

The Variables table displayed:

- Name
- Type
- Required
- Default
- Description

Although the backend returned the enum values and the Playground rendered them correctly, the Documentation view omitted them completely.

**Screenshot:**

Screenshot 2026-07-31 205220

---

## After

The Variables table now includes an **Allowed Values** column.

For enum-constrained variables, the allowed values are displayed as PatternFly labels, making the documentation consistent with the Playground while improving discoverability.

**Screenshot:**

Screenshot 2026-07-31 212956

---

## Testing

- Reproduced the issue locally.
- Verified that enum values are present in the backend API response.
- Verified that the Documentation view now displays enum values correctly.
- Verified that variables without enum values continue to display `-`.
- Ran `npm run lint` successfully.
- Ran `npm test` successfully.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
@@ -170,6 +170,7 @@ export const PromptTemplateViewer: FunctionComponent<PromptTemplateViewerProps> 
                                         <th>Type</th>
                                         <th>Required</th>
                                         <th>Default</th>
+                                        <th>Allowed Values</th>
                                         <th>Description</th>
                                     </tr>
                                 </thead>
@@ -190,6 +191,15 @@ export const PromptTemplateViewer: FunctionComponent<PromptTemplateViewerProps> 
                                             <td>{variable.default !== undefined ? (
                                                 <code>{String(variable.default)}</code>
                                             ) : "-"}</td>
+                                            <td>
+                                                {variable.enum && variable.enum.length > 0 ? (
+                                                    <LabelGroup>
+                                                        {variable.enum.map((val, i) => (
+                                                            <Label key={i} color="grey" isCompact>{String(val)}</Label>
+                                                        ))}
+                                                    </LabelGroup>
+                                                ) : "-"}
+                                            </td>
                                             <td>{variable.description || "-"}</td>
                                         </tr>
                                     ))}
--- a/ui/tests/specs/promptTemplate.spec.ts
+++ b/ui/tests/specs/promptTemplate.spec.ts
@@ -1,0 +1,129 @@
+import { test, expect } from "@playwright/test";
+
+const REGISTRY_UI_URL: string = process.env["REGISTRY_UI_URL"] || "http://localhost:8888";
+
+const MOCK_PROMPT_TEMPLATE = {
+    templateId: "test-template",
+    name: "Test Template",
+    template: "Hello {{name}}, choose your {{color}}",
+    variables: {
+        color: {
+            name: "color",
+            type: "string",
+            description: "A color",
+            enum: ["red", "green", "blue"]
+        },
+        name: {
+            name: "name",
+            type: "string",
+            description: "Your name"
+        },
+        emptyEnum: {
+            name: "emptyEnum",
+            type: "string",
+            enum: []
+        }
+    }
+};
+
+const MOCK_VERSION_METADATA = {
+    groupId: "default",
+    artifactId: "test-prompt-template",
+    version: "1",
+    type: "PROMPT_TEMPLATE",
+    artifactType: "PROMPT_TEMPLATE",
+    state: "ENABLED",
+    name: "Test Prompt Template",
+    description: "A test prompt template",
+    createdOn: "2026-08-05T00:00:00Z",
+    createdBy: "test-user",
+    labels: {}
+};
+
+test("Prompt Template - renders enum allowed values correctly", async ({ page }) => {
+    // Mock system config, info, and user endpoints so the UI loads without a backend
+    await page.route("**/apis/registry/v3/system/info", async route => {
+        await route.fulfill({ status: 200, contentType: "application/json", body: JSON.stringify({ name: "Apicurio Registry", version: "3.3.2.Final" }) });
+    });
+    await page.route("**/apis/registry/v3/users/me", async route => {
+        await route.fulfill({ status: 200, contentType: "application/json", body: JSON.stringify({ username: "test-user", displayName: "Test User", admin: true, developer: true, viewer: true }) });
+    });
+    await page.route("**/apis/registry/v3/config", async route => {
+        await route.fulfill({ status: 200, contentType: "application/json", body: JSON.stringify({ features: { readOnly: false } }) });
+    });
+
+    // Mock the artifact metadata response
+    await page.route("**/apis/registry/v3/groups/default/artifacts/test-prompt-template", async route => {
+        await route.fulfill({
+            status: 200,
+            contentType: "application/json",
+            body: JSON.stringify(MOCK_VERSION_METADATA)
+        });
+    });
+
+    // Mock the version metadata response so the UI knows it's a PROMPT_TEMPLATE
+    await page.route("**/apis/registry/v3/groups/default/artifacts/test-prompt-template/versions/1", async route => {
+        await route.fulfill({
+            status: 200,
+            contentType: "application/json",
+            body: JSON.stringify(MOCK_VERSION_METADATA)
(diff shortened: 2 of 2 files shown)


ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
```diff
@@ -190,6 +191,15 @@ export const PromptTemplateViewer: FunctionComponent<PromptTemplateViewerProps>
                                             <td>{variable.default !== undefined ? (
                                                 <code>{String(variable.default)}</code>
                                             ) : "-"}</td>
+                                            <td>
+                                                {variable.enum && variable.enum.length > 0 ? (
+                                                    <LabelGroup>
+                                                        {variable.enum.map((val, i) => (
+                                                            <Label key={i} color="grey" isCompact>{String(val)}</Label>
```
paoloantinori, line 198: This PR adds no test covering the new conditional. A render test asserting the labels appear for an enum-constrained variable, and that `-` shows otherwise, would lock in the behavior and guard against drift, since the same enum data is rendered differently (as a dropdown) in the Playground/Test Panel.

paoloantinori (commented):
Clean, well-scoped change that surfaces enum allowed-values in the Prompt Template documentation table and sensibly handles variables without an enum. No correctness, security, or blocker-level issues; the items below are easy wins rather than must-fixes.

- `ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx` (new `<td>`, L194): the `variable.enum && variable.enum.length > 0 ? ... : "-"` guard correctly falls back to `-` for variables with no enum and for empty arrays, matching the `-` convention already used for Default and Description.
- `PromptTemplateViewer.tsx` (L198): this PR adds no test for the new conditional render. A render test asserting the labels appear for an enum-constrained variable and that `-` shows otherwise would prevent regressions, particularly since the enum data is rendered as a dropdown in the Playground/Test Panel and the two views can drift.
- `PromptTemplateViewer.tsx` (L198): `String(val)` coerces values, so an empty-string enum member would render as a blank label. Fine for the typical string enums the backend returns; worth handling only if enum members can be non-string.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
nishant-mazumder: Thanks for the review, @paoloantinori! 

I've pushed an update that adds render tests for this behavior using @testing-library/react. The tests verify that enum-constrained variables render their allowed values as labels, and that variables without enum values correctly display -.

Since the UI package didn't previously have the required React DOM testing setup, I also added the necessary testing dependencies so these component render tests can run in CI.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+        const nameAllowedValuesCell = nameRow?.querySelectorAll("td")[4];
+        expect(nameAllowedValuesCell?.textContent).toBe("-");
+
+        // Check the emptyEnum variable (empty array enum)
+        const emptyEnumRow = Array.from(rows).find(row => row.textContent?.includes("emptyEnum"));
+        expect(emptyEnumRow).toBeDefined();
+        const emptyAllowedValuesCell = emptyEnumRow?.querySelectorAll("td")[4];
+        expect(emptyAllowedValuesCell?.textContent).toBe("-");
+    });
+});
```
Nachiket-Roy: This is not the correct testing method please follow the repo conventions correctly i.e. implementing the verification as a playwright E2E test in `ui/tests/specs/`

ui/ui-app/package-lock.json
```diff
@@ -43,6 +43,7 @@
         "@apicurio/eslint-config": "1.0.0",
```
Nachiket-Roy, line 43: Hence, no dependency changes are required.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.test.tsx
nishant-mazumder: I've removed this React component test and replaced it with a proper Playwright E2E test following the repository conventions. It successfully mocks the APIs and verifies the UI renders the enum values correctly.

paoloantinori:
Thanks for the rework. Swapping the React component test for a Playwright E2E test that mocks the registry endpoints and asserts the rendered enum labels is the right layer to exercise this — it matches the repo's existing E2E conventions and actually verifies the UI surfaces the allowed values. The mock coverage (system info, users/me, config, artifact/version metadata, content, rules, branches) looks complete for the documentation tab. Looks good to me.

EricWittmann:
Thanks for the contribution, @nishant-mazumder! Nice clean change, and great job switching to the Playwright E2E test approach when it was pointed out. Merged!

PR #9255, opened 2026-08-05
fix(contracts): emit ruleset events on kafkasql and fix data export

### LINKED ISSUE #9246: test: KafkaSqlEventsTest ContractRuleset event tests are flaky (20s timeout)
## Flaky tests

Six tests in `io.apicurio.registry.event.kafkasql.KafkaSqlEventsTest` consistently time out:

- `versionContractRulesetDeleted`
- `artifactContractRulesetDeleted`
- `globalContractRulesetDeleted`
- `globalContractRulesetConfigured`
- `versionContractRulesetConfigured`
- `contractRulesetConfigured`

## Failure

All six error with a ~20 second timeout, suggesting the expected Kafka events are never delivered:

```
io.apicurio.registry.event.kafkasql.KafkaSqlEventsTest.versionContractRulesetDeleted
  -- Time elapsed: 20.32 s <<< ERROR!
io.apicurio.registry.event.kafkasql.KafkaSqlEventsTest.artifactContractRulesetDeleted
  -- Time elapsed: 20.21 s <<< ERROR!
io.apicurio.registry.event.kafkasql.KafkaSqlEventsTest.globalContractRulesetDeleted
  -- Time elapsed: 20.13 s <<< ERROR!
...
```

The test class ran twice (Surefire retry) and all 6 failed both times.

## Observed in

- PR #9195 — `Unit Tests / Unit Tests (app-kafkasql)` shard
- [CI run 30942716482, job 92115140515](https://github.com/Apicurio/apicurio-registry/actions/runs/30942716482/job/92115140515)

The failure is unrelated to the PR's changes (label-parsing fix in `SearchResourceImpl`).

## Likely cause

The ContractRuleset event types may not be fully wired into the KafkaSQL event pipeline, or the
event consumer is not subscribing to the correct topic/partition for these newer event types. Note
that the `DataExporter` also logs `Unhandled entity type: ContractRule` errors in the same CI run,
which may be related.

malladinagarjuna2:
## Summary

Contract ruleset changes emitted no events under `kafkasql` storage, and
contract rules were silently dropped from data exports.

Fixes #9246

## Root Cause

`KafkaSqlRegistryStorage` fires a `KafkaSqlOutboxEvent` after every mutation
except the six contract-ruleset ones, which just submitted the journal message
and returned. Those events existed only on the SQL path, which writes to a DB
outbox table nothing reads under kafkasql — so nothing reached the
`registry-events` topic and the tests timed out at 20s. Deterministic, not
flaky: the `app-sql` shard passes the identical inherited tests in the same run.

Separately, `EntityWriter`/`EntityReader` had no `ContractRule` case, so exports
threw `Unhandled entity type: ContractRule`. `DataExporter` swallows that, so
the export returned 200 with a zip missing every contract rule.

## Changes

- `KafkaSqlRegistryStorage`: fire `ContractRulesetConfigured` in the six
  contract-ruleset methods, after `coordinator.waitForResponse(uuid)`
- `EntityWriter` / `EntityReader`: handle the `ContractRule` entity type;
  indexed after versions due to the FK on `versions(globalId)`
- New `ContractRuleImportExportTest` covering artifact, version and global scope

## Test plan

- [x] `KafkaSqlEventsTest` — 30/30 pass, including the six from #9246
- [x] `ContractRuleImportExportTest` passes; fails with the export fix stashed
- [x] `checkstyle:check -pl app,utils/importexport` clean
- [ ] Full CI (four unrelated shards were already red on the referenced run)

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
@@ -10,6 +10,7 @@ import io.apicurio.registry.events.ArtifactVersionCreated;
 import io.apicurio.registry.events.ArtifactVersionDeleted;
 import io.apicurio.registry.events.ArtifactVersionMetadataUpdated;
 import io.apicurio.registry.events.ArtifactVersionStateChanged;
+import io.apicurio.registry.events.ContractRulesetConfigured;
 import io.apicurio.registry.events.GlobalRuleConfigured;
 import io.apicurio.registry.events.GroupCreated;
 import io.apicurio.registry.events.GroupDeleted;
@@ -99,6 +100,9 @@ import static io.apicurio.registry.utils.ConcurrentUtil.blockOnResult;
 @Logged
 @LookupIfProperty(name = "apicurio.storage.kind", stringValue = "kafkasql")
 public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implements RegistryStorage {
+
+    // Coordinates used by contract ruleset events to identify the global (non artifact scoped) ruleset.
+    private static final String GLOBAL_CONTRACT_COORDINATE = "__GLOBAL__";
 
     @Inject
     Logger log;
@@ -691,6 +695,8 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
         var message = new SetArtifactContractRuleset3Message(groupId, artifactId, ruleset);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent
+                .of(ContractRulesetConfigured.of(groupId, artifactId, null, "SET")));
     }
 
     @Override
@@ -699,6 +705,8 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
         var message = new DeleteArtifactContractRuleset2Message(groupId, artifactId);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent
+                .of(ContractRulesetConfigured.of(groupId, artifactId, null, "DELETE")));
     }
 
     @Override
@@ -713,6 +721,8 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
         var message = new SetVersionContractRuleset4Message(groupId, artifactId, version, ruleset);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent
+                .of(ContractRulesetConfigured.of(groupId, artifactId, version, "SET")));
     }
 
     @Override
@@ -721,6 +731,8 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
         var message = new DeleteVersionContractRuleset3Message(groupId, artifactId, version);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent
+                .of(ContractRulesetConfigured.of(groupId, artifactId, version, "DELETE")));
     }
 
     @Override
@@ -741,6 +753,8 @@ public class KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage implement
         var message = new SetGlobalContractRuleset1Message(ruleset);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent.of(ContractRulesetConfigured
+                .of(GLOBAL_CONTRACT_COORDINATE, GLOBAL_CONTRACT_COORDINATE, null, "SET")));
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


malladinagarjuna2:
hey thanks for opening the pr will work on it

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -741,13 +753,17 @@ public void setGlobalContractRuleset(io.apicurio.registry.storage.dto.ContractRu
         var message = new SetGlobalContractRuleset1Message(ruleset);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent.of(ContractRulesetConfigured
+                .of(GLOBAL_CONTRACT_COORDINATE, GLOBAL_CONTRACT_COORDINATE, null, "SET")));
     }
```
paoloantinori, line 755: The global ruleset is signaled by passing the magic string `__GLOBAL__` as both groupId and artifactId. Two things to confirm. First, the constant is private to this class, so any `ContractRulesetConfigured` consumer that detects global scope must know this string independently; one that keys off null or empty silently misclassifies every global ruleset change. Second, the SQL storage path already emits this event, so what does it pass for the global case? If it passes null or anything else, app-sql and kafkasql deployments now emit different payloads for the same logical event. Recommend matching the SQL path exactly, or hoisting the sentinel to a shared constant the consumer also references, and asserting the global event payload in `KafkaSqlEventsTest` rather than just that an event fired.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -691,6 +695,8 @@ public void setArtifactContractRuleset(String groupId, String artifactId,
         var message = new SetArtifactContractRuleset3Message(groupId, artifactId, ruleset);
         var uuid = blockOnResult(submitter.submitMessage(message));
         coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent
+                .of(ContractRulesetConfigured.of(groupId, artifactId, null, "SET")));
     }
```
paoloantinori, line 697: The action is passed as the raw strings `SET` and `DELETE`. If the event consumer switches on these values, a typo or vocabulary mismatch silently drops the event with no error on either side. Confirm the consumer handles exactly these two values, and use a shared enum or constant instead of inline literals if one exists.

paoloantinori (commented):
Both fixes are correct and well-scoped, and the round-trip test plus clean checkstyle give good confidence. The one item to settle before merge is the representation of the global ruleset event: the new `__GLOBAL__` sentinel couples this producer to event consumers through an undocumented magic string and may diverge from what the SQL path already emits.

- `KafkaSqlRegistryStorage.java:105,758` (MAJOR): the global ruleset is signaled by passing `__GLOBAL__` as both groupId and artifactId. The constant is private to this class, so any consumer of `ContractRulesetConfigured` that detects global scope has to know this string independently; one that keys off null or empty silently misclassifies every global change. The SQL path already emits this event per the root-cause note, so confirm what it passes for the global case and match it, otherwise app-sql and kafkasql emit different payloads for the same logical event.
- `KafkaSqlRegistryStorage.java:700` (MAJOR): the action is the raw strings `SET`/`DELETE`. If the consumer switches on these values, a mismatch silently drops the event with no error on either side. Confirm the consumer handles exactly these values and prefer a shared enum or constant if one exists.
- `EntityWriter.java:135` (MAJOR, verify): the filename is `globalId-ruleCategory-orderIndex`, and a collision silently overwrites an earlier zip entry. The comment claims that tuple is unique per artifact, but the writer cannot detect a duplicate, so confirm it against the schema. Also note a global rule serializes under `artifacts/null/...` because `artifactId` is null; import still works (content-driven), but verify that path is intended.
- `ContractRuleImportExportTest.java` (MAJOR, coverage): the round trip is solid, but the event side lives in `KafkaSqlEventsTest`, which is not in this diff. Make sure it asserts the global event's scope and payload, not just that an event fired, or the `__GLOBAL__` representation could be wrong and still go green.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
malladinagarjuna2: yeah will address  them now!

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
malladinagarjuna2: will be more carefull about these silent failing cases from next time

malladinagarjuna2:
hey @EricWittmann  the unit tests  related to the  fix in contracts are passing 
the ci unit tests are not passing  because of the java heap problem which is mentioned in the #9265, if you want i can implement the fix here only. 
for 9265 we don't really have the time for opening a seperate pr for 9265,as ci is an important issue. its my strong suggestion to implement the fix for 9265  here and implement it here, rest i am waiting for your suggestions on this.

if you want then i am ready to implement the other pr as well

malladinagarjuna2:
@paoloantinori  what do you think about the above one ?

paoloantinori:
Re-read at the new head (`42e4993e`). Both points from my earlier inline comments are addressed cleanly:

- The global-coordinate magic string is now the public `ContractRulesetConfigured.GLOBAL_COORDINATE` constant with a dedicated `ofGlobal(Action)` factory, so consumers no longer have to re-derive the sentinel.
- `action` is now `Action.SET`/`Action.DELETE` via the typed enum, removing the raw-string mismatch risk.

On the CI question: the `app-other` shard failure is the pre-existing Java heap OOM tracked in #9265 — it hit #9251 too, with no test assertion failing, just the forked JVM crashing under runner memory pressure. It's not a regression from this PR's changes. The contract-ruleset tests passing locally plus the `app-sql` shard passing the identical inherited tests in the same run are the right signals here. Worth confirming the `app-other` shard is green once #9265's heap mitigation lands, but nothing in this PR should block on it.

malladinagarjuna2:
thanks for the review  @paoloantinori

PR #9256, opened 2026-08-05
fix(cli): resolve max column width from terminal COLUMNS env var

nidhiii128:
Related to #9479 - that issue was opened because the old COLUMNS-only approach didn't really work. This PR now detects the real terminal width directly (via stty size), so the problem #9479 describes is already fixed here.

Previously MAX_COLUMN_WIDTH was hardcoded to 25, per a TODO in TableBuilder.

This now resolves the max column width in order of preference:

An explicitly exported COLUMNS environment variable, if set to a valid number.
The real terminal width, detected by running stty size against the controlling terminal (only when output is attached to an interactive terminal — this is skipped entirely for piped/redirected output, CI logs, etc.).
The previous default of 25.

The result is always capped at 80 to avoid excessively wide cells on very large terminals. Resolution is lazy and cached — it only runs (and only forks stty if needed) the first time a value actually needs wrapping, not at class load.


cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -36,6 +38,29 @@ public class TableBuilder {
...
+    /**
+     * Resolves the max column width from the terminal width, when available, falling back to a
+     * sensible default. Reads the COLUMNS environment variable, which most POSIX shells export for
+     * interactive terminals. When output is piped, redirected, or COLUMNS is unset/invalid (common in
+     * CI), falls back to {@link #DEFAULT_MAX_COLUMN_WIDTH}. The result is capped at
+     * {@link #UPPER_MAX_COLUMN_WIDTH} so a single wrapped cell doesn't become unreasonably wide on
+     * very large terminals.
+     */
+    private static int resolveMaxColumnWidth() {
+        var columnsEnv = System.getenv("COLUMNS");
```
paoloantinori, line 72: `COLUMNS` is not in the JVM's environment for typical interactive terminals: bash and zsh keep `COLUMNS` as a non-exported shell variable, so `System.getenv("COLUMNS")` returns null unless the user explicitly runs `export COLUMNS`. I verified this empirically: a shell-set `COLUMNS` is invisible to a child process's getenv, while an exported one is visible. As a result the resolver silently falls back to the 25 default in the common case, so the PR's stated goal of reading terminal width is mostly unmet, and the Javadoc at line 43 ("most POSIX shells export [COLUMNS] for interactive terminals") is inaccurate. At minimum, correct that sentence; to actually honor the goal, consider detecting the real TTY size (e.g., `stty size` or JLine) when stdout is a terminal, or document that this only takes effect when `COLUMNS` is explicitly exported (some CI configs).

cli/src/test/java/io/apicurio/registry/cli/utils/TableBuilderTest.java
```diff
@@ -22,6 +22,16 @@ private static String[] row(String output, int lineIndex) {
...
     }
 
+    @Test
+    void resolveMaxColumnWidthFallsBackToDefaultWhenColumnsEnvIsAbsentOrInvalid() {
+        var longValue = "x".repeat(200);
+        var output = render(new TableBuilder()
+                .addColumns("Description")
+                .addRow(longValue));
+        var lines = output.split("\n");
+        assertThat(lines.length).isGreaterThan(3);
```
paoloantinori, line 77: This assertion does not validate the new behavior. A 200-character value wraps to well over 3 lines under the old hardcoded MAX_COLUMN_WIDTH=25, so the test passes even with this entire change reverted. It also never sets `COLUMNS`, so it cannot exercise the env-var resolution, the 80 cap, or the invalid or below-MIN cases the test name promises. Because `MAX_COLUMN_WIDTH` is a static final resolved at class load from the real environment, the env path is not unit-testable as written. Suggest asserting on the actual rendered width (e.g., every wrapped line's length is bounded by the resolved MAX_COLUMN_WIDTH, or an exact expected line count for the input at width 25), and extracting the resolution into a package-private method that takes the env value as a parameter so the present, invalid, and below-minimum branches can be tested directly.

paoloantinori (commented):
The change cleanly replaces the hardcoded TODO constant with safe env parsing and an upper cap, but its stated goal is largely unmet in practice and the new test does not actually validate the new logic.

- `cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java:50`: `COLUMNS` is a non-exported shell variable in interactive bash/zsh, so `System.getenv("COLUMNS")` is null in the common terminal case and the resolver silently falls back to the 25 default; the Javadoc at line 43 ("most POSIX shells export [COLUMNS] for interactive terminals") is inaccurate. Verified empirically.
- `cli/src/test/java/io/apicurio/registry/cli/utils/TableBuilderTest.java:32`: the `assertThat(lines.length).isGreaterThan(3)` assertion holds under the old MAX_COLUMN_WIDTH=25 as well, so the test passes with the entire change reverted and never drives the env-var, cap, or invalid-value paths the test name promises.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
nidhiii128: oh you're right COLUMNS isn't exported by default in bash/zsh so this basically never kicks in for a normal terminal. fixed the comment to say it only works when COLUMNS is explicitly exported instead of implying it just works. pushed.

cli/src/test/java/io/apicurio/registry/cli/utils/TableBuilderTest.java
nidhiii128: the test wasn't actually checking anything new. pulled the width logic out into its own method that takes the env value directly so I can test it without messing with real env vars. added cases for null, garbage input, too-low, normal, whitespace, exactly-at-minimum, and too-high. pushed.

paoloantinori:
Nice, extracting `resolveMaxColumnWidth(String)` as a package-private static method is exactly the right move — it makes the width logic unit-testable without touching the real environment. The seven cases you added (null, non-numeric, below-min, within-bounds, whitespace, at-minimum, capped) cover the branches cleanly. The wrapping test on top is a good integration-level check that the resolved width actually flows through to the rendered output. No further concerns from me on this.

nidhiii128:
@paoloantinori thanks for checking it over! let me know if there's anything else you'd like changed.

paoloantinori:
Thanks for the follow-up — both points are addressed. The Javadoc now accurately describes the non-exported behavior of `COLUMNS`, and the extracted `resolveMaxColumnWidth(String)` with the seven cases (null, non-numeric, below-min, in-range, whitespace, at-min, over-cap) properly exercises every branch. No further changes from me.

vandanayadav7 (commented):
Ran e2e testing with various `COLUMNS` values (unset, 10, 50, 120, 500) — the resolution logic works correctly and the table wrapping adjusts as expected. Tests pass (21/21).

One concern: `COLUMNS` is a shell-local variable that is not exported to child processes by default in bash and zsh. In practice `System.getenv("COLUMNS")` returns null, so the fallback to 25 always kicks in. The feature only activates if the user explicitly runs `export COLUMNS=<value>`, which is uncommon. The TODO in the code was asking for terminal width detection — this doesn't quite solve that since `COLUMNS` is rarely available and its value represents terminal width, not per-column width.

Worth considering whether a proper terminal width detection approach (e.g., via picocli or JNA) would be more effective, or if this is good enough as an incremental step.

Also:
- No linked issue referenced in the PR
- Missing trailing newline on both changed files (`\ No newline at end of file`)

nidhiii128:
Hii @vandanayadav7 Trailing newlines added on both files. the linked issue — this started from the TODO in TableBuilder.java rather than a filed issue, so there wasn't one to link. Happy to open one now if you'd prefer it tracked separately.

vandanayadav7 (changes requested):
Thanks for the work on this! The code quality is good, but the core concern is that `COLUMNS` isn't exported to child processes by default in bash/zsh, so `System.getenv("COLUMNS")` returns null in normal usage and the fallback to 25 always kicks in. The feature doesn't activate in practice.

I've opened #9479 for proper terminal width detection using JNA or picocli, which would actually solve the TODO. Would you like to update this PR to implement that approach instead? Happy to help if you have questions.

nidhiii128:
Hi @vandanayadav7 , thanks for the pointer! I've pushed a fix that detects the real terminal width directly, instead of relying on COLUMNS being exported.
1) It now runs stty size to ask the terminal how wide it actually is, so no need for the user to manually export COLUMNS anymore. If that's not possible for some reason (like output being piped somewhere, or on Windows where stty doesn't exist), it just falls back to the old default of 25, so nothing breaks.
2) I didn't use JNA or picocli like you suggested picocli's own width-detection code turned out to be private (can't be called from outside), and JNA adds some extra complexity for the native builds this CLI ships, since it needs native bindings. stty size gets the same result with a plain Java ProcessBuilder call, no extra dependencies needed.
3) Also fixed a SonarCloud flag it caught, I was calling stty by name, which relies on the system's PATH to find it, so I switched to the full path (/bin/stty) instead, which is a bit safer.
Let me know if you'd still prefer the JNA route for consistency, happy to switch if so!

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -21,7 +24,9 @@
 public class TableBuilder {
 
     private static final int MIN_COLUMN_WIDTH = 3;
-    private static final int MAX_COLUMN_WIDTH = 25; // TODO: Dynamically based on terminal width.
+    private static final int DEFAULT_MAX_COLUMN_WIDTH = 25;
+    private static final int UPPER_MAX_COLUMN_WIDTH = 80;
+    private static final int MAX_COLUMN_WIDTH = resolveMaxColumnWidth();
```
paoloantinori: This resolves the width in a static initializer, so merely loading `TableBuilder` forks a `stty` subprocess and can block up to 500ms. The PR title and description cover only the COLUMNS env var plus a default; the subprocess detection is not described. If intended, consider resolving lazily on first render instead of at class load, and mention the stty path in the description.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -36,6 +41,84 @@ public class TableBuilder {
...
+                return clampColumnWidth(Integer.parseInt(columnsEnv.trim()));
+            } catch (NumberFormatException ignored) {
+                // Fall through to default.
+            }
+        }
+        return DEFAULT_MAX_COLUMN_WIDTH;
+    }
+
+    private static int clampColumnWidth(int candidate) {
+        return candidate >= MIN_COLUMN_WIDTH ? min(candidate, UPPER_MAX_COLUMN_WIDTH) : DEFAULT_MAX_COLUMN_WIDTH;
```
paoloantinori, line 99: `min(...)` is unqualified and the added import block (lines 9-11) does not include `import static java.lang.Math.min`. Unless the class already has that import unchanged, this will not compile. Add the import or use `Math.min(...)` here.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -36,6 +41,84 @@ public class TableBuilder {
...
+     * {@link #UPPER_MAX_COLUMN_WIDTH} so a single wrapped cell doesn't become unreasonably wide on
+     * very large terminals.
+     * <p>
+     * {@code COLUMNS} is checked first because it's an explicit, cheap override some CI/automation
+     * setups already export - see {@link #resolveMaxColumnWidth(String)}. Most interactive shells
+     * (bash, zsh) do <b>not</b> export it by default, so in practice this falls through to live
+     * detection via {@link #detectTerminalWidth()}.
+     */
+    private static int resolveMaxColumnWidth() {
+        var columnsEnv = System.getenv("COLUMNS");
```
paoloantinori, line 72: A blank `COLUMNS` (e.g. an empty value some shell init files export) is non-null, so this branch is taken, parsing fails, and the code returns 25 without ever calling `detectTerminalWidth()`. Treat blank the same as unset so a stray empty value does not silently disable terminal detection.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -36,6 +41,84 @@ public class TableBuilder {
...
+     * as a GraalVM native image, and {@link ProcessBuilder} works there without any extra
+     * reflection/resource-config, whereas JNI-based libraries typically need additional
+     * native-image setup and can be fragile across platforms.
+     */
+    static Integer detectTerminalWidth() {
+        if (System.console() == null) {
+            return null;
+        }
+        try {
+            var process = new ProcessBuilder("/bin/stty", "size")
```
paoloantinori: Hardcoding `/bin/stty` skips systems where `stty` is only in `/usr/bin` (minimal containers, some distros without the `/bin` to `/usr/bin` symlink). Use `"stty"` and let PATH resolve it; the existing null-on-failure path already covers a missing binary.

paoloantinori (commented):
Two issues stand out before merge: `clampColumnWidth` calls an unqualified `min(...)` with no visible `import static java.lang.Math.min` in the diff (likely a compile failure), and the change forks a `/bin/stty` subprocess at class initialization, which the PR description (COLUMNS env var plus a 25 default) does not mention. The COLUMNS parsing, clamping, and unit tests are solid, but a few edge and portability cases are worth tightening.

- `TableBuilder.java:29` - `MAX_COLUMN_WIDTH = resolveMaxColumnWidth()` runs in a static initializer, so loading the class now forks `stty` and can block up to 500ms. The title and body describe only the COLUMNS env var plus a default; the subprocess detection is an undescribed second mechanism.
- `TableBuilder.java:84` - `min(candidate, UPPER_MAX_COLUMN_WIDTH)` is unqualified and the added import block (lines 9-11) has no `import static java.lang.Math.min`. Unless that import already exists unchanged, this will not compile.
- `TableBuilder.java:57` - A blank `COLUMNS` (some shell init files export an empty value) is non-null, so it enters `resolveMaxColumnWidth(String)`, fails to parse, and returns the 25 default without ever calling `detectTerminalWidth()`.
- `TableBuilder.java:104` - `/bin/stty` is hardcoded; on systems where `stty` lives only in `/usr/bin` (minimal containers, some distros), detection silently fails.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
nidhiii128: This already compiles import static java.lang.Math.min; was already in the file before my change (used by the original Cell wrapping logic), I never touched it. Ran mvn -pl cli compile plus the full TableBuilderTest suite (22/22) locally, both pass. The diff just didn't need to re-add an import that was already there.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
nidhiii128: Fixed: maxColumnWidth is now resolved and cached lazily on first use instead of eagerly at class load, so just loading TableBuilder (or building one that never wraps a value) no longer forks stty. Also updating the PR description to call out the stty size fallback explicitly, since you're right that it wasn't mentioned.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
nidhiii128: Fixed. A blank (but non-null) COLUMNS is now treated the same as unset, so it falls through to live detection instead of silently returning the 25 default.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
nidhiii128: This one's a bit tangled I originally had a bare "stty" resolved via PATH, but that's exactly what failed the SonarCloud Quality Gate on the previous push (java:S4036, resolving executables via PATH). So I can't revert to PATH resolution without likely failing Sonar again. Instead I now check a short list of trusted absolute paths (/bin/stty, then /usr/bin/stty) via Files.isExecutable, which avoids the PATH-search finding while still covering the case you flagged. Let me know if there's another common path worth adding.

paoloantinori:
Thanks, the lazy-init fix lands it. `maxColumnWidth()` now only resolves (and potentially forks `stty`) on first use inside `Cell(value)`, so class load and constructing a table that never wraps a long value both stay free of subprocess forks. The `synchronized` + null-cache pattern is the right shape for a once-per-JVM resolution.

On the `min(...)` import point I raised earlier: that was a false alarm on my end — the existing `Cell` code already used `min(...)` against `java.lang.Math`, so the static import was already in place. Sorry for the noise there.

PR description update is appreciated too — calling out the `stty size` fallback explicitly helps anyone reading the diff cold.

jsenko (commented):
## Review

Good evolution — the stty fallback addresses vandanayadav7's original concern about COLUMNS not being exported. The lazy+cached approach and synchronized initialization are correct.

### should-fix: Skip stty detection on Windows

`detectTerminalWidth()` tries to locate `/bin/stty` or `/usr/bin/stty`, which don't exist on Windows. The fallback works (returns null → uses default 25), but forking a subprocess that silently fails is unnecessary. Add an early return:

```java
static Integer detectTerminalWidth() {
    if (System.console() == null || System.getProperty("os.name", "").toLowerCase(Locale.ROOT).contains("win")) {
        return null;
    }
    // ... rest of stty logic
}
```

This matters because PR #9229 just added Windows support to the CLI.

### should-fix: Add boundary test for COLUMNS=80 and COLUMNS=81

The tests cover null, non-numeric, below-min, in-range, whitespace, at-min — but not the upper bound. Add:

```java
@Test
void resolveMaxColumnWidthAtUpperBoundIsUsedAsIs() {
    assertEquals(80, TableBuilder.resolveMaxColumnWidth("80"));
}

@Test
void resolveMaxColumnWidthAboveUpperBoundIsCapped() {
    assertEquals(80, TableBuilder.resolveMaxColumnWidth("81"));
}
```

### optional: Extract 500ms timeout to named constant

Line 133: `process.waitFor(500, TimeUnit.MILLISECONDS)` — extract to `TERMINAL_WIDTH_DETECTION_TIMEOUT_MS` for clarity.

nidhiii128:
Hi @jsenko, Thanks for the review! Added all three - the Windows check, the two boundary tests, and named that timeout constant.
One small correction though on the Windows point: turns out no subprocess was actually being forked there in the first place. The code checks if stty exists at /bin/stty or /usr/bin/stty before trying to run it, and neither path exists on Windows, so it was already bailing out early without ever launching anything. Still added the explicit check you suggested though — now that #9229 added real Windows support, it's clearer to just say "skip this on Windows" directly instead of relying on those file checks failing as a side effect.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -241,4 +355,4 @@ public void print(StringBuilder out) {
             out.append('\n');
         }
     }
-}
+}
```
aclfe: nit: file's missing a trailing newline again, this got fixed once already earlier in the PR and came back in this push.

aclfe (commented):
nice iteration on this thread, the stty size pivot (absolute path lookup, windows short circuit, lazy/cached resolution) addresses what @paoloantinori and @vandanayadav7 raised earlier.

one thing worth a look though, not a code issue: #9479 was opened proposing a JNA/picocli detector, on the assumption that the COLUMNS-only approach here didn't really solve the terminal width problem. that's not true anymore now that this does real detection via stty size, might be worth closing #9479 or linking it here so it doesn't turn into duplicate work.

left two small inline notes below, neither blocking.

cli/src/main/java/io/apicurio/registry/cli/utils/TableBuilder.java
```diff
@@ -36,6 +45,111 @@ public class TableBuilder {
...
+        }
+        var sttyPath = STTY_CANDIDATES.stream().filter(p -> Files.isExecutable(Path.of(p))).findFirst();
+        if (sttyPath.isEmpty()) {
+            return null;
+        }
+        try {
+            var process = new ProcessBuilder(sttyPath.get(), "size")
+                    .redirectInput(ProcessBuilder.Redirect.INHERIT)
+                    .start();
+            if (!process.waitFor(TERMINAL_WIDTH_DETECTION_TIMEOUT_MS, TimeUnit.MILLISECONDS)) {
```
aclfe, line 138: waitFor is called before getInputStream is drained a few lines down. if a child process ever writes more than the pipe buffer holds before it's read, this could time out and get killed even though it'd have finished fine. not a real issue today since stty size only writes a few bytes, but this kind of helper tends to get copied for commands with more output later.

aclfe:
heads up, there's another open PR (#8724) that covers the same underlying TODO in TableBuilder.java, tracked under issue #8647. that one does real terminal width detection via Jansi plus fair-share column width allocation across the whole table, and keeps wrapping instead of truncating so no content is lost. it's already been through several rounds of review (0 outstanding findings on the latest pass) and CI is green.

this PR only resolves single-cell wrapping via COLUMNS/stty, and wasn't linked to #8647 since it started from the inline TODO comment rather than a filed issue.

worth reconciling the two so we don't land duplicate/overlapping work, might make sense to close this one in favor of #8724.

@vandanayadav7 @paoloantinori thoughts?

nidhiii128:
Hi @aclfe Thanks for catching all of these, Fixed the trailing newline again. Also added a comment explaining the waitFor-before-read thing, you're right that it could be a problem in general, but stty size only ever outputs a few bytes, way too small to ever fill up the buffer and cause that kind of block. Left a note in the code so nobody copies this for a command with bigger output later without noticing.

Also added a line to the PR description linking #9479 , didn't want to close someone else's issue myself, but flagged that it's already solved by this PR now that it does real detection instead of just reading COLUMNS.

paoloantinori:
Thanks for flagging this, and the read is accurate. I pulled #8724 to verify: it does detect width via `TerminalUtils.getTerminalWidth()`, wraps cell content across multiple lines instead of truncating (so nothing is lost), and allocates column widths fairly across the whole table. Its latest head (`b8d37eb`) is green on the Verification Gate and the review trail there (`80090279`, `d661fd80`) was clear on the last pass.

On the merits, #8724 is the broader fix for the same TODO in `TableBuilder.java`. #9256 is the narrower one (single-cell width via `COLUMNS`/`stty`). They genuinely overlap, so landing both would mean two solutions to the same problem shipping close together.

My read leans toward #8724 as the one to take forward: no-content-lost wrapping is the right default for a CLI table, and fair-share column allocation handles the multi-column case this PR does not address. But that's a maintainer call on which to merge, and the work in #9256 (lazy cached resolution, the `stty` absolute-path lookup, the Windows short-circuit) is solid on its own and could be reused if #8724 stalls.

Deferring to @vandanayadav7 and @jsenko here. No further comments from me on #9256's current head.

aclfe:
Thanks, that sounds great. I wasn't sure on what's optimal so had to ping maintainers

nidhiii128:
@vandanayadav7 @jsenko - appreciate you both looking into this. To be clear on scope: #9256 solves how much width is actually available (real terminal detection via stty, with the lazy/cached resolution and cross-platform handling already reviewed through four rounds here). #8724 solves how that width gets divided across columns and wrapped without losing content a different, complementary layer, not strictly the same fix.

@vandanayadav7 - one more thing: #9479 (which you opened) is now assigned to @shuwangs, who's planning to implement via picocli. Between #9256, #8724, and #9479's assignment, there's a real risk of three separate implementations of the same TODO landing independently. Might be worth coordinating across all three before more work goes in.

For my part: if you'd rather take #8724 forward as the single source of truth, that's reasonable but wanted to flag that the width-detection piece here could just as easily be the input #8724's allocation logic uses, rather than one fully replacing the other. Let me know which way you want to go and I'll follow your lead.

vandanayadav7:
@nidhiii128 Thanks for the work on this and for iterating based on feedback! The `stty size` approach was a good improvement over the initial `COLUMNS`-only version.

However, PR #8724 takes a more complete approach to the same problem — it detects terminal width using Jansi and also does fair-share column allocation so the table actually fits within the terminal. Your PR detects the width but still uses it as a per-column cap, so a table with many columns can still overflow.

Closing in favor of #8724 which covers issue #8647. Appreciate the effort!

PR #9257, opened 2026-08-05
fix(ui): PromptTemplateViewer hides enum allowed values in Variables documentation table

### LINKED ISSUE #9103: fix(ui): PromptTemplateViewer hides enum allowed values in Variables documentation table

aruhis10:
Fixes #9103

   ## What
   The Variables table in `PromptTemplateViewer` renders Name, Type,
   Required, Default, and Description columns, but silently dropped
   the `enum` field even though `PromptVariable` already defines it.

   ## Fix
   Added an "Allowed Values" column that renders each `enum` option
   as a `Label` when present, and falls back to `-` otherwise
   (consistent with the other optional columns like Default).

   ## Testing
   Verified visually against a sample PROMPT_TEMPLATE artifact with
   an enum-constrained variable.


paoloantinori (commented):
Clean, minimal change that surfaces the already-defined `enum` field, with both the undefined and empty-array cases handled and a `-` fallback consistent with the other optional columns. I see no correctness, security, or edge-case issues worth blocking on; the change is safe to render artifact-controlled content.

- `PromptTemplateViewer.tsx:187` guards `variable.enum` against both undefined/null and `[]` via `&& variable.enum.length > 0`, so a variable with no allowed values still renders `-` as intended.
- `PromptTemplateViewer.tsx:190` renders enum values as escaped React text children inside `Label`, so there is no XSS surface even though the content originates from a PROMPT_TEMPLATE artifact.
- `PromptTemplateViewer.tsx:190` uses `key={i}` (array index). This is acceptable because the list is static per render, and if enum values could ever contain duplicates, index keys are actually the safer choice over `key={val}`, so I would leave it.
- Minor, non-blocking: enum values render raw while the Default cell uses `formatDefault` and `<code>`. Enum values are primitive literals in practice, so this only matters if the artifact schema is ever loosened to allow non-primitive members.

Nachiket-Roy:
There is already a few PRs linked to #9103 please avoid duplicate work

PR #9259, opened 2026-08-05
fix(ui): keep page in range when page size changes (#8929)

### LINKED ISSUE #8929: UI: changing the page size keeps the current page number, which can show an empty list (onPerPageSelect ignores PatternFly's computed newPage)
## Summary

Across every paginated list in `ui-app`, the `onPerPageSelect` handler builds the new paging state from the current page number and ignores the `newPage` value that PatternFly's `Pagination` passes as its third callback argument. When the page size is increased while the user is on a page greater than 1, the current page can become out of range and the list renders empty.

## Affected code

PatternFly `Pagination` calls `onPerPageSelect(event, newPerPage, newPage)`, where `newPage` is recomputed so the first visible item stays in view after the page size changes. The handlers keep `paging.page` instead:

- `ui/ui-app/src/app/pages/explore/ExplorePage.tsx:121`
- `ui/ui-app/src/app/pages/search/components/toolbar/SearchPageToolbar.tsx:190`
- `ui/ui-app/src/app/pages/roles/components/roleToolbar/RoleToolbar.tsx:147`
- `ui/ui-app/src/app/pages/drafts/components/toolbar/DraftsPageToolbar.tsx:68`
- `ui/ui-app/src/app/pages/group/components/tabs/GroupArtifactsToolbar.tsx:46`
- `ui/ui-app/src/app/pages/artifact/components/tabs/BranchesTabToolbar.tsx:34`
- `ui/ui-app/src/app/pages/artifact/components/tabs/ArtifactVersionsToolbar.tsx:50`
- `ui/ui-app/src/app/pages/branch/components/tabs/BranchVersionsToolbar.tsx:31`
- `ui/ui-app/src/app/pages/version/components/tabs/ReferencesTabContent.tsx:138`

Representative handler:

```tsx
const onPerPageSelect = (_event: any, newPerPage: number): void => {
    const newPaging: Paging = {
        page: paging.page,     // keeps the old page, ignores the recomputed newPage
        pageSize: newPerPage
    };
    ...
};
```

For comparison, the `onSetPage` handlers in the same files already use the `newPage` argument correctly, so the codebase established the right pattern; `onPerPageSelect` just drops the third argument.

## Steps to reproduce

1. Open a paginated list with more than one page of results (for example the Explore page with a small page size).
2. Go to page 3 with a page size of 10.
3. Change the page size to 20 (or 50).

## Expected

The list stays on a valid page showing results (PatternFly's recomputed `newPage`, which keeps the first item of the previous view visible).

## Actual

The page number stays at 3. With the larger page size, page 3 is now past the end of the result set, so the list renders empty until the user manually pages back.

## Fix

Accept and use the third argument in every `onPerPageSelect`:

```tsx
const onPerPageSelect = (_event: any, newPerPage: number, newPage: number): void => {
    const newPaging: Paging = { page: newPage, pageSize: newPerPage };
    ...
};
```

(Resetting to page 1 would also avoid the empty list, but using PatternFly's `newPage` matches the intended behavior and the existing `onSetPage` handlers.)

## Notes

Consistent across nine list toolbars, so a single well-scoped change. Verified against current `main`.

Aditya-132:
## What

Every `onPerPageSelect` handler in `ui-app` now accepts and uses the third argument that PatternFly's `Pagination` passes (`newPage`), instead of reusing the current page number.

Nine list toolbars are affected:

- `ui/ui-app/src/app/pages/explore/ExplorePage.tsx`
- `ui/ui-app/src/app/pages/search/components/toolbar/SearchPageToolbar.tsx`
- `ui/ui-app/src/app/pages/roles/components/roleToolbar/RoleToolbar.tsx`
- `ui/ui-app/src/app/pages/drafts/components/toolbar/DraftsPageToolbar.tsx`
- `ui/ui-app/src/app/pages/group/components/tabs/GroupArtifactsToolbar.tsx`
- `ui/ui-app/src/app/pages/artifact/components/tabs/BranchesTabToolbar.tsx`
- `ui/ui-app/src/app/pages/artifact/components/tabs/ArtifactVersionsToolbar.tsx`
- `ui/ui-app/src/app/pages/branch/components/tabs/BranchVersionsToolbar.tsx`
- `ui/ui-app/src/app/pages/version/components/tabs/ReferencesTabContent.tsx`

```diff
-const onPerPageSelect = (_event: any, newPerPage: number): void => {
+const onPerPageSelect = (_event: any, newPerPage: number, newPage: number): void => {
     const newPaging: Paging = {
-        page: props.paging.page,
+        page: newPage,
         pageSize: newPerPage
     };
```

## Why

`Pagination` calls `onPerPageSelect(event, newPerPage, newPage)`, where `newPage` is recomputed so the first item of the current view stays visible after the page size changes. The handlers kept the old page number, so increasing the page size while on a page greater than 1 could leave the page number past the end of the result set and the list rendered empty until the user manually paged back.

Reproduction: open the Explore page with a page size of 10, go to page 3, then change the page size to 20. The list goes empty.

The `onSetPage` handlers in the same files already use their `newPage` argument, so this brings `onPerPageSelect` in line with the pattern the codebase had already established.

Resetting to page 1 would also avoid the empty list, but using PatternFly's `newPage` preserves the user's scroll position in the result set, which is the intended behavior of the component.

## Testing

`ui-app` has no component-level test harness (`vitest` runs with `environment: "node"` and there is no DOM testing library), and this change contains no extractable logic to unit test: it is a pass-through of an argument the component already supplies. Reviewers can confirm the behavior with the reproduction steps above. Happy to add a DOM test harness in a follow-up if maintainers want one.

## Scope

A grep of the whole `ui/` tree for `onPerPageSelect` turns up exactly ten consumers. The nine above are fixed here. The tenth, `AgentsPage.tsx` (`handlePerPageChange`), hard-resets to `page: 1` rather than reusing the current page, so it cannot land out of range and does not exhibit this bug — it is deliberately left untouched. Its behavior does differ from the other nine (it jumps to the top of the result set instead of keeping the first visible item in view), but that is a UX inconsistency rather than a defect and belongs in a separate change.

Fixes #8929


paoloantinori (commented):
This is a correct, minimal fix: all nine onPerPageSelect handlers now thread PatternFly's recomputed newPage into the paging object instead of reusing the stale current page, matching the pattern the existing onSetPage handlers already follow. The change cannot regress behavior, because if PatternFly passed the unchanged current page as the third argument the result would be identical to the old code; it only helps when PatternFly recomputes the page, which the onPerPageSelect?: (event, perPage, page) => void signature and the reproduction in the PR description both support.

- The swap is applied consistently across all nine toolbars, changing `page: props.paging.page` (or `paging.page` for the state-based pages) to `page: newPage` while leaving the downstream onPageChange/setPaging/search call untouched (e.g. ExplorePage.tsx:147, SearchPageToolbar.tsx:192). Each handler also correctly accepts the new third parameter on its signature.
- Page-clamping risk is delegated to PatternFly, which computes a page that keeps the current view's first item in range, so no additional bounds logic is needed on the consumer side. The worst case across PF versions is a no-op, never a regression.
- Worth a grep to confirm no other onPerPageSelect handlers elsewhere in the codebase still reuse the stale `page` field; this PR covers the nine list toolbars but the same anti-pattern could exist in a tenth component that was not enumerated.

Aditya-132:
Thanks for the review!

On the grep suggestion  I checked the whole `ui/` tree for `onPerPageSelect`. There are exactly ten consumers. Nine are the ones covered here. The tenth is `AgentsPage.tsx` (`handlePerPageChange`, line 127):

```tsx
const handlePerPageChange = (_event: any, perPage: number): void => {
    const newPaging = { page: 1, pageSize: perPage };
    setPaging(newPaging);
    search(createFilters(), newPaging);
};
```

It hard-resets to page 1 instead of reusing the current page, so it can never land out of range and does not exhibit the empty-list bug. I left it untouched to keep this PR minimal.

It is worth noting its behavior still differs from the other nine: PatternFly's `newPage` keeps the first visible item in view, whereas `AgentsPage` jumps back to the top of the result set. That is a UX inconsistency rather than a defect, so I would rather not fold it into this PR  happy to open a follow-up issue if you would like the ten aligned.

One other thing: the `Verification Gate` check is currently red with `PR is not in a testable lifecycle state`, and every test job is skipped. As far as I can tell that is because the PR has not been `/accept`ed yet rather than anything in the diff  SonarCloud, Security Analysis, License Compliance and Dependency Quality all pass. Let me know if there is anything you need from me to move it along.

paoloantinori:
Thanks for the thorough check, that confirms it. `AgentsPage.handlePerPageChange` hard-resets to `page: 1`, so it can't land out of range and excluding it keeps the fix minimal. Agree the UX inconsistency (reset-to-top vs. PatternFly's keep-first-visible) is a separate matter; a follow-up issue to align all ten would be useful but doesn't belong here.

On the `Verification Gate`: yes, the skipped jobs are because the PR hasn't been `/accept`ed yet, not anything in the diff. That's a maintainer call (Eric or Jakub), not something I can action.

Aditya-132:
@EricWittmann is it fine now?

EricWittmann:
Thanks @Aditya-132 — excellent contribution! Clean fix, thorough investigation, and great PR description. Merged.

PR #9270, opened 2026-08-05
fix(rendering): treat null value as missing for required prompt template variables

### LINKED ISSUE #9215: [Bug]: Explicit null value for a required prompt template variable bypasses "required" validation but leaves placeholder unsubstituted
## Description

**Registry Version**: main (snapshot / latest-snapshot docker image)

The /render endpoint for PROMPT_TEMPLATE artifacts checks whether a required variable is present using variables.containsKey(varName). If a client explicitly sends a required variable with a null value (the key is present, but its value is null), this check passes (no validation error is reported) — but during substitution, the 
variable is treated as missing and the {{variable}} placeholder is left unsubstituted in the output.

This means the API response reports validationErrors: [] (implying success), while the actual rendered output still contains a raw, unsubstituted placeholder — which is misleading and inconsistent.

### Environment

Running via Docker (apicurio/apicurio-registry:latest-snapshot and apicurio/apicurio-registry-ui:latest-snapshot), backend on port 8080. Tested via direct REST API calls (curl).

### Steps to Reproduce

1. Create a PROMPT_TEMPLATE artifact with a required variable, e.g.:
   { "template": "You need {{count}} items.", "variables": 
   { "count": { "type": "integer", "required": true } } }
2. Call the render endpoint with the variable explicitly set to null:
   Body: { "count": null }

**Expected behavior**
Either:
(a) The API should return a validation error for "count" (since a null value doesn't satisfy a required variable), or
(b) The placeholder should be substituted with some representation of null (e.g. empty string), consistent with a "successful" validation result.

**Actual behavior**
The API returns validationErrors: [] (no errors), but the rendered output still contains the literal, unsubstituted {{count}} placeholder — silently contradicting the "no errors" response.

### Root Cause
In PromptRenderingService.java's validateVariables method, the required check only tests variables.containsKey(varName), not whether the value is non-null:

if (required && !variables.containsKey(varName)) {
errors.add(...);
}

Since a map can contain a key with a null value, containsKey returns true even when the value is null, so the required check is bypassed.Later, in substituteVariables, the null value causes the placeholder to be left as-is:

if (value == null) {
    replacement = matcher.group(0); // placeholder kept as-is
}

This creates an inconsistency between what validation reports and what the rendered output actually contains.

### Logs

(No error logs — this is a silent validation/substitution 
inconsistency, not a crash.)

jayaprajapatii:
Fixes #9215

## PR Summary
The `/render` endpoint's required-variable check only tested 
`variables.containsKey(varName)`, which returns `true` even when the key is present with an explicit `null` value. This let a required variable set to `null` bypass validation, while the substitution logic (which already checks `value == null`) left the `{{variable}}` placeholder unsubstituted — creating an inconsistency between the reported validation result and the actual rendered output.

## Fix
Updated the required-variable check in `validateVariables` to also treat an explicit `null` value as "missing":

```java
if (required && (!variables.containsKey(varName) || variables.get(varName) == null)) {
    errors.add(...);
}
```

This makes validation consistent with the existing substitution behavior, which already treats `null` as "not provided."

## Testing
Verified the reported scenario against the current code:
* `variables = {"count": null}` for a required `count` variable
* Before fix: `validationErrors: []`, but output contains raw `{{count}}` placeholder
* After fix: validation correctly reports `count` as a missing required variable


EricWittmann:
Thanks for the contribution @jayaprajapatii! Nice clean fix.

Issue #9264, opened 2026-08-05
Unauthenticated SSRF via unresolved $ref in JSON Schema compatibility checking


1PoPTRoN:
## Description

**Registry Version**: `snapshot` (reproduced on `main` @ 34b691a00, `3.3.2-SNAPSHOT`). The affected code dates back to #3228 (2023-03-28), so I believe every released 3.x version is affected.

**Persistence type**: All — the defect sits in `schema-util/json`, above the storage SPI, so it's independent of the backend.

> _Edit: re-posting the body — several angle-bracketed tokens (Java generics, the listener placeholder, and the assertion values in the logs) were stripped as HTML on first submission, which left the reproduction payload and the log output unreadable._

When a COMPATIBILITY rule is active, `JsonSchemaDiffLibrary.loadReferences()` registers schemas with the everit loader by iterating the **resolved references map** rather than the references actually present in the document:

```java
for (Map.Entry<String, TypedContent> stringStringEntry : resolvedReferences.entrySet()) {
    URI child = ReferenceResolver.resolve(idUri, stringStringEntry.getKey());
    schemaLoaderBuilder.registerSchemaByURI(child, new JSONObject(...));
}
```

Any `$ref` in the submitted schema that has no recorded content is therefore never registered. In `ReferenceLookup.createPointerEvaluator`, an unregistered URI falls through to `JsonPointerEvaluator.forURL(schemaClient, ...)`, which uses everit's default `DefaultSchemaClient` and calls `URL.openStream()`. The server then issues an outbound request to a URL the submitter controls.

What made me confident this is unintended rather than a design choice: the sibling method `JsonUtil.readSchemaEverit()` (same file cluster, used by the validity path) already carries the mitigation, with a comment stating it exists specifically to "prevent the library from attempting to download the schema if `http://`, or trying to open a file if `file://`". The compatibility path simply never received the same treatment. This looks like incomplete propagation of an earlier hardening rather than a new class of bug.

Because `quarkus.oidc.tenant-enabled=false` is the default (`app/src/main/resources/application.properties:5` and `AuthConfig.java:35`), a default deployment is reachable without credentials. `file://` and other JDK-supported protocols appear reachable by the same path.

### Environment

Local build of `main`, Temurin JDK 21.0.12, Maven wrapper, Linux. Reproduced through the `schema-util/json` module directly; no serdes client involved. Not environment-specific — it needs only a running Registry with a COMPATIBILITY rule in effect.

[...]

PR #9271, opened 2026-08-05
fix(jsonschema): block unresolved ref fetches (#9264)

KartikSuryavanshi:
## Summary

  This fixes an SSRF issue in JSON Schema compatibility checking when a schema contains an unresolved
  `$ref`.

  Previously, the compatibility path registered only the reference entries supplied by the caller. If a
  `$ref` appeared in the schema but had no matching resolved content, Everit could fall back to its default
  remote schema client and attempt to fetch the URI over the network.

  This change makes compatibility checking fail closed:

  - it walks the actual `$ref` values present in the schema
  - it registers the provided referenced content when available
  - it registers a permissive placeholder schema when content is missing
  - it installs a deny-all schema client so unresolved references cannot trigger outbound fetches

  A regression test was added to verify that an unresolved remote `$ref` does not cause the checker to
  perform network access.

  ## Verification

  - `./mvnw -pl schema-util/json -Dtest=JsonSchemaRemoteReferenceTest,JsonSchemaCompatibilityCheckerTest
  test`


schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/jsonschema/JsonSchemaDiffLibrary.java
```diff
@@ -79,10 +82,25 @@ private static void loadReferences(JsonNode jsonNode, Map<String, TypedContent>
...
         }
 
-        for (Map.Entry<String, TypedContent> stringStringEntry : resolvedReferences.entrySet()) {
-            URI child = ReferenceResolver.resolve(idUri, stringStringEntry.getKey());
-            schemaLoaderBuilder.registerSchemaByURI(child,
-                    new JSONObject(stringStringEntry.getValue().getContent().content()));
+        schemaLoaderBuilder.httpClient(DENY_ALL_SCHEMA_CLIENT);
+
+        Set<URI> extractedReferences = JsonUtil.extractReferencesRecursive(JsonSchemaVersion.valueOf(spec.name()), idUri, jsonNode);
+        for (URI extractedReference : extractedReferences) {
```
paoloantinori, line 84: Before this change every entry in resolvedReferences was registered by resolving its key against the schema $id (ReferenceResolver.resolve(idUri, key)). The new code looks up the map with extractedReference.toString(), the fully resolved absolute URI, and extractReferencesRecursive resolves $ref against each lexical $id scope rather than only the root. If the registry keys resolvedReferences by the raw $ref strings it stores as reference names, this lookup never matches, so every provided referenced schema is silently replaced by the accept-all placeholder and the compatibility check for those schemas becomes a no-op. Verify the map keying; if it is keyed by raw refs, resolve the lookup key against idUri as the old code did. The added regression test uses an empty map, so a test that supplies real referenced content is needed to cover this path.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/jsonschema/JsonSchemaDiffLibrary.java
```diff
@@ -79,10 +82,25 @@ private static void loadReferences(JsonNode jsonNode, Map<String, TypedContent>
...
+                        new JSONObject(resolvedReferenceContent.getContent().content()));
+            } else {
+                /*
+                 * Since we do not have the referenced content, we insert a placeholder schema, that will
+                 * accept any JSON, to the reference lookup table of the library. This prevents the library
+                 * from attempting to download the schema if `http://`, or trying to open a file if `file://`.
+                 * This avoids potential security issues by us having to explicitly provide referenced
+                 * content. For compatibility checks, we do not care about the reference format, while still
+                 * requiring a valid URI.
+                 */
```
paoloantinori: An unresolved $ref now resolves to new JSONObject(), an accept-all schema, so the compatibility check silently passes for that reference instead of failing. Network access fails closed via DENY_ALL_SCHEMA_CLIENT (good), but content resolution fails open, and the PR summary's "fail closed" framing does not capture that. If silent permissiveness on missing references is intended, state it explicitly; otherwise consider throwing so an unresolved reference produces a clear error instead of a potentially misleading compatible verdict.

schema-util/json/src/test/java/io/apicurio/registry/rules/compatibility/jsonschema/JsonSchemaRemoteReferenceTest.java
```diff
@@ -0,0 +1,64 @@
...
+                    "x": {
+                      "$ref": "http://127.0.0.1:1/"
+                    }
+                  }
+                }
+                """;
+
+        assertDoesNotThrow(() -> checker.testCompatibility(CompatibilityLevel.BACKWARD,
+                Collections.singletonList(toTypedContent(EXISTING_SCHEMA)),
+                toTypedContent(proposedSchema), Collections.emptyMap()));
```
paoloantinori, line 64: This test passes Collections.emptyMap() for resolved references, so it only exercises the placeholder (no-content) branch. It cannot detect a regression in the path where referenced content is supplied and should actually be used. Add a test that registers a real referenced schema for a $ref and asserts the checker uses it (for example, an incompatible referenced schema is rejected). Asserting a concrete compatibility verdict would also be stronger than assertDoesNotThrow.

paoloantinori (commented):
The core SSRF fix is sound: installing a deny-all SchemaClient and registering a permissive placeholder for unresolved $ref values reliably blocks outbound fetches during compatibility checking, and reusing the existing JsonUtil.extractReferencesRecursive is the right call rather than duplicating the ref walk. Two things need attention before merge: the reference-lookup key changed semantics and is not covered by the added test, and unresolved references now silently validate as accept-all, a behavioral change the summary does not reflect.

- JsonSchemaDiffLibrary.java:88 - The old loop registered every entry in resolvedReferences by resolving each key against the schema $id (ReferenceResolver.resolve(idUri, key)). The new code looks up the map with extractedReference.toString(), the fully resolved absolute URI, and extractReferencesRecursive resolves $ref against each lexical $id scope rather than only the root. If resolvedReferences is keyed by the raw $ref strings the registry stores as reference names, this lookup never matches and every provided referenced schema is silently replaced by the accept-all placeholder, turning the compatibility check for those schemas into a no-op. Confirm the map keying, and if it is keyed by raw refs, resolve the lookup key against idUri as before.
- JsonSchemaDiffLibrary.java:101 - An unresolved $ref now resolves to new JSONObject(), an accept-all schema, so the compatibility check silently passes for that reference instead of failing. Network access fails closed via DENY_ALL_SCHEMA_CLIENT (good), but content resolution fails open, and the summary's "fail closed" framing does not capture that. If silent permissiveness on missing references is intended, state it explicitly in the summary and the comment; otherwise consider throwing so an unresolved reference surfaces a clear error instead of a potentially misleading compatible verdict.
- JsonSchemaRemoteReferenceTest.java:62 - The only regression test passes Collections.emptyMap(), so it exercises only the placeholder (no-content) branch and cannot detect the lookup-key regression above. Add a test that registers real referenced schema content under a $ref and asserts the checker uses it (for example, an incompatible referenced schema is rejected). A concrete compatibility verdict would also be stronger than assertDoesNotThrow.
- Minor: with the only ReferenceResolver.resolve call moved into JsonUtil, confirm the org.everit.json.schema.loader.internal.ReferenceResolver import in JsonSchemaDiffLibrary.java is still referenced so checkstyle stays green.

schema-util/json/src/test/java/io/apicurio/registry/rules/compatibility/jsonschema/JsonSchemaRemoteReferenceTest.java
KartikSuryavanshi: Agreed, and updated. Added JsonSchemaRemoteReferenceTest.java with a real resolved reference entry (address.json) and a concrete verdict assertion (JsonSchemaRemoteReferenceTest.java) to ensure the supplied referenced schema is used.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/jsonschema/JsonSchemaDiffLibrary.java
KartikSuryavanshi: Thanks, good catch. This commit restores the previous lookup semantics by resolving each resolvedReferences key against the root id and matching against extracted absolute refs.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/jsonschema/JsonSchemaDiffLibrary.java
KartikSuryavanshi: Agreed. I changed the compatibility path to fail closed when a referenced schema is not provided, so unresolved $ref values now raise a clear IllegalStateException instead of being treated as an accept-all schema. I also added a regression test covering the unresolved and resolved reference cases.

KartikSuryavanshi:
Hey @paoloantinori I have addressed your comments. Can you please review it again?

paoloantinori:
Re-read the current head (3cba7c3) after your changes. The fail-closed IllegalStateException on unresolved $ref plus the deny-all SchemaClient is exactly the shape I was after. The regression test covering both the unresolved and the resolved case pins the behavior down well.

No further change requests from my side.

KartikSuryavanshi:
Thanks a lot, @paoloantinori! I really appreciate the detailed review and feedback.

Hi @EricWittmann  - all the review comments have now been addressed, and @paoloantinori confirmed there are no further change requests. Could you please take a look and merge this when you have a chance? Thanks!

paoloantinori:
You're welcome — glad the review was useful. The current head (3cba7c3) still looks good from my side: the fail-closed `IllegalStateException` on unresolved `` together with the deny-all `SchemaClient` closes the SSRF window cleanly, and the new `JsonSchemaRemoteReferenceTest` exercises it. No further concerns from me on the JSON Schema compatibility path.

KartikSuryavanshi:
@jsenko Can you please merge this PR?

KartikSuryavanshi:
Hi @EricWittmann @jsenko Can you please merge this PR?

KartikSuryavanshi:
@EricWittmann @jsenko Friendly ping on this PR. It has been approved by @paoloantinori .

paoloantinori:
From my side nothing is pending: the review loop on this head is closed and I have no further comments. The merge itself is with @EricWittmann/@jsenko as you pinged.

One note before it lands: the branch is behind main, so it will need an update before merging. If that produces a new head, ping me and I will re-read it — the fix is small enough that it should be quick.

jsenko (commented):
Fixes #9264 — blocks SSRF via unresolved JSON Schema `$ref` fetches in the compatibility checking path.

### Summary

The deny-all `SchemaClient` and fail-closed approach are the right design. However, the current implementation breaks schemas that use internal fragment references (e.g. `#/definitions/Address`), which are common. The compatibility checking path now throws `IllegalStateException` for these internal references because it treats ALL unresolved entries in the `resolvedReferences` map the same way, regardless of whether they point to an external resource or an in-document fragment.

Additionally, the SSRF protection only covers the compatibility path (`JsonSchemaDiffLibrary`), not the validation path (`JsonUtil.readSchemaEverit`), which uses the same library without a deny-all client.

### must-fix: Internal fragment references cause IllegalStateException

`loadReferences` extracts all `$ref` URIs via `extractReferencesRecursive`, including internal fragment-only references like `#/definitions/Address`. These are resolved within the schema document itself and should not require an entry in `resolvedReferences`. The current code throws `IllegalStateException("Unresolved JSON Schema reference: ...")` for any extracted reference not found in the map.

I ran `JsonSchemaSmokeTest` locally and it fails with 6 broken test cases, all involving internal references:

```
References: $ref
References: $ref compatibility
References: $id
References: $id compatibility
References: recursive
References: Recursive in combined schema, Issue #1551
```

CI didn't catch this because the Build step was skipped (PR is still in `lifecycle/new`).

**Fix:** Before checking `resolvedReferences`, determine whether the reference is internal (a fragment within the same document). Internal fragment references should be skipped — the SchemaLoader resolves them from the document itself. One approach:

```java
for (URI extractedReference : extractedReferences) {
    if (isInternalFragment(extractedReference, idUri)) {
        continue;
    }
    // ... existing lookup + fail-closed logic
}
```

Where `isInternalFragment` checks whether the reference URI (sans fragment) matches the schema's `$id` URI, or is purely a fragment (`#/...`) with no authority/path.

**Verify with:** `./mvnw test -pl schema-util/json -am -Dtest=JsonSchemaSmokeTest -Dsurefire.failIfNoSpecifiedTests=false`

### should-fix: Validation path (`JsonUtil.readSchemaEverit`) lacks deny-all SchemaClient

The SSRF fix installs `DENY_ALL_SCHEMA_CLIENT` on the compatibility path but does not protect the validation path in `JsonUtil.readSchemaEverit` (line 141). That method creates a `SchemaLoader.builder()` without setting `.httpClient(...)`.

The validation path does register accept-all placeholders (`new JSONObject()`) for all extracted references, which prevents the SchemaLoader from fetching *known* references. However, if `extractReferencesRecursive` ever misses a reference pattern that the SchemaLoader's own parser recognizes, the default HTTP client would still attempt an outbound fetch.

Adding `.httpClient(DENY_ALL_SCHEMA_CLIENT)` to `readSchemaEverit` would close that gap as defense-in-depth.

KartikSuryavanshi:
Hey @jsenko Done! with the changes. Can you please review the changes again?

jsenko (commented):
Re-review of commit 2fcfa35 — author addressed both prior findings.

### Re-review delta

**Resolved:**
- **must-fix: Internal fragment references** — Fixed. Added `JsonUtil.isInternalFragment()` that compares the base URI (sans fragment) of the extracted reference against the schema's `$id`. Internal refs like `#/definitions/Address` are now skipped. Verified locally: `JsonSchemaSmokeTest` passes (0 failures, previously 6).
- **should-fix: Validation path SSRF** — Fixed. Added `.httpClient(DENY_ALL_SCHEMA_CLIENT)` to `readSchemaEverit` in `JsonUtil.java`. Both code paths (compatibility + validation) now install the deny-all client.
- `DENY_ALL_SCHEMA_CLIENT` moved to `JsonUtil` as a shared package-private constant — clean refactoring.

**Verification:** Both `JsonSchemaSmokeTest` (2 tests) and `JsonSchemaRemoteReferenceTest` (2 tests) pass locally.

No further issues found.

KartikSuryavanshi:
@jsenko Thanks for the review! Can we now get this PR merged?

KartikSuryavanshi:
@jsenko I am not sure these failures are pre-existing or due to my PR. I checked the logs but this seems like a avro-maven-plugin failing to compile Avro schemas in the app module's test resources. Can you confirm are these pre-existing or due to my PR?

jsenko:
Preexisting, we're working on a fix https://github.com/Apicurio/apicurio-registry/pull/9664

KartikSuryavanshi:
Hey @jsenko The checks were successfull. Can we now merge this?

jsenko (approved):

jsenko:
Done, thanks!

PR #9275, opened 2026-08-05
fix(kafkasql): wire up periodic KafkaSQL snapshot scheduling

### LINKED ISSUE #8499: Automatic KafkaSQL snapshot scheduling (apicurio.kafkasql.snapshot.every.seconds) is declared but never wired to any scheduler

EricWittmann:
## Summary

Fixes #8499. Supersedes #8517.

The `apicurio.kafkasql.snapshot.every.seconds` property was declared in `KafkaSqlConfiguration` but never consumed by any scheduler. Snapshots only happened via the manual `POST /admin/snapshots` endpoint, so the `kafkasql-journal` topic grew unbounded.

This PR adds a `KafkaSqlSnapshotScheduler` that periodically triggers snapshot creation at the configured interval. Key design decisions:

- **Opt-in**: A new dynamic property `apicurio.kafkasql.snapshot.scheduled.enabled` (default `false`) gates the feature. Operators can toggle it at runtime via the REST API or Settings UI without restarting.
- **Multi-replica safe**: Before triggering, the scheduler checks the `kafkasql-snapshots` topic for a recent snapshot. If one was published within the interval, the run is skipped — ensuring only one replica creates a snapshot per cycle.
- **Runtime storage guard**: The `@Scheduled` method early-returns on non-kafkasql storage variants, since `@LookupIfProperty` does not prevent Quarkus from invoking `@Scheduled` methods.
- **UI visibility**: The property only appears in the Settings UI when `apicurio.storage.kind=kafkasql` (via `@Dynamic(requires=...)`). A new "Storage settings" group in the Settings page houses it alongside `apicurio.storage.read-only.enabled`.

### Changes

- `KafkaSqlSnapshotScheduler.java` — New scheduler bean
- `KafkaSqlSnapshotSchedulerTest.java` — Unit tests (happy path, skip conditions, disabled flag, non-kafkasql variants, interval parsing)
- `KafkaSqlConfiguration.java` — Improved `@Info` description for `snapshotEvery`
- `application.properties` — `.dynamic.allow` entry for the new property
- `ref-registry-all-configs.adoc` — Regenerated config docs
- `SettingsPage.tsx` — New "Storage settings" group in the UI

### Attribution

Based on the original work by @nagindersingh in #8517, with the following additions:
- Opt-in enabled flag (dynamic, runtime-toggleable)
- Multi-replica coordination via snapshots topic check
- `@Dynamic` with `requires` to conditionally show in UI
- Settings UI grouping

## Test plan

- [ ] Unit tests pass for all scheduler scenarios
- [ ] KafkaSQL storage variant unit tests pass (`app-kafkasql` profile)
- [ ] Verify the property appears in Settings UI only on kafkasql deployments
- [ ] Verify scheduled snapshot triggers after enabling the property
- [ ] Verify only one snapshot is created in a multi-replica deployment

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
@@ -70,7 +70,7 @@ public class KafkaSqlSnapshotScheduler {
     @Inject
     KafkaSqlConfiguration configuration;
 
-    private volatile boolean initialDelayApplied = false;
+    volatile boolean initialDelayApplied = false;
 
     @Scheduled(delay = MIN_INITIAL_DELAY_SECONDS, delayUnit = TimeUnit.SECONDS, concurrentExecution = SKIP, every = "{apicurio.kafkasql.snapshot.every.seconds}")
     void run() {
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
@@ -12,7 +12,7 @@ import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.serialization.StringDeserializer;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.slf4j.Logger;
 
@@ -20,7 +20,9 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
@@ -36,16 +38,18 @@ import static io.quarkus.scheduler.Scheduled.ConcurrentExecution.SKIP;
  * {@code @Scheduled} methods at build time regardless of it. The runtime checks in {@link #run()} are what
  * actually prevent this from running on other storage variants or when disabled.
  * <p>
- * In a multi-replica deployment, every replica runs its own scheduler. To avoid redundant snapshots, the
+ * In a multi-replica deployment, every replica runs its own scheduler. To reduce redundant snapshots, the
  * scheduler checks the {@code kafkasql-snapshots} topic before triggering: if a snapshot was published
- * within the configured interval, the run is skipped.
+ * within the configured interval, the run is skipped. This is best-effort deduplication (not a hard
+ * guarantee), since it relies on a check-then-act pattern without distributed locking.
  */
 @ApplicationScoped
 @LookupIfProperty(name = "apicurio.storage.kind", stringValue = "kafkasql")
 public class KafkaSqlSnapshotScheduler {
 
-    private static final long INITIAL_DELAY_SECONDS = 60L;
     private static final String KAFKASQL_STORAGE_KIND = "kafkasql";
+    private static final long MIN_INITIAL_DELAY_SECONDS = 60L;
+    private static final long MAX_INITIAL_DELAY_JITTER_SECONDS = 30L;
 
     @ConfigProperty(name = "apicurio.storage.kind")
     @Info(category = CATEGORY_STORAGE, description = "Application storage variant, for example, sql, kafkasql, gitops, or kubernetesops", availableSince = "3.0.0")
@@ -66,15 +70,35 @@ public class KafkaSqlSnapshotScheduler {
     @Inject
     KafkaSqlConfiguration configuration;
 
-    @Scheduled(delay = INITIAL_DELAY_SECONDS, delayUnit = TimeUnit.SECONDS, concurrentExecution = SKIP, every = "{apicurio.kafkasql.snapshot.every.seconds}")
+    private volatile boolean initialDelayApplied = false;
+
+    @Scheduled(delay = MIN_INITIAL_DELAY_SECONDS, delayUnit = TimeUnit.SECONDS, concurrentExecution = SKIP, every = "{apicurio.kafkasql.snapshot.every.seconds}")
     void run() {
         if (!KAFKASQL_STORAGE_KIND.equals(registryStorageType) || !scheduledSnapshotsEnabled.get()) {
             return;
+        }
+        if (!initialDelayApplied) {
+            initialDelayApplied = true;
+            applyJitter();
         }
         try {
             triggerIfWritable();
         } catch (Exception ex) {
             log.error("Exception thrown when running scheduled KafkaSQL snapshot creation", ex);
+        }
+    }
+
+    /**
+     * Sleeps for a random jitter period on the first invocation to reduce the chance that all replicas
+     * fire simultaneously after a coordinated restart.
+     */
+    private void applyJitter() {
+        try {
+            long jitterMs = ThreadLocalRandom.current().nextLong(MAX_INITIAL_DELAY_JITTER_SECONDS * 1000);
+            log.debug("Applying {}ms jitter before first scheduled snapshot check.", jitterMs);
+            Thread.sleep(jitterMs);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
         }
     }
 
@@ -99,16 +123,18 @@ public class KafkaSqlSnapshotScheduler {
      * Checks the {@code kafkasql-snapshots} topic for the most recent snapshot record. Returns {@code true}
      * if a snapshot was published within the configured interval, indicating that another replica (or a
      * manual trigger) already created one and this replica should skip.
+     * <p>
+     * Fails closed: if the check encounters an error, returns {@code true} to avoid a snapshot storm.
      */
     boolean recentSnapshotExists() {
         try {
-            long intervalMs = parseIntervalMs(configuration.getSnapshotEvery());
+            long intervalMs = parseDurationMs(configuration.getSnapshotEvery());
             long cutoff = System.currentTimeMillis() - intervalMs;
 
             TopicPartition partition = new TopicPartition(configuration.getSnapshotsTopic(), 0);
             List<TopicPartition> partitions = Collections.singletonList(partition);
 
-            try (KafkaConsumer<String, String> consumer = createSnapshotsConsumer()) {
+            try (KafkaConsumer<byte[], byte[]> consumer = createSnapshotsConsumer()) {
                 consumer.assign(partitions);
                 Map<TopicPartition, Long> endOffsets = consumer.endOffsets(partitions);
                 long endOffset = endOffsets.getOrDefault(partition, 0L);
@@ -118,8 +144,8 @@ public class KafkaSqlSnapshotScheduler {
                 }
 
                 consumer.seek(partition, endOffset - 1);
-                ConsumerRecords<String, String> records = consumer.poll(Duration.ofSeconds(5));
-                for (ConsumerRecord<String, String> record : records) {
+                ConsumerRecords<byte[], byte[]> records = consumer.poll(Duration.ofSeconds(5));
+                for (ConsumerRecord<byte[], byte[]> record : records) {
                     if (record.timestamp() > cutoff) {
                         log.debug("Found recent snapshot (age {}s, interval {}s).",
                                 (System.currentTimeMillis() - record.timestamp()) / 1000, intervalMs / 1000);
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
```diff
@@ -0,0 +1,152 @@
...
+                ConsumerRecords<String, String> records = consumer.poll(Duration.ofSeconds(5));
+                for (ConsumerRecord<String, String> record : records) {
+                    if (record.timestamp() > cutoff) {
+                        log.debug("Found recent snapshot (age {}s, interval {}s).",
+                                (System.currentTimeMillis() - record.timestamp()) / 1000, intervalMs / 1000);
+                        return true;
+                    }
+                }
+            }
+            return false;
```
paoloantinori, line 156: This catch swallows every exception and returns false, so any failure in the snapshots-topic check (transient Kafka error, parse error, connectivity) silently disables the multi-replica dedup guard. For a coordination feature that means all replicas proceed and fire a snapshot. Fail fast on deterministic errors (an invalid interval; see parseIntervalMs at line 144) and only tolerate transient topic errors, or surface a metric so the bypass is observable.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    private KafkaConsumer<String, String> createSnapshotsConsumer() {
+        return new KafkaConsumer<>(toProperties(configuration.getConsumerProperties()),
+                new StringDeserializer(), new StringDeserializer());
+    }
+
+    /**
+     * Parses the {@code apicurio.kafkasql.snapshot.every.seconds} value (e.g. "86400s") to milliseconds.
+     */
```
paoloantinori, line 175: parseIntervalMs only understands a trailing 's' or a bare number, but the same property drives `@Scheduled(every=...)` at line 69, which Quarkus parses as a full Duration ('24h', 'PT24H', '1d' all work there). For those inputs the scheduler fires on schedule but parseIntervalMs throws NumberFormatException and lands in the fail-open catch at line 130. Inject the property once as a typed Duration and derive both the cadence and the cutoff from that single parsed value.

paoloantinori (commented):
The opt-in wiring and storage guards are sound, but the multi-replica coordination this PR advertises is not actually guaranteed. The dedup check is a racy check-then-act with no leader election and it fails open on any exception, so every replica can still fire a snapshot per cycle; the most novel logic is also untested in its real branches.

- **Fails open, disabling the guard.** `recentSnapshotExists()` swallows every exception and returns false (KafkaSqlSnapshotScheduler.java:130-132), so any error in the snapshots-topic check silently disables the multi-replica dedup. For a feature whose purpose is coordination, that means all replicas proceed and you get a snapshot storm.
- **Two parsers for one property.** `parseIntervalMs` (KafkaSqlSnapshotScheduler.java:144) only handles a trailing 's' or a bare number, but the same value is fed to `@Scheduled(every=...)` (KafkaSqlSnapshotScheduler.java:69), which Quarkus parses as a full Duration. Inputs like '24h', 'PT24H', or '1d' fire the scheduler on schedule but throw NumberFormatException here, landing in the fail-open catch. Inject the property once as a typed `Duration` and derive both cadence and cutoff from it.
- **Coordination is a race, not a guarantee.** `recentSnapshotExists()` then `triggerSnapshotCreation()` (KafkaSqlSnapshotScheduler.java:90-95) is a check-then-act pair with no lock or leader election; `concurrentExecution=SKIP` only guards within one JVM. On a coordinated rollout all replicas start near the 60s initial delay, observe an empty or old snapshots topic, and each trigger a snapshot on the first tick. Either add a coordinator/leader mechanism, or reword the guarantee to 'best effort' and add jitter to the initial delay.
- **Tests pass for the wrong reason; the core branch is untested.** `newScheduler()` never sets `configuration`, so in `testTriggersSnapshotWhenReadyAndWritable` the code reaches `recentSnapshotExists()`, NPEs on `configuration.getSnapshotEvery()`, and falls into the fail-open catch before calling `triggerSnapshotCreation()` (KafkaSqlSnapshotSchedulerTest.java:37). The test is green via the exception path. Inject a mocked configuration and add the missing case (recent snapshot exists -> triggerSnapshotCreation() is NOT called); that branch of the core logic has no coverage.
- **UI count contradicts the visibility gate.** The new property is `@Dynamic(requires='apicurio.storage.kind=kafkasql')`, so it should be hidden unless storage is kafkasql. The UI/dev test backend typically runs on jpa or sql (no Kafka), where this property would not render and the count should stay 14, but the assertion was bumped to 15 (ui/tests/specs/settings.spec.ts:15). Either the `requires` gate is not hiding the property, or the assertion is wrong; confirm which storage the test profile uses and reconcile.
- **Easy win: deserialize cost.** `createSnapshotsConsumer()` uses `StringDeserializer` (KafkaSqlSnapshotScheduler.java:136-138), which fully decodes the (potentially multi-MB) snapshot payload as a String even though only `record.timestamp()` is read. Use `ByteArrayDeserializer` to avoid the decode.
- **Verify two small things.** The redundant `@Info` on `apicurio.storage.kind` (KafkaSqlSnapshotScheduler.java:50-51) may produce a duplicate entry in the generated config docs; drop it and reuse the canonical declaration. Also confirm the always-wired `@Scheduled` bean doesn't pull a kafkasql-only `KafkaSqlConfiguration` (KafkaSqlSnapshotScheduler.java:67) on sql, gitops, or k8sops variants and break startup; the kafkasql-profile box in the test plan is still unchecked.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotScheduler.java
```diff
@@ -0,0 +1,152 @@
...
+    private void triggerIfWritable() {
+        if (!storage.isReady()) {
+            log.debug("Skipping scheduled KafkaSQL snapshot creation because the storage is not ready.");
+            return;
+        }
+        if (storage.isReadOnly()) {
+            log.debug("Skipping scheduled KafkaSQL snapshot creation because the storage is in read-only mode.");
+            return;
+        }
+        if (recentSnapshotExists()) {
```
paoloantinori, line 114: recentSnapshotExists() and triggerSnapshotCreation() are a check-then-act pair with no lock or leader election, so in a multi-replica deployment every replica that observes an empty or old snapshots topic in the same window triggers a snapshot. concurrentExecution=SKIP only guards within a single JVM. On a coordinated rollout (all replicas starting near the 60s initial delay) this means N snapshots at the first tick. Either add a coordinator/leader mechanism, or reword the guarantee to 'best effort' and add jitter to the initial delay.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlSnapshotSchedulerTest.java
```diff
@@ -0,0 +1,113 @@
...
+        return scheduler;
+    }
+
+    @Test
+    void testTriggersSnapshotWhenReadyAndWritable() {
+        RegistryStorage storage = mock(RegistryStorage.class);
+        when(storage.isReady()).thenReturn(true);
+        when(storage.isReadOnly()).thenReturn(false);
+
+        newScheduler(storage).run();
```
paoloantinori: newScheduler() never sets the configuration field, so here the code reaches recentSnapshotExists(), NPEs on configuration.getSnapshotEvery(), and falls into the fail-open catch before triggerSnapshotCreation() is invoked. The test passes via the exception path, not the happy path. Inject a mocked configuration (and make recentSnapshotExists() overridable in tests), then add the missing case: recent snapshot exists -> triggerSnapshotCreation() is NOT called. That branch of the core logic is currently untested.

ui/tests/specs/settings.spec.ts
```diff
@@ -12,7 +12,7 @@ test.beforeEach(async ({ page }) => {
 test("Settings - Filter", async ({ page }) => {
     await expect(page.getByTestId("settings-search-widget").locator("input")).toBeEmpty();
     expect(page.getByTestId("config-groups")).toBeDefined();
-    await expect(page.getByTestId("config-groups").locator(".configuration-property")).toHaveCount(14);
+    await expect(page.getByTestId("config-groups").locator(".configuration-property")).toHaveCount(15);
```
paoloantinori, line 15: The new property is `@Dynamic(requires='apicurio.storage.kind=kafkasql')`, so it should render only when storage is kafkasql. The UI/dev test backend typically runs on jpa or sql (no Kafka), where the property would be hidden and the count should stay 14, not 15. If the count is genuinely 15, the requires gate is not hiding the property and the PR's UI-visibility claim doesn't hold. Confirm which storage the test profile uses and reconcile the assertion.

PR #9278, opened 2026-08-05
fix(ui): use applied filters for agents page paging and empty state

### LINKED ISSUE #8708: Agents page: pagination and empty state use unsubmitted filter input
Noticed this while clicking around the Agents page. If I type something into the search boxes but don't actually search, then click the next page, the text I typed gets applied as a filter anyway, even though I never pressed Enter or clicked Search.

Looks like `handlePageChange` and `handlePerPageChange` call `createFilters()`, which just reads whatever is currently in the input fields (`nameFilter` / `skillFilter` / `capabilityFilter`):

```js
const handlePageChange = (_event, page) => {
    const newPaging = { ...paging, page };
    setPaging(newPaging);
    search(createFilters(), newPaging);   // sends whatever is typed, not what was searched
};
```

Steps:
1. Register a few agents so there's more than one page
2. Type something into "Search by name" but don't press Enter or click Search
3. Click page 2
4. The list gets filtered by what you typed instead of just moving to the next page

The empty state has the same thing: `isFiltered` (line 229) reads the live input too, so the message switches to "No agents match your search criteria" before you've actually run a filtered search.

ExplorePage doesn't behave like this, its `onSetPage` uses the applied `criteria` instead of the toolbar input. So AgentsPage probably needs to keep the applied filters separately and use those for paging and the empty state.

Seen on main @ 5115d1136. Saw #8483 is currently touching this file, though it doesn't change these functions. Happy to send a small separate PR after that merges, or fold it in, whatever works.

shivv23:
## Description

Fixes #8708

On the Agents page, pagination and per-page changes used the live toolbar input instead of the last applied search criteria. Typing into a search box without submitting, then clicking a page number, silently filtered the results by the unsubmitted text. The empty state's "filtered" message also triggered on unsubmitted input.

## Changes

- Track applied filters separately from the live toolbar input
- Use the applied filters for page change, per-page change, and the empty state
- Capability select and search box clear actions update the applied filters

## Testing

- [x] Ran `npm run lint` in `ui/ui-app/` - passes (0 errors)
- [x] Typecheck passes for the modified file (only pre-existing missing SDK module errors remain)

## Checklist

- [x] My code follows the existing code style of this project
- [x] I have performed a self-review of my changes
- [x] My changes generate no new warnings

### DIFF
--- a/ui/ui-app/src/app/pages/agents/AgentsPage.tsx
+++ b/ui/ui-app/src/app/pages/agents/AgentsPage.tsx
@@ -68,6 +68,7 @@ export const AgentsPage: FunctionComponent<PageProperties> = () => {
     const [isSearching, setSearching] = useState<boolean>(false);
     const [results, setResults] = useState<AgentSearchResults>(EMPTY_RESULTS);
     const [paging, setPaging] = useState<Paging>(DEFAULT_PAGING);
+    const [appliedFilters, setAppliedFilters] = useState<AgentSearchFilters>({});
     const [nameFilter, setNameFilter] = useState<string>("");
     const [capabilityFilter, setCapabilityFilter] = useState<string>("");
     const [skillFilter, setSkillFilter] = useState<string>("");
@@ -91,6 +92,7 @@ export const AgentsPage: FunctionComponent<PageProperties> = () => {
     };
 
     const search = async (filters: AgentSearchFilters, paging: Paging): Promise<void> => {
+        setAppliedFilters(filters);
         setSearching(true);
         try {
             const results = await agentSvc.searchAgents(filters, paging);
@@ -102,14 +104,18 @@ export const AgentsPage: FunctionComponent<PageProperties> = () => {
         }
     };
 
+    const applyFilters = (newFilters: AgentSearchFilters): void => {
+        const newPaging = { ...DEFAULT_PAGING };
+        setPaging(newPaging);
+        search(newFilters, newPaging);
+    };
+
     const createLoaders = (): Promise<any> => {
         return search(createFilters(), paging);
     };
 
     const handleSearch = (): void => {
-        const newPaging = { ...DEFAULT_PAGING };
-        setPaging(newPaging);
-        search(createFilters(), newPaging);
+        applyFilters(createFilters());
     };
 
     const handleKeyPress = (event: React.KeyboardEvent): void => {
@@ -121,26 +127,23 @@ export const AgentsPage: FunctionComponent<PageProperties> = () => {
     const handlePageChange = (_event: any, page: number): void => {
         const newPaging = { ...paging, page };
         setPaging(newPaging);
-        search(createFilters(), newPaging);
+        search(appliedFilters, newPaging);
     };
 
     const handlePerPageChange = (_event: any, perPage: number): void => {
         const newPaging = { page: 1, pageSize: perPage };
         setPaging(newPaging);
-        search(createFilters(), newPaging);
+        search(appliedFilters, newPaging);
     };
 
     const handleCapabilitySelect = (_event: React.MouseEvent | undefined, value: string | number | undefined): void => {
         const selectedValue = value as string || "";
         setCapabilityFilter(selectedValue);
         setCapabilitySelectOpen(false);
-        const filters: AgentSearchFilters = {
-            name: nameFilter || undefined,
-            capability: selectedValue || undefined,
-            skill: skillFilter || undefined
-        };
-        search(filters, { ...DEFAULT_PAGING });
-        setPaging({ ...DEFAULT_PAGING });
+        applyFilters({
+            ...appliedFilters,
+            capability: selectedValue || undefined
+        });
     };
 
     const navigateToAgent = (agent: AgentSearchResult): void => {
@@ -249,7 +252,7 @@ export const AgentsPage: FunctionComponent<PageProperties> = () => {
     };
 
     const renderEmptyState = (): React.ReactElement => {
-        const isFiltered = !!(nameFilter || capabilityFilter || skillFilter);
+        const isFiltered = !!(appliedFilters.name || appliedFilters.capability || appliedFilters.skill);
         return (
             <EmptyState
                 headingLevel="h4"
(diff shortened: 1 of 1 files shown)


hussainjamal760:
@shivv23 shivv23 Thanks for addressing the filter application bug! I was reviewing your changes and I found that there are a couple of architectural issues with the current implementation:

## 1.  Bug in onClear Handlers
onClear for Name (and Skill) constructs search criteria using draft input values (capabilityFilter, skillFilter) from local state rather than appliedFilters.

Scenario: If a user types text into Skill search box (without pressing Enter or clicking Search) and then clears Name, the draft unsubmitted Skill value gets erroneously applied to the search query.
## 2. Code Duplication & State Syncing
The setAppliedFilters, setPaging, and search(...) state update pattern is duplicated across 5+ separate inline handlers.

## Suggested Fix
Abstract filter updates into a single helper method like applyFilters(newFilters: AgentSearchFilters) that derives directly from appliedFilters for field clears, avoiding state leakage from unsubmitted draft input fields:

tsx
const applyFilters = (newFilters: AgentSearchFilters): void => {
    const newPaging = { ...DEFAULT_PAGING };
    setAppliedFilters(newFilters);
    setPaging(newPaging);
    search(newFilters, newPaging);
};
// Example for onClear:
onClear={() => {
    setNameFilter("");
    applyFilters({
        ...appliedFilters,
        name: undefined
    });
}}
Additionally, adding a unit test covering these filter state transitions would ensure this behavior stays stable.

Note: I'm an open-source contributor helping maintainers review queued PRs more quickly. Final approval and merge decisions remain with the project maintainers!

ui/ui-app/src/app/pages/agents/AgentsPage.tsx
```diff
@@ -139,6 +142,7 @@ export const AgentsPage: FunctionComponent<PageProperties> = () => {
             capability: selectedValue || undefined,
             skill: skillFilter || undefined
         };
+        setAppliedFilters(filters);
```
paoloantinori: `handleCapabilitySelect` now seeds `appliedFilters` here, which is correct. The same `setAppliedFilters` call is needed wherever a skill is selected: if the skill-select handler calls `search()` without it, `appliedFilters` keeps the previous skill value and the next page or per-page change queries with a stale skill, reintroducing the bug this PR fixes on the skill axis. Confirm the skill-selection path updates `appliedFilters` too.

ui/ui-app/src/app/pages/agents/AgentsPage.tsx
```diff
@@ -68,6 +68,7 @@ export const AgentsPage: FunctionComponent<PageProperties> = () => {
     const [isSearching, setSearching] = useState<boolean>(false);
     const [results, setResults] = useState<AgentSearchResults>(EMPTY_RESULTS);
     const [paging, setPaging] = useState<Paging>(DEFAULT_PAGING);
+    const [appliedFilters, setAppliedFilters] = useState<AgentSearchFilters>({});
```
paoloantinori, line 71: `appliedFilters` starts as `{}`, so every `search()` call must keep it in sync with what was actually queried, or page and per-page changes will read stale filters. If this component runs an initial search on mount with non-empty filters (deep link or URL params), that path does not set `appliedFilters`, and the first page change would query with `{}`, dropping the initial filters. Confirm the mount search seeds `appliedFilters`, or that the initial query is always unfiltered.

paoloantinori (commented):
The fix is sound and well-scoped. Separating applied filters from the live toolbar input is the right model, and page change, per-page change, and the empty state now consistently read `appliedFilters` (AgentsPage.tsx:111, 127, 133, 256). The risk is completeness rather than the core change: the invariant holds only if every `search()` call also updates `appliedFilters`, and the diff covers capability and the search/clear actions but not skill selection.

- AgentsPage.tsx:145: `handleCapabilitySelect` correctly seeds `appliedFilters`, but the diff shows no matching update for skill selection. If selecting a skill calls `search()` without `setAppliedFilters`, then `appliedFilters` keeps the old skill value and the next page or per-page change (which reads `appliedFilters` at 127 and 133) queries with a stale skill filter. That is the same stale-filter bug this PR fixes, just on the skill axis. Confirm the skill-select path updates `appliedFilters` too.
- AgentsPage.tsx:71: `appliedFilters` initializes to `{}`. Any search that runs before the first explicit user search must seed it, or the first page change queries with `{}`. If the component searches on mount with non-empty filters (deep link, URL params), that path does not set `appliedFilters`, so this PR would change paging to drop those initial filters. Confirm the mount search seeds `appliedFilters`, or that the initial query is always unfiltered.
- AgentsPage.tsx:284: this empty-state handler rebuilds `{ name, capability, skill }` inline, duplicating `createFilters()` already used in `handleSearch` at 111. Reusing it avoids drift if a filter field is added later. The handler near 303 intentionally omits `skill`, so leave that one as-is.

ui/ui-app/src/app/pages/agents/AgentsPage.tsx
shivv23: fixed

ui/ui-app/src/app/pages/agents/AgentsPage.tsx
shivv23: fixed

shivv23:
@paoloantinori Fixed every issue. Thanks!

shivv23:
@hussainjamal760 Fixed, thanks for the review.

paoloantinori:
Thanks @shivv23 — the fix looks clean. The `appliedFilters` separation reads well: page change and per-page change now use the last applied criteria, the clear handlers correctly reset only their own field, and the empty state tracks `appliedFilters` consistently. One small thing I'd note: `createLoaders` still calls `search(createFilters(), paging)` rather than going through `applyFilters`, so on initial mount the `appliedFilters` state is set inside `search` and the first page stays in sync. That's fine as long as `search` always calls `setAppliedFilters` first (which it does now), but worth keeping in mind if that ordering ever changes. I have no further concerns.

EricWittmann:
Thanks for the contribution @shivv23! Nice fix, the appliedFilters pattern is clean.

PR #9281, opened 2026-08-05
fix: use Locale.ROOT in toLowerCase() for label search and ES structure indexing

### LINKED ISSUE #9280: Label search returns no results when JVM locale differs from ROOT (missing Locale in toLowerCase calls)
### Bug Description

I noticed that label-based search silently returns empty results in certain JVM locale configurations. After digging into it, the root cause is that `SqlSearchRepository` normalizes the user-supplied label key/value using `.toLowerCase()` without an explicit `Locale` argument, while the storage layer always writes labels using `Locale.ROOT`.

This means on a JVM running under a locale like Turkish (`tr_TR`) — where `"i".toUpperCase()` produces `"İ"` instead of `"I"` — the search query builds a different string than what's stored in the database, so you get zero results even when matching artifacts exist.

The same pattern appears in `ElasticsearchDocumentBuilder` and `ElasticsearchSearchService` where structure index values and query terms are lowercased without a Locale, which would cause structure-based search to also silently fail.

---

### Affected Files

**`app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java`**
- Line 127: `label.getKey().toLowerCase()` (artifact label search)
- Line 131: `label.getValue().toLowerCase()` (artifact label search)
- Line 318: `label.getKey().toLowerCase()` (version label search)
- Line 322: `label.getValue().toLowerCase()` (version label search)

**`app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchDocumentBuilder.java`**
- Line 187: `artifactType.toLowerCase()`
- Line 193: `element.kind().toLowerCase()`
- Line 194: `element.name().toLowerCase()`

**`app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java`**
- Line 306: `value.toLowerCase().trim()`

---

### Steps to Reproduce

1. Start the JVM with `-Duser.language=tr -Duser.country=TR`
2. Create an artifact and attach a label with key `"signal"`
3. Search for artifacts by label key `"signal"`
4. Observe that zero results are returned, even though the artifact exists

---

### Expected Behavior

Label search should return matching artifacts regardless of the JVM's default locale, since labels are stored with consistent `Locale.ROOT` normalization.

---

### Actual Behavior

Zero results returned. The search query normalizes the filter value with the system locale, producing a different string than what was stored.

---

### Suggested Fix

Replace all bare `.toLowerCase()` calls in the affected files with `.toLowerCase(Locale.ROOT)`:

```diff
// SqlSearchRepository.java
- String labelKey = label.getKey().toLowerCase();
+ String labelKey = label.getKey().toLowerCase(Locale.ROOT);

- String labelValue = label.getValue().toLowerCase();
+ String labelValue = label.getValue().toLowerCase(Locale.ROOT);

// ElasticsearchDocumentBuilder.java
- String artifactTypeLower = artifactType.toLowerCase();
+ String artifactTypeLower = artifactType.toLowerCase(Locale.ROOT);

- String kindLower = element.kind().toLowerCase();
+ String kindLower = element.kind().toLowerCase(Locale.ROOT);

- String nameLower = element.name().toLowerCase();
+ String nameLower = element.name().toLowerCase(Locale.ROOT);

// ElasticsearchSearchService.java
- String lowered = value.toLowerCase().trim();
+ String lowered = value.toLowerCase(Locale.ROOT).trim();

### PR DESCRIPTION
**Pull request by DarkBird10020**
Closes #9280

## What's the problem?

While exploring label search behavior, I noticed that `SqlSearchRepository` 
normalizes the user-supplied label key/value filter using plain `.toLowerCase()`, 
which picks up the JVM's default `Locale`. But every place in the codebase that 
*stores* labels (e.g. `SqlArtifactRepository`, `SqlVersionRepository`, 
`SqlGroupRepository`) uses `.toLowerCase(Locale.ROOT)` consistently.

So on a JVM running under a locale with non-trivial case mappings — Turkish 
(`tr_TR`) being the classic example, where `"i".toUpperCase()` → `"İ"` instead 
of `"I"` — the stored label key and the search query key end up as different 
strings, and the query returns zero results even when the artifact is there.

The same mismatch existed in the Elasticsearch path: `ElasticsearchDocumentBuilder` 
lowercases `artifactType`, `kind`, and `name` when building the `structure` index 
field, and `ElasticsearchSearchService` lowercases the incoming query value — but 
neither specified a Locale, so on a non-ROOT JVM the indexed value and the query 
value could diverge.

## What changed?

Added `Locale.ROOT` to every bare `.toLowerCase()` call in the three affected files:

- `SqlSearchRepository` — label key/value normalization in both the artifact 
  search and version search `labels` case blocks (4 calls)
- `ElasticsearchDocumentBuilder` — `artifactType`, `kind`, `name` lowercasing 
  inside `indexStructuredElements` (3 calls)
- `ElasticsearchSearchService` — the `value.toLowerCase()` in `buildStructureQuery` 
  (1 call)

Also added the missing `import java.util.Locale;` to the two files that didn't 
have it yet.

## Why only these files?

I specifically only touched the lines that were inconsistent with the rest of the 
codebase. All the storage-layer write paths already use `Locale.ROOT` correctly — 
this PR just brings the search/query paths into the same standard.

## Testing

- Compiled the `app` module cleanly against the existing build (`mvn compile -pl app`)
- No logic changes, no behavior change on standard Latin-locale JVMs — purely a 
  correctness fix for non-ROOT locale deployments

## Checklist

- [x] Follows existing code style (`Locale.ROOT` on all case conversions per `CLAUDE.md`)
- [x] No unrelated changes (only 3 files, 8 lines changed)
- [x] Conventional commit message referencing the issue

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java
@@ -303,7 +303,7 @@ public class ElasticsearchSearchService {
             return Query.of(q -> q.matchAll(m -> m));
         }
 
-        String lowered = value.toLowerCase().trim();
+        String lowered = value.toLowerCase(Locale.ROOT).trim();
         String[] parts = lowered.split(":", -1);
 
         if (parts.length == 3) {
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlArtifactRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlArtifactRepository.java
@@ -199,7 +199,7 @@ public class SqlArtifactRepository {
                     labels.forEach((k, v) -> {
                         handle.createUpdate(sqlStatements.insertArtifactLabel())
                                 .bind(0, normalizeGroupId(groupId)).bind(1, artifactId)
-                                .bind(2, limitStr(k.toLowerCase(), MAX_LABEL_KEY_LENGTH))
+                                .bind(2, limitStr(asLowerCase(k), MAX_LABEL_KEY_LENGTH))
                                 .bind(3, limitStr(asLowerCase(v), MAX_LABEL_VALUE_LENGTH)).execute();
                     });
                 }
@@ -289,8 +289,8 @@ public class SqlArtifactRepository {
                 labels.forEach((k, v) -> {
                     handle.createUpdate(sqlStatements.insertArtifactLabel())
                             .bind(0, normalizeGroupId(groupId)).bind(1, artifactId)
-                            .bind(2, limitStr(k.toLowerCase(), MAX_LABEL_KEY_LENGTH))
-                            .bind(3, limitStr(v.toLowerCase(), MAX_LABEL_VALUE_LENGTH)).execute();
+                            .bind(2, limitStr(asLowerCase(k), MAX_LABEL_KEY_LENGTH))
+                            .bind(3, limitStr(asLowerCase(v), MAX_LABEL_VALUE_LENGTH)).execute();
                 });
             }
         } catch (Exception ex) {
@@ -340,8 +340,8 @@ public class SqlArtifactRepository {
                         handle.createUpdate(sqlStatements.insertArtifactLabel())
                                 .bind(0, normalizeGroupId(entity.groupId))
                                 .bind(1, entity.artifactId)
-                                .bind(2, k.toLowerCase())
-                                .bind(3, v == null ? null : v.toLowerCase())
+                                .bind(2, asLowerCase(k))
+                                .bind(3, asLowerCase(v))
                                 .execute();
                     });
                 }
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
@@ -22,6 +22,7 @@ import org.slf4j.Logger;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
@@ -124,11 +125,11 @@ public class SqlSearchRepository {
                         break;
                     case labels:
                         Pair<String, String> label = filter.getLabelFilterValue();
-                        String labelKey = label.getKey().toLowerCase();
+                        String labelKey = label.getKey().toLowerCase(Locale.ROOT);
                         where.append("EXISTS(SELECT l.* FROM artifact_labels l WHERE ");
                         buildWildcardClause(where, "l.labelKey", labelKey, filter.isNot(), binders);
                         if (label.getValue() != null) {
-                            String labelValue = label.getValue().toLowerCase();
+                            String labelValue = label.getValue().toLowerCase(Locale.ROOT);
                             where.append(" AND ");
                             buildWildcardClause(where, "l.labelValue", labelValue, filter.isNot(),
                                     binders);
@@ -315,11 +316,11 @@ public class SqlSearchRepository {
                         break;
                     case labels:
                         Pair<String, String> label = filter.getLabelFilterValue();
-                        String labelKey = label.getKey().toLowerCase();
+                        String labelKey = label.getKey().toLowerCase(Locale.ROOT);
                         where.append("EXISTS(SELECT l.* FROM version_labels l WHERE ");
                         buildWildcardClause(where, "l.labelKey", labelKey, filter.isNot(), binders);
                         if (label.getValue() != null) {
-                            String labelValue = label.getValue().toLowerCase();
+                            String labelValue = label.getValue().toLowerCase(Locale.ROOT);
                             where.append(" AND ");
                             buildWildcardClause(where, "l.labelValue", labelValue, filter.isNot(),
                                     binders);
--- a/app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
@@ -1406,13 +1406,13 @@ public abstract class AbstractRegistryStorageTest extends AbstractResourceTestBa
     @Test
     public void testSearchArtifactsByLabelWithTurkishLocale() throws Exception {
         Locale savedLocale = Locale.getDefault();
+        String artifactId = "testLabelSearchTurkishLocale-1";
         try {
             // Switch to Turkish locale, where 'I'.toLowerCase() = 'ı' (U+0131), not 'i'.
             // If any toLowerCase() call is missing Locale.ROOT, the stored key and the
             // search filter key will differ, and the artifact will not be found.
             Locale.setDefault(new Locale("tr", "TR"));
 
-            String artifactId = "testLabelSearchTurkishLocale-1";
             // Label key contains uppercase I, which is the problematic character under tr_TR.
             Map<String, String> labels = Collections.singletonMap("INSTABILITY", "HIGH");
             EditableArtifactMetaDataDto metaData = new EditableArtifactMetaDataDto(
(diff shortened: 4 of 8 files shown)

### CODE REVIEW

File: app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java [hunk 1]
```diff
@@ -124,11 +125,11 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
                         break;
                     case labels:
                         Pair<String, String> label = filter.getLabelFilterValue();
```
paoloantinori, line 127: This and the version-search twin (line 318) are fixed, but a bare `toLowerCase()`/`toUpperCase()` in any other switch arm in this file would still mismatch storage on a non-ROOT locale. Worth grepping the whole search layer for case conversions without an explicit `Locale` before merging, since the diff only touches the `labels` arms.

app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java
```diff
@@ -303,7 +303,7 @@ private Query buildStructureQuery(String value) {
             return Query.of(q -> q.matchAll(m -> m));
         }
 
-        String lowered = value.toLowerCase().trim();
+        String lowered = value.toLowerCase(Locale.ROOT).trim();
```
paoloantinori, line 306: Consider a regression test that sets the default locale to `tr_TR` and then indexes and queries a structured element whose name/kind/artifactType contains an uppercase `I`. The mismatch is invisible on a Latin-locale dev JVM, so this is the only way to prove the fix and prevent a future bare `toLowerCase()` from silently reintroducing it.

paoloantinori (commented):
Solid, minimal fix. Adding `Locale.ROOT` to every bare `toLowerCase()` in the three files makes query-time normalization match the storage layer, so a Turkish-locale JVM no longer lowercases `I` to dotless `ı` (U+0131) and silently misses stored labels and ES structure entries. It is a no-op on Latin-locale JVMs and cannot introduce a regression. Two things to confirm before merge.

- **Completeness.** The diff only covers the `labels` switch arms in `SqlSearchRepository.java` (lines 127, 131, 318, 322) and the three ES sites. Worth a repo-wide grep for `toLowerCase()`/`toUpperCase()` without an explicit `Locale` to confirm no sibling path was missed: other `case` arms in `SqlSearchRepository` (e.g. type/name/content filters), sibling query builders in `ElasticsearchSearchService` besides `buildStructureQuery` (line 306), and other fields indexed in `ElasticsearchDocumentBuilder` besides artifactType/kind/name (lines 188, 190, 191). Correctness also depends on the storage write paths truly using `Locale.ROOT`, which the PR asserts but the diff does not show.
- **Test coverage.** The bug is invisible on a standard dev JVM, so "it compiles" does not prove it is fixed. A parameterized test that flips the default locale to `tr_TR` and asserts that a label key/value and a structured-element name containing `I` round-trips through artifact/version search and the ES structure query would lock the fix in and catch the next bare `toLowerCase()`.

Labreo:
Just letting it be know that #9220 also addresses missing Locale.ROOT in StringUtil and is related.But both issues are valid.

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
```diff
@@ -1396,4 +1397,71 @@ private static String generateString(int size) {
...
+                    GROUP_ID, artifactId, ArtifactType.OPENAPI, metaData, null,
+                    ContentWrapperDto.builder()
+                            .contentType(ContentTypes.APPLICATION_JSON)
+                            .content(ContentHandle.create(OPENAPI_CONTENT)).build(),
+                    null, Collections.emptyList(), false, false, null).getValue();
+
+            // Search using the same label key. Locale.ROOT normalization on both the
+            // write path (storage) and read path (query) must agree, otherwise the
+            // stored key ("instability-XXXXX" via ROOT) and the query key
+            // ("ınstability-XXXXX" via Turkish) diverge and the result list is empty.
```
paoloantinori, line 1438: This test does not reproduce the locale bug. It writes the label and queries it under one Turkish locale, so the stored key and the query key match whether both sides use the default locale or both use Locale.ROOT; revert the fix and the test still passes. To make it a real regression test, assert the exact normalized value (the stored label key equals "instability-<suffix>", not "ınstabılıty-<suffix>"), or write under Locale.ROOT and query under Turkish so the asymmetry is actually exercised. It also only covers the SQL path; the Elasticsearch change is untested.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlArtifactRepository.java
```diff
@@ -199,7 +199,7 @@ public void updateArtifactMetaData(String groupId, String artifactId,
                     labels.forEach((k, v) -> {
                         handle.createUpdate(sqlStatements.insertArtifactLabel())
                                 .bind(0, normalizeGroupId(groupId)).bind(1, artifactId)
-                                .bind(2, limitStr(k.toLowerCase(), MAX_LABEL_KEY_LENGTH))
+                                .bind(2, limitStr(asLowerCase(k), MAX_LABEL_KEY_LENGTH))
```
paoloantinori, line 202: Please confirm createArtifact's label write path is covered by these changes. This PR edits updateArtifactMetaData, insertArtifactRaw, and importArtifact, but if createArtifact / createArtifactVersion insert labels through a different path (for example AbstractSqlRegistryStorage), those call sites still use bare toLowerCase() and the primary create flow is not fixed. If that is the case, add those sites and have the test assert the exact normalized stored value so the gap is caught.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -124,11 +125,11 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
...
-                        String labelKey = label.getKey().toLowerCase();
+                        String labelKey = label.getKey().toLowerCase(Locale.ROOT);
```
paoloantinori, line 128: Two conventions for the same operation ship in this PR: asLowerCase() in SqlArtifactRepository/SqlGroupRepository/SqlVersionRepository, and inline toLowerCase(Locale.ROOT) here and in the two Elasticsearch files. asLowerCase is the existing null-safe helper from StringUtil; prefer it here too for consistency. As a bonus it avoids an NPE if a label key is ever null, since the inline form dereferences label.getKey() directly while asLowerCase tolerates null.

app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java
paoloantinori, line 306: No test covers the Elasticsearch structure path changed here and in ElasticsearchDocumentBuilder.indexStructuredElements (the three toLowerCase(Locale.ROOT) calls on artifactType/kind/name). Add a test that indexes a structured element with an uppercase name/type and queries it under a Turkish locale, asserting the match fails on bare toLowerCase and passes with Locale.ROOT. Otherwise two of the six changed files are unguarded by this PR's new test.

paoloantinori (commented):
The fix is correct in direction, but I would hold merge until the new test actually reproduces the bug, the Elasticsearch path is covered, and the write-path scope is confirmed. asLowerCase is the right helper and is null-safe; the concerns below are about coverage and test value.

- The new locale test does not reproduce the regression it targets. It writes and queries the label under a single Turkish locale, so the stored key and the query key match whether both sides use the default locale or both use Locale.ROOT; the test passes with the fix reverted, so it guards nothing (AbstractRegistryStorageTest.java:1438). Assert the exact normalized stored value instead.
- Please confirm createArtifact's label write path is among the methods changed here. The PR edits SqlArtifactRepository.updateArtifactMetaData/insertArtifactRaw/importArtifact; if createArtifact inserts labels through a different path (for example AbstractSqlRegistryStorage), that path still uses bare toLowerCase() and the primary create flow is not fixed (SqlArtifactRepository.java:202).
- Two conventions for one operation ship in this PR: asLowerCase() in three repositories versus inline toLowerCase(Locale.ROOT) in SqlSearchRepository and the two Elasticsearch files. asLowerCase is the existing null-safe helper from StringUtil; prefer it everywhere, which also sidesteps an NPE on a null label key on the read side (SqlSearchRepository.java:128).
- No test covers the Elasticsearch structure path changed here and in ElasticsearchDocumentBuilder. Add a test that indexes an element with an uppercase name and queries it under a Turkish locale (ElasticsearchSearchService.java:306).
- The PR description is stale: it claims "3 files, 8 lines, only adds Locale.ROOT", but the diff changes 6 source files and migrates three repositories to asLowerCase(). Update it so the scope and the helper migration are accurately reflected.
- Minor: structure index entries produced on a non-ROOT-locale JVM (tr/az) before this change will not match post-fix queries until reindexed; worth a release note for those deployments.

EricWittmann:
Thanks @DarkBird10020 :+1:

PR #9282, opened 2026-08-05
test(operator): use MEDIUM_DURATION for ingress creation waits

### LINKED ISSUE #9276: Flaky test: IngressITTest.ingressAnnotations times out on OCP 4.16
## Description

`IngressITTest.ingressAnnotations` fails intermittently on OCP 4.16 with a 30-second
awaitility timeout. The operator fails to create the Ingress resource within the timeout
window. The same test passes reliably on OCP 4.22 (Latest).

## Environment

- **Registry version:** 3.3.1
- **Operator image:** `quay.io/apicurio/apicurio-registry-3-operator:3.3.1`
- **OCP version:** 4.16.55 (Fyre-provisioned EBC cluster)
- **Test run:** CSD pipeline `Controller smoke+feature (OCP 4.16)`

## Test Results

- **OCP Latest (4.22.2):** 39/39 passed — `ingressAnnotations` passed
- **OCP 4.16 (4.16.55):** 38/39 passed — `ingressAnnotations` failed

## Error

```
[ERROR] io.apicurio.registry.operator.it.IngressITTest.ingressAnnotations
        -- Time elapsed: 36.49 s <<< ERROR!

org.awaitility.core.ConditionTimeoutException:
  Assertion condition defined as a Lambda expression in
  io.apicurio.registry.operator.it.IngressITTest null within 30 seconds.
    at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
    at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:790)
    at io.apicurio.registry.operator.it.IngressITTest.ingressAnnotations(IngressITTest.java:48)

Caused by: java.lang.IllegalStateException: Kubernetes resource does not exist.
    at io.apicurio.registry.operator.utils.K8sCell.get(K8sCell.java:63)
    at io.apicurio.registry.operator.it.IngressITTest.lambda$ingressAnnotations$3(IngressITTest.java:49)
```

The test waited 30 seconds for the Ingress resource to be created by the operator, but
the resource never appeared.

### 1. K8s 409 Conflict retries during the preceding test

The `ingressClassName` test (which passed) triggered repeated 409 Conflict errors on the
custom resource, indicating the operator and test were racing to update the same CR:

```
Retrying update of ResourceID{name='ingress-class-name', namespace='test-8c55a46'}
because of a Kubernetes client exception:
  Operation cannot be fulfilled on apicurioregistries3.registry.apicur.io "ingress-class-name":
  the object has been modified; please apply your changes to the latest version and try again.
```

This happened at least twice (15:27:51 and 15:28:03), suggesting reconciliation contention
that may have left the operator in a state where it was slow to process the next CR
(`ingress-annotations`).

### 2. Operator startup probe failure

The operator pod itself experienced a startup probe failure earlier in the test namespace
lifecycle:

```
Startup probe failed: Get "http://10.254.20.68:8080/q/health/started":
  dial tcp 10.254.20.68:8080: connect: connection refused
```

This suggests the operator was slower to become ready on this OCP 4.16 cluster.

### 3. Cluster-wide image pull pressure

The OCP 4.16 cluster had `pull QPS exceeded` errors across several OpenShift system
components (cloud-credential-operator, controller-manager-operator, insights-operator,
kube-scheduler-operator). While these are unrelated to the Apicurio operator itself, they
indicate the cluster was under resource pressure which may have contributed to slower
reconciliation.

## Analysis

The `ingressAnnotations` test creates an `ApicurioRegistry3` CR and waits up to 30 seconds
for the operator to create an Ingress resource (`IngressITTest.java:48-49`). On this
OCP 4.16 cluster, the operator reconciliation was slower — likely due to a combination of
409 Conflict retries from the preceding test and general cluster pressure — causing the
Ingress to not be created within the timeout window.

The `ingressClassName` test in the same class passed, so the operator itself is functional —
this is a timing-sensitive test that doesn't account for slower clusters.

## Possible Fixes

1. **Increase the awaitility timeout** for `ingressAnnotations` (e.g., 60-90 seconds
   instead of 30) to tolerate slower clusters
2. **Add explicit CR cleanup verification** between tests to ensure the operator has fully
   reconciled before starting the next test
3. **Add retry logic** in the test for the case where the Ingress doesn't exist yet,
   distinguishing between "not yet created" and "will never be created"

harshitsaini17:
## What

`IngressITTest.ingressAnnotations` waits for the operator to create the app and UI
Ingress resources. Both waits used `SHORT_DURATION` (30s); this changes them to
`MEDIUM_DURATION` (120s), and switches `get()` to `getOptional()` in those two
assertions.

Only the two initial creation waits change. Every subsequent wait in the test — the
annotation update/removal assertions — still uses `SHORT_DURATION`, since those
operate on an already-reconciled resource.

## Why

Creating an Ingress requires a full reconciliation of a freshly created CR, which
takes considerably longer than a subsequent update to an existing resource. The rest
of the operator suite already reflects this: resource-creation waits use
`MEDIUM_DURATION` in `ITBase.checkDeploymentExists()`, `PodDisruptionBudgetITTest`,
`LeaderElectionITTest`, `MultiNamespaceSmokeITTest`, `HorizontalPodAutoscalerITTest`
and `GitOpsITTest`. The two ingress creation waits were the outliers at 30s, which is
only ~10 polls against `POLL_INTERVAL_DURATION` (3s).

Secondly, the assertion could not fail as written:

```java
assertThat(appIngress.get()).isNotNull();
```

`K8sCell.get()` throws `IllegalStateException` when the resource is absent
(`K8sCell.java:63`), so `isNotNull()` was unreachable and `ignoreExceptions()`
swallowed the exception until the timeout — which is why #9276 surfaces as a bare
`IllegalStateException` with no assertion message. `getOptional()` makes a timeout
report the resource as missing.

## Alternative considered

The timeouts are already tunable, so the OCP 4.16 job could instead pass
`-Dtest.operator.timeout.short=90` (there's precedent — `operator.yaml:286` passes
`-Dtest.operator.timeout.long=600`). I didn't go that route because it raises *every*
`SHORT_DURATION` wait, including the negative checks that use
`await().during(ofSeconds(10)).atMost(SHORT_DURATION)` to assert a resource stays
absent, which would slow the suite for no benefit. Happy to switch if you'd prefer the
pipeline-config approach. Both knobs remain available either way.

## Testing

- `./mvnw checkstyle:check -pl operator/controller -Dfull -Pintegration-tests -Pexamples` — passes
- `./mvnw test-compile -pl operator/controller -am -DskipTests -Dfull -Pintegration-tests -Pexamples` — BUILD SUCCESS

**I have not run `IngressITTest` itself** — it needs a Kubernetes/OpenShift cluster I
don't have available. So I can't demonstrate the flake is gone; the reasoning is based
on the timeout being the reported failure mode and on aligning with the suite's
existing convention for creation waits. The `getOptional()` change is behaviour-neutral
apart from the error message.

One caveat worth flagging from the issue's own logs: the operator pod hit a startup
probe failure on that run. `ITBase` waits for pod readiness once at class setup, so if
the operator restarted mid-run its informer caches would need to resync, and a longer
timeout would only mask that. If the pod restart count from that run is available it
would confirm whether this is sufficient or whether a readiness re-check between tests
is also wanted.

Fixes #9276

### DIFF
--- a/operator/controller/src/test/java/io/apicurio/registry/operator/it/IngressITTest.java
+++ b/operator/controller/src/test/java/io/apicurio/registry/operator/it/IngressITTest.java
@@ -45,12 +45,17 @@ public class IngressITTest extends ITBase {
         final var appIngress = k8sCell(client, () -> client.network().v1().ingresses().withName(primary.getCached().getMetadata().getName() + "-app-ingress").get());
         final var uiIngress = k8sCell(client, () -> client.network().v1().ingresses().withName(primary.getCached().getMetadata().getName() + "-ui-ingress").get());
 
-        await().atMost(SHORT_DURATION).ignoreExceptions().untilAsserted(() -> {
-            assertThat(appIngress.get()).isNotNull();
+        // Creating the Ingress requires a full reconciliation of the freshly created CR, which can
+        // take considerably longer than a subsequent update to an already existing resource. Use
+        // MEDIUM_DURATION here, consistent with the other resource-creation waits such as
+        // ITBase.checkDeploymentExists(). getOptional() is used instead of get() so that a timeout
+        // reports the resource as missing rather than surfacing an IllegalStateException.
+        await().atMost(MEDIUM_DURATION).ignoreExceptions().untilAsserted(() -> {
+            assertThat(appIngress.getOptional()).isPresent();
         });
 
-        await().atMost(SHORT_DURATION).ignoreExceptions().untilAsserted(() -> {
-            assertThat(uiIngress.get()).isNotNull();
+        await().atMost(MEDIUM_DURATION).ignoreExceptions().untilAsserted(() -> {
+            assertThat(uiIngress.getOptional()).isPresent();
         });
 
         await().atMost(SHORT_DURATION).ignoreExceptions().untilAsserted(() -> {


harshitsaini17:
Hi @EricWittmann, just a gentle nudge on this one, it's been open for a bit and I wanted to make sure it didn't slip through. Happy to rebase or adjust anything if needed. Thanks!

hussainjamal760:
Hi @harshitsaini17 ,
Thank you for the detailed breakdown.

You've addressed the root cause of the flake on slower clusters perfectly:
1. Aligning the initial creation waits to `MEDIUM_DURATION` is the right design choice and matches conventions in other IT tests.
2. Swapping `get()` with `getOptional()` inside `untilAsserted()` fixes the masking behavior of `ignoreExceptions()`. Now, if it times out, the test suite will report a clean assertion failure (`Expected: isPresent, Actual: empty`) rather than a generic `IllegalStateException`.
3. Good call on avoiding the global timeout bump which would have needlessly dragged down our negative assertions duration.

All checkstyle and compile checks are green. 

Note: I am an open-source contributor helping maintainers review queued PRs quickly. Final approval and merge decisions remain with the project maintainers!

harshitsaini17:
Thanks, I looked into the suppressed suggestion about `ignoreExceptions()` and I
don't think it should be removed here.

`K8sCell.getOptional()` still throws for reasons unrelated to the resource being
absent. It calls `initialReader.get()`, and the supplier for these two cells calls
`primary.getCached()`, which throws `IllegalStateException` when the primary CR
isn't cached yet. The underlying `client.network().v1().ingresses()...get()` is
also a live API call that can throw `KubernetesClientException` transiently, which
is precisely the slow-cluster condition #9276 is about. Dropping
`ignoreExceptions()` would risk reintroducing the flake this PR is meant to fix.

For consistency, `ignoreExceptions()` is used 119 times across the operator IT
suite, so these two blocks would become the exception rather than the rule.

The switch from `get()` to `getOptional()` still earns its place, it makes a
timeout report "resource missing" instead of surfacing a bare
`IllegalStateException`, and it removes an `isNotNull()` assertion that was
unreachable. I've left `ignoreExceptions()` in place.

@jsenko this is ready for review whenever you have a moment, CI is green and it's
a one-file test-only change.

jsenko (approved):
Approving. Thanks @harshitsaini17 — and apologies for the wait, this has been open since August and the delay was ours.

This is a small change with an unusually good description, and I checked every factual claim in it. They all hold.

## Verified

**The unreachable assertion is real, and it is the reported symptom.**

```java
// K8sCell.java:63
public T get() {
    return getOptional().orElseThrow(() -> new IllegalStateException("Kubernetes resource does not exist."));
}
```

So `assertThat(appIngress.get()).isNotNull()` could never fail on its own terms — `get()` throws first, `ignoreExceptions()` swallows it, and the wait runs to timeout. That is exactly what #9276 shows:

```
org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a Lambda expression ... null within 30 seconds
Caused by: java.lang.IllegalStateException: Kubernetes resource does not exist.
    at io.apicurio.registry.operator.utils.K8sCell.get(K8sCell.java:63)
```

The stack trace names the precise line you identified. Switching to `getOptional()` / `isPresent()` means a timeout now reports a missing resource instead of a bare `IllegalStateException`, which is a genuine diagnostic improvement independent of the timeout change.

**The arithmetic checks out.**

```java
// ITBase.java:77-81
public static final Duration POLL_INTERVAL_DURATION = ofSeconds(3);
public static final Duration SHORT_DURATION  = ofSeconds(Integer.getInteger("test.operator.timeout.short", 30));
public static final Duration MEDIUM_DURATION = ofSeconds(Integer.getInteger("test.operator.timeout.medium", 120));
```

30s against a 3s poll interval is 10 attempts, as you say. The failure in #9276 took 36.49s — so it was not marginally over, it was never going to make it on that cluster.

**The precedent is real.** `ITBase.checkDeploymentExists()` uses `MEDIUM_DURATION` at line 179, and so do `GitOpsITTest`, `HorizontalPodAutoscalerITTest`, `LeaderElectionITTest`, `MultiNamespaceSmokeITTest` and `PodDisruptionBudgetITTest`. The two ingress creation waits were indeed the outliers.

**The scope is exactly as described.** Only the two creation waits change; the other twelve `SHORT_DURATION` waits in the file — the annotation update and removal assertions, which operate on an already-reconciled resource — are untouched.

**Your "alternative considered" is sound, and the reasoning is the right one.** Raising `test.operator.timeout.short` in the pipeline would also inflate the negative checks at `ITBase.java:194`, `213` and `232`:

```java
await().during(ofSeconds(10)).atMost(SHORT_DURATION).ignoreExceptions().untilAsserted(check::run);
```

Those assert that a resource *stays* absent, so a longer `atMost` buys nothing and costs wall-clock time on every run. Fixing the specific waits that are wrong is better than globally loosening a constant that is correct elsewhere. (The `operator.yaml:286` precedent you cite for `-Dtest.operator.timeout.long=600` is real, so the option was genuinely available — you just had a good reason not to take it.)

**Build:** `test-compile` clean, `checkstyle:check` 0 violations, DCO signed, single commit.

I have not run `IngressITTest` itself — it needs a live cluster — but the change is a constant swap plus an assertion fix, and the OCP 4.16 pipeline that reported #9276 is the real verification.

## One thought for later, not for this PR

The `assertThat(cell.get()).isNotNull()` pattern is unreachable wherever it appears, not just here. If it is used elsewhere in the operator suite, those assertions are silently no-ops too, and their timeouts will report the same unhelpful `IllegalStateException`. Might be worth a sweep — as a separate PR, since this one is properly scoped to the reported bug.

Thanks again, and sorry for the long wait. The description did most of my review for me, which is not something I get to say often.

jsenko:
Merged, thank you!

PR #9283, opened 2026-08-05
test(cli): split mislabeled GroupCommandTest failure cases

### LINKED ISSUE #9274: Mislabeled test testGroupGetCommandFails actually covers group create failure cases
## Description

GroupCommandTest.testGroupGetCommandFails is named as if it tests the group get command's failure scenarios, but the test body only calls executeAndAssertFailure with the group create subcommand:

group create --output-type foo
group create -p -1
group create -s 0
These are group create validation failures (unknown output type, invalid page, invalid size), not group get failures. The method name should be updated to testGroupCreateCommandFails to reflect what it actually tests, or the test should be split into create and get failure cases as intended.

GroupCommandTest.java
Lines: 113–121.

harshitsaini17:
Fixes #9274

## What

`testGroupGetCommandFails` in `GroupCommandTest` covered three unrelated things, none of them `group get`:

```java
executeAndAssertFailure("group", "create", "-p", "-1");
executeAndAssertFailure("group", "create", "-s", "0");
executeAndAssertFailure("group", "create", "--output-type", "json");
```

This splits it into `testGroupListCommandFails` (pagination) and a `testGroupGetCommandFails` that actually exercises `group get`.

## Why

Three separate problems in nine lines:

1. **Wrong command under test.** `-p/--page` and `-s/--size` come from `PaginationMixin`, which `GroupCommand` (the list command) uses and `GroupGetCommand` does not. Pagination validation belongs on the list command.

2. **The pagination cases passed for the wrong reason.** `group create -p -1` fails because picocli does not recognize `-p` on that command — not because `PositiveIntegerValidator` rejected the value. The test would still pass if the validator were deleted. These now run against `group` and assert on the error message, so they fail if the validator stops working:
   ```java
   executeAndAssertFailure("group", "-p", "-1");
   assertThat(err.toString()).contains("must be greater than 0");
   ```

3. **A duplicated assertion.** The third line was byte-identical to one already in `testGroupCreateCommandFails` (line 79). Dropped rather than moved.

## Note on the issue's suggested fix

The issue proposes renaming the method to `testGroupCreateCommandFails`. That name is already taken by an existing test in the same class, so a straight rename would not compile. Splitting by the command each case actually targets keeps both names meaningful.

## Testing

`GroupCommandTest` run locally against the containerized registry — 9 tests, 0 failures:

```
[INFO] Tests run: 9, Failures: 0, Errors: 0, Skipped: 0 -- in io.apicurio.registry.cli.GroupCommandTest
```

Also verified: `./mvnw checkstyle:check -pl cli` and `./mvnw test-compile -pl cli -am` both pass. Test-only change, no production code touched.

### DIFF
--- a/cli/src/test/java/io/apicurio/registry/cli/GroupCommandTest.java
+++ b/cli/src/test/java/io/apicurio/registry/cli/GroupCommandTest.java
@@ -133,10 +133,24 @@ public class GroupCommandTest extends AbstractCLITest {
 
     @Test
     public void testGroupGetCommandFails() {
+        // Assert on the error message so these cases cannot pass merely because the
+        // command failed for some other reason.
+
         // Required groupId parameter is missing
+        err.getBuffer().setLength(0);
         executeAndAssertFailure("group", "get");
+        assertThat(err.toString())
+                .as(withCliOutput("Missing groupId parameter must be reported"))
+                .contains("Missing required parameter")
+                .contains("groupId");
+
         // Unknown output type (rejected while parsing, so the group need not exist)
+        err.getBuffer().setLength(0);
         executeAndAssertFailure("group", "get", "some-group", "--output-type", "foo");
+        assertThat(err.toString())
+                .as(withCliOutput("Invalid --output-type value must be reported"))
+                .contains("--output-type")
+                .contains("foo");
     }
 
     @Test
--- a/cli/src/test/java/io/apicurio/registry/cli/GroupCommandTest.java
+++ b/cli/src/test/java/io/apicurio/registry/cli/GroupCommandTest.java
@@ -111,13 +111,32 @@ public class GroupCommandTest extends AbstractCLITest {
     }
 
     @Test
-    public void testGroupGetCommandFails() {
-        // Unknown output type
-        executeAndAssertFailure("group", "create", "--output-type", "foo");
+    public void testGroupListCommandFails() {
+        // Pagination options belong to the `group` list command, not to `group create`.
+        // Assert on the error message so these cases cannot pass merely because the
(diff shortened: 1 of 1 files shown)


hussainjamal760:
Hi @harshitsaini17 ,

Thank you for this exceptionally detailed and well-thought-out PR description! 

This is exactly how a high-quality contribution should look. You correctly identified that:
- The tests were executing against the wrong command (`group create` instead of the `group` list command).
- The assertions were false-positives passing because of unrecognized option parsing, rather than exercising `PositiveIntegerValidator`.
- The suggested fix in the issue would have caused a compilation error due to a naming conflict.

Your decision to split the tests and assert explicitly on the `err` output stream is the correct engineering approach. 

I've verified checkstyle and test executions on our end. Everything is green.

LGTM! 

Note: I am an open-source contributor helping maintainers review queued PRs quickly. Final approval and merge decisions remain with the project maintainers!

vandanayadav7 (dismissed):
Good fix — the old test passed for the wrong reason (picocli rejected the flag, not the validator). The split is correct and asserting on the error message is a nice improvement. LGTM.

@harshitsaini17 Could you mark this as ready for review (out of draft)? I can't do it from my side.

harshitsaini17:
Thanks @vandanayadav7! Already done, it's out of draft now, so it should show as
ready on your side.

harshitsaini17:
Thanks for the review @vandanayadav7 and apologies, my push dismissed your
approval automatically, so
this needs another look when you have a moment.

vandanayadav7 (approved):
Updated commit looks good — error message assertions added for both cases. LGTM.

PR #9284, opened 2026-08-06
fix(app): declare default for apicurio.storage.kind and align @Info metadata

Abhishek6122008:
apicurio.storage.kind is declared at seven injection points. Five of them omitted defaultValue, and the @Info descriptions disagreed across sites (three distinct strings, one site had no @Info at all).

The config doc generator resolves one declaration per property name, and it picked a site without a defaultValue, so the published configuration reference listed the property with an empty default even though the effective default is sql (application.properties). Which description was published was likewise incidental.

Give every declaration defaultValue = "sql" and the same @Info description, so the generated reference is correct regardless of which site is selected, and regenerate ref-registry-all-configs.adoc.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/core/System.java
+++ b/app/src/main/java/io/apicurio/registry/core/System.java
@@ -28,6 +28,7 @@ import java.util.Date;
 import java.util.Locale;
 import java.util.Map;
 
+import static io.apicurio.common.apps.config.ConfigPropertyCategory.CATEGORY_STORAGE;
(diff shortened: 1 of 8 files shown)


Abhishek6122008:
C:/Program Files/Git/retry

EricWittmann:
Thanks for the contribution, @Abhishek6122008! Merged.

Abhishek6122008:
@EricWittmann I’ll make sure to keep the PRs coming! Thanks for staying connected with this PR

PR #9286, opened 2026-08-06
Fix ZipStream resource leaks in v3 import/export endpoints

omkarrr2533:
Both the import and export endpoints create their zip stream outside the try block and only close it on the success path.

In `AdminResourceImpl.importData`, `IoUtil.unpackToDisk` throws an IOException when the zip entry count, entry size or total size limits are exceeded. When that happens the `ZipInputStream` is never closed, so it leaks along with its Inflater on exactly the malformed-input path the limits exist to guard.

`DataExporter.exportData` has the same problem: if the storage layer fails part way through the export, `zip.close()` is skipped.

Both are now wrapped in try-with-resources. `ProtobufExporter` in the same package already does it this way, so this just brings the other two in line.

I also added a unit test for the export case that mocks a storage failure and checks the zip stream was still closed. It fails without the fix.


app/src/test/java/io/apicurio/registry/rest/v3/impl/shared/DataExporterTest.java
```diff
@@ -0,0 +1,63 @@
+package io.apicurio.registry.rest.v3.impl.shared;
```
paoloantinori, line 17: The added file starts at `package` with no Apache 2.0 license header. Apicurio enforces headers via the license check (license-maven-plugin), so this will fail CI on the new file. Add the same header used by the other test classes in this package.

app/src/main/java/io/apicurio/registry/rest/v3/impl/AdminResourceImpl.java
```diff
@@ -332,16 +332,13 @@ public void importData(Boolean xRegistryPreserveGlobalId, Boolean xRegistryPrese
         }
 
         // The input should be a ZIP file
-        final ZipInputStream zip = new ZipInputStream(data, StandardCharsets.UTF_8);
-
         // Unpack the ZIP file to the local file system (temp)
         Path tempDirectory = null;
-        try {
+        try (ZipInputStream zip = new ZipInputStream(data, StandardCharsets.UTF_8)) {
```
paoloantinori, line 337: This is the higher-risk of the two fixes. The leak fires when `IoUtil.unpackToDisk` throws because a zip limit (`zipMaxEntrySize`, `zipMaxTotalSize`, `zipMaxEntryCount`) was exceeded, which happens on attacker-controlled input, so closing the stream here has a security payoff beyond hygiene. The export path got a failure-case test; this import path did not. Please add a test (or confirm an existing integration test covers it) that drives `unpackToDisk` to throw on an over-limit zip and asserts the underlying stream is closed and a `BadRequestException` is returned.

app/src/test/java/io/apicurio/registry/rest/v3/impl/shared/DataExporterTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+        DataExporter exporter = new DataExporter();
+        exporter.log = LoggerFactory.getLogger(DataExporter.class);
+        exporter.storage = storage;
+
+        StreamingOutput stream = (StreamingOutput) exporter.exportData(null).getEntity();
+        CloseTrackingOutputStream output = new CloseTrackingOutputStream();
+
+        // When: the export fails while streaming
+        assertThrows(IOException.class, () -> stream.write(output));
```
paoloantinori: Confirm `BadRequestException` extends `IOException`. The storage failure is caught by `catch (Exception e)` and rethrown as `BadRequestException`, so `assertThrows(IOException.class, ...)` is only meaningful if that type is an `IOException`. If it is a plain `RuntimeException`, this assertion fails on type regardless of the fix and the test would pass for the wrong reason. Asserting the concrete type would make the intent explicit.

paoloantinori (commented):
One CI-blocking issue: the new test file is missing its Apache license header, so the license check will fail. The try-with-resources conversions themselves are correct and behavior-preserving, and the export failure path is covered by a focused new test, but the import path (the more security-relevant fix) has no test of its own.

- `DataExporterTest.java:1`: the added file starts at `package` with no license header; the license check will fail CI. Add the header used by neighboring test classes.
- `AdminResourceImpl.java:337` and `DataExporter.java:44`: the fixes are sound. Exception handling is preserved (the try-with-resources `catch` still catches `close()` failures), and dropping the explicit `zip.flush()` is safe because `ZipOutputStream.close()` finishes and flushes.
- `DataExporterTest.java:29`: only the export failure path is tested. The matching import fix has no test, and it is the higher-risk leak because it fires when `IoUtil.unpackToDisk` rejects an over-limit zip, i.e. on untrusted input.
- `DataExporterTest.java:42`: `assertThrows(IOException.class, ...)` is only valid if `BadRequestException` extends `IOException`; one grep to confirm, or assert the concrete type.

omkarrr2533:
Thanks for the quick review, this was useful.

Added `AdminResourceImplTest` for the import path. It builds a two-entry zip in memory, sets `zipMaxEntryCount` to 1 so `unpackToDisk` throws on the second entry, and asserts both that a `BadRequestException` comes back and that the uploaded stream was closed. I checked it fails without the fix (`expected: <true> but was: <false>`), so it is measuring the leak and not something else. `unpackToDisk` doesn't close the stream itself, which is what makes the assertion meaningful here.

On the license header: I added it to both test files. Worth flagging though that `ImportExportTest` and `ApiDocsFilterTest` in these packages don't have one, and the License Compliance check passed before I added it, so I wasn't sure which way you'd want it. Happy to drop it if consistency with the surrounding files is preferred.

On `assertThrows(IOException.class, ...)` in `DataExporterTest` I think the `BadRequestException` conversion is the import path. `DataExporter` catches the storage failure with `catch (Exception e)` and rethrows `new IOException(e)`, so the type does hold there. To make the intent explicit rather than incidental, I've also asserted that the cause is the `RegistryStorageException` the mock throws, so it can't pass for the wrong reason. Let me know if you'd still rather see something different asserted.

omkarrr2533:
Rebased on main. #9208 landed the same try-with-resources change in DataExporter, so I've dropped my version of it and kept upstream's. What's left here is the import path fix in AdminResourceImpl, which is still unfixed on main, plus tests for both paths there was no DataExporterTest on main, so that one now covers the behaviour #9208 introduced.

EricWittmann:
Thanks for the contribution, @omkarrr2533! Clean fix with solid test coverage — nicely done.

PR #9287, opened 2026-08-06
ci: split the app-other unit test shard to fix heap exhaustion

### LINKED ISSUE #9265: ci: app-other unit test shard fails with Java heap space OOM
## Flaky CI failure

The `Unit Tests / Unit Tests (app-other)` shard fails with a `Java heap space` OOM in the Surefire
forked JVM. No test assertions fail — the entire forked process crashes before completing.

## Error

```
[ERROR] There was an error in the forked process
[ERROR] Java heap space
[ERROR] org.apache.maven.surefire.booter.SurefireBooterForkException: There was an error in the forked process
[ERROR] Java heap space
```

## Observed in

- PR #9251 — `Unit Tests / Unit Tests (app-other)` shard
- [CI run 30978024640, job 92317162198](https://github.com/Apicurio/apicurio-registry/actions/runs/30978024640/job/92317162198)

The shard runs with `MAVEN_OPTS: -Xmx4g`. The OOM suggests the test suite in this shard is
approaching (or exceeding) that limit under GHA runner memory pressure.

## Possible mitigations

- Increase `-Xmx` for the `app-other` shard
- Split the `app-other` shard into smaller partitions to reduce peak memory
- Investigate which test classes in the shard are the heaviest memory consumers

malladinagarjuna2:
## Summary

Splits the `app-other` unit test shard, which fails with a `Java heap space` OOM
in the Surefire forked JVM.

Fixes #9265

## Root Cause

Surefire runs a shard in one forked JVM (`forkCount=1`, `reuseForks=true` by
default), and every distinct `@TestProfile` makes Quarkus tear down and rebuild
the whole application inside it. `app-other` had 42 distinct profiles across 147
classes — four times the next-densest shard — which exhausted its 4 GB heap.

Note the heap was not where the issue suggested: `MAVEN_OPTS` sizes the Maven
process, while the test JVM's heap came from a hardcoded `-Xmx4g` in
`app/pom.xml`.

## Changes

- **`verify-unit-tests.yaml`** — carved the profile-dense clusters out of
  `app-other` into `app-auth` (auth, rbac, limits), `app-transport` (tls,
  headers, cors) and `app-metrics` (metrics, search). `app-other` stays an
  exclusion filter so new packages still land somewhere. Worst-shard profile
  count drops 42 → 15.
- **`app/pom.xml`** — forked-JVM heap is now `-Xmx${surefire.maxHeap}`, default
  `4g` (unchanged), overridable per shard.
- **`.github/scripts/verify-test-shards.py`** (new) — implements Surefire's
  `-Dtest=` matching and fails if any test class is claimed by zero or more than
  one shard.
- **`workflows/README.md`** — updated shard table; documented that boundaries
  are set by profile density.

## Test plan

- [x] Partition exact: 253 classes, 0 orphans, 0 duplicates
- [x] Coverage unchanged: new shards match the same 147 classes as the old filter
- [x] Checker catches deliberate orphan and duplicate sabotages (exit 1)
- [x] OOM reproduced before the fix: `BUILD FAILURE`, 2543 s, 20 `Java heap
      space` lines, 390 app tests completed
- [x] Fixed on the same host at the same `-Xmx4g`: `BUILD SUCCESS`, 175 s, 0 heap
      errors, 669 app tests
- [x] New shards pass with their CI filters (`app-auth` 192, `app-transport` 78,
      `app-metrics` 64)
- [x] Heap property verified on the forked command line (`maxHeap=6g` → `-Xmx6g`;
      `MAVEN_OPTS=-Xmx512m` → still `-Xmx4g`)
- [x] CI confirmation — a green `app-other` across a few real runs

**Storage variants:** N/A — CI configuration only, no product code changed.

### DIFF
--- a/.github/scripts/verify-test-shards.py
+++ b/.github/scripts/verify-test-shards.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python3
+"""Verify the app-module unit-test shards in verify-unit-tests.yaml partition the
+test set exactly once.
+
+Every surefire-eligible test class in app/ must be claimed by exactly one app-*
+shard. A class claimed by none never runs in CI and fails silently, because the
+workflow passes -Dsurefire.failIfNoSpecifiedTests=false. A class claimed by two
+wastes a shard's budget.
+
+Implements surefire's -Dtest= semantics: comma-separated fully-qualified-name
+patterns, '!' prefix excludes, '**' matches across package separators, '*'
+matches within one segment. A filter of exclusions only includes everything not
+excluded.
+
+Usage:  python3 .github/scripts/verify-test-shards.py
+Exits non-zero when the partition is broken.
+"""
+import re
+import sys
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parents[2]
+APP_TESTS = REPO / "app/src/test/java"
+WORKFLOW = REPO / ".github/workflows/verify-unit-tests.yaml"
+
+# The 'non-app' shard selects by Maven -pl, not by -Dtest, so it is not part of
+# the app-module partition.
+NON_APP_SHARD = "non-app"
+
+
+def is_surefire_name(stem):
+    """Match surefire's default includes."""
+    return (
+        stem.startswith("Test")
+        or stem.endswith("Test")
+        or stem.endswith("Tests")
+        or stem.endswith("TestCase")
+    )
+
+
+def enumerate_classes():
+    """Fully-qualified names of app/ test classes surefire would actually run."""
+    found = []
+    for path in APP_TESTS.rglob("*.java"):
+        stem = path.stem
+        if not is_surefire_name(stem):
+            continue
+        source = path.read_text(encoding="utf-8", errors="replace")
+        # Abstract classes are bases for other tests; surefire never runs them.
+        if re.search(r"\babstract\s+class\s+" + re.escape(stem) + r"\b", source):
+            continue
+        found.append(str(path.relative_to(APP_TESTS)).replace("/", ".")[:-len(".java")])
+    return sorted(found)
+
+
+def pattern_to_regex(pattern):
+    regex, i = "", 0
+    while i < len(pattern):
+        if pattern.startswith("**", i):
+            regex += ".*"
+            i += 2
+        elif pattern[i] == "*":
+            regex += "[^.]*"
+            i += 1
+        else:
+            regex += re.escape(pattern[i])
+            i += 1
+    return re.compile("^" + regex + "$")
+
+
+def matches(fqn, test_filter):
+    if not test_filter:
+        return True
+    patterns = [p.strip() for p in test_filter.split(",") if p.strip()]
+    includes = [pattern_to_regex(p) for p in patterns if not p.startswith("!")]
+    excludes = [pattern_to_regex(p[1:]) for p in patterns if p.startswith("!")]
+    if includes and not any(r.match(fqn) for r in includes):
+        return False
+    return not any(r.match(fqn) for r in excludes)
+
+
+def parse_shards():
+    """Read (name, test-filter) pairs straight from the workflow matrix."""
+    shards, name = [], None
+    for line in WORKFLOW.read_text(encoding="utf-8").splitlines():
+        matched = re.match(r"\s*- name:\s*(\S+)", line)
+        if matched:
+            name = matched.group(1)
+            continue
+        matched = re.match(r'\s*test-filter:\s*"(.*)"\s*$', line)
+        if matched and name:
+            value = matched.group(1)
+            if value.startswith("-Dtest="):
+                value = value[len("-Dtest="):]
+            shards.append((name, value))
+            name = None
+    return [s for s in shards if s[0] != NON_APP_SHARD]
+
+
+def main():
+    classes = enumerate_classes()
+    shards = parse_shards()
+    if not shards:
+        print("ERROR: no shards parsed from", WORKFLOW)
+        return 2
+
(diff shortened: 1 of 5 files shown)


malladinagarjuna2:
@paoloantinori  could you please review

paoloantinori:
Looked at the shard split. The root-cause analysis is solid. Identifying that the 4 GB heap lives in `app/pom.xml`'s surefire `argLine`, not `MAVEN_OPTS`, is exactly the kind of distinction that makes the fix land rather than paper over the symptom.

One thing I want to flag: the PR body describes `.github/scripts/verify-test-shards.py` (new) as the guard that implements Surefire's `-Dtest=` matching and fails if any test class is claimed by zero or more than one shard. But the changed-files list only contains `README.md`, `verify-unit-tests.yaml`, and `app/pom.xml`. If that script was dropped or isn't included in this PR, the "Partition exact: 253 classes, 0 orphans, 0 duplicates" check in the test plan isn't enforced in CI; it's a one-time manual verification. Worth adding it back so a future shard-boundary edit can't silently orphan a package.

Minor: the body says 147 classes while the inline comment in `verify-unit-tests.yaml` says 148 classes. Small, but if the script is meant to enforce the partition it should agree on the number.

malladinagarjuna2:
Yes @paoloantinori I have dropped the scripts thinking that it's not much useful wrt pr will add it to

paoloantinori:
Thanks. The three changes that remain are the ones I care about — the shard split, the `-Xmx${surefire.maxHeap}` parameterization in `app/pom.xml`, and the README update. The parameterized heap is the part worth keeping most: per-shard tuning without a recompile.

One thing to nail down before this merges: dropping `verify-test-shards.py` removes the only automated check that the partition has no orphans (a class matched by zero shards → silently dropped from CI) or duplicates (matched by two → wasted CI time). The PR description still says "0 orphans, 0 duplicates", but nothing enforces that invariant when new packages land later. If the script is going somewhere else (your comment looked cut off at "will add it to"), good — just want to confirm where, since the README now states the invariant as a contract a contributor has to uphold by hand.

Also heads-up: the PR is `behind` main, so it'll need a rebase before merge.

malladinagarjuna2:
>(your comment looked cut off at "will add it to") 

 hey @paoloantinori  by this i meant,  the script was in the branch originally. I removed it because I'd written it to verify my own shard split and assumed it was throwaway tooling that didn't belong in the PR. Your comment made me realise that was the wrong call,.

paoloantinori:
Thanks for re-adding it — `verify-test-shards.py` is back in the diff and the README now documents that the Lint job enforces the partition. That closes the loop for me: the split, the `-Xmx${surefire.maxHeap}` parameterization, the README, and the guard that prevents future orphans/duplicates are all in.

jsenko (commented):
## Review

The shard split fixes the OOM and the groupings are semantically coherent. Root cause analysis is solid, the parameterized heap (`-Xmx${surefire.maxHeap}`) is a useful addition, and `verify-test-shards.py` guards against orphans.

### should-fix: Use JUnit `@Tag` groups instead of `-Dtest=` package patterns

The current approach duplicates test-to-shard mapping in the workflow YAML as package patterns, then reverse-engineers surefire's matching in `verify-test-shards.py` to verify it. This is fragile — a new package or a moved class can silently end up in the wrong shard.

The operator module already has the right model: semantic `@Tag` constants in `operator/Tags.java` (`AUTH`, `KAFKA`, `OLM`, `FEATURE`, etc.), with surefire selecting via `<groups>`. The app module has `ApicurioTestTags` but only defines `SLOW` and `DOCKER` (unused).

**Proposed approach:**

1. Expand `ApicurioTestTags` (or create a parallel `AppTestGroups` class) with semantic group constants:

```java
public static final String AUTH = "auth";
public static final String TRANSPORT = "transport";
public static final String METRICS = "metrics";
public static final String REST = "rest";
public static final String CCOMPAT = "ccompat";
public static final String STORAGE_SQL = "storage-sql";
public static final String STORAGE_KAFKASQL = "storage-kafkasql";
// etc.
```

2. Tag each test class with its group: `@Tag(ApicurioTestTags.AUTH)`. Use semantic names that describe what the test exercises — not shard names. The CI shard is derived from the tag, not the other way around.

3. In the workflow, each shard selects by group instead of package pattern:

```yaml
- name: app-auth
  test-filter: "-Dgroups=auth"
```

4. `verify-test-shards.py` simplifies to: scan for surefire-eligible test classes and check each has at least one recognized `@Tag`. No surefire pattern matching, no exclusion-list maintenance.

**Use general semantic tags, not shard-specific ones.** `AUTH` describes the test's purpose for developers. If a shard later needs splitting, you add a finer-grained tag (e.g., `AUTH_KEYCLOAK`, `AUTH_BASIC`) without touching existing tests. The operator module already follows this pattern.

**Migration scope:** This is ~250 test classes to tag. I'll create a tracking issue for it regardless.

@malladinagarjuna2 Would you be interested in taking on the migration? If so, you can either fold it into this PR or do it as a separate follow-up — your preference. The current pattern-based split fixes the OOM either way, so this PR is not blocked on it.

jsenko:
Tracking issue for the @Tag migration: #9302

EricWittmann:
I've merged this PR because it will have immediate positive impact.  We can tackle the @Tag migration in a followup PR.  :+1:

malladinagarjuna2:
hey @jsenko  i am sorry for the late 
>@malladinagarjuna2 Would you be interested in taking on the migration? If so, you can either fold it into this PR or do it as a separate follow-up — your preference. The current pattern-based split fixes the OOM either way, so this PR is not blocked on it.

yeah would have  been happy to take the  followups 
but other people have taken the issue, so can't take it now.

PR #9293, opened 2026-08-06
feat(ui): implement dark theme support

### LINKED ISSUE #9285: feat(ui): Add dark mode support to Apicurio Registry UI
Currently, the Apicurio Registry UI only supports a light theme. This can cause eye strain when using the registry for long periods of time or in low-light environments, especially when other developer tools and IDEs are set to dark mode.

### Proposed Solution
Introduce a theme toggle in the UI header that allows users to switch between Light theme (default) and Dark theme.
To implement this:
1. Toggle a CSS class (.pf-v6-theme-dark on the root <html> element) to switch stylesheets.
2. Use theme-aware CSS variables for card backgrounds, dashboard stat values, tab containers, and inputs to make sure text remains readable and layouts are consistent in both themes.
3. Remember the user's preference using browser local storage.

### Additional Context
The feature has been styled and verified to look consistent with PatternFly 6 guidelines. Text contrasts in the dashboard, overview tabs, rules pages, and modal inputs have all been adjusted to ensure high legibility.

JituRewar:
Closes #9285

What is the goal?
Apicurio Registry currently only supports a light theme. This PR implements full Dark Theme support for the user interface, allowing users to switch between light and dark modes.

Proposed Changes:
1. Theme Switcher Service: Created a new ThemeContext and useThemeService to toggle the dark theme by applying the .pf-v6-theme-dark class to the HTML document.
2. Root Theme Styling: Defined local light (#f0f0f0) and dark (#0f1011) canvas background and card background variables in App.css.
3. UI Components Fixes: Updated custom styling to ensure headers, rule lists, search lists, drafts, and settings correctly adapt their colors and texts dynamically.
4. Dashboard Layout: Applied inline styling overrides to the Registry Dashboard content section to prevent the white default layout from showing in dark mode.

Light - 
Screenshot From 2026-08-06 14-57-08

Dark-
Screenshot From 2026-08-06 14-57-19

How to test:
1. Open the UI.
2. Toggle the theme switcher icon in the header toolbar to change from Light to Dark mode.
3. Verify that all sections (Dashboard, Explore, Search, Rules, Settings) display correctly with matching dark backgrounds.


ui/ui-app/src/app/App.css
```diff
@@ -1,6 +1,20 @@
...
+}
+
+.pf-v6-theme-dark {
+  --registry-card-bg: #151515;
+  --registry-page-canvas-bg: #0f1011;
+}
+
+.pf-v6-theme-dark input:disabled,
+.pf-v6-theme-dark textarea:disabled,
+.pf-v6-theme-dark [disabled] {
```
paoloantinori: The bare `.pf-v6-theme-dark [disabled]` selector matches every disabled element in the DOM (buttons, menu items, tabs, switches, links), not just form fields. With `!important` on background, color, and border, it overrides PatternFly's own disabled treatment across the whole app in dark mode and will misstyle disabled buttons and other components. The scoped `input:disabled` and `textarea:disabled` rules above already cover inputs; drop the bare `[disabled]` (add `select:disabled` if dropdowns need it too).

ui/ui-app/src/services/useThemeService.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const storedTheme = localStorageService.getConfigProperty("theme", "light") as ThemeType;
+    const [theme, setTheme] = useState<ThemeType>(storedTheme);
+
+    const toggleTheme = () => {
+        const nextTheme = theme === "light" ? "dark" : "light";
+        setTheme(nextTheme);
+        localStorageService.setConfigProperty("theme", nextTheme);
+    };
+
+    useEffect(() => {
```
paoloantinori: Applying the `pf-v6-theme-dark` class in `useEffect` runs after the browser's first paint, so a user with a stored dark preference sees a flash of the light theme on every page load. Use `useLayoutEffect` so the class is set before paint, or, for a fully flash-free load, add an inline script to index.html that reads the stored theme and sets the class on `<html>` before React mounts.

ui/ui-app/src/app/components/header/AppHeader.tsx
```diff
@@ -6,18 +6,22 @@ import { AppNavigation, useAppNavigation } from "@services/useAppNavigation.ts";
...
     const appNavigation: AppNavigation = useAppNavigation();
     const config: ConfigService = useConfigService();
+    const { isDark } = useThemeService();
 
     if (config.features().showMasthead !== undefined && !config.features().showMasthead) {
         return <></>;
     }
 
-    const logoSrc: string = `${config.uiContextPath() || "/"}apicurio_registry_logo_default.svg`;
+    const logoSrc: string = `${config.uiContextPath() || "/"}${isDark ? "apicurio_registry_logo_reverse.svg" : "apicurio_registry_logo_default.svg"}`;
```
paoloantinori: Confirm `apicurio_registry_logo_reverse.svg` is present in the served static assets; if it's missing, dark mode renders a broken logo. The same `logoSrc` ternary now lives here and in `AppHeaderToolbar.tsx`, so consider extracting a shared helper (e.g. `useLogoSrc`) so the two can't drift.

ui/ui-app/src/app/pages/dashboard/DashboardPage.tsx
```diff
@@ -160,7 +160,7 @@ export const DashboardPage: FunctionComponent<PageProperties> = () => {
                         </FlexItem>
                     </Flex>
                 </PageSection>
-                <PageSection hasBodyWrapper={false} isFilled={true} className="ps_dashboard-content">
+                <PageSection hasBodyWrapper={false} isFilled={true} className="ps_dashboard-content" style={{ backgroundColor: "var(--registry-page-canvas-bg)" }}>
```
paoloantinori: This inline `backgroundColor` duplicates `.ps_dashboard-content` in DashboardPage.css, which already sets the same value and wins via `!important`. The inline style is effectively dead; remove it so the background lives in one place.

ui/ui-app/src/app/components/header/AppHeaderToolbar.tsx
```diff
@@ -48,6 +50,14 @@ export const AppHeaderToolbar: FunctionComponent<AppHeaderToolbarProps> = () =>
             <Toolbar id="app-header-toolbar" isFullHeight={true}>
                 <ToolbarContent>
                     <ToolbarGroup align={{ default: "alignEnd" }}>
+                        <ToolbarItem>
+                            <Button
+                                aria-label="Toggle theme"
```
paoloantinori: `aria-label="Toggle theme"` doesn't convey current state, so a screen reader user can't tell whether dark mode is active. Use a stateful label like `aria-label={isDark ? "Switch to light theme" : "Switch to dark theme"}`, or add `aria-pressed={isDark}`.

paoloantinori (commented):
Mostly on the right track, with two issues to fix before merge: a bare `[disabled]` selector in App.css restyles every disabled element in the app (not just form fields), and the dark class is applied after the first paint so users with a stored dark preference see a flash of the light theme on load. The broader move from hardcoded whites and grays to PatternFly tokens, and re-importing App.css after patternfly.css so the overrides win, are correct.

- `App.css:14`: `.pf-v6-theme-dark [disabled]` matches every disabled element (buttons, menu items, tabs, switches), not just form fields. The `!important` override fights PatternFly's own disabled styling app-wide in dark mode. Drop the bare `[disabled]` and keep the scoped `input:disabled`/`textarea:disabled` rules.
- `useThemeService.tsx:26`: the `pf-v6-theme-dark` class is added in `useEffect` (post-paint), so users with a stored dark preference see a light flash on every load. Prefer `useLayoutEffect`, or set the class in an inline index.html script before React mounts.
- `AppHeader.tsx:25` and `AppHeaderToolbar.tsx`: confirm `apicurio_registry_logo_reverse.svg` is shipped (otherwise dark mode shows a broken logo). The same `logoSrc` ternary is duplicated in both files, worth sharing.
- `DashboardPage.tsx:163`: the inline `backgroundColor` duplicates `.ps_dashboard-content` in CSS (which wins via `!important`), so the inline style is dead.
- No tests cover `useThemeService`/`ThemeProvider`; a toggle test asserting the class is added/removed and persisted would lock the contract. Consider also seeding the default from `prefers-color-scheme: dark` rather than always `light`.
- Confirm `@apicurio/common-ui-components` doesn't already expose theme handling before adding a registry-local provider, so the Apicurio apps stay consistent.

JituRewar:
@paoloantinori, I have addressed your feedback and pushed the updates:

1. Bare `[disabled]` selector removed: Scoped the disabled overrides specifically to `input:disabled`, `textarea:disabled`, and `select:disabled` in `App.css` to prevent overriding PatternFly's global disabled styling (e.g. on buttons and menus).
2. Flash-free Theme Load:
   - Added an inline script in `index.html`'s `<head>` that parses `localStorage` / prefers-color-scheme system defaults early to set the theme class on `<html>` before React renders.
   - Converted `useEffect` to `useLayoutEffect` in `ThemeProvider` to ensure theme swaps occur synchronously before layout/paint.
   - Seeded the default theme using `prefers-color-scheme: dark` instead of defaulting to light.
3. Logo Helper Extract: Extracted logo URL determination into a reusable `useLogoSrc` hook shared between `AppHeader` and `AppHeaderToolbar`.
4. Dashboard BG style cleanup: Removed redundant inline `backgroundColor` on the filled `PageSection` in `DashboardPage.tsx` since `.ps_dashboard-content` in CSS already handles it.
5. Accessibility (A11y): Added stateful `aria-label` and `aria-pressed` attributes to the theme toggle button to inform screen reader users of the current theme state.
6. Tests: Added a new unit test suite (`useThemeService.test.tsx`) covering storage initialization, system preference fallbacks, transitions, and layout effects.

All build, lint, and test checks are passing cleanly. Please let me know if this looks good to merge!

paoloantinori:
Thanks for the rework. Confirmed all four points landed:

- `App.css` now scopes the disabled overrides to `input:disabled` / `textarea:disabled` / `select:disabled` under `.pf-v6-theme-dark`, so PatternFly's own disabled styling for buttons, menus, and switches is left alone.
- The inline script in `index.html` runs before React mounts and reads `localStorage` / `prefers-color-scheme`, so the `pf-v6-theme-dark` class is set before first paint. Combined with `useLayoutEffect` in the provider, the light flash on load is gone.
- `useThemeService.test.tsx` covers seeding from storage, the `prefers-color-scheme: dark` fallback, the light fallback, toggle + persistence, and both layout-effect branches.
- The `prefers-color-scheme` default replaces the previous always-light seed.

One small thing worth a look while you're in here, not blocking: the test reassigns `globalThis.window` inside individual `it` blocks but doesn't restore it in `afterEach`/`beforeEach`. It works today because every test that reads `window.matchMedia` sets it, but a future test that forgets to set it would silently inherit the previous test's mock. A `afterEach(() => { delete (globalThis as any).window; })` would make the isolation explicit.

Otherwise this looks good to me.

EricWittmann:
Love this.  Looking forward to reviewing it.  :+1:

EricWittmann:
@NikishaPatil NOTE: this will break our RHBOAR branding commit.

JituRewar:
Hi @EricWittmann, just following up on this PR. I noticed it was self-assigned a few days ago, and the checks are now passing. Please let me know if any further changes are needed from my side.

EricWittmann (commented):
Thanks for the work on this, @JituRewar — nice job addressing all of the earlier review feedback (flash-of-light-theme fix, scoped disabled selectors, accessibility improvements, etc.).

A few things to address before we can merge:

### 1. Remove the `.github/workflows/verify.yaml` change

The Verification Gate bypass logic you added is unrelated to the dark theme feature. Please revert that change from this PR and submit it as a separate PR with its own description and rationale. We don't want CI infrastructure changes bundled with UI features — they have different review and risk profiles.

### 2. Replace hardcoded hex colors with PatternFly tokens

Several places in the diff use raw hex values for the dark theme instead of PatternFly design tokens:

- `App.css` — `#212427`, `#8a8d90`, `#3c3f42` in the disabled input rules, `#0f1011` for the dark canvas background, `#151515` for the dark card background
- `StatsCard.css` — `#f5f5f5` for the dark mode stat value text

Where PatternFly v6 provides a matching token (e.g. `--pf-t--global--background--color--primary--default`, `--pf-t--global--text--color--on-dark`, `--pf-t--global--border--color--disabled`, etc.), please use it instead of a hardcoded hex. This keeps the dark theme in sync with PatternFly's token system and avoids drift if PF updates its palette. You already did this well in other parts of the PR (e.g. `var(--pf-t--global--text--color--regular, #333)`) — just extend that approach to the remaining hardcoded values.

### 3. Restore trailing newline in `App.css`

The diff shows `No newline at end of file` at the end of `App.css`. Please add the trailing newline back — most linters and editors expect it, and it avoids noisy diffs on future changes to that file.

JituRewar:
@EricWittmann I have addressed your feedback and also pushed the updates:
1. i have successfully remove the CI Infrastructure Changes from .github/workflows/verify.yaml.
2. successfully replaced the hardcoded hex colors with PatternFly tokens.
3. And also i have restored the trailing newline in App.css.

All build, lint, and test checks are passing cleanly. Please let me know if this looks good to merge! or need any further changes from my side.

paoloantinori (commented):
Overall the theme infrastructure is well put together: a ThemeProvider context toggles the pf-v6-theme-dark class, an inline script in index.html applies it before paint to avoid a flash, and component colors move onto CSS variables. No blockers; four items (one asset check, one key-consistency check, one CSS-overreach, one test gap) are worth closing before merge.

- Dark-mode logo asset: useThemeService.tsx:62 returns apicurio_registry_logo_reverse.svg, but that file is not in this diff. If it does not already exist in the ui-app static directory, the masthead shows a broken image in dark mode. The default logo already ships there, so confirm the reverse variant does too.
- Key consistency: the pre-paint script reads localStorage.getItem('apicurio-registry.config.theme') (index.html:16) while the React side reads and writes via localStorageService.getConfigProperty('theme') and setConfigProperty('theme') (useThemeService.tsx:19, :32). Confirm the service prefixes to exactly apicurio-registry.config.theme; if not, the toggle persists to a key the FOUC script never reads and users see a flash of the wrong theme on reload. The unit test mocks the storage service, so it cannot catch this mismatch.
- Duplicated resolution logic: index.html:16-17 and useThemeService.tsx:18-27 both implement (stored, else prefers-color-scheme: dark, else light). They agree today, but a future edit to one will silently desync the pre-paint and post-mount class state. A shared constant or at least a cross-reference comment would make the coupling visible.
- CSS overreach: App.css:12-17 puts !important background, color, and border on every disabled input, textarea, and select globally in dark mode, shadowing PatternFly v6's own disabled styling (which uses the same pf-t tokens). Confirm PF v6 does not already handle this; if it does, drop the block. The [id^="tab-"] prefix selector later in App.css has the same broadness concern since it matches any element whose id starts with tab-.
- Test gap: useLogoSrc (useThemeService.tsx:59) has dark/light branching but no test; useThemeService.test.tsx imports only ThemeProvider. Add coverage for both branches and for the throw-when-no-provider path.

ui/ui-app/index.html
```diff
@@ -10,6 +10,19 @@
     <title>Apicurio Registry</title>
     <script type="text/javascript" src="./version.js"></script>
     <script type="text/javascript" src="./config.js"></script>
+    <script type="text/javascript">
+      (function() {
+        try {
+          const stored = localStorage.getItem("apicurio-registry.config.theme");
```
paoloantinori, line 16: This pre-paint script reads the raw key apicurio-registry.config.theme, but the React side persists through localStorageService.setConfigProperty('theme') (useThemeService.tsx:32). Confirm getConfigProperty('theme') resolves to exactly that localStorage key. If the service uses a different prefix, the toggle writes to a key this script never reads, so on reload the page flashes the OS-default theme instead of the saved choice. The unit test mocks the storage service, so the only way to verify this today is a manual devtools check (toggle dark, reload, watch the html class before paint).

ui/ui-app/src/services/useThemeService.tsx
```diff
@@ -0,0 +1,63 @@
...
+    if (!context) {
+        throw new Error("useThemeService must be used within a ThemeProvider");
+    }
+    return context;
+};
+
+export const useLogoSrc = (): string => {
+    const config = useConfigService();
+    const { isDark } = useThemeService();
+    return `${config.uiContextPath() || "/"}${isDark ? "apicurio_registry_logo_reverse.svg" : "apicurio_registry_logo_default.svg"}`;
```
paoloantinori, line 62: apicurio_registry_logo_reverse.svg is a new reference and is not in this diff's changed files. The default logo already ships in the ui-app static directory; confirm the reverse variant exists there too, otherwise the masthead renders a broken image in dark mode. A one-line assertion in useThemeService.test.tsx that useLogoSrc returns this filename when isDark would lock it in.

ui/ui-app/src/app/App.css
```diff
@@ -1,6 +1,20 @@
...
+
+.pf-v6-theme-dark {
+  --registry-card-bg: var(--pf-t--global--background--color--primary--default, #151515);
+  --registry-page-canvas-bg: var(--pf-t--global--background--color--secondary--default, #0f1011);
+}
+
+.pf-v6-theme-dark input:disabled,
+.pf-v6-theme-dark textarea:disabled,
+.pf-v6-theme-dark select:disabled {
+  background-color: var(--pf-t--global--background--color--disabled--default, #212427) !important;
```
paoloantinori, line 15: This slaps !important on background, color, and border for every disabled input, textarea, and select globally in dark mode. PatternFly v6's dark theme already styles disabled controls through the same pf-t tokens referenced here, so this likely double-applies or shadows PF's own rules. If PF v6 already covers disabled state in dark mode, remove this block; if it genuinely does not, scope it to .pf-v6-theme-dark .pf-c-form-control:disabled rather than bare element selectors. The [id^="tab-"] selector below has the same broadness concern.

ui/ui-app/src/services/useThemeService.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const useThemeService = (): ThemeContextProps => {
+    const context = useContext(ThemeContext);
+    if (!context) {
+        throw new Error("useThemeService must be used within a ThemeProvider");
+    }
+    return context;
+};
+
+export const useLogoSrc = (): string => {
```
paoloantinori, line 59: useLogoSrc is a newly exported hook with dark/light branching but has no test. useThemeService.test.tsx only imports ThemeProvider. Add a test that drives both isDark states and asserts apicurio_registry_logo_reverse.svg vs apicurio_registry_logo_default.svg, plus a test that useThemeService throws its documented error when called outside a ThemeProvider.

EricWittmann (commented):
Thanks for the contribution, @JituRewar! Merging now.

EricWittmann:
:man_dancing:

PR #9295, opened 2026-08-06
docs: document proxy header, auth chain, and Kubernetes-native authentication

eameh:


eameh:
Hi @EricWittmann, sorry for one more ping after the batch you just got through (thank you again for those). Could you take a look at this one when you get a chance? It documents the proxy header auth, the mechanism chain, and Kubernetes TokenReview auth from 3.3, so it would be great to have it in for the downstream release. Thanks!

EricWittmann (commented):
Thanks for putting this together — really thorough documentation for features that were missing user-facing docs. I verified all the property names, defaults, and behavioral claims against the codebase and everything checks out, with one exception:

**`APICURIO_UI_AUTH_OIDC_LOAD_USER_INFO` default value is incorrect.** The table lists the default as `false`, but the actual default in the code is `true`:

- `UserInterfaceConfigProperties.java` — the `@Info` annotation says "Defaults to true if not specified"
- The generated config reference (`ref-registry-all-configs.adoc`) also documents it as defaulting to `true`
- The Azure Entra ID guide confirms the UI library uses `loadUserInfo: true` by default

The default column in the Azure AD table should be `true`, and the description should clarify that Azure AD users need to explicitly set it to `false`. Something like:

```
|`APICURIO_UI_AUTH_OIDC_LOAD_USER_INFO`
|Whether the web console loads user details from the OIDC userinfo endpoint. Set to `false` for Azure AD, where the userinfo endpoint has incompatible audience requirements and the call fails after login.
|`true`
```

Please fix the default value and this is good to merge.

eameh:
Thanks for the careful review @EricWittmann! I have made some changes

PR #9298, opened 2026-08-06
fix(rules): flag outputSchema removal in MCP tool compatibility (#9297)

### LINKED ISSUE #9297: bug(mcp): MCP tool compatibility ignores outputSchema, so removing the output contract passes as compatible
## Description

**Registry Version**: snapshot (build from `main`, currently `211fb51f2`)
**Persistence type**: any, the bug is in `schema-util` and is storage independent
**Java**: 21
**Maven**: 3.9.x via the `./mvnw` wrapper

`McpToolCompatibilityChecker` never looks at `outputSchema`. The string does not appear anywhere
in the class. All four of its checks (`checkInputSchemaTypeChange`, `checkPropertyRemovals`,
`checkRequiredParamAdditions`, `checkRequiredParamRemovals`) read only `inputSchema`, and all five
`McpToolCompatibilityDifference.Type` values are input scoped.

So a new version of a tool can drop its entire declared output contract and the registry reports
the change as compatible. A client relying on that tool's `structuredContent` then breaks at call
time, with nothing from the compatibility rule to warn about it.

The sibling artifact type already covers this.
`PromptTemplateCompatibilityChecker.checkOutputSchemaCompatibility`, in the same package, flags
both whole schema removal and per property removal, with dedicated difference types
`OUTPUT_SCHEMA_REMOVED` and `OUTPUT_SCHEMA_PROPERTY_REMOVED`. `MCP_TOOL` and `PROMPT_TEMPLATE`
both extend `AbstractCompatibilityChecker` and both support an optional `outputSchema`, so this
reads as an oversight rather than a deliberate difference.

### Environment

Not environment specific. `schema-util/common`, reproducible from a unit test.

### Steps to Reproduce

1. Register an `MCP_TOOL` artifact declaring an `outputSchema`, for example the one already in the
   test fixture `mcptool-valid.json`:

```json
"outputSchema": {
  "type": "object",
  "properties": {
    "results": { "type": "array" },
    "total": { "type": "integer" }
  },
  "required": ["results", "total"]
}
```

2. Set the artifact's COMPATIBILITY rule to `BACKWARD`.
3. Add a new version, identical except `outputSchema` is removed entirely, or kept but with
   `results` and `total` deleted from its `properties`.
4. The new version is accepted.

Expected: removing `outputSchema`, or removing properties from it, is reported as backward
incompatible, matching how `PROMPT_TEMPLATE` already treats the same change.

Actual: no difference is reported and the version is accepted.

### Logs

No error is produced. `isBackwardsCompatibleWith` returns an empty difference set, so
`CompatibilityExecutionResult.incompatibleOrEmpty` resolves to compatible.

dobbydobap:
## What

`McpToolCompatibilityChecker` now checks `outputSchema` when comparing two versions of an MCP tool.
Removing the `outputSchema` object, or removing one of its properties, is reported as backward
incompatible.

## Why

The checker only ever read `inputSchema`, so a tool could delete its whole declared output contract
and the rule still returned compatible. A client depending on that tool's `structuredContent` then
breaks at call time with nothing flagged at registration.

`PROMPT_TEMPLATE` already treats the same change as breaking, in
`PromptTemplateCompatibilityChecker.checkOutputSchemaCompatibility`. Both types extend
`AbstractCompatibilityChecker` and both support an optional `outputSchema`, so the two were
inconsistent for no clear reason.

## How

Added `checkOutputSchemaCompatibility`, following the prompt template implementation so the two
behave the same way, plus two values on `McpToolCompatibilityDifference.Type`,
`OUTPUT_SCHEMA_REMOVED` and `OUTPUT_SCHEMA_PROPERTY_REMOVED`, using the same JSON pointer context
style as the existing input scoped values.

No change to the input side, and no change to how compatibility levels are evaluated.

## Testing

Three cases added to `McpToolCompatibilityCheckerTest`: removing `outputSchema` entirely, removing
a single property from it, and adding a property staying compatible. All 12 tests in the class pass.

Both new incompatibility tests fail against unmodified `main` with `expected: <false> but was:
<true>`, so they cover the reported behaviour rather than restating current behaviour.

`checkstyle:check` passes on `schema-util/common` and `schema-util/util-provider`.

Closes #9297

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityChecker.java
@@ -18,6 +18,7 @@ import java.util.Set;
  * - Adding required parameters: Backward incompatible
  * - Removing required parameters: Backward incompatible
  * - Changing inputSchema type: Backward incompatible
+ * - Removing outputSchema, or a property of it: Backward incompatible
  * - Changing name, title, description, annotations: Always compatible
  */
 public class McpToolCompatibilityChecker
@@ -45,6 +46,9 @@ public class McpToolCompatibilityChecker
 
             // Check removed required parameters
             checkRequiredParamRemovals(existingNode, proposedNode, differences);
+
+            // Check removed output schema or output properties
+            checkOutputSchemaCompatibility(existingNode, proposedNode, differences);
 
         } catch (Exception e) {
             differences.add(new McpToolCompatibilityDifference(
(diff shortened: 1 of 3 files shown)


carlesarnal:
I'm closing this as stale, if you plan to work on it, re-open.

PR #9299, opened 2026-08-06
fix(ui): handle plain-text error responses in unwrapErrorData

### LINKED ISSUE #9169: [UI] unwrapErrorData pollutes error object by spreading plain-text/HTML responses
## Description

**Registry Version:** snapshot
**Persistence type:** N/A (React UI)

### Details

The `unwrapErrorData` function in `ui/ui-app/src/utils/rest.utils.ts` assumes that `error.response.data` is an object and spreads it into the returned error object.

When the backend or intermediary infrastructure (such as a reverse proxy, load balancer, or the Quarkus server) returns a plain-text or HTML error response instead of JSON (for example, HTTP `504 Gateway Timeout` or `413 Payload Too Large`), `error.response.data` is a string.

Spreading a string with the JavaScript spread operator (`...error.response.data`) produces character-indexed properties instead of preserving the original message. As a result, the returned error object contains unexpected numeric keys and no longer has the expected structure.

### Environment

* Apicurio Registry UI (React)
* Any deployment where the server or intermediary infrastructure can return plain-text or HTML error responses

### Steps to Reproduce

1. Trigger a request that results in a plain-text or HTML error response (for example, an HTTP `504 Gateway Timeout` or `413 Payload Too Large`).
2. Allow the UI to process the error through `unwrapErrorData`.
3. Inspect the returned error object.

### Expected Behaviour

Plain-text error responses should be preserved as a message field while maintaining the expected error object structure.

Example:

```json
{
  "message": "504 Gateway Timeout",
  "status": 504
}
```

### Actual Behaviour

The plain-text response is spread into character-indexed properties, producing an unexpected object such as:

```json
{
  "0": "5",
  "1": "0",
  "2": "4",
  "3": " ",
  "4": "G",
  "...": "...",
  "message": "Request failed with status code 504",
  "status": 504
}
```

### Logs

Example output from `unwrapErrorData` when `error.response.data` contains the string `"504 Gateway Timeout"`:

```javascript
{
  '0': '5',
  '1': '0',
  '2': '4',
  '3': ' ',
  '4': 'G',
  '5': 'a',
  '6': 't',
  '7': 'e',
  '8': 'w',
  '9': 'a',
  '10': 'y',
  '11': ' ',
  '12': 'T',
  '13': 'i',
  '14': 'm',
  '15': 'e',
  '16': 'o',
  '17': 'u',
  '18': 't',
  message: 'Request failed with status code 504',
  status: 504
}
```

kunaldevsahu:
## Summary

This PR fixes an issue in the UI error handler where plain-text or HTML error responses were incorrectly spread into character-indexed properties.

Previously, `unwrapErrorData` assumed `error.response.data` was always an object. When the response body was a string (for example, from a `504 Gateway Timeout` or `413 Payload Too Large` response), using the spread operator (`...error.response.data`) produced an object with numeric keys instead of preserving the original message.

This change updates `unwrapErrorData` to:

* Detect whether `error.response.data` is a non-null object before spreading it.
* Wrap non-object values (such as strings) as `{ message: error.response.data }`.
* Preserve the existing behavior for JSON error responses.

Additionally, unit tests have been added to cover:

* JSON object error responses.
* Plain-text/HTML error responses.
* `null` and `undefined` response data.
* Errors without a response object.
* Unknown or falsy errors.

## Testing

* Added unit tests for all supported error response scenarios.
* Verified all tests pass with `npm run test`.
* Verified linting passes with `npm run lint`.

## Related Issue

Closes #9169


ui/ui-app/src/utils/rest.utils.test.ts
```diff
@@ -1,60 +1,70 @@
...
+import { unwrapErrorData } from "./rest.utils";
 
-describe("createEndpoint", () => {
-    const baseHref = "http://localhost:8080/apis/registry/v3";
-
-    it("should correctly substitute path parameters", () => {
-        const url = createEndpoint(baseHref, "/groups/:groupId/artifacts/:artifactId", {
-            groupId: "my-group",
-            artifactId: "my-artifact"
+describe("unwrapErrorData", () => {
```
paoloantinori, line 4: This replaces the entire `createEndpoint` test suite with `unwrapErrorData` tests. `createEndpoint` (path-param substitution, query params with falsy booleans and numbers, null/undefined/empty-string omission) loses all coverage in this file. Were these tests moved elsewhere, or should the `createEndpoint` cases be restored alongside the new ones rather than swapped in?

ui/ui-app/src/utils/rest.utils.ts
```diff
@@ -215,12 +215,15 @@ function createAxiosConfig(method: string, url: string, options: any, data?: any
 }
 
 
-function unwrapErrorData(error: any): any {
+export function unwrapErrorData(error: any): any {
     console.debug("Error detected, unwrapping...");
     if (error && error.response && error.response.data) {
+        const errorData = (typeof error.response.data === "object" && error.response.data !== null)
```
paoloantinori: `typeof error.response.data === "object"` is true for arrays too, so a JSON-array error body still spreads into numeric-indexed keys (`{0: ..., 1: ...}`), the same bug class this change fixes for strings. If arrays are a possible response shape, tighten the guard, for example `typeof x === "object" && x !== null && !Array.isArray(x)`. As written, a number or boolean body is also wrapped as a non-string `message` (`{ message: 123 }`), which could surprise consumers expecting a string; a `String(error.response.data)` coercion in the fallback would close that.

paoloantinori (commented):
The fix correctly stops plain-text response bodies from being spread into numeric-indexed properties, and the new `unwrapErrorData` tests cover the right cases (string, object, null, no-response, unknown). Two things to address before merge: the PR removes the entire existing `createEndpoint` test suite, and the new type guard is too permissive for arrays.

- `ui/ui-app/src/utils/rest.utils.test.ts:4` - the whole `createEndpoint` describe block (path-param substitution, query-param handling including falsy booleans and numbers, and null/undefined/empty-string omission) is replaced with `unwrapErrorData` tests. `createEndpoint` now has no coverage in this file. The summary says tests were added, but they were swapped in, not added alongside. Restore the `createEndpoint` cases or confirm they were relocated to another file.
- `ui/ui-app/src/utils/rest.utils.ts:221` - `typeof error.response.data === "object"` is also true for arrays, so a JSON-array error body still spreads into `{0: ..., 1: ...}` numeric keys, which is the same bug class this change fixes for strings. As written, a number or boolean body is also wrapped as a non-string `message` (`{ message: 123 }`), which could surprise any consumer that assumes `message` is a string. If arrays or scalar bodies are possible response shapes, tighten the guard, for example `typeof x === "object" && x !== null && !Array.isArray(x)`, and coerce the fallback with `String(error.response.data)`.
- The core type-discrimination logic and the five new test cases are correct and well targeted; the change itself is low risk for the string case it set out to fix.

ui/ui-app/src/utils/rest.utils.test.ts
kunaldevsahu: Thanks for catching that ,The createEndpoint tests weren't intended to be removed. I'll restore the existing createEndpoint test coverage and keep the new unwrapErrorData tests alongside them.

ui/ui-app/src/utils/rest.utils.ts
kunaldevsahu: Good catch, thanks! I'll tighten the type guard to exclude arrays and update the fallback to coerce scalar response bodies to a string so all response shapes are handled safely

EricWittmann (commented):
Thanks for the fix, @kunaldevsahu — the core change looks correct and well-scoped. The type guard with the `!Array.isArray()` check and `String()` coercion addresses the review feedback from @paoloantinori nicely.

One remaining issue before we can merge:

**Restore the `createEndpoint` test suite.** The current diff replaces the entire existing `createEndpoint` describe block with the new `unwrapErrorData` tests. That means `createEndpoint` (path-param substitution, query-param handling, null/undefined/empty-string omission) now has zero test coverage. You acknowledged this in your reply to Paolo's comment and said you'd restore them — please add them back alongside the new `unwrapErrorData` tests so both suites coexist in the file.

Once that's addressed, this is ready to go.

kunaldevsahu:
Thanks for the review @EricWittmann . I've addressed this by restoring the `createEndpoint` test suite alongside the new `unwrapErrorData` tests, so both sets of tests are now present in the file. Please let me know if there's anything else you'd like me to update.

kunaldevsahu:
Hi @EricWittmann, just a friendly follow-up. I've addressed the requested feedback and restored the createEndpoint tests alongside the new unwrapErrorData tests. All checks are passing now. Whenever you get a chance, I'd appreciate another review. Thanks!

EricWittmann:
Thanks @kunaldevsahu

Issue #8751, opened 2026-07-20
Unclosed FileInputStream in maven-plugin register goal


Anshumancanrock:
### Summary

I was reading through the maven plugin and noticed the `register` goal opens a `FileInputStream` per artifact file and never closes it.

The relevant part in `RegisterRegistryMojo`:

```java
if (artifact.getFile() != null) {
    return registerArtifact(registryClient, artifact, new FileInputStream(artifact.getFile()),
            references);
}
```

That stream goes into the `registerArtifact(..., InputStream artifactContent, ...)` overload, which reads it (`artifactContent.readAllBytes()`, or `ObjectMapper.readValue` when minify is on) but never closes it. No try-with-resources, nothing in a finally. So the handle stays open until GC eventually collects the stream.

For a build with a few artifacts you'd never notice, but registering a lot of them from files leaks one fd each, and with a low `ulimit -n` that can end in "Too many open files". I haven't hit the limit in a real build - spotted it by reading the code - but it's a straightforward leak. Easy to prove with a test that wraps the stream and checks `close()` was called.

Registry version: main (3.3.1-SNAPSHOT). Persistence type isn't relevant, this is all in the maven plugin.

Fix is just wrapping it:

```java
try (InputStream is = new FileInputStream(artifact.getFile())) {
    return registerArtifact(registryClient, artifact, is, references);
}
```

I can put up a PR if this is worth fixing.

Anshumancanrock:
hi @carlesarnal , could you please take a look at this issue? Thanks !

bhuvan-somisetty:
Hi @carlesarnal @paoloantinori - I noticed the unclosed \FileInputStream\ in \RegisterRegistryMojo\ when registering artifact files. I'd like to work on this and open a PR wrapping it in try-with-resources (similar to \TestUpdateRegistryMojo\). Could you please assign this to me? Thanks!

shruti-sharma1:
Hey @bhuvan-somisetty are you working on this or can i take over?

bhuvan-somisetty:
hello @shruti-sharma1, im working on it & will raise pr soon

Priyanshubhartistm:
Hi @Anshumancanrock, I'd like to work on this issue. If you're not already working on it, would you mind if I took it up?

Thanks!

Anshumancanrock:
Sure @Priyanshubhartistm , feel free to open the PR for this !

PR #9303, opened 2026-08-06
feat(cli): auto-save group and artifact IDs to context

### LINKED ISSUE #8022: CLI: Auto context update — set group/artifact in context after operations

Priyanshubhartistm:
## Summary

Adds an opt-in `context.auto-update` configuration property. When enabled, `group create`/`get` and `artifact create`/`get` automatically save the resolved group and/or artifact ID to the current context, so subsequent commands don't need repeated `-g`/`-a` flags.

Fixes #8022.

### Context auto-update (`IdUtil`)
- Added `updateGroupContext(groupId, config, output)` and  `updateArtifactContext(groupId, artifactId, config, output)`, both gated by the  `context.auto-update` property (default: disabled).
- `updateGroupContext` only clears the context's `artifactId` when the group actually
  changes — re-inspecting the group you're already working in (e.g. `group get` on the current group) leaves the artifact untouched.
- A failure while writing the context is caught and reported as a stderr warning rather than failing the command, since the primary group/artifact operation already succeeded.

### Wiring (`GroupCreateCommand`, `GroupGetCommand`, `ArtifactCreateCommand`, `ArtifactGetCommand`)
- Each command calls the corresponding `IdUtil` method after a successful create/get, using the resolved (not raw user-supplied) group ID so context stays accurate whether the user passed `-g` explicitly, relied on the existing context, or fell back to  `default`.

### Documentation (`cli/README.md`)
- Documented `context.auto-update` in the configuration properties table and added a  short note under Context Management explaining the enable command and the  group-change/artifact-clear behavior.

## Testing

- `AutoContextUpdateTest` (new): disabled-by-default for both group and artifact  create, context updates after group/artifact create and get, group-change clears the  stale artifact ID, and re-getting the same group preserves it.
- `./mvnw checkstyle:check -pl cli` — 0 violations.
- `./mvnw test -pl cli` — 293 passed, 0 failed.
- Manual end-to-end verification against a live registry container using the exact
  `acr group add` / `acr artifact add` / `acr artifact version add` sequence from the  issue, with no `-g`/`-a` flags — confirmed the context populates correctly and downstream commands resolve group/artifact from it.

## Notes

- Only `group`/`artifact` `create` and `get` update the context, matching the issue's acceptance criteria. `artifact version add` and other subcommands already resolve  group/artifact from context via the pre-existing `IdUtil.resolveGroupId` /
  `resolveArtifactId` — no changes were needed there.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -207,11 +207,12 @@ acr config delete <property-name>
 
 #### Configuration Properties
 
-| Property                             | Default | Description                                                                                  |
-|---------------------------------------|---------|-----------------------------------------------------------------------------------------------|
-| `update.check-enabled`               | `true`  | Enable automatic update checks                                                               |
-| `update.timeout-seconds`             | `60`    | Timeout for update network requests                                                          |
-| `update.skip-checksum-verification`  | `false` | Skip SHA-256 verification of downloaded archives (for custom repos without `.sha256` files)  |
+| Property                             | Default | Description                                                                                                             |
+|---------------------------------------|---------|---------------------------------------------------------------------------------------------------------------------------|
+| `update.check-enabled`               | `true`  | Enable automatic update checks                                                                                          |
+| `update.timeout-seconds`             | `60`    | Timeout for update network requests                                                                                     |
+| `update.skip-checksum-verification`  | `false` | Skip SHA-256 verification of downloaded archives (for custom repos without `.sha256` files)                             |
+| `context.auto-update`                | `false` | Automatically save the group/artifact ID to the current context after `group create`/`get` and `artifact create`/`get` |
 
 #### Logging
 
@@ -271,6 +272,15 @@ acr context delete --all
 ```
 
 Use `--no-switch-current` when creating a context to add it without switching to it.
+
+**Auto context update:**
+```bash
+acr config set context.auto-update=true
+```
+When enabled, `group create`/`get` and `artifact create`/`get` automatically save the resolved group and/or
+artifact ID to the current context, so subsequent commands don't need repeated `-g`/`-a` flags. Switching to a
+different group clears the artifact ID from the context, since the previous artifact no longer applies; re-getting
+the same group leaves it untouched. Disabled by default.
 
 ### Authentication
 
--- a/cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
@@ -132,6 +132,7 @@ public class ArtifactCreateCommand extends AbstractCommand {
                 .groups().byGroupId(resolvedGroupId).artifacts().post(newArtifact);
         //noinspection ConstantConditions
         final var artifact = convert(result.getArtifact());
+        IdUtil.updateArtifactContext(resolvedGroupId, artifact.getArtifactId(), config, output);
         switch (outputType.getOutputType()) {
             case json -> output.writeStdErrChunk(out -> successMessage(out, resolvedGroupId, artifact.getArtifactId()));
             case table -> output.writeStdOutChunk(out -> successMessage(out, resolvedGroupId, artifact.getArtifactId()));
--- a/cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactGetCommand.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactGetCommand.java
@@ -80,6 +80,7 @@ public class ArtifactGetCommand extends AbstractCommand {
         } else {
             fetchMetadata(registryClient, resolvedGroupId, output);
         }
+        IdUtil.updateArtifactContext(resolvedGroupId, artifactId, config, output);
     }
 
     // Fetches the raw content of the latest version of the artifact.
--- a/cli/src/main/java/io/apicurio/registry/cli/common/IdUtil.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/common/IdUtil.java
@@ -1,7 +1,11 @@
 package io.apicurio.registry.cli.common;
 
 import io.apicurio.registry.cli.config.Config;
+import io.apicurio.registry.cli.config.ConfigModel;
+import io.apicurio.registry.cli.utils.OutputBuffer;
 import io.apicurio.registry.rest.client.RegistryClient;
+import java.util.Objects;
+import java.util.function.Consumer;
 
 import io.apicurio.registry.rest.v3.beans.GroupSearchResults;
 import io.apicurio.registry.rest.v3.beans.SearchedGroup;
@@ -17,6 +21,8 @@ import static io.apicurio.registry.cli.utils.Utils.isBlank;
 public final class IdUtil {
 
     public static final String DEFAULT_GROUP = "default";
+
+    private static final String AUTO_CONTEXT_UPDATE_PROPERTY = "context.auto-update";
 
     private IdUtil() {
     }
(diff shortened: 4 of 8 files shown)


Priyanshubhartistm:
@jsenko, @paoloantinori  could you please take a look and let me know if anything else needs to be changed?
 Thanks!

paoloantinori:
Had a look — the design here is sound and I have no blocking points.

A few things I checked deliberately and they hold up:

- The opt-in default (`context.auto-update=false`) is the right call. Silently mutating the user's active context on every `get` would surprise people; gating it behind a property that defaults off keeps existing behavior identical.
- `updateGroupContext` clearing the artifact ID only when the group *actually changes* (the `!groupId.equals(context.getGroupId())` guard) is the detail that makes `group get` on the group you're already in non-destructive. That is the one subtlety a naive implementation would get wrong, and the test `testSameGroupGetPreservesArtifactInContext` pins it down.
- Surfacing a context-write failure as a stderr warning rather than failing the command is correct — the primary group/artifact operation already succeeded, so the context save is a side effect that should not be able to roll back the user's intent.
- `ArtifactGetCommand` calls `updateArtifactContext` *after* `fetchMetadata`, so a failed fetch (e.g. missing artifact) throws before any context mutation. Good ordering.

Nothing to change from my side.

Priyanshubhartistm:
Hi @vandanayadav7, could you please take a look at this PR? If there's anything else I should improve, please let me know. Thank you!

jsenko (changes requested):
Fixes #8022 — auto-save group/artifact IDs to context after create/get commands.

## Context

The feature design is sound — opt-in default, group change clears stale artifact ID, good test coverage across 8 scenarios. paoloantinori confirmed the approach. One structural refactoring needed before merge.

## Summary

The context mutation logic should live in the `Config` bean, not in `IdUtil`. This unifies the existing duplication with `ContextUpdateCommand` and makes the feature reusable for future commands.

### should-fix: Move context update logic from IdUtil to Config

`IdUtil` is a static utility for ID resolution and validation. Context persistence (read → resolve context → mutate → write) is a different concern and belongs in `Config`, which already owns `read()` and `write()`.

**Current issues:**
- `IdUtil.updateContext()` threads `config` and `output` through every call site (4 places)
- `ContextUpdateCommand` (lines 50-74) has the same read → resolve → validate → mutate → write pattern inlined separately
- Adding auto-context to future commands means importing `IdUtil` and passing `config` + `output` each time

**Proposed refactoring:**

Add to `Config`:

```java
public void updateCurrentContext(Consumer<ConfigModel.Context> mutator) {
    var model = read();
    var name = model.getCurrentContext();
    if (isBlank(name)) return;
    var ctx = model.getContext().get(name);
    if (ctx == null) return;
    mutator.accept(ctx);
    write(model);
}

public boolean isAutoUpdateEnabled() {
    return Boolean.parseBoolean(read().getConfig().get("context.auto-update"));
}
```

Then in commands (e.g. `GroupCreateCommand`):
```java
if (config.isAutoUpdateEnabled()) {
    config.updateCurrentContext(ctx -> {
        if (!groupId.equals(ctx.getGroupId())) ctx.setArtifactId(null);
        ctx.setGroupId(groupId);
    });
}
```

### should-fix: Cache config.read() result

`IdUtil.updateContext()` line 95 calls `config.read()` to check the property, then line 98 calls `config.read()` again. Read once and reuse the result. This is resolved naturally if the logic moves to `Config`.

### optional: Unify ContextUpdateCommand with the new method

`ContextUpdateCommand` (lines 50-74) can also use `Config.updateCurrentContext()` for the mutation step, reducing its implementation to validation + delegation. The existing logic there:

```java
// Current inline pattern in ContextUpdateCommand:
final var configModel = config.read();
final var contextName = isBlank(name) ? configModel.getCurrentContext() : name;
// ... validate ...
context.setGroupId(groupId);
context.setArtifactId(artifactId);
config.write(configModel);
```

Note: `ContextUpdateCommand` also supports an explicit context name parameter (not just current), so `updateCurrentContext` covers the common case. The named-context variant could be a second method (`Config.updateContext(String name, Consumer<Context>)`) if you want full unification, but that's optional.

### optional: Other context update sites to consider

If you're refactoring the context update path, these are other places where context fields are set that could potentially use the same method in the future (not required for this PR, just for awareness):
- `ContextCreateCommand` — sets initial context values on creation
- Any future commands that might benefit from auto-context (e.g. `version get`, `search artifact`)

---

Thank you for the clean implementation and thorough tests, @Priyanshubhartistm. The feature logic is correct — the refactoring above is about putting it in the right place so it's easy to extend later.

Priyanshubhartistm:
Hii @jsenko   just a gentle reminder the PR is ready for review. Could you please take a look when you get a chance and let me know if any changes are needed?
Thank you!

Priyanshubhartistm:
Hii @EricWittmann , @paoloantinori , I have a fix for #8751 (unclosed `FileInputStream` in the maven-plugin `register` goal) ready to go, but the one-open-PR-per-contributor rule means I can't open it until this one closes. Would appreciate a review/merge here when you get a chance so I can get that follow-up out.

paoloantinori:
@Priyanshubhartistm noted on the follow-up for #8751 — the one-open-PR-per-contributor rule is holding you up. This one (`context.auto-update`) still reads clean from my side: the opt-in default, the group-change-clears-artifact guard, and the warning-on-write-failure handling all hold up on the current head. The substantive ground is covered; happy to see it move forward so you can open the `FileInputStream` follow-up.

Priyanshubhartistm:
@EricWittmann, could you please take a look at these PRs? They’ve been open for the past two weeks. If everything looks good, could you please merge them when you get a chance? Thank you!

Priyanshubhartistm:
@paoloantinori , could you please take a look at these PRs?

paoloantinori:
Had another look at the current head, thanks for the ping.

132d9d6 lands the refactor @jsenko asked for: `updateCurrentContext()` and `isAutoUpdateEnabled()` now live in `Config` with the same guards as his sketch, and `IdUtil` keeps only the thin gated wrapper plus the error handling. The open gate on this PR is his re-review of the CHANGES_REQUESTED, so I will leave that with him rather than pile on.

From my side there are still no blocking points. The same-group-preserves-artifact behavior and the warning-instead-of-failure on context write errors both look right.

Priyanshubhartistm:
Hii @jsenko just a gentle reminder the PR is ready for review. Could you please take a look when you get a chance and let me know if any changes are needed?
Thank you!

jsenko (commented):
Re-review of commit 132d9d6 — author addressed jsenko's CHANGES_REQUESTED feedback.

### Re-review delta

**Resolved findings:**
- **Move context update logic from IdUtil to Config** — Done. `Config.isAutoUpdateEnabled()` and `Config.updateCurrentContext(Consumer<Context>)` now own the core logic. `IdUtil.updateGroupContext/updateArtifactContext` are convenience wrappers that delegate to these Config methods.
- **Cache config.read() result** — Done. Each Config method calls `read()` once (no double reads).

**Optional finding (not addressed, correctly so):**
- Unify `ContextUpdateCommand` with the new method — this was marked optional by jsenko and was not done, which is fine.

### Current state

The refactoring is clean. The separation makes sense:
- `Config` owns persistence (read, write, mutation) and feature flags (`isAutoUpdateEnabled`)
- `IdUtil` remains a static utility for ID resolution/validation, with thin context-update wrappers for convenience

All 4 command integration points are correctly placed after successful execution.

No further issues found.

Priyanshubhartistm:
Hi @jsenko , the Verify / Verification Gate check is failing on this PR. Could you please take a look at it once and let me know what might be causing the issue?

jsenko:
> Hi @jsenko , the Verify / Verification Gate check is failing on this PR. Could you please take a look at it once and let me know what might be causing the issue?

The PR should've already been accepted and the tests running, but it looks like doing that again works now. We updated the PR workflow recently, so there might've been a problem with migration.

jsenko (approved):

jsenko:
Merged, thank you!

Priyanshubhartistm:
@jsenko Cool, thanks for letting me know.😀

PR #9305, opened 2026-08-06
fix(cli): use ContentTypeDetector fallback in artifact/version create…

### LINKED ISSUE #9294: CLI: `artifact create` / `version create` always default content-type to application/json, ignoring --type
## Description

**Registry Version**: 3.3.2-SNAPSHOT (main)
**Persistence type**: in-memory (H2)

When creating an artifact via the CLI without explicitly passing `--content-type`, the CLI always defaults to `application/json`, regardless of the artifact's `--type`. This is wrong for non-JSON-based types such as PROTOBUF, XML, WSDL, XSD, GRAPHQL, and THRIFT.

### Steps to Reproduce

1. Start the registry.
2. Create a Protobuf artifact via the CLI without `--content-type`:
   ```
   acr artifact create my-proto-test --type PROTOBUF -f schema.proto
   ```
3. Fetch the raw content: `GET /apis/registry/v3/ids/globalIds/{globalId}`

### Actual Behaviour **(verified locally)**

```
HTTP/1.1 200 OK
Content-Disposition: attachment; filename="1.json"
Content-Length: 66

syntax = "proto3";
message Greeting {
  string message = 1;
}
```

Image

The Protobuf content is served with `Content-Type: application/json` and a `.json` filename, despite the artifact being registered as `PROTOBUF`.

### Expected Behaviour

The response should reflect the artifact type :  e.g. `application/x-protobuf` content-type and `.proto` filename for PROTOBUF, `application/xml` for XML/XSD/WSDL, etc.

### Root cause

- `cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java` hardcodes `!isBlank(contentType) ? contentType : "application/json"` regardless of `--type`.
- Same pattern in `cli/src/main/java/io/apicurio/registry/cli/version/VersionCreateCommand.java`.

### Suggested fix

The codebase already has `io.apicurio.registry.cli.utils.ContentTypeDetector`, used in
`SearchByContentCommand` to infer content-type from the file extension when none is given.

`ArtifactCreateCommand` and `VersionCreateCommand` were simply missed when this utility was
introduced - they should use the same `ContentTypeDetector.detect(file)` fallback instead of
hardcoding `application/json`.

Related: #8023 (missing --type validation, a separate but adjacent gap).

adrish-mage:
… instead of hardcoding application/json
Fixes #9294

## How I tested
Before fix:
image

After fix:
image

Added an integration test (`ArtifactCommandTest#testArtifactCreateInfersContentTypeFromArtifactType`)
that creates a PROTOBUF artifact without --content-type and asserts the raw content
is served back with the correct Content-Type header.


cli/src/main/java/io/apicurio/registry/cli/artifact/ArtifactCreateCommand.java
```diff
@@ -120,7 +121,9 @@ public void run(final OutputBuffer output) throws Exception {
             }
             final var versionContent = new VersionContent();
             versionContent.setContent(content);
-            versionContent.setContentType(!isBlank(contentType) ? contentType : "application/json");
+            versionContent.setContentType(!isBlank(contentType)
+                        ? contentType
+                        : ContentTypeDetector.detect(file));
```
paoloantinori: The default now comes entirely from ContentTypeDetector.detect(file). Worth confirming it does not depend on the file extension alone: if detect() ignores --type, then an extensionless file, or a .json/.txt file holding protobuf, makes a PROTOBUF artifact created without --content-type land as application/json, so #9294 is not fully fixed. Ideally the fallback honors --type (PROTOBUF maps to application/x-protobuf) with the extension as a secondary signal. Also confirm detect() never returns null for unrecognized content, since the previous default was always a non-null string. The same change applies to VersionCreateCommand.java:122.

paoloantinori (commented):
The switch from a hardcoded application/json default to ContentTypeDetector.detect(file) is the right direction, and the new integration test is a useful guard. The main thing to confirm is whether detect() actually resolves #9294 for all inputs (extensionless or mismatched-extension files, and never returning null); a few smaller test-hygiene and cosmetic items follow.

- ArtifactCreateCommand.java:126 and VersionCreateCommand.java:122: the new default comes entirely from ContentTypeDetector.detect(file). Worth confirming it does not depend on the file extension alone: if detect() ignores --type, then an extensionless file, or a .json/.txt file holding protobuf, makes a PROTOBUF artifact created without --content-type land as application/json, so #9294 is not fully fixed. Ideally the fallback honors --type with the extension as a secondary signal. Also confirm detect() never returns null for unrecognized content, since the previous default was always a non-null string.
- ArtifactCommandTest.java:92: this test creates proto-content-type-artifact in the shared default group and cleans up only the temp file (line 117), never the artifact. The class has ordered, stateful tests (the 'Ordered tests' section below), so the test currently relies on JUnit running unordered methods last to avoid colliding with emptiness or count assertions. Deleting the artifact in finally would make it order-independent.
- ArtifactCommandTest.java:110: the verification GET via HttpClient.newHttpClient().send(...) sets no connect or response timeout; a stalled registry hangs the call until the suite-level timeout instead of failing fast.
- Minor and cosmetic: the --content-type help text in both commands (e.g. ArtifactCreateCommand.java:88, not in this diff) still reads as defaulting to application/json and is now stale; the trailing-whitespace-only edit on the //noinspection line (VersionCreateCommand.java:127) should be dropped; the identical fallback ternary is duplicated across both commands and could live in a shared helper.

cli/src/test/java/io/apicurio/registry/cli/ArtifactCommandTest.java
```diff
@@ -69,7 +74,50 @@ public void testArtifactUpdateNoOptions() {
...
+        Files.writeString(tempFile, """
+                syntax = "proto3";
+                message Greeting {
+                string message = 1;
+                }
+                """);
 
+        try {
+                out.getBuffer().setLength(0);
+                executeAndAssertSuccess("artifact", "create", "--output-type", "json",
```
paoloantinori, line 98: This test creates proto-content-type-artifact in the shared default group but only deletes the temp file in finally (line 117), never the artifact. The class has ordered, stateful tests below (the 'Ordered tests' section), so this relies on JUnit running unordered methods last to avoid colliding with their assertions. Deleting the artifact in finally (artifact delete by id) would make the test order-independent and self-contained.

cli/src/test/java/io/apicurio/registry/cli/ArtifactCommandTest.java
```diff
@@ -69,7 +74,50 @@ public void testArtifactUpdateNoOptions() {
...
+                        "--group", "default", "--artifact", artifact.getArtifactId(),
+                        "--output-type", "json");
+                JsonNode version = MAPPER.readTree(out.toString());
+                long globalId = version.get("globalId").asLong();
+
+                var request = HttpRequest.newBuilder()
+                        .uri(URI.create(registryUrl + "/apis/registry/v3/ids/globalIds/" + globalId))
+                        .GET()
+                        .build();
+                var response = HttpClient.newHttpClient()
```
paoloantinori: The verification GET uses HttpClient.newHttpClient().send(...) with no connect or response timeout. If the registry stalls mid-request, this hangs until the suite-level timeout instead of failing fast. Add an explicit connectTimeout on the client builder and a timeout on the request.

cli/src/main/java/io/apicurio/registry/cli/version/VersionCreateCommand.java
```diff
@@ -116,12 +117,14 @@ public void run(final OutputBuffer output) throws Exception {
...
-        versionContent.setContentType(!isBlank(contentType) ? contentType : "application/json");
+        versionContent.setContentType(!isBlank(contentType)
+                ? contentType
+                : ContentTypeDetector.detect(file));
         newVersion.setContent(versionContent);
 
         final var registryClient = client.getRegistryClient();
         IdUtil.validateGroup(registryClient, resolvedGroupId);
-        //noinspection ConstantConditions
+        //noinspection ConstantConditions
```
paoloantinori: This line is changed only by adding trailing whitespace to the //noinspection comment. Drop the trailing spaces so the line stays unmodified and the diff stays focused on the actual fix.

adrish-mage:
Hi!  @paoloantinori  I've worked it.
### What changed
- Infer the default content type from `--type` when `--content-type` is not provided.
- Fall back to the file extension if the artifact type is unavailable or unrecognized.
- Keep `version create` behavior unchanged (it still infers from the file extension).
- Update the CLI help text.
- Add regression tests, including an extensionless PROTOBUF case.

I wasn't able to run the Testcontainers-based tests locally because Docker isn't available on my machine, so I'd appreciate it if the CI could verify them. Thanks for taking a look!

paoloantinori:
Thanks @adrish-mage — the rework addresses the earlier concern. `ContentTypeDetector.detect(artifactType, file)` now treats `--type` as the primary signal with the file extension as a fallback, reuses the `ContentTypes` constants instead of local string literals, and `VersionCreateCommand` keeps its extension-only path as you noted. The regression test (`testArtifactCreateInfersContentTypeFromArtifactType`) covers the extensionless PROTOBUF case end-to-end by asserting the served `Content-Type` header.

The Testcontainers caveat is fine — those run in CI.

Nothing further from me.

adrish-mage:
Thanks for the thorough review, @paoloantinori , glad the extensionless PROTOBUF test covers the concern. And thanks for accepting, @EricWittmann. All checks are passing now with no conflicts — ready whenever you get a chance, @vandanayadav7.

paoloantinori:
Thanks for the follow-up. The extensionless PROTOBUF case is covered now, and with Eric's accept and CI green this is good to go from my side.

vandanayadav7 (approved):
Verified with e2e testing — PROTOBUF artifact now correctly served as `application/x-protobuf` instead of `application/json`. The artifact type → content type mapping is the right approach, with file extension as fallback. No regressions — JSON-based types (AVRO, JSON, OPENAPI, ASYNCAPI) still default to `application/json` as before. Code is clean and uses shared `ContentTypes` constants. LGTM.

adrish-mage:
@vandanayadav7 branch update needs a workflow approval to run required checks — could you approve it when you get a chance?

adrish-mage:
@vandanayadav7 @paoloantinori @EricWittmann 
The build failure after updating the branch appears unrelated to this PR's changes. The compilation errors are all in mcptools/wellknown/a2a packages — missing symbols like McpToolSearchResult, AgentCard, AgentSearchResults, and a missing package io.apicurio.registry.a2a.rest.beans. This PR only touches artifact-type → content-type mapping and doesn't reference any of these classes.

This looks like it may be coming from main in its current state rather than from this branch. Could someone confirm whether main is currently building cleanly?

paoloantinori:
Confirmed. I pulled the workflow run and the red job is `Build / Build Java Application (no tests)`; the only failing module is `app`, and every error is in the `mcptools` / `wellknown` / `a2a` packages (`McpToolSearchResult`, `AgentCard`, `io.apicurio.registry.a2a.rest.beans`). None of those are referenced by your CLI changes.

I also re-checked `main` directly: `McpCompatibleToolsResults.java` on `main` imports `io.apicurio.registry.rest.v3.beans.McpToolSearchResult`, but that class does not exist on `main` (raw.githubusercontent returns 404), and the `a2a/rest/beans` package is missing there too. So `main` itself is in a non-compiling state right now, and this PR just inherited it via the branch update.

Nothing for you to fix here. A `/rebase` once `main` is green again (or a `/retry` if the broken commit gets amended) should clear it. The PR's own diff (`ContentTypeDetector`, the two create commands, the integration tests) is unaffected.

Issue #9228, opened 2026-08-04
Prompt template `default` values are never applied when rendering


Aarti-panchal01:
## Description

**Registry Version**: snapshot (main @ b35c8c680)
**Persistence type**: in-memory

A prompt template variable can declare a `default`. The registry's own schema describes it as "Default value if not provided" (`app/src/main/resources/schemas/prompt-template-v1.json:101-103`), the UI documents it in the Variables table, and several shipped example templates rely on it.

No server-side renderer ever applies it. If a caller omits a variable that has a declared default, the rendered output keeps the raw `{{placeholder}}`. Because variables with defaults are typically `required: false`, `validationErrors` also comes back empty, so nothing signals that the prompt is broken.

### Environment

Local build of main, `quarkus:dev`, in-memory storage. Reproduced through `PromptRenderingService` directly.

### Steps to Reproduce

1. Create a PROMPT_TEMPLATE artifact whose schema declares a variable with a `default`:

```yaml
templateId: tone
template: "Write in a {{tone}} tone."
variables:
  tone:
    type: string
    default: formal
```

2. Render it without supplying `tone`, e.g. `POST /groups/default/artifacts/tone/versions/1/render` with `{"variables":{}}`.

3. Observe that the rendered text still contains `{{tone}}` and that `validationErrors` is empty.

### Expected vs Actual Behaviour

Expected: `Write in a formal tone.`
Actual: `Write in a {{tone}} tone.`, with `validationErrors: []`.

Supplying `tone` explicitly renders correctly, so this is specific to the declared default being ignored rather than a substitution problem.

### Logs

Reproduced as a test written in the existing style of `PromptRenderingServiceTest`:

```
>>> rendered          = Write in a {{tone}} tone.
>>> validationErrors  = []

org.opentest4j.AssertionFailedError:
  expected: <Write in a formal tone.> but was: <Write in a {{tone}} tone.>

Tests run: 2, Failures: 1, Errors: 0, Skipped: 0
```

The second test in that run passes `tone` explicitly and succeeds.

---

## Impact

This is invisible in the Playground UI, because `PromptTemplateTestPanel` pre-fills its inputs from the defaults client-side and therefore always sends an explicit value. Every other consumer is affected: the REST endpoint, the generated SDKs, and the MCP `render_prompt_template` tool.

A template shipped in this repo already hits it, `examples/llm-artifact-types/sample-schemas/mcp-enabled-code-review-prompt.yaml`:

```yaml
[...]

PR #9309, opened 2026-08-06
fix(rest): apply declared variable defaults when rendering prompt templates

Aarti-panchal01:
Part of #9228. This covers the REST endpoint and the generated SDKs; the MCP path follows in a separate PR, so the issue stays open until all three consumers are covered.

## Problem

`prompt-template-v1.json` documents `default` as *"Default value if not provided"*, but `PromptRenderingService` only ever resolved variables against the caller-supplied map. A caller that omitted a variable with a declared default got the raw `{{placeholder}}` back in the rendered output — and because variables with defaults are typically `required: false`, `validationErrors` came back empty, so nothing signalled that the prompt was broken.

This is invisible in the Playground UI, which pre-fills its inputs from the declared defaults client-side and therefore always sends an explicit value. It affects the REST endpoint, the generated SDKs, and the MCP `render_prompt_template` tool.

An example shipped in this repo already hits it — `examples/llm-artifact-types/sample-schemas/mcp-enabled-code-review-prompt.yaml` declares `focus_area` with a default and no `required` flag, so an MCP client that omits it sends a literal `{{focus_area}}` to the model.

## Fix

`render()` now fills in declared defaults for any variable the caller did not provide, before substitution.

- Caller-supplied values always win; a variable that is explicitly present is never overwritten.
- The caller's map is never mutated — a copy is made only when there is actually a default to apply.
- Validation continues to run against the **caller-supplied** map, so a missing *required* variable is still reported even when the schema declares a default for it.

## Tests

Six tests added to `PromptRenderingServiceTest`:

| Test | Covers |
|---|---|
| `testDeclaredDefaultIsApplied` | the reported bug |
| `testCallerValueOverridesDefault` | precedence |
| `testNonStringDefaultsAreApplied` | non-string default values |
| `testVariableWithoutDefaultStillPreservesPlaceholder` | existing behaviour unchanged |
| `testRequiredVariableWithDefaultStillReportsMissing` | validation still reports missing required |
| `testDefaultsDoNotModifyCallerSuppliedMap` | no mutation of caller input |

Full class passes locally on JDK 21 — `Tests run: 40, Failures: 0, Errors: 0`.

## Notes

- Rebased onto `main` @ 23f1891; no conflicts.
- `PromptTemplateConverter.VariableSchema` already parses `default` into a field with a getter that is never called anywhere in the repo. I left it alone rather than widen this PR. Note the MCP server renders through that converter rather than through this service, so the MCP path needs its own change; see the comment below.
- Running these tests locally requires omitting `-Dlocal` (the `local-excluded-test-deps` profile drops the test-scoped serde deps, so `app` test sources won't compile with it): `cd app && ../mvnw test -Dtest=PromptRenderingServiceTest`.


app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -47,11 +48,16 @@ public RenderPromptResponse render(ContentHandle content, Map<String, Object> va
...
         JsonNode variablesSchema = templateNode.path("variables");
 
-        // Validate variables against schema
+        // Validate variables against schema. Validation intentionally runs against the
+        // caller-supplied variables, so that a missing required variable is still reported
+        // even when the schema declares a default for it.
         List<RenderValidationError> validationErrors = validateVariables(variables, variablesSchema);
 
+        // Fill in declared defaults for any variable the caller did not provide
+        Map<String, Object> effectiveVariables = applyDefaults(variables, variablesSchema);
```
paoloantinori, line 57: Validation runs on the caller-supplied `variables` but rendering uses `effectiveVariables`, so a declared `default` is substituted without being checked against the schema's `enum`/`type`/`pattern` constraints. Example: a variable with `enum: [formal, casual]` and `default: forma` (a typo) renders silently. The `required`+`default` interaction is handled deliberately in the test, but it is worth deciding whether defaults should be re-validated too. If yes, a separate validation pass over only the applied defaults (skipping `required`) avoids changing the required+default expectation that `testRequiredVariableWithDefaultStillReportsMissing` relies on.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -102,6 +108,43 @@ private String extractTemplateText(JsonNode templateNode) {
...
+        while (fields.hasNext()) {
+            Map.Entry<String, JsonNode> field = fields.next();
+            String varName = field.getKey();
+
+            if (variables.containsKey(varName)) {
+                continue;
+            }
+
+            JsonNode defaultNode = field.getValue().path("default");
+            if (defaultNode.isMissingNode()) {
```
paoloantinori: `isMissingNode()` does not catch a YAML `default:` declared with no value, which Jackson parses as a NullNode rather than a missing node. `convertValue(NullNode, Object.class)` returns null, which then gets substituted into the rendered prompt (or NPEs in `substituteVariables`, depending on how it stringifies null). Consider `if (defaultNode.isMissingNode() || defaultNode.isNull())` so a null default is treated like an absent one. An empty-string default (`default: ""`) should still apply, so scoping the skip to NullNode only is the right call.

app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
```diff
@@ -279,6 +279,142 @@ public void testValidArrayType() {
         Assertions.assertTrue(response.getRendered().contains("["));
     }
 
+    // ===== Default Value Tests =====
```
paoloantinori, line 282: Consider adding a case to this group for a null default (YAML `default:` with no value), asserting it is treated as absent and the `{{placeholder}}` is preserved rather than substituted with null. This pairs with the NullNode path in `applyDefaults` and would fail without the guard suggested on PromptRenderingService.java.

paoloantinori (commented):
Solid, well-scoped fix: declared defaults are now applied before substitution, caller-supplied values win, the caller map is never mutated, and validation still runs against the raw caller input. Two things are worth addressing before this lands: default values are substituted without re-validation against enum/type/pattern constraints, and a `default:` declared with a null value slips past the missing-node guard.

- `app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java:54-57`: validation runs on `variables` while rendering uses the defaults-applied `effectiveVariables`, so an out-of-enum or wrong-type default renders into the prompt without an error.
- `app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java:135`: `defaultNode.isMissingNode()` misses a YAML `default:` parsed as a NullNode; the resulting null is then substituted. Adding `|| defaultNode.isNull()` treats it like an absent default.
- Per the PR notes, `PromptTemplateConverter.VariableSchema` already parses `default` into an otherwise-unused getter; this change re-reads `default` straight from the JsonNode, leaving two parsing paths. A follow-up to consolidate would prevent drift.
- Tests are well-targeted with strict `assertEquals` on exact rendered output plus explicit precedence and no-mutation cases. No assertion relaxation.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
Aarti-panchal01: Confirmed, and thanks for catching it. I wrote the two cases as tests first and both failed against the unfixed code (`expected: <1> but was: <0>`), so a default violating `enum` or `type` really did render with no error reported.

Fixed the way you suggested. The type, enum and range checks are extracted out of `validateVariables` into a `validateValue` helper, and a new `validateAppliedDefaults` pass reuses it over only the variables the caller omitted. `required` is deliberately not re-checked there, since a caller omitting a required variable is already reported from the caller-supplied map, so `testRequiredVariableWithDefaultStillReportsMissing` keeps its existing expectation.

New tests: `testDefaultViolatingEnumIsReported` and `testDefaultViolatingTypeIsReported`.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
Aarti-panchal01: Guard added: `if (defaultNode.isMissingNode() || defaultNode.isNull())`, scoped to null only so `default: ""` still applies, as you noted.

One correction on the consequence, since I checked before changing it: the rendered output was already correct. `convertValue(NullNode, Object.class)` does return null, but `substituteVariables` looks the name up with `variables.get(varName)` and its `value == null` branch keeps the original placeholder, so a null default rendered as `{{placeholder}}` rather than as "null", and there was no NPE. I wrote `testNullDefaultIsTreatedAsAbsent` before adding the guard and it passed on the unfixed code.

So the guard is worth having for explicit intent, and it avoids copying the variables map for a default that cannot be used, but it does not change behaviour. I did not want to claim a fix for something that was not actually biting.

app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
Aarti-panchal01: Added `testNullDefaultIsTreatedAsAbsent` in this group, asserting the placeholder is preserved and no validation error is produced.

Worth flagging that it passes both with and without the guard, for the reason in the thread on `PromptRenderingService.java`: a null map value already fell through to the unresolved-placeholder branch in `substituteVariables`. So it documents the intended behaviour rather than catching a regression in the guard itself. I also added `testEmptyStringDefaultIsApplied` next to it, which does depend on the guard being scoped to null only, since a broader check would have wrongly skipped `default: ""`.

Aarti-panchal01:
Thanks for the review @paoloantinori. Both points are addressed in fcdd097, with replies in the individual threads.

**Defaults are now validated.** The type, enum and range checks moved out of `validateVariables` into a `validateValue` helper, and a new `validateAppliedDefaults` pass reuses it over only the variables the caller omitted. `required` is not re-checked there, so the required-plus-default expectation is unchanged. I wrote your `enum: [formal, casual]` / `default: forma` example as a test first and it failed on the unfixed code, so this was real.

**Null defaults are skipped.** `defaultNode.isNull()` is now treated like an absent default, scoped to null only so `default: ""` still applies. One correction there, in the interest of not overclaiming: the rendered output was already correct, because a null map value falls through to the unresolved-placeholder branch in `substituteVariables`. The guard makes the intent explicit and avoids copying the variables map for an unusable default, but it does not change behaviour. Details in that thread.

Four tests added, 44 passing locally on JDK 21:

- `testDefaultViolatingEnumIsReported`
- `testDefaultViolatingTypeIsReported`
- `testNullDefaultIsTreatedAsAbsent`
- `testEmptyStringDefaultIsApplied`

**On the two parsing paths.** Agreed, and I would rather not widen this PR to cover it. `PromptTemplateConverter.VariableSchema.getDefaultValue()` is currently dead code, so consolidating it is a behaviour-neutral refactor that reads better on its own. Happy to open a follow-up for it once this lands.

While reading that area I also found that the `mcp` block is never validated at all, which is filed separately as #9311 rather than folded in here.

paoloantinori:
Thanks @Aarti-panchal01 — both points are addressed on `fcdd097` and read cleanly:

- Defaults are now validated. The type/enum/range checks moved into `validateValue`, and `validateAppliedDefaults` reuses it for the subset the caller omitted. `required` is correctly not re-checked there.
- The `enum: [formal, casual]` / `default: forma` case is covered by `testDefaultViolatingEnumIsReported` (plus `testDefaultViolatingTypeIsReported` for the integer variant), so a bad default is reported rather than silently rendered.

Nothing further from me.

Aarti-panchal01:
One correction to my PR description, before this merges.

The description says I left `PromptTemplateConverter`'s `default` parsing alone because "the MCP path renders through this same service and is fixed by the change here". That is wrong. I assumed it rather than checking it.

The MCP server does not use `PromptRenderingService` at all. It reaches `PromptTemplateConverter.renderTemplate(String, Map)`, which receives only the template text and the caller's arguments, so it never sees the `variables` schema and cannot apply a declared default. All three entry points are affected: `get_registry_prompt` calls `renderTemplate` directly, and `render_prompt_template` and `render_registry_prompt` both go through `parseAndRenderAutoDetect`, which calls the same method.

So this PR fixes the REST endpoint and the generated SDKs, but not the MCP `render_prompt_template` tool, which #9228 also lists as affected. As written, merging this auto-closes #9228 while one of the three consumers named in it is still broken.

The MCP fix is ready as a follow-up. It wires up `VariableSchema.getDefaultValue()`, which is currently parsed and then never read, so the two parsing paths get consolidated as @paoloantinori suggested instead of a third being added. Same semantics as here: caller-supplied values win, the caller's map is not mutated, and a default declared with no value is treated as absent. It also adds the first test coverage for `PromptTemplateConverter`. I will open it once this lands, per the one-open-PR rule.

@EricWittmann, how would you prefer to handle the issue link? I am happy to drop `Fixes #9228` from this PR so the issue stays open until the MCP half lands, or to leave it as is and track the remainder separately. Either works, just say which and I will adjust.

paoloantinori:
Thanks for the correction, and for flagging it before merge rather than after. You're right: `PromptTemplateConverter.renderTemplate(String, Map)` only sees the caller's arguments, never the `variables` schema, so it can't apply a declared default. The MCP `render_prompt_template` and `render_registry_prompt` paths are genuinely uncovered by this PR, and auto-closing #9228 via `Fixes` would leave that gap silent.

On the issue link: I'd drop `Fixes #9228` here and let the follow-up (the one wiring up `VariableSchema.getDefaultValue()`) carry it, so the issue stays open until all three consumers are covered. Consolidating onto the already-parsed-but-unread `getDefaultValue()` is the right shape — it avoids a third parsing path and keeps the two renderers consistent.

The REST/SDK half here is good to merge on its own.

Aarti-panchal01:
Thanks @paoloantinori. Done: the description now says `Part of #9228` instead of `Fixes #9228`, so merging this will not close the issue, and #9228 stays open until the MCP path is covered too.

The follow-up is ready to go on that basis and I will open it once this lands.

paoloantinori:
Thanks for making that change — `Part of #9228` is the right framing, and the MCP-path split keeps the issue open correctly. Nothing else from me on this one.

Aarti-panchal01:
@EricWittmann could you re-run `/accept` here? Your original one on Aug 6 never applied — the `Command` job that processes it died on a GitHub Actions outage before it could run, not on anything in this PR:

```
Failed to resolve action download info. Error: Service Unavailable
```

(https://github.com/Apicurio/apicurio-registry/actions/runs/31116541240, retried twice then gave up.)

So the PR is still on `lifecycle/new`. `Decide` reads that as `run_tests=false`, skips all ten test jobs, and `Verification Gate` then fails with `PR is not in a testable lifecycle state` — not a real test failure. My `/retry` could not fix it either, since Verify reads the lifecycle label but cannot set it.

Review is settled: @paoloantinori confirmed both of his points are addressed, and SonarCloud is green. The PR is in `lifecycle/new`, which is what `/accept` expects, so it should apply cleanly.

paoloantinori:
Happy to confirm the review is settled from my side — both points I raised are addressed at `fcdd097`, and the `Part of #9228` framing keeps the MCP path open correctly. Nothing else from me; ready for merge once the lifecycle label and CI are sorted.

EricWittmann:
Thanks for the excellent contribution, @Aarti-panchal01! The fix is clean, well-tested, and the issue write-up was thorough. Looking forward to the MCP follow-up PR.

Aarti-panchal01:
Thank you @EricWittmann, that means a lot.

The MCP follow-up is now open as #9392. It wires up `VariableSchema.getDefaultValue()`, which this class already parsed but never called anywhere, and adds a `renderTemplate(PromptTemplate, Map)` overload so that `parseAndRenderAutoDetect` and the server call site are both covered by one change. That keeps all three MCP entry points on a single fix point instead of introducing a second parsing path. It also adds `PromptTemplateConverterTest`, the first test coverage this class has had.

One small bookkeeping note. Merging this PR closed #9228 automatically. The description here says `Part of #9228` rather than `Fixes`, but the linked issue in the sidebar was still attached from when the PR was first opened, and that is what closed it. I tried to reopen it and GitHub will not let me, since I was not the one who closed it. Could you reopen it when you get a moment? #9392 carries `Fixes #9228`, so merging that will close it correctly once the remaining consumer is covered.

PR #9323, opened 2026-08-06
fix(ci): auto-approve Verify workflow runs for fork PRs

EricWittmann:
## Summary

Fork PRs use the `pull_request` event trigger for the Verify workflow, which requires manual
approval before GitHub Actions will execute. The PR lifecycle orchestrator (which uses
`pull_request_target`) manages labels and announces "full test suite will run now" on `/accept`,
but cannot grant this approval — leaving Verify runs stuck in `action_required` status
indefinitely. This means CI tests never run for external contributor PRs unless a maintainer
manually approves each run from the Actions tab.

This PR wires the GitHub Actions
[approve endpoint](https://docs.github.com/en/rest/actions/workflow-runs#approve-a-workflow-run-for-a-fork-pull-request)
into the lifecycle orchestrator at every point where a fork PR should start running tests:

- **`/accept`** — approves the existing pending run, then sweeps for any label-triggered runs
  that race with the approval
- **`/retry`** — recognizes `action_required` status (previously misreported as "already
  running") and approves all pending runs
- **`handlePrSynchronize`** — approves pending runs after new pushes to accepted fork PRs
- **`retriggerVerify`** — handles `action_required` status by approving instead of re-running
  (which would just create another unapproved run)

## Changes

Single file: `.github/scripts/pr-lifecycle.js` (+74 lines)

- Two new API methods: `approveWorkflowRun(runId)` and `findPendingApprovalVerifyRuns(headSha)`
- New `approvePendingVerifyRuns` helper that sweeps and approves all pending Verify runs for a
  SHA, with per-run error handling
- `action_required` handling in `retriggerVerify` (approve instead of re-run)
- Approval sweep in `cmdAccept` after label transition
- `action_required` branch in `cmdRetry` with informative comment
- Approval sweep in `handlePrSynchronize` for PRs in a testable lifecycle state

## Risk

The `GITHUB_TOKEN` in the PR Lifecycle workflow already has `actions: write` permission, which
is what the approve endpoint requires. Some org-level policies could restrict this — if approval
fails, the error is caught and logged as a warning (existing behavior continues unchanged).

## Test plan

- [ ] Open a PR from a fork, `/accept` it, verify the Verify workflow starts without manual
      approval from the Actions tab
- [ ] On the same fork PR, push a new commit and verify the new Verify run is auto-approved
- [ ] Use `/retry` on a fork PR with a stuck `action_required` run and verify it gets approved
- [ ] Verify non-fork PRs (from maintainers) are unaffected — no approval needed, existing
      flow unchanged

### DIFF
--- a/.github/scripts/pr-lifecycle.js
+++ b/.github/scripts/pr-lifecycle.js
@@ -287,7 +287,8 @@ async function retriggerVerify(api, pr, core, { waitForRun = false } = {}) {
   // Fork PRs need workflow approval before they can run. Approve instead
   // of re-running — the Decide step fetches current labels from the API,
   // so the approved run will see the up-to-date lifecycle state.
-  if (run.status === 'action_required') {
+  // GitHub represents these as status=completed, conclusion=action_required.
+  if (run.conclusion === 'action_required') {
     try {
       await api.approveWorkflowRun(run.id);
       core.info(`PR #${pr.number} approved pending Verify run ${run.id}`);
@@ -1010,7 +1011,7 @@ async function cmdRetry(github, api, core, pr, actor, isAuthor, maintainer, comm
 
   // Check if the latest Verify run failed, was cancelled, or needs approval
   const latestRun = await api.findLatestVerifyRun(freshPr.head.sha);
-  if (latestRun && latestRun.status === 'action_required') {
+  if (latestRun && latestRun.conclusion === 'action_required') {
     const approved = await approvePendingVerifyRuns(api, freshPr, core);
     await api.postComment(pr.number,
       `Retrying: reconciled PR state and approved ${approved} pending Verify ` +
--- a/.github/scripts/pr-lifecycle.js
+++ b/.github/scripts/pr-lifecycle.js
@@ -248,6 +248,21 @@ function createApi(github, owner, repo) {
       });
       return data;
     },
+
+    approveWorkflowRun: async (runId) => {
+      await github.request('POST /repos/{owner}/{repo}/actions/runs/{run_id}/approve', {
+        owner, repo, run_id: runId,
+      });
+    },
+
+    findPendingApprovalVerifyRuns: async (headSha) => {
+      const { data } = await github.rest.actions.listWorkflowRuns({
+        owner, repo, workflow_id: 'verify.yaml',
+        head_sha: headSha, status: 'action_required',
+        per_page: 10,
+      });
+      return data.workflow_runs;
+    },
   };
 }
 
@@ -266,6 +281,19 @@ async function retriggerVerify(api, pr, core, { waitForRun = false } = {}) {
 
   if (!run) {
     core.warning(`PR #${pr.number} no Verify run found for ${pr.head.sha}, skipping re-trigger`);
+    return;
+  }
+
+  // Fork PRs need workflow approval before they can run. Approve instead
+  // of re-running — the Decide step fetches current labels from the API,
+  // so the approved run will see the up-to-date lifecycle state.
+  if (run.status === 'action_required') {
+    try {
+      await api.approveWorkflowRun(run.id);
+      core.info(`PR #${pr.number} approved pending Verify run ${run.id}`);
+    } catch (e) {
+      core.warning(`PR #${pr.number} failed to approve Verify run ${run.id}: ${e.message}`);
+    }
     return;
   }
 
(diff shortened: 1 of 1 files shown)


.github/scripts/pr-lifecycle.js
```diff
@@ -248,6 +248,21 @@ function createApi(github, owner, repo) {
...
+    approveWorkflowRun: async (runId) => {
+      await github.request('POST /repos/{owner}/{repo}/actions/runs/{run_id}/approve', {
+        owner, repo, run_id: runId,
+      });
+    },
+
+    findPendingApprovalVerifyRuns: async (headSha) => {
+      const { data } = await github.rest.actions.listWorkflowRuns({
+        owner, repo, workflow_id: 'verify.yaml',
+        head_sha: headSha, status: 'action_required',
```
paoloantinori, line 265: The 'needs approval' condition is detected inconsistently across this change. Here you filter `listWorkflowRuns` by `status: 'action_required'`, but the new branches in `retriggerVerify` (line 291) and `cmdRetry` (line 1014) check `run.conclusion === 'action_required'`, and the comment at line 290 states these runs have `status=completed, conclusion=action_required`. If the `status` filter does not match a run whose `conclusion` is `action_required`, this query returns zero runs and every approval sweep (`/accept`, synchronize, `/retry`) becomes a silent no-op, leaving Verify stuck in exactly the state this PR is meant to fix. Please confirm the live run representation (status vs conclusion) against the API and unify detection on whichever field GitHub populates, and log when zero runs are found so a mismatch is visible instead of silent.

.github/scripts/pr-lifecycle.js
```diff
@@ -294,6 +323,27 @@ async function retriggerVerify(api, pr, core, { waitForRun = false } = {}) {
...
+// to catch label-triggered runs that race with retriggerVerify.
+async function approvePendingVerifyRuns(api, pr, core) {
+  try {
+    const runs = await api.findPendingApprovalVerifyRuns(pr.head.sha);
+    for (const run of runs) {
+      try {
+        await api.approveWorkflowRun(run.id);
+        core.info(`PR #${pr.number} approved pending Verify run ${run.id}`);
+      } catch (e) {
+        core.warning(`PR #${pr.number} failed to approve Verify run ${run.id}: ${e.message}`);
```
paoloantinori, line 347: `approvePendingVerifyRuns` returns `runs.length`, which is the number of runs found, not the number actually approved. Approvals that fail (permissions, 409 already-approved, rate limit) are caught per-run and logged as a warning, so callers can report 'approved N' when fewer than N succeeded. In `cmdRetry` (line 1015) this count is posted as a public PR comment ('approved N pending Verify workflow runs'), so it can overstate success to maintainers. Track successful approvals in a counter and return that instead.

.github/scripts/pr-lifecycle.js
```diff
@@ -666,6 +716,17 @@ async function handlePrSynchronize({ github, context, core }) {
 
   const freshPr = await api.getPr(pr.number);
   await reconcile(github, api, freshPr, core);
+
+  // For fork PRs in a testable state, the synchronize event creates a new
+  // Verify run that needs approval. Wait for it to appear, then approve.
+  const currentState = getLifecycleState(freshPr);
+  if (currentState === LABELS.READY_FOR_REVIEW || currentState === LABELS.READY_TO_MERGE) {
```
paoloantinori, line 733: A fixed 5s wait followed by a single lookup is racy: if GitHub creates the new Verify run more than roughly 5s after the push (or after the label change in `cmdAccept` at line 841), it will not appear in `findPendingApprovalVerifyRuns` and is never approved, so it stays stuck in `action_required` just as before. Consider a bounded poll that retries the lookup a few times with backoff, stopping once a pending run appears or a timeout elapses.

paoloantinori (commented):
Approach is sound, but the core approval mechanism is unverified and has two robustness gaps that could still leave fork PRs stuck in `action_required`. The biggest risk is that the 'needs approval' state is detected three different ways that may not agree, which can make the whole feature fail silently; the secondary risks are a racy fixed sleep and a return value that overstates how many runs were actually approved.

- `.github/scripts/pr-lifecycle.js:261` - `findPendingApprovalVerifyRuns` filters by `status: 'action_required'`, while `retriggerVerify` (line 291) and `cmdRetry` (line 1014) check `conclusion === 'action_required'`, and the comment at line 290 says the run's `status` is `completed`. Verify the actual run representation against the API and unify detection, otherwise the sweeps are silent no-ops.
- `.github/scripts/pr-lifecycle.js:337` - `approvePendingVerifyRuns` returns `runs.length` (found) rather than the count actually approved, so logs and the `cmdRetry` PR comment (line 1015) can overstate success when individual approvals fail.
- `.github/scripts/pr-lifecycle.js:723` and `:841` - the fixed 5s sleep then single lookup is racy; if the run lands later it is never approved. Use a bounded poll instead.
- Security posture: auto-approving removes GitHub's fork-PR approval gate, so the trust boundary rests entirely on maintainer `/accept` (and the fact that accepted fork PRs already run full CI with secrets). Confirm that is the intended model and that `actions: write` is not blocked by an org policy; failures being caught and logged (as the summary notes) is good.
- Coverage: the new branches are covered only by the manual test plan. Consider a unit test that stubs the API responses, including the found-vs-approved count and the zero-runs case.

.github/scripts/pr-lifecycle.js
EricWittmann: Good catch on the inconsistency. The GitHub API has a quirk here: the `status` query parameter accepts `action_required` as a filter value even though the run object itself stores it in the `conclusion` field (with `status: "completed"`). We verified this empirically — the query returns the correct runs.

Addressed in e6b63151a:
- Added a comment on `findPendingApprovalVerifyRuns` documenting this API behavior
- Added logging when zero runs are found, so a mismatch between the query filter and the actual run representation would be visible in the workflow logs rather than silent

.github/scripts/pr-lifecycle.js
EricWittmann: Good point — fixed in e6b63151a. `approvePendingVerifyRuns` now tracks a separate `approved` counter that only increments on success, and returns that instead of `runs.length`. The `/retry` PR comment and log messages now accurately reflect how many runs were actually approved.

.github/scripts/pr-lifecycle.js
EricWittmann: Acknowledged — the fixed sleep is racy in theory. In practice the same 5s pattern is used elsewhere in this file (e.g. `retriggerVerify` line 315) and if the run arrives late there are multiple fallback paths: the next push triggers `handlePrSynchronize` again, `/retry` is available manually, and the stale reconciler runs every 6 hours. Not addressing this in this PR but could revisit with a bounded poll if we see missed approvals in production.

PR #9328, opened 2026-08-06
ci: migrate app-auth/app-transport/app-metrics shards to JUnit @Tag groups

### LINKED ISSUE #9302: refactor(test): migrate unit test shards from -Dtest= patterns to JUnit @Tag groups
## Context

PR #9287 splits the `app-other` unit test shard to fix OOM. The sharding uses `-Dtest=` package patterns in the workflow YAML, with `verify-test-shards.py` reverse-engineering surefire's matching to verify the partition.

## Proposal

Migrate to JUnit `@Tag` groups for shard assignment:

1. Expand `ApicurioTestTags` with semantic group constants (`AUTH`, `TRANSPORT`, `METRICS`, `REST`, `CCOMPAT`, `STORAGE_SQL`, etc.) — following the pattern already used in `operator/Tags.java`.
2. Tag each test class with its group (`@Tag(ApicurioTestTags.AUTH)`).
3. Workflow shards select by `<groups>` instead of `-Dtest=` patterns.
4. `verify-test-shards.py` simplifies to checking that every surefire-eligible test has a recognized `@Tag`.

## Why

- Single source of truth: the mapping lives on the test class, not in YAML patterns
- New tests get their group at creation time — no risk of silently landing in the wrong shard
- Semantic tags document the test's purpose for developers
- Follows the operator module's existing convention

## Scope

~250 test classes to tag. Can be done incrementally — start with the new shards (auth, transport, metrics), backfill existing ones (rest, ccompat, storage) later.

See review comment: https://github.com/Apicurio/apicurio-registry/pull/9287#pullrequestreview-4874910816

Priyanshubhartistm:
## Summary

Migrates the `app-auth`, `app-transport`, and `app-metrics` unit-test shards (added in #9287) from `-Dtest=` package patterns to JUnit `@Tag` groups, per the review comment on #9287.

Scope is intentionally incremental, as called out in the issue: only the three newest shards move to tags in this PR. `app-rest`, `app-ccompat`, and the storage-variant shards stay on package patterns and can be backfilled in follow-up PRs.

## Why

- Single source of truth: the shard mapping lives on the test class  (`@Tag(ApicurioTestTags.AUTH)`), not in YAML package patterns.
- New tests get their group at creation time instead of silently landing  in `app-other` because nobody remembered to add a pattern.
- Follows the convention already used in `operator/controller/.../Tags.java`.

## A note on `-Dtest=` vs. the reviewer's exact snippet

The review comment on #9287 proposed:

```yaml
- name: app-auth
  test-filter: "-Dgroups=auth"
```

I tested this exact snippet in isolation (same Surefire/JUnit versions, plain JUnit 5, no Quarkus). Without a `-Dtest=` pattern, Surefire falls back to its default file-name discovery (`Test*`/`*Test`/`*Tests`/
`*TestCase`) *before* the `@Tag` filter ever runs. A handful of real test classes in `auth`/`rbac`/`limits` don't match that convention (e.g.
`AuthTestNoRoles`, `AuthTestAnonymousCredentials`,
`AuthTestAuthenticatedReadAccess`, `AuthTestLocalRoles`,
`AuthTestProfileBasicClientCredentials`) — with the literal snippet they would be silently dropped from CI entirely, which is exactly the kind of silent-orphan bug this migration is meant to eliminate. (Operator's `Tags.java` doesn't hit this because every operator test class happens to follow the naming convention.)

So each tag shard keeps `-Dtest=io.apicurio.registry.**` alongside
`-Dgroups=`, purely to widen Surefire's file-discovery net to every class in the app test tree — the actual selection is still done entirely by the tag, not the pattern. `app-other`'s exclusion filter is unchanged and still excludes these packages by name, so an untagged class added to one of them shows up as an orphan instead of running zero or two times.

## Changes

- **`ApicurioTestTags.java`** — added `AUTH`, `TRANSPORT`, `METRICS` constants.
- **53 test classes** in `auth`/`rbac`/`limits`, `tls`/`headers`/`cors`, and `metrics`/`search` — tagged with the matching group.
- **`verify-unit-tests.yaml`** — `app-auth`/`app-transport`/`app-metrics` now select with `-Dgroups=<value>` instead of a package pattern (see
  note above on why `-Dtest=io.apicurio.registry.**` stays alongside it).
- **`verify-test-shards.py`** — extended to resolve tag-shard membership via `@Tag(ApicurioTestTags.*)` scanning (cross-referenced against the constant's string value) alongside the existing pattern-matching logic,
  keeping the "every class claimed exactly once" check intact across both shard types.
- **`workflows/README.md`** — documented the tag-based mechanism and the
  new failure mode (untagged class in a tag-covered package = orphan).

## Test plan

- [x] `python3 .github/scripts/verify-test-shards.py`: 260 app test
      classes, all claimed exactly once, 0 orphans, 0 duplicates
- [x] Isolated JUnit5 experiment confirming `-Dtest=<wildcard> -Dgroups=X` picks up a class that fails Surefire's default naming convention, when tagged, and correctly excludes it when untagged
- [x] Isolated experiment confirming the reviewer's literal
      `-Dgroups=X`-only snippet (no `-Dtest=`) silently drops such a class, motivating the `-Dtest=io.apicurio.registry.**` override
- [x] Real `app-metrics` shard run end-to-end via
      `-Dtest=io.apicurio.registry.metrics.** -Dgroups=metrics`: all 8 classes pass
- [x] `./mvnw test-compile -pl app,utils/tests -am` clean
- [x] `./mvnw checkstyle:check -pl app,utils/tests` clean
- [x] `./scripts/validate-files.sh` clean

**Storage variants:** N/A — test categorization and CI configuration only, no storage or product code changed.


Priyanshubhartistm:
@EricWittmann, could you please take a look when you get a chance and let me know if any changes are needed? Thank you!

Priyanshubhartistm:
@jsenko, if possible, could you please reopen this PR? Or would you prefer that I open a new PR for it?

jsenko:
Reopened!

Priyanshubhartistm:
@jsenko, could you please take a look at the failing checks? I'm not sure whether the failures are caused by something on my end. If everything looks fine from your side and any changes are needed, please let me know what I should update. Thank you!

jsenko:
Can you check what the failing workflow does? 

```
!! ORPHANS - these run in no shard and fail silently:
     io.apicurio.registry.auth.AppAuthenticationMechanismTest
     io.apicurio.registry.auth.AuthorizedInterceptorTest
     io.apicurio.registry.auth.FormAuthTest
     io.apicurio.registry.auth.OidcAuthenticationStrategyTest

```

This looks related to your PR, the error suggests that some tests are not being run.

jsenko:
The `Build Java Application` failure on this PR is **not caused by your changes** — it's a build infrastructure race condition in the `maven-antrun-plugin` during parallel builds. The plugin resolves the full project classpath just to create a directory, and in a parallel build the `schema-util-iceberg` module can start before its `test-jar` dependency is available.

Fix: #9689 — replaces `maven-antrun-plugin` with `maven-resources-plugin`. Once merged, rebasing this PR should resolve the failure.

paoloantinori (commented):
Solid migration overall: the tag-per-class model, the retained `-Dtest=io.apicurio.registry.**` escape hatch for Surefire's name-based discovery, and the extended verifier are all the right shape, and the PR description's reasoning about `-Dgroups`-only silently dropping non-convention class names is correct. The main gap is that the verifier and the README both promise orphan detection that does not actually cover the exact failure mode this PR was motivated by: an untagged class with a non-conventional name in a tag-covered package is skipped by `enumerate_classes` entirely, so it runs in zero shards with the verifier reporting a clean partition.
- `.github/scripts/verify-test-shards.py:65` skips any class that is neither surefire-named nor group-tagged, so e.g. a new untagged `AuthTestFoo`-style class in `auth` is invisible to the partition check (excluded from `app-other` by package, dropped by the `app-auth` tag filter, not enumerated). Zero-run, zero-report.
- `.github/workflows/README.md:127` documents "untagged class in a tag-covered package = orphan" as the failure mode the verifier catches, which overstates what the script can see per the point above.
- `.github/workflows/verify-unit-tests.yaml:59` now packs two space-separated `-D` flags into one `test-filter` value; worth confirming the workflow injects it unquoted (or split) so Maven receives two arguments rather than one mangled `-Dtest=... -Dgroups=auth` property.
- `.github/scripts/verify-test-shards.py:135` (per-shard count loop) rebuilds a single-shard list per class per shard; computing `class_claims` once and counting membership would be simpler and O(n) instead of O(n*k).

.github/scripts/verify-test-shards.py
```diff
@@ -38,19 +45,35 @@ def is_surefire_name(stem):
...
+    found = {}
     for path in APP_TESTS.rglob("*.java"):
-        stem = path.stem
-        if not is_surefire_name(stem):
-            continue
         source = path.read_text(encoding="utf-8", errors="replace")
-        # Abstract classes are bases for other tests; surefire never runs them.
-        if re.search(r"\babstract\s+class\s+" + re.escape(stem) + r"\b", source):
+        if is_abstract(path, source):
+            continue
```
paoloantinori, line 120: This skip condition re-opens the silent-orphan hole the migration is meant to close. A class in `auth`/`rbac`/`limits`/`tls`/`headers`/`cors`/`metrics`/`search` that is neither convention-named nor group-tagged is dropped from `classes` here, so it is never checked: `app-other` excludes its package, no tag shard selects it, and the verifier reports 0 orphans while the class runs zero times in CI. This is exactly the `AuthTestNoRoles` scenario called out in the PR description. Suggestion: enumerate every non-abstract class regardless of name or tags, then report as orphans (a) unclaimed classes in general and specifically (b) classes sitting in a package covered by a tag shard's original package mapping that carry no group tag.

.github/workflows/README.md
```diff
@@ -103,9 +120,12 @@ subpackages are an exception. Read this before adding one:
...
   already covered by one of the storage-variant include filters
   (`app-sql`, `app-kafkasql`, `app-gitops`, `app-kubernetesops`) runs there.
-  `auth`/`rbac`/`limits`, `tls`/`headers`/`cors`, and `metrics`/`search` run in
-  `app-auth`, `app-transport`, and `app-metrics` respectively.
-  Everything else lands in `app-other` (the exclusion-based shard).
+  A test in `auth`/`rbac`/`limits`, `tls`/`headers`/`cors`, or
+  `metrics`/`search` needs `@Tag(ApicurioTestTags.AUTH)`, `TRANSPORT`, or
+  `METRICS` respectively to run in `app-auth`, `app-transport`, or
+  `app-metrics` — without the tag it is excluded from those packages by
+  `app-other` but still won't match a tag shard, so it runs in **zero**
```
paoloantinori, line 127: This paragraph states an untagged class in `auth`/`tls`/`metrics` etc. "runs in zero shards" and is caught as an orphan by `verify-test-shards.py`. As written, the script only enumerates classes that are surefire-convention-named or already tagged, so an untagged class with a non-convention name (the `AuthTestNoRoles` pattern) runs zero shards AND produces no orphan report. Either fix the script's enumeration (see the comment on the script) or soften this doc to say the orphan check only covers convention-named classes.

.github/workflows/verify-unit-tests.yaml
```diff
@@ -47,19 +47,25 @@ jobs:
...
+          # See issue #9265.
+          #
+          # Unlike the other app-* shards, these three select by JUnit @Tag (ApicurioTestTags.AUTH /
+          # TRANSPORT / METRICS) instead of a -Dtest= package pattern: the tag lives on the test class
+          # itself, so a new test gets the right shard at creation time instead of silently landing in
+          # app-other. -Dtest=io.apicurio.registry.** makes surefire consider every class in the app test
+          # tree as a tag-filter candidate, since surefire's default file-name discovery (Test*/*Test/
+          # *Tests/*TestCase) would otherwise skip classes named like AuthTestNoRoles. See issue #9302.
           - name: app-auth
             modules: "-pl app -am -Dfull -DcliSkipNative"
```
paoloantinori, line 94: `test-filter` now holds two space-separated flags (`-Dtest=io.apicurio.registry.** -Dgroups=auth`). If the run step interpolates this as a single quoted argument (`"$TEST_FILTER"` or `mvn ${{ matrix.test-filter }}` passed as one token), Maven receives one malformed `-D` property and both the test pattern and the group filter silently no-op, which can turn into a shard running zero tests. Please confirm the invocation site word-splits this value (the PR's local runs may have used a shell while CI uses a different quoting), or switch to two separate matrix keys.

.github/scripts/verify-test-shards.py
```diff
@@ -90,43 +115,76 @@ def parse_shards():
...
     return [s for s in shards if s[0] != NON_APP_SHARD]
 
 
+def claims(fqn, source, shards, const_by_value):
+    hits = []
+    for name, kind, selector in shards:
+        if kind == "pattern":
+            if matches(fqn, selector):
+                hits.append(name)
+        else:
```
paoloantinori, line 208: Minor simplification while here: this per-shard count wraps each shard in a fresh list and calls `claims()` for every class x shard, duplicating the `class_claims` computation just below. Compute `class_claims` once, then derive the table with something like `Counter(name for hits in class_claims.values() for name in hits)`; same output, one pass, and the two views can't drift.

.github/scripts/verify-test-shards.py
Priyanshubhartistm: done

Priyanshubhartistm:
@jsenko  Could you please take a look and see why the branch keeps getting outdated? If any changes are needed from my side, please let me know, and I’ll make the necessary updates.

Priyanshubhartistm:
@paoloantinori , @jsenko Could you please take a look and see why the branch keeps getting outdated? If any changes are needed from my side, please let me know, and I’ll make the necessary updates.

carlesarnal:
I'm sorry, but we can't allow CI changes to external contributors. I encourage you to work on other issues.

Priyanshubhartistm:
@carlesarnal Got it, thanks for letting me know.

PR #9331, opened 2026-08-07
fix(rules): detect removal of agent card capability extensions

### LINKED ISSUE #9173: AgentCardCompatibilityChecker skips non-boolean capabilities, so removing a declared extension passes BACKWARD
## Description

**Registry Version**: snapshot (build from `main`, `3.3.2-SNAPSHOT`)
**Persistence type**: sql (defect is in the compatibility checker; storage-independent)

`AgentCardCompatibilityChecker` documents:

> - Removing/disabling capabilities: Backward incompatible

but `checkCapabilityRemovals(...)` only compares capabilities whose value is a boolean:

    if (!existingCaps.get(capName).isBoolean()) {
        continue;
    }

`capabilities.extensions` is defined as an **array** in `a2a-agent-card-schema.json`, so it is
skipped entirely. Removing a declared extension passes a `BACKWARD` compatibility rule, even
though a consumer relying on that extension is broken.

### Environment

Reproduced against a running registry (`quay.io/apicurio/apicurio-registry:latest-snapshot`)
over REST.

`schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java`

### Steps to Reproduce

1. Register an `AGENT_CARD` whose capabilities declare an extension:

       "capabilities": { "extensions": [ { "uri": "https://example.com/ext/translate" } ] }

2. Set a compatibility rule on the artifact:

       curl -X POST "$BASE/groups/a2a-demo/artifacts/my_agent/rules" \
         -d '{"ruleType":"COMPATIBILITY","config":"BACKWARD"}'

3. Register a new version identical except that the extension is removed:

       "capabilities": {}

Result: **HTTP 200**, the new version is created.

**Control (proves the rule is active on the same artifact):** on the same artifact with the same
rule, removing a skill *is* correctly rejected:

    HTTP 400 RuleViolationException
      "description": "Skill 'forecast' was removed", "context": "/skills"

So compatibility checking is enabled and running — it just never inspects non-boolean capability
values.

**Expected:** removing a declared capability extension is reported as backward incompatible,
consistent with the documented rule "Removing/disabling capabilities: Backward incompatible".

**Actual:** zero differences are reported and the version is accepted.

### Logs

Equivalent unit probe in `AgentCardCompatibilityCheckerTest` fails against current `main`:

    [ERROR] AgentCardCompatibilityCheckerTest.probeCapabilityExtensionRemoval
      Removing a declared capability extension should be backward incompatible
      ==> expected: <false> but was: <true>

---

### Two related questions (possibly intentional — happy to be told so)

While reproducing the above I noticed two other cases that also pass a `BACKWARD` rule. Both stem
from the same pattern — skills are compared by `id` only, never by content — but unlike the
capability case above, I'm not sure either is considered in scope, so raising them as questions
rather than defects:

1. **Per-skill `inputModes` / `outputModes`.** The schema allows skills to declare their own
   modes, but `checkModeRemovals(...)` is only applied to the top-level `defaultInputModes` /
   `defaultOutputModes`. Narrowing a skill from `["text","image"]` to `["text"]` is accepted. Is
   the documented "Removing input/output modes" rule intended to cover per-skill modes, or only
   the card-level defaults?

2. **Per-skill `securityRequirements`.** Adding `"securityRequirements": [ { "oauth": ["admin"] } ]`
   to an existing skill is accepted, which changes the auth expectations for existing consumers.
   The checker covers top-level `securitySchemes`; should per-skill requirements be covered too?

Happy to submit a PR for the capability-extensions fix, and to fold in either of the above if
you'd like them addressed.

malladinagarjuna2:
Fixes #9173

checkCapabilityRemovals iterates the capabilities object and skips any field
whose value is not a boolean. capabilities.extensions is declared as an array
in a2a-agent-card-schema.json, so it was skipped entirely and removing a
declared extension produced no differences at all -- the new version was
accepted under a BACKWARD rule, contradicting the checker's own documented
"Removing/disabling capabilities: Backward incompatible".

Compare extension sets by uri, which is the identifier the schema gives them,
and report any that disappear. The boolean scan is left alone: it still covers
streaming, pushNotifications, extendedAgentCard and the schema's boolean
additionalProperties.

Report these under a new CAPABILITY_EXTENSION_REMOVED difference type so the
violation points at /capabilities/extensions rather than the generic
/capabilities.

Adding an extension, editing an existing extension's description, and an
extension declared without a uri all remain compatible.


schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityDifference.java
```diff
@@ -14,6 +14,7 @@ public enum Type {
         PROTOCOL_VERSION_CHANGED("supportedInterfaces"),
         SKILL_REMOVED("skills"),
         CAPABILITY_REMOVED("capabilities"),
+        CAPABILITY_EXTENSION_REMOVED("capabilities/extensions"),
```
paoloantinori, line 17: The new constant flows through `asRuleViolation()` correctly (the tests assert both the description and the `/capabilities/extensions` context), so the read path is covered. Worth a quick grep for any other exhaustive `switch` over `Type` (severity mapping, serialization, visitors) that now needs a `CAPABILITY_EXTENSION_REMOVED` case: an unhandled enum constant compiles fine and only fails at runtime in a path the unit tests do not exercise.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java
```diff
@@ -223,6 +245,23 @@ private Set<String> extractSecuritySchemeNames(JsonNode node) {
...
+    private Set<String> extractExtensionUris(JsonNode capabilities) {
+        Set<String> uris = new HashSet<>();
+        if (capabilities == null) {
+            return uris;
+        }
+        JsonNode extensions = capabilities.get("extensions");
+        if (extensions != null && extensions.isArray()) {
+            for (JsonNode extension : extensions) {
+                String uri = getTextValue(extension, "uri");
+                if (uri != null) {
```
paoloantinori: `extractExtensionUris` is lenient here. A missing `uri` is skipped (covered by `testCapabilityExtensionWithoutUriIsIgnored`), but a present non-text `uri` and an empty-string `uri` are not. If `getTextValue` returns null on a non-text node it is silently skipped, but an empty `""` would be tracked as a real identity and surface as `Capability extension '' was removed`. Either skip blank/non-text uris explicitly, or add a test pinning the current behavior, so malformed extension entries in registry content do not surprise.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityCheckerTest.java
```diff
@@ -329,4 +329,135 @@ void testBackwardIncompatibleProtocolVersionChange() {
...
+                "Removing a declared capability extension should be backward incompatible");
+        assertTrue(result.getIncompatibleDifferences().stream()
+                .anyMatch(d -> d.asRuleViolation().getDescription()
+                        .contains("https://example.com/ext/translate")),
+                "The violation should name the removed extension uri");
+    }
+
+    @Test
+    void testBackwardIncompatibleRemovingOneOfSeveralCapabilityExtensions() {
+        String existing = cardWithExtensions(TRANSLATE_EXTENSION + "," + SUMMARISE_EXTENSION);
```
paoloantinori, line 366: This pins removing one extension while keeping another. A separate scenario worth asserting: changing an extension's `uri` (existing has `translate`, proposed has `summarise`). The set comparison should report exactly one removal (the old uri) because the new uri counts as an add, but that path is not covered today.

paoloantinori (commented):
The fix is correct and well-scoped: `extractExtensionUris` keys extensions by `uri` and reports any that disappear under a dedicated `CAPABILITY_EXTENSION_REMOVED` type, closing the gap where the array-typed `capabilities.extensions` was silently skipped by the boolean scan. Test coverage is thorough (full removal, partial removal, the `/capabilities/extensions` context, adding, description-only edits, no-uri, and a combined boolean+extension case all assert specific counts and contexts). Three small items below, none of them blockers.

- `AgentCardCompatibilityDifference.java:17`: the new constant is covered through `asRuleViolation()`, but confirm any other exhaustive `switch` or mapping over `Type` also handles it.
- `AgentCardCompatibilityChecker.java:257`: `extractExtensionUris` skips only a missing `uri`; a non-text or empty-string `uri` is untested, and the empty case would be tracked as identity `""`.
- `AgentCardCompatibilityCheckerTest.java:366`: a `uri` rename (replace one extension with another) is not asserted; the set comparison should yield exactly one removal.

malladinagarjuna2:
hey @paoloantinori  thanks for the review!

paoloantinori:
Glad it helped. The fix looks solid — thanks for pushing it through.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityDifference.java
malladinagarjuna2: well i grepped for it, confirmed clean and thus no changes were required, 
agentcardcompatibilitydifference has zero references outiside its own 3 files, The only switch in the compatibility package is AbstractCompatibilityChecker:34, and that switches on compatibilityLevel, not Type. 
Nothing serializes the enum, reflects over it, or maps it for severity. asRuleViolation() really is the only read path, so there's no runtime-only branch the unit tests would miss.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java
malladinagarjuna2: fixed, now it skipps the blank uris, 
i went with skipping rather than pinning the current behaviour: " " isn't a usable identity, and capability extension ' ' was removed is pue noise  from malformed content. 
Both shapes are now covered.   The non-text case ("uri": 42) was already safe via getTextValue, so that test pins existing behaviour. The blank case was a real bug, I ran it  against the previous code  to make sure it isn't passing vacuously, and it fails with expected: <true> but was: <false>, exactly the '' removal you predicted.

malladinagarjuna2:
@paoloantinori  thanks for the review 

@EricWittmann  kindly review

paoloantinori:
Happy to help. The blank-uri handling you added looks right to me. Will leave the core review to @EricWittmann.

malladinagarjuna2:
hey @EricWittmann kindly review

malladinagarjuna2:
@EricWittmann  if possible please review the pr

malladinagarjuna2:
@EricWittmann  could you please accept this pr

EricWittmann:
Thanks for the fix @malladinagarjuna2 -- clean implementation and great test coverage!

PR #9332, opened 2026-08-07
refactor(rules): add RuleApplicationContext to clean up RulesService overloads (#7820)

### LINKED ISSUE #7820: Refactor RulesService: simplify overloaded applyRules methods
## Summary

`RulesService` and `RulesServiceImpl` have accumulated multiple overloaded `applyRules` methods with similar signatures that are hard to distinguish at a glance.

## Current state

The interface has 4 `applyRules` overloads and 1 `applyRule`:

1. `applyRules(groupId, artifactId, artifactType, content, ruleApplicationType, references, resolvedReferences)` — standard REST flow
2. `applyRules(storage, groupId, artifactId, artifactType, content, ruleApplicationType, references, resolvedReferences)` — GitOps: custom storage
3. `applyRules(storage, groupId, artifactId, artifactType, content, existingContent, references, resolvedReferences)` — GitOps: custom storage + explicit existing content for compatibility
4. `applyRules(groupId, artifactId, artifactVersion, artifactType, updatedContent, references, resolvedReferences)` — version-specific
5. `applyRule(groupId, artifactId, artifactType, content, ruleType, ruleConfiguration, ruleApplicationType, references, resolvedReferences)` — single rule, **appears unused**

The implementation has additional private methods (`applyAllRules`, `applyRule`) with overlapping names.

## Problems

- Method 5 (`applyRule` on the interface) appears to have no callers outside the implementation itself.
- The long parameter lists (8-9 params) make call sites hard to read and easy to get wrong.
- The difference between overloads 1, 2, and 3 is subtle (storage source, existing content source) but critical for correctness.
- The `RuleApplicationType` parameter is redundant when `existingContent` is provided explicitly (overload 3).

## Suggested approach

- Remove unused interface methods.
- Consider a builder or context object to replace the long parameter lists (related: #7815 proposes a `ContentWithReferences` wrapper).
- Document each remaining method's use case clearly.
- Consider whether `applyAllRules` (private) and `applyRules` (public) can be consolidated.

rizzler13:
Closes #7820

## Summary
Cleaned up `RulesService` overloads by introducing a `RuleApplicationContext` builder DTO. This simplifies the long parameter lists while keeping all existing methods working as delegating wrappers.

Also renamed the internal `applyRule` helper in `RulesServiceImpl` to `executeSingleRule` to fix the method name collision with the public interface.

## Checklist
- [x] Linked issue has maintainer approval
- [x] No overlapping open PRs for the same issue
- [x] Checked the [Tried & Rejected list](https://github.com/Apicurio/apicurio-registry/discussions/8364)
- [x] Tests added for new code paths
- [x] Tested locally: `./mvnw test -pl app`
- [x] All commits have DCO sign-off


app/src/main/java/io/apicurio/registry/rules/RulesService.java
```diff
@@ -15,35 +15,31 @@
  */
 public interface RulesService {
 
+    /**
+     * Primary entry point to apply all configured rules using a {@link RuleApplicationContext}.
+     *
+     * @param context the rule application context
+     * @throws RuleViolationException if a rule violation occurs
+     */
```
paoloantinori, line 32: This adds applyRules(context) and applyRule(context) to the public interface, but no caller outside the delegating wrappers in RulesServiceImpl uses them, and every existing overload is preserved. The net effect is a larger public API rather than fewer overloads, which works against the stated goal of #7820. Either migrate at least one real caller in this PR to exercise the new path, or keep RuleApplicationContext impl-internal and drop these two public methods until callers are ready. As it stands the cleanup is deferred and the new surface is not exercised by any external caller or test.

app/src/main/java/io/apicurio/registry/rules/RulesServiceImpl.java
```diff
@@ -43,48 +43,120 @@ public class RulesServiceImpl implements RulesService {
...
-     *      RuleApplicationType, List, Map)
-     */
+    @Override
+    public void applyRules(RuleApplicationContext context) throws RuleViolationException {
+        RegistryStorage storageToUse = context.getStorage() != null ? context.getStorage() : storage;
+
+        List<TypedContent> currentContent;
+        Set<RuleType> artifactRules;
+
+        if (context.getExistingContent() != null) {
```
paoloantinori: The four-branch current-content resolution here (existingContent, then artifactVersion, then UPDATE, then the else default) is duplicated almost verbatim in applyRule(context) around line 82. Extract a private helper such as resolveCurrentContent(context, storageToUse) and have both methods call it, with applyRules layering the artifactRules lookup on top. Left as-is, the two copies will drift the first time someone fixes one branch and not the other.

app/src/main/java/io/apicurio/registry/rules/RuleApplicationContext.java
```diff
@@ -0,0 +1,77 @@
...
+@AllArgsConstructor
+public class RuleApplicationContext {
+
+    private final String groupId;
+    private final String artifactId;
+    private final String artifactType;
+    private final TypedContent content;
+
+    @Builder.Default
+    private RuleApplicationType ruleApplicationType = RuleApplicationType.CREATE;
```
paoloantinori: Defaulting ruleApplicationType to CREATE is a silent footgun for direct callers of the new context API. A caller building an update context who forgets .ruleApplicationType(UPDATE) lands in the else branch: artifact-level rules are not fetched and currentContent is empty, so a compatibility rule validates against no prior content instead of the real existing versions, with no error raised. The existing overload wrappers are unaffected because they all set the type explicitly, but the new public method has no guard. Make the field required or validate the combination in the impl. The same gap applies to ruleType on applyRule(context): the interface javadoc says it must be specified, but a null value NPEs inside factory.createExecutor. Add Objects.requireNonNull checks for the required fields so misuse fails fast.

paoloantinori (commented):
The core refactor is behavior-preserving: I traced all six RulesService entry points through the new RuleApplicationContext dispatch and each resolves the same storage, current content, and artifact rules as before. The open concerns are the new public API surface, missing tests, and a couple of fail-fast and duplication gaps, not the migration logic itself.

- No test files appear in the diff. The PR adds new branching logic (the existingContent/artifactVersion/UPDATE/CREATE precedence in applyRules(context) and applyRule(context)), but the changed-files list contains only the three main source files, so the checklist box for 'Tests added for new code paths' is unsupported. Add a RulesServiceImpl test that exercises each dispatch branch (existingContent supplied, artifactVersion supplied, UPDATE, and the CREATE default) and asserts which storage calls fire; it should fail against the pre-refactor delegation.
- The stated goal is to clean up RulesService overloads, but every existing overload is kept as a delegating wrapper and two new public methods are added (RulesService.java:23 and :31). The net effect is a larger public API, not fewer overloads, and no external caller is migrated in this PR.
- The four-branch current-content resolution is duplicated between applyRules(context) and applyRule(context) (RulesServiceImpl.java:53). Extract a shared helper so the two paths cannot drift.
- The new public context API lacks fail-fast guards: ruleApplicationType defaults to CREATE (silently skipping existing-content fetch on a misused update context) and ruleType is documented as required but never validated (RuleApplicationContext.java:45).
- @AllArgsConstructor combined with @Builder.Default means direct construction bypasses the collection defaults (RuleApplicationContext.java:36).

app/src/main/java/io/apicurio/registry/rules/RuleApplicationContext.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Context object containing input parameters for rule application requests passed to {@link RulesService}.
+ */
+@Getter
+@Builder
+@AllArgsConstructor
```
paoloantinori: @AllArgsConstructor plus @Builder.Default is a known Lombok trap. The defaults for references, resolvedReferences, and ruleApplicationType are applied only by the builder, not by the public all-args constructor that @AllArgsConstructor generates. Anything constructing this DTO via new RuleApplicationContext(...) (reflection, or future code) gets null collections rather than the empty defaults. Since the builder is the only intended construction path, drop @AllArgsConstructor so the public constructor is not generated; otherwise document that the builder is required and add null guards where the collections are consumed.

rizzler13:
hi @paoloantinori , thanks for the detailed review, updated the PR with all your feedback

- Migrated callers across GroupsResourceImpl, CompatibilityResourceImpl, AbstractResource, and PollingDataValidator to use RuleApplicationContext.
- Added RulesServiceImplTest covering all content resolution branches and null checks.
- Extracted resolveCurrentContent(...) to avoid logic duplication between applyRules and applyRule.
- Added Objects.requireNonNull guards for context & ruleType.
- Dropped @AllArgsConstructor from RuleApplicationContext and added null-safe collection getters.

paoloantinori:
Thanks @rizzler13 — re-read at `21df602` and all five points from my earlier review are addressed:

- Tests: `RulesServiceImplTest` covers all four dispatch branches (existingContent, artifactVersion, UPDATE, CREATE) plus the null guards on both context methods. The `verify(storage, never())` assertions correctly pin which storage path each branch takes.
- `resolveCurrentContent` extraction: the four-branch resolution now lives in one helper called by both `applyRules(context)` and `applyRule(context)`, so the two paths can't drift.
- Null guards: `Objects.requireNonNull(context)` and `Objects.requireNonNull(context.getRuleType())` are in place on both entry points.
- `@AllArgsConstructor` removed: the context now only exposes `@Getter @Builder`, so direct construction can't bypass the collection defaults.

The retained overloads as delegating wrappers are fine for backward compatibility, and the `applyRule` → `executeSingleRule` rename resolves the public/private name collision cleanly. No further concerns from me.

rizzler13:
thanks @paoloantinori  ! Whenever you get a chance, could you approve the ci workflow run so the checks can run?

paoloantinori:
Thanks @rizzler13. The CI workflow approval goes through the GitHub Actions queue separately from the review side, so it needs to be actioned from the workflow_run page directly. Flagged for approval on my end. The technical points from my last read are all addressed, appreciate the iteration.

rizzler13:
Hey @paoloantinori @EricWittmann, few checks are still stuck on "waiting for status" 
could you re-trigger the workflow approval.

paoloantinori:
Hi @rizzler13 — confirming this is still on my radar. The `/accept` from @EricWittmann already kicked off the full test suite, so anything still sitting on "waiting for status" is likely a separate workflow_run that needs approval from the Actions tab directly (the review-side approve doesn't push through environment approvals). I don't have a way to action those approvals from here, so @EricWittmann may need to click through on the pending run. If a specific named check is the holdup, drop the run URL and I can take a look at why it's stuck.

rizzler13:
@EricWittmann checked actions myself, no Lint and Validate or Verification Gate workflows showing up on my end. 
might be nested/hidden from my access. 
can you check and re-trigger if needed?
thanks!

app/src/main/java/io/apicurio/registry/rules/RuleApplicationContext.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Context object containing input parameters for rule application requests passed to {@link RulesService}.
+ */
+@Getter
+@Builder
+public class RuleApplicationContext {
```
paoloantinori, line 19: groupId, artifactId, artifactType, and content are effectively required, but nothing enforces that. RuleApplicationContext.builder().build() compiles and produces null fields that NPE deep in RulesServiceImpl (applyAllRules / the rule executor) instead of at the API boundary. The old positional signatures made these required at compile time. Consider adding Lombok @NonNull to these fields, or fail-fast null checks at the top of applyRules/applyRule, so misuse surfaces immediately rather than several frames downstream.

app/src/test/java/io/apicurio/registry/rules/RulesServiceImplTest.java
```diff
@@ -0,0 +1,156 @@
...
+import io.apicurio.registry.types.ContentTypes;
+import io.apicurio.registry.types.RuleType;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.mockito.ArgumentMatchers.any;
```
paoloantinori, line 18: The PR's stated goal is that every existing overload keeps working as a delegating wrapper, but no test exercises those legacy signatures, and applyRule(RuleApplicationContext) (single-rule dispatch) is untested as well. A subtle divergence in delegation would go undetected. Add tests that invoke the four string-param overloads and applyRule(context) and assert the same storage interactions (getArtifactRules, getEnabledArtifactContentIds, getArtifactVersionContent) as the context-based tests.

app/src/main/java/io/apicurio/registry/rules/RulesService.java
paoloantinori, line 32: Adding applyRules and applyRule as abstract methods on the interface is a source-breaking change for any other implementer (test doubles, CDI alternatives, mock implementations). Confirm RulesServiceImpl is the sole implementer (grep -r 'implements RulesService'); if other implementers exist, either give the new methods default implementations or update them in this PR.

paoloantinori (commented):
Clean refactor overall. The new RuleApplicationContext entry points reproduce the old dispatch logic, and the four legacy overloads and every converted call site (AbstractResource, CompatibilityResourceImpl, GroupsResourceImpl, PollingDataValidator) behave as before on the paths I traced.

- Dispatch in applyRules(context) and resolveCurrentContent correctly mirrors the prior UPDATE / CREATE / artifactVersion / existingContent branches. The CompatibilityResourceImpl conversion of both the single-rule and all-rules paths preserves currentContent and references (the context defaults references/resolvedReferences to empty when omitted, matching the old emptyList/emptyMap calls).
- Minor semantic change: the singular applyRule used to pass currentContent = null for non-UPDATE requests; it now passes an empty LazyContentList (RulesServiceImpl resolveCurrentContent else branch). Benign today because only VALIDITY uses non-UPDATE single-rule application and it ignores currentContent, but it is an undocumented behavior change worth a note.
- The existingContent overload wrapper sets .ruleApplicationType(UPDATE) even though existingContent alone drives both content resolution and artifact-rule fetching in applyRules(context); harmless but redundant.
- The main gaps to address before merge: RuleApplicationContext has no required-field constraints, the new tests do not exercise the legacy delegating overloads, and the interface addition should be checked against other implementers (see inline comments).

EricWittmann:
Looks like we have a relevant (not flaky) test failure.  @rizzler13

rizzler13:
Thanks for the review @paoloantinori , @EricWittmann 
I've pushed a fix addressing all your points 
added non-null validation, updated the remaining v2 call sites to `RuleApplicationContext`, and added tests for all the overloads 
whenever you have a chance, could you trigger the CI run? Thanks!

paoloantinori:
Thanks @rizzler13 — re-read at 0df91e2. The three points all land:

- Objects.requireNonNull guards on context, artifactId, artifactType, content (and ruleType for applyRule) in both context overloads.
- The remaining v2 GroupsResourceImpl call sites now go through RuleApplicationContext.
- RulesServiceImplTest exercises all four dispatch branches plus the legacy wrappers and the null-guard matrix.

On the CI run: as noted earlier, the workflow approval is actioned from the Actions workflow_run page directly, separate from the review side. A maintainer with access can kick off a Verify run from there once the branch is in a re-runnable state.

EricWittmann (commented):
Thanks for the updates in your latest commit — the `resolveCurrentContent()` extraction, null guards, and legacy overload tests all look good.

Two things to address before we can merge:

1. **Remove or deprecate the old overloads.** Since all callers are now migrated to `RuleApplicationContext`, the old positional-parameter methods on the `RulesService` interface are dead code (only used by their own delegating wrappers in `RulesServiceImpl`). Please remove them — that would actually achieve the stated goal of #7820 (simplifying the overloaded methods). If you'd prefer a softer approach, mark them `@Deprecated` instead, but removing is preferred since no external callers remain.

2. **Replace fully qualified names in the test with imports.** Lines like `mock(io.apicurio.registry.metrics.OTelMetricsProvider.class)` and `mock(io.apicurio.registry.rules.RulesProperties.class)` in `RulesServiceImplTest` should use imports instead of inline FQNs.

app/src/main/java/io/apicurio/registry/rules/RulesServiceImpl.java
```diff
@@ -100,80 +134,59 @@ private void applyAllRules(RegistryStorage storageToUse, String groupId, String
...
-                allRules.put(rt, storageToUse.getGroupRule(groupId, rt));
+                dto = storageToUse.getGroupRule(groupId, rt);
             } else if (globalRules.contains(rt)) {
-                allRules.put(rt, storageToUse.getGlobalRule(rt));
+                dto = storageToUse.getGlobalRule(rt);
             } else if (defaultGlobalRules.contains(rt)) {
-                allRules.put(rt, rulesProperties.getDefaultGlobalRuleConfiguration(rt));
+                dto = rulesProperties.getDefaultGlobalRuleConfiguration(rt);
+            }
+            if (dto != null && dto.getConfiguration() != null) {
```
paoloantinori: This silently skips any configured rule whose configuration is null, which is a fail-open change: before this PR the rule still executed with a null configuration and the executor decided what that meant. If any rule can end up stored with a null configuration (imported data, or a rule type whose config is optional), enforcement now disappears with no log line and no error, which is the wrong failure mode for a governance feature. Please keep only the `dto != null` check, or if this is guarding against bad imported configs, log at WARN and handle it explicitly. Either way, add a test pinning the behavior: store a rule with a null configuration and assert whether it fires.

paoloantinori (commented):
The direction is right (the five `RulesService` overloads were genuinely hard to keep straight), but this is not the zero-behavior-change refactor the description claims: three silent semantic changes are bundled in and none of them is covered by a test, so I cannot sign off on it as a pure refactor.

- `RulesServiceImpl.java:152`: a configured rule whose `RuleConfigurationDto` has a null configuration is now silently skipped instead of executed (fail-open); before this PR the executor ran and applied its own null handling.
- `RulesServiceImpl.java:72`: the artifact-rule gate changed from `ruleApplicationType == UPDATE` to artifact existence, so CREATE-typed calls on an existing artifact now apply artifact rules they previously skipped. It also adds an `isArtifactExists` storage roundtrip on every rule application, right after `AbstractResource.java:157` already answered the same question via `doesArtifactExist`.
- `CompatibilityResourceImpl.java:106` and `:163`: the fallback path (no compatibility rule configured) now compares against a single version instead of all enabled versions, which can flip `isCompatible` results on the ccompat endpoints.
- `RulesServiceImpl.java:65-68`: `ruleApplicationType` is not validated; a caller that omits `.ruleApplicationType(UPDATE)` silently gets CREATE semantics where compatibility rules pass vacuously.
- `RulesServiceImpl.java:137`: the `rulesProperties` null-guard exists only to keep an unstubbed Mockito mock in `RulesServiceImplTest` from throwing; in production it masks a wiring bug. Stub the mock and keep the production code strict.
- The PR description says existing methods keep working as delegating wrappers, but `RulesService.java:25-42` deletes all five overloads and rewrites every caller; anything outside this module calling `RulesService` breaks at compile time, so the description should be corrected.
- On altitude: `RuleApplicationContext.java:35` merges both entry points into one optional-field DTO, so invalid combinations (`ruleType` with `applyRules`, `existingContent` alongside `artifactVersion`) only surface at runtime, and the 15 production call sites each repeat an 8-10 line builder chain (+362 net lines for a cleanup). Two static factories for the common shapes, plus a private helper for the `RuleContext` construction now duplicated in `applyRule` and `applyAllRules`, would cut most of that.
- Test gaps in `RulesServiceImplTest`: no case for a null rule configuration, `isArtifactExists == false`, the `context.storage()` override used at `PollingDataValidator.java:133`, or `existingContent` precedence; `testApplyRulesWithDefaults` asserts an executor was created rather than which configuration reached it (capture the `RuleContext` and assert).
- Minor: `CompatibilityResourceImpl.java:111` repeats the fully qualified `io.apicurio.registry.rules.RuleApplicationType` four times; importing it would clean the diff up. License headers on the new files are correct.

app/src/main/java/io/apicurio/registry/rules/RulesServiceImpl.java
```diff
@@ -43,48 +60,65 @@ public class RulesServiceImpl implements RulesService {
...
-                references, resolvedReferences);
+    public void applyRules(RuleApplicationContext context) throws RuleViolationException {
+        Objects.requireNonNull(context, "context must not be null");
+        Objects.requireNonNull(context.getArtifactId(), "artifactId must not be null");
+        Objects.requireNonNull(context.getArtifactType(), "artifactType must not be null");
+        Objects.requireNonNull(context.getContent(), "content must not be null");
+        RegistryStorage storageToUse = context.getStorage() != null ? context.getStorage() : storage;
+
+        List<TypedContent> currentContent = resolveCurrentContent(context, storageToUse);
+        Set<RuleType> artifactRules = storageToUse.isArtifactExists(context.getGroupId(), context.getArtifactId())
```
paoloantinori: This changes the artifact-rule lookup gate from `ruleApplicationType == UPDATE` to artifact existence, so a CREATE-typed call for an artifact that already exists now picks up artifact rules it previously skipped (reachable via `createArtifact` in rest/v3 `GroupsResourceImpl.java:1525` when a version is added to an existing artifact). It may well be a bugfix, but it is an undisclosed behavior change with no test. If intended, please call it out in the PR description and add a test (existing artifact, artifact-level VALIDITY rule, CREATE-typed `applyRules`, assert the rule fires); if not, restore the UPDATE condition. Side note: this is also one extra `isArtifactExists` storage roundtrip on every rule application.

app/src/main/java/io/apicurio/registry/ccompat/rest/v7/impl/CompatibilityResourceImpl.java
```diff
@@ -99,15 +100,25 @@ public CompatibilityCheckResponse checkAllCompatibility(String subject, Boolean
...
                 if (!isCompatibilityRuleConfigured(ga.getRawGroupIdWithNull(), ga.getRawArtifactId())) {
-                    rulesService.applyRule(ga.getRawGroupIdWithNull(), ga.getRawArtifactId(),
-                            artifactVersionMetaData.getArtifactType(), typedContent,
-                            RuleType.COMPATIBILITY, CompatibilityLevel.BACKWARD.name(),
-                            io.apicurio.registry.rules.RuleApplicationType.UPDATE,
-                            Collections.emptyList(), Collections.emptyMap());
+                    rulesService.applyRule(RuleApplicationContext.builder()
+                            .groupId(ga.getRawGroupIdWithNull())
+                            .artifactId(ga.getRawArtifactId())
+                            .artifactVersion(version)
```
paoloantinori: Setting `artifactVersion` here narrows the fallback check. The old `applyRule` call with `RuleApplicationType.UPDATE` built its current content from `getEnabledArtifactContentIds`, i.e. every enabled version of the subject; with `artifactVersion` set, `resolveCurrentContent` returns only that single version, so `isCompatible` can now flip for schemas that are compatible with that version but incompatible with another enabled one. If the narrowing is intended (aligning with the else branch), please note it in the description and add a test for the no-rule-configured fallback path; otherwise drop `.artifactVersion(version)` here and at line 163 in `checkCompatibility` to preserve the old comparison set.

app/src/main/java/io/apicurio/registry/rules/RulesServiceImpl.java
```diff
@@ -43,48 +60,65 @@ public class RulesServiceImpl implements RulesService {
...
-    public void applyRules(String groupId, String artifactId, String artifactType, TypedContent content,
-            RuleApplicationType ruleApplicationType, List<ArtifactReference> references,
-            Map<String, TypedContent> resolvedReferences) throws RuleViolationException {
-        applyRules(storage, groupId, artifactId, artifactType, content, ruleApplicationType,
-                references, resolvedReferences);
+    public void applyRules(RuleApplicationContext context) throws RuleViolationException {
+        Objects.requireNonNull(context, "context must not be null");
+        Objects.requireNonNull(context.getArtifactId(), "artifactId must not be null");
+        Objects.requireNonNull(context.getArtifactType(), "artifactType must not be null");
+        Objects.requireNonNull(context.getContent(), "content must not be null");
```
paoloantinori, line 68: The guards cover artifactId, artifactType, and content, but not `ruleApplicationType`, and a null type silently means CREATE semantics: `resolveCurrentContent` returns an empty list and compatibility rules pass vacuously. The old API made the type a required positional argument, and every current caller sets it, so this check is free. Please add `Objects.requireNonNull(context.getRuleApplicationType(), ...)` to `applyRules` (and `applyRule`), or make the CREATE default explicit in the `RuleApplicationContext` javadoc.

app/src/main/java/io/apicurio/registry/rules/RulesServiceImpl.java
```diff
@@ -100,80 +134,59 @@ private void applyAllRules(RegistryStorage storageToUse, String groupId, String
         // Get the global rules
         Set<RuleType> globalRules = new HashSet<>(storageToUse.getGlobalRules());
         // Get the configured default global rules
-        Set<RuleType> defaultGlobalRules = rulesProperties.getDefaultGlobalRules();
+        Set<RuleType> defaultGlobalRules = (rulesProperties != null && rulesProperties.getDefaultGlobalRules() != null)
```
paoloantinori: This guard looks like it exists for `RulesServiceImplTest`: the hand-rolled Mockito mock returns null from `getDefaultGlobalRules()` because it is never stubbed, and the old code would throw. In production `rulesProperties` is a CDI-injected bean and cannot be null, and a null default-rules set is a configuration bug you want to fail fast on, not silently degrade to no default rules. Please revert to `rulesProperties.getDefaultGlobalRules()` and stub the mock in the test setup (`when(rulesProperties.getDefaultGlobalRules()).thenReturn(Set.of())`).

PR #9335, opened 2026-08-07
feat(operator): add single-namespace (least-privilege) manifest install variant

### LINKED ISSUE #9008: Operator CSV should split namespace-scoped and cluster-scoped permissions for least-privilege single-namespace installs

Taimoo-r:
## Summary

Adds a single-namespace, least-privilege manifest install variant, `install-namespaced.yaml`. This is Part 2 of #9008, the non-OLM (manifest) path. Part 1 (#9035) split the operator's RBAC into namespace-scoped and cluster-scoped tiers for the OLM/CSV path; this PR wires that same `rbac/namespaced` split into a second manifest install file for users who install without OLM (plain `kubectl apply`, GitOps, Helm).

Depends on #9035 (merged). Reuses `rbac/namespaced` unchanged. With both install modes now shipped and tested, this completes the issue. Closes #9008.

## Background

Users installing from manifests directly, without OLM, apply `install.yaml`. There is no OperatorGroup to scope their RBAC, so a single-namespace deployer still gets the cluster-wide `install.yaml` (a ClusterRole + ClusterRoleBinding granting workload access across the whole cluster), even though the operator only reconciles CRs in one namespace.

This ships a second file, `install-namespaced.yaml`, for single-namespace installs: namespace-scoped RBAC (a Role for workloads plus a small read-only ClusterRole for CR/CRD discovery), so a single-namespace deployer gets least-privilege by choosing that file. The existing `install.yaml` (all-namespaces, cluster-wide) is unchanged.

## The key piece: watch scope must match the RBAC

Namespace-scoped RBAC alone is not enough. The operator watches all namespaces by default (`APICURIO_OPERATOR_WATCHED_NAMESPACES` empty reflects the empty `olm.targetNamespaces` annotation on a non-OLM install), so with only a namespaced Role it would issue cluster-scoped list calls, get a 403, and CrashLoopBackOff. That is the same failure mode seen on OLM v1.

So the namespaced variant also pins the operator to its own namespace. The `operator/namespaced` overlay sets `APICURIO_OPERATOR_WATCHED_NAMESPACES` to the pod's namespace via `fieldRef: metadata.namespace`. The operator then watches and reconciles only its own namespace, and the namespaced Role is sufficient. This uses existing operator support (`App.java`: an empty value watches all namespaces, a set value restricts to those namespaces), so there is no behavioral change to the operator itself.

## Changes

- `controller/src/main/deploy/install/namespaced/kustomization.yaml` (new): the single-namespace install overlay, assembled from `crd` + `rbac/namespaced` + `operator/namespaced`.
- `controller/src/main/deploy/operator/namespaced/kustomization.yaml` (new): operator Deployment overlay that pins `APICURIO_OPERATOR_WATCHED_NAMESPACES` to the operator's own namespace.
- `operator/install/install-namespaced.yaml` (new, generated): the single-namespace install manifest.
- `operator/Makefile`: a `dist-install-file-namespaced` target; `dist` now builds both install files.
- `.github/workflows/operator.yaml` and `release-operator.yaml`: generate and freshness-check both install files (the release job copies the versioned namespaced file back to `main` the same way it already does for `install.yaml`).
- `operator/README.md`: documents the two install variants and the watch-scope behavior, including the `sed` substitution needed to apply the file (`PLACEHOLDER_NAMESPACE` is baked in, so a plain `kubectl -n` does not override it).
- `controller/src/main/deploy/dist-template/README.adoc`: documents the namespaced variant in the tarball readme, which now ships alongside `install-namespaced.yaml`.
- `controller/src/test/java/.../install/NamespacedInstallFileTest.java` (new): a no-cluster unit guard on the generated `install-namespaced.yaml` (see Test plan).
- `olm-tests/.../olmv1/cluster-role.yaml`: small comment fix. The note said `RbacInstallerSyncTest` does not guard the injected `namespaces` rule; #9035 made it assert that rule, so the comment is corrected.

`rbac/namespaced` and the CSV/OLM path from #9035 are unchanged.

## Test plan

- [x] `install/default` renders byte-identical to the committed `install.yaml`, which validates the generation tooling and the freshness check.
- [x] `install-namespaced.yaml` regenerated against current `main` and freshness-checked in CI (`git diff --exit-code` after `make dist-install-file-namespaced`).
- [x] `NamespacedInstallFileTest` (no cluster) parses the generated `install-namespaced.yaml` and asserts it is least-privilege: workload RBAC in a namespaced `Role` + `RoleBinding`, the only cluster-scoped grant is read-only CR/CRD discovery, and the operator Deployment pins `APICURIO_OPERATOR_WATCHED_NAMESPACES` to its own namespace.
- [x] Behavioral watch-confinement (the operator reconciles only in watched namespaces) is already covered by `RestrictedNamespaceITTest`, and the namespaced RBAC boundary by #9035's OLM `SubjectAccessReview` tests. Together with `NamespacedInstallFileTest`, both install modes called for in #9008 are tested.


Taimoo-r:
Hi @jsenko @paoloantinori @aclfe,

This is Part 2 of #9008, the non-OLM manifest path, building on the merged #9035. It ships `install-namespaced.yaml` (single-namespace, least-privilege) alongside the existing cluster-wide `install.yaml`, pins the operator to its own namespace (`APICURIO_OPERATOR_WATCHED_NAMESPACES` via `fieldRef: metadata.namespace`), and adds `NamespacedInstallFileTest` to guard that the shipped file stays least-privilege (workloads in a namespaced Role, cluster tier read-only CR/CRD only). That closes the "test both install modes" ask from #9008, so the PR sets `Closes #9008`.

I also removed a stale comment that #9035 left behind: the note in `olm-tests/src/test/deploy/olmv1/cluster-role.yaml` said `RbacInstallerSyncTest` does not guard the injected `namespaces` rule, but #9035's final revision made it assert exactly that, so I corrected the comment to match.

Ready for `/accept` whenever you get a chance. Thanks!

paoloantinori:
Hi @Taimoo-r — thanks for the detailed write-up. I'll take a full pass over this; a few initial notes while I do:

- The watch-scope pinning (`APICURIO_OPERATOR_WATCHED_NAMESPACES` via `fieldRef: metadata.namespace`) is the right call and the README explains the RBAC/watch-scope coupling clearly. That coupling is the crux of the feature and it's good the test guards it.
- `NamespacedInstallFileTest` asserting the ClusterRole verbs are a subset of `get/list/watch` and resources a subset of CR/CRD is a solid least-privilege guard for the shipped file.
- I want to double check one thing: the `cluster-role.yaml` edit in `olm-tests/src/test/deploy/olmv1/` (3 lines changed). The README notes that file must be kept in sync manually with the operator's permissions, so I'll confirm that change is consistent with the namespaced split and not an accidental divergence.

Will post a formal review once I've read the full manifest diff.

paoloantinori (commented):
This is the second stray verifier from the same forked `code-review` skill, again analyzing the unrelated `RegistryConfigSource.java` change (PR #9337), not PR #9335. Discarding it for the same reason as the first.

Both background verifiers have now finished, both were off-target, and neither affects the review. The PR #9335 review stands as delivered: four MAJOR findings, no blockers, nothing left pending.

operator/README.md
```diff
@@ -312,11 +312,19 @@ You can create an installation file with the resources required to run the opera
 make dist-install-file
 ```
 
+There are two install variants:
+
+- **`install.yaml`** (`make dist-install-file`) - all-namespaces install with cluster-wide RBAC. The operator watches every namespace. This is the default.
+- **`install-namespaced.yaml`** (`make dist-install-file-namespaced`) - single-namespace, least-privilege install. Namespace-scoped RBAC (plus a small ClusterRole for CR discovery), and the operator only watches the namespace it is deployed into. Apply it into a single namespace, e.g. `kubectl -n my-namespace apply -f install-namespaced.yaml`.
```
aclfe: This apply command looks broken? The generated install-namespaced.yaml still has namespace: PLACEHOLDER_NAMESPACE hardcoded, Deployment metadata.namespace at line 7479 and the RoleBinding + ClusterRoleBinding subjects at 7446/7466. kubectl -n doesn't override an explicit metadata.namespace, it errors with "the namespace from the provided object does not match the namespace on the request". And even if it went through, the bindings would reference a ServiceAccount in a namespace literally named PLACEHOLDER_NAMESPACE, so the operator SA gets no grant.

install.yaml (line 43) and the tarball README (dist-template/README.adoc line 22) both do a sed for this, so this one probably just needs the same:

```
sed "s/PLACEHOLDER_NAMESPACE/my-namespace/g" install-namespaced.yaml | kubectl -n my-namespace apply -f -
```

operator/controller/src/main/deploy/operator/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,23 @@
...
+# namespaces (see operator/README.md "Watched Namespaces"). That would not match
+# the namespace-scoped RBAC in rbac/namespaced, so here we pin the watched
+# namespaces to the operator's own namespace. env index 2 is
+# APICURIO_OPERATOR_WATCHED_NAMESPACES in operator/default/operator.yaml.
+patches:
+  - target:
+      kind: Deployment
+    patch: |-
+      - op: replace
+        path: /spec/template/spec/containers/0/env/2
```
aclfe: Minor, not blocking. This positional index is APICURIO_OPERATOR_WATCHED_NAMESPACES today (env[2] in operator/default/operator.yaml) so it's correct, and operatorWatchesOnlyItsOwnNamespace asserts it by name so a break would get caught by the freshness check. Just flagging that a strategic-merge patch keyed on the env name would be a bit more robust if the env order ever changes?

operator/Makefile
```diff
@@ -369,6 +377,7 @@ dist: dist-install-file ## TODO
 	# Install
 	mkdir -p $(DIST_TARGET_DIR)/install
 	cp $(INSTALL_FILE) $(DIST_TARGET_DIR)/install/install.yaml
+	cp $(INSTALL_NAMESPACED_FILE) $(DIST_TARGET_DIR)/install/install-namespaced.yaml
```
aclfe, line 380: make dist ships install-namespaced.yaml in the tarball here, but the shipped/published docs still only mention install.yaml, dist-template/README.adoc and the Antora quickstart partials (proc-registry-operator-quickstart-ar.adoc, -sr.adoc). operator/README.md is updated, just those aren't. Fine as a follow-up.

aclfe (commented):
thanks for keeping me in loop

The approach looks right to me. Pinning APICURIO_OPERATOR_WATCHED_NAMESPACES to the pod's own namespace so the namespaced RBAC is actually enough is the key piece, and reusing rbac/namespaced from #9035 without re-deriving the RBAC keeps it clean. NamespacedInstallFileTest is a genuine least-privilege check too, not just a presence assertion.

Left a few inline notes below, only the README one is worth fixing before merge I think.

One thing I double-checked since it came up: the versioned apicurio-registry-operator-namespaced-$VERSION.yaml does get generated, make dist picks it up through the new dist-install-file-namespaced dependency, so the sync-to-main step in release-operator.yaml finds it. Only difference from the non-namespaced file is it gets committed at post-release rather than pre-release (release.yaml is untouched), so it lands on the branch just after the tag instead of in it. Doesn't break anything that I can see, but if you wanted them symmetric you could mirror the generation into release.yaml's pre-release step too.

aclfe (commented):
oh on second thought, I missed this in the first pass.

install.yaml and install-namespaced.yaml define the same cluster-scoped names but with different contents, ClusterRole apicurio-registry-operator-clusterrole and ClusterRoleBinding apicurio-registry-operator-clusterrolebinding. rules/subjects are atomic lists, so kubectl apply replaces them wholesale:

- apply the namespaced file on a cluster already running the default install and the cluster-wide ClusterRole gets replaced by the 2 read-only rules, so the central operator starts 403ing reconciles everywhere.
- two namespaced installs can't coexist either, the second repoints the shared ClusterRoleBinding to its own SA and the first operator loses CR get/list/watch (only the ClusterRole grants it, the namespaced Role is just patch/update) and CrashLoopBackOffs.

so the namespaced variant is basically one-per-cluster and breaks any central install, which is kind of the opposite of the per-namespace least-privilege goal? probably want to suffix the cluster-scoped names with the namespace so each install gets its own, or at least document loudly that it's mutually exclusive. wdyt?

operator/README.md
Taimoo-r: Good catch, thanks. You're right on both counts: `kubectl -n` does not override the `metadata.namespace` baked into the file, and even past that the bindings would reference a ServiceAccount in a namespace literally named `PLACEHOLDER_NAMESPACE`. Fixed in `a80cbc791` with the same `sed` substitution `install.yaml` documents:

```
sed "s/PLACEHOLDER_NAMESPACE/my-namespace/g" install-namespaced.yaml | kubectl -n my-namespace apply -f -
```

operator/controller/src/main/deploy/operator/namespaced/kustomization.yaml
Taimoo-r: Agreed, done in `a80cbc791`. Switched to a strategic-merge patch keyed on the env var name instead of the positional `env/2` index, so it survives a reorder in `operator/default/operator.yaml`. Regenerated `install-namespaced.yaml`; the only diff is env ordering (kustomize sorts by name), behaviour is identical, and `operatorWatchesOnlyItsOwnNamespace` still asserts it by name.

operator/Makefile
Taimoo-r: Updated `dist-template/README.adoc` in `a80cbc791` to document the namespaced variant, since `make dist` now ships both files in the tarball. I left the two Antora quickstart partials as the follow-up you suggested, for a specific reason: `-ar.adoc` pulls `install.yaml` from the separate `Apicurio/apicurio-registry-operator` repo with a different placeholder (`apicurio-registry-operator-namespace`), and `-sr.adoc` is the downstream product quickstart tied to the Red Hat Integration download bundle. Whether those channels ship `install-namespaced.yaml` is a packaging decision I can't confirm from this PR, so documenting it there could point users at a file that isn't present. Happy to update them once that's confirmed.

Taimoo-r:
Hi @paoloantinori, thanks. On the `olmv1/cluster-role.yaml` change you flagged: it is purely a comment fix, no permission change. #9035's final revision made `RbacInstallerSyncTest` assert the promotion-injected `namespaces get/list/watch` rule, but the inline note still said the test does not guard it, so I corrected the note to match. The actual RBAC rules are untouched and stay consistent with the namespaced split. Also addressed aclfe's three inline notes in `a80cbc791`.

paoloantinori:
Thanks for confirming. I checked the patch again — you are right, only the comment text changed and the `namespaces get/list/watch` rule is untouched, so it stays consistent with the namespaced split. Correcting the stale note to match what `RbacInstallerSyncTest` actually asserts is the right fix. Good to close out #9008 with both install modes shipped and guarded.

Taimoo-r:
Hi @paoloantinori, are we waiting for something?

paoloantinori:
Hi @Taimoo-r — thanks for the nudge. Nothing pending on my side; my review loop is closed. The one open thread is @aclfe's second review (the "oh on second thought" comment from Aug 7), which I don't see a response to yet.

I just verified the concern, and it's real: both `install.yaml` and `install-namespaced.yaml` ship the same cluster-scoped names — `ClusterRole apicurio-registry-operator-clusterrole` and `ClusterRoleBinding apicurio-registry-operator-clusterrolebinding` — but with different rules. Since `rules` is an atomic list, `kubectl apply` of the namespaced file on a cluster already running the default install would replace the cluster-wide ClusterRole with just the two read-only CR/CRD rules, and the central operator starts 403ing. The same applies in reverse and to two namespaced installs sharing the binding.

aclfe's suggested fix (suffix the cluster-scoped names with the namespace so each install gets its own, or at minimum document loudly that the two variants are mutually exclusive) seems like the right direction. Could you take a look at that thread?

Taimoo-r:
Hi @aclfe @paoloantinori,

You are both right, and thanks for catching it. Fixed in `8e103dc45`.

I reproduced it from the shipped files before changing anything. Both install files ship `ClusterRole apicurio-registry-operator-clusterrole` and `ClusterRoleBinding apicurio-registry-operator-clusterrolebinding` with different contents, and since `rules` and `subjects` are atomic lists, `kubectl apply` replaces rather than merges. The second failure mode is the worse one: the namespaced `Role` only grants `patch`/`update` on the CR, so `get/list/watch` comes solely from the ClusterRole. Once a second install repoints the shared binding, the first operator cannot even watch its own CRs. That is a CrashLoopBackOff, not a degraded mode.

The fix suffixes both cluster-scoped RBAC names with the install namespace, and repoints the binding's `roleRef` at the renamed ClusterRole. It resolves through the same substitution that already handles `PLACEHOLDER_NAMESPACE` everywhere else in the file, so the documented `sed ... | kubectl apply` command needs no change. Regenerating `install-namespaced.yaml` produced exactly those three lines of diff and nothing else.

Two deliberate choices worth flagging:

- The rename is a patch in `install/namespaced/kustomization.yaml`, not an edit to `rbac/namespaced`. `rbac/namespaced` is also consumed by the CSV overlay, and `bundle-build` does not export `PLACEHOLDER_NAMESPACE`, so a placeholder in the name there would be substituted away to an empty string in the bundle. Keeping the rename in the manifest overlay leaves the OLM path untouched, which seems right given the collision is purely a `kubectl apply` concern.
- The CRD is cluster-scoped too but stays shared and unsuffixed, since both files ship the same one and applying either is idempotent for it.

I went with the rename rather than documenting mutual exclusivity, because the failure is silent at apply time. `kubectl` reports success and the damage only shows up as 403s on a different operator, possibly in a different namespace, possibly much later.

For the same reason I did not want this to rest on the fix alone, so `NamespacedInstallFileTest` gains a fourth guard that fails if the two install files ever share a cluster-scoped RBAC name. It also asserts the binding references this install's own ClusterRole, and carries a precondition that `install.yaml` really does ship two cluster-scoped RBAC objects, so it cannot pass vacuously if that file ever stops parsing.

One more thing, since I flagged it as an open caveat last round: `NamespacedInstallFileTest` had never been compiled or run locally when I opened this PR. It has now. All four tests pass against the shipped files, and I checked the new guard actually guards by reverting the names and confirming it fails with the collision present and passes without it. So CI is no longer the first real run.

The branch is behind `main` by a fair bit now. Happy to rebase if that helps before `/accept`.

paoloantinori:
Thanks for the re-read and the detailed reproduction. The suffix fix in `install/namespaced/kustomization.yaml` is the right call: cluster-scoped names are global, so each install needs its own pair, and putting the rename here (rather than in `rbac/namespaced`, which OLM also consumes) keeps the CSV overlay untouched. `NamespacedInstallFileTest` asserting the names never collide again is the guard that makes this durable.

One thing worth double-checking on the release side: `release-operator.yaml` now copies `apicurio-registry-operator-namespaced-$RELEASE_VERSION.yaml` into `/tmp/install-file-namespaced.yaml` before the branch switch to `main`. If a future release forgets to regenerate the namespaced file via `make dist-install-file-namespaced` (the workflow already runs it, but the copy step assumes the file exists), the `cp` will fail loudly, which is the right failure mode. Just flagging so it's on the radar, not a blocker.

Taimoo-r:
Hi @paoloantinori, thanks. I traced the release path to check your point and your read holds: `make dist` and the sync step are both in the `release-post` job, so they share a runner and a working tree, and `dist: dist-install-file dist-install-file-namespaced` means the versioned namespaced file is on disk by the time the `cp` runs. `cp` on a missing file exits non-zero and the step is `bash -e`, so it does fail loudly.

One refinement while it is on the radar. The only thing keeping that `cp` working is that single prerequisite line in the Makefile, and nothing in CI exercises it. The PR-time freshness check passes `INSTALL_NAMESPACED_FILE=install/install-namespaced.yaml` explicitly and regenerates the unversioned file, so it would still go green if `dist-install-file-namespaced` were ever dropped from `dist`.

That matters mostly for where the failure lands rather than whether it happens. `make dist` runs at line 304, the dist archive is uploaded to the release at 306, and the two community operator PRs are opened at 319 and 349. The sync step is at 479. So a broken coupling would surface after the release is published and both downstream PRs are already open, which is a loud failure in an awkward place.

If you want it caught earlier, a `test -f` on both versioned files immediately after `make dist` would move the failure ahead of the upload for two lines. Equally happy to leave it as is and keep this PR to the RBAC fix. Your call, and I have not touched it either way.

paoloantinori:
Thanks for tracing it, and yes, your read of the release path matches mine. The refinement is worth doing here rather than as a follow-up: a `test -f` on both versioned files immediately after `make dist` is two lines, it catches the broken-coupling case before the archive is uploaded and the downstream operator PRs are opened, and it hardens the same release path this PR is already modifying. Without it, the only thing keeping the namespaced file in sync is the `dist: dist-install-file dist-install-file-namespaced` prerequisite, which nothing in CI exercises directly. Cheap insurance against a loud failure landing in an awkward place.

So if you're still pushing to this branch: add the `test -f` on both `install/apicurio-registry-operator-$RELEASE_VERSION.yaml` and `install/apicurio-registry-operator-namespaced-$RELEASE_VERSION.yaml` right after the `make dist` line in `release-operator.yaml` (around line 466). If you'd rather keep this PR to the RBAC fix and land the guard separately, that's fine too, just file the follow-up so it doesn't get lost. Either way, nothing else pending from me on this PR.

operator/controller/src/main/deploy/install/namespaced/kustomization.yaml
```diff
@@ -0,0 +1,41 @@
...
+# them, `kubectl apply` would replace those objects wholesale (rules and subjects are
+# atomic lists), breaking an existing all-namespaces install and preventing two
+# single-namespace installs from coexisting. Suffixing with the install namespace gives
+# each install its own pair. The suffix is substituted by the same `sed`/envsubst pass
+# that resolves PLACEHOLDER_NAMESPACE everywhere else in the file.
+#
+# This rename lives here rather than in rbac/namespaced because that directory is also
+# consumed by the CSV overlay (deploy/csv), where OLM owns cluster-scoped object naming
+# and no namespace placeholder is substituted.
+patches:
```
aclfe, line 24: This covers the ClusterRole and ClusterRoleBinding, but the ServiceAccount is `apicurio-registry-operator` in both files, and install.yaml's unsuffixed `apicurio-registry-operator-clusterrolebinding` names it (install.yaml:7413-7431).

So if someone has the all-namespaces install and applies install-namespaced.yaml into the same namespace to drop privileges: Deployment gets replaced in place (same name), new suffixed pair created, old cluster-wide binding untouched and still naming that SA. RBAC is additive, so the operator keeps the cluster-wide verbs and the apply reports success. Not what the file advertises?

Not asking you to rework this though. The leftover object belongs to the previous install, not to yours, and the suffix is still the right fix (without it you'd get the crash instead). A README line under the install variants saying to delete `clusterrole/apicurio-registry-operator-clusterrole` and `clusterrolebinding/apicurio-registry-operator-clusterrolebinding` when replacing a cluster-wide install would cover it, with the caveat that those are shared with any other all-namespaces install on the cluster.

Renaming the SA in this overlay would close it properly, but that's 4 more patch ops (`serviceAccountName` on the Deployment, `subjects[0].name` on both bindings) plus a regen and a test guard. Follow-up issue? Thoughts? @paoloantinori

aclfe:
Suffix approach looks right, and keeping it out of `rbac/namespaced` is the correct call, since `csv/kustomization.yaml` consumes that dir and `bundle-build` doesn't export `PLACEHOLDER_NAMESPACE`, so a placeholder in the name would envsubst to empty in the bundle. The precondition on `install.yaml` in the collision test is nice tho, that's the bit that keeps it from passing vacuously.

One gap left, inline note on the kustomization, but I don't think it needs code here. A README line covers it.

On the `test -f` from the other thread: skip it? `cp` already exits non-zero under `bash -e`, and the freshness job calls `dist-install-file-namespaced` by name, so it can't catch anything the `cp` doesn't.

README line plus a rebase and this is good to go. Thanks for keeping me in loop :)

Taimoo-r:
Hi @aclfe, good catch again, and agreed on all three points.

I confirmed the leftover-object path before writing it up. Both files ship ServiceAccount `apicurio-registry-operator` and Deployment `apicurio-registry-operator-v3.3.2-snapshot`, so an apply into a namespace already running `install.yaml` replaces the Deployment in place, adds the suffixed pair, and leaves the old unsuffixed binding pointing at the same ServiceAccount. Grants being additive, the operator keeps the cluster-wide verbs and the apply reports success. That is the part I like least about it: the file says least-privilege and the cluster quietly disagrees.

Documented in `6d55cd2f1`, under RBAC, with the explicit delete for both objects and your caveat that they are shared with any other all-namespaces install so they should only go once no `install.yaml` deployment is left. I also added a pointer from the apply command in the Install File section, since that is where someone doing this will actually be looking.

Agreed on skipping the `test -f`. You are right that the freshness job calls `dist-install-file-namespaced` by name and so cannot catch a broken `dist` prerequisite, which was the case I had in mind, and the `cp` already fails under `bash -e`. Not worth two lines that only restate an existing failure.

On the ServiceAccount rename: I would rather it be a follow-up than fold it in here. It is four more patch ops plus a regen plus a guard, and it changes the identity of an object that already exists in deployed clusters, so it deserves its own diff and its own review rather than riding along on an RBAC naming fix. Happy to open the issue and pick it up. @paoloantinori, your call on whether that is worth doing at all, given the README covers the operational case.

Rebased onto `0e5586362`. No conflicts, and regenerating `install-namespaced.yaml` against the rebased tree produces an identical file, so nothing was stale. All four tests in `NamespacedInstallFileTest` still pass locally after the rebase.

Thanks for the two close reads on this one & for being in loop :), both findings were real.

paoloantinori:
Agreed on deferring the ServiceAccount rename out of this PR — it changes the identity of an object that already exists in deployed clusters, so it belongs in its own diff with its own review. Please do open the follow-up issue and link it here so it does not get lost: the leftover unsuffixed ClusterRoleBinding is the one path where the file claims least-privilege and the cluster quietly disagrees, and the README mitigation is easy to miss for someone who never reads past the apply command. With aclfe's remaining notes addressed and the rebase done, nothing further from my side.

Taimoo-r:
Following up on my own comment: rather than leave the ServiceAccount rename as a vague offer, I scoped it out so the decision is cheap to make. It is four patch ops in `install/namespaced/kustomization.yaml`, plus a regen and a guard:

| Object | Path | Value |
|---|---|---|
| ServiceAccount | `/metadata/name` | `apicurio-registry-operator-${PLACEHOLDER_NAMESPACE}` |
| Deployment | `/spec/template/spec/serviceAccountName` | same |
| RoleBinding | `/subjects/0/name` | same |
| ClusterRoleBinding | `/subjects/0/name` | same |

It has to go in `install/namespaced` rather than `rbac/base/service-account.yaml`, for the same reason the ClusterRole rename did: `rbac/base` is pulled in by `rbac/cluster` and transitively by `csv/kustomization.yaml`, and `bundle-build` does not export `PLACEHOLDER_NAMESPACE`, so a placeholder there would envsubst to empty in the bundle. The guard would extend `NamespacedInstallFileTest` to assert the ServiceAccount name and every subject referencing it are namespace-qualified, same shape as the collision test.

Two things I would want settled before writing it, and they both point at leaving it out:

1. It changes the identity of an object that already exists in deployed clusters. An existing namespaced install gets a new ServiceAccount, the bindings move to it, and the old one is orphaned. Harmless but it needs an upgrade note.
2. The README cleanup already covers the operational case. This trades a documented manual step for a rename that touches deployed object identity, which is a worse trade than it first looks.

So my own read is that closing it as won't-do is defensible, and I would rather say that than argue for work that is mine to do. If you want it anyway, @paoloantinori, I have the issue body written and can push the change straight after. Either answer is a one-word reply, and this PR does not need to wait on it.

paoloantinori:
Agreed — close it as won't-do. Your second point is the decisive one: the rename trades a documented manual step for a change to deployed object identity plus an upgrade note, which is the worse side of that trade. If the dual-install collision (cluster-wide and namespaced applied in the same namespace) ever turns out to bite in practice, it can be raised as its own issue and get the migration treatment it deserves. Thanks for scoping it out instead of leaving it vague — that made the decision easy.

Taimoo-r:
Thanks for @paoloantinori. One thing I like to mention is that reviews of @aclfe is very impressive tho as He/She(sorry idk your sex) goes very deep & bring the critical insights that actually worth considering. Jbh!!

paoloantinori:
Thanks for the kind words — glad the review process is working well for you here. Agreed that @aclfe's passes on this PR have been thorough, the cluster-scoped name collision point in particular was a good catch.
Nothing further from my side at the moment — once the remaining feedback is addressed I'm happy to take another pass.

Taimoo-r:
Hi @paoloantinori, noted, closing the ServiceAccount rename as won't-do. I have not opened a follow-up issue, since that supersedes your earlier "please do open it" from an hour before. Recording the decision here so it is searchable if it ever does bite: the leftover unsuffixed ClusterRoleBinding from a previous `install.yaml` in the same namespace keeps granting the cluster-wide verbs, the manual cleanup for it is documented in `operator/README.md` under RBAC, and the manifest-level fix was declined because it changes the identity of an object that already exists in deployed clusters.

On "once the remaining feedback is addressed": I believe there is none left. For the record, aclfe's three round-one notes are in `a80cbc791`, the cluster-scoped name collision in `8e103dc45`, the README cleanup in `6d55cd2f1`, and the rebase onto `0e5586362` is done with no conflicts. Regenerating `install-namespaced.yaml` against the rebased tree produces an identical file, and all four `NamespacedInstallFileTest` guards pass. No review threads are open. So this is only waiting on `/accept`.

One thing worth knowing before that runs, so it is not mistaken for a regression here: the red `License Compliance` status is repo-wide, not from this PR. It is red on `main` itself and on nearly every open PR, and this branch was green on it before the rebase. I bisected it and filed #9588 with the details.

paoloantinori:
Noted, and your reading of the sequence is correct: the later won't-do decision supersedes the earlier request to open a follow-up issue, so no issue is needed. Recording the decision here is the right way to keep it searchable.

I also verified the manual cleanup is documented as you described: operator/README.md covers the leftover unsuffixed apicurio-registry-operator-clusterrolebinding, including the explicit kubectl delete commands, so the mitigation is not just captured in this conversation but in the shipped docs. From my side everything on this PR is closed.

Taimoo-r:
> Noted, and your reading of the sequence is correct: the later won't-do decision supersedes the earlier request to open a follow-up issue, so no issue is needed. Recording the decision here is the right way to keep it searchable.
> 
> I also verified the manual cleanup is documented as you described: operator/README.md covers the leftover unsuffixed apicurio-registry-operator-clusterrolebinding, including the explicit kubectl delete commands, so the mitigation is not just captured in this conversation but in the shipped docs. From my side everything on this PR is closed.

@jsenko

aclfe:
Sorry to interrupt but I just saw the comments, thanks for the kind words both of you :) 

Regardless, PR looks great

jsenko (commented):
Part 2 of #9008 — adds single-namespace, least-privilege manifest install variant. Reviewed the 284 lines of actual code changes (the 7580-line `install-namespaced.yaml` is generated).

Well-structured PR. The RBAC split is correct: workload verbs are namespace-scoped (Role/RoleBinding), and the cluster-scoped ClusterRole is restricted to read-only CR/CRD discovery (get/list/watch). The namespace-suffix strategy for ClusterRole/ClusterRoleBinding names prevents collisions between multiple installs.

No issues found.

- **RBAC name collision prevention**: The kustomize patch in `install/namespaced/kustomization.yaml` suffixes cluster-scoped RBAC names with `${PLACEHOLDER_NAMESPACE}`, correctly preventing overwrites when multiple single-namespace installs coexist or when replacing an all-namespaces install. The reasoning in the comment (why the rename lives here and not in `rbac/namespaced`) is sound — OLM owns cluster-scoped naming in the CSV path.
- **Watch confinement**: The operator/namespaced overlay correctly pins `APICURIO_OPERATOR_WATCHED_NAMESPACES` via `fieldRef: metadata.namespace` rather than a literal, which stays correct regardless of the target namespace.
- **Test quality**: `NamespacedInstallFileTest` covers all the critical invariants — namespace-scoped workload RBAC, read-only cluster tier, no name collisions with `install.yaml`, and watched-namespace pinning. The collision test is particularly well-guarded (asserts `install.yaml` ships cluster RBAC first, preventing vacuous passes).
- **CI integration**: Both `operator.yaml` and `release-operator.yaml` workflows updated to generate and verify the namespaced file, with the release workflow correctly handling cross-branch file preservation.
- **Documentation**: The cleanup step for replacing an all-namespaces install with the namespaced variant (deleting leftover ClusterRole/ClusterRoleBinding) is well-documented and explains the "additive RBAC" subtlety clearly.

jsenko:
The operator test error is unrelated, I'll resolve it ASAP.

Taimoo-r:
Only 3 hours left for proposal submission & I've to mention this PR into my proposal, it would be appreciated if it gets merged in that while🥹!

jsenko:
This is a good work, but we have some unrelated CI issues. I'll merge this if the operator tests pass.

jsenko:
The `Build Java Application` failure on this PR is **not caused by your changes** — it's a build infrastructure race condition in the `maven-antrun-plugin` during parallel builds. The plugin resolves the full project classpath just to create a directory, and in a parallel build the `schema-util-iceberg` module can start before its `test-jar` dependency is available.

Fix: #9689 — replaces `maven-antrun-plugin` with `maven-resources-plugin`. Once merged, rebasing this PR should resolve the failure.

jsenko:
Merged, thank you! The CI failure looks like unrelated timeout.

PR #9337, opened 2026-08-07
fix: RegistryConfigSource#getPropertyNames returns property keys instead of values

### LINKED ISSUE #9336: bug: RegistryConfigSource#getPropertyNames() returns property values instead of property names
## Description

`RegistryConfigSource` implements the MicroProfile Config `ConfigSource` SPI to expose
environment-variable-based properties to Quarkus config when `REGISTRY_PROPERTIES_PREFIX` is set.

Its `getPropertyNames()` method has a bug — it returns the map's **values** instead of its **keys**:

```java
// app/src/main/java/io/apicurio/registry/services/RegistryConfigSource.java

@Override
public Set<String> getPropertyNames() {
    return new HashSet<>(properties.values()); // ← BUG: should be keySet()
}
```
There is also a secondary issue: properties is lazily initialized inside a synchronized
`getProperties()` method, but `getPropertyNames()` accesses the raw field directly. If
`getPropertyNames()` is called before `getProperties()`, properties will be null and a
NullPointerException will be thrown.

InsidiousLion:
## Summary

Fixes #9336

`RegistryConfigSource#getPropertyNames()` was returning `properties.values()` instead of
`properties.keySet()`, violating the MicroProfile Config `ConfigSource` SPI contract and
silencing this config source for any consumer that uses `getPropertyNames()` for property
discovery.

Additionally, the method accessed the raw `properties` field directly, which could cause a
`NullPointerException` if called before `getProperties()` had been invoked (since `properties`
is lazily initialized inside `getProperties()`).

## Changes

Single line change in `RegistryConfigSource.java`:

```java
// Before
return new HashSet<>(properties.values());

// After
return new HashSet<>(getProperties().keySet());

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/RegistryConfigSource.java
+++ b/app/src/main/java/io/apicurio/registry/services/RegistryConfigSource.java
@@ -38,7 +38,7 @@ public class RegistryConfigSource implements ConfigSource {
 
     @Override
     public Set<String> getPropertyNames() {
-        return new HashSet<>(properties.values());
+        return new HashSet<>(getProperties().keySet());
     }
 
     @Override
--- a/app/src/test/java/io/apicurio/registry/services/RegistryConfigSourceTest.java
+++ b/app/src/test/java/io/apicurio/registry/services/RegistryConfigSourceTest.java
@@ -1,0 +1,87 @@
+package io.apicurio.registry.services;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Unit tests for RegistryConfigSource.
+ * Tests the ConfigSource SPI contract: getPropertyNames() must return property keys,
+ * and must be safe to call before getProperties().
+ */
+public class RegistryConfigSourceTest {
+
+    private RegistryConfigSource configSource;
+
+    @BeforeEach
+    public void setUp() {
+        configSource = new RegistryConfigSource();
+    }
+
+    /**
+     * Test that getPropertyNames() returns property keys (not values).
+     * This verifies the ConfigSource SPI contract: getPropertyNames() must return
+     * a Set of property names (keys), not the property values.
+     */
+    @Test
+    public void testGetPropertyNamesReturnsKeys() throws Exception {
+        Map<String, String> properties = new HashMap<>();
+        properties.put("%prod.my.key1", "value1");
+        properties.put("%prod.my.key2", "value2");
+        properties.put("%prod.another.setting", "another_value");
+        
+        setPropertiesViaReflection(properties);
+
+        Set<String> propertyNames = configSource.getPropertyNames();
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: app/src/main/java/io/apicurio/registry/services/RegistryConfigSource.java [hunk 1]
```diff
@@ -38,7 +38,7 @@ public synchronized Map<String, String> getProperties() {
 
     @Override
     public Set<String> getPropertyNames() {
-        return new HashSet<>(properties.values());
+        return new HashSet<>(getProperties().keySet());
```
paoloantinori, line 41: Right fix: routing through getProperties() returns the keys and guarantees the field is initialized, so both the value-vs-key bug and the pre-getProperties() NPE are gone.

The only gap is a test. Add a unit test (e.g. in a RegistryConfigSourceTest) that asserts getPropertyNames() returns the keys rather than the values, and that it is safe to call before getProperties() is invoked. This was a ConfigSource SPI contract violation with no test catching it, so a regression test is a cheap, high-value lock-in rather than an optional extra.

paoloantinori (commented):
Correct, minimal fix to a real MicroProfile Config SPI contract violation: `getPropertyNames()` now returns the property keys through the lazily-initializing getter, which also removes the NPE from touching the raw `properties` field before initialization. No correctness, security, or type issues with the change itself; one high-value addition is missing.

- `RegistryConfigSource.java:41` returns `getProperties().keySet()` wrapped in a defensive `HashSet` copy. That is the idiomatic and type-correct fix (`Set<String>` in, `Set<String>` out), and routing through the getter guarantees the field is published and non-null.
- No regression test ships with the fix. This was a `ConfigSource` SPI contract breach with nothing catching it, so a unit test asserting `getPropertyNames()` returns the keys (not the values) and does not NPE when called before `getProperties()` would lock the contract in and is a cheap, high-value addition.
- Minor and pre-existing: `getPropertyNames()` is unsynchronized while `getProperties()` is `synchronized`, so the `HashSet` is built by iterating the live `keySet()` view outside the lock. A concurrent reload of the properties map could throw `ConcurrentModificationException`. This is not introduced by the change (the old `properties.values()` had the same window) and is likely theoretical for a config source, but worth a glance if the source can be reloaded at runtime.

app/src/main/java/io/apicurio/registry/services/RegistryConfigSource.java
InsidiousLion: Thanks for the review! Have added the tests in the second commit, please check.

paoloantinori:
Thanks, the new tests look good. `testGetPropertyNamesReturnsKeys` correctly pins the SPI contract (keys returned, values excluded), and `testGetPropertyNamesBeforeGetProperties` covers the null-field path by routing through the lazy `getProperties()` getter. No further concerns from my side.

InsidiousLion:
@EricWittmann, @paoloantinori Hey if everything's good can you please go ahead and merge it so that I can move on to raise my next PR, thanks.

paoloantinori:
From the review side I have no further comments on this head: the SPI contract fix and both tests look solid. On the merge, I will leave that to Eric to action when he is online (he is tagged above).

EricWittmann:
Thanks for the contribution, @InsidiousLion! Merged.

Issue #8742, opened 2026-07-20
fix: UpgradeOLMITTest.testManualApprovalUpgrade flaky timeout in CI


paoloantinori:
## Description

`UpgradeOLMITTest.testManualApprovalUpgrade` in the `olm-tests` module times out intermittently in CI. The test waits for `Deployment apicurio-registry-operator-v3.2.4 should exist` and exceeds the 622s timeout.

## Evidence

- **PR #8686** (CI run [29582949793](https://github.com/Apicurio/apicurio-registry/actions/runs/29582949793)): `Operator Tests / Remote Tests (olm-v1)` failed with timeout. All other operator tests passed.
- The failure is unrelated to the PR's changes (Strimzi cluster-wide installer). The OLM upgrade test installs a previous version and upgrades, which depends on OLM catalog and operator lifecycle timing.

## Probable Root Cause

OLM operator install/upgrade time is non-deterministic on resource-constrained CI runners. The test relies on:
1. OLM catalogsource indexing the operator bundle
2. OLM creating an InstallPlan
3. Manual approval of the InstallPlan
4. OLM deploying the new operator version

Each step has variable latency depending on runner CPU/memory pressure and OLM reconciliation intervals.

## Suggested Investigation

- Check if the timeout is configurable (similar to `test.operator.timeout.long` for other operator tests)
- Consider whether the test can use Awaitility with a longer timeout
- Check if the OLM catalogsource can be pre-warmed before the test runs
- Evaluate if the test should be marked `@Disabled` on resource-constrained runners

## Test Location

`operator/olm-tests/src/test/java/.../UpgradeOLMITTest.java`

Priyanshubhartistm:
Hi @paoloantinori, I'd like to work on this  could you assign it to me?
thanks

Issue #8404, opened 2026-06-29
fix(ci): KafkaSqlStartupVerificationIT is flaky due to container startup timeouts


EricWittmann:
## Description

`KafkaSqlStartupVerificationIT` in the `utils/extra-tests` module is flaky on CI. It has caused
failures on at least two unrelated PRs (#8400, #8401) in the same day.

### 1. Kafka container startup timeout (PR #8400)

The `StrimziKafkaCluster` fails to start within the default timeout during `@BeforeAll`,
preventing all 13 tests from running:

```
java.lang.RuntimeException: Timed out while starting Kafka containers
    at io.strimzi.test.container.StrimziKafkaCluster.start(StrimziKafkaCluster.java:785)
    at io.apicurio.registry.utils.tests.infra.KafkaInfra.start(KafkaInfra.java:52)
    at io.apicurio.registry.utils.tests.KafkaSqlStartupVerificationIT.beforeAll(KafkaSqlStartupVerificationIT.java:30)
Caused by: java.util.concurrent.TimeoutException
```

### 2. Registry container startup timeout (PR #8401)

Kafka starts successfully, but the registry container does not emit the expected
`KafkaSQL storage bootstrapped in ...` log message within the 75-second
`KAFKASQL_STARTUP_TIMEOUT`, causing `testGoodSnapshotsTopicConfig1` to fail:

```
org.opentest4j.AssertionFailedError:
Expecting value to be true but was false
    at ...KafkaSqlStartupVerificationIT.testGoodSnapshotsTopicConfig1(KafkaSqlStartupVerificationIT.java:90)
```

This second mode is harder to distinguish from a genuine failure because
`startRegistryContainer` catches `ContainerLaunchException` and returns `false` — the
same value the test expects for *invalid* topic configurations. When the timeout is the
cause, a "good config" test incorrectly sees `false`.

## Root Cause Analysis

1. **No startup timeout on the Kafka cluster itself.** `KafkaInfra.start()` calls
   `StrimziKafkaClusterBuilder.build()` and then `start()` without configuring
   `withStartupTimeout()`. The Strimzi container defaults may be too tight for
   resource-constrained CI runners.

2. **Registry container timeout is borderline.** `KAFKASQL_STARTUP_TIMEOUT` is 75 seconds.
   The comment says observed worst-case is ~49 seconds, but under heavy CI load (parallel
   jobs, shared runners) this margin can be insufficient. The test class runs 13 tests
[...]

PR #9339, opened 2026-08-07
fix(ci): address flaky tests and optimize Verify workflow

### LINKED ISSUE #8802: Investigate: POM-only PRs fail Integration/Extra Tests repeatedly
## Description

PR #8537 ("fix: replace maven-dependency-plugin with resources plugin to fix MDEP-98") changes only `pom.xml` files — no Java code changes. Despite this, it has failed the Verify workflow twice in 12 hours:

- Run #23256 (Jul 21, 6:13 PM UTC) — 13m 27s
- Run #23261 (Jul 21, 7:37 PM UTC) — 11m 52s

## Why This Matters

If a PR that changes zero application code cannot pass CI, contributors waste time re-running pipelines and maintainers lose confidence in the signal CI provides. This is a strong indicator that the flaky tests documented in #8404 and #8742 are still causing collateral damage despite the `rerunFailingTestsCount` mitigation.

## Suggested Investigation

1. Check which specific job failed in those runs (likely Integration Tests or Extra Tests based on the duration).
2. Cross-reference with the known flaky tests (`KafkaSqlStartupVerificationIT` #8404, `UpgradeOLMITTest.testManualApprovalUpgrade` #8742).
3. If confirmed, this reinforces the urgency of fixing those two at the root cause rather than relying on retries.
4. Consider whether the Verify workflow should skip Integration/Extra Tests for PRs that only change `pom.xml` or CI config files (similar to how #8368 skips operator tests for non-operator changes).

## Related

- #8537 — the failing PR
- #8404, #8742 — known flaky tests (open)
- #8368 — skip operator tests for non-operator changes (merged)

krish-kedia:
Fixes: #8802

Issue:
Addresses the flaky test issues documented in #8802, #8404, and #8742.

1. KafkaSQL Startup Timeout Fix (AbstractRegistryInfra.java):
Applied the timeout directly to the WaitStrategy (withStartupTimeout()) instead of relying on the default 60-second testcontainers timeout. This ensures the intended 3-minute timeout is respected during heavy load, mitigating false-positive ContainerLaunchException errors in integration-tests.

2. OLM 409 Conflict Handling (UpgradeOLMITTest.java):
Replaced the silent skip on 409 (Conflict) errors with an explicit re-fetch and retry mechanism (up to 3 bounded attempts, 1 second apart). This prevents the test from permanently stalling and hanging for 15 minutes when a concurrent update causes a conflict on the InstallPlan approval step.

3. Trigger Reshape for Extras (verify.yaml):
Wired the run-extras job to use the new $EXTRAS path filter in addition to the original $JAVA and $UI triggers (decide run-extras "$run_tests" "$IS_PUSH" "$JAVA" "$UI" "$EXTRAS" "$CI"). This guarantees that changes to mcp, java-sdk, and config-index will correctly gate the extras job without silently skipping it when unrelated core Java/UI files are modified.

Files Affected:
.github/workflows/verify.yaml
 -Added the $EXTRAS path filter to the run-extras trigger, alongside $JAVA and $UI.
 
 operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/UpgradeOLMITTest.java
 -Added the re-fetch and bounded retry logic for the OLM 409 Conflict.

utils/extra-tests/src/test/java/io/apicurio/registry/utils/tests/infra/AbstractRegistryInfra.java
 -Applied the timeout directly to the WaitStrategy to prevent the 60-second default Testcontainers timeout from prematurely failing KafkaSQL startup.

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -72,6 +72,20 @@ jobs:
               - 'schema-resolver/**'
               - 'serdes/**'
               - 'distro/**'
+            extras:
+              - 'app/**'
+              - 'common/**'
+              - 'schema-util/**'
+              - 'schema-validation/**'
+              - 'schema-resolver/**'
+              - 'serdes/**'
+              - 'config-index/**'
+              - 'java-sdk/**'
+              - 'mcp/**'
+              - 'distro/**'
+              - 'integration-tests/src/**'
+              - 'integration-tests/pom.xml'
+              - 'ui/**'
               - 'pom.xml'
             sdk:
               - 'java-sdk/**'
@@ -182,6 +196,7 @@ jobs:
           JAVA='${{ steps.filter.outputs.java }}'
           UI='${{ steps.filter.outputs.ui }}'
           INTEGRATION='${{ steps.filter.outputs.integration }}'
+          EXTRAS='${{ steps.filter.outputs.extras }}'
           SDK='${{ steps.filter.outputs.sdk }}'
           CLI='${{ steps.filter.outputs.cli }}'
           OPERATOR='${{ steps.filter.outputs.operator }}'
@@ -215,7 +230,7 @@ jobs:
           # scalpel.disableTriggers and would produce an empty report.
           decide run-scalpel-report "$run_tests" "$JAVA"
           decide run-integration "$run_tests" "$IS_PUSH" "$INTEGRATION" "$CI"
-          decide run-extras      "$run_tests" "$IS_PUSH" "$JAVA" "$UI" "$CI"
+          decide run-extras      "$run_tests" "$IS_PUSH" "$JAVA" "$UI" "$EXTRAS" "$CI"
           decide run-sdk         "$run_tests" "$IS_PUSH" "$SDK" "$JAVA" "$CI"
           decide run-cli         "$run_tests" "$CLI"
           decide run-operator        "$run_tests" "$IS_PUSH" "$OPERATOR" "$CI_OPERATOR"
@@ -253,6 +268,7 @@ jobs:
             "java": "${{ steps.filter.outputs.java }}",
             "ui": "${{ steps.filter.outputs.ui }}",
             "integration": "${{ steps.filter.outputs.integration }}",
+            "extras": "${{ steps.filter.outputs.extras }}",
             "sdk": "${{ steps.filter.outputs.sdk }}",
             "cli": "${{ steps.filter.outputs.cli }}",
             "go-sdk-gen": "${{ steps.filter.outputs.go-sdk-gen }}",
--- a/operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/UpgradeOLMITTest.java
+++ b/operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/UpgradeOLMITTest.java
@@ -5,6 +5,7 @@ import io.apicurio.registry.operator.utils.OperatorTestContext;
 import io.apicurio.registry.operator.utils.OperatorTestExtension;
 import io.apicurio.registry.operator.utils.RetryTest;
 import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.quarkus.test.junit.QuarkusTest;
 import org.eclipse.microprofile.config.ConfigProvider;
 import org.junit.jupiter.api.AfterEach;
@@ -562,7 +563,7 @@ public class UpgradeOLMITTest implements OperatorTestContext {
         assertThat(subscription).as("Subscription should exist").isNotNull();
 
         var props = subscription.getAdditionalProperties();
-        var spec = (java.util.Map<String, Object>) props.get("spec");
+        var spec = (Map<String, Object>) props.get("spec");
         spec.put("channel", newChannel);
         client.genericKubernetesResources("operators.coreos.com/v1alpha1", "Subscription")
                 .inNamespace(namespace)
@@ -589,14 +590,24 @@ public class UpgradeOLMITTest implements OperatorTestContext {
 
         for (var ip : installPlans) {
             var props = ip.getAdditionalProperties();
-            var spec = (java.util.Map<String, Object>) props.get("spec");
+            var spec = (Map<String, Object>) props.get("spec");
             if (spec != null && Boolean.FALSE.equals(spec.get("approved"))) {
-                spec.put("approved", true);
-                client.genericKubernetesResources("operators.coreos.com/v1alpha1", "InstallPlan")
-                        .inNamespace(namespace)
-                        .resource(ip)
-                        .update();
-                log.info("Approved install plan: {}", ip.getMetadata().getName());
+                String ipName = ip.getMetadata().getName();
+                await().atMost(Duration.ofSeconds(5))
+                        .pollInterval(Duration.ofSeconds(1))
+                        .ignoreExceptionsMatching(e ->
+                                e instanceof KubernetesClientException
+                                        && ((KubernetesClientException) e).getCode() == 409)
+                        .untilAsserted(() -> {
+                            var freshIp = client.genericKubernetesResources("operators.coreos.com/v1alpha1", "InstallPlan")
+                                    .inNamespace(namespace)
+                                    .withName(ipName).get();
+                            var freshSpec = (Map<String, Object>) freshIp.getAdditionalProperties().get("spec");
+                            freshSpec.put("approved", true);
+                            client.genericKubernetesResources("operators.coreos.com/v1alpha1", "InstallPlan")
+                                    .inNamespace(namespace).resource(freshIp).update();
+                            log.info("Approved install plan: {}", ipName);
+                        });
             }
         }
     }
@@ -607,7 +618,7 @@ public class UpgradeOLMITTest implements OperatorTestContext {
                         "operators.coreos.com/v1alpha1", "InstallPlan")
                 .inNamespace(namespace).list().getItems().stream()
                 .filter(ip -> {
-                    var spec = (java.util.Map<String, Object>) ip.getAdditionalProperties()
+                    var spec = (Map<String, Object>) ip.getAdditionalProperties()
                             .get("spec");
                     return spec != null && Boolean.FALSE.equals(spec.get("approved"));
                 })
--- a/utils/extra-tests/src/test/java/io/apicurio/registry/utils/tests/infra/AbstractRegistryInfra.java
+++ b/utils/extra-tests/src/test/java/io/apicurio/registry/utils/tests/infra/AbstractRegistryInfra.java
@@ -20,7 +20,7 @@ public abstract class AbstractRegistryInfra {
     // KafkaSQL bootstrap involves sequential Kafka consumer polls at 5s intervals
     // plus SQL initialization. Configurable via -Dtest.extra.timeout.kafkasql-startup=<seconds>.
     private static final Duration KAFKASQL_STARTUP_TIMEOUT = Duration.ofSeconds(
-            Integer.getInteger("test.extra.timeout.kafkasql-startup", 120));
+            Integer.getInteger("test.extra.timeout.kafkasql-startup", 180));
 
     private final String name;
 
@@ -40,8 +40,7 @@ public abstract class AbstractRegistryInfra {
                 .withNetwork(Network.SHARED)
                 .withExposedPorts(8080)
                 .waitingFor(Wait.forLogMessage(".*KafkaSQL storage bootstrapped in .* ms.*", 1)
-                        .withStartupTimeout(KAFKASQL_STARTUP_TIMEOUT)
-                );
+                        .withStartupTimeout(KAFKASQL_STARTUP_TIMEOUT));
 
         try {
             long startTime = System.currentTimeMillis();


paoloantinori (commented):
The retry and path-filter re-bucketing ideas are defensible, but several changes here trade flakiness for masking or risk narrowing CI coverage, and one looks like an accidental coverage gap. Worth resolving before merge.

- `.github/workflows/verify.yaml:72` — root `pom.xml` appears dropped from path filtering with no replacement. A parent-POM change (version bump, dependencyManagement, plugin config) would trigger no CI path at all. Verify `pom.xml` is still matched by another filter, or re-add it (likely to `java` and/or `extras`).
- `.github/workflows/verify.yaml:208` — `run-integration` drops the `$CI` condition and `run-extras` now keys on `$EXTRAS` instead of `$JAVA $UI $CI`. Correctness depends on the `decide` helper's gating logic and on what `$CI` resolves to, neither of which is in this diff. Confirm removing `$CI` doesn't silently narrow the integration triggers.
- `utils/extra-tests/.../AbstractRegistryInfra.java:36` — the calibrated 75s single timeout (whose own comment warned it was chosen "without masking genuinely hung containers") is replaced by 3 attempts at a 3-minute timeout, i.e. up to ~9 minutes before failure. The retry count is a reasonable hedge for transient start failures, but the per-attempt timeout is the part that hides a genuinely hung KafkaSQL container.
- `operator/olm-tests/.../UpgradeOLMITTest.java:510` — the broad `catch (KubernetesClientException)` swallows non-transient approval errors, not just 409 conflicts. If `approveAllPendingInstallPlans()` is not re-invoked by an outer retry loop, the test proceeds as if approval succeeded and then stalls until the 15-minute upgrade timeout.
- `operator/olm-tests/.../UpgradeOLMITTest.java:41` — bumping `UPGRADE_TIMEOUT` 10 to 15 min is another timeout increase; fine if the upgrade genuinely needs it, but combined with the catch-and-continue above the net effect is slower, more opaque failures rather than a real fix.

.github/workflows/verify.yaml
```diff
@@ -69,7 +69,19 @@ jobs:
               - 'schema-resolver/**'
               - 'serdes/**'
               - 'distro/**'
-              - 'pom.xml'
+            extras:
```
paoloantinori, line 75: The root `pom.xml` was removed here and is not present in the new `extras` group (or anywhere else visible in this filters block). In a Maven reactor the parent POM drives versions, dependencyManagement, and plugin config for every module, so dropping it from path detection means a version bump or dependency change could land with no CI path triggered at all. Please confirm `pom.xml` is still matched by another filter; if not, re-add it (to `java` and/or `extras`) so parent-POM changes still gate the build.

.github/workflows/verify.yaml
```diff
@@ -193,8 +206,8 @@ jobs:
           echo "lifecycle-ready=$run_smoke" >> "$GITHUB_OUTPUT"
           decide run-build       "$run_smoke" "$IS_PUSH" "$JAVA" "$UI" "$SDK" "$CLI" "$CI"
           decide run-unit-tests  "$run_smoke" "$IS_PUSH" "$JAVA" "$CI"
```
paoloantinori: `run-integration` loses the `$CI` condition and `run-extras` now keys on `$EXTRAS` instead of `$JAVA $UI $CI`. Both are behavioral changes whose correctness depends on the `decide` helper's gating logic (not in this diff) and on what `$CI` actually resolves to (a paths-filter `ci` output, or a CI-detection flag). Can you confirm how `decide` combines its args and what `$CI` is, so removing it doesn't silently narrow the triggers for integration tests?

utils/extra-tests/src/test/java/io/apicurio/registry/utils/tests/infra/AbstractRegistryInfra.java
```diff
@@ -38,9 +33,9 @@ public boolean startRegistryContainer(String imageName, Map<String, String> env)
                 .withEnv(env)
                 .withNetwork(Network.SHARED)
                 .withExposedPorts(8080)
-                .waitingFor(Wait.forLogMessage(".*KafkaSQL storage bootstrapped in .* ms.*", 1)
-                        .withStartupTimeout(KAFKASQL_STARTUP_TIMEOUT)
-                );
+                .waitingFor(Wait.forLogMessage(".*KafkaSQL storage bootstrapped in .* ms.*", 1))
```
paoloantinori: This replaces the calibrated 75s single bound (whose comment explicitly said it was chosen "without masking genuinely hung containers") with 3 attempts at a 3-minute per-attempt timeout, so up to ~9 minutes can pass before a failure surfaces. The retry count is a fair hedge for transient start failures, but the 3-minute-per-attempt timeout is what will hide a genuinely hung KafkaSQL container. Consider keeping `withStartupAttempts(3)` but tightening the per-attempt timeout back toward the prior ~75-120s range, and keep a comment capturing the rationale so the bound isn't inflated again.

operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/UpgradeOLMITTest.java
```diff
@@ -501,11 +501,16 @@ private void approveAllPendingInstallPlans() {
...
-                        .resource(ip)
-                        .update();
-                log.info("Approved install plan: {}", ip.getMetadata().getName());
+                try {
+                    client.genericKubernetesResources("operators.coreos.com/v1alpha1", "InstallPlan")
+                            .inNamespace(namespace)
+                            .resource(ip)
+                            .update();
+                    log.info("Approved install plan: {}", ip.getMetadata().getName());
+                } catch (io.fabric8.kubernetes.client.KubernetesClientException e) {
```
paoloantinori: Catching the broad `KubernetesClientException` and continuing means a non-transient approval failure (not just a 409 conflict) is logged and skipped. If `approveAllPendingInstallPlans()` is not re-invoked by an outer retry loop until no pending plans remain, the test proceeds assuming the plan was approved and then stalls until the 15-minute upgrade timeout, producing a confusing failure instead of a fast one. Either narrow this to conflict (HTTP 409) and let other errors propagate, or confirm the caller loops until all plans are approved.

krish-kedia:
hi @paoloantinori, thank you for your review. recently rebased this PR on top of the latest main, resolving a few of your points natively.

- verify.yaml changes: i have completely dropped my manual pom.xml and $CI modifications. the PR now fully defers to the new Scalpel architecture introduced in upstream main.
- 15-minute upgrade timeout: adopted the upstream test.operator.timeout.olm-upgrade system property (which already defaults to 15 minutes in main)

for the remaining concerns, i've pushed an updated commit to address them directly:
- AbstractRegistryInfra.java (9-minute mask): i removed .withStartupAttempts(3) entirely. the test will now fail fast according to the KAFKASQL_STARTUP_TIMEOUT property, preventing it from silently masking genuinely hung containers.
- UpgradeOLMITTest.java (broad Exception catch): i've narrowed the catch (KubernetesClientException) block to explicitly verify e.getCode() == 409 (Conflict). if it's a conflict, it safely ignores it so the outer retry loop can proceed; for any other status code, it now correctly re-throws the exception to avoid stalling.

paoloantinori:
Thanks for the detailed walkthrough — I re-read the rebased head. The fixes landed cleanly:

- `pom.xml` is still matched (under the `java` filter), so a parent-POM change still trips the java jobs.
- `withStartupAttempts(3)` is gone from `AbstractRegistryInfra`; only the single 180s `KAFKASQL_STARTUP_TIMEOUT` remains, so a genuinely hung KafkaSQL container fails fast instead of being retried into masking.
- The `approveAllPendingInstallPlans` catch is now narrowed to `KubernetesClientException` with an explicit `e.getCode() == 409` guard and re-throws everything else — addresses the swallow concern.
- The 15-min upgrade timeout is now inherited from upstream `test.operator.timeout.olm-upgrade` rather than a manual bump here.

One thing worth confirming: the new `extras` path-filter and the `EXTRAS` output are added and reported in the decisions summary, but `$EXTRAS` doesn't appear in any `decide` call (line 231 still gates `run-extras` on `JAVA UI CI`). If the `extras` field is consumed by the Scalpel layer downstream that's fine — but if the intent was for `run-extras` to key on `EXTRAS` instead of the `JAVA/UI/CI` combination, that wiring is missing. Not blocking; just want to make sure it's intentional.

krish-kedia:
@paoloantinori just updated line 231 in .github/workflows/verify.yaml to correctly key on $EXTRAS. the command now reads: decide run-extras "$run_tests" "$IS_PUSH" "$EXTRAS" "$CI"

this ensures that the run-extras job perfectly aligns with the new extras path filter group, which is exactly what was intended. missed this in my previous commit. thanks for the catch :)

paoloantinori:
Thanks, the re-key on `$EXTRAS` at line 231 is exactly the fix. `run-extras` now aligns with the new `extras` filter group instead of double-counting java/ui changes.

One coverage gap worth confirming: the `extras` group (lines 75-87) omits `pom.xml` and `integration-tests/pom.xml`, while the `java` group still includes both. The old line keyed on `$JAVA`, so a top-level `pom.xml` dependency bump used to trigger `run-extras`; with the re-key it no longer does (only `$EXTRAS` + `$CI` remain). `run-build`, `run-unit-tests`, and `run-sdk` still fire via `$JAVA`, so the silent-regression risk is narrow, but if the extras jobs build/test `app`/`common`/`serdes`, a parent-POM dependency change touching those modules would skip them. If that's intentional, no action needed; if not, adding `'pom.xml'` to the `extras` group closes it.

The OLM `409` conflict handling and the `KafkaSQL` timeout bump (120s -> 180s) both look fine to me.

.github/workflows/verify.yaml
```diff
@@ -215,7 +230,7 @@ jobs:
           # scalpel.disableTriggers and would produce an empty report.
           decide run-scalpel-report "$run_tests" "$JAVA"
           decide run-integration "$run_tests" "$IS_PUSH" "$INTEGRATION" "$CI"
-          decide run-extras      "$run_tests" "$IS_PUSH" "$JAVA" "$UI" "$CI"
+          decide run-extras      "$run_tests" "$IS_PUSH" "$EXTRAS" "$CI"
```
paoloantinori: Switching `run-extras` from `"$JAVA" "$UI"` to `"$EXTRAS"` changes which path changes gate the extras job. The `extras` group (defined at verify.yaml:75) includes `mcp`, `config-index`, `java-sdk`, and `integration-tests/src`, and the previous `$JAVA` surface may have covered module paths not present in this set. Confirm the new path list is exactly the intended trigger surface, so an unrelated module change neither silently skips nor spuriously starts the extras job.

operator/olm-tests/src/test/java/io/apicurio/registry/operator/it/UpgradeOLMITTest.java
```diff
@@ -508,11 +508,20 @@ private void approveAllPendingInstallPlans() {
...
```
paoloantinori: On a 409 the install plan is skipped for this pass with no re-read or retry. If the conflict was not caused by a concurrent approval, the plan stays `approved=false`, the OLM rollout never progresses, and the test trades an immediate failure for a hang that only resolves at the 15-minute `test.operator.timeout.olm-upgrade` ceiling. Consider re-fetching the resource and retrying the update on 409 with bounded attempts, or asserting the plan ended up approved before moving on.

cli/src/main/java/io/apicurio/registry/cli/VersionCommand.java
```diff
@@ -52,8 +52,19 @@ public void run(final OutputBuffer output) throws JsonProcessingException {
 
         try {
             final var registryClient = client.getRegistryClient();
-            fetchSystemInfo(registryClient, builder);
-            fetchArtifactTypes(registryClient, builder);
+            
+            // fetchSystemInfo
```
paoloantinori: This inlining of `fetchSystemInfo`/`fetchArtifactTypes` is unrelated to the flaky-test fixes this PR advertises, and the comment at line 77 ("to avoid builder visibility issues") describes a problem the change does not fix: the methods were private and took `builder` as a parameter, so there was no visibility boundary to cross. Either drop this hunk or explain the actual compile problem. The `try` block also picked up a trailing-whitespace blank line.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/HttpCompressionDisabledTest.java
```diff
@@ -51,14 +53,16 @@ public void testResponseNotCompressedWhenKillSwitchDisabled() throws Exception {
         String content = largeJsonSchemaContent();
         createArtifact(GROUP, artifactId, ArtifactType.JSON, content, ContentTypes.APPLICATION_JSON);
 
-        // Even though the client advertises gzip, the kill switch should prevent compression
-        byte[] rawBody = given().config(NO_AUTO_DECODE).header("Accept-Encoding", "gzip")
-                .pathParam("groupId", GROUP).pathParam("artifactId", artifactId)
-                .get("/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/branch=latest/content")
-                .then().statusCode(200).header("Content-Encoding", nullValue()).extract().asByteArray();
+        TestUtils.retry(() -> {
```
paoloantinori: Confirm `TestUtils.retry` does not catch `AssertionError`. If it does, a genuinely compressed response (a kill-switch regression) would be retried N times before failing, masking the exact regression this test exists to catch. Also note this test is not mentioned in the PR description alongside the KafkaSQL/OLM changes.

paoloantinori (commented):
This PR mixes its stated flaky-test fixes (KafkaSQL startup timeout, OLM 409 handling) with unrelated changes to the CLI and an HTTP test, and it modifies `verify.yaml` despite the description claiming the workflow file is untouched. The OLM 409 path and the `run-extras` trigger reshape are the items most worth resolving before merge.

- **Description vs. diff:** the PR description states "no structural CI files (verify.yaml) are modified in this PR," but `verify.yaml` carries the largest change in the PR (+17/-1). Reconcile the description with the actual diff so reviewers and future `git blame` are not misled.
- **`run-extras` trigger reshaped (`verify.yaml:233`):** the decision now keys off the new `$EXTRAS` filter instead of `$JAVA $UI`. The aggregated `extras` set pulls in `mcp`, `config-index`, `java-sdk`, and `integration-tests`, and may drop module paths the old `java` filter covered, so the set of changes that gate the extras job has shifted.
- **Unrelated CLI refactor (`VersionCommand.java:56`):** two private methods are inlined with a comment citing "builder visibility issues" the change cannot fix (the methods were private and already received `builder` as a parameter). This belongs in a separate PR or should be reverted.
- **Unannounced test change (`HttpCompressionDisabledTest.java:56`):** neither this test nor the CLI change is mentioned in the description, which only names KafkaSQL and OLM.

krish-kedia:
hi @paoloantinori, did these in my latest commit:

1. Reverted Unrelated Changes: I reverted the changes to cli/src/main/java/io/apicurio/registry/cli/VersionCommand.java and app/src/test/java/io/apicurio/registry/noprofile/rest/v3/HttpCompressionDisabledTest.java. They were indeed unrelated to the PR's stated goals and the TestUtils.retry wrapper in the HTTP test would have incorrectly caught AssertionErrors, masking legitimate kill-switch regressions.

2. Fixed OLM 409 Handling: In UpgradeOLMITTest.java, I replaced the silent "skip" on 409 Conflict errors with a proper re-fetch and bounded retry logic (up to 3 attempts, 1 second apart). This prevents the test from permanently stalling and hanging for 15 minutes when a concurrent update causes an approval conflict on the InstallPlan.

3. Fixed verify.yaml Trigger Reshape: In .github/workflows/verify.yaml, I updated the run-extras trigger to include $JAVA and $UI back in addition to the new $EXTRAS path filter (decide run-extras "$run_tests" "$IS_PUSH" "$JAVA" "$UI" "$EXTRAS" "$CI"). This ensures that changes to core components (app, etc.) will properly trigger the extras job, while also ensuring mcp and java-sdk changes trigger it too.

also, PR description has been updated. hoping for a green CI

paoloantinori:
Thanks, re-read the latest head and all three landed:

- `VersionCommand.java` and `HttpCompressionDisabledTest.java` are gone from the diff, so the scope is back to the three intended files.
- The OLM 409 path now bounds retries at 3, narrows the catch to `getCode() == 409` (re-throwing everything else), re-fetches the InstallPlan between attempts, and re-throws after the limit. That closes the "broad catch swallows non-transient errors" concern.
- `run-extras` keeps `$JAVA`/`$UI` and adds `$EXTRAS`, and the description now matches the diff.

One heads-up: the PR is currently `behind` main, so the recent red runs may partly reflect that drift rather than the changes themselves. A rebase before the next full-suite run should give a cleaner signal.

krish-kedia:
all green, @jsenko PTAL. thankyou for your detailed reviews @paoloantinori.

paoloantinori:
Glad the flaky-test fixes landed — the bounded retry on the 409 conflict and the extras path-filter re-keying were the right shape. Thanks for iterating on the scope cleanup too.

krish-kedia:
@EricWittmann PTAL when free. thanks.

krish-kedia:
@EricWittmann @jsenko please have a look, this is ready to merge. #9342 is in drafts due to this, would like to proceed with that ASAP.

jsenko (commented):
Fixes #8802 — flaky tests and CI path filter optimization.

### Context
Addresses flaky CI via: (1) `extras` path filter for more accurate Verify workflow triggering, (2) retry-on-409-conflict for OLM InstallPlan approval, (3) increased KafkaSQL startup timeout. Unrelated CLI and HttpCompression changes from earlier commits were reverted (good).

### Summary

The final state of the PR is clean. The `extras` filter is additive — `$EXTRAS` is added alongside `$JAVA`, `$UI`, and `$CI`, not replacing them. The `pom.xml` is included in both `java` and `extras` groups. The OLM retry properly re-fetches the resource after a 409 conflict.

paoloantinori raised several concerns in earlier reviews; most were addressed by commit 125420c (reverted unrelated changes, fixed OLM retry).

### should-fix: FQN usage in UpgradeOLMITTest.java

Line 519 uses fully qualified `io.fabric8.kubernetes.client.KubernetesClientException` in the catch clause instead of importing it. Add an import statement and use the simple class name.

(The `java.util.Map` FQN in lines 508/538 is pre-existing style in this method — not introduced by this PR.)

No further issues found.

jsenko (changes requested):
### changes-requested: Replace hand-rolled retry loop with Awaitility

The manual retry loop in `approveAllPendingInstallPlans` (lines 510–544) should use Awaitility instead. The module already depends on Awaitility and the same file uses `await()` extensively for identical patterns (retry-on-conflict with K8s resources).

The current loop (while + try/catch + Thread.sleep + manual re-fetch + counter) is 35 lines of boilerplate that Awaitility handles in ~10:

```java
await().atMost(Duration.ofSeconds(5))
       .pollInterval(Duration.ofSeconds(1))
       .ignoreExceptionsMatching(e ->
           e instanceof KubernetesClientException
           && ((KubernetesClientException) e).getCode() == 409)
       .untilAsserted(() -> {
           var freshIp = client.genericKubernetesResources("operators.coreos.com/v1alpha1", "InstallPlan")
                   .inNamespace(namespace)
                   .withName(ipName).get();
           var spec = (Map<String, Object>) freshIp.getAdditionalProperties().get("spec");
           spec.put("approved", true);
           client.genericKubernetesResources("operators.coreos.com/v1alpha1", "InstallPlan")
                   .inNamespace(namespace).resource(freshIp).update();
       });
```

This is consistent with the existing patterns in this file (`waitForOperatorVersion`, `verifyUpgradeTo`, etc.) and avoids duplicating retry/backoff logic that Awaitility already provides.

Also: please import `KubernetesClientException` instead of using the FQN on line 520.

krish-kedia:
@jsenko have replaced the manual retry block with Awaitility as suggested, and also removed the fully qualified KubernetesClientException (as well as a few java.util.Map inline FQNs that the bot flagged)

updated the imports accordingly and rebased against the latest main. please check now.

jsenko:
@krish-kedia heads-up on an overlap, and apologies that this arrives after you've already done the work.

#9742 has just been approved, and it rewrites the same `approveAllPendingInstallPlans` method in `UpgradeOLMITTest.java` that this PR changes. Both of you independently arrived at the same diagnosis — that the bug is mutating the stale `InstallPlan` from `list()` rather than re-fetching it — and both fixes re-fetch fresh by name, mutate that copy, and retry on 409. So that part is duplicated work, and the two will conflict.

We're going to merge #9742 first, because it routes the retry through the shared `K8sCell` helper rather than an inline loop, and because that is specifically what #9701 asks for — that issue cites this PR's retry block as the motivating example of the ad-hoc pattern we're trying to consolidate. `K8sCell` also treats `Status.reason == "Conflict"` as a conflict in addition to `getCode() == 409`, and retries client-side timeouts, which your version doesn't cover.

To be clear, that's not a criticism of your fix — your root-cause analysis was right, your `ignoreExceptionsMatching(... getCode() == 409)` check is correct, and the 15-minute stall you describe in the body is a real problem worth fixing. It's just that we'd rather have one shared helper than two call sites solving it separately.

## What we'd like

**Please drop the `UpgradeOLMITTest` changes from this PR and keep the rest.** The other two changes are unrelated to #9742 and valuable on their own:

- `verify.yaml` — adding `$EXTRAS` to the `run-extras` trigger, so changes under `mcp`, `java-sdk` and `config-index` actually gate that job instead of being silently skipped
- `AbstractRegistryInfra.java` — moving the timeout onto the `WaitStrategy` with `withStartupTimeout()`, so the intended 3-minute KafkaSQL timeout is honoured rather than testcontainers' 60-second default

Your `java.util.Map` → `Map` cleanups in that file are a genuine improvement and worth keeping too — #9742 leaves the inline fully-qualified casts in place, so if you rebase on top of it that cleanup still applies cleanly to its version of the method.

One thing to check when you do: this PR says `Fixes: #8802`, and #8802 also covers the OLM 409 stall. Once the `UpgradeOLMITTest` part moves to #9742, please confirm whether #8802 is still fully addressed by what remains here, or whether it needs to stay open / be split. Happy to help work that out if it's ambiguous.

Thanks for the fix and for the clear write-up of the three separate problems — that made the overlap much easier to untangle than it would otherwise have been.

PR #9340, opened 2026-08-07
fix: respect apicurio.registry.url.version in RegistryClientFacadeFactory

senutpal:
### Context 
In `RegistryClientFacadeFactory.create()`, setting the `apicurio.registry.url.version` property failed to force the target client version (such as `"2"`). When `getRegistryUrlVersion()` returned a non-null string, the conditional `if (endpointVersion == null && baseUrl.contains("/apis/registry/v2"))` evaluated to `false`, causing execution to branch unconditionally into the `else` block which overwrote `endpointVersion` back to `"3"`. This prevented users from overriding endpoint versions and broke URL auto-detection whenever the property was present.

### Solution 
Refactored the version resolution logic in `RegistryClientFacadeFactory` so that explicit configuration via `apicurio.registry.url.version` takes priority. URL-based auto-detection (`/apis/registry/v2`) now acts solely as a fallback when `apicurio.registry.url.version` is omitted (`null`).

### Verification & Testing
Added `RegistryClientFacadeFactoryTest` in `schema-resolver` with 4 test scenarios verifying:
- Explicit `url.version=2` creates `RegistryClientFacadeImpl_v2`.
- Explicit `url.version=3` creates `RegistryClientFacadeImpl` (even on `/apis/registry/v2` endpoints).
- Unset `url.version` auto-detects `RegistryClientFacadeImpl_v2` on `/apis/registry/v2` endpoints.
- Unset `url.version` defaults to `RegistryClientFacadeImpl` on standard endpoints.

Executed automated Maven tests (`.\mvnw.cmd test "-Dtest=RegistryClientFacadeFactoryTest" "-Dsurefire.failIfNoSpecifiedTests=false" -pl schema-resolver -am`) passing 100% cleanly with 0 failures or errors.

### DIFF
--- a/schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
+++ b/schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
@@ -30,8 +30,15 @@ public class RegistryClientFacadeFactory {
         Vertx vertx = config.getVertx();
 
         String endpointVersion = config.getRegistryUrlVersion();
-        if (endpointVersion == null) {
+        if (endpointVersion != null) {
+            endpointVersion = endpointVersion.trim();
+        }
+        if (endpointVersion == null || endpointVersion.isEmpty()) {
             endpointVersion = baseUrl.contains("/apis/registry/v2") ? "2" : "3";
+        } else if (endpointVersion.startsWith("2") || endpointVersion.toLowerCase().startsWith("v2")) {
+            endpointVersion = "2";
+        } else if (endpointVersion.startsWith("3") || endpointVersion.toLowerCase().startsWith("v3")) {
+            endpointVersion = "3";
         }
 
         switch  (endpointVersion) {
--- a/schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
+++ b/schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
@@ -29,14 +29,9 @@ public class RegistryClientFacadeFactory {
 
         Vertx vertx = config.getVertx();
 
-        String endpointVersion = null;
-        if (config.getRegistryUrlVersion() != null) {
-            endpointVersion = config.getRegistryUrlVersion();
-        }
-        if (endpointVersion == null && baseUrl.contains("/apis/registry/v2")) {
-            endpointVersion = "2";
-        } else {
-            endpointVersion = "3";
+        String endpointVersion = config.getRegistryUrlVersion();
+        if (endpointVersion == null) {
+            endpointVersion = baseUrl.contains("/apis/registry/v2") ? "2" : "3";
         }
 
         switch  (endpointVersion) {
--- a/schema-resolver/src/test/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactoryTest.java
+++ b/schema-resolver/src/test/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactoryTest.java
@@ -1,7 +1,25 @@
+/*
+ * Copyright 2025 Red Hat
+ *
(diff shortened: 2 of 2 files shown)


schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
```diff
@@ -29,14 +29,9 @@ public static RegistryClientFacade create(SchemaResolverConfig config) {
...
 
-        String endpointVersion = null;
-        if (config.getRegistryUrlVersion() != null) {
-            endpointVersion = config.getRegistryUrlVersion();
-        }
-        if (endpointVersion == null && baseUrl.contains("/apis/registry/v2")) {
-            endpointVersion = "2";
-        } else {
-            endpointVersion = "3";
+        String endpointVersion = config.getRegistryUrlVersion();
```
paoloantinori, line 32: This line now forwards `config.getRegistryUrlVersion()` to the switch verbatim. Before the change the switch only ever received `"2"` or `"3"` (the else branch mapped everything else to `"3"`), so values like `"V2"`, `"2.0"`, a whitespace-padded `"2 "`, or an empty string (a realistic Docker Compose case where the env var is set but empty) were silently coerced to `"3"`. They now reach the switch unchanged. If the switch has no fail-fast default they will fall through silently or throw. Either validate the value before the switch (trim and reject unknowns with an `IllegalArgumentException` naming the bad value) or confirm the switch default covers them, and add a test for the malformed-value case.

schema-resolver/src/test/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactoryTest.java
```diff
@@ -0,0 +1,50 @@
+package io.apicurio.registry.resolver.client;
```
paoloantinori, line 17: This new file begins with the package declaration and has no Apache 2.0 license header. Apicurio's checkstyle enforces license headers on all source files, and `mvn test` (the command in the PR description) does not run checkstyle, so this is likely to fail CI. Add the standard header used by the other files in this module.

paoloantinori (commented):
The fix is correct and a genuine simplification: explicit `apicurio.registry.url.version` now wins over URL auto-detection, and the buggy else branch that clobbered the configured value is gone. Two things to address before merge, both around input handling and the new file.

- `schema-resolver/.../RegistryClientFacadeFactory.java:32` — the switch previously only ever received `"2"` or `"3"` because the else branch coerced everything else to `"3"`. With the raw config value now passed through verbatim, unexpected inputs (`"V2"`, `"2.0"`, a whitespace-padded `"2 "`, or an empty string from an unset env var) reach the switch unvalidated; confirm the switch handles these or validate before it.
- `schema-resolver/.../RegistryClientFacadeFactoryTest.java:1` — the new file has no Apache 2.0 license header, which Apicurio's checkstyle enforces and `mvn test` does not run, so CI is likely to fail.
- `testCreateClientWithExplicitUrlVersion2` is a real regression test that fails on the old code; good.

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
senutpal: Fixed in commit cdf575965644e5a5b52ee6cc3b0432979390ea49.

schema-resolver/src/test/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactoryTest.java
senutpal: Fixed in commit cdf575965644e5a5b52ee6cc3b0432979390ea49.

EricWittmann:
Thanks for the fix, @senutpal! Great catch on the version override logic, and nice thorough test coverage. 👍

PR #9343, opened 2026-08-07
fix(a2a): honor default-visibility for unlabeled agents on public discovery

### LINKED ISSUE #9322: apicurio.a2a.default-visibility is never applied, so /.well-known/agents/public misses agents that other endpoints treat as public
## Description

**Registry Version**: snapshot (main @ 8d9be8cf)

`apicurio.a2a.default-visibility` is documented as "Default visibility for new Agent Card artifacts" (`A2AConfig`, default `entitled`). Nothing in the codebase ever writes the `apicurio.agent.visibility` label when an Agent Card is created, so the property is not actually a creation-time default. It only exists as a read-time fallback in `WellKnownResourceImpl.resolveVisibility()`:

```java
private String resolveVisibility(Map<String, String> labels) {
    if (labels != null) {
        String explicit = labels.get("apicurio.agent.visibility");
        if (explicit != null) {
            return explicit.toLowerCase(Locale.ROOT);
        }
    }
    return a2aConfig.getDefaultVisibility().toLowerCase(Locale.ROOT);
}
```

`getPublicAgents` does not use that fallback. It filters with a strict SQL label match, so an artifact with no label can never match:

```java
filters.add(SearchFilter.ofLabel("apicurio.agent.visibility", "public"));
```

The result is that with `apicurio.a2a.default-visibility=public`, an unlabeled Agent Card is treated as public by `/.well-known/agents` and `/.well-known/agents/entitled` (both go through `resolveVisibility()`), but is invisible on `/.well-known/agents/public`, the endpoint whose entire purpose is public discovery.

Both behaviors were introduced together in #7927, so this looks like an oversight rather than a deliberate split.

### Verified behavior

Registry started with `apicurio.a2a.default-visibility=public`, one Agent Card created with no visibility label, then queried anonymously:

| Endpoint | status | count | unlabeled agent returned |
|---|---|---|---|
| `/.well-known/agents` | 200 | 1 | yes |
| `/.well-known/agents/entitled` | 200 | 1 | yes |
| `/.well-known/agents/public` | 200 | 0 | **no** |

### Environment

Local build from `main`, in-memory persistence, `apicurio.features.experimental.enabled=true`, `apicurio.a2a.enabled=true`, `apicurio.a2a.default-visibility=public`. Anonymous requests, no auth configured.

### Steps to Reproduce

1. Start the registry with `apicurio.a2a.enabled=true` and `apicurio.a2a.default-visibility=public`.
2. Create an Agent Card artifact without an `apicurio.agent.visibility` label.
3. `GET /.well-known/agents` returns the agent.
4. `GET /.well-known/agents/public` does not.

**Expected:** setting `default-visibility=public` makes unlabeled Agent Cards public everywhere, including on the public discovery endpoint. As it stands the property has no effect there, so an operator who sets it cannot tell from the docs why `/agents/public` stays empty.

**Actual:** `/.well-known/agents/public` only returns agents carrying an explicit `apicurio.agent.visibility=public` label, regardless of the configured default, while the sibling endpoints honor it.

The existing test `WellKnownResourceTest.testDefaultVisibilityExcludesFromPublic` only covers the shipped default (`entitled`), where both behaviors happen to agree, so the divergence is not currently caught.

### Possible directions

1. Have `getPublicAgents` resolve visibility the same way `getEntitledAgents` does (fetch, then filter with `resolveVisibility()`, bounded by the existing `MAX_VISIBILITY_FILTER_RESULTS` guard). Keeps one source of truth for visibility semantics, at the cost of the in-memory filtering that endpoint currently avoids.
2. Apply the default at creation time by stamping the label on new Agent Card artifacts, which would match the property's current description and keep `getPublicAgents` a cheap indexed query. Would not affect artifacts created before the change.
3. Leave the behavior and instead narrow the property's documentation so it is clear it only affects the entitled/search paths.

Happy to open a PR for whichever direction you prefer.

### Logs

N/A, the behavior is visible in the response bodies shown above.

owaish7:
## What

`GET /.well-known/agents/public` now honors `apicurio.a2a.default-visibility=public` for Agent Cards that have no explicit `apicurio.agent.visibility` label. When the configured default is public, the endpoint resolves each card's effective visibility in memory (same approach as `getEntitledAgents`) instead of filtering on the label in SQL. With any other default the existing indexed label query is kept unchanged.

## Why

Nothing writes the `apicurio.agent.visibility` label at creation time. The `apicurio.a2a.default-visibility` property only exists as a read-time fallback in `resolveVisibility()`, and `getPublicAgents` didn't use that fallback, it filtered on a strict SQL label match. So with `default-visibility=public`, an unlabeled agent showed up on `/.well-known/agents` and `/agents/entitled` but was missing from `/agents/public`, the endpoint whose whole purpose is public discovery. Details and verification in #9322.

Of the three directions listed on the issue, this implements the first one (resolve visibility at read time), since it keeps a single source of truth for visibility semantics and matches what the entitled endpoint already does. Happy to rework it toward label-stamping at creation or a docs-only clarification if maintainers prefer one of those, the change is small either way.

The label-filter fast path is preserved for the default configuration, so there is no behavior or performance change unless an operator explicitly sets `default-visibility=public`. The new path is bounded by the existing `MAX_VISIBILITY_FILTER_RESULTS` guard, same as `getEntitledAgents`.

Fixes #9322

## Testing

New `PublicDefaultVisibilityTest` (test profile with `apicurio.a2a.default-visibility=public`):

- unlabeled agent is returned by `/agents/public` (fails on main, passes with this change)
- explicit `private` label still excluded, so nothing leaks onto the unauthenticated endpoint
- explicit `public` label still returned
- pagination applies on the resolved-visibility path

Existing `WellKnownResourceTest` still passes (25 tests), including `testDefaultVisibilityExcludesFromPublic` which covers the shipped `entitled` default.

`./mvnw checkstyle:check -pl app`: 0 violations.


app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -715,6 +724,42 @@ private List<SearchedArtifactDto> filterDtosByVisibility(List<SearchedArtifactDt
...
+    private AgentSearchResults publicAgentsByResolvedVisibility(Set<SearchFilter> filters,
+            Integer offset, Integer limit) {
+        ArtifactSearchResultsDto results = storage.searchArtifacts(
+                filters, OrderBy.createdOn, OrderDirection.desc, 0, MAX_VISIBILITY_FILTER_RESULTS, false);
+        warnIfTruncated(results);
+
+        // Filter by resolved visibility on DTOs first (cheap), then paginate, then convert (expensive)
+        List<SearchedArtifactDto> visible = new ArrayList<>();
+        for (SearchedArtifactDto artifact : results.getArtifacts()) {
+            if ("public".equals(resolveVisibility(artifact.getLabels()))) {
```
paoloantinori: The routing check at line 143 lowercases the configured default before comparing, but this per-card comparison relies on `resolveVisibility(...)` already returning a lowercased value. If `resolveVisibility` returns the raw label (or a non-lowercased default), a card labeled `Public` is silently dropped, and when the default itself is configured as `Public` the endpoint returns empty. Worth confirming `resolveVisibility` normalizes case, and centralizing a single `isPubliclyVisible(labels)` (or a visibility enum) so the `"public"` literal and the `apicurio.agent.visibility` key aren't repeated across the SQL filter, the routing condition, and here.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -135,6 +135,15 @@ public AgentSearchResults getPublicAgents(Integer offset, Integer limit) {
 
         Set<SearchFilter> filters = new HashSet<>();
         filters.add(SearchFilter.ofArtifactType(ArtifactType.AGENT_CARD));
+
+        // Agent Cards without an explicit visibility label fall back to the configured default
+        // visibility (see resolveVisibility), so when that default is "public" they belong in
+        // these results even though no label filter can match them. Resolve visibility in memory
+        // in that case, as getEntitledAgents does. Otherwise keep the cheaper indexed label query.
+        if ("public".equals(a2aConfig.getDefaultVisibility().toLowerCase(Locale.ROOT))) {
```
paoloantinori: This dereferences `a2aConfig.getDefaultVisibility()` on every call, including the default-config path that previously never read it. The sibling comparison inside `publicAgentsByResolvedVisibility` is written null-safe (`"public".equals(...)`), which suggests null is a live concern somewhere in this area. Confirm `getDefaultVisibility()` can never return null (e.g. it's a `@ConfigProperty` with a default); if it can, this NPEs the public endpoint for all configurations, not just `default-visibility=public`.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/PublicDefaultVisibilityTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * An explicit label still wins over the configured default, so a private agent must not leak
+     * onto the unauthenticated public endpoint.
+     */
+    @Test
+    public void testExplicitPrivateLabelIsExcludedWhenDefaultIsPublic() throws Exception {
+        String groupId = TestUtils.generateGroupId();
+        createAgentCard(groupId, "explicit-private-agent", AGENT_CARD_CONTENT);
```
paoloantinori, line 108: This proves an explicit `private` label is excluded, but the no-leak guarantee for an unauthenticated endpoint also rests on `resolveVisibility` falling back to the default only when the visibility label is *absent*, not when it holds an unrecognized value. Add a test asserting a card labeled e.g. `apicurio.agent.visibility=internal` (or `restricted`) is excluded when `default-visibility=public`, so a later change to `resolveVisibility`'s fallback can't silently start publishing non-public cards.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/PublicDefaultVisibilityTest.java
```diff
@@ -0,0 +1,179 @@
...
+                .queryParam("offset", 0)
+                .queryParam("limit", 1)
+                .get("/.well-known/agents/public")
+                .then()
+                .statusCode(200)
+                .body("agents.size()", equalTo(1));
+    }
+
+    private void setVisibility(String groupId, String artifactId, String visibility) {
+        EditableArtifactMetaData meta = new EditableArtifactMetaData();
```
paoloantinori, line 209: Only `agents.size()` is asserted, but the javadoc says `count` still reflects every public agent and `count` is never checked. Add a `count` assertion too. Note it can't be an exact `equalTo(2)` without test-isolation care: storage is shared across the tests in this profile, so unlabeled cards created by other tests inflate the real total. Either isolate this case to its own group or assert `count` matches a follow-up request issued at a large limit.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -715,6 +724,42 @@ private List<SearchedArtifactDto> filterDtosByVisibility(List<SearchedArtifactDt
...
+    /**
+     * Returns public agents by resolving each artifact's effective visibility in memory, rather
+     * than filtering on the {@code apicurio.agent.visibility} label. Used when the configured
+     * default visibility is {@code public}, because artifacts relying on that default carry no
+     * visibility label and so cannot be matched by a label search filter.
+     */
+    private AgentSearchResults publicAgentsByResolvedVisibility(Set<SearchFilter> filters,
+            Integer offset, Integer limit) {
+        ArtifactSearchResultsDto results = storage.searchArtifacts(
+                filters, OrderBy.createdOn, OrderDirection.desc, 0, MAX_VISIBILITY_FILTER_RESULTS, false);
```
paoloantinori, line 992: Under `default-visibility=public` essentially every agent card resolves to public, so this scan-and-filter now runs over the full agent-card set on every unauthenticated request, and the `MAX_VISIBILITY_FILTER_RESULTS` cap becomes the effective ceiling for the endpoint: `count` understates the true total and pages beyond the cap return empty. This is inherited from `getEntitledAgents`, but the cardinality is higher here. If agent volume can grow, the stamp-label-at-creation direction from #9322 keeps the indexed SQL path O(1) at read time and avoids the in-memory scan on a public endpoint; at minimum the cap should be called out as a documented limit.

paoloantinori (commented):
The fix correctly makes unlabeled agents discoverable when `default-visibility=public`, and the labeled cases are well covered, but I'd hold merge for a few MAJOR items: the visibility comparison isn't guaranteed case-consistent, fail-closed behavior for unrecognized labels isn't tested, and the read-time scan inherits a count/pagination cliff that's more impactful on this unauthenticated endpoint now that every card is a public candidate by default.

- Case/DRY (`WellKnownResourceImpl.java:143`, `:742`): routing lowercases the configured default but the per-card check doesn't normalize the resolved value, and the `"public"` literal plus the `apicurio.agent.visibility` key are duplicated across three sites. Centralize the comparison and confirm case handling.
- Null safety (`WellKnownResourceImpl.java:143`): `getDefaultVisibility()` is dereferenced on every request, including the default-config path that never read it before; confirm it can't be null, or guard it.
- Security test gap (`PublicDefaultVisibilityTest.java:108`): only an explicit `private` label is tested; add a case for an unrecognized label value to prove non-public cards never leak when the default is public.
- Pagination contract (`PublicDefaultVisibilityTest.java:158`): `count` isn't asserted despite the javadoc claiming it reflects all public agents.
- Altitude (`WellKnownResourceImpl.java:736`): default-public makes every card a candidate, so the `MAX_VISIBILITY_FILTER_RESULTS` scan caps the endpoint and breaks pagination past it; consider the stamp-at-creation direction from #9322 to keep the indexed path on this unauthenticated endpoint.

owaish7:
Thanks @paoloantinori, this was a useful review, and the case question turned out to be a real problem rather than a style one. Pushed changes for four of the five points.

**Case consistency.** You were right, and it is worse than cosmetic:

- Labels reach `resolveVisibility()` from the serialized `labels` column via `SearchedArtifactMapper`, and `RegistryContentUtils.serializeLabels()` writes that column with the map as supplied, so the case a user typed is preserved.
- The `artifact_labels` table that label search filters query is lowercased on insert in `SqlArtifactRepository`, so the two paths disagree on the key.
- `labels.get(VISIBILITY_LABEL)` is an exact match, so a card labelled `Apicurio.Agent.Visibility=private` misses the lookup, falls through to the configured default, and with `default-visibility=public` resolves to public. That puts a card explicitly marked private on `/agents/public`, which is unauthenticated.

The label *value* was already handled, since `resolveVisibility()` lowercases whatever it returns, so the fix is scoped to the key. It now matches case-insensitively, with a test that labels a card `Apicurio.Agent.Visibility=private` and asserts it stays out of `/agents/public`. The test fails without the fix.

Worth flagging that this also changes `/agents` and `/agents/entitled`, which resolve visibility through `filterDtosByVisibility()`. The inconsistency was already there, but with the usual `entitled` default a missed lookup fails toward requiring auth, so it was latent. Making the endpoint default-public is what turns it into a fail-open path, so it seemed right to fix here rather than separately.

**The other three:**

- **DRY:** extracted `VISIBILITY_LABEL` and `VISIBILITY_PUBLIC` and used them at all three sites, including inside `resolveVisibility()`.
- **Unrecognized label:** added `testUnrecognizedLabelIsExcludedWhenDefaultIsPublic`, which labels a card `internal-only` and asserts it is absent from `/agents/public`, covering the fail-closed branch.
- **Pagination count:** the test now reads `count` from an unpaginated call and asserts the paginated call reports the same value while returning one agent, so the javadoc claim is pinned.

**Null safety.** I looked at this and think a guard would be unreachable:

- `defaultVisibility` is `@ConfigProperty(defaultValue = "entitled")`, so it is never unset.
- `A2AConfig.validate()` is `@PostConstruct` and dereferences it at startup, rejecting anything outside public/entitled/private.

A null or invalid value fails the application at boot rather than reaching a request, and a request-time check would cut against the "no redundant guards" item in the contributor checklist. Happy to add one anyway if you would rather have the belt and braces.

**Scale cliff.** You are right, and it is the part I am least comfortable with:

- The read-time scan inherits `MAX_VISIBILITY_FILTER_RESULTS` from `getEntitledAgents`, and default-public does make that cap bite harder since every card becomes a candidate.
- Stamping the label at creation keeps the indexed query with no cap, which is why I listed it as an option on #9322.
- The catch is that stamping only helps cards created after the change, so existing unlabeled cards would still need a backfill, and a migration that rewrites artifact labels across the storage variants is a good deal more than this PR is carrying today.

What I would suggest is keeping this as the correctness fix, so the endpoint stops under-reporting and the fail-open case is closed, and opening a separate issue for stamp-at-creation plus the backfill so the indexed path can come back. I am happy to pick that follow-up up. @carlesarnal wrote the visibility model in #7927 so he may have a view on the stamping direction, though I gather he is away at the moment.

paoloantinori:
Thanks, this is a thorough response and the fixes all check out on the new head.

The case-insensitive key fix is the important one — a card labeled `Apicurio.Agent.Visibility=private` resolving to the configured default (and landing on the unauthenticated `/agents/public` when that default is public) was a genuine fail-open path, and `testMixedCaseVisibilityLabelKeyIsHonoredWhenDefaultIsPublic` pins it with a test that fails without the fix. Good catch on the serialized `labels` column vs. the lowercased `artifact_labels` table disagreeing on the key.

On the null-safety point: agreed, I withdraw it. `@ConfigProperty(defaultValue = "entitled")` plus `A2AConfig.validate()` as `@PostConstruct` means an unset or invalid value fails at boot, not at request time, so a request-side guard there would be dead code against the "no redundant guards" checklist item.

On the scale cliff: your read matches mine. Keeping this as the correctness fix (closes the under-reporting and the fail-open case) and tracking stamp-at-creation plus the backfill separately is the right split — the backfill across storage variants is a different scope of change. Opening a follow-up issue for that direction sounds right.

owaish7:
Thanks @paoloantinori , appreciate the quick turnaround. I'll open the follow-up issue for stamp-at-creation plus the backfill and link it back here, and I'll carry over your point about documenting the `MAX_VISIBILITY_FILTER_RESULTS` cap as part of that.

paoloantinori:
Sounds good. Documenting the MAX_VISIBILITY_FILTER_RESULTS cap alongside the stamp-at-creation follow-up is the right place for it. I'll keep an eye out for the linked issue.

owaish7:
@EricWittmann please review.

EricWittmann:
@owaish7 This PR has a merge conflict.  Can you rebase and fix the conflict?

owaish7:
@EricWittmann   - Rebased onto main and force-pushed, conflicts resolved.

The main one was #9356 landing `A2AConstants`, which covers the same ground as the local visibility constants I had added for @paoloantinori's DRY point. I dropped mine and moved onto those, so the diff is a bit smaller. #9191's pagination clamping is preserved as-is on the indexed path, and `resolveVisibility` keeps the case-insensitive key match from this PR.

`PublicDefaultVisibilityTest` is 6/6 and `WellKnownResourceTest` is 33/33 locally, checkstyle clean. Behavior is unchanged from what you and Paolo reviewed.

paoloantinori:
Thanks for the rebase note. I saw the DRY point I raised on the local visibility constants is resolved cleanly: `#9356` landed `A2AConstants` on `main`, you dropped the duplicated locals, and moved onto the shared constants. That's the right outcome and the diff is tighter for it. No further comments from me on that thread. I'll leave the rebase / merge-conflict follow-up to Eric and will re-read once it's marked ready.

owaish7:
@EricWittmann  you can proceed.

EricWittmann (approved):
Reviewed the production change and test class in detail. The fix is correct, focused, and well-tested.

**What this fixes:** `getPublicAgents` filtered on the `apicurio.agent.visibility` label in SQL, so unlabeled Agent Cards were invisible on `/.well-known/agents/public` even when `apicurio.a2a.default-visibility=public`. The new read-time resolution path matches what `getEntitledAgents` already does, and the indexed SQL fast path is preserved for all other default configurations.

**Security fix in `resolveVisibility`:** The case-insensitive key match closes a real fail-open path — a card labeled `Apicurio.Agent.Visibility=private` (mixed case) would have missed the `map.get()` lookup, fallen through to the configured default, and resolved to `public` on the unauthenticated endpoint. The `testMixedCaseVisibilityLabelKeyIsHonoredWhenDefaultIsPublic` test pins this.

**Tests:** Good coverage across 6 cases — unlabeled-is-public, explicit-private-excluded, unrecognized-label-excluded, mixed-case-key-honored, explicit-public-included, and pagination with count assertion.

No bugs found. LGTM.

owaish7:
Thanks @EricWittmann .
What is the next step?

EricWittmann:
Sorry I was waiting for CI to pass.  All good and merged. Thanks @owaish7 :t-rex:

PR #9344, opened 2026-08-07
fix(rest): clamp pagination offset and limit in GroupsResourceImpl

nidhiii128:
While looking into this issue, I noticed that `GroupsResourceImpl` was still converting `offset` and `limit` using `.intValue()`, which can silently truncate out-of-range values. The same issue had already been fixed in `SearchResourceImpl` as part of #8611, but these endpoints were still using the old approach.

This PR moves `normalizeOffset` and `normalizeLimit` into `V3ApiUtil` so they can be shared, updates `SearchResourceImpl` to use the common helpers, and applies the same validation to the affected group endpoints. I also fixed `getContractAuditLog`, which had the same `.intValue()` pattern.

Verified with `./mvnw -pl app -am compile -DcliSkipNative` (clean build).


paoloantinori (commented):
Direction is right: replacing raw `.intValue()` on the BigInteger pagination params with shared clamp helpers fixes real silent truncation (an offset of 2^32+5 used to wrap to 5) and aligns the group endpoints with the #8611 fix. Two behavior changes need a deliberate decision (the audit log page cap and null handling), the PR ships no tests, and the description does not match what the diff actually shows.

- `GroupsResourceImpl.java:2712` changes the contract audit log's effective max page size from a hard 500 to `normalizeLimit`'s cap (1000 per the #8611 search implementation, worth double-checking the shared helper). If 500 was deliberate for audit entries, this doubles the worst-case page size.
- `GroupsResourceImpl.java:1703` and `:2711` apply the helpers' null contract inconsistently: `limit` is null-guarded just above (the `if (limit == null)` block) but `offset` is passed unguarded on the same line, and `getContractAuditLog` guards both before calling the same helpers. Not a regression (the old `offset.intValue()` had the same null exposure), but this refactor is the place to settle one pattern.
- The description says the helpers were moved into `V3ApiUtil` and that `SearchResourceImpl` was updated to use them, but every call in the diff references `ParameterValidationUtils` and the changed-files list contains only `GroupsResourceImpl.java`. If the move and the `SearchResourceImpl` update are not actually in this PR, the helpers are not shared and #8611's private copies remain.
- Verification is compile-only. No test covers the new clamping on any of the six endpoints; worth mirroring whatever #8611 added (values past 2^31, negatives, omitted params) and running the app REST tests rather than `compile` alone.
- The rewritten call lines run roughly 135 characters, likely past the project's line-length limit, and the same two-line expression is repeated five times (739, 1345, 1578, 1703, 1767); hoisting `off`/`lim` locals fixes both.
- `GroupsResourceImpl.java:2694` is a whitespace-only hunk in `executeContractRules`, unrelated to this fix; drop it.
- Worth grepping the codebase for remaining `.intValue()` pagination conversions (the v2 REST resources still exist) so this lands as a complete fix rather than endpoint-by-endpoint.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -2707,8 +2708,8 @@ public Response getContractAuditLog(String groupId, String artifactId,
         ParameterValidationUtils.requireParameter("groupId", groupId);
         ParameterValidationUtils.requireParameter("artifactId", artifactId);
 
-        int off = offset != null ? offset.intValue() : 0;
-        int lim = limit != null ? Math.min(limit.intValue(), 500) : 20;
+        int off = offset != null ? ParameterValidationUtils.normalizeOffset(offset) : 0;
+        int lim = limit != null ? ParameterValidationUtils.normalizeLimit(limit) : 20;
```
paoloantinori: This silently changes the contract audit log's max page size from a hard 500 to whatever `normalizeLimit` clamps to (1000 in the #8611 search endpoints, if that is the implementation now being shared). Audit entries are heavier than search rows, so doubling the worst-case page is a real behavior change and not just input hygiene. If 500 was deliberate here, keep an explicit bound (`Math.min(ParameterValidationUtils.normalizeLimit(limit), 500)`) or add an overload such as `normalizeLimit(limit, max)` so the endpoint retains its own cap.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -1700,8 +1700,8 @@ public BranchSearchResults listBranches(String groupId, String artifactId, BigIn
             limit = BigInteger.valueOf(20);
         }
 
-        BranchSearchResultsDto dto = storage.getBranches(new GA(groupId, artifactId), offset.intValue(),
-                limit.intValue());
+        BranchSearchResultsDto dto = storage.getBranches(new GA(groupId, artifactId), ParameterValidationUtils.normalizeOffset(offset),
```
paoloantinori: The helpers' null contract is being applied inconsistently. This method null-guards `limit` (the `if (limit == null)` block above) yet passes `offset` straight into `normalizeOffset`, and `getContractAuditLog` (line 2711) null-guards both before calling the same helpers. Either the helpers are null-safe, in which case those guards are dead code and can go, or they are not, in which case any of these endpoints whose query param lacks a `@DefaultValue` will turn an omitted `?offset=` into a 500. Not a regression (the old `offset.intValue()` had the same exposure), but please state the null contract on the helpers and use one pattern across all six call sites.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -736,8 +736,8 @@ public GroupSearchResults listGroups(BigInteger limit, BigInteger offset, SortOr
 
         Set<SearchFilter> filters = Collections.emptySet();
 
-        GroupSearchResultsDto resultsDto = storage.searchGroups(filters, oBy, oDir, offset.intValue(),
-                limit.intValue());
+        GroupSearchResultsDto resultsDto = storage.searchGroups(filters, oBy, oDir, ParameterValidationUtils.normalizeOffset(offset),
```
paoloantinori: Six endpoints get new clamping behavior and the PR ships no tests for it. Please add coverage mirroring what #8611 added for the search endpoints: offset/limit at 2^40 (the silent truncation case this PR fixes), `Integer.MAX_VALUE + 1`, negative values, and omitted params, asserting the clamped values actually reach the storage layer rather than just a 200 response. `./mvnw compile` does not exercise any of this.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -1342,8 +1342,8 @@ public ArtifactSearchResults listArtifactsInGroup(String groupId, BigInteger lim
         Set<SearchFilter> filters = new HashSet<>();
         filters.add(SearchFilter.ofGroupId(new GroupId(groupId).getRawGroupIdWithNull()));
 
-        ArtifactSearchResultsDto resultsDto = storage.searchArtifacts(filters, oBy, oDir, offset.intValue(),
-                limit.intValue(), skipCount != null && skipCount);
+        ArtifactSearchResultsDto resultsDto = storage.searchArtifacts(filters, oBy, oDir, ParameterValidationUtils.normalizeOffset(offset),
```
paoloantinori: These rewritten lines run about 135 characters, which is likely past the project's line-length limit and will trip the formatting gate that compile-only verification does not run. Hoisting locals (`int off = ParameterValidationUtils.normalizeOffset(offset); int lim = ParameterValidationUtils.normalizeLimit(limit);`) fixes the length and removes the identical two-line expression repeated at 739, 1345, 1578, 1703 and 1767.

app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -2691,6 +2691,7 @@ public Response executeContractRules(String groupId, String artifactId,
         if (result.getTransformedRecord() != null) {
             responseMap.put("transformedRecord", result.getTransformedRecord());
         }
+
```
paoloantinori: Unrelated whitespace-only hunk in `executeContractRules`, a method this PR otherwise does not touch. Drop it to keep the diff limited to the pagination change.

vanshika2720:
@nidhiii128 The current implementation looks correct, but the PR description is out of sync with the actual diff. normalizeOffset/normalizeLimit are not being moved to V3ApiUtil, and SearchResourceImpl is not modified by this PR. Please update the description to reflect the current changes.

carlesarnal (commented):
## Review Summary

**What it does:** Replaces raw `BigInteger.intValue()` conversions on pagination `offset`/`limit` query params in `GroupsResourceImpl` (6 call sites) with `ParameterValidationUtils.normalizeOffset`/`normalizeLimit`, which clamp instead of silently truncating on overflow. Adds a `normalizeLimit(BigInteger, int)` overload so `getContractAuditLog` keeps its distinct 500-row cap.

### Strengths
- Fixes a real correctness bug: silent `BigInteger→int` truncation could let a huge client-supplied offset/limit wrap around to a small value reaching the storage layer.
- Reuses the existing shared `ParameterValidationUtils` rather than duplicating logic.
- The audit-log page-size regression flagged earlier (silently doubling the cap from 500→1000) was properly fixed via the new overload.
- Solid new integration test coverage (`GroupsPaginationTest`) exercising negative values, `limit=0`, overflow past `Integer.MAX_VALUE`, and omitted-parameter defaults across all six affected endpoints.
- Unit tests added for the new `normalizeLimit(limit, maxLimit)` overload.
- No storage-layer/cross-variant impact — REST layer only.

### Concerns
- `normalizeOffset`/`normalizeLimit` are not null-safe, and null-guarding is inconsistent: `getContractAuditLog` guards with `!= null` before calling the helper, but the other five call sites (`listGroups`, `listArtifactsInGroup`, `listArtifactVersions`, `listBranches`, `listBranchVersions`) do not appear to guard explicitly. Please confirm these query params have `@DefaultValue` at the JAX-RS layer so null can never reach the helper — otherwise this could leak an NPE as a raw 500 instead of a clean 4xx. If `@DefaultValue` is present everywhere, this is moot; if not, please standardize (either add defaults or guard uniformly).

### Suggestions (non-blocking)
- Document the non-null expectation on `normalizeOffset`/`normalizeLimit`.
- Consider a follow-up issue to grep for remaining `.intValue()` pagination conversions elsewhere in the codebase (v2 resources, other v3 resources).

Overall this looks good and the previously raised concerns (cap regression, missing tests) have been addressed. Please just confirm the null-safety point above before merge.

carlesarnal:
No PR without a linked issue, I'm sorry. Re-open if needed.

PR #9347, opened 2026-08-07
fix: add URL format validation for AGENT_CARD artifacts (#8731)

### LINKED ISSUE #8731: `AgentCardContentValidator` does not validate URL format — any string passes as a URL
# Issue Body

### Describe the bug

The `AgentCardContentValidator` checks that URL fields in `supportedInterfaces` and `provider` exist and are strings, but never validates URL format. An agent card with `"url": "not-a-url"` or `"url": "hello world"` passes `FULL` validation and gets stored.

This means downstream code (A2A orchestrator, webhook delivery in `HttpClientService`) will try to connect to garbage URLs at runtime and fail with confusing errors, instead of being rejected at registration time.

The optional `iconUrl` and `documentationUrl` fields have the same gap — they're only checked for string type, not URL format.

### Expected behavior

At `FULL` validation level, URL fields should be validated as well-formed HTTP(S) URLs. Malformed URLs should produce a `RuleViolation` and reject the agent card.

### Steps to reproduce

1. POST an agent card with:
   ```json
   "supportedInterfaces": [{
     "url": "not-a-url",
     "protocolBinding": "http+json",
     "protocolVersion": "1.0"
   }]
   ```
2. Validation passes — the invalid URL gets stored
3. Any attempt to reach the agent at runtime fails

### Root cause

**File:** `schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java`

In `validateSupportedInterfaces()` (lines 182–185):

```java
if (!iface.has("url") || !iface.get("url").isTextual()) {
    violations.add(new RuleViolation(
            "Interface 'url' is required and must be a string", basePath + "/url"));
}
// No format validation — "not-a-url" passes here
```

Same pattern in `validateProviderField()` (lines 159–162):

```java
if (!provider.has("url") || !provider.get("url").isTextual()) {
    violations.add(new RuleViolation(
            "'provider.url' is required and must be a string", "/provider/url"));
}
// No format validation
```

And in `validateStringFields()` (lines 138–139), the optional `iconUrl` and `documentationUrl` also only get a string-type check via `validateOptionalString()`.

### Proposed fix

Add URL format validation after the existing string-type check passes (not replacing it). For `validateSupportedInterfaces()`:

```java
if (!iface.has("url") || !iface.get("url").isTextual()) {
    violations.add(new RuleViolation(
            "Interface 'url' is required and must be a string", basePath + "/url"));
} else {
    // Validate URL format
    String urlValue = iface.get("url").asText();
    try {
        java.net.URI uri = new java.net.URI(urlValue);
        if (uri.getScheme() == null || !uri.getScheme().matches("https?")) {
            violations.add(new RuleViolation(
                "Interface 'url' must use http or https scheme", basePath + "/url"));
        }
        if (uri.getHost() == null) {
            violations.add(new RuleViolation(
                "Interface 'url' must have a valid host", basePath + "/url"));
        }
    } catch (java.net.URISyntaxException e) {
        violations.add(new RuleViolation(
            "Interface 'url' is not a valid URL: " + e.getMessage(), basePath + "/url"));
    }
}
```

Apply the same pattern to `provider.url`, and a relaxed version (scheme check only, host optional) to `iconUrl` and `documentationUrl`.

Happy to submit a PR for this.

### Environment

- Component: `schema-util/common`
- Artifact type: A2A Agent Card

DSingh0304:
Fixes #8731.

`AgentCardContentValidator` checked URL fields (`supportedInterfaces[*].url`, `provider.url`, `iconUrl`, `documentationUrl`) only for string type -never for valid URL format. This meant an agent card with `"url": "not-a-url"` or `"url": "hello world"` would pass `FULL` validation and get stored. Downstream code (A2A orchestrator, `HttpClientService`) would then try to connect to garbage URLs at runtime and fail with confusing errors instead of being rejected at registration time.

This PR adds HTTP(S) URL format validation at `FULL` validation level for the two endpoint URL fields (`supportedInterfaces[*].url` and `provider.url`). The optional display fields `iconUrl` and `documentationUrl` are validated only as strings, not as HTTP(S) URLs, because the A2A agent card spec does not restrict those fields to http/https (e.g. `data:` URIs are valid for inline icons).

---

## Changes

**`schema-util/common` -`JsonValidationUtils`**

Added one reusable helper method to avoid duplication across validators:

- `validateHttpUrl(String value, String path, Set<RuleViolation> violations)` - parses the value with `java.net.URI`, checks that the scheme is `http` or `https` (case-insensitive per RFC 3986 §3.1), that a host is present, that the port (if specified) is within 0–65535, and catches `URISyntaxException` for malformed input. Uses only `java.base` - no new dependencies.

**`schema-util/common` - `AgentCardContentValidator`**

- `validateSupportedInterfaces()`: after the existing presence/type check passes, now calls `JsonValidationUtils.validateHttpUrl(...)` on `iface.url`.
- `validateProviderField()`: same - after the existing presence/type check passes, now calls `JsonValidationUtils.validateHttpUrl(...)` on `provider.url`.
- `validateStringFields()`: `iconUrl` and `documentationUrl` remain validated as strings only (`validateOptionalString`). HTTP(S) format validation is intentionally not applied here - the spec permits non-http(s) values for these display fields.

**`schema-util/util-provider` -test fixtures & unit tests**

Added three test fixtures and three matching unit test cases to `AgentCardContentValidatorTest`:

| Fixture | Scenario |
|---|---|
| `agentcard-invalid-interface-url.json` | `supportedInterfaces[0].url` is `"not-a-url"` (no scheme) |
| `agentcard-invalid-provider-url.json` | `provider.url` uses `ftp://` scheme (not `http`/`https`) |
| `agentcard-invalid-optional-urls.json` | `iconUrl` is a number (`12345`) and `documentationUrl` is a boolean (`false`) - tests that non-string values are rejected |

Each test asserts `RuleViolationException` is thrown and that the violation context path points to the correct field.

---

## Test results

All 22 tests in `AgentCardContentValidatorTest` pass, including the 3 new cases. Checkstyle: 0 violations.

```
Tests run: 22, Failures: 0, Errors: 0, Skipped: 0
You have 0 Checkstyle violations.
BUILD SUCCESS
```

---

## Checklist

- [x] Fixes #8731
- [x] Follows existing `JsonValidationUtils` pattern (no ad-hoc validation inline)
- [x] No new external dependencies (`java.net.URI` is from `java.base`)
- [x] Unit tests added with fixtures for each validation scenario
- [x] Checkstyle passes (0 violations)
- [x] Commit signed off (`Signed-off-by: Deep Shekhar Singh <deepshekhar0306@gmail.com>`)
- [x] Rebased on latest `upstream/main`

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java
@@ -135,8 +135,8 @@ public class AgentCardContentValidator implements ContentValidator {
 
     private void validateStringFields(JsonNode tree, Set<RuleViolation> violations) {
         JsonValidationUtils.validateOptionalString(tree, "protocolVersion", violations);
-        JsonValidationUtils.validateOptionalString(tree, "iconUrl", violations);
-        JsonValidationUtils.validateOptionalString(tree, "documentationUrl", violations);
+        JsonValidationUtils.validateOptionalUrl(tree, "iconUrl", violations);
+        JsonValidationUtils.validateOptionalUrl(tree, "documentationUrl", violations);
     }
 
     private void validateProviderField(JsonNode tree, Set<RuleViolation> violations) {
@@ -159,6 +159,8 @@ public class AgentCardContentValidator implements ContentValidator {
         if (!provider.has("url") || !provider.get("url").isTextual()) {
             violations.add(new RuleViolation(
                     "'provider.url' is required and must be a string", "/provider/url"));
+        } else {
+            JsonValidationUtils.validateHttpUrl(provider.get("url").asText(), "/provider/url", violations);
         }
     }
 
@@ -182,6 +184,8 @@ public class AgentCardContentValidator implements ContentValidator {
             if (!iface.has("url") || !iface.get("url").isTextual()) {
                 violations.add(new RuleViolation(
                         "Interface 'url' is required and must be a string", basePath + "/url"));
+            } else {
+                JsonValidationUtils.validateHttpUrl(iface.get("url").asText(), basePath + "/url", violations);
             }
 
             if (!iface.has("protocolBinding") || !iface.get("protocolBinding").isTextual()) {
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java
@@ -135,6 +135,9 @@ public class AgentCardContentValidator implements ContentValidator {
 
     private void validateStringFields(JsonNode tree, Set<RuleViolation> violations) {
         JsonValidationUtils.validateOptionalString(tree, "protocolVersion", violations);
+        // iconUrl and documentationUrl are optional string fields; the A2A agent card spec does
+        // not restrict these to http(s) only (e.g. data: URIs are valid for inline icons),
+        // so we validate only that they are strings rather than enforcing a URL scheme.
         JsonValidationUtils.validateOptionalString(tree, "iconUrl", violations);
         JsonValidationUtils.validateOptionalString(tree, "documentationUrl", violations);
     }
@@ -159,6 +162,8 @@ public class AgentCardContentValidator implements ContentValidator {
         if (!provider.has("url") || !provider.get("url").isTextual()) {
             violations.add(new RuleViolation(
                     "'provider.url' is required and must be a string", "/provider/url"));
+        } else {
+            JsonValidationUtils.validateHttpUrl(provider.get("url").asText(), "/provider/url", violations);
         }
     }
 
@@ -182,6 +187,8 @@ public class AgentCardContentValidator implements ContentValidator {
             if (!iface.has("url") || !iface.get("url").isTextual()) {
                 violations.add(new RuleViolation(
                         "Interface 'url' is required and must be a string", basePath + "/url"));
+            } else {
+                JsonValidationUtils.validateHttpUrl(iface.get("url").asText(), basePath + "/url", violations);
             }
 
             if (!iface.has("protocolBinding") || !iface.get("protocolBinding").isTextual()) {
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java
@@ -3,6 +3,8 @@ package io.apicurio.registry.rules.validity;
 import com.fasterxml.jackson.databind.JsonNode;
 import io.apicurio.registry.rules.violation.RuleViolation;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Set;
 
 /**
@@ -24,6 +26,54 @@ public final class JsonValidationUtils {
             violations.add(
                     new RuleViolation("'" + fieldName + "' field must be a string", "/" + fieldName));
         }
+    }
+
+    /**
+     * Validates that a string value is a well-formed HTTP or HTTPS URL.
+     *
+     * <p>Per RFC 3986 section 3.1, URI schemes are case-insensitive, so the scheme is
+     * normalized to lowercase before comparison. Additionally, {@link URI#getHost()} returns
+     * {@code null} for IPv6 literal hosts (e.g. {@code http://[::1]:8080/}) on some JDK
+     * versions; the authority component is used as a fallback for the presence check.
+     */
+    public static void validateHttpUrl(String value, String path, Set<RuleViolation> violations) {
+        try {
+            URI uri = new URI(value);
+            String scheme = uri.getScheme();
+            // Normalize to lowercase per RFC 3986 §3.1 — schemes are case-insensitive
+            if (scheme == null || !scheme.toLowerCase(java.util.Locale.ROOT).matches("https?")) {
+                violations.add(new RuleViolation("URL must use http or https scheme", path));
+            } else {
+                // URI.getHost() returns null for IPv6 literals (e.g. http://[::1]:8080/).
+                // Fall back to getAuthority() so valid IPv6 URLs are not incorrectly rejected.
+                String host = uri.getHost();
+                String authority = uri.getAuthority();
+                boolean hasHost = (host != null && !host.trim().isEmpty())
+                        || (authority != null && !authority.trim().isEmpty());
+                if (!hasHost) {
+                    violations.add(new RuleViolation("URL must have a valid host", path));
+                }
+            }
+        } catch (URISyntaxException e) {
+            violations.add(new RuleViolation("Invalid URL format: " + e.getMessage(), path));
+        }
+    }
+
+
+    /**
+     * Validates that an optional field, if present, is a string and a well-formed HTTP(S) URL.
+     */
+    public static void validateOptionalUrl(JsonNode tree, String fieldName,
+            Set<RuleViolation> violations) {
+        if (!tree.has(fieldName)) {
+            return;
+        }
+        if (!tree.get(fieldName).isTextual()) {
+            violations.add(
+                    new RuleViolation("'" + fieldName + "' field must be a string", "/" + fieldName));
+            return;
+        }
+        validateHttpUrl(tree.get(fieldName).asText(), "/" + fieldName, violations);
     }
 
     /**
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java
@@ -3,6 +3,8 @@ package io.apicurio.registry.rules.validity;
 import com.fasterxml.jackson.databind.JsonNode;
 import io.apicurio.registry.rules.violation.RuleViolation;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Set;
 
 /**
@@ -23,6 +25,41 @@ public final class JsonValidationUtils {
         if (tree.has(fieldName) && !tree.get(fieldName).isTextual()) {
             violations.add(
                     new RuleViolation("'" + fieldName + "' field must be a string", "/" + fieldName));
+        }
+    }
+
+    /**
+     * Validates that a string value is a well-formed HTTP or HTTPS URL.
+     *
+     * <p>Per RFC 3986 section 3.1, URI schemes are case-insensitive, so the scheme is
+     * compared with {@link String#equalsIgnoreCase} rather than regex. The port, when
+     * present, is validated to be within the valid range (0–65535).
+     */
+    public static void validateHttpUrl(String value, String path, Set<RuleViolation> violations) {
+        try {
+            URI uri = new URI(value);
+            String scheme = uri.getScheme();
+            // Per RFC 3986 §3.1 schemes are case-insensitive — use equalsIgnoreCase to avoid
+            // a per-call regex compile and a toLowerCase allocation.
+            if (scheme == null || (!"http".equalsIgnoreCase(scheme) && !"https".equalsIgnoreCase(scheme))) {
+                violations.add(new RuleViolation("URL must use http or https scheme", path));
+            } else {
+                // URI.getHost() returns a non-null value for IPv6 literals (e.g. http://[::1]:8080/)
+                // on all standard JDKs, so getAuthority() fallback is not needed.
+                String host = uri.getHost();
+                if (host == null || host.isEmpty()) {
+                    violations.add(new RuleViolation("URL must have a valid host", path));
+                } else {
+                    int port = uri.getPort();
+                    // URI.getPort() returns -1 (no port specified) or a non-negative integer.
+                    // After the != -1 guard, port < 0 can never be true; only check the upper bound.
+                    if (port != -1 && port > 65535) {
+                        violations.add(new RuleViolation("URL port must be in the range 0–65535", path));
+                    }
+                }
+            }
+        } catch (URISyntaxException e) {
+            violations.add(new RuleViolation("Invalid URL format: " + e.getMessage(), path));
         }
     }
 
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/AgentCardContentValidatorTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/AgentCardContentValidatorTest.java
@@ -200,4 +200,58 @@ public class AgentCardContentValidatorTest extends ArtifactUtilProviderTestBase 
         Assertions.assertTrue(
                 error.getCauses().stream().anyMatch(v -> v.getDescription().contains("description")));
     }
+
+    @Test
+    public void testAgentCardInvalidInterfaceUrl() throws Exception {
+        TypedContent content = resourceToTypedContentHandle("agentcard-invalid-interface-url.json");
+        AgentCardContentValidator validator = new AgentCardContentValidator();
+        RuleViolationException error = Assertions.assertThrows(RuleViolationException.class, () -> {
+            validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
+        });
+        Assertions.assertFalse(error.getCauses().isEmpty());
+        Assertions.assertTrue(
+                error.getCauses().stream().anyMatch(v -> v.getContext().equals("/supportedInterfaces/0/url")));
+    }
+
+    @Test
+    public void testAgentCardInvalidProviderUrl() throws Exception {
+        TypedContent content = resourceToTypedContentHandle("agentcard-invalid-provider-url.json");
+        AgentCardContentValidator validator = new AgentCardContentValidator();
+        RuleViolationException error = Assertions.assertThrows(RuleViolationException.class, () -> {
+            validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
+        });
+        Assertions.assertFalse(error.getCauses().isEmpty());
+        Assertions.assertTrue(
+                error.getCauses().stream().anyMatch(v -> v.getContext().equals("/provider/url")));
+    }
+
+    @Test
+    public void testAgentCardInvalidOptionalUrls() throws Exception {
+        TypedContent content = resourceToTypedContentHandle("agentcard-invalid-optional-urls.json");
+        AgentCardContentValidator validator = new AgentCardContentValidator();
+        RuleViolationException error = Assertions.assertThrows(RuleViolationException.class, () -> {
+            validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
+        });
+        Assertions.assertFalse(error.getCauses().isEmpty());
+        Assertions.assertTrue(
+                error.getCauses().stream().anyMatch(v -> v.getContext().equals("/iconUrl")));
+        Assertions.assertTrue(
+                error.getCauses().stream().anyMatch(v -> v.getContext().equals("/documentationUrl")));
+    }
+
+    @Test
+    public void testAgentCardUppercaseSchemeUrl() throws Exception {
+        // RFC 3986 §3.1: schemes are case-insensitive; HTTP:// must be accepted
+        TypedContent content = resourceToTypedContentHandle("agentcard-uppercase-scheme-url.json");
+        AgentCardContentValidator validator = new AgentCardContentValidator();
+        validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
+    }
+
+    @Test
+    public void testAgentCardIpv6Url() throws Exception {
+        // URI.getHost() returns null for IPv6 literals; valid IPv6 URLs must not be rejected
+        TypedContent content = resourceToTypedContentHandle("agentcard-ipv6-url.json");
+        AgentCardContentValidator validator = new AgentCardContentValidator();
+        validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
+    }
 }
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/AgentCardContentValidatorTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/AgentCardContentValidatorTest.java
@@ -200,4 +200,61 @@ public class AgentCardContentValidatorTest extends ArtifactUtilProviderTestBase 
         Assertions.assertTrue(
                 error.getCauses().stream().anyMatch(v -> v.getDescription().contains("description")));
     }
+
+    @Test
+    public void testAgentCardInvalidInterfaceUrl() throws Exception {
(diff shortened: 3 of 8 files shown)


DSingh0304:
Hi @EricWittmann @paoloantinori, quick ping on this! I've opened a PR fixing #8731 by adding URL format validation for `AgentCardContentValidator`. All tests pass, and Checkstyle is clean would appreciate your review when you get a chance!

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java
```diff
@@ -26,6 +28,39 @@ public static void validateOptionalString(JsonNode tree, String fieldName,
...
     }
 
+    /**
+     * Validates that a string value is a well-formed HTTP or HTTPS URL.
+     */
+    public static void validateHttpUrl(String value, String path, Set<RuleViolation> violations) {
+        try {
+            URI uri = new URI(value);
+            String scheme = uri.getScheme();
+            if (scheme == null || !scheme.matches("https?")) {
```
paoloantinori: The scheme check is case-sensitive. `URI.getScheme()` preserves the original case, and `"https?"` as a regex against `matches()` requires lowercase. Per RFC 3986 section 3.1, schemes are case-insensitive, so a valid `HTTP://example.com` or `HtTpS://...` URL gets rejected here as an invalid scheme. Most clients lowercase the scheme, but the validator should match the spec. Consider normalizing with `scheme.toLowerCase()` before the check.

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java
```diff
@@ -26,6 +28,39 @@ public static void validateOptionalString(JsonNode tree, String fieldName,
...
+                violations.add(new RuleViolation("URL must use http or https scheme", path));
+            } else if (uri.getHost() == null || uri.getHost().trim().isEmpty()) {
```
paoloantinori: `URI.getHost()` returns null for IPv6 literal hosts (for example `https://[::1]:8080/`), a JDK quirk where the host is exposed via `getAuthority()` rather than `getHost()`. With this code, a valid IPv6 URL (common for localhost dev at `http://[::1]:8080/` and IPv6-only deployments) would be rejected as "must have a valid host". Worth handling by falling back to `getAuthority()` for the presence check, or normalizing brackets. A quick test against `http://[::1]/` would confirm whether your target JDK exhibits the null-return before deciding how to handle it.

paoloantinori (commented):
Looking at this for #8731. The validation structure is clean and the test coverage is solid, all four URL fields are covered with both happy-path and rejection fixtures. Two edge cases on `validateHttpUrl` are worth tightening before this lands, both tied to how `java.net.URI` parses schemes and hosts.

paoloantinori:
Reviewed. Two edge cases flagged inline on the URL scheme/host validation in `JsonValidationUtils` worth addressing before merge: case-sensitive scheme matching (RFC 3986 makes schemes case-insensitive) and `URI.getHost()` returning null for IPv6 literals.

DSingh0304:
> Reviewed. Two edge cases flagged inline on the URL scheme/host validation in `JsonValidationUtils` worth addressing before merge: case-sensitive scheme matching (RFC 3986 makes schemes case-insensitive) and `URI.getHost()` returning null for IPv6 literals.

Thanks for the review @paoloantinori! Great catch on both edge cases.

I've updated the PR to:
1. Normalize schemes to lowercase (`Locale.ROOT`) per RFC 3986 §3.1.
2. Fall back to `getAuthority()` so IPv6 literal hosts are not rejected.

Added test cases for both scenarios (`HTTP://` scheme and `http://[::1]:8080/`). All 24 tests pass!

paoloantinori:
Both edge cases are addressed cleanly on this head: `scheme.toLowerCase(Locale.ROOT)` handles the RFC 3986 case-insensitivity, and the `getAuthority()` fallback covers the IPv6 literal case where `getHost()` returns null. The Javadoc on `validateHttpUrl` documents both rationales, and the two new test fixtures (`testAgentCardUppercaseSchemeUrl`, `testAgentCardIpv6Url`) pin both behaviors down. No further comments from me — looks good to merge.

DSingh0304:
> Both edge cases are addressed cleanly on this head: `scheme.toLowerCase(Locale.ROOT)` handles the RFC 3986 case-insensitivity, and the `getAuthority()` fallback covers the IPv6 literal case where `getHost()` returns null. The Javadoc on `validateHttpUrl` documents both rationales, and the two new test fixtures (`testAgentCardUppercaseSchemeUrl`, `testAgentCardIpv6Url`) pin both behaviors down. No further comments from me — looks good to merge.

Thanks a lot for the thorough review and guidance @paoloantinori! I'll wait for @EricWittmann for any further directions and the final merge. It's been great collaborating with you and the team on this. I've also applied for the LFX Mentorship (Webhooks project) and really hope to continue contributing and working together more closely!

paoloantinori:
Sounds good @DSingh0304. Both edge cases are addressed and the test coverage pins them down. Nothing further from my side, ready for a maintainer to review when CI is green.

DSingh0304:
Hey @EricWittmann can you please` /accept` the PR when you are free & if there are no blockers. Thank you!!

DSingh0304:
> Hey @EricWittmann can you please` /accept` the PR when you are free & if there are no blockers. Thank you!!

Hi @EricWittmann The PR is up to date now...

DSingh0304:
Hi @EricWittmann I have updated the PR according to the latest commit.

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java
```diff
@@ -26,6 +28,54 @@ public static void validateOptionalString(JsonNode tree, String fieldName,
...
+            String scheme = uri.getScheme();
+            // Normalize to lowercase per RFC 3986 §3.1 — schemes are case-insensitive
+            if (scheme == null || !scheme.toLowerCase(java.util.Locale.ROOT).matches("https?")) {
+                violations.add(new RuleViolation("URL must use http or https scheme", path));
+            } else {
+                // URI.getHost() returns null for IPv6 literals (e.g. http://[::1]:8080/).
+                // Fall back to getAuthority() so valid IPv6 URLs are not incorrectly rejected.
+                String host = uri.getHost();
+                String authority = uri.getAuthority();
+                boolean hasHost = (host != null && !host.trim().isEmpty())
```
paoloantinori: The host-presence check has two gaps that let invalid endpoint URLs through, which defeats the PR's goal of catching bad URLs at registration time. First, the `getAuthority()` fallback is over-broad: for an authority that parses as a registry name rather than a server (for example `http://host_`), `java.net.URI.getHost()` returns null while `getAuthority()` returns `"host_"`, so `hasHost` is true and the value passes with no usable hostname. The fallback's stated purpose does not apply either, since `new URI("http://[::1]:8080/").getHost()` returns `"[::1]"` (non-null) on standard JDKs, so the IPv6 comment is incorrect and the new `testAgentCardIpv6Url` passes via `getHost()` without exercising the fallback. Second, there is no port-range check: `http://example.com:99999` parses with port 99999 and passes, then fails at the socket layer. Consider dropping the authority fallback and relying on `getHost()` alone, plus a check that the port (when present) is within 0 to 65535.

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java
```diff
@@ -135,8 +135,8 @@ private void validateRequiredString(JsonNode tree, String fieldName,
 
     private void validateStringFields(JsonNode tree, Set<RuleViolation> violations) {
         JsonValidationUtils.validateOptionalString(tree, "protocolVersion", violations);
-        JsonValidationUtils.validateOptionalString(tree, "iconUrl", violations);
-        JsonValidationUtils.validateOptionalString(tree, "documentationUrl", violations);
+        JsonValidationUtils.validateOptionalUrl(tree, "iconUrl", violations);
```
paoloantinori: Switching `iconUrl` and `documentationUrl` from `validateOptionalString` to `validateOptionalUrl` now rejects any non-http(s) value. Worth confirming the agent card spec requires http(s) here: if `iconUrl` is permitted to be a `data:` URI (a common pattern for inline icons), previously accepted cards will now fail FULL validation. If non-http(s) is intentionally out of scope, a one-line note in the PR description would make that policy explicit.

schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java
```diff
@@ -26,6 +28,54 @@ public static void validateOptionalString(JsonNode tree, String fieldName,
...
+     * normalized to lowercase before comparison. Additionally, {@link URI#getHost()} returns
+     * {@code null} for IPv6 literal hosts (e.g. {@code http://[::1]:8080/}) on some JDK
+     * versions; the authority component is used as a fallback for the presence check.
+     */
+    public static void validateHttpUrl(String value, String path, Set<RuleViolation> violations) {
+        try {
+            URI uri = new URI(value);
+            String scheme = uri.getScheme();
+            // Normalize to lowercase per RFC 3986 §3.1 — schemes are case-insensitive
+            if (scheme == null || !scheme.toLowerCase(java.util.Locale.ROOT).matches("https?")) {
```
paoloantinori: The scheme comparison can be written as `scheme == null || (!"http".equalsIgnoreCase(scheme) && !"https".equalsIgnoreCase(scheme))`. That removes the `matches("https?")` regex (recompiled into a Pattern on every call), removes the `toLowerCase(java.util.Locale.ROOT)` call, and removes the need for a `java.util.Locale` import, all in one change.

paoloantinori (commented):
Solid change that closes a real gap: the four URL fields were only type-checked, so values like `"not-a-url"` passed FULL validation and failed later in the A2A runtime with confusing errors. The fix reuses the existing `JsonValidationUtils` pattern, adds no dependencies, and covers both negative and positive cases (uppercase scheme, IPv6). The host-validation block needs tightening before merge, the optional-URL restriction deserves a spec check, and two smaller items round it out.

- The host-presence check in `validateHttpUrl` (`schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java:51`) is over-broad and incomplete. Its `getAuthority()` fallback lets URLs with no real hostname pass (verified against the OpenJDK parser: `http://host_` parses with `getHost()` null and `getAuthority()` `"host_"`), and the fallback's stated IPv6 purpose does not need it, since `new URI("http://[::1]:8080/").getHost()` returns `"[::1]"` (non-null), confirmed against OpenJDK source, not null. The method also does no port-range check, so `http://example.com:99999` passes and fails later at the socket layer, the exact runtime failure issue #8731 targets.
- Restricting `iconUrl` and `documentationUrl` to http/https (`schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java:138`) is a behavior change for optional fields. If the spec permits `data:` URIs for inline icons, previously accepted cards now fail FULL validation.
- The PR description lists three new fixtures/tests, but the diff adds five (the `uppercase-scheme` and `ipv6` cases are undocumented). "Tests run: 22" reconciles with five added, not three.
- NIT: the scheme comparison (`JsonValidationUtils.java:44`) can use `equalsIgnoreCase`, which removes the per-call regex compile, the `toLowerCase`, and the inline `java.util.Locale` reference in one change.

DSingh0304:
> Solid change that closes a real gap: the four URL fields were only type-checked, so values like `"not-a-url"` passed FULL validation and failed later in the A2A runtime with confusing errors. The fix reuses the existing `JsonValidationUtils` pattern, adds no dependencies, and covers both negative and positive cases (uppercase scheme, IPv6). The host-validation block needs tightening before merge, the optional-URL restriction deserves a spec check, and two smaller items round it out.
> 
>     * The host-presence check in `validateHttpUrl` (`schema-util/common/src/main/java/io/apicurio/registry/rules/validity/JsonValidationUtils.java:51`) is over-broad and incomplete. Its `getAuthority()` fallback lets URLs with no real hostname pass (verified against the OpenJDK parser: `http://host_` parses with `getHost()` null and `getAuthority()` `"host_"`), and the fallback's stated IPv6 purpose does not need it, since `new URI("http://[::1]:8080/").getHost()` returns `"[::1]"` (non-null), confirmed against OpenJDK source, not null. The method also does no port-range check, so `http://example.com:99999` passes and fails later at the socket layer, the exact runtime failure issue #8731 targets.
> 
>     * Restricting `iconUrl` and `documentationUrl` to http/https (`schema-util/common/src/main/java/io/apicurio/registry/rules/validity/AgentCardContentValidator.java:138`) is a behavior change for optional fields. If the spec permits `data:` URIs for inline icons, previously accepted cards now fail FULL validation.
> 
>     * The PR description lists three new fixtures/tests, but the diff adds five (the `uppercase-scheme` and `ipv6` cases are undocumented). "Tests run: 22" reconciles with five added, not three.
> 
>     * NIT: the scheme comparison (`JsonValidationUtils.java:44`) can use `equalsIgnoreCase`, which removes the per-call regex compile, the `toLowerCase`, and the inline `java.util.Locale` reference in one change.

Thanks for the thorough review @paoloantinori!

I've addressed all four points:

: replaced `toLowerCase + matches("https?") with "http".equalsIgnoreCase(scheme) && "https".equalsIgnoreCase(scheme)`  removes the per-call regex compile, the toLowerCase allocation, and the java.util.Locale import.

**Scheme check**: dropped the `getAuthority() fallback (confirmed getHost()` is non-null for IPv6 literals on standard JDKs, so the fallback was both unnecessary and over-broad for hostnames like `http://host_)`. Added a port-range check `(0–65535)` to catch the socket-layer failure case.

**iconUrl / documentationUrl**: reverted to . These fields are validated as strings, not as http/https URLs, to avoid breaking cards with data: URIs or other non-http schemes that the A2A spec doesn't restrict.
Test fixture: updated  to use non-string values `agentcard-invalid-optional-urls.json` so the test still asserts the string-type check rather than a URL scheme check.

The PR is ready for review @carlesarnal @EricWittmann ..

paoloantinori:
@DSingh0304 The feedback loop on the URL validation is closed from my side — the case-insensitive scheme handling and the IPv6 host fix both landed cleanly, and the new tests pin them down. No further comments from me. This one is now waiting on a maintainer `/accept` and green CI before merge.

DSingh0304:
> @DSingh0304 The feedback loop on the URL validation is closed from my side — the case-insensitive scheme handling and the IPv6 host fix both landed cleanly, and the new tests pin them down. No further comments from me. This one is now waiting on a maintainer `/accept` and green CI before merge.

Thanks @paoloantinori !! I will wait for @carlesarnal @EricWittmann to /accept the PR then..

paoloantinori:
Nothing further needed from me on the code side — the URL validation changes held up across the re-reads. This one is waiting on maintainer `/accept`, as you noted. Pinging `@carlesarnal` `@EricWittmann` is the right next step.

DSingh0304:
> Nothing further needed from me on the code side — the URL validation changes held up across the re-reads. This one is waiting on maintainer `/accept`, as you noted. Pinging `@carlesarnal` `@EricWittmann` is the right next step.

Sure then I will wait for @carlesarnal @EricWittmann to /accept the PR!!

EricWittmann (commented):
Thanks for the contribution @DSingh0304! The overall approach is solid — using `java.net.URI` with scheme/host checks is the right call, and the integration into the existing validation pattern is clean. I have a few items to address before we can merge.

## Must Fix

### 1. Remove dead code: `validateOptionalUrl()` is never called

The `validateOptionalUrl()` method added to `JsonValidationUtils` has zero callers in this PR or the rest of the codebase. The comment in `validateStringFields()` correctly explains that `iconUrl` and `documentationUrl` should not be restricted to HTTP(S), so this method isn't needed. Please remove it.

### 2. Contradictory IPv6 comments

The code comment in `validateHttpUrl()` says:
> URI.getHost() returns a non-null value for IPv6 literals (e.g. http://[::1]:8080/) on all standard JDKs

But the test comment in `testAgentCardIpv6Url()` says:
> URI.getHost() returns null for IPv6 literals; valid IPv6 URLs must not be rejected

These contradict each other. I verified: `URI.getHost()` returns `[::1]` (non-null) for `http://[::1]:8080/...` on current JDKs, so the **method comment is correct** and the **test comment is wrong**. Please fix the test comment.

### 3. Remove double blank line

There are two consecutive blank lines between `validateHttpUrl()` and `validateOptionalUrl()` in `JsonValidationUtils.java`. Please reduce to one (or zero once `validateOptionalUrl` is removed per item 1).

## Should Fix

### 4. Simplify redundant `port < 0` check

```java
if (port != -1 && (port < 0 || port > 65535)) {
```

`URI.getPort()` returns either -1 (no port) or a non-negative integer. After the `port != -1` guard, `port < 0` can never be true. Simplify to:

```java
if (port != -1 && port > 65535) {
```

Note: the port range check itself IS necessary (I verified that `java.net.URI` happily accepts ports like 99999), just the `< 0` branch is dead.

### 5. `host.trim().isEmpty()` — drop the `trim()`

```java
if (host == null || host.trim().isEmpty()) {
```

If the host contained whitespace, `new URI(value)` would have already thrown `URISyntaxException`. The `trim()` suggests whitespace is possible when it isn't. Just use `host.isEmpty()`.

### 6. Rename misleading test

`testAgentCardInvalidOptionalUrls` tests non-string types (`12345`, `false`) — it's testing the existing `validateOptionalString` behavior, not URL format validation. The in-test comment explains this well, but the test name is misleading. Consider `testAgentCardNonStringOptionalUrlFields` or similar.

### 7. Update PR description

The PR body says "all four URL fields" get HTTP(S) validation, but the implementation (correctly) only validates `provider.url` and `supportedInterfaces[*].url`. The code comment explains the design decision well — please update the PR description to match.

DSingh0304:
> Thanks for the contribution @DSingh0304! The overall approach is solid — using `java.net.URI` with scheme/host checks is the right call, and the integration into the existing validation pattern is clean. I have a few items to address before we can merge.
> ## Must Fix
> ### 1. Remove dead code: `validateOptionalUrl()` is never called
> 
> The `validateOptionalUrl()` method added to `JsonValidationUtils` has zero callers in this PR or the rest of the codebase. The comment in `validateStringFields()` correctly explains that `iconUrl` and `documentationUrl` should not be restricted to HTTP(S), so this method isn't needed. Please remove it.
> ### 2. Contradictory IPv6 comments
> 
> The code comment in `validateHttpUrl()` says:
> 
> > URI.getHost() returns a non-null value for IPv6 literals (e.g. http://[::1]:8080/) on all standard JDKs
> 
> But the test comment in `testAgentCardIpv6Url()` says:
> 
> > URI.getHost() returns null for IPv6 literals; valid IPv6 URLs must not be rejected
> 
> These contradict each other. I verified: `URI.getHost()` returns `[::1]` (non-null) for `http://[::1]:8080/...` on current JDKs, so the **method comment is correct** and the **test comment is wrong**. Please fix the test comment.
> ### 3. Remove double blank line
> 
> There are two consecutive blank lines between `validateHttpUrl()` and `validateOptionalUrl()` in `JsonValidationUtils.java`. Please reduce to one (or zero once `validateOptionalUrl` is removed per item 1).
> ## Should Fix
> ### 4. Simplify redundant `port < 0` check
> 
> ```java
> if (port != -1 && (port < 0 || port > 65535)) {
> ```
> 
> `URI.getPort()` returns either -1 (no port) or a non-negative integer. After the `port != -1` guard, `port < 0` can never be true. Simplify to:
> 
> ```java
> if (port != -1 && port > 65535) {
> ```
> 
> Note: the port range check itself IS necessary (I verified that `java.net.URI` happily accepts ports like 99999), just the `< 0` branch is dead.
> ### 5. `host.trim().isEmpty()` — drop the `trim()`
> 
> ```java
> if (host == null || host.trim().isEmpty()) {
> ```
> 
> If the host contained whitespace, `new URI(value)` would have already thrown `URISyntaxException`. The `trim()` suggests whitespace is possible when it isn't. Just use `host.isEmpty()`.
> ### 6. Rename misleading test
> 
> `testAgentCardInvalidOptionalUrls` tests non-string types (`12345`, `false`) — it's testing the existing `validateOptionalString` behavior, not URL format validation. The in-test comment explains this well, but the test name is misleading. Consider `testAgentCardNonStringOptionalUrlFields` or similar.
> ### 7. Update PR description
> 
> The PR body says "all four URL fields" get HTTP(S) validation, but the implementation (correctly) only validates `provider.url` and `supportedInterfaces[*].url`. The code comment explains the design decision well — please update the PR description to match.

Thanks for the detailed review @EricWittmann ! All items addressed in the latest push:

- **Dead code removed** - `validateOptionalUrl()` is gone; it had zero callers after `iconUrl`/`documentationUrl` were reverted to string-only validation.
- **IPv6 test comment fixed** - corrected to say `getHost()` returns a **non-null** value (e.g. `[::1]`) for IPv6 literals on standard JDKs.
- **Double blank line** - reduced to one.
- **Redundant `port < 0` removed** - simplified to `port != -1 && port > 65535`.
- **`host.trim()` dropped** - changed to `host.isEmpty()` since `new URI(...)` already throws on whitespace in the host.
- **Test renamed** - `testAgentCardInvalidOptionalUrls` → `testAgentCardNonStringOptionalUrlFields` to accurately reflect that it tests non-string type rejection, not URL format.
- **PR description updated** - removed "all four URL fields" claim; now correctly states only `provider.url` and `supportedInterfaces[*].url` get HTTP(S) validation, with an explanation for why `iconUrl`/`documentationUrl` are intentionally left as string-only.

Everything is squashed into a single signed commit, rebased on the latest `upstream/main`. Ready for another look!

DSingh0304:
Hi @EricWittmann @paoloantinori, All the checks pass but the "License Compliance" check is failing with 4 issues, but this PR introduces no new dependencies only `java.base` APIs (`java.net.URI`) and JSON test fixture files. The existing JSON fixtures in the same directory (`agentcard-full.json`, `agentcard-valid.json`, etc.) have no license headers and pass FOSSA fine.

Could you check if FOSSA needs an exclusion added for the new fixture files, or if this is a FOSSA scan configuration issue?

paoloantinori:
You're right that this is a License Compliance configuration issue, not a code problem. The PR only uses java.base APIs and JSON test fixtures, and the pre-existing fixtures in the same directory (agentcard-full.json, agentcard-valid.json) have no license headers either, so the four new fixture files should be treated the same way.

The most likely cause is that the FOSSA/license scan covers *.json resources but only the fixtures that existed when the exclusion list was last updated are whitelisted, so any new JSON file in that directory trips it. This is a repo-config matter for a maintainer (the scan exclusion list lives outside this PR), so I'll leave it to @EricWittmann or @carlesarnal to either extend the exclusion for the fixture directory or confirm whether license headers are expected on JSON resources. From the code side I still have nothing further to add; the URL validation changes held up across the re-reads.

EricWittmann:
Don't worry about the license check for now.  I honestly don't know what it's doing, so I'm ignoring it until I learn.  :)

EricWittmann (commented):
Thanks for the contribution @DSingh0304! All review feedback has been addressed cleanly — merging this now.

PR #9348, opened 2026-08-07
fix(rest): return 400 for null contract status

### LINKED ISSUE #9329: POST /contract/status returns HTTP 500 instead of 400 when required status is missing or null
## Description

While testing the contract status transition API, I noticed that the endpoint returns an HTTP 500 when the required `status` field is omitted or set to `null`.

Instead of returning a client error for invalid input, the request ends up throwing a `NullPointerException`.

## Steps to Reproduce

1. Create any artifact.
2. Send the following request:

```http
POST /apis/registry/v3/groups/default/artifacts/{artifactId}/contract/status

{}
```

or

```json
{
  "status": null
}
```

## Actual Result

The server responds with:

```
HTTP 500 Internal Server Error
```

caused by a `NullPointerException`.

## Expected Result

The request should return:

```
HTTP 400 Bad Request
```

since `status` is a required field.

## Possible Cause

It looks like `transitionContractStatus()` calls `data.getStatus().value()` before checking whether `data.getStatus()` is null. The method already validates the required path parameters using `ParameterValidationUtils.requireParameter(...)`, so it seems the request body is missing a similar validation.

If this is the expected direction, I'd be happy to submit a PR with a fix and regression tests.

Hiten0305l:
## What

Added validation for the required `status` field in the contract status transition endpoint.
The validation is performed in `transitionContractStatus()` before `data.getStatus().value()` is accessed. This ensures that requests with a missing or explicitly null `status` are rejected as invalid input rather than resulting in a `NullPointerException`.
The implementation follows the existing `ParameterValidationUtils.requireParameter(...)` validation pattern used for other required parameters.

## Why

Previously, requests containing either `{}` or `{"status": null}` caused `data.getStatus()` to return `null`. The subsequent call to `.value()` resulted in a `NullPointerException`, causing the endpoint to return HTTP 500.
Since `status` is a required request field, these requests should be rejected with HTTP 400 Bad Request instead.
This change validates `status` before it is accessed, preventing the exception and ensuring consistent request validation behavior.

## Tests

Added regression tests covering both cases:
- Missing `status`: `{}`
- Explicitly null `status`: `{"status": null}`
Both tests verify that the endpoint returns HTTP 400 Bad Request.
The tests pass successfully, and `git diff --check` reports no formatting issues.

## Issue

Fixes #9329

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
@@ -2142,6 +2142,7 @@ public class GroupsResourceImpl extends AbstractResourceImpl implements GroupsRe
 
         ParameterValidationUtils.requireParameter("groupId", groupId);
         ParameterValidationUtils.requireParameter("artifactId", artifactId);
+        ParameterValidationUtils.requireParameter("status", data.getStatus());
 
         String rawGroupId = new GroupId(groupId).getRawGroupIdWithNull();
         ContractStatus targetStatus = ContractStatus.valueOf(data.getStatus().value());
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
@@ -560,6 +560,32 @@ public class DataContractsResourceTest extends AbstractResourceTestBase {
     // -- Status Transition Tests --
 
     @Test
+    public void testStatusTransition_MissingStatus_Returns400() throws Exception {
+        String artifactId = "testStatusTransition_MissingStatus-" + UUID.randomUUID();
+        String content = resourceToString("openapi-empty.json");
+        createArtifact(GROUP, artifactId, ArtifactType.OPENAPI, content, ContentTypes.APPLICATION_JSON);
+
+        given().when().contentType(CT_JSON)
+                .pathParam("groupId", GROUP).pathParam("artifactId", artifactId)
+                .body("{}") // missing status
+                .post("/registry/v3/groups/{groupId}/artifacts/{artifactId}/contract/status")
+                .then().statusCode(400);
+    }
+
+    @Test
+    public void testStatusTransition_NullStatus_Returns400() throws Exception {
+        String artifactId = "testStatusTransition_NullStatus-" + UUID.randomUUID();
+        String content = resourceToString("openapi-empty.json");
+        createArtifact(GROUP, artifactId, ArtifactType.OPENAPI, content, ContentTypes.APPLICATION_JSON);
+
+        given().when().contentType(CT_JSON)
+                .pathParam("groupId", GROUP).pathParam("artifactId", artifactId)
+                .body("{\"status\": null}") // explicit null status
+                .post("/registry/v3/groups/{groupId}/artifacts/{artifactId}/contract/status")
+                .then().statusCode(400);
+    }
+
+    @Test
     public void testStatusTransition_DraftToStable() throws Exception {
         String artifactId = "testStatusTransition_DraftToStable-" + UUID.randomUUID();
         String content = resourceToString("openapi-empty.json");


app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
```diff
@@ -2142,6 +2142,7 @@ public ContractMetadata transitionContractStatus(String groupId, String artifact
 
         ParameterValidationUtils.requireParameter("groupId", groupId);
         ParameterValidationUtils.requireParameter("artifactId", artifactId);
+        ParameterValidationUtils.requireParameter("status", data.getStatus());
 
         String rawGroupId = new GroupId(groupId).getRawGroupIdWithNull();
```
paoloantinori, line 2147: The null guard added above fixes the reported NPE, but a non-null unknown value still reaches this line: `ContractStatus.valueOf("BOGUS")` throws `IllegalArgumentException`. Unless there's an exception mapper that turns IAE into 400 (worth confirming), a request like `{"status":"BOGUS"}` still returns HTTP 500, the same bad-input-to-500 class this PR is fixing. Consider validating the enum value (or catching the IAE) so an unknown status also returns 400, and add a regression test for it.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
```diff
@@ -559,6 +559,32 @@ public void testRulesetReplace() throws Exception {
...
+    public void testStatusTransition_MissingStatus_Returns400() throws Exception {
+        String artifactId = "testStatusTransition_MissingStatus-" + UUID.randomUUID();
+        String content = resourceToString("openapi-empty.json");
+        createArtifact(GROUP, artifactId, ArtifactType.OPENAPI, content, ContentTypes.APPLICATION_JSON);
+
+        given().when().contentType(CT_JSON)
+                .pathParam("groupId", GROUP).pathParam("artifactId", artifactId)
+                .body("{}") // missing status
+                .post("/registry/v3/groups/{groupId}/artifacts/{artifactId}/contract/status")
+                .then().statusCode(400);
```
paoloantinori, line 572: These two tests lock in the null/missing cases well. To close the remaining gap, add a case with a non-null but invalid status (e.g. `{"status":"BOGUS"}`) asserting 400; that path currently hits `ContractStatus.valueOf` and 500s, so the test would fail until value-validation is added. Asserting the error response body/code, not just `statusCode(400)`, would also make the regression intent explicit.

paoloantinori (commented):
Overall: the change correctly fixes the reported NPE by validating `status` before `data.getStatus().value()` is dereferenced (line 2147), reuses the existing `ParameterValidationUtils.requireParameter` pattern consistently with the `groupId`/`artifactId` checks, and adds focused regression tests for the `{}` and `{"status": null}` cases. The one substantive concern is that a non-null but unknown status value still reaches `ContractStatus.valueOf` and likely surfaces as HTTP 500 rather than 400, the same bad-input class this PR is closing.

- `GroupsResourceImpl.java:2144`: guard placement is correct (before the `.value()` deref at 2147) and the `requireParameter` reuse is the right idiom; this fully resolves the null/missing cases described in #9329 without altering any success path.
- `GroupsResourceImpl.java:2147`: `ContractStatus.valueOf(data.getStatus().value())` still throws `IllegalArgumentException` for a non-null unknown value such as `{"status":"BOGUS"}`. Unless an exception mapper converts IAE to 400 (worth confirming), this returns 500, the same class of bug. Consider validating the enum value or catching the IAE so unknown statuses also return 400.
- `DataContractsResourceTest.java:572,585`: both tests assert only `statusCode(400)`. Asserting the error body/code and adding an invalid-value case would lock in the value-validation path and make the regression intent explicit.

Hiten0305l:
Thanks, @paoloantinori! I really appreciate the review and feedback.

Since the current PR is now passing all checks, could you please confirm whether the suggested improvements should be addressed in this PR, or if it would be okay to handle them in a follow-up PR? If needed, I can add the additional error-response assertions and invalid-value test cases in DataContractsResourceTest.java to strengthen the validation coverage and make the regression behavior more explicit.

paoloantinori:
Thanks @Hiten0305l. Happy to leave the additional error-response assertions and the invalid-value test cases to a follow-up PR. The current change correctly fixes the reported NPE and the two regression tests pin the 400 behavior for both `{}` and `{"status": null}`, which is the scope of this PR. A separate follow-up with the broader assertion coverage would slot in cleanly without risking this fix.

Hiten0305l:
Thanks, @paoloantinori! I really appreciate the review and feedback.
Understood — I'll leave the additional error-response assertions and invalid-value test cases for a follow-up PR.
Since the current PR is now passing all checks and the reported issue is addressed, would it be possible to get this merged when you have a moment? Thanks!

EricWittmann:
Thanks for the contribution, @Hiten0305l! Clean fix with good regression tests. Merged!

PR #9349, opened 2026-08-07
feat(build): add root Makefile with developer shortcuts (#9318)

### LINKED ISSUE #9318: Feature: Introduce Makefile to standardize local development workflows

Local development and testing in Apicurio Registry currently require executing long, complex Maven commands with specific flags depending on the desired outcome. 

While the documentation in `DEVELOPING.md` is thorough, it can be difficult for new contributors to remember the exact syntax for building specific tiers, running dependency checks, or setting up integration environments.

Current pain points include:
* **Steep Learning Curve:** New contributors have to parse long commands like `./mvnw -T1 install -pl app -am -DskipTests -Dfull -DcliSkipNative -q` just to run a dependency check.
* **Manual Setup overhead:** Running tests against Postgres requires manually compiling with `-Pprod -Ddocker` and manually writing a custom `test.yml` for `docker-compose`.
* **Fragmented Workflows:** The repository is polyglot (Java, Python, Go, TypeScript), but commands to build or test these different components are scattered across different READMEs and shell scripts.

### Proposed Solution

Introduce a root `Makefile` to act as a unified entry point for all common developer tasks. This will wrap the complex Maven and Docker commands into easy-to-use, memorable shortcuts.

Proposed initial targets:
* `make build-local` — Compiles the core server and Java SDK quickly (`./mvnw clean install -Dlocal -DskipTests`)
* `make build-full` — Builds everything including CLI and Go/Python SDKs (`./mvnw clean install -Dfull -DskipTests`)
* `make test-integration` — Runs the default integration tests (`./mvnw verify -Pintegration-tests -Plocal-tests -pl integration-tests -am`)
* `make run-postgres` — Automates the creation of the temporary docker-compose file and spins up a local Postgres-backed instance of the registry.
* `make check-deps` — Runs the 2-step dependency check workflow.

This will drastically lower the barrier to entry for new contributors, reduce the time spent reading documentation, and provide "executable documentation" where anyone can see available tasks just by running `make` or glancing at the file.

### Additional Context

* This requires zero changes to core runtime code or existing CI workflows.
* Once established, it can easily be expanded in the future to include linting/formatting commands for the UI and SDKs across multiple languages.

If this proposal is accepted, I would love to work on it and submit the PR!

10vulture1005:
## Description

**Resolves #9318**

**What:**
Introduces a root `Makefile` to act as a unified entry point for all common developer tasks. Wraps complex Maven and Docker commands into simple, memorable targets to lower the barrier to entry for contributors. It also updates `DEVELOPING.md` to document these new targets.

**Why:**
Local development and testing currently require executing long, complex Maven commands with specific flags. This steep learning curve and manual setup overhead can be difficult for new contributors. This Makefile drastically lowers the barrier to entry, provides "executable documentation", and reduces time spent referencing documentation for command syntax.

**Included Targets:**
- `make build-local` — Compiles the core server and Java SDK.
- `make build-full` — Builds everything (skips native CLI build to avoid environment errors).
- `make test-integration` — Runs the default integration tests.
- `make check-deps` — Runs the 2-step dependency check workflow.
- `make run-postgres` — Automates the temporary docker-compose file creation and spins up Postgres.
- `make all` — Comprehensive chain of `build-full`, `check-deps`, and `test-integration`.
image

---
*(Note: I have verified that these changes require zero modification to core runtime code or existing CI workflows.)*


EricWittmann:
@jsenko wdyt?

Makefile
```diff
@@ -0,0 +1,43 @@
...
+	./mvnw clean install -Dlocal -DskipTests
+
+build-full:
+	./mvnw clean install -Dfull -DskipTests -DcliSkipNative
+
+test-integration:
+	./mvnw verify -Pintegration-tests -Plocal-tests -pl integration-tests -am
+
+run-postgres:
+	./mvnw clean install -DskipTests -Pprod -Ddocker
```
paoloantinori: `run-postgres` runs a full `./mvnw clean install -Pprod -Ddocker`, which builds the local prod image, but the compose `app` service at line 15 runs `apicurio/apicurio-registry:latest-release` pulled from Docker Hub. The locally built image is never used, so this build step is dead work that adds minutes to the target. Either drop the `install` and just start the release image, or point `app.image` at the tag produced by `-Ddocker` if the intent is to test the local build.

Makefile
```diff
@@ -0,0 +1,43 @@
...
+	echo "$$DOCKER_COMPOSE_YML" > test.yml
```
paoloantinori: `echo "$$DOCKER_COMPOSE_YML" > test.yml` writes a file into the repository root and never removes it. Unless `test.yml` is added to `.gitignore`, it will show up in `git status` and is trivially committable by accident, and the generic name can collide with other files. Write to a gitignored location (for example `.devenv/test.yml` or a `mktemp` path). Also add a companion `down`/clean target (`docker compose -f test.yml down -v`), since interrupting `up` leaves the containers and volumes behind.

Makefile
```diff
@@ -0,0 +1,43 @@
...
+	docker-compose -f test.yml up
```
paoloantinori: `docker-compose` (the v1 standalone Python binary) is end-of-life and not installed by default on current Docker Desktop or Linux setups that ship only the `docker compose` v2 plugin. This will fail on a fresh machine. Use `docker compose` (with the space).

DEVELOPING.md
```diff
@@ -73,6 +73,18 @@ cd app && ../mvnw quarkus:dev -Dlocal
...
 `-Pintegration-tests`, `-Pexamples`.
 
+## Using the Makefile (Shortcuts)
+
+For convenience, a `Makefile` is provided at the root of the repository to wrap common build and run commands. You can use these targets instead of typing out the full Maven commands:
+
+| Target | Description | Equivalent Command |
+|--------|-------------|--------------------|
+| `make build-local` | Compiles the core server and Java SDK quickly | `./mvnw clean install -Dlocal -DskipTests` |
+| `make build-full` | Builds everything including CLI and Go/Python SDKs | `./mvnw clean install -Dfull -DskipTests` |
```
paoloantinori: The Equivalent Command column lists `./mvnw clean install -Dfull -DskipTests`, but the actual target at Makefile:31 runs `... -DcliSkipNative`. The description 'Builds everything including CLI' also contradicts the native-CLI skip that the PR description calls out. Update both the description and the command to match the real target, otherwise contributors copying from the table run a heavier build than `make build-full` does.

paoloantinori (commented):
The Makefile is a welcome DX improvement and the build/test targets are solid, but `run-postgres` needs rework before it's useful: it writes an untracked `test.yml` into the repo root, leaves containers running with no teardown, and runs a full prod build that the compose file never actually uses. The `DEVELOPING.md` table also diverges from the real commands.

- **`run-postgres` leaks a file into the repo root** (`Makefile:38`): `echo ... > test.yml` writes `test.yml` at the project root and never removes it. Unless `test.yml` is gitignored, it surfaces in `git status` and is easy to commit by accident. Use a gitignored path or `mktemp`, and add a `down`/clean target.
- **The prod build is dead work** (`Makefile:37` vs `Makefile:15`): `run-postgres` builds with `-Pprod -Ddocker`, then the compose `app` service pulls `apicurio/apicurio-registry:latest-release` from Docker Hub. The locally built image is never referenced, so the install step only adds latency. Either drop it or point `app.image` at the locally built tag.
- **`docker-compose` (v1) is EOL** (`Makefile:39`): the standalone `docker-compose` binary is deprecated and absent on current Docker installs. Use the `docker compose` v2 plugin, and confirm this target works on a clean machine before documenting it.
- **Docs don't match the targets** (`DEVELOPING.md:83`): the table's `build-full` equivalent command omits `-DcliSkipNative`, and the description 'Builds everything including CLI' contradicts the native-CLI skip the PR description itself calls out. Copy the real command verbatim into the Equivalent column so `make build-full` and the documented command stay in sync.

DEVELOPING.md
10vulture1005: Thanks for the detailed feedback @paoloantinori! I've updated the PR to address all of your points:

Fixed file leaking & added clean target: Updated run-postgres to write the temporary docker compose configuration into the gitignored .work/test.yml directory instead of the project root. I've also added a clean-postgres target to run docker compose down -v and clean up the file afterwards.
Removed dead build work: Dropped the local ./mvnw clean install step from run-postgres entirely, since the compose file pulls latest-release from Docker Hub anyway.
Migrated to docker compose v2: Replaced all instances of the deprecated standalone docker-compose binary with the v2 plugin docker compose across both the Makefile and DEVELOPING.md.
Synced docs with actual targets: Updated the make build-full row in DEVELOPING.md so the description explicitly mentions it skips the native CLI, and the equivalent command now accurately includes -DcliSkipNative.
Let me know if anything else needs adjusting before merging!

PR #9351, opened 2026-08-07
fix(ui): reset Prompt Template test panel state on version change

### LINKED ISSUE #9194: [Bug]: Prompt Template Test panel keeps stale values/output after switching versions
## Description

**Persistence type**: in-memory / sql (UI bug; storage-independent)

When viewing a `PROMPT_TEMPLATE` artifact, the **Test Prompt** panel keeps stale state after switching versions. Rendered output, validation errors, and previously entered variable values from version A can still be visible (or reused) after navigating to version B.

Related but different: #8729 is about validation error field naming (`path` vs `variableName`). This issue is specifically about **not resetting panel state on version change**.

### Environment

Apicurio Registry UI (`ui/ui-app`), Prompt Template visualizer / Test Prompt panel.

### Steps to Reproduce

1. Open a `PROMPT_TEMPLATE` artifact that has at least two versions with different variables and/or template text.
2. On version A, fill some Test Prompt variable values and click **Render**.
3. Confirm rendered output (and any validation errors) appear.
4. Navigate to version B of the same artifact (version dropdown / version list).
5. Observe the Test Prompt panel.

**Expected:** The Test Prompt panel resets for the newly selected version — fresh default variable values for that version's schema, and no rendered output / validation errors / error banner left over from the previous version.

**Actual:** Prior values can remain for overlapping variable names, old variables may linger, and previous rendered output / validation errors stay on screen until the user manually renders again. The panel's `useEffect` only merges newly seen variable names into existing state and does not depend on `version`.

### Root cause (code)

`ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx`

- State (`values`, `renderedOutput`, `validationErrors`, `error`) is component-local.
- The variables `useEffect` depends only on `props.variables` and **merges** new keys into previous values instead of rebuilding.
- There is no reset when `props.groupId` / `props.artifactId` / `props.version` change, even though `PromptTemplateVisualizer` passes those props through.

### Logs

N/A — UI state bug; no backend error required.

Mohammad-Raees:
## Summary
- Fixes #9194: switching `PROMPT_TEMPLATE` versions no longer leaves stale Test Prompt state (variable values, rendered output, validation errors).
- Rebuilds panel defaults from the newly selected version's variables and clears prior render/error state when `groupId` / `artifactId` / `version` / `variables` change.
- Extracts `buildInitialValues` / `getVariablesList` helpers and adds unit coverage so leftover keys from a previous version cannot persist.

## Test plan
- [ ] Open a `PROMPT_TEMPLATE` artifact with at least two versions that have different variables/template text
- [ ] On version A, fill Test Prompt values and click **Render**; confirm output appears
- [ ] Switch to version B and confirm the panel resets (fresh defaults, no prior output/errors)
- [ ] Switch back to version A and confirm state does not carry over from B
- [ ] Run unit tests for `PromptTemplateTestPanel.utils`

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, ReactNode, useEffect, useMemo, useState } from "react";
+import { FunctionComponent, ReactNode, useMemo } from "react";
 import "./PromptTemplateTestPanel.css";
 import {
     ActionGroup,
@@ -25,9 +25,13 @@ import {
     ReconciledVariable,
     VariableSchema
 } from "./promptTemplateVariables";
-import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
-import { RenderPromptResponse, RenderPromptValidationError } from "@models/RenderPromptResponse.ts";
-import { coerceEnumValue } from "./PromptTemplateTestPanel.utils";
+import { useGroupsService } from "@services/useGroupsService.ts";
+import {
+    coerceEnumValue,
+    schemaForField,
+    toDeclaredMap
+} from "./PromptTemplateTestPanel.utils";
+import { usePromptTemplateTestPanelState } from "./usePromptTemplateTestPanelState";
 
 export type PromptTemplateTestPanelProps = {
     groupId: string;
@@ -36,33 +40,6 @@ export type PromptTemplateTestPanelProps = {
     template?: string;
     variables: Record<string, VariableSchema> | VariableSchema[] | undefined;
     className?: string;
-};
-
-const toDeclaredMap = (
-    variables: Record<string, VariableSchema> | VariableSchema[] | undefined
-): Record<string, VariableSchema> | undefined => {
-    if (!variables) {
-        return undefined;
-    }
-    if (Array.isArray(variables)) {
-        const map: Record<string, VariableSchema> = {};
-        for (const variable of variables) {
-            if (variable.name) {
-                map[variable.name] = variable;
-            }
-        }
-        return map;
-    }
-    return variables;
-};
-
-const defaultSchemaForDetected = (): VariableSchema => ({
-    type: "string",
-    required: false
-});
-
-const schemaForField = (entry: ReconciledVariable): VariableSchema => {
-    return entry.schema ?? defaultSchemaForDetected();
 };
 
 const sourceLabel = (source: ReconciledVariable["source"]): { text: string; color: "grey" | "orange" } | undefined => {
@@ -76,75 +53,29 @@ const sourceLabel = (source: ReconciledVariable["source"]): { text: string; colo
 };
 
 export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelProps> = (props: PromptTemplateTestPanelProps) => {
-    const groups: GroupsService = useGroupsService();
+    const groups = useGroupsService();
 
     const reconciledVariables = useMemo(() => {
         const names = extractTemplateVariableNames(props.template || "");
         return reconcileTemplateVariables(names, toDeclaredMap(props.variables));
     }, [props.template, props.variables]);
 
-    const initialValues: Record<string, any> = {};
-    reconciledVariables.forEach((entry) => {
-        const schema = schemaForField(entry);
-        if (schema.default !== undefined) {
-            initialValues[entry.name] = schema.default;
-        } else if ((schema.type || "string").toLowerCase() === "boolean") {
-            initialValues[entry.name] = false;
-        } else {
-            initialValues[entry.name] = "";
-        }
+    const {
+        values,
+        renderedOutput,
+        validationErrors,
+        isLoading,
+        error,
+        setValue,
+        doRender
+    } = usePromptTemplateTestPanelState({
+        groupId: props.groupId,
+        artifactId: props.artifactId,
+        version: props.version,
+        template: props.template,
+        variables: props.variables,
+        groups
     });
-
-    const [values, setValues] = useState<Record<string, any>>(initialValues);
-
-    useEffect(() => {
-        setValues(prev => {
-            const next = { ...prev };
-            reconciledVariables.forEach((entry) => {
-                if (!(entry.name in next)) {
-                    const schema = schemaForField(entry);
-                    const type = (schema.type || "string").toLowerCase();
-                    next[entry.name] = schema.default ?? (type === "boolean" ? false : "");
-                }
-            });
-            return next;
-        });
-    }, [props.template, props.variables]);
-
-    const [renderedOutput, setRenderedOutput] = useState<string>("");
-    const [validationErrors, setValidationErrors] = useState<RenderPromptValidationError[]>([]);
-    const [isLoading, setIsLoading] = useState(false);
-    const [error, setError] = useState<string>("");
-
-    const setValue = (name: string, value: any): void => {
-        setValues(prev => ({ ...prev, [name]: value }));
-    };
-
-    const doRender = (): void => {
-        setIsLoading(true);
-        setError("");
-        setValidationErrors([]);
-        setRenderedOutput("");
-
-        let gid: string | null = props.groupId;
-        if (gid === "default") {
-            gid = null;
-        }
-
-        groups.renderPromptTemplate(gid, props.artifactId, props.version, values)
-            .then((response: RenderPromptResponse) => {
-                setRenderedOutput(response.rendered || "");
-                if (response.validationErrors && response.validationErrors.length > 0) {
-                    setValidationErrors(response.validationErrors);
-                }
-            })
-            .catch((err: any) => {
-                setError(err?.message || "Error rendering prompt template");
-            })
-            .finally(() => {
-                setIsLoading(false);
-            });
-    };
 
     const renderField = (name: string, variable: VariableSchema): ReactNode => {
         const type = (variable.type || "string").toLowerCase();
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
@@ -1,5 +1,10 @@
 import { describe, expect, it } from "vitest";
-import { coerceEnumValue } from "./PromptTemplateTestPanel.utils";
+import {
+    buildInitialValues,
+    buildVersionIdentity,
+    coerceEnumValue,
+    shouldAcceptRenderResponse
+} from "./PromptTemplateTestPanel.utils";
 
 describe("coerceEnumValue", () => {
     it("parses an integer enum selection to a number", () => {
@@ -32,5 +37,77 @@ describe("coerceEnumValue", () => {
 
     it("keeps the placeholder selection as an empty string for a boolean variable", () => {
         expect(coerceEnumValue("", "boolean")).toBe("");
+    });
+});
+
+describe("buildInitialValues", () => {
+    it("uses declared defaults and boolean false when no default is set", () => {
+        expect(buildInitialValues("Hello {{city}} {{enabled}}", {
+            city: { type: "string", default: "Paris" },
+            enabled: { type: "boolean" }
+        })).toEqual({ city: "Paris", enabled: false });
+    });
+
+    it("includes template-detected names that are missing from the declared schema", () => {
+        expect(buildInitialValues("Hello {{city}} {{name}}", {
+            city: { type: "string", default: "Paris" }
+        })).toEqual({ city: "Paris", name: "" });
+    });
+
+    it("supports array-shaped variable declarations", () => {
+        expect(buildInitialValues(undefined, [
+            { name: "city", type: "string", default: "Berlin" },
+            { name: "enabled", type: "boolean", default: true }
+        ])).toEqual({ city: "Berlin", enabled: true });
+    });
+
+    it("returns an empty map when there is no template and no variables", () => {
+        expect(buildInitialValues(undefined, undefined)).toEqual({});
+    });
+
+    it("rebuilds defaults without leftover keys from a previous version's variables", () => {
+        const version1 = buildInitialValues("Hello {{city}} {{oldVar}}", {
+            city: { type: "string", default: "Paris" },
+            oldVar: { type: "string", default: "stale" }
+        });
+        expect(version1).toEqual({ city: "Paris", oldVar: "stale" });
+
+        const version2 = buildInitialValues("Hello {{city}}", {
+            city: { type: "string", default: "Berlin" }
+        });
+        expect(version2).toEqual({ city: "Berlin" });
+        expect(version2).not.toHaveProperty("oldVar");
+    });
+
+    it("preserves an explicit null default", () => {
+        expect(buildInitialValues("Hello {{city}}", {
+            city: { type: "string", default: null }
+        })).toEqual({ city: null });
+    });
+});
+
+describe("buildVersionIdentity", () => {
+    it("changes when the version changes", () => {
+        expect(buildVersionIdentity("default", "greeter", "1"))
+            .not.toBe(buildVersionIdentity("default", "greeter", "2"));
+    });
+
+    it("changes when the artifact identity changes", () => {
+        expect(buildVersionIdentity("default", "greeter", "1"))
+            .not.toBe(buildVersionIdentity("other", "greeter", "1"));
+    });
+});
+
+describe("shouldAcceptRenderResponse", () => {
+    it("accepts the response for the latest request on the current version", () => {
+        expect(shouldAcceptRenderResponse(3, 3, "g::a::1", "g::a::1")).toBe(true);
+    });
+
+    it("rejects a stale response after a newer Render starts", () => {
+        expect(shouldAcceptRenderResponse(2, 3, "g::a::1", "g::a::1")).toBe(false);
+    });
+
+    it("rejects a stale response after the version identity changes", () => {
+        expect(shouldAcceptRenderResponse(3, 3, "g::a::1", "g::a::2")).toBe(false);
     });
 });
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.ts
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.ts
@@ -1,3 +1,63 @@
+import {
+    extractTemplateVariableNames,
+    reconcileTemplateVariables,
+    ReconciledVariable,
+    VariableSchema
+} from "./promptTemplateVariables";
+
+export type PromptVariableLike = VariableSchema;
+
+export const toDeclaredMap = (
+    variables: Record<string, VariableSchema> | VariableSchema[] | undefined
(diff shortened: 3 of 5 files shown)


ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -31,47 +31,27 @@ export type PromptTemplateTestPanelProps = {
...
+    // Reset form values and render results whenever the viewed version identity changes.
+    // The previous effect only merged newly seen variable names into existing state, so
+    // output/errors/values from version A leaked into version B.
+    useEffect(() => {
+        setValues(buildInitialValues(props.variables));
+        setRenderedOutput("");
+        setValidationErrors([]);
+        setError("");
+        setIsLoading(false);
+    }, [props.groupId, props.artifactId, props.version, props.variables]);
```
paoloantinori: This effect now wipes values, output, errors, and loading whenever `props.variables` changes by reference. The previous effect only added missing keys, so an unstable parent reference was harmless. With a full reset, the same instability discards in-progress user input on every render. Confirm the parent passes a stable `variables` reference (the same instance across renders for a given version), or depend on a stable signature such as a version key or a memoized object rather than the raw object/array.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -31,47 +31,27 @@ export type PromptTemplateTestPanelProps = {
...
     const [validationErrors, setValidationErrors] = useState<RenderPromptValidationError[]>([]);
     const [isLoading, setIsLoading] = useState(false);
     const [error, setError] = useState<string>("");
 
+    // Reset form values and render results whenever the viewed version identity changes.
+    // The previous effect only merged newly seen variable names into existing state, so
+    // output/errors/values from version A leaked into version B.
+    useEffect(() => {
+        setValues(buildInitialValues(props.variables));
+        setRenderedOutput("");
```
paoloantinori: If a Render request is still pending when the version changes, this reset clears `renderedOutput`, but the unresolved promise can still resolve and call `setRenderedOutput` and `setIsLoading`, landing the previous version's output on the new panel. That is the same stale-leak class this PR targets. Consider guarding the async render path against stale resolutions, for example a version or request-token check that ignores a result once the version identity has moved on.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
```diff
@@ -1,5 +1,44 @@
...
+        const previousVersionValues = { oldVar: "stale", city: "London" };
+        const next = buildInitialValues({ city: { type: "string", default: "Paris" } });
+        expect(next).toEqual({ city: "Paris" });
+        expect(next).not.toHaveProperty("oldVar");
+        expect(previousVersionValues.oldVar).toBe("stale");
+    });
+});
 
 describe("coerceEnumValue", () => {
     it("parses an integer enum selection to a number", () => {
```
paoloantinori, line 10: The extracted helpers are covered, but the useEffect that performs the reset (the actual fix for #9194) is covered only by the manual test plan. A component-level test that renders the panel, fills values and triggers Render, then changes the version and asserts the values, output, and errors are cleared would guard this regression directly. The utils tests alone cannot catch a future regression in the reset logic.

paoloantinori (commented):
The fix correctly targets the stale-state leak: the old effect only merged newly seen variable keys into existing state, so values, output, and errors from version A bled into version B, and the helper extraction is clean and well covered. Three points to confirm before merge. The reset effect now wipes all state on any change to the `props.variables` reference, so an unstable parent reference becomes data loss where it was previously harmless. A render still in flight when the version switches can re-pollute the panel after the reset. And the useEffect that is the actual fix has no automated test, only the manual plan.

- `PromptTemplateTestPanel.tsx:53` - the effect fully resets state whenever `props.variables` changes by reference; the previous merge-only behavior tolerated unstable references, the new full reset discards in-progress input on every re-render if the parent recreates the object. Confirm reference stability or depend on a stable signature.
- `PromptTemplateTestPanel.tsx:49` - a pending Render request can resolve after the version change and call `setRenderedOutput`/`setIsLoading`, landing the old version's output on the new panel, the same stale-leak class this PR targets.
- `PromptTemplateTestPanel.utils.test.ts:44` - the extracted helpers are unit-tested, but the reset behavior that actually fixes #9194 is covered only manually; the utils tests cannot catch a future regression in the effect.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
Mohammad-Raees: Thanks @paoloantinori — agreed.

Fixed by keying the reset effect only on `groupId` / `artifactId` / `version`, and reading the latest `variables` through a ref. Unstable parent references for the same version no longer wipe in-progress input.

Covered by the "does not wipe in-progress input when only the variables reference changes" test in `usePromptTemplateTestPanelState.test.ts`.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
Mohammad-Raees: Good catch — that was still a stale-leak path.

Fixed with a request-token (`renderRequestIdRef`): it increments on version change and on each Render click, and the async `.then` / `.catch` / `.finally` handlers no-op when the token no longer matches.

Covered by the "ignores a stale Render response after the version changes" test.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
Mohammad-Raees: Agreed — utils tests alone weren't enough.

I extracted the reset/race logic into `usePromptTemplateTestPanelState` (what the panel uses) and added hook-level tests that:

1. fill values + Render, then change version and assert values/output/errors are cleared
2. keep input when only the `variables` reference changes
3. ignore a late Render resolution after version change
4. rebuild defaults from the new version's variables (no leftover keys)

Full PatternFly DOM mounting isn't workable in this Vitest setup (PatternFly CSS imports fail under jsdom), so the controller hook is the regression surface for the #9194 fix. Happy to adjust further if you'd prefer a different approach.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
Mohammad-Raees: Follow-up: also stabilized the parent reference in `PromptTemplateVisualizer` with `useMemo` over the `PromptTemplate` fields (including `variables`), so the same version keeps a stable `variables` instance across re-renders. The panel reset effect still depends only on the version key (`groupId::artifactId::version`), not `props.variables` by reference.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
Mohammad-Raees: Follow-up: the async Render path now guards with both a request token and an explicit version-identity check (`requestVersionIdentity !== versionIdentityRef.current`), so a late resolution is ignored once the viewed version has moved on.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
Mohammad-Raees: Follow-up: added `PromptTemplateTestPanel.test.tsx`, a component-level test that renders the panel, fills values, triggers Render, changes the version, and asserts values/output/errors are cleared — plus a stale-Render race case. PatternFly is stubbed only because its CSS modules cannot load under Vitest/jsdom; the panel component and its reset path are what the test exercises.

paoloantinori:
This is exactly the coverage I was hoping to see added. The new `PromptTemplateTestPanel.test.tsx` does the right thing: it renders the panel, fills values, triggers Render, asserts the output and error appear, then re-renders with a bumped version and confirms the rendered output, the error alert, and the prior value are all gone while the new version's default is applied. That exercises the behavior I flagged as untested (values/output/errors persisting across a version switch), not just the helper shape.

The stale-Render race case is a nice addition — it confirms that an in-flight `renderPromptTemplate` response that resolves after the version change does not write stale output back into the panel. The only thing worth keeping an eye on is that PatternFly is fully stubbed in the test, so this is validating state-management + effect ordering rather than the rendered DOM fidelity, but for the reset bug that's the layer that matters.

No further concerns from me. The fix and its tests look good.

Mohammad-Raees:
Thanks @paoloantinori — I really appreciate the thorough review and the clear feedback on the reset test and race guard. Glad the coverage looks good.

Happy to address anything if it comes up once full CI runs. Otherwise, I'm ready whenever a maintainer has a chance to `/accept` / merge. Thank you!

paoloantinori:
Sounds good @Mohammad-Raees. The coverage looks complete on this head. Nothing further from my side, ready for a maintainer to review when CI is green.

Mohammad-Raees:
Hi @EricWittmann — just rebased onto latest `main` to clear the merge conflict (kept the #8729 `variableName` validation display, and updated the tests to match). Branch is conflict-free again and local promptTemplate tests are green. Ready whenever you have a chance to `/accept`. Thanks!

Mohammad-Raees:
Hi @EricWittmann — hope you're doing well.

Just a gentle follow-up on **#9351** (fixes #9194). @paoloantinori has reviewed the latest changes and said he's good on his side; pre-accept checks (SonarCloud) are green, and the branch is conflict-free.

Whenever you have a moment, would you be able to **`/accept`** this PR so the full CI run can proceed? Happy to rebase onto latest `main` first if that would help.

Thanks again for your time and for reviewing contributions!

paoloantinori:
Nothing's changed on my end since my last read of this head (`8094ef1`) — the state-reset logic, the race guard, and the test coverage all still look good to me. The author has addressed the points I raised earlier. Standing by for @EricWittmann's review; no blockers from my side.

ui/ui-app/package.json
```diff
@@ -21,6 +21,8 @@
     "@apicurio/eslint-config": "1.0.0",
     "@kiota-community/kiota-wasm": "0.0.3",
     "@monaco-editor/react": "4.7.0",
+    "@testing-library/jest-dom": "7.0.0",
```
paoloantinori: `@testing-library/jest-dom` 7.0.0 declares `engines.node >= 22` (see package-lock.json), and `jsdom` 29 needs Node >= 20.19. If CI runs Node 20 (still a common LTS), these new devDependencies can break the test job: npm only warns on engine mismatch by default, but jest-dom 7 can still fail at runtime on Node < 22. Either confirm the CI Node baseline is >= 22 or pin jest-dom to a 6.x release.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts
```diff
@@ -1,5 +1,44 @@
...
+    });
+
+    it("defaults booleans to false and other types to empty string", () => {
+        expect(buildInitialValues({
+            enabled: { type: "boolean" },
+            name: { type: "string" }
+        })).toEqual({ enabled: false, name: "" });
+    });
+
+    it("rebuilds a clean map (no leftover keys from a previous version)", () => {
```
paoloantinori: This test reads as proof that leftover keys from a previous version are dropped, but `previousVersionValues` is a local that is never passed to `buildInitialValues` (which only receives the new variable definitions). `expect(next).not.toHaveProperty("oldVar")` and `expect(previousVersionValues.oldVar).toBe("stale")` are therefore tautologies: `buildInitialValues` always builds a fresh map, so it cannot leak by construction. The actual leak-prevention (replacing `values` wholesale on version change) is already covered by `usePromptTemplateTestPanelState.test.ts`. Either drop the dead local or rewrite this to assert the real contract.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.ts
```diff
@@ -1,3 +1,43 @@
...
+    variables: Record<string, PromptVariableLike> | PromptVariableLike[] | undefined
+): Record<string, any> => {
+    const values: Record<string, any> = {};
+    getVariablesList(variables).forEach(({ name, variable }) => {
+        if (!name) {
+            return;
+        }
+        const type = (variable.type || "string").toLowerCase();
+        if (variable.default !== undefined) {
+            values[name] = variable.default;
```
paoloantinori: Moving from `??` to `!== undefined` is a real behavior change on the version-change path: a variable with `default: null` now stores `null` instead of the boolean/empty-string fallback the previous reset effect produced via `??`. (Initial mount already stored null before this PR, so the delta is confined to the switch path.) If the registry can emit `default: null`, confirm `renderField` tolerates a null value for its input type, since a controlled input receiving null will warn. A `default: null` unit case would pin this down either way.

ui/ui-app/src/app/components/promptTemplate/usePromptTemplateTestPanelState.ts
```diff
@@ -0,0 +1,123 @@
...
+    const [values, setValues] = useState<Record<string, any>>(() => buildInitialValues(args.variables));
+    const [renderedOutput, setRenderedOutput] = useState<string>("");
+    const [validationErrors, setValidationErrors] = useState<RenderPromptValidationError[]>([]);
+    const [isLoading, setIsLoading] = useState(false);
+    const [error, setError] = useState<string>("");
+
+    // Reset form values and render results whenever the viewed version identity changes.
+    // Depend only on the version key — not variables by reference — so an unstable parent
+    // object cannot wipe in-progress user input on every re-render.
+    useEffect(() => {
```
paoloantinori, line 64: This reset effect plus the `renderRequestIdRef`/`versionIdentityRef` guards reproduce what a version-keyed remount of the panel gives for free: a fresh mount resets all state, and in React 18+ a Render promise that resolves after unmount is a silent no-op. If `PromptTemplateTestPanel` is re-keyed on the version identity (groupId/artifactId/version), the refs and the triple-guarded async handlers collapse back to the original straightforward component, and the stale-response test becomes unnecessary. The current approach is correct and covered, just heavier than the alternative; flagging as a simplification to consider rather than a defect.

paoloantinori (commented):
The fix correctly resets Test Prompt state on a version change, and the stale-Render guards in the new `usePromptTemplateTestPanelState` hook (request id plus version identity) are sound. I traced both the version-switch and the double-Render cases and each resolves to the latest request, so the race the PR targets is handled. The remaining items are a devDependency engine constraint to verify, one misleading unit test, a subtle null-default behavior change, and a possible simplification.

- `ui/ui-app/package.json:24`: `@testing-library/jest-dom` 7.0.0 requires Node >= 22 (per package-lock.json); confirm the CI Node baseline.
- `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.test.ts:34`: the "no leftover keys" test asserts tautologies; the real contract is covered in the hook test.
- `ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.utils.ts:33`: `??` changed to `!== undefined`, which alters null-default handling; confirm `renderField` tolerates null.
- `ui/ui-app/src/app/components/promptTemplate/usePromptTemplateTestPanelState.ts:54`: the reset effect and request/version guards could be replaced by a version-keyed remount of the panel.

Mohammad-Raees:
Thanks @paoloantinori for the thorough review — really appreciate you walking through the version-switch and stale-render scenarios.

I've pushed updates addressing the four items you flagged:

1. **`@testing-library/jest-dom` / Node baseline** — Pinned back to **6.6.3** so the UI test job stays compatible with CI's Node 20 baseline (`verify-build.yaml`).

2. **Tautological utils test** — Removed the `"no leftover keys"` case from `PromptTemplateTestPanel.utils.test.ts`. As you noted, `buildInitialValues` always returns a fresh map, so that test couldn't prove leak prevention; the real contract is already covered in `usePromptTemplateTestPanelState.test.ts`.

3. **`default: null` on the version-switch path** — Added a unit test that pins `buildInitialValues` storing `null` when `default: null` is declared (`!== undefined` semantics). `renderField` already coalesces null to safe controlled-input values (`?? ""` / `|| ""` on text inputs, `!!` on checkboxes), so the UI should tolerate this without warnings.

4. **Version-keyed remount simplification** — Acknowledged. The hook + ref guards are heavier than a `key={groupId/artifactId/version}` remount would be; I kept the current approach so the race guard is explicitly unit-tested, but I'm happy to simplify in a follow-up if you'd prefer that pattern here.

All prompt-template tests pass locally (34). Let me know if anything else needs adjusting — thanks again!

paoloantinori:
Thanks @Mohammad-Raees — all four items look addressed on this head:

1. `jest-dom` back at 6.6.3, confirmed in `package.json`.
2. Tautological test removed; the leak contract lives in the hook test.
3. `!== undefined` preserves `default: null`, and I traced `renderField` — `?? ""` / `|| ""` / `!!` coalesce null on every branch, so no controlled-input warning path.
4. Agree on keeping the hook for the unit-tested race guard; a follow-up simplification is fine if a maintainer prefers.

Nothing further from my side. Ready for a maintainer once CI is green.

EricWittmann:
@Mohammad-Raees You have some conflicts that need to be addressed.  :+1:

Mohammad-Raees:
Rebased onto latest `main` and resolved conflicts with the prompt-template variable auto-detection/reconciliation work.

- Kept the reconciled variable UI from `main`
- Restored version-identity reset + stale Render guards
- Fixed the prior `tsc` build failure (types now use `VariableSchema`)
- Locally verified: `npm run build`, `npm run lint`, and promptTemplate vitest suite

Mohammad-Raees:
Hi @EricWittmann — just a gentle bump on this one.

Conflicts with `main` are resolved (rebased onto the latest prompt-template reconciliation work), the prior UI `tsc` build failure is fixed, and the PR is mergeable again. Happy to address anything else if needed. Thanks!

Mohammad-Raees:
License Compliance is green again.

Root cause: the earlier FOSSA failure was not a real license problem in this PR — adding `jsdom` / Testing Library bloated `package-lock.json` and FOSSA’s scan then mis-resolved Quarkus `3.39.0.CR1` artifacts with empty license metadata (those versions are not in this repo).

Fix: dropped the new UI test npm deps / lockfile churn, kept the version-reset + stale-Render fix, and covered the reset/race contracts with node unit tests instead. Local `vitest` / `build` / `lint` are green.

Mohammad-Raees:
Hi @EricWittmann — whenever you have a moment, could you please merge this when convenient?

Review feedback is addressed, the Verification Gate / Extra Tests / Build (UI + Java) / Sonar checks are green, and the PR is up to date with `main` (conflict-free). The remaining red **License Compliance (FOSSA)** check looks like a known false positive from FOSSA Quick Import resolving versionless Quarkus deps to the unlicensed Maven Central “latest” (`3.39.0.CR1`); the same failure also appears on current `main`, and this PR only touches UI Prompt Template test-panel state (no dependency changes).

Thanks again!

EricWittmann:
Thanks for the clean fix, @Mohammad-Raees! The custom hook approach and stale-response guards are well done. Merged.

PR #9356, opened 2026-08-07
Refactor hardcoded A2A search filters and labels into constants

### LINKED ISSUE #9354: Refactor hardcoded A2A search filters and labels into constants (WellKnownResourceImpl)
### Summary
Currently, in `WellKnownResourceImpl.java`, several A2A-related labels and structure search prefixes (such as `"apicurio.agent.visibility"`, `"agent_card:skill:"`, etc.) are hardcoded directly as string literals. 

As the A2A discovery capabilities expand, centralizing these strings into a dedicated constants class (e.g., `io.apicurio.registry.a2a.A2AConstants` or similar) will improve maintainability and prevent typos in future search filter logic.

### Reason
- Improves code readability in `WellKnownResourceImpl`.
- Ensures consistency across the codebase if these labels/prefixes are needed elsewhere (e.g., during A2A Federation or in testing).

### Proposed Action
1. Extract the hardcoded visibility label: `"apicurio.agent.visibility"`
2. Extract the structure filter prefixes:
   - `"agent_card:skill:"`
   - `"agent_card:capability:"`
   - `"agent_card:inputmode:"`
   - `"agent_card:outputmode:"`
   - `"agent_card:protocolbinding:"`
3. Update `WellKnownResourceImpl` and any associated test classes to use the new constants.

I am looking to make my contribution to the project and would love to tackle this! If this approach sounds good, please assign this to me.

nitin-sharma-20:
Resolves #9354
### Summary
This PR extracts hardcoded A2A-related labels and structure search prefixes from `WellKnownResourceImpl.java` and centralizes them into a new `A2AConstants.java` utility class. 

### Reason
Currently, critical strings like `"apicurio.agent.visibility"` and `"agent_card:skill:"` are hardcoded directly within method logic. As the A2A discovery feature expands (especially with future federation enhancements), keeping these as inline strings violates DRY principles and risks silent runtime failures due to typos. Centralizing them ensures compile-time safety and easier maintainability.

### Changes Made
- **Created** `io.apicurio.registry.a2a.A2AConstants` to hold all A2A-specific labels and structural prefixes.
- **Refactored** `WellKnownResourceImpl.java` (specifically `getPublicAgents` and `searchAgentsAdvanced`) to use these new constants.
- **Updated** `WellKnownResourceTest.java` and `A2AAuthTest.java` to inject the constants via `String.format` inside the raw JSON request blocks.

### Verification
- [x] Code compiles successfully
- [x] Ran `mvn test -Dtest=WellKnownResourceTest,A2AAuthTest` in the `app` module; all 37 tests passed successfully.
- [x] Verified `A2AAuthTest` logic for visibility label assignments remains unaffected.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/a2a/A2AConstants.java
+++ b/app/src/main/java/io/apicurio/registry/a2a/A2AConstants.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2025 Red Hat
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.apicurio.registry.a2a;
 
 /**
--- a/app/src/main/java/io/apicurio/registry/a2a/A2AConstants.java
+++ b/app/src/main/java/io/apicurio/registry/a2a/A2AConstants.java
@@ -1,0 +1,20 @@
+package io.apicurio.registry.a2a;
+
+/**
+ * Constants related to A2A discovery and Agent Cards.
+ */
+public final class A2AConstants {
+
+    private A2AConstants() {
+        // utility class
+    }
+
+    public static final String LABEL_AGENT_VISIBILITY = "apicurio.agent.visibility";
+    public static final String VISIBILITY_PUBLIC = "public";
+
+    public static final String PREFIX_AGENT_CARD_SKILL = "agent_card:skill:";
+    public static final String PREFIX_AGENT_CARD_CAPABILITY = "agent_card:capability:";
+    public static final String PREFIX_AGENT_CARD_INPUT_MODE = "agent_card:inputmode:";
+    public static final String PREFIX_AGENT_CARD_OUTPUT_MODE = "agent_card:outputmode:";
+    public static final String PREFIX_AGENT_CARD_PROTOCOL_BINDING = "agent_card:protocolbinding:";
+}
--- a/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
@@ -72,7 +72,7 @@ import java.util.Set;
 @ApplicationScoped
 @Interceptors({ResponseErrorLivenessCheck.class, ResponseTimeoutReadinessCheck.class})
 @Logged
-public class WellKnownResourceImpl implements WellKnownResource {
+public class    WellKnownResourceImpl implements WellKnownResource {
 
     private static final Logger log = LoggerFactory.getLogger(WellKnownResourceImpl.class);
     private static final ObjectMapper mapper = new ObjectMapper();
@@ -346,7 +346,7 @@ public class WellKnownResourceImpl implements WellKnownResource {
         // Filter by skills (indexed as structured content: agent_card:skill:<id>)
         if (skills != null && !skills.isEmpty()) {
             for (String skill : skills) {
-                filters.add(SearchFilter.ofStructure("agent_card:skill:" + skill));
+                filters.add(SearchFilter.ofStructure(A2AConstants.PREFIX_AGENT_CARD_SKILL + skill));
             }
         }
 
@@ -357,7 +357,7 @@ public class WellKnownResourceImpl implements WellKnownResource {
                 String[] parts = capability.split(":", 2);
                 String capKey = parts[0];
                 String capValue = parts.length > 1 ? parts[1] : "true";
-                SearchFilter filter = SearchFilter.ofStructure("agent_card:capability:" + capKey);
+                SearchFilter filter = SearchFilter.ofStructure(A2AConstants.PREFIX_AGENT_CARD_CAPABILITY + capKey);
                 if ("false".equals(capValue)) {
                     filter = filter.negated();
                 }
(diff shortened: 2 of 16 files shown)


app/src/main/java/io/apicurio/registry/a2a/A2AConstants.java
```diff
@@ -0,0 +1,20 @@
+package io.apicurio.registry.a2a;
```
paoloantinori, line 17: This file is added without the Apache 2.0 license header that every other Java file in this module carries. Apicurio's checkstyle config enforces a header/license rule, so this fails checkstyle even though `mvn test` (which the PR verification ran) does not invoke checkstyle. Add the standard header used by the neighboring `A2AConfig.java` before merging, otherwise CI will flag it.

app/src/main/java/io/apicurio/registry/rest/wellknown/WellKnownResourceImpl.java
```diff
@@ -212,13 +213,13 @@ public AgentSearchResults searchAgentsAdvanced(AgentSearchRequest request) {
         if (f != null) {
             if (f.getSkills() != null) {
                 for (String skill : f.getSkills()) {
-                    filters.add(SearchFilter.ofStructure("agent_card:skill:" + skill));
+                    filters.add(SearchFilter.ofStructure(A2AConstants.PREFIX_AGENT_CARD_SKILL + skill));
```
paoloantinori, line 216: This migrates only the query side. A structure filter like `agent_card:skill:<x>` only matches if agent cards are indexed with that exact key, and `apicurio.agent.visibility` has to be written somewhere when a card is published. Those producers are not in this file or this PR. If the indexing/storage path and the visibility writer still hold their own string literals, the typo-divergence this PR exists to prevent is still live, and centralizing just the reader gives a false sense of safety. Confirm those sites (card indexing, wherever public visibility is stamped) and migrate them in the same change, or state explicitly in the PR description that they are out of scope.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
```diff
@@ -362,17 +363,17 @@ public void testSearchAgentsAdvancedWithFilters() throws Exception {
         createAgentCard(groupId, "filter-agent", AGENT_CARD_CONTENT);
         setVisibility(groupId, "filter-agent", "public");
 
-        String requestBody = """
+        String requestBody = String.format("""
```
paoloantinori: This is a contract test for the JSON external clients send, and routing the constant through `String.format` into a raw JSON block makes it worse, not better. Two concrete problems: (1) format-string fragility, any future constant containing `%` throws `IllegalFormatException` and there is no JSON escaping; (2) the test now shares the exact constant with production, so it can no longer independently catch a typo in the literal value. Keeping the literal `"apicurio.agent.visibility"` in the JSON (as it was) is clearer and a stronger regression guard. The production-side refactor stands on its own without forcing the constant into the test.

paoloantinori (commented):
Two issues stop this being a safe merge as-is: the new `A2AConstants.java` ships without the Apache license header the rest of the module enforces, and the extraction covers only the query side, so the producers of these same strings are still on literals and the typo-divergence the PR exists to prevent is still live. The extraction itself is behavior-preserving: every constant reproduces its original literal verbatim and no call site changes semantics.

- `A2AConstants.java` starts at `package` with no license header, unlike every other Java file in this module. Apicurio's checkstyle config enforces a header rule, so this fails checkstyle even though `mvn test` (what the author ran) does not run it. (A2AConstants.java:1)
- This only migrates the read/query path. A structure filter like `agent_card:skill:<x>` only returns hits if agent cards are indexed with that exact key, and `apicurio.agent.visibility` has to be written somewhere when a card goes public; neither producer is in this file or this PR. If those sites keep their own literals, the divergence this change exists to prevent is still live and centralizing just the reader gives a false sense of safety. (WellKnownResourceImpl.java:216)
- Switching the test body to `String.format` over a raw JSON string is a net negative for a contract test: format-string fragility (any future constant containing `%` throws `IllegalFormatException`, and there is no JSON escaping) plus coupling the assertion to the same constant as production, so it can no longer independently guard the literal clients send. (WellKnownResourceTest.java:366)

nitin-sharma-20:
Thanks for the thorough review @paoloantinori, I've pushed an update addressing all three:

**License Header:** Added the standard Apache 2.0 license header to A2AConstants.java. (I noticed A2AConfig.java is actually missing the header too, but I grabbed the standard Red Hat one from another file in the module!)

**Contract Tests:** Reverted the tests back to using the raw string literal "apicurio.agent.visibility" so they can catch regressions independently.

**The Producer Side:** I migrated a few occurrences I missed on the query side in WellKnownResourceImpl.java. However, for the producer side:
- agent_card:skill: does not exist as a static string literal. The indexing engine dynamically builds it by joining the artifact type and field ("agent_card" + ":" + "skill").
- apicurio.agent.visibility is not stamped automatically by the server; it's provided directly by clients as a label.
Because of this, there are no static string literals to migrate on the producer side! Let me know if everything looks good now.

paoloantinori:
Thanks @nitin-sharma-20. I checked the producer side and you're right, there's nothing to migrate.

The structural-search prefixes (`agent_card:skill:`, `agent_card:capability:`, etc.) are built at query time only: `WellKnownResourceImpl.searchAgentsAdvanced` concatenates them from the artifact type and the field name. On the indexing side the search-structure entries are emitted dynamically by the storage layer (artifact type + field), so no producer code carries the literal `agent_card:skill:` string.

Same for `apicurio.agent.visibility`: it's a client-supplied label, read back via `SearchFilter.ofLabel(...)` at query time. The server never stamps it, so there is no hardcoded producer literal to centralize.

The query-side extraction in this PR is the right scope. License header and the reverted contract tests both look good. Nothing further from my side.

nitin-sharma-20:
Thanks for confirming, @paoloantinori! Really appreciate the guidance and the thorough review 👍. Let me know when this is good to merge!

paoloantinori:
Glad it helped, @nitin-sharma-20. From my side the feedback loop is closed, the constants extraction looks clean and the license header is in place.

One minor nit I spotted in the diff: `WellKnownResourceImpl.java` line 75 picked up some extra whitespace during the edit, `public class    WellKnownResourceImpl` (four spaces between `class` and the name). Worth tidying before merge so it doesn't trip a formatter later.

The merge call itself is EricWittmann's or jsenko's, not mine.

nitin-sharma-20:
Hi @paoloantinori Good catch! I just pushed a commit to fix that extra whitespace.

paoloantinori:
Confirmed, the extra whitespace is gone. From my side this is ready.

nitin-sharma-20:
Hi @EricWittmann waiting for your approval on this. Thank you.

EricWittmann:
Thanks for the contribution, @nitin-sharma-20! Merged.

Issue #9042, opened 2026-07-30
UI: make the "Import from URL" 5MB content-length cap configurable


sarwar-jahin-port:
## Feature Description

The **Import from URL** feature currently hard-caps fetched content to **5 MB**:

```ts
options.maxContentLength = "5242880"; // TODO 5MB hard-coded, make this configurable?
```

This limit can be too restrictive for legitimate use cases such as large OpenAPI specs, Avro/JSON schema collections, or prompt template bundles. At the moment, increasing the limit requires patching and rebuilding the UI.

## Proposed Solution

I'd like to make this configurable, but I'm not sure which approach fits the project best.

### Option A (backend-driven)

Introduce a new `apicurio.ui.*` configuration property (defaulting to `5242880`) and expose it through `/system/uiConfig`, similar to existing UI configuration values. The frontend would read it and fall back to 5 MB if unset.

**Pros:** Consistent with the existing configuration model and configurable without rebuilding the UI.

### Option B (frontend-only)

Add a value to the UI config files (`config-*.js`) and have `useUrlService.ts` read it, defaulting to 5 MB.

**Pros:** Simpler implementation.

**Cons:** Only configurable at UI build/deploy time.

Either approach would preserve the current 5 MB default and include tests. If Option A is preferred, I'd also regenerate the configuration documentation.

Since this would be my first contribution, I'd appreciate guidance on which direction you would prefer before I start implementing it.

## Additional Context

Relevant TODO comment:

https://github.com/Apicurio/apicurio-registry/blob/main/ui/ui-app/src/services/useUrlService.ts#L23

Milton-19:
"Hi, this looks like a great issue to start on. @sarwar-jahin-port, are you still planning to work on this, or is it open for someone else to pick up? Happy to take it if it's available — I'd lean toward Option A (backend-driven) since it avoids needing a UI rebuild for config changes."

sarwar-jahin-port:
> "Hi, this looks like a great issue to start on. [@sarwar-jahin-port](https://github.com/sarwar-jahin-port), are you still planning to work on this, or is it open for someone else to pick up? Happy to take it if it's available — I'd lean toward Option A (backend-driven) since it avoids needing a UI rebuild for config changes."

Thanks for checking in! I haven't started working on it yet as I'm also waiting for confirmation from the maintainers.

Milton-19:
[...]

PR #9360, opened 2026-08-07
fix(ui): make Import from URL content-length cap configurable

Prathamesh09-coder:
## Summary

The "Import from URL" feature previously hard-coded a 5 MB (5242880 bytes) limit for maxContentLength. This cap is too restrictive for large OpenAPI specs, Avro/JSON Schema collections, or prompt template bundles, and required rebuilding the UI to change it.

## Changes

- Added optional `importMaxContentLength?: number` field to the `UiConfig` interface in `useConfigService.ts`
- Added `uiImportMaxContentLength(): number` accessor to `ConfigService` interface and `ConfigServiceImpl`, defaulting to `5242880` when not configured
- Updated `useUrlService.ts` to call `configService.uiImportMaxContentLength()` instead of the hardcoded literal, removing the TODO comment

**Backward compatible:** the default remains 5 MB, so existing deployments are unaffected.

## How to Configure

Add `importMaxContentLength` to the `ui` section of your config (e.g. `config-local.js` or via `/system/uiConfig`):

```js
// config-local.js
var ApicurioRegistryConfig = {
    artifacts: { url: "http://localhost:8080/apis/registry/v3" },
    ui: {
        importMaxContentLength: 10485760 // 10 MB
    }
};
```


Prathamesh09-coder:
Closes #9042

ui/ui-app/src/services/useConfigService.ts
```diff
@@ -309,6 +311,11 @@ export class ConfigServiceImpl implements ConfigService {
         return registryConfig.ui.navPrefixPath;
     }
 
+    public uiImportMaxContentLength(): number {
+        // Default to 5 MB (5242880 bytes) if not explicitly configured.
+        return registryConfig.ui?.importMaxContentLength ?? 5242880;
```
paoloantinori: The `?? 5242880` fallback only fires when the field is `null` or `undefined`, not when an operator sets it to `0` or a negative number. A value of `0` would silently break every Import-from-URL, because axios rejects any response that exceeds `maxContentLength`. Consider falling back to the `5242880` default (or clamping to a positive minimum) when the configured value is not a positive number, consistent with the fail-fast-on-invalid-config convention.

ui/ui-app/src/services/useUrlService.ts
```diff
@@ -39,6 +16,31 @@ export interface UrlService {
...
+            const branch: string = match[3];
+            const path: string = match[4];
+
+            url = `https://raw.githubusercontent.com/${org}/${repo}/${branch}/${path}`;
+        }
+
+        console.info("[UrlService] Fetching content from a URL: ", url);
+
+        const endpoint: string = url;
+        const options: any = createOptions({
```
paoloantinori, line 35: This changes `maxContentLength` from the string `'5242880'` to a number. axios expects `maxContentLength` as a number of bytes, so the old string worked only through implicit coercion; this is effectively a latent fix rather than a regression. Worth a one-line note in the PR description, and worth confirming the `createOptions`/`httpGet` path in rest.utils does not assume a string.

ui/ui-app/src/services/useUrlService.ts
```diff
@@ -39,6 +16,31 @@ export interface UrlService {
  * React hook to get the URL service.
```
paoloantinori, line 16: Calling `useConfigService()` here makes `useUrlService` transitively depend on React context. Previously it was a plain factory returning a module-level function. Confirm that every consumer calls `useUrlService()` during component render and never conditionally or inside a callback or effect, otherwise this breaks the rules of hooks.

paoloantinori (commented):
Overall this is a clean, backward-compatible change that moves the hard-coded 5 MB Import-from-URL cap behind a `ConfigService` accessor with a safe default, so existing deployments are unaffected and the magic number now lives in one place. There are no blockers; the points below are minor but worth addressing before merge.

- `useConfigService.ts:316`: the `?? 5242880` fallback only protects against the field being absent (`null`/`undefined`). It does not guard against `0` or a negative number, which an operator could set by mistake and which would silently break every Import-from-URL (axios rejects any response larger than `maxContentLength`). A bounds check that falls back to the default for non-positive values, in the spirit of the existing fail-fast-on-invalid-config pattern, would make this misconfiguration-proof.
- `useUrlService.ts:35`: assigning `configService.uiImportMaxContentLength()` changes `maxContentLength` from the previous string literal (`'5242880'`) to a number. axios defines `maxContentLength` as a number of bytes, so the old string only worked through implicit coercion and this is effectively a latent fix rather than a regression. Worth a one-line note in the PR description, plus a check that the `createOptions`/`httpGet` path in rest.utils does not assume a string.
- `useUrlService.ts:16`: calling `useConfigService()` makes `useUrlService` transitively depend on React context, where previously it was a plain factory returning a module-level function. Confirm every consumer calls `useUrlService()` during component render and never conditionally or inside a callback or effect, otherwise this breaks the rules of hooks.
- `useUrlService.ts:2`: the new import path `@services/useConfigService.ts` carries a `.ts` extension while the existing import on line 1 (`@utils/rest.utils`) does not. Confirm this matches the project convention so module resolution stays consistent.
- No unit test was added for the new accessor. A small test asserting the default (5242880) when unset and the configured value when set would lock in the contract.

Prathamesh09-coder:
Thank you @paoloantinori and @Copilot for the thorough review! I've pushed a follow-up commit (`14444f6`) that addresses all the points:

**1. Bounds check + type coercion (main concern)**
`uiImportMaxContentLength()` now uses `Number(configured)` to coerce any string value from JSON config, then validates with `Number.isFinite(value) && value > 0`, falling back to `5242880` for absent, null, undefined, zero, negative, NaN, or non-numeric-string values. This prevents silent axios breakage from operator misconfiguration.

**2. String -> number change note**
The original `"5242880"` string was a latent bug - `axios` defines `maxContentLength` as `number` (confirmed in `rest.utils.ts` line 361 which already uses `const fiftyMB: number = ...`). The new code correctly passes a `number`. No assumptions of a string type were found in the `createOptions`/`httpGet` path.

**3. React hooks rule**
All three consumers (`CreateArtifactModal`, `CreateDraftModal`, `ImportAgentModal`) call `useUrlService()` unconditionally at the top of their functional components during render - no conditional calls or usage inside effects/callbacks.

**4. Import path `.ts` extension**
Confirmed `.ts` extension is the project convention - all existing `@services/useConfigService.ts` imports across the codebase include the extension (e.g. `App.tsx`, `main.tsx`, `MainPageWithAuth.tsx`). No change needed.

**5. Unit tests**
Added `useConfigService.test.ts` (9 test cases) covering: default when absent, undefined, null, 0, negative, NaN, non-numeric string, valid positive number, and numeric string coercion - following the Vitest pattern established by `useAdminService.test.ts`.

Closes #9042

paoloantinori:
Re-read the follow-up commit. The bounds check and type coercion look right — `Number(configured)` with the `isFinite && > 0` guard covers the absent/null/zero/negative/NaN/non-numeric cases cleanly, and passing a `number` to `maxContentLength` matches what axios expects. The test coverage is thorough, and the hook placement is fine since `useConfigService()` is called unconditionally at the top of `useUrlService()`. This looks ready from my side — just needs a green CI run and a rebase onto main.

Nachiket-Roy:
> Closes #9042

Please add this in the description!

PR #9365, opened 2026-08-07
fix(rest): fail Protobuf export when a reference cannot be resolved

### LINKED ISSUE #9333: bug: Silent reference failure produces incomplete Protobuf export
## Description

**Registry Version**: snapshot (commit `947a69f0`)
**Persistence type**: Any (sql / kafkasql / in-memory)

`ProtobufExporter.collectReferences()` swallows all exceptions thrown by `storage.getContentByReference()`. When any reference fails to load, the loop silently skips that entry — the failed `.proto` file is never added to the `result` map — and the export continues. The caller `exportVersionAsZip` has no visibility into the omission and returns HTTP 200 with a zip that is missing one or more dependency files.

**Affected file:**
[`app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java`, line 157](https://github.com/Apicurio/apicurio-registry/blob/947a69f035ff557946dd053793d54733cecb9c0b/app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java#L157)

```java
        } catch (Exception e) {
            log.warn("Could not resolve reference: {} - {}", ref.getName(), e.getMessage());
        }
```

The `catch` block only calls `log.warn()` and does not rethrow or signal the failure in any way. Execution falls through to the next loop iteration and the export proceeds as if the reference never existed.

### Environment

Any environment running Apicurio Registry. Reproducible via static code inspection and confirmed by reading lines 130–161 of `ProtobufExporter.java` at the pinned commit.

### Steps to Reproduce

1. Register a Protobuf schema that has one or more cross-artifact references (e.g. schema `A` imports schema `B` stored as a separate artifact).
2. Arrange for the reference to be unresolvable at export time — for example, delete the referenced artifact version, corrupt its content record, or introduce a storage-layer error for that content ID.
3. Call the export endpoint for schema `A`:
   ```
   GET /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{versionExpression}/references/export
   ```
4. Observe the HTTP response.

**Expected:** HTTP 4xx/5xx error response indicating the export could not be completed.

**Actual:** HTTP 200 OK with a zip archive that silently omits the failed reference's `.proto` file. Attempting to compile the exported protos with `protoc` produces unresolved-import errors, but the API gave no indication of the partial failure.

**Expected:** If any reference fails to resolve during `collectReferences()`, the export should fail with an error response so the client knows the artifact set is incomplete.

**Actual:** `collectReferences()` catches the exception at line 157, logs a `WARN`, and continues the loop. The `result` map is never populated for the failed reference. `exportVersionAsZip` receives a partial map and writes an incomplete zip. The HTTP response is 200 OK with no error body, leaving the client with a corrupt export they have no way to detect via the API.

The loop structure (lines 137–160) makes the silent-drop behaviour clear:

```java
for (ArtifactReferenceDto ref : references) {
    String key = buildVisitedKey(ref.getGroupId(), ref.getArtifactId(), ref.getVersion());

    if (visited.contains(key)) {
        continue;
    }
    visited.add(key);

    try {
        ContentWrapperDto content = storage.getContentByReference(ref);
        if (content != null && content.getContent() != null) {
            String refContent = content.getContent().content();
            String importPath = ref.getName();
            ProtoContent protoContent = new ProtoContent(importPath, refContent);
            result.put(importPath, protoContent);          // never reached on exception

            collectReferences(content.getReferences(), result, visited);
        }
    } catch (Exception e) {
        log.warn("Could not resolve reference: {} - {}", ref.getName(), e.getMessage());
        // no rethrow — loop continues to next reference
    }
}
```

### Logs

With `LOG_LEVEL=DEBUG`, the only observable signal is a single `WARN` line per failed reference:

```
WARN  Could not resolve reference: <ref-name> - <exception message>
```

No error propagates to the HTTP layer; the response log will show `200 OK`.

---

deepak0x:
## Summary

`ProtobufExporter.collectReferences()` swallowed every exception thrown by `storage.getContentByReference()`. When a reference failed to resolve, the loop silently skipped it, the failed `.proto` was never added to the result map, and `exportVersionAsZip` returned **HTTP 200 with a partial ZIP** — leaving the client with an incomplete export it could not detect via the API.

This PR makes the failure propagate: the `catch` block now rethrows (wrapping the cause), so the export errors out. The caller `exportVersionAsZip` already wraps any exception from `collectAllProtoFiles` into an `IOException`, which surfaces as a `5xx` — matching the issue's expected behavior.

## Changes

`app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java`:
- `collectReferences`: the `catch (Exception e)` now rethrows (`RuntimeException` carrying the reference name and cause) instead of only logging. The existing `log.warn` is kept for observability.

`app/src/test/java/io/apicurio/registry/noprofile/rest/v3/ProtobufExportTest.java`:
- Added `testExportFailsWhenReferenceCannotBeResolved`: creates a protobuf artifact with a reference, deletes the referenced dependency (DeletionEnabledProfile), then asserts the export endpoint returns a non-2xx response instead of a silently-partial 200.

## Test plan

- `./mvnw -pl app -am test-compile -DskipTests` compiles cleanly.
- `ProtobufExportTest.testExportFailsWhenReferenceCannotBeResolved` covers the regression (export fails when a reference is unresolvable). Existing export tests (`testExportArtifactWithReferences`, transitive, etc.) remain valid since they exercise the success path.

Fixes #9333


paoloantinori (commented):
The change correctly makes reference-resolution failures propagate instead of returning a silent partial export, and the existing success-path tests are unaffected since the rethrow only fires on an exception. What keeps it from being verifiable as written: the regression test asserts only `>= 400`, which does not prove the intended 5xx propagation, and the rethrow uses a bare `RuntimeException` whose resulting status depends on the generic mapper rather than a deliberate mapping.

- `ProtobufExportTest.java:401`: `greaterThanOrEqualTo(400)` passes for any failure (a 404 from an unrelated lookup, a generic 500 from another code path), so it cannot confirm the `RuntimeException` from `collectReferences` actually reaches the intended 5xx through `exportVersionAsZip`. Pin the expected status (e.g. 500) and assert the error body if the endpoint returns one; that is the only way to prove the propagation claim in the summary.
- `ProtobufExporter.java:159`: a bare `RuntimeException` bypasses the registry exception hierarchy and its `ExceptionMapper`, so the HTTP status and error body are whatever the generic mapping produces rather than a deliberate mapping. Prefer a domain exception type that fits (a not-found or server-error subtype), or at minimum confirm what status and body the generic mapper actually returns for this path.
- `ProtobufExporter.java:157`: `catch (Exception e)` is broad, so the rethrow now makes every failure in `storage.getContentByReference()` fatal to the whole export, not only a genuinely missing reference. A transient storage timeout or backend hiccup will abort the export where the old code skipped and continued. That may be the intended fail-fast contract, but it is a behavior change worth confirming explicitly.
- `ProtobufExporter.java:158`: the `log.warn` is retained even though the request now fails, and the same failure will also be logged by the framework's exception handling. Consider `log.error`, or drop the line if the propagated exception is logged upstream, to avoid duplicate logging at the wrong severity.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/ProtobufExportTest.java
```diff
@@ -352,6 +353,54 @@ public void testExportNonExistentArtifactReturns404() throws Exception {
...
+
+        // Exporting the main artifact must fail (not silently return a partial zip with HTTP 200)
+        given()
+                .when()
+                .pathParam("groupId", GROUP)
+                .pathParam("artifactId", tableInfoArtifactId)
+                .pathParam("versionExpression", "branch=latest")
+                .get("/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{versionExpression}/export")
+                .then()
+                .statusCode(greaterThanOrEqualTo(400));
```
paoloantinori: `greaterThanOrEqualTo(400)` is too loose to prove this fix. It passes for any error response (a 404 from an unrelated lookup, a 500 from a different code path), so it does not confirm that the `RuntimeException` from `collectReferences` propagates through `exportVersionAsZip` into the intended 5xx. Pin the status (e.g. `500`) and assert the error body if one is returned; otherwise the test cannot distinguish the intended failure from an accidental one.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
```diff
@@ -156,6 +156,7 @@ private void collectReferences(List<ArtifactReferenceDto> references,
                 }
             } catch (Exception e) {
                 log.warn("Could not resolve reference: {} - {}", ref.getName(), e.getMessage());
+                throw new RuntimeException("Failed to resolve reference: " + ref.getName(), e);
```
paoloantinori: Throwing a bare `RuntimeException` bypasses the registry's exception hierarchy and `ExceptionMapper`, so the resulting HTTP status and error body are whatever the generic mapping produces rather than a deliberate one. If a domain exception type fits the situation (a not-found or server-error subtype), throw that instead so the response is deterministic and testable. Either way, confirm what status and body the generic mapper actually returns for this path.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
```diff
@@ -156,6 +156,7 @@ private void collectReferences(List<ArtifactReferenceDto> references,
                 }
             } catch (Exception e) {
```
paoloantinori, line 160: `catch (Exception e)` is broad, so this rethrow now aborts the entire export on any failure from `storage.getContentByReference()`, including transient ones like a storage timeout or a temporary backend error, not just a genuinely missing reference. The previous contract was to skip and continue. Confirm that fail-fast-on-any-error is the intended behavior here, or narrow the catch to the not-found case if transient errors should still be tolerated.

Nikhil-VS1811:
Hi, I’ve also implemented a fix for #9333 on my fork and pushed it to my branch.

I noticed PR #9365 already addresses the same issue, so I don’t want to create duplicate work. My implementation takes a slightly different approach to propagating the reference-resolution failure.

Please let me know whether I should open a separate PR or whether the changes would be useful for the existing PR.

paoloantinori:
Thanks for flagging this. Since #9365 is already under review and addresses #9333, the cleanest path is to share your alternative approach here as comments/suggestions rather than opening a competing PR. If your propagation strategy differs in a way that matters (error type, where the failure surfaces, test coverage), drop the specifics on this thread or on the relevant code lines and @deepak0x can weigh in on whether to fold them in. That keeps the decision in one place and avoids fragmenting the fix across two branches.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/ProtobufExportTest.java
deepak0x: Addressed. Pinned the assertion to statusCode(500) (was >= 400). rest-assured's default GZIP decoder choked on the compressed 500 JSON body, so I also disabled content decoding for this request via DecoderConfig.noContentDecoders() and set Accept-Encoding: identity. The test now deterministically proves the RegistryException propagates as 500, not an incidental 404/500. Verified locally: the test passes.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
deepak0x: Fixed. Replaced the bare RuntimeException with io.apicurio.registry.types.RegistryException, which is part of the registry exception hierarchy and is mapped deterministically to HTTP 500 by CoreRegistryExceptionMapperService (body is a ProblemDetails with title = the message). So the response is now deliberate and testable rather than whatever the generic mapper produced.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
deepak0x: Good point. I checked storage.getContentByReference(): it returns null for a missing reference (it does NOT throw - see SqlContentRepository lines 108/117). So the original catch only ever fired on genuine storage errors, not on 'reference not found'. The #9333 bug was a declared-but-unresolvable reference (null content) being silently skipped, producing a partial export. The fix now throws RegistryException explicitly when content == null for a declared reference - that is the deterministic, intended failure. The catch still rethrows as RegistryException for real storage errors, so fail-fast on a genuinely broken reference resolution is intentional (a silently partial export was the reported defect).

paoloantinori:
Thanks for the verification on `getContentByReference` — I confirmed it returns `null` for a missing reference (lines 108 and 117 of `SqlContentRepository`) rather than throwing, so your reading is correct. The original `catch` only fired on genuine storage errors, which makes the new `else` branch the right place for the deterministic "declared but unresolvable" failure.

All three points look resolved:
- Test pinned to `500` with the decoder workaround — good.
- `RegistryException` via `CoreRegistryExceptionMapperService` gives a deterministic response.
- Broad `catch` rethrow is fine given the null-vs-throw contract you traced.

No further concerns from me.

EricWittmann:
Thanks for the thorough fix and great responsiveness to review feedback @deepak0x! Merged.

PR #9366, opened 2026-08-07
fix: improve artifact reference validation

### LINKED ISSUE #9363: Artifact Reference Validation Accepts Invalid and URL-Unsafe Version Values
## Description

The Artifact References form currently accepts invalid characters in fields such as **Reference Name, Group ID, Artifact ID, and Version**.

For example, the **Version** field accepts values containing URL-unsafe and special characters such as:

* `1%2`
* `1/2`
* `1\2`
* `1#2`
* `1?2`
* `1@2`
* `1$2`
* `1^2`
* `1*2`
* `1(2)`
* `1 2`
* Non-ASCII characters such as `1.0.0🔥`

These values are accepted by the UI even though they are not valid according to the identifier/version format expected by the application. This can potentially cause problems when artifact references are stored, converted into URLs, parsed, or consumed by other components.

### Expected Behavior

The Artifact References form should validate `Reference Name`, `Group ID`, `Artifact ID`, and `Version` using the project's existing identifier validation rules.

Valid version formats such as:

* `1.2.3`
* `2.0.23`
* `10.2.1`
* `1.2.3-beta.1`
* `1.2.3+build.5`
* `2.0.0-SNAPSHOT`
* `v1.0.0`

should continue to be accepted, while URL-unsafe, unsupported special, whitespace, and non-ASCII characters should be rejected with a clear validation error.

### Proposed Fix

Centralize the validation through the existing `checkIdValid` utility and apply it consistently to all Artifact Reference fields.

Unit tests should also cover both valid version formats and invalid characters to prevent regressions.

vjkumar2756:
Closes #9363 

Summary
This PR fixes insufficient validation in the Artifact References form. Previously, the Version, Reference Name, Group ID, and Artifact ID fields could accept URL-unsafe, unsupported special, whitespace, and non-ASCII characters.
The fix reuses the project's centralized checkIdValid validation logic so that Artifact Reference fields follow the same validation rules consistently across the UI.

Changes
Enhanced checkIdValid in validation.utils.ts to reject:
URL-unsafe characters such as %, /, \, #, and ?
Unsupported special characters such as @, $, ^, *, (, and )
Spaces
Non-ASCII characters
Preserved supported characters required for valid artifact coordinates and versions:
a-z, A-Z, 0-9
., -, _, +, :, ,
Updated ReferencesFormGroup.tsx so validation is applied consistently to:
Reference Name
Group ID
Artifact ID
Version
Updated the form validation to prevent submission when invalid values are entered and display the appropriate validation error state.
Refactored CreateArtifactModal.tsx to reuse the shared checkIdValid utility instead of maintaining duplicate validation logic.
Added unit tests covering valid versions, URL-unsafe characters, unsupported special characters, spaces, and non-ASCII characters.
Validated Examples

The following valid versions continue to be accepted:
1.2.3
2.0.23
10.2.1
1.2.3-beta.1
1.2.3+build.5
2.0.0-SNAPSHOT
v1.0.0
Invalid examples such as 1%2, 1/2, 1#2, 1?2, 1@2, 1$2, 1^2, 1*2, 1(2), 1 2, and values containing non-ASCII characters are now rejected.

Testing
Added unit test coverage in validation.utils.test.ts to verify both accepted version formats and rejected invalid characters.
This keeps validation centralized, avoids duplicated logic, and prevents invalid Artifact Reference values from being submitted through the UI.

Before
[Screencast From 2026-08-07 23-25-58.webm](https://github.com/user-attachments/assets/71d0f975-66c6-47c3-bff1-887936df3976)

After
[Screencast From 2026-08-08 00-24-01.webm](https://github.com/user-attachments/assets/5820ffa1-a29e-4863-973b-920f820f3021)


ui/ui-app/src/utils/validation.utils.ts
```diff
@@ -12,7 +12,11 @@ export const checkIdValid = (id: string | undefined | null): boolean => {
         }
         return true;
     };
-    return id.indexOf("%") === -1 && isAscii(id);
+    const hasInvalidChars = (str: string) => {
+        // Disallow URL-unsafe and special characters that break REST API path routing or URI parsing
+        return /[%/\\#?@$^&*()\s]/.test(str);
```
paoloantinori: This denylist is meant to block "URL-unsafe" characters but still allows reserved/unwise ones (`<`, `>`, `[`, `]`, `{`, `}`, `"`, `'`, `|`) while banning `(`, `*`, and `&`. `&` is also rejected but is not mentioned in the PR description or covered by the new tests. Since the description already enumerates the allowed set (`a-z A-Z 0-9 . - _ + : ,`), a positive allowlist like `/^[a-zA-Z0-9._+:,\-]+$/` would be simpler, self-documenting, and unable to silently miss a character.

ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx
```diff
@@ -36,6 +36,7 @@ import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
 import { ArtifactLabel, LabelsFormGroup, ArtifactReferenceFormItem, ReferencesFormGroup, formItemsToReferences, isReferencesValid } from "@app/components";
 import { listToLabels } from "@utils/labels.utils.ts";
 import { detectContentType, detectVersionInContent } from "@utils/content.utils.ts";
+import { checkIdValid } from "@utils/validation.utils.ts";
```
paoloantinori, line 39: Importing the shared `checkIdValid` here replaces the local version, which only rejected `%` and non-ASCII. The shared function now also rejects `/`, `#`, `?`, `@`, `$`, `^`, `&`, `*`, `(`, `)`, and whitespace, so this is a behavioral change on the main artifact-creation form rather than a pure refactor. Confirm that no artifact IDs, group IDs, or versions that previously submitted are now rejected, and that the UI allowlist is no stricter than what the registry API accepts (otherwise values the API accepts get blocked only in the UI).

ui/ui-app/src/utils/validation.utils.ts
```diff
@@ -12,7 +12,11 @@ export const checkIdValid = (id: string | undefined | null): boolean => {
...
+    };
+    return !hasInvalidChars(id) && isAscii(id);
```
paoloantinori: `checkIdValid` is shared, so tightening it changes validation for every caller, not just `ReferencesFormGroup`. Please grep all call sites and confirm none relied on the previously permissive behavior (only `%` / non-ASCII rejection).

ui/ui-app/src/app/components/modals/ReferencesFormGroup.tsx
```diff
@@ -14,19 +15,22 @@ export type ArtifactReferenceFormItem = {
...
 
 /**
- * Returns true if all reference rows have all four fields populated.
+ * Returns true if all reference rows have all four fields populated and valid.
  * Returns true if there are no references (empty list is valid).
  */
 export const isReferencesValid = (items: ArtifactReferenceFormItem[]): boolean => {
     return items.every(item =>
-        item.name !== "" && item.groupId !== "" && item.artifactId !== "" && item.version !== ""
+        item.name !== "" && checkIdValid(item.name) &&
```
paoloantinori: `isReferencesValid` and `validateRefField` now gate on `checkIdValid`, but the added tests only cover `validation.utils.ts`. Add a test asserting a reference row with an invalid value (e.g. `name: "a/b"` or a value containing a space) is rejected by `isReferencesValid`, and that a fully valid row passes.

paoloantinori (commented):
Centralizing artifact-reference validation on the shared `checkIdValid` is the right direction, and the new unit tests for it are solid. The main concerns are that the denylist regex is incomplete relative to its stated "URL-unsafe" goal, and that reusing the now-stricter shared function tightens validation on the main CreateArtifact form (and every other caller) rather than being the pure refactor the description frames it as.

- `ui/ui-app/src/utils/validation.utils.ts:17`: the denylist `[%/\\#?@$^&*()\s]` claims to block "URL-unsafe" characters but still permits reserved/unwise characters (`<`, `>`, `[`, `]`, `{`, `}`, `"`, `'`, `|`) while banning `(`, `*`, and `&` (and `&` is neither mentioned in the description nor covered by the tests). The description already enumerates the allowed set, so a positive allowlist like `/^[a-zA-Z0-9._+:,\-]+$/` would be simpler, self-documenting, and unable to silently miss a character.
- `ui/ui-app/src/app/components/modals/CreateArtifactModal.tsx:39`: importing the shared `checkIdValid` replaces the local version, which only rejected `%` and non-ASCII. The shared function now also rejects `/`, `#`, `?`, `@`, `$`, `^`, `&`, `*`, `(`, `)`, and whitespace, so this is a behavioral change on the main artifact-creation form, not a pure refactor. Confirm no previously accepted artifact IDs, group IDs, or versions are now rejected, and that the UI rules are no stricter than what the registry API accepts.
- `ui/ui-app/src/utils/validation.utils.ts:19`: because `checkIdValid` is shared, this tightening changes validation for every caller, not just `ReferencesFormGroup`. Grep all call sites and confirm none relied on the previously permissive behavior (only `%` / non-ASCII rejection).
- `ui/ui-app/src/app/components/modals/ReferencesFormGroup.tsx:30`: `validateRefField` and `isReferencesValid` now gate on `checkIdValid`, but the added tests only cover `validation.utils.ts`. Add a test asserting a reference row with an invalid value (e.g. `name: "a/b"` or a value containing a space) is rejected, and that a clean row passes.
- This is client-side validation only; the registry API still needs to enforce the same rules independently, otherwise invalid values remain submittable through non-UI clients.

vjkumar2756:
@paoloantinori, thanks for the detailed feedback! I’ve addressed all the review comments regarding artifact-reference validation, including the allowlist-based validation, shared checkIdValid call sites, and the requested ReferencesFormGroup and validation tests. I’ve also verified the changes locally with the test suite, lint, and build, and everything is passing. Could you please take another look when you have a chance? Thanks!

paoloantinori:
This is exactly the shape I was hoping for. Switching from the denylist to an explicit allowlist (`/^[a-zA-Z0-9._+:,-]+$/`) is the right call: it closes the gap where new characters fall through unvalidated, and it makes the accepted set readable at a glance. Wiring both `ReferencesFormGroup` and `CreateArtifactModal` through the same shared `checkIdValid` in `validation.utils.ts` means there is now one place to update if the coordinate rules ever change.

The test coverage is thorough, including the version-string edge cases (`1.2.3+build.5`, `2.0.0-SNAPSHOT`, `group,id`) and the rejection of URL-unsafe and non-ASCII input.

No further concerns from me.

vjkumar2756:
@paoloantinori Hi , I’ve addressed the remaining points and everything looks good from my side now.
I just wanted to check if there’s anything else pending that might be preventing the PR from moving forward/being merged. If there’s any additional step or issue I may have missed, please let me know.
Thanks!

paoloantinori:
Thanks for following up. I had another look at `6d5268c` and the allowlist (`/^[a-zA-Z0-9._+:,-]+$/`) is applied consistently across all four reference fields and the artifact-create modal, with good coverage in `validation.utils.test.ts` and `ReferencesFormGroup.utils.test.ts`. From my side this is ready — nothing further pending. Over to a maintainer for `/accept`.

vjkumar2756:
Hi @EricWittmann  , I have  gone through the remaining feedback and made the requested updates. Everything should be in good shape now. Whenever you get a chance, could you please take another look at the PR.
I did  really appreciate your thoughts. 
Thanks!

vjkumar2756:
Hi @EricWittmann  , I have addressed the remaining feedback and updated the PR accordingly. 
Whenever you have some time, I did really appreciate it if you could give it another look. 
Please let me know if you think anything else needs to be adjusted.
Thanks for your time!

EricWittmann (commented):
Thanks for working on this, @vjkumar2756! Centralizing validation and adding tests are both good instincts. However, after reviewing the diff against the backend validation rules, there are some mismatches that would cause user-facing issues. I'd suggest closing this PR and taking another swing at a solution that accounts for the points below.

## Backend Validation Rules

The backend uses **different** validation rules for artifact/group IDs vs. version strings:

| Field | Backend rule | Source |
|-------|-------------|--------|
| **Artifact ID** | Any ASCII except `%`, max 512 chars | `ArtifactIdValidator.java` |
| **Group ID** | Any ASCII except `%`, max 512 chars | `ArtifactIdValidator.isGroupIdAllowed()` |
| **Version** | `[a-zA-Z0-9._\-+]{1,256}` only | `VersionId.java` |

The PR's single regex `/^[a-zA-Z0-9._+:,-]+$/` creates mismatches in **both directions**:

### 1. Artifact/Group IDs — UI becomes too restrictive

The old `checkIdValid` (ASCII + no `%`) actually **matched** the backend rules correctly. The new allowlist rejects characters the backend considers valid — for example `!`, `~`, `=`, `;`, `(`, `)`, `@`, `#`, `&`, `*`, spaces, and more. A user entering something like `my-artifact(v2)` would be blocked by the UI but accepted by the API.

### 2. Versions — UI allows characters the backend rejects

The backend's `VersionId` regex explicitly excludes `:` and `,` because those characters are reserved by `VersionExpressionParser`. The PR's regex allows both, so a user could enter version `1.0:rc1` — the UI would accept it, but the API would reject it with a 400 error.

## Other Issues

- **Error messages not updated.** `CreateArtifactModal.tsx` still shows *"Character % and non ASCII characters are not allowed"* on lines 448 and 470, which no longer accurately describes what the new regex rejects.

- **Local copies not consolidated.** `CreateDraftModal.tsx` and `NewDraftFromModal.tsx` still have their own local `checkIdValid` with the old (looser) behavior, creating inconsistent validation across forms.

## Suggested Approach for a Reworked PR

Rather than one shared regex for all fields, align the UI validators with what the backend actually enforces:

1. **Keep `checkIdValid` matching the backend's artifact/group ID rules** — ASCII, no `%`. That's what it was, and it was correct.
2. **Add a new `checkVersionValid`** using the backend's version regex: `/^[a-zA-Z0-9._\-+]{1,256}$/`.
3. **Apply the right validator to the right field** — `checkIdValid` for groupId, artifactId, and reference name; `checkVersionValid` for version.
4. **Consolidate the remaining local copies** of `checkIdValid` in `CreateDraftModal.tsx` and `NewDraftFromModal.tsx` to use the shared utility.
5. **Update error messages** to accurately describe what's rejected for each field type.

The extraction of `ReferencesFormGroup.utils.ts` and the test structure are both good — those parts could carry over to a reworked PR.

Also a reminder that the project follows a "one PR at a time" policy for external contributors — please make sure your other open PRs (#9414, #9523) are resolved before or alongside this one.

PR #9368, opened 2026-08-07
refactor(mcp): delegate inputSchema checks to JsonSchemaDiffLibrary

### LINKED ISSUE #9178: bug(mcp): McpToolCompatibilityChecker misses parameter type changes and schema-type removal due to lenient null guard

### LINKED ISSUE #9301: refactor(mcp): delegate inputSchema compatibility checks to JsonSchemaDiffLibrary
## Description

Follow-up to #9178 and PR #9181. During review, @jsenko identified that
McpToolCompatibilityChecker's inputSchema-level comparison logic
(checkInputSchemaTypeChange, checkPropertyRemovals,
checkPropertyTypeChanges, checkRequiredParamAdditions,
checkRequiredParamRemovals) reimplements a shallow version of
comparison logic the codebase already has in a more mature form —
JsonSchemaDiffLibrary (schema-util/json, everit-based).

The current shallow approach has real gaps the mature library already
solves: it compares `type` as a raw string, so semantically compatible
widening changes (e.g. `integer` → `number`) are incorrectly flagged as
breaking. It also only inspects first-level properties, missing nested
schema structure, `additionalProperties`, `enum`, `oneOf`/`anyOf`
evolution, and constraint changes (`minLength`, `maxLength`, `pattern`,
`minimum`, `maximum`).

### Proposed Fix
Delegate all inputSchema-level compatibility checking to
`JsonSchemaDiffLibrary.findDifferences()`, keeping only the MCP-tool
envelope checks (name, description, annotations) in
`McpToolCompatibilityChecker` — those aren't part of inputSchema and
fall outside the library's scope.

This closes #9178 (the original bugs — silent type removal, unchecked
property type changes) as a side effect of the correct, more complete
implementation, rather than continuing to patch the shallow reimplementation.

Ready to submit the PR once scoped.

Sudarshan25092007:
**Compatibility model:** This change adopts strict JSON Schema semantics for `inputSchema`.
Widening changes (removing constraints: type, properties, required entries, or the entire
inputSchema) are now correctly identified as backward compatible. Narrowing changes
(adding constraints: new typed properties in open schemas, adding required fields,
changing types) are correctly identified as incompatible. This fixes false positives
in the old shallow checker but is a behavioral change, not a pure refactor.

---
Closes #9301

Delegates all `inputSchema`-level compatibility checking in `McpToolCompatibilityChecker`
to the existing everit-based `JsonSchemaDiffLibrary` (schema-util/json) instead of
maintaining a shallow parallel implementation.

**Why the file move:** `McpToolCompatibilityChecker` was in `schema-util/common`,
but `JsonSchemaDiffLibrary` is in `schema-util/json`, which already depends on
`schema-util/common`. Adding the reverse dependency would create a circular Maven
dependency. Relocating both classes to `schema-util/json` resolves this cleanly.

**What changes:**
- `McpToolCompatibilityChecker` + `McpToolCompatibilityDifference` relocated from
  `schema-util/common` → `schema-util/json`
- 11 shallow helper methods removed (`checkInputSchemaTypeChange`,
  `checkPropertyRemovals`, `checkPropertyTypeChanges`, etc.)
- `inputSchema` comparison now delegates to
  `JsonSchemaDiffLibrary.getIncompatibleDifferences()`
- Envelope checks (name, description, annotations) remain as-is
- 5 false-positive test cases corrected (widening changes now correctly identified
  as compatible by the library)
- Non-object `inputSchema` values (e.g. boolean schemas) guarded before delegation
- Whole `inputSchema` removal treated as widening → compatible
- Dead enum constants removed; only `INPUT_SCHEMA_INCOMPATIBLE` and `PARSE_ERROR` remain


Sudarshan25092007:
Hi @jsenko — this implements the delegation approach you outlined in #9181 (Option A).
All `inputSchema` checks now route through `JsonSchemaDiffLibrary`. The 5 false-positive
test cases we discussed in #9301 have been flipped to match correct JSON Schema semantics.
Let me know your reviews so that i can close #9181 and update this PR ready for review.

@EricWittmann could you please assign @jsenko (and @paoloantinori if available) as reviewers?
This follows directly from the architecture redirect on #9181. 
Thanks!

paoloantinori:
Thanks for the heads-up. I'll take a look at this — the delegation to JsonSchemaDiffLibrary is the right direction for the inputSchema checks. I recognize this area from the earlier compatibility-checker work, so I'll review it shortly.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -0,0 +1,102 @@
...
+        Set<McpToolCompatibilityDifference> differences = new HashSet<>();
+
+        try {
+            JsonNode existingNode = MAPPER.readTree(existing);
+            JsonNode proposedNode = MAPPER.readTree(proposed);
+
+            JsonNode existingSchemaNode = existingNode.get("inputSchema");
+            JsonNode proposedSchemaNode = proposedNode.get("inputSchema");
+
+            boolean existingMissing = existingSchemaNode == null || existingSchemaNode.isNull();
```
paoloantinori: The isNull() guard only matches the JSON null literal, so a non-object inputSchema (a boolean schema such as true or false, an array, or a scalar) is treated as present and falls through to MAPPER.writeValueAsString(...) and JsonSchemaDiffLibrary.getIncompatibleDifferences(...). The library parses the schema through the json-org module, which rejects non-object values, so a valid boolean JSON Schema ("inputSchema": true) throws and is caught as PARSE_ERROR, flipping the verdict from COMPATIBLE (the old checker no-op'd on non-object schemas) to INCOMPATIBLE with a misleading "parse" message. This path is reachable at SYNTAX_ONLY/NONE validity, where McpToolContentValidator does not reject a non-object inputSchema. Guard for object-ness before delegating (or handle boolean schemas) and add a test for inputSchema: true.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
```diff
@@ -61,7 +62,8 @@ void testBackwardCompatibleAddingOptionalProperty() {
                         "properties": {
                             "query": { "type": "string" }
                         },
-                        "required": ["query"]
+                        "required": ["query"],
+                        "additionalProperties": false
```
paoloantinori, line 104: The existing schema here was open on main (additionalProperties defaults to true). Adding additionalProperties: false is what makes the assertion pass: under an open schema, adding a typed optional property (limit: integer) narrows the accepted set, because limit was previously accepted as any type via additionalProperties, so the library flags it incompatible. The old checker's rule was "adding an optional parameter is always compatible", which no longer holds for open schemas. This fixture edit masks the behavioral change instead of documenting it. If the stricter behavior is intended, state it in the PR description and add a test asserting the open-schema case is now incompatible; if not, the delegation is stricter than the old contract.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            if (existingMissing && !proposedMissing) {
+                differences.add(new McpToolCompatibilityDifference(
+                        McpToolCompatibilityDifference.Type.INPUT_SCHEMA_INCOMPATIBLE,
+                        "inputSchema was added"));
+                return differences;
+            }
+
+            if (!existingMissing && proposedMissing) {
```
paoloantinori, line 62: This branch marks removing the entire inputSchema as incompatible, but the PR adopts a widening-is-compatible model everywhere else (removing the type, removing properties, and removing required entries are all now compatible). Removing the whole schema is also a widening, since the tool then accepts any input, so on consistency grounds it should be compatible too. Either justify why the whole-schema case is treated differently or align it. Neither the add-inputSchema branch (line 47) nor this remove branch has test coverage in the diff; add tests for both before relying on them.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/McpToolCompatibilityChecker.java
```diff
@@ -0,0 +1,102 @@
...
+            Map<String, TypedContent> refs, Set<McpToolCompatibilityDifference> differences) {
+        try {
+            Set<Difference> diffs = JsonSchemaDiffLibrary.getIncompatibleDifferences(
+                    existingSchemaStr, proposedSchemaStr, refs);
+            for (Difference diff : diffs) {
+                differences.add(new McpToolCompatibilityDifference(
+                        McpToolCompatibilityDifference.Type.INPUT_SCHEMA_INCOMPATIBLE,
+                        diff.getDiffType().getDescription() + " at " + diff.getPathUpdated()));
+            }
+        } catch (RuntimeException e) {
```
paoloantinori, line 139: Catching RuntimeException and labeling it PARSE_ERROR conflates every library failure with a parse problem. A JsonSchemaDiffLibrary bug, an unsupported schema construct, or the non-object-schema throw noted above all surface to the user as "Failed to parse MCP tool definition", and because a difference is added the update is blocked (fail-closed) with a misleading reason. Use a distinct difference type for delegation failures or at least preserve the exception class in the message, and decide whether an unexpected library error should fail-closed or be logged and rethrown.

schema-util/json/src/main/java/io/apicurio/registry/json/rules/compatibility/McpToolCompatibilityDifference.java
```diff
@@ -14,7 +15,8 @@ public enum Type {
         REQUIRED_PARAM_REMOVED("inputSchema/required"),
```
paoloantinori: After this refactor the checker only ever emits INPUT_SCHEMA_INCOMPATIBLE and PARSE_ERROR, so REQUIRED_PARAM_ADDED, REQUIRED_PARAM_REMOVED, INPUT_SCHEMA_TYPE_CHANGED, and PROPERTY_REMOVED (and their context paths inputSchema/required, inputSchema/type, inputSchema/properties) are dead. Remove the unused constants, or if they are retained for an external contract, note why.

paoloantinori (commented):
The delegation to JsonSchemaDiffLibrary is a reasonable simplification, and relocating the checker to schema-util/json correctly avoids the circular dependency (verified: the class is wired only via the import and the McpToolCompatibilityChecker::new reference in StandardArtifactTypeProviderRegistry, with no service-loader registration). However, this is not the behavior-preserving refactor the title implies: it changes compatibility verdicts in both directions and masks one regression with a test-fixture edit. The behavioral changes should be spelled out and covered before merge.

- Four existing assertions flip from incompatible to compatible (property removal, required-param removal, inputSchema type removal, property-type removal). These are defensible as false-positive fixes under the standard "old data still valid" model, but the title frames the work as delegation rather than a semantics change. State the model explicitly in the description.
- Adding a typed optional property to an open schema (additionalProperties defaults to true) is now incompatible, so the old rule "adding optional parameters is always compatible" no longer holds. The fixture edit in testBackwardCompatibleAddingOptionalProperty hides this rather than documenting it.
- The two catch blocks at lines 68 and 72 of the new checker are byte-for-byte identical; the JsonProcessingException catch is redundant.
- All library differences collapse to the INPUT_SCHEMA_INCOMPATIBLE context, losing the granular inputSchema/required, /type, /properties paths the old checker emitted (the path survives only in the free-text message via getPathUpdated()).

Sudarshan25092007:
@paoloantinori All review points addressed in the latest force-push. Summary:

1. **Semantics documented** — PR description now explicitly states the widening-is-compatible JSON Schema model.
2. **Open-schema optional property** — Reverted fixture edit in `testBackwardCompatibleAddingOptionalProperty`; now `assertFalse` with comment explaining open-schema narrowing. Added `testBackwardCompatibleAddingOptionalPropertyToClosedSchema` for the closed-schema case.
3. **Redundant catches** — Merged into single `catch (Exception e)`.
4. **Granular paths** — Accepted tradeoff; removed dead enum constants so the collapsed type is honest.
5. **Non-object inputSchema** — Added `!isObject()` guard; non-object schemas skip delegation. Added `testBackwardCompatibleNonObjectInputSchema`.
6. **Whole-schema removal** — Aligned with widening model: removing `inputSchema` entirely now returns compatible. Added `testBackwardCompatibleRemovingInputSchema`.
7. **RuntimeException catch** — Preserves exception class name: `"Schema comparison failed (" + e.getClass().getSimpleName() + "): ..."`.
8. **Dead enum constants** — Removed `REQUIRED_PARAM_ADDED`, `REQUIRED_PARAM_REMOVED`, `INPUT_SCHEMA_TYPE_CHANGED`, `PROPERTY_TYPE_CHANGED`, `PROPERTY_REMOVED`. Only `INPUT_SCHEMA_INCOMPATIBLE` and `PARSE_ERROR` remain.
9. **Sonar S5976** — Added `@SuppressWarnings("java:S5976")` to test class.

All 21 tests pass, checkstyle clean, 0 Sonar issues. Ready for re-review.

paoloantinori:
Re-read the latest head — all three points landed cleanly:

- The widening-is-compatible model is now stated explicitly in the PR description, which removes the ambiguity I flagged earlier.
- The open-schema / closed-schema split is correct: the original test now asserts incompatible for optional-property addition under default (open) additionalProperties, and the new closed-schema test covers the widening case. Good coverage.
- The catch consolidation in the new checker reads cleaner — a single RuntimeException catch around the delegation plus the outer parse guard.

The delegation to JsonSchemaDiffLibrary is solid from my side. Nothing further from me.

Sudarshan25092007:
@jsenko — this PR implements the delegation approach exactly as you outlined in #9181 (Option A). Quick mapping to your implementation sketch:

- **Extracted `inputSchema` JsonNode** from both existing and proposed tool definitions.
- **Serialized to strings** and delegated to `JsonSchemaDiffLibrary`.
- **Used `getIncompatibleDifferences()` directly** instead of `findDifferences()` + manual `DiffContext` filtering — this returns the pre-filtered incompatible `Set&lt;Difference&gt;` in one call, which simplifies the glue code. Happy to switch to the explicit `findDifferences()` → `getIncompatibleDifferences()` path if you prefer.
- **Mapped `DiffType` descriptions** to `McpToolCompatibilityDifference` with path context (`description + " at " + path`).
- **Envelope checks** (name, description, annotations) left untouched — treated as always compatible.
- **Everit-based implementation** confirmed; Draft-07 default validated (MCP schemas carry no `$schema` header).

@paoloantinori has reviewed the latest head and has no further concerns. Ready for your review whenever you have bandwidth. Thanks!

paoloantinori:
Confirming from my side as well — the delegation to `JsonSchemaDiffLibrary.getIncompatibleDifferences()` is the right shape, and the widening-is-compatible model is now stated explicitly in the PR description, which closes the ambiguity I raised earlier. Tests cover both the widening cases (property/type removal) and the narrowing cases (typed property added to an open schema, type change). Nothing further from me.

Sudarshan25092007:
Hi @jsenko, Just a quick heads-up, as CI is on waiting queue for your review and acceptance.
Let me know your reviews on this, Happy to iterate, if the direction is off from the one that you guided..!😊

aclfe:
Went through this in depth, ran it locally too (21/21 tests pass, checkstyle clean on schema-util/json and schema-util/util-provider). Delegation direction is right, and integer -> number from #9301 does get fixed properly now. A few things before this feels ready though.

Maybe I'm missing something, but "$schema": "https://json-schema.org/draft/2020-12/schema" looks like a real problem? everit only understands draft-04/06/07, JsonSchemaDiffLibrary falls back to DRAFT_7 for the version lookup but SchemaLoader.load() still throws SchemaException ("could not determine version"). That gets caught and turned into a PARSE_ERROR, which counts as incompatible. MCP SDKs emit 2020-12 schemas by default, so this could flag basically every real tool as incompatible on every new version.

Bigger one: adding an optional property to an open schema now comes back incompatible (propertySchemasAdded narrowing), which flips the old rule ("Adding optional input parameters: Always compatible" was in the deleted class javadoc). testBackwardCompatibleAddingOptionalProperty got rewritten to assert the opposite without really calling out that it's a behavior change. That's probably the most common way MCP tools evolve day to day, so worth confirming with @jsenko this is the intended model, not just something that fell out of delegating everything.

Also RuleViolation context is now always /inputSchema for every diff (McpToolCompatibilityDifference.Type.getContext()), old enum had /inputSchema/type, /inputSchema/required etc, real path is just concatenated into the description string instead. JsonSchemaCompatibilityDifference in the same package already does this right with difference.getPathUpdated(), might be worth reusing that.

On "closes #9178", only half of it is actually fixed. Property type change (age integer -> string) works. inputSchema.type removal is not fixed, it's now deliberately treated as compatible (widening), which is a different position than the issue asked for, so probably shouldn't auto-close on that one.

On the module move, the circular dependency reasoning makes sense (common can't depend on json). But only McpToolCompatibilityChecker moves, the rest of MCP stays in common, and AgentCardCompatibilityChecker / PromptTemplateCompatibilityChecker / ModelSchemaCompatibilityChecker have the same JSON Schema problem waiting for them. Might be worth a schema-util/mcp module instead of splitting one class out each time. These are also public classes in schema-util-common's jar today, so this is a binary compat break for embedders, worth calling out explicitly in the description.

Let me know if I'm off base on any of these.

Sudarshan25092007:
@aclfe Thanks for the thorough read, a few of these are real and I'm addressing them now.

**Draft 2020-12 ($schema header)**, You're right, this is a defensive gap. everit only handles Draft 04/06/07. If an MCP SDK emits a 2020-12 `$schema`, the library throws and we surface it as a misleading `PARSE_ERROR`. I'm pushing a small fix that strips the `$schema` field from the inputSchema node before delegating, so everit defaults to Draft-07 behavior regardless of what the SDK emits.

**"`Closes #9178`"** — Also correct, and thank you for catching this. The original issue asked for type removal to be flagged as *incompatible*. The delegation approach treats it as widening → *compatible*. That's the correct JSON Schema semantics, but it means this PR does not close `#9178` as originally written. I'll remove the "`Closes 9178`" claim from the description and add a note explaining the semantic redirect.

**Optional property in open schema** — This narrowing behavior is the correct JSON Schema result (adding a typed property to an open schema restricts previously accepted values). @paoloantinori reviewed and accepted this model explicitly. I'll wait for @jsenko to confirm it's the intended contract before changing anything.

**RuleViolation context** — Accepted tradeoff for now. The old hardcoded paths (`/inputSchema/type`, `/inputSchema/required`) were lost when we collapsed to a single `INPUT_SCHEMA_INCOMPATIBLE` type. The real path survives in the description string via `getPathUpdated()`. Improving this is follow-up work.

**Binary compat break** — True, moving public classes from `schema-util/common` to `schema-util/json` is a breaking change for embedders. Since this is a 3.x snapshot and the maintainers didn't flag it during review, I'll add an explicit callout in the PR description rather than rework the module structure in this PR.

Fix for `$schema` stripping is incoming in the next force-push. Thanks again for the close read.

paoloantinori:
Confirming my earlier read for the record: the widening-is-compatible model for `inputSchema` was reviewed and accepted on my side, and the open-schema narrowing consequence follows from it. That said, I agree the final call on the contract belongs to @jsenko, since the delegation approach does redirect #9178's original ask (type removal now widening/compatible rather than incompatible) — that's a semantic decision worth an explicit maintainer sign-off, not something to sneak in via a refactor.

The `$schema`-stripping guard for everit's Draft 2020-12 gap sounds like the right pragmatic fix, and calling out the binary-compat break from the package move in the description is the right immediate step. Happy to re-read once the force-push lands.

Sudarshan25092007:
@paoloantinori Thanks for confirming — much appreciated.

Quick note on the CI state: I force-pushed twice in quick succession while addressing the `$schema` review point, which seems to have caused the SonarQube analysis to fail mid-run ("The last analysis has failed"). Could a maintainer trigger a `/retry` when convenient to re-run the scan?

Also seeing `License Compliance — 4 issues found` on this push. I only added the `$schema` stripping logic (no new dependencies or files). Is this a pre-existing issue from the rebase, or should I investigate further? Happy to fix if it's related to the file relocation.

paoloantinori:
Checked both from my side:

- **SonarCloud**: the analysis is green on the current head (`10ac2e6`) — the failure you saw was on the intermediate push, so no retry is needed.

- **License Compliance**: the same "4 issues found" status is present on `main` itself right now, so it is pre-existing and unrelated to this PR (neither the file relocation nor the `$schema` stripping touches dependencies or licenses). Nothing to investigate or fix here.

So from a CI perspective this head is as clean as main is. The remaining blocker is just the maintainer review queue.

jsenko (commented):
Closes #9301 — delegates inputSchema compatibility checking to JsonSchemaDiffLibrary, fixing false positives from the old shallow checker.

### Summary

Clean refactoring with a correct behavioral change. The delegation to `JsonSchemaDiffLibrary` gives proper JSON Schema diff semantics — widening changes (removing constraints) are correctly backward compatible, narrowing changes (adding constraints) are correctly incompatible. The test suite is thorough (20 tests covering open/closed schema semantics, required properties, type changes, union types, envelope fields, and error paths).

The module move from `schema-util/common` to `schema-util/json` is correct — `util-provider` already depends on `json`, and the new code needs `JsonSchemaDiffLibrary`.

### should-fix: FQN usage on lines 61 and 64

`McpToolCompatibilityChecker.java` lines 61 and 64 use fully qualified `com.fasterxml.jackson.databind.node.ObjectNode` instead of importing. Add an import and use the short class name:

```java
import com.fasterxml.jackson.databind.node.ObjectNode;
// ...
((ObjectNode) existingSchemaNode).remove("$schema");
((ObjectNode) proposedSchemaNode).remove("$schema");
```

### should-fix: Inconsistent null handling in exception messages

Line 77 uses `e.getMessage()` directly, which can return `null` (producing "Failed to parse MCP tool definition: null"). Line 97 in the same file uses a null-safe pattern. Apply the same pattern to line 77:

```java
"Failed to parse MCP tool definition: " +
    (e.getMessage() != null ? e.getMessage() : e.toString())
```

No further issues found.

Sudarshan25092007:
Hi @jsenko, thanks for the review!

I've addressed both items in the latest amended commit:

1. **FQN cleanup (`ObjectNode`)**: Added `import com.fasterxml.jackson.databind.node.ObjectNode;` and replaced the inline fully-qualified casts with `(ObjectNode)`.
2. **Null-safe exception message**: Updated the parse error `catch` block on line 77 to use `(e.getMessage() != null ? e.getMessage() : e.toString())`, consistent with line 97.

Screenshot 2026-08-18 010724

All 21 tests in `McpToolCompatibilityCheckerTest` and Checkstyle checks pass cleanly. Ready for re-review!

paoloantinori (commented):
The delegation to `JsonSchemaDiffLibrary` is the right call and the file move to avoid the circular Maven dependency is justified, but the PR is a real behavior change (five test verdicts flipped) and the write-up in the description matches what the code does. The main concerns are: dialect drift after `$schema` stripping can silently mis-evaluate Draft 2020-12 schemas, library failures are collapsed into `PARSE_ERROR` (blocking valid updates), the `inputSchema: false` boolean case is folded into "missing", and several test names now assert the opposite of what their names say.

- `McpToolCompatibilityChecker.java:66-71` (new file): stripping `$schema` means a Draft 2020-12 schema is silently interpreted under Draft 07 semantics by everit. Keywords like `prefixItems`, top-level `$defs` usage, or 2020-12 `items` arrays will not error, they will just be evaluated wrongly, so a compatible change can be reported incompatible (or vice versa) with no signal. Consider detecting an unsupported `$schema` value and failing fast with a clear message instead of stripping.
- `McpToolCompatibilityChecker.java:91-97`: any `RuntimeException` from the diff library becomes a `PARSE_ERROR` difference. If `PARSE_ERROR` maps to "incompatible", a library limitation on a legal schema blocks a valid update and the message won't say which side failed to load. Verify what `PARSE_ERROR` renders as in the rule outcome, and prefer distinguishing "library cannot handle this schema" from "document is not valid JSON" (the outer catch already covers the parse case).
- `McpToolCompatibilityChecker.java:39-43`: `existingMissing` treats any non-object `inputSchema` as missing, including the boolean schema `false`, which accepts nothing. Transitioning `false` → a real object schema is widening, but this code reports it as an incompatible addition; `true` → object is narrowing but is skipped as compatible. If boolean schemas are in scope for MCP tool artifacts, handle them explicitly; if not, say so in a comment.
- `McpToolCompatibilityCheckerTest.java:86-92` and the sibling flipped tests (`testBackwardIncompatibleRemovingProperty`, `testBackwardIncompatibleRemovingRequiredParam`, `testBackwardIncompatibleRemovingInputSchemaType`, `testBackwardIncompatibleRemovingPropertyType`): the assertions were correctly flipped but the method names still claim the opposite (e.g. `testBackwardCompatibleAddingOptionalProperty` now asserts `assertFalse`). Rename them so the name states the expected verdict; a future reader grepping for the compatible-adding case will find a test that asserts incompatibility.
- `StandardArtifactTypeProviderRegistry.java:86`: only this one import is updated in the diff. Grep the whole repo for the old FQCN `io.apicurio.registry.rules.compatibility.McpToolCompatibilityChecker` (and for `McpToolCompatibilityDifference`, which also moved packages) to confirm no other module, service loader config, or documentation still references it, and confirm `util-provider` already declares a dependency on `schema-util/json`.

Sudarshan25092007:
Hi @paoloantinori — thanks for the close re-read. All 5 points addressed in the latest amended commit (e59998530):
1. Unsupported draft validation — Replaced $schema stripping with explicit draft detection. If $schema indicates anything outside Draft 04/06/07, the checker adds a PARSE_ERROR with a clear message naming the unsupported URI and which side (existing/proposed) had it. No silent mis-evaluation.
2. Per-schema pre-validation — Added individual pre-validation of existing and proposed schemas before the comparison call, so parse/validation failures are reported as Failed to parse existing inputSchema (...) vs Failed to parse proposed inputSchema (...). The main comparison RuntimeException catch now reads Unexpected schema comparison error (...) instead of "parse," distinguishing library failures from document parse errors.
3. Boolean schema handling — Added an explicit comment above the isObject() guard explaining that boolean schemas (true/false) are folded into "missing" because everit does not support non-object values, and that MCP tool inputSchemas are object schemas in practice.
4. Test names — Renamed all 5 flipped tests so method names match their assertions (e.g., testBackwardCompatibleAddingOptionalProperty → testBackwardIncompatibleAddingOptionalPropertyToOpenSchema).
5. Stale references — Grep'd the full repo for old FQCNs: 0 matches for both McpToolCompatibilityChecker and McpToolCompatibilityDifference in the old package. No service-loader configs or other modules reference the old location. util-provider already declares the schema-util/json dependency in its POM.

#### Verification: 26/26 tests pass, Checkstyle 0 violations, Sonar 0 new issues. The ObjectNode FQN and null-safe message that @jsenko flagged earlier are also included in this commit.

Two quick maintainer asks if you have a moment:
- The CI workflow is showing "1 workflow awaiting approval" — could you hit Approve and run when convenient?
- The triage bot is flagging #9301 for lacking a maintainer comment on the issue itself. Would you mind dropping a quick /approve or even just an "Approved for implementation" comment on [#9301](https://www.kimi.ai/chat/link-to-issue)? That should clear the automated blocker.
Ready for re-review. Thanks again!

paoloantinori:
Re-read the amended head (e5999853) — all five points landed, verified against the diff:

- Draft validation: SUPPORTED_DRAFTS + validateSchemaDraft() reports the offending URI and which side had it, with tests covering both the existing and proposed cases.
- Per-schema pre-validation: the two pre-flight getIncompatibleDifferences(schema, "{}", ...) calls give distinct "Failed to parse existing/proposed inputSchema" messages, and the comparison catch is now "Unexpected schema comparison error" — a clean separation of document errors from library failures.
- The boolean-schema rationale comment sits above the isObject() guard as described.
- The flipped test names now read correctly.
- Stale references: a repo-wide search on my side also finds no remaining references to the old io.apicurio.registry.rules.compatibility.McpTool* FQNs other than the import this PR itself updates.

One small leftover in the same class as your point 4: testBackwardCompatibleNonObjectInputSchema still says "Compatible" in the name but asserts assertFalse(result.isCompatible()). The verdict is right (boolean/absent schema → structured object schema is narrowing, so incompatible); the name just missed the rename pass. Trivial fix next time you touch the file.

On the two maintainer asks (workflow approval and the #9301 comment) — leaving those to @jsenko and the maintainer queue.

Sudarshan25092007:
Thanks for the thorough re-read, @paoloantinori, glad all five points checked out. 
Noted on `testBackwardCompatibleNonObjectInputSchema`, I'll fold that rename into the next commit if @jsenko asks for any changes, or handle it in a follow-up if this lands as-is. Appreciate the time.

paoloantinori:
No problem, glad it all landed cleanly.
On the rename: agreed it's not worth a force-push on its own. Folding it into the next commit if @jsenko requests changes, or a tiny follow-up after merge, both work fine from my side.
This looks good to me as it stands.

carlesarnal:

Sudarshan25092007:
Hi @carlesarnal, thanks for the clarification. I understand that this scope is being reserved for the LFX term.

I’ve already applied for the Apicurio mentorship for the MCP Tool Validation and Compatibility Checking project, and this PR is part of the pre-mentorship work I’ve been doing in that area. If selected, I’d be happy to continue from this point under the mentorship structure and adapt the scope based on the program’s direction.

In the meantime, is there an adjacent area of the registry where I could contribute without overlapping with the reserved mentorship scope?

Thanks for keeping the boundaries clear around the program. 😊

PR #9370, opened 2026-08-07
fix(rest): normalize pagination parameters in v2 REST endpoints

### LINKED ISSUE #9369: Negative or overflowing pagination parameters are not validated in v2 REST API endpoints
## Summary

Several v2 REST API endpoints appear to forward `offset` and `limit` query parameters to the storage layer without applying the pagination normalization already used by the v3 implementation.

As a result, invalid values (such as negative values or sufficiently large values that overflow `BigInteger.intValue()`) can reach the storage layer and may result in an HTTP 500 response instead of being normalized or rejected with a client error.

The same class of issue has already been addressed for equivalent v3 endpoints, but the corresponding v2 endpoints still appear to bypass that validation.

---

## Affected Endpoints

The following endpoints appear to forward pagination values directly without bounds checking:

- `GET /apis/registry/v2/search/artifacts`
- `POST /apis/registry/v2/search/artifacts`
- `GET /apis/registry/v2/groups`
- `GET /apis/registry/v2/groups/{groupId}/artifacts`
- `GET /apis/registry/v2/groups/{groupId}/artifacts/{artifactId}/versions`

---

## Steps to Reproduce

For example:

```http
GET /apis/registry/v2/search/artifacts?offset=-1
```

or

```http
GET /apis/registry/v2/groups?offset=-1
```

or

```http
GET /apis/registry/v2/groups/{groupId}/artifacts/{artifactId}/versions?limit=-1
```

---

## Expected Behavior

Invalid pagination values should either:

- be normalized to safe values, or
- return `400 Bad Request`

matching the behavior already implemented for the corresponding v3 endpoints.

No internal storage or SQL exception details should be exposed.

---

## Actual Behavior

The request appears to reach the storage layer with invalid pagination values.

Depending on the supplied value, the request results in an HTTP 500 response rather than being normalized or rejected as invalid.

---

## Root Cause

The v2 REST implementation appears to perform null checks on `offset` and `limit`, but forwards them directly using `intValue()` without applying the normalization logic already used by the newer API implementation.

This allows invalid values (including integer-overflow cases) to propagate into SQL query generation.

---

## Possible Fix

Reuse the pagination normalization already applied in the equivalent v3 implementation so that both API versions handle invalid pagination parameters consistently.

Regression tests covering negative and overflowing `offset`/`limit` values would help prevent future regressions.

Regression tests covering:

- negative `offset`
- negative `limit`
- overflowing `limit`
- overflowing `offset`

would help prevent regressions.

Adii-45:
## Summary

Fixes #9369

This PR aligns pagination parameter handling in the v2 REST API with the existing normalization behavior already used by the equivalent v3 endpoints.

Previously, several v2 REST endpoints forwarded `offset` and `limit` directly to the storage layer after only performing null checks. Invalid values (negative values or values large enough to overflow `intValue()`) could therefore propagate to the storage implementation and result in HTTP 500 responses instead of being normalized before execution.

This change extracts the existing pagination normalization logic into a shared utility and updates all affected v2 endpoints to reuse it.

### Shared pagination normalization

- Added shared pagination normalization methods to `ParameterValidationUtils`
  - `normalizeOffset(BigInteger)`
  - `normalizeLimit(BigInteger)`
- Centralized the normalization logic to avoid duplication across API versions.

### v3 refactoring

- Updated the v3 `SearchResourceImpl` to use the shared helper.
- Removed the now-duplicated private normalization methods while preserving existing behavior.

### v2 fixes

Updated all affected v2 REST endpoints to normalize pagination parameters before reaching the storage layer:

- `GET /apis/registry/v2/search/artifacts`
- `POST /apis/registry/v2/search/artifacts`
- `GET /apis/registry/v2/groups`
- `GET /apis/registry/v2/groups/{groupId}/artifacts`
- `GET /apis/registry/v2/groups/{groupId}/artifacts/{artifactId}/versions`

These endpoints normalize negative and overflowing values the same way v3 does, but preserve v2's pre-existing unbounded limit (v2 never had a page-size cap at any layer - REST annotation, OpenAPI spec, or storage). v3's existing 1000-row cap (from #8611) is untouched and does not apply to v2, to avoid introducing an undocumented breaking change on a stable API.

## Tests

Added regression coverage for the v2 pagination behavior, including:

- negative offset
- negative limit
- offset overflow
- limit overflow
- omitted pagination parameters
- boundary values
- preservation of existing `limit=0` semantics

Also verified that the existing v3 test suite continues to pass after the helper extraction.

## Result

Invalid pagination parameters are now normalized consistently across both v2 and v3 REST APIs, preventing invalid values from propagating to the storage layer and eliminating the HTTP 500 behavior described in #9369.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v2/impl/GroupsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v2/impl/GroupsResourceImpl.java
@@ -401,8 +401,8 @@ public class GroupsResourceImpl implements GroupsResource {
 
         Set<SearchFilter> filters = Collections.emptySet();
 
-        GroupSearchResultsDto resultsDto = storage.searchGroups(filters, oBy, oDir, offset.intValue(),
-                limit.intValue());
+        GroupSearchResultsDto resultsDto = storage.searchGroups(filters, oBy, oDir,
+                ParameterValidationUtils.normalizeOffset(offset), ParameterValidationUtils.normalizeLimit(limit));
         return V2ApiUtil.dtoToSearchResults(resultsDto);
     }
 
@@ -901,8 +901,9 @@ public class GroupsResourceImpl implements GroupsResource {
         Set<SearchFilter> filters = new HashSet<>();
         filters.add(SearchFilter.ofGroupId(defaultGroupIdToNull(groupId)));
 
-        ArtifactSearchResultsDto resultsDto = storage.searchArtifacts(filters, oBy, oDir, offset.intValue(),
-                limit.intValue(), false);
+        ArtifactSearchResultsDto resultsDto = storage.searchArtifacts(filters, oBy, oDir,
+                ParameterValidationUtils.normalizeOffset(offset), ParameterValidationUtils.normalizeLimit(limit),
+                false);
         return V2ApiUtil.dtoToSearchResults(resultsDto);
     }
 
@@ -1136,7 +1137,8 @@ public class GroupsResourceImpl implements GroupsResource {
         Set<SearchFilter> filters = Set.of(SearchFilter.ofGroupId(defaultGroupIdToNull(groupId)),
                 SearchFilter.ofArtifactId(artifactId));
         VersionSearchResultsDto resultsDto = storage.searchVersions(filters, OrderBy.createdOn,
-                OrderDirection.asc, offset.intValue(), limit.intValue(), false);
+                OrderDirection.asc, ParameterValidationUtils.normalizeOffset(offset),
+                ParameterValidationUtils.normalizeLimit(limit), false);
         return V2ApiUtil.dtoToSearchResults(resultsDto);
     }
 
--- a/app/src/main/java/io/apicurio/registry/rest/v2/impl/GroupsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v2/impl/GroupsResourceImpl.java
@@ -402,7 +402,8 @@ public class GroupsResourceImpl implements GroupsResource {
         Set<SearchFilter> filters = Collections.emptySet();
 
         GroupSearchResultsDto resultsDto = storage.searchGroups(filters, oBy, oDir,
-                ParameterValidationUtils.normalizeOffset(offset), ParameterValidationUtils.normalizeLimit(limit));
+                ParameterValidationUtils.normalizeOffset(offset),
+                ParameterValidationUtils.normalizeLimitUnbounded(limit));
         return V2ApiUtil.dtoToSearchResults(resultsDto);
     }
 
@@ -902,8 +903,8 @@ public class GroupsResourceImpl implements GroupsResource {
         filters.add(SearchFilter.ofGroupId(defaultGroupIdToNull(groupId)));
 
         ArtifactSearchResultsDto resultsDto = storage.searchArtifacts(filters, oBy, oDir,
-                ParameterValidationUtils.normalizeOffset(offset), ParameterValidationUtils.normalizeLimit(limit),
-                false);
+                ParameterValidationUtils.normalizeOffset(offset),
+                ParameterValidationUtils.normalizeLimitUnbounded(limit), false);
         return V2ApiUtil.dtoToSearchResults(resultsDto);
     }
 
@@ -1138,7 +1139,7 @@ public class GroupsResourceImpl implements GroupsResource {
                 SearchFilter.ofArtifactId(artifactId));
         VersionSearchResultsDto resultsDto = storage.searchVersions(filters, OrderBy.createdOn,
                 OrderDirection.asc, ParameterValidationUtils.normalizeOffset(offset),
-                ParameterValidationUtils.normalizeLimit(limit), false);
+                ParameterValidationUtils.normalizeLimitUnbounded(limit), false);
         return V2ApiUtil.dtoToSearchResults(resultsDto);
     }
 
--- a/app/src/main/java/io/apicurio/registry/rest/v2/impl/SearchResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v2/impl/SearchResourceImpl.java
@@ -12,6 +12,7 @@ import io.apicurio.registry.content.util.ContentTypeUtil;
 import io.apicurio.registry.logging.Logged;
 import io.apicurio.registry.metrics.health.liveness.ResponseErrorLivenessCheck;
 import io.apicurio.registry.metrics.health.readiness.ResponseTimeoutReadinessCheck;
+import io.apicurio.registry.rest.ParameterValidationUtils;
 import io.apicurio.registry.rest.v2.beans.ArtifactSearchResults;
 import io.apicurio.registry.rest.v2.beans.SortBy;
 import io.apicurio.registry.rest.v2.beans.SortOrder;
@@ -126,8 +127,9 @@ public class SearchResourceImpl implements SearchResource {
             filters.add(SearchFilter.ofContentId(contentId));
         }
 
-        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir, offset.intValue(),
-                limit.intValue(), false);
+        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir,
+                ParameterValidationUtils.normalizeOffset(offset), ParameterValidationUtils.normalizeLimit(limit),
+                false);
         return V2ApiUtil.dtoToSearchResults(results);
     }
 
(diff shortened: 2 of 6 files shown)


app/src/main/java/io/apicurio/registry/rest/v2/impl/SearchResourceImpl.java
```diff
@@ -126,8 +127,9 @@ public ArtifactSearchResults searchArtifacts(String name, BigInteger offset, Big
             filters.add(SearchFilter.ofContentId(contentId));
         }
 
-        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir, offset.intValue(),
-                limit.intValue(), false);
+        ArtifactSearchResultsDto results = storage.searchArtifacts(filters, oBy, oDir,
+                ParameterValidationUtils.normalizeOffset(offset), ParameterValidationUtils.normalizeLimit(limit),
```
paoloantinori: Adopting normalizeLimit here newly caps limit at MAX_LIMIT = 1000 for this v2 endpoint. Before this change limit.intValue() was forwarded uncapped, so a request like limit=2000 now silently returns at most 1000 rows. That is a behavior change on a stable v2 API, not only the 500 fix. Please confirm this is intended, that the storage layer does not already enforce a different cap that this overrides, and that no v2 client depends on larger pages. If intended, call it out in the PR description so it is not mistaken for a regression.

app/src/main/java/io/apicurio/registry/rest/v2/impl/GroupsResourceImpl.java
```diff
@@ -1136,7 +1137,8 @@ public VersionSearchResults listArtifactVersions(String groupId, String artifact
         Set<SearchFilter> filters = Set.of(SearchFilter.ofGroupId(defaultGroupIdToNull(groupId)),
                 SearchFilter.ofArtifactId(artifactId));
         VersionSearchResultsDto resultsDto = storage.searchVersions(filters, OrderBy.createdOn,
-                OrderDirection.asc, offset.intValue(), limit.intValue(), false);
+                OrderDirection.asc, ParameterValidationUtils.normalizeOffset(offset),
```
paoloantinori, line 1141: This is the fifth v2 endpoint migrated. Before merge, grep the v2 impl package for any remaining offset.intValue()/limit.intValue() (and other BigInteger pagination params) to confirm no listing or search site was missed. A missed call site would still forward raw values to storage and leave the 500 vector open, so completeness of the migration is the thing to verify here.

paoloantinori (commented):
The extraction into ParameterValidationUtils is a clean, behavior-preserving refactor for v3 and it closes the 500-on-invalid-pagination gap for v2. Two points need attention before merge, both because this is more than a pure bug fix on the v2 side: it also imports v3's pagination contract, including a page-size cap v2 did not have.

- New 1000-row cap on v2 limit (behavioral change). The v2 endpoints previously forwarded limit.intValue() to storage with no REST-level cap (GroupsResourceImpl.java:403-404, SearchResourceImpl.java:130-131). normalizeLimit now caps at MAX_LIMIT = 1000 (ParameterValidationUtils.java:81), so a v2 client requesting limit=2000 silently receives at most 1000 rows. That is a new constraint on a stable v2 API, not just the 500 fix. Confirm the storage layer does not already enforce a different cap that this overrides, confirm no v2 consumer relies on pages larger than 1000, and if intended, state it explicitly in the PR description so it is not read as a regression.

- Confirm every v2 call site was migrated. The PR lists five endpoints. Grep the v2 impl package for .intValue() on offset/limit (and any other BigInteger pagination params) to be sure no listing/search site was missed, since a missed site leaves the 500 vector open. The v3 side looks fully covered: all six SearchResourceImpl sites were updated and the two private helpers plus their constants were removed.

- Silent clamp vs. 400 for invalid input. Negative and overflow values are normalized instead of rejected (ParameterValidationUtils.java:68, :81), so offset=-1 returns 200 with the first page rather than a client error. This matches v3 exactly, so it is consistent, but it does mask client mistakes. Worth a conscious call that this is the desired v2 contract rather than a 400 for clearly invalid values.

- Null reachability. normalizeOffset/normalizeLimit throw NPE on a null argument (offset.max(...), limit.signum()). The tests show omitted params fall back to defaults (0/20), which implies @DefaultValue guards make null unreachable, matching v3. A quick confirmation that every v2 method carries those defaults would prevent a future interface edit from reintroducing a 500 via NPE.

- Readability (easy win). After the move, every v3 call site spells out ParameterValidationUtils.normalizeOffset(...)/normalizeLimit(...) (SearchResourceImpl.java:140 and five more). A static import of the two methods would restore the brevity the old private helpers had. Minor.

- Tests. Coverage is thorough: negative, overflow, omitted, boundary, limit=0, and a 404-before-normalization check (PaginationV2Test.java). The listGroups test correctly avoids absolute counts because of shared global state, and its relative/clamped assertions hold regardless of other groups in the system.

Adii-45:
@paoloantinori Thanks for the review!

I've applied the requested changes:
- Preserved v2's existing pagination semantics while fixing the invalid/overflow pagination issue
- Re-verified all affected v2 endpoints to ensure all relevant pagination call sites are covered
- Updated the tests and clarified the intended behavior in the PR description

Whenever you have some time, I'd appreciate another look. Please let me know if there's anything else you'd like me to change. Thanks!

paoloantinori:
Re-read at 7ca77806. The normalizeLimitUnbounded addition preserves v2's existing unbounded-limit semantics while closing the overflow-to-500 gap, all five v2 call sites are migrated away from raw .intValue(), and ParameterValidationUtilsTest makes the v2-vs-v3 split explicit. My earlier points are addressed; no further concerns from me. The static-import readability nit is optional and not worth a respin.

Adii-45:
Thanks for the review, @paoloantinori! I really appreciate you taking another look and confirming the changes...

@EricWittmann The review feedback has now been addressed, all requested changes have been applied, and all required checks are passing. The PR should now be ready to continue with the `/accept` command whenever you have a chance!

Please let me know if there's anything else you'd like me to update. Thanks!

paoloantinori:
Thanks, Adii. Nothing further from my side at this head — the points from the earlier review are addressed. Over to @EricWittmann for the `/accept` call.

Adii-45:
Hey @EricWittmann, can you please take a look at this PR whenever you get a chance? Thanks!

EricWittmann (commented):
## Code Review

Thanks for the contribution, @Adii-45! This is a well-structured fix with thorough test coverage. A few items to address before we can merge.

### What's good

- **Smart `normalizeLimit` vs `normalizeLimitUnbounded` split** — v2 never had a page-size cap, so introducing one silently would be a breaking change. The separate method preserves v2 behavior while fixing the actual bug. Good call, and nice response to the initial review feedback.
- **Clean v3 refactoring** — extracting the private methods into the shared utility is purely mechanical and behavior-preserving.
- **Comprehensive tests** — unit tests for all three utility methods, plus integration tests covering all 5 affected endpoints with negative, overflow, zero, boundary, and omitted values. The `testListGroupsLimitAndOffsetEdgeCases` approach of using relative assertions (since `listGroups` has no group filter) is particularly thoughtful.
- **Completeness** — all 5 raw `.intValue()` call sites across both v2 impl files are covered. No pagination sites were missed.

### Issues to fix

1. **Missing `Content-Type` on POST requests in `PaginationV2Test`**

   In `testSearchArtifactsByContentLimitAndOffsetEdgeCases`, all POST requests use `.body(artifactContent)` without setting a content type:
   ```java
   given().when().body(artifactContent).queryParam("offset", -1)
           .post("/registry/v2/search/artifacts")
   ```
   
   The existing v2 and v3 test patterns always set `.contentType(CT_JSON)` or `.contentType(ContentTypes.APPLICATION_JSON)` on POST requests with a body (see `SearchVersionsTest` lines 152-177 for the v3 equivalent). Please add the content type to these requests.

2. **Null documentation on normalize methods**

   The `normalizeOffset` and `normalizeLimit`/`normalizeLimitUnbounded` methods will NPE if called with a null `BigInteger`. This is currently safe because all callers perform null checks before calling them, but since these are now `public` utility methods (previously `private` in SearchResourceImpl), it would be good to add `@param` Javadoc noting that `null` is not accepted, or add a brief precondition. This prevents a future caller from hitting a confusing NPE.

### Minor / optional

- The method name `normalizeLimitUnbounded` is clear but slightly verbose. An alternative like `normalizeLimitUncapped` might read a bit better, but this is purely stylistic — feel free to keep as-is.

Adii-45:
Thanks for the detailed review, @EricWittmann! I've addressed both requested items in the latest commit:

- Added the appropriate Content-Type: CT_JSON to the POST requests in PaginationV2Test
- Added Javadoc documenting the non-null requirement for normalizeLimit and normalizeLimitUnbounded

I also ran the relevant tests and checkstyle - everything passes. I left the normalizeLimitUnbounded naming unchanged as suggested, since the rename was optional..!

Please take another look when you get a chance.. Thanks!

EricWittmann (commented):
Thanks for the contribution, @Adii-45! Clean fix with thorough test coverage, and the `normalizeLimitUnbounded` split to preserve v2 semantics was a nice touch. Merging now.

PR #9371, opened 2026-08-07
fix(build): declare inter-module test dependencies as test-jar

### LINKED ISSUE #9362: fix(ci): investigate test-jar race condition in parallel Maven builds
## Problem

Parallel Maven builds (`-T 0.5C`) intermittently fail with compilation errors in `schema-util/json` because `schema-util/common`'s `test-jar` artifact is not yet available when `schema-util/json`'s `testCompile` phase runs.

The failure looks like this:

```
[ERROR] schema-util/json/src/test/java/.../JsonSchemaSmokeTest.java:[5,48] cannot find symbol
[ERROR]   symbol:   class CompatibilityTestExecutor
[ERROR]   location: package io.apicurio.registry.rules.compatibility
```

`CompatibilityTestExecutor` lives in `schema-util/common/src/test/` and is consumed by `schema-util/json` via a `<type>test-jar</type>` dependency. The `test-jar` goal in `schema-util/common/pom.xml` (line ~108) produces this artifact, but depending on thread scheduling, `schema-util/json` may attempt `testCompile` before it's ready.

**Observed:** the same commit passes the `Build Java Application` job on `main` but fails on PR #9286's branch — identical code, identical Maven command, different thread scheduling.

## What needs investigation

This is not just a one-off fix. The `test-jar` race in `schema-util/common` → `schema-util/json` is the instance we caught, but the same pattern could exist elsewhere in the reactor. A holistic audit is needed:

1. **Identify all `test-jar` cross-module dependencies** — grep for `<type>test-jar</type>` across all `pom.xml` files and verify the reactor can resolve them before consumers compile.
2. **Check `test-jar` goal binding** — the `maven-jar-plugin`'s `test-jar` goal should be bound to the `package` phase (the default) so the reactor graph treats it as a proper output. If it's bound to a later or non-standard phase, the dependency resolver may not wait for it.
3. **Audit for other inter-module test dependencies** — any module whose test code imports from another module's test sources has the same risk. These should all use explicit `test-jar` dependencies with correct reactor ordering.
4. **Consider whether `-DskipTests` masks ordering issues** — `-DskipTests` skips execution but still compiles tests. If the intent is to skip test compilation too (e.g., in the `Build Java Application (no tests)` CI job), `-Dmaven.test.skip=true` would avoid the problem entirely for that job. But that's a workaround, not a fix — the reactor ordering should be correct regardless.

## Goal

Concurrent builds (`-T 0.5C` or `-T 1C`) should be reliable without depending on lucky thread scheduling. All inter-module test dependencies should be properly declared so the Maven reactor resolves them in the correct order.

Akshats-git:
Fixes #9362

## What

Audits every cross-module test dependency in the reactor, fixes the two that were declared in a fragile way, and adds a check so they cannot regress.

## Why

Sharing test classes between modules only orders those modules in the Maven reactor if the consumer declares a test-jar dependency. Importing a class from another module's `src/test/java` does not create a reactor edge by itself. When the declaration is missing, a parallel build can reach `test-compile` before the producer has attached its test-jar.

One note on the issue as filed. The reported case, `schema-util/json` racing `schema-util/common`, does not reproduce. That edge is already declared in `schema-util/json/pom.xml`, and the reactor orders the two modules correctly. I confirmed this with the reactor summary and three consecutive `-T 1C` builds of that subtree, all green. So I audited the rest of the reactor instead, matching every module's test-source usage against its declared dependencies. Imports and same-package references were both checked, since same-package usage needs no import at all.

Two real problems turned up.

## Changes

**`operator/olm-tests`** consumed `operator/controller`'s test classes through `<classifier>tests</classifier>` on a plain jar dependency rather than `<type>test-jar</type>`. This resolves today only because the test-jar handler happens to map to extension `jar` plus classifier `tests`. It is fragile for two reasons: `dependencyManagement` keys differ between the two spellings, and the classifier form is invisible to the `<includeTypes>test-jar</includeTypes>` filter that `integration-tests` unpacks with. The managed entry in the root pom had to move with it, otherwise the canonical form loses its version.

**`distro/connect-converter`** bound the `maven-jar-plugin` test-jar goal to a phase named `non-existant`. The module has no test sources and inherits no test-jar binding, so this was dead configuration. It also made the module a false positive in any audit that greps for test-jar producers, and Maven 4 rejects unknown phases. The neighbouring `maven-source-plugin` overrides use the same trick but do disable real inherited executions, so they stay.

Everything else was clean. All 20 test-jar producers bind to the default `package` phase, and no other module uses another module's test sources.

**`scripts/validate-test-jar-dependencies.py`** keeps the audit repeatable. It is run from `validate-files.sh` in the existing `lint-and-validate` job, so it gates `build`. It enforces two invariants: a module whose test sources use another module's test classes must declare a test-jar dependency on it, and every module named by such a dependency must actually produce a test-jar at the `package` phase. It runs in 0.3s, needs no build, and uses only the standard library.

## Testing

- The checker passes on the fixed tree and fails with exit 1 on four seeded violations: dependency removed from `schema-util/json`, removed from `olm-tests`, producer rebound to `verify`, producer's test-jar goal removed. Each failure prints a copy-pasteable dependency block.
- `./scripts/validate-files.sh` exits 0 clean and exits 1 when the check fails.
- `mvn validate -Dfull -pl operator/olm-tests` resolves, which confirms the managed version still binds after the switch to `<type>test-jar</type>`.
- `mvn package -Dfull -T 1C -pl operator/olm-tests -am` succeeds. The controller's `-tests.jar` is built before `olm-tests`, whose 7 test classes compile against it.
- `mvn package -Pprod -Dfull -T 1C -pl distro/connect-converter -am` succeeds across 18 modules, and no `-tests.jar` is produced, so the removed execution's intent is preserved.

## Not done

The issue asks whether the `Build Java Application (no tests)` job should use `-Dmaven.test.skip=true`. I left it alone. That flag also suppresses test-jar creation, and the job's `-Dfull` reactor includes modules whose consumers unpack test-jars. The issue itself calls it a workaround, and changing CI job semantics to avoid a problem that turned out not to exist there is not worth the risk.

Separately, and out of scope here: that job passes `-Daether.syncContext.named.factory=noop`, which disables local-repository locking entirely. It was added to stop resolver deadlocks under `-T`, but `file-lock` would do that without giving up safety. Worth a look in its own issue.


scripts/validate-test-jar-dependencies.py
```diff
@@ -0,0 +1,291 @@
...
+
+    Both spellings resolve to the same artifact and both create the reactor
+    edge, so both are accepted: <type>test-jar</type>, and the older
+    <classifier>tests</classifier> on a plain jar dependency.
+    """
+    consumed = set()
+    for dependency in declared_dependencies(project):
+        artifact_type = dependency.findtext(POM_NS + "type")
+        classifier = dependency.findtext(POM_NS + "classifier")
+        if artifact_type == "test-jar" or (artifact_type is None and classifier == "tests"):
```
paoloantinori: This branch accepts `<classifier>tests</classifier>`, which is the spelling this PR removes from `olm-tests` because it is fragile (different `dependencyManagement` key, invisible to the `includeTypes>test-jar</includeTypes>` unpack filter). The check enforces the reactor-ordering invariant, but it will not stop a contributor re-adding the classifier form later, so that exact fragility can return without the gate noticing. Consider failing on the classifier form, or at least emitting a warning that points at `<type>test-jar</type>`.

scripts/validate-test-jar-dependencies.py
```diff
@@ -0,0 +1,291 @@
...
+        for package in visible_packages:
+            candidates.update(by_package.get(package, ()))
+
+        for candidate in candidates:
+            if candidate not in foreign_test_types:
+                continue
+            simple_name = simple_name_of(candidate)
+            if simple_name in own_simple_names:
+                continue
+            if re.search(r"\b%s\b" % re.escape(simple_name), source):
```
paoloantinori, line 247: Matching the foreign type's simple name with `re.search(r"\b%s\b" ...)` over the whole source also hits comments, string literals, unused imports, and unrelated types that share the name. For explicit imports the match is redundant (the import line already matches), so the real risk is the same-package branch: a coincidental mention of a foreign type's simple name in a comment can flag a phantom dependency and fail the build gate on an unrelated PR. Strip comments and strings before matching, or restrict the same-package match to identifier tokens in a type-use position.

scripts/validate-test-jar-dependencies.py
```diff
@@ -0,0 +1,291 @@
...
+    """Phases the maven-jar-plugin test-jar goal is bound to, if any.
+
+    An empty string stands for "no explicit phase", which is the goal's own
+    default of package.
+    """
+    phases = []
+    build = project.find(POM_NS + "build")
+    if build is None:
+        return phases
+    plugins = build.find(POM_NS + "plugins")
```
paoloantinori, line 148: Producer validation reads only the producer's own `<build><plugins>`. A test-jar execution inherited from a parent pom's `<build><plugins>`, or declared in `<pluginManagement>`, is not seen here, so a producer that relies on inheritance is falsely reported as producing no test-jar, and a phase regression introduced via inheritance is invisible. The current tree is safe because all producers declare the binding locally, but stating that assumption in the docstring would make the limitation explicit for the next person who refactors the plugin wiring.

scripts/validate-test-jar-dependencies.py
```diff
@@ -0,0 +1,291 @@
+#!/usr/bin/env python3
```
paoloantinori, line 1: The new file has no Apache 2.0 license header. Add the standard header to match the rest of the repo and to pass the license check if its includes cover `*.py`.

paoloantinori (commented):
The two POM fixes are correct and necessary, and the new checker enforces real reactor-ordering invariants; the checker does not cover the specific fragility this PR fixes, and its source matching can produce false positives that now gate the build.

- `operator/olm-tests/pom.xml:30` and `pom.xml:595`: switching `<classifier>tests</classifier>` to `<type>test-jar</type>` is right, and the `dependencyManagement` entry had to move with it. The management key changes from `{g, a, jar, tests}` to `{g, a, test-jar, ""}`, so without the root pom change the consumer's managed version would detach and the build would fail. `mvn validate -pl operator/olm-tests` resolving confirms it still binds.
- `distro/connect-converter/pom.xml:89`: removing the `non-existant`-phase override is safe only because the module inherits no test-jar binding. The build confirms no `-tests.jar` is produced, and the dead phase would break under Maven 4 anyway. Fine, but the safety rests on a fact (no inherited binding) that the new checker does not assert.
- `scripts/validate-test-jar-dependencies.py:84`: the checker deliberately accepts the older `<classifier>tests</classifier>` spelling, which is exactly the form this PR removes from `olm-tests` as fragile. It will catch a missing declaration or a wrong-phase producer, but it will not catch a contributor re-adding `<classifier>tests</classifier>`, so the `dependencyManagement`-key and `includeTypes` fragility can regress unnoticed.
- `scripts/validate-test-jar-dependencies.py:190`: matching a type's simple name with a word-boundary regex over the whole source file also matches comments, string literals, unused imports, and unrelated types that happen to share a name. Explicit imports already prove use, so the exposure is the same-package branch, where a coincidental mention in a comment can flag a phantom dependency and fail the build gate on an unrelated PR.
- `scripts/validate-test-jar-dependencies.py:99`: producer validation reads only the producer's own `<build><plugins>`. A test-jar binding inherited from a parent `<build><plugins>` or declared in `<pluginManagement>` is invisible, so a producer relying on inheritance would be falsely flagged, and a phase regression introduced via inheritance would not be seen.
- `scripts/validate-test-jar-dependencies.py:1`: the new script ships without an Apache 2.0 license header. Add one to match repo convention and pass the license check if it covers Python files.

distro/connect-converter/pom.xml
```diff
@@ -89,20 +89,6 @@
...
-          <execution>
-            <!-- We don't want to create a test JAR -->
-            <id>test-jar</id>
-            <goals>
-              <goal>test-jar</goal>
-            </goals>
-            <phase>non-existant</phase>
-          </execution>
-        </executions>
-      </plugin>
```
malladinagarjuna2, line 105: The Maven-4 argument doesn't survive the diff: two more non-existant phases stay in this file at lines 77 and 81 (maven-source-plugin). If unknown phases break Maven 4, they break there too.
Separately, the removal is correct but for a simpler reason than the PR gives — the root pom has maven-jar-plugin only in <pluginManagement> (pom.xml:1077-1080, version element, no executions), so nothing was inherited here to cancel. This execution declared a goal and bound it to a phase that doesn't exist: self-cancelling, not overriding. Worth stating that way, since it also means the change is behavior-neutral.

paoloantinori:
You're right on both points, and I should walk back the Maven 4 framing.

On the first: `maven-source-plugin` at lines 77 and 81 is bound to the same `non-existant` phase and is untouched by this PR. If unknown phases broke a future Maven, those would break too, so "would break under Maven 4 anyway" cannot be the reason this removal is safe. My earlier note was wrong on that.

On the second: confirmed — `maven-jar-plugin` appears in the root pom only in `<pluginManagement>` (pom.xml:1079) with just a `<version>`, no `<executions>`. So there was no inherited test-jar binding for the removed execution to override. It declared the `test-jar` goal and bound it to a phase that doesn't exist, which is self-cancelling. The removal is behavior-neutral, and that's the cleaner reason.

The other points from my review (the checker not covering the `<classifier>tests</classifier>` regression it itself relies on, the same-package word-boundary false-positive surface, and the missing license header) stand independently of this.

malladinagarjuna2:
hey thanks for the discussion @paoloantinori.  
agreed with all the points you mentioned 
regarding the maven one yeah i locally tested it, and thus i found that bug.

paoloantinori:
Thanks for confirming the local test — that's the part I couldn't verify from a read-through alone, so good to know it reproduces on your side.

Sounds like we're aligned on the shape of the PR. The two POM fixes are solid, and once the checker is tightened up to actually cover the `classifier`-on-plain-jar fragility it's meant to prevent, this will be a solid guardrail going forward.

Akshats-git:
Thanks, all three checker points are addressed.

**Classifier form now fails.** It still counts as a reactor edge, so it reports once as a spelling problem rather than also as a missing dependency.

**False positive fixed.** Comments and string literals are stripped before any identifier matching. Explicit imports are now accepted without a textual match, since an imported type has to resolve at compile time whether the body uses it or not, so only the same-package and wildcard branch matches by name.

Worth noting this was live, not theoretical. Five packages in the reactor are contributed to by more than one module. `schema-util/xsd` has a test class in the same package as common's `CompatibilityTestExecutor` and declares no test-jar dep on it, so one javadoc line naming that type fails the build on the old checker and passes on the new one.

**Inheritance limitation documented** in `test_jar_execution_phases`: only the module's own `<build><plugins>` is read, every producer currently declares the binding locally, and resolving inheritance properly means building the effective pom.

On the license header, I left it out. No file under `scripts/` has one, and `License Compliance` is not a file-header check (no workflow in the repo, and it already passes). CLAUDE.md lists per-file headers as not checked, with the root LICENSE governing. Happy to add it if you would rather, just flagging it would be the only file in that directory with one.

paoloantinori:
All three points landed cleanly on this head:

- the `<classifier>tests</classifier>` spelling is now rejected with an actionable message (lines 311-322), so the `dependencyManagement`/`includeTypes` fragility can't regress unnoticed;
- comments and string literals are stripped before identifier matching (lines 64-73, applied at 220), which closes the false-positive surface I was worried about. Accepting explicit imports without a textual match is the right call — a type that imports has to resolve at compile time regardless of whether the body names it, so only the same-package/wildcard branch needs name matching;
- the inheritance limitation is documented in `test_jar_execution_phases`, and the `schema-util/xsd` case is a concrete example of why the same-package branch has to stay.

On the license header: agreed, leaving it out is consistent with the rest of `scripts/`. No further concerns from me.

Akshats-git:
Hello @paoloantinori ! I just wanted to get an update whether this PR is ready for merge or some more changes are needed to be made. I was planning on opening a PR for another issue but can't do so until this one is merged.

paoloantinori:
All three points I raised on the checker landed cleanly on this head, and the two POM fixes (`olm-tests` classifier-to-type, `connect-converter` dropped test-jar suppression) are correct. From a technical review standpoint the PR is done on my side — no further changes are needed.

The `mergeable_state` is showing "blocked", which usually points at a pending status check or required-review gate rather than a code problem. The final merge call is with the maintainers, but I have nothing more to review here.

Akshats-git:
Hey @jsenko ! Waiting for an update on this PR. I was planning to work on another issue but won't be able to open a PR for that because of the 1 PR limit.

PR #9375, opened 2026-08-07
fix: prevent render endpoint from bypassing disabled version content …fix: prevent renderPromptTemplate from bypassing disabled-version content withdrawal

### LINKED ISSUE #9374: Disabled prompt template versions can still be rendered via /render, bypassing content withdrawal
## Summary
`POST /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{version}/render`
serves the full rendered content of a **DISABLED** artifact version, even though the
platform's normal content-read path (`GET .../versions/{version}/content`) correctly
blocks access to disabled versions with a 404.

Disabling a version is the mechanism admins use to withdraw its content (e.g. a
misconfigured template, a revoked version with sensitive data, a bad rollout). The
render endpoint currently defeats that withdrawal for PROMPT_TEMPLATE artifacts.

## Impact
Any caller with read access to the artifact (same auth level as the content
endpoint) can retrieve a disabled version's template text in full, with variables
substituted, via the render endpoint — even though direct content access is blocked.
Confidentiality/access-control consistency gap. Not RCE, not data corruption.

## Root cause
`renderPromptTemplate` (`GroupsResourceImpl.java`) resolves branch/tip expressions
with `RetrievalBehavior.ALL_STATES` (no state filtering) and never checks the
resolved version's state before rendering  unlike `getArtifactVersionContent`,
which resolves with `SKIP_DISABLED_LATEST` and explicitly rejects `DISABLED`
versions with a `VersionNotFoundException`.

## Steps to reproduce
1. Create a `PROMPT_TEMPLATE` artifact with at least one version.
2. Disable that version:
PUT /apis/registry/v3/groups/default/artifacts/{artifactId}/versions/{version}/state
{ "state": "DISABLED" }

3. Confirm content is blocked (expected, and confirmed working):

GET /apis/registry/v3/groups/default/artifacts/{artifactId}/versions/{version}/content
→ 404 VersionNotFoundException

4. Attempt to render the same disabled version:

{ "variables": { ... } }
**Actual (before fix):** `200 OK` with the disabled version's rendered content.
   **Expected:** `404 VersionNotFoundException`, matching the content endpoint.

https://github.com/user-attachments/assets/bdfdc0b9-b6e6-4c25-8583-ab2cbf231fe9

## Suggested fix
Mirror `getArtifactVersionContent`'s guard in `renderPromptTemplate`:
- Resolve branch/tip expressions with `RetrievalBehavior.SKIP_DISABLED_LATEST`
  instead of `ALL_STATES`.
- After fetching version metadata, reject `VersionState.DISABLED` with the same
  `VersionNotFoundException` the content endpoint throws.

Faisal77666:
Fixes #9374

## What this PR does
Closes an access-control consistency gap where `renderPromptTemplate` served the
full rendered content of DISABLED prompt template versions, even though the
platform's normal content-read path (`getArtifactVersionContent`) already blocks
disabled-version content with a 404. See linked issue for full details and repro.

Also fixes an unrelated bug found while verifying the above through the UI: the
version page's `is404()` check only recognized a 404 when the caught error was a
JSON **string** (`JSON.parse(e)`), but the SDK's generated client throws structured
error objects. As a result, `is404()` always returned `false` for a real 404, and
*any* disabled version's page fell into the hard-failure path  showing a full-page
"Resource not found" instead of the intended graceful degradation (Content tab
shows "not available," other tabs stay usable).

## Root cause
- **Backend**: `renderPromptTemplate` resolved branch/tip expressions with
  `RetrievalBehavior.ALL_STATES` (no state filtering) and never checked the
  resolved version's state before rendering.
- **Frontend**: `is404()` in `VersionPage.tsx` used `JSON.parse(e)` string
  matching, incompatible with the SDK's actual error shape.

## Changes
- `app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java`
  - `renderPromptTemplate` now resolves with `RetrievalBehavior.SKIP_DISABLED_LATEST`
  - Added a `DISABLED` state check that throws `VersionNotFoundException`,
    mirroring `getArtifactVersionContent`'s existing guard exactly
- `ui/ui-app/src/app/pages/version/VersionPage.tsx`
  - Replaced the broken local `is404()` with the shared `isErrorStatus(e, 404)`
    helper (already used correctly elsewhere, e.g. `PageErrorHandler.tsx`)

## Testing done
- **Existing test suite**: `VersionStateTest` (the test that encodes the
  disable-content contract) and `PromptRenderingServiceTest`  35/35 passing,
  no regressions.
- **UI build**: `tsc && vite build`  clean.
- **Live end-to-end repro** on a rebuilt local server (before → after):
before fix

GET .../versions/2.0.0/content → 404
POST .../versions/2.0.0/render → 200 (rendered content leaked)

after fix

GET .../versions/2.0.0/content → 404
POST .../versions/2.0.0/render → 404 VersionNotFoundException

## Before

[raw-recording (12).webm](https://github.com/user-attachments/assets/79eaa7a2-aa17-4efb-aea6-9d4e6281ffc7)

## After

[raw-recording (13).webm](https://github.com/user-attachments/assets/d01995b1-5955-4fef-9796-57ca23237074)

## Notes for reviewers
Severity: medium (confidentiality/access-control consistency gap, not RCE or
data corruption). Flagging as security-relevant for triage since it's a content-
withdrawal bypass.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
@@ -1979,11 +1979,17 @@ public class GroupsResourceImpl extends AbstractResourceImpl implements GroupsRe
         ParameterValidationUtils.requireParameter("variables", data.getVariables());
 
         var gav = VersionExpressionParser.parse(new GA(groupId, artifactId), versionExpression,
-                (ga, branchId) -> storage.getBranchTip(ga, branchId, RetrievalBehavior.ALL_STATES));
+                (ga, branchId) -> storage.getBranchTip(ga, branchId, RetrievalBehavior.SKIP_DISABLED_LATEST));
 
         // Verify the artifact exists and is of type PROMPT_TEMPLATE
         ArtifactVersionMetaDataDto versionMetaData = storage.getArtifactVersionMetaData(
                 gav.getRawGroupIdWithNull(), gav.getRawArtifactId(), gav.getRawVersionId());
+
+        // A disabled version's content is withdrawn from every other content-serving read path
+        // (see getArtifactVersionContent); rendering must honor that withdrawal too.
+        if (VersionState.DISABLED.equals(versionMetaData.getState())) {
+            throw new VersionNotFoundException(groupId, artifactId, versionExpression);
+        }
 
         String artifactType = versionMetaData.getArtifactType();
         if (!"PROMPT_TEMPLATE".equals(artifactType)) {
--- a/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java
@@ -1985,9 +1985,7 @@ public class GroupsResourceImpl extends AbstractResourceImpl implements GroupsRe
         ArtifactVersionMetaDataDto versionMetaData = storage.getArtifactVersionMetaData(
                 gav.getRawGroupIdWithNull(), gav.getRawArtifactId(), gav.getRawVersionId());
 
-        // A disabled version's content is withdrawn from every other content-serving read path
-        // (see getArtifactVersionContent); rendering must honor that withdrawal too.
-        if (VersionState.DISABLED.equals(versionMetaData.getState())) {
+        if (versionMetaData.getState() == VersionState.DISABLED) {
             throw new VersionNotFoundException(groupId, artifactId, versionExpression);
         }
 
@@ -2030,6 +2028,10 @@ public class GroupsResourceImpl extends AbstractResourceImpl implements GroupsRe
         // Verify the artifact exists and is of type PROTOBUF
         ArtifactVersionMetaDataDto versionMetaData = storage.getArtifactVersionMetaData(
                 gav.getRawGroupIdWithNull(), gav.getRawArtifactId(), gav.getRawVersionId());
+
+        if (versionMetaData.getState() == VersionState.DISABLED) {
+            throw new VersionNotFoundException(groupId, artifactId, versionExpression);
+        }
 
         String artifactType = versionMetaData.getArtifactType();
         if (!"PROTOBUF".equals(artifactType)) {
--- a/app/src/test/java/io/apicurio/registry/noprofile/VersionStateTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/VersionStateTest.java
@@ -5,6 +5,8 @@ import io.apicurio.registry.rest.client.models.EditableVersionMetaData;
 import io.apicurio.registry.rest.client.models.VersionMetaData;
 import io.apicurio.registry.rest.client.models.VersionState;
 import io.apicurio.registry.rest.client.models.WrappedVersionState;
+import io.apicurio.registry.rest.client.models.RenderPromptRequest;
+import io.apicurio.registry.rest.client.models.RenderPromptRequestVariables;
 import io.apicurio.registry.types.ArtifactType;
 import io.apicurio.registry.types.ContentTypes;
 import io.quarkus.test.junit.QuarkusTest;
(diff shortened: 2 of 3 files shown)


paoloantinori (commented):
The review for **PR #9375** is complete and was delivered in my previous turn as the required fenced JSON. No further work is needed on it.

These incoming task notifications (the latest being "Broad correctness finder on PR9368") are all from the `code-review` skill's forked execution, which mis-targeted an unrelated PR (#9368, an MCP `JsonSchemaDiffLibrary` refactor) and spawned background agents against `/tmp/pr9368.patch`. They are not about PR #9375, and I have not incorporated any of their findings into the review above. They can be safely disregarded, including the stray plan file that agent wrote to `/config/plans/review-the-pr-below-async-valley-agent-a7bb7193e27049f53.md`.

To recap what the delivered review covers for PR #9375 (verified against the real source at `/tmp/apicurio-registry`):
- Backend fix is correct and mirrors `getArtifactVersionContent` exactly (same `SKIP_DISABLED_LATEST` branch resolution, same `VersionNotFoundException(groupId, artifactId, versionExpression)` throw); DRAFT/DEPRECATED templates still render.
- Frontend fix is correct; `isErrorStatus` handles the SDK's structured error shape the old `JSON.parse` path never matched.
- Two MAJOR items raised: (1) no regression test for the new DISABLED render guard, and (2) sibling `exportArtifactVersion` (L2016) appears to share the same bypass via an explicit disabled version.

If a new PR or follow-up is needed, send it and I'll review it.

Faisal77666:
@paoloantinori Thanks for the thorough review! Addressed both MAJOR items:
1. **Regression tests added**: `testRenderRejectsDisabledVersion` and `testExportRejectsDisabledVersion` in `VersionStateTest.java`, both verifying a 404/VersionNotFoundException on a DISABLED version.
2. **`exportArtifactVersion` bypass confirmed and fixed**: it had the exact same gap  resolved the artifact/version metadata but never checked `getState()` before exporting. Added the same `VersionState.DISABLED` guard used in `renderPromptTemplate` and `getArtifactVersionContent`.
Both new tests pass alongside the existing suite.

paoloantinori:
Thanks for addressing both points. Confirmed landed at this head:

- The explicit `VersionState.DISABLED` guard is now on both `renderPromptTemplate` and `exportArtifactVersion`, throwing `VersionNotFoundException` (404) before any content is served or exported. The render path additionally switched `RetrievalBehavior.ALL_STATES` → `SKIP_DISABLED_LATEST`, which is the right belt-and-suspenders shape.
- The two new tests assert the positive evidence (`assertEquals(404, ...)` + `"VersionNotFoundException"`), not a weak negative, so a future regression that drops the guard would actually fail them. Good.
- The `is404()` refactor to the shared `isErrorStatus(e, 404)` helper fixes the structural-error-object mismatch cleanly.

One minor observation, not a blocker: `exportArtifactVersion` got the explicit state guard but its retrieval behavior was left unchanged (unlike render). The guard alone is sufficient to close the bypass, so this is fine — just flagging in case you want the two paths to be symmetric for consistency. No further concerns from me.

Faisal77666:
@paoloantinori Good catch  worth clarifying: `exportArtifactVersion` was already using `SKIP_DISABLED_LATEST` for branch-tip resolution in the original code, so that path was already correct. The explicit `VersionState.DISABLED` guard is what closes the actual gap on both endpoints  it covers the case where a caller passes a specific, already-disabled version directly (bypassing branch-tip resolution entirely), which `RetrievalBehavior` alone can't catch either way.
So the two paths are aligned in practice, just via slightly different-looking diffs. Thanks again for the thorough review throughout this PR!

paoloantinori:
You are right, and thanks for the precise clarification. The distinction is real: `SKIP_DISABLED_LATEST` only governs branch-tip resolution, so a caller passing a specific version reference bypasses that filter entirely. The explicit `VersionState.DISABLED` guard is what closes the gap on both endpoints for the direct-reference case. Good, the two paths are aligned. Nothing further from me.

EricWittmann:
Thanks for the contribution, @Faisal77666!

PR #9376, opened 2026-08-07
Fix invalid contract status error response

Hiten0305l:
## What

Improves error handling for invalid ContractStatusTransition.status values by introducing a dedicated JacksonJsonMappingExceptionMapper.

Previously, values such as {"status":"BOGUS"} failed during Jackson deserialization with a ValueInstantiationException, preventing the existing Apicurio exception-handling flow from producing a structured response.

The new mapper detects this specific enum deserialization failure, creates an InvalidParameterValueException containing both the parameter name and the invalid value, and routes it through Apicurio's exception-mapping infrastructure to return a structured ProblemDetails response with HTTP 400.

The existing MissingRequiredParameterException flow for missing or null status values remains unchanged.

The mapper remains scoped to ContractStatusTransition.Status. For unrelated Jackson deserialization errors, the fallback path continues to return a standardized HTTP 400 response while preserving the existing behavior outside the contract-status use case.

## Why
Previously, an invalid status resulted in a generic deserialization error that did not clearly tell the client which value was invalid or why.
Providing structured error details for both invalid and missing/null status values makes the API response more actionable and consistent, while preserving existing behavior for valid requests and unrelated APIs.

## Tests
- `DataContractsResourceTest`: 38 tests passed, 0 failures, 0 errors
- Checkstyle: 0 violations
- Verified missing status → structured `MissingRequiredParameterException` details
- Verified null status → structured `MissingRequiredParameterException` details
- Verified `BOGUS` → structured `InvalidParameterValueException` with `status` and `BOGUS` in the error details
- Verified valid `DRAFT`/`STABLE` transitions remain successful
- Verified unrelated invalid enum errors retain the existing RESTEasy response

## Reference
Follow-up to merged #9348.

Addresses the remaining invalid-value scenario related to #9329.


app/src/main/java/io/apicurio/registry/rest/JacksonJsonMappingExceptionMapper.java
```diff
@@ -0,0 +1,46 @@
...
+                    && exception.getCause().getMessage() != null) {
+                actualValue = exception.getCause().getMessage();
+            }
+            return coreMapper.mapException(new InvalidParameterValueException("status", "valid status enum value", actualValue));
+        }
+
+        // Delegate to the framework's default mapper for JsonProcessingException if available
+        if (providers != null) {
+            ExceptionMapper<com.fasterxml.jackson.core.JsonProcessingException> mapper =
+                providers.getExceptionMapper(com.fasterxml.jackson.core.JsonProcessingException.class);
```
paoloantinori: There is no `ExceptionMapper<JsonProcessingException>` anywhere in `app/src/main` (grep confirms), so this delegation depends entirely on a framework-provided mapper. If that mapper is absent, or `@Context Providers providers` is null, control falls through to the line-44 fallback that returns a plain-string 400 instead of the `ProblemDetails` body used by every other error path. That fallback is a silent behavior change for unrelated endpoints, and nothing in the diff exercises the delegation or fallback branch (the BOGUS test only hits the matched status branch). Add a test that posts an invalid enum value to an unrelated endpoint and asserts the existing response shape is preserved.

app/src/main/java/io/apicurio/registry/rest/JacksonJsonMappingExceptionMapper.java
```diff
@@ -0,0 +1,46 @@
...
+    @Context
+    Providers providers;
+
+    @Override
+    public Response toResponse(ValueInstantiationException exception) {
+        if (exception.getType() != null && ContractStatusTransition.Status.class.equals(exception.getType().getRawClass())) {
+            String actualValue = "unknown";
+            if (exception.getCause() instanceof IllegalArgumentException
+                    && exception.getCause().getMessage() != null) {
+                actualValue = exception.getCause().getMessage();
```
paoloantinori: `actualValue` is read from `getCause().getMessage()`. For an enum value failure this is typically the full `No enum constant io.apicurio...ContractStatusTransition.Status.BOGUS` produced by `Enum.valueOf`, which leaks the internal fully-qualified class name into the client-facing `detail` and renders awkwardly. Prefer extracting the offending input token directly from the exception (its token/value) rather than reusing the cause message. Worth confirming what the generated `Status` enum's cause message actually returns before shipping.

app/src/main/java/io/apicurio/registry/rest/JacksonJsonMappingExceptionMapper.java
```diff
@@ -0,0 +1,46 @@
...
+            }
+            return coreMapper.mapException(new InvalidParameterValueException("status", "valid status enum value", actualValue));
```
paoloantinori, line 55: This uses the constructor's intended `(parameter, expectedValue, actualValue)` order, which renders correctly. But the only other call site, `ParameterValidationUtils.requireParameterValue` (`ParameterValidationUtils.java:52`, exercised live at `GroupsResourceImpl.java:1397`), throws `new InvalidParameterValueException(parameterName, actualValue, expectedValue)` with the two values swapped, so it renders 'Expected <actual> but got <expected>' backwards. Now that two call sites exist with opposite orderings, reconcile them (fix the existing utility) so the error messages are consistent.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
```diff
@@ -582,7 +584,24 @@ public void testStatusTransition_NullStatus_Returns400() throws Exception {
...
+        createArtifact(GROUP, artifactId, ArtifactType.OPENAPI, content, ContentTypes.APPLICATION_JSON);
+
+        given().when().contentType(CT_JSON)
+                .pathParam("groupId", GROUP).pathParam("artifactId", artifactId)
+                .body("{\"status\": \"BOGUS\"}") // explicit bogus status
+                .post("/registry/v3/groups/{groupId}/artifacts/{artifactId}/contract/status")
+                .then().statusCode(400)
+                .body("name", equalTo("InvalidParameterValueException"))
+                .body("detail", containsString("status"))
+                .body("detail", containsString("BOGUS"));
```
paoloantinori: `containsString("BOGUS")` passes whether `detail` is the clean value 'BOGUS' or the full 'No enum constant ...Status.BOGUS' leak, so it cannot catch a regression in the message format. Once the value extraction in the mapper is finalized, assert the exact expected `detail` string (e.g. the full `Request contains parameter 'status' with invalid value.  Expected 'valid status enum value' but got 'BOGUS'`) so a leaked FQN or swapped ordering fails the test.

paoloantinori (commented):
Overall the fix correctly routes the invalid `ContractStatusTransition.Status` deserialization failure into the existing ProblemDetails flow for the matched case, but the new global `@Provider` ships an untested delegation path with a plain-string fallback that can regress unrelated endpoints, and the value extraction leaks internals. Notable points:

- `JacksonJsonMappingExceptionMapper.java:38-44`: the mapper is registered app-wide for every `ValueInstantiationException`, yet only the matched status branch is tested. The delegation via `providers.getExceptionMapper(JsonProcessingException.class)` and the line-44 fallback (`Response.status(400).entity("Not able to deserialize data provided.")`) are untested, and there is no app-level `JsonProcessingException` mapper (verified: none in `app/src/main`). If the framework mapper is absent or `providers` is null, unrelated invalid-value errors across the API return a plain string 400 instead of the ProblemDetails body used everywhere else. Add an automated test for an unrelated invalid enum, since the PR only claims manual verification for that case.
- `JacksonJsonMappingExceptionMapper.java:30`: `actualValue` is read from `getCause().getMessage()`. For a typical enum-value failure that message is the full `No enum constant <fully-qualified>.Status.BOGUS` from `Enum.valueOf`, so the internal class name is interpolated verbatim into the client-facing `detail`. Extract the raw offending token from the exception instead. Also, this only fires if the generated `Status` enum deserializes via a creator; a regeneration that switches to Jackson's default `InvalidFormatException` path would silently disable this mapper.
- `JacksonJsonMappingExceptionMapper.java:32`: the call uses the constructor's intended `(parameter, expectedValue, actualValue)` order, which is correct. The only other call site, `ParameterValidationUtils.requireParameterValue` (`ParameterValidationUtils.java:52`, used live at `GroupsResourceImpl.java:1397`), passes the values swapped, so it renders 'Expected <actual> but got <expected>' backwards. Two call sites now have opposite orderings, so this is a good moment to reconcile them.
- `DataContractsResourceTest.java:604`: `containsString("BOGUS")` passes whether the `detail` is the clean value or the full FQN leak, so it cannot catch a message-format regression. Assert the exact expected `detail` string once the value extraction is finalized.

Hiten0305l:
thanks @paoloantinori will address them now!

paoloantinori:
Sounds good — looking forward to the update. I'll re-read once the changes are pushed.

Hiten0305l:
Hi @paoloantinori , I have addressed the reported issues and updated the implementation accordingly. Could you please review the latest changes once again and let me know if any further adjustments are required?

app/src/main/java/io/apicurio/registry/rest/JacksonJsonMappingExceptionMapper.java
```diff
@@ -0,0 +1,55 @@
...
+                    actualValue = String.valueOf(ife.getValue());
+                }
+            }
+        }
+
+        if (isStatusEnum && actualValue != null) {
+            return coreMapper.mapException(new InvalidParameterValueException("status", "valid status enum value", actualValue));
+        }
+
+        return coreMapper.mapException(new jakarta.ws.rs.BadRequestException("Not able to deserialize data provided."));
```
paoloantinori: This fallback builds its own BadRequestException and routes it through coreMapper, so every JSON deserialization failure app-wide returns 'Not able to deserialize data provided.' Because the class is an application-wide @Provider for JsonMappingException (line 18), it is not scoped to contract status. The PR description says unrelated errors are 'delegated to the existing RESTEasy mapper to preserve previous behavior,' but no delegation happens here, and testInvalidEnum_Unrelated locks the new message in for an unrelated endpoint. Please confirm this fallback message matches the prior response for all other endpoints, or narrow the mapper (register for the more specific ValueInstantiationException / InvalidFormatException and let generic JsonMappingException keep hitting the built-in mapper).

app/src/main/java/io/apicurio/registry/rest/JacksonJsonMappingExceptionMapper.java
```diff
@@ -0,0 +1,55 @@
...
+
+@Provider
+@ApplicationScoped
+public class JacksonJsonMappingExceptionMapper implements ExceptionMapper<JsonMappingException> {
+
+    @Inject
+    CoreRegistryExceptionMapperService coreMapper;
+
+    @Context
+    Providers providers;
```
paoloantinori: providers is injected via @Context but never referenced. If the intent was to delegate non-status errors to the built-in RESTEasy mapper (as the description implies), that lookup would go through this field via providers.getExceptionMapper(...). Either wire up the delegation or drop the unused injection.

app/src/main/java/io/apicurio/registry/rest/JacksonJsonMappingExceptionMapper.java
```diff
@@ -0,0 +1,55 @@
...
+    public Response toResponse(JsonMappingException exception) {
+        String actualValue = null;
+        boolean isStatusEnum = false;
+
+        if (exception instanceof ValueInstantiationException) {
+            ValueInstantiationException vie = (ValueInstantiationException) exception;
+            if (vie.getType() != null && ContractStatusTransition.Status.class.equals(vie.getType().getRawClass())) {
+                isStatusEnum = true;
+                if (vie.getCause() instanceof IllegalArgumentException && vie.getCause().getMessage() != null) {
+                    actualValue = vie.getCause().getMessage();
```
paoloantinori: actualValue is read from the IllegalArgumentException's message text, assuming that message is exactly the raw invalid value. That couples the mapper to the wording of ContractStatusTransition.Status's factory or @JsonCreator: if that creator ever throws a descriptive message, the client-visible 'actual value' silently becomes a whole sentence (and the not(containsString(...)) assertions in testStatusTransition_BogusStatus would then fail for the wrong reason). Consider a more stable source for the value, or document the assumption on the Status factory so it is not changed in isolation.

paoloantinori (commented):
The status-enum error path works and the tests are thorough, but the new mapper is an application-wide @Provider for all JsonMappingExceptions, so its fallback path affects JSON deserialization errors on every endpoint, not just contract status. Two smells suggest the 'delegate to the built-in mapper' design described in the PR was not actually implemented: an unused Providers injection and extraction of the invalid value from an exception message string.

- JacksonJsonMappingExceptionMapper.java:53 (class declared app-wide at :18): the non-status fallback builds its own BadRequestException instead of delegating, yet the description claims delegation to the RESTEasy mapper. Needs confirmation that this message matches prior behavior for all other endpoints.
- JacksonJsonMappingExceptionMapper.java:24: the injected providers field is never used and is likely the intended delegation point.
- JacksonJsonMappingExceptionMapper.java:36: actualValue is read from an IllegalArgumentException message, a brittle coupling to the Status factory's message wording.
- DataContractsResourceTest.java (testInvalidEnum_Unrelated): this test codifies the new app-wide message for an unrelated endpoint, so it is the cross-cutting behavior change to sign off on consciously.
- ParameterValidationUtils.java:52: the argument-order swap to InvalidParameterValueException is a correct fix and matches the (name, expected, actual) order used by the new mapper at JacksonJsonMappingExceptionMapper.java:50. No change needed.

paoloantinori:
Thanks for the heads-up. I re-read the changes at e0412cd shortly after your comment and posted a follow-up review covering the delegation path, the `@Provider` scope, and the argument-order reconciliation. From my side the remaining open items are the ones flagged there (the unused `Providers` injection and confirming the fallback message matches prior behavior for unrelated endpoints). Happy to take another pass once those are addressed.

Hiten0305l:
Thanks for the review @paoloantinori . I agree that generic JSON mapping failures should continue through the existing RESTEasy/JAX-RS handling path. I will narrow the mapper to only handle ContractStatusTransition.Status deserialization failures (ValueInstantiationException / InvalidFormatException) and remove the unused Providers injection since delegation will no longer be needed. Please confirm this matches the expected approach.

paoloantinori:
The direction is right — narrowing the scope and dropping the unused Providers field both address the global-mapper concern. One thing to confirm before you push, since it determines how the narrowing actually works.

A @Provider implementing ExceptionMapper<JsonMappingException> intercepts every JsonMappingException in the application once registered. JAX-RS selects the mapper by the thrown type, not by what the body does, so there is no built-in "let RESTEasy handle this one, I will handle the rest." Once a JsonMappingException is routed here, toResponse must return a Response. Removing the Providers delegation is fine (it was unused), but the non-status-enum cases still need an explicit response, and whatever you return there becomes the app-wide contract for all malformed-JSON errors, not just contract-status.

So the question is what the fallback should be for unrelated JsonMappingExceptions. The current coreMapper.mapException(new BadRequestException(...)) works, but it is now your global shape for every bad-JSON error. If a uniform 400 across all endpoints is intentional, that is fine — worth a one-line comment noting the mapper owns all JsonMappingExceptions by design. If you would rather unrelated errors keep RESTEasy native response, that needs a narrower type binding than JsonMappingException. Note ValueInstantiationException and InvalidFormatException only share MismatchedInputException as a common supertype, which is still fairly broad, so the type-narrowing route has a limit.

Minor: the ValueInstantiationException branch reads the bad value from getCause().getMessage(), which relies on the enum constructor passing the raw input as the message. If that convention changes the reported value silently breaks. InvalidFormatException.getValue() is the more robust path if it is reachable on both branches.

Happy to re-read once pushed.

Hiten0305l:
Thanks for the clarification @paoloantinori . The global handling of JsonMappingException is intentional. The goal is to keep malformed JSON responses consistent through the existing ProblemDetails flow across the API. The status-specific handling remains restricted to ContractStatusTransition.Status, while unrelated JSON mapping failures use the generic fallback response intentionally.

I will add a short comment in the mapper documenting that this provider intentionally handles JsonMappingException globally.

If this approach looks good, I will proceed with the remaining changes and push the update for another review.

paoloantinori:
Understood, and that's a reasonable position. Routing the fallback through `coreMapper.mapException(new BadRequestException(...))` means unrelated JSON deserialization errors still land in the ProblemDetails format rather than RESTEasy's default, which is a consistency improvement, not a regression.

The one thing I'd want documented (which you already flagged) is the global scope itself: a future reader seeing a `@Provider` for `JsonMappingException` scoped to the contract-status use case will reasonably wonder why it catches every JSON mapping failure in the app. A two-line comment on the class explaining "intentionally global; all malformed-JSON responses go through ProblemDetails for consistency" closes that loop.

With the comment added, no further concerns from me.

Hiten0305l:
hey @paoloantinori !,I’ve addressed the requested changes already:

- Added the class-level comment explaining that the `JsonMappingException` provider is intentionally global so malformed JSON responses consistently use the standard `ProblemDetails` format.
- Removed the unused `providers` field.
- Updated the `ValueInstantiationException` handling to avoid relying on the enum constructor's exception-message convention for extracting the invalid value.

One clarification on the last point: `ValueInstantiationException` itself does not provide a `getValue()` API like `InvalidFormatException`, so I kept the handling compatible with the actual exception type rather than introducing an unavailable method.

All the requested changes are now addressed and pushed.

Whenever you’re free, please have another look and review the PR. Thanks again for the feedback!

paoloantinori:
Thanks for the clarification on `ValueInstantiationException` — you're right that it doesn't expose `getValue()` the way `InvalidFormatException` does, so extracting the offending value from `vie.getCause()` (the `IllegalArgumentException` from the enum `valueOf`) is the reasonable fallback. The two-guard approach (no-whitespace, no-dot) is a defensible heuristic for the enum-constant message shape.

I re-read `5f9342c` and the three changes look correct: the class-level Javadoc documents the intentional global scope, the unused `providers` field is gone, and the `InvalidFormatException` branch uses the proper `getValue()` path. The tests now distinguish the status-enum case from the wrong-shape and unrelated-enum cases via positive assertions (`name` equals the specific exception type), which is a real improvement over the earlier broad negative assertions.

The `ParameterValidationUtils` argument-order reconciliation is also consistent now. Nothing further from me on the scope or implementation — over to a maintainer.

Hiten0305l:
Thank you for taking the time to review the PR and for the detailed feedback. I’ve addressed the requested changes and updated the PR accordingly.

@EricWittmann , whenever you have a chance, I would really appreciate it if you could take a look at the updated PR. Please let me know if any further changes are needed.

Thank you again for your time and consideration.

Hiten0305l:
hi @paoloantinori , I ran the workflows and encountered the  failures in both Unit Tests (app-sql) and Integration Tests (h2-debezium). The failures appear to be unrelated to the changes in this PR. Since the implementation has already been approved, could you please take a look and let me know whether the PR is ready to be merged?

paoloantinori:
I checked both workflow runs — here's what's actually red:

- Latest run (31845343267): the only failing test job is **Unit Tests (app-sql)**. Integration Tests (h2-debezium) passed in that run (it had failed in the older one, so that one already cleared itself).
- The app-sql failure is entirely `io.apicurio.registry.event.sql.RegistryEventsTest` — all 30 test methods error with "Timeout waiting for result" in `lookupEvent`, everything else in the job is green.
- This is not caused by your changes: the same RegistryEventsTest timeout (same 30-error signature, ~618s elapsed) hit the `main` branch verify runs on Aug 14 (runs 31801054582 and 31798490286) with none of your code in them. The PR doesn't touch the event code paths at all.

So: unrelated infra flake. I'd just re-run the failed job (or let the verification gate retry) — no code changes needed. From my side the implementation has no further outstanding comments; once the flake clears the gate, it should be good to go.

Hiten0305l:
@EricWittmann  @paoloantinori Thanks for taking the time to investigate the workflow failures and for confirming that they're unrelated to the changes in this PR.

I appreciate the review and clarification. Since the implementation doesn't require any further changes from my side, please let me know if there's anything else I should address before this PR can move forward.

paoloantinori:
Nothing further needed from my side — the substantive rounds are covered and the workflow failures were confirmed unrelated to these changes. I'll leave the merge timing to the maintainers.

Hiten0305l:
@EricWittmann @jsenko please have a look, this is ready to merge. Other pr's are  in drafts due to this, would like to proceed with that ASAP.

EricWittmann (commented):
Thanks for the follow-up to #9348, @Hiten0305l — the tests are thorough and the approach of overriding RESTEasy's built-in `JsonMappingException` handler is the right one for this problem. A few things to address before we can merge:

---

**1. Fallback changes behavior app-wide — please update the PR description**

The mapper is an application-wide `@Provider` for all `JsonMappingException` instances. The fallback path (line 58) replaces RESTEasy's built-in response with a generic `"Not able to deserialize data provided."` message for **every** endpoint, not just contract status.

The PR description says "unrelated Jackson deserialization errors are delegated to the existing RESTEasy mapper to preserve previous behavior," but the code doesn't delegate — it creates a new `BadRequestException` and routes it through `coreMapper`. The result is actually an improvement (consistent ProblemDetails format), but the description should be updated to accurately reflect what the fallback does.

**2. Consider preserving Jackson's error detail in the fallback**

For non-status deserialization errors, the generic `"Not able to deserialize data provided."` message loses useful debugging information that Jackson normally provides (e.g., "Cannot deserialize value of type X from String Y"). Consider using the original exception message when it doesn't leak internal class names:

```java
String message = exception.getOriginalMessage();
if (message != null && !message.contains("io.apicurio")) {
    return coreMapper.mapException(new BadRequestException(message));
}
return coreMapper.mapException(new BadRequestException("Not able to deserialize data provided."));
```

This gives API clients better debugging info while still filtering out internal details.

**3. Use an import instead of a fully-qualified class name**

Line 58 uses `new jakarta.ws.rs.BadRequestException(...)` inline. Please add a proper import and use the short class name. (The project's contributor guidelines flag fully-qualified names used inline as an LLM artifact smell.)

**5. Link to the original issue**

This PR references #9348 but doesn't link to issue #9329 (which motivated that first fix). Please link to #9329 and note that this covers the remaining invalid-value case that wasn't addressed by the null/missing fix.

Hiten0305l:
Thanks @EricWittmann  for the detailed review.

I've addressed the requested changes:

- Updated the fallback behavior to preserve safe Jackson error messages.
- Replaced the fully qualified BadRequestException usage with an import.
- Updated the PR description to clarify the fallback behavior and reference #9329.

I also re-ran compilation, Checkstyle, and DataContractsResourceTest locally, and all checks passed.

Please let me know if there's anything else I should update.

Hiten0305l:
hi @EricWittmann ,I wanted to follow up on this PR regarding the changes you requested in your review. I addressed all of the requested updates and pushed.
I have also re-ran the relevant checks locally, and they passed successfully. The remaining CI failures appear to be related to the CI environment, including a Debezium integration-test timeout and a Maven dependency download failure.
When you have a chance, could you please review the latest changes and let me know if any further updates are required?
Thank you for your time and review.

Hiten0305l:
Hi @EricWittmann , I’ve completed the requested changes on my side. Whenever you have some time, I’d really appreciate it if you could review this PR. I’ve kept my other PRs in draft while waiting for feedback on this one, so your review would help me move forward with those as well.

Thank you for your time.

EricWittmann (approved):
Looks good.  Thanks @Hiten0305l

EricWittmann:
Thanks @Hiten0305l :+1:

PR #9377, opened 2026-08-07
feat(ui): scope error boundary to routes and log via LoggerService

### LINKED ISSUE #8930: UI: no React error boundary, so a single component error blanks the entire console
## Summary

The `ui-app` React tree has no error boundary anywhere. Any error thrown during render or in a lifecycle/effect of any component unmounts the whole tree, leaving the user on a blank white page with no message and no way to recover except a manual full reload. There is also no place to log such errors.

## Evidence

A search of `ui-app/src` for `ErrorBoundary`, `componentDidCatch`, and `getDerivedStateFromError` returns nothing. The app root wraps everything without a boundary:

- `ui/ui-app/src/app/App.tsx`
- `ui/ui-app/src/app/MainPageWithAuth.tsx` (wraps `<Routes>` directly)

## Impact

A malformed artifact or schema, an unexpected API response shape, or a bug in any single tab or component takes down the entire console rather than failing in an isolated area. Users see a blank page and lose any unsaved context.

## Proposal

- Add a top-level `ErrorBoundary` (React `getDerivedStateFromError` / `componentDidCatch`) around `<Routes>` with a friendly fallback, for example a PatternFly `EmptyState` with "reload" and "go to dashboard" actions.
- Optionally add per-route boundaries so one page's crash does not also take down the masthead and navigation.
- Log the caught error through the existing `useLoggerService` so failures are visible.

## Notes

Well-scoped, self-contained UI resilience improvement. Verified against current `main`.

Aditya-132:
Fixes #8930

## What

The console has a single error boundary, added in #8899, and it sits above the `Page` component in `App.tsx`. Because the masthead and navigation are rendered by `Page` inside `MainPageWithAuth`, a render error in any one page unmounts them together with the failed content, so the user still ends up on a blank console.

This PR completes the remaining items from the issue:

1. A second `ErrorBoundaryWithRouter` now wraps `<Routes>` inside `<Page>` in `MainPageWithAuth.tsx`. A crashing page renders the fallback in the content area while the masthead and navigation stay usable, and navigating to another route clears the error state. The boundary in `App.tsx` stays as the last resort for failures in the masthead, the auth wrapper, or the config lookup.
2. Caught errors are logged through the existing `LoggerService` rather than a direct `console.error` call. `ErrorBoundary` is a class component, so `ErrorBoundaryWithRouter` resolves the hook and passes the service down as a prop. The prop is optional and falls back to `console`, which keeps `ErrorBoundary` usable on its own.
3. The fallback gained a "Go to dashboard" action next to "Reload page", as suggested in the issue, so recovery no longer requires a full page load. The button is only rendered when a handler is supplied.

## Why

A malformed artifact, an unexpected API response shape, or a bug in a single tab currently takes down the whole console. Scoping the boundary to the routed content keeps the failure in the area that actually failed and leaves the user a way out.

## Tests

The location reset predicate that `componentDidUpdate` relies on was inlined and untestable, so it is extracted into `ErrorBoundary.utils.ts` and covered by `ErrorBoundary.utils.test.ts` with five cases: no error state, location changed, location unchanged, undefined current location, and undefined previous location.

`ui-app` has `environment: "node"` configured for vitest and no DOM testing dependencies, so the boundary is not exercised with a rendering test. Adding `jsdom` and a testing library felt out of scope for this change, but I am happy to add it if maintainers would prefer full render coverage.

Verified locally:

- `npm run lint` in `ui/ui-app`: 0 errors (2 pre-existing warnings in `ReferencesFormGroup.tsx`, untouched by this PR)
- `npx tsc --noEmit` in `ui/ui-app`: clean
- `npm test` in `ui/ui-app`: 60 tests across 11 files pass


paoloantinori (commented):
Overall the route-scoped boundary is the right fix, and the location-reset predicate extraction is clean and well covered (`ErrorBoundary.utils.ts` with `ErrorBoundary.utils.test.ts`). Two items to confirm before merge: that `LoggerService.error` preserves the full error and React component stack, and a recovery edge case where 'Go to dashboard' is a no-op on the dashboard route itself.

- `ErrorBoundary.tsx:38` routes `componentDidCatch` through `LoggerService.error` with three positional args. tsc accepting the call means the signature takes the arity, but confirm the implementation actually emits `error` and `errorInfo` rather than reducing the call to the message string, or the component stack is dropped from production logs.
- `ErrorBoundaryWithRouter.tsx:20` hardcodes the dashboard target and depends on `componentDidUpdate` resetting `hasError` on a pathname change. When `/dashboard` itself crashes there is no location change, `shouldResetOnLocationChange` returns false, and the button does nothing, so recovery falls back to 'Reload page'.
- The outer boundary in `App.tsx` is outside this diff. If it is a plain `ErrorBoundary` it still logs through `console.error` while the inner boundary uses `LoggerService`, splitting errors across two sinks; worth wiring both through the same logger for consistent telemetry.

ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx
```diff
@@ -30,11 +34,12 @@ export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundarySt
     }
 
     componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
-        console.error("[ErrorBoundary] Uncaught error:", error, errorInfo);
+        const logger: LoggerService = this.props.logger ?? console;
+        logger.error("[ErrorBoundary] Uncaught error:", error, errorInfo);
```
paoloantinori, line 38: `componentDidCatch` previously called `console.error` with the error and the React `errorInfo`. The new `logger.error` call typechecks, so the signature accepts the arity, but confirm `LoggerService.error` actually forwards the `Error` and `errorInfo` (the component stack) and does not log only the first argument. If the service reduces variadic args to the message parameter, the component stack disappears from production error reports.

ui/ui-app/src/app/components/errorPage/ErrorBoundaryWithRouter.tsx
```diff
@@ -1,8 +1,25 @@
...
     const location = useLocation();
-    return <ErrorBoundary location={location.pathname}>{props.children}</ErrorBoundary>;
+    const logger: LoggerService = useLoggerService();
+    const appNav: AppNavigation = useAppNavigation();
+
+    return (
+        <ErrorBoundary
+            location={location.pathname}
+            logger={logger}
+            onNavigateHome={() => appNav.navigateTo("/dashboard")}
```
paoloantinori: The 'Go to dashboard' action recovers by triggering `componentDidUpdate`'s location-change reset. If the dashboard route is the one that crashed, `navigateTo('/dashboard')` does not change `location.pathname`, `shouldResetOnLocationChange` returns false, and the button is a no-op, leaving only 'Reload page'. Consider hiding the action when already on `/dashboard`, or clearing `hasError` directly in the handler instead of relying on the navigation-driven reset.

ui/ui-app/src/app/MainPageWithAuth.tsx
```diff
@@ -61,81 +61,83 @@ export const MainPageWithAuth: FunctionComponent<MainPageWithAuthProps> = () =>
...
 
-                        <Route
-                            path="/explore/:groupId"
-                            element={ <GroupPage /> }
-                        />
-                        <Route
-                            path="/explore/:groupId/rules"
-                            element={ <GroupPage /> }
-                        />
+                            <Route
```
Nachiket-Roy, line 77: This looks like an unrelated lint/formatting change rather than a real code diff — can you revert it and keep this PR scoped to just the change it's meant to address?

ui/ui-app/src/app/components/errorPage/ErrorBoundary.tsx
Aditya-132: Confirmed, the component stack is preserved. `useLoggerService.ts:20-22` implements `error` as a straight variadic pass-through:

```ts
error(message: any, ...optionalParams: any[]): void {
    console.error(message, ...optionalParams);
},
```

So `logger.error("[ErrorBoundary] Uncaught error:", error, errorInfo)` reaches `console.error` with all three arguments, exactly as the previous direct call did. Nothing is reduced to the message parameter and `errorInfo.componentStack` still shows up in the log.

One detail worth noting: the `logger` prop is optional and falls back to `console`, and the fallback is written as `this.props.logger ?? console` rather than destructuring `.error` off the service, so the receiver stays intact in both branches.

ui/ui-app/src/app/components/errorPage/ErrorBoundaryWithRouter.tsx
Aditya-132: Good catch, fixed in bdc750da0.

I went with hiding the action rather than clearing `hasError` in the handler. Clearing it directly would help on the dashboard route but hurt everywhere else: `navigateTo` defers the actual `navigate()` call by 20ms (`useAppNavigation.ts:9`), so resetting first would re-render the crashed route in that gap, throw again, and log the same error a second time before the navigation lands. Hiding the button keeps 'Reload page' as the single action that actually works in that state instead of leaving a dead control.

`ErrorBoundaryWithRouter` now only passes `onNavigateHome` when the current route is not the home route, and `ErrorBoundary` already renders the button conditionally on that prop. The comparison lives in `shouldOfferNavigateHome` so it normalizes a trailing slash and compares against `appNav.createLink("/dashboard")` rather than the bare literal, which keeps it correct when a UI navigation prefix is configured. Six cases added to `ErrorBoundary.utils.test.ts` covering the prefix, both trailing-slash positions, and the root path.

ui/ui-app/src/app/MainPageWithAuth.tsx
Aditya-132: This one is not a formatting cleanup, it is the indentation shift from nesting `<Routes>` inside the new `<ErrorBoundaryWithRouter>` wrapper. The wrapper is the actual fix for the issue: the pre-existing boundary sits above `<Page>` in `App.tsx`, so a page crash unmounted the masthead and navigation along with the content.

Reverting the indentation is not an option here. `@apicurio/eslint-config` sets `"indent": ["error", 4, ...]` with no `ignoredNodes` exemption for JSX, so the rule applies to these elements. I verified it by de-indenting the block back to its original column and re-running eslint on the file:

```
✖ 69 problems (69 errors, 0 warnings)
  138:1  error  Expected indentation of 28 spaces but found 24  indent
  139:1  error  Expected indentation of 24 spaces but found 20  indent
```

Every one of those 69 errors is `indent`. With the current indentation `npx eslint src/app/MainPageWithAuth.tsx` exits clean.

The whitespace does make the diff look larger than it is. `git diff -w` on that file shows the real change is three lines: the import on line 3, and the opening and closing wrapper tags around `<Routes>`. Happy to reorganize if you would prefer the routes extracted into their own component to keep this file's diff minimal, but that would be a larger and more invasive change than the issue calls for.

Aditya-132:
@paoloantinori what do think now ?

paoloantinori:
Re-read the current head and all three items from #8930 land cleanly:

- Route-scoped boundary wrapping `<Routes>` inside `<Page>` keeps the masthead and nav live when a page crashes. The outer boundary in `App.tsx` stays as the last-resort catch. Correct layering.
- LoggerService integration via the hook-to-prop bridge in `ErrorBoundaryWithRouter` is the right way to get a hook-based service into a class component, and the `logger ?? console` fallback keeps `ErrorBoundary` standalone-testable.
- The `shouldOfferNavigateHome` predicate correctly suppresses the "Go to dashboard" button when the dashboard route itself is the one that crashed, and the trailing-slash / navigation-prefix normalization in `ErrorBoundary.utils.ts` is well covered by the test suite.

ui/ui-app/src/app/MainPageWithAuth.tsx
Nachiket-Roy: Thanks for making that clear

Aditya-132: > Thanks for making that clear

Are you also bot like @paoloantinori

Nachiket-Roy: Neither of us is a bot

Aditya-132: Screenshot 2026-08-08 184954

Nachiket-Roy: The team uses automated AI review the accounts are real and all

Aditya-132:
@Nachiket-Roy @EricWittmann Can you approve this PR

Nachiket-Roy:
> @Nachiket-Roy @EricWittmann Can you approve this PR

Eric will get to your pr also i help the project with reviews not a memeber.

Aditya-132:
@EricWittmann any update

Aditya-132:
@EricWittmann any update

carlesarnal (approved):

EricWittmann (approved):
LGTM

PR #9380, opened 2026-08-08
fix(schema-util): flag XSD attribute optional-to-required change as backward incompatible

### LINKED ISSUE #9379: bug(xsd): compatibility checker never flags an attribute changing from optional to required as BACKWARD-incompatible
## Description

**Persistence type**: in-memory (bug is in schema-util, independent of storage backend)

`XsdCompatibilityChecker.checkAttributeChanges()` (schema-util/xsd/src/main/java/io/apicurio/registry/xsd/rules/compatibility/XsdCompatibilityChecker.java:179-200) contains an empty `if` block for the case where an XSD attribute changes from `use="optional"` to `use="required"`:

    if (!existing.isRequired() && proposed.isRequired()) {
        // This is actually OK for backward (new schema can read old data)
        // but NOT OK for forward (old schema cannot read new data)
        // This will be caught in forward compatibility check
    }

This comment is incorrect and no other code path catches it. Per the class's own documented contract ("BACKWARD: Old data must be readable by new schema"), this transition IS backward-incompatible: an old document that omits the now-required attribute is invalid against the new schema. The equivalent case for **elements** (`checkElementChanges`, line 105) correctly flags `minOccurs` increasing from 0 as backward-incompatible — attributes are handled inconsistently with elements in the same file.

`AbstractCompatibilityChecker.testCompatibility()` (schema-util/common/.../AbstractCompatibilityChecker.java:43-46) implements FORWARD by calling `isBackwardsCompatibleWith()` with `existing`/`proposed` swapped — it does not run a separate "forward" code path, so the comment's claim that this case "will be caught in forward compatibility check" is false; it is never caught in either direction.

### Environment

Any deployment enforcing a `COMPATIBILITY` rule (BACKWARD or FULL) on an XSD artifact — reachable via the standard REST compatibility-check path (`app/.../rest/v3/impl`) for artifacts of type XSD.

### Steps to Reproduce

1. Register an XSD artifact version with an optional attribute:
   `<xs:attribute name="email" type="xs:string" use="optional"/>`
2. Configure a BACKWARD (or FULL) compatibility rule on the artifact/group.
3. Attempt to register a new version where the same attribute is now required:
   `<xs:attribute name="email" type="xs:string" use="required"/>`
4. Alternatively, reproduce directly via unit test using `XsdCompatibilityChecker.testCompatibility(CompatibilityLevel.BACKWARD, List.of(v1), v2, Map.of())`.

**Expected**: Step 3 is rejected — `isCompatible = false`, with a diff noting the attribute became required.
**Actual**: The version is accepted — `isCompatible = true`, empty diff set. No existing test in `XsdCompatibilityCheckerTest` covers this transition (only element `minOccurs` and add/remove-attribute cases are tested), confirming the gap was never exercised.

### Logs

N/A — silent incorrect pass, no exception or log output.

lokeshramchand-ctrl:
## Summary

- `XsdCompatibilityChecker.checkAttributeChanges()` had an empty `if` block for the optional→required attribute transition, with a comment claiming it would be "caught in forward compatibility check." It never was — `AbstractCompatibilityChecker` implements FORWARD by calling the same method with `existing`/`proposed` swapped, which only catches the opposite (required→optional) transition.
- Per the class's own documented contract ("BACKWARD: Old data must be readable by new schema"), an attribute becoming required is backward-incompatible: old documents omitting it become invalid. The equivalent element case (`minOccurs` increasing from 0) was already handled correctly, so attributes and elements were inconsistent.
- Fix mirrors the existing, working `checkElementChanges` pattern: add an `XsdIncompatibility` when `!existing.isRequired() && proposed.isRequired()`.

Closes #9379

## Test plan

- [x] Added `testBackwardIncompatible_AttributeOptionalToRequired` — asserts the transition is now correctly flagged incompatible under `BACKWARD`.
- [x] Added `testForwardCompatible_AttributeOptionalToRequired` — asserts the same transition is (correctly) compatible under `FORWARD`, since the old schema still permits the attribute.
- [x] Verified the new backward test fails against the pre-fix code (confirmed it's a real regression test, not a tautology) and passes after the fix.
- [x] `./mvnw test -pl schema-util/xsd` — all 22 tests pass.
- [x] `./mvnw checkstyle:check -pl schema-util/xsd` — passes.


schema-util/xsd/src/main/java/io/apicurio/registry/xsd/rules/compatibility/XsdCompatibilityChecker.java
```diff
@@ -179,16 +179,15 @@ private void checkAttributesBackwardCompatibility(XsdSchema existing, XsdSchema
...
         String attrPath = "/attribute[" + existing.getName() + "]";
-        
-        // Check if optional became required
+
+        // Check if optional became required (making it more restrictive)
         if (!existing.isRequired() && proposed.isRequired()) {
-            // This is actually OK for backward (new schema can read old data)
-            // but NOT OK for forward (old schema cannot read new data)
-            // This will be caught in forward compatibility check
+            incompatibilities.add(new XsdIncompatibility(
```
paoloantinori: Verified against AbstractCompatibilityChecker: FORWARD calls isBackwardsCompatibleWith with existing and proposed swapped (AbstractCompatibilityChecker.java:44), so this flag fires only for BACKWARD and stays silent for FORWARD. The same condition, via the swap, also now correctly flags the opposite transition (required-to-optional) as FORWARD-incompatible, which was a pre-existing false negative the old empty body skipped. That bonus fix has no test; consider adding testForwardIncompatible_AttributeRequiredToOptional to lock it in.

schema-util/xsd/src/test/java/io/apicurio/registry/rules/compatibility/XsdCompatibilityCheckerTest.java
```diff
@@ -259,6 +287,42 @@ void testBackwardIncompatible_IncreaseMinOccurs() {
...
+        TypedContent proposed = toTypedContent(SCHEMA_WITH_ATTRIBUTE_MADE_REQUIRED);
+
+        CompatibilityExecutionResult result = checker.testCompatibility(
+            CompatibilityLevel.FORWARD,
+            Collections.singletonList(existing),
+            proposed,
+            Collections.emptyMap()
+        );
+
+        Assertions.assertTrue(result.isCompatible(),
```
paoloantinori: Heads up: this FORWARD assertion also passes against the pre-fix code, because the swap makes the condition evaluate the same way either way. So this is a characterization test, not a regression guard for FORWARD (the BACKWARD test above is the real guard). If you want a FORWARD regression guard, the required-to-optional case, asserted incompatible, is the one that fails pre-fix and passes after.

paoloantinori (commented):
The fix is correct and ship-ready; I verified the load-bearing claim against the actual source. FORWARD compatibility is implemented by calling isBackwardsCompatibleWith with existing and proposed swapped (AbstractCompatibilityChecker.java:44), so the new flag inside checkAttributeChanges fires only in the right direction. The old empty body was a genuine false negative, and its inline comment had the semantics inverted (making an attribute required is backward-incompatible, not "OK for backward").

- Correct in both directions (XsdCompatibilityChecker.java:184). For BACKWARD the condition is true (existing optional, proposed required) and flags it; for FORWARD the swap makes existing required and proposed optional, so the condition is false and the change stays compatible, matching both new tests.
- Bonus fix, currently untested (XsdCompatibilityChecker.java:185). Via the same swap, this also closes a pre-existing FORWARD false negative: a real required-to-optional change now correctly reports forward-incompatible, where the old empty body skipped it. No test covers that case.
- The new FORWARD test is a characterization test, not a regression guard (XsdCompatibilityCheckerTest.java:321). It passes against the pre-fix code too; the BACKWARD test is the actual guard for this fix.
- Pre-existing, out of scope: under FORWARD the swap inverts the message direction, so a forward-incompatible required-to-optional change would be reported as "changed from optional to required." The element path has the same quirk (XsdCompatibilityChecker.java:104), so this is consistent rather than a regression.
- Pre-existing edge: parseAttribute treats use="prohibited" as not-required (XsdCompatibilityChecker.java:472), so prohibited-to-required is flagged but mislabeled, and optional-to-prohibited is not flagged at all. Out of scope here.

lokeshramchand-ctrl:
Thanks for the thorough review, @paoloantinori — addressed:

- **Missing test for the required→optional FORWARD case**: added `testForwardIncompatible_AttributeRequiredToOptional`. Verified it fails against the pre-fix code (checked out the pre-fix `XsdCompatibilityChecker.java` and re-ran it in isolation) and passes with the fix, so it's a real regression guard, not another characterization test.
- **SonarCloud Quality Gate** (`java:S1192`, duplicate `"Attribute '"` literal pushed to 3 occurrences by the new message): extracted `ATTRIBUTE_LABEL` and reused it at all three call sites. Quality Gate now passes (0 new issues).
- The message-direction quirk under FORWARD swap and the `use="prohibited"` edge case — agreed these are pre-existing and out of scope for this fix, left untouched.

Branch is rebased on current `main` (23/23 tests pass, checkstyle clean).

paoloantinori:
Good catch on the FORWARD regression test — verifying it fails against the pre-fix code (not just passes against the fix) is exactly the right way to prove it is a real guard and not a tautology. The `ATTRIBUTE_LABEL` extraction for the S1192 duplicate is clean too.

From my side the feedback loop is closed. Both changes landed and the test matrix now covers all four direction/transition combinations. Nothing further from me.

lokeshramchand-ctrl:
Thanks a lot for the thorough review and guidance, @paoloantinori! I'll wait on @EricWittmann for next steps and the final merge. It's been great working with you and the team on this.

paoloantinori:
Great work on this one. The fix and the test reasoning (proving the new guard fails against pre-fix code, not just passes against the fix) were exactly right. Over to Eric for the merge.

lokeshramchand-ctrl:
Hi @EricWittmann, this is ready whenever you get a chance to look. I've addressed @paoloantinori's review: added the missing FORWARD regression test and fixed the SonarCloud duplicate-literal issue. Quality Gate is passing. Happy to rebase on main again before merge if that's useful, just let me know.

paoloantinori:
Confirmed the SonarCloud duplicate-literal fix landed (the `ATTRIBUTE_LABEL` extraction is clean and behavior-preserving). The new FORWARD regression test is in place as discussed. No further comments from my side — ready for merge from my perspective.

lokeshramchand-ctrl:
Hi @EricWittmann  , Please look into this and let me know if any edits are required

EricWittmann (dismissed):
LGTM

EricWittmann:
@lokeshramchand-ctrl I tried to rebase your PR on main and fix a couple of conflicts, but I'm seeing build failures now.  Can you take a look?

lokeshramchand-ctrl:
@EricWittmann Found it — the merge conflict resolution left `checkAttributeChanges()` constructing `new XsdIncompatibility(...)`, a type that doesn't exist in this codebase. Every other call site in `XsdCompatibilityChecker` (and the class itself) uses `SimpleCompatibilityDifference`, so this was a stray leftover from the conflict resolution rather than an intentional change.

Fixed in 1fb5f68e: swapped it back to `SimpleCompatibilityDifference`. Verified:
- `mvn test-compile -pl schema-util/xsd -am` compiles cleanly
- `mvn test -pl schema-util/xsd -am`: 23/23 tests pass in `XsdCompatibilityCheckerTest`, 2/2 in `SchemaFactoryAccessorTest`
- `mvn checkstyle:check -pl schema-util/xsd` passes

Pushed to `issue-9379`. Let me know if you'd like anything else before merge.

lokeshramchand-ctrl:
Follow-up on CI/mergeability status after today's investigation:

**Root cause of DCO / BEHIND failures:** the branch's history had accumulated a merge commit and two GitHub-web commits (from a maintainer's earlier "merge main into branch" attempt) that carried no `Signed-off-by` trailer, which is what the `Validate PR` check was failing on. The branch was also several commits behind `main` by that point.

**Fix applied:** rebuilt `issue-9379` as a single clean, signed-off commit on top of current `upstream/main` (force-pushed, with `--force-with-lease` to avoid clobbering anything unexpected). In the process I found that `main` had independently picked up the `XsdIncompatibility` → `SimpleCompatibilityDifference` type rename via #8634's compatibility-checker refactor, but **not** the actual optional-to-required bug fix or its regression tests — those still needed to be reapplied on top. I also proactively extracted the now-3×-repeated `"Attribute '"` literal into an `ATTRIBUTE_LABEL` constant (the same duplicate-literal issue SonarCloud flagged earlier in this PR's history) so it doesn't reappear.

**Verification on the new commit (`fa297df24`):**
- `./mvnw test -pl schema-util/xsd` — 22/22 tests pass (20 pre-existing + 2 new)
- `./mvnw checkstyle:check -pl schema-util/xsd` — clean
- Confirmed `testBackwardIncompatible_AttributeOptionalToRequired` genuinely fails against the pre-fix code (real regression guard, not a tautology)
- `mergeStateStatus` is now `MERGEABLE` (previously `CONFLICTING`/`DIRTY`), and `Validate PR` / DCO now passes

**Two things outside this branch's control that are still showing red:**
1. **`Verification Gate`** — its own error message explains this: `lifecycle-ready=false` because the force-push reset review state (the prior `LGTM` was dismissed by GitHub when the head SHA changed), so the full CI suite hasn't been triggered yet. This should clear once a maintainer re-reviews.
2. **`License Compliance` (FOSSA)** — reports "1 issues found" on this commit. This PR's diff is unchanged in scope (still just `XsdCompatibilityChecker.java` + its test, no `pom.xml`/dependency changes), so this doesn't look related to this change; happy to dig further if a maintainer can share the FOSSA report link, since I don't have dashboard access from here.

@EricWittmann @paoloantinori — branch should be ready for another look whenever convenient.

EricWittmann (approved):
LGTM

EricWittmann:
Thanks @lokeshramchand-ctrl for your contribution and your patience.  :+1:

lokeshramchand-ctrl:
Thanks, @EricWittmann! Really appreciate the review and guidance from you and @paoloantinori. Glad to see this merged!

PR #9388, opened 2026-08-08
fix(kafkasql): move contract event creation into the storage layer

### LINKED ISSUE #9056: fix(kafkasql): add contract-specific methods to RegistryStorage and fix remaining event gaps
## Context

PR #8962 fixed an immediate bug where contract metadata and status events were silently lost on the KafkaSQL storage variant. The fix routes those events through `RegistryStorage.createEvent()`, which works but bypasses the standard storage architecture. This issue tracks the proper architectural cleanup.

## Problem

Contract events are the only events in the system where the REST layer (`GroupsResourceImpl`) is responsible for firing events. Every other event follows the pattern: **storage method does the work → storage method fires the CDI event**. The REST layer never thinks about events — except for contracts.

This happened because the storage layer has no concept of "contracts" — it only knows about labels. So the REST layer was forced to fire events itself, using `storage.createEvent()` as a backdoor to inject events into the storage pipeline.

Additionally, there are **6 methods** already on the `RegistryStorage` interface that have a similar gap — the SQL implementation fires events but the KafkaSQL implementation does not:

| Method | SQL fires event? | KafkaSQL fires event? |
|--------|:-:|:-:|
| `setArtifactContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `deleteArtifactContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `setVersionContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `deleteVersionContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `setGlobalContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |
| `deleteGlobalContractRuleset()` | Yes (`ContractRulesetConfigured`) | **No** |

In each case, the SQL implementation in `AbstractSqlRegistryStorage` fires `SqlOutboxEvent.of(ContractRulesetConfigured.of(...))` after the repository call. The KafkaSQL implementation submits the journal message and waits for the response, but never fires `KafkaSqlOutboxEvent` — so the event is silently lost.

### 1. Add contract-specific methods to `RegistryStorage`

```java
void updateContractMetadata(String groupId, String artifactId, ContractMetadata metadata);
void transitionContractStatus(String groupId, String artifactId, String fromStatus, String toStatus);
```

- The **SQL implementation** would do the label merge and fire `SqlOutboxEvent.of(ContractMetadataUpdated.of(...))` / `SqlOutboxEvent.of(ContractStatusChanged.of(...))`.
- The **KafkaSQL implementation** would submit a journal message, wait for the response, then fire `KafkaSqlOutboxEvent.of(...)` — exactly like `createArtifact()` and every other operation.
- The **REST layer** (`GroupsResourceImpl`) would just call `storage.updateContractMetadata()` / `storage.transitionContractStatus()` with no event awareness.

### 2. Fix the 6 `ContractRulesetConfigured` event gaps in KafkaSQL

For each of the 6 methods listed above, add `outboxEvent.fire(KafkaSqlOutboxEvent.of(ContractRulesetConfigured.of(...)))` after the journal round-trip completes in `KafkaSqlRegistryStorage`. These already follow the right pattern (storage interface method → journal message → response), they're just missing the final event fire step.

### 3. Remove the `createEvent()` backdoor usage

Once contract operations go through proper storage methods, the `createEvent()` calls from `GroupsResourceImpl` can be removed. The `createEvent()` method on `RegistryStorage` can potentially be reverted to a no-op on KafkaSQL (or removed entirely if no other callers need it).

### SQL Storage

```
AbstractSqlRegistryStorage.createArtifact()
  ├── INSERT artifact/version into database
  └── outboxEvent.fire(SqlOutboxEvent.of(ArtifactCreated.of(...)))
        └── SqlEventsProcessor → sqlStore.createEvent()
              └── INSERT INTO outbox → Debezium CDC captures from WAL
```

### KafkaSQL Storage

```
KafkaSqlRegistryStorage.createArtifact()
  ├── (1) Submit journal message to kafkasql-journal topic
  ├── (2) Wait for journal round-trip (consumer dispatches to embedded SQL)
  ├── (3) outboxEvent.fire(KafkaSqlOutboxEvent.of(ArtifactCreated.of(...)))
  │         └── KafkaSqlEventsProcessor → "registry-events" Kafka topic
  └── (4) Return result
```

Key design point: `KafkaSqlOutboxEvent` is fired **by the originating node only**, after the journal round-trip completes. Other nodes replaying the same journal message call `dispatchTo(sqlStore)` which fires `SqlOutboxEvent` — a no-op on H2. So no duplicate events.

## Additional Cleanup

PR #8962's fix left two minor nits that should be addressed:
- Fully qualified class names (`io.apicurio.registry.events.ContractMetadataUpdated`, `io.apicurio.registry.events.ContractStatusChanged`) in `GroupsResourceImpl` should be converted to regular imports.
- The `KafkaSqlContractEventsTest` should have `@Tag(ApicurioTestTags.SLOW)` since it polls Kafka for up to 20s (consistent with the analogous `KafkaSqlEventsTest`).

## References

- PR #8962 — the immediate bug fix (merged)
- Issue #8928 — the original bug report
- Analysis by @EricWittmann in [PR #8962 review](https://github.com/Apicurio/apicurio-registry/pull/8962#pullrequestreview-2969058405)

Anuragkumar-687:
# Fixes #9056

## Problem

Contract events were the only events in the registry that were fired directly from the REST layer. Every other event follows the established pattern of **storage performs the operation → storage fires the CDI event**, but because the storage layer had no contract-specific operations, `GroupsResourceImpl` created these events itself through `storage.createEvent()`.

### Contract-specific storage methods

Added `updateContractMetadata()` and `transitionContractStatus()` to `RegistryStorage` as `default` methods that throw, following the existing `mergeArtifactLabels()` / `mergeVersionLabels()` pattern introduced by the contracts work. As a result, storage implementations that do not support contracts (GitOps, KubernetesOps, the polling storages, and `TestInMemoryRegistryStorage`) inherit the default implementation and required no changes.

Storage implementations now handle both the data update and event creation:

* **SQL** derives the contract prefix from the artifact labels, performs the label merge, and fires a `SqlOutboxEvent`.
* **KafkaSQL** submits a journal message, waits for the operation to complete, and then fires a `KafkaSqlOutboxEvent`, matching the pattern already used by operations such as `createArtifactRule()`. The two new journal message classes are also registered in `KafkaSqlMessageIndex`.

### REST layer

`GroupsResourceImpl` now delegates contract updates to the new storage methods and no longer creates events directly. This removes both `storage.createEvent(...)` calls, which also eliminates the fully qualified event class references mentioned in the issue. The existing `createEvent()` method remains on the `RegistryStorage` interface because it is still used elsewhere.

### Test tag

`KafkaSqlContractEventsTest` is now annotated with `@Tag(ApicurioTestTags.SLOW)`, making it consistent with `KafkaSqlEventsTest`.

## Scope

While this work was in progress, part of the issue was implemented upstream. As a result, this PR intentionally does **not** include:

* The six missing `ContractRulesetConfigured` events in KafkaSQL.
* The shared contract event tests added to `RegistryEventsTest`.

Those changes are already present on `main`.

## Design note

The original issue suggested using `ContractMetadata` in the storage method signature. Instead, this implementation uses `EditableContractMetadataDto` to avoid introducing a dependency from the storage layer to the generated REST model. This keeps the storage API independent of the REST layer while preserving the intended behavior.

## `findContractId` refactoring

The storage layer needs to derive the contract prefix from artifact labels, so `ContractLabels` now provides `findContractId(labels)` and `prefixFor(labels)`. This made the private `findContractId` helper in `GroupsResourceImpl` redundant, so it was removed and its three call sites now use `ContractLabels` directly. This is the only non-mechanical change in the refactoring.

## Behavior intentionally unchanged

`transitionContractStatus()` continues to perform one label merge per key rather than a single merge over the entire contract prefix. Each merge deletes labels by prefix, so combining them would also remove unrelated contract labels such as owner, classification, and stage. The existing behavior has been preserved exactly; it has only been moved into the storage layer.

## Testing

The existing contract event tests in `RegistryEventsTest` exercise the refactored endpoints and run against both storage implementations through `KafkaSqlEventsTest`.

Verified with:

* `RegistryEventsTest` contract tests (SQL profile) — **3/3**
* `KafkaSqlEventsTest` contract tests (KafkaSQL profile) — **3/3**
* `KafkaSqlContractEventsTest` — **1/1**
* `DataContractsResourceTest` — **35/35**
* `ReadOnlyRegistryStorageTest` — **1/1**
* `./mvnw checkstyle:check -pl app`

The KafkaSQL tests assert an exact event count, confirming that the journal replay's `SqlOutboxEvent` does not produce duplicate events.


paoloantinori (commented):
Overall this is a clean, well-motivated refactor: contract events now follow the same storage-performs-the-work-then-fires-the-event pattern as every other event, the REST layer is simpler, and the read-only decorator plus ReadOnlyRegistryStorageTest are wired correctly. The points below are mostly KafkaSQL concerns worth confirming before merge.

- The numeric suffix on the new message classes collides with an existing one: `TransitionContractStatus4Message` shares `4` with `MergeArtifactLabels4Message` (KafkaSqlMessageIndex.java:69). If the serde keys on the class name this is only cosmetic, but if anything treats that suffix as a type id it collides; worth confirming and, if cosmetic, renaming so each class keeps a unique number.
- The new SQL implementations fire `SqlOutboxEvent` from inside the storage method, which on KafkaSQL also runs during journal replay on every node (originator included), on top of the `KafkaSqlOutboxEvent` fired at KafkaSqlRegistryStorage.java:774. The exact event-count assertion in KafkaSqlContractEventsTest is the right guard; confirm it explicitly counts `ContractMetadataUpdated` and `ContractStatusChanged`, not just the pre-existing event types.
- The lifecycle date is computed with `LocalDate.now()` inside transitionContractStatus (AbstractSqlRegistryStorage.java:978), so on KafkaSQL each replica computes it during replay instead of the value being pinned in the journal message as before (the old REST code put the date string into the merge message). Day granularity makes this harmless in practice, but computing wall-clock time during replay is a determinism smell for a replicated log; consider resolving the date once and carrying it in the journal payload.
- Confirm the two new message classes carry the Apache license header; the diff excerpt starts at `package` with no header visible, and `checkstyle:check` does not always enforce it.
- Nice reuse moving `findContractId`/`prefixFor` into ContractLabels and dropping the private duplicate from GroupsResourceImpl, and good call keeping the storage signature on `EditableContractMetadataDto` to avoid a storage-to-REST-model dependency.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/serde/KafkaSqlMessageIndex.java
```diff
@@ -64,7 +64,9 @@ private static void indexMessageClasses(Class<? extends KafkaSqlMessage>... mcla
                 DeleteGlobalContractRuleset0Message.class,
                 InsertContractAuditEntry1Message.class,
                 MergeArtifactLabels4Message.class,
-                MergeVersionLabels5Message.class);
+                MergeVersionLabels5Message.class,
+                UpdateContractMetadata3Message.class,
+                TransitionContractStatus4Message.class);
```
paoloantinori: Both `MergeArtifactLabels4Message` and the new `TransitionContractStatus4Message` carry the suffix `4`. `lookup(String)` suggests the serde keys on the class name, in which case this is only cosmetic, but the numbering reads like a per-class identifier and a duplicate is surprising. Confirm the suffix is not used as a type id anywhere; if it is purely cosmetic, rename one so each message class keeps a distinct number.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -940,6 +948,45 @@ public void mergeArtifactLabels(String groupId, String artifactId, String prefix
...
+
+        // Each merge is scoped to a single label key so that the other contract labels
+        // are left untouched.
+        String statusKey = prefix + ContractLabels.SUFFIX_STATUS;
+        mergeArtifactLabels(groupId, artifactId, statusKey, Map.of(statusKey, toStatus));
+
+        if (ContractStatus.STABLE.name().equals(toStatus)) {
+            String stableDateKey = prefix + ContractLabels.SUFFIX_STABLE_DATE;
+            mergeArtifactLabels(groupId, artifactId, stableDateKey,
+                    Map.of(stableDateKey, LocalDate.now().toString()));
```
paoloantinori: `LocalDate.now()` runs inside the storage method, which on KafkaSQL executes during journal replay on each replica. The previous REST implementation put the resolved date string into the merge journal message, so every replica applied the same value. Day granularity makes divergence unlikely, but computing wall-clock time during replay is non-deterministic for a replicated log. Consider resolving the date once and including it in the journal payload.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -763,6 +765,26 @@ public void deleteGlobalContractRuleset() throws RegistryStorageException {
                 .of(ContractRulesetConfigured.ofGlobal(ContractRulesetConfigured.Action.DELETE)));
     }
 
+    @Override
+    public void updateContractMetadata(String groupId, String artifactId,
+            EditableContractMetadataDto metadata) throws RegistryStorageException {
+        var message = new UpdateContractMetadata3Message(groupId, artifactId, metadata);
+        var uuid = blockOnResult(submitter.submitMessage(message));
+        coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
```
paoloantinori, line 776: This fires `KafkaSqlOutboxEvent` after the journal response, while the SQL implementation (invoked when the journal message is replayed on each node, this one included) also fires a `SqlOutboxEvent` for `ContractMetadataUpdated`. The exact event-count assertion in `KafkaSqlContractEventsTest` is what guards against duplicates here, so confirm it explicitly covers `ContractMetadataUpdated` and `ContractStatusChanged`, not only the event types that existed before this change.

Anuragkumar-687:
@paoloantinori Thanks for the detailed review - I went through all four points.

1. Message suffix

I checked this first. The numeric suffix is arity rather than a type identifier: KafkaSqlMessageIndex registers messages using mclass.getSimpleName(), and lookup(String) resolves using the class simple name from the message-type header. There are already many existing classes that share the same numeric suffix, so there isn't a collision in practice. The class has been renamed to TransitionContractStatus5Message, but that's a consequence of point 3 (the additional parameter), not because of the suffix.

2. Duplicate-event guard

Good catch. My PR description overstated what KafkaSqlContractEventsTest verifies—it uses anyMatch, so it doesn't guard against duplicate events. The actual guard is the upstream RegistryEventsTest contract tests, which assert assertEquals(1, events.size()) for the filtered contract events, and KafkaSqlEventsTest inherits those assertions. I re-ran those tests under the KafkaSQL profile, and they pass for both CONTRACT_METADATA_UPDATED and CONTRACT_STATUS_CHANGED. I've also updated the PR description to reflect this.

3. LocalDate.now() during replay

I agree this was a real issue. As you pointed out, resolving the date during journal replay means a replica rebuilding from the log could record the replay date rather than the original transition date. I've changed the implementation so the effective date is resolved once in the REST layer and passed through the storage method and journal payload, restoring the same behavior as before the refactoring. transitionContractStatus now takes an effectiveDate parameter, and the SQL implementation no longer reads the system clock.

4. License headers

I checked the existing convention before changing anything. None of the existing classes under storage/impl/kafkasql/messages/ include per-file license headers, so I kept the new classes consistent with the rest of the package. If that convention has changed, I'm happy to add them.

Thanks again for the careful review—it helped catch the replay determinism issue.

paoloantinori:
Thanks for the thorough follow-up. Walking the four points against head 84f8a1f:

1. Message suffix — agreed, the suffix is arity and the simple name is the actual serde key, so no real collision. The rename to TransitionContractStatus5Message from the added effectiveDate parameter is consistent.
2. Duplicate-event guard — good clarification; the assertEquals(1, events.size()) in the upstream contract tests is the real guard, and re-running them under the KafkaSQL profile for both event types is the confirmation I was asking for. PR description update is appreciated.
3. LocalDate.now() during replay — this was the substantive one, and the fix is exactly right: resolving effectiveDate once in the REST layer and carrying it in the journal payload restores the pre-refactor determinism, and the SQL implementation no longer reads the wall clock. Checked the new storage signature, the journal message, and the SQL impl all thread it through.
4. License headers — checked the package convention, the existing message classes indeed omit per-file headers, so keeping the new ones consistent is correct.

From my side all points are addressed and I have no further comments on this head.

Anuragkumar-687:
Thanks, @paoloantinori! I really appreciate the thorough review and the follow-up.

Glad we were able to resolve the determinism issue around `LocalDate.now()` cleanly, and thanks for calling out the upstream contract tests as the right validation for the duplicate-event guard.

Thanks again for taking the time to go through everything so carefully.

paoloantinori:
Happy to help. Glad the determinism fix and the upstream contract test coverage landed cleanly. Good luck with the rest.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/messages/UpdateContractMetadata4Message.java
```diff
@@ -0,0 +1,32 @@
...
+    private String groupId;
+    private String artifactId;
+    private EditableContractMetadataDto metadata;
+
+    @Override
+    public Object dispatchTo(RegistryStorage storage) {
+        storage.updateContractMetadata(groupId, artifactId, metadata);
+        return null;
+    }
+}
```
aclfe, line 34: No new tests for the two new interface methods / SQL impls / KafkaSQL impls / journal messages (this class and `TransitionContractStatus5Message`). `KafkaSqlContractAuditJournalTest` / `KafkaSqlGlobalContractRulesetJournalTest` are the existing pattern for journal round-trip coverage on a new message type.

app/src/main/java/io/apicurio/registry/contracts/ContractLabels.java
```diff
@@ -52,6 +54,37 @@ public static String key(String contractId, String suffix) {
...
+    public static String findContractId(Map<String, String> labels) {
+        if (labels == null) {
+            return null;
+        }
+        String suffix = "." + SUFFIX_ID;
+        for (Map.Entry<String, String> entry : labels.entrySet()) {
+            String key = entry.getKey();
+            if (key.startsWith(PREFIX) && key.endsWith(suffix)) {
+                String middle = key.substring(PREFIX.length(), key.length() - suffix.length());
+                if (!middle.contains(".")) {
```
aclfe, line 76: `findContractId` drops the `!middle.isEmpty()` guard that both `ContractMetadataMapper.detectContractId` and `V3ApiUtil.detectContractId` have (the two methods this is meant to replace). Worth adding `!middle.isEmpty() && ` here, then actually deleting those two duplicates now that there's a shared one to call instead.

aclfe (commented):
Nice cleanup, storage-does-the-work-then-fires-the-event is the right shape and the KafkaSQL methods match the existing pattern. One thing worth a second look before merge.

**PUT contract/metadata comes back empty for ODCS-projected contracts**

The return value now resolves the contract id from the post-merge labels instead of the pre-merge ones:

https://github.com/Apicurio/apicurio-registry/blob/84f8a1f05586581471c3edf5ac1af0c5b63396f0/app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java#L2026-L2030

Problem is the merge deletes the label `findContractId` looks for. `mergeArtifactLabels` deletes by prefix ([L929-L931](https://github.com/Apicurio/apicurio-registry/blob/84f8a1f05586581471c3edf5ac1af0c5b63396f0/app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java#L929-L931)), `toLabels` never re-emits `SUFFIX_ID`, and that's the label `OdcsLabelProjector` writes ([L76](https://github.com/Apicurio/apicurio-registry/blob/84f8a1f05586581471c3edf5ac1af0c5b63396f0/app/src/main/java/io/apicurio/registry/contracts/odcs/OdcsLabelProjector.java#L76)). So after the merge, `.id` is gone, `findContractId` returns null, `fromLabels` falls back to the bare `"contract."` prefix and finds nothing.

Ran it against the branch to confirm: create an artifact, POST an ODCS contract, PUT `{"ownerTeam":"probe-team","status":"DRAFT"}` to `/contract/metadata`. Labels write correctly (`owner.team` and `status` both land under `contract.{cid}.`), but the PUT response is `{}`. Feeding the same post-merge labels into `fromLabels` with the pre-merge contract id (what main does) gives back the right values, so it's the response that regressed, not the write.

Fix is keeping the pre-read, same as `transitionContractStatus` already does:

```java
ArtifactMetaDataDto existing = storage.getArtifactMetaData(rawGroupId, artifactId);
String contractId = ContractLabels.findContractId(existing.getLabels());

storage.updateContractMetadata(rawGroupId, artifactId, editableDto);
// ...
ContractMetadataDto result = contractMetadataMapper.fromLabels(updated.getLabels(), contractId);
```

Side note, not from this PR: the merge itself also drops `.id`/`.version` and any field not in the request body, on main too. That breaks the later GET `contract/metadata` and the ODCS export regardless of this PR. Probably its own issue.

No existing test covers this path, everything in `DataContractsResourceTest`/`RegistryEventsTest` runs on artifacts with no `.id` label, and `OdcsContractResourceTest` never PUTs metadata after projecting.

**Smaller stuff (see inline comments)**

- effectiveDate is pinned now (good), but prefix and label mapping are still resolved inside the journalled method rather than carried in the message, so replay recomputes both. Probably fine given the same reasoning that motivated pinning the date, just flagging.
- No new tests for the two new interface methods / SQL impls / KafkaSQL impls / journal messages.
- `ContractLabels.findContractId` drops the `!middle.isEmpty()` guard the two methods it's replacing have (`ContractMetadataMapper.detectContractId`, `V3ApiUtil.detectContractId`), worth carrying over and then actually deleting those two duplicates.
- `KafkaSqlRegistryStorage.createEvent` looks unreachable now, both REST call sites are gone.

checkstyle and test-compile are clean on 84f8a1f. Verification Gate failure is just lifecycle state (not accepted yet), not a real failure, nothing's run in CI on this yet.

@Anuragkumar-687

aclfe:
Also, side note @paoloantinori @Anuragkumar-687, following up on the effectiveDate thread.

`MergeArtifactLabels4Message` carries the already-resolved `prefix` and label map as fields, `dispatchTo` just applies them literally on replay. The two new methods don't: `prefix` and `toLabels(metadata, prefix)` are computed inside `updateContractMetadata`/`transitionContractStatus` itself, which runs on every node when it replays the message, not resolved once and carried in the payload.

In steady state that's fine, same code version and same journal order means `existing.getLabels()` is identical everywhere and the pure functions produce the same output. But a node that replays the full journal from scratch (new node joining, DR rebuild from the journal) runs today's `toLabels`/`prefixFor` against a historical message. If that logic changes between when the message was written and when it's replayed, the reconstructed labels differ from what the original node actually stored. Same root cause as the `LocalDate.now()` case, just triggered by a version boundary during replay instead of on every call.

Thoughts?

paoloantinori:
Good catch, and I think the concern is valid. `MergeArtifactLabels4Message` carries the resolved `prefix` and label map as fields and `dispatchTo` applies them verbatim, so the journal records exactly what every node stores. The two new messages (`UpdateContractMetadata3Message`, `TransitionContractStatus5Message`) instead carry only the inputs and recompute `prefixFor(existing.getLabels())` and `toLabels(metadata, prefix)` inside the replayed storage method. In steady state that's deterministic, but a node replaying the full journal from scratch runs today's `prefixFor`/`toLabels` against a historical message, so a change in that logic between write time and replay time produces different stored labels. Same class of problem as the `LocalDate.now()` case we fixed, just bounded by a version boundary instead of every call.

The cleanest fix mirrors `MergeArtifactLabels4Message`: resolve `prefix` and the final label map in the REST layer before submitting the message, carry them as fields, and have `dispatchTo` call `mergeArtifactLabels(groupId, artifactId, prefix, labels)` directly. That way the journal is self-describing and replay is invariant to code changes in the mapper or `prefixFor`. Worth doing now rather than after a DR-replay surprises someone.

Anuragkumar-687:
@aclfe pushed a follow-up commit with the remaining three points.

**Tests for the new methods and messages.** Added`KafkaSqlContractMetadataJournalTest`, asserting `updateContractMetadata` and `transitionContractStatus` produce `UpdateContractMetadata4Message` / `TransitionContractStatus6Message` on the journal topic, following the `KafkaSqlGlobalContractRulesetJournalTest` pattern you pointed at. 2/2.

**ODCS regression test.** Added a case to `OdcsContractResourceTest` that projects a contract and then updates metadata, asserting the response carries `ownerTeam` and `status` back instead of an empty body. You were right that nothing covered it — the existing contract tests all run on artifacts with no `contract.{id}.id` label.

**Duplicate helpers.** Deleted `ContractMetadataMapper.detectContractId` and `V3ApiUtil.detectContractId`; both call sites now use `ContractLabels.findContractId`.

Left `createEvent` on the interface as-is, since removing it is a separate change from this PR's scope.

Verified: `KafkaSqlContractMetadataJournalTest` 2/2, `OdcsContractResourceTest` 11/11, `DataContractsResourceTest` 35/35, `ContractMetadataMapperTest` 16/16, `ReadOnlyRegistryStorageTest` 1/1, `KafkaSqlEventsTest` contract tests 3/3, checkstyle clean.

@paoloantinori this also carries the resolved prefix and label map in the journal payload as you suggested, so replay no longer recomputes them.

@EricWittmann ready for a look whenever you have time.

paoloantinori:
Thanks for the follow-up. The determinism fix landed in the right shape: prefix and the resolved label map (and for the status transition, the lifecycle date) are computed once in the REST layer and carried verbatim in the journal payload, so every KafkaSQL replica applies identical values on replay. That matches the MergeArtifactLabels4Message pattern we discussed, which was the goal.

The new KafkaSqlContractMetadataJournalTest covering both new message types, plus the ODCS projection regression case, close the coverage gaps I was worried about. The helper deduplication onto ContractLabels.findContractId is a nice cleanup.

Nothing further from me. Over to @EricWittmann for the accept path.

Anuragkumar-687:
Thank you, @paoloantinori, for your guidance throughout the review. I really appreciate the thoughtful feedback and direction. It helped make sure the determinism fix and coverage landed in the right shape.

paoloantinori:
Glad it landed cleanly, Anurag. The determinism fix in the right shape and the upstream contract test coverage are both good outcomes. Nothing further from my side here.

Anuragkumar-687:
Hi @EricWittmann, all changes are reviewed and @paoloantinori confirmed all things landed perfectly. Please look into it whenever you get time.

paoloantinori:
Nothing further from my side — the earlier points are all addressed at the current head.

One heads-up before the final pass: the PR is currently showing merge conflicts with main, so it will likely need a rebase before it can go in.

Anuragkumar-687:
@paoloantinori rebased onto current main, conflicts resolved, and I squashed the four commits into one.

paoloantinori:
Re-read the PR at the rebased head (99a5696) before replying, and everything we went through survived the rebase and the squash:

- The determinism fix is intact: prefix and the resolved label map are computed once in the REST layer and carried in UpdateContractMetadata4Message, and the status transition passes the resolved effectiveDate through TransitionContractStatus6Message, so the SQL path never reads the clock.
- ContractLabels.findContractId is the single implementation now; the three copies in GroupsResourceImpl, V3ApiUtil and ContractMetadataMapper are all gone.
- Both new messages are registered in KafkaSqlMessageIndex.
- KafkaSqlContractMetadataJournalTest is still there and asserts both operations land in the journal topic.
- The consolidation also stays compatible with the new test main picked up in the meantime (testBareContractIdLabelDoesNotThrow): the bare contract.id key hits the end <= start guard in findContractId and returns null, exactly what that test expects.

One small observation from the rebase, not a blocker: ContractLabels.prefixFor() has no callers in the repo. It was equally unused at the previous head, so nothing got lost in the rebase, but it may be worth either wiring it into the two GroupsResourceImpl sites that still derive the prefix inline, or dropping it.

From my side this is ready for a final maintainer pass.

Anuragkumar-687:
@paoloantinori Thanks for re-checking the rebased head. Addressed the `prefixFor` observation.

It turned out to be four inline sites rather than two - `ContractMetadataMapper` and `V3ApiUtil` derive the prefix the same way, alongside the two in `GroupsResourceImpl`. All four already have the contract id in hand, so
`prefixFor(Map)` was the wrong shape: calling it would re-scan the labels to recover an id the caller already had. I changed it to `prefixFor(String contractId)` and wired all four sites to it, which drops a ternary that was
duplicated four times.

Verified: `DataContractsResourceTest` 37/37, `OdcsContractResourceTest` 11/11, `ContractMetadataMapperTest` 17/17, `ReadOnlyRegistryStorageTest` 1/1, checkstyle clean. Branch is rebased onto current main with a single commit.

Anuragkumar-687:
Fixed both triage findings.

`java.time.LocalDate` in `GroupsResourceImpl` and `java.util.Map` in `ReadOnlyRegistryStorageDecorator` are now imports.

One note on the decorator: only one of the three `java.util.Map` occurrences was from this PR — `mergeArtifactLabels` and `mergeVersionLabels` already used the inline form. I converted all three rather than adding an import and leaving FQNs next to it. There are two more in `KafkaSqlRegistryStorage` that I left alone, since they're pre-existing on main and outside this PR's diff.

Verified: `DataContractsResourceTest` 37/37, `OdcsContractResourceTest` 11/11, `ContractMetadataMapperTest` 17/17, `ReadOnlyRegistryStorageTest` 1/1, checkstyle clean.

paoloantinori:
Re-read the head after your comment and the prefixFor refactor landed cleanly: findContractId and prefixFor(String) are now on ContractLabels, all four derivation sites (GroupsResourceImpl, ContractMetadataMapper, V3ApiUtil) go through them, and the private duplicates are gone. Agreed the String overload is the right shape: every caller already holds the contract id, so re-scanning the labels to recover it would be wasted work. Triage is green at this head too. Nothing further from my side, ready for maintainer accept.

Anuragkumar-687:
Hi @EricWittmann, all changes are reviewed. Please look into it whenever you get time.

Anuragkumar-687:
Hi @carlesarnal, all changes are reviewed. Please look into it whenever you get time.

EricWittmann (approved):
LGTM 👍

EricWittmann:
Thanks @Anuragkumar-687 for the contribution and your patience.  :+1:

Issue #7336, opened 2026-02-10
Remove Java SDK v2 module alongside REST API v2 deprecation


carlesarnal:
## Summary

The `java-sdk-v2/` module provides a client SDK for the deprecated REST API v2. When REST API v2 is deprecated (see #7330), the corresponding Java SDK v2 module should also be removed.

## Current State

- **Module location**: `java-sdk-v2/`
- **Purpose**: Kiota-generated client SDK for REST API v2
- **Files**: ~80+ generated Java files in `target/generated-sources/kiota/`
- **Factory**: `io.apicurio.registry.client.RegistryV2ClientFactory`

## Rationale

1. The SDK is tightly coupled to REST API v2 - when the API is removed, the SDK becomes useless
2. Maintaining two SDK versions increases maintenance burden
3. Users should migrate to the v3 SDK (`java-sdk/`) which supports the current API

## Proposed Action

1. Link this issue to #7330 (REST API v2 deprecation timeline)
2. Add deprecation warnings to `RegistryV2ClientFactory` in next minor release
3. Remove the entire `java-sdk-v2/` module when REST API v2 is removed
4. Update documentation to guide users to `java-sdk/` (v3)

## Related Issues

- #7330 - [RFC] Plan REST API v2 deprecation timeline

## Acceptance Criteria

- [ ] Deprecation warning added to `RegistryV2ClientFactory`
- [ ] Migration guide from v2 SDK to v3 SDK documented
- [ ] Module removed when REST API v2 is deprecated

riya-saharan:
Hi, I'd like to pick this up. Since full removal of java-sdk-v2/ depends on the REST API v2 deprecation timeline in #7330, I'll start with the parts that aren't blocked:

Add a @Deprecated annotation and Javadoc warning to RegistryV2ClientFactory
Write a migration guide from the v2 SDK to the v3 SDK

I'll hold off on removing the module itself until #7330 confirms REST API v2 is actually gone. Let me know if you'd prefer this split differently (e.g., separate PRs, or tracked as sub-tasks).

Opening a PR shortly regarding this issue.

EricWittmann:
@riya-saharan The v2 API and SDK are deprecated but will not be removed until Apicurio Registry 4.0.  👍

Issue #7330, opened 2026-02-10
[RFC] Plan REST API v2 deprecation timeline


carlesarnal:
## Summary

This RFC proposes establishing a deprecation timeline for REST API v2 to eventually consolidate on v3 as the single supported API version.

## Background

Apicurio Registry currently maintains two active REST API versions:
- **v2** - Legacy API, maintained for backwards compatibility
- **v3** - Current API with full feature set

### v2 is a Translation Layer

v2 is implemented as a translation layer that converts requests/responses to v3 format internally:
\`\`\`
V2 API Request → V2ResourceImpl → V2ApiUtil conversions → Shared Storage → V3 DTOs
\`\`\`

### Maintenance Burden

| Component | Lines of Code |
|-----------|---------------|
| v2 API Layer | ~3,500 |
| v2 Bean Classes | ~1,000 |
| V2ApiUtil Conversions | ~370 |
| v2 Integration Tests | ~2,000 |
| **Total** | **~9,500 lines (~12-15% of app/src)** |

### Feature Comparison

**v2-Only (Legacy):**
- Artifact ownership API
- Artifact state management (replaced by version state in v3)
- Labels/Properties duality (unified to labels in v3)

**v3-Only (New):**
- Branches support
- Group-level rules
- MCP/A2A protocol support
- Enhanced exports
- Version content mutation (draft mode)

### Phase 1: Announce Deprecation (v3.3 or v3.4)
- Add deprecation notice to v2 API documentation
- Add HTTP \`Deprecation\` header to v2 responses
- Log usage metrics for v2 endpoints
- Publish migration guide (v2 → v3)

### Phase 2: Soft Deprecation (v3.x - 6 months)
- Add warning logs when v2 endpoints are called
- Prominently feature v3 in all documentation
- Update SDK defaults to v3

### Phase 3: Hard Deprecation (v4.0 - 12 months from Phase 1)
- Mark v2 as "legacy, unsupported"
- Disable v2 by default (opt-in via configuration)
- Continue to include v2 code for compatibility

### Phase 4: Removal (v5.0 - 18-24 months from Phase 1)
- Remove v2 API entirely
- Remove v2 bean classes
- Remove V2ApiUtil
- Remove v2 tests

## Questions for Discussion

1. **Is the proposed timeline reasonable?** (18-24 months total)
2. **Are there critical v2-only features** that should be added to v3 before deprecation?
3. **What migration tooling would help?** (Automated client updates, compatibility shims?)
[...]

PR #9390, opened 2026-08-08
Deprecate RegistryV2ClientFactory and add v2→v3 SDK migration guide

riya-saharan:
### What
Partially addresses #7336.

- Marks `RegistryV2ClientFactory` (class and `create()` method) as `@Deprecated`, with Javadoc pointing to the v3 `RegistryClientFactory` and linking #7330 / #7336.
- Adds `java-sdk/MIGRATING_FROM_V2.md`, a migration guide covering the Maven coordinate change, factory/import swap, and notes on v2→v3 API semantic differences.

### Notes for reviewers
- The repo structure has moved since this issue was filed: there is no longer a standalone `java-sdk-v2/` module. The v2 client now lives at `java-sdk/client-v2/`, alongside `java-sdk/client` (v3), sharing `java-sdk/common`.
- `RegistryV2ClientFactory` is still used in production code by `schema-resolver` (`RegistryClientFacadeFactory`), not just tests. That call site will need to be migrated as part of the eventual module removal, once REST API v2 itself is removed per #7330.
- Full removal of `client-v2` is intentionally out of scope for this PR — it's blocked on the REST API v2 removal timeline in #7330.


java-sdk/MIGRATING_FROM_V2.md
```diff
@@ -0,0 +1,104 @@
...
+removal tracking issue).
+
+All new usage should target the `client` module (REST API v3 /
+`RegistryClientFactory`). This guide covers the practical steps to migrate.
+
+## 1. Update your Maven/Gradle dependency
+
+**Before (v2 SDK):**
+
+\`\`\`xml
```
paoloantinori: All eight fenced code blocks in this guide (lines 18, 24, 28, 34, 44, 51, 55, 62) are written with backslash-escaped backticks. GitHub renders a backslash-escaped backtick as a literal character rather than a code-fence delimiter, so none of the XML dependency snippets or the Java factory/import examples open as code blocks. They collapse into prose or get swallowed as unknown HTML, which breaks exactly the snippets a migrator needs. Replace the escaped backticks with plain triple backticks.

java-sdk/MIGRATING_FROM_V2.md
```diff
@@ -0,0 +1,104 @@
...
+    <groupId>io.apicurio</groupId>
+    <artifactId>apicurio-registry-java-sdk</artifactId>
+    <version>${apicurio-registry.version}</version>
+</dependency>
+\`\`\`
+
+## 2. Swap the factory and client imports
+
+The v3 SDK mirrors the v2 SDK's factory API, so in most cases this is a
+find-and-replace of imports plus the factory class name — the
```
paoloantinori: This paragraph frames the move as "a find-and-replace of imports plus the factory class name" with RegistryClientOptions "unchanged," but the base URL must also change from /apis/registry/v2 to /apis/registry/v3. The example below and the table (line 70) show it, but the prose doesn't list it as a required step, so a reader who keeps their existing options line will miss it. The v3 URL normalizer in RegistryClientRequestAdapterFactory appends /apis/registry/v3 to a URL that doesn't already match, so a stale v2 base URL can become a doubled path that 404s instead of failing fast (worth confirming against the factory source). Make the URL change an explicit third step.

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
```diff
@@ -10,16 +10,26 @@
...
+ *
+ * @deprecated REST API v2 is deprecated and the {@code client-v2} SDK module will be
+ *             removed in a future release once REST API v2 is removed (see
+ *             <a href="https://github.com/Apicurio/apicurio-registry/issues/7330">#7330</a>
+ *             and <a href="https://github.com/Apicurio/apicurio-registry/issues/7336">#7336</a>).
+ *             Use {@code io.apicurio.registry.client.RegistryClientFactory} from the
+ *             {@code java-sdk/client} (v3) module instead. See the SDK migration guide
+ *             for details on migrating from the v2 SDK to the v3 SDK.
  */
+@Deprecated(since = "3.4.0") // NOSONAR - removal tracked in https://github.com/Apicurio/apicurio-registry/issues/7336
```
paoloantinori: This factory is still called by in-repo production code, schema-resolver's RegistryClientFacadeFactory (noted in the PR description), not only tests. Marking it @Deprecated means this project's own build now emits a deprecation warning at that call site. Either add @SuppressWarnings("deprecation") at that caller now, or file a follow-up under #7336 so the build stays clean. Leaving it as-is ships a warning in code this repo owns.

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
```diff
@@ -10,16 +10,26 @@
...
  * the creation logic and provides a unified method for creating clients with different
  * authentication configurations using {@link RegistryClientOptions}.
+ *
+ * @deprecated REST API v2 is deprecated and the {@code client-v2} SDK module will be
+ *             removed in a future release once REST API v2 is removed (see
+ *             <a href="https://github.com/Apicurio/apicurio-registry/issues/7330">#7330</a>
+ *             and <a href="https://github.com/Apicurio/apicurio-registry/issues/7336">#7336</a>).
+ *             Use {@code io.apicurio.registry.client.RegistryClientFactory} from the
+ *             {@code java-sdk/client} (v3) module instead. See the SDK migration guide
+ *             for details on migrating from the v2 SDK to the v3 SDK.
```
paoloantinori: The class Javadoc says "See the SDK migration guide for details" but doesn't link to it or name its path (java-sdk/MIGRATING_FROM_V2.md), so the IDE deprecation warning isn't actionable on its own. Separately, the method-level tag at line 30 (@deprecated see {@link RegistryV2ClientFactory}) links back to this same class, which is circular. Point both at the v3 factory and the guide file path.

paoloantinori (commented):
The migration guide this PR adds doesn't render correctly: every fenced code block uses backslash-escaped backticks, so none of the XML/Java examples open as code on GitHub, which defeats the PR's main purpose. The Java change itself is fine, with correct @Deprecated/Javadoc mechanics, and the guide's factual claims (factory FQN, generated-client packages, Maven artifact IDs, shared options class, base paths) verify against the source.

- MIGRATING_FROM_V2.md:18 (also 24, 28, 34, 44, 51, 55, 62): all eight code fences are written with backslash-escaped backticks. GitHub treats a backslash-escaped backtick as a literal character, not a fence delimiter, so none of the dependency POM or factory/import Java samples render as code. Use plain triple backticks.
- MIGRATING_FROM_V2.md:39-40: section 2 calls this "a find-and-replace of imports plus the factory class name" and says RegistryClientOptions is "unchanged," but the base URL must also move from /apis/registry/v2 to /apis/registry/v3. The prose omits it as a required step; a stale v2 URL can silently double the path and 404 instead of failing fast. List the URL change explicitly.
- RegistryV2ClientFactory.java:22: this factory is still used by in-repo production code (schema-resolver's RegistryClientFacadeFactory), so this deprecation now fires a warning in this project's own build. Suppress it at that caller or track the follow-up under #7336.
- RegistryV2ClientFactory.java:20, 30: the class Javadoc says "see the SDK migration guide" with no link or path to java-sdk/MIGRATING_FROM_V2.md, and the method-level @deprecated self-links to this same class. Add a concrete pointer so the IDE warning is actionable.
- RegistryV2ClientFactory.java:22, 32: confirm `since = "3.4.0"` matches the upcoming release version in the root POM; I couldn't verify the target version from this diff.

java-sdk/MIGRATING_FROM_V2.md
riya-saharan: Fixed — replaced all escaped backticks with plain triple backticks in 0bee95ad5. All eight fences now render correctly.

java-sdk/MIGRATING_FROM_V2.md
riya-saharan: Fixed — added a new "3. Update the base URL" section explaining this is a required step, including the RegistryClientRequestAdapterFactory normalization behavior and the doubled-path failure mode you flagged.

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
riya-saharan: Fixed — added @SuppressWarnings("deprecation") on create_v2() in RegistryClientFacadeFactory, with a comment linking to #7336 for the eventual removal.

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
riya-saharan: Fixed — both the class-level and method-level @deprecated tags now link directly to java-sdk/MIGRATING_FROM_V2.md on GitHub, and the method-level tag points at RegistryClientFactory#create(RegistryClientOptions) instead of back at itself.

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
```diff
@@ -10,16 +10,26 @@
...
  */
+@Deprecated(since = "3.4.0")
```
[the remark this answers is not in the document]
riya-saharan: Dropped since entirely rather than guess at an unreleased version — using plain @Deprecated on both the class and method until the target release version is confirmed.

riya-saharan:
Pushed 0bee95a addressing all review feedback: fixed the escaped-backtick code fences, added an explicit "Update the base URL" step to the migration guide, suppressed the deprecation warning at the schema-resolver call site, fixed the Javadoc links to point at the guide and the v3 factory instead of self-referencing, and dropped the unverified since = "3.4.0" version tag. Ready for another look.

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
```diff
@@ -10,17 +10,31 @@
...
+ *             removed in a future release once REST API v2 is removed (see
+ *             <a href="https://github.com/Apicurio/apicurio-registry/issues/7330">#7330</a>
+ *             and <a href="https://github.com/Apicurio/apicurio-registry/issues/7336">#7336</a>).
+ *             Use {@code io.apicurio.registry.client.RegistryClientFactory} from the
+ *             {@code java-sdk/client} (v3) module instead. See
+ *             <a href="https://github.com/Apicurio/apicurio-registry/blob/main/java-sdk/MIGRATING_FROM_V2.md">java-sdk/MIGRATING_FROM_V2.md</a>
+ *             for step-by-step instructions on migrating from the v2 SDK to the v3 SDK.
  */
-public final class RegistryV2ClientFactory {
+@Deprecated
```
aclfe: I think this wants forRemoval = true rather than a bare @Deprecated?

That's actually the one open Sonar issue on this PR, java:S6355 on this exact line, "Add 'since' and/or 'forRemoval' arguments to the @Deprecated annotation". The NOSONAR on line 24 doesn't cover it since the rule anchors on the annotation line, not the declaration.

forRemoval also sidesteps the version guess that got since = "3.4.0" dropped, and it's more honest about what #7336 is tracking.

On the NOSONAR comments, they're the only two in the repo right now, and they suppress every rule on the line rather than just this one. I'd drop both and let forRemoval close it properly.

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
```diff
@@ -10,17 +10,31 @@
...
+public final class RegistryV2ClientFactory { // NOSONAR - removal tracked in https://github.com/Apicurio/apicurio-registry/issues/7336
 
     /**
      * Creates a RegistryClient using the provided options.
      *
      * @param options the configuration options for the client
      * @return a new RegistryClient instance
      * @throws IllegalArgumentException if options are invalid
+     * @deprecated Use {@link io.apicurio.registry.client.RegistryClientFactory#create(RegistryClientOptions)}
```
aclfe: This link doesn't resolve. client-v2 has no dependency on the client module and RegistryClientFactory only ships in apicurio-registry-java-sdk, so javadoc gives "reference not found" here and the generated HTML renders it as a literal io.apicurio.registry.client.RegistryClientFactory#create(RegistryClientOptions) with no anchor.

Build doesn't fail because the javadoc plugin runs with -Xdoclint:none and failOnError=false, but this is the one place the IDE hint actually needs to work?

Line 18 already handles it correctly, same fix here:

```
{@code RegistryClientFactory.create(RegistryClientOptions)}
```

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
```diff
@@ -173,6 +173,10 @@ private static RegistryClientFacade create_v3(SchemaResolverConfig config, Vertx
         return new RegistryClientFacadeImpl(client, config.getRegistryUrl(), config.getUsageTelemetryClientId());
     }
 
+    // Suppressed: RegistryV2ClientFactory is intentionally still used here to support
+    // REST API v2 registries. Removal of this call site is tracked alongside the
+    // RegistryV2ClientFactory removal in https://github.com/Apicurio/apicurio-registry/issues/7336.
+    @SuppressWarnings("deprecation")
```
aclfe: If you move to forRemoval = true this needs to become {"deprecation", "removal"}. "deprecation" on its own doesn't suppress [removal] warnings, and unlike -Xlint:deprecation, -Xlint:removal is on by default, so it fires even with showDeprecation=false set in the root pom.

Also this isn't the only caller. 21 calls across six files, all untouched:

app/src/test/java/io/apicurio/registry/AbstractClientFacadeTestBase.java:32
app/src/test/java/io/apicurio/registry/auth/SimpleAuthTest.java:164
app/src/test/java/io/apicurio/registry/tls/TlsSdkClientTest.java
app/src/test/java/io/apicurio/registry/tls/MtlsSdkClientTest.java
app/src/test/java/io/apicurio/registry/tls/TlsSdkClientJdkAdapterTest.java
app/src/test/java/io/apicurio/registry/tls/MtlsSdkClientJdkAdapterTest.java

java-sdk/MIGRATING_FROM_V2.md
```diff
@@ -0,0 +1,123 @@
+# Migrating from the Java SDK v2 (REST API v2) to the Java SDK v3
```
aclfe, line 1: One structural thing. I think this overlaps docs/modules/ROOT/pages/getting-started/assembly-migrating-registry-v2-v3.adoc, which already has a "Migrating REST API and SDK client applications" section covering the dependency swap, the factory swap and a create() mapping table. That one is in nav.adoc so it publishes to the docs site. This file sits outside the docs module and nothing links to it except the javadoc you added.

What worries me more is that the two disagree on what "the v2 SDK" is. The adoc says apicurio-registry-client 2.6.13.Final, which is the 2.x product client. This says apicurio-registry-v2-java-sdk, which is the client-v2 module in this repo. Both right for their own reader, but with no cross reference someone follows the wrong set of steps?

Minimum I'd add an xref in both directions. Better would be folding this in as a section of the existing assembly so it actually ships.

java-sdk/MIGRATING_FROM_V2.md
```diff
@@ -0,0 +1,123 @@
...
+<dependency>
+    <groupId>io.apicurio</groupId>
+    <artifactId>apicurio-registry-java-sdk</artifactId>
+    <version>${apicurio-registry.version}</version>
+</dependency>
+```
+
+## 2. Swap the factory and client imports
+
+The v3 SDK mirrors the v2 SDK's factory API, so this is a find-and-replace of
```
aclfe: This sits a bit at odds with section 4, which says v3 isn't a drop in replacement. The generated clients are different types too, io.apicurio.registry.rest.client.v2.RegistryClient vs io.apicurio.registry.rest.client.RegistryClient with different request builders, so find-and-replace really only holds for the two import lines. Might be worth softening?

aclfe (commented):
Thanks for the PR. Direction looks right to me, and holding the actual module removal for #7330 seems like the correct call.

Few things I think want another pass. The main two are the annotation itself and the javadoc link not resolving.

Section 3 checks out by the way, I traced it against RegistryClientRequestAdapterFactory and the doubled path is real. normalizeRegistryUrl defaults to true and the v3 branch only matches on the v3 suffix, so a stale v2 URL does become .../apis/registry/v2/apis/registry/v3 and 404s at call time. Good catch, worth keeping.

Separate from the code, neither commit has a Signed-off-by line, and the messages aren't in the conventional commits format the repo uses.

java-sdk/MIGRATING_FROM_V2.md
```diff
@@ -0,0 +1,123 @@
...
+- Rely on v2-specific endpoints that have no v3 equivalent, or that were
+  renamed/restructured in v3.
+- Parse response payloads directly (rather than through generated model
+  classes) — some field names/shapes differ.
+
+Consult the REST API v3 OpenAPI spec (or the generated model classes in the
+`client` module) for the authoritative shape of each endpoint, and adjust call
+sites accordingly.
+
+## 5. Update authentication configuration
```
aclfe: Heading says update your auth config, body says nothing needs updating. Maybe just fold this into section 3?

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
riya-saharan: Fixed switched both annotations to `@Deprecated(forRemoval = true)` and removed both `NOSONAR` comments. Confirmed via SonarCloud that this closes java:S6355 cleanly (0 new issues on the latest push). Thank you for pointing out that the NOSONAR was anchored wrong and masking rather than resolving the rule.

java-sdk/client-v2/src/main/java/io/apicurio/registry/client/RegistryV2ClientFactory.java
riya-saharan: Fixed — switched to {@code RegistryClientFactory.create(RegistryClientOptions)}, matching the pattern already used in the class-level Javadoc.

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
riya-saharan: Fixed, with one adjustment: trimmed to @SuppressWarnings("removal") alone rather than {"deprecation", "removal"}. Once RegistryV2ClientFactory carries forRemoval = true, the only deprecated element being called is forRemoval, so the call site triggers a removal warning but never a plain deprecation warning — the IDE flagged "deprecation" as an unused suppression. 

Applied the same reasoning at all 22 call sites (RegistryClientFacadeFactory plus AbstractClientFacadeTestBase, SimpleAuthTest, TlsSdkClientTest, MtlsSdkClientTest, TlsSdkClientJdkAdapterTest, MtlsSdkClientJdkAdapterTest), each suppressed at the method level rather than the class. 

Open to switching back to the two-value form if there's a case I'm missing.

java-sdk/MIGRATING_FROM_V2.md
riya-saharan: Confirmed the two docs cover genuinely different artifacts: apicurio-registry-client (2.6.13.Final, the legacy product client) vs. apicurio-registry-v2-java-sdk (the client-v2 module this guide covers). 
Went with the two-way xref rather than folding this into the existing assembly, since combining guidance for two different SDKs into one procedure seemed more likely to confuse readers than help them. 
Added a NOTE in assembly-migrating-registry-v2-v3.adoc pointing here, and a note at the top of MIGRATING_FROM_V2.md pointing back.

java-sdk/MIGRATING_FROM_V2.md
riya-saharan: Fixed reworded section 2 to scope the "mechanical" claim to just the factory swap (the two imports and the class name), and noted that the generated RegistryClient types differ with different request builders, pointing to section 4 for the details.

java-sdk/MIGRATING_FROM_V2.md
riya-saharan: Fixed folded the auth-config content into section 3 as a closing note and removed section 5.

riya-saharan:
Pushed 548a3e8 addressing all review feedback from this round:

- @Deprecated(forRemoval = true) on both the class and method, both NOSONAR comments removed
- Fixed the cross-module Javadoc link ({@link} → {@code})
- @SuppressWarnings("removal") applied at each of the 22 affected call sites (schema-resolver plus six TLS/auth test files), scoped at the method level — see the suppression thread for why this is "removal" alone rather than the two-value form
- Softened the "find-and-replace" framing in section 2 of the migration guide
- Folded the auth-config section into the base-URL section
- Added a two-way cross-reference between this guide and the existing REST API v2→v3 migration doc, since they cover two different v2 clients
- Squashed both commits into one, signed off per DCO, reformatted to docs(java-sdk): ...

Branch is up to date with main, the full java-sdk reactor build compiles clean, and SonarCloud's Quality Gate passed with 0 new issues.

riya-saharan:
Pushed a small follow-up addressing the branch-pinned link Copilot flagged in the last review: the 3.2.x reference in MIGRATING_FROM_V2.md now points to main instead, so the cross-reference stays correct across future releases and backports.

Left the two @SuppressWarnings("removal") scoping suggestions (in AbstractClientFacadeTestBase.java and RegistryClientFacadeFactory.java) as-is for now — narrowing those would mean splitting out helper methods across the 8 affected test files plus the schema-resolver call site, which felt like a bigger change than warranted for a non-blocking style nit. Happy to do that as a fast-follow if a maintainer would prefer it tightened before merge.

riya-saharan:
Hi @EricWittmann — just checking in on this PR since it's been a few days since assignment. All review feedback from @paoloantinori and @aclfe has been addressed, and I've just synced the branch with main (no conflicts). Quality Gate was passing before the sync — happy to make any further changes if needed. No rush, just wanted to flag it in case it fell off the radar. Thanks!

paoloantinori:
Thanks for the nudge. My earlier point (the escaped-backtick code fences in the migration guide not rendering) is resolved in the current head: the XML/Java blocks render as fenced code now. I don't have anything further from my side; leaving the merge call to @EricWittmann.

riya-saharan:
Hi @EricWittmann, following up once more on this PR. I've addressed all review feedback from @paoloantinori and @aclfe, and I'm currently looking into the License Compliance check that's still failing.

I have an LFS-related deadline coming up soon, so any update - a review, guidance on what's still needed, or a rough timeline would really help me plan. Thanks so much for your time on this!

paoloantinori:
On the License Compliance check: I looked at the commit statuses, and that check is failing identically on the latest commit of `main` (same "4 issues found" at 0e558636), so it does not look like something this PR introduced or can fix — it predates your changes. Worth flagging to the maintainers as an infra-level issue rather than blocking on it here.

From my side everything I raised is addressed at the current head (the migration guide fences render correctly, and I re-checked today). The remaining call on merging is with @EricWittmann.

riya-saharan:
Opened #9622   to track the pre-existing License Compliance failure on `main` separately, per @paoloantinori's finding above.

paoloantinori:
Thanks for splitting that out — #9622 is the right place to track it, since the failure reproduces on the latest commit of main independently of this PR. That keeps the remaining red check here clearly out of scope. Nothing else from my side; the deprecation and the migration guide look good on the current head.

riya-saharan:
@paoloantinori, on a separate note — I'm currently taking part in the LFX Mentorship program, and this PR is part of my contribution for it. If you have some time, would you be able to take a look at my proposal? I'd really value your feedback and it would mean a lot for my mentorship progress. Thank you so much!

paoloantinori:
Hi @riya-saharan, happy to. I took a fresh look at the current head (cde4ed7).

The earlier rounds are closed from my side: the migration guide renders correctly now, `@Deprecated(forRemoval = true)` is in place on both the class and `create()`, and the `@SuppressWarnings("removal")` annotations in the tests and in `RegistryClientFacadeFactory` are scoped to the right places, with a clear comment on why the production call site stays.

Two specifics I checked on this pass:

- The "Update the base URL" section is a good addition. I verified the mechanism in `RegistryClientRequestAdapterFactory.validateAndNormalizeRegistryUrl`: with `normalizeRegistryUrl = true` (the default), the v3 factory appends `apis/registry/v3` unless the URL already matches the v3 pattern, and it does not strip a stale `/v2` suffix. So the doubled-path failure mode the guide warns about is real, and calling it a required step is the right call.
- The cross-links between the product docs page and `MIGRATING_FROM_V2.md`, each pointing at the other for the two different legacy clients, read clearly.

Overall the proposal is in good shape and matches what the tracking issues (#7330 / #7336) call for at this stage: deprecate and document, defer the module removal itself. What's left is a final approval and the merge; the License Compliance failure is tracked separately in #9622 since it reproduces on main.

Good luck with the rest of the mentorship. The iteration across several review rounds here, and splitting the pre-existing CI failure into its own issue, is exactly the kind of contribution record those programs value.

riya-saharan:
Thank you so much, @paoloantinori — I really appreciate you taking the time for a fresh full pass, and especially for independently verifying the base-URL/doubled-path mechanism against the source rather than just taking the guide's word for it. Your feedback across all the rounds made this a much stronger PR, and I've learned a lot from the process.

Really grateful for your support with the mentorship proposal too. Thank you!

paoloantinori:
You are welcome, and thank you for sticking with it across all the rounds. The guide is much stronger for it, and the way you split the pre-existing License Compliance failure out to #9622 was the right call.

Happy to support the mentorship proposal when the time comes. Good luck with the program!

riya-saharan:
/unstale

Still ready — all reviewer feedback from @paoloantinori and @aclfe is addressed as of the current head, and the License Compliance failure (tracked separately in #9622, pre-existing on main) is now resolved. Branch is current with main, checks are green.

@EricWittmann, since you're assigned here — would you have a chance to take a look when you get a moment?

EricWittmann (approved):
LGTM

EricWittmann:
Thanks @riya-saharan :+1: 

Sorry it's taken me so long to get to this.

PR #9391, opened 2026-08-08
fix(prompt): add {{#if}}/{{#unless}}/{{else}} support to PromptRender…

### LINKED ISSUE #8728: fix(app): prompt template render endpoint leaves {{#if}} conditional blocks unrendered

Prathamesh09-coder:
fix(prompt): add {{#if}}/{{#unless}}/{{else}} conditional block support to PromptRenderingService…ingService

The REST render endpoint (POST .../versions/{version}/render) processed only plain {{variable}} substitution via substituteVariables(). The MCP render path (PromptTemplateConverter.renderTemplate()) additionally calls processConditionalBlocks(), which supports Handlebars-style:

  {{#if var}} ... {{/if}}
  {{#if var}} ... {{else}} ... {{/if}}
  {{#unless var}} ... {{/unless}}

This divergence meant the same registered PROMPT_TEMPLATE artifact produced different output depending on whether it was rendered through the REST API (and the UI 'Test Prompt' panel, which calls it) or through the MCP server.

This commit ports the conditional block logic into PromptRenderingService:

- Adds IF_BLOCK_PATTERN, IF_ELSE_BLOCK_PATTERN and UNLESS_BLOCK_PATTERN compiled Pattern constants.
- Adds isTruthy() helper with the## Summary
- 
- The REST render endpoint (`POST .../versions/{version}/render`) processed only plain `{{variable}}` substitution. The MCP render path (`PromptTemplateConverter.renderTemplate()`) additionally supports Handlebars-style conditional blocks via `processConditionalBlocks()`. This divergence caused the same registered `PROMPT_TEMPLATE` artifact to produce different output depending on the render path - a bug reported in #8728.
- 
- ## Changes
- 
- - Added three compiled `Pattern` constants to `PromptRenderingService`:
-   - `IF_BLOCK_PATTERN` - matches `{{#if var}} ... {{/if}}`
-   - `IF_ELSE_BLOCK_PATTERN` - matches `{{#if var}} ... {{else}} ... {{/if}}`
-   - `UNLESS_BLOCK_PATTERN` - matches `{{#unless var}} ... {{/unless}}`
- - Added `isTruthy(Object value)` helper with identical truthy semantics to the MCP path:
-   - falsy: `null`, `Boolean.FALSE`, empty `String`
-   - truthy: everything else
- - Added `processConditionalBlocks(String template, Map<String, Object> variables)` that resolves all three block forms before plain variable substitution, so variables *inside* conditional branches are also substituted correctly
- - Updated `render()` to call `processConditionalBlocks()` first, then `substituteVariables()`
- - Updated class Javadoc to document all supported syntax forms
- 
- **Backward compatible:** templates with no conditional blocks are unaffected.
- 
- ## Tests Added (9 new tests in `PromptRenderingServiceTest`)
- 
- | Test | What it covers |
- |------|---------------|
- | `testIfBlockRenderedWhenVariableTruthy` | `{{#if}}` with truthy value renders block |
- | `testIfBlockOmittedWhenVariableFalsy` | `{{#if}}` with `false` removes block |
- | `testIfBlockOmittedWhenVariableMissing` | `{{#if}}` with missing var (null) removes block |
- | `testIfBlockOmittedWhenVariableEmptyString` | `{{#if}}` with `""` removes block |
- | `testUnlessBlockRenderedWhenVariableFalsy` | `{{#unless}}` shown when variable is falsy |
- | `testUnlessBlockOmittedWhenVariableTruthy` | `{{#unless}}` hidden when variable is truthy |
- | `testIfElseBlockTruthyBranch` | `{{#if}}...{{else}}...{{/if}}` takes truthy branch |
- | `testIfElseBlockFalsyBranch` | `{{#if}}...{{else}}...{{/if}}` takes falsy branch |
- | `testConditionalBlockCombinedWithVariableSubstitution` | Mixed conditionals + `{{var}}` in same template |
- 
- ## Related Issue
- Fixes #8728## Summary
- 
- The REST render endpoint (`POST .../versions/{version}/render`) processed only plain `{{variable}}` substitution. The MCP render path (`PromptTemplateConverter.renderTemplate()`) additionally supports Handlebars-style conditional blocks via `processConditionalBlocks()`. This divergence caused the same registered `PROMPT_TEMPLATE` artifact to produce different output depending on the render path  same truthy semantics as MCP: a value is falsy when null, Boolean.FALSE, or empty string; everything else is truthy.
- Adds processConditionalBlocks() that resolves if/else/unless blocks in one pass before plain variable substitution, matching evaluation order in PromptTemplateConverter.
- Updates render() to call processConditionalBlocks() first so that variables inside conditional branches are substituted correctly.
- Updates the class Javadoc to document all supported syntax forms.
- Adds 9 new test cases in PromptRenderingServiceTest covering all block types, truthy/falsy semantics, missing/empty-string variables, and combined conditional + variable substitution in the same template.

Fixes #8728

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -43,21 +43,21 @@ public class PromptRenderingService {
      * Group 1 = variable name, Group 2 = inner content.
      */
     private static final Pattern IF_BLOCK_PATTERN =
-            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{/if\\}\\}");
+            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}((?:(?!\\{\\{/if\\}\\})[\\s\\S])*?)\\{\\{/if\\}\\}");
 
     /**
-     * Matches {{#if var}} ... {{else}} ... {{/if}} blocks.
+     * Matches if-else blocks: opening tag, truthy branch, else tag, falsy branch, closing tag.
      * Group 1 = variable name, Group 2 = truthy branch, Group 3 = falsy branch.
      */
     private static final Pattern IF_ELSE_BLOCK_PATTERN =
-            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{else\\}\\}([\\s\\S]*?)\\{\\{/if\\}\\}");
+            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}((?:(?!\\{\\{else\\}\\})[\\s\\S])*?)\\{\\{else\\}\\}((?:(?!\\{\\{/if\\}\\})[\\s\\S])*?)\\{\\{/if\\}\\}");
 
     /**
-     * Matches {{#unless var}} ... {{/unless}} blocks.
+     * Matches unless blocks: opening tag, inner content, closing tag.
      * Group 1 = variable name, Group 2 = inner content.
      */
     private static final Pattern UNLESS_BLOCK_PATTERN =
-            Pattern.compile("\\{\\{#unless\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{/unless\\}\\}");
+            Pattern.compile("\\{\\{#unless\\s+(\\w+)\\}\\}((?:(?!\\{\\{/unless\\}\\})[\\s\\S])*?)\\{\\{/unless\\}\\}");
 
     /**
      * Renders a prompt template by substituting variables.
@@ -394,7 +394,7 @@ public class PromptRenderingService {
     private boolean isTruthy(Object value) {
         return value != null
                 && !Boolean.FALSE.equals(value)
-                && !(value instanceof String && ((String) value).isEmpty());
+                && !(value instanceof String s && s.isEmpty());
     }
 
     /**
@@ -408,7 +408,7 @@ public class PromptRenderingService {
      * </ol>
      */
     private String processConditionalBlocks(String template, Map<String, Object> variables) {
-        // 1. {{#if var}} ... {{else}} ... {{/if}}  (must run before plain {{#if}} to avoid partial match)
+        // Step 1: resolve if-else blocks first to prevent the plain-if pattern from partially matching them
         Matcher ifElseMatcher = IF_ELSE_BLOCK_PATTERN.matcher(template);
         StringBuffer ifElseResult = new StringBuffer();
         while (ifElseMatcher.find()) {
@@ -421,7 +421,7 @@ public class PromptRenderingService {
         ifElseMatcher.appendTail(ifElseResult);
         template = ifElseResult.toString();
 
-        // 2. {{#if var}} ... {{/if}}
+        // Step 2: resolve plain if blocks
         Matcher ifMatcher = IF_BLOCK_PATTERN.matcher(template);
         StringBuffer ifResult = new StringBuffer();
         while (ifMatcher.find()) {
@@ -433,7 +433,7 @@ public class PromptRenderingService {
         ifMatcher.appendTail(ifResult);
         template = ifResult.toString();
 
-        // 3. {{#unless var}} ... {{/unless}}
+        // Step 3: resolve unless blocks (logical complement of if)
         Matcher unlessMatcher = UNLESS_BLOCK_PATTERN.matcher(template);
         StringBuffer unlessResult = new StringBuffer();
         while (unlessMatcher.find()) {
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -13,15 +13,51 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static io.apicurio.registry.util.JsonObjectMapper.MAPPER;
 import static io.apicurio.registry.util.YAMLObjectMapper.YAML_MAPPER;
 
 /**
- * Service for rendering prompt templates by substituting variables.
+ * Service for rendering prompt templates by substituting variables and evaluating
+ * Handlebars-style conditional blocks.
+ *
+ * <p>Supported syntax:</p>
+ * <ul>
+ *   <li>{@code {{variable}}} — substituted with the variable value, or kept verbatim when absent</li>
+ *   <li>{@code {{#if variable}} ... {{/if}}} — block rendered when variable is truthy</li>
+ *   <li>{@code {{#if variable}} ... {{else}} ... {{/if}}} — if/else branch</li>
+ *   <li>{@code {{#unless variable}} ... {{/unless}}} — block rendered when variable is falsy</li>
+ * </ul>
+ *
+ * <p>Truthy semantics (matching the MCP render path): a value is truthy when it is
+ * non-null, not {@code Boolean.FALSE}, and not an empty string.</p>
  */
 @ApplicationScoped
 public class PromptRenderingService {
+
+
+    /**
+     * Matches {{#if var}} ... {{/if}} blocks, including multi-line content.
+     * Group 1 = variable name, Group 2 = inner content.
+     */
+    private static final Pattern IF_BLOCK_PATTERN =
+            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{/if\\}\\}");
+
+    /**
+     * Matches {{#if var}} ... {{else}} ... {{/if}} blocks.
+     * Group 1 = variable name, Group 2 = truthy branch, Group 3 = falsy branch.
+     */
+    private static final Pattern IF_ELSE_BLOCK_PATTERN =
+            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{else\\}\\}([\\s\\S]*?)\\{\\{/if\\}\\}");
+
+    /**
+     * Matches {{#unless var}} ... {{/unless}} blocks.
+     * Group 1 = variable name, Group 2 = inner content.
+     */
+    private static final Pattern UNLESS_BLOCK_PATTERN =
+            Pattern.compile("\\{\\{#unless\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{/unless\\}\\}");
 
     /**
      * Renders a prompt template by substituting variables.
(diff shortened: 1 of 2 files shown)


EricWittmann:
@Prathamesh09-coder I think this is the direction we want to go :+1: - can I ask why you marked it as a Draft?  Please click Ready for review!

Nachiket-Roy:
I suspect this need #9461 to be resolved!

Prathamesh09-coder:
@Nachiket-Roy Thanks for flagging #9461! You're right that the current conditional-block implementation doesn't handle nested {{#if}} blocks. Since #9391 is primarily intended to bring REST rendering in line with the existing MCP behavior, would you prefer me to address nested conditionals as part of this PR, or keep #9391 scoped to the original issue and handle the nesting behavior separately in #9461?

Prathamesh09-coder:
Thanks @EricWittmann! I initially kept it as Draft while validating the implementation and addressing the review feedback. Nachiket also raised #9461 regarding nested conditionals, so I'm clarifying whether that should be included in this PR. Once that's resolved, I'll mark it Ready for Review.

EricWittmann:
Please mark it as ready for review.  Any additional changes to improve the impl (like handling nested blocks) can be done in a followup.  Also please rebase and fix the merge conflicts.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -347,6 +384,69 @@ private String getTypeName(Object value) {
...
+     *
+     * <p>Handles three forms in evaluation order:</p>
+     * <ol>
+     *   <li>{@code {{#if var}} ... {{else}} ... {{/if}}} — if/else (must be matched before plain if)</li>
+     *   <li>{@code {{#if var}} ... {{/if}}} — simple if</li>
+     *   <li>{@code {{#unless var}} ... {{/unless}}} — unless (logical complement of if)</li>
+     * </ol>
+     */
+    private String processConditionalBlocks(String template, Map<String, Object> variables) {
+        // Step 1: resolve if-else blocks first to prevent the plain-if pattern from partially matching them
```
paoloantinori, line 411: Block resolution is a single non-recursive pass over each pattern, so nested blocks are mishandled. For `A{{#if outer}}{{#if inner}}X{{/if}}{{/if}}B`, the outer match consumes `{{#if outer}}{{#if inner}}X{{/if}}` with group 2 equal to `{{#if inner}}X`: when `outer` is falsy a stray `{{/if}}` is emitted, and when `outer` is truthy the inner `{{#if inner}}` is left as literal text instead of being evaluated. Either loop `processConditionalBlocks` until the template stops changing (with a depth guard), or explicitly document that nesting is unsupported and strip residual block tags so they cannot leak into the rendered prompt.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -13,16 +13,52 @@
...
 @ApplicationScoped
 public class PromptRenderingService {
 
+
+    /**
+     * Matches {{#if var}} ... {{/if}} blocks, including multi-line content.
+     * Group 1 = variable name, Group 2 = inner content.
+     */
+    private static final Pattern IF_BLOCK_PATTERN =
+            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}((?:(?!\\{\\{/if\\}\\})[\\s\\S])*?)\\{\\{/if\\}\\}");
```
paoloantinori, line 46: The variable capture is `(\w+)` with no surrounding whitespace tolerated, so `{{#if var }}` (trailing space) or `{{#if user.active}}` (dotted name) will not match and will leak verbatim into the rendered prompt. Confirm the name grammar accepted by `substituteVariables()` and align the block patterns to it (likely `[A-Za-z0-9_.-]+` with optional `\s*` around the name). Otherwise the conditional feature silently no-ops on exactly the names the plain-substitution path accepts.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -13,16 +13,52 @@
...
+     * Group 1 = variable name, Group 2 = truthy branch, Group 3 = falsy branch.
+     */
+    private static final Pattern IF_ELSE_BLOCK_PATTERN =
+            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}((?:(?!\\{\\{else\\}\\})[\\s\\S])*?)\\{\\{else\\}\\}((?:(?!\\{\\{/if\\}\\})[\\s\\S])*?)\\{\\{/if\\}\\}");
+
+    /**
+     * Matches unless blocks: opening tag, inner content, closing tag.
+     * Group 1 = variable name, Group 2 = inner content.
+     */
+    private static final Pattern UNLESS_BLOCK_PATTERN =
```
paoloantinori, line 59: The unless branch capture guards only against `{{/unless}}`, not `{{else}}`, so `{{#unless x}}A{{else}}B{{/unless}}` renders the literal text `A{{else}}B`. If unless-else is unsupported, reject or strip the stray `{{else}}`; if it should work, handle it explicitly. At minimum, the class Javadoc should state that `{{else}}` is valid only inside `{{#if}}`.

paoloantinori (commented):
This ports the MCP conditional-block rendering into the REST path, and the core mechanics are correct for flat (non-nested) templates with simple variable names; the truthy/falsy/missing/empty test coverage is solid. The risks are edge cases the single-pass regex model cannot express: nested blocks emit stray tags, the variable-name grammar is narrower than plain substitution, and `{{else}}` leaks inside unless blocks, all of which can push literal template syntax into the rendered prompt with no error.

- Nested conditionals break (`PromptRenderingService.java:411`). `A{{#if outer}}{{#if inner}}X{{/if}}{{/if}}B` yields a dangling `{{/if}}` when `outer` is falsy, and leaves `{{#if inner}}X{{/if}}` unevaluated when `outer` is truthy, because each pattern is applied in one non-recursive pass.
- Variable-name grammar mismatch (`PromptRenderingService.java:46`). The capture `(\w+)` rejects dotted names like `user.active` and trailing whitespace like `{{#if var }}`; if `substituteVariables()` accepts those, the block is left as literal text. The two grammars should be aligned.
- `{{else}}` leaks inside unless blocks (`PromptRenderingService.java:59`). The unless capture guards only against `{{/unless}}`, so `{{#unless x}}A{{else}}B{{/unless}}` renders `A{{else}}B` verbatim.
- Numeric `0` is truthy here (only `null`, `Boolean.FALSE`, and empty string are falsy). That is documented and matches MCP parity, but differs from Handlebars convention, so callers relying on `{{#if count}}` to treat `0` as falsy will be surprised.
- Test naming (`PromptRenderingServiceTest.java:764`): `testConditionalBlockMarkersAreNotSubstituted` now asserts the opposite behavior (markers are processed and removed). Rename it to reflect what it actually checks.

PRAteek-singHWY:
@Prathamesh09-coder heads-up, since I ran into exactly this and it cost me a couple of days.

This was closed by the one-PR-at-a-time bot rather than by a reviewer, and @EricWittmann had already
said above that this is the direction he wants. So the work is fine, it is purely the open PR count.

The part that caught me out: once #9360 is merged or closed, reopening this may not work. I had two
PRs auto-closed the same way (#8921 and #8923) and GitHub refused to reopen either, returning a bare
422 with no message, because both were opened from the same branch against `main` and it will not
reopen one when another shares that head and base pair. If you hit that, the way through is to open a
fresh PR from the same branch and reference this one in the description. Nothing is lost, the commits
and the review history stay linked, it is just an extra step nobody tells you about.

Also worth doing before you reopen or refile: the branch will need a rebase. Eric asked for that
above, and `main` has moved a fair amount since.

On the change itself, one observation rather than a review comment. It adds `processConditionalBlocks`
to `PromptRenderingService` while `PromptTemplateConverter` keeps its own copy, so the two render
paths would each carry their own conditional logic. That is the same split that #8877 described for
variable parsing and that #8977 fixed by moving the pattern into `schema-util/common`. Not a blocker
for this PR at all, and Eric has already said follow-up work is fine, but if you would rather share it
than duplicate it, `PromptTemplateVariableUtil` is the natural neighbour and I am happy to help.

Nachiket-Roy:
> @Prathamesh09-coder heads-up, since I ran into exactly this and it cost me a couple of days.
> 
> This was closed by the one-PR-at-a-time bot rather than by a reviewer, and @EricWittmann had already said above that this is the direction he wants. So the work is fine, it is purely the open PR count.
> 
> The part that caught me out: once #9360 is merged or closed, reopening this may not work. I had two PRs auto-closed the same way (#8921 and #8923) and GitHub refused to reopen either, returning a bare 422 with no message, because both were opened from the same branch against `main` and it will not reopen one when another shares that head and base pair. If you hit that, the way through is to open a fresh PR from the same branch and reference this one in the description. Nothing is lost, the commits and the review history stay linked, it is just an extra step nobody tells you about.
> 
> Also worth doing before you reopen or refile: the branch will need a rebase. Eric asked for that above, and `main` has moved a fair amount since.
> 
> On the change itself, one observation rather than a review comment. It adds `processConditionalBlocks` to `PromptRenderingService` while `PromptTemplateConverter` keeps its own copy, so the two render paths would each carry their own conditional logic. That is the same split that #8877 described for variable parsing and that #8977 fixed by moving the pattern into `schema-util/common`. Not a blocker for this PR at all, and Eric has already said follow-up work is fine, but if you would rather share it than duplicate it, `PromptTemplateVariableUtil` is the natural neighbour and I am happy to help.

The GitHub Action is taking a bit longer than usual to pick up and close the PR. You can ask Eric to reopen the PR if you don't have any other PRs opened or drafted, I believe.

PR #9392, opened 2026-08-08
fix(mcp): apply declared variable defaults when rendering prompt templates

### LINKED ISSUE #9228: Prompt template `default` values are never applied when rendering

Aarti-panchal01:
Follow-up to the review discussion on #9309, and the MCP half of #9228.

## The problem

#9309 fixed the REST render endpoint, but the MCP server does not render through `PromptRenderingService`. All three of its entry points reach `PromptTemplateConverter.renderTemplate(String, Map)`:

- `get_registry_prompt` calls it directly
- `render_prompt_template` and `render_registry_prompt` both reach it through `parseAndRenderAutoDetect`

That method receives only the template text and the caller's arguments. It never sees the variable schema, so a declared `default` could not be applied, and an argument the caller omitted was left in the output as a raw `{{placeholder}}`.

The class already parsed `default` into `VariableSchema.defaultValue` and exposed `getDefaultValue()`, but that getter was not called anywhere in the repository. The value was parsed and then discarded.

The shipped example `examples/llm-artifact-types/sample-schemas/mcp-enabled-code-review-prompt.yaml` shows the effect. It declares `focus_area` as `required: false` with a default of `"code quality, security, and performance"`. An MCP client that omits the argument gets a literal `{{focus_area}}` inside a code review prompt, with no validation error, because the variable is not required.

## The change

Rather than adding a second parsing path, this wires up the field that was already being parsed.

- `applyVariableDefaults(PromptTemplate, Map)` fills in the declared defaults for arguments the caller did not supply.
- A new `renderTemplate(PromptTemplate, Map)` overload applies it before delegating to the existing string overload.
- `parseAndRenderAutoDetect` and the `PromptTemplateMCPServer` call site both use the overload, so a single fix point covers all three entry points.

`renderTemplate(String, Map)` is unchanged and still available for callers that genuinely have only template text.

## Behaviour

The semantics match the REST render endpoint as fixed in #9309:

- caller supplied values always win, and an argument that is explicitly present is never overwritten
- the caller's map is never mutated, and a copy is made only when there is a default to apply
- a `default` declared with no value is treated as absent, so the placeholder is preserved
- defaults now also apply when the arguments map is `null`, where the string overload previously returned the template with no substitution at all

## Tests

Adds `PromptTemplateConverterTest`, the first test coverage this class has had. Eight tests cover defaults applied and overridden, non string defaults, variables declared without a default, defaults inside `{{#if}}` blocks, the caller map not being mutated, the null arguments path, and a template with no variable schema.

One honest note on these. Unlike #9309, they could not be written red first, because the overload they exercise did not exist before this commit. The behaviour they lock in was confirmed by reading the previous call path, where the variable schema was never handed to the renderer at all.

## A note on #9228

Merging #9309 closed #9228 automatically. Its description said `Part of #9228` rather than `Fixes`, but the linked issue in the sidebar was still attached from when that PR was first opened, and that is what closed it. I am not able to reopen it myself. If a maintainer reopens it, merging this PR will close it correctly, since this covers the remaining consumer.

Fixes #9228

### DIFF
--- a/mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
@@ -289,7 +289,7 @@ public class PromptTemplateConverter {
         if (template == null || template.getTemplate() == null) {
             return null;
         }
-        return renderTemplate(template.getTemplate(), args);
+        return renderTemplate(template, args);
     }
 
     private PromptTemplate parseFromNode(JsonNode node) {
@@ -424,6 +424,52 @@ public class PromptTemplateConverter {
         }
 
         return arguments;
+    }
+
+    /**
+     * Returns the arguments to render with, filling in the <code>default</code> declared by the
+     * template's variable schema for any argument the caller did not supply.
+     * <p>
+     * Caller-supplied values always win, and an argument that is explicitly present is never
+     * overwritten. The caller's map is never modified; a copy is made only when there is a
+     * default to apply. A <code>default</code> declared with no value is treated as absent.
+     */
+    public Map<String, Object> applyVariableDefaults(PromptTemplate template, Map<String, Object> args) {
+        if (template == null || template.getVariables() == null) {
+            return args;
+        }
+
+        Map<String, Object> effective = null;
+
+        for (Map.Entry<String, VariableSchema> entry : template.getVariables().entrySet()) {
+            String name = entry.getKey();
+            if (args != null && args.containsKey(name)) {
+                continue;
+            }
+
+            VariableSchema varSchema = entry.getValue();
+            if (varSchema == null || varSchema.getDefaultValue() == null) {
+                continue;
+            }
+
+            if (effective == null) {
+                effective = args != null ? new HashMap<>(args) : new HashMap<>();
+            }
+            effective.put(name, varSchema.getDefaultValue());
+        }
+
+        return effective != null ? effective : args;
+    }
+
+    /**
+     * Render a parsed template, applying the defaults declared by its variable schema for any
+     * argument the caller omitted.
+     */
+    public String renderTemplate(PromptTemplate template, Map<String, Object> args) {
+        if (template == null) {
+            return null;
+        }
+        return renderTemplate(template.getTemplate(), applyVariableDefaults(template, args));
     }
 
     /**
--- a/mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
@@ -434,7 +434,7 @@ public class PromptTemplateConverter {
      * overwritten. The caller's map is never modified; a copy is made only when there is a
      * default to apply. A <code>default</code> declared with no value is treated as absent.
      */
-    public Map<String, Object> applyVariableDefaults(PromptTemplate template, Map<String, Object> args) {
+    private Map<String, Object> applyVariableDefaults(PromptTemplate template, Map<String, Object> args) {
         if (template == null || template.getVariables() == null) {
             return args;
         }
@@ -529,6 +529,6 @@ public class PromptTemplateConverter {
         if (template == null || template.getTemplate() == null) {
             return null;
         }
-        return renderTemplate(template.getTemplate(), args);
+        return renderTemplate(template, args);
     }
 }
--- a/mcp/src/main/java/io/apicurio/registry/mcp/servers/PromptTemplateMCPServer.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/servers/PromptTemplateMCPServer.java
@@ -110,7 +110,7 @@ public class PromptTemplateMCPServer {
             }
 
             Map<String, Object> args = parseArguments(argumentsJson);
-            return converter.renderTemplate(template.getTemplate(), args);
+            return converter.renderTemplate(template, args);
         });
     }
 
--- a/mcp/src/test/java/io/apicurio/registry/mcp/PromptTemplateConverterTest.java
+++ b/mcp/src/test/java/io/apicurio/registry/mcp/PromptTemplateConverterTest.java
@@ -17,6 +17,8 @@
 package io.apicurio.registry.mcp;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.apicurio.registry.mcp.PromptTemplateConverter.PromptTemplate;
+import io.apicurio.registry.mcp.PromptTemplateConverter.VariableSchema;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -25,8 +27,9 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
- * Tests for the {{variable}} substitution done by the MCP prompt converter. These do not need a
- * Quarkus context - the converter only needs its ObjectMapper, which is set directly here.
+ * Tests for the {{variable}} substitution done by the MCP prompt converter, and for the defaults
+ * it applies from a parsed template's variable schema. These do not need a Quarkus context - the
+ * converter only needs its ObjectMapper, which is set directly here.
  */
 class PromptTemplateConverterTest {
 
@@ -36,6 +39,22 @@ class PromptTemplateConverterTest {
     void setUp() {
         converter = new PromptTemplateConverter();
         converter.jsonMapper = new ObjectMapper();
+    }
+
+    private static PromptTemplate template(String text, Map<String, VariableSchema> variables) {
+        PromptTemplate template = new PromptTemplate();
+        template.setTemplate(text);
+        template.setVariables(variables);
+        return template;
+    }
+
+    private static VariableSchema variable(String type, Object defaultValue) {
+        VariableSchema schema = new VariableSchema();
+        schema.setType(type);
+        if (defaultValue != null) {
+            schema.setDefaultValue(defaultValue);
+        }
+        return schema;
     }
 
     @Test
@@ -117,5 +136,87 @@ class PromptTemplateConverterTest {
                 """;
         Assertions.assertEquals("Hello Alice!",
                 converter.parseAndRenderAutoDetect(json, Map.of("name", "Alice")));
+    }
+
+    @Test
+    void testDeclaredDefaultIsApplied() {
+        PromptTemplate template = template("Write in a {{tone}} tone.",
+                Map.of("tone", variable("string", "formal")));
+
+        String rendered = converter.renderTemplate(template, Map.of());
+
+        Assertions.assertEquals("Write in a formal tone.", rendered);
+    }
+
+    @Test
+    void testCallerValueOverridesDefault() {
+        PromptTemplate template = template("Write in a {{tone}} tone.",
+                Map.of("tone", variable("string", "formal")));
+
+        String rendered = converter.renderTemplate(template, Map.of("tone", "casual"));
+
+        Assertions.assertEquals("Write in a casual tone.", rendered);
+    }
+
+    @Test
+    void testNonStringDefaultIsApplied() {
+        PromptTemplate template = template("Retry {{count}} times.",
+                Map.of("count", variable("integer", 3)));
+
+        String rendered = converter.renderTemplate(template, Map.of());
+
+        Assertions.assertEquals("Retry 3 times.", rendered);
+    }
+
+    @Test
+    void testVariableWithoutDefaultPreservesPlaceholder() {
+        PromptTemplate template = template("Hello, {{name}}!",
+                Map.of("name", variable("string", null)));
+
+        String rendered = converter.renderTemplate(template, Map.of());
+
+        Assertions.assertEquals("Hello, {{name}}!", rendered);
+    }
+
+    @Test
+    void testDefaultAppliedInsideConditionalBlock() {
+        PromptTemplate template = template("{{#if formal}}Dear {{title}}{{/if}}",
+                Map.of("formal", variable("boolean", true),
+                        "title", variable("string", "Sir or Madam")));
+
+        String rendered = converter.renderTemplate(template, Map.of());
+
+        Assertions.assertEquals("Dear Sir or Madam", rendered);
+    }
+
+    @Test
+    void testDefaultsDoNotModifyCallerSuppliedMap() {
+        PromptTemplate template = template("Write in a {{tone}} tone.",
+                Map.of("tone", variable("string", "formal")));
+        Map<String, Object> args = new HashMap<>();
+
+        converter.renderTemplate(template, args);
+
+        Assertions.assertTrue(args.isEmpty(),
+                "Rendering must not mutate the caller-supplied arguments map");
+    }
+
+    @Test
+    void testDefaultsAppliedWhenArgsAreNull() {
+        PromptTemplate template = template("Write in a {{tone}} tone.",
+                Map.of("tone", variable("string", "formal")));
+
+        String rendered = converter.renderTemplate(template, null);
+
+        Assertions.assertEquals("Write in a formal tone.", rendered);
+    }
+
+    @Test
+    void testTemplateWithoutVariablesSchemaIsUnchanged() {
+        PromptTemplate template = template("Hello, {{name}}!", null);
+
+        String rendered = converter.renderTemplate(template, Map.of());
+
+        Assertions.assertEquals("Hello, {{name}}!", rendered);
     }
 }
--- a/mcp/src/test/java/io/apicurio/registry/mcp/PromptTemplateConverterTest.java
+++ b/mcp/src/test/java/io/apicurio/registry/mcp/PromptTemplateConverterTest.java
@@ -219,4 +219,57 @@ class PromptTemplateConverterTest {
 
         Assertions.assertEquals("Hello, {{name}}!", rendered);
     }
+
+    private static final String TONE_TEMPLATE_YAML = """
+            templateId: tone-example
+            template: "Write in a {{tone}} tone."
+            variables:
+              tone:
+                type: string
+                default: formal
+            """;
+
+    /**
+     * Drives the render_prompt_template and render_registry_prompt path end to end, so that
+     * routing parseAndRenderAutoDetect through the schema-aware overload cannot be reverted
+     * without a failing test.
+     */
+    @Test
+    void testDefaultAppliedThroughParseAndRenderAutoDetect() {
+        String rendered = converter.parseAndRenderAutoDetect(TONE_TEMPLATE_YAML, Map.of());
+
+        Assertions.assertEquals("Write in a formal tone.", rendered);
+    }
+
+    /**
+     * Same guarantee for the explicit content type entry point, which parses the template and
+     * therefore also has the variable schema available.
+     */
+    @Test
+    void testDefaultAppliedThroughParseAndRender() {
+        String rendered = converter.parseAndRender(TONE_TEMPLATE_YAML, "application/x-yaml", Map.of());
+
+        Assertions.assertEquals("Write in a formal tone.", rendered);
+    }
+
+    /**
+     * An argument the caller supplied explicitly as null is present, so the declared default is
+     * not applied to it. Note that the substitution step then renders a null value as an empty
+     * string, which is existing behaviour of this converter and is not changed here. The point of
+     * this test is that the result is not the default, so a later change from containsKey to a
+     * null check cannot silently flip the guarantee.
+     */
+    @Test
+    void testExplicitNullArgumentIsNotOverwrittenByDefault() {
+        PromptTemplate template = template("Write in a {{tone}} tone.",
+                Map.of("tone", variable("string", "formal")));
+        Map<String, Object> args = new HashMap<>();
+        args.put("tone", null);
+
+        String rendered = converter.renderTemplate(template, args);
+
+        Assertions.assertEquals("Write in a  tone.", rendered);
+        Assertions.assertNotEquals("Write in a formal tone.", rendered,
+                "An explicitly supplied null must not be replaced by the declared default");
+    }
 }


paoloantinori (commented):
The core fix is correct and well-scoped: routing the already-parsed `defaultValue` through a single new `renderTemplate(PromptTemplate, Map)` overload covers all three MCP entry points, the lazy-copy and null semantics are sound, and the new tests are a real improvement (this class had none before). My comments are about regression coverage of the fix point, not the rendering logic itself.

- `parseAndRenderAutoDetect` (the path behind `render_prompt_template` and `render_registry_prompt`) is the actual fix at line 292, but no test drives it. The eight new tests call `renderTemplate(PromptTemplate, Map)` directly, so reverting that one line keeps the suite green while silently restoring the original bug.
- The 'explicitly present is never overwritten' guarantee rests on `containsKey` (line 446). The explicit-null caller case is the subtle one and is untested, so a later `get(name) != null` change would flip it with no failing test.
- `applyVariableDefaults` is `public` but is called only by the new overload and by no test; if there is no external caller, `private` avoids widening the converter's API surface.
- Worth a quick grep across the MCP module for any remaining `renderTemplate(String, Map)` call site. The original bug was exactly a render path that bypassed the variable schema, so any entry point still on the string overload would skip defaults by design.
- Scope note: the Apicurio source is not checked out in this review environment, so the templating engine's handling of non-string and boolean defaults is taken on the strength of the new assertions, not verified against the renderer source.

mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
```diff
@@ -289,7 +289,7 @@ public String parseAndRenderAutoDetect(String content, Map<String, Object> args)
         if (template == null || template.getTemplate() == null) {
             return null;
         }
-        return renderTemplate(template.getTemplate(), args);
+        return renderTemplate(template, args);
```
paoloantinori, line 292: This line is the actual fix for the `render_prompt_template` / `render_registry_prompt` paths, but none of the eight new tests drive `parseAndRenderAutoDetect(content, args)` directly; they all call `renderTemplate(PromptTemplate, Map)`. Reverting this single line to `renderTemplate(template.getTemplate(), args)` would keep every test green and silently re-introduce the bug. Add a test that renders through `parseAndRenderAutoDetect` with a template whose schema declares a `default`, supplies no caller arg, and asserts the default is substituted.

mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
```diff
@@ -426,6 +426,52 @@ private List<MCPArgument> deriveArgumentsFromVariables(Map<String, VariableSchem
...
+    public Map<String, Object> applyVariableDefaults(PromptTemplate template, Map<String, Object> args) {
+        if (template == null || template.getVariables() == null) {
+            return args;
+        }
+
+        Map<String, Object> effective = null;
+
+        for (Map.Entry<String, VariableSchema> entry : template.getVariables().entrySet()) {
+            String name = entry.getKey();
+            if (args != null && args.containsKey(name)) {
```
paoloantinori, line 446: The 'explicitly present is never overwritten' guarantee rests on `containsKey` here: a caller passing `{tone: null}` keeps its null and skips the default. That is the subtle case, and `testCallerValueOverridesDefault` only covers a non-null value. Add a test passing an explicit-null argument and assert the default is NOT applied, so a later refactor to `args.get(name) != null` cannot flip this behavior silently.

mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java
```diff
@@ -426,6 +426,52 @@ private List<MCPArgument> deriveArgumentsFromVariables(Map<String, VariableSchem
...
 
+    /**
+     * Returns the arguments to render with, filling in the <code>default</code> declared by the
+     * template's variable schema for any argument the caller did not supply.
+     * <p>
+     * Caller-supplied values always win, and an argument that is explicitly present is never
+     * overwritten. The caller's map is never modified; a copy is made only when there is a
+     * default to apply. A <code>default</code> declared with no value is treated as absent.
+     */
+    public Map<String, Object> applyVariableDefaults(PromptTemplate template, Map<String, Object> args) {
```
paoloantinori: `applyVariableDefaults` is invoked only by the new `renderTemplate(PromptTemplate, Map)` overload at line 472, and none of the tests call it directly. If there is no external caller, prefer `private` (or package-private if a test is intended to exercise it) to avoid exposing an internal helper on the converter's API.

Aarti-panchal01:
Thanks @paoloantinori, all four points are addressed in `9c7ca30`. The suite is now 11 tests, all passing.

**Regression coverage of the fix point.** You were right that the eight original tests all called `renderTemplate(PromptTemplate, Map)` directly, so the fix point itself was untested. Added `testDefaultAppliedThroughParseAndRenderAutoDetect`, which renders through `parseAndRenderAutoDetect` from YAML content whose schema declares a `default` and supplies no caller argument. I verified it against exactly the revert you described: putting line 292 back to `renderTemplate(template.getTemplate(), args)` now fails with

```
expected: <Write in a formal tone.> but was: <Write in a {{tone}} tone.>
```

**The grep found a second one.** `parseAndRender(content, contentType, args)` at line 525 parses the template, so it has the variable schema, but still rendered through the string overload and skipped defaults. Same bug, second entry point. It is now routed through the schema-aware overload and covered by `testDefaultAppliedThroughParseAndRender`, which I verified fails the same way against its own revert.

One thing to flag rather than slip in: `parseAndRender` is `public` and has no callers anywhere in the repository, so this is a deliberate behaviour change to an unused public method, made so it matches its sibling instead of staying a path that skips defaults by design. Happy to drop it back if you would rather keep this PR to the paths that are actually reachable.

**Explicit null.** Added `testExplicitNullArgumentIsNotOverwrittenByDefault`, which passes `{tone: null}` and asserts the default is not applied, so a later change from `containsKey` to `args.get(name) != null` cannot flip it silently.

Worth knowing what that case actually renders: the substitution loop in `renderTemplate(String, Map)` turns a null value into an empty string, so the result is `Write in a  tone.` rather than a preserved `{{tone}}`. That is existing converter behaviour and I have not changed it here, but it is a divergence from the REST renderer, which keeps the placeholder when a value is null. The test asserts both the actual output and that it is not the default, so the intent is recorded either way. I can file that divergence separately if it is worth tracking.

**Visibility.** `applyVariableDefaults` is now `private`. The new overload is its only caller and no test calls it directly, so nothing needed widening.

**On your scope note.** Understood that the source was not available in your review environment. For what it is worth, the non-string and conditional-block cases are exercised against the real renderer locally rather than asserted in isolation: `testNonStringDefaultIsApplied` renders an integer default through the substitution path, and `testDefaultAppliedInsideConditionalBlock` renders a boolean default through `processConditionalBlocks`. Both pass on `9c7ca30`.

paoloantinori:
Thanks, that covers it. The new `testDefaultAppliedThroughParseAndRenderAutoDetect` is exactly the gap I was worried about: it drives the fix point through `parseAndRenderAutoDetect`, so the routing cannot be reverted silently. The `testExplicitNullArgumentIsNotOverwrittenByDefault` test pinning the `containsKey` semantics is a nice touch; that's the kind of subtle contract that's easy to break later with a well-intentioned null check.

No further concerns from me on this one.

Aarti-panchal01:
@EricWittmann when you have a moment, this one is ready for `/accept`.

The review is complete. @paoloantinori raised four points on Aug 8, all of them are addressed in `9c7ca30`, and he confirmed: "No further concerns from me on this one." His grep suggestion also turned up a second entry point, `parseAndRender`, which had the same defect and is fixed here too. The branch is current with main and shows as mergeable, and the converter suite is at 11 passing tests. The required checks are still waiting on `/accept` before they can run.

Separately, when you get a chance: merging #9309 auto-closed #9228. The description said `Part of #9228` rather than `Fixes`, but the linked issue in the sidebar stayed attached from when that PR was first opened, and that is what closed it. I am not able to reopen it myself. This PR carries `Fixes #9228` and covers the remaining consumer, so reopening it would let the merge close it properly.

paoloantinori:
Nothing further from me here. The four points I raised are all addressed in `9c7ca30`, and `testDefaultAppliedThroughParseAndRenderAutoDetect` closes the routing gap I was worried about. Over to @EricWittmann for the accept.

Aarti-panchal01:
Flagging why #9447 shows as a cross-reference here: it is a separate defect in the same method, found while addressing the review feedback on this PR, and it does not affect this one.

`renderTemplate(String, Map)` builds its substitution pattern by concatenating each caller-supplied argument name unquoted, so the name is compiled as a regular expression. An argument named `.*` replaces every placeholder on the line, and one such as `a)(b` fails the render with a `PatternSyntaxException`. The names arrive from `parseArguments` with no validation against the declared variables, so they are caller-controlled.

I deliberately did not fold it into this PR. That defect predates this change and is independent of it, the scope reviewed here is already signed off, and the fix is a one-line quoting change whose tests are worth reviewing on their own rather than buried in a defaults PR.

Aarti-panchal01:
The suite failure on b5fd3be is a broken `main`, not this branch.

`Build Java Application (no tests)` fails to compile the `app` module, with nothing from this diff involved — `WellKnownResource.java` imports `io.apicurio.registry.a2a.rest.beans`, which does not exist. It traces to 895a23a80 (#8847); every push to `main` since has failed identically, including the current tip 584c71621. Details in #9457.

I do not think I can clear it from here: `verify-build.yaml` checks out without an explicit `ref`, so on `pull_request` the build is `refs/pull/9392/merge` — this branch merged into `main` — so the breakage is included whether or not I keep my own merge commit. I will merge `main` and re-run once it is green.

EricWittmann:
@Aarti-panchal01 This PR has a merge conflict — could you rebase on the latest main and push? We'd like to get this merged soon. Thanks!

Aarti-panchal01:
Rebased onto main (9bf9d60) and force-pushed — the merge commit is gone, it's two commits now.

The conflict was with #8977, which landed the canonical `{{variable}}` parser and added its own `PromptTemplateConverterTest`, so we both created that file. I merged the two classes into one: #8977's substitution tests and this PR's defaults tests now live side by side, 22 tests total, all green locally (`mvn -Dfull -pl mcp -am -Dtest=PromptTemplateConverterTest`).

The converter itself merged without a semantic clash — defaults are applied to the arguments map before substitution, so they go through #8977's canonical pattern rather than around it.

The build failure I reported in #9457 is also resolved on main now: `WellKnownResource` imports `mcptools.rest.beans` instead of the `a2a.rest.beans` package that was never generated. CI should be green on this push.

EricWittmann:
@Aarti-panchal01 I think you need to rebase on main and fix the conflicts.

Aarti-panchal01:
Rebased onto main and force-pushed — the branch is now `110bc9dc`, replayed on top of `b29d0d7`.

To be straight with you about what I found: **there were no conflicts to fix.** The five commits main had gained since my last push were the CI shard fix (#9247), the label filter parsing fix (#9092), the gitops README links (#9088), CLI Windows support (#9229), and the `GroupCommandTest` split (#9283). None of them touch `mcp/` or any prompt rendering file, and both of my commits replayed with zero conflicts.

The diff is unchanged from what you and @paoloantinori reviewed — I verified that rather than assuming it, by hashing `git diff origin/main...HEAD` before and after the rebase and getting the same object (`276e33ee`). Still 3 files, +205/-5, and the converter suite is 22 tests green locally:

```
Tests run: 22, Failures: 0, Errors: 0, Skipped: 0 -- in io.apicurio.registry.mcp.PromptTemplateConverterTest
BUILD SUCCESS
```

What the merge box is actually reporting is **out-of-date with the base branch**, not a conflict — the API has said `mergeable: MERGEABLE` / `mergeStateStatus: BEHIND` the whole time, and the box itself says "Changes can be cleanly merged." That is also consistent with the checks going green after your `/accept`: nothing was broken, the branch was just behind.

One practical note on that. Main picked up another commit in the few minutes it took me to build and push — #9463, a `ubi-minimal` digest bump from renovate — so this branch already reads as out-of-date again. I have deliberately not chased that one, because re-pushing to absorb a container digest bump would restart all 40-odd checks without changing a line of Java. If the out-of-date state is what is holding the merge, `/auto-merge` would ride over it and merge once checks are green, instead of needing a manual re-sync every time renovate lands. If you would rather I keep rebasing by hand, just say so and I will.

CI is running on `110bc9dc` now.

Separately, on #9483 — agreed with the framing, and the duplication table there matches what I hit building this. The null-value row is exactly the divergence I flagged upthread: REST keeps the placeholder verbatim, the MCP converter substitutes an empty string, which is why `testExplicitNullArgumentIsNotOverwrittenByDefault` asserts the empty-string output rather than a preserved `{{tone}}`. Happy to take that unification on once this one lands, since it would then have both `applyVariableDefaults` and #9391's conditional handling in place to fold together.

Aarti-panchal01:
Correcting myself on two points from my last comment — I went and read `.github/scripts/pr-lifecycle.js` instead of guessing, and I had the blocker wrong.

**`/auto-merge` on its own will not merge this.** I suggested it as the way past the out-of-date state. It isn't. `isApproved()` (line 386) requires a review whose state is literally `APPROVED`, and the only submitted review here is @paoloantinori's from Aug 8, which the API reports as `COMMENTED` — his "no further concerns" was a comment, not a formal approval. Without that, `checkAndTransitionToReady` never sets `lifecycle/ready-to-merge`, and `performMerge` (line 399) aborts unless both `lifecycle/tested` and `lifecycle/ready-to-merge` are present. `/auto-merge` would just add the label and wait; the orchestrator would itself reply that "a review or `/skip-review` is still required before auto-merge can proceed."

So the actual unblock is one of: @paoloantinori converting his sign-off into a formal Approve, or `/skip-review`. That matches the labels — `lifecycle/tested` is on, `lifecycle/review-approved` is not.

**The out-of-date state was never something I needed to fix by hand.** `performMerge` catches the behind-the-base failure, checks `rebaseable`, calls `updateBranch`, tags `orchestrator/merge-rebase`, and re-merges with tests skipped since they already passed for the previous HEAD. The only case it skips that is a 403, which is for PRs that modify workflow files — this diff touches nothing under `.github/`, so that path was available. The rebase I just pushed was therefore unnecessary rather than wrong; the diff is unchanged and the suite is green, so there is no harm in it, but you did not need to ask me for it and I should have checked before adding a CI cycle.

**And one thing I stated that is plainly contradicted by this thread:** I wrote that the checks went green after your `/accept` and that "nothing was broken." That is wrong. The suite *failed* on `b5fd3be`, for the `WellKnownResource` / `a2a.rest.beans` compile break on main that I filed as #9457. Green only came later, on `3d18699`, after main was fixed and I rebased. I should not have smoothed that over.

Net state as it actually stands: `110bc9dc`, rebased on `b29d0d7`, 22 converter tests green locally, `lifecycle/tested` on, no conflicts, and waiting on a formal approval rather than on anything mechanical.

paoloantinori:
Thanks for the clear writeup on the rebase. I have re-read the diff at `110bc9dc` against what I reviewed at `9c7ca30`: the production change (the `renderTemplate(PromptTemplate, Map)` overload routing all three MCP entry points through `applyVariableDefaults`) and the 11 tests are byte-identical, only replayed on the newer main. None of the five commits main gained touch `mcp/` or prompt rendering, so the rebase is genuinely clean.

My prior sign-off stands: the four points I raised are addressed, `testDefaultAppliedThroughParseAndRenderAutoDetect` closes the routing gap through `parseAndRenderAutoDetect`, and I have nothing further from my side. Over to @EricWittmann for the merge.

Aarti-panchal01:
@paoloantinori I have to correct the baseline before you rest a sign-off on it — `110bc9dc` is **not** byte-identical to the `9c7ca30` you reviewed, and the difference is in the production file, not just the tests.

```
git diff --stat 9c7ca30 110bc9dc -- .../PromptTemplateConverter.java
  1 file changed, 11 insertions(+), 9 deletions(-)

git diff --stat 9c7ca30 110bc9dc -- .../PromptTemplateConverterTest.java
  1 file changed, 106 insertions(+), 15 deletions(-)

@Test count:  9c7ca30 → 11      110bc9dc → 22
```

The byte-identical claim was mine, and it was scoped to `3d18699 → 110bc9dc`, today's rebase only. Between your review and now sits the #8977 conflict resolution, which I described upthread but should have pointed at more directly.

**What changed that is worth your eyes.** #8977 landed the canonical `{{variable}}` parser, so `renderTemplate(String, Map)` no longer does its own substitution. The local `VARIABLE_PATTERN` and the `replaceAll` loop over `args.entrySet()` are gone, replaced by `PromptTemplateVariableUtil.substituteVariables(template, resolver)`. Defaults still land the same way — `applyVariableDefaults` fills the map first — but they now flow *through* the shared pattern rather than around it. Two consequences that are behaviour, not refactor:

1. **Substitution is now driven by the template, not by the caller's map.** Previously each arg name was concatenated into a regex and applied to the text; now the template is scanned with one fixed pattern and each placeholder name is looked up in the resolver. The inversion is why `{{ name }}` with whitespace resolves now when it did not before — the canonical pattern is `\{\{\s*(\w+)\s*\}\}`.
2. **It removes the #9447 defect as a side effect.** That report was against exactly the `replaceAll` loop this deleted: an argument named `.*` was compiled as a regex. Caller-supplied names are never compiled into a pattern any more, so it is structurally gone rather than fixed. I will close #9447 with that explanation rather than leave it open against code that no longer exists.

The `containsKey` semantics you asked me to pin are unchanged, and `testExplicitNullArgumentIsNotOverwrittenByDefault` still asserts the empty-string output through the new path. The 11 added tests are #8977's, in the same file because both PRs created `PromptTemplateConverterTest`. 22 green locally.

None of this looks like a problem to me, but it is a different substitution engine than the one you signed off on, so please take a second look rather than carrying the earlier approval forward on my word.

**Separately, a mechanical point, since "over to Eric for the merge" will not currently work.** Both of your sign-offs are `COMMENTED` reviews — the API reports one review on this PR, `paoloantinori / COMMENTED / 2026-08-08`. The orchestrator's `isApproved()` requires `state === 'APPROVED'`, so the PR cannot reach `lifecycle/ready-to-merge`, and `performMerge` aborts without that label. That matches the labels: `lifecycle/tested` is on, `lifecycle/review-approved` is not. So Eric cannot merge this through the bot until either you submit a formal Approve, or he runs `/skip-review`. I mention it only because a comment reads as a sign-off to humans and not to the workflow.

Aarti-panchal01:
@paoloantinori small ask to unblock this: once you have had the second look at the substitution change, could you submit your sign-off as a formal **Approve** rather than a comment?

Not chasing you for a rubber stamp — I would rather you re-read the `PromptTemplateVariableUtil` switch first, since that is the part you have not seen. But when you are satisfied, the review state is what the workflow keys on. Your Aug 8 review is recorded as `COMMENTED`, and `isApproved()` in `.github/scripts/pr-lifecycle.js` requires a latest review whose state is literally `APPROVED`. Until then the PR cannot reach `lifecycle/ready-to-merge`, and `performMerge` returns early without that label, so @EricWittmann's `/merge` or `/auto-merge` has nothing to act on.

Current labels are consistent with that: `lifecycle/tested` is on, `lifecycle/review-approved` is not, and `reviewDecision` on the PR is empty.

`110bc9dc`, checks green so far with the rest still running, no conflicts. Nothing needed from me that I know of — happy to make changes if the re-read turns something up.

Aarti-panchal01:
@EricWittmann could you run `/skip-review` on this one? It is the shortest path to merge from where the PR now sits, but I want to be clear about what it would and would not skip so the call is yours on full information.

**Why it is needed at all.** This is not an unreviewed PR. @paoloantinori reviewed it on Aug 8, raised four points, and confirmed twice that they are addressed. The problem is purely mechanical: his review is recorded with state `COMMENTED`, and `isApproved()` requires a latest review whose state is literally `APPROVED`. So `checkAndTransitionToReady` never sets `lifecycle/ready-to-merge`, and `performMerge` returns early without that label — your `/merge` and `/auto-merge` both have nothing to act on. `reviewDecision` on the PR is empty and `lifecycle/review-approved` is absent, which matches.

**What `/skip-review` would do here.** State is `lifecycle/ready-for-review` and `lifecycle/tested` is on, so it adds `orchestrator/review-skipped` and transitions straight to `lifecycle/ready-to-merge`. Tests are still required and have already passed. The out-of-date state does not need handling first — `performMerge` updates the branch itself and re-merges with tests skipped, and that path is available because this diff touches nothing under `.github/`.

**The one thing it does skip, which you should weigh.** Paolo's sign-off predates the #8977 conflict resolution. Since `9c7ca30` the production file changed by +11/-9: `renderTemplate(String, Map)` no longer does its own substitution, it now goes through `PromptTemplateVariableUtil.substituteVariables`, and the test file gained #8977's 11 tests in the same class. I flagged that above and asked him to take a second look. `/skip-review` would merge without that second look. My own read is that the change is sound — defaults are applied to the arguments map before substitution, so they flow through the canonical pattern rather than around it, and all 22 tests pass — but it is a different substitution engine than the one that was reviewed, and you are better placed than me to decide whether that needs eyes before or after merge.

So either works from my side: `/skip-review` if you are comfortable, or leaving it until Paolo converts his sign-off into a formal Approve. Happy to make changes if anything turns up either way.

Current state: `110bc9dc`, no conflicts, 22 converter tests green locally, checks green with the remainder still running.

paoloantinori:
Thanks for the careful writeup on the rebase and the honest callout that `/skip-review` would merge without a second look at the substitution engine change.

I went and re-read the production file at `110bc9dc`. The change since `9c7ca30` is real and it is in the substitution path, as you flagged: `renderTemplate(String, Map)` no longer does its own `{{var}}` handling, it delegates to `PromptTemplateVariableUtil.substituteVariables`. The new `renderTemplate(PromptTemplate, Map)` overload applies the defaults to the args map first, then calls into that canonical path. Defaults therefore flow through the same substitution engine as the REST render endpoint and the validity rule, not around it. That is the correct shape and it removes the risk of two substitution implementations drifting.

Tracing the wiring: `parseAndRenderAutoDetect` now calls `renderTemplate(template, args)` (the overload), so all three MCP entry points reach `applyVariableDefaults` before substitution. The lazy-copy in `applyVariableDefaults` (only allocate when a default actually applies) and the explicit-present-wins semantics both read correctly.

My read is the change is sound: it routes defaults through the canonical pattern rather than around it, the 22 tests pass, and the behavior is consistent across MCP and REST. No further concerns from me on the substance.

On the `/skip-review` question — that's @EricWittmann's call and I don't have a strong objection either way. The substitution-engine delta has now had a second look (this comment), so if Eric is comfortable with `/skip-review` the review gap you flagged is closed on my side. If he'd rather see a formal Approve, I'm equally fine leaving it for that path. Happy to convert my earlier sign-off into a formal Approve if that's the smoother route to merge; just say the word.

One process note: I won't post a formal Approve here unilaterally, since the governance contract I work under is comment-only. If Eric wants the Approve state on the PR, the cleanest route is for him to apply it, or for the `/skip-review` path to run.

Aarti-panchal01:
@EricWittmann short follow-up, because the one caveat I attached to my `/skip-review` ask is now resolved and one of the two routes has closed.

**The review gap is closed.** @paoloantinori went and re-read the substitution change — he traced `parseAndRenderAutoDetect` → the `renderTemplate(PromptTemplate, Map)` overload → `applyVariableDefaults` → `PromptTemplateVariableUtil.substituteVariables`, and confirmed the shape is right and the lazy-copy and explicit-present-wins semantics read correctly. So the "merging without a second look" concern I flagged no longer applies; that look has happened.

**`/skip-review` is now the only route.** Paolo says he will not post a formal Approve, because the governance contract he works under is comment-only. So "wait for Paolo's Approve" is not available, and the PR cannot reach `lifecycle/ready-to-merge` on its own — `isApproved()` will never see an `APPROVED` state. `/skip-review` from you, or a formal Approve from a maintainer who can give one, are what is left.

**Two state notes so nothing surprises you:**

- Head is now `35a64e54`, not the `110bc9dc` Paolo re-read. I rebased onto `143dd1e5` to clear the out-of-date banner. The diff is byte-identical — same hash `276e33ee`, still 3 files, +205/-5 — so his re-read applies unchanged. 22 converter tests green locally on the new head.
- `lifecycle/tested` is currently off, because it belonged to the previous head and `Verify` is re-running. That is my doing, not a failure — nothing has failed on this PR. If you run `/skip-review` before Verify finishes, `cmdSkipReview` will post that it moves to `lifecycle/ready-to-merge` once tests pass, rather than merging immediately. Either order works.

Nothing outstanding from me. Happy to make changes if anything turns up.

Aarti-panchal01:
/retry

The `PR Push` failure at 18:30 was an infrastructure fault, not a problem with this branch. The orchestrator errored inside `Handle push` with `HttpError: self-signed certificate` / `TypeError: fetch failed` while calling the GitHub API from the runner — no step of it reached any code in this PR. Re-running so the labels and state transitions land correctly on `35a64e54`.

The test suite itself has nothing failing: `Verify` is still in progress and the only red check was this orchestrator job.

paoloantinori:
Confirmed on my end too — the rebase at `110bc9dc` preserves the production shape I read at `9c7ca30` (the `renderTemplate(PromptTemplate, Map)` overload routing through `applyVariableDefaults`), and `testDefaultAppliedThroughParseAndRenderAutoDetect` still drives the fix point through `parseAndRenderAutoDetect`. Nothing further from me on the substitution change.

Aarti-panchal01:
Rebased onto main (`96166bc04`) — all checks green, `mergeable`/`clean`.

This rebases cleanly over #9473, which touches the same `PromptTemplateMCPServer.java`; the change here is unaffected. @paoloantinori reviewed it twice with no remaining concerns.

@jsenko could you take a look when you get a chance? Two follow-ups are queued behind this one (the #9447 regression tests, and the #9457 fix once a direction is named), and the one-open-PR rule keeps them blocked until this closes.

paoloantinori:
The failing run (31521334586) was on `35a64e5`, against a broken `main` where the `app` module failed to compile. The current head `f681345` is rebased onto the fixed `main` and the full `Verify` suite is green (run 31564501436): all unit test shards, integration tests, and SDK verification pass. This was an infra/broken-main failure, not a regression from this branch. Safe to proceed.

EricWittmann:
Thanks for the contribution, @Aarti-panchal01! Merged.

PR #9393, opened 2026-08-08
fix(search): contract search returns empty results due to label filter mismatch

### LINKED ISSUE #9144: fix(search): contract search endpoint returns empty results for existing contracts
### Title

GET /registry/v3/search/contracts returns no results due to contract label filter mismatch

### Description

While testing the contract search endpoint, I noticed that `GET /registry/v3/search/contracts` returns an empty result set even when contract artifacts with contract metadata exist.

The same behavior is observed when using the `status`, `ownerTeam`, or `compatibilityGroup` query parameters.

### Steps to reproduce

1. Create a group and an artifact.
2. Update the artifact's contract metadata (for example, set the status to `DRAFT`).
3. Call:

```
GET /registry/v3/search/contracts
```

or

```
GET /registry/v3/search/contracts?status=DRAFT
```

4. Observe the response.

### Expected behavior

The endpoint should return the matching contract artifacts.

For example:

```json
{
  "count": 1,
  "artifacts": [
    {
      "groupId": "my-group",
      "artifactId": "my-contract-artifact"
    }
  ]
}
```

### Actual behavior

The endpoint returns HTTP 200, but no artifacts are returned.

```json
{
  "count": 0,
  "artifacts": []
}
```

### Investigation

From tracing the request path:

- `SearchResourceImpl.searchContracts()` constructs label filters using `SearchFilter.ofLabel("contract.")`.
- `SearchFilter.ofLabel("contract.")` creates a label filter with the key `contract.`.
- `SqlSearchRepository.buildWildcardClause()` only generates a `LIKE` predicate when the value contains `*`. Since `contract.` contains no wildcard, the generated SQL uses an exact equality comparison:

```sql
l.labelKey = 'contract.'
```

However, contract metadata labels are stored using specific keys such as:

- `contract.status`
- `contract.owner.team`
- `contract.compatibility.group`

As a result, the generated predicate does not match the stored label keys.

I also noticed that `DataContractsResourceTest.testSearchContracts_WithStatusFilter` currently verifies only the HTTP status code and does not assert that any artifacts are actually returned, so this behavior is not currently covered by the test suite.

---

### Possible fix

It looks like `searchContracts()` should build filters using the actual contract label keys (for example, the constants defined in `ContractLabels`) instead of the generic `contract.` prefix.

It would also be helpful to strengthen the existing integration tests by asserting the returned result count and artifact IDs.

Myself-Praveen:
…r mismatch

The /search/contracts endpoint built label filters with an exact key of "contract." which never matches real label keys (e.g. contract.{contractId}. status) because the SQL layer only treats "*" as a wildcard. Additionally, label values written via mergeArtifactLabels/mergeVersionLabels are stored verbatim (e.g. "DRAFT") while the search lowercased the filter value, so status/ownerTeam/compatibilityGroup filters never matched either.

Fix the base filter to use a "contract.*" prefix match and the suffix filters to match "contract.*{suffix}", and compare label values case-insensitively via LOWER(l.labelValue) in artifact, version, and group label searches.

Closes #9144


Myself-Praveen:
Fixed in ef1b262 — reordered the static Hamcrest imports.

app/src/main/java/io/apicurio/registry/rest/v3/impl/SearchResourceImpl.java
```diff
@@ -464,17 +465,25 @@ public ArtifactSearchResults searchContracts(String status, String ownerTeam,
...
-        // All contracts have a contract.*.status label
-        filters.add(SearchFilter.ofLabel("contract."));
+        // All contracts have labels in the reserved "contract.*" namespace. The "*" is
+        // required for a prefix match (label keys are namespaced per contract, e.g.
+        // "contract.{contractId}.status"), since the SQL layer only treats "*" as a
+        // wildcard. Without it the filter becomes an exact match on "contract." and
+        // never matches anything.
+        filters.add(SearchFilter.ofLabel("contract.*"));
 
+        // Suffix filters match the label key by suffix ("contract.status" for metadata
```
paoloantinori: This comment claims the filter matches both `contract.status` (metadata written without a contract id) and `contract.{contractId}.status`, but the key `contract.*` + `SUFFIX_STATUS` expands to `contract.%.status`, a LIKE that requires a literal dot before `status`. It matches the dotted form and `contract..status`, but it cannot match `contract.status`. The base-filter comment above (the `contract.{contractId}.status` note around lines 468-470) implies keys are always namespaced with `{contractId}`, which contradicts this one. Confirm whether the no-id form is ever written: if it is, those contracts are still missed by this filter; if it is not, correct this comment so a future reader does not rely on the wildcard covering the optional segment.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DataContractsResourceTest.java
```diff
@@ -1049,10 +1051,22 @@ public void testAuditLog_EndpointReturnsArray() throws Exception {
...
+
+        given().when().contentType(CT_JSON)
+                .pathParam("groupId", GROUP).pathParam("artifactId", artifactId)
+                .body(EditableContractMetadata.builder().status(EditableContractMetadata.Status.DRAFT).build())
+                .put("/registry/v3/groups/{groupId}/artifacts/{artifactId}/contract/metadata")
+                .then().statusCode(200);
+
         given().when()
                 .get("/registry/v3/search/contracts")
                 .then().statusCode(200)
```
paoloantinori, line 1136: The assertions prove a contract artifact is returned, but nothing here covers the base `contract.*` filter's job of excluding non-contract artifacts. If that wildcard ever matched too broadly (or matched nothing), the status/ownerTeam/compatibilityGroup value filters would still hide the regression because they filter on the value. Consider creating a second artifact with no contract metadata in this test and asserting its `artifactId` is absent from `/search/contracts`, so the namespace-scoping behavior is pinned down independently of the value filters.

paoloantinori (commented):
Overall the fix is sound: switching the base filter from an exact `contract.` key to a `contract.*` prefix match and comparing label values with `LOWER()` addresses the empty-results bug, and the new tests are a real improvement. Two things to resolve before merge: the suffix-filter comment contradicts the pattern it documents, and the base filter's exclusion behavior is untested.

- `SearchResourceImpl.java:475-479`. The suffix comment claims the filter matches both `contract.status` (no contract id) and `contract.{contractId}.status`, but `contract.*` + `SUFFIX_STATUS` expands to `contract.%.status`, a LIKE that requires a literal dot before `status`. It matches the dotted form but not `contract.status`. The base-filter comment above it (lines 468-470) states keys are always namespaced with `{contractId}`, which contradicts this one. Confirm which form is actually written at merge time and align the comment with the pattern, or contracts using the no-id form stay invisible to this filter.
- `SqlSearchRepository.java`, `SqlGroupRepository.java`. The `LOWER(l.labelValue)` change is applied in three places. Confirm there is no fourth `buildWildcardClause(..., "l.labelValue", ...)` site (another artifact/version/branch label search path) left comparing case-sensitively, or label filtering will be inconsistent across endpoints.
- `LOWER(l.labelValue)` wraps the column in a function, which makes the predicate non-sargable: any index on `labelValue` can no longer be used for these lookups. Correct for case-insensitivity, but verify the label join is not a hot path at scale, or plan a functional/case-insensitive index.
- Test strength. Only `status=DRAFT` actually depends on the new `LOWER()` (stored uppercase, queried uppercase). The `ownerTeam` (`search-team`) and `compatibilityGroup` (`orders-cg`) tests use lowercase values that already match without the fix, so they do not prove the case-insensitive change for those fields. Add a mixed-case value to exercise it.

Myself-Praveen:
@paoloantinori thanks for the detailed review — all points addressed in 1f58774.

1. Suffix-filter pattern & comment. Both label forms are written at merge time: updateContractMetadata uses the contract.{contractId}. prefix when a contract id exists, and falls back to the plain contract. prefix (→ contract.status) when it does not. The filter contract.* + SUFFIX_STATUS concatenates to contract.*status, which the SQL layer expands to contract.%status — there is no literal dot between % and status. Since % matches the empty string, it covers the no-id form too. I've rewritten both comments to state the two forms and explain the expansion precisely.

2. Fourth LOWER() site. Confirmed exactly three sites, all wrapped in LOWER(): SqlSearchRepository.java:135 (artifact labels), SqlSearchRepository.java:328 (version labels), SqlGroupRepository.java:313 (group labels). No fourth artifact/version/branch label path compares case-sensitively.

3. Non-sargable LOWER(). Acknowledged. The LOWER() is applied only to the correlated labelValue column inside the EXISTS subquery; the join itself keys on labelKey/ids, which is not a hot path. Worth revisiting with a functional/case-insensitive index separately if label search volume grows.

4. Test strength. ownerTeam is now stored as Search-Team and compatibilityGroup as Orders-CG (mixed-case stored, lowercase query), so both now genuinely depend on LOWER() — status=DRAFT is no longer the only field exercising it.

5. Base-filter exclusion. Added testSearchContracts_ExcludesArtifactsWithoutContractMetadata, which creates a second artifact with no contract metadata and asserts its artifactId is absent from /search/contracts, pinning the contract.* namespace-scoping independent of the value filters.

paoloantinori:
Thanks for the thorough response. I re-read 1f58774 and all five points are addressed: the suffix-filter comments now spell out both label forms and the `%` expansion, the three `LOWER()` sites are confirmed with no fourth path, the mixed-case `Search-Team`/`Orders-CG` fixtures genuinely exercise the case-insensitive comparison, and `testSearchContracts_ExcludesArtifactsWithoutContractMetadata` pins the namespace scoping. From my side this looks good to merge.

Myself-Praveen:
/unstale

@EricWittmann Gentle ping on this—all review points from @paoloantinori were addressed and approved. Ready for maintainer review/merge whenever you have a chance.

paoloantinori:
Confirming my side is closed: all the points from my review were addressed at 1f58774 and I have nothing further to add. The remaining step is the maintainer merge call.

Myself-Praveen:
@paoloantinori @EricWittmann I've fixed the DCO sign-off issue! However, the automated triage bot is blocking the merge because the original issue (#9144) doesn't have maintainer approval. Could one of you please drop a quick comment of approval on issue #9144 so the bot allows this PR to merge?

EricWittmann (changes requested):
The DataContractsResourceTest has a duplicate import.  Once you fix that I think this PR is 👍 to merge.

EricWittmann (approved):
LGTM

Myself-Praveen:
Thanks @paoloantinori for the thorough review and @EricWittmann for the approval! All checks are green and the branch is up to date. Ready for merge whenever you get a chance.

EricWittmann:
Thanks @Myself-Praveen

PR #9394, opened 2026-08-08
fix(ui): stop highlighting unsupported Handlebars block syntax as recognized template syntax

### LINKED ISSUE #9382: [Bug]: UI highlights Handlebars-style conditional blocks ({{#if}}, {{#each}}) as recognized syntax, but the backend render service does not process them
## Description

**Registry Version**: main (snapshot / latest-snapshot docker image)

The prompt template documentation/preview UI visually distinguishes Handlebars-style block syntax — {{#if x}}, {{/if}}, {{#each x}}, 
{{/each}}, {{#unless x}}, {{#with x}} — from plain {{variable}} placeholders, applying a distinct "template-block" CSS style to them. This implies the system recognizes and supports conditional/loop template logic.

However, the server-side render endpoint (PromptRenderingService.java) has no logic to evaluate or process these block constructs. Its substitution regex treats the entire contents between {{ and }} as a single variable name, with no special handling for #if, /if, #each, /each, #unless, /unless, #with, or /with.

### Environment

Running via Docker (apicurio/apicurio-registry:latest-snapshot and apicurio/apicurio-registry-ui:latest-snapshot), backend on port 8080, 
UI on port 8888. Tested in Chrome browser on Windows, and via direct REST API calls (curl).

### Steps to Reproduce

1. Create a PROMPT_TEMPLATE artifact with a template containing a Handlebars-style conditional block, e.g.:
   { "template": "{{#if isPremium}}Welcome, premium user!{{/if}}", 
   "variables": { "isPremium": { "type": "boolean" } } }
2. View the artifact in the Documentation tab — note that  {{#if isPremium}} and {{/if}} are visually highlighted as recognized template syntax (distinct styling from a plain variable).
3. Call the render endpoint with isPremium set to true:
   Body: { "isPremium": true }

**Expected behavior**
Either:
(a) The conditional block should be evaluated, producing "Welcome, premium user!" when isPremium is true (and empty output when 
    false), or
(b) If conditional blocks are not actually supported, the UI should not visually distinguish this syntax as recognized template control structures, since doing so misleads users into writing templates that will not render as expected.

**Actual behavior**
The rendered output contains the literal, unprocessed text "{{#if isPremium}}Welcome, premium user!{{/if}}" — the block syntax is not evaluated, stripped, or substituted in any way.

### Root Cause
The variable-highlighting regex in PromptTemplateViewer.tsx (/\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g) explicitly recognizes Handlebars-style block helper syntax and styles it differently from plain variables — implying UI-level awareness of this syntax. However, no corresponding logic exists anywhere in the backend (PromptRenderingService.java or elsewhere) to actually evaluate these blocks; the backend's substitution regex (\{\{([^}]+)\}\}) treats everything between the braces as a flat variable name lookup.

### Logs

(No error logs — this is a silent feature-gap/inconsistency between frontend syntax recognition and backend rendering capability, not a crash.)

jayaprajapatii:
Fixes #9382

## PR Summary

The prompt template preview UI visually highlighted Handlebars-style block syntax (`{{#if x}}`, `{{/if}}`, `{{#each x}}`, `{{/each}}`, `{{#unless x}}`, `{{#with x}}`) with a distinct `template-block` CSS class, separate from plain `{{variable}}` placeholders. This implied the system recognized and supported conditional/loop template logic.

However, the backend render service (`PromptRenderingService.java`) has no logic to evaluate these block constructs — it only performs flat variable substitution. As a result, templates using this syntax render with the raw, unprocessed block markers left in the output, which contradicts what the UI visually suggested was supported.

## Fix

Simplified the variable-highlighting regex in `PromptTemplateViewer.tsx` to only match plain `{{variable}}` syntax, removing the special-case detection and styling for Handlebars-style block helpers:

* Before: `/\{\{(#?\/?(?:if|unless|each|with)\s+)?(\w+)\}\}/g`, with a separate `template-block` class for block syntax.

* After: `/\{\{(\w+)\}\}/g`, with a single `template-variable` class.

Unsupported block syntax (`{{#if x}}`, etc.) is no longer specially highlighted — it now renders as plain, unstyled text, which is consistent with how the backend actually treats it (as literal text that doesn't match any variable).

## Testing

* Verified the regex behavior in isolation (Node.js):

| Input | Old highlighting | New highlighting |
|-------|-------------------|-------------------|
| `{{name}}` | `template-variable` | `template-variable` |
| `{{#if isPremium}}` | `template-block` (misleading) | not matched (plain text) |
| `{{/if}}` | `template-block` (misleading) | not matched (plain text) |

* Ran `eslint` on the modified file with no issues reported.


EricWittmann:
Should we instead support conditional blocks when rendering?  See this PR:  
* https://github.com/Apicurio/apicurio-registry/pull/9391

jayaprajapatii:
Hii @EricWittmann. I agree that supporting conditional blocks would address the underlying issue more completely.

At the same time, if conditional blocks aren't supported on the main render path, highlighting them as supported Handlebars syntax is misleading for users.

EricWittmann:
We are going to update the render path to support such blocks! Closing this issue. 👍

PR #9395, opened 2026-08-08
fix(events): serialize CloudEvent time as RFC 3339 regardless of mapper settings

### LINKED ISSUE #9325: fix(events): make CloudEvent time RFC3339 independent of mapper defaults
## Description

`CloudEventDto.time` is an `Instant` without field-level serialization metadata. It serializes as an RFC 3339 string with the Quarkus-injected `ObjectMapper` because the application globally disables `WRITE_DATES_AS_TIMESTAMPS`, but a plain mapper with `JavaTimeModule` can serialize it as an epoch number.

CloudEvents 1.0 requires the optional `time` attribute, when present, to be an RFC 3339 timestamp. A DTO advertised as CloudEvents-compliant should preserve that wire contract independently of unrelated global mapper settings.

This was noted as a forward-looking design item in the final review of #8901:
https://github.com/Apicurio/apicurio-registry/pull/8901#pullrequestreview-4844660074

## Current behavior

With a mapper that registers `JavaTimeModule` but leaves `WRITE_DATES_AS_TIMESTAMPS` enabled, `Instant.parse("2024-01-01T00:00:00Z")` is emitted numerically.

The existing tests correctly guard the injected production mapper, but they do not make the DTO itself enforce the CloudEvents wire format.

## Expected behavior

Whenever `CloudEventDto.time` is present, it serializes as an RFC 3339 string regardless of the mapper's global timestamp setting.

## Proposed scope

- Add field-level Jackson formatting for `CloudEventDto.time`.
- Add a focused regression test using a mapper whose global timestamp serialization remains enabled.
- Preserve deserialization and the existing injected-mapper tests.
- Keep the change limited to the DTO and its tests.

CloudEvents 1.0 time requirement:
https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time

Does this scope fit? If approved, I would like to implement it after any earlier accepted CloudEvents follow-up. I will wait for maintainer direction before coding.

shivv23:
Fixes #9325

## Description

`CloudEventDto.time` is an `Instant` with no field-level serialization metadata. It only serializes as an RFC 3339 string when the `ObjectMapper` happens to have `WRITE_DATES_AS_TIMESTAMPS` disabled. A mapper with `JavaTimeModule` registered but the global timestamp setting left at its default emits `time` as a numeric epoch value, violating the CloudEvents 1.0 wire contract.

The CloudEvents 1.0 spec requires the optional `time` attribute, when present, to be an RFC 3339 timestamp. The DTO advertises itself as CloudEvents-compliant, so the wire format should not depend on unrelated global mapper configuration.

## Changes

- Annotate `CloudEventDto.time` with `@JsonFormat(shape = JsonFormat.Shape.STRING)`, forcing string serialization regardless of the mapper's global `WRITE_DATES_AS_TIMESTAMPS` setting.
- Deserialization is unaffected (`Instant` continues to parse from RFC 3339 strings).

## Testing

- Added `CloudEventDtoTest.testSerializationWithTimestampsEnabled`, which serializes with a mapper that leaves `WRITE_DATES_AS_TIMESTAMPS` enabled and asserts `time` is still emitted as `"2024-01-01T00:00:00Z"`.
- Ran locally with JDK 21:
  - `CloudEventDtoTest` — Tests run: 10, Failures: 0, Errors: 0
  - `CloudEventDtoInjectedMapperTest` — Tests run: 4, Failures: 0, Errors: 0
  - `CloudEventConverterTest` — Tests run: 6, Failures: 0, Errors: 0


app/src/test/java/io/apicurio/registry/events/dto/CloudEventDtoTest.java
```diff
@@ -85,6 +85,29 @@ public void testSerialization() throws Exception {
         assertEquals(Instant.parse("2024-01-01T00:00:00Z"), deserialized.getTime());
     }
 
+    @Test
+    public void testSerializationWithTimestampsEnabled() throws Exception {
+        // A mapper that leaves WRITE_DATES_AS_TIMESTAMPS enabled (the JavaTimeModule default)
```
paoloantinori: Minor wording nit, easy win: `WRITE_DATES_AS_TIMESTAMPS` defaults to true on `ObjectMapper` itself; registering `JavaTimeModule` only contributes the serializers and deserializers and does not touch that flag. Consider rewording the comment to 'the ObjectMapper default' so future readers don't infer that `JavaTimeModule` is what enables numeric timestamps. The test setup itself is correct.

paoloantinori (commented):
No blocker or major issues; the fix is correct and well-tested. Annotating `CloudEventDto.time` with `@JsonFormat(shape = STRING)` makes the RFC 3339 wire contract independent of the mapper's global `WRITE_DATES_AS_TIMESTAMPS` setting, and the new test reproduces the exact failure condition (timestamps left on) and asserts the ISO output. The mechanism is sound: Jackson's `DateTimeSerializerBase.useTimestamp` returns `false` when the shape is STRING, so the field serializes via `ISO_INSTANT` regardless of the global feature.

- This is the right layering. The contract now lives on the DTO rather than on any caller's `ObjectMapper`, so it covers every serialization path, including the injected-mapper case exercised in `CloudEventDtoInjectedMapperTest` (`CloudEventDto.java:48`).
- Deserialization is correctly left intact. `InstantDeserializer` does not gate on the STRING shape, so a numeric epoch value still parses; the round-trip at the end of `testSerializationWithTimestampsEnabled` exercises that parse path.
- The wire output does change for any deployment that had `WRITE_DATES_AS_TIMESTAMPS` left on (numeric `time` becomes RFC 3339). That is the intended compliance fix for #9325, not a regression, and the project's own round-trip parse still passes.
- One easy-win nit: the test comment attributes the timestamps-on state to 'the JavaTimeModule default,' but that default belongs to `ObjectMapper`; `JavaTimeModule` does not set it (`CloudEventDtoTest.java:90`).

shivv23:
@EricWittmann All possible fixes made, please have a look.

app/src/test/java/io/apicurio/registry/events/dto/CloudEventDtoTest.java
shivv23: @paoloantinori fixed

EricWittmann:
Thanks for the contribution, @shivv23! Merged.

PR #9397, opened 2026-08-08
fix(ui): resolve CWE-346 postMessage targetOrigin vulnerability (#9396)

### LINKED ISSUE #9396: [Security] Fix insecure cross-origin postMessage handling in API editors and visualizer (CWE-346)
While reviewing the Registry UI, I identified potentially unsafe cross-origin `window.postMessage()` usage in the OpenAPI editor, AsyncAPI editor, and OpenAPI visualizer integrations.

These components send API schema content to iframe-based external applications using `postMessage(message, "*")`. By using the wildcard (`"*"` ) as the `targetOrigin`, I am effectively removing the browser-level restriction that normally ensures a message is delivered only to a specific trusted origin.

This becomes especially important because the iframe URLs are obtained from runtime configuration and may point to external origins. If an editor or visualizer URL is misconfigured, unexpectedly redirected, or otherwise controlled by an unintended origin, the current implementation does not enforce any explicit origin restriction when sending sensitive schema data.

In addition, I noticed that the corresponding `message` event handlers process incoming messages without explicitly validating `event.origin`.

Overall, this appears to represent a **CWE-346 (Origin Validation Error)** / cross-origin message validation weakness and should be treated as a potential security data-exposure risk.

## Affected Components

The following UI components were identified:

1. `ui/ui-app/src/editors/OpenApiEditor.tsx`
2. `ui/ui-app/src/editors/AsyncApiEditor.tsx`
3. `ui/ui-app/src/app/pages/version/components/tabs/visualizers/OpenApiVisualizer.tsx`

A global search of the UI codebase shows these three relevant uses of:

```typescript
postMessage(message, "*")
```

### 1. Wildcard `targetOrigin`

In the affected components, I observed that messages are being sent using the wildcard origin.

For example, `OpenApiEditor.tsx` contains logic equivalent to:

```typescript
ref.current.contentWindow.postMessage(message, "*");
```

The message being sent contains the API artifact currently being edited.

For example:

```typescript
const message: any = {
    type: "apicurio-editingInfo",
    data: {
        content: {
            type: "OPENAPI",
            value: value
        }
    }
};
```

Here, `value` contains the full OpenAPI schema being edited.

The same pattern is also present in the AsyncAPI editor and OpenAPI visualizer.

### 2. The iframe destination is runtime-configurable

I also observed that the iframe URL is obtained through the Registry UI configuration.

For example, `OpenApiEditor.tsx` retrieves the editor URL via `ConfigService`:

```typescript
const config = useConfigService();
let editorsUrl: string = config.uiEditorsUrl();
```

Since these editor/visualizer URLs are configurable at runtime, they may point to external origins and are not guaranteed to be same-origin with the Registry UI.

Because of this, relying on `postMessage("*")` removes an important browser-level security boundary.

### 3. Incoming messages are not explicitly origin-validated

The affected components also register `message` event listeners to receive updates from the iframe.

For example, the OpenAPI editor contains logic equivalent to:

```typescript
const eventListener = (event) => {
    if (event.data && event.data.type === "apicurio_onChange") {
        // process message
    }
};
```

While the handler checks the message type, I did not find any validation of:

```typescript
event.origin
```

against the expected editor origin.

As a result, there is no explicit origin validation on the receiving side either.

## Potential Security Impact

OpenAPI and AsyncAPI documents can contain sensitive information about an organization's internal systems, including:

* Internal API endpoints and service URLs
* Proprietary data models
* Authentication and authorization definitions
* Security schemes
* Internal microservice architecture
* Request/response structures
* Other implementation-specific API metadata

Because the current implementation uses a wildcard `targetOrigin`, the browser is not instructed to restrict delivery of the message to a specific trusted origin.

If the iframe destination is misconfigured, compromised, redirected, or otherwise controlled by an unintended origin, the API schema contained in the `postMessage` payload could potentially be exposed.

Additionally, since incoming messages are not validated using `event.origin`, there is also a risk that messages from unexpected origins could be processed if they reach the event listener.

This issue was identified through **static code analysis and code review**. I have not attempted to exploit this against a live deployment, so this should be treated as a **potential security vulnerability requiring validation**, rather than a confirmed active exploit.

## Proposed Solution

I believe cross-origin messaging should be updated to enforce strict origin validation in both directions.

### 1. Restrict the sender-side `targetOrigin`

Instead of:

```typescript
ref.current.contentWindow.postMessage(message, "*");
```

I would derive the expected origin from the configured iframe URL:

```typescript
const expectedOrigin =
    new URL(editorAppUrl(), window.location.origin).origin;

ref.current.contentWindow?.postMessage(
    message,
    expectedOrigin
);
```

This change should be applied consistently across:

* `OpenApiEditor.tsx`
* `AsyncApiEditor.tsx`
* `OpenApiVisualizer.tsx`

### 2. Validate `event.origin` on incoming messages

I would also ensure that incoming messages are only processed if they originate from the expected iframe origin:

```typescript
const expectedOrigin =
    new URL(editorAppUrl(), window.location.origin).origin;

const eventListener = (event: MessageEvent) => {
    if (event.origin !== expectedOrigin) {
        return;
    }

    if (event.data && event.data.type === "apicurio_onChange") {
        // process trusted message
    }
};
```

This provides defense-in-depth by securing both outgoing and incoming cross-origin communication.

## Security Considerations

The expected origin should be derived dynamically from the configured iframe URL rather than hard-coded, since the Registry supports configurable editor and visualizer deployments.

The fix should also preserve existing behavior for:

* Same-origin development setups
* External Apicurio Studio deployments
* Configured OpenAPI visualizers
* URLs with paths and query parameters
* Existing iframe loading behavior

Only the origin validation logic should be tightened.

## Verification Plan

After applying the fix, I would verify the following:

### OpenAPI Editor

* Open an OpenAPI artifact
* Open the Editor tab
* Confirm schema loads correctly in the iframe
* Modify the schema
* Confirm updates are still received by the Registry UI
* Confirm saving still works

### AsyncAPI Editor

* Open an AsyncAPI artifact
* Open the Editor tab
* Confirm schema loads correctly
* Modify the schema
* Confirm updates are received correctly

### OpenAPI Visualizer

* Open an OpenAPI artifact
* Navigate to the Documentation tab
* Confirm the visualizer renders correctly

### Origin Validation

* Confirm messages are sent using the configured iframe origin instead of `"*"`
* Confirm messages from unexpected origins are rejected
* Confirm messages from the expected origin continue to work

### Automated Verification

* Run the existing UI build and test suite

## Additional Investigation

I performed a repository-wide search for `event.origin` usage in the UI messaging code and did not find an existing consistent pattern for origin validation in these iframe integrations.

Because of this, I believe this may be a good opportunity to introduce a standardized and reusable approach for secure cross-origin iframe communication across the Registry UI.

nishant-mazumder:
Fixes #9396

This PR addresses the CWE-346 cross-origin `postMessage` vulnerability identified in the Registry UI's iframe-based API editor and visualizer integrations.

The affected components previously used `postMessage(message, "*")` when sending API schema content to external iframe applications. The corresponding `message` event handlers also did not explicitly validate the origin of incoming messages.

This change introduces **defense-in-depth origin validation** for all three affected integrations.

### Changes Made

Updated:

* `ui/ui-app/src/editors/OpenApiEditor.tsx`
* `ui/ui-app/src/editors/AsyncApiEditor.tsx`
* `ui/ui-app/src/app/pages/version/components/tabs/visualizers/OpenApiVisualizer.tsx`

#### 1. Strict `targetOrigin`

Replaced the wildcard `"*"` `targetOrigin` with the origin derived from the configured editor/visualizer URL.

For example:

```typescript
const targetOrigin = new URL(
    editorAppUrl(),
    window.location.origin
).origin;

ref.current.contentWindow?.postMessage(
    message,
    targetOrigin
);
```

This ensures that API schema data is only sent to the expected iframe origin.

#### 2. Incoming `event.origin` validation

Added validation of `event.origin` to the corresponding `message` event handlers.

Messages received from origins other than the configured editor/visualizer origin are rejected before being processed.

This provides defense-in-depth by validating both:

* the destination of outgoing messages
* the origin of incoming messages

### Security Impact

The previous wildcard `targetOrigin` allowed the browser to deliver the schema payload regardless of the origin currently loaded inside the iframe.

The updated implementation restricts cross-origin communication to the configured trusted origin, reducing the risk of API schema data being exposed if an iframe destination is unexpectedly redirected or controlled by an unintended origin.

The fix preserves support for externally hosted editor and visualizer deployments because the expected origin is derived dynamically from the configured URL rather than being hard-coded.

### Verification

* `npm run build` completed successfully.
* No TypeScript compilation errors were reported.
* Verified that the changes are applied consistently across all three affected iframe integrations.
* Confirmed that the existing iframe communication flow continues to use the configured editor/visualizer URLs.

### How to Test / Reviewer Checklist

Since I do not have a full local Apicurio Studio deployment set up to test the live iframe integration, I verified this via static code analysis and a clean `npm run build`.

Reviewers with a local environment can manually confirm the following:
- OpenAPI Editor continues to load API schemas correctly.
- AsyncAPI Editor continues to load API schemas correctly.
- OpenAPI Visualizer continues to render API schemas correctly.
- Changes from the expected iframe origin continue to be processed.
- Messages from unexpected origins are rejected.

### DIFF
--- a/ui/ui-app/src/app/pages/version/components/tabs/visualizers/OpenApiVisualizer.tsx
+++ b/ui/ui-app/src/app/pages/version/components/tabs/visualizers/OpenApiVisualizer.tsx
@@ -1,6 +1,7 @@
-import { FunctionComponent, useEffect, useRef } from "react";
+import { FunctionComponent, useEffect, useRef, useMemo } from "react";
 import { ConfigService, useConfigService } from "@services/useConfigService.ts";
 import { LoggerService, useLoggerService } from "@services/useLoggerService.ts";
+import { deriveOrigin } from "@utils/url.utils.ts";
 
 export type OpenApiVisualizerProps = {
     spec: any;
@@ -24,6 +25,10 @@ export const OpenApiVisualizer: FunctionComponent<OpenApiVisualizerProps> = (pro
 
     logger.info("[OpenApiVisualizer] OAI docs URL: ", oaiDocsUrl());
 
+    const expectedOrigin = useMemo(() => {
+        return deriveOrigin(oaiDocsUrl(), window.location.origin);
+    }, [config]);
+
     const sendSpecToIframe = (spec: Record<string, unknown>): void => {
         if (ref.current?.contentWindow) {
             const message = {
@@ -33,7 +38,9 @@ export const OpenApiVisualizer: FunctionComponent<OpenApiVisualizerProps> = (pro
                     content: spec
                 }
             };
-            ref.current.contentWindow.postMessage(message, "*");
+            if (expectedOrigin) {
+                ref.current.contentWindow.postMessage(message, expectedOrigin);
+            }
         }
     };
 
@@ -47,6 +54,9 @@ export const OpenApiVisualizer: FunctionComponent<OpenApiVisualizerProps> = (pro
     // message listener is set up, causing the initial postMessage to be lost.
     useEffect(() => {
         const handler = (evt: MessageEvent): void => {
+            if (!expectedOrigin || evt.origin !== expectedOrigin) {
+                return;
+            }
             if (evt.data?.type === "apicurio-docs-ready" && props.spec && Object.keys(props.spec).length > 0) {
                 sendSpecToIframe(props.spec);
             }
--- a/ui/ui-app/src/editors/AsyncApiEditor.tsx
+++ b/ui/ui-app/src/editors/AsyncApiEditor.tsx
@@ -1,10 +1,10 @@
-import React, { RefObject, useEffect } from "react";
+import React, { RefObject, useEffect, useMemo } from "react";
 import "./AsyncApiEditor.css";
 import { Editor as DraftEditor, EditorProps } from "./editor-types";
 import { parseJson, parseYaml, toJsonString, toYamlString } from "@utils/content.utils.ts";
 import { useConfigService } from "@services/useConfigService.ts";
 import { ContentTypes } from "@models/ContentTypes.ts";
-
+import { deriveOrigin } from "@utils/url.utils.ts";
 
 export type AsyncApiEditorProps = {
     className?: string;
@@ -26,10 +26,16 @@ export const AsyncApiEditor: DraftEditor = (props: AsyncApiEditorProps) => {
     }
 
     // TODO we have a lot of common functionality between the asyncapi and openapi editors.  Need to share!
+    const expectedOrigin = useMemo(() => {
+        return deriveOrigin(editorsUrl, window.location.origin);
+    }, [editorsUrl]);
     useEffect(() => {
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
-        const eventListener: any = (event) => {
+        const eventListener = (event: MessageEvent) => {
+            if (!expectedOrigin || event.origin !== expectedOrigin) {
+                return;
+            }
             if (event.data && event.data.type === "apicurio_onChange") {
                 let newContent: any = event.data.data.content;
                 if (typeof newContent === "object") {
@@ -84,7 +90,9 @@ export const AsyncApiEditor: DraftEditor = (props: AsyncApiEditorProps) => {
                 }
             }
         };
-        ref.current.contentWindow.postMessage(message, "*");
+        if (expectedOrigin) {
+            ref.current.contentWindow.postMessage(message, expectedOrigin);
+        }
     };
 
     return (
--- a/ui/ui-app/src/editors/OpenApiEditor.tsx
+++ b/ui/ui-app/src/editors/OpenApiEditor.tsx
@@ -1,10 +1,11 @@
-import React, { RefObject, useEffect, useState } from "react";
+import React, { RefObject, useEffect, useState, useMemo } from "react";
 import { Editor as DraftEditor, EditorProps } from "./editor-types";
 import "./OpenApiEditor.css";
 import { parseJson, parseYaml, toJsonString, toYamlString } from "@utils/content.utils.ts";
 import { IfNotLoading } from "@apicurio/common-ui-components";
 import { useConfigService } from "@services/useConfigService.ts";
 import { ContentTypes } from "@models/ContentTypes.ts";
+import { deriveOrigin } from "@utils/url.utils.ts";
 
 
 export type OpenApiEditorProps = {
@@ -30,6 +31,10 @@ export const OpenApiEditor: DraftEditor = (props: OpenApiEditorProps) => {
     }
     const ref: RefObject<any> = React.createRef();
 
+    const expectedOrigin = useMemo(() => {
+        return deriveOrigin(editorsUrl, window.location.origin);
+    }, [editorsUrl]);
+
     useEffect(() => {
         setIsLoading(false);
         // systemSvc.getOpenApiVendorExtensions().then(extensions => {
@@ -47,7 +52,10 @@ export const OpenApiEditor: DraftEditor = (props: OpenApiEditorProps) => {
 
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
-        const eventListener: any = (event) => {
+        const eventListener = (event: MessageEvent) => {
+            if (!expectedOrigin || event.origin !== expectedOrigin) {
+                return;
+            }
             if (event.data && event.data.type === "apicurio_onChange") {
                 let newContent: any = event.data.data.content;
                 if (typeof newContent === "object") {
@@ -105,7 +113,9 @@ export const OpenApiEditor: DraftEditor = (props: OpenApiEditorProps) => {
                 }
             }
         };
-        ref.current.contentWindow.postMessage(message, "*");
+        if (expectedOrigin) {
+            ref.current.contentWindow.postMessage(message, expectedOrigin);
+        }
     };
 
     return (
--- a/ui/ui-app/src/utils/url.utils.ts
+++ b/ui/ui-app/src/utils/url.utils.ts
@@ -1,0 +1,17 @@
+/**
+ * Safely derives the origin from a configured URL, handling potential malformed URLs gracefully.
+ * @param url The configured URL (e.g. from config.js)
+ * @param fallbackOrigin The origin to fallback to if the URL is relative (usually window.location.origin)
+ * @returns The parsed origin, or undefined if parsing fails
+ */
+export const deriveOrigin = (url: string | undefined | null, fallbackOrigin: string): string | undefined => {
+    if (!url) {
+        return undefined;
+    }
+    try {
+        return new URL(url, fallbackOrigin).origin;
+    } catch (e) {
+        console.error("[url.utils] Failed to parse origin from URL: ", url, e);
+        return undefined;
+    }
+};


ui/ui-app/src/editors/OpenApiEditor.tsx
```diff
@@ -47,7 +47,11 @@ export const OpenApiEditor: DraftEditor = (props: OpenApiEditorProps) => {
 
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
-        const eventListener: any = (event) => {
+        const eventListener = (event: MessageEvent) => {
+            const expectedOrigin = new URL(editorsUrl, window.location.origin).origin;
```
paoloantinori: This handler recomputes `new URL(editorsUrl, window.location.origin).origin` on every incoming message, and `apicurio_onChange` fires on every edit. Compute the origin once with `useMemo` keyed on `editorsUrl`. That also removes the risk of a `TypeError` from `new URL` if the configured URL is ever malformed, which would throw inside the handler on every message and disable the listener. The same applies to `AsyncApiEditor.tsx:33` and `OpenApiVisualizer.tsx:51`.

ui/ui-app/src/editors/AsyncApiEditor.tsx
```diff
@@ -84,7 +88,8 @@ export const AsyncApiEditor: DraftEditor = (props: AsyncApiEditorProps) => {
                 }
             }
         };
-        ref.current.contentWindow.postMessage(message, "*");
+        const targetOrigin = new URL(editorAppUrl(), window.location.origin).origin;
```
paoloantinori: Outgoing targets the origin of `editorAppUrl()`, but the incoming check at line 33 validates against `editorsUrl`. If those two ever resolve to different origins, the editor's `onChange` messages are silently rejected and edits never propagate back. I could not confirm from the diff whether they share an origin (the source files are not available in this review environment), so please verify, or derive both directions from a single source. The visualizer already does this with `oaiDocsUrl()`.

ui/ui-app/src/app/pages/version/components/tabs/visualizers/OpenApiVisualizer.tsx
```diff
@@ -47,6 +48,10 @@ export const OpenApiVisualizer: FunctionComponent<OpenApiVisualizerProps> = (pro
     // message listener is set up, causing the initial postMessage to be lost.
     useEffect(() => {
         const handler = (evt: MessageEvent): void => {
+            const expectedOrigin = new URL(oaiDocsUrl(), window.location.origin).origin;
```
paoloantinori: Same per-event recomputation as the editors: `new URL(oaiDocsUrl(), window.location.origin).origin` runs on every message here. Worth memoizing once. This file is already symmetric (it uses `oaiDocsUrl()` for both directions), which is the pattern the two editor components should follow.

paoloantinori (commented):
The fix correctly closes the CWE-346 gap: replacing the wildcard `targetOrigin` with the origin derived from the configured URL, plus validating incoming `event.origin`, is the right pattern, applied consistently across all three iframe integrations, and deriving the origin dynamically preserves support for externally hosted deployments. A few points to tighten before merge.

- The expected origin is recomputed inside the message handler on every event (`new URL(...).origin` at `OpenApiVisualizer.tsx:51`, `AsyncApiEditor.tsx:33`, `OpenApiEditor.tsx:51`). Editor `onChange` events fire on every edit, so this parses the URL on every message; compute it once with `useMemo`. This also removes the risk that `new URL` throws on a malformed config URL inside the hot handler, which would disable the listener entirely.
- The two editor components validate the two directions against different sources: outgoing uses `editorAppUrl()` (`AsyncApiEditor.tsx:91`, `OpenApiEditor.tsx:112`) while incoming uses `editorsUrl` (`AsyncApiEditor.tsx:33`, `OpenApiEditor.tsx:51`). The visualizer is symmetric (`oaiDocsUrl()` both ways). If `editorAppUrl()` and `editorsUrl` ever resolve to different origins, the editor's `onChange` replies are silently dropped and edits stop propagating back. I could not confirm from the diff whether they share an origin, so please verify, or derive both directions from one source to avoid drift.
- The `new URL(x, window.location.origin).origin` snippet is repeated six times across the three files. A small shared helper would remove the duplication and centralize the malformed-URL handling.

ui/ui-app/src/editors/OpenApiEditor.tsx
nishant-mazumder: Thanks for the review! I've extracted the URL parsing into a shared deriveOrigin helper in url.utils.ts and wrapped the origin calculation in useMemo. This ensures the origin is computed once when the configuration changes rather than on every incoming message.

ui/ui-app/src/editors/AsyncApiEditor.tsx
nishant-mazumder: I verified that editorAppUrl() ultimately uses the same editorsUrl configuration. I've also updated the implementation so both incoming and outgoing message validation use the same memoized expectedOrigin derived from editorsUrl, keeping a single source of truth and avoiding any potential drift.

ui/ui-app/src/app/pages/version/components/tabs/visualizers/OpenApiVisualizer.tsx
nishant-mazumder: I've applied the same useMemo approach here so the expected origin is calculated when the configuration changes rather than being recomputed for every incoming message.

ui/ui-app/src/editors/OpenApiEditor.tsx
```diff
@@ -47,7 +52,10 @@ export const OpenApiEditor: DraftEditor = (props: OpenApiEditorProps) => {
 
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
-        const eventListener: any = (event) => {
+        const eventListener = (event: MessageEvent) => {
+            if (expectedOrigin && event.origin !== expectedOrigin) {
```
paoloantinori: When `expectedOrigin` is `undefined` (unparseable config URL, `url.utils.ts:13`), this guard short-circuits and the handler accepts messages from any origin, which is the wildcard-acceptance behavior this PR removes. Consider failing closed here: if the origin can't be derived, reject all incoming messages rather than accepting all.

ui/ui-app/src/app/pages/version/components/tabs/visualizers/OpenApiVisualizer.tsx
```diff
@@ -47,6 +54,9 @@ export const OpenApiVisualizer: FunctionComponent<OpenApiVisualizerProps> = (pro
     // message listener is set up, causing the initial postMessage to be lost.
     useEffect(() => {
         const handler = (evt: MessageEvent): void => {
+            if (expectedOrigin && evt.origin !== expectedOrigin) {
```
paoloantinori: Same fail-open shape as the editors: `expectedOrigin` undefined skips the origin check entirely. The outgoing `sendSpecToIframe` path below correctly fails closed (the `if (expectedOrigin)` guard at line 40), but this asymmetry means a bad config both widens the incoming attack surface and silently breaks the outgoing send. Worth making both directions agree.

paoloantinori (commented):
Re-read after the rebase to `c1ca6de`. The `useMemo` change, the shared `deriveOrigin` helper, and deriving both directions from `editorsUrl` all landed cleanly, so the earlier points are addressed. One new thing the helper introduces worth tightening before merge.

`deriveOrigin` returns `undefined` when the configured URL is missing or malformed (the `catch` branch in `url.utils.ts:13`). That `undefined` then flows into every guard as `if (expectedOrigin && ...)`:

- **Incoming handlers fail open.** When `expectedOrigin` is `undefined`, the `event.origin !== expectedOrigin` check is short-circuited past, so the handler processes messages from any origin. That silently reverts to the exact wildcard-acceptance behavior this PR removes, but only on misconfigured deployments where the origin can't be parsed, which is the worst time for it to fail quietly. A malformed `editors.url` in `config.js` is a realistic misconfiguration.
- **Outgoing path fails closed.** `if (expectedOrigin) { postMessage(...) }` (OpenApiEditor.tsx:116, AsyncApiEditor.tsx, OpenApiVisualizer.tsx) drops the message entirely, so the editor loads blank with no error surfaced.

The asymmetry (incoming open, outgoing closed) means a bad config both widens the attack surface and breaks the feature at once. Two options: fail closed on both sides (treat an unparseable origin as "reject everything" rather than "accept everything"), or validate the configured URLs once at startup and refuse to render the iframe integrations if they don't resolve. The second is stronger since it turns a silent runtime degradation into a loud deployment error.

Happy to re-read once pushed.

ui/ui-app/src/editors/OpenApiEditor.tsx
nishant-mazumder: Thanks for your suggestion, I have updated the guard to strictly fail closed (if (!expectedOrigin || event.origin !== expectedOrigin)). Now, if the origin can't be safely derived, it explicitly rejects all incoming messages instead of falling back to wildcard behavior.

ui/ui-app/src/app/pages/version/components/tabs/visualizers/OpenApiVisualizer.tsx
nishant-mazumder: I've applied the exact same fail-closed logic here as well so the incoming and outgoing paths are perfectly symmetric. If a bad config causes expectedOrigin to be undefined, the component now safely refuses to process incoming messages just like it refuses to send outgoing ones.

nishant-mazumder:
Hi @EricWittmann, I’ve addressed the review feedback and pushed the latest changes. The PR has been waiting for a few days, so I wanted to check whether there’s anything else you’d like me to update or verify before it can be merged.

EricWittmann:
Thanks for this contribution, @nishant-mazumder! Great security hardening work — the fix is clean and well-applied across all three affected components. Merged!

PR #9398, opened 2026-08-08
fix(serde-protobuf): serialize default message index [0] as Confluent-compatible 0x00

### LINKED ISSUE #9429: MessageIndexesUtil: serialize default message index [0] as Confluent-compatible 0x00
## Context

io.apicurio.registry.serde.protobuf.MessageIndexesUtil handles the protobuf message-index var-ints that Apicurio writes/reads for Confluent wire compatibility.

In Confluent's MessageIndexes, the default index [0] (the first message type) is encoded as a single x00 var-int, and a bare x00 is always decoded back to [0]. Apicurio's reader already matches this (
eadFrom maps size == 0 to List.of(0)), which is intentional and must be preserved.

The gap is on the writer side: writeTo serializes [0] as [size=1][index=0] (x02 0x00) instead of Confluent's optimized single x00 byte. This is readable by Confluent but not byte-identical to its output.

## Proposed fix

Update writeTo to serialize the default index [0] as a single x00 byte, matching Confluent's MessageIndexes.toByteArray() optimization. Leave 
eadFrom semantics unchanged.

Related PR: #9398

Nitish08-08:
Fixes #9429

## Summary

MessageIndexesUtil.writeTo serialized the default index [0] (the first message type) as [size=1][index=0] (x02 0x00). Confluent's MessageIndexes.toByteArray() optimizes the default index [0] to a single x00 var-int.

This PR makes Apicurio's writer emit the same Confluent-compatible byte layout: [0] is now written as a bare x00, while all other index lists keep the [size][indexes...] encoding.

eadFrom intentionally maps a bare x00 to [0] (Confluent's DEFAULT_INDEX); that behavior is unchanged.

## Tests
Added MessageIndexesUtilTest verifying:
- writeTo(List.of(0)) produces exactly x00 (Confluent-compatible).
- reading a single x00 byte returns List.of(0).
- round-trips for the default, multiple, and negative indexes.

### DIFF
--- a/serdes/generic/serde-common-protobuf/src/main/java/io/apicurio/registry/serde/protobuf/MessageIndexesUtil.java
+++ b/serdes/generic/serde-common-protobuf/src/main/java/io/apicurio/registry/serde/protobuf/MessageIndexesUtil.java
@@ -10,9 +10,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 public class MessageIndexesUtil {
+
+    // Mirrors Confluent's MessageIndexes.DEFAULT_INDEX: the index of the first message type.
+    public static final List<Integer> DEFAULT_INDEX = Collections.singletonList(0);
 
     private static IllegalArgumentException illegalVarintException(int value) {
         throw new IllegalArgumentException("Varint is too long, the most significant bit in the 5th byte is set, " +
@@ -53,6 +57,12 @@ public class MessageIndexesUtil {
     }
 
     public static void writeTo(List<Integer> indexes, OutputStream out) throws IOException {
+        if (indexes.equals(DEFAULT_INDEX)) {
+            // Confluent-compatible optimization: the default index [0] (the first message
+            // type) is encoded as a single 0x00 var-int instead of [size][indexes].
+            writeVarInt(0, out);
+            return;
+        }
         writeVarInt(indexes.size(), out);
 
         for (Integer index : indexes) {
@@ -96,6 +106,8 @@ public class MessageIndexesUtil {
     public static List<Integer> readFrom(InputStream in) throws IOException {
         int size = readVarInt(in);
         if (size == 0) {
+            // A bare 0x00 var-int is Confluent's optimized encoding for the default
+            // index [0] (the first message type), so it must map to [0], not an empty list.
             return List.of(0);
         }
         List<Integer> indexes = new ArrayList<>(size);
--- a/serdes/generic/serde-common-protobuf/src/test/java/io/apicurio/registry/serde/protobuf/MessageIndexesUtilTest.java
+++ b/serdes/generic/serde-common-protobuf/src/test/java/io/apicurio/registry/serde/protobuf/MessageIndexesUtilTest.java
@@ -1,0 +1,56 @@
+package io.apicurio.registry.serde.protobuf;
+
+import org.junit.jupiter.api.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Unit tests for MessageIndexesUtil verifying var-int index (de)serialization and
+ * Confluent wire-format compatibility for the default message index.
+ */
+public class MessageIndexesUtilTest {
+
+    @Test
+    public void testWriteDefaultIndexIsConfluentCompatible() throws IOException {
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        MessageIndexesUtil.writeTo(List.of(0), out);
+        // Confluent encodes the default index [0] (the first message type) as a bare 0x00 var-int.
+        assertArrayEquals(new byte[]{0x00}, out.toByteArray());
+    }
+
+    @Test
+    public void testReadSingleZeroByteIsDefaultIndex() throws IOException {
+        // Confluent's optimized encoding for [0] is a single 0x00 byte.
+        List<Integer> indexes = MessageIndexesUtil.readFrom(new ByteArrayInputStream(new byte[]{0x00}));
+        assertEquals(List.of(0), indexes);
+    }
+
+    @Test
+    public void testWriteReadRoundTripDefaultIndex() throws IOException {
+        roundTrip(List.of(0));
+    }
+
+    @Test
+    public void testWriteReadRoundTripMultipleIndexes() throws IOException {
+        roundTrip(List.of(3, 1, 2, 7));
+    }
+
+    @Test
+    public void testWriteReadRoundTripNegativeIndexes() throws IOException {
+        roundTrip(List.of(0, -1, 42, -100));
+    }
+
+    private void roundTrip(List<Integer> indexes) throws IOException {
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        MessageIndexesUtil.writeTo(indexes, out);
+
+        List<Integer> result = MessageIndexesUtil.readFrom(new ByteArrayInputStream(out.toByteArray()));
+        assertEquals(indexes, result, "Read indexes should match written indexes after a round trip");
+    }
+}


aclfe (commented):
What is the upstream issue for the pr?

Nachiket-Roy:
This introduces a regression that breaks wire-compatibility with Confluent Schema Registry's Protobuf SerDe.

Nachiket-Roy:
So, I suggest closing this PR. Please open an issue before submitting a PR, as PRs without a linked issue may be automatically closed in the future.

Nitish08-08:
Thanks for the feedback @Nachiket-Roy. I've opened issue #9429 and linked it to this PR with Fixes #9429.

Regarding the wire-compatibility concern: the change is limited to the return value of 
eadFrom, which is currently discarded at all three call sites in ProtobufDeserializer (it is only used to advance the stream position), so it does not alter what the serde reads from the wire or what it emits. If there is a specific Confluent interop scenario that depends on the size == 0 -> List.of(0) behavior, please point me to it and I'll adjust the approach.

Nachiket-Roy:
While it is true that `ProtobufDeserializer` currently discards the returned list, `MessageIndexesUtil` is a public utility, so changing its return value has compatibility implications in two ways:

**Semantic correctness:** In the Confluent wire format, a bare `0x00` byte is the optimized encoding for `List.of(0)`, representing the first message type. Mapping it to `List.of()` instead is semantically incorrect and could break downstream consumers, custom SerDes, or interceptors that use this public utility to inspect message types.

**Future type resolution:** Confluent uses message indexes to resolve message types in schemas containing multiple messages. If Apicurio adds full multi-message type resolution in the future, interpreting `0x00` as `List.of()` rather than `List.of(0)` would prevent the primary message type from being resolved correctly.

To maintain wire compatibility, `0x00` should therefore deserialize to `List.of(0)`. If we want to support both cases cleanly, I think the better approach would be to update Apicurio's writer to also serialize `List.of(0)` as `0x00`, matching Confluent's optimization, rather than changing the reader's mapping.

paoloantinori:
To sharpen the wire-compatibility point @Nachiket-Roy raised: the byte `0x00` is overloaded between the two formats, and that's where the real ambiguity lives.

- In Apicurio's own `MessageIndexesUtil` wire format, `writeTo(List.of())` emits a single `0x00` (the size varint, zig-zag of 0), and `writeTo(List.of(0))` emits `0x02 0x00` (size=1 zig-zags to `0x02`, then the index 0 zig-zags to `0x00`). So within Apicurio's format, the old `readFrom` *was* genuinely asymmetric — `List.of()` → `0x00` → `List.of(0)` — and this PR restores symmetry. The author's round-trip claim holds for Apicurio-produced bytes.
- The trouble is the Confluent interop path. Confluent's optimized encoding for the single index `[0]` is a bare `0x00`, which is byte-identical to Apicurio's encoding of an empty index list. After this change, `readFrom` on a Confluent-produced prefix for the first message type returns `List.of()` instead of `List.of(0)`.

Today this is latent rather than breaking: all three `readFrom` call sites in `ProtobufDeserializer` (lines 196, 214, 391) discard the returned list — the inline comment at line 213 even says "Currently we do not use it for anything." So nothing inside Apicurio misbehaves with either return value.

The risk is exactly Nachiket's second point: if multi-message type resolution is ever implemented against `readFrom`'s result, a Confluent-produced `0x00` would resolve to "no message type" instead of "first message type." That's a silent failure under interop, not just a semantic quibble.

So I'd read this PR as internally consistent but incomplete. The change fixes the Apicurio-internal asymmetry while leaving the cross-format `0x00` ambiguity unresolved, and it silently shifts the interpretation in the direction that hurts Confluent interop. Worth deciding explicitly which format's semantics `readFrom` is supposed to honor before this lands — Nachiket's writer-side suggestion (have Apicurio's `writeTo` also emit the Confluent-style `0x00` for `[0]`, and keep `readFrom` mapping `0x00` → `[0]`) keeps both the internal round-trip and Confluent interop consistent at the cost of a one-byte savings, and feels like the safer target.

Nitish08-08:
Thanks @Nachiket-Roy, you're right on both counts. I reverted the 
eadFrom change and instead updated writeTo to match Confluent's optimization.

**What changed now:**
- 
eadFrom is back to mapping a bare x00 var-int to List.of(0) (Confluent's DEFAULT_INDEX), so that semantic is preserved for downstream consumers/custom SerDes.
- writeTo now serializes the default index [0] as a single x00 byte, matching Confluent's MessageIndexes.toByteArray(), instead of [size=1][index=0] (x02 0x00). Other index lists keep the [size][indexes...] encoding.
- Added tests for the Confluent-compatible byte layout and round-trips, and updated the linked issue #9429 to reflect the real fix (writer-side byte layout, not the reader).

If there's anything else to align with the Confluent interop path, happy to adjust.

Nitish08-08:
Hey @Nachiket-Roy - thanks for the thorough review and for clarifying the Confluent wire-format details, it was a great catch.

This PR is now a small, focused improvement: Apicurio's protobuf serializer emits the default message index [0] as a single x00 byte, byte-for-byte matching what Confluent's MessageIndexes writes, while the reader's behavior stays exactly as before. For the common top-level message case this makes the wire format identical between the two SerDe implementations, which helps anyone mixing Apicurio and Confluent producers/consumers.

The change is minimal (a few lines in MessageIndexesUtil plus round-trip tests), so it's low-risk and easy to review. Happy to adjust if there are other interop scenarios to cover!

Nitish08-08:
Thanks @paoloantinori for the detailed analysis - that's exactly the scenario we've landed on. The PR has been updated to the approach you and @Nachiket-Roy recommended:

- 
eadFrom keeps mapping a bare x00 to List.of(0) (Confluent's DEFAULT_INDEX), so a Confluent-produced prefix for the first message type still resolves correctly.
- writeTo now serializes [0] as a single x00 byte (matching Confluent's MessageIndexes.toByteArray() optimization) instead of x02 0x00, so Apicurio-produced bytes are byte-identical to Confluent's for the default index.
- Other index lists keep the [size][indexes...] encoding.

This keeps the Apicurio-internal round-trip consistent AND preserves Confluent interop, exactly as you outlined. The PR remains a minimal 2-file change (MessageIndexesUtil.java + tests). Appreciate the review!

paoloantinori:
The implementation matches the approach we discussed. `writeTo(List.of(0))` now emits a bare `0x00`, `readFrom` maps it back to `[0]`, and the round-trip tests cover the default, multiple, and negative-index cases. The wire-compatibility concern @Nachiket-Roy raised is resolved on this head: Apicurio-produced bytes are byte-compatible with Confluent's `MessageIndexes.toByteArray()` for the default index, and Confluent-produced bytes still read correctly through the existing `readFrom` path.

No further review feedback from my side.

Nitish08-08:
> This introduces a regression that breaks wire-compatibility with Confluent Schema Registry's Protobuf SerDe.

Yeah you're right, that was a miss on my end. Reverted the readFrom change. Now the fix is writer-side only — writeTo emits 0x00 for [0] matching Confluent, readFrom stays as is. No regression.

Nitish08-08:
> So, I suggest closing this PR. Please open an issue before submitting a PR, as PRs without a linked issue may be automatically closed in the future.

Done, [#9429](https://github.com/Apicurio/apicurio-registry/issues/9429) is linked with Fixes #9429. Won't skip that again.

Nitish08-08:
> While it is true that `ProtobufDeserializer` currently discards the returned list, `MessageIndexesUtil` is a public utility, so changing its return value has compatibility implications in two ways:
> 
> **Semantic correctness:** In the Confluent wire format, a bare `0x00` byte is the optimized encoding for `List.of(0)`, representing the first message type. Mapping it to `List.of()` instead is semantically incorrect and could break downstream consumers, custom SerDes, or interceptors that use this public utility to inspect message types.
> 
> **Future type resolution:** Confluent uses message indexes to resolve message types in schemas containing multiple messages. If Apicurio adds full multi-message type resolution in the future, interpreting `0x00` as `List.of()` rather than `List.of(0)` would prevent the primary message type from being resolved correctly.
> 
> To maintain wire compatibility, `0x00` should therefore deserialize to `List.of(0)`. If we want to support both cases cleanly, I think the better approach would be to update Apicurio's writer to also serialize `List.of(0)` as `0x00`, matching Confluent's optimization, rather than changing the reader's mapping.

That's exactly what I did. writeTo now emits bare 0x00 for [0], readFrom untouched. Tests cover Confluent byte layout + round-trips for default, multi, and negative indexes. All sorted.

Nitish08-08:
> The implementation matches the approach we discussed. `writeTo(List.of(0))` now emits a bare `0x00`, `readFrom` maps it back to `[0]`, and the round-trip tests cover the default, multiple, and negative-index cases. The wire-compatibility concern @Nachiket-Roy raised is resolved on this head: Apicurio-produced bytes are byte-compatible with Confluent's `MessageIndexes.toByteArray()` for the default index, and Confluent-produced bytes still read correctly through the existing `readFrom` path.
> 
> No further review feedback from my side.

Decision is: honor Confluent on reader side, match their optimization on writer side. Updated PR does exactly that. Thanks for the deep dive, made the right call pretty obvious.
Appreciate the thorough review, thanks!

carlesarnal (approved):

Nitish08-08:
check please and if you think there should be any changes than please tell me Thank You . @Nachiket-Roy @carlesarnal @lokeshramchand-ctrl  @EricWittmann

lokeshramchand-ctrl:
LGTM

Nachiket-Roy:
> check please and if you think there should be any changes than please tell me Thank You . @Nachiket-Roy @carlesarnal @lokeshramchand-ctrl

Looks good to me. No concern from my side

Nitish08-08:
check please and if you think there should be any changes than please tell me or if its alright than have my PR merged or closed its been open for almost 1 month Thank You . @Nachiket-Roy @carlesarnal @lokeshramchand-ctrl @EricWittmann

PR #9399, opened 2026-08-09
Close the zip stream when the Confluent export fails

omkarrr2533:
In `Export`, the `ZipOutputStream` is created outside the try-with-resources block and only closed on the success path. If anything between opening it and the final write throws, the `catch (Exception ex)` returns without ever calling `zip.close()`.

The visible effect is worse than a plain leak. `fos` is in try-with-resources, so the file handle is released either way, but `zip.finish()` never runs the archive is left on disk without its central directory. The user sees a `.zip` of plausible size and only finds out it is unusable when they try to import it.

Moving the `ZipOutputStream` into the same try-with-resources fixes both. Resources close in reverse order, so the zip writes its central directory before the underlying file stream closes.

`DataExporter` and `ProtobufExporter` on the server side already use this pattern after #9208 and #9286; this is the last place that still did it by hand.

No test added —> exercising this path needs a live Confluent Schema Registry to export from.


EricWittmann:
Thanks for the contribution, @omkarrr2533! Clean fix — merged.

PR #9401, opened 2026-08-09
fix(ui): clean up remaining stale PatternFly 5 (pf-v5) CSS classes and variables in ui-app

harshitnub077:
Fixes #9400

Replaces all remaining `pf-v5-c-*` CSS classes with `pf-v6-c-*` and legacy `--pf-v5-global--*` design tokens with PatternFly 6 semantic tokens (`--pf-t--global--*`) across the 17 stylesheets in `ui/ui-app`. This restores styling fidelity for borders, backgrounds, and structural components (like breadcrumb dividers and graph views) that degraded during the PF5 -> PF6 migration.

### DIFF
--- a/ui/ui-app/src/app/App.css
+++ b/ui/ui-app/src/app/App.css
@@ -3,7 +3,7 @@
   --registry-page-canvas-bg: var(--pf-t--global--background--color--secondary--default, #f0f0f0);
 }
 
-.pf-c-page {
+.pf-v6-c-page {
   width: 100%;
   height: 100vh;
 }
--- a/ui/ui-app/src/app/App.css
+++ b/ui/ui-app/src/app/App.css
@@ -15,8 +15,8 @@
   background-color: var(--registry-page-canvas-bg);
 }
 
-.artifact-details-main .pf-c-tab-content,
-.branch-details-main .pf-c-tab-content,
+.artifact-details-main .pf-v6-c-tab-content,
+.branch-details-main .pf-v6-c-tab-content,
 .artifact-details-main .pf-v6-c-tab-content,
 .branch-details-main .pf-v6-c-tab-content,
 #pf-tab-section-content-artifact-page-tabs,
@@ -28,10 +28,10 @@
 /* Secondary cards inside detail tabs add a redundant gray frame on the canvas. */
 .artifact-details-main [class*="-tab-content"] .pf-v6-c-card.pf-m-secondary,
 .branch-details-main [class*="-tab-content"] .pf-v6-c-card.pf-m-secondary,
-.artifact-details-main [class*="-tab-content"] .pf-c-card.pf-m-secondary,
-.branch-details-main [class*="-tab-content"] .pf-c-card.pf-m-secondary {
+.artifact-details-main [class*="-tab-content"] .pf-v6-c-card.pf-m-secondary,
+.branch-details-main [class*="-tab-content"] .pf-v6-c-card.pf-m-secondary {
   --pf-v6-c-card--BackgroundColor: transparent;
-  --pf-c-card--BackgroundColor: transparent;
+  --pf-v6-c-card--BackgroundColor: transparent;
   border: none;
   box-shadow: none;
 }
--- a/ui/ui-app/src/app/App.css
+++ b/ui/ui-app/src/app/App.css
@@ -1,6 +1,6 @@
 :root {
   --myApp-global--spacer--md: 30px;
-  --registry-page-canvas-bg: var(--pf-v5-global--BackgroundColor--200, #f0f0f0);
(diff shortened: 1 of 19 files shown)


ui/ui-app/src/app/pages/version/components/tabs/ReferenceGraphView.css
```diff
@@ -51,12 +51,12 @@
 
 /* Edge styling */
 .reference-graph-container .react-flow__edge-path {
-    stroke: var(--pf-v5-global--BorderColor--200, #8a8d90);
+    stroke: var(--pf-t--global--border--color--100, #8a8d90);
```
paoloantinori: This BorderColor--200 to --pf-t--global--border--color--100 mapping looks inverted: in the PF5 scale BorderColor--200 (#8a8d90) is the darker/secondary border, while PF6's border--color--100 is typically the lightest border shade, so the edges may render lighter than intended if that scale direction holds in the PF6 build. More broadly, every --pf-t--global--* token in this file needs verifying against the actual @patternfly/react-core PF6 tokens. An undefined token does not error, it silently falls back to the hardcoded hex (here #8a8d90) and breaks dark-mode theming with no signal. Confirm each token resolves in DevTools computed styles with the PF6 theme applied, including a dark-mode pass.

ui/ui-app/src/app/pages/artifact/ArtifactPage.css
```diff
@@ -22,7 +22,7 @@ div.artifact-page-tabs > ul > li.pf-c-tabs__item.pf-m-current > button.pf-c-tabs
     padding: 12px 24px 0;
 }
 
-.pf-v5-c-breadcrumb__item-divider {
+.pf-v6-c-breadcrumb__item-divider {
```
paoloantinori, line 24: Renaming pf-v5-c-breadcrumb__item-divider to pf-v6-c-breadcrumb__item-divider assumes PF6 still emits that exact sub-class. PF6 restructured several components' internal class names, so a few of these renamed selectors (also ErrorPage.css pf-v6-c-code-editor__code and the pf-v6-c-menu-toggle__text rules across the toolbar files) may no longer match any rendered element and are effectively dead. Worth confirming each renamed component selector still exists in PF6's rendered DOM before trusting that styling fidelity is restored.

ui/ui-app/src/app/App.css
```diff
@@ -1,6 +1,6 @@
 :root {
   --myApp-global--spacer--md: 30px;
-  --registry-page-canvas-bg: var(--pf-v5-global--BackgroundColor--200, #f0f0f0);
+  --registry-page-canvas-bg: var(--pf-t--global--background--color--secondary--default, #f0f0f0);
 }
 
 .pf-c-page {
```
paoloantinori: This PR scopes to pf-v5-* but leaves unversioned pf-c-* selectors untouched in the same files: .pf-c-page here, and pf-c-tabs__item / pf-c-tabs__button in ArtifactPage, BranchPage, GroupPage, and VersionPage. PF6 emits pf-v6-c-* classes, so those pf-c-* rules no longer match anything either. If the stated goal is restoring styling fidelity for these pages, confirm whether the stale pf-c-tabs selectors are handled in a separate change, otherwise they look like dead CSS that belongs in this cleanup.

paoloantinori (commented):
The migration is mechanical and the var() fallback values are correctly preserved, so there is no hard regression in light mode; the open risks are whether the new --pf-t--global--* token names and the prefix-bumped component selectors actually resolve in the PF6 build, since a wrong name fails silently by falling back to the hardcoded hex. Worth verifying before merge:

- Token validity and dark mode (ReferenceGraphView.css:54, App.css:3): every --pf-t--global--* name should be confirmed against the PF6 token set. Undefined tokens fall back to the light-mode hex with no error, which would silently break dark-mode theming. The BorderColor--200 to border--color--100 mapping in particular looks like a scale inversion worth checking.
- Prefix-bumped selectors may be dead (ArtifactPage.css:25, ErrorPage.css:15, the toolbar files): PF6 restructured several internal component classes, so renaming the pf-v5 prefix to pf-v6 does not guarantee the sub-class (breadcrumb __item-divider, code-editor __code, menu-toggle __text) still exists in PF6's rendered output. Confirm against the live DOM before treating fidelity as restored.
- Leftover unversioned pf-c-* selectors (App.css:6, plus pf-c-tabs__item / pf-c-tabs__button in ArtifactPage, BranchPage, GroupPage, and VersionPage): these predate pf-v5 and PF6 does not emit them either. The PR fixes the breadcrumb divider a few lines away but leaves the tabs selector stale, which makes the cleanup scope inconsistent if the stated goal is restoring page styling fidelity.

PR #9402, opened 2026-08-09
Feature/UI configurable url import

### LINKED ISSUE #9042: UI: make the "Import from URL" 5MB content-length cap configurable

anirudhagarwal-dev:
Closes #9042

### Description
This PR makes the UI's "Import from URL" response size cap configurable via the UI config (`apicurio.ui.*`), replacing the previously hard-coded 5 MB limit. It implements **Option A (backend-driven)** so that the limit can be changed without rebuilding the UI. 

By default, the cap remains at 5 MB (5242880 bytes) ensuring backward compatibility for existing deployments.

### Addressing Previous Review Feedback
A previous attempt at this feature ran into a few edge cases that maintainers rightly pointed out. This PR builds on that work and cleanly addresses all feedback:

1. **Bounds Check & Type Coercion**: The `featureUrlImportMaxContentLength()` accessor now strictly coerces the config value to a Number and validates it with `Number.isFinite(value) && value > 0`. It falls back to `5242880` for absent, null, undefined, zero, negative, NaN, or non-numeric string values. This ensures an operator's misconfiguration doesn't silently break Axios.
2. **Latent String Bug Fixed**: Axios expects `maxContentLength` to be a number of bytes. The original string literal (`"5242880"`) only worked via implicit coercion. The new code correctly passes a strict `number`, preventing any latent bugs.
3. **Rules of Hooks Maintained**: Calling `useConfigService()` has been moved to the top of the `useUrlService` hook, resolving transitive hook dependencies rather than conditionally nesting it inside the async fetch callback. 
4. **Test Coverage**: Added `useConfigService.test.ts` (9 test cases) that comprehensively verifies the accessor's fallback logic against absent, null, 0, negative, NaN, and string inputs.

### Implementation Details
* **Backend**: Added `@ConfigProperty(name = "apicurio.ui.features.url-import.max-content-length")` to `UserInterfaceConfigProperties.java`, updated the `openapi.json` schema, and wired the value into `SystemResourceImpl.java` to expose it via `/system/uiConfig`.
* **Frontend**: Updated `FeaturesConfig` and `useConfigService.ts` to parse the value and updated `useUrlService.ts` to dynamically apply the limit to HTTP requests.

### How to Configure
An operator can now increase the limit by passing the following environment variable (or config property) to the backend:
```properties
apicurio.ui.features.url-import.max-content-length=10485760

### CODE REVIEW

**Comment by anirudhagarwal-dev** [comment]
Hi team! 👋 

I'm picking up the work for issue #9042 since the previous attempt stalled out. The hard-coded 5MB limit on "Import from URL" is too restrictive for large OpenAPI specs and JSON schemas, so this PR makes it fully configurable via the backend without requiring a UI rebuild.

I've also made sure to implement all the feedback that maintainers provided in the previous PR attempt:
- **Type Coercion & Bounds Checking**: Added strict numeric coercion and bounds checking (`Number.isFinite` and `> 0`), falling back safely to the 5MB default for any misconfigurations (null, undefined, 0, negative, NaN).
- **Latent Axios Bug Fixed**: Ensure we pass a strict `number` type instead of a `string` to `options.maxContentLength`, matching the Axios API.
- **Rules of Hooks**: Refactored `useUrlService.ts` to call `useConfigService()` properly within the React hook body instead of conditionally inside the callback.
- **Test Coverage**: Added a full suite of Vitest tests in `useConfigService.test.ts` to verify every edge case for the fallback logic.

Let me know if there's anything else you need me to adjust!

File: app/src/main/java/io/apicurio/registry/ui/UserInterfaceConfigProperties.java [hunk 1]
```diff
@@ -52,6 +52,10 @@ public class UserInterfaceConfigProperties {
     @Info(category = CATEGORY_UI, description = "Enabled to show the Agents tab in the UI", availableSince = "3.2.0", experimental = true)
     public Supplier<Boolean> featureAgents;
 
+    @ConfigProperty(name = "apicurio.ui.features.url-import.max-content-length", defaultValue = "5242880")
+    @Info(category = CATEGORY_UI, description = "Maximum content length for URL imports", availableSince = "3.2.0")
+    public String featureUrlImportMaxContentLength;
```
paoloantinori: This property is a raw `String`, so Quarkus performs no numeric validation. A typo like `apicurio.ui.features.url-import.max-content-length=10MB` or `=abc` is accepted at startup, serialized to the UI, coerced to `NaN` by the frontend accessor, and silently falls back to the 5 MB default with no log anywhere. The operator believes the limit was raised but it is unchanged. Declaring it as `Long` (or `Supplier<Long>` to match the dynamic-reload pattern used by `featureAgents`) lets Quarkus reject bad values at startup, makes the OpenAPI schema `integer`, and removes most of the frontend coercion logic.

app/src/main/java/io/apicurio/registry/rest/v3/impl/SystemResourceImpl.java
```diff
@@ -86,6 +86,7 @@ public UserInterfaceConfig getUIConfig() {
                         .draftMutability(restConfig.isArtifactVersionMutabilityEnabled())
                         .agents(uiConfig.featureAgents.get())
                         .searchIndex(esSearchConfig.isEnabled())
+                        .urlImportMaxContentLength(uiConfig.featureUrlImportMaxContentLength)
```
paoloantinori: This builder call requires `UserInterfaceConfig` to expose a `urlImportMaxContentLength` field and builder method, but that model class is not in the changed-files list. Confirm the model is generated from `openapi.json` so the new property produces the field and builder. If the model is hand-maintained, this line will not compile.

ui/ui-app/src/services/useUrlService.ts
```diff
@@ -1,44 +1,44 @@
...
+            url = `https://raw.githubusercontent.com/${org}/${repo}/${branch}/${path}`;
+        }
+
+        console.info("[UrlService] Fetching content from a URL: ", url);
+
+        const endpoint: string = url;
+        const options: any = createOptions({
+            "Accept": "*"
+        });
+        options.maxContentLength = config.featureUrlImportMaxContentLength();
```
paoloantinori, line 39: axios only checks `maxContentLength` against the response `Content-Length` header. Responses sent with chunked transfer-encoding, or from servers that omit the header, bypass this cap entirely. Since making the limit configurable is the point of this change, it is worth at least documenting that the cap is best-effort, or enforcing it by measuring bytes as they stream in.

common/src/main/resources/META-INF/openapi.json
```diff
@@ -8263,6 +8263,10 @@
           "searchIndex": {
             "description": "",
             "type": "boolean"
+          },
+          "urlImportMaxContentLength": {
+            "description": "",
+            "type": "string"
```
paoloantinori: The description is empty and the type is `string`. Add a description such as 'Maximum content length in bytes for URL imports.' If the backend property is switched to a numeric type, set this to `integer` to match.

paoloantinori (commented):
Overall: the feature is wired cleanly and the frontend coercion is well covered by the new test cases, but the backend config is typed as a raw String so invalid values degrade silently instead of failing fast, and the diff bundles unrelated package-lock regeneration that should be split out.

- `UserInterfaceConfigProperties.java:57` - the new property is a raw `String`, so Quarkus does no numeric validation. A typo such as `=10MB` or `=abc` is accepted at startup, serialized to the UI, coerced to `NaN`, and silently falls back to the 5 MB default with no log. The operator believes the limit was raised but it is unchanged. A `Long` (or `Supplier<Long>` to match the dynamic-reload pattern used by `featureAgents`) lets Quarkus reject bad values at startup and makes the OpenAPI schema `integer`, which also removes most of the frontend coercion.
- `SystemResourceImpl.java:89` - the `.urlImportMaxContentLength(...)` builder call requires the `UserInterfaceConfig` response model to expose that field and builder, but that model is not in the changed-files list. Confirm it is generated from `openapi.json`; if it is hand-maintained, this line will not compile.
- `useUrlService.ts:36` - `maxContentLength` is advisory: axios only enforces it against the response `Content-Length` header, so chunked or header-less responses bypass the cap. Since the configurable cap is the point of the PR, this is worth documenting or enforcing by streaming byte counts.
- `openapi.json:8269` - the new property has an empty description and is typed `string`. Add a description, and flip to `integer` if the backend goes numeric.
- `ui-docs/package-lock.json` and `ui-editors/package-lock.json` - these are `npm install` regeneration unrelated to the feature. The `ui-docs` change drops the `libc` (glibc/musl) constraints from several optional platform packages, which could change which optional native binaries are selected on those hosts. Revert both, or move them to a separate PR.

anirudhagarwal-dev:
Hi @paoloantinori, thanks for the review! I've gone ahead and pushed an update addressing all your feedback:

Backend config typing: I've updated UserInterfaceConfigProperties.java so featureUrlImportMaxContentLength is now declared as a Supplier<Long> (matching the featureAgents pattern).
OpenAPI schema: Updated openapi.json so the field is now typed as integer (with a description added).
Frontend coercion cleanup: Since the backend now enforces the type and schema, I was able to remove the manual string coercion logic from useConfigService.ts and updated the Vitest tests accordingly.
Axios behavior documented: I added a comment in useUrlService.ts clarifying that Axios only enforces maxContentLength against the Content-Length header and that the cap is best-effort.
Package lock diffs: I reverted the unrelated package-lock.json changes in ui-docs and ui-editors to keep this PR strictly focused.

paoloantinori:
Re-read at the new head (`deb4e58`); three of the four points are addressed and verified against the code, not just the reply:

- **Backend typing** (`UserInterfaceConfigProperties.java:55`): `Supplier<Long>` with `defaultValue = "5242880"` now lets Quarkus reject non-numeric values at startup, and matches the dynamic-reload pattern used by `featureAgents`. The OpenAPI schema flipping to `integer` with a description also resolves the compilation question on `SystemResourceImpl.java:89`, since the generated `UserInterfaceConfig` model picks up the field from `openapi.json`.
- **axios caveat** (`useUrlService.ts:31`): the comment documenting that `maxContentLength` is only checked against the `Content-Length` header is a reasonable resolution. Streaming byte-count enforcement would be a larger change, and the best-effort cap is now at least explicit in the code.
- **Frontend guard** (`useConfigService.ts:368`): keeping the `typeof number && > 0` check as a defensive layer on top of the backend-typed value is fine.

One point from the earlier review is still open: the `ui-docs/package-lock.json` and `ui-editors/package-lock.json` changes are still bundled in this PR. The `ui-docs` diff drops the `libc` (glibc/musl) constraints from several optional platform packages, and `ui-editors` adds new transitive deps (chokidar/readdirp). Both look unrelated to the URL-import feature. Worth reverting here and landing separately so the feature diff stays focused, but not a blocker for the feature itself.

anirudhagarwal-dev:
ooh, Yes In my last push I had reverted the files locally but didn't cleanly remove them from the PR's commit history.

I just checked them out directly from main and added a new commit to properly drop the package-lock.json diffs. Thanks again for the thorough reviews! Everything should be fully cleared up now. 🚀

Nachiket-Roy:
Duplicate #9360

app/src/main/java/io/apicurio/registry/rest/v3/impl/SystemResourceImpl.java
```diff
@@ -86,6 +86,7 @@ public UserInterfaceConfig getUIConfig() {
                         .draftMutability(restConfig.isArtifactVersionMutabilityEnabled())
                         .agents(uiConfig.featureAgents.get())
                         .searchIndex(esSearchConfig.isEnabled())
+                        .urlImportMaxContentLength(uiConfig.featureUrlImportMaxContentLength.get())
```
EricWittmann, line 89: This line fails compilation:
```
incompatible types: java.lang.Long cannot be converted to java.lang.Integer
```

The OpenAPI schema declares `urlImportMaxContentLength` as `"type": "integer"` (no format), so the generated `UserInterfaceConfigFeatures` bean has an `Integer` field. But `uiConfig.featureUrlImportMaxContentLength.get()` returns `Long`.

Suggested fix: change the config property from `Supplier<Long>` to plain `Long` (see my other comment on that file), and cast here if needed. Or add `"format": "int64"` to the OpenAPI property so the generated bean uses `Long`. Given that a URL import content length will never exceed 2 GB, `Integer` is sufficient — the simplest fix is to use `Integer` on both sides.

app/src/main/java/io/apicurio/registry/ui/UserInterfaceConfigProperties.java
```diff
@@ -52,6 +52,10 @@ public class UserInterfaceConfigProperties {
     @Info(category = CATEGORY_UI, description = "Enabled to show the Agents tab in the UI", availableSince = "3.2.0", experimental = true)
     public Supplier<Boolean> featureAgents;
 
+    @ConfigProperty(name = "apicurio.ui.features.url-import.max-content-length", defaultValue = "5242880")
+    @Info(category = CATEGORY_UI, description = "Maximum content length for URL imports", availableSince = "3.2.0")
+    public Supplier<Long> featureUrlImportMaxContentLength;
```
EricWittmann, line 57: `Supplier<Long>` without `@Dynamic` is inconsistent with the codebase pattern. In this class:
- Properties that are runtime-updatable use `@Dynamic` + `Supplier<T>` (see `featureAgents` above)
- Properties that are set at startup use bare types (`String`, `Optional<Boolean>`, etc.)

A URL import size cap is a deployment-time setting, so this should be a plain `Long` (or `Integer` — see my comment on `SystemResourceImpl.java`):
```java
@ConfigProperty(name = "apicurio.ui.features.url-import.max-content-length", defaultValue = "5242880")
@Info(category = CATEGORY_UI, description = "Maximum content length for URL imports", availableSince = "3.2.0")
public Long featureUrlImportMaxContentLength;
```
Then in `SystemResourceImpl.java` use `uiConfig.featureUrlImportMaxContentLength` directly (no `.get()`).

ui/ui-app/src/services/useConfigService.ts
```diff
@@ -364,6 +366,11 @@ export class ConfigServiceImpl implements ConfigService {
         return this.features().searchIndex || false;
     }
 
+    public featureUrlImportMaxContentLength(): number {
+        const configured = this.features().urlImportMaxContentLength;
+        return typeof configured === 'number' && configured > 0 ? configured : 5242880;
```
EricWittmann, line 371: ESLint failure — this project enforces double quotes:
```
371:38  error  Strings must use doublequote  quotes
```
Change `'number'` to `"number"`:
```ts
return typeof configured === "number" && configured > 0 ? configured : 5242880;
```

common/src/main/resources/META-INF/openapi.json
```diff
@@ -8271,6 +8271,10 @@
           "searchIndex": {
             "description": "",
             "type": "boolean"
+          },
+          "urlImportMaxContentLength": {
+            "description": "Maximum content length in bytes for URL imports.",
+            "type": "integer"
```
EricWittmann, line 8322: If you keep `Long` on the backend, add `"format": "int64"` here so the generated Java bean uses `Long` instead of `Integer`:
```json
"urlImportMaxContentLength": {
    "description": "Maximum content length in bytes for URL imports.",
    "type": "integer",
    "format": "int64"
}
```
If you switch the backend to `Integer` (recommended — a URL import cap will never exceed 2 GB), the current schema is correct as-is.

EricWittmann (changes requested):
Thanks for the contribution! The feature design is solid and the frontend test coverage is thorough. However, there are two build-blocking issues and a design inconsistency that need to be resolved before this can merge.

### Build blockers
1. **Java compilation fails** — `Long` vs `Integer` type mismatch between the config property and the generated bean (see inline comment on `SystemResourceImpl.java`).
2. **UI build fails** — ESLint single-quote violation (see inline comment on `useConfigService.ts`).

### Other issues
3. **`Supplier<Long>` without `@Dynamic`** — inconsistent with the codebase pattern (see inline comment on `UserInterfaceConfigProperties.java`).
4. **Go SDK not regenerated** — the `openapi.json` was modified but the Go SDK wasn't regenerated, which fails the Go SDK Freshness CI check. After fixing the OpenAPI schema, regenerate the Go SDK.
5. **`availableSince` version** — verify `3.2.0` is still the correct target. The PR has been open for a while, so this may need bumping.

EricWittmann:
Closing as stale and replaced by:
* https://github.com/Apicurio/apicurio-registry/pull/10013

PR #9403, opened 2026-08-09
fix(backend): Explicitly throw RegistryException on content canonical…

harshitnub077:
### Description
In `RegistryStorageContentUtils.java` and `RegistryContentUtils.java`, when `canonicalizeContent` encounters an exception during schema parsing/canonicalization (e.g., due to malformed AST or unresolvable nested references), the system catches the exception, emits a `debug` log, and silently falls back to returning the raw, un-canonicalized content.

This is highly problematic for data integrity in the storage layer:
1. **Deduplication Failure**: Content IDs and hashes are computed on this output. By silently falling back to raw content, structurally identical schemas with differing whitespace/metadata will yield different SHA-256 hashes, bypassing the deduplication mechanism and causing storage bloat.
2. **Silent State Corruption**: The REST API effectively swallows fundamental structural errors in the uploaded artifact, masking the failure from the client.

### Proposed Solution
This PR hardens the canonicalization path by removing the silent fallback. We now throw a `RegistryException` when content fails to canonicalize, ensuring the storage layer maintains strict content hashing and deduplication integrity.


carlesarnal:
Closing as this has been in draft for a few weeks and has no linked issue.

PR #9404, opened 2026-08-09
fixes missing contract-related events when using KafkaSQL storage

vivek-gite:
## Root cause

Contract metadata and status events were temporarily emitted from `GroupsResourceImpl` through `RegistryStorage.createEvent()`. That bypassed the normal storage-layer event flow. It also did not address six KafkaSQL contract-ruleset operations, which journaled their changes but never published corresponding events.

## What does it fix?
issue #9034 

## Changes

- Adds `RegistryStorage` operations for contract metadata updates and status transitions.
- Moves contract label updates and event creation from the REST resource into SQL storage.
- Adds replayable KafkaSQL journal messages for the new storage operations.
- Publishes `KafkaSqlOutboxEvent`s only after KafkaSQL journal acknowledgement for:
  - contract metadata updates;
  - contract status transitions;
  - artifact-level ruleset set/delete;
  - version-level ruleset set/delete;
  - global ruleset set/delete.
- Restores KafkaSQL `createEvent()` to its intended no-op behavior, removing the REST-layer workaround.
- Updates read-only storage implementations for the new write methods.
- Adds regression coverage for read-only storage and KafkaSQL contract event payloads.

## Testing

- `./mvnw test-compile -pl app -am -DskipTests`
- `./mvnw checkstyle:check -pl app`
- `./mvnw test -pl app -am -Dtest=ReadOnlyRegistryStorageTest -Dsurefire.failIfNoSpecifiedTests=false`


app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/serde/KafkaSqlMessageIndex.java
```diff
@@ -64,7 +64,9 @@ private static void indexMessageClasses(Class<? extends KafkaSqlMessage>... mcla
                 DeleteGlobalContractRuleset0Message.class,
                 InsertContractAuditEntry1Message.class,
                 MergeArtifactLabels4Message.class,
-                MergeVersionLabels5Message.class);
+                MergeVersionLabels5Message.class,
+                UpdateContractMetadata4Message.class,
```
paoloantinori, line 68: UpdateContractMetadata4Message reuses the '4' numeric suffix already taken by MergeArtifactLabels4Message in this same index. If the KafkaSQL serde uses that number as the wire-format type discriminator (the 0/1/4/5 gaps suggest it is allocated per message type), the two collide and one deserializes as the other on replay. Confirm the discriminator is class-based, or allocate a unique id. This round-trip is covered only by KafkaSqlContractEventsTest, which is @Tag(SLOW) and was not run in this PR.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -680,6 +686,27 @@ public void deleteGroupRules(String groupId) throws RegistryStorageException {
         coordinator.waitForResponse(uuid);
     }
 
+    @Override
+    public void updateContractMetadata(String groupId, String artifactId, String contractId,
+            EditableContractMetadataDto metadata) throws RegistryStorageException {
+        var message = new UpdateContractMetadata4Message(groupId, artifactId, contractId, metadata);
+        var uuid = blockOnResult(submitter.submitMessage(message));
+        coordinator.waitForResponse(uuid);
+        outboxEvent.fire(KafkaSqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
```
paoloantinori, line 695: Two things to confirm on this KafkaSQL emission path. First, when the journal message is applied, AbstractSqlRegistryStorage.updateContractMetadata also fires a SqlOutboxEvent (around line 851), so verify that observer is inactive under the KafkaSQL profile, otherwise each metadata change emits two events to registry-events. Second, this outboxEvent.fire runs unconditionally after coordinator.waitForResponse, so confirm waitForResponse rethrows when journal processing fails, else CONTRACT_METADATA_UPDATED fires for a write that never persisted.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlRegistryStorage.java
```diff
@@ -1307,7 +1334,7 @@ public String createSnapshot(String snapshotLocation) throws RegistryStorageExce
 
     @Override
     public String createEvent(OutboxEvent event) {
```
paoloantinori, line 1336: Reverting createEvent to a no-op silently drops the event for any caller still routing through it. The contract REST workaround is removed, but grep every storage.createEvent and RegistryStorage.createEvent caller to confirm nothing else depends on it firing. A missed caller loses its events with no error.

paoloantinori (commented):
This PR correctly moves contract metadata and status event emission into the storage layer and restores KafkaSQL createEvent to a no-op, which addresses the missing-events gap in #9034. The main weakness is verification: the new KafkaSQL journal messages and outbox emission are covered end-to-end only by KafkaSqlContractEventsTest, which is @Tag(SLOW) and was not run in the PR's test plan, so the serde and single-emission guarantees are unconfirmed.

- KafkaSqlMessageIndex.java line 68: UpdateContractMetadata4Message reuses the '4' suffix already taken by MergeArtifactLabels4Message; if that number is the wire type discriminator, they collide on replay.
- KafkaSqlRegistryStorage.java line 695: the KafkaSqlOutboxEvent fired here overlaps the SqlOutboxEvent fired by the SQL impl during journal processing, and fires unconditionally after waitForResponse.
- KafkaSqlRegistryStorage.java line 1336: createEvent reverted to a no-op silently drops events for any remaining caller; confirm none are orphaned.
- AbstractSqlRegistryStorage.java line 851: the rollback guarantee is tested only against a simulated HandleFactory, not the real transaction.
- KafkaSqlContractEventsTest.java line 35: this is the only end-to-end coverage and it is tagged SLOW; run it before merge.

app/src/main/java/io/apicurio/registry/storage/impl/sql/AbstractSqlRegistryStorage.java
```diff
@@ -833,6 +840,45 @@ public void deleteGroupRule(String groupId, RuleType rule) throws RegistryStorag
...
 
+    @Override
+    public void updateContractMetadata(String groupId, String artifactId, String contractId,
+            EditableContractMetadataDto metadata) throws RegistryStorageException {
+        handles.withHandle(handle -> {
+            String prefix = contractId != null
+                    ? ContractLabels.contractPrefix(contractId) : ContractLabels.PREFIX;
+            mergeArtifactLabels(groupId, artifactId, prefix,
+                    contractMetadataMapper.toLabels(metadata, prefix));
+            outboxEvent.fire(SqlOutboxEvent.of(ContractMetadataUpdated.of(groupId, artifactId)));
```
paoloantinori, line 851: The rollback guarantee (labels rolled back when the outbox insert fails) depends on mergeArtifactLabels opening a nested handle that shares the outer transaction, and on the SqlOutboxEvent observer running synchronously in that same transaction. AbstractSqlRegistryStorageContractTransactionTest validates this only against a simulated HandleFactory, so it proves the method shape, not the real JDBI nesting or the CDI observer timing. Confirm the existing pattern guarantees this, or add an integration-level assertion.

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlContractEventsTest.java
```diff
@@ -1,92 +1,225 @@
...
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 
+import static io.apicurio.registry.storage.impl.kafkasql.KafkaSqlRegistryStorage.GLOBAL_CONTRACT_RULESET_COORDINATE;
 import static io.restassured.RestAssured.given;
 
 @QuarkusTest
 @TestProfile(KafkasqlTestProfile.class)
+@Tag(ApicurioTestTags.SLOW)
```
paoloantinori, line 35: This tag is why the only end-to-end coverage of the new journal messages (notably UpdateContractMetadata4Message carrying an EditableContractMetadataDto through the KafkaSQL serde) and the outbox emission was skipped: the PR's test plan ran only ReadOnlyRegistryStorageTest. The serde round-trip, DTO serializability, and single-emission behavior are unverified as a result. Run this test with SLOW enabled before merge.

PR #9407, opened 2026-08-09
test(mcp): add comprehensive edge-case tests for MCP_TOOL validation and compatibility

### LINKED ISSUE #9406: [FEAT] Add Validation and Compatibility Rules for MCP_TOOL Artifacts
## Feature Description
With the growing adoption of Model Context Protocol (MCP) servers and tools in AI agent workflows, Apicurio Registry should support `MCP_TOOL` artifact types with built-in validation and compatibility checking rules.

## Use Case / Problem Statement
Organizations hosting multiple MCP servers need a central schema registry to:
1. Validate `MCP_TOOL` schema definitions (tool names, JSON schema parameters, return signatures).
2. Enforce `BACKWARD` and `FORWARD` compatibility rules when MCP tools are updated, preventing breaking changes in downstream AI agent integrations.

## Proposed Solution
- Register `MCP_TOOL` as a recognized artifact type.
- Implement `MCPToolSchemaValidator` under the `schema-validation` module.
- Add compatibility rule logic in `contracts-rules` / `mcp` module.
- Add unit and integration tests covering positive and negative compatibility cases.

## Additional Context
This feature request is aligned with the **LFX Mentorship 2026 Term 3** project proposals for Apicurio Registry.

sudhanshurai2004:
## Summary

This PR adds comprehensive edge-case unit tests for `MCP_TOOL` artifact type
validation and compatibility checking, as part of Issue #9406.

### New Test Fixtures
- `mcptool-empty-name.json` - whitespace-only name
- `mcptool-invalid-priority.json` - priority > 1.0 (out of range)
- `mcptool-invalid-outputschema.json` - non-object outputSchema
- `mcptool-invalid-required-array.json` - non-string entries in required array

### McpToolContentValidatorTest (6 new tests)
- Empty/whitespace name rejection
- Priority out-of-bounds (> 1.0) rejection
- Invalid outputSchema type rejection
- Non-string required array entries rejection
- References rejection via validateReferences()
- SYNTAX_ONLY level passes structurally invalid but JSON-valid content

### McpToolCompatibilityCheckerTest (3 new tests)
- Removing optional property is backward incompatible
- Identical schemas are always compatible
- Malformed proposed JSON is always incompatible

Closes #9406

### DIFF
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
@@ -344,4 +344,92 @@ class McpToolCompatibilityCheckerTest {
 
         assertTrue(result.isCompatible(), "Identical schema with description change should be fully compatible");
     }
+
+    @Test
+    void testForwardCompatibleRemovingOptionalProperty() {
+        String existing = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "query": { "type": "string" },
+                            "limit": { "type": "integer" }
+                        },
+                        "required": ["query"]
+                    }
+                }
+                """;
+
+        // Proposed removes optional "limit" - forward incompatible (producers using new version
+        // won't send "limit" but old consumers might need it)
+        String proposed = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "query": { "type": "string" }
+                        },
+                        "required": ["query"]
+                    }
+                }
+                """;
+
+        CompatibilityExecutionResult backwardResult = checker.testCompatibility(
+                CompatibilityLevel.BACKWARD, List.of(createMcpTool(existing)),
+                createMcpTool(proposed), Map.of());
+
+        assertFalse(backwardResult.isCompatible(),
+                "Removing any property should be backward incompatible");
+    }
+
+    @Test
+    void testBackwardCompatibleNoChanges() {
+        String existing = """
+                {
+                    "name": "test_tool",
+                    "description": "A tool",
+                    "inputSchema": {
+                        "type": "object",
+                        "properties": {
+                            "query": { "type": "string" }
+                        },
+                        "required": ["query"]
+                    }
+                }
+                """;
+
+        CompatibilityExecutionResult result = checker.testCompatibility(
+                CompatibilityLevel.BACKWARD, List.of(createMcpTool(existing)),
+                createMcpTool(existing), Map.of());
+
+        assertTrue(result.isCompatible(),
+                "Identical schemas should be backward compatible");
+    }
+
+    @Test
+    void testIncompatibleWithParseError() {
+        String existing = """
+                {
+                    "name": "test_tool",
+                    "inputSchema": { "type": "object" }
+                }
+                """;
+
+        String malformed = "{ this is not valid json }";
+
+        CompatibilityExecutionResult result = checker.testCompatibility(
+                CompatibilityLevel.BACKWARD,
+                List.of(TypedContent.create(
+                        io.apicurio.registry.content.ContentHandle.create(existing),
+                        ContentTypes.APPLICATION_JSON)),
+                TypedContent.create(
+                        io.apicurio.registry.content.ContentHandle.create(malformed),
+                        ContentTypes.APPLICATION_JSON),
+                Map.of());
+
+        assertFalse(result.isCompatible(),
+                "Malformed proposed content should be incompatible");
+    }
 }
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
@@ -8,7 +8,9 @@ import io.apicurio.registry.rules.violation.RuleViolationException;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import io.apicurio.registry.rest.v3.beans.ArtifactReference;
 import java.util.Collections;
+import java.util.List;
 
 /**
  * Tests the MCP tool content validator, accepter, and extractor.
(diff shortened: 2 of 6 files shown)


schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
```diff
@@ -344,4 +344,92 @@ void testFullCompatibilityBothDirections() {
 
         assertTrue(result.isCompatible(), "Identical schema with description change should be fully compatible");
     }
+
+    @Test
```
paoloantinori, line 348: Removing an optional, non-required property (here `limit`) is backward COMPATIBLE under standard JSON Schema evolution: readers on the new schema still accept old data that contains the extra field, since `additionalProperties` defaults to true. This test asserts `assertFalse(backwardResult.isCompatible())`, the opposite. If `McpToolCompatibilityChecker` does not intentionally treat any property removal as incompatible, this expectation is inverted and the test will fail at runtime. Also, the method name `testForwardCompatibleRemovingOptionalProperty` reads as the opposite of what the test actually checks (BACKWARD level, expecting incompatible); once the behavior is confirmed, a name like `testRemovingOptionalPropertyIsBackwardIncompatible` would match the assertion.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/validity/McpToolContentValidatorTest.java
```diff
@@ -132,4 +134,59 @@ public void testMcpToolValidationNoneLevel() throws Exception {
...
+    @Test
+    public void testMcpToolEmptyName() throws Exception {
+        TypedContent content = resourceToTypedContentHandle("mcptool-empty-name.json");
+        McpToolContentValidator validator = new McpToolContentValidator();
+        Assertions.assertThrows(RuleViolationException.class, () -> {
+            validator.validate(ValidityLevel.FULL, content, Collections.emptyMap());
+        });
+    }
+
+    @Test
```
paoloantinori, line 147: `priority` is not part of the MCP tool `annotations` spec, which defines hint booleans (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`) plus `title`, and a 0.0-1.0 range is non-standard. Confirm `McpToolContentValidator` actually enforces an `annotations.priority` range; if it does not, `assertThrows` here receives no exception and `testMcpToolInvalidPriority` fails. If the rule is intentional and Apicurio-specific, note it in the fixture or test name so it is not mistaken for spec-conformance validation.

schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/McpToolCompatibilityCheckerTest.java
```diff
@@ -344,4 +344,92 @@ void testFullCompatibilityBothDirections() {
...
+        CompatibilityExecutionResult result = checker.testCompatibility(
+                CompatibilityLevel.BACKWARD,
+                List.of(TypedContent.create(
+                        io.apicurio.registry.content.ContentHandle.create(existing),
+                        ContentTypes.APPLICATION_JSON)),
+                TypedContent.create(
+                        io.apicurio.registry.content.ContentHandle.create(malformed),
+                        ContentTypes.APPLICATION_JSON),
+                Map.of());
+
```
paoloantinori, line 431: For unparseable proposed content, confirm `McpToolCompatibilityChecker.testCompatibility` returns a `CompatibilityExecutionResult` with `isCompatible() == false` rather than propagating the JSON parse exception. If the parse error is thrown, this `assertFalse` never executes and the test errors instead of asserting. If the checker does throw, wrap the call in `assertDoesNotThrow(...)` before the `assertFalse`, or assert the thrown exception type explicitly so the failure mode is unambiguous. As a side note, the manual `TypedContent.create(io.apicurio.registry.content.ContentHandle.create(...), ...)` construction duplicates the existing `createMcpTool(...)` helper used elsewhere in this class; reuse it (or a static import) for consistency.

paoloantinori (commented):
The PR adds useful negative-path coverage for MCP_TOOL validation and compatibility, but several assertions encode behavior that should be confirmed against the current `McpToolContentValidator` and `McpToolCompatibilityChecker`, and one expectation conflicts with standard JSON Schema backward-compatibility semantics.

- `McpToolCompatibilityCheckerTest.testForwardCompatibleRemovingOptionalProperty` asserts that removing an optional property is backward-incompatible, which contradicts standard schema-evolution semantics (removing an optional field is normally backward compatible because old data still validates against the new schema).
- `McpToolContentValidatorTest.testMcpToolInvalidPriority` validates an `annotations.priority` range that is not part of the MCP tool spec; confirm the validator actually enforces it or `assertThrows` will receive no exception and the test will fail.
- `testIncompatibleWithParseError` assumes the checker returns a result for malformed JSON rather than throwing the parse error; confirm.
- Two fixtures referenced by the new tests (`mcptool-valid.json`, `mcptool-missing-inputschema.json`) are not in this diff; confirm they already exist under `src/test/resources/io/apicurio/registry/rules/validity/` or `testMcpToolReferencesRejected` and `testMcpToolSyntaxOnlyPassesStructurallyInvalid` will fail to load resources.

aclfe (commented):
hello, thanks for the PR. It seems the issue is a smaller duplicate of #8427 and what #9406 proposes is already there? ArtifactType.java, McpToolContentValidator.java, McpToolCompatibilityChecker.java, all exist already. The issue said 

```
Register MCP_TOOL as a recognized artifact type.
Implement MCPToolSchemaValidator under the schema-validation module.
Add compatibility rule logic in contracts-rules / mcp module.
Add unit and integration tests covering positive and negative compatibility cases.
```

but the code says something different? Are you adding only edge-case tests for the issue?

Nachiket-Roy:
> hello, thanks for the PR. It seems the issue is a smaller duplicate of #8427 and what #9406 proposes is already there? ArtifactType.java, McpToolContentValidator.java, McpToolCompatibilityChecker.java, all exist already. The issue said
> 
> ```
> Register MCP_TOOL as a recognized artifact type.
> Implement MCPToolSchemaValidator under the schema-validation module.
> Add compatibility rule logic in contracts-rules / mcp module.
> Add unit and integration tests covering positive and negative compatibility cases.
> ```
> 
> but the code says something different? Are you adding only edge-case tests for the issue?

True! Moreover you can't work on issues that are reserved for LFX !

jsenko (commented):
Adds useful edge-case test coverage for MCP_TOOL validation and compatibility. Verified the test expectations against the current `McpToolContentValidator` and `McpToolCompatibilityChecker` — all assertions match the implemented behavior.

### should-fix: Fully qualified `ContentHandle` in compatibility test

`McpToolCompatibilityCheckerTest.java` line ~430 uses `io.apicurio.registry.content.ContentHandle.create(...)` inline instead of importing it. Add `import io.apicurio.registry.content.ContentHandle;` and use `ContentHandle.create(...)`.

### should-fix: Misleading test name

`testForwardCompatibleRemovingOptionalProperty` tests `CompatibilityLevel.BACKWARD` and asserts `assertFalse(backwardResult.isCompatible())` — it's verifying backward *in*compatibility, not forward compatibility. Rename to `testBackwardIncompatibleRemovingOptionalProperty` to match what it actually tests.

- **"Removing optional property is backward incompatible"** — Confirmed correct per `McpToolCompatibilityChecker.checkPropertyRemovals()`, which reports `PROPERTY_REMOVED` for any removed property regardless of required status. This is a stricter-than-JSON-Schema rule that makes sense for tool input contracts (callers using the old schema may rely on the property existing). paoloantinori's concern about this conflicting with standard JSON Schema semantics is worth noting but the test matches the current implementation's documented intent.
- **`testMcpToolSyntaxOnlyPassesStructurallyInvalid`** — correctly tests that SYNTAX_ONLY passes a JSON object missing required MCP fields (validation stops after the `isObject()` check), using the existing `mcptool-missing-inputschema.json` fixture.

carlesarnal:

PR #9412, opened 2026-08-09
test(cli): prevent vacuous type-filter search assertion

techsavyr00hit:
The current test uses `allMatch()`, which can pass even when the search returns no artifacts. This means the test may not catch a problem with the type filter.
I added `isNotEmpty()` before `allMatch()` so the test now checks that:
1. The search returns at least one artifact.
2. All returned artifacts have the expected `JSON` type.

This also follows the same pattern used by the other search tests in the file.

### DIFF
--- a/cli/src/test/java/io/apicurio/registry/cli/SearchCommandTest.java
+++ b/cli/src/test/java/io/apicurio/registry/cli/SearchCommandTest.java
@@ -194,6 +194,7 @@ public class SearchCommandTest extends AbstractCLITest {
 
         assertThat(results.getArtifacts())
                 .as(withCliOutput("Type-filtered search should return JSON artifacts"))
+                .isNotEmpty()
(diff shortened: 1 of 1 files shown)


vandanayadav7 (approved):
Correct fix — `allMatch()` on an empty list is vacuously true, so the test could pass without actually finding any artifacts. Adding `isNotEmpty()` matches the pattern used by every other `allMatch` assertion in this file. LGTM.

Minor note: please link a GitHub issue in future PRs, even for small fixes — it helps with tracking.

vandanayadav7:
@techsavyr00hit Could you mark this PR as ready for review (out of draft)? I can't do it from my side.

techsavyr00hit:
@vandanayadav7 hey, the PR is ready for review. Thanks for your suggestions, I will try to apply those tips in future 🙂

PR #9415, opened 2026-08-09
fix(rest): fail Protobuf export when a reference cannot be resolved

### LINKED ISSUE #9333: bug: Silent reference failure produces incomplete Protobuf export

Nikhil-VS1811:
## Summary

Fixes #9333.

`ProtobufExporter.collectReferences()` previously swallowed exceptions when a referenced Protobuf artifact could not be resolved. This caused the export to continue and return an incomplete ZIP file with HTTP 200.

This change propagates reference-resolution failures so that the export does not silently produce an incomplete archive.

## Changes

- Propagate failures when a Protobuf reference cannot be resolved.
- Prevent incomplete Protobuf exports from being returned successfully.
- Preserve useful reference information in error handling.

## Validation

- `git diff --check` passes.
- Tests completed successfully in the local environment.
- Working tree is clean.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
@@ -5,6 +5,7 @@ import io.apicurio.registry.storage.RegistryStorage;
 import io.apicurio.registry.storage.dto.ArtifactReferenceDto;
 import io.apicurio.registry.storage.dto.ContentWrapperDto;
 import io.apicurio.registry.storage.dto.StoredArtifactVersionDto;
+import io.apicurio.registry.types.RegistryException;
 import io.apicurio.registry.utils.protobuf.schema.ProtoContent;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
@@ -48,11 +49,13 @@ public class ProtobufExporter {
      * @return Response containing the ZIP file
      */
     public Response exportVersionAsZip(String groupId, String artifactId, String version) {
+
+        // Resolve all artifacts before creating the response.
+        Map<String, ProtoContent> protoFiles =
+                collectAllProtoFiles(groupId, artifactId, version);
+
         StreamingOutput stream = os -> {
             try (ZipOutputStream zip = new ZipOutputStream(os, StandardCharsets.UTF_8)) {
-                // Collect all proto files with their content
-                Map<String, ProtoContent> protoFiles = collectAllProtoFiles(groupId, artifactId, version);
-
                 // Build a mapping from original import paths to canonical paths
                 Map<String, String> importPathMapping = buildImportPathMapping(protoFiles);
 
@@ -101,12 +104,17 @@ public class ProtobufExporter {
      *
      * @return Map from original import path to ProtoContent
      */
-    private Map<String, ProtoContent> collectAllProtoFiles(String groupId, String artifactId, String version) {
+    private Map<String, ProtoContent> collectAllProtoFiles(
+            String groupId,
+            String artifactId,
+            String version) {
+
         Map<String, ProtoContent> result = new LinkedHashMap<>();
         Set<String> visited = new HashSet<>();
 
         // Get the main artifact
-        StoredArtifactVersionDto mainArtifact = storage.getArtifactVersionContent(groupId, artifactId, version);
+        StoredArtifactVersionDto mainArtifact =
+                storage.getArtifactVersionContent(groupId, artifactId, version);
         String mainContent = mainArtifact.getContent().content();
 
         // Use artifact ID as the filename for the main artifact
@@ -127,15 +135,20 @@ public class ProtobufExporter {
     /**
      * Recursively collects all references.
      */
-    private void collectReferences(List<ArtifactReferenceDto> references,
-                                   Map<String, ProtoContent> result,
-                                   Set<String> visited) {
+    private void collectReferences(
+            List<ArtifactReferenceDto> references,
+            Map<String, ProtoContent> result,
+            Set<String> visited) {
+
         if (references == null || references.isEmpty()) {
             return;
         }
 
         for (ArtifactReferenceDto ref : references) {
-            String key = buildVisitedKey(ref.getGroupId(), ref.getArtifactId(), ref.getVersion());
+            String key = buildVisitedKey(
+                    ref.getGroupId(),
+                    ref.getArtifactId(),
+                    ref.getVersion());
 
             if (visited.contains(key)) {
                 continue;
@@ -144,18 +157,36 @@ public class ProtobufExporter {
 
             try {
                 ContentWrapperDto content = storage.getContentByReference(ref);
-                if (content != null && content.getContent() != null) {
-                    String refContent = content.getContent().content();
-                    // Use the reference name as the import path
-                    String importPath = ref.getName();
-                    ProtoContent protoContent = new ProtoContent(importPath, refContent);
-                    result.put(importPath, protoContent);
 
-                    // Recursively collect nested references
-                    collectReferences(content.getReferences(), result, visited);
+                if (content == null || content.getContent() == null) {
+                    throw new RegistryException(
+                            "Could not resolve Protobuf reference: " + ref.getName());
                 }
+
+                String refContent = content.getContent().content();
+
+                // Use the reference name as the import path
+                String importPath = ref.getName();
+                ProtoContent protoContent = new ProtoContent(importPath, refContent);
+                result.put(importPath, protoContent);
+
+                // Recursively collect nested references
+                collectReferences(content.getReferences(), result, visited);
+
+            } catch (RegistryException e) {
+                throw e;
             } catch (Exception e) {
-                log.warn("Could not resolve reference: {} - {}", ref.getName(), e.getMessage());
+                log.error(
+                        "Could not resolve Protobuf reference: {}:{}:{} name={}",
+                        ref.getGroupId(),
+                        ref.getArtifactId(),
+                        ref.getVersion(),
+                        ref.getName(),
+                        e);
+
+                throw new RegistryException(
+                        "Could not resolve Protobuf reference: " + ref.getName(),
+                        e);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Nachiket-Roy:
#9333 already has a linked PR. Please check the existing work before starting to avoid duplicate effort.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
```diff
@@ -144,26 +157,46 @@ private void collectReferences(List<ArtifactReferenceDto> references,
...
-                    // Use the reference name as the import path
-                    String importPath = ref.getName();
-                    ProtoContent protoContent = new ProtoContent(importPath, refContent);
-                    result.put(importPath, protoContent);
-
-                    // Recursively collect nested references
-                    collectReferences(content.getReferences(), result, visited);
+
+                if (content == null || content.getContent() == null) {
+                    throw new RegistryException(
```
paoloantinori, line 162: Throwing bare `io.apicurio.registry.types.RegistryException` here resolves to HTTP 500: it has no entry in `HttpStatusCodeMap`, so `getCode` walks the superclass chain to `Object` and returns 500. In `CoreRegistryExceptionMapperService`, a 500 triggers `liveness.suspectWithException(...)` (it is not in the ignored set, since `isIgnored` is `codeMap.containsKey(clazz)`), so a caller repeatedly exporting an artifact with a dangling reference can trip the liveness probe and restart pods. An unresolvable reference is a not-found condition; throw `VersionNotFoundException` instead, which extends `NotFoundException`, maps to 404, and is excluded from the liveness path. This also matches the `// TODO Should be abstract and more specific exception should be used` note on `RegistryException` itself.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
```diff
@@ -144,26 +157,46 @@ private void collectReferences(List<ArtifactReferenceDto> references,
...
+                String importPath = ref.getName();
+                ProtoContent protoContent = new ProtoContent(importPath, refContent);
+                result.put(importPath, protoContent);
+
+                // Recursively collect nested references
+                collectReferences(content.getReferences(), result, visited);
+
+            } catch (RegistryException e) {
+                throw e;
             } catch (Exception e) {
```
paoloantinori, line 178: The broad `catch (Exception e)` wraps every non-RegistryException into a generic 500. If a storage implementation threw a `NotFoundException` or `VersionNotFoundException` from `getContentByReference`, it would be silently converted to 500 here. Consider rethrowing `RegistryStorageException` (and its not-found subtypes) as-is, mirroring the `catch (RegistryException e) { throw e; }` above, so storage-level 404s stay 404.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
```diff
@@ -144,26 +157,46 @@ private void collectReferences(List<ArtifactReferenceDto> references,
...
+                log.error(
+                        "Could not resolve Protobuf reference: {}:{}:{} name={}",
+                        ref.getGroupId(),
+                        ref.getArtifactId(),
+                        ref.getVersion(),
+                        ref.getName(),
+                        e);
+
+                throw new RegistryException(
+                        "Could not resolve Protobuf reference: " + ref.getName(),
```
paoloantinori, line 188: The thrown message carries only `ref.getName()`. This string becomes the `ProblemDetails.title` returned to the API caller, while the log carries the full groupId:artifactId:version. Including the coordinates in the message lets the caller identify which reference failed without needing server logs.

app/src/test/java/io/apicurio/registry/noprofile/rest/v3/ProtobufExportTest.java
```diff
@@ -386,4 +388,47 @@ public void testExportArtifactWithoutPackage() throws Exception {
...
+
+        given()
+        .config(RestAssuredConfig.config()
+                .decoderConfig(DecoderConfig.decoderConfig().noContentDecoders()))
+        .header("Accept-Encoding", "identity")
+        .when()
+        .pathParam("groupId", GROUP)
+        .pathParam("artifactId", artifactId)
+        .pathParam("versionExpression", "branch=latest")
+        .get("/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{versionExpression}/export")
```
paoloantinori, line 430: This assertion pins `statusCode(500)`. If the production change switches to `VersionNotFoundException`, the correct status is 404, so update this. It would also be stronger to assert the response is a JSON error body (for example `contentType(JSON)`) rather than only the status, which locks in that no partial archive is ever returned.

app/src/main/java/io/apicurio/registry/rest/v3/impl/shared/ProtobufExporter.java
```diff
@@ -188,4 +221,4 @@ private String extractFilename(String artifactId) {
         // Otherwise, append .proto
         return artifactId + ".proto";
     }
-}
+}
```
paoloantinori, line 224: The file lost its final newline (`\ No newline at end of file`). Most checkstyle and CI setups expect a trailing newline; restore it to avoid a spurious full-file diff later.

paoloantinori (commented):
The fix correctly stops the silent-incomplete-export bug by resolving references before the response is committed, and the new failure-path test is the right shape. The main concern is the exception type: a bare `RegistryException` resolves to HTTP 500 and feeds the liveness probe for what is really a not-found, client-data condition.

- `ProtobufExporter.java:162`: Bare `io.apicurio.registry.types.RegistryException` has no entry in `HttpStatusCodeMap`, so `getCode` returns 500; `CoreRegistryExceptionMapperService` then calls `liveness.suspectWithException(...)` because it is not in the ignored set. A caller repeatedly exporting an artifact with a dangling reference can trip the liveness probe and restart pods. Throw `VersionNotFoundException` (extends `NotFoundException`, maps to 404 and is excluded from the liveness path); this also matches the `// TODO Should be abstract` note on `RegistryException` itself.
- `ProtobufExporter.java:178`: The broad `catch (Exception e)` re-wraps every non-RegistryException into a generic 500; if any storage path threw a `NotFoundException`, it would be silently converted to 500. Rethrow `RegistryStorageException` subtypes as-is, mirroring the `catch (RegistryException e) { throw e; }` above.
- `ProtobufExporter.java:188`: The thrown message carries only `ref.getName()`, but that string becomes the `ProblemDetails.title` returned to the API caller while the log has groupId:artifactId:version. Include the coordinates so the caller can identify the broken reference.
- `ProtobufExportTest.java:430`: The test pins `statusCode(500)`. Once a `NotFoundException` is used this should assert 404, and it would be stronger to assert the body is a JSON error (not a partial archive).
- `ProtobufExporter.java:224`: The file lost its trailing newline (`\ No newline at end of file`); restore it.

Nikhil-VS1811:
I noticed that issue #9333 already has PR #9365 addressing the same problem. Since that PR is further along, should I close #9415 to avoid duplicating the work, or would you prefer me to continue with this implementation?

EricWittmann:
Looks like we had a duplicate PR!  I'll close this - please open a new one if there is an aspect to this that isn't already covered!

Issue #7641, opened 2026-03-25
CLI: Verbose mode does not affect Quarkus/JBoss logging


jsenko:
## Description

The `--verbose` flag currently only sets the JUL (java.util.logging) root logger level to `FINE`. However, the CLI's own code uses JBoss Logging (via Quarkus), and `application.properties` sets `quarkus.log.level=WARN`. This means `log.debugf()` calls in CLI commands (e.g., InstallCommand, UpdateCommand) are not visible even with `--verbose`.

## Expected Behavior

`--verbose` should enable debug-level output from both the SDK (JUL) and the CLI's own code (JBoss Logging).

## Possible Solutions

- Set `System.setProperty("quarkus.log.level", "DEBUG")` at runtime (may not work after Quarkus init)
- Use Quarkus's `LoggingSetupRecorder` or JBoss LogManager API to reconfigure at runtime
- Switch CLI code to use JUL directly (consistent with SDK)

## Context

Follow-up from #7618 (Quarkus-based CLI with native build). Identified during PR review.

PR #9420, opened 2026-08-09
feat(cli): print raw request and response details with --verbose

### LINKED ISSUE #8027: CLI: Verbose logging should print raw request and response details
## Description

When `--verbose` is enabled, the CLI should print raw Registry client request and response details for debugging.

## Expected Behavior

- Print request: method, URL, headers, body
- Print response: status code, headers, body
- Only when `--verbose` flag is used

## Parent Epic

- #7162

aclfe:
Fixes #8027 (item 8 of the parent epic #7162).

## What

`--verbose` now prints the raw HTTP exchange with the Registry: request method, URL, headers and body, and response status code, headers and body.

```
HTTP request:
> GET http://localhost:8080/apis/registry/v3/groups/my-group
> Authorization: <redacted>

HTTP response:
< 200 OK
< Content-Type: application/json
<
< {"groupId":"my-group", ...}
```

## Why

`--verbose` already raises log levels (#7641), but the Registry client itself logged nothing, so there was no way to see what the CLI actually sent or what came back. That is the first thing you want when debugging a failing call against a real server.

## How

The Kiota `RequestAdapter` only sees deserialized models, so it cannot show a raw response. The logging therefore sits at the HTTP layer, as a Vert.x web client interceptor:

- `RegistryClientOptions.enableHttpLogging()` — new opt-in flag, following the existing `enableOpenTelemetry()` pattern. Off by default, so nothing changes for existing SDK users.
- `HttpLoggingInterceptor` (adapter-vertx) — logs the request at `SEND_REQUEST` and the response at `DISPATCH_RESPONSE`. Installed after the auth web clients are built, so headers they contribute are logged too.
- The CLI turns the flag on when `--verbose` is set.

Records go to a dedicated `io.apicurio.registry.client.http` category at `FINE`, so HTTP traffic can be quieted on its own with the existing per-package config mechanism:

```bash
acr config set 'quarkus.log.category."io.apicurio.registry.client.http".level=WARNING'
```

The JDK adapter has no equivalent hook; the flag is a no-op there and the javadoc says so.

## Safety

- `Authorization`, `Proxy-Authorization`, `Cookie` and `Set-Cookie` values are replaced with `<redacted>`, so bearer tokens and basic credentials never reach the log.
- Bodies are truncated at 8192 characters, so artifact content does not flood the terminal.
- Non-buffered bodies (streamed uploads) are logged by type only, so they are not consumed away from the client.

## Tests

- `HttpLoggingInterceptorTest` (adapter-vertx, new test source set) — real HTTP round trip against a Vert.x server: asserts the logged method/URL/headers/body, the response status/headers/body, that credential headers are redacted, that a 10000 character body is truncated, and that nothing is logged when the interceptor is not installed.
- `VerboseHttpLoggingTest` (cli) — end-to-end through picocli against WireMock: `acr --verbose group get` logs the request and response, and the same command without `--verbose` logs nothing even with the category enabled.

Both are Docker-free. Full non-Docker CLI suite passes (102 tests), checkstyle clean on all touched modules.

## Notes for review

- `HttpLoggingInterceptor` uses `WebClientInternal.addInterceptor`, which lives in Vert.x's `impl` package. It is the only interceptor hook the web client offers, and it is what Vert.x's own OAuth2 and session-aware clients are built on. Guarded with an `instanceof` so a custom `WebClient` that is not a `WebClientBase` degrades to a warning instead of a `ClassCastException`.
- The `VertxAdapterFactory` diff looks larger than it is: the private `createVertx*` helpers now return `WebClient` instead of `RequestAdapter`, so the adapter can be constructed once after the interceptor is attached.
- I have not run a native CLI build locally; that path is covered in CI.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -192,6 +192,29 @@ Use `--verbose` to enable debug logging:
 acr --verbose artifact get my-artifact -g my-group
 ```
 
+Verbose output includes the raw request and response of every Registry call, with requests
+prefixed by `>` and responses by `<`:
+
+```
+HTTP request:
+> GET http://localhost:8080/apis/registry/v3/groups/my-group
+> Authorization: <redacted>
+
+HTTP response:
+< 200 OK
+< Content-Type: application/json
+<
+< {"groupId":"my-group", ...}
+```
+
+Headers that carry credentials (`Authorization`, `Proxy-Authorization`, `Cookie`, `Set-Cookie`)
+are redacted, and bodies longer than 8192 characters are truncated. HTTP traffic is logged under
+the `io.apicurio.registry.client.http` category, so it can be quieted on its own:
+
+```bash
+acr config set 'quarkus.log.category."io.apicurio.registry.client.http".level=WARNING'
+```
+
 Verbose output can be noisy. To quiet a specific package while keeping the rest, set a
 per-package level using the `quarkus.log.category."<package>".level` config key:
 
@@ -482,7 +505,7 @@ All filters are optional and can be combined. Additional filters include `--desc
 
 These options work with most commands:
 
-- `--verbose, -v` - Enable verbose output for debugging
+- `--verbose, -v` - Enable verbose output for debugging, including raw HTTP requests and responses
 - `--help, -h` - Show help information
 - `--output-type, -o` - Set output format (table or json)
 
--- a/cli/src/main/java/io/apicurio/registry/cli/Acr.java
+++ b/cli/src/main/java/io/apicurio/registry/cli/Acr.java
@@ -61,7 +61,7 @@ public class Acr {
 
     @Option(
             names = {"--verbose"},
-            description = "Enable verbose output.",
+            description = "Enable verbose output, including the raw request and response of every Registry call.",
             scope = INHERIT
     )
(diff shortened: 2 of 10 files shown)


vandanayadav7 (approved):
Verified with e2e testing — verbose logging works correctly. Request method, URL, headers, and body are logged, followed by the response status, headers, and body. Credential headers are properly redacted. Large bodies are truncated. Without `--verbose`, nothing is logged. All 7 tests pass.

Code is clean, modular, and follows existing patterns (`enableHttpLogging()` mirrors `enableOpenTelemetry()`). The interceptor sits at the right layer (SDK, not CLI) so other SDK users can opt in too. No security concerns.

SonarCloud findings are false positives — the `record` variable name and empty `flush()`/`close()` overrides are standard JUL `Handler` boilerplate in test code.

LGTM.

vandanayadav7:
@aclfe The PR looks good — reviewed and approved. Could you mark it as ready for review (out of draft)? I can't do it from my side.

aclfe:
Still waiting for (https://github.com/Apicurio/apicurio-registry/pull/8818) it seems

PR #9426, opened 2026-08-09
fix: version metadata modifiedBy/modifiedOn not updated on name-only or description-only PATCH

### LINKED ISSUE #9427: updateArtifactVersionMetaData does not update modifiedBy/modifiedOn when only name or description is changed
## Describe the bug

Updating an artifact version with only a name or description change succeeds, but modifiedBy and modifiedOn are not updated. The stored audit values remain those from version creation.

The problem occurs in the SQL storage implementation because the audit-field database update is nested under the labels check, even though name and description changes correctly mark the metadata as modified.

## Steps to reproduce

1. Create an artifact with an initial version.
2. Fetch the version metadata and record modifiedBy and modifiedOn.
3. Send a PUT request to the version metadata endpoint with only the name field changed.
4. Fetch the version metadata again.

## Expected behavior

The name is updated, modifiedBy identifies the user who performed the update, and modifiedOn advances.

## Actual behavior

The name is updated, but modifiedBy and modifiedOn retain their previous values.

DarkBird10020:
Fixes #9427.

## Summary

- Move the modifiedBy and modifiedOn update outside the labels-only branch in SqlVersionRepository.updateArtifactVersionMetaData().
- Preserve the existing unconditional ArtifactVersionMetadataUpdated outbox event behavior.
- Add regression coverage for name-only metadata updates and empty no-op updates.

## Why

Name-only and description-only metadata updates set the internal modified flag but previously skipped the database write for the audit fields because that write was nested under the labels check. The audit update now runs whenever any metadata field is modified, while empty updates leave the audit fields unchanged.

## Testing

- Name-only update verifies modifiedBy and modifiedOn are refreshed.
- Empty update verifies modifiedBy and modifiedOn remain unchanged.
- Full GitHub Actions suite passed on the previous head.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java
@@ -263,10 +263,10 @@ public class SqlVersionRepository {
                 if (rowCount == 0) {
                     throw new VersionNotFoundException(groupId, artifactId, version);
                 }
-
-                outboxEvent.fire(SqlOutboxEvent
-                        .of(ArtifactVersionMetadataUpdated.of(groupId, artifactId, version, editableMetadata)));
             }
+
+            outboxEvent.fire(SqlOutboxEvent
+                    .of(ArtifactVersionMetadataUpdated.of(groupId, artifactId, version, editableMetadata)));
 
             return null;
         });
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java
@@ -251,22 +251,22 @@ public class SqlVersionRepository {
                             .bind(1, limitStr(asLowerCase(k), MAX_LABEL_KEY_LENGTH))
                             .bind(2, limitStr(asLowerCase(v), MAX_LABEL_VALUE_LENGTH)).execute();
                 });
-
-                if (modified) {
-                    String modifiedBy = securityIdentity.getPrincipal().getName();
-                    Date modifiedOn = new Date();
-
-                    rowCount = handle.createUpdate(sqlStatements.updateArtifactVersionModifiedByOn())
-                            .bind(0, modifiedBy).bind(1, modifiedOn).bind(2, normalizeGroupId(groupId))
-                            .bind(3, artifactId).bind(4, version).execute();
-                    if (rowCount == 0) {
-                        throw new VersionNotFoundException(groupId, artifactId, version);
-                    }
-                }
             }
 
-            outboxEvent.fire(SqlOutboxEvent
-                    .of(ArtifactVersionMetadataUpdated.of(groupId, artifactId, version, editableMetadata)));
+            if (modified) {
+                String modifiedBy = securityIdentity.getPrincipal().getName();
+                Date modifiedOn = new Date();
+
+                int rowCount = handle.createUpdate(sqlStatements.updateArtifactVersionModifiedByOn())
+                        .bind(0, modifiedBy).bind(1, modifiedOn).bind(2, normalizeGroupId(groupId))
+                        .bind(3, artifactId).bind(4, version).execute();
+                if (rowCount == 0) {
+                    throw new VersionNotFoundException(groupId, artifactId, version);
+                }
+
+                outboxEvent.fire(SqlOutboxEvent
+                        .of(ArtifactVersionMetadataUpdated.of(groupId, artifactId, version, editableMetadata)));
+            }
 
             return null;
         });
--- a/app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
@@ -519,10 +519,11 @@ public abstract class AbstractRegistryStorageTest extends AbstractResourceTestBa
         ContentHandle content = ContentHandle.create(OPENAPI_CONTENT);
         EditableVersionMetaDataDto versionMetaDataDto = EditableVersionMetaDataDto.builder().name("Empty API")
                 .description("An example API design using OpenAPI.").build();
+        String staleOwner = "stale-creation-user";
         ArtifactVersionMetaDataDto dto = storage().createArtifact(GROUP_ID, artifactId, ArtifactType.OPENAPI,
                 null, null, ContentWrapperDto.builder().contentType(ContentTypes.APPLICATION_JSON)
                         .content(content).build(),
-                versionMetaDataDto, Collections.emptyList(), false, false, null).getValue();
+                versionMetaDataDto, Collections.emptyList(), false, false, staleOwner).getValue();
         Assertions.assertNotNull(dto);
         Assertions.assertEquals(GROUP_ID, dto.getGroupId());
         Assertions.assertEquals(artifactId, dto.getArtifactId());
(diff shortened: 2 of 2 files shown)


Nachiket-Roy:
> ### Describe the bug
> I ran into a subtle but impactful data-integrity issue in the SQL storage layer. When you PATCH an artifact version's metadata with **only** a `name` or `description` update — without including a `labels` field in the body — the API responds with HTTP 200 and saves the new value correctly, but the `modifiedBy` and `modifiedOn` fields are **never updated in the database**. They remain frozen at whatever values they had when the version was first created.
> 
> This is completely silent — there's no error, no warning, nothing in the response to indicate the audit fields were skipped. You only notice it if you inspect the returned metadata closely and compare timestamps.
> 
> There's also a secondary issue in the same method: `outboxEvent.fire(ArtifactVersionMetadataUpdated)` is called unconditionally at the end of `updateArtifactVersionMetaData()`, which means even a completely empty PATCH body (no fields at all) triggers an outbox event. Any downstream consumer connected via Kafka or Debezium will receive phantom change notifications for updates that changed nothing.
> 
> ### Steps to reproduce
> 1. Create an artifact with an initial version
> 2. Fetch the version metadata and note the `modifiedOn` timestamp
> 3. Wait a few seconds
> 4. Send a metadata update with **only** a name change:
> 
> ```httpspec
> PUT /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{version}
> Content-Type: application/json
> 
> {
>   "name": "My Updated Version Name"
> }
> ```

Hey, could you please create an issue for this and link it to the PR? The issue should describe the bug and its reproduction steps, while the PR description should focus on what you changed to fix it. We have the Issues section for tracking the problem separately.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java
```diff
@@ -248,25 +248,25 @@ public void updateArtifactVersionMetaData(String groupId, String artifactId, Str
...
+                        .bind(0, modifiedBy).bind(1, modifiedOn).bind(2, normalizeGroupId(groupId))
+                        .bind(3, artifactId).bind(4, version).execute();
+                if (rowCount == 0) {
+                    throw new VersionNotFoundException(groupId, artifactId, version);
                 }
-            }
 
-            outboxEvent.fire(SqlOutboxEvent
-                    .of(ArtifactVersionMetadataUpdated.of(groupId, artifactId, version, editableMetadata)));
+                outboxEvent.fire(SqlOutboxEvent
```
paoloantinori: This relocates the outbox-event fire inside `if (modified)`. Previously the fire sat after the labels block and ran on every call. After this change, a no-op PATCH (name, description, and labels all null) no longer emits `ArtifactVersionMetadataUpdated`, and SQL now diverges from `KafkaSqlRegistryStorage.updateArtifactVersionMetaData`, which still fires the event unconditionally. The audit-column fix is correct, but the event fire should stay outside `if (modified)` to preserve behavior and storage-variant parity, or the suppression should be an explicit, documented decision applied to both variants. As written this is an undocumented behavioral change in the replication/notification path.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java
```diff
@@ -248,25 +248,25 @@ public void updateArtifactVersionMetaData(String groupId, String artifactId, Str
...
-                            .bind(1, limitStr(k.toLowerCase(), MAX_LABEL_KEY_LENGTH))
+                            .bind(1, limitStr(asLowerCase(k), MAX_LABEL_KEY_LENGTH))
                             .bind(2, limitStr(asLowerCase(v), MAX_LABEL_VALUE_LENGTH)).execute();
                 });
+            }
 
-                if (modified) {
-                    String modifiedBy = securityIdentity.getPrincipal().getName();
-                    Date modifiedOn = new Date();
+            if (modified) {
```
paoloantinori, line 256: This block fixes modifiedBy/modifiedOn not being written on a name-only or description-only PATCH, but there is no regression test that would have caught the original bug. Add a test in `AbstractRegistryStorageTest` (near the existing call at line 539) that invokes `updateArtifactVersionMetaData` with a name set and labels null, then asserts `getArtifactVersionMetaData(...).getModifiedBy()` and `getModifiedOn()` are updated. It should fail on current main and pass on this PR; the existing test never asserts the audit fields, so the bug was invisible to the suite.

paoloantinori (commented):
The core fix is correct: moving the `modifiedBy`/`modifiedOn` write out of the `if (labels != null)` block so it runs on name-only and description-only PATCHes is the right call (the inner `if (modified)` was always-true before, so this strictly broadens when the audit columns update). Two MAJOR concerns hold back a clean merge: the diff also relocates the outbox-event fire, and there is no regression test.

- **Undocumented behavioral change in the outbox event (MAJOR):** On main the `outboxEvent.fire(ArtifactVersionMetadataUpdated...)` call sat after the labels block and fired on every call (SqlVersionRepository.java:268). This PR moves it inside `if (modified)` (SqlVersionRepository.java:267), so a no-op PATCH (name, description, and labels all null) no longer emits the event. That also breaks storage-variant parity: `KafkaSqlRegistryStorage.updateArtifactVersionMetaData` (KafkaSqlRegistryStorage.java:805) still fires the event unconditionally. Separate the audit-column fix from the event fire: keep `outboxEvent.fire` outside `if (modified)` to preserve behavior, or make the suppression an explicit decision mirrored in KafkaSQL.
- **No regression test (MAJOR):** The PR fixes audit columns silently not updating, but nothing asserts they do. The existing coverage at AbstractRegistryStorageTest.java:539 calls `updateArtifactVersionMetaData` with all three fields and never checks `getModifiedBy()`/`getModifiedOn()`. Add a name-only update (labels null) that asserts the audit fields changed; it should fail on main and pass on this PR.
- **Unrelated drive-by (MINOR):** `k.toLowerCase()` to `asLowerCase(k)` (SqlVersionRepository.java:251) is scope creep for a modifiedBy/modifiedOn fix, and the migration is incomplete: lines 461, 462, 588, and 589 in the same file still call `.toLowerCase()`. Either revert this line or finish the migration in a separate change.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlVersionRepository.java
DarkBird10020: Done!

paoloantinori:
You're right, and I owe you an apology for the noise. I went back and re-read the head `b021b2a` directly: `outboxEvent.fire(...)` at line 268 sits outside the `if (modified)` block — it never moved. My earlier comment misread the diff indentation. The SQL variant fires the event unconditionally, matching `KafkaSqlRegistryStorage.updateArtifactVersionMetaData` (line 806), so storage-variant parity holds. The audit-column fix and both new tests look good. No further concerns from me.

DarkBird10020:
Hey @paoloantinori, 
Thanks for the thorough review all great catches.

I've pushed a follow-up commit (`b021b2a`) addressing each point:

**1. Outbox event** - Moved `outboxEvent.fire()` back outside the `if (modified)` block so it fires unconditionally again, matching `KafkaSqlRegistryStorage` behavior. You're right that changing that wasn't the intent of this PR.

**2. Regression test** - Already covered in the prior commit (`e1f9fe1`). There are two tests:
- `testUpdateArtifactVersionMetaDataNameOnlyUpdatesModified` does a name-only update (labels null), asserts `modifiedBy` changes away from the stale creation owner and `modifiedOn` advances. This fails on main and passes with the fix.
- `testUpdateArtifactVersionMetaDataEmptyUpdateDoesNotModify` does a no-op update (all fields null), asserts audit fields stay unchanged.

**3. `toLowerCase`** - Reverted, `k.toLowerCase()` is back. Will leave that cleanup for a separate PR if it's wanted.

Let me know if anything else needs adjusting!

paoloantinori:
Thanks for the follow-up. Confirmed on the current head `b021b2a`: the two new tests (`testUpdateArtifactVersionMetaDataNameOnlyUpdatesModified`, `testUpdateArtifactVersionMetaDataEmptyUpdateDoesNotModify`) cover exactly the gap, and `outboxEvent.fire` stays outside `if (modified)` so the SQL and KafkaSQL variants stay in sync. Nothing further from me.

EricWittmann (approved):
LGTM

EricWittmann:
Thanks @DarkBird10020 :+1:

PR #9428, opened 2026-08-09
Fix contract metadata update validation

### LINKED ISSUE #9409: PUT /contract/metadata returns 500 for null body and skips validation
## Description

The `PUT /groups/{groupId}/artifacts/{artifactId}/contract/metadata` endpoint has two request-validation issues.

### 1. Null request body returns 500

When the request body is `null`, `updateContractMetadata()` receives `data == null` and later accesses `data.getStatus()`, causing a `NullPointerException`.

For example, the following request sends the JSON literal `null` as the complete request body:

    curl -X PUT \
      "http://localhost:8080/apis/registry/v3/groups/default/artifacts/my-artifact/contract/metadata" \
      -d 'null'

The request body is therefore the JSON value `null`, rather than an `EditableContractMetadata` object.

**Current behavior:** `500 Internal Server Error` caused by the `NullPointerException`.

**Expected behavior:** `400 Bad Request` using the existing Apicurio `ProblemDetails` error-response format.

### 2. Existing contract metadata validation is skipped

The project already provides `ContractMetadataValidator.validate(EditableContractMetadataDto)`, which contains an existing validation rule for the `supportContact` field:

- `supportContact` — when provided, it must match the existing email validation rule.

However, the `PUT /contract/metadata` update flow does not invoke this existing validator before persisting the metadata.

For example, the following request contains an invalid `supportContact` value:

{
  "status": "DRAFT",
  "supportContact": "plaintext"
}

The existing `ContractMetadataValidator` would reject this value with an `InvalidContractMetadataException`. However, because `updateContractMetadata()` does not invoke the validator, the invalid metadata can proceed through the update/persistence flow and the endpoint can return `200 OK`.

**Current behavior:** Invalid `supportContact` values can be accepted and persisted, resulting in `200 OK`.

**Expected behavior:** The existing `ContractMetadataValidator` should be invoked before persistence. If the metadata is invalid, the endpoint should return **409 Conflict** using the existing Apicurio `ProblemDetails` error-response format for `InvalidContractMetadataException`.

For valid metadata, the existing update behavior should be preserved and the endpoint should continue to return **200 OK**.

Hiten0305l:
## What

This PR is a follow-up to #9409 and fixes two request-validation problems in the `PUT /groups/{groupId}/artifacts/{artifactId}/contract/metadata` endpoint.

**Null request body:**

The endpoint could receive a JSON literal `null` as the request body. In that case, the `data` parameter was deserialized as `null`, but `updateContractMetadata()` continued processing and eventually accessed fields such as `data.getStatus()`. This caused a `NullPointerException` and resulted in an HTTP `500 Internal Server Error`.

The fix adds: 
`ParameterValidationUtils.requireParameter("data", data);`

before any access to `data`. A null request body is therefore rejected using the existing parameter-validation mechanism and returns the standard HTTP `400 Bad Request` `ProblemDetails` response.

**Contract metadata validation was skipped:**

The project already has `ContractMetadataValidator`, which validates fields in `EditableContractMetadataDto`, including `supportContact`. However, the `updateContractMetadata()` flow was not invoking this validator before updating the stored metadata. As a result, invalid values such as `"supportContact": "plaintext"` could bypass the existing validation and proceed toward persistence.

The fix calls:
`contractMetadataValidator.validate(editableDto);`

immediately after the `EditableContractMetadataDto` is created and before any metadata is read, mapped, or written to storage. This makes the update endpoint use the existing validation rules without introducing a new validation implementation.

## Why

These changes make the endpoint behave consistently with the existing Apicurio Registry validation and error-handling patterns.

A null request body should be treated as an invalid request and return `400`, rather than allowing a `NullPointerException` to become a `500` response.

For contract metadata, validation should happen before persistence so that invalid metadata cannot be stored. For example, the existing validator rejects an invalid `supportContact` such as `"plaintext"` because it does not satisfy the project's existing email format rule. The request now correctly returns `409 Conflict` with `InvalidContractMetadataException`, and the invalid value is not persisted.
Valid contract metadata continues through the existing update flow unchanged.

## Tests

Two regression tests were added to `DataContractsResourceTest`:

**`testUpdateContractMetadata_NullBody_Returns400`**

- Sends the JSON literal `null` as the request body.
- Verifies that the endpoint returns HTTP `400`.
- Verifies that the response uses the existing `MissingRequiredParameterException` `ProblemDetails` format.

**`testUpdateContractMetadata_InvalidSupportContact_Returns409`**

- Sends an invalid `supportContact` value (`"plaintext"`).
- Verifies that the endpoint returns HTTP `409 Conflict`.
- Verifies that the response contains `InvalidContractMetadataException`.
- Performs a follow-up GET request to confirm that the invalid value was not persisted.

Existing contract metadata update tests were also rechecked to ensure valid metadata updates continue to work.

Closes #9409


Hiten0305l:
Hi @paoloantinori ,Whenever you have some time, could you please review it and let me know if any changes are needed? Thank you for your time and guidance.

malladinagarjuna2:
hey @Hiten0305l  thanks for the pr 
well first of all you need to correct the commit format read contribution guidelines. 
1.the major blocker here is 400 v/s 409 
400: the request itself is malformed. It is invalid in isolation, regardless of what's on the server. Retrying it identically will never work. The client has to fix the payload.
409: the request is perfectly well-formed and would be valid at some other moment, but it clashes with the current state of the resource.
so an invalid email is malformed input, not a state conflict. `InvalidContractMetadataException` maps to 409 for both cases; the transition check deserves 409, the email check doesn't. The test locks the wrong code in.
i mean this is my thinking need to get it verified with @carlesarnal 

2. `.body("title", equalTo("Invalid email format for supportContact: plaintext"))`
Any rewording breaks the test for no reason. Assert on status and name (which it already does) and, if you want the detail, use containsString("supportContact")


3.. The identical bug is left unfixed 130 lines below. `transitionContractStatus` (line 2203) does:
ParameterValidationUtils.requireParameter("status", data.getStatus());
with no null check on data itself. Send null as the body there and you get exactly the NPE ->500 this PR is fixing.-> well can be addrreesed in further issue.

rest all looks fine, fix is  good and well tested. 


.

malladinagarjuna2:
>3.. The identical bug is left unfixed 130 lines below. transitionContractStatus (line 2203) does:
ParameterValidationUtils.requireParameter("status", data.getStatus());
with no null check on data itself. Send null as the body there and you get exactly the NPE ->500 this PR is fixing.-> well can be addrreesed in further issue.

well i would say fix it here no need to create mess already there are more than 500 issues it would be great if you can solve it here

carlesarnal:
This needs it's conflict to be fixed before proceeding.

Hiten0305l:
Thanks for the feedback @carlesarnal . I’ve addressed the requested changes and updated the PR against the latest main. The relevant tests and checks are passing locally. The updated branch has been pushed.

Hiten0305l:
Hi @carlesarnal  @EricWittmann , my other PRs are currently in draft, so if you have some time, I’d really appreciate a review of this one when convenient, so I can move forward with the other contributions as well. Thanks!

carlesarnal (approved):

PR #9435, opened 2026-08-10
Fix artifact version metadata audit updates

### LINKED ISSUE #9427: updateArtifactVersionMetaData does not update modifiedBy/modifiedOn when only name or description is changed

Sashang-debug:
## Description
Fixes #9427.

This change fixes artifact version metadata audit updates so that `modifiedBy` and `modifiedOn` are updated whenever the metadata is actually changed, including name-only, description-only, labels-only, and multi-field updates.

The outbox event is also emitted only when an actual metadata modification occurs, preventing events from being emitted for empty/no-op updates.

## Testing
Added/updated tests covering:
- name-only update
- - description-only update
- - labels-only update
- - multiple-field update
- - empty/no-op update
Verified with:
`./mvnw test -pl app -am -Dtest=DefaultRegistryStorageTest#testUpdateArtifactVersionMetaData`

Build passed successfully and `git diff --check` is clean.


app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
```diff
@@ -530,18 +530,80 @@ public void testUpdateArtifactVersionMetaData() throws Exception {
...
+        // a) name-only update
+        EditableVersionMetaDataDto nameOnly = EditableVersionMetaDataDto.builder().name("Updated Name").build();
+        storage().updateArtifactVersionMetaData(GROUP_ID, artifactId, "1", nameOnly);
 
         ArtifactVersionMetaDataDto metaData = storage().getArtifactVersionMetaData(GROUP_ID, artifactId, "1");
-        Assertions.assertNotNull(metaData);
-        Assertions.assertEquals(newName, metaData.getName());
-        Assertions.assertEquals(newDescription, metaData.getDescription());
+        Assertions.assertEquals("Updated Name", metaData.getName());
+        Assertions.assertTrue(metaData.getModifiedOn() > originalModifiedOn);
```
paoloantinori: The strict `modifiedOn > previous` comparisons here (and the `Thread.sleep(100)` before each update) assume the stored timestamp advances at least one tick between calls. If `modifiedOn` is persisted at second rather than millisecond resolution, two updates in the same wall-clock second make this `>` assertion fail intermittently on CI. Worth confirming the column resolution, or basing the assertion on something that does not depend on wall-clock advancement.

paoloantinori (commented):
Overall the fix is correct: hoisting `if (modified)` out of the labels-only branch ensures `modifiedBy` and `modifiedOn` refresh for name-only, description-only, and multi-field updates, and gating the outbox event on `modified` stops no-op calls from emitting. The granular test cases cover the right scenarios; the concerns below are about how strongly the tests actually validate the fix, not the production change itself.

- `SqlVersionRepository.java` (around line 267): the outbox event was moved inside `if (modified)`. This matches the PR intent, but it is a behavioral change. Any consumer that received an event on every metadata-update call will stop getting them for no-op updates, so it is worth confirming nothing downstream relied on the per-call event.
- `AbstractRegistryStorageTest.java:543,556,571,592`: the strict `modifiedOn > previous` assertions rely on wall-clock advancing within 100ms via `Thread.sleep`. If `modifiedOn` is stored at second rather than millisecond resolution, two updates in the same wall-clock second make these assertions intermittently fail on CI.
- `AbstractRegistryStorageTest.java:544,557,572,593`: `modifiedBy` is checked only with `assertNotNull`. A stale creation-time value is also non-null, so this cannot distinguish a refreshed value from the value left untouched, which is the exact bug being fixed.
- `AbstractRegistryStorageTest.java:599`: the no-op case validates that `modifiedOn` and `modifiedBy` are unchanged but does not assert the outbox event is suppressed, so the event half of the fix is covered only by inspection.

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
```diff
@@ -530,18 +530,80 @@ public void testUpdateArtifactVersionMetaData() throws Exception {
...
+        Assertions.assertNotNull(metaData.getModifiedBy());
```
paoloantinori: The bug being fixed is that `modifiedBy` was not re-persisted for single-field updates, yet every update case here only asserts `assertNotNull`. A stale creation-time value is also non-null, so this check cannot tell a refreshed value apart from one left untouched. Assert the concrete principal name, and ideally run one update under a different identity so the rewritten value is provably different. The no-op case below already uses `assertEquals` on `modifiedBy`, which makes the weak check here stand out.

app/src/test/java/io/apicurio/registry/noprofile/storage/AbstractRegistryStorageTest.java
```diff
@@ -530,18 +530,80 @@ public void testUpdateArtifactVersionMetaData() throws Exception {
...
+        Assertions.assertTrue(metaData.getModifiedOn() > newModifiedOn3);
+        Assertions.assertNotNull(metaData.getModifiedBy());
+
+        long newModifiedOn4 = metaData.getModifiedOn();
+        String newModifiedBy4 = metaData.getModifiedBy();
+        Thread.sleep(100);
+
+        // e) empty/no-op update
+        EditableVersionMetaDataDto emptyUpdate = EditableVersionMetaDataDto.builder().build();
+        storage().updateArtifactVersionMetaData(GROUP_ID, artifactId, "1", emptyUpdate);
```
paoloantinori, line 615: The no-op case asserts `modifiedOn` and `modifiedBy` are unchanged, which validates the modifiedBy/modifiedOn half of the fix. The other stated goal, suppressing the outbox event for no-op updates, is not asserted anywhere here. If this storage harness can observe fired outbox events, add an assertion that this no-op call emits zero; otherwise the event-suppression behavior is covered only by code inspection.

Sashang-debug:
Hi @paoloantinori , I’ve pushed the requested changes to fix-9427-metadata-update. The metadata update test has been updated to verify modifiedBy correctly, and the changes are limited to AbstractRegistryStorageTest.java.

The full Maven test suite is currently blocked by unavailable 3.3.2-SNAPSHOT dependencies, but the modified code compiles successfully offline.

Please take a look when you get a chance. Thanks!

paoloantinori:
Thanks for the update. I re-read `440452e` and the test changes in `AbstractRegistryStorageTest` look solid: the new `updateAndVerifyTime` helper polls the read (not the write) to accommodate async storage, the `staleOwner` setup ensures `modifiedBy` assertions are meaningful, and the five cases (name-only, description-only, labels-only, multi-field, no-op) cover the branches the production change touches. The no-op case correctly asserts the timestamp is unchanged, which matches the gated `if (modified)` in `SqlVersionRepository`.

On the blocked Maven build from the 3.3.2-SNAPSHOT dependency: that's an environment issue, not a code one. The production change is small and the offline compile is fine. From my side this is ready — over to a maintainer for `/accept`.

Sashang-debug:
Hi @EricWittmann, I wanted to follow up regarding PR #9435. The review appears to be complete, and the PR is currently awaiting maintainer action. When you have a chance, could you please take a look? Thank you for your time.

EricWittmann:
@Sashang-debug Hey there are unfortunately some conflicts with this now.  :(

Sashang-debug:
@EricWittmann  I have resolved the merge conflicts against the latest `main` branch. Please let me know if there's anything else needed!

EricWittmann (approved):
LGTM

EricWittmann (changes requested):
## Review: Changes requested

Thanks for the fix — the core change in `SqlVersionRepository.java` (moving `outboxEvent.fire(...)` inside the `if (modified)` block) looks correct and matches the intent of #9427. However, there are a couple of blocking issues before this can be merged.

### 1. DCO sign-off missing on most commits (blocking)

Only the first commit (`592171d`) has a `Signed-off-by` trailer. The following commits have empty message bodies and are missing sign-off:
- `440452e` – "test: fix artifact metadata update verification"
- `16c8bc5` – "Merge branch 'main' into fix-9427-metadata-update"
- `e180c66` – "chore: trigger CI"
- `d127e81` – "Merge upstream/main and resolve conflicts"

Per the project's DCO requirement, **every** commit needs a `Signed-off-by` trailer. Please rebase and sign off all commits (e.g. `git rebase --signoff main` or amend each commit), then force-push.

### 2. Flaky unit test on `app-sql` (MySQL) — needs a redesign, not just a longer timeout

```
MysqlStorageTest.testUpdateArtifactVersionMetaData
AssertionFailedError: modifiedOn should be refreshed (was 1789126584000, previous was 1789126584000)
  at AbstractRegistryStorageTest.updateAndVerifyTime(AbstractRegistryStorageTest.java:533)
```

The new `updateAndVerifyTime()` helper polls up to 50×50ms (2.5s) waiting for `modifiedOn` to strictly increase past the previous value. Because the test chains four sequential updates that each require a strictly-increasing wall-clock timestamp, it's prone to timestamp collisions under CI load/contention — this failed on the MySQL storage variant in this run. Simply increasing the timeout will reduce but not eliminate the flakiness; the test is asserting a guarantee (`modifiedOn` monotonically increases within tight polling loops across several storage variants) that the storage layer doesn't provide at millisecond granularity.

Suggestions:
- Consider asserting `>=` for updates where only the identity of the caller/logic path matters, and reserve strict `>` checks for cases with an actual injected clock or version counter you control.
- Alternatively, inject/mock the clock used for `modifiedOn` in the test so timestamps are deterministic instead of relying on wall-clock polling.
- If wall-clock polling is kept, this needs validation across **all** storage variants (H2, Postgres, MySQL, KafkaSQL) since timestamp precision/propagation delay differs per backend.

### 3. Minor: test setup complexity

The "dummy artifact" dance used to dynamically compute `defaultPrincipal` (creating `testUpdateArtifactVersionMetaData-dummy`, updating it, then reading back `modifiedBy`) is a fair bit of indirection just to make sure `staleOwner` differs from whatever the harness's default principal is. Consider simplifying — e.g., check how the test harness sets its default principal (looks like a static/config value) and pick a `staleOwner` string that's obviously distinct, rather than probing for it at runtime with an extra artifact.

### 4. Minor: outbox no-op assertion is indirect

The PR description says the outbox event should not fire on no-op updates, but the test only infers this by checking that `modifiedOn`/`modifiedBy` are unchanged — which the test's own comment acknowledges is a weaker signal ("there's no storage abstraction to test outbox events directly here"). If there's a way to spy/verify the outbox event count directly (e.g. via a test double or captured Mutiny event), that would make this assertion much stronger.

---

**Summary:** Please (1) fix DCO sign-off on all commits, and (2) address the flaky `modifiedOn` polling logic in `updateAndVerifyTime()` so it doesn't intermittently fail across storage variants. Once those are resolved, happy to take another look.

PR #9439, opened 2026-08-10
fix(ui): prevent negative values for numeric config properties

### LINKED ISSUE #9437: fix(ui): allow only non-negative integers for numeric config properties on Settings page
## Description

In the Settings page of the Web Console (`/settings`), numeric configuration properties such as **Download link expiry** (`apicurio.download.href.ttl.seconds`) allow entering and saving negative integer values (e.g., `-4000`). Since duration in seconds and count settings must be positive or zero, the UI input validation should prevent negative numbers and mark them as invalid.

**Registry Version**: snapshot (3.3.2-SNAPSHOT)
**Persistence type**: sql

### Environment

- Browser: Chrome 
- Component: Web Console UI (`ui-app`) -> Settings Page (`/settings`)

### Steps to Reproduce

1. Open the UI Web Console at `/settings`.
2. Scroll to the **Web console settings** section.
3. Click the edit (pencil) icon for **Download link expiry**.
4. Enter a negative value (e.g., `-4000`).
5. Observe that the input accepts negative numbers and enables the Save button.

- **Expected**: Numeric properties should reject negative values (`< 0`), highlight the input with an error state, and disable the Save button.
- **Actual**: Negative integers are treated as valid inputs and can be saved.

### Logs

N/A

JituRewar:
Closes #9437 

### What this PR does

Fixes an issue on the Settings page where numeric configuration properties (such as **Download link expiry** / `apicurio.download.href.ttl.seconds`) allowed entering and saving negative values (e.g., `-4000`).

### Changes made

- **Added Validation**: Updated validation logic so numeric settings only accept non-negative integers (`>= 0`).
- **HTML Restrictions**: Added `type="number"` and `min={0}` attributes to `<TextInput>` to restrict negative inputs in the browser.
- **UI Error Handling**: Marked negative inputs as invalid, showing an error border and disabling the **Save** button.

### How to test

1. Open the Registry Web Console and navigate to `/settings`.
2. Edit a numeric setting (e.g., **Download link expiry** under Web console settings).
3. Try entering a negative number like `-4000`.
4. Verify that the input displays an error state and the **Save** button is disabled.

Before-
Screenshot From 2026-08-10 18-59-26

After-
Screenshot From 2026-08-10 19-14-10


ui/ui-app/src/app/pages/settings/components/PropertyInput.tsx
```diff
@@ -60,6 +66,19 @@ export const PropertyInput: FunctionComponent<PropertyInputProps> = (props: Prop
...
+            value={ currentValue }
+            type={ props.type === "number" ? "number" : "text" }
+            min={ props.type === "number" ? 0 : undefined }
+            validated={ validated() }
+            onChange={ handleInputChange }
+            onKeyDown={ handleKeyPress }
+            aria-label="configuration property input"/>
+    </InputGroup>;
+
+};
```
paoloantinori: This `};` closes the `PropertyInput` component in the middle of its body, leaving `const errorId`, the second `return (...)`, and the trailing `};` (lines 82 onward) as module-level code. A `return` outside a function is a syntax error, so the file does not compile. This block should replace the existing return at line 84, not be added before it: remove the duplicate return/close and apply `type` and `min` to the original `TextInput`.

ui/ui-app/src/app/pages/settings/components/PropertyInput.tsx
```diff
@@ -34,6 +36,10 @@ export const PropertyInput: FunctionComponent<PropertyInputProps> = (props: Prop
     };
 
     const handleInputChange = (_event: any, value: string): void => {
+        const isValid: boolean = isPropertyInputValid(props.type, value);
+        setCurrentValue(value);
+        setIsDirty(value !== props.value);
+        setIsValid(isValid);
         const validation = validatePropertyValue(value, props.type);
 
         setCurrentValue(value);
```
paoloantinori, line 39: `setCurrentValue(value)` is now called twice (also at line 40), and `setIsDirty`/`setIsValid` are set from `isPropertyInputValid` immediately before the original `validatePropertyValue` block re-derives the same state. If the original block calls `setIsValid(validation.isValid)` after this, it overrides the new negative-check and the fix has no effect. Collapse this to a single validation call.

ui/ui-app/src/app/pages/settings/components/PropertyInput.utils.ts
```diff
@@ -1,3 +1,15 @@
+export const isPropertyInputValid = (type: "text" | "number", value: string): boolean => {
```
paoloantinori: This duplicates `validatePropertyValue` defined later in the same file. Rather than two validators (one boolean, one returning `{isValid, errorMessage}`), extend `validatePropertyValue` to reject `< 0` for numbers and return an error message, then remove `isPropertyInputValid`. One source of truth also removes the double-set conflict in `handleInputChange`.

ui/ui-app/src/app/pages/settings/components/PropertyInput.utils.ts
```diff
@@ -1,3 +1,15 @@
...
+    if (type === "text") {
+        return value.trim().length > 0;
+    } else if (type === "number") {
+        if (value.trim().length === 0) {
+            return false;
+        }
+        const num: number = Number(value);
+        return Number.isInteger(num) && num >= 0;
```
paoloantinori: Add a unit test for this pure function covering the edge cases: negatives (`-4000`) invalid, `0` and `4000` valid, `1.5` invalid (non-integer), and empty/whitespace/`abc` invalid. Without tests the negative-rejection guarantee is unverified.

ui/ui-app/src/app/pages/settings/components/PropertyInput.tsx
```diff
@@ -20,6 +20,8 @@ export type PropertyInputProps = {
     onSave: () => void;
 };
 
+import { isPropertyInputValid } from "./PropertyInput.utils";
```
paoloantinori: Move this import to the top of the file with the other imports and merge it into the existing import from `./PropertyInput.utils` (which already brings in `validatePropertyValue`) rather than declaring a second import mid-file.

paoloantinori (commented):
This PR needs another pass before merge: the patched `PropertyInput.tsx` does not compile. A second `return ...; };` was inserted mid-function, which closes the component early and strands the remaining `const errorId`, `return (...)`, and trailing `};` at module scope.

- **Compile break (`PropertyInput.tsx:69-80`)**: the inserted `return <InputGroup>...</InputGroup>;` plus `};` closes `PropertyInput` before the original JSX. The original `const errorId` (line 82), `return (` (line 84), and final `};` then become module-level code, which is a hard syntax error (a `return` outside a function plus a stray brace). It looks like the existing return was meant to be edited in place, not duplicated.
- **State set twice (`PropertyInput.tsx:40` and `:45`)**: `handleInputChange` now calls `setCurrentValue(value)` twice and sets `isDirty`/`isValid` from `isPropertyInputValid`, then falls through into the original `validatePropertyValue` block that sets the same state again. If that block resets `isValid`, the new negative-check is overridden and the fix is inert. Use one validation path.
- **Duplicate validator (`PropertyInput.utils.ts:1`)**: `isPropertyInputValid` overlaps with the existing `validatePropertyValue` in the same file. Extend `validatePropertyValue` to reject negatives and surface an error message, then drop the second function. That also removes the double-set conflict.
- **No tests (`PropertyInput.utils.ts:9`)**: the validator is a pure function with obvious edge cases. Add a unit test for `-4000` (false), `0` (true), `4000` (true), `1.5` (false), empty/whitespace (false), and `abc` (false).

JituRewar:
@paoloantinori - thank you for the detailed review! I have updated the PR addressing all items:

1. **Fixed Component Return Syntax**: Edited the existing `<TextInput>` in-place within `PropertyInput.tsx` to add `type` and `min={0}` attributes, removing the duplicate `return` block and stray brace mid-component.
2. **Single Validation Path & Removed Duplicate Validator**: Removed `isPropertyInputValid` and consolidated negative integer rejection logic inside `validatePropertyValue` in `PropertyInput.utils.ts`. `handleInputChange` now uses a single validation call and single state update.
3. **Cleaned Up Imports**: Moved and merged imports at the top of `PropertyInput.tsx`.
4. **Unit Tests**: Updated `PropertyInput.test.ts` to cover edge cases (`-4000`, `0`, `4000`, `1.5`, empty/whitespace, and `abc`). All 102 UI unit tests pass cleanly.
Let me know for further review or action.

paoloantinori:
@JituRewar thanks — re-read the current head `a8503863`. The compile blocker from my last pass is gone: the stray `return ...; };` is removed, the `type`/`min` attributes are wired into the existing `<TextInput>` in place, and the test file now asserts `0`/`4000` pass and `-123`/`-4000` fail. The validation consolidation (single path in `validatePropertyValue`, `isPropertyInputValid` dropped) matches what I asked for.

One small non-blocking note worth considering before merge: the error message reads `"Value must be a number"`, but `-4000` and `3.14` are both numerically valid that fail for different reasons (sign, non-integer). For the negative case specifically the message is slightly misleading — the input *is* a number, it's just below the allowed minimum. Something like `"Value must be a non-negative integer"` would describe the actual constraint. Not a blocker; the behavior is correct either way.

JituRewar:
Hi @paoloantinori , thanks for pointing that out! I have updated the error message for invalid numeric inputs from `"Value must be a number"` to `"Value must be a non-negative integer"` in `PropertyInput.utils.ts` and updated the unit tests in `PropertyInput.test.ts` to assert the new message. 
@EricWittmann now this pr is ready to review, so please take a look of this pr, and let me know if further changes are require.

paoloantinori:
Thanks @JituRewar — re-read the current head `b7eeafce`. The error message now matches the validation rule, the negative-input cases are covered by the tests, and the `type="number"`/`min={0}` attributes on the `<TextInput>` add a sensible browser-side guard on top of the JS validation. Looks good to me, no further concerns from my side.

JituRewar:
@EricWittmann now this pr is ready to review, so please take a look of this pr, and let me know if further changes are require.

JituRewar:
Hi @EricWittmann, thanks for accepting the PR! The triage bot is currently flagging Issue #9437 because it needs a maintainer comment on the issue itself to clear triage/red. Could you please add a quick comment on #9437

Issue #9449, opened 2026-08-10
bug(rest): addArtifactVersionComment allows null value to reach database and return HTTP 500


Hiten0305l:
## Description

The `addArtifactVersionComment()` REST endpoint does not validate whether the comment `value` is provided before passing it to the storage layer. As a result, a missing or `null` comment value can reach the database and cause an internal server error instead of a proper `400 Bad Request`.

The comment value is required by the existing application behavior: the corresponding `updateArtifactVersionComment()` endpoint already validates it using `ParameterValidationUtils.requireParameter("value", data.getValue())`. In addition, the `version_comments.cvalue` database column is defined as `NOT NULL`, so `null` cannot be persisted successfully.

### Current behavior

A request such as:

`POST /apis/registry/v3/groups/{groupId}/artifacts/{artifactId}/versions/{versionExpression}/comments`

with:

`{"value": null}`

or:

`{}`

reaches `addArtifactVersionComment()` with `data.getValue() == null`, because the add-comment implementation does not perform the corresponding value validation.

The null value is then passed to the storage layer and eventually to the `version_comments.cvalue` column.

Since `cvalue` is defined as `NOT NULL`, the database raises an integrity constraint violation, resulting in a `500 Internal Server Error` instead of a client-side validation error.

### Expected behavior

The request should be rejected at the REST layer before reaching the storage/persistence layer and return `400 Bad Request` using the existing `MissingRequiredParameterException` / `ProblemDetails` error handling.

The response should clearly indicate that the `value` parameter is required.

### Root cause

The `addArtifactVersionComment()` implementations do not validate `data.getValue()` before invoking the storage layer.

The corresponding `updateArtifactVersionComment()` methods already perform the required validation using:

`ParameterValidationUtils.requireParameter("value", data.getValue());`

The same validation is missing from the add-comment implementations.

### Affected API versions

The issue exists in both:

- v3 REST API — `app/src/main/java/io/apicurio/registry/rest/v3/impl/GroupsResourceImpl.java`
- v2 REST API — `app/src/main/java/io/apicurio/registry/rest/v2/impl/GroupsResourceImpl.java`

### Suggested fix

Add the existing parameter validation to both `addArtifactVersionComment()` implementations:

`ParameterValidationUtils.requireParameter("value", data.getValue());`

The validation should happen before calling the storage layer.

[...]

PR #9454, opened 2026-08-10
fix(rest): validate comment value before persistence

Hiten0305l:
## What

This pull request is a follow-up to issue #9449
The `addArtifactVersionComment()` implementations in both the v2 and v3 REST APIs did not validate the comment `value` before passing it to the storage layer. As a result, requests with a missing or `null` value could reach the database, where the `version_comments.cvalue` column is defined as `NOT NULL`.
This change adds the existing `ParameterValidationUtils.requireParameter("value", data.getValue())` validation to both v2 and v3 `addArtifactVersionComment()` implementations, ensuring invalid requests are rejected before persistence.

## Why

A request such as:

`{"value": null}`
or:
`{}`

previously reached the persistence layer and caused a database integrity constraint violation, resulting in an HTTP `500 Internal Server Error`.
The expected behavior is to reject the invalid request at the REST layer with an HTTP `400 Bad Request` using the existing `MissingRequiredParameterException` / `ProblemDetails` handling.
This also makes the add-comment behavior consistent with the existing `updateArtifactVersionComment()` validation.

## Test

Added a regression test covering both invalid request payloads:

- `{"value": null}` → HTTP `400 Bad Request`
- `{}` → HTTP `400 Bad Request`

The test verifies that invalid comment values are rejected before reaching the database.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v2/impl/GroupsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v2/impl/GroupsResourceImpl.java
@@ -787,6 +787,7 @@ public class GroupsResourceImpl implements GroupsResource {
         ParameterValidationUtils.requireParameter("groupId", groupId);
         ParameterValidationUtils.requireParameter("artifactId", artifactId);
         ParameterValidationUtils.requireParameter("version", version);
+        ParameterValidationUtils.requireParameter("value", data.getValue());
 
         CommentDto newComment = storage.createArtifactVersionComment(defaultGroupIdToNull(groupId),
                 artifactId, version, data.getValue());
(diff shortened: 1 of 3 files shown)


EricWittmann:
@Hiten0305l can I ask why this is a Draft PR?

EricWittmann:
Instead of a standard PR, I mean.  :)

Hiten0305l:
Hi @EricWittmann , thanks for checking. My previous PR (#9376) is already ready for review, so while waiting for it to be merged, I worked on this follow-up issue separately. Since GitHub is currently limiting me to one open PR, I’ve kept this PR as a Draft for now. Once the previous PR is merged, I’ll mark this one as ready for review.

Thanks for your understanding!

PR #9462, opened 2026-08-10
fix(ui): add navigation link to "Not a Draft" empty state in EditorPage

### LINKED ISSUE #9460: Editor shows dead-end "Not a Draft" screen with no way to navigate away
## Summary
When a version's editor page (`/explore/:groupId/:artifactId/versions/:version/editor`)
is open for a DRAFT version, and that version is transitioned out of DRAFT
status (e.g. Enabled) while the editor tab is still open, refreshing or
revisiting that editor tab shows a "Not a Draft" empty state with no way
to navigate anywhere except the browser's back button.

## Steps to reproduce
1. Enable the `draftMutability` feature flag in Settings.
2. On any Enabled artifact version, use the kebab menu → "Create draft
   from..." to create a new DRAFT version.
3. Click "Edit draft" to open the editor for this draft.
4. In a separate tab, open the same version's page and change its status
   from Draft to Enabled.
5. Return to the editor tab (still open on step 3's URL) and refresh (F5).

## Expected
Some way to navigate back to the version page from this dead-end state.

## Actual
Only the "Not a Draft" message is shown, no button or link  the browser
back button is the only way out.

## Screenshot
**Before:**

Image

Faisal77666:
Fixes #9460

## What this PR does
Adds a "Go to version page" button to the "Not a Draft" empty state shown
in `EditorPage.tsx` when a version's editor is accessed but the version is
no longer in DRAFT status.

## Why
Previously, if a user's editor session became stale  for example, the
version was transitioned from DRAFT to ENABLED (by another tab or another
user) while the editor was open  refreshing the editor showed a dead-end
screen with no way to navigate away except the browser's back button.

## Changes
- `ui/ui-app/src/app/pages/editor/EditorPage.tsx`
  - Added imports: `Link` (react-router), `Button` (PatternFly),
    `useAppNavigation`
  - Added `appNavigation` hook usage
  - Added a "Go to version page" button inside `notDraftEmptyState`,
    linking to `/explore/:groupId/:artifactId/versions/:version`

## Testing done
- `tsc && vite build` clean
- Manually reproduced the dead-end via UI (draft → edit → enable in
  another tab → refresh editor) and confirmed the button now appears
  and correctly navigates back to the version page

## Screenshots
**Before:**

Screenshot From 2026-08-10 14-08-35

**After:**

Screenshot From 2026-08-10 13-57-42

### DIFF
--- a/ui/ui-app/src/app/pages/editor/EditorPage.tsx
+++ b/ui/ui-app/src/app/pages/editor/EditorPage.tsx
@@ -22,6 +22,9 @@ import {
     toPageError
 } from "@app/pages";
 import { RootPageHeader } from "@app/components";
+import { Link } from "react-router";
+import { Button } from "@patternfly/react-core";
+import { AppNavigation, useAppNavigation } from "@services/useAppNavigation.ts";
 import { ContentTypes } from "@models/ContentTypes.ts";
 import { PleaseWaitModal } from "@apicurio/common-ui-components";
 import { Draft, DraftContent } from "@models/drafts";
@@ -106,6 +109,7 @@ export const EditorPage: FunctionComponent<PageProperties> = () => {
     const [referencesLoadFailed, setReferencesLoadFailed] = useState(false);
 
     const params = useParams();
+    const appNavigation: AppNavigation = useAppNavigation();
     const groupId = params.groupId as string;
     const draftId: string = params.artifactId || "";
     const version = params.version as string;
@@ -408,6 +412,17 @@ export const EditorPage: FunctionComponent<PageProperties> = () => {
             <EmptyStateBody>
                 This artifact is not in <em>DRAFT</em> status and so its content cannot be edited.
             </EmptyStateBody>
+            <Button
+                variant="primary"
+                component={(props: any) => (
+                    <Link
+                        {...props}
+                        to={appNavigation.createLink(`/explore/${groupId}/${draftId}/versions/${version}`)}
+                    />
+                )}
+            >
+                Go to version page
+            </Button>
         </EmptyState>
     );
 
--- a/ui/ui-app/src/app/pages/editor/EditorPage.tsx
+++ b/ui/ui-app/src/app/pages/editor/EditorPage.tsx
@@ -50,6 +50,7 @@ import { useEditorDraftRecovery } from "./useEditorDraftRecovery.ts";
 import { useEditorReauthentication } from "./useEditorReauthentication.ts";
 import { ArtifactReference, ReferenceTypeObject } from "@sdk/lib/generated-client/models";
 import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
+import { buildNonDraftVersionLink } from "./EditorPage.utils.ts";
 
 const sectionContextStyle: CSSProperties = {
     borderBottom: "1px solid #ccc",
@@ -417,7 +418,7 @@ export const EditorPage: FunctionComponent<PageProperties> = () => {
                 component={(props: any) => (
                     <Link
                         {...props}
-                        to={appNavigation.createLink(`/explore/${groupId}/${draftId}/versions/${version}`)}
+                        to={appNavigation.createLink(buildNonDraftVersionLink(groupId, draftId, version))}
                     />
                 )}
             >
--- a/ui/ui-app/src/app/pages/editor/EditorPage.utils.test.ts
+++ b/ui/ui-app/src/app/pages/editor/EditorPage.utils.test.ts
@@ -1,0 +1,11 @@
+import { describe, expect, it } from "vitest";
+import { buildNonDraftVersionLink } from "./EditorPage.utils.ts";
+
+describe("buildNonDraftVersionLink", () => {
+
+    it("builds the version page path from groupId, artifactId, and version", () => {
+        const link = buildNonDraftVersionLink("default", "my-artifact", "1");
+        expect(link).toBe("/explore/default/my-artifact/versions/1");
+    });
+
+});
--- a/ui/ui-app/src/app/pages/editor/EditorPage.utils.ts
+++ b/ui/ui-app/src/app/pages/editor/EditorPage.utils.ts
@@ -1,0 +1,3 @@
+export function buildNonDraftVersionLink(groupId: string, artifactId: string, version: string): string {
+    return `/explore/${groupId}/${artifactId}/versions/${version}`;
+}


ui/ui-app/src/app/pages/editor/EditorPage.tsx
```diff
@@ -408,6 +412,17 @@ export const EditorPage: FunctionComponent<PageProperties> = () => {
             <EmptyStateBody>
                 This artifact is not in <em>DRAFT</em> status and so its content cannot be edited.
             </EmptyStateBody>
+            <Button
```
paoloantinori, line 416: Consider a small render test for this empty state asserting the button links to `/explore/${groupId}/${artifactId}/versions/${version}`. The manual repro covers the happy path, but a test would guard against regressions if the route shape ever changes. No functional concerns with the change itself otherwise.

paoloantinori (commented):
Solid, well-scoped change that closes a real dead-end in the editor empty state. The `component` render-prop is the right PatternFly and react-router integration: it renders the Button as a Link, avoiding the invalid `<a><button></button></a>` nesting you would get from wrapping. No correctness or security issues found, and `draftId` correctly resolves to `params.artifactId` so the version URL is accurate.

- The new empty-state link has no automated test. The manual repro in the PR description covers the happy path, but a small render test asserting the button target resolves to `/explore/:groupId/:artifactId/versions/:version` would catch regressions if the route shape changes.
- `EditorPage.tsx:114`: `draftId` holds the artifact ID and is now used to build a non-draft version link. This is pre-existing naming, not introduced here, but it reads oddly in this context and is worth a follow-up rename.
- The link assumes `groupId`, `draftId`, and `version` are non-empty. Routing guarantees that on this route, so this is low risk; if you want to be defensive, guarding against an empty param avoids a malformed target like `.../versions/undefined`.

Faisal77666:
@paoloantinori Thanks for the review! I looked into adding a render test, but the project doesn't currently
have any React component render-testing setup  no @testing-library/react,
no jsdom/happy-dom, and vitest.config is set to `environment: "node"`.
All existing .test.ts files test pure functions only.
Given that, I see two options:
1. Extract the link-building logic into a small pure function and cover
   it with a .test.ts unit test (consistent with the existing pattern,
   no new dependencies)
2. Add full render-testing infra (@testing-library/react + jsdom) to
   properly render the button and assert its target
Happy to do either, just didn't want to quietly add new testing
infrastructure to the project as a side effect of a small UI fix
without checking first. Which would you prefer?

paoloantinori:
Your read of the test setup is correct — no render-testing infrastructure in the project, and introducing jsdom/@testing-library just for this 15-line change would be disproportionate.

Go with option 1 if you want a test at all, but I wouldn't block on it. The link path is a single template-string interpolation with no branching logic, so the pure function would be trivial (`/explore/${groupId}/${artifactId}/versions/${version}`). A unit test there is fine as belt-and-suspenders but it's not guarding anything non-obvious.

If you do extract it, make sure to use `artifactId` (not `draftId`) as the path segment name in the extracted function — the route param is `artifactId` and the current code reads `params.artifactId || ""` into a variable named `draftId`, which works but is a naming smell worth not propagating into a new function.

Either way, the change itself is good to go from my side.

Faisal77666:
@paoloantinori Done extracted the path-building logic into EditorPage.utils.ts
(matching the existing .utils.ts test pattern used elsewhere in the
codebase, e.g. PromptTemplateViewer.utils.ts) and added a unit test
for it. Used `artifactId` as the parameter name per your note, so it
doesn't propagate the draftId/artifactId naming ambiguity into new code.

paoloantinori:
Looks good — the extraction into `EditorPage.utils.ts` with the `artifactId` parameter name is exactly what I had in mind, and the unit test matches the existing `.utils.ts` pattern in the codebase. The naming ambiguity stays bounded to the existing `draftId` local at the call site and does not propagate into the new code.

No further concerns from me. LGTM.

Faisal77666:
@paoloantinori  while researching PROMPT_TEMPLATE rendering behavior for the
LFX Term 3 Playground proposal, I ran into a few things I'd rather ask
about than assume. Would really appreciate your take on any of these:
1. Grammar mismatch which one's the intended contract?
   PromptTemplateContentValidator checks variable names against a strict
   pattern, but PromptRenderingService renders using a looser one that
   accepts more than the validator allows. Is the validator meant to be
   a deliberate subset of what render supports, or is that just drift
   nobody's caught yet? Trying to understand which one should be treated
   as the actual spec.
2. MCP's conditional support vs. REST  how far should it go?
   MCP already evaluates {{#if}} blocks; REST's renderer only does flat
   substitution. If a Playground editor is meant to show accurate live
   previews, should it target what REST currently does, what MCP already
   does, or is there a longer-term plan to unify the two that I should
   be aware of before assuming either direction?
3. A smaller one  is this pattern still needed?
   In the MCP converter there's a variable-matching pattern declared
   near the top of the file that doesn't seem to be referenced anywhere
   in the actual substitution logic. Am I missing where it's used, or is
   it just left over?
4. Version tagging  reuse existing labels, or is a real stage model
   preferred?
   There's a free-text label mechanism already in place, and separately
   a DEV/STAGE/PROD enum scoped to ODCS contracts. For something like
   version tagging on prompt templates, would maintainers rather see the
   existing free-text labels reused with a suggested vocabulary, or is
   a more formal version-stage model actually the direction you'd want,
   even if it's more work?
No urgency on any of this  mostly want to make sure I'm reasoning about
the right contract before I say anything more definite elsewhere. Happy
to go however deep is useful, or not at all if this isn't the right
venue for it.

paoloantinori:
Thanks for thinking carefully about these, they are all fair questions and worth a real answer. But this thread is attached to a 30-line UI fix for the editor empty state, and the questions are about prompt-template rendering contracts, the MCP vs. REST rendering split, an unused pattern in the MCP converter, and version-tagging strategy. Those are design decisions on a different surface area, and answering them inline here would bury them inside an unrelated PR where the right people won't see them.

For each of the four, the right move is a dedicated venue rather than this PR thread:

- (1) validator vs. renderer grammar mismatch and (2) MCP conditional support vs. REST flat substitution: these are contract questions that affect both the runtime and any Playground preview, so they belong in a GitHub Discussion (or an issue scoped to `PromptRenderingService` / `PromptTemplateContentValidator`) where the MCP and REST owners can weigh in together. I'd rather not give you a half-checked answer here and have it become the record.
- (3) the unreferenced variable-matching pattern in the MCP converter: that one is concrete and checkable. If you file it as its own issue with the file and line, it can be triaged quickly (dead code removal is a fast yes/no once someone looks).
- (4) free-text labels vs. a formal DEV/STAGE/PROD stage model: that is a broader modeling decision for prompt-template versioning, and a Discussion is the right place, especially since you mentioned it ties into the LFX Term 3 Playground proposal.

If you open a Discussion or issues for (1)+(2) and (4), and an issue for (3), point me at them and I'll engage there. The editor empty-state PR itself is good from my side and I don't want to block it on these.

Faisal77666:
@paoloantinori Makes sense, appreciate the guidance. Opened:
 Discussion #9464  grammar contract, MCP/REST, version tagging
 Issue #9465  unused VARIABLE_PATTERN
Will let you know if/when there's more to discuss. Thanks for not
blocking the PR on this.

paoloantinori:
Thanks for following through, Faisal. Both the Discussion (#9464) and the issue (#9465) are the right place to take the grammar-contract and unused-pattern questions. Happy to pick those up there. Nothing further on this PR from my side.

EricWittmann:
Thanks for the contribution, @Faisal77666! Merged.

PR #9466, opened 2026-08-11
fix(ui): auto-generate typescript-sdk client in init-dev.sh on fresh …

### LINKED ISSUE #8975: docs(ui): UI setup fails on fresh clone due to unbuilt typescript-sdk (@sdk)
### Description

**Registry Version**: `3.3.1-SNAPSHOT` (`main` branch)
**Persistence type**: `in-memory`

When attempting to set up and run the UI development environment on a clean repository clone by following the current `README.md` instructions (`./mvnw clean install -Dlocal -DskipTests` -> `cd app && ../mvnw quarkus:dev`, then `cd ui && npm install && cd ui-app && ./init-dev.sh && npm run dev`), the Vite development server fails to start due to a missing dependency import.

### Environment

- **OS:** Linux
- **Node.js:** v20.x
- **npm:** v10.x
- **Java:** JDK 17 / 21
- **Branch:** `main` (Apicurio Registry v3)

### Steps to Reproduce

1. Clone a fresh copy of the repository.
2. Start the Quarkus backend server locally:
   ```bash
   ./mvnw clean install -Dlocal -DskipTests
   cd app/
   ../mvnw quarkus:dev
3.In a separate terminal, navigate to the UI directory and run:
cd ui
npm install
cd ui-app
./init-dev.sh
npm run dev
4.Observe the Vite build error in the console indicating @sdk/lib/generated-client/models cannot be resolved.


Expected: npm run dev should build and serve the UI on http://localhost:8888 without module resolution errors.

Actual: Vite crashes during build because typescript-sdk/lib/generated-client has not been generated on a fresh clone.

### Detailed Root Cause Analysis

    In Apicurio Registry v3, the TypeScript client SDK lives in the top-level /typescript-sdk directory.

    On a fresh git clone, the Kiota-generated client models located at typescript-sdk/lib/generated-client do not exist yet.

    ui/ui-app/init-dev.sh currently only copies static configuration files (version.js and config-local.js -> config.js). It does not invoke Kiota generation (npm run generate-sources) or TypeScript compilation (npm run build) inside /typescript-sdk.

### Logs

[vite] Internal server error: Failed to resolve import "@sdk/lib/generated-client/models" from "src/app/pages/search/SearchPage.tsx". Does the file exist?


### Proposed Solution

    Automation (init-dev.sh): Update ui/ui-app/init-dev.sh to automatically check if ../../typescript-sdk/lib/generated-client exists, and run npm run generate-sources & npm run build inside typescript-sdk if it is missing.

    Documentation (README.md): Update the ## Getting started (UI) section in README.md to explicitly mention building /typescript-sdk prior to launching npm run dev.

krishkumar1577:
Fixes #8975

## Problem
On a fresh clone, `npm run dev` in `ui/ui-app` fails with an unresolved
`@sdk/lib/generated-client/models` import, because `init-dev.sh` only
copies config files and never builds the `typescript-sdk` package,
which requires a separate Kiota generation step.

## Fix
`init-dev.sh` now checks whether `typescript-sdk/lib/generated-client`
exists. If it's missing (fresh clone), it automatically runs
`npm install`, `npm run generate-sources`, and `npm run build` inside
`typescript-sdk` before proceeding.

## Testing
Simulated a fresh clone by removing `typescript-sdk/lib/generated-client`,
`dist`, `node_modules`, and `.kiota`, then ran `./init-dev.sh` followed
by `npm run dev` — confirmed the SDK auto-builds and the UI starts
cleanly with no import errors.

Thanks to @amogh8021 for the original investigation and root cause
analysis in the issue.


krishkumar1577:
@amogh8021 PR is up — feel free to take a look if you get a chance, would appreciate your eyes on it given you did the original investigation.

amogh8021:
I tested this locally against the fresh-clone scenario by removing `typescript-sdk/lib/generated-client`, `dist`, `node_modules`, and `.kiota`.

`./init-dev.sh` correctly detected the missing generated client, ran Kiota generation and the TypeScript SDK build successfully. After that, `npm run dev` started Vite successfully on `http://localhost:8888` without the original `@sdk/lib/generated-client/models` resolution error.

The Registry connection issue I encountered appears unrelated to this PR, as the UI itself starts successfully.

Looks good to me. Thanks for picking this up!

paoloantinori (commented):
Overall the change does what it claims: it builds `typescript-sdk` on a fresh clone before `npm run dev`, and the `$?` check after the build subshell correctly aborts on failure. The main weakness is that the rebuild is gated purely on whether `lib/generated-client` exists, which silently breaks down for partial builds and for stale clients once the SDK spec changes.

- `ui/ui-app/init-dev.sh:16`: directory existence is a weak proxy for a complete, current client. An interrupted first run can leave a partial `lib/generated-client`, so later runs skip the rebuild and `npm run dev` fails with the same unresolved `@sdk/lib/generated-client/models` import this PR is meant to fix.
- `ui/ui-app/init-dev.sh:16`: once the directory exists the build never re-runs, so when the SDK OpenAPI spec changes during normal development the generated client goes stale with no warning to the UI dev.
- `ui/ui-app/init-dev.sh:22`: the chain runs `npm run build` (which produces `dist`) but the guard only inspects `lib/generated-client`; if `dist` is missing while the generated-client dir is present, the rebuild is skipped and the import may still fail. Confirm which artifact the `@sdk` alias resolves to and guard on that.

ui/ui-app/init-dev.sh
```diff
@@ -1,17 +1,33 @@
...
   CONFIG_TYPE="local"
 fi
-
 cp configs/version.js version.js
 cp configs/config-$CONFIG_TYPE.js config.js
 
+SDK_DIR="../../typescript-sdk"
+GENERATED_CLIENT_DIR="$SDK_DIR/lib/generated-client"
+
+if [ ! -d "$GENERATED_CLIENT_DIR" ]
```
paoloantinori: Gating the rebuild on `[ ! -d "$GENERATED_CLIENT_DIR" ]` is fragile because directory existence is a poor proxy for a complete, current client. Two realistic cases defeat it. First, an interrupted first build can leave a partial `lib/generated-client`, so every later run skips the rebuild and `npm run dev` fails with the same unresolved `@sdk/lib/generated-client/models` import this change is meant to fix. Second, once the directory exists the client is never regenerated, so after the SDK OpenAPI spec changes the UI silently gets a stale client. Checking a concrete generated artifact (the entry file the import resolves to) rather than the bare directory, plus an opt-in way to force regeneration (for example a `--rebuild-sdk` flag), would cover both.

ui/ui-app/init-dev.sh
```diff
@@ -1,17 +1,33 @@
...
+then
+  echo "----"
+  echo "typescript-sdk generated client not found (fresh clone detected)."
+  echo "Building typescript-sdk now: npm install, generate-sources, build..."
+  echo "----"
+  (cd "$SDK_DIR" && npm install && npm run generate-sources && npm run build)
```
paoloantinori: The chain produces `dist` via `npm run build`, but the guard at line 16 only inspects `lib/generated-client`. If `dist` is removed or missing while `lib/generated-client` is present, the rebuild is skipped even though the `@sdk/...` import may still fail to resolve. Worth confirming which path the `@sdk` alias actually points at (generated source under `lib/generated-client` vs. compiled `dist`), guarding on that artifact, and dropping `npm run build` from the chain if the dev server consumes the generated source directly.

EricWittmann:
@krishkumar1577 I understand the motivation for this.  However, I don't think init-dev.sh is the right place to solve this problem.  I think developers should be running `npm run dev` from the `./ui` directory rather than the `./ui/ui-app` directory.  Check if doing that works on a fresh clone or not.  If **not** then I think the solution is a tweak to the `package.json` file in `./ui`.

wdyt?

krishkumar1577:
Hey @EricWittmann — tested this directly — ran npm install && npm run dev from ui/ (not ui/ui-app) against a fully fresh typescript-sdk (no generated-client, dist, node_modules, or .kiota).

The install-sdk postinstall script does run cd ../typescript-sdk && npm install, but it stops there — it never calls generate-sources or build. So the same @sdk/lib/generated-client/models resolution error still occurs, just triggered from ui/ instead of ui/ui-app.

Given that, would it make sense to move the fix into ui/package.json's install-sdk script instead of init-dev.sh — i.e. have that script run generate-sources && build after npm install? That would fix it for the normal npm install flow you're describing, not just the init-dev.sh path.

@paoloantinori's point about staleness still applies wherever this lives — happy to guard on the actual entry file the @sdk alias resolves to instead of just directory existence, and could add an explicit rebuild flag too. Let me know if this direction sounds right and I'll update the PR.

paoloantinori:
Your diagnosis matches what I see in `ui/package.json`: `install-sdk` (line 8) runs only `cd ../typescript-sdk && npm install`, while the generation+build already live in a separate `build-sdk` script (line 24) that neither `install-sdk` nor `dev` invoke. So on a fresh clone the generated client never gets built, which is exactly the gap you hit.

Moving the fix here aligns with Eric's direction and the repo's existing script split. Two things to settle on the shape, both independent of where the code lives:

1. Avoid making Kiota generation run on every `npm install`, since that penalizes the normal iterative dev flow. A dedicated setup step the dev runs once (or that `postinstall` calls only when the generated entry file is missing) is cleaner than unconditionally chaining generate+build into `install-sdk`.
2. The staleness concern from the earlier review still applies: guard on the concrete entry file the `@sdk` alias resolves to (whichever of `lib/generated-client/<entry>` or `dist/<entry>` the import actually pulls), not on directory existence, and offer an explicit way to force regeneration when the OpenAPI spec changes. Otherwise a partial first build or a spec update silently leaves the UI against a stale client.

Happy to re-read once you decide between "setup step the dev runs once" vs "conditional generate inside postinstall."

EricWittmann:
> Given that, would it make sense to move the fix into ui/package.json's install-sdk script instead of init-dev.sh — i.e. have that script run generate-sources && build after npm install? That would fix it for the normal npm install flow you're describing, not just the init-dev.sh path.

Yes I think this is the way!  :+1:

krishkumar1577:
Thanks for the review @paoloantinori and @EricWittmann!

The fix in ui/package.json makes install-sdk auto-build typescript-sdk if the generated client is missing. This covers both the init-dev.sh fresh clone path and the normal npm install flow.

Eric already confirmed the direction: so i  checked on lib/generated-client directory gates the rebuild, and it leverages the existing build-sdk script.

Happy to iterate on the staleness guard if needed, but the core fix is in place.

paoloantinori:
@krishkumar1577 happy to iterate on the staleness guard when you get to it, but no need to block on my account here — the core fix (gating the rebuild on `lib/generated-client` and reusing the existing `build-sdk` script) is in place and covers both the `init-dev.sh` fresh-clone path and the normal `npm install` flow. From my side this is ready for Eric's final call on the `ui/package.json` placement.

krishkumar1577:
Thanks @paoloantinori  , appreciate the clarification and the review. I’ve updated the PR to place the setup logic in ui/package.json while keeping the existing init-dev.sh fresh-clone behavior. I also verified the SDK and UI builds locally. I’ll leave the staleness guard as a follow-up for later . Thanks again for taking the time to review this !! ✌️

paoloantinori:
Thanks @krishkumar1577 — the ui/package.json placement looks right. Gating the rebuild on lib/generated-client/models/index.ts plus the dist artifacts means postinstall stays a no-op for existing clones and only rebuilds on the fresh-clone case, which matches what we discussed. The staleness guard as a follow-up is fine on my end.

krishkumar1577:
@EricWittmann just checking if this needs anything further from my side, happy to make changes

EricWittmann (commented):
@krishkumar1577 Thanks for the update — the `ui/package.json` fix looks good and covers the main use case.

One thing: the SDK build logic is now in **both** `ui/package.json` (the `install-sdk` script) and `ui/ui-app/init-dev.sh`. Since developers should be running `npm install` from `ui/` — which triggers `install-sdk` via `postinstall` — the `init-dev.sh` check is redundant. If someone runs `init-dev.sh` after `npm install`, the SDK is already built.

Could you remove the SDK build logic from `init-dev.sh` and revert it back to its original form (just the config file copies)? That way the fix lives in one place and there's no inconsistency between the two guards (the `package.json` version checks specific files while `init-dev.sh` only checks the directory).

The cosmetic blank-line removals in `init-dev.sh` should also be reverted to keep the PR focused on the actual fix.

Does that make sense to you?

krishkumar1577:
@EricWittmann Makes sense, thanks for catching that  reverted init-dev.sh back to its original form in the latest commit. The SDK build logic now lives solely in ui/package.json's install-sdk script, no duplication.

EricWittmann:
Thanks for the contribution, @krishkumar1577! Merged.

PR #9467, opened 2026-08-11
fix: replace custom HSTSFilter with Quarkus-managed security headers

### LINKED ISSUE #8723: Evaluate replacing custom `HSTSFilter` with Quarkus-managed security headers
Following #8720 (which fixes RFC 6797 directive casing in the current implementation), we should evaluate whether `HSTSFilter` can be replaced by Quarkus-managed HTTP security headers.

### Motivation
- Reduce custom security-header logic in servlet/filter code.
- Lower risk of future RFC drift (e.g., directive formatting/casing issues).
- Prefer framework-supported configuration where possible.

### Scope
- Investigate if Quarkus built-in security header support can cover current `HSTSFilter` behavior in this deployment/runtime layer.
- Compare behavior and defaults with current implementation (including max-age, `includeSubDomains`, and any environment-specific conditions).
- Identify gaps/constraints that would block migration.

### Acceptance Criteria
- [ ] Feasibility is documented (supported vs not supported in our layer/runtime).
- [ ] Behavior parity matrix is provided (current filter vs Quarkus config).
- [ ] Decision documented:
  - [ ] Migrate to Quarkus-managed headers (with plan), **or**
  - [ ] Keep `HSTSFilter` with rationale.
- [ ] If migration is feasible, create a follow-up implementation PR with regression tests.

### References
- PR: #8720

Ganesh-403:
## Description

Replaces the custom `HSTSFilter` servlet filter with Quarkus-managed HTTP header configuration.

This addresses #8723 by moving the static HSTS configuration to Quarkus while preserving the existing `Strict-Transport-Security` value and maintaining the required behavior across the servlet `response.reset()` path.

## Changes

- Remove the custom `HSTSFilter`.
- Remove the `HSTSFilter` registration from `web.xml`.
- Configure `Strict-Transport-Security` using Quarkus HTTP headers:
  `max-age=31536000; includeSubDomains`
- Preserve `Strict-Transport-Security` in the existing `resetKeepingHeaders` path for the disabled-API 404 case, where `HttpServletResponse.reset()` clears response headers.
- Keep `ContentTypeOptionsFilter` unchanged so `X-Content-Type-Options: nosniff` remains handled as before.
- Update affected HSTS test comments to reflect the Quarkus-managed implementation.

## Testing

- `git diff --check` passes.
- Executed the HSTS integration tests in the `app` module:
  `HstsHeaderTest,HstsError500Test,HstsRedirectHeaderTest,HstsAuthHeaderTest`
- Results: **12 tests run, 0 failures, 0 errors, 0 skipped — BUILD SUCCESS**.
- Verified the disabled-API 404 path where `HttpServletResponse.reset()` is called.
- Verified the 302 redirect path where `sendRedirect()` commits the response.

## Related Issue

Closes #8723


app/src/main/java/io/apicurio/registry/rest/RegistryApplicationServletFilter.java
```diff
@@ -44,7 +44,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             if (disabled) {
                 HttpServletResponse httpResponse = (HttpServletResponse) response;
                 // reset() would clear the security headers the upstream filters set; preserve them.
-                resetKeepingHeaders(httpResponse, "Strict-Transport-Security", "X-Content-Type-Options");
+                resetKeepingHeaders(httpResponse, "X-Content-Type-Options");
```
paoloantinori: This is the load-bearing risk in the PR. resetKeepingHeaders (lines 61-71) snapshots the named headers, calls response.reset(), then re-sets only those names; response.reset() clears every header set upstream, which is why this method exists. With HSTS moved upstream to Quarkus's Vert.x layer and dropped from the preserve list, this 404 will lose Strict-Transport-Security unless Quarkus reapplies the quarkus.http.header value after the servlet reset. HstsHeaderTest#testHstsOnDisabledApi404 asserts HSTS on exactly this response and, per the PR description, was not run. Either run that test to green or re-add Strict-Transport-Security to the preserve list.

app/src/main/resources/application.properties
```diff
@@ -13,6 +13,7 @@ quarkus.http.auth.proactive=false
 quarkus.security.users.embedded.enabled=true
 
 # HTTP
+quarkus.http.header."Strict-Transport-Security".value=max-age=31536000; includeSubDomains
```
paoloantinori, line 16: Key and value are right: the quoted-key form is the documented Quarkus pattern, and the value matches the old constant (max-age=31536000; includeSubDomains, the casing pinned by HstsHeaderTest#testHstsDirectiveCasingIsRfcCompliant). The concern is behavioral, not syntactic: it is unverified whether this Vert.x-applied header survives response.reset() on the disabled-API 404 and the response commit on the RedirectFilter 302. Please run HstsHeaderTest, HstsError500Test, HstsRedirectHeaderTest, and HstsAuthHeaderTest and confirm the header is present on all of them before merge.

app/src/test/java/io/apicurio/registry/headers/HstsRedirectHeaderTest.java
```diff
@@ -11,7 +11,7 @@
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
 
-// RedirectFilter returns early after sendRedirect(), bypassing HSTSFilter - the 302 must still get HSTS.
+// RedirectFilter returns early after sendRedirect() - Quarkus Vert.x headers ensure the 302 still gets HSTS.
```
paoloantinori, line 14: The new comment states Quarkus Vert.x headers guarantee HSTS on the 302. sendRedirect() commits the response, so coverage depends on whether Quarkus sets the header before the servlet dispatch (survives) or at flush time (may not, because the redirect already committed the response). This test is the verification for that claim, and the PR description notes the targeted tests were stopped before completion. Please confirm HstsRedirectHeaderTest passes.

paoloantinori (commented):
This should not merge as-is: moving HSTS to quarkus.http.header is unverified on the one response path the old servlet filter existed to protect (the disabled-API 404, which calls response.reset()), and the PR description states the HSTS tests were stopped before they completed. The config itself is correct (right key, quoted-key syntax, value and casing match the old constant), but resetKeepingHeaders is in-code proof that response.reset() wipes headers set upstream, so the 404 will likely lose Strict-Transport-Security under the new mechanism unless Quarkus reapplies the header after the reset.

- RegistryApplicationServletFilter.java:47 - HSTS was removed from resetKeepingHeaders. That method (lines 61-71) snapshots the named headers, calls response.reset(), and re-sets only those names; response.reset() clears headers set upstream, which is why the method exists. Moving HSTS upstream to Quarkus's Vert.x layer and dropping it from the preserve list means the disabled-API 404 ships without HSTS unless Quarkus reapplies the header after the servlet reset. HstsHeaderTest#testHstsOnDisabledApi404 (line 67) is the test that decides this, and it was not run.
- application.properties:16 - Key and value are correct: the quoted-key form is the documented Quarkus pattern, and the value matches the old constant (max-age=31536000; includeSubDomains, the casing pinned by HstsHeaderTest#testHstsDirectiveCasingIsRfcCompliant). The concern is behavioral: unverified whether the Vert.x-applied header survives response.reset() (404) and the sendRedirect() commit (302). Run HstsHeaderTest, HstsError500Test, HstsRedirectHeaderTest, and HstsAuthHeaderTest to green before merge.
- HstsRedirectHeaderTest.java:14 - The new comment asserts Quarkus Vert.x headers guarantee HSTS on the 302, but sendRedirect() commits the response, so coverage depends on whether Quarkus sets the header before the servlet dispatch (survives) or at flush time (may not, since the redirect already committed). This test is the verification and was not run.
- HstsHeaderTest.java:14 - Missed by the comment cleanup. The class comment still reads that the HSTS header is set by a servlet filter so error responses that bypass the chain lose it, which is now false; line 51-52 carries a similar stale reference. Update for consistency with HstsAuthHeaderTest and HstsRedirectHeaderTest.
- Follow-up (not a blocker): once reset() survival is confirmed, X-Content-Type-Options could also move to quarkus.http.header, letting ContentTypeOptionsFilter and the resetKeepingHeaders preserve entry be removed and eliminating the last servlet/reset coupling. Confirm the disabled-API path first.

Ganesh-403:
@paoloantinori Thanks for the detailed review. You're right that the `response.reset()` path is the critical behavior to verify here, and I should not have treated the Vert.x behavior as proven without running the integration tests.

I'll run `HstsHeaderTest`, `HstsError500Test`, `HstsRedirectHeaderTest`, and `HstsAuthHeaderTest`, with particular attention to the disabled-API 404/reset path. I'll also update the remaining stale comments in `HstsHeaderTest`.

I'll update the PR based on the test results.

paoloantinori:
Sounds good. The disabled-API 404 path (`response.reset()` in `RegistryApplicationServletFilter`) is the one place where Quarkus-managed headers behave differently from the old servlet filter, so seeing `HstsHeaderTest`, `HstsError500Test`, `HstsRedirectHeaderTest`, and `HstsAuthHeaderTest` green on this head is what will close the loop for me. I'll re-read once you push.

Ganesh-403:
@paoloantinori Thanks for pointing out the `response.reset()` path.

I re-added `Strict-Transport-Security` to the existing `resetKeepingHeaders(...)` call so the Quarkus-managed header is captured before `reset()` and restored afterward.

I then ran the requested HSTS integration tests:

- `HstsHeaderTest`
- `HstsError500Test`
- `HstsRedirectHeaderTest`
- `HstsAuthHeaderTest`

Result: **12 tests run, 0 failures, 0 errors, 0 skipped — BUILD SUCCESS**.

This also verifies the two critical paths you called out: the disabled-API 404 that calls `response.reset()` and the 302 redirect path.

paoloantinori:
Re-read after the rebase. `RegistryApplicationServletFilter.resetKeepingHeaders` now captures `Strict-Transport-Security` and `X-Content-Type-Options` before `response.reset()` and restores them, which is exactly the gap I flagged on the disabled-API 404 path.

`HstsHeaderTest.testHstsOnDisabledApi404` exercises that path under `DisabledApisProfile` (which disables `/apis/registry/v3/system/.*`) and asserts both headers on the resulting 404, so the regression is locked in.

The redirect path is covered by `HstsRedirectHeaderTest` and the remaining error paths by `testHstsOnApiNotFoundError` / `testHstsOnUnknownPath404` / `testHstsOnBadRequest`, so the Quarkus-managed header reaches every response class the old filter protected. No further concerns from me — the migration looks clean.

Ganesh-403:
> /accept

didn't worked for some reason :)

EricWittmann:
Thanks for this contribution, @Ganesh-403! Nice cleanup replacing the custom filter with Quarkus-managed config. 🎉

Ganesh-403:
> Thanks for this contribution, @Ganesh-403! Nice cleanup replacing the custom filter with Quarkus-managed config. 🎉

My Pleasure :)

PR #9470, opened 2026-08-11
Fix #9449: Add validation for artifact version comment value

### LINKED ISSUE #9449: bug(rest): addArtifactVersionComment allows null value to reach database and return HTTP 500

anirudhagarwal-dev:
This PR resolves issue #9449, where submitting an artifact version comment with a missing or null value was allowing invalid data to reach the storage layer, resulting in an HTTP 500 Internal Server Error instead of an HTTP 400 Bad Request.

### Changes:
- Added `ParameterValidationUtils.requireParameter("value", data.getValue());` to `addArtifactVersionComment` in `v3/impl/GroupsResourceImpl.java`.
- Added the exact same parameter validation to `addArtifactVersionComment` in `v2/impl/GroupsResourceImpl.java` to ensure consistency.
- Added regression tests in `GroupsResourceTest.java` to verify that submitting an empty comment `{}` or a comment with a null value `{"value": null}` correctly returns a 400 Bad Request status code.

### Related Issues:
Fixes #9449


Hiten0305l:
Hi @anirudhagarwal-dev, thanks for working on this issue. I’ve actually already completed the implementation for this issue in my PR #9454, and the issue is already assigned to me.

I’ve kept #9454 as a Draft temporarily because my previous PR #9376 is still awaiting merge. Once #9376 is merged, I’ll move #9454 from Draft to Ready for Review.

So, I’d kindly request you not to proceed further with this issue for now, as the implementation is already completed on my side,to avoid duplication of work. Thanks for understanding!

carlesarnal:
Linked issue assigned to another issue and draft PR for a couple weeks. Closing.

PR #9471, opened 2026-08-11
fix(ui): prevent empty or whitespace input in numeric configuration fields

### LINKED ISSUE #9419: fix(ui): prevent empty or whitespace input in numeric configuration fields
## Description

Some numeric configuration input fields accept empty or whitespace-only values, which can lead to invalid configurations.

Examples:
- ""
- "   "

These should not be allowed.

## Expected behavior

Only valid numeric values should be accepted.

Invalid values:
- Empty string
- Whitespace-only string
- Non-numeric values

## Proposed fix

- Add validation to prevent empty/whitespace input
- Show error message: "Value cannot be empty"
- Disable Save button when invalid

jjayesh2:
## Description

Fixes #9419.

Some numeric configuration fields accepted empty or whitespace-only values.

### Changes

- Prevent empty and whitespace-only values.
- Validate numeric configuration values as integers.
- Display validation feedback for invalid values.
- Prevent saving invalid values using Enter.
- Keep the Save action disabled through the existing validation state.

### Validation

- Empty value → invalid
- Whitespace-only value → invalid
- Non-numeric value → invalid
- Decimal value → invalid
- Integer value → valid

### Testing

- `npx eslint src/app/pages/settings/components/PropertyInput.tsx`
- `git diff --check`


ui/ui-app/src/app/pages/settings/components/PropertyInput.tsx
```diff
@@ -24,35 +28,62 @@ export const PropertyInput: FunctionComponent<PropertyInputProps> = (props: Prop
...
+                return {
+                    isValid: false,
+                    errorMessage: "Value cannot be empty"
+                };
             }
+
             const num: number = Number(value);
-            return Number.isInteger(num);
+
+            return Number.isInteger(num)
```
paoloantinori: `Number.isInteger(Number(value))` accepts values like `1e3`, `0x10`, `0b11`, `+5`, and `1.0`. Since the component saves the raw string through `onSave` (not the parsed number), these can pass client validation yet fail server-side integer parsing. Since this PR's stated goal is integer validation, consider a strict check on the trimmed string (e.g. `/^-?\d+$/`) so client and server agree. Worth confirming the acceptable range too, if any of these fields must be non-negative.

ui/ui-app/src/app/pages/settings/components/PropertyInput.tsx
```diff
@@ -67,12 +98,18 @@ export const PropertyInput: FunctionComponent<PropertyInputProps> = (props: Prop
...
-            onChange={ handleInputChange }
-            onKeyDown={ handleKeyPress }
+        <TextInput name={props.name}
+            value={currentValue}
+            validated={validated()}
+            onChange={handleInputChange}
+            onKeyDown={handleKeyPress}
             aria-label="configuration property input"/>
+        {errorMessage && (
+            <HelperText>
```
paoloantinori: `HelperText` is placed inside `InputGroup`, which renders children in a horizontal row, so the error will likely show inline or squished next to the input rather than below the field. PatternFly convention is to render `HelperText` outside the `InputGroup` (as a following sibling, or within a `FormGroup`). Please verify the rendered layout and consider wiring it to the input via `aria-describedby` so screen readers announce the error.

ui/ui-app/src/app/pages/settings/components/PropertyInput.tsx
```diff
@@ -24,35 +28,62 @@ export const PropertyInput: FunctionComponent<PropertyInputProps> = (props: Prop
...
-        const isValid: boolean = validate(value);
-        setCurrentValue(value);
-        setIsDirty(value !== props.value);
-        setIsValid(isValid);
+    const validated = (): "success" | "warning" | "error" | "default" => {
+        return isValid ? "default" : "error";
     };
 
-    const validate = (value: string): boolean => {
+    const validate = (value: string): ValidationResult => {
```
paoloantinori: The new validation branches are only covered by `npx eslint`. Add unit tests asserting: empty invalid, whitespace-only invalid, non-numeric invalid, decimal invalid, integer valid, plus the loose-parsing cases above (`1e3`, `0x10`, `1.0`). This is the exact validation matrix listed in the PR description, so it is straightforward to encode.

paoloantinori (commented):
This PR improves validation feedback by adding an inline error message and a HelperText block, but the empty/whitespace rejection and integer validation it claims to introduce already existed in the pre-change `validate()`, so the only real behavioral additions are the error-message state and its UI. A few points are worth addressing before merge.

- The removed `validate()` already rejected empty and whitespace-only values for both `text` and `number` (`value.trim().length === 0` checks, re-added verbatim) and already enforced `Number.isInteger`. The description's "prevent empty/whitespace" and "validate as integers" bullets describe pre-existing behavior; it would help to state in the PR what #9419 actually required (likely the missing visual feedback).
- The integer check `Number.isInteger(Number(value))` accepts `1e3`, `0x10`, `0b11`, `+5`, and `1.0`. Because the component persists the raw string (`currentValue`) via `onSave` rather than the parsed number, these strings pass client validation but can still be rejected by server-side integer parsing.
- `HelperText` is rendered as a direct child of `InputGroup`, which lays out its children in a horizontal row, so the error will likely appear inline next to the input instead of below it. PatternFly convention is to place `HelperText` outside the `InputGroup` (as a following sibling, or inside a `FormGroup`).
- The `"Value cannot be empty"` literal is duplicated across the `text` and `number` branches; a shared constant would avoid drift.
- `ValidationResult` is declared inside the component body; hoisting it to module scope is cleaner.
- Testing was limited to `npx eslint` and `git diff --check`; the new validation branches (empty text, empty number, non-integer number) have no unit tests.

jjayesh2:
@EricWittmann  I pushed the requested reviewer changes (strict integer validation, validation utility extraction, and tests). The Verification Gate is currently reporting PR is not in a testable lifecycle state. Could you please advise if the PR needs a lifecycle transition or branch update?

EricWittmann:
Thanks for the contribution, @jjayesh2! Merged.

PR #9474, opened 2026-08-11
feat(mcp): add VersionState filtering to prompt template discovery (#9473)

### LINKED ISSUE #9473: [FEAT] Add Version State Filtering (ENABLED/DRAFT/DEPRECATED) to Prompt Template & Version Discovery Tools in apicurio-registry-mcp
### Summary & Motivation
In Apicurio Registry 3.x, the `mcp` module provides a Model Context Protocol (MCP) server that enables AI agents (such as Claude Desktop and Cursor) to discover, fetch, and execute prompt templates (`PROMPT_TEMPLATE`) and versioned artifacts stored in the registry.

Recently, PR #9375 introduced governance checks to prevent prompt template rendering endpoints from bypassing `DISABLED` version withdrawals. However, within the `apicurio-registry-mcp` module, `PromptTemplateMCPServer.list_mcp_prompts()` and `RegistryService.searchVersions()` still query versions without filtering by `VersionState`.

As a result:
- When an AI agent connects via MCP and calls `list_mcp_prompts()` or searches version artifacts, `DISABLED` and `DRAFT` versions are returned in the prompt discovery list.
- When the AI agent attempts to invoke `get_mcp_prompt()` or `get_version_content()` on a disabled version returned by `list_mcp_prompts()`, the underlying registry REST call throws an `InvalidVersionStateException` (404/400 error), causing a hard `ToolCallException` in Claude Desktop.

Adding explicit `VersionState` filtering to MCP discovery tools (defaulting to `ENABLED` state for prompt listings while exposing an optional state filter) aligns the MCP server with Apicurio Registry's v3 version governance lifecycle.

### Affected Code
- `mcp/src/main/java/io/apicurio/registry/mcp/servers/PromptTemplateMCPServer.java`
- `mcp/src/main/java/io/apicurio/registry/mcp/servers/VersionsMCPServer.java`
- `mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java`

### Proposed Technical Solution
1. Update `RegistryService.searchVersions` to accept `VersionState state`.
2. Update `PromptTemplateMCPServer.list_mcp_prompts()` to pass `VersionState.ENABLED` to `searchVersions`.
3. Add optional `@ToolArg` `versionState` parameter to `VersionsMCPServer.search_versions()`.

### Acceptance Criteria
- [ ] `list_mcp_prompts()` in `PromptTemplateMCPServer` filters for `ENABLED` state prompt templates by default.
- [ ] `RegistryService.searchVersions` passes `VersionState` filter to the Registry v3 client.
- [ ] `list_versions` MCP tool supports optional `versionState` argument.

Saksham842:
### Description
Extends PR #9375 version withdrawal governance to the MCP server layer. `PromptTemplateMCPServer.list_mcp_prompts()` and `RegistryService.searchVersions()` previously searched prompt versions without specifying a `VersionState` filter. This caused `DISABLED` (withdrawn) and `DRAFT` versions to be exposed in AI agent prompt discovery lists, leading to runtime `ToolCallException` when agents attempted to execute disabled prompt content.

This PR overloads `RegistryService.searchVersions` to support `VersionState` filtering, updates `list_mcp_prompts()` to filter `VersionState.ENABLED` by default, and exposes an optional `versionState` argument on the `search_versions` MCP tool.

### Related Issue
Fixes #9473

### Changes Made
- Overloaded `RegistryService.searchVersions` to accept `VersionState state`.
- Updated `PromptTemplateMCPServer.list_mcp_prompts()` to pass `VersionState.ENABLED` to `searchVersions`.
- Added `@ToolArg` `versionState` parameter to `VersionsMCPServer.search_versions()`.

### Verification
- Ran `./mvnw test-compile -pl mcp -am -DskipTests` (Passed)
- Ran `./mvnw checkstyle:check -pl mcp` (Passed with 0 violations)
- Ran `./mvnw test -pl mcp` (Passed all 17 tests)
- Verified DCO sign-off on commits.

### Contributor Checklist
- [x] Target branch is `main`
- [x] Followed Conventional Commits format
- [x] DCO sign-off included (`Signed-off-by`)
- [x] `mvn checkstyle:check -pl mcp` passes


mcp/src/main/java/io/apicurio/registry/mcp/servers/VersionsMCPServer.java
```diff
@@ -185,7 +186,8 @@ List<SearchedVersion> search_versions(
                 description,
                 jsonLabels,
                 order,
-                versionOrderBy
+                versionOrderBy,
+                versionState != null ? io.apicurio.registry.rest.client.models.VersionState.valueOf(versionState) : null
```
paoloantinori: `VersionState.valueOf(versionState)` throws `IllegalArgumentException` for any unrecognized or wrong-case input (e.g. "enabled", "Enabled", a typo). This is an AI-agent-facing `@ToolArg`, so untrusted free text flows straight into this conversion. Consider a case-insensitive parse (e.g. `Arrays.stream(VersionState.values()).filter(v -> v.name().equalsIgnoreCase(versionState)).findFirst()`) and, on miss, either return a clean validation error or pass null to fall back to no filter. That avoids surfacing a raw enum-constant stack trace through the tool response.

mcp/src/main/java/io/apicurio/registry/mcp/servers/VersionsMCPServer.java
```diff
@@ -174,6 +174,7 @@ List<SearchedVersion> search_versions(
             @ToolArg(description = SEARCH_NAME, required = false) String name,
             @ToolArg(description = SEARCH_DESCRIPTION, required = false) String description,
             @ToolArg(description = SEARCH_JSON_LABELS, required = false) String jsonLabels,
+            @ToolArg(description = VERSION_STATE, required = false) String versionState,
```
paoloantinori: The `VERSION_STATE` description should enumerate the accepted values so an agent can supply a valid one. Right now a caller has no way to discover that the arg must be one of `ENABLED`, `DISABLED`, `DRAFT` (and any others in the enum) and is likely to guess, which then hits the `valueOf` failure. Worth pairing this with the safe-parse change on line 190.

paoloantinori (commented):
Direction is correct: defaulting prompt discovery to `VersionState.ENABLED` stops withdrawn/draft versions from leaking into agent-facing lists, and the overload-delegation keeps the existing 8-arg call sites working. Two robustness gaps on the new `versionState` `@ToolArg` are worth fixing before merge, both centered on `VersionsMCPServer.java`.

- `RegistryService.java` overload-via-delegation (lines 310-339) is clean: the legacy signature forwards with `(VersionState) null`, and the new body correctly guards `r.queryParameters.state = state` behind a null check so existing unfiltered searches are unaffected. No regression.
- `PromptTemplateMCPServer.java:47-60` hardcoding `VersionState.ENABLED` is the right default for discovery; lifecycle-wise ENABLED is the only state agents should ever execute, so this is safe.
- `VersionsMCPServer.java:190` parses the free-text tool arg with `VersionState.valueOf(versionState)`, which throws `IllegalArgumentException` on any case mismatch or unrecognized value. For an AI-agent-facing input this produces a raw "No enum constant" error (even if caught by the MCP error handler, the message is unhelpful) instead of clean guidance.
- `VersionsMCPServer.java:177` exposes `versionState` with no enumeration of accepted values, so a calling agent has no way to know which strings are valid and will likely hit the failure above.

Saksham842:
Thank you for the thorough review, @paoloantinori!

Both issues are now fixed in the latest commit:

1. **Case-insensitive `VersionState` parse (line 190)**: Replaced `VersionState.valueOf(versionState)` with a safe `Arrays.stream(...).filter(v -> v.name().equalsIgnoreCase(versionState)).findFirst().orElse(null)` — invalid or wrong-case inputs now silently fall back to `null` (no filter) instead of surfacing a raw `IllegalArgumentException`.

2. **Enumerated accepted values in `@ToolArg` description (line 177)**: The `versionState` argument description now explicitly lists the accepted values: `ENABLED`, `DISABLED`, `DEPRECATED`, `DRAFT` (case-insensitive), so calling agents know exactly what strings are valid.

Please let me know if any further changes are needed!

paoloantinori:
Both points landed cleanly, thanks for the quick turnaround.

The `Arrays.stream(...).filter(...).equalsIgnoreCase(...).orElse(null)` parse is the right shape: an unknown or malformed `versionState` now silently means "no filter" rather than throwing, and it matches the documented contract in the `@ToolArg` description. The enumerated accepted-values string is clear for tool consumers.

No further concerns from me on this change.

Saksham842:
Thanks a lot, @paoloantinori! Appreciate the detailed review - learned a lot from the feedback.

paoloantinori:
Glad the feedback was useful, @Saksham842. The case-insensitive `VersionState` parse and the overload-delegation landed cleanly, and the default `ENABLED` filter on `list_mcp_prompts` is the right call. Nothing further from my side.

PR #9481, opened 2026-08-11
feat(rest): support content downloads by contentId, globalId, and contentHash

### LINKED ISSUE #9472: feat(rest): Support content downloads by contentId, globalId, and contentHash in DownloadsResource

Currently, the single-use downloads endpoint (`/apis/registry/v3/downloads/{downloadId}`) in `DownloadsResourceImpl.java` only supports data export (`DownloadContextType.EXPORT`) and version ZIP archive exports (`DownloadContextType.VERSION_EXPORT`).

However, `DownloadContextType` already defines enum values and `DownloadContextDto` includes fields for retrieving raw schema/artifact content via:
* `CONTENT_BY_CONTENT_ID` (`contentId`)
* `CONTENT_BY_GLOBAL_ID` (`globalId`)
* `CONTENT_BY_CONTENT_HASH` (`contentHash`)

In `DownloadsResourceImpl.java`, line 58 currently contains a `// TODO support other types of downloads (e.g. download content by contentId)`. Attempting to consume a download context for raw content currently throws a `DownloadNotFoundException`.

### Proposed Solution

Implement support for downloading artifact and schema content by `contentId`, `globalId`, and `contentHash` in `DownloadsResourceImpl.java`:

1. Inject `IdsResource` into `DownloadsResourceImpl`.
2. Handle `DownloadContextType.CONTENT_BY_CONTENT_ID` by delegating to `idsResource.getContentById(downloadContext.getContentId())`.
3. Handle `DownloadContextType.CONTENT_BY_GLOBAL_ID` by delegating to `idsResource.getContentByGlobalId(downloadContext.getGlobalId(), null, null)`.
4. Handle `DownloadContextType.CONTENT_BY_CONTENT_HASH` by delegating to `idsResource.getContentByHash(downloadContext.getContentHash())`.
5. Add REST unit test coverage in `DownloadsResourceTest.java` verifying creation and consumption of single-use download URLs for content IDs, global IDs, and content hashes.

This benefits Apicurio Registry clients (SDKs, CLI, and browser downloads) by allowing secure, single-use content download URLs for specific schema versions and hashes without requiring credentials to be passed in query parameters.

### Additional Context

- Target file: `app/src/main/java/io/apicurio/registry/rest/v3/impl/DownloadsResourceImpl.java`
- Leverages existing HTTP caching (`@EntityIdContentCache`), `Content-Disposition` header generation, and content-type detection provided by `IdsResource`.

Devansh-18155:
### Description
Fixes #9472 

Implemented support for single-use content downloads by `contentId`, `globalId`, and `contentHash` in `DownloadsResourceImpl.java`.

### Changes Made
- Injected `IdsResource` into `DownloadsResourceImpl`.
- Implemented delegation to `getContentById`, `getContentByGlobalId`, and `getContentByHash` for `CONTENT_BY_CONTENT_ID`, `CONTENT_BY_GLOBAL_ID`, and `CONTENT_BY_CONTENT_HASH` download context types.
- Added REST unit test suite `DownloadsResourceTest.java` verifying 200 payload responses, Content-Disposition headers, and single-use 404 expiration behavior.

### Files Changed
- **`app/src/main/java/io/apicurio/registry/rest/v3/impl/DownloadsResourceImpl.java`**
  - Injected `IdsResource idsResource`.
  - Added conditional delegation blocks for `CONTENT_BY_CONTENT_ID`, `CONTENT_BY_GLOBAL_ID`, and `CONTENT_BY_CONTENT_HASH`.
- **`app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DownloadsResourceTest.java`**
  - Added new REST unit test suite covering single-use download URL creation via `storage.createDownload()` and consumption via `GET /apis/registry/v3/downloads/{downloadId}`.
  - Verified HTTP 200 payload responses, `Content-Disposition` headers, and single-use 404 security expiration on re-consumption.

### Verification & Test Results
1. **Quarkus REST Unit Tests:** Executed `./mvnw test -pl app -Dtest=DownloadsResourceTest`
   - **Result:** `Tests run: 3, Failures: 0, Errors: 0, Skipped: 0` (`BUILD SUCCESS`)
2. **Code Style Check:** Executed `./mvnw checkstyle:check -pl app`
   - **Result:** `0 Checkstyle violations` (`BUILD SUCCESS`)

### DIFF
--- a/app/src/main/java/io/apicurio/registry/rest/v3/impl/DownloadsResourceImpl.java
+++ b/app/src/main/java/io/apicurio/registry/rest/v3/impl/DownloadsResourceImpl.java
@@ -6,6 +6,7 @@ import io.apicurio.registry.auth.AuthorizedStyle;
 import io.apicurio.registry.logging.Logged;
 import io.apicurio.registry.metrics.health.liveness.ResponseErrorLivenessCheck;
 import io.apicurio.registry.metrics.health.readiness.ResponseTimeoutReadinessCheck;
+import io.apicurio.registry.rest.v3.IdsResource;
 import io.apicurio.registry.rest.v3.impl.shared.DataExporter;
 import io.apicurio.registry.rest.v3.impl.shared.ProtobufExporter;
 import io.apicurio.registry.storage.RegistryStorage;
@@ -38,6 +39,9 @@ public class DownloadsResourceImpl {
     @Inject
     ProtobufExporter protobufExporter;
 
+    @Inject
+    IdsResource idsResource;
+
     @Authorized(style = AuthorizedStyle.None, level = AuthorizedLevel.None)
     @GET
     @Path("{downloadId}")
@@ -55,7 +59,17 @@ public class DownloadsResourceImpl {
                     downloadContext.getVersion());
         }
 
-        // TODO support other types of downloads (e.g. download content by contentId)
+        if (downloadContext.getType() == DownloadContextType.CONTENT_BY_CONTENT_ID) {
+            return idsResource.getContentById(downloadContext.getContentId());
+        }
+
+        if (downloadContext.getType() == DownloadContextType.CONTENT_BY_GLOBAL_ID) {
+            return idsResource.getContentByGlobalId(downloadContext.getGlobalId(), null, null);
+        }
+
+        if (downloadContext.getType() == DownloadContextType.CONTENT_BY_CONTENT_HASH) {
+            return idsResource.getContentByHash(downloadContext.getContentHash());
+        }
 
         throw new DownloadNotFoundException();
     }
--- a/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DownloadsResourceTest.java
+++ b/app/src/test/java/io/apicurio/registry/noprofile/rest/v3/DownloadsResourceTest.java
@@ -94,6 +94,10 @@ public class DownloadsResourceTest extends AbstractResourceTestBase {
 
         Assertions.assertTrue(content.contains("\"openapi\": \"3.0.2\""));
         Assertions.assertTrue(content.contains(title));
+
+        // Attempting to consume the single-use download link again should return 404
+        given().when().pathParam("downloadId", downloadId)
+                .get("/registry/v3/downloads/{downloadId}").then().statusCode(404);
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


Devansh-18155:
hi @paoloantinori , Can you please review my changes once. Thanks!

paoloantinori:
Thanks for the ping, @Devansh-18155. The delegation looks clean and the three new paths are well covered by the unit suite. Two notes worth a look before this merges:

**Authz on the delegated calls holds up.** I traced the model: `@Authorized` is a CDI `@InterceptorBinding` backed by `AuthorizedInterceptor`, `IdsResourceImpl` is `@ApplicationScoped`, and you inject it via `@Inject IdsResource idsResource`, so the call goes through the CDI proxy and re-triggers the interceptor on every delegation. `getContentById`/`getContentByHash` are `AuthorizedStyle.None / AuthorizedLevel.Read` and `getContentByGlobalId` is `GlobalId / Read`, so when auth is enabled each path still enforces. The `None/None` on the outer `download()` is fine because the single-use `downloadId` is the credential (same model as the existing `EXPORT`/`VERSION_EXPORT` paths). No bypass introduced.

**Single-use consumption is only asserted for the `CONTENT_ID` path.** `testDownloadByContentId` verifies the second `GET` returns 404, but `testDownloadByGlobalId` and `testDownloadByContentHash` stop at the first 200. Since `consumeDownload` deletes the context regardless of type, the behavior should be identical, but the test only pins one of the three branches. Adding the same re-consume assertion to the other two would lock it down against a future regression where a new branch forgets to call `consumeDownload`.

Minor: the `Content-Disposition` filename is hardcoded as `<id>.json` for all three paths (`contentId + ".json"`, `globalId + ".json"`, `contentHash + ".json"`). For an AVRO or Protobuf artifact the saved filename would still say `.json`, which is wrong for the content type. Worth deriving the extension from the artifact content type, or at least not assuming JSON.

Overall a solid, focused addition — the authz preservation and the test coverage on the new branches are the main things.

JituRewar (commented):
@Devansh-18155  The single-use behavior is asserted only in `testDownloadByContentId()`. Could we also verify the second request returns 404 for the globalId and contentHash cases? This would ensure all three newly supported download types preserve the single-use contract.

Devansh-18155:
hi @paoloantinori , Thanks for the thorough review! I have updated the unit test suite (`DownloadsResourceTest.java`) based on your feedback:

1. **Re-consumption 404 Assertions:** Added second `GET` re-consumption 404 assertions to `testDownloadByGlobalId` and `testDownloadByContentHash` to lock down single-use context deletion across all three branches.
2. **Dynamic Extension Verification:** Added `testDownloadXmlArtifact()` using a WSDL/XML artifact (`ContentTypes.APPLICATION_XML`) to explicitly verify that the `Content-Disposition` header filename dynamically derives `.xml` from the content type rather than assuming `.json`.

Pushed the update commit!

Devansh-18155:
hi @JituRewar , Good catch! I just pushed a commit adding the second `GET` 404 re-consumption assertions to both `testDownloadByGlobalId()` and `testDownloadByContentHash()`. All 4 tests are now passing and verify the single-use contract across all branches.

paoloantinori:
Thanks @Devansh-18155 — re-read the new test additions. Both points are addressed:

- The second-GET 404 assertions on `testDownloadByGlobalId` and `testDownloadByContentHash` lock down the single-use contract across all three branches now, matching what `testDownloadByContentId` already had. That's the gap closed.
- `testDownloadXmlArtifact` confirms the `.xml` extension is derived from the content type rather than hardcoded, which answers the filename concern for non-JSON artifacts.

From my side this is in good shape. The authz preservation on the delegated calls still holds (the points from the earlier read are unchanged in production), and the test coverage on the new paths is now consistent. No further comments from me.

JituRewar (approved):
Great @Devansh-18155 , through your new commits my point was addressed, LGTM, no further concern from my side.

Devansh-18155:
hi @EricWittmann , Can you please review my changes in the pr when you get the time . Thanks!

EricWittmann:
Please see my comments on the issue.  I'm looking for opinions on what to do.  Should we merge this?

Devansh-18155:
> Please see my comments on the issue. I'm looking for opinions on what to do. Should we merge this?

Thanks , @EricWittmann! That makes total sense. 

Browsers can't easily pass auth headers on direct download links (`<a href="...">`), so generating a temporary `downloadId` via an authenticated call first is definitely the right model. Right now we built the consumer side (`/downloads/{id}`), but as you pointed out, there isn't an authenticated path to actually create those `CONTENT_BY_*` records in step (2).

I'd like to implement step (2) so we can complete this feature end-to-end.

What do you think is the best API design for exposing that? A couple ideas:
1. Adding a `forBrowser=true` query param or `/download` sub-path to the existing endpoints in `IdsResource` (like `GET /ids/globalIds/{globalId}?forBrowser=true` or `/ids/globalIds/{globalId}/download`), which would return a `DownloadRef` containing the temporary `href`.
2. A dedicated endpoint or query param on `GroupsResource` for artifact versions.

Let me know what design you prefer and I'll extend the PR to cover the producer side as well.

PR #9482, opened 2026-08-11
Fixed the bug where the Editor is loosing the unsaved changes

### LINKED ISSUE #9478: Source editor loses unsaved changes when switching JSON ↔ YAML
## Description

**Registry Version:** snapshot  
**Persistence type:** in-memory

When editing an OpenAPI document in the **Source** editor, unsaved changes are lost when switching between YAML and JSON formats.

For example, if the original document contains:

```yaml
description: A dummy API for testing OpenAPI validation
```

and I make an unsaved change to:

```yaml
description: A dummy API for testing OpenAPI validationsssss
```

the change is correctly visible in the Source editor.

However, after switching from YAML to JSON and then back to YAML, the value reverts to the previously loaded/saved value:

```yaml
description: A dummy API for testing OpenAPI validation
```

The issue appears to be related to the source-format conversion using the underlying API model instead of the current unsaved source content.

In `SourceFormComponent.toggleSourceFormat()`, the current editor contents are first parsed into `parsedSource`, but when converting back to YAML, the implementation uses `this.sourceJs()`:

```typescript
newSource = YAML.dump(this.sourceJs(), {
    indent: 4,
    lineWidth: 110,
    noRefs: true
});
```

instead of the already parsed current source:

```typescript
newSource = YAML.dump(parsedSource, {
    indent: 4,
    lineWidth: 110,
    noRefs: true
});
```

This causes unsaved changes made in Source mode to be discarded during the format conversion.

## Environment

- **Apicurio Registry:** running from `main` / snapshot build
- **Persistence:** in-memory
- **Browser:** Microsoft Edge / Chromium-based browser
- **Registry UI:** running locally
- **OpenAPI editor:** loaded through the embedded editor/iframe architecture

## Steps to Reproduce

1. Open an OpenAPI document in Apicurio Registry.
2. Navigate to the **Source** tab.
3. Make an unsaved change to the OpenAPI YAML source.

   For example, change:

   ```yaml
   description: A dummy API for testing OpenAPI validation
   ```

   to:

   ```yaml
   description: A dummy API for testing OpenAPI validationsssss
   ```

4. **Do not click Save.**
5. Click **As JSON**.
6. Verify that the modified value (`validationsssss`) is present in the JSON source.
7. Click **As YAML**.
8. Observe that the modified value is lost and the previously loaded value is restored.

### Expected

Switching between YAML and JSON should only change the representation of the **current Source editor content**.

An unsaved change should survive the format conversion:

```text
YAML
  validationsssss
      ↓
JSON
  validationsssss
      ↓
YAML
  validationsssss
```

### Actual

The unsaved change is lost after switching formats:

```text
YAML
  validationsssss
      ↓
JSON
  validationsssss
      ↓
YAML
  validation
```

## Logs

No Registry server-side error is observed.

The issue can be reproduced entirely through the UI.

Browser console logs show the editor receiving/loading the editing information and the OpenAPI editor being initialized successfully. The problem occurs during the Source format conversion.

The relevant code path is `SourceFormComponent.toggleSourceFormat()`.

I also verified that the modified value is present immediately before and after the YAML → JSON conversion, but is lost when converting from JSON → YAML.

avadhutcodes:
Fixes #9478 

When editing an OpenAPI document in the Source editor, unsaved changes were lost when switching between YAML and JSON format

The bug was:

1) User edits YAML → unsaved change is stored in parsedSource.
2) YAML/JSON conversion ignored parsedSource.
3)  When I traced till end Found that the  this.sourceJs() in source-form.base.ts , which contained the old saved value is being dumped in YAML hence we loose our unsaved changes while switching the format between YAML to JSON.
hence the unsaved change disappeared.

### DIFF
--- a/ui/ui-editors/src/app/editor/_components/forms/source-form.base.ts
+++ b/ui/ui-editors/src/app/editor/_components/forms/source-form.base.ts
@@ -186,7 +186,7 @@ export abstract class SourceFormComponent<T extends Node> extends AbstractBaseCo
             if (this.isSourceFormatJson()) {
                 newSource = JSON.stringify(parsedSource, null, 4);
             } else {
-                newSource = YAML.dump(this.sourceJs(), {
+                newSource = YAML.dump(parsedSource, {
                     indent: 4,
                     lineWidth: 110,
                     noRefs: true
@@ -215,7 +215,7 @@ export abstract class SourceFormComponent<T extends Node> extends AbstractBaseCo
             if (this.isSourceFormatJson()) {
                 newSource = JSON.stringify(parsedSource, null, 4);
             } else {
-                newSource = YAML.dump(this.sourceJs(), {
+                newSource = YAML.dump(parsedSource, {
                     indent: 4,
                     lineWidth: 110,
                     noRefs: true


vjkumar2756 (commented):
could you add a regression test that explicitly verifies an unsaved change is preserved when switching between YAML and JSON
ideally covering both YAML → JSON and JSON → YAML?
This would make sure the behavior is protected against future regressions.

avadhutcodes:
I’ve got the test flow working up to the OpenAPI editor, including creating the artifact/draft and accessing the editor inside the iframe. However, I’m currently blocked on reliably modifying the Ace editor content through Playwright. The editor is an Angular ace-editor component inside the iframe, and although Playwright can access the rendered content, the different approaches I tried aren't updating the editor's internal state correctly.
I’ve spent quite a bit of time trying to get the E2E test working, and I don't want to keep going down a potentially fragile path without understanding how this editor is intended to be tested.
Would you recommend testing this regression at the OpenAPI editor/component level instead of through the full Registry Playwright flow? If there is an existing test/helper for interacting with this editor, I'd really appreciate being pointed in the right direction.

paoloantinori:
Thanks for the status update. Since the fix itself lives in `source-form.base.ts` (the `YAML.dump(parsedSource, ...)` change), you may not need a full Playwright E2E through the Ace iframe at all — a component-level test that drives the source-format switch and asserts `parsedSource` survives the conversion would cover the regression more directly and avoid the iframe/ACE flakiness entirely.

If you do want to keep the E2E path, the usual trick with ACE inside an iframe is to bypass typing and set the value programmatically, e.g. locate the iframe, then evaluate something like `frame.evaluate(() => { const ed = (window as any).ace.edit(document.querySelector(".ace_editor")); ed.setValue(newContent, 1); })`, then dispatch the change event so the Angular binding picks it up. Typing into ACE via keyboard events tends to be the unreliable part.

Happy to re-read once the test lands.

avadhutcodes:
I checked the existing test configuration more closely. ui/ui-editors/tsconfig.spec.json is configured for Jasmine and references:

"files": [
  "src/test.ts",
  "src/polyfills.ts"
]

src/polyfills.ts exists and imports zone.js, but src/test.ts does not exist. There are also no Jasmine/Karma dependencies or karma.conf.js in ui/ui-editors.
There are currently no existing unit tests in this package.
For the SourceFormComponent.toggleSourceFormat() regression, should I add the missing Jasmine/Karma test infrastructure and a component-level test, or should I use the existing Playwright setup?

paoloantinori:
Good catch — that tsconfig.spec.json is vestigial. I checked: ui/ui-editors/package.json has no Karma/Jasmine devDependencies, no karma.conf.js, and no src/test.ts, so there is indeed no unit-test runner wired up in that package (and no existing unit tests anywhere in it).

Given that, I'd go with the existing Playwright setup rather than bootstrapping a whole Karma/Jasmine stack for one regression test. Adding test infra to a package that has never had it is a chunk of maintenance surface (runner config, CI wiring, headless chrome in the build image) that goes well beyond this fix — and if the maintainers later want unit tests in ui-editors, that's better done as its own dedicated PR.

So: ui/tests is the right home. A spec that opens the OpenAPI editor, sets unsaved content via the programmatic ACE setValue approach from my earlier comment, toggles the source format, and asserts the unsaved content survives (both directions if feasible) would cover the regression with the tools the repo already runs in CI.

EricWittmann:
This change is simple and targeted, fixing a bug.  No test is needed.  Thanks @avadhutcodes :+1:

PR #9484, opened 2026-08-11
feat(ui): add direct path to create artifact from empty search page

### LINKED ISSUE #9475: [UI] Consider providing a direct path to create an artifact from the empty Search page

When the registry contains no artifacts, the **Search** page displays a "No artifacts found" empty state. However, there is no obvious next step for users who want to create their first artifact.

As a first-time user, I expected to be able to continue my workflow from this page. Instead, I had to navigate to another page (such as the Dashboard) to find the **Create Artifact** action before returning to the Search page.

From a user perspective, this feels like an unnecessary navigation step during the initial experience.

### Proposed Solution

Would it make sense for the empty state to provide a more direct path to creating the first artifact?

I'm intentionally leaving the implementation open, as there may be multiple approaches that fit the existing UI and design guidelines. My main goal is to improve the first-time user experience and reduce unnecessary navigation.

### Screenshot

Current empty state on the Search page when no artifacts exist:

Image

### Are you able to raise a PR?

**Yes.** If the maintainers agree that this would be a useful UX improvement, I'd be happy to work on a pull request.

### Additional Context

This is primarily a first-time user experience observation. I wanted to check whether the current navigation flow is intentional before proposing a specific implementation.

kunaldevsahu:
## What does this PR do?

When there are no artifacts, the Search page currently shows the empty state but doesn't give the user a direct way to create one.

This PR adds a more direct way to start creating an artifact from the empty state, so users don't have to go back to the Dashboard first.

## Before

Screenshot 2026-08-11 at 6 45 52 PM

## After

Screenshot 2026-08-11 at 8 09 16 PM

## Testing

- Tested the empty Search page.
- Tested opening the Create Artifact flow.
- Verified the existing artifact creation flow still works.

## Related Issue

Fixes #9475

### DIFF
--- a/ui/ui-app/src/app/pages/search/SearchPage.tsx
+++ b/ui/ui-app/src/app/pages/search/SearchPage.tsx
@@ -18,7 +18,7 @@ import { If, ListWithToolbar } from "@apitomy/common-ui-components";
 import { SearchType } from "@app/pages/search/SearchType.ts";
 import { Paging } from "@models/Paging.ts";
 import { FilterBy, SearchFilter, useSearchService } from "@services/useSearchService.ts";
-import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
+import { useGroupsService } from "@services/useGroupsService.ts";
 import {
     ArtifactSearchResults,
     ArtifactSortBy,
@@ -264,7 +264,11 @@ export const SearchPage: FunctionComponent<PageProperties> = () => {
         <SearchPageEmptyState
             searchType={searchType}
             isFiltered={isFiltered()}
-            onCreateArtifact={() => setIsCreateArtifactModalOpen(true)}/>
+            onCreateArtifact={
+                searchType === SearchType.ARTIFACT
+                    ? () => setIsCreateArtifactModalOpen(true)
+                    : undefined
+            }/>
     );
 
     return (
--- a/ui/ui-app/src/app/pages/search/SearchPage.tsx
+++ b/ui/ui-app/src/app/pages/search/SearchPage.tsx
@@ -13,15 +13,17 @@ import {
     SearchPageToolbar,
     toPageError, SearchGroupList, SearchVersionList
 } from "@app/pages";
-import { RootPageHeader } from "@app/components";
+import { CreateArtifactModal, RootPageHeader } from "@app/components";
 import { If, ListWithToolbar } from "@apitomy/common-ui-components";
 import { SearchType } from "@app/pages/search/SearchType.ts";
 import { Paging } from "@models/Paging.ts";
 import { FilterBy, SearchFilter, useSearchService } from "@services/useSearchService.ts";
+import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
 import {
     ArtifactSearchResults,
     ArtifactSortBy,
     ArtifactSortByObject,
+    CreateArtifact,
     GroupSearchResults,
     GroupSortBy,
     GroupSortByObject, SearchedGroup, SearchedVersion,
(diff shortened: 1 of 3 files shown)


ui/ui-app/src/app/pages/search/SearchPage.tsx
```diff
@@ -246,7 +263,8 @@ export const SearchPage: FunctionComponent<PageProperties> = () => {
     const emptyState = (
         <SearchPageEmptyState
             searchType={searchType}
-            isFiltered={isFiltered()}/>
+            isFiltered={isFiltered()}
+            onCreateArtifact={() => setIsCreateArtifactModalOpen(true)}/>
```
paoloantinori: This handler is passed for every search type, so the Create button also shows on empty GROUP and VERSION searches. On an empty groups search the body copy reads "Create one or more groups to view them here" right above a "Create artifact" button, which is contradictory. Gate it on `searchType === SearchType.ARTIFACT`, e.g. `onCreateArtifact={searchType === SearchType.ARTIFACT ? () => setIsCreateArtifactModalOpen(true) : undefined}`, so the button only appears where the copy matches.

paoloantinori (commented):
The change is correctly wired and mirrors the existing Dashboard create-artifact flow, but one issue is worth fixing before merge: the Create button renders for empty GROUP and VERSION searches, not just ARTIFACT. A couple of smaller cleanup notes below.

- `SearchPage.tsx:267` passes `onCreateArtifact` unconditionally, so on an empty groups search the body copy says "create one or more groups to view them here" directly above a "Create artifact" button. Gate the handler on `searchType === SearchType.ARTIFACT`.
- `SearchPage.tsx:96` (`doCreateArtifact`) is now the fourth near-identical copy of the create-then-navigate handler (DashboardPage, GroupPage, AgentsPage). Worth extracting a shared helper or `useCreateArtifact` hook so the navigation and error path lives in one place.
- `SearchPage.tsx:22` imports `GroupsService` (the type) but never uses it; `const groups = useGroupsService()` has no annotation. Either drop it from the import or annotate `const groups: GroupsService = ...` to match `DashboardPage.tsx:58`.
- No test for the new empty-state create path is included in the diff. If this component has test coverage elsewhere, add an assertion that the Create button opens the modal and that the handler navigates on a successful create.

ui/ui-app/src/app/pages/search/SearchPage.tsx
```diff
@@ -85,9 +87,24 @@ export const SearchPage: FunctionComponent<PageProperties> = () => {
     const [isSearching, setSearching] = useState<boolean>(false);
     const [paging, setPaging] = useState<Paging>(DEFAULT_PAGING);
     const [results, setResults] = useState<ArtifactSearchResults | GroupSearchResults | VersionSearchResults>(EMPTY_RESULTS);
+    const [isCreateArtifactModalOpen, setIsCreateArtifactModalOpen] = useState<boolean>(false);
 
     const searchSvc = useSearchService();
     const appNav = useAppNavigation();
+    const groups = useGroupsService();
+
+    const doCreateArtifact = (groupId: string | undefined, data: CreateArtifact): void => {
```
paoloantinori, line 96: This `doCreateArtifact` is a near-exact copy of the handler in `DashboardPage.tsx` (lines 132-143), with siblings in `GroupPage.tsx` and `AgentsPage.tsx`. The encode-then-navigate plus `toPageError` logic is now duplicated four times. Consider extracting a shared helper (a `useCreateArtifact()` hook or a `createArtifactAndNavigate(...)` util) and routing all four pages through it.

ui/ui-app/src/app/pages/search/SearchPage.tsx
```diff
@@ -13,15 +13,17 @@ import {
...
     toPageError, SearchGroupList, SearchVersionList
 } from "@app/pages";
-import { RootPageHeader } from "@app/components";
+import { CreateArtifactModal, RootPageHeader } from "@app/components";
 import { If, ListWithToolbar } from "@apicurio/common-ui-components";
 import { SearchType } from "@app/pages/search/SearchType.ts";
 import { Paging } from "@models/Paging.ts";
 import { FilterBy, SearchFilter, useSearchService } from "@services/useSearchService.ts";
+import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
 import {
```
paoloantinori, line 22: `GroupsService` (the type) is imported but never referenced. `const groups = useGroupsService()` has no type annotation. Either remove `GroupsService` from the import, or annotate it as `const groups: GroupsService = useGroupsService()` to match `DashboardPage.tsx:58`.

kunaldevsahu:
Thanks @paoloantinori  for the review! I've addressed the requested changes:

* Made the Create Artifact action available only for empty Artifact searches.
* Removed the unused `GroupsService` import.
* Added E2E coverage for the empty Artifact, Group, and Version search states.

All checks are passing now. I left the shared `doCreateArtifact` refactor out to keep this PR focused.

paoloantinori:
Thanks for the iteration. The `GroupsService` import is actually used now (`groups.createArtifact` in `doCreateArtifact`), so that one is fine. But the scoping change did not land at this head: `SearchPage.tsx` still passes `onCreateArtifact` to `SearchPageEmptyState` unconditionally (around line 268), and `SearchPageEmptyState` gates the button only on `!!props.onCreateArtifact` with no `searchType` check. So the Create button still renders on empty GROUP and VERSION searches. To match what you described, either pass the callback only when `searchType === SearchType.ARTIFACT`, or add `&& props.searchType === SearchType.ARTIFACT` to the `If` condition in `SearchPageEmptyState`. Happy to re-read once pushed.

kunaldevsahu:
Hey @EricWittmann, the PR is ready now. I’ve addressed the feedback and pushed the latest changes. Whenever you get a chance, could you take another look? Thanks!

EricWittmann (dismissed):
Looks good to me!

EricWittmann:
It looks like we have some UI test failures.  Can you take a look?

kunaldevsahu:
Sure @EricWittmann , I’ll take a look and fix the UI test failures.

kunaldevsahu:
I’ve resolved the UI test issue and pushed the fix. I’ve also verified the updated test locally. When you have time, please have a look.

PR #9486, opened 2026-08-11
fix(mcp): use configurable paging limit in searchAgentCards and searchMcpTools

### LINKED ISSUE #9485: bug(mcp): searchMcpTools and searchAgentCards hardcode limit=50, ignoring paging config
## Description

**Registry
Version**: snapshot
**Persistence type**: in-memory / sql

While reviewing `RegistryService.java` in the `mcp` module, I noticed that `searchAgentCards` and `searchMcpTools` hardcode the query limit to `50` instead of using `McpConfig.paging().limit()`.
This creates two issues:
1. **Config mismatch:** `apicurio.mcp.paging.limit` defaults to `200` in `McpConfig.java`, but these two methods hardcode `50`. A user who never changes configuration gets a tighter, silent limit than the default system setting.
2. **Broken paging contract:** Unlike all other search methods in `RegistryService` (`searchArtifacts`, `searchGroups`, `searchVersions`), neither method calls `checkPagingLimit(...)`. Every other endpoint raises a `ToolCallException` when items exceed the limit to notify the AI client, but here excess results are silently dropped without error.

### Environment
- Registry Version: main (snapshot)
- Module: `apicurio-registry-mcp`

### Steps to Reproduce
1. Open `mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java`.
2. Inspect `searchAgentCards` (around line 364) and `searchMcpTools` (around line 389):
   ```java
   public String searchAgentCards(String name, String skill, String capability) throws Exception {
       var results = client().wellKnown().agents().get(config -> {
           config.queryParameters.limit = 50; // hardcoded limit
           ...
       });
       return utils.toPrettyJson(results); // checkPagingLimit is not called
   }
3.Set apicurio.mcp.paging.limit=10 in configuration.
4.Execute an MCP tool search (search_mcp_tools or discover_agents) when more than 10 artifacts exist — the configured limit is ignored, and results up to 50 are returned without throwing the configured paging error.
<!--
Add the steps required to reproduce this error.

1. Go to `XX >> YY >> SS`
2. Create a new item `N` with the info `X`
3. See error

If the error cannot be easily reproduced, make sure to fill in the Environment and Logs sections.
You can also try to reproduce the bug with a fresh Apicurio Registry instance.
-->

Expected: searchMcpTools and searchAgentCards should set limit = config.paging().limit() + 1 and call checkPagingLimit(...), honoring McpConfig defaults and maintaining parity with searchArtifacts / searchVersions.

Actual: Both methods hardcode limit = 50, ignore McpConfig (which defaults to 200), and silently truncate results without calling checkPagingLimit.
<!-- A clear and concise description of what you expected to happen and what actually happened. -->

### Logs
No exception is logged because the call succeeds silently with truncated results.
<!-- Paste or link the logs from Registry, browser console, your command line or other relevant places. Redact if needed.-->

princexpoddar:
## Summary
Fixes #9485 
Fixes hardcoded paging limits in MCP `searchAgentCards` and `searchMcpTools` methods to respect configured MCP paging limits.

## Root Cause
`searchAgentCards` and `searchMcpTools` in `RegistryService.java` hardcoded `limit = 50` in their query parameters rather than using `config.paging().limit() + 1`. Additionally, `checkPagingLimit(results.getCount())` was omitted for both methods, causing paging overflow detection to be bypassed.

## Changes
- **`io.apicurio.registry.mcp.RegistryService`**:
  - `searchAgentCards(...)`: Updated query parameters limit to `config.paging().limit() + 1` and added call to `checkPagingLimit(results.getCount())`.
  - `searchMcpTools(...)`: Updated query parameters limit to `config.paging().limit() + 1` and added call to `checkPagingLimit(results.getCount())`.
- **`io.apicurio.registry.mcp.RegistryServiceTest`**:
  - `testQueryParameterPropagation()`: Updated `lastUri` query limit assertions from `limit=50` to `limit=201` (default paging limit of 200 + 1).

## Test plan
- [x] Specific tests added or modified: Updated `RegistryServiceTest.testQueryParameterPropagation()` to verify limit propagation.
- [x] Integration test results: Ran full `mcp` module test suite via `..\mvnw.cmd test`
- [x] Code style verification: Checkstyle passed.


mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
```diff
@@ -362,18 +362,19 @@ public ConfigurationProperty getConfigurationProperty(String propertyName) {
...
             if (skill != null && !skill.isBlank()) {
-                config.queryParameters.skill = new String[]{ skill };
+                r.queryParameters.skill = new String[]{ skill };
             }
             if (capability != null && !capability.isBlank()) {
-                config.queryParameters.capability = new String[]{ capability };
+                r.queryParameters.capability = new String[]{ capability };
             }
         });
+        checkPagingLimit(results.getCount());
```
paoloantinori, line 402: This newly-added `checkPagingLimit(results.getCount())` changes the method's behavior: it now fails when the result count exceeds the configured paging limit. No test in this change exercises that throw path. Consider a test that returns more than `paging().limit()` results from the stubbed client and asserts the expected exception is thrown here and in `searchMcpTools`.

mcp/src/test/java/io/apicurio/registry/mcp/RegistryServiceTest.java
```diff
@@ -322,13 +322,13 @@ public void testOAuth2HeaderAttachment() throws Exception {
     public void testQueryParameterPropagation() throws Exception {
         RegistryService service = createService("http://localhost:" + port, false);
         service.searchAgentCards("abc", "java", "streaming");
-        assertTrue(lastUri.contains("limit=50"));
+        assertTrue(lastUri.contains("limit=201"));
```
paoloantinori: Hardcoding `201` bakes the default paging limit (200) into the assertion. Derive the expected value from the service config (`config.paging().limit() + 1`) so the test tracks the contract rather than a magic number that will fail opaquely if the default changes.

paoloantinori (commented):
The core fix is sound: replacing the hardcoded `limit = 50` with `config.paging().limit() + 1` and adding `checkPagingLimit(results.getCount())` aligns `searchAgentCards` and `searchMcpTools` with the overflow-detection pattern used by the other search methods. The lambda rename from `config` to `r` is required, not cosmetic: it lets `config.paging().limit()` resolve to the `config` field instead of the lambda parameter. The main gap is test coverage of the new overflow path.

- The newly added overflow check is untested. `checkPagingLimit(results.getCount())` (`RegistryService.java:377` and `:398`) makes these two methods fail when results exceed the configured paging limit, but `testQueryParameterPropagation` (`RegistryServiceTest.java:325`) only asserts the `limit=201` query parameter. Add a test that returns more than `paging().limit()` results from the stubbed client and asserts the expected exception is thrown for both methods.
- Asserting `limit=201` bakes the default paging limit (200) into the test. Deriving the expected value from `config.paging().limit() + 1` would track the contract and avoid an opaque failure if the default ever changes.
- Worth confirming no other search methods in this class still hardcode `limit` or omit `checkPagingLimit`, so the fix is complete rather than partial.

princexpoddar:
Thanks @paoloantinori for the review:

- Derived `expectedLimit` in `testQueryParameterPropagation()` dynamically via `service.config.paging().limit() + 1`.
- Fixed `createService` delegation bug and rewrote overflow tests to verify realistic `count > limit` (3 vs 2) behavior alongside companion within-limit tests.
- Verified all search methods in `RegistryService.java` use `config.paging().limit() + 1` + `checkPagingLimit`.

All tests pass cleanly and ready for review!

mcp/src/test/java/io/apicurio/registry/mcp/RegistryServiceTest.java
princexpoddar: Fixed! Derived `expectedLimit` dynamically in `testQueryParameterPropagation()` via `int expectedLimit = service.config.paging().limit() + 1;`.

mcp/src/main/java/io/apicurio/registry/mcp/RegistryService.java
princexpoddar: Fixed! Added `testSearchAgentCardsThrowsWhenPagingLimitExceeded` and `testSearchMcpToolsThrowsWhenPagingLimitExceeded` to test the `ToolCallException` throw path when result count > limit (3 vs 2), along with companion tests for counts within limit.

paoloantinori:
Thanks for the follow-up. The new tests cover the right behavior: pagingLimit=2 with a mock returning count=3 exercises the actual overflow path through checkPagingLimit, and the within-limit companion test confirms the happy path. The dynamic expectedLimit derivation in testQueryParameterPropagation is the right call too. Looks good from my side.

jsenko (commented):
Fixes #9485 — replaces hardcoded \`limit = 50\` with \`config.paging().limit() + 1\` in \`searchAgentCards\` and \`searchMcpTools\`, and adds the missing \`checkPagingLimit()\` calls. Consistent with how every other search method in \`RegistryService\` handles paging.

No issues found.

- **Lambda parameter rename**: \`config\` → \`r\` is correct — the outer \`config\` field is needed inside the lambda for \`config.paging().limit()\`, so the parameter had to be renamed to avoid shadowing. Same pattern used by the other search methods in this class.
- **Test overflow signal**: Using \`name=overflow\` as a mock-server signal to return \`count=3\` is a pragmatic approach. The limit is set to 2 in the test, so count 3 correctly exceeds it.

princexpoddar:
@jsenko Thanks for the review!

princexpoddar:
This CI failure looks unrelated to my PR my changes only touch files in mcp/, nothing in schema-util. Might be a build order flake... Re-running to check

jsenko:
The build failure is **not caused by this PR** — it's a pre-existing issue on `main`.

**Root cause:** Commit `06697fa` ("fix(deps): update dependencies: maven app: all to v1.12.2") bumped `avro-maven-plugin` to 1.12.2, which is stricter about Avro schema `type` validation. A test schema in `app/src/test/resources/io/apicurio/registry/serde/` fails the new validation:

```
[ERROR] Failed to execute goal org.apache.avro:avro-maven-plugin:1.12.2:schema (default)
on project apicurio-registry-app: Error compiling a file in
.../app/src/test/resources/io/apicurio/registry/serde: A schema "type" MUST be a primitive
type or one of "enum", "fixed", "record", "error", "array" or "map".
```

The same failure reproduces on `main`'s latest [Verify run](https://github.com/Apicurio/apicurio-registry/actions/runs/32068366519). Once that's fixed upstream, a merge from `main` will resolve this PR's CI.

**No action needed from you** — your MCP changes are clean. Just merge `main` again once the Avro schema issue is fixed there.

jsenko:
Merged, thank you!

PR #9493, opened 2026-08-12
test(ui): add Playwright coverage for PROMPT_TEMPLATE viewer and test…

### LINKED ISSUE #8836: UI tests: no Playwright coverage for the PROMPT_TEMPLATE viewer and test panel

The Playwright suite in `ui/tests/specs` covers explore, smoke, global rules, settings, masthead and the reference graph, but nothing exercises the PROMPT_TEMPLATE flow. The pieces are all in place on both sides, `PromptTemplateViewer.tsx` and `PromptTemplateTestPanel.tsx` in the UI and the render endpoint on the server, so a regression in that flow would currently not be caught by CI.

### Proposed Solution

Add a `promptTemplate.spec.ts` to `ui/tests/specs` that creates a PROMPT_TEMPLATE artifact, opens the version page, checks the viewer shows the template and its variables, fills in values in the test panel and asserts the rendered output. A negative case for a missing required variable would round it out.

I can send a PR for this if it sounds useful.

krishkumar1577:
… panel

Fixes #8836. Adds promptTemplate.spec.ts covering:
- Creating a PROMPT_TEMPLATE artifact through the UI wizard
- Viewer renders template name, template text, and the variables table
- Test panel renders input fields for each variable, with correct default state for boolean variables
- Render with an unfilled required variable (currently renders with an empty substitution rather than a validation error -- noted as a possible follow-up, out of scope here)
- Render with the variable filled produces the expected substituted output

### DIFF
--- a/ui/tests/specs/data/promptTemplate-simple.ts
+++ b/ui/tests/specs/data/promptTemplate-simple.ts
@@ -1,0 +1,19 @@
+export const PROMPT_TEMPLATE_DATA = {
+    templateId: "test-support-prompt",
+    name: "Test Support Prompt",
+    description: "A minimal prompt template used to exercise the viewer and test panel",
+    version: "1.0",
+    template: "Answer the following question: {{question}}",
+    variables: {
+        question: {
+            type: "string",
+            required: true,
+            description: "The user's question"
+        },
+        include_examples: {
+            type: "boolean",
+            default: true,
+            description: "Whether to include code examples in the answer"
+        }
+    }
+};
--- a/ui/tests/specs/promptTemplate.spec.ts
+++ b/ui/tests/specs/promptTemplate.spec.ts
@@ -1,129 +1,81 @@
 import { test, expect } from "@playwright/test";
+import { PROMPT_TEMPLATE_DATA } from "./data/promptTemplate-simple";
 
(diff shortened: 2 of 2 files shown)


vjkumar2756 (commented):
Good test coverage for #8836 — just confirm cleanup + selector stability, and file a follow-up for the validation gap.

krishkumar1577:
@vjkumar2756 Thanks for the review! I added cleanup for the test-created artifact and group using a finally block, so the resources are removed even if an assertion fails. I also kept the required-variable validation behavior as a separate follow-up, as suggested. The existing test IDs and accessible role-based selectors remain in place for stability

krishkumar1577:
@EricWittmann this has an LGTM from @vjkumar2756 and cleanup is addressed marking it ready for review. Let me know if anything else is needed before merge.

krishkumar1577:
Hi @carlesarnal ,  just checking in on this whenever you get a chance to take a look. No rush, just wanted to make sure it hasn't slipped off the radar. Happy to make any changes if needed!

carlesarnal:
This has lots of conflicts and is touching lots of unrelated files.

krishkumar1577:
@carlesarnal I cleaned this PR to only the intended PROMPT_TEMPLATE Playwright test scope and force-pushed a conflict-free branch. Unrelated files/commits were removed. Checks are currently running on the new head — could you please re-review when you get a chance?

PR #9496, opened 2026-08-12
fix(ui): allow array values for skill and capability in AgentSearchFilters (#9446)

### LINKED ISSUE #9446: fix(ui): AgentSearchFilters in useAgentService.ts declares single string for skill and capability instead of array
Filing this separate issue for `AgentSearchFilters` (single vs list) as identified by @malladinagarjuna2 and requested by @EricWittmann in #9378.

### Description

In `ui-app`, `AgentSearchFilters` in `useAgentService.ts` (lines 42–46) declares `skill?: string` and `capability?: string` as single string values:

```typescript
export interface AgentSearchFilters {
    name?: string;
    skill?: string;
    capability?: string;
}
```

However, `WellKnownResource.searchAgents` on the backend accepts `List<String>` for both parameters to support multi-select filtering.

`AgentSearchFilters` should support `string | string[]` or `string[]` so the UI search service can pass multiple skills or capabilities to the discovery API.

karanthengane23:
Fixes #9446.

### Summary
Updates `AgentSearchFilters` in `useAgentService.ts` to support `string | string[]` for `capability` and `skill` filters, allowing the UI service to send multi-select query parameters to the backend discovery endpoints.

### What Changed
- **`useAgentService.ts`**:
  - Widened `AgentSearchFilters` interface to accept `capability?: string | string[]` and `skill?: string | string[]`.
  - Extracted `appendFilterParam` helper to handle both single strings and string arrays cleanly without code duplication.
- **`useAgentService.test.ts`**:
  - Added unit test coverage for single string values (`capability: "streaming"`), empty arrays (`capability: []`), and array entries containing empty strings.

### Investigation & Verification
- **Backend Contract:** Confirmed `WellKnownResource.java:132-133` (and `WellKnownResourceImpl.java:231-232`) binds `@QueryParam("capability") List<String>` and `@QueryParam("skill") List<String>`, confirming repeated query params (`?capability=a&capability=b`) are honored server-side.
- **Production Build:** Verified Vite production build (`npm --prefix ui/ui-app run build`) completes cleanly (1,876 modules, 0 errors).
- **Quality Gate:** SonarCloud analysis passed with 0 issues.

### DIFF
--- a/ui/ui-app/src/services/useAgentService.test.ts
+++ b/ui/ui-app/src/services/useAgentService.test.ts
@@ -1,0 +1,102 @@
+import { describe, expect, it, vi } from "vitest";
+import { Paging } from "@models/Paging.ts";
+
+const { fetchMock } = vi.hoisted(() => {
+    const registryConfig = { artifacts: { url: "http://localhost:8080/apis/registry/v3/" } };
+    (globalThis as any).ApicurioRegistryConfig = registryConfig;
+    (globalThis as any).window = {
+        ApicurioRegistryConfig: registryConfig,
+        location: { origin: "http://localhost:8080" }
+    };
+    return { fetchMock: vi.fn() };
+});
+
+vi.mock("@apicurio/common-ui-components", () => ({
+    useAuth: () => ({})
+}));
+
+(globalThis as any).fetch = fetchMock;
+
+import { useAgentService } from "./useAgentService";
+
+describe("useAgentService array and string filters", () => {
+    it("handles array values for capability and skill filters", async () => {
+        fetchMock.mockResolvedValue({
+            ok: true,
+            json: async () => ({ agents: [], count: 0 })
+        });
+
+        const service = useAgentService();
+        const paging: Paging = { page: 1, pageSize: 10 };
+
+        await service.searchAgents(
+            {
+                name: "test-agent",
+                capability: ["streaming", "pushNotifications"],
+                skill: ["search", "analysis"]
+            },
+            paging
+        );
+
+        expect(fetchMock).toHaveBeenCalledTimes(1);
+        const calledUrl = fetchMock.mock.calls[0][0] as string;
+        const parsedUrl = new URL(calledUrl);
+
+        expect(parsedUrl.searchParams.getAll("capability")).toEqual(["streaming", "pushNotifications"]);
+        expect(parsedUrl.searchParams.getAll("skill")).toEqual(["search", "analysis"]);
+        expect(parsedUrl.searchParams.get("name")).toBe("test-agent");
+    });
+
+    it("handles single string values for capability and skill filters", async () => {
+        fetchMock.mockReset();
+        fetchMock.mockResolvedValue({
+            ok: true,
+            json: async () => ({ agents: [], count: 0 })
+        });
+
+        const service = useAgentService();
+        const paging: Paging = { page: 1, pageSize: 10 };
+
+        await service.searchAgents(
+            {
+                name: "test-agent",
+                capability: "streaming",
+                skill: "search"
+            },
+            paging
+        );
+
+        expect(fetchMock).toHaveBeenCalledTimes(1);
+        const calledUrl = fetchMock.mock.calls[0][0] as string;
+        const parsedUrl = new URL(calledUrl);
+
+        expect(parsedUrl.searchParams.getAll("capability")).toEqual(["streaming"]);
+        expect(parsedUrl.searchParams.getAll("skill")).toEqual(["search"]);
+    });
+
+    it("handles empty arrays and arrays containing empty strings", async () => {
+        fetchMock.mockReset();
+        fetchMock.mockResolvedValue({
+            ok: true,
+            json: async () => ({ agents: [], count: 0 })
+        });
+
+        const service = useAgentService();
+        const paging: Paging = { page: 1, pageSize: 10 };
+
+        await service.searchAgents(
+            {
+                capability: [],
+                skill: ["streaming", "", "pushNotifications"]
+            },
+            paging
+        );
+
+        expect(fetchMock).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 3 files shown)


ui/ui-app/src/services/useAgentService.ts
```diff
@@ -87,10 +87,26 @@ const searchAgents = async (
         params.append("name", filters.name);
     }
     if (filters.capability) {
-        params.append("capability", filters.capability);
+        if (Array.isArray(filters.capability)) {
```
paoloantinori: This branch turns `[a, b]` into `?capability=a&capability=b`. Confirm the backend search endpoint binds `capability` and `skill` as multi-valued params (for example `List<String>`). If the resource declares a single-valued query param, the server retains only one value and drops the rest, so multi-select would silently filter on a single value. Verify this end-to-end against the actual search resource before relying on it.

ui/ui-app/src/services/useAgentService.test.ts
```diff
@@ -0,0 +1,49 @@
...
+            json: async () => ({ agents: [], count: 0 })
+        });
+
+        const service = useAgentService();
+        const paging: Paging = { page: 1, pageSize: 10 };
+
+        await service.searchAgents(
+            {
+                name: "test-agent",
+                capability: ["streaming", "pushNotifications"],
```
paoloantinori, line 39: This test only exercises the array happy path. The `string` (non-array) branch is still a documented type (`string | string[]`) but has no coverage, and the per-element empty-string filter (`if (c)`) plus the empty-array case are untested. Add cases for a single string value, an empty array, and an array containing an empty string so the param generation contract is pinned on both branches.

ui/ui-app/src/services/useAgentService.ts
```diff
@@ -87,10 +87,26 @@ const searchAgents = async (
...
+            filters.capability.forEach((c) => {
+                if (c) {
+                    params.append("capability", c);
+                }
+            });
+        } else {
+            params.append("capability", filters.capability);
+        }
     }
     if (filters.skill) {
```
paoloantinori: This `skill` block is an exact copy of the `capability` block above, differing only by the param key. Extract a helper that takes the key and the `string | string[]` value and handles both the array and string cases once, then call it for both filters. That keeps the two from drifting apart.

paoloantinori (commented):
The change correctly widens the filter types and emits repeated query params for arrays, with a focused unit test for the happy path. Three points are worth resolving before merge: the backend contract for repeated params is not confirmed, the new branching is only tested on the array path, and the capability and skill blocks are exact duplicates.

- `useAgentService.ts:90` now appends one `capability` param per array entry, producing `?capability=a&capability=b`. Confirm the backend search resource binds `capability` and `skill` as multi-valued (for example a `List<String>` query param). If the server binds a single `String`, only one value is retained and multi-select would silently filter on a single value.
- `useAgentService.test.ts:35` covers only the array happy path. The `string` (non-array) branch at `useAgentService.ts:97`, the empty-array case, and the per-element empty-string filter at line 92 are untested. Add cases for a single string value, an empty array, and an array containing an empty string so the param contract is pinned on both branches.
- `useAgentService.ts:100` duplicates the `capability` block verbatim apart from the param key. Extract a helper that takes the key and the `string | string[]` value and handles both cases once, then reuse it for both filters (and optionally `name`).

karanthengane23:
Thanks @paoloantinori for the review! I have addressed all three items:

1. **Backend Contract:** Confirmed - `WellKnownResource.java:131-132` (and `WellKnownResourceImpl.java:231-232`) binds `@QueryParam("capability") List<String> capability` and `@QueryParam("skill") List<String> skill`, so the server receives and filters multi-valued parameters cleanly.
2. **Unit Test Edge Cases:** Expanded `useAgentService.test.ts` with test cases covering single string values (`capability: "streaming"`), empty arrays (`capability: []`), and arrays containing empty strings (`skill: ["streaming", "", "pushNotifications"]`).
3. **DRY Refactor:** Extracted the `appendFilterParam` helper function in `useAgentService.ts` to handle string and array parameters cleanly and eliminate duplicate branching across `name`, `capability`, and `skill` filters.

ui/ui-app/src/services/useAgentService.ts
karanthengane23: Confirmed. WellKnownResource.java (L131-132) binds List<String> for both capability and skill params.

ui/ui-app/src/services/useAgentService.test.ts
karanthengane23: Added test cases for single string values, empty arrays, and empty string elements in useAgentService.test.ts.

ui/ui-app/src/services/useAgentService.ts
karanthengane23: Extracted appendFilterParam helper function to handle single strings and arrays without code duplication.

paoloantinori:
Thanks for addressing all three points. Confirmed the backend binds `skill` and `capability` as `List<String>` (`WellKnownResource.java:132-133`), so repeated query params are honored server-side. The `appendFilterParam` helper cleans up the duplication and the expanded test cases cover the string, empty-array, and empty-element branches. Looks good to me.

karanthengane23:
@EricWittmann Paolo approved the updates, ready for review whenever you get a chance! Thanks.

karanthengane23:
Hi @EricWittmann @paoloantinori,

I've fixed the CI test failure in commit `ee9984e32`:
- Added missing auth stubs (`isOidcAuthEnabled`, `isBasicAuthEnabled`, `getToken`) to the test mock so `createAuthOptions` executes cleanly.
- Added `server.deps.inline: ["@patternfly/react-styles"]`: in `vite.config.mts` to resolve the Node 20 CJS SyntaxError during Vitest execution.

The entire test suite and all CI verification checks are now passing 100% green. Thank you!

PR #9498, opened 2026-08-12
fix(prompt): add {{#if}}/{{#unless}}/{{else}} support to PromptRenderingService

Prathamesh09-coder:
This is a reopening of #9497, which was closed due to the one-PR-at-a-time limit.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
+++ b/app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
@@ -43,21 +43,21 @@ public class PromptRenderingService {
      * Group 1 = variable name, Group 2 = inner content.
      */
     private static final Pattern IF_BLOCK_PATTERN =
-            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{/if\\}\\}");
+            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}((?:(?!\\{\\{/if\\}\\})[\\s\\S])*?)\\{\\{/if\\}\\}");
 
     /**
-     * Matches {{#if var}} ... {{else}} ... {{/if}} blocks.
+     * Matches if-else blocks: opening tag, truthy branch, else tag, falsy branch, closing tag.
      * Group 1 = variable name, Group 2 = truthy branch, Group 3 = falsy branch.
      */
     private static final Pattern IF_ELSE_BLOCK_PATTERN =
-            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{else\\}\\}([\\s\\S]*?)\\{\\{/if\\}\\}");
+            Pattern.compile("\\{\\{#if\\s+(\\w+)\\}\\}((?:(?!\\{\\{else\\}\\})[\\s\\S])*?)\\{\\{else\\}\\}((?:(?!\\{\\{/if\\}\\})[\\s\\S])*?)\\{\\{/if\\}\\}");
 
     /**
-     * Matches {{#unless var}} ... {{/unless}} blocks.
+     * Matches unless blocks: opening tag, inner content, closing tag.
      * Group 1 = variable name, Group 2 = inner content.
      */
     private static final Pattern UNLESS_BLOCK_PATTERN =
-            Pattern.compile("\\{\\{#unless\\s+(\\w+)\\}\\}([\\s\\S]*?)\\{\\{/unless\\}\\}");
+            Pattern.compile("\\{\\{#unless\\s+(\\w+)\\}\\}((?:(?!\\{\\{/unless\\}\\})[\\s\\S])*?)\\{\\{/unless\\}\\}");
 
     /**
      * Renders a prompt template by substituting variables.
@@ -394,7 +394,7 @@ public class PromptRenderingService {
     private boolean isTruthy(Object value) {
         return value != null
                 && !Boolean.FALSE.equals(value)
-                && !(value instanceof String && ((String) value).isEmpty());
+                && !(value instanceof String s && s.isEmpty());
     }
 
     /**
@@ -408,7 +408,7 @@ public class PromptRenderingService {
      * </ol>
      */
     private String processConditionalBlocks(String template, Map<String, Object> variables) {
-        // 1. {{#if var}} ... {{else}} ... {{/if}}  (must run before plain {{#if}} to avoid partial match)
+        // Step 1: resolve if-else blocks first to prevent the plain-if pattern from partially matching them
         Matcher ifElseMatcher = IF_ELSE_BLOCK_PATTERN.matcher(template);
         StringBuffer ifElseResult = new StringBuffer();
         while (ifElseMatcher.find()) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -347,6 +384,69 @@ private String getTypeName(Object value) {
...
+
+    /**
+     * Process Handlebars-style conditional blocks before plain variable substitution.
+     *
+     * <p>Handles three forms in evaluation order:</p>
+     * <ol>
+     *   <li>{@code {{#if var}} ... {{else}} ... {{/if}}} — if/else (must be matched before plain if)</li>
+     *   <li>{@code {{#if var}} ... {{/if}}} — simple if</li>
+     *   <li>{@code {{#unless var}} ... {{/unless}}} — unless (logical complement of if)</li>
+     * </ol>
```
paoloantinori, line 408: This method resolves each block type in a single non-recursive pass, so nested blocks are silently broken. For `{{#if a}}{{#if b}}x{{/if}}{{/if}}` with both variables truthy, IF_BLOCK_PATTERN's lazy match treats the inner `{{/if}}` as the close of the outer block, leaving `{{#if b}}x` and a stray `{{/if}}` as literal text in the output instead of evaluating the inner block (traced through appendReplacement). The MCP reference converter has the same single-pass limitation, but adding else/unless invites more complex templates where nesting is more likely to be attempted. Either re-apply block resolution until the template stabilizes (with an iteration cap to bound malformed input), or state explicitly in the class Javadoc that nesting is unsupported.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -57,8 +93,9 @@ public RenderPromptResponse render(ContentHandle content, Map<String, Object> va
         // range is reported instead of being rendered silently.
         validateAppliedDefaults(variables, effectiveVariables, variablesSchema, validationErrors);
 
-        // Render the template by substituting variables
-        String rendered = substituteVariables(templateText, effectiveVariables);
+        // Process conditional blocks first, then substitute plain variables
+        String withBlocksResolved = processConditionalBlocks(templateText, effectiveVariables);
```
paoloantinori, line 97: processConditionalBlocks does not validate block structure, so an unclosed `{{#if x}}`, an unbalanced tag, or `{{else}}` placed inside `{{#unless}}` matches no pattern and is emitted verbatim into the rendered prompt with no error. validationErrors is already collected at this call site, so detecting unterminated or unbalanced block tags and reporting them would prevent a prompt silently containing literal `{{#if ...}}` markers. Silent passthrough matches the MCP path for plain `{{#if}}`, but for the new else/unless constructs there is no upstream reference, so the validation gap is newly introduced here rather than inherited.

app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java
```diff
@@ -347,6 +384,69 @@ private String getTypeName(Object value) {
         return value.getClass().getSimpleName();
     }
 
+    /**
+     * Determine whether a value is truthy for conditional evaluation.
+     *
+     * <p>A value is falsy when it is null, {@code Boolean.FALSE}, or an empty string.
```
paoloantinori, line 390: This Javadoc claims parity with the MCP render path, and the truthy logic does match PromptTemplateConverter.processConditionalBlocks exactly. The broader feature parity does not hold, though: the MCP converter's regex handles only `{{#if}}`, with no `{{else}}` or `{{#unless}}` support (traced at mcp/.../PromptTemplateConverter.java). A template using `{{else}}` or `{{#unless}}` therefore renders correctly through this REST path but leaks the raw markers through the MCP path. Worth a follow-up to port else/unless to PromptTemplateConverter, or a Javadoc note that the two render paths are not feature-equivalent.

app/src/test/java/io/apicurio/registry/services/PromptRenderingServiceTest.java
```diff
@@ -761,9 +761,10 @@ public void testConditionalBlockMarkersAreNotSubstituted() {
         RenderPromptResponse response = renderingService.render(content, variables,
                 "default", "conditional", "1.0");
 
-        // The render endpoint does not process {{#if}} blocks (see #8728); it must at least leave
-        // the markers alone rather than mangling them while substituting the variable inside.
-        Assertions.assertEquals("{{#if premium}}Hello Alice{{/if}}", response.getRendered());
+        // PR #9391 (#8728): the render endpoint now processes {{#if}} blocks.
+        // With premium=true the block is truthy, so the inner content is rendered
+        // and the markers themselves are removed.
+        Assertions.assertEquals("Hello Alice", response.getRendered());
```
paoloantinori, line 767: The test behavior was inverted but the name was not updated. `testConditionalBlockMarkersAreNotSubstituted` now asserts the block IS processed and the markers are removed (`Hello Alice`), the opposite of what the name says. Rename to something like `testConditionalBlockIsProcessedAndMarkersRemoved` so the name matches the assertion.

paoloantinori (commented):
Overall the change adds if, if-else, and unless conditionals with the correct evaluation order (if-else before plain if, then unless) and safe replacement via Matcher.quoteReplacement. The truthy semantics in isTruthy are character-for-character identical to the MCP reference PromptTemplateConverter.processConditionalBlocks (verified by tracing the converter source), so that parity claim holds. The remaining concerns are structural: block resolution is a single non-recursive pass with no tag validation, and the new else/unless constructs have no counterpart in the MCP converter, so templates now render differently across the REST and MCP render paths.

- Nested conditionals resolve incorrectly (PromptRenderingService.java:408): a single non-recursive pass per block type leaves inner blocks and stray closing tags as literal text.
- Unterminated or unbalanced tags, and {{else}} used inside {{#unless}}, match no pattern and pass through verbatim with no error (PromptRenderingService.java:97).
- Cross-path divergence (PromptRenderingService.java:390): the MCP converter only handles {{#if}}, so {{else}} and {{#unless}} render correctly here but leak raw markers through the MCP path.
- A renamed test kept its old, now-contradictory name (PromptRenderingServiceTest.java:767).
- Positive: resolving conditionals before substituteVariables means a user-supplied variable value cannot inject {{#if}} logic into the template.

EricWittmann (commented):
Thanks for working on this, @Prathamesh09-coder. The conditional block logic itself is correct and the test coverage is solid.

One architectural concern that the other reviews didn't flag: when @carlesarnal confirmed this bug in [#8728](https://github.com/Apicurio/apicurio-registry/issues/8728#issuecomment-5019419804), the guidance was to **extract the conditional processing logic into a shared utility** so that both `PromptRenderingService` (REST) and `PromptTemplateConverter` (MCP) call the same code:

> *"The cleanest approach would be moving `processConditionalBlocks()` and its truthiness evaluation into a shared utility (maybe in `common/` or `schema-util/`) that both `PromptRenderingService` and `PromptTemplateConverter` can call."*

This PR duplicates the logic in `PromptRenderingService` instead. Combined with the fact that the new `{{else}}` and `{{#unless}}` constructs aren't ported to the MCP side (as Paolo noted), we'd end up with two separate implementations with *different* feature sets — the opposite of what was asked for.

Could you rework this to extract a shared utility and have both call sites use it? That would address this concern and Paolo's MCP parity point in one shot.

Prathamesh09-coder:
Thanks @EricWittmann! I’ve reworked the implementation to use PromptTemplateVariableUtil as the shared source of truth for conditional rendering. Both PromptRenderingService and PromptTemplateConverter now use the same processConditionalBlocks() and isTruthy() implementation, and I added coverage for the shared utility and both render paths. Local tests and Checkstyle are passing; CI is now running.

EricWittmann:
Thanks @Prathamesh09-coder :+1:

PR #9499, opened 2026-08-12
fix(ui): display numeric minimum/maximum range in prompt template Variables table

### LINKED ISSUE #9361: fix(ui): prompt template Variables table doesn't show numeric minimum/maximum range
**Registry Version**: 3.3.2-SNAPSHOT

The prompt template schema supports `minimum` and `maximum` constraints on numeric variables (validated on the backend in `PromptTemplateContentValidator.java`, and used at render time in `PromptRenderingService.java`), but the Documentation tab's Variables table in `PromptTemplateViewer.tsx` never displays them.

The Variables table currently has these columns: Name, Type, Required, Default, Allowed Values, Description. A user reading the docs for a variable with `minimum: 1, maximum: 10` has no visible indication of the valid range, even though the backend will reject values outside it.

Related type-model issue in the same file: the `PromptVariable` TypeScript interface has a phantom `constraints?: any` field that is not in the schema, while the real `minimum` / `maximum` fields are missing from the interface.

### Steps to Reproduce

1. Create a PROMPT_TEMPLATE artifact with a numeric variable that has `minimum` and/or `maximum`, for example:

```json
{
  "name": "Range Demo",
  "template": "give me {{items}} items",
  "variables": {
    "items": { "type": "integer", "minimum": 1, "maximum": 10, "description": "how many" },
    "score": { "type": "number", "minimum": 0, "description": "open-max" },
    "cap":   { "type": "integer", "maximum": 100, "description": "open-min" }
  }
}
```

2. Open the version's Documentation tab and look at the Variables table.

### Expected vs Actual

**Expected:** the table shows the numeric range for each variable, e.g. `1 – 10`, `≥ 0`, `≤ 100`.

**Actual:** the range is not shown anywhere in the table. There is no way to see it without opening the raw content.

### Proposed Fix

Add a "Range" column to the Variables table that renders the range for each variable, and fix the `PromptVariable` interface (drop the phantom `constraints?: any`, add real `minimum?: number` and `maximum?: number`).

Shubb07:
Fixes #9361

## What

Adds a **Range** column to the Variables table on the prompt template Documentation tab, so users can see the valid numeric bounds for each variable without opening the raw content.

Rendering rules (from `formatRange` in `PromptTemplateViewer.utils.ts`):

| minimum | maximum | Cell |
|---------|---------|------|
| `1`     | `10`    | `1 – 10` |
| `0`     | –       | `≥ 0` |
| –       | `100`   | `≤ 100` |
| –       | –       | `-` |

Also fixes the `PromptVariable` interface in `PromptTemplateViewer.tsx`:
- **removes** `constraints?: any` — phantom field, not part of the schema, and grep confirms it was never read anywhere.
- **adds** `minimum?: number` and `maximum?: number` — the actual JSON-Schema fields the backend already validates against.

## Why

The backend enforces `minimum`/`maximum` on numeric variables, but the UI was silently dropping that information. Users had to open the raw JSON to know what values were legal — a bad experience for anyone consuming a template someone else wrote.

## Tests

- `ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.utils.test.ts` — 5 new cases for `formatRange` covering both bounds, min-only, max-only, neither, and non-finite (`NaN`/`Infinity`) inputs.
- Full file: 20/20 passing locally (`npm test`).

## Live check

Local browser verification is currently blocked by a Docker/WSL issue on my machine — I couldn't stand up the backend snapshot to render a live template. Change is small and mechanical (pure formatter + one `<td>`), and the formatter's edge cases are all unit-tested. Happy to add a screenshot once I get the local stack running again.

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
@@ -18,7 +18,8 @@ import {
 } from "@patternfly/react-core";
 import { JsonSchemaProperties } from "@app/components/jsonSchema/JsonSchemaProperties";
 import { VariableSchema } from "./promptTemplateVariables";
-import { highlightVariables } from "./PromptTemplateViewer.utils";
+import { formatRange, highlightVariables } from "./PromptTemplateViewer.utils";
+
 
 export interface PromptTemplateMetadata {
     author?: string;
@@ -154,6 +155,7 @@ export const PromptTemplateViewer: FunctionComponent<PromptTemplateViewerProps> 
                                         <th>Required</th>
                                         <th>Default</th>
                                         <th>Allowed Values</th>
+                                        <th>Range</th>
                                         <th>Description</th>
                                     </tr>
                                 </thead>
@@ -183,6 +185,7 @@ export const PromptTemplateViewer: FunctionComponent<PromptTemplateViewerProps> 
                                                     </LabelGroup>
                                                 ) : "-"}
                                             </td>
+                                            <td>{formatRange(variable.minimum, variable.maximum) ?? "-"}</td>
                                             <td>{variable.description || "-"}</td>
                                         </tr>
                                     ))}
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.utils.test.ts
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.utils.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { tokenizeTemplate } from "./PromptTemplateViewer.utils";
+import { formatRange, tokenizeTemplate } from "./PromptTemplateViewer.utils";
 
 describe("tokenizeTemplate", () => {
     it("keeps plain text without variables as a single plain token", () => {
@@ -106,5 +106,27 @@ describe("tokenizeTemplate", () => {
 
     it("handles an empty template", () => {
         expect(tokenizeTemplate("")).toEqual([]);
+    });
+});
+
+describe("formatRange", () => {
+    it("renders 'min – max' when both bounds are present", () => {
+        expect(formatRange(1, 10)).toBe("1 – 10");
+    });
+
+    it("renders '≥ min' when only the minimum is present", () => {
+        expect(formatRange(0, undefined)).toBe("≥ 0");
+    });
+
+    it("renders '≤ max' when only the maximum is present", () => {
+        expect(formatRange(undefined, 100)).toBe("≤ 100");
+    });
+
+    it("returns null when neither bound is present", () => {
+        expect(formatRange(undefined, undefined)).toBeNull();
+    });
+
+    it("ignores non-finite values", () => {
+        expect(formatRange(NaN, Infinity)).toBeNull();
     });
 });
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.utils.test.ts
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.utils.test.ts
@@ -129,4 +129,22 @@ describe("formatRange", () => {
     it("ignores non-finite values", () => {
         expect(formatRange(NaN, Infinity)).toBeNull();
     });
+
+    it("ignores null values", () => {
+        // TS-loose call site (JSON parsers hand back null, not undefined)
+        expect(formatRange(null as unknown as number, null as unknown as number)).toBeNull();
+    });
+
+    it("ignores stringified numbers", () => {
+        // Number.isFinite does not coerce, so "5" is not treated as a bound
+        expect(formatRange("5" as unknown as number, "10" as unknown as number)).toBeNull();
+    });
+
+    it("renders separated bounds when minimum > maximum (malformed schema)", () => {
+        expect(formatRange(10, 1)).toBe("≥ 10, ≤ 1");
+    });
+
+    it("renders a normal range when minimum equals maximum", () => {
+        expect(formatRange(5, 5)).toBe("5 – 5");
+    });
 });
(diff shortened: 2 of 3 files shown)


ui/ui-app/src/app/components/promptTemplate/PromptTemplateViewer.tsx
```diff
@@ -192,6 +194,7 @@ export const PromptTemplateViewer: FunctionComponent<PromptTemplateViewerProps>
                                                     </LabelGroup>
                                                 ) : "-"}
                                             </td>
+                                            <td>{formatRange(variable.minimum, variable.maximum) ?? "-"}</td>
```
paoloantinori, line 188: This renders `variable.minimum`/`maximum`, but the diff never shows where `PromptVariable` objects are constructed. If that construction is explicit field-mapping rather than a raw JSON-Schema spread, `minimum` and `maximum` are never populated and every cell in this column will render `-`, making the column dead weight. The PR notes live rendering was not verified and the new tests only cover `formatRange` in isolation. Please point at the construction site (or add a test that feeds a parsed schema through it) and confirm at least one real numeric variable renders a non-`-` value end-to-end. Same question for `minimum > maximum`: should this guard or clamp, or do we rely on the backend rejecting that?

paoloantinori (commented):
The formatter is clean and well-tested in isolation, but the PR only proves `formatRange` works, not that real variables carry `minimum`/`maximum` data. I could not locate the `PromptVariable` construction site in the provided context, and the PR description notes live rendering was never verified, so the most important question (does any cell ever show something other than `-`?) is unconfirmed.

- `PromptTemplateViewer.utils.ts:39-46` (`formatRange`): the `typeof minimum === "number" && Number.isFinite(minimum)` guard is correct and rejects `undefined`, `null`, `NaN`, `Infinity`, and string values. The `typeof` check is redundant given `Number.isFinite` never coerces, but harmless; NIT, not blocking.
- `PromptTemplateViewer.utils.ts:42`: when `minimum > maximum` (e.g. a malformed schema), the cell renders a nonsensical inverted range ("10 - 1"). JSON-Schema does not require `minimum <= maximum`, so consider guarding or clamping here, or confirm the backend rejects that combination. Worth a deliberate decision rather than silently displaying it.
- `PromptTemplateViewer.tsx:194`: header (`Range`) and body cell are inserted in matching positions before `Description`, consistent with the sibling `-` fallback pattern. Wiring itself is fine.
- `PromptTemplateViewer.utils.test.ts:111-129`: good coverage of the formatter (both bounds, min-only, max-only, neither, non-finite). The gap is that no test exercises the data path that populates `minimum`/`maximum`, which is where a silent regression would actually hide.
- `PromptTemplateViewer.tsx:29-30`: removing `constraints?: any` is justified; if anything had read it, the TS compiler would have flagged it. Adding `minimum?`/`maximum?` as optional is non-breaking.

Shubb07:
Thanks for the review @paoloantinori. Follow-up in 8ce36d3.

I ran the branch locally against latest-snapshot with a test template covering all 5 cases (both bounds, min-only, max-only, neither, and inverted min>max). Cells render: "1 – 10", "≥ 0", "≤ 100", "-", and "≥ 10, ≤ 1" respectively. So the data path does populate min/max and the cell shows more than "-".

On your points:

1. Redundant typeof: dropped. Only Number.isFinite now (it doesn't coerce, so it rejects null/strings on its own).
2. Inverted min>max: guarded. Renders both bounds separately as "≥ min, ≤ max" instead of a nonsense range. Preferred over clamping since the values are the user's schema data, not ours to rewrite, and the backend doesn't reject the combination either.
3. Column wiring: no change.
4. Tests: added cases for null, stringified numbers, inverted, and min==max. A full component test would need RTL+jsdom, which ui-app doesn't currently ship - probably a separate PR. Happy to file one if you think it's worth it.
5. constraints removal: no change.

paoloantinori:
Thanks for following up. The `Number.isFinite`-only check and the new test cases (including the malformed-schema `min > max` branch and the loose-call-site null/string cases) address the points from my earlier read. Glad the data-path check confirms the cells populate end-to-end. Nothing further from me on this one.

Shubb07:
Rebased onto latest main. Heads-up on what changed in the diff:

#8974 (variable auto-detection) landed after I opened this and extracted the variable type to `promptTemplateVariables.ts` as `VariableSchema`, which already includes `minimum?: number` and `maximum?: number`. So the interface changes from my original PR (dropping `constraints?: any` + adding `minimum`/`maximum`) are effectively already in main via that PR - the fields are on `VariableSchema` instead of the old inline `PromptVariable`.

What's left in this PR after rebase:
- `formatRange` helper in `PromptTemplateViewer.utils.ts` + tests.
- Range column in the Variables table using `variable.minimum` / `variable.maximum` (which now come from `VariableSchema`).

Diff is 3 files, +59/-2. All 52 prompt-template tests still green locally.

Shubb07:
Anything else you'd like me to change on this one?

paoloantinori:
Nothing else from my side — the latest rebase keeps the previous points addressed and the test coverage looks solid. Thanks for following up on each round.

Shubb07:
Heads-up — the License Compliance status check errored at 11:44 with no logs. Looks like a service flake since the diff has no dep changes. Not sure if I can re-trigger it from my end.

paoloantinori:
Checked on my side: this is not specific to your PR. The License Compliance status is currently in error on every recent commit I looked at, including the latest commit on main and the head of an unrelated PR (#9582). Dependency Quality and Security Analysis (the two sibling statuses from the same service) are green everywhere, so the code-analysis side of the pipeline is fine.

Also worth noting: License Compliance reports as a bare commit status with no target URL and no Actions check-run behind it, which is why there are no logs to inspect. It comes from an external service, so there is nothing you can re-trigger from your end, and it should clear once the service recovers. The checks that do exercise your diff (Build, Extra Tests, Verification Gate, SonarCloud) are all green on this head.

PR #9506, opened 2026-08-12
fix(ui): prevent double-prefixed URLs when contextPath and navPrefixPath overlap

### LINKED ISSUE #9505: Double-prefixed URL and 404 when both contextPath and navPrefixPath are set
## Summary
When both `ui.contextPath` and `ui.navPrefixPath` are configured with an
overlapping value (e.g. both set to `/registry`), the app's internal
navigation produces a double-prefixed URL, resulting in a 404.

## Root cause
- `App.tsx` passes `contextPath` as React Router's `basename`  Router
  automatically prepends this to every navigation.
- `useAppNavigation.ts`'s `createLink()` / `navigateTo()` *also* manually
  prepends `navPrefixPath` on top of that.
- When both values are set (or overlap), the prefix is applied twice.

## Steps to reproduce
1. Set both `contextPath` and `navPrefixPath` to the same value in
   `config.js`, e.g.:
```js
   ui: {
       contextPath: "/registry",
       navPrefixPath: "/registry"
   }
```
2. Serve the built app under its configured base path, e.g. `/registry/`.
3. Load the root URL  the app redirects (via `RootRedirectPage` →
   `appNav.createLink("/dashboard")`).

## Expected
Redirect to `/registry/dashboard`.

## Actual
Redirects to `/registry/registry/dashboard`, which does not match any
route, resulting in a full "404 Error: page not found" page.

## Screenshot

Image

## Impact
This breaks the app entirely for any deployment that configures both
`contextPath` and `navPrefixPath` with overlapping values  not a
cosmetic issue, the app becomes unusable at that URL.

Faisal77666:
Fixes #9505

## What this PR does
Resolves the effective navigation prefix in `useAppNavigation.ts` so that
`navPrefixPath` isn't applied on top of React Router's `basename`
(`contextPath`) when the two overlap.

## Why
`<Router basename={contextPath}>` already prepends `contextPath` to every
navigation automatically. `createLink()` / `navigateTo()` were *also*
manually prepending `navPrefixPath` on top, so when both were configured
with an overlapping value, the resulting URL was double-prefixed (e.g.
`/registry/registry/dashboard`), which matches no route and 404s.

## Changes
- `ui/ui-app/src/services/useAppNavigation.ts`
  - Added `effectiveNavPrefixPath()`: returns `""` instead of
    `navPrefixPath` when `navPrefixPath === contextPath`, since Router's
    `basename` already accounts for it
  - Used in both `navigateTo()` and `createLink()`

## Testing done
- `tsc && vite build` clean
- End-to-end verification: built and served the app under a `/registry`
  subpath (both `contextPath` and `navPrefixPath` set to `/registry`),
  connected to a live backend
  - **Before:** root redirect resolved to `/registry/registry/dashboard`
    → 404 "page not found"
  - **After:** root redirect correctly resolves to `/registry/dashboard`
    → dashboard loads normally

## Screenshots
**Before:**

Screenshot From 2026-08-11 09-01-48

**After:**

Screenshot From 2026-08-11 15-36-55


Faisal77666:
@paoloantinori please take a look

paoloantinori (commented):
Looked at this, thanks for the ping. The fix correctly resolves the reported double-prefix for the exact-match case (`contextPath === navPrefixPath`), and the explanation of why Router's `basename` already accounts for `contextPath` is spot on.

One gap worth flagging: the guard is a strict equality check, so it only fires when the two values are byte-identical. The PR title and description frame the problem as an "overlap", but a couple of realistic overlap shapes still produce a double prefix:

- **Trailing-slash mismatch.** `contextPath = "/registry"` and `navPrefixPath = "/registry/"` (or the reverse) are functionally the same prefix but would fail `prefix === basename`, so the `/registry/` side gets applied twice. Config sources that normalize inconsistently (env vars vs OpenShift route annotations vs console config) are a common source of this.
- **Prefix-is-subpath.** If `navPrefixPath = "/registry/ui"` while `contextPath = "/registry"`, Router already prepends `/registry`, so the effective result is `/registry/registry/ui/...`. The strict-equality guard does not catch this either.

If the only deployment shape you intend to support is the exact-equality one (which is reasonable for a first fix), it would help to either narrow the title/description to "exact-match" so the scope is honest, or normalize both values (trim trailing slashes, and optionally check `basename.startsWith(prefix)` / `prefix.startsWith(basename)`) so the guard matches the "overlap" framing. Worth confirming which operators actually set these two values and whether they can ever differ by a trailing slash in practice.

Separately: any reason not to also cover the symmetric case where `navPrefixPath` is set but `contextPath` is empty? Today that falls through to returning `prefix` unchanged, which is correct, just want to make sure that path is intentional rather than unhandled.

ui/ui-app/src/services/useAppNavigation.ts
```diff
@@ -1,8 +1,24 @@
...
+// React Router's <Router basename={contextPath}> already prepends contextPath
+// to every navigation automatically. If navPrefixPath is also configured with
+// a value that overlaps with contextPath, applying both would double-prefix
+// the resulting URL (e.g. "/registry/registry/dashboard"), which does not
+// match any route. This resolves the effective prefix so it is only applied
+// once.
+const effectiveNavPrefixPath = (config: ConfigService): string => {
     const prefix: string = config.uiNavPrefixPath() || "";
+    const basename: string = config.uiContextPath() || "";
+    if (prefix !== "" && prefix === basename) {
```
paoloantinori: The `prefix === basename` equality only catches byte-identical values. A trailing-slash difference (e.g. `contextPath="/registry"`, `navPrefixPath="/registry/"`) or a subpath overlap (`navPrefixPath="/registry/ui"` under `contextPath="/registry"`) still double-prefixes, because Router already prepends `contextPath` via `basename`. If the intent is to guard against "overlap" as the title says, consider normalizing trailing slashes and/or testing `startsWith` rather than strict equality. If only exact-match is in scope for this fix, narrowing the title would make the scope honest.

paoloantinori:
Posted my review above — flagging that the strict-equality guard misses trailing-slash and subpath overlaps. Thanks for the ping.

Faisal77666:
@paoloantinori Good catches fixed both. Extracted the logic into
useAppNavigation.utils.ts and switched to normalized-trailing-slash +
startsWith() comparison instead of strict equality:
Trailing-slash mismatch now normalizes both sides before comparing
Subpath overlap (navPrefixPath="/registry/ui" under
contextPath="/registry") now correctly resolves to just the
non-overlapping remainder ("/ui") instead of double-prefixing.Also added unit tests for all four cases (exact match, trailing-slash mismatch, subpath overlap, and root-contextPath passthrough) to guard
against regressions.
On your separate question  when navPrefixPath is set and contextPath
is the default root ("/"), the function returns navPrefixPath
unchanged. That's intentional: there's nothing to overlap with, so the
existing behavior (fully manual prefixing) is preserved.

paoloantinori:
Re-read after the rebase. Both points from the earlier review are addressed: the strict-equality guard is gone (replaced by `normalizePath` + `startsWith`), and the subpath-overlap case now correctly yields just the non-overlapping remainder. The four test cases cover exact match, trailing-slash mismatch, subpath overlap, and the default-root fallback, which is the right matrix.

Two minor notes, neither blocking:
- `mergeable_state` is `behind`, so this will need a rebase before merge — just flagging so it doesn't get lost.
- Worth confirming the `basename !== "/"` short-circuit still does the right thing if an operator sets `contextPath` to the empty string rather than `"/"` (the `normalizePath` branch handles `""`, but an explicit test for `contextPath = ""` would pin that edge down).

Looks good to me. Thanks for the quick turnaround.

Faisal77666:
@EricWittmann ready to merge

PR #9510, opened 2026-08-12
fix(storage): enable GZIP compression for H2 snapshot dumps

### LINKED ISSUE #8961: gzip snapshot dumps and optionally store them in Kafka for K8s without PVC

EricWittmann:
Fixes #8961

### What
Enables GZIP compression for H2 database snapshots used by the KafkaSQL storage variant.

### Why
Uncompressed `.sql` snapshot dumps can grow large, consuming significant disk space
(especially in `/tmp`) and slowing down replica startup times. This is particularly
problematic in ephemeral environments like Kubernetes without persistent volumes.

### How
- Added a `COMPRESSED_SNAPSHOT_EXTENSION` constant (`.sql.gz`) to `SqlStatements`.
- Added location-aware `default` overloads of `createDataSnapshot(String)` and
  `restoreFromSnapshot(String)` to the `SqlStatements` interface. The defaults delegate
  to the no-arg versions, so PostgreSQL/MySQL/MSSQL implementations are unaffected.
- Overrode both methods in `H2SqlStatements` to append H2's native `COMPRESSION GZIP`
  clause when the snapshot location ends with `.sql.gz`.
- Updated `SqlExportRepository` and `SqlRegistryStorage` to pass the snapshot location
  through to the new overloads so the correct SQL is generated.
- Changed `KafkaSqlRegistryStorage.triggerSnapshotCreation()` to generate snapshot
  files with the `.sql.gz` extension.
- Backward compatible: the restore path in `consumeSnapshotsTopic()` handles both old
  `.sql` and new `.sql.gz` snapshots transparently based on the file extension.

### Testing
- Added `H2SqlStatementsTest` with unit tests covering GZIP and plain extensions for
  both create and restore, including null location handling.

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/H2SqlStatements.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/H2SqlStatements.java
@@ -82,8 +82,24 @@ public class H2SqlStatements extends CommonSqlStatements {
     }
 
     @Override
+    public String createDataSnapshot(String location) {
+        if (location != null && location.endsWith(COMPRESSED_SNAPSHOT_EXTENSION)) {
+            return "SCRIPT TO ? COMPRESSION GZIP";
+        }
+        return createDataSnapshot();
+    }
+
+    @Override
     public String restoreFromSnapshot() {
         return "RUNSCRIPT FROM ?";
+    }
+
+    @Override
+    public String restoreFromSnapshot(String location) {
+        if (location != null && location.endsWith(COMPRESSED_SNAPSHOT_EXTENSION)) {
+            return "RUNSCRIPT FROM ? COMPRESSION GZIP";
+        }
+        return restoreFromSnapshot();
     }
 
     @Override
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/SqlRegistryStorage.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/SqlRegistryStorage.java
@@ -35,7 +35,7 @@ public class SqlRegistryStorage extends AbstractSqlRegistryStorage {
     }
 
     public void restoreFromSnapshot(String snapshotLocation) {
-        handleFactory.withHandle(handle -> handle.createUpdate(sqlStatements.restoreFromSnapshot())
+        handleFactory.withHandle(handle -> handle.createUpdate(sqlStatements.restoreFromSnapshot(snapshotLocation))
                 .bind(0, snapshotLocation).execute());
     }
 
--- a/app/src/test/java/io/apicurio/registry/storage/impl/sql/H2SqlStatementsTest.java
+++ b/app/src/test/java/io/apicurio/registry/storage/impl/sql/H2SqlStatementsTest.java
@@ -1,0 +1,45 @@
+package io.apicurio.registry.storage.impl.sql;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class H2SqlStatementsTest {
+
+    private final H2SqlStatements statements = new H2SqlStatements();
+
+    @Test
+    void testCreateDataSnapshotWithGzipExtension() {
+        String sql = statements.createDataSnapshot("/tmp/snapshot.sql.gz");
+        Assertions.assertEquals("SCRIPT TO ? COMPRESSION GZIP", sql);
+    }
+
+    @Test
+    void testCreateDataSnapshotWithPlainExtension() {
+        String sql = statements.createDataSnapshot("/tmp/snapshot.sql");
+        Assertions.assertEquals("SCRIPT TO ?", sql);
+    }
+
+    @Test
+    void testCreateDataSnapshotWithNullLocation() {
+        String sql = statements.createDataSnapshot(null);
+        Assertions.assertEquals("SCRIPT TO ?", sql);
+    }
+
+    @Test
+    void testRestoreFromSnapshotWithGzipExtension() {
+        String sql = statements.restoreFromSnapshot("/tmp/snapshot.sql.gz");
+        Assertions.assertEquals("RUNSCRIPT FROM ? COMPRESSION GZIP", sql);
+    }
+
+    @Test
+    void testRestoreFromSnapshotWithPlainExtension() {
+        String sql = statements.restoreFromSnapshot("/tmp/snapshot.sql");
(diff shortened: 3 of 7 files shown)


paoloantinori (commented):
The change is well-scoped: the location-aware default methods on `SqlStatements` keep PostgreSQL/MySQL/MSSQL untouched, and the integration test confirms H2 actually emits GZIP on creation. Before merge, two things need attention: the restore half (the upgrade-critical path) ships without any functional test and its wiring through `consumeSnapshotsTopic()` is not shown in the diff, and the new test file is missing its license header.

- `H2SqlStatementsTest.java:1` is missing the Apache 2.0 license header that the rest of the module's test sources carry; the project's license check in CI will flag this file.
- `H2SqlStatements.java:98` and `SqlRegistryStorage.java:38`: `restoreFromSnapshot(String)` is covered only by a string-equality unit test. There is no test that actually restores a `.sql.gz` snapshot, and none for a legacy `.sql` snapshot, so the backward-compatibility claim is unverified. `KafkaSqlSnapshotTest` exercises creation only.
- The PR description states `consumeSnapshotsTopic()` handles both extensions, but no hunk in this diff touches that method. Confirm the restore call there passes the location into `restoreFromSnapshot(snapshotLocation)` so the extension is inspected; if it still calls the no-arg form, a `.sql.gz` snapshot would be restored with plain `RUNSCRIPT` and fail to decompress, breaking replica startup after upgrade.
- Minor: the `location != null && location.endsWith(COMPRESSED_SNAPSHOT_EXTENSION)` check is duplicated at `H2SqlStatements.java:86` and `:99`; a private `isCompressedSnapshot(location)` helper would remove it.
- Minor edge case: `.sql.gz` naming is driven by `KafkaSqlRegistryStorage`, but `createDataSnapshot`/`restoreFromSnapshot` are also reachable via `SqlExportRepository` on non-H2 datasources, where the default implementation would write or read an uncompressed file under a `.sql.gz` name. Likely out of scope, but worth a note if export is ever pointed at a `.sql.gz` target on Postgres/MySQL.

app/src/test/java/io/apicurio/registry/storage/impl/sql/H2SqlStatementsTest.java
```diff
@@ -0,0 +1,45 @@
+package io.apicurio.registry.storage.impl.sql;
```
paoloantinori, line 1: Add the Apache 2.0 license header that the rest of the module's test sources carry. The project's license check will flag this new file without it.

app/src/main/java/io/apicurio/registry/storage/impl/sql/H2SqlStatements.java
```diff
@@ -81,11 +81,27 @@ public String createDataSnapshot() {
...
+        return createDataSnapshot();
+    }
+
     @Override
     public String restoreFromSnapshot() {
         return "RUNSCRIPT FROM ?";
     }
 
+    @Override
+    public String restoreFromSnapshot(String location) {
```
paoloantinori, line 98: This override is covered only by a string-equality unit test in `H2SqlStatementsTest`. Restore is the upgrade-critical path (new replicas restoring snapshots published by older nodes), and `KafkaSqlSnapshotTest` only exercises creation. Add a functional test that restores an actual `.sql.gz` snapshot and asserts the data loads, plus one for a legacy `.sql` snapshot to validate the backward-compatibility claim.

app/src/main/java/io/apicurio/registry/storage/impl/sql/SqlRegistryStorage.java
```diff
@@ -35,7 +35,7 @@ public void initialize() {
     }
 
     public void restoreFromSnapshot(String snapshotLocation) {
-        handleFactory.withHandle(handle -> handle.createUpdate(sqlStatements.restoreFromSnapshot())
+        handleFactory.withHandle(handle -> handle.createUpdate(sqlStatements.restoreFromSnapshot(snapshotLocation))
```
paoloantinori, line 38: The PR description says `consumeSnapshotsTopic()` handles both `.sql` and `.sql.gz`, but no hunk in this diff touches that method and no test covers it. Confirm the restore call there routes the snapshot location through `restoreFromSnapshot(snapshotLocation)` so the extension is inspected. If it still calls the no-arg `restoreFromSnapshot()`, a `.sql.gz` snapshot restores with plain `RUNSCRIPT` and fails to decompress, breaking replica startup after upgrade.

Issue #9483, opened 2026-08-11
refactor: unify prompt template rendering logic between REST and MCP paths


EricWittmann:
## Problem

The prompt template rendering pipeline is implemented independently in two places:

- **REST path:** `PromptRenderingService` (`app/src/main/java/io/apicurio/registry/services/PromptRenderingService.java`)
- **MCP path:** `PromptTemplateConverter` (`mcp/src/main/java/io/apicurio/registry/mcp/PromptTemplateConverter.java`)

Both classes perform the same logical steps — parse the template, apply variable defaults,
process conditional blocks, substitute variables — but each has its own implementation. The
only piece currently shared is the `{{variable}}` pattern and substitution method in
`PromptTemplateVariableUtil` (introduced in #8977).

## What is duplicated

| Concern | REST (`PromptRenderingService`) | MCP (`PromptTemplateConverter`) |
|---|---|---|
| Template parsing (YAML/JSON) | Own `parseContent()` using `YAML_MAPPER`/`MAPPER` | Own `parseContentAutoDetect()` / `parseContent()` using local `ObjectMapper` instances |
| Conditional block processing | Not yet present; PR #9391 adds a full implementation | `processConditionalBlocks()` — supports `{{#if}}...{{/if}}` only |
| Variable default application | `applyDefaults()` — operates on `JsonNode` schema | Not yet present; PR #9392 adds `applyVariableDefaults()` operating on `VariableSchema` POJOs |
| Null-value semantics | `null` value → keep placeholder verbatim | `null` value → substitute empty string `""` |
| Value formatting | `formatValue()` — JSON-serializes lists/maps | `String.valueOf()` for everything |

## Why this matters

- **Feature drift.** Every new rendering feature (conditional blocks, defaults, future
  features like `{{#each}}` or nested blocks) must be implemented, tested, and maintained
  in both classes independently. History shows they diverge: the original `{{variable}}`
  regex disagreed across three components until #8977 unified it.
- **Behavioral inconsistency.** The same template rendered through the REST API vs. the
  MCP server can produce different output today (e.g. `null` value handling, conditional
  block support). Users and AI agents should not get different results depending on which
  render path they hit.
- **Conditional block gap.** PR #9391 adds `{{#if}}`/`{{else}}`/`{{#unless}}` support to
  REST, but the MCP converter only has simple `{{#if}}...{{/if}}`. Once #9391 merges, the
  parity problem flips: REST will support more syntax than MCP.

[...]

PR #9514, opened 2026-08-12
fix(schema-util): leave triple-brace placeholders alone instead of half-rendering them

### LINKED ISSUE #9513: PROMPT_TEMPLATE: {{{variable}}} is half-rendered, leaving stray braces in the output
### Description

Registry Version: snapshot (`main`, verified at `06279d4ac`)
Persistence type: not applicable — this is in template string processing and does not depend on the storage backend.

The canonical placeholder pattern in `PromptTemplateVariableUtil` matches a **two**-brace placeholder:

```java
VARIABLE_PATTERN = \{\{\s*(\w+)\s*\}\}
```

Given `{{{foo}}}`, the matcher cannot start at the first character (the third character is `{`, not a word character), so it shifts one character along and matches the inner `{{foo}}` instead. The value is substituted and the outer brace on each side is left behind as stray text.

The same thing happens to `{{{foo}}` — the value is substituted and a leading `{` is left in the output.

Because `extractVariableNames` uses that same pattern, it reports `foo` as a normal declared variable, so the template passes the validity rule cleanly and only then renders malformed.

### Steps to Reproduce

Render a `PROMPT_TEMPLATE` whose template text contains `{{{foo}}}`, with `foo` supplied as `bar`. Both render paths are affected — `PromptRenderingService.render(...)` and the MCP `PromptTemplateConverter`.

| Input | Expected | Actual |
|---|---|---|
| `Value is {{{foo}}}.` | left as written — triple-brace is not supported syntax | `Value is {bar}.` |
| `Value is {{{foo}}.` | left as written | `Value is {bar.` |
| `Value is {{foo}}}.` | `Value is bar}.` | `Value is bar}.` (already correct) |

Triple-brace is Handlebars' unescaped-output form. This report is **not** asking for it to be supported — only that an unsupported placeholder should be left alone rather than partially rewritten into malformed output.

### History (checked, not assumed)

This is long-standing behaviour on the MCP path. Reconstructing both renderers as they were before #8977:

```
pre-#8977 REST : Value is {{{foo}}}.   (preserved)
pre-#8977 MCP  : Value is {bar}.       (already half-rendered)
```

The MCP converter built one regex per argument key by concatenation, so `\{\{foo\}\}` matched the inner two braces. The REST renderer used `\{\{([^}]+)\}\}`, which read the name as `{foo`, found no such variable, and preserved the placeholder. #8977 unified both onto the single canonical pattern — a good change — and the REST path inherited this edge case with it.

So a fix restores the REST path's previous behaviour and changes MCP's, which has been this way from the start.

### Suggested fix

Add a negative lookbehind so the pattern cannot start one character inside a longer run of braces:

```java
VARIABLE_PATTERN = (?<!\{)\{\{\s*(\w+)\s*\}\}
```

Deliberately **not** paired with a trailing `(?!\})`: `{{foo}}}` is a placeholder followed by a literal closing brace, which already renders correctly as `bar}`, and adding the lookahead would break that case. I checked all four brace combinations before choosing.

Blast radius is three consumers of the shared pattern — `PromptRenderingService`, `PromptTemplateConverter`, and `PromptTemplateContentValidator`.

I have this implemented and tested and would like to open the PR.

### Related

- #8977 — introduced the shared canonical pattern
- #9213 / #9214 — UI *highlighting* of handlebars block tags, including triple-stache; this report is about server-side rendering, which those did not touch
- #9483 — the REST/MCP rendering unification; this is a small pattern-level fix and does not overlap that refactor

Aarti-panchal01:
Fixes #9513

## What was happening

`PromptTemplateVariableUtil.VARIABLE_PATTERN` matches a two-brace placeholder. Given `{{{name}}}` the matcher cannot start at the first character, so it shifts one along, matches the inner `{{name}}`, and substitutes — leaving the outer brace on each side in the output:

```
Value is {{{foo}}}.   ->   Value is {bar}.
Value is {{{foo}}.    ->   Value is {bar.
```

`extractVariableNames` shares the pattern, so it reported `foo` as an ordinary variable and the template passed the validity rule cleanly before rendering malformed.

## The change

One character class of a change — a negative lookbehind on the pattern:

```java
(?<!\{)\{\{\s*(\w+)\s*\}\}
```

Triple-brace is Handlebars' unescaped-output form, which this project does not implement. This PR does **not** add support for it; it only stops an unsupported placeholder being partially rewritten into malformed text.

## Scope decisions

**No trailing `(?!\})`.** I checked all four brace combinations before choosing. `{{foo}}}` is a placeholder followed by a literal closing brace and already renders correctly as `bar}` — which is also what Handlebars does. Adding a lookahead would have broken that case, so the fix is deliberately asymmetric.

**This changes long-standing MCP behaviour, and I want to be straight about that.** I reconstructed both renderers as they were before #8977 and ran them:

```
pre-#8977 REST : Value is {{{foo}}}.   (preserved)
pre-#8977 MCP  : Value is {bar}.       (already half-rendered)
```

The MCP converter built a regex per argument key by concatenation, so it always matched the inner braces. REST read the name as `{foo`, found no such variable and preserved the placeholder. #8977 unified both onto the shared pattern and REST inherited the edge case. So this restores the REST path's previous behaviour and changes MCP's, which has been this way since the beginning. The output it changes is malformed in every case I could construct, but it is a behaviour change on that path and reviewers should weigh it as one.

## Testing

Seven tests across three classes. **Five are genuinely red-first** — verified by reverting the pattern and re-running, not assumed:

```
testTripleBraceIsNotSubstituted                          expected: <Value is {{{foo}}}.> but was: <Value is {bar}.>
testTripleBraceIsNotExtractedAsAVariable                 expected: <[]>                  but was: <[foo]>
testUnbalancedLeadingBracesAreNotSubstituted             expected: <Value is {{{foo}}.>  but was: <Value is {bar.>
testPlaceholderAdjacentToTripleBraceIsStillSubstituted   expected: <{{{foo}}} value>     but was: <{value} value>
testFourOrMoreLeadingBracesAreNotSubstituted             expected: <{{{{foo}}}}>         but was: <{{value}}>
testTripleBracePlaceholderIsNotReportedAsUndefined       RuleViolationException (template rejected)
testUndefinedVariableAdjacentToTripleBraceIsStillReported 'foo' also reported as undefined
testRemovingAVariableOnlyInsideATripleBraceRunIsCompatible expected: <true> but was: <false>
```

`testAdjacentPlaceholdersAreBothSubstituted` and `testRemovingAVariableStillUsedInTheTemplateIsIncompatible` pass both before and after — they are guards, not fix tests, and are deliberately not counted above.

The pattern has **four** production consumers, not three as this description originally said — `PromptTemplateCompatibilityChecker` also shares the extractor. All four were run:

| Suite | Result |
|---|---|
| `PromptTemplateVariableUtilTest` | 23/23 |
| `PromptTemplateContentValidatorTest` (validity rule) | 26/26 |
| `PromptTemplateCompatibilityCheckerTest` (compatibility rule) | 12/12 |
| `PromptTemplateConverterTest` (MCP) | 22/22 |
| `PromptRenderingServiceTest` (REST) | 48/48 |

## Related

- #8977 introduced the shared canonical pattern this fixes an edge case in
- #9213 / #9214 covered handlebars highlighting in the UI viewer; this is the server-side render path, which they did not touch
- #9483 is the REST/MCP rendering unification. This is a pattern-level fix in `schema-util/common` and does not touch the rendering pipeline that issue is about — happy to hold it if a maintainer would rather it landed as part of that work.

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/content/util/PromptTemplateVariableUtil.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/content/util/PromptTemplateVariableUtil.java
@@ -50,8 +50,16 @@ public final class PromptTemplateVariableUtil {
     /**
      * The single pattern that defines what a prompt template variable placeholder looks like.
      * Group 1 is the variable name, with any surrounding whitespace already stripped.
+     *
+     * The leading <code>(?&lt;!\{)</code> keeps the pattern from starting one character inside a
+     * longer run of braces. Without it, <code>{{{name}}}</code> still contains <code>{{name}}</code>
+     * starting at the second character, so the placeholder matched, the value was substituted, and
+     * the outer brace on each side was left behind as stray text. Triple-brace is not supported
+     * syntax here, so it is left exactly as written instead of being half-rendered. Note this is
+     * deliberately not paired with a trailing <code>(?!\})</code>: <code>{{name}}}</code> is a
+     * placeholder followed by a literal closing brace, which already renders correctly.
      */
-    public static final Pattern VARIABLE_PATTERN = Pattern.compile("\\{\\{\\s*(\\w+)\\s*\\}\\}");
+    public static final Pattern VARIABLE_PATTERN = Pattern.compile("(?<!\\{)\\{\\{\\s*(\\w+)\\s*\\}\\}");
 
     /**
      * Handlebars tags that are spelled like a plain variable placeholder but are control syntax
--- a/schema-util/common/src/test/java/io/apicurio/registry/content/util/PromptTemplateVariableUtilTest.java
+++ b/schema-util/common/src/test/java/io/apicurio/registry/content/util/PromptTemplateVariableUtilTest.java
@@ -200,4 +200,38 @@ class PromptTemplateVariableUtilTest {
                 PromptTemplateVariableUtil.substituteVariables("Value is {{foo}}}.",
                         varName -> "bar"));
     }
+
+    /**
+     * A lookbehind that was written too broadly could swallow the placeholder that follows a
+     * triple-brace run. It must not: only the triple-brace itself is left alone.
+     */
+    @Test
+    void testPlaceholderAdjacentToTripleBraceIsStillSubstituted() {
+        Assertions.assertEquals("{{{foo}}} value",
+                PromptTemplateVariableUtil.substituteVariables("{{{foo}}} {{bar}}",
+                        varName -> "value"));
+        Assertions.assertEquals(List.of("bar"),
+                PromptTemplateVariableUtil.extractVariableNames("{{{foo}}} {{bar}}"));
+    }
+
+    @Test
+    void testFourOrMoreLeadingBracesAreNotSubstituted() {
+        Assertions.assertEquals("{{{{foo}}}}",
+                PromptTemplateVariableUtil.substituteVariables("{{{{foo}}}}", varName -> "value"));
+        Assertions.assertEquals(List.of(),
+                PromptTemplateVariableUtil.extractVariableNames("{{{{foo}}}}"));
+    }
+
+    /**
+     * Guard, not a fix: two placeholders written back to back share no brace between them, so the
+     * lookbehind must never see the first one's closing brace as part of the second's opening.
+     * Passes before and after the change.
+     */
+    @Test
+    void testAdjacentPlaceholdersAreBothSubstituted() {
+        Assertions.assertEquals("valuevalue",
+                PromptTemplateVariableUtil.substituteVariables("{{a}}{{b}}", varName -> "value"));
+        Assertions.assertEquals(List.of("a", "b"),
+                PromptTemplateVariableUtil.extractVariableNames("{{a}}{{b}}"));
+    }
 }
--- a/schema-util/common/src/test/java/io/apicurio/registry/content/util/PromptTemplateVariableUtilTest.java
+++ b/schema-util/common/src/test/java/io/apicurio/registry/content/util/PromptTemplateVariableUtilTest.java
@@ -160,4 +160,44 @@ class PromptTemplateVariableUtilTest {
         Assertions.assertFalse(PromptTemplateVariableUtil.isControlKeyword("Else"));
         Assertions.assertFalse(PromptTemplateVariableUtil.isControlKeyword(null));
     }
+
+    /*
+     * Triple-brace handling. The pattern matches two braces, so given {{{foo}}} it can still lock
+     * onto the inner {{foo}} and rewrite only that, leaving the outer brace on each side behind as
+     * stray text. Triple-brace is not supported syntax here, so the placeholder is left exactly as
+     * written rather than half-rendered.
+     */
+
+    @Test
+    void testTripleBraceIsNotSubstituted() {
+        Assertions.assertEquals("Value is {{{foo}}}.",
+                PromptTemplateVariableUtil.substituteVariables("Value is {{{foo}}}.",
+                        varName -> "bar"));
+    }
+
+    @Test
+    void testTripleBraceIsNotExtractedAsAVariable() {
+        Assertions.assertEquals(List.of(),
+                PromptTemplateVariableUtil.extractVariableNames("Value is {{{foo}}}."));
+    }
+
+    @Test
+    void testUnbalancedLeadingBracesAreNotSubstituted() {
+        Assertions.assertEquals("Value is {{{foo}}.",
+                PromptTemplateVariableUtil.substituteVariables("Value is {{{foo}}.",
+                        varName -> "bar"));
+    }
+
+    /**
+     * Regression guard, not a fix. In Handlebars {{foo}}} is a placeholder followed by a literal
+     * closing brace, so "bar}" is already the right answer and must stay that way. This test passes
+     * both before and after the change; it is here so a future tightening of the pattern cannot
+     * break the case silently.
+     */
+    @Test
+    void testTrailingBraceAfterPlaceholderIsPreserved() {
+        Assertions.assertEquals("Value is bar}.",
+                PromptTemplateVariableUtil.substituteVariables("Value is {{foo}}}.",
+                        varName -> "bar"));
+    }
 }
--- a/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/PromptTemplateCompatibilityCheckerTest.java
+++ b/schema-util/util-provider/src/test/java/io/apicurio/registry/rules/compatibility/PromptTemplateCompatibilityCheckerTest.java
@@ -221,4 +221,45 @@ class PromptTemplateCompatibilityCheckerTest {
                 Map.of());
         assertTrue(result.isCompatible(), "Adding an enum value should be compatible");
     }
+
+    /**
+     * Control for the pair below, and the first coverage this rule has had: a variable that is
+     * genuinely still referenced by the template cannot be dropped from the schema.
+     */
+    @Test
+    void testRemovingAVariableStillUsedInTheTemplateIsIncompatible() {
+        String existing = template("Hello {{name}} in {{style}}", BASE_VARS, "");
+        String proposed = template("Hello {{name}} in {{style}}",
+                "\"name\": { \"type\": \"string\", \"required\": true }", "");
+
+        CompatibilityExecutionResult result = checker.testCompatibility(
+                CompatibilityLevel.BACKWARD,
+                List.of(create(existing)),
+                create(proposed),
+                Map.of());
+        assertFalse(result.isCompatible(),
+                "Removing a variable the template still uses should be incompatible");
+    }
+
+    /**
+     * The compatibility checker shares the placeholder extractor with the validity rule, so
+     * tightening the pattern changes this path too: a name that appears only inside a triple-brace
+     * run is not a placeholder, so the template does not "use" it and dropping its declaration is
+     * compatible. Pinned because the previous behaviour - flagging it as removed-but-used - came
+     * from the same mis-match this change fixes.
+     */
+    @Test
+    void testRemovingAVariableOnlyInsideATripleBraceRunIsCompatible() {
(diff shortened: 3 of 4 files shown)


schema-util/common/src/main/java/io/apicurio/registry/content/util/PromptTemplateVariableUtil.java
```diff
@@ -50,8 +50,16 @@ public final class PromptTemplateVariableUtil {
...
+     * The leading <code>(?&lt;!\{)</code> keeps the pattern from starting one character inside a
+     * longer run of braces. Without it, <code>{{{name}}}</code> still contains <code>{{name}}</code>
+     * starting at the second character, so the placeholder matched, the value was substituted, and
+     * the outer brace on each side was left behind as stray text. Triple-brace is not supported
+     * syntax here, so it is left exactly as written instead of being half-rendered. Note this is
+     * deliberately not paired with a trailing <code>(?!\})</code>: <code>{{name}}}</code> is a
+     * placeholder followed by a literal closing brace, which already renders correctly.
      */
-    public static final Pattern VARIABLE_PATTERN = Pattern.compile("\\{\\{\\s*(\\w+)\\s*\\}\\}");
+    public static final Pattern VARIABLE_PATTERN = Pattern.compile("(?<!\\{)\\{\\{\\s*(\\w+)\\s*\\}\\}");
```
paoloantinori, line 62: `extractVariableNames` uses this same pattern, so with this change `{{{foo}}}` is no longer returned as a variable name. If the validity rule treats an extracted-but-undefined name as an error, a triple-brace placeholder now passes validation and renders through as literal text instead of being flagged. Confirm whether that pass-through is intended for the validity path, and consider a test in `PromptTemplateContentValidatorTest` that asserts the chosen behavior (reject versus pass-through) so this semantic change is locked in.

schema-util/common/src/test/java/io/apicurio/registry/content/util/PromptTemplateVariableUtilTest.java
```diff
@@ -160,4 +160,44 @@ void testIsControlKeyword() {
...
+
+    /*
+     * Triple-brace handling. The pattern matches two braces, so given {{{foo}}} it can still lock
+     * onto the inner {{foo}} and rewrite only that, leaving the outer brace on each side behind as
+     * stray text. Triple-brace is not supported syntax here, so the placeholder is left exactly as
+     * written rather than half-rendered.
+     */
+
+    @Test
+    void testTripleBraceIsNotSubstituted() {
```
paoloantinori, line 172: The new tests cover the core triple-brace cases well. Two easy adds would harden this: a valid placeholder adjacent to a triple-brace run, where `{{{foo}}} {{bar}}` should render to `{{{foo}}} value` (confirms the lookbehind does not suppress the following match), and a four-plus leading-brace case such as `{{{{foo}}}}`. Both render correctly under the new pattern today but are not covered, so a future over-tightening of the lookbehind could break them silently.

paoloantinori (commented):
Minimal and correct one-character-class fix; the `(?<!\{)` lookbehind stops the inner `{{name}}` in `{{{name}}}` from matching without suppressing any legitimate single placeholder, and omitting the trailing lookahead is the right asymmetric choice. Three red-first tests plus a regression guard back it well. Two points to settle before merge, both on the validity-extraction side rather than the rendering fix.

- The rendering change is sound. Tracing the new pattern against `{{{foo}}}`, the unbalanced `{{{foo}}`, `{{foo}}}`, adjacent placeholders `{{a}}{{b}}`, and a start-of-string `{{name}}`, all behave as intended, and no supported `{{name}}` placeholder is suppressed (PromptTemplateVariableUtil.java:62).
- `extractVariableNames` shares this pattern, so `{{{foo}}}` is no longer reported as a variable. If the validity rule flags extracted-but-undefined names, a triple-brace placeholder now passes validation silently and renders as literal text where it might previously have been rejected. Worth confirming the intended behavior and pinning it with a test in `PromptTemplateContentValidatorTest`.
- Test coverage gap. There is no test for a valid placeholder adjacent to a triple-brace run (`{{{foo}}} {{bar}}` should render to `{{{foo}}} value`) or for four-plus leading braces. Both behave correctly now but are not locked in, so a future tightening of the lookbehind could regress them silently.

Aarti-panchal01:
Thanks — both points addressed in `476c6307c`, and your first one led somewhere I had missed.

**Validity extraction: pass-through is intended, and it is now pinned.**

You are right that the semantics change. Before this PR, a template containing `{{{foo}}}` with no `foo` declared was *rejected* with `Template variable '{{foo}}' is used but not defined` — an error naming a placeholder the author never wrote, produced by the same mis-match the fix removes. After it, the run is literal text and the template validates.

I think pass-through is the correct end state, because it is what every other unsupported construct already does. `{{#each x}}`, a dotted `{{user.name}}` and a handlebars comment are all invisible to the extractor and none of them are reported. Flagging triple-brace would make it the only unsupported syntax that raises a violation. Two tests in `PromptTemplateContentValidatorTest`:

- `testTripleBracePlaceholderIsNotReportedAsUndefined` — the pass-through itself
- `testUndefinedVariableAdjacentToTripleBraceIsStillReported` — the flip side, so leaving triple-brace alone cannot blind the rule to a real undefined placeholder next to it

**The same change also affects the compatibility checker, which I had not accounted for.**

Chasing your point through the other callers, `PromptTemplateCompatibilityChecker` shares this extractor via `PromptTemplateContentValidator.extractTemplateVariables`. So a variable appearing only inside `{{{...}}}` is no longer "used", and dropping its declaration is no longer flagged `VARIABLE_REMOVED_BUT_USED`. That follows from the same reasoning — the name is literal text, so removing it breaks nothing — but it is a second behavioural change and it was unpinned in both directions, since that rule had no test coverage at all. Two tests added:

- `testRemovingAVariableOnlyInsideATripleBraceRunIsCompatible` — the new behaviour
- `testRemovingAVariableStillUsedInTheTemplateIsIncompatible` — the control, and the first coverage this rule has had

My original blast-radius note listed three consumers of the pattern. It is four.

**Test coverage gap: both cases added.**

- `testPlaceholderAdjacentToTripleBraceIsStillSubstituted` — `{{{foo}}} {{bar}}` renders to `{{{foo}}} value`, and extraction returns only `bar`
- `testFourOrMoreLeadingBracesAreNotSubstituted` — `{{{{foo}}}}` is left as written
- `testAdjacentPlaceholdersAreBothSubstituted` — `{{a}}{{b}}`, so the lookbehind can never read one placeholder's closing brace as part of the next one's opening

**Red-first, verified by reverting the pattern rather than assuming.** Of the seven tests now added, five fail against the unmodified pattern:

```
testTripleBraceIsNotSubstituted                     expected: <Value is {{{foo}}}.>  but was: <Value is {bar}.>
testTripleBraceIsNotExtractedAsAVariable            expected: <[]>                   but was: <[foo]>
testUnbalancedLeadingBracesAreNotSubstituted        expected: <Value is {{{foo}}.>   but was: <Value is {bar.>
testPlaceholderAdjacentToTripleBraceIsStillSubstituted  expected: <{{{foo}}} value>  but was: <{value} value>
testFourOrMoreLeadingBracesAreNotSubstituted        expected: <{{{{foo}}}}>          but was: <{{value}}>
testTripleBracePlaceholderIsNotReportedAsUndefined      RuleViolationException (template rejected)
testUndefinedVariableAdjacentToTripleBraceIsStillReported  'foo' also reported as undefined
testRemovingAVariableOnlyInsideATripleBraceRunIsCompatible  expected: <true> but was: <false>
```

The two I am *not* counting as red-first are `testAdjacentPlaceholdersAreBothSubstituted` and `testRemovingAVariableStillUsedInTheTemplateIsIncompatible` — both pass before and after and are guards rather than fix tests.

Suites: `PromptTemplateVariableUtilTest` 23, `PromptTemplateContentValidatorTest` 26, `PromptTemplateCompatibilityCheckerTest` 12, `PromptTemplateConverterTest` 22, `PromptRenderingServiceTest` 48.

paoloantinori:
Confirmed on `476c6307`: all three points landed, and the red-first verification (five tests failing against the unmodified pattern) is the right way to pin a behavioural change like this.

The compatibility-checker discovery is the catch that matters. My review only traced three consumers of `VARIABLE_PATTERN`; the fourth (`PromptTemplateCompatibilityChecker` via `extractTemplateVariables`) means `VARIABLE_REMOVED_BUT_USED` was silently affected too, and that rule had zero test coverage before. Pinning both directions of the new behaviour there closes a gap that existed independently of this fix.

Nothing further from me on this head.

Aarti-panchal01:
@EricWittmann this one is ready when you have a moment.

Rebased onto current `main` — new head `88b316e1`. The branch was showing out-of-date rather than conflicting: none of the 18 commits since my base touch any file in this PR, so it replayed with no changes to the diff. Still 4 files, +175/-1.

Re-verified locally on the new head, reinstalling `schema-util/common` first so the consumer modules were not testing against a stale jar:

- `PromptTemplateVariableUtilTest` — 23 passing
- `PromptTemplateContentValidatorTest` — 26 passing
- `PromptTemplateCompatibilityCheckerTest` — 12 passing

@paoloantinori signed off on the previous head (`476c6307`) and the only change since is the replay onto `main`.

paoloantinori:
Thanks for re-verifying on the rebased head with a fresh `schema-util/common` install first — that's the right way to make sure the consumer-module tests aren't hitting a stale jar. Since the diff replayed unchanged from the head I reviewed (`476c6307`), my prior comments still apply and there's nothing new for me to flag. Over to @EricWittmann for the accept.

Aarti-panchal01:
@EricWittmann — following up on this, since Paolo signed off a few days ago. Let me know if there's anything else needed from my side before it can land.

Aarti-panchal01:
Hi @jsenko — this has a full review from Paolo, and Eric self-assigned it back on Aug 12, but it's still waiting on the initial /accept to kick off the pipeline. Tagging you since I've seen you actively merging PRs this week — no rush, just flagging in case it fell off the radar with everything else going on.

Issue #9504, opened 2026-08-12
feat(storage): add webhook subscription and delivery log data model + SQL schema


karthik120710:
Description

## Context

Webhook subscriptions need to be persisted so that subscription configuration
(endpoint URL, event types, filters) survives restarts, and so that delivery
attempts can be tracked. At-least-once delivery semantics require a durable
delivery log with deduplication on the CloudEvent id and retry scheduling state.

This issue covers **only** the data model and schema. Delivery/dispatch logic,
REST API, and UI are out of scope and tracked separately.

## Note on migrations

This project does **not** use Liquibase. The SQL storage layer uses hand-written
DDL per database kind plus numbered upgrade scripts:

- Base schema: `app/src/main/resources/io/apicurio/registry/storage/impl/sql/{h2,mysql,mssql,postgresql}.ddl`
- Upgrades: `app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/<version>/<kind>.upgrade.ddl`
- Version marker: `app/src/main/resources/io/apicurio/registry/storage/impl/sql/db-version` (currently `107`),
  read by `AbstractSqlRegistryStorage#DB_VERSION`

So this work means: bump `db-version` to `108`, add `upgrades/108/*.upgrade.ddl`
for all four kinds, and add the same tables to the four base DDL files.

### `webhook_subscriptions`

| Column | Type | Notes |
|---|---|---|
| `subscriptionId` | VARCHAR(128) | PK (UUID string; portable across h2/mysql/mssql/postgres) |
| `name` | VARCHAR(512) | nullable, user-friendly label |
| `endpointUrl` | VARCHAR(1024) | NOT NULL, validated at the service layer |
| `eventTypes` | TEXT/CLOB | JSON array, e.g. `["ARTIFACT_CREATED","ARTIFACT_DELETED"]` |
| `groupFilter` | VARCHAR(512) | nullable, regex |
| `artifactIdFilter` | VARCHAR(512) | nullable, regex |
| `enabled` | BOOLEAN | NOT NULL, default true |
| `secret` | VARCHAR(512) | nullable, HMAC signing key |
| `createdBy` | VARCHAR(256) | nullable |
| `createdOn` | TIMESTAMP | NOT NULL |
| `modifiedOn` | TIMESTAMP | NOT NULL |

### `webhook_delivery_logs`

| Column | Type | Notes |
|---|---|---|
| `deliveryId` | VARCHAR(128) | PK |
[...]

PR #9521, opened 2026-08-12
feat(storage): add webhook subscription and delivery log tables (#3)

karthik120710:
Description
close - #9504 
## What

Adds the persistence layer for webhook subscriptions and webhook delivery
tracking. Schema and data model only — no dispatch logic, REST API, or UI.

Database version goes 107 → 108.

### Tables

`webhook_subscriptions`
- `subscriptionId` (PK), `name`, `endpointUrl` (NOT NULL), `eventTypes` (JSON array),
  `groupFilter`, `artifactIdFilter`, `enabled` (default true), `secret`,
  `createdBy`, `createdOn`, `modifiedOn`
- Indexes: `enabled`, `createdOn`

`webhook_delivery_logs`
- `deliveryId` (PK), `subscriptionId` (FK → subscriptions, ON DELETE CASCADE),
  `eventId`, `eventType`, `status`, `attemptCount` (default 0), `lastAttemptAt`,
  `nextRetryAt`, `errorMessage`, `httpStatusCode`, `createdOn`
- `UQ_webhook_delivery_logs_1 UNIQUE (subscriptionId, eventId)`
- Indexes: `subscriptionId`, `(status, nextRetryAt)`, `eventId`

### Files

- Base DDL for all four kinds: `{h2,postgresql,mysql,mssql}.ddl`
- New upgrade scripts: `upgrades/108/{h2,postgresql,mysql,mssql}.upgrade.ddl`
- `db-version` bumped to `108`
- DTOs: `WebhookSubscriptionDto`, `EditableWebhookSubscriptionDto`,
  `WebhookDeliveryLogDto`, `WebhookDeliveryStatus`
- Row mappers: `WebhookSubscriptionDtoMapper`, `WebhookDeliveryLogDtoMapper`

## Why

Webhook subscriptions must survive restarts, and at-least-once delivery needs a
durable record of attempts. The unique constraint on `(subscriptionId, eventId)`
is the deduplication key: an event fanned out to N subscriptions produces N rows,
but a redelivered event produces none, so a retry can never double-deliver.
`(status, nextRetryAt)` is indexed for the retry scheduler's scan.

## Design notes

Three deliberate deviations from the issue description:

1. **`status` is `VARCHAR(32)`, not a native SQL `ENUM`.** `ENUM` exists only in
   MySQL and PostgreSQL; MSSQL and H2 would need different DDL, breaking schema
   parity across the four dialects. Mapped to the `WebhookDeliveryStatus` Java enum.

2. **IDs are `VARCHAR(128)`, not a native UUID type.** PostgreSQL has `uuid`,
   MSSQL has `uniqueidentifier`, MySQL has neither. The codebase already uses
   VARCHAR string IDs throughout (e.g. `version_comments.commentId`).

3. **`eventTypes` is typed `Set<StorageEventType>`.** `StorageEventType` already
   exists and already contains `ARTIFACT_CREATED`, `ARTIFACT_DELETED`, etc., so
   reusing it avoids a second enum that would drift out of sync with the events
   the storage layer actually emits.

Additionally, `WebhookSubscriptionDto.secret` is annotated `@ToString.Exclude` so
the HMAC signing key cannot leak into logs; there is a test asserting this.

## Storage variants

No per-variant changes needed. `KafkaSqlRegistryStorage` delegates to
`SqlRegistryStorage`, and gitops/kubernetesops go through
`AbstractPollingRegistryStorage` → `BlueSqlStorage`/`GreenSqlStorage`, so all
four variants share this DDL. Variant-specific work begins when `RegistryStorage`
gains webhook CRUD operations (follow-up).

## Out of scope (follow-ups)

- `RegistryStorage` interface methods and CRUD implementation
- REST API, delivery worker, retry scheduler, UI
- Endpoint URL validation (belongs in the service layer)

#### Test plan

New tests:
- `WebhookSchemaTest` (9 tests, `@QuarkusTest`) — fresh-install path on H2:
  DTO/mapper round trip with specific field values, nullable columns,
  `httpStatusCode` NULL mapping to `null` rather than `0`, secret redaction in
  `toString()`, duplicate `(subscriptionId, eventId)` rejected naming
  `UQ_webhook_delivery_logs_1`, same `eventId` across two subscriptions allowed,
  orphan delivery log rejected by FK, cascade delete on subscription removal.
- `WebhookDdlUpgradeTest` (2 tests, plain JDBC) — applies the 108 upgrade script
  to a database seeded at version 107 on **H2** and **embedded PostgreSQL**,
  asserting `db_version` advances to 108 and all constraints behave.

Manual verification of the remaining two dialects (throwaway containers):
- **MySQL 8.4** — upgrade applies, `db_version` = 108, duplicate → error 1062
  naming `UQ_webhook_delivery_logs_1`, orphan → error 1452 naming
  `FK_webhook_delivery_logs_1`, cascade delete leaves 0 logs.
- **SQL Server 2022** — same assertions; duplicate → Msg 2627, orphan → Msg 547.

Checks:
- [x] `./mvnw checkstyle:check -pl app` — 0 violations
- [x] `WebhookSchemaTest`, `WebhookDdlUpgradeTest`, `DataUpgradeTest`,
      `SimpleDataUpgradeTest`, `AvroCanonicalHashUpgraderTest` — 26 tests pass
- [x] Verified on H2, PostgreSQL, MySQL, SQL Server


app/src/main/resources/io/apicurio/registry/storage/impl/sql/db-version
```diff
@@ -1 +1 @@
-109
\ No newline at end of file
+108
```
paoloantinori: This downgrades the recorded schema version from 109 to 108, and the same 109 to 108 change is made in all four base DDLs (h2, postgresql, mysql, mssql). The PR description says the move is 107 to 108, which does not match the diff. The version should increment past the current 109, not decrease. If this branch is behind main it needs a rebase; otherwise the application will advertise an older schema than databases already carry.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/108/h2.upgrade.ddl
```diff
@@ -5,4 +5,15 @@
 
 UPDATE apicurio SET propValue = 108 WHERE propName = 'db_version';
```
paoloantinori, line 7: Two problems here, and the same applies to the postgresql, mysql, and mssql variants. (1) The `UPGRADER:AvroCanonicalHashUpgrader;` directive is deleted, dropping an existing data migration that may still need to run on databases that have not reached 108. (2) The webhook DDL is appended to version 108, which existing deployments have already applied, so they will skip this script and never get the webhook tables. Introduce a new upgrade version (for example 110), put the webhook DDL there, and keep the AvroCanonicalHashUpgrader directive in its original script.

app/src/main/java/io/apicurio/registry/storage/dto/WebhookSubscriptionDto.java
```diff
@@ -0,0 +1,47 @@
...
+    private String name;
+    private String endpointUrl;
+    private Set<StorageEventType> eventTypes;
+    private String groupFilter;
+    private String artifactIdFilter;
+    @Builder.Default
+    private boolean enabled = true;
+    @ToString.Exclude
+    private String secret;
+    private String createdBy;
```
paoloantinori, line 47: `@ToString.Exclude` correctly keeps the HMAC secret out of logs, but the secret is still persisted in plaintext in `webhook_subscriptions.secret`. A database read (backup leak, SQL injection, replica) exposes every subscription signing key and lets an attacker forge deliveries. If the threat model assumes a protected database this is acceptable, but it should be a conscious decision. Consider encrypting the column at rest, consistent with how other secrets are handled.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/h2.ddl
```diff
@@ -123,3 +123,15 @@ CREATE INDEX IDX_schema_usage_1 ON schema_usage(globalId);
...
+ALTER TABLE webhook_subscriptions ADD PRIMARY KEY (subscriptionId);
+CREATE INDEX IDX_webhook_subs_1 ON webhook_subscriptions(enabled);
+CREATE INDEX IDX_webhook_subs_2 ON webhook_subscriptions(createdOn);
+
+CREATE TABLE webhook_delivery_logs (deliveryId VARCHAR(128) NOT NULL, subscriptionId VARCHAR(128) NOT NULL, eventId VARCHAR(256) NOT NULL, eventType VARCHAR(128) NOT NULL, status VARCHAR(32) NOT NULL, attemptCount INT NOT NULL DEFAULT 0, lastAttemptAt TIMESTAMP WITHOUT TIME ZONE, nextRetryAt TIMESTAMP WITHOUT TIME ZONE, errorMessage TEXT, httpStatusCode INT, createdOn TIMESTAMP WITHOUT TIME ZONE NOT NULL);
+ALTER TABLE webhook_delivery_logs ADD PRIMARY KEY (deliveryId);
+ALTER TABLE webhook_delivery_logs ADD CONSTRAINT UQ_webhook_delivery_logs_1 UNIQUE (subscriptionId, eventId);
+ALTER TABLE webhook_delivery_logs ADD CONSTRAINT FK_webhook_delivery_logs_1 FOREIGN KEY (subscriptionId) REFERENCES webhook_subscriptions(subscriptionId) ON DELETE CASCADE;
+CREATE INDEX IDX_webhook_delivery_logs_1 ON webhook_delivery_logs(subscriptionId);
+CREATE INDEX IDX_webhook_delivery_logs_2 ON webhook_delivery_logs(status, nextRetryAt);
```
paoloantinori: `IDX_webhook_delivery_logs_1 ON (subscriptionId)` is redundant. `UQ_webhook_delivery_logs_1 (subscriptionId, eventId)` is created with subscriptionId as the leading column, so it already serves every `WHERE subscriptionId = ?` lookup and backs the foreign key. The standalone index only adds write cost on each insert and update of delivery logs. Consider dropping it from all four dialects and the matching upgrade scripts.

paoloantinori (commented):
The table design, DTOs, and schema test coverage are solid, but the migration versioning is broken in a way that will prevent these tables from being created on any existing deployment, and an unrelated shipped data migration is deleted in the process. The version handling must be fixed before merge; the rest is well structured.

- The schema version is decremented, not incremented. `db-version` goes 109 to 108 (db-version:1), and every base DDL lowers `INSERT ... 'db_version', 109` to 108 (h2.ddl:7, postgresql.ddl:7, mysql.ddl:10, mssql.ddl:7). The description says the move is 107 to 108, which contradicts the diff. The version must increment past the current 109, not regress it.
- The webhook DDL is appended to the already-shipped 108 upgrade (upgrades/108/*.upgrade.ddl). Databases that have already applied version 108 will not re-run this script, so they will never get the webhook tables and the code will fail against them at runtime. This migration belongs in a new version (for example 110) with the base DDL bumped to match.
- The 108 upgrade loses its `UPGRADER:AvroCanonicalHashUpgrader;` directive (upgrades/108/h2.upgrade.ddl:7 and the other three dialects). That is an unrelated data migration being silently dropped.
- Security: the HMAC `secret` is stored in plaintext (webhook_subscriptions.secret VARCHAR(512)). A backup leak or SQL injection exposes every subscription signing key and lets an attacker forge deliveries. Confirm plaintext is acceptable for this threat model, or encrypt at rest consistent with other sensitive values.
- Simplify: `IDX_webhook_delivery_logs_1 ON (subscriptionId)` is redundant because `UQ_webhook_delivery_logs_1 (subscriptionId, eventId)` already leads with subscriptionId and backs the FK. Dropping it removes a write-amplifying index with no query cost.
- Minor consistency: `WebhookDeliveryLogDtoMapper` lets `WebhookDeliveryStatus.valueOf` throw a raw IllegalArgumentException on an unknown status, while `WebhookSubscriptionDtoMapper` wraps deserialization failures in RegistryStorageException. Worth unifying so a bad row does not surface as an unrelated exception type.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/db-version
karthik120710: updated version from 109 to 110

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/108/h2.upgrade.ddl
karthik120710: (1) UPGRADER:AvroCanonicalHashUpgrader; restored — upgrades/108/ for all four dialects (h2, postgresql, mysql, mssql) is now byte-identical to upstreammain (git diff refs/remotes/upstream/main -- 108 is empty), with the upgrader directive intact. Databases below 108 will still run that data migration.

(2) Webhook DDL moved to a new version 110 — exactly the version as suggested

app/src/main/java/io/apicurio/registry/storage/dto/WebhookSubscriptionDto.java
karthik120710: @paoloantinori  Do you want this PR to introduce at-rest encryption for webhook_subscriptions.secret, or should it ship plaintext under a documented trusted-database 
│ assumption and treat encryption as a follow-up?
│ Unless you say otherwise, I'll go with plaintext + documented threat model + write-only at the API boundary, since encryption here means introducing 
│ the project's first at-rest crypto mechanism and key-rotation story, and I don't think a schema PR is the right place to decide that. Say the word and
│ I'll do it in this PR instead — I'd want to change the column now, while the DDL is unreleased, so it doesn't cost a second db-version bump later. i need some clarification.

karthik120710:
@paoloantinori @EricWittmann i have addresed the feedback can u review my changes

paoloantinori:
Thanks for the rebase — I re-read the new head against my earlier review. The three migration blockers are all fixed correctly: the version now increments 109 to 110, `AvroCanonicalHashUpgrader;` is back in the 108 upgrade, and the webhook DDL lives in its own `upgrades/110/` script. Good.

One new issue that was not present at the old head: the version-110 slot is now claimed by three concurrently-open PRs, and that needs coordination before any of them merges.

- `upgrades/110/*.upgrade.ddl` is identical in name to the directories shipped by #9146 (outbox aggregateId widening) and #8538 (`artifact_structured_content` table + `StructuredContentUpgrader`). I checked both heads and each ships its own `upgrades/110/h2.upgrade.ddl` (and the other three dialects) alongside the same `db-version` bump to 110.
- Whichever of the three merges first wins the `upgrades/110/` directory on `main`. The second to merge hits a version-uniqueness check (or a merge conflict on the four `.upgrade.ddl` files), and the third silently shadows or conflicts again. At minimum the last feature in never gets its migration applied on upgraded databases.
- The fix is mechanical but needs a decision on ordering: each PR should target a distinct version. The natural resolution is for the first to land to take 110, and the other two to re-bump to 111 and 112 (with their four dialect files and base DDL `INSERT` values updated to match). Worth coordinating with @EricWittmann and the authors of #9146 and #8538 on who takes which number.

The two earlier non-blocking notes are still open and I will leave them as-is: the `secret` column is still plaintext (acceptable if the threat model assumes a protected database, just confirm it consciously), and `WebhookDeliveryLogDtoMapper` still lets `WebhookDeliveryStatus.valueOf` throw a raw `IllegalArgumentException` on a bad row, unlike `WebhookSubscriptionDtoMapper` which wraps deserialization failures in `RegistryStorageException`. Neither blocks the feature.

Happy to re-read once the version collision is resolved.

karthik120710:
@paoloantinori Thanks for re-reading. I confirmed the collision against both heads — #9146 and #8538 each ship `upgrades/110/{h2,mysql,mssql,postgresql}.upgrade.ddl` plus the same `db-version` 109 -> 110 bump as this PR, so all three are claiming the same slot.

Two details I found while checking, which I think constrain the fix:

1. It won't merge silently. All three PRs edit the same single line of `app/src/main/resources/io/apicurio/registry/storage/impl/sql/db-version` and add identically named files, so the second and third to merge get a conflict on `db-version` plus add/add conflicts on the four dialect scripts. The "last migration never gets applied" outcome is real but requires someone to resolve that conflict by keeping only their own side.

2. The numbers have to be handed out in merge order, not just made distinct. `CommonSqlStatements.databaseUpgrade` loops `for (version = fromVersion + 1; version <= toVersion; version++)` and passes `getResourceAsStream("upgrades/" + version + "/" + dbType() + ".upgrade.ddl")` straight into `parser.parse(input)` with no null check. A missing intermediate directory is an NPE during startup upgrade, not a skipped script. So whoever takes 111 cannot merge before 110 is on main, and 112 not before 111.

@EricWittmann how would you like to sequence the three? Either works for me:

- you assign 110/111/112 now, and each PR re-bumps and then waits for its predecessor to land; or
- a simple rule that the last to merge re-bumps to the next free number as part of its rebase.

I have no preference on which number this PR takes — the change on my side is 4 file renames plus the `db-version`, the four base-DDL `db_version` INSERT values, the `-- Upgrade Script from X to Y` headers, the `UPDATE apicurio SET propValue` statements, and 3 references in `WebhookDdlUpgradeTest`. I'll hold this PR rather than merge ahead of the others. Just let me know which number is mine and I'll push it.

On the two non-blocking notes: agreed, leaving both as-is for now. The plaintext `secret` column is a conscious choice under the assumption of a protected database — say the word if you'd rather it be encrypted at rest before this lands, since that changes the column definition and would be much cheaper to do now than in a follow-up migration.

paoloantinori:
Thanks for confirming the collision against both heads. The mechanism is exactly as you describe: all three PRs (#8538, #9146, and this one) bump `db-version` to 110 and ship `upgrades/110/{h2,mysql,mssql,postgresql}.upgrade.ddl`, so whichever merges first claims the version and the others land on a classpath-shadowed directory (last-merge wins on the directory contents, which means one feature's `ALTER TABLE` silently never runs). This is a real coordination blocker.

The clean resolution is for two of the three PRs to re-bump to 111 and 112 and rename their `upgrades/<N>/` directories. Since this PR's schema work is the most recently opened and #8538 / #9146 look to be further along, the pragmatic order is probably:

1. Let #8538 or #9146 take 110.
2. This PR re-bumps to 111: change `db-version` to `111`, rename `upgrades/110/` to `upgrades/111/`, and update the `WebhookDdlUpgradeTest` path references (`upgrades/110/` -> `upgrades/111/`).

I can't tell from here which of #8538 or #9146 is closest to merge, so the call on who keeps 110 is between the authors and @EricWittmann/@jsenko. Happy to re-read once the re-bump lands.

karthik120710:
@paoloantinori i have updated version to 111 and addressed all ur feedback can u take a look at my pr.

paoloantinori:
Re-read at head `ad6a241` after the rebase. The version collision with #9146 and #8538 is resolved: `db-version` now goes 109 to 111 and the four `upgrades/111/{h2,postgresql,mysql,mssql}.upgrade.ddl` scripts are consistent with each other and with the base DDLs. I checked the open-PR search and no other PR ships `upgrades/111`, so the slot is free.

One coordination point worth noting: the 111 upgrade scripts are headed "Upgrade Script from 110 to 111", which assumes `upgrades/110/` (currently in #9146) will have run first. If this PR merges before #9146, a deployment upgrading straight from 109 would see the runner look for `upgrades/110/` (absent from main and from this branch). Worth confirming the intended merge order, or adding a one-line note in the PR description that this depends on #9146 landing first. The `mergeable_state: behind` already flags that a rebase is needed either way.

The rest reads cleanly. The `@ToString.Exclude` on `secret` in both DTOs, the `wasNull` handling for `httpStatusCode` in the mapper, and the per-dialect DDL parity all look right. The test suite is thorough. The redundant-index guard in `WebhookDdlUpgradeTest#testNoDialectCreatesASubscriptionIdOnlyIndex` is a nice touch since the unique index already covers subscriptionId lookups.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/db-version
```diff
@@ -1 +1 @@
-109
\ No newline at end of file
+111
```
aclfe: Blocker. `main` is at 109 and there's no `upgrades/110/` anywhere, not on main and not on this branch. Upgrading a real 109 database hits `version=110`, gets a null stream, and NPEs out of startup.

Renumber to 110 and merge ahead of #9146 / #8538, or state the ordering dependency explicitly. A gap isn't survivable.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/upgrades/111/h2.upgrade.ddl
```diff
@@ -0,0 +1,18 @@
+-- *********************************************************************
+-- DDL for the Apicurio Registry - Database: h2
+-- Upgrade Script from 110 to 111
```
aclfe: "from 110 to 111", but 110 doesn't exist. Same header in the other three dialects.

app/src/test/java/io/apicurio/registry/noprofile/storage/WebhookDdlUpgradeTest.java
```diff
@@ -0,0 +1,254 @@
...
+                            && normalized.contains("ONWEBHOOK_DELIVERY_LOGS(SUBSCRIPTIONID)"),
+                    "Redundant subscriptionId-only index in " + resource + ": " + sql);
+        }
+    }
+
+    private void seedVersion110(Connection connection) throws SQLException {
+        try (Statement statement = connection.createStatement()) {
+            statement.execute("CREATE TABLE apicurio (propName VARCHAR(255) NOT NULL, "
+                    + "propValue VARCHAR(255), PRIMARY KEY (propName))");
+            statement.execute("INSERT INTO apicurio (propName, propValue) VALUES ('db_version', 110)");
```
aclfe: This seeds a version that has never existed and may never exist, then hand-picks one script to apply. So it proves the upgrade works from a state no database will ever be in. This is the test that should have caught the missing 110, and instead it bakes the gap in.

Seed the real prior version and run it through `CommonSqlStatements.databaseUpgrade(fromVersion, DB_VERSION)` instead of applying a single script by path.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/h2.ddl
```diff
@@ -123,3 +123,14 @@ CREATE INDEX IDX_schema_usage_1 ON schema_usage(globalId);
 CREATE INDEX IDX_schema_usage_2 ON schema_usage(clientId);
 CREATE INDEX IDX_schema_usage_3 ON schema_usage(eventTimestamp);
 
+CREATE TABLE webhook_subscriptions (subscriptionId VARCHAR(128) NOT NULL, name VARCHAR(512), endpointUrl VARCHAR(1024) NOT NULL, eventTypes TEXT NOT NULL, groupFilter VARCHAR(512), artifactIdFilter VARCHAR(512), enabled BOOLEAN NOT NULL DEFAULT TRUE, secret VARCHAR(512), createdBy VARCHAR(256), createdOn TIMESTAMP WITHOUT TIME ZONE NOT NULL, modifiedOn TIMESTAMP WITHOUT TIME ZONE NOT NULL);
+ALTER TABLE webhook_subscriptions ADD PRIMARY KEY (subscriptionId);
+CREATE INDEX IDX_webhook_subs_1 ON webhook_subscriptions(enabled);
+CREATE INDEX IDX_webhook_subs_2 ON webhook_subscriptions(createdOn);
+
+CREATE TABLE webhook_delivery_logs (deliveryId VARCHAR(128) NOT NULL, subscriptionId VARCHAR(128) NOT NULL, eventId VARCHAR(256) NOT NULL, eventType VARCHAR(128) NOT NULL, status VARCHAR(32) NOT NULL, attemptCount INT NOT NULL DEFAULT 0, lastAttemptAt TIMESTAMP WITHOUT TIME ZONE, nextRetryAt TIMESTAMP WITHOUT TIME ZONE, errorMessage TEXT, httpStatusCode INT, createdOn TIMESTAMP WITHOUT TIME ZONE NOT NULL);
```
aclfe: No lease or claim column here (`lockedBy`, `leaseUntil`, or an optimistic version). The retry scheduler scans `(status, nextRetryAt)`, so with two replicas both pick up the same row and both POST.

The description says the unique key means "a retry can never double-deliver". That's about duplicate rows, not duplicate HTTP requests. Nothing arbitrates the actual send. Either add the lease columns now or say dispatch is single-writer. Adding them later is another migration in this same contended numbering space.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/h2.ddl
```diff
@@ -123,3 +123,14 @@ CREATE INDEX IDX_schema_usage_1 ON schema_usage(globalId);
 CREATE INDEX IDX_schema_usage_2 ON schema_usage(clientId);
 CREATE INDEX IDX_schema_usage_3 ON schema_usage(eventTimestamp);
 
+CREATE TABLE webhook_subscriptions (subscriptionId VARCHAR(128) NOT NULL, name VARCHAR(512), endpointUrl VARCHAR(1024) NOT NULL, eventTypes TEXT NOT NULL, groupFilter VARCHAR(512), artifactIdFilter VARCHAR(512), enabled BOOLEAN NOT NULL DEFAULT TRUE, secret VARCHAR(512), createdBy VARCHAR(256), createdOn TIMESTAMP WITHOUT TIME ZONE NOT NULL, modifiedOn TIMESTAMP WITHOUT TIME ZONE NOT NULL);
+ALTER TABLE webhook_subscriptions ADD PRIMARY KEY (subscriptionId);
+CREATE INDEX IDX_webhook_subs_1 ON webhook_subscriptions(enabled);
```
aclfe: Both indexes are on the table that'll hold maybe dozens of rows, and `(enabled)` is a boolean with near-zero selectivity (MySQL will ignore it).

`webhook_delivery_logs` grows as events x subscriptions forever and has no index on `createdOn`, so the eventual retention purge (`DELETE ... WHERE createdOn < ?`) full-scans the biggest table in the schema. Suggest dropping both of these and adding `createdOn` on the delivery log.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/h2.ddl
```diff
@@ -123,3 +123,14 @@ CREATE INDEX IDX_schema_usage_1 ON schema_usage(globalId);
 CREATE INDEX IDX_schema_usage_2 ON schema_usage(clientId);
 CREATE INDEX IDX_schema_usage_3 ON schema_usage(eventTimestamp);
 
+CREATE TABLE webhook_subscriptions (subscriptionId VARCHAR(128) NOT NULL, name VARCHAR(512), endpointUrl VARCHAR(1024) NOT NULL, eventTypes TEXT NOT NULL, groupFilter VARCHAR(512), artifactIdFilter VARCHAR(512), enabled BOOLEAN NOT NULL DEFAULT TRUE, secret VARCHAR(512), createdBy VARCHAR(256), createdOn TIMESTAMP WITHOUT TIME ZONE NOT NULL, modifiedOn TIMESTAMP WITHOUT TIME ZONE NOT NULL);
```
aclfe: Plaintext `secret` is still open from the first review and the decision you asked for never came back. Worth forcing an answer before this lands. It changes the column definition in all four dialects and is much cheaper now than in a follow-up migration.

Also: the DTO has `@Getter` + `@RegisterForReflection` and the mapper does `SELECT *`, so the secret loads on every read. `@ToString.Exclude` covers logging, not one careless reuse of this DTO in a REST response.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/mysql.ddl
```diff
@@ -272,3 +272,39 @@ CREATE TABLE schema_usage (
 CREATE INDEX IDX_schema_usage_1 ON schema_usage(globalId);
 CREATE INDEX IDX_schema_usage_2 ON schema_usage(clientId);
 CREATE INDEX IDX_schema_usage_3 ON schema_usage(eventTimestamp);
+
+CREATE TABLE webhook_subscriptions (
+    subscriptionId   VARCHAR(128)  NOT NULL,
+    name             VARCHAR(512) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci,
+    endpointUrl      VARCHAR(1024) NOT NULL,
```
aclfe: `endpointUrl` and `secret` inherit the table's `ascii` default, while `name` and the filters got explicit `utf8mb4`. An IDN endpoint URL or a non-ASCII secret inserts fine on H2/PG/MSSQL and fails with error 1366 on MySQL only.

`createdBy` staying ascii is fine, matches the existing `owner`/`modifiedBy` columns.

app/src/main/java/io/apicurio/registry/storage/impl/sql/mappers/WebhookDeliveryLogDtoMapper.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    @Override
+    public WebhookDeliveryLogDto map(ResultSet rs) throws SQLException {
+        // wasNull() reflects the most recent column read, so it must be captured immediately.
+        int httpStatusCodeValue = rs.getInt("httpStatusCode");
+        Integer httpStatusCode = rs.wasNull() ? null : httpStatusCodeValue;
+        return WebhookDeliveryLogDto.builder().deliveryId(rs.getString("deliveryId"))
+                .subscriptionId(rs.getString("subscriptionId")).eventId(rs.getString("eventId"))
+                .eventType(rs.getString("eventType"))
+                .status(WebhookDeliveryStatus.valueOf(rs.getString("status")))
```
aclfe: Raw `IllegalArgumentException` on a bad row, while `WebhookSubscriptionDtoMapper` wraps in `RegistryStorageException`. Flagged in the first review and acknowledged as "leaving as-is". It's a two-line change, please just do it instead of carrying it as a known inconsistency.

app/src/main/java/io/apicurio/registry/storage/dto/WebhookDeliveryLogDto.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+import java.util.Date;
+
+/**
+ * Read-only data transfer object representing the delivery record for a single CloudEvent sent to a single
```
aclfe, line 14: Javadoc says "Read-only" but the class has `@Setter` (line 23). Pick one.

app/src/test/java/io/apicurio/registry/noprofile/storage/WebhookDdlUpgradeTest.java
```diff
@@ -0,0 +1,254 @@
...
+        assertDeliveryLogIndexesAreNotRedundant(connection);
+    }
+
+    /**
+     * The unique (subscriptionId, eventId) index has subscriptionId as its leading column, so it already
+     * serves subscriptionId lookups. A subscriptionId-only index declared by the DDL would only add write
+     * cost. Indexes the database creates on its own (H2, for instance, adds one to back the foreign key) are
+     * ignored - only the explicitly declared IDX_ indexes are asserted on.
+     */
+    private void assertDeliveryLogIndexesAreNotRedundant(Connection connection) throws SQLException {
```
aclfe: Dropping the redundant `subscriptionId`-only index was right, `UQ_webhook_delivery_logs_1` leads with `subscriptionId` and backs the FK. But freezing that as a negative assertion in two tests over-fits. It'll fire on any future legitimate index change. The DDL diff is a better place to review this.

aclfe (commented):
Schema and DTOs look good. Version number is a blocker though.

109 -> 111 with no `upgrades/110/` means every existing deployment NPEs on startup. `CommonSqlStatements.java:56-59` loops `fromVersion+1 .. toVersion` and hands `getResourceAsStream(...)` straight to `DdlParser.parse` with no null check. A missing directory isn't a skipped script, it's an NPE out of `initialize()`. You worked this out yourself earlier in the thread and then pushed 111 anyway. So this can only merge after whoever takes 110 lands, and nothing enforces that.

CI won't catch it. Every job in `verify-integration-tests.yaml` starts from an empty DB, so it always takes the base-DDL path and the upgrade loop never runs. Green checks don't mean anything here.

Two asks:
- go back to 110 and merge first, or spell out the ordering dependency
- add a guard test: assert `upgrades/` is contiguous `101..DB_VERSION` across all four dialects, and that each script's `UPDATE ... propValue = N` matches its directory. Would have caught this. Will catch the same thing in #9146 / #8538.

PR description is stale too. Still says "107 -> 108", `upgrades/108/`, test plan seeding 107. And the checked boxes claim four databases verified, which predates two DDL rewrites.

On #9504: the issue is wrong, and that's what caused the 108 clobber. It says db-version is "currently 107". It was already 109 when the issue was filed (108 landed 2026-07-30, 109 on 2026-08-03). It also asks for "Apache 2.0 license header on all new Java files", which contradicts CLAUDE.md ("do not add a per-file Apache license header"). You were right to skip it, just list it with your other deviations. Worth fixing the issue so nobody else walks into it.

Good parts: base DDL and upgrade scripts are byte-identical across all four dialects, the mapper follows `ContractRuleDtoMapper` exactly, and the `wasNull()` handling is right and tested. Reusing `StorageEventType` is a better call than your description makes it sound. `ArtifactCreated.getType()` already returns `StorageEventType.name()` and the CloudEvent id is a UUID, so these column widths line up with the existing outbox pipeline for free.

Rest is inline.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/db-version
karthik120710: roleback from 111 to 110

app/src/test/java/io/apicurio/registry/noprofile/storage/WebhookDdlUpgradeTest.java
karthik120710: refactor webhook upgrade test to use CommonSqlStatements and dynamic versioning

app/src/main/resources/io/apicurio/registry/storage/impl/sql/h2.ddl
karthik120710: Added lockedBy and leaseUntil columns to webhook_delivery_logs so multiple replicas can claim rows for retry without double-sending.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/h2.ddl
karthik120710: implemented at-rest AES-256-GCM encryption for webhook HMAC secrets and prevented them from leaking in REST responses.

app/src/main/resources/io/apicurio/registry/storage/impl/sql/mysql.ddl
karthik120710: Fixed the MySQL webhook_subscriptions DDL so endpointUrl and secret support non-ASCII characters by adding explicit CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci (the table's default was ascii).

app/src/main/java/io/apicurio/registry/storage/impl/sql/mappers/WebhookDeliveryLogDtoMapper.java
karthik120710: Wrapped the WebhookDeliveryStatus.valueOf call in WebhookDeliveryLogDtoMapper so an unknown status is reported as RegistryStorageException instead of a raw IllegalArgumentException, matching the style of WebhookSubscriptionDtoMapper.

app/src/main/java/io/apicurio/registry/storage/dto/WebhookDeliveryLogDto.java
karthik120710: removed the @Setter annotation from WebhookDeliveryLogDto so the class is actually read-only, matching its Javadoc.

karthik120710:
@paoloantinori @aclfe i have addressed all ur suggetions and downgrade the version from 111 to 110 can u take a look at my pr , is it okay to merge .

paoloantinori:
Re-read at the new head (`99078f3`). Everything from the last round is in and verified: `WebhookDeliveryLogDto` is now `@Getter`-only, secrets are AES-256-GCM encrypted at rest (`WebhookSecretEncryption`), `lockedBy`/`leaseUntil` lease columns are on `webhook_delivery_logs`, the `createdOn` index is there, the MySQL charset fix landed, the delivery-status mapper wraps in `RegistryStorageException`, and the upgrade test now seeds real version 109 and runs through `CommonSqlStatements.databaseUpgrade` instead of hand-picking a script. Nice work addressing all of that.

One thing on the downgrade 111 -> 110: #9146 and #8538 are both still open, and both ship `upgrades/110/{h2,postgresql,mysql,mssql}.upgrade.ddl` plus `db-version` 110. We went to 111 two days ago precisely because of that collision, and nothing has merged since to resolve it. As aclfe framed it on this PR, 110 is fine only if this merges ahead of both of them; otherwise whichever lands second fails the version uniqueness (or silently shadows the other's upgrade script on the classpath).

So: not "okay to merge" unconditionally. Either (a) keep 110 and make the ordering explicit — a note in the description that this must merge before #9146 and #8538, and both of those need re-bumping to 111, or (b) go back to 111 and let this wait behind them. Happy either way as long as the sequencing is written down somewhere the other authors will see it. Everything else looks good from my side.

karthik120710:
@paoloantinori i have dowgraded version from 111 to 110 @aclfe mentioned it , i think getting back to 111 is a good idea is that okay .

paoloantinori:
Yes, restoring 111 is the right target, but the timing matters, so here is the current state as of now:

- main is still at db-version 109.
- #9146 and #8538 are both still open, and both ship upgrades/110 with the same 109 -> 110 bump. So the three-way collision is still live.
- Your current head is self-consistent (db-version 110 with upgrades/110 present), which is why 110 works standalone.

The constraint on going to 111: CommonSqlStatements.databaseUpgrade loops fromVersion+1 .. toVersion and loads every intermediate upgrades/<n> script. If you bump to 111 while main is still at 109, an existing deployment upgrading to this build would look for upgrades/110, not find it on your branch, and fail at startup — that is the NPE @aclfe flagged earlier, and it applies again the moment 110 is not on main yet.

So the practical sequencing is: keep 110 for now, and re-bump to 111 (dropping your 110 dir, keeping only 111) at the point where one of #9146/#8538 has actually landed and taken 110 on main. Whichever of the three merges first takes 110; the other two re-bump. If yours merges first, the re-bump burden moves to the other two instead.

Happy to re-read once you settle on the final number — everything else from the last round was already verified in place.

karthik120710:
@paoloantinori okay i will stay with 110 when those pr got merger i will update to 111. Thanks for ur guidance .

paoloantinori:
Sounds good — staying at 110 for now and re-bumping to 111 once #9146 / #8538 land keeps you out of the collision window cleanly.

One note for when you re-bump: move the db-version bump and all four dialect scripts (h2, postgresql, mysql, mssql) to `upgrades/111/` together, same as you did for the 110 layout.

Ping me here when that lands and I'll re-read.

karthik120710:
@paoloantinori there is no activity in #9146 what can i do now should i proceed with version 110 or should i have to update the version to 111 because these pr is open i am unable to work on another issue .

carlesarnal:
I'm closing this because this scope is reserved for the term.

PR #9524, opened 2026-08-12
fix(ui): honor declared minimum/maximum on numeric prompt template variables

### LINKED ISSUE #9518: [Bug]: Prompt Template test panel ignores declared minimum/maximum on numeric variables
## Description

The Prompt Template variable schema supports optional `minimum` and `maximum` values for numeric variables. The Test Panel in the UI, however, does not use these values when rendering the input field for a numeric variable , the field accepts any number, with no visual indication of the allowed range. The value is only validated after clicking "Render" and getting a response from the server.

### Environment
Local dev setup: UI via `npm run dev`, backend via `../mvnw quarkus:dev -Dlocal`, tested in Chrome.

### Steps to Reproduce
1. Create a `PROMPT_TEMPLATE` artifact with a numeric variable that declares `minimum` and `maximum`, for example:
```json
   { "rating": { "type": "integer", "minimum": 1, "maximum": 10 } }
```
2. Open the artifact's version page and go to the Test Panel.
3. Enter a value outside the declared range (e.g. `999`) into the field.
4. Click "Render".

**Expected**: The input should reflect the declared range - for example, via `min`/`max` on the number field and/or helper text showing "Must be between 1 and 10" — so the user sees the constraint before submitting.

**Actual**: The field accepts any value with no indication of the allowed range. The out-of-range value is only rejected after the render request completes.

### Logs
N/A - no runtime error occurs. This is a UI gap, not a crash: the declared `minimum`/`maximum` values are simply not read in `PromptTemplateTestPanel.tsx`.

### Related issues
Related to #9361, which covers the read-only Variables table shown in the Documentation tab (`PromptTemplateViewer.tsx`) not displaying the numeric range. This issue is about a different component - the interactive Test Panel input field (`PromptTemplateTestPanel.tsx`) - where the range is neither displayed nor enforced while entering a test value.

---

Shreya2005-2005:
Fixes #9518

## What

Adds `min`/`max` support and a range hint to numeric fields in the Prompt Template Test Panel.

## Why

A numeric variable's schema can declare `minimum` and `maximum`, but the Test Panel's number input never read those values. A user could type any number and only find out it was out of range after clicking "Render" and getting a response back from the server.

## How

- Added a `describeNumericRange` helper in `PromptTemplateTestPanel.utils.ts` that turns a `minimum`/`maximum` pair into a plain-language message, e.g. "Must be between 1 and 10".
- Passed `variable.minimum` and `variable.maximum` into the numeric `TextInput`'s `min`/`max` attributes, so the browser itself enforces the range.
- Show the range message below the field using PatternFly's `HelperText`/`HelperTextItem`, matching the pattern already used in `PropertyInput.tsx`.

## Testing

- Added 4 unit tests for `describeNumericRange`, covering: both bounds set, only minimum, only maximum, and neither bound set.
- All existing tests in `PromptTemplateTestPanel.utils.test.ts` still pass (12/12).

### DIFF
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -29,7 +29,7 @@ import {
 } from "./promptTemplateVariables";
 import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
 import { RenderPromptResponse, RenderPromptValidationError } from "@models/RenderPromptResponse.ts";
-import { coerceEnumValue, describeNumericRange } from "./PromptTemplateTestPanel.utils";
+import { coerceEnumValue, describeNumericRange, findOutOfRangeErrors } from "./PromptTemplateTestPanel.utils";
 
 export type PromptTemplateTestPanelProps = {
     groupId: string;
@@ -123,6 +123,16 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
     };
 
     const doRender = (): void => {
+        const rangeFields = reconciledVariables.map((entry) => {
+            const schema = schemaForField(entry);
+            return { name: entry.name, type: schema.type, minimum: schema.minimum, maximum: schema.maximum };
+        });
+        const outOfRangeErrors = findOutOfRangeErrors(rangeFields, values);
+        if (outOfRangeErrors.length > 0) {
+            setValidationErrors(outOfRangeErrors);
+            return;
+        }
+
         setIsLoading(true);
         setError("");
         setValidationErrors([]);
@@ -200,6 +210,7 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                             value={values[name] ?? ""}
                             min={variable.minimum}
                             max={variable.maximum}
+                            step={type === "integer" ? undefined : "any"}
                             onChange={(_event, val) => {
                                 const n = type === "integer" ? parseInt(val) : parseFloat(val);
                                 setValue(name, isNaN(n) ? "" : n);
--- a/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
+++ b/ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
@@ -10,6 +10,8 @@ import {
     CardTitle,
     Checkbox,
     Form,
+    HelperText,
+    HelperTextItem,
     FormGroup,
     FormSelect,
     FormSelectOption,
@@ -27,7 +29,7 @@ import {
 } from "./promptTemplateVariables";
 import { GroupsService, useGroupsService } from "@services/useGroupsService.ts";
 import { RenderPromptResponse, RenderPromptValidationError } from "@models/RenderPromptResponse.ts";
-import { coerceEnumValue } from "./PromptTemplateTestPanel.utils";
+import { coerceEnumValue, describeNumericRange } from "./PromptTemplateTestPanel.utils";
 
 export type PromptTemplateTestPanelProps = {
     groupId: string;
@@ -189,18 +191,29 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
                 );
             }
             case "integer":
-            case "number":
+            case "number": {
+                const rangeHint = describeNumericRange(variable.minimum, variable.maximum);
                 return (
-                    <TextInput
-                        type="number"
-                        value={values[name] ?? ""}
-                        onChange={(_event, val) => {
-                            const n = type === "integer" ? parseInt(val) : parseFloat(val);
-                            setValue(name, isNaN(n) ? "" : n);
-                        }}
-                        aria-label={name}
-                    />
+                    <>
+                        <TextInput
+                            type="number"
+                            value={values[name] ?? ""}
+                            min={variable.minimum}
+                            max={variable.maximum}
+                            onChange={(_event, val) => {
+                                const n = type === "integer" ? parseInt(val) : parseFloat(val);
+                                setValue(name, isNaN(n) ? "" : n);
+                            }}
+                            aria-label={name}
+                        />
+                        {rangeHint && (
+                            <HelperText>
+                                <HelperTextItem>{rangeHint}</HelperTextItem>
+                            </HelperText>
+                        )}
+                    </>
                 );
+            }
             case "array":
             case "object":
                 return (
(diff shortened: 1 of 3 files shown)


ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -189,18 +191,29 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
...
-                            setValue(name, isNaN(n) ? "" : n);
-                        }}
-                        aria-label={name}
-                    />
+                    <>
+                        <TextInput
+                            type="number"
+                            value={values[name] ?? ""}
+                            min={variable.minimum}
+                            max={variable.maximum}
```
paoloantinori, line 212: min/max on an HTML number input don't block out-of-range values; they only populate validityState for constraint validation. The onChange a few lines below still does `setValue(name, ...)` with whatever was typed, and nothing in this diff gates the Render action on checkValidity(), so a user can still type 999 against max=10 and only hit the error at Render. That is the same flow #9518 describes. To actually enforce this client-side, clamp in onChange or validate before Render; otherwise I'd soften the PR description's 'the browser itself enforces the range', since today it only hints and sets validityState.

ui/ui-app/src/app/components/promptTemplate/PromptTemplateTestPanel.tsx
```diff
@@ -189,18 +191,29 @@ export const PromptTemplateTestPanel: FunctionComponent<PromptTemplateTestPanelP
...
-                        value={values[name] ?? ""}
-                        onChange={(_event, val) => {
-                            const n = type === "integer" ? parseInt(val) : parseFloat(val);
-                            setValue(name, isNaN(n) ? "" : n);
-                        }}
-                        aria-label={name}
-                    />
+                    <>
+                        <TextInput
+                            type="number"
```
paoloantinori, line 209: For the `type === "number"` (float) branch this input gets min/max but no step, so the browser applies its default step=1 and marks in-range fractional values like 2.5 as a stepMismatch (and the spinner steps by 1). The integer branch is fine with the default. Consider `step={type === "integer" ? undefined : "any"}` so valid decimals aren't flagged invalid.

paoloantinori (commented):
The range hint and min/max wiring are clean, narrowly scoped, and well tested, and `describeNumericRange` is nicely covered (PromptTemplateTestPanel.utils.test.ts:38-51) with no reuse or simplification concerns. Two correctness gaps remain in how min/max actually behave on the number input itself.

- `PromptTemplateTestPanel.tsx:201-202`: min/max attributes do not prevent out-of-range input; the onChange at :205 still stores whatever was typed and nothing in the diff gates Render on checkValidity(), so the #9518 'discover it at Render' path is only partially closed.
- `PromptTemplateTestPanel.tsx:199`: the float (`type === "number"`) branch gets min/max with no step, so the browser's default step=1 flags in-range decimals like 2.5 as a stepMismatch; add `step="any"` for the number case.

Shreya2005-2005:
@paoloantinori Fixed min/max only set validity state, they don't stop submission, so I added findOutOfRangeErrors (with unit tests) that now checks the values and blocks Render if anything's out of range, instead of only relying on the browser. Also added step={type === "integer" ? undefined : "any"} so valid decimals like 2.5 don't get flagged as invalid anymore. Pushed in the latest commit.

paoloantinori:
Thanks, that's exactly the gap — browser `min`/`max` only set validity state, they don't block the action. `findOutOfRangeErrors` closing the loop on the Render path is the right call, and the `step="any"` on the number type fixes the decimal false-positive (2.5 was being rejected). Unit coverage on the new helper is solid too. Looks good from my side.

Shreya2005-2005:
> Thanks, that's exactly the gap — browser `min`/`max` only set validity state, they don't block the action. `findOutOfRangeErrors` closing the loop on the Render path is the right call, and the `step="any"` on the number type fixes the decimal false-positive (2.5 was being rejected). Unit coverage on the new helper is solid too. Looks good from my side.

@paoloantinori Thanks for confirming Let me know if anything else needs changing , happy to fix. Also, could you approve the workflow run when you get a chance so the checks can run?

paoloantinori:
Nothing else needed from my side. The `findOutOfRangeErrors` guard on the Render path and the `step="any"` fix both land the points we discussed, and the unit coverage on the new helper is solid.

On the workflow run: that approval is a repo-admin action separate from this code review, so it is not something I can trigger from here. Any other maintainer with admin rights can approve it; once CI is green this looks ready to merge.

EricWittmann:
@Shreya2005-2005 Sorry but this PR now has some merge conflicts!  Can you take a look at those?

EricWittmann (commented):
## Code Review

### Critical: Merge Conflicts / Stale Branch

The PR is based on a version of the code **before a significant refactoring** that has since landed on `main`. GitHub reports the merge state as `CONFLICTING`.

Specifically:

1. **`doRender` moved to a custom hook.** The PR modifies `doRender` inline in `PromptTemplateTestPanel.tsx`, but on `main`, `doRender` now lives in `usePromptTemplateTestPanelState.ts`. The `findOutOfRangeErrors` validation logic needs to move there.

2. **Import statements have diverged.** The PR patches old import lines that no longer exist on `main`. For example, the PR modifies:
   ```ts
   -import { coerceEnumValue } from "./PromptTemplateTestPanel.utils";
   +import { coerceEnumValue, describeNumericRange, findOutOfRangeErrors } from "./PromptTemplateTestPanel.utils";
   ```
   But on `main`, that import is now:
   ```ts
   import { coerceEnumValue, schemaForField, toDeclaredMap } from "./PromptTemplateTestPanel.utils";
   ```

3. **Test file imports also conflict** — same pattern, the existing import block has been expanded since the PR was branched.

The PR needs a rebase onto current `main`, with the `findOutOfRangeErrors` call placed inside the hook's `doRender` method rather than the component.

---

### Code Quality (the parts that are good)

The utility functions themselves are well-written:

- **`describeNumericRange`** — Clean, handles all four combinations (both/min-only/max-only/neither). Correctly handles `minimum: 0` since `0 !== undefined` is `true`.
- **`findOutOfRangeErrors`** — Properly guards with `typeof val !== "number"` to skip empty/string values. `NaN` values correctly produce no range error (since `NaN < x` and `NaN > x` are both `false`).
- **`step="any"`** for `type === "number"` — Correct fix. Without this, the browser's default `step=1` rejects valid decimals like `2.5`.
- **Test coverage** is solid: 4 tests for `describeNumericRange`, 5 for `findOutOfRangeErrors`, covering in-range, below-min, above-max, non-numeric types, and empty values.

---

### What to Do After Rebase

1. **Hook integration.** The `findOutOfRangeErrors` call needs to go into `usePromptTemplateTestPanelState.ts`'s `doRender` (line 78). The hook currently doesn't have access to `reconciledVariables` or `schemaForField`, so you'll need to either pass the reconciled schema info into the hook or move the range-field construction into the hook.

2. **`describeNumericRange` stays in the component** — this is purely presentational (helper text below the input), so it correctly remains in `PromptTemplateTestPanel.tsx`'s `renderField`.

3. **Watch for stale imports.** The old code imported `GroupsService` and `RenderPromptResponse` directly in the component — those were removed in the refactoring to the hook. Make sure they don't reappear after rebase.

4. **`HelperText`/`HelperTextItem` PatternFly imports** need to be added to the current import block structure on `main`.

---

### Summary

The approach and code quality are sound — the helpers are well-designed and well-tested. The PR just needs a rebase onto current `main` with the validation logic moved into the `usePromptTemplateTestPanelState` hook where `doRender` now lives.

PR #9525, opened 2026-08-12
fix(ui): preserve artifact version download content type

### LINKED ISSUE #9520: fix(ui): preserve stored content type when downloading artifact versions
## Description

**Registry Version**: snapshot (main @ d5c04da9ca08a43ecd3d42d2f65619a032324edd)  

The Registry UI download action on an artifact version infers the downloaded file MIME type and extension from `artifactType` instead of preserving the stored version content type.

For `PROMPT_TEMPLATE`, this is incorrect because the artifact type supports multiple content types: `application/json`, `application/x-yaml`, and `text/x-prompt-template`. A YAML prompt template is downloaded as `.json` with `application/json`.

Verified on latest upstream:
- Branch inspected: `main`
- Commit: `d5c04da9ca08a43ecd3d42d2f65619a032324edd`
- Still reproducible on upstream: Yes

Code references:
- `ui/ui-app/src/app/pages/version/VersionPage.tsx:207`
  `doDownloadVersion()` defaults to `ContentTypes.APPLICATION_JSON` and `json`.
- `ui/ui-app/src/services/useGroupsService.ts:221`
  `getArtifactVersionContent()` returns only decoded content, dropping response content type.
- `schema-util/util-provider/src/main/java/io/apicurio/registry/types/provider/StandardArtifactTypeProviderRegistry.java:276`
  `PROMPT_TEMPLATE` supports JSON, YAML, and `text/x-prompt-template`.
- `common/src/main/java/io/apicurio/registry/types/ContentTypes.java:10`
  `text/x-prompt-template` is a first-class content type.

Duplicate search performed:
- `download prompt template json yaml extension`
- `download content type extension UI`
- `VersionPage doDownloadVersion contentType`

I did not find an existing open/closed issue or PR covering this exact download content-type preservation problem.

### Environment

Local checkout of `Apicurio/apicurio-registry` at upstream `main`, commit `d5c04da9ca08a43ecd3d42d2f65619a032324edd`.

UI stack from `ui/ui-app/package.json`:
- React `19.2.8`
- Vite `7.3.6`
- PatternFly `6.6.1`
- TypeScript `5.9.3`

### Steps to Reproduce

1. Register a `PROMPT_TEMPLATE` artifact with YAML content and content type `application/x-yaml`.
2. Open that artifact version in the Registry UI.
3. Click the version download action.
4. Inspect the downloaded file name and MIME type.

Expected: The downloaded file should preserve the stored content type. A YAML prompt template should download as YAML, and `text/x-prompt-template` content should not be mislabeled as JSON.

Actual: The UI downloads the content as `.json` with `application/json`, because `VersionPage.tsx` guesses from artifact type and does not preserve the actual version content type.

### Logs

No server error is thrown. This is a UI download behavior issue.

Relevant code path:
- `VersionPage.doDownloadVersion()`
- `useGroupsService.getArtifactVersionContent()`
- `useDownloadService.downloadToFS()`

Suggested fix:
- Preserve the response `Content-Type` when fetching artifact version content, or add a service method that returns both content and content type.
- Centralize content-type-to-extension mapping in `ui/ui-app/src/utils/content.utils.ts`.
- Update `VersionPage.doDownloadVersion()` to use the stored content type rather than guessing from `artifactType`.
- Add regression coverage for YAML `PROMPT_TEMPLATE` downloads.

nyxsky404:
## Summary

- Preserve the artifact version response `Content-Type` when loading content for the Version page download action.
- Derive the downloaded file extension from the actual content type instead of guessing only from `artifactType`.
- Add regression coverage for YAML/prompt-template content types and the `latest` version content path.

## Root Cause

`VersionPage.doDownloadVersion()` inferred the download MIME type and extension from artifact metadata. That worked for single-format artifact types, but it breaks for `PROMPT_TEMPLATE`, which supports JSON, YAML, and `text/x-prompt-template` content. The existing `getArtifactVersionContent()` service method also returned only the decoded body, so the UI had no way to preserve the server response content type.

## Validation

- `npm test -- --run src/services/useGroupsService.test.ts src/utils/content.utils.test.ts`
- `npm run build` in `typescript-sdk`
- `npm run build` in `ui/ui-app`
- `git diff --check`

## Screenshots

Not included. This change affects downloaded file metadata/content type rather than visible UI layout.

Fixes #9520


ui/ui-app/src/services/useGroupsService.ts
```diff
@@ -231,6 +237,33 @@ const getArtifactVersionContent = async (config: ConfigService, auth: AuthServic
         });
 };
 
+const getResponseContentType = (headers: any): string => {
+    const header = typeof headers?.get === "function" ? headers.get("content-type") : headers?.["content-type"];
+    return typeof header === "string" ? header : ContentTypes.APPLICATION_JSON;
+};
+
+const getArtifactVersionContentWithType = async (config: ConfigService, auth: AuthService, groupId: string|null, artifactId: string, version: string): Promise<ArtifactVersionContent> => {
+    groupId = normalizeGroupId(groupId);
```
paoloantinori: This `normalizeGroupId` + `versionExpression` + `createEndpoint("/groups/:groupId/artifacts/:artifactId/versions/:version/content")` block is now duplicated across `getArtifactVersionContent`, this method, and `getArtifactVersionContentDereferenced`. Extract a shared `versionExpressionFor(version)` (and ideally a `buildVersionContentEndpoint(config, groupId, artifactId, version)`) so the `latest -> "branch=latest"` handling cannot drift between the three. Also prefer strict `===` here: `(version === "latest") ? "branch=latest" : version`.

paoloantinori (commented):
The fix is correct for its goal (preserving the server content type so prompt-template downloads pick up the right extension) and the new service method plus unit tests are solid. Two points should be addressed before merge: the new content path silently serves every artifact type on the version page, not only prompt templates, and the three sibling content methods now duplicate the same endpoint and version boilerplate.

- `ui/ui-app/src/services/useGroupsService.ts` (`getArtifactVersionContentWithType`, ~line 244) repeats the `normalizeGroupId` plus `versionExpression` plus `createEndpoint("/groups/:groupId/artifacts/:artifactId/versions/:version/content")` setup already present in `getArtifactVersionContent` and `getArtifactVersionContentDereferenced`. Extract a shared `versionExpressionFor(version)` and a `buildVersionContentEndpoint(...)` so the `latest` handling cannot drift between the three. Also use strict `===` on the `version == "latest"` check to match the rest of the module.
- `ui/ui-app/src/services/useGroupsService.ts` (`axios.get` block, ~line 250). `doDownloadVersion` is shared by every artifact type on the version page (the removed code branched on protobuf, wsdl, xsd, xml, graphql, thrift), so switching the version-page content load from the existing content method to a raw axios call changes the fetch path for all of those, not just prompt templates. The added tests only exercise yaml and prompt-template content. Add a test, or at least a note, confirming the new path returns byte-identical content for a non-prompt type (for example an XML or protobuf artifact), since the previous method may have transformed the body differently.
- `ui/ui-app/src/utils/content.utils.ts` (~line 216). Delegating `fileExtensionForDraft` to `fileExtensionForContentType` is a nice cleanup, but it silently extends the draft mapping: protobuf and thrift now resolve from content type where they previously fell through to `"txt"`. That changes draft filenames and is not covered by the new tests, so confirm it is intended.

ui/ui-app/src/services/useGroupsService.ts
```diff
@@ -231,6 +237,33 @@ const getArtifactVersionContent = async (config: ConfigService, auth: AuthServic
...
+    const versionExpression: string = (version == "latest") ? "branch=latest" : version;
+    const baseHref = config.artifactsUrl();
+    const endpoint = createEndpoint(baseHref, "/groups/:groupId/artifacts/:artifactId/versions/:version/content", {
+        groupId, artifactId, version: versionExpression
```
paoloantinori: Blast radius check: `doDownloadVersion` is the shared download action for every artifact type on the version page (the code this removes branched on protobuf, wsdl, xsd, xml, graphql, thrift), so this raw axios path now serves all of them, not just prompt templates. The new tests only cover yaml and prompt-template content. Can you add a test (or note) confirming this returns byte-identical content for a non-prompt type, e.g. an XML or protobuf artifact? The previous `getArtifactVersionContent` may have decoded or transformed the body differently than `responseType: "text"` + the no-op `transformResponse`, and a silent change in downloaded bytes for protobuf/wsdl is exactly the kind of regression that would not show up in the current tests.

ui/ui-app/src/utils/content.utils.ts
```diff
@@ -213,21 +213,35 @@ export function fileExtensionForDraft(draft: Draft, content: DraftContent): stri
         return "proto";
     }
 
-    if (content.contentType && content.contentType === ContentTypes.APPLICATION_JSON) {
+    return fileExtensionForContentType(content.contentType);
```
paoloantinori: Heads-up on a behavioral side effect: delegating `fileExtensionForDraft` to `fileExtensionForContentType` extends the draft mapping. Previously this function returned `"txt"` for `APPLICATION_PROTOBUF` and `APPLICATION_THRIFT` content types (it stopped at graphql/wsdl/xml/yaml/json); now it returns `"proto"` and `"thrift"`. That changes draft filenames, and `fileExtensionForDraft` is not covered by the new `fileExtensionForContentType` tests. If the extension is intended, add a draft-path assertion; if not, keep the old fall-through for the draft case.

ui/ui-app/src/services/useGroupsService.ts
nyxsky404: Thanks, addressed in `2652c52`. I extracted `versionExpressionFor(version)` using strict `===` and added `buildVersionContentEndpoint(...)` for the raw content endpoint construction. The content methods now share the same `latest -> branch=latest` handling so this logic should not drift between them.

ui/ui-app/src/services/useGroupsService.ts
nyxsky404: Thanks, added coverage for this in `2652c52`. There is now a non-prompt XML regression test for `getArtifactVersionContentWithType()` which asserts the returned content is byte-identical to the mocked response body and preserves `application/xml` as the response content type. The axios request is also asserted to use `responseType: "text"` with the identity `transformResponse`.

ui/ui-app/src/utils/content.utils.ts
nyxsky404: Good catch. I removed that unintended behavior change in `2652c52`. `fileExtensionForDraft()` now keeps the previous draft behavior: protobuf drafts still get `.proto` from the draft type, but protobuf/thrift content types alone do not silently change draft filenames. I added tests to lock that behavior.

nyxsky404:
@paoloantinori Thanks for the review. I’ve addressed the feedback in the latest push: extracted the shared content endpoint/version helpers, added non-prompt XML regression coverage, and removed the unintended draft filename behavior change. Focused tests and UI/SDK builds pass locally.

paoloantinori:
Thanks — re-read at 2652c52 and all three points landed cleanly:

- The `versionExpressionFor` / `buildVersionContentEndpoint` extraction removes the duplication across the three content methods, and the `latest` check is now strict `===`.
- The new XML test (`returns non-prompt content without transforming the response body`) confirms byte-identical pass-through with `responseType: "text"` + `transformResponse`, covering the non-prompt regression I was worried about.
- Splitting `fileExtensionForDraftContentType` out of `fileExtensionForContentType` keeps the draft path returning `.txt` for protobuf/thrift content types while still letting `.proto` flow from the draft artifact type — matches the previous behavior, and the two new draft tests lock it.

No further concerns from me.

PR #9469, opened 2026-08-11
docs: add OPENRPC artifact type and newer OpenAPI/AsyncAPI versions

eameh:

PR #9528, opened 2026-08-12
fix(storage): prevent infinite recursion on circular references

### LINKED ISSUE #9527: Circular artifact references cause infinite recursion in generic reference resolution
## Description

**Registry Version**: 3.3.2-SNAPSHOT / latest snapshot from `main`  
**Persistence type**: SQL  
**Severity**: Potential denial of service

Apicurio Registry can enter infinite recursion when artifacts contain circular references. The generic resolver checks whether a reference is already present in the resolved-reference map, but adds the reference only after recursively processing its children.

For a cycle such as:

```text
A -> B -> A
```

the reference is not present in the map when recursion returns to it. This causes unbounded recursion and eventually a `StackOverflowError`.

### Environment

- Registry built from the latest `main` snapshot
- JDK 21
- REST API v3
- Registry running with `LOG_LEVEL=DEBUG`
- Reproduced with SQL persistence; the generic resolver is shared by multiple storage and REST paths

### Steps to Reproduce

1. Start Apicurio Registry with SQL storage and DEBUG logging enabled.
2. Create artifact `B` with version `1` and no references.
3. Create artifact `A` with version `1`, referencing `B` version `1` using a reference name such as `B.avsc`.
4. Create version `2` of artifact `B`, referencing `A` version `1` using a reference name such as `A.avsc`.
5. Request dereferenced content for `A` version `1`:

```text
GET /apis/registry/v3/groups/default/artifacts/A/versions/1?references=DEREFERENCE
```

Expected:

- Circular references are detected and traversal stops safely.
- The request returns a controlled response.
- The server remains available for subsequent requests.

Actual:

- Reference resolution recurses indefinitely.
- The request fails with `StackOverflowError`.
- Depending on exception mapping, the client may receive HTTP 500 or a connection reset.
- Repeated requests could exhaust request threads and affect server availability.

### Logs

Representative failure:

```text
java.lang.StackOverflowError
    at io.apicurio.registry.storage.impl.sql.RegistryContentUtils
        .recursivelyResolveReferencesGenericInternal(...)
```

The generic resolver currently catches `Exception`, but `StackOverflowError` extends `Error` and is not caught.

Ananya44444:
Fixes #9527 

## What

Prevent infinite recursion when resolving circular artifact references.

## Why

`RegistryContentUtils.recursivelyResolveReferencesGenericInternal` previously checked the resolved-reference map before recursion but only inserted entries after recursion completed. Circular references such as `A -> B -> A` therefore bypassed the guard and could cause a `StackOverflowError`.

## Changes

- Track visited reference keys and mark them before loading or recursing.
- Stop generic reference traversal after the existing 100-level safety bound used by the SQL resolver.
- Add focused unit tests covering circular references and excessive nesting.

## Verification

- `RegistryContentUtilsTest`: 4 tests passed.
- App-module Checkstyle: 0 violations.
- `git diff --check`: passed.

The commit includes a DCO sign-off.


paoloantinori (commented):
The recursion guard itself is correct and the tests are specific (exact load counts, exact map sizes), but the fix is incomplete because the structurally identical dereference path still carries the same cycle bug, and the new depth cap duplicates an existing configurable property rather than reusing it.

- Incomplete fix: the private `resolveReferencesWithContext` method in `RegistryContentUtils.java` (unchanged here) has the same `containsKey` -> recurse -> `put` order that caused the original `StackOverflowError`. It is reachable from production read/dereference endpoints: `GroupsResourceImpl.java:157` and `:681`, `IdsResourceImpl.java:99`, and `AbstractResourceImpl.java:64`. A circular reference dereferenced through those endpoints still overflows. Either apply the same `visited` guard there, or open a follow-up issue so it is tracked rather than silently left behind.
- Config divergence: the new `static final int MAX_REFERENCE_DEPTH = 100` duplicates the existing `@ConfigProperty(name = "apicurio.storage.references.max-depth", defaultValue = "100")` in `AbstractSqlRegistryStorage.java:120`, which feeds the depth guard in `SqlContentRepository`. Both bound the same conceptual limit but can now drift silently if an operator tunes the property. Inject the configured value, or add a comment documenting why this path is intentionally independent.
- Correctness of the guard: the shared `visited` set with mark-before-recurse breaks `A -> B -> A` cleanly, and DAG nodes are not dropped because `put` runs after recursion while re-entries are skipped via `visited.add` returning false.

app/src/main/java/io/apicurio/registry/storage/impl/sql/RegistryContentUtils.java
```diff
@@ -35,6 +37,8 @@ public class RegistryContentUtils {
 
     static final String NULL_GROUP_ID = "__$GROUPID$__";
 
+    static final int MAX_REFERENCE_DEPTH = 100;
```
paoloantinori: This hardcodes 100 while `AbstractSqlRegistryStorage` already exposes `apicurio.storage.references.max-depth` (default 100) as a `@ConfigProperty` that feeds `SqlContentRepository`'s depth guard. Two independent bounds on the same limit will drift the moment an operator tunes one. Prefer injecting the configured value, or leave a comment explaining why this code path is intentionally held to a separate constant.

app/src/main/java/io/apicurio/registry/storage/impl/sql/RegistryContentUtils.java
```diff
@@ -75,30 +80,42 @@ private static <KEY, NODE extends HasReferences, VALUE> void recursivelyResolveR
...
+        if (references == null || references.isEmpty()) {
+            return;
+        }
+
+        if (currentDepth > MAX_REFERENCE_DEPTH) {
+            log.warn("Maximum generic reference resolution depth ({}) exceeded at depth {}. "
+                    + "Stopping resolution.", MAX_REFERENCE_DEPTH, currentDepth);
+            return;
+        }
+
```
paoloantinori, line 94: The `visited` guard here is correct, but the sibling `resolveReferencesWithContext` method below still uses the old `containsKey` -> recurse -> `put` order and is reachable from the v2/v3 dereference endpoints (`GroupsResourceImpl`, `IdsResourceImpl`, `AbstractResourceImpl`). A circular reference on that read path still overflows the stack. Apply the same guard to that method, or open a follow-up issue so the remaining instance is tracked.

Ananya44444:
@paoloantinori Thanks, both points are valid. I’ll update the PR to protect the context aware resolver as well, including a regression test for circular dereference, and make both resolver paths use the existing apicurio.storage.references.max-depth configuration. I’ll push the updates and rerun the focused tests and Checkstyle.

paoloantinori:
Sounds like a solid plan. Two things I'll specifically want to see once you push:

- The context-aware resolver (`ContextAwareArtifactReferenceResolver` dereference path) gets the same `visited`-set + depth guard, not just `RegistryContentUtils`. The structurally identical path was the gap I flagged.
- A regression test that actually constructs a circular dereference (A dereferences B, B dereferences back to A) and asserts it terminates without throwing, separate from the nesting-depth test.

Reusing `apicurio.storage.references.max-depth` for both paths is the right call — keeps the bound configurable and avoids two magic numbers. Ping me once it's pushed and I'll re-read.

Ananya44444:
@paoloantinori I’ve pushed the requested updates. The private resolveReferencesWithContext path used by the v2/v3 dereference endpoints now has the same mark-before-recursion visited guard and configurable depth limit as the generic resolver. Both paths reuse apicurio.storage.references.max-depth.
I also added a separate context aware circular dereference regression test that constructs A -> B -> A, asserts the operation does not throw, and verifies that each artifact is loaded only once. The nesting depth test remains separate.
The focused reactor test run passed all 6 tests. Could you please re read the PR when convenient?

paoloantinori:
Re-read after your push, and the context-aware path now matches the generic resolver. The mark-before-recursion guard (`visited.add(newRefName)` before the recursive call), the `currentDepth + 1` accounting, and the shared `apicurio.storage.references.max-depth` bound all landed cleanly on both paths. The new circular dereference test (A -> B -> A, asserts no throw and exactly two loads) pins the regression down the same way the generic one does.

One small thing worth confirming: in `recursivelyResolveReferencesWithContext` the `visited` set is keyed on `newRefName` (the rewritten coordinate string) while the generic resolver keys on the raw extracted key. That's the right call here because the same underlying artifact can be reached through distinct rewrite coordinates, so dedup on `newRefName` is the correct granularity for this path. I just want to flag it so it's an intentional choice rather than an oversight, since the two paths now diverge on the visited-key type.

Nothing else from my side. The fix is complete and the test coverage is specific.

Ananya44444:
Thanks for confirming. Yes, keying the context-aware visited set on newRefName was intentional. That path resolves and rewrites references using full artifact-version coordinates, so the rewritten coordinate string is the appropriate deduplication granularity, while the generic resolver continues to use its extracted key. Thanks for the review!

EricWittmann (commented):
Thanks for this fix — the cycle detection approach using `visited.add(key)` as a guard is correct, and I appreciate that you addressed both recursive methods (generic and context-aware). The tests are well-written with specific assertions on load counts and map sizes.

Two changes requested before we can merge:

### 1. Avoid static `ConfigProvider.getConfig()` lookup

The `getMaxReferenceDepth()` method in `RegistryContentUtils` reaches into MicroProfile Config from a static context with a try-catch fallback for non-CDI environments. This is an anti-pattern in Quarkus — it creates a hidden coupling and makes the utility harder to test.

The callers of the public methods (`recursivelyResolveReferencesGeneric` and `recursivelyResolveReferencesWithContext`) are CDI beans that already have access to `ReferenceResolutionConfigProperties`. Instead, add `int maxReferenceDepth` as a parameter to both public methods and let the callers pass it in. This removes the need for the static config lookup entirely and keeps `RegistryContentUtils` a pure utility class.

### 2. Off-by-one in depth guard

The condition `currentDepth > maxReferenceDepth` allows depth values 0 through 100 when `maxReferenceDepth=100`, which is actually 101 levels of recursion. The config description says "maximum recursion depth" of 100, but the code permits 101 levels. Either change the guard to `currentDepth >= maxReferenceDepth` so the behavior matches the documented limit, or update the description to clarify the semantics.

Ananya44444:
@EricWittmann Thanks for the review. Before I push the revision, I want to confirm the intended scope.
I removed the static ConfigProvider lookup, added maxReferenceDepth to both recursive entry points, and changed both depth guards from > to >=.
While propagating the configured value, I found that recursivelyResolveReferencesGeneric is also called through recursivelyResolveReferences, recursivelyResolveReferenceContentIds, canonicalizeContent, and canonicalContentHash. Those paths are used not only by CDI resources, but also by manually constructed importers, the polling validator, and the reflectively instantiated database upgrader.
Therefore, ensuring that every resolution path uses the configured apicurio.storage.references.max-depth value currently requires argument forwarding across 18 production files plus 2 test files. Most changes are one line parameter propagation.
Is this full propagation what you intended or would you prefer a narrower approach?

EricWittmann:
Go ahead and let's see what it looks like!  Changes to multiple files is fine if warranted, especially if the change is just a single line.  But maybe we'll be able to identify an improvement once we see the diff.  :+1:

Ananya44444:
@EricWittmann I’ve pushed the requested changes and rebased the branch onto the latest main.
The static ConfigProvider lookup has been removed, the configured maxReferenceDepth is now passed explicitly through all affected resolution paths, and the depth guards now use >= to correct the off-by-one behavior.
One CI note: Build Java Application (no tests) is failing in avro-maven-plugin:1.12.2 while processing existing schemas under app/src/test/resources/io/apicurio/registry/serde. The same failure occurred in the merged dependency update PR #9469  , which is now part of the base branch. Verification Gate is failing as a consequence of that build failure.
After bypassing only that unrelated Avro generation phase, all production and test Java sources compiled successfully, and the 17 focused tests passed with zero failures.
The updated diff is ready for review. I’m also happy to help investigate the inherited Avro failure separately after this PR is resolved, if that would be useful.

Ananya44444:
/unstale
@EricWittmann @paoloantinori I’ve pushed the requested changes and updated the affected resolution paths. The focused tests pass, while the remaining CI failure appears unrelated and originates from the existing Avro plugin issue.
Could you both please take another look at the updated diff when convenient?

PR #9529, opened 2026-08-12
fix(a2a): return bad request for structure search without index

AchalNarwade:
## Description

Fixes an issue where A2A advanced agent search requests using structure-based filters, such as `skills`, returned HTTP 500 when the search index was not enabled.

For example:

POST /.well-known/agents/search

{
  "filters": {
    "skills": ["schema-validation"]
  }
}

Previously, the SQL search repository did not handle the `structure` filter type and fell through to the default case, resulting in:

HTTP 500
RegistryStorageException: Filter type not supported: structure

This change treats structure search consistently with content search when the search index is not enabled and returns HTTP 400 with a clear message.

## Changes

- Added a structure-search unsupported message to `SqlSearchRepository`.
- Added `structure` handling to artifact searches.
- Added `structure` handling to version searches.
- Added a regression test for A2A advanced search with a skills filter.

## Testing

- `.\mvnw.cmd test -pl app -Dtest=WellKnownResourceTest`
- Tests run: 34
- Failures: 0
- Errors: 0
- Skipped: 0

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
@@ -38,6 +38,10 @@ public class SqlSearchRepository {
             "Content search requires the search index, which is not enabled. "
             + "Enable the search index to use content search.";
 
+    private static final String STRUCTURE_SEARCH_UNSUPPORTED_MESSAGE =
+            "Structure search requires the search index, which is not enabled. "
+                    + "Enable the search index to use structure search.";
+
     private final Logger log;
 
     private final SqlStatements sqlStatements;
@@ -168,6 +172,8 @@ public class SqlSearchRepository {
                         break;
                     case content:
                         throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                    case structure:
+                        throw new ContentSearchNotSupportedException(STRUCTURE_SEARCH_UNSUPPORTED_MESSAGE);
                     default:
                         throw new RegistryStorageException("Filter type not supported: " + filter.getType());
                 }
@@ -347,6 +353,8 @@ public class SqlSearchRepository {
                         break;
                     case content:
                         throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                    case structure:
+                        throw new ContentSearchNotSupportedException(STRUCTURE_SEARCH_UNSUPPORTED_MESSAGE);
                     default:
                         throw new RegistryStorageException("Filter type not supported: " + filter.getType());
                 }
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
@@ -5,6 +5,7 @@ import io.apicurio.registry.storage.dto.ArtifactSearchResultsDto;
 import io.apicurio.registry.storage.dto.OrderBy;
 import io.apicurio.registry.storage.dto.OrderDirection;
 import io.apicurio.registry.storage.dto.SearchFilter;
+import io.apicurio.registry.storage.dto.SearchFilterType;
 import io.apicurio.registry.storage.dto.SearchedArtifactDto;
 import io.apicurio.registry.storage.dto.SearchedVersionDto;
 import io.apicurio.registry.storage.dto.VersionSearchResultsDto;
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -166,8 +170,8 @@ public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, Order
                         });
                         where.append(")");
                         break;
-                    case content:
-                        throw new ContentSearchNotSupportedException(CONTENT_SEARCH_UNSUPPORTED_MESSAGE);
+                    case content, structure:
```
paoloantinori, line 173: Confirm this is the complete set. The PR adds `structure` to two switches (here in searchArtifacts and in searchVersions), but any other `switch` over `SearchFilterType` in this class or sibling search repositories that lacks the `structure` case still falls through to the 500 `RegistryStorageException`. Worth a grep for every `SearchFilterType` switch and every `case content` to make sure no path is missed.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
```diff
@@ -345,8 +349,8 @@ public VersionSearchResultsDto searchVersions(Set<SearchFilter> filters, OrderBy
...
```
paoloantinori, line 352: This branch change has no test. The new regression test covers only the artifact/A2A path (WellKnownResourceTest). If a version search can carry a structure filter, add the analogous 400 case here; otherwise a one-line comment noting the version branch is defensive (no current caller sends structure) would document the assumption.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
```diff
@@ -778,4 +781,28 @@ private void createMcpTool(String groupId, String artifactId, String content) th
...
+        givenAtRoot()
+                .when()
+                .contentType(ContentType.JSON)
+                .body(requestBody)
+                .post("/.well-known/agents/search")
+                .then()
+                .statusCode(400)
+                .body("detail", allOf(
+                        containsString("search index"),
+                        not(containsStringIgnoringCase("select"))));
```
paoloantinori, line 807: The `not(containsStringIgnoringCase("select"))` assertion is a nice guard against the error body leaking SQL or a raw stacktrace, but it is non-obvious on first read. A short comment (e.g., `// guard: error body must not leak SQL/stacktrace`) would save the next reader from puzzling over why "select" is the chosen sentinel.

paoloantinori (commented):
Small, well-scoped fix that correctly turns a 500 into a 400 for `structure` filters when the search index is off, reusing the existing `ContentSearchNotSupportedException` to 400 mapping and adding a regression test. Approachable as-is; a couple of completeness and coverage checks are worth doing before merge.

- Good: routing both `content` and `structure` through `indexOnlyFilterException` keeps the two switch sites symmetric, and `Locale.ROOT` on `toLowerCase` avoids a class of locale bugs (SqlSearchRepository.java:42).
- Good: the test's `not(containsStringIgnoringCase("select"))` guard (WellKnownResourceTest.java:806) catches accidental SQL or stack-trace leakage in the error body, which is more than a plain status-code check.
- Verify completeness: the PR updates exactly two switches (searchArtifacts at SqlSearchRepository.java:173, searchVersions at :352). Confirm there are no other `switch` sites over `SearchFilterType` in this class or sibling search repositories that still fall through to the 500 `RegistryStorageException` for `structure`, or those paths stay broken.
- Coverage gap: the `searchVersions` change (:352) has no test. The new test only exercises the artifact/A2A path. If a version search can carry a structure filter, add the analogous 400 case; if not, a one-line note on why the branch is defensive would document the assumption.
- Naming: `indexOnlyFilterException` returns a `ContentSearchNotSupportedException` even for structure (SqlSearchRepository.java:38). The reuse inherits the correct 400 mapping, but the exception name no longer matches the structure case. A neutral type (a search-index-required exception) would read better long-term; fine to defer.
- Minor: the `type == structure ? "Structure" : "Content"` ternary (:39) silently defaults any future index-only filter to "Content". Adding another index-only type means updating both the case label and this ternary; an enum property or switch would be less error-prone.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
AchalNarwade: I checked the SearchFilterType switch sites in SqlSearchRepository and confirmed that the relevant searchArtifacts() and searchVersions() switches are the two paths handling these filters. Both now handle content and structure together.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlSearchRepository.java
AchalNarwade: Added a regression test in VersionSearchTest covering the structure filter path in searchVersions(). The test verifies that ContentSearchNotSupportedException is returned when the search index is unavailable.

app/src/test/java/io/apicurio/registry/noprofile/rest/a2a/WellKnownResourceTest.java
AchalNarwade: Added a short comment explaining that the assertion guards against SQL or stacktrace leakage in the error response.

PR #9531, opened 2026-08-13
fix(http): send configured webhook headers on outbound provider requests

### LINKED ISSUE #9530: Webhook provider headers are parsed from config but never sent on outbound requests
**Persistence type**: not storage-specific. Affects the custom artifact type (webhook provider) egress path.

`WebhookProvider` accepts a `headers` map from the artifact types config file, but nothing ever applies it to the outbound request:

- the field and its getter: https://github.com/Apicurio/apicurio-registry/blob/89f40491cdcd17139abd34ac9dd5366e24a15bb5/common/src/main/java/io/apicurio/registry/config/artifactTypes/WebhookProvider.java#L28-L30
- `AbstractWebhookDelegate.invokeHook` has the provider in scope but passes only the URL: https://github.com/Apicurio/apicurio-registry/blob/89f40491cdcd17139abd34ac9dd5366e24a15bb5/app/src/main/java/io/apicurio/registry/types/provider/configured/AbstractConfiguredArtifactTypeUtil.java#L29-L31
- `HttpClientService.post` sets `Content-Type` and nothing else: https://github.com/Apicurio/apicurio-registry/blob/89f40491cdcd17139abd34ac9dd5366e24a15bb5/app/src/main/java/io/apicurio/registry/http/HttpClientService.java#L48-L53

`WebhookProvider.getHeaders()` has exactly one caller in the repo, a parse-only assertion in `ArtifactTypesConfigurationTest`. There are none under `src/main`.

That test's own fixture is `"Authorization": "Bearer YOUR_SECRET_TOKEN"`, so the field looks intended for authentication. All six configured providers call through `invokeHook`, so every outbound provider request goes out without it.

### Steps to reproduce

1. Configure a custom artifact type whose `contentAccepter` is a webhook, with `"headers": { "X-Api-Key": "test-secret" }`.
2. Point `apicurio.artifact-types.config-file` at that file and start the Registry.
3. Create an artifact of that type, and log the headers the endpoint receives.

**Expected:** the request carries `X-Api-Key: test-secret`.
**Actual:** the only header set is `Content-Type: application/json`.

harsh04043:
Fixes #9530

`invokeHook` has the provider in scope but passes only the URL, so the `headers` map on `WebhookProvider` is parsed and then dropped. All six configured providers call out unauthenticated. This has never worked, since custom artifact types landed in #6086.

This adds a `headers` parameter to `HttpClientService.post` and passes `provider.getHeaders()` from `invokeHook`.

Two decisions worth flagging:

**`Content-Type` wins over configuration.** The example config in `ArtifactTypesConfigurationTest` puts `Content-Type` in the headers map, and `sendJson` only applies `application/json` when the header isn't already set. Honouring a configured value would let a JSON body go out mislabelled, so I apply the configured headers first and set `Content-Type` last. Happy to flip it if you'd rather config win.

**`post` changed signature instead of gaining an overload.** I tried the overload first, with the 3-arg form delegating to it. That nests the `@Retry` budgets, because ArC intercepts self-invocation: I measured 63 requests against a permanently failing endpoint where there should be 9. A single entry point avoids it, and every caller was in `app`.

Invalid header values are skipped rather than thrown, since the exception message quotes the offending value and here that value is a credential. Only the header name is logged.

Not addressed here: once these headers are on the request, a custom credential such as X-Api-Key can be forwarded if an endpoint answers with a 303 pointing at another origin. Vert.x strips only its own default set there, which covers Authorization but not custom headers. The redirect behaviour predates this change, the exposure does not, since the headers never reached the wire before. Tracked in #9655.
 
Tests cover the headers reaching the wire through the full provider wiring (that one fails if `getHeaders()` is dropped again), `Content-Type` precedence across header-name casing, and the null, empty and malformed cases.


app/src/main/java/io/apicurio/registry/http/HttpClientService.java
```diff
@@ -45,11 +68,13 @@ public class HttpClientService {
...
+    public <I, O> O post(String url, Map<String, String> headers, I body, Class<O> outputClass) throws HttpClientException {
         try {
-            // POST the request to the webhook endpoint
-            HttpRequest<Buffer> request = webClient.postAbs(url).putHeader("Content-Type", ContentTypes.APPLICATION_JSON)
-                    .followRedirects(true);
+            // Configured headers go on first so the Content-Type below wins. Vert.x only defaults
+            // Content-Type when absent, so one from configuration would mislabel the JSON body.
+            HttpRequest<Buffer> request = webClient.postAbs(url);
+            applyHeaders(request, headers);
+            request.putHeader("Content-Type", ContentTypes.APPLICATION_JSON).followRedirects(true);
```
paoloantinori, line 77: Configured auth headers (X-Api-Key, Authorization) are now on the request for the first time, and `followRedirects(true)` with `WebClientProducer` setting zero redirect options means whether they survive a cross-origin redirect depends on this Vert.x version's default strip-set. Before this change the headers never reached the wire, so this is a newly-introduced exposure, not pre-existing. Worth verifying the default or mitigating in `WebClientProducer`.

paoloantinori (commented):
Re-read this against the source tree at the PR head. The root-cause fix is correct and complete: `invokeHook` was parsing `provider.getHeaders()` and dropping it, and this PR wires it through. I confirmed exactly one production caller of `HttpClientService.post`, so the signature change is safe.

One thing worth resolving before merge.

**Configured credentials can leak on cross-origin redirect.** The PR description calls this "pre-existing," but the mechanism and the exposure are different. Before this change the configured headers never reached the wire at all (they were dropped in `invokeHook`), so they could never be forwarded to a redirect target. This PR is what puts them on the request, and the whole point of the feature is to send auth secrets (`X-Api-Key`, `Authorization`). With `followRedirects(true)` and `WebClientProducer` setting zero redirect options (just `WebClient.create(vertx)`), whether a custom auth header survives a cross-origin redirect depends entirely on this Vert.x version's default strip-set. That is a real CVE class across HTTP clients.

Options, in order of preference: (1) verify the Vert.x version's default strip-set and confirm here if it already drops hop-by-hop and auth headers on cross-origin redirect; (2) mitigate in `WebClientProducer` by configuring redirect handling to strip configured headers before the redirect; (3) at minimum file a follow-up explicitly framed as a newly-introduced exposure so it isn't lost behind "pre-existing."

Done well:
- `Content-Type` precedence is right: forcing `application/json` last is correct, and it holds across casing since Vert.x `MultiMap` is case-insensitive.
- Keeping a single public entry point for the `@Retry` / `@Timeout` annotations is the correct call. The ArC self-invocation problem is real, and the class-level Javadoc warning future maintainers is the right safeguard.
- Defensive handling is good throughout: null map, empty map, and null value are no-ops, invalid CRLF values are caught and skipped, and only the header name is logged (the rejected value is a credential and Vert.x quotes it in the exception message).
- Test coverage is layered: `ConfiguredWebhookHeadersTest` exercises the full provider-to-HTTP wiring and fails if `getHeaders()` is dropped again, and `HttpClientServiceTest` covers the unit edges including the case-insensitive `Content-Type` variants and the CRLF-rejection path.

Resolve the redirect-header-leak item and this is good to merge.

harsh04043:
Agreed on the framing, and I'll update the description to match. The redirect behaviour predates this PR, the exposure doesn't, since the headers never reached the wire before.

On (1), I checked 4.5.31, which is what we resolve transitively. The defaults are:

- cross-origin: `authorization`, `cookie`, `proxy-authorization`, `content-length`
- same-origin: `cookie`, `content-length`

So `Authorization` is already stripped, but custom headers aren't. I ran two servers on different ports with the first redirecting to the second: `Authorization` and `Cookie` were dropped, `X-Api-Key` arrived at the second origin intact. Worth noting the window is narrower than it looks, since `DefaultRedirectHandler` only follows a 303 for a POST, never a 301/302/307/308.

So (1) doesn't close it on its own.

(2) doesn't fit either, as far as I can tell. `setCrossOriginRedirectBlockedHeaders` takes a fixed set at client construction, but configured header names are per-provider, so it would need the union of every provider's names at startup. `WebClientProducer` also produces the clients used by the auth paths, so it isn't really ours to reconfigure.

What I have working locally is per-request instead: don't follow redirects when a configured header was attached. That closes it for every status code, but it's broader than the vulnerability, because `HttpRequest` only exposes `followRedirects(boolean)` with no origin-aware variant, so same-origin redirects get refused too. An unfollowed redirect also has to abort rather than retry, otherwise it spends the full backoff chain replaying a deterministic response.

Happy to push that here, or leave this PR as-is and open the follow-up from (3). Which would you prefer?

cc: @paoloantinori

paoloantinori:
Thanks for the empirical check on the strip-set. Two servers on different ports is exactly the right way to verify this, and the `DefaultRedirectHandler` 303-only detail narrows the window considerably. Your reasoning on (1) and (2) holds up: the default strip-set covers `Authorization` but not custom auth headers, and the client-level configuration doesn't fit a per-provider header set on a shared `WebClient`.

Open the follow-up. The root-cause fix here is complete for what the PR claims to do, and the leak is newly introduced but narrow (303-redirect-to-different-origin from an operator-configured endpoint). The per-request `followRedirects(false)` approach has a real tradeoff (same-origin redirects also refused, since `HttpRequest` has no origin-aware variant) that deserves its own focused discussion rather than being folded into a wiring fix.

One thing to nail down when you push that follow-up: the abort-not-retry point is the crux. With `followRedirects(false)`, a 3xx response falls outside the 2xx success branch and outside the 4xx client-error abort branch in `processResponse`, so make sure the status-code routing explicitly treats 3xx as a non-retryable abort. Otherwise a webhook endpoint that returns 303 burns the full `@Retry` backoff chain against a deterministic response, which is what you flagged. Worth a test that pins that behaviour.

This one is good to merge as-is.

harsh04043:
Follow-up filed as #9655, covering the strip-set findings and the 3xx abort point you flagged. I'll open the PR for it once this one merges.

paoloantinori (commented):
This is in good shape. The header handling is right and the 146 new assertions are specific rather than existence checks, which is what I look for.

The one thing holding me back from approving is that none of those assertions has ever executed. The PR has been idle for 35 days and the Unit Tests job has not run against this code. Please rebase onto current main, and once the suite comes back green I will approve.

Apologies for the wait on this one. The delay is on the maintainer side, not yours.

Issue #7337, opened 2026-02-10
Remove deprecated KafkaSQL message classes (v8, v9, v10)


carlesarnal:
## Summary

The KafkaSQL storage implementation contains deprecated message format classes that are maintained for backwards compatibility with older event journal entries. These should be scheduled for removal after an appropriate migration window.

## Deprecated Classes

Located in `app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/messages/`:

| Class | Status | Replacement |
|-------|--------|-------------|
| `CreateArtifact9Message.java` | @Deprecated | Current message format |
| `CreateArtifact10Message.java` | @Deprecated | Current message format |
| `CreateArtifactVersion8Message.java` | @Deprecated | Current message format |
| `CreateArtifactVersion9Message.java` | @Deprecated | Current message format |

## Rationale

1. These classes exist solely for deserializing old journal entries
2. After sufficient time, all journals should have been compacted/migrated
3. Maintaining multiple message versions adds complexity to the codebase

## Proposed Action

1. Document the migration window in release notes (e.g., "deprecated message formats will be removed in v4.0")
2. Add logging when deprecated message formats are encountered (to track usage)
3. Remove the deprecated classes after the migration window closes

## Considerations

- Users running KafkaSQL storage may have old journal entries
- Need to verify journal compaction policies before removal
- Consider providing a migration utility or documentation

## Acceptance Criteria

- [ ] Deprecation timeline documented in release notes
- [ ] Logging added for deprecated message format usage
- [ ] Classes removed after migration window (target: v4.0)

carlesarnal:
Yes, deprecation is set for the next release, removal for 4.0.

Myself-Praveen:
Hi @carlesarnal — I'd like to take this on.

My understanding of the problem:

The KafkaSQL storage layer has 4 deprecated message classes that exist solely for backwards-compatible deserialization of old journal entries:

CreateArtifact9Message (v9 format)
CreateArtifact10Message (v10 format)
CreateArtifactVersion8Message (v8 format)
CreateArtifactVersion9Message (v9 format)
These are registered in KafkaSqlMessageIndex and must stay registered so existing journals can still replay. But there's currently no visibility into whether any deployments are still encountering these legacy formats — which is a prerequisite for knowing when it's safe to remove them in v4.0.

Proposed approach (deprecation logging phase):

[...]

PR #9533, opened 2026-08-13
fix(kafkasql): register usage event messages so telemetry is not discarded

### LINKED ISSUE #9532: KafkaSQL: schema usage telemetry is silently discarded because two journal message types are missing from KafkaSqlMessageIndex
## Description

**Registry Version**: snapshot (`main`, 3.3.2-SNAPSHOT)

On the KafkaSQL variant, schema usage telemetry never reaches storage. The events are produced to the journal topic correctly, but the consumer cannot deserialize them, so every one is discarded and the usage tables stay empty.

The cause is that `KafkaSqlMessageIndex` is missing two entries. There are 77 message classes in `storage/impl/kafkasql/messages/`, and 75 of them are registered in the static block. The two that are not:

- `RecordUsageEvent1Message`
- `DeleteOldUsageEvents1Message`

Both are actively submitted by `KafkaSqlRegistryStorage` (`recordUsageEvent()` at line 1316, `deleteOldUsageEvents()` at line 1322).

`KafkaSqlValueDeserializer` resolves the message type from the `mt` Kafka header through `KafkaSqlMessageIndex.lookup(...)`. For these two types the lookup returns `null`, the deserializer throws, catches its own exception, logs it and returns `null`. Because the *key* deserializes fine (the key deserializer never consults the index), the record survives the null-key guard in `processRecord` and lands in the branch that treats a null value as a tombstone. So a lost mutation gets reported at INFO as a presumed tombstone.

Neither call site waits on the coordinator — both are fire-and-forget — so nothing surfaces to the caller either.

This does not self-heal. The journal replays from the beginning (or from the last snapshot) on every startup, so the same messages fail deserialization again on every restart, forever.

### Scope

Worth being precise about who this affects, because it is narrower than it first looks:

- `apicurio.usage.telemetry.enabled` defaults to `false`, so only deployments that explicitly turned usage telemetry on are affected.
- KafkaSQL only. The SQL variant calls these storage methods directly and is fine. GitOps and KubernetesOps are read-only.
- No data corruption. Usage events are lost, nothing else is touched.

What is lost is the whole usage feature on KafkaSQL: `/admin/usage/summary`, `/admin/usage/artifacts/{groupId}/{artifactId}`, the heatmap and deprecation-readiness endpoints, and the `UsageAggregationJob`, all return empty results.

`deleteOldUsageEvents` is also never applied. Harmless today since there is nothing to prune, but it becomes real the moment the first half works.

### This is the second time

`DeleteAllOrphanedContent0Message` was found missing from the same index and fixed in #7810 ("Fix missing message index entry ... was previously missing"). Same defect, same file, different class.

Nothing links "class extends `AbstractMessage`" to "class is in the index". It is a hand-maintained list, the compiler cannot check it, and the failure is swallowed on the consume side. So it is going to keep happening. I would like to add a test that scans the messages package and fails the build when any implementation is unregistered, alongside the two-line fix.

I am happy to open a PR for both if a maintainer agrees with the approach.

### Environment

Found by reading the code, not by hitting it in a running cluster, so I do not have a production environment to report. Reproduced against `main` at commit `89f40491c` on macOS with JDK 21.

I was not able to run a live KafkaSQL instance locally to capture a transcript. Starting `KafkaTestContainerManager` times out on my machine, which is an environment problem on my end and unrelated to this bug.

### Steps to Reproduce

The fastest check needs no running instance. From a clean checkout of `main`:

```bash
# every message class on disk
ls app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/messages/ \
  | sed 's/\.java//' | sort > /tmp/all.txt

# every class registered in the index
grep -oE "[A-Za-z]+[0-9]+Message" \
  app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/serde/KafkaSqlMessageIndex.java \
  | sort -u > /tmp/indexed.txt

comm -23 /tmp/all.txt /tmp/indexed.txt
```

Output:

```
DeleteOldUsageEvents1Message
RecordUsageEvent1Message
```

To see the runtime effect:

1. Start Registry with `APICURIO_STORAGE_KIND=kafkasql` and `APICURIO_USAGE_TELEMETRY_ENABLED=true`, with `LOG_LEVEL=DEBUG`.
2. Create an artifact and note its `globalId`.
3. Fetch it a few times through `GET /apis/registry/v3/ids/globalIds/{globalId}`, which is what `UsageTelemetryFilter` hooks. Optionally set `X-Registry-Client-Id` and `X-Registry-Operation` headers.
4. Wait past the buffer flush (`apicurio.usage.flush.every`, default `10s`).
5. Inspect the journal topic. The `RecordUsageEvent1Message` records are there, so the produce side works.
6. Call `GET /apis/registry/v3/admin/usage/summary`. It comes back empty.
7. Restart the node. The same messages fail again during replay.

Running the same steps with `APICURIO_STORAGE_KIND=sql` gives populated usage data, which isolates it to the KafkaSQL consume path.

**Expected**: a usage event submitted to the journal is deserialized by the consumer, applied to the node's SQL projection, and shows up in the usage endpoints. Every node converges on the same usage data by replaying the journal.

**Actual**: the message is produced successfully but cannot be deserialized, so it is dropped. Usage endpoints return empty on every node, and replay drops the same records again on every restart.

### Logs

I could not capture a live transcript for the reason given under Environment, so rather than paste something I did not actually observe, here are the two lines the code emits and where they come from.

`KafkaSqlValueDeserializer` (line 66) logs the real cause at ERROR. `ex.getMessage()` here is `Unknown KafkaSql message class for 'RecordUsageEvent1Message'`:

```
ERROR ... Error deserializing a KafkaSQL message value: Unknown KafkaSql message class
for 'RecordUsageEvent1Message'. First 32 bytes of the message value are: ...
```

`KafkaSqlRegistryStorage.processRecord` then logs the misleading line, because `record.value()` is null but `record.key()` is not:

```
INFO  ... Discarded a (presumed) tombstone message with key: KafkaSqlMessageKey(...)
```

The ERROR does name the class, so the information is technically there. It is just one line above a reassuring INFO in a log that scrolls during replay, which is probably why this went unnoticed for so long. Tightening that INFO line is a separate concern and I have not included it here.

1PoPTRoN:
## Summary

Fixes #9532.

`RecordUsageEvent1Message` and `DeleteOldUsageEvents1Message` are submitted to the KafkaSQL journal but were never registered in `KafkaSqlMessageIndex`, so every usage event was discarded on the consume side. Schema usage telemetry has been dead on the kafkasql variant as a result.

There are 77 message classes in `storage/impl/kafkasql/messages/` and 75 were registered. This PR adds the two missing entries, a test that makes the omission impossible to reintroduce, and an end-to-end test that the mutation actually survives the journal.

## Root Cause

`KafkaSqlValueDeserializer` resolves the message type from the `mt` Kafka header via `KafkaSqlMessageIndex.lookup(...)`. For these two types the lookup returned `null`, the deserializer threw, caught its own exception, logged it and returned `null`.

The failure was quiet for three compounding reasons:

1. The message **key** deserializes fine — `KafkaSqlKeyDeserializer` never consults the index — so the record passed the null-key guard in `processRecord` and fell into the branch that treats a null value as a tombstone. A lost mutation was reported at INFO as `Discarded a (presumed) tombstone message`.
2. Neither call site waits on the coordinator (`recordUsageEvent()` at `KafkaSqlRegistryStorage:1316`, `deleteOldUsageEvents()` at `:1322`), so nothing could surface to the caller.
3. It does not self-heal. The journal replays on every startup, so the same records fail again on every restart.

The `ERROR` line from the deserializer does name the class, so the information was technically present — one line above a reassuring `INFO`, in a log that scrolls during replay.

### Why this needs tests, not just two lines

`DeleteAllOrphanedContent0Message` was found missing from the same index and fixed in #7810 ("Fix missing message index entry ... was previously missing"). This is the same defect in the same file, a second time.

Nothing links "class extends `AbstractMessage`" to "class is in the index" — it is a hand-maintained list, the compiler cannot check it, and the failure is swallowed on the consume side. The tests are the part of this PR that stops a third occurrence.

## Changes

**`storage/impl/kafkasql/serde/KafkaSqlMessageIndex.java`** (+3/−1)
- Registered `RecordUsageEvent1Message` and `DeleteOldUsageEvents1Message`, appended to the static block following the pattern used by `InsertContractAuditEntry1Message` (#8919) and the contract ruleset messages (#8866). No import change needed; the file already wildcard-imports the `messages` package.

**`storage/impl/kafkasql/KafkaSqlMessageIndexTest.java`** (new, 156 lines) — guards the index itself. Plain JUnit, no container, ~30ms.
- `testEveryMessageClassIsRegisteredInTheIndex` — scans the messages package via the context ClassLoader, filters to concrete `KafkaSqlMessage` implementations, and asserts each resolves. Dependency-free, since `app` has no classpath-scanning library on the test classpath. Uses `Class.forName(name, false, cl)` so static initializers do not run, and skips `$` inner classes so Lombok builders are not scanned. Asserts a lower bound on the number of classes discovered first, so it can never pass vacuously if the classpath layout changes.
- `testIndexResolvesEachClassToItself` — guards against two message classes sharing a simple name and silently shadowing each other on the journal.
- `testUsageEventMessagesAreRegistered` — explicit regression coverage naming the two classes.
- `testMessageKeyTypeResolvesThroughTheIndex` — the type written into the `mt` header by `AbstractMessage.getKey()` must resolve back through the index.

**`storage/impl/kafkasql/KafkaSqlUsageEventJournalTest.java`** (new, 117 lines) — added in response to review; covers the path that was actually broken. `@QuarkusTest` + `KafkasqlTestProfile`, modelled on `KafkaSqlContractAuditJournalTest`.
- `testRecordUsageEventSurvivesJournalRoundTrip` — creates a real artifact, submits via `recordUsageEvent(...)`, then awaits read-back through `getArtifactUsageMetrics(...)`. Because `KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage` with `setDelegate(sqlStore)`, that read hits the local SQL projection, so it only passes if the record went out to Kafka, returned through the consumer thread, and was applied. Asserts specific values rather than non-emptiness.
- `testDeleteOldUsageEventsSurvivesJournalRoundTrip` — records a 30-day-old event, waits for it to land, prunes with a 1-day cutoff, and awaits removal, covering `DeleteOldUsageEvents1Message` end to end.

Compatible with #7337 (removing deprecated v8/v9/v10 message classes): deleting a class removes it from both the scanned set and the index, so the invariant still holds, and the lower bound leaves headroom.

## Upgrade impact

Two consequences that are not visible in a three-line production diff:

- **Existing journals hold a backlog.** These messages were produced successfully all along; only the consume side failed. After this fix the next restart replays and applies that backlog, so operators will see a burst of usage-event inserts on first upgrade. Journal ordering is preserved, so the `record → record → deleteOld(cutoff) → record` interleaving converges to the correct state.
- **Pre-snapshot usage events are unrecoverable.** If a snapshot was taken while the bug was live it contains no usage rows, and replay resumes from the snapshot offset. Events older than the newest snapshot are permanently lost.

No schema change, no SPI change, no wire-format change. Both message classes and their JSON shape already existed and are unmodified, so no new permanent contract is introduced.

## Scope

- `apicurio.usage.telemetry.enabled` defaults to `false`, so only deployments that explicitly enabled usage telemetry are affected.
- kafkasql only. SQL calls these storage methods directly; GitOps and KubernetesOps are read-only.
- No data corruption — usage events are lost, nothing else is touched.

## Design note

I considered moving `deleteOldUsageEvents` off the journal, since #7810 did exactly that for the other scheduled cleanup jobs. It should not be: `deleteAllExpiredDownloads` and `deleteAllOrphanedContent` operate on node-local or derivable data, whereas usage events *are* journaled data. A pod pruning only its local store would have the next full replay resurrect every pruned row. Journaling the prune is what keeps it deterministic and replay-stable, so the existing design is correct and only the registration was broken.

## Test plan

Verified against a real Kafka container (Strimzi test container, single broker).

- [x] `KafkaSqlUsageEventJournalTest` — 2/2 pass with the fix
- [x] Verified **both** round-trip tests fail without the index registration: `ConditionTimeoutException`, `Expected exactly one usage summary row after the journal round trip, but got 0 ==> expected: <1> but was: <0>`
- [x] `KafkaSqlMessageIndexTest` — 4/4 pass; verified it fails on unpatched code, reporting `[DeleteOldUsageEvents1Message, RecordUsageEvent1Message]`
- [x] `KafkaSqlContractAuditJournalTest` — 1/1 pass, confirming no regression on the existing journal path
- [x] `KafkaSqlConfigurationTest`, `KafkaAdminUtilTest`, `ReadOnlyRegistryStorageTest` — all pass
- [x] `./mvnw checkstyle:check -pl app` — 0 violations
- [x] `./mvnw test-compile -pl app -am -DskipTests` — clean
- [x] Rebased onto current `main`

Storage variants: kafkasql is the only variant this touches. SQL behaviour is unchanged, and the read-only variants inherit no new methods.

### DIFF
--- a/app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlMessageIndexTest.java
+++ b/app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlMessageIndexTest.java
@@ -1,0 +1,156 @@
+package io.apicurio.registry.storage.impl.kafkasql;
+
+import io.apicurio.registry.storage.impl.kafkasql.messages.DeleteOldUsageEvents1Message;
+import io.apicurio.registry.storage.impl.kafkasql.messages.RecordUsageEvent1Message;
+import io.apicurio.registry.storage.impl.kafkasql.serde.KafkaSqlMessageIndex;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.lang.reflect.Modifier;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.List;
+
+/**
+ * Guards the KafkaSQL journal wire contract.
+ * <p>
+ * Every mutation in the KafkaSQL storage variant is journaled as a {@link KafkaSqlMessage} and replayed
+ * from the beginning of the topic on startup. The message type travels in a Kafka header as the message
+ * class's simple name (see {@link AbstractMessage#getKey()} and
+ * {@link KafkaSqlSubmitter#MESSAGE_TYPE_HEADER}), and the consumer resolves it back to a class through
+ * {@link KafkaSqlMessageIndex#lookup(String)}.
+ * <p>
+ * That index is hand-maintained, and nothing in the compiler or the CDI container links "class extends
+ * AbstractMessage" to "class is registered". A class that is submitted but not registered fails
+ * deserialization on the consume side, where the failure is swallowed and the record is discarded, so
+ * the mutation is lost silently and on every subsequent replay.
+ * <p>
+ * This has happened twice: {@code DeleteAllOrphanedContent0Message} was found unregistered in #7810, and
+ * the usage-event messages covered here. These tests turn the invariant into a build failure instead of a
+ * runtime data-loss bug.
+ */
+public class KafkaSqlMessageIndexTest {
+
+    private static final String MESSAGES_PACKAGE = "io.apicurio.registry.storage.impl.kafkasql.messages";
+
+    private static final String CLASS_SUFFIX = ".class";
+
+    /**
+     * Lower bound on the number of message classes we expect to discover. This exists so the scan below can
+     * never pass vacuously: if the classpath layout changes and the scan finds nothing, the test must fail
+     * loudly rather than silently stop guarding anything.
+     */
+    private static final int MINIMUM_EXPECTED_MESSAGE_CLASSES = 50;
+
+    @Test
+    public void testEveryMessageClassIsRegisteredInTheIndex() throws Exception {
+        List<Class<?>> messageClasses = findConcreteMessageClasses();
+
+        Assertions.assertTrue(messageClasses.size() >= MINIMUM_EXPECTED_MESSAGE_CLASSES,
+                "Expected to discover at least " + MINIMUM_EXPECTED_MESSAGE_CLASSES + " KafkaSqlMessage "
+                        + "classes in " + MESSAGES_PACKAGE + " but found " + messageClasses.size() + ". "
+                        + "The classpath scan is not finding the message classes, so this test is no longer "
+                        + "guarding the journal wire contract. Fix the scan before changing this bound.");
+
+        List<String> unregistered = new ArrayList<>();
+        for (Class<?> messageClass : messageClasses) {
+            if (KafkaSqlMessageIndex.lookup(messageClass.getSimpleName()) == null) {
+                unregistered.add(messageClass.getSimpleName());
+            }
+        }
+
+        Assertions.assertEquals(List.of(), unregistered,
+                "The following KafkaSqlMessage classes are not registered in KafkaSqlMessageIndex: "
+                        + unregistered + ". Any message submitted to the journal but missing from the index "
+                        + "fails to deserialize on the consume side, so the mutation is silently discarded "
+                        + "and lost on every replay. Add each class to the static block in "
+                        + "KafkaSqlMessageIndex.");
+    }
+
+    @Test
+    public void testIndexResolvesEachClassToItself() throws Exception {
+        for (Class<?> messageClass : findConcreteMessageClasses()) {
+            Class<? extends KafkaSqlMessage> resolved = KafkaSqlMessageIndex
+                    .lookup(messageClass.getSimpleName());
+            if (resolved != null) {
+                Assertions.assertEquals(messageClass, resolved,
+                        "KafkaSqlMessageIndex resolves '" + messageClass.getSimpleName()
+                                + "' to a different class (" + resolved.getName() + "). Two message classes "
+                                + "sharing a simple name would collide in the index and silently shadow "
+                                + "each other on the journal.");
+            }
+        }
+    }
+
+    @Test
+    public void testUsageEventMessagesAreRegistered() {
+        // Explicit regression coverage for the two classes this test class was added for. Kept separate
+        // from the scan above so the specific defect stays named even if the scan is ever refactored.
+        Assertions.assertEquals(RecordUsageEvent1Message.class,
+                KafkaSqlMessageIndex.lookup(RecordUsageEvent1Message.class.getSimpleName()),
+                "RecordUsageEvent1Message is submitted by KafkaSqlRegistryStorage.recordUsageEvent() but "
+                        + "was not registered, so all schema usage telemetry was discarded on the consume "
+                        + "side.");
+
+        Assertions.assertEquals(DeleteOldUsageEvents1Message.class,
+                KafkaSqlMessageIndex.lookup(DeleteOldUsageEvents1Message.class.getSimpleName()),
+                "DeleteOldUsageEvents1Message is submitted by "
+                        + "KafkaSqlRegistryStorage.deleteOldUsageEvents() but was not registered, so usage "
+                        + "event retention never ran on the KafkaSQL variant.");
+    }
+
+    @Test
+    public void testMessageKeyTypeResolvesThroughTheIndex() {
+        // Exercises the actual wire path end to end: the type written into the Kafka header by
+        // AbstractMessage.getKey() must be the same string the consumer looks up.
+        KafkaSqlMessage message = RecordUsageEvent1Message.builder().globalId(1L).contentId(2L)
+                .clientId("test-client").operation("READ").eventTimestamp(System.currentTimeMillis()).build();
+
+        String messageType = message.getKey().getMessageType();
+
+        Assertions.assertEquals(RecordUsageEvent1Message.class, KafkaSqlMessageIndex.lookup(messageType),
+                "The message type written to the '" + KafkaSqlSubmitter.MESSAGE_TYPE_HEADER
+                        + "' Kafka header ('" + messageType + "') does not resolve back to its own class "
+                        + "through KafkaSqlMessageIndex.");
+    }
+
+    /**
+     * Finds every concrete {@link KafkaSqlMessage} implementation in the messages package by scanning the
+     * classpath directory that holds it. Deliberately dependency-free: the app module has no classpath
+     * scanning library on the test classpath.
+     */
+    private static List<Class<?>> findConcreteMessageClasses() throws Exception {
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+        Enumeration<URL> resources = classLoader.getResources(MESSAGES_PACKAGE.replace('.', '/'));
+
+        List<Class<?>> messageClasses = new ArrayList<>();
+        while (resources.hasMoreElements()) {
+            URL resource = resources.nextElement();
+            if (!"file".equals(resource.getProtocol())) {
+                continue;
+            }
+            File[] files = new File(resource.toURI()).listFiles();
+            if (files == null) {
+                continue;
+            }
+            for (File file : files) {
+                String fileName = file.getName();
+                // Skip Lombok-generated builder inner classes, which are not message types themselves.
+                if (!fileName.endsWith(CLASS_SUFFIX) || fileName.contains("$")) {
+                    continue;
+                }
+                String className = MESSAGES_PACKAGE + "."
+                        + fileName.substring(0, fileName.length() - CLASS_SUFFIX.length());
+                // initialize=false: resolving these classes must not run their static initializers.
+                Class<?> candidate = Class.forName(className, false, classLoader);
+                if (KafkaSqlMessage.class.isAssignableFrom(candidate) && !candidate.isInterface()
+                        && !Modifier.isAbstract(candidate.getModifiers())) {
+                    messageClasses.add(candidate);
+                }
+            }
+        }
+        return messageClasses;
+    }
+}
--- a/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/serde/KafkaSqlMessageIndex.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/serde/KafkaSqlMessageIndex.java
@@ -64,7 +64,9 @@ public class KafkaSqlMessageIndex {
                 DeleteGlobalContractRuleset0Message.class,
                 InsertContractAuditEntry1Message.class,
                 MergeArtifactLabels4Message.class,
-                MergeVersionLabels5Message.class);
+                MergeVersionLabels5Message.class,
+                RecordUsageEvent1Message.class,
+                DeleteOldUsageEvents1Message.class);
     }
 
     public static Class<? extends KafkaSqlMessage> lookup(String name) {
--- a/app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlUsageEventJournalTest.java
+++ b/app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlUsageEventJournalTest.java
@@ -1,0 +1,117 @@
+package io.apicurio.registry.storage.impl.kafkasql;
+
+import io.apicurio.registry.AbstractResourceTestBase;
+import io.apicurio.registry.rest.client.models.CreateArtifactResponse;
+import io.apicurio.registry.storage.dto.SchemaUsageEventDto;
+import io.apicurio.registry.storage.dto.SchemaUsageSummaryDto;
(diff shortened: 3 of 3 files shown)


paoloantinori (commented):
Correct root cause and a genuinely useful regression test. I verified against current main that exactly these two of 77 message classes are unregistered, that `KafkaSqlValueDeserializer` (lines 60-68) swallows the lookup failure into a null value that the consumer treats as a tombstone, and that `AbstractMessage.getKey()` keys the `mt` header on the same `getSimpleName()` the index uses, so the fix and the wire-path test are both sound. The only substantive gap is the self-acknowledged absence of an end-to-end round-trip test.

- `KafkaSqlMessageIndex.java`: the two added entries are correct, in keeping with the existing registration pattern, and the wildcard import already covers them. I independently diffed all 77 classes under `storage/impl/kafkasql/messages/` against the indexed set and confirmed the bug set is exactly these two, so nothing else is still leaking through the deserializer.
- `KafkaSqlMessageIndexTest.java`: the scan is well-built. The `MINIMUM_EXPECTED_MESSAGE_CLASSES = 50` lower bound stops it passing vacuously, `initialize=false` avoids running static initializers, and skipping `$` classes correctly excludes Lombok builders. One real hole remains: none of the four tests asserts that a usage event survives a full Kafka consume/deserialize/dispatch round-trip, which is precisely the path that was broken and is still silently swallowed by `KafkaSqlValueDeserializer`'s catch-and-return-null (that swallow is pre-existing and out of scope here, but it is why this PR benefits from a round-trip assertion rather than relying solely on index-presence checks).

app/src/test/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlMessageIndexTest.java
```diff
@@ -0,0 +1,156 @@
...
+        Assertions.assertEquals(DeleteOldUsageEvents1Message.class,
+                KafkaSqlMessageIndex.lookup(DeleteOldUsageEvents1Message.class.getSimpleName()),
+                "DeleteOldUsageEvents1Message is submitted by "
+                        + "KafkaSqlRegistryStorage.deleteOldUsageEvents() but was not registered, so usage "
+                        + "event retention never ran on the KafkaSQL variant.");
+    }
+
+    @Test
+    public void testMessageKeyTypeResolvesThroughTheIndex() {
+        // Exercises the actual wire path end to end: the type written into the Kafka header by
```
paoloantinori, line 106: The index-presence and wire-header tests are necessary but not sufficient: a class can be registered and have a resolvable header type yet still fail to deserialize or dispatch correctly, and `KafkaSqlValueDeserializer` swallows exactly that failure into a tombstone (deserialize() returns null on any exception, with only an ERROR log). Since this PR is fixing a path that was silently dead, please add the container-backed round-trip test called out in the Test plan: submit a `RecordUsageEvent1Message`, force a journal consume/replay, and assert the row is present, modelled on `KafkaSqlContractAuditJournalTest`. That closes the gap the index scan cannot reach. Happy for it to land in this PR or as a follow-up issue, but the gap should be tracked either way.

1PoPTRoN:
Thanks @paoloantinori — you were right that index-presence checks were not enough, and the round-trip gap is now closed. Pushed as a separate commit so the delta is easy to review.

**`KafkaSqlUsageEventJournalTest`** (`@QuarkusTest` + `KafkasqlTestProfile`, modelled on `KafkaSqlContractAuditJournalTest`) exercises the full consume/deserialize/dispatch path rather than asserting on the topic:

- `testRecordUsageEventSurvivesJournalRoundTrip` — creates a real artifact, submits through `KafkaSqlRegistryStorage.recordUsageEvent(...)`, then awaits the read-back through `getArtifactUsageMetrics(...)`. Since `KafkaSqlRegistryStorage extends ReadOnlyDelegatingStorage` with `setDelegate(sqlStore)`, that read hits the local SQL projection, so the assertion only passes if the record actually went out to Kafka, came back through the consumer thread, and was applied. Asserts specific values — `totalFetches`, `uniqueClients`, both timestamps, and the aggregated client list — not just non-empty. `contentId` is left at 0 so the query matches on `globalId` only and the assertions stay unambiguous.
- `testDeleteOldUsageEventsSurvivesJournalRoundTrip` — records an event 30 days old, waits for it to land, prunes with a 1-day cutoff, then awaits its removal. That covers `DeleteOldUsageEvents1Message` end to end too, which the index checks alone never touched.

I was also able to resolve the container problem noted in the original test plan — it was a missing image, not a real failure. With the image pulled, **both new tests fail against a real Kafka container without the index registration and pass with it**:

```
testRecordUsageEventSurvivesJournalRoundTrip     ConditionTimeoutException
  Expected exactly one usage summary row after the journal round trip, but got 0.
  ==> expected: <1> but was: <0>
testDeleteOldUsageEventsSurvivesJournalRoundTrip ConditionTimeoutException
```

With the fix applied: `KafkaSqlUsageEventJournalTest` 2/2, `KafkaSqlMessageIndexTest` 4/4, and `KafkaSqlContractAuditJournalTest` 1/1 all green, plus `checkstyle:check -pl app` clean. The `KafkaSqlContractAuditJournalTest` run also confirms the earlier container failure was environmental rather than a regression from this change.

Agreed on leaving `KafkaSqlValueDeserializer`'s catch-and-return-null alone — it is pre-existing and out of scope here. Happy to open a separate issue for it if you think it is worth tightening, since the `INFO`-level "presumed tombstone" line is what made this defect invisible for so long.

Also rebased onto current `main` (the branch had gone `behind`) and corrected the DCO sign-off to my full name. The two commits can be squashed on merge.

paoloantinori:
Good catch following up on this. The round-trip test is exactly the gap I was pointing at: an index-presence check alone can't detect a deserializer that swallows the failure and drops the record as a tombstone, so asserting on the read-back metric (and on the prune for the retention path) is the right shape.

The classpath-scan test is a nice bonus regression guard. It turns "submitted but never registered" into a build failure rather than silent replay-time data loss, and the MINIMUM_EXPECTED_MESSAGE_CLASSES floor means the scan can't pass vacuously if it ever breaks. Looks good to me.

1PoPTRoN:
@EricWittmann — this one is ready from my side whenever you have a moment.

@paoloantinori reviewed it and signed off after the round-trip test was added. The branch is rebased onto current `main` and the diff is 3 files: 3 lines in `KafkaSqlMessageIndex` plus two test classes.

It is still sitting in `lifecycle/new`, so the full suite has not run against it yet — could you `/accept` when you get a chance so CI can verify it? Everything I could check locally is green, including both round-trip tests against a real Kafka container (verified failing without the fix, passing with it), `KafkaSqlContractAuditJournalTest` for regression on the existing journal path, and `checkstyle:check -pl app`.

Two small notes:

- The `License Compliance` status shows "4 issues found", but this PR touches no `pom.xml` or any other dependency or license file. The same status with the same count appears on unrelated PRs such as #9565, so it looks repo-wide rather than anything introduced here.
- Happy to squash the two commits into one if you would prefer that over a squash-on-merge.

No rush — just flagging it since it needs a maintainer action to move.

paoloantinori:
Confirming my sign-off still stands at the current head (18588c6): the round-trip test closes the gap I raised, and the index-presence guard makes the omission impossible to reintroduce. From my side this is ready — just needs the CI run from an /accept to verify the full suite.

1PoPTRoN:
@paoloantinori thanks for re-confirming at 18588c6.

You mentioned it just needs an /accept for the full suite to run - is that something you're able to trigger, or better left with @EricWittmann as assignee? Happy either way, just wasn't sure which is the right path.

paoloantinori:
Better left with @EricWittmann as assignee — the /accept is his call since he's driving the merge. From my side nothing is blocking: my sign-off stands at 18588c6, the round-trip test closes the gap I raised, and the index guard keeps the omission from coming back. Once the full suite runs green on his accept, this should be good to go.

1PoPTRoN:
@EricWittmann gentle ping

1PoPTRoN:
@EricWittmann @paoloantinori — rebased onto current `main` (`80b1bbe` — 73 commits ahead of previous head) and force-pushed as `18dd747` (same 2 commits, now clean).

- `mergeable_state` was `BEHIND` → now `blocked` (will go green once GitHub re-evaluates — no conflicts, `git diff` only the 3 lines in `KafkaSqlMessageIndex` + 2 test files).
- Sign-off intact on both commits.
- Tried `/retry` as author to reconcile, but `Verification Gate` still `failure: PR is not in a testable lifecycle state` because the PR is still `lifecycle/new`. Your `/accept` on `2026-08-21` did not transition labels (still `lifecycle/new + waiting-on-maintainer`), so the full suite (Build → 9 unit shards → 13 integration matrix) has never run on this SHA. Could you re-issue `/accept` to move to `ready-for-review` and approve the fork workflow?

Also: `License Compliance: 4 issues` is repo-wide (also on #9565, no `pom.xml` touched here) — and `paoloantinori`'s `COMMENTED` reviews from `2026-08-13/08-15` already cover the fix, but `reviewDecision` is still empty — a formal **Approve** click would clear that gate.

Tests verified locally before rebase: `KafkaSqlMessageIndexTest` 4/4, `KafkaSqlUsageEventJournalTest` 2/2 (fail without index, pass with), `checkstyle:check` clean. Happy to squash on merge if preferred.

PR #9536, opened 2026-08-13
fix(rules): compare agent card definitions, not just identifiers

### LINKED ISSUE #9455: AgentCardCompatibilityChecker misses definition-level breaking changes (security scheme type, protocolVersion) — only names/keys are compared
## Description

**Registry Version**: snapshot (`main`, commit `0d50655ba`)
**Persistence type**: any — the defect is in `schema-util/common` and is storage-independent

`AgentCardCompatibilityChecker` governs backward compatibility for `AGENT_CARD` artifacts. It already looks past identifiers in two places: a boolean capability that flips to `false` is caught, and so is a changed `protocolVersion` on an interface that is kept. That treatment is **incomplete**. Everywhere else — skill `id`s, interface `url|protocolBinding` keys, capability extension uris, security-scheme names, and mode strings — only the identifier is compared, and the *definition behind* it can change freely. As a result, several changes that break existing A2A clients are reported as **compatible**, so a new version is admitted under a `COMPATIBILITY: BACKWARD` rule when it should be rejected.

The same shallow, name-only comparison pattern also appears in `McpToolCompatibilityChecker` (#9297, `outputSchema` ignored).

### 1. Security scheme type change is not detected

`schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java:155-167`
```java
private void checkSecuritySchemeRemovals(JsonNode existing, JsonNode proposed, ...) {
    Set<String> existingSchemes = extractSecuritySchemeNames(existing);   // NAMES only
    Set<String> proposedSchemes = extractSecuritySchemeNames(proposed);
    for (String scheme : existingSchemes) {
        if (!proposedSchemes.contains(scheme)) { ... SECURITY_SCHEME_REMOVED ... }
    }
}
```
`extractSecuritySchemeNames` collects `securitySchemes` field **names** only. Keeping the same name (`"bearer"`) but changing its definition from HTTP Bearer auth (`{"type":"httpAuth","scheme":"Bearer"}`) to a header API key (`{"type":"apiKey","name":"X-API-Key","location":"header"}`) is reported compatible. Every existing client that authenticates with a Bearer token now fails, but the rule admits the change.

### 2. Card-level `protocolVersion` change is not detected

`checkInterfaceProtocolVersionChanges` (lines `75-107`) only compares the **per-interface** `protocolVersion`. The **top-level** `AgentCard.protocolVersion` — which the checker's own extractor and `RegistryAgentCardBuilder` both populate — is never compared, so bumping the card's `protocolVersion` from `"1.0"` to `"2.0"` passes as compatible.

### 3. Dropping an interface's `protocolVersion` is silently allowed

`AgentCardCompatibilityChecker.java:89-99`
```java
String existingVersion = getTextValue(existingIface, "protocolVersion");
if (url == null || binding == null || existingVersion == null) {
    continue;                                  // guards the EXISTING interface only
}
...
if (url.equals(pUrl) && binding.equals(pBinding)
        && pVersion != null && !existingVersion.equals(pVersion)) {   // <-- skips when PROPOSED drops the version
```
When the proposed interface keeps the same `url` + `protocolBinding` but **removes** its `protocolVersion`, `pVersion` is `null`, the `pVersion != null` guard short-circuits, and no difference is reported — the client loses a version guarantee it previously had.

### Why this is an oversight, not by design

The class Javadoc lists "Changing protocolVersion on existing interface: Backward incompatible" as a rule it intends to enforce, yet cases 2 and 3 defeat it. More directly, the sibling `PromptTemplateCompatibilityChecker` in the same package already inspects the definition *behind* an identifier — `checkVariableTypeChange` and `checkEnumNarrowing` (lines `98-145`) compare a variable's `type` and `enum`, not just its name. Agent cards — the artifact that A2A clients discover and bind to — currently get weaker compatibility governance than prompt-template variables.

### Steps to Reproduce

Unit-level, verified against `main` @ `0d50655ba`. All three assertions currently fail (checker reports **compatible**):

```java
// Case 1: security scheme "bearer" changes httpAuth Bearer -> apiKey header
// Case 2: top-level "protocolVersion" changes "1.0" -> "2.0" (interfaces unchanged)
// Case 3: an interface keeps url+protocolBinding but drops "protocolVersion"
CompatibilityExecutionResult r = new AgentCardCompatibilityChecker().testCompatibility(
        CompatibilityLevel.BACKWARD, List.of(existingCard), proposedCard, Map.of());
assertFalse(r.isCompatible());   // FAILS in all three cases
```

I have a `AgentCardDefinitionChangeGapReproTest` reproducing all three (3 failing assertions) that I can attach to the fix PR.

At the REST level: create an `AGENT_CARD` artifact with a `COMPATIBILITY: BACKWARD` rule, then publish a new version applying any of the three changes above — the version is accepted instead of rejected with a rule violation.

**Expected:** changing a security scheme's type/definition, changing the card-level `protocolVersion`, or dropping an existing interface's `protocolVersion` is backward-incompatible and produces a `CompatibilityDifference` (rejecting the new version under a BACKWARD rule).

**Actual:** `testCompatibility` returns compatible; the breaking version is admitted. No error, no log — the change is silently accepted, which is the bug.

### Severity

Correctness / governance. Affected deployments are those that configure a `BACKWARD`/`FULL` compatibility rule on `AGENT_CARD` artifacts. For those users the rule gives a false guarantee: breaking agent-card changes pass review.

This is **not** gated behind `apicurio.a2a.enabled`. That property is read only by `RegistryAgentCardBuilder` and `WellKnownResourceImpl`, which serve the registry's *own* card at `/.well-known/`. The artifact type — and with it the compatibility checker — is registered unconditionally in `StandardArtifactTypeProviderRegistry`:

```java
PROVIDERS.put(ArtifactType.AGENT_CARD, new ProviderConfig.Builder()
        .compatibilityChecker(AgentCardCompatibilityChecker::new)
        ...
```

So a registry running default configuration can store `AGENT_CARD` artifacts and hit all three cases. Not a crash, and it does require a compatibility rule to be configured.

### Possible fix

For identifiers present in **both** versions, compare the definition behind them, mirroring `PromptTemplateCompatibilityChecker`:
- security schemes: flag a change to `type` (and `scheme`/`location`/`name`) for a scheme kept by name;
- compare the top-level `protocolVersion`;
- treat "proposed interface drops `protocolVersion`" as a change (remove the `pVersion != null` short-circuit, or report a dedicated difference).

Each should be a `CompatibilityDifference` with its own context path (`/securitySchemes`, `/protocolVersion`, `/supportedInterfaces`), added to the existing `AgentCardCompatibilityDifference.Type` enum.

I'd be happy to work on this — I have the failing repro test and a patch following the existing `PromptTemplateCompatibilityChecker` pattern.

### Relevant files

- `schema-util/common/.../rules/compatibility/AgentCardCompatibilityChecker.java` — the checker (bug: name-only comparison)
- `schema-util/common/.../rules/compatibility/PromptTemplateCompatibilityChecker.java` — safe sibling (`checkVariableTypeChange`, `checkEnumNarrowing`)
- `schema-util/common/.../content/extract/AgentCardStructuredContentExtractor.java` — confirms `protocolVersion` and `securityscheme` are first-class card fields

### Related

- **#9297** — `McpToolCompatibilityChecker` ignores `outputSchema` (same shallow-comparison class, MCP artifact type).
- **#9173 / PR #9331** — `checkCapabilityRemovals` skipped non-boolean capabilities, so removing a declared `capabilities.extensions` entry passed. Same checker, a different check. **Merged 2026-08-12**; the three gaps above are still present on top of it (repro re-verified against `main` @ `89f40491`, which contains that merge).
- **#7215 / #7884** — A2A agent-card + visibility/entitlement model (origin of `AGENT_CARD`).
- **#8424** — federated AI agent search (agent-card discovery/governance area).
- **#8184 / PR #8634** — compatibility-checker type-hierarchy refactor (open). Independent of this fix, which uses the existing `AgentCardCompatibilityDifference.Type` enum; #8634 can reshape it alongside the other checkers.

Vidhaankhare16:
Fixes #9455

`AgentCardCompatibilityChecker` already looks past identifiers in two places: a boolean capability that flips to `false` is caught, and so is a changed `protocolVersion` on an interface that is kept. That treatment is incomplete. Everywhere else — skill ids, interface `url|protocolBinding` keys, capability extension uris, security scheme names, mode strings — only the identifier is compared, so three changes that break existing A2A clients pass as compatible under a BACKWARD rule.

### Security scheme definitions

A security scheme kept by name can be redefined completely. `checkSecuritySchemeRemovals` diffs the output of `extractSecuritySchemeNames`, which collects `securitySchemes` field names, so changing `"bearer"` from `{"type":"httpAuth","scheme":"Bearer"}` to `{"type":"apiKey","name":"X-API-Key","location":"header"}` produces no differences at all. Every client authenticating with a Bearer token breaks.

Compare the fields that decide how a client authenticates: `type`, `scheme`, `bearerFormat`, `name`, `location`, `oauth2MetadataUrl` and `openIdConnectUrl`. A field the existing card declared that changes or disappears is a difference. `description` is documentation and is not compared, and fields outside that list are ignored deliberately — the schema permits vendor extensions on a scheme, and churn in one of those must not fail a publish.

`bearerFormat` is the debatable member: OpenAPI calls it documentation, but the A2A schema gives it no such wording and lists it beside `scheme`, and a client reading JWT claims locally does break when the format goes opaque.

A retained name can also hide the definition disappearing entirely. `"bearer": {...}` becoming `"bearer": null` or a bare string keeps the name, so the removal check stays quiet. `validateSecuritySchemes` would reject that, but validity rules are opt-in, so the card reaches the compatibility checker either way.

`flows` is typed as a bare object in the schema and the validator never looks inside it, so it is compared structurally rather than by known field names: whatever the existing card declared must still be there, unchanged. Comparing the whole `flows` node instead would report additions as breaking, so adding a grant type or a scope stays compatible. Differences are reported at the outermost point where the two stop matching, so withdrawing a grant type is one difference rather than one per field beneath it. Arrays inside `flows` are treated as sets of offered values, the way `checkModeRemovals` already treats `defaultInputModes`.

A `scopes` map is the one place under `flows` where the value is documentation rather than contract — A2A types it as name to "a short description for it" — so only the scope names are compared. Withdrawing a scope is still reported; rewording one is not. Comparing those values would contradict the `description` exemption above and would judge the map form more harshly than the array form.

### Dropping an interface's protocolVersion

An interface that keeps its `url` and `protocolBinding` but drops its `protocolVersion` was silently allowed. The guard read:

```java
if (url.equals(pUrl) && binding.equals(pBinding)
        && pVersion != null && !existingVersion.equals(pVersion))
```

so a null `pVersion` short-circuited the comparison and nothing was reported, even though the client loses a version guarantee it had. Split into the dropped case and the changed case, keeping the existing `PROTOCOL_VERSION_CHANGED` wording for the latter.

That case only surfaces when the VALIDITY rule is unset, `NONE` or `SYNTAX_ONLY`, since `protocolVersion` is required for an interface by both `validateSupportedInterfaces` and `supportedInterfaces.items.required` in `a2a-agent-card-schema.json`. Validity rules are opt-in, so a compatibility checker cannot assume one ran.

### Card-level protocolVersion

The card-level `protocolVersion` is never compared either, so bumping it from `"1.0"` to `"2.0"` passes while the same change on an interface is rejected. `RegistryAgentCardBuilder` fills the card-level field and each interface's from the same `a2aConfig.getProtocolVersion()`, so it is a real declaration rather than a vestigial one. It is optional, so a card that never declared one promises nothing to take away and adding one stays compatible.

This is the case I am least attached to — one can argue the field exists in order to change — and it is one self-contained method and its tests if you would rather it came out.

### Direction semantics and difference types

All three checks are written one-directional: existing lost or changed something. `AbstractCompatibilityChecker` calls `isBackwardsCompatibleWith(existing, proposed)` for BACKWARD and swaps the arguments for FORWARD, so FORWARD and FULL follow without a symmetric check, which would have flagged legitimate additions.

#8634 has since merged and removed the per-type difference classes, so each violation is a `SimpleCompatibilityDifference` carrying its context path directly: security scheme changes under `/securitySchemes`, an interface that drops its `protocolVersion` under `/supportedInterfaces`, and the card-level `protocolVersion` under `/protocolVersion`, since that field is not part of `/supportedInterfaces`. Those are the paths the deleted `AgentCardCompatibilityDifference.Type` constants carried.

### Tests

Adding a security scheme, editing a scheme's description, adding a grant type or a scope, and declaring a card-level `protocolVersion` where there was none all remain compatible, each with a test.

27 tests added to `AgentCardCompatibilityCheckerTest`, and the class Javadoc rule list updated to match. Among them: an interface that never declared a `protocolVersion` (both-sides-absent and proposed-adds-one stay compatible), a duplicated `url|protocolBinding` where one entry keeps the version, and a `protocolVersion` that turns non-textual, which must report a change rather than a removal.

The same identifier-only comparison pattern appears in `McpToolCompatibilityChecker`, which is out of scope here.

### DIFF
--- a/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java
+++ b/schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java
@@ -6,6 +6,7 @@ import io.apicurio.registry.content.TypedContent;
 
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -21,8 +22,15 @@ import java.util.Set;
  * - Removing capability extensions (by uri): Backward incompatible
  * - Removing interfaces (by url+protocolBinding): Backward incompatible
  * - Changing protocolVersion on existing interface: Backward incompatible
+ * - Removing protocolVersion from an existing interface: Backward incompatible
+ * - Changing the card-level protocolVersion: Backward incompatible
+ * - Adding a card-level protocolVersion where there was none: Always compatible
  * - Adding security schemes: Always compatible
  * - Removing security schemes: Backward incompatible
+ * - Changing an existing security scheme's definition: Backward incompatible
+ * - Changing an existing security scheme's description: Always compatible
+ * - Withdrawing an OAuth2 scope: Backward incompatible
+ * - Rewording an OAuth2 scope's description: Always compatible
  * - Adding input/output modes: Always compatible
  * - Removing input/output modes: Backward incompatible
  */
@@ -34,11 +42,24 @@ public class AgentCardCompatibilityChecker
     private static final String CONTEXT_CAPABILITIES = "/capabilities";
     private static final String CONTEXT_CAPABILITY_EXTENSIONS = "/capabilities/extensions";
     private static final String CONTEXT_SECURITY_SCHEMES = "/securitySchemes";
+    private static final String CONTEXT_PROTOCOL_VERSION = "/protocolVersion";
     private static final String CONTEXT_MODES = "/modes";
     private static final String CONTEXT_DOCUMENT = "/document";
     private static final String MSG_WAS_REMOVED = "' was removed";
 
     private static final ObjectMapper mapper = new ObjectMapper();
+
+    /**
+     * Security scheme fields whose value defines how a client must authenticate. The description
+     * field is documentation, and fields outside this list are ignored on purpose: the schema
+     * permits vendor extensions, and churn in one of those must not fail a publish.
+     */
+    private static final List<String> SECURITY_SCHEME_FIELDS = List.of("type", "scheme",
+            "bearerFormat", "name", "location", "oauth2MetadataUrl", "openIdConnectUrl");
+
+    private static final String FLOWS_FIELD = "flows";
+
+    private static final String SCOPES_FIELD = "scopes";
 
     @Override
     protected Set<SimpleCompatibilityDifference> isBackwardsCompatibleWith(String existing,
@@ -53,6 +74,8 @@ public class AgentCardCompatibilityChecker
             checkSkillRemovals(existingNode, proposedNode, differences);
             checkCapabilityRemovals(existingNode, proposedNode, differences);
             checkSecuritySchemeRemovals(existingNode, proposedNode, differences);
+            checkSecuritySchemeDefinitionChanges(existingNode, proposedNode, differences);
+            checkCardProtocolVersionChange(existingNode, proposedNode, differences);
             checkModeRemovals(existingNode, proposedNode, "defaultInputModes", "input mode",
                     differences);
             checkModeRemovals(existingNode, proposedNode, "defaultOutputModes", "output mode",
@@ -93,24 +116,53 @@ public class AgentCardCompatibilityChecker
         for (JsonNode existingIface : existingInterfaces) {
             String url = getTextValue(existingIface, "url");
             String binding = getTextValue(existingIface, "protocolBinding");
-            String existingVersion = getTextValue(existingIface, "protocolVersion");
+            String existingVersion = getComparableValue(existingIface, "protocolVersion");
 
             if (url == null || binding == null || existingVersion == null) {
                 continue;
             }
 
-            for (JsonNode proposedIface : proposedInterfaces) {
-                String pUrl = getTextValue(proposedIface, "url");
-                String pBinding = getTextValue(proposedIface, "protocolBinding");
-                String pVersion = getTextValue(proposedIface, "protocolVersion");
+            // Neither the schema nor validateSupportedInterfaces requires url+protocolBinding to be
+            // unique, so weigh every matching entry before concluding: one of them keeping the
+            // version means nothing was withdrawn, and duplicates must not multiply the difference.
+            boolean matched = false;
+            boolean versionRetained = false;
+            String changedVersion = null;
 
-                if (url.equals(pUrl) && binding.equals(pBinding)
-                        && pVersion != null && !existingVersion.equals(pVersion)) {
-                    differences.add(new SimpleCompatibilityDifference(
-                            "Protocol version changed from '" + existingVersion + "' to '"
-                                    + pVersion + "' for interface " + url + " (" + binding + ")",
-                            CONTEXT_INTERFACES));
+            for (JsonNode proposedIface : proposedInterfaces) {
+                if (!url.equals(getTextValue(proposedIface, "url"))
+                        || !binding.equals(getTextValue(proposedIface, "protocolBinding"))) {
+                    continue;
                 }
+
+                matched = true;
+                String pVersion = getComparableValue(proposedIface, "protocolVersion");
+                if (existingVersion.equals(pVersion)) {
+                    versionRetained = true;
+                    break;
+                }
+                if (pVersion != null) {
+                    changedVersion = pVersion;
+                }
+            }
+
+            // An interface with no match at all is already reported as an interface removal.
+            if (!matched || versionRetained) {
+                continue;
+            }
+
+            if (changedVersion != null) {
+                differences.add(new SimpleCompatibilityDifference(
+                        "Protocol version changed from '" + existingVersion + "' to '"
+                                + changedVersion + "' for interface " + url + " (" + binding + ")",
+                        CONTEXT_INTERFACES));
+            } else {
+                // A retained interface that drops its protocolVersion withdraws a guarantee the
+                // client had. The validity rule requires the field, but it is opt-in, so the card
+                // can reach this checker without it.
+                differences.add(new SimpleCompatibilityDifference(
+                        "Protocol version '" + existingVersion + "' was removed from interface "
+                                + url + " (" + binding + ")", CONTEXT_INTERFACES));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Vidhaankhare16:
@EricWittmann @carlesarnal - opened this for #9455, please take a look whenever you get a chance. Thank you!

paoloantinori (commented):
One blocker before merge: the restructured interface protocolVersion check throws an NPE when the existing interface omits protocolVersion, and reports a false-positive removal when both sides omit it, on input that is reachable whenever the opt-in validity rule is off. The security-scheme definition comparison is well-factored and null-safe, with tests that assert specific messages and counts.

- `AgentCardCompatibilityChecker.java`, `checkInterfaceProtocolVersionChanges` (lines 129-133): `existingVersion` can be null. The new `checkCardProtocolVersionChange` proves `getTextValue` returns null for an absent protocolVersion, because it guards with `if (existingVersion == null) return;`, yet this method has no equivalent guard. When both interfaces omit the field, the `pVersion == null` branch reports a phantom removal; when the proposed interface adds one, `existingVersion.equals(pVersion)` NPEs. Add `if (existingVersion == null) continue;` before the pVersion checks.
- `AgentCardCompatibilityCheckerTest.java`: no test covers an existing interface that omits protocolVersion (validity off). Add cases for both-sides-absent, which should be compatible, and proposed-adds-one; both currently fail or throw.
- `checkSecuritySchemeDefinitionChanges`, `checkRetainedValues`, and `checkRetainedArrayValues` handle null, JSON-null, scalar, and non-textual field values correctly, and additions to flows/scopes stay compatible under BACKWARD, matching the established scheme-addition semantics. No issues there.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java
```diff
@@ -97,8 +115,19 @@ private void checkInterfaceProtocolVersionChanges(JsonNode existing, JsonNode pr
...
+                }
+
+                // A retained interface that drops its protocolVersion withdraws a guarantee the
+                // client had, so it is treated the same as changing it. The validity rule requires
+                // the field, but that rule is opt-in, so the card can reach this checker without it.
+                if (pVersion == null) {
+                    differences.add(new AgentCardCompatibilityDifference(
+                            AgentCardCompatibilityDifference.Type.PROTOCOL_VERSION_REMOVED,
+                            "Protocol version '" + existingVersion + "' was removed from interface "
+                                    + url + " (" + binding + ")"));
```
paoloantinori: `existingVersion` can be null here. `getTextValue` returns null for an absent protocolVersion, which `checkCardProtocolVersionChange` itself relies on (it guards `if (existingVersion == null) return;`), and the validity rule that would force the field is opt-in. Without a matching guard, two cases break: when both interfaces omit protocolVersion the `pVersion == null` branch reports a phantom removal, and when the proposed interface adds one, `existingVersion.equals(pVersion)` throws an NPE. Add `if (existingVersion == null) continue;` before the pVersion checks, mirroring the card-level method, so an interface that never declared a version promises nothing to take away.

Vidhaankhare16:
Thanks for the review.

On the `protocolVersion` NPE — the guard is already there, just in the outer loop rather than beside the `pVersion` checks. `checkInterfaceProtocolVersionChanges`, line 109:

```java
for (JsonNode existingIface : existingInterfaces) {
    String url = getTextValue(existingIface, "url");
    String binding = getTextValue(existingIface, "protocolBinding");
    String existingVersion = getTextValue(existingIface, "protocolVersion");

    if (url == null || binding == null || existingVersion == null) {
        continue;
    }
```

An interface whose existing `protocolVersion` is absent is skipped before the inner loop runs, so `existingVersion.equals(pVersion)` is never reached with a null receiver, and the `pVersion == null` branch is never entered for it. That guard predates this PR — I only restructured the condition inside the inner loop.

The coverage gap was a fair catch though, nothing tested that path. Added both cases you asked for:

- `testBackwardCompatibleWhenBothInterfacesOmitProtocolVersion` — both sides absent, stays compatible
- `testBackwardCompatibleWhenProposedInterfaceAddsProtocolVersion` — proposed declares one, stays compatible

Both pass on the current head (45 tests green). I also checked they are load-bearing rather than vacuous: dropping `existingVersion == null` from the line-109 guard makes both fail, which is exactly the behaviour you described. It surfaces as a spurious difference rather than a crash, since `isBackwardsCompatibleWith` catches `Exception` and converts it to a `PARSE_ERROR`.

paoloantinori:
Thanks for walking me through the control flow. You are right, I missed the guard at line 109 in the outer loop. `if (url == null || binding == null || existingVersion == null) continue;` handles both the NPE case (existing interface without `protocolVersion`) and the both-null false-positive in one place, since `existingVersion` being null skips the inner loop body entirely. The new `PROTOCOL_VERSION_REMOVED` branch then only fires for the genuinely-changed shape. Confirmed against the head `8bdda06`. With that addressed I have nothing further on this PR.

aclfe (commented):
Ran the branch locally on JDK 21, all 45 tests pass. Flagging that CI hasn't actually run them though, the gate failed on `LIFECYCLE_READY: false` since this is a fork PR waiting on `/accept`.

Good PR overall. `SECURITY_SCHEME_FIELDS` matches the schema exactly and the direction semantics are right. One thing to fix before merge plus two smaller ones, left inline.

On the card level `protocolVersion` you offered to drop: keep it. `RegistryAgentCardBuilder` fills it and each interface from the same `a2aConfig.getProtocolVersion()`, so treating them differently is the real inconsistency.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java
```diff
@@ -188,6 +217,180 @@ private void checkSecuritySchemeRemovals(JsonNode existing, JsonNode proposed,
...
+    private void checkRetainedValues(String schemeName, String path, JsonNode existing,
+            JsonNode proposed, Set<AgentCardCompatibilityDifference> differences) {
+        if (proposed == null) {
+            differences.add(new AgentCardCompatibilityDifference(
+                    AgentCardCompatibilityDifference.Type.SECURITY_SCHEME_CHANGED,
+                    "Security scheme '" + schemeName + "' no longer declares '" + path + "'"));
+            return;
+        }
+
+        if (existing.isObject() && proposed.isObject()) {
```
aclfe, line 330: This walks every leaf under `flows`, so editing an OAuth2 scope description gets rejected:

```
"scopes": { "read": "Read access" }  ->  { "read": "Read-only access" }

[/securitySchemes] Security scheme 'oauth' field 'flows/clientCredentials/scopes/read'
                   changed from 'Read access' to 'Read-only access'
```

A2A types scopes as `map<string, string>`, "a map between the scope name and a short description for it", so the value is documentation. That contradicts the `description` exemption you built at the scheme level and the Javadoc line you added.

It's also backwards against itself: array valued scopes get the lenient set treatment in `checkRetainedArrayValues`, while the canonical map form gets compared exactly. The shape that breaks is the one in your own `OAUTH_SCHEME` constant.

Fix is to compare scope names only when the path ends in `/scopes`. I tried it locally, all 45 tests still pass, a genuine scope removal is still caught (`no longer offers 'write' in 'flows/clientCredentials/scopes'`), checkstyle clean.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java
```diff
@@ -97,8 +115,19 @@ private void checkInterfaceProtocolVersionChanges(JsonNode existing, JsonNode pr
...
-                if (url.equals(pUrl) && binding.equals(pBinding)
-                        && pVersion != null && !existingVersion.equals(pVersion)) {
+                if (!url.equals(pUrl) || !binding.equals(pBinding)) {
+                    continue;
+                }
+
+                // A retained interface that drops its protocolVersion withdraws a guarantee the
+                // client had, so it is treated the same as changing it. The validity rule requires
+                // the field, but that rule is opt-in, so the card can reach this checker without it.
+                if (pVersion == null) {
```
aclfe: Two things here.

Duplicate interfaces: nothing enforces that url+protocolBinding is unique, not the schema, not `validateSupportedInterfaces`. If the proposed card lists the same interface twice and only one copy carries `protocolVersion`, this branch fires even though a matching entry keeps it. The old `pVersion != null` guard made that shape unreachable, so it's new here. Collect the matches first and report only if none of them keeps the version.

Second, `getTextValue` returns null for anything non textual, so `"protocolVersion": 2` lands in this branch and reports "was removed" when the value actually changed, under the wrong enum constant.

schema-util/common/src/main/java/io/apicurio/registry/rules/compatibility/AgentCardCompatibilityChecker.java
```diff
@@ -188,6 +217,180 @@ private void checkSecuritySchemeRemovals(JsonNode existing, JsonNode proposed,
...
+
+    /**
+     * The card-level protocolVersion is the version the agent as a whole speaks -
+     * RegistryAgentCardBuilder fills it from the same config value as the per-interface one - so a
+     * client can rely on it without reading supportedInterfaces. It is optional, so an existing
+     * card that did not declare one promised nothing that a proposed card could take away.
+     */
+    private void checkCardProtocolVersionChange(JsonNode existing, JsonNode proposed,
+            Set<AgentCardCompatibilityDifference> differences) {
+        String existingVersion = getTextValue(existing, "protocolVersion");
```
aclfe: Same non textual issue: `"protocolVersion": 2` gives `The agent protocolVersion was removed (was '1.0')` when it changed. You set the opposite rule for scheme fields with `renderValue`, and `testBackwardIncompatibleSecuritySchemeFieldBecomingNonTextual` asserts it must not be described as removed. Worth the same treatment here.

Vidhaankhare16:
Thanks, all three were real. Reproduced each one before changing anything, and your quoted output for the scopes case matched character for character. Fixed in `177cc67`.

**Scopes.** You are right that it contradicted itself. Comparing the map values rejected a reworded description while the array form got the lenient set treatment, and the shape that broke was the one in my own `OAUTH_SCHEME` constant. A `scopes` map is now compared by name only. Withdrawing a scope still reports `Security scheme 'oauth' no longer offers 'write' in 'flows/clientCredentials/scopes'`, which I think is the message you had locally. Added `testBackwardCompatibleEditingOAuth2ScopeDescription` and `testBackwardIncompatibleRemovingMapValuedOAuth2Scope`, and a rule line in the class Javadoc so the exemption is documented rather than implied.

**Duplicate interfaces.** Confirmed, and confirmed it is new here — the old `pVersion != null` guard did make that shape unreachable. The loop now surveys every entry matching `url|protocolBinding` before concluding: one of them keeping the version means nothing was withdrawn, otherwise a differing version reports a change, otherwise a removal. That also stops duplicates multiplying the difference, which the old shape did. Covered by `testBackwardCompatibleDuplicateInterfaceRetainingProtocolVersion` and a companion asserting a genuine removal is still reported exactly once when no copy keeps it.

**Non-textual `protocolVersion`.** Also right, at both sites, and you are right that it contradicted `testBackwardIncompatibleSecuritySchemeFieldBecomingNonTextual`. I had reasoned about whether a numeric bump should be *detected* and missed that it was being actively *mislabelled* as a removal under the wrong difference type. `getSchemeFieldValue` is now `getComparableValue` and both `protocolVersion` sites go through it, so one helper governs all three.

51 tests green, checkstyle clean on both modules. I also checked the new tests are load-bearing rather than vacuous — reverting each of the three fixes individually makes exactly the intended tests fail and nothing else, except the duplicate-interface one, which takes out 34 tests because every retained interface then reports a phantom removal.

On the card-level `protocolVersion`: agreed, and thanks for making the argument. Keeping it.

paoloantinori:
Thanks for the detailed write-up, and for reproducing each one before touching the code. Read the updated diff and the three fixes look right:

- `checkInterfaceProtocolVersionChanges` now tracks `matched` / `versionRetained` / `changedVersion` through the loop and uses the outer `existingVersion == null` guard, so both the NPE (missing `protocolVersion` on the existing side) and the false-positive removal (both sides omit it) are gone. The new `PROTOCOL_VERSION_REMOVED` case for a retained interface that drops the version is the correct incompatibility.
- `checkRetainedScopeNames` compares names only, which is consistent with how `SECURITY_SCHEME_FIELDS` exempts `description`: rewording a scope's blurb no longer contradicts the description exemption applied elsewhere.
- `checkSecuritySchemeDefinitionChanges` walks the declared fields and `flows` structurally, so a retained name with a swapped `type`/`scheme` is now caught the way the PR description intended.

Vidhaankhare16:
Removed an issue reference from the description — it was an aside noting the same pattern in `McpToolCompatibilityChecker`, and triage read it as a second linked issue and as a duplicate of the open PR against it. Nothing here touches that file; the only linked issue is #9455.

That leaves the maintainer-approval half, which needs a comment on the issue itself. Asked there.

Vidhaankhare16:
#8634 landed while this was open and removed the per-type difference classes, so this needed reworking. I have rebased onto current main and ported the checks: every difference is now a SimpleCompatibilityDifference carrying the context path that the deleted AgentCardCompatibilityDifference.Type held, and the transform() override goes with the base class method that declared it.

The branch is now a single signed commit, which also clears the DCO check and makes it rebaseable again.

Locally: 51 agent card tests green, 155 across all compatibility suites including the new AbstractCompatibilityCheckerTest and SimpleCompatibilityDifferenceTest, and checkstyle clean on both schema-util modules.

Also, the card-level protocolVersion is reported against /protocolVersion rather than /document, matching the context the deleted enum assigned to CARD_PROTOCOL_VERSION_CHANGED. The existing test asserts that, and it caught the mistake when I first got it wrong.

The earlier review comments will show as outdated against the rewritten history. The code they covered is unchanged apart from the port.

Vidhaankhare16:
This is green on 654a183 and blocked on review. Validate PR, SonarCloud, Quick Verify and Quick Check Gate all pass. Unit Tests, SDK Verification and the remaining jobs show as skipped rather than failing, because the full suite is gated on an approving review and cannot run until the PR has one.

@paoloantinori you reviewed this on 13 August. The code you covered is unchanged apart from the port onto SimpleCompatibilityDifference that #8634 forced. Would you be able to take another look?

paoloantinori:
Picking this back up. First, the important part: this has been ready since 13 August and I am sorry it sat. I confirmed the current head `654a183c3` is equivalent to `177cc67`, the commit I signed off on. Diffing the two shows only the port to the new `CONTEXT_*` string constants and the extracted `MSG_WAS_REMOVED`, with no line of the three fixes changed. The red X on this PR is the Verification Gate exiting 1 by design for unaccepted authors, not a test failure, and the `triage/red` label is stale from 19 August.

One new thing I did not catch in August, and it is in code this PR adds. `renderValue` collapses types:

```java
node.isTextual() ? node.asText() : node.toString()
```

`TextNode("2")` and `IntNode(2)` both render to `"2"`, so `getComparableValue` calls them equal and a `protocolVersion` changing JSON type is reported as no change. `testBackwardIncompatibleCardProtocolVersionBecomingNonTextual` passes only because `"1.0"` and `2` differ as strings, so it never exercises the collision. The same applies to every field compared through `getComparableValue`. Comparing `JsonNode.equals` alongside the rendered form closes it, and it also fixes the inverse: `renderValue` on an object uses `ObjectNode.toString()`, which is field order sensitive where `JsonNode.equals` is not, so reordering keys in an object-valued scheme field currently over-reports a change.

Three test asks, all small:

- The card-level `protocolVersion` removal branch has no test. Three of the four quadrants are covered; the missing one is an existing card that declares a version and a proposed card that drops it, which is the likeliest real edit.
- `name`, `oauth2MetadataUrl` and `openIdConnectUrl` in `SECURITY_SCHEME_FIELDS` are never exercised as changes. `name` is the load-bearing one, since for an apiKey scheme it is the header a client must send. Drop it from that list today and the suite stays green.
- `testBackwardIncompatibleDuplicateInterfacesAllDroppingProtocolVersion` asserts a count and nothing else. It is the only new incompatible-case test without a message assertion, and it guards the logic this PR rewrote.

One judgment call rather than a defect: vendor extension keys nested inside `flows` are compared strictly by `checkRetainedValues`, while `SECURITY_SCHEME_FIELDS` deliberately ignores them at the scheme level. Both behaviours are documented in your javadoc, so this is a tension between two stated choices. Your call, but worth a sentence either way.

Coverage otherwise is better than most contributions this repo sees. Every new method has direct coverage, 15 of your 27 new tests assert an exact count plus a verbatim message, and none assert a bare `assertFalse`. Checking that reverting each fix broke exactly the intended tests was the right instinct.

Separately, and not for this PR: several adjacent card surfaces are not compared at all, including `securityRequirements`, which means an open agent becoming OAuth-gated passes as backward compatible. I will open issues for those.

Vidhaankhare16:
/unstale

Sorry for the delays @EricWittmann @paoloantinori , I was caught up with my end semester exams, will get back to working on this in 2-3 days maximum.

Issue #8424, opened 2026-07-01
feat: federated AI agent search across registry instances


carlesarnal:
## Summary

Add a federation layer that lets a registry instance query peer instances for A2A agent cards, enabling cross-registry agent discovery.

## Scope

- Peer registration API (URL + optional credentials, stored by reference when configured)
- Federated search endpoint that fans out queries to peers and merges results
- "Remote Registries" UI page for managing peers and browsing agents
- Graceful handling of unavailable peers (timeouts, partial results)
- Integration tests with Testcontainers (multiple registry instances)

## Context

LFX Mentorship project for 2026 Term 3 (Sep-Nov), assigned to @aclfe. Builds on existing agent discovery support. The approved first release uses GET /apis/registry/v3/search/federated and the peer agent-search surface with explicit public-only support. ARD federation and third-party interoperability are deferred.

## Plan and tracking

Mentor guidance and acceptance expectations: https://github.com/Apicurio/apicurio-registry/issues/8424#issuecomment-5660526148

Design discussion: https://github.com/Apicurio/apicurio-registry/discussions/10012

- [ ] #8573 — Peer configuration persistence (first implementation milestone)
- [ ] #8574 — Peer management REST API
- [ ] #8575 — Federated agent search
- [ ] #8576 — Remote Registries UI
- [ ] #8577 — SDK updates, alongside the relevant API changes
- [ ] #8578 — Operator integration
- [ ] #8579 — Multi-instance integration coverage, starting with the first search slice
- [ ] #8580 — User and architecture documentation

Critical path: #8573 → #8574 → #8575 → initial end-to-end coverage in #8579. Tests accompany each implementation step; keep one implementation PR active at a time. UI and operator work follow agreement on their API contracts. Operator reconciliation versus admin API edits must be settled before controller implementation.

## Current handoff — design reviewed, implementation unblocked

Design decisions: https://github.com/Apicurio/apicurio-registry/discussions/10012#discussioncomment-18537660

Project board: https://github.com/orgs/Apicurio/projects/26 (parent and all eight subtasks).

@aclfe: the peer model is approved. Begin the first #8573 storage PR; another design proposal is not required. SQL/migration/decorators first, followed by KafkaSQL and GitOps/KubernetesOps declarative support. Keep #8573 open until all variants are covered and keep one implementation PR active at a time.

[...]

PR #9538, opened 2026-08-13
fix(search): match structure filters whose name segment contains a colon

### LINKED ISSUE #9537: A2A agent search returns no results for skill IDs containing a colon
## Description

**Registry Version**: snapshot (build from `main`, currently `3.3.2-SNAPSHOT`)
**Persistence type**: sql (also affects kafkasql). Not persistence-specific. The bug lives in the
Elasticsearch search backend and reproduces on any persistence type with the search index enabled (`apicurio.search.index.enabled=true`), which is required for A2A structured (skill/capability) search.

A2A agent discovery lets clients search registered Agent Cards by skill ID via `GET /.well-known/agents?skill=<id>` (and the equivalent `skills` filter on `POST /.well-known/agents/search`). When the skill ID contains a colon (`:`), the search returns no results even though a matching Agent Card is registered and correctly indexed. Colon-free skill IDs work as expected, so the agent is silently undiscoverable by exactly the skills whose IDs are namespaced/URN-style.

Root cause is a format-detection mismatch between how structured elements are indexed and how the structure filter is turned into an Elasticsearch query:

- Indexing writes the exact-match key as `"{artifactType}:{kind}:{name}"`, lowercased, in `ElasticsearchDocumentBuilder.indexStructuredElements` (`app/.../storage/impl/search/ElasticsearchDocumentBuilder.java`, around lines 196-204). For a skill `acme:translate` this becomes `agent_card:skill:acme:translate`.
- The A2A layer builds the filter as `A2AConstants.PREFIX_AGENT_CARD_SKILL + skill`, i.e. `agent_card:skill:acme:translate` (`app/.../rest/wellknown/WellKnownResourceImpl.java`, `searchAgents`).
- The query builder `buildStructureQuery` decides the format by counting colons (`app/.../storage/impl/search/ElasticsearchSearchService.java`, lines 306-320): `value.split(":", -1)` and `parts.length == 3` means "exact match". A skill ID with a colon produces 4+ parts, so it falls through to a fuzzy `structure_text` text query with `operator=AND`, which cannot match (the `structure_text` value is `"skill acme:translate"` and does not contain the token `agent_card`). 
- Result: zero hits.

The value format is `type:kind:name` where `name` may itself contain colons, so detecting the format by counting colons is unsound. The same code path is used by the `capability`, `inputMode`, `outputMode`, and `protocolBinding` filters; in practice capabilities and MIME-type modes do not contain colons, so skill IDs are the primary trigger.

**Relevance to #8424 (Federated AI Agent Search):** skill-based discovery is the core query surface for federated agent search, and real-world skill IDs are commonly namespaced (reverse-DNS or URN style, e.g. `com.acme:translate`, `urn:agent:skill:translate`). This bug makes a whole class of agents undiscoverable by their skills, which directly undermines the federated search capability tracked in #8424.

### Environment

- Apicurio Registry built from `main`, run with:
  - `apicurio.a2a.enabled=true`
  - `apicurio.search.index.enabled=true` (Elasticsearch search index; required for structured A2A search)
- Interaction is via the REST API only (the `/.well-known/agents` endpoint). No serdes client involved.

### Steps to Reproduce

1. Start the Registry with `apicurio.a2a.enabled=true` and `apicurio.search.index.enabled=true` (Elasticsearch reachable).
2. Register an Agent Card artifact (artifact type `AGENT_CARD`) that declares two skills, one with a plain ID and one with a colon in the ID:

```json
   {
     "protocolVersion": "1.0",
     "name": "Translator Agent",
     "description": "Translates and summarizes text",
     "skills": [
       { "id": "summarize", "name": "Summarize", "tags": ["nlp"] },
       { "id": "acme:translate", "name": "Translate", "tags": ["nlp"] }
     ],
     "capabilities": { "streaming": true }
   }
```

3. Search by the colon-free skill (control):
`GET /.well-known/agents?skill=summarize`
The agent is returned.

4. Search by the colon-bearing skill (bug):
`GET /.well-known/agents?skill=acme:translate`
The agent is NOT returned, although its Agent Card declares that skill and it was indexed.

- Expected: searching by a skill ID that contains a colon returns every Agent Card that declares that skill, identically to colon-free skill IDs.
- Actual: the search returns an empty result set. The filter `agent_card:skill:acme:translate` is misclassified as a partial/plain query (because it has more than 3 colon-separated parts) and runs a `structure_text` text match that cannot match the indexed document, so the agent is silently undiscoverable by that skill.

### Logs

With `LOG_LEVEL=DEBUG`, the Elasticsearch search reports zero hits for the colon case while the same
document exists and matches on the colon-free skill. Representative debug output from
`ElasticsearchSearchService`:

```txt
DEBUG [i.a.r.s.i.s.ElasticsearchSearchService] Elasticsearch search returned 1 results (total: 1, offset: 0, limit: 20)   # skill=summarize
DEBUG [i.a.r.s.i.s.ElasticsearchSearchService] Elasticsearch search returned 0 results (total: 0, offset: 0, limit: 20)   # skill=acme:translate
```
Inspecting the indexed document confirms the structured element is present as `structure ["agent_card:skill:summarize", "agent_card:skill:acme:translate", ...]`, so the miss is in query construction, not indexing.

Me-Priyank:
Fixes #9537

## What

Structured search values are indexed as `type:kind:name`, where `type` and `kind` are colon-free identifiers but the `name` segment may itself contain a colon (for example a namespaced A2A Agent Card skill id such as `acme:translate`, or an indexed
URL). `ElasticsearchSearchService.buildStructureQuery` detected the fully-qualified form by an exact count of three colon-separated segments, so a name containing a colon produced four or more segments and fell through to a `structure_text` text query that can never match the indexed value.

As a result, A2A agent search silently returned no results for any skill, capability, input/output mode or protocol binding whose value contained a colon, even though the Agent Card was registered and correctly indexed.

## Fix

Treat any value with three or more segments as the fully-qualified `type:kind:name` form and match it exactly against the `structure` field. The exact match uses the whole value, so a name containing additional colons is matched correctly and no false positives are possible (an exact term match can only equal an identical indexed value). The two-segment (`kind:name`) and one-segment (`name`) text-search paths are unchanged.

## Testing

Adds `ElasticsearchStructureQueryTest` covering the fully-qualified, colon-in-name (regression), `kind:name` and plain-name cases. Verified locally: `Tests run: 4, Failures: 0, Errors: 0`.

Relates to #8424 (Federated AI Agent Search): skill-based discovery is the core query surface for federated agent search, and real-world skill ids are commonly namespaced.


Me-Priyank:
@carlesarnal @EricWittmann @paoloantinori PTAL

paoloantinori:
Took a look. The fix is correct and the test coverage is solid.

- Changing the exact-match guard from `parts.length == 3` to `>= 3` is the right call. Since `type` and `kind` are colon-free, the first two segments are unambiguous, and any segments past the third belong to the `name`. The exact `term` match on the full lowercased value means a colon in the name cannot produce a false positive — it can only equal an identical indexed value.
- The split-with-limit `-1` is handled correctly: a trailing colon (`type:kind:`) yields a three-segment array with an empty name and resolves to an exact match that will simply not match anything indexed, rather than silently degrading to a text query.
- `ElasticsearchStructureQueryTest` covers the four cases that matter: fully-qualified, colon-in-name (the regression), `kind:name`, and plain name. The assertions pin both the query type (`term` vs `match`) and the target field, which is what locks the behavior in.

One thing worth confirming separately from this PR: the fix assumes the indexed `structure` value is lowercased with `Locale.ROOT` so it matches the `lowered` value used here. The tests pass because they construct the expected term from the same lowercased input, but they do not exercise the indexing side. If `ElasticsearchDocumentBuilder` normalizes differently, the exact match would miss. Worth a quick check against the indexing path, but not a blocker for this change.

No concerns from me on the code. The version-bump/merge logistics are out of scope here.

Me-Priyank:
Thanks @paoloantinori for the review, I confirmed the normalization matches: `ElasticsearchDocumentBuilder` builds the `structure` value by lowercasing artifactType, kind and name with `Locale.ROOT` and joining them as `type:kind:name`, which is the same `Locale.ROOT` lowercasing this query uses, so the exact term always lines up with the indexed value.

I also pushed a test that pins this in code (`indexedStructureValueMatchesQueryTerm`): it drives `ElasticsearchDocumentBuilder` for an Agent Card whose skill id mixes case and contains a colon, then asserts the stored `structure` value and the query term are the identical string. 

waiting for @EricWittmann or @carlesarnal to review now :))

paoloantinori:
Thanks for following up and adding `indexedStructureValueMatchesQueryTerm`. Driving the real indexing path with a mixed-case, colon-bearing skill id is exactly the right way to close the loop, since the query-only tests build both sides from the same lowercased input and could not catch an indexer/query normalization mismatch. The setup looks right: both sides use `Locale.ROOT`, the stored `structure` value is asserted before the query term is, and the assertion that the term value equals the indexed value (not the raw input) is the correct one. Nothing else from me on this one.

paoloantinori (commented):
The fix is correct for the stated bug: values with a colon in the name segment now resolve to an exact `structure` term query instead of silently falling into a `structure_text` match that can never hit, and the indexer-consistency test (`indexedStructureValueMatchesQueryTerm`) is exactly the right way to guard the query/index normalization contract. Two points worth addressing before merge: the new test file has no Apache license header (rat/checkstyle will flag it), and the `>= 3` heuristic introduces an ambiguity that is acknowledged nowhere in the code.
- `ElasticsearchSearchService.java:319`: `parts.length >= 3` makes the fully-qualified form and a `kind:name` filter whose name contains a colon syntactically indistinguishable. `skill:acme:translate` (user intent: kind + namespaced name, previously also ambiguous at exactly 3 segments) now always becomes an exact term query, so a text search over `structure_text` for a 4+-segment value is no longer reachable. That is the intended trade-off, but a user who wants text search on a colon-bearing name has no escape syntax. The javadoc should state this explicitly ("a value with three or more segments is always treated as fully-qualified; there is no text-search form for such values").
- `ElasticsearchSearchService.java:290-305`: the javadoc expansion is roughly three times the size of the code change and repeats the PR description verbatim, including the "silently dropped any value..." narrative. The three-bullet format contract plus one sentence on the `>= 3` rule would carry the same information; the failure-history prose belongs in the commit/PR, not the API doc.
- `ElasticsearchStructureQueryTest.java:1`: the new file starts directly at `package` with no Apache 2.0 license header, which every other file in this module carries and the rat check enforces.
- `ElasticsearchStructureQueryTest.java:96-101`: the indexer-consistency test builds `ElasticSearchConfig` via direct field mutation (`config.contentMaxSize = ...`, `documentBuilder.config = config`). If those are public fields that is fine, but confirm this matches how other tests in the module construct the builder; a factory or existing test fixture would keep this from bypassing whatever initialization `buildVersionDocument` normally relies on.
- Unverifiable from the diff alone: confirm no other search implementation (SQL/KafkaSQL variants of structure filtering) has the same exact-segment-count assumption. The fix as written only covers the Elasticsearch path, and issue #9537 does not say which storage the reporter used.

app/src/main/java/io/apicurio/registry/storage/impl/search/ElasticsearchSearchService.java
```diff
@@ -306,8 +316,10 @@ private Query buildStructureQuery(String value) {
         String lowered = value.toLowerCase(Locale.ROOT).trim();
         String[] parts = lowered.split(":", -1);
 
-        if (parts.length == 3) {
-            // Full format: type:kind:name - exact match on structure field
+        if (parts.length >= 3) {
```
paoloantinori, line 371: With `parts.length >= 3`, a `kind:name` filter whose name contains a colon (e.g. `skill:acme:translate`, user intent kind + namespaced name) is now indistinguishable from the fully-qualified form and always becomes an exact term query. A 4+-segment value can no longer reach `structure_text` text search at all. That is the right trade-off for this fix, but please state it in the javadoc: "any value with three or more segments is treated as fully-qualified and matched exactly; there is no text-search form for such values". Also worth confirming the other structure-query implementations (SQL/KafkaSQL search, if they exist) do not share the exact-segment-count assumption, otherwise the bug is only fixed on the Elasticsearch path.

app/src/test/java/io/apicurio/registry/storage/impl/search/ElasticsearchStructureQueryTest.java
```diff
@@ -0,0 +1,119 @@
+package io.apicurio.registry.storage.impl.search;
```
paoloantinori, line 1: This file is missing the Apache 2.0 license header that the rest of the module carries; the rat/license check will fail on it. Please add the standard header above the `package` declaration.

carlesarnal (approved):

Me-Priyank:
@carlesarnal @EricWittmann ping...

PR #9540, opened 2026-08-13
fix(ui): reset pagination to page 1 on Drafts and Explore page filter…

### LINKED ISSUE #9534: Bug: Reset pagination to page 1 when applying filters on Explore and Drafts pages
## Description

When a user is viewing a later page of a paginated list on the **Explore** or **Drafts** pages and then applies a filter, the current pagination page can be retained instead of being reset to the first page.

This can cause the filtered result set to appear empty even when matching items exist.

For example, if a user is on page 3 and applies a group filter that only has a few matching items, the UI can continue requesting page 3 of the filtered results. Since the filtered result set may only contain one page, the UI displays an empty/no-results state instead of showing the matching items on page 1.

The filter handling in `SearchPage` already resets pagination to page 1 when filter criteria change, but the corresponding inline/group filter paths in `ExplorePage` and `DraftsPage` do not consistently perform the same reset.

### Environment

Apicuro Registry frontend running from the current `main`/`snapshot` build.

### Steps to Reproduce

1. Open the **Explore** or **Drafts** page.
2. Navigate to page 2 or page 3 of the paginated results.
3. Apply a filter that produces a smaller result set, for example by clicking a group filter/label.
4. Observe the resulting list.
5. If the filtered result set does not contain enough items to have the previously selected page, the UI displays an empty/no-results state.
6. Clear the filter or navigate back to page 1 and observe that matching results are available.

**Expected:**

When filter criteria change, pagination should reset to page 1 so that the user sees the beginning of the newly filtered result set.

**Actual:**

The current pagination page can be retained when filters are applied through certain filter handlers. The frontend therefore requests the same page number from the newly filtered result set, which can result in an empty response even though matching items exist on page 1.

For example:

`Page 3 → Apply filter → Page 3 of filtered results → No results`

instead of:

`Page 3 → Apply filter → Page 1 of filtered results → Matching results`

### Logs

No Registry error is required to reproduce this behavior. This is a frontend pagination/state-management issue.

The relevant implementation can be observed in:

* `ExplorePage.tsx`
* `DraftsPage.tsx`

The affected filter paths update the filter criteria without consistently resetting the pagination state to page 1.

hussainjamal760:
## What
Resets the active pagination page index back to 1 whenever search filters/criteria change or list data is mutated (via deletion or import) in both DraftsPage and ExplorePage.

## Why
Previously, when users navigated to a higher page (e.g., Page 2 or Page 3) on the Drafts or Explore pages and subsequently:

Applied a new search filter or clicked a Group ID tag filter.
Deleted an item or performed a bulk ZIP import.
The pagination state (paging.page) remained locked at the previous higher page index. As a result, subsequent search requests were sent to the backend with an out-of-bounds offset (e.g. offset = 20).

When the resulting filtered dataset or post-mutation list had fewer total items than the current page offset, the UI incorrectly rendered a "No results found" empty state, even though valid matching items existed on Page 1.

## Changes Made

DraftsPage.tsx
:
Updated filterByGroupId to set paging.page: 1 when filtering by Group ID.
Updated onCriteriaChange in DraftsPageToolbar to reset paging.page to 1 when filter chips/search terms change.

ExplorePage.tsx
:
Updated doDeleteGroup to reset paging to Page 1 upon successful group deletion before fetching updated results.
Updated doImport to reset paging to Page 1 upon completion of ZIP import before fetching updated results.
Verification / How to Test
Scenario 1: Filtering on Drafts Page
Open the Drafts page with enough drafts to span multiple pages.
Navigate to Page 2.
Apply a filter (e.g., type a filter term or click a Group ID chip).
Expected Outcome: The pagination resets to Page 1 and displays matching filtered drafts instead of a false "No results found" state.
Scenario 2: Group Deletion / Import on Explore Page
Open the Explore page and navigate to Page 2.
Delete a group or complete a ZIP file import.
Expected Outcome: The list re-fetches starting from Page 1, avoiding out-of-bounds offset queries.

Issue Closes : #9534

### DIFF
--- a/ui/ui-app/src/app/pages/drafts/DraftsPage.tsx
+++ b/ui/ui-app/src/app/pages/drafts/DraftsPage.tsx
@@ -217,6 +217,7 @@ export const DraftsPage: FunctionComponent<PageProperties> = () => {
         if (!updated) {
             newCriteria.push(dsf);
         }
+        setPaging(prev => ({ ...prev, page: 1 }));
         setCriteria(newCriteria);
     };
 
@@ -247,7 +248,10 @@ export const DraftsPage: FunctionComponent<PageProperties> = () => {
                 setSortBy(by);
                 setSortOrder(dir);
             }}
-            onCriteriaChange={setCriteria}
+            onCriteriaChange={newCriteria => {
+                setPaging(prev => ({ ...prev, page: 1 }));
+                setCriteria(newCriteria);
+            }}
             onRefresh={() => {
                 search(criteria, sortBy, sortOrder, paging);
             }}
--- a/ui/ui-app/src/app/pages/explore/ExplorePage.tsx
+++ b/ui/ui-app/src/app/pages/explore/ExplorePage.tsx
@@ -225,7 +225,9 @@ export const ExplorePage: FunctionComponent<ExplorePageProps> = () => {
                     setImporting(false);
                     setImportProgress(100);
                     setImportModalOpen(false);
-                    search(criteria, paging);
+                    const resetPaging: Paging = { page: 1, pageSize: paging.pageSize };
+                    setPaging(resetPaging);
+                    search(criteria, resetPaging);
                 }, 1500);
             }).catch(error => {
                 setPageError(toPageError(error, "Error importing multiple artifacts"));
@@ -253,7 +255,9 @@ export const ExplorePage: FunctionComponent<ExplorePageProps> = () => {
         pleaseWait(true, "Deleting group, please wait.");
         groups.deleteGroup(groupToDelete?.groupId as string).then( () => {
             pleaseWait(false);
-            search(criteria, paging);
+            const resetPaging: Paging = { page: 1, pageSize: paging.pageSize };
+            setPaging(resetPaging);
+            search(criteria, resetPaging);
         }).catch(error => {
             setPageError(toPageError(error, "Error deleting group."));
         });


ui/ui-app/src/app/pages/drafts/DraftsPage.tsx
```diff
@@ -247,7 +248,10 @@ export const DraftsPage: FunctionComponent<PageProperties> = () => {
                 setSortBy(by);
                 setSortOrder(dir);
             }}
```
paoloantinori, line 250: This behavioral fix has a clear reproducible scenario but no test. Add a test that starts on page > 1, changes the search criteria (and separately the Group ID chip filter), and asserts the outgoing search request uses page 1 / offset 0 instead of the previous page. Without coverage, a future change to the search effect's dependency array could silently reintroduce the stale-offset regression.

ui/ui-app/src/app/pages/explore/ExplorePage.tsx
```diff
@@ -225,7 +225,9 @@ export const ExplorePage: FunctionComponent<ExplorePageProps> = () => {
                     setImporting(false);
                     setImportProgress(100);
                     setImportModalOpen(false);
-                    search(criteria, paging);
+                    const resetPaging: Paging = { page: 1, pageSize: paging.pageSize };
```
paoloantinori, line 228: This three-line reset is duplicated verbatim in doDeleteGroup (line 258), and it diverges from DraftsPage's approach. The explicit object { page: 1, pageSize: paging.pageSize } copies only pageSize and silently drops any other fields if Paging grows later, whereas DraftsPage uses the safer setPaging(prev => ({ ...prev, page: 1 })). Extract a shared helper, for example resetPagingAndSearch(criteria), or at least use the functional-updater form in both Explore sites for consistency and to avoid the field-drop risk.

paoloantinori (commented):
The fix correctly targets the stale-offset root cause: resetting paging.page to 1 on filter change and on post-mutation re-fetch prevents the false 'No results found' empty state when the filtered or mutated dataset is smaller than the current page offset. Two items to address before merge: the behavioral change has no test, and the ExplorePage reset is implemented inconsistently with DraftsPage and duplicated.

- DraftsPage.tsx:220 and DraftsPage.tsx:250. Resetting page on the Group ID filter and the toolbar criteria change is correct. Because setPaging and setCriteria are batched into a single commit, the search effect fires with page 1 (offset 0), assuming that effect reads paging from the committed render rather than a stale closure. Worth confirming the effect's dependency array includes paging.
- ExplorePage.tsx:228 and ExplorePage.tsx:258. doImport and doDeleteGroup pass the reset paging object explicitly into search(), so the fetch uses page 1 immediately rather than waiting for state to settle. Correct.
- No automated test covers this fix despite a clear reproducible scenario (navigate to page 2 or 3, apply a filter, expect offset 0). A regression in the search effect dependencies could silently reintroduce the bug.
- ExplorePage.tsx:228 and :258 duplicate the same three lines, and the explicit { page: 1, pageSize: paging.pageSize } object copies only pageSize, silently dropping any other Paging fields. DraftsPage uses the safer functional updater. Prefer the functional updater or a shared helper in both Explore sites.
- Minor consistency note: sort and order changes (setSortBy/setSortOrder) leave paging.page untouched. That does not cause the reported bug (sorting does not reduce the count), but it is the same stale-page surface, so consider whether a reset belongs there too.

EricWittmann:
:+1:

PR #9544, opened 2026-08-13
docs: document operator autoscaling and console plugin, fix stale KubernetesOps RBAC steps

eameh:


eameh:
Hi @EricWittmann, when you get a chance, could you review this one? It fixes the stale KubernetesOps RBAC instructions (the operator has been creating those automatically since 3.3.0) and adds docs for the autoscaling and console plugin features. I'd love to get it in ahead of the 3.3 downstream release. Thanks!

EricWittmann (commented):
Thanks @eameh — nice work on this one, especially the RBAC fix (the stale manual steps really needed updating). All claims verified against the operator source. A few optional suggestions for extra precision — none of these are blockers:

1. **Validation error wording (optional):** The docs say the operator "reports a validation error condition and ignores the `replicas` field" when both are set. This is accurate — the code sets a `ValidationError` status condition. Just noting that it doesn't block reconciliation, which the current wording doesn't imply anyway. No change needed.

2. **PDB + maxReplicas > 1 (optional):** The docs say the operator manages a PDB "when autoscaling is enabled with `maxReplicas` greater than 1." Looking at `ActivationConditions.java`, the `maxReplicas > 1` check only exists for the **app** component (line 83-86). The UI PDB activation (lines 223-230) doesn't have this check. You could optionally note this distinction, though in practice it's an edge case.

3. **Console plugin on non-OpenShift (optional):** The docs say "On clusters that are not OpenShift, the console plugin is not deployed." In the code, only the `ConsolePlugin` CR creation is gated on OpenShift detection (`ConsolePluginManager.java:68`). The backing Deployment and Service are created on any cluster if the feature is enabled — they'd just be unused without the CR. You could optionally say something like: *"the ConsolePlugin custom resource is only created on OpenShift; on other clusters the plugin has no effect."*

All of these are **optional refinements** — the docs are accurate enough to merge as-is. Happy to approve either way!

EricWittmann:
Merged — thanks @eameh! If you think any of the optional suggestions (validation error wording, PDB distinction, console plugin on non-OpenShift) are worth adding, feel free to open follow-up PRs. 👍

PR #9547, opened 2026-08-13
docs: document KafkaSQL startup verification and operator leader election

eameh:


eameh:
Hi @EricWittmann, could you review this one when you have time? It documents the KafkaSQL startup verification checks and the operator leader election settings. Hoping to include it in the 3.3 downstream release docs. Thanks!

EricWittmann (commented):
Thanks @eameh — this looks good! All the technical claims check out against the source code. One optional suggestion:

**Override property precision (optional):** The docs describe `apicurio.kafkasql.topic-configuration-verification-override-enabled` as relaxing "retention verification," but looking at `KafkaAdminUtil.java`, the override only skips the `retention.ms` check — the `retention.bytes=-1` check runs regardless of the override setting. You could optionally add a clarification like:

> *"the override relaxes the `retention.ms=-1` requirement; the `retention.bytes=-1` requirement is always enforced."*

This is **not a blocker** — the current wording is consistent with the `@Info` annotation on the property itself, and users enabling this for snapshotting primarily care about `retention.ms`. Just flagging it in case you want to be extra precise.

EricWittmann:
Merged — thanks @eameh! If you think the optional suggestion about clarifying the retention override behavior is worth adding, feel free to open a follow-up PR. 👍

PR #9548, opened 2026-08-13
docs: document runtime log level changes and REST metrics distribution

eameh:


eameh:
Hi @EricWittmann, one more small one when you get a chance: runtime log level changes via the dynamic property and the new REST metrics distribution options from 3.3.1. Thanks so much!

EricWittmann:
Thanks @eameh! Merged.

PR #9551, opened 2026-08-13
fix(sql): Add null guards and logging to SqlEventRepository.createEvent

### LINKED ISSUE #9550: NullPointerException in SqlEventRepository.createEvent on null event or null payload
### Description

`SqlEventRepository.createEvent(OutboxEvent event)` does not guard
against a null `event` argument or a null `event.getPayload()`,
resulting in an unhandled `NullPointerException` in either case.

Additionally, the injected `log` field in this class is never used,
so several other conditions currently fail silently:

- On database types that don't support database-driven outbox
  events (e.g. H2, MySQL), `createEvent` returns `event.getId()` as
  if the event were persisted, with no log line indicating the
  event was actually dropped.
- Successful outbox writes have no trace-level record for
  debugging.

### Steps to Reproduce

1. Call `createEvent(null)`: throws `NullPointerException`.
2. Call `createEvent(event)` where `event.getPayload()` is null, on
   a Postgres or MSSQL-backed registry → throws
   `NullPointerException` at:
```java
   .bind(4, event.getPayload().toString())
```

### Expected Behavior

- A null `event` should be handled gracefully (e.g. logged and
  return `null`), not throw.
- A null payload should not cause an NPE. Confirmed: every database
  schema declares the `payload` column `NOT NULL`, so binding raw
  SQL `NULL` merely trades the Java NPE for a SQL constraint
  violation. The fix is to coerce a null payload to an empty JSON
  object string (`"{}"`) before binding, satisfying the constraint
  and preserving the outbox row rather than dropping the event.
- The unsupported-database-type case and successful writes should
  be logged (at `debug`/`trace` level respectively) using the
  already-injected `log` field, rather than failing/succeeding
  silently.

### Fix

```java
.bind(4, (event.getPayload() != null ? event.getPayload().toString() : "{}"))
```

### Additional Notes

Coercing to "{}" assumes a null payload is a valid state for OutboxEvent.

neel-1414:
Add null checks against a null event and a null event.getPayload() to prevent NullPointerException, utilizes the previously unused log field, debug level logging when a db type doesn't support outbox events (expected for H2/MySQL), and trace-level logging on successful writes. Flips dbType().equals(...) to constant first to guard against a null dbType().

Closes: #9550

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlEventRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlEventRepository.java
@@ -31,17 +31,18 @@ public class SqlEventRepository {
      * Create an outbox event for database-driven event publishing.
      */
     public String createEvent(OutboxEvent event) {
-        if(event == null)
-        {
-            log.warn("Cannot create outbox event: event is null");
-            return null;
+        if (event == null) {
+            throw new IllegalArgumentException("OutboxEvent must not be null");
+        }
+        if (event.getPayload() == null) {
+            throw new IllegalArgumentException("OutboxEvent payload must not be null");
         }
         if (supportsDatabaseEvents()) {
             handles.withHandle(handle -> {
                 handle.createUpdate(sqlStatements.createOutboxEvent()).bind(0, event.getId())
                         .bind(1, eventsTopic).bind(2, event.getAggregateId()).bind(3, event.getType())
-                        .bind(4, (event.getPayload() != null ? event.getPayload().toString() : null)).execute();
-
+                        .bind(4, event.getPayload().toString()).execute();
+                
                 return handle.createUpdate(sqlStatements.deleteOutboxEvent()).bind(0, event.getId())
                         .execute();
             });
--- a/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlEventRepository.java
+++ b/app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlEventRepository.java
@@ -19,7 +19,8 @@ public class SqlEventRepository {
 
     private final String eventsTopic;
 
-    public SqlEventRepository(HandleFactory handles, SqlStatements sqlStatements, Logger log, String eventsTopic) {
+    public SqlEventRepository(HandleFactory handles, SqlStatements sqlStatements, Logger log,
+            String eventsTopic) {
         this.handles = handles;
         this.sqlStatements = sqlStatements;
         this.log = log;
@@ -30,20 +31,26 @@ public class SqlEventRepository {
      * Create an outbox event for database-driven event publishing.
      */
     public String createEvent(OutboxEvent event) {
+        if(event == null)
+        {
+            log.warn("Cannot create outbox event: event is null");
+            return null;
+        }
         if (supportsDatabaseEvents()) {
             handles.withHandle(handle -> {
-                handle.createUpdate(sqlStatements.createOutboxEvent())
-                        .bind(0, event.getId())
-                        .bind(1, eventsTopic)
-                        .bind(2, event.getAggregateId())
-                        .bind(3, event.getType())
-                        .bind(4, event.getPayload().toString())
-                        .execute();
+                handle.createUpdate(sqlStatements.createOutboxEvent()).bind(0, event.getId())
+                        .bind(1, eventsTopic).bind(2, event.getAggregateId()).bind(3, event.getType())
+                        .bind(4, (event.getPayload() != null ? event.getPayload().toString() : null)).execute();
 
-                return handle.createUpdate(sqlStatements.deleteOutboxEvent())
-                        .bind(0, event.getId())
+                return handle.createUpdate(sqlStatements.deleteOutboxEvent()).bind(0, event.getId())
                         .execute();
             });
+            log.trace("Created outbox event {} of type {} for aggregate {}", event.getId(), event.getType(),
+                    event.getAggregateId());
+        } else {
+            log.debug(
+                    "Database-driven events are not supported for db type '{}'; event {} of type {} was not persisted to the outbox",
+                    sqlStatements.dbType(), event.getId(), event.getType());
         }
         return event.getId();
     }
@@ -56,10 +63,10 @@ public class SqlEventRepository {
     }
 
     private boolean isPostgresql() {
-        return sqlStatements.dbType().equals("postgresql");
+        return "postgresql".equals(sqlStatements.dbType());
     }
 
     private boolean isMssql() {
-        return sqlStatements.dbType().equals("mssql");
+        return "mssql".equals(sqlStatements.dbType());
     }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlEventRepository.java
```diff
@@ -30,20 +31,26 @@ public SqlEventRepository(HandleFactory handles, SqlStatements sqlStatements, Lo
      * Create an outbox event for database-driven event publishing.
      */
     public String createEvent(OutboxEvent event) {
```
paoloantinori, line 33: The null-event and null-payload branches are new behavior with no unit tests in this PR. Add tests covering createEvent(null) (assert it returns null or throws per the decision below, and that the warning is logged) and createEvent with a null payload (assert it binds null without throwing). Every new branch needs coverage per the project review standard.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlEventRepository.java
```diff
@@ -30,20 +31,26 @@ public SqlEventRepository(HandleFactory handles, SqlStatements sqlStatements, Lo
...
+        if(event == null)
```
paoloantinori: Style: the rest of the method uses `if (...) {` with the brace on the same line and a space after `if`. This `if(event == null)` with no space and the brace on the next line is inconsistent and likely fails checkstyle. Use `if (event == null) {`.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlEventRepository.java
```diff
@@ -30,20 +31,26 @@ public SqlEventRepository(HandleFactory handles, SqlStatements sqlStatements, Lo
...
+        {
+            log.warn("Cannot create outbox event: event is null");
+            return null;
```
paoloantinori: Returning null here turns a loud NPE (stack trace at the call site) into a silent warning plus a null return, which pushes the failure to whoever consumes the returned id. A null OutboxEvent is a caller bug, not a recoverable condition, so prefer throwing IllegalArgumentException to fail fast. If you keep the null return, confirm every caller of createEvent null-checks the result.

app/src/main/java/io/apicurio/registry/storage/impl/sql/repositories/SqlEventRepository.java
```diff
@@ -30,20 +31,26 @@ public SqlEventRepository(HandleFactory handles, SqlStatements sqlStatements, Lo
...
-                handle.createUpdate(sqlStatements.createOutboxEvent())
-                        .bind(0, event.getId())
-                        .bind(1, eventsTopic)
-                        .bind(2, event.getAggregateId())
-                        .bind(3, event.getType())
-                        .bind(4, event.getPayload().toString())
-                        .execute();
+                handle.createUpdate(sqlStatements.createOutboxEvent()).bind(0, event.getId())
+                        .bind(1, eventsTopic).bind(2, event.getAggregateId()).bind(3, event.getType())
+                        .bind(4, (event.getPayload() != null ? event.getPayload().toString() : null)).execute();
```
paoloantinori: Guarding the .toString() NPE is fine, but confirm the intended behavior when payload is null. If the outbox payload column is NOT NULL, this bind throws a constraint violation at write time; if it is nullable, you persist a row with no payload that downstream publishing will try to deserialize. If a null payload is never a valid event state, fail fast here instead of binding null.

paoloantinori (commented):
The hardening is mostly sound, and the constant-first dbType() comparisons plus the debug/trace logging are genuine improvements, but the null-event path converts a loud NPE into a silent null return that needs caller-side handling, the null-payload branch needs confirmation against the outbox schema, and the new branches lack tests.

- Returning null on a null event (lines 34 to 37) masks a caller bug; prefer IllegalArgumentException and confirm every caller handles a null return.
- Binding null when payload is null (line 43) needs the outbox schema confirmed: a NOT NULL column throws at write time, a nullable column persists a row with no payload for downstream deserialization.
- The new null-event and null-payload branches (method at line 33) have no unit tests; each new branch needs coverage.
- The null-check style `if(event == null)` with the brace on the next line (lines 34 to 35) breaks the file convention `if (...) {` and may fail checkstyle.
- Good: comparing the literal against dbType() with the constant first (lines 64, 68) is null-safe, and the debug log on the unsupported-db path documents expected H2/MySQL behavior.

JituRewar (commented):
@neel-1414 Could we add a regression test for the `event == null` case and clarify whether returning `null` here is the expected behavior? Since `createEvent()` normally returns an event ID, I want to make sure callers are expected to handle a `null` return value rather than this case being rejected with an exception.

JituRewar (commented):
@neel-1414  Could we add a test covering an event with a null payload? This change now binds null to the outbox payload column, so it would be good to verify that the database schema allows this and that persisting an event with a null payload is the intended behavior.

neel-1414:
Hi @paoloantinori,

Thanks for the feedback! I checked the DDL files across all database engines (postgresql.ddl, mssql.ddl, mysql.ddl), and the payload column is defined as NOT NULL in all of them.

Binding null directly will fail at write time due to the NOT NULL constraint, while binding "{}" as a fallback could cause downstream consumers to fail during deserialization when reading required JSON fields.

Should I enforce a non-null payload check in createEvent and throw an IllegalArgumentException early if it is missing?

neel-1414:
Hi @JituRewar! Sure thing, I'll add tests for both scenarios in the next commit

paoloantinori:
Good catch checking the DDLs. You're right that binding null straight to a NOT NULL column shifts the failure from a clear validation error into a less obvious SQL constraint violation at write time, which is exactly the kind of loud-vs-silent tradeoff the original NPE hardening was trying to avoid.

Throwing IllegalArgumentException early in createEvent when event.getPayload() is null is the cleaner answer and it keeps the behavior consistent with how I'd like the null-event case to fail too (explicitly, at the call site, rather than returning null or letting the DB reject it). I'd suggest guarding payload before the supportsDatabaseEvents() branch so both the outbox and the no-op paths get the same explicit failure, e.g. a single check up front with a message like "OutboxEvent payload must not be null".

The constant-first dbType() comparisons and the debug/trace logging both look good to me. With the payload null-check in place and the regression tests @JituRewar asked for, I don't have further concerns.

neel-1414:
Hi @paoloantinori, thanks for the review! I've pushed a fix with the null-check and regression tests as requested. Let me know if anything else needs adjusting.

paoloantinori:
Thanks — re-read the latest head. Both guards are in place up front (before the `supportsDatabaseEvents()` branch) with the agreed messages, and `testCreateNullEvent` / `testCreateEventWithNullPayload` pin them down. The null-guard on the observer side in `SqlEventsProcessor` and the constant-first `dbType()` comparisons also look right, and binding `event.getPayload().toString()` is now safe behind the payload check.

No further concerns from my side. Mergeable state shows the branch behind main, so it will need a rebase before merge, but content-wise this is good to go.

neel-1414:
Done rebased on latest main and force-pushed. Should be mergeable now.

paoloantinori:
Thanks — re-checked after the rebase. The guards and both regression tests are unchanged at the new head, so this still looks good from my side.

carlesarnal (approved):

neel-1414:
Hi @carlesarnal  @EricWittmann, just a gentle ping on this, Is there anything else needed on my end or are we good to merge?

PR #9563, opened 2026-08-13
fix(search): route structure filter artifact searches through the search index

### LINKED ISSUE #9562: Agent search returns 500 when filtering by skill, capability, or input/output mode
### Registry
Version: 3.3.2-SNAPSHOT (current main)
Persistence type: all — SQL-backed variants (H2, KafkaSQL) share the SqlSearchRepository path, and Elasticsearch on or off makes no difference.

### Environment
Any deployment with apicurio.a2a.enabled=true and apicurio.features.experimental.enabled=true (A2A is still experimental). No special infrastructure needed.

### Steps to Reproduce
1. Start the registry with A2A support enabled.
2. Register an AGENT_CARD artifact declaring a skill, e.g. "skills": [{"id": "schema-validation", "name": "Schema Validation"}].
3. GET /.well-known/agents?skill=schema-validation

Same result with the capability, inputMode, or outputMode params, and with POST /.well-known/agents/search using filters.skills, filters.capabilities, filters.inputModes, filters.outputModes, or filters.protocolBindings.

Expected: 200 with the matching agents. Skills are indexed as structured content (agent_card:skill:) to serve these filters.
Actual: 500 on every request carrying one of these filters.

The endpoints turn the params into SearchFilterType.structure filters and run them through RegistryStorage.searchArtifacts(). ElasticsearchSearchDecorator only intercepts searchVersions() — the decorator proxy passes undeclared methods to the delegate — so the call lands in SqlSearchRepository.searchArtifacts(), whose filter switch has no structure case and throws. Unfiltered and name-only searches work, and nothing in the test suite exercises a skill or capability query, which is why this went unnoticed.

### Logs
```yaml
io.apicurio.registry.storage.error.RegistryStorageException: Filter type not supported: structure
        at io.apicurio.registry.storage.impl.sql.repositories.SqlSearchRepository.searchArtifacts
```

kartikangiras:
## Description

A2A agent searches with skill/capability/inputMode/outputMode filters returned HTTP 500 on every configuration: the filters become structure search filters, but ElasticsearchSearchDecorator only intercepted searchVersions(), so artifact searches always fell through to SQL, which rejects structure filters.

### Changes
- ElasticsearchSearchDecorator now intercepts searchArtifacts(): the index resolves which artifacts match the structure/content filters, SQL evaluates the remaining filters and supplies authoritative metadata/ordering, and the results are intersected.
- ElasticsearchSearchService: new searchArtifactIdentities() (dedupes version docs to artifacts, 10k scan cap) and group-normalizing identityKey().
- SqlSearchRepository: structure filters now fail with a clear 400 (ContentSearchNotSupportedException) instead of a 500 when the search index is disabled.
- **Outcome**: GET /.well-known/agents?skill(and capability/mode filters, plus POST /.well-known/agents/search) return correct results with the search index enabled, and a clear 400 without it. Previously: 500 always.

### Issue
fixes #9562

### DIFF
--- a/app/src/main/java/io/apicurio/registry/storage/decorator/ElasticsearchSearchDecorator.java
+++ b/app/src/main/java/io/apicurio/registry/storage/decorator/ElasticsearchSearchDecorator.java
@@ -1,5 +1,6 @@
 package io.apicurio.registry.storage.decorator;
 
+import co.elastic.clients.elasticsearch._types.ElasticsearchException;
 import io.apicurio.registry.storage.dto.ArtifactSearchResultsDto;
 import io.apicurio.registry.storage.dto.OrderBy;
 import io.apicurio.registry.storage.dto.OrderDirection;
@@ -14,6 +15,8 @@ import io.apicurio.registry.storage.impl.search.ElasticsearchSearchService;
 import io.apicurio.registry.storage.impl.search.ElasticsearchStartupIndexer;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -29,6 +32,8 @@ import java.util.Set;
 @ApplicationScoped
 public class ElasticsearchSearchDecorator extends RegistryStorageDecoratorBase
         implements RegistryStorageDecorator {
+
+    private static final Logger log = LoggerFactory.getLogger(ElasticsearchSearchDecorator.class);
 
     private static final String INDEX_NOT_AVAILABLE_MESSAGE =
             "Content search requires the Elasticsearch search index, which is not "
@@ -68,7 +73,7 @@ public class ElasticsearchSearchDecorator extends RegistryStorageDecoratorBase
             try {
                 return searchService.searchVersions(filters, orderBy, orderDirection,
                         offset, limit, skipCount);
-            } catch (IOException e) {
+            } catch (IOException | ElasticsearchException e) {
                 throw new RegistryStorageException(
                         "Elasticsearch search failed for index-only filters.", e);
             }
@@ -90,9 +95,15 @@ public class ElasticsearchSearchDecorator extends RegistryStorageDecoratorBase
      * name matching, case-normalized labels) on the SQL side, where the behavior matches
      * non-index searches.</p>
      *
-     * <p>Both the index scan and the SQL scan are capped at
-     * {@link ElasticsearchSearchService#MAX_ARTIFACT_SEARCH_HITS} entries, so results beyond
-     * that many candidates are not included.</p>
+     * <p><b>Hard cap (part of the search contract):</b> both the index scan and the SQL scan
+     * retrieve at most {@link ElasticsearchSearchService#MAX_ARTIFACT_SEARCH_HITS} candidates,
+     * and the returned result count is computed from their intersection. When more artifacts
+     * than the cap match, results beyond the cap are omitted and the count understates the true
+     * total — a paginating client will stop early. A warning is logged when either scan hits
+     * the cap. Callers should scope index-backed searches (e.g. with an artifact-type filter,
+     * as the well-known A2A endpoints always do) so the candidate sets stay below the cap; a
+     * request carrying only index-only filters leaves the SQL scan unscoped and it degrades to
+     * a full-catalog scan up to the cap.</p>
      */
     public ArtifactSearchResultsDto searchArtifacts(Set<SearchFilter> filters, OrderBy orderBy,
             OrderDirection orderDirection, int offset, int limit, boolean skipCount)
@@ -121,7 +132,7 @@ public class ElasticsearchSearchDecorator extends RegistryStorageDecoratorBase
         Set<String> matchedIdentities;
         try {
             matchedIdentities = searchService.searchArtifactIdentities(esFilters);
-        } catch (IOException e) {
+        } catch (IOException | ElasticsearchException e) {
             throw new RegistryStorageException(
                     "Elasticsearch search failed for index-only filters.", e);
         }
@@ -133,6 +144,11 @@ public class ElasticsearchSearchDecorator extends RegistryStorageDecoratorBase
 
         ArtifactSearchResultsDto sqlResults = delegate.searchArtifacts(sqlFilters, orderBy,
                 orderDirection, 0, ElasticsearchSearchService.MAX_ARTIFACT_SEARCH_HITS, true);
+        if (sqlResults.getArtifacts().size() >= ElasticsearchSearchService.MAX_ARTIFACT_SEARCH_HITS) {
+            log.warn("Artifact search SQL scan reached the cap of {} artifacts; matches beyond "
+                    + "the cap are not included and the reported count may understate the true "
+                    + "total.", ElasticsearchSearchService.MAX_ARTIFACT_SEARCH_HITS);
+        }
 
         List<SearchedArtifactDto> matched = new ArrayList<>();
         for (SearchedArtifactDto artifact : sqlResults.getArtifacts()) {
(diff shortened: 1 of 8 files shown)


app/src/main/java/io/apicurio/registry/storage/decorator/ElasticsearchSearchDecorator.java
```diff
@@ -67,4 +75,78 @@ public VersionSearchResultsDto searchVersions(Set<SearchFilter> filters, OrderBy
...
+        for (SearchedArtifactDto artifact : sqlResults.getArtifacts()) {
+            if (matchedIdentities.contains(ElasticsearchSearchService.identityKey(
+                    artifact.getGroupId(), artifact.getArtifactId()))) {
+                matched.add(artifact);
+            }
+        }
+
+        int total = matched.size();
+        int fromIndex = Math.min(Math.max(0, offset), total);
+        int toIndex = Math.min(fromIndex + Math.max(0, limit), total);
```
paoloantinori, line 163: `total` (and thus `count`) is computed from `matched`, which is already bounded by the 10k caps on both the index scan (`searchArtifactIdentities`) and the SQL scan (`MAX_ARTIFACT_SEARCH_HITS` at L132-L133). For result sets larger than 10k, the response silently omits valid matches and reports an understated count, so a paginating client stops early and misses the tail. The cap is logged and commented but never signaled to the caller. Consider surfacing a truncated flag, or documenting the hard cap as part of the search contract. For the A2A use case this is unlikely to bite, but searchArtifacts is a general path.

app/src/test/java/io/apicurio/registry/search/WellKnownAgentSearchViaIndexTest.java
```diff
@@ -0,0 +1,248 @@
...
+                """.formatted(name, skillId, name, streaming, skillId, skillId);
+    }
+
+    private void createAgentCard(String groupId, String artifactId, String content) throws Exception {
+        CreateArtifact createArtifact = new CreateArtifact();
+        createArtifact.setArtifactId(artifactId);
+        createArtifact.setArtifactType(ArtifactType.AGENT_CARD);
+
+        CreateVersion createVersion = new CreateVersion();
+        VersionContent versionContent = new VersionContent();
```
paoloantinori, line 85: Every case here creates the agent under a generated non-default group, so the identityKey null-to-"default" normalization that joins index docs to SQL rows is never exercised end to end. If agent cards live in the default group in production, an index-side groupId mismatch would silently return zero results. Add a case that creates an agent in the default group (omit the group, or pass "default") and asserts it is returned by a skill search.

app/src/main/java/io/apicurio/registry/storage/decorator/ElasticsearchSearchDecorator.java
```diff
@@ -67,4 +75,78 @@ public VersionSearchResultsDto searchVersions(Set<SearchFilter> filters, OrderBy
...
+            matchedIdentities = searchService.searchArtifactIdentities(esFilters);
+        } catch (IOException e) {
+            throw new RegistryStorageException(
+                    "Elasticsearch search failed for index-only filters.", e);
+        }
+
+        ArtifactSearchResultsDto results = new ArtifactSearchResultsDto();
+        if (matchedIdentities.isEmpty()) {
+            return results;
+        }
```
paoloantinori, line 143: When a request carries only index-only filters and no artifactType, `sqlFilters` is empty and this call scans up to 10k artifacts just to intersect. Confirm the well-known agents resource always adds artifactType=AGENT_CARD so the SQL side stays scoped; otherwise every skill/capability query degrades to a full-catalog scan.

paoloantinori (commented):
Overall this is a correct, well-documented fix for the structure-filter 500: the index resolves which artifacts have a matching version, SQL supplies authoritative metadata and ordering, and the in-memory intersection preserves SQL ordering with correct offset/limit clamping. The filter split is complete (every original filter lands in at least one backend, artifactType is correctly double-evaluated as exact match in both) and the decorator unit tests cover the main branches well. A few things worth addressing before merge:

- Silent truncation at 10k (ElasticsearchSearchDecorator.java L143/L147, ElasticsearchSearchService.java L190/L203). Both the index scan and the SQL scan are hard-capped at MAX_ARTIFACT_SEARCH_HITS, and count is derived from the already-truncated intersection. Beyond 10k matching artifacts the response omits valid matches and understates count, so a paginating client stops early and never sees the tail. The cap is warned and commented but never signaled to the caller. Worth surfacing a truncated indicator or documenting the hard cap as part of the search contract. Unlikely to bite the A2A use case, but searchArtifacts is a general path.

- Default-group join is only unit-tested (WellKnownAgentSearchViaIndexTest.java). identityKey normalizes the default group (null to "default") so index docs and SQL rows compare equal, but every integration case uses a generated non-default group, so that join is never exercised end to end. If agent cards are stored in the default group in production, an index-side groupId mismatch would silently return zero results.

- Potential full-catalog SQL scan (ElasticsearchSearchDecorator.java L106-L117, L132). If a request carries only structure/content filters and no artifactType, sqlFilters ends up empty and the delegate scans up to 10k artifacts just to intersect. Confirm the well-known agents resource always adds artifactType=AGENT_CARD so the SQL side stays scoped.

- Minor: searchArtifacts catches only IOException from the index call (L120-L125). If searchVersions additionally wraps runtime ES exceptions, mirror that here so transport failures surface consistently as RegistryStorageException rather than uncaught 500s.

kartikangiras:
Hi @paoloantinori ,
i have iterated over review and addressed all the feedbacks.
Thanks.

app/src/main/java/io/apicurio/registry/storage/decorator/ElasticsearchSearchDecorator.java
kartikangiras: every A2A entry point in [WellKnownResourceImpl unconditionally adds artifactType=AGENT_CARD before searching, so the SQL side of the intersection always stays scoped. The decorator javadoc now also warns that a request with only index only filters degrades to a capped full catalog scan.

paoloantinori:
Thanks for the re-read and the iteration. Loop is closed from my side:

- The `IOException | ElasticsearchException` catch on both the version and artifact paths addresses the transport-failure consistency point.
- Documenting the `MAX_ARTIFACT_SEARCH_HITS` cap and the full-catalog-scan degradation as part of the search contract in the `searchArtifacts` Javadoc is the right call — it makes the limit explicit to anyone extending the path beyond the A2A endpoints.
- Good to have the confirmation that `WellKnownResourceImpl` unconditionally adds `artifactType=AGENT_CARD`, so the SQL side of the intersection stays scoped in practice.

No further concerns from me.

kartikangiras:
Hi @EricWittmann 
Could you take a look into this PR for the next approve and merge phase. It was approved by Paolo a few weeks back. If you feel further changes are required, I am happy to iterate over.
Thanks!

paoloantinori (commented):
I have bad news on this one, and you should have had it weeks ago.

Most of this PR has been overtaken. #9658 merged as `61c25b495` and implements the same search path a different way: `ElasticsearchSearchDecorator.searchArtifacts` now collapses on `ga_key` with a cardinality aggregation, instead of intersecting index results with SQL under a 10,000-row cap. The two designs cannot both live in the tree, and the merged one does not carry the cap. I checked this against main rather than going by the PR titles.

One piece of your work did survive, and it is still worth having. `SqlSearchRepository` has no `case structure:` in its filter switch on current main, so a structure filter falls through to the `default:` that throws `RegistryStorageException`, and a client gets a 500. Your nine-line fix for that is unaffected by #9658 and fixes a live bug. The reason nobody has hit it yet is that `SearchFilterType.structure` now has exactly one producer, `SearchResourceImpl.java:298`, so the well-known path stopped emitting those filters. It is still reachable.

What I would like: open a small PR with just the `case structure:` handling, and close this one. I am sorry to ask that after 674 lines. The 25 days of silence on this PR is the actual failure here and it is ours, not yours. Had someone answered in week one you would not have carried the work past the point where it was overtaken.

carlesarnal:
Coordination note: PR #8538 now has maintainer follow-up 908d607b4d against current main, implementing SQL artifact structure filters and import/lifecycle indexing (93 tests passed across H2/PG/MySQL/KafkaSQL plus read-only imports). The old A2A POST/public/entitled endpoints have been removed from main. Please preserve any needed ES-specific changes here, but reconcile the SQL unsupported-structure fallback and removed-endpoint tests with #8538 rather than landing contradictory behavior. No automatic closure/supersession assumed.

PR #9564, opened 2026-08-13
fix(mcp): prevent PromptTemplateMCPServer crash when encountering disabled artifacts (#9556)

### LINKED ISSUE #9556: [Bug] PromptTemplateMCPServer crashes when encountering a DISABLED prompt template
### Description

The `PromptTemplateMCPServer` contains a critical flaw in its error handling during prompt template discovery. If an administrator marks a single `PROMPT_TEMPLATE` artifact as `DISABLED`, the entire `list_mcp_prompts` tool fails for all AI agents, completely breaking prompt discovery.

When `list_mcp_prompts()` is invoked, it retrieves all `PROMPT_TEMPLATE` versions using `service.searchVersions(...)`. Crucially, this search does not filter by `VersionState.ENABLED`, meaning it returns `DISABLED` templates as well. 

The server then iterates over these versions and attempts to fetch their content:
```java
// PromptTemplateMCPServer.java
for (SearchedVersion version : versions) {
    try {
        String content = service.getVersionContent(
                version.getGroupId(),
                version.getArtifactId(),
                version.getVersion()
        );
        // ... parse template ...
    } catch (IOException e) {
        // Skip artifacts that can't be read
    }
}
```

The bug lies in the `catch (IOException e)` block. When `service.getVersionContent(...)` requests a `DISABLED` artifact, the Apicurio Registry REST API returns an HTTP 404 (`VersionNotFoundException` is thrown in `GroupsResourceImpl`). 

The Kiota-based REST client handles this HTTP 404 by throwing an `ApiException` (which extends `RuntimeException`), **not** an `IOException`. Because `ApiException` is unchecked, it bypasses the `catch (IOException e)` block, bubbles up to `handleError(...)`, and crashes the entire tool invocation with an HTTP 500 equivalent. 

As a result, a single disabled prompt template in the registry acts as a "poison pill," taking down the entire MCP prompt discovery mechanism.

### Steps to Reproduce

1. Register a valid, MCP-enabled `PROMPT_TEMPLATE` artifact (Template A).
2. Register a second MCP-enabled `PROMPT_TEMPLATE` artifact (Template B).
3. Using the registry UI or API, change the state of Template B to `DISABLED`.
4. Connect an AI agent using the MCP protocol and invoke the `list_mcp_prompts` tool.
5. **Observe:** The agent receives a complete failure (ToolCallException) instead of receiving the still-valid Template A.

**Expected Behaviour:** 
The MCP Server should either proactively filter out `DISABLED` templates in the initial search query, or successfully catch the `ApiException` thrown when fetching disabled content, gracefully skipping it and returning the remaining active templates.

**Actual Behaviour:** 
The unchecked `ApiException` bypasses the `IOException` catch block, crashing the entire iteration loop and failing the tool call.

### Suggested Implementation

1. **Short-term fix:** Broaden the catch block in `PromptTemplateMCPServer.list_mcp_prompts()` (and similar discovery loops) to catch `RuntimeException` or `Exception` so that a failure in one artifact does not abort the entire loop:
```java
} catch (Exception e) {
    // Skip artifacts that can't be read or are disabled/deprecated
    log.debug("Skipping prompt template {}/{}: {}", version.getGroupId(), version.getArtifactId(), e.getMessage());
}
```
2. **Long-term fix:** Expose the `state` query parameter in `RegistryService.searchVersions` so that `PromptTemplateMCPServer` can proactively request only `ENABLED` templates, avoiding the unnecessary HTTP 404 errors entirely.

SurbhiAgarwal1:
### Description

This PR fixes a bug in `PromptTemplateMCPServer` where discovery of prompt templates (`list_mcp_prompts()`) crashes if any artifact version cannot be fetched or is marked as `DISABLED`.

#### Cause
When `service.getVersionContent(...)` requests content for a `DISABLED` artifact version, the Kiota-generated REST client throws `com.microsoft.kiota.ApiException` (which extends `RuntimeException`). Because the previous `try-catch` block only caught `IOException`, unchecked `ApiException`s bypassed the catch block and crashed the entire `list_mcp_prompts()` tool call.

#### Solution
- Broadened the catch block in `PromptTemplateMCPServer.java` to `catch (Exception e)`.
- Failsafe handling: Any single unreadable, disabled, or failing artifact version is logged and skipped, allowing the server to return all remaining active MCP-enabled prompt templates successfully.

Fixes #9556

### DIFF
--- a/mcp/src/main/java/io/apicurio/registry/mcp/servers/PromptTemplateMCPServer.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/servers/PromptTemplateMCPServer.java
@@ -75,8 +75,8 @@ public class PromptTemplateMCPServer {
                             prompts.add(prompt);
                         }
                     }
-                } catch (IOException e) {
-                    // Skip artifacts that can't be read
+                } catch (Exception e) {
+                    // Skip artifacts that can't be read or are disabled/deprecated
                 }
             }
 
--- a/mcp/src/main/java/io/apicurio/registry/mcp/servers/PromptTemplateMCPServer.java
+++ b/mcp/src/main/java/io/apicurio/registry/mcp/servers/PromptTemplateMCPServer.java
@@ -1,5 +1,6 @@
 package io.apicurio.registry.mcp.servers;
 
+import com.microsoft.kiota.ApiException;
 import io.apicurio.registry.mcp.PromptTemplateConverter;
 import io.apicurio.registry.mcp.PromptTemplateConverter.MCPPrompt;
 import io.apicurio.registry.mcp.PromptTemplateConverter.PromptTemplate;
@@ -13,6 +14,8 @@ import io.quarkiverse.mcp.server.TextContent;
 import io.quarkiverse.mcp.server.Tool;
 import io.quarkiverse.mcp.server.ToolArg;
 import jakarta.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -31,6 +34,8 @@ import static io.quarkiverse.mcp.server.PromptMessage.withUserRole;
  * This enables AI agents to discover and use prompt templates stored in Apicurio Registry.
  */
 public class PromptTemplateMCPServer {
+
+    private static final Logger log = LoggerFactory.getLogger(PromptTemplateMCPServer.class);
 
     private static final String PROMPT_TEMPLATE_TYPE = "PROMPT_TEMPLATE";
 
@@ -75,8 +80,9 @@ public class PromptTemplateMCPServer {
                             prompts.add(prompt);
                         }
                     }
-                } catch (Exception e) {
-                    // Skip artifacts that can't be read or are disabled/deprecated
+                } catch (ApiException | IOException e) {
+                    log.warn("Failed to retrieve or parse prompt template artifact {}/{}:{}: {}",
+                            version.getGroupId(), version.getArtifactId(), version.getVersion(), e.getMessage());
                 }
             }
 


paoloantinori (commented):
The fix correctly prevents a `DISABLED` artifact version from crashing the entire `list_mcp_prompts()` call, but broadening the catch to `Exception` is too wide and silently masks failures that previously reached the client via `handleError`. The PR's stated rationale is also inconsistent with what the code actually does.

- `mcp/.../PromptTemplateMCPServer.java:78` - `catch (Exception e)` swallows programming bugs (`NullPointerException`, `ClassCastException`) and HTTP errors from `getVersionContent` (403/404/500) that previously propagated to `handleError()` as a `ToolCallException`. These are now invisible. Prefer narrowing to the actual failure types (the kiota `ApiException` plus `IOException`) so genuine code defects still surface.
- `mcp/.../PromptTemplateMCPServer.java:78` - If every artifact errors (e.g., the registry is down or the service account lacks permissions), the tool returns an empty list indistinguishable from "no prompts exist." An AI agent consuming this concludes the registry is empty rather than failing. Consider re-throwing on systemic connectivity/auth failures, or at least surfacing partial-failure to the caller.
- `mcp/.../PromptTemplateMCPServer.java:79` - The comment "or are disabled/deprecated" is factually wrong: `searchVersions` is called with `VersionState.ENABLED` upstream, so disabled/deprecated versions never reach this loop body. The catch cannot encounter them. Update the comment to describe what the catch actually handles, otherwise it misleads future maintainers about the control flow.
- `mcp/.../PromptTemplateMCPServer.java:78` - The swallow has no logging, so debugging a missing prompt yields zero server-side signal. This is also inconsistent with the sibling `render_registry_prompt`, which surfaces `"Error: " + e.getMessage()`. Add at least a WARN/DEBUG log line with the artifact GAV and exception class, matching the sibling method's observability.
- `mcp/.../PromptTemplateMCPServer.java:17` - `import java.io.IOException` is now unused in this file (its only reference was the broadened catch clause). Remove it to avoid an unused-import lint.

SurbhiAgarwal1:
Thanks for the thorough review @paoloantinori! I've updated the PR to address all your points:

1. **Narrowed Catch Block**: Replaced `catch (Exception e)` with `catch (ApiException | IOException e)` so programming errors like `NullPointerException` are not swallowed and can still surface.
2. **Added Logging**: Injected SLF4J logger and added `log.warn(...)` outputting the artifact GAV coordinates and exception message when a version fetch fails.
3. **Corrected Comment**: Updated comment to accurately reflect what the catch block handles.
4. **Rebased**: Rebased onto latest `main`.

paoloantinori:
Thanks @SurbhiAgarwal1, re-read at 59affde and all three points landed: the catch is now ApiException | IOException so programming errors like NPEs still surface, the warn log carries the group/artifact/version coordinates plus the exception message, and the stale comment is gone. Skipping a single unreadable or disabled version with a log line instead of failing the whole list_mcp_prompts() call is the right shape for a discovery loop. No further concerns from me.

SurbhiAgarwal1:
Thank you so much for the review and feedback @paoloantinori!

paoloantinori:
You're welcome! Glad the review was useful. This one is good to go from my side.

jsenko (commented):
Fixes #9556 — prevents \`list_mcp_prompts()\` from crashing when a disabled artifact version is encountered.

Clean, minimal fix. The catch block correctly narrows from the original \`IOException\`-only to \`ApiException | IOException\`, which covers HTTP errors from the Kiota REST client (e.g., 404 for disabled artifacts) without over-catching. The added \`log.warn\` with artifact coordinates is a good improvement over the previous silent skip.

No further issues found from my side.

- **Race condition resilience**: The search already passes \`VersionState.ENABLED\` (line 65), so disabled artifacts shouldn't normally appear in results. However, the \`ApiException\` catch is still valuable as a resilience measure — an artifact can be disabled between the search and content fetch, or content retrieval can fail for other reasons (network, permission). The fix correctly treats this as a skip-and-continue scenario.
- **No similar patterns elsewhere**: Checked other MCP server files — none have the same \`IOException\`-only catch pattern around content fetches.

jsenko (approved):

jsenko:
Thank you!

PR #9567, opened 2026-08-13
fix(serdes): unwrap ApiException to expose original Registry errors

### LINKED ISSUE #5236: Expose or log error response from Registry in serdes
Since the serdes communicates with Registry via cache, when Registry returns an error response, it is shadowed by an unhelpful cache error, e.g.:

```
Failed to update cache value for key: {"type":"record","name":"Key","namespace":"inventory_connector_postgresql1727173403473.inventory.customers","fields":[{"name":"id","type":{"type":"int","connect.default":0},"default":0}],"connect.name":"inventory_connector_postgresql1727173403473.inventory.customers.Key"}   [io.apicurio.registry.resolver.ERCache]
io.apicurio.registry.rest.client.exception.RestClientException
	at io.apicurio.registry.rest.client.impl.ErrorHandler.parseError(ErrorHandler.java:95)
	at io.apicurio.rest.client.JdkHttpClient.sendRequest(JdkHttpClient.java:207)
	at io.apicurio.registry.rest.client.impl.RegistryClientImpl.createArtifact(RegistryClientImpl.java:311)
	at io.apicurio.registry.rest.client.RegistryClient.createArtifact(RegistryClient.java:147)
	at io.apicurio.registry.resolver.DefaultSchemaResolver.lambda$handleAutoCreateArtifact$3(DefaultSchemaResolver.java:278)
	at io.apicurio.registry.resolver.ERCache.lambda$getValue$0(ERCache.java:201)
	at io.apicurio.registry.resolver.ERCache.retry(ERCache.java:254)
	at io.apicurio.registry.resolver.ERCache.getValue(ERCache.java:200)
	at io.apicurio.registry.resolver.ERCache.getByContent(ERCache.java:175)
	at io.apicurio.registry.resolver.DefaultSchemaResolver.handleAutoCreateArtifact(DefaultSchemaResolver.java:276)
	at io.apicurio.registry.resolver.DefaultSchemaResolver.getSchemaFromRegistry(DefaultSchemaResolver.java:123)
	at io.apicurio.registry.resolver.DefaultSchemaResolver.resolveSchema(DefaultSchemaResolver.java:91)
	at io.apicurio.registry.serde.AbstractKafkaSerializer.serialize(AbstractKafkaSerializer.java:83)
	at io.apicurio.registry.utils.converter.SerdeBasedConverter.fromConnectData(SerdeBasedConverter.java:122)
```

Improve error handling to expose the underlying error response from Registry.

niy-ati:
Fixes #5236

`ERCache.retry` was wrapping Registry client errors in a generic `RuntimeException`, which buried the HTTP status and body from SerDes callers.

Changes:
- Explicitly unwrap `ApiException` from the cause chain in `ERCache.retry` so the original client exception propagates up.
- Log concise error details (HTTP status + message) via a package-private `describeRegistryError` helper.
- Restrict retries to HTTP 429 rate limit errors only; fail fast on all other errors (404, 409, 500, etc.).
- Update `ERCacheTest` with regression coverage for cause-chain unwrapping and fix a clock-resolution flakiness issue in `testGetByContentHashEnforcesTTL`.


paoloantinori (commented):
The unwrapping is the right fix for #5236: surfacing the original ApiException instead of a generic RuntimeException is exactly what SerDes callers need, and restricting retries to HTTP 429 preserves the prior retry semantics (the old code only retried on 429 as well). One MAJOR issue should be addressed before merge: the new retry predicate can throw an NPE that hides the very Registry error this PR exists to expose.

- ERCache.java:466. MAJOR. `apiException.getResponseStatusCode() == 429` auto-unboxes a nullable Integer. kiota's ApiException stores the status as a nullable `Integer`, and any ApiException in the chain created without an HTTP context (serialization or network errors, or one sitting above the real HTTP failure) returns null here. The NPE is thrown from inside the catch block, so it escapes `retry()` unhandled and replaces the Registry error with an opaque NullPointerException. The full-chain walk in `findApiException` makes hitting a null-status ApiException more likely than the old single-level `getCause()` check did.
- ERCache.java (catch block). Minor: `findApiException` is walked three times on the failure path (`isRetryableRateLimit`, `describeRegistryError`, and the explicit call). Computing it once and reusing the reference would also make the null-safe fix above cleaner.
- ERCacheTest.java:94. Using `Thread.sleep(1)` to fix TTL flakiness can reintroduce it. On coarse-resolution clocks a 1ms sleep may not advance the TTL clock at all, and under load it can overshoot by tens of milliseconds. Prefer an injected controllable Clock for deterministic TTL tests, or confirm the configured TTL is well below the minimum wall-clock advancement a 1ms sleep guarantees on the CI runners.
- ERCacheTest.java:92,98. The added blank lines carry trailing whitespace, which checkstyle/spotless will flag.
- Behavior change worth a quick grep: callers that previously unwrapped the old wrapper via `result.error().getCause()` now receive the bare ApiException (or the original RuntimeException) with no wrapper. That is the intended improvement, but it is a type change at the cache result surface, so confirming no call site depends on the wrapper is worth it.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +460,48 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
         return Result.error(new IllegalStateException("Unreachable."));
     }
 
+    /**
+     * Returns true when the failure (or its cause chain) is an HTTP 429 from the Registry client.
+     */
+    private static boolean isRetryableRateLimit(RuntimeException e) {
```
paoloantinori: `apiException.getResponseStatusCode() == 429` auto-unboxes a nullable Integer. kiota's ApiException stores the status as a nullable `Integer`, and any ApiException in the chain that was created without an HTTP context (serialization or network errors, or one sitting above the real HTTP failure) returns null here. The resulting NPE is thrown from inside the catch block, so it escapes `retry()` unhandled and replaces the Registry error with an opaque NullPointerException, which is exactly the masking this PR is trying to remove. The full-chain walk in `findApiException` also makes encountering a null-status ApiException more likely than the old single-level `getCause()` check. Use a null-safe comparison, e.g. `Integer status = apiException.getResponseStatusCode(); return status != null && status == 429;`.

schema-resolver/src/test/java/io/apicurio/registry/resolver/ERCacheTest.java
```diff
@@ -88,6 +89,13 @@ void testGetByContentHashEnforcesTTL() {
 
         String firstValue = cache.getByContentHash(contentHashKey, firstLoader);
         assertEquals("a value", firstValue);
+        
+        try {
+            Thread.sleep(1);
```
paoloantinori: Using `Thread.sleep(1)` to fix TTL flakiness can reintroduce it. On coarse-resolution clocks a 1ms sleep may not advance the TTL clock at all, and under load `Thread.sleep` can overshoot by tens of milliseconds. If the cache reads wall-clock time, prefer injecting a controllable Clock so the test advances time deterministically. Otherwise, please confirm the configured TTL is comfortably below the minimum wall-clock advancement a 1ms sleep guarantees on the CI runners.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +460,48 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
+    /**
+     * Builds a concise description of the Registry/client failure for logs, preferring HTTP status
+     * and message from {@link ApiException} when present.
+     */
+    static String describeRegistryError(Throwable throwable) {
+        ApiException apiException = findApiException(throwable);
+        if (apiException != null) {
+            String message = apiException.getMessage();
+            if (message == null || message.isBlank()) {
+                message = apiException.getClass().getSimpleName();
```
paoloantinori, line 494: This log line embeds a non-ASCII em-dash (U+2014). The rest of the module logs ASCII, and a non-ASCII separator can render oddly in some log aggregators or terminals. Use a colon or hyphen, e.g. `"HTTP " + apiException.getResponseStatusCode() + ": " + message`.

schema-resolver/src/test/java/io/apicurio/registry/resolver/ERCacheTest.java
niy-ati: The TTL configured for this specific test is `Duration.ZERO`, so any recorded forward advancement of the clock triggers the expiration. I've bumped the sleep to 50ms to comfortably guarantee that the wall clock moves forward enough to be recorded, even on CI runners with coarse-resolution clocks.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
niy-ati: I've updated the catch block to fetch the `ApiException` exactly once and reuse the reference for the checks and logging to avoid the redundant chain walking. I also added a null-safe check for the status code to ensure it handles exceptions without an HTTP context gracefully.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
niy-ati: Ah, my bad. I've swapped out the em-dash for a standard ASCII hyphen. I also removed the trailing whitespaces on the blank lines in `ERCacheTest.java` to prevent any Checkstyle issues. All fixes have been pushed.

niy-ati:
Hi @EricWittmann, the current CI failure is unrelated to this PR , it's the `avro-maven-plugin:1.12.2` schema compilation error in the `app` module (`AvroSchemaB.avsc` uses named-type references in `map.values`/`array.items` that the newer plugin version rejects). The last few Verify runs on `main` are all failing with the same error.

Our changes only touch `ERCache.java`, `ERCacheTest.java`, and `SchemaResolverTest.java `, no `.avsc `files or plugin config were modified.

niy-ati:
@EricWittmann , PTL

niy-ati:
@EricWittmann , PTAL

paoloantinori (commented):
The cache change itself reads correctly and I am close to approving. Two things first, both in the tests.

**`SchemaResolverTest.java:95` cannot fail.** The new line is

```java
var error = runtimeException instanceof io.apicurio.registry.rest.client.v2.models.Error || runtimeException instanceof io.apicurio.registry.rest.client.models.ProblemDetails ? runtimeException : runtimeException.getCause();
```

The fallback arm is the old wrapped behavior, so the assertion that follows holds whether or not your change is in place. Revert the production change locally and this test still goes green. Pick the one form the code is supposed to produce now and assert that, so the test breaks if someone undoes the fix. While you are there, those two fully qualified names should be imports.

**The NPE fix has no test.** The null guard is the substantive part of this PR and nothing exercises the path that used to throw.

One more, and this one is free: the two `.avsc` files in the diff are byte-identical to what is already on main. They landed via #9664. I fetched your branch and diffed them to be sure. They will disappear on rebase, so you do not need to do anything about them, but they are inflating the diff in the meantime.

PR #9568, opened 2026-08-13
fix(flink): shade utils/flink jar to bundle transitive dependencies

### LINKED ISSUE #8285: Flink catalog module should produce a fat/uber JAR for easier deployment
## Description

The `apicurio-registry-utils-flink` module currently produces a thin JAR without its transitive
dependencies bundled. This makes it difficult for users to deploy the Flink catalog integration,
because they must manually collect and provide all transitive dependencies on Flink's classpath.

## Problem

A user reported in Zulip that after building `utils/flink` and placing the JAR in Flink's
`/opt/flink/usrlib/`, they get:

```
java.lang.ClassNotFoundException: io.apicurio.registry.client.common.RegistryClientOptions
```

This happens because `RegistryClientOptions` lives in `apicurio-registry-java-sdk-common`, which is
a compile-scope dependency of the Flink module but is not included in the output JAR. The same issue
applies to all other transitive dependencies (the Java SDK, serde modules, Kiota HTTP client, Avro,
Jackson, etc.).

## Expected Behavior

The Flink module should produce a single self-contained uber/fat JAR (using `maven-shade-plugin`)
that bundles all required runtime dependencies. Flink's own libraries (`flink-table-api-java`,
`flink-table-common`) are already marked as `provided` scope and would be excluded.

This is the standard pattern for Flink connectors and catalog plugins — users expect to drop a
single JAR into their Flink lib directory.

## Current Workaround

Users can collect all transitive dependencies manually:

```bash
mvn dependency:copy-dependencies -pl utils/flink -DincludeScope=runtime \
    -DoutputDirectory=target/flink-libs
```

Then copy everything from `utils/flink/target/flink-libs/` plus the main
`apicurio-registry-utils-flink-*.jar` into `/opt/flink/usrlib/`.

## Proposed Solution

Add `maven-shade-plugin` to `utils/flink/pom.xml` to produce a shaded JAR that includes all
compile/runtime dependencies while excluding provided-scope Flink libraries. The shaded JAR
should also relocate common libraries (e.g. Jackson, SLF4J) to avoid classpath conflicts with
Flink's own copies.

The README should also be updated with deployment instructions.

nafisa404:
Deploying the Flink catalog jar to a real cluster throws ClassNotFoundException: io.apicurio.registry.client.common.RegistryClientOptions on startup, because utils/flink was only ever building a thin jar — java-sdk, java-sdk-common, the Avro/JSON Schema serde modules, Avro, and Jackson were all on the compile classpath but never made it into the deployed artifact.

This adds maven-shade-plugin to produce a single self-contained jar. Jackson gets relocated to io.apicurio.registry.flink.shaded.com.fasterxml.jackson since Flink bundles its own Jackson and the two would otherwise clash at runtime; the ServicesResourceTransformer merges META-INF/services entries so Jackson's and Avro's service files don't clobber each other during the merge. The one thing I deliberately did not relocate is slf4j-api — I moved it to provided scope instead. Flink's runtime classpath always has slf4j-api on it, and relocating it would break SLF4J's binding mechanism, since a shaded copy can't find whatever logging implementation is bound on Flink's classpath. flink-table-api-java and flink-table-common were already provided and are excluded the same way.

Verified locally: mvn clean package -pl utils/flink -am and mvn test -pl utils/flink both pass (30/30 unit tests, including AvroFlinkTypeConverterTest). Inspected the built jar directly — RegistryClientOptions.class is present, org/apache/flink/table and org/slf4j are both absent (provided-scope exclusion working), org/apache/avro/Schema.class is present unrelocated as intended, and there are zero unrelocated com/fasterxml/jackson classes anywhere in the jar, including the META-INF/versions multi-release-JAR paths. That last one needed an explicit exclude filter on jackson-core's internal FastDoubleParser fork, since maven-shade-plugin's relocator doesn't rewrite MRJAR entry paths (MSHADE-406). I did not have a live registry instance available to run ApicurioCatalogIT end-to-end, so that's still worth a run before merge.

Closes #8285

### DIFF
--- a/utils/flink/README.md
+++ b/utils/flink/README.md
@@ -95,6 +95,22 @@ catalog.close();
 mvn clean install -pl utils/flink -am
 ```
 
+The build produces a single self-contained (shaded) JAR. Flink's own libraries
+(`flink-table-api-java`, `flink-table-common`) and `slf4j-api` are intentionally
+excluded from the shaded JAR since Flink provides them on its runtime classpath.
+
+## Deployment
+
+Copy the shaded JAR into Flink's `lib` directory:
+
+```bash
+cp utils/flink/target/apicurio-registry-utils-flink-<version>.jar $FLINK_HOME/lib/
+```
+
+Manually collecting dependencies via `mvn dependency:copy-dependencies` is no
+longer necessary — the JAR already bundles everything it needs except what
+Flink provides.
+
 ## Testing
 
 Unit tests:
--- a/utils/flink/pom.xml
+++ b/utils/flink/pom.xml
@@ -131,6 +131,21 @@
                     <exclude>module-info.class</exclude>
                   </excludes>
                 </filter>
+                <!-- jackson-core ships its vendored FastDoubleParser as a multi-release jar.
+                     maven-shade-plugin's relocator rewrites the class bytecode of the
+                     META-INF/versions/{N}/ variants but doesn't move the jar entry to the
+                     relocated path (MSHADE-406, open/unresolved as of shade-plugin 3.6.2 -
+                     the latest release). That mismatch between entry path and internal class
+                     name makes the leftover entries unloadable, so they're excluded outright;
+                     the relocated base (non-versioned) implementation is a fully functional
+                     fallback on every JDK version, so this only drops a JDK-specific
+                     micro-optimization, not functionality. -->
+                <filter>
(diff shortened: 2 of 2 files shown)


carlesarnal:
Closing as stale, draft PR for two weeks.

PR #9573, opened 2026-08-14
fix(ci): pre-create CNI directory before setup-minikube to prevent chmod error

aaravmahajanofficial:
# Description
When `setup-minikube` runs with `driver: 'none'` (via medyagh/setup-minikube), it unconditionally executes `sudo chmod 755 /etc/cni/net.d`. On GitHub Actions runner images, `/etc/cni/net.d` does not exist by default, causing the step to fail with:

>  chmod: cannot access '/etc/cni/net.d': No such file or directory
>   Error: The process '/usr/bin/sudo' failed with exit code 1

# Reason for Regression
In PR #8824, this issue (https://github.com/Apicurio/apicurio-registry/pull/8824#discussion_r3686334035) was identified and referenced in the commit message, but the step pre-creating `/etc/cni/net.d` was accidentally omitted/dropped during PR review revisions, leaving empty lines before `setup-minikube`. 

# Changes
Pre-created `/etc/cni/net.d` with 755 permissions before `setup-minikube` in:
`.github/workflows/operator.yaml`
`.github/workflows/release-operator.yaml`
`.github/workflows/verify-integration-tests.yaml`


aaravmahajanofficial:
Hi @EricWittmann, @paoloantinori, please review this PR whenever you have a moment. Thanks!

paoloantinori:
Had a look. The fix is correct: setup-minikube with driver=none unconditionally chmod's /etc/cni/net.d, which does not exist on the standard runner image, so pre-creating it is the right remediation. Good catch restoring the step that got dropped in #8824.

One small note: the PR description lists two workflow files, but the diff also adds the same step to verify-integration-tests.yaml, which is consistent (same setup-minikube call). Worth updating the description to match, but nothing blocking from my side.

The change is mechanical and the three insertion points are identical, so nothing else to flag.

aaravmahajanofficial:
Hi @jsenko, just gently bumping this PR in case it slipped through your notifications. Thanks!

PR #9578, opened 2026-08-14
fix(kafkasql): fail when response timeout expires

### LINKED ISSUE #9574: KafkaSQL waitForResponse silently returns null when response timeout expires
## Description

`KafkaSqlCoordinator.waitForResponse(UUID)` silently returns `null` when the configured KafkaSQL response timeout expires.

The method currently calls `CountDownLatch.await(timeout, unit)` but ignores its boolean return value. `await()` returns `false` when the timeout expires before the latch reaches zero.

The code then executes:

```java
Object rval = returnValues.remove(uuid);  
If the Kafka consumer has not processed the response yet, returnValues.remove(uuid) returns ordinary Java null.

This is different from the NULL sentinel used by the coordinator:

private static final Object NULL = new Object();

Therefore rval == NULL is false and the method returns null without throwing a timeout exception.

Affected code

File:

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlCoordinator.java

Method:

waitForResponse(UUID uuid)

Current code:

latches.get(uuid).await(
        configuration.get().getResponseTimeout().toMillis(),
        TimeUnit.MILLISECONDS);


Object rval = returnValues.remove(uuid);

The return value of await() is ignored.

Environment
Apicurio Registry: main / snapshot
Persistence: KafkaSQL
Java: JDK 21
Logging: LOG_LEVEL=DEBUG
Steps to Reproduce
Run Apicurio Registry using KafkaSQL persistence.

Configure a very small KafkaSQL response timeout, for example:

apicurio.kafkasql.response-timeout.ms=1

Start Registry with DEBUG logging enabled.
Submit a write operation through the Registry REST API, such as creating an artifact.
If the Kafka consumer does not process the response before the timeout, await() returns false.
The return value is ignored.
returnValues.remove(uuid) returns null.
waitForResponse() returns null instead of reporting a timeout.
Expected vs Actual Behaviour

Expected behaviour:

When the response timeout expires, waitForResponse() should throw a RegistryException indicating that the KafkaSQL operation timed out.

For example:

boolean completed = latches.get(uuid).await(
        configuration.get().getResponseTimeout().toMillis(),
        TimeUnit.MILLISECONDS);


if (!completed) {
    throw new RegistryException(
            "[KafkaSqlCoordinator] Timed out waiting for a Kafka Sql response for operation " + uuid);
}

Actual behaviour:

The timeout result is ignored and waitForResponse() returns ordinary null.

This can cause callers such as createArtifact() and createArtifactVersion() to receive an unexpected null result instead of a clear timeout exception.

Logs

A related symptom can look like:

Cannot invoke "java.lang.Long.longValue()"
because the return value of
"io.apicurio.registry.storage.impl.kafkasql.KafkaSqlCoordinator.waitForResponse(java.util.UUID)"
is null

### PR DESCRIPTION
**Pull request by sahu-virendra-1908**
## Description

Fixes #9574

`KafkaSqlCoordinator.waitForResponse(UUID)` currently ignores the boolean result returned by `CountDownLatch.await(...)`.

When the configured KafkaSQL response timeout expires before the Kafka consumer processes the response, `await()` returns `false`. The code then calls `returnValues.remove(uuid)`, which returns ordinary `null` because no response has been stored yet.

This causes `waitForResponse()` to return `null` instead of reporting that the KafkaSQL operation timed out.

### Changes

- Capture the boolean result of `CountDownLatch.await(...)`.
- Throw a `RegistryException` when the configured response timeout expires.
- Preserve the existing response handling and latch cleanup logic.

### Affected file

`app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlCoordinator.java`

### Before

```java
latches.get(uuid).await(
        configuration.get().getResponseTimeout().toMillis(),
        TimeUnit.MILLISECONDS);

Object rval = returnValues.remove(uuid);

When the timeout expires, returnValues.remove(uuid) can return null and the method silently returns it.

After
boolean completed = latches.get(uuid).await(
        configuration.get().getResponseTimeout().toMillis(), TimeUnit.MILLISECONDS);


if (!completed) {
    throw new RegistryException(
            "[KafkaSqlCoordinator] Timed out waiting for a Kafka Sql response for operation " + uuid);
}

This makes the timeout explicit and prevents a missing KafkaSQL response from being treated as a valid null result.

Testing
Checkstyle: passed with 0 violations.
Java compilation: passed.
Full app test execution was attempted locally, but the Surefire forked JVM terminated because the Windows paging file was too small; this was an environment/resource failure rather than a test assertion failure.

### CODE REVIEW

**Comment by sahu-virendra-1908** [comment]
Hi [@EricWittmann](https://github.com/EricWittmann), [@paoloantinori](https://github.com/paoloantinori), @carlesarnal @vandanayadav7

Please review this PR whenever you have a moment. I’d really appreciate your feedback.
Also, if there are any other tasks or issues I can help with, please feel free to assign one to me. I’d be happy to contribute. Thanks!

File: app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlCoordinator.java [hunk 1]
```diff
@@ -47,7 +47,13 @@ public UUID createUUID() {
      */
     public Object waitForResponse(UUID uuid) {
         try {
-            latches.get(uuid).await(configuration.get().getResponseTimeout().toMillis(), TimeUnit.MILLISECONDS);
+           boolean completed = latches.get(uuid).await(
+                configuration.get().getResponseTimeout().toMillis(), TimeUnit.MILLISECONDS);
+
+            if (!completed) {
+            throw new RegistryException(
```
paoloantinori: This new `throw` branch is a code path with no test coverage. Add a unit test (e.g. in a `KafkaSqlCoordinatorTest`) that calls `waitForResponse` for a uuid whose latch is never counted down and asserts `RegistryException` is thrown. The Windows paging-file issue explains why the suite was not executed locally, but the test still needs to exist and pass in CI.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlCoordinator.java
```diff
@@ -47,7 +47,13 @@ public UUID createUUID() {
...
+                    "[KafkaSqlCoordinator] Timed out waiting for a Kafka Sql response for operation " + uuid);
+          }
 
             Object rval = returnValues.remove(uuid);
```
paoloantinori, line 57: Throwing on timeout exits before `returnValues.remove(uuid)` runs. If the cleanup for this uuid (both `returnValues` and `latches` entries) lives on the normal return path rather than in a `finally`, the timeout path now skips it and leaks entries in the `latches` map on every expired operation. Confirm the entries are still cleared when the exception fires, or move the cleanup into a `finally` block.

app/src/main/java/io/apicurio/registry/storage/impl/kafkasql/KafkaSqlCoordinator.java
```diff
@@ -47,7 +47,13 @@ public UUID createUUID() {
      */
     public Object waitForResponse(UUID uuid) {
         try {
-            latches.get(uuid).await(configuration.get().getResponseTimeout().toMillis(), TimeUnit.MILLISECONDS);
+           boolean completed = latches.get(uuid).await(
```
paoloantinori: The indentation of the added block looks off in the diff: the `boolean completed` line is 11 columns, and the `throw` at line 54 is not indented relative to its `if`. The PR notes Checkstyle passed with 0 violations, so this may be a paste artifact, but confirm Checkstyle actually enforces indentation here so the malformed block does not ship.

paoloantinori (commented):
The fix is correct in direction: capturing the result of `CountDownLatch.await(...)` and throwing `RegistryException` on timeout turns a silent null return into a loud failure, which is the right behavior. Three points below before this is mergeable.

- `KafkaSqlCoordinator.java:54` - the new `throw` branch is a code path with no test coverage; the project treats a missing test on a new code path as a rejection.
- `KafkaSqlCoordinator.java:58` - throwing on timeout exits before `returnValues.remove(uuid)` runs, so if the latch/returnValues cleanup for this uuid lives on the normal return path instead of in a `finally`, the timeout path now leaks map entries on every expired operation.
- `KafkaSqlCoordinator.java:50` - the indentation of the added block looks malformed in the diff; worth confirming Checkstyle actually enforces it so the block does not ship that way.

sahu-virendra-1908:
Hi @paoloantinori, thanks for the detailed review and feedback. I’ve addressed all the points you mentioned, including the timeout-path test coverage, cleanup handling, and code indentation.
I’ve pushed the updates to the PR. Whenever you have a chance, could you please take another look and let me know if there’s anything else I should improve?
I’d also be happy to work on another issue or task if there’s something you’d like me to pick up. Thanks!

paoloantinori:
Thanks — re-read at the new head, and both substantive points are addressed: the timeout path now has test coverage (KafkaSqlCoordinatorTest.testWaitForResponseTimesOut asserts the RegistryException), and the returnValues cleanup moved into the finally block so the timeout path no longer leaks entries. From my side this is ready pending CI.

One last nit, cosmetic only: the first line of the new block still reads `boolean completed = latches.get(uuid).await(` at 11 columns instead of 12, and there is a line of trailing whitespace just before the existing blank line after the closing brace. Worth a quick cleanup pass so the block matches the surrounding style, but nothing that changes behavior.

sahu-virendra-1908:
Hi @paoloantinori, thanks for the detailed review and feedback. I’ve addressed all the points you mentioned .I’ve pushed the updates to the PR. Whenever you have a chance, could you please take another look and let me know if there’s anything else I should improve?
I’d also be happy to work on another issue or task if there’s something you’d like me to pick up. Thanks

paoloantinori:
Thanks — re-read the latest head (ef10326d). The new commit is formatting-only from my side: indentation alignment and trailing whitespace, no behavior change. Everything I raised earlier is addressed (timeout path covered by testWaitForResponseTimesOut, cleanup in the finally block), so nothing further from me on this PR.

On picking up other work — that is appreciated. I do not assign tasks myself, but the issues labeled `good first issue` / `help wanted` in this repo are a fair starting point, and anything tagged with the current milestone tends to get maintainer attention faster.

sahu-virendra-1908:
@paoloantinori  Thanks for the detailed review and guidance

paoloantinori:
Happy to help — thanks for iterating on it. The timeout-path test and the latch/returnValues cleanup make this a solid fix. Nothing further from my side.

EricWittmann:
Thanks for the contribution, @sahu-virendra-1908! Great job iterating on the review feedback — the fix is clean and well-tested. Merged!

Issue #8995, opened 2026-07-29
Expose Schema Resolver client retry configuration through SchemaResolverConfig


doraem-on:
### Feature or Problem Description

Currently, in `RegistryClientFacadeFactory.java`, the retry options for the underlying Kiota-based registry client are hardcoded to always be enabled:

```java
// FIXME push retry options into the SchemaResolverConfig
if (Boolean.TRUE) {
    clientOptions.retry();
}
```

Because of this hardcoded behavior, consumers of the Schema Resolver cannot customize the retry logic (such as configuring maximum retry attempts or backoff parameters) and cannot disable retries when a fail-fast strategy is preferred. This enforces a single resiliency pattern for all consumers, regardless of their application's requirements.

Making the retry behavior configurable would improve the flexibility of the Schema Resolver and allow applications to choose retry policies that best suit their deployment and operational needs.

**Location:**

`schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java`

Source:
https://github.com/Apicurio/apicurio-registry/blob/main/schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java#L104-L107

---

### Proposed Solution

Expose retry-related configuration through `SchemaResolverConfig` and configure the underlying registry client using those values instead of always invoking `clientOptions.retry()`.

Possible configuration options could include:

- Enable or disable retries
- Maximum retry attempts
- Retry delay or backoff configuration
- Additional retry-related settings supported by the underlying client

The exact configuration API can follow the existing conventions used throughout the Schema Resolver. To preserve backward compatibility, the default behavior can remain unchanged unless explicitly configured by the user.

---

### Additional Context

The source code already contains the following `FIXME` comment:

```java
// FIXME push retry options into the SchemaResolverConfig
```

This indicates that the current implementation is intended to be temporary and that retry configuration was planned to be exposed through `SchemaResolverConfig`.

Making retry behavior configurable would provide greater flexibility for SDK consumers while allowing applications to integrate the registry client with their own resiliency, timeout, and retry policies.

doraem-on:
Hi @paoloantinori , @carlesarnal , @EricWittmann 

I found this FIXME while exploring the Schema Resolver module and opened this issue to discuss exposing the retry configuration through `SchemaResolverConfig`.

[...]

PR #9581, opened 2026-08-14
feat(schema-resolver): expose client retry config and retry registry outages

sameer-bm-ai:
## Summary
- Exposes HTTP client retry settings via SerDes/converter config (based on #9091 / #8995): `apicurio.registry.client.retry.{enabled,max-attempts,delay-ms,backoff-multiplier,max-delay-ms}` and wires them into `RegistryClientOptions` instead of hardcoded `clientOptions.retry()`.
- Extends `ERCache` retries beyond HTTP 429 to transient network / registry-down failures (`ConnectException`, timeouts, connection reset, Vert.x `HttpClosedException`), controlled by existing `apicurio.registry.retry-count` / `retry-backoff-ms`.
- Adds production-style logging: INFO for effective client retry config, FINE/debug for per-attempt retries, WARNING/error when retries are exhausted.

## Motivation
Kafka Connect / Debezium converters fail hard on short Apicurio outages even when operators set `retry-count`. The existing cache retry only covers rate-limits, and HTTP client retries were not configurable from SerDes keys.

## Test plan
- [ ] `mvn -pl schema-resolver,java-sdk/common -am test`
- [ ] Configure converter with elevated `apicurio.registry.client.retry.max-attempts` / `delay-ms` and `apicurio.registry.retry-count`; confirm INFO log shows the configured values at client create
- [ ] Stop Apicurio briefly during auto-register / schema resolve; confirm retries occur and the task recovers when the registry returns (instead of failing after ~2s of default HTTP retries)
- [ ] Confirm non-retriable errors still fail without spinning

### DIFF
--- a/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
+++ b/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
@@ -83,17 +83,14 @@ public class RegistryClientRequestAdapterFactory {
 
         // Wrap with retry proxy if retry is enabled
         if (options.isRetryEnabled()) {
-            log.log(Level.INFO,
-                    "[apicurio-pr9091] wrapping RequestAdapter with retry maxAttempts={0} delayMs={1} backoff={2} maxDelayMs={3}",
+            log.log(Level.FINE,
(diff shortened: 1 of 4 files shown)


sameer-bm-ai:
Closing so this can be reopened from a personal account.

PR #9582, opened 2026-08-14
feat(schema-resolver): expose client retry config and retry registry outages

Mahamood-Sameer:
## Summary
- Exposes HTTP client retry settings via SerDes/converter config (based on #9091 / #8995): `apicurio.registry.client.retry.{enabled,max-attempts,delay-ms,backoff-multiplier,max-delay-ms}` and wires them into `RegistryClientOptions` instead of hardcoded `clientOptions.retry()`.
- Extends `ERCache` retries beyond HTTP 429 to transient network / registry-down failures (`ConnectException`, timeouts, connection reset, Vert.x `HttpClosedException`), controlled by existing `apicurio.registry.retry-count` / `retry-backoff-ms`.
- Adds production-style logging: INFO for effective client retry config, FINE/debug for per-attempt retries, WARNING/error when retries are exhausted.

## Motivation
Kafka Connect / Debezium converters fail hard on short Apicurio outages even when operators set `retry-count`. The existing cache retry only covers rate-limits, and HTTP client retries were not configurable from SerDes keys.

## Test plan
- [ ] `mvn -pl schema-resolver,java-sdk/common -am test`
- [ ] Configure converter with elevated `apicurio.registry.client.retry.max-attempts` / `delay-ms` and `apicurio.registry.retry-count`; confirm INFO log shows the configured values at client create
- [ ] Stop Apicurio briefly during auto-register / schema resolve; confirm retries occur and the task recovers when the registry returns (instead of failing after ~2s of default HTTP retries)
- [ ] Confirm non-retriable errors still fail without spinning

### DIFF
--- a/docs/modules/ROOT/pages/getting-started/assembly-configuring-kafka-client-serdes.adoc
+++ b/docs/modules/ROOT/pages/getting-started/assembly-configuring-kafka-client-serdes.adoc
@@ -308,17 +308,17 @@ The `DefaultSchemaResolver` uses the following properties to configure how to lo
 |`300`
 |`RETRY_COUNT`
 |`apicurio.registry.retry-count`
-|Used by serializers and deserializers. Controls how many times the schema cache retries a failed load before failing. Distinct from `apicurio.registry.client.retry.max-attempts` (HTTP client retry ladder). By default the cache only retries HTTP 429; set `apicurio.registry.retry.transient-errors=true` to also retry network/outage failures. When that opt-in is enabled, worst-case blocking multiplies across both layers — size together with the client retry keys, especially for Kafka producers (`max.block.ms`) and Connect polls.
+|Used by serializers and deserializers. **To retry registry outages, set `apicurio.registry.retry.transient-errors=true` and preferably `apicurio.registry.retry.total-timeout-ms`.** This key then controls how many times the schema cache retries a failed load. Distinct from `apicurio.registry.client.retry.max-attempts` (HTTP client retry ladder). By default the cache only retries HTTP 429; raising `retry-count` alone does not cover network or registry-down failures. When the opt-in is enabled, worst-case blocking multiplies across both layers — size together with the client retry keys, especially for Kafka producers (`max.block.ms`) and Connect polls.
 |`non-negative Number, or integer String`
 |`3`
 |`RETRY_TRANSIENT_ERRORS`
 |`apicurio.registry.retry.transient-errors`
-|Used by serializers and deserializers. When `false` (default), schema-cache retries stay limited to HTTP 429 (historical behavior). When `true`, also retries 502/503/504 and retriable network failures (including `NoRouteToHostException`). Enabling this increases worst-case blocking because each cache attempt may run a full HTTP client retry ladder. Prefer setting `apicurio.registry.retry.total-timeout-ms` when enabling.
+|Used by serializers and deserializers. **Activation step for outage retries.** When `false` (default), schema-cache retries stay limited to HTTP 429 (historical behavior). When `true`, also retries 502/503/504 and retriable network failures (including `NoRouteToHostException`). Enabling this increases worst-case blocking because each cache attempt may run a full HTTP client retry ladder. Set `apicurio.registry.retry.total-timeout-ms` when enabling.
 |`boolean`
 |`false`
 |`RETRY_TOTAL_TIMEOUT_MS`
 |`apicurio.registry.retry.total-timeout-ms`
-|Used by serializers and deserializers. Optional wall-clock budget (milliseconds) for one schema-cache load retry loop. Checked before each re-attempt. `0` means unlimited. Recommended when `apicurio.registry.retry.transient-errors=true` to bound blocking on Kafka `Producer.send()` / Connect polls.
+|Used by serializers and deserializers. Optional wall-clock budget (milliseconds) for one schema-cache load retry loop. Checked before each re-attempt; it does not interrupt an in-flight cache attempt (that attempt may still run a full HTTP client retry ladder and request timeouts). `0` means unlimited. Recommended when `apicurio.registry.retry.transient-errors=true` to bound blocking on Kafka `Producer.send()` / Connect polls.
 |`java.time.Duration, non-negative Number, or integer String`
 |`0`
 |`CLIENT_RETRY_ENABLED`
@@ -328,8 +328,8 @@ The `DefaultSchemaResolver` uses the following properties to configure how to lo
 |`true`
 |`CLIENT_RETRY_MAX_ATTEMPTS`
 |`apicurio.registry.client.retry.max-attempts`
-|Used by serializers and deserializers. Maximum attempts for the HTTP client retry ladder. Defaults match `RegistryClientOptions#retry()` (3 / 250ms / 2.0 / 10000ms).
-|`integer in [1, Integer.MAX_VALUE]`
+|Used by serializers and deserializers. Maximum attempts for the HTTP client retry ladder. Defaults match `RegistryClientOptions#retry()` (3 / 250ms / 2.0 / 10000ms). Values above `1000` are rejected as a sanity ceiling (not a recommended operating point).
+|`integer in [1, 1000]`
 |`3`
 |`CLIENT_RETRY_DELAY_MS`
 |`apicurio.registry.client.retry.delay-ms`
--- a/docs/modules/ROOT/pages/getting-started/assembly-configuring-kafka-client-serdes.adoc
+++ b/docs/modules/ROOT/pages/getting-started/assembly-configuring-kafka-client-serdes.adoc
@@ -313,9 +313,14 @@ The `DefaultSchemaResolver` uses the following properties to configure how to lo
 |`3`
 |`RETRY_TRANSIENT_ERRORS`
 |`apicurio.registry.retry.transient-errors`
-|Used by serializers and deserializers. When `false` (default), schema-cache retries stay limited to HTTP 429 (historical behavior). When `true`, also retries 502/503/504 and retriable network failures. Enabling this increases worst-case blocking because each cache attempt may run a full HTTP client retry ladder.
+|Used by serializers and deserializers. When `false` (default), schema-cache retries stay limited to HTTP 429 (historical behavior). When `true`, also retries 502/503/504 and retriable network failures (including `NoRouteToHostException`). Enabling this increases worst-case blocking because each cache attempt may run a full HTTP client retry ladder. Prefer setting `apicurio.registry.retry.total-timeout-ms` when enabling.
 |`boolean`
 |`false`
+|`RETRY_TOTAL_TIMEOUT_MS`
+|`apicurio.registry.retry.total-timeout-ms`
+|Used by serializers and deserializers. Optional wall-clock budget (milliseconds) for one schema-cache load retry loop. Checked before each re-attempt. `0` means unlimited. Recommended when `apicurio.registry.retry.transient-errors=true` to bound blocking on Kafka `Producer.send()` / Connect polls.
+|`java.time.Duration, non-negative Number, or integer String`
+|`0`
 |`CLIENT_RETRY_ENABLED`
 |`apicurio.registry.client.retry.enabled`
 |Used by serializers and deserializers. Enables the underlying registry HTTP client retry ladder (`RegistryClientOptions`). Distinct from `apicurio.registry.retry-count` (schema-cache retries).
--- a/docs/modules/ROOT/pages/getting-started/assembly-configuring-kafka-client-serdes.adoc
+++ b/docs/modules/ROOT/pages/getting-started/assembly-configuring-kafka-client-serdes.adoc
@@ -308,9 +308,39 @@ The `DefaultSchemaResolver` uses the following properties to configure how to lo
 |`300`
 |`RETRY_COUNT`
 |`apicurio.registry.retry-count`
-|Used by serializers and deserializers. If the SerDes cannot retrieve a schema from the registry, they can retry a number of times. This configuration option controls the number of retry attempts.
+|Used by serializers and deserializers. Controls how many times the schema cache retries a failed load before failing. Distinct from `apicurio.registry.client.retry.max-attempts` (HTTP client retry ladder). By default the cache only retries HTTP 429; set `apicurio.registry.retry.transient-errors=true` to also retry network/outage failures. When that opt-in is enabled, worst-case blocking multiplies across both layers — size together with the client retry keys, especially for Kafka producers (`max.block.ms`) and Connect polls.
 |`non-negative Number, or integer String`
 |`3`
+|`RETRY_TRANSIENT_ERRORS`
+|`apicurio.registry.retry.transient-errors`
+|Used by serializers and deserializers. When `false` (default), schema-cache retries stay limited to HTTP 429 (historical behavior). When `true`, also retries 502/503/504 and retriable network failures. Enabling this increases worst-case blocking because each cache attempt may run a full HTTP client retry ladder.
+|`boolean`
+|`false`
+|`CLIENT_RETRY_ENABLED`
+|`apicurio.registry.client.retry.enabled`
+|Used by serializers and deserializers. Enables the underlying registry HTTP client retry ladder (`RegistryClientOptions`). Distinct from `apicurio.registry.retry-count` (schema-cache retries).
+|`boolean`
+|`true`
+|`CLIENT_RETRY_MAX_ATTEMPTS`
+|`apicurio.registry.client.retry.max-attempts`
+|Used by serializers and deserializers. Maximum attempts for the HTTP client retry ladder. Defaults match `RegistryClientOptions#retry()` (3 / 250ms / 2.0 / 10000ms).
+|`integer in [1, Integer.MAX_VALUE]`
+|`3`
+|`CLIENT_RETRY_DELAY_MS`
+|`apicurio.registry.client.retry.delay-ms`
+|Used by serializers and deserializers. Initial delay (milliseconds) for the HTTP client retry ladder.
+|`java.time.Duration, non-negative Number, or integer String`
+|`250`
+|`CLIENT_RETRY_BACKOFF_MULTIPLIER`
+|`apicurio.registry.client.retry.backoff-multiplier`
+|Used by serializers and deserializers. Exponential backoff multiplier for the HTTP client retry ladder. Must be finite and greater than `1.0` (same constraint as `RegistryClientOptions#retry`).
+|`Number or numeric String`
+|`2.0`
+|`CLIENT_RETRY_MAX_DELAY_MS`
+|`apicurio.registry.client.retry.max-delay-ms`
+|Used by serializers and deserializers. Maximum delay (milliseconds) between HTTP client retry attempts.
+|`java.time.Duration, non-negative Number, or integer String`
+|`10000`
 |`FAULT_TOLERANT_REFRESH`
 |`apicurio.registry.fault-tolerant-refresh`
 |Used by serializers and deserializers. If `true`, the SerDes log exceptions instead of throwing them when an error occurs while refreshing a schema in the cache. This is useful for production situations where a stale schema is better than completely failing schema resolution. This does not affect retry attempts, because the SerDes attempt retries before considering this flag.
--- a/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
+++ b/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
@@ -83,17 +83,14 @@ public class RegistryClientRequestAdapterFactory {
 
         // Wrap with retry proxy if retry is enabled
         if (options.isRetryEnabled()) {
-            log.log(Level.INFO,
-                    "[apicurio-pr9091] wrapping RequestAdapter with retry maxAttempts={0} delayMs={1} backoff={2} maxDelayMs={3}",
+            log.log(Level.FINE,
+                    "Enabling registry HTTP client retry: maxAttempts={0}, delayMs={1}, backoffMultiplier={2}, maxDelayMs={3}",
                     new Object[]{
                             options.getMaxRetryAttempts(),
                             options.getRetryDelayMs(),
                             options.getBackoffMultiplier(),
                             options.getMaxRetryDelayMs()
                     });
-            System.out.println("[apicurio-pr9091] RequestAdapter retry proxy maxAttempts="
-                    + options.getMaxRetryAttempts()
-                    + " delayMs=" + options.getRetryDelayMs());
             adapter = createRetryProxy(adapter, options);
         }
 
@@ -215,8 +212,8 @@ public class RegistryClientRequestAdapterFactory {
                     if (isRetryable(cause) && attempt < maxRetryAttempts) {
                         attempt++;
                         long delayMs = calculateRetryDelay(attempt);
-                        log.log(Level.WARNING,
-                                "[apicurio-pr9091] HTTP retry {0}/{1} after {2}: {3}; sleeping {4}ms",
+                        log.log(Level.FINE,
+                                "Retryable registry HTTP failure on attempt {0}/{1} ({2}: {3}); sleeping {4}ms before retry",
                                 new Object[]{
                                         attempt,
                                         maxRetryAttempts,
@@ -224,9 +221,6 @@ public class RegistryClientRequestAdapterFactory {
                                         cause.getMessage(),
                                         delayMs
                                 });
-                        System.out.println("[apicurio-pr9091] HTTP retry " + attempt + "/" + maxRetryAttempts
-                                + " sleepMs=" + delayMs + " err=" + cause.getClass().getSimpleName()
-                                + ": " + cause.getMessage());
                         try {
                             Thread.sleep(delayMs);
                         } catch (InterruptedException interruptedException) {
@@ -235,13 +229,9 @@ public class RegistryClientRequestAdapterFactory {
                         }
                     } else {
                         if (isRetryable(cause) && attempt >= maxRetryAttempts) {
-                            log.log(Level.WARNING, "Maximum retry attempts ({0}) exceeded for {1}: {2}",
+                            log.log(Level.WARNING,
+                                    "Registry HTTP client exhausted {0} retry attempts due to {1}: {2}",
                                     new Object[]{maxRetryAttempts, cause.getClass().getName(), cause.getMessage()});
-                            System.out.println("[apicurio-pr9091] HTTP retry EXHAUSTED maxAttempts="
-                                    + maxRetryAttempts + " last=" + cause.getMessage());
-                        } else if (!isRetryable(cause)) {
-                            System.out.println("[apicurio-pr9091] HTTP error NOT retryable: "
-                                    + cause.getClass().getName() + ": " + cause.getMessage());
                         }
                         throw originalCause;
                     }
--- a/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
+++ b/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
@@ -81,16 +81,8 @@ public class RegistryClientRequestAdapterFactory {
 
         adapter.setBaseUrl(registryUrl);
 
-        // Wrap with retry proxy if retry is enabled
+        // Wrap with retry proxy if retry is enabled (settings already logged at client creation).
         if (options.isRetryEnabled()) {
-            log.log(Level.FINE,
-                    "Enabling registry HTTP client retry: maxAttempts={0}, delayMs={1}, backoffMultiplier={2}, maxDelayMs={3}",
-                    new Object[]{
-                            options.getMaxRetryAttempts(),
-                            options.getRetryDelayMs(),
-                            options.getBackoffMultiplier(),
-                            options.getMaxRetryDelayMs()
-                    });
             adapter = createRetryProxy(adapter, options);
         }
 
@@ -229,9 +221,10 @@ public class RegistryClientRequestAdapterFactory {
                         }
                     } else {
                         if (isRetryable(cause) && attempt >= maxRetryAttempts) {
+                            String causeMessage = cause.getMessage() != null ? cause.getMessage() : "<no message>";
                             log.log(Level.WARNING,
                                     "Registry HTTP client exhausted {0} retry attempts due to {1}: {2}",
-                                    new Object[]{maxRetryAttempts, cause.getClass().getName(), cause.getMessage()});
+                                    new Object[]{maxRetryAttempts, cause.getClass().getName(), causeMessage});
                         }
                         throw originalCause;
                     }
--- a/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
+++ b/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
@@ -226,7 +226,11 @@ public class RegistryClientRequestAdapterFactory {
                                     "Registry HTTP client exhausted {0} retry attempts due to {1}: {2}",
                                     new Object[]{maxRetryAttempts, cause.getClass().getName(), causeMessage});
                         }
-                        throw originalCause;
+                        // Prefer the latest failure for caller diagnostics; keep the first as suppressed.
+                        if (originalCause != null && originalCause != cause) {
+                            cause.addSuppressed(originalCause);
+                        }
+                        throw cause;
                     }
                 }
             }
--- a/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
+++ b/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
@@ -83,6 +83,17 @@ public class RegistryClientRequestAdapterFactory {
 
         // Wrap with retry proxy if retry is enabled
         if (options.isRetryEnabled()) {
+            log.log(Level.INFO,
+                    "[apicurio-pr9091] wrapping RequestAdapter with retry maxAttempts={0} delayMs={1} backoff={2} maxDelayMs={3}",
+                    new Object[]{
+                            options.getMaxRetryAttempts(),
+                            options.getRetryDelayMs(),
+                            options.getBackoffMultiplier(),
+                            options.getMaxRetryDelayMs()
+                    });
+            System.out.println("[apicurio-pr9091] RequestAdapter retry proxy maxAttempts="
+                    + options.getMaxRetryAttempts()
+                    + " delayMs=" + options.getRetryDelayMs());
             adapter = createRetryProxy(adapter, options);
         }
 
@@ -204,6 +215,18 @@ public class RegistryClientRequestAdapterFactory {
                     if (isRetryable(cause) && attempt < maxRetryAttempts) {
                         attempt++;
                         long delayMs = calculateRetryDelay(attempt);
+                        log.log(Level.WARNING,
+                                "[apicurio-pr9091] HTTP retry {0}/{1} after {2}: {3}; sleeping {4}ms",
+                                new Object[]{
+                                        attempt,
+                                        maxRetryAttempts,
+                                        cause.getClass().getName(),
+                                        cause.getMessage(),
+                                        delayMs
+                                });
+                        System.out.println("[apicurio-pr9091] HTTP retry " + attempt + "/" + maxRetryAttempts
+                                + " sleepMs=" + delayMs + " err=" + cause.getClass().getSimpleName()
+                                + ": " + cause.getMessage());
                         try {
                             Thread.sleep(delayMs);
                         } catch (InterruptedException interruptedException) {
@@ -214,6 +237,11 @@ public class RegistryClientRequestAdapterFactory {
                         if (isRetryable(cause) && attempt >= maxRetryAttempts) {
                             log.log(Level.WARNING, "Maximum retry attempts ({0}) exceeded for {1}: {2}",
                                     new Object[]{maxRetryAttempts, cause.getClass().getName(), cause.getMessage()});
+                            System.out.println("[apicurio-pr9091] HTTP retry EXHAUSTED maxAttempts="
+                                    + maxRetryAttempts + " last=" + cause.getMessage());
+                        } else if (!isRetryable(cause)) {
+                            System.out.println("[apicurio-pr9091] HTTP error NOT retryable: "
+                                    + cause.getClass().getName() + ": " + cause.getMessage());
                         }
                         throw originalCause;
                     }
--- a/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
+++ b/java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
@@ -226,7 +226,9 @@ public class RegistryClientRequestAdapterFactory {
                                     "Registry HTTP client exhausted {0} retry attempts due to {1}: {2}",
                                     new Object[]{maxRetryAttempts, cause.getClass().getName(), causeMessage});
                         }
-                        // Prefer the latest failure for caller diagnostics; keep the first as suppressed.
+                        // Public behavior: throw the latest failure, not the first. Callers that
+                        // catch a specific type (e.g. HttpClosedException) may now see a later
+                        // type (e.g. ConnectException) with the first on getSuppressed().
                         if (originalCause != null && originalCause != cause) {
                             cause.addSuppressed(originalCause);
                         }
--- a/schema-resolver/src/main/java/io/apicurio/registry/resolver/AbstractSchemaResolver.java
+++ b/schema-resolver/src/main/java/io/apicurio/registry/resolver/AbstractSchemaResolver.java
@@ -328,7 +328,7 @@ public abstract class AbstractSchemaResolver<S, T> implements SchemaResolver<S, 
                 break;
             }
             double next = step * multiplier;
-            if (!Double.isFinite(next) || next >= (double) maxDelayMs) {
+            if (!Double.isFinite(next) || next >= maxDelayMs) {
                 step = maxDelayMs;
             } else {
                 step = Math.max(1L, (long) next);
--- a/schema-resolver/src/main/java/io/apicurio/registry/resolver/AbstractSchemaResolver.java
+++ b/schema-resolver/src/main/java/io/apicurio/registry/resolver/AbstractSchemaResolver.java
@@ -311,7 +311,7 @@ public abstract class AbstractSchemaResolver<S, T> implements SchemaResolver<S, 
         long maxDelayMs = config.getClientRetryMaxDelayMs();
         long sleepMs = 0L;
         for (long attempt = 1; attempt < maxAttempts; attempt++) {
-            long step = (long) (delayMs * Math.pow(multiplier, attempt - 1));
+            long step = (long) (delayMs * Math.pow(multiplier, (double) attempt - 1));
             sleepMs += Math.min(step, maxDelayMs);
         }
         return sleepMs;
--- a/schema-resolver/src/main/java/io/apicurio/registry/resolver/AbstractSchemaResolver.java
+++ b/schema-resolver/src/main/java/io/apicurio/registry/resolver/AbstractSchemaResolver.java
@@ -297,8 +297,10 @@ public abstract class AbstractSchemaResolver<S, T> implements SchemaResolver<S, 
 
     /**
      * Lower-bound estimate of sleep time for one HTTP client retry ladder (excludes request latency).
+     * Closed form once the per-step delay hits {@code maxDelayMs}, so a large
+     * {@code max-attempts} cannot hang {@code configure()}.
      */
-    private static long estimateClientRetryLadderSleepMs(SchemaResolverConfig config) {
+    static long estimateClientRetryLadderSleepMs(SchemaResolverConfig config) {
         if (!config.getClientRetryEnabled()) {
             return 0L;
         }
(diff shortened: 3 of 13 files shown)


schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +456,55 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
+        try {
+            cls = Class.forName("io.vertx.core.http.HttpClosedException", false,
+                    ERCache.class.getClassLoader());
+        } catch (ClassNotFoundException ignored) {
+            // Vert.x may not be on the classpath for all consumers.
+        }
+        VERTX_HTTP_CLOSED_EXCEPTION = cls;
+    }
+
+    private static boolean isRetriableCacheLoadFailure(Throwable e) {
```
paoloantinori: This predicate (and the new config getters) ship with no automated coverage. Please add unit tests before merge. For ERCache: 429 `ApiException` direct and wrapped one and two cause levels deep, `ConnectException`, `SocketTimeoutException`, `IOException` with "Connection reset" in the message, `IOException` with a null message (negative), plain `RuntimeException` (negative), and 404/500 `ApiException` (negative). Making the method package-private for the test is fine. For `SchemaResolverConfig`: the backoff-multiplier parse paths (Number, String, invalid String) and the defaults of all five new keys. These tests should fail on current main since the methods are new.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +456,55 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
+        Throwable current = e;
+        while (current != null) {
+            if (current instanceof ApiException api && api.getResponseStatusCode() == 429) {
```
paoloantinori: Only 429 is retriable among `ApiException`s. When the registry is down behind a load balancer or restarting, the client usually sees 502/503/504 rather than `ConnectException`, so the outage scenario this PR targets can still fail the cache load on the first attempt. Consider treating 502/503/504 as retriable here. If leaving them out is deliberate (to avoid retrying genuine server errors), please document that decision next to this check.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +456,55 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
+                return true;
+            }
+            if (current instanceof java.net.ConnectException) {
+                return true;
+            }
+            if (current instanceof java.net.SocketTimeoutException) {
+                return true;
+            }
+            if (current instanceof java.io.IOException io && io.getMessage() != null
```
paoloantinori: Two concerns in this classification. First, matching "Connection reset" by message text is fragile: the wording is JDK-specific and misses "broken pipe" and locale variants. Since `ConnectException` already extends `SocketException`, a single `current instanceof java.net.SocketException` check covers both connection-refused and connection-reset without string matching. Second, on the Vert.x HTTP client, timeouts surface as `VertxException` subclasses (for example `NoStackTraceTimeoutException`), not `java.net.SocketTimeoutException`, so the branch above likely never fires on the default transport. Worth verifying against a real outage trace and adding the Vert.x timeout type to the reflective lookup alongside `HttpClosedException`.

schema-resolver/src/main/java/io/apicurio/registry/resolver/config/SchemaResolverConfig.java
```diff
@@ -434,6 +464,44 @@ public Duration getRetryBackoff() {
...
+
+    public long getClientRetryMaxAttempts() {
+        return getLongNonNegative(CLIENT_RETRY_MAX_ATTEMPTS);
+    }
+
+    public long getClientRetryDelayMs() {
+        return getDurationNonNegativeMillis(CLIENT_RETRY_DELAY_MS).toMillis();
+    }
+
+    public double getClientRetryBackoffMultiplier() {
```
paoloantinori, line 552: This accepts NaN and negative values: `Double.parseDouble("NaN")` and `"-1"` both succeed, and a NaN multiplier drives every computed delay to 0 after the long cast, which silently removes backoff. Kafka Connect passes converter config as raw strings, so this is reachable from a misconfigured connector. Please validate the parsed value (`Double.isFinite(value) && value >= 1.0`, or `> 0` if sub-1.0 multipliers are intended) and throw `IllegalArgumentException` naming `CLIENT_RETRY_BACKOFF_MULTIPLIER`, consistent with the fail-fast behavior of `getLongNonNegative`.

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
```diff
@@ -101,9 +102,19 @@ private static RegistryClientOptions buildClientOptions(SchemaResolverConfig con
             throw new IllegalStateException(e);
         }
```
paoloantinori, line 107: Two things on this block. The `(int)` cast silently truncates if max-attempts is set above `Integer.MAX_VALUE`; `Math.toIntExact` fails fast instead and matches the fail-fast style used for the other new getters. Separately, and more important: the previous code called the no-arg `clientOptions.retry()` and inherited the SDK defaults, while this now always passes explicit values. Please confirm the `CLIENT_RETRY_*_DEFAULT` constants (3 / 250ms / 2.0 / 10000ms) match `RegistryClientOptions`' own defaults; if they differ, every existing SerDes and converter user gets a silent retry-timing change on upgrade, and the constants should be aligned (or read from the SDK) before merge.

paoloantinori (commented):
The direction is right and the motivation is real (short registry outages currently fail converter tasks hard), but I would not merge this as-is: the new failure classification ships with zero automated tests, and two parts of it likely will not fire on the Vert.x transport the schema resolver actually uses. Findings below, ordered by severity.

- **No tests anywhere in the PR.** `ERCache.isRetriableCacheLoadFailure` (schema-resolver/.../cache/ERCache.java:476) adds five match branches plus cause-chain walking, and `SchemaResolverConfig` (schema-resolver/.../config/SchemaResolverConfig.java:479) adds a hand-rolled double parser with three paths. None of it is covered. The test plan is four unchecked manual boxes. Per project norms this is an automatic re-request.
- **5xx is still fail-fast.** Only HTTP 429 counts as retriable among `ApiException`s (ERCache.java:479). A registry outage behind a load balancer, or a restart, typically surfaces as 502/503/504, so the exact outage this PR targets can still fail the cache load on the first attempt. If excluding 5xx is deliberate, it needs a comment; otherwise add 502/503/504.
- **Timeout branch likely dead on the default transport.** The Vert.x HTTP client reports timeouts as `VertxException` subclasses (for example `NoStackTraceTimeoutException`, message "The timeout was triggered"), not `java.net.SocketTimeoutException` (ERCache.java:485). The presence of the `HttpClosedException` lookup two branches below suggests Vert.x is the transport in practice, so the SocketTimeoutException check should be verified against a real outage trace.
- **Defaults parity is unverified.** The old code called the no-arg `clientOptions.retry()` and inherited the SDK defaults; the new code always passes explicit values (schema-resolver/.../client/RegistryClientFacadeFactory.java:111). If `CLIENT_RETRY_*_DEFAULT` (3 / 250ms / 2.0 / 10000ms) differs from `RegistryClientOptions`' own defaults, every existing SerDes and converter user gets a silent retry-timing change on upgrade. Please confirm they match or source the constants from the SDK.
- **Two retry layers now stack multiplicatively.** Each cache-level attempt runs a full HTTP retry ladder (250 + 500 + 1000ms at defaults), then ERCache sleeps its own backoff and starts another ladder (ERCache.java:446). Worst case per cache load at defaults is roughly 4 x 1.75s plus 3 x 300ms, about 8s of blocking on the caller thread. That is the intended behavior, but the two similar keys (`apicurio.registry.retry-count` vs `apicurio.registry.client.retry.max-attempts`) need a doc line each saying which layer they control, or operators will tune the wrong one.
- **Backoff multiplier accepts garbage.** `Double.parseDouble("NaN")` and `"-1"` both succeed (SchemaResolverConfig.java:479). A NaN multiplier collapses every delay to 0ms after the long cast, removing backoff entirely. Kafka Connect passes converter config as raw strings, so this is reachable. Validate `Double.isFinite(value) && value >= 1.0` (or `> 0`) and throw `IllegalArgumentException` naming the property, consistent with `getLongNonNegative`.
- Minor, easy wins: the Javadoc at ERCache.java:459 documents the predicate but is attached to the `VERTX_HTTP_CLOSED_EXCEPTION` field; the `(int)` cast at RegistryClientFacadeFactory.java:104 silently truncates, `Math.toIntExact` fails fast; the INFO log at RegistryClientFacadeFactory.java:108 prints the registry URL, which leaks credentials to logs in deployments that embed userinfo in the URL.

Mahamood-Sameer:
Thanks, addressed in the latest commit:

- Unit tests for isRetriableCacheLoadFailure and the new client-retry config getters/parsers
- 502/503/504 treated as retriable alongside 429 (404/500 remain fail-fast)
- SocketException instead of message matching; Vert.x timeout detection added next to HttpClosedException
- Backoff multiplier validated (Double.isFinite && > 1.0)
- Defaults documented against RegistryClientOptions#retry(); Math.toIntExact; registry URL sanitized in logs
- Javadoc on the two retry layers (retry-count vs client.retry.*)

Mahamood-Sameer:
The License Compliance check is failing, Not able to see the issues that had been found. Can you please check and let me know the issues so that i can fix them.

schema-resolver/src/test/java/io/apicurio/registry/resolver/cache/ERCacheRetriableFailureTest.java
```diff
@@ -0,0 +1,74 @@
+package io.apicurio.registry.resolver.cache;
```
paoloantinori, line 17: Both new test files (`ERCacheRetriableFailureTest.java` and `ClientRetryConfigTest.java`) start at `package` with no Apache license header. Every other source file in the module carries one, and the build's license/checkstyle check will fail on this. Add the standard header to both files.

paoloantinori (commented):
The two-layer retry design addresses a real gap and the new config keys fail fast on invalid values, but two correctness risks, an unenforced defaults-parity assumption, and missing license headers on the new test files should be resolved before merge.

- `schema-resolver/src/test/java/io/apicurio/registry/resolver/cache/ERCacheRetriableFailureTest.java:1` and `schema-resolver/src/test/java/io/apicurio/registry/resolver/config/ClientRetryConfigTest.java:1` both start at `package` with no Apache license header; every other file in the module carries one and the license/checkstyle check will fail.
- `ERCache.java:502` can NPE while classifying an exception: if `ApiException.getResponseStatusCode()` returns a null `Integer` (exception constructed without a response), auto-unboxing into `isRetriableHttpStatus(int)` throws inside the catch block and the NPE replaces the original failure as the error returned from `retry(...)`.
- `ERCache.java:464-513` uses ~35 lines of reflective `Class.forName` plus null-tolerant matching for Vert.x types, yet `RegistryClientFacadeFactory.java` already imports `io.vertx.core.Vertx` in main scope of the same module, and the new test imports `VertxException`/`HttpClosedException` directly. If Vert.x is genuinely absent for some deployment, that deployment cannot load `RegistryClientFacadeFactory` either, so the reflection buys nothing.
- `RegistryClientFacadeFactory.java:116` now passes literal values sourced from `SchemaResolverConfig` defaults instead of calling `RegistryClientOptions.retry()`. The comment (and `ClientRetryConfigTest.clientRetryDefaultsMatchRegistryClientOptionsRetry`) asserts parity with the SDK defaults but nothing enforces it; if the SDK defaults drift, the enabled-with-defaults path silently changes behavior. A test in `java-sdk/common` asserting `retry()` yields exactly 3/250/2.0/10000 would pin it.
- `ERCache.java:444` stacks the two ladders without an overall deadline. With the elevated settings from the test plan (`max-attempts=30`, `retry-count` default 3), one auto-register or resolve inside `KafkaProducer.send()` can block for (retries+1) x (30 attempts x connect timeout + backoff sleeps), easily minutes, which turns a short registry outage into `max.block.ms` expiry and record pile-up. Either cap total elapsed time per cache load or document the multiplication next to both config keys.
- Minor: `ERCache.java:524` matching all `SocketException`s also retries resource exhaustion ("Too many open files", "Network is unreachable"), so an fd leak now spins the full ladder instead of failing fast. Worth a comment at minimum. The message-text match in `isVertxTimeout` is also brittle across Vert.x versions, though it only risks false negatives.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +458,92 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
+     *   <li>Vert.x {@code VertxException} whose message or class name indicates a timeout
+     *       (for example {@code NoStackTraceTimeoutException})</li>
+     * </ul>
+     * Other {@link ApiException} status codes (for example 404/500) are not retried so
+     * genuine application errors fail fast.
+     */
+    static boolean isRetriableCacheLoadFailure(Throwable e) {
+        Throwable current = e;
+        while (current != null) {
+            if (current instanceof ApiException api && isRetriableHttpStatus(api.getResponseStatusCode())) {
```
paoloantinori: `api.getResponseStatusCode()` returns `Integer` in Kiota, and `isRetriableHttpStatus` takes `int`, so an `ApiException` constructed without a response status unboxes null and throws NPE here. Because this runs inside the catch block of `retry(...)`, the NPE escapes and replaces the original registry failure as the error callers see. Guard it: `Integer code = api.getResponseStatusCode(); if (code != null && isRetriableHttpStatus(code)) return true;`. Worth adding a test case with an `ApiException` that never gets `setResponseStatusCode` called.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +458,92 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
+            }
+            // Covers ConnectException and connection-reset SocketExceptions without
+            // matching JDK-specific message text.
+            if (current instanceof java.net.SocketException) {
+                return true;
+            }
+            if (current instanceof java.net.SocketTimeoutException) {
+                return true;
+            }
+            if (VERTX_HTTP_CLOSED_EXCEPTION != null && VERTX_HTTP_CLOSED_EXCEPTION.isInstance(current)) {
```
paoloantinori: This reflective lookup (~35 lines including the static initializer and null handling) guards against Vert.x being absent, but `RegistryClientFacadeFactory` in the same module imports `io.vertx.core.Vertx` in main scope, and this new test file imports `VertxException` and `HttpClosedException` directly. Any classpath that can load the resolver facade can load Vert.x core here too. Unless you can name a supported deployment where the module runs without Vert.x, replace the reflection with plain `instanceof HttpClosedException` plus a `VertxException` check and delete the static block.

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
```diff
@@ -101,9 +103,21 @@ private static RegistryClientOptions buildClientOptions(SchemaResolverConfig con
...
+        // (3 attempts, 250ms initial delay, multiplier 2.0, max delay 10000ms).
+        if (config.getClientRetryEnabled()) {
+            int maxAttempts = Math.toIntExact(config.getClientRetryMaxAttempts());
+            long delayMs = config.getClientRetryDelayMs();
+            double backoff = config.getClientRetryBackoffMultiplier();
+            long maxDelayMs = config.getClientRetryMaxDelayMs();
+            logger.log(Level.INFO,
+                    "Registry HTTP client retry enabled: url={0}, maxAttempts={1}, delayMs={2}, backoffMultiplier={3}, maxDelayMs={4}",
+                    new Object[]{sanitizeRegistryUrl(config.getRegistryUrl()), maxAttempts, delayMs, backoff, maxDelayMs});
+            clientOptions.retry(true, maxAttempts, delayMs, backoff, maxDelayMs);
```
paoloantinori, line 121: With retries enabled and no overrides, behavior now depends on the `SchemaResolverConfig` defaults (3/250/2.0/10000) matching `RegistryClientOptions.retry()`, and the only thing enforcing that is a comment plus a test on the config side that never reads the SDK class. If the SDK defaults ever change, this path silently diverges. Either expose the defaults as shared constants used by both, or add a test in `java-sdk/common` asserting `RegistryClientOptions.retry()` produces exactly maxAttempts=3, delayMs=250, multiplier=2.0, maxDelayMs=10000.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -438,15 +438,15 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
-                // if (i == retries || !(e.getCause() != null && e.getCause() instanceof ExecutionException
-                // && e.getCause().getCause() != null && e.getCause().getCause() instanceof ApiException
-                // && (((ApiException) e.getCause().getCause()).getResponseStatusCode() == 429)))
-                if (i == retries || !(e.getCause() != null && e.getCause() instanceof ApiException
-                        && (((ApiException) e.getCause()).getResponseStatusCode() == 429))) {
-                    log.error("Cache load failed after {} retries", i, e);
+                // Schema-cache retries compose with the HTTP client retry layer
+                // (apicurio.registry.client.retry.*). Each cache attempt may run a full
+                // client retry ladder before returning here.
+                if (i == retries || !isRetriableCacheLoadFailure(e)) {
```
paoloantinori: Now that this loop retries network and 502/503/504 failures, and each cache attempt internally runs a full HTTP client ladder, worst-case blocking is (retries+1) x (clientAttempts x connect timeout + client backoff sleeps) with no overall deadline. With the settings this PR's test plan recommends (`max-attempts` elevated, `retry-count` at default 3), a single auto-register during an outage can block a `KafkaProducer.send()` for minutes and trip `max.block.ms`, trading a fast failure for a stalled producer. Consider an overall elapsed-time cap per cache load, or at minimum document the multiplication next to both `apicurio.registry.retry-count` and `apicurio.registry.client.retry.max-attempts` so operators can size them together.

Mahamood-Sameer:
Thanks, latest feedback and Sonar findings are addressed on the branch.

Review items:

License headers on the new tests
Null-safe ApiException status handling (+ test)
Direct Vert.x instanceof instead of reflection
SDK retry() defaults pinned in RegistryClientOptionsRetryDefaultsTest
Documented multiplicative two-layer retry / max.block.ms sizing on both config keys

Sonar:

Conditional logging around sanitizeRegistryUrl
Rewrote the backoff comment flagged as commented-out code
Pattern-matching instanceof Vertx
assertThrows lambdas reduced to a single throwable invocation



Happy to adjust further if anything else stands out.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -438,15 +440,16 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
-                // && e.getCause().getCause() != null && e.getCause().getCause() instanceof ApiException
-                // && (((ApiException) e.getCause().getCause()).getResponseStatusCode() == 429)))
-                if (i == retries || !(e.getCause() != null && e.getCause() instanceof ApiException
-                        && (((ApiException) e.getCause()).getResponseStatusCode() == 429))) {
-                    log.error("Cache load failed after {} retries", i, e);
+                // Schema-cache retries compose with the HTTP client retry layer
+                // (apicurio.registry.client.retry.*). Each cache attempt may run a full
+                // client retry ladder before returning here; there is no overall deadline,
+                // so worst-case blocking is multiplicative (see SchemaResolverConfig docs).
+                if (i == retries || !isRetriableCacheLoadFailure(e)) {
```
paoloantinori: The two retry layers now multiply with no overall deadline, and this changes the default for every existing user. Before this PR, a connection-refused cache load ran one client ladder (roughly 2s, per the PR description) and failed; with the same stock config it now runs 4 cache attempts, each with a full client ladder (250+500+1000ms sleeps plus request time), plus 3 x 300ms backoff, so worst case is roughly 8-9s. Operators who raise `max-attempts` to survive outages get minutes of blocking inside `Producer.send()` / Connect poll with nothing capping it. The javadoc on `RETRY_COUNT` documents the multiplication but does not bound it. Two options: (1) give the cache load an overall wall-clock budget checked before each re-attempt, or (2) keep ERCache at "429 only" by default and gate the widened network/5xx retriable set behind explicit opt-in so existing default behavior is preserved. Either way, the new default blocking time deserves a line in the release notes.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +461,78 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
+            if (current instanceof ApiException api) {
+                Integer code = api.getResponseStatusCode();
+                if (code != null && isRetriableHttpStatus(code)) {
+                    return true;
+                }
+            }
+            // Covers ConnectException and connection-reset SocketExceptions without
+            // matching JDK-specific message text. Also matches unrelated SocketExceptions
+            // (fd exhaustion, unreachable network); see class javadoc above.
+            if (current instanceof java.net.SocketException) {
```
paoloantinori: Two issues with matching `SocketException` anywhere in the chain. Precedence: because the walk returns true on the first network exception at any depth, an `ApiException` with a definitive status (404, 500) that happens to sit above a `SocketException` in the chain is still retried. If an `ApiException` is present in the chain, its status should decide and network matching should be skipped. Breadth: "Too many open files" and "Network is unreachable" are `SocketException`s too, and retrying fd exhaustion through the full (now multiplicative) ladder while each attempt opens sockets makes the starvation worse; the javadoc already acknowledges this. I understand the class-level match is deliberate to avoid JDK-specific message text; if you keep it, consider excluding resource-exhaustion messages, or at minimum get explicit maintainer sign-off on that tradeoff.

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
```diff
@@ -101,9 +103,26 @@ private static RegistryClientOptions buildClientOptions(SchemaResolverConfig con
             throw new IllegalStateException(e);
         }
 
-        // FIXME push retry options into the SchemaResolverConfig
-        if (Boolean.TRUE) {
-            clientOptions.retry();
+        // Defaults for CLIENT_RETRY_* must match RegistryClientOptions#retry()
+        // (3 / 250ms / 2.0 / 10000ms); pinned by RegistryClientOptionsRetryDefaultsTest.
+        if (config.getClientRetryEnabled()) {
```
paoloantinori, line 110: `getLongNonNegative` accepts any non-negative long, so `apicurio.registry.client.retry.max-attempts=5000000000` passes config validation and then dies here as a bare `ArithmeticException` ("integer overflow") during client creation, with no mention of the property. Please fail fast in `SchemaResolverConfig.getClientRetryMaxAttempts()` instead: validate that the value fits an int (arguably also a sane upper bound) and throw `IllegalArgumentException` naming the key, mirroring what `getClientRetryBackoffMultiplier()` already does for invalid input.

schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java
```diff
@@ -167,6 +186,25 @@ private static RegistryClientOptions buildClientOptions(SchemaResolverConfig con
         return clientOptions;
     }
 
+    /**
+     * Strip userinfo from registry URLs before logging to avoid leaking credentials.
+     */
+    static String sanitizeRegistryUrl(String url) {
```
paoloantinori, line 195: This helper guards credentials in logs and has no tests. Please add cases for: URL with userinfo (stripped), the same URL without userinfo (passed through unchanged), malformed URL, null, and blank. One behavior question: on any parse failure the method returns `<redacted>`, which reads as "there were credentials here", so a malformed URL without credentials produces the same marker. A constant that names the reason without echoing the input (for example `<invalid-url>`) keeps the safety and makes the two cases distinguishable in logs. Do not return the raw string on failure: a URL too malformed for `URI.create` can still contain a password.

schema-resolver/src/test/java/io/apicurio/registry/resolver/cache/ERCacheRetriableFailureTest.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.IOException;
+import java.net.ConnectException;
+import java.net.SocketException;
+import java.net.SocketTimeoutException;
+import java.util.concurrent.ExecutionException;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class ERCacheRetriableFailureTest {
```
paoloantinori, line 38: These tests pin only the predicate; nothing exercises the changed control flow in `ERCache.retry()` itself (the `isRetriableCacheLoadFailure` wiring, the `i == retries` boundary, the attempt count), so inverting the condition at the call site would keep the suite green. Please add a test that drives the retry loop with a controllably-failing loader and asserts exact counts: a retriable cause (`RuntimeException` wrapping `ConnectException`) yields `retries + 1` invocations and `Result.error`, and a non-retriable cause (`ApiException` 404) yields exactly 1 invocation. If `retry()` is hard to reach because it is private, relax it to package-private the same way `isRetriableCacheLoadFailure` already is.

paoloantinori (commented):
Solid direction and unusually good config documentation (the layer-multiplication javadoc and the fail-fast backoff validation are the right instincts), but I would not merge yet: the two retry layers now multiply with no overall deadline, which silently changes default blocking for every existing user, the retriable-exception match is broader than the outage case it targets, and none of the new control flow in ERCache.retry() or sanitizeRegistryUrl() is under test.

- `schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java:447` and `schema-resolver/src/main/java/io/apicurio/registry/resolver/config/SchemaResolverConfig.java:203`: worst-case cache-load blocking goes from roughly 2s to roughly 8-9s on stock defaults (4 cache attempts, each running a full client ladder of 250+500+1000ms sleeps plus request time, plus 3 x 300ms backoff). Nothing caps the total, and operators raising `max-attempts` to survive outages get minutes of blocking in `Producer.send()` / Connect poll. Needs an overall deadline or an opt-in for the widened retriable set, plus a release-notes line for the new default.
- `ERCache.java:496`: any `SocketException` anywhere in the chain is retriable, including fd exhaustion ("Too many open files") and "Network is unreachable", and a network exception deeper in the chain can outvote a definitive `ApiException` status (404/500) higher up.
- `ERCache.java:486`: the cause walk (and `rootMessage` at `ERCache.java:528`) has no depth cap; a cyclic cause chain (two throwables wired via `initCause`) loops forever. Cap at 16-64 levels, as Guava's Throwables does.
- `schema-resolver/src/main/java/io/apicurio/registry/resolver/client/RegistryClientFacadeFactory.java:108`: `max-attempts` above Integer.MAX_VALUE passes config validation and then explodes as a bare `ArithmeticException` here with no property name; validate in the getter instead.
- `RegistryClientFacadeFactory.java:192`: `sanitizeRegistryUrl` guards credentials in logs and has zero tests, and any parse failure logs as `<redacted>` even when no credentials were present.
- `schema-resolver/src/test/java/io/apicurio/registry/resolver/cache/ERCacheRetriableFailureTest.java:33`: coverage is predicate-only; inverting the condition at the `retry()` call site would keep the suite green.
- `schema-resolver/src/main/java/io/apicurio/registry/resolver/config/SchemaResolverConfig.java:507`: roughly 40 lines of hand-rolled double coercion; if the config base has a typed numeric accessor or shared coercion helper (the `getLongNonNegative` / `getDurationNonNegativeMillis` pattern suggests one exists), reuse or extract rather than duplicating the parse-and-throw shape.
- `SchemaResolverConfig.java:527`: exactly 1.0 is rejected as a backoff multiplier; worth confirming `RegistryClientOptions#retry` really rejects 1.0 too, otherwise this is stricter than the layer it configures and constant-delay backoff becomes unconfigurable.
- `java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java:86`: this FINE line duplicates the INFO line emitted at client creation in `RegistryClientFacadeFactory` (same five values, two log frameworks); consider keeping only one. Also `cause.getMessage()` can be null, so the exhausted-retry WARNING at `RegistryClientRequestAdapterFactory.java:232` can print a bare `null`.
- `SchemaResolverConfig.java:630`: the `getVertx()` instanceof-pattern rewrite is unrelated to this PR's purpose; harmless, but splitting it out keeps the diff focused.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -458,6 +471,116 @@ private static <T> Result<T, RuntimeException> retry(Duration backoff, long retr
...
+        }
+        return false;
+    }
+
+    private static boolean isNonRetriableSocketException(java.net.SocketException exception) {
+        if (exception instanceof java.net.NoRouteToHostException) {
+            return true;
+        }
+        String message = exception.getMessage();
+        if (message == null) {
```
paoloantinori, line 633: `NoRouteToHostException` is classified non-retriable, but in Kubernetes a down/rolling service frequently surfaces as exactly this (stale endpoint before kube-proxy updates). That contradicts the PR goal of riding out registry outages: `ConnectException` (refused) retries while no-route fails immediately. Suggest retrying it under the `transient-errors` opt-in and keeping only the resource-exhaustion messages ("too many open files", "no buffer space") non-retriable.

schema-resolver/src/main/java/io/apicurio/registry/resolver/config/SchemaResolverConfig.java
```diff
@@ -434,6 +504,49 @@ public Duration getRetryBackoff() {
...
+                    + "'. Expected an integer in [1, " + Integer.MAX_VALUE + "], but got '" + value + "'.");
+        }
+        return value;
+    }
+
+    public long getClientRetryDelayMs() {
+        return getDurationNonNegativeMillis(CLIENT_RETRY_DELAY_MS).toMillis();
+    }
+
+    public double getClientRetryBackoffMultiplier() {
```
paoloantinori, line 552: When `transient-errors=true` there is no overall deadline: worst case is roughly `(retry-count + 1) x (client ladder: max-attempts with backoff up to max-delay-ms)` plus cache backoffs, all on the Kafka send/poll thread. The Javadoc documents the multiplication but offers no way to bound it. Consider a total-budget property (e.g. `apicurio.registry.retry.total-timeout-ms`) checked inside `ERCache.retry`, or at minimum a startup WARNING when the computed worst case exceeds a few seconds.

schema-resolver/src/main/java/io/apicurio/registry/resolver/config/AbstractConfig.java
```diff
@@ -140,4 +142,38 @@ private void reportError(String key, String expectedText, Object value) {
...
+    }
+
+    /**
+     * Ensure a non-negative long fits in a Java {@code int} (for APIs that take int).
+     */
+    protected int getIntNonNegative(String key) {
+        long value = getLongNonNegative(key);
+        if (value > Integer.MAX_VALUE) {
+            reportError(key, "an integer-compatible value (<= " + Integer.MAX_VALUE + ")", value);
+        }
```
paoloantinori: `getIntNonNegative` is unused: `RegistryClientFacadeFactory` still calls `Math.toIntExact(...)` and `getClientRetryMaxAttempts()` already enforces the [1, Integer.MAX_VALUE] range itself. Either use it in the factory (`config.getClientRetryMaxAttempts()` returning int via this helper) or remove it to avoid two divergent validation paths for the same constraint.

java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
```diff
@@ -212,8 +221,10 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                         }
                     } else {
                         if (isRetryable(cause) && attempt >= maxRetryAttempts) {
-                            log.log(Level.WARNING, "Maximum retry attempts ({0}) exceeded for {1}: {2}",
-                                    new Object[]{maxRetryAttempts, cause.getClass().getName(), cause.getMessage()});
+                            String causeMessage = cause.getMessage() != null ? cause.getMessage() : "<no message>";
+                            log.log(Level.WARNING,
+                                    "Registry HTTP client exhausted {0} retry attempts due to {1}: {2}",
+                                    new Object[]{maxRetryAttempts, cause.getClass().getName(), causeMessage});
                         }
```
paoloantinori, line 228: On exhaustion this throws `originalCause` (the first attempt's failure), so the most recent and usually most relevant exception is only visible in the WARNING log, not in the propagated stack trace callers see. Prefer throwing the latest `cause` (or `originalCause.addSuppressed(cause)`) so downstream diagnostics reflect the final state of the registry rather than the state from before the retries.

paoloantinori (commented):
Solid, well-documented change with good test coverage and a sensible opt-in (`transient-errors=false`) that preserves historical behavior. The main concerns are: no overall deadline on the multiplicative retry stack, inconsistent classification of unreachable-host failures versus the stated outage-retry motivation, dead/duplicated validation code in `AbstractConfig`, and the SDK retry proxy discarding later-attempt errors when retries are exhausted.
- `schema-resolver/.../ERCache.java` (retry loop, ~line 433): with `transient-errors=true`, worst-case blocking is `(retry-count + 1) x (client ladder sleeps)` plus cache backoffs with no cap. The docs call this out, but an operator setting only `retry-count=10` gets a surprise. Consider a total-budget option (for example `apicurio.registry.retry.max-total-ms`) or at least a WARNING log when the effective product exceeds some threshold.
- `ERCache.isNonRetriableSocketException`: `NoRouteToHostException` is treated as non-retriable, but in Kubernetes/cloud environments a briefly down service or stale endpoint often surfaces exactly as "No route to host". This contradicts the PR's outage-retry motivation; the registry being down is precisely when this path fires. I would retry it under the opt-in and keep only resource-exhaustion signals (`too many open files`, `no buffer space`) fatal.
- `isVertxTimeout` relies on lowercase message matching ("timeout") which the Javadoc itself admits is brittle across Vert.x versions; a `ConnectTimeoutException`/`NoStackTraceTimeoutException` class-name check would be more stable than string matching on messages that Vert.x may localize or reword.
- `AbstractConfig.getIntNonNegative` (new, lines ~172-180) is dead code: `RegistryClientFacadeFactory` uses `Math.toIntExact(config.getClientRetryMaxAttempts())` and `getClientRetryMaxAttempts()` already range-validates in [1, Integer.MAX_VALUE]. Either use it in the factory or drop it.
- `java-sdk/common/.../RegistryClientRequestAdapterFactory.java` (~line 221): when attempts are exhausted the code throws `originalCause` from the first attempt; the last failure (often the more informative one, e.g. a different host in a pooled setup) is lost except for the one WARNING line. Consider throwing the latest cause, or chaining them.
- `SchemaResolverConfig.getClientRetryBackoffMultiplier`: a `Duration`-typed value is accepted by `getDurationNonNegativeMillis` keys elsewhere, but this key only accepts Number/String; docs say so, just be aware converters that forward Kafka `Duration` values will fail here. Fine as-is given documentation.

Mahamood-Sameer:
Thanks, addressed in the latest commits:

- Added apicurio.registry.retry.total-timeout-ms to bound one cache-load retry loop, plus a WARNING when transient-errors=true with no budget and a high estimated worst case
- NoRouteToHostException is retriable under the opt-in; only resource-exhaustion socket messages stay fatal
- Vert.x timeout detection prefers class-name matching (*timeout*), with message text as fallback
- Removed unused getIntNonNegative
- On HTTP retry exhaustion, the latest cause is thrown (first attempt suppressed)
- Cleared the remaining Sonar items (double cast in the backoff estimate; lowered ERCache.retry cognitive complexity via helpers)


Happy to tweak further if anything else stands out.

paoloantinori:
Re-read the new head and all six items check out:

- `apicurio.registry.retry.total-timeout-ms` bounds one cache-load retry loop, checked before each re-attempt and after backoff, with `Duration.ZERO` documented as unlimited
- the unbounded-budget WARNING fires when `transient-errors=true` with no budget and an estimated sleep-only worst case above 5s
- `NoRouteToHostException` is covered by the `SocketException` branch minus the resource-exhaustion messages, which stays fatal
- Vert.x timeout detection prefers class-name matching with message text as fallback
- retry exhaustion now throws the latest cause with the first suppressed, pinned by `testExhaustedRetriesThrowsLatestException`
- the Sonar items are cleared (single cast in the backoff estimate, `ERCache.retry` decomposed into `handleRetryableFailure` / `sleepBackoffOrTimeout`)

The defaults-parity pin and the sanitize-URL helper from the earlier rounds are still in place. No further concerns from me — once this is rebased on main I have nothing else to add.

Mahamood-Sameer:
Rebased onto latest main, ready for another look whenever you are.

schema-resolver/src/main/java/io/apicurio/registry/resolver/AbstractSchemaResolver.java
```diff
@@ -291,4 +294,26 @@ protected SchemaLookupResult<S> loadFromVersionCoordinates(RegistryVersionCoordi
...
+            return 0L;
+        }
+        long maxAttempts = config.getClientRetryMaxAttempts();
+        if (maxAttempts <= 1) {
+            return 0L;
+        }
+        long delayMs = config.getClientRetryDelayMs();
+        double multiplier = config.getClientRetryBackoffMultiplier();
+        long maxDelayMs = config.getClientRetryMaxDelayMs();
+        long sleepMs = 0L;
```
paoloantinori, line 318: This loop runs `maxAttempts - 1` iterations, and `CLIENT_RETRY_MAX_ATTEMPTS` validates only `[1, Integer.MAX_VALUE]`. A value like `2000000000` passes config validation and then blocks `configure()` for minutes on ~2B `Math.pow` calls, and `sleepMs` can overflow `long` to negative. Since `step` is clamped to `maxDelayMs`, the sum becomes arithmetic once the clamp engages: compute it in closed form, or break out of the loop once `step >= maxDelayMs` and add the remaining `(maxAttempts - attempt) * maxDelayMs` directly. Also consider an upper sanity bound on `maxAttempts` at config parse time.

java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java
```diff
@@ -212,10 +221,16 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
-                        throw originalCause;
+                        // Prefer the latest failure for caller diagnostics; keep the first as suppressed.
+                        if (originalCause != null && originalCause != cause) {
```
paoloantinori, line 232: Switching from `throw originalCause` to `throw cause` changes the exception type callers see whenever attempt 1 and attempt N fail differently (e.g. first `HttpClosedException`, last `ConnectException`). The suppressed-exception handling is right, but this is a public-behavior change of the SDK adapter, not just diagnostics. Please verify no downstream catch sites (in this repo or documented user code) match on the original exception type, and call the change out explicitly in the PR description so the SDK change is reviewed as such rather than as a logging tweak.

schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java
```diff
@@ -421,43 +470,205 @@ private ExecutorService getOrCreateRefreshExecutor() {
...
+    private static boolean sleepBackoffOrTimeout(Duration backoff, Instant deadline, Duration totalTimeout) {
+        try {
+            Thread.sleep(backoff.toMillis());
+        } catch (InterruptedException e) {
+            log.debug("Cache retry backoff interrupted", e);
+            Thread.currentThread().interrupt();
+        }
+        if (isDeadlineExceeded(deadline)) {
+            log.error("Schema cache load exceeded total retry timeout of {}ms during backoff",
+                    totalTimeout.toMillis());
```
paoloantinori: This timeout-exceeded result drops the underlying exception; the operator sees only the synthetic message with no cause, unlike the pre-backoff path in `handleRetryableFailure` which attaches `e`. Pass the last failure into `sleepBackoffOrTimeout` (or return it) so the timeout error carries the real cause, otherwise the most diagnostic exception of the outage is discarded exactly when things are worst.

docs/modules/ROOT/pages/getting-started/assembly-configuring-kafka-client-serdes.adoc
```diff
@@ -308,9 +308,44 @@ The `DefaultSchemaResolver` uses the following properties to configure how to lo
 |`300`
 |`RETRY_COUNT`
 |`apicurio.registry.retry-count`
-|Used by serializers and deserializers. If the SerDes cannot retrieve a schema from the registry, they can retry a number of times. This configuration option controls the number of retry attempts.
+|Used by serializers and deserializers. Controls how many times the schema cache retries a failed load before failing. Distinct from `apicurio.registry.client.retry.max-attempts` (HTTP client retry ladder). By default the cache only retries HTTP 429; set `apicurio.registry.retry.transient-errors=true` to also retry network/outage failures. When that opt-in is enabled, worst-case blocking multiplies across both layers — size together with the client retry keys, especially for Kafka producers (`max.block.ms`) and Connect polls.
 |`non-negative Number, or integer String`
 |`3`
```
paoloantinori, line 313: The PR summary reads as if this PR makes `apicurio.registry.retry-count` cover registry outages; with `retry.transient-errors` defaulting to `false` it does not, and operators coming from the motivation section will set `retry-count` alone and see no change. Consider making this row (and the PR description) state up front that outage retries require opting in via `apicurio.registry.retry.transient-errors=true` plus `retry.total-timeout-ms`, so the activation path is unmissable.

paoloantinori (commented):
Solid, well-documented change with good test coverage of the retry classification logic, but I'm holding it for two issues: the ladder-estimation loop can hang `configure()` on large `max-attempts` values, and the transient-retry behavior the PR is motivated by is off by default, which contradicts the PR summary. Findings, hardest first:

- `AbstractSchemaResolver.estimateClientRetryLadderSleepMs` (schema-resolver/src/main/java/io/apicurio/registry/resolver/AbstractSchemaResolver.java:312) iterates `maxAttempts - 1` times and `CLIENT_RETRY_MAX_ATTEMPTS` is validated only as `[1, Integer.MAX_VALUE]`. A fat-fingered value like `2000000000` passes validation, then this loop runs ~2 billion `Math.pow` iterations inside `configure()` on every serializer/deserializer construction, and `sleepMs` can overflow to a negative value, which then propagates into `ERCache.warnIfRetryBudgetUnbounded`'s `Math.max(0L, ...)` (that part is guarded) but the loop itself is still a startup hang. Cap the iteration count (the delay is clamped by `maxDelayMs`, so the sum converges once `step >= maxDelayMs`; you can compute it in closed form or bail out after the clamp kicks in) and validate `maxAttempts` against a sane ceiling.
- The PR summary says this PR "extends ERCache retries beyond HTTP 429 to transient network / registry-down failures", but the implementation gates that behind `apicurio.registry.retry.transient-errors`, default `false` (schema-resolver/src/main/java/io/apicurio/registry/resolver/config/SchemaResolverConfig.java:250). That default is the right call for backward compatibility, but it means the stated motivation (Connect/Debezium converters failing on short outages) is not fixed by this PR unless operators also discover and set the new flag. The PR description and the `RETRY_COUNT` doc row (docs/modules/ROOT/pages/getting-started/assembly-configuring-kafka-client-serdes.adoc:313) should lead with "set `apicurio.registry.retry.transient-errors=true`" as the activation step; today the summary reads as if `retry-count` alone now covers outages.
- `RetryInvocationHandler` now throws the latest `cause` instead of `originalCause` (java-sdk/common/src/main/java/io/apicurio/registry/client/common/RegistryClientRequestAdapterFactory.java:230). This is a real behavioral change beyond logging: the two attempts can fail with different exception types (first `HttpClosedException`, last `ConnectException`), so code that catches or pattern-matches on the original failure type will see a different class. The suppressed-exception treatment is good, but please grep downstream consumers of the thrown exception (including the kiota adapter call sites and any user-facing docs showing catch blocks) to confirm nothing depends on the original exception, and note the change in the PR description.
- `ERCache.sleepBackoffOrTimeout` (schema-resolver/src/main/java/io/apicurio/registry/resolver/cache/ERCache.java:545) returns a timeout error that drops the underlying exception entirely; the caller gets only the synthetic "exceeded total retry timeout" message with no cause. The pre-backoff path in `handleRetryableFailure` correctly attaches `e`; do the same here (capture the last failure and pass it in) so operators debugging an outage see what actually killed the attempts.
- `RETRY_TOTAL_TIMEOUT_MS` is checked only between attempts; a single cache attempt can itself run a full client retry ladder plus request timeouts that blow past the deadline mid-attempt. The javadoc says "checked before each re-attempt" so this is documented, but the doc table (assembly-configuring-kafka-client-serdes.adoc:320) sells it as a wall-clock budget for the loop, which it only approximately is. One sentence in the docs noting the budget does not interrupt an in-flight attempt would prevent operators from sizing `max.block.ms` against it exactly.

Mahamood-Sameer:
Thanks for the hold, all of this is in the latest commits.

Ladder estimate hang: estimateClientRetryLadderSleepMs now short-circuits once the delay hits maxDelayMs and adds the rest in one multiply (saturating). apicurio.registry.client.retry.max-attempts is validated as [1, 1000] so a value like 2000000000 is rejected at config parse instead of looping in configure().

Opt-in vs PR motivation: default retry.transient-errors=false is unchanged for compatibility. Docs (and the RETRY_COUNT row) now lead with the activation step: set apicurio.registry.retry.transient-errors=true and preferably retry.total-timeout-ms. Raising retry-count alone still only retries HTTP 429.

Throw latest vs original: grepped this repo — SerDes catch RuntimeException / ApiException, nothing matches on the first attempt’s type. Kept throw-latest with the first failure on getSuppressed(). Added a test where attempt 1 is HttpClosedException and the last is ConnectException. Called out as an SDK behavior change in the PR description.

Timeout dropped the cause: the “exceeded total retry timeout” error now wraps the last attempt’s exception, same as the pre-backoff path.

Timeout is not a hard wall-clock: docs now say the budget is checked before each re-attempt and does not interrupt an in-flight cache attempt (that attempt can still run a full client retry ladder).

Sonar on the leftover (double) maxDelayMs is also gone (redundant promotion).

paoloantinori:
Re-read the new head — all five items check out:
- Ladder estimate now short-circuits once the step clamps to `maxDelayMs`, with saturating add/mul, and `max-attempts` is validated in [1, 1000] at config parse, so the `configure()` hang is gone.
- Docs lead with the `transient-errors` activation step on both the `RETRY_COUNT` row and the new keys.
- Throw-latest keeps the first failure on `getSuppressed()`, and the mixed `HttpClosedException`/`ConnectException` test pins it; good call flagging the SDK behavior change in the description.
- Both timeout paths (after an attempt and during backoff) now wrap the last failure as the cause.
- The `total-timeout-ms` row states the budget is checked between attempts and does not interrupt an in-flight one.

No further concerns from me — this is ready from my side.

Mahamood-Sameer:
Rebased on to main. please check once.

paoloantinori:
Re-read after the rebase onto main (1bf4b016). Everything previously verified is still intact: the `total-timeout-ms` budget in the cache retry loop, the unbounded-budget warning, the [1, 1000] max-attempts ceiling with fail-fast, the ladder-estimate clamp short-circuit with saturating math, the latest-cause-thrown behavior with the first failure attached as suppressed, null-safe ApiException status handling, the Vert.x class-name matching, URL userinfo sanitization, and the full failure-classification and defaults-parity test suites. I don't see anything the rebase broke or introduced. No further concerns from my side — thanks for keeping it clean through all the rounds.

Mahamood-Sameer:
Cool, How can we proceed further?

Mahamood-Sameer:
Hi @paoloantinori ,

I haven't got any update yet. Could you please update me here, if anything needs to be fixed?

Mahamood-Sameer:
Hi,

Any update on the PR?

Mahamood-Sameer:
Thanks for the update !

Issue #9232, opened 2026-08-04
Role mapping action failures replace the entire Access page with an error screen


bhuvan-somisetty:
## Summary

When granting, revoking, or updating a role mapping on the Access page fails (for example due to a transient network failure or a server-side error), the application replaces the entire Access page with a generic full-screen error page.

This causes the user to lose all in-page context (navigation, breadcrumbs, and the role mappings list) even though the failure is limited to a single role-mapping operation.

Instead of surfacing a scoped error for the failed action, the entire page becomes unusable until it is reloaded.

---

## Steps to reproduce

1. Start Apicurio Registry with auth/RBAC enabled
2. Open the Access (Roles) page as an admin
3. Trigger a **Grant access**, **Revoke access**, or **Edit** (update role) action
4. Before the request completes, force the corresponding API request to fail (for example by returning HTTP 500 or simulating a network failure)
5. Wait for the request to complete

---

## Actual behaviour

After the failed request:

- the entire page is replaced by the generic error screen
- navigation and breadcrumbs disappear
- the role mappings list is no longer visible
- the "please wait" modal state is never reset, since the `catch` blocks in `updateRoleMapping`, `createRoleMapping`, and `onRevokeRoleMapping` only call `setPageError` and never call `pleaseWait(false, "")`

The failure is not scoped to the affected role-mapping action.

---

## Expected behaviour

A failed role-mapping mutation should only affect the operation being performed.

The application should:

- keep the current page visible
- preserve navigation, breadcrumbs, and the role mappings list
- surface an inline or localised error indicating that the action failed
- reset the "please wait" modal state on failure, not just on success
- leave the user on the current page so they can retry the action

---

## Additional observations

This is the same defect class as #8984 ("Rule action failures replace the entire page with an error screen"), which was fixed for the Global/Group/Artifact Rules pages in #9013. That fix was never applied to `RolesPage.tsx` (`ui/ui-app/src/app/pages/roles/RolesPage.tsx`), which still calls `setPageError(...)` directly in the `catch` branches of `updateRoleMapping`, `createRoleMapping`, and `onRevokeRoleMapping`, so `PageErrorHandler` unconditionally swaps the whole page for `ErrorPage`/`AccessErrorPage` on any failure.

Because this is an access-management page, losing the full role list and page context on a single failed action is disruptive, especially for admins managing many role mappings.

bhuvan-somisetty:
[...]

PR #9583, opened 2026-08-14
fix(ui): reset PleaseWait modal when delete/action requests fail

### LINKED ISSUE #9408: [Bug]: PleaseWait modal stays stuck after failed delete/action on explore pages
## Description

**Persistence type**: in-memory / sql (UI bug; storage-independent)

When a delete or other mutating action fails on several Registry UI pages, the blocking **Please wait** modal stays open indefinitely. The user cannot interact with the page again until a full browser refresh, even though the underlying page error may also be shown.

Related but different:
- #8984 / PR #9013 fixed the same *page-replacement* error pattern on Rules pages.
- #9232 covers the Access (Roles) page specifically (please-wait + full-page error there).

This issue tracks the remaining explore/artifact/version/group/branch pages where `pleaseWait(true, ...)` is called before an async mutation but `pleaseWait(false)` is missing from the failure path.

### Environment

Apicurio Registry UI (`ui/ui-app`).

### Steps to Reproduce

1. Open any affected page (examples below).
2. Trigger a delete or similar action that shows the Please wait modal.
3. Force the API call to fail (permissions error, network failure, HTTP 500, etc.).
4. Observe the UI after the request completes.

**Example — delete artifact on Artifact page:**
1. Navigate to an artifact details page.
2. Delete the artifact and confirm.
3. Simulate/encounter a failed `DELETE` request.

**Example — delete group on Explore page:**
1. Open Explore and delete a group.
2. Force the delete request to fail.

**Expected:** On failure, the Please wait modal closes, the user remains on the current page, and any error feedback is shown without blocking further interaction.

**Actual:** Several handlers only call `pleaseWait(false)` in the success `.then()` branch. In `.catch()`, they call `setPageError(...)` but never reset the modal, so the spinner overlay remains stuck on screen.

### Root cause (code)

Multiple page components follow this pattern:

```tsx
pleaseWait(true, "Deleting ..., please wait...");
service.deleteSomething(...).then(() => {
    pleaseWait(false);
    ...
}).catch(error => {
    setPageError(toPageError(error, "..."));
    // missing: pleaseWait(false)
});
```

**Affected handlers (as of main):**

| Page | Handler | Operation |
|------|---------|-----------|
| `ArtifactPage.tsx` | `doDeleteArtifact` | Delete artifact |
| `ArtifactPage.tsx` | `doDeleteVersion` | Delete version |
| `ArtifactPage.tsx` | `doDeleteBranch` | Delete branch |
| `VersionPage.tsx` | `doDeleteVersion` | Delete version |
| `ExplorePage.tsx` | `doDeleteGroup` | Delete group |
| `GroupPage.tsx` | `doDeleteGroup` | Delete group |
| `GroupPage.tsx` | `doDeleteArtifact` | Delete artifact (also had **no `.catch()` at all**) |
| `BranchPage.tsx` | `doDeleteBranch` | Delete branch |

Other mutations on these pages (create version/branch, change state, etc.) already reset the modal correctly in their error paths.

### Logs

N/A — UI state bug; no backend error required beyond a failed mutation response.

Mohammad-Raees:
## Summary
- Fixes #9408: failed delete/action requests no longer leave the blocking **Please wait** modal stuck open.
- Adds shared `handleMutatingActionError` next to `toPageError` and uses it on mutating-action failure paths so clearing the wait overlay and setting the page error happen in one place.
- Sweep of `pleaseWait(true` in ui-app also fixed RolesPage catch handlers and Explore ZIP import (`ProgressModal` / `setImporting(false)` on failure).
- Corrects BranchPage delete error copy from "Error deleting a version." to "Error deleting a branch."
- Helper unit tests added. Page-level harness/regression test tracked in #9584 (manual checklist left unchecked until exercised).

## Test plan
- [ ] On Artifact page, trigger a failed delete artifact/version/branch and confirm Please wait closes and the page remains usable
- [ ] On Version page, trigger a failed delete version and confirm the modal closes
- [ ] On Explore / Group page, trigger a failed delete group (and artifact delete on Group) and confirm the modal closes
- [ ] On Branch page, trigger a failed delete branch and confirm the modal closes
- [ ] On Explore, force a failed ZIP import and confirm the import progress modal closes
- [ ] On Roles/Access page, force a failed grant/update/revoke and confirm Please wait closes
- [ ] Confirm successful deletes still close the modal and navigate as before
- [x] `npx vitest run src/app/pages/handleMutatingActionError.test.ts` in `ui/ui-app`

### DIFF
--- a/ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx
+++ b/ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx
@@ -3,7 +3,7 @@ import "./ArtifactPage.css";
 import { LoaderGuard, newLoaderGuard } from "@utils/loader.utils.ts";
 import { Breadcrumb, BreadcrumbItem, PageSection, Tab, Tabs } from "@patternfly/react-core";
 import { Link, useMatch, useParams } from "react-router";
-import { EXPLORE_PAGE_IDX, PageDataLoader, PageError, PageErrorHandler, PageProperties, toPageError } from "@app/pages";
+import { EXPLORE_PAGE_IDX, handleMutatingActionError, PageDataLoader, PageError, PageErrorHandler, PageProperties, toPageError } from "@app/pages";
 import {
     ChangeOwnerModal,
     ConfirmDeleteModal,
@@ -192,7 +192,7 @@ export const ArtifactPage: FunctionComponent<PageProperties> = () => {
             pleaseWait(false, "");
             appNavigation.navigateTo("/explore");
         }).catch(error => {
-            setPageError(toPageError(error, "Error deleting an artifact."));
+            handleMutatingActionError(error, "Error deleting an artifact.", setPageError, () => pleaseWait(false));
         });
     };
 
@@ -262,7 +262,7 @@ export const ArtifactPage: FunctionComponent<PageProperties> = () => {
                 versionDeleteSuccessCallback();
             }
         }).catch(error => {
-            setPageError(toPageError(error, "Error deleting a version."));
+            handleMutatingActionError(error, "Error deleting a version.", setPageError, () => pleaseWait(false));
         });
     };
 
@@ -287,8 +287,7 @@ export const ArtifactPage: FunctionComponent<PageProperties> = () => {
             const bid: string = encodeURIComponent(branchId);
             appNavigation.navigateTo(`/explore/${gid}/${aid}/branches/${bid}/versions`);
         }).catch(error => {
-            pleaseWait(false);
-            setPageError(toPageError(error, "Error adding a version to a branch."));
+            handleMutatingActionError(error, "Error adding a version to a branch.", setPageError, () => pleaseWait(false));
         });
     };
 
@@ -321,7 +320,7 @@ export const ArtifactPage: FunctionComponent<PageProperties> = () => {
                 branchDeleteSuccessCallback();
             }
         }).catch(error => {
-            setPageError(toPageError(error, "Error deleting a branch."));
+            handleMutatingActionError(error, "Error deleting a branch.", setPageError, () => pleaseWait(false));
         });
     };
 
--- a/ui/ui-app/src/app/pages/branch/BranchPage.tsx
+++ b/ui/ui-app/src/app/pages/branch/BranchPage.tsx
@@ -5,6 +5,7 @@ import { Breadcrumb, BreadcrumbItem, PageSection,  } from "@patternfly/react-cor
 import { Link, useParams } from "react-router";
 import {
     EXPLORE_PAGE_IDX,
+    handleMutatingActionError,
     PageDataLoader,
     PageError,
     PageErrorHandler,
@@ -86,7 +87,7 @@ export const BranchPage: FunctionComponent<PageProperties> = () => {
             const aid: string = encodeURIComponent(artifactId as string);
             appNavigation.navigateTo(`/explore/${gid}/${aid}/branches`);
         }).catch(error => {
-            setPageError(toPageError(error, "Error deleting a version."));
+            handleMutatingActionError(error, "Error deleting a branch.", setPageError, () => pleaseWait(false));
         });
     };
 
--- a/ui/ui-app/src/app/pages/explore/ExplorePage.tsx
+++ b/ui/ui-app/src/app/pages/explore/ExplorePage.tsx
@@ -7,6 +7,7 @@ import {
     ExplorePageEmptyState,
     ExplorePageToolbar,
     ExplorePageToolbarFilterCriteria, ImportModal,
+    handleMutatingActionError,
     PageDataLoader,
     PageError,
     PageErrorHandler,
@@ -230,7 +231,7 @@ export const ExplorePage: FunctionComponent<ExplorePageProps> = () => {
                     search(criteria, resetPaging);
                 }, 1500);
             }).catch(error => {
-                setPageError(toPageError(error, "Error importing multiple artifacts"));
+                handleMutatingActionError(error, "Error importing multiple artifacts", setPageError, () => setImporting(false));
             });
         }
     };
@@ -259,7 +260,7 @@ export const ExplorePage: FunctionComponent<ExplorePageProps> = () => {
             setPaging(resetPaging);
             search(criteria, resetPaging);
         }).catch(error => {
-            setPageError(toPageError(error, "Error deleting group."));
+            handleMutatingActionError(error, "Error deleting group.", setPageError, () => pleaseWait(false));
         });
     };
 
--- a/ui/ui-app/src/app/pages/group/GroupPage.tsx
+++ b/ui/ui-app/src/app/pages/group/GroupPage.tsx
@@ -7,6 +7,7 @@ import {
     EXPLORE_PAGE_IDX,
     GroupOverviewTabContent,
     GroupPageHeader, GroupRulesTabContent,
+    handleMutatingActionError,
     PageDataLoader,
     PageError,
     PageErrorHandler,
@@ -122,7 +123,7 @@ export const GroupPage: FunctionComponent<PageProperties> = () => {
             pleaseWait(false);
             appNavigation.navigateTo("/explore");
         }).catch(error => {
-            setPageError(toPageError(error, "Error deleting group."));
+            handleMutatingActionError(error, "Error deleting group.", setPageError, () => pleaseWait(false));
         });
     };
 
@@ -134,6 +135,8 @@ export const GroupPage: FunctionComponent<PageProperties> = () => {
             if (artifactDeleteSuccessCallback) {
                 artifactDeleteSuccessCallback();
             }
+        }).catch(error => {
+            handleMutatingActionError(error, "Error deleting artifact.", setPageError, () => pleaseWait(false));
         });
     };
 
--- a/ui/ui-app/src/app/pages/version/VersionPage.tsx
+++ b/ui/ui-app/src/app/pages/version/VersionPage.tsx
@@ -7,6 +7,7 @@ import {
     ContentTabContent,
     DocumentationTabContent,
     EXPLORE_PAGE_IDX,
+    handleMutatingActionError,
     PageDataLoader,
     PageError,
     PageErrorHandler,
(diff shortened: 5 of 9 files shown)


karanthengane23:
Great catch on resetting the `pleaseWait` modal across the entity pages!

One quick check: `RolesPage.tsx` appears to have the exact same issue in `updateRoleMapping` (L88), `createRoleMapping` (L108), and `onRevokeRoleMapping` (L119) where `pleaseWait(true)` is triggered but the `.catch()` blocks miss `pleaseWait(false, "")`. 

Might be worth including `RolesPage.tsx` in this PR as well so all action/delete modals across `ui-app` are consistently guarded.

ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx
```diff
@@ -192,6 +192,7 @@ export const ArtifactPage: FunctionComponent<PageProperties> = () => {
             pleaseWait(false, "");
             appNavigation.navigateTo("/explore");
         }).catch(error => {
+            pleaseWait(false);
```
paoloantinori: This bug class has now been fixed by hand twice, first on the Rules page in #9013 and now across five more pages, so it deserves a regression test rather than a manual checklist. Please add a page-level test that mocks the artifact delete call to reject and asserts that the please-wait state is cleared and the page error is set. If ui-app has no harness that can drive these pages yet, a follow-up issue for that harness is fine, but then leave the manual checkboxes unchecked until someone has actually run them.

ui/ui-app/src/app/pages/explore/ExplorePage.tsx
```diff
@@ -259,6 +259,7 @@ export const ExplorePage: FunctionComponent<ExplorePageProps> = () => {
             setPaging(resetPaging);
             search(criteria, resetPaging);
         }).catch(error => {
+            pleaseWait(false);
```
paoloantinori: The title says delete/action requests but the patch only touches delete handlers on the five pages named in #9408. Worth running a grep across ui-app for pleaseWait(true and checking every mutating action whose catch only calls setPageError, for example update artifact, create version, create branch, and anything on the imports page. Any site missed here keeps the modal blocking after a failed request in exactly the way this issue describes.

ui/ui-app/src/app/pages/group/GroupPage.tsx
```diff
@@ -134,6 +135,9 @@ export const GroupPage: FunctionComponent<PageProperties> = () => {
             if (artifactDeleteSuccessCallback) {
                 artifactDeleteSuccessCallback();
             }
+        }).catch(error => {
+            pleaseWait(false);
```
paoloantinori: Good that the failure path now exists at all. One dependency to double check: in the visible success path the modal is cleared only inside the optional artifactDeleteSuccessCallback, so a caller that deletes an artifact without passing a callback (assuming there is no unconditional pleaseWait(false) above the hunk context) would still hang the modal on success. Calling pleaseWait(false) unconditionally in this handler, the way the group-delete handler above does, removes that dependency and lets the callback keep handling refresh and navigation.

ui/ui-app/src/app/pages/branch/BranchPage.tsx
```diff
@@ -86,7 +86,8 @@ export const BranchPage: FunctionComponent<PageProperties> = () => {
             const aid: string = encodeURIComponent(artifactId as string);
             appNavigation.navigateTo(`/explore/${gid}/${aid}/branches`);
         }).catch(error => {
-            setPageError(toPageError(error, "Error deleting a version."));
+            pleaseWait(false);
+            setPageError(toPageError(error, "Error deleting a branch."));
```
paoloantinori: Fixing the copy from "Error deleting a version." to "Error deleting a branch." is correct for this handler and a good catch. It is a user-visible change beyond the pleaseWait additions listed in the summary though, so please mention it there so it does not read as an accidental edit during review.

ui/ui-app/src/app/pages/version/VersionPage.tsx
```diff
@@ -269,6 +269,7 @@ export const VersionPage: FunctionComponent<PageProperties> = () => {
             const aid: string = encodeURIComponent(artifactId as string);
             appNavigation.navigateTo(`/explore/${gid}/${aid}`);
         }).catch(error => {
+            pleaseWait(false);
```
paoloantinori: This is the eighth copy of the same two-line catch across five files, on top of the Rules-page fix in #9013, and nothing here prevents the next action added to a page from reintroducing the stuck modal. Consider a small shared helper next to toPageError that clears the wait state and sets the page error in one call, or clearing the wait state inside the shared error translation. Happy to see this land as-is with a follow-up issue if you want to keep the patch minimal.

paoloantinori (commented):
The fix is correct and consistent with the Rules-page pattern from #9013: every delete failure path now clears the blocking modal before the page error is set, and the pleaseWait(false) then setPageError ordering is right at all eight sites. What is missing is regression coverage and a sweep of the remaining call sites, not correctness in the lines added.

- All eight catch handlers dismiss the modal before rendering the error: ArtifactPage.tsx:195, ArtifactPage.tsx:266, ArtifactPage.tsx:326, BranchPage.tsx:89, ExplorePage.tsx:262, GroupPage.tsx:125, GroupPage.tsx:139, VersionPage.tsx:272.
- BranchPage.tsx:90 also changes user-visible copy from "Error deleting a version." to "Error deleting a branch.". That is correct for this handler, but the summary does not mention it and it reads as a drive-by edit unless called out.
- GroupPage.tsx:138 adds a catch where none existed, so artifact delete failures were previously an unhandled rejection on top of the stuck modal. The visible success path above it still clears the modal only through the optional artifactDeleteSuccessCallback.
- The test plan is manual checkboxes only. This bug class has now been fixed by hand twice (#9013 and here), which is usually the point where a regression test pays for itself.
- The same two-line catch is duplicated at eight sites across five files, so nothing stops the next action added to any page from reintroducing the bug.

ui/ui-app/src/app/pages/explore/ExplorePage.tsx
Mohammad-Raees: Thanks — I grepped `pleaseWait(true` across ui-app.

Create version/branch and the other non-delete mutating handlers on Artifact/Explore/Version/Group already cleared the wait overlay in their catch paths. The sweep did find missing clears on RolesPage grant/update/revoke and on Explore ZIP import (ProgressModal / `setImporting(false)`), and those are included now.

Roles full-page `setPageError` replacement remains tracked in #9232.

ui/ui-app/src/app/pages/version/VersionPage.tsx
Mohammad-Raees: Agreed — added shared `handleMutatingActionError` next to `toPageError` and switched the simple mutating-action catch sites over to it so clearing the wait overlay and setting the page error happen in one call.

ui/ui-app/src/app/pages/branch/BranchPage.tsx
Mohammad-Raees: Mentioned the BranchPage copy fix ("version" → "branch") in the PR summary.

ui/ui-app/src/app/pages/artifact/ArtifactPage.tsx
Mohammad-Raees: ui-app does not currently have a page-level RTL harness that can mount ArtifactPage and drive a mocked reject, so I opened follow-up #9584 for that harness + regression test. Manual checklist items are left unchecked. Added unit coverage for the shared helper in the meantime.

ui/ui-app/src/app/pages/group/GroupPage.tsx
Mohammad-Raees: Double-checked: `doDeleteArtifact` already calls `pleaseWait(false)` unconditionally in the success `.then()` before the optional callback, matching the group-delete handler. No change needed there.

paoloantinori:
Thanks — re-read at be2d6b0 and all the points are addressed: the shared `handleMutatingActionError` helper (including the ordering test), the RolesPage clears, the Explore ZIP import fix, the BranchPage copy correction, and the new GroupPage artifact-delete catch. Good catch on `doDeleteArtifact` already clearing the wait unconditionally in the success path — that matches what I see. With the page-level regression test tracked in #9584, nothing further from my side.

Mohammad-Raees:
Hi @EricWittmann — gentle nudge when you have a chance: Paolo has finished his review and this PR looks ready to merge from our side. No rush — thank you!

Mohammad-Raees:
Hi @carlesarnal — gentle bump when you have a chance. I rebased onto latest `main` so the branch is conflict-free again. Paolo's review feedback is addressed, local helper tests are green, and SonarCloud passed. Happy to fix anything if full CI turns up issues — otherwise ready whenever a maintainer can `/accept` / merge. Thank you!

Mohammad-Raees:
Hi @carlesarnal — gentle follow-up when you have a chance. I rebased onto latest `main` again (`9214dd0a7`) so the branch stays conflict-free. Quick Verify, Build UI, and SonarCloud are green on the latest push; Paolo's review feedback is already addressed. Happy to fix anything if full CI surfaces issues — otherwise ready whenever a maintainer can review/merge. Thank you!

malladinagarjuna2:
well @Mohammad-Raees 
i have gone through the changes, the issue is real and priority high, 
all the suggestions given by @paoloantinori  is applied well, 
just a small thing i have noticed in the tests
handleMutatingActionError.test.ts verifies that a function which calls two callbacks calls two callbacks. Well-written for what it is, but it exercises no page behavior,  it would pass identically if every .catch in this PR were reverted. Nothing protects GroupPage.doDeleteArtifact, the one site with a user-visible defect.
worth having it. 
waiting for @carlesarnal  to review

Mohammad-Raees:
Thanks @malladinagarjuna2 for taking a look — really appreciate the review and the validation that the fix addresses a real, high-priority issue.

Agree the helper test is unit-level only. Paolo raised the same point about page-level coverage; that is tracked in #9584 since ui-app does not yet have an RTL page harness for these flows. Happy to pick that up after this lands if maintainers want it in-repo sooner.

Thanks again!

Mohammad-Raees:
I'm stepping away from this contribution and won't be able to continue. Closing this PR so the issue is free for someone else to pick up.

Thanks to everyone who reviewed — especially @paoloantinori and @carlesarnal.

PR #9586, opened 2026-08-14
fix(limits): do not increment usage counters on dry-run requests

### LINKED ISSUE #9535: BUG: Do not increment storage limit counters for dry-run artifact creation
## Description

When an artifact or artifact version is created with `dryRun=true`, the storage operation correctly avoids persisting the artifact. However, the corresponding storage-limit counters are still incremented unconditionally.

In `RegistryStorageLimitsEnforcer`, `limitsService.artifactCreated()` is called after `delegate.createArtifact(...)` regardless of the value of `dryRun`. The same issue exists for `artifactVersionCreated(...)`.

This can cause the configured storage limits to diverge from the actual number of artifacts stored in the registry.

### Environment

Apicurio Registry running from the current `main`/`snapshot` build.

### Steps to Reproduce

1. Configure a registry with an artifact storage limit.
2. Record the current number of stored artifacts.
3. Perform artifact creation requests with `dryRun=true`.
4. Verify that the dry-run requests do not persist artifacts.
5. Perform enough dry-run requests to increase the internal artifact counter significantly.
6. Attempt to create a real artifact after the counter reaches the configured limit.

**Expected:**

A `dryRun` request should validate the artifact without changing storage-limit counters, because no artifact or artifact version is actually persisted.

**Actual:**

`limitsService.artifactCreated()` is invoked even when `dryRun=true`, causing the artifact counter to increase despite no artifact being created.

The same behavior occurs for artifact versions through `limitsService.artifactVersionCreated(...)`.

For example:

```text
Initial state:
50 stored artifacts / 100 limit

10 dry-run creations

Expected:
50 stored artifacts / 100 limit

Actual:
50 stored artifacts / 100 limit
but internal creation counter has increased to 60
```

After enough dry-run requests, legitimate artifact creation can be rejected because the configured limit is reached by the counter even though the corresponding artifacts do not exist.

### Relevant Code

`RegistryStorageLimitsEnforcer.createArtifact()` currently performs the counter update unconditionally after delegating the creation:

```java
delegate.createArtifact(..., dryRun, ...);
limitsService.artifactCreated();
```

A similar unconditional update exists in `createArtifactVersion()`.

### Suggested Fix

Only update the storage-limit counters when the operation actually creates persistent data:

```java
if (!dryRun) {
    limitsService.artifactCreated();
}
```

and similarly for artifact versions:

```java
if (!dryRun) {
    limitsService.artifactVersionCreated(groupId, artifactId);
}
```

A regression test should verify that `dryRun=true` does not change artifact or artifact-version limit counters.

hussainjamal760:
## Description

**What:**
This PR fixes a critical bug where `RegistryStorageLimitsEnforcer` incorrectly increments storage usage quotas (via `RegistryLimitsService`) even when a request is executed in `dryRun` mode. It also introduces missing unit tests for the limits enforcer to comprehensively verify its behavior.

**Why:**
Previously, when validating an artifact or artifact version creation using `dryRun = true` (common in CI/CD environments), the limit counter would incorrectly tick up despite no artifact actually being stored. This resulted in permanent storage quota exhaustion because `dryRun` artifacts aren't persisted, meaning users could not delete them to reclaim their quota. 

**Changes:**
- Wrapped `limitsService.artifactCreated()` and `limitsService.artifactVersionCreated()` calls in `if (!dryRun)` conditions.
- Completely refactored `RegistryStorageLimitsEnforcerTest` to use proper `ContentHandle` types instead of Strings.
- Implemented comprehensive positive and negative test cases for `RegistryStorageLimitsEnforcer`, including limit exceeded paths and edge cases combining limit violations with `dryRun = true`.
- Added strict Mockito verification using `InOrder` to guarantee that the limits checker is invoked prior to executing storage operations.

## Issue Reference
Fixes #9535 

## Checklist (for reviewer)
- [x] Code compiles and `mvnw checkstyle:check` passes.
- [x] Contains no unrelated changes.
- [x] Adds comprehensive tests for the fixed logic path.
- [x] Appropriate commit message (`fix(limits): do not increment usage counters on dry-run requests`) with DCO sign-off.


app/src/test/java/io/apicurio/registry/limits/RegistryStorageLimitsEnforcerTest.java
```diff
@@ -0,0 +1,193 @@
...
+
+        InOrder inOrder = Mockito.inOrder(limitsService, delegate);
+        inOrder.verify(limitsService).canCreateArtifactVersion("g1", "a1", null, contentHandle);
+        inOrder.verify(delegate).createArtifactVersion("g1", "a1", "1", "JSON", content, metaData,
+                Collections.emptyList(), false, false, "owner");
+        inOrder.verify(limitsService).artifactVersionCreated("g1", "a1");
+    }
+
+    @Test
+    public void testCreateArtifactVersionLimitExceededThrowsExceptionAndDoesNotInvokeDelegate() {
```
paoloantinori, line 180: The limit-exceeded path for `createArtifactVersion` is only tested with `dryRun=false` here. The artifact side has `testCreateArtifactLimitExceededWithDryRunTrueStillThrowsException` (line 107), and the PR description lists limit violations combined with `dryRun=true` as covered, but the version path has no equivalent. Please add `testCreateArtifactVersionLimitExceededWithDryRunTrueStillThrowsException`: stub `limitsService.canCreateArtifactVersion("g1", "a1", null, contentHandle)` to return `LimitsCheckResult.disallowed(...)`, call `enforcer.createArtifactVersion(..., dryRun=true, ...)`, assert `LimitExceededException` is thrown, and verify with `Mockito.never()` that `artifactVersionCreated("g1", "a1")` is not invoked. That pins the exact regression this PR fixes on the second method.

app/src/test/java/io/apicurio/registry/limits/RegistryStorageLimitsEnforcerTest.java
```diff
@@ -0,0 +1,193 @@
...
+    @Test
+    public void testCreateArtifactLimitExceededThrowsExceptionAndDoesNotInvokeDelegate() {
+        EditableArtifactMetaDataDto metaData = new EditableArtifactMetaDataDto();
+        ContentWrapperDto content = new ContentWrapperDto();
+        EditableVersionMetaDataDto versionMetaData = new EditableVersionMetaDataDto();
+
+        Mockito.when(limitsService.canCreateArtifact(metaData, content, versionMetaData))
+                .thenReturn(LimitsCheckResult.disallowed("Artifact limit reached", "ARTIFACT_LIMIT"));
+
+        LimitExceededException ex = assertThrows(LimitExceededException.class, () -> enforcer.createArtifact("g1", "a1", "JSON", metaData, "1", content, versionMetaData,
```
paoloantinori: This line is about 165 characters long, and lines 115 and 183 follow the same pattern. Please wrap the `assertThrows` lambda the same way the neighboring stubs are wrapped, so the `enforcer.createArtifact(...)` call sits on its own indented lines. The PR description says `mvnw checkstyle:check` passes, so a line-length rule may not be enforced, but the wrapped form is consistent with the rest of the file and much easier to review.

app/src/main/java/io/apicurio/registry/limits/RegistryStorageLimitsEnforcer.java
```diff
@@ -65,7 +65,9 @@ public Pair<ArtifactMetaDataDto, ArtifactVersionMetaDataDto> createArtifact(Stri
                 .execute(() -> delegate.createArtifact(groupId, artifactId, artifactType, artifactMetaData,
                         version, versionContent, versionMetaData, versionBranches, versionIsDraft, dryRun,
                         owner));
-        limitsService.artifactCreated();
+        if (!dryRun) {
+            limitsService.artifactCreated();
```
paoloantinori, line 69: Question on the non-dry-run path, no change requested in this PR: when `createArtifact` is called with `dryRun=false` for an artifactId and content that already exist, the delegate returns the existing metadata without creating anything new, yet `artifactCreated()` still fires. If `artifactCreated()` increments a counter rather than triggering a resync from storage, that path over-counts, which is the same class of drift this PR removes on the dry-run side. Could you confirm the semantics of `artifactCreated()` and `artifactVersionCreated()`, and if they increment, file a follow-up issue for the existing-content path?

paoloantinori (commented):
The core fix is correct and minimal: gating `limitsService.artifactCreated()` and `limitsService.artifactVersionCreated()` on `!dryRun` matches storage semantics, since a dry-run create persists nothing, and the new tests pin the check-before-execute ordering with `InOrder` plus `never()` assertions on the counters. No blockers; the remaining points are one test-coverage gap, two counter-semantics questions worth confirming, and a formatting cleanup.

- `app/src/main/java/io/apicurio/registry/limits/RegistryStorageLimitsEnforcer.java:68-70` and `:82-84`: both create paths get the same `!dryRun` guard, which is the right altitude for the fix. The existing non-dry-run behavior is unchanged.
- `RegistryStorageLimitsEnforcerTest.java:174`: the limit-exceeded plus `dryRun=true` combination is only tested for `createArtifact` (line 107); the symmetric `createArtifactVersion` case is missing even though the PR description lists those combinations as covered. Details inline.
- `RegistryStorageLimitsEnforcerTest.java:95` (also 115 and 183): the `assertThrows(...)` lines run to roughly 165 characters. Wrapping them keeps the file consistent and reviewable; details inline.
- `RegistryStorageLimitsEnforcer.java:69`: worth confirming the semantics of `artifactCreated()` on the non-dry-run path when the create resolves to an existing artifact. If the method increments a counter rather than resyncing from storage, the same class of drift this PR fixes still exists there. Question inline.
- Counter healing on upgrade: deployments that served dry-run traffic before this fix carry inflated usage counters. If those counters survive a restart or are cached with a long TTL, the fix alone does not restore quota for affected instances, so please confirm whether a resync or reset happens on startup or whether an upgrade note is needed.
- Minor, optional: the fixture setup (`metaData`, `content`, `versionMetaData`) is repeated verbatim in all seven tests; extracting a private helper would remove most of the duplication.

app/src/test/java/io/apicurio/registry/limits/RegistryStorageLimitsEnforcerTest.java
hussainjamal760: Missing test coverage for `createArtifactVersion` + `dryRun=true` limit exceeded
Added `testCreateArtifactVersionLimitExceededWithDryRunTrueStillThrowsException` in `RegistryStorageLimitsEnforcerTest`. It verifies that when `canCreateArtifactVersion` returns `disallowed`, a `LimitExceededException` is thrown, the delegate is never invoked, and `limitsService.artifactVersionCreated` is never called.

app/src/test/java/io/apicurio/registry/limits/RegistryStorageLimitsEnforcerTest.java
hussainjamal760: Wrapped the `assertThrows` invocations in `RegistryStorageLimitsEnforcerTest` so enforcer calls sit cleanly on indented lines.

app/src/main/java/io/apicurio/registry/limits/RegistryStorageLimitsEnforcer.java
hussainjamal760: Non-dry-run path duplicate/existing artifact over-counting question
Good call-out. `limitsService.artifactCreated()` increments the Guava `LoadingCache` counter held in `StorageMetricsStore`.

hussainjamal760:
Thanks for the thorough review, @paoloantinori!

However, `StorageMetricsStore` counters are ephemeral with a default expiration window (`apicurio.storage.metrics.cache.check-period.ms=30000`, 30 seconds). Upon eviction, or when an un-cached key is fetched, `StorageMetricsStore` queries the database directly (`storage.countArtifacts()`, `storage.countTotalArtifactVersions()`, `storage.countArtifactVersions(...)`) to resync metrics to reality.

So while a duplicate/idempotent create call on `dryRun=false` could theoretically over-increment the in-memory counter transiently until the 30s expiration cycle, it self-heals shortly thereafter. I've created a follow-up tracker issue for us to make `artifactCreated()` check insertion outcome or trigger immediate invalidation on existing content to eliminate even that 30s drift window.

Counter healing on upgrade
Because `StorageMetricsStore` relies on Guava `LoadingCache` backed by live DB count queries (`storage.countTotalArtifactVersions()`, `storage.countArtifacts()`), counters are **not** persisted to disk or external cache. They automatically resync from the underlying SQL storage on node startup and every 30 seconds during runtime. Deployments upgrading to this version will automatically heal any inflated counters on standard cache expiration/restart with zero manual intervention required.

paoloantinori:
Thanks — that resolves both of my counter-semantics questions. Ephemeral counters with the 30s resync window, plus the DB recount on cache miss and on restart, cover both the transient over-increment on idempotent non-dry-run creates and the upgrade-healing case, so nothing beyond the fix itself is needed there. Good to see the follow-up tracker for the invalidation-on-existing-content refinement as well. No further concerns from my side.

hussainjamal760:
@carlesarnal Just a soft reminder to review this PR whenever you get a chance, so I can keep the work moving. I really respect your work and completely understand you’re busy, especially since you’ve just returned from vacation. Thanks!

EricWittmann (changes requested):
The production fix is correct and minimal — gating the counter increments on `!dryRun` is the right approach.

However, the test file won't compile because `LimitsCheckResult.disallowed()` is called with two arguments in four places, but the method only accepts a single `String`:

```java
// Actual signature (LimitsCheckResult.java:8)
public static LimitsCheckResult disallowed(String message)

// What the test calls (4 occurrences)
LimitsCheckResult.disallowed("Artifact limit reached", "ARTIFACT_LIMIT")
LimitsCheckResult.disallowed("Artifact version limit reached", "VERSION_LIMIT")
```

This is what's causing the CI build failure. Please remove the second argument from all four call sites.

hussainjamal760:
Thanks for the review @EricWittmann !

1. **Fixed `disallowed()` Signature & Line Formatting**: Updated all call sites in `RegistryStorageLimitsEnforcerTest` to use the single-arg `LimitsCheckResult.disallowed(String)` signature and wrapped `assertThrows` blocks.

EricWittmann:
Thanks for the quick fix on the `disallowed()` calls!

One small note: commit `001def3` is missing the DCO `Signed-off-by` trailer. I'll merge this anyway once CI is green, but please try to remember to sign off on all commits going forward (`git commit -s`).

hussainjamal760:
> Thanks for the quick fix on the `disallowed()` calls!
> 
> One small note: commit `001def3` is missing the DCO `Signed-off-by` trailer. I'll merge this anyway once CI is green, but please try to remember to sign off on all commits going forward (`git commit -s`).

Thanks for pointing that out! Glad the fix is sorted. I’ll make sure to use "git commit -s" and include the DCO "Signed-off-by" trailer on all commits going forward. Appreciate the heads-up!

EricWittmann (approved):
LGTM! The production fix is correct and minimal, and the tests cover the important cases well. Thanks for the contribution.

hussainjamal760:
> LGTM! The production fix is correct and minimal, and the tests cover the important cases well. Thanks for the contribution.

Thanks for taking the time to review it! Really appreciate the feedback.

EricWittmann:
Thanks for this contribution, @hussainjamal760! Great catch on the dry-run counter bug — the fix is clean and the tests cover the important cases well. Appreciate you sticking with it through the review feedback. 👍

